Issue #2831, opened 2016-06-09
Redesign The Effects Plugins


RebeccaLaVie:
It has been requested by some [here](https://github.com/LMMS/lmms/pull/2806#issuecomment-224297891) that all of the Effects Plugins are redesigned to fit with the new theme. There are a lot of plugins and some people have already started collaborating with me on this. Bellow, I have listed all of the elements and design standards that will be needed to design matching plugins that look consistent. I think it's important that if someone would like to contribute a design for one of the plugins that they say so in the comments bellow so that two people don't redesign the same plugin.

**FONTS**
The font that I am using is Orbitron Bold in ALL CAPS with 1.50 tracking in 11px. The font (and preferably all elements) have an 8px space around them. For best results, export your final SVG as a Cairo .png (I prefer to use the Save A Copy function here).

If the Cairo .png leaves some of the elements looking a bit to crisp, you can always export the text with Cairo and the background with Inkscape. Use the "Color to Alpha" option in GIMP to cut out the bg and then place it on the background. This is only necessary if you are creating a more complex graphic.

**BACKGROUND**
Let's try to keep backgrounds that are going to be flush with the Title Bars the same as the "Plugin BG" background in my source files.

**GRADIENTS**
Please note that your eyes follow the direction that a gradient is facing. For example, the gradient that I used in the LFO controller was linear so that your eyes would be drawn from the title down to the content of the plugin. If the gradient pointing at an odd angle, the plugin will be a bit less intuitive for the user. Gradients should be subtle when possible. This is not exactly a rule, but because many prefer flat design and many others don't I try to keep it somewhere in between.

**KNOBS**
The knobs were designed to look slightly transparent in the center, but they actually a solid color. This is just the green with an overlay of the background color to give it the same transparent look, while making sure that the knobs are perfectly visible in all parts of the program.

**BUTTONS**
You will find various buttons in my .SVG source files. Please reference the SENDS buttons. I usually just create pixel perfect text from scratch because it doesn't take me very much time but if you want to use a font, just make sure that you are using good hinting when you export. Inactive buttons are lighter with white text, active buttons are darker with green text. Again, you should easily be able to resize the buttons in my source files so don't worry about creating that from scratch.

**SLIDERS**
Please see the Faders Knobs in my source files for a reference.

**OTHER BACKGROUNDS**
[...]

PR #3093, opened 2016-10-27
Redesign dual filter (and enable theme based styling of effect control dialogs)

simonvanderveldt:
Fixes part of #2831

image

This uses proper layouts and CSS styling from the theme to layout and style the Dual Filter control dialog instead of hardcoded pixel values and pixmaps.

I had to change a relatively large amount of files to enable the styling using the theme/CSS for the plugin/effect control dialogs. I believe passing the parent is the correct way to do it, but please let me know if that's not the case.

P.S. The whitespace seems to a bit messed up, just noticed there's a [coding conventions page](https://github.com/LMMS/lmms/wiki/Coding-conventions), I'll fix those up after/when I get some feedback.

### DIFF
--- a/data/themes/default/style.css
+++ b/data/themes/default/style.css
@@ -854,6 +854,11 @@ NesInstrumentView Knob {
 	qproperty-lineWidth: 2;
 }
 
+DualFilterControlDialog  {
+	background: red;
+  color: blue;
+}
+
 /* palette information  */
 
 LmmsPalette {
--- a/data/themes/default/style.css
+++ b/data/themes/default/style.css
@@ -349,11 +349,13 @@ EffectSelectDialog QScrollArea {
 /* the inner boxes in LADSPA effect windows */
 
 EffectControlDialog QGroupBox {
-	background: #262b30;
-	margin-top: 1ex;
-	padding: 10px 2px 1px;
-	border-radius: 4px;
-	border: none;
+  background: qlineargradient( x1:0 y1:0, x2:0 y2:1, stop:0 #22282e, stop:1 #1c202a);
+	padding-top: 1ex;
+  border: none;
+  border-top-style: solid;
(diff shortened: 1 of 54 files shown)


RebeccaDeField:
Aesthetically, I think this looks good :+1:

@Mentioning some people that might have some more feedback for you...
@Umcaruje @simonvanderveldt @BaraMGB

simonvanderveldt:
Seems like either I messed someting up or it was already broken with regards to the `VstEffectControls`. For some reason `VstEffect` doesn't seem to build for me even when I run a `build all`. I don't see a build target for `VstEffect` either. Does anyone know how I can build it?

BaraMGB:
I don't think that is the right approach here. In the effect plugins we don't use the css. You should reverse all that and only redesign the plugin by code. 

If you want to do it with QLayout feel free. But I made the experience that QLayout the code blow up with no benefits. Because effect plugins can't be resized by the user. Move() is just okay for this.

Umcaruje:
@simonvanderveldt the build is failing on this, plus such a change to CSS based styling is not something I think we should do for 1.2.

Can you please split this into 2 different PR's?  Do a plugin redesign in inkscape like we've been doing so far to be consistent, and then have a seperate PR that does this for all plugins and then we can have this discussion there.

simonvanderveldt:
> @simonvanderveldt the build is failing on this, plus such a change to CSS based styling is not something I think we should do for 1.2.
> 
> Can you please split this into 2 different PR's? Do a plugin redesign in inkscape like we've been doing so far to be consistent, and then have a seperate PR that does this for all plugins and then we can have this discussion there.

@Umcaruje thanks for the feedback. I don't have experience with Inkscape + I've just moved to a new house and have a lot of work to do there, so I guess someone else should pick this one up.

PR #3094, opened 2016-10-27
Replace int by int32_t in aeffectx.h structures

moleculext:
Replace int by int32_t in aeffectx.h structures to ensure that structure's sizes does not vary with platform or compiler. Size of int is not guaranteed to always be 4 bytes.

Moreover, the required header <stdint.h> should be included at the beginning of this file because of int32_t reference. This change is not included in this pull request.

Thanks,
François

### DIFF
--- a/include/aeffectx.h
+++ b/include/aeffectx.h
@@ -26,10 +26,10 @@
 #ifndef AEFFECTX_H
 #define AEFFECTX_H
 
-#define CCONST(a, b, c, d)( ( ( (int) a ) << 24 ) |		\
-				( ( (int) b ) << 16 ) |		\
-				( ( (int) c ) << 8 ) |		\
-				( ( (int) d ) << 0 ) )
+#define CCONST(a, b, c, d)( ( ( (int32_t) a ) << 24 ) |		\
+				( ( (int32_t) b ) << 16 ) |		\
+				( ( (int32_t) c ) << 8 ) |		\
+				( ( (int32_t) d ) << 0 ) )
 
 const int audioMasterAutomate = 0;
 const int audioMasterVersion = 1;
@@ -146,17 +146,17 @@ class VstMidiEvent
 {
 public:
 	// 00
-	int type;
+	int32_t type;
 	// 04
-	int byteSize;
+	int32_t byteSize;
 	// 08
-	int deltaFrames;
+	int32_t deltaFrames;
 	// 0c?
-	int flags;
+	int32_t flags;
 	// 10?
-	int noteLength;
+	int32_t noteLength;
 	// 14?
-	int noteOffset;
+	int32_t noteOffset;
 	// 18
 	char midiData[4];
 	// 1c?
@@ -186,7 +186,7 @@ class VstEvents
 {
 public:
 	// 00
-	int numEvents;
+	int32_t numEvents;
 	// 04
 	void *reserved;
 	// 08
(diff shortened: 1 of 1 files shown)


grejppi:
> This change is not included in this pull request.

Why? 😮

moleculext:
> Why? :open_mouth:

The reverse-engineering comments state that the size of the "int" member must be 4 bytes:
// 00
int type;
// 04
int byteSize;
// 08
int deltaFrames;

And the macro `CCONST` would not work with longer or shorter integer type.

This is modification I'm doing for [VST Preset Generator](http://vst-preset-generator.org) project, and I felt that it would be great to forward it to the upstream file.

(By the way thanks for this amazing work at providing an open-source alternative to Steinberg SDK!)

liushuyu:
Well… so do anyone else can confirm this change is good enough to merge?

tresf:
> Well… so do anyone else can confirm this change is good enough to merge?

The OP answered the wrong question.  @grejppi specifically asked why a header include wasn't included and the OP never answered (although the OP did however give a good explanation justifying the change).

moleculext:
@tresf @grejppi sorry, I misread your comment. The missing header is now committed at 5ebcf19!

jasp00:
The patch looks harmless, but what does it actually fix? This interface was meant for 32-bit x86. There are pointers to x86 functions. Size of pointers is assumed to be 4 bytes; by the way, I am amazed that this stuff works on 64-bit Windows. We should keep compatibility with binary plug-ins made with Steinberg's SDK. If Steinberg supported a new platform where int were 8 bytes, we would know whether to use int or int32_t. If the new platform were not supported and it could not execute x86 code, why would int32_t be better than int?

So, should this be merged? I am not sure. How should a new platform behave?

tresf:
> but what does it actually fix? [...] I am amazed that this stuff works on 64-bit Windows. 

Shouldn't that be reason enough to merge it or am I missing something here?

moleculext:
@jasp00 This patch is not about pointer size.  Function pointers are always referenced by `intprt_t` and as far as I can check on 64bit platform (Linux and Windows), the pointer function size is right. So this header works with 64bit plugins.

However, I agree that we must check binary VST plugin on platform where int size is not 4 bytes and check that the structure sizes like AEffect is the same.
According to [this article](https://en.wikipedia.org/wiki/64-bit_computing#64-bit_data_models), such platforms are not mainstream (HAL, SPARC64, UNICOS,...) and I can't check VST plugins on them. 

For the CCONST macro, the input must be int32_t because of the bit-shift.

I'm not active LMMS developper (yet?), and this contribution is only to give back and thank you for your work. So feel free to accept or reject this pull request.

Best,
François

jasp00:
>I am amazed that this stuff works on 64-bit Windows.

I meant that (I thought) the original header worked on 64-bit Windows without modification, because pointers change from 4 to 8 bytes and I never tested that. I was wrong, since tobydox did 65a0313807e8a62315ec412657bbd865c69643a9 to fix that environment. Please ignore my comment.

>I'm not active LMMS developper (yet?), and this contribution is only to give back and thank you for your work. 

You are welcome. We need patches like this and you are more active than me right now. I am just trying to understand what you are trying to accomplish.

Why would SPARC64 want to use 32-bit integers rather than 64-bit ones? Is there any binary plug-in for SPARC64 compiled with Steinberg's SDK?

zonkmachine:
Can't we leave the testing to the users in the next RC? We would get notified quite rapidly if anything changed on the VST side.  :smiley:

jasp00:
>Can't we leave the testing to the users in the next RC?

While this patch would fix some environments if we are wrong, it would break those environments if we are right. We have no information (e.g., binary plug-in for SPARC64 compiled with Steinberg's SDK) to know whether this patch must be applied.

`<stdint.h>` should be included, that is correct.

tresf:
As @zonkmachine has mentioned, SPARK64 users can open bug reports as they wish.  This architecture is now very rare and even more rare on the end-user desktop.

Easy enough to revert.  Merging.

PR #2672, opened 2016-03-13
Render the mixer levels in a dB FS scale

michaelgregorius:
Adds the option to render the mixer levels in dB FS. By default this option is disabled so that classes which inherit from `Fader` are not affected by this change. However, in the code of the `FxMixerView` this feature is enabled so that the mixer shows the levels in dB FS.

The ability to render in dB FS is exported as a property so that it can also be set in a style sheet (not used as of now). The new property is called `levelsDisplayedInDBFS`.

Showing the levels in dB FS (which is a logarithmic scale) gives a less "fidgety" impression when the levels are moving. This can be seen in the following movies, especially when channel "FX 1" goes into clipping towards the end of the movies.

Linear levels | Levels in dB FS
------------ | -------------
lmms-linearmixer | lmms-dbmixer
(-40 dB, 0.8 dB) | (-40 dB, 12 dB)

The new mixer shows a range of (-40 dB, 12 dB). This provides the user with more insight into how much a channel goes into clipping. The previous implementation corresponded to a range of (-40 dB, 0.8 dB) which pretty much only showed you that you are clipping. This is also best illustrated towards the end of the movies. The linear levels just bounce against the ceiling when clipping while the logarithmic levels have still some headroom.

PR #3095, opened 2016-10-28
dBV is actually mislabeled dBFS

Umcaruje:
So, after some looking around, I realised LMMS already used dBFS values but labeled them as dBV. according to wikipedia, dBV is a decibel that's relative to 1V of voltage, which does not quite fit in software. 

dBFS on the other hand is relative to the maximum value of a digital full scale, which in our case is 1 in the linear scale, so the conclusion is that we are indeed actually using dBFS in LMMS.

Also @michaelgregorius also came to this conclusion in #2672 but probably never got around to this rename job.

### DIFF
--- a/src/core/ConfigManager.cpp
+++ b/src/core/ConfigManager.cpp
@@ -154,7 +154,7 @@ void ConfigManager::upgrade()
 	{
 		return;
 	}
-	
+
 	ProjectVersion createdWith = m_version;
 	
 	if ( createdWith.setCompareType(ProjectVersion::Build) < "1.1.90" )
--- a/src/core/ConfigManager.cpp
+++ b/src/core/ConfigManager.cpp
@@ -136,6 +136,16 @@ void ConfigManager::upgrade_1_1_90()
 	}
 }
 
+	
+void ConfigManager::upgrade_1_1_91()
+{		
+	// rename displaydbv to displaydbfs
+	if ( !value( "app", "displaydbv" ).isNull() ) {
+		setValue( "app", "displaydbfs", value( "app", "displaydbv" ) );
+		deleteValue( "app", "displaydbv" );
+	}
+}
+
 
 void ConfigManager::upgrade()
 {
@@ -144,12 +154,17 @@ void ConfigManager::upgrade()
 	{
 		return;
 	}
-
+	
 	ProjectVersion createdWith = m_version;
 	
 	if ( createdWith.setCompareType(ProjectVersion::Build) < "1.1.90" )
 	{
 		upgrade_1_1_90();
+	}
+	
+	if ( createdWith.setCompareType(ProjectVersion::Build) < "1.1.91" )
+	{
+		upgrade_1_1_91();
 	}
 	
 	// Don't use old themes as they break the UI (i.e. 0.4 != 1.0, etc)
@@ -546,5 +561,3 @@ void ConfigManager::saveConfigFile()
 	outfile.write( xml.toUtf8() );
 	outfile.close();
 }
-
-
--- a/src/gui/MainWindow.cpp
+++ b/src/gui/MainWindow.cpp
@@ -1200,10 +1200,10 @@ void MainWindow::updateViewMenu()
 	// that is safe to change on the fly. There is probably some
 	// more elegant way to do this.
 	QAction *qa;
-	qa = new QAction(tr( "Volume as dBV" ), this);
-	qa->setData("displaydbv");
+	qa = new QAction(tr( "Volume as dBFS" ), this);
+	qa->setData("displaydbfs");
 	qa->setCheckable( true );
-	qa->setChecked( ConfigManager::inst()->value( "app", "displaydbv" ).toInt() );
+	qa->setChecked( ConfigManager::inst()->value( "app", "displaydbfs" ).toInt() );
 	m_viewMenu->addAction(qa);
 
 	// Maybe this is impossible?
@@ -1244,9 +1244,9 @@ void MainWindow::updateConfig( QAction * _who )
 	QString tag = _who->data().toString();
 	bool checked = _who->isChecked();
 
-	if( tag == "displaydbv" )
+	if( tag == "displaydbfs" )
 	{
-		ConfigManager::inst()->setValue( "app", "displaydbv",
+		ConfigManager::inst()->setValue( "app", "displaydbfs",
 						 QString::number(checked) );
 	}
 	else if ( tag == "tooltips" )
--- a/src/gui/widgets/Fader.cpp
+++ b/src/gui/widgets/Fader.cpp
@@ -321,9 +321,9 @@ void Fader::setPeak_R( float fPeak )
 // update tooltip showing value and adjust position while changing fader value
 void Fader::updateTextFloat()
 {
-	if( ConfigManager::inst()->value( "app", "displaydbv" ).toInt() && m_displayConversion )
+	if( ConfigManager::inst()->value( "app", "displaydbfs" ).toInt() && m_displayConversion )
 	{
-		s_textFloat->setText( QString("Volume: %1 dBV").
+		s_textFloat->setText( QString("Volume: %1 dBFS").
 				arg( 20.0 * log10( model()->value() ), 3, 'f', 2 ) );
 	}
 	else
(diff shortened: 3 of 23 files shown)


BaraMGB:
Agreed!

src/gui/MainWindow.cpp
```diff
@@ -1200,10 +1200,10 @@ void MainWindow::updateViewMenu()
 	// that is safe to change on the fly. There is probably some
 	// more elegant way to do this.
 	QAction *qa;
-	qa = new QAction(tr( "Volume as dBV" ), this);
-	qa->setData("displaydbv");
+	qa = new QAction(tr( "Volume as dBFS" ), this);
+	qa->setData("displaydbfs");
 	qa->setCheckable( true );
-	qa->setChecked( ConfigManager::inst()->value( "app", "displaydbv" ).toInt() );
+	qa->setChecked( ConfigManager::inst()->value( "app", "displaydbfs" ).toInt() );
```
tresf: Is this going to need an `upgrade()`?

Umcaruje: Do we even do upgrades for `.lmmsrc.xml`? If so, yes I can write an upgrade.

tresf: Albeit rather recently, [yes we do](https://github.com/LMMS/lmms/blob/master/src/core/ConfigManager.cpp#L140:L163).

Umcaruje: Ok, I'll look into it.

Umcaruje: @tresf done https://github.com/LMMS/lmms/pull/3095/commits/5a7acd322095b5a124fe349b008031e90004dceb

tresf: 👍

tresf:
I don't know enough about dbfs to merge, but otherwise the PR is good to merge.  I did notice some inconsistent whitespace formatting that this commit introduces.  We should probably decide eventually which one of these is correct, but this PR isn't the place for that decision.

```diff
     if (foo) {
           // foo
+          whitepsace option 1
-whitespace option 2
           // bar
     }
```

BaraMGB:
For what I've learned in Wikipedia dB FS is the unit used by all digital audio signals. For analog signals is dbv used (voltage).

Umcaruje:
@tresf please comment on the code where you see that issue. This is a simple find/replace job, and I didn't touch the whitespace.

src/core/ConfigManager.cpp
```diff
@@ -144,14 +154,19 @@ void ConfigManager::upgrade()
 	{
 		return;
 	}
-
+
```
tresf: whitespace

Umcaruje: Oh, ok, I accidentaly added whitespace there. I'll remove it.

src/gui/widgets/Fader.cpp
```diff
@@ -400,12 +400,12 @@ void Fader::paintDBFSLevels(QPaintEvent * ev, QPainter & painter)
...
 
 	// Draw right levels
-	float const rightSpan = ampToDbv(m_fPeakValue_R) - minDB;
+	float const rightSpan = ampToDbfs(m_fPeakValue_R) - minDB;
 	int peak_R = height * rightSpan * fullSpanReciprocal;
 	QRect const drawRectR( center, height - peak_R, width, peak_R ); // Source and target are identical
 	painter.drawPixmap( drawRectR, *m_leds, drawRectR );
 
-	float const persistentRightPeakDBFS = ampToDbv(m_persistentPeak_R);
+	float const persistentRightPeakDBFS = ampToDbfs(m_persistentPeak_R);
```
zonkmachine: `ampToDbfs( qMax<float>( 0.0001, m_persistentPeak_R ) )` ?
@Umcaruje ampToDbfs( [Must_be_greater_than_zero!](https://github.com/LMMS/lmms/blob/master/include/lmms_math.h#L268:L274) ) 
I'm debugging with a [new secret weapon](https://github.com/LMMS/lmms/issues/3408#issuecomment-284804410) and you've got a floating point error here. I triggered it by just starting lmms in gdb with @michaelgregorius super hack and when it loaded I just launched the Empty template. Boom!  Stuck the qMax code in there and the issue went away. Care to take a look at this?

```gdb
0x00007ffff47d51ac in feraiseexcept (__excepts=4) at ../sysdeps/x86/fpu/bits/fenv.h:135
135	../sysdeps/x86/fpu/bits/fenv.h: No such file or directory.
(gdb) bt
#0  0x00007ffff47d51ac in feraiseexcept (__excepts=4) at ../sysdeps/x86/fpu/bits/fenv.h:135
#1  __log10f (x=0) at w_log10f.c:32
#2  0x00000000005ec737 in ampToDbfs (amp=0) at /home/zonkmachine/builds/lmms/lmms/include/lmms_math.h:273
#3  0x00000000005ee1eb in Fader::paintDBFSLevels (this=0x1650ef0, ev=0x7fffffffb5f0, painter=...)
    at /home/zonkmachine/builds/lmms/lmms/src/gui/widgets/Fader.cpp:409
```

Umcaruje: We should use [`safeAmpToDbfs`](https://github.com/LMMS/lmms/blob/master/include/lmms_math.h#L249) here then? That has built in protection for zeores. I'll take a more thorough look soon.

unfa:
I think that in the FX Mixer, it'd make more sense to label the fader positions using "dB" units not "dBFS", because then +6 dBFS sounds weird. +6dB is perfectly fine though.

src/gui/widgets/Fader.cpp
zonkmachine: > We should use safeAmpToDbfs here then?

Yes. This looks a bit odd however: https://github.com/LMMS/lmms/blob/master/include/lmms_math.h#L251
` amp == 0.0f` There is no test here for negative input. I'm not up to date on the math here but this reference hints at an error for negative arguments.
http://www.cplusplus.com/reference/cmath/log10/

>  If x is negative, it causes a domain error.

Umcaruje: Yes, but we never use negative values for linear amplitude though, so it shouldn't really be even considered. *scratches head* and the variables are initialized with the value of zero: https://github.com/Umcaruje/lmms/blob/e3aad60661b9581f12e3d91bfc2158618e1ca571/src/gui/widgets/Fader.cpp#L73-L74

Though you loaded the Empty template, and there are no instruments there. Does this happen with the default template?

zonkmachine: I'm failing to reproduce that right now as I just get the crash right away. I think I might have mistaken and just thought I managed to launch something while it had already closed down business. Anyway.
Line 387 reads: `float const persistentLeftPeakDBFS = ampToDbfs(m_persistentPeak_L);`
The `bt full` readings for line 387 reads:


```
#3  0x00000000005ee070 in Fader::paintDBFSLevels (this=0x16556b0, ev=0x7fffffffbb90, painter=...)
    at /home/zonkmachine/builds/lmms/lmms/src/gui/widgets/Fader.cpp:387
        peak_L = 0
        persistentLeftPeakDBFS = -nan(0x7fbba4)
        width = 11
        center = 12
        maxDB = 9
        leftSpan = 0
        rightSpan = 4.59163468e-41
        drawRectR = {x1 = 0, y1 = 0, x2 = 0, y2 = 0}
        persistentPeak_R = 32767
        peak_R = -18464
        persistentRightPeakDBFS = 4.59163468e-41
        height = 116
        minDB = -42.0000038
        fullSpanReciprocal = 0.019607842
        drawRectL = {x1 = 0, y1 = 116, x2 = 10, y2 = 115}
        persistentPeak_L = -17520
```

Umcaruje: >persistentPeak_L = -17520

That's super weird...

zonkmachine: `int persistentPeak_L = height * (1 - (persistentLeftPeakDBFS - minDB) ...`

It's overflowing. 
`unsigned int persistentPeak_L ` seem to be a fix. Not tested this a lot though.

zonkmachine: If I start up and press play directly it's cool. If I wait a couple of seconds and do nothing it crashes.

michaelgregorius: Shouldn't we rather open a new issue for this one?

zonkmachine: Yes, you're right.

zonkmachine: Continued here: https://github.com/LMMS/lmms/issues/3420

PR #3110, opened 2016-11-09
Changed the arpeggio functions: InstrumentFunction.h and ccp

Rikislav:
As I am an enthusiast user of the arpeggio function wanted to add some functionality, like single note volume and pan.

Changed the ChordTable structure into a QVector, the int8 key to a new structure etc.

The chords are still hardcoded but the final objective is to build up a dedicated sequence editor and put data into a file

### DIFF
--- a/data/locale/it.ts
+++ b/data/locale/it.ts
@@ -3440,8 +3440,8 @@ Puoi rimuovere e muovere i canali con il menù contestuale, cliccando con il tas
         <translation>Modalità logaritmica</translation>
     </message>
     <message>
-        <source>Please enter a new value between -96.0 dBV and 6.0 dBV:</source>
-        <translation>Inserire un nuovo valore tra -96.0 dBV e 6.0 dBV:</translation>
+        <source>Please enter a new value between -96.0 dBFS and 6.0 dBFS:</source>
+        <translation>Inserire un nuovo valore tra -96.0 dBFS e 6.0 dBFS:</translation>
     </message>
     <message>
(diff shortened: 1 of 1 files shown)


Rikislav:
Thanx! I really need as much input as possible!!!

2016-11-09 20:34 GMT+01:00 Oskar Wallgren <notifications@github.com>:

CMakeLists.txt
```diff
@@ -401,9 +401,9 @@ IF(LMMS_BUILD_OPENBSD)
 	FIND_PACKAGE(Sndio)
 	IF(SNDIO_FOUND)
 		SET(LMMS_HAVE_SNDIO TRUE)
-        	SET(STATUS_SNDIO "OK")
-    	ELSE()
-        	SET(STATUS_SNDIO "<not found or not supported on this platform>")
+					SET(STATUS_SNDIO "OK")
+			ELSE()
+					SET(STATUS_SNDIO "<not found or not supported on this platform>")
```
zonkmachine: Too many tabs here (line 404-406). If you press `view` in the top of this file you can see how it looks with the GitHub formatting. I think you might have the tabs in your editor set to a low value.

Rikislav:
Thanks for the input, I am fixing things and formatting the code according to standard (was using a beautifier).

My pull request was probably premature, it compiled it on my linux machine but I didn't check it deep (I  see now I still have a long way to go...).

I think the idea is good tough, and when the things get right I'll try to pull another request!

PR #3111, opened 2016-11-10
Fix several i18n-related defects

liushuyu:
Fix the problem that some translations unable to render to the interface.
Mostly caused by lacking `Q_OBJECT` and the order of multiple inheritance.

Add translation support for `MidiJACK`

---
This change is [Reviewable](https://reviewable.io/reviews/lmms/lmms/3111)

### DIFF
--- a/data/locale/en.ts
+++ b/data/locale/en.ts
@@ -4088,6 +4088,19 @@ Please visit http://lmms.sf.net/wiki for documentation on LMMS.</source>
     </message>
 </context>
 <context>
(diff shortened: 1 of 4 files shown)


liushuyu:
Before you ask...
- Why `//:` comments? 
These comments will also be extracted by  Qt's`lupdate` and add to comment section of the translation file

liushuyu:
If there's no objections, I will merge this in 16 hours.

liushuyu:
Merged #3111

PR #3114, opened 2016-11-11
Disable MIDI export temporarily

Umcaruje:
So we can move on and release the RC2. If its necessary, we even release 1.2 without midi export.

This, if it gets merged should take the 'critical' badge off of #2384 

Another option is to make midi export an easter egg, as a value in `.lmmsrc.xml` that you would edit to get experimental midi export

### DIFF
--- a/plugins/CMakeLists.txt
+++ b/plugins/CMakeLists.txt
@@ -49,7 +49,7 @@ IF("${PLUGIN_LIST}" STREQUAL "")
 		LadspaEffect
 		lb302
 		MidiImport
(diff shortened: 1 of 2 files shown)


zonkmachine:
> So we can move on and release the RC2

:+1:

Umcaruje:
I'm going to merge this in 24 hours. This is an easily revertable change, and I want to get RC2 released.

tonychee7000:
I had made MIDI Export work, see #3224

Issue #3079, opened 2016-10-15
Goto feature


musikBear:
Enhancement for 1.2.x or **later**
I suggest to let the user jump to a selected bar in the project.
A way to do that could be the current **Bar-Beat-Tick** Timer-objects.
https://lmms.io/wiki/images/b/bf/Timer.png
The user could right-click, and mouse-wheel-scroll the numbers/ or drag. When play is pressed, the project would then start at the position accordingly to the chosen values.
Design-wise it would be nice is the color of the numbers changed to another color, when the numbers was right-clicked. That would make the feature more intuitive

liushuyu:
Possible implementation in PR #3115

Sawuare:
I doubt the usefulness of this feature. I think users navigate in the song editor based on patterns and colors, not units of time.

PR #3115, opened 2016-11-12
Add Goto feature to TimeDisplayWidget

liushuyu:
I have tried to implement the proposed feature in #3079 . Not sure if it will fulfill the request.
There are changes/fixes made to `Song.h`: 
1. I have changed some `int` type to `qint64` type (on most common machines, this should be `long long`) to prevent from overflow.
2. I have to move `setPlayPos` from `private` to `public`.

Some known problems:
1. The `goto` dialog is unable to input the value over 9999 minutes.
2. Sometimes the LCD meter will still display negative value despite the fact that I have changed the data type.
3. Because the time is converted to milliseconds after `getMilliseconds` and then converted to ticks before `setPlayPos`, so there are around 50 ms of offset in minutes/milliseconds mode, around 1 tick of offset in bars/ticks mode.

### DIFF
--- a/include/Song.h
+++ b/include/Song.h
@@ -277,6 +277,7 @@ public slots:
 	void clearProject();
 
 	void addBBTrack();
+    void setPlayPos(qint64 ticks, PlayModes playMode );
 
 
 private slots:
@@ -317,8 +318,6 @@ private:
 		return m_playPos[m_playMode].getTicks() * Engine::framesPerTick() + 
 			m_playPos[m_playMode].currentFrame();
 	}
-	
-	void setPlayPos( tick_t ticks, PlayModes playMode );
 
 	void saveControllerStates( QDomDocument & doc, QDomElement & element );
 	void restoreControllerStates( const QDomElement & element );
--- a/include/TimeDisplayWidget.h
+++ b/include/TimeDisplayWidget.h
@@ -30,6 +30,7 @@
 #include <QHBoxLayout>
 
 #include "LcdWidget.h"
+#include "TimeInputDialog.h"
 
 
 class TimeDisplayWidget : public QWidget
@@ -42,10 +43,12 @@ public:
 
 protected:
 	virtual void mousePressEvent( QMouseEvent* mouseEvent );
+    virtual void contextMenuEvent( QContextMenuEvent *event );
 
 
 private slots:
 	void updateTime();
+    void timeJump();
 
 
 private:
@@ -64,6 +67,7 @@ private:
 	LcdWidget m_majorLCD;
 	LcdWidget m_minorLCD;
 	LcdWidget m_milliSecondsLCD;
+    TimeInputDialog* m_timeinputbox;
 
 } ;
 
--- a/include/TimeInputDialog.h
+++ b/include/TimeInputDialog.h
@@ -1,0 +1,69 @@
+/*
+ * TimeInputDialog.h - dialog for input playback time
+ *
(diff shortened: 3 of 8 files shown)


liushuyu:
I will first try to fix the whitespace thing... And align/justify the layout of the code...

liushuyu:
Well, I don't want to idle for so long...
So if there's no objections within 50 hours, other members or I may merge this.

BaraMGB:
Please let me check this first. I'll look into it tomorrow.

liushuyu:
Okay 😃

BaraMGB:
Hi @liushuyu . Can you please explain me, why you changed the data type of ticks from tick_t to qint64? And if qint64 is better (because it's bigger?) shouldn't we change that in lmms_basics.h?

I've tested your changes. Works here. But I would expect to change the position with mouse wheel in the spinboxes, too. What do you think?

liushuyu:
>Hi @liushuyu . Can you please explain me, why you changed the data type of ticks from tick_t to qint64? And if qint64 is better (because it's bigger?) shouldn't we change that in lmms_basics.h?

Yes,  it's bigger. And I originally wanted to change that in `lmms_basics.h`, but I am afraid this change may break something that I untested. `qint64` has larger range than `int`, the size of the latter may also varies on some machines (like SPARC64). 

>I've tested your changes. Works here. But I would expect to change the position with mouse wheel in the time widget directly, too. What do you think?

I also want to implement that, but I am considering if the playback cursor should be set as you scroll the mouse wheel or set it after one second pause after you have scrolled to a specific value (that's act like rubber band).

BaraMGB:
Okay, I'm not sure with the type change. It could break something we don't see in the moment. Is qint64 necessary for your implementation? If not I feel better to have tick_t here. We can change the size of tick_t in a other pull request. 

Are there some other opinions here? @jasp00?

softrabbit:
32 bits is enough to express 5598 seconds at a 999 BPM tempo and a resolution of 768 ticks per quarter note, that's roughly 1.5 hours. AFAIK a quarter note in LMMS is really only 48 ticks, which would put the limit at a bit over 24 hours. 

So, I'd say making tick_t 64 bits wide can wait, unless there's an concrete problem somewhere if fixes.

liushuyu:
@BaraMGB @softrabbit The problem here is you can't control what users input. Because we have 4 digit LED display, so users may think they can enter a value up to 9999 minutes. When this scenario occurs, the value passed to `setPlayPos` will be negative, since it is beyond the limit.

Currently, I can think of the following possible solutions:
1. Limit the input range and change `qint64 setPlayPos` back to `tick_t setPlayPos`
2. Change `tick_t` to 64-bit value
3. I don't know, you will fill in the blank: _______________________________

BTW, for the inaccuracy problem, I don't know if there's a solution to this. I wonder how `TimeLineWidget` sets the play position. Any fresh ideas?

Have a nice day.

BaraMGB:
@>9999min this should be catched up by your line edit I think.

liushuyu:
>@>9999min this should be catched up by your line edit I think.

No one can enter a value more than 9999 minutes in that input box

BaraMGB:
@liushuyu okay, I have to check this. And if this is a bug, we should fix this in a separate pull request. But there is something more I don't like about this pull request: the pop up window. We should always find a solution without a pop up. Perhaps with dragging up and down or a line edit.

PhysSong:
@liushuyu Why not `int64_t`?

PR #3120, opened 2016-11-14
Delay Plugin Redesign

RebeccaDeField:
## In this pull request I have:

- Redesigned the plugin
- Renamed the knobs for consistency
- Added a grid
- Changed the size of the plugin
- Fixed alignment 

## Screenshot

@Umcaruje @BaraMGB 

Related: #2831


tresf:
Thanks @RebeccaDeField .

In fear of sounding like a broken record... the contrast between the canvas and the rest of the UI is too low and could benefit from an increase in contrast.

RebeccaDeField:
@tresf Actually, we're on the same page right now because upon reviewing my PR I already started to modify my design with the same concern. I have an idea that I believe will solve the problem.

RebeccaDeField:
@tresf What do you think about this?
shot-2016-11-13_19-50-36

tresf:
@RebeccaDeField it certainly helps the slider on the right although the canvas is still hard to distinguish.  Could this perhaps benefit from a grid pattern similar to some of the other plugins?

RebeccaDeField:
@tresf Do you think that the plugin would benefit from a the green and gray grid from a usability standpoint? If so, I think we should definitely add a grid.

If it's just a matter of the distinguishing where the edges of the canvas are, maybe we can use a more apparent border to accomplish this?
unnamed
shot-2016-11-13_20-15-57

tresf:
> Do you think that the plugin would benefit from a grid from a usability standpoint? If so, I think we should definitely add a grid.

Yes, I think so.  Perhaps even long-term swap the Y axis and make the mouse responsive outside of the canvas area.

RebeccaDeField:
> Yes, I think so.


@tresf Then I will make it so #1. :smirk: What kind of spacing do you think the grid should have? Would this grid benefit from major and minor lines, or do you think we should stick to something more simple?

tresf:
> What kind of spacing do you think

Since one measurement is 0 - 1 and the other is 0 - 5, a 10 x 10 grid perhaps?  Major lines in the center?

RebeccaDeField:
@tresf Your thoughts on this?

tresf:
@RebeccaDeField that looks great.

Umcaruje:
I like this, though I find that 5 seconds worth of delay is quite useless with that XY pad, especially cause it's linear. But that's talk for another issue.

@RebeccaDeField Please change `REGEN` to the appropriate abbreviation we chose for Feedback on previous plugins, as that's what that is. Also, the Rate and the Amount are related to a LFO, so I wonder if we should indicate that visually somehow. Otherwise, top notch work!

P.S. I couldn't help but notice that that fader is rendering as linear, it would be good to set it to render in a dbFS scale

RebeccaDeField:
> Please change REGEN to the appropriate abbreviation we chose for Feedback on previous plugins, as that's what that is.

:+1: 

> Also, the Rate and the Amount are related to a LFO, so I wonder if we should indicate that visually somehow.

Maybe change the labels to LFO AMNT and LFO RATE, or did you have something else in mind?

>  It would be good to set it to render in a dbFS scale

Sorry, but I don't know how to set this. A few pointers?

Thank you :)

Umcaruje:
>Sorry, but I don't know how to set this. A few pointers?

Should be as simple to as adding a call to make the faders render in dbfs like 
```c++
outFader->setLevelsDisplayedInDBFS();
```

here: https://github.com/LMMS/lmms/blob/master/plugins/Delay/DelayControlsDialog.cpp#L79

This is how it's done for the FX mixer: https://github.com/LMMS/lmms/blob/master/src/gui/FxMixerView.cpp#L284-L286


>Maybe change the labels to LFO AMNT and LFO RATE, or did you have something else in mind?

Sure, if they end up fitting, otherwise a line and the text LFO over the knobs could also work.

RebeccaDeField:
This has been updated then tested by @BaraMGB myself. I'm merging.

Issue #3129, opened 2016-11-20
Build problem on Fedora x86_64: Relocatable linking with relocations from format elf64-x86-64 (/usr/lib64/wine/libwinecrt0.a(exe_entry.o)) to format elf32-i386 (RemoteVstPlugin.sBWziW.o) is not supported winebuild: ld failed with status 1 winegcc: winebuild failed


jasonMisaq:
A similar issue has already been discussed on https://github.com/LMMS/lmms/issues/15, but those instructions are mostly for Ubuntu and I could not solve this on Fedora x86_64. Also, if anyone can tell me where I should use the workaround given by user @tobydox on the same page linked above, I would appreciate it.  Currently, the complete output for my make is as follows:   
```
/home/mhd/src/lmms/build
[mhd@localhost build]$ make
[  0%] Built target manpage
[  0%] Built target bin2res
[  0%] Automatic moc for target lmmsobjs
[  0%] Built target lmmsobjs_automoc
[ 17%] Built target lmmsobjs
[ 17%] Automatic moc for target lmms
[ 17%] Built target lmms_automoc
[ 17%] Built target lmms
[ 17%] Built target audiofileprocessor
[ 18%] Built target kicker
[ 18%] Built target tripleoscillator
[ 19%] Built target amplifier
[ 20%] Built target bassbooster
[ 20%] Built target bitinvader
[ 21%] Built target bitcrush
[ 21%] Built target carlabase
[ 22%] Built target carlapatchbay
[ 22%] Built target carlarack
[ 23%] Built target crossovereq
[ 24%] Built target delay
[ 25%] Built target dualfilter
[ 26%] Built target dynamicsprocessor
[ 28%] Built target eq
[ 29%] Built target flanger
[ 29%] Built target hydrogenimport
[ 30%] Built target ladspabrowser
[ 31%] Built target ladspaeffect
[ 33%] Built target caps
[ 34%] Built target tap_eq
[ 34%] Built target tap_dynamics_m
[ 35%] Built target tap_vibrato
[ 35%] Built target tap_autopan
[ 35%] Built target tap_rotspeak
[ 35%] Built target tap_dynamics_st
[ 35%] Built target tap_tubewarmth
[ 35%] Built target tap_chorusflanger
[ 36%] Built target tap_deesser
[ 36%] Built target tap_tremolo
[ 36%] Built target tap_limiter
[ 36%] Built target tap_eqbw
[ 36%] Built target tap_echo
[...]

PR #3130, opened 2016-11-21
workaround for broken wineg++

jasonMisaq:
Only changes paths to how they are addressed in Fedora

### DIFF
--- a/plugins/vst_base/CMakeLists.txt
+++ b/plugins/vst_base/CMakeLists.txt
@@ -37,11 +37,11 @@ IF(LMMS_HOST_X86_64)
 	# workaround for broken wineg++ in WINE 1.4 (shipped e.g. with Ubuntu Precise)
 	EXEC_PROGRAM( ${WINE_CXX} ARGS "-v -m32 /dev/zero" OUTPUT_VARIABLE WINEBUILD_OUTPUT)
 	IF("${WINEBUILD_OUTPUT}" MATCHES ".*x86_64-linux-gnu/wine/libwinecrt0.a.*")
-	SET(EXTRA_FLAGS ${EXTRA_FLAGS} -nodefaultlibs /usr/lib/i386-linux-gnu/wine/libwinecrt0.a -luser32 -lkernel32 -lgdi32)
+		SET(EXTRA_FLAGS ${EXTRA_FLAGS} -nodefaultlibs /usr/lib/i386-linux-gnu/wine/libwinecrt0.a -luser32 -lkernel32 -lgdi32)
 	ENDIF()
 	#The following check works on Fedora systems 
 	IF("${WINEBUILD_OUTPUT}" MATCHES ".*lib64/wine/libwinecrt0.a.*")
-        SET(EXTRA_FLAGS ${EXTRA_FLAGS} -nodefaultlibs /usr/lib/i386/wine/libwinecrt0.a -luser32 -lkernel32 -lgdi32)
+		SET(EXTRA_FLAGS ${EXTRA_FLAGS} -nodefaultlibs /usr/lib/i386/wine/libwinecrt0.a -luser32 -lkernel32 -lgdi32)
 	ENDIF()
 ENDIF(LMMS_HOST_X86_64)
 
--- a/plugins/vst_base/CMakeLists.txt
+++ b/plugins/vst_base/CMakeLists.txt
@@ -36,13 +36,12 @@ IF(LMMS_HOST_X86_64)
 
 	# workaround for broken wineg++ in WINE 1.4 (shipped e.g. with Ubuntu Precise)
 	EXEC_PROGRAM( ${WINE_CXX} ARGS "-v -m32 /dev/zero" OUTPUT_VARIABLE WINEBUILD_OUTPUT)
-	if("${WINEBUILD_OUTPUT}" MATCHES ".*x86_64-linux-gnu/wine/libwinecrt0.a.*")
+	IF("${WINEBUILD_OUTPUT}" MATCHES ".*x86_64-linux-gnu/wine/libwinecrt0.a.*")
 	SET(EXTRA_FLAGS ${EXTRA_FLAGS} -nodefaultlibs /usr/lib/i386-linux-gnu/wine/libwinecrt0.a -luser32 -lkernel32 -lgdi32)
-	endif()
-	#The following check works in Fedora systems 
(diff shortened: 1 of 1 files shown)


tresf:
@jasonMisaq thanks!

Since the original work-around is still perfectly valid for Ubuntu, can you please keep it in there?  Also, the comment above it should be simplified to simply state `# workaround for broken wineg++` without the specifics about Ubuntu.

tresf:
@jasonMisaq thanks!  Can you double-check the whitespace formatting?  It's a bit off now.  If you can't see it with your editor, take a look at it in the files: https://github.com/LMMS/lmms/pull/3130/files

jasonMisaq:
@tresf does this create syntax error? I am not much familiar with cmake files syntax

plugins/vst_base/CMakeLists.txt
```diff
@@ -36,8 +36,12 @@ IF(LMMS_HOST_X86_64)
...
 	# workaround for broken wineg++ in WINE 1.4 (shipped e.g. with Ubuntu Precise)
 	EXEC_PROGRAM( ${WINE_CXX} ARGS "-v -m32 /dev/zero" OUTPUT_VARIABLE WINEBUILD_OUTPUT)
-	if("${WINEBUILD_OUTPUT}" MATCHES ".*x86_64-linux-gnu/wine/libwinecrt0.a.*")
-		SET(EXTRA_FLAGS ${EXTRA_FLAGS} -nodefaultlibs /usr/lib/i386-linux-gnu/wine/libwinecrt0.a -luser32 -lkernel32 -lgdi32)
+	IF("${WINEBUILD_OUTPUT}" MATCHES ".*x86_64-linux-gnu/wine/libwinecrt0.a.*")
+	SET(EXTRA_FLAGS ${EXTRA_FLAGS} -nodefaultlibs /usr/lib/i386-linux-gnu/wine/libwinecrt0.a -luser32 -lkernel32 -lgdi32)
+	ENDIF()
+	#The following check works on Fedora systems 
+	IF("${WINEBUILD_OUTPUT}" MATCHES ".*lib64/wine/libwinecrt0.a.*")
+        SET(EXTRA_FLAGS ${EXTRA_FLAGS} -nodefaultlibs /usr/lib/i386/wine/libwinecrt0.a -luser32 -lkernel32 -lgdi32)
```
tresf: @jasonMisaq this looks like spaces instead of a tab, which is causing the extra indent.

tresf:
> @tresf does this whitespace formatting create a syntax error? I am not much familiar with cmake files syntax

@jasonMisaq no it does not, it's purely cosmetic but part of our code quality standards.

We haven't converted all of our source to use the same exact format, but we are pretty strict about matching a file to what it had before for readability purposes.

Once you fix that indent, we'll merge.  Thanks again!

jasp00:
I am not sure this workaround is necessary. According to https://github.com/LMMS/lmms/issues/3129#issuecomment-261873431, 32-bit Wine was not installed in the first place.

jasonMisaq:
I do not really know @jasp00, but it solved https://github.com/LMMS/lmms/issues/3129 though. According to [Wine](https://wiki.winehq.org/FAQ#How_do_I_create_a_32_bit_wineprefix_on_a_64_bit_system.3F), there are significant bugs with 64 bit systems. So, I guess that's why installing 32-bit versions (i686) solved the build issue.

tresf:
> According to https://github.com/LMMS/lmms/issues/3129#issuecomment-261873431, 32-bit Wine was not installed in the first place.

Yes, but according to #3129 the build errors are confusing and obscure.  Won't this yield more sane error messages, such as `/usr/lib/i386-linux-gnu/wine/libwinecrt0.a not found`?  This seems much more sane over the current ` relocations from format elf64-x86-64 to format elf32-i386 is not supported`, no?

jasonMisaq:
@tresf, you are right, I was able to confirm this by removing wine-devel.i686 and wine.i686. Then, I built again by running `make`, and got this error:  
```
winegcc: File does not exist: /usr/lib/wine/libwinecrt0.a
plugins/vst_base/CMakeFiles/vstbase.dir/build.make:61: recipe for target 'plugins/RemoteVstPlugin' failed
make[2]: *** [plugins/RemoteVstPlugin] Error 2
CMakeFiles/Makefile2:8768: recipe for target 'plugins/vst_base/CMakeFiles/vstbase.dir/all' failed
make[1]: *** [plugins/vst_base/CMakeFiles/vstbase.dir/all] Error 2
Makefile:138: recipe for target 'all' failed
make: *** [all] Error 2
```
But, note that this is true considering the fact that we applied the workaround. Maybe if we did not, we would get the error we got in the first place (`relocations from format elf64-x86-64....`)

tresf:
> and got this error

This seems much more sane of an error.  Can you also test @jasp00's theory that reverting the patch has no effect on the end-result when the proper dependencies are met?  That way we at least know.

jasonMisaq:
Yes, @tresf I installed wine-devel.i686 again, reverted the patch. Removed CMakeCache.txt, did a `make clean`, and then ran `make` again. This time it built fine. I do not understand why did the error came up in the first place. So, maybe @jasp00 is right, the Fedora part (or even the Ubuntu part?) of the workaround might not be required. 

Now my ./plugins/vst_base/CMakeLists.txt looks like this:  
```
# workaround for broken wineg++ in WINE 1.4 (shipped e.g. with Ubuntu Precise)
        EXEC_PROGRAM( wineg++ ARGS "-v -m32 /dev/zero" OUTPUT_VARIABLE WINEBUILD_OUTPUT)
        IF("${WINEBUILD_OUTPUT}" MATCHES ".*x86_64-linux-gnu/wine/libwinecrt0.a.*")
                SET(EXTRA_FLAGS ${EXTRA_FLAGS} -nodefaultlibs /usr/lib/i386-linux-gnu/wine/libwinecrt0.a -luser32 -lkernel32 -lgdi32)
        ENDIF()
        #The following check works on Fedora systems
        #IF("${WINEBUILD_OUTPUT}" MATCHES ".*lib64/wine/libwinecrt0.a.*")
        #       SET(EXTRA_FLAGS ${EXTRA_FLAGS} -nodefaultlibs /usr/lib/i386/wine/libwinecrt0.a -luser32 -lkernel32 -lgdi32)
        #ENDIF()

```

tresf:
>  Fedora part (or even the Ubuntu part?) of the workaround might not be required.

Correct.  Since it yields a sane error message, I don't think we need to remove it.  Ideally we'd detect the environment properly never get into this logic, but that is for another day. 👍

PR #3140, opened 2016-12-01
Graph: add Control+click feature for compressor usage

DeRobyJ:
Ctrl+click now creates two lines for a compressor-like graph, to use in Waveshaper and Dynamics Processor.

I had to set the initial and start values because the drawLineAt method changes the final values of the line slightly, so using ctrl+click multiple times would result in first and last values being changed.

In my opinion, this needs some fixes, but I felt like sharing it anyway.

My original intent was not to reset initial and final values, but to keep old ones and draw lines using those values and the mouse position. This would have allowed for more generalist usage, like for Multitap Echo.

pizza3
(in this screen, I used the new feature and then pressed "-1dB" some times)

### DIFF
--- a/src/gui/widgets/Graph.cpp
+++ b/src/gui/widgets/Graph.cpp
@@ -159,20 +159,7 @@ void Graph::mousePressEvent( QMouseEvent * _me )
 {
 	if( _me->button() == Qt::LeftButton )
 	{
-		if ( !( _me->modifiers() & Qt::ShiftModifier ) )
-		{
-			// get position
-			int x = _me->x();
-			int y = _me->y();
-
-			changeSampleAt( x, y );
-
-			// toggle mouse state
-			m_mouseDown = true;
-			setCursor( Qt::BlankCursor );
-			m_lastCursorX = x;
-		}
-		else
+		if ( _me->modifiers() & Qt::ShiftModifier )
 		{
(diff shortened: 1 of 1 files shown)


Umcaruje:
I would find this confusing like a user. A better solution would be to actually add ratio/knee/threshold knobs to the dynamic processor and have a freeline mode that would turn the knobs off for drawing custom curves.

DeRobyJ:
Well the two features aren't contrasting. We may have both

JohannesLorenz:
As this is a feature and stable-1.2 has feature freeze for a long time, can you please re-target this to master?

PhysSong:
For compressor, now there is #5458.

DeRobyJ:
Of course, and it's magnificent!

That said, that is a Compressor, a mixing effect, this is a Wave Shaper, a sound design tool!

My wording wasn't really clear when I did this pull request. If you feel that this feature is not useful anymore, I'm ok with it. I just have one question before closing it: when ad how do you use the Waveshaper? What shapes do you draw?

When I did this change, I did it because every time I used the waveshaper I had to draw that compressor-like shape. Maybe that's just my case.

DouglasDGI:
The feature requested here will be resolved by #4367.

DeRobyJ:
Oh, then I guess I can close this ahah

Great stuff!

PR #3171, opened 2016-12-20
Arpeggiator, swap positions of Random and DownUp

zonkmachine:
arpdirmodel

I put an upgrade path in DataFile.cpp and that's a first for me so I would be happy if someone reviewed this.

### DIFF
--- a/src/core/DataFile.cpp
+++ b/src/core/DataFile.cpp
@@ -883,6 +883,21 @@ void DataFile::upgrade_1_1_91()
 			el.setAttribute( a, ( el.attribute( a ) == "0" ) ? "1" : "0" );
 		}
 	}
+
+	list = elementsByTagName( "arpeggiator" );
+	for( int i = 0; !list.item( i ).isNull(); ++i )
+	{
+		QDomElement el = list.item( i ).toElement();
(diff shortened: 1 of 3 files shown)


src/core/DataFile.cpp
```diff
@@ -883,6 +883,21 @@ void DataFile::upgrade_1_1_91()
 			el.setAttribute( a, ( el.attribute( a ) == "0" ) ? "1" : "0" );
 		}
 	}
+
+	list = elementsByTagName( "arpeggiator" );
+	for( int i = 0; !list.item( i ).isNull(); ++i )
+	{
+		QDomElement el = list.item( i ).toElement();
+		// Swap elements ArpDirRandom and ArpDirDownAndUp
+		if( el.attribute( "arpdir" ).toInt() == 3 )
```
tresf: What is the advantage to parsing an `int` in this instance over doing a string comparison?

zonkmachine: I tried that first and it wouldn't compile. I had another go and it came out like this instead. Looks cleaner so I'm changing.

```
		if( el.attribute( "arpdir" ) == "3" )
		{
			el.setAttribute( "arpdir", "4" );
		}
```

Issue #3169, opened 2016-12-19
Sample tracks eats buffers


zonkmachine:
Since #1088 `LMMS Memory Manager` (~LMMS 1.1.0), buffers acquired through the SamplePlayHandle() aren't handed back after playing. If you stop a sample played from a SampleTrack the buffer is properly released but if the sample plays all the way through it's gone.
This is best demonstrated by activating the metronome and hitting play. At 180 bpm it will take almost three minutes to chomp up all 512 of them at my machine and close down business.

```
BufferManager: out of buffers
Aborted (core dumped)
```

These leaked buffers will stay gone as long as LMMS is running, even if you load another project or a default one. Gone...

softrabbit:
There's a SampleBuffer being created in one of the constructors, https://github.com/LMMS/lmms/blob/master/src/core/SamplePlayHandle.cpp#L38
...but never deleted?

zonkmachine:
Yeah, I noticed that one and I tried to just wrap a `sharedObject::ref( )` around it like in the other constructors but didn't know how to do it properly.

softrabbit:
Like `m_sampleBuffer( sharedObject::ref( new SampleBuffer( sampleFile ) ) )` ? That's my guess on how to do it.

The bug might be somewhere else, but that small flaw jumped out right away when reading the code.

zonkmachine:
> Like m_sampleBuffer( sharedObject::ref( new SampleBuffer( sampleFile ) ) ) ? That's my guess on how to do it.

That's exactly what I tried. Just tried it again and it makes no difference. I'm going to have one more go at fixing this.

zonkmachine:
It looks like we first create an m_sampleBuffer in the SampleTrackTCO
https://github.com/LMMS/lmms/blob/master/src/tracks/SampleTrack.cpp#L88
and then another one in SamplePlayHandle(). ?

zonkmachine:
I think I'll pass this one... Rude little buffers.  :angry:

softrabbit:
Crashes quite reliably at bar 128:0:0 using metronome. No crash for me when simulating metronome in a bar & beat-track using the same samples. 

[...]

PR #3173, opened 2016-12-21
Don't run metronome when project is empty

zonkmachine:
On completely empty project we get a crash when the metronome is activated so we test for this.

_Edit, bug description moved from #3169: The fastest crash I get with the metronome is if I delete all tracks or create an empty track from template and then hit play. I think this is a separate problem though as earlier versions you could access the metronome only from within the Piano Roll so there had to be at least one track to run it._


zonkmachine:
With this fix it will not crash but in some odd instances where you:
 - have tracks in the Song Editor and...
 - an empty BBTrack that...
 - was made originally empty from a template and...
 - hit play in that BBTrack
Under these circumstances you will get a single metronome tick and then silence. Just add an instrument in there and it will tick away again. Not many will ever stumble upon this.

zonkmachine:
Solid, merging...

PR #2526, opened 2016-01-28
Fix outdated copyright info, fix license not showing in the about dialog

Umcaruje:
Fixes #2508 

Pics: 
screenshot from 2016-01-28 23 18 00
screenshot from 2016-01-28 23 18 19
screenshot from 2016-01-28 23 18 39

PR #3174, opened 2016-12-23
corrected timestamp on de.ts

Serkan-devel:
2015, not 2014


zonkmachine:
Hi!

Thanks for the patch, but...
a) Translations are managed on our project page at Transiffex:
https://www.transifex.com/lmms/lmms
b) I don't think the Copyright string is really translated but managed centrally and if it shows the wrong value it's just waiting to be refreshed prior to the release. I think @liushuyu can shed some light on this.

Umcaruje:
This was fixed in #2526. the de.ts file hasn't been updated, as it's not translated to a 100% on transifex.

PR #3090, opened 2016-10-25
Envelopes are on now by default

mikobuntu:
~~fixes~~ Temporary workaround for #3086

I set a minimal attack just to quash any artefacts caused by a loud instrument, which i believe 0.002 secs will not be noticeable or feel like there is any delay until the audio starts. I set both Hold and Decay to 0 so that we only hear sound as long as a key is pressed. Sustain was obviously set to full i.e 1 to allow the note to hold while being pressed and most importantly release was kept at the default of 0.1 , which seemed to give the best results ( no clicks on key release ) [ smaller values might work ]. 
Setting m_amountModel to 1.0 ensures that amount of the Envelope is at maximum ( which switches it on by default ).

PR #3176, opened 2016-12-25
Revert "Envelopes are on now by default"

### LINKED ISSUE #3143: Drums sound totally different in beat+bassline editor after envelopes are on by default
Drums (especially kicks, snares, crashes, rides and other longer sounds) sound very different in library and when added to project.

When I add drums to project I have to put so much effort now to make them sound again like in library.

Just drag and drop kick04 and snare07 to B+B and listen how wrong it sounds. How this even got merged?

zonkmachine:
Fixes #3143 

@mikobuntu 
I think we should revert #3090 for the time being and sneak back and fix the issue after rc2.
I haven't looked at what it would take to fix the beat note envelopes but if we go with default envelopes on, which I'm also leaning toward, then maybe it would take two separate constructors(?).


mikobuntu:
Yes I'm happy for this to be reverted and we can look for a way to fix this issue in the near future.

Umcaruje:
>I think we should revert #3090 for the time being and sneak back and fix the issue after rc2.

:+1:

PR #3202, opened 2017-01-02
ReverbSC: Initial implementation

PaulBatchelor:
Hello,

Here is my initial implementation of the Sean Costello reverb algorithm, implemented as a native LMMS plugin. The full DSP inner loop code has been made, and a very minimal GUI has been added with four knobs: input gain, size, color, and output gain. Feel free to muck about with the name and the looks.

I've done my best to keep the coding convention consistent with everything around it, but it's possible I've missed a few things. One thing you'll note is that the DSP code itself is extracted from my Soundpipe library. These files are largely untouched, as I've built Soundpipe to be largely drag-n-drop (AudioKit uses soundpipe in this way already). For this reason, this style consistency is different. I felt this to be fine. 

One thing I'm not entirely sure about is a field I assume to be the plugin ID, which I have set to 0x123 right now. This will most likely need to be changed.

Using this plugin is pretty straight forward. Open up LMMS, and add ReverbSC as a track effect. One trouble I had initially was that the reverb tail would abruptly shut off immediately after the note was finished. This was remedied by setting the decay time for the gate to be very high. Is this a standard thing for all effects in LMMS? Ideally, it would be better to remove it entirely for this effect. The reverb needs more space than the gate is currently giving. 

Let me know if I'm doing everything right! 

-P


PaulBatchelor:
Travis failed. This is my fault. I forgot to remove the "soundpipe.h" header file. I will fix this.

PaulBatchelor:
Windows is bugging out with -Werror flag enabled. Turns out there was a warning I missed, having to do with the RNG. I'll try and see if I can get that warning to pop up on my own machine and fix that.

PaulBatchelor:
@tresf The only test that doesn't pass here is unrelated to my PR. I believe it will go away when it is merged because of #3201.

BaraMGB:
there's a bug: the reverb tail stoppes playing if the instrument produces no sound anymore.

PaulBatchelor:
Bumping for visibility. I'm new to LMMS, so I don't the protocol for how things get reviewed/merged. Apologies for the noise. What needs to happen for this to be merged?

tresf:
> Bumping for visibility. I'm new to LMMS, so I don't the protocol for how things get reviewed/merged. Apologies for the noise. What needs to happen for this to be merged?

Testing and feedback really.  A good start is to make a win32 build for users.  I'll see what I can do about that.

zonkmachine:
I'll be testing this when I'm home after the weekend. Maybe if you add a screenshot of it you can trigger the in-house artists...

PaulBatchelor:
@tresf @zonkmachine thanks guys! let me know if you have any questions.

Also, here is what the plugin looks like right now:
shot

tresf:
Here's a win32 build for testers like @musikBear.  https://github.com/LMMS/lmms/pull/3202

tresf:
https://github.com/tresf/lmms/releases/tag/v1.2.0-rc2-reverbsc

tresf:
> I had initially was that the reverb tail would abruptly shut off immediately after the note was finished. 

This doesn't occur with the other reverb plugins.  I think this is a bug.

tresf:
> One thing I'm not entirely sure about is a field I assume to be the plugin ID, which I have set to 0x123 right now. This will most likely need to be changed.

They [all seem to be `0x0100`](https://github.com/LMMS/lmms/blob/master/plugins/MultitapEcho/MultitapEcho.cpp#L39).  I guess it should be `0x0100`.

Early feedback... I really love the warmth and tone of this reverb; can't wait to compose with it.  It's unlike the others we ship.  I feel the default knob positions are good for this type of effect plugin.


>  muck about with the name

Name proposal: "Warm Reverb"

> the looks

It's workable -- Good enough to ship.  The graphics team may have some fun ideas for it, but it shouldn't hold up the release.

Plugins are modular too, this can make it in 1.2 rather safely.

PaulBatchelor:
> This doesn't occur with the other reverb plugins. I think this is a bug.

One thing that seemed to fix it was moving this line here to the end of the process loop:

https://github.com/PaulBatchelor/lmms/blob/master/plugins/ReverbSC/ReverbSC.cpp#L94

The idea being that the accumulated signal would be the wet + dry signal to include the reverb tail, as opposed to just the dry signal.

plugins/ReverbSC/ReverbSC.cpp
```diff
@@ -0,0 +1,141 @@
...
+	SPFLOAT dcblkL, dcblkR;
+	
+	ValueBuffer * inGainBuf = m_reverbSCControls.m_inputGainModel.valueBuffer();
+	ValueBuffer * sizeBuf = m_reverbSCControls.m_sizeModel.valueBuffer();
+	ValueBuffer * colorBuf = m_reverbSCControls.m_colorModel.valueBuffer();
+	ValueBuffer * outGainBuf = m_reverbSCControls.m_outputGainModel.valueBuffer();
+
+	for( fpp_t f = 0; f < frames; ++f )
+	{
+		outSum += buf[f][0]*buf[f][0] + buf[f][1]*buf[f][1];
```
zonkmachine: Shouldn't outsum be an absolute number like:
`outSum += fabs( buf[f][0]*buf[f][0] ) + fabs( buf[f][1]*buf[f][1] );`

If I interpret this correctly the [checkGate function](https://github.com/LMMS/lmms/blob/48cc3bb5ef85bac8cc7c01b11dd3139c547a0677/src/core/Effect.cpp#L140:L162) basically just tests if the signal is lower than the gate value over one frame, so if the input is a wave with a period that is divisible with the buffer size it would see an input of 0, silence and cut out.

PaulBatchelor: You are squaring the left and right signal, so the fabs would be a redundant operation.

PaulBatchelor: Also, would it be okay if I move this operation to the end of the for-loop? That way, the reverb tail also keeps the effect on. At the moment, there's a bit of a bug where the gate immediately shuts off after a note has been played, cutting off the tail.

zonkmachine:
I've tried it now for a couple of days and like it a lot. It sounds good and the interface is simple.
1.2 :+1: 

Some thoughts:
In the following project I'm just repeating the same note while automating the reverbs input/output levels in opposite directions. 
[reverbTestAutomation.mmp.zip](https://github.com/LMMS/lmms/files/712711/reverbTestAutomation.mmp.zip)

1 - Are both level controls really needed? They seem to essentially control the same level.
2 - There is noise on looping back. This can be controlled by setting the levels a bit slower, morphing, to the starting values before looping back. Is there something else that can be done here to spare our beloved user the shots and pops? Clearing out a buffer or so?

plugins/ReverbSC/ReverbSC.cpp
zonkmachine: ^^ Oh, dear. But off course... I learnt that math when I was 12.
^ Sure. I looked at the other plugins and they use either method.

PaulBatchelor:
Thank you for the test file! These are good questions.

1. While they do very similar things, I would say both gain stages are useful. It is easy to "overdrive" the reverb when it's decay is very long, so trimming down the input to the signal without attenuating the output signal is nice to have. Output gain is just a great tool for getting just the right level. 

2. The popping noise sounds like a sudden DC jump. Adding a small attack to your instrument gets rid of the loud this loud pop (I call this a de-clicking envelope). Does this solve the issue in a good enough way? There's still something weird that happens on loop back, but now it's not jarring.

zonkmachine:
> Adding a small attack to your instrument gets rid of the loud this loud pop (I call this a de-clicking envelope). Does this solve the issue in a good enough way?

Yes. No further questions. There are other issues open tracking envelope related noise/clicks.

PaulBatchelor:
I pushed a new change that would allow the reverb to stay on longer (literally moving one line of code). 

The error I seem to be getting here is because you have -Werror flag, and that cmake is using -std=c++11 on a C file:

https://travis-ci.org/LMMS/lmms/jobs/196288079#L3233

tresf:
> The error I seem to be getting here is because you have -Werror flag, and that cmake is using -std=c++11 on a C file:

That's my fault, per 607d3f4.

I decided to make an assumption our own plugins would be purely `c++11` or not `c++11` to simplify the cmake files.  This allows us to quickly set `c++11` for our entire project and disable for individual plugins.  This has the downside of throwing warnings on non-cpp files.

The workaround is to use the opl2 method: https://github.com/LMMS/lmms/blob/master/plugins/opl2/CMakeLists.txt#L3-L5

```cmake
# Enable C++11 for CXXFLAGS only
REMOVE_DEFINITIONS(-std=c++0x)
SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++0x")
```

jasp00:
>CMake doesn't easily allow the removal of `CXX_FLAGS`

You may try:
`STRING(REPLACE "-std=c++0x" "" CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})`

tresf:
@PaulBatchelor I'm going to go slightly off topic in your thread about the CXX flags... 


> >CMake doesn't easily allow the removal of `CXX_FLAGS`

> You may try:
> `STRING(REPLACE "-std=c++0x" "" CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})`

@jasp00 That works.  My thought process was to keep it consistent with `src/CMakeLists.txt`.  About 2 years ago, @Lukas-W added this to `/CMakeLists.txt` which was eventually moved to `src/CMakeLists.txt`.  I think it's cleaner than doing substring replacements and I also felt that we would eventually move all of our own LMMS-made classes to CPP for consistency.  Thoughts welcome.

Lukas-W:
It might be worth noting that nowadays CMake has a better way to specify the C++ version per target, which is the property [`CXX_STANDARD`](https://cmake.org/cmake/help/v3.1/prop_tgt/CXX_STANDARD.html) with possible values `98`, `11` and `14`.
```cmake
set_property(TARGET lmms PROPERTY CXX_STANDARD 11)
```

CMake automatically detects whether it's compiling C or C++ code, and only adds the compile option to the command line for C++ files.
This property is initiaized by `CMAKE_CXX_STANDARD`, so we should be able to just do `set(CMAKE_CXX_STANDARD 11)` once and be done with it (except for files that are C++ but fail with C++11).

plugins/ReverbSC/ReverbSC.cpp
```diff
@@ -0,0 +1,141 @@
...
+{
+	if( !isEnabled() || !isRunning () )
+	{
+		return( false );
+	}
+
+	double outSum = 0.0;
+	const float d = (dryLevel() + 1) * 0.5;
+	const float w = (wetLevel() + 1) * 0.5;
+
```
zonkmachine: With regards to the use of the dry/wet knob (https://github.com/LMMS/lmms/issues/3261) I think maybe it's best to stick with the way it's done in the rest of plugins for now? The knob now shows -0.99 when you have 0.0 wet level. I find that a bit confusing.

PaulBatchelor: Noted. With #3261, it's best to be consistent for now. I'll make that change.

zonkmachine: Did you look into this? I've created a project for the [rc3 release](https://github.com/LMMS/lmms/projects/2) and have added ReverbSC to it.

PaulBatchelor: No not yet. Will do now.

zonkmachine:
When the Travis build is done, some 2 weeks from now...  I plan on squashing this to one commit, we happily do this around this place, and then merge. Suggested message

```
* ReverbSC - Plugin. lmplementation of the Sean Costello reverb algorithm
```
Cool?

PaulBatchelor:
@zonkmachine I appreciate the heads up on this. Go for it!

zonkmachine:
@PaulBatchelor ReverbSC doesn't seem to have a method to change sample rate. This means the rate will stay the default when you export so if you choose another rate and/or oversampling the reverb will come out wrong. One of the things I should have tested at an early stage...
The mixer emits `sampleRateChanged() `and the other plugins who do time related stuff has some way to deal with it. I don't know what would be a good example really. Here's one anyway, from the Delay plugin.
https://github.com/LMMS/lmms/blob/master/plugins/Delay/DelayControls.cpp#L41

PaulBatchelor:
@zonkmachine Ah, okay. I can look into that. 

So, the easy part is telling Soundpipe what the new SR is. That's just a matter of setting a variable in the struct, like I do when I initialized the plugin:

https://github.com/LMMS/lmms/blob/master/plugins/ReverbSC/ReverbSC.cpp#L53

One possible issue I forsee lies is reallocating memory for RevSC. The delay lines are computed at init-time based on the engine samplerate. The only way to do this is by destroying the instance of RevSC and creating a new instance with the new samplerate. Not ideal if this is an operation on a thread.

PaulBatchelor:
Looks like delay is allocating memory when the sampleRateChanged() method is called, so it doesn't seem unheard of:

https://github.com/LMMS/lmms/blob/master/plugins/Delay/StereoDelay.cpp#L82

This uses "new" C++ operator, while my code is using malloc in C.

zonkmachine:
@PaulBatchelor I made an issue to keep track of this https://github.com/LMMS/lmms/issues/3401

plugins/ReverbSC/ReverbSC.cpp
```diff
@@ -0,0 +1,141 @@
+/*
+ * ReverbSC.cpp - A native reverb based on an algorithm by Sean Costello
+ *
```
jasp00: Please add a copyright notice.

plugins/ReverbSC/base.c
```diff
@@ -0,0 +1,214 @@
+#include <stdlib.h>
```
jasp00: License information?

plugins/ReverbSC/base.h
```diff
@@ -0,0 +1,84 @@
+#include <stdint.h>
```
jasp00: License information?

plugins/ReverbSC/dcblock.c
```diff
@@ -0,0 +1,52 @@
...
+ * DCblock
+ *
+ * This code has been extracted from the Csound opcode "dcblock".
+ * It has been modified to work as a Soundpipe module.
+ *
+ * Original Author(s): Perry R. Cook
+ * Year: 1995
+ * Location: Opcodes/biquad.c
+ *
+ */
```
jasp00: License information?

plugins/ReverbSC/dcblock.h
```diff
@@ -0,0 +1,11 @@
+typedef struct {
```
jasp00: License information?

plugins/ReverbSC/revsc.c
```diff
@@ -0,0 +1,276 @@
...
+ * RevSC
+ *
+ * This code has been extracted from the Csound opcode "reverbsc".
+ * It has been modified to work as a Soundpipe module.
+ *
+ * Original Author(s): Sean Costello, Istvan Varga
+ * Year: 1999, 2005
+ * Location: Opcodes/reverbsc.c
+ *
+ */
```
jasp00: License information?

plugins/ReverbSC/revsc.h
```diff
@@ -0,0 +1,28 @@
+typedef struct {
```
jasp00: License information?

plugins/ReverbSC/dcblock.c
zonkmachine: I think it would be this license from csound.
https://github.com/csound/csound/blob/develop/COPYING

zonkmachine: Soundpipe was switched fro LGPL to MIT license here:
https://github.com/PaulBatchelor/Soundpipe/blob/master/LICENSE

PR #3206, opened 2017-01-03
Add pitch range CC handling

### LINKED ISSUE #1967: MIDI import: Pitch bend range CCs ignored and wrong default
First the easy part, the default range should be adjusted for instrument tracks imported from MIDI:
http://www.blitter.com/~russtopia/MIDI/~jglatt/tech/midispec/wheel.htm

> The GM spec recommends that MIDI devices default to using the entire range of possible Pitch Wheel message values (ie, 0x0000 to 0x3FFF) as +/- 2 half steps transposition (ie, 4 half-steps total range). The Pitch Wheel Range (or Sensitivity) is adjusted via an RPN controller message.

And second, if there are any pitch bend range messages, they should be handled. And they aren't plain CCs, I'm afraid:
http://kurzweil.com/knowledgebase/pc1x/internal_voices-program_mode/33/

> First you send cc# 100 with a value of 0; then cc#101 with a value of 0. This turns on reception for setting pitch bend with the Data controller (#6). Then you send cc# 6 with a value of 2 (in semitones; this will give you a whole step up and a whole step down from the center).

liushuyu:
Added one basic and widely used RPN CC handling: Pitch Sensitivity (aka Pitch Range)
CC code sequence:
```
101, 0  ; Indicate a RPN-MSB entry
100, 0  ; Indicate a RPN-MSB entry
; The combination above (0, 0) indicates to change Pitch Sensitivity (aka Pitch Range)
6, <value in semitones / 127>
38, 0 ; End
```
This will fix #1967

---
This change is [Reviewable](https://reviewable.io/reviews/lmms/lmms/3206)

### DIFF
--- a/plugins/MidiImport/MidiImport.cpp
+++ b/plugins/MidiImport/MidiImport.cpp
@@ -390,7 +390,8 @@ bool MidiImport::readSMF( TrackContainer* tc )
 		QString trackName = QString( tr( "Track" ) + " %1" ).arg( t );
 		Alg_track_ptr trk = seq->track( t );
 		pd.setValue( t + preTrackSteps );
-		int prevCC = 0;
+		int rpn_msb = -1;
+		int rpn_lsb = -1;
 
 		for( int c = 0; c < 129; c++ )
 		{
@@ -509,6 +510,16 @@ bool MidiImport::readSMF( TrackContainer* tc )
 
 								break;
 
+							case 6:
+								if ( rpn_lsb == 0 && rpn_msb == 0 )
+								{
+									objModel = ch->it->pitchRangeModel();
+									cc = ceil( cc * 127.0f );
+									printf( "%f: Pitch range changed to: %f\n", evt->time, cc );
+								}
+
+								break;
+
 							case 7:
 								objModel = ch->it->volumeModel();
 								cc *= 100.0f;
@@ -519,37 +530,13 @@ bool MidiImport::readSMF( TrackContainer* tc )
 								cc = cc * 200.f - 100.0f;
 								break;
 
-							case 6:
-								if ( prevCC == 201 )
-								{
-									objModel = ch->it->pitchRangeModel();
-									cc = floor( cc * 128.0f );
-									printf( "%f: Pitch range changed to: %f\n", evt->time, cc );
-								}
-
-								break;
-
 							case 100:
-								if( prevCC == 101 && cc == 0.0f )
-								{
-									prevCC = 201;
-								}
-								else
-								{
-									prevCC = 100;
-								}
+								rpn_msb = cc;
 
 								continue;
 
 							case 101:
-								if( prevCC == 100 && cc == 0.0f )
-								{
-									prevCC = 201;
-								}
-								else
-								{
-									prevCC = 101;
-								}
+								rpn_lsb = cc;
 
 								continue;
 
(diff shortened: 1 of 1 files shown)


tonychee7000:
I would say two `flags` might be better. The `prevCC = 201 ` looks strange.

liushuyu:
Because I need to prepare for other CC message other than (0,0)

tonychee7000:
I see that.


在 2017年01月04日 12:49, liushuyu 写道:

softrabbit:
I'm not entirely sure about the code. What if there's a ```CC 101, X ; CC 100,0```, won't that be interpreted the same for any X? I believe the order of these CCs isn't dictated by the MIDI standard.

Here's how I'd implement the RPN logic: https://github.com/softrabbit/lmms/blob/midi_import/plugins/MidiImport/MidiImport.cpp#L520. I think it's the more future-proof way, in case other RPNs are needed at some point.

tonychee7000:
Okay, CC101 means RPN MSB and CC100 means RPN LSB. RPN MSB seems to be
always 0 or 127(Null). While RPN LSB = 0 and MSB = 0, it means Pitch
Bend Sensitivity. 

I was read your code. Pitch bend should be multiplied by pitchRangeMode,
or you will hear incorrect tones after you import MIDI file. here is a
MIDI file you can test https://cfp.vim-cn.com/cbcNg

You need to use base64 to decode it. And it will be correct while you
hear a octave.

在 2017年01月04日 16:31, Raine M. Ekman 写道:

tonychee7000:
Okay, CC101 means RPN MSB and CC100 means RPN LSB. RPN MSB seems to be always 0 or 127(Null). While RPN LSB = 0 and MSB = 0, it means Pitch Bend Sensitivity. 

| CC101 RPN MSB | CC100 RPN LSB | Meaning |
| -----------------: | -----------------: | ---------- |
|                         0 | 0 | Pitch Bend Sensitivity |
| 0 | 1 | Fine Tune |
| 0 | 2 |	Coarse Tune |
| 127 | 127 |	Null |


I was read your code. Pitch bend should be multiplied by pitchRangeModel, or you will hear incorrect tones after you import MIDI file. here is a MIDI file you can test https://cfp.vim-cn.com/cbcNg

Base64 decoder is needed. And it will be correct while you hear a octave.

softrabbit:
> I was read your code. Pitch bend should be multiplied by pitchRangeModel, or you will hear incorrect tones after you import MIDI file. here is a MIDI file you can test https://cfp.vim-cn.com/cbcNg

I never said it was complete :smile:. Looks like I was going to post-process the pitch ranges in case they happened in a different track in the file, but that might be a pretty rare situation. Just simply multiplying while reading the track will probably work fine for something like 99% of all files. 

I guess #1967 can be closed when this is merged.

tonychee7000:
Aha, it's done.


在 2017年01月04日 17:21, Raine M. Ekman 写道:

liushuyu:
Well, some more annoying things occurred and proved this solution is not applicable. Close this out for future reference.

softrabbit:
> Well, some more annoying things occurred and proved this solution is not applicable.

Please, document them for future reference.

liushuyu:
Okay. So write this to wiki or just leave here as comments?

"Raine M. Ekman" <notifications@github.com> wrote:

Well, some more annoying things occurred and proved this solution is not
applicable.

Please, document them for future reference.

—
You are receiving this because you modified the open/close state.

Reply to this email directly, view it on GitHub
<https://github.com/LMMS/lmms/pull/3206#issuecomment-270770248>, or mute
the thread

softrabbit:
Comment here, this is the most likely place for people to look for info.

liushuyu:
>Comment here, this is the most likely place for people to look for info.

Okay, I will do

PR #3208, opened 2017-01-04
Fix build with Clang

### LINKED ISSUE #3073: Error building  lmms  with clang
Building with clang results in:
`AutomatableModel.h:310:8: error: private field 'm_oldValue' is not used`

Full log:
http://clang.debian.net/logs/2016-08-30/lmms_1.1.3-3_unstable_clang.log

jasp00:
Closes #3073.
Several issues were present, most of them because of third-party projects. System header `gig.h` may need to be overridden. This request adds a Clang build to Travis CI.


.travis/osx..script.sh
```diff
@@ -5,4 +5,4 @@ if [ $QT5 ]; then
         export CMAKE_PREFIX_PATH="$(brew --prefix qt55)"
 fi
 
-cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DWANT_QT5=$QT5 -DUSE_WERROR=OFF ..
+cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DWANT_QT5=$QT5 -DUSE_WERROR=ON ..
```
tresf: We can simply remove this `DUSE_WERROR` now, right?

.travis/linux..install.sh
```diff
@@ -15,3 +15,9 @@ else
 fi
 
 sudo apt-get install -y $PACKAGES
+
+# Fix header
```
tresf: @jasp00 should this be a cmake task so that it doesn't show up for non-travis builds too?

.travis/osx..script.sh
tresf: @jasp00 thanks I didn't realize this was our own flag.  This was added and only ever enabled for Mac travis builds.  My guess is @Lukas-W did this just to get Travis to pass for the time being.  If you've fixed all the clang warnings, is it needed anymore?  https://github.com/LMMS/lmms/commit/9cc1a5931c34535436a2fb82e51031121b3a53d1

tresf: To be more specific, couldn't we default it for `DCMAKE_BUILD_TYPE=Debug` like we do for stripping binaries, or is this flag a useful one to have around regardless?

plugins/LadspaEffect/swh/analogue_osc_1416.c
```diff
@@ -175,7 +175,6 @@ static void runAnalogueOsc(LADSPA_Handle instance, unsigned long sample_count) {
 	float otm2 = plugin_data->otm2;
 	unsigned int rnda = plugin_data->rnda;
 	unsigned int rndb = plugin_data->rndb;
-	blo_h_tables * tables = plugin_data->tables;
```
tresf: Once these are merged, we should issue a PR against https://github.com/swh/ladspa

jasp00:
Request for upstream at https://github.com/swh/ladspa/pull/44.

.travis/osx..script.sh
jasp00: `USE_WERROR` is not strictly necessary because you can use `CFLAGS=-Wno-error` and `CXXFLAGS=-Wno-error`. `USE_WERROR` should be dropped if Clang warnings go away.

.travis/linux..install.sh
jasp00: Non-Travis builds may not have sudo permission. We could add a task to warn about it and maybe fix it, but this is an issue that libgig upstream has dealt with and distros should handle their versions.

tresf: > Non-Travis builds may not have sudo permission.

Can't we leverage a cmake task which places the custom header in user-space?  Even if it's fixed upstream it will take a while for it to make it to the various -dev channels so it seems like we could hit two birds with one stone and it's not the only header we write at configure/compile time.

jasp00: >Can't we leverage a cmake task which places the custom header in user-space?

Yes, we can.

jasp00:
>Are there advantages to keeping Clang + Qt4 that you can think of?

We can easily enable this combination if the advantage arises.

jasp00:
Do you mind if I merge this one? It means an extra build for pull requests.

tresf:
@jasp00 yes however can you explain the severity of items like re-writing headers on build, notably qvector?

jasp00:
>can you explain the severity of items like re-writing headers on build, notably qvector?

The changes are harmless by themselves. The problem I see now is if a user does these steps:
1. Build in a broken environment.
2. Upgrade the environment.
3. Rebuild.

I will add a parameter to enable the rewriting.

tresf:
> I will add a parameter to enable the rewriting.

I meant, what breaks in the build if we don't rewrite those headers?  Is it only so that we can have stricter compiler warning settings?

jasp00:
>Is it only so that we can have stricter compiler warning settings?

It may be so. The errors could actually be warnings. I have not built with laxer settings.

tresf:
> I have not built with laxer settings.

I fit passes with laxer settings, can we consider removing it?  What is the lesser of two evils here?

jasp00:
>What is the lesser of two evils here?

Stricter settings are the best bet. Do not worry about rewriting, it is only meant for Travis CI. Users will not enable this option, unless they want to build with Clang, without warnings, and bad system headers.

Lukas-W:
How about just using Qt 5.7.1 which [includes the fix](https://github.com/qt/qtbase/blob/v5.7.1/src/corelib/tools/qvector.h#L766)? Homebrew [seems to provide 5.7.1](https://github.com/Homebrew/homebrew-core/blob/master/Formula/qt5.rb#L6).

tresf:
> How about just using Qt 5.7.1 which includes the fix? Homebrew seems to provide 5.7.1.

That works.  Since Qt5.7 still breaks macdeployqt we should probably merge in https://github.com/tresf/lmms/commit/f24f3d94bf3e9400003ebb962a77718f020a642f shortly afterward so that packaging isn't adversely affected.

jasp00:
>How about just using Qt 5.7.1 which includes the fix?

Linux build runs on Trusty, so an older Qt 5 should be supported anyway.

tresf:
> > It may be so. The errors could actually be warnings. I have not built with laxer settings.

> Linux build runs on Trusty, so an older Qt 5 should be supported anyway.

Sure but if the problem has been fixed upstream, isn't this overkill?

jasp00:
>Sure but if the problem has been fixed upstream, isn't this overkill?

No. Trusty is not fixed. Some users build on Trusty. Travis builds on Trusty. As you can see, the builds are working. You need to configure with `-DFIX_SYSTEM_HEADERS=ON` to enable the rewriting, so it is controlled. Besides, `/usr/include/gig.h` needs to be fixed too, the fix being in experimental packages on some distros.

tresf:
> needs to be fixed

Mac builds with Clang on each release.  I feel your use of the word "need" is unwarranted.  If we need to rewrite headers, great.  If you are just trying to avoid compilation warnings, I strongly feel this is overkill.

Lukas-W:
I agree with Tres. Another option to easily silence the warnings is just telling Clang not to show any warnings for these files by [registering them as system headers](https://clang.llvm.org/docs/UsersManual.html#controlling-diagnostics-in-system-headers). Adding a command line argument like `--system-header-prefix=/usr/` should silence them.

tresf:
> `--system-header-prefix=/usr` should silence them.

Thanks for sharing.  FYI, we'll need to be careful to make this version dependant due to this bug:

http://stackoverflow.com/a/23342233/3196753

tresf:
**Edit:** and yes, I believe we're affected by `-isystem-prefix vs. --system-header-prefix` due to [macOS 10.8 ~= XCode 5.1.1 ~= Clang 3.4](https://gist.github.com/yamaya/2924292#file-xcode-clang-vers-L71-L74)

jasp00:
@tresf, you have [asked](https://github.com/LMMS/lmms/pull/3208#issuecomment-274817593) what is worse, whether to build with laxer settings or to rewrite system headers. I have answered, stricter settings are better. Warnings in one system might become errors in another. Warnings are solved by fixing them, not by hiding them.

You have [asked](https://github.com/LMMS/lmms/pull/3208#discussion_r94716882) for a header rewrite. I have given you that and it works. Could you tell me what is actually wrong with fixing system headers? Why a bad header is better than a good one?

tresf:
The advantage of keeping headers as-is is is simplicity, which the project needs more of.

I feel that is easier to uncomment two lines of code prefixed with `# TODO: Remove once Travis ships with Qt 5.7` then to patch source-code on the fly.  Developers should see and report the errors (warnings) upstream.  I think the patching is overestimating the importance of these warnings.    If the argument is that a standard user wouldn't be using the rewrites, then the severity argument is flawed.  Users will build and use the software they develop, no?  If I'm wrong, please explain.

jasp00:
>`# TODO: Remove once Travis ships with Qt 5.7`

We can add comments about which versions are fixed. Travis is not the only affected environment.

>Developers should see and report the errors (warnings) upstream.

They will see the errors, they may [report](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=850179) the errors, and they will use `-DFIX_SYSTEM_HEADERS=ON` or fix the headers themselves.

>If the argument is that a standard user wouldn't be using the rewrites, then the severity argument is flawed.

A standard user will not use the rewrite because she will not get the error (warning). The special user that builds with Clang, warning-clean, without sudo powers or the knowledge to fix a header, will use `-DFIX_SYSTEM_HEADERS=ON`.

include/debug.h
```diff
@@ -34,6 +34,10 @@
 #ifdef LMMS_DEBUG
 	#include <assert.h>
 #else
+	// TODO: assert is a standard macro, we should choose another name
```
tresf: > we should choose another name

Isn't this quite common practice though?

tresf:
@jasp00 I'd like to merge this, but I think @Lukas-W's proposal is cleaner for the headers. https://github.com/LMMS/lmms/pull/3208#issuecomment-275694687

include/debug.h
jasp00: >Isn't this quite common practice though?

Using `assert` for operations with side effects? Yes, it is a common error. We should not change the meaning of standard `assert`.

jasp00:
>I think @Lukas-W's proposal is cleaner for the headers.

>Adding a command line argument like `--system-header-prefix=/usr/` should silence them.

Please, let us not hide bugs. I do not understand your skepticism. If a user wants to rewrite her headers, she must say "I want to rewrite my headers, fix them for me". What is wrong with that? Do we not provide other workarounds?

If we should not offer a rewriting option, then let me fix the Travis environment only. If fixing on setup is not acceptable, then let me add an extra repository so that fixed packages can be used. Let us build on a fixed environment.

tresf:
> Please, let us not hide bugs.

I've read the commit as well as the comments and I'm not sure what's being fixed.  Can you be explicit in the PR so that we can dispute the difference between a `bug` and a benign `warning` ?

jasp00:
Warnings are not errors, but they are not benign either. Warnings in one system might become errors in another.

I will add code comments that point to this report. The problems are:

1.
```
/usr/include/gig.h:434:27: error: ISO C++11 does not allow access declarations;
      use using declarations instead
```
If we use C++11, there is no choice but to fix the header.

2.
```
/usr/include/x86_64-linux-gnu/qt5/QtCore/qvector.h:767:21: warning: destination
      for this 'memmove' call is a pointer to class containing a dynamic class
      'QPixmap'; vtable pointer will be overwritten [-Wdynamic-class-memaccess]
```
Turning off this warning would hide bad uses of `memmove`, either direct or indirect. An example would be to try to switch objects of parent and child classes.

Lukas-W:
@jasp00 

> 1. If we use C++11, there is no choice but to fix the header.

So let's just not use C++11 for `GigPlayer` until libgig supports it upstream. 

> 2. Turning off this warning would hide bad uses of memmove

Adding a `static_cast` like you do [here](https://github.com/LMMS/lmms/pull/3208/files#diff-95e351a3805a1dafa85bf20b81d086e6R104) is no different. We'd still be calling `memmove` on a dynamic class. The cast merely silences the warning.

Could you point us to the template instantiation we're talking about?

jasp00:
>until libgig supports it upstream

Upstream does support it, distros do not. Anyway, we can wait five years.

>let's just not use C++11 for `GigPlayer`

Agreed if the other rewrite is not needed.

>Adding a `static_cast` like you do here is no different. We'd still be calling `memmove` on a dynamic class. The cast merely silences the warning.

Because this specific case is safe.

>Could you point us to the template instantiation we're talking about?

These are the [problematic](http://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/tools/qvector.h?h=5.6#n757) and [fixed](http://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/tools/qvector.h#n763) versions.

Lukas-W:
> Because this specific case is safe.

And yet you rewrite the headers, silencing the warning for all uses. Still, what specific case are we talking about here? Links to `qvector.h` don't answer that question.

jasp00:
>And yet you rewrite the header, silencing the warning for all uses.

Let me explain. Turning off the warning using `-Wno-dynamic-class-memaccess` hides any bad use of `memmove`, `memset`, `memcpy`, etc. It is fine to silence the warning in `qvector.h` because this specific use of `memmove` in the header is correct by itself; this is what upstream does. I do not silence other bad calls to `memmove`, `memset`, etc.

>Still, what specific case are we talking about here? Links to `qvector.h` don't answer that question.

Any call to `QVector::erase()` with dynamic classes. One case is [`NStateButton.cpp`](https://github.com/LMMS/lmms/blob/master/src/gui/widgets/NStateButton.cpp#L47).

Lukas-W:
> in qvector.h […] this specific use of memmove in the header is correct by itself

Doesn't that depend on the template argument, i.e. the type you store in the vector?

> One case is NStateButton.cpp.

It also appears to be the only case. I suggest we just add `#pragma clang diagnostic ignored "-Wdynamic-class-memaccess"` to `NStateButton.cpp`.

jasp00:
>Doesn't that depend on the template argument, i.e. the type you store in the vector?

The warning will be triggered depending on the type, but the algorithm is always correct.

>I suggest we just add `#pragma`

That technique would hide other cases and should be avoided. Indeed, the one `erase()` in `NStateButton.cpp` is the only case. `#pragma` effects can be limited using `push`, `QT_VERSION` and `pop`, and it works.

However, when disabling C++11 for GIG Player, I get:
```
/usr/include/x86_64-linux-gnu/qt5/QtCore/qcompilerdetection.h:562:6: error: Qt
      requires a C++11 compiler and yours does not seem to be that.
```
There is no choice.

Preventing the user to fix her system means discarding a valid tool. No one is giving a justification not to rewrite headers.

Lukas-W:
> I believe there are no objections to the rewrite option

There are objections, please don't merge.

As far as I can see, both @tresf and I have objected to rewriting any of the two headers, and have also stated why we think so and have suggested multiple alternatives. Yet you write "No one is giving a justification not to rewrite headers". For the record, these are my reasons: In the Qt case, it's overkill to introduce this to merely fix a warning. Generally speaking, I strongly oppose fixing upstream bugs (let alone warnings) in our code, especially if upstream already fixed them. I don't like the complexity introduced with this kind of fixes, and I don't feel it's even our responsibility.

As for `QVector`: There are several methods of silencing the warning that don't involve regex-fixing a header. We can also just build with relaxed settings.

As for `gig.h`: Upstream has fixed this two years ago in version 4.0. We could just require that version for the `GigPlayer` plugin. Let's look into that option first.

jasp00:
Yes, before I showed there is no choice, @tresf and you did state an objection, overkill (excessive complexity), which is not a justification unless there is a simpler way to achieve the same result, and there is not. You have suggested several alternatives and I have tried them all. Without this solution, you are just banning environments.

>There are several methods of silencing the warning that don't involve regex-fixing a header

And all of them could hide potential bugs.

>We can also just build with relaxed settings.

Which means that we do not fix warnings in continuous integration.

>We could just require that version for the `GigPlayer` plugin.

You are forbidding any usable Debian system (and Ubuntu, of course) with Clang. Is requiring a full package any simpler than rewriting a header?

I am repeating myself. This is an *optional* flag. Do we not have other workarounds? No one is giving a *sensible* justification. Please explain how to accomplish the same result in a simpler way. If you do not understand the code, then I add more comments.

tresf:
> excessive complexity [...] is not a justification unless there is a simpler way to achieve the same result

I'll reiterate the discussion... 

1. Qt ships with a header that barks with a warning on clang compiler.  If we want to leverage strict Travis-CI compiler settings for the Clang environment (I think we all do) we need to find a way to get past this warning.
   ```
   /usr/include/x86_64-linux-gnu/qt5/QtCore/qvector.h:767:21: warning: destination
         for this 'memmove' call is a pointer to class containing a dynamic class
         'QPixmap'; vtable pointer will be overwritten [-Wdynamic-class-memaccess]
   ```

   A proposal by @Lukas-W is to instruct the compiler that this is a system header that we have no control over, either through an `ifdef` flag on each `qvector.h` include or alternately through a compile setting.  @jasp00 is worried `-Wno-dynamic-class-memaccess` for the entire project is unsafe.  @jasp00 also advises against `--system-header-prefix=/foo/bar`.  I personally feel  `--system-header-prefix=/foo/bar` is an adequate compromise.

2. Gig player simply won't work on Clang unless updated.  Some repos already have the updated GigPlayer, but Travis-CI does not.  This affects users on many popular distributions building LMMS using Clang using GigPlayer -- something LMMS has never really offered prior to this PR.
   ```/usr/include/gig.h:434:27: error: ISO C++11 does not allow access declarations;
      use using declarations instead
   ```
   A proposal by @Lukas-W is to drop GigPlayer support < 4.0 as 4.0+ fixes this problem.  @Lukas-W did not specify if he meant drop it altogether or if he meant drop it only for Clang.  @jasp00 has concerns that the entire feature shouldn't be dropped over a trivial header change.  This concern has merit.  I personally feel that GigPlayer is largely unused plugin which has never worked on these systems with Clang to begin with, so this is a few feature and CAN wait for upstream to fix it.  For those rare occurrences that need GigPlayer without access to the newer version on the mentioned compiler, they can edit the header themselves, since the header is useless on Clang anyway.  i.e. They're going to need it fixed for any other projects that use the header so why burned the project?

Whether we use it or not, @jasp00 I have to thank you for the work that has gone into this header patching as well as the patience and time taken to justify the changes however stating that all of the points above are not "sensible justification[s]" is weighing this feature higher than simplicity itself, which we have a clear divide over. Once which we'll need to overcome to get this into Travis. :)

jasp00:
Pending SWH fixes merged, https://github.com/swh/ladspa/pull/47.

jasp00:
I am all for simplicity, but I am running out of alternatives. My concern is the Travis environment. May I just copy the headers in the Travis scripts? `sha256sum` plus `cp`, it is simple and we do more complex stuff.

tresf:
The decision was made not to rewrite system headers in this fashion, but mark the headers as `SYSTEM` headers and simplify the codebase.

There are several Clang fixes in here that haven't made it into the codebase, but many of them are for 3rd party LADSPA plugins and have since been moved to submodules.  New PRs should be opened at the upstream homes, then we'll fast forward the submodules once accepted.

PR #3211, opened 2017-01-06
Update .travis.yml

liushuyu:
Fix Travis autodeploy

---
This change is [Reviewable](https://reviewable.io/reviews/lmms/lmms/3211)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ deploy:
     on:
         tags: true
         all_branches: true
-        condition: "($TARGET_OS = win*) && ($QT5 = true)"
+        condition: "(${TARGET_OS} == win*) && (${QT5} == true)"
     repo: LMMS/lmms
 notifications:
     webhooks:


tresf:
@liushuyu can you explain this please?  Looks different than the documentation.

https://docs.travis-ci.com/user/deployment#Conditional-Releases-with-on%3A

liushuyu:
According to their documentation, this expression will be pasted to `if [[ <condition> ]]; then <deploy>; fi` and I just wonder if this will work, because this is a valid Bash expression.

liushuyu:
And I suspect if this is a big of Travis CI. It says it can't execute `getcwd()` normally, this is impossible unless the current folder has been removed.

jasp00:
The solution was to add `skip_cleanup`, you should revert to the previous condition. New one is invalid (no wildcards, case-sensitive).

jasp00:
Use `"$TARGET_OS != linux"`.

Umcaruje:
>Use "$TARGET_OS != linux".

This would make travis build packages for mac as well, and would end up uploading double files for windows. That's why I changed the condition

Umcaruje:
>no wildcards, case-sensitive

so `condition: "($TARGET_OS = win32 || $TARGET_OS=win64) && $QT5 = True"`?

jasp00:
My bad, I was thinking about `[` instead of `[[`; it is case-sensitive, though. So:
`condition: "$TARGET_OS = win* && $QT5 = True"`

zonkmachine:
Superseded by 4db1db5ecaa183b24525296e8d344bf789b78f00, closing.

PR #3212, opened 2017-01-06
Xpressive plugin

gnudles:
sorry about the tabs. my editor adds spaces instead of tabs and Iv'e noticed it too late.
the expressive plugin is not included. this is only some fixes to support it. (it uses QPlainTextEdit and also the new convolution method in Graph.cpp)


Umcaruje:
Hello, thanks for contributing to LMMS. Can you please tell us more about what this plugin is about? Also, you'll need to set your editor to use tabs, or your pull request won't be accepted.

gnudles:
Can one please only perform the changes in  src/gui/editors/AutomationEditor.cpp  src/gui/TrackContainerView.cpp src/gui/PianoView.cpp in one's branch, and merge them to the master branch.
right now I don't have the time to remake that commit... and these are important changes.
And for your question, my plugin uses expression parser so one can write his own formula to generate sound. something like "sine_wave(t*f)+square_wave(t*f*cent(3))*0.4"

tresf:
@gnudles I think you forgot to add your actual plugin source files.  You can add them with `git add`.  You seem to only  have committed the changes to existing files.

> Can one please only perform the changes in `src/gui/editors/AutomationEditor.cpp` `src/gui/TrackContainerView.cpp src/gui/PianoView.cpp` in one's branch, and merge them to the master branch.

I was just going to ask you the same question.  Please open a Pull Request explaining these important changes and why so they can be merged in without relying on the plugin testing and approval.

gnudles:
the plugin is not ready to be released yet but it's working. the problem is that it uses ExprTk (https://github.com/ArashPartow/exprtk) which is licensed under CPL. is it OK?

jasp00:
>which is licensed under CPL.

This seems to be a [problem](https://www.gnu.org/licenses/license-list.html#CommonPublicLicense10).

gnudles:
oh no! I guess I'll have to use MuParser which uses MIT license. or write my own (which is not considered as a good idea)

jasp00:
Or you may want to explain your situation to ArashPartow if you are serious about using the library. It would not be the first [project](http://www.wykobi.com/license.html) under GPL.

gnudles:
That's a good Idea. First, because ExprTk is capable of using "float" type which is less consumable. Second, I added support for inserting a function to the parser with internal data. and it is harder to do so in MuParser.
BTW. ExprTk uses exception handling, and LMMS is compiled with the counter exception flag.
I removed the exceptions from ExprTk but I don't fill comfort about it. any ideas?

ArashPartow:
https://github.com/ArashPartow/exprtk/pull/8#issuecomment-271165839

Umcaruje:
Closing this pull request, superseded by #3222 and #3221

PR #3213, opened 2017-01-06
Move private functions to private files (example)

jasp00:
One frequently asked question is how to avoid writing private member functions to public headers. I propose the following solution: move all private information that is not required to be known to a private file; this comprises non-virtual functions, non-slots, static members, and types. Runtime behavior does not change.

`DataFile` is a good example. You may add a private helper function, such as an upgrade method, and `make` will just recompile `DataFile.cpp`.

### DIFF
--- a/src/include/DataFile_priv.h
+++ b/src/include/DataFile_priv.h
@@ -1,0 +1,39 @@
+#ifndef DATA_FILE_PRIVATE_HEADER
+#define DATA_FILE_PRIVATE_HEADER "DataFile_priv.h"
+#else
+
+static Type type( const QString& typeName );
+static QString typeName( Type type );
+
+void cleanMetaNodes( QDomElement de );
+
+// helper upgrade routines
+void upgrade_0_2_1_20070501();
+void upgrade_0_2_1_20070508();
+void upgrade_0_3_0_rc2();
+void upgrade_0_3_0();
+void upgrade_0_4_0_20080104();
+void upgrade_0_4_0_20080118();
+void upgrade_0_4_0_20080129();
+void upgrade_0_4_0_20080409();
+void upgrade_0_4_0_20080607();
+void upgrade_0_4_0_20080622();
+void upgrade_0_4_0_beta1();
+void upgrade_0_4_0_rc2();
+void upgrade_1_0_99();
+void upgrade_1_1_0();
+void upgrade_1_1_91();
+
+void upgrade();
+
+void loadData( const QByteArray & _data, const QString & _sourceFile );
+
+
+struct EXPORT typeDescStruct
+{
+	Type m_type;
+	QString m_name;
+} ;
+static typeDescStruct s_types[TypeCount];
+
+#endif
--- a/include/DataFile.h
+++ b/include/DataFile.h
@@ -104,39 +104,9 @@ public:
 
 
 private:
-	static Type type( const QString& typeName );
-	static QString typeName( Type type );
-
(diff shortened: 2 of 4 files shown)


tresf:
@jasp00 can you explain the real-world use-case/need for this?  Currently few projects use our dev headers.  If cleanliness it he concern the `DataFile.h` you mention could very well just abolish the upgrade routines and stash them in a dedicated class (right?)

jasp00:
>can you explain the real-world use-case/need for this?

The case is any coder developing, debugging, optimizing, etc. any non-small C++ project. In these cases, it is important to minimize recompilation. In C++, you may make a change to the private interface of a class, such as adding an auxiliary function, but triggers the recompilation of all related classes, even though they cannot use that private interface.

To solve this problem, there are some decoupling techniques. A popular one is the [pImpl idiom](http://en.cppreference.com/w/cpp/language/pimpl), but it has some drawbacks, such as double `new`/`delete`. Another one is the use of [factories](https://en.wikipedia.org/wiki/Factory_%28object-oriented_programming%29), but methods become virtual.

The solution I am presenting does not have any runtime penalty. Like the alternative techniques, it is a matter of organization and learning the pattern. While this procedure can benefit any project, this request is not because of developers that use LMMS headers, but because of developers of the LMMS project itself.

tresf:
How are friend classes affected?

tresf:
@jasp00 how do other projects deal with the inevitable header file explosion?  As it is, we already have > 200 header files in `lmms/include`.  Have projects like LMMS gone away from the `/include` design and adopted a side-by-side `Class.cpp, Class.h`  type structure?

jasp00:
When dealing with friend classes, the required methods go to the public header, but the friend class may include the private header instead, even a third header if that suits better.

Now I am checking with Qt 5. Projects with many headers use folders (QtCore, QtGui, etc.). > 200 header files are fine; we should have more, one per class. I do not know about a C++ project similar to LMMS in size, but projects do decouple; Qt uses pImpl. As far as I can tell, the `/include/Class.h`, `/src/Class.cpp`, `/src/Class_priv.h` structure has never been publicly implemented, so we would be on the innovation side.

jasp00:
What about this one? Yet another upgrade path (#3281). Do we go ahead?

Another option is to use an auxiliary friend class. Private methods in the target class become static methods in the auxiliary class, and operations like:
`myPrivateCalc() + x + y;`
turn into:
`t->myPrivateCalc() + t->x + t->y;`
which makes code less movable.

tresf:
> What about this one? Yet another upgrade path (#3281). Do we go ahead?

I think it's a good time to add this feature to master although the more headers we split like this the the more confusing it is going to be to a new developer.

Also this has it's main benefits on benign​ yet commonly edited files like `DataFile.cpp`.  I'm not sure the added complexity is going to be worth it on every header, but mostly those with a large list of constantly growing private functions.

Are you aware of any others that could really speed up builds with this design?

I vote we rebase and merge but I think we owe ourselves a bit of direction in regards to how we police this moving forward.

src/include/DataFile_priv.h
```diff
@@ -0,0 +1,39 @@
+#ifndef DATA_FILE_PRIVATE_HEADER
```
tresf: To match our current standards, this should be `DATA_FILE_PRIV_H`.

tresf:
Closing, the OP has left the project.  This is a noble concept and Qt uses it for some of it's classes to reduce compile time, but it's getting closed out for now.

PR #3216, opened 2017-01-07
Iterate PulseAudio mainLoop and wait for context state

fholmer:
Hi,

after looking into an issue where Pulse Audio did not produce sound, I came to the following conclusion:
in order for the context state to advance, mainloop have to iterate

### DIFF
--- a/src/core/audio/AudioPulseAudio.cpp
+++ b/src/core/audio/AudioPulseAudio.cpp
@@ -227,9 +227,9 @@ void AudioPulseAudio::run()
 	// connect the context
 	pa_context_connect( context, NULL, (pa_context_flags) 0, NULL );
 
-        while (!m_connectedSemaphore.tryAcquire()) {
-          pa_mainloop_iterate(mainLoop, 1, NULL);
-        }
+	while (!m_connectedSemaphore.tryAcquire()) {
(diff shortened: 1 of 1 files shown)


tresf:
@fholmer minor change, but can you adjust your branch to use `tab` instead of `space` to copy the existing file formatting?  It will be reflected in your pull request automatically.

tresf:
Thanks.  @fholmer since this loop seems to keep trying over and over...  is there any chance of deadlock with this while loop?

fholmer:
Yes, if context never advance to ready state or failed state then we have a deadlock, because m_connectedSemaphore is not released.

This is a issue already present in the code and this patch does not solve this issue.

The only improvement here is that we are calling pa_mainloop_iterate while we wait for the semaphore.

The call to pa_mainloop_iterate then solves the deadlock I was having where context never advance to ready state.  :)

tresf:
> The call to pa_mainloop_iterate then solves the deadlock I was having where context never advance to ready state. :)

Just read [the api](https://freedesktop.org/software/pulseaudio/doxygen/mainloop_8h.html#a491422aebe5487800d8be0bf4153cb81), reads about right, thanks.

PR #3217, opened 2017-01-07
MIDI import changed into drumstick

tonychee7000:
It's exciting.
We have strong tools to import MIDI now.
Here is some compatibility problems with Qt4 you know.
The rest work is yours.


tonychee7000:
Oops, I pulled wrong repo.

PR #3219, opened 2017-01-08
Fix opening of project files on macOS

### LINKED ISSUE #665: Mimetype Support (Apple)
The current OS X build does not register the MMP/MMPZ file extension with the Apple OS X operating system.

image

This bug report is a placeholder for this feature as it applies to the OS X DMG installer.

tresf:
This is intended to fix the problem where LMMS won't open a project that has been double-clicked on macOS.

Closes #665

Although this fix is Mac-specific, it should have no negative impact on Windows or Linux.

### DIFF
--- a/src/gui/MainApplication.cpp
+++ b/src/gui/MainApplication.cpp
@@ -1,0 +1,68 @@
+/*
+ * MainApplication.cpp - Main QApplication handler
+ *
+ * Copyright (c) 2017-2017 Tres Finocchiaro <tres.finocchiaro/at/gmail.com>
+ *
+ * This file is part of LMMS - http://lmms.io
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this program (see COPYING); if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA 02110-1301 USA.
+ *
+ */
+#include "MainApplication.h"
+
+#include <QDebug>
+#include <QFileOpenEvent>
+
+#include "Engine.h"
+#include "GuiApplication.h"
+#include "MainWindow.h"
+#include "Song.h"
+
+MainApplication::MainApplication(int& argc, char** argv) :
+	QApplication(argc, argv),
+	m_queuedFile() {}
+
+bool MainApplication::event(QEvent* event)
+{
+	switch(event->type())
+	{
+		case QEvent::FileOpen:
+		{
+			QFileOpenEvent * fileEvent = static_cast<QFileOpenEvent *>(event);
+			if(event)
+			{
+				// Handle the project file
+				m_queuedFile = fileEvent->file();
+				if(Engine::getSong())
+				{
+					if(gui->mainWindow()->mayChangeProject(true))
+					{
+						qDebug() << "Loading file " << m_queuedFile;
+						Engine::getSong()->loadProject(m_queuedFile);
+					}
+				}
+				else
+				{
+					qDebug() << "Queuing file " << m_queuedFile;
+				}
+				return true;
+			}
+			// Fall through to default
+		}
+		default:
+			return QApplication::event(event);
+	}
+}
--- a/include/MainApplication.h
+++ b/include/MainApplication.h
@@ -1,0 +1,43 @@
+/*
+ * MainApplication.h - Main QApplication handler
+ *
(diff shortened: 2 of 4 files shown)


jasp00:
>    * What is the best approach to prevent crash?

It looks like you are trying to load the project before the engine is initialized. If `Engine::getSong()` is null, then delay the loading; you should edit `fileToLoad` in `main.cpp` somehow.

>    * How is best to incorporate feature into codebase?

You should use a separate file for the class. `LmmsApplication` or `MainApplication` would be better names since this the `QApplication` derivative.

tresf:
> It looks like you are trying to load the project before the engine is initialized.

That makes sense.

> If Engine::getSong() is null, then delay the loading; you should edit fileToLoad in main.cpp somehow.

I'm not sure the second part of that statement, but it appears we could be missing quite a bit of logic here.  Shouldn't this be decoupled from `main.cpp` and placed into some GUI init function? https://github.com/LMMS/lmms/blob/1f3947408d1ed8c4d91c8e3215b58a94addece63/src/core/main.cpp#L714-L899


> You should use a separate file for the class. LmmsApplication or MainApplication would be better names since this the QApplication derivative.

Sure but even if it's a single function only used in `main.cpp`?  I find the lack of mid-declaration overriding in this language to be quite frustrating, all we need is the `::event(...)` function. 😐

tresf:
OK I think I get this...

We have two separate use-cases.

1.  LMMS is open, just open the file (not a new QApplication)
2.  New QApplication 

The ladder simply needs to set a member variable in main.cpp instructing which file to load, right?

tresf:
..  or would it be simpler to just loop and wait for the Engine to be non-NULL and hit two birds?

jasp00:
>Sure but even if it's a single function only used in `main.cpp`?

It is about organization. When you see `EventApp` in a backtrace (or in any source file), it is faster to get to the code with `find -name EventApp.cpp` than `grep -r "EventApp::method(" source_dirs`.

Regarding the implementation:

1. In `event()`, set `m_eventFileToLoad` if song is null.
2. In [`main.cpp`](https://github.com/LMMS/lmms/blob/1f3947408d1ed8c4d91c8e3215b58a94addece63/src/core/main.cpp#L843), if `fileToLoad` is empty, set it to `app->eventFileToLoad()`.

tresf:
Won't that require dynamic casting back in main.cpp since app is a QApplication, not a EventApp?


> When you see EventApp in a backtrace (or in any source file), it is faster to get to the code with find -name EventApp.cpp than grep -r "EventApp::method(" source_dirs

Fair enough but the code is riddled with nested classes, I assume you wanna keep'm separated moving forward?  I just can't bare to add two new files for such a small change, but if that's the way to do it, that's just fine.

jasp00:
>Won't that require dynamic casting ?

Static casting will do since you are running the GUI case.

>the code is riddled with nested classes, I assume you wanna keep'm separated moving forward?

Outer classes in a file with a different name are a problem. Nested classes are fine.

>I just can't bare to add two new files for such a small change

I understand how you feel, though I would add those files. On the other hand, LMMS is not that big yet. If you decide to keep `EventApp` in `main.cpp`, you should keep the class declaration too.

tresf:
> If you decide to keep EventApp in main.cpp, you should keep the class declaration too.

You mean the header contents?  That's how I had it originally but it was god awful ugly.  I'll split it out, it may come in handy for other Mac events or perhaps non-Mac events down the road.

tresf:
> I understand how you feel, though I would add those files. On the other hand, LMMS is not that big yet. If you decide to keep `EventApp` in `main.cpp`, you should keep the class declaration too.

Done.

> Static casting will do since you are running the GUI case.

Done.

> In `event()`, set `m_eventFileToLoad` if song is null.

Done.

> In `main.cpp`, if `fileToLoad` is empty, set it to `app->eventFileToLoad()`

Done, in a slightly different way.  If the event file is not empty, we use it.

This is ready for another review.  Tested on already running instances as well as newly fired instances.  Thanks for the pointers.

src/gui/MainApplication.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include "MainWindow.h"
+#include "Engine.h"
+#include "Song.h"
+
+MainApplication::MainApplication(int& argc, char** argv) :
+	QApplication(argc, argv),
+	m_queuedFile() {}
+
+bool MainApplication::event(QEvent* event)
+{
```
tresf:
```cpp
if(a)
{
        // THIS
        if(b)
        {
                // IS
        }
elseif(c)
{
        if(d)
        {
        // I LOATH INDIVIDUAL LINE BRACES
        }
}
```

```cpp
if(a) {
        // THIS
        if(b) {
                // IS WHY
        }
} elseif(c) {
        if(d) {
                // I PROPOSE WE STOP THE MADNESS
        }
}    
```

tresf:
If there are no objections, I will merge this in 2 days.

src/gui/MainApplication.cpp
```diff
@@ -0,0 +1,68 @@
...
+	switch(event->type())
+	{
+		case QEvent::FileOpen:
+		{
+			QFileOpenEvent * fileEvent = static_cast<QFileOpenEvent *>(event);
+			if(event)
```
jasp00: This condition will always be true.

tresf: Can you please explain?  Also, why drop the non-obvious fall-through comment?

PR #3221, opened 2017-01-08
Small fixes:

gnudles:
In PianoView.cpp fixed the focus policy so that if one would like to use
a QPlainTextEdit in his plugin, he'd be able to do so.
In TrackContainerView.cpp and AutomationEditor.cpp there's a problem
that if you use a Right To Left locale, all the GUI turn RTL, but in
those specific editors, things get very messy, because they were made up
with the assumption of LTR.


tresf:
@gnudles thanks!

PR #3222, opened 2017-01-08
Xpressive plugin

gnudles:
First Preview of the X-Pressive Plugin
    (exprtk.hpp is not included, get it from my exprtk fork in the branch
    internal_data_functions)
    available keys:
    f- note's frequency. available only in the output expressions
    t- time in seconds. in the Waves (W1,W2,W3) it's in the range [0,1)
    key- the note's keyboard key. available only in the output expressions.
    v- the note's velocity (divided by 255.0 so it is in the range [0,1]).
    available only in the output expressions
    A1,A2,A3- general purpose knobs (you can control them with the
    automations). available only in the output expressions.
    W1,W2,W3- precalculated wave forms. can be also load from file. you can
    use them only in the output expressions
    available functions:
    cent(x)- gives pow(2,x/1200)
    rand()- random number generator. in range [-1,1). each call gives other
    value.
    randv(i)- random vector (with pseudo infinite integer cells). the values
    are in range [-1,1). it's stays consistent only across the note
    playback. so each note playback will get other vector (even on the same
    key).
    sinew- sine wave with period of 1.
    saww- saw wave with period of 1.
    squarew- square wave with period of 1.
    trianglew- triangle wave with period of 1.
    expw- exponent wave with period of 1.
    expnw- another exponent wave with period of 1.
    moogw- moog wave with period of 1.
    moogsaww- moog-saw wave with period of 1.
    you can use * % ^ / + - pow sin log pi etc.


gnudles:
The checks fails because the missing exprtk.hpp. it is not included yet because of license issues

gnudles:
Arash Partow the author of ExprTk, said that he'd be happy to provide it for my plug-in under either MIT license or Boost license. Is it acceptable? and which one is preferred?

tresf:
@gnudles http://softwareengineering.stackexchange.com/a/105926

Looks like you should be safe to snag it MIT and then restrict it GPL.  We don't use MIT internally AFAIK so we shouldn't start with this library.  Interested to hear feedback from @jasp00 on this one.

gnudles:
I uploaded a new version.

jasp00:
>We don't use MIT internally AFAIK so we shouldn't start with this library.

MIT license should be referred to as Expat license, since MIT has used different licenses. We are currently using Expat for Portsmf (MidiImport) and sfxr. It is fine to use Expat; Boost license is [good](https://www.gnu.org/licenses/license-list.html#boost) too.

Issue #2611, opened 2016-02-25
Switching to Qt5


tresf:
Many of our users are already compiling against Qt5, but most of them tend to be on Linux.  This topic is related to switching all builds to Qt5 permanently and what's needed prior to that switch.
- For starters, we need more testing.
  - Our first Win64 build which utilizes Qt5 is located here:<br> *(removed, please download 1.2.0 from the [downloads](https://lmms.io/download) page)*
  - Our first Mac build which utilizes Qt5 is located here:<br> *(removed, please download 1.2.0 from the [downloads](https://lmms.io/download) page)*
- Bugs (please check these off as they're fixed)
  - [x] All: Frequent crashes related to `QVector` https://github.com/LMMS/lmms/issues/2192#issuecomment-193905099.
  - [x] Mac: Spash window appears above SettingsDialog, MessageDialog (fixed via 1058ea4)
  - [x] All: Visual artifacts in MDI window decorations ([preview](https://cloud.githubusercontent.com/assets/6345473/13335913/6a2a6f48-dbe1-11e5-9497-720be1d11750.png)) (superseded by #2516)
  - [x] All: Menu bar colors are hard to read ([preview](https://cloud.githubusercontent.com/assets/6345473/13536660/2a22bb60-e20f-11e5-8b23-0fda60c21fa6.png))  (fixed via 5f48d1c)
  - [x] All: Inactive QMdiWindow titles are hard to read  ([preview](https://cloud.githubusercontent.com/assets/6345473/13536646/1a019436-e20f-11e5-81ec-56afdffe8a46.png)) (superseded by #2516)
  - [x] All: Language dropdown is blank (fixed via https://github.com/LMMS/lmms/commit/85011cdcf72fbba721bb0c10f140bfabf7a35725)
  - [x] All: White line on edge of all `QWidgets` ([preview](https://cloud.githubusercontent.com/assets/6345473/13538614/ab01bace-e219-11e5-8638-48204a41a6b7.png)) (superseded by #2516)
  - [X] All: `QMdiWindow` windowing controls are incorrectly colored ([preview](https://cloud.githubusercontent.com/assets/6345473/13336044/f96ca838-dbe1-11e5-8930-6a54b698b00a.png)) (`wont-fix`, see https://github.com/LMMS/lmms/issues/2611#issuecomment-193942180)
[...]

Issue #793, opened 2014-05-31
Enhancement: Midi Out Plugin


StakeoutPunch:
It would be neat to have a dedicated midi out instrument that's only function is to send midi to instruments in LMMS listening on whatever midi channel it is set to.

As I've read it is possible to use an empty ATP to send midi out, but there is no way to bridge that to instruments inside LMMS without using external programs/devices (which in turn breaks the project if they are configured exactly the same each time the project is opened, which is a pain).

Uses of this would be to use one instrument to trigger many layered things, using midi to control parameters in VST effects that can't be controlled with LMMS's automation, and possibly, once implemented, have a role in exporting midi. There are probably more uses to this but eh. I just want a way to bounce midi around inside LMMS, which currently can't be done within the program to my knowledge.

PR #3224, opened 2017-01-09
MIDI Import feature ported to drumstick and new MIDI Export feature

senseab:
Now I have changed origin portsmf of MIDI Import to drumstick
And with some new features.
* Fix issue #1967
* Adding Overture (.ove) and Cakewalk (.wrk) project file support. with drumstick
* New `MidiExport` plugin

Origin `MidiImport` plugin changed to `SmfImport`


plugins/SmfImport/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+# Importer for SMF-like file format, such as .mid, .ove, .wrk
+
+INCLUDE(BuildPlugin)
+
+# check for drumstick-file
+PKG_CHECK_MODULES(DRUMSTICK_FILE REQUIRED drumstick-file>=1.1.0)
+IF(NOT DRUMSTICK_FILE_FOUND)
+        MESSAGE(FATAL_ERROR "LMMS requires libdrumstick-file and drumstick-dev >= 1.0.11 - please install, remove CMakeCache.txt and try again!")
```
softrabbit: Looks like the as of now unreleased Ubuntu 17.04 will have libdrumstick 1.0.2. Is there something in 1.0.11 that is absolutely required? (https://launchpad.net/ubuntu/+source/libdrumstick)

tonychee7000: well，i was using Archlinux，drumstick may newer. i will fix it later

tonychee7000: Okay, it's typo

Umcaruje:
@tonychee7000 locale updates are done over transifex now, this should not be in the commit for other languages, just English. CC @liushuyu

tonychee7000:
Sorry, i would fix it

Umcaruje:
Also you'll need to add drumstick to packages needed for travis, and mingw packages will need to be made. What are the advantages of drumstick over our current implementation?

tonychee7000:
Reply @Umcaruje 

1. I have not understand travis yet, and I don't know how to make mingw packages. So it will be better if there were someone help.
2. As you see, drumstick can read Overture and Cakewalk project. Then we can support more file types.
3. Drumstick is Qt-based library for MIDI, using Signal/Slot mechanism.
4. It would be easy to procceed with RPN Event when I changed pitch bend range. See #1967 and #3206 

If any doubts, ask @liushuyu

zonkmachine:
`MMS requires libdrumstick-file and drumstick-dev >= 1.0.2`

Debian Jessie use drumstick 0.5.0

```
Changes for v1.0.2
...
Compilation minimum requirements for all platforms: CMake 3.0 and Qt 5.1 
```

We still use Qt 4 on Linux

tonychee7000:
is there any libdrumstick newer  version at Debian ？if not, i will test it

zonkmachine:
> is there any libdrumstick newer version at Debian ？

In experimental only. 1.0.2 https://packages.debian.org/search?suite=experimental&arch=any&searchon=names&keywords=drumstick
I don't know if Debian folks have a problem using packages from experimental.

QT 4 is about to get dropped anyway so that is probably not a big thing. Also I don't know if the libstick version in Debian is a problem. I only use Ubuntu based systems so I'm used to using ppa's. It's a thing with Debian to not update their stable releases so I guess they are cool with running an older lmms too.

tonychee7000:
Got that, I would say that Qt 5 might be more effective. 
But I still wonder the lowest version for my codes, to get the most widely compatibility

zonkmachine:
> But I still wonder the lowest version for my codes, to get the most widely compatibility

Then it looks like maybe 0.5.0 is the best candidate.

tonychee7000:
i will test it

liushuyu:
I would like to say:
1. This PR should be squashed into one commit, for the early commit messages were non-sense.
2. Need more testing, for the changes affected many aspects
3. Test drumstick package under Windows. Seems like `drumstick` is intended for Linux only, so we may need some workarounds to port or compile it for Windows (via MSYS or Cygwin, or just dig out drumstick-file components)
4. For the sake of new comers, @tonychee7000 had expressed the hardness of finding "clues" of each and every APIs in LMMS, let's just try to leave more comments in our code. Not only for developers. A bunch of translators asked me about the context of messages in LMMS, I feel like we need to do something about this.

tonychee7000:
Tested on Debian with drumstick 0.5.0, works well.

tonychee7000:
@liushuyu It seems that I overwrite your commit for mistake. so could you repair it again?

tonychee7000:
I was trying to resolve travis problem and `drumstick-file` needs build from source. Now I can build `libdrumstick-file.dll` successfully. But travis seems need some tests. Code will be merged after all is okay. @Umcaruje

tonychee7000:
In addition, MIDI Export will be availiable with help of drumstick

tonychee7000:
Okay, here is a script that can build `drumstick-file` with mingw
```bash
drumstick_ver=$1

cd /tmp

wget https://master.dl.sourceforge.net/project/drumstick/${drumstick_ver}/drumstick-${drumstick_ver}.tar.gz
tar xf drumstick-${drumstick_ver}.tar.gz
cd drumstick-${drumstick_ver}

# We just need library
sed -i "/ADD_SUBDIRECTORY(utils)/d" CMakeLists.txt
sed -i "/add_subdirectory(utils)/d" CMakeLists.txt

mkdir -p build && cd build

export MINGW=/opt/mingw64
export MINGW32=/opt/mingw32
export PROCESSOR=x86_64
export PROCESSOR32=i686
export MINGW_TOOL_PREFIX=${MINGW}/bin/${PROCESSOR}-w64-mingw32-
export MINGW_TOOL_PREFIX32=${MINGW32}/bin/${PROCESSOR32}-w64-mingw32-

export CMAKE_OPTS="-DUSE_WERROR=ON
                   -DLIB_SUFFIX=''
                   -DCMAKE_INSTALL_PREFIX=$MINGW
                   -DCMAKE_PREFIX_PATH=$MINGW 
                   -DMINGW_PREFIX=$MINGW 
                   -DMINGW_PREFIX32=$MINGW32
                   -DCMAKE_SYSTEM_PROCESSOR=$PROCESSOR
                   -DCMAKE_SYSTEM_PROCESSOR32=$PRPCESSOR32
                   -DWIN64=TRUE 
                   -DCMAKE_SYSTEM_NAME=Windows 
                   -DCMAKE_SYSTEM_VERSION=1 
                   -DCMAKE_FIND_ROOT_PATH=$MINGW 
                   -DCMAKE_C_COMPILER=${MINGW_TOOL_PREFIX}gcc 
                   -DCMAKE_CXX_COMPILER=${MINGW_TOOL_PREFIX}g++ 
                   -DCMAKE_RC_COMPILER=${MINGW_TOOL_PREFIX}gcc
                   -DSTRIP=${MINGW_TOOL_PREFIX}strip
                   -DWINDRES=${MINGW_TOOL_PREFIX}windres
                   -DPKG_CONFIG_EXECUTABLE=${MINGW_TOOL_PREFIX}pkgconfig
                   -DQT_BINARY_DIR=$MINGW/bin
                   -DQT_QMAKE_EXECUTABLE=$MINGW/bin/qmake"
                     
export PATH=$PATH:$MINGW/bin

if [ $QT5 ];then
    (cd ../utils; sed -i "s/add_sub/#add_sub/g" CMakeLists.txt)
fi

cmake ../ $CMAKE_OPTS
make && sudo make install
```
There is an important thing that if use Qt4, drumstick should use `0.5.0` or use `>=1.0.0` if use Qt5.

But I was not sure what can I do for travis. It was f***ing difficult for me.

So anyone help?

PS: OSX build might use this
```bash
drumstick_ver=$1

cd /tmp

wget https://master.dl.sourceforge.net/project/drumstick/${drumstick_ver}/drumstick-${drumstick_ver}.tar.gz
tar xf drumstick-${drumstick_ver}.tar.gz
cd drumstick-${drumstick_ver}

# We just need library
sed -i "/ADD_SUBDIRECTORY(utils)/d" CMakeLists.txt
sed -i "/add_subdirectory(utils)/d" CMakeLists.txt

mkdir -p build && cd build

cmake ../ -DCMAKE_BUILD_TYPE=RelWithDebInfo -DUSE_WERROR=OFF -DLIB_SUFFIX=''
make && sudo make install
```

tresf:
> if use Qt4, drumstick should use 0.5.0 or use >=1.0.0 if use Qt5.  

Can you try:

```cmake
# Drumstick is Qt version dependent
IF(QT_VERSION_MAJOR EQUAL 4)
   PKG_CHECK_MODULES(REQUIRED DRUMSTICK drumstick-file>=0.5)
ELSE()
   PKG_CHECK_MODULES(REQUIRED DRUMSTICK drumstick-file>=1.0)
ENDIF()
```

If that doesn't work, you can check `WANT_QT5` variable.

> But I was not sure what can I do for travis. It was f***ing difficult for me.

For travis, we use three sources for precompiled binaries:

1. apt (Linux)
2. brew (MacOS)
3. @tobydox's apt (Mingw)

If the package is not available in all locations, the build should be commented out on that platform until the repos have binaries ready.

tonychee7000:
@tresf 

>Can you try:
>
>```cmake
># Drumstick is Qt version dependent
>IF(QT_VERSION_MAJOR EQUAL 4)
>   PKG_CHECK_MODULES(REQUIRED DRUMSTICK drumstick-file>=0.5)
>ELSE()
>   PKG_CHECK_MODULES(REQUIRED DRUMSTICK drumstick-file>=1.0)
>ENDIF()
>```
>If that doesn't work, you can check `WANT_QT5` variable.

Thanks for your help, CMake can require different version of drumstick with Qt version.

>For travis, we use three sources for precompiled binaries:
>
>1. apt (Linux)
>2. brew (MacOS)
>3. @tobydox's apt (Mingw)
>
>If the package is not available in all locations, the build should be commented (or naturally skipped) on that platform until the repos have binaries ready.

Actually, I was tested them on travis. `drumstick-file` can build well, but LMMS source always have some problems

Qt4 with MinGW (32/64)
-------

```travis
Linking CXX shared module ../midiexport.dll
/opt/mingw32/lib/gcc/i686-w64-mingw32/4.9.4/../../../../i686-w64-mingw32/bin/ld: cannot find -lQtCore
make[2]: *** [plugins/midiexport.dll] Error 1
make[1]: *** [plugins/MidiExport/CMakeFiles/midiexport.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

```
Linking CXX shared module ../smfimport.dll
/opt/mingw32/lib/gcc/i686-w64-mingw32/4.9.4/../../../../i686-w64-mingw32/bin/ld: cannot find -lQtCore
make[2]: *** [plugins/smfimport.dll] Error 1
make[1]: *** [plugins/SmfImport/CMakeFiles/smfimport.dir/all] Error 2
make: *** [all] Error 2
```

Qt4 on OSX
-------
```
/opt/mingw64/lib/gcc/x86_64-w64-mingw32/4.9.4/../../../../x86_64-w64-mingw32/bin/ld: cannot find -lQtCore
make[2]: *** [plugins/midiexport.dll] Error 1
make[1]: *** [plugins/MidiExport/CMakeFiles/midiexport.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

```
Linking CXX shared module ../smfimport.dll
/opt/mingw64/lib/gcc/x86_64-w64-mingw32/4.9.4/../../../../x86_64-w64-mingw32/bin/ld: cannot find -lQtCore
make[2]: *** [plugins/smfimport.dll] Error 1
make[1]: *** [plugins/SmfImport/CMakeFiles/smfimport.dir/all] Error 2
make: *** [all] Error 2
```

Qt5 with MinGW (32/64)
-------

```
Linking CXX shared module ../smfimport.dll
CMakeFiles/smfimport.dir/objects.a(commonReader.cpp.obj):commonReader.cpp:(.text+0x628): undefined reference to `SmfMidiChannel::setName(QString)'
CMakeFiles/smfimport.dir/objects.a(commonReader.cpp.obj):commonReader.cpp:(.text+0x757): undefined reference to `SmfMidiChannel::create(TrackContainer*, QString)'
CMakeFiles/smfimport.dir/objects.a(commonReader.cpp.obj):commonReader.cpp:(.text+0xd86): undefined reference to `SmfMidiChannel::create(TrackContainer*, QString)'
CMakeFiles/smfimport.dir/objects.a(commonReader.cpp.obj):commonReader.cpp:(.text+0x1130): undefined reference to `SmfMidiChannel::create(TrackContainer*, QString)'
CMakeFiles/smfimport.dir/objects.a(commonReader.cpp.obj):commonReader.cpp:(.text+0x12c5): undefined reference to `SmfMidiCC::putValue(MidiTime, AutomatableModel*, float)'
CMakeFiles/smfimport.dir/objects.a(commonReader.cpp.obj):commonReader.cpp:(.text+0x1ac6): undefined reference to `SmfMidiCC::create(TrackContainer*, QString)'
CMakeFiles/smfimport.dir/objects.a(commonReader.cpp.obj):commonReader.cpp:(.text+0x1e62): undefined reference to `SmfMidiChannel::create(TrackContainer*, QString)'
CMakeFiles/smfimport.dir/objects.a(commonReader.cpp.obj):commonReader.cpp:(.text+0x2082): undefined reference to `SmfMidiCC::putValue(MidiTime, AutomatableModel*, float)'
CMakeFiles/smfimport.dir/objects.a(commonReader.cpp.obj):commonReader.cpp:(.text+0x285b): undefined reference to `SmfMidiCC::create(TrackContainer*, QString)'
CMakeFiles/smfimport.dir/objects.a(commonReader.cpp.obj):commonReader.cpp:(.text+0x2adb): undefined reference to `SmfMidiCC::SmfMidiCC()'
CMakeFiles/smfimport.dir/objects.a(commonReader.cpp.obj):commonReader.cpp:(.text+0x2b1e): undefined reference to `SmfMidiChannel::SmfMidiChannel()'
CMakeFiles/smfimport.dir/objects.a(commonReader.cpp.obj):commonReader.cpp:(.text+0x32ee): undefined reference to `SmfMidiChannel::create(TrackContainer*, QString)'
CMakeFiles/smfimport.dir/objects.a(commonReader.cpp.obj):commonReader.cpp:(.text+0x3460): undefined reference to `SmfMidiChannel::addNote(Note&)'
make[2]: *** [plugins/smfimport.dll] Error 1
make[1]: *** [plugins/SmfImport/CMakeFiles/smfimport.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

Qt5 on OSX
---------

```
ld: library not found for -ldrumstick-file
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [plugins/libsmfimport.so] Error 1
make[1]: *** [plugins/SmfImport/CMakeFiles/smfimport.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

So what can I do next?

For more information, see [here](https://travis-ci.org/tonychee7000/lmms/builds/194438965)

plugins/SmfImport/CMakeLists.txt
```diff
@@ -1,9 +1,17 @@
 # Importer for SMF-like file format, such as .mid, .ove, .wrk
 
 # check for drumstick-file
-PKG_CHECK_MODULES(DRUMSTICK_FILE REQUIRED drumstick-file>=0.5.0)
+IF(WANT_QT5)
+    PKG_CHECK_MODULES(DRUMSTICK_FILE REQUIRED drumstick-file>=1.0.0)
+ELSE(WANT_QT5)
```
tresf: Please use `ELSE()` ... `ENDIF()`  syntax, it's valid and easier to read.

tonychee7000: Okay, I will fix it

tonychee7000:
For travis, now Qt5 with MinGW is okay

see [here](https://travis-ci.org/tonychee7000/lmms/builds/194686942)

But don't worry, i will squash that branch after all works well.

tresf:
> For travis, now Qt5 with MinGW is okay

Great.  Linux Qt5 won't work unless you shim yet another compile script for 14.04 since it defaults to `libdrumstick0`, which is Qt4 only.  When I had previously stated to disable this for platforms which didn't offer binaries, I spoke prematurely since this is a replacement for MIDI import.

We generally don't script tarball downloads and installs into our build process, but rather work with the upstream mirror process.  @tobydox is usually pretty quick to offer up the DLLs but we'll need someone comfortable with issuing a PR to homebrew as well.

I'm still concerned about the lack of `libdrumstick1` for Ubuntu 14.04 as it breaks our build system.  Any advice from `apt` packagers is appreciated.

tresf:
@tonychee7000 Now I see you're working from two branches.  Please merge the build helpers from `travis-t` into this branch until the mirrors can be setup.

tonychee7000:
@tresf Yep, so I would say all the things I was doing was just a temporarily solution.

Codes in my `travis-t` branch need to be reviewed again and to be squashed when merging

But you may also see my another branch `export_midi_support`, lmms can generate a pretty MIDI file.

I would merge it to the current branch we are at now. after codes of midi-export branch is adapted to your standard.

So could we merge that to a new branch?

>Great. Linux Qt5 won't work unless you shim yet another compile script for 14.04 since it defaults to `libdrumstick0`, which is Qt4 only. When I had previously stated to disable this for platforms which didn't offer binaries, I spoke prematurely since this is a replacement for MIDI import.

Just because I am using Archlinux now, for some library may be newer. But you're right, it seems a tough thing.

PS: travis haven't passed all tests yet, and I need some help for that. see [here](https://travis-ci.org/tonychee7000/lmms/jobs/194696040)

tresf:
I can't get past `pkgconfig` step on my system (14.04, mingw64).

I think anyone helping will need exactly the changes you've made.

```
CMake Error at /usr/share/cmake-2.8/Modules/FindPkgConfig.cmake:283 (message):
  A required package was not found
Call Stack (most recent call first):
  /usr/share/cmake-2.8/Modules/FindPkgConfig.cmake:337 (_pkg_check_modules_internal)
  plugins/SmfImport/CMakeLists.txt:5 (PKG_CHECK_MODULES)
```

tonychee7000:
So what can I do?

tonychee7000:
I am a newbie on Qt and cross compile and need help for this.

tresf:
>  I need some help for that. see here.... I am a newbie on Qt and cross compile and need help for this.

You just need to adjust your Windows compilation script to use separate prefixes for 64 and 32 bit.  I'll comment on the `travis-t` branch.  This would be much easier if we were working from a single branch.

tresf:
> You just need to adjust your Windows compilation script to use separate prefixes for 64 and 32 bit

Nevermind, your script above is different from that on `travis-t`.   Are you on discord?

tonychee7000:
>Nevermind, your script above is different from that on `travis-t`. Are you on discord?

I have just modified them, so you may subject to `travis-t`

tresf:
> I have just modified them, so you may subject to travis-t

The following variables aren't needed in your 64-bit mingw script: MINGW32, PROCESSOR32, MINGW_TOOL_PREFIX32, DMINGW_PREFIX32, DCMAKE_SYSTEM_PROCESSOR32,DWIN64=TRUE

The only differences between 32-bit and 64-bit compilations should be the prefix and arch.  I would recommend combining the build scripts into one.

Why are you [rewriting the `CMakeLists.txt` file](https://github.com/tonychee7000/lmms/blob/travis-t/cmake/nsis/build_mingw64_drumstick.sh#L46)?  And why aren't you doing this for 32-bit?

tresf:
Also, it appears the `-lQCore4` syntax for Qt5 can be `Qt5::Core` per https://github.com/LMMS/lmms/blob/master/plugins/vst_base/CMakeLists.txt#L9-L13.

tonychee7000:
>Why are you [rewriting the CMakeLists.txt file](https://github.com/tonychee7000/lmms/blob/travis-t/cmake/nsis/build_mingw64_drumstick.sh#L46)? And why aren't you doing this for 32-bit?

Actually, 32-bit also rewrite `CMakeLists.txt` for `drumstick/utils` depend on `Qt5Svg`

tresf:
I think QtCore4 can be removed actually.  The `BUILD_PLUGIN` macro seems to handle most of this automatically.  The examples I linked don't use the `BUILD_PLUGIN` macro, yours does.

tonychee7000:
Qt5 on OSX with static link passed.

tresf:
**Mac:**
Does `-DCMAKE_INSTALL_PREFIX=/usr/local` help the Mac builds (Homebrew install prefix?)

**Qt4:**

There seems to be something wrong with how `libdrumstick 0.5.0` is looking for Qt4.  The first run only of configure shows a bunch of warnings about not finding QtCore, and I think that's causing the grief later in the build.  I'm not sure if Travis is suppressing it or not, but this is how it appears on my machine:
https://gist.github.com/tresf/69c097353bdb8e9c6f55b73d44e83246

For now, can you try to add `export PKG_CONFIG_PATH=/opt/mingw$2/lib/pkgconfig` and see if it succeeds?  If so, there may be a bug or misconfiguration with the installation.

tonychee7000:
**Mac:**
It seems the default install path is `/usr/local`

When I use static link(`-DSTATIC_DRUMSTICK=ON` when build drumstick) with Qt5, it will pass. otherwise(use framework, might bug of drumstick) it fails.

**Qt4:**
Still fail :(

tresf:
> When I use static link(-DSTATIC_DRUMSTICK=ON when build drumstick) with Qt5, it will pass. otherwise(use framework, might bug of drumstick) it fails.

That is OK, we can re-link it [at bundling time](https://github.com/LMMS/lmms/blob/master/cmake/apple/install_apple.sh.in#L62-L72). 

> Qt4 Still fail :(

Please move it to `build_mingw32|64.sh`
https://github.com/tonychee7000/lmms/commit/125b55e5bdbf15c6fbd37fb5a798ab51ed1014c2#diff-eb7d01858da9e5272e82592903f4797d

```diff
if [ "$1" = "-qt5" ]; then
+     # pkgconfig bug with libdrumstick<=0.50
+     export PKG_CONFIG_PATH=$MINGW/lib/pkgconfig
      QT5=True
fi
```

tonychee7000:
>Please move it to `build_mingw32|64.sh`

Qt5 is okay now.

But Qt4 on MinGW or OSX are all fail. will this fix work?

tonychee7000:
I have checked `link.txt` and find this
```
usr/bin/cmake -E remove -f CMakeFiles/smfimport.dir/objects.a
/opt/mingw64/bin/x86_64-w64-mingw32-ar cr CMakeFiles/smfimport.dir/objects.a @CMakeFiles/smfimport.dir/objects1.rsp
/opt/mingw64/bin/x86_64-w64-mingw32-g++     -shared -o ../smfimport.dll -Wl,--major-image-version,0,--minor-image-version,0 -Wl,--whole-archive CMakeFiles/smfimport.dir/objects.a -Wl,--no-whole-archive  -L/opt/mingw64  -L/home/tc/lmms/build  -L/home/tc/lmms -Wl,-Bstatic -lQtGui4 -lQtXml4 -lQtCore4 -Wl,-Bdynamic ../../src/liblmms.dll.a -ldrumstick-file -lQtCore -lQtGui4 -lQtCore4 -lkernel32 -luser32 -lgdi32 -lwinspool -lshell32 -lole32 -loleaut32 -luuid -lcomdlg32 -ladvapi32
```

I wonder how to get rid of this `-lQtCore`

jasp00:
Since we are moving to Qt 5 and the plug-in depends on Qt, I would enable this plug-in only on Qt 5. Library source should be uploaded because the library is not available in some distros (experimental packages are not acceptable). I would not drop `MidiImport` (is it broken?) until this new plug-in is more tested. The core should not depend on the specific plug-in, but conditional compilation could be used in the meantime.

tonychee7000:
@jasp00 I have made new `MidiExport` plugin and worked. So should I merge it into this branch?

tresf:
> Since we are moving to Qt 5 and the plug-in depends on Qt, I would enable this plug-in only on Qt 5. 

Perhaps this is the PR which [pulls the plugin on Qt4](https://github.com/LMMS/lmms/issues/2611) for Mac and Windows then.

tonychee7000:
Not bad, but I would say is there necessary for building Qt4 on MinGW and OSX?

tresf:
> is there necessary for building Qt4 on MinGW and OSX?

Probably not.

tonychee7000:
Me too, so I might merge that `travis-t` into this brach. :|

src/gui/MainWindow.cpp
```diff
@@ -308,12 +308,13 @@ void MainWindow::finalize()
 					SLOT( exportProjectTracks() ),
 					Qt::CTRL + Qt::SHIFT + Qt::Key_E );
 
-	// temporarily disabled broken MIDI export				
-	/*project_menu->addAction( embed::getIconPixmap( "midi_file" ),
+#ifdef LMMS_HAVE_FLUIDSYNTH
```
softrabbit: I don't think exporting MIDI should depend on fluidsynth. Import, possibly.

tonychee7000: It's depend on `Sf2 Player`

tresf: I agree with @softrabbit.  The days of MIDI = cheesy instrument bank are in the past but current version warns when no default soundfont is loaded, so perhaps decoupling this should be a separate bug/pr?

tonychee7000: Well, you can change them as you wish. 

In my opinion, LMMS should have ablity to output a pretty MIDI file, for I had thousands of MIDI file to listen.

tonychee7000: @tresf @softrabbit How do you think a new plugin just for MIDI-out like FL Studio?

I have no good idea about that.

tonychee7000:
Just a suggestion, you can create a new branch if necessary.

src/gui/MainWindow.cpp
softrabbit: @tonychee7000, let's not discuss that plugin here. There's issue #793 and possibly others for that.

tonychee7000: I see that.

difarem:
```
Object::connect: No such signal ����::signalSMFTimeSig( int, int, int, int )
Object::connect: No such signal ����::signalSMFTempo( int )
Object::connect: No such signal ����::signalSMFText( int, QString )
Object::connect: No such signal ����::signalSMFError( QString )
Object::connect: No such signal ����::signalSMFCtlChange( int, int, int )
Object::connect: No such signal ����::signalSMFPitchBend( int, int )
Object::connect: No such signal ����::signalSMFNoteOn( int, int, int )
Object::connect: No such signal ����::signalSMFNoteOff( int, int, int )
Object::connect: No such signal ����::signalSMFProgram( int, int )
Object::connect: No such signal ����::signalSMFHeader( int, int, int )
Object::connect: No such signal ����::signalSMFTrackStart()
```

Just tried building the PR, and it looks like it segfaults when trying to import any MIDI file.

cmake/apple/build_osx_drumstick.sh
```diff
@@ -0,0 +1,23 @@
...
+mkdir -p build && cd build
+
+if [ $QT5 ];then
+    STATIC_OPTS=ON
+else
+    STATIC_OPTS=OFF
+fi
+
+cmake ../ -DUSE_WERROR=OFF -DLIB_SUFFIX='' -DSTATIC_DRUMSTICK=$STATIC_OPTS
+make && sudo make install
```
tresf: This file is near identical to Linux.  They should be consolidated.  It can live in `linux` until this library is ported to homebrew.

cmake/nsis/build_mingw_drumstick.sh
```diff
@@ -0,0 +1,54 @@
+drumstick_ver=$1
```
tresf: This can stay as-is, but this should probably be ported to https://github.com/LMMS/lmms/blob/master/cmake/msys/msys_helper.sh like @DragonEagle did in #3369 or alternately added directly to the mingw mirros that Toby maintains for us.  Toby generally does this on-request.

I think first though we can pull Qt4 support, which is an open task, so we wouldn't hold up the merge for this script alone.

tresf:
> it segfaults when trying to import any MIDI file.

> it segfaults when trying to import any MIDI file.

Seems to be similar to this issue... http://stackoverflow.com/a/14774274/3196753

@tonychee7000 may need to add the class name in front of the slots `::signalSMFTimeSig`, etc.

I assume you're on the right build of drumstick? Qt5 builds against `drumstick `1.1.x`.  Qt4 builds against `drumstick 0.5.x`.

tonychee7000:
> @tonychee7000 may need to add the class name in front of the slots `::signalSMFTimeSig`, etc.

I have added that class name above `::signal***`, but it won't work with output

```
QObject::connect: No such signal drumstick::QSmf::drumstick::QSmf::signalSMFTimeSig( int, int, int, int ) in /home/tonychyi/工程/lmms/plugins/SmfImport/midiReader.cpp:49
QObject::connect: No such signal drumstick::QSmf::drumstick::QSmf::signalSMFTempo( int ) in /home/tonychyi/工程/lmms/plugins/SmfImport/midiReader.cpp:52
QObject::connect: No such signal drumstick::QSmf::drumstick::QSmf::signalSMFText( int, QString ) in /home/tonychyi/工程/lmms/plugins/SmfImport/midiReader.cpp:55
QObject::connect: No such signal drumstick::QSmf::drumstick::QSmf::signalSMFError( QString ) in /home/tonychyi/工程/lmms/plugins/SmfImport/midiReader.cpp:58
QObject::connect: No such signal drumstick::QSmf::drumstick::QSmf::signalSMFCtlChange( int, int, int ) in /home/tonychyi/工程/lmms/plugins/SmfImport/midiReader.cpp:61
QObject::connect: No such signal drumstick::QSmf::drumstick::QSmf::signalSMFPitchBend( int, int ) in /home/tonychyi/工程/lmms/plugins/SmfImport/midiReader.cpp:64
QObject::connect: No such signal drumstick::QSmf::drumstick::QSmf::signalSMFNoteOn( int, int, int ) in /home/tonychyi/工程/lmms/plugins/SmfImport/midiReader.cpp:67
QObject::connect: No such signal drumstick::QSmf::drumstick::QSmf::signalSMFNoteOff( int, int, int ) in /home/tonychyi/工程/lmms/plugins/SmfImport/midiReader.cpp:70
QObject::connect: No such signal drumstick::QSmf::drumstick::QSmf::signalSMFProgram( int, int ) in /home/tonychyi/工程/lmms/plugins/SmfImport/midiReader.cpp:73
QObject::connect: No such signal drumstick::QSmf::drumstick::QSmf::signalSMFHeader( int, int, int ) in /home/tonychyi/工程/lmms/plugins/SmfImport/midiReader.cpp:76
QObject::connect: No such signal drumstick::QSmf::drumstick::QSmf::signalSMFTrackStart() in /home/tonychyi/工程/lmms/plugins/SmfImport/midiReader.cpp:79
```

Here are duplicate `drumstick::QSmf`

difarem:
>I assume you're on the right build of drumstick? Qt5 builds against drumstick 1.1.x. Qt4 builds against drumstick 0.5.x.

Yep, building with Qt5 seems to have solved the issue.

On another note, perhaps the instrument used when exporting a MIDI could be determined with a dropdown on the instrument window, instead of forcing all instruments to be a sf2 player.

PhysSong:
Now there are a bunch of merge conflicts. If this work still has some values, someone may redo this work.
@tonychee7000 Do you have any ideas/opinions what should we do?

tonychee7000:
Reply @PhysSong 

Oops, several years.

Well,  someone can redo this feature referring to my code or just check where conflict is and merge them by manual. Might a lot work to do. 

In my code, `MidiImport` was replaced by `SmfImport` for drumstick supports Overture(`.ove`) and old Cakewalk(`.wrk`) project file. I just want a accurate name.

`MidiExport` was totally ported to drumstick. Abandon the current code might do.

A new branch might needed?

PhysSong:
Well, there have been several changes:
- `MidiExport` changed a lot in #3733, mainly due to #2384
- `master` now requires Qt5, not Qt4
- We have experimental MSVC support

There are some more changes. Anyway, it may be very difficult to resolve merge conflicts...

tonychee7000:
@PhysSong 

* `MidiExport` I haven't considered about B&B tracks or more, just export tracks which use `sf2player`. It might be redo.
* I tested Qt5 and all codes was based by Qt5, just forget Qt4.
* I have no idea about MSVC, even it worked on Travis at that time.

I will show what I have done on `MidiExport`, all based on drumstick.

1. Call `setFileFormat()` and set MIDI file version to `1`
2. Call `setDivision(960)` 
3. Scan all tracks which use `sf2player` and call `setTracks() + 1` on drumstick
4. `AutomationTrack` was bonded to MIDI channel.
5. Write drum track with MIDI channel 10.
6. Write global event which had track 0
7. List all notes and automations by order
8. Calculate the time difference between 2 events.
9. Write to some channel
10. Write `EndOfTrack` event finally

Considering `MidiExport` on `master` has changed a lot, Redo might be an option.

PhysSong:
> Considering `MidiExport` on `master` has changed a lot, Redo might be an option.

Some remarks in case someone wants to redo works in this PR:
- The license of drumstick has been changed to GPL3+ in [r464](https://sourceforge.net/p/drumstick/code/464/).
- Overture support is removed in [r437](https://sourceforge.net/p/drumstick/code/437/).

Note that those changes are in their trunk, but not in their latest release([r391](https://sourceforge.net/p/drumstick/code/391/)).

PhysSong:
Updates:
- The latest Drumstick release is 2.7.2(as of 2022.11.12), which defaults to Qt 6(Qt 5 option is also available, but some distributions build it with Qt 6)
- As of Drumstick 2.0, it's licensed under GPL3+ while LMMS is still under GPL2+

zonkmachine:
> As of Drumstick 2.0, it's licensed under GPL3+ while LMMS is still under GPL2+

What does this mean practically?

tresf:
> What does this mean practically?

It just means we need to bump LMMS to GPL3+ in order to bundle it.  Any dependencies that don't have the "+" in "GPL2+" would no longer be able to be bundled.  I think we went over this in another thread, but my seaching was not fruitful. https://github.com/LMMS/lmms/issues?q=is%3Aissue+is%3Aopen+gpl3

zonkmachine:
> It just means we need to bump LMMS to GPL3+ in order to bundle it. Any dependencies that don't have the "+" in "GPL2+" would no longer be able to be bundled. I think we went over this very specifically in another thread, but my seaching was not fruitful. https://github.com/LMMS/lmms/issues?q=is%3Aissue+gpl3

OK. So, if I understand this correctly, big if..., we wouldn't be able to ship it with the AppImage but for the rest we could link it to a package.

tresf:
> > It just means we need to bump LMMS to GPL3+ in order to bundle it. Any dependencies that don't have the "+" in "GPL2+" would no longer be able to be bundled. I think we went over this very specifically in another thread, but my seaching was not fruitful. https://github.com/LMMS/lmms/issues?q=is%3Aissue+gpl3
> 
> OK. So, if I understand this correctly, big if..., we wouldn't be able to ship it with the AppImage but for the rest we could link it to a package.

When you way "wouldn't be able to ship", I'm not sure if you mean the GPL2 (without +) or GPL3 (before we upgrade), but I'll try to make it as simple as possible, based on my understanding of the license.
* GPL2+ can upgrade to GPL3+ at any time. without a license change.  You can just call the completed work GPL3+, however many will consider this misleading.
* GPL3+ cannot downgrade to GPL2+. 
* GPL2 (without the +) cannot upgrade to GPL3+.
* Regardless of this, our plugin system can fall into the "Standard Interface" clause of either GPL license.
   * This means that technically, as long as our plugin-specific code is compatible, we're OK.
   * This may cause confusion or be considered misleading if certain directories of the project were a different license, but this is technically allowed using an interface or plugin system.
   * At time of writing this, I do not believe we use a standard interface for the MIDI import, so this library specifically would require either a refactor to leverage the "Standard Interface" clause, or an upgrade to the project as a whole to comply with GPL3+.
*  Switching the project to GPL3+ is already a requirement for VST3 support, so this is probably the best path.  We just need to make sure we check each plugin and dependency for compatibility, and if exceptions exist, try to leverage the "Standard Interface" clause.  Anything not using a plugin system (e.g. uses headers and linking) is a hard-requirement to be GPL3/GPL3+ compatible or needs a significant refactor to be used as a standardized plugin.

With regards to the word "shipping"... the bundle of GPL3+ and non-GPL3+ components is referred to as "aggregate" in the license and seems to be allowed, although I've seen projects such as FoxIt leverage install-time downloaders (e.g. GhostScript) to avoid any potential legal issues.

zonkmachine:
> With regards to the word "shipping".

I think I meant "bundle".

Thanks for the clarification. I'm removing this from the current milestone.

Rossmaxx:
@Veratil didn't you rewrite midi import (and also plan to fix midi export) so is there anything you need here?

Rossmaxx:
> Well, someone can redo this feature referring to my code or just check where conflict is and merge them by manual. Might a lot work to do.
In my code, MidiImport was replaced by SmfImport for drumstick supports Overture(.ove) and old Cakewalk(.wrk) project file. I just want a accurate name.
MidiExport was totally ported to drumstick. Abandon the current code might do.
A new branch might needed?

I see this as an indication for closing.

Veratil:
> @Veratil didn't you rewrite midi import (and also plan to fix midi export) so is there anything you need here?

Yes, MIDI import is pretty much done locally (but I don't like the code in the current state it is). Export is not done yet.

PR #3229, opened 2017-01-09
Fixed #3226. Beat velocity fixes for both themes.

### LINKED ISSUE #3226: Beat velocities are not shown correctly on the classic theme
image
Problem introduced by #2960. The graphics for the classic theme should be updated.

Noticed by @karmux in https://github.com/LMMS/lmms/issues/3225#issuecomment-271285053

karmux:
Fixed beat velocity on classic theme.
Fixed visual difference between 0 velocity and no step on both themes.


Umcaruje:
Will test this out soon.

Umcaruje:
I would suggest rendering the `step_btn_on_0.png` for every note, and rendering the 200 graphic over that. It will mimic the previous behavior, and the jump from 0% to 1% won't look so sudden.

src/tracks/Pattern.cpp
```diff
@@ -1037,16 +1037,13 @@ void PatternView::paintEvent( QPaintEvent * )
 			{
 				const int vol = n->getVolume();
 				p.drawPixmap( x, y, stepoffl );
+				p.drawPixmap( x, y, stepon0 );
 				if( vol > 0 )
```
Umcaruje: And this if statement is not really needed, I mean when volume is 0 the opacity will be also 0, effectively rendering nothing.

Umcaruje: The only reason the statement might stay is CPU usage, but that is so minor that I don't think it should affect this.

Umcaruje:
This looks good to me.
image

Tagging @RebeccaDeField to see if she approves the look of the 0 velocity beats. :+1: 

Checked out the classic theme too, looking good.

RebeccaDeField:
@Umcaruje Thanks for the heads up. The design the 0 velocity beats in your screenshot work well with rest of the theme to me :+1:

BaraMGB:
So this can be merged?

tresf:
> So this can be merged?

Testing and feedback suggests this yes.  Note, this change will break any 1.2-RC1 and 1.2-RC2 3rd party themes (which is fine, but something the devs/community should be aware of)

zonkmachine:
OK, merging based on feedback. I tried to test this PR but the branch was deleted. :cry:

PR #3231, opened 2017-01-10
Add NSHighResolutionCapable to lmms.plist.in

wileyyugioh:
This will enable lmms in OSX to stop opening in low resolution mode and instead take advantage of retina display. It makes the title bar and text look clearer, but icons and images are unaffected and will still remain blurry.

Before
before

After
after

### DIFF
--- a/cmake/apple/lmms.plist.in
+++ b/cmake/apple/lmms.plist.in
@@ -143,5 +143,7 @@
                 </dict>
             </dict>
         </array>
+        <key>NSHighResolutionCapable</key>
+        <string>True</string>
   </dict>
 </plist>


tresf:
@wileyyugioh thanks.  The screenshots you've provided don't do this justice.

| Before | After |
|--------|-------|
|image | image |


Merging as I see no downside in having this.  Thanks!

Umcaruje:
Would blurry icons be fixed by using .svgs, or we need to make double resolution PNGs?

tresf:
> Would blurry icons be fixed by using .svgs

Yes.

PR #3238, opened 2017-01-11
SoundIO- fixed up crashes when device cannot be opened.

gnudles:
LMMS crashes when soundio cannot open device. it tries to free pointer with unset value.


src/core/audio/AudioSoundIo.cpp
```diff
@@ -195,7 +196,8 @@ void AudioSoundIo::onBackendDisconnect(int err)
 AudioSoundIo::~AudioSoundIo()
 {
 	stopProcessing();
-	soundio_destroy(m_soundio);
+	if (m_soundio)
+		soundio_destroy(m_soundio);
```
AndiEcker: m_soundio = NULL;

tresf:
@gnudles thanks for the patch.

gnudles:
:+1:

Issue #3235, opened 2017-01-11
CONTROLLER RACK - Strange Window Width


miketurn:
Hello Everyone,

Obviously not a huge deal, I actually like when windows are more stretchy, but seems odd, that the stretching capabilities of "Controller Rack" window is overly customizable, meaning is does not have a fixed width.

I will admit I have not really looked into using this feature, so I may be unaware of something, or of a possible future plan for this window, which if that is the case I apologize.

Are there any future plans for adding other items to the "Controller Rack" other then the "LFO Controller"?

Thank You

PR #3240, opened 2017-01-12
makes controller rack fixed in width

BaraMGB:
fixes https://github.com/LMMS/lmms/issues/3235


zonkmachine:
@BaraMGB It's possible to make the Controller window to disappear by grabbing the bottom frame and moving up. It should have a minimum height.

BaraMGB:
I will look into this.

zonkmachine:
The same thing goes with the **about** window and the **export** window ( second window with audio file settings ).

PR #3254, opened 2017-01-13
corrects precision for sample track volume knob

BaraMGB:
the volume knob on the sample track gots now the same precision as the instrument volume knob: one position behind the decimal point.


BaraMGB:
This is a simple one. merge

Issue #3234, opened 2017-01-11
BROWSER - Add a Close "X"


miketurn:
Hello Everyone,

I think a close "X" added in the upper right hand corner of the browser would be useful and would follow the theme of all other windows.
In ADDITION to the close "X" it may also be nice to program the ability to close the browser by double clicking on the "6 Dot" dragging grip located in the middle of the right side of the browser.

Not big items but may be nice in the future if others agree.
Thank You

tresf:
Like this?

![image](https://cloud.githubusercontent.com/assets/6345473/21831381/52d30ca6-d771-11e6-89d7-57b77bb21c9e.png)

miketurn:
@tresf 
Correct

Maybe even just giving the browser a title bar like all the other windows.
Where it has an     icon at the left   /   title   /   X (to close)
Tough to say because I am aware that it is a special kind of sidebar, but it might make it fit in a little more, if desired.


While I have you here for a second, sometimes it is hard to tell if I should group my suggestions on a specific topic, such as say "The Browser" or to separate them.  I like them separate, but sometimes they are very minor suggestions I feel weird making single posts for.  I feel they would probably get more attention if separate.
(Sorry for asking this here, I can remove this part of the post after, once I receive an answer)

Like for example,  I would recommend maybe removing those instructions underneathe the red "X" in the picture you provided above.
"Drag an instrument..........existing instrument track."

Thank You For your quick response and for the graphic provided.

Oh by the way, the "p" in "Instrument plugins" needs capitalizing :)

tresf:
> I would recommend maybe removing those instructions underneathe the red "X" in the picture you provided above.

I think everyone is in agreement, but we need to make it a bit more UX consistent with the file browser (Right Click context menu, double-click) then the help text can be removed.

> Oh by the way, the "p" in "Instrument plugins" needs capitalizing :)

Please make the [change yourself](https://github.com/LMMS/lmms/blob/master/src/gui/PluginBrowser.cpp) and we'll merge.  Please tag @liushuyu in your commit so he knows to update transiflex translation files.

miketurn:
> Please make the change yourself and we'll merge. Please tag @liushuyu in your commit so he knows to update transiflex translation files.
I didn't realize I would be allowed to do that, I can see where the text needs to be updated on the link you provided, I would change it but I don't know the process of how to actually do it.
Believe me, if I did I would "P" all over that code :)

tresf:
[...]

PR #3256, opened 2017-01-13
SideBarButton set text below the icon

BaraMGB:
this would fix https://github.com/LMMS/lmms/issues/3234#issuecomment-271785914
2017-01-13-104015_1920x1080_scrot

### DIFF
--- a/data/themes/default/style.css
+++ b/data/themes/default/style.css
@@ -494,7 +494,7 @@ SideBar QToolButton:pressed {
 }
 
 SideBar QToolButton:checked {
-        background: qlineargradient(spread:reflect, x1:0, y1:0, x2:0, y2:1, stop:0 #4b525b,  stop:1 #2f333a);
+	background: qlineargradient(spread:reflect, x1:0, y1:0, x2:0, y2:1, stop:0 #4b525b, stop:1 #2f333a);
 	border: none;
 	font-size: 12px;
 }
--- a/data/themes/default/style.css
+++ b/data/themes/default/style.css
@@ -494,7 +494,7 @@ SideBar QToolButton:pressed {
 }
 
 SideBar QToolButton:checked {
-	background: none;
+	background: #2f353b;
 	border: none;
 	font-size: 12px;
 }
--- a/data/themes/default/style.css
+++ b/data/themes/default/style.css
@@ -494,7 +494,7 @@ SideBar QToolButton:pressed {
 }
 
 SideBar QToolButton:checked {
-	background: #2f353b;
+        background: qlineargradient(spread:reflect, x1:0, y1:0, x2:0, y2:1, stop:0 #4b525b,  stop:1 #2f333a);
 	border: none;
 	font-size: 12px;
 }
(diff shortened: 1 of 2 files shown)


mikobuntu:
:+1:  great job :)

tresf:
Unrelated to this PR, but the side-tabs needs more contrast from each other.  The description is indistinguishable from the two icons.

mikobuntu:
@tresf something like this design in my mockup could work?
sidebar

tresf:
> @tresf something like this design in my mockup could work?

👍

BaraMGB:
> something like this design in my mockup could work?

I guess this one can easily made in the styles.css

mikobuntu:
@BaraMGB Yes i was thinking the same. I could probably dig out my css book and have a look to see how to do this, well along with qtcreator :) , tho I hope the experts step in and do this i.e @RebeccaDeField et all :)

BaraMGB:
in the styles.css line 497 changed to:
`	background: #2f353b;`
looks like:
2017-01-13-201705_1280x800_scrot

Should I add this to this PR?

mikobuntu:
@BaraMGB did adding the background conflict with the original PR, as i see the icon is now below the text in your last image?

BaraMGB:
I changed the styles.css in master for be quick. ;)

BaraMGB:
2017-01-13-214328_1280x800_scrot

looks not bad.

tresf:
Better but still requires more contrast.  The Instrument Buttons are about as close as we should allow (despite the similarities in color, they have additional bevel and shading which makes them standout from the background).


image

RebeccaDeField:
This is looking good, people :) I do agree with @tresf that if we are adding this to enhance the contrast, we might want to step up the contrast a bit more.

RebeccaDeField:
Here are some colors that should work well for the sidebar tab.

**Horizontal gradient:**
Stop 1: #4b525b 
Stop 2: #2f333a

Should look something like this


@BaraMGB

BaraMGB:
Thanks to @RebeccaDeField :+1:  

2017-01-23-131900_1920x1080_scrot

data/themes/default/style.css
```diff
@@ -494,7 +494,7 @@ SideBar QToolButton:pressed {
 }
 
 SideBar QToolButton:checked {
-	background: none;
+        background: qlineargradient(spread:reflect, x1:0, y1:0, x2:0, y2:1, stop:0 #4b525b,  stop:1 #2f333a);
```
tresf: @BaraMGB code style formatting... You've used spaces instead of a tab.

src/gui/widgets/SideBar.cpp
```diff
@@ -110,8 +110,9 @@ SideBar::~SideBar()
 void SideBar::appendTab( SideBarWidget *widget )
 {
 	SideBarButton *button = new SideBarButton( orientation(), this );
-	button->setText( widget->title() );
+	button->setText( " " + widget->title() );
```
tresf: Can you explain this?  Is this for deference purposes?

tresf: Never mind, it's a space, I thought it was empty.  Please disregard.

tresf:
Just one minor code formatting change, then we can merge this.

BaraMGB:
done ;)

Issue #3182, opened 2016-12-27
Shouldn't saving preset save FX channel as 0?


karmux:
When you save some instrument preset then also FX channel is saved.
If you listen previously saved preset in another project it plays either silence (when FX channel does not exist) or some random effects that happen to be on that channel.
If channel were saved as 0 then it plays user created presets consistently with factory presets.
Currently one can solve this by changing channel to 0 before saving preset.

tresf:
> Shouldn't saving preset [files] [default the] FX channel [to] 0

👍

softrabbit:
I disagree. I think this should be handled through ignoring any saved FX channel on load. That way you won't have to reset the channel when you load a preset into a track you already assigned to a channel.

Umcaruje:
I agree with @softrabbit we shouldn't save FX channel information in presets at all.

tresf:
Well @softrabbit is saying to go ahead and save it, just ignore it on load.

Umcaruje:
Well, maybe the best of both worlds? Save with FX0, to avoid problems on preview and ignore on load.

tresf:
@Umcaruje I'd be more interested to see if we can abolish the FX channel keys from the XML without causing a crash.  Might hit two birds, so to speak.

tresf:
```diff
- <instrumenttrack pan="30.0333" fxch="1" usemasterpitch="1" pitchrange="2" pitch="-90.9667" basenote="46" vol="48.1">
                                 ^-------
+ <instrumenttrack pan="30.0333" usemasterpitch="1" pitchrange="2" pitch="-90.9667" basenote="46" vol="48.1">
                                 ^------- (REMOVED)
```

Tested, works. 😄 

![screen shot 2016-12-30 at 5 37 15 pm](https://cloud.githubusercontent.com/assets/6345473/21573784/a1df5dc2-ceb6-11e6-9f00-6fd44b2541dc.png)

tresf:
Can be shimmed in the `saveSettings` function of the `InstrumentTrack` class:

https://github.com/LMMS/lmms/blob/master/src/tracks/InstrumentTrack.cpp#L1561:L1594

tresf:
@softrabbit I tested this... it seems to also preserve the existing FX channel selection when dragged over an existing instrument button.

softrabbit:
Ideally the fix should:
- preserve channel selection when loaded into an existing track 
- work with old presets (just not saving the channel doesn't)
- not break preview
[...]

Issue #3021, opened 2016-09-05
24-bit export


unfa:
Sometimes 32 bit-export is an overkill, but I still needs more than 16 bits.

This could save a lot of diskspace, you know :)

zonkmachine:
I had a quick look at this and the library we use to write sound files, [libsndfile](http://www.mega-nerd.com/libsndfile/#Features), has support for this so it should be a pretty easy fix.

WinLin97:
Pretty easy? So... When will this be fixed?

Umcaruje:
There's nobody currently working on this issue. We're short on devs. Do you
volunteer?

michaelgregorius:
> Pretty easy? So... When will this be fixed?

Now. :) Implemented with pull request #3516.

michaelgregorius:
The pull request now also adds some other improvements to the export dialog:
* Enable variable bit rate for Ogg exports.
* Only show relevant widgets depending on the selected export format.

See the comments in #3516 for more details and screenshots.

WinLin97:
Great, big thanks Michael! :D

PR #3259, opened 2017-01-14
Xpressive plugin

gnudles:
Xpressive plugin allows you to write your own expressions for sound creation.


AndiEcker:
+1 - WOW - looks like another interesting instrument

2017-01-14 19:53 GMT+00:00 gnudles <notifications@github.com>:

zonkmachine:
xpressive

:open_mouth: Totally tripping right now. It's so cool. Like if Bit Invader grew up and took classes in engineering.

gnudles:
What should be done with this error? :
/tmp/cc62tQLz.s: Fatal error: can't write CMakeFiles/xpressive.dir/exprfront.cpp.obj: File too big

ArashPartow:
> What should be done with this error? :

The version of ld shipped with mingw64 is **Literally Broken** - last good version was circa GCC4.3.  In short to get rid of the build error you need to do the following two things:

**Step 1.** 
Add the following only to the mingw64 build options in CMake:

**`-Wa,-mbig-obj`**


**Step 2.** 
In order to make the ExprTk TU section size smaller add the following defines either in the mingw64 CMake build line 
 
- -Dexprtk_disable_break_continue
- -Dexprtk_disable_sc_andor
- -Dexprtk_disable_enhanced_features
- -Dexprtk_disable_string_capabilities
- -Dexprtk_disable_rtl_io_file

Or **#ifdef** on mingw64 just prior to the **#include "exprtk.hpp"** and define 

```
#define exprtk_disable_break_continue
#define exprtk_disable_sc_andor
#define exprtk_disable_enhanced_features
#define exprtk_disable_string_capabilities
#define exprtk_disable_rtl_io_file
```

gnudles:
oops, I'll replace
`IF(MINGW_PREFIX)`
with
`IF(WIN64)`
I invoke the parser with:
```
	parser_t::settings_store sstore;
	sstore.disable_all_logic_ops();
	sstore.disable_all_assignment_ops();
	sstore.disable_all_control_structures();
	parser_t parser(sstore);
```
So I don't understand why these defines should be used only in a certain build. Furthermore, I added them to reduce build time.

tresf:
@ArashPartow Thanks for such specific insight into this issue.  From what I'm reading, we can set **Step 1** globally, right?

https://github.com/LMMS/lmms/blob/master/cmake/modules/MinGWCrossCompile.cmake#L17-L22

, and

https://github.com/LMMS/lmms/blob/master/cmake/modules/MinGWCrossCompile.cmake#L61-L67

ArashPartow:
I think it should be something like:

`IF(MINGW_PREFIX and WIN64)`

As **WIN64** probably implies **ALL** windows 64 builds, perhaps including MSVC


-------------------


Perhaps online _assignment_ and _loop control_structures_ as you might want if-statements/ternary-statements with logical operators to be made available:

```
parser_t::settings_store sstore;
sstore.disable_all_assignment_ops();
sstore.disable_control_structure(settings_t::e_ctrl_for_loop   );
sstore.disable_control_structure(settings_t::e_ctrl_while_loop );
sstore.disable_control_structure(settings_t::e_ctrl_repeat_loop);
parser_t parser(sstore);
```


-------------------


> So I don't understand why these defines should be used only in a certain build. 

It's explained in my post **above**.


-------------------


> I do want all the users to be able to use
> (t>0)&(t<1)
> I saw that using the flag exprtk_disable_sc_andor changes the implementation of the "&" and "|" signs.

Yeah get rid of it from the build line. It wont save much for section size issues and it is probably better off to have it available/present.

ArashPartow:
@tresf  Nothing should be global. All of this is related to a broken version of  **mingw64** - So all the changes must only be specific to 64-bit builds using mingw64 and no other compilers or architectures. 

The build changes presented are a hack/kludge to reduce the section size of the TU so as to allow the broken **ld** that is shipped with **mingw64**  to build the final executable binary. 

The **_defines_** disable certain features such as expression optimizations so as to reduce the section size.

tresf:
> @tresf Nothing should be global. All of this is related to a broken version of mingw64 - So all the changes must only be specific to 64-bit builds using mingw64 and no other compilers or architectures.

Please look at the lines that are linked.  https://github.com/LMMS/lmms/pull/3259#issuecomment-273540427

ArashPartow:
ooh yeah I missed this line:

https://github.com/LMMS/lmms/blob/master/cmake/modules/MinGWCrossCompile.cmake#L10

Should be all good if it is only within that  **IF(LMMS_BUILD_MSYS)**  scope.

tresf:
> Should be all good if it is only within that IF(LMMS_BUILD_MSYS) scope.

Great but to be precise, I believe it needs to be for both scopes linked or moved into a separate conditional.

ArashPartow:
@tresf Now I'm confused. According to the build artifacts the 32-bit mingw build was successful however the 64-bit mingw build **failed**.

On line:

https://github.com/LMMS/lmms/blob/master/cmake/modules/MinGWCrossCompile.cmake#L18

The "IF(WIN64)" seems to suggest 64-bit but the comments around and in the scope and the "COMPILER32" stuff seem to suggest 32-bit. 

I'll let you guys figure it out. However you end up doing it my recommendation is that it should only apply to 64-bit builds using mingw64.

tresf:
Correct, we explicitly build the 32-bit host wrapper on 64-bit.  Naturally, the description would have to change if this were added.

softrabbit:
Actually, this shouldn't be (just) a plugin. @gnudles, could you get the expression evalution worked into the standard oscillator class(es)? That way it could be used in at least TripleOscillator or Organic as well, maybe some other plugin too.

(I haven't tried this plugin yet, but I like the idea.)

gnudles:
@softrabbit , I think that the next step is to create a sound scripting language, compile it with llvm (or some other jit engine) on runtime, and make a generic plug-in infrastructure... But then you need to add llvm to the dependency list...
@tresf what's the bottom line? 
@zonkmachine - it's better to use `integrate(f)` instead of `t*f` because it stays stable with frequency change (with base detuning). integrate(f) holds internal counter that counts `f` and return the sum divided by sample_rate.

plugins/xpressive/exprfront.cpp
```diff
@@ -0,0 +1,483 @@
...
+		{
+			if (index<0)
+				return 0;
+			std::mt19937 generator (m_rseed);
+			generator.discard((int)index);
+			return dist<T>(generator);
+		}
+		int m_rseed;
+	};
+}*/
```
zapashcanon: A comment to explain why this is commented would be better I guess.

plugins/xpressive/exprfront.cpp
```diff
@@ -0,0 +1,483 @@
...
+	{exprtk::disable_has_side_effects(*this);}
+
+	inline float operator()(const float& index)
+	{
+		if (index<0 || std::isnan(index) || std::isinf(index))
+			return 0;
+		const int data_size=sizeof(random_data)/sizeof(int);
+		int xi=(int)index;
+		int si=m_rseed%data_size;
+		int sa=m_rseed/data_size;
```
zapashcanon:
```c++
const int xi = [...];
const int si = [...];
const int sa = [...];
```

plugins/xpressive/exprfront.cpp
```diff
@@ -0,0 +1,483 @@
...
+		int ix=(int)x;
+		//to disable interpolation uncomment:
+		//return static_cast<WaveSample*>(w)->samples[ix];
+
+		float xfrc=fraction(x);
+		if (xfrc==0)
+		{
+			return vec[ix];
+		}
+		int ixnext=(ix+1)%size;
```
zapashcanon:
```c++
const int ixnext=(ix+1)%size;
```

plugins/xpressive/exprfront.cpp
```diff
@@ -0,0 +1,483 @@
...
+	inline T operator()(const T& index)
+	{
+		T x=positiveFraction(index);
+		x=size*x;
+
+		int ix=(int)x;
+		//to disable interpolation uncomment:
+		//return static_cast<WaveSample*>(w)->samples[ix];
+
+		float xfrc=fraction(x);
```
zapashcanon:
```c++
const float xfrc=fraction(x);
```

plugins/xpressive/exprfront.cpp
```diff
@@ -0,0 +1,483 @@
...
+	vec(v),
+	size(s)
+	{}
+
+	inline T operator()(const T& index)
+	{
+		T x=positiveFraction(index);
+		x=size*x;
+
+		int ix=(int)x;
```
zapashcanon:
```c++
const int ix=(int)x;
```

plugins/xpressive/exprfront.cpp
```diff
@@ -0,0 +1,483 @@
...
+	WaveValueFunction(const T* v, std::size_t s)
+	: exprtk::ifunction<T>(1),
+	vec(v),
+	size(s)
+	{}
+
+	inline T operator()(const T& index)
+	{
+		T x=positiveFraction(index);
+		x=size*x;
```
zapashcanon:
```c++
const T x = size * positiveFraction(index);
```

plugins/xpressive/expressive_plugin.cpp
```diff
@@ -0,0 +1,884 @@
...
+	}
+}
+
+void expressive::smooth(float smoothness,const graphModel * in,graphModel * out)
+{
+	out->setSamples(in->samples());
+	if (smoothness>0)
+	{
+		int guass_size=smoothness*5;
+		guass_size|=1;
```
zapashcanon:
```c++
const int guass_size = (smoothness * 5) | 1;
```

plugins/xpressive/expressive_plugin.cpp
```diff
@@ -0,0 +1,884 @@
...
+		int guass_center=guass_size/2;
```
zapashcanon:
```c++
const int guass_center=guass_size/2;
```

plugins/xpressive/expressive_plugin.cpp
```diff
@@ -0,0 +1,884 @@
...
+		float delta=smoothness;
```
zapashcanon:
```c++
const float delta=smoothness;
```

plugins/xpressive/expressive_plugin.cpp
```diff
@@ -0,0 +1,884 @@
...
+		float a=1/(sqrtf(2*F_PI)*delta);
```
zapashcanon:
```c++
const float a=1/(sqrtf(2*F_PI)*delta);
```

plugins/xpressive/expressive_plugin.cpp
```diff
@@ -0,0 +1,884 @@
...
+			expr.add_cyclic_vector("W2",e->m_graphW2.samples(),e->m_graphW2.length());
+			expr.add_cyclic_vector("W3",e->m_graphW3.samples(),e->m_graphW3.length());
+		}
+		expr.setIntegrate(&i,sample_rate);
+
+		bool parse_ok=expr.compile();
+
+		if (parse_ok) {
+			e->m_exprValid.setValue(0);
+			int length = m_raw_graph->length();
```
zapashcanon:
```c++
const int length = m_raw_graph->length();
```

plugins/xpressive/expressive_plugin.cpp
```diff
@@ -0,0 +1,884 @@
...
+			expr.add_constant("A1", e->m_parameterA1.value());
+			expr.add_constant("A2", e->m_parameterA2.value());
+			expr.add_constant("A3", e->m_parameterA3.value());
+			expr.add_cyclic_vector("W1",e->m_graphW1.samples(),e->m_graphW1.length());
+			expr.add_cyclic_vector("W2",e->m_graphW2.samples(),e->m_graphW2.length());
+			expr.add_cyclic_vector("W3",e->m_graphW3.samples(),e->m_graphW3.length());
+		}
+		expr.setIntegrate(&i,sample_rate);
+
+		bool parse_ok=expr.compile();
```
zapashcanon:
```c++
const bool parse_ok=expr.compile();
```

plugins/xpressive/expressive_plugin.cpp
```diff
@@ -0,0 +1,884 @@
...
+		e->m_outputExpression[1] = text;
+		break;
+	}
+
+	if (text.size()>0)
+	{
+		ExprFront expr(text.constData());
+		float t,f=10,key=2,rel=0,v=0.5;
+		unsigned int i;
+		unsigned int sample_rate=m_raw_graph->length();
```
zapashcanon:
```c++
float t;
const float f=10,key=2,rel=0,v=0.5;
unsigned int i;
const unsigned int sample_rate=m_raw_graph->length();
```

plugins/xpressive/expressive_plugin.cpp
```diff
@@ -0,0 +1,884 @@
...
+		expr.add_variable("t", t);
```
zapashcanon: `t` not initialized ?

plugins/xpressive/expressive_plugin.cpp
```diff
@@ -0,0 +1,884 @@
...
+	updateLayout();
+}
+
+expressiveView::~expressiveView()
+{
+}
+
+static void clearGraph(graphModel * g)
+{
+	int length = g->length();
```
zapashcanon:
```c++
const int length = g->length();
```

plugins/xpressive/expressive_plugin.cpp
```diff
@@ -0,0 +1,884 @@
...
+	m_smoothW2.loadSettings(_this,"smoothW2");
+	m_smoothW3.loadSettings(_this,"smoothW3");
+	m_parameterA1.loadSettings(_this,"A1");
+	m_parameterA2.loadSettings(_this,"A2");
+	m_parameterA3.loadSettings(_this,"A3");
+	m_panning1.loadSettings(_this,"PAN1");
+	m_panning2.loadSettings(_this,"PAN2");
+	m_relTransition.loadSettings(_this,"RELTRANS");
+
+	int size = 0;
```
zapashcanon: Not sure for this one, as I don't know the effect of the next function calls, but:

```c++
const int size = 0;
```

plugins/xpressive/expressive_plugin.cpp
```diff
@@ -0,0 +1,884 @@
...
+		delete exprO2;
+}
+float saw_wave(float);
+float sin_wave(float);
+void exprSynth::renderOutput(fpp_t _frames, sampleFrame *_buf)
+{
+
+	float o1,o2,pn1,pn2;
+	float new_freq=nph->frequency();
+	float freq_inc=(new_freq-frequency)/_frames;
```
zapashcanon:
```c++
const float new_freq=nph->frequency();
const float freq_inc=(new_freq-frequency)/_frames;
```

plugins/xpressive/expressive_plugin.cpp
gnudles: It works, but I'll initialize it for clarity.

gnudles: I mean, the value is set before it's evaluated...

plugins/xpressive/expressive_plugin.cpp
gnudles: Let's keep it that way...

tresf:
Is this ready for merge?

zonkmachine:
> Is this ready for merge?

I think it is.

I've seen a very minor issue. The presets are not saved with unique names. It's not obvious when you save a preset that the file name and the preset name isn't necessarily the same. One could argue that this is something we could change actually.

data/presets/X-Pressive/X-Pressive1.xpf
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0"?>
+<!DOCTYPE lmms-project>
+<lmms-project version="1.0" creator="LMMS" creatorversion="1.2.0-rc2.6" type="instrumenttracksettings">
+  <head/>
+  <instrumenttracksettings muted="0" type="0" name="X-Pressive" solo="0">
```
zonkmachine: `name="X-Pressive"`

The preset name that shows up when you load an instrument isn't the name of the `.xpf` but this setting. You need to edit the name in the rename dialogue on the top of the plugin.

gnudles: Tell you the truth, I'm not sure that those presets should be merge. And even if they do, I think they should have more meaningful naming... I'm not so good at naming sounds so... if someone volunteers... 
Also if someone made nice presets which he wants them to be merged...

zonkmachine: The user will want to have some good presets to look at for this one. In any case I don't think this should hold back a merge. You or someone else can tweak the names later and it's easier for people to try it out and make presets if it's merged anyway.

plugins/xpressive/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+INCLUDE(BuildPlugin)
+
+SET(CMAKE_CXX_FLAGS "-Dexprtk_disable_sc_andor -Dexprtk_disable_enhanced_features -Dexprtk_disable_break_continue -Dexprtk_disable_comments -Dexprtk_disable_string_capabilities -Dexprtk_disable_rtl_io_file -Dexprtk_disable_rtl_vecops ${WERROR_FLAGS}")
```
jasp00: Please include previous value of `CMAKE_CXX_FLAGS`. You should apply all these definitions only for the MinGW-w64 build, as recommended at https://github.com/LMMS/lmms/pull/3259#issuecomment-273447592.

plugins/xpressive/expressive_plugin.cpp
```diff
@@ -0,0 +1,926 @@
+/*
+ * expressive_plugin.cpp - instrument which uses a mathematical formula parser
+ *
+ * Copyright (c) 2016-2017 Orr Dvori
```
jasp00: The convention is to add an email address, but it is not strictly necessary.

plugins/xpressive/expressive_plugin.cpp
```diff
@@ -0,0 +1,926 @@
...
+ * 
+ * This file is part of LMMS - http://lmms.io
```
jasp00: https.

plugins/xpressive/expressive_plugin.cpp
```diff
@@ -0,0 +1,926 @@
...
+ */
+
+#include "expressive_plugin.h"
+
+#include <QDomElement>
+//#include <QDebug>
+#include "SubWindow.h"
+#include "GuiApplication.h"
+#include "MainWindow.h"
+#include "exprfront.h"
```
jasp00: Please sort headers if possible.

plugins/xpressive/expressive_plugin.cpp
```diff
@@ -0,0 +1,926 @@
...
+Plugin::Descriptor PLUGIN_EXPORT xpressive_plugin_descriptor = { STRINGIFY(
+		PLUGIN_NAME), "X-Pressive", QT_TRANSLATE_NOOP("pluginBrowser",
+		"Mathematical expression parser"), "Orr Dvori", 0x0100,
+		Plugin::Instrument, new PluginPixmapLoader("logo"), NULL, NULL };
+
+}
+
+exprSynth::exprSynth(const WaveSample *gW1, const WaveSample *gW2, const WaveSample *gW3, ExprFront *_exprO1, ExprFront *_exprO2,
+					NotePlayHandle *_nph, const sample_rate_t _sample_rate, const FloatModel* _pan1, const FloatModel* _pan2, float _rel_trans)
+	:exprO1(_exprO1),exprO2(_exprO2),W1(gW1),W2(gW2),W3(gW3),nph(_nph),sample_rate(_sample_rate),pan1(_pan1),pan2(_pan2),rel_transition(_rel_trans)
```
jasp00: All kind of types begin with an upper case letter. Member variables are prefixed with `m_`.

Although this is work in progress, please follow the format style (one line per member) used in other parts of LMMS. Other developers may need to fix your code or help you.

plugins/xpressive/expressive_plugin.cpp
```diff
@@ -0,0 +1,926 @@
...
+{
+	note_sample=0;
```
jasp00: Infix operators should have a space before and after.

plugins/xpressive/expressive_plugin.cpp
```diff
@@ -0,0 +1,926 @@
...
+
+	exprO1->compile();
+	exprO2->compile();
+
+}
+
+exprSynth::~exprSynth()
+{
+	if (exprO1)
+		delete exprO1;
```
jasp00: You should use explicit blocking with `{` `}`.

plugins/xpressive/expressive_plugin.cpp
```diff
@@ -0,0 +1,926 @@
...
+	if (exprO2)
+		delete exprO2;
+}
+float saw_wave(float);
```
jasp00: It is a good idea to declare internal functions as `static`.

plugins/xpressive/expressive_plugin.cpp
```diff
@@ -0,0 +1,926 @@
...
+ *
+ *	class expressive
+ *
+ *	lmms - plugin
+ *
+ ***********************************************************************/
+#define GRAPH_LENGTH 4096
+
+expressive::expressive(InstrumentTrack * _instrument_track) :
+		Instrument(_instrument_track, &xpressive_plugin_descriptor), m_graphO1(
```
jasp00: Please do not add excessive indentation.

plugins/xpressive/expressive_plugin.h
```diff
@@ -0,0 +1,254 @@
...
+
+	virtual void saveSettings( QDomDocument & _doc,
+							QDomElement & _parent );
+	virtual void loadSettings( const QDomElement & _this );
+
+	virtual QString nodeName() const;
+
+	virtual f_cnt_t desiredReleaseFrames() const
+	{
+		return( 64 );
```
jasp00: Return without parenthesis.

plugins/xpressive/expressive_plugin.h
```diff
@@ -0,0 +1,254 @@
...
+	FloatModel m_panning2;
+	FloatModel m_relTransition;
+	float m_A1,m_A2,m_A3;
+	WaveSample m_W1, m_W2, m_W3;
+
+
+	
+	BoolModel m_exprValid;
+	
+	friend class expressiveView;
```
jasp00: Be aware this is not a clean model/view separation.

plugins/xpressive/expressive_plugin.h
```diff
@@ -0,0 +1,254 @@
...
+		return instance;
+	}
+
+private:
+	expressiveHelpView();
+	static expressiveHelpView * instance;
+	static QString HelpText;
+
+};
+class SubWindow;
```
jasp00: Please move these declarations to the top.

plugins/xpressive/expressive_plugin.cpp
```diff
@@ -0,0 +1,926 @@
...
+
+exprSynth::~exprSynth()
+{
+	if (exprO1)
+		delete exprO1;
+	if (exprO2)
+		delete exprO2;
+}
+float saw_wave(float);
+float sin_wave(float);
```
jasp00: Where are these functions used?

plugins/xpressive/exprfront.cpp
```diff
@@ -0,0 +1,533 @@
...
+		return powf(2,x/12);
+	}
+};
+static freefunc1<float,harmonic_semitone,true> harmonic_semitone_func;
+
+struct float_random
+{
+	static inline float process()
+	{
+		return 1.0f - rand() * 2.0f / RAND_MAX;
```
jasp00: `rand` is not thread-safe. It should be replaced later.

plugins/xpressive/exprfront.cpp
```diff
@@ -0,0 +1,533 @@
...
+		m_data->cyclics.push_back(wvf);
+		return m_data->symbol_table.add_function(name,*wvf);
+	}
+}
+size_t find_occurances(const std::string& haystack, const char * const needle)
+{
+	size_t last_pos=0;
+	size_t count = 0;
+	size_t len=strlen(needle);
+	while (1)
```
jasp00: You should use the standard true constant.

plugins/xpressive/expressive_plugin.cpp
gnudles: The repository is full with this mistaken address...
It can be done once with find/replace.

plugins/xpressive/expressive_plugin.cpp
gnudles: leftovers...

plugins/xpressive/expressive_plugin.cpp
gnudles: What do you mean? those two tabs or the list below?

plugins/xpressive/expressive_plugin.cpp
zonkmachine: It's was fixed a week ago so this is the only instance left with http:

plugins/xpressive/CMakeLists.txt
gnudles: But I want to disable those features on any platform...

plugins/xpressive/expressive_plugin.h
gnudles: I saw it in bitInvader (I started by copying bitInvader)...

plugins/xpressive/CMakeLists.txt
gnudles: Oh... now it's not compiling because exprtk.hpp still throws exception in line 2733
@ArashPartow, should I just make a little fix like:
if (stride > 4)
            {
-               throw std::invalid_argument("token_scanner() - Invalid stride value");
+               stride = 4;
            }
????

plugins/xpressive/expressive_plugin.cpp
zonkmachine: The tabs!

gnudles:
I fixed everything that @jasp00 noted,

> Please include previous value of CMAKE_CXX_FLAGS
 but unfortunately when I added the previous value, it couldn't be compiled, because there is still another throw statement in exprTk that slipped through all those disabling defines.

simonvanderveldt:
Cool stuff!
Small question: Shouldn't new plugins use the new styling? See #2831
Or will this be done after this PR is merged?

zonkmachine:
> I fixed everything that @jasp00 noted,

@gnudles Awesome! Push the changes.


> > Please include previous value of CMAKE_CXX_FLAGS
> 
> but unfortunately when I added the previous value, it couldn't be compiled, because there is still another throw statement in exprTk that slipped through all those disabling defines.
> When it will be fixed, I think we are ready for merge...


Is there an upstream issue about this we can keep track of?


> Small question: Shouldn't new plugins use the new styling? See #2831
> Or will this be done after this PR is merged?

#2831 Is concerning the native effect plugins specifically.

simonvanderveldt:
> #2831 Is concerning the native effect plugins specifically.

@zonkmachine yeah, I always forget there's a distinction between instruments and effects :x

gnudles:
> @gnudles Awesome! Push the changes.

But if I do, it won't compile

> Is there an upstream issue about this we can keep track of?

What do you mean?

plugins/xpressive/CMakeLists.txt
ArashPartow: @gnudles  which plaform are you building where you see that exception being thrown?

gnudles: The exception is not being thrown. But I can't compile it with the flag -fno-exceptions

plugins/xpressive/expressive_plugin.h
jasp00: If you have some spare time, please use inline getters and setters instead of `friend`.

plugins/xpressive/exprfront.cpp
```diff
@@ -0,0 +1,610 @@
...
+0x478eb75a, 0x898370a6, 0x73e7cca3, 0x34ace0ad,
+0xc8ecb388, 0x5375c3aa, 0x9c194d87, 0x1b65246d,
+0xca000bcf, 0x8a0fb13d, 0x81b957b0, 0xac627bfb,
+0xc0fe47e5, 0xf3db0ad8, 0x1c605c7d, 0x5f579884,
+0x63e079b5, 0x3d96f7cf, 0x3edd46e9, 0xc347c61e,
+0x7b2b2a0e, 0x63dfcf51, 0x596781dd, 0x80304c4d,
+0xa66f8b47
+};
+
+inline unsigned int rotate(unsigned int x, const int b)
```
jasp00: You should call this `rotateLeft`. Also, you should check if `b == 0` because you could cause undefined behavior.

jasp00:
@gnudles
Regarding `exprtk.patched.hpp`, if you use `assert()` and define `NDEBUG`, you silence errors. You should print a message to standard error, even once.

@zonkmachine
>I probably have the weakest hardware of the developers

Are you able to compile `exprfront.cpp`? It is too big for my system. I will not be able to help if we do not split the file.

gnudles:
@jasp00, it shouldn't reach there anyway...

jasp00:
>it shouldn't reach there anyway...

So it will be a cold function. Could you write this with your words?
```
#define xAssert() perror("exprtk exception")
```

zonkmachine:
> Are you able to compile exprfront.cpp?

I can compile the latest commit and it runs fine, if that is what you meant.

jasp00:
>I can compile the latest commit and it runs fine

If you `touch exprfront.cpp` and `time make`, how long does it take?

zonkmachine:
```
real	0m52.185s
user	0m48.320s
sys	0m1.800s
```

gnudles:
Maybe it should be an external plug-in?

jasp00:
>Maybe it should be an external plug-in?

No, I will split the file if necessary.

gnudles:
I don't like to use exprtk.patched.hpp, I prefer to use the original, but currently I cannot do so. I'm waiting for @tresf or @ArashPartow for further guidelines.

tresf:
> Small question: Shouldn't new plugins use the new styling? See #2831

For effects plugins, certainly.  For instrument plugins like this one... Ideally, but not mandatory.

Why not mandatory?  Well, this would generally require the OP to grant commit access to his/her repository to a designer like @RebeccaDeField requiring coordination/collaboration.  I'll defer that question to the OP and @RebeccaDeField if that can happen prior to merge.  If not, we can follow up with this task afterward.

> I don't like to use exprtk.patched.hpp, I prefer to use the original, but currently I cannot do so. I'm waiting for @tresf or @ArashPartow for further guidelines.

Can you or @jasp00 please summarize what's left (or the problem) on this file/task?  If the patch can be automated, @jasp00 has written some nice upgrade routines in other PRs we can leverage but if it's too complex to do, I'd like to know the harm in bundling the patched file.

We can also ship a `.diff` file with the `.hpp` and use that at compile time as often the `.diff` is easier to merge with upstream changes over a patched snapshot.

But I'm just speaking hypothetically as I don't know enough about the history of this source file to be able to help.  Perhaps my name was tagged accidentally?

gnudles:
@tresf, the question is - do I have to respect the -fnoexceptions flag, because it forces me to make small changes in the exprtk utility.

tresf:
> 

@jasp00 does some custom cmake tasks on his unmerged `clang` branch.  The method would bear some similarities.

First, he issues a custom cmake [`COMMAND`](https://github.com/jasp00/lmms/blob/clang/plugins/GigPlayer/CMakeLists.txt#L15-L34) which could be adapted to apply a diff.

Second, he shims the custom file into the [`BUILD_PLUGIN`](https://github.com/jasp00/lmms/blob/clang/plugins/GigPlayer/CMakeLists.txt#L38-L42) command.

His is configurable due to the reliance on a system header.  Yours can be hard-coded.

plugins/xpressive/CMakeLists.txt
```diff
@@ -1,7 +1,16 @@
...
 
-SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DNDEBUG -Dexprtk_disable_sc_andor -Dexprtk_disable_enhanced_features -Dexprtk_disable_break_continue -Dexprtk_disable_comments -Dexprtk_disable_string_capabilities -Dexprtk_disable_rtl_io_file -Dexprtk_disable_rtl_vecops ${WERROR_FLAGS}")
+SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -I${CMAKE_CURRENT_BINARY_DIR} -Dexprtk_disable_sc_andor -Dexprtk_disable_return -Dexprtk_disable_enhanced_features -Dexprtk_disable_break_continue -Dexprtk_disable_comments -Dexprtk_disable_string_capabilities -Dexprtk_disable_rtl_io_file -Dexprtk_disable_rtl_vecops ${WERROR_FLAGS}")
 IF(WIN64)
 	SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wa,-mbig-obj")
 ENDIF()
-BUILD_PLUGIN(xpressive expressive_plugin.cpp exprfront.cpp expressive_plugin.h MOCFILES expressive_plugin.h EMBEDDED_RESOURCES "${CMAKE_CURRENT_SOURCE_DIR}/*.png")
+#file(GENERATE OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/exprtk.hpp"
+#	INPUT "${CMAKE_CURRENT_SOURCE_DIR}/exprtk.hpp.origin")
+file(COPY "${CMAKE_CURRENT_SOURCE_DIR}/exprtk.hpp.origin" DESTINATION "${CMAKE_CURRENT_BINARY_DIR}")
```
tresf: `exprtk.hpp` is just fine.  No need to call it `origin`.

plugins/xpressive/exprtk_patch
```diff
@@ -0,0 +1,64 @@
+57,58c57,69
```
tresf: `exprtk.hpp.diff` or `exprtk.diff` would be the accepted file extension.  This will trigger syntax highlighting in most code editors.

tresf:
@gnudles this looks great.  I left a few comments.

I've been thinking about the best place to put the patched `.hpp` file and I'm interested to know opinions on this:

* **ORIGINAL**: `plugins/xpressive/exprtk.hpp`
* **PATCH**: `plugins/xpressive/exprtk.hpp.diff`
* **COMPILE**: `plugins/xpressive/exprtk.patched.hpp`
* **GITIGNORE**:  `plugins/xpressive/.gitignore` entry: `*.patched.*`

Apologies for the slight change of direction.  It think leaving it in the source directory but omitting it from git is cleanest.

RebeccaDeField:
> I'll defer that question to the OP and @RebeccaDeField if that can happen prior to merge. If not, we can follow up with this task afterward.

I am currently busy wrapping up the effects plugins with some other contributors. I would suggest that if we want to redesign all of the other plugins, we follow up with that project later and once [this](#2831) is done. We would want to decide on some consistent design conventions for those plugins (I will go into this later because I want to stay on topic), I think this can be pulled in without a redesign for the time being.

plugins/xpressive/CMakeLists.txt
gnudles: I renamed it in order that it won't be included by the preprocessor...

tresf: @gnudles I'm pretty sure the `cmake` configure step runs before all of that.  Can you explain?

jasp00:
>I've been thinking about the best place to put the patched `.hpp` file

Please be simple, just use one consolidated file. Prior to merging, we should keep the original file around to make a patch for upstream.

>do I have to respect the -fnoexceptions flag

@gnudles, try and see if you can use exceptions.

plugins/xpressive/CMakeLists.txt
gnudles: I'm sorry, I don't understand the question...
If you have better place for the cmake patch script then tell me...

gnudles:
I don't know why it stopped compile. Any suggestions? it looks like the PLUGIN_NAME define is being ignored...

jasp00:
>Any suggestions?

Please edit `cmake/modules/BuildPlugin.cmake` to stop including `${CMAKE_SOURCE_DIR}/src/gui`. Then include `embed.h` instead of `embed.cpp`.

Umcaruje:
So we don't want this plugin in 1.2 after all?

jasp00:
From [build log](https://travis-ci.org/LMMS/lmms/jobs/218398018),
```
Fatal error: can't write CMakeFiles/xpressive.dir/exprsynth.cpp.obj: File too big
```
Please chop the code.

gnudles:
yeah, it should be ok now... I tried to enable the variable-constant optimizations (exprtk_disable_enhanced_features).
I also moved the patching process to occur in the build time.

gnudles:
I think it's ready to be merged... Isn't it?

jasp00:
>I also moved the patching process to occur in the build time.

[Please](https://github.com/LMMS/lmms/pull/3259#issuecomment-289730206) be simple, use consolidated files while keeping the original header.

What error do you get if you remove `-Dexprtk_disable_` flags from the normal build?

gnudles:
> What error do you get if you remove -Dexprtk_disable_ flags from the normal build?

It make the compilation process slower and it adds more throw statements...

> Please be simple, use consolidated files while keeping the original header.
But @tresf said that it's OK...

gnudles:
BTW, the error in the build system is not my fault...

jasp00:
>It make the compilation process slower

But it yields more features, right? Compilation speed may be decreased by splitting logic.

>and it adds more throw statements...

Which you know how to handle. Did you try to build with exceptions? Does ExprTk use any callbacks?

>But @tresf said that it's OK...

As @tresf [points out](https://github.com/LMMS/lmms/pull/3259#issuecomment-289231753), we have full control over the bundled source. We are not using `dpatch` or `quilt`, thus working with patches is more complicated. So please be simple and use consolidated files.

gnudles:
> Does ExprTk use any callbacks?

Yes. it receives some callbacks but I only use callbacks in polymorphism manner, I mean through a class overridden method, and not as a bare function. 
Remeber when you told me on the walk-through :
> Please include previous value of CMAKE_CXX_FLAGS

Beforehand, when it was without the previous value of CMAKE_CXX_FLAGS, it was compiled without the -fnoexceptions flag, and everything worked correctly, at least on my machine (linux 64bit). Maybe it poses some problems on Windows(R) or OS/X, although it shouldn't.
If exceptions will be turned on, I won't have to patch the lib ( except for the exprtk_disable_return flag that I added, but we can just ignore the fact that return statement exists... )

> But it yields more features, right?

I've already told you that those features are an overkill. I do not want to enable assignment operation, and enabling while-loops may lead to infinite loop and even if it won't, it's hard to render 44100*2 samples per second when you do too much operation without JIT compiler...
About the flags:
exprtk_disable_sc_andor - I'm not so sure about it because and/or operations still working without it... (maybe it does some optimizations).
exprtk_disable_return - I made that flag to disable return statement.
exprtk_disable_break_continue - disables some control statements
Dexprtk_disable_comments - disable parsing comments in the expression.
Dexprtk_disable_string_capabilities - disable parsing strings and manipulate them in the expression.
Dexprtk_disable_rtl_io_file - disable file I/O
Dexprtk_disable_rtl_vecops - disable vector operations... I haven't tried that feature, but I don't think it's useful in our case...

> Compilation speed may be decreased by splitting logic.

It's not so simple to split this file. You need to master all the inner dependencies and you should also remember that it's a templates file.
I know it's annoying to compile this file, and it's the largest source file in LMMS for sure, but you have the choise if to insert it or not.

tresf:
> Please be simple, just use one consolidated file. Prior to merging, we should keep the original file around to make a patch for upstream.

Patching is fine.  It's a clean 61-line diff.  I do agree we should add a `FIXME` or `TODO` as a reminder to file the patch upstream.

@jasp00 if you require this patched prior to merging, perhaps you can help direct us [where to submit it](http://www.partow.net/programming/exprtk/index.html).

The main reasoning for recommending to patch this file is @gnudles [didn't want to include the patched version](https://github.com/LMMS/lmms/pull/3259#issuecomment-288686998) and I don't blame him since the diff is small and the original source is huge.

That conversation happened 5 days ago, so I think we had assumed no news was good news. :)

gnudles:
The patch is combined from two main changes:
1) a workaround to compile with -fnoexception flag...
2) a flag to disable return statement... 
So AFAIK if you want to push the changes to upstream, only change #2 should be considered...
BUT! I talked with Arash, and he said he will do it ([change #2](https://github.com/gnudles/lmms/pull/2#issuecomment-291642100)) .

> perhaps you can help direct us where to submit it.

Perhaps via pull request in github?

tresf:
> Perhaps via pull request in github?

My apologies, I didn't realize @ArashPartow was the upstream author of the library.

Are you saying upstream can take all of our changes, or no?  If they can't, we should keep the diff.  If they can, we should PR. 👍

gnudles:
They shouldn't, because regular users of exprtk.hpp can use exceptions in their environment... ;)
And the other change is (probably) already being done by the author.
Just to be sure, when you say "upstream" you mean the exprtk repository, managed by Arash, right?

plugins/xpressive/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+INCLUDE(BuildPlugin)
+
+SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Dexprtk_disable_sc_andor -Dexprtk_disable_return -Dexprtk_disable_break_continue -Dexprtk_disable_comments -Dexprtk_disable_string_capabilities -Dexprtk_disable_rtl_io_file -Dexprtk_disable_rtl_vecops ${WERROR_FLAGS}")
+IF(WIN32 OR WIN64)
```
tresf: [`WIN32`](https://cmake.org/cmake/help/v3.0/variable/WIN32.html) actually covers both cases, `WIN64` can safely be removed.

tresf:
Ok, so I'm reading through the history of this and I feel @jasp00 can be more specific with the review process.

@jasp00 keeps stating things like "split the file", "keep it simple" yet @gnudles has clearly stated that not all of these patches aren't upstream worthy.

Can you please nudge this PR in the right direction?  If not, we'll merge as-is.  This PR is very clean and the `diff` file is sane and small.

If we can leverage exceptions for this plugin, please instruct @gnudles to do so.  If we can't, the `diff` is an acceptable approach because it will make for quicker merging from upstream.

At one point @softrabbit asked if we could make this library part of the base code.  Should that be handled in a separate PR?

> Just to be sure, when you say "upstream" you mean the exprtk repository, managed by Arash, right?

Yes.  And on a side note, thanks for the work on this so far.  @jasp00 has recently re-joined our team and brings a lot of expertise, so PRs are getting more attention to fine detail.  I hope that's ok. :)

ArashPartow:
@gnudles @tresf

> a flag to disable return statement...
> So AFAIK if you want to push the changes to upstream, only change # 2 should be considered...
> BUT! I talked with Arash, and he said he will do it (change #2) .


I think I may have misunderstood the _actual_ requirement here.

My understanding was you wanted to disable the use of the return statement in expressions. So i went and added an extra settings mode to the parser settings.

But reading this thread, it seems your intention was not only to disable return statements but to also allow for a build with **fno-exceptions**?

If that is correct, we'll probably need to add the if-defs @gnudles proposed earlier on. Please let me know if that's what you want to do.

tresf:
> If that is correct, we'll probably need to add the if-defs @gnudles proposed earlier on. Please let me know if that's what you want to do.

@ArashPartow thanks kindly for keeping track of this PR. :)  I had to re-read this as well... yes, it seems to be two-fold however I'll let @gnudles and @jasp00 make that call.

ArashPartow:
@tresf @gnudles 

I've added both [options](https://github.com/gnudles/lmms/pull/4#issue-219736929), you can choose to use whichever is most appropriate.

ArashPartow:
@gnudles

In the xpressive's CMakeLists.txt you have the following:

```
IF(WIN32 OR WIN64)
   SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wa,-mbig-obj -Dexprtk_disable_enhanced_features")
ENDIF()
```

My understanding was that defining exprtk_disable_enhanced_features was only required for the 32-bit MinGW version, and was not required for the 64-bit version.

Is this not the case anymore?

tresf:
> My understanding was that defining exprtk_disable_enhanced_features was only required for the 32-bit MinGW version, and was not required for the 64-bit version.

In that case perhaps... 

```diff
- IF(WIN32 OR WIN64)
+ IF(WIN32 AND NOT WIN64)
     SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wa,-mbig-obj -Dexprtk_disable_enhanced_features")
  ENDIF()
```

tresf:
Or perhaps... 

```cmake
IF(WIN32)
   SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wa,-mbig-obj")
   IF (NOT WIN64)
      SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Dexprtk_disable_enhanced_features")
   ENDIF()
ENDIF()

**Comment by ArashPartow** [comment]
@tresf the last option seems most appropriate.

**Comment by gnudles** [comment]
> My understanding was that defining exprtk_disable_enhanced_features was only required for the 32-bit MinGW version, and was not required for the 64-bit version.

At the following build:
https://travis-ci.org/LMMS/lmms/builds/192966413
It failed on "file too large" on WIN64 so I disabled many features including the exprtk_disable_enhanced_features on ALL platforms (mistakenly).
But, recently I discovered that this flag only adds optimization... so I enabled the enhanced features on ALL platforms except WIN64. and then the WIN32 build failed with "file too large" error, so I disabled the enhanced features on both of them.

> Edit: or perhaps...

What @tresf suggests seems good.

> I've added both options, you can choose to use whichever is most appropriate.

Thanks a lot! (it indeed able to throw return_exception() if not disabled...)

CONCLUSIONS - 
I'll fix the CMAKE for the WIN64/32 flags.
I'll merge the new exprtk library.
@jasp00 should instruct me about the -fno-exceptions if he wants to...

**Comment by gnudles** [comment]
> IF (NOT WIN64)

Wait, now I'm confused... @ArashPartow you said before that the ld of Mingw for WIN64 is broken, not for WIN32...

**Comment by gnudles** [comment]
You can see in the log : https://travis-ci.org/LMMS/lmms/jobs/219417335
that the fail occurs on WIN64. I'm updating the Cmake as I suggested...

**Comment by gnudles** [comment]
I must leave tomorrow's morning, and I won't have an internet connection for a while. Can somebody take control?

**Comment by tresf** [comment]
@gnudles yeah, paraphrasing https://github.com/LMMS/lmms/pull/3259#issuecomment-273447592 that is what it would appear.

Note, to be as technical as possible... if this problem is with `mingw-w64`, the yes it would affect both `mingw-w64` 32-bit and `mingw-w64` 64-bit alike.  (Yes, it's [exactly as confusing](https://sourceforge.net/projects/mingw-w64/) as it reads).

And to nitpick, @ArashPartow has told us that this bug doesn't necessarily effect MSVC++, so to be proper, we should validate `MINGW_PREFIX`, as we may offer MSVC++ support in the next year or so.

**Comment by tresf** [comment]
> Can somebody take control?

This rarely works with forked repositories, so if this needs to happen we would have to supercede your PR, or you'd have to grant us commit access on your fork.  If we know how long you'll be gone for, that would help the decision.

**Comment by gnudles** [comment]
I hope it did it.
I might have a random access in about two weeks from now.

**Comment by Umcaruje** [comment]
Github already allows this without you providing us access to your whole fork: 

https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/
https://help.github.com/articles/committing-changes-to-a-pull-request-branch-created-from-a-fork/

**Comment by jasp00** [comment]
>@jasp00 should instruct me about the -fno-exceptions if he wants to...

* https://github.com/LMMS/lmms/pull/3259#issuecomment-289730206: "@gnudles, try and see if you can use exceptions."
* https://github.com/LMMS/lmms/pull/3259#issuecomment-291482763: "Did you try to build with exceptions?"

>If we can leverage exceptions for this plugin, please instruct @gnudles how best to do so.

Build with `-fexceptions`. Do you know what a try-block is?

>I've already told you that those features are an overkill.

You do not like those features, fine, do not use them, but allow other users to choose.

>It's not so simple to split this file.

It is not that hard, take your time and ask for guidance, upstream is cooperating. You should be able to enable all features and maybe drop MinGW flags.

>diff is an acceptable approach

It is not the preferred form for making modifications, you do not edit the diff file. A simple alternative would be to ship the patched file and the diff to the original. Anyway, we can try not to have any diff.

**Comment by tresf** [comment]
> Do you know what a try-block is?

No, not in C++ and not in LMMS.  I was hoping you could help out.

> It is not the preferred form for making modifications

For some projects it is.  MacPorts uses this technique and it's for identical purposes (to fix compilation or compatibility problems not yet acknowledged, fixed or intended to be included by upstream)

> Anyway, we can try not to have any diff.

I agree that sounds like the best option, but I will not prevent this feature from being merged on that reason alone.

**Comment by jasp00** [comment]
Are you taking care of this, @tresf?

A [try-block](http://en.cppreference.com/w/cpp/language/try_catch) is a sequence of one try-clause and one or more catch-clauses. For this plug-in, this should work:

```c++
	try
	{
		// ExprTk calls
	}
	catch(...)
	{
		qWarning("ExprTk exception");
	}
```

tresf:
Not yet.  I won't​ be able to for some weeks. If you have the bandwidth and desire, please do.

JordanKyser:
I see that this is an advanced BitInvader plug-in, and you can put your formula for a particular wave (as in soundwave, not the .wav file) sample. I like BitInvader, and I will sure use Xpressive when it is implemented in LMMS.

gnudles:
@ArashPartow, Compiling with some disable flags seem to cause some troubles...

ArashPartow:
Looks like a whole bunch of things were missed in the merge.  In any case I've made a new PR: [ 	e330a93](https://github.com/gnudles/lmms/pull/6)

zonkmachine:
It passes Travis, builds fine and works like a charm. :whale:

plugins/xpressive/.gitignore
```diff
@@ -0,0 +1,2 @@
+#ignore the patched file
+exprtk.patched.hpp
```
tresf: Has the patched file been removed? (can this file be removed now?)

cmake/modules/BuildPlugin.cmake
```diff
@@ -9,7 +9,7 @@ MACRO(BUILD_PLUGIN PLUGIN_NAME)
 	CMAKE_PARSE_ARGUMENTS(PLUGIN "" "" "MOCFILES;EMBEDDED_RESOURCES;UICFILES;LINK" ${ARGN})
 	SET(PLUGIN_SOURCES ${PLUGIN_UNPARSED_ARGUMENTS})
 
-	INCLUDE_DIRECTORIES(${CMAKE_CURRENT_BINARY_DIR} ${CMAKE_BINARY_DIR} ${CMAKE_SOURCE_DIR}/include ${CMAKE_SOURCE_DIR}/src/gui)
+	INCLUDE_DIRECTORIES(${CMAKE_CURRENT_BINARY_DIR} ${CMAKE_BINARY_DIR} ${CMAKE_SOURCE_DIR}/include)
```
tresf: I'm sorry if this was already explained previously, but was `${CMAKE_SOURCE_DIR}/src/gui` deemed as unnecessary for all of our plugins and was lingering from something else?

tresf: Update, this appears to [date back 10 years](https://github.com/gnudles/lmms/blob/e637495/cmake/modules/BuildPlugin.cmake#L50) and I see no evidence it's needed as headers aren't in that directory.  Should be safe to remove. 👍

plugins/xpressive/.gitignore
gnudles: The patch was removed due to compilation with the -fexceptions flag and the release of new exprtk version that allow one to disable return statement.

gnudles: All calls for Exprtk are now surrounded by try-blocks.

gnudles:
I think it is ready to be merge. The only disadvantage is that you can't draw a wave shape by hand. (I disabled that possibility when I began working on this plug-in to make things simpler). Enabling that require a little work, and I don't have the time for that right now...
Anyway, I recently added the tempo variable to the evaluator, and this is awesome because you can make a tempo aware synthesizer.

Umcaruje:
@gnudles can you please change the base branch to `stable-1.2`, not master

zonkmachine:
> @gnudles can you please change the base branch to stable-1.2, not master

Is this a matter of just [changing the setting](https://github.com/blog/2224-change-the-base-branch-of-a-pull-request)?

Hussam-Eddin-Alhomsi:
I really like the idea behind the name of the plug-in that it handles mathematical expressions and the name describes it really well but I suggest we change the name of the plug-in from "**X-Pressive**" to "**X-pressive**" because capitalizing the the letter p in the first one makes it like **ExPressive** rather than "**Expressive**".

zonkmachine:
Tested the graph draw feature. Nice! You can't draw unless the smooth knob value is 0 as I understand it and the crossmark is inactive to reflect this. It works.

Umcaruje:
@zonkmachine I don't have time to test this, is this ready for a merge?

zonkmachine:
> @zonkmachine I don't have time to test this, is this ready for a merge?

I think so and @gnudles comment some 5 posts up reads:

> I think it is ready to be merge.

That's good enough for me.
@tresf merge?

Lukas-W:
Not good to merge. Due to the rebasing to `stable-1.2` this includes some commits from `master` that are not present in `stable-1.2`, such as de3b344edd4ebb7021b589869e8eaff69a8840d1 and 9f905bce3eee4e24a2338d9d8a104af6bdbec8ad.

Umcaruje:
>Not good to merge. Due to the rebasing to stable-1.2 this includes some commits from master that are not present in stable-1.2, such as de3b344 and 9f905bc.

Oh damn how did I miss this? @gnudles since your changes are mostly focused on a single folder, I would recommend backing up your changes, hard-resetting to `stable-1.2` and reapplying your changes on top of that. Note that you'll need to force-push

Umcaruje:
Or if someone else has something better in mind it'd be interesting to see how one can get out of this mess

Lukas-W:
I'd say we just merge it into `master`, as the it was intended originally. We'll never release 1.2 if we don't start taking "feature freeze" seriously. 

We can release 1.3 as quickly as possible after 1.2 is out (we could go into feature freeze pretty much immediately). We should move towards a much quicker release cycle anyway. Our last feature release 1.1.0 was in 2014!

Umcaruje:
>I'd say we just merge it into master, as the PR was intended originally. We'll never release 1.2 if we don't start taking "feature freeze" seriously.

Yeah I guess you're right.

zonkmachine:
So, bump this to 1.3?

zonkmachine:
@gnudles I looked through the commits and these are the ones that seem to be already in stable-1.2. Mind you, I'm not super savvy with git...

 `*` Marks merge commits


`Small fixes:	`								953d59ffd7c6126ed66dc0660f8ab9e137fc373b
`Added Buzzer Beater to demos (#3482)	`					32f07ac317b42dda42a1d00b0a3b1fa1feef0655
`Simpler SF2 Player file dialog. Copied from GIG Player. Opens correct…`		09c9ccb34dc6bb54954da9291ddfed3b02b4ebbd
`Fixed #3182. Always using master channel for preset previews. (#3503)` 		2df56829a31151f448acb48cdacbcc96c0a0e35e
`Dual Filter Plugin Redesign (#3484)`						2f51062d1e6b144a0ffe5229ba445f262d62ff3f
`EQ plugin now responds to wet / dry control`					a7154df004be92e1692b41f3967476eb0819b0a4
`Revert "'Off beat' beat notes to MelodyPattern" `				082dbed8e9eb7194c96e0f51e79b6e7d84032a03
 `* Merge pull request #3520 from curlymorphic/eqWetDry`				0190256801eab505967c43b44ddfc3f713451f31
` * Merge pull request #3522 from curlymorphic/delaySync`				77496188ae7c6e12f0b432e69db7b7325857de98
`Disable HiDPI on Windows`							d2e50df3ce2f8a56fd1750cc077eff3e76a6867d
` * Merge pull request #3531 from tresf/hidpi` 					6a5255e3795cf2079e3d1bebef5f32ef38876cff
`Make factory samples relative (#3510)` 						7e3ee14cf1a2d3e2975256c1fc2b8fb37d4c1c63
`LFO controller now has correct frequency with multiple connections. `		1848ed20d9802c3615138a3d67a883df8d7c34dc
`Fix automation processing in BB tracks (#3481)` 					75077f6200a5aee3a5821aae48a3b8466ed8714a
`Merge pull request #3535 from curlymorphic/lfo `					ff4ffd17f41ffd04ab179492868c40a55c591f35
`Implement 24 bit support for WAV export (#3021) `				4196a3041535010e7fa70c554cbd690a32a5ca91
`Simplification of code and structure in the area of the file export ` 		d5fc38ebd9c45000a745aa0b2dd8d5d25bf1a2fb
`Only show relevant widgets on the export dialog` 				4ac6aa909d5e804a8101e340fd5d1d88d2faa6ca
`Add the option to enable variable bit rates to the export dialog` 		823788919a1d3d411162eba425c6a6df1d03055e
`Add copyright info to the file OutputSettings.h` 				f06eac3f91bd77f51f5be1b560817ef8ab8fb84e
`Make the 24 bit export the default selection in the export dialog` 		4b3309a4e991d01542e980cd07237ae2106e01e3
`Rename entry for bit depth from "24 Bit Float" to "24 Bit Integer"` 		dfeb6bea3a415ffb9757ff4b7018cd7714bf0ee6
` * Merge pull request #3552 from michaelgregorius/3021-24-Bit-Export-For…` 	08201c202494c25ab8a151c98118dfa0359500e6
`Remove Limited Session` 								290556e43d200952e85a9f54607af9501621d484
`Provide support for fallback config values (#3551)` 				17b37e02a986c05b5a1af16f2852061b088a68eb
`don't change sampleTco length if tempo changed (#3543)` 				69d09f01865771c9604be002a3568d72ca9a90d2
`ReverbSC: Method to change samplerate (#3401) (#3403)` 				f390fd1723452f56740c4feb5ebcc2a6debd6da3
`Update Discord invite to link to #welcome (#3556)` 				b0ae31c2ffa96432e25ddbefe90b62270483cdae
`Remove glitching from the Flanger and Delay plugins (#3524)` 			b432707b9beba0bee1e37281bfa1c5e98e977172
`Flanger LFO rate syncronised (#3521)` 						300058df0430848697a9ef6f0175857822d4cdd9
`Replacement metronome samples (#3513)` 						c0b910ec578889065b8e46734bfdb36f14f74403
` * Merge branch 'stable-1.2' into xpressive_plugin`				887eeda6567a327cc73f458182b537f7f3d280cf

zonkmachine:
As the branch has already been switched back to master and this needs some more work, I bump this to the 1.3.0 milestone.

zonkmachine:
@tresf If I've understood this correctly the conflicts here will be easier to fix once https://github.com/LMMS/lmms/pull/3607 is merged, no?

tresf:
> If I've understood this correctly the conflicts here will be easier to fix once #3607 is merged, no?

Looks like that but I think it's just suffering from some branch switching.  Someone with some patience should be able to rebase against master again without losing work.  I say this because the conflicts above aren't actually edited in this PR.

zonkmachine:
> Someone with some patience...

OK. I've fixed it.
 * Branched master
 * Cherry-picked all commits on `gnudles:xpressive_plugin`...
   * that was from @gnudles except the first one, 953d59ffd7c6126ed66dc0660f8ab9e137fc373b, which is already in master.
   * that was from **ArashPartow** .
   * but omitted the merge commits.

Compiled, runs and still has the groove. :man_dancing: 

@gnudles @tresf Do you want me to push this to a separate branch or should I force push it over **gnudles/lmms** directly?

zonkmachine:
@tresf Fixed this but now there's a merge conflict over here: https://github.com/gnudles/lmms/pull/8

tresf:
> @tresf Fixed this but now there's a merge conflict over here: gnudles#8

Right, I guess we would expect that.  Any PRs against his PR would go into conflict.

> It looks like the easiest way is to reset this branch to before my fixup and then merge that one and cherry-pick it to my fixup branch?

The offending PR is on someone else's fork and was performed by @ArashPartow (the author of exprtk).  He's best to answer this question.  The unknown branch may be because this PR was handled elsewhere and he deleted his branch.  He's best to answer that question but at a glance, it appears you may be able to hand-edit them into this PR prior to merging.  The changes are minor.

zonkmachine:
```diff
-               typedef std::iterator_traits<typename generator::token_list_t::iterator>::difference_type diff_t;
+               typedef std::iterator_traits<generator::token_list_t::iterator>::difference_type diff_t;
```

tresf:
Oh, I see.   We if you've already incorporated them, the PR is moot.

zonkmachine:
```diff
-               typedef std::iterator_traits<typename generator::token_list_t::iterator>::difference_type diff_t;
+               typedef std::iterator_traits<generator::token_list_t::iterator>::difference_type diff_t;
```

Sorry, I pressed send there by mistake. The second (+) line above is indeed in [exprtk.hpp upstream](https://raw.githubusercontent.com/ArashPartow/exprtk/master/exprtk.hpp) but neither of the lines is anywhere in the versions I have. Either I've messed up or the left behind PR was never incorporated for a reason. Let's wait a couple of days for some feedback @gnudles and/or @ArashPartow.

zonkmachine:
> The second (+) line above is indeed in exprtk.hpp upstream but neither of the lines is anywhere in the versions I have.

_Edit: apart from the backed up original..._

gnudles:
@zonkmachine, you can always get the newest exprtk.hpp from it's repository. Thanks for the fixes BTW.
The exprtk.hpp in the plugin is the original file without any change, so it should be OK to take the newest.

zonkmachine:
Aight, upated! :wink:

zonkmachine:
> Not good to merge. Due to the rebasing to stable-1.2 this includes some commits from master that are not present in stable-1.2, such as de3b344 and 9f905bc.

I believe this is fixed now. Merge?

Lukas-W:
> Merge?

👍

zonkmachine:
@tresf This is a big one. Does anything get squashed or just merge it as it is?

tresf:
@zonkmachine if we wanted we could isolate the Graph additions and plugin cmake cleanup, but they're actually relatively small, I vote squash 'em all.

zonkmachine:
> I vote squash 'em all.

I'm fine with that but how about keeping the two main contributors apart and squashing the exprtk.hpp commits by Arash (amended by my update) separately so it gets recognized on the contributors list. I'm a bit sentimental that way. I tested it in a separate branch and the rebase comes out with no bugs and there is no diff against `gnudles:xpressive_plugin`.

It would look something like:

```
commit fa59472eed44f5e40d011bf3366874949720be89
Author: Orr Dvori <dvoreader@gmail.com>
Date:   Wed May 17 22:43:31 2017 +0300

    X-Pressive plugin - added graph drawing feature

commit acee8d3c6aefdcb7eaa047ce83881e3852ee2701
Author: ArashPartow <partow@gmail.com>
Date:   Sat Jan 14 05:21:23 2017 +1100

    Update ExprTk to tip of branch

commit 9b646c5c01266ef9229a6ff3f49c4ba32fcdb356
Author: Orr Dvori <dvoreader@gmail.com>
Date:   Sun Jan 8 14:31:01 2017 +0200

    X-Pressive Plugin
    
    Synthesizer plugin based on Bitinvader and 'The C++ Mathematical
    Expression Toolkit Library (ExprTk)'
    
    Some initial functions.
    
    available keys:
    f- note's frequency. available only in the output expressions
    t- time in seconds. in the Waves (W1,W2,W3) it's in the range [0,1)
    key- the note's keyboard key. available only in the output expressions.
    v- the note's velocity (divided by 255.0 so it is in the range [0,

...
```

tresf:
In my experience squash maintains the contributors.

zonkmachine:
Ah, OK! Then just a Squash and merge.

zonkmachine:
> In my experience squash maintains the contributors.

It doesn't [look like it](https://github.com/zonkmachine/lmms/graphs/contributors?from=2017-06-09&to=2017-06-26&type=a).

tresf:
What is the concern?  It gives him credit?

image

zonkmachine:
> What is the concern? It gives him credit?

Yes, gnudles but not Arash. It's not a big concern though, I'm more trying to get to grips with git and github.

ArashPartow:
@zonkmachine - Gnudles did all the heavy lifting, I'm if I don't get through.  :D

gnudles:
Maybe we should credit Arash in the plugin itself in the help window.
BTW, shouldn't it be merged to stable1.2?

zonkmachine:
> BTW, shouldn't it be merged to stable1.2?

Agree. In the works here: https://github.com/LMMS/lmms/pull/3845

Unesty:
How to add external sample to W?

gnudles:
you have the load button

Unesty:
I tried to load sample and it became distorted. Is that how supposed to be?

gnudles:
the wave is loaded to a buffer with 4096 samples or so (4096 floats or shorts, can't recall). 4096 is a huge buffer compared to other plugins in lmms. but if you load a long sample it will become distorted (resampled). it is not a bug, it's a feature :)

PR #3273, opened 2017-01-18
Fix inverted zoom direction

### LINKED ISSUE #3232: Inverting the Zoom Direction
Is there a way to invert the zoom controls for the scroll wheel? Moving the scroll wheel in the direction that zooms in prior versions appears to zoom out instead in 1.2 RC2 and vice versa. Is this a bug, or is there an option to invert this somewhere?

I'm using the 64bit version of 1.2 RC2 on Windows 7.

It also might be worth noting that when I initially launched this version, the color scheme matched the old version except for several editor backgrounds which were solid black. Restarting the application appeared to fix that though.

zonkmachine:
fixes #3232 

Inverts the zoom direction in Piano Roll, Automation Editor and Song Editor.


BaraMGB:
:1+

PR #3278, opened 2017-01-20
Added support for official wine-devel and wine-staging packages.

### LINKED ISSUE #3245: cmake doesn't find non-stable Wine on Ubuntu
I'm using Kubuntu 16.10 64-bit and can compile with VST support without problems using wine-stable 1.8 from Ubuntu repositories. It seems not possible to compile when Wine is either:
- wine-development 1.9 from Ubuntu repositories
- wine-devel 2.0 from Wine repositories

In both cases I installed 64 bit and also 32 bit packages.

I removed CMakeCache.txt before running cmake. `cmake ..` output:

`-- Could NOT find Wine (missing:  WINE_LIBRARIES WINE_INCLUDE_DIRS)`

Wine development files are located at `/opt/wine-devel/include/wine/windows`.
Wine libraries are located at `/opt/wine-devel/lib` and `/opt/wine-devel/lib64`.

karmux:
Fixes #3245.


plugins/vst_base/CMakeLists.txt
```diff
@@ -39,23 +39,43 @@ IF(LMMS_HOST_X86_64)
...
 		SET(EXTRA_FLAGS ${EXTRA_FLAGS} -nodefaultlibs /usr/lib/i386-linux-gnu/wine/libwinecrt0.a -luser32 -lkernel32 -lgdi32)
 	ENDIF()
-	#The following check works on Fedora systems 
-	IF("${WINEBUILD_OUTPUT}" MATCHES ".*lib64/wine/libwinecrt0.a.*")
+	# The following check works on Fedora systems
+	IF("${WINEBUILD_OUTPUT}" MATCHES "/usr/lib/lib64/wine/libwinecrt0.a.*")
 		SET(EXTRA_FLAGS ${EXTRA_FLAGS} -nodefaultlibs /usr/lib/i386/wine/libwinecrt0.a -luser32 -lkernel32 -lgdi32)
 	ENDIF()
+	# Development
+	IF("${WINEBUILD_OUTPUT}" MATCHES "/opt/wine-devel/lib64/wine/libwinecrt0.a.*")
```
jasp00: I question this workround for a recent system, specially for a package from WineHQ. There is no way currently to mix 64-bit and 32-bit objects, so this scenario is not possible.

plugins/vst_base/CMakeLists.txt
```diff
@@ -39,23 +39,43 @@ IF(LMMS_HOST_X86_64)
...
+	COMMAND ${WINE_CXX}
+	ARGS "-I\"${CMAKE_BINARY_DIR}\""
+		"-I\"${CMAKE_SOURCE_DIR}/include\""
+		"-I\"${CMAKE_INSTALL_PREFIX}/include/wine/windows\""
+		"-I\"${CMAKE_INSTALL_PREFIX}/include\""
+		-I/usr/include/wine/windows
+		-I/opt/wine-devel/include
+		-L/opt/wine-devel/lib
+		-I/opt/wine-staging/include
+		-I/opt/wine-staging/lib
```
jasp00: This breaks if you have multiple versions of Wine. You should include directories depending on the detected configuration. On the other hand, the expected way is to configure with `WINE_CXX_FLAGS`.

Do you mean `-L/opt/wine-staging/lib`?

plugins/vst_base/CMakeLists.txt
karmux: I'm confused by this comment. What I did is just added a 2 paths to look for libwinecrt0.a.
In addition to /usr/lib/x86_64-linux-gnu/wine/ and /usr/lib/lib64/wine/ it now looks also into /opt/wine-devel/lib64/wine/ and /opt/wine-staging/lib64/wine/. I didn't introduce any new mixing of 32bit and 64bit.

plugins/vst_base/CMakeLists.txt
karmux: Yes, it's a typo, should be -L.

plugins/vst_base/CMakeLists.txt
jasp00: This workaround is currently here because some people do not know they have to install 32-bit Wine and do not understand the linker error. If you install the package from WineHQ, you have 64-bit and 32-bit Wine, so you cannot get this error because 32-bit is selected. Did you actually run into this problem?

plugins/vst_base/CMakeLists.txt
tresf: I agree with @jasp00 here... isn't there a way to assume this in `FindWine.cmake` once `FIND_LIBRARY` returns?

plugins/vst_base/CMakeLists.txt
karmux: If I remove those lines, I can't compile with Wine devel or staging.

cmake/modules/FindWine.cmake
```diff
@@ -7,8 +7,8 @@
 #  WINE_DEFINITIONS - Compiler switches required for using wine
 #
 
-FIND_PATH(WINE_INCLUDE_DIR windows/windows.h PATH_SUFFIXES wine)
-FIND_LIBRARY(WINE_LIBRARY NAMES wine PATH_SUFFIXES wine i386-linux-gnu/wine)
+FIND_PATH(WINE_INCLUDE_DIR windows/windows.h PATHS /opt/wine-devel/include /opt/wine-staging/include PATH_SUFFIXES wine)
```
karmux: I've now written this part using CMAKE_PREFIX_PATH.

plugins/vst_base/CMakeLists.txt
karmux: I've made a new changeset that detects if devel or stable is installed and includes directories of only one wine version.

plugins/vst_base/CMakeLists.txt
```diff
@@ -31,6 +31,17 @@ BUILD_PLUGIN(vstbase vst_base.cpp VstPlugin.cpp VstPlugin.h communication.h MOCF
...
+SET(WINE_INCLUDES_AND_LIBS -I/usr/include/wine/windows)
+
+# Development
+IF(EXISTS "/opt/wine-devel/lib64/libwine.so")
+	SET(WINE_INCLUDES_AND_LIBS -I/opt/wine-devel/include -L/opt/wine-devel/lib)
+ENDIF()
+# Staging
+IF(EXISTS "/opt/wine-staging/lib64/libwine.so")
+	SET(WINE_INCLUDES_AND_LIBS -I/opt/wine-staging/include -L/opt/wine-staging/lib)
+ENDIF()
```
jasp00: Overriding again. You should use `WINE_INCLUDE_DIR` and `WINE_LIBRARY`.

karmux: `WINE_INCLUDE_DIR` gets value `/opt/wine-devel/include/wine` and `wine/*.h` headers cannot be found from there. Path needs to be `/opt/wine-devel/include`.

`WINE_LIBRARY` gets value `/opt/wine-devel/lib/libwine.so` and this cannot be linked. Path needs to be `/opt/wine-devel/lib`.

tresf: @karmux @jasp00 can't we calculate it in [`FindWine.cmake`](https://github.com/LMMS/lmms/blob/master/cmake/modules/FindWine.cmake#L11) then and call it something like `WINE_LIBRARY_PATH` ?

> `wine/*.h` headers cannot be found from [`wine/`]

Sounds like a bug with `FindWine.cmake`, no?

jasp00: >`WINE_INCLUDE_DIR` gets value `/opt/wine-devel/include/wine` [...] Path needs to be `/opt/wine-devel/include`.
>
> `WINE_LIBRARY` gets value `/opt/wine-devel/lib/libwine.so` [...] Path needs to be `/opt/wine-devel/lib`.

So? Use `STRING()`.

plugins/vst_base/CMakeLists.txt
```diff
@@ -39,23 +39,42 @@ IF(LMMS_HOST_X86_64)
...
+	# Staging
+	IF("${WINEBUILD_OUTPUT}" MATCHES "/opt/wine-staging/lib64/wine/libwinecrt0.a.*")
+		SET(EXTRA_FLAGS ${EXTRA_FLAGS} -nodefaultlibs /opt/wine-staging/lib/wine/libwinecrt0.a -luser32 -lkernel32 -lgdi32)
+	ENDIF()
 ENDIF(LMMS_HOST_X86_64)
 
 SET(WINE_CXX_FLAGS "" CACHE STRING "Extra flags passed to wineg++")
 
+STRING(REPLACE "include/wine" "include" WINE_INCLUDE_DIR ${WINE_INCLUDE_DIR})
+STRING(REPLACE "lib/libwine.so" "lib" WINE_LIBRARY ${WINE_LIBRARY})
```
jasp00: You should output to new variables.

plugins/vst_base/CMakeLists.txt
```diff
@@ -39,23 +39,42 @@ IF(LMMS_HOST_X86_64)
...
-		OUTPUTS ../RemoteVstPlugin
-		)
+	SOURCE "${CMAKE_CURRENT_SOURCE_DIR}/RemoteVstPlugin.cpp"
+	COMMAND ${WINE_CXX}
+	ARGS "-I\"${CMAKE_BINARY_DIR}\""
+		"-I\"${CMAKE_SOURCE_DIR}/include\""
+		"-I\"${CMAKE_INSTALL_PREFIX}/include/wine/windows\""
+		"-I\"${CMAKE_INSTALL_PREFIX}/include\""
+		"-I\"${WINE_INCLUDE_DIR_REPLACED}\""
+		"-L\"${WINE_LIBRARY_REPLACED}\""
```
jasp00: With `_REPLACED`, it looks like you are using an alternate directory and library. Do these variables name directories? Then you should use something like `WINE_INCLUDE_BASE_DIR` and `WINE_LIBRARY_DIR`.

karmux: You are correct. I'll fix it.

jasp00:
Code looks good. @karmux, did you test Wine under `/usr`? If so, I will merge.

karmux:
@jasp00 it compiles and runs with Wine under `/usr`

PR #3287, opened 2017-01-24
Fix importing MIDI file from command line

### LINKED ISSUE #2562: Import midi file from command line hangs
When I try an import a midi file from the command line it will fail and open an empty project that will hang and if I press play will show time signatures as the top picture. When I force quit however the recovery file will be a correctly imported file ( bottom picture ). The same file will import just fine from the gui.

Example file: [mrjoe_m.mid.zip](https://github.com/LMMS/lmms/files/129271/mrjoe_m.mid.zip) 
`./lmms --import mrjoe_m.mid`

midiimport

zonkmachine:
Fixes #2562 

Call `Engine::updateFramesPerTick()` as suggested by @michaelgregorius

### DIFF
--- a/src/core/ImportFilter.cpp
+++ b/src/core/ImportFilter.cpp
@@ -90,6 +90,8 @@ void ImportFilter::import( const QString & _file_to_import,
 					QMessageBox::Ok,
 					QMessageBox::NoButton );
 	}
+
+	Engine::updateFramesPerTick();
 }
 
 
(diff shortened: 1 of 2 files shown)


jasp00:
This does not look right. If the model is inconsistent, then the fix should be in the model rather than in a GUI object.

zonkmachine:
> then the fix should be in the model rather than in a GUI object.

Right. Should this be done in the MidiImport.cpp instead? I just noticed that we also don't set the tempo for the project and `Song::setTempo();` calls `Engine::updateFramesPerTick()` so setting the tempo will fix this.

zonkmachine:
I'm more off than I thought with this. I'll close this for now since there is development going on with the drumstick branch. Interestingly that branch sees the precise same issues.

Issue #3205, opened 2017-01-03
LMMS does not allow samples larger than 300MB and/or longer than 90 minutes


Umcaruje:
This issue popped up as I was browsing the [Glitch Artists Collective: Tool Time](https://www.facebook.com/groups/GACToolTime/permalink/1696492363975805/) facebook group. One of the users said it was unable to load files that were >100MB, so I rendered out 2 hours of white noise in audacity and tried to load it, and voilà, this message popped up in the terminal: 
```
refusing to load sample files bigger than 100 MB
```

This error is generated in `SampleBuffer.cpp` where the loading breaks if the file is too large: 
https://github.com/LMMS/lmms/blob/919ca8b4d71a79cfb899b0a81c8bcc49cc8d08b3/src/core/SampleBuffer.cpp#L204-L208

Now, [this change was added](https://github.com/LMMS/lmms/commit/a4ba61882854135a292661933c46d1163b667daf) by @tobydox back in '09 to fix a bug - `#2458375`, which I can't find and it's either forever gone or buried on sourceforge somewhere.

Now, I don't want to blindly remove that break without knowing the context of the bug, but 100MB is very limiting when dealing with uncompressed audio, not just for the purpose of databending, but also when remixing longer tracks and importing stems, so input on this is appreciated.

zonkmachine:
Found a cached page here:
[[ 2458375 ] Open large WAV file](http://web.archive.org/web/20090108102928/http://sourceforge.net/tracker/index.php?func=detail&aid=2458375&group_id=105168&atid=640434)

```
Submitted By: Nobody/Anonymous - nobody
Date Submitted: 2008-12-22 11:59
Last Updated By: Item Submitter - Tracker Item Submitted
Date Last Updated: No updates since submission
Number of Comments: 0
Number of Attachments: 0
Category: (?) Crash
Group: (?) Current Release
Assigned To: (?) Nobody/Anonymous
Priority: (?) 5
Status: (?) Open
Resolution: (?) None
Summary: (?) Open large WAV file
Private: (?) No
From the "my computer" browser in LMMS, open a 1.5 hour long WAV file.
After a while, LMMS will crash.
```

That's pretty unspecific. Maybe just beef up the size of the file or set it relative to available ram? Can you make it crash if you remove the limit?

tresf:
@Umcaruje we can bump it to 200MB if you like, but I just tested against a 900MB WAV and there are some usability issues with it.

**Sample Track** (900MB WAV)
[...]

PR #3293, opened 2017-01-25
Some issues with large files

zonkmachine:
In response to two related crashes in SampleBuffer.cpp as [commented in](https://github.com/LMMS/lmms/issues/3205#issuecomment-274969914) #3205
I substituted the command line messages for a message box.
I haven't touched the original 100 MB size limit and have added a 1 hour length check.

----

FIx explained: 
I add a bool, `fileLoadError`, which is set to false, should the file fail on one of the tests. I add a test for file length as a compressed file of small size when decompressed will take a much bigger size so the important fact to know for LMMS is not the file size but the length of the sample. This operation takes a bit of time to carry out compared to the file size test so we wait to see if that one passes and then carry on testing the length of the sample.
If the file is marked fileLoadError == true, the buffer will be written with one sample, value 0, and after removing possible locks a message is shown. The message is only for information and is shown after the locks are removed so the project isn't frozen while waiting for 'OK' to be pressed. The old message was to the command line and of little use to an ordinary user.

Some issues that remain with opening large files:
 - Previewing large files is sluggish. The file tests works however but large file within the limit are still loaded in their entirety. If you open large files you do it from the track container rather than the menu.
 - After failing to load a file on being to large the buffer is zeored and you might want to keep the old file in case of  a fail. In order to fix this you (probably) need to test the file earlier, before calling `SampleBuffer::update()` .

### DIFF
--- a/src/core/SampleBuffer.cpp
+++ b/src/core/SampleBuffer.cpp
@@ -22,7 +22,6 @@
  *
  */
 
-
 #include "SampleBuffer.h"
 
 
@@ -58,6 +57,7 @@
 #include "DrumSynth.h"
 #include "endian_handling.h"
 #include "Engine.h"
+#include "GuiApplication.h"
 #include "interpolation.h"
 #include "Mixer.h"
 #include "templates.h"
@@ -223,7 +223,7 @@ void SampleBuffer::update( bool _keep_settings )
 			}
 		}
 
-		if( ! fileLoadError )
+		if( !fileLoadError )
 		{
 #ifdef LMMS_HAVE_OGGVORBIS
 			// workaround for a bug in libsndfile or our libsndfile decoder
@@ -291,11 +291,19 @@ void SampleBuffer::update( bool _keep_settings )
 
 	if( fileLoadError )
 	{
-		QMessageBox::information( NULL,
-				"Fail to open file",
-				"Audio files are limited to 100 MB "
-				"in size and 1 hour of playing time",
+		QString message = "Audio files are limited to 100 MB "
+				"in size and 1 hour of playing time";
+		if( gui )
+		{
+			QMessageBox::information( NULL,
+				"Fail to open file", message,
 					QMessageBox::Ok );
+		}
+		else
+		{
+			fprintf( stderr, "%s\n", message.toUtf8().constData() );
+			exit( EXIT_FAILURE );
+		}
 	}
 }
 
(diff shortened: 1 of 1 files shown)


zonkmachine:
This is ready for review.

zonkmachine:
I moved the message box to the end as it would interrupt the song playing. First take care of everything and remove lock, then disappoint the user...

Me thinks `SampleBuffer::update( ... )` is trying to do a bit too much. The file checks could be done in a separate function, and in advance of updating it, and then we can avoid writing blank, 1 sample tracks, on loading a new sample that fails. I think it's more useful to keep the old one in case this happens.

zonkmachine:
I've updated the first post with an explanation of the fix. I think this is a reasonable fix for now.

tresf:
> I've updated the first post with an explanation of the fix. I think this is a reasonable fix for now.

You've added a `QMessageBox` to a non-gui class, so I still think we [need to check for a GUI via `if(gui) ...`](https://github.com/LMMS/lmms/blob/master/include/GuiApplication.h#L81) despite the near-impossible likelihood of a command-line render with a 60 minute sample.

zonkmachine:
OK, fixed that. Stumbled upon https://github.com/LMMS/lmms/issues/3338 so I had to backport to 6137fcc5a5a187750017422d183d282fa8d29a37 to test this.


```
$ ./lmms -r ~/lmms/projects/buggar/largeSamples/90min.mmp -o test.wav
...
Loading project...
Audio files are limited to 100 MB in size and 1 hour of playing time
$
```

src/core/SampleBuffer.cpp
```diff
@@ -291,11 +291,19 @@ void SampleBuffer::update( bool _keep_settings )
...
+		if( gui )
+		{
+			QMessageBox::information( NULL,
+				"Fail to open file", message,
 					QMessageBox::Ok );
+		}
+		else
+		{
+			fprintf( stderr, "%s\n", message.toUtf8().constData() );
+			exit( EXIT_FAILURE );
```
PhysSong: @zonkmachine I think the behavior is inconsistent. With GUI, you can load and render project files with large audio files. However, it's impossible with CLI.
I believe it's okay to remove `exit( EXIT_FAILURE );`.

zonkmachine: So, if you make a project containing a sample of allowed size and later edit that sample to become large/longer than the limit, exporting from the command line will abort when loading the project instead of just exporting the other tracks... I did that. :clown_face: 

> I believe it's okay to remove `exit( EXIT_FAILURE );`.

:+1:

PhysSong: Addressed via 1ab1280843793f16f17d709a3bc47b8c2f5d4e30.

Issue #3298, opened 2017-01-26
 Piano Roll Note Resize


binyominzeev:
When trying to resize an existing note on Piano Roll, sometimes it moves the note instead of resizing. According to my experience, as I hover the right end of the note, the mouse cursor changes to resize. But if the mouse pointer is left to the right end, it still moves. If the mouse pointer is more to the right it resizes. Still, it's often wearysome to click exactly to the right side, especially when trying to resize a several notes. I think it would make more sense to resize whenever the mouse cursor show resize, even when it's a bit more to the left.

zonkmachine:
This seem to be true on zoom greater than 100%.

musikBear:
if more than 2/3 of the mouse-icon is inside the note, i can reproduce, but if the mouse is moved in from right, and the note are dragged as the icon change to *size-note*, then RC2.8 on windows resizes and do not move the note

binyominzeev:
I use Ubuntu 15.10 (but tried on 16.04 as well, and it is the same), 1.2.0-rc2 and it keeps on moving the note. What's even worse is when I have two adjacent notes of the same pitch, I am unable to move the mouse outside, unless I move the second note away from the way, like this (screenshot doesn't show the resize cursor, but it was there, in both cases):

![2017-01-26-204102_2304x1024_scrot](https://cloud.githubusercontent.com/assets/4866177/22347350/f2edd160-e407-11e6-9c0a-5145ba5000fc.png)

![2017-01-26-204107_2304x1024_scrot](https://cloud.githubusercontent.com/assets/4866177/22347355/f78ccece-e407-11e6-8dbb-2cdc3d3d13f8.png)

zonkmachine:
It seem to have been an unintentional side effect of 8504633103f665cb69d6732b110d5c8cb40a6c6c

musikBear:
@binyominzeev Ahhh
> I use Ubuntu 15.10 (but tried on 16.04 as well, and it is the same)

So this linux.., sorry i tested windows. I will shut up

PR #3304, opened 2017-01-27
Note resize area width

zonkmachine:
Fixes https://github.com/LMMS/lmms/issues/3298

I've tested it over the whole zoom range and this seem to fix it good.


zonkmachine:
Sorry! Thorough test but systematically wrong. Now there's an issue grabbing
short notes. Going over this again....

zonkmachine:
No. Short notes on lower zoom settings only gets a resize option and this may or may not be an issue but it's in old stable so this was nothing new.

PR #3309, opened 2017-01-30
Set Stk::showWarnings() via LMMS_DEBUG

zonkmachine:
Fixup of #3172

### DIFF
--- a/plugins/stk/mallets/mallets.cpp
+++ b/plugins/stk/mallets/mallets.cpp
@@ -620,8 +620,10 @@ malletsSynth::malletsSynth( const StkFloat _pitch,
 		Stk::setSampleRate( _sample_rate );
 		Stk::setRawwavePath( QDir( ConfigManager::inst()->stkDir() ).absolutePath()
 						.toLatin1().constData() );
+#ifndef LMMS_DEBUG
 		Stk::showWarnings( false );
-	
+#endif
+
 		m_voice = new ModalBar();
 	
 		m_voice->controlChange( 16, _control16 );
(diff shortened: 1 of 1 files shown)


tresf:
@zonkmachine I was going to comment on this before merge... 

We should really consider configuring this debug value into a header to avoid unnecessary pre-compiler tags all over our code.

Perhaps a nice home for it would be here: https://github.com/LMMS/lmms/blob/master/src/lmmsconfig.h.in

tresf:
Sorry... bad example... `LMMS_VERSION` is a better example.  https://github.com/LMMS/lmms/blob/master/src/lmmsversion.h.in.  Note how the logic in the code can [use it directly](https://github.com/LMMS/lmms/blob/master/src/gui/AboutDialog.cpp#L43), rather than changing the precompile behavior.  This means we could do something like:  `Stk::showWarnings(LMMS_DEBUG)` instead. :)

tresf:
Also, side-note... probably overkill to have both `lmmsconfig.h.in` as well as `lmmsversion.h.in`  We should consider consolidating these. `/offtopic`.

zonkmachine:
> This means we could do something like: Stk::showWarnings(LMMS_DEBUG) instead. :)

OK, interesting so...

```
#cmakedefine LMMS_DEBUG 1
#ifndef LMMS_DEBUG
#define LMMS_DEBUG 0
#endif

...

Stk::showWarnings(LMMS_DEBUG);
```

tresf:
@zonkmachine poorly documented as many of the `cmake` features, but I believe `#cmakedefine01 LMMS_DEBUG` will do the trick per https://github.com/adobe/webkit/blob/master/Source/cmakeconfig.h.cmake#L4

Yes, you read that right, `#cmakedefine01` :)

PR #3145, opened 2016-12-06
Fix #2047 ("TripleOscillator: Oscillators are getting out of sync")

michaelgregorius:
Fix for #2047

Change all phase related variables in `Oscillator` from `float` to `double`.
The parameters for the `getSample` methods have not been changed to `double`
as it should suffice to only update the phase variables with `double`
precision.

Issue #1048, opened 2014-07-31
KILLER BUG: Inifinitely Loud Silence (ILS or rather NaN)


unfa:
**INTRODUCTION**
I'm fighting this beast for years now. With recent LMMS releases it has been worse than ever. Almost every project I start gets infected by this weird "virus" and I can't play the song or render it.

This is a very serious problem, as it makes the whole program unusable. No need to mention the frustration it causes.

I feel in need to lead an investigation to find the root of the problem and solve this once and for all. And I need your help.

**MY SETUP**
I am running KX Studio 14.04b 64-bit on a ASUS-K52J laptop with Core i5 processor and 8GB of RAM.
LMMS 1.0.3 from KX Studio repositories.

**WHAT IS IT?**
The bug makes L and/or R channel of LMMS master output mute with huge DC offset.
This "loud silence" has a very special property: it'll mute your whole computer, and no filters can remove it (AFAIK). The problem seems to attack projects that reach a certain level of development (small projects doesn't seem to be affected).

**How it looks like**
![osc](https://cloud.githubusercontent.com/assets/491704/3770781/bab757da-18ee-11e4-9fbe-e8a274369ed9.png)
The oscilloscope should show you a massive DC offset. But sometimes it just looks silent, regardless of the instruments playing and signal visible in the mixer strips.

Sometimes the ILS (Infintely Loud Silence) mutes all mixer channels, sometimes none, but master, sometimes just a few of them, as can be seen below:
![master](https://cloud.githubusercontent.com/assets/491704/3771085/d5aeae3c-18f1-11e4-9492-8bb855689666.png)
![mix](https://cloud.githubusercontent.com/assets/491704/3770990/b4996986-18f0-11e4-8deb-3f563fd0e310.png)

Here's a screenshot form Audacity, viewing a rendered project affected by ILS:
![auda1](https://cloud.githubusercontent.com/assets/491704/3771249/2e5f69ee-18f3-11e4-9e5b-016bc5edb751.png)

**How it sounds like**
Here's a sample from the rendered project seen on the screenshot above:
http://www.freesound.org/people/unfa/sounds/243833/

Here's a video I recorded to explain the problem (I sent it to the mailing list before):
http://youtu.be/YgXWC91DWVc

**FACTS**
1. The problems seems to start after the project is rendered for the first time. Before the first rendering it hardly ever occurs;
2. After rendering affected project to WAV I get an nearly empty file with huge empty space inside (checked with hexdump);
[...]

PR #3313, opened 2017-01-31
Ladspastuff

zonkmachine:
Dyson Compressor

 NaN - Release time 0.0f  = divide with 0 [here](https://github.com/LMMS/lmms/blob/master/plugins/LadspaEffect/swh/dyson_compress_1403.c#L375):
`	float rgainfilter = 1.0f / (release_time * sample_rate);`

In the solution: `port_range_hints[DYSONCOMPRESS_RELEASE_TIME].LowerBound = 0.0000001; `
Maybe use a constant for 0.0000001 ... 'weeLittleFloat' = 0.0000001 ?
Are there any suitable math constants to use? I just put in the smalles number that would give 0 and 1 back in the knobs extreme positions.

Also, I found a generally odd looking line [here](
https://github.com/LMMS/lmms/blob/master/plugins/LadspaEffect/swh/dyson_compress_1403.c#L446:L452)
`if (compressionratio == 0.50f)` should probably be `<` or `>`
I thought it could potentionally produce a glitch if you sweep the compression knob across 0.5f as it would change to another algorithm for that value only but I haven't managed to produce anything buggy because of it. I haven't dived into the math of it yet.

---

LADSPA debugging code

Put general debugging code for LADSPA so we can stop those NaN's and whatever may be hiding in there. In response to #1048 and discussion [here](https://github.com/LMMS/lmms/pull/3145#issuecomment-267848275) .
@tresf :) I've looked at cmakedefine01 but I don't think any of that applies to this case, right?

### DIFF
--- a/plugins/LadspaEffect/LadspaEffect.cpp
+++ b/plugins/LadspaEffect/LadspaEffect.cpp
@@ -24,8 +24,10 @@
  */
 
 
+#include <QDebug>
 #include <QMessageBox>
 
+#include "debug.h"
 #include "LadspaEffect.h"
 #include "DataFile.h"
 #include "AudioDevice.h"
@@ -241,6 +243,15 @@ bool LadspaEffect::processAudioBuffer( sampleFrame * _buf,
 					for( fpp_t frame = 0; 
 						frame < frames; ++frame )
 					{
+						if (	LADSPA_DEBUG &&
+							( isnan( pp->buffer[frame] ) ||
+							isinf( pp->buffer[frame] ) ) )
+						{
+							qDebug() << "name: " << m_descriptor->Label;
+							assert( ! isnan( pp->buffer[frame] ) );
+							assert( ! isinf( pp->buffer[frame] ) );
+						}
+
 						_buf[frame][channel] = d * _buf[frame][channel] + w * pp->buffer[frame];
 						out_sum += _buf[frame][channel] * _buf[frame][channel];
 					}
--- a/include/debug.h
+++ b/include/debug.h
@@ -32,8 +32,10 @@
 // additional range-checkings) should be compiled
 
 #ifdef LMMS_DEBUG
+#define LADSPA_DEBUG TRUE
 	#include <assert.h>
 #else
+#define LADSPA_DEBUG FALSE
 	#define assert(x) ((void)(x))
 #endif
 #include <cstdio>
--- a/plugins/LadspaEffect/swh/dyson_compress_1403.c
+++ b/plugins/LadspaEffect/swh/dyson_compress_1403.c
@@ -14,6 +14,7 @@
 #define         __USE_ISOC9X    1
 
 #include <math.h>
+#include <float.h>
 
 #include "ladspa.h"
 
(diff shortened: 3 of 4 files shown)


tresf:
> Are there any suitable math constants to use? I just put in the smalles number that would give 0 and 1 back in the knobs extreme positions.

`FLT_MIN`, I believe.

> LADSPA debugging code... Put general debugging code for LADSPA so we can stop those NaN's and whatever may be hiding in there. In response to #1048 and discussion here .  @tresf :) I've looked at cmakedefine01 but I don't think any of that applies to this case, right?

Hmm... http://stackoverflow.com/a/5354352/3196753

According to this, we can pull out preprocessor flags here too if we do it correct.  Not my area of expertise, but I'm sure some others will know.

zonkmachine:
> Hmm... http://stackoverflow.com/a/5354352/3196753
> 
> According to this, we can pull out preprocessor flags here too if we do it correct. Not my area of expertise, but I'm sure some others will know.

`#define assert(ignore) ((void)0)`
There is [debug.h](https://github.com/LMMS/lmms/blob/48cc3bb5ef85bac8cc7c01b11dd3139c547a0677/include/debug.h) .

Umcaruje:
@zonkmachine do you need some debugging work? I have some projects that are basically NaN farms :smiley:

zonkmachine:
> do you need some debugging work? I have some projects that are basically NaN farms

Yes! zip and send! :tractor:

plugins/LadspaEffect/LadspaEffect.cpp
```diff
@@ -241,6 +243,14 @@ bool LadspaEffect::processAudioBuffer( sampleFrame * _buf,
 					for( fpp_t frame = 0; 
 						frame < frames; ++frame )
 					{
+						if (	isnan( pp->buffer[frame] ) ||
```
zonkmachine: @tresf I need your cmakedefine01 here to go `if (	LMMS_DEBUG ||` or maybe an `LADSPA_DEBUG` that is set by the former. If I try to cmakedefine01 LMMS_DEBUG that isn't working because it's already defined.

Umcaruje:
There, have an ILS farm. It reliably looses sound after bar 9 on both windows and linux.
[Syntask x Umcaruje x Greippi - Musaa.zip](https://github.com/LMMS/lmms/files/745973/Syntask.x.Umcaruje.x.Greippi.-.Musaa.zip)

If I stumble upon any other reliably problematic projects, I'll keep you posted ;)

tresf:
> @tresf I need your cmakedefine01 here to go if ( LMMS_DEBUG || or maybe an LADSPA_DEBUG that is set by the former. If I try to cmakedefine01 LMMS_DEBUG that isn't working because it's already defined.

Can you expand on this statement?  I can't find much about `LMMS_DEBUG` in general.  Where is it defined already?

zonkmachine:
> There, have an ILS farm. It reliably looses sound after bar 9 on both windows and linux.

```
Label:  Limiter 
ID: 0x8521
Name:  Calf Limiter LADSPA 
 ... Assertion `! isnan( pp->buffer[frame] )' failed.
```

If I nudge the volume on the bitinvader and the cowbell, which are the two tracks acting up like this, before the sounds are played, I can get the song to play all the way to the end. I wonder if that means it could be our bug? That the plug-in isn't set up properly... I think the next step in this case is that you should try it with the latest calf plug-ins and see if there is any difference.
Upstream fix that could be related: https://github.com/calf-studio-gear/calf/commit/ec8afdda9926b45d0f62c9edc4437c4ff40c5bd6

Umcaruje:
>I think the next step in this case is that you should try it with the latest calf plug-ins and see if there is any difference.

Does Calf even support their LADSPA versions, or have they switched to LV2 completely? If they still work on the LADSPAs, we could just pull in the changes, right?

Umcaruje:
Looks like they removed support for ladspa since version `0.0.19-rc3`

https://github.com/calf-studio-gear/calf/commit/460642bfae534679b4094e6887714828c90d8172

zonkmachine:
> Looks like they removed support for ladspa since version 0.0.19-rc3

And our calf changelog reads `Version 0.0.60.0 (unreleased)`
So I guess we just patch what we can. I'll try and see if there are any fixes committed in the Linux repositories that we can use.

PhysSong:
Wrong clicks:(

zonkmachine:
The debug part of this PR, 28b30d6d42eaaa0d9c0ae3d5d25b82568f505b46, is superceded by https://github.com/LMMS/lmms/pull/3687 . Closing.
The other issues will be addressed in a separate PR or upstream.

PR #3322, opened 2017-02-04
Avoid crash on gui when rendering from command line and no write access

### LINKED ISSUE #2902: trying to show a QMessageBox when no gui is present in command line
when using command line for rendering if the output or config file aren't writable, it can't create the QWidget showing the error message, because there is no gui
What do you think of including `GuiApplication.h` on the respective classes, and `if (gui)` shows the QMessageBox, or else a console message?

zonkmachine:
Fixes #2902

Something like this?

### DIFF
--- a/src/core/audio/AudioFileDevice.cpp
+++ b/src/core/audio/AudioFileDevice.cpp
@@ -27,6 +27,7 @@
 
 #include "AudioFileDevice.h"
 #include "ExportProjectDialog.h"
+#include "GuiApplication.h"
 
 
 AudioFileDevice::AudioFileDevice( const sample_rate_t _sample_rate,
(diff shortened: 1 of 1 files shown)


zonkmachine:
The other option is to fix it in [AudioFileDevice.cpp](https://github.com/LMMS/lmms/blob/master/src/core/audio/AudioFileDevice.cpp#L51:L64)
Add "GuiApplication.h" as suggested in #2902 printf something and EXIT_FAILURE when no gui. Both works just fine.

zonkmachine:
Pushed method two on a separate commit for comparison. I vote no. 2

src/core/audio/AudioFileDevice.cpp
```diff
@@ -50,17 +51,26 @@ AudioFileDevice::AudioFileDevice( const sample_rate_t _sample_rate,
...
 	{
-		QMessageBox::critical( NULL,
-			ExportProjectDialog::tr( "Could not open file" ),
-			ExportProjectDialog::tr( "Could not open file %1 "
+		QString title, message;
+		title = ExportProjectDialog::tr( "Could not open file" );
+		message = ExportProjectDialog::tr( "Could not open file %1 "
 						"for writing.\nPlease make "
 						"sure you have write-"
 						"permission to the file and "
```
zonkmachine: `write-permission`?

-> `write permission`

zonkmachine:
Settled for no.2 . Done... :steam_locomotive:

src/core/audio/AudioFileDevice.cpp
```diff
@@ -50,17 +51,27 @@ AudioFileDevice::AudioFileDevice( const sample_rate_t _sample_rate,
...
+
+		if( gui )
+		{
+			QMessageBox::critical( NULL, title, message,
+						QMessageBox::Ok,
+						QMessageBox::NoButton );
+		}
+		else
+		{
+			printf( "%s\n", message.toUtf8().constData() );
```
jasp00: `fprintf( stderr, "%s\n",`

zonkmachine: Fixed!

zonkmachine:
Thanks! Merge...

Issue #2636, opened 2016-03-05
Determine how the Beat and Bassline Editor should work


RebeccaDeField:
From what people have been saying, it seems like the current BB editor is not ideal, and it's more from a technical standpoint than just aesthetic.

> What would the Beat and Bassline editor ideally look/work like? Maybe after we have that established, we can then come up with a solution based on those standards. ~ Rebecca

> We're still trying to determine what it should work like... I think ideally, the patterns would all grow and shrink together (something I'm trying to tackle right now). Then perhaps a way to prevent squishing the steps smaller than their natural size. I'll keep you posted as to my progress. This code is all over the place, so this is taking much more time than I had anticipated. ~ @tresf

tresf:
Related: #2443

tresf:
From a step-size perspective, I think all are in agreement that the steps need to be a predicable size, so that the pixmaps can be rendered correctly.

From a UI perspective there are some options... 
 * Make BB editor more like the song-editor in the fact that it can have variable length patterns and all patterns are positioned and looped based on their location in the BB timeline.
  * Currently the BB editor stretches patterns to the BB editor's width
  * Preventing patterns from stretching may mean we need a scroll area like the piano roll
  * Scroll area suggest we have a zoom factor as well
 * Keep BB editor fixed 
  * This is likely to be faster to implement from a code perspective
  * All patterns get adjusted to the BB editor's max pattern width
  * Steps never appear under a certain size
  * BB editor can quickly become wider than the average screen if long BB patterns are chosen

I had tried option 2 above but was having a hard time keeping the track widths in sync with each-other.  There seemed to be a missing way to quick expand/contract a pattern.

liftarn:
On feature that would be nice to have is syncopated beats. Instead of just 1234 you would have 1a2a3a4a. You can do it as it is by either double the bpm and use half of them or going into the piano roll view, but neither of them is very user friendly.

musikBear:
@liftarn you mean *sving* right?
I hope #2255 will be in consideration. Building continuous sample in succession would open complete new powerful ways of using lmms.

liftarn:
@musikBear Sving is not something I'm familiar with. Swing I know what it is, but that is changing the timing of the play. I'm talking about adding beats at the positions between the 4/4 beats. So instead of 1 2 3 4 you have 1 and 2 and 3 and 4 and.

RebeccaDeField:
[...]

PR #3325, opened 2017-02-05
Update Pattern.cpp

doyousketch2:
So BB Editor changes with time signature.

If you select 2/2 time, you get 2/2 time in your BB Editor.
If you select 3/4 time, then that's what you get.

Doesn't change the Add More button, or anything like that.
It still doubles 'em so that should be fine.

Goes along with [2636](https://github.com/LMMS/lmms/issues/2636)

### DIFF
--- a/src/tracks/Pattern.cpp
+++ b/src/tracks/Pattern.cpp
@@ -309,6 +309,7 @@ void Pattern::rearrangeAllNotes()
 
 
 
+
 void Pattern::clearNotes()
 {
 	instrumentTrack()->lock();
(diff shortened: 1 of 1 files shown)


doyousketch2:
I hate git...  Realized after I posted the code, it needs to update the length of the pattern to fit on the main pattern page.  I like the way it looks, but I'll need to dig into the code some more to see how that's done.

tresf:
> I hate git...

If you sync your branch up with master and push, it auto-closes the PRs.  No worries, just open a new one when ready. 👍

Issue #3311, opened 2017-01-31
Add Long Label Cutoffs To Tracks In The Song Editor and BB Editor


miketurn:
(This was brought up in a separate thread)
Add Long Label Cutoffs To Tracks In The Song Editor and BB Editor.
When labels get too long add the "..." to the end of them, like the Mixer labels have.

PR #3328, opened 2017-02-06
elide the text on track button if it's too lon...

BaraMGB:
closes https://github.com/LMMS/lmms/issues/3311


tresf:
@BaraMGB this wasn't solely added by you, but I see calls to `setText` in the paint event, which seems the wrong place to do this.

Previous to your commit, the `renameFinished` event fired `setText`.  Your commit removed that which made me take a closer look.

For example... `setText(" ")` called [over-and-over every paint event](https://github.com/BaraMGB/lmms/blob/b101f090f549c0e14de8ec1d6ee9581090402de3/src/gui/widgets/TrackLabelButton.cpp#L197) when `compacttrackbuttons` is true seems unecessary.

I assume the elide is easiest this way due to the possibility of the button changing size (which is a good thing to capture) but to be proper, that should be tackled with an `eventFilter` emits `QEvent::Resize`.

If we decide to merge this as-is, we should have a follow-up task to handle resizes and redraws more gracefully.

BaraMGB:
acknowledged. I work on this.

src/gui/widgets/TrackLabelButton.cpp
```diff
@@ -64,8 +64,17 @@ TrackLabelButton::TrackLabelButton( TrackView * _tv, QWidget * _parent ) :
 	}
 
 	setIconSize( QSize( 24, 24 ) );
-	setText( " " );
+	if( ConfigManager::inst()->value( "ui", "compacttrackbuttons" ).toInt() )
```
tresf: @BaraMGB how well does compact buttons work on the constructor?  Previously it would fire a live-redraw but now I'd expect it to only take effect after the buttons are destroyed and recreated.  May be better suited for `update()` instead.

BaraMGB: Okay, we change the elide width dynamically to the button width. And call the nameChanged slot on resize. 

If the buttons width smaller the a certain value we set the text to "".

BaraMGB:
done @tresf

zonkmachine:
Tests fine. `Linuxmint/Mate desktop/Qt4` :+1:

BaraMGB:
Wanna wait for @tresf

tresf:
My only remaining concern was in regards to `QToolButton::paintEvent( _pe );` being called on each `paintEvent`.

BaraMGB:
`QToolButton::paintEvent( _pe );` seems necessary for painting the rest of the button.

tresf:
> QToolButton::paintEvent( _pe ); seems necessary for painting the rest of the button.

Have you tried simply calling [`QWidget::update()`](https://github.com/LMMS/lmms/blob/9e85d7c66ed28254039f986adb2799aa386740b6/plugins/GigPlayer/GigPlayer.cpp#L1042-L1044) in the rename function?  I'm only basing my criticism on the rest of the codebase, please feel free to prove me wrong. 👍

jasp00:
>My only remaining concern was in regards to `QToolButton::paintEvent( _pe );` being called on each `paintEvent`.

`TrackLabelButton` inherits from `QToolButton`. It is normal to call the parent method.

BaraMGB:
Can we merge this? @tresf

PR #3331, opened 2017-02-06
Auto-save - Fix double negations and some cleanup

zonkmachine:
Some auto-save fixes

- Cleaned up the double negations
- Simplified the evaluation in autoSave()
- Some minor changes


zonkmachine:
OK. Thanks for the review!

PR #3333, opened 2017-02-07
add some minimum sizies for some windows

BaraMGB:
- adds a minimum height for **controller rack window**
- adds minimum size for **about dialog** and **export dialog** (setup dialog)

### DIFF
--- a/src/gui/dialogs/about_dialog.ui
+++ b/src/gui/dialogs/about_dialog.ui
@@ -10,6 +10,12 @@
     <height>357</height>
    </rect>
   </property>
+  <property name="minimumSize">
+   <size>
+    <width>558</width>
+    <height>357</height>
+   </size>
+  </property>
   <property name="windowTitle">
    <string>About LMMS</string>
   </property>
--- a/src/gui/dialogs/export_project.ui
+++ b/src/gui/dialogs/export_project.ui
@@ -10,6 +10,12 @@
     <height>412</height>
    </rect>
   </property>
+  <property name="minimumSize">
+   <size>
+    <width>519</width>
+    <height>412</height>
+   </size>
+  </property>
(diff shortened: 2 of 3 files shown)


zonkmachine:
Yes. Those are good sizes. Maybe the size of at least the **export dialog** should be fixed as the fields looks a bit odd when you stretch it.

BaraMGB:
bildschirmfoto vom 2017-02-07 14-51-53

I fear on different languages it look different.

zonkmachine:
> I fear on different languages it look different.

Right. Forgot about those.

src/gui/dialogs/about_dialog.ui
```diff
@@ -10,6 +10,12 @@
     <height>357</height>
    </rect>
   </property>
+  <property name="minimumSize">
+   <size>
+    <width>558</width>
```
Umcaruje: the numbers seem odd to me, why not go with something like `560x360`?

src/gui/dialogs/export_project.ui
```diff
@@ -10,6 +10,12 @@
     <height>412</height>
    </rect>
   </property>
+  <property name="minimumSize">
+   <size>
+    <width>519</width>
```
Umcaruje: same here, why not `520x410`

tresf: To me it would be more odd (statistical anomaly) that a round number were coincidentally the right number.

BaraMGB: I took the numbers which are used as initial sizes for that windows.

Umcaruje: Ok, no need to change them then, it's just that I found them odd at a first glance ;)

Issue #3299, opened 2017-01-26
Piano Roll Window Opens With Narrow Width


miketurn:
Not sure if this is on purpose but when you open your first instance of the piano roll the window opens up with a very narrow default width, causing you to have to stretch it out to see everything.

BaraMGB:
:1+

mikobuntu:
Agreed . I think the default size should show all tools on the toolbar at least.

tresf:
This was partially fixed in https://github.com/LMMS/lmms/issues/2288; fixed when loading a saved project but never addressed an empty project file.

mikobuntu:
It seems if we use the same width as the Song_Editor window, it allows all tools and drop-down menus to be shown and also opening and closing the Piano_Roll looks aesthetically better ( as though someone cared  enough to align the Song_Editor and Piano_roll's dimensions at least )

PR #3334, opened 2017-02-07
set new initial width of piano roll editor

BaraMGB:
fixes https://github.com/LMMS/lmms/issues/3299

set the size equal to automation editor window.

### DIFF
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -81,6 +81,7 @@ typedef AutomationPattern::timeMap timeMap;
 
 
 // some constants...
+const int INITIAL_PIANOROLL_WIDTH = 860;
 const int INITIAL_PIANOROLL_HEIGHT = 480;
 
 const int SCROLLBAR_SIZE = 12;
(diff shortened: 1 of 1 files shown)


tresf:
The toolbar method seemed a bit more dynamic.  Is there a reason why this began failing (collapsible toolbar perhaps?)

mikobuntu:
@tres I'm presuming the collapsible toolbar method will still work after this PR gets implemented as it seems to only initialize a set width on opening ? , If so I think at least the user will still get these option if they decide to shrink the Piano_Roll window. 
I just find that I have to stretch out the window every single time I use LMMS which can be annoying ( less clicks/dragging window size means more music being created ) .

BaraMGB:
>The toolbar method seemed a bit more dynamic. Is there a reason why this began failing (collapsible toolbar perhaps?)

@tresf yes, I guess thats the reason. The toolbar is broken into parts. I don't quite get it. Seems it uses the standard Qt Toolbar. But I don't see how to get a sizehint() from it.

tresf:
@mikobuntu I think you misread my concern.  My concern is why this calculated value became hard-coded, not so much about the toolbar behavior.

@BaraMGB my concern here is that as we start to write more code around high DPI displays, we need to move away from magic width and height parameters.

Since this fix is a hard requirement of 1.2, I'll merge as-is, but we need to move away from magic numbers as a project goal.

Issue #3338, opened 2017-02-09
Project with Sample Track crash on export from command line


zonkmachine:
Projects containing Sample Tracks fails on export from command line since the big update in 43a77a0219fc95fd017ae7af99d98a7cbb04bdd2
@BaraMGB 

```
Loading project...
QObject::connect: Cannot connect (null)::positionMarkerMoved() to SampleTCO::playbackPositionChanged()
Segmentation fault (core dumped)

...

Core was generated by `./lmms -r /home/zonkmachine/lmms/projects/buggar/largeSamples/11min.mmp -o test'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x000000000054fb5c in GuiApplication::songEditor (this=0x0) at /home/zonkmachine/builds/lmms/lmms/include/GuiApplication.h:54
54		SongEditorWindow* songEditor() { return m_songEditor; }
(gdb) bt
#0  0x000000000054fb5c in GuiApplication::songEditor (this=0x0) at /home/zonkmachine/builds/lmms/lmms/include/GuiApplication.h:54
#1  0x000000000061fa56 in SampleTCO::SampleTCO (this=0x7f2b2db97ee0, _track=0x7f2b2db97ae0)
    at /home/zonkmachine/builds/lmms/lmms/src/tracks/SampleTrack.cpp:87
#2  0x0000000000622469 in SampleTrack::createTCO (this=0x7f2b2db97ae0) at /home/zonkmachine/builds/lmms/lmms/src/tracks/SampleTrack.cpp:677
#3  0x00000000005590b3 in Track::loadSettings (this=0x7f2b2db97ae0, element=...) at /home/zonkmachine/builds/lmms/lmms/src/core/Track.cpp:2146
#4  0x0000000000549d24 in SerializingObject::restoreState (this=0x7f2b2db97b00, element=...)
    at /home/zonkmachine/builds/lmms/lmms/src/core/SerializingObject.cpp:70
#5  0x000000000051b1b2 in JournallingObject::restoreState (this=0x7f2b2db97b00, _this=...)
    at /home/zonkmachine/builds/lmms/lmms/src/core/JournallingObject.cpp:92
#6  0x000000000055852c in Track::create (element=..., tc=0x274eb40) at /home/zonkmachine/builds/lmms/lmms/src/core/Track.cpp:2009
[...]

PR #3341, opened 2017-02-10
fix crash on export per command line with sample track

BaraMGB:
this should fix https://github.com/LMMS/lmms/issues/3338

@zonkmachine can you look at this?

### DIFF
--- a/src/tracks/SampleTrack.cpp
+++ b/src/tracks/SampleTrack.cpp
@@ -74,7 +74,10 @@ SampleTCO::SampleTCO( Track * _track ) :
 
 	//care about positionmarker
 	TimeLineWidget * timeLine = Engine::getSong()->getPlayPos( Engine::getSong()->Mode_PlaySong ).m_timeLine;
-	connect( timeLine, SIGNAL( positionMarkerMoved() ), this, SLOT( playbackPositionChanged() ) );
+	if( timeLine )
+	{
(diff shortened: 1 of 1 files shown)


zonkmachine:
Now the file loads and exports fine but never hands back to main() . The Render Manger is stuck reporting 100% .

zonkmachine:
Sometimes it renders just fine. I think I've seen this before. Just once in a while a render will fail to complete. It may be that you've poked another bug a bit. I'll continue testing this to see if I can find a pattern in it.

BaraMGB:
Does it depends on the project?

zonkmachine:
> Does it depends on the project?

It seem to depend on the length of the samples. I'm testing this with noise samples of various length and with a 15 second sample in the project there is about 50% chance of the renderer getting stuck after export.

zonkmachine:
Sorry. I went back 50 commits and this bug is there also. It's hard to see a real pattern in it though other than the sample track being involved. This commit fixes the issue reported and should be merged. I can probe into the hanging renderer a bit more and write a new issue for it.

zonkmachine:
The hanging render on longer Sample Tracks is an SDL thing.
I'm not done bisecting this but it may have happened around these commits 8497d39eec78f3f23f0581898e18868025d5070e e55ec80c3d956dea8674afc0755bc87831c64ce6

zonkmachine:
New issue opened for the SDL issues here: https://github.com/LMMS/lmms/issues/3344
I suggest you merge this.

BaraMGB:
Okay. Thanks for the investigation

Issue #3339, opened 2017-02-09
crossover  equalizer: distortion from dials


musikBear:
RC.2.8 win 32
This image shows random notes with automation of the 1/2-dial in crossover equalizer 
![crossoverbuttons](https://cloud.githubusercontent.com/assets/6662066/22795120/cbabdf72-eef5-11e6-988c-82fc7c85f4f7.JPG)
Listen to this recording
https://soundcloud.com/musikbear/crossoverdials
The distorting are where the slope of the automation-track makes the dial turn
The clean sound is on the plateaus
Can that be fixed?

michaelgregorius:
Should be fixed with #3343.

zonkmachine:
Closed in https://github.com/LMMS/lmms/pull/3343

PR #3343, opened 2017-02-10
Fix issue #3339 by not clearing filter history on control change

michaelgregorius:
Do not clear the filter histories when the crossover control has changed,
e.g. via automation.

Add a new method CrossoverEQEffect::clearFilterHistories that's called
whenever the filter histories need to be cleared, e.g. after loading a
crossover EQ. It would be beneficial to also call this method when the
effect is enabled again after being disabled but it seems there is no
was to find out that this event has happened. One could implement it in
the process method by storing the current state in a member and
comparing it to the state at the time of the last process call but this
is something that should be provided by the framework.

### DIFF
--- a/plugins/CrossoverEQ/CrossoverEQ.cpp
+++ b/plugins/CrossoverEQ/CrossoverEQ.cpp
@@ -95,23 +95,17 @@ bool CrossoverEQEffect::processAudioBuffer( sampleFrame* buf, const fpp_t frames
 	if( m_needsUpdate || m_controls.m_xover12.isValueChanged() )
 	{
 		m_lp1.setLowpass( m_controls.m_xover12.value() );
-		m_lp1.clearHistory();
 		m_hp2.setHighpass( m_controls.m_xover12.value() );
-		m_hp2.clearHistory();
 	}
 	if( m_needsUpdate || m_controls.m_xover23.isValueChanged() )
 	{
 		m_lp2.setLowpass( m_controls.m_xover23.value() );
-		m_lp2.clearHistory();
 		m_hp3.setHighpass( m_controls.m_xover23.value() );
-		m_hp3.clearHistory();
 	}
 	if( m_needsUpdate || m_controls.m_xover34.isValueChanged() )
 	{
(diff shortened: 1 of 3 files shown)


zonkmachine:
Testing this with an Automation Track. It solves the problem when I use Linear and Cubic Hermite progression but creates new glitches on Discrete progression.

zonkmachine:
Interestingly it's only on a positive flank, when the automation jumps from a lower to a higher value, that I get a glitch.

michaelgregorius:
I think that the pops are created because the parameter changes are not smoothed at all. So instead of interpreting the incoming parameter value directly they'd need to go through a simple 1 pole low pass like the one described here:
http://www.musicdsp.org/archive.php?classid=3#257

However, that might be a more general problem and is something for another issue. So I propose to merge my changes as a first improvement and perhaps add a new issue that describes the problem and that might also be used to investigate whether similar problems show up in other areas as well.

zonkmachine:
> So I propose to merge my changes as a first improvement and perhaps add a new issue that describes the problem and that might also be used to investigate whether similar problems show up in other areas as well.

Sounds good to me. :+1: 

Any guess as to why I only see this on a rising signal? I've seen pops before but have never noticed this behaviour. It could simply be that I just haven't thought about it but since the behaviour is systematic it could be that we do something wrong here. It works the same way with an LFO square wave.

zonkmachine:
Merge?

michaelgregorius:
@zonkmachine, I'd say let's merge it. It's not perfect but already better than before.

zonkmachine:
> However, that might be a more general problem and is something for another issue. So I propose to merge my changes as a first improvement and perhaps add a new issue that describes the problem and that might also be used to investigate whether similar problems show up in other areas as well.

Will you open this new issue? Close https://github.com/LMMS/lmms/issues/3339 ?

zonkmachine:
> Will you open this new issue? Close #3339 ?

Fixed! New issue here: https://github.com/LMMS/lmms/issues/3421

michaelgregorius:
Thanks for the merge and the creation of the new issue, @zonkmachine!

Issue #3345, opened 2017-02-12
Separator shouldn't display at bottom of "File" menu on Mac OS X


follower:
This seems to be a regression related to the move from Qt4 to Qt5.

Notice there is a separator at the bottom of the "File" menu which is unnecessary and shouldn't be there:
![lmms-menu-separator-issue-qt5](https://cloud.githubusercontent.com/assets/189962/22863154/250bb8d8-f1a0-11e6-9d93-f3b2273fe01d.png)

Version information (this is on Mac OS X 10.8.x):
![lmms-menu-separator-issue-qt5-version-info](https://cloud.githubusercontent.com/assets/189962/22863157/2519e994-f1a0-11e6-967d-474ac9aba158.png)

For comparison here is the same menu with Qt4:
![lmms-menu-separator-issue-qt4](https://cloud.githubusercontent.com/assets/189962/22863155/250c2ca0-f1a0-11e6-8ff4-b4c98b3c99ed.png)

And version information for Qt4 build on same system:
![lmms-menu-separator-issue-qt4-version-info](https://cloud.githubusercontent.com/assets/189962/22863156/251787b2-f1a0-11e6-862c-5bbe07f932ee.png)

The reason this occurs is because Qt automatically moves the "Quit" menu item from the file menu to the "~~Apple~~ Application Menu" as is usual for the platform. Normally any redundant separator is also removed at the same time.

This appears to be a known issue for Qt: ["[QTBUG-40071] separatorsCollapsible does not work when menu items are moved to the application menu on Mac OS X"](https://bugreports.qt.io/browse/QTBUG-40071).

I've added further detail to the QTBUG entry but have been unable to determine an underlying cause so far.

Before I realised it was an issue with Qt I developed a simple workaround which results in:

![lmms-menu-separator-issue-qt5-workaround](https://cloud.githubusercontent.com/assets/189962/22863264/a16d6240-f1a1-11e6-9c9b-39099602f67a.png)

Given that it appears to be a known issue that's not been resolved in nearly 3 years I'd suggest using the workaround in lmms in the meantime.

If desired I can make a PR with the workaround (with further documentation):

```
diff --git a/src/gui/MainWindow.cpp b/src/gui/MainWindow.cpp
index 4a5bacb..aa943c8 100644
--- a/src/gui/MainWindow.cpp
+++ b/src/gui/MainWindow.cpp
@@ -315,7 +315,9 @@ void MainWindow::finalize()
[...]

PR #3350, opened 2017-02-13
Workaround Qt5 bug that fails to remove separator at end of File/Help menus

follower:
See #3345 for details.

### DIFF
--- a/src/gui/MainWindow.cpp
+++ b/src/gui/MainWindow.cpp
@@ -315,7 +315,16 @@ void MainWindow::finalize()
 					SLOT( exportProjectMidi() ),
 					Qt::CTRL + Qt::Key_M );*/
 
+//
+// Conditional compilation to workaround Qt5 bug that fails to remove
+// separator at end of menu when the Quit menu item is relocated to
+// Application Menu on Mac OS X / macOS. For details see:
+//  * <https://github.com/LMMS/lmms/issues/3345>
+//  * <https://bugreports.qt.io/browse/QTBUG-40071>
+//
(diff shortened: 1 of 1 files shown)


follower:
Note: I've only tested this fix on Qt5 + Mac OS X 10.8.x.

tresf:
@follower thanks, your test results are sufficient.

I think your comment could be shorted some.

`Conditional compilation to workaround Qt5 bug` is redundant since you're using precompiler flags and the flag has Qt in the name.  I'd shrink it to simply:

```cpp
// Per https://bugreports.qt.io/browse/QTBUG-40071
```

Or if you prefer to leave an explanation:

```cpp
// remove dangling separator at end of menu per https://bugreports.qt.io/browse/QTBUG-40071
```

Also note that removing the specificity of `Quit` also allows others to copy and past this patch in other areas moving forward without fear of contextual comment typos.

follower:
Thanks for the review.

I suspected I might get that feedback--I've never been accused of being excessively terse. :)

I can make the change as suggested but wanted to mention part of the motivation for the wording was to make it clear that it was the preprocessor conditional that was the workaround not the addition of the separator. I'd normally put the comment on the `#if` line but figured it'd make the line too long in this case.

Will go with:

    // Prevent dangling separator at end of menu per https://bugreports.qt.io/browse/QTBUG-40071

follower:
Should now exclude Qt 5.6+. Tested with Qt 5.5.1.

tresf:
@follower thanks!!!

Issue #3301, opened 2017-01-27
Automation Editor Nodes Don't Always Delete


miketurn:
Not sure if anyone else has experienced this, not a big deal.
When you have the pencil tool selected and you try to use your Right mouse button to delete a node (or point), sometimes they delete fine but other times even though the crosshairs/arrow is in the node circle, if you are a little off center the nodes don't delete.

tresf:
Sounds similar to https://github.com/LMMS/lmms/issues/391.

miketurn:
@tresf 
(Again not a big deal, it can just feel a little odd)
I am not sure, I think in the case of what I am referring to is that although the node is a round circle it is only programmed to delete it from its center point, so if you try to delete it off center of that node with a it does not register it.  Yet if you hold down your RIGHT mouse button and drag across, the second the purple crosshairs touch the node at its side it gets deleted.

In testing this I also found a minor minor bug which I will create a new post for.

zonkmachine:
This has greatly improved since 1.2.0-rc4 but is still a bit insensitive. In order for this to be fixed completely you could do something like a search to find the closest automation point and delete that one.

tecknixia:
I believe I may have fixed this problem in this PR:
#5232

The cursor now turns into a hand on the edges of the circle, and as soon as it does, when you right click it will delete.

I can't say the same for when you're holding and dragging the mouse, the same while loop and if statement is being used, but for some reason during a mouse move, it won't detect until the center (just on the left side).  Although, when you're holding and moving the mouse, it's real easy to just swipe the center, and doesn't seem to hinder the user experience.

PR #3352, opened 2017-02-13
Some automation pattern fixes

### LINKED ISSUE #391: Automation "Locked Out" Bug, Affects Exporting
**Edit:** This bug is with the eraser behavior and quantization.  If automation events were created with quantization of 64 and then later are edited with quantization 16, it does not properly delete the events in between.  This behavior is not present in 0.4.15.

In addition, this bug is also to address an unknown cause where the automation does not get honored on export (which arguable should be its own bug, once I can reproduce this predictably).

---
## Original Post

I have a bug to report regarding the new automation editor.

It seems there are scenarios where the old automation points can't be removed and also they playback properly but don't export properly.

I'm inclined to say this is remnant of the old "straight line" trick where you could draw a straight line by holding a modifier key and clicking two positions, but whatever the cause, I now have an automation track that I cannot make adjustments to.

This picture is an Audacity rendering of the track next to the LMMS automation.
image
In the case of this one, instead of graduating from 50% to 100% when rendering, the volume just spikes up to 100% without any graduation.

Also, notice the unique three dots, then three dots?

I tried CTRL + A to select all dots and delete, but CTRL+A doesn't seem to do anything.  I tried clicking a few more buttons and gave up.

For this particular track, I was trying to export to WAV for the new LMMS Bandcamp album and what's most interesting is the old events seem to playback fine on the song editor, but once I export, they have undesired effects.

This bug is two fold as it prevents edits to some old automation events and creates undesired effects to exported tracks.

For now I'm going to have to remove 0.9.91 and re-install 0.4.15 to get the track rendered for the album.  I'll put 0.9.91 back on once I'm finished rendering so that I can hopefully find more bugs like this.  If the developer requires an `mmpz`, please contact me individually for this particular track.

-Tres

zonkmachine:
Fixes #391
 * In response to one part of issue https://github.com/LMMS/lmms/issues/391, comment [here](https://github.com/LMMS/lmms/issues/391#issuecomment-72783217), in which you can't delete Automation Points that were made under a finer quantization. Don't consider quantization when deleting and just delete the value at the position.
We should maybe iterate through the neighbouring positions if there are no values at the current positon as it's a bit hard currently to hit the right spot for deletion. You have a similar problem when dragging an Automation Point which I haven't touched in this PR. When you select an Automation Point made under finer quantization you will instead create a new point at the quantized position.
 * Let an automation point clear the area it covers.
 * Triplets and no quantization option. This gives the Automation Editor the same quantization options as the other editors.
 * Improved erase action when moving mouse across Automation Points.
 * Eighth note, new default quantization value.

### DIFF
--- a/include/AutomationPattern.h
+++ b/include/AutomationPattern.h
@@ -79,11 +79,10 @@ public:
 	MidiTime putValue( const MidiTime & _time, const float _value,
 						const bool _quant_pos = true );
 
-	void removeValue( const MidiTime & _time,
-					  const bool _quant_pos = true );
+	void removeValue( const MidiTime & time );
 
 	MidiTime setDragValue( const MidiTime & _time, const float _value,
-						   const bool _quant_pos = true );
+						const bool _quant_pos = true );
 
 	void applyDragValue();
 
--- a/include/AutomationPattern.h
+++ b/include/AutomationPattern.h
@@ -76,13 +76,17 @@ public:
 	MidiTime timeMapLength() const;
 	void updateLength();
 
-	MidiTime putValue( const MidiTime & _time, const float _value,
-						const bool _quant_pos = true );
+	MidiTime putValue( const MidiTime & time,
+				const float value,
+				const bool quantPos = true,
+				const bool controlKey = false );
 
 	void removeValue( const MidiTime & time );
 
-	MidiTime setDragValue( const MidiTime & _time, const float _value,
-						const bool _quant_pos = true );
+	MidiTime setDragValue( const MidiTime & time,
+				const float value,
+				const bool quantPos = true,
+				const bool controlKey = false );
 
 	void applyDragValue();
 
--- a/src/core/AutomationPattern.cpp
+++ b/src/core/AutomationPattern.cpp
@@ -237,18 +237,13 @@ MidiTime AutomationPattern::putValue( const MidiTime & _time,
 
 
 
-void AutomationPattern::removeValue( const MidiTime & _time,
-									 const bool _quant_pos )
+void AutomationPattern::removeValue( const MidiTime & time )
 {
 	cleanObjects();
 
-	MidiTime newTime = _quant_pos ?
-				Note::quantized( _time, quantization() ) :
-				_time;
-
-	m_timeMap.remove( newTime );
-	m_tangents.remove( newTime );
-	timeMap::const_iterator it = m_timeMap.lowerBound( newTime );
+	m_timeMap.remove( time );
+	m_tangents.remove( time );
+	timeMap::const_iterator it = m_timeMap.lowerBound( time );
 	if( it != m_timeMap.begin() )
 	{
 		--it;
@@ -267,7 +262,7 @@ void AutomationPattern::removeValue( const MidiTime & _time,
 
 
 /**
- * @brief Set the position of the point that is being draged.
+ * @brief Set the position of the point that is being dragged.
  *        Calling this function will also automatically set m_dragging to true,
  *        which applyDragValue() have to be called to m_dragging.
  * @param the time(x position) of the point being dragged
@@ -648,7 +643,7 @@ void AutomationPattern::processMidiTime( const MidiTime & time )
 			}
 			else if( valueAt( time ) != value )
 			{
-				removeValue( time, false );
+				removeValue( time );
 			}
 		}
 	}
--- a/src/core/AutomationPattern.cpp
+++ b/src/core/AutomationPattern.cpp
@@ -204,23 +204,34 @@ void AutomationPattern::updateLength()
 
 
 
-MidiTime AutomationPattern::putValue( const MidiTime & _time,
-							const float _value,
-							const bool _quant_pos )
+MidiTime AutomationPattern::putValue( const MidiTime & time,
+					const float value,
+					const bool quantPos,
+					const bool controlKey )
 {
 	cleanObjects();
 
-	MidiTime newTime = _quant_pos ?
-				Note::quantized( _time, quantization() ) :
-				_time;
+	MidiTime newTime = quantPos ?
+				Note::quantized( time, quantization() ) :
+				time;
 
-	m_timeMap[newTime] = _value;
+	m_timeMap[ newTime ] = value;
 	timeMap::const_iterator it = m_timeMap.find( newTime );
+
+	// Remove control points that are covered by the new points
+	// quantization value. Control Key to override
+	if( ! controlKey )
+	{
+		for( int i = newTime + 1; i < newTime + quantization(); ++i )
+		{
+			AutomationPattern::removeValue( i );
+		}
+	}
 	if( it != m_timeMap.begin() )
 	{
 		--it;
 	}
-	generateTangents(it, 3);
+	generateTangents( it, 3 );
 
 	// we need to maximize our length in case we're part of a hidden
 	// automation track as the user can't resize this pattern
@@ -270,14 +281,16 @@ void AutomationPattern::removeValue( const MidiTime & time )
  * @param true to snip x position
  * @return
  */
-MidiTime AutomationPattern::setDragValue( const MidiTime & _time, const float _value,
-					   const bool _quant_pos )
+MidiTime AutomationPattern::setDragValue( const MidiTime & time,
+						const float value,
+						const bool quantPos,
+						const bool controlKey )
 {
 	if( m_dragging == false )
 	{
-		MidiTime newTime = _quant_pos  ?
-					Note::quantized( _time, quantization() ) :
-					_time;
+		MidiTime newTime = quantPos  ?
+				Note::quantized( time, quantization() ) :
+							time;
 		this->removeValue( newTime );
 		m_oldTimeMap = m_timeMap;
 		m_dragging = true;
@@ -288,10 +301,10 @@ MidiTime AutomationPattern::setDragValue( const MidiTime & _time, const float _v
 
 	for( timeMap::const_iterator it = m_timeMap.begin(); it != m_timeMap.end(); ++it )
 	{
-		generateTangents(it, 3);
+		generateTangents( it, 3 );
 	}
 
-	return this->putValue( _time, _value, _quant_pos );
+	return this->putValue( time, value, quantPos, controlKey );
 
 }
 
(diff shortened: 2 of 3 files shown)


zonkmachine:
Found a better solution that keeps the quantization but defaults for minimal value.

Umcaruje:
On an offtopic note, I think the automation editor should support all quantization values, it can't do triplets atm.

zonkmachine:
> it can't do triplets atm.

https://github.com/LMMS/lmms/pull/3358

src/core/AutomationPattern.cpp
```diff
@@ -243,8 +243,9 @@ void AutomationPattern::removeValue( const MidiTime & _time,
 	cleanObjects();
 
 	MidiTime newTime = _quant_pos ?
-				Note::quantized( _time, quantization() ) :
-				_time;
+				Note::quantized( _time,
+					DefaultTicksPerTact / 64 ) : // finest quantization
```
Umcaruje: Should this be adjusted to `DefaultTicksPerTact / 192` per #3358?

zonkmachine: Yes, or maybe reverted back to the first commit.

Umcaruje: Hmm yes, the first commit does look better, cause we never want to limit the deletion of automation points really, at any quantization.

zonkmachine:
OK. Reverted back to the first commit.

zonkmachine:
Tested something. Right now the quantized position only covers the starting point. Let it instead also cover the length of the Automation Point by just wiping clean the time positions it covers.

zonkmachine:
The point to be deleted is determined on a pixel basis, so you may need to zoom in to get to all time values. With no quantization on `AutomationPattern::removeValue()` it gets overall slightly harder to delete an Automation Point but it depends on how you use the mouse. With the last fix it gets pretty easy. You just 'turn the quantization up' and write over the old values. Needs more testing.

zonkmachine:
I bunched the three automation pattern/editor fixes together. Ready for testing

zonkmachine:
@Umcaruje A compromise on `AutomationPattern::removeValue()`. Keep quantization and also delete the raw time value at the same time. The quantized delete makes deleting quantized positions snappy and you don't get non deletable positions.

zonkmachine:
@Umcaruje I fixed the algorithm for erase so it's smoother and catches Automation Points around it proportional to the zooming factor, but I don't know how it works on different resolutions. Can you test this and especially erasing Automation Points when right click + moving across them at various zoom levels?

zonkmachine:
Merge?

src/gui/editors/AutomationEditor.cpp
```diff
@@ -139,6 +139,9 @@ AutomationEditor::AutomationEditor() :
 	}
 	m_quantizeModel.addItem( "1/192" );
 
+	connect(&m_quantizeModel, SIGNAL(dataChanged()), this, SLOT(setQuantization()));
+	m_quantizeModel.setValue( m_quantizeModel.findText( "1/4" ) );
```
Umcaruje: I've been using the automation editor lately, and find the 1/8 quantization to work best for my needs, tbqh. 1/4 feels too limiting and 1/16 is too much.

zonkmachine: OK. Changing to 1/8.

zonkmachine:
> * Let an automation point clear the area it covers.

@Umcaruje Did you test this change? It gives the Automation Point the same length as it's quantization. I think it's a really smooth solution. You could make something like a shift key override for this but I don't think it would be used a whole lot.

Umcaruje:
Hmm, I'm not sure how much I like the deleting solution, now that I've tested it. It would be nice if it catches the 20-40px around the point, but not the whole length. It just surprises me from a UX perspective. I record automation sometimes and I delete unnecessary points, and while doing that I hold my right mouse key. With the new behaviour this just wipes clean along the Y axis, which feels weird. 

>Did you test this change? It gives the Automation Point the same length as it's quantization. I think it's a really smooth solution. You could make something like a shift key override for this but I don't think it would be used a whole lot.

Yes, it actually is something I would use rather than my previous method of erasing. I'd implement a shift override, because LMMS doesn't allow to have two automation points at the same position. This makes ramping automation hard to do, so I usually use the lowest quantisation to put the previous point as close to the next one. The new behaviour would make changing those points hard.

zonkmachine:
> Hmm, I'm not sure how much I like the deleting solution, now that I've tested it. It would be nice if it catches the 20-40px around the point, but not the whole length.

Is this about erasing when sweeping with the mouse? For me the automation point is deleted when the mouse is at about 'one automation point diameter' distance. What is your screen resolution?

> Yes, it actually is something I would use rather than my previous method of erasing. I'd implement a shift override, because LMMS doesn't allow to have two automation points at the same position.

OK. Will try the shift fix..

zonkmachine:
> It would be nice if it catches the 20-40px around the point,

Right, I think I got you. Some space to counter this #3301 ?
I've tried to deal with it but I pretty much failed there unfortunately. I'll maybe look at it again but not right now. I'm taking the 'automation point length' hack out for now and send it back to the drawing board.

zonkmachine:
> >You could make something like a shift key override for this but I don't think it would be used a whole lot.
> 
> Yes, it actually is something I would use rather than my previous method of erasing. I'd implement a shift override, because LMMS doesn't allow to have two automation points at the same position. This makes ramping automation hard to do, so I usually use the lowest quantisation to put the previous point as close to the next one. The new behaviour would make changing those points hard.

OK. Shift key was busy so I used the Ctrl key.

zonkmachine:
I think this works well.
@Umcaruje Can you test this and see if the `<Ctrl>` button works to override the quantization length function?

@tresf I think this closes https://github.com/LMMS/lmms/issues/391 though there is more than one issue in that ticket but it looks like the export bug part is a bit dated.

tresf:
@zonkmachine I don't have time to build and test this, so if someone else can confirm that automation points smaller than default Quantization can be edited/deleted with this PR, the bug should be closed.

>  the export bug part is a bit dated.

I had originally assumed they were related, but on second glance it should be reopened as a separate item if still valid after this is closed.

zonkmachine:
> I had originally assumed they were related, but on second glance it should be reopened as a separate item if still valid after this is closed.

I had a second look at it and it is off course the same issue. Multiples of 3 is the positions for 1/64 quantization so if you erase with 1/16 quantization you'll get those values. Just duplicated the bug.

dots

zonkmachine:
But then again you have the issue of the volume automation and that's not really related...

zonkmachine:
If there are no further comments I'll merge this later this evening.

Umcaruje:
So I did further testing, the Ctrl override works neatly, but I think we should display a Hint so users actually know that it exists.

The deleting behaviour is not something that I really like, cause it will make accidental deletes much more common imo. Example: 
gifrecord_2017-03-04_110711

lukeoftheaura:
all seems working to me, I agree with @Umcaruje there should be a hint for the ctrl override.

zonkmachine:
whatsup

Qt cuts off the last words sometimes?

```cpp
	m_quantizeComboBox->setWhatsThis( tr( "Quantization. Sets the smallest "
				"step size for the Automation Point. By default "
				"this also sets the length of the Automation "
				"Point, clearing out other points in the range. "
				"Press <Ctrl> to override this behaviour." ) );
```

zonkmachine:
> The deleting behaviour is not something that I really like, cause it will make accidental deletes much more common imo. Example:

Dropping the quantization on delete makes it more or less a necessity. It's painful enough in 1.1.3 .
Should I make the action a bit closer to the point to delete?

zonkmachine:
@lukeoftheaura @Umcaruje Thanks for testing. I made the erase/delete a bit less aggressive and the whatsthis now reads:

```
setWhatsThis( tr( "Quantization. Sets the smallest "
				"step size for the Automation Point. By default "
				"this also sets the length, clearing out other "
				"points in the range. Press <Ctrl> to override "
				"this behaviour." ) );
```

zonkmachine:
I'm not sure what's wrong in the animation though. It looks pretty good and I see no accidental deletions in it.

ps. I'm still on schedule to merge this later on... :smiley:

include/AutomationPattern.h
```diff
@@ -76,14 +76,17 @@ class EXPORT AutomationPattern : public TrackContentObject
 	MidiTime timeMapLength() const;
 	void updateLength();
 
-	MidiTime putValue( const MidiTime & _time, const float _value,
-						const bool _quant_pos = true );
+	MidiTime putValue( const MidiTime & time,
+				const float value,
+				const bool quantPos = true,
+				const bool controlKey = false );
```
jasp00: Core objects should not have references to the control key, which is a UI concept.

zonkmachine: Ouch! I'm looking into this now but I'm not sure how to fix that. The objective is to be able to turn off the feature to give the control point a 'minimum length' according to it's quantization value.

jasp00: You have two choices:
1. The clean way: Move the feature to the view object, which reacts to the control key.
2. The easy way: Keep the feature in the model and rename `controlKey` to something like `ignoreSurroundingPoints`.

zonkmachine: > 2. The easy way: Keep the feature in the model and rename `controlKey` to something like `ignoreSurroundingPoints`.

Do you mean like:
```diff
diff --git a/include/AutomationPattern.h b/include/AutomationPattern.h
index 480b69f..1bb5c03 100644
--- a/include/AutomationPattern.h
+++ b/include/AutomationPattern.h
@@ -79,7 +79,7 @@ public:
        MidiTime putValue( const MidiTime & time,
                                const float value,
                                const bool quantPos = true,
-                               const bool controlKey = false );
+                               const bool ignoreSurroundingPoints = false );
 
        void removeValue( const MidiTime & time );
 
diff --git a/src/core/AutomationPattern.cpp b/src/core/AutomationPattern.cpp
index ffe13ff..7996af3 100644
--- a/src/core/AutomationPattern.cpp
+++ b/src/core/AutomationPattern.cpp
@@ -208,7 +208,7 @@ void AutomationPattern::updateLength()
 MidiTime AutomationPattern::putValue( const MidiTime & time,
                                        const float value,
                                        const bool quantPos,
-                                       const bool controlKey )
+                                       const bool ignoreSurroundingPoints )
 {
        cleanObjects();
 
@@ -221,7 +221,7 @@ MidiTime AutomationPattern::putValue( const MidiTime & time,
 
        // Remove control points that are covered by the new points
        // quantization value. Control Key to override
-       if( ! controlKey )
+       if( ! ignoreSurroundingPoints )
        {
                for( int i = newTime + 1; i < newTime + quantization(); ++i )
                {

```

jasp00: Yes.

zonkmachine: Thanks! Fixed here: https://github.com/LMMS/lmms/pull/3459

Issue #3353, opened 2017-02-13
Impossible to change the track name when the instrument window is open 


Umcaruje:
![gifrecord_2017-02-13_225132](https://cloud.githubusercontent.com/assets/6282045/22905049/126b1530-f23f-11e6-819e-3c2aac55633b.gif)

When double clicking the track label button that is active (window open), it is impossible to change the name of the track. The QLineEdit appears for a very short time, but double clicking will just close and open the window again.

CC @BaraMGB

BaraMGB:
Can not reproduce here. But, you can rename by right click too.

Umcaruje:
Weird, are you on Qt5 or Qt4 @BaraMGB?

BaraMGB:
Okay, can reproduce on current master. But can't on RC2. Something is happend in the meantime.

BaraMGB:
I guess it was me with this: https://github.com/LMMS/lmms/commit/c5aa22663e555524d34813dd00bf350e5d5ba33b

musikBear:
FYI RC 2.8 windows x32 was fine

PR #3356, opened 2017-02-14
fix double click rename on track label button

BaraMGB:
fixes https://github.com/LMMS/lmms/issues/3353

### DIFF
--- a/src/gui/widgets/TrackLabelButton.cpp
+++ b/src/gui/widgets/TrackLabelButton.cpp
@@ -162,9 +162,9 @@ void TrackLabelButton::mouseDoubleClickEvent( QMouseEvent * _me )
 
 
 
-void TrackLabelButton::mouseReleaseEvent(QMouseEvent *_me)
+void TrackLabelButton::mouseReleaseEvent( QMouseEvent *_me )
 {
-	if( m_buttonRect.contains( _me->globalPos(), true ) )
+	if( m_buttonRect.contains( _me->globalPos(), true ) && m_renameLineEdit->isHidden() )
 	{
 		QToolButton::mousePressEvent( _me );
 	}


Umcaruje:
It can be clicked now, but here's another thing that happens now: When you double click to rename, the window gets toggled, if it was open, it will get closed when double clicked and vice versa. This should not happen and the window should stay in the same state when the button is double clicked to rename.

tresf:
If `rename()` is already being handled, we should considering adding an exit on `mouseReleaseEvent` when `QEvent::MouseButtonDblClick` occurs as to not double-invoke.

BaraMGB:
>If rename() is already being handled, we should considering adding an exit on mouseReleaseEvent when QEvent::MouseButtonDblClick occurs as to not double-invoke.

I guess that's exactly what this PR is about. ;)

>It can be clicked now, but here's another thing that happens **now**: When you double click to rename, the window gets toggled, if it was open, it will get closed when double clicked and vice versa. This should not happen and the window should stay in the same state when the button is double clicked to rename.

This behavior isn't new. I build 1.1.90 and if you double click on the button the instrument window appears and the QDialog for renaming appears. I guess it is impossible to distinguish a first click of a double click from a single click. 

I tested it on a file browser. If I click a folder, the folder is marked. If I double click a folder the folder will be marked before it will be opened. 

We should consider to change the button function for only open the instrument window rather to toggle on/off (after 1.2)

tresf:
>  I guess it is impossible to distinguish a first click of a double click from a single click.

This probably isn't new to Qt.  All modern GUI programming languages support differentiating single from double.  Please do not make this assumption.

BaraMGB:
>This problem isn't new to Qt. All modern GUI programming languages support differentiating single from double. Please do not make this assumption.

Okay, what should we do?

tresf:
The more I read on this problem, the more disheartening it is!  I've now found three threads that talk about a timer.  I don't know.  Perhaps I have to eat my words.  This is quite common in GUI programming.  I'm quite puzzled if Qt doesn't have a way to handle it.

BaraMGB:
I've tested the timer method. The delay is very annoying. That is not a solution for lmms.

tresf:
@BaraMGB what if we leverage the OS double-click interval instead of rolling our own guessed timeout?

According to [this thread](http://www.qtcentre.org/threads/53610-Qt5-mouseDoubleClickEvent), we should be able to probe the OS and filter for single-clicks, right?

```cpp
ulong doubleClickInterval = static_cast<ulong>(qApp->styleHints()->mouseDoubleClickInterval());
doubleClick = e->timestamp - mousePressTime < doubleClickInterval && button == mousePressButton;
 
// ...
 
if (doubleClick) {
        mousePressButton = Qt::NoButton;
        const QEvent::Type doubleClickType = frameStrut ? QEvent::NonClientAreaMouseButtonDblClick : QEvent::MouseButtonDblClick;
        QMouseEvent dblClickEvent(doubleClickType, localPoint, localPoint, globalPoint,
                                  button, buttons, e->modifiers);
        dblClickEvent.setTimestamp(e->timestamp);
        QGuiApplication::sendSpontaneousEvent(window, &dblClickEvent);
}
```

If we set the timeout value to `static_cast<ulong>(qApp->styleHints()->mouseDoubleClickInterval());` it should feel as fast and natural as the OS, right?  Or does it still feel bad?

BaraMGB:
@tresf  
>what if we leverage the OS double-click interval instead of rolling our own guessed timeout?

In my tests I used `qApp->mouseDoubleClickInterval()` which takes on my system 400 ms. The computer can not know if the user wants to click a second time before the user clicks. That is the problem here. 

This button is a design fail. The actions wich will be triggered by click and double click should be related somehow. (_mark and open_ for example) In our case they are not related to each other.

Umcaruje:
>This button is a design fail. 

If you think this can't be solved by timers, I vote on removing the drag function of the buttons, as double clicking to rename is something much more useful imho.

tresf:
> This button is a design fail. The actions wich will be triggered by click and double click should be related somehow. (mark and open for example) In our case they are not related to each other.

Yes.  The more I observe single versus double on OS and application items, the more this statement is true.  Double always invokes single.

The only possible solution to this problem is to set the window visible and rename it in the title bar of the instrument window.  Renaming on button can be done via right click too I suppose, but what you're saying makes perfect sense.  We shouldn't have mutually exclusive actions between single and double.  Double should invoke single.

BaraMGB:
>If you think this can't be solved by timers, I vote on removing the drag function of the buttons, as double clicking to rename is something much more useful imho.

No, this has nothing to do with the drag function. Look at my comment above:

>This behavior isn't new. I build 1.1.90 and if you double click on the button the instrument window appears and the QDialog for renaming appears.

BaraMGB:
>The only possible solution to this problem is to set the window visible and rename it in the title bar of the instrument window. 

I had a similar idea, too. There's a lineedit in the instrument window for this.

Umcaruje:
Well since the tracks can be renamed with the right click, I vote we remove the double click feature, and just allow drags. Also on double click we can make a `TextFloat` appear with text similar to `Hint: You can rename Instrument tracks by Right-clicking on them`

Umcaruje:
Though this might propose a problem if we in the future add a context menu to this button. Then we wouldn't have a 1 step action for rename.

tresf:
> Well since the tracks can be renamed with the right click

I think this is part of the UI/UX discrepancies that we really need to tackle.  Double-click is quite common and right-click context menu, rename are quite common but right right rename is ran non-standard.

Umcaruje:
Yeah, well I don't get the need to drag the instrument using it's label cause we have the grip on the side. Also the settings button can easily just become a right click context menu, not too much need for it, except it just takes space.

tresf:
> Yeah, well I don't get the need to drag the instrument using it's label cause we have the grip on the side.

Most items these days you can long-hold and move, select.  This is the case for most dock-style bars and most touch-devices as well.

> Also the settings button can easily just become a right click context menu, not too much need for it, except it just takes space.

I don't think I've ever used the gear except remove/clone, and agreed.  Can't we just put this into the compact design and make it default?

BaraMGB:
>Yeah, well I don't get the need to drag the instrument using it's label cause we have the grip on the side. 

Again: the problematic behavior has nothing to do with the dragging of the button!

>When you double click to rename, the window gets toggled, if it was open, it will get closed when double clicked and vice versa.

Absolutely nothing new! This behavior has always been like that. Please test it in 1.1 or earlier releases.

Umcaruje:
As @BaraMGB this behaviour is noted and we could open up a seperate issue to rework the UX of the track labels. This fixes the issue I reported so I'm merging.

PR #3062, opened 2016-10-03
Piano Roll and Automation editor grid redesign (w/ @BaraMGB)

Umcaruje:
Fixes #3019 and #2308

Thank you very much @BaraMGB for collaborating on this Pull Request with me and helping me out with my problems.

screenshot from 2016-10-03 20 34 37
screenshot from 2016-10-03 20 34 18

The grid is ligher, the background alternates shades. I also did a general CSS cleanup, fixed the indentation and stuff.

Feedback and testing welcome.

PR #3358, opened 2017-02-14
Triplets in Automation Editor

zonkmachine:
:tractor:

### DIFF
--- a/src/gui/editors/AutomationEditor.cpp
+++ b/src/gui/editors/AutomationEditor.cpp
@@ -1959,9 +1959,10 @@ void AutomationEditor::zoomingYChanged()
 
 void AutomationEditor::setQuantization()
 {
-	int quantization = m_quantizeModel.value() < 7
-			? 1 << m_quantizeModel.value()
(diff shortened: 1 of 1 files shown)


Umcaruje:
Cool, though I think we should also allow the maximum quantisation amount of `1/192`, to be absolutely consistent with the piano roll. Also the drawing of triplets is fixed with #3062

zonkmachine:
> I think we should also allow the maximum quantisation amount of 1/192

Tried it, pushed it. I think maybe they are a bit too hard to erase without a proper select tool.

> Also the drawing of triplets is fixed with #3062

:+1:

zonkmachine:
> I think maybe they are a bit too hard to erase without a proper select tool.

No, it's about the same.

zonkmachine:
Needs testing and review. I can merge this into https://github.com/LMMS/lmms/pull/3358 if that simplifies things.

zonkmachine:
> I can merge this into #3352 if that simplifies things.

Closing this, now in https://github.com/LMMS/lmms/pull/3352

PR #3361, opened 2017-02-16
Offbeatnotes

### LINKED ISSUE #1681: Piano Roll to Step Sequencer Bug
Steps to reproduce:
1. Create four notes on the first four tombstones in a BB pattern
2. Double click the fifth tombstone to open in piano roll
3. Ctrl-drag to select the first four notes
4. Shift drag them one 1/2 note to the left (tombstones 9-12)
5. Close piano roll. Tombstones 9-12 should be lit.
6. Add a note on tombstone 6. Tombstones 9-12 should go dark.

This can be reproduced in many more situations than this one, it's just a specific example tat's easy to reproduce.

zonkmachine:
Fixes #1681

From a discussion that starts here: https://github.com/LMMS/lmms/issues/1681#issuecomment-277537185
 - ~~When you add an ordinary note to a beat pattern the pattern is visually turned into a melody pattern, but the beat notes aren't represented in it.~~ _Fixed in #3375_
 - When a beat note is shifted off it's 16th part position it's no longer visible. Instead, turn the pattern into a melody pattern.

offbeatnotes
_Suggested fix: Pattern switch to MelodyPattern when one or more beat notes are not visible in the BBEditor._

### DIFF
--- a/include/Pattern.h
+++ b/include/Pattern.h
@@ -88,7 +88,7 @@ public:
 	{
 		return m_patternType;
 	}
-
+	void checkType();
 
 	// next/previous track based on position in the containing track
 	Pattern * previousPattern() const;
@@ -132,7 +132,6 @@ private:
 	MidiTime beatPatternLength() const;
 
 	void setType( PatternTypes _new_pattern_type );
-	void checkType();
 
 	void resizeToFirstTrack();
 
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -2451,6 +2451,11 @@ void PianoRoll::dragNotes( int x, int y, bool alt, bool shift, bool ctrl )
 
 					note->setPos( MidiTime( pos_ticks ) );
 					note->setKey( key_num );
+					// If dragging beat notes check if pattern should be MelodyPattern
+					if( note->length() < 0 )
+					{
+						m_pattern->checkType();
+					}
 				}
 			}
 		}
--- a/src/tracks/Pattern.cpp
+++ b/src/tracks/Pattern.cpp
@@ -369,7 +369,9 @@ void Pattern::checkType()
 	NoteVector::Iterator it = m_notes.begin();
 	while( it != m_notes.end() )
 	{
-		if( ( *it )->length() > 0 )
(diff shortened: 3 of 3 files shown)


src/gui/editors/PianoRoll.cpp
```diff
@@ -2451,7 +2451,7 @@ void PianoRoll::dragNotes( int x, int y, bool alt, bool shift, bool ctrl )
 
 					note->setPos( MidiTime( pos_ticks ) );
 					note->setKey( key_num );
-					if( note->length() <= 0 )
+					if( note->length() < 0 )
```
zonkmachine: Because '0' length notes are no more.

zonkmachine:
Needs testing/review.

The first change is pretty straight forward and I think 'merge is go' for that one.

As for the second change I have no better idea on store. It works as it is but it's probably a matter of taste more than anything else. Pros, there are no more hidden notes in BBTracks and it's easy to implement. If you add or drag a note off beat you get an instant representation of it as the pattern is updated. Cons, you maybe simply prefer the older way and if you have more than one off beat note it may be annoying to find them to go back to a BBTrack. Maybe off beat beat track notes could be given a diffferent colour in the song editor?

zonkmachine:
> The first change is pretty straight forward and I think 'merge is go' for that one.

Merged in https://github.com/LMMS/lmms/pull/3375 the other feature needs review.

include/Pattern.h
```diff
@@ -88,7 +88,7 @@ class EXPORT Pattern : public TrackContentObject
 	{
 		return m_patternType;
 	}
-
+	void checkType();
```
zonkmachine: I made this public to be able to call it from `PianoRoll.cpp` .

zonkmachine:
Merge?

zonkmachine:
No objections so I will merge this tomorrow.

zonkmachine:
@jasp00 Unfortunately this breaks the upgrade of void [DataFile::upgrade_1_2_0_rc3()](https://github.com/LMMS/lmms/blob/master/src/core/DataFile.cpp#L903)
Unfortunately I won't have time to address this in any other way than reverting the commit.

Issue #3349, opened 2017-02-13
Remove menu icons from non-document menu actions on macOS/OS X


follower:
(_Originally this bug was this but see #3362:_ 

As previously mentioned (https://github.com/LMMS/lmms/issues/2890#issuecomment-273351599) the new white menu icons do not display correctly on Mac OS X 10.8 with Qt5:

![lmms-menu-icons-display-issue-qt5](https://cloud.githubusercontent.com/assets/189962/22863434/c9ec8d7e-f1a4-11e6-988d-14f14aa94810.png)

Related version info:

![lmms-menu-icons-display-issue-qt5-version-info](https://cloud.githubusercontent.com/assets/189962/22863458/36929612-f1a5-11e6-9e9a-a4803add1295.png)

I assume if they display on later OS X versions it's due to a change in menu background colour?
)

_Now:_

FWIW, I think on Mac OS X / macOS icons should not be displayed on menus at all, as it's not "standard practice" on the platform and IMO makes lmms appear "less native". Further details on request. :)

It appears specifying [`Qt::AA_DontShowIconsInMenus`](http://doc.qt.io/qt-5/qt.html#ApplicationAttribute-enum) would enable this but I've not tested it yet.

tresf:
> IMO makes lmms appear "less native"

Probably the more important part of this is the UI experience.  OSs which can't align text in a menu should have the icons turned off, period.  Please issue a Pull Request with `Qt::AA_DontShowIconsInMenus` activated for MacOS only and we can finally put this issue to rest. 👍

tresf:
Sounds good.  I noticed the icons don't seem so cosmetically weird if the icons aren't white.  For reference here's a side-by-side with Google Chrome:

![image](https://cloud.githubusercontent.com/assets/6345473/22896104/da46428a-f1ec-11e6-9ffb-39a09d87799d.png)  ![image](https://cloud.githubusercontent.com/assets/6345473/22896001/77de8684-f1ec-11e6-9196-fe4c9c8f5b70.png)

tresf:
I just read a bug report with users complaining about this identical behavior, except when in dark mode.  Apparently there's an unmerged feature to help... https://codereview.qt-project.org/#/c/115120/

Linked from https://bugreports.qt.io/browse/QTBUG-42109.

For the short-term, we can use a quick hack, such as inverting the color palette for these icons. http://doc.qt.io/qt-5/qimage.html#invertPixels

follower:
[...]

PR #3363, opened 2017-02-16
Remove non-document menu icons on macOS / OS X

follower:
See: #3349

### DIFF
--- a/src/gui/MainWindow.cpp
+++ b/src/gui/MainWindow.cpp
@@ -894,6 +894,7 @@ void MainWindow::updateRecentlyOpenedProjectsMenu()
 			m_recentlyOpenedProjectsMenu->addAction(
 					embed::getIconPixmap( "project_file" ), *it );
 #ifdef LMMS_BUILD_APPLE
+			m_recentlyOpenedProjectsMenu->actions().last()->setIconVisibleInMenu(false); // Workaround for https://bugreports.qt.io/browse/QTBUG-44565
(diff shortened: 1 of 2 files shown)


src/gui/MainWindow.cpp
```diff
@@ -893,6 +893,10 @@ void MainWindow::updateRecentlyOpenedProjectsMenu()
 		{
 			m_recentlyOpenedProjectsMenu->addAction(
 					embed::getIconPixmap( "project_file" ), *it );
+#ifdef LMMS_BUILD_APPLE
+			m_recentlyOpenedProjectsMenu->actions().last()->setIconVisibleInMenu(false); // Workaround for https://bugreports.qt.io/browse/QTBUG-44565
```
tresf: I feel this would be cleaner

```cpp
m_recentlyOpenedProjectsMenu->actions().last()->setIconVisibleInMenu(false); // Per QTBUG-44565
m_recentlyOpenedProjectsMenu->actions().last()->setIconVisibleInMenu(true);
```

follower: I did wonder about URL vs bug number but went with the full URL since that's what we used in #3350.

I'd tend toward still indicating it's a workaround, so would prefer to go with:
```
m_recentlyOpenedProjectsMenu->actions().last()->setIconVisibleInMenu(false);  // QTBUG-44565 workaround
```

follower:
Thanks for helping getting this in merge-able state and merging it! :)

PR #3364, opened 2017-02-16
[WIP!] Open to bar

Spekular:
Currently only "works" in the automation editor. First time pressing "open to bar" does not work. Second press and after will scroll the automation editor scrollbar correctly, but not the viewport. Wiggling the scrollbar updates the viewport.

m_playMode needs to be set to PlaySong for scrolling to work. I'm currently changing this mode, scrolling, and then putting it back, using functions I added to Song.cpp/Song.h. I'm not entirely sure if it resets correctly (I don't know how to invoke different play modes and test if it goes back), or if this is a good method for changing playmode.


Spekular:
My biggest holdup at the moment is whether or not it's ok to change playmode, even if it's set back after. I feel like this would cause issues that might be hard to notice. One option could be to gray out "open to bar" if we're not in PlaySong mode. Thoughts?

tresf:
@Spekular this PR has gone stale.  Please either update it or close it out.

Spekular:
@tresf rebased to master if that's what you meant. If you mean actual updates then I should close this and the vertical zoom PR, as they're both open because I hit a wall and figured someone might have leads on what to do if I showed my progress.

tresf:
>  If you mean actual updates then I should close this and the vertical zoom PR, as they're both open because I hit a wall and figured someone might have leads on what to do if I showed my progress.

Yes, I meant update the progress on this.  That's your call.

src/gui/AutomationPatternView.cpp
```diff
@@ -89,6 +96,22 @@ void AutomationPatternView::update()
 
 
 
+
+void AutomationPatternView::allowOpenToBar()
+{
+	if(Engine::getSong()->playMode() == 1)
+	{ 
+		TextFloat::displayMessage("Right Playmode, allow");
```
Spekular: If graying out "open to playhead" based on playMode is ok, this should allow the option to be selected in context menus. Right now it just displays a message.

src/gui/AutomationPatternView.cpp
```diff
@@ -89,6 +96,22 @@ void AutomationPatternView::update()
...
+	}//_cm.a.setEnabled(true); }
+	else
+	{
+		TextFloat::displayMessage("Disallow/Gray Out");
```
Spekular: If graying out "open to playhead" based on playMode is ok, this should disallow/gray out the option in context menus. Right now it just displays a message.

src/gui/editors/AutomationEditor.cpp
```diff
@@ -2563,6 +2563,37 @@ void AutomationEditorWindow::open(AutomationPattern* pattern)
...
+{
+	//How many bars we need to scroll. Playhead position - pattern start position
+	int offset = 0;
+	if (position.getTact() < Engine::getSong()->getTacts()) { offset = Engine::getSong()->getTacts() - position.getTact(); }
+
+	//Open the pattern
+	open(pattern);
+
+	//This scrolls to the right part
+	//12 scroll values per 16th note
```
Spekular: 12 * 16 = 192. Scrollbar is set in 1/192th notes

Spekular:
Closing. Issues:
- [ ] Doesn't work at all when first used
- [ ] Doesn't update view in automation editor until the user nudges the scrollbar
- [ ] playMode weirdness
- [ ] Ought to be available for every editor

Issue #2951, opened 2016-07-30
"cannot find -lasound" (compiling on Lubuntu for Windows)


T0NIT0RMX:
Hi,

I'm trying to compile LMMS on Lubuntu 14.04 for Windows,
I followed -> https://github.com/LMMS/lmms/wiki/Compiling-lmms-(Windows)

But when I get to step "Package"
I type :
`make package`

And when it's complete, I get :
![lasound](https://cloud.githubusercontent.com/assets/12613367/17272364/e5bd8efa-5694-11e6-9a02-0c5555fbcdff.PNG)

I installed all dependencies, I'm not the only one to have this issue, it also happened to @Umcaruje 

I hope I provided all needed infos, if you need more, just ask and I'll reply ASAP.
Thanks by advance for your help.

zonkmachine:
try:
`sudo apt-get install libasound2-dev`

Umcaruje:
Actually, this is building with mingw, which doesn't provide alsa, and alsa is disabled in the cmake configuration. I'm not sure what is wrong, but I'm expiriencing this same issue.

tresf:
AFAIK, `lmms.exe` is build from the `src/CMakeLists.txt` which explicitly links against ALSA if it's found and [stores it in a variable called `LMMS_REQUIRED_LIBS`](https://github.com/zonkmachine/lmms/blob/master/src/CMakeLists.txt#L121).

I agree with @Umcaruje, this should be empty but AFAIK the issue should happen if running `make` or `make package`.

If you take `${ASOUND_LIBRARY}` out of `LMMS_REQUIRED_LIBS` does it succeed?  If so, there may be something fishing going on with [`FindAlsa.cmake`](https://github.com/LMMS/lmms/blob/master/cmake/modules/FindAlsa.cmake) (or something residual from a previous build, perhaps).

T0NIT0RMX:
@zonkmachine Thanks, but already installed.

@tresf : I think I got it to work thanks to your help
I proceeded like this :
In `src/CMakeLists.txt` I took `${ASOUND_LIBRARY}` out of `LMMS_REQUIRED_LIBS`

Umcaruje:
Ok, so the actual issue seems to be with portaudio, as `asound` seems to be its required library(This is from the CMakeCache.txt):
[...]

PR #3369, opened 2017-02-19
Fixes build errors under mingw (3113)

DragonEagle:
Modified build bash scripts for mingw to fix errors when building under mingw.  Specifically when compiling gigplayer and malletstk.  The build now compiles without error on mingw systems with a clean install. There are still some issues with make package and some dlls that can't be found, but the program compiles and runs.  It is beyond my abilities with cmake to fix these.

-E. Allen Soard

### DIFF
--- a/cmake/build_mingw32.sh
+++ b/cmake/build_mingw32.sh
@@ -8,7 +8,7 @@ else
 	CMAKE_OPTS="$CMAKE_OPTS -DLMMS_BUILD_MSYS=1"
 fi
 
-export PATH=$PATH:$MINGW/bin
+export PATH=$MINGW/bin:$PATH
 export CFLAGS="-march=pentium3 -mtune=generic -mpreferred-stack-boundary=5 -mfpmath=sse"
 export CXXFLAGS="$CFLAGS"
 
--- a/cmake/build_mingw64.sh
+++ b/cmake/build_mingw64.sh
@@ -8,7 +8,7 @@ else
 	CMAKE_OPTS="$CMAKE_OPTS -DLMMS_BUILD_MSYS=1"
 fi
 
-export PATH=$PATH:$MINGW/bin
+export PATH=$MINGW/bin:$PATH
 
 if [ "$1" = "-qt5" ]; then
         QT5=True
--- a/cmake/msys/fetch_ppa.sh
+++ b/cmake/msys/fetch_ppa.sh
@@ -20,8 +20,8 @@ temp_file=/tmp/ppa_listing_$$
 temp_temp_file=/tmp/ppa_listing_temp_$$
 
 skip_files="binutils openssl flac libgig libogg libvorbis x-bootstrap zlib"
-skip_files="$skip_files x-runtime gcc qt_4 qt5 fltk x-stk pkgconfig" 
-skip_files="$skip_files libjpeg glib2 libpng"
+skip_files="$skip_files x-runtime gcc qt_4 qt5 x-stk pkgconfig" 
+skip_files="$skip_files glib2 libpng"
 
 echo "Connecting to $PPA_HOST to get list of packages..."
 wget -qO- $PPA_URL |grep "Filename:" > $temp_file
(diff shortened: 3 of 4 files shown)


tresf:
@DragonEagle I'm intrigued you're using these msys2 helpers for mingw.  Is this on a Linux machine or Windows?  They were written as helpers for the Windows build environments, but we don't Travis-CI test Windows, so the only way to know the impact of these on Windows is to run the painstaking task of [setting up a Windows build environment](https://github.com/LMMS/lmms/wiki/Compiling-lmms-(On%20Windows)) which I personally won't have time for for a while.

DragonEagle:
This is on a windows machine.  I meant msys.  sorry.

cmake/build_mingw32.sh
```diff
@@ -8,7 +8,7 @@ else
 	CMAKE_OPTS="$CMAKE_OPTS -DLMMS_BUILD_MSYS=1"
 fi
 
-export PATH=$PATH:$MINGW/bin
+export PATH=$MINGW/bin:$PATH
```
Umcaruje: Will these changes affect linux builds though?

tresf: > `export PATH=$MINGW/bin:$PATH`
> > Will these changes affect linux builds though?

They should be benign and Travis-CI passes, which is a good sign.

What this change does is force the path to prefer mingw's BIN **first** to any system BIN.  I'd expect this behavior as well and we use it in other areas.  If I were to harbor a guess, I'd say this is just an oversight that's been copied and pasted since our first builds (see https://github.com/lmms/lmms/commit/d3516cd5c51d88b1ba6bd0428c45137552f98186#diff-7e1f009e6b046eedd463b471771e7a05R2)

Umcaruje: Ah, ok, this might actually be the cause of #2951, I'll check if this change fixes that.

tresf: > this might actually be the cause of #2951, I'll check if this change fixes that.

It shouldn't help that problem.  This should only fix problems where the same library exists in both `$PATH` as well as `$MINGW/bin` which I wouldn't expect to solve #2951.  This is just a change which fixes the order of resolution.  #2951 is more likely related to something with `pkg-config` or `CMAKE_PREFIX_PATH` since the library isn't found at all.

DragonEagle: Actually might fix it.  I made the change because the right pkg-config wasn't being found so it could find libsndfile.  I got a successful build after that,  and I wasn't aware of #2951.

tresf: >  pkg-config wasn't being found 

Good point.

Umcaruje: I can confirm this change fixed #2951 for me. My machine and (@T0NIT0RMX's too) probably had portaudio dev files installed alongside the mingw ones, so it was picking them up for pkg-config and looking for alsa, which is probably required for linux portaudio.

Umcaruje: Great job @DragonEagle :D

tresf: 👍 @Umcaruje let's commit that change directly then (no PR).  That will close out your issue and then @DragonEagle can rebase his PR as needed.

Umcaruje: Well I was planning on actually setting a windows environment tommorow, and testing this PR, and possibly merging it, so if I encounter problems on my windows environment, I'll do the direct change :+1:

tresf: @Umcaruje sounds good.  FYI to you @DragonEagle and anyone else, we may want to start (eventually) publishing to [`scoop.sh`](http://scoop.sh/) which is Homebrew clone for Windows.

Umcaruje:
This works amazingly well @DragonEagle this is a terrific job you've done. The only issue I encountered is that the libgig download gives a 404, but that is a problem with the linuxsampler website, and will probably get fixed. I went around the issue by using a web archive link at line 209:
```bash
wget https://web.archive.org/web/20160330120039/http://download.linuxsampler.org/packages/libgig-$gigver.tar.bz2 -O $HOME/gig-source.tar.xz
```

I'm merging this. Thank you for your contribution, this makes compiling on Windows a breeze 🎉

Issue #3223, opened 2017-01-09
LMMS crash when open one of my project


hashhakaj:
Hello Everyone
I did share this project one year ago and put note it will crash your LMMS 100% when it begins play from claps beat bars section at the end of claps bar, colored in red, if you do extend it.
I've done this with different windows hardware and even I've test it with Mac. all have the same situation.
this experiment is done with LMMS 1.1.3
However, I've test the project in latest version of LMMS 1.2. and it crashed at claps beat faster than in LMMS 1.1.3 .

So, I hope that problem be fixed.
project link
https://lmms.io/lsp/?action=show&file=7616

tresf:
Confirmed.  Thanks for the bug report.

Umcaruje:
Looks like LMMS is running out of buffers. Backtrace: 
```
BufferManager: out of buffers
BufferManager: out of buffers

Thread 8 "QThread" received signal SIGABRT, Aborted.
[Switching to Thread 0x7fffa7074700 (LWP 18198)]
0x00007ffff4244428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
54	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt full
#0  0x00007ffff4244428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
        resultvar = 0
        pid = 18183
        selftid = 18198
#1  0x00007ffff424602a in __GI_abort () at abort.c:89
        save_stage = 2
        act = {__sigaction_handler = {sa_handler = 0x7fff94026010, sa_sigaction = 0x7fff94026010}, sa_mask = {__val = {140735995655936, 140735995656368, 
              140737330365679, 2802268544, 140735676571690, 124560769757, 9644727278355117056, 0, 4294967041, 140737333684128, 10, 140737333684152, 140737293141872, 
              140735995660032, 9644727278355117056, 140735995656480}}, sa_flags = 0, sa_restorer = 0x3}
        sigs = {__val = {32, 0 <repeats 15 times>}}
#2  0x00007ffff67fcf15 in qt_message_output(QtMsgType, char const*) () from /usr/lib/x86_64-linux-gnu/libQtCore.so.4
No symbol table info available.
#3  0x00007ffff67fd371 in ?? () from /usr/lib/x86_64-linux-gnu/libQtCore.so.4
No symbol table info available.
[...]

PR #3372, opened 2017-02-19
Envelope time modifier

zonkmachine:
envelope

Just some things I've wanted to test since https://github.com/LMMS/lmms/pull/411
Also in response to issue https://github.com/LMMS/lmms/issues/3223
I think It's a good idea to make dangerously long envelopes a conscious decision. 

I never liked the  0 - 2.0 value for the envelope times. This fixes it right up and instead adds a check box to turn on longer envelope times.

#### TODO
 * The upgrade path isn't 'tuned in' just yet.
 * The check box context menu lacks a 'help' button which this knob would benefit from.
 * One could also scrap the check box and just tweak the knob value range to 0 - 1.0 .

@jasp00 You suggest here to make the envelope knobs response linear. If this is still a plan now would probably be a good time.
https://github.com/LMMS/lmms/blob/master/src/core/EnvelopeAndLfoParameters.cpp#L407


zonkmachine:
I bump the buffer amount to 1024. It's the fastest way to reduce the risk of running out of buffers like in https://github.com/LMMS/lmms/issues/3223 . You could go higher but there is a time penalty to it. I haven't looked at precisely how much the cost in CPU cycles is to do this. From my testing, pretty much negligible.

jasp00:
> You suggest here to make the envelope knobs response linear. If this is still a plan now would probably be a good time.

It is still a plan because it is an optimization to avoid the calculations from `expKnobVal()`; for a (-1.0, 2.0) range, it means to switch to (-1.0, 4.0) and upgrade old data. This can be done in a different PR.

>I never liked the 0 - 2.0 value for the envelope times. This fixes it right up and instead adds a check box to turn on longer envelope times.

Then the knob is probably not the proper widget.

Jousboxx:
I think I have found a bug: When I open a project file that was made using normal master LMMS, the envelopes on my instruments are way too long. I tried a few different project files and I've seen this behavior on all of them. When I made a project using my build of this pull request, it saved and reopened properly. Here's links to a video of this behavior and the MMPZ. 

https://drive.google.com/open?id=0B8dZ3IVOKKA6VXN2aTdzT0JkSGc

lmms.io/lsp/?action=show&file=10488

Interesting feature, could be very useful :)

grejppi:
The choice between 1x and 5x feels arbitrary to me. What if it was an LCD spin box with a range of 1-5 instead?

zonkmachine:
I'm closing for the time being.

zonkmachine:
PS. Thanks for the reviews.

PR #3381, opened 2017-02-23
Envelope Generator - Don't divide by zero

zonkmachine:
Some think it's rude so I cleaned up `EnvelopeAndLfoParameters.cpp`  a bit. I caught this because the new reverb went NaN and things shut down when I turned up the decay.
I made `const f_cnt_t minimumFrames` and didn't touch the knob values even though you could give them saner minimum values but you would still need to set some of the f_cnt_t variables in `EnvelopeAndLfoParameters::updateSampleVars()` so I went this way instead.

In the LFO `m_lfoAttackFrames` also use `minimumFrames`.

### DIFF
--- a/src/core/EnvelopeAndLfoParameters.cpp
+++ b/src/core/EnvelopeAndLfoParameters.cpp
@@ -34,6 +34,8 @@
 extern const float SECS_PER_ENV_SEGMENT = 5.0f;
 // how long should be one LFO-oscillation maximal?
 extern const float SECS_PER_LFO_OSCILLATION = 20.0f;
+// minimum number of frames for ENV/LFO stages that mustn't be '0'
+const f_cnt_t minimumFrames = 1;
 
 
 EnvelopeAndLfoParameters::LfoInstances * EnvelopeAndLfoParameters::s_lfoInstances = NULL;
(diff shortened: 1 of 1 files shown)


zonkmachine:
> removed from In Progress in Release 1.2.0 RC3

Removing from RC3 project as this is nothing new and needs more time.

Umcaruje:
@zonkmachine is this targeted for 1.2?

zonkmachine:
> @zonkmachine is this targeted for 1.2?

No. I don't think we should do any changes for 1.2 concerning NaN's and 0 divisions etc. It's not one bug and it's not introduced in 1.2 . It sucks a bit but I won't have time to work on this now.

zonkmachine:
> It's not one bug...

I had a second go at this as #3687 mad this so much easier. The ones in the envelope seems clearly defined and straight forward to just give a minimum frame of 0.

This one is ready to be merged into master. It should clear up a couple of the NaN oriented bugs for sure.
It also makes debugging with the new `WANT_DEBUG_FPE` flag more practical as you currently can't start lmms with any instruments because it will give a floating point error as soon as 
`EnvelopeAndLfoParameters::updateSampleVars()` is run.

zonkmachine:
Some final adjustments. To test this commit I suggest compiling with and without `WANT_DEBUG_FPE` to see the difference.

I intend to merge this in a day.

PR #3382, opened 2017-02-23
Include past automation patterns in processing

### LINKED ISSUE #662: Make automation read back to the last automation's clip last value
In Ardour, no matter where you play your track, it's always updating all automated values according to what was the last value in the automation track.

I know Ardour doens't use "clips" and each track has only one assigned parameter, but I think this still could be implemented in LMMS, to some nice effect, and less work needed for extending the clips manually just to get a consistent output.

(Real Life) Example:

My track has a HighPass filter on the Master bus.
In the very beginning the filter is automated to sweep from 500 Hz to 0 Hz.
If you however play the begginning of the track (and the filter's knob is set to 500 Hz) and jump to the middle of the track (a few bars after the automation clip ends), the filter won't update to the last value of automation clip, it'll stay where the pleayhead has last read.

I'd like LMMS' automation to "look back" to the last value of the previeos clip, avery time the playhad is manually moved by the user, this way the automated values would be in sync to what you'd get when you render the track from start to finish.

Lukas-W:
Closes #662.

lmms-662

I think that ideally, automated values should update when setting the playing position, not just when playing. Opinions on that?


tresf:
Holy mackerel! 🎉

zonkmachine:
Tested. :+1: 

> I think that ideally, automated values should update when setting the playing position, not just when playing. Opinions on that?

Probably yes to this but there are other cases where values aren't updated until you press play. For instance when you import a midi file and it's tempo isn't updated until you play.

Lukas-W:
Unfortunately this PR causes an issue where some automation patterns are ignored, because automation tracks are processed in the order they appear in the song editor. See the following screenshot, where the second track overwrites the first one:

image


> For instance when you import a midi file and it's tempo isn't updated until you play.

Isn't that just a global automation track on the tempo? Doesn't sound like desired behavior to me.

Lukas-W:
> some automation patterns are ignored

Fixed via c42eb2e.

I did a quick test implementing "Updating when setting the position", but I can't seem to find a good place in the code to insert the updating procedure at.
lmms-662-2

Lukas-W:
Do we want to include this PR in 1.2?

Umcaruje:
>Do we want to include this PR in 1.2?

I really like this feature, but we should be in a feature freeze, so not really sure. Interested in what others have in mind. I'm fine either way really.

tresf:
> Do we want to include this PR in 1.2?

It's a pretty easy fix to revert if it causes problems.  If someone can provide some testing, that would certainly help.  @musikBear is usually good for this if we can get him a win32 installer.

musikBear:
@Lukas-W Marvellous 🎆

musikBear:
@Lukas-W Marvellous 🎆  
That one is indeed a good feature.
If its decided that it goes past the freeze, and into 1.2, just '@' me a notification for the test-file

include/Song.h
```diff
@@ -97,6 +97,7 @@ class EXPORT Song : public TrackContainer
 
 
 	void processNextBuffer();
+	void processAutomations(const TrackList& tracks, MidiTime timeStart, fpp_t frames, int tcoNum);
```
jasp00: Method should be private.

include/Track.h
```diff
@@ -146,6 +146,8 @@ class TrackContentObject : public Model, public JournallingObject
 		return m_selectViewOnCreate;
 	}
 
+	/// Returns true if and only if a->startPosition() < b->startPosition()
+	static bool comparePosition(const TrackContentObject* a, const TrackContentObject* b);
```
jasp00: This is only needed inside `Song` and uses a public `TrackContentObject` function. It should a static function in `Song.cpp`.

src/core/Song.cpp
```diff
@@ -415,6 +409,68 @@ void Song::processNextBuffer()
...
+			tco_begin = tcoNum;
+			tco_end = tcoNum + 1;
+		}
+
+		for (int i = tco_begin; i < tco_end; i++)
+		{
+			TrackContentObject* tco = track->getTCO(i);
+
+			// Skip this TCO if it's muted or not contained in the time period we're processing
+			if( tco == NULL || tco->isMuted() || tco->startPosition() > time_end)
```
jasp00: `tco == NULL` is an error. It should not be hidden.

include/Song.h
Lukas-W: I'm going to add a unit test for this feature, so I'll leave it public. We can gladly refactor to split this into a public and a private part some day.

include/Track.h
Lukas-W: We may end up using it elsewhere. Also, it's quite obviously a function related to `TrackContentObject`, so I don't think moving it to a mostly unrelated file is a good idea. Whether `TrackContentObject` should know it's own position is debatable of course, but that's how it is right now.

src/core/Song.cpp
Lukas-W: I agree, but I didn't write that part, I just moved it and didn't dare remove check introduced by `@tobydox` 9 years ago. Feel free to add something like a `qCritical` message or to just remove the check.

include/Song.h
jasp00: If this is really private, you [should not](http://stackoverflow.com/a/7394311) be testing private methods. You should publish when necessary.

include/Track.h
jasp00: I do not agree, but your choice is not bad.

include/Song.h
Lukas-W: > you should not be testing private methods

Good thing this one's public then. 😉 As I said, we can refactor this some day, but it's out of scope for this PR.

jasp00: Why cannot `processAutomations()` be private until there is a real need to be public? May I do the cut+paste?

Lukas-W:
Added tests and some refactorizations, which also makes `processAutomations` private, @jasp00.

jasp00:
Good. Could you add the scope trick for `automatedValuesAt()`? Something like `privateAndTest` that is defined to `private` normally and to `public` on tests.

Lukas-W:
> Could you add the scope trick for automatedValuesAt()?

What? Why?

Lukas-W:
I think it's better to not merge this with 1.2. It may be a low risk, but should take the term "feature freeze" seriously if we want to have a working 1.2 release any time soon.
I'll create a new `stable-1.2` branch, then merge this into `master`.

jasp00:
Perhaps you should not have added so many changes to the PR, but this is a bug fix. I do not know why #662 is tagged as "enhancement". After all your work, if you are not going to finish, I will.

Lukas-W:
It may be a bug fix from the user perspective, but from the code perspective, it can be considered a rewrite of automation processing. So I suggest not merging this without extensive testing, which is better done in part of the next release so we don't hold up this one even longer.

> if you are not going to finish, I will.

What's not finished here?

tresf:
> I do not know why #662 is tagged as "enhancement". After all your work, if you are not going to finish, I will.

I think @jasp00's point here (hard to read between the lines) is that if this is a bug, it can be merged as a bug-fix after the freeze... whereas as an enhancement, it would wait until the next release.

In regards to "why" this is an enhancement (I labeled it as such), it is because the automation behavior was behaving correctly, just not intuitively.  e.g. the old behavior was not a bug and it was not wrong.

From a delivery perspective, this feature adds a tremendous amount of value to 1.2 so if we could get a precompiled version out for our testers, that could certainly influence the candidacy.

Lukas-W:
Here you go: [lmms-win32-pr-3382-g429da920c.zip](https://github.com/LMMS/lmms/files/865270/lmms-win32-pr-3382-g429da920c.zip)

Lukas-W:
Updated: [lmms-win32-pr-3382-v1.2.0-rc2-113-ge2808837d.zip](https://github.com/LMMS/lmms/files/865536/lmms-win32-pr-3382-v1.2.0-rc2-113-ge2808837d.zip)

tresf:
Testing feedback from @Specular:

> @tresf 26 automation tracks and no playback issues. Tested reading backwards on a few and it works. :ok_hand:

On a side note, there are some items that are being duplicated, namely all instruments as well as the menu item for LADSPA Plugin browser.  I can't see it being related, but wanted to relay that feedback.

Lukas-W:
> On a side note, there are some items that are being duplicated

This probably surfaced only now because plugin search paths were broken until 7251c84f1c22b9803de067cae2a1012d2112af26. Fixed via 0d77cefc13700fedc53d5bf908f7ddc5745140fb.

So it sounds like no one's objecting to having this in 1.2 (except me). Merging 👍

jasp00:
>> Could you add the scope trick for `automatedValuesAt()`?
>What?

```c++
privateAndTest:
    static AutomatedValueMap automatedValuesAt(const Track::tcoVector& tcos, MidiTime time);
```

>Why?

Because `automatedValuesAt()` should not be used by other classes.

Tests should be under `tests/src`.

Lukas-W:
> Because automatedValuesAt() should not be used by other classes.

I understand what the `private` keyword is for. But there's no point in making this function private and introducing a "scope trick" doesn't exactly encourage good design either. It's a static, read-only function. No damage done with this one public.

> Tests should be under `tests/src`.

They are.

jasp00:
>No damage done with this one public.

How does the "scope trick" discourage good design? It observes encapsulation restrictions. If you publish a private method, other classes may depend on implementation details.

Issue #2688, opened 2016-03-19
Anti-Aliasing in automation editior


TheTravelingSpaceman:
For me (a relatively new user), a immediately recognisable issue is that the interface is not as well rounded as other music production software. The biggest eye-sore was the aliasing of the automation editor graph. It's probably not that noticeable to users with proper HD screens, but since I'm a scrub with a 720p display, it's horrible.

![example](https://cloud.githubusercontent.com/assets/17950515/13899563/e0326612-edf9-11e5-93f0-fcbd1c3e881b.png)

I suggest (going to try and do myself) implementing anti-aliasing for displaying the graphs of the automation editor and wherever else needed.

tresf:
Words like "horrible" are rude and don't help get buy in for features like this.

Also, please stop comparing us to other DAWs.  State the problem, any insight, etc and leave the opinions out please.

I agree, this would be a nice addition.

Best regards.

TheTravelingSpaceman:
Apologies, I was only referring to the way it displays on my screen compared to a proper one(which most people probably have). It wasn't an attempt to insult anyone, just a way of describing an effect exaggerated by my bad screen. I will limit my vocabulary in the future for the more sensitive readers.

musikBear:
:imp: devils lawyer : Will antializing in song-editor cause a higher cpu usage? 
If so, i would say no thank you.

Umcaruje:
No, it shouldn't. We don't redraw the Patterns all the time, only when they are changed, so this shouldn't affect CPU usage at all.

michaelgregorius:
Fixed with #2694.

zonkmachine:
![hint](https://cloud.githubusercontent.com/assets/6368949/13907271/f02895f4-eeea-11e5-88c7-f9b06787baf7.png)

:+1:

TheTravelingSpaceman:
Seems like there are colour inconsistencies in the shading when the graph changes from a cure to a straight line:
![colors](https://cloud.githubusercontent.com/assets/17950515/13918218/6debc96a-ef6e-11e5-8dd3-a773145ddd2a.png)

tresf:
@TheTravelingSpaceman good catch.  Some of this code has recently changed (f136ba3).  Tagging @Umcaruje in case he has any insight.

If you reset your code base to before f136ba3 (i.e. `git reset --hard b7d5b2d`) does this strange rendering still occur?

[...]

PR #3385, opened 2017-02-25
[WIP] Fixes Ailiasing in AutomationPatternView

TheTravelingSpaceman:
This is in reference to #2688. This is however not complete since the Automation editor also needs to be Anti-Aliased and there is still a bug:

_The graph/pattern is not being drawn in certain circumstances. It seems to be when it isn't referencing anything but still has a set of data points. The picture below shows the pattern in the editor and right above it it can be seen to not be displaying in the PatternView._
auto

PS: (Please let me know if I'm doing anything in the wrong way. Submitting issues, pull requests etc. This is the first time trying to)

### DIFF
--- a/src/gui/AutomationPatternView.cpp
+++ b/src/gui/AutomationPatternView.cpp
@@ -291,6 +291,7 @@ void AutomationPatternView::paintEvent( QPaintEvent * )
 	lin2grad.setColorAt( 0.5, col );
 	lin2grad.setColorAt( 0, col.darker( 150 ) );
 
+	p.setRenderHints( QPainter::Antialiasing, true );
 	for( AutomationPattern::timeMap::const_iterator it =
 						m_pat->getTimeMap().begin();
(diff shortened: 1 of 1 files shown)


TheTravelingSpaceman:
Should I have posted this before the bug I mentioned has been fixed? I'm not sure if the workflow works so that I need to be 100% confident in a pull request or that it's still acceptable to post a pull request with known errors as a work in progress?

zonkmachine:
> ... or that it's still acceptable to post a pull request with known errors as a work in progress?

This. Now you can get feedback and we know what is being worked on.

TheTravelingSpaceman:
I think I'm closing this pull request because I requested to merge with my master branch which makes it harder to work with on my side... So I created a separate branch on my forked repository and I will create another pull request from that branch... (I don't think there is a way to change this pull request to a different branch on my forked repo)

PR #3386, opened 2017-02-26
Aliasing in AutomationPatternView and AutomationEditor fixed

TheTravelingSpaceman:
Exactly the same as this pull request: #3385 but now from a different branch on my side. (Sorry still learning) It's referencing issue #2688

### DIFF
--- a/src/gui/AutomationPatternView.cpp
+++ b/src/gui/AutomationPatternView.cpp
@@ -293,6 +293,7 @@ void AutomationPatternView::paintEvent( QPaintEvent * )
 	lin2grad.setColorAt( 0.5, col );
 	lin2grad.setColorAt( 0, col.darker( 150 ) );
 
+	p.setRenderHints( QPainter::Antialiasing, true );
 	for( AutomationPattern::timeMap::const_iterator it =
 						m_pat->getTimeMap().begin();
 					it != m_pat->getTimeMap().end(); ++it )
@@ -317,25 +318,37 @@ void AutomationPatternView::paintEvent( QPaintEvent * )
 		float *values = m_pat->valuesAfter( it.key() );
 		for( int i = it.key(); i < (it + 1).key(); i++ )
 		{
-			float value = values[i - it.key()];
+			float x1value = values[i - it.key()];
+			float x2value = values[i - it.key() + 1];
+			if (i+1 == (it + 1).key()) x2value = x1value; //Catches last value not being there
+
 			const float x1 = x_base + i * ppt /
 						MidiTime::ticksPerTact();
 			const float x2 = x_base + (i + 1) * ppt /
 						MidiTime::ticksPerTact();
 			if( x1 > ( width() - TCO_BORDER_WIDTH ) ) break;
 
+			//Creates Polygon
+			QPainterPath path;
+			path.moveTo(QPoint(x1,0));
+			path.lineTo(QPoint(x1,x1value));
+			path.lineTo(QPoint(x2,x2value));
+			path.lineTo(QPoint(x2,0));
+			path.lineTo(QPoint(x1,0));
+
 			if( gradient() )
 			{
-				p.fillRect( QRectF( x1, 0.0f, x2 - x1, value ), lin2grad );
+				p.fillPath(path,lin2grad);
 			}
 			else
 			{
-				p.fillRect( QRectF( x1, 0.0f, x2 - x1, value ), col );
+				p.fillPath(path,col);
 			}
 		}
 		delete [] values;
 	}
 
+	p.setRenderHints( QPainter::Antialiasing, false );
 	p.resetMatrix();
 	
 	// bar lines
--- a/src/gui/AutomationPatternView.cpp
+++ b/src/gui/AutomationPatternView.cpp
@@ -280,7 +280,7 @@ void AutomationPatternView::paintEvent( QPaintEvent * )
 
 	const float y_scale = max - min;
 	const float h = ( height() - 2 * TCO_BORDER_WIDTH ) / y_scale;
-	const float ppTact  = ppt / MidiTime::ticksPerTact();
+	const float ppTick  = ppt / MidiTime::ticksPerTact();
 
 	p.translate( 0.0f, max * height() / y_scale - TCO_BORDER_WIDTH );
 	p.scale( 1.0f, -h );
@@ -301,7 +301,7 @@ void AutomationPatternView::paintEvent( QPaintEvent * )
 	{
 		if( it+1 == m_pat->getTimeMap().end() )
 		{
-			const float x1 = x_base + it.key() * ppTact;
+			const float x1 = x_base + it.key() * ppTick;
 			const float x2 = (float)( width() - TCO_BORDER_WIDTH );
 			if( x1 > ( width() - TCO_BORDER_WIDTH ) ) break;
 			if( gradient() )
@@ -318,19 +318,20 @@ void AutomationPatternView::paintEvent( QPaintEvent * )
 		float *values = m_pat->valuesAfter( it.key() );
 
 		QPainterPath path;
-		QPointF origin = QPointF(x_base + it.key() * ppTact,0.0f);
+		QPointF origin = QPointF(x_base + it.key() * ppTick,0.0f);
 		path.moveTo( origin );
-		path.moveTo( QPointF( x_base + it.key() * ppTact,values[0] ) );
+		path.moveTo( QPointF( x_base + it.key() * ppTick,values[0] ) );
+		float x;
 		for( int i = it.key() + 1; i < ( it + 1 ).key(); i++ )
 		{
-			const float x1 = x_base + i * ppTact;
-			if( x1 > ( width() - TCO_BORDER_WIDTH ) ) break;
+			x = x_base + i * ppTick;
+			if( x > ( width() - TCO_BORDER_WIDTH ) ) break;
 			float value = values[ i - it.key() ];
-			path.lineTo( QPointF( x1,value ) );
+			path.lineTo( QPointF( x, value ) );
 
 		}
-		path.lineTo( x_base + ( ( it + 1 ).key() ) * ppTact,values[ ( it + 1 ).key() - 1 - it.key() ] );
-		path.lineTo( x_base + ( ( it + 1 ).key() ) * ppTact,0.0f );
+		path.lineTo( x_base + ( ( it + 1 ).key() ) * ppTick, values[ ( it + 1 ).key() - 1 - it.key() ] );
+		path.lineTo( x_base + ( ( it + 1 ).key() ) * ppTick, 0.0f );
 		path.lineTo( origin );
 
 		if( gradient() )
--- a/src/gui/AutomationPatternView.cpp
+++ b/src/gui/AutomationPatternView.cpp
@@ -319,27 +319,27 @@ void AutomationPatternView::paintEvent( QPaintEvent * )
 
 		QPainterPath path;
 		QPointF origin = QPointF(x_base + it.key() * ppTact,0.0f);
-		path.moveTo(origin);
-		path.moveTo(QPointF(x_base + it.key() * ppTact,values[0]));
-		for( int i = it.key() + 1; i < (it + 1).key(); i++ )
+		path.moveTo( origin );
+		path.moveTo( QPointF( x_base + it.key() * ppTact,values[0] ) );
+		for( int i = it.key() + 1; i < ( it + 1 ).key(); i++ )
 		{
 			const float x1 = x_base + i * ppTact;
 			if( x1 > ( width() - TCO_BORDER_WIDTH ) ) break;
-			float value = values[i - it.key()];
-			path.lineTo(QPointF(x1,value));
+			float value = values[ i - it.key() ];
+			path.lineTo( QPointF( x1,value ) );
 
 		}
-		path.lineTo(x_base + ((it + 1).key()) * ppTact,values[(it + 1).key() - 1 - it.key()]);
-		path.lineTo(x_base + ((it + 1).key()) * ppTact,0.0f);
-		path.lineTo(origin);
+		path.lineTo( x_base + ( ( it + 1 ).key() ) * ppTact,values[ ( it + 1 ).key() - 1 - it.key() ] );
+		path.lineTo( x_base + ( ( it + 1 ).key() ) * ppTact,0.0f );
+		path.lineTo( origin );
 
 		if( gradient() )
 		{
-			p.fillPath(path,lin2grad);
+			p.fillPath( path, lin2grad );
 		}
 		else
 		{
-			p.fillPath(path,col);
+			p.fillPath( path, col );
 		}
 		delete [] values;
 	}
--- a/src/gui/AutomationPatternView.cpp
+++ b/src/gui/AutomationPatternView.cpp
@@ -280,7 +280,7 @@ void AutomationPatternView::paintEvent( QPaintEvent * )
 
 	const float y_scale = max - min;
 	const float h = ( height() - 2 * TCO_BORDER_WIDTH ) / y_scale;
-	const int MidiTpT = MidiTime::ticksPerTact();
+	const float ppTact  = ppt / MidiTime::ticksPerTact();
 
 	p.translate( 0.0f, max * height() / y_scale - TCO_BORDER_WIDTH );
 	p.scale( 1.0f, -h );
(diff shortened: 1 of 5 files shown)


tresf:
> Sorry still learning

No need to apologize.  There's no easy way to switch off of a master branch, this was the right thing to do.  The only portion that may be worth carrying over is any remaining discussion items such as this:

> "After many hours of trying different things I am quite happy with what happens when the following code changes are applied. It solves the anti-aliasing issue and by drawing polygons instead of rectangles and uses the anti aliasing hint. I also made sure there are no shading inconsistencies.
> 
> For some reason however it sometimes doesn't want to draw an automation that is not linked to anything. I believe it is because the values ranges are between 0 and 1 and not 0 and 100 but this is also the case for the master branch and for some reason it works. Can anyone shed some light on it?"

Umcaruje:
>For some reason however it sometimes doesn't want to draw an automation that is not linked to anything. I believe it is because the values ranges are between 0 and 1 and not 0 and 100 but this is also the case for the master branch and for some reason it works. Can anyone shed some light on it?

By changing all the `QPoint`s to `QPointF`s you solve the problem, but a new glitch occurs: 
screenshot from 2017-02-26 17 42 14

There's again transparency in the painter paths from the huge amount of transforming and antialiasing that happens here.

Could we try to push all the values into a singular path and then draw that path?

TheTravelingSpaceman:
Ok awesome! Seems to work now without and shading inconsistencies or leaving out somethings... Now all that needs to be done is the same for the editor itself. @Umcaruje Thanks for pointing out the QPointF thing. It was driving me crazy.

src/gui/AutomationPatternView.cpp
```diff
@@ -313,27 +314,35 @@ void AutomationPatternView::paintEvent( QPaintEvent * )
...
-			if( gradient() )
-			{
-				p.fillRect( QRectF( x1, 0.0f, x2 - x1, value ), lin2grad );
-			}
-			else
-			{
-				p.fillRect( QRectF( x1, 0.0f, x2 - x1, value ), col );
-			}
+		}
+		path.lineTo(x_base + ((it + 1).key()) * ppt / MidiTime::ticksPerTact(),values[(it + 1).key() - 1 - it.key()]);
```
Umcaruje: ` ppt / MidiTime::ticksPerTact()` appears a lot of time in the code, you could make it a seperate variable so we don't calculate this all the time.

TheTravelingSpaceman:
> ppt / MidiTime::ticksPerTact() appears a lot of time in the code, you could make it a seperate variable so we don't calculate this all the time.

Done!

src/gui/AutomationPatternView.cpp
```diff
@@ -278,6 +278,7 @@ void AutomationPatternView::paintEvent( QPaintEvent * )
 
 	const float y_scale = max - min;
 	const float h = ( height() - 2 * TCO_BORDER_WIDTH ) / y_scale;
+	const int MidiTpT = MidiTime::ticksPerTact();
```
zonkmachine: I think what was suggested was to do away with a computation by doing it once, such as:
`const int MidiTpT = ppt / MidiTime::ticksPerTact();`

TheTravelingSpaceman:
Done... I was quite silly there

Umcaruje:
Cool, could you also adjust the spaces in parenthesis to be consistent with the rest of the file? I know this is not required anymore, but it really stands out.

Also would you do the same thing for the automation editor in this PR? So we could kill two birds with one stone

src/gui/AutomationPatternView.cpp
```diff
@@ -278,6 +278,7 @@ void AutomationPatternView::paintEvent( QPaintEvent * )
 
 	const float y_scale = max - min;
 	const float h = ( height() - 2 * TCO_BORDER_WIDTH ) / y_scale;
+	const float ppTact  = ppt / MidiTime::ticksPerTact();
```
Umcaruje: This should actually be named `ppTick`, cause `ppt` already means pixels per tact, and you're dividing it with the number of ticks in a tact :slightly_smiling_face:

src/gui/AutomationPatternView.cpp
```diff
@@ -313,27 +314,35 @@ void AutomationPatternView::paintEvent( QPaintEvent * )
...
 		{
-			float value = values[i - it.key()];
-			const float x1 = x_base + i * ppt /
-						MidiTime::ticksPerTact();
-			const float x2 = x_base + (i + 1) * ppt /
-						MidiTime::ticksPerTact();
+			const float x1 = x_base + i * ppTact;
 			if( x1 > ( width() - TCO_BORDER_WIDTH ) ) break;
+			float value = values[ i - it.key() ];
+			path.lineTo( QPointF( x1,value ) );
```
Umcaruje: There should be a space after the comma: `x1, value`

src/gui/AutomationPatternView.cpp
```diff
@@ -313,27 +314,35 @@ void AutomationPatternView::paintEvent( QPaintEvent * )
...
+		path.moveTo( origin );
+		path.moveTo( QPointF( x_base + it.key() * ppTact,values[0] ) );
+		for( int i = it.key() + 1; i < ( it + 1 ).key(); i++ )
 		{
-			float value = values[i - it.key()];
-			const float x1 = x_base + i * ppt /
-						MidiTime::ticksPerTact();
-			const float x2 = x_base + (i + 1) * ppt /
-						MidiTime::ticksPerTact();
+			const float x1 = x_base + i * ppTact;
```
Umcaruje: as there is no `x2` now, this variable could be called just `x` to avoid confusion.

src/gui/AutomationPatternView.cpp
```diff
@@ -313,27 +314,35 @@ void AutomationPatternView::paintEvent( QPaintEvent * )
...
-			if( gradient() )
-			{
-				p.fillRect( QRectF( x1, 0.0f, x2 - x1, value ), lin2grad );
-			}
-			else
-			{
-				p.fillRect( QRectF( x1, 0.0f, x2 - x1, value ), col );
-			}
+		}
+		path.lineTo( x_base + ( ( it + 1 ).key() ) * ppTact,values[ ( it + 1 ).key() - 1 - it.key() ] );
```
Umcaruje: same here, space after comma

src/gui/AutomationPatternView.cpp
```diff
@@ -313,27 +314,35 @@ void AutomationPatternView::paintEvent( QPaintEvent * )
...
+		path.lineTo( x_base + ( ( it + 1 ).key() ) * ppTact,0.0f );
```
Umcaruje: same as previous comment

src/gui/AutomationPatternView.cpp
TheTravelingSpaceman: What is a tact?

Umcaruje: It's the german/european word for a Bar in music: https://en.wikipedia.org/wiki/Bar_(music)

src/gui/AutomationPatternView.cpp
```diff
@@ -313,27 +314,36 @@ void AutomationPatternView::paintEvent( QPaintEvent * )
 		}
 
 		float *values = m_pat->valuesAfter( it.key() );
-		for( int i = it.key(); i < (it + 1).key(); i++ )
+
+		QPainterPath path;
+		QPointF origin = QPointF(x_base + it.key() * ppTick,0.0f);
```
Umcaruje: Missing spaces in parenthesis (sorry I somehow missed this one)

TheTravelingSpaceman: *( sorry I somehow missed this one )

TheTravelingSpaceman:
Ok so I managed to add anti-aliasing in the Automation Editor with some minor points that need addressing:

- The Shading is inconsistent (Although it was inconsistent in some themes before)

- I'm not using  the is_selected Boolean (Can't seem to find it's use)

- I don't catch the points if they aren't in the "Visible area" as was done before

This is the result thus far:
before
After:
after

Keep in mind this is at extreme zoom in

tresf:
@TheTravelingSpaceman @Umcaruje that shading won't ever (EVER) line up unless we either set a fixed gradient (yuck).

Can we just abolish it?  I see this is being done against the old theme, has the new theme gotten rid of the gradient or is it still there?

TheTravelingSpaceman:
> Can we just abolish it?
Are you talking about the gradient?

The new theme works just fine since it doesn't use a gradient.
image

tresf:
> The new theme works just fine

👍

Umcaruje:
>The new theme works just fine since it doesn't use a gradient.

Actually it still does use one, but I agree this can be removed really. @RebeccaDeField thoughts?

src/gui/editors/AutomationEditor.cpp
```diff
@@ -1308,12 +1308,26 @@ void AutomationEditor::paintEvent(QPaintEvent * pe )
 				}
 
 				float *values = m_pattern->valuesAfter( it.key() );
-				for( int i = 0; i < (it+1).key() - it.key(); i++ )
-				{
 
-					drawLevelTick( p, it.key() + i, values[i],
-									is_selected );
+				float nextValue;
+				if ( m_pattern->valuesAfter( ( it + 1 ).key() ) != NULL )
```
Umcaruje: You should have brackets for the if/else statement, and there should be no space between `if` and the parenthesis: `if(...`

src/gui/editors/AutomationEditor.cpp
```diff
@@ -1308,12 +1308,26 @@ void AutomationEditor::paintEvent(QPaintEvent * pe )
...
+				path.moveTo( QPointF( xCoordOfTick( it.key() ), yCoordOfLevel( 0 ) ) );
+				for( int i = 0; i < ( it + 1 ).key() - it.key(); i++ )
+				{	path.lineTo( QPointF( xCoordOfTick( it.key() + i ), yCoordOfLevel( values[i] ) ) );
+					//drawLevelTick( p, it.key() + i, values[i], is_selected );
+					
 				}
+				path.lineTo( QPointF( xCoordOfTick( ( it + 1 ).key() ), yCoordOfLevel( nextValue ) ) );
+				path.lineTo( QPointF( xCoordOfTick( ( it + 1 ).key() ), yCoordOfLevel( 0 ) ) );
+				path.lineTo( QPointF( xCoordOfTick( it.key() ), yCoordOfLevel( 0 ) ) );
+				p.fillPath( path, graphColor() );
```
Umcaruje: The `is_selected` variable was used when a particular automation point was selected, but this is disabled and broken atm, it would requre some CSS changes to be properly implemented, and we can save that for when that feature gets fixed.

RebeccaDeField:
> Actually it still does use one, but I agree this can be removed really. @RebeccaDeField 

The gradient just for a touch of depth and it can definitely be flattened. I do think it's important to keep some transparency so that the grid can show through for usability. I'm thinking we could give the whole automation pattern the same transparency as the top stop.

src/gui/editors/AutomationEditor.cpp
TheTravelingSpaceman: So can I delete the  code referencing the is_selected variable and related code?

src/gui/editors/AutomationEditor.cpp
TheTravelingSpaceman: I thought it looked neater without the brackets but I'll add them to make it consistent?

TheTravelingSpaceman:
> I'm thinking we could give the whole automation pattern the same transparency as the top stop.

What's the top stop? 

And is it a problem that the Travis-CI failed it's tests?

tresf:
> What's the top stop?

I believe @RebeccaDeField is recommending to use the top-most automation point as the gradient end point for the entire section so that they draw without segmentation.

RebeccaDeField:
It is currently a gradient with two stops (top and bottom). If we wanted to take away the gradient, we would only have one shade of transparency. I'm suggesting that we use the transparency of [`stop:0`](https://github.com/LMMS/lmms/blob/master/data/themes/default/style.css#L26) which seems to be 180.

Sorry that was confusing :sweat_smile:

TheTravelingSpaceman:
@RebeccaDeField  So do we change that in the CSS or the c++ code? I can't seem to find the right place to limit it in the c++ code :/

Umcaruje:
@TheTravelingSpaceman I think it would be wiser to actually comment out the code rather than remove it, cause we want that feature back sometime.

TheTravelingSpaceman:
So? What's the next step?

TheTravelingSpaceman:
@Umcaruje @RebeccaDeField  @tresf

RebeccaDeField:
@TheTravelingSpaceman Sorry for the late reply, this somehow got buried in my inbox. You will need to remove the gradient in the [style.css](https://github.com/LMMS/lmms/blob/master/data/themes/default/style.css#L26) file if that is what you're asking. :blush:

TheTravelingSpaceman:
@Umcaruje Should be fine to commit now

zonkmachine:
Pretty! I've tested this and it works really well. :+1:

TheTravelingSpaceman:
Cool let's merge then @zonkmachine @tresf @Umcaruje

zonkmachine:
@Umcaruje merge?

Umcaruje:
>@Umcaruje merge?

I'll test today and merge if all is well.

src/gui/editors/AutomationEditor.cpp
Umcaruje: You're still missing `{}` brackets here

src/gui/editors/AutomationEditor.cpp
Umcaruje: It's good when it's commented out as now.

Umcaruje:
Did another test. It looks really smooth, but you're off by a pixel when drawing in the automation editor: 
screenshot from 2017-03-08 16 43 38

As you see, there's a pixel gap on the bottom.

Also there seems to be some overflowing on the top end of the automation editor, this was probably not introduced with this PR, but it's just something that I've noticed now.
gifrecord_2017-03-08_164846


P.S. Sorry for my late response, life's been busy.

TheTravelingSpaceman:
@Umcaruje I've looked at the two problems you've indicated and can't find a proper way to fix it (That I understand and am confident that it won't break something elsewhere). 

I did "fix" the bottom pixel error in the final commit but if feels a bit hackish. Maybe the person who wrote the code originally should take a look (@jmuzz)

The problems however has nothing to do with the things I've added and can therefore be fixed separately in another PR. I would like to fix it but unfortunately I'm starting school tomorrow.

zonkmachine:
> Did another test. It looks really smooth, but you're off by a pixel when drawing in the automation editor:
> ...
> As you see, there's a pixel gap on the bottom.

But this gap is there already in 1.0.3

> I did "fix" the bottom pixel error in the final commit but if feels a bit hackish.

Works, but maybe not if fixing the top margin error too.

> Also there seems to be some overflowing on the top end of the automation editor, this was probably not introduced with this PR, but it's just something that I've noticed now.

https://github.com/LMMS/lmms/blob/master/include/AutomationEditor.h#L176
If we don't take the bottom gap into account, Changing `TOP_MARGIN` value to ~20 seem to take care of this. I think rather the theming folks needs to take a look at this as it's a separate issue.

Umcaruje:
>But this gap is there already in 1.0.3

Huh, I never noticed that. Guess I never looked hard enough. Anyways @TheTravelingSpaceman the last commit doesn't fix it completely. I suggest you revert it, we merge this and open a seperate issue for the top margin and the pixel gap.

TheTravelingSpaceman:
Reverted as requested. Should be fine to merge now

zonkmachine:
> Reverted as requested. Should be fine to merge now @Umcaruje

I'm pretty sure he meant reverting  940c766b616dbfccdd09ef233ac4898a1674d08f

TheTravelingSpaceman:
@zonkmachine I thought that was what I was doing. Was under the impression that the command reverts to the state that the commit was at that point. So just making sure now... I need to execute the following code to have it right:
`git revert 4e127a7`
`git revert 940c766`

zonkmachine:
> `git revert 4e127a7`
> `git revert 940c766`

Yes, this works.

For the future. I usually do `rebase -i` and remove the lines of the commits I want removed as it's neater ( first back up! ) .

tresf:
`revert` reverts a single commit.  `reset --hard` rewinds back in time to a particular commit.

Umcaruje:
:+1:

RebeccaDeField:
Nice work @TheTravelingSpaceman! :grin:

PR #3387, opened 2017-02-26
Barlines now end at end of automation

TheTravelingSpaceman:
The barlines in the Automation pattern view ended "early" where the data points ended for the pattern. It looks more consistent to end wherever the pattern is dragged to.

Before:
before

After:
after

I'm not sure that the limit I changed is the right limit. It's a good idea for someone who knows exactly what each variable represents to have a look over. It does seem to work and not cause any problems.

### DIFF
--- a/src/gui/AutomationPatternView.cpp
+++ b/src/gui/AutomationPatternView.cpp
@@ -344,7 +344,8 @@ void AutomationPatternView::paintEvent( QPaintEvent * )
 	{
 		const int tx = x_base + static_cast<int>( ppt * t ) - 2;
 		p.drawLine( tx, TCO_BORDER_WIDTH, tx, TCO_BORDER_WIDTH + lineSize );
-		p.drawLine( tx,	rect().bottom() - ( lineSize + TCO_BORDER_WIDTH ), tx, rect().bottom() - TCO_BORDER_WIDTH );
+		p.drawLine( tx,	rect().bottom() - ( lineSize + TCO_BORDER_WIDTH ),
+					tx, rect().bottom() - TCO_BORDER_WIDTH );
 	}
 
 	// recording icon for when recording automation
--- a/src/gui/AutomationPatternView.cpp
+++ b/src/gui/AutomationPatternView.cpp
@@ -343,12 +343,8 @@ void AutomationPatternView::paintEvent( QPaintEvent * )
 	for( tact_t t = 1; t < width() - TCO_BORDER_WIDTH; ++t )
 	{
 		const int tx = x_base + static_cast<int>( ppt * t ) - 2;
-		if( tx < ( width() - TCO_BORDER_WIDTH * 2 ) )
-		{
-			p.drawLine( tx, TCO_BORDER_WIDTH, tx, TCO_BORDER_WIDTH + lineSize );
-			p.drawLine( tx,	rect().bottom() - ( lineSize + TCO_BORDER_WIDTH ),
-						tx, rect().bottom() - TCO_BORDER_WIDTH );
-		}
+		p.drawLine( tx, TCO_BORDER_WIDTH, tx, TCO_BORDER_WIDTH + lineSize );
+		p.drawLine( tx,	rect().bottom() - ( lineSize + TCO_BORDER_WIDTH ), tx, rect().bottom() - TCO_BORDER_WIDTH );
 	}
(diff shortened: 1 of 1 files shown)


src/gui/AutomationPatternView.cpp
```diff
@@ -340,7 +340,7 @@ void AutomationPatternView::paintEvent( QPaintEvent * )
 	const int lineSize = 3;
 	p.setPen( c.darker( 300 ) );
 
-	for( tact_t t = 1; t < m_pat->timeMapLength().getTact(); ++t )
+	for( tact_t t = 1; t < width() - TCO_BORDER_WIDTH; ++t )
 	{
 		const int tx = x_base + static_cast<int>( ppt * t ) - 2;
 		if( tx < ( width() - TCO_BORDER_WIDTH * 2 ) )
```
zonkmachine: Perhaps it's possible to remove this if( ... ) since the tests now look similar?

TheTravelingSpaceman:
Question on rebasing: 
So there is a new commit that has recently been merged to lmms/master. This causes my entire forked repo to be 4 commits behind. So to fix this I run `git pull upstream master` to update my master branch. This updates my master branch to be in sync with the master branch of the main repo. 

I have, however, been working on this branch "Barlines", which is not updated. I know it doesn't really matter for this situation but it might be needed in the future to make sure the branch I'm working on is compatible with the new merges. So from what I understand, I should "rebase" the branch. So I run the following code: `git checkout Barlines && git rebase master`. This all goes smoothly but when I want to sync my local branch with my personal repo there are some issues. 

I then run `git push origin master` to update my remote repo with the new merge. All goes well and I can even verify on the github website that my master branch is now up to date. When I run `git push origin Barlines` I get the following error:

>  ! [rejected]        Barlines -> Barlines (non-fast-forward)
error: failed to push some refs to 'git@github.com:TheTravelingSpaceman/lmms.git'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Integrate the remote changes (e.g.
hint: 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.

Umcaruje:
>When I run git push origin Barlines I get the following error:

After rebasing you need to force push, since the branch can't be fast forwarded: `git push -f origin Barlines`

Protip: to avoid always stating which branch you push to, push once with the `--set-upstream` option: `git push --set-upstream origin Barlines` Now, next time you push you can just say `git push` and it will always push to that branch :smile_cat:

Umcaruje:
Also, unless the new change makes a merge conflict on your branch, there is generally very little need for rebasing your Pull Request, as an outdated PR when merged won't actually revert the new changes that are not on the branch, even though it might look like something like this could happen.

TheTravelingSpaceman:
So is this ready for a merge?

src/gui/AutomationPatternView.cpp
```diff
@@ -340,15 +340,11 @@ void AutomationPatternView::paintEvent( QPaintEvent * )
...
 	{
 		const int tx = x_base + static_cast<int>( ppt * t ) - 2;
-		if( tx < ( width() - TCO_BORDER_WIDTH * 2 ) )
-		{
-			p.drawLine( tx, TCO_BORDER_WIDTH, tx, TCO_BORDER_WIDTH + lineSize );
-			p.drawLine( tx,	rect().bottom() - ( lineSize + TCO_BORDER_WIDTH ),
-						tx, rect().bottom() - TCO_BORDER_WIDTH );
-		}
+		p.drawLine( tx, TCO_BORDER_WIDTH, tx, TCO_BORDER_WIDTH + lineSize );
+		p.drawLine( tx,	rect().bottom() - ( lineSize + TCO_BORDER_WIDTH ), tx, rect().bottom() - TCO_BORDER_WIDTH );
```
zonkmachine: The old format of this line, broken into two lines, is more along the recommended 80 character lines width in the [coding conventions](https://github.com/LMMS/lmms/wiki/Coding-conventions). I suggest you change this back to the way it was ( but with `tx, ` 5 tabs in... ) .

TheTravelingSpaceman:
@zonkmachine Fixed! @Umcaruje

zonkmachine:
Tested and looks fine, merging!

Issue #3389, opened 2017-02-28
Light and Dark patches in Piano roll and automation editor


TheTravelingSpaceman:
The alternate bars in the piano roll and automation editor is slightly differently shaded. This is to easily identify the size of a bar. The shading however moves relative to the bars when zoomed in or out (And then panned to the left and right)

Correct shading: (Lines up with the bars)
![image](https://cloud.githubusercontent.com/assets/17950515/23409924/129718e0-fdd6-11e6-8209-c75bfdeaae24.png)

Incorrect shading:
![image](https://cloud.githubusercontent.com/assets/17950515/23409960/323f13c8-fdd6-11e6-9618-f5dccddb4a4c.png)

zonkmachine:
@Umcaruje

PR #3391, opened 2017-03-01
fix alternate shading of pianoroll and automation editor on zooming

BaraMGB:
fixes https://github.com/LMMS/lmms/issues/3389


Umcaruje:
@BaraMGB you're awesome

PR #3399, opened 2017-03-04
Haiku update

korli:
The first commit from waddlesplash allows to run as root.
The two others commits fix the build on Haiku, though the one with unistd.h isn't specific to Haiku.


tresf:
Why would you want to run as root?

Also, we have an override flag that can be used, does it not work?

korli:
Running as root is the default on Haiku. I can remove that commit if you prefer.

tresf:
If that is the norm, OK, then people can run it by default, it is just strange to read as a Linux/Unix user.

korli:
@waddlesplash your opinion on the "Allow running as root on Haiku." commit?

waddlesplash:
@tresf Haiku has POSIX multiuser, but the GUI all runs as root for BeOS compatibility. That won't change till after R1, most likely.

So keeping that change in there makes sense to me.

tresf:
> So keeping that change in there makes sense to me.

👍

tresf:
@korli for consistency can we switch the statement to `defined(LMMS_BUILD_HAIKU)`

https://github.com/LMMS/lmms/blob/master/include/lmms_math.h#L37

korli:
It's now updated.

tresf:
> It's now updated.

Thanks.  Will merge once Travis succeeds.

Issue #3402, opened 2017-03-05
Make the commandline renderer work better


PaulBatchelor:
Hello LMMS-ers,

It seems pretty clear to me based on issues like #3396 #3207 #2934 #2845 #3360 that there are fundamental issues with the LMMS audio engine, particularly when it comes to rendering audio to disk.

It is hope that one of my next contributions to LMMS will a step forward in solving the issues above by improving the command line renderer (aka "lmms -r"). 

To me, a successful CLI renderer is one that can bounce an audio file without any clicks or hiccups, regardless of what audio/MIDI drivers are being used. It would seem that this is not the case right now.  

Based on my initial evaluations of the code, this will involve the following:
- Ensuring that no realtime audio driver starts (JACK/ALSA/PortAudio/CoreAudio etc..). I believe there are threading issues involving JACK and the Renderer class working at the same time
- Ensuring that no realtime MIDI starts up. 
- If possible, remove the use of threads in the render loop (MixerWorkerThread, I'm looking at you)

As I write this, I am painfully aware of that mind-numbing tasks that await me, but this is the sort of thing that must be done. Since I'm still new to the LMMS codebase, I welcome any and all help/guidance on the matter. 

All the best,

-P

PaulBatchelor:
Getting some promising results. Posting my tweaks in the "cli_fixes" branch in my fork of LMMS:

https://github.com/paulbatchelor/lmms/tree/cli_fixes

It looks like I botched up some git rebasing. Don't worry, I plan on redoing my changes in a proper PR ;)

PaulBatchelor:
Closing this to continue conversation in #3404

PR #3404, opened 2017-03-05
Fixes to CLI renderer

PaulBatchelor:
This PR contains some tweaks which address #3402 

This fix should now allow individuals to render audio using "lmms -r" without having to worry about the glitches or pops from before. 

I noticed that calling the CLI renderer was unnecessarily calling JACK, and possibly interfering with the renderer processing thread. What I did was ensure that the audio/MIDI devices were set to "dummy" when renderOnly mode was set to true. After applying this fix, I was able to render offline without having things skip or crash. 

Note that this fix is only for the CLI renderer, and not for the internal LMMS exporter (this still crashes for me).

### DIFF
--- a/include/Mixer.h
+++ b/include/Mixer.h
@@ -157,7 +157,7 @@ public:
 		}
 	} ;
 
-	void initDevices();
+	void initDevices( bool renderOnly );
 	void clear();
 
 
--- a/src/core/Engine.cpp
+++ b/src/core/Engine.cpp
@@ -70,7 +70,7 @@ void LmmsCore::init( bool renderOnly )
 	s_projectJournal->setJournalling( true );
 
 	emit engine->initProgress(tr("Opening audio and midi devices"));
-	s_mixer->initDevices();
+	s_mixer->initDevices( renderOnly );
 
 	PresetPreviewPlayHandle::init();
 	s_dummyTC = new DummyTrackContainer;
--- a/src/core/Mixer.cpp
+++ b/src/core/Mixer.cpp
@@ -203,10 +203,18 @@ Mixer::~Mixer()
 
 
 
-void Mixer::initDevices()
+void Mixer::initDevices( bool renderOnly )
 {
-	m_audioDev = tryAudioDevices();
-	m_midiClient = tryMidiClients();
+	bool success_ful = false;
+	if( renderOnly ) {
+		m_audioDev = new AudioDummy( success_ful, this );
+		m_audioDevName = AudioDummy::name();
+		m_midiClient = new MidiDummy;
+		m_midiClientName = MidiDummy::name();
+	} else {
+		m_audioDev = tryAudioDevices();
+		m_midiClient = tryMidiClients();
+	}
 }
 
 


BaraMGB:
Shouldn't every render process independent of the audio device? No matter of cli or gui?

PaulBatchelor:
Yes, but lmms is still kicking on JACK when invoking the command line renderer. 

The GUI exporter is tougher to fix because ideally you still want to come back to your audio device after rendering, and shutting it down and starting it back up again is clumsy.

jasp00:
You should save `renderOnly` in `Mixer()`, then skip device operations rather than using dummy back ends.

Lukas-W:
Merged via 10e28c26b9527f52c7f049b5cde8c4c98efdf3ea and applied @jasp00's suggestion to save `renderOnly` via 435c898781d724a3535a0b296612bc77fc2cb3d5. Using dummy devices will work just fine. Thanks @PaulBatchelor!

jasp00:
>Using dummy devices will work just fine.

If you set dummy devices, you overwrite the loaded configuration, which we may want to save again for whatever reason; e.g., if a user wants to keep track of recently rendered projects.

Lukas-W:
@jasp00 I'm not sure what you're saying. I tried doing the following:

1. Started LMMS, opened `abc.mmp`, closed LMMS
2. Ran `lmms -r xyz.mmp`
3. Started LMMS

"Recent files" then showed `xyz.mmp` and `abc.mmp` in that order, which is pretty much what I'd expect. Am I getting you wrong?

tresf:
>  you overwrite the loaded configuration

I decoded this as "Now your audio backend in `~/.lmmsrc.xml` is Dummy, dummy! :)  But I too was confused.

Lukas-W:
> Now your audio backend in ~/.lmmsrc.xml is Dummy, dummy! :)

Tried that too but it's not the case, so I guess we'll need to wait for @jasp00 to clarify this :)

jasp00:
I clarify. First, I am not saying that this is not working now, but it is prone to error; the example about the recent list is not relevant. Although it is not documented nor enforced, it is assumed that `audioDevName()` returns the preferred audio device name. This can be seen in [`SetupDialog.cpp`](https://github.com/LMMS/lmms/blob/master/src/gui/SetupDialog.cpp#L815).

If this assumption remains, code is refactored, and device changes are saved, the developer will think that it works, merge, and forget about it. Later, a user files an issue:

Alice: LMMS does not sound anymore.
Bob: What were you doing?
Alice: I rendered a project from the command line.
Bob: You are right. What is going on?
Carol: Do you remember #3404?
Bob: Remember what? Oh, now your back ends are dummy.

These potential side effects should be avoided.

Lukas-W:
No one's making that assumption in the code, and I really don't see how anyone would ever make that assumption. `audioDevName` is used in exactly two places:

1. [src/core/midi/MidiJack.cpp#L73](https://github.com/LMMS/lmms/blob/29f832034ae04301d49b7f2fe087a0bab7ef57aa/src/core/midi/MidiJack.cpp#L73): Here, the audio device's name is checked for being equal to `AudioJack::name()` to re-use the existing connection.
2. [src/gui/SetupDialog.cpp#L811](https://github.com/LMMS/lmms/blob/29f832034ae04301d49b7f2fe087a0bab7ef57aa/src/gui/SetupDialog.cpp#L811), the  line you linked: The audio device's name is only saved as a preference if there is no current preference, which is not a bad idea, and is in no way assuming that `audioDevName()` always returns the preferred device. This is only the case when an user opens LMMS's GUI for the first time ever.

Anyway, via a5aa5e7d08174caf371427b834540212b3251068 I added comments and refactored 1. to not use `audioDevName()`, so only a single usage remains and it's now super-clear that `audioDevName()` returns the *current* audio device's name.

jasp00:
@Lukas-W, you write that `audioDevName()` "Returns the current audio device's name", but this is not true, it just returns `m_audioDevName`. If you change the device via `Mixer::setAudioDevice()`, the variable is not updated.

Lukas-W:
> If you change the device via Mixer::setAudioDevice(), the variable is not updated.

I'd consider that a bug then.

jasp00:
>I'd consider that a bug then.

It is not, you are changing the meaning. I am not saying that you should not change it, but now there is an inconsistency that was not before.

PhysSong:
@Lukas-W I think you should cherry-pick it to stable-1.2 because it is a bug fix.

Lukas-W:
@PhysSong `stable-1.2` already contains this fix, see this PR's commit 5e536319b77548718c6b7bdd770cdbce3f5bcf3b. The branch `stable-1.2` was created after this PR was merged.

PhysSong:
@Lukas-W You're right. I didn't know that because there were changes after that commit.

Issue #3284, opened 2017-01-23
TimeLineWidget overflows


zonkmachine:
With compile option `-DCMAKE_BUILD_TYPE=DEBUG`
If you left-click and drag on the timeline you will eventually get negative numbers. On my machine this happens around ~5m19s at 140bpm. This is a bit early for the counter to overflow. It doesn't happen when playing, only if you click the timeline at a certain intervals. It looks like the computation is done by a too small integer when clicking the timeline.

 To reproduce:
 - On a debug build. `-DCMAKE_BUILD_TYPE=DEBUG`
 - In Song Editor, zoom out to 12.5 or 25%
- Click on the timeline and drag. The time may or may not overflow to a negative number at intervals.

zonkmachine:
Could be this bug in gcc: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=49949

zonkmachine:
@Lukas-W When I first bisected this it pointed to 13357e57c43644cb82413da2d2abac3da1fe4f51 but that didn't make much sense unless it really is a gcc bug and the one I link to above fits the bill. It's a bit on the exotic side but I'm closing this as `not our bug`.

zonkmachine:
I may have closed this prematurely.  I'm trying to alter the optimization flags from `-O2` to something else to see if I can verify the bug. How do you change them? If I search for `-O2` I will find output from cmake like `CMakeFiles/CMakeOutput.log:/usr/bin/cc   -O2 -g -Wall ...` but I can't find the source of it. There is the CMakeCache.txt but if I change the flags in there it doesn't seem to affect the compile in any way.

tresf:
@zonkmachine have you already tried `... -DCMAKE_CXX_FLAGS_DEBUG=-O0 -DCMAKE_C_FLAGS_DEBUG=-O0` ?

Lukas-W:
I can reproduce this with any build type, not just Debug. The gcc bug you linked doesn't look related as it's about complex numbers.

This particular overflow happens at [src/gui/TimeLineWidget#L374](https://github.com/LMMS/lmms/blob/v1.2.0-rc2/src/gui/TimeLineWidget.cpp#L374): 
```cpp
Engine::getSong()->setMilliSeconds(((((t.getTicks()))*60*1000/48)/Engine::getSong()->getTempo()));
//                                     ^~~~~~~~~~~~~~~~~~~~~^
//                                 product too big for 32bit int
```

[...]

PR #3410, opened 2017-03-06
Fix TimeLineWidget overflow

zonkmachine:
In response to #3284


Lukas-W:
I think it would be best to fix this all occurrences (see [comment](https://github.com/LMMS/lmms/issues/3284#issuecomment-274781046)) by providing a function that does this, and using that instead of having the same calculation seven times in a single file.

zonkmachine:
> I think it would be best to fix all occurrences (see comment) by providing a function that does this, and using that instead of having the same calculation seven times in a single file.

OK. I left #3284 open so maybe continue there?

Lukas-W:
Yup.

Issue #3310, opened 2017-01-31
Right Mouse Button To Erase Nodes In Automation Editor Cursor Icon Does Not Change


miketurn:
When you are in the Automation Editor and you hold down the RIGHT mouse button to erase the automation node points, the mouse cursor Pencil icon does not change to an Eraser, like it does in the Piano Roll.

zonkmachine:
Closed in #3414

PR #3414, opened 2017-03-07
Right Mouse Button To Erase Nodes In Automation Editor Cursor Icon...

uro5h:
Right Mouse Button To Erase Nodes In Automation Editor Cursor Icon Does Not Change #3310

### DIFF
--- a/include/AutomationEditor.h
+++ b/include/AutomationEditor.h
@@ -240,6 +240,8 @@ private:
 
 	EditModes m_editMode;
 
+	bool m_mouseDownLeft; //true if left click is being held down
+	bool m_mouseDownRight; //true if right click is being held down
 
 	TimeLineWidget * m_timeLine;
 	bool m_scrollBack;
--- a/include/AutomationEditor.h
+++ b/include/AutomationEditor.h
@@ -240,7 +240,6 @@ private:
 
 	EditModes m_editMode;
 
-	bool m_mouseDownLeft; //true if left click is being held down
 	bool m_mouseDownRight; //true if right click is being held down
 
 	TimeLineWidget * m_timeLine;
--- a/src/gui/editors/AutomationEditor.cpp
+++ b/src/gui/editors/AutomationEditor.cpp
@@ -103,6 +103,8 @@ AutomationEditor::AutomationEditor() :
 	m_y_delta( DEFAULT_Y_DELTA ),
 	m_y_auto( true ),
 	m_editMode( DRAW ),
+	m_mouseDownLeft( false ),
+	m_mouseDownRight( false ),
 	m_scrollBack( false ),
 	m_barLineColor( 0, 0, 0 ),
 	m_beatLineColor( 0, 0, 0 ),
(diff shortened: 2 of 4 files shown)


include/AutomationEditor.h
```diff
@@ -240,6 +240,8 @@ protected slots:
 
 	EditModes m_editMode;
 
+	bool m_mouseDownLeft; //true if left click is being held down
```
Umcaruje: Since you end up only using `m_mouseDownRight` you should remove the `m_mouseDownLeft` as it's not used anywhere.

src/gui/editors/AutomationEditor.cpp
```diff
@@ -1412,7 +1432,17 @@ void AutomationEditor::paintEvent(QPaintEvent * pe )
 	// draw current edit-mode-icon below the cursor
 	switch( m_editMode )
 	{
-		case DRAW: cursor = s_toolDraw; break;
+		case DRAW:
+			if( m_mouseDownRight )
+			{
+				cursor = s_toolErase;
+			}
+			else
```
Umcaruje: You could also add a change for the cursor when automation points are moved, like in the piano roll: https://github.com/LMMS/lmms/blob/85ed63be852d4da2555da49f93d8fd0e18cc3f6c/src/gui/editors/PianoRoll.cpp#L3136-L3152

So we keep consitency.

Umcaruje:
Tested out the PR, works nice, but some changes are needed.

P.S. Lepo je videti da postoje ljudi u Beogradu koji se zanimaju za LMMS :)
(English translation: Nice to see that there are people in Belgrade that take interest in LMMS)

include/AutomationEditor.h
uro5h: Agreed, will remove this one. Also, suggested change(s) might be applicable to PianoRoll as it is never used also.

https://github.com/LMMS/lmms/blob/85ed63be852d4da2555da49f93d8fd0e18cc3f6c/src/gui/editors/PianoRoll.cpp#L1925

(Only line where it is used)

src/gui/editors/AutomationEditor.cpp
uro5h: 👍

uro5h:
Hvala (Thanks), @Umcaruje .

🍻

uro5h:
@Umcaruje,

I have made suggested changes, also removed unused member from `PianoRoll` class. Do you require `squashed` pull requests?

Also there is an issue with triggering `MousePressEvent`, then `MouseReleaseEvent` on automation point (without `MouseMoveEvent`), which doesn't delete the point, but I guess I can file a new issue for it, as I believe it is related to state handling (`m_action`) instead of just cursor switching. What do you think?

Umcaruje:
>Do you require squashed pull requests?

We just squash on merge with Github, so there's no need for you to do it manually.

>Also there is an issue with triggering MousePressEvent, then MouseReleaseEvent on automation point (without MouseMoveEvent), which doesn't delete the point, but I guess I can file a new issue for it, as I believe (based on a quick look) it is related to state handling (m_action) instead of cursor switching, therefore irrelevant for this pull request. What do you think?

Yeah, if that issue is outside of the scope of this PR, feel free to open a new issue for it. Better to have several small PR's than one big one :)

I'll test this tommorow, the code LGTM, if all goes well in testing, I'll merge :+1:

Issue #1693, opened 2015-01-26
Song examples on LMMS


badosu:
Let's use this issue as a placeholder for questions and planning on how we should be able to have songs bundled with LMMS and even if we should have.

Sti2nd:
It could be done as now, pull request by the artist himself or on behalf of others, and then approved by at least three uninvolved persons.

Many have suggested competitions for it too, but I guess the price is somewhat less attractive if you cannot self promote in all text fields you can find? :fish: :cat:  What do you think?

Spekular:
If anyone knows if it's possible to add spoiler tags or otherwise compress text please let me know.

>I’d like to propose multiple licenses.
> ###### CC0: 
>I think CC0 is perfectly suited to samples. Crediting every person who made a sample you use would quickly get out of hand. CC0 would also be suited to presets, but I’m not sure they have any license at the moment? I don’t think it’s an absurd idea that they could/would though.
>Project files as CC0, well. From a user perspective CC0 is the best possible scenario for anything. They can use it however they like with no legal issues, sweet! As an artist though (This is hypothetical, as in *if* I were an artist), I wouldn’t want to give away any serious work. I’m not saying it’s not possible to get CC0 projects, and it’s great if we have them, but maybe not only CC0? I personally would only release as CC0 if I decided on it from the start, as in, ”I’m gonna make a song for other people to use”.
> ###### CC-BY:
>	Any CC-BY samples would have to be darn nice. I can see sample packs or other kind of “luxury” samples requiring attribution though, maybe. If presets had licenses I think CC-BY would work well here. The ones we have now are good and should obviously be CC0/no license, but once again “preset packs”/”luxury” presets or some artist’s signature sound hand crafted presets could be CC-BY perhaps.
>	CC-BY is also where I think a project file’s natural license would be if it’s open. It makes sense to give attribution if you remix or borrow from someone else’s project, I’d consider it courtesy. If maybe Pitbull would do this people wouldn’t think he created the melody in Sugar (Blue (Da Ba Dee)) or the vocal in Good Feeling (Came out suspiciously close after Avicii sampled the same vocal). But enough of Mr 305 Worldwide palabras español y listado ciudades.
> ###### CC-BY-SA: 
Samples and presets? Nope. Projects? Sure, seems fair to me.
> ###### CC-BY-NC: 
Samples/presets maaaybe, projects once again seems fair.
[...]

PR #3418, opened 2017-03-11
Create Jousboxx-BuzzerBeater

Jousboxx:


Umcaruje:
For those out of the loop, this is the winning entry from the Kicker challenge and I think it would be a valid addition to our demos. One thing though, you didn't actually provide the project file, so this PR is invalid. Also what were you thinking for the license of this song. The general consensus of the community on #1693 is that all song examples should be CC0. I personally am ok with any CC license, but CC0 seems to be a requirement for new song additions.

I'm going to close this PR to avoid clutter, but if you need help making a proper pull request please continue the discussion here :)

Jousboxx:
Cool :). I was trying to figure out how to add the project file but then I got sidetracked and forgot. I will try again today! 

As far as the license, I don't really care as long as nobody posts the song as if it is their original work. Remixes, using in YouTube videos, etc. is fine.

PR #3425, opened 2017-03-14
Fix Overflow in Fader.cpp

### LINKED ISSUE #3420: Fader.cpp - Floating point error
As discussed here: https://github.com/LMMS/lmms/pull/3095#pullrequestreview-25678353

We get a floating point error in [Fader.cpp#L387](https://github.com/LMMS/lmms/blob/master/src/gui/widgets/Fader.cpp#L387) and [Fader.cpp#L409](https://github.com/LMMS/lmms/blob/master/src/gui/widgets/Fader.cpp#L409).
```float const persistentRightPeakDBFS = ampToDbfs(m_persistentPeak_R)```
`int m_persistentPeak_R` seem to be overflowing and `ampToDbfs()` mustn't be fed a value of 0 and below.

Possible fix: ```ampToDbfs( qMax<float>( 0.0001, m_persistentPeak_R ) )```

---

```gdb
0x00007ffff47d51ac in feraiseexcept (__excepts=4) at ../sysdeps/x86/fpu/bits/fenv.h:135
135	../sysdeps/x86/fpu/bits/fenv.h: No such file or directory.
(gdb) bt
#0  0x00007ffff47d51ac in feraiseexcept (__excepts=4) at ../sysdeps/x86/fpu/bits/fenv.h:135
#1  __log10f (x=0) at w_log10f.c:32
#2  0x00000000005ec737 in ampToDbfs (amp=0) at /home/zonkmachine/builds/lmms/lmms/include/lmms_math.h:273
#3  0x00000000005ee1eb in Fader::paintDBFSLevels (this=0x1650ef0, ev=0x7fffffffb5f0, painter=...)
    at /home/zonkmachine/builds/lmms/lmms/src/gui/widgets/Fader.cpp:409
```

```gdb
#3  0x00000000005ee070 in Fader::paintDBFSLevels (this=0x16556b0, ev=0x7fffffffbb90, painter=...)
    at /home/zonkmachine/builds/lmms/lmms/src/gui/widgets/Fader.cpp:387
        peak_L = 0
        persistentLeftPeakDBFS = -nan(0x7fbba4)
        width = 11
        center = 12
        maxDB = 9
        leftSpan = 0
        rightSpan = 4.59163468e-41
        drawRectR = {x1 = 0, y1 = 0, x2 = 0, y2 = 0}
        persistentPeak_R = 32767
        peak_R = -18464
        persistentRightPeakDBFS = 4.59163468e-41
        height = 116
        minDB = -42.0000038
        fullSpanReciprocal = 0.019607842
        drawRectL = {x1 = 0, y1 = 116, x2 = 10, y2 = 115}
        persistentPeak_L = -17520
```

Umcaruje:
Fixes #3420

I just implemented @zonkmachine's idea. I don't have the superdebug code set up so testing is appreciated.

### DIFF
--- a/src/gui/widgets/Fader.cpp
+++ b/src/gui/widgets/Fader.cpp
@@ -379,7 +379,7 @@ void Fader::paintDBFSLevels(QPaintEvent * ev, QPainter & painter)
 
 
 	// Draw left levels
-	float const leftSpan = ampToDbfs(m_fPeakValue_L) - minDB;
+	float const leftSpan = ampToDbfs(qMax<float>(0.0001, m_fPeakValue_L)) - minDB;
 	int peak_L = height * leftSpan * fullSpanReciprocal;
 	QRect drawRectL( 0, height - peak_L, width, peak_L ); // Source and target are identical
 	painter.drawPixmap( drawRectL, *m_leds, drawRectL );
@@ -401,7 +401,7 @@ void Fader::paintDBFSLevels(QPaintEvent * ev, QPainter & painter)
 
 
 	// Draw right levels
-	float const rightSpan = ampToDbfs(m_fPeakValue_R) - minDB;
+	float const rightSpan = ampToDbfs(qMax<float>(0.0001, m_fPeakValue_R)) - minDB;
 	int peak_R = height * rightSpan * fullSpanReciprocal;
 	QRect const drawRectR( center, height - peak_R, width, peak_R ); // Source and target are identical
 	painter.drawPixmap( drawRectR, *m_leds, drawRectR );
--- a/src/gui/widgets/Fader.cpp
+++ b/src/gui/widgets/Fader.cpp
@@ -384,7 +384,7 @@ void Fader::paintDBFSLevels(QPaintEvent * ev, QPainter & painter)
 	QRect drawRectL( 0, height - peak_L, width, peak_L ); // Source and target are identical
 	painter.drawPixmap( drawRectL, *m_leds, drawRectL );
 
-	float const persistentLeftPeakDBFS = ampToDbfs(m_persistentPeak_L);
+	float const persistentLeftPeakDBFS = ampToDbfs(qMax<float>(0.0001, m_persistentPeak_L));
 	int persistentPeak_L = height * (1 - (persistentLeftPeakDBFS - minDB) * fullSpanReciprocal);
 	// the LED's have a  4px padding and we don't want the peaks
 	// to draw on the fader background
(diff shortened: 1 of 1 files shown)


zonkmachine:
:-O

```
(gdb) 
(gdb) bt full
#0  0x00007ffff47d51ac in feraiseexcept (__excepts=4) at ../sysdeps/x86/fpu/bits/fenv.h:135
No locals.
#1  __log10f (x=0) at w_log10f.c:32
No locals.
#2  0x00000000005eee07 in ampToDbfs (amp=0) at /home/zonkmachine/builds/lmms/lmms/include/lmms_math.h:273
No locals.
#3  0x00000000005f067f in Fader::paintDBFSLevels (this=0x1b80360, ev=0x7fffffff8e60, painter=...)
    at /home/zonkmachine/builds/lmms/lmms/src/gui/widgets/Fader.cpp:382
        peak_L = 0
        persistentLeftPeakDBFS = -nan(0x7f8e74)
        width = 11
        center = 12
...
```

I didn't get this before but  `ampToDbfs (amp=0)` is because m_fPeakValue_L and m_fPeakValue_R is initialised to 0.0f , [here](https://github.com/LMMS/lmms/blob/master/src/gui/widgets/Fader.cpp#L71:L72) and [here](https://github.com/LMMS/lmms/blob/master/src/gui/widgets/Fader.cpp#L113:L114)... More qMax?
I think this probably depends on https://github.com/LMMS/lmms/pull/3381 for testing apart form the debug code, as envelope floating point errors will shut down LMMS immediately and after this you will get the Fader issue. However, I'm not yet confident with the envelope testing/fix but I'm still surprised by this result. I'll have to poke the envelope some more and get back to you on this.

Umcaruje:
Does this still happen on the Empty template only? or in general?

zonkmachine:
In general. This crash was new so I haven't looked into it that much.
No crash on just loading default.mpt or Empty.mpt .
Open mixer and click 'add channel', boom! It's the initialisation to zero that's fed to `ampToDbfs()`.
I think either we initialise it to something over 0.0f , use the safe version that hands back -inf for 0.0f, or use the same qMax fix. Go with the qMax fix. I think it's most reasonable to go with the same fix.
tested, works... :-)

`float const leftSpan = ampToDbfs( qMax<float>( 0.0001, m_fPeakValue_L ) ) - minDB;`

Umcaruje:
What happens if you change all the instances to SafeAmpToDBFS? do you get the crash?

Umcaruje:
If you don't have time I can do it in the PR too

zonkmachine:
> What happens if you change all the instances to SafeAmpToDBFS? do you get the crash?

The function hands back -inf which creates a crash in the line after.

```
0x00000000005f06b2 in Fader::paintDBFSLevels (this=0x1b866d0, ev=0x7fffffffb470, painter=...)
    at /home/zonkmachine/builds/lmms/lmms/src/gui/widgets/Fader.cpp:383
383		int peak_L = height * leftSpan * fullSpanReciprocal;
```

zonkmachine:
> If you don't have time I can do it in the PR too

:+1:

Umcaruje:
Let's see, I'll make all the values initalize at something other than 0

zonkmachine:
> Let's see, I'll make all the values initalize at something other than 0

I don't think there's any real point doing this other than to show that the value can't be 0 or lower. But it wont really work as a fix since we got negative values, probably from an overflowing int., in the original thread so you'll likely end up having to use qMax anyway.

Umcaruje:
k then, qmax it is :smile:

Umcaruje:
@zonkmachine done. I didn't add the code to the min and max values, they're constants defined in the FXMixerView

PR #3429, opened 2017-03-15
Remove unused includes

Lukas-W:


src/core/main.cpp
```diff
@@ -44,11 +43,7 @@
 #endif
 
 #ifdef LMMS_HAVE_SCHED_H
-#include <sched.h>
-#endif
-
-#ifdef LMMS_HAVE_SYS_TIME_H
-#include <sys/time.h>
+#include "sched.h"
```
jasp00: Why the double quotes?

src/core/midi/MidiSndio.cpp
```diff
@@ -2,8 +2,6 @@
 
 /* license */
 
-#include "MidiSndio.h"
```
jasp00: Mistake.

PR #3432, opened 2017-03-16
Travis speed-ups

Lukas-W:
Reduces running time on Travis by

* Enabling [ccache](https://ccache.samba.org/)
* Caching mingw dependencies from `@tobydox`'s launchpad PPA, because downloads have been quite slow from there

Cuts down running time from the previous ~10-20 minutes to [2-3 minutes](https://travis-ci.org/Lukas-W/lmms/builds/211714144) (best case) for each build except OSX, where installing dependencies can still take around 5 minutes.


tresf:
When I first read this I accidentally thought `ccache` related to the mingw caching, which it is not.

How valuable is [`ccache`](https://ccache.samba.org/performance.html) for our own machines?  Does this add yet another layer of complexity to our build environment when enabled or should we consider having this on by default for all developers?

eagles051387:
Hi Tres, 

I quote the overview from the ccache website. 

ccache is a compiler cache. It speeds up recompilation [1] by caching
previous compilations and detecting when the same compilation is being
done again. Supported languages are C, C++, Objective-C and
Objective-C++. 

What I am not sure is what would need to be done if anything to
integrate ccache. This overall should help speed up compile times for
us. 

--- 

Regards, 

Jonathan Aquilina 

On 2017-03-16 14:24, Tres Finocchiaro wrote:

> When I first read this I accidentally thought ccache related to the mingw caching, which it is not. 
> 
> How valuable is ccache [1] for our own machines? Does this add yet another layer of complexity to our build environment when enabled or should we consider having this on by default for all developers? 
> 
> --
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub [2], or mute the thread [3].
 

Links:
------
[1] https://ccache.samba.org/performance.html
[2] https://github.com/LMMS/lmms/pull/3432#issuecomment-287055229
[3]

.travis/linux.win32.install.sh
```diff
@@ -1,16 +1,30 @@
...
 
-if [ $QT5 ]; then
-	PACKAGES="$PACKAGES mingw32-x-qt5base"
-else
-	PACKAGES="$PACKAGES mingw32-x-qt"
-fi
+export MINGW_PACKAGES
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+. $DIR/linux.win.download.sh win32
```
jasp00: `.` is not necessary, you may run the script in a separate process.

.travis/linux.win.download.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+set -e
+
+CACHE_DIR=$TRAVIS_BUILD_DIR/apt_mingw_cache/$1
+mkdir -p $CACHE_DIR
+
+pushd $CACHE_DIR
+
+apt-get --print-uris --yes install $MINGW_PACKAGES | grep ^\' | cut -d\' -f2 > downloads.list
+wget -N --input-file downloads.list
```
jasp00: `wget` is not secure, you are throwing away authentication. Use `apt-get download`.

CMakeLists.txt
```diff
@@ -477,6 +477,14 @@ ELSE(WIN32)
 	SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC -DPIC")
 ENDIF(WIN32)
 
+# use ccache
+FIND_PROGRAM(CCACHE_FOUND ccache)
+IF(CCACHE_FOUND AND USE_CCACHE)
```
jasp00: You should declare `USE_CCACHE` and test it before looking for ccache.

.travis/linux.win32.install.sh
```diff
@@ -1,16 +1,30 @@
...
+
+PACKAGES="nsis cloog-isl libmpc3 qt4-linguist-tools mingw32 $MINGW_PACKAGES"
 
 sudo apt-get install -y $PACKAGES
+
+# ccache 3.2 is needed because mingw32-x-gcc is version 4.9, which causes cmake
+# to use @file command line passing, which in turn ccache 3.1.9 doesn't support
+pushd /tmp
+wget http://archive.ubuntu.com/ubuntu/pool/main/c/ccache/ccache_3.2.3-1_amd64.deb
```
jasp00: Authentication...

.travis/linux.win32.script.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 
-export CMAKE_OPTS="-DUSE_WERROR=ON"
+export CMAKE_OPTS="-DUSE_WERROR=ON -DUSE_CCACHE=ON"
```
jasp00: This should not be enabled for release builds.

jasp00:
>should we consider having this on by default for all developers?

No, please. If the expected hit rate is low, which would be normal for a developer, ccache is [detrimental](https://ccache.samba.org/performance.html). This does not mean ccache is useless for everyone.

@Lukas-W, I would cache all PPA setup because it may solve some builds that fail to download the GPG key.

.travis/linux.win.download.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+set -e
+
+CACHE_DIR=$TRAVIS_BUILD_DIR/apt_mingw_cache/$1
+mkdir -p $CACHE_DIR
+
+pushd $CACHE_DIR
```
Lukas-W: Why `cd`?

.travis/linux.win.download.sh
Lukas-W: `apt-get download` doesn't download dependencies. Can you explain how I am throwing away authentication?

.travis/linux.win.download.sh
jasp00: One `cd` is simpler than `pushd`/`popd`.

.travis/linux.win32.install.sh
```diff
@@ -1,16 +1,30 @@
...
+	mingw32-x-libgig mingw32-x-libsoundio $MINGW_PACKAGES"
 
-if [ $QT5 ]; then
-	PACKAGES="$PACKAGES mingw32-x-qt5base"
-else
-	PACKAGES="$PACKAGES mingw32-x-qt"
-fi
+export MINGW_PACKAGES
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```
Lukas-W: @tresf introduced this line in 67334a84c270e38a8d5a209970db073445e40bad, I copied it. Maybe he can explain the motivation behind it.

.travis/linux.win.download.sh
jasp00: `wget` does not download dependencies either.

After downloading a package, it should be checked against `Packages`. Read about [secure apt](https://wiki.debian.org/SecureApt).

.travis/linux.win.download.sh
Lukas-W: Ah! I pop back to the previous directory so I don't break scripts running this file.

.travis/linux.win.download.sh
Lukas-W: > wget does not download dependencies either.

But `apt-get --print-uris install <packages>` prints all URIs that need to be downloaded.

> After downloading a package, it should be checked against Packages.

I expect `apt-get` to verify the integrity of packages in the cache prior to installing them. Are you sure it doesn't that?

.travis/linux.win32.install.sh
tresf: > How about using `${TRAVIS_BUILD_DIR}/.travis`?

Knowing the script's location is portable in event  of 1. Refactor or 2. copy to `/tmp`, which I copied from @Lukas-W [here](https://github.com/LMMS/lmms/commit/40af0683280673fe52674d35e2378172a6473f0c#diff-6303d41f3ce806de8cd0b7fd3d2e622dR17).  See what just happened there? 😄

Lukas-W: 😆 I didn't remember that commit. Thanks for explaining it to me.

.travis/linux.win.download.sh
jasp00: >I expect `apt-get` to verify the integrity of packages

You are right, it does. `dpkg` does not.

.travis/linux.win32.install.sh
jasp00: OK.

Lukas-W:
> This should not be enabled for release builds.

Shouldn't be a problem with mingw builds. Any idea for Linux builds though (we may build Linux packages from Travis some day)? When enabling the `ccache` flag in `.travis.yml`, Travis [shadows gcc with ccache versions](https://github.com/travis-ci/travis-build/blob/ef82e82457f82c6518d090544bb533dc0faf0063/lib/travis/build/script/cpp.rb#L14).

jasp00:
>Shouldn't be a problem with mingw builds.

Why not?

Lukas-W:
> Why not?

Because they're using a cross-compiler in a non-standard path (`/opt/mingw` or something like that), which means they're not shadowed by Travis's symlinks. This is the reason I added the `-DUSE_CCACHE` flag in the first place, so we can just disable that on release builds.

Lukas-W:
Disabled ccache in release builds via 3f91f0126e9b5a6902c5a4a2de59970e478fd6e7. Merging

.travis/linux.win32.install.sh
```diff
@@ -1,16 +1,30 @@
...
+
+PACKAGES="nsis cloog-isl libmpc3 qt4-linguist-tools mingw32 $MINGW_PACKAGES"
 
 sudo apt-get install -y $PACKAGES
+
+# ccache 3.2 is needed because mingw32-x-gcc is version 4.9, which causes cmake
+# to use @file command line passing, which in turn ccache 3.1.9 doesn't support
+pushd /tmp
+wget http://archive.ubuntu.com/ubuntu/pool/main/c/ccache/ccache_3.2.3-1_amd64.deb
+sudo dpkg -i ccache_3.2.3-1_amd64.deb
```
jasp00: @Lukas-W, this is a security hazard. Please do something about it.

liushuyu: @jasp00 So you mean the downloaded package doesn't undergo under integrity check and it is vulnerable to MITM attack?

tresf: This is really the only argument, that we're bypassing package validation by installing directly, rather than from `apt`.  There are some utilities to validate signatures on packages but I didn't have luck on 12.04 with any of them... Perhaps 14.04 will be better... 

 * `dpkg-sig --verify foo.deb` -- Did nothing
 * `debsig-verify` -- Dependency conflicts, couldn't install

Lukas-W: I already added a sha check via a9a28a8ea885955dbb580f9a5177d67e3a6cdf34. Sorry for not commenting about it here.

jasp00: It looks like the SHA-256 sum is correct, according to [Packages](http://archive.ubuntu.com/ubuntu/dists/xenial/main/binary-amd64/Packages.gz).

PR #3075, opened 2016-10-13
suggest a fix for precision on knob dragging

BaraMGB:
fixes https://github.com/LMMS/lmms/issues/3074 closed because it's a duplicate of:
fixes https://github.com/LMMS/lmms/issues/3024
~~only for knobs at first.~~

PR #3437, opened 2017-03-19
revert parts of #3075

BaraMGB:
This reverts a part of https://github.com/LMMS/lmms/pull/3075#pullrequestreview-21029234 which introduce a bug in the fader context menu. I wanted to fix the floating point bug (151.2 % -> 1.512 %). But this was the wrong aproach.


jasp00:
Fine, but you could have fixed the original issue too. It is in the same function.

Issue #2869, opened 2016-06-22
Controller will not saved in right order


BaraMGB:
It's a little bit complicated to reproduce. Here's a step by step instruction:
1. new project / add 2 FX channels in Mixer
2. add lfo controller in Controller Rack
3. in BB editor make some steps in kicker and route kicker in FX1
4. add a Peak Controller on FX1
5. connect FX2 Fader with Controller2 (this should be the Peak Controller )
6. connect TripleOSC Volume knob to controller 1 (the LFO from step 2.)
7. save Project / restart LMMS and load Project again

Behavior: the controller connections are in reversed order. The lfo is now connected to the fader and the peak to the volume knob. Also the order in Controller Rack are reverse and both are named "controller1"

Kubuntu 14.04 64bit Qt4 build, latest master.

zonkmachine:
This problem/regression started with PR https://github.com/LMMS/lmms/pull/2222 (35486293e36b4ba211460ed23c6281df2d652524) that caused a crash when loading a project with a Peak Controller in it, reported as Issue #2318 some two weeks later .
#2318 was closed in #2456.
Since we need to repopen a project as devised above to test the bug we can't actually bisect the code from this point to where the crash has been fixed. I'm trying to bisect this further to see if there are any changes in the save files that could hint to the cause but I haven't spotted anything yet.

Related Issues/Pull requests:
#2222 PR (@michaelgregorius)
#2407 Issue: PeakControllerEffect and PeakController circular destruction
#2408 (PR, closed #2407 @grejppi)
#2456 ( PR, fixed crash, closed #2318 @softrabbit)

zonkmachine:
If you save the example project above in master and open it in stable-1.1.3 it will open correctly.

michaelgregorius:
### The problem
Here's what happens:

1. LMMS loads the FX Mixer from the file. Because the Peak Controller is a mix between an FX and a controller it gets loaded as an FX when the mixer "FX 1" is restored. In `PeakControllerEffect::PeakControllerEffect` the Peak Controller is added to the list of controllers by calling `Engine::getSong()->addController( m_autoController )`. Because no controllers have been loaded yet the Peak Controller is now the **first** controller in the list.
2. LMMS loads the controllers and adds them via `Song::addController`. For each controller this method checks whether the controller is already contained in the list and only adds it if it's not:
  * The LFO controller is loaded first from the file. It is added because it is not contained in the list of controllers.
[...]

PR #3438, opened 2017-03-19
fixes the wrong order of controller by loading a project

BaraMGB:
fixes https://github.com/LMMS/lmms/issues/2869

the problem:

> Here's what happens:
> 1. LMMS loads the FX Mixer from the file. Because the Peak Controller is a mix between an FX and a controller it gets loaded as an FX when the mixer "FX 1" is restored. In `PeakControllerEffect::PeakControllerEffect` the Peak Controller is added to the list of controllers by calling `Engine::getSong()->addController( m_autoController )`. Because no controllers have been loaded yet the Peak Controller is now the **first** controller in the list.
> 2. LMMS loads the controllers and adds them via `Song::addController`. For each controller this method checks whether the controller is already contained in the list and only adds it if it's not:
>    - The LFO controller is loaded first from the file. It is added because it is not contained in the list of controllers.
>    - The Peak Controller is now also loaded from the file. It is not added because it is already contained in the list of controllers as it was added during the load of the FX chains.

What I did:

On adding a controller it checks if the controller is in the list. If so it removes the controller from the list and appends it again. So it is in the right order.

```
if( controller )
	{
		if( m_controllers.contains( controller ) )
		{
			int index = m_controllers.indexOf( controller );
			if( index != -1 )
			{
				m_controllers.remove( index );
				emit controllerRemoved( controller );
			}
		}
		m_controllers.append( controller );
		emit controllerAdded( controller );

		this->setModified();
	}
```

### DIFF
--- a/plugins/peak_controller_effect/peak_controller_effect.cpp
+++ b/plugins/peak_controller_effect/peak_controller_effect.cpp
@@ -67,7 +67,10 @@ PeakControllerEffect::PeakControllerEffect(
 	m_autoController( NULL )
 {
 	m_autoController = new PeakController( Engine::getSong(), this );
-	Engine::getSong()->addController( m_autoController );
+	if( !Engine::getSong()->isLoadingProject() )
+	{
+		Engine::getSong()->addController( m_autoController );
+	}
 	PeakController::s_effects.append( this );
 }
 
--- a/src/core/Song.cpp
+++ b/src/core/Song.cpp
@@ -1419,17 +1419,8 @@ void Song::setModified()
 
 void Song::addController( Controller * controller )
 {
-	if( controller )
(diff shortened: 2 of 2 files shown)


BaraMGB:
@zonkmachine @michaelgregorius

zonkmachine:
Tested, works! I've tested complex patterns, both saved from 1.1.3, master and this PR and they now open up correctly.

jasp00:
The patch is incorrect. It changes the meaning of `addController()`. The (understandable) bug is `PeakControllerEffect`'s fault. The constructor should check `Engine::getSong()->isLoadingProject()`.

BaraMGB:
@jasp00 thank you.

PR #3440, opened 2017-03-19
Update MainApplication.cpp

jasp00:
Drop useless condition. Tweaks #3219.


tresf:
Then can we switch to `dynamic_cast`?  This code could be an unsafe cast if non-File events happen.  Advice appreciated.

tresf:
Also, what's the problem with the code comment?

```c++
// Fall through to default
```

This is a safe-guard comment.  Switch-cases that use selective fall-through can be hard to spot or flagged as mistakes when they are not.  This comment was intentional.

tresf:
Doh!  Nevermind.... We're already testing the type implicitly with `case QEvent::FileOpen:`, thus removing the need for non-obvious fall-through.  Merging.

Issue #3351, opened 2017-02-13
LMMS hangs when opening a project


Umcaruje:
I was contacted over discord by Outer Mind, so he sent me a file that makes LMMS hang when opened. The program doesn't even get to the loading dialog, it hangs immideately. 

Here's the file: [Outer_Mind_-_Starlight.mmpz.zip](https://github.com/LMMS/lmms/files/772229/Outer_Mind_-_Starlight.mmpz.zip)

It opens fine on 1.1.3.

Here's a backtrace:
```gdb
Starting program: /home/umcaruje/github/lmmstest/build/lmms 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Gtk-Message: Failed to load module "pantheon-filechooser-module"
[New Thread 0x7fffd9129700 (LWP 25266)]
[New Thread 0x7fffd8928700 (LWP 25267)]
[New Thread 0x7fffd3fff700 (LWP 25268)]
VST sync support disabled in your configuration
[New Thread 0x7fffdb94e700 (LWP 25270)]
[New Thread 0x7fffdb8cd700 (LWP 25271)]
[New Thread 0x7fffa19a0700 (LWP 25272)]
[New Thread 0x7fffa119f700 (LWP 25273)]
[New Thread 0x7fffd8127700 (LWP 25274)]
[New Thread 0x7fff9bfff700 (LWP 25278)]
SuspendRefNum error
JackClient::Execute error name = lmms
[Thread 0x7fffd8127700 (LWP 25274) exited]
[New Thread 0x7fff9b3f1700 (LWP 25279)]
Jack server was closed but clients are still allocated, cleanup...
Cleanup client ref = 6
Server is not running
[Thread 0x7fffdb8cd700 (LWP 25271) exited]
[Thread 0x7fffdb94e700 (LWP 25270) exited]
[New Thread 0x7fffdb94e700 (LWP 25280)]
[New Thread 0x7fffdb8cd700 (LWP 25281)]
[New Thread 0x7fffd91ef700 (LWP 25282)]
^C
Thread 1 "lmms" received signal SIGINT, Interrupt.
0x00007ffff6c989d3 in ?? () from /usr/lib/x86_64-linux-gnu/libQtXml.so.4
(gdb) thread apply all backtrace

Thread 14 (Thread 0x7fffd91ef700 (LWP 25282)):
[...]

PR #3441, opened 2017-03-19
Drop useless loop from upgrade_1_2_0_rc3

jasp00:
Loop with `k` index does not make sense.


jasp00:
Fixes https://github.com/LMMS/lmms/issues/3351#issuecomment-287574705.

zonkmachine:
Tested, works like a charm. Thanks!

PR #3445, opened 2017-03-20
Add COPYING-CMAKE-SCRIPTS

### LINKED ISSUE #3436: Sndio back end license
@devnexen, please clarify the license for the following files:
```
cmake/modules/FindSndio.cmake
include/AudioSndio.h
include/MidiSndio.h
src/core/audio/AudioSndio.cpp
src/core/midi/MidiSndio.cpp
```

jasp00:
Copy license from CMake 2.6.0. Closes #3436.

### DIFF
--- a/cmake/modules/COPYING-CMAKE-SCRIPTS
+++ b/cmake/modules/COPYING-CMAKE-SCRIPTS
@@ -1,0 +1,22 @@
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions
+are met:
+
+1. Redistributions of source code must retain the copyright
+   notice, this list of conditions and the following disclaimer.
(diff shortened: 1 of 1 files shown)


tresf:
For those reviewing this PR wondering the context of why this file is being supplied and why it is named as such, it's an upstream thing.

https://github.com/lmms/lmms/blob/master/cmake/modules/FindAlsa.cmake#L14
https://github.com/lmms/lmms/blob/master/cmake/modules/FindSndio.cmake#L9
https://github.com/lmms/lmms/blob/master/cmake/modules/FindOggVorbis.cmake#L16

I do have concerns about the perceived/implied license of non-licensed files once this is introduced, but that can be handed in a separate discussion.

PR #3446, opened 2017-03-21
Add SF2 support for macOS

### LINKED ISSUE #649: SF2 hangs on open (Apple)
Software freezes and causes the user to "Force Quit" when opening any project with at least one soundfont .
- This crash occurs with SF2 files as relative OR absolute paths.
- This crash occurs with SDL OR PortAudio.
- This crash DOES NOT occur when using Dummy Audio.

The soundfont player works if added manually, but when reopening the project, the software hangs on a dialog that says "Opening project" usually around 70-80% with the spinning wheel which never goes away.

The screenshot below (sorry, not ver descriptive) is the screen that the user will see when the freeze occurs.  The default project is loaded in the background while the "Opening project" screen appears and then the pinwheel appears and the software has crashed.

screen shot 2014-04-26 at 12 25 10 pm

Possibly related:  http://www.juce.com/forum/topic/coreaudio-possible-deadlock-when-moving-mavericks

Backtrace:
https://gist.github.com/tresf/230b20f083403a11a3c7

tresf:
Leveraging Homebrew, we can switch off the problematic `enable-coreaudio` flag by hosting our own [modified `fluid-synth.rb` brew formula](https://github.com/LMMS/lmms/pull/3446/files#diff-8e78be35d8f17194403c91cb8191b525R28).

Closes #649


jasp00:
Could you add license information to `cmake/apple/fluid-synth.rb ` following these [guidelines](https://github.com/LMMS/lmms/wiki/Copyright-information)? Besides that, the patch looks fine.

tresf:
> Could you add license information to cmake/apple/fluid-synth.rb following these guidelines? Besides that, the patch looks fine.

Those guidelines don't apply, it's not our code.  This is a one-liner directly from Homebrew upstream.  I'd like to attach something, but I'd rather not make the diff any larger than one line.  Advice is welcome.

tresf:
> [...] original contributions are assumed to be under GPL-2.0+. However, we must be sure that content under an incompatible license has not been copied from another project

Doesn't apply.... GPL 2.0 is completely valid in this case, since the BSD-2 can be downgraded to GPL 2.0.

> All files should have copyright information. It is general convention to include a full notice, but we only need these lines:

The copyright information is missing from the file, so this is an upstream bug.  Even upstream specified  "Homebrew contributors" which is against the standard you've just created.  This is superfluous.  I'd argue this file isn't even source code.  It's a config file.  There's really no copyrightable logic in it from a legal perspective.

I've added it, but this is nitpicking to a new level.

tresf:
> Besides that, the patch looks fine.

Merging.

jasp00:
Thanks.

>Doesn't apply.... GPL 2.0 is completely valid in this case, since the BSD-2 can be downgraded to GPL 2.0.

@tresf, you know it, I know it, but a newcomer will not.

>I'd argue this file isn't even source code.

Yet there is a class with a definition of bottle, install and test. As I see it, it is not clearly a configuration file. Better safe than sorry.

GamingStar:
I dragged the sf2 player onto the channel rack and the program crashed :/

tresf:
@GamingStar thanks for your unit testing today on [Discord](https://discord.gg/5kSc32Z).    From what I am reading online, `Illegal Instruction: 4` is often caused by either an API compatibility difference or sometimes caused by compiler optimizers.

When we swapped out the fluidsynth library on your machine with one that was built using debug flags (to strip all optimization flags), it still crashed in the same fashion.

Some addition information... You're on MacOS 10.12.  My build machine is on 10.11 and one other community member on Discord was on 10.12 as well however on your machine exhibits this crash.

The most effective way to troubleshoot this crash is to get a better backtrace of the problem, and that is done by creating a debug build and running a debugger against the codebase.

This takes a while to configure, build and then debug but without a machine to reproduce on our hands our tied.

Our apple build tutorial is here: https://github.com/LMMS/lmms/wiki/Compiling-lmms-(Apple)

tresf:
Here's some more information on it... 

https://github.com/muammar/mkchromecast/issues/4#issuecomment-222174145

In the event of `mkchromecast` project, he was able to build from a VM to fix this (this is what we do for stable releases, so this may be a viable option).

This is a lengthier process due to performance issues with macOS in a VM, but I'll begin the process now and link a new installer once completed.

tresf:
@GamingStar can you try [this build](https://github.com/tresf/lmms/releases/download/v1.2.0-rc2-sf2/lmms-1.2.0-rc2-sf2-mac10.11_e.dmg)?

PR #3450, opened 2017-03-22
adds doxygen documentation to SubWindow.cpp

BaraMGB:
I began to write documentation for my work on lmms. Perhaps someone with more expierence on this can guide me a little bit.


jasp00:
You should document public and protected interfaces in the headers. Copyright notices should not be the place for documentation. You should explain what the functions are supposed to do without giving implementation details; documentation is like a contract, try not to describe behavior you may want to change later.

Umcaruje:
Changed the PR to address `stable-1.2`

BaraMGB:
@jasp00 I'll try to make it better. I think I understand what you mean. 

@umcaruje thank you.

Umcaruje:
@BaraMGB is there more work that you want to do on this or can we merge?

BaraMGB:
@Umcaruje yes. But I'm short in time at the moment. But soon it will be better and I can develop again.

tresf:
Bad documentation is better than no documentation.  No need to bike-shed over where it lives.  Doxygen is meant to be exported as an API, so whether this is in a single comment or a proper parsable comment is rather arbitrary until we publish an API for our developers.

Merging.  @BaraMGB can improve this down the road, no PR needed, just commit straight to master. 👍

PR #3455, opened 2017-03-26
Fix scripts using shellcheck

zapashcanon:
Hi,

I ran [shellcheck](https://github.com/koalaman/shellcheck) on the whole repository and fixed each error/warning.

### DIFF
--- a/.travis/linux.win32.install.sh
+++ b/.travis/linux.win32.install.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -e
 
-if [ $QT5 ]; then
+if [ "$QT5" ]; then
 	MINGW_PACKAGES="mingw32-x-qt5base"
 else
 	MINGW_PACKAGES="mingw32-x-qt"
@@ -16,10 +16,13 @@ MINGW_PACKAGES="mingw32-x-sdl mingw32-x-libvorbis mingw32-x-fluidsynth mingw32-x
 export MINGW_PACKAGES
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-. $DIR/linux.win.download.sh win32
+
+# shellcheck disable=SC1090
+. "$DIR/linux.win.download.sh" win32
 
 PACKAGES="nsis cloog-isl libmpc3 qt4-linguist-tools mingw32 $MINGW_PACKAGES"
 
+# shellcheck disable=SC2086
 sudo apt-get install -y $PACKAGES
 
 # ccache 3.2 is needed because mingw32-x-gcc is version 4.9, which causes cmake
--- a/.travis/linux.win64.install.sh
+++ b/.travis/linux.win64.install.sh
@@ -3,10 +3,11 @@ set -e
 
 # First, install 32-bit deps
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-. $DIR/linux.win32.install.sh
+# shellcheck disable=SC1090
+. "$DIR/linux.win32.install.sh"
 
 
-if [ $QT5 ]; then
+if [ "$QT5" ]; then
 	MINGW_PACKAGES="mingw64-x-qt5base"
 else
 	MINGW_PACKAGES="mingw64-x-qt"
@@ -21,6 +22,8 @@ MINGW_PACKAGES="mingw64-x-sdl mingw64-x-libvorbis mingw64-x-fluidsynth mingw64-x
 export MINGW_PACKAGES
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-. $DIR/linux.win.download.sh win64
+# shellcheck disable=SC1090
+. "$DIR/linux.win.download.sh" win64
 
+# shellcheck disable=SC2086
 sudo apt-get install -y $MINGW_PACKAGES
--- a/.travis/osx..install.sh
+++ b/.travis/osx..install.sh
@@ -2,18 +2,19 @@
 
 PACKAGES="cmake pkgconfig fftw libogg libvorbis libsndfile libsamplerate jack sdl stk portaudio node fltk"
 
-if [ $QT5 ]; then
+if [ "$QT5" ]; then
 	PACKAGES="$PACKAGES homebrew/versions/qt55"
 else
 	PACKAGES="$PACKAGES cartr/qt4/qt"
 fi
 
+# shellcheck disable=SC2086
 brew install $PACKAGES ccache
 
 # Recompile fluid-synth without CoreAudio per issues #649
 # Changes to fluid-synth.rb must be pushed to URL prior to use
 url=$(git remote get-url origin)
 branch=$(git symbolic-ref --short HEAD)
-brew install --build-from-source $url/raw/$branch/cmake/apple/fluid-synth.rb
+brew install --build-from-source "$url/raw/$branch/cmake/apple/fluid-synth.rb"
 
 sudo npm install -g appdmg
--- a/.travis/osx..script.sh
+++ b/.travis/osx..script.sh
@@ -1,8 +1,10 @@
 #!/usr/bin/env bash
 
-if [ $QT5 ]; then
+if [ "$QT5" ]; then
         # Workaround; No FindQt5.cmake module exists
-        export CMAKE_PREFIX_PATH="$(brew --prefix qt55)"
+        CMAKE_PREFIX_PATH="$(brew --prefix qt55)"
+        export CMAKE_PREFIX_PATH
 fi
 
+# shellcheck disable=SC2086
 cmake $CMAKE_FLAGS -DUSE_WERROR=OFF ..
--- a/cmake/msys/msys_helper.sh
+++ b/cmake/msys/msys_helper.sh
@@ -21,27 +21,27 @@ if [ $? -ne 0 ]; then
 fi
 
 info "Preparing the git directory..."
-mkdir $HOME/.git; touch $HOME/.git/config > /dev/null &2>1
+mkdir "$HOME/.git"; touch "$HOME/.git/config" > /dev/null &2>1
 git config --global http.sslverify false
 
 info "Cloning the repository..."
-if [ -d ./lmms ]; then 
+if [ -d ./lmms ]; then
 	warn "  - Skipping, ./lmms already exists"
-else 
+else
 	git clone -b $branch https://github.com/$fork/lmms.git
 fi
 
 
 info "Fetching ppa using cmake/msys/fetch_ppas.sh..."
-if [ -d $HOME/ppa ]; then
+if [ -d "$HOME/ppa" ]; then
         warn "  - Skipping, $HOME/ppa already exists"
 else
-	./lmms/cmake/msys/fetch_ppa.sh	
+	./lmms/cmake/msys/fetch_ppa.sh
 fi
 
 
 info "Extracting debs to $HOME/ppa/opt/, etc..."
-if [ -d $HOME/ppa/opt ]; then
+if [ -d "$HOME/ppa/opt" ]; then
         warn "  - Skipping, $HOME/ppa/opt already exists"
 else
         ./lmms/cmake/msys/extract_debs.sh
@@ -49,15 +49,15 @@ fi
 
 info "Preparing library merge, making all qt headers writable..."
 chmod u+w /mingw64/include/qt4 -R
-chmod u+w /mingw32/include/qt4 -R 
+chmod u+w /mingw32/include/qt4 -R
 
 info "Merging mingw headers and libraries from ppa over existing system libraries..."
 find /mingw64 |grep sndfile.h
 
-if [ $? -ne 0 ]; then 
-	\cp -r $HOME/ppa/opt/mingw* /
-else 
-	warn "  - Skipping, sndfile.h has already been merged" 
+if [ $? -ne 0 ]; then
+	command cp -r "$HOME/ppa/opt/mingw"* /
+else
+	warn "  - Skipping, sndfile.h has already been merged"
 fi
 
 fltkver="1.3.3"
@@ -69,18 +69,18 @@ stkver="4.5.1"
 
 info "Downloading and building fltk $fltkver"
 
-mingw_root="/$(echo $MSYSTEM|tr '[:upper:]' '[:lower:]')"
+mingw_root="/$(echo "$MSYSTEM"|tr '[:upper:]' '[:lower:]')"
 which fluid
-if [ $? -ne 0 ]; then 
-	wget http://fltk.org/pub/fltk/$fltkver/fltk-$fltkver-source.tar.gz -O $HOME/fltk-source.tar.gz
+if [ $? -ne 0 ]; then
+	wget http://fltk.org/pub/fltk/$fltkver/fltk-$fltkver-source.tar.gz -O "$HOME/fltk-source.tar.gz"
 	if [ $? -ne 0 ]; then
-		err "ERROR: Could not download fltk.  Exiting."	
+		err "ERROR: Could not download fltk.  Exiting."
 	fi
-	tar zxf $HOME/fltk-source.tar.gz -C $HOME/
-	pushd $HOME/fltk-$fltkver
+	tar zxf "$HOME/fltk-source.tar.gz" -C "$HOME/"
+	pushd "$HOME/fltk-$fltkver"
 
 	info "  - Compiling fltk $fltkver..."
-	./configure --prefix=$mingw_root --enable-shared
+	./configure --prefix="$mingw_root" --enable-shared
 
 	make
 
@@ -90,25 +90,25 @@ if [ $? -ne 0 ]; then
 	if [ $? -ne 0 ]; then
         	err "ERROR: Could not build/install fltk -- Zyn needs this.  Exiting."
 	fi
-	
-#	ln -s $mingw_root/usr/local/bin/fluid.exe $mingw_root/bin/fluid.exe 	
+
+#	ln -s $mingw_root/usr/local/bin/fluid.exe $mingw_root/bin/fluid.exe
 	popd
 else
-	warn "  - Skipping, fluid binary already exists" 
+	warn "  - Skipping, fluid binary already exists"
 fi
 
 info "Downloading and building libogg $oggver"
 
-if [ ! -e $mingw_root/lib/libogg.dll.a ]; then 
-	wget http://downloads.xiph.org/releases/ogg/libogg-$oggver.tar.xz -O $HOME/libogg-source.tar.xz
+if [ ! -e "$mingw_root/lib/libogg.dll.a" ]; then
+	wget http://downloads.xiph.org/releases/ogg/libogg-$oggver.tar.xz -O "$HOME/libogg-source.tar.xz"
 	if [ $? -ne 0 ]; then
-		err "ERROR: Could not download libogg.  Exiting."	
+		err "ERROR: Could not download libogg.  Exiting."
 	fi
-	tar xf $HOME/libogg-source.tar.xz -C $HOME/
-	pushd $HOME/libogg-$oggver
+	tar xf "$HOME/libogg-source.tar.xz" -C "$HOME/"
+	pushd "$HOME/libogg-$oggver"
 
 	info "  - Compiling libogg $oggver..."
-	./configure --prefix=$mingw_root
+	./configure --prefix="$mingw_root"
 
 	make
 
@@ -116,7 +116,7 @@ if [ ! -e $mingw_root/lib/libogg.dll.a ]; then
 	make install
 
 	# for some reason libgig needs this
-	./configure --prefix=/opt$mingw_root
+	./configure --prefix="/opt$mingw_root"
 
 	make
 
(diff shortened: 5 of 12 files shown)


cmake/msys/msys_helper.sh
```diff
@@ -21,43 +21,43 @@ if [ $? -ne 0 ]; then
...
 
 info "Merging mingw headers and libraries from ppa over existing system libraries..."
 find /mingw64 |grep sndfile.h
 
-if [ $? -ne 0 ]; then 
-	\cp -r $HOME/ppa/opt/mingw* /
-else 
-	warn "  - Skipping, sndfile.h has already been merged" 
+if [ $? -ne 0 ]; then
+	command cp -r "$HOME"/ppa/opt/mingw* /
```
tresf: @zapashcanon can you please provide a reference how `\cp` can be replaced with `command cp`.

plugins/zynaddsubfx/zynaddsubfx/style.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
```
tresf: I don't believe we use this file at all and it's not our code.  This change is unnecessary.

cmake/msys/msys_helper.sh
zapashcanon: Not sure about this, but, to fix the sc error I had a look at: https://github.com/koalaman/shellcheck/wiki/SC1001

I hoped the CI would tell me if the change was correct.

.travis/linux.win64.install.sh
```diff
@@ -3,10 +3,12 @@ set -e
 
 # First, install 32-bit deps
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-. $DIR/linux.win32.install.sh
+# Maybe find a better fix: see https://github.com/koalaman/shellcheck/wiki/SC1090
+# shellcheck disable=SC1090
```
tresf: Can you please explain this in you PR rather than in code?  Disabling the shellcheck is perfectly adequate.  People can pull up `SC1090` on their own.

.travis/osx..install.sh
```diff
@@ -2,18 +2,18 @@
...
 
-brew install $PACKAGES ccache
+brew install "$PACKAGES" ccache
 
 # Recompile fluid-synth without CoreAudio per issues #649
 # Changes to fluid-synth.rb must be pushed to URL prior to use
 url=$(git remote get-url origin)
 branch=$(git symbolic-ref --short HEAD)
-brew install --build-from-source $url/raw/$branch/cmake/apple/fluid-synth.rb
+brew install --build-from-source "$url"/raw/"$branch"/cmake/apple/fluid-synth.rb
```
tresf: What is wrong with `"$url/raw/$branch/cmake/apple/fluid-synth.rb" ?  Is there something this tool has against "$foo/$bar"?  It's much more readable than "$foo"/"$bar".  This comment can be copied and pasted throughout, the PR.

plugins/zynaddsubfx/zynaddsubfx/style.sh
zapashcanon: Fixed.

.travis/osx..install.sh
zapashcanon: The tool has nothing against `"$url/raw/$branch/cmake/apple/fluid-synth.rb"` ! I guess it would work too.

It has a problem with: `$url/raw/$branch/cmake/apple/fluid-synth.rb` saying: `-- SC2086: Double quote to prevent globbing and word splitting.`

See [SC2086](https://github.com/koalaman/shellcheck/wiki/SC2086).

I treated error/warning like this using search/replace, so, indeed, when there are multiple occurrence on the same line I could have use only one pair of `"`. With my coloration it's not hard to read at all, but yes, here on GitHub it looks bad.

Wiil fix this.

.travis/linux.win32.install.sh
```diff
@@ -16,11 +16,13 @@ MINGW_PACKAGES="mingw32-x-sdl mingw32-x-libvorbis mingw32-x-fluidsynth mingw32-x
...
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-. $DIR/linux.win.download.sh win32
+# Maybe find a better fix: see https://github.com/koalaman/shellcheck/wiki/SC1090
+# shellcheck disable=SC1090
+. "$DIR"/linux.win.download.sh win32
 
 PACKAGES="nsis cloog-isl libmpc3 qt4-linguist-tools mingw32 $MINGW_PACKAGES"
 
-sudo apt-get install -y $PACKAGES
+sudo apt-get install -y "$PACKAGES"
```
tresf: You've probably found this in Travis-CI already, but in this case the packages should explicitly have spaces between them.  You'll want to add your `# shellcheck disable=FOO` in front of this guy too.

liushuyu: On macOS environment as well

.travis/osx..script.sh
```diff
@@ -1,8 +1,9 @@
 #!/usr/bin/env bash
 
-if [ $QT5 ]; then
+if [ "$QT5" ]; then
         # Workaround; No FindQt5.cmake module exists
-        export CMAKE_PREFIX_PATH="$(brew --prefix qt55)"
+        CMAKE_PREFIX_PATH="$(brew --prefix qt55)"
```
tresf: Can you explain why these need to be split?

.travis/osx..script.sh
```diff
@@ -1,8 +1,9 @@
...
+        export CMAKE_PREFIX_PATH
 fi
 
-cmake $CMAKE_FLAGS -DUSE_WERROR=OFF ..
+cmake "$CMAKE_FLAGS" -DUSE_WERROR=OFF ..
```
tresf: Is this correct?  Won't be run into same problems as `apt-get` ? I'd expect `cmake` to actually want the spaces although it may be intelligent enough to take it this way, I don't think it's right.

.travis/osx..script.sh
zapashcanon: https://github.com/koalaman/shellcheck/wiki/SC2155

.travis/osx..script.sh
zapashcanon: Yep, I'm waiting Travis to tell. :)

cmake/msys/msys_helper.sh
tresf: @zapashcanon this script is for Windows-only, so it won't be tested by Travis-CI.  I'm reading up on `\cp` in general and the above logic seems to be the proper replacement.  Apparently as a safeguard, the `cp` command is aliased to something like `cp -i` which is `interactive` mode.  Both `\cp` as well as `command cp` bypasses interactive mode, thus justifying the change.   In interactive terminal, `\cp` is valid, but in a script it appears to be both unnecessary as well as invalid.

zapashcanon: > it won't be tested by Travis-CI

Indeed, I thought there was something like AppVeyor... :/

Is there someone using windows here ? Change should be ok anyway, but, still a good idea to test IMHO.

.travis/linux.win64.install.sh
zapashcanon: I'll do some ugly debugging using Travis here. I'll remove the two lines anyway once it's done. Just a reminder to myself. :)

.travis/osx..script.sh
tresf: That seems to be `set -e` motivated.  More here:  http://stackoverflow.com/a/22636016/3196753

Since macOS [won't update to bash 4.0](https://blog.bennycornelissen.nl/bash-4-x-on-osx-the-bash-that-apple-wont-ship/), this is necessary, thanks.

.travis/osx..script.sh
tresf: Please don't rely on Travis-CI, but test on your own.  If there's one or none parameters, you will get a false positive.

zapashcanon: I mean, I hope Travis will fail so I know it's wrong. If it pass, then.. further investigation needed.

.travis/linux.win32.install.sh
zapashcanon: Fixed everywhere.

.travis/osx..script.sh
liushuyu: CMake can correctly interpret this quoted thing

.travis/osx..install.sh
zapashcanon: Actually, there's only two places where there is many var in the same path. Here and [there](https://github.com/LMMS/lmms/pull/3455/files#diff-e2415930ea7e22e5bcb8d5016798757dR48).

.travis/osx..script.sh
tresf: That's a good point.  It's presumptuous.  Please remove.

.travis/osx..install.sh
tresf: I stopped counting at 10 instances.

.travis/osx..script.sh
Lukas-W: > CMake can correctly interpret this quoted thing, I have tested it

I have tested it and it does not, the same goes with `apt-get`. Putting quotes around will pass `$CMAKE_FLAGS` as a single argument. For example, `-DA=2 -DB=3` will be interpreted as if you passed `-DA="2 -DB=3"`.

.travis/osx..install.sh
zapashcanon: Could you show me one more ? Seems like we're not talking about the same thing... :/

.travis/osx..script.sh
zapashcanon: Done.

.travis/osx..install.sh
tresf: "$DIR", "$PPA_ROOT", "$HOME" are all used in areas they fall exactly into this category.

zapashcanon: Ok, I first thought you were talking only of the cases where there are many var in the same path.

zapashcanon: Fixed.

.travis/linux.win64.install.sh
zapashcanon: Ok, so, after looking into this, seems to me it would be useless to do something else. Just disabling the sc is enough IMHO.

zapashcanon:
I think it's ready for merge.

.travis/linux.win32.install.sh
```diff
@@ -16,10 +16,13 @@ MINGW_PACKAGES="mingw32-x-sdl mingw32-x-libvorbis mingw32-x-fluidsynth mingw32-x
 export MINGW_PACKAGES
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-. $DIR/linux.win.download.sh win32
+# Maybe find a better fix: see https://github.com/koalaman/shellcheck/wiki/SC1090
```
tresf: Please remove this comment. :)

cmake/msys/msys_helper.sh
```diff
@@ -90,33 +90,33 @@ if [ $? -ne 0 ]; then
...
-	warn "  - Skipping, fluid binary already exists" 
+	warn "  - Skipping, fluid binary already exists"
 fi
 
 info "Downloading and building libogg $oggver"
 
-if [ ! -e $mingw_root/lib/libogg.dll.a ]; then 
-	wget http://downloads.xiph.org/releases/ogg/libogg-$oggver.tar.xz -O $HOME/libogg-source.tar.xz
+if [ ! -e "$mingw_root/lib/libogg.dll.a" ]; then
+	wget http://downloads.xiph.org/releases/ogg/libogg-$oggver.tar.xz -O "$HOME/libogg-source.tar.xz"
```
tresf: Why doesn't `shellcheck` bark at this one?  Shouldn't the URL be escaped?

cmake/msys/msys_helper.sh
```diff
@@ -126,33 +126,33 @@ if [ ! -e $mingw_root/lib/libogg.dll.a ]; then
...
+	warn "  - Skipping, libogg binary already exists"
 fi
 
 
 info "Downloading and building libvorbis $vorbisver"
 
-if [ ! -e $mingw_root/lib/libvorbis.dll.a ]; then 
-	wget http://downloads.xiph.org/releases/vorbis/libvorbis-$vorbisver.tar.xz -O $HOME/libvorbis-source.tar.xz
+if [ ! -e "$mingw_root/lib/libvorbis.dll.a" ]; then
+	wget http://downloads.xiph.org/releases/vorbis/libvorbis-$vorbisver.tar.xz -O "$HOME/libvorbis-source.tar.xz"
```
tresf: Same, I'd expect this to be quoted like every other concatenated string.

cmake/msys/msys_helper.sh
```diff
@@ -162,32 +162,32 @@ if [ ! -e $mingw_root/lib/libvorbis.dll.a ]; then
...
-	warn "  - Skipping, libvorbis binary already exists" 
+	warn "  - Skipping, libvorbis binary already exists"
 fi
 
 info "Downloading and building flac $flacver"
 
-if [ ! -e $mingw_root/lib/libFLAC.dll.a ]; then 
-	wget http://downloads.xiph.org/releases/flac/flac-$flacver.tar.xz -O $HOME/flac-source.tar.xz
+if [ ! -e "$mingw_root/lib/libFLAC.dll.a" ]; then
+	wget http://downloads.xiph.org/releases/flac/flac-$flacver.tar.xz -O "$HOME/flac-source.tar.xz"
```
tresf: Same, I'd expect this to be quoted like every other concatenated string.

cmake/msys/msys_helper.sh
```diff
@@ -197,54 +197,54 @@ if [ ! -e $mingw_root/lib/libFLAC.dll.a ]; then
...
-	warn "  - Skipping, libvorbis flac already exists" 
+	warn "  - Skipping, libvorbis flac already exists"
 fi
 
 info "Downloading and building libgig $gigver"
 
-if [ ! -e $mingw_root/lib/libgig/libgig.dll.a ]; then 
-	wget http://download.linuxsampler.org/packages/libgig-$gigver.tar.bz2 -O $HOME/gig-source.tar.xz
+if [ ! -e "$mingw_root/lib/libgig/libgig.dll.a" ]; then
+	wget http://download.linuxsampler.org/packages/libgig-$gigver.tar.bz2 -O "$HOME/gig-source.tar.xz"
```
tresf: Same, I'd expect this to be quoted like every other concatenated string.

cmake/msys/msys_helper.sh
```diff
@@ -197,54 +197,54 @@ if [ ! -e $mingw_root/lib/libFLAC.dll.a ]; then
...
-	warn "  - Skipping, libgig binary already exists" 
+	warn "  - Skipping, libgig binary already exists"
 fi
 
 info "Downloading and building stk $stkver"
 
-if [ ! -e $mingw_root/lib/libstk.dll ]; then 
-	wget http://ccrma.stanford.edu/software/stk/release/stk-$stkver.tar.gz -O $HOME/stk-source.tar.xz
+if [ ! -e "$mingw_root/lib/libstk.dll" ]; then
+	wget http://ccrma.stanford.edu/software/stk/release/stk-$stkver.tar.gz -O "$HOME/stk-source.tar.xz"
```
tresf: Same, I'd expect this to be quoted like every other concatenated string.

cmake/msys/msys_helper.sh
zapashcanon: Because a few lines before: `oggver="1.3.2"`, so it's okay. :)

cmake/msys/msys_helper.sh
zapashcanon: Same.

cmake/msys/msys_helper.sh
zapashcanon: Same.

cmake/msys/msys_helper.sh
zapashcanon: Same.

cmake/msys/msys_helper.sh
zapashcanon: Same.

.travis/linux.win32.install.sh
zapashcanon: Missed that one, thanks !

cmake/msys/msys_helper.sh
tresf: >  Because a few lines before: oggver="1.3.2", so it's okay. :)

And it knows there will never be a space in the value?

 * If `"$QT5"` were really so dangerous in an `if` statement, how come Ubuntu 12.04 and 14.04 have never had a problem with it?

zapashcanon: That we won't set something like: `oggver="1 3 2"` in the future ? If this is what you're asking, the answer is: of course no.

If your question is that it knows that `oggver` is not an env var and is not modified by the script between the declaration and the use of the var etc. ; the answer is yes.

I plan to add shellcheck with astyle and kwstyle, so if a change like this is made, we will be warned.

But I don't see why we should quote this as it's currently not needed, probably won't (space in a version number in an url is not a common thing), makes the code harder to read and also because we'll be able to prevent this by adding shellcheck in the ci chain.

tresf: > But I don't see why we should quote this as it's currently not needed

Because people will continue to copy and paste it (as they have previously), and we may calculate the values.

> But I don't see why we should quote this as it's currently not needed, probably won't (space in a version number in an url is not a common thing)

And `$HOME` is?  I've used a lot of systems and I can promise you `%20` in a URL is more common than `<space>` in a `/home` path.

> makes the code harder to read

@zapashcanon arguably... all this PR does is make the code harder to read, that portion is quite irrelevant.

Adding double quotes to the URL does a two things in my eyes.

1. Helps make the script consistent.  For those that don't have the accuracy of shellcheck, it's a good habit!
2. Helps copy/paste, especially if the version becomes parametrized or calculated down the road.  This is very likely for scripts like this that are growing rather rapidly, link against specific versions and have a lot of redundant code.

That said, it's fine to merge without.  I won't hold up the patch for this alone.

>  plan to add shellcheck with astyle and kwstyle

We should add it separately.  We only have a handful of scripts, and this is one more greenlight we can have for the PR process.

Thanks for your work on this!

zapashcanon: > Because people will continue to copy and paste it (as they have previously), and we may calculate the values.

Indeed...

> And $HOME is? I've used a lot of systems and I can promise you %20 in a URL is more common than <space> in a /home path.

Space in an url yes, but this is a version number, in a file name in an url... who's crazy enough to put a space here ? :p We're not setting `$HOME` here, so...

> @zapashcanon arguably... all this PR does is make the code harder to read, that portion is quite irrelevant.

I agree other parts are making the code harder to read, but, I think it's not a reason to avoid doing it when it's possible. :)

> We should add it separately. We only have a handful of scripts, and this is one more greenlight we can have for the PR process.

Actually, I though I'd add a separate build for things like this, with a first PR for shellcheck (just a matter of adding a shell script and a few lines to travis config file), then, doing a second PR to add astyle and kwstyle in this build as it will recquire more changes. But, I'll talk of this somewhere else.

> That said, it's fine to merge without. I won't hold up the patch for this alone.

Yep, I think (and would prefer if) the PR can be merged like this too.

tresf: 👍

jasp00:
IMHO, scripts should not have `# shellcheck disable` comments, shellcheck should be smart enough. You may want to move the multiple commands in `.travis.yml` to a script.

zapashcanon:
> IMHO, scripts should not have # shellcheck disable comments, shellcheck should be smart enough. 

I agree this is kinda ugly. But actually, I don't see how shellcheck could guess. The only way I'm thinking of, is that it kept a list of every command and options existing in the world with which user probably want to split on space. Don't seems a good idea to me...

Anyway, if you look at [Exceptions](https://github.com/koalaman/shellcheck/wiki/SC2086#exceptions), it is saying that we should have used a function or an array. Do you think this is the way to go ? If yes, I could do a PR to remove `# shellcheck disable` comments.

> You may want to move the multiple commands in `.travis.yml` to a script.

Seems like a good idea. Will do.

tresf:
The exceptions are safe and clean.  No more scripts please.  Arrays are fine.  Functions get disgusting.  We shouldn't and won't clobber solid logic with superfluous logic because of a helper.  The exceptions have been justified perfectly.

jasp00:
>I don't see how shellcheck could guess.

There should be two different checks. A non-space-delimited one (` $VAR/foo `) where quoting is very likely to be safe, and a space-delimited one (` $VAR `) where quoting is not clear. We would disable the latter one.

>No more scripts please.

Then, how do you shellcheck `.travis.yml` commands?

>Arrays are fine.

Are you sure? I do not remember CMake handling arrays and CMake may depend on these variables.

zapashcanon:
> There should be two different checks. A non-space-delimited one ($VAR/foo) where quoting is very likely to be safe, and a space-delimited one ($VAR) where quoting is not clear. We would disable the latter one.

Then, if we use something like `$PATH`, it won't be checked as we disabled it... And still, we would have to disable it by adding a comment before each line in this case, which would be exactly like what we currently have no ? Or we don't disable this check, but, we're in the same case than before then.
(Note that I'm not sure I understood your proposal.)

tresf:
> Arrays are fine.
> > Are you sure? 

Thanks for asking.   I've put a lot of thought into this and I find there's no clean way to fix the `sudo apt-get install FOO BAR` problem without making the script quite unreadable.

Even if the array technique works (shellcheck has some decent examples of this), it's an unnecessary safeguard because we aren't parameterizing the string.  Our usage is a safe usage (since package names can't contain spaces).  Shellcheck is throwing a false-positive.  These false-positives are exactly what the exception comments were created for.  They're scoped properly.  We're using them correctly.

A design change (parameterizing the string in an unsafe way) may warrant the use of an array.

In regards to functions and sourcing external files:
* Functions in bash require forward declaration.  This is harder to read and still suffers the problem @zapashcanon mentions "still, we would have to disable it by adding a comment before each line in this case"
* Sourcing external files seem unwarranted at this point.
  * The `apt-get` script call shouldn't be decoupled from the string in which it requires unless we have a good design reason.  One good design reason may be to offer CI build support for other platforms, and then we can source `ubuntu_14.04.txt` in Travis yet direct users to source `centos_7.0.txt`, etc for Fedora/RHEL, etc.  We should consider this long-term, but it's out of scope for this PR.
 * The `cmake` call would also lose a significant amount of readability if externally sourced.  After making a second review, I feeel the areas which expose potential problems have already been addressed.

jasp00:
>if we use something like `$PATH`

Could you show an example?

>we would have to disable it by adding a comment before each line in this case, which would be exactly like what we currently have no ?

No, the check would be disabled from the command line.

>I've put a lot of thought into this and I find there's no clean way to fix the `sudo apt-get install $FOO $BAR` problem without making the script quite unreadable.

shellcheck is not smart enough and should not handle these cases, unless it is aware of commands such as `apt-get`.

zapashcanon:
> Could you show an example?

Something like:

```sh
make DESTDIR=$INSTALL_DIR install
```

> No, the check would be disabled from the command line.

Ok, but then, as you said, it would be a category where splitting on space is not clear, so we would probably miss some cases, like the above example.

> shellcheck is not smart enough and should not handle these cases, unless it is aware of commands such as apt-get.

I still fail to see how it could know which cases it should not handle. And keeping a list of commands seems not feasible to me. Even if it did, it would have to know the whole man of each command, because we could have things like: `apt-get install -c=$PATH_TO_CONFIG_FILE $PACKAGES`.

jasp00:
```
make DESTDIR=$INSTALL_DIR install
```
This is not an example because it would hit the non-space-delimited rule and quote `$INSTALL_DIR`.

>I still fail to see how it could know which cases it should not handle.

shellcheck should not know which cases it should not handle. It should know which cases it can handle and refrain in unknown situations. Maintaining a list of well-known commands and their syntax is feasible; an example is the `\cp` case.

tresf:
Since this PR is in good shape, we should move any supplemental conversation to discord.  This conversation started when a post-merge comment was added, but I think we're all OK with the way it's written; the way it's merged. 👍

PR #3458, opened 2017-03-26
Always highlight the beats in PianoRoll and AutomationEditor using only time signature

z-up:
The wrong lines were highlighted in PianoRoll and Automation Editor when using triplets.

Duration of the beat is defined only by the denominator of the time signature and it does not depend on whether or not you use triplets. In the previous version of the code the beat duration changed, when you used triplets.

When using triplets you could also highlight the BEGINNING of each triplet. But in the previous version of the code, when you chose for example 1/12 in the quantization level, each 1/6 note was highlighted, that is the first and the third note of the first triplet, the second note of the second triplet, the first and third notes of the third triplet, and so on, which does not make any sense.

In this version of code, simply each beat (defined only by the time signature) is highlighted.

screenshots illustrating the issues.
image
image

Edit by @Lukas-W: Included images from external hoster

### DIFF
--- a/src/gui/editors/AutomationEditor.cpp
+++ b/src/gui/editors/AutomationEditor.cpp
@@ -1253,13 +1253,6 @@ void AutomationEditor::paintEvent(QPaintEvent * pe )
 		int ticksPerBeat = DefaultTicksPerTact /
 			Engine::getSong()->getTimeSigModel().getDenominator();
 
-		// triplet mode occurs if the note quantization isn't a multiple of 3
-		// note that the automation editor does not support triplets yet
-		if( AutomationPattern::quantization() % 3 != 0 )
-		{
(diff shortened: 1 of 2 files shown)


Umcaruje:
Cherry picked to `stable-1.2` via https://github.com/LMMS/lmms/commit/8416cd60debab1eb95eae164d441138cc36702a7

z-up:
>Thanks for the contribution

Thank you from a happy LMMS user=)

>the beat lines need some redesigning now on very low quantization levels

I personally think that it is "logical" to consistently show the beat lines, so you could always see where 2, 3 and 4 are, even if the line pattern starts to look a bit irregular when you choose 1/2 or 1/3 in the quantization level. But on the other hand when the grid lines are so sparse, it's harder to get lost where you are in the bar, so may be you could just hide the beat lines.

HummusSamurai:
I love those usability changes!

Issue #3394, opened 2017-03-02
Multitrack recording button and countdown clock


Gabrielxd195:
Hello!. My name is Gabriel Lopez, and I am Latin from Dominican Republic, so I do not speak English and everything I wrote is translated, so forgive my spelling mistakes. First of all I would like to tell you that I am not a developer and that I am starting to get into music, with lmms I am taking my first steps and that is why I want to give two of many ideas that I have to improve linux multimedia studio, lmms is versatile and powerful but le There are many things to improve and add, such as: The button of simultaneous recording of multiple tracks and the countdown clock at the time of recording. 
![boton multigrabacion actual](https://cloud.githubusercontent.com/assets/26127693/23489277/c2b4fd36-fec7-11e6-9364-cb498ee68834.png)
 That's the current interface, with the two useless recording buttons. (These images edit them with gimp, so forgive the mistakes)

![boton de grabacion multipista](https://cloud.githubusercontent.com/assets/26127693/23490167/20bfb222-fecd-11e6-9968-edc8c8ff2a61.png)
And this is the interface suggested by me. As in the case of ardour or pro tools, the buttons on the tracks marked with the record button on (red button) are recording simultaneously, while the other disabled instruments (gray button) do not record. The recording buttons of the song editor when pressed, give the command to record on the marked tracks (red button).

![cuenta regresiva actual](https://cloud.githubusercontent.com/assets/26127693/23490327/61b6adb6-fece-11e6-953f-a42a5550498f.png)
As far as I've seen, lmms does not have a multitrack record button, but it also does not have a countdown (the 4 seconds before recording) at the moment of pressing the record button on the tracks, causing it to lose rhythm and time. So I suggest a clock that marks the allocated seconds before recording.
![reloj de conteo regresivo](https://cloud.githubusercontent.com/assets/26127693/23490367/b04031be-fece-11e6-9b32-d22e753c4d9e.png)

musikBear:
You can already do both in lmms

In respect to: 
* selective export: 
mute the tracks you do not want in your export
LMMS will also (shortly) allow you to export one track at a time, to separate sound-files 

* countdown
insert one-bar of sharp ticks before the part you plan to record, then you can listen for the ticks and be prepared to play on your keyboard, when the fourth tick rings out.
A countdown clock has also been proposed before, btw

Gabrielxd195:
[...]

PR #3467, opened 2017-03-30
[WIP] Add record count down

liushuyu:
**Do not merge this until it is told to do so**

Tagging my "mentor" @jasp00 

In respond to #3394

### DIFF
--- a/include/MainWindow.h
+++ b/include/MainWindow.h
@@ -32,6 +32,7 @@
 
 #include "ConfigManager.h"
 #include "SubWindow.h"
+#include "LcdSpinBox.h"
 
 class QAction;
 class QDomElement;
@@ -168,6 +169,7 @@ public slots:
 	void toggleFxMixerWin();
 	void togglePianoRollWin();
 	void toggleControllerRack();
+    void updateCountDownDuration();
 
 	void updatePlayPauseIcons();
 
@@ -224,6 +226,7 @@ private:
 
 	QBasicTimer m_updateTimer;
 	QTimer m_autoSaveTimer;
+    QTimer m_recordCountDownTimer;
 	int m_autoSaveInterval;
 
 	friend class GuiApplication;
@@ -231,6 +234,10 @@ private:
 	QMenu * m_viewMenu;
 
 	ToolButton * m_metronomeToggle;
+
+    LcdSpinBox * m_countDownClock;
+
+    LcdSpinBoxModel * m_countDownValue;
 
 	SessionState m_session;
 
@@ -243,6 +250,8 @@ private slots:
 	void updateViewMenu( void );
 	void updateConfig( QAction * _who );
 	void onToggleMetronome();
+    void startRecordCountDown();
+    void decrementRecordCountDown();
 
 
 signals:
--- a/include/Song.h
+++ b/include/Song.h
@@ -254,6 +254,10 @@ public:
 		return m_timeSigModel;
 	}
 
+    void setRecordDelay( int time );
+    int getRecordDelay();
+    void startRecordCountDown();
+
 
 public slots:
 	void playSong();
@@ -358,6 +362,7 @@ private:
 	PlayModes m_playMode;
 	PlayPos m_playPos[Mode_Count];
 	tact_t m_length;
+    int m_recordDelay;
 
 	const Pattern* m_patternToPlay;
 	bool m_loopPattern;
@@ -384,6 +389,7 @@ signals:
 	void controllerAdded( Controller * );
 	void controllerRemoved( Controller * );
 	void updateSampleTracks();
+    void recordCoundDown();
 
 } ;
 
--- a/src/gui/MainWindow.cpp
+++ b/src/gui/MainWindow.cpp
@@ -71,6 +71,7 @@ MainWindow::MainWindow() :
 	m_recentlyOpenedProjectsMenu( NULL ),
 	m_toolsMenu( NULL ),
 	m_autoSaveTimer( this ),
+    m_recordCountDownTimer( this ),
 	m_viewMenu( NULL ),
 	m_metronomeToggle( 0 ),
 	m_session( Normal )
@@ -209,9 +210,11 @@ MainWindow::MainWindow() :
 
 	connect( Engine::getSong(), SIGNAL( playbackStateChanged() ),
 				this, SLOT( updatePlayPauseIcons() ) );
+    connect( &m_recordCountDownTimer, SIGNAL( timeout()), this, SLOT(decrementRecordCountDown()) );
+    connect( Engine::getSong(), SIGNAL( recordCoundDown()), this, SLOT(startRecordCountDown()) );
+    m_countDownValue = new LcdSpinBoxModel();
+    m_countDownValue->setRange( 0, 99 );
 }
-
-
 
 
 MainWindow::~MainWindow()
@@ -552,6 +555,12 @@ void MainWindow::finalize()
 								m_toolBar );
 	controllers_window->setShortcut( Qt::Key_F11 );
 
+    m_countDownClock = new LcdSpinBox(2, this, tr("Count Down Clock"));
+
+    m_countDownClock->setLabel( tr("SEC") );
+    m_countDownClock->setModel( m_countDownValue );
+    connect( m_countDownValue, SIGNAL(dataChanged()), this, SLOT( updateCountDownDuration() ) );
+
 	m_toolBarLayout->addWidget( song_editor_window, 1, 1 );
 	m_toolBarLayout->addWidget( bb_editor_window, 1, 2 );
 	m_toolBarLayout->addWidget( piano_roll_window, 1, 3 );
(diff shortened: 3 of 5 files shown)


include/MainWindow.h
```diff
@@ -168,6 +169,7 @@ public slots:
 	void toggleFxMixerWin();
 	void togglePianoRollWin();
 	void toggleControllerRack();
+    void updateCountDownDuration();
```
Umcaruje: Spaces instead of tabs for indentation. You want to change this right away so you don't have more work down the way :)

liushuyu: Yeah, it's annoying my editor will convert all the tabs to spaces, I will try to configure it not to do so.

src/gui/MainWindow.cpp
```diff
@@ -552,13 +555,20 @@ void MainWindow::finalize()
 								m_toolBar );
 	controllers_window->setShortcut( Qt::Key_F11 );
 
+    m_countDownClock = new LcdSpinBox(2, this, tr("Count Down Clock"));
+
+    m_countDownClock->setLabel( tr("SEC") );
```
Umcaruje: The countdown should really be in bars/beats, not seconds. When you record you always want to be relative to the beat, because the countdown actually gives musicians time to align themselves with the beat. An arbitrary second value can't do that.

liushuyu: Well, in this case, we may want to implement a conversion algorithm and makes it public in `Engine` or  somewhere, this problem was also a headache in my previous PR (Go to feature), it will introduce offsets when doing the conversions.

include/MainWindow.h
liushuyu: Fixed, rest of the unfixed parts are staged locally, will be pushed soon.

src/gui/MainWindow.cpp
liushuyu: And maybe here we can toggle on the metronome automatically to assist the recording preparation even better?

Umcaruje: Yes, absolutely. The metronome ought to be more intuitive, that's a step in the right direction.

liushuyu: I came up with such conversion method:
image

The equation (1) is our original algorithm found in TimeDisplayWidget.cpp:112
The equation (2) is to solve the equation (1) given `Tick` value to express `Msec` as a function of `Tick tempo` and `TicksPerTack`
The equation (4) is to ease manual verification of whether the equation (2) works correctly.
As far as I could tell, given the precise value of `ticks` (preserving at least 1 decimal point) can give the very close output (around +- 3 msec at 999 BPM)

src/gui/MainWindow.cpp
```diff
@@ -1298,15 +1308,34 @@ void MainWindow::onToggleMetronome()
...
 
+void MainWindow::decrementRecordCountDown()
+{
+	int currentReadOut = m_countDownValue->value();
+	if ( currentReadOut - 1 < 0) {
+		m_recordCountDownTimer.stop();
+		return;
+	}
+	m_countDownValue->setValue( currentReadOut - 1 );
+}
```
zapashcanon: Maybe removing `return;` and adding a `else` would make this easier to read ?

src/gui/MainWindow.cpp
```diff
@@ -1298,15 +1308,34 @@ void MainWindow::onToggleMetronome()
...
 
 
 void MainWindow::toggleControllerRack()
 {
-	toggleWindow( gui->getControllerRackView() );
+    toggleWindow( gui->getControllerRackView() );
```
Umcaruje: missed one spot with spaces here

src/gui/MainWindow.cpp
```diff
@@ -552,13 +555,20 @@ void MainWindow::finalize()
...
+	connect( m_countDownValue, SIGNAL(dataChanged()), this, SLOT( updateCountDownDuration() ) );
+
 	m_toolBarLayout->addWidget( song_editor_window, 1, 1 );
 	m_toolBarLayout->addWidget( bb_editor_window, 1, 2 );
 	m_toolBarLayout->addWidget( piano_roll_window, 1, 3 );
 	m_toolBarLayout->addWidget( automation_editor_window, 1, 4 );
 	m_toolBarLayout->addWidget( fx_mixer_window, 1, 5 );
 	m_toolBarLayout->addWidget( project_notes_window, 1, 6 );
 	m_toolBarLayout->addWidget( controllers_window, 1, 7 );
+    m_toolBarLayout->addWidget( m_countDownClock, 1, 8 );
```
Umcaruje: spaces

src/gui/MainWindow.cpp
Umcaruje: In the final version don't use 1250 like a magic number, cause the TicksPerTact can change in the future. It's better to have a constant that's `24000/TicksPerTact`

src/gui/MainWindow.cpp
jasp00: @liushuyu's choice is good. I am used to the `return` pattern; it allows to decrease indentation, which helps reading.

include/MainWindow.h
```diff
@@ -224,6 +226,7 @@ public slots:
 
 	QBasicTimer m_updateTimer;
 	QTimer m_autoSaveTimer;
+	QTimer m_recordCountDownTimer;
```
jasp00: These timer facilities are not needed in the GUI, the model should handle the countdown.

include/MainWindow.h
```diff
@@ -243,6 +250,8 @@ private slots:
 	void updateViewMenu( void );
 	void updateConfig( QAction * _who );
 	void onToggleMetronome();
+	void startRecordCountDown();
+	void decrementRecordCountDown();
```
jasp00: These are not needed.

include/Song.h
```diff
@@ -254,6 +254,10 @@ class EXPORT Song : public TrackContainer
 		return m_timeSigModel;
 	}
 
+	void setRecordDelay( int time );
+	int getRecordDelay();
+	void startRecordCountDown();
```
jasp00: `startRecordCountDown` would be private because the countdown should be integrated with `record`.

src/gui/editors/PianoRoll.cpp
```diff
@@ -4357,15 +4357,17 @@ void PianoRollWindow::stop()
 
 void PianoRollWindow::record()
 {
-	m_editor->record();
+	Engine::getSong()->startRecordCountDown();
+	QTimer::singleShot( Engine::getSong()->getRecordDelay() * 1000 , m_editor, SLOT( record() ) );
```
jasp00: Since the countdown will be based on bars and the metronome is going to play, you should rely on the rendering process and tick count instead of timers.

liushuyu:
@jasp00 Thanks for your feedback! I will fix the code accordingly.

src/gui/MainWindow.cpp
```diff
@@ -552,13 +555,20 @@ void MainWindow::finalize()
 								m_toolBar );
 	controllers_window->setShortcut( Qt::Key_F11 );
 
+	m_countDownClock = new LcdSpinBox(2, this, tr("Count Down Clock"));
+
+	m_countDownClock->setLabel( tr("SEC") );
+	m_countDownClock->setModel( m_countDownValue );
+	connect( m_countDownValue, SIGNAL(dataChanged()), this, SLOT( updateCountDownDuration() ) );
```
liushuyu: Ahh, there's a small problem, I don't know why the `manualChange` wasn't fired during my testing.

src/gui/editors/PianoRoll.cpp
liushuyu: This is because I haven't familiar with the rendering process of LMMS, I will try to understand it before doing this.

src/gui/MainWindow.cpp
jasp00: `manualChange` works. What object were you connecting?

liushuyu:
Well, I am writing to let you guys know I am still alive and I just have been fully occupied, hopefully I will have some time this weekend.
And I don't know how to make metronome tick while hold the record process, but I'll try to figure this out myself first before asking for help.

src/gui/MainWindow.cpp
liushuyu: >`manualChange` works. What object were you connecting?

I was connecting`m_countDownClock`, and I found the slot is never got called.

liushuyu:
Okay, I feel like to ditch all the work here, because I think I need a clear mind to implement a better approach.
So I think the implementation should go as follows:
1. ~~Add count down clock~~ This can be preserved, but the model should be in `Song` class
2. In `Mixer` class, add an exception to metronome trigger condition to make it tick when counting down, and together a count down process here.  (IIRC, should be in `Mixer::renderNextBuffer()` ?)
3. To prevent note input during count down, we need to modify each`record()` implementation to block input when in counting down status?

mirk0dex:
Any updates?

src/gui/MainWindow.cpp
Rossmaxx: No this is fine. We follow early returns too.

Rossmaxx:
> Any updates?

Looks like this PR is dead

Issue #3071, opened 2016-10-11
Integrate a code formatter into the build


jasp00:
A code formatter, mainly for C++, should be integrated into the build. The formatter should follow the [coding conventions](https://github.com/LMMS/lmms/wiki/Coding-conventions), and the configuration file should become the reference. It should be able to format files (for developers modifying code) and check whether files follow the format (for the build process). Only a subset should be formatted, excluding third-party files.

Existing solutions (`astyle`, `bcpp`, `indent`, `trueprint`) should be explored before developing a custom tool.

zapashcanon:
I was thinking about integrating astyle a few days ago as a first PR. I no one object it, I'll do it. :)

jasp00:
I am not sure astyle should be used. It does not fix the issue, although it would deal with whitespace formatting, which is the most frequent case. Handling whitespace is the easy part, the tool should be able to detect header guards, order of includes, class names, variables, members, parameters and returns; it should allow extensions or hooks if it does not cover these cases. Are you sure astyle is the best option?

tresf:
Putting the dev team on a single, solid IDE can solve all of these problems as well.

midi-pascal:
Qt Creator? Lmms is a Qt app after all! :smile_cat:

teeberg:
Maybe an [`.editorconfig`](http://editorconfig.org/) file could do the job too. That wouldn't force us to use a single IDE but would let us use any IDE that supports reading these files.

jasp00:
>a single, solid IDE can solve all of these problems as well.

If there is an IDE that can do such formatting, we can extract the formatter.

tresf:
> If there is an IDE that can do such formatting, we can extract the formatter.

I'm talking big league stuff like `CLion` or a commercial IDE that actually helps do code simplification, formatting suggestions, standard helping.  I'm not sure how Qt Creator compares from a standards perspective but I'd be happy to reach out and request a FOSS discount from a commercial IDE provider.

The reason I put emphasis on this is if we set an IDE standard, it will encourage video tutorials for getting developers involved, lower the barrier to entry and catch the problem as the code is happening, rather than at the time of the pull request.

Also, after a few PRs we'll find ourselves proactively fixing formatting (the IDE will yell at us if we don't), which is long overdue.  I've never used CLion but I provide it as an example because I use a sibling product, IntelliJ daily and it has git integration as well as before/after diffs  that are very helpful to see the red/green prior to committing, PRing.

[...]

Issue #3426, opened 2017-03-14
Code Styling - Tabs Versus Spaces


tresf:
I assume the decision to use tabs over spaces is historical due to the use of editors like `vi`.

Since we're supporting many platforms and multiple versions of those platforms, configuring QtCreator for each machine can get tedious.

First, I'll make the assumption that QtCreator is the preferred IDE for most contributors.  This has never been made official by the dev team (I personally use whatever is on the machine -- gedit, kate, notepad++, vi, TextWrangler) but I've spoken with many C++/Qt developers over the last year that use it exclusively, and now I'm starting to use it so it's the basis of discussion.
 
 * Currently, configuring QtCreator is a [4-step process](https://github.com/LMMS/lmms/pull/2033#issuecomment-98895801).
    * Is this something that can be solved by a preference file in the source repo?
 * The argument of tabs vs. spaces is a [very old and often heated one](http://softwareengineering.stackexchange.com/a/66) but switching to spacing (e.g. `4` spaces instead of the default tab width of `8`) can prevent the code from "walking too far to the right".

 * A major **downside** to switching from tabs to spaces is it's not default for the majority of text editors.

Is this a standard worth changing moving forward?  What is the primary IDE of choice among developers and how do you handle this problem in other large projects?

This isn't meant to start any flame wars.  I'm well aware this is a sensitive topic.

Spekular:
"switching to spacing (e.g. 4 spaces instead of the default tab width of 8) can prevent the code from "walking to far to the right"."

Can't this be fixed by adjusting tab width in your editor of choice? I don't see any argument for spaces unless they've been used in the past and we're keeping consistency.

tresf:
> Can't this be fixed by adjusting tab width in your editor of choice?

That depends on

```diff
-    what you consider the "editor" 
+	what you consider the"editor" 
```

... since GitHub is what PR reviews are performed on.

>  I don't see any argument for spaces

The S.O. article [here](http://softwareengineering.stackexchange.com/a/66) mentions a few good arguments for both sides.

PaulBatchelor:
I sure do love me my spaces, and I have to switch my preferences in vim just to do stuff in LMMS. However, I think efforts can be better spent elsewhere.

Spekular:
@tresf seems github tab size can be changed.

http://stackoverflow.com/questions/8833953/how-to-change-tab-size-on-github/23522945#23522945

Umcaruje:
[...]

PR #3470, opened 2017-03-31
[WIP] Add KWStyle and AStyle to the build

zapashcanon:
Hi,

First, I'd like to say that this is far from being finished.

So, as it's been said in #3071, I started following the current conventions.

I'd like to first talk about max. line length and space vs. tabs.

Current convention is to try to limit line length to 80 char. I didn't checked all files, but I found lines with 211 char. I think we should just stop breaking lines. I don't know which software are used by LMMS developer, but personally, I'm using atom or nvim and they both display too long lines in a convenient way, e.g. :

capture du 2017-03-31 19-33-14

I think, nowadays this is not useful anymore as every text editor should be handling this for you.

Would like to hear what do you think about this.

Currently, I set the limit to 250. I think we could set it to something even bigger, just to prevent some abuse, like 1000 char. Or we could just disable this check.

Then, tabs. I don't see a single reason why to use tabs over spaces nowadays. I don't know a single editor who can't replace tabs by space for you when you press tab. Tabs leads to strange display depending on your editor, so we won't see the same thing depending on our editor etc. So, I'd like that we switch to a 4 space indentation.

I'd like to fix those two points first, because they're leading to troubles with other checks.

To test the current KWStyle config, from the `lmms` folder:

```sh
KWStyle -v -xml KWStyle.xml -d src/core
```

### DIFF
--- a/.travis/install.sh
+++ b/.travis/install.sh
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+
+if [ "$TYPE" = 'style' ]; then
+  sudo apt install astyle kwstyle shellcheck
+else
+  # shellcheck disable=SC1090
+  . "$TRAVIS_BUILD_DIR/.travis/$TRAVIS_OS_NAME.$TARGET_OS.install.sh"
+fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ cache:
         - $HOME/.ccache
 matrix:
     include:
+        - env: TYPE=style
         - env: TARGET_OS=win32
         - env: TARGET_OS=win64
         - os: osx
@@ -18,17 +19,13 @@ matrix:
         - os: osx
           env: QT5=True
 before_install:
-   - . ${TRAVIS_BUILD_DIR}/.travis/${TRAVIS_OS_NAME}.${TARGET_OS}.before_install.sh
+    - . ${TRAVIS_BUILD_DIR}/.travis/before_install.sh
 install:
-    - . ${TRAVIS_BUILD_DIR}/.travis/${TRAVIS_OS_NAME}.${TARGET_OS}.install.sh
+    - . ${TRAVIS_BUILD_DIR}/.travis/install.sh
 before_script:
-    - mkdir build && cd build
-    - export CMAKE_FLAGS="-DWANT_QT5=$QT5 -DUSE_WERROR=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo"
-    - if [ -z "$TRAVIS_TAG" ]; then export CMAKE_FLAGS="$CMAKE_FLAGS -DUSE_CCACHE=ON"; fi
+    - . ${TRAVIS_BUILD_DIR}/.travis/before_script.sh
 script:
-    - . ${TRAVIS_BUILD_DIR}/.travis/${TRAVIS_OS_NAME}.${TARGET_OS}.script.sh
-    - make -j4
-    - if [[ $TARGET_OS != win* ]]; then make tests && ./tests/tests; fi;
+    - . ${TRAVIS_BUILD_DIR}/.travis/script.sh
 after_script:
     - ccache -s
 before_deploy: make package
@@ -49,4 +46,4 @@ notifications:
            - https://webhooks.gitter.im/e/1ac7fc698195981a9227
        on_success: change  # options: [always|never|change] default: always
        on_failure: always  # options: [always|never|change] default: always
-       on_start: never     # options: [always|never|change] default: always 
+       on_start: never     # options: [always|never|change] default: always
--- a/.travis/before_install.sh
+++ b/.travis/before_install.sh
@@ -1,0 +1,6 @@
+#!/usr/bin/env bash
+
+if [ "$TYPE" != 'style' ]; then
+  # shellcheck disable=SC1090
+  . "$TRAVIS_BUILD_DIR/.travis/$TRAVIS_OS_NAME.$TARGET_OS.before_install.sh"
+fi
--- a/.travis/before_script.sh
+++ b/.travis/before_script.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+
+if [ "$TYPE" != 'style' ]; then
+  mkdir build && cd build || exit
+
+  export CMAKE_FLAGS="-DWANT_QT5=$QT5 -DUSE_WERROR=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo"
+
+  if [ -z "$TRAVIS_TAG" ]; then export CMAKE_FLAGS="$CMAKE_FLAGS -DUSE_CCACHE=ON"; fi
+fi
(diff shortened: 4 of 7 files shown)


Spekular:
Conversely, I don't see a single reason to use spaces.

zapashcanon:
> Conversely, I don't see a single reason to use spaces.

Consistent display. Allow fine-grained alignment by tools like AStyle.

Spekular:
1. Set tabs to show as whatever length you want. 
2. Using tabs for indentation doesn't mean you can't use spaces for alignment.

zapashcanon:
> Set tabs to show as whatever length you want.

How do I set that on GitHub ? Plus, I don't want to set this for each project I'm working on, some are using 2 space tabs, some others are using 4 space tabs etc.

> Using tabs for indentation doesn't mean you can't use spaces for alignment.

Indeed, but it'll lead to error if you check for tabs indentation only. Or you allow tabs + space indentation, and it leads to a mess.

tresf:
**Tabs vs. Spaces**

 * LMMS thread https://github.com/LMMS/lmms/issues/3426
 * Satirical video:  https://www.youtube.com/watch?v=SsoOG6ZeyUI

I'm on the spaces side of the fence, personally, but I seem to be part of the minority (by a considerable margin).  Reasoning is the the way spaces display -- It is is consistent among every editor on earth.  Switching to 3 or 4 spaces is quite common in editors.  Tabs generally consume 8 characters, which makes this whole "80 char width" problem even worse.

**80 Char Width**

Most developers are fine going over, but about 1/4 of our active developers still prefer to limit it to 80 due to display preferences.  I know @jasp00 has stated he prefers 80 characters.  I'm less concerned with the decision... it really must be abolished.  I'm more concerned with how that will affect contributors like @jasp00.  When we make this decision, we need to know how it will impact our contributors and that they have a valid workaround.

> *`jasp00` said in #3071*
> @zapashcanon, I suggest you first implement a solution that follows current conventions as closely as possible and tell what rules you need to break. Other rules can be modified later.

Current conventions need to be changed.  These tasks are really coupled.  Unfortunate for @zapashcanon now his PR is going to be spammed with opinion.  Hopefully we can keep this civil.

zapashcanon:
> his PR is going to be spammed with opinion

Actually, I don't like most of the conventions (about `int* x` vs. `int * x` and `int *x`, or `{` stuff). Fortunately, this won't be discussed in this PR. because this aren't necessary changes to add KWStyle and AStyle. :) So, after two two above, it will be fine I hope haha.

tresf:
> I don't like most of the conventions (about `int*` ... )
> Fortunately, this won't be discussed in this PR

Well, now it will because you brought it up! :)

My vote is to put the modifier by the type, not the by the variable.

e.g. 

| Code | ✅  | ⛔️  |
|-------|------|-----|
|`int * x`|`int* x` | `int *x`.|
|`int [] x`|`int[] x` | `int x[]`.|

`</opinion>`

zapashcanon:
Looks like we agree on this too. ;)

tresf:
> and do some kinda poll in it so we'll be able to make some progress ?

We could but it's not a democracy.  I feel the active contributors get to decide.

zapashcanon:
Would you agree on giving me a list of those contributors so we ask them ? Because, otherwise, I feel like this will quickly get stuck.

tresf:
> Would you agree on giving me a list of those contributors so we ask them ? Because, otherwise, I feel like this will quickly get stuck.

The [broken](https://github.com/LMMS/lmms.io/pull/221) `CONTRIBUTORS` file has a good list.  Here's [how we did it before we broke it](https://github.com/LMMS/lmms/blob/stable-1.1/CMakeLists.txt#L360-L370). This only shows who's contributed, not how recent.  Sorting commits by name would help us understand if they developer is long gone or not.

BaraMGB:
The only thing I need are clear rules how to format the code. I don't prefer anything.

zapashcanon:
@BaraMGB, maybe this wasn't clear in the first post, but anyway, you'll have something like: `./astyle.sh` to do all the correct whitespace formatting changes for you. For the other changes, the CI won't pass if something is wrong with an error message to tell you the problem. Would it be ok for you ? We'd also update the wiki about coding conventions if they're changed I guess. :)

@tresf, I think I can find the list of contributors. ;) It's more like I don't feel like I should decide myself who is active enough to decide on this, as I'm not (currently at least) an active contributor. If you could clarify `editing code on a semi-regular basis`, it would help me. :) Or maybe, we could just let people decide if they consider themselves being active enough ?

tresf:
Hopefully this is a good start... 

Wallacoloo, Lukas-W, softrabbit, jasp00, Umcaruje, BaraMGB, grejppi, michaelgregorius, zonkmachine, liushuyu, tresf

Some other recent contributors whose opinion would weigh in:  karmux, Spekular, PaulBatchelor, follower, DeRobyJ, StCyr, zapashcanon

RebeccaDeField was left out because I'm making an assumption she has little investment in this topic. :)

zapashcanon:
> I stick with my advice

It was to stay as close as possible to current conventions, that's what I'm trying to do. All other things can stay as is, but those two can't imho.

> even less about all of them at the same time

That's why I opened separated issues.

> Indent with tabs

So it means no space indentation on line breaking. Which will just lead to have cases with something like 40 char. of indentation then 40 chars of real stuff, on this on many lines, making the code really hard to read.

> Either wrap to 80 characters

So we have currently 1415 lines of code with more than 80 char. If we don't use spaces, it won't help to fix this. 

> not handle any wrapping

If we do this for everything then, what's the point of using a formatter ?

Still, I don't see any valid reason to use tabs or to limit line length to 80 char.

Spekular:
Using tabs does NOT lead to misalignment because you can use spaces for this. If this creates errors then disable that check (because it's throwing errors for perfectly good code) or get a checker smart enough to allow alignment spaces.

KWStyle.xml
```diff
@@ -0,0 +1,131 @@
+<Project>LMMS</Project>
```
tresf: Please move this to `.travis`.  Also, I recommend making the filename lowercase, which is what we do for all non-class files.

tresf:
> Either wrap to 80 characters or not handle any wrapping.

@jasp00 if you can live with no (mandatory) wrapping, we can make that call now and close out #3471.

.travis/install.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+if [ "$TYPE" = 'style' ]; then
+  sudo apt install astyle kwstyle shellcheck
```
tresf: Is it irony that this file uses 2-character indents? 😄

jasp00:
> if you can live with no (mandatory) wrapping, we can make that call now and close out #3471.

One issue is the coding convention and another one is what a formatting tool should handle.

KWStyle.xml
zapashcanon: This is useless, I can just remove it if you want ?

Ok, I choosed this name because it was the one given as an example on the doc. Will do.

.travis/install.sh
zapashcanon: Wasn't intended haha.

zapashcanon:
I'm closing, will do it another way.

PR #3476, opened 2017-04-01
Add shellcheck to the build

zapashcanon:
Hi,

Just add a build called 'style', which will be used for kwstyle and astyle too (I hope at least).

Currently it's checking for each `.sh` file in `.travis/` and in `.cmake/`, others scripts are not ours so we won't check against them. If there are others scripts I forgot, just tell me.

I also moved some of the code in `.travis.yml` to separated `.sh` files as @jasp00 advised, so they can be checked (shellcheck found a few things in them btw) and also to ease the process of adding a "style build" which has nothing in common with other builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ cache:
         - $HOME/.ccache
 matrix:
     include:
+        - env: TYPE=style
         - env: TARGET_OS=win32
         - env: TARGET_OS=win64
         - os: osx
@@ -17,20 +18,9 @@ matrix:
         - env: QT5=True TARGET_OS=win64
         - os: osx
           env: QT5=True
-before_install:
-   - . ${TRAVIS_BUILD_DIR}/.travis/${TRAVIS_OS_NAME}.${TARGET_OS}.before_install.sh
-install:
-    - . ${TRAVIS_BUILD_DIR}/.travis/${TRAVIS_OS_NAME}.${TARGET_OS}.install.sh
-before_script:
-    - mkdir build && cd build
-    - export CMAKE_FLAGS="-DWANT_QT5=$QT5 -DUSE_WERROR=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo"
-    - if [ -z "$TRAVIS_TAG" ]; then export CMAKE_FLAGS="$CMAKE_FLAGS -DUSE_CCACHE=ON"; fi
-script:
-    - . ${TRAVIS_BUILD_DIR}/.travis/${TRAVIS_OS_NAME}.${TARGET_OS}.script.sh
-    - make -j4
-    - if [[ $TARGET_OS != win* ]]; then make tests && ./tests/tests; fi;
-after_script:
-    - ccache -s
+install: ${TRAVIS_BUILD_DIR}/.travis/install.sh
+script: ${TRAVIS_BUILD_DIR}/.travis/script.sh
+after_script: ${TRAVIS_BUILD_DIR}/.travis/after_script.sh
 before_deploy: make package
 deploy:
     provider: releases
@@ -43,10 +33,3 @@ deploy:
         all_branches: true
         condition: '("$TARGET_OS" = win??) && "$QT5"'
     repo: LMMS/lmms
-notifications:
-    webhooks:
-       urls:
-           - https://webhooks.gitter.im/e/1ac7fc698195981a9227
-       on_success: change  # options: [always|never|change] default: always
-       on_failure: always  # options: [always|never|change] default: always
-       on_start: never     # options: [always|never|change] default: always 
--- a/.travis/install.sh
+++ b/.travis/install.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -e
+
+if [ "$TYPE" = 'style' ]; then
+	sudo apt-get -yqq update
+	sudo apt-get install shellcheck
+else
+	"$TRAVIS_BUILD_DIR/.travis/$TRAVIS_OS_NAME.$TARGET_OS.before_install.sh"
+	"$TRAVIS_BUILD_DIR/.travis/$TRAVIS_OS_NAME.$TARGET_OS.install.sh"
+fi
--- a/.travis/script.sh
+++ b/.travis/script.sh
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+
+set -e
+
+if [ "$TYPE" = 'style' ]; then
+
+	# shellcheck disable=SC2046
+	shellcheck $(find -O3 "$TRAVIS_BUILD_DIR/.travis/" "$TRAVIS_BUILD_DIR/cmake/" -type f -name '*.sh' -o -name "*.sh.in")
+
+else
+
+	mkdir build
+	cd build
+
+	export CMAKE_FLAGS="-DWANT_QT5=$QT5 -DUSE_WERROR=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo"
+
+	if [ -z "$TRAVIS_TAG" ]; then
+		export CMAKE_FLAGS="$CMAKE_FLAGS -DUSE_CCACHE=ON"
+	fi
+
+	"$TRAVIS_BUILD_DIR/.travis/$TRAVIS_OS_NAME.$TARGET_OS.script.sh"
+
+	make -j4
+
+	if [[ $TARGET_OS != win* ]]; then
+
+		make tests
+		tests/tests
+
+	fi
+fi
--- a/.travis/after_script.sh
+++ b/.travis/after_script.sh
@@ -1,0 +1,7 @@
+#!/usr/bin/env bash
+
+set -e
+
+if [ "$TYPE" != 'style' ]; then
+	ccache -s
+fi
--- a/.travis/linux..before_install.sh
+++ b/.travis/linux..before_install.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -e
+
 sudo add-apt-repository ppa:andrewrk/libgroove -y
 sudo sed -e "s/trusty/precise/" -i \
 	/etc/apt/sources.list.d/andrewrk-libgroove-trusty.list
--- a/.travis/linux..install.sh
+++ b/.travis/linux..install.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -e
+
 PACKAGES="cmake libsndfile-dev fftw3-dev libvorbis-dev  libogg-dev
 	libasound2-dev libjack-dev libsdl-dev libsamplerate0-dev libstk0-dev
 	libfluidsynth-dev portaudio19-dev wine-dev g++-multilib libfltk1.3-dev
--- a/.travis/linux..script.sh
+++ b/.travis/linux..script.sh
@@ -1,4 +1,6 @@
 #!/usr/bin/env bash
 
+set -e
+
 # shellcheck disable=SC2086
 cmake -DUSE_WERROR=ON $CMAKE_FLAGS ..
--- a/.travis/linux.win.download.sh
+++ b/.travis/linux.win.download.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+
 set -e
 
 CACHE_DIR=$TRAVIS_BUILD_DIR/apt_mingw_cache/$1
--- a/.travis/linux.win32.before_install.sh
+++ b/.travis/linux.win32.before_install.sh
@@ -1,4 +1,6 @@
 #!/usr/bin/env bash
 
+set -e
+
 sudo add-apt-repository ppa:tobydox/mingw-x-trusty -y
 sudo apt-get update -qq
--- a/.travis/linux.win32.install.sh
+++ b/.travis/linux.win32.install.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+
 set -e
 
 if [ "$QT5" ]; then
@@ -15,10 +16,7 @@ MINGW_PACKAGES="mingw32-x-sdl mingw32-x-libvorbis mingw32-x-fluidsynth mingw32-x
 
 export MINGW_PACKAGES
 
-DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-
-# shellcheck disable=SC1090
-. "$DIR/linux.win.download.sh" win32
+"$TRAVIS_BUILD_DIR/.travis/linux.win.download.sh" win32
 
 PACKAGES="nsis cloog-isl libmpc3 qt4-linguist-tools mingw32 $MINGW_PACKAGES"
 
@@ -29,6 +27,6 @@ sudo apt-get install -y $PACKAGES
 # to use @file command line passing, which in turn ccache 3.1.9 doesn't support
 pushd /tmp
 wget http://archive.ubuntu.com/ubuntu/pool/main/c/ccache/ccache_3.2.4-1_amd64.deb
-sha256sum -c $TRAVIS_BUILD_DIR/.travis/ccache.sha256
+sha256sum -c "$TRAVIS_BUILD_DIR/.travis/ccache.sha256"
 sudo dpkg -i ccache_3.2.4-1_amd64.deb
 popd
--- a/.travis/linux.win32.script.sh
+++ b/.travis/linux.win32.script.sh
@@ -1,4 +1,6 @@
 #!/usr/bin/env bash
 
+set -e
+
 export CMAKE_OPTS="$CMAKE_FLAGS -DUSE_WERROR=ON"
 ../cmake/build_mingw32.sh
--- a/.travis/linux.win64.before_install.sh
+++ b/.travis/linux.win64.before_install.sh
@@ -1,3 +1,5 @@
 #!/usr/bin/env bash
 
-. .travis/linux.win32.before_install.sh
+set -e
+
+"$TRAVIS_BUILD_DIR/.travis/linux.win32.before_install.sh"
--- a/.travis/linux.win64.install.sh
+++ b/.travis/linux.win64.install.sh
@@ -1,11 +1,9 @@
 #!/usr/bin/env bash
+
 set -e
 
 # First, install 32-bit deps
-DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-# shellcheck disable=SC1090
-. "$DIR/linux.win32.install.sh"
-
+"$TRAVIS_BUILD_DIR/.travis/linux.win32.install.sh"
 
 if [ "$QT5" ]; then
 	MINGW_PACKAGES="mingw64-x-qt5base"
@@ -21,9 +19,7 @@ MINGW_PACKAGES="mingw64-x-sdl mingw64-x-libvorbis mingw64-x-fluidsynth mingw64-x
 
 export MINGW_PACKAGES
 
-DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-# shellcheck disable=SC1090
-. "$DIR/linux.win.download.sh" win64
+"$TRAVIS_BUILD_DIR/.travis/linux.win.download.sh" win64
 
 # shellcheck disable=SC2086
 sudo apt-get install -y $MINGW_PACKAGES
--- a/.travis/linux.win64.script.sh
+++ b/.travis/linux.win64.script.sh
@@ -1,4 +1,6 @@
 #!/usr/bin/env bash
 
+set -e
+
 export CMAKE_OPTS="$CMAKE_FLAGS -DUSE_WERROR=ON"
 ../cmake/build_mingw64.sh
--- a/.travis/osx..before_install.sh
+++ b/.travis/osx..before_install.sh
@@ -1,3 +1,5 @@
 #!/usr/bin/env bash
 
+set -e
+
 brew update
--- a/.travis/osx..install.sh
+++ b/.travis/osx..install.sh
@@ -1,20 +1,40 @@
 #!/usr/bin/env bash
 
-PACKAGES="cmake pkgconfig fftw libogg libvorbis libsndfile libsamplerate jack sdl stk portaudio node fltk"
+set -e
+
+PACKAGES="cmake pkg-config fftw libogg libvorbis libsndfile libsamplerate jack sdl stk portaudio node fltk"
 
 if [ "$QT5" ]; then
-	PACKAGES="$PACKAGES homebrew/versions/qt55"
+	PACKAGES="$PACKAGES qt@5.5"
 else
-	PACKAGES="$PACKAGES cartr/qt4/qt"
+	PACKAGES="$PACKAGES cartr/qt4/qt@4"
 fi
 
+if "${TRAVIS}"; then
+   PACKAGES="$PACKAGES ccache"
+fi
+
+# removing already installed packages from the list
+for p in $(brew list); do
+	PACKAGES=${PACKAGES//$p/}
+done;
+
 # shellcheck disable=SC2086
-brew install $PACKAGES ccache
+brew install $PACKAGES
 
 # Recompile fluid-synth without CoreAudio per issues #649
 # Changes to fluid-synth.rb must be pushed to URL prior to use
-url=$(git remote get-url origin)
-branch=$(git symbolic-ref --short HEAD)
-brew install --build-from-source "$url/raw/$branch/cmake/apple/fluid-synth.rb"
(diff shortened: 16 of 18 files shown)


.travis.yml
```diff
@@ -49,4 +46,4 @@ notifications:
            - https://webhooks.gitter.im/e/1ac7fc698195981a9227
```
tresf: We don't use gitter anymore.  We can probably shut this down.

.travis/before_script.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+if [ "$TYPE" != 'style' ]; then
+  mkdir build && cd build || exit
+
+  export CMAKE_FLAGS="-DWANT_QT5=$QT5 -DUSE_WERROR=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo"
+
+  if [ -z "$TRAVIS_TAG" ]; then export CMAKE_FLAGS="$CMAKE_FLAGS -DUSE_CCACHE=ON"; fi
```
tresf: This was likely kept a one-liner to keep the `.yml` clean reading.  Please break this out now that it's in a shell script. :)

.travis/install.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+if [ "$TYPE" = 'style' ]; then
+  sudo apt install shellcheck
```
tresf: Should this be `apt-get` ?

tresf:
Would it make more sense to force this as a Linux-environment run?  In the current state, it's assuming items like `apt` are available, which really shouldn't happen on anything except Linux environments.

.travis/install.sh
zapashcanon: From the debian admin. handbook:

> APT is a vast project, whose original plans included a graphical interface. It is based on a library which contains the core application, and apt-get is the first front end — command-line based — which was developed within the project. apt is a second command-line based front end provided by APT which overcomes some design mistakes of apt-get.

IMHO using `apt` is better.

.travis.yml
zapashcanon: I removed the whole section as it looked like it was only for gitter.

.travis/before_script.sh
zapashcanon: Done.

zapashcanon:
> to force this as a Linux-environment

It's taking the default from `.travis.yml` which is `dist: trusty` so I don't think it's necessary.

.travis/install.sh
tresf: Thanks.  We use `apt-get` everywhere else, let's use that for consistency.   If there turn out to be advantages using `apt` over `apt-get`, we'll switch the whole project to use it (wiki, travis, etc).

tresf:
> It's taking the default from .travis.yml which is dist: trusty so I don't think it's necessary.

Fair enough, but these should still only live in the linux scripts and you can just add [another block](https://github.com/LMMS/lmms/blob/master/.travis/linux..install.sh#L11) to the apt-get logic rather than segmenting them which also addresses https://github.com/LMMS/lmms/pull/3476/files#r109296139.

.travis/install.sh
zapashcanon: Done.

tresf:
Adding them to Linus may also benefit us if some of these tools shift from Linux to homebrew, etc as history has proven sometimes non-Ubuntu packages [sometimes run newer](https://github.com/LMMS/lmms/pull/3208#issuecomment-274851256).

zapashcanon:
> these should still only live in the linux scripts and you can just add another block to the apt-get logic rather than segmenting them

So, you wan't me only to merge the `apt-get` part ? I'm not sure because, for others files, like `.travis/script.sh` they're not merged, would make it harder to understand. So I'd rather merge all of them, which will be more complicated in the case of `.travis/script.sh`, or even better, not merge them at all and stay like this which seems clearer to me.

> which also addresses https://github.com/LMMS/lmms/pull/3476/files#r109296139.

When I click here, I only get a diff. Not sure what you're talking about here.

> Adding them to Linus may also benefit us if some of these tools shift from Linux to homebrew, etc as history has proven sometimes non-Ubuntu packages sometimes run newer.

I don't understand sorry. Why adding them to linux would help if one day there's a more up to date homebrew version ?

tresf:
> would make it harder to understand

I thought of that too.  I think it's fine.  That way all CI platforms can have the formatting if needed.  It's more granular.

> for others files, like `.travis/script.sh`

The `.yml` really can go back.  I don't see a reason they can't live directly in the platforms scripts that they run on.

> I don't understand sorry. 

I realized that and had edited my post. :)  The point was if a utility that we're using for style or code quality is dated, often the Homebrew environment will get the bleeding edge first.  This has happened time and time again (Qt5, GigPlayer, etc) so having `apt-get` calls in the root of `install` when they aren't platform independent seems like it could have scalability problem if we switch.  Just a forethought.

> When I click here, I only get a diff. 

It got squashed out.  Was the `apt-get` comment, which has since been fixed anyway. 👍

zapashcanon:
> I thought of that too. I think it's fine. That way all CI platforms can have the formatting if needed. It's more granular.

Sorry, I don't understand what you mean. But, what would you all the CI platforms to have the formatting ? And I don't understand what you want me to do.

> The .yml really can go back. I don't see a reason they can't live directly in the platforms scripts that they run on.

Because now, shellcheck is going to check those files. In `.travis/before_script.sh` as an example it spotted: `mkdir build && cd build` and told me to replace it by: `mkdir build && cd build || exit` (in case cd fails).

My point is really that we have `build` builds for many platform and a `style` build which should be separated. If someday we decide we don't want to run this `style` build on trusty, we will just have to do something like:

```yml
        - os: osx
          env: TYPE=style
```

And just edit the `install.sh` file. Seems more clear to me than mixing build builds and style build.

tresf:
I think the `|| exit` could  be improved.  We should return non-zero, probably `set -e`

The other points make sense and you're right they should stay for now.  Thanks!

.travis.yml
```diff
@@ -18,17 +19,13 @@ matrix:
...
-    - export CMAKE_FLAGS="-DWANT_QT5=$QT5 -DUSE_WERROR=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo"
-    - if [ -z "$TRAVIS_TAG" ]; then export CMAKE_FLAGS="$CMAKE_FLAGS -DUSE_CCACHE=ON"; fi
+    - . ${TRAVIS_BUILD_DIR}/.travis/before_script.sh
 script:
-    - . ${TRAVIS_BUILD_DIR}/.travis/${TRAVIS_OS_NAME}.${TARGET_OS}.script.sh
-    - make -j4
-    - if [[ $TARGET_OS != win* ]]; then make tests && ./tests/tests; fi;
+    - . ${TRAVIS_BUILD_DIR}/.travis/script.sh
 after_script:
     - ccache -s
```
jasp00: You should move this command to `.travis/script.sh`.

.travis/before_install.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+if [ "$TYPE" != 'style' ]; then
+  # shellcheck disable=SC1090
+  . "$TRAVIS_BUILD_DIR/.travis/$TRAVIS_OS_NAME.$TARGET_OS.before_install.sh"
+fi
```
jasp00: You may merge this script with `.travis/install.sh`.

.travis/script.sh
```diff
@@ -0,0 +1,21 @@
...
+
+  # shellcheck disable=SC1090
+  . "$TRAVIS_BUILD_DIR/.travis/$TRAVIS_OS_NAME.$TARGET_OS.script.sh"
+
+  make -j4
+
+  if [[ $TARGET_OS != win* ]]; then
+
+    make tests
+    ./tests/tests
```
jasp00: Why the dot?

.travis.yml
Lukas-W: No you shouldn't. Moving it to `script` will cause it to affect the build result.

jasp00: >Moving it to `script` will cause it to affect the build result.

I am assuming @zapashcanon will drop sourcing and act accordingly. `ccache -s` prints a statistics summary, are you worried that `ccache` may fail?

Lukas-W: I don't see how sourcing relates to this. 

> are you worried that ccache may fail?

If it fails, I don't want our CI to fail because of that, obviously.

Regardless, it doesn't belong in `script` because it's not part of our tests.

jasp00: @zapashcanon, you should move this command to `.travis/after_script.sh` and execute `ccache` when appropriate.

.travis.yml
```diff
@@ -18,17 +19,13 @@ matrix:
         - os: osx
           env: QT5=True
 before_install:
-   - . ${TRAVIS_BUILD_DIR}/.travis/${TRAVIS_OS_NAME}.${TARGET_OS}.before_install.sh
+    - . ${TRAVIS_BUILD_DIR}/.travis/before_install.sh
```
zapashcanon: I'm not sure what you mean. Moving everything related to `before_install` to `install` and then remove `before_install` ?

.travis.yml
zapashcanon: By appropriate, you mean not when running the `style` build right ?

.travis/script.sh
zapashcanon: No idea, that's how it was before in `travis.yml`.

.travis/before_script.sh
```diff
@@ -0,0 +1,11 @@
...
+
+if [ "$TYPE" != 'style' ]; then
+  mkdir build && cd build || exit
+
+  export CMAKE_FLAGS="-DWANT_QT5=$QT5 -DUSE_WERROR=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo"
+
+  if [ -z "$TRAVIS_TAG" ]; then
+    export CMAKE_FLAGS="$CMAKE_FLAGS -DUSE_CCACHE=ON"
+  fi
+fi
```
zapashcanon: Done.

.travis/before_install.sh
zapashcanon: Done.

.travis/before_install.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+if [ "$TYPE" != 'style' ]; then
+  # shellcheck disable=SC1090
+  . "$TRAVIS_BUILD_DIR/.travis/$TRAVIS_OS_NAME.$TARGET_OS.before_install.sh"
```
zapashcanon: Done for the indenting part.

zapashcanon: How do you want me to remove sourcing ? I could do:

`chmod +x file && ./file` or `bash file`, no idea what's best.

zapashcanon:
@jasp00, some of your comments are now on outdated code but not fixed. Just to make sure you don't miss my comments.

.travis.yml
jasp00: I guess so. If ShellCheck does not compile, then `ccache` makes no sense.

.travis/before_install.sh
jasp00: You should set permissions in the repository, like `cmake/build_mingw32.sh`.

.travis/script.sh
jasp00: I see. Plain `tests/tests` should work.

.travis.yml
jasp00: `install: ${TRAVIS_BUILD_DIR}/.travis/install.sh`

.travis.yml
zapashcanon: Seems a useless to me. `ccache` indeed makes no sense, but it'll just print: "nothing changed" and it's done. Is it worth to add a ~10 lines script when keeping it in one line is the same ?

.travis/script.sh
zapashcanon: Done.

.travis/before_install.sh
zapashcanon: I'm not sure what you meant, so I just did a chmod.

.travis.yml
zapashcanon: Doesn't work, have to do a chmod or call bash directly.

zapashcanon:
@tresf @jasp00, I think I'm done with each review. Anything missing ?

tresf:
@zapashcanon the `chmod +x` technique is fine.  Have we also considered trying fixing this flag [on the file itself](http://stackoverflow.com/questions/21691202/how-to-create-file-execute-mode-permissions-in-git-on-windows)?

zapashcanon:
@tresf setting this flag seems interesting. Do you want me to do it ? Otherwise, I'd like this to get merged so I can start adding things to the style job.

.travis.yml
jasp00: It will work if you set execute permissions.

.travis.yml
jasp00: If `ccache -s` output is irrelevant, then you should not run `ccache -s`. By ~10 lines, do you mean 2? No matter how many lines, it is worth because ShellCheck will check the script.

jasp00:
>@tresf setting this flag seems interesting. Do you want me to do it ?

Please, set the flag and drop `chmod` calls.

.travis.yml
zapashcanon: Yep', so if you want it to fit in a single line I'll have to do something like: `chmod +x files && file`, here, using `bash file` looked better to me.

jasp00: Do you know how to [set permissions](http://stackoverflow.com/questions/21691202/how-to-create-file-execute-mode-permissions-in-git-on-windows) in the repository or not?

tresf:
@zapashcanon thanks.  I know some of these things seem mundane, but it makes the feature cleaner as a whole and sets a good example (such as the execute flag).  Thanks for the patience with this.

.travis/install.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+if [ "$TYPE" = 'style' ]; then
+	sudo apt-get -yqq update
+	sudo apt-get install shellcheck
+else
+
```
tresf: Minor, but there seem to be some additional newlines in this `if [ ]` block as well as the the `else` block in the `install.sh` file.

.travis.yml
zapashcanon: Done.

.travis.yml
zapashcanon: Yep, just did it. I didn't know it was possible to do this with git.

.travis/install.sh
zapashcanon: Fixed.

zapashcanon:
@tresf, it's ok, I understand. I'm even glad to know now that git is able to handle permissions.

.travis/install.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
```
jasp00: You should add `set -e` to all scripts except `after_script.sh`.

.travis/script.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+
+if [ "$TYPE" = 'style' ]; then
+
+	# shellcheck disable=SC2046
+	shellcheck $(find -O3 "$TRAVIS_BUILD_DIR/.travis/" "$TRAVIS_BUILD_DIR/cmake/" -type f -name '*.sh')
+
+else
+
+	mkdir build && cd build || exit
```
jasp00: Could you use simple commands and drop `exit`?

.travis/install.sh
zapashcanon: Done.

.travis/script.sh
zapashcanon: `mkdir build && cd build` was here before in the `.yml`. I added the exit but with `set -e` it's useless now.

.travis/install.sh
zapashcanon: I'm not sure why I shouldn't add it into `after_script.sh`. If `after_script.sh` return a non-0 exit code, Travis won't say the build failed. That's exactly the purpose of the `after_script` in the `.yml` file.

tresf: It won't hurt in `after_script` but to @jasp00's point, the `ccache` statistics aren't really critical for a pass/fail and it's a single-line, so `set -e` on the whole script hasn't yet been warranted over something like `exit $?`.  But @zapashcanon you're right, it's best to add it.

zapashcanon: Ok, done.

jasp00: As I see `after_script`, it is useful for statistics, non-critical tests, custom notifications (not webhooks), dumps, etc. You probably want to run all of them, so `set -e` is not convenient. Right now, it does not matter.

jasp00: Could you add `set -e` to `linux..before_install.sh`, `linux.win32.before_install.sh`, `linux.win32.install.sh`, etc.?

zapashcanon: Just did it.

tresf:
@zapashcanon Homebrew seems to be picky about the install command.  If a package is already installed, it returns a non-zero code and bombs out the script.  This is now exposed by `set -e`.

I'd recommend we keep the package listing incase we need to invoke this from another build system and make it variable.

```bash
# skip preinstalled packages
if ! "${TRAVIS}"; then
   PACKAGES="$PACKAGES cmake pkgconfig node"
fi
```

Alternately, I believe `brew reinstall` will return warning instead of error, but it also adds unecessary time reinstalling packages that are already installed.

zapashcanon:
@tresf, I tried something else, which seems better to me. Currrently not working, but I'll fix it tomorrow. Do you agree with the idea ? I'm getting the list of already installed packages from `$PACKAGES`, then I remove them from `$PACKAGES` and finally install.

Moreover, there's this error too: `Error: No such keg: /usr/local/Cellar/fftw`. I've no idea what this is as I (almost) don't use MacOS  or homebrew. Any clue ?

tresf:
> Do you agree with the idea 

Absolutely.  This is what I find recommended most places.  It could even be extended to install them one-by-one and remove the bleeding $FOO, which would probably make @jasp00 happy.

> `No such keg: /usr/local/Cellar/fftw`

It's simply the first already-installed package... Yet another side-effect of `set -e` I'm afraid.  I think we need to do something like `p in PACKAGES` and catch errors from `brew list`  (`|| true` or whatever).   Probably don't even need to concatenate, just use the conditional to call `brew install` one by one.  Man what a shitty side effect this `!0` is causing. :)

zapashcanon:
> to install them one-by-one

I thought about that, but I was afraid it would make the installation process longer.

tresf:
> I thought about that, but I was afraid it would make the installation process longer.

It shouldn't.

tresf:
I was wrong @zapashcanon... it does make a difference.  

Here's my benchmarking

| `foo bar` | `foo && bar` |
|----------|----------------|
| 9.357s  | 12.21s  |
| 7.596s  |  10.598s  |
| 8.261s  | 12.209  |

Probably best to build out the string in this case.

zapashcanon:
@tresf, I replaced `for p in $(brew list $PACKAGES); do` by `for p in $(brew list); do`. It worked. I also had to replace `pkgconfig` by `pkg-config`.

Now it fails on: `branch=$(git symbolic-ref --short HEAD)`, I tried something but it didn't work. I'm not sure how to fix this. As you're the one who wrote this, have you any idea ?

tresf:
@zapashcanon the git command should be valid, it just gets the current branch name.  Custom homebrew formulas are hosted `.rb` files, so we have to point to the GitHub URL and that part makes sure it's the `stable-1.2`, `master`, etc.

I'm guessing the script just doesn't know where the git repository is.  Can you try something like `push ${TRAVIS_BUILD_DIR}` (or `cd`, etc) and see if that resolves it?

zapashcanon:
I tried `cd $TRAVIS_BUILD_DIR`, didn't work.

I'm not sure the command is valid, see [this](https://travis-ci.org/LMMS/lmms/jobs/221617670#L1118). Looks like we're in a detached HEAD state, no ? I don't know how travis get the git repo. etc.

zapashcanon:
[git-symbolic-ref](https://git-scm.com/docs/git-symbolic-ref):

> git symbolic-ref will exit with status 0 if the contents of the symbolic ref were printed correctly, with status 1 if the requested name is not a symbolic ref, or 128 if another error occurs.

Travis build [log](https://travis-ci.org/LMMS/lmms/jobs/222136614#L1204):

> fatal: ref HEAD is not a symbolic ref
> 
> The command "${TRAVIS_BUILD_DIR}/.travis/install.sh" failed and exited with 128 during .

tresf:
@zapashcanon it appears weren't not the only ones wanting this.

Can we add a fallback to use `${TRAVIS_PULL_REQUEST_BRANCH}`?  I say fallback because --
 despite the home -- this script may some day be run outside of a travis environment.  e.g. Our Mac tutorial [currently recommends it](https://github.com/LMMS/lmms/wiki/Compiling-lmms-(Apple)#homebrew) although I'll soon be removing this in favor of a [more comprehensive quickstart](https://github.com/tresf/lmms/wiki/Compiling#dependencies-quickstart).

Back on topic... here's the Travis-CI documentation around the branch name:
https://github.com/travis-ci/travis-ci/issues/1633#issuecomment-250500078

zapashcanon:
Ok, so, from the latest test:

```sh
pwd:

/Users/travis/build/LMMS/lmms

url=https://github.com/LMMS/lmms.git

git branch:

* (HEAD detached at FETCH_HEAD)

  master

git status:

HEAD detached at FETCH_HEAD

nothing to commit, working tree clean

fatal: ref HEAD is not a symbolic ref
```

So `git symbolic-ref --short HEAD` doesn't work because we're in detached head state, I've got no idea how to fix this properly.

I could do something like:

```sh
if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
    branch="master"
else
    branch=$(git symbolic-ref --short HEAD)
fi
```

But it won't work for PR to `stable-1.2` or stuff like this.

tresf:
I think we're on the same page.  It is paramount we know the branch name.  If it becomes too much if a struggle we can research how Homebrew does local `.rb` files or hardlink a gist.

zapashcanon:
@tresf, it works now, it's a bit hacky but it should do the trick.

tresf:
That's perfect.

zapashcanon:
So, do you see anything left to do before merging ?

tresf:
@zapashcanon I found a script that we missed... `cmake/apple/install_apple.sh.in` (due to the non-matching `*.sh`).  We can append ` -or -name '*.sh.in'` to the `find` command to add it.  I just tried it there are going to be quite a few patches.

I'm not sure how well `shellcheck` does with unconfigured cmake files but at a glance, the cmake variables don't seem to be causing a problem.

zapashcanon:
@tresf, I fixed everything except one thing which seems strange to me.

We have:

```sh
STK_RAWWAVE=$HOME/stk-*/rawwaves

if [ ! -d "$STK_RAWWAVE" ]; then
   STK_RAWWAVE=$(brew --prefix stk)/share/stk/rawwaves
fi
```

Shellcheck doesn't like that. What's the `*` here for ? Is the folder name composed of some version number ? I did a quick fix with:

```sh
STK_RAWWAVE_LIST=( $HOME/stk-*/rawwaves )
STK_RAWWAVE=${STK_RAWWAVE_LIST[0]}

if [ ! -d "$STK_RAWWAVE" ]; then
   STK_RAWWAVE=$(brew --prefix stk)/share/stk/rawwaves
fi
```

But I don't know what's this code and I probably can't test it so...

tresf:
@zapashcanon your assumptions are exactly correct.  The purpose is a bit more vague... perhaps it could benefit from a comment but the raw waves must be copied at bundle time and at the time of writing that script, MacPorts didn't have libstk available.

Since then, we've moved to Homebrew, so I kept the home-built rawwaves directory for backwards compatibility.

zapashcanon:
Ok, so what should I do ? Is the current script ok ?

tresf:
@zapashcanon sorry I was giving a background story on it all.  It looks good.  Merging.  Note, to other devs... we will not be back-porting this to stable-1.2.

zapashcanon:
Thanks for the review @tresf and @jasp00 ! :)

PhysSong:
@tresf @jasp00 shellcheck now fails because the trusty image used by Travis CI got updated.
See [this](https://blog.travis-ci.com/2017-06-21-trusty-updates-2017-Q2-launch).

tresf:
@zapashcanon spearheaded this PR, I'll see if he's interested in fixing it.  Seems like the [documentation is straightforward](https://github.com/koalaman/shellcheck/wiki/SC2181).

zapashcanon:
I'll open a PR soon, should be easy.

Issue #3473, opened 2017-04-01
Drop #shellcheck comments


jasp00:
Follows #3455.

Scripts should not have `# shellcheck disable` comments and multiple commands in `.travis.yml` should be moved to separate scripts that could be checked.

tresf:
Please elaborate.  @zapashcanon and myself both feel #3455 is in good shape.  Please specifically point out the major problem that the PR creates.

zapashcanon:
> multiple commands in .travis.yml should be moved to separate scripts that could be checked.

Done in #3470.

jasp00:
>Please specifically point out the major problem that the PR creates.

While the PR works and it is a fine addition, it introduces the need for `# shellcheck disable`; a smart tool would not need such comments. The more or less important disadvantages are:
1. Readability, the comment does not explain the code.
2. Dependency on a checking tool.
3. A newcomer must know (and an old developer must remember) that such comment is needed for common and legal uses of variables.

tresf:
> Readability, the comment does not explain the code.

This happens fairly often in programming languages (`CMAKE_POLICY`, `@SupressWarnings`, `#pragma) , bash just doesn't have a better way to do it.  Adding more scripts is not the answer.

> Dependency on a checking tool.

This script has no dependency on any checking tools.  This argument is invalid.

> A newcomer must know (and an old developer must remember) that such comment is needed for common and legal uses of variables.

If not, Travis will (eventually) warn them once we get it into our CI.  This is intended.

Please provide a better argument.  The time spent discussing such this topic adds little value.  It should be put in more useful places.

jasp00:
>This happens fairly often in programming languages

Which does not make it less of a disadvantage. You are citing workarounds, `CMAKE_POLICY` is a temporary solution while a real fix is being developed.

>Adding more scripts is not the answer.

Answer to what?

>This script has no dependency on any checking tools.

What happens if you remove the comment?

>Travis will (eventually) warn them once we get it into our CI.

May you post the shellcheck error?

>This is intended.

This error could be avoided if shellcheck were smarter, so I am proposing an enhancement.

tresf:
> This error could be avoided if shellcheck were smarter, so I am proposing an enhancement.

Complain to shellcheck.  This is staying as-is.

jasp00:
>Complain to shellcheck.

Someone may.

>This is staying as-is.

Why? You could disable these checks from the command-line invocation.

@zapashcanon, may you post the shellcheck error?

[...]

PR #3477, opened 2017-04-02
Document ShellCheck comments issue

jasp00:
Documents #3473.


tresf:
This is superfluous, safe and the shellcheck-happy version is only going to make this script harder to read.  Closing.

PR #3478, opened 2017-04-02
Document SC1090 issue

jasp00:
SC1090 cases can be fixed if the included file does not provide elements for the running script.


zapashcanon:
I just removed every sourcing in the PR to add ShellCheck to the build.

PR #3479, opened 2017-04-02
Document optimization regarding std::remove_if

jasp00:
Since global automation track is never muted, this use of `std::remove_if` can be optimized. Related header may be dropped.


tresf:
Superseded by 96a0030, which removed `std::remove_if` code, making comment obsolete.

PR #3486, opened 2017-04-04
Do not use cache on release

jasp00:
On release, packages from the cache should not be used and ccache should not be downloaded.

### DIFF
--- a/.travis/linux.win32.install.sh
+++ b/.travis/linux.win32.install.sh
@@ -17,8 +17,10 @@ export MINGW_PACKAGES
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
-# shellcheck disable=SC1090
-. "$DIR/linux.win.download.sh" win32
+if [ -z "$TRAVIS_TAG" ]; then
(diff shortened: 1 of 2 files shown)


.travis/linux.win32.install.sh
```diff
@@ -27,8 +29,10 @@ sudo apt-get install -y $PACKAGES
 
 # ccache 3.2 is needed because mingw32-x-gcc is version 4.9, which causes cmake
 # to use @file command line passing, which in turn ccache 3.1.9 doesn't support
```
tresf: Since `TRAVIS_TAG` isn't immediately associated with `ccache`, we should have a comment explaining why `ccache` is disabled for tagged releases.

Lukas-W:
I don't see any harm in leaving it the way it is. Please explain why you think this is necessary, @jasp00.

jasp00:
Since ccache is not used, skipping ccache steps is an optimization. About ignoring the cache, I was anticipating enhancements; I will drop that part.

Lukas-W:
Closing as there's been no update in almost three months.

PR #3496, opened 2017-04-09
Ladspa plugin small memory leak fix

devnexen:


jasp00:
This patch is wrong because you are not supposed to free `const` memory. You should try a different approach.

plugins/LadspaEffect/calf/src/plugin.cpp
```diff
@@ -334,9 +334,13 @@ void ladspa_plugin_metadata_set::prepare(const plugin_metadata_iface *md, LADSPA
     param_count = md->get_param_count(); // XXXKF ladspa_instance<Module>::real_param_count();
     
     const ladspa_plugin_info &plugin_info = md->get_plugin_info();
+	std::string strName(plugin_info.name);
```
jasp00: You should use spaces in this file.

plugins/LadspaEffect/calf/src/plugin.cpp
```diff
@@ -334,9 +334,13 @@ void ladspa_plugin_metadata_set::prepare(const plugin_metadata_iface *md, LADSPA
...
+	strName += " LAPSDA";
+	char *Name = new char[strName.size() + 1];
+	strcpy(Name, strName.c_str());
     descriptor.UniqueID = plugin_info.unique_id;
     descriptor.Label = plugin_info.label;
-    descriptor.Name = strdup((std::string(plugin_info.name) + " LADSPA").c_str());
```
jasp00: `strdup` is fine. You just need to save the result somewhere else. It is not as efficient, but correctness is more important.

plugins/LadspaEffect/calf/src/plugin.cpp
```diff
@@ -467,13 +475,16 @@ ladspa_plugin_metadata_set::~ladspa_plugin_metadata_set()
     delete []descriptor.PortNames;
     delete []descriptor.PortDescriptors;
     delete []descriptor.PortRangeHints;
+	delete []descriptor.Name;
```
jasp00: You are not supposed to delete `const` memory.

plugins/LadspaEffect/calf/src/plugin.cpp
```diff
@@ -467,14 +469,17 @@ ladspa_plugin_metadata_set::~ladspa_plugin_metadata_set()
...
 #if USE_DSSI
     if (presets)
         presets->clear();
     if (preset_descs)
         preset_descs->clear();
     delete presets;
     delete preset_descs;
+    free(descriptorForDssiName);
 #endif
+    delete metadata;
```
jasp00: `metadata` is `const`. You should go to the point where `metadata` is created, make sure you keep a reference, and release it on destruction; you may need a function with `__attribute__((destructor))`.

tresf:
@devnexen I'm sorry this took so long to review, the original reviewer is no longer with the project.  Calf has been moved to a [submodule](https://github.com/lmms/veal) since #3940 and I missed this patch.

Currently @JohannesLorenz is working on getting Calf updated to 0.90, so this patch could have a home in either the current 0.18~ish version on the `ladspa` branch or the 0.90 version on the `master` branch.

Closing since the code this PR is against is no longer part of the LMMS codebase.

PR #3502, opened 2017-04-13
Simpler SF2 Player file dialog.

karmux:
Copied from GIG Player. Opens correct SF2 plugin directory when running LMMS from build directory without install. Used to open LMMS build directory before.

### DIFF
--- a/plugins/sf2_player/sf2_player.cpp
+++ b/plugins/sf2_player/sf2_player.cpp
@@ -546,7 +546,7 @@ void sf2Instrument::playNote( NotePlayHandle * _n, sampleFrame * )
 	{
 		return;
 	}
-	
+
 	const f_cnt_t tfp = _n->totalFramesPlayed();
 
 	if( tfp == 0 )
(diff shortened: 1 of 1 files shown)


Umcaruje:
I think this fix should be targeted at `stable-1.2`.

tresf:
> just to explain, the problem occured when you had an already loaded soundfont, and the file dialog wouldn't open at its location, rather opened at the last browsed folder.

I think we have a few areas where this bug occurs (e.g. saving a project, opening a project).  I find myself desiring this same behavior in other dialogs.

>  I can confirm the fix

Great we can merge.

> I think this fix should be targeted at stable-1.2.

Agreed.

tresf:
Well, changing the branch didn't do well.  Perhaps we can cherry-picked back onto stable.

image

Umcaruje:
@tresf I think we should change the default branch to `stable-1.2` on github. Also I think all future small fixes like this can be targeted to stable-1.2 with no problems @karmux

tresf:
> @tresf I think we should change the default branch to stable-1.2 on github. Also I think all future small fixes like this can be targeted to stable-1.2 with no problems @karmux

We did that for `stable-1.0` as well since so many bugfixes were hitting it in a short period of time.  I'll do that now however we should switch it back shortly after release.

tresf:
> I'll do that now

Done.

tresf:
Cherry-picked to `stable-1.2` via 09c9ccb

PR #3503, opened 2017-04-13
Fixed #3182. Always using master channel for preset previews.

### LINKED ISSUE #3182: Shouldn't saving preset save FX channel as 0?

karmux:
Fixed #3182.

### DIFF
--- a/include/InstrumentTrack.h
+++ b/include/InstrumentTrack.h
@@ -166,7 +166,7 @@ public:
 	{
 		return &m_baseNoteModel;
 	}
-	
+
 	int baseNote() const;
 
 	Piano *pianoModel()
@@ -207,6 +207,8 @@ public:
 		return &m_effectChannelModel;
 	}
 
+	void setPreviewMode( const bool );
+
 
 signals:
 	void instrumentChanged();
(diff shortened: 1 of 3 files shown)


Umcaruje:
Tested this out, works like a charm!
How I tested: routed an instrument to channel 5, added a clipper there to crazily distort the sound, so I know it's getting routed, and made a preset.

When previewing the preset goes to master, when adding it back to song editor, goes to FX 5 :+1: 

The code looks good to me too, only thing I didn't notice that `m_previewMode` is not getting initalised to `false` and I don't know if that can make some problems, but I haven't encountered any while testing so it might not be needed.

Umcaruje:
I think this fix should be targeted at `stable-1.2`.

tresf:
Merged  and cherry-picked to `stable-1.2` via 2df5682.

PR #3509, opened 2017-04-22
SDL version changed from SDL1.2 to SDL2

speedy-software-ch:
You still need to change Build Settings to build against SDL2 *.so / *.dll file.

### DIFF
--- a/include/AudioSdl.h
+++ b/include/AudioSdl.h
@@ -29,8 +29,8 @@
 
 #ifdef LMMS_HAVE_SDL
 
-#include <SDL/SDL.h>
-#include <SDL/SDL_audio.h>
+#include <SDL2/SDL.h>
+#include <SDL2/SDL_audio.h>
 
 #include "AudioDevice.h"
 #include "AudioDeviceSetupWidget.h"


Umcaruje:
This failed on Travis on all platforms, mostly cause you didn't edit travis to use sdl2 on linux and we don't have mingw packages for SDL2. 

What are the advantages of using SDL2 in LMMS performance wise? Do we get any other benefits from this switch?

tresf:
We can add this safely, especially if it's a one liner to the source code, but careful consideration need to be made.
 
1. Add a new cmake module [`cmake/modules/FindSDL2.cmake`](https://github.com/tcbrindle/sdl2-cmake-scripts/blob/master/FindSDL2.cmake)
2. Edit [`CMakeLists.txt`](https://github.com/LMMS/lmms/blob/master/CMakeLists.txt#L233) to still set `LMMS_HAVE_SDL` when either SDL1 or SDL2 is found.  This is done for [backwards compatibility purposes](https://github.com/LMMS/lmms/search?utf8=%E2%9C%93&q=LMMS_HAVE_SDL+&type=).
3. Add a new variable `LMMS_HAVE_SDL2` to [`CMakeLists.txt`](https://github.com/LMMS/lmms/blob/master/CMakeLists.txt#L233) when SDL2 if found.
3. Add a new variable [`LMMS_HAVE_SDL2`)(https://github.com/LMMS/lmms/blob/master/src/lmmsconfig.h.in#L20) to the list of variables that we expose to our C++ code, though the header.
4. Finally, add support for it through the C++ code... 
   ```c++
   
   #if defined(LMMS_HAVE_SDL)
   
   #ifdef LMMS_HAVE_SDL2
   #include <SDL2/SDL2.h>
   #else
   #include <SDL2/SDL.h>
   #endif
   
   #include <SDL2/SDL_audio.h>
   ```

If this seems like overkill, please be assured it's not.  We're [not the only project asking this question](http://stackoverflow.com/a/29447053/3196753).  This solution should be scalable to most platforms and have the smallest impact on the codebase.

speedy-software-ch:
i'll try to change all things you mentioned tresf

Umcaruje:
Also I think this change is more suited for master and not 1.2. Milestoning this for 1.3

tresf:
> I think this change is more suited for master and not 1.2.

It's a trivial change and touches no substantial code, so it doesn't really matter.  I still don't know the benefit of SDL2 and/or if it's something the world will want / need for the next major release, but from a code perspective it's benign to our codebase if the OP has time to complete.

tresf:
Superseded by #3732, targeted at `master`.  Closing.

PR #3515, opened 2017-04-28
Changed email formats

ollieperree:
Replaced "/at/" with "@", "/dot/" with "." etc.

### DIFF
--- a/doc/AUTHORS
+++ b/doc/AUTHORS
@@ -1,37 +1,37 @@
 Tobias Doerffel
-  <tobydox/at/users/dot/sourceforge/dot/net>
+  <tobydox@users.sourceforge.net>
   Development Manager
 
 Jonathan Aquilina
-  <eagles051387/at/gmail/dot/com>
+  <eagles051387@gmail.com>
   Project Manager
 
 Paul Giblock
-  <drfaygo/at/gmail/dot/com>
+  <drfaygo@gmail.com>
   Development
 
 Danny McRae
-  <khjklujn/at/yahoo.com>
+  <khjklujn@yahoo.com>
   Development
 
 Javier Serrano Polo
-  <jasp00/at/terra/dot/es>
+  <jasp00@terra.es>
   Development
   
 Lukas Wohlschläger
-  <lukaswhl/at/gmail/dot/com>
+  <lukaswhl@gmail.com>
   Development
 
 Andrew Kelley
-  <superjoe30/at/gmail/dot/com>
+  <superjoe30@gmail.com>
   Development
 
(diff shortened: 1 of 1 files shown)


tresf:
@ollieperree this is done this way intentionally to thwart off email scrapers/bots.  Can you please explain the purpose of this more?

ollieperree:
@tresf I just noticed that the format was inconsistent. Alternatively, the normal symbols could be substituted in when this file is displayed in the about dialog.

tresf:
@ollieperree thanks for the attention to detail.

The `AUTHORS` file used to separate the software authors from the code contributors, but it's actually very inaccurate and outdated.  We need to remove this file and simply keep the `CONTRIBUTORS` file, which is [hanging out in purgatory for a while](https://github.com/LMMS/lmms.io/pull/198#discussion_r101218217).

Umcaruje:
>The AUTHORS file used to separate the software authors from the code contributors, but it's actually very inaccurate and outdated. We need to remove this file and simply keep the CONTRIBUTORS file, which is 

Though some people are not in the contributors file, like the sample authors and some artwork authors. Also it doesn't really differentiate who did what like in the authors file.

tresf:
We'll have to find a way to hack them into contributors.  The maintenance task can so this.

In regards to *what* they contributed, I have a "who gives a ***" attitude because our most active contributors do a lot more than code LMMS let alone a specific feature.

Perhaps a compromise is to add a category each dev can pick that accompanies our mail map.

tresf:
@ollieperree I'm going to request we close this since the file is so terribly outdated it really serves us no purpose to fix these without the sweeping changes mentioned above which rely on a maintenance task from our web server to be fixed first.  Anyone's welcome to tackle this, but I don't really see a benefit.

Although @Umcaruje brings a good point about **what** each person contributes, it's still quite misleading when it puts people like [`@eagles051387`](https://github.com/LMMS/lmms/commits/master?author=eagles051387) in the listing.  (Sorry Jonathan!)

In regards to the email formatting it was done as a [sweeping change by Toby in 2006](https://github.com/ollieperree/lmms/commit/052224c8e2c1213e1a86010886c77176edbea2c6) when spam emails were prevelant (Gmail was released as a beta in 2004).

Furthermore, this inconsistency is all over the codebase as well.  `gmail.com` vs `gmail/dot/com` and cleaning it up seems to have little or no value and is really up to the author.  Some may not want an email exposed at all.

So although the project is perfectly fine removing the 11-year antiquated anti-spam measure, I'd rather see the time spent on fixing the contributors list.

PR #3516, opened 2017-04-28
Implement 24 bit support for WAV export (#3021) and improve the export dialog

michaelgregorius:
Add a new value of "24 Bit Float" to the "Depth" combo box in the
project export dialog.

Add a new enum value to `ProjectRenderer::Depth` and extend the evaluation
of the different enum values in `ProjectRenderer`.

Add the new case of a depth of 24 to `AudioFileWave` and remove some
repetition with regards to `SF_FORMAT_WAV` in the code. It's only set once
now and then the depth is "added" in a switch statement.

### DIFF
--- a/include/OutputSettings.h
+++ b/include/OutputSettings.h
@@ -1,0 +1,84 @@
+/*
+ * OutputSettings.h - Stores the settings for file rendering
+ *
+ * Copyright (c) 2008-2009 Tobias Doerffel <tobydox/at/users.sourceforge.net>
+ *
+ * This file is part of LMMS - https://lmms.io
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this program (see COPYING); if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA 02110-1301 USA.
+ *
+ */
+
+#ifndef OUTPUT_SETTINGS_H
+#define OUTPUT_SETTINGS_H
+
+
+class OutputSettings
+{
+public:
+	enum BitDepth
+	{
+		Depth_16Bit,
+		Depth_24Bit,
+		Depth_32Bit,
+		NumDepths
+	};
+
+	class BitRateSettings
+	{
+	public:
+		BitRateSettings(bitrate_t bitRate, bool isVariableBitRate) :
+			m_bitRate(bitRate),
+			m_isVariableBitRate(isVariableBitRate)
+		{}
+
+		bool isVariableBitRate() const { return m_isVariableBitRate; }
+		void setVariableBitrate(bool variableBitRate = true) { m_isVariableBitRate = variableBitRate; }
+
+		bitrate_t getBitRate() const { return m_bitRate; }
+		void setBitRate(bitrate_t bitRate) { m_bitRate = bitRate; }
+
+	private:
+		bitrate_t m_bitRate;
+		bool m_isVariableBitRate;
+	};
+
+public:
+	OutputSettings( sample_rate_t sampleRate,
+			BitRateSettings const & bitRateSettings,
+			BitDepth bitDepth ) :
+		m_sampleRate(sampleRate),
+		m_bitRateSettings(bitRateSettings),
+		m_bitDepth(bitDepth)
+	{
+	}
+
+	sample_rate_t getSampleRate() const { return m_sampleRate; }
+	void setSampleRate(sample_rate_t sampleRate) { m_sampleRate = sampleRate; }
+
+	BitRateSettings const & getBitRateSettings() const { return m_bitRateSettings; }
+	void setBitRateSettings(BitRateSettings const & bitRateSettings) { m_bitRateSettings = bitRateSettings; }
+
+	BitDepth getBitDepth() const { return m_bitDepth; }
+	void setBitDepth(BitDepth bitDepth) { m_bitDepth = bitDepth; }
+
+private:
+	sample_rate_t m_sampleRate;
+	BitRateSettings m_bitRateSettings;
+	BitDepth m_bitDepth;
+};
+
+#endif
--- a/include/OutputSettings.h
+++ b/include/OutputSettings.h
@@ -2,6 +2,7 @@
  * OutputSettings.h - Stores the settings for file rendering
  *
  * Copyright (c) 2008-2009 Tobias Doerffel <tobydox/at/users.sourceforge.net>
+ * Copyright (c) 2017 Michael Gregorius <michael.gregorius.git/at/arcor[dot]de>
  *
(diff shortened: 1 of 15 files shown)


michaelgregorius:
Commit 3a14d3b43475cdf019aa6518b086c955af911097 adjust the export dialog so that only the relevant widgets are shown for the selected export type. This enables the removal of the info label that stated: "Please note that not all of the parameters above apply for all file formats."

Example Wave export:
3516-wave-export

Example Ogg export:
3516-ogg-export

michaelgregorius:
Commit 0ba4df406c608c512b6cdbd115a97a8e25a5c9e0 adds the option to use variable bit rates for Ogg exports in the export dialog:

3516-variable-bitrate

Please refer to the commit message for an explanation of how it works with the current code. The commit only connects the new widget to the already existing variable bit rate code.

include/OutputSettings.h
```diff
@@ -0,0 +1,84 @@
+/*
+ * OutputSettings.h - Stores the settings for file rendering
+ *
+ * Copyright (c) 2008-2009 Tobias Doerffel <tobydox/at/users.sourceforge.net>
```
Umcaruje: You should put your name here and the correct date

src/gui/dialogs/export_project.ui
```diff
@@ -81,25 +81,57 @@
...
+           <item>
+            <widget class="QComboBox" name="depthCB">
+             <item>
+              <property name="text">
+               <string>16 Bit Integer</string>
+              </property>
+             </item>
+             <item>
+              <property name="text">
+               <string>24 Bit Float</string>
```
Umcaruje: 24 Bit is Integer, not Float.

Umcaruje:
Tested this out, worked like a charm. I also believe the 24 bit option should become the default one, since 24-bit is the de-facto standard in digital uncompressed audio. 
***
While you're poking around this code, Could you look into this?
screenshot from 2017-04-29 19 52 06
screenshot from 2017-04-29 19 51 52
Whenever you want to export over an existing file you get two prompts, but there should be only one.

zonkmachine:
> Whenever you want to export over an existing file you get two prompts, but there should be only one.

My bad, I'm looking into it.

Umcaruje:
>My bad, I'm looking into it.

Sorry for not opening a separate issue earlier, didn't get around to it

include/OutputSettings.h
michaelgregorius: Done with commit 02a36d138758d4d0433fdef8f21a1e5b9a0c1034.

src/gui/dialogs/export_project.ui
michaelgregorius: What do you think about renaming the entries to something like:
* 16 bit PCM
* 24 bit PCM
* 32 bit float

This would be similar to how Audacity names the formats when importing raw data.

michaelgregorius:
@Umcaruje Commit 8ecae3e59c2b244efa714b6daa4527ca3241b773 makes the 24 bit option the default selection in the dialog. Shall we also reorder the options in the combo box for the bit depth? Currently it sorted by ascending bit depth, i.e. 16, 24, 32. However, if 24 bit becomes the default it might make sense to put it at the top.

Spekular:
Not umcaruje, but FWIW:
-Yes to adding the "float" descriptor to 32 bit
-No to having 24 at the top. I think ascending makes more sense, and it's what we use for Interpolation and Oversampling.

src/gui/dialogs/export_project.ui
Umcaruje: I think `Integer` is more understandable to the end user than `PCM`. Also don't they all use PCM?

michaelgregorius: You're right. In the end all three methods use PCM to encode the signal. I only think that the term "Integer" already goes too much into the technical direction of programming. Put differently: a [search for "16 bit integer"](https://www.google.de/#q=16+bit+integer) will not be as informative for the end user as a [search "16 bit PCM"](https://www.google.de/#q=16+bit+pcm). The latter returns much more audio related results.

Umcaruje:
>Currently it sorted by ascending bit depth, i.e. 16, 24, 32. However, if 24 bit becomes the default it might make sense to put it at the top.

No, it should be kept as-is, like @Spekular said. Putting it on top will just add confusion

src/gui/dialogs/export_project.ui
Umcaruje: Integer is a common English word. Also, Both Bitwig and FL Studio use the `16 Bit Integer, 24 Bit Integer, 32 Bit Float` naming. 

Whoever's working with digital audio will need to learn the difference between integer and floats. PCM is not helpful at all since they all use PCM, and as such you can omit that from the labels. Also, if you search for bit depth, the first thing that'll pop out is a very useful wiki article that explains it all.

michaelgregorius: Commit 954659fa8e91fa42837ca213912d0be421faff5b changes the string from "24 Bit Float" to "24 Bit Integer". It seems that both options ("Integer" vs. "PCM") are used by other audio software, e.g. REAPER also uses "24 Bit PCM" and "32 Bit Float". However, in the end the naming is not that important to me and we shall just use one of the two standards. :)

You are also right that users who want to learn about the feature can just search for "bit depth".

michaelgregorius:
Can I merge this one into master? And shall it go into 1.2 as well?

Umcaruje:
I'd say yes to this being in 1.2, idk about the others tho.

zonkmachine:
1.2 ... :+1:

michaelgregorius:
Merged to master with 241d5f7479023d2903a2658307223672c2b623f3. What's the correct way to merge these changes into 1.2? Do I have to cherrypick the commits?

zonkmachine:
@tresf @michaelgregorius  I would like to see this in 1.2 RC3.

michaelgregorius:
@tresf @zonkmachine If everybody is OK with it I'd happily merge this into 1.2.

However, there is still the open question on how to do this best. After committing this to `master` I noticed that `stable-1.2` is merged regularly into `master` and that it might have been better to directly develop and offer this for `stable-1.2`. I guess I could cherry pick the relevant commits and merge them to `stable-1.2`. However, I wonder whether this would cause problems later on when `stable-1.2` is merged into `master` again. Or is git clever enough that it would figure out that these changes are in `master` already?

speedy-software-ch:
good quesion. would like to know the answer too as to a PR of mine also had such a conflict.

tresf:
Normally, stable goes to master, not the other way around.  Conflicts must be manually resolved.  It sounds like hell and it can be if you don't understand the conflicting code well enough to fix it.  This is why we kept stable-1.2 called master for so long.  It wasn't ready.

Resolving conflicts must be done manually and generally through command line.  We'll get to it.

michaelgregorius:
@tresf @zonkmachine I have added pull request #3552 which integrates the changes into `stable-1.2`. If there are problems with the next merge from `stable-1.2` into `master` then it might make sense to revert the following commits from master: b1df8fd, 515b28d, 3a14d3b, 0ba4df4, 02a36d1, 8ecae3e, 954659f.

Let's hope that git is clever enough to see that the changes from 1.2 are already in master.

michaelgregorius:
I did a local test merge from `3021-24-Bit-Export-For-1.2` into a copy/branch of `master` and everything worked fine. Seems that git is clever after all. 😉

Umcaruje:
>Seems that git is clever after all. :wink:

Yeah, and to whomever does this again, you can just directly cherry pick to stable-1.2 if you have push access, no need for a pull request :wink:

michaelgregorius:
I have merged the changes to 1.2 so that they can be tested with the next RC.

Umcaruje:
Amazing work Michael!

PR #3521, opened 2017-05-01
Flanger LFO rate syncronised

### LINKED ISSUE #3335: Plugin Flanger-lmms is 'backward' in tempo-RATE
If TEMPO is increased the output gets slower
Recreate:
Insert the plugin
set
* DELAY 0
* RATE 2.5
* AMT max
* FB 0.5

play/ listen -flanger is medium fast

right-click RATE, choose sync-to 1/32

play/ listen

Now the flanger is slow, not 1/32 as expected

curlymorphic:
The LFO rate was not correctly synchronizing to tempo

This has been rectified, to utilize the TempoSyncKnob as intended, returning a period,
instead of a frequency. The knob now reports the correct values in the GUI.

fixes #3335

### DIFF
--- a/plugins/Flanger/FlangerControls.cpp
+++ b/plugins/Flanger/FlangerControls.cpp
@@ -35,7 +35,7 @@ FlangerControls::FlangerControls( FlangerEffect *effect ) :
 	EffectControls ( effect ),
 	m_effect ( effect ),
 	m_delayTimeModel(0.001, 0.0001, 0.050, 0.0001,  this, tr( "Delay Samples" ) ) ,
-	m_lfoFrequencyModel( 0.25, 0.01, 5, 0.0001, 20000.0 ,this, tr( "Lfo Frequency" ) ),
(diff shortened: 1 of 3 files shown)


zonkmachine:
With the chosen values the tempo sync menu now leaves out the longest time signatures, whole note and eight beats. I think long sweep times is especially useful for flangers. Here is what it looked like before.

flangertempo

curlymorphic:
@zonkmachine 

Good catch, I will revise

curlymorphic:
I have updated the PR, to restore the syncing of 8 beats notes, 
@zonkmachine nice work for spotting that.

zonkmachine:
> I have updated the PR, to restore the syncing of 8 beats notes,

Works. The 5 second maximum period is in my opinion much to small. I want to be able to sweep the flanger over a long timer, a minute or so. Like a real hippie.

curlymorphic:
I have adjusted the maximum period to be 60 seconds, This does sound nice now,

PR #3522, opened 2017-05-01
Correct the Delay syncronisation

### LINKED ISSUE #3276: Delay - Inconsistent time
Since 4f76241e6803a61377fe32ef522378d9c5b1b7e5 the timing is off. With an older project it plays back correct but when syncing to project tempo the delay is 4 times too fast. Tempo Sync one whole note and you get a quarter note instead.

```

zonkmachine@zonkmachine:~/builds/lmms/lmms$ git bisect bad
4f76241e6803a61377fe32ef522378d9c5b1b7e5 is the first bad commit
commit 4f76241e6803a61377fe32ef522378d9c5b1b7e5
Author: Dave French <dave.french3@googlemail.com>
Date:   Sat Apr 18 18:40:41 2015 +0100

    Delay and Flanger, removed noise from delay when automating length
    
    The original delay code, was setting it read index relative to the length
    of the delay. This fixes that, now when the length is changed the read
    index stays in the correct place in the buffer.

```

curlymorphic:
The Delay plugin had an issue with the delay knob having the incorrect max value,
this was resulting in incorrectly scaled times
fixes #3276

### DIFF
--- a/plugins/Delay/DelayControls.cpp
+++ b/plugins/Delay/DelayControls.cpp
@@ -32,7 +32,7 @@
 DelayControls::DelayControls( DelayEffect* effect ):
 	EffectControls( effect ),
 	m_effect ( effect ),
-	m_delayTimeModel( 0.5, 0.01, 5.0, 0.0001, 20000.0, this, tr( "Delay Samples" )) ,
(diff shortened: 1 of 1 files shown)


zonkmachine:
Tested this and it works fine. Playing and exporting.

PR #3524, opened 2017-05-01
Remove glitching from the Flanger and Delay plugins

curlymorphic:
The previous delay code was incorrectly not utalising the whole buffer, causing glitches when
incressing the delay time, due to outputting incorrect data, This was apparent when using the
lfo in the Delay and Flanger plugins.

This has been rectified. The read index is now offset from the write index. and the complete buffer
is used in a circular fashon.

### DIFF
--- a/plugins/Delay/StereoDelay.cpp
+++ b/plugins/Delay/StereoDelay.cpp
@@ -36,7 +36,7 @@ StereoDelay::StereoDelay( int maxTime, int sampleRate )
 	m_maxLength = maxTime * sampleRate;
 	m_length = m_maxLength;
 
-	m_index = 0;
+	m_writeIndex = 0;
 	m_feedback = 0.0f;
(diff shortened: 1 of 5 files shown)


zonkmachine:
It looks like there still are some glitches in there. In the example there are glitches and occasional beeps, like a buffer replaying the bitinbader sine once in a while.
[flangernoise2.mmp.zip](https://github.com/LMMS/lmms/files/998566/flangernoise2.mmp.zip)

curlymorphic:
@zonkmachine Thank you for the project, That reviled a bug with short delay times, The previous code used the lfo in a way that could result in negative delay lengths, the cause of randomness.

I have revised the code to eliminate this effect.

zonkmachine:
I've tested this PR and https://github.com/LMMS/lmms/pull/3521 together now and I think they should be merged.

PR #3525, opened 2017-05-01
Projectexportfixes

zonkmachine:
If writing over an existing file on export you are prompted twice. See comment here:
https://github.com/LMMS/lmms/pull/3516#issuecomment-298184068

@Umcaruje I had this fix lying around already but hadn't tested it properly. It looks like a fix but it should


zonkmachine:
Borked this PR. You can't reopen a PR after force pushing to a closed one...
New one here: https://github.com/LMMS/lmms/pull/3526

PR #2230, opened 2015-08-02
Set default suffix in "Save as..." file save dialog

zonkmachine:
When you "save as..." without adding suffix manually, LMMS will add a suffix but not test if the file "name + suffix" exists and will gladly write over an existing file.
This is true on Ubuntu Mate at least.
This patch adds suffix mmp/mmpz as default to VersionedSaveDialog when called in saveProjectAs() which fixes the issue.
When I searched the issue I got the impression that this could be mostly a Linux thing but I have not possibility to test on any other OS.

PR #3526, opened 2017-05-01
Fix export - double dialog windows on writing over existing file

zonkmachine:
If writing over an existing file on export you are prompted twice. See comment here:
https://github.com/LMMS/lmms/pull/3516#issuecomment-298184068
Issue introduced here: https://github.com/LMMS/lmms/pull/2230

@Umcaruje I had this fix lying around already but hadn't tested it properly. It looks like a fix but needs more testing.

### DIFF
--- a/src/core/Song.cpp
+++ b/src/core/Song.cpp
@@ -1398,13 +1398,14 @@ void Song::exportProject( bool multiExport )
 		efd.setWindowTitle( tr( "Select file for project-export..." ) );
 	}
 
+	QString suffix = "wav";
+	efd.setDefaultSuffix( suffix );
 	efd.setAcceptMode( FileDialog::AcceptSave );
-
 
 	if( efd.exec() == QDialog::Accepted && !efd.selectedFiles().isEmpty() &&
 					 !efd.selectedFiles()[0].isEmpty() )
 	{
-		QString suffix = "";
+
 		QString exportFileName = efd.selectedFiles()[0];
 		if ( !multiExport )
 		{
@@ -1416,17 +1417,16 @@ void Song::exportProject( bool multiExport )
 				// Get first extension from selected dropdown.
 				// i.e. ".wav" from "WAV-File (*.wav), Dummy-File (*.dum)"
 				suffix = efd.selectedNameFilter().mid( stx + 2, etx - stx - 2 ).split( " " )[0].trimmed();
+				exportFileName.remove( "." + suffix );
 				if ( efd.selectedFiles()[0].endsWith( suffix ) )
 				{
-					suffix = "";
+					if( VersionedSaveDialog::fileExistsQuery( exportFileName + suffix,
+							tr( "Save project" ) ) )
+					{
+						exportFileName += suffix;
+					}
 				}
 			}
-		}
-
-		if( VersionedSaveDialog::fileExistsQuery( exportFileName + suffix,
-				tr( "Save project" ) ) )
-		{
-			exportFileName += suffix;
 		}
 
 		ExportProjectDialog epd( exportFileName, gui->mainWindow(), multiExport );
--- a/src/core/Song.cpp
+++ b/src/core/Song.cpp
@@ -1417,7 +1417,7 @@ void Song::exportProject( bool multiExport )
 				// Get first extension from selected dropdown.
 				// i.e. ".wav" from "WAV-File (*.wav), Dummy-File (*.dum)"
 				suffix = efd.selectedNameFilter().mid( stx + 2, etx - stx - 2 ).split( " " )[0].trimmed();
-				exportFileName.remove( "." + suffix );
+				exportFileName.remove( "." + suffix, Qt::CaseInsensitive );
 				if ( efd.selectedFiles()[0].endsWith( suffix ) )
 				{
 					if( VersionedSaveDialog::fileExistsQuery( exportFileName + suffix,


UnityParadox:
Export showed 2 overwrite dialogs before checkout. After checkout, 1 overwrite dialog is shown. Pull request works as expected.

src/core/Song.cpp
```diff
@@ -1416,19 +1417,18 @@ void Song::exportProject( bool multiExport )
 				// Get first extension from selected dropdown.
 				// i.e. ".wav" from "WAV-File (*.wav), Dummy-File (*.dum)"
 				suffix = efd.selectedNameFilter().mid( stx + 2, etx - stx - 2 ).split( " " )[0].trimmed();
+				exportFileName.remove( "." + suffix );
```
tresf: Two observations about `remove(...)`...

1.  By defaul it's case insensitive by default which can cause issues on Mac and Windows, since they're both case-insensitive (yes, Mac is too!)
2.  This will remove all instances of e.g. `.wav`, so there are some edge cases where `my.wav.file.wav` might get snagged, no?

zonkmachine:
> By defaul it's case insensitive by default which can cause issues on Mac and Windows, since they're both case-insensitive (yes, Mac is too!)

Fixed... I think.

> This will remove all instances of e.g. .wav, so there are some edge cases where my.wav.file.wav might get snagged, no?

Yes. Will think. Think not easy. Sad.

Umcaruje:
@zonkmachine can we merge or do you have something to add to this PR?

qnebra:
Tested locally, fixed double export dialog issue for me.

zonkmachine:
> @zonkmachine can we merge or do you have something to add to this PR?

I'm not near my build machine so I can't work on this right now but I'm happy with a merge. @tresf had made some points above that I also can't test or do anything about right now.

Umcaruje:
Ok no objections to a merge for 3 days, merging this. If there are any issues, they can be opened on the tracker.

zonkmachine:
> If there are any issues, they can be opened on the tracker.

Cool! I'll test this later on and there are still questions from @tresf unanswered. Will get to that.

tresf:
There are still some logical problems.  Linux can have `.foo` and `.FOO` side-by-side (valid albeit confusing) whereas Windows and MacOS cannot, so that logic is a bit presumptuous and these edge-cases will be very confusing in the very rare event they occur.

There's also the double-extension problem, where `.remove()` makes no guarantee that the extension is at the end of the file.   Again, very rare edge-case but when it happen to someone it won't be obvious.  For example `ocean.waves.wav` would become `oceanes.wav`. :)

zonkmachine:
@tresf I'm looking at fixing the two issues you mention above. I haven't found a way to force any actual bug to appear though.

The case sensitivity issue should be fixed by:

```diff
diff --git a/src/core/Song.cpp b/src/core/Song.cpp
index ba2659a..6e9b382 100644
--- a/src/core/Song.cpp
+++ b/src/core/Song.cpp
@@ -1414,7 +1414,11 @@ void Song::exportProject( bool multiExport )
                                // Get first extension from selected dropdown.
                                // i.e. ".wav" from "WAV-File (*.wav), Dummy-File (*.dum)"
                                suffix = efd.selectedNameFilter().mid( stx + 2, etx - stx - 2 ).split( " " )[0].trimmed();
-                               exportFileName.remove( "." + suffix, Qt::CaseInsensitive );
+                               Qt::CaseSensitivity cs = Qt::CaseSensitive;
+#if defined(LMMS_BUILD_APPLE) || defined(LMMS_BUILD_WIN32) || defined(LMMS_BUILD_WIN64)
+                               cs = Qt::CaseInsensitive;
+#endif
+                               exportFileName.remove( "." + suffix, cs );
                                if ( efd.selectedFiles()[0].endsWith( suffix ) )
                                {
```

However. I intend to fix the problem of using `remove()` by using `QFileInfo().completeBaseName()` to 'remove' the suffix and I suspect that also takes care of the other issue. `QFileInfo()` doesn't mention case sensitivity anywhere so I guess it's all taken into account internally.

tresf:
FYI, `LMMS_BUILD_WIN32` is always true for `LMMS_BUILD_WIN64`, only 32 is needed.  If `QFileInfo` works as described (even when the target file doesn't yet exist) it sounds like a plan.

zonkmachine:
> If `QFileInfo` works as described (**even when the target file doesn't yet exist**)

Actually no. It operates on existing files. I'll come up with something else...

PhysSong:
I think we can address the `remove()` issue by using [`chop()`](http://doc.qt.io/qt-5/qstring.html#chop) with [`endsWith()`](http://doc.qt.io/qt-5/qstring.html#endsWith).
However, I've found another issue: the `remove()` tries to `..wav` or similar because `suffix` starts with `.`.

zonkmachine:
Thanks for chipping in! I'll look into chop() with endsWith().
I tried `exportFileName.truncate( exportFileName.lastIndexOf( suffix, -1, cs ) );` in different ways but bumped into what I think is the same issue you mention above with remove(). I'm not treating dots with suffixes consistently so I'll have to look closer into this.

zonkmachine:
Also just bumped into [QTBUG-59401](https://bugreports.qt.io/browse/QTBUG-59401), an issue with dots in the path and `setDefaultSuffix()`.

zonkmachine:
Here is the original Qt issue I tried to solve in https://github.com/LMMS/lmms/pull/2230 with probing the suffix. https://bugreports.qt.io/browse/QTBUG-11352
I think I referred to it earlier but that info got lost when I rebased the PR.

PR #3530, opened 2017-05-02
Equalizer plugin, refinement to analysis display

### LINKED ISSUE #2703: Equalizer shows high frequency content for sine wave at 110 Hz
Playing a sine wave at 110 Hz through an Equalizer results in some signal shown in the high frequency ranges, e.g. 5 kHz. As far as I know the sine in LMMS is computed using the `sin` function from the math library so I'd expect no signal to show in the higher frequencies because these frequency contents should already be well beyond -100 dB FS.

Here is a screenshot that shows the effect:
equalizer-signal-for-sine

curlymorphic:
The spectral analysis was using a rectangle window, leading to high spectral leakage.
This pull request uses the Blackman-Harris window to give a display more representative of the audio.

Fixes #2703

### DIFF
--- a/plugins/Eq/EqSpectrumView.cpp
+++ b/plugins/Eq/EqSpectrumView.cpp
@@ -1,9 +1,8 @@
 /* eqspectrumview.cpp - implementation of EqSpectrumView class.
 *
-* Copyright (c) 2014 David French <dave/dot/french3/at/googlemail/dot/com>
+* Copyright (c) 2014-2017, David French <dave/dot/french3/at/googlemail/dot/com>
 *
 * This file is part of LMMS - https://lmms.io
-*
 * This program is free software; you can redistribute it and/or
 * modify it under the terms of the GNU General Public
 * License as published by the Free Software Foundation; either
@@ -38,6 +37,20 @@ EqAnalyser::EqAnalyser() :
 	m_inProgress=false;
 	m_specBuf = ( fftwf_complex * ) fftwf_malloc( ( FFT_BUFFER_SIZE + 1 ) * sizeof( fftwf_complex ) );
 	m_fftPlan = fftwf_plan_dft_r2c_1d( FFT_BUFFER_SIZE*2, m_buffer, m_specBuf, FFTW_MEASURE );
+
+	//initialize Blackman-Harris window, constants taken from
+	//https://en.wikipedia.org/wiki/Window_function#A_list_of_window_functions
+	const float a0 = 0.35875;
+	const float a1 = 0.48829;
+	const float a2 = 0.14128;
+	const float a3 = 0.01168;
+
+	for(int i = 0; i < FFT_BUFFER_SIZE; i++)
+	{
+		m_fftWindow[i] = ( a0 - a1 * cosf( 2 * F_PI * i / (float)FFT_BUFFER_SIZE - 1 )
+									  + a2 * cosf( 4 * F_PI * i / (float)FFT_BUFFER_SIZE-1)
+									  - a3 * cos( 6 * F_PI * i / (float)FFT_BUFFER_SIZE - 1.0 ));
+	}
 	clear();
 }
 
@@ -83,6 +96,12 @@ void EqAnalyser::analyze( sampleFrame *buf, const fpp_t frames )
 		m_sampleRate = Engine::mixer()->processingSampleRate();
 		const int LOWEST_FREQ = 0;
 		const int HIGHEST_FREQ = m_sampleRate / 2;
+
+		//apply FFT window
+		for( int i = 0; i < FFT_BUFFER_SIZE; i++ )
+		{
+			m_buffer[i] = m_buffer[i] * m_fftWindow[i];
+		}
 
 		fftwf_execute( m_fftPlan );
 		absspec( m_specBuf, m_absSpecBuf, FFT_BUFFER_SIZE+1 );
--- a/plugins/Eq/EqSpectrumView.h
+++ b/plugins/Eq/EqSpectrumView.h
@@ -32,7 +32,6 @@
 
 
 const int MAX_BANDS = 2048;
-
 class EqAnalyser
 {
 public:
@@ -61,6 +60,7 @@ private:
 	int m_sampleRate;
 	bool m_active;
 	bool m_inProgress;
+	float m_fftWindow[FFT_BUFFER_SIZE];
 };
 
 


plugins/Eq/EqSpectrumView.cpp
```diff
@@ -38,6 +38,21 @@ EqAnalyser::EqAnalyser() :
 	m_inProgress=false;
 	m_specBuf = ( fftwf_complex * ) fftwf_malloc( ( FFT_BUFFER_SIZE + 1 ) * sizeof( fftwf_complex ) );
 	m_fftPlan = fftwf_plan_dft_r2c_1d( FFT_BUFFER_SIZE*2, m_buffer, m_specBuf, FFTW_MEASURE );
+
+	//initilise BalckmanHarris window, constants taken from
+	//https://en.wikipedia.org/wiki/Window_function#A_list_of_window_functions
```
tresf: spelling... 

```diff
-	//initilise BalckmanHarris window, constants taken from
+	//initialize BlackmanHarris window, constants taken from
```

Umcaruje: Also there's a dash there, `Blackman-Harris`

tresf:
@michaelgregorius do you have time to test and approve this pull request?

michaelgregorius:
@tresf, @curlymorphic: I had a look at the changed code and also compiled the branch myself to test it with a simple sine wave at 110Hz again.

First something that I noticed while looking at the code. It might be worthwhile to put the code that generates the window into a separate class in case we decide later to also implement other windowing functions.

As for the practical test it seems that unfortunately the patch doesn't have a big effect in the tested case.

Here's the EQ from master:
3530-sine-from-master

And here's the same signal with the patch applied:
3530-sine-from-branch

For reference here's the same signal shown in the frequency analyzer from REAPER using a buffer size of 2048:
reaper-blackman-harris-2048

Please note that LMMS cuts off the display at around -18 dB FS and that that's the reason why the plots don't look almost identical.

Judging by the plots and looking at the changed code the implementation seems to be correct. However, it seems that the main culprit for the high leakage is the buffer size of 2048 that's used for the FFT. If I set the buffer size to 32768 in the REAPER frequency analyzer the peak for the test signal becomes very sharp and does not spill over.

I think the following features might be interesting for the analyzer:
* The ability to switch the windowing function.
* The ability to switch the buffer sizes.
* The ability to switch the scales from -18 dB to smaller values.

The reason for the last request is that some people mix with signals peaking at around -18 dB to have enough headroom. If you do this in LMMS and send such a channel through the EQ you will likely see almost nothing in the plot window.

Sorry, for requesting all this here. I usually hate pull requests that request more and more features myself. :wink:

If you want we can add separate issues for these requests.

curlymorphic:
@michaelgregorius 

Thanks for the in-depth report.

I will have to check the range on the display when I'm at my pc, but its alot more than the -18dB, its just squashed vertically when compared to reapers output. The +-18dB is the gain 'for the eq controls. It could be an idea to add the analysis scale on the right hand side.
What I am going to do is write the content of the fft buffer to a csv file and plot the curves, as I want to chech the display when over a larger y axis.

"First something that I noticed while looking at the code. It might be worthwhile to put the code that generates the window into a separate class in case we decide later to also implement other windowing functions"

I would agree. The fft-helper class, but that would cause confusion, as there is already a function there, that's suposed to do the windowing, that iirc is used in the spectrum analyzer plugin. However there are some problems with it. It calculates the window incorrectly, it also calculates the window for each sample played, 2 trig functions and 2 divides per sample is just silly. 
I don't want to change the code for another plugin this close to a release.

"I think the following features might be interesting for the analyzer:

The ability to switch the windowing function.
The ability to switch the buffer sizes.
The ability to switch the scales from -18 dB to smaller values."

These are all good ideas, I was tempted to add these to this pull request but i decided againt adding new features at this stage. I do think it would be a good idea to add these suggested features, but in a seperate pr aimed at master. I can do a mock up of the additional features, in the not to distant future, but i would prefer to fix current bugs first.

"If you want we can add separate issues for these requests." Yes please.

Umcaruje:
I think you should try out the von Hann window, it's the default in calf analyzer and works pretty good. When using the blackman-harris windowing in the calf analyzer it added a lot of weird harmonics to the sine waves, while the von Hann window really seemed like the best choice.

Umcaruje:
>If I set the buffer size to 32768

Wouldn't that make the latency on the EQ go through the roof?

curlymorphic:
> Wouldn't that make the latency on the EQ go through the roof?
 
Its the trade off between frequency and time resolution.


 
> I think you should try out the von Hann window, 

Thats a great idea, maybe for the update aimed after 1.2 release, we should collate a list of what windowing functions users would like, and any additional functionally required  then plan and evaluate the expansion. I am keen to do this, could be a nice team project.

michaelgregorius:
> Wouldn't that make the latency on the EQ go through the roof?

Yes, bigger windows definitively introduce some latency. However, I assume that this would only affect the graphical display and not the processing of the signal itself.

> I am keen to do this, could be a nice team project.

I agree and I think that LMMS needs more of such team efforts/projects. Right now LMMS seems to be mainly developed by individual developers who contribute with pull requests. Obviously there is a limit to what a single developer can do and this keeps the project from implementing the bigger changes that are needed/wanted like a better separation between the core and GUI classes, realtime safe operations or the single window GUI.

curlymorphic:
@Umcaruje 
I have changed the fft window to use a Von Hann function, to see what you prefer?

I will remove the unused windowing function code, when a decision has been made as to what is preferred.

PhysSong:
Maybe it would be better to separate window initializing code to some new functions and implement switching between window functions.
If it is not needed for now, just providing windowing toggle button would be better.

michaelgregorius:
@curlymorphic I have added #3624, #3625 and #3626 to deal with the aforementioned features separately.

@PhysSong I agree that it would be a good idea to have a class or a collection of functions that can generate different types of windowing functions that can be reused in different places.

zonkmachine:
> @Umcaruje
> I have changed the fft window to use a Von Hann function, to see what you prefer?
> 
> I will remove the unused windowing function code, when a decision has been made as to what is preferred.

I have no preferred windowing function myself. I've tested this and it works fine. I think we can merge it as it is with a von Hann window and the commented out Blackman-Harris function and then get back to it after 1.2 is released.

PhysSong:
Some windowing functions(Hamming and von Hann) are already implemented in fft_helpers.h, and I think it is possible to reuse them. If someone needs other window functions, simply adds it into fft_helpers.h and use it in plugins.

zonkmachine:
> Some windowing functions(Hamming and von Hann) are already implemented in fft_helpers.h, and I think it is possible to reuse them. If someone needs other window functions, simply adds it into fft_helpers.h and use it in plugins.

I think that can be fixed later here: https://github.com/LMMS/lmms/issues/3625
We're pressed to release 1.2.0 and this is a pretty good place to merge this in my opinion.

@LMMS/developers Merge?

PhysSong:
Someone might want to see not windowed spectrum as before. I think windowing on/off button should exist.

Umcaruje:
@zonkmachine I really want to hear from @michaelgregorius on which windowing function is better, so we can clean up the code before merging.

michaelgregorius:
@Umcaruje I don't think there is something like a "better" function. As can be seen here they all have certain tradeoffs: https://en.wikipedia.org/wiki/Window_function#A_list_of_window_functions

It seems that the main lobe of the FFT is more pronounced with Blackman-Harris. So I'd use Blackman-Harris for now and then implement the switching of windowing functions as part of #3625. One option for the switching should be something like "Off (Rectangular)" so that it can also be turned off.

curlymorphic:
@Umcaruje @michaelgregorius 

If we can make a decision  I will amend the pull request to suit. My vote would be Blackman-Harris, 

I would like to stick with the decision of not changing the fft_helpers class this close to a release. I do plan on implementing the windowing functions in the fft_helper class,  but I shall aim that pull request at master due to changes affecting more than just this plugin. Changing the Equalizer plugin to use this code can then be handled in #3625.

michaelgregorius:
@curlymorphic Sound reasonable and good to me!

curlymorphic:
@PhysSong is already working on updating the fft_helper class for the master branch  in #3625 
Sorry, I only just noticed :(

Umcaruje:
Yeah, Blackman-Harris gets a :+1: from me as well then, this should be merged

Issue #2510, opened 2016-01-17
Scaling issues on high DPI displays


samkcarlile:
This is my first time submitting an issue here, so If I'm doing it incorrectly, please let me know.

I'm on Windows 10 and would like to use LMMS, however, being on a Surface Book with a high DPI display, I can't comfortably use this software without squinting to see the various menu items and text. This application doesn't seem to scale when I scale Windows up. Is this a known issue or is it just me?

tresf:
Closing, marking as duplicate of #455, #2209, #1913, #82.

musikBear:
@tresf Looks like all topics around this issue specific *resolution on high-end screens/displays* has been closed, or consolidated with font issues.
Imo at least one with *high-resolution monitors* should be kept open. 
Some users even now has 4K monitors:
http://lmms.io/forum/viewtopic.php?f=7&t=4843&p=15906#p15906
But are *-every-* 'normal' application not suffering from exact that same issue?
How does anything, even *Word*, manage a 4K monitor?
Surely *this* cant just be a LMMS issue. :confounded: 
I can only think of advising users with 4K monitors, to select a different *general* resolution. Can anyone suggest something better?
Imo. Schedule 1.2.x

tresf:
> Imo. Schedule 1.2.x

Stop.

> I can only think of advising users with 4K monitors, to select a different general resolution. Can anyone suggest something better?

No.

> How does anything, even Word, manage a 4K monitor?

Perfect.

> But are -every- 'normal' application not suffering from exact that same issue?

No, they don't.  Install QT Linguist on Windows.  It does not suffer this scaling problem.

> Some users even now has 4K monitors:

The exact resolution is immaterial, the fact that it doesn't scale is the problem.

> @tresf Looks like all topics around this issue specific resolution on high-end screens/displays has been closed, or consolidated with font issues.

Yes, we've been making progress on some fronts by removing hard-coded values, but they still may be wrong as we aren't testing each CSS patch with the super-HD folks in mind.

[...]

PR #3531, opened 2017-05-04
Disable HiDPI on Windows

tresf:
This will address #2510 on Windows by disabling High DPI in LMMS until proper Hi-DPI support is added to the codebase.  Thanks @BaraMGB for pointing me to the [forums](https://lmms.io/forum/viewtopic.php?f=7&t=24335&p=44747#p44747) where this was being evaluated.

Installer available for testing here: [`lmms-rc2.hidpi-win64.exe`](https://github.com/tresf/lmms/releases/download/v1.2.0-rc2.113/lmms-1.2.0-rc2.hidpi-win64.exe)

Before:
image

After:
image


BaraMGB:
This works like a charme. It has to be clear that the whole UI is a little bit blurred. But for people with hight resolution screens it is a ability to work with lmms. I don't know how blurred it is on a 4K screen. I only work with  FullHD screen here but I could work with it.

tay10r:
@tresf Thank you so much for bringing this up! It was not at all usable on 4K until this!

PR #3532, opened 2017-05-05
Detachable windows

### LINKED ISSUE #1259: Multiple Monitors
I am not sure if this idea or issue has been discussed elsewhere,  but I was thinking it would be great (I don't know how tough it would be to implement this) if lmms subwindows(things such as the "Song-Editor" and "Mixer") could be separate rather than confined to the single window provided, such as in Gimp. I feel like this would be really useful in properly using more monitors. 
untitled
Of course, the windows will have the proper window style (Based on operating system). 

Or it can be like Blender's system where you can pop out windows by pressing shift and dragging. And ideas on this?

I'm new to this and I'm not sure if I posted this on the right spot. If not, please tell me where these things go. 

Edit: I would love to contribute, but I don't think my coding is up to par yet. Hopefully, when I get better.

lukas-w:
Allows detaching a window from LMMS's main window, making things like working on multiple screens easier.

Closes #1259

Quick demonstration:
lmms-window-detach


BaraMGB:
Wow! That's really awesome!

karmux:
Very useful functionality especially when using multi-monitor setup! :+1:

I played a little bit with it and found some bugs:
1. When detached instruments are maximized then all instruments stay maximized inside LMMS until LMMS runs. I think there shouldn't be maximize button and functionality for instrument windows at all.
2. Instruments, mixer, project notes and controller rack can't be closed once detached. Clicking on X-button does nothing.
3. Trying to close instrument when detached crashes LMMS every time.
4. Detaching any window and then clicking minimize freezes this window inside LMMS main window and the subwindow border icons/labels don't look correct. Expected functionality is to minimize detached window same way like main window.
5. Detaching any window and then clicking maximize maximizes it inside LMMS main window and scrolling main window too much to the right. Expected functionality is to maximize detached window same way like main window.

Lukas-W:
@karmux Thanks for testing! I'll have a look at the issues you pointed out as soon as I find the time. I already noticed number 2 (X-button doesn't work), but couldn't find the cause.

Jousboxx:
This is super useful!! 

I can't seem to make a windows build of it though. The build seems to go without a hitch, but when I open the app, it doesn't have the detach button. This might be my own error, but maybe someone else should look into it.

I can confirm all the issues that @karmux pointed out. Also, getting the detached window back into lmms is a little unwieldy. I basically need to re-open the window (for exmaple, clicking the FX-Mixer button to re-attach the FX-Mixer). Not exactly a bug, but it would be useful if the minimize button reattached the window, or if there was another button entirely.

Good job, looking forward to be able to use this for major producing :)

NickAcPT:
Sorry for the bump.
@Jousboxx I agree with you when you gave the idea that the minimize button would reattach the window into LMMS.
But, if they were going to add a button on the window to attach it back, they would need to adjust the window's content to account for the button.
Also, if you were talking about adding a button to the title bar, that would be a bit complicated and hard to maintain.
###### PS: Sorry to bother...

Jousboxx:
@NickAcPT Yeah, my idea is that since the minimize button reattaches, they wouldn't have to add another button. The minimize button on the main LMMS window should probably shrink all child windows so there is a way to minimize them.

PhysSong:
> This uses some Qt5-only features, so it's best to wait with merging this until we've switched

Now we've, so I think we can continue working on this.
> When detached instruments are maximized then all instruments stay maximized inside LMMS until LMMS runs. I think there shouldn't be maximize button and functionality for instrument windows at all.

It's already fixed.
> Mixer, project notes and controller rack can't be closed once detached. Clicking on X-button does nothing.

> Trying to close instrument when detached crashes LMMS every time.

`vstSubWin`, `ControllerDialog`, `EffectControlDialog`, `FxMixerView`, `SongEditor`(this is fine because  `SongEditorWindow` doesn't), `ControllerRackView`, `ProjectNotes` and `IntrumentTrackWindow` class overrides `closeEvent` and ignores close event. See https://github.com/LMMS/lmms/pull/1495#issuecomment-68191344 for why we have such workarounds.

> Detaching any window and then clicking minimize freezes this window inside LMMS main window and the subwindow border icons/labels don't look correct. Expected functionality is to minimize detached window same way like main window.

> Detaching any window and then clicking maximize maximizes it inside LMMS main window and scrolling main window too much to the right. Expected functionality is to maximize detached window same way like main window.

I guess the event is propagated to `SubWindow` in some way. Installing an event filter which blocks propagation seems to fix it.

PhysSong:
Found some more issues:
- Window size is not preserved when using some window manager on X11(compiz on Ubuntu 18.04, unity session in my case). There's a [known issue](http://doc.qt.io/qt-5/application-windows.html#x11-peculiarities) with `frameGeometry` on X11 so I'm not sure if we can work around this.
- When switching instrument, detached windows get attached automatically.

PhysSong:
> if the minimize button reattached the window, or if there was another button entirely.

If the minimize button is used for attaching, wouldn't it be confusing? Also, I can't find platform-independent way to add such a button to the title bar unless we create a custom window class which would be quite difficult. There's an easier workaround, wrapping window content and window controls in a layout. It might be a little bit ugly, but it's easy.

@lukas-w may I continue this work?

lukas-w:
> @lukas-w may I continue this work?

Go for it. :+1:

PhysSong:
Fixed almost all bugs reported by @karmux. As a side effects, minimize button doesn't re-attach windows anymore. I can restore the behavior, but I think that might be confusing.
Bugs in https://github.com/LMMS/lmms/pull/3532#issuecomment-391538209 still needs to be fixed.

karmux:
@PhysSong thank you! During a quick test I couldn't reproduce any of my previously reported bugs.
While testing I discovered now that I can't close sample track instrument in detached mode. Maybe happens because I have #3866 merged.

PhysSong:
> Maybe happens because I have #3866 merged.

I think that's correct. IIRC I copied some code from instrument track and it had some issue(fixed in this PR).
When/If that one is merged, I will add a fix.

lukas-w:
> As a side effects, minimize button doesn't re-attach windows anymore.

It never did. This was just suggested by @Jousboxx in https://github.com/LMMS/lmms/pull/3532#issuecomment-299733177.

pwepwe973:
hello,
how to enable this option please?
thank you

PhysSong:
@pwepwe973 This feature is in development and not a part of released versions. Sorry.

winniehell:
@PhysSong Can you please summarize the current state of this? Is it ready to be merged?

I'm willing to help get this finished if possible.

PhysSong:
> @PhysSong Can you please summarize the current state of this?

I unintentionally abandoned this one, but I can restart working on this.
> Is it ready to be merged?

Not yet, mainly due to https://github.com/LMMS/lmms/pull/3532#issuecomment-391538209.

pwepwe973:
you are the best!!!!! thank you for this option

PhysSong:
> When switching instrument, detached windows get attached automatically.

This should probably tested again once #5808 is merged.

PhysSong:
> > When switching instrument, detached windows get attached automatically.
> 
> This should probably be tested again once #5808 is merged.

I checked it again and noticed that the bug still exists with slightly different behavior. It looks like `InstrumentTrackWindow::toggleVisibility()` and `InstrumentTrackWindow::viewInstrumentInDirection()` require some modifications.

Rossmaxx:
What's the status as of now?

pwepwe973:
I have never changed version since I kept this version with the detachment of the windows because I use the others on 2 screens, even if there are new things but I am not interested in WHAT INTERESTS ME IS TO BE ABLE TO WORK in a practical way with the 2 screens

luzpaz:
What's blocking the rebase  ?

Rossmaxx:
merge conflicts maybe, and the fact that the author hasn't been here in a long time.

messmerd:
@PhysSong Are you still working on this? If not, I can try getting it to a state where it's ready to merge.

PhysSong:
> @PhysSong Are you still working on this? If not, I can try getting it to a state where it's ready to merge.

@messmerd No, please feel free to take over.

messmerd:
I fixed the bug that prevented detached editor windows (Song Editor, Piano Roll, etc.) from closing when "X" is clicked.

These are the known bugs that still remain:
- Detached microtuner config window doesn't close when "X" is clicked
- When switching instruments, detached windows get attached automatically
- The LMMS taskbar icon changes to the icon for the most recently detached window
- Some detached windows look really bad when maximized, so they should be prevented from maximizing if possible
- The "Add" button in the Controller Rack is not visible at first when attached

SpomJ:
One more thing, on Hyprland (Wayland), spawned windows that are closed can never reopen. I'll try look into this when I have time.

SpomJ:
Upon further inspection, it seems that the issue described is precisely this - The song editor buttons and the detached window status are out of sync.

If you switch to a closed instrument with arrow buttons from a detached window, its state toggles in song editor, but the window does not close.

The inverse is also true - if you manually toggle the window from the song editor it also doesn't toggle, however this time after you close the window, reopening does nothing. The only way to get it back is through the same arrow buttons.

SpomJ:
We should also check if the returned position for the window pos are 0-0 which appears to indicate whatever compositor is used doesn't support the window tracking. It might be better to ditch the 'seamless snapping' altogethrt and snap the window back to where it was detached from. Or do both and apply the second only if the first fails.

SpomJ:
> The "Add" button in the Controller Rack is not visible at first when attached

This and perhaps some other issues of this type could be solved by closing the window, then reopening it instead of trying to move it, since as far as I can tell everything remains the same if we did that.

SpomJ:
We should also find whatever is responsible for blocking resizing internally. As I theorize it now, there might be some flag which prevents resizing as a possibility in embedded mode, and if we want to be consistent we just need to adapt it for WMs.

I think that because the size persists when resetting back:
image

messmerd:
@SpomJ Thanks for testing this out.

I originally decided to contribute to this PR simply because no one else was working on it. While I did refactor it to cut down on duplicate code and fix a couple of bugs, I honestly don't really understand how Qt's window-related stuff works, so the changes I made that weren't refactoring-related were just me experimenting and seeing what worked.

I'm not sure I'd be able to fix the remaining issues, and I'm tied up with some other LMMS work right now, so if you or anyone else wants to contribute to this PR, please feel free. I can help review and test anything you come up with.

SpomJ:
I think the thing should be rebased before doing anything, which I have no clue how to do (-:
If someone rebased it, they would be very welcome.

Other than that, I think I managed to track the instrument issue down to the 

```cpp
InstrumentTrackWindow * InstrumentTrackView::topLevelInstrumentTrackWindow()
{
	InstrumentTrackWindow * w = nullptr;
	for( const QMdiSubWindow * sw :
				getGUI()->mainWindow()->workspace()->subWindowList(
											QMdiArea::ActivationHistoryOrder ) )
	{
		if( sw->isVisible() && sw->widget()->inherits( "lmms::gui::InstrumentTrackWindow" ) )
		{
			w = qobject_cast<InstrumentTrackWindow *>( sw->widget() );
		}
	}

	return w;
}
```
which appears to not care about the windowing at all, although I might be wrong here.

SpomJ:
[0001-fix-detached-windows-being-shown-after-toggleVisibil.patch.txt](https://github.com/user-attachments/files/17791639/0001-fix-detached-windows-being-shown-after-toggleVisibil.patch.txt)
I'm not sure how do I write to this repo, so I resort to this. This should fix the window jank I mentioned above.

SpomJ:
> The "Add" button in the Controller Rack is not visible at first when attached

On a closer look, what actually happens is the fake window gets the wrong dimensions at first, you can see the button peeking from below as well as the window also being cropped horizontally. No idea what causes it though.

SpomJ:
As if this wasn't enough, I noticed that some (currently knob, but might be others too) tooltips appear inside main window instead of popping over like they should. This well be fun to fix (no)

messmerd:
Shoot, I think I just messed up the merge...
I will open a new PR using a feature branch that isn't directly in the LMMS repo, since this has been causing nothing but problems.

messmerd:
I've opened #7586 which supersedes this PR.

lukas-w:
> I will open a new PR using a feature branch that isn't directly in the LMMS repo

This branch being directly in the LMMS repo may be a major reason it's still receiving contributions after no less than seven years, instead of slowly dying or disappearing in some now-inactive developer's repository that no-one else can push to. If this feature is to actually land in LMMS one day, I believe it's better for it to have its home in this repo, as explicit and visible part of this project & team. I made a deliberate choice to push this (and other branches in the past) directly to the main repo for exactly those reasons - so that they can be seamlessly continued by any LMMS member in the (now very real) event that I become unable to do so.

> [being] directly in the LMMS repo [...] has been causing nothing but problems.

This statement feels ungrounded and unrelated to your merge problems, I'm not sure what problems it would cause but I admit I haven't been following the conversation here. The location of an upstream branch doesn't change anything about a developer's ability to successfully merge two branches on their machine though - the process is the same either way.

lukas-w:
I've reset the branch to a8fddfcd7b60b735eed0fc3288f83b35c658fe53, the commit immediately preceding the messed-up merge (7fbd1f1e18c4a5ed32f6c52b0136ea5d0d3c6a99, this link may lead to a 404 error eventually). Feel free to give merging another shot. Rebasing as a strategy to pull in changes from `master` to a branch that's already been pushed - especially one that's so old - should generally be avoided with very few exceptions as it makes it harder for everyone else to sync changes and also pretty much defeats the point of Git's non-linear model - although I do realize that this opinion is somehow controversial in this project.

lukas-w:
I had a go at merging master, but not 100% sure on some of the merge conflicts as I don't have time to properly review the commits made to this branch. Feel free to just reset the branch again to revert the branch if I broke something and you'd rather do it yourself :)

SpomJ:
That's fun. Now there are _two_ of essentially the same repo. I guess I'll commit to both -_-

SpomJ:
So I'm starting to understand what's going on with resizing issues.

Essentially what happens is that in detached mode window size is only controlled by topmost widget (that being window frame), while the actual size restrictions were always announced for the subwindow.

What we (I) have been doing for a while here, is hunting down the restrictions end essentially setting both. Which is wrong and shouldn't happen, however I don't see another way to do this. If anyone comes up with a more sensible solution, they would be very welcome.

SpomJ:
Until then I will continue to push double sizing.

lukas-w:
> That's fun. Now there are _two_ of essentially the same repo. I guess I'll commit to both -_-

Apologies, my intent wasn't to have two competing pull requests. I'm waiting on feedback whether my `master` merge in this one is fine or not, so that we may close out one of the PRs.

Also I wasn't aware you had write access to @messmerd's fork.

SpomJ:
> These are the known bugs that still remain:
> 
> * Detached microtuner config window doesn't close when "X" is clicked
> 
> * When switching instruments, detached windows get attached automatically
> 
> * The LMMS taskbar icon changes to the icon for the most recently detached window
> 
> * Some detached windows look really bad when maximized, so they should be prevented from maximizing if possible
> 
> * The "Add" button in the Controller Rack is not visible at first when attached

Meanwhile I think I managed to fix some of these.

At this point known ones are:
- When switching instruments, detached windows get attached automatically (by design??)
- The LMMS taskbar icon changes to the icon for the most recently detached window (Don't have a taskbar, can't confirm)
- Some detached windows look really bad when maximized, so they should be prevented from maximizing if possible (partially solved, idk if any are left)
- ~~Controller Rack "Add" button invisible~~ (fixed, was caused by wrong minimal size, a sort-of-regression caused by 4b3bf9f2a5d69a1c3e21c7d27a57e7f0a53477fd)
- ~~Detached microtuner config being unable to close~~ (fixed, missed QWidget into DetachableWidget conversion)

\+ Newly discovered:
- Knob hints appear on main window instead of being pop-up
- Windows get translated to 0-0 when attached (Issue on Wayland side, but a fix would be nice in the meantime)

SpomJ:
> Quick demonstration:
> lmms-window-detach

I don't really like the backwards transition here, it feels quite counter-intuitive and this is what I believe to be the cause of detached windows being attached on instrument switch.

SpomJ:
> I'm waiting on feedback whether my `master` merge in this one is fine or not, so that we may close out one of the PRs.

I have no idea either. I guess we should ask @messmerd ?

SpomJ:
I did some PRs to #7586 in the meantime, idk what do we do with this branch... Do we just pull from #7586?

messmerd:
@SpomJ Since you've been working on this feature (thank you by the way!), I think it's up to you which branch(es) you want to contribute to. I'm just happy to see progress on this, so I'm fine with either option.

messmerd:
@lukas-w Your `master` merge appears similar to what I did as far as I can tell. I was mainly concerned with `SubWindow::adjustTitleBar()`, so I added a couple TODO comments there when I was resolving the conflicts.

SpomJ:
The controller rack behavior also seems really weird, and I don't know what's causing it. It loads fine, but then snaps the subwindow to what appears to be embed size for no apparent reason. My commit is apparently a workaround for this, but that shouldn't be happening in the first place.

lukas-w:
> I don't really like the backwards transition here, it feels quite counter-intuitive

Yeah I'm not very fond of this either, it was a poor choice but at the time felt like the easiest solution without introducing more complex changes. Some ideas that come to mind that may be better:

1. Re-attach on window close
2. Re-attach on keyboard shortcut, which could be displayed briefly as a popup hint when detaching
3. Have the detached window keep the custom window frame to keep access to the detach/attach button. Probably the most intuitive for most users using floating window managers, but more work to implement. May make it necessary to replicate some of `QMdiSubWindow`'s features to e.g.  that can't be overwritten in a subclass. This may have limited support on some Wayland compositors.
4. Be fancy and find a way to allow simply dragging a subwindow out of and into the main window to trigger a detach or attach respectively; similar to how browsers allow moving tabs to a new window by dragging the tab & dropping it outside of the window in an empty area. I like this most from a usability perspective, but dragging into the window again will not work with Wayland as far as I know.

I think 1. is the sanest choice here because it's very simple, would work on any desktop / window manager and while I don't know if I'd call it intuitive, users will probably discover it quickly which can't be said about the current solution at all. The downside being that users who actually want to get rid of the window altogether will have to click twice.


> I did some pushes to #7586 in the meantime, idk what do we do with this branch... Do we just pull from #7586?

We can't just pull because #7586 is rebased on a newer `master`, so merging it back to `feature/detach-window` would duplicate all commits made there. Hence my aversion to unnecessary rebasing :grin: It makes collaboration harder. If I have your OK though, I'll cherry-pick the new commits made there so that work may continue targeting the original branch. :+1:

SpomJ:
> 1. Re-attach on window close

This would actually be really nice if it could be toggled in settings. If only we had actual settings...

> 3\. Have the detached window keep the custom window frame to keep access to the detach/attach button. Probably the most intuitive for most users using floating window managers, but more work to implement. May make it necessary to replicate some of `QMdiSubWindow`'s features to e.g.  that can't be overwritten in a subclass. This may have limited support on some Wayland compositors.

This one is... interesting. Client-side decorations _are_ a thing, and would probably work beautifully if implemented properly. I have two complaints against this, however. First, it would probably take a bit of unnecessary boilerplate, if Qt can make this happen at all. And second, I rely on tiling WMs and consider title bars ugly :]



> 4\. Be fancy and find a way to allow simply dragging a subwindow out of and into the main window

at this point it would be easier to just get rid of `QMdiSubWindow` altogether and just detach all the windows. This messes up the usual 2d-scrolled landscape and makes windowing unnecessarily janky, especially if you have other windows floating alongside LMMS.

---

So out of all these the most pleasing to me (i'm biased btw) is a shortcut, although again adding shortcuts without a way to change them makes me sad. If I understood you correctly, you essentially want to make close button an attach one, which would be nice aside from the fact it would require patches for non-persistent things (like microtuner) where closing them makes more sense (and therefore makes the code just so slightly more obese).

SpomJ:
> If I have your OK though, I'll cherry-pick the new commits made there so that work may continue targeting the original branch.

Do as you please (and on the way you can squash two commits where I screwed up size constraints on itw and almost immediately patched it afterwards :)

I'm happy to contribute to whatever branch I have write access to.

SpomJ:
Regarding [remaining issues](https://github.com/LMMS/lmms/pull/3532#issuecomment-2483806758).

I'll reiterate my thoughts on resizing issues. For whatever reason SubWindow (actually, widget minimalSizeHint) does not respect the widget's size constraints, and only snaps to requested size once moved/resized by user. Calling `resize()` on `SubWindow` does _not_ check for these constraints. For this reasons, some constraints in code are set for `subWin` instead, giving tolerable results; however this does not impact detached windows in any way since only the embed (without the SubWindow) gets detached.

This is what caused "Add" on controller rack not to be visible, resulting in me using a bad workaround, and this is what causes constraint issues specifically on SlicerT which I tried solving for several days at this point and intend to give up soon. I could use the same workaround, but this doesn't get us anywhere in the long run, since at the end of the day it is still very much a workaround.

I have no clue on how to fix this, and apparently no one in #dev-support either cares or knows either.

---

Other than that, basically no issues remain. I can't test the icons thing, so I'm probably the last person to fix it. The attachment thing seems to be a design decision, and can be reverted with relative ease in favor of anything we discussed. Window translation is annoying, but can probably be fixed just by throwing the window to center in `attach()` which shouldn't be that hard.

---

So ignoring the first part, this is pretty much one step away from being mergeable.

lukas-w:
> And second, I rely on tiling WMs and consider title bars ugly

Same. I don't mind a slim bar though if it provides important application features. Firefox draws its own title bar and it never bothered me because it actually has a purpose by integrating the tab bar. If it didn't it wouldn't be possible to attach a single tab window back into another one so that's a win to me.

> at this point it would be easier to just get rid of `QMdiSubWindow` altogether and just detach all the windows

This would be awesome for us two tiling window manager users out there but a GIMP-2.0-trauma-triggering constant-window-misplacing nightmare for everyone else. :ghost:

lukas-w:
@SpomJ I rebased your commits on top of `feature/detach-window` while squashing b477e6a6130c2017b5349e252d9830665c54996b and 8fa469da87a993cc9ad08c06f56b049f4a2cc081.

SpomJ:
I submitted a PR (#7592) to this, sort of fixing the wayland bbug

SpomJ:
I fixed all the bugs I know of and can test.

It's probably best to leave attachment behavior as it is, at least for now. We could probably hook attachment to minimize button instead, if there even is a call for that. I'd mainly go for shortcuts, but there's ongoing work on them, so it's probably best not to touch it for now.

I can't check anything about window icons, since y'know, wayland.

From further testing the best I can do about window constraints is set up a helper function which syncs them since setting either of these is not sufficient and there doesn't seem to be a way to dynamically infer one from another.

Is there anything that prevents this from exiting the draft stage?

messmerd:
@SpomJ The taskbar icon bug isn't occurring for me anymore, at least when I built this branch locally with Qt 5.15.2. Maybe it was a Qt bug that got fixed upstream, or your changes fixed it somehow.

messmerd:
The only bug I've found so far is that detached windows can be resized past the minimum size. Other than that, everything seems to be working fine. I'll continue testing later today.

SpomJ:
What are those windows? Only thing I know of still doing this is SlicerT and it seems to be reworked anyways, so I think tweaking it now would do more bad than good.

messmerd:
SlicerT, Microtuner config, and the LADSPA plugin browser.

And Tap Tempo is resizable when detached but it shouldn't be.

messmerd:
Found some more windows with resizing issues:
VST effect controls window, VST parameter windows (this also is missing its window's close button and the "Close" push button inside the window doesn't work), LADSPA effect windows, VST GUI windows

SpomJ:
I've come up with two solutions for this.

1. ```cpp
void SubWindow::applySizeRestrictions()
{
        if (!widget())
        {
                return;
        }
        setMinimumHeight(widget()->minimumHeight() + titleBarHeight() + frameWidth());
        setMinimumWidth(widget()->minimumWidth() + 2 * frameWidth());
        setMaximumHeight(widget()->maximumHeight() + titleBarHeight() + frameWidth());
        setMaximumWidth(widget()->maximumWidth() + 2 * frameWidth());
}
```
this is a function that is supposed to be called after widget dimensions are set. I already implemented it and can confirm it works perfectly.

Drawbacks: potential for oversight, code verbosity increases.

2. Make a subclass of QWidget (perhaps write this into `DetachableWidget`) that overrides `setMinimum*`, `setMaximum*` and `setFixed*` by doing essentially this again. This would only require swapping the inherited class, which in most cases is already DetachableWidget.

Drawbacks: subWin now needs to be stored inside the widget (perhaps it already is?), lots of boilerplate written in DetachableWidget (perhaps we should macro it?).

If anyone comes up with a more approachable version of 2., I'd be happy to implement it. Otherwise I can send 1. as soon as I'm done sticking it everywhere it needs to be.

**Comment by SpomJ** [comment]
I should have fixed SlicerT, Microtuner, LadspaBrowser and Tap Tempo.

Regardng VSTs - Rack appears to wannabe-constrained (and thus, not intended to maximize) which is weird, considering its max size is _strictly below_ its scroll area width (same applies to mixer btw). I also lack _any_ embed protocols to test how VST windows behave.

Not sure what you mean by {VST,LADSPA} effect windows??

**Comment by SpomJ** [comment]
I also just discovered that setting fixed width does _not_ prevent maximizing. Oh well, what a great thing Qt is.

**Comment by messmerd** [comment]
I tested the Linux AppImage that was automatically built for this PR, and there were some issues. The detach button didn't display correctly and the taskbar icon bug was back. However, when I built it locally, these issues were gone. I'm using Qt 5.15.2 locally, while the CI uses Qt 5.12.8, so they were likely upstream Qt bugs that got fixed.

#7316 updates our Qt version on Linux to Qt 5.15.2, so it would probably fix the CI issues here. I've just been waiting on people to review that PR.

**Comment by SpomJ** [comment]
@messmerd Since mixer can't be resized anymore, shouldn't er disable maximizing?

**Comment by SpomJ** [comment]
It was added back in c85b4982b048d4067ed5470788febd6f451c7de9 for some reason. This was written long before I joined the discussion so I didn't really question it lol

So should we revert that line?

**Comment by messmerd** [comment]
I haven't been following the recent work on the Mixer. If the Mixer in this PR deviates from the resize behavior on master, it should probably be reverted.

**Comment by SpomJ** [comment]
Also what's the desired behavior for VeSTige controller window? rn it's very janky, doesn't have a title nor a maximize button and its maximum width is less than embedded area width.

Should I unfix the width so that it can be maximized and resized? (Here it could also be beneficial to tamper with scrollarea to snap the controls into the empty space, but this is _really_ out of scope now)
or should I just set the flag to stop maximizing so that the behavior persists

**Comment by SpomJ** [comment]
#7688 fixes everything remaining I know of.

If no new issues are found, after its merge this pr will be actually ready this time.

**Comment by SpomJ** [comment]
Another issue I found with this is that opening any track in piano roll snaps it back to main window. This is probably due to the fact it "reopens" when the track is opened and this triggers the snap-back behavior buttons are supposed to do.

Something tells me reopening really shouldn't snap windows back.

**Comment by sqrvrt** [comment]
Is there even a point in differentiation between `DetachableWidget` and `DetachableWindow`? From what I can see the only difference in them is the latter applies to `QMainWindow` (in codebase - things like Piano Roll, Song Editor, etc.) instead of `Qwidget`.

But `QMainWindow` inherits `QWidget`, shouldn't we be able to apply `DetachableWidget` to windows too?

**Comment by messmerd** [comment]
@SpomJ 
> Is there even a point in differentiation between `DetachableWidget` and `DetachableWindow`?

They *are* mostly the same, so if one of them could be removed, that would be awesome.

**Comment by sqrvrt** [comment]
Heard some people say detaching straight up doesn't work on MacOS. If that's the case, perhaps we should add a check for mac in `detach()`?

**Comment by sqrvrt** [comment]
btw working on #8091 right now, if there are any concerns about code style/structure I'd be happy to know. It's apparently a near-half rewrite of the way this currently works (it seems I overestimated the size of this PR lol), so I'm afraid to leave redundant functionality somewhere.

**Comment by sqrvrt** [comment]
IMO it would also be interesting to have a setting to make all windows detached by default. What do you think?

**Comment by sqrvrt** [comment]
Can someone explain `SubWindow::eventFilter()`? I don't immediately see it applied anywhere in the class and it doesn't seem to be referenced outside too. And yet it compares `*obj` to child `widget()`. Does this mean it's never actually different from `QMdiSubWindow::eventFilter`?

File: src/gui/SampleTrackWindow.cpp [hunk 1]
```diff
@@ -257,27 +253,15 @@ void SampleTrackWindow::toggleVisibility(bool on)
...
-	}
-
 	m_stv->setFocus();
 	m_stv->m_tlb->setChecked(false);
 }
 
 
 
-void SampleTrackWindow::saveSettings(QDomDocument& doc, QDomElement & element)
+void SampleTrackWindow::saveSettings([[maybe_unused]] QDomDocument& doc, QDomElement& element)
```
sqrvrt: Is `[[maybe_unused]]` even needed here? I think it shouldn't give off a warning as a function parameter.

messmerd: It appears to be unnecessary - at least on GCC. I'll try removing it.

EDIT: Removing it worked.

sqrvrt:
`PianoRollWindow::showEvent()` calls `hide()` which softlocks the UI.
Conclusion: don't-show-piano-roll is being reworked

<details><summary>details</summary>
**Edit:**

To clarify, it looks like showing a TextFloat from a hidden detached window crashes the X11 server in its entirety, and when running wayland-native softlocks the GUI (though strace still shows _some_ activity).

Steps to reproduce:

1. Add two or more non-empty clips to an instrument track
2. Add an empty clip
3. Open the empty clip in Piano Roll
4. Detach piano roll
5. Remove the empty clip (=> Piano Roll window is hidden)
6. Open piano roll window with a toolbar button or the shortcut

On XWayland it appears to show an empty TextFloat before it dies

**Edit 2:**

It wasn't actually  a Textfloat itself, but rather the subsequent `hide()`. For some reason `PianoRollWindow::hide()` inside `PianoRollWindow::showEvent()` just deadlocks the entire thing and lags the host, possibly due to an infinite recursion.
</details>

messmerd (approved):
I've looked over the code and it looks fine to me.

I also did some testing on Linux Mint 22 + X11, and only noticed a few minor issues:
- The LADSPA plugin browser is maximizable when attached
- There's some weirdness when grabbing the About dialog and moving it around. When I grab it, the title bar of the attached Piano Roll window is also highlighted as if I was dragging it.
- Maximizing an attached window (such as the Song Editor) adds the LMMS icon to the File/Edit/View/... toolbar at the top for some reason.
- When detaching a window, it appears detached slightly below the position where it was attached. So if you were to repeatedly detach and reattach a window, it would slide down the screen. I don't think it's taking into account the height of the window's title bar.

These issues are all pretty minor though, so in the interest of getting this PR merged, I'm fine with dealing with these in a follow-up PR.

messmerd:
@yohannd1 did some testing and posted the results on Discord. I'm copying them here:

> Ok, did some testing here - per system report:
>
> - Niri (Linux, Wayland): spawned windows are not floating, but this might not be LMMS's fault (Niri is a tiling WM);,
> - Openbox (Linux, X11): works fine overall,
> - Windows 10: both MinGW64 and MSVC versions worked fine.,
>
> I found two issues that apply to all versions:
>
> - When a detached window is closed and it becomes a subwindow again, it always has its position limited to the visible area of the workspace. If possible I think it'd be good to not limit it, but that would probably be really bad for multi-monitor setups when the workspace is only on one screen.,
>
> - As messmerd mentioned on github, the window detaches on a position slightly below what would be intended - but I'm not sure if it is necessarily because of the height of the title bar, because the title bar on openbox is thinner than the attached-subwindow titlebar and it still goes around as much down as on windows 10, which has a really thick title bar.

And sqrvrt responded to these points: https://discord.com/channels/203559236729438208/712444046236450839/1464397740296376603

szeli1 (approved):
I briefly tested this PR and I couldn't find any major issues. I did not review the code. I found that If I maximize detached windows, then close them, they aren't maximized in lmms. Maybe this behavior is more desirable than keeping the window maximized.

messmerd:
Great work everyone!

Especially @sqrvrt - thanks to all the work you did reworking the whole thing, we were able to get this PR merged this decade rather than next.

As mentioned before, there are some known issues in this PR, though they are relatively minor and cosmetic - not significant enough to justify holding up this PR any longer. I'm hoping the momentum and excitement generated from this feature will help us get those remaining issues fixed relatively quickly.

PR #3535, opened 2017-05-07
LFO controller now has correct frequency with multiple connections.

### LINKED ISSUE #3064: LFO speed multiplied by number of connections
Steps to reproduce:
1) Add some LFO in the controller rack.
2) Add some FX effect to some instrument and connect Dry/Wet to created LFO.
3) Add another FX effect to same instrument and connect Dry/Wet to created LFO.

Expected result: LFO speed is same if both FX effects are active and when one of them is muted.

Actual result: LFO speed is multiplied by number of connections active.

Does not happen always but often. Usually after some time of playing. Stopping song and playing again makes LFO speed normal for a while.

Edit: Sometimes even LFO with single connection becomes several times faster than defined. Seems to be not related to CPU usage.

curlymorphic:
The LFO controller was playing at incorrect frequencies when their were multiple connections.

This was caused by the buffer counters not being correctly set.

m_bufferLastUpdated is now correctly set to the current frame upon updating
the buffer.

fixes #3064 

@Umcaruje 
Can you test this for me please, as you confirmed you could reproduce the bug in the original report

### DIFF
--- a/src/core/LfoController.cpp
+++ b/src/core/LfoController.cpp
@@ -113,6 +113,7 @@ void LfoController::updateValueBuffer()
 	}
 
 	m_currentPhase = absFraction( phase - m_phaseOffset );
+	m_bufferLastUpdated = s_periods;
 }
 
 void LfoController::updatePhase()


karmux:
@curlymorphic good find! I tested this and I can confirm that in this PR #3064 is fixed :+1:

zonkmachine:
Great! #3064 and also another [bug report](https://lmms.io/forum/viewtopic.php?f=7&t=6019) in the forums is fixed here. :+1:

Jousboxx:
Seems to work!

Windows build for anyone who is ~~n't on a real os~~ interested
https://drive.google.com/open?id=0B8dZ3IVOKKA6SWRWMXlwdXRVaXM

Umcaruje:
:+1:

Issue #181, opened 2014-01-28
Turn Autosave ON by default (and some other improvements)


unfa:
Simple as that.

I guess no user will be happy for saving extra 100 kB of disk for the cost of loosing extra 5 hours of work ;)

Make Autosave Feature ON by default!

And if you can - make it also save w few files, and prevent it from overwriting (without a warning!) itself when having two or three instances open. Every instance could have a random session ID and save the autosave files with this session-ID prefix like:

```
 57567-recovery-01.mmpz
 57567-recovery-02.mmpz
 57567-recovery-03.mmpz
 96817-recovery-01.mmpz
 57567-recovery-02.mmpz
 12422-recovery-01.mmpz
 29500-recovery-01.mmpz
```

Also a dedicated folder for autosave files could be nice, and a button for opening file browser in that directory would be great too.

unfa:
Or maybe the autosave can be handled by a separate thread, that operates on low priority (high nice value) so that it won't block DSP processing during playback?

softrabbit:
Autosave _is_ disabled during playback, right? https://github.com/LMMS/lmms/blob/master/src/gui/MainWindow.cpp#L1343

musikBear:
> Autosave is disabled during playback, right?

autosave should be disabled in all situations except idle. And definitively be optional.
If autosave kicks in as you are moving a block in edit-mode, The block 'drops' off, In automation editor, you draw wrong. It is completely annoying, and **must** be optional, also a user time setting for like 5, 10, 15, 30 mins between autosave, would be advice able - *But* !!
*Definitively optional* as it is now 
I think a tiny icon in main bar, just flashing a couple of times, would be better, then the user has the choice to do a version save, or a regular save, or nothing.
Here are autosave profiles for cpu usage (0.4.9 i believe it was.. Toby gracefully gave me a exe where it was disabled, then it became optional)
![spikedrecoversavewithmanualtask](https://cloud.githubusercontent.com/assets/6662066/6917197/b894d0fe-d7a2-11e4-9b6c-fedf6b60f122.jpg)

Sti2nd:
Wait, what is wrong with using your processor again?

musikBear:
@Sti2nd  -you are right, nothing wrong with that, but (on my system at least) this process results in lmms un-responsiveness, and a break in the functionality for ~1-2 secd -Thats the bad thing, about it, but i have a new angle on this, i just have to formulate this new proposal for a new kind of autosave.

[...]

PR #3088, opened 2016-10-24
Option to allow auto save while playing

zonkmachine:
Since most systems have no problems saving a project while playing we could add this as an option.
The problem here is that if you edit a project while playing (or just do very short stops so the autosave doesn't have time enough to start) you may end up not getting a backup for a very long time and you wouldn't know about it.

autosaveagain

PR #3541, opened 2017-05-08
Fix regression where autosave function is off by default

zonkmachine:
Autosave regression https://github.com/LMMS/lmms/issues/181#issuecomment-299703956

Introduced here: https://github.com/LMMS/lmms/commit/901fea51d064df9fd2665c6b737f7a70ed18b4db

### DIFF
--- a/src/gui/SetupDialog.cpp
+++ b/src/gui/SetupDialog.cpp
@@ -119,8 +119,8 @@ SetupDialog::SetupDialog( ConfigTabs _tab_to_open ) :
 #endif
 	m_backgroundArtwork( QDir::toNativeSeparators( ConfigManager::inst()->backgroundArtwork() ) ),
 	m_smoothScroll( ConfigManager::inst()->value( "ui", "smoothscroll" ).toInt() ),
-	m_enableAutoSave( ConfigManager::inst()->value( "ui", "enableautosave" ).toInt() ),
-	m_enableRunningAutoSave( ConfigManager::inst()->value( "ui", "enablerunningautosave" ).toInt() ),
+	m_enableAutoSave( !ConfigManager::inst()->value( "ui", "enableautosave" ).toInt() ),
+	m_enableRunningAutoSave( !ConfigManager::inst()->value( "ui", "enablerunningautosave" ).toInt() ),
 	m_saveInterval(	ConfigManager::inst()->value( "ui", "saveinterval" ).toInt() < 1 ?
 					MainWindow::DEFAULT_SAVE_INTERVAL_MINUTES :
 			ConfigManager::inst()->value( "ui", "saveinterval" ).toInt() ),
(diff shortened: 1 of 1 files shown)


tresf:
@zonkmachine thanks.  Although this fixes the problem in the UI, can you help explain what's happening in the config?  Is this value being stored incorrectly as  code looks backwards now. :)

tresf:
Ok... I think I see what happened... @jasp00 complained about double-negatives in our variable names, but we don't currently seem to have a way to default an unconfigured property to `enabled`, right?

zonkmachine:
I'm basically just copying the logic of the other 'default on' buttons. SetupDialog.cpp and I aren't really on speaking terms...

zonkmachine:
Merge?


> but we don't currently seem to have a way to default an unconfigured property to enabled, right?

Yes, something like that. In any case I've never had an easy time when I've touched that file.

tresf:
> Merge?

No unfortunately now the code is backwards, so the patch is masking the problem.

@jasp00's advice was pragmatic but impractical.  We either revert the variable naming to what the project uses and keep this type of patch, or spend the time necessary to have default-on in our config file.

zonkmachine:
> We either revert the variable naming to what the project uses and keep this type of patch,

I didn't get this. 



> or spend the time necessary to have default-on in our config file.

Sorry. I don't have much time now.

tresf:
I'm not sure how else to phrase this properly.... 

```cpp
m_enableAutoSave( !ConfigManager::inst()->value( "ui", "enableautosave" ).toInt() ),
```

This essentially says 
```cpp
m_enableAutoSave = !enableautosave;
```

These are the same conceptual concerns [@jasp00 raised](https://github.com/LMMS/lmms/pull/3088#pullrequestreview-20149999) in #3088.

There's likely a better way to handle this all in our `ConfigManager` class to actually fix the underlying problem of default int value is zero.  Alternately, we can go back to the old `disableautosave` verbiage and just let it be zero naturally, but the logic proposed in this PR is incorrect.

tresf:
Ok... I'd appreciate a second set of eyes on this, but if what I'm observing is true, all of our booleans can never be true, which punishes anyone that wants a "default on" value without a "default off" name.

Here's a hacked proposal... 

Note, this has some implications... namely, it potentially enables `animateafp`, `displaydbfs` and `enableautosave` by default.  It's the wrong way, but illustrates the problem.

```diff
const QString & ConfigManager::value( const QString & cls,
					const QString & attribute ) const
{
	if( m_settings.contains( cls ) )
	{
		for( stringPairVector::const_iterator it =
						m_settings[cls].begin();
					it != m_settings[cls].end(); ++it )
		{
			if( ( *it ).first == attribute )
			{
				return ( *it ).second ;
			}
		}
	}
-	static QString empty
+	// Default enable flags to true
+	static QString empty = attribute.startsWith("enable") || 
		attributes.startsWith("display")  ||
		attributes.startsWith("animate") ? "1" : "";
	return empty;
}
```

The better proposal... add new function that accepts a default value when one can't be found... (pseudocode, untested)

```cpp
const QString & ConfigManager::value( const QString & cls, const QString & attribute, const QString & defaultVal ) const
{
	QString val = value(cls, attribute);
        return val.isEmpty() ? defaultVal : val;
}

zonkmachine:
> Alternately, we can go back to the old disableautosave verbiage and just let it be zero blank -- resolving to zero -- naturally, but the logic proposed in this PR is incorrect.

So, just reverting 901fea51d064df9fd2665c6b737f7a70ed18b4db ?
I suggest you start a separate issue with the above proposed code. I'm not going to have time to implement anything like that here and now.

tresf:
> So, just reverting 901fea5 ?

That was my original thought, but please read the [bottom of the post](https://github.com/LMMS/lmms/pull/3541#issuecomment-300361815), I provided a better fix that's only a few lines and allows us to keep the `enabled` flag.

zonkmachine:
> That was my original thought, but please read the bottom of the post, I provided a better fix that's only a few lines and allows us to keep the enabled flag.

Still not within my time limit. What I can do is revert 901fea5. I won't have time to look any deeper into things right now and for a possible long time coming.

tresf:
> Still not within my time limit. 

Someone else will have to tackle it then.  This PR is as good as a bug report since it contains the valid info.  We'll either supercede this will a new PR or use this PR directly for the code.

tresf:
Superseded by https://github.com/LMMS/lmms/pull/3551.

PR #3032, opened 2016-09-11
Multiple instance recover files

zonkmachine:
Fixes #181

Design request. Not ready for code review but I'm curious as to how y'all think the recover file system should behave.
It is currently a working rudimentary implementation of recovery files for multiple sessions.
LMMS is given a unique 8 char hex ID and sticks this to the recover files belonging to the session. On recovery the ID is reused as session ID. The files are sorted so if you have more than one `.recover.mmp` the system will suggest the most recent for recovery.
Limited Session is dropped and maybe all sessions should be dropped?
Maybe we shouldn't prompt for recover files at all on program launch but just open up a session and if you had a crash there could be an indicator (red R or something) on the MainWindow, maybe superimposed upon the `open existing project` icon.? Don't know. Ideas?

PR #3545, opened 2017-05-08
Remove Limited Session

zonkmachine:
As Limited Session is useful for precisely no one at all, I'm removing it. This is commit one from the https://github.com/LMMS/lmms/pull/3032 rebased against stable-1.2 .

### DIFF
--- a/include/MainWindow.h
+++ b/include/MainWindow.h
@@ -110,8 +110,7 @@ public:
 	enum SessionState
 	{
 		Normal,
-		Recover,
-		Limited,
+		Recover
 	};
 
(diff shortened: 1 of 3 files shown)


zonkmachine:
@tresf @Umcaruje This is good stuff. The Limited Session only really worked for me while I was actively working with the PR introducing it itself but it is of no value to keep.This is not a good substitute for multiple session recover files but just confuse the user. I will merge this is in a day or so if there are no objections.

zonkmachine:
No objections, merge. :tractor:

Umcaruje:
:+1:

PR #3550, opened 2017-05-10
VeSTige opens correct folder

karmux:
This pull request fixes opening correct VST folder in previously saved projects with existing VeSTige instruments. Project has to be saved using code in this PR. After next project load VST folders open correctly. Copied logic from SF2 and GIG Players.

### DIFF
--- a/plugins/GigPlayer/GigPlayer.cpp
+++ b/plugins/GigPlayer/GigPlayer.cpp
@@ -1066,7 +1066,6 @@ void GigInstrumentView::showFileDialog()
 	types << tr( "GIG Files (*.gig)" );
 	ofd.setNameFilters( types );
 
-	QString dir;
 	if( k->m_filename != "" )
 	{
 		QString f = SampleBuffer::tryToMakeAbsolute( k->m_filename );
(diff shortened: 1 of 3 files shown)


tresf:
@karmux thanks.

First, what happens if the VST does not exist at the given path?  Does it know to fallback to the VST directory?

Second, have you considered consolidating this into the `FileDialog` constructor or a `FileDialog` helper?  This seems like it's only going to be more useful as time goes on and it will cleanup the codebase.  If we're worried about changing too many files on `stable-1.2`, this could be a follow-up task for master.

Reason being... we already have areas where the existing constructor could simplify the code since the existing constructor supports `FileDialog::FileDialog( QWidget *parent, const QString &caption, const QString &directory, const QString &filter )` however I feel we can make the constructor even better by taking a file path and resolving its parent using your logic with a fallback path if it's not found.

karmux:
@tresf Good that you asked that. It didn't default to configured VST plugin directory if the VST plugin did not exist at the given path. SF2 and GIG players seemed to default to plugins directory defined in config. I fixed this in similar way to GIG Player.

I was thinking about some general helper functionality for all these instrument plugins but that would be safer to implement for next version after 1.2 I think.

tresf:
@karmux thanks!

PR #3551, opened 2017-05-11
Provide support for fallback config values

tresf:
Supersedes #3541 to address autosave regression https://github.com/LMMS/lmms/issues/181#issuecomment-299703956.

Previous to this PR, there was no way to provide a "default on" style config value.  The effect of this was variable names such as `disableautosave` causing double negatives in our codebase making it hard to read.  Attempts to change them have resulted in a mismatch between desired default and actual default.  This should allow config values to have default values.

e.g. 

```diff
- m_enableAutoSave( ConfigManager::inst()->value( "ui", "enableautosave" ).toInt() ),
+ m_enableAutoSave( ConfigManager::inst()->value( "ui", "enableautosave", "1" ).toInt() ),
```

Booleans are parsed as integers.  A default value of `1` will make a checkbox true.  This will allow us to start moving toward more properly named variables as explained very well here https://github.com/LMMS/lmms/pull/3088#discussion_r99468228.

### DIFF
--- a/include/ConfigManager.h
+++ b/include/ConfigManager.h
@@ -217,6 +217,9 @@ public:
 
 	const QString & value( const QString & cls,
 					const QString & attribute ) const;
+	const QString & value( const QString & cls,
+					const QString & attribute,
+					const QString & defaultVal) const;
 	void setValue( const QString & cls, const QString & attribute,
 						const QString & value );
 	void deleteValue( const QString & cls, const QString & attribute);
--- a/include/ConfigManager.h
+++ b/include/ConfigManager.h
@@ -219,7 +219,7 @@ public:
 					const QString & attribute ) const;
 	const QString & value( const QString & cls,
 					const QString & attribute,
-					const QString & defaultVal) const;
+					const QString & defaultVal ) const;
 	void setValue( const QString & cls, const QString & attribute,
 						const QString & value );
 	void deleteValue( const QString & cls, const QString & attribute);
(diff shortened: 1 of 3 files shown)


Umcaruje:
Tested this one out. 
image
After removing `.lmmsrc.xml` autosave and playback cursor is on :+1:

include/ConfigManager.h
```diff
@@ -217,6 +217,9 @@ class EXPORT ConfigManager
 
 	const QString & value( const QString & cls,
 					const QString & attribute ) const;
+	const QString & value( const QString & cls,
+					const QString & attribute,
+					const QString & defaultVal) const;
```
Umcaruje: you should have a space between the parenthesis : `defaultVal )`

src/gui/SetupDialog.cpp
```diff
@@ -131,7 +131,7 @@ SetupDialog::SetupDialog( ConfigTabs _tab_to_open ) :
 	m_syncVSTPlugins( ConfigManager::inst()->value( "ui",
 							"syncvstplugins" ).toInt() ),
 	m_animateAFP(ConfigManager::inst()->value( "ui",
-						   "animateafp").toInt() ),
+						   "animateafp", "1").toInt() ),
```
Umcaruje: same here: `"1" )`

tresf:
> Other than the two formatting issues, this looks good to me.

Formatting issues fixed.

PR #3501, opened 2017-04-12
Fixed #3498. Show missing pixmaps in VeSTige plugin.

karmux:

PR #3554, opened 2017-05-12
Show missing pixmaps in VST effects dialog.

karmux:
Show missing pixmaps in VST effects dialog.
m_openPresetButton consistency with VeSTige.

Same as #3501 but for VST effects dialog.


tresf:
@karmux is this the last of them or would it be worth sweeping the entire project?

karmux:
@tresf I haven't noticed any more missing pixmaps. Hopefully this is the last one.

tresf:
Just grep'd the source code and couldn't find any obvious offenders either.  Merging.  Thanks!

PR #2599, opened 2016-02-22
Allow WidgetTab to use artwork tabs

StCyr:
This PR relates to https://github.com/LMMS/lmms/issues/2528

Br,

Cyrille

PR #3569, opened 2017-05-21
Allow WidgetTab to use artwork tabs (@StCyr)

Umcaruje:
This is an effort to bring #2599 to a closure. The original PR went stale, and most of the work has been done, so it'd be a shame for it to never get implmeneted, as I think it's a nifty feature.

I rebased the original commits to stable-1.2 and will address the remaining issues. I couldn't push to @StCyr's branch cause I don't have access and github didn't have the feature to allow edits form mantainers at the time of that PR.

Here's how it looks now:  
image
I'll be changing the colors shortly. @RebeccaLaVie input appreciated.


Umcaruje:
Adressed @tresf and @jasp00 's [comments](https://github.com/LMMS/lmms/pull/2599/files#r71420352) from the original PR. The name is used as a tooltip without a seperate variable. The names are as follows:
1. Plugin
1. Envelope, filter & LFO
1. Chord stacking & arpeggio
1. Effects
1. MIDI settings
1. Miscellaneous

Umcaruje:
After a session with the community on discord, this is how the tabs look now: 
image

There are some spacing issues that I'd like to address tomorrow.

src/gui/widgets/TabWidget.cpp
```diff
@@ -133,88 +198,100 @@ void TabWidget::resizeEvent( QResizeEvent * )
...
+		// Draw a text tab or a artwork tab.
+		if ( m_usePixmap )
 		{
-			p.setPen( QColor( 32, 48, 64 ) );
-			p.fillRect( tab_x_offset, 2, ( *it ).nwidth - 6, 10, cap_col );
+			// Fixes tab's width, because original size is only correct for text tabs
+			( *it ).nwidth = tab_width;
+
+			// Get artwork
+      QPixmap artwork( embed::getIconPixmap( ( *it ).pixmap ) );
```
Umcaruje: note to self: whitespace here

qnebra:
Tested and works nicely, not causing errors on usage.

Umcaruje:
@LMMS/developers anyone want to review the code on this? Would be appreciated

include/TabWidget.h
```diff
@@ -57,16 +79,26 @@ class TabWidget : public QWidget
 private:
 	struct widgetDesc
 	{
-		QWidget * w;	// ptr to widget
-		QString name;	// name for widget
-		int nwidth;	// width of name when painting
+		QWidget       * w;	   	// ptr to widget
+		const char    *pixmap; 	// artwork for the widget
```
tresf: Comment can be formatted like the others. :)

tresf:
@Umcaruje this looks pretty good.  I haven't reviewed it for any syntax errors, but the change appears to be quite minor and straightforward despite the number of files.

The major change I see is the draw code has moved from resize event to paint event which may have slight performance implications, but the draw routines are greatly simplified as well.  Unfortunately I don't have the experience with updating paint routines to know where they should go from a design perspective.

Umcaruje:
>The major change I see is the draw code has moved from resize event to paint event which may have slight performance implications, but the draw routines are greatly simplified as well. Unfortunately I don't have the experience with updating paint routines to know where they should go from a design perspective.

Well those were @StCyr's changes, and it makes sense to me to draw in the paint event. Anyways I didn't notice any performance impacts and neither did @qnebra from what I've gathered.

This looks ready to merge then. Will fix the comment formatting.

Umcaruje:
image
Here's the final look of the icons, realised I didn't post a screenshot before

Umcaruje:
I'll leave this sitting for 24 hours, and merge after that.

tresf:
>  it makes sense to me to draw in the paint event. 

This is the part I'm a bit confused about as it was working before and only being triggered on construction or resize.

But the way he made it is more consistent with how we do it today in other classes, so it's only an observation.

To that point, calling `update()` on construction seems quite redundant, no?  It made sense before when the constructor wasn't using `paintEvent`, but now that it is it should be called naturally as part of the widget constructor.  It's not harmful `update()` can be called just about anywhere, but I think it's redundant.  If you look at other classes (e.g. `Fader`, `PixmapButton` that override `paintEvent()`, they do not call `update()` in the constructor.

Umcaruje:
>It's not harmful update() can be called just about anywhere, but I think it's redundant.

I agree, will look into removing it.

Umcaruje:
@tresf done.

Issue #3166, opened 2016-12-17
Time signature changing doesn't update song editor grid immediately


BaraMGB:
You have to click in the grid for updating the grid or the time line after changing the time signature. 
Edit: ~~only if the song is not playing~~
Edit2: only on position 0

PR #3572, opened 2017-05-22
fix time signature updates timelinewidget also on pos 0

BaraMGB:
fix https://github.com/LMMS/lmms/issues/3166


BaraMGB:
This is an easy one liner. merge.

PR #3573, opened 2017-05-24
Proper Case Renames

miketurn:
@liushuyu 
Please merge my "Patch 1" changes if this is done correctly.

@tresf 
As per our discussion a while back, you mention that in order to make a batch of changes, I should select the "Patch 1" in my own profile, make all the changes to it, and then make one pull request for any changes in that batch (Patch 1).

Sorry to bother you guys, but can someone just check that I have done this correctly.
I think I have a friend that will be joining the site and will be helping with minor updates like these.
I will show him how to do this exactly the way that I did to my "Patch 1", so I just want to make sure that I am doing this right so that I will not be teaching him incorrectly.

Thank You

### DIFF
--- a/src/gui/PluginBrowser.cpp
+++ b/src/gui/PluginBrowser.cpp
@@ -50,7 +50,7 @@ static bool pluginBefore( const Plugin::Descriptor* d1, const Plugin::Descriptor
 
 
 PluginBrowser::PluginBrowser( QWidget * _parent ) :
-	SideBarWidget( tr( "Instrument plugins" ),
+	SideBarWidget( tr( "Instrument Plugins" ),
 				embed::getIconPixmap( "plugins" ).transformed( QTransform().rotate( 90 ) ), _parent )
 {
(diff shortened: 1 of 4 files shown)


Umcaruje:
I'm sorry but is this really the case we want?

Sentence case would make a lot more sense to me here rather than capitalizing each word.

karmux:
`Instrument Plugins` and `Project Notes` makes them consistent with other window titles but for other changes I agree with @Umcaruje. They are more like sentences not titles.

tresf:
Thanks for the feedback.  I examined various parts of the interface and we use "Title Case" for window titles and sections and "Sentence case" for buttons and sentences.

@liushuyu FYI, this will affect translations.

tresf:
Commit history is strange for this one.  GitHub split this into 4b92651d4f011952e58c11da6e858c4e6884115e and https://github.com/LMMS/lmms/commit/1f66f62ed76b85d8ccfdb1eb7ca9cca9595f5553.  Not sure why, just an FYI.

Issue #3590, opened 2017-05-30
sampletrack mutes the playing sample by muting a different one


BaraMGB:
![mutingsampletrack](https://cloud.githubusercontent.com/assets/6502580/26573085/add4c1b2-451c-11e7-9538-bdc66b076f52.jpg)

By muting/unmuting a sample on a sampletrack the playing sample will be muted, too.

PhysSong:
Changing sample file of other sample track object also mutes the playing sample track, too.

https://github.com/LMMS/lmms/blob/stable-1.2/src/tracks/SampleTrack.cpp#L168
Engine::mixer()->removePlayHandlesOfTypes( getTrack(), PlayHandle::TypeSamplePlayHandle );
in SampleTCO::playbackPositionChanged()

It removes play handle(s) of currently playing sample(s). I think it causes this problem.

PR #3133, opened 2016-11-26
Let sample tracks play from any song position

BaraMGB:
Okay, let's test this. I hope I found a solution that works for all.

PR #3591, opened 2017-05-30
fixes sample muting bug for sampletrack

BaraMGB:
fixes : https://github.com/LMMS/lmms/issues/3590

### DIFF
--- a/include/SampleTrack.h
+++ b/include/SampleTrack.h
@@ -153,6 +153,7 @@ public:
 
 public slots:
 	void updateTcos();
+	void setPlayingTcos( bool isPlaying );
 
 private:
 	FloatModel m_volumeModel;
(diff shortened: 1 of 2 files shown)


PhysSong:
Good.
Maybe #3133 have caused #3590.
Could we change SampleTrack::updateTcos() into more simple form?

BaraMGB:
>Could we change SampleTrack::updateTcos() into simpler form?

Yes, thinking about that brings me to another problem. If we have more than one sample track this would not work. 

I'm on it.

BaraMGB:
>~~Yes, thinking about that brings me to another problem. If we have more than one sample track this would not work.~~

That's not true. I updated the PR.

BaraMGB:
If no one has objections about this I'll merge in few hours.

tresf:
FYI, cherry-picked to `liushuyu/merge` via 8c17b39.

PR #1891, opened 2015-03-21
Use Qt's Resource System (2nd approach)

Lukas-W:
This PR does pretty much the same as #1742, but takes a less invasive approach, touching much less code. In contrast to my first approach, `embed::getIconPixmap` stays (I removed it and replaced all calls in #1742), but uses Qt's resource system and `QDir`'s search paths internally.

~~Even though I didn't test it, it should now be~~ It's now possible to drop in SVGs instead of PNGs in themes without changing any code.

It's already been possible to replace any plugin's artwork by adding a file called something like e.g. `monstro_logo.png` to the theme folder. LMMS will then use that one instead of the plugin's built-in file. This PR changes this to `monstro/logo.png` instead (subdirectory vs underscore), just for the sake of easy implementation. I could not find any theme that replaces a plugin's artwork, so I hope this will have no impact on existing themes. Please let me know if I'm wrong and if there are such themes out there (on the LSP).

PR #3607, opened 2017-06-04
Merge `stable-1.2` into `master`

liushuyu:
This change is [Reviewable](https://reviewable.io/reviews/lmms/lmms/3607)


liushuyu:
Just to make sure `master` got all the fixes big and small from 1.2 branch

And I found new text strings were introduced but not yet exist in `master` branch

karmux:
Changes to from #3604 for plugins/vst_base/CMakeLists.txt are missing.
I suggest to keep short lines, just remove quoting.

tresf:
@karmux good point... that whole block needs to be rewritten because `stable-1.2` added the `-L` in the `${EXTRA_FLAGS}` whereas master seems to have added `-L` explicitly in the command string.  I also am curious how the value of `${WINE_LIBRARY_DIR}` compares to e.g. `/usr/lib/i386-linux-gnu/wine/`, as the prior would be cleaner.  Perhaps we just shim it into `FindWine.cmake` and cleanup the `vst_base/CMakeLists.txt`.

> I suggest to keep short lines, just remove quoting.

I'm a fan of that.  While we're at it, we should make it simpler in general... 

`vst_base/CMakeLists.txt`
```cmake
	ARGS -I${CMAKE_BINARY_DIR}
 		-I${CMAKE_SOURCE_DIR}/include
		# TODO:  Is this next line even a valid directory?
		-I${CMAKE_INSTALL_PREFIX}/include/wine/windows
		-I${CMAKE_INSTALL_PREFIX}/include
		-I${WINE_INCLUDE_BASE_DIR}
		# TODO: Is next line is redundant with -L/foo/bar in ${EXTRA_FLAGS}?
		-L${WINE_LIBRARY_DIR}
		${CMAKE_CURRENT_SOURCE_DIR}/RemoteVstPlugin.cpp
		-ansi -mwindows -lpthread -fno-omit-frame-pointer
		${EXTRA_FLAGS} ${WINE_BUILD_FLAGS} -o ../RemoteVstPlugin
	# Ensure correct file extension
	COMMAND sh -c "mv ../RemoteVstPlugin.exe ../RemoteVstPlugin || true"
```

liushuyu:
PR is now refreshed and updated with changes from both branches

karmux:
Merge?

liushuyu:
>Merge?

Who can merge PRs is always a mystery in LMMS. 😄 
 
And Travis CI is failing due to unable to resolve `homebrew` dependencies, maybe someone would like to investigate the problem.

tresf:
> Who can merge PRs is always a mystery in LMMS.

Well we have a varying mix of competency in this project so some PRs are easier to merge than other.  This one is a monster.  @liushuyu @karmux if you two can agree on "OK", then do it.

karmux:
We should do smaller merges more often to avoid monsters like this.

Issues that me and @tresf pointed out are fixed now. Looks OK by me. There were just 4 conflicting files in this merge.

liushuyu:
>We should do smaller merges more often to avoid monsters like this.

I can't be more agree on this point. I personally think if this PR were not proposed, a "mega" PR will appear and put on the desk sooner or later.

I am also okay with this, except that Travis CI is not feeling good.

plugins/ReverbSC/CMakeLists.txt
```diff
@@ -7,18 +7,21 @@ SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++0x")
...
-	ReverbSCControls.h 
-	ReverbSCControlDialog.h 
+	MOCFILES
+	ReverbSCControls.h
+	ReverbSCControlDialog.h
 	base.h
 	revsc.h
 	dcblock.h
-	EMBEDDED_RESOURCES artwork.png logo.png
+	EMBEDDED_RESOURCES "${CMAKE_CURRENT_SOURCE_DIR}/*.png"
```
PhysSong: Why these three lines appear twice?

plugins/ReverbSC/CMakeLists.txt
```diff
@@ -7,18 +7,21 @@ SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++0x")
...
-	MOCFILES 
 	ReverbSC.h
-	ReverbSCControls.h 
-	ReverbSCControlDialog.h 
+	MOCFILES
+	ReverbSCControls.h
+	ReverbSCControlDialog.h
 	base.h
 	revsc.h
 	dcblock.h
```
PhysSong: Why these three lines appear twice?

plugins/ReverbSC/CMakeLists.txt
PhysSong: And this line shouldn't be changed, because of https://github.com/LMMS/lmms/commit/9f905bce3eee4e24a2338d9d8a104af6bdbec8ad#diff-ea4a8d6b4968ac81d0ffb07a45a4c7e6 in #1891.

plugins/ReverbSC/CMakeLists.txt
liushuyu: This is because `git` sometimes do wrong merges, need to correct this by doing a new `git merge` action (cleaner approach)

src/CMakeLists.txt
```diff
@@ -45,8 +45,9 @@ INCLUDE(GenQrc)
 ADD_GEN_QRC(LMMS_RCC_OUT lmms.qrc
 	"${CMAKE_SOURCE_DIR}/doc/AUTHORS"
 	"${CMAKE_SOURCE_DIR}/LICENSE.txt"
-	"${CMAKE_SOURCE_DIR}/doc/CONTRIBUTORS"
-)
+	"${CMAKE_SOURCE_DIR}/doc/CONTRIBUTORS")
+SET(LMMS_ER_H "${CMAKE_CURRENT_BINARY_DIR}/embedded_resources.h")
+ADD_CUSTOM_COMMAND(OUTPUT "${LMMS_ER_H}" COMMAND "${BIN2RES}" ARGS ${lmms_EMBEDDED_RESOURCES} > "${LMMS_ER_H}" DEPENDS bin2res)
```
PhysSong: Here, too.

PhysSong:
I think this PR is fine except failing travis-ci and these three problems.

liushuyu:
>I think this PR is fine except failing travis-ci and these three problems.

Thanks for your review work. I'll try a new merge asap.

This PR is very large in scale, which should be avoided. And we need to consider first push stable fixes to `master` then back port to `stable` branch instead of pushing to `stable` branch and then merge the fix to `master

tresf:
> first push stable fixes to `master` then back port to `stable` branch instead of pushing to `stable` branch and then merge the fix to `master`

I'm not sure what additional benefit this adds besides being able to use the word "back port" :)

Once the branches have diverged significantly, the conflicts will arise regardless.  If we do decide to do the way you've proposed, it can be as simple as keeping `master` the default branch (always) but the `stable` to `master` is how the predecessors did it and it was fine.

plugins/vst_base/CMakeLists.txt
```diff
@@ -61,16 +61,15 @@ STRING(REPLACE " " ";" WINE_BUILD_FLAGS ${CMAKE_CXX_FLAGS} " " ${CMAKE_EXE_LINKE
...
-	ARGS "-I\"${CMAKE_BINARY_DIR}\""
-		"-I\"${CMAKE_SOURCE_DIR}/include\""
-		"-I\"${CMAKE_INSTALL_PREFIX}/include/wine/windows\""
-		"-I\"${CMAKE_INSTALL_PREFIX}/include\""
-		"-I\"${WINE_INCLUDE_BASE_DIR}\""
-		"-L\"${WINE_LIBRARY_DIR}\""
-		"\"${CMAKE_CURRENT_SOURCE_DIR}/RemoteVstPlugin.cpp\""
+	ARGS -I${CMAKE_BINARY_DIR}
+		-I${CMAKE_SOURCE_DIR}/include
+		-I${WINE_INCLUDE_BASE_DIR}
```
PhysSong: @karmux Is the deletion of two includes intentional?

karmux: @PhysSong yes

.travis/osx..install.sh
```diff
@@ -2,7 +2,7 @@
 
 set -e
 
-PACKAGES="cmake pkg-config fftw libogg libvorbis libsndfile libsamplerate jack sdl stk portaudio node fltk"
+PACKAGES="cmake pkgconfig fftw libogg libvorbis lame libsndfile libsamplerate jack sdl libgig libsoundio stk portaudio node fltk"
```
PhysSong: If I don't change pkgconfig to pkg-config, Travis-CI succeeds. Please check and fix this.

tresf: `pkgconfig` is an [alias to `pkg-config`](https://github.com/Homebrew/homebrew-core/blob/master/Aliases/pkgconfig).  Travis is wrong.

tresf: > If I change pkgconfig to pkg-config, Travis-CI succeeds. Please check and fix this.

I think this is a coincidence.  Travis is erroring on `shell_session_update: command not found` which seems to be a [ruby issue](https://github.com/direnv/direnv/issues/210#issuecomment-269259123).  We can either merge with Mac failing, or try the recommended fix.

PhysSong: I think it might conflicts with `for` loop below it. @tresf could you test?

tresf: On what evidence?  Here's another bug report with people complaining 14 days ago.  https://github.com/travis-ci/travis-ci/issues/6307#issuecomment-280644135

tresf: It's most likely to using [xcode 8.2](https://github.com/liushuyu/lmms/blob/5ec242bf29859d71cc0554c25f2f89bbcd17ecfd/.travis.yml#L15) since the Travis-CI bug report mentions that multiple times.  We can merge with this failing, it's fine.

zapashcanon: @PhysSong is right, it's about the `for loop`.

It's because homebrew fails if we try to install a package which is already there. (Because of `set -e`).

So, first we remove already existing packages from the list of packages to install.
Unfortunately, it won't work with the wrong alias. Because, when we ask homebrew the list of installed packages, it tells pkg-config and not pkgconfig. So we don't remove it from the list, so we try to install it but it's already there, so homebrew fails...

To avoid this, I opened an issue on homebrew a few time ago, but people closed it without providing what I would consider a good fix. So... we should use `pkg-config`.

tresf: @zapashcanon thanks for the detailed explanation.  So let's hard-code a `pkgconfig` search as well (right?) to avoid the conflict and then only inject `pkg-config` if neither `pkgconfig` or `pkg-config` is found in the list of installed packages.  Does that make sense?

PhysSong: @tresf That may make sense. However, I think leave `pkg-config` is the better.
It was `pkg-config` since #3476 by @zapashcanon,  See https://github.com/LMMS/lmms/commit/7844b3e289e3627bf89d5c3048ac2ab03b33fff5#diff-199dc21cb5e108771eecb4815a82b450R5.

PhysSong: In addition, I found `shell_session_update: command not found` exists in some successful builds. I think it is a seperated issue.

zapashcanon: > I think leave `pkg-config` is the better

I thinks it's the better too. Homebrew shouldn't change the displayed output. Here it's just because of the merge. It won't happen again I guess.

PhysSong:
I think a wrong merge have done, or there's a mistake while resolving confilcts. Should @liushuyu or any other person should try a cleaner and right merge?

karmux:
@PhysSong what's the issue still? 

There were only 4 conflicting files in this merge:
```
CONFLICT (content): Merge conflict in src/CMakeLists.txt
CONFLICT (content): Merge conflict in plugins/vst_base/CMakeLists.txt
CONFLICT (content): Merge conflict in plugins/ReverbSC/CMakeLists.txt
CONFLICT (content): Merge conflict in .travis/osx..install.sh
```

liushuyu:
>@PhysSong what's the issue still?

Well, I think. @PhysSong wants to do a cleaner and correct merge from the ground up. It's harder though, but I suppose it could have a much cleaner history graph

tresf:
> Well, I think. @PhysSong wants to do a cleaner and correct merge from the ground up. It's harder though, but I suppose it could have a much cleaner history graph

@PhysSong please do. 👍 You may use @liushuyu's PR as a baseline for comparison and then we can close this out.

PhysSong:
@liushuyu Could you tell me how did you make and add commits?

liushuyu:
>@liushuyu Could you tell me how did you make and add commits?

Do you want to rework the merge and start everything from the ground up or just fix mine? The procedures can be different

PhysSong:
@tresf Which merge method should be used? If I know that, I'll check and do some tests with it.

liushuyu:
>@tresf Which merge method should be used? If I know that, I'll check and do some tests with it.

I personally recommend regular 3 way `git merge` in this case. Using rebase will mess up everything ( 'cause i have tried)

tresf:
> @tresf Which merge method should be used? If I know that, I'll check and do some tests with it.

This isn't my area of expertise.  Please feel free to try a few things out.

PhysSong:
I tried both `git merge` and `git rebase`, 
```
git checkout -b new_master stable-1.2
git rebase master
```
gave me the best result.
However, It will make branch comparison in Github like this: This branch is 105 commits ahead, 69 commits behind LMMS:stable-1.2.
I saw Github says @liushuyu's merge branch is `48 commits ahead, 5 commits behind LMMS:stable-1.2. ` @liushuyu Please tell me how did you do.

liushuyu:
PhysSong Which branch did you take as base branch? It should be `master`

liushuyu:
And @PhysSong please use `git merge ` in this case, rebase will override the history making GitHub think there are new commits added

PhysSong:
@liushuyu Yes. `git rebase` has the problem you said.
If I use `git merge`, however, `git` fails to handle cherry-picked commits properly. See [this](https://davitenio.wordpress.com/2008/09/27/git-merge-after-git-cherry-pick-avoiding-duplicate-commits/) for more details.

As I know, any of these methods doesn't generate correct and clean commit histories(unless someone force-push commits into master - definitely not good).

liushuyu:
Problem solved.
@PhysSong Please review the merge again and see if there's something worth noting

PhysSong:
@liushuyu Seems good, except three comment changes in #3614 isn't included.
It isn't necessary, but including them would be better.

liushuyu:
>@liushuyu Seems good, except three comment changes in #3614 isn't included.
It isn't necessary, but including them would be better.

 It's included, but a new separate commit was not created for the merge.

PhysSong:
@liushuyu The include directory change in 242d80d is included. However, Three comment changes aren't.

liushuyu:
>@liushuyu The include directory change in 242d80d is included. However, Three comment changes aren't.

Ah. Okay, so you think it's very vital to include those changes as well?

Let's just merge it earlier if there's no other problems that affect the actual functionality to avoid merge the **** over and over and over and over again

PhysSong:
@liushuyu Okay. Merging it now and editing those lines later would be better. Now it seems to be fine.

liushuyu:
Manually push my branch directly to `master`, avoid using GitHub merge buttons to prevent from further f*** up.

PhysSong:
@liushuyu I agree. Direct pushing would be better than pressing GitHub merge button.

PhysSong:
Some commits are added after your merge, so direct pushing is improper now.
@liushuyu I did a [new merge](https://github.com/PhysSong/lmms/tree/merge) based on yours. If you don't mind, @tresf use this.

tresf:
> If you don't mind, @tresf use this.

That looks very clean.  I'm leaving the merging up to you guys though, I hope that's OK.

PhysSong:
@liushuyu One error now: `pkg-config`, not `pkgconfig`.
It was correct in your last merge.

PhysSong:
I pushed https://github.com/LMMS/lmms/commit/cc3822141ee7a304f616cb0130e29f2d8b7a714f, a new merge.
@liushuyu Since it is hard to give you credit directly, I mentioned you in the commit message.

Issue #3608, opened 2017-06-04
Crash when clicking "Quantize" in a Piano-Roll with no pattern


styken:
Reproduce:
1 - open pianoroll
2 - press quantize

Texxiliator:
Nothing happens 
atleast for me

Texxiliator:
What OS are you using?

karmux:
I can reproduce.

DeRobyJ:
I cannot reproduce in Windows 8.1 or Windows 7, 64-bit

styken:
Do you have an empty roll? I only have the text "Please open a pattern...".
Then I press 'Q' to the left of the scissors.

Hussam-Eddin-Alhomsi:
The issue's title should be "Quantizing in Piano-Roll causes a crash.".

zonkmachine:
Closed in https://github.com/LMMS/lmms/pull/3609

PR #3609, opened 2017-06-04
Fixed LMMS crash when pressing Q in not existing piano roll.

karmux:
Steps to reproduce:
1. Open LMMS
2. Open Piano-Roll
3. Click on Q
4. LMMS crashes.

This change fixes it.
Link to issue: #3608.

### DIFF
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -3918,31 +3918,33 @@ int PianoRoll::quantization() const
 
 void PianoRoll::quantizeNotes()
 {
-	NoteVector notes = getSelectedNotes();
+	if( m_pattern != NULL ) {
+		NoteVector notes = getSelectedNotes();
 
-	if (notes.empty())
-	{
-		for (Note* n : m_pattern->notes())
+		if( notes.empty() )
 		{
-			notes.push_back(n);
-		}
-	}
-
-	for (Note* n : notes)
-	{
-		if (n->length() == MidiTime(0))
-		{
-			continue;
+			for( Note* n : m_pattern->notes() )
+			{
+				notes.push_back(n);
+			}
 		}
 
-		Note copy(*n);
-		m_pattern->removeNote(n);
-		copy.quantizePos(quantization());
-		m_pattern->addNote(copy);
-	}
+		for( Note* n : notes )
+		{
+			if (n->length() == MidiTime(0))
+			{
+				continue;
+			}
 
-	update();
-	gui->songEditor()->update();
+			Note copy(*n);
+			m_pattern->removeNote(n);
+			copy.quantizePos(quantization());
+			m_pattern->addNote(copy);
+		}
+
+		update();
+		gui->songEditor()->update();
+	}
 }
 
 
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -805,7 +805,7 @@ void PianoRoll::setBackgroundShade( const QColor & c )
 
 
 
-void PianoRoll::drawNoteRect( QPainter & p, int x, int y, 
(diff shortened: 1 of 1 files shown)


src/gui/editors/PianoRoll.cpp
```diff
@@ -3918,31 +3918,33 @@ int PianoRoll::quantization() const
 
 void PianoRoll::quantizeNotes()
 {
-	NoteVector notes = getSelectedNotes();
+	if( m_pattern != NULL ) {
```
michaelgregorius: While this definitively works I'd propose to use the same style like in `PianoRoll::cutSelectedNotes()`:
```C++
if( ! hasValidPattern() )
{
	return;
}
```
This make it more obvious what's tested and also keeps the indentation level of the method down.

src/gui/editors/PianoRoll.cpp
```diff
@@ -805,7 +805,7 @@ void PianoRoll::setBackgroundShade( const QColor & c )
 
 
 
-void PianoRoll::drawNoteRect( QPainter & p, int x, int y, 
+void PianoRoll::drawNoteRect( QPainter & p, int x, int y,
 				int width, const Note * n, const QColor & noteCol,
```
zonkmachine: You have one requested fix in here together with 14 lines that are formatting changes like whitespace at the end etc. This makes it a bit hard to review. I suggest for the next time to commit changes like this separately and name the commit 'fixup' or 'whitespace' or something similar.

tresf: @zonkmachine agreed.  Harder to cherry-pick too as the more lines that get modified increases the chances of a merge conflict.

karmux: I need to switch off editor's automatic cleanups :)

src/gui/editors/PianoRoll.cpp
```diff
@@ -3918,31 +3918,33 @@ int PianoRoll::quantization() const
 
 void PianoRoll::quantizeNotes()
 {
-	NoteVector notes = getSelectedNotes();
+	if( hasValidPattern() ) {
```
michaelgregorius: This is already better than before but I was also asking to use the following coding pattern:
```C++
if ( !hasValidPattern() )
{
  return;
}

NoteVector notes = getSelectedNotes();

// ...
// Everything that's inside the `if` statement follows here
```
This is an often used pattern when the scope of an `if` statement encompasses a whole method. By doing it like this you also get rid of one indentation level for the code that's inside the `if` clause which makes the code more readable.

src/gui/editors/PianoRoll.cpp
```diff
@@ -3918,33 +3918,36 @@ int PianoRoll::quantization() const
 
 void PianoRoll::quantizeNotes()
 {
-	if( hasValidPattern() ) {
-		NoteVector notes = getSelectedNotes();
+	if( ! hasValidPattern() )
```
michaelgregorius: That's what I meant. Thanks!

zonkmachine: Clean and simple. Retested, merging!

tresf:
FYI, this commit is [already to merge to master](https://github.com/liushuyu/lmms/blob/merge/src/gui/editors/PianoRoll.cpp#L3921) in `liushuyu/merge`.

src/gui/editors/PianoRoll.cpp
```diff
@@ -3918,27 +3918,32 @@ int PianoRoll::quantization() const
...
 {
+	if( ! hasValidPattern() )
+	{
+		return;
+	}
+
 	NoteVector notes = getSelectedNotes();
 
-	if (notes.empty())
+	if( notes.empty() )
```
grejppi: This is a new function, and therefore the old coding style (wrt spaces inside parentheses) should not apply. I realize this is late, but still 😅

PR #3617, opened 2017-06-06
Fix #3616 by preventing fold overs for clipped data

### LINKED ISSUE #3616: 24-bit export noise. Integer overflow.
At levels over +/- 1.0f the signal will fold over.

24bitclip

@michaelgregorius

michaelgregorius:
Use libsndfile functionality to prevent fold overs when exporting clipped data. The fold overs occurred when exporting with a bit depth of 24 bit.


tresf:
@liushuyu please add this to #3607

PR #3621, opened 2017-06-09
Some fixes to recent files

### LINKED ISSUE #528: New Project opens with old paramters
valid for loop-indicator-state & windows-states
Significance :Low

loop-indicators
In Song-editor & piano-roll:
A 'New Project' will have the loop-indicators in the same posistions, and state as the last saved project. It is more intuitive if 'New Project'loop-indicators are in the first beat, and off.

Windows-states
'New Project' open with the same windows-states as the last saved project. For instance -if all windows was closed, the 'New Project' has an empty GUI.

(cheked Main tempo, time-sign, master-vol & master-pitch -all are correctly set to default values)

zonkmachine:
The recent file system feels a bit glitchy. If you work with a saved file and then start a new project, or a new project from template that you decide to scratch, the next time you launch LMMS you will get back the previous saved work again. I find this a bit unintuitive as you've done other work in between.

---

### Fixes

 - If the last project was a template we create a new project.
 - If there is a recovery file present and you discard it we create a new project. The project launched could be defective or, if .lmmsrc.xml wasn't written, an earlier project.
 - Add a factory default data/projects/templates/default.mpt. Fixes #528

### DIFF
--- a/data/projects/templates/default.mpt
+++ b/data/projects/templates/default.mpt
@@ -1,0 +1,89 @@
+<?xml version="1.0"?>
+<!DOCTYPE lmms-project>
+<lmms-project version="1.0" creator="LMMS" creatorversion="1.2.0-rc3.15" type="song">
+  <head timesig_numerator="4" mastervol="100" timesig_denominator="4" bpm="140" masterpitch="0"/>
+  <song>
+    <trackcontainer width="600" x="5" y="5" maximized="0" height="300" visible="1" type="song" minimized="0">
+      <track muted="0" type="0" name="TripleOscillator" solo="0">
+        <instrumenttrack pan="0" fxch="0" usemasterpitch="1" pitchrange="1" pitch="0" basenote="57" vol="100">
+          <instrument name="tripleoscillator">
+            <tripleoscillator phoffset2="0" userwavefile0="" finer0="0" userwavefile1="" finer1="0" userwavefile2="" finer2="0" coarse0="0" coarse1="-12" coarse2="-24" finel0="0" finel1="0" modalgo1="2" modalgo2="2" finel2="0" pan0="0" modalgo3="2" pan1="0" stphdetun0="0" pan2="0" stphdetun1="0" wavetype0="0" stphdetun2="0" wavetype1="0" wavetype2="0" vol0="33" vol1="33" phoffset0="0" phoffset1="0" vol2="33"/>
+          </instrument>
+          <eldata fres="0.5" ftype="0" fcut="14000" fwet="0">
+            <elvol lspd_denominator="4" sustain="0.5" pdel="0" userwavefile="" dec="0.5" lamt="0" latt="0" rel="0.1" amt="0" x100="0" att="0" lpdel="0" hold="0.5" lspd_syncmode="0" lshp="0" lspd="0.1" ctlenvamt="0" lspd_numerator="4"/>
+            <elcut lspd_denominator="4" sustain="0.5" pdel="0" userwavefile="" dec="0.5" lamt="0" latt="0" rel="0.1" amt="0" x100="0" att="0" lpdel="0" hold="0.5" lspd_syncmode="0" lshp="0" lspd="0.1" ctlenvamt="0" lspd_numerator="4"/>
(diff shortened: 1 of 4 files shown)


zonkmachine:
> - if the last project was a template we create a new project. ( This PR )

Works fine. Clears the action up good. Some more testing and I'll merge this later.

zonkmachine:
> - if the last project was a template we create a new project. ( This PR )

Seem to be fixed for when you create project from a template but not when the project is the `default.mpt` .

zonkmachine:
> > - if the last project was a template we create a new project. ( This PR )
> 
> Seem to be fixed for when you create project from a template but not when the project is the default.mpt .

No. It works if you save a project as default project. `Song::createNewProject()` first tries to find `default.mpt` in your templates directory. Next it looks in [`factoryProjectsDir()`](https://github.com/LMMS/lmms/blob/stable-1.2/src/core/Song.cpp#L916:L917) and this it cant find because there seem to be no `default.mpt` among the [factory templates](https://github.com/LMMS/lmms/tree/stable-1.2/data/projects/templates), so it proceeds to generate a hard coded project and this isn't really a file so it doesn't get registered in the recent files list.

@tresf Should there be a default.mpt in the factory templates?

tresf:
> @tresf Should there be a default.mpt in the factory templates?

It certainly sounds more logical than a hard-coded project.

zonkmachine:
Works like a charm!

zonkmachine:
> - Add a factory default data/projects/templates/default.mpt

I believe this also takes care of issue https://github.com/LMMS/lmms/issues/528

src/core/ConfigManager.cpp
```diff
@@ -282,7 +282,8 @@ void ConfigManager::addRecentlyOpenedProject( const QString & file )
 {
 	QFileInfo recentFile( file );
 	if( recentFile.suffix().toLower() == "mmp" ||
-			recentFile.suffix().toLower() == "mmpz" )
+			recentFile.suffix().toLower() == "mmpz" ||
+				recentFile.suffix().toLower() == "mpt" )
```
Umcaruje: You could remove the one tab of indentation here so these two lines are at the same level

src/core/main.cpp
```diff
@@ -846,13 +846,15 @@ int main( int argc, char * * argv )
 		else if( ConfigManager::inst()->
 				value( "app", "openlastproject" ).toInt() &&
 			!ConfigManager::inst()->
-				recentlyOpenedProjects().isEmpty() )
+				recentlyOpenedProjects().isEmpty() &&
+			!recoveryFilePresent )
```
Umcaruje: And here you could add another tab

data/projects/templates/default.mpt
```diff
@@ -0,0 +1,89 @@
+<?xml version="1.0"?>
+<!DOCTYPE lmms-project>
+<lmms-project version="1.0" creator="LMMS" creatorversion="1.2.0-rc3.15" type="song">
```
Umcaruje: I wonder what we should do with the creator version, this way we'd need to update it for every version, otherwise the users would keep getting the 'Project version mismatch' tool tip on each new project, right?

src/core/ConfigManager.cpp
zonkmachine: Fixed

src/core/main.cpp
zonkmachine: I added even more for a casual look. Bohemic chic.

tresf:
Interesting point about creator version.  A `default.mpt.in` could address this problem, or we could suppress project version warnings for template files.

src/core/main.cpp
Umcaruje: Haha, it'll maybe look weird in some editors that don't have the same spaces for tabs like github does, don't you think?

zonkmachine:
I thought about suppressing warnings too but we should really keep our own made stuff up to date, making such a sing a bit superfluous. Update the projects for a new release makes most sense to me.
We should have a release protocol to follow with things like this defined.

tresf:
Cherry-picked and ready for master via 9edaa59.

PR #3630, opened 2017-06-11
Open .mpt files via the file open dialogue.

zonkmachine:
untitled

Just found myself needing this again for the tenth time or so.
I also omitted the project suffixes from translation as is done in `MainWindow::saveProjectAs()`

### DIFF
--- a/src/gui/MainWindow.cpp
+++ b/src/gui/MainWindow.cpp
@@ -854,7 +854,9 @@ void MainWindow::openProject()
 {
 	if( mayChangeProject(false) )
 	{
-		FileDialog ofd( this, tr( "Open Project" ), "", tr( "LMMS (*.mmp *.mmpz)" ) );
+		FileDialog ofd( this, tr( "Open Project" ), "",
+				tr( "LMMS Project" ) + " (*.mmpz *.mmp);;" +
+				tr( "LMMS Project Template" ) + " (*.mpt)" );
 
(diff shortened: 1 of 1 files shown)


zonkmachine:
Oops! This works horribly bad as I need to also call `createNewProjectFromTemplate()` Closing for now.

zonkmachine:
No. I works just fine. Loads and saves .mpt files. I have second thoughts about the idea though. It seem to go against the idea of a template to be able to access it and save over it that easy. It would need a warning message on trying to save over a template at least. Thoughts?

tresf:
If we mirror other applications the default behavior of opening a template is to create new, but the default behavior of saving is to overwrite the template itself.  Does  that help?

tresf:
What I mean is, double-clicking the template should create a new project.  We don't really encourage distributing templates and they aren't yet registered with the various OSs, so whichever direction we take we need a clear distinction between open and edit.

eagles051387:
I think if anything templates should be read only forcing you to save it as a new project.

zonkmachine:
Thanks for the feedback!

At the time I made this PR I was working with the default templates themselves and under those circumstances the behaviour felt really impractical. Looking at this today I think it may be a good candidate to just close. I may go over the case more thoroughly after 1.2 is released.

zonkmachine:
Closing for now.

PR #3643, opened 2017-06-17
Add tooltip for refresh button in FileBrowser

jack126guy:
It seems that some users (on Discord) don't know what this button does, and I noticed that it doesn't have a tooltip. This patch should help at least a bit.


zonkmachine:
@jack126guy Thanks for the PR. The code around this has changed in master where this button has been moved to the top. Can you please rebase this PR against master instead?

jack126guy:
Done. It seems appropriate to change the base of this PR to `master`, so I've done that.

zonkmachine:
Thanks, merging!

Issue #3652, opened 2017-06-21
Text missing on the envelope tabs with Classic theme.


zonkmachine:
Since 53772c052ba8d0de1bb82b22249895e4737cfdf6 #3569, `Allow WidgetTab to use artwork tabs (@StCyr)`, some text is missing with the Classic Theme. I've only noticed this on the instrument plugins envelope tab as shown below. 'GENERAL SETTINGS' and the row above the envelopes beginning with 'TARGET' is missing. @Umcaruje 

![classic](https://user-images.githubusercontent.com/6368949/27405891-910b5ab4-56d3-11e7-8620-e78f847eb72a.png)

Umcaruje:
I'll work on this once my exams are over (25th)

PhysSong:
Not only these text but also other texts(for example, some texts in setting window) have been disappeared.

PhysSong:
It's the problem of style.css and easy to fix. I sent a PR to fix this.

PhysSong:
@Umcaruje did you check my PR?

Umcaruje:
@PhysSong yes but the colors need work, I wonder if I should push to your branch or make a new PR alltogether

PhysSong:
@Umcarjue Okay. I'll grant you push access.

PR #3664, opened 2017-06-24
Fix for misaligned controls icon in VeSTige plugin.

karmux:
track_op_menu* has different dimensions in classic theme and default theme. As track_op_menu* was only used in VeSTige and there's already identical theme icons under name trackop* then I decided to make it part of VeSTige. Just resizing or moving icons don't fix default theme while keeping classic theme as it is.

VeSTige in classic theme:

screenshot_20170624_213629

VeSTige in default theme before fix:

screenshot_20170624_213521

VeSTige in default theme after fix:

screenshot_20170624_230437


PhysSong:
Maybe related to #3652?

plugins/VstEffect/VstEffectControlDialog.cpp
```diff
@@ -88,18 +88,18 @@ VstEffectControlDialog::VstEffectControlDialog( VstEffectControls * _ctl ) :
 		m_managePluginButton->setCheckable( false );
 		m_managePluginButton->setCursor( Qt::PointingHandCursor );
 		m_managePluginButton->setActiveGraphic( PLUGIN_NAME::getIconPixmap(
-							"track_op_menu" ) );
+							"controls_active" ) );
 		m_managePluginButton->setInactiveGraphic( PLUGIN_NAME::getIconPixmap(
-							"track_op_menu" ) );
+							"controls" ) );
```
PhysSong: @karmux You should add two renamed images in `BUILD_PLUGIN()`, in `plugins/VstEffect/CMakeLists.txt`. I think changing `logo.png` to `*.png` will be fine.

PhysSong: The problem was only in `master`. Now fixed via 4ca6cdb.

PR #3667, opened 2017-06-26
Fix scripts

zapashcanon:
Hi,

See this [comment](https://github.com/LMMS/lmms/pull/3476#issuecomment-310925442).

### DIFF
--- a/.travis/script.sh
+++ b/.travis/script.sh
@@ -4,6 +4,9 @@ set -e
 
 if [ "$TYPE" = 'style' ]; then
 
+	# SC2185 is disabled because of: https://github.com/koalaman/shellcheck/issues/942
+	# once it's fixed, it should be enabled again
+	# shellcheck disable=SC2185
 	# shellcheck disable=SC2046
 	shellcheck $(find -O3 "$TRAVIS_BUILD_DIR/.travis/" "$TRAVIS_BUILD_DIR/cmake/" -type f -name '*.sh' -o -name "*.sh.in")
 
--- a/cmake/msys/msys_helper.sh
+++ b/cmake/msys/msys_helper.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -eu
+
 # Git repo information
 fork="lmms"	# i.e. "lmms" or "tobydox"
 branch="master"	# i.e. "master" or "stable-1.2"
@@ -15,7 +17,6 @@ function warn() { echo -e "\n${yellow}$1${plain}"; }
 function err() { echo -e "\n${red}$1${plain}"; exit 1;}
 
 info "Checking for mingw environment"
-
 if ! env | grep MINGW; then
 	err "  - Failed. Please relaunch using MinGW shell"
 fi
@@ -31,20 +32,18 @@ else
 	git clone -b $branch https://github.com/$fork/lmms.git
 fi
 
-
 info "Fetching ppa using cmake/msys/fetch_ppas.sh..."
 if [ -d "$HOME/ppa" ]; then
         warn "  - Skipping, $HOME/ppa already exists"
 else
-	./lmms/cmake/msys/fetch_ppa.sh
+	lmms/cmake/msys/fetch_ppa.sh
 fi
-
 
 info "Extracting debs to $HOME/ppa/opt/, etc..."
 if [ -d "$HOME/ppa/opt" ]; then
         warn "  - Skipping, $HOME/ppa/opt already exists"
 else
-        ./lmms/cmake/msys/extract_debs.sh
+        lmms/cmake/msys/extract_debs.sh
 fi
 
 info "Preparing library merge, making all qt headers writable..."
@@ -52,7 +51,6 @@ chmod u+w /mingw64/include/qt4 -R
 chmod u+w /mingw32/include/qt4 -R
 
 info "Merging mingw headers and libraries from ppa over existing system libraries..."
-
 if ! find /mingw64 | grep sndfile.h; then
 	command cp -r "$HOME/ppa/opt/mingw"* /
 else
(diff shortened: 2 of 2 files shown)


PhysSong:
Great! Tracis CI build is now passing.

cmake/msys/msys_helper.sh
```diff
@@ -222,9 +212,8 @@ if [ ! -e "$mingw_root/lib/libgig/libgig.dll.a" ]; then
 	make install
 
 	mv "$mingw_root/lib/bin/libakai-0.dll" "$mingw_root/bin"
-	mv "$mingw_root/lib/bin/libgig-7.dll" "$mingw_root/bin"
 
-	if [ $? -ne 0 ]; then
+	if ! mv "$mingw_root/lib/bin/libgig-7.dll" "$mingw_root/bin"; then
```
tresf: This is where `shellcheck` shows it's value.  If we need both DLLs, we'll need to test for both.

PhysSong: I agree. This problem also existed before, and we need to fix it if needed.

PhysSong:
```bash
if ! ( mv "$mingw_root/lib/bin/libakai-0.dll" "$mingw_root/bin" && mv "$mingw_root/lib/bin/libgig-7.dll" "$mingw_root/bin" );then
```
@tresf Would it be better?

zapashcanon: There's something I don't understand. Why don't we (I) add `set -e` on the previous PR on this script ? That would allow us to remove almost all these tests...

PhysSong: @zapashcanon I think so.
If you add `set -e`, however, printing custom error message will be unable. Is there alternative method? Or, is it okay to disable printing custom error message?

tresf: Sure... we're moving to`set -e` for everything else. :+1:.  We've sort of repurposed a fetch/untar script as a build helper so this file is growing in complexity which can affect how it should respond to errors.

Ideally we'd just move away from this technique and leverage something crowdsourced like `scoop` instead, but we're not there yet.

zapashcanon: I moved to `set -e` in a new commit. There's no need for custom error message IMO: we print a lot of info about what we're doing, so the user will know what failed, plus, the shell will print something too. What do you think ?

PhysSong:
@tresf I suggest you to merge ASAP if there's no problem. Some people might think Travis CI is failing because of Critical bug(s). By merging this, We can prevent such misunderstandings.

tresf:
The message to relaunch in mine shell is there for a reasons since this can be unobvious to the user.  Please being it back as well as red error message.

tresf:
*mingw shell

zapashcanon:
> as well as red error message

Are you just talking about the relauch in mingw shell or every error message ?

I think most of the error message are useless, e.g.:

```sh
info "Downloading and building fltk $fltkver"

 if ! which fluid; then
    wget http://fltk.org/pub/fltk/$fltkver/fltk-$fltkver-source.tar.gz -O "$HOME/fltk-source.tar.gz"
```

If `wget fails`, no need to do: `err "ERROR: Could not download fltk.  Exiting."` IMO, we already told the user what we're doing, if it fails, it will be clear.

If you disagree, so, what you want is just my first commit right ? I can just remove the second commit... But I think we should keep it.

tresf:
> I think most of the error message are useless [...] If wget fails, no need to do: `err` ...

Agreed, just the mingw one, it's not obvious at all, but to that point we may have others that aren't obvious down the road.  `set -e` is a convenience, but we have to keep in the ability to issue a human warning when stdout is inadequate which from my experience happens about 10% of the time.

zapashcanon:
> we have to keep in the ability to issue a human warning (or error) when stderr is inadequate

We probably could do something with `trap`.

> Agreed, just the mingw one

I'll fix it in the second commit then.

PhysSong:
@zapashcanon @tresf Now it seems to be ready for merge. Is it needs any more changes?

tresf:
Were the indents addressed?  I understand there is a need for master to clear Travis-CI, but "ready for merge" is subjective.

cmake/msys/msys_helper.sh
```diff
@@ -207,26 +164,18 @@ info "Downloading and building libgig $gigver"
...
 	info "  - Installing libgig..."
 	make install
 
 	mv "$mingw_root/lib/bin/libakai-0.dll" "$mingw_root/bin"
-	mv "$mingw_root/lib/bin/libgig-7.dll" "$mingw_root/bin"
-
-	if [ $? -ne 0 ]; then
-        	err "ERROR: Could not build/install libgig -- gigplayer needs this.  Exiting."
-	fi
+  mv "$mingw_root/lib/bin/libgig-7.dll" "$mingw_root/bin"
```
PhysSong: Needs indentation?

.travis/script.sh
```diff
@@ -4,6 +4,9 @@ set -e
 
 if [ "$TYPE" = 'style' ]; then
 
+  # SC2185 is disabled because of: https://github.com/koalaman/shellcheck/issues/942
+  # once it's fixed, it should be enabled again
+  # shellcheck disable=SC2185
```
PhysSong: Here three lines, too.

zapashcanon:
@tresf, I didn't noticed the indent, fixed now.

tresf:
Thanks for the fixes!  Merging.  If we encounter new issues we'll submit them as new PRs/patches.

tresf:
@zapashcanon the merge button gave credit to me do to the way the commits were added.  Apologies.

zonkmachine:
> @zapashcanon the merge button gave credit to me do to the way the commits were added

`git commit --amend --author="Author Name <email@address.com>"`

zonkmachine:
@tresf No. He's credited. This was in master and our default is stable-1.2 so once we switch to master these commits will be visible. If I'm even on topic here...

tresf:
@zonkmachine you're right. 👍 

https://github.com/LMMS/lmms/commits/master

zapashcanon:
>  Merging.

Thanks.

> If we encounter new issues we'll submit them as new PRs/patches.

I agree on this.

PR #3678, opened 2017-07-03
MIDI - Don't apply base velocity to all controller values

zonkmachine:
I noticed the sustain pedal was affected by setting the base velocity. Turn down the velocity under MIDI settings and eventually the threshold for value 'on' (64) is no longer reached. I've moved that part of the code in `MidiPort::processInEvent()` to within the part handling note on/off/keypressure.
I haven't tried this with other control messages than sustain yet.

### DIFF
--- a/src/core/midi/MidiPort.cpp
+++ b/src/core/midi/MidiPort.cpp
@@ -129,11 +129,11 @@ void MidiPort::processInEvent( const MidiEvent& event, const MidiTime& time )
 			{
 				return;
 			}
-		}
 
-		if( fixedInputVelocity() >= 0 && inEvent.velocity() > 0 )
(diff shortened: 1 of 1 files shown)


zonkmachine:
I'll merge this tomorrow if there are no objections.

[After touch](https://github.com/zonkmachine/lmms/blob/62438a1bb94597fe793ab7a5409b9b4e5612a573/src/core/midi/MidiPort.cpp#L126) messages are filtered together with key on/off and the pressure level is set to the velocity level. This behaviour looks a bit suboptimal if you ask me. I think this probably needs some more design and is better handled in master. In any case this is not a new behaviour for 1.2.0 .

PR #3691, opened 2017-07-10
Fix #3674: Pattern names difficult to read

### LINKED ISSUE #3674: Pattern names difficult to read
blocknamesrc3

Difference between notes and letters are not sufficient, if they overlap -Maby more dark in the letters? (mouse over tooltips still shows 'open in piano-roll')

michaelgregorius:
Make the pattern names better readable by rendering them on top of a semitransparent black rectangle. Elide the text and make it render a bit more to the right.

Use the font properties that are defined in the CSS to draw the pattern labels. This provides flexibility with regards to the font properties that are used (size, font family, etc.).

Adjust the CSS for the default theme and the classic theme.

### DIFF
--- a/data/themes/classic/style.css
+++ b/data/themes/classic/style.css
@@ -596,6 +596,7 @@ TrackContentObjectView   {
 PatternView {
 	background-color: rgb( 119, 199, 216 );
 	color: rgb( 187, 227, 236 );
+	font-size: 11px;
 }
 
 /* sample track pattern */
--- a/data/themes/default/style.css
+++ b/data/themes/default/style.css
@@ -615,6 +615,7 @@ TrackContentObjectView   {
 PatternView {
 	background-color: #21A14F;
 	color: rgba(255,255,255,220);
+	font-size: 11px;
 }
 
 /* sample track pattern */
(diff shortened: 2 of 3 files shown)


Umcaruje:
This change has brought an UI inconsistency with the other patterns:
screenshot from 2017-07-15 23 23 19

I vote the same rectangle background gets implemented in other TCO's. Also for theming purposes, the rectangle color should be available through CSS, so the opacity can be freely changed.

Sorry for reviewing the PR this late, I never got around to it.

michaelgregorius:
@Umcaruje Solved with #3704. For now I have only added the method `TrackContentObjectView::paintTextLabel` which is then called in the paint code of all inheriting classes. However, in the long run it might make sense to move the painting code that's shared by all pattern types to `TrackContentObjectView` and to only keep the specialized logic in the inherited pattern views.

Umcaruje:
>However, in the long run it might make sense to move the painting code that's shared by all pattern types to TrackContentObjectView and to only keep the specialized logic in the inherited pattern views

Agree on that 100%, checking out the PR now.

michaelgregorius:
Inconsistencies have been fixed with the merge of #3704.

Issue #3698, opened 2017-07-15
Melody patterns do not make good use of height when enlarged


michaelgregorius:
The current implementation draws the notes of melody patterns as lines regardless of the pattern's size:
![patternlineslargeold](https://user-images.githubusercontent.com/9293269/28239007-2e27383c-6961-11e7-81b0-e6148a547885.png)

Also the notes are always painted underneath the text label even if the pattern is enlarged.

michaelgregorius:
Fixed with #3699. The pattern above looks as follows at larger heights:

![3698 - pattern new](https://user-images.githubusercontent.com/9293269/28239100-d2c7b068-6963-11e7-98f3-caae901d9b18.png)

Please note that at larger heights the notes are moved completely below the text label. For patterns with a small height like in the second track the notes are drawn underneath the text label to make best use of the constrained space. #3699 also has an animated GIF that shows the dynamic aspects when the pattern is resized.

michaelgregorius:
Here's a variation where the black keys are indicated in the pattern as well (only drawn when the notes are drawn as rectangles):

![3698 - variant with black key indicators](https://user-images.githubusercontent.com/9293269/28239301-a0037d6e-6969-11e7-9d13-1e88b5973d06.png)

BaraMGB:
Nice idea!

Mark-Agent003:
It seems like the notes gain more opacity the bigger they get. Was that intended?

michaelgregorius:
@Mark-Agent003 I think what you observe is caused by the fact that for the notes in the third pattern you see more of their darker border than their inner fill. The code does not change the color or opacity depending on the note height.

Umcaruje:
@michaelgregorius I find the black keys idea would just add unneeded complexity, and can't figure out a valid usecase for it. I like what you've done with the drawing, but I think the rectangles shouldn't have any border, just be plain single-colored, it will look sharper and nicer

michaelgregorius:
@Umcaruje I agree that showing the black keys rather adds clutter than useful information and that it might make the song editor very busy.

Regarding the note borders: my first implementation had notes without any borders. However, for some patterns the borders really help to convey a bit more information about whats happening in the pattern. Here's a pattern without borders:

![3698 - notes without border](https://user-images.githubusercontent.com/9293269/28246503-81d4b30a-6a1b-11e7-9228-ea086598a2c4.png)

On first sight it looks like the pattern starts with one long note. Here's the same pattern with borders:

[...]

PR #3699, opened 2017-07-15
Draw notes as rectangles and only move them below the text label at small heights

michaelgregorius:
Draw the notes as antialiased rectangles with a slightly darker border if the height of the pattern is larger than 64 pixels. Otherwise draw them as lines (as before).

Move the notes underneath the text label at small sizes but otherwise draw them below to enhance readability.

Always draw at least an octave of notes onto the pattern, so that patterns with just a single note will not look strange, i.e. make sure that a patterns is not completely filled with a single note. If there is additional space to fill it will be rather filled at the top of the pattern so that single notes have a high chance of being drawn below the pattern text.

Add a border of four pixels at the top and bottom of the area where the notes are drawn so they don't get pushed under the pattern borders.

Set the transparency of the text label to 100.

Technical details:
Transform the `QPainter` so that notes drawn into a [0,1] x [0,1] coordinate system are drawn at the correct position on the pattern widget. 

Other fixes:
* Remove some redundant checks whether the pattern holds notes and decrease the indentation depth by two tabs for most of the code.
* Simplify the drawing of the inner and outer border.
* Use `numeric_limits` to initialize the minimum and maximum key.
* Remove an unnecessary assignment to a pixmap
* Move code closer to where it is needed
* Remove an unnecessary call to `QPainter::end()`

### DIFF
--- a/src/tracks/Pattern.cpp
+++ b/src/tracks/Pattern.cpp
@@ -879,7 +879,7 @@ void PatternView::paintEvent( QPaintEvent * )
 
 	QLinearGradient lingrad( 0, 0, 0, height() );
 	QColor c;
-	bool muted = m_pat->getTrack()->isMuted() || m_pat->isMuted();
+	bool const muted = m_pat->getTrack()->isMuted() || m_pat->isMuted();
 	bool current = gui->pianoRoll()->currentPattern() == m_pat;
 	bool beatPattern = m_pat->m_patternType == Pattern::BeatPattern;
 
@@ -901,11 +901,10 @@ void PatternView::paintEvent( QPaintEvent * )
 		p.fillRect( rect(), c );
 	}
 
-	const float ppt = fixedTCOs() ?
-			( parentWidget()->width() - 2 * TCO_BORDER_WIDTH )
-					/ (float) m_pat->length().getTact() :
-				( width() - 2 * TCO_BORDER_WIDTH )
-					/ (float) m_pat->length().getTact();
+	// Compute pixels per tact
+	const int baseWidth = fixedTCOs() ? parentWidget()->width() : width();
+	const float pixelsPerTact = ( baseWidth - 2 * TCO_BORDER_WIDTH ) / (float) m_pat->length().getTact();
+	const float pixelsPerTick = pixelsPerTact / MidiTime::ticksPerTact();
 
 	const int x_base = TCO_BORDER_WIDTH;
 
@@ -924,18 +923,18 @@ void PatternView::paintEvent( QPaintEvent * )
 			int min_key = 9999999;
 			int total_notes = 0;
 
-			for( NoteVector::Iterator it = m_pat->m_notes.begin();
-					it != m_pat->m_notes.end(); ++it )
+			for (Note const * note : m_pat->m_notes)
 			{
-				max_key = qMax( max_key, ( *it )->key() );
-				min_key = qMin( min_key, ( *it )->key() );
-				central_key += ( *it )->key();
+				int const key = note->key();
+				max_key = qMax( max_key, key );
+				min_key = qMin( min_key, key );
+				central_key += key;
 				++total_notes;
 			}
 
 			if( total_notes > 0 )
 			{
-				central_key = central_key / total_notes;
+				central_key /= total_notes;
 				const int keyrange = qMax( qMax( max_key - central_key, central_key - min_key ), 1 );
 
 				// debug code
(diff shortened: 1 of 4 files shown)


michaelgregorius:
Here's an animated GIF that shows the dynamic aspect of moving notes underneath the text label and also the switching from lines to rectangles (click on the image to see it at its original size):
lmms-pattern-new

Fixes #3698.

src/tracks/Pattern.cpp
```diff
@@ -870,23 +875,24 @@ void PatternView::paintEvent( QPaintEvent * )
 
 	setNeedsUpdate( false );
 
-	m_paintPixmap = m_paintPixmap.isNull() == true || m_paintPixmap.size() != size()
-		? QPixmap( size() ) : m_paintPixmap;
+	if (m_paintPixmap.isNull() || m_paintPixmap.size() != size())
```
Umcaruje: This is a much cleaner than the previous approach. We do this logic for all the other TCO's, can you update that logic too?

michaelgregorius: @Umcaruje Done with commit b7b2204cbbd2f8247a50ed8becb3e3e07ff876fd.

PhysSong:
It doesn't work properly in my build, which uses Qt 5.5.1 and runs on Ubuntu 16.04. I'm trying to figure out what goes wrong.

michaelgregorius:
@PhysSong You are right. With Qt5 there are some problems. Here's how it looks on my machine:

3698 - qt5 bug

I will also troubleshoot when I find the time.

michaelgregorius:
@PhysSong The problem seems to be related to the pen size. With Qt 4 it seems that the pen width is preserved if you define a transform whereas with Qt 5 it does not seem to be preserved. If I add the following code before painting the note pattern it works:
```C++
QPen pen = p.pen();
pen.setWidth(0);
p.setPen(pen);
```

I have just learned that a `QPen` with a width of 0 is a cosmetic pen which always draws with a border width of 1. I think this might be a nice solution anyway and it might counter the effect that @Mark-Agent003 has observed [here](https://github.com/LMMS/lmms/issues/3698#issuecomment-315538265).

michaelgregorius:
@PhysSong The problem should be fixed with commit 4021db2468bc1e4790212c2c8906438e67999479. Can you please also check on your machine? Thanks for testing this!

PhysSong:
@michaelgregorius Now works fine.

src/tracks/Pattern.cpp
```diff
@@ -870,23 +875,24 @@ void PatternView::paintEvent( QPaintEvent * )
...
-	QLinearGradient lingrad( 0, 0, 0, height() );
-	QColor c;
-	bool muted = m_pat->getTrack()->isMuted() || m_pat->isMuted();
+	bool const muted = m_pat->getTrack()->isMuted() || m_pat->isMuted();
 	bool current = gui->pianoRoll()->currentPattern() == m_pat;
 	bool beatPattern = m_pat->m_patternType == Pattern::BeatPattern;
 
 	// state: selected, normal, beat pattern, muted
-	c = isSelected() ? selectedColor() : ( ( !muted && !beatPattern )
+	QColor c = isSelected() ? selectedColor() : ( ( !muted && !beatPattern )
```
Umcaruje: Sorry for noticing just now, this would also be a good change in the other files, as well as the `QLinearGradient` one down below

src/tracks/Pattern.cpp
```diff
@@ -1096,17 +1151,18 @@ void PatternView::paintEvent( QPaintEvent * )
...
+		// inner border
 		p.setPen( c.lighter( current ? 160 : 130 ) );
 		p.drawRect( 1, 1, rect().right() - TCO_BORDER_WIDTH,
 			rect().bottom() - TCO_BORDER_WIDTH );
 
-	// outer border
-	p.setPen( ( current && !beatPattern ) ? c.lighter( 130 ) : c.darker( 300 ) );
-	p.drawRect( 0, 0, rect().right(), rect().bottom() );
+		// outer border
+		p.setPen( current ? c.lighter( 130 ) : c.darker( 300 ) );
```
Umcaruje: nice catch here, I missed that when working on this.

PhysSong:
@michaelgregorius Please merge it if there's no objection.
@Umcaruje Do you think it is fine to merge?

tresf:
From a styling perspective are there any objections to the boldness of the notes now?  I'm not opposing, but the thin notes care a much lighter weight versus their expanded counterparts.

src/tracks/Pattern.cpp
michaelgregorius: @Umcaruje I think in the long run it's better to refactor the drawing code of the TCO view implementations so that we do not have copies of that code in the first place. See also [my comment here](https://github.com/LMMS/lmms/pull/3691#issuecomment-315613020).

PhysSong:
@michaelgregorius @Umcaruje Merge?

Umcaruje:
Please do not rush things. I wouldn't go for a merge yet we're still
discussing the design

PhysSong:
@Umcaruje Okay. I was asking your opinion.
Please merge it when there's nothing th discuss anymore.

src/tracks/Pattern.cpp
Umcaruje: Agreed.

michaelgregorius:
If no one objects I will merge this one soon.

The last commit exposes the note border and note fill color (for the muted and non-muted case) to the style sheets.

Issue #2633, opened 2016-03-04
Crash when closing


tresf:
May only happen with DummyAudio, but no user-way to change it, since closing it crashes and the `~.lmmsrc.xml` file never gets written, so I consider this critical as it potentially makes master branch unusable on Mac to the average user.
- The crash occurs around here: [`Mixer.cpp#L214:L228`](https://github.com/LMMS/lmms/blob/master/src/core/Mixer.cpp#L214:L228)
- Seems to be stuck here: [`AudioDummy.h#L88`](https://github.com/LMMS/lmms/blob/master/include/AudioDummy.h#L88)

~~I haven't provided a stack trace because I feel the OS provided one has enough information to investigate this.~~

**Edit:** Switching audio device to `PortAudio/CoreAudio` causes the crash to occur in a different place with a different backtrace, so this may not be specifically related to `stopProcessing` but rather to some `free()` memory call somewhere.  A full backtrace is most likely needed here for proper investigation.

Apple provided crash report:

```
Crashed Thread:        8  QThread

Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Application Specific Information:
abort() called
*** error for object 0x7f95c2ea5510: pointer being freed was not allocated


Thread 0:: Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib          0x00007fff8c9cddb6 __psynch_cvwait + 10
1   libsystem_pthread.dylib         0x00007fff8c869728 _pthread_cond_wait + 767
2   QtCore                          0x00000001047d740f QWaitConditionPrivate::wait_relative(unsigned long) + 287
3   QtCore                          0x00000001047d724b QWaitConditionPrivate::wait(unsigned long) + 43
4   QtCore                          0x00000001047d7122 QWaitCondition::wait(QMutex*, unsigned long) + 162
5   QtCore                          0x00000001047d68ee QThread::wait(unsigned long) + 110
6   io.lmms                         0x000000010358aea6 AudioDummy::stopProcessing() + 38
7   io.lmms     [...]
```

tresf:
Came across this comment while working on another bug.  May be related to the crash:  https://github.com/LMMS/lmms/pull/2339#issuecomment-139929668

Lukas-W:
[...]

PR #3700, opened 2017-07-15
Calling delete[] without new[]

zonkmachine:
Possible fix for https://github.com/LMMS/lmms/issues/2633
Crashing on close with DummyAudio

A [comment](https://github.com/LMMS/lmms/issues/2633#issuecomment-195655709) from @Lukas-W points to the crash taking place on [deleting a pointer](https://github.com/LMMS/lmms/blob/master/include/AudioDummy.h#L104) to an array. The latter wasn't created with new[] and that's supposedly a no-no. I've tested this but since I couldn't replicate the crash itself I can't verify the fix.

### DIFF
--- a/include/AudioDummy.h
+++ b/include/AudioDummy.h
@@ -93,7 +93,8 @@ private:
 		while( true )
 		{
 			timer.reset();
-			const surroundSampleFrame* b = mixer()->nextBuffer();
+			const surroundSampleFrame* b = new surroundSampleFrame[ mixer()->framesPerPeriod() ];
+			b = mixer()->nextBuffer();
 			if( !b )
 			{
 				break;


Lukas-W:
@zonkmachine I think the crash I was referring to has already been fixed in 293926a1dd4c8cb5e84ab1d22f3026147fd5d6b1.

Lukas-W:
@PhysSong That's probably what @zonkmachine meant to write, yeah.

zonkmachine:
Right. I knew I was treading into foreign ground... Closing. I'll update #2633 with some more modest finds.

PhysSong:
@Lukas-W Can you reproduce the bug now?

Lukas-W:
@PhysSong Nope.

Issue #2112, opened 2015-06-17
Bar:Beat:Tick / Min:Sec:Msec timer values disagree with playhead


Wallacoloo:
Steps to reproduce (I recommend dialing the tempo down to ~20 bpm to make things easier to see):
1. Open LMMS and create an empty pattern on the tripleosc track.
2. Open said pattern in piano roll, hit play.
3. Hit pause in the piano roll.
4. In the song editor, move the playhead to some non-zero position.
5. Hit play in the song editor.

The Bar:Beat:Tick counter should reset and begin counting from 1:1:0 (or 0:0:000 if M:s:ms), however the playhead will _not_ reset, and they will thus be in a disagreement over what time it is. 

If notes are added to the pattern, you will hear them play according to where the playhead is, and not what the timer says.

A gif for your viewing pleasure:

![lmms-timer](https://cloud.githubusercontent.com/assets/1210751/8199374/dfddecb6-14a3-11e5-964d-d14f4d71c2ba.gif)

tresf:
Related?  #1673

Wallacoloo:
@tresf it very well might, though I did have a bit of trouble following that report's description. I could not reproduce it by following the steps in the description.

However, @badosu's interpretation of the bug report does still seem to apply.

Although not shown in the gif, when I hit play in the song editor, the piano roll's playhead does reset to 0. It may be that the timer is still tracking the piano roll's playhead at this point and that's why the reset occurs. However, the piano roll's playhead remains at 0 as the song is playing.

Wallacoloo:
So I did the first 4 steps and then used gdb to put a watch on the Song's `m_elapsedMilliSeconds` variable, which drives the time lcds. 

I got this:
```
(gdb) c
Continuing.
Hardware watchpoint 2: *(double*)0xa22958

Old value = 1392
New value = 0
0x00000000004efe2c in Song::stop (this=0xa22450) at /home/colin/proj/lmms/src/core/Song.cpp:626
626			m_elapsedMilliSeconds = 0;
(gdb) bt
#0  0x00000000004efe2c in Song::stop (this=0xa22450) at /home/colin/proj/lmms/src/core/Song.cpp:626
#1  0x00000000004efea0 in Song::playSong (this=0xa22450) at /home/colin/proj/lmms/src/core/Song.cpp:442
[...]

PR #3701, opened 2017-07-15
Remove explicit time calculations from Song and TimeLineWidget

michaelgregorius:
Removes some explicit time calculations from `Song` and `TimeLineWidget` as part of #3284. See [this comment](https://github.com/LMMS/lmms/issues/3284#issuecomment-315542540).

### DIFF
--- a/include/Song.h
+++ b/include/Song.h
@@ -101,14 +101,22 @@ public:
 	{
 		return m_nLoadingTrack;
 	}
+
 	inline int getMilliseconds() const
 	{
 		return m_elapsedMilliSeconds;
 	}
-	inline void setMilliSeconds( float ellapsedMilliSeconds )
+
+	inline void setToTime( MidiTime const & midiTime )
 	{
-		m_elapsedMilliSeconds = ellapsedMilliSeconds;
+		m_elapsedMilliSeconds = midiTime.getTimeInMilliseconds(getTempo());
 	}
+
+	inline void setToTimeByTicks(tick_t ticks)
+	{
+		m_elapsedMilliSeconds = MidiTime::ticksToMilliseconds(ticks, getTempo());
+	}
+
 	inline int getTacts() const
 	{
 		return currentTact();
--- a/src/core/Song.cpp
+++ b/src/core/Song.cpp
@@ -258,8 +258,7 @@ void Song::processNextBuffer()
 		if( m_playPos[m_playMode] < tl->loopBegin() ||
 					m_playPos[m_playMode] >= tl->loopEnd() )
 		{
-			m_elapsedMilliSeconds = 
-				( tl->loopBegin().getTicks() * 60 * 1000 / 48 ) / getTempo();
+			setToTime(tl->loopBegin());
 			m_playPos[m_playMode].setTicks(
 						tl->loopBegin().getTicks() );
 			emit updateSampleTracks();
@@ -316,8 +315,7 @@ void Song::processNextBuffer()
 					ticks %= ( maxTact * MidiTime::ticksPerTact() );
 
 					// wrap milli second counter
-					m_elapsedMilliSeconds = 
-						( ticks * 60 * 1000 / 48 ) / getTempo();
+					setToTimeByTicks(ticks);
 
 					m_vstSyncController.setAbsolutePosition( ticks );
 				}
@@ -335,10 +333,7 @@ void Song::processNextBuffer()
 				if( m_playPos[m_playMode] >= tl->loopEnd() )
 				{
 					m_playPos[m_playMode].setTicks( tl->loopBegin().getTicks() );
-
-					m_elapsedMilliSeconds =
-						( ( tl->loopBegin().getTicks() ) * 60 * 1000 / 48 ) /
-							getTempo();
+					setToTime(tl->loopBegin());
 				}
 				else if( m_playPos[m_playMode] == tl->loopEnd() - 1 )
 				{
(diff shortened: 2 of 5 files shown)


PhysSong:
Good. Could you also fix two more places where `* 60 * 1000 / 48` appears? They are in Song.cpp and you can use `MidiTime::ticksToMilliseconds()`.

PhysSong:
Coding conventions looks inconsistent a little bit but isn't wrong.

michaelgregorius:
@PhysSong Good catch! I had also checked the file for other occurrences (searching for "48" but somehow it did not occur to me that I can replace two more calls by using on of the methods that I had just added. 😄

I have removed two more calls and squashed the two commits into one.

PhysSong:
@michaelgregorius You don't need to squash commits manually. They can be squashed when PR is merged.

zonkmachine:
@PhysSong The main reason for squashing the commits yourself, the way I see it, is to have control of the final commit message. If someone else does it for you, which is a risk at places like this, they may just press merge and all the commit messages will be automatically stacked one after the other.

PhysSong:
@michaelgregorius I'm working on #2112. If some problems occur due to this change, I'll fix it. I suggest merging.

PhysSong:
I'm going to fix #2112. When I finish my work, I'll make a pull request against stable-1.2 and my PR needs this fix. Is it okay to rebase to stable-1.2 and change the base branch?

tresf:
> I'll make a pull request against stable-1.2 and my PR needs this fix. Is it okay to rebase to stable-1.2 and change the base branch?

If this bug is not new to 1.2, it shouldn't be targeted at 1.2.  We are trying to freeze the codebase.

michaelgregorius:
Are you sure that the fix for #2112 should be targeted for 1.2? It's not in the milestones for 1.2 and at a certain point we should really get finished with 1.2. By the way, even if this fix goes into master first we can always cherry-pick it for 1.2.

@tresf What do you think?

michaelgregorius:
@tresf Seems like we posted at the same time. :smile:

PhysSong:
Though it isn't a new bug and changes the codebase, it doesn't make something unstable or change the codebase a lot. Moreover, it is a bug fix.

tresf:
I still don't see a compelling reason to merge this into stable.  This code is called all over the place, I'd rather not change it at the RC4 state so close to release.

PhysSong:
Okay. Go to master.

include/Song.h
```diff
@@ -101,14 +101,22 @@ class EXPORT Song : public TrackContainer
 	{
 		return m_nLoadingTrack;
 	}
+
 	inline int getMilliseconds() const
 	{
 		return m_elapsedMilliSeconds;
 	}
-	inline void setMilliSeconds( float ellapsedMilliSeconds )
```
PhysSong: I prefer leaving `setMilliSeconds()` to removing, but it's not bad. We can add this again when needed.

michaelgregorius: @PhysSong Yes, we can always add it back if needed. However, for now it look like the milliseconds are rather an internal representation of the `Song` class and that other client "think" in different times, e.g. `MidiTime`.

karmux:
Time display is not changing during playing anymore with this change. Time only changes when you click somewhere on timeline but not during playing.

zonkmachine:
I see this too. I've compiled both with and without debug flags.

PhysSong:
>Time display is not changing during playing anymore with this change.

I haven't tested yet, but I will fix it if I can reproduce.

zonkmachine:
> I haven't tested yet, but I will fix it if I can reproduce.

I can't answer for @michaelgregorius but why not leave him some time to respond? I'm pretty sure he knows what's up with this one already.

PhysSong:
@zonkmachine I think @michaelgregorius can fix it. However, I haven't seen this after my work for #2112.
If @michaelgregorius can fix it soon, I'll let him do.

src/core/Song.cpp
```diff
@@ -393,9 +388,7 @@ void Song::processNextBuffer()
 		framesPlayed += framesToPlay;
 		m_playPos[m_playMode].setCurrentFrame( framesToPlay +
 								currentFrame );
-		m_elapsedMilliSeconds += 
-			( ( framesToPlay / framesPerTick ) * 60 * 1000 / 48 ) 
-				/ getTempo();
+		m_elapsedMilliSeconds += MidiTime::ticksToMilliseconds( framesToPlay / framesPerTick, getTempo());
```
PhysSong: I think this line caused bug. Possible fix is: change
`double MidiTime::ticksToMilliseconds(tick_t ticks, bpm_t beatsPerMinute)` to
`double MidiTime::ticksToMilliseconds(double ticks, bpm_t beatsPerMinute)`.

PhysSong: Or adding `MidiTime::framesToMillisecond(frame_t ticks, const float framesPerTick, bpm_t beatsPerMiute)`?

michaelgregorius:
Problem is fixed with commit bcdb5ecb5af503839070a77edb23e1731c74856e. I have decided to add a second version of `MidiTime::ticksToMilliseconds` which takes a double as the argument for the ticks.

Thanks for the hint, @PhysSong!

zonkmachine:
Tested, fixed!

The somewhat spooky thing is that I actually started a review and the action around the timeline was the only thing I got around to test and still I didn't spot it... :/

michaelgregorius:
@zonkmachine Thanks for testing! Seems to have been quite some elusive bug then. I think I only tested whether the play head moves or not.

PR #3704, opened 2017-07-16
Consistent text rendering for different pattern types and exposure of the background color to style sheets

michaelgregorius:
Render the text labels of the different types of pattern consistently:
3674-consistent-patterns

Do not render any label in case the trimmed text becomes empty, i.e. if the user enters a single space (" ") for the name this would not be rendered.

Expose the background color of the pattern text label to the style sheets.

### DIFF
--- a/data/themes/default/style.css
+++ b/data/themes/default/style.css
@@ -620,6 +620,7 @@ TrackContentObjectView   {
 	qproperty-selectedColor: #006B65;
 	qproperty-BBPatternBackground: #373d48;
 	qproperty-textColor: #fff;
+	qproperty-textBackgroundColor: rgba(0, 0, 0, 75);
 	qproperty-textShadowColor: rgb(0,0,0,200);
 	qproperty-gradient: false; /* boolean property, set true to have a gradient */
 	
--- a/data/themes/default/style.css
+++ b/data/themes/default/style.css
@@ -622,13 +622,14 @@ TrackContentObjectView   {
 	qproperty-textColor: #fff;
 	qproperty-textShadowColor: rgb(0,0,0,200);
 	qproperty-gradient: false; /* boolean property, set true to have a gradient */
+	
+	font-size: 11px;
 }
 
 /* instrument pattern */
 PatternView {
 	background-color: #21A14F;
 	color: rgba(255,255,255,220);
-	font-size: 11px;
 }
 
 /* sample track pattern */
--- a/data/themes/classic/style.css
+++ b/data/themes/classic/style.css
@@ -613,6 +613,7 @@ TrackContentObjectView   {
 	qproperty-selectedColor: rgb( 0, 125, 255 );
 	qproperty-BBPatternBackground: rgb( 80, 80, 80 );
 	qproperty-textColor: rgb( 255, 255, 255 );
+	qproperty-textBackgroundColor: rgba(0, 0, 0, 75);
 	qproperty-textShadowColor: rgb( 0, 0, 0 );
 	qproperty-gradient: true; /* boolean property, set true to have a gradient */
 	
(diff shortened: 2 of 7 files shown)


PhysSong:
Looks fine. Let's wait for @Umcaruje.

data/themes/default/style.css
```diff
@@ -620,15 +620,17 @@ TrackContentObjectView   {
 	qproperty-selectedColor: #006B65;
 	qproperty-BBPatternBackground: #373d48;
 	qproperty-textColor: #fff;
+	qproperty-textBackgroundColor: rgba(0, 0, 0, 75);
 	qproperty-textShadowColor: rgb(0,0,0,200);
```
Umcaruje: Now that we actually have a shaded background, is there even a need for a shadow for the text?

Umcaruje:
This looks good. The only problem I see is that the eliding code now won't show text when a pattern is very small. Example: 
gifrecord_2017-07-18_151425

data/themes/default/style.css
tresf: > Now that we actually have a shaded background, is there even a need for a shadow for the text?

Should be tested with custom colors.

image

michaelgregorius:
@Umcaruje What do you think about the following solution? As long as the elided text is longer than one character it is used. If the text is shorter than two characters, i.e. it only consists of "..." or becomes completely empty due to the elision, we just render the trimmed text so that anything is drawn at all:

lmms-elide-example

data/themes/default/style.css
michaelgregorius: @tresf The CSS now gives great flexibility with regards to the colors. So if the user decided to make the patterns white it would be wise to take a different color for the text background or the text color itself. I also don't think that the current text shadow implementation works well with HDPI displays in some of the examples above because it would only give a line that is one pixel wide.

@Umcaruje If we decide to drop the text shadows then I guess we could also remove the `qproperty-textShadowColor` property.

tresf: > @tresf The CSS now gives great flexibility with regards to the colors. So if the user decided to make the patterns white it would be wise to take a different color for the text background or the text color itself

I'm talking about testing the built-in functionality.  This has nothing to do with custom themes or custom CSS, this has to do with testing this feature against the color capabilities of our BB patterns.

tresf:
I don't mind the cut-off behavior of the pattern name now.  Perhaps we change the tooltip to display the full name instead of the pointless "double-click" hint.  As @budislav has previously mentioned, we shouldn't be hiding one-time-use help text inside our UI. :)

data/themes/default/style.css
michaelgregorius: @tresf Ah, ok. It looks like this:

3704-colored-bb-patterns

tresf: @michaelgregorius that looks fantabulous.

Umcaruje: I still kinda like it with the shadow now that I look at it, makes it more readable imo

Umcaruje:
@michaelgregorius I like your cut-off proposal, and I think eliding does add a bit of class and consistency, but @tresf makes a very good point on utilising the tooltip, but I think that could be done in another PR.

data/themes/default/style.css
michaelgregorius: @Umcaruje Commit 822893619e92247bfa01b8461f0afe29cccbd351 adds the ability to render shadows back. The rationale is that if a design does not want to use shadows it can just set the property `qproperty-textShadowColor` to something completely transparent. That way we are more flexible.

tresf:
> and I think eliding does add a bit of class and consistency, 

I don't disagree with this point either.  It's more consistent with the rest of the software for sure.  I use quite a few of these smaller patterns in my productions, so I do think I'll have a bunch of `S...` named patterns which makes me lean in the cut-off direction, but both implementations have merit.

data/themes/default/style.css
michaelgregorius: With shadows:

3704-colored-bb-patterns-shadowed

For my taste it looks cleaner without the shadow. It might also make sense to investigate whether it would make sense to use [QGraphicsDropShadowEffect](https://doc.qt.io/qt-4.8/qgraphicsdropshadoweffect.html) to render the shadow. But that would be something for another pull request.

michaelgregorius:
So is everybody OK with merging this one? It provides new options and with the current implementation it's even possible to completely reinstate the old style.

Issue #3685, opened 2017-07-05
NaN introduced with sample-exact fix in 1.0.91


zonkmachine:
NaN in a project file from [this comment](https://github.com/LMMS/lmms/pull/3313#issuecomment-276763420) in PR #3313 submitted by @Umcaruje 
_Edit: I edited down the original song to a minimal project._
[issue3685.mmp.zip](https://github.com/LMMS/lmms/files/1126497/issue3685.mmp.zip)

The project goes silent just as the BitInvader sounds for the first time in bar 5.
`<lmms-project version="1.0" creator="LMMS" creatorversion="1.0.0" type="song">`
It was created with lmms 1.0.0 and the project breaks in lmms-1.0.91

If you turn the volume knob on the BitInvader track or if you render the project before you hit play the bug isn't triggered. I've been trying to spot something like an undefined variable in there but have come up with nothing so far.

The bug was introduced in 23433a70b577bc67f2836c0ae98a4dc48ae40f93

```
23433a70b577bc67f2836c0ae98a4dc48ae40f93 is the first bad commit
commit 23433a70b577bc67f2836c0ae98a4dc48ae40f93
Author: Vesa <contact.diizy@nbl.fi>
Date:   Mon Jun 30 01:59:18 2014 +0300

    Sample-exact models: improve
    - Remove the redundant hasSampleExactData() function. Instead, signal lack of s.ex.data by returning a NULL in valueBuffer()
    - Cache s.ex.buffers and only update them once per period
    - Make valueBuffer() in AutomatableModel threadsafe so that it can be used for NPH's sharing the same model
    - Add sample-exactness to instrumenttrack's vol & pan knobs
```

musikBear:
There is more to this 
Try this:
* Open the FX-mixer. 
* Focus *Master*
* rightclick and select 'remove unused channels'

Not only does the project sounds different, the issue with BitInvader is also gone. Somehow effects in unused channels are still influencing the playback. These channels are for presets thats *no longer in the project* !

[issue3685_unusedChannelRemoved.zip](https://github.com/LMMS/lmms/files/1127824/issue3685_unusedChannelRemoved.zip)

zonkmachine:
[...]

Issue #3408, opened 2017-03-06
Envelope triggered NaN


zonkmachine:
Edit: Brief intro
 - We produce NaN.
 - NaN is triggered in instrument envelopes. Specifically by changing release time while playing.
 - This may only be during the first loop.
 - NaN introduced here. 9e1cdd0441bca0697f582be3b85d8e2d4c152c72
 - An uninitialized variable?

To reproduce.
 - Load test project:[envelopeNaN.mmp.zip](https://github.com/LMMS/lmms/files/825468/envelopeNaN.mmp.zip)

 - Play until loop. Sound disappears? If yes, proof. If no, reload project and try again.

---

Verbose intro

The sound went out, NaN style, when I tweaked the decay on an instrument while the song was playing. I set out to see if I could pinpoint the source and I eventually managed to find a way to reproduce it reliably enought to be able to bisect this and the bug seem to have been introduced in 9e1cdd0441bca0697f582be3b85d8e2d4c152c72

```
commit 9e1cdd0441bca0697f582be3b85d8e2d4c152c72
Author: Vesa <contact.diizy@nbl.fi>
Date:   Sun Aug 3 14:49:45 2014 +0300

    Start work on replacing/removing global locks
```

The test project is a simple melody with a bitinvader synt having the volume envelopes release knob controlled by an automation track. There is a reverb on the end for detection purposes as having a reverb seem to enormously aggravate the issue. The NaN is however introduced before the effect chain.
If you play this track the sound may go silent somewhere during the first loop but if it makes it to the second you're safe and need to reopen the project and repeat 'play' until it eventually crashes. For me it will go silent at least one time out of four. Since the issue seem to clear itself up after the first run, I suspect an uninitialised variable somewhere but so far I haven't managed to fix this. We divide with zero in the envelope which I set out to 'fix' in issue #3381 but fixing that doesn't affect this issue at all. I have only seen this with the release button and I think only on the first loop.
If you introduce an `assert( ! isnanf( some buffer ) ); ` in the input of the FX chain you will see that the NaN is detected on either:

 - The first loop cycle on a project launched from the command line (or automatically from last session).
 - Immediately on pressing play on a project launched from the gui.

This is related to https://github.com/LMMS/lmms/issues/1048 and https://github.com/LMMS/lmms/pull/3313 . I think this one should be fixed first as it's introduced first in the sound chain.

[...]

PR #3706, opened 2017-07-18
Allways remove infs/nans

### LINKED ISSUE #1048: KILLER BUG: Inifinitely Loud Silence (ILS or rather NaN)

zonkmachine:
Brief

Testing to switch the inf/nan tests that are performed on export to be 'on' allways.
This seem to work well. For some 2.5% increase in cycles most of the NaN oriented bugs can be transformed from a project cutting out all sound at a specific point or from start, to just a moderate glitch.
I currently lack real test cases to assess this.

---

Verbose

I split the original PR into two separate commits and added a new one.

2fc1802 Turn on sanitizers in the FX-Mixer. This effectively prevents the whole projects to go silent because of one channel with a bad signal.

a85b02c Turn on sanitizers in the effect chains. This prevents bad data from shutting down the channel. Basically all NaN ( Warning! Will get loud without 47ac277 )

47ac277 As of the previious commit some dangerously large signals can come through the mix.

> ... has some amazing new noise going on with this PR

Fixed. Clamp values to, for the time being, +/-1.0f . I understand introducing hard limiting like this could trigger some but I see no reason to allow overly large signals. Don't other projects perform some form of clipping?
At this level a glitch will propagate through a reverb with long decay more like a distinct tick. A slight performance hit. +~0.2%.

---

Profiling

Test Project: Skiessi - Onion
Result: ~2.5% is spent in `MixHelpers::sanitize()` and `MixHelpers::addSanitizedMultiplied()`routines.

```
$ operf ./lmms

zonkmachine@zonkmachine:~/builds/lmms/lmms/build$ opreport --callgraph
CPU: AMD64 family15h, speed 3.4e+06 MHz (estimated)
Counted CPU_CLK_UNHALTED events (CPU Clocks not Halted) with a unit mask of 0x00 (No unit mask) count 100000
samples  %        image name               symbol name
-------------------------------------------------------------------------------
433771   14.1108  caps.so                  void Clip::one_cycle<&(store_func(float*, int, float, float))>(int)
  433771   100.000  caps.so                  void Clip::one_cycle<&(store_func(float*, int, float, float))>(int) [self]
-------------------------------------------------------------------------------
390387   12.6995  libQtGui.so.4.8.6        /usr/lib/x86_64-linux-gnu/libQtGui.so.4.8.6
  390387   100.000  libQtGui.so.4.8.6        /usr/lib/x86_64-linux-gnu/libQtGui.so.4.8.6 [self]

/////

61433     1.9870  lmms                     MixHelpers::sanitize(float (*) [2], int)
  61433    100.000  lmms                     MixHelpers::sanitize(float (*) [2], int) [self]
8048      0.2603  lmms                     MixHelpers::addSanitizedMultiplied(float (*) [2], float const (*) [2], float, int)
  8048     100.000  lmms                     MixHelpers::addSanitizedMultiplied(float (*) [2], float const (*) [2], float, int) [self]
```

fixes #1048 #3685

### DIFF
--- a/src/core/EffectChain.cpp
+++ b/src/core/EffectChain.cpp
@@ -202,11 +202,8 @@ bool EffectChain::processAudioBuffer( sampleFrame * _buf, const fpp_t _frames, b
 	{
 		return false;
 	}
-	const bool exporting = Engine::getSong()->isExporting();
-	if( exporting ) // strip infs/nans if exporting
-	{
-		MixHelpers::sanitize( _buf, _frames );
-	}
+
+	MixHelpers::sanitize( _buf, _frames );
 
 	bool moreEffects = false;
 	for( EffectList::Iterator it = m_effects.begin(); it != m_effects.end(); ++it )
@@ -214,10 +211,7 @@ bool EffectChain::processAudioBuffer( sampleFrame * _buf, const fpp_t _frames, b
 		if( hasInputNoise || ( *it )->isRunning() )
 		{
 			moreEffects |= ( *it )->processAudioBuffer( _buf, _frames );
-			if( exporting ) // strip infs/nans if exporting
-			{
-				MixHelpers::sanitize( _buf, _frames );
-			}
+			MixHelpers::sanitize( _buf, _frames );
 		}
 	}
 
--- a/src/core/FxMixer.cpp
+++ b/src/core/FxMixer.cpp
@@ -117,7 +117,6 @@ void FxChannel::unmuteForSolo()
 void FxChannel::doProcessing()
 {
 	const fpp_t fpp = Engine::mixer()->framesPerPeriod();
-	const bool exporting = Engine::getSong()->isExporting();
 
 	if( m_muted == false )
 	{
@@ -140,25 +139,21 @@ void FxChannel::doProcessing()
 				if( ! volBuf && ! sendBuf ) // neither volume nor send has sample-exact data...
 				{
 					const float v = sender->m_volumeModel.value() * sendModel->value();
-					if( exporting ) { MixHelpers::addSanitizedMultiplied( m_buffer, ch_buf, v, fpp ); }
-					else { MixHelpers::addMultiplied( m_buffer, ch_buf, v, fpp ); }
+					MixHelpers::addSanitizedMultiplied( m_buffer, ch_buf, v, fpp );
 				}
 				else if( volBuf && sendBuf ) // both volume and send have sample-exact data
 				{
-					if( exporting ) { MixHelpers::addSanitizedMultipliedByBuffers( m_buffer, ch_buf, volBuf, sendBuf, fpp ); }
-					else { MixHelpers::addMultipliedByBuffers( m_buffer, ch_buf, volBuf, sendBuf, fpp ); }
+					MixHelpers::addSanitizedMultipliedByBuffers( m_buffer, ch_buf, volBuf, sendBuf, fpp );
 				}
 				else if( volBuf ) // volume has sample-exact data but send does not
 				{
 					const float v = sendModel->value();
-					if( exporting ) { MixHelpers::addSanitizedMultipliedByBuffer( m_buffer, ch_buf, v, volBuf, fpp ); }
-					else { MixHelpers::addMultipliedByBuffer( m_buffer, ch_buf, v, volBuf, fpp ); }
+					MixHelpers::addSanitizedMultipliedByBuffer( m_buffer, ch_buf, v, volBuf, fpp );
 				}
 				else // vice versa
 				{
 					const float v = sender->m_volumeModel.value();
-					if( exporting ) { MixHelpers::addSanitizedMultipliedByBuffer( m_buffer, ch_buf, v, sendBuf, fpp ); }
-					else { MixHelpers::addMultipliedByBuffer( m_buffer, ch_buf, v, sendBuf, fpp ); }
+					MixHelpers::addSanitizedMultipliedByBuffer( m_buffer, ch_buf, v, sendBuf, fpp );
 				}
 				m_hasInput = true;
 			}
--- a/src/core/MixHelpers.cpp
+++ b/src/core/MixHelpers.cpp
@@ -82,6 +82,10 @@ bool sanitize( sampleFrame * src, int frames )
 				src[f][c] = 0.0f;
 				found = true;
 			}
+			else
+			{
+				src[f][c] = qBound( -4.0f, src[f][c], 4.0f );
+			}
 		}
 	}
 	return found;


zonkmachine:
The demo project in #3408, [envelopeNaN.mmp.zip](https://github.com/LMMS/lmms/files/825468/envelopeNaN.mmp.zip) has some amazing new noise going on with this PR.

What happens is probably something like this: Some faults give you infinite values and these propagated through the sound chain will metamorphose into NaN. The NaN will make lmms go silent.
When we add the sanitizer the NaN/inf values are sorted out but some crap will get through in the form of short glitches around where the infs. would have been, because you'll have values leading up to/from the infinites. These will now pass straight through and if you have a reverb in there... Magic!!!

zonkmachine:
If you only turn on the Sanitizers in the FX-Mixer the sound will cut out on just the troubled channel and not affect the whole mix. This could be a better option.

PhysSong:
>I see no reason to allow overly large signals.

Some projects use them as a intermediate signals. If we clip them, some projects may give different sound output.

And I have a question: what if an instrument generates NaN and its output goes to master FX channel directly? If there is something like master reverb, the result will be worse.

zonkmachine:
> Some projects use them as a intermediate signals. If we clip them, some projects may give different sound output.

Yes. This is a destructive PR in that sense. You could have a switch to allow higher signals. I don't know how many projects would be affected though. It would probably be more user dependent, the ones that has the habit of driving a large signal through the FX channel.


> what if an instrument generates NaN and its output goes to master FX channel directly? If there is something like master reverb, the result will be worse.

Not worse but equally shitty I think. I've kept an eye open on the master channel but so far I haven't seen it acting up because of this in any way. And I've hit it with NaN intentionally. I'll try and come up with a more angry test... :)

tresf:
If this fixes the notorious bug where the mixer from goes silent, this comes with a lot of value.  Can we make a decision on this and merge?

zonkmachine:
> If this fixes the notorious bug where the mixer from goes silent

It does. I think it should be merged and I think perhaps 1.2 is a better target than master.

PhysSong:
A sample file may contain inf/NaN stuffs. In that case, this PR won't help to fix such issues.
I guess `PlayHandle` is also a candidate for sanitization, but I don't know how it will affect performance.

Anyway, Clipping signal at +/- 1.0f doesn't sound good for me. I think it should be some appropriate larger level(ex. 4.0f~10.0f) for intermediate signals and 1.0f(or slightly larger value) for master FX channel.

zonkmachine:
> I think it should be some appropriate larger level(ex. 4.0f~10.0f) for intermediate signal

Yes, I started out with 4.0f in the first version I pushed and in testing this 10.0f was about where the glitches would start to be too loud.

zonkmachine:
> A sample file may contain inf/NaN stuffs. 

Interesting, I didn't think of this possibility. I went looking for some audio test suites containing various defective sound files for further testing but I couldn't find any. There are functions to generate borked floats like http://en.cppreference.com/w/cpp/numeric/math/nan so I may try to make some. Qt5 has functions to generate non-finite doubles but I haven't made the switch yet.

PhysSong:
@zonkmachine I have a .wav file with NaN value. I can provide it if you want.

zonkmachine:
Yes please!

zonkmachine:
Changing base for the time being as my master branch isn't anywhere near sane.

zonkmachine:
> Anyway, Clipping signal at +/- 1.0f doesn't sound good for me. I think it should be some appropriate larger level(ex. 4.0f~10.0f) for intermediate signals

Changed to +/-4.0f

PhysSong:
[nan2.wav.zip](https://github.com/LMMS/lmms/files/1493526/nan2.wav.zip)

@zonkmachine Here is a .wav file full with NaN. I made it using hex editor.

zonkmachine:
> @zonkmachine Here is a .wav file full with NaN. I made it using hex editor.

Super! Can you fix one with infinites too? :)

As you suspected an audio file to Master will sneak past the checks. As will any sound that doesn't have an FX unit in it's path.

zonkmachine:
> As will any sound that doesn't have an FX unit in it's path.

Actually, sound via a mixer channel is catching NaN even without an FX unit. It's only direct to Master that's an issue.

unfa:
I guess no one would want to use Infs/NaNs in a creative way. One might use a very loud intermediate signal to do something crazy and break effects in creative ways with that, but invalid float samples could probably be sanitized safely, without hurting creativity.

zonkmachine:
> One might use a very loud intermediate signal to do something crazy and break effects in creative ways

But how much louder? What levels do other studios, like Ardour and Bitwig, use for clipping?
With my hardware a 10.0 clipping level is a bit too audible for my taste when the glitches come. I think the VST specification use +/-1.0 for audio signals but I don't know if that also apply to hosts and if developers really stay within those levels.

zonkmachine:
> One might use a very loud intermediate signal to do something crazy and break effects in creative ways with that, but invalid float samples could probably be sanitized safely, without hurting creativity.

@unfa Here's the thing. The NaN's will cut the sound for a channel or even a whole mix. If you remove nan's/infinites, there will instead be an audible glitch in it's place where the sound is on it's way to/from infinitely high. This needs to be limited or it will give you a wall of noise when the glitch is followed by a reverb unit. Right now the suggested level for limiting is +/-4.0 . I've tried to trigger some of the glitches we've seen with ladspa pluggins in audacity and if they have a limit it seem to be much higher than that and I think we should err on being less generous in lmms.

Have you seen any similar glitches in other daw's? Ardour?

zonkmachine:
> Anyway, Clipping signal at +/- 1.0f doesn't sound good for me. I think it should be some appropriate larger level(ex. 4.0f~10.0f) for intermediate signals and 1.0f(or slightly larger value) for master FX channel.

> Right now the suggested level for limiting is +/-4.0

As I've understood it, exporting will clip at +/-1.0 anyway so I think this should be the clipping level. I believe this is the signal level limit for VST's too.


> A sample file may contain inf/NaN stuffs. In that case, this PR won't help to fix such issues.

I think this is a separate issue and should probably be fixed in SampleBuffer.cpp

zonkmachine:
> @zonkmachine Here is a .wav file full with NaN. I made it using hex editor.

OK. I made a project with nan2.wav
1. As a Sample Track playing at the very beginning.
2. A note playing at the 5th measure.

[nanwav.mmp.zip](https://github.com/LMMS/lmms/files/1772994/nanwav.mmp.zip)

Without this fix the sound will cut out from when the nan ridden file starts sounding and you need to reload the project or bypass the reverb to get the sound back. With this fix the sound will cut out with the bad sound file playing but it will come back again as soon as the sample is done playing. So, a bit of an improvement but it still needs fixing on the sample side. SampleBuffer.cpp...

tresf:
I don't have much to offer this because I don't know the DSP code well enough to understand what this will break.

My initial concerns were that stuff like C*CLIP would have problems, but it doesn't appear to.  I'm sure this is simply a lack of understanding where sanitize is being called.  Anyway, I did two tracks with BEFORE and AFTER... In both cases, the tracks play fine.  Both tracks make heavy use of effects plugins and to the naked ear, I can't hear a difference.

There are slight visual differences, but I believe these are directly attributed to nondeterministic plugins and effects, which will create the differences regardless of this PR.

@zonkmachine I think we can merge this, but it would be nice to have a real-world track in the wild to unit-test against (rather than the one that as made intentionally to test this scenario).

screen shot 2018-04-14 at 11 07 20 am


image

zonkmachine:
> I don't have much to offer this because I don't know the DSP code well enough to understand what this will break.

We already do this sanitizing but only on export. The only difference you'd see should be:
 - No more NaN issues when not exporting.
 - No more noises from feeding delay/reverb with extreme signals.
 - Possible side effect from the forced clipping at +/-4.0 .

> I think we can merge this

I'll go about it tomorrow.

tresf:
> We already do this sanitizing but only on export. 

That makes sense.  I thought the code was already there for export.  Ok, yes, in that case we definitely need to merge this.

PhysSong:
> Possible side effect from the forced clipping at +/-4.0 .

It seems to cause clipping for one of our demo projects, `Jousboxx - BuzzerBeater`. I guess some channels in the project reach the level 5.0\~6.0 and are affected.

zonkmachine:
> `Jousboxx - BuzzerBeater`

I don't know what to look for with this one as it's clipping with or without this PR. I exported as tracks and 11 out of 56 tracks was clipping but the rendered tracks are also clipped in the render process to +/-1.0 so it needs some more probing into which ones go above +/-4.0.
Do you think we should do something with this prior to rc6. Bump up the clip level to +/-10 ?

tresf:
The team is in talks about reverting this INF/NAN sanitization.  I tested the example project in this bug report and was unable to reproduce the bug on nightly, however I'm still able to reproduce the bug with the project files from #1048.

Some arguments for removal:
* Performance.  Quoting @LostRobotMusic 
   > These hundreds of thousands of checks done every single second inside of LMMS are being done to protect the program from NaN-generating plugins.  I would first like to know if one of these plugins even exists in the first place.  If it does, I would then like to test it in other programs to see how they handle the situation.  It would be embarrassing if this whole time LMMS's infamously poor performance has been due to protections from plugins that don't even exist.
* Not-Our-Bug.  Quoting @LostRobotMusic
   > es.  We should fix the issue by fixing the source of the infinitely loud nothing, rather than trying to cover it up (and ruining performance as a result). 

So far, no measured evidence of performance loss has been provided.  This should be quantified.  Furthermore, despite both of these bug reports containing steps to reproduce, there was no attempt by @LostRobotMusic to quantify "plugins that don't even exist".  A simple re-testing of #1048 shows that this problem is still very real.

I did do a quick look for the problematic plugin.  My initial investigation shows `zita-reverb` as the likely cause.  We don't ship this plugin with LMMS, so we don't (easily) have the ability to patch it.  If someone can link a repository where this plugin is maintained, it may add merit -- or at least scope -- to the "plugins that don't even exist" claim (since patching it would support this argument by one less plugin).

tresf:
> [nan2.wav.zip](https://github.com/LMMS/lmms/files/1493526/nan2.wav.zip)
> 
> @zonkmachine Here is a .wav file full with NaN. I made it using hex editor.

Audacity playhead just freezes with this file loaded, supporting @LostRobotMusic's point about other DAWs not sanitizing.

... however, Audacity also can be recovered by removing the problematic track, whereas LMMS requires a restart.

LostRobotMusic:
> the "plugins that don't even exist" claim

> there was no attempt by @LostRobotMusic to quantify "plugins that don't even exist". 

That is because I never once made the claim that they don't exist, and implying that I did is blatantly dishonest and deeply unprofessional.  It's impossible to quantify such a claim without showing every plugin that has ever been made, which obviously cannot be done.

Here's what I said in that same message:

>  I would first like to know if one of these plugins even exists in the first place.  If it does, I would then like to test it in other programs to see how they handle the situation.

I said I would like an example of one of these plugins so it can be used in testing.  I never once made the claim that they do not exist, though I did emphasize their overwhelming rarity and the fact that invalid output is solely the fault of the plugin.  I also said that if the plugin doesn't work properly in LMMS, it won't work properly in other DAWs either, which is demonstrably true.

> ... however, Audacity also can be recovered by removing the problematic track, whereas LMMS requires a restart.

This is false, LMMS doesn't require a restart.  The issue goes away once all of the offending plugins are removed or disabled, exactly like Audacity.  The issue LMMS has to deal with is the fact that these NaNs can "infect" future plugins in the signal chain, causing all of them to output NaNs just like the original offending plugin.  It's for this reason that reloading the project file is oftentimes necessary.  Restarting the entirety of LMMS is not, in any of the situations I've experienced.

tresf:
>  It's impossible to prove such a claim without showing every plugin that has ever been made, which obviously cannot be done.

The absence of this bug was used as supporting argument for its removal, quoting (Discord)

> For the record, I've had sanitization disabled in my personal-use build for over two years now and I haven't hit a single issue.

... when combined with targeted, speculative statements of performance, these statements often become persuasive in conversation, quoting: (Discord)

> I think I was the one who originally said we need to remove the sanitation entirely. 

Responses in support of this statement, or to #7323, of which this statement takes credit for the idea of:

> NaN could be annoying, but at the same time, it's not something we should be catching
> I agree with that removing sanitation but it should be thoroughly tested, so that we don't get any regressions.
> Reducing complexity will definitely help us in the long run.
> Remove the setting; the major benefit of this PR was optimization

... 

... 

> This is false, LMMS doesn't require a restart. [...] It's for this reason that reloading the project file is oftentimes necessary

Thanks, this drastically speeds up testing.

The requested example has been provided.  I did try to test this plugin in Audacity but I couldn't recreate the problem, likely due to incorrect parameters.  Testing was taking a long time due to constants restarts of LMMS -- I understand now there was a quicker way to test -- while trying to isolate the parameters that cause the NaNs.

I won't be dedicating any more time to this myself, but I hope that narrowing down a problematic plugin can help the project decide if it's worthwhile.

With regards to professionalism (either in Disord or on the bug tracker), I believe this is a conversation for another medium, perhaps Discussions.

zonkmachine:
> narrowing down a problematic plugin can help the project decide if it's worthwhile.

Lv2, swh, Dyson Compressor. Turn release to 0.

PR #3707, opened 2017-07-19
Show the names of patterns in their tool tip

michaelgregorius:
Adjust `PatternView::update` to show the name of the pattern as the tool tip (regardless of whether it is a beat and bassline or a melody pattern).

Override the public slot `TrackContentObjectView::update` for `AutomationPatternView` and `BBTCOView`. Implement it similar to `PatternView::update`.

### DIFF
--- a/include/AutomationPatternView.h
+++ b/include/AutomationPatternView.h
@@ -44,6 +44,7 @@ public:
 public slots:
 	/// Opens this view's pattern in the global automation editor
 	void openInAutomationEditor();
+	virtual void update();
 
 
 protected slots:
--- a/include/BBTrack.h
+++ b/include/BBTrack.h
@@ -98,6 +98,8 @@ public:
 	}
 	void setColor( QColor _new_color );
 
+public slots:
+	virtual void update();
 
 protected slots:
 	void openInBBEditor();
--- a/src/gui/AutomationPatternView.cpp
+++ b/src/gui/AutomationPatternView.cpp
@@ -55,8 +55,7 @@ AutomationPatternView::AutomationPatternView( AutomationPattern * _pattern,
 
 	setAttribute( Qt::WA_OpaquePaintEvent, true );
 
-	ToolTip::add( this, tr( "double-click to open this pattern in "
-						"automation editor" ) );
+	ToolTip::add(this, m_pat->name());
 	setStyle( QApplication::style() );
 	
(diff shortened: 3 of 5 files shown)


michaelgregorius:
@tresf, @Umcaruje This one implements the tool tip changes that we discussed in #3704.

PhysSong:
@michaelgregorius How about previous tooltips? Do you think it is fine to remove them?

michaelgregorius:
@PhysSong I think we can remove the previous tool tips. There seems to be [some consensus](https://github.com/LMMS/lmms/pull/3704#issuecomment-316151630) that the help system should work differently anyway.

If some work is done on a new help system the texts can be added again. They seemed a bit strange anyway as most of them started with lower case letters.

michaelgregorius:
@Umcaruje Can you please review?

The sample track already shows the file name in the tool tip if a sample was loaded. I now wonder whether it makes sense to also give it a text label which shows the filename as it is the only one that has none:

3674-consistent-patterns

The downside would be that this label would be shown all the time as it is not possible to set a new name / label.

mikobuntu:
@michaelgregorius I agree with giving the sample track a text label to keep it consistent with the other tracks. If the user wants to remove it we have the " reset name " switch for this anyway. keeping the tooltip will do no harm in my humble opinion.

tresf:
@mikobuntu I think his point was this feature hasn't been implemented yet, so it's all or nothing.

screen shot 2017-07-20 at 11 09 52 pm

PhysSong:
And rename, too.

mikobuntu:
@tresf I see, in that case i think text labels should be implemented for the sample track. I say this because I've just noticed that the way it is now ( with only having a tooltip ) and having sample folders within sample folders on my box that the text string in the tooltip can be extremely long ( and confusing ) especially with some of my sample naming conventions.

michaelgregorius:
How about showing the file name without path in the label and showing the full path in the tool tip?

3707-sampletracklabel

PhysSong:
>How about showing the file name without path in the label and showing the full path in the tool tip?

:thumbs_up:

michaelgregorius:
Showing the file name without path in the label has been pushed with commit 379e1020abd3f3cdad5930bfd08e0dc8a81cc6d6:

3707-sampletracklabel

Issue #3537, opened 2017-05-07
MIDI Sustain ignored when envelope is on.


Umcaruje:
While doing some testing for #1700 I noticed another bug.
In non-midi-based instruments when you have the envelope on, the sustain pedal gets ignored. 

Video of the behavior (first part of the video is the piano roll recording problem):
https://youtu.be/PA0mxaMIVrc?t=33s

zonkmachine:
Confirmed. This bug is in 1.0.3 .

zonkmachine:
I'm trying to bisect this one and #1700 .

zonkmachine:
This doesn't seem connected directly to #1700 and seem to never have worked. I've gone back to around 0.4.15 .

zonkmachine:
For some reason the 'close issue on merge task' didn't work for this one.
Fixed in #3730

tresf:
> For some reason the 'close issue on merge task' didn't work for this one.
> Fixed in #3730

Because the PR wasn't against the [default branch](https://help.github.com/articles/closing-issues-using-keywords/).

zonkmachine:
Ah, difawlt brawnch...

PR #3710, opened 2017-07-20
fixed recording of sustained midi notes

### LINKED ISSUE #1700: sustain pedal not being recorded in piano roll
Windows 8.1 64bit
LMMS 1.1.0 64bit (and 32bit)

I've seen 2 other issues noted that were fixed earlier last year on this but I'm not able to record the sustain.  I assume my issue is the same and the MIDI data sent by my midi keyboard is not something LMMS recognizes.

I've logged the midi here:

 TIMESTAMP IN PORT STATUS DATA1 DATA2 CHAN NOTE EVENT  
 0005C341   2   2     B0    40    7F    1  ---  CC: Pedal (Sustain)  
 0005C436   2   2     B0    40    00    1  ---  CC: Pedal (Sustain)   

Thanks for any help you can provide.

serdnab:
Fixes #1700.
This recovers the behavior done by 4b340f7d5f10297bd3f7198bb86877823fe89ede, and later lost.
When a midinoteff comes from midi, and sustain pedal is pressed, the noteplayhandle pointer don't get nullified, so later, when sustain pedal is released, it can catch that pointer to emit the signal that recording of note is finished.

### DIFF
--- a/src/tracks/InstrumentTrack.cpp
+++ b/src/tracks/InstrumentTrack.cpp
@@ -309,25 +309,22 @@ void InstrumentTrack::processInEvent( const MidiEvent& event, const MidiTime& ti
 				{
 					m_sustainPedalPressed = true;
 				}
-				else
+				else if (isSustainPedalPressed())
 				{
-					if (isSustainPedalPressed())
+					for (NotePlayHandle*& nph : m_notes)
 					{
-						for (NotePlayHandle*& nph : m_notes)
+						if (nph && nph->isReleased())
 						{
-							if (nph && nph->isReleased())
+							if( nph->origin() ==
+								nph->OriginMidiInput)
 							{
-								if( nph->origin() ==
-									nph->OriginMidiInput)
-								{
-									nph->setLength(
+								nph->setLength(
 									MidiTime( static_cast<f_cnt_t>(
 									nph->totalFramesPlayed() /
 									Engine::framesPerTick() ) ) );
-									midiNoteOff( *nph );
-								}
-								nph = NULL;
+								midiNoteOff( *nph );
 							}
+							nph = NULL;
 						}
 					}
 					m_sustainPedalPressed = false;
--- a/src/tracks/InstrumentTrack.cpp
+++ b/src/tracks/InstrumentTrack.cpp
@@ -273,7 +273,14 @@ void InstrumentTrack::processInEvent( const MidiEvent& event, const MidiTime& ti
 				// be deleted later automatically)
 				Engine::mixer()->requestChangeInModel();
 				m_notes[event.key()]->noteOff( offset );
-				m_notes[event.key()] = NULL;
+
+				if (!(isSustainPedalPressed()) ||
+					!(m_notes[event.key()]->origin() ==
+					m_notes[event.key()]->OriginMidiInput))
+				{
+					m_notes[event.key()] = NULL;
+				}
+
 				Engine::mixer()->doneChangeInModel();
 			}
 			eventHandled = true;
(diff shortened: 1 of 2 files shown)


src/tracks/InstrumentTrack.cpp
```diff
@@ -304,6 +311,28 @@ void InstrumentTrack::processInEvent( const MidiEvent& event, const MidiTime& ti
 				}
 				else
 				{
+					if (isSustainPedalPressed())
+					{
+						for (NotePlayHandle*& nph : m_notes)
+						{
+							if (nph != NULL)
+							{
+								if (nph->isReleased())
```
PhysSong: I think using && is better than using many if() clauses. In if (A && B), B isn't evaluated when A is false, maybe you know it already.
So it is safe to write like if (nph != NULL && nph->isReleased()) and that will make the code cleaner.

tresf: >  is safe to write like if (nph != NULL && nph->isReleased()) 

Or if you believe in such a thing... :) `if (nph && nph->isReleased())`

PhysSong:
Should the target branch be master? #1700 is in 1.2.0 milestone.

zonkmachine:
> Should the target branch be master? #1700 is in 1.2.0 milestone.

Yes, or we backport to stable-1.2.

@serdnab I've tested this and it works fine. It can't be tested completely because of https://github.com/LMMS/lmms/issues/3537.
As commented in the review you can cut down the number of if()'s.
Would it be rude to suggest you to have a look at #3537 too?

serdnab:
Made the shortening of the if's.

> Would it be rude to suggest you to have a look at #3537 too?
I will take a look at it.

src/tracks/InstrumentTrack.cpp
```diff
@@ -304,6 +311,25 @@ void InstrumentTrack::processInEvent( const MidiEvent& event, const MidiTime& ti
 				}
 				else
 				{
+					if (isSustainPedalPressed())
```
zonkmachine: I think you can write this into an `if else` with the `else` above to simplify it further.
And then lines 324-326 can get one tab each.

zonkmachine:
> I will take a look at it.

Awesome!

I restarted the Travis build. It sometimes will fail for reasons that are not related to the changes made.

serdnab:
Addressed the remarks.
I realized that the "else" is not necessary, if not `isSustainPedalPressed()`, `m_sustainPedalPressed` is already false.

zonkmachine:
> I realized that the "else" is not necessary, if not isSustainPedalPressed(), m_sustainPedalPressed is already false.

So, drop the test and just let there be: `else`

serdnab:
I believe you misunderstood me.
The `if(isSustainPedalPressed())` is necessary because `m_sustainPedalPressed` can be true or false, what is not necessary is a possible "else" for the  `if(isSustainPedalPressed())` clause.

zonkmachine:
> I believe you misunderstood me.

I believe you're right there... :)
Looking at this again.

> what is not necessary is a possible "else" for the if(isSustainPedalPressed()) clause.

This is true, but the `if( event.controllerValue() > MidiMaxControllerValue/2 )` above it still needs the `else`. You're testing a state, but this is the event that sets the state. Maybe you can show how you're thinking in code instead?

PhysSong:
@serdnab Or you can drop `else` if you put current `else if` clause first, though current code looks better. I suggest keeping this code.

serdnab:
I am not explaining well, then you are not understanding.
Anyway, the code I intend is the current one.

zonkmachine:
The only difference I see is that with dropping the test you have a built in safe guard in case of an unexpected state. When we get a pedal off command when `m_sustainPedalPressed` is `false`. This may never happen and/or may not have any side effects at all. I don't know if I'm being picky here. If other developers are cool with this I vote merge.

PhysSong:
@zonkmachine I suggest merging without more change.

zonkmachine:
OK. Squashed and merged. Thank you so much Andrés for fixing a long standing bug!

PhysSong:
@zonkmachine backport?

PR #3711, opened 2017-07-20
Fix time display

### LINKED ISSUE #2112: Bar:Beat:Tick / Min:Sec:Msec timer values disagree with playhead

PhysSong:
It fixes #2112, ~~also fixes https://github.com/LMMS/lmms/pull/3701#issuecomment-316512622.
If https://github.com/LMMS/lmms/pull/3701#issuecomment-316512622 is fixed by @michaelgregorius(or other person), I'll drop the fix for https://github.com/LMMS/lmms/pull/3701#issuecomment-316512622 and modify my PR.~~

Edit: @michaelgregorius have done fix by bcdb5ec, so commif 8ffebbb have been dropped.

### DIFF
--- a/include/Song.h
+++ b/include/Song.h
@@ -104,17 +104,36 @@ public:
 
 	inline int getMilliseconds() const
 	{
-		return m_elapsedMilliSeconds;
+		return m_elapsedMilliSeconds[m_playMode];
 	}
 
-	inline void setToTime( MidiTime const & midiTime )
+	inline int getMilliseconds(PlayModes playMode) const
 	{
-		m_elapsedMilliSeconds = midiTime.getTimeInMilliseconds(getTempo());
+		return m_elapsedMilliSeconds[playMode];
+	}
+
+	inline void setToTime(MidiTime const & midiTime)
+	{
+		m_elapsedMilliSeconds[m_playMode] = midiTime.getTimeInMilliseconds(getTempo());
+		m_playPos[m_playMode].setTicks(midiTime.getTicks());
+	}
+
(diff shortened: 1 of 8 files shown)


michaelgregorius:
@PhysSong My commit to fix #3701 has introduced a conflict but it should hopefully be possible to resolve by dropping 8ffebbb0436b4fe8b6b9c0896970578755a0a7b9 as you have already mentioned. The difference to your fix is that I have introduced a second methods so that the version with `tick_t` can still be used.

PhysSong:
@michaelgregorius Okay. I'll drop that commit and fix my PR.

PhysSong:
I found that `m_elapsedTicks` and `m_elapsedTacts` are no longer used. Could I remove them?

michaelgregorius:
If they are no longer used I'd say we should remove them.

tresf:
Tested, works.

Issue #3713, opened 2017-07-21
ExportProjectDialog may break on some build configs


irrenhaus3:
The file-scope function convertIndexToExportFileFormat(int) in gui/ExportProjectDialog.cpp implements the following hard mapping onto ProjectRenderer::ExportFileFormats values:  
0:WAV  
1:OGG  
2:MP3  
However, the actually availabe menu items that correspond to these indices are subject to build configurations.  
In a build config where LMMS_HAVE_OGGVORBIS is undefined, but LMMS_HAVE_MP3LAME is defined, the OGG export menu item would not exist, thus the MP3 item index would go off-by-one and be 1 instead of 2. Because the mapping function doesn't respect these options, selecting MP3 export would then show the controls for the vorbis renderer and possibly generate invalid settings for the mp3 renderer.  
  
Possible solutions:
- Check the build configurations for optional export types within the convertIndexToExportFileFormat function and use this information to handle possible offsets.
- Do not have a hard-coded order at all.

michaelgregorius:
@irrenhaus3 Can you please check if you can also fix the problem described [here](https://github.com/LMMS/lmms/pull/3615#pullrequestreview-51988486)? It's a related problem.

irrenhaus3:
@michaelgregorius @PhysSong Should be fixed via https://github.com/irrenhaus3/lmms/commit/ac932dcab3f08ef666169dff3590d0b643742bf4

PhysSong:
Merged #3714.

michaelgregorius:
@irrenhaus3 Thanks!

PR #3714, opened 2017-07-21
Respect build options in ExportProjectDialog

irrenhaus3:
Addresses issue #3713 by implementing a workaround on the existing hard-coded ordering. This is obviously not the optimal solution because the core problem is the hard-coded ordering itself.

### DIFF
--- a/src/gui/ExportProjectDialog.cpp
+++ b/src/gui/ExportProjectDialog.cpp
@@ -194,9 +194,13 @@ ProjectRenderer::ExportFileFormats convertIndexToExportFileFormat(int index)
 	case 0:
 		return ProjectRenderer::WaveFile;
 	case 1:
+#ifdef LMMS_HAVE_OGGVORBIS
 		return ProjectRenderer::OggFile;
 	case 2:
+#endif
+#ifdef LMMS_HAVE_MP3LAME
 		return ProjectRenderer::MP3File;
+#endif // If neither ogg nor mp3 are available, index 1 is OOB; fall through to default case
 	default:
 		Q_ASSERT(false);
 		break;
--- a/src/gui/ExportProjectDialog.cpp
+++ b/src/gui/ExportProjectDialog.cpp
@@ -66,8 +66,8 @@ ExportProjectDialog::ExportProjectDialog( const QString & _file_name,
 			// add to combo box
 			fileFormatCB->addItem( ProjectRenderer::tr(
 				ProjectRenderer::fileEncodeDevices[i].m_description ),
-                QVariant(ProjectRenderer::fileEncodeDevices[i].m_fileFormat) // format tag; later used for identification
-            );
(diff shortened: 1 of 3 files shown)


tresf:
> Do not have a hard-coded order at all

This seems like the best solution.

src/gui/ExportProjectDialog.cpp
```diff
@@ -194,9 +194,13 @@ ProjectRenderer::ExportFileFormats convertIndexToExportFileFormat(int index)
 	case 0:
 		return ProjectRenderer::WaveFile;
 	case 1:
+#ifdef LMMS_HAVE_OGGVORBIS
 		return ProjectRenderer::OggFile;
 	case 2:
+#endif
+#ifdef LMMS_HAVE_MP3LAME
 		return ProjectRenderer::MP3File;
+#endif // If neither ogg nor mp3 are available, index 1 is OOB; fall through to default case
```
tresf: This is hard to read and will be even harder to add new items to.  Let's instead parse "OGG, WAV, MP3" from the dropdown and make our assumptions there.

irrenhaus3: Agreed. But I don't think parsing strings is the best solution either. I think it's best to use the User Data field in QItem to tag each combo box item directly with the ProjectRenderer::ExportFileFormat, converted to int. That way, we can extract this value in onFileFormatChanged, convert it back to the enum type, and work directly with it. This would also make this dialog a lot more extensible for the purpose of adding more renderers.  
I'll work on it and update this pull request once I have something that works.

tresf: >  I don't think parsing strings is the best solution either.

The dialog mandates a file extension filter.  Parsing string is perfectly sane here.  I prefer enum approaches too and we need them elsewhere, but I see no logic flaw in `.ogg` = `OGG` at this point.  It's not going to change.

tresf:
@irrenhaus3 this is very clean, thank you.  Can you please reformat the code to use tabs instead of spaces?  Sorry for the inconvenience.

PhysSong:
@irrenhaus3 @tresf Did you test for every configuration?

tresf:
I did not test this PR at all.

irrenhaus3:
@PhysSong @tresf  I didn't test it for every configuration, just for (OGG enabled, MP3 disabled) and (both enabled). However, this PR specifically removes the need to check these configurations in the UI code and replaces them with runtime checks on whatever is available after this initialization: https://github.com/irrenhaus3/lmms/blob/6368bf1c886b961cbafe168c5d650eede62eb980/src/core/ProjectRenderer.cpp#L39  
It should therefore work for any arbitrary flag configuration and continue to work when new flags are added in the future.

PhysSong:
I'll do a final test and merge it soon. Thanks for your work, @irrenhaus3!

PhysSong:
Tested and works good. I'll merge it soon if there's no objection.

PhysSong:
Backported via b83c1bd.

PR #3722, opened 2017-07-23
Recover file fixes

zonkmachine:
I'll collect a bunch of, hopefully, last tweaks to the recover file system here.

 - Remove extra calls to autoSave() and a separate function runAutoSave() which was only related to the Limited Session that was dropped from the code in 290556e43d200952e85a9f54607af9501621d484.

 - Don't auto-save while playing by default. ( From criticism after RC3 where it created issues with glitches for users with weaker machines )

### DIFF
--- a/include/MainWindow.h
+++ b/include/MainWindow.h
@@ -105,8 +105,6 @@ public:
 		return m_autoSaveTimer.interval();
 	}
 
-	void runAutoSave();
-
 	enum SessionState
 	{
 		Normal,
--- a/src/core/main.cpp
+++ b/src/core/main.cpp
@@ -919,7 +919,6 @@ int main( int argc, char * * argv )
 		if( autoSaveEnabled )
 		{
 			gui->mainWindow()->autoSaveTimerReset();
-			gui->mainWindow()->autoSave();
 		}
 	}
 
--- a/src/gui/MainWindow.cpp
+++ b/src/gui/MainWindow.cpp
@@ -821,7 +821,6 @@ void MainWindow::createNewProject()
 	{
 		Engine::getSong()->createNewProject();
 	}
-	runAutoSave();
 }
 
 
@@ -840,7 +839,6 @@ void MainWindow::createNewProjectFromTemplate( QAction * _idx )
 		Engine::getSong()->createNewProjectFromTemplate(
 			dirBase + _idx->text() + ".mpt" );
 	}
-	runAutoSave();
 }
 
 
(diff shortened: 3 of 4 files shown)


tresf:
This PR had an in answered question when it was merged.

zonkmachine:
I missed that. Where? What was the question?

tresf:
> I missed that. Where? What was the question?

It's a commit review, so I can't hard-link it.  Please scroll through the commit code, you can't miss it.

zonkmachine:
Sorry, I force pushed and the comments seem gone. I can't find them in my mail either.

tresf:
screenshot_20170730-130535

zonkmachine:
When a review says `Pending`, it is only visible to you.
https://help.github.com/articles/reviewing-proposed-changes-in-a-pull-request/

Let's see, 7 days ago (review date) that really was a one line fix and the rest is removing old logic from the defunct Limited Session. I introduced `void MainWindow::runAutoSave()` to enforce saving whenever a new project was started to ensure there was a recovery.mmp. This would just be confusing to leave in as I've seen questions about why the recovery.mmp is spamming...

If you mean the [last update from today](https://github.com/LMMS/lmms/pull/3722/files#diff-523fb9d5bd2e855f54159b86476a3b20R1521) before I rebased the PR, no. If you just change the bool "1" the update function still is hard coded "true" so I did it this way. Still hard coded but now false.

tresf:
> When a review says Pending, it is only visible to you.
https://help.github.com/articles/reviewing-proposed-changes-in-a-pull-request/

Thanks that explains it.

So when does auto-save kick in?  Is it running in the background?  What triggers it?  Yes, the spamming of auto-save files is annoying and I tend to agree with that observation, but in my opinion, that's a failure to clean-up after successful save (or when explicitly click "Do not save").

If this feature causes performance issues, it's likely a threading prioritization issue.

zonkmachine:
> So when does auto-save kick in? Is it running in the background? What triggers it?

The timer is started [here](https://github.com/LMMS/lmms/blob/stable-1.2/src/core/main.cpp#L921) and the first save now is done after the first time-out of [m_autoSaveTimer](https://github.com/LMMS/lmms/blob/stable-1.2/src/gui/MainWindow.cpp#L194:L208).


Two of the most common complaints around the auto-save function are both user related:
1.  Error message: 'Could not open /recover.mmp for writing...'
Most likely from the user setting the working directory to `'/'`.
2. User doesn't save at all but uses auto-save as backup and therefore doesn't have any real safety net so when the crash finally comes they hit the forum with a post like: 'I had the auto-save on and worked for 2 days without turning the computer off but then it crashed and all was lost... How do I get my project back?'

Issue #3679, opened 2017-07-03
Add FLAC export support


PhysSong:
LMMS supports wav, ogg, and mp3 exports. Many other DAWs support FLAC export feature.
Although it is not strongly needed, it can be a convenient feature because FLAC is famous and widely used lossless audio encoding.

Since `libflac` and `libflac++` is dependencies of libsndio, no additional dependency will be needed.

musikBear:
compared to wave, what is the most obvious advantage?
Is the advantage significant (enough) to lead to *yet* a new feature?

PhysSong:
Compared to .wav, .flac file has smaller size. Furthermore, FLAC is lossless.

tresf:
@musikBear this isn't so much of a "feature" as MP3 support.  FLAC is a common encoding and it adds value.  Let's not bikeshed over such a trivial request please.

softrabbit:
> Since libflac and libflac++ is dependencies of libsndio, no additional dependency will be needed.

And then there's libsndfile, which LMMS already uses for audio file export (WAV at least). It has handled FLAC since 2009.

Umcaruje:
I actually find flac to be the best of the audio formats. It's lossless, and it doesn't hold as much space as a wav file. A win-win.

trollworkout:
Exporting FLAC is not really that useful. Maybe the ability to import FLAC into  a wav track would be more useful. You can always export wav and convert it to flac.  FLAC is more of a portable lossless audio format . It helps you get CD quality but in a smaller package . I see no other benefit.

tresf:
> Exporting FLAC is not really that useful. Maybe the ability to import FLAC into a wav track would be more useful.

You just explained a use-case.  If you're exporting something that's to be used again later, FLAC. :)

> You can always export wav and convert it to FLAC

The same can be said for MP3 as well, but it's one of our biggest complaints.  We should strive to support encoding and decoding equally and consistently unless there's a good reason not to.

Umcaruje:
>Maybe the ability to import FLAC into a wav track would be more useful.

Actually, you can load FLAC files in LMMS just fine, with no trouble at all.

>You can always export wav and convert it to FLAC.

Yes but as @tresf the same can go with any other format, having to use external software to have a flac file is just bad UX.

trollworkout:
I mean here's my take on it . Flac is great format because you know high quality smaller package .  So is PERFECT for a sample library. 

[...]

PR #3731, opened 2017-07-26
Add FLAC export and related options

irrenhaus3:
Implements #3679 by adding a FLAC rendering device to LMMS' Project renderer and related GUI configuration widgets in the export window.  
The renderer is implemented in terms of libsndfile and **hardens the dependency on it**, requiring a minimal version of 1.0.18.  
Additionally, since the FLAC compression level settings was not supported in libsndfile until version 1.0.26, a new CPP flag is added, called LMMS_HAVE_SF_COMPLEVEL. It is undefined if the sndfile version is less than 1.0.26. The "Compression Level" GUI element is only visible (and the setting received from it is only respected in the renderer) if LMMS_HAVE_SF_COMPLEVEL is defined.  
  
Unrelated to this new feature, this PR also introduces a minor bug fix regarding the output file extension. If the output format doesn't match the file extension set by the user before the export dialog window opens, the extension will automatically be adjusted when the rendering process starts.


src/core/audio/AudioFileFlac.cpp
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+#ifdef LMMS_HAVE_SF_COMPLEVEL
+	double compression = getOutputSettings().getCompressionLevel();
+	sf_command(m_sf,SFC_SET_COMPRESSION_LEVEL,&compression,sizeof(double));
+#endif
+
+	m_sf = sf_open(
+#ifdef LMMS_BUILD_WIN32
+		outputFile().toLocal8bit().constData(),
```
zonkmachine: `toLocal8bit()` -> `toLocal8Bit()`

irrenhaus3: Lol, and here I was wondering why travis exploded. Thanks!

zonkmachine:
> #include <QtGui/QAction>

For this you should be able to just go:
`#include <QAction>`

zonkmachine:
> plugins/sf2_player/ui_patches_dialog.h. The offending line is
> #include <QtGui/QAction>

I actually can't find that file or the #include line anywhere.

src/gui/ExportProjectDialog.cpp
```diff
@@ -79,30 +79,37 @@ ExportProjectDialog::ExportProjectDialog( const QString & _file_name,
...
 		}
 	}
+	
+	int const MAX_LEVEL=8;
+	for(int i=0; i<=MAX_LEVEL; ++i)
+	{
+		QString info="";
+		if (i==0){ info = tr("(fastest)", nullptr, QApplication::UnicodeUTF8); }
+		if (i==4){ info = tr("(default)", nullptr, QApplication::UnicodeUTF8); }
+		if (i==MAX_LEVEL){ info = tr("(smallest)", nullptr, QApplication::UnicodeUTF8); }
```
PhysSong: @irrenhaus3 I suggest using `tr("(something)")` without additional variable(s).

PhysSong: Here Qt5 build fails, too.

irrenhaus3:
My bad about that autogenerated SF2player GUI header. Turns out I didn't properly clean up my build directory and that particular header was left over from a previous QT4 build. Oops...

src/gui/ExportProjectDialog.cpp
irrenhaus3: Good catch, thank you. The symbol was removed in the newest commit.

include/AudioFileFlac.h
```diff
@@ -0,0 +1,48 @@
+/*
+ * AudioFileFlac.h - Audio device which encodes a wave stream into a FLAC file.
+ */
```
PhysSong: You should put the copyright information here.

src/core/audio/AudioFileFlac.cpp
```diff
@@ -0,0 +1,95 @@
+#include <memory>
```
PhysSong: Here, too.

PhysSong:
However, the rest things work great. I tested it with both 1.0.25 and 1.0.28.

PhysSong:
I found one more bug just now: It breaks "Export tracks" path setting. If I choose directory '/foo/bar', it tries to export tracks to '/foo/bar.mp3/1_asdf.mp3'.

PhysSong:
One more request: please add WANT_FLAC option to CMakeLists.txt and let the code be built with libsndfile 1.0.11~1.0.17 without flac export.

tresf:
> let the code be built with libsndfile 1.0.11~1.0.17 without flac export.

As stated in the PR, `sndfile >= 1.0.18` will be the new minimum version on our master branch.  This was discussed on Discord.  Legacy systems that do not have `sndfile >= 1.0.18` will have to use `stable-1.2` or older.  Why?  Because Ubuntu 10.04 offered 1.0.21, and that distribution is now 7 years old and was end of life 2 years ago. :)

PhysSong:
@tresf Quite true.
@irrenhaus3 However, WANT_FLAC and status message in CMakeLists.txt is still needed.

tresf:
> However, WANT_FLAC and status message in CMakeLists.txt is still needed.

Why?

PhysSong:
>Why?
@tresf Both OGG and MP3 have such things, so user can enable/disable such features manually.
https://github.com/LMMS/lmms/blob/stable-1.2/CMakeLists.txt#L55

tresf:
Because they are (or were) separate libraries.  Don't add a toggle for historical purposes please.  Instead, remove OGG.

PhysSong:
>Because they are (or were) separate libraries.

Yes.
>Don't add a toggle for historical purposes please.

I don't agree. They were added because of historical purposes, but I think they are still useful.
I can disable some features if I don't want them, for build speed and some other purposes.
>Instead, remove OGG toggle.

I agree that it should be, if FLAC toggle is not added, for consistency.

irrenhaus3:
@PhysSong: I agree with @tresf on this one. I really can't picture a use-case where a post-1.2 LMMS end user would want to disable FLAC, but keep WAV, especially since they're both implemented in terms of the same library. While I agree that it might make some sense in development and testing, this particular feature doesn't noticably increase build times or interfere with unrelated system components.  
As such, I'd like to refrain from implementing a WANT_FLAC mode for now. If the need for it ever arises, it can easily be done anyway. But in this PR, I'll only do it if the majority of developers agrees on it.

PhysSong:
@irrenhaus3 Then could you wait for other developers' opinion for CMakeLists.txt issue?
However, I think the code needs no more change(s) after @zonkmachine's [suggestions](https://github.com/LMMS/lmms/pull/3731#issuecomment-318950581) are applied.

zonkmachine:
Removing the ogg toggle option is better than adding a flac toggle. Keep it simple.

src/core/ProjectRenderer.cpp
```diff
@@ -42,6 +43,11 @@ const ProjectRenderer::FileEncodeDevice ProjectRenderer::fileEncodeDevices[] =
 	{ ProjectRenderer::WaveFile,
 		QT_TRANSLATE_NOOP( "ProjectRenderer", "WAV-File (*.wav)" ),
 					".wav", &AudioFileWave::getInst },
+    { ProjectRenderer::FlacFile,
```
tresf: @irrenhaus3 this needs formatting, spaces versus tabs. :)

tresf:
A follow-up task to merging this should be to update the required `libsndfile` version on our [`Compiling`](https://github.com/LMMS/lmms/wiki/Compiling#libraries) wiki page.

PhysSong:
@irrenhaus3 Then could you remove WANT_OGG toggle and related flags?

irrenhaus3:
Update for the interested reader not on the LMMS Discord server:  
In a discussion on the developers channel, we found that refactoring the OGG exporter to use sndfile and removing the WANT_OGGVORBIS flag are separate issues from FLAC export and will be handled in their own PR.

src/core/audio/AudioFileFlac.cpp
```diff
@@ -0,0 +1,119 @@
...
+#ifdef LMMS_BUILD_WIN32
+		outputFile().toLocal8Bit().constData(),
+#else
+		outputFile().toUtf8().constData(),
+#endif
+		SFM_WRITE,
+		&m_sfinfo
+	);
+
+	sf_command(m_sf,SFC_SET_CLIPPING,nullptr,SF_TRUE);
```
tresf: It is conventional to put spaces between parameters.

PhysSong: Like `foo(b, a, r);`.

PhysSong:
I recommend adding a status message for FLAC, like WAV one. It's not necessary, but I think having this will be better.

gi0e5b06:
+1 for WANT_FLAC
Anything that use external libraries should be flagged.

PhysSong:
Great! Now everything looks fine.
I'll merge it soon, if there's no objection.

zthurman:
I'm probably being a dumb but I'm not seeing this in the front end options on lmms 1.2.0-rc7 running on Arch or the lmms-git version 1.2.0-rc7.11.

Am I understanding correctly from the above that I have to build from source and tweak out a bit in order to use this enhancement?

zonkmachine:
> I'm probably being a dumb but I'm not seeing this in the front end options on lmms 1.2.0-rc7 running on Arch or the lmms-git version 1.2.0-rc7.11.

It's because this has been added to master and will be in lmms-1.3.

> Am I understanding correctly from the above that I have to build from source and tweak out a bit in order to use this enhancement?

You either build `master` or you can cherry-pick commit https://github.com/LMMS/lmms/commit/e9a406311969397a9d4219eb80671d27db826e87 to stable-1.2 .
I've seen unofficial builds around for master that you could try out but I don't know precisely where or if they are up to date. Check out our support channel at discord, https://discord.gg/3sc5su7 (`#support`), and ask there. You'll get help much sooner there I think. This is a closed pull request and late comments on those tend to get little attention.

PR #3742, opened 2017-08-01
Fix deadlock when exporting is finished, if the project has any Sample TCO(s).

### LINKED ISSUE #3344: SDL - problems with export of project with Sample Track
As discussed in PR https://github.com/LMMS/lmms/pull/3341 there is some issues with exporting projects containing a Sample Track. I've only seen this when using **SDL**. **ALSA** works fine and other sound engines I haven't tried.

The renderer process will simply hang and never return to main(). If you do a `--rendertracks` export it will stop on the first Sample Track it tries tor render.

I'm not done bisecting this but it may have happened around these commits 8497d39eec78f3f23f0581898e18868025d5070e e55ec80c3d956dea8674afc0755bc87831c64ce6

PhysSong:
Fixes #3344.
Calls `stopProcessing()` when the rendering ends, to prevent `requsetChangeInModels()` from waiting for `runChangesInModels()`, which causes a deadlock.
https://github.com/LMMS/lmms/issues/3344#issuecomment-318878520 is another solution, but this one is more fundamental than https://github.com/LMMS/lmms/issues/3344#issuecomment-318878520.

Tested for both GUI and CLI, and both single-file rendering and rendering per track, works well.

### DIFF
--- a/include/Mixer.h
+++ b/include/Mixer.h
@@ -421,6 +421,7 @@ private:
 
 	friend class LmmsCore;
 	friend class MixerWorkerThread;
+	friend class ProjectRenderer;
 
 } ;
 
--- a/src/core/ProjectRenderer.cpp
+++ b/src/core/ProjectRenderer.cpp
@@ -176,8 +176,8 @@ void ProjectRenderer::run()
 
 	Engine::getSong()->startExport();
 	Engine::getSong()->updateLength();
-	//skip first empty buffer
-	Engine::mixer()->nextBuffer();
+    //skip first empty buffer
+    Engine::mixer()->nextBuffer();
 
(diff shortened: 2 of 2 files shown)


zonkmachine:
Tested, works!

Exported from gui and command line, single samples and complex projects.
As for the code of it I'm not qualified to answer... :)  I suggest merge.

PhysSong:
Merge?

zonkmachine:
@PhysSong I'll 'bump' this to master too.

PhysSong:
@zonkmachine Does it really needed? It will be in `master` once we merge `stable-1.2` into`master`.

zonkmachine:
?! You're right there. I don't know where I got the idea that this was the way to go actually...
Probably from the large merge of `stable-1.2` into `master` we had which got a bit cumbersome.

Issue #2612, opened 2016-02-26
Tab MISC | 'Enable the use of Master-pitch' Diodes has slave behavior


musikBear:
PR 17. feb. g23d284 win32

To reproduce:
Insert any number of presets or sample tracks in B&B-editor
Add some notes.

Open top-track
Go to MISC. 
Toggle the diode for MASTER-PITCH
Open any track below the top track
The diode for  MASTER-PITCH will have changed to the value of the top-track as a slave.

Does current windows Master (still) have this behavior?

Demofile: https://lmms.io/lsp/download_file.php?file=8152&name=masterpitchErrors.mmp

Edit: Similar issue closed as duplicate: https://github.com/LMMS/lmms/issues/2613

zonkmachine:
There's absolutely something fishy here. Is this only in a Beat track?

musikBear:
@zonkmachine ..dont know why i did not test that too :mag_right: 
No, song-editor has the same bug.

zonkmachine:
Names gets scrambled too.
Here are the tracks `TripleOscillator`, `1` and `2` selected but the opened instruments are named
`5`, `4` and `TripleOscillator` respectively. This is most likely related.
![masterpitchbug](https://cloud.githubusercontent.com/assets/6368949/18066499/fb86099a-6e38-11e6-844d-2e85348d4e10.png)

zonkmachine:
I got a crash while probing this that could be related. :beer: 

[bt full](https://gist.github.com/zonkmachine/b95543f56ccf0b9d040442497f595594)

```gdb
Segmentation fault (core dumped)
zonkmachine@zonkmachine:~/builds/lmms/lmms/build$ gdb ./lmms core
GNU gdb (Ubuntu 7.7.1-0ubuntu5~14.04.2) 7.7.1
...
Reading symbols from ./lmms...done.
[New LWP 25825]
[New LWP 25826]
[New LWP 25828]
[New LWP 25830]
[New LWP 25829]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Core was generated by `./lmms'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00000000004e75cf in AutomatableModel::value<bool> (this=0x0, frameOffset=0) at /home/zonkmachine/builds/lmms/lmms/include/AutomatableModel.h:134
[...]

PR #3753, opened 2017-08-06
"misc" view now shows the model state, of the track use of master pitch

serdnab:
Fix for #2612.
When an already created instance of `InstrumentTrackWindow` is used, the master pitch `GroupBox` isn't refreshed.
Now, if this is the case, when `modelChanged()` is invoked, the refresh is done.

### DIFF
--- a/include/InstrumentMidiIOView.h
+++ b/include/InstrumentMidiIOView.h
@@ -72,6 +72,11 @@ public:
 	InstrumentMiscView( InstrumentTrack *it, QWidget* parent );
 	~InstrumentMiscView();
 
+	GroupBox * getPitchGroupBox()
+	{
+		return m_pitchGroupBox;
+	}
+
 private:
 
 	GroupBox * m_pitchGroupBox;
--- a/include/InstrumentMidiIOView.h
+++ b/include/InstrumentMidiIOView.h
@@ -72,7 +72,7 @@ public:
 	InstrumentMiscView( InstrumentTrack *it, QWidget* parent );
 	~InstrumentMiscView();
 
-	GroupBox * getPitchGroupBox()
+	GroupBox * PitchGroupBox()
 	{
 		return m_pitchGroupBox;
 	}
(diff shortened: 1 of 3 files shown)


PhysSong:
Good observation! I haven't tested it yet, but It should be work.

gi0e5b06:
getPitchGroupBox() should be renamed pitchGroupBox(), according to Qt and Lmms coventions.
Use the read accessor and not directly the field:
m_miscView->getPitchGroupBox()->setModel(&m_track->m_useMasterPitchModel);
should be:
m_miscView->pitchGroupBox()->setModel(m_track.useMasterPitchModel());
wgich requires to have the models mapped as properties, like volume.

Umcaruje:
I tested this and it works like a charm :+1: But @gi0e5b06 is right it seems inconsistent with the rest of the software so that'd need to change before a merge.

PhysSong:
I think @gi0e5b06 is right, too.
>getPitchGroupBox() should be renamed pitchGroupBox()

Yes, it's our convention.
>Use the read accessor and not directly the field

It is recommended, but I think it's not necessary.

serdnab:
> getPitchGroupBox() should be renamed pitchGroupBox()

> Yes, it's our convention.

Addressed

include/InstrumentMidiIOView.h
```diff
@@ -72,7 +72,7 @@ class InstrumentMiscView : public QWidget
 	InstrumentMiscView( InstrumentTrack *it, QWidget* parent );
 	~InstrumentMiscView();
 
-	GroupBox * getPitchGroupBox()
+	GroupBox * PitchGroupBox()
```
zonkmachine: Lower case here: `PitchGroupBox()` -> `pitchGroupBox()`

PhysSong: Yes. It's also our convention.

PhysSong: @serdnab This is [our convention](https://github.com/LMMS/lmms/wiki/Coding-conventions), too.

PhysSong:
@serdnab?

serdnab:
Done

zonkmachine:
Tested! This fixes the issue.
Merge?

PR #3761, opened 2017-08-10
Fix producing of NaN from Env/LFO parameter change while playing.

### LINKED ISSUE #3408: Envelope triggered NaN

PhysSong:
Fixes #3408.

Guarantee thread safety to ensure `fillLevel()` not to read value from wrong buffer address, because of buffer reallocation in `updateSampleVars()`.

### DIFF
--- a/src/core/EnvelopeAndLfoParameters.cpp
+++ b/src/core/EnvelopeAndLfoParameters.cpp
@@ -291,7 +291,7 @@ void EnvelopeAndLfoParameters::fillLevel( float * _buf, f_cnt_t _frame,
 						const f_cnt_t _release_begin,
 						const fpp_t _frames )
 {
-	m_paramMutex.lock();
+	QMutexLocker m(&m_paramMutex);
 
 	if( _frame < 0 || _release_begin < 0 )
 	{
@@ -330,8 +330,6 @@ void EnvelopeAndLfoParameters::fillLevel( float * _buf, f_cnt_t _frame,
 			env_level * ( 0.5f + *_buf ) :
 			env_level + *_buf;
 	}
-
-	m_paramMutex.unlock();
 }
 
 
@@ -406,7 +404,7 @@ void EnvelopeAndLfoParameters::loadSettings( const QDomElement & _this )
 
 void EnvelopeAndLfoParameters::updateSampleVars()
 {
-	m_paramMutex.lock();
+	QMutexLocker m(&m_paramMutex);
 
 	const float frames_per_env_seg = SECS_PER_ENV_SEGMENT *
 				Engine::mixer()->processingSampleRate();
@@ -536,8 +534,6 @@ void EnvelopeAndLfoParameters::updateSampleVars()
 	}
 
 	m_bad_lfoShapeData = true;
-
-	m_paramMutex.unlock();
 
 	emit dataChanged();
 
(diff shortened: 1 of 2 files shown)


src/core/EnvelopeAndLfoParameters.cpp
```diff
@@ -291,6 +291,8 @@ void EnvelopeAndLfoParameters::fillLevel( float * _buf, f_cnt_t _frame,
 						const f_cnt_t _release_begin,
 						const fpp_t _frames )
 {
+	m_paramMutex.lock();
```
michaelgregorius: Please use a [QMutexLocker](https://doc.qt.io/qt-4.8/qmutexlocker.html) here to ensure safe unlocking in cases of exceptions, etc.

src/core/EnvelopeAndLfoParameters.cpp
```diff
@@ -402,6 +406,8 @@ void EnvelopeAndLfoParameters::loadSettings( const QDomElement & _this )
 
 void EnvelopeAndLfoParameters::updateSampleVars()
 {
+	m_paramMutex.lock();
```
michaelgregorius: Please use a [QMutexLocker](https://doc.qt.io/qt-4.8/qmutexlocker.html) here to ensure safe unlocking in cases of exceptions, etc.

PhysSong:
@michaelgregorius Done via 667ee11.

PhysSong:
I'll merge it after 12~24 hours if there's no objection.

PR #3770, opened 2017-08-16
Fix hang when remote plugin is missing

DomClark:
Invalidate the plugin if the remote plugin executable is missing, so LMMS doesn't hang on RemotePlugin::waitForInitDone.

### DIFF
--- a/src/core/RemotePlugin.cpp
+++ b/src/core/RemotePlugin.cpp
@@ -196,6 +196,7 @@ bool RemotePlugin::init( const QString &pluginExecutable,
 		qWarning( "Remote plugin '%s' not found.",
 						exec.toUtf8().constData() );
 		m_failed = true;
+		invalidate();
 		unlock();
 		return failed();
 	}


zonkmachine:
Hello, thanks for the PR!

> so LMMS doesn't hang on RemotePlugin::waitForInitDone.

I'm not usually into the 'remote' stuff but I'm currently testing this out.
How do I make lmms hang? I've tried and rename the zynaddsubfx executable and some vst's that are used in a project and then relaunched lmms but I get no hang there. Do you have a recipe for this?

DomClark:
Hello,

The hang happens when the remote executable (i.e. RemoteZynAddSubFx.exe, RemoteVstPlugin.exe or RemoteVstPlugin32.exe) is missing, and LMMS tries to use the associated plugin (i.e. ZynAddSubFx or VeSTige). The issue occurs for me (Windows 10 64-bit) on opening the Zyn GUI or loading a VST of the appropriate type (64/32 bit).

What happens is this (for VeSTige, Zyn does something similar, line numbers are master at time of writing):


`vestigeInstrument::loadFile` (plugins/vestige/vestige.cpp:233) is called with the name of the plugin library, and passes this to the constructor of a new VstPlugin on line 254.

`VstPlugin::VstPlugin` (plugins/vst_base/VstPlugin.cpp:83) takes this, and calls `tryLoad( REMOTE_VST_PLUGIN_FILEPATH )` on line 102. `REMOTE_VST_PLUGIN_FILEPATH` is defined somewhere in the cmake files, but is ultimately `plugins/RemoteVstPlugin` or something like that.

`VstPlugin::tryLoad` (plugins/vst_base/VstPlugin.cpp:134) then passes `REMOTE_VST_PLUGIN_FILEPATH` on to `init`, along with `false`.

`RemotePlugin::init` (core/RemotePlugin.cpp:167) attempts to convert its first argument to an actual file path, and tests (line 194) whether such a file exists. If so, everything is fine and we continue to start the executable. If not, `m_failed` is set to `true` and the function returns. This is the case we're looking at here.

Back in `VstPlugin::tryLoad`, we encounter our problem. On line 183 `waitForInitDone` is called regardless of whether or not we succeeded loading the executable.

`RemotePlugin::waitForInitDone` (include/RemotePlugin.h:772) calls `waitForMessage( IdInitDone, true)`

`RemotePluginBase::waitForMessage` (include/RemotePlugin.h:1071) is then stuck in an infinite loop on lines 1075-1082; `isInvalid()` is `false`, `messagesLeft()` is always `false` since the remote executable is not running, so we aren't receiving any messages, so we hit `continue` and the only thing we do for the rest of time is process non-user-input events.


Of course, this could be fixed by simply aborting `VstPlugin::tryLoad` if we don't manage to load anything, but calling `invalidate()` fits in with the other remote process issue handling case (core/RemotePlugin.cpp:65, which happens if the process dies unexpectedly).

Obviously this isn't a condition that should ever happen normally (I ran into it while chasing another bug, turns out MSYS2 refuses to build RemoteVstPlugin32 for some reason), but I feel it ought to be handled nonetheless.

Umcaruje:
This seems like a simple enough change, should we merge @DomClark @zonkmachine ?

zonkmachine:
Sure. It's one line. What could possibly go wrong? :P
I'll leave it to you and @DomClark to decide on this one.

zonkmachine:
:tractor:

Issue #3507, opened 2017-04-20
Feature request: Option to disable animated expand/collapse of panels


Oddwarg:
This is a minor issue, and I am using Windows 7.

In the Instrument Plugins menu, moving the mouse over the instruments will make the panels slowly expand and collapse, offsetting the items below in the list and creating an artificial delay before the user can read the full description. This behaviour slows down user interaction.

![image](https://cloud.githubusercontent.com/assets/6345473/25237230/124c3c4e-25b8-11e7-92df-50d487c872a5.png)

Ideally, the UI should be redesigned so that the appearance of the description does not offset (nor occlude) the items below in the list. However, the ability to either speed up or entirely disable the animation would be a helpful patch.

tresf:
I believe this is a `10ms` call to `updateHeight()`.

https://github.com/LMMS/lmms/blob/master/src/gui/PluginBrowser.cpp#L235

Simply changing [`+1` to `+10`](https://github.com/LMMS/lmms/blob/master/src/gui/PluginBrowser.cpp#L221) should do the trick for faster animation.

Umcaruje:
I agree on this.

zonkmachine:
I believe this was fixed in #3773

PR #3773, opened 2017-08-23
UI: New behavior for the Plugin Browser

irrenhaus3:
This PR makes the following changes to the GUI:  
* Plugin browser widgets no longer resize on mouse-over; the list stays at a fixed size at all times.
* The detailed information for every plugin is no longer displayed within the plugin widgets. Instead, the title, icon and text of the header widget above the list is updated on the fly.
* The default theme now highlights plugin widgets on mouse-over.
  
See the following demonstration.
Demo
  
~~**Note** This PR is currently _not_ ready to merge. There's still a major issue regarding the height of the description text label, causing said text to be cropped in some circumstances. The main reason I'm opening it anyway is for people to look over it, review it and suggest improvements and solutions to existing problems.~~  
Edit2: Since we're now using tool-tips to display plugin details, the header height problem is obsolete. The PR is ready for merging.
  
Edit: This PR sort-of implements #3507 in that it removes the resizing behaviour altogether.

### DIFF
--- a/data/themes/default/style.css
+++ b/data/themes/default/style.css
@@ -537,7 +537,7 @@ PluginDescWidget {
 }
 
 PluginDescWidget:hover {
-	background-color: qlineargradient(spread:reflect, x1:0, y1:0, x2:0, y2:1, stop:0 #2E343B, stop:1 #1E2326);
+	background-color: qlineargradient(spread:reflect, x1:0, y1:0, x2:0, y2:1, stop:0 #363F4B, stop:1 #4A545F);
 	color: #d1d8e4;
 }
 
--- a/data/themes/default/style.css
+++ b/data/themes/default/style.css
@@ -537,7 +537,7 @@ PluginDescWidget {
 }
 
 PluginDescWidget:hover {
-	background-color: qlineargradient(spread:reflect, x1:0, y1:0, x2:0, y2:1, stop:0 #363F4B, stop:1 #4A545F);
+	background: qlineargradient(spread:reflect, x1:0, y1:0, x2:0, y2:1, stop:0 #7C8799, stop:1 #343840);
 	color: #d1d8e4;
 }
 
--- a/include/PluginBrowser.h
+++ b/include/PluginBrowser.h
@@ -62,14 +62,13 @@ class PluginDescWidget : public QWidget
 	Q_OBJECT
 public:
 	PluginDescWidget( const Plugin::Descriptor & _pd, QWidget * _parent );
-	virtual ~PluginDescWidget();
 
 
 protected:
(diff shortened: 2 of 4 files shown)


Spekular:
This looks great. I do still think that having the description at the top might not be ideal. Normally I'd assume users' eyes "seek" to where their mouse is, and with this design they need to dart back and forth between wherever they are the list and the top to see information about the plugin.

I think an alternative would be to display a tooltip. This accomplishes the goal of unifying user focus + mouse + information, but doesn't rely on the expanding description.

Sawuare:
@irrenhaus3
Looks great!
Is the color of the highlighting here the same as the color of the highlighting of other GUI elements like the buttons in the editors?

@Spekular
I agree. This way, users who are familiar with the program and the instrument plug-ins (these users usually have tooltips disabled) can disable tooltips so they don't have to see these descriptions.

irrenhaus3:
@Sawuare No, I just put together an arbitrary highlighting gradient, because I didn't know there was a standard for it. I'll change it to use the standard instead. :smile:  
I'll also change the plugin description to display in a tooltip instead of the header. This comes along with a lot of reverts though, so it may take a while. But I definitely agree that it's a better option than having to deal with text size issues in a constrained space.

irrenhaus3:
> so it may take a while.  
  
Or not, lol.

Spekular:
With this done, I suppose the header for Instrument Plugins could be shortened to the same height as the other tabs? I also really hate to nitpick when you've already fixed the things I complained about, but does the tooltip stay inside the window properly?

If my LMMS window is so short that a plugin in the list ends up right at the bottom edge, the tooltip should hopefully shift up so it doesn't go offscreen. Otherwise it could be wonky or impossible to read it.

image

If it does go offscreen, it should hopefully let the users scroll down to read the whole text. This is what the current implementation does, so while less convenient, it wouldn't be a regression for the tooltip to act this way.

Thanks so much for working on this!

irrenhaus3:
@Spekular If the location of the tool-tip is properly within the desktop bounds, it displays underneath the cursor (if necessary, partly outside the LMMS window borders). If it's at the bottom edge of the desktop, though, it shifts upwards and displays above the cursor, so it's still fully contained in the visible desktop area. Same goes for behavior at the left/right desktop edges.  
I do believe this is default QT behavior for tool-tips, so we don't need to take care of that in client code.
  
Also, there's no need to feel bad about bringing stuff like this up. I'm not one to take code reviews personally. :wink:

tresf:
Cleans up code, makes the UI more consistent, approved by UI coders.  Another benefit is the average user may wait for the animation to finish each time, so this has a slight productivity gain. 
 Tested, works, merging.

Issue #3757, opened 2017-08-07
Consecutive notes won't work with sustain


Umcaruje:
Regression caused by #3710, reverting the commit removes the bug.

Steps to reproduce: 
1. Connect any instrument to a midi controller with a sustain pedal/button
1. Hold down the sustain pedal/button
1. Press any key more than one time
1. Only the first key press will register and the consecutive keys won't work or produce any sound until you release the sustain pedal.
 
If you revert the said commit, in case of `stable-1.2`: 
```
git revert d87a8b3da4976b669600972c4e731f8a4f6a6fb5
```
The behavior is as expected, they keys keep playing the same sound and layering over one another.

If my steps are unclear I can provide a video.

Tagging @serdnab as the author of the PR.

PhysSong:
Confirmed.

[Here](https://github.com/LMMS/lmms/blob/d87a8b3da4976b669600972c4e731f8a4f6a6fb5/src/tracks/InstrumentTrack.cpp#L277-L282) m_notes[event.key()] is not cleared if sustain pedal is pressed, so the [note-on condition](https://github.com/LMMS/lmms/blob/d87a8b3da4976b669600972c4e731f8a4f6a6fb5/src/tracks/InstrumentTrack.cpp#L250) is not satisfied.

PhysSong:
Possible solution:
* Remove if clause which guards `m_notes[event.key()] = NULL;`.
* Store sustained notes in other place.
* Read note handle list from the new place [here](https://github.com/LMMS/lmms/blob/d87a8b3da4976b669600972c4e731f8a4f6a6fb5/src/tracks/InstrumentTrack.cpp#L314).

zonkmachine:
> * Remove if clause which guards m_notes[event.key()] = NULL;.

This, tried it and it works. One drawback though. Today, if you strike a midi keyboard and then enable it via the instrument, 'Enable Midi Input' drop-down menu, you will have notes stuck on. You can 'unstick' these notes by hitting the same note again. When I remove the 'if' mentioned above the notes are stuck bad and I have no method to silence them. This is an issue already and should be fixed. Maybe running `InstrumentTrack::silenceAllNotes(` when connecting a keyboard?

PhysSong:
**All of** these three things should be done. If not, these issue will take a place.
I'll complete my solution tomorrow and open a PR if needed.

zonkmachine:
OK. Slow day over here.. 8)

PhysSong:
[...]

PR #3774, opened 2017-08-24
same note layering when sustain pedal is pressed

serdnab:
fix for #3757 
I moved the signal emission for note ended to where all notes (sustained and not sustained) go, the release process on `NotePlayHandle:play`

### DIFF
--- a/src/core/NotePlayHandle.cpp
+++ b/src/core/NotePlayHandle.cpp
@@ -252,8 +252,17 @@ void NotePlayHandle::play( sampleFrame * _working_buffer )
 	if( m_released && (!instrumentTrack()->isSustainPedalPressed() ||
 		m_releaseStarted) )
 	{
-		m_releaseStarted = true;
+		if (m_releaseStarted == false)
+		{
 
+			if( m_origin == OriginMidiInput )
+			{
+				setLength( MidiTime( static_cast<f_cnt_t>( totalFramesPlayed() / Engine::framesPerTick() ) ) );
+				m_instrumentTrack->midiNoteOff( *this );
+			}
+
+			m_releaseStarted = true;
+		}
 		f_cnt_t todo = framesThisPeriod;
 
 		// if this note is base-note for arpeggio, always set
@@ -379,16 +388,6 @@ void NotePlayHandle::noteOff( const f_cnt_t _s )
 				MidiEvent( MidiNoteOff, midiChannel(), midiKey(), 0 ),
 				MidiTime::fromFrames( _s, Engine::framesPerTick() ),
 				_s );
-	}
-
-	// inform attached components about MIDI finished (used for recording in Piano Roll)
-	if (!instrumentTrack()->isSustainPedalPressed())
-	{
-		if( m_origin == OriginMidiInput )
-		{
-			setLength( MidiTime( static_cast<f_cnt_t>( totalFramesPlayed() / Engine::framesPerTick() ) ) );
-			m_instrumentTrack->midiNoteOff( *this );
-		}
 	}
 }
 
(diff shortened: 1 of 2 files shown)


PhysSong:
Haven't tested yet, but looks good. I'll test it soon.

PhysSong:
Tested. It works great, and I can't find any regressions caused by this.

src/core/NotePlayHandle.cpp
```diff
@@ -252,8 +252,17 @@ void NotePlayHandle::play( sampleFrame * _working_buffer )
 	if( m_released && (!instrumentTrack()->isSustainPedalPressed() ||
 		m_releaseStarted) )
 	{
-		m_releaseStarted = true;
+		if (m_releaseStarted == false)
```
zonkmachine: Don't mix styles. This would be comparably to the rest of your changes in this PR and surrounding code.

```cpp
if( m_releaseStarted == false )
```

PhysSong: `if (a)` is new convention, `if( a )` is old one. Many things are written with old convention, which make the code inconsistent for recently edited files.
I suggest using new style in line 252.

zonkmachine: OK. ;)

PhysSong:
It'd be better to either change line 252 and 258 to new style, or use old convention in line 255.

zonkmachine:
> So, I have to change something?

We had a discussion on this over at the discord developer channel. Since the conventions have changed the code base is inconsistent anyway. Merged. Thanks for fixing this!

zonkmachine:
Backported in efd0d34f2babd0020d0a1d76c4dc3c4458e67458

zonkmachine:
An oops with this PR. Recording single streamed instruments just disappeared...

serdnab:
> An oops with this PR. Recording single streamed instruments just disappeared...

@PhysSong I will use your proposed solution (here https://github.com/LMMS/lmms/issues/3757#issuecomment-320881503).
Do you want to code yourself?
If not, I will be glad of doing it myself.

PhysSong:
I'm going to fix #3777. I'll try to fix this too.
If I say I've failed, please do this @serdnab.

serdnab:
I will take a look at #3777,

> An oops with this PR. Recording single streamed instruments just disappeared...

 and are you on this?

PhysSong:
>Maybe open a separate issue?

:+1: I think so.

PhysSong:
I tried my suggestion in https://github.com/LMMS/lmms/issues/3757#issuecomment-320881503. It seems to fix both issues, but it is an incomplete solution for #3777.

serdnab:
@PhysSong I have a solution for #3777 but I'm waiting your PR because I will use the place where you store the sustained notes.

PhysSong:
@serdnab What is your solution for #3777?

serdnab:
> What is your solution for #3777?

Assign an unique ID for every instance of `Note` and then on `finishRecordNote` at `PianoRoll.cpp` compare by that ID.

PhysSong:
@serdnab That can be a permanent and fundamental solution, if the uniqueness is guaranteed and the ID won't be changed. I'll send a PR for single-stream recording fix.

serdnab:
The problem is not that instruments are single-stream, it is that they don't have a release.
Some instruments, even if they have envelope off, have an internal release.
Try with Vibed instrument and you will see that when envelope is on it records normally.

zonkmachine:
> Try with Vibed instrument and you will see that when envelope is on it records normally.

Confirmed. **sfxr** too.

zonkmachine:
> An oops with this PR. Recording single streamed instruments just disappeared...


> @PhysSong Is looking at this one. Maybe open a separate issue?


> The problem is not that instruments are single-stream, it is that they don't have a release.
> Some instruments, even if they have envelope off, have an internal release.

I don't have time to write up an issue for this right now. RC4?

PhysSong:
I will open a PR that fixes recording single-stream instruments tomorrow. RC4 is really coming :)

PR #3776, opened 2017-08-24
Fix some VST deadlocks/hangs

### LINKED ISSUE #1008: Freeze on loading project with Kontakt VST plugin
Hey there,
[Kontakt 5 Player](http://www.native-instruments.com/en/products/komplete/samplers/kontakt-5-player/) works fine with LMMS 1.0.2 until I save and load my project.
While loading the project, the plugin loads but LMMS freezes before finishing. I think it has trouble loading the plugin settings.
My OS is Win 8 64 Bit.

### LINKED ISSUE #1183: VSTs crash sometimes when parameters are changed
This happens to me with Synth1 and the Syelnth1 demo, to name a couple. After about 5 - 10 minutes of tweaking knobs, LMMS freezes up. It happens when I change a parameter, and I don't think it matters which. It loads fine though, and if I don't change any parameters, it doesn't crash.

Some notes:
The five minutes isn't exact. That's about how long it takes for it to crash, but it has nothing to do with the time.

When I went to test this today, it crashed synth1 when starting and then didn't crash at all after that. I'm sure I didn't change anything except possibly my behavior. (I started synth1, opened up the parameter list and tweaked knobs, then switched to the GUI, then switched to sylenth1 demo without restarting or making a new vestige instance and repeated the same thing - this took at least half an hour.)

I can do further testing if anybody would like me to. This crashing after 5 minutes happened at least 4 times last time I was trying this (same LMMS version and plugin versions).

I'm running Windows 7, 64 bit, but I'm using 32 bit VSTs as far as I know. I can check this if it's important

### LINKED ISSUE #1468: VST/(i) Debug message.
Hi ,

I always get a debug ( _non fatal and has been happening since the 0.4.x series_ ) message whenever I load a VST instrument or effect into LMMS :-        **" failed getting shared memory "**

The debug message originates from include/RemotePlugin.h:1161: and include/RemotePlugin.h:1180:

What I would like answered:- Is shmem actually being allocated to the VST's?

Is there an understandable way to show what ( _if any_ ) shared memory is being allocated?

I have had a brief look at '''icps''' and also searching the pid of the RemoteVstPlugin.exe process through cat /proc/*/maps, but I'm a bit unsure of what I'm looking for.

thanks Mikobuntu :

### LINKED ISSUE #2157: Crash after clicking on Dexed VST "init" button several times
Hello everyone,

I noticed that when you open Dexed VST (https://github.com/asb2m10/dexed) on LMMS and click several times on the "init" button (sometimes even 2 times are enough), LMMS freezes and needs to be closed/opened again to be usable. I have also made the same test on FL Studio 12 and couldn't reproduce the issue.

Used LMMS configuration:
Buffer size: 64
Latency: 1.5ms

My PC configuration:
Windows 7 SP1
INTEL Core i5-3210M @2.5GHz
64 Bits
8GB of ram
VIA High Definition Audio
INTEL HD 4000 + Geforce 645M GT

### LINKED ISSUE #2784: Switching Helm presets crashes LMMS
[Helm](http://tytel.org/helm/) is open source. I used the 64-bit Windows version and reproduced on two different machines.
1. Load Helm to VeSTige
2. Drop some notes in the piano roll and hit play
3. Click Browse in Helm and switch presets

It might crash the first time or after a couple of preset changes.
LMMS 1.1.90.
Helm 0.6.0.

### LINKED ISSUE #3274: Kontakt VST (among others) crashes after reloading the project
I have been using LMMS for a while now, and it's a very usefull program. The biggest problem I have is that there are a few vst's that will crash LMMS. At first they will load, but after I save the project and open it again, it crashes. And it usually pops up an error saying something about the Maize sampler player was unable to load this instrument. If more information is needed, let me know!

P.S. I use Windows 10

DomClark:
Fixes #1008, fixes #1183, fixes #3274.

All problems referred to here I've (re)produced on 64-bit Windows 10.

I've experienced #1183 (or something matching the description), and it is caused by a deadlock during autosave. During the autosave, LMMS sends a message to the plugin (generally `IdSaveSettingsToFile`) and waits for a response. This message is picked up by the processing thread in RemoteVstPlugin and posted to the main UI thread. Usually the UI thread then receives the message, saves a chunk to a temporary file and sends a response. However, in the case of the crash, the main thread is processing some other window message (from what I've found, it always seems to be [`WM_SETCURSOR`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms648382(v=vs.85).aspx)). Messages such as this one, if not handled by a window's WindowProc, are instead sent to the parent window. Thus if the plugin doesn't handle WM_SETCURSOR, it is forwarded to LMMS and the OS waits for the message to be processed. This is the deadlock; LMMS's UI thread is blocking on RemoteVstPlugin's, and vice versa.

This PR fixes the issue by passing `true` as a second argument to `RemotePluginBase::waitForMessage` in any method which may be called from LMMS's UI thread, causing LMMS to continue to process events while waiting for the reply. Note that this does not need to be done for ZynAddSubFx, as it runs in a separate window and thus will not forward such window messages to LMMS.

This does, however, introduce a new issue, where the UI thread may attempt to communicate with RemoteVstPlugin in response to a new event while still waiting on a reply to an earlier message. This will cause the UI thread to attempt to acquire the plugin mutex twice and thus block on itself. This isn't an issue usually, as such events tend to be user-input driven and the message loop in `RemotePluginBase::waitForMessage` specifies `QEventLoop::ExcludeUserInputEvents` as an argument to `QCoreApplication::processEvents`. However, autosaving is carried out on the UI thread and is triggered by a timer, and so may cause an issue here. Thus this PR moves autosaving to a new thread, and in general any automatic VST communication should be carried out in a different thread from now on so as to avoid deadlock. Additionally, this also causes message-waiting code that is sometimes called from the UI thread, and sometimes from other threads, to always run a message loop in `RemotePluginBase::waitForMessage`. This is unnecessary when not on the UI thread and will waste CPU cycles, so this PR also modifies `waitForMessage` to ignore its second argument if not called from the UI thread.

\#1008 is a similar issue, and has the same fix. Kontakt, when loading settings back in, attempts to display a modal dialog. This appears to involve dispatching a message to the VST's parent window, although I have been unable to discover which message this is as it is dispatched and remains undelivered entirely inside the Windows kernel. Nevertheless, the previous changes have fixed this and projects using Kontakt can now be loaded successfully.

This PR also fixes another VST-related hang. Currently, RemoteVstPlugin receives messages in its processing thread, and, unless these are messages designed to be processed in real time (i.e. MIDI events and sound processing), they are sent to be handled by the UI thread by wrapping them in a Windows user message and passing it to `PostThreadMessage` along with a handle to the UI thread. The UI thread then processes these messages in its message loop as it receives them. However, there is a problem: [thread messages are eaten by modal loops](https://blogs.msdn.microsoft.com/oldnewthing/20050426-18/?p=35783). Such modal loops are how Windows implements modal dialogs; thus any message that is sent while a modal dialog is open will disappear, causing a hang if LMMS is waiting for a reply. This can be seen, for example, on 1.2.0-rc3 by opening a modal dialog on a VST plugin (e.g. by clicking Kontakt's 'Options' button), then leaving it open while the autosave kicks in. Autosaving has been moved to a new thread, but there are other messages from which LMMS expects a reply, so the problem should be addressed. This is fixed by using a standard WindowProc as a callback. The dummy window currently used for receiving timer events has been repurposed to receive all messages sent to the UI thread, and the `PostThreadMessage` calls have been replaced by `PostMessage`. There are a couple of new synchronization constructs introduced as it is now possible for the VST to re-call the WindowProc whilst handling a message.

Now fixes #2784, fixes #2157, fixes #1468 too.

### DIFF
--- a/include/MainWindow.h
+++ b/include/MainWindow.h
@@ -29,6 +29,7 @@
 #include <QtCore/QTimer>
 #include <QtCore/QList>
 #include <QMainWindow>
+#include <QThread>
 
 #include "ConfigManager.h"
 #include "SubWindow.h"
@@ -246,6 +247,13 @@ signals:
 	void periodicUpdate();
 	void initProgress(const QString &msg);
 
+} ;
+
+class AutoSaveThread : public QThread
+{
+	Q_OBJECT
+public:
+	void run();
 } ;
 
 #endif
--- a/include/RemotePlugin.h
+++ b/include/RemotePlugin.h
@@ -424,6 +424,7 @@ enum RemoteMessageIDs
 	IdChangeSharedMemoryKey,
 	IdChangeInputCount,
 	IdChangeOutputCount,
+	IdChangeInputOutputCount,
 	IdShowUI,
 	IdHideUI,
 	IdSaveSettingsToString,
@@ -917,6 +918,15 @@ public:
 	{
 		m_outputCount = _i;
 		sendMessage( message( IdChangeOutputCount ).addInt( _i ) );
+	}
+
+	void setInputOutputCount( int i, int o )
+	{
+		m_inputCount = i;
+		m_outputCount = o;
+		sendMessage( message( IdChangeInputOutputCount )
+				.addInt( i )
+				.addInt( o ) );
 	}
 
 	virtual int inputCount() const
--- a/include/RemotePlugin.h
+++ b/include/RemotePlugin.h
@@ -1072,6 +1072,14 @@ RemotePluginBase::message RemotePluginBase::waitForMessage(
 							const message & _wm,
 							bool _busy_waiting )
 {
+#ifndef BUILD_REMOTE_PLUGIN_CLIENT
+	if( _busy_waiting )
+	{
+		// No point processing events outside of the main thread
+		_busy_waiting = QThread::currentThread() ==
+					QCoreApplication::instance()->thread();
+	}
+#endif
 	while( !isInvalid() )
 	{
 #ifndef BUILD_REMOTE_PLUGIN_CLIENT
--- a/plugins/vst_base/RemoteVstPlugin.cpp
+++ b/plugins/vst_base/RemoteVstPlugin.cpp
@@ -44,6 +44,10 @@
 
 #ifdef LMMS_BUILD_LINUX
 
+#ifndef NOMINMAX
+#define NOMINMAX
+#endif
+
 #ifndef O_BINARY
 #define O_BINARY 0
 #endif
@@ -67,6 +71,7 @@
 
 
 #include <vector>
+#include <queue>
 #include <string>
 
 
@@ -113,7 +118,7 @@ class RemoteVstPlugin;
 
 RemoteVstPlugin * __plugin = NULL;
 
-DWORD __GuiThreadID = 0;
+HWND __MessageHwnd = NULL;
 
 
 
@@ -244,8 +249,38 @@ public:
 		pthread_mutex_unlock( &m_pluginLock );
 	}
 
+	inline bool isProcessing() const
+	{
+		return m_processing;
+	}
+
+	inline void setProcessing( bool processing )
+	{
+		m_processing = processing;
+	}
+
+	inline void queueMessage( const message & m ) {
+		m_messageList.push( m );
+	}
+
+	inline bool shouldGiveIdle() const
+	{
+		return m_shouldGiveIdle;
+	}
+
+	inline void setShouldGiveIdle( bool shouldGiveIdle )
+	{
+		m_shouldGiveIdle = shouldGiveIdle;
+	}
+
+	void idle();
+	void processUIThreadMessages();
+
 	static DWORD WINAPI processingThread( LPVOID _param );
-	static DWORD WINAPI guiEventLoop( LPVOID _param );
+	static bool setupMessageWindow();
+	static DWORD WINAPI guiEventLoop();
+	static LRESULT CALLBACK messageWndProc( HWND hwnd, UINT uMsg,
+						WPARAM wParam, LPARAM lParam );
 
 
 private:
@@ -303,6 +338,10 @@ private:
 	bool m_initialized;
 
 	pthread_mutex_t m_pluginLock;
+	bool m_processing;
+
+	std::queue<message> m_messageList;
+	bool m_shouldGiveIdle;
 
 
 	float * * m_inputs;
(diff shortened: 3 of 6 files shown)


zonkmachine:
I've tested the auto-save function. It's running like before. :+1: 
I don't use a whole lot of vst stuff so someone else will have to test this.

@LMMS/developers Maybe bump this for RC4

PhysSong:
>Maybe bump this for RC4

Good Idea. We should let users test it before stable 1.2.0 release.

tresf:
@DomClark thanks for this patch.  I'm not sure if we have any developers familiar with this logic to be able to properly review so as @zonkmachine and @PhysSong have suggested this is going to get it's best approval if we merge and release with the next release candidate -- RC4.

@DomClark are there any items left todo or is this ready?  The code quality and format are top notch and thank you for researching and fixing the `winegcc` issue via 5c1e48d.

I vote we merge this to `stable-1.2` once @DomClark gives the nod.

DomClark:
Latest commit fixes #2784, #2157. `RemoteVstPlugin::updateInOutCount` is synchronized with `RemoteVstPlugin::process` so shared memory is never freed during audio processing. A flag is cleared in `RemoteVstPlugin::updateInOutCount` to indicate shared memory is invalid, then is set again upon receiving the new shared memory key from LMMS. The handling of the `IdChangeSharedMemoryKey` message has been moved to the processing thread to ensure the pointer to shared memory is changed outside of any call to `RemotePluginClient::doProcessing` (I feel this is clearly better than the alternative of continuing to have it sent to the UI thread and synchronizing on that anyway). Also, `RemotePluginClient::setInputOutputCount` has been added as an alternative to `setInputCount` and `setOutputCount` so the shared memory is only reallocated once, stopping the validity flag being set prematurely. (As a side effect, this fixes #1468, which was due to trying to attach the shared memory reallocated by `setInputCount` when `setOutputCount` had already started reallocating a second time.)

zonkmachine:
> Fixes #1008, #1183, #3274.

If you want to automatically close multiple issues on merge they each need [their own keyword](https://help.github.com/articles/closing-issues-using-keywords/#closing-multiple-issues).

DomClark:
Ah, fixed it, thank you. I'm happy with the code now, you can merge when you're ready.

Umcaruje:
I don't think I've ever seen so many issues closed by one PR. Amazing job @DomClark

Issue #3747, opened 2017-08-03
Some MIDI events(ex. sustain pedal) should be ignored while exporting


PhysSong:
To reproduce:

1. Enable MIDI input in instrument window.
2. Put some notes.
3. Hold sustain pedal.
4. Export project.

Result: sustain pedal affects to the export, which is undesirable.
Tested with latest master.

zonkmachine:
You don't need to export the project. Swap step 4 above with, `4. Play project`.
This behaviour is already there in 1.0.3

PhysSong:
Even more crazy thing: If you press keys from your MIDI keyboard while exporting, they **will be** rendered.

Umcaruje:
Tbh all midi events should be completely ignored while exporting. I had to re-render a lot of times when I actually accidentally pressed some keys on my midi keyboard while exporting

zonkmachine:
So, something like this...
```cpp
void InstrumentTrack::processInEvent( ... )
{
    if( Engine::getSong()->isExporting() )
    {
        return;
    }
...

```

zonkmachine:
I pulled the above code in https://github.com/LMMS/lmms/pull/3778. It's a partial fix and a good compromise for 1.2.0 .

PR #3778, opened 2017-08-26
Suppress midi in events on export

zonkmachine:
Addresses https://github.com/LMMS/lmms/issues/3747. This suppresses midi events (sustain pedal, key press) triggered during export. It does however not clear out midi events in effect while starting the export.

### DIFF
--- a/src/tracks/InstrumentTrack.cpp
+++ b/src/tracks/InstrumentTrack.cpp
@@ -237,6 +237,11 @@ MidiEvent InstrumentTrack::applyMasterKey( const MidiEvent& event )
 
 void InstrumentTrack::processInEvent( const MidiEvent& event, const MidiTime& time, f_cnt_t offset )
 {
+	if( Engine::getSong()->isExporting() )
+	{
+		return;
+	}
+
 	bool eventHandled = false;
 
 	switch( event.type() )


PhysSong:
If sustin pedal have been pressed before export, it will be exported. Before rendering, all notes should be silenced and sustain pedal should be marked as released.

zonkmachine:
> If sustin pedal have been pressed before export, it will be exported.

Which this PR doesn't take care of. It's a partial fix as mentioned [here](https://github.com/LMMS/lmms/issues/3747#issuecomment-325161279). I don't have time to dive into the issue more.

zonkmachine:
I'm merging this tomorrow if there are no other complaints. https://github.com/LMMS/lmms/issues/3747 isn't marked for 1.2.0 and shouldn't be. This PR will make the issue less prominent.

PhysSong:
I'm going to test this after my works for rc4 finish. If I find any regressions, I'll request changes. Otherwise, it's good:+1:

zonkmachine:
> Tested. Just a note:
> If you press keys/sustain pedal before exporting, there are no ways to correct during the export.

OK. I think I've already commented on this.

Merging :whale:

PR #3779, opened 2017-08-27
Fix some memory leaks

lukas-w:


src/core/Song.cpp
```diff
@@ -1515,36 +1514,31 @@ void Song::removeController( Controller * controller )
...
+	return ( m_errors.length() > 0 );
 }
 
 
 
-QString* Song::errorSummary()
+QString Song::errorSummary()
 {
-	QString* errors = new QString();
+	QString errors = m_errors.join('\n') + '\n';
```
irrenhaus3: Constructor of QString expects a const char* here. Change the '\n's to "\n" to make them string literals.

lukas-w: Yup, although It's the `join` function that expects `const char*`, not the constructor. Appears to be a Qt5-only feature.

lukas-w: Fixed via 670ca4f

Issue #777, opened 2014-05-27
Segfault when job queue is full


diizy:
Just documenting the backtrace here for future reference

```
#0  operator= (this=0xb32ca1a0, value=0xaac05e90) at /usr/include/qt4/QtCore/qbasicatomic.h:182
No locals.
#1  operator= (value=0xaac05e90, this=0xb32ca1a0) at /usr/include/qt4/QtCore/qatomic.h:145
No locals.
#2  MixerWorkerThread::JobQueue::addJob (this=0x82ad700, _job=0xaac05e90)
    at /home/vesa/tmp/lmms-ui/src/core/MixerWorkerThread.cpp:53
No locals.
#3  0x08173e57 in addJob (_job=<optimized out>) at /home/vesa/tmp/lmms-ui/include/MixerWorkerThread.h:86
No locals.
#4  fillJobQueue<QList<PlayHandle*> > (_opMode=MixerWorkerThread::JobQueue::Static, _vec=...)
    at /home/vesa/tmp/lmms-ui/include/MixerWorkerThread.h:98
        it = <optimized out>
#5  Mixer::renderNextBuffer (this=0x89d1ed8) at /home/vesa/tmp/lmms-ui/src/core/Mixer.cpp:380
        timer = {begin = {tv_sec = 1401133078, tv_usec = 755525}}
        it_rem = {i = 0xaac03298}
        last_metro_pos = {<MidiTime> = {m_ticks = -1, static s_ticksPerTact = 192}, m_timeLine = 0x0, 
          m_timeLineUpdate = true, m_currentFrame = 0}
        p = {<MidiTime> = {m_ticks = <optimized out>, static s_ticksPerTact = 192}, m_timeLine = <optimized out>, 
          m_timeLineUpdate = 152, m_currentFrame = 0}
#6  0x08174437 in Mixer::fifoWriter::run (this=0x8e9ed80) at /home/vesa/tmp/lmms-ui/src/core/Mixer.cpp:952
        buffer = 0xaac020e8
        b = <optimized out>
#7  0xb696eeb0 in ?? () from /usr/lib/i386-linux-gnu/libQtCore.so.4
No symbol table info available.
#8  0xb7701d4c in start_thread () from /lib/i386-linux-gnu/libpthread.so.0
No symbol table info available.
#9  0xb6186bae in clone () from /lib/i386-linux-gnu/libc.so.6
No symbol table info available.
```

lukas-w:
[...]

PR #3783, opened 2017-08-30
Fix "out of buffers" crash

### LINKED ISSUE #3223: LMMS crash when open one of my project

lukas-w:
Fixes #3223 

Replaces `BufferManager`'s implementation with simple calls to `MemoryManager::alloc. Re-use buffers where they are allocated (AudioPort.cpp & PlayHandle.cpp). I don't experience any performance regression from this change.

My initial approach to fix #3223 was to re-implement `BufferManager` using a lock-free stack, see b6421d4. However, I couldn't get Travis to build with it, and I realized `BufferManager::acquire` is only used in two places that can be optimized easily.


src/core/BufferManager.cpp
```diff
@@ -86,43 +59,6 @@ void BufferManager::clear( surroundSampleFrame * ab, const f_cnt_t frames,
 
 void BufferManager::release( sampleFrame * buf )
 {
-	if (buf == nullptr) return;
-	int i = s_releasedIndex.fetchAndAddOrdered( 1 );
-	s_released[ i ] = buf;
-	//qDebug( "released buffer: %p - index %d", buf, i );
+	MM_FREE(buf);
```
Umcaruje: I'm a nitpick, but space missing here too. Thanks for fixing the other lines.

lukas-w:
@Umcaruje I fixed some inconsistencies. However, I'd like to avoid touching any code just for the sake of fixing code format, because it makes reviewing commit history (especially using `git blame`) a lot harder.

include/MemoryManager.h
```diff
@@ -109,17 +109,17 @@ struct MmAllocator
...
 	}
 
-	void deallocate(T* p, std::size_t)
+	void deallocate( T* p, std::size_t )
 	{
 		MemoryManager::free( p );
 	}
 
-	typedef std::vector<T, MmAllocator<T> > vector;
+	typedef std::vector<T, MmAllocator<T>> vector;
```
Umcaruje: looks like the space here was removed again, have an error while compiling.
```
error: ‘>>’ should be ‘> >’ within a nested template argument list
```

lukas-w: Right, should have remembered from 73728d79da858b2ee07db5df4e1bd032bdd89911

Umcaruje:
The original project from the bug report still crashes for me with this PR. backtrace:
```gdb
Thread 11 "QThread" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fff859a0700 (LWP 7287)]
0x0000000000579116 in QGenericAtomicOps<QBasicAtomicOps<8> >::storeRelease<ThreadableJob*, ThreadableJob*> (newValue=0x7fffe7ff8f48, _q_value=@0xffffffff40998c60: <error reading variable>)
    at /usr/include/x86_64-linux-gnu/qt5/QtCore/qgenericatomic.h:111
111	        *static_cast<volatile T *>(&_q_value) = newValue;
(gdb) bt full
#0  0x0000000000579116 in QGenericAtomicOps<QBasicAtomicOps<8> >::storeRelease<ThreadableJob*, ThreadableJob*> (newValue=0x7fffe7ff8f48, _q_value=@0xffffffff40998c60: <error reading variable>)
    at /usr/include/x86_64-linux-gnu/qt5/QtCore/qgenericatomic.h:111
No locals.
#1  QBasicAtomicPointer<ThreadableJob>::storeRelease (this=0xffffffff40998c60, newValue=0x7fffe7ff8f48) at /usr/include/x86_64-linux-gnu/qt5/QtCore/qbasicatomic.h:259
No locals.
#2  0x0000000000578f14 in QAtomicPointer<ThreadableJob>::operator= (this=0xffffffff40998c60, other=...) at /usr/include/x86_64-linux-gnu/qt5/QtCore/qatomic.h:178
No locals.
#3  0x0000000000578968 in MixerWorkerThread::JobQueue::addJob (this=0x9d1a60 <MixerWorkerThread::globalJobQueue>, _job=0x7fffe7ff8f48) at /home/umcaruje/github/lmmstest/src/core/MixerWorkerThread.cpp:57
No locals.
#4  0x0000000000555538 in MixerWorkerThread::addJob (_job=0x7fffe7ff8f48) at /home/umcaruje/github/lmmstest/include/MixerWorkerThread.h:87
No locals.
#5  0x00000000005753fa in MixerWorkerThread::fillJobQueue<QList<PlayHandle*> > (_vec=..., _opMode=MixerWorkerThread::JobQueue::Static) at /home/umcaruje/github/lmmstest/include/MixerWorkerThread.h:99
        it = {i = 0x7fff7806cf78}
#6  0x0000000000571fe5 in Mixer::renderNextBuffer (this=0xba56c0) at /home/umcaruje/github/lmmstest/src/core/Mixer.cpp:438
        last_metro_pos = {<MidiTime> = {m_ticks = -1, static s_ticksPerTact = 192}, m_timeLine = 0x0, m_currentFrame = 0}
        song = 0xe2d990
        currentPlayMode = Song::Mode_PlaySong
        p = {<MidiTime> = {m_ticks = 8212, static s_ticksPerTact = 192}, m_timeLine = 0x2a9df60, m_currentFrame = 312.87793}
        playModeSupportsMetronome = true
        it_rem = {i = 0x7fff780013b0}
        fxMixer = 0xea2800
#7  0x0000000000573c3f in Mixer::fifoWriter::run (this=0xb1f3e0) at /home/umcaruje/github/lmmstest/src/core/Mixer.cpp:1102
        buffer = 0x7fff780257e0
        b = 0xe0ce10
        frames = 256
#8  0x00007ffff4b0c7be in ?? () from /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
No symbol table info available.
#9  0x00007ffff7bc16ba in start_thread (arg=0x7fff859a0700) at pthread_create.c:333
        __res = <optimized out>
        pd = 0x7fff859a0700
        now = <optimized out>
        unwind_buf = {cancel_jmp_buf = {{jmp_buf = {140735434852096, 8462573384777015085, 1, 140737488344511, 140735434852800, 11662208, -8462806189127819475, -8462555276406087891}, mask_was_saved = 0}}, priv = {pad = {0x0, 0x0, 0x0, 
              0x0}, data = {prev = 0x0, cleanup = 0x0, canceltype = 0}}}
        not_first_call = <optimized out>
        pagesize_m1 = <optimized out>
        sp = <optimized out>
        freesize = <optimized out>
        __PRETTY_FUNCTION__ = "start_thread"
#10 0x00007ffff44223dd in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:109
No locals.
```

lukas-w:
@Umcaruje That's an unrelated crash (job queue is full), see https://github.com/LMMS/lmms/issues/777#issuecomment-325200871.

Umcaruje:
Ah, ok, good, this fixes the out of buffers issue and I can't seem to pick up any issues with it. This is ready to merge.

src/core/PlayHandle.cpp
```diff
@@ -22,18 +22,23 @@
  *
  */
  
-#include "PlayHandle.h"
 #include "BufferManager.h"
+#include "Engine.h"
+#include "Mixer.h"
+#include "PlayHandle.h"
```
PhysSong: Why this line have been moved from above? It should stay above `#include "BufferManager.h"` according to our convention, and that way seems more consistent for me.

include/MemoryManager.h
```diff
@@ -103,6 +103,25 @@ class EXPORT MemoryManager
...
+	}
+
+	void deallocate( T* p, std::size_t )
+	{
+		MemoryManager::free( p );
+	}
+
+	typedef std::vector< T, MmAllocator<T> > vector;
+};
+
```
PhysSong: These parts aren't used. Are there some reasons you leave them in our codebase?
Some more inconsistencies: you mixed `< foo >` and `<foo>`, it makes the code looks inconsistent.

PhysSong:
It seems to fix the issue clearly, anyway.

[test.mmp.zip](https://github.com/LMMS/lmms/files/1313663/test.mmp.zip)

This project file shows the bug clearly and it seems to be fixed by this PR.

src/core/PlayHandle.cpp
lukas-w: No reason, probably an unintentional change.

include/MemoryManager.h
lukas-w: > Are there some reasons you leave them in our codebase?

I plan on using them at a later point in time. Using an [Allocator](http://en.cppreference.com/w/cpp/concept/Allocator) class is the standard C++ way of having custom allocators, and should be preferred over using the macros `MM_ALLOC` and `MM_FREE`.

> you mixed < foo > and <foo>, it makes the code look inconsistent.

This has already been discussed by @Umcaruje and me in the comments here. C++03 doesn't allow `>>` for nested template arguments (it always interprets it as the stream operator `>>`). We can't use `>>` here, because the headers is included by plugins that are compiled without C++11 support. See d1aa39bfd194b0645aeadb647808c7ed15ce8f7e.

lukas-w: > if you are going to use this later, I think vector is not very good name

Well it's a vector, so why not call it `vector`? What would you call it?

PhysSong: You're right, but you used both `template<typename T>` and `template< class U >` above. Is that intentional?

lukas-w: Ah, no that's not intentional.

PhysSong:
@lukas-w I think it is ready to merge. However, I suggest 
fixing inconsistent and unintentional things before you merge. Would [this pach](https://gist.github.com/PhysSong/52346bb1d02aa6c98d97a765768e20ad) be fine?

lukas-w:
> Would this pach be fine?

@PhysSong Sure. Feel free to apply it if you find the time before I do.

lukas-w:
Thanks @PhysSong. Merging.

src/core/PlayHandle.cpp
```diff
@@ -60,6 +65,10 @@ void PlayHandle::doProcessing()
 
 void PlayHandle::releaseBuffer()
 {
-	BufferManager::release( m_playHandleBuffer );
```
PhysSong: @lukas-w Why did you move this to destructor?

lukas-w: @PhysSong Because I moved buffer allocation to the constructor. The buffer's lifetime matches the `PlayHandle`'s lifetime now, instead of being allocated everytime `doProcessing` is called.

PR #3497, opened 2017-04-10
Support VST windows on Qt5

jasp00:
This fixes #2855, but subwindows are not used. It is a simpler alternative to #3475, uses separate windows like ZynAddSubFX, and unifies Windows and Linux behavior. Since there is no embedding, more clients are supported. The inability to embed native windows properly is likely because of Qt; the embedding implementation is left, just in case Qt is fixed.

This is the continuation of #2856, but disables FluidSynth because of #3490. You may reenable it prior to merging, but I recommend you not to do so.

PR #3475, opened 2017-04-01
Add X11EmbedContainer Qt5 port

lukas-w:
Alternative solution for issue #2855 (other PRs trying to solve the same issue are ~~#2856~~ #3497  and #3295). I ported most of `QX11EmbedContainer` from Qt4 to Qt5 for this. I haven't looked at the code in detail, so it's quite possible most of it is not even used in our case.

It'd be great if we could get some more test results for this. :)

PR #3786, opened 2017-09-01
Add all the Qt5 Linux VST implementations

### LINKED ISSUE #2855: VSTi plugins don't show in Subwindow on Qt5 build (probably only X11)

lukas-w:
**Current AppImage for testing:** https://transfer.sh/pIL0f/lmms-1.2.0-rc4.106-linux-x86_64.AppImage

This PR adds the Qt5+Linux implementations from #3295, #3475 and #3497, as suggested in https://github.com/LMMS/lmms/pull/3497#issuecomment-315554077.

Introduces CMake options for specifying the implementation to use at compile time:

* Turning off `LMMS_EMBED_VST` results in VSTs being opened in separate window (#3497)
* Turning on `LMMS_EMBED_VST` will embed VSTs in subwindows. This is the default.
  * Turning on `LMMS_EMBED_VST_X11` will use `QX11EmbedContainer` or the [Qt5 port](https://github.com/lukas-w/qt5-x11embed) of it (#3475).
  * Turning off `LMMS_EMBED_VST_X11` will use Qt5's new `QWidget::createWindowContainer` method (unavailable with Qt4). (#3295) 

The last option has various bugs on some Qt versions. However, when I re-tested it recently with Qt 5.9.1, it worked just as well as #3475, suggesting that Qt's internal XEMBED implementation was fixed. So it might supersede #3475 eventually.

It would be great if we could make this an option that's user-configurable at runtime, ~~but I don't have any time to implement that currently.~~ Done.

Fixes #2855, Closes #3475, closes #3497.


PhysSong:
Are there some ways to make them configurable on runtime, not on the compilation time? People need to re-compile LMMS if they want to switch to other VST implementations, which might be inconvenient.

tresf:
> People need to re-compile LMMS if they want to switch to other VST implementations, which might be inconvenient.

@lukas-w has already acknowledged this:

> It would be great if we could make this an option that's user-configurable at runtime, but I don't have any time to implement that currently.

qnebra:
And maybe add libqt5x11extras5-dev package to a LMMS depencies list, if this isn't installed on system this PR cannot be builded.

lukas-w:
@qnebra `libqt5x11extras5-dev`is already present in the Travis script, we just need to remember adding it to the wiki when this PR is merged.

It seems your system's lacking `qtbase5-private-dev` (assuming you're on Ubuntu).

qnebra:
@lukas-w Yep, I am on Ubuntu and qtbase5-private-dev isn't installed.

plugins/vst_base/RemoteVstPlugin.cpp
```diff
@@ -1807,6 +1851,13 @@ DWORD WINAPI RemoteVstPlugin::guiEventLoop( LPVOID _param )
 	while( quit == false && GetMessage( &msg, NULL, 0, 0 ) )
 	{
 		TranslateMessage( &msg );
+
+		if( msg.message == WM_SYSCOMMAND && msg.wParam == SC_CLOSE )
+		{
+			_this->destroyEditor();
```
PhysSong: @lukas-w When I merged this to my local branch(up-to-date `stable-1.2`) it fails to build. The failure is because of fe98a9a by @DomClark. It seems to work after I change `_this` to `__plugin`, but I'm not sure this is the right way...

karmux:
cmake fails because of sndio when I merge this pull request.

PhysSong:
@karmux Could you check environment variables and run CMake again? This works fine on my side.

plugins/vst_base/RemoteVstPlugin.cpp
DomClark: Previously `_this` was passed as an argument to `guiEventLoop`, but after I moved the window message handling code to `messageWndProc`, there was no need for it any more so I removed it. Using `__plugin` is safe here, and has the same value `_this` would have done, so the change is correct, but this code may well be better off with all the other message handling code in `messageWndProc`. I have no experience with Linux programming, so I may well be wrong, but on Windows other modal loops can be run, e.g. by modal dialogs or window drag-resizing, in which case the messages are still sent to the WindowProc but will not be picked up by code in the main modal loop. This could lead to your message being missed if it is dispatched while a different loop is currently in control, unless the code is moved over.

PhysSong: @DomClark Thanks for explanation. :+1:

karmux:
Couldn't `cmake` because `src/3rdparty/qt5-x11embed` was empty.
I finally figured out that I had to run `git submodule update --init`.

PhysSong:
Works great for VST instruments, but seems to fail to embed VST plug-in correctly. @lukas-w Can you reproduce this?
GCC 7.2.0, Qt 5.5.1, Wine-Staging 2.16, Ubuntu 16.04 x86_64.

lukas-w:
> Works great for VST instruments, but seems to fail to embed VST plug-in correctly

You mean VST effects? I assume you're using the default behaviour. Will have a look. 👍

PhysSong:
Yes. X11embed one. createWindowContainer method is buggy on Qt 5.5 so I was unable to test that one.

lukas-w:
@PhysSong Confirmed. If I recall correctly, they did work in #3475, so I guess I broke it during merging.

lukas-w:
Fixed.

PhysSong:
Good. I'll test it soon. :+1:

PhysSong:
Now it works.

karmux:
Now all VST instruments and effects are displayed with some offset inside window. About +5x and -20y.

screenshot_20170925_000831

lukas-w:
> Seems to be Qt 5.7 bug with `QWidget::createWindowContainer`. LMMS_EMBED_VST=OFF works well for me.

Did you explicitly turn off `LMMS_EMBED_VST_X11`? It should default to `ON` on Linux. Does `-DLMMS_EMBED_VST=ON -DLMMS_EMBED_VST_X11=ON` work for you?

karmux:
@lukas-w no then offset is when using `QX11EmbedContainer`. With `QWidget::createWindowContainer` VST UI just dissapear after few clicks and never comes back.

wine-2.15 and Qt 5.7.1.

lukas-w:
@karmux Did your system change since you tested https://github.com/LMMS/lmms/pull/3475#issuecomment-316520179? Can you reproduce the same offset problem with #3475, or just with this PR? Could you give more detail about your system (distro, desktop, window manager)?

karmux:
@lukas-w `qt5-vst-2` branch has no offset or other glitches or bugs. My system hasn't changed since then. I'm using Kubuntu 17.04, KDE, Qt 5.7.1 and Wine-Staging 2.15.

PhysSong:
When I merged latest `stable-1.2` into this, build errors in non-linux platforms:
```
CMake Error: Cannot determine link language for target "qx11embedcontainer".
CMake Error: CMake can not determine linker language for target: qx11embedcontainer
```

PhysSong:
Again, this needs merge fix. See the [Travis-CI log](https://travis-ci.org/PhysSong/lmms/jobs/291395594#L1201-L1202)(on my fork).

lukas-w:
@PhysSong We're well aware, thank you. Repeating your comment and deleting your old one isn't helping. If you have time to investigate, please do.

This PR also needs to have the embed method made a selectable option at run-time. #2855 appears to be the biggest issue blocking 1.2, so any help is appreciated. Anyone who has the time, feel free to work on it and open a PR against this PR's branch `fix/qt5-vst`.

PhysSong:
> Repeating your comment and deleting your old one isn't helping.

Sorry. The deletion was unintended.

@lukas-w I'm already on this. Once my work is done, I'll open a PR. :smile:

PhysSong:
Solved: #3873 adds `ADD_SUBDIRECTORY(3rdparty)` in the beginning of `src/CMakeLists.txt`, whereas this PR adds that after some lines. If it is added prior to `SET(CMAKE_AUTOMOC ON)`, it generates error.
Doing a right merge is the solution.

plugins/vst_base/VstPlugin.cpp
```diff
@@ -629,6 +568,60 @@ void VstPlugin::idleUpdate()
...
+			createUI( NULL, false );
+		}
+
+		QWidget * w = pluginWidget();
+		if( w )
+		{
+#		ifdef LMMS_BUILD_WIN32
+			// hide sw, plugin window wrapper on win32
+			// this is obtained from pluginWidget()
+			if( isEffect )
```
PhysSong: Here `isEffect` is undefined. I think you should store it from `createUI()` or somehow drop it if possible.

src/gui/SetupDialog.cpp
```diff
@@ -137,12 +139,14 @@ SetupDialog::SetupDialog( ConfigTabs _tab_to_open ) :
 	m_displayWaveform(ConfigManager::inst()->value( "ui",
 						   "displaywaveform").toInt() ),
 	m_disableAutoQuit(ConfigManager::inst()->value( "ui",
-						   "disableautoquit").toInt() )
+						   "disableautoquit").toInt() ),
+	m_vstEmbedMethod(ConfigManager::inst()->value( "ui",
+						   "vstembedmethod", "xembed"))
```
PhysSong: Is the default value `"xembed"` fine for non-linux platforms?

src/gui/SetupDialog.cpp
```diff
@@ -1044,6 +1063,8 @@ void SetupDialog::accept()
 	ConfigManager::inst()->setValue( "ui", "disableautoquit",
 					QString::number( m_disableAutoQuit ) );
 	ConfigManager::inst()->setValue( "app", "language", m_lang );
+	ConfigManager::inst()->setValue( "ui", "vstembedmethod",
+					m_vstEmbedComboBox->currentData().toString() );
```
PhysSong: `QComboBox::currentData()` was introduced in Qt 5.2 and breaks Qt4 compatibility. Are there any alternatives? Or just drop Qt4 support now?

PhysSong:
And `#include <QWindow>` should be included for Windows build(currently not, because of the `LMMS_BUILD_LINUX` condition).

src/gui/SetupDialog.cpp
PhysSong: If I understand correctly, `ConfigManager::inst()->vstEmbedMethod()` could be used instead.

src/gui/SetupDialog.cpp
PhysSong: I found an alternative: `m_vstEmbedComboBox->itemData(m_vstEmbedComboBox->currentIndex()).toString()`.

lukas-w: I suggest we drop Qt4 support with this PR. It's too much work to maintain both versions, and the VST issue is the last remaining deal-breaker with Qt5.

src/gui/SetupDialog.cpp
lukas-w: `ConfigManager::inst()->vstEmbedMethod()` is better here indeed, thanks.

tresf:
> I suggest we drop Qt4 support with this PR. It's too much work to maintain both versions, and the VST issue is the last remaining deal-breaker with Qt5.

I was under the impression we were going to keep Qt4 available in `stable-1.2` source and remove it from master.

lukas-w:
@tresf So was I, but I'm getting increasingly frustrated with the number of solutions we maintain here, so I don't feel like implementing Qt4 support in this PR just to have it removed in `master` immediately is worth the time. It looks like we'll be shipping 1.2 with Qt5 AppImages anyway.

tresf:
> It looks like we'll be shipping 1.2 with Qt5 AppImages anyway

What we ship and what we offer to the various distros is two separate things though.  The fact is that Qt4 works perfect and Qt5 needs workarounds.  I haven't reviewed the code and I don't want to minimize the efforts here, but killing off Qt4 for a feature that's working seems a bit drastic.  In my opinion, decoupling Qt4 going to hit several areas of code and should be done gracefully, no? 🤷‍♂️

gi0e5b06:
VST on Qt5 + wine 1.8.7-1ubuntu1 is working fine.
https://github.com/gi0e5b06/lmms/blob/master/plugins/vst_base/VstPlugin.cpp

BTW, we should not differentiate between effects and instruments. I suggest to put the vst effect into its own separate window (and eventually to bring the controls back to the lmms effect window).
- simpler code
- easier to move the vst in real windows it the user wants that
- easier to manage the sizes of the different windows

lukas-w:
@tresf 

> The fact is that Qt4 works perfect and Qt5 needs workarounds.

In my tests, this branch's "xembed" option works a lot better than the Qt4 implementation, which has always been buggy. I'll supply an AppImage soon for better testing (do our AppImages bundle Wine?).

> In my opinion, decoupling Qt4 (properly) is going to hit several areas of code and should be done gracefully, no?

Of course, I wouldn't remove Qt4-specific code on `stable-1.2`, we'd only do that on `master`. `stable-1.2` would merely get its Qt4 Travis builds disabled and official support for Qt4 dropped (maybe some CMake changes to make `-DWANT_QT5=ON` default).

Anyway, @PhysSong says he's working on Qt4 support, which renders this discussion irrelevant :)

PhysSong:
> Anyway, @PhysSong says he's working on Qt4 support, which renders this discussion irrelevant :)

Supporting Qt4 here doesn't require too many changes. A simple patch will make it work on Qt4 with disabled `QWidget::createWindowContainer()` method(`"qt"` embedding).
And I'll reintroduce an embedding method with native Win32 API on Windows. It can be implemented on LMMS' side(previously it required a Win32 hack in `RemoteVstPlugin`). I'm going to do some tests to find differences between two ways.
Embedding on LMMS' side will make the code much simpler than embedding on remote side. However, I can switch back to original "self-embedding" if the new method turns out to cause problems.

PhysSong:
I found two bugs while testing this PR.
First bug: LMMS crashes when pressing keys after removing VST effects, for every embedding methods.
I got a crash in `InstrumentTrackView::topLevelInstrumentTrackWindow()`. The plugin widget is deleted when removing effects, but the subwindow seems not.

Second bug: When I open a VST with "No embedding" mode and try to hide it after switching to any other embedding mode, nothing happens(VeSTige) or LMMS crashes(VST effects).
I think we should store the embedding method in `VstPlugin` or apply embedding setting change after restarting.

lukas-w:
@PhysSong Thanks for the feedback. Fixed the crash via a8311a7, stored embedding method in `VstPlugin` via f9f4d0c and added support for changing the embed method without restarting LMMS via 358a251.

PhysSong:
> Fixed the crash via a8311a7, stored embedding method in VstPlugin via f9f4d0c

Tested, cool!

> added support for changing the embed method without restarting LMMS via 358a251

Seems to work for VeSTige only. I'm not sure whether reloading already loaded plugins is good, or not.

FYI, I almost finished my work for Qt4 compatibility and Win32 API embedding.

PhysSong:
For some reason VST effect dialogs always have width around 250 pixel when I use Qt embedding. `setFixedSize(sizeHint())` seems to fix this, but I'm not sure why the bug occurs and is this solution correct.
Tested with Qt 5.5.1, but I wonder if this is fixed in later version of Qt.

PhysSong:
@lukas-w What should we do before merging this? Are there anything left other than testing?

karmux:
Window offsets that I had before are gone after pulling updates from this branch.

lukas-w:
@PhysSong I'd like to investigate the VST Effect dialog width issue first, and then upload another AppImage for more testing. Not sure where the width problem comes from...

PhysSong:
@lukas-w `QX11EmbedContainer` overrides `QWidget::minimumSizeHint` but `QWindowContainer` doesn't provide proper size hint. Can it be the reason for the weird size of VST effect dialogs?

lukas-w:
@PhysSong I'm not sure... We call `setFixedSize` on the plugin widget and the VST Effect layout should respect that. Anyway, for some reason ae488b89cd6cc824e1e92fc3c105676be964741f fixes it as well. Almost any change to the dialog's layout triggered the bug to disappear, so I suspect it's some strange Qt hick-up.

I pushed the width fix and one more Qt-embedding fix and will upload an AppImage shortly.

lukas-w:
AppImage: https://transfer.sh/hnqHw/lmms-1.2.0-rc4.106-linux-x86_64.AppImage

PhysSong:
> added support for changing the embed method without restarting

I'm not sure how existing VST windows should behave. I think this(reinstantiation) should be applied to VST effects for consistency if we keep this behavior. I wonder if user can enable/disable this behavior...

PhysSong:
In [qt5-x11embed/CMakeLists.txt](https://github.com/lukas-w/qt5-x11embed/blob/022b39a1d496d72eb3e5b5188e5559f66afca957/CMakeLists.txt#L1): 
```
cmake_minimum_required(VERSION 2.8.12 FATAL_ERROR)
```
@lukas-w As far as I understand, it will harden CMake dependency to 2.8.12 for Qt5 build, right? I think it won't matter too much(for example, Ubuntu 14.04 ships with both 2.8.12 and 3.5.1), but it should be clarified.

zonkmachine:
> AppImage: https://transfer.sh/pIL0f/lmms-1.2.0-rc4.106-linux-x86_64.AppImage

Tested. The few tracks I have with vst's in them works as expected.

lukas-w:
> I'm not sure how existing VST windows should behave. I think this(reinstantiation) should be applied to VST effects for consistency if we keep this behavior.

VST instruments re-load with the new embedding method. However, the way VST effects are implemented makes it difficult to have the same behaviour for them. Should we just remove support for it altogether?

lukas-w:
@PhysSong 

> As far as I understand, it will harden CMake dependency to 2.8.12 for Qt5 build, right?

No, not really. From the [docs](https://cmake.org/cmake/help/v3.0/command/cmake_minimum_required.html):

> The `FATAL_ERROR` option is accepted but ignored by CMake 2.6 and higher. It should be specified so CMake versions 2.4 and lower fail with an error instead of just a warning.

PhysSong:
> > As far as I understand, it will harden CMake dependency to 2.8.12 for Qt5 build, right?

> No, not really. From the docs:

Okay. I got it.
> However, the way VST effects are implemented makes it difficult to have the same behaviour for them.

I'll look into that if I have enough time.
> Should we just remove support for it altogether?

I think we should, if we can't implement that for VST effects.

PhysSong:
> However, the way VST effects are implemented makes it difficult to have the same behaviour for them. Should we just remove support for it altogether?

I'm not sure, but it may cause some issues to implement the same behavior for VST effects. I vote on removing. If that behavior is needed, it can be done later.

lukas-w:
@PhysSong Disabled automatic reloading via 19f48898a705f22b063036565ac5dd67f480fabe. 

Merge?

PhysSong:
:+1:

qnebra:
I tested this with latest possible build from this branch, my God, it is wonderful. No escaping, no glitches, presets work very well, add multiple VST to project, add multiple VST effects to VST instrument and all works.

lukas-w:
@qnebra Glad to hear that :) 

Merging 💣

qnebra:
@lukas-w Now it's time for really real testing on production.

PR #3789, opened 2017-09-03
Fix Vibed when the "P" parameter has non-zero value

### LINKED ISSUE #3663: Playing specific notes causes distortion when the "P" parameter is set to more than 0 in "Vibed".
**Before you re-produce:**
The volume of the distortion caused by this bug is extremely loud and causes clipping even if you lower the volume of the instrument plug-in or the master mixer fader. You should decrease the volume of your playback device to something low (e.g., 20%). The distortion causes the master fader to look like this:
image

To re-produce:
(1) Open a project.
(2) Add the default "Vibed" instrument plug-in.
(3) Set the value of the "P" parameter to more than 0 (e.g., 0.05).
(4) Play a note from the following list, these are only the notes I tested atm from C4 to B5 (two octaves): C4/G4/A#4/D#5/E5/B5.

Changing the "Octave", "Length" and "Detune" parameters changes which notes cause the distortion.
Changing any other parameter in the "plugin" tab doesn't change which notes cause the distortion.
Enabling the "Imp" parameter removes the distortion.

PhysSong:
Fixes #3663. Current code reads value data from wrong indexes.
To correct it, ` - 1` have been added.

### DIFF
--- a/plugins/vibed/vibrating_string.h
+++ b/plugins/vibed/vibrating_string.h
@@ -130,7 +130,7 @@ private:
 				offset =  ( m_randomize / 2.0f -
 						m_randomize ) * r;
 				_dl->data[i] = _scale *
-						_values[_dl->length - i] +
+						_values[_dl->length - i - 1] +
 						offset;
 			}
 			for( int i = _pick; i < _dl->length; i++ )


PhysSong:
Merge? Or change the base branch?

Issue #3795, opened 2017-09-07
Song editor mode is changed permanently with Ctrl key(master branch)


PhysSong:
To reproduce:
1. Set the song editior to draw mode.
2. Press Ctrl key, in order to change to select mode(temporarily).
3. Click any other subwindows(ex. piano roll, bb editor, etc.)
4. Release Ctrl. 

Now the song editor is in the select mode, which is undesirable. Note that it isn't there in 1.2 branch, since this behavior was introduced in #3649, which was merged into `master` branch.

PhysSong:
Tagging @BaraMGB as the author of that PR.

BaraMGB:
Thank you. I will look into this.

musikBear:
duplicate of piano-roll situation: 
Sti2end made a list over problem with icon-hangs/ changes here
https://github.com/LMMS/lmms/issues/906#issuecomment-48371989
Dont know if the issue is the same, but maby ?

BaraMGB:
Closed by https://github.com/LMMS/lmms/issues/3796

PR #3796, opened 2017-09-07
fix CTRL in songeditor by losing focus

BaraMGB:
fixes https://github.com/LMMS/lmms/issues/3795


PhysSong:
It's because of uninitialized `m_crtlAction`. Whenever it crashes, that have non-`NUL`L value in constructor.
Initializating it to `NULL` fixes crash :+1:

BaraMGB:
Thanks for investigation! I will add the initialization.

Issue #3777, opened 2017-08-26
Glitch with layered sustained notes


zonkmachine:
Got one more glitch with recording sustain that I unfortunately missed when testing. Some layered notes seem to finish early and the lost length is added to the following note. True both for chords and single notes played multiple times when under sustain.

Example. C-maj played struck twice:
![sustain](https://user-images.githubusercontent.com/6368949/29738776-58bf9b38-8a2d-11e7-8af5-f3ef3416c2db.png)

Should have been:
![sustain2](https://user-images.githubusercontent.com/6368949/29738779-621a45de-8a2d-11e7-890c-e693d40a0dd0.png)

@serdnab

PhysSong:
https://github.com/LMMS/lmms/blob/v1.2.0-rc3/src/gui/editors/PianoRoll.cpp#L3486
The if clause compares only the notes' keys, not their positions. So lengths can be mixed.

zonkmachine:
I can't replicate this. May have been fixed in https://github.com/LMMS/lmms/commit/81966fa1a68293a5dfa37408da993bc95ac8402d . Need to test this more though.

zonkmachine:
> https://github.com/LMMS/lmms/blob/v1.2.0-rc3/src/gui/editors/PianoRoll.cpp#L3486
> The if clause compares only the notes' keys, not their positions. So lengths can be mixed.

If it can be replicated I guess a fix then would be as easy as changing this line from:
https://github.com/LMMS/lmms/blob/afd22c8b74e23f6a89f2fac9ae596dd16979f4f7/src/gui/editors/PianoRoll.cpp#L3486

to:
```cpp
 if (it->key() == n.key() || it->pos() == n.pos())
```

PhysSong:
> `if (it->key() == n.key() || it->pos() == n.pos())`

Shouldn't it be `&&`?

zonkmachine:
That is what I meant. Tried it and it didn't record anything. :-/

PhysSong:
`InstrumentTrack::processInEvent()` doesn't save the time offset in `NotePlayHandle`s when received `MidiNoteOn`, that's why the `&&` doesn't work.
I think the fix is from #3803. One problem is that it assumes notes in a `Pattern` is always sorted by position. It might be fine because `Pattern::addNote()` always ensure that, at least for now.

zonkmachine:
OK. I think we're fine closing this then.

PR #3803, opened 2017-09-11
Fix recording of single streamed instruments(regression in #3774)

PhysSong:
Fixes https://github.com/LMMS/lmms/pull/3774#issuecomment-325090531, reverts #3774 and provides a new fix for #3757 using the solution in https://github.com/LMMS/lmms/issues/3757#issuecomment-320881503.
As a side effect, It works as a temporary solution for #3777. However, #3777 still needs a permanent solution like https://github.com/LMMS/lmms/pull/3774#issuecomment-327640947.

### DIFF
--- a/include/InstrumentTrack.h
+++ b/include/InstrumentTrack.h
@@ -237,6 +237,7 @@ private:
 	MidiPort m_midiPort;
 
 	NotePlayHandle* m_notes[NumKeys];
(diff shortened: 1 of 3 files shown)


zonkmachine:
I won't be able to test this as I'm away from home.

PhysSong:
Okay... Is there anyone who can test it?

PhysSong:
Merge?

PR #3805, opened 2017-09-12
Don't use VST-provided memory when loading chunks

### LINKED ISSUE #493: LMMS hangs when certain VST effects are reloaded in a saved project
(0.92 xp 32 bit)
I looked at one of the vst-instruments (Kairatune.dll) that lmms does work with. The vst loads correct in an empty vst-container, but after project is saved, it is imposible to reopen. So is any saved instrument made (.xpf) with the vst.
I believe the reason is the chunk in the vst definition
An analyse shows enourmous size of this entry:

Lines   1
Non-empty lines 1
Characters  1728819
Characters without blanks   1728819
Commented char count    0
Commented char count percent    0

the result is the same for both instrument .xpf, and for .mmp -eg same numbers
What could be the reason for lmms to generate a chunk of this size?
If that could be understood, lmms could perhaps load this and other vsts. that does not work now
(btw - teh same chunk in a Synth1 project is 3500 char in size.
(zipped files :http://www.sendspace.com/file/8mmpjj

DomClark:
Fixes #493. May potentially break some other plugins, see [this comment on the issue](https://github.com/LMMS/lmms/issues/493#issuecomment-328919040).

### DIFF
--- a/include/aeffectx.h
+++ b/include/aeffectx.h
@@ -101,6 +101,7 @@ const int effEditOpen = 14;
 const int effEditClose = 15;
 const int effEditIdle = 19;
 const int effEditTop = 20;
+const int effSetChunk = 24;
 const int effProcessEvents = 25;
 const int effGetEffectName = 45;
 const int effGetVendorString = 47;
--- a/plugins/vst_base/RemoteVstPlugin.cpp
+++ b/plugins/vst_base/RemoteVstPlugin.cpp
@@ -1385,20 +1385,7 @@ void RemoteVstPlugin::loadPresetFile( const std::string & _file )
 
 void RemoteVstPlugin::loadChunkFromFile( const std::string & _file, int _len )
 {
-	char * buf = NULL;
-
-	void * chunk = NULL;
-	// various plugins need this in order to not crash when setting
-	// chunk (also we let the plugin allocate "safe" memory this way)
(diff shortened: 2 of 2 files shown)


Umcaruje:
I tested this on Linux Qt4, this fixes the crash when loading the VST. I also opened Synth1, and it performed as expected. 

The plugin has another quirk on linux, it's GUI disappears when interacted with, but this was not introduced with this PR, so I believe it would deem an issue of it's own. (I'd like to test with qt5 first before making one though)
gifrecord_2017-09-18_221823

musikBear:
> The plugin has another quirk on linux

@Umcaruje Looks a lot like a re-draw issue, so the behaviour could be different with different graphic hw.
I see this in other GUIs, but there it appears as black blocks. They are fixed *on my hw*, by pushing the UI slowly into the edge of the screen, and bringing it back.

tresf:
@musikBear there may also be Wine logs that have some errors.  P.S. Stop using Windows XP.

zonkmachine:
> P.S. Stop using Windows XP.

@musikBear I think what @tresf is refering to here, more specifically, is a **later** OS (Windows Vista **not** included).

musikBear:
💸  is actually the issue.. Not laziness or incompetence 💸

PR #3808, opened 2017-09-14
Fixing few possible mem leaks for ladspa plugin effects loading.

devnexen:


lukas-w:
This should probably be submitted upstream as well: https://github.com/tomszilagyi/tap-plugins

PR #2826, opened 2016-06-06
MDI Subwindow Decoration for VeSTige instruments closes #2824

BaraMGB:
Can't test this. I fight with the `RemotePluginClient::shmget:` bug. 
Can someone test this please?

Regards

PR #3809, opened 2017-09-15
Don't let plugins make the main window transparent

DomClark:
Some plugins, e.g. Blue Cat's VSTs, have an opacity control which, in LMMS, makes the main window transparent. This PR blocks this behaviour under Windows by preventing the `WS_EX_LAYERED` extended window style from being set.

### DIFF
--- a/include/MainApplication.h
+++ b/include/MainApplication.h
@@ -25,13 +25,22 @@
 #ifndef MAINAPPLICATION_H
 #define MAINAPPLICATION_H
 
+#include "lmmsconfig.h"
+
 #include <QApplication>
+
+#ifdef LMMS_BUILD_WIN32
+#include <windows.h>
+#endif
 
 class MainApplication : public QApplication
 {
 public:
 	MainApplication(int& argc, char** argv);
 	bool event(QEvent* event);
+#ifdef LMMS_BUILD_WIN32
+	bool winEventFilter(MSG* msg, long* result);
+#endif
 	inline QString& queuedFile()
 	{
 	    return m_queuedFile;
--- a/src/gui/MainApplication.cpp
+++ b/src/gui/MainApplication.cpp
@@ -62,3 +62,24 @@ bool MainApplication::event(QEvent* event)
 			return QApplication::event(event);
 	}
 }
+
+#ifdef LMMS_BUILD_WIN32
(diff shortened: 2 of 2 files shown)


tresf:
@DomClark thanks.  This looks pretty safe to merge.  Does this prevent all transparencies or just those invoked from remote plugins?

DomClark:
It prevents all transparencies achieved with the Windows API which require setting an extended window style, for all windows whose messages are processed inside `QCoreApplication::exec()`. This includes all remote-plugin-invoked transparencies and excludes all Qt ones, so should be safe to include. Windows doesn't, however, provide a way to find out which process or window is making the modifications, so this could well also prevent third-party applications designed to make windows transparent from doing so to LMMS. I'm not sure if you would consider that a problem or not.

Also, I've made a slight mistake where windows that already have the `WS_EX_LAYERED` style will lose it if their extended style is modified, I'll fix that up now.

tresf:
> I'm not sure if you would consider that a problem or not.

Not at all. 👍

Umcaruje:
Ok so I tested this on Linux Qt4, and turning the transparency knob has the same behaviour with or without this PR:
gifrecord_2017-09-18_220655

The window basically becomes uninitialized and copies whatever is underneath it, and can't be reset without removing and adding the effect again. So this PR fixes the problem on windows, but not so much on linux.

zonkmachine:
> Ok so I tested this on Linux Qt4...

This fix is `#ifdef LMMS_BUILD_WIN32`
@DomClark Is it possible to fix this for Linux/Wine too?

Merge?

PhysSong:
FYI #2826 seems to resolve this issue, both on Qt4 and Qt5(if #3786 is merged).

zonkmachine:
> FYI #2826 seems to resolve this issue on linux, both on Qt4 and Qt5(if #3786 is merged).

OK. So Linux is fixed elsewhere. Merge this?

DomClark:
Turns out this only works for Qt4. In Qt5 native event filtering is achieved via [`QCoreApplication::installNativeEventFilter`](https://doc.qt.io/qt-5/qcoreapplication.html#installNativeEventFilter). I'll implement this once I've got a Qt5 build environment working, unless anyone else wants to do so first.

PR #3813, opened 2017-09-19
Enable HiDPI Scaling (Qt 5.6+ only)

flynn16:
(Qt 5.6+ only) This pull request will make LMMS scale automatically on HiDPI displays and solve the [GUI scaling issue](https://github.com/LMMS/lmms/issues/2510#issuecomment-330239441) _[at least for me]_.

### DIFF
--- a/cmake/linux/lmms.desktop
+++ b/cmake/linux/lmms.desktop
@@ -10,7 +10,7 @@ Comment[ca]=Producció fàcil de música per a tothom!
 Comment[fr]=Production facile de musique pour tout le monde !
 Comment[pl]=Prosta produkcja muzyki dla każdego!
 Icon=lmms
-Exec=env QT_X11_NO_NATIVE_MENUBAR=1 lmms %f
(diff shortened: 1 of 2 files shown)


tresf:
@flynn16 thanks!

>  solve the GUI scaling issue.

As far as we know, this solution will only work properly on Linux, correct?  Just want to be as specific as possible as this scaling issue occurs on Mac and Windows as well.

This should be targeted against `stable-1.2` so we can make it part of the RC4 release.  We can try to switch it for you from this PR.

flynn16:
> This should be targeted against `stable-1.2` so we can make it part of the RC4 release. We can try to switch it for you from this PR.

Sure. Please help me switch over to `stable-1.2`.

tresf:
@flynn16 it doesn't look possible due to the divergence of the branches.  If you are OK making a new PR against `stable-1.2`, that would be ideal.  Sorry for the inconvenience.

flynn16:
I opened a new PR #3814

PR #3820, opened 2017-09-23
Rewrite the setup dialog

### LINKED ISSUE #3681: Restart warning on setup dialog itself
A less annoying alternative for a restart warning popping up after closing the setup dialog is to display the message on the dialog itself, like so:
lmms-restart

This would also make the option "Show restart warning after changing settings" unnecessary.

Sawuare:
General
Performance
Audio
MIDI
Paths

### Changes:
This slightly modifies ConfigManager and its related files.

• Enhanced code quality and style.
• Commented some code.
• Improved translatable strings.
• Improved GUI and fixed minor GUI bugs.
• Improved the look for plugins embedding settings. 1c9ff9a
• Divided the MISC tab into 2 tabs: GUI and Projects, and moved some of its items.
• Moved some items to more appropriate sections.
• Changed the restart warning from a QMessageBox to a QLabel that is shown when a setting that requires restarting is changed.
• Removed the useless context help button. 8689e51

Closes #3681.


lukas-w:
Before & after screenshots by @Sawuare from Discord:

image
image

include/ConfigManager.h
```diff
@@ -93,184 +114,165 @@ class EXPORT ConfigManager
...
-#ifdef LMMS_HAVE_FLUIDSYNTH
-	QString m_defaultSoundfont;
-#endif
-	QString m_backgroundArtwork;
+	QString m_gigDir;
+	QString m_themeDir;
+	QString m_backgroundPicFile;
+	QString m_lmmsRcFile;
 	QStringList m_recentlyOpenedProjects;
+	QString m_version;
```
BaraMGB: In the configmanager. Cpp in the constructor the order of initialization of this members is different than the declaration here. I guess therefore Travis check failed.

Sawuare: Fixed the problem with https://github.com/LMMS/lmms/pull/3820/commits/edac2edf01b7a9f2d68f8292c77784c5ef811fee.
Thanks! 👍

michaelgregorius:
Don't know if this was already addressed in one of the commits but I wouldn't keep the option "Display restart warning after changing settings" where it is shown here:

image

Perhaps it can be moved into the GUI section because it affects the GUI behavior.

tresf:
That pop-up needs to be abolished.  We should reuse that same area to show a warning as soon as an item is clicked.

Sawuare:
> That pop-up needs to be abolished. We should reuse that same area to show a warning as soon as an item is clicked.

@tresf So you agree with the first comment in #3681?
If so, should that be done in this PR?

Sawuare:
I want the `restartWarning` label (see ded1765) to show when a setting that requires restarting is changed and hide when it's at its default (last saved) value, but I don't know how to implement that.
I already have a list of changes that require restarting and changes that don't.

Any help is appreciated. :)

Sorry for referencing the first issue in ded1765. 😅

tresf:
@Sawuare you should keep the restart warning accessible and just set the visibility of it.  In regards to how to trigger the visibility, I would recommend using multiple signals.  You can either have a boolean flag to control the visibility (e.g. quick `return` on those that shouldn't affect the message) or you can simply choose to leave the slot out from those signals that don't need it.

The boolean is cleaner as it's a one-word change `true|false` as we find ways to alleviate the need to restart.

https://forum.qt.io/topic/38121/solved-single-signal-to-multiple-slots/5

Sawuare:
I tried a lot to make it possible to scroll to the bottom of the paths scroll area, but without success, so I need help.
The paths widget's code is here: https://github.com/Sawuare/lmms/blob/0f52be7c2dcbe08575e4060351c1a95ab253ea07/src/gui/SetupDialog.cpp#L779-L977

Sawuare:
I think hiding the [restart warning](https://github.com/Sawuare/lmms/blob/8f143825aac605b2bb9420adc201f39fabafdba0/src/gui/SetupDialog.cpp#L983-L986) when a setting that requires restarting is at its default value requires much work but has little benefit. Thoughts?

Wallacoloo:
@Sawuare I think it's perfectly reasonable to not hide it. If somebody wants that functionality, they could submit a future PR that adds it. But I don't think the fact that the restart warning isn't hidden should prevent this PR from being accepted.

PhysSong:
@Sawuare  I fixed some issues in your merge. It seems like some changes in #4128 was missing.

PhysSong:
I think I might do that in a week.

PhysSong:
Resolved conflicts. `SetupDialog.cpp` was almost a hell...
Anyway, I'd ask reverting 87fec931f3ac1a759433d06cace3b915af2d4ef0. Please don't merge this right now...

Sawuare:
I made commit 87fec93 because the `stk/rawwaves` directory shouldn't be user-configurable, but I think didn't do that correctly. I couldn't test the commit because there were missing dependencies. I will revert it for now.

src/gui/SetupDialog.cpp
```diff
@@ -145,7 +145,7 @@ SetupDialog::SetupDialog(ConfigTabs tab_to_open) :
 {
 	setWindowIcon(embed::getIconPixmap("setup_general"));
 	setWindowTitle(tr("Settings"));
-	setWindowFlags(windowFlags() & ~Qt::WindowContextHelpButtonHint);
+	setWindowFlag(Qt::WindowContextHelpButtonHint, false);
```
PhysSong: From https://doc.qt.io/qt-5/qwidget.html#setWindowFlag:
> This function was introduced in Qt 5.9.

I think it's better to revert this for compatibility and add a TODO comment instead.

Sawuare: I didn't notice that Travis CI failed. Good catch! Fixed in 8c5a88c.

include/SetupDialog.h
```diff
@@ -67,148 +67,136 @@ protected slots:
...
-	bool m_warnAfterSetup;
+	// General settings widgets.
 	bool m_displaydBFS;
+	bool m_tooltips;
+	bool m_displayWaveform;
+	bool m_printNoteLabels;
+	bool m_compactTrackButtons;
+	bool m_oneInstrumentTrackWindow;
 	bool m_MMPZ;
 	bool m_disableBackup;
```
PhysSong: Maybe out of scope, but the name of this variable should be `m_enableBackup`.

src/gui/SetupDialog.cpp
```diff
@@ -324,1040 +279,1047 @@ SetupDialog::SetupDialog( ConfigTabs _tab_to_open ) :
...
+	ConfigManager::inst()->setValue("app", "nanhandler",
+					QString::number(m_NaNHandler));
+	ConfigManager::inst()->setValue("mixer", "hqaudio",
+					QString::number(m_hqAudioDev));
+	ConfigManager::inst()->setValue("mixer", "framesperaudiobuffer",
+					QString::number(m_bufferSize));
+	ConfigManager::inst()->setValue("mixer", "mididev",
+					m_midiIfaceNames[m_midiInterfaces->currentText()]);
+	ConfigManager::inst()->setValue("ui", "vstalwaysontop",
+					QString::number(m_vstAlwaysOnTop));
```
PhysSong: It's natural to move this between `vstembedmethod` and `syncvstplugins`.

src/gui/SetupDialog.cpp
```diff
@@ -30,291 +31,245 @@
...
+	m_enableAutoSave(ConfigManager::inst()->value(
+			"ui", "enableautosave", "1").toInt()),
+	m_enableRunningAutoSave(ConfigManager::inst()->value(
+			"ui", "enablerunningautosave", "0").toInt()),
+	m_smoothScroll(ConfigManager::inst()->value(
+			"ui", "smoothscroll").toInt()),
+	m_animateAFP(ConfigManager::inst()->value(
+			"ui", "animateafp", "1").toInt()),
+	m_vstEmbedMethod(ConfigManager::inst()->vstEmbedMethod()),
+	m_syncVSTPlugins(ConfigManager::inst()->value(
```
PhysSong: You should initialize `m_vstAlwaysOnTop` here.
```suggestion
	m_vstAlwaysOnTop(ConfigManager::inst()->value(
			"ui", "vstalwaysontop", "1").toInt()),
	m_syncVSTPlugins(ConfigManager::inst()->value(
```

include/SetupDialog.h
PhysSong: The name is misleading, but I think this should go into a separate pull request.

src/gui/SetupDialog.cpp
PhysSong: Sorry, there should be no `"1"`.

Reflexe:
A lot of changes are automatic and minor. That causes finding "real" changes much harder.  Would it be OK if I'll modify this PR to include only major changes, without style and formatting changes?

Sawuare:
@Reflexe Yes, please, but keep the order of statements as-is.

Sawuare:
This PR is currently a mess with many stupid changes. I will hopefully tackle it next month and make it review-able.

src/gui/SetupDialog.cpp
```diff
@@ -30,291 +31,247 @@
...
+			"ui", "smoothscroll").toInt()),
+	m_animateAFP(ConfigManager::inst()->value(
+			"ui", "animateafp", "1").toInt()),
+	m_vstEmbedMethod(ConfigManager::inst()->vstEmbedMethod()),
+	m_vstAlwaysOnTop(ConfigManager::inst()->value(
+			"ui", "vstalwaysontop").toInt()),
+	m_syncVSTPlugins(ConfigManager::inst()->value(
+			"ui", "syncvstplugins").toInt()),
+	m_disableAutoQuit(ConfigManager::inst()->value(
+			"ui", "disableautoquit").toInt()),
```
PhysSong: I just noticed the default value of `m_syncVSTPlugins` and `m_disableAutoQuit` was 1 before this PR, but it's 0 now. This PR should have not been touched the default value.
Sorry for not noticing this before.

Sawuare: Oh, sorry about that.
I should just add `"1"` as the last argument to `value()`, right?

PhysSong: Yes. That's what you only need to do.

Sawuare: Addressed in https://github.com/LMMS/lmms/commit/9f0baab93ae18f88c07dff76a0563f73200b649f.

Issue #3677, opened 2017-07-02
Some problems with VST. Again.


Andrew344:
VST preset files doesn’t appear in LMMS search window. But, in Windows file manager all  is OK. 
It occurs in LMMS 1.1.3 , 1.2.0 rc2, rc3. If I write file name in search line and press enter, the preset file will open successful. But, if I click button, it doesn't happen.
![default](https://user-images.githubusercontent.com/28777908/27772852-1828ac84-5f73-11e7-90ae-a3cc60b76584.png)

Howewer, after reloading LMMS project, even if project was saved with need preset, VST plugin settings still returns to primary preset.

tresf:
Could a typo in the translation file cause this? (the comma?)

https://github.com/LMMS/lmms/blob/32f07ac317b42dda42a1d00b0a3b1fa1feef0655/data/locale/ru.ts#L7206

Andrew344:
I change the language on English,  and now files are visible.
At this moment, I do not checked the code on link , but anyway, VST plugin settings still returns to primary preset after reloading project.

Andrew344:
Later, I will analyze it, and correct mistakes. 
Thanks for help

musikBear:
Ahhh.. sounds like the 'umlaut' problem 
ä, ö, ü, æ, ø, å 
amo. -can be problems in filenames as well as foldernames

tresf:
@musikBear do you have evidence of this problem?  His folders do not contain the characters you mention, only the save dialog, which shouldn't affect it.  My investigation points to a translation issue.  You should be able to unit test this using the raw translation file and running it though Qt Linguist.  Once that is done you or someone can work with @liushuyu to fix.

oeai:
maybe because of this?
    <message>
        <source>.fxp</source>
        <translation>.fxp</translation>
    </message>
    <message>
        <source>.FXP</source>
        <translation>.FXP</translation>
    </message>
    <message>
        <source>.FXB</source>
        <translation>.FXB</translation>
    </message>
    <message>
        <source>.fxb</source>
        <translation>.fxb</translation>
</message>

oeai:
i remember that there were the same problem with converting to ogg, when values for converting were taken from the selected string (that was also translated) and same problem happened in chinese.
[...]

PR #3016, opened 2016-09-04
Allow PROJECT_YEAR override

jasp00:
`PROJECT_YEAR` depends on build time, which is a problem with reproducible builds. ~~This patch allows the year to be set on configure. This solution is needed until a bot includes the information in the source tree.~~
This patch allows to use information from the folder `release-info`, which will come from a tarball generated in the Travis build. The command to build `lmms-VERSION-release-info.tar.gz` is `make release-info` and requires Git.

PR #3826, opened 2017-09-25
About dialog & export dialog improvements.

Sawuare:
**Changes:**
• Improved code style.
• Improved translatable strings.
• Improved the geometry of the dialogs.
• Improved the position of some items in the dialogs.

**TODO:**
• In the about dialog, make the color of the link the same as the color of the strings in the tabs.
I couldn't find the exact color code...
image

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,7 +22,7 @@ STRING(TOUPPER          "${CMAKE_PROJECT_NAME}" PROJECT_NAME_UCASE)
 # Updated by maintenance tasks
 SET(PROJECT_YEAR 2015)
 
-SET(PROJECT_AUTHOR      "LMMS developers")
+SET(PROJECT_AUTHOR      "LMMS Developers")
 SET(PROJECT_URL         "https://lmms.io")
 SET(PROJECT_EMAIL       "lmms-devel@lists.sourceforge.net")
 SET(PROJECT_DESCRIPTION "${PROJECT_NAME_UCASE} - Free music production software")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,7 +22,7 @@ STRING(TOUPPER          "${CMAKE_PROJECT_NAME}" PROJECT_NAME_UCASE)
 # Updated by maintenance tasks
 SET(PROJECT_YEAR 2015)
 
-SET(PROJECT_AUTHOR      "LMMS Developers")
+SET(PROJECT_AUTHOR      "LMMS developers")
 SET(PROJECT_URL         "https://lmms.io")
 SET(PROJECT_EMAIL       "lmms-devel@lists.sourceforge.net")
 SET(PROJECT_DESCRIPTION "${PROJECT_NAME_UCASE} - Free music production software")
--- a/src/core/ProjectRenderer.cpp
+++ b/src/core/ProjectRenderer.cpp
@@ -37,55 +37,54 @@
 #include "sched.h"
 #endif
 
+
 const ProjectRenderer::FileEncodeDevice ProjectRenderer::fileEncodeDevices[] =
 {
-
-	{ ProjectRenderer::WaveFile,
-		QT_TRANSLATE_NOOP( "ProjectRenderer", "WAV-File (*.wav)" ),
-					".wav", &AudioFileWave::getInst },
-	{ ProjectRenderer::FlacFile,
-		QT_TRANSLATE_NOOP("ProjectRenderer", "FLAC-File (*.flac)"),
-		".flac",
-		&AudioFileFlac::getInst
+	{ProjectRenderer::WaveFile,
+			QT_TRANSLATE_NOOP("ProjectRenderer", "WAV (*.wav)"),
+			".wav", &AudioFileWave::getInst},
+	{ProjectRenderer::FlacFile,
+			QT_TRANSLATE_NOOP("ProjectRenderer", "FLAC (*.flac)"),
+			".flac",
+			&AudioFileFlac::getInst
 	},
-	{ ProjectRenderer::OggFile,
-		QT_TRANSLATE_NOOP( "ProjectRenderer", "Compressed OGG-File (*.ogg)" ),
-					".ogg",
+	{ProjectRenderer::OggFile,
+			QT_TRANSLATE_NOOP("ProjectRenderer", "OGG (*.ogg)"),
+			".ogg",
 #ifdef LMMS_HAVE_OGGVORBIS
-					&AudioFileOgg::getInst
+			&AudioFileOgg::getInst
 #else
-					NULL
+			NULL
 #endif
-									},
-	{ ProjectRenderer::MP3File,
-		QT_TRANSLATE_NOOP( "ProjectRenderer", "Compressed MP3-File (*.mp3)" ),
-					".mp3",
+	},
+	{ProjectRenderer::MP3File,
+			QT_TRANSLATE_NOOP("ProjectRenderer", "MP3 (*.mp3)"),
+			".mp3",
 #ifdef LMMS_HAVE_MP3LAME
-					&AudioFileMP3::getInst
+			&AudioFileMP3::getInst
 #else
-					NULL
+			NULL
 #endif
-									},
-	// ... insert your own file-encoder-infos here... may be one day the
-	// user can add own encoders inside the program...
+	},
+	// Insert your own file-encoder infos here.
+	// Maybe one day the user can add own encoders inside the program.
 
-	{ ProjectRenderer::NumFileFormats, NULL, NULL, NULL }
-
-} ;
-
+	{ProjectRenderer::NumFileFormats, NULL, NULL, NULL}
+};
 
 
 
-ProjectRenderer::ProjectRenderer( const Mixer::qualitySettings & qualitySettings,
-					const OutputSettings & outputSettings,
-					ExportFileFormats exportFileFormat,
-					const QString & outputFilename ) :
-	QThread( Engine::mixer() ),
-	m_fileDev( NULL ),
-	m_qualitySettings( qualitySettings ),
-	m_oldQualitySettings( Engine::mixer()->currentQualitySettings() ),
-	m_progress( 0 ),
-	m_abort( false )
+
+ProjectRenderer::ProjectRenderer(const Mixer::qualitySettings & qualitySettings,
+			const OutputSettings & outputSettings,
+			ExportFileFormats exportFileFormat,
+			const QString & outputFilename) :
+	QThread(Engine::mixer()),
+	m_fileDev(NULL),
+	m_qualitySettings(qualitySettings),
+	m_oldQualitySettings(Engine::mixer()->currentQualitySettings()),
+	m_progress(0),
+	m_abort(false)
 {
 	AudioFileDeviceInstantiaton audioEncoderFactory = fileEncodeDevices[exportFileFormat].m_getDevInst;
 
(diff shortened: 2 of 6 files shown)


Spekular:
I'm not a fan of the about dialog changes, but the rest looks great!

Sawuare:
Are there specific about dialog changes that you don't like?

Spekular:
Changing the BG to the same everywhere means your eyes are less drawn to the relevant portion of the window. The tabs also look awkward now.

Also, I hadn't checked these at full resolution, but I'm not too crazy about the "depth" you've added on the outlines, and especially not on the progress bar. This looks further from the single window/flat look we're steering towards than it is now.

tresf:
The current gray color is terrible for text because it is neither dark or light enough for text contrast.  It's also inconsistent with the rest of the UI currently.  I'm in agreement with @Sawuare, it needs to be darker.  If it needs to be a different shade, we can cherry-pick one from the rest of the UI.  I would use the effects chain as an example of how to show a border, depth and content inside a container.

image

@Spekular please try to be constructive in your comments so that @Sawuare can have concrete changes to make.

Spekular:
>your eyes are less drawn to the relevant portion of the window

>This looks further from the single window/flat look we're steering towards than it is now.

I fail to see how these aren't constructive.

As for the tabs, I think my issue is that the tabs in the background are brighter than those in the foreground, confusing the matter of which is in focus (especially since text is the same color on both). Also, our other tab-like elements (sidebar, tabs in instruments, tabs in settings) have the same size tabs for active and inactive. The large height difference in the active tab looks out of place in comparison, especially now that the about section is vertically shorter.

In addition, the export dialog sections are inset (thicker line on top indicates shadow from light source above), whereas the sections of the about screen pop out (drop shadow below them).

We've made decisions in the past based around moving towards the look set forth in budislav's single window concept, and the constant width of the previous borders is more in line with that look than the inset/popping one that this PR has.

Sawuare:
The BG color wasn't changed by me, but I like it.
The outlines weren't changed by me either, but I also like them. 😃

Spekular:
Then your before/after screenshots are giving a bad comparison, no? If this PR isn't changing the outline it should be the same in both.

Sawuare:
> Then your before/after screenshots are giving a bad comparison, no? If this PR isn't changing the outline it should be the same in both.

You're right. The before pictures are from RC3 😅, I will update them to be from master.

Spekular:
Great, thanks!

Spekular:
As far as I can tell, all that changed is that the windows are smaller and the export dialogue checkboxes were moved out from "quality settings". Is this correct? If so I have no objections.

tresf:
The visual differences may be due to the installed (Qt5) version versus the compiled (Qt4) version.  We don't have instructions written to compile against Qt5 yet.

CMakeLists.txt
```diff
@@ -22,7 +22,7 @@ STRING(TOUPPER          "${CMAKE_PROJECT_NAME}" PROJECT_NAME_UCASE)
 # Updated by maintenance tasks
 SET(PROJECT_YEAR 2015)
 
-SET(PROJECT_AUTHOR      "LMMS Developers")
+SET(PROJECT_AUTHOR      "LMMS developers")
```
tresf: Why lowercase here?  Is this necessary change as part of this PR or was it something that was never reverted after testing?

src/core/ProjectRenderer.cpp
```diff
@@ -37,55 +37,54 @@
...
 {
-
-	{ ProjectRenderer::WaveFile,
-		QT_TRANSLATE_NOOP( "ProjectRenderer", "WAV-File (*.wav)" ),
-					".wav", &AudioFileWave::getInst },
-	{ ProjectRenderer::FlacFile,
-		QT_TRANSLATE_NOOP("ProjectRenderer", "FLAC-File (*.flac)"),
-		".flac",
-		&AudioFileFlac::getInst
+	{ProjectRenderer::WaveFile,
```
tresf: We need to stop translating stuff like `*.mp3`.  This is only going to cause issues down the road where a bad translation breaks a feature.

Umcaruje: Yeah according to #3677 this is already making problems in other places

tresf:
@Sawuare I'm going to be a bit critical here and other devs feel free to chime in and refute, but I really don't see the benefit of doing the code cleanup in this PR.  If this PR is about condensing the UI, that's what it should do and only that.  There are somewhat unrelated and random changes here such as formatting, removing "Compressed" from the MP3 and OGG file extensions and they're hard to review when the formatting has changed so much.

Also as @Spekular has pointed out above, the before and after screenshots are bad examples of the changes due to the mixed Qt versions.

We all want the formatting to get cleaner in the codebase, but I think that's a candidate for a separate all-breaking PR that should stand on its own.

Umcaruje:
>We all want the formatting to get cleaner in the codebase, but I think that's a candidate for a separate all-breaking PR that should stand on its own.

This. @tresf said it all, don't have anything to add. I feel the same way.

gi0e5b06:
@tresf Sure. A dozen of PR would certainly be useful. In fact, one PR for each changed line of code should be the standard.

@Sawuare Good work. Let's merge and move on.

BTW, I'm going to add two checkboxes in the export dialog:
 - remove silence (at the beginning and the end)
 - normalize audio output (peak)

src/core/ProjectRenderer.cpp
Sawuare: I just removed the space after `{` here.
Do you want me to keep the strings but make them untranslatable? If so, is removing `QT_TRANSLATE_NOOP` enough?

Sawuare:
I'm going to reverse all formatting changes and remove the before and after images. 👍 

@gi0e5b06 
Your suggested check boxes are a great idea, but doing them in a separate PR is the best thing to do. 😉

Sawuare:
@tresf 

> removing "Compressed" from the MP3 and OGG file extensions and they're hard to review when the formatting has changed so much.

You're right about the formatting stuff, but I removed "Compressed" and other stuff because I think they belong in the documentation (when it's created), not in the program itself.

Umcaruje:
>Let's merge and move on.

I'm sorry but you are in no position to say that.

>BTW, I'm going to add two checkboxes in the export dialog:
>
>remove silence (at the beginning and the end)
>normalize audio output (peak)

You do realize those changes will never make it into the codebase if you don't play by the rules? We can discuss all you want, but one PR per feature is how things are done here, and the majority of the community agrees on it.

src/gui/AboutDialog.cpp
```diff
@@ -52,9 +52,12 @@ AboutDialog::AboutDialog(QWidget* parent) :
 	copyrightLabel->setText( copyrightLabel->text().
 					arg( LMMS_PROJECT_COPYRIGHT ) );
 
-	authorLabel->setPlainText( embed::getText( "AUTHORS" ) );
+	authorLabel->setPlainText(
+					embed::getText( "AUTHORS" ) );
 
-	licenseLabel->setPlainText( embed::getText( "LICENSE.txt" ) );
+	/*involvedLabel->setPlainText(
```
Umcaruje: Why did you disable the contributors file, it contains all of our github contributors and it's more accurate than the authors file (though the authors file contains some people that wouldn't be credited through github)

lukas-w: > it contains all of our github contributors

@Umcaruje Not anymore, jasp00 removed the CMake logic that created the file in 30f1e52be35ed35f8e0e40f876620ce49007be62. He planned to replace that by a maintenance task on our server, but it appears it was never implemented correctly. That's probably why @Sawuare removed it. Related: LMMS/lmms.io/pull/198, #3016

Umcaruje: Then, should we update the authors file with some missing people? Or revert @jasp00's changes, since he sent a mail saying he won't be able to contribute for some time.

Sawuare: I didn't know the story of the contributors tab. I just removed it because it didn't display anything other than: "Contributors ordered by number of commit:".

tresf:
I'll take over the `CONTRIBUTORS` file before `stable-1.2`, just backlogged on the task.  Please keep it in the codebase @Sawuare. 

> I removed "Compressed" and other stuff because I think they belong in the documentation (when it's created), not in the program itself.

After typing that I took a look at some other programs and they don't use the word "Compressed" either.  From a translation perspective, and it's more to translate.  Removing it seems warranted.

> A dozen of PR would certainly be useful

@gi0e5b06 No, that would not be useful.

> one PR for each changed line of code should be the standard.

Although I find this "revolting defiance" to be amusing, it's better left in our Discord chat server.   There are some pros and cons to "reformatting as you go" and we can talk about that on Discord if it is going to be a problem with your contributions.  In this case, @Sawuare is making his first PR, so we're trying to set reasonable standards moving forward, but they aren't set in stone.  https://lmms.io/chat, then click on `#devtalk`.

src/core/ProjectRenderer.cpp
```diff
@@ -41,15 +41,15 @@ const ProjectRenderer::FileEncodeDevice ProjectRenderer::fileEncodeDevices[] =
...
 					".wav", &AudioFileWave::getInst },
 	{ ProjectRenderer::FlacFile,
-		QT_TRANSLATE_NOOP("ProjectRenderer", "FLAC-File (*.flac)"),
+		QT_TRANSLATE_NOOP("ProjectRenderer", "FLAC (*.flac)"),
 		".flac",
 		&AudioFileFlac::getInst
 	},
 	{ ProjectRenderer::OggFile,
-		QT_TRANSLATE_NOOP( "ProjectRenderer", "Compressed OGG-File (*.ogg)" ),
+		QT_TRANSLATE_NOOP( "ProjectRenderer", "OGG (*.ogg)" ),
```
lukas-w: While we're at it, can we make this say "Ogg-Vorbis" instead? This would correctly reflect the fact that Ogg is merely the container format, while Vorbis is the audio codec that's being used here.

tresf: From wikipedia.org:

> Before 2007, the .ogg filename extension was used for all files whose content used the Ogg container format. Since 2007, the Xiph.Org Foundation recommends that .ogg only be used for Ogg Vorbis audio files. The Xiph.Org Foundation decided to create a new set of file extensions and media types to describe different types of content such as .oga for audio only files, .ogv for video with or without sound (including Theora), and .ogx for multiplexed Ogg.[4]

"OGG" should be just fine.

Sawuare:
The "Export as loop" option does not work, I'll comment it out if there are no objections.

zonkmachine:
> I'll comment it out if there are no objections.

Don't. If it doesn't work it's a bug and should be fixed. Is this bug in master only?

zonkmachine:
> It is in RC.3 and master.

OK. I'll have a look.

zonkmachine:
Tested. The option marked '**Export as loop (remove end silence)**' works just fine. Can you please describe steps to reproduce the issue?

Sawuare:
To reproduce:
1, Create a pattern for any instrument.
2. Put 2 8th notes after each other.
3. Export with the `'Export as loop (remove end silence)` option.

There will be silence after the notes.
image

zonkmachine:
Not a bug.

Export as loop and default export together.

looptest

By default an extra bar is added to allow for notes and delay effects to end. You can choose to export as a loop but it will not end the sample after the right note but still fill out the whole last bar.

Here is a recent issue suggesting an option to remove silence. https://github.com/LMMS/lmms/issues/3830

Sawuare:
Thank you for clarifying.

Perhaps we change the option's name to something clearer, like "Export as loop (remove extra bar)"?

CMakeLists.txt
Sawuare: It's the latter. I will capitalize it.

tresf:
Looks good.
screen shot 2017-11-14 at 11 56 47 pm
screen shot 2017-11-14 at 11 55 21 pm

Issue #3822, opened 2017-09-24
TCOs in BB editor are draggable since #3649


PhysSong:
Steps to reproduce:
1. Create a new project.
2. Open BB editor.
3. Add TCOs by any possible ways(put notes, add automation track or sample track, etc.)
4. Try to drag them.

Note that this behavior is only in `master` branch.
@BaraMGB

BaraMGB:
I'll have a look.

Umcaruje:
Closed via #3832

PR #3832, opened 2017-09-28
don't allow dragging TCOs in BBEditor

BaraMGB:
fixes https://github.com/LMMS/lmms/issues/3822

### DIFF
--- a/src/core/Track.cpp
+++ b/src/core/Track.cpp
@@ -688,7 +688,7 @@ void TrackContentObjectView::paintTextLabel(QString const & text, QPainter & pai
 void TrackContentObjectView::mousePressEvent( QMouseEvent * me )
 {
 	setInitialMousePos( me->pos() );
-	if( me->button() == Qt::LeftButton && fixedTCOs() == false )
+	if( !fixedTCOs() && me->button() == Qt::LeftButton )
 	{
 		if( me->modifiers() & Qt::ControlModifier )
 		{
@@ -776,7 +776,7 @@ void TrackContentObjectView::mousePressEvent( QMouseEvent * me )
 		{
 			m_tco->toggleMute();
 		}
-		else if( me->modifiers() & Qt::ShiftModifier && fixedTCOs() == false )
+		else if( me->modifiers() & Qt::ShiftModifier && !fixedTCOs() )
 		{
 			remove();
 		}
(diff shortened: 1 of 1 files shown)


lukas-w:
For better modularity: Can you try implementing the fix in the affected subclass (`BBTCOView`?) instead of using `dynamic_cast` in the super class? I.e. override `mousePressEvent` in `BBTCOView` and stop the events from being propagated there.

BaraMGB:
I'm not sure. I looked into this. But a BBTCOView is this one:

bildschirmfoto von 2017-09-28 14-29-45

The issue appears with all TCOs in a BBTrackContainer.

tresf:
I was under the impression [`fixedTCOs`](https://github.com/LMMS/lmms/blob/32f07ac317b42dda42a1d00b0a3b1fa1feef0655/include/BBEditor.h#L75) was the boolean flag that stored whether or not a pattern was resizable.

lukas-w:
Didn't know about `fixedTCOs`, looks like this is what we need here. 👍

BaraMGB:
Sometimes it's easier than imagined. 🙄

src/core/Track.cpp
```diff
@@ -688,9 +688,7 @@ void TrackContentObjectView::paintTextLabel(QString const & text, QPainter & pai
 void TrackContentObjectView::mousePressEvent( QMouseEvent * me )
 {
 	setInitialMousePos( me->pos() );
-
-	BBTrackContainer * isbbEditor = dynamic_cast<BBTrackContainer*>( getTrackContentObject()->getTrack()->trackContainer() );
-	if( !isbbEditor && me->button() == Qt::LeftButton )
+	if( me->button() == Qt::LeftButton && fixedTCOs() == false )
```
Umcaruje: You could just to `!FixedTCOs()` since it's a boolean

Umcaruje:
Just tested this out, works well, fixes the bug :+1:

src/core/Track.cpp
BaraMGB: I wanted to be consist with the rest of the code. But if there are no objections I do it for the whole If statement.

Umcaruje: `!FixedTCOs()` because I literally read it "If it's not a Fixed TCO"

Issue #2303, opened 2015-08-30
Zooming in song editor doesn't zoom to part under cursor


MightyPork:
This is more of a "missing feature" than a bug - should be easy to fix.

When you zoom in in the song editor, it simply changes horizontal scale without moving the scrollbar, and the thing you were trying to zoom is usually somewhere off the screen. It'd be great if it changed h-scrollbar to keep the under-mouse part visible.

![LMMS bug - zoom moves to start](http://cdn.makeagif.com/media/1-26-2016/3iHwRt.gif)
[Here's the bug captured (YouTube video)](https://youtu.be/XSELW7VWDQw)

![LMMS zoom bug - expected behavior](http://cdn.makeagif.com/media/1-26-2016/Z7n6GS.gif)
[Expected behavior (YouTube video)](https://youtu.be/YHcLewlrMtM) (after some video cutting with kdenlive ^^)

LocoMatt:
I agree with @MightyPork it's very annoying.

Umcaruje:
Took the liberty of updating the original post with gifs made from the videos.

PhysSong:
Implemented in `master` via #3835.

PR #3835, opened 2017-09-28
Zooming with mouse wheel center

Premik:
Fixes the #2303

### DIFF
--- a/include/SongEditor.h
+++ b/include/SongEditor.h
@@ -135,6 +135,7 @@ private:
 
 	bool m_scrollBack;
 	bool m_smoothScroll;
+	int m_widgetWidthTotal;
 
 	EditMode m_mode;
 	EditMode m_ctrlMode; // mode they were in before they hit ctrl
--- a/src/gui/editors/AutomationEditor.cpp
+++ b/src/gui/editors/AutomationEditor.cpp
@@ -1634,6 +1634,17 @@ void AutomationEditor::wheelEvent(QWheelEvent * we )
 			x--;
 		}
 		x = qBound( 0, x, m_zoomingXModel.size() - 1 );
+
+		int mouseX = (we->x() - VALUES_WIDTH)* MidiTime::ticksPerTact();
+		// ticks based on the mouse x-position where the scroll wheel was used
+		int ticks = mouseX / m_ppt;
+		// what would be the ticks in the new zoom level on the very same mouse x
+		int newTicks = mouseX / (DEFAULT_PPT * m_zoomXLevels[x]);
+
+		// scroll so the tick "selected" by the mouse x doesn't move on the screen
+		m_leftRightScroll->setValue(m_leftRightScroll->value() + ticks - newTicks);
+
+
 		m_zoomingXModel.setValue( x );
 	}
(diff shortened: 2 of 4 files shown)


Umcaruje:
Hi Premik, thanks for contributing to LMMS, I tested this out and it works well, and the code is by the guidelines, but can you implement this in the automation editor as well, so we have consistent UX over the software?

Premik:
@Umcaruje I've noticed this too. But it is another bug or missfeature which was there before my change. Somehow the automation track doesn't expand to new bars as the control poins are being added (like the other editors do). And if you zoom in a bit more the Automation editor won't let you scroll after the Pattern end (I bet it would be the Bar no. 4 in your case). Try with shift+mouse wheel for example. You can workaround this by expanding the Pattern in the song editor.

 Maybe a picture would be better. Below the first automation track with the long Pattern is ok. But the second one is dodgy.
automationscroll


@zonkmachine Can you provide more details about that Piano Roll issue?

PhysSong:
According to the [`QAbstractSlider` documentation](http://doc.qt.io/qt-5/qabstractslider.html#value-prop), it looks like a problem with scrolls' range.

Premik:
@PhysSong True. The zooming would center to the cursor as far as it is possible to do so. I think that is where the confusion came from. For example when zooming to some bars at the very beginning or very end of a clip those can never be in the middle of the screen (where the mouse would be).
This works exactly the same way as selecting the zoom level from the top-left drop down first and then scrolling with the scroll-bar. And I don't think it is something that should be changed.

For the automation editor I suggest to open separate ticket and not mix it with this one. As those are not related.

tresf:
@Umcaruje, this PR shows 2 changes requested.  Is there anything remaining before merging?

zonkmachine:
> @Umcaruje I've noticed this too. But it is another bug or missfeature which was there before my change. Somehow the automation track doesn't expand to new bars as the control poins are being added (like the other editors do). And if you zoom in a bit more the Automation editor won't let you scroll after the Pattern end (I bet it would be the Bar no. 4 in your case). Try with shift+mouse wheel for example. You can workaround this by expanding the Pattern in the song editor.

Could have been this I saw. If the rest of you are happy with this then go ahead and merge it.

Wallacoloo:
Just merged this locally and gave it a try - impressively it's still working fine with no merge conflicts even after 8 months. It's undeniably better than what we have in place right now - I feel we can do whatever further work is needed to improve zoom behavior in the automation editor and piano roll sometime down the road after merging.

@Umcaruje Github has this marked as not ready to merge because "Umcaruje requested changes". This is a useful Github feature that I don't want to blow off - but you didn't reply back in November when tresf asked if you were happy to merge it. I'll give you another chance to speak up, otherwise I'm going to merge this in a few days.

Umcaruje:
Sorry about this, I completely forgot about this PR. OP’s explanation for the behavior i expirienced is good I’m merging.

Wallacoloo:
@Umcaruje No worries. Thanks for reviewing it.

@Premik Thanks for the contribution! (And sorry we let it sit here inactive/unmerged for so long >.<)

PR #3837, opened 2017-09-28
fixing C abs/C++ abs mix up

devnexen:
from 6.2, it will clang/libc++ thus moving from gcc 4.9 to this compiler for testing openbsd build which raised a little issue with abs call mixing up with its C++ counterpart. casting rate field to int explicitly. A similar fix would follow for stable-1.2


tresf:
> A similar fix would follow for stable-1.2

We've been trying to sync from `stable-1.2` to master and we've made `stable-1.2` our default branch temporarily to encourage that.  Would you be opposed to directing this at `stable-1.2` please?  We'll be back on master once `stable-1.2` is finally... well, stable. :)

devnexen:
I just noticed it I m catching up on this branch :-)

tresf:
@devnexen sorry, I hadn't realized they had drifted.  We'll merge them separately then, thanks.

Issue #3721, opened 2017-07-23
Track restored to large size when saved after switching instruments in instrument window


michaelgregorius:
Steps to reproduce:
1. Start LMMS with the default project (with at least one Triple Oscillator).
2. Add another Triple Oscillator (or another instrument).
3. Open the instrument window of one of the instruments.
4. Switch between instruments using the left / right arrow on the top of the instrument window.
4. Save.
5. Load the file that you have just saved.

The last selected instrument will be restored with a very large height that coincides with the instrument window's height.

The main cause is that there are two methods which store different types of heights as the attribute `height` in the node `track` under the parent `trackcontainer`:
1. The method `Track::saveSettings` stores the height of the track under that node.
2. The method `InstrumentTrackWindow::saveSettings` stores the window geometry of the instrument window by delegating to `MainWindow::saveWidgetState`. This step stores the height of the instrument window into the aforementioned node which causes the faulty restoring of the track height.

The switching of the instrument windows is relevant for this bug because at one point this action calls `InstrumentTrackView::getInstrumentTrackWindow` which installs a hook for the serialization so that `InstrumentTrackWindow::saveSettings` is called as well. Without this action no hook is installed and everything is fine. You can see this if you add a break point to `SerializingObject::setHook` and then execute the steps described above.

So everything action that leads to the installation of the aforementioned hook or a direct call to `InstrumentTrackWindow::saveSettings` will likely trigger this bug.

See the closed pull request #3692 for more details.

michaelgregorius:
Here's an example of a malformed `track` DOM element in the written project file:
```XML
<track width="262" x="1667" y="0" maximized="0" muted="0" height="472" visible="1"
       type="0" minimized="0" name="TripleOscillator" tab="0" solo="0">
```
The attributes are stored by the following classes / methods.

`Track::saveSettings`: type, name, muted, solo, height

`InstrumentTrackWindow::saveSettings`: tab

`InstrumentTrackWindow::saveSettings` (via `MainWindow::saveWidgetState`): visible, minimized, maximized, x, y, width, height

As you can see the `height` attribute is used twice which leads to the wrong height being used when the track is restored.

To solve this problem it will need to be discussed whether it's correct that all these entities write to the `track` element. It might for example make sense that the instrument window stores its state to an `instrumenttrackwindow` element. It might also be unnecessary that the class `InstrumentTrackWindow` stores its window state at all (see below). So an alternative structure might look like this (if the window still stores its state):
```XML
[...]

PR #3840, opened 2017-09-29
Save track height under a separate name, 'trackheight'

zonkmachine:
Temporary fix for https://github.com/LMMS/lmms/issues/3721.

### DIFF
--- a/src/core/Track.cpp
+++ b/src/core/Track.cpp
@@ -2086,7 +2086,7 @@ void Track::saveSettings( QDomDocument & doc, QDomElement & element )
 	element.setAttribute( "solo", isSolo() );
 	if( m_height >= MINIMAL_TRACK_HEIGHT )
 	{
-		element.setAttribute( "height", m_height );
+		element.setAttribute( "trackheight", m_height );
 	}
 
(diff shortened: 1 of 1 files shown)


zonkmachine:
@PhysSong Merge?

PhysSong:
It should work without compatibility problem, but I want to know how @michaelgregorius thinks.

michaelgregorius:
So if I understand correctly then files stored with this patch will differentiate between `trackheight` which stores the height of the track as shown in the Song Editor and `height` which stores the height of the instrument window. Therefore loading and saving of such new files will work without the two classes interfering with each other. If an old file is loaded the new attribute `trackheight` will not be found and in these cases the tracks will be restored to their default height. Is this correct? If that's the case then the patch looks good to me.

@zonkmachine, @PhysSong: Thanks for working on this issue!

zonkmachine:
>Is this correct?

Correct. I've tested this and it fixes the issue. Merge?

michaelgregorius:
@zonkmachine I'd say let's merge this.

zonkmachine:
:tractor:

PR #3841, opened 2017-09-29
Stop some autosave crashes

DomClark:
Move autosaving back to the main thread as `Song::saveProjectFile` doesn't appear to be thread-safe. Instead prevent autosave triggering during main thread VST communication by tracking call depth to `RemotePlugin::waitForMessage`.

Prevent autosave triggering during project load.

### DIFF
--- a/include/MainWindow.h
+++ b/include/MainWindow.h
@@ -249,11 +249,4 @@ signals:
 
 } ;
 
-class AutoSaveThread : public QThread
-{
-	Q_OBJECT
-public:
-	void run();
-} ;
-
 #endif
--- a/include/RemotePlugin.h
+++ b/include/RemotePlugin.h
@@ -621,6 +621,11 @@ public:
 			fetchAndProcessNextMessage();
 		}
 	}
+
+	static bool isMainThreadWaiting()
+	{
+		return waitDepthCounter() > 0;
+	}
 #endif
 
 	virtual bool processMessage( const message & _m ) = 0;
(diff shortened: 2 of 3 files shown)


zonkmachine:
> Move autosaving back to the main thread as Song::saveProjectFile doesn't appear to be thread-safe.

Right. Until recently auto-saving was not active while playing for some reason. I guess not being thread safe was the reason. We added the option to turn this on and I've been monitoring the save files for issues but haven't found any. Would this PR address this issue too?

DomClark:
> Would this PR address this issue too?

I shouldn't think so, unfortunately; autosaving always used to be in the main thread, I moved it out in #3776 as some other changes there would cause it to run during VST communication and hang trying to acquire a mutex twice in the same thread. That particular change would cause occasional crashes due to some concurrent modification to the song, so this PR is primarily a new, safer attempt at not autosaving during VST communication.

I'm not sure why autosave isn't always active during playback, but given that there's an option to enable it, it seems less likely that it's because it might crash and more likely that it's because it's CPU intensive.

zonkmachine:
> ... more likely that it's because it's CPU intensive.

Yes. We/I weren't completely sure. It didn't work on XP for instance because it would overload the system.

zonkmachine:
Tested. Unfortunately I can't replicate the issue so my test is of little value. I've tried and import large files with vst tracks but I can't make it crash and auto-save will not trigger at all for me, with or without this fix until the project has finished loading.

zonkmachine:
@DomClark @PhysSong Since I don't see any new issues from this, maybe we should merge it and leave the testing to the user?

PR #3844, opened 2017-09-30
Fix #3842: Opening a project with LB302 with specific waveforms produces a noise burst.

### LINKED ISSUE #3842: Opening a project with LB302 with specific waveforms produces a noise burst.
🔊 

To reproduce:
1. Load an LB302 instrument.
2. Change the waveform to one of the following underlined types:
image
3. Save the project (the format of the project file doesn't matter).
4. Open the project.

_An example project file:_ [LB302 noise burst.zip](https://github.com/LMMS/lmms/files/1346412/LB302.noise.burst.zip)

michaelgregorius:
Fix the problem described in #3842 by removing a duplicate initialization of a VCA variable. Also clean up some code by
* moving the initialization of the VCA related variables into the constructor initializer list and
* introducing an enum for the VCA mode.


zonkmachine:
Fix confirmed. Code looks fine...

zonkmachine:
I don't know where but I vaguely remember suggesting this should go in master.
It should be backported to 1.2 .

Issue #3800, opened 2017-09-10
Automations continue after the end of their TCOs.


Sawuare:
![automations tcos bug](https://user-images.githubusercontent.com/25503477/30249122-4f90262e-963d-11e7-9e51-6cd62537fa1d.gif)

_The project file:_ [Automations TCOs bug.zip](https://github.com/LMMS/lmms/files/1290519/Automations.TCOs.bug.zip)

The automation in the GIF and the project automates the volume of BitInvader.
Automation TCO 1 is the same as 2, but 1 is shortened.
Automation TCO 1 behaved like 2, but it should have behaved like 3, by stopping the automation at its end and mapping the last value in the automation to the automated parameter.

PhysSong:
Related: #3382, #662.

musikBear:
Confirm win32 RC3
Could be related to the smart addition of setting automation value according to play-head pos - eg #3382

zonkmachine:
As @musikBear points out this is from #3382. It's by design. @tresf Right?

Umcaruje:
@zonkmachine I'm pretty sure that this is a bug, because the value should stay from the last value of the automation clip, not from the values that are hidden. @lukas-w do you agree?

zonkmachine:
Right. Now I got it. It's a bug, sorry.

lukas-w:
Definitely a bug.

PhysSong:
There are some tricky things. For discrete progression, the value should stay the last value **before** the end point(the value **at** the end point shouldn't be included), while it should be the value at the end point for other progressions.
Anyway, I found that `TrackContainer::automatedValuesAt()` doesn't check automation patterns' length. We may check the length there.

Issue #3464, opened 2017-03-29
Automations inside patterns don't work anymore


karmux:
After 6004edae5c5108560caf9d1f67789b5e47329212 automations inside patterns don't work anymore. 

- Create B+B pattern
- Add automation track inside pattern
- Connect some random knob or slider (e.g. master volume) to automation track
- Draw something into automation track
- Play and connected knob or slider stays still

Automations in song editor work.

Lukas-W:
@karmux Thanks for reporting! #3481 should fix it once merged.

karmux:
@Lukas-W thanks! #3481 fixes it.

PR #3845, opened 2017-10-01
Xpressivebackport

zonkmachine:
As discussed just briefly on discord I'm backporting xpressive pluggin to stable-1.2 as its stable, was originally targeted for 1.2 (switched to master after issues with a bad merge that is now fixed) and it would otherwise be much to long time before the creator of the pluggin will see his work in use.

I had to tweak it some and this may or may not be an issue when merging back to master.

fb691b7 shouldn't be merged into master.


lukas-w:
👎, because "feature freeze" means feature freeze. Ignoring this will delay release even more (just take a look at #3464 and #3800, regressions from #3382, which was merged into `stable-1.2`). Let's focus on bug fixes only.

I suggest starting a `stable-1.3` branch based on `master` immediately after releasing 1.2 and merging some of the many feature PRs waiting to be integrated. Xpressive will be part of it.

zonkmachine:
Downvoted on discord #developer... :cry:

PR #3849, opened 2017-10-02
Update the grid after changing the quantization value in automation editor.

Sawuare:
Fixes that changing the quantization value in automation editor doesn't update the grid until clicking or moving inside of it, by adding a call to `update()` at the end of `setQuantization()`.


zonkmachine:
Backported to stable-1.2 in 0dbbdd9f4c35d3227a8812c10c16081f37725895

Issue #340, opened 2014-02-19
Arpeggios, sort mode, and release


eagles051387:
Hello, I'm on LMMS 0.4.14-rc1, and there is a problem with arpeggios.
This existed for a while, but I just did a bit more debugging. When the release of the instrument overlaps with the next note in sequence, a note will either a) fail to play, or b) play two notes together when the instrument is set to "sort" mode.

repsej:
This is still an issue (1.1.3 win64) - and I think it has gotten worse recently.  

I've made a very simple example that shows it.

![image](https://cloud.githubusercontent.com/assets/116525/7784754/5bcf5d14-0171-11e5-986a-aa3a6d8397e7.png)

There are just two half notes in the piano roll and the arpeggio time setting is also set to half a note.

Playing this yields a lot of overlapping notes and strange clicking noise.

Worth noticing:
 - Changing the tempo of the song changes the bug behaviour.
 - Changing gate, range, sorting mode dosn't affect the bug.
 - The bug only comes after the first looping (also it disappears from time to time in some strange pattern)

repsej:
A short video with some of the arpeggio problems.

https://youtu.be/tDRmTxxEYSA

musikBear:
interesting! and tchanging the BMP, other weird things happens
when we get below 50 the 'double-note' is replaced by silence. The playhead moves over the note, but there is no sound.
I think that might be the case also at higher bmp, but there its not possible to discriminate the silence  ?
fi:
at 60 bmp. arp: 2 octaves. TeimeSync : ½-note
beat1 sound
beat2 silence
beat3 sound
beat4 silence
or..
1+ 2- 3+ 4-
second run is different:
1* 2- 3- 4+
where '*' is double note
3..8 as 2.
9.
1* 2- 3* 4+
That is reproduceable, and could(?) give insight in the off-sequence

musikBear:
How does current Master do in *this* respect?
Should this actually be marked at least **Bug** or isent it in fact **Critical** ?

Umcaruje:
!? This bug isn't critical, neither its even milestoned, so it isn't such a high priority.

musikBear:
@Umcaruje #2589 was marked critical, this is same thing, but its fine that its now marked 'bug'. :+1 
Sort-mode *may* also be ok in current Master, because #2589 is closed. 
Could you do a test of sort-mode on Master?

zonkmachine:
[...]

Issue #3342, opened 2017-02-10
Tab FUNC / New arpeggio : several issues


musikBear:
OS  : winXP sp3
LMMS 1.1.90-RC2.8

**Issue description:**
Arpeggio is mixed up, and has several errors

**Recreate         :**
New project
insert a c-maj chord with 2 bar length notes
FUNC-TAB/ ARP
Set:
* GATE 10%
* TIME synch to ½ note
* Rest default

play
FREE works as SYNC should work
-All 3 notes in the chord is played simultaneous 
Mark all notes and move them ½ bar fwd
play
There are now one extra hit ½ a bar from the end of the notes
If you move them 1 bar, There are now one extra hit as the playhead starts second cycle at 0

change to SORT
play
One of the notes will not play arpeggio, but plays one long note, then 2 notes play shortly, then 1 note

Change to SYNC
play
first a short all-notes artefact occur. Then one note play a long note, then all notes play simultaneous

Could it bee that a wrong commit has been used for RC?

---
Do anyone want a file?

zonkmachine:
I don't think there is a drastic difference from 1.1.3 .

> There are now one extra hit as the playhead starts second cycle at 0

Thanks! I knew there was an extra hit in there somewhere but I thought it was the master note sounding with the arpeggio note. It's an extra trailing note that comes back as a ghost note when looping.

zonkmachine:
> There are now one extra hit as the playhead starts second cycle at 0

I think I nailed this one at least but the arpeggio sync problems is a bit tricky.

musikBear:
> I think I nailed this one at least 

Good job! Over time i think it would make sense to re-furbish arpeggio. Options like in Oatmeal and Synth, where different delays on selected notes are possible, would open for amazing new sound.

> but the arpeggio sync problems are a bit tricky.

Believe the arp has been bugged from day one, has always only had success with the *sorted* setting

zonkmachine:
Fixed in #3858  #3867

PR #3858, opened 2017-10-05
Fixing the arpeggiator

zonkmachine:
Fixing some of the issues with the arpeggiator.

-  [ ] The decay of the last arpeggio note is added to the length of the master note so after the first run there will be extra notes played along with the arpeggio, forming chords or doubling the notes for increased volume.
 `* * *Fix reverted* * *`

-  [x] Sort mode. If the envelopes are off, the notes that should be silent are playing as ordinary notes while waiting for their turn. Solving this by using the same hack as the 'skip' algorithm as it had the same issue.
Set as master note.

-  [ ]  Sort mode. All notes are playing at the beginning of the notes. Fixed this by simplifying the algorithm to set
`int cur_frame` . A bit experimental this one. This issue was introduced in 1.1 .
`* * *Fix reverted* * * `

Adresses https://github.com/LMMS/lmms/issues/3342

---

**Issues remaining**

 * Sort mode. Sometimes notes are skipped.
 * Sort mode. There is no way to know in which order the notes are being handled without displacing them. This is a pretty good way to deal with it actually but it would be better if the default would be to start with the lowest note and then work our way up. To fix this we probably need to sort the cnphv after key value.
 * Skip function. This may only apply to lb302... If you drag an lb302 over a track with envelopes those affect the lb302 in such a way that notes keep sounding, in a legato kind of way, for as long as notes after it are skipped.

### DIFF
--- a/src/core/InstrumentFunctions.cpp
+++ b/src/core/InstrumentFunctions.cpp
@@ -305,7 +305,7 @@ InstrumentFunctionArpeggio::InstrumentFunctionArpeggio( Model * _parent ) :
 	m_arpCycleModel( 0.0f, 0.0f, 6.0f, 1.0f, this, tr( "Cycle steps" ) ),
 	m_arpSkipModel( 0.0f, 0.0f, 100.0f, 1.0f, this, tr( "Skip rate" ) ),
 	m_arpMissModel( 0.0f, 0.0f, 100.0f, 1.0f, this, tr( "Miss rate" ) ),
-	m_arpTimeModel( 100.0f, 25.0f, 2000.0f, 1.0f, 2000, this, tr( "Arpeggio time" ) ),
+	m_arpTimeModel( 200.0f, 25.0f, 2000.0f, 1.0f, 2000, this, tr( "Arpeggio time" ) ),
 	m_arpGateModel( 100.0f, 1.0f, 200.0f, 1.0f, this, tr( "Arpeggio gate" ) ),
 	m_arpDirectionModel( this, tr( "Arpeggio direction" ) ),
 	m_arpModeModel( this, tr( "Arpeggio mode" ) )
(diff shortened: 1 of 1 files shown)


zonkmachine:
Note:
Arpeggiatore sometimes drops notes when handling more than one base note at a time. This is true for all sync modes and is not related to this PR.

zonkmachine:
This fixes https://github.com/LMMS/lmms/issues/340
Testing appreciated. I'll merge this tomorrow if no one objects.

src/core/InstrumentFunctions.cpp
```diff
@@ -378,9 +378,7 @@ void InstrumentFunctionArpeggio::processNote( NotePlayHandle * _n )
 	// used for calculating remaining frames for arp-note, we have to add
 	// arp_frames-1, otherwise the first arp-note will not be setup
 	// correctly... -> arp_frames frames silence at the start of every note!
-	int cur_frame = ( ( m_arpModeModel.value() != FreeMode ) ?
```
zonkmachine: I'm dropping this test altogether but perhaps it should be kept and changed to `m_arpModeModel.value() = SyncMode` ?

zonkmachine:
I went a head and fixed up some other issues. Merge will wait.

zonkmachine:
I think I'm done with the arpeggiator now for 1.2 . I reverted one of the changes but the rest is fine to go and also took the liberty of bumping the default arp time from 100ms to 200ms. half the speed that is. Else it will just fire away and scare our beloved beginners. The fixes here basically takes care of regressions from 1.1.3.

I'll merge this in a day if there is no objection.

zonkmachine:
Two of the fixes didn't pass testing and had to be reverted. I'll open a new ticket (meta) for the remaining issues once this PR is merged.

PR #3860, opened 2017-10-07
Fix warning in automation unit test

PhysSong:
Fixes wrong usage of `Track::createTCOsForBB()`, which creates BBTCOs instead of automation track's TCOs. It doesn't cause errors now, but it does something definitely wrong.
It caused `called Track::getTCO( 0 ), but TCO 0 doesn't exist` message. This PR fixes that.

Note that the target branch can be switched to `master` safely if this shouldn't go into `stable-1.2`.


tresf:
> Note that the target branch can be switched to `master`

Seems like a better candidate for master. 👍

Issue #3741, opened 2017-07-31
Can not open recent files (KDE)


gi0e5b06:
Qt can add automatically accelarators for each menu entries and KDE does that by default. So a recent file like myproject.mmp becomes mypr&oject.mmp and can not be opened.
As a workaround, if a file is missing and contains an ampersand, the ampersand is removed and the file is tried again. Implemented. Works. May not work if your original filename contains one or more '&'.

PS: implementing a clean solution would require 30-40 lines of code, so way over the allowed PR size limit.

Spekular:
There's no restriction on pull request line length.

zonkmachine:
> would require 30-40 lines of code, so way over the allowed PR size limit.

What size limit? where did you see this?

gi0e5b06:
BTW, that should go into stable-1.2 too.
https://github.com/gi0e5b06/lmms/blob/ab1b45a32a6213ebbe09bda8c809380404c5ae4e/src/gui/MainWindow.cpp#L907
```
void MainWindow::openRecentlyOpenedProject( QAction * _action )
{
	if ( mayChangeProject(true) )
	{
		QString f = _action->text();
		// workaround for KDE
		QFileInfo recentFile(f);
		if(!recentFile.exists()&&(f.indexOf('&')>=0))
		{
			f=f.replace("&","");
		}
		//end
		setCursor( Qt::WaitCursor );
		Engine::getSong()->loadProject( f );
		setCursor( Qt::ArrowCursor );
	}
}
```

irrenhaus3:
@gi0e5b06 If you have a clean solution to the problem, please follow these steps:  
* Create a new git branch in your fork, based on this repo's stable-1.2 branch
* Implement your solution on that branch
* Create a new Pull Request of your branch into stable-1.2 of this repo.
  
There are no size restrictions on PRs, although complex or large changes may take longer to review. See also: https://github.com/LMMS/lmms/wiki/Submitting-a-patch  
Also, could you explain where you saw the stuff about PR size limitations? If that's in the wiki somewhere, it needs to be taken out.

gi0e5b06:
@irrenhaus3 I posted the code. Feel free to go from there.

zonkmachine:
Closed in https://github.com/LMMS/lmms/pull/3872

PR #3872, opened 2017-10-10
Fix recent files

BumblingCoder:
This is a quick fix for issue #3741 which causes templates and recent files to not work in KDE.


tresf:
@BumblingCoder interesting one-liner.  Can you help explain this?

tresf:
Sorry, read the last commit only.  Disregard previous comment. 👍

BumblingCoder:
This is currently still broken if the & character appears in the file name.

tresf:
String replacement can most certainly cause other issues.  Would it be possible to kill this ridiculous behavior?  It's just going to cause more problems.

Here's a solution [from the KDE5 bug tracker](https://bugs.kde.org/show_bug.cgi?id=337491#c21)... 

This solution seems to be written for KDE desktop environments only.  I assume this is safe, since this problem doesn't occur on Mac or Windows but should be tested on Gnome3, Pantheon, whatever isn't Qt5 based but I would expect the `if (!d)` to address it.

> **cmdrkotori 2017-07-22 05:36:43 UTC**
> Here's how I solved it. It's a bit ugly.
> ```c++
> #include <dlfcn.h>
> 
> void Helpers::disableSystemAccel(QWidget *what)
> {
>     void *d = dlopen("libKF5WidgetsAddons.so", RTLD_LAZY);
>     if (!d)
>         return;
>     typedef void (*DisablerFunc)(QWidget *);
>     DisablerFunc setNoAccel;
>     setNoAccel = reinterpret_cast<DisablerFunc>(dlsym(d, "_ZN19KAcceleratorManager10setNoAccelEP7QWidget"));
>     if (setNoAccel)
>         setNoAccel(what);
>     dlclose(d);
> }
> ```
>
> Call this function in your object's constructor.

BumblingCoder:
It looks like that may be the better option. There needs to be some level of string replacement though, because right now files with the & character don't work correctly.

src/gui/MainWindow.cpp
```diff
@@ -76,6 +94,9 @@ MainWindow::MainWindow() :
 	m_metronomeToggle( 0 ),
 	m_session( Normal )
 {
+#ifdef LMMS_BUILD_LINUX
```
tresf: Perhaps a `#ifndef` Mac and Windows check would be sufficient here instead to avoid non-Linux KDEs from being impacted (e.g. BSD, etc).

BumblingCoder: My concern there is that it would be basically untested. The CMake file change would also have to be set up this way as well.

tresf: What do you mean by "untested"?

BumblingCoder: CI doesn't cover platforms other than windows, mac, and linux, and I don't have a machine to attempt building for anything else. This code is very much not portable, and if allowances are being for other platforms the code should be compiled for them at some point.

tresf: @BumblingCoder perhaps there's some confusion... I'm asking you disable the option for Mac + Windows, which will inadvertently enable it for all others.  Enabling it for Linux-only is going to just cause this same issue to happen for (e.g.) OpenBSD users that chose the KDE desktop.  Regardless, it's a safe call for all platforms everywhere since the library won't be found, I was just trying to be consistent with the "probably only on KDE desktop" logic that I presume you were choosing.

BumblingCoder: There is stuff in the CMake files about Haiku. Is this something that we intend to support? I would be more ok making this change say linux or bsd than saying not windows or mac.

tresf: >  I would be more ok making this change say linux or bsd than saying not windows or mac.

Well Haiku isn't BSD, it's BeOS and it's an edge-case (and it can't run the full KDE desktop yet apparently).  However, most Unixes can run KDE.

I'd rather blacklist than whitelist, it's more scalable.  KDE will never be a primary desktop on Windows or MacOS in the foreseeable future.  I can't say the same for new/undocumented Unixes.

BumblingCoder: My concern isn't really about where KDE is common, but about putting in stuff depending on dlopen() and specifying where to run that by specifying the platforms it isn't available or needed on rather than the ones it is. I could go either way on this though.

tresf: > putting in stuff depending on `dlopen()` and specifying where to run that by specifying the platforms it isn't available or needed

Great point.  As BeOS seems to be the major exception from what we support.  Even [Solaris supports `dlopen()`](https://docs.oracle.com/cd/E23824_01/html/821-1465/dlopen-3c.html).   Until we require `dlopen` for something else, I would suggest simply blacklisting Haiku along side Windows and Mac then.  Down the road if we need `dlopen` for a future feature, we can [leverage a cross-platform wapper](https://www.gnu.org/software/libtool/manual/html_node/Using-libltdl.html).

src/gui/MainWindow.cpp
```diff
@@ -64,6 +64,24 @@
 
 #include "lmmsversion.h"
 
+#ifdef LMMS_BUILD_LINUX
+#include <dlfcn.h>
+//This function is slightly modified from a solution to a KDE bug
+//Found at https://bugs.kde.org/show_bug.cgi?id=337491#c21
+void DisableSystemAccel(QWidget *what)
+{
+	void *d = dlopen("libKF5WidgetsAddons.so", RTLD_LAZY);
```
tresf: Nitpicking, but... 
1. We should prefer something more descriptive such as `void *handle = dlopen(...)`.
2. This function name should start with a lowercase letter.
3. I find the name to be agnostic to hardware acceleration, please give it a name which is less agnostic, e.g. `disableAutoMnemonics` or something about "[Keys](http://doc.qt.io/qt-4.8/accelerators.html)".
4. We have a lot of workaround code we've borrowed and a simple link should be enough.  e.g. "Disable auto-keyboard accelerators for KDE per https://bugs.kde.org/show_bug.cgi?id=337491"

BumblingCoder: The function name is referring to keyboard accelerators, but I agree, the naming isn't great.

BumblingCoder:
The latest version of this should work on any platform it is needed on.

gi0e5b06:
@BumblingCoder >This is currently still broken if the & character appears in the file name.

You can use this:
`		{
			f=f.replace("&&","£lukas-w closed my PR£");
			f=f.replace("&","");
			f=f.replace("£lukas-w closed my PR£","&");
		}
`

src/CMakeLists.txt
```diff
@@ -147,6 +152,7 @@ SET(LMMS_REQUIRED_LIBS
 	${SAMPLERATE_LIBRARIES}
 	${SNDFILE_LIBRARIES}
 	${EXTRA_LIBRARIES}
+	${KDE_FIX_LIBRARIES}
```
zonkmachine: The PR can't be merged since rpmalloc. Reabse and just stick rpmalloc to the end.

zonkmachine:
I don't have KDE but templates and recent files works as it should with this PR applied under linuxmint/mate.

BumblingCoder:
The bug was only present with Qt5 on KDE. The issue is that on KDE Qt loads platform theme stuff that adds accelerators to all of the menu items. I guess it is good to know that it doesn't break things on other desktops though. :)

zonkmachine:
> I guess it is good to know that it doesn't break things on other desktops

Yeah, I tested what I could.

Do you want me to push the rebased version to `BumblingCoder:fix-recent-files` ?

BumblingCoder:
I'm fine with you pushing the rebase. The soonest I could do it is about 6 hours from now, and I'm not particularly familiar with git anyway.

zonkmachine:
Aight, fixed! Looks about right... :)

src/gui/MainWindow.cpp
```diff
@@ -64,6 +64,26 @@
...
+	void *libraryHandle = dlopen("libKF5WidgetsAddons.so", RTLD_LAZY);
+	if (!libraryHandle) {
+		//KDE not installed, nothing to do.
+		return;
+	}
+	using DisablerFunc = void(*)(QWidget*);
+	DisablerFunc setNoAccelerators =
+			reinterpret_cast<DisablerFunc>(dlsym(libraryHandle,
+					"_ZN19KAcceleratorManager10setNoAccelEP7QWidget"));
+	if (setNoAccelerators) {
```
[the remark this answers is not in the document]
BumblingCoder: There isn't actually anything in that document that says that braces should be on the next line. ( Or in what cases they should.)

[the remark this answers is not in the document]
tresf: More of a consistency with the file being edited than a formal convention.  It's fine as-is.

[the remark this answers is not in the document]
zonkmachine: OK. We should look over the documentation if there's something missing.
Is the braces the only thing that's holding this PR back now?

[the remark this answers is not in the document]
PhysSong: The braces don't matter too much, but I think it needs some more works(ex. `libKF5WidgetsAddons.so.5` for ubuntu-based systems, (optional)using `QLibrary`).

src/gui/MainWindow.cpp
```diff
@@ -70,7 +70,7 @@
 void disableAutoKeyAccelerators(QWidget* mainWindow)
 {
 	using DisablerFunc = void(*)(QWidget*);
-	QLibrary kf5WidgetsAddon("libKF5WidgetsAddons.so");
+	QLibrary kf5WidgetsAddon("KF5WidgetsAddons");
```
PhysSong: Good, but it may not work on Ubuntu if `libkf5widgetsaddons-dev` is not installed. A simple solution is here:
```
// try with version number(for some systems like Ubuntu)
if (!kf5WidgetsAddon.load())
{
	kf5WidgetsAddon.setFileNameAndVersion("KF5WidgetsAddons", 5);
	kf5WidgetsAddon.load();
}
```

src/gui/MainWindow.cpp
```diff
@@ -64,6 +65,21 @@
 
 #include "lmmsversion.h"
 
+#if !defined(LMMS_BUILD_WIN32) && !defined(LMMS_BULID_APPLE) && !defined(LMMS_BUILD_HAIKU)
```
PhysSong: Is this check still needed? Since @BumblingCoder switched the loader to `QLibrary`, it can be removed safely. If there are some reason to keep this, however, leaving is okay, too.

BumblingCoder: It should be safe to run the code on all platforms. Pointless on the ones it is disabled on, but safe.

tresf: It saves a few cpu cycles to preprocess it out.  I agree with the logic, especially when loading libraries.  Why add an unnecessary step to MainWindow for systems which can't experience the bug?

PhysSong: > Why add an unnecessary step to MainWindow for systems which can't experience the bug?

Okay then. Keep it if leaving makes sense.

src/gui/MainWindow.cpp
PhysSong: @BumblingCoder May I push a commit addressing this? Or you may do that yourself if you want.

BumblingCoder: If you want to pull this tonight you can do it, otherwise I can deal with it tomorrow evening. I think it is safe to always try to load version 5 though, so we could just throw the 5 in the constructor.

PhysSong: > I think it is safe to always try to load version 5 though, so we could just throw the 5 in the constructor.

I'm not quite sure, but it should work if `libKF5WidgetsAddons.so.5` exists. Since the bug is a KDE 5 + Qt 5 issue, that looks safe for me. I'll wait for you, since you said you can do that tomorrow.

PhysSong:
Merge?

tresf:
:+1:

src/gui/MainWindow.cpp
```diff
@@ -1500,7 +1518,7 @@ void MainWindow::fillTemplatesMenu()
 	{
 		m_templatesMenu->addAction(
 					embed::getIconPixmap( "project_file" ),
-					( *it ).left( ( *it ).length() - 4 ) );
+					( *it ).left( ( *it ).length() - 4 ).replace("&", "&&") );
```
lukas-w: Should line `1500` have this change as well?

PR #3878, opened 2017-10-13
Track operations fixes and other stuff.

### LINKED ISSUE #3592: Moving and dragging & dropping tracks enhancement.
* When clicking on a "move-grip" or the track label  or any empty space in a track, this is what happens:
bug 1
"Move-grip", "Actions for this track" and "Mute this track" disappear but "Solo" doesn't.

* Dragging & dropping a track can only be done after pressing & holding Ctrl then clicking & dragging from the "move-grip", but moving a track isn't limited to the "move-gip" only.
I think we should limit the ability of moving and dragging & dropping tracks to these:
image

Last **edit** was in 13 October, 2017.

Sawuare:
Fixes #3592 and other stuff.

Before | After
-------|------
image | image

**Changes:**
• Clicking the move grip of a track doesn't hide the track's mute and options buttons.
• Clicking the move grip of a track changes the move grip's image to one with white squares instead of gray.
• Changed the cursor shape that shows when moving tracks to be vertical.
• Improved the position and padding of the gear image.
• Removed the `trackop_c.png` and `trackop_h.png` files and some of their related code in the `style.css` file of the default theme.


gi0e5b06:
I can only approve (as I have made the same changes in my PR, including the TODO item).

Sawuare:
Could you commit the fix for the TODO item here?

gi0e5b06:
It's here: https://github.com/gi0e5b06/lmms/blob/master/src/core/Track.cpp
Moving is limited to the grip.
I don't remember if I changed something for the drag&drop. Anyway I think it should cover the whole area, the difference being made by the mimetype.

gi0e5b06:
Also I like the lines 1876-1880: https://github.com/gi0e5b06/lmms/blob/86a54872b9d00fbfc338f1864a903bf4d690f962/src/core/Track.cpp#L1876

```
if( m_trackView->isMovingTrack() )
	{
		p.setPen(Qt::red);
		p.drawLine(r.x(),r.y(),r.x(),r.y()+r.height()-1);
	}
```
It draws a small red line on the left border when you're moving the track.

Sawuare:
I tested your suggestion.
The line makes the grip look crammed with the white squares, and isn't enough alone.
I think the white squares on the grip are enough.

tresf:
Note, this prevents the previous behavior of dragging a track by "any random empty part of the track label" behavior that had been working previously.

Since no one has made a UX argument against this, and it actually fixes the drag handle, merging.  If you want the old way, you'll have to argue it back.  From a UI perspective we really don't allow rearranging of items like this in any other part of the software except the pattern editors, which aren't UI input elements.

Although I personally prefer a "long hold, then drag" like most modern software uses, that should be filed as a new feature.

src/core/Track.cpp
```diff
@@ -2713,6 +2710,12 @@ void TrackView::dropEvent( QDropEvent * de )
  */
 void TrackView::mousePressEvent( QMouseEvent * me )
 {
+	if( me->x()>10 ) // 10 = The width of the grip + 2 pixels to the left and right.
```
PhysSong: This breaks horizontal track resizing. I think this check should be moved to below.

src/core/Track.cpp
```diff
@@ -2745,7 +2748,7 @@ void TrackView::mousePressEvent( QMouseEvent * me )
 		{
 			m_action = MoveTrack;
```
PhysSong: To here.

Sawuare: Tested. Fixes the issue as expected.

Sawuare: Direct commit or PR?

PhysSong: Direct commit looks fine.

Sawuare: Done via https://github.com/LMMS/lmms/commit/e8debf9015979c5d51e5440a78b4dbb2ea2e91b9. 👍

Issue #3880, opened 2017-10-14
Arpeggiator overhaul


zonkmachine:
Some issues with the arpeggiator and the note sorting algorithms.

If you play an arpeggio over one note only things generally work fine. There is the issue with occasionally dropped notes but if you use the function to skip notes and only arpeggiate over one note at a time, there is a chance that you will experience not a single glitch. If however you choose to arpeggiate over a chord, then you will most likely bump into a couple of the following issues.

---

 - [ ] Skipped/extra notes. Sometimes notes are skipped or extra notes are sounding. Such an extra note can be the same as the actual note intended, increasing its volume.

 - [ ] The notes are only sorted after position and not after key. This is apparent in the arpeggiator when you're trying to arpeggiate over more than one key. It's difficult to get the notes to follow a defined route.
Proposed fix: 3862cfe3d1346e11a6dc5dff44b1e8a9b21b1bb5

 - [ ] The order of notes handed to the arpeggiator is inverted. If you're arpeggiating over more than one note the starting point is the highest note and direction down. This is probably a unique response for lmms as the usual is to start from the lowest one and walking up from there. Maybe an option to switch direction?
This may be a regression but because of all the other issues in the arpeggiator at the earlier version it's difficult to test properly.

 - [ ] Sorting notes is made both in `Pattern::addNote(...)` and in `Pattern::rearrangeAllNotes()`. The former is triggered on key 'on' and the latter on key 'off' but also works more generally on actions in the PianoRoll. The sort in `Pattern::addNote(...)` may thus be unnecessary. I've tested to drop this code and it seem to work just fine. Notes recorded through Midi keyboard are sorted only through Pattern::addNote so some rework seem to be needed to use `Pattern::rearrangeAllNotes()`.
Test: 84abaae3d31bffe07191f2a80bed9cab32111bc1

 - [ ] Pressing an existing note in the Piano Roll plays it back. When you release the key this triggers a note sort via `Pattern::rearrangeAllNotes():` the same way as a drag/shift action. This even though nothing really has changed to the pattern. It may be a non issue as the extra cpu cycles involved doesn't appear to be screaming high.

-  [x]  Sort mode. All notes are playing at the beginning of the notes. This is a regression from 1.1 and was introduced in 6650dd356dbdd08739ad69153b0c7020e39787b9.
Fixed in 2d583db990385506bff7020f4b4109976af5835f

[...]

Issue #3879, opened 2017-10-13
Crash on 1-tick loops


gi0e5b06:
LMMS crashes on playing 1-tick loops, but 2-ticks loops are fine.
(rc4 master alsa)
[memory20171013.zip](https://github.com/LMMS/lmms/files/1383828/memory20171013.zip)

qnebra:
In my case lmms holds this 1-tick loop fine, very glitchy but no crash at all. But I have compiled version with merged bunch of PRs.

gi0e5b06:
You tried the new file? If yes, then that means it is not a logical mistake...
(btw, which OS and audio?)
Not an out of memory mistake.
Maybe a memory mistake (Segmentation fault (core dumped)) thread-related?
For me, it is not glitchy (cpu usage is low because only the instruments that don't use a lot of cpu are active) but it crashes after half a second. No problem with the 2-tick loop.

qnebra:
I tried this one you provide in first post of this thread.
(Ubuntu 17.10, lmms use SDL2)

gi0e5b06:
I tried the other audio drivers, same problem.
I run it under gdb and it doesn't crash.
It seems @PhysSong has the same problem (or at least it looks like) #3881 (8th comment).

PhysSong:
I got a crash and the backtrace looks like one in #777. If there are no evidences which show this isn't #777, this should be closed as a duplicate.

PhysSong:
You may want to take a look at #2539.

PhysSong:
@gi0e5b06 Have you got any backtrace(s)? If so, please post it. It will help a lot.

gi0e5b06:
Thanks @qnebra and @PhysSong, it helps a lot. It was indeed #777 (May 28, 2014).
I changed the size of the queue and it doesn't crash anymore:
MixerWorkerThread.h:
```
#define JOB_QUEUE_SIZE 2048`
```
I also added this:
MixerWorkerThread::JobQueue:
```
		if( (int) m_queueSize >= JOB_QUEUE_SIZE-16 )
			qFatal("MixerWorkerThread::JobQueue full (%d)",JOB_QUEUE_SIZE);
```
I haven't had time to fix the problem but I will. It just needs a better queue.

PR #3881, opened 2017-10-14
Note sorting algorithm rework

zonkmachine:
Addresses some of the issues in https://github.com/LMMS/lmms/issues/3880

 Sorting after position and key. Probably only needed with the arpeggiator. This works well and doesn't seem to use a whole lot cpu cycles. The notes are sorted on input from mouse or keyboard in ` Pattern::addNote()`. Actions on the notes in the Piano Roll also triggers a sort via `Pattern::rearrangeAllNotes()` with a helper function in note.h . The sorting is duplicated when adding notes in the Piano Roll and this has its own post in https://github.com/LMMS/lmms/issues/3880 but is probably a wontfix.

I've pulled these against stable-1.2 but they can just as well go into master or be split over the two.

### DIFF
--- a/src/tracks/Pattern.cpp
+++ b/src/tracks/Pattern.cpp
@@ -271,7 +271,7 @@ Note * Pattern::noteAtStep( int _step )
 void Pattern::rearrangeAllNotes()
 {
 	// sort notes by start time
-	qSort(m_notes.begin(), m_notes.end(), Note::lessThan );
+	std::sort(m_notes.begin(), m_notes.end(), Note::lessThan);
 }
 
 
--- a/src/tracks/Pattern.cpp
+++ b/src/tracks/Pattern.cpp
@@ -213,28 +213,7 @@ Note * Pattern::addNote( const Note & _new_note, const bool _quant_pos )
 	}
 
 	instrumentTrack()->lock();
-	if( m_notes.size() == 0 || m_notes.back()->pos() <= new_note->pos() )
-	{
-		m_notes.push_back( new_note );
-	}
-	else
-	{
-		// simple algorithm for inserting the note between two
-		// notes with smaller and greater position
-		// maybe it could be optimized by starting in the middle and
-		// going forward or backward but note-inserting isn't that
-		// time-critical since it is usually not done while playing...
-		long new_note_abs_time = new_note->pos();
-		NoteVector::Iterator it = m_notes.begin();
-
-		while( it != m_notes.end() &&
-					( *it )->pos() < new_note_abs_time )
-		{
-			++it;
-		}
-
-		m_notes.insert( it, new_note );
-	}
+	m_notes.insert(std::upper_bound(m_notes.begin(), m_notes.end(), new_note, Note::lessThan), new_note);
 	instrumentTrack()->unlock();
 
 	checkType();
--- a/include/Note.h
+++ b/include/Note.h
@@ -113,11 +113,19 @@ public:
 	void quantizeLength( const int qGrid );
 	void quantizePos( const int qGrid );
 
-	static inline bool lessThan( Note * &lhs, Note * &rhs )
+	static inline bool lessThan( const Note * lhs, const Note * rhs )
 	{
 		// function to compare two notes - must be called explictly when
 		// using qSort
-		return (bool) ((int) ( *lhs ).pos() < (int) ( *rhs ).pos());
+		if( (int)( *lhs ).pos() < (int)( *rhs ).pos() )
+		{
+			return true;
+		}
+		else if( (int)( *lhs ).pos() > (int)( *rhs ).pos() )
+		{
+			return false;
+		}
+		return ( (int)( *lhs ).key() > (int)( *rhs ).key() );
 	}
 
 	inline bool selected() const


zonkmachine:
The simplest approach here may be to just duplicate the position/key sorting algorithm for Pattern::addNote().

zonkmachine:
The latest (third) commit reverts the first one and implement a position/key note sort for `Pattern::addNote()`

Here is an example output from playing chords and single notes with a keyboard.
```
        <pattern steps="16" muted="0" type="1" name="TripleOscillator" pos="0">
          <note pan="0" key="41" vol="173" pos="72" len="96"/>
          <note pan="0" key="47" vol="173" pos="72" len="96"/>
          <note pan="0" key="50" vol="168" pos="72" len="96"/>
          <note pan="0" key="50" vol="179" pos="252" len="84"/>
          <note pan="0" key="45" vol="173" pos="264" len="84"/>
          <note pan="0" key="53" vol="165" pos="264" len="84"/>
          <note pan="0" key="41" vol="182" pos="420" len="84"/>
          <note pan="0" key="47" vol="176" pos="468" len="24"/>
          <note pan="0" key="50" vol="168" pos="540" len="24"/>
        </pattern>
```

zonkmachine:
Note that old patches will open up like before and all will be fine. The next time you edit a track it will be sorted with the new method and this will probably be only an advantage. Probably. Maybe you like the notes to shift around a bit, and you'll only really notice this in arpeggiated tracks with more than two simultaneous notes.
If this is not wanted you can just _open/press a note in the piano roll to trigger a sort/save_ the project in 1.1.3 and it'll go wonky again.

zonkmachine:
Got a random crash. I don't think it's related to this PR though. The project had been running for maybe an hour or so. I was looping two 3osc tracks with an automation track for the Master pitch, two notes detuned in the one track and an LFO detuning the other synth.
`m_frequencyNeedsUpdate = true;` LFO ... or Note detune ... or Master pitch?

[bt](https://gist.github.com/zonkmachine/211524d32e670682c8afbdaa1d2c6ccc#file-gistfile1-txt-L4) / [bt full](https://gist.github.com/zonkmachine/211524d32e670682c8afbdaa1d2c6ccc#file-gistfile1-txt-L51)
```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x000000000062e0ea in NotePlayHandle::setFrequencyUpdate (this=0x81) at /home/zonkmachine/builds/lmms/lmms/include/NotePlayHandle.h:263
263			m_frequencyNeedsUpdate = true;
(gdb) bt
#0  0x000000000062e0ea in NotePlayHandle::setFrequencyUpdate (this=0x81) at /home/zonkmachine/builds/lmms/lmms/include/NotePlayHandle.h:263
#1  0x00000000006270a1 in InstrumentTrack::updateBaseNote (this=0x7fce29e77620) at /home/zonkmachine/builds/lmms/lmms/src/tracks/InstrumentTrack.cpp:543
#2  0x000000000062710a in InstrumentTrack::updatePitch (this=0x7fce29e77620) at /home/zonkmachine/builds/lmms/lmms/src/tracks/InstrumentTrack.cpp:552
#3  0x000000000063f89f in InstrumentTrack::qt_static_metacall (_o=0x7fce29e77620, _c=QMetaObject::InvokeMetaMethod, _id=6, _a=0x7fffe6625550)
    at /home/zonkmachine/builds/lmms/lmms/build/src/moc_InstrumentTrack.cpp:68
#4  0x00007fce35cf187a in QMetaObject::activate (sender=0x7fce29e78898, m=<optimized out>, local_signal_index=<optimized out>, argv=0x0)
    at kernel/qobject.cpp:3539
#5  0x0000000000642f4b in Model::dataChanged (this=0x7fce29e78898) at /home/zonkmachine/builds/lmms/lmms/build/src/moc_Model.cpp:102
#6  0x0000000000642e09 in Model::qt_static_metacall (_o=0x7fce29e78898, _c=QMetaObject::InvokeMetaMethod, _id=0, _a=0x7fffe66256b0)
    at /home/zonkmachine/builds/lmms/lmms/build/src/moc_Model.cpp:51
#7  0x00007fce35cf187a in QMetaObject::activate (sender=0x2ee0210, m=<optimized out>, local_signal_index=<optimized out>, argv=0x0)
    at kernel/qobject.cpp:3539
#8  0x000000000063b4ab in ControllerConnection::valueChanged (this=0x2ee0210)
    at /home/zonkmachine/builds/lmms/lmms/build/src/moc_ControllerConnection.cpp:104
#9  0x000000000063b354 in ControllerConnection::qt_static_metacall (_o=0x2ee0210, _c=QMetaObject::InvokeMetaMethod, _id=0, _a=0x7fcde40521b0)
    at /home/zonkmachine/builds/lmms/lmms/build/src/moc_ControllerConnection.cpp:52
```

PhysSong:
`NotePlayHandle::setFrequencyUpdate (this=0x81)` was called in
https://github.com/LMMS/lmms/blob/54f3eccad77103ff7441470cbfc5739180f3c1ba/src/tracks/InstrumentTrack.cpp#L543
and it may indicate an invalid reference of iterator occured.

PhysSong:
`m_processHandles` is used in non-thread-safe way(i.e. without proper synchronization). `(this=0x81)` means `*it == 0x81`, which is an invalid address.
Anyway, I'm pretty sure it's unrelated with this PR. For this kind of crashes, I'll open a separate issue. :)

gi0e5b06:
#3879 seems related to this NotePlayHandle race condition.

zonkmachine:
This should go into master. I'll switch it over.

zonkmachine:
I switched back to stable 1.2 and this is ready for a review. Switching sorting order was a matter of flipping over two 'lesser than' signs and this is now less buggy than before. See the note on backward compatibility [here](https://github.com/LMMS/lmms/pull/3881#issuecomment-336606142).

src/tracks/Pattern.cpp
```diff
@@ -213,7 +213,7 @@ Note * Pattern::addNote( const Note & _new_note, const bool _quant_pos )
 	}
 
 	instrumentTrack()->lock();
-	if( m_notes.size() == 0 || m_notes.back()->pos() <= new_note->pos() )
+	if( m_notes.size() == 0 || m_notes.back()->pos() < new_note->pos() )
```
lukas-w: As far as I understand the code, all lines 216 to 245 do is insert the new note at the right position (sorted by the criteria in `lessThan`). If so, we can reduce this to a single line:
```cpp
m_notes.insert(std::upper_bound(m_notes.begin(), m_notes.end(), new_note, Note:lessThan), new_note);
```
This reduces code complexity, improves performance by doing a binary search, and removes the duplicate compare code by using the `lessThan` method.

zonkmachine: > As far as I understand the code, all lines 216 to 245 do is insert the new note at the right position (sorted by the criteria in lessThan).

Yes indeed. Oneliner, nice but I can't make it work.

Computer says no. 8)

```cpp
/usr/include/c++/4.8/bits/stl_algo.h: In instantiation of ‘_FIter std::upper_bound(_FIter, _FIter, const _Tp&, _Compare) [with _FIter = Note**; _Tp = Note*; _Compare = bool (*)(Note*&, Note*&)]’:
/home/zonkmachine/builds/lmms/lmms/src/tracks/Pattern.cpp:216:90:   required from here
/usr/include/c++/4.8/bits/stl_algo.h:2543:31: error: invalid initialization of reference of type ‘Note*&’ from expression of type ‘Note* const’
    if (__comp(__val, *__middle))
                               ^
```

lukas-w: Ah, you'll have to change `Note::lessThan` to take `const` pointers.

zonkmachine: Perfect, thanks!

zonkmachine:
I took the opportunity to practice and updated the other sorting function to std::sort() as qSort() is obsolete.

Now I think I'm done. The notes are correctly sorted for both functions and there is no code changes needed to save/load the patterns correctly.

PR #3882, opened 2017-10-14
Benchmark the user/system time usage for exports. 

softrabbit:
I think this could be useful for quick checks on what the effects on performance are when modifying code. Startup, project loading and teardown are ignored as this only concerns the export phase.

Add -DWANT_DEBUG_RTIME=true to the cmake options to enable this feature.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,6 +67,7 @@ OPTION(WANT_VST_NOWINE	"Include partial VST support (without wine)" OFF)
 OPTION(WANT_WINMM	"Include WinMM MIDI support" OFF)
 OPTION(WANT_QT5		"Build with Qt5" OFF)
 OPTION(WANT_DEBUG_FPE	"Debug floating point exceptions" OFF)
+OPTION(WANT_DEBUG_RTIME "Debug render times" OFF)
 
 
 IF(LMMS_BUILD_APPLE)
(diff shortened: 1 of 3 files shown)


tresf:
Any major downside to just having this on all the time?  This would be helpful for all releases I think. :+1:

softrabbit:
> Any major downside to just having this on all the time? This would be helpful for all releases I think

This code only runs before and after export, so there shouldn't be any apart from cluttering the terminal one starts LMMS from. And that's only one line.

gi0e5b06:
I totally agree with the idea. Especially to have it encapsulated inside a pragma.
Is the name standard? (it is not really debugging)
Most important: a standard printf format so we can extract and compare the results easily.

gi0e5b06:
A bit more thinking.

1) It would be nice ti benchmark a few other main tasks:
- lmms loading
- project loading
- project saving

2) These tasks are long-lasting so don't need clock resolution. A portable way to measure the duration should be possible. OTOH, having user time and system time is important too.

3) Output should have the following:
- "PERFLOG"
- date, time (Y:m:d H:i:s)
- task name ("ExportingSong", "ProjectLoading", ...)
- times
- project name (optional)
- __FILE__, __LINE__ (optional)

4) -DWANT_PERFLOG

gi0e5b06:
I wrapped your code into a class (and two macros)
https://github.com/gi0e5b06/lmms/commit/dfbb3b0c8998d91c934a900d9040ed5c5ea5d332

softrabbit:
@gi0e5b06 Looks good to me, make a PR from that class and we can continue there.

tresf:
> make a PR from that class and we can continue there./

Done via #3974.

lukas-w:
Superseded by #3974.

PR #3883, opened 2017-10-15
Fix sfxr buffer noise

### LINKED ISSUE #3383: sfxr - buffer noise
The usual buffer stuff. There is noise with other instruments with the decay if using the general envelope (ENV/LFO).

_Example. May get a bit loud..._
[sfxrbuffer.mmp.zip](https://github.com/LMMS/lmms/files/798365/sfxrbuffer.mmp.zip)

PhysSong:
Fixes #3383 by zeroing buffer out. Actually I don't exactly know why it works...

### DIFF
--- a/plugins/sfxr/sfxr.cpp
+++ b/plugins/sfxr/sfxr.cpp
@@ -464,6 +464,7 @@ void sfxrInstrument::playNote( NotePlayHandle * _n, sampleFrame * _working_buffe
 	}
 	else if( static_cast<SfxrSynth*>(_n->m_pluginData)->isPlaying() == false )
 	{
+		memset(_working_buffer + offset, 0, sizeof(sampleFrame) * frameNum);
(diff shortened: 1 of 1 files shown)


zonkmachine:
> Actually I don't exactly know why it works...

It doesn't. The bug is apparently intermittent. It takes some tweaking. In the test project in #3383 you can turn down the decay on sfxr (on sfxr, not the general pluggin). That triggered it for me.

zonkmachine:
I tested this wrong. Sorry, redoing it!

zonkmachine:
Tested it again properly. The fix off course works. :tractor: 
I think you can go ahead and just merge it.

zonkmachine:
> Actually I don't exactly know why it works...

It's the same noise and the same solution as for the buffer noise in sfxr and it works fine. All I need to know ;) .

Issue #3775, opened 2017-08-24
Envelope + (Rapidly) varying tempo give undesirable output.


PhysSong:
[TempoNaN.mmp.zip](https://github.com/LMMS/lmms/files/1247655/TempoNaN.mmp.zip)

The tempo model is connected to a LFO controller. Instruments with envelope sounds weird, some notes may be skipped.
If arpeggio is enabled, it sounds more glitchy. Infs and/or NaNs are also generated.

PhysSong:
`NotePlayHandle::resize( const bpm_t _new_tempo )` changes `m_totalFramesPlayed`, so `EnvTotalFrames` and `EnvReleaseBegin` in `InstrumentSoundShaping::processAudioBuffer()` can have wrong values.
`NotePlayHandle::resize()` has introduced in 52d9ca47, which was done in 2007. I don't know why it exists and why it should be there.

PhysSong:
That function is there because notes should be resized when the tempo changes. However, it shouldn't change `m_totalFramesPlayed`. Instead of this, **left frames** should be recalculated.
There are some exceptions, like notes from MIDI input or arpeggio sub-notes(if tempo-sync is disabled, at least). Creating a new variable `m_shouldBeResized` may be handy, in my opinion.

zonkmachine:
This sounds to me like its another case where it would be easier to trouble shoot/test if https://github.com/LMMS/lmms/issues/2606 is fixed as some of the symptoms are overlapping. Dropped notes etc.

PR #3884, opened 2017-10-15
Fix destructor call in NotePlayHandleManager

PhysSong:
Fixes a regression from #3783 due to missing destructor call. It should have been added in 42e67d2, but not.
Since `BufferManager::releaseBuffer()` call is moved in #3783, buffer isn't freed(by `MM_FREE()`) in `NotePlayHandle::done()`. It isn't a memory leak because  `MemoryManager` works as a garbage collector, but decreases usable memory.
This PR also makes sure `PlayHandle::m_processingLock`'s destructor is called.

### DIFF
--- a/src/core/NotePlayHandle.cpp
+++ b/src/core/NotePlayHandle.cpp
@@ -128,7 +128,7 @@ NotePlayHandle::NotePlayHandle( InstrumentTrack* instrumentTrack,
 }
 
 
-void NotePlayHandle::done()
+NotePlayHandle::~NotePlayHandle()
 {
 	lock();
 	noteOff( 0 );
@@ -599,7 +599,7 @@ NotePlayHandle * NotePlayHandleManager::acquire( InstrumentTrack* instrumentTrac
 
 void NotePlayHandleManager::release( NotePlayHandle * nph )
 {
-	nph->done();
+	nph->NotePlayHandle::~NotePlayHandle();
 	s_mutex.lockForRead();
 	s_available[ s_availableIndex.fetchAndAddOrdered( 1 ) + 1 ] = nph;
 	s_mutex.unlock();
--- a/include/NotePlayHandle.h
+++ b/include/NotePlayHandle.h
@@ -66,8 +66,7 @@ public:
 					NotePlayHandle* parent = NULL,
 					int midiEventChannel = -1,
 					Origin origin = OriginPattern );
-	virtual ~NotePlayHandle() {}
-	void done();
+	virtual ~NotePlayHandle();
 
 	void * operator new ( size_t size, void * p )
 	{


zonkmachine:
I tested it with the 'football theme' from the original issue, https://github.com/LMMS/lmms/issues/3223, and it works fine.

src/core/NotePlayHandle.cpp
```diff
@@ -160,6 +160,8 @@ void NotePlayHandle::done()
 	if( buffer() ) releaseBuffer();
 
 	unlock();
+
+	this->PlayHandle::~PlayHandle();
```
lukas-w: It should never be necessary to call destructors explicitly in C++. If `PlayHandle`'s destructor is never called to release the buffer, something else is wrong here - if it is called, this patch will lead to the buffer being freed twice.

PhysSong: @lukas-w Usually yes, but I think *no* in this situation. Since `NotePlayHandle` uses memory allocated by `NotePlayHandleManager` and its destructor is never called, `NotePlayHandle::done()` should do what destructor does.
If you really don't like this, you may consider redesigning or removing `NotePlayHandleManager`.

PhysSong: Note: https://stackoverflow.com/questions/14187006/is-calling-destructor-manually-always-a-sign-of-bad-design

lukas-w: Then it's a bug in `NotePlayHandleManager`. The destructor should be called in `NotePlayHandleManager::release` because it's `NotePlayHandleManager::acquire` where placement-new is used. Also, it looks like `NotePlayHandle::done()`'s code actually belongs in `~NotePlayHandle`.

PhysSong: Then renaming `NotePlayHandle::done()` to `NotePlayHandle::~NotePlayHandle()` and calling it in the `NotePlayHandleManager::release` would just be fine?

lukas-w: Yep, I think it's fine this way 👍 As far as I can see, `done()` is only called in `NotePlayHandleManager::release()` and simply plays the role of a destructor.

PhysSong:
Now `NotePlayHandle::release()` will call parent classes' destructor correctly, always once. #1088 looks really bad for now...

PhysSong:
Side note: `NotePlayHandleManager` seems not safe from [ABA problem](https://en.wikipedia.org/wiki/ABA_problem). 
For example, `s_available[ s_availableIndex.fetchAndAddOrdered( 1 ) + 1 ] = nph;` isn't an atomic operation. It consists of two atomic operations: fetch-and-add `s_availableIndex` and store in `s_available` with the index. If a context switch occurs between them and some other operations are done, it may lead to wrong results.
Though it won't make a trouble frequently(since it is a rare case), this should be fixed later if it matters.

zonkmachine:
Cool destructor. You can even hear it working... . ;)
(turn the volume down and loop it)
[destructor.mmp.zip](https://github.com/LMMS/lmms/files/1390467/destructor.mmp.zip)

lukas-w:
@zonkmachine That file stops playing for me after a couple of seconds both with and without this PR merged.

zonkmachine:
I bisected wrongly. Loosing my favourite game. :(

lukas-w:
Don't worry, you'll have enough chances (a.k.a. "bugs") to play this game again. :)

lukas-w:
@zonkmachine I did some debugging and the error seems to be within bitinvader not zeroing the buffer, so I added a line in `PlayHandle` do just that and guard against instruments that don't (#3883 comes to mind). Can you confirm 6fc4577f102e5bb60cc246a426fb9ed8b4ac3be9 fixes it?

zonkmachine:
It's definitely an improvement. There is no longer a thunderclap coming out of the reverb. I still get a NaN and the noise cuts out though when moving the tempo slider while looping.

I ran it with the fpe debug and got the backtrace below. It reminds me of a comment [here](https://github.com/LMMS/lmms/issues/3408#issuecomment-284881794) which also had a problem file with a bitinvader in it.

```
[Switching to Thread 0x7fffdddde700 (LWP 19879)]
0x000000000052499f in InstrumentSoundShaping::processAudioBuffer (this=0x7fffeab2aca0, buffer=0x7fffeab2f5e0, frames=256, n=0x7fffeab218e8)
    at /home/zonkmachine/builds/lmms/lmms/src/core/InstrumentSoundShaping.cpp:206
206					float new_cut_val = EnvelopeAndLfoParameters::expKnobVal( cutBuffer[frame] ) *
(gdb) bt
#0  0x000000000052499f in InstrumentSoundShaping::processAudioBuffer (this=0x7fffeab2aca0, buffer=0x7fffeab2f5e0, frames=256, n=0x7fffeab218e8)
    at /home/zonkmachine/builds/lmms/lmms/src/core/InstrumentSoundShaping.cpp:206
#1  0x000000000062b3ae in InstrumentTrack::processAudioBuffer (this=0x7fffeab29720, buf=0x7fffeab2f5e0, frames=256, n=0x7fffeab218e8)
    at /home/zonkmachine/builds/lmms/lmms/src/tracks/InstrumentTrack.cpp:203
#2  0x00007fffdc3b3704 in bitInvader::playNote (this=0x7fffeab2c5a0, _n=0x7fffeab218e8, _working_buffer=0x7fffeab2f5e0)
    at /home/zonkmachine/builds/lmms/lmms/plugins/bit_invader/bit_invader.cpp:301
#3  0x000000000062c41e in InstrumentTrack::playNote (this=0x7fffeab29720, n=0x7fffeab218e8, workingBuffer=0x7fffeab2f5e0)
    at /home/zonkmachine/builds/lmms/lmms/src/tracks/InstrumentTrack.cpp:483

(gdb) bt full
#0  0x000000000052499f in InstrumentSoundShaping::processAudioBuffer (this=0x7fffeab2aca0, buffer=0x7fffeab2f5e0, frames=256, n=0x7fffeab218e8)
    at /home/zonkmachine/builds/lmms/lmms/src/core/InstrumentSoundShaping.cpp:206
        new_cut_val = 6025
        fcv = 25
        cutBuffer = 0x7fffddddd6c0
        resBuffer = 0x7fffddddd2b0
        old_filter_cut = 6025
        old_filter_res = 0
        frv = 0,790000021
        envTotalFrames = 32316
        envReleaseBegin = -370
#1  0x000000000062b3ae in InstrumentTrack::processAudioBuffer (this=0x7fffeab29720, buf=0x7fffeab2f5e0, frames=256, n=0x7fffeab218e8)
    at /home/zonkmachine/builds/lmms/lmms/src/tracks/InstrumentTrack.cpp:203
        offset = 0
        vol = 2,35106045e-38
        pan = 0 '\000'
        vv = {vol = {-1,07653206e+26, 4,59163468e-41}}
        DefaultVolumeRatio = 0,00999999978
#2  0x00007fffdc3b3704 in bitInvader::playNote (this=0x7fffeab2c5a0, _n=0x7fffeab218e8, _working_buffer=0x7fffeab2f5e0)
    at /home/zonkmachine/builds/lmms/lmms/plugins/bit_invader/bit_invader.cpp:301
        frames = 256
        offset = 0
        ps = 0x7fffeab2fde0
```

PhysSong:
@zonkmachine It looks like #3775.

zonkmachine:
Yes, that looks like it.

lukas-w:
@zonkmachine So I guess it's not related to this PR or #3783?

zonkmachine:
Yes, job done here.

PR #3873, opened 2017-10-10
Replace MemoryManager implementation with rpmalloc

lukas-w:
Fixes #3865, by integrating [rpmalloc](https://github.com/rampantpixels/rpmalloc).

Quick – probably not very meaningful – benchmark, using Skiessi-RandomProjectNumber14253:

|   | time user  |  time system | elapsed  |  cpu |
|:---|---:|---:|---:|---:|
| stable-1.2  | 23.26  |  1.33 |  8.09 | 303%  |
| malloc/free in MemoryManager | 21.30  | 1.30  | 6.57  | 344%  |
| rpmalloc [override](https://github.com/rampantpixels/rpmalloc/blob/master/rpmalloc/malloc.c), replaces all memory allocations  | 21.66  | 1.45  | 6.47  | 357%  |
| rpmalloc (this PR)  | 21.15  | 1.45  | 6.71  | 337%  |

PR #3887, opened 2017-10-16
Use weak loading of libjack

tresf:
WeakJack is a third party tool that abstracts the JACK Application Binary Interface for use with various jack versions.  It helps make LMMS a portable application on Linux.

* This used to be part of #3688.  It was separated to make tracking (blaming, cherry-picking, submoduling) easier.
* This is 100% 3rd-party code and should eventually be submoduled.


tresf:
Travis-CI passes for Linux and Windows.  I'm not waiting for Apple, this has been tested several times in the other PR.  Merging.

lukas-w:
Can we move the third-party code to a dedicated directory, such as `src/3rdparty`, similar to what I've done in #3873 or #3475? This will make it clear to everyone that it's not part of LMMS, and we can easily exclude it from tools like automatic code formatting.

Also, maybe we should add a file or a comment somewhere that documents the upstream source of the code.

tresf:
@lukas-w I agree fundamentally, although I see little benefit in doing refactoring in `stable-1.2` branch.  We could triage these refactoring items into a new milestone so that we don't forget them?  `ArashPartow/exprtk` is another candidate.

lukas-w:
Sure 👍

Issue #512, opened 2014-03-26
Windows runtime terminating in an unusual way


eagles051387:
every time i try to open the lmms its stop and tells me (thisapplication has requested the runtime to terminate it in and unusual way ) and than closes the program some one help me please i look on youtube and cant find a video for my problem only video on how to use it

BeaQueen:
I tried all the solutions shared on this post and LMMS (version 1.1.3) is still not running on my computer {Windows 10 64bits}.
[Intel i7-6700K 4.GHz, 16RAM GB, Sound card Approx APPPCI71 PCI 7.1]  
![lmms](https://cloud.githubusercontent.com/assets/16983449/13492440/511f7d20-e138-11e5-935a-833b809b06cf.png)

musikBear:
@BeaQueen This thread is for a general crash when LMMS is closed *down* fully normally with a click in the red-x, or the file context-menu.
If i understand you correctly, you cant run LMMS at all?
That is different, and does not belong *here*
''I dont have an emediate 'fix' for your issue. Others uses win10 successfully, but try
http://lmms.io/forum/viewforum.php?f=7

tresf:
> @BeaQueen This thread is for a general crash when LMMS is closed down fully normally with a click in the red-x, or the file context-menu.

No it's not, it's identical to her problem.

> That is different, and does not belong here

It's OK here.  Although it's a known issue, it's a quite severe one and @BeaQueen has at least posted to the topic versus creating a new duplicate.

@BeaQueen have you tried to reinstall your sound card drivers?  I've have a computer start working simply by uninstalling the sound card driver and getting a new driver from Windows.  In other cases the opposite has helped.

You may also want to move your `%USERPROFILE%/.lmmsrc.xml` file to see if you can force a fallback on to SDL.  Alternatively you can try to copy a config file from another computer with `DummyAudio` selected and see if that will help it load.

@musikBear is correct in the sense that you'll get more end-user help on the forums.  Most of the devs don't run Windows at all, and those that do, can't reproduce the problem, making this bug very difficult to debug from a development perspective, leaving the forums a better home for discussion.

musikBear:
Why dig in the DB to recreate my mistake. I realized i was on the wrong track emediately and removed the comment?
@BeaQueen  If you already has tried as you wrote
> I tried all the solutions shared on this post

Then the driver thing is also old news.
One way to get a working LMMS could perhaps be to try the 32x windows version, or perhaps this
[...]

PR #3889, opened 2017-10-16
Disable portaudio on Windows

Umcaruje:
#512 Has been happening for more and more users lately, we have people who can't even open the software on their machines. I provided them with a build with portaudio removed and their issues were gone.

Portaudio was always hard to set up on windows for me, and I find SDL works every time. If WASAPI support is required, I think we should focus on libsoundio, even though that would mean dropping support for windows XP. ASIO is also a planned feature for libsoundio: https://github.com/andrewrk/libsoundio/issues/23

Anyways, in conclusion, Portaudio causes more problems than benfits on windows, and I think it should be disabled.


Umcaruje:
Unless there are objections, I'll merge this in 24 hours.

lukas-w:
It's easy enough to revert in case there are any objections. 👍

CMakeLists.txt
```diff
@@ -85,12 +85,14 @@ IF(LMMS_BUILD_WIN32)
...
 	SET(WANT_JACK OFF)
 	SET(WANT_PULSEAUDIO OFF)
+	SET(WANT_PORTAUDIO OFF)
 	SET(WANT_SOUNDIO OFF)
 	SET(WANT_WINMM ON)
 	SET(LMMS_HAVE_WINMM TRUE)
 	SET(STATUS_ALSA "<not supported on this platform>")
 	SET(STATUS_JACK "<not supported on this platform>")
 	SET(STATUS_PULSEAUDIO "<not supported on this platform>")
+	SET(STATUS_PORTAUDIO "<disabled on this platform")
```
PhysSong: Missing `>` here.

lukas-w: [Fixed](https://github.com/LMMS/lmms/commit/20ea945cee6030cb5872ff27c095976c64eaf631)

Issue #3286, opened 2017-01-24
[suggestion / feature request] sf3 player


trebmuh:
The SF3 format has been invented by the lead developer of Musescore and is implemented in Musescore (http://musescore.org/) and Polyphone (http://polyphone-soundfonts.com/).

The SF3 format is the same than the SF2 one but the samples are in OGG and not in WAV. This in order to allow the soundfonts file weight to be heavily reduced.

It'd be great to have this support in LMMS (if that's not already the case and I missed it).

tresf:
@trebmuh thanks for the bug report.

This is really a `fluidsynth` task.  If you agree, please close this out in favor of upstream implementation.  https://sourceforge.net/p/fluidsynth/tickets/142/

trebmuh:
@tresf : thanks for the pointer; really appreciated. It sounds great.

Just to be 100% sure of my understanding, can you confirm/infirm the following: LMMS is using fluidsynth for the sf2 management. So, if/when the support for sf3 is added to fluidsynth, then LMMS will be able to deal with the sf3 soundfonts. Am I right here?

Going further, if I'm able to manage to build a local fluidsynth with the patch from sourceforge, would that mean that LMMS will be able to deal with sf3 straight, or would LMMS need to be rebuild as well to do so?

tresf:
> So, if/when the support for sf3 is added to fluidsynth, then LMMS will be able to deal with the sf3 soundfonts. Am I right here?

Yes.  We may have to add the `.sf3` file extension, but that should only take a few lines of code.

> Going further, if I'm able to manage to build a local fluidsynth with the patch from sourceforge, would that mean that LMMS will be able to deal with sf3 straight, or would LMMS need to be rebuild as well to do so?

If your changes are published LMMS could technically point to your source instead.  This isn't necessarily a quick task.    See this thread for a similar situation:  https://github.com/LMMS/lmms/pull/3224#issuecomment-274689192

Probably the best way to get this in is to get upstream to accept it and then it will naturally flow down stream.

trebmuh:
>> So, if/when the support for sf3 is added to fluidsynth, then LMMS will be able to deal with the sf3 soundfonts. Am I right here?

> Yes. We may have to add the .sf3 file extension, but that should only take a few lines of code.

That sounds great.

> If your changes are published LMMS could technically point to your source instead. This isn't necessarily a quick task. See this thread for a similar situation: #3224 (comment)
[...]

Issue #1807, opened 2015-03-02
Splitting -- Placeholder folders (and files) in user directory for Ladspa, VSTs, themes and templates


curlymorphic:
This issue is to keep track of progress in #1541 as it covers a lot or areas. I am planing on breaking it down into the following areas

* [x] **Settings Dialog**
 * [x] Add Scrollbar to path selection list
 * [x] Move STK raw waves,  Default SF2, to bottom of list
 * [x] This dialog needs instructions and possibly enhanced to allow multiple selections. It's not obvious that you can have more than one search path for some of these values, that should be spelled out IMO.  
 * [x] Artwork should be renamed "Themes" in my opinion, kept on paths page.

* [x] **Folders**
 * [x] Sample, GIG, SF2, folders created (#1541)
 * [x] Add folders to setting tab
 * [x] Connect up to make these folders the default when open this type of file.    

* [x] **Plugins**
 * [x] Add user directories,  
 * [x] Options added to setting dialog

  
  
I plan on keeping this initial post updated

curlymorphic:
If we are to have a Miscellaneous tab on the setting dialog, we are going to need an icon. I am unfamiliar with lmms artwork, do we already have an icon we can use, or what is required in making one. My gimp skill are not the best, is anyone able to help this this please?

![image](https://cloud.githubusercontent.com/assets/7412852/6408250/8e132d76-be48-11e4-8de6-d888c4315018.png)

curlymorphic:
ty for cleaning up my original post :)

Sti2nd:
I am not against a new tab, but I want to point out that you could possibly make the window scrollable if you think some settings belong to for example the path tab

curlymorphic:
> you could possibly make the window scrollable

Thats a nice idea. It would give us room to expand with ease, and have mimimal impact, on both the user experience, and current code base.

tresf:
> I am not against a new tab, but I want to point out that you could possibly make the window scrollable if you think some settings belong to for example the path tab

True... we could just put them at the bottom.

In regards to the artwork, if you make a new button, please make it as ugly as possible.  It makes people cringe and then they make a better one.

curlymorphic:
![image](https://cloud.githubusercontent.com/assets/7412852/6448954/5499d8a4-c115-11e4-9d4c-7d346f866336.png)
There is the new icon :)  until someone donates better 

if we need one. Lets see what everyone thinks?

tresf:
[...]

PR #3895, opened 2017-10-19
Enable sf3 extension

tresf:
Upstream fluidsynth has added ogg encoded soundfont support.  This won't be available until fluidsynth binaries are updated, so this should only be made available in the `master` branch for now.

Per #3286


Sawuare:
After merging this, I think we should rename everything named `SF2` to `SF` (in the GUI at least), for example `SF2 directory` to `SF directory`.

Spekular:
At that point it might be worth extending to Soundfont. SF by itself
doesn't say much to me.

trebmuh:
Nice. I just did a build from GIT + adding this patch, and using fluidsynth 1.1.8 .
Result is that I've been able to get sound from the "SF2 Player" when loading the FluidR3Mono_GM.sf3 file. Awesome and congratz @tresf !

Something I noticed is the name of the "SF2 Player" module which now could be misleading. Shouldn't be renamed to "SF Player" or "SoundFont Player" as suggested by @Spekular (which is a clever suggestion I'd say)?

tresf:
AFAIK, the `sf2` folder is a relatively new addition and hasn't been made part of a stable release.  This was carefully decided here: https://github.com/LMMS/lmms/issues/1807.  It's something we should be able to change if we can agree on something.  We can even talk about an upgrade path for intermediary (RC) users although this may be overkill.  I'd be a fan of `samples/soundfonts` personally.  It's what I've been using myself for years.

From a naming perspective, I'm not sure `sf2` is really that inaccurate or even bad.  It's like making a folder called "MP3s" and having some `M4A` files in there.

Slightly off-topic... I have a vision forward that would encompass all bank files (`sf2`, `sf3`, `sfz`, `sfArk`, `gig`).  This would strip any file format branding and replace it with a modularized, non-branded "instrument patch" instrument, which when moving towards a [single-window design](#1911) should be more future proof.  The plugin could be called `sf2` still, but the idea of a patch (I propose) should be agnostic to the user.  They'd just drag it in like any other sample. `</offtopic>`.

Mark-Agent003:
@tresf, (to encourage your slightly off-topic vision) I've thought through that same concept, and I really like that you've thought of this too. I agree on having patch banks and samples together, it makes sense. As for sf3 files in LMMS, thanks, it will be appreciated.

tresf:
So @LMMS/developers do we rename `samples/sf2` to `samples/soundfonts` to be nomanclature-friend with the `sf3` extension or do we leave it as-is?

If we change it, this decision should make it into `stable-1.2` to avoid future discrepancies.

👍 - Rename `sf2` folder to `soundfonts`
👎 - Leave `sf2` alone, keep it `sf2` for the foreseeable future.

Wallacoloo:
According to [Wikipedia](https://en.wikipedia.org/wiki/SoundFont), the SoundFont spec is still at version 2.04, and hasn't changed since 2005 - I think it's safe to consider it final. So what is `sf3`? Is it actually a widely recognized & supported format? If not, then I think the simpler thing would be to preserve the sf2 nomenclature.

But actually I vote for the change to `soundfonts` anyway because it provides more context to those unfamiliar with the `sf2` acronym.

Umcaruje:
I actually think all these files, such as `.sf2`, `.sf3`, `.pat` and `.gig` should go into a single folder called `multisamples`, this is a pretty common name among other DAWs, but this should perhaps happen when all the players get merged into a single instrument.

tresf:
Cherry-picked to stable via https://github.com/LMMS/lmms/commit/de98781df4d5b66d1e273167aad05915c42ebcf1.  Merging.

Note, although this closes #3286, official SF3 support won't arrive to LMMS until 1.3.0, or before then by running a build from master.

Issue #3896, opened 2017-10-20
Difficult to move and delete automation points


musikBear:
rc4 win 32
Irradiate/ random in occurrence
  
recreate: 
* Open an automation
* insert any curve with ~10 points

The problem seams to be most common, at high resolutions:
* set magn to 800
* set Q to 1/192

The points are tiny, and it is difficult to 'grab' one. Quite often a new point will appear instead. 
If the points were bigger, i believe this issue would be solved
Deleting one point often removes more than one

Redo works perfect! 👍

zonkmachine:
Something is going on there. I'll have a look.

zonkmachine:
> The points are tiny, and it is difficult to 'grab' one. Quite often a new point will appear instead.
> If the points were bigger, i believe this issue would be solved

It's not the size of the points but the quantization level versus the zoom level. If there is a control point within the distance for the quantization the point will snap to the same level as the cursor and then you can drag it. I'm touching up the code so it'll be much smoother.

musikBear:
@zonkmachine great 👍 -But *should* the points be that tiny? I would prefer the medium-big points, for easy mouse-grabbing. Same size in all magnifications. Imo no need to scale points

PR #3902, opened 2017-10-22
Fix Automation Point delete radius and size

zonkmachine:
Exchange defective algorithm for Automatin Point delete radius with a new function adapted from `AutomationEditor::drawLine()`,
Increase minimal Automation Point radius by 50%.

The sensitivity for grabbing/moving an Automation Point is still a bit off on the fines quantization levels.

Fixes https://github.com/LMMS/lmms/issues/3896

### DIFF
--- a/src/gui/editors/AutomationEditor.cpp
+++ b/src/gui/editors/AutomationEditor.cpp
@@ -1100,7 +1100,7 @@ inline void AutomationEditor::drawAutomationPoint( QPainter & p, timeMap::iterat
 {
 	int x = xCoordOfTick( it.key() );
 	int y = yCoordOfLevel( it.value() );
-	const int outerRadius = qBound( 2, ( m_ppt * AutomationPattern::quantization() ) / 576, 5 ); // man, getting this calculation right took forever
+	const int outerRadius = qBound( 3, ( m_ppt * AutomationPattern::quantization() ) / 576, 5 ); // man, getting this calculation right took forever
 	p.setPen( QPen( vertexColor().lighter( 200 ) ) );
 	p.setBrush( QBrush( vertexColor() ) );
 	p.drawEllipse( x - outerRadius, y - outerRadius, outerRadius * 2, outerRadius * 2 );
--- a/src/gui/editors/AutomationEditor.cpp
+++ b/src/gui/editors/AutomationEditor.cpp
@@ -589,6 +589,7 @@ void AutomationEditor::mousePressEvent( QMouseEvent* mouseEvent )
 							m_editMode == DRAW ) ||
 					m_editMode == ERASE )
 			{
+				m_drawLastTick = pos_ticks;
 				m_pattern->addJournalCheckPoint();
 				// erase single value
 				if( it != time_map.end() )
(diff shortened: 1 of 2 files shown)


zonkmachine:
I found a much better way by adapting `AutomationEditor::drawLine()`.

src/gui/editors/AutomationEditor.cpp
```diff
@@ -507,6 +507,7 @@ void AutomationEditor::mousePressEvent( QMouseEvent* mouseEvent )
 			// get tick in which the user clicked
 			int pos_ticks = x * MidiTime::ticksPerTact() / m_ppt +
 							m_currentPosition;
+			m_drawLastTick = pos_ticks;
```
zonkmachine: I just, cunningly, shot the line draw function to pieces with this line. 8P

zonkmachine: Fixed.

src/gui/editors/AutomationEditor.cpp
```diff
@@ -735,14 +771,13 @@ void AutomationEditor::mouseMoveEvent(QMouseEvent * mouseEvent )
...
-			int resolution = 1 + zoom * zoom;
-			for( int i = -resolution; i < resolution; ++i )
+			// removing automation point
+			if( pos_ticks < 0 )
 			{
-				m_pattern->removeValue( MidiTime( pos_ticks + i ) );
+				pos_ticks = 0;
 			}
+			removePoints( m_drawLastTick, pos_ticks );
+			Engine::getSong()->setModified();
```
PhysSong: It will mark the project as modified if no points are removed. It should be moved into `removePoints()`.

zonkmachine: :+1:

zonkmachine: I remove points without checking if there are any there. So if I add ```bool pointsRemoved = false;``` and lets the delete loop set it to true, it is still going to set the project to changed as soon as the mouse moves just a bit without really removing anything.

zonkmachine: The behaviour right now is, deleting points with right click sets the project to changed but deleting by dragging over the point doesn't.

zonkmachine: @PhysSong I don't know quite how to move ```Engine::getSong()->setModified();``` to `removePoints()`. This is already an improvement from before when setModified() wasn't called at all. I suggest merging this 'as is', or I can have another look at it if someone can point me in the right direction but for now I think it's low priority.

PhysSong: `Engine::getSong()->setModified();` is an UX problem. So I think it's out of scope a little bit. Since the point of this PR is on GUI part, merging as-is looks fine. Dropping that will be fine, too.
I can create a PR which fixes the UX issue later. I think it will be the best way. :+1:

src/gui/editors/AutomationEditor.cpp
```diff
@@ -680,6 +681,39 @@ void AutomationEditor::mouseReleaseEvent(QMouseEvent * mouseEvent )
...
+	int deltax = qAbs( x1 - x0 );
+	int x = x0;
+	int xstep;
+
+	if( deltax < AutomationPattern::quantization() )
+	{
+		return;
+	}
+
+	if( x0 < x1)
```
PhysSong: This looks somewhat odd. Space after `(`, but no space before `)`.

zonkmachine: Fixed.

src/gui/editors/AutomationEditor.cpp
zonkmachine: All right. I'll merge this tomorrow.

PR #3904, opened 2017-10-23
Fix preset preview issues with peak controller(s)

### LINKED ISSUE #3897: Presets with peak controller crash LMMS after previewing

PhysSong:
Fixes #3897.
This PR ensures peak controller effects from preview track don't instantiate a peak controller and replaces itself to a dummy effect by setting `m_okay` to `false`.
Here's a preset for test: [PeakController.xpf.zip](https://github.com/LMMS/lmms/files/1407220/PeakController.xpf.zip)

### DIFF
--- a/plugins/peak_controller_effect/peak_controller_effect.cpp
+++ b/plugins/peak_controller_effect/peak_controller_effect.cpp
@@ -26,6 +26,7 @@
 
 #include "Controller.h"
 #include "Song.h"
+#include "PresetPreviewPlayHandle.h"
 #include "PeakController.h"
 #include "peak_controller_effect.h"
 #include "lmms_math.h"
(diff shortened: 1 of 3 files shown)


PhysSong:
The difference between this and @gi0e5b06's approach is: this don't create peak controller while @gi0e5b06's fix create it and not add to the song. I'll compare two ways and find better way...

gi0e5b06:
@PhysSong If you don't create the peak controller(s), the preset may sound totally different. Which just brings this other issue #3906 . OTOH, the consequences of creating these temporary controllers are still unknown but I didn't notice anything strange so far (after fixing some double-ckick bugs in the file browser).

PhysSong:
@gi0e5b06 That is a long-standing issue. However, your suggestion doesn't look harmful dor me. I 'll take a look at this more. :smile:

PhysSong:
Partially switched to @gi0e5b06's approach. Waiting for tests...

zonkmachine:
Tests fine. Issue fixed with this PR.

PR #3912, opened 2017-10-24
Fix monstro's osc 3 vol knob values. 

### LINKED ISSUE #3911: Monstro's osc 3 vol knob always uses percentages.
Monstro's osc 3 vol knob with the "Display volume as dBFS" setting enabled:
capture20171024164622676
image

It uses percentages instead of levels in dBFS.

Sawuare:
Fixes #3911.


zonkmachine:
I'll merge this after Travis is done.

Issue #3763, opened 2017-08-11
Vertical Zoom in Song Editor


Spekular:
With the recent improvement to vertically larger patterns in the song editor (PR 3699), a vertical zoom makes more sense than before. This would allow every track to be scaled vertically by the same amount, and could be more convenient than shift+dragging even if you only want to focus on one track.

Right now the following shortcuts use the mouse wheel:

| Modifier | Modifier | Mousewheel = | Direction | Operation |
| ---: | ---: | :---: | :--- | :--- |
|  |  | Mousewheel = | Vertical | Scroll |
|  | Shift + | Mousewheel = |  Horizontal | Scroll |
|  | Alt + | Mousewheel = | Horizontal | Scroll |
|  | Ctrl + | Mousewheel = | Horizontal | Zoom |

Since there are two horizontal scroll shortcuts, ether Shift or Alt could in theory be used for vertical zoom. However, I think a modified Ctrl would make more sense. This would lead to something like this (excluding alt for brevity):

| Modifier | Modifier | Mousewheel = | Direction | Operation |
| ---: | ---: | :---: |  :--- | :--- |
|  |  | Mousewheel = | Vertical | Scroll |
|  | Shift + | Mousewheel = | Horizontal | Scroll |
|  | Ctrl + | Mousewheel = | Horizontal | Zoom |
| Shift + | Ctrl + | Mousewheel = | Vertical | Zoom |

However, this would mean that an unmodified scroll is vertical, and an unmodified zoom is horizontal. Perhaps these should be unified, so an unmodified scroll and zoom operate on the same axis. I would then propose changing a regular unmodified mousewheel scroll to move horizontally, leading to:

| Modifier | Modifier | Mousewheel = | Direction | Operation |
| ---: | ---: | :---: | :--- | :--- |
|  |  | Mousewheel = | Horizontal | Scroll |
|  | Shift + | Mousewheel = | Vertical | Scroll |
|  | Ctrl + | Mousewheel = | Horizontal |  Zoom |
| Ctrl + | Shift + | Mousewheel = | Vertical |  Zoom |

This would make all mousewheel events horizontal by default. Shift becomes the "vertical" modifier, and Ctrl remains the "zoom" modifier.

Spekular:
Here's a rough mockup, although I'm not sure if one is necessary. I haven't actually built a version with the drawing changes so this isn't even properly representative.

[...]

PR #3917, opened 2017-10-26
[WIP] Vertical Zoom

Spekular:
I decided it was about time I took #3763 into my own hands. My very, *very*, incompetent hands.

Right now, shift scroll:

1. Changes a zoom-selector to the right of the horizontal scroll selector. This works fine except that it also changes the song editor to select mode while shift is held, and the selectors need padding and labels/icons.
2. Triggers a big mess that tries to set all tracks to be 300 units tall. This needs to actually set height properly based on the track's base height and the song editor's zoom level.

The big mess goes through a bunch of classes that it maybe shouldn't because I can't wrap my brain around how everything is connected. The tracks don't actually visually change size until you save and reload the project, and BB tracks are unaffected because why not.

<sup>send help</sup>

### DIFF
--- a/plugins/LadspaEffect/caps/basics.h
+++ b/plugins/LadspaEffect/caps/basics.h
@@ -184,7 +184,7 @@ class Plugin {
 		inline sample_t getport_unclamped (int i)
 			{
 				sample_t v = *ports[i];
-				return (isinf (v) || isnan(v)) ? 0 : v;
+				return (__builtin_isinf_sign(v) || __builtin_isnan(v)) ? 0 : v;
 			}
 
 		/* get port value and clamp to port range */
--- a/src/gui/editors/SongEditor.cpp
+++ b/src/gui/editors/SongEditor.cpp
@@ -50,7 +50,6 @@
 #include "TimeDisplayWidget.h"
 #include "AudioDevice.h"
 #include "PianoRoll.h"
-//Spekular was here
 #include "Track.h"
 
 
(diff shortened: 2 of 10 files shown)


qnebra:
At this moment, vertical zooming not works.

Spekular:
@qnebra yeah, this isn't actually working yet. I wouldn't recommend testing it with any actual projects either, since it sets track heights absurdly high.

qnebra:
Maybe limit vertical zoom to max 400%?

plugins/LadspaEffect/caps/basics.h
```diff
@@ -184,7 +184,7 @@ class Plugin {
 		inline sample_t getport_unclamped (int i)
 			{
 				sample_t v = *ports[i];
-				return (isinf (v) || isnan(v)) ? 0 : v;
+				return (__builtin_isinf_sign(v) || __builtin_isnan(v)) ? 0 : v;
```
Spekular: Unrelated change required for me to compile.

src/gui/editors/SongEditor.cpp
```diff
@@ -50,6 +50,7 @@
 #include "TimeDisplayWidget.h"
 #include "AudioDevice.h"
 #include "PianoRoll.h"
+#include "Track.h"
```
Spekular: Don't actually know if this is necessary

Spekular:
Closing. Issues:
- [ ] Relies on changing the actual size of the tracks rather than changing how they're drawn
- [ ] Doesn't affect BB tracks
- [ ] Doesn't show up until the project is saved and reloaded
- [ ] Sets all tracks to a fixed size instead of actually using the zoom factor

Issue #3914, opened 2017-10-26
Cannot adjust individual selected note velocities in piano roll


gandalf3:
According to https://github.com/LMMS/lmms/pull/1804, I should be able to alt+left click and drag to change the individual velocities of the selected notes. However, nothing happens when I do this. Other modifiers (or no modifiers at all) cause all of the selected notes to be set to the same velocity.

LMMS 1.2.0-rc3.56
Archlinux

musikBear:
i think you point the wrong spot. Its on the velosity columns, not the note
![velositysetting](https://user-images.githubusercontent.com/6662066/32054321-58075e9a-ba5e-11e7-96c6-d22134ee74c9.JPG)
If you click once the velocity change
If you double-click  the input-box for velocity change (picture) appear.
This is a help request, so you should use forum, *unless* none of the abowe works in linux, then its a bug, and then it need to be tested in rc4

PhysSong:
> alt+left click and drag to change the individual velocities of the selected notes

You can do that with left click + drag or Alt + scroll. As @musikBear pointed, it should be done on the velocity section.

gandalf3:
Tested with rc4, same behavior. I am dragging on the velocity columns.
<kbd>Alt</kbd>+scroll works, but it also works on unselected notes too:

![lmms_pianoroll_velocity_demo](https://user-images.githubusercontent.com/11167504/32087840-8bd5d1c2-ba93-11e7-84de-57360e9e08bf.gif)

musikBear:
>  adjustment of the selected notes individually by dragging along
@gandalf3 
No that you cant do, but you can *un-select* one specific note in a selection-group with sh+left-click, and then you can adjust that *unselected note* and re-select it to the group with yet another sh+left-click.
-Again, its a help-request:

gandalf3:
@musikBear That sounds horrifically tedious, and it also doesn't work for me; clicking the velocity bar while notes are selected adjusts all selected notes.

Is there now no way to quickly adjust the velocity of some notes while ignoring others? If so, I feel like this is a major regression (this used to be possible). If regressions not considered bugs, is there a separate place for them? (feature requests?)

musikBear:
> That sounds horrifically tedious

Ye.. it is a bit convulated. Agreed..
But i guess its a compromise between being able to select a group, and set their velosity at once, and also being able to set one note.
What is 'missing' is a *select-in-selection*, but honestly; Have you ever seen any program have that?!
[...]

Issue #1810, opened 2015-03-02
Find an alternative for Alt keybindings for X/Linux users


badosu:

Sti2nd:
So the problem is that you cannot first press ALT and then use the mouse in Linux.

My thoughts: It would be really great to have just a single key because then it can be combined with fast resizing of notes in Piano Roll (and possibly other functionality ofc). Now in LMMS you can click and hold and while holding hit SHIFT to do fast resize, add the ALT key and you can do fine tuning quick too.

mikobuntu:
I like the idea of just a single key i.e ALT , my preference would be to use the window management overrides of QT for Linux Xwindows. 
ALT has become or always has been a well known modifier that users have become accustomed to .

Umcaruje:
>I like the idea of just a single key i.e keeping ALT intact as it is.... , my preference would be to use the window management overrides of QT for Linux Xwindows. 

I agree with mikobuntu, this would be an ideal solution to this problem (If its currently possible).

tresf:
> preference would be to use the window management overrides

I'm not sure this can be controlled on the application level.  This is very wishful thinking.

From what I'm reading, you're dealing with global desktop settings here, not application-specific overrides.

I use ALT+Click quite a bit on Linux so I wouldn't be a fan of changing this default for the desktop personally.

mikobuntu:
"I use ALT+Click quite a bit on Linux"

Yeah this will be the same for a lot of users and most likely that a user will be multi tasking, not just using LMMS. I think what @badosu had mentioned in the previous thread leading to this FR, was having a temp override, i.e only during the actual mouse/key gestures and not as I had initially proposed a global override.

tresf:
>  was having a temp override

How though?  That is my question... is this even possible?

Sti2nd:
If it is possible... Idk. I guess mikobuntu mean that in the Piano Roll when mouse over the green volume bars LMMS will override, but note elsewhere.

mikobuntu:
@Sti2nd  yes that is what I mean and in any other situation where ALT key bindings are used within LMMS.  @tres " is this even possible? " IDK , hopefully :)

mikobuntu:
http://qt-project.org/faq/answer/how_can_i_catch_altf4_in_my_qt_application

could something like this example work?

musikBear:
[...]

Issue #1475, opened 2014-12-20
User definable shortcuts


Sti2nd:
Proposal: A new setting tab where you can redefine the keys used for each shortcut. You could allow MIDI controllers here so that stop and play could be started with it, for example. https://github.com/LMMS/lmms/issues/1071

Why? Cause we are different, we have different computers and different equipment.

Umcaruje:
@Sti2nd can you please extend this request with some ideas how this could be implemented, and a mockup?

softrabbit:
Maybe something here helps for the keyboard part? 
http://doc.qt.io/qt-5.4/qshortcut.html
http://stackoverflow.com/questions/13487352/how-to-create-a-hotkey-field-in-qt

Sti2nd:
I think how Musescore does it was what I had in mind. Whatever is easy to make, that link ^ is ok as well, maybe even the exact same as Musescore use Qt as well?

![image](https://cloud.githubusercontent.com/assets/6373430/8580121/0d6a773c-25b9-11e5-865d-a8208c602566.png)

Hussam-Eddin-Alhomsi:
I like how Audacity handles keyboard shortcuts.
* Easy viewing and searching system.
* You choose an action, and you can either set a shortcut for it or remove the current one.
To set a shortcut for your action:
(1) You click on white space and perform you shortcut. The program recognizes what key you pressed and displays them for you.
(2) If you set a shortcut for an action and another action already has it, a message box will show, warning you.
* You can import/export shortcut from/to XML files.
* You can return the shortcuts to their defaults.

Screenshots:
![image](https://user-images.githubusercontent.com/25503477/28240026-fe27885a-6981-11e7-8c82-042f53d54ecf.png)
![image](https://user-images.githubusercontent.com/25503477/28240028-05de9ac0-6982-11e7-8aea-73ffb531eff1.png)

PR #3923, opened 2017-10-28
Allow individual velocity/pan changes with alt

gandalf3:
This commit changes the behavior introduced in 6e3d4f431 to allow using
alt to change the velocity/panning of multiple *selected* notes by dragging.

Screencapture showing piano roll velocity/pan click+drag behavior with and without Alt held)

I believe this was the originally intended behavior, but it never worked for me.

Fixes https://github.com/LMMS/lmms/issues/3914

### DIFF
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -2118,9 +2118,14 @@ void PianoRoll::mouseMoveEvent( QMouseEvent * me )
 				bool isUnderPosition = n->withinRange( ticks_start, ticks_end );
 				// Play note under the cursor
 				if ( isUnderPosition ) { testPlayNote( n ); }
-				// If note is the one under the cursor or is selected when alt is
-				// not pressed
-				if ( ( isUnderPosition && !isSelection() ) || ( n->selected() && !altPressed ) )
+				// If note is:
+				// Under the cursor, when there is no selection
+				// Selected, and alt is not pressed
+				// Under the cursor, selected, and alt is pressed
+				if ( ( isUnderPosition && !isSelection() ) ||
+					  ( n->selected() && !altPressed ) ||
+					  ( isUnderPosition && n->selected() && altPressed )
+					)
 				{
 					if( m_noteEditMode == NoteEditVolume )
 					{


tresf:
@gandalf3 thanks for the PR!  This reads like a bug and should be targeted against `stable-1.2` so that it makes it into the stable release.

src/gui/editors/PianoRoll.cpp
```diff
@@ -2116,9 +2116,14 @@ void PianoRoll::mouseMoveEvent( QMouseEvent * me )
...
 				// Play note under the cursor
 				if ( isUnderPosition ) { testPlayNote( n ); }
-				// If note is the one under the cursor or is selected when alt is
-				// not pressed
-				if ( ( isUnderPosition && !isSelection() ) || ( n->selected() && !altPressed ) )
+				// If note is:
+				// Under the cursor, when there is no selection
+				// Selected, and alt is not pressed
+				// Under the cursor, selected, and alt is pressed
+				if ( ( isUnderPosition && !isSelection() ) || \
```
tresf: Cosmetic but the [`\` before the newline](https://stackoverflow.com/a/14570370/3196753) is not needed since this line is not part of a preprocessor statement.  It's also inconsistent with the formatting of this file.

gandalf3:
@tresf Thanks, I've removed the `\`s and switched to stable-1.2

src/gui/editors/PianoRoll.cpp
```diff
@@ -2118,9 +2118,14 @@ void PianoRoll::mouseMoveEvent( QMouseEvent * me )
...
-				// not pressed
-				if ( ( isUnderPosition && !isSelection() ) || ( n->selected() && !altPressed ) )
+				// If note is:
+				// Under the cursor, when there is no selection
+				// Selected, and alt is not pressed
+				// Under the cursor, selected, and alt is pressed
+				if ( ( isUnderPosition && !isSelection() ) ||
+					  ( n->selected() && !altPressed ) ||
+					  ( isUnderPosition && n->selected() && altPressed )
+					)
```
zonkmachine: On Linux `<Alt>` + `<Left Mouse Button>` by default moves the whole application window. I think this is true on all desktops. This is nothing you've missed here but I had to change
`bool altPressed = me->modifiers() & Qt::AltModifier;` to 
`bool altPressed = me->modifiers() & Qt::ControlModifier;`
in order to test this. It works as intended on Linux with this change applied. My suggestion is to simply change `<Alt>` for `<Ctrl>` . Two issues fixed.

gandalf3:
@zonkmachine Isn't <kbd>Ctrl</kbd> drag used for selection? When <kbd>Ctrl</kbd>+dragging here I get a box selection.

I use <kbd>Super</kbd> in place of <kbd>Alt</kbd> for window manipulations, as many other programs (e.g. Blender, Inkscape) use <kbd>Alt</kbd> + drag for stuff. I don't see a problem with programs using <kbd>Alt</kbd>. 

In fact, \<rant>  I've never met a single program which binds anything to <kbd>Super</kbd> by default, which seems to be conventionally reserved for window managers. Why distros bind <kbd>Alt</kbd> to the window manager by default (instead of <kbd>Super</kbd>), I do not understand. \</rant>

 Regardless, that's a complaint for distribution maintainers. In my opinion, the solution is to choose shortcuts which make sense without worrying about <kbd>Alt</kbd> and permit the user to rebind keys as they see fit. Obviously that's a much bigger task, though I see it already exists: https://github.com/LMMS/lmms/issues/1475

Anyway, I just wanted to put my two cents in. If I'm mistaken and <kbd>Ctrl</kbd> doesn't conflict, I'll happily update the PR.

Umcaruje:
I agree with @gandalf3, <kbd>Alt</kbd> seems to be the right modifier for this. my distro also doesn't have any shortcuts with <kbd>Alt</kbd>, they're all with <kdb>Super</kbd>.

zonkmachine:
It looks like the Alt + Move Window is a Gnome thing and one that many ask how to disable.

> Isn't Ctrl drag used for selection?

It indeed is.

zonkmachine:
Merge?

tresf:
> Merge?

Yes.  I can confirm that on Ubuntu 14.04 (and likely most other Linux desktops), the ALT key will drag the window, but a valid workaround is to simply add the SHIFT modifier and this feature works as expected.  So, hold SHIFT + ALT, then keep holding and click.

untitled

Side note... there's a bit of order-of-operation issues with modifiers and mouse clicks, but those are existing throughout most of LMMS.  i.e. If you click first and then add the modifier afterward, the behavior is different.  Fixing these other UX issues should be discussed separately but wanted to mention this incase someone testing this gets hung up. :)

zonkmachine:
There are supposedly ways to turn this feature off but so far it hasn't worked for me. It's an **upstream** anyway. I think this should be merged.

zonkmachine:
`<Alt>+<Left click>` issue (drag window) on some Linux versions discussed here: https://github.com/LMMS/lmms/issues/1810

PR #3929, opened 2017-11-01
Fix wrong value interpolation

### LINKED ISSUE #3859: Instrument 'on' glitch
On instrument tracks. If the volume is really low, as is often the case with some of the louder pluggins (3osc, bitinvader...) , there is shot noise the first time the track is switched on. When you press play the first time after loading a project or if the track was saved in a muted state, when you unmute it.

Steps to reproduce.
Default project, turn volume all the way down.
Enter one note and save the project.
Reload the project and press play.
Pop!
Press  stop and then play again... Silence!
Mute the instrument, save and reload.
Play and then unmute. Pop!

This is a regression on master after 1.1 and was possibly introduced in 23433a70b577bc67f2836c0ae98a4dc48ae40f93 but that commit has turned out as a possible false positive when bisecting before (https://github.com/LMMS/lmms/issues/3685).
@PhysSong

PhysSong:
Fixes #3859.

### DIFF
--- a/src/core/AutomatableModel.cpp
+++ b/src/core/AutomatableModel.cpp
@@ -42,8 +42,6 @@ AutomatableModel::AutomatableModel( DataType type,
 	Model( parent, displayName, defaultConstructed ),
 	m_dataType( type ),
 	m_scaleType( Linear ),
-	m_value( val ),
-	m_initValue( val ),
 	m_minValue( min ),
 	m_maxValue( max ),
 	m_step( step ),
@@ -59,6 +57,7 @@ AutomatableModel::AutomatableModel( DataType type,
 	m_hasSampleExactData( false )
 
 {
+	m_value = fittedValue( val );
 	setInitValue( val );
 }
 
@@ -523,14 +522,8 @@ float AutomatableModel::controllerValue( int frameOffset ) const
 
 ValueBuffer * AutomatableModel::valueBuffer()
 {
-	// if we've already calculated the valuebuffer this period, return the cached buffer
-	if( m_lastUpdatedPeriod == s_periodCounter )
-	{
-		return m_hasSampleExactData
-			? &m_valueBuffer
-			: NULL;
-	}
 	QMutexLocker m( &m_valueBufferMutex );
+	// if we've already calculated the valuebuffer this period, return the cached buffer
 	if( m_lastUpdatedPeriod == s_periodCounter )
 	{
 		return m_hasSampleExactData
@@ -626,6 +619,7 @@ void AutomatableModel::setInitValue( const float value )
 	m_initValue = fittedValue( value );
 	bool journalling = testAndSetJournalling( false );
 	setValue( value );
+	m_oldValue = m_value;
 	setJournalling( journalling );
 	emit initValueChanged( value );
 }


zonkmachine:
A first test. I'm not quite sure if this test tells anything but I cherry-picked 600e361 into master and tested it with the fpe debug option. The project I tested it with had a calf limiter just as the one in #3685/#3859. We had two separate fixes that will allow the project to play as normally, patching audio_fx.cpp or setting m_valueOld to '0' for new value buffers, and now this third proper fix for the value buffer. They all work and makes the failing projects pass but when you test for floating point errors, they will all come out crashing. The only difference is that with this latest effort it will play for two bars, but they will all crash with a similar backtrace.


master with 600e361
Crash after 2 bars
test project:  [fpeonstart.mmp.zip](https://github.com/LMMS/lmms/files/1436583/fpeonstart.mmp.zip)

```
0x00007fffd96d433e in LadspaEffect::processAudioBuffer (this=0x10000f0020, _buf=0x1000011df0, _frames=256)
    at /home/zonkmachine/builds/lmms/lmms/plugins/LadspaEffect/LadspaEffect.cpp:245
245							out_sum += _buf[frame][channel] * _buf[frame][channel];
(gdb) bt full
#0  0x00007fffd96d433e in LadspaEffect::processAudioBuffer (this=0x10000f0020, _buf=0x1000011df0, _frames=256)
    at /home/zonkmachine/builds/lmms/lmms/plugins/LadspaEffect/LadspaEffect.cpp:245
        frame = 114
        pp = 0x19f9780
        port = 2
        proc = 0 '\000'
        frames = 256
        o_buf = 0x0
        sBuf = 0x7fffb0d83280
        channel = 0 '\000'
        w = 0,469999999
        is_running = 127
        out_sum = 0,00017636669005594019
        d = 0,529999971
#1  0x00000000005178d9 in EffectChain::processAudioBuffer (this=0x19ba3a0, _buf=0x1000011df0, _frames=256, hasInputNoise=true)
    at /home/zonkmachine/builds/lmms/lmms/src/core/EffectChain.cpp:216
        it = 0x19c4550
        exporting = false
        moreEffects = false
#2  0x0000000000582411 in AudioPort::processEffects (this=0x10000d0398) at /home/zonkmachine/builds/lmms/lmms/src/core/audio/AudioPort.cpp:98
        more = false
#3  0x0000000000582dd3 in AudioPort::doProcessing (this=0x10000d0398) at /home/zonkmachine/builds/lmms/lmms/src/core/audio/AudioPort.cpp:221
        fpp = 256
        me = false
```

PhysSong:
Yes. I had the same issue before. That's an underflow, which seems harmless there.

PhysSong:
You may remove the [underflow check](https://github.com/LMMS/lmms/blob/c765249723ff6e460f6d45eee6f05bd0fbeec16e/src/core/main.cpp#L236) and test it again. Underflow can cause problems, but it isn't harmful for some situations such as level checking. In LMMS, inf and NaN are more important than denormals.

zonkmachine:
That worked. Interesting, what shows that there are denormals involved here?

PhysSong:
> Interesting, what shows that there are denormals involved here?

I inserted `printf` to check the value of `_buf[frame][channel]`. It was small enough to generate denormals when squared.

zonkmachine:
> In LMMS, inf and NaN are more important than denormals.

Maybe we should have the underflow check disabled by default for now then. The debugger is a bit too nervous right now.

PhysSong:
> The debugger is a bit too nervous right now.

Totally agree. In my case, LMMS fails to start with FPE debugging for some reason.

zonkmachine:
Also, we do have some tweak to keep denormals under control.
[disable_denormals();](https://github.com/LMMS/lmms/blob/c765249723ff6e460f6d45eee6f05bd0fbeec16e/src/core/main.cpp#L249) as included in `denormals.h`
As I understand it numbers of this size are simply substituted with '0'.


> LMMS fails to start with FPE debugging for some reason.

`gdb gdb` ?

Issue #296, opened 2014-02-11
Better handling of 3rd party dependencies


lukas-w:
_Taken and adapted from my original comment in #287_
_See another discussion of this subject in #96_

Allow me to say, that the way external dependencies are handled and embedded in LMMS's source is just not a practicable way, and all in all a mess.

In my opinion, one should avoid including source from other projects as far as it's possible. It adds unnecessary compile time for libraries that may be shipped by the linux distro anyway, makes it very difficult to maintain the source, to update the library from upstream, and to push own commits to upstream. Especially the last three points can cause a lot of work.

Basically, it's a matter of principle: Imagine every project out there shipping it's dependencies in the source tree, every project using and installing a different version of the needed libraries. This would lead to a lot of versions of the same library being installed next to each other at the same time which is a huge waste of disk space. It also contradicts the principle and intentions of shared libraries and package management.

Starting my argument at the other end: We can't include every single library we need, for instance we're never going to ship Wine or Qt or Jack or whatever because it's too huge and just not our job as that's what shared libraries and package managers are for.
There are good reasons for the fact that including external sources is no common practice and is something that is generally discouraged to do.

Even though I would head to removing as many third party sources from the tree as possible, I (partly) understand Toby's reasons for including them.
So if we're going to continue shipping external sources (the question about whether to do this in the first place shall not be the point of my message now), we should at least find a better way to do this.

I therefore suggest:
- **Have a special directory for dependencies**  
  Move all 3rd party libraries to one single directory in the source root, let's name it `extern`. The root directory would then look something like this:
  
  ```
  lmms/
  ├── cmake/
  ├── data/
  ├── extern/
  |      ├── calf
  |      ├── caps
  |      ├── …
  |      └── ZynAddSubFx/
  ├── include/
  ├── plugins/
  ├── src/
  ├── CMakeLists.txt
  ├── …
  └── README
  ```
- **Use git subtree**
  
  Use `git subtree` to manage the libraries. Basically it works by still having the source code in the tree, but knowing about where it comes from. This way pulling or pushing updates from or to upstream is made possible. 
  Note that it causes no extra efforts for users who for instance just want to clone the git repository and compile lmms (in contrast to git submodule, which requires users to explicitly run a command for downloading the directores).
  Subtree has been introduced in git-1.7.11. _Update:_ I just noticed this feature may be not included or not enabled in some distributions by default. It is not needed as long as you don't want to push changes in the subproject though.
- **Use CMake's ExternalProject module**
  
[...]

Issue #3876, opened 2017-10-11
Log file in LMMS


NETMANSKY:
LMMS lacks log file to trap bugs. It will help devs to fix troubles.

tresf:
Some context https://github.com/LMMS/lmms/issues/3561#issuecomment-302405321 as to why this is so important, at least on Windows.

gi0e5b06:
Logs make every thing slow.Logs never help with bugs.
Use qWarning/qDebug when coding/testing.
Use qFatal for runtime checks (during the development phase).
Keep those statements as comments once your code is stable.

tresf:
> Of course, stderr should be visible on every OS. If it is not on Windows, redirect the stream to a file.

No, it's simply not there at all.  `mconsole` is needed.  We may need to create an `lmms-console.exe` for debugging purposes.  It's a shitty limitation of the OS unfortunately.

tresf:
> redirect the stream to a file.

For the record though, Wine has access to it, so that's what I've used historically, but it's not the same environment, so it's a crapshoot.

gi0e5b06:
@tresf Check this:
http://doc.qt.io/qt-5/qtglobal.html#qInstallMessageHandler

tresf:
> @tresf Check this: http://doc.qt.io/qt-5/qtglobal.html#qInstallMessageHandler

Context please?

What I'm stating is that the console is gone in a GUI app for Windows.  You need to compile it as a console app if you want output and that console will appear until closed.  It's a limitation of the OS.  What are you suggesting?

tresf:
@gi0e5b06 Check this:
https://stackoverflow.com/a/3370017/3196753

gi0e5b06:
Good enough for Windows.

```
#ifdef LMMS_BUILD_WIN32
	stderr = ::fopen("lmms-err.txt","a");
	stdout = ::fopen("lmms-out.txt","a");
#endif

```

tresf:
So logging is only a performance problem when it affects the system you use?

gi0e5b06:
Writing to a file is faster than writing to the console.
Not really needed for Linux because every user knows 2>&1 1>/dev/null

tresf:
> Logs make every thing slow.Logs never help with bugs.
> Writing to a file is faster than writing to the console.

[...]

PR #3931, opened 2017-11-02
Switch swh to submodule

tresf:
Moves the "swh" LADSPA plugins to pull from upstream `swh/ladspa` by building the source from a git submodule.  This requires a few new build tools over what we use today, namely `perl` and `perl-xml`.

**Before**

```
plugins/LadspaEffect/swh/acconfig.h
plugins/LadspaEffect/swh/alias_1407.c
plugins/LadspaEffect/swh/allpass_1895.c
[... 109 total files ...]
```

**After**
```
plugins/LadspaEffect/swh/CMakeLists.txt
plugins/LadspaEffect/swh/ladspa@ 5edf95c -> Points to http://github.com/swh/ladspa
[...]
```

**New Dependencies** 
```
perl libxml2-utils libxml-perl liblist-moreutils-perl
```
Once merged, the dependencies must be updated on the wiki.  Supercedes #3489, related #296


gi0e5b06:
swh-lv2
swh-plugins (ladspa)

SWH is a package on most GNU/Linux distros. No need to make it a submodule, no need to have it as part of LMMS.

tresf:
> SWH is a package on most GNU/Linux distros. No need to make it a submodule, no need to have it as part of LMMS.

Thanks.  We know.  But this is the exception and not the rule and the systems you mention have packagers that already take advantage of this.

So in summary, the statement is not true for any use-case we care about... OUR installers.

Our installers include: Windows, MacOS & Linux AppImages (our ONLY official distribution methods).

Bundled installers need this and we will continue to bundle in this fashion until a better solution is provided

gi0e5b06:
Windows: https://www.fosshub.com/Audacity.html/LADSPA_plugins-win-0.4.15.exe
MacOS: https://www.fosshub.com/Audacity.html/swh-plugins-mac-0.4.15.zip
AppImage: no one is using that (and anyway there is the package)

It would be better to support DSSI, LV2, VST3 and Nyquist plugins instead.
And Audio Units on MacOS, and native VST on both MacOS and Linux.

tresf:
> It would be better to support DSSI, LV2, VST3 and Nyquist plugins instead.
And Audio Units on MacOS, and native VST on both MacOS and Linux.

You're going off topic.

> MacOS: https://www.fosshub.com/Audacity.html/swh-plugins-mac-0.4.15.zip

I'm not sure the point.  The fact that some unrelated 3rd party provides binaries doesn't negate the value this PR has.  We may eventually decouple our packaging as well, but that's out of scope.  This PR explains its purpose and it's something that's been discussed for several months.  Perhaps you need to jump on Discord? https://lmms.io/chat

tresf:
> AppImage: no one is using that (and anyway there is the package)

Again, off-topic.  If you have issues with AppImages or the popularity thereof, please place them in the appropriate channels (#devtalk on Discord or any of the AppImage issues on this tracker).

On a side note, you've been asked several times to improve the quality of your contributions on GitHub. 
 This is your first warning.  After three, you will be removed from our tracker.  Let's keep comments on-topic and progressive, please.  If you have a problem with this, we can discuss privately or publicly on Discord.

lukas-w:
> This is your first warning.

I feel like this already is the third warning: (1) https://github.com/LMMS/lmms/pull/3422#issuecomment-328307219  (2) https://github.com/LMMS/lmms/issues/3876#issuecomment-337651881

tresf:
This is successfully compiling on all platforms.  I'd like to point a few things out... 

1. I put a submodule guard in [here](https://github.com/LMMS/lmms/pull/3931/files#diff-3571db6c539aaf0a3404e117e5bdcaefR15) to avoid any strangeness with non-empty directories.  We can likely standardize this into a macro (e.g. `SUBMODULE_CHECK(swh/configure.ac)`) and use it as a standard moving forward.  It's a very basic check but it can help us move contributors away from confusing error messages.  This will become more important as we add more submodules.  So far, we have `malloc`, `zynaddsubfx` and `swh` .  This list is going to grow.

2. This PR contains patches that don't belong in LMMS source code but are required for compilation.  [Upstream PRs have been opened](https://github.com/swh/ladspa/pulls?utf8=%E2%9C%93&q=is%3Apr%20author%3Atresf%20) and will supercede those patches.  I'm going to give Steve Harris (swh) a few days to merge, which he's very good about.  Then I'll bump the submodule to the `HEAD` of upstream.

3. If anyone has feedback on the ["rube-goldberg"](https://en.wikipedia.org/wiki/Rube_Goldberg) approach to building this project, speak now.  I've considered asking upstream to switch to `cmake`, but the author no longer maintains this, so this would have to be done by a community member instead.  The perl/xml template dependency makes this project a bad candidate for cmake I feel, but thoughts are welcome.

4. Speaking of ^rube goldberg, there's an inefficiency added to compilation where the `.xml` files are coerced into `.c` files every time `cmake` is run.  These should be cached and only updated as the `.xml` files are changed, but I don't know the best way to do this.  Advice is appreciated.  If I can't figure it out, I'll leave the `FIXME` in there and attack it at a later time.

lukas-w:
> 1. We can likely standardize this [submodule guard] into a macro (e.g. `SUBMODULE_CHECK(swh/configure.ac)`) and use it as a standard moving forward. It's a very basic check but it can help us move contributors away from confusing error messages.

Good idea 👍. We can even run `git` to fetch the submodule automatically if it's missing.

> 3. If anyone has feedback on the "rube-goldberg" approach to building this project, speak now.

I'm slightly worried about what this means for building on Windows, especially regarding the perl dependency. It may be more portable to generate the `.c` files once, upload them on e.g. Github Releases, and fetch them at build time using `ExternalProject_Add`. But then again, the `.c` files will likely need fixing for MSVC anyway.

> 4. […] there's an inefficiency added to compilation where the .xml files are coerced into .c files every time cmake is run. These should be cached and only updated as the .xml files are changed, but I don't know the best way to do this.

Use [`add_custom_target`](https://cmake.org/cmake/help/v3.0/command/add_custom_target.html) or [`add_custom_command`](https://cmake.org/cmake/help/v3.0/command/add_custom_command.html) instead of `execute_process`. They have a `DEPENDS` option where you can specify a list of files to watch for changes.

Something like this could work:

```cmake
FOREACH(_item ${XML_SOURCES})
	SET(out_file "${_plugin}.c")
	GET_FILENAME_COMPONENT(_plugin "${_item}" NAME_WE)
	ADD_CUSTOM_COMMAND(
		OUTPUT ${out_file}
		COMMAND ./makestub.pl ${_item} > ${out_file}
		DEPENDS ${_item}
		WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/swh
		VERBATIM
	)
	LIST(APPEND PLUGIN_SOURCES ${out_file})
ENDFOREACH()
```

Then use the resulting `PLUGIN_SOURCES` list instead of globbing.

tresf:
@lukas-w done via 38d0e37 as well as some other cleanup and the first draft of the submodule/git script.

lukas-w:
Great 👍 I'll have another look at the CMake scripts soon.

cmake/modules/CheckSubmodule.cmake
```diff
@@ -0,0 +1,67 @@
...
+	STRING(REPLACE "/"  ";" PARTS ${RELATIVE_FILE})
+	LIST(GET PARTS 0 SUBMODULE_NAME)
+	SET(ABSOLUTE_FILE "${CMAKE_CURRENT_SOURCE_DIR}/${RELATIVE_FILE}")
+	SET(SUBMODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/${SUBMODULE_NAME}")
+
+	# Attempt to do lazy clone
+	IF(NOT EXISTS "${ABSOLUTE_FILE}")
+		MESSAGE("-- Submodule '${SUBMODULE_NAME}' missing")
+		MESSAGE("--   Cloning '${SUBMODULE_NAME}' using git")
+		CHECK_COMMAND(git)
```
lukas-w: I think it's better to just use `FIND_PACKAGE(Git)` here instead. No need for an extra CMake module.

tresf: I'll switch to `FIND_PACKAGE(Git)` and use `GIT_EXECUTABLE` variable, thanks.  This module was originally added to look for some of the automake/autoconf/autopoint tools that weren't trivial to locate, but I've since realized we don't need them if we build with cmake.  Thanks.  I'll retire the module for now.

tresf:
The following upstream patches have been merged:

* https://github.com/swh/ladspa/pull/56 - Remove broken lookahead_limiter (So we can glob everything)
* https://github.com/swh/ladspa/pull/53 - Use List::MoreUtils instead of List::Util (Ubuntu 14.04 compat)
* https://github.com/swh/ladspa/pull/51 - Allow older versions of gettext (Ubuntu 14.04 compat)

The following upstream patches are still open:

* https://github.com/swh/ladspa/pull/52 - Windows compat fixes

.. once merged all [patches and plugin blacklisting](https://github.com/tresf/lmms/blob/ea69da7b0a9a7ddbac167ba743b717129ecf5e1f/plugins/LadspaEffect/swh/CMakeLists.txt#L5-L30) can safely be removed from this PR and we can simply fast-forward the submodule to be in sync with swh/ladspa upstream master.

lukas-w:
Merge now or wait for https://github.com/swh/ladspa/pull/52 to be accepted?

tresf:
My instinct is to wait.

tresf:
Upstream has been merged.  Squashed to two separate commits.  This PR should be "rebased and merged" once Travis is looking OK.

tresf:
Dependencies for Ubuntu have been updated.  Note, I have not updated Fedora, Debian or the others at this time.  I also do not recommend updating Qt4 listing as we will drop Qt4 support with the master branch as soon as someone has the time to do so.

Another disclaimer for Windows coders... (e.g. @DomClark), this will also break `master` builds on Windows desktop with `WANT_SWH` enabled until we have the tools installed and documented properly.  Our Windows build tutorial is grossly outdated since switching to Qt5, so this is a separate task and any help is appreciated.

PR #3937, opened 2017-11-04
Init Automation Tracks on song start

zonkmachine:
If the automation value isn't the same as the controlled values default (or the last value prior to stop) that value will skip into place on the first frame and which can lead to an audible artifact. 

Example project with filter sweep. It will present a 'tick' on play, either on load/play, or on stop/play a 'tick' sound of varying level depending on where in the song you stop and rewind to '0'.
[sweep.mmp.zip](https://github.com/LMMS/lmms/files/1443426/sweep.mmp.zip)

### DIFF
--- a/src/core/Song.cpp
+++ b/src/core/Song.cpp
@@ -266,13 +266,22 @@ void Song::processNextBuffer()
 	}
 
 	f_cnt_t framesPlayed = 0;
+	f_cnt_t framesToPlay = Engine::mixer()->framesPerPeriod() - framesPlayed;
+
+	float currentFrame = m_playPos[m_playMode].currentFrame();
+	if( ( f_cnt_t ) currentFrame == 0 )
+	{
+		processAutomations(trackList, m_playPos[m_playMode], framesToPlay);
+	}
+
 	const float framesPerTick = Engine::framesPerTick();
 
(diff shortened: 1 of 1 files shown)


src/core/Song.cpp
```diff
@@ -266,13 +266,22 @@ void Song::processNextBuffer()
 	}
 
 	f_cnt_t framesPlayed = 0;
+	f_cnt_t framesToPlay = Engine::mixer()->framesPerPeriod() - framesPlayed;
+
+	float currentFrame = m_playPos[m_playMode].currentFrame();
+	if( ( f_cnt_t ) currentFrame == 0 )
+	{
+		processAutomations(trackList, m_playPos[m_playMode], framesToPlay);
```
zonkmachine: Well, it works. I'm still reading up on `Song::processAutomations( ... )` though.

PhysSong:
I think it should be fixed on `AutomatableModel`'s side, not the `Song`'s side.
Anyway, this is a little bit tricky problem. I'm going to work on this issue and open a new PR if I find a better way.

zonkmachine:
> I think it should be fixed on AutomatableModel's side, not the Song's side.

Aight... model stuff. Close this then. :)

PR #3939, opened 2017-11-04
Remove text from project notes of factory templates.

### LINKED ISSUE #3613: Remove Message From Project Notes
Doesn't necessarily matter, but probably can remove the
"Put down your project notes here." from the Project Notes window.
In 2017, probably don't need to tell people how to use a text editor, especially when the window itself is called "Project Notes" :)
This message also causes the user to have to highlight and delete it, in order to start typing.
Sure many will not care about this but, just thought I would mention it.

Sawuare:
Closes #3613.


tresf:
Rebased against `stable-1.2`, merging.

Issue #3636, opened 2017-06-13
Improve the cancel function of loading projects.


zonkmachine:
When testing to load, play and close lmms I have had an on/off problem with projects not loading completely. As it turns out I had forgotten that there is a cancel button when loading and pressed play too fast which instead triggered the cancel action. I think this button should be pressed specifically and not be tied to any key but perhaps `ESC` . Or we could have some message function confirming the cancel. A text float or whatnot.
Maybe it's best to also clean out the project on a cancel?

Umcaruje:
I agree that when we cancel the project, it should get cleaned out.

BiRD4:
I think ESC and ENTER are both fine to use. I also agree that there should be a message to confirm the cancel, but I think this should keep the project from continuing to load until the user has confirmed the cancel. However, I think it might be nice to have the option to either clean out the project or keep only what has already been loaded.

zonkmachine:
Bump @PhysSong You're working on other issues around project loading already (#3611,  #3670) so maybe you could have a look at this?

PhysSong:
@zonkmachine I think it won't be hard. I'll have a look.
I'll suggest something or make some PRs if I can.

zonkmachine:
Here is an issue reported in the LMMS forum that could be related:
https://lmms.io/forum/viewtopic.php?f=7&t=8505
I'm not 100% sure it is related though but I've tried and cancel big projects and have gotten some positively odd results that wasn't obviously broken if you didn't know that a cancel had taken place. I'm labelling this as a bug instead.

PhysSong:
Someone can cancel project loading accidentally add save the project without noticing that isn't loaded completely. Then some unloaded track will be lost.
Such accidents would be prevented. At least, we should notify user of the cancellation.

musikBear:
Imo pressing cancel mean that the user made a mistake in the project selection, or simply does not want to use lmms now at all.
Cancel should just as @zonkmachine says clean away the project from screen, and perhaps ask the user if he also want to quit lmms completely.
The situation that @PhysSong depict, and that maby even is the reason for forum8505, should never happen. Eg, partial loading should be prevented!

PhysSong:
@musikBear I agree. Partial loading brings many problems.
Sometimes it makes sense, but usually not.

PhysSong:
@zonkmachine Do you know how other programs handle cancellation of project loading? If yes, which method would be better?

tresf:
[...]

Issue #3670, opened 2017-06-28
LMMS crashes when project loading is cancelled


PhysSong:
How to reproduce:
- Open demo project Jousboxx-BuzzerBeater.mmpz
- Press Esc key when progress is about 20%

Tested version: 1.2.0-rc3

The error message is: `ASSERT: "c != NULL" in file /home/hjsong/lmms-stable-1.2/src/core/Song.cpp, line 1264`

In the project, FX Ch11 uses peak controller from 'Sidechain' in B/B track, but this track wasn't load yet when I cancel loading.

Here's another example [test.mmp.zip](https://github.com/LMMS/lmms/files/1107624/test.mmp.zip).

Song::restoreControllerStates() should handle this situation. Unfortunate
ly it doesn't.

If there's someone who has an idea, please write a comment.

Edit: When I removed Q_ASSERT(), there was no crash. However, in the context menu of knob which was connected to peak controller, LMMS says there is an controller connection(there can't be such connections because the controller isn't loaded).

zonkmachine:
> Edit: When I removed Q_ASSERT(), there was no crash. However, in the context menu of knob which was connected to peak controller, LMMS says there is an controller connection(there can't be such connections because the controller isn't loaded).

I think this is how the user will experience it as the assert is probably only there if you build with the debug  flag.

PhysSong:
Then It is fine to fix the bug only?

PhysSong:
Fixed in #3941, with #3636.

PR #3941, opened 2017-11-05
Default project when cancelling project loading

zonkmachine:
If a user isn't aware that the loading of a project has been
cancelled, it may be in an incomplete state. Saving such a project
will overwrite the original file and result in data loss. This is
solved by loading the default project on cancelling project loading.

Fixes https://github.com/LMMS/lmms/issues/3636

### DIFF
--- a/src/core/Song.cpp
+++ b/src/core/Song.cpp
@@ -86,6 +86,7 @@ Song::Song() :
 	m_playing( false ),
 	m_paused( false ),
 	m_loadingProject( false ),
+	m_isCancelled( false ),
 	m_playMode( Mode_None ),
 	m_length( 0 ),
 	m_patternToPlay( NULL ),
@@ -1074,7 +1075,7 @@ void Song::loadProject( const QString & fileName )
 		}
 	}
 
-	while( !node.isNull() )
+	while( !node.isNull() && !isCancelled() )
 	{
 		if( node.isElement() )
 		{
@@ -1132,6 +1133,13 @@ void Song::loadProject( const QString & fileName )
 	Engine::projectJournal()->setJournalling( true );
 
 	emit projectLoaded();
+
+	if( isCancelled() )
+	{
+		m_isCancelled = false;
+		createNewProject();
+		return;
+	}
 
 	if ( hasErrors())
 	{
@@ -1278,7 +1286,7 @@ void Song::saveControllerStates( QDomDocument & doc, QDomElement & element )
 void Song::restoreControllerStates( const QDomElement & element )
 {
 	QDomNode node = element.firstChild();
-	while( !node.isNull() )
+	while( !node.isNull() && !isCancelled() )
 	{
 		Controller * c = Controller::create( node.toElement(), this );
 		Q_ASSERT( c != NULL );
--- a/include/Mixer.h
+++ b/include/Mixer.h
@@ -155,6 +155,7 @@ public:
 
 	void initDevices();
 	void clear();
+	void clearNewPlayHandles();
 
 
 	// audio-device-stuff
--- a/include/Song.h
+++ b/include/Song.h
@@ -229,6 +229,16 @@ public:
 		return m_loadingProject;
 	}
 
+	void loadingCancelled()
+	{
+		m_isCancelled = true;
+	}
+
+	bool isCancelled()
+	{
+		return m_isCancelled;
+	}
+
 	bool isModified() const
 	{
 		return m_modified;
@@ -358,6 +368,7 @@ private:
 	volatile bool m_paused;
 
 	bool m_loadingProject;
+	bool m_isCancelled;
 
 	QStringList m_errors;
 
--- a/include/Song.h
+++ b/include/Song.h
@@ -33,6 +33,7 @@
 #include "TrackContainer.h"
 #include "Controller.h"
 #include "MeterModel.h"
+#include "Mixer.h"
 #include "VstSyncController.h"
 
 
@@ -232,6 +233,7 @@ public:
 	void loadingCancelled()
 	{
 		m_isCancelled = true;
+		Engine::mixer()->clearNewPlayHandles();
 	}
 
 	bool isCancelled()
--- a/src/core/Mixer.cpp
+++ b/src/core/Mixer.cpp
@@ -501,6 +501,20 @@ void Mixer::clear()
 
 
 
+void Mixer::clearNewPlayHandles()
+{
+	requestChangeInModel();
+	for( LocklessListElement * e = m_newPlayHandles.popList(); e; )
+	{
+		LocklessListElement * next = e->next;
+		m_newPlayHandles.free( e );
+		e = next;
+	}
+	doneChangeInModel();
+}
+
+
+
 // removes all play-handles. this is necessary, when the song is stopped ->
 // all remaining notes etc. would be played until their end
 void Mixer::clearInternal()
--- a/src/core/TrackContainer.cpp
+++ b/src/core/TrackContainer.cpp
@@ -33,12 +33,14 @@
 #include "AutomationTrack.h"
 #include "BBTrack.h"
 #include "BBTrackContainer.h"
+#include "embed.h"
 #include "TrackContainer.h"
 #include "InstrumentTrack.h"
 #include "Song.h"
 
 #include "GuiApplication.h"
 #include "MainWindow.h"
+#include "TextFloat.h"
 
 TrackContainer::TrackContainer() :
 	Model( NULL ),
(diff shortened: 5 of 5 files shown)


src/core/Song.cpp
```diff
@@ -1150,6 +1151,12 @@ void Song::loadProject( const QString & fileName )
 	m_modified = false;
 	m_loadOnLaunch = false;
 
+	if( isCancelled() )
+	{
+		m_isCancelled = false;
+		createNewProject();
```
zonkmachine: This should be done earlier. At least before [the error summary](https://github.com/LMMS/lmms/pull/3941/files#diff-bc7aaaf624969ba372535c9eab5ed490R1137).

PhysSong:
Seems to work fine. However, I want to test this more for some reason. After it, I'll either approve it or request some changes.

zonkmachine:
I've tested to not adding cancelled programs to the recent files list but found that rather unintuitive actually. The brain seem to add it in there for my inner vision anyway and when I reach for the recent project list I'm 'off by one'.

PhysSong:
@zonkmachine Do you have a backtrace? Can you provide some more details?

zonkmachine:
```
(gdb) bt full
#0  0x0000000000523362 in InstrumentPlayHandle::play (this=0x1c93c5c0, _working_buffer=0x1000cf0020)
    at /home/zonkmachine/builds/lmms/lmms/include/InstrumentPlayHandle.h:46
        nphv = {{p = {static shared_null = {ref = {_q_value = 12591}, alloc = 0, begin = 0, end = 0, sharable = 1, array = {0x0}}, d = 0x7fffb1e47c30}, 
            d = 0x7fffb1e47c30}}
        nphsLeft = false
#1  0x0000000000544abf in PlayHandle::doProcessing (this=0x1c93c5c0) at /home/zonkmachine/builds/lmms/lmms/src/core/PlayHandle.cpp:58
No locals.
#2  0x0000000000523221 in ThreadableJob::process (this=0x1c93c5c0) at /home/zonkmachine/builds/lmms/lmms/include/ThreadableJob.h:74
No locals.
#3  0x00000000005393b3 in MixerWorkerThread::JobQueue::run (this=0x9634e0 <MixerWorkerThread::globalJobQueue>)
    at /home/zonkmachine/builds/lmms/lmms/src/core/MixerWorkerThread.cpp:74
        job = 0x1c93c5c0
        i = 0
        processedJob = false
#4  0x000000000053960d in MixerWorkerThread::startAndWaitForJobs () at /home/zonkmachine/builds/lmms/lmms/src/core/MixerWorkerThread.cpp:147
No locals.
#5  0x00000000005340c7 in Mixer::renderNextBuffer (this=0xc44190) at /home/zonkmachine/builds/lmms/lmms/src/core/Mixer.cpp:440
        playModeSupportsMetronome = false
        it_rem = {i = 0x7fffa80013e8}
        fxMixer = 0xc2e3c0
        last_metro_pos = {<MidiTime> = {m_ticks = -1, static s_ticksPerTact = 192}, m_timeLine = 0x0, m_currentFrame = 0}
        song = 0xc4bf80
        currentPlayMode = Song::Mode_None
        p = {<MidiTime> = {m_ticks = 0, static s_ticksPerTact = 192}, m_timeLine = 0x0, m_currentFrame = 0}
#6  0x0000000000535c2b in Mixer::fifoWriter::run (this=0x10e50f0) at /home/zonkmachine/builds/lmms/lmms/src/core/Mixer.cpp:1103
        buffer = 0x7fffa80031f0
        b = 0xc8c440
        frames = 256
#7  0x00007ffff685a32f in QThreadPrivate::start (arg=0x10e50f0) at thread/qthread_unix.cpp:349
        __clframe = {__cancel_routine = 0x7ffff6859300 <QThreadPrivate::finish(void*)>, __cancel_arg = 0x10e50f0, __do_it = 1, 
          __cancel_type = <optimized out>}
        thr = 0x10e50f0
        data = <optimized out>
        objectName = {static null = {<No data fields>}, static shared_null = {ref = {_q_value = 18604}, alloc = 0, size = 0, 
            data = 0x7ffff6cc293a <QString::shared_null+26>, clean = 0, simpletext = 0, righttoleft = 0, asciiCache = 0, capacity = 0, reserved = 0, 
            array = {0}}, static shared_empty = {ref = {_q_value = 180}, alloc = 0, size = 0, data = 0x7ffff6cc291a <QString::shared_empty+26>, clean = 0, 
            simpletext = 0, righttoleft = 0, asciiCache = 0, capacity = 0, reserved = 0, array = {0}}, d = 0x7ffff6cc2920 <QString::shared_null>, 
          static codecForCStrings = 0x0}
#8  0x00007ffff7bc4184 in start_thread (arg=0x7fffb1e48700) at pthread_create.c:312
        __res = <optimized out>
        pd = 0x7fffb1e48700
        now = <optimized out>
        unwind_buf = {cancel_jmp_buf = {{jmp_buf = {140736177932032, 2699179435038421571, 1, 0, 140736177932736, 140736177932032, -2699333285049191869, 
                -2699197564588445117}, mask_was_saved = 0}}, priv = {pad = {0x0, 0x0, 0x0, 0x0}, data = {prev = 0x0, cleanup = 0x0, canceltype = 0}}}
        not_first_call = <optimized out>
        pagesize_m1 = <optimized out>
        sp = <optimized out>
        freesize = <optimized out>
        __PRETTY_FUNCTION__ = "start_thread"
#9  0x00007ffff3f8affd in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:111

```

zonkmachine:
` QThreadPrivate::start (`

So it kicks off a new thread somewhere.

zonkmachine:
What I was doing is cancelling the project at the very end of loading unfa/spoken. I've seen this more than once and only after hitting the 90% mark.

PhysSong:
I got it. I think you were loading a project with single-streamed instrument(s) and canceled it. I have a simple way to fix that, related to mixer. May I just push that to your branch tomorrow?

zonkmachine:
> May I just push that to your branch tomorrow?

You may.

src/core/Song.cpp
```diff
@@ -1133,6 +1134,13 @@ void Song::loadProject( const QString & fileName )
 
 	emit projectLoaded();
 
+	if( isCancelled() )
```
PhysSong: I think this should be right after the `while` loop. But then LMMS will crash if you cancel loading a project which has single-streamed instruments. However, I have a fix for this as I mentioned before, and I'm going to push this today.

Anyway it works nice for projects without such instruments. :+1:

PhysSong:
FYI, it fixes #3670 too.

PhysSong:
Instead of direct push, I'll provide a patch file for that. I was a little bit busy...
> I've tested to not adding cancelled programs to the recent files list but found that rather unintuitive actually.

What do you exactly mean there?

zonkmachine:
>> I've tested to not adding cancelled programs to the recent files list but found that rather unintuitive actually.
> 
> What do you exactly mean there?

Not very important. If I break loading before [this line](https://github.com/LMMS/lmms/blob/7ed9bea9fbded61b4ba2d32113e442c38f84a9b9/src/core/Song.cpp#L1130), or move it to after the break, the project will not be bumped up to the top of the recent projects list. I tried this and was surprised at how unintuitive it felt. I'm pretty sure I would get used to it quickly but as a first impression it was a bit odd.

zonkmachine:
> Instead of direct push, I'll provide a patch file for that.

Aight, works fine for me!

PhysSong:
> the project will not be bumped up to the top of the recent projects list.

Looks minor, but a little bit tricky... Users may cancel project loading because of several reasons. For some situation(ex. clicked wrong file), that behavior is appropriate, but current one may be better for other situations.

tresf:
> Not very important. If I break loading before this line, or move it to after the break, the project will not be bumped up to the top of the recent projects list. I tried this and was surprised at how unintuitive it felt. I'm pretty sure I would get used to it quickly but as a first impression it was a bit odd.

As a baseline comparison I just tested this with Microsoft Excel (it takes a while to load as well) and the behavior is identical to what @zonkmachine describes feels normal.  i.e. Even if you cancel the loading, it still shows at the top of recent list.

PhysSong:
Sorry to say that, but I lost my works by accident. Adding a function in `Mixer` class will be useful.
```c++
void Mixer::clearNewPlayHandles()
{
	requestChangeInModel();
	for( LocklessListElement * e = m_newPlayHandles.popList(); e; )
	{
		LocklessListElement * next = e->next;
		m_newPlayHandles.free( e );
		e = next;
	}
	doneChangeInModel();
}
```
To fix the crash, call this function when project loading is canceled.

zonkmachine:
> To fix the crash, call this function when project loading is canceled.

OK, thanks! https://github.com/LMMS/lmms/pull/3941/commits/14182a310b337a5f68aa879478b8ed826de89b8b
I'm calling `Mixer::clearNewPlayHandles()` from Song.h and add Mixer.h as a dependency. No more crashes reproducible. :)

zonkmachine:
> For some situation(ex. clicked wrong file), that behavior is appropriate,

This was the reason I tried not adding it to the list. The current behaviour is to add a project to the recent list even if it's cancelled. I think it's best to leave it as it is for now though I have no real opinion on how this should best be handled. If someone feels strongly about it then I think it's better to look into it for 1.3 or later.

Merging.

PR #3945, opened 2017-11-06
Fix visual glitch with automation patterns

### LINKED ISSUE #3935: Automation pattern glitch
Since the Anti-Aliasing fix of the Automation Editor (899e386df71bb0d88e0b6cd873348ec2b9bf9ad1)  there is a step left at the last automation point. The pattern is somehow off by one.

automationpatternoff

PhysSong:
Fixes #3935.

### DIFF
--- a/src/gui/AutomationPatternView.cpp
+++ b/src/gui/AutomationPatternView.cpp
@@ -319,6 +319,16 @@ void AutomationPatternView::paintEvent( QPaintEvent * )
 
 		float *values = m_pat->valuesAfter( it.key() );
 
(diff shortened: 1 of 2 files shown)


zonkmachine:
Fixed, in the Automation Editor. You can actually see this in the pattern from the Song Editor too, but it's not as prominent.

PhysSong:
> You can actually see this in the pattern from the Song Editor too

Then I can add a fix for that, too. :smile:

PhysSong:
Now both issues are fixed.

PR #3947, opened 2017-11-06
Enable sample track recording

### LINKED ISSUE #3943: Sample track standalone recorder
Hello guys, 

I'm currently working on a standalone sound recorder for LMMS. Most of the code is already there (`SampleRecordHandle`). The only thing I've implemented is the actual recording for each device type (e.g. `ALSA`, `JACK`, ...). 

Surprisingly, for JACK, it was pretty trivial so now JACK and PortAudio support recording and therefore support sample track recording too.

Just before I take another step, am I doing it right? Way was this feature disabled?

Thank you all :)

Reflexe:
## Download links for testing: 
[here](https://github.com/Reflexe/lmms/releases/tag/recording-dream).

many thanks to @tresf!

## Progress
- [x] Support capture in major interfaces:
  - [x] JACK
  - [x] PulseAudio
  - [x] SDL
- [x] Make recording work.
- [x] Make recording usable for humans
    - [x] Recording should record into a new TCO in the *track* or tracks that has been marked as record.  
        - [x] ~~If no track was marked, it shall create a new track and mark it as record~~.
    - [x] (Optional) Leave a way to record into a specific TCO (the current behaviour).
         -  [x] (Optional) Change the color of a TCO that has been set as record.
    - [x] ~~(Need to decide about the correct behaviour) Record according to another sample track: start and stop according to it.~~
    - [x] ~~Add per sample track audio channel modifiers (e.g. solo right, right both).~~
    - [x] Hide the regular record button on the song editor.

## Issues
 - [x] JACK does not connect lmms to the system input by default.
 - [x] Sometimes with SDL, a few seconds last does not get captured by lmms. (most of the times)
 - [x] For some reason, choosing a global record channel does not work correctly. 
Closes #3943


gi0e5b06:
1) Great. Looks good for me.
I would merge the Jack part so we can test.

2) Please add Alsa.

3) While you're working on these files, could you change the printf() to qWarning()/qCritical() ?
Not urgent but it has do be done anyway.

lukas-w:
> While you're working on these files, could you change the printf() to qWarning()/qCritical() ?

No, this has nothing to do with this PR. Don't encourage people to do PRs the way you do them.

qnebra:
In my case compiler got this:

/home/lukasz/lmms/compile/lmms/src/core/audio/AudioSdl.cpp: In member function ‘void AudioSdl::sdlInputAudioCallback(Uint8*, int)’:
/home/lukasz/lmms/compile/lmms/src/core/audio/AudioSdl.cpp:312:30: error: no matching function for call to ‘Mixer::pushInputFrames(sample_t (*&)[2], fpp_t&, bool)’
          true /* applyGain */);

When lmms is compiled with your patch.

BaraMGB:
I have tested this for ubuntu/Jack. I can record my voice in LMMS :tada:. That's great. I found some issues with the sample track resize from left, which makes no sense in an empty or fresh recorded tco. It should be set to 0 in this case.

Reflexe:
@qnebra, My bad. I forgot to commit some files.

Reflexe:
@BaraMGB I think I've fixed the issue with resizing from the left. Can you test it again please?

BaraMGB:
Okay, tested. Works. 

Next problems: 

1. When I start the recording in the middle of the TCO:
bildschirmfoto von 2017-11-08 23-32-46
The recorded sample moved to the beginning of the TCO:
bildschirmfoto von 2017-11-08 23-33-17

2. Recording in a loop doesn't work at all. When the Playhead moves back to the beginning of the loop, the recording stops and goes into the play mode. 

3. We should find a way for the user to choose the channel(s). When I want to record my voice I use only the left channel of my interface. But in the current implementation I have my voice only in the left speaker.

Reflexe:
Current issues:
 * Record button does not work (Press record, press play)
 * Loops don't work more than once (See @BaraMGB's comment)

BaraMGB:
Here is a little bug, yet. When you recording start in the middle of the TCO you add a startTimeOffset() now. This is correct. But now: If you wanna record in this TCO the recording don't start at the beginning of the TCO. The recording starts only if the Playhead reaches the start+startTimeOffset() position. 

I hope you understand me.

src/core/audio/AudioJack.cpp
```diff
@@ -405,6 +430,21 @@ int AudioJack::processCallback( jack_nframes_t _nframes, void * _udata )
...
+		jack_default_audio_sample_t *channel_buffer = m_tempInBufs[c];
+
+		for( jack_nframes_t frame = 0; frame < _nframes; ++frame )
+		{
+			m_inBuffer[frame][c] = channel_buffer[frame] / gain;
+		}
+	}
+
+	mixer()->pushInputFrames (m_inBuffer.get (), _nframes);
+
```
PhysSong: This whole logic should be run if and only if JACK is running(in other words, `m_stopped` is false). Otherwise, it will cause deadlock after you export audio with sample rate different from JACK's processing sample rate.
LMMS doesn't deactivate the client when the `AudioJack` interface stops processing.

Reflexe: Just to be sure, where is the deadlock actually happening?

PhysSong: JACK callback calls `requestChangeInModel()` which waits for mixer, before the mixer starts. The mixer can't be started because the function blocks starting the mixer thread.
Relevant parts of backtrace:
```gdb
Thread 10 (Thread 0x7fffd7887700 (LWP 27891)):
#0  pthread_cond_wait@@GLIBC_2.3.2 ()
    at ../sysdeps/unix/sysv/linux/x86_64/pthread_cond_wait.S:185
#1  0x00007ffff48baa5b in QWaitCondition::wait(QMutex*, unsigned long) ()
   from /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#2  0x00000000005322ed in Mixer::requestChangeInModel() ()
#3  0x000000000053239f in Mixer::pushInputFrames(float const (*) [2], int, bool) ()
#4  0x000000000057a9ef in AudioJack::processCallback(unsigned int, void*) ()


Thread 1 (Thread 0x7ffff7ee0900 (LWP 27879)):
#0  syscall () at ../sysdeps/unix/sysv/linux/x86_64/syscall.S:38
#1  0x00007ffff48b18b8 in QBasicMutex::lockInternal() ()
   from /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#2  0x00007ffff48b19ad in QMutex::lock() ()
   from /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#3  0x0000000000532299 in Mixer::requestChangeInModel() ()
#4  0x00000000005572b2 in SampleBuffer::update(bool) ()
#5  0x00007ffff4ac8d2a in QMetaObject::activate(QObject*, int, int, void**) ()
   from /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#6  0x00000000005336bb in Mixer::restoreAudioDevice() ()
```
FYI `QMetaObject::activate` is called here: https://github.com/LMMS/lmms/blob/03e9889605fbe729bc479412b9d1767f5b39103d/src/core/Mixer.cpp#L641

Reflexe: Fixed.

src/core/audio/AudioJack.cpp
```diff
@@ -403,7 +428,23 @@ int AudioJack::processCallback( jack_nframes_t _nframes, void * _udata )
...
+            jack_default_audio_sample_t *channel_buffer = m_tempInBufs[c];
+
+            for( jack_nframes_t frame = 0; frame < _nframes; ++frame )
+            {
+                m_inBuffer[frame][c] = channel_buffer[frame] / gain;
+            }
+        }
+
+        mixer()->pushInputFrames (m_inBuffer.get (), _nframes);
+    }
```
PhysSong: You should use tabs instead of spaces from Line 431 to here.

include/AudioPulseAudio.h
```diff
@@ -71,12 +71,16 @@ class AudioPulseAudio : public AudioDevice, public QThread
 
 
 	void streamWriteCallback( pa_stream * s, size_t length );
+    void streamReadCallback(pa_stream *s, size_t length);
```
PhysSong: Spaces.

src/core/audio/AudioPulseAudio.cpp
```diff
@@ -42,23 +42,30 @@ static void stream_write_callback(pa_stream *s, size_t length, void *userdata)
 }
 
 
-
+static void stream_read_callback(pa_stream *s, size_t length, void *userdata)
+{
+    static_cast<AudioPulseAudio *>( userdata )->streamReadCallback( s, length );
```
PhysSong: Here, too.

src/gui/editors/SongEditor.cpp
```diff
@@ -655,6 +656,8 @@ SongEditorWindow::SongEditorWindow(Song* song) :
 	m_editor(new SongEditor(song)),
 	m_crtlAction( NULL )
 {
+    bool isRecordSupported = Engine::mixer()->audioDev()->supportsCapture();
```
PhysSong: And here

src/gui/editors/SongEditor.cpp
```diff
@@ -839,3 +877,8 @@ void SongEditorWindow::keyReleaseEvent( QKeyEvent *ke )
 		}
 	}
 }
+
+SampleTrack::RecordingChannel SongEditorWindow::globalRecordChannel() const
+{
+    return m_globalRecordChannel;
```
PhysSong: And finally here.

src/core/Mixer.cpp
```diff
@@ -309,8 +309,9 @@ bool Mixer::criticalXRuns() const
 
 
 
-void Mixer::pushInputFrames( sampleFrame * _ab, const f_cnt_t _frames )
-{
+void Mixer::pushInputFrames(const sampleFrame * _ab, const f_cnt_t _frames,
+							 bool shouldApplyMasterGain)
+{
```
PhysSong: I found an unnecessary tab here.

Reflexe:
@PhysSong, Thanks! I'll use `NULL` to reserve the previous behaviour.

src/core/audio/AudioPulseAudio.cpp
```diff
@@ -265,35 +306,59 @@ void AudioPulseAudio::streamWriteCallback( pa_stream *s, size_t length )
 {
 	const fpp_t fpp = mixer()->framesPerPeriod();
 	surroundSampleFrame * temp = new surroundSampleFrame[fpp];
```
Reflexe: We need to memset it to zero, or just make sure getNextBuffer always fill the buffer with zeros if there is no data.

tresf:
@Reflexe says this is ready for review and merge.  For those capable, please test this out.  If someone has a chance to upload some installers, please do so.

src/tracks/SampleTrack.cpp
```diff
@@ -587,7 +587,12 @@ SampleTrack::SampleTrack( TrackContainer* tc ) :
 							tr( "Volume" ) ),
 	m_panningModel( DefaultPanning, PanningLeft, PanningRight, 0.1f,
 					this, tr( "Panning" ) ),
-	m_audioPort( tr( "Sample track" ), true, &m_volumeModel, &m_panningModel, &m_mutedModel )
+	m_audioPort( tr( "Sample track" ), true, &m_volumeModel, &m_panningModel, &m_mutedModel ),
+	m_recordingChannelModel(RecordingChannel::None,
```
PhysSong: The order of declaration and initialization differs. It causes compilation warning. This is why Travis failed.

Reflexe: I've got overwhelmed with the amount of warnings clang had threw on me (For some reason, `-Weverything` was on)

BaraMGB:
@Reflexe thank you for implementing this to lmms. This feature is a big jump forward. I'll test this out after Xmas.

mendomusic:
I cannot compile your fork.

```
Installation Summary
--------------------
* Install Directory           : /usr/local

Supported audio interfaces
--------------------------
* ALSA                        : OK
* JACK                        : OK (weak linking enabled)
* OSS                         : OK
* Sndio                       : <not found or not supported on this platform>
* PortAudio                   : OK
* libsoundio                  : OK
* PulseAudio                  : OK
* SDL                         : OK, using SDL2

Supported MIDI interfaces
-------------------------
* ALSA                        : OK
* OSS                         : OK
* Sndio                       : <not found or not supported on this platform>
* JACK                        : OK (weak linking enabled)
* WinMM                       : <not supported on this platform>
* AppleMidi                   : <not supported on this platform>

Supported file formats for project export
-----------------------------------------
* WAVE                        : OK
* FLAC                        : OK
* OGG/VORBIS                  : OK
* MP3/Lame                    : OK

Optional plugins
----------------
* Carla Patchbay & Rack       : OK
* SoundFont2 player           : OK
* Stk Mallets                 : OK
* VST-instrument hoster       : not found, please install (lib)wine-dev (or similar) - 64 bit systems additionally need gcc-multilib and g++-multilib
* VST-effect hoster           : not found, please install (lib)wine-dev (or similar) - 64 bit systems additionally need gcc-multilib and g++-multilib
* CALF LADSPA plugins         : OK
* CAPS LADSPA plugins         : OK
* CMT LADSPA plugins          : OK
* TAP LADSPA plugins          : OK
* SWH LADSPA plugins          : OK
* GIG player                  : OK

Developer options
-----------------------------------------
* Debug FP exceptions         : Disabled


-----------------------------------------------------------------
IMPORTANT:
after installing missing packages, remove CMakeCache.txt before
running cmake again!
-----------------------------------------------------------------

```

tresf:
@mendomusic your log doesn't show an error.  You'll have to share some more.  Perhaps in #support on [Discord](https://lmms.io/chat)

mendomusic:
@tresf 
CMakeError.log    --> https://pastebin.com/FLzHFgB5
CMakeOutput.log --> https://pastebin.com/09x41HGB

tresf:
@mendomusic that error suggests that you don't have the alsa development libraries installed.

mendomusic:
@tresf thanks. what package do I exactly need (debian based distro)?

mendomusic:
@Reflexe Could you provide a little instructionon how to enable recording?

src/core/audio/AudioSdl.cpp
```diff
@@ -107,10 +161,24 @@ void AudioSdl::stopProcessing()
 {
 	if( SDL_GetAudioStatus() == SDL_AUDIO_PLAYING )
```
PhysSong: @Reflexe LMMS crashes when rendering with SDL2 + this PR because this function doesn't stop playback correctly. You need to use `SDL_GetAudioDeviceStatus(m_outputDevice)` for SDL2 to fix this, since you used `SDL_OpenAudioDevice()`.

src/core/audio/AudioJack.cpp
```diff
@@ -411,6 +436,22 @@ int AudioJack::processCallback( jack_nframes_t _nframes, void * _udata )
...
+
+			for( jack_nframes_t frame = 0; frame < _nframes; ++frame )
+			{
+				m_inBuffer[frame][c] = channel_buffer[frame] / gain;
+			}
+		}
+
+		mixer()->pushInputFrames (m_inBuffer.get (), _nframes);
+	}
+
```
PhysSong: Trailing whitespace here.

Reflexe: I couldn't find any whitespace there, are you talking about the newline?

PhysSong: I can find two spaces in Line 454, at least in your repository.

tresf: 454 has two spaces after it.  I can't see it with the naked eye myself.  Not sure how @PhysSong spotted it.  Ninja perhaps? :D

Reflexe: Yes, I see it. Not sure how I missed it, I searched for spaces. Maybe it was github's merge that inserted it.

PhysSong: I found that in the diff page of the merge commit.

tresf:
He's aware.  Email to him:

> > Toby,
> > 
> > Is SDL2 available or on the roadmap? We have a recording feature a new dev is working on.
> >
> > -Tres

> Hi Tres,
>
> packaging SDL2 should be feasible, I'll try to deal with that soon.
> 
> Toby

Note, this would be on the new 16.04 PPA, available by brute force on 14.04. :)

tresf:
There seems to be a growing trend of using [Discord](https://lmms.io/chat) for features and bug reports.  Let's keep them here.  @Umcaruje @StakeoutPunch you're reporting some good information but placing it in the wrong place.  Anyway... Here's what @Umcaruje stated on Discord:

> *@Umcaruje wrote:*
> jack
> @Reflexe here's my feedback: When using jack LMMS didn't automatically connect to my system inputs, so I had to manually connect. Another big issue I see is that the sample track doesn't update while you're recording, now this is very bad imo because you don't know if you're recording or not.
> 
> Also when deleting another TCO from the sample track the recording stopped so I guess that's a bug
> I am so glad this is being worked on and thank you a lot for your work on this highly demanded feature
we should switch travis to use SDL2 on linux too
> since that's the default backend

In regards to SDL2, just toggle it on via `.travis/linux..install.sh` script please.  This PR seems to be the only open PR that needs it, so it can safely live here.

PhysSong:
For AppImage, we may need a PPA or some other installation source, since `SDL < 2.0.5` doesn't support recording(and Trusty ships with SDL2 2.0.2).

tresf:
@PhysSong it may put us into dependency hell, but we can try PPA pinning from 18.04.  16.04 still doesn't offer > 2.0.5.  Here's an example of PPA pinning d7aeee0.

tobydox:
SDL2 for Win32/Win64 is ready at https://launchpad.net/~tobydox/+archive/ubuntu/mingw-w64

src/tracks/SampleTrack.cpp
```diff
@@ -556,7 +569,7 @@ void SampleTCOView::paintEvent( QPaintEvent * pe )
 
 		p.setBrush( QBrush( textColor() ) );
 		p.drawEllipse( 4, 5, 4, 4 );
-	}*/
+	}
```
Sawuare: These recording indicators can overlay the sample track TCO name, for example: image
How about a different color shade for TCOs as a recording indicator?

Sawuare: Also, the current recording indicators can get cut when zooming out in the song editor.
Rec

Reflexe: thanks! I think it would be better if we'll just use a red circle and a different background.

PhysSong:
@Reflexe Can you continue this work? If you feel you can't, someone may help out.
This is a nice feature, so I hope this won't be stale for a long time.

Reflexe:
@PhysSong  Yes. Sorry for the delay. I was very busy with work. I now have a long vacation to work and hopefully finish this PR.

Reflexe:
I don't know if I should improve it more: I could add an abstraction layer for SampleBuffer and then represent it as list of buffers instead of one contagious buffer (that, to support sampleTCO rendering while recording). 

However, I think it would require more work and may introduce bugs in other areas that assume SampleBuffer is a contagious area.

Please let me know what you think.

data/themes/default/style.css
```diff
@@ -616,6 +616,7 @@ TrackContentObjectView   {
 	qproperty-textColor: #fff;
 	qproperty-textBackgroundColor: rgba(0, 0, 0, 75);
 	qproperty-textShadowColor: rgba(0,0,0,200);
+        qproperty-recordingBackgroundColor: rgba(139,25,25,0.38);
```
PhysSong: Spaces vs. Tabs

src/tracks/SampleTrack.cpp
```diff
@@ -495,9 +495,14 @@ void SampleTCOView::paintEvent( QPaintEvent * pe )
...
-	c = isSelected() ? selectedColor() : ( muted ? mutedBackgroundColor() 
-		: painter.background().color() );
+	if (isSelected ())
+		c = selectedColor ();
+	else if (muted)
+		c = mutedBackgroundColor ();
+	else if (m_tco->isRecord ())
+		c = recordingBackgroundColor ();
+	else
+		c = painter.background ().color ();
```
PhysSong: FYI, You don't need to add spaces between function names and parentheses.

Reflexe: It's my IDE that do it automatically. I'll update it.

PhysSong:
> I don't know if I should improve it more: I could add an abstraction layer for SampleBuffer and then represent it as list of buffers instead of one contagious buffer (that, to support sampleTCO rendering while recording).

Do you mean GUI rendering? If so, I guess you want to read recorded data from `SampleBuffer` while recording. Why do you want to represent that as list of buffers?

>However, I think it would require more work and may introduce bugs in other areas that assume SampleBuffer is a contagious area.

If you think so, let us search for better alternatives. I think introducing bugs is not too bad if we can't find any better ways. `master` is the development branch now, and we may fix bugs whenever before 1.3 releases.

Reflexe:
> Do you mean GUI rendering? If so, I guess you want to read recorded data from SampleBuffer while recording. Why do you want to represent that as list of buffers?
The SampleBuffer does not get created until the recording has been completed (that is way you don't see anything in the recorded TCO). Currently, the SampleBuffer, as opposed to the recording buffer is represented as a contiguous area of memory (while the recording buffer represented as a series of buffers).

> If you think so, let us search for better alternatives. I think introducing bugs is not too bad if we can't find any better ways. master is the development branch now, and we may fix bugs whenever before 1.3 releases.
I tried to look for alternatives and in the end, decided to ask here before I start implementing this.

musikBear:
One word WOW! @Reflexe 🎆 
I have one issue on my x32. The volume is significantly louder in the recorded clip, than in the original:

CMakeLists.txt
```diff
@@ -472,7 +472,7 @@ ELSE()
 ENDIF(WANT_DEBUG_FPE)
 
 # check for libsamplerate
-PKG_CHECK_MODULES(SAMPLERATE REQUIRED samplerate>=0.1.8)
+PKG_CHECK_MODULES(SAMPLERATE REQUIRED samplerate>=0.1.9)
```
PhysSong: FYI, Ubuntu 16.04 still ships with `libsamplerate` 0.1.8. To avoid the compile issue with old `libsamplerate`, you can use `const_cast` instead. I've looked into the code of the library, and it looks safe.

Reflexe: Oky doky :D I'll do that.

src/core/SampleBuffer.cpp
```diff
@@ -1403,8 +1380,7 @@ QString SampleBuffer::tryToMakeRelative( const QString & file )
 {
 	if( QFileInfo( file ).isRelative() == false )
 	{
-		// Normalize the path
-		QString f = QFileInfo( file ).canonicalFilePath().replace( QDir::separator(), '/' );
+		QString f = QString( file ).replace( QDir::separator(), '/' );
```
PhysSong: You should revert this change. FYI, this part is changed again in `stable-1.2` and will by synced back into `master`.

Reflexe: Thanks! I wasn't sure what I did wrong.  Done.

PhysSong:
A bug reported by @Sawuare: 
> Also, when recording in a TCO that has a file loaded, CPU usage gets very high and nothing is recorded.

I guess `SampleBuffer::m_audioFile` should be set to an empty `QString` before recording.

PhysSong:
One more minor issue: recorded samples are off by 1 sample because of `m_data.resize (1, {0,0});`. It could be worked around if needed, but it may be hard to make sure the workaround doesn't break anything.

src/core/SampleBuffer.cpp
```diff
@@ -144,32 +132,32 @@ void SampleBuffer::sampleRateChanged()
...
 	bool fileLoadError = false;
-	if( m_audioFile.isEmpty() && m_origData != NULL && m_origFrames > 0 )
+	if( m_audioFile.isEmpty() && m_origData.size () > 0 )
 	{
 		// TODO: reverse- and amplification-property is not covered
 		// by following code...
-		m_data = MM_ALLOC( sampleFrame, m_origFrames );
-		memcpy( m_data, m_origData, m_origFrames * BYTES_PER_FRAME );
+		auto previousFrames = frames ();
+		m_data = std::move(m_origData);
```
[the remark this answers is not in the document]
Reflexe: I removed `m_origFrames`. Do you mean `m_origData`?

src/core/Mixer.cpp
```diff
@@ -331,6 +330,12 @@ void Mixer::pushInputFrames( sampleFrame * _ab, const f_cnt_t _frames )
 	}
 
 	memcpy( &buf[ frames ], _ab, _frames * sizeof( sampleFrame ) );
+
+	if (!shouldApplyMasterGain) {
+		applyMasterGainToInputBuffer (&buf[ frames ], _frames, DEFAULT_CHANNELS,
```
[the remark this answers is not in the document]
Reflexe: I have no idea. I just reversed the actions order for the output :D Apparently there are some differences.

include/SamplePlayHandle.h
```diff
@@ -38,7 +38,7 @@ class AudioPort;
 class SamplePlayHandle : public PlayHandle
 {
 public:
-	SamplePlayHandle( SampleBuffer* sampleBuffer );
+	SamplePlayHandle(SampleBuffer* sampleBuffer , bool shouldCreateAudioPort=true);
```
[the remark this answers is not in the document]
Reflexe: About the formatting: what about just running clang-format on the project whole?

Reflexe:
> One more minor issue: recorded samples are off by 1 sample because of m_data.resize (1, {0,0});. It could be worked around if needed, but it may be hard to make sure the workaround doesn't break anything

I think it could also work with an empty sample buffer. It should work. I will check that.

src/core/SampleBuffer.cpp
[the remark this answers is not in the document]
PhysSong: Yes.

src/core/Mixer.cpp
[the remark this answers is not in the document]
PhysSong: Master gain have no effects on audio backends, so I think this should be dropped.
Another reason: if one records sample.when master volume is set to 0, NaN will be recorded.

include/SamplePlayHandle.h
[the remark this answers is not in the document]
PhysSong: We're planning to reformat entire project after releasing the last 1.2.x version.

PhysSong:
`PlayHandle` stuffs in `SampleTrack` are complex since #3133. I found some more issues:
- LMMS crashes or hangs when TCO being recorded is repeatedly resized around play head
In this case, multiple `SampleRecordHandle` instances try to add data to one TCO. It won't work as expected and may cause crash/hanging.

- Samples are recorded over the end of TCO
I don't know if that's desired, but I think it may be unclear for some users.

PhysSong:
>  Loops don't work more than once

When the playhead moves back, `Song::updateSampleTracks` signal is emitted. It's connected to `SampleTCO::playbackPositionChanged` and it calls `Mixer::removePlayHandlesOfTypes` with type `PlayHandle::TypeSamplePlayHandle`. `SampleRecordHandle` constructor sets play-handle type to `TypeSamplePlayHandle`, so recording handles are removed when jumping back to loop start point.

src/core/SamplePlayHandle.cpp
```diff
@@ -32,25 +32,26 @@
...
+	if (shouldCreateAudioPort)
+		setAudioPort( new AudioPort( "SamplePlayHandle", false ) );
 }
 
 
 
 
 SamplePlayHandle::SamplePlayHandle( const QString& sampleFile ) :
-	SamplePlayHandle( new SampleBuffer( sampleFile ) )
+	SamplePlayHandle( new SampleBuffer( sampleFile ) , false)
```
PhysSong: According tohttps://github.com/LMMS/lmms/pull/4310/files#diff-154de9db984f4e39bfeb6e62700ecc83L40, `true` should be passed.

src/core/SampleBuffer.cpp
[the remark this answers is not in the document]
Reflexe: I've ended up reworking all `SampleBuffer` and removing the "god function" `update()` and splitting it into multiple functions. The problem is: It is a lot of lines. Should I open another PR for this and block this pr?

src/core/SampleBuffer.cpp
```diff
@@ -1140,12 +1134,12 @@ QString & SampleBuffer::toBase64( QString & _dst ) const
 
 
 
-SampleBuffer * SampleBuffer::resample( const sample_rate_t _src_sr,
-						const sample_rate_t _dst_sr )
+std::unique_ptr<SampleBuffer> SampleBuffer::resample(const sample_rate_t _src_sr,
```
PhysSong: Why did you declare it as returning `SampleBuffer` in the header? Its instance can't be copied because it's a child of `QObject`.

Reflexe:
I've probably forgot to commit `SampleBuffer.h`.

src/core/SampleBuffer.cpp
```diff
@@ -214,54 +199,49 @@ void SampleBuffer::update( bool _keep_settings )
...
+		if ( dataFrames == 0 || fileLoadError )  // if still no frames, bail
 		{
 			// sample couldn't be decoded, create buffer containing
 			// one sample-frame
-			m_data = MM_ALLOC( sampleFrame, 1 );
-			memset( m_data, 0, sizeof( *m_data ) );
-			m_frames = 1;
+			m_data.clear ();
 			m_loopStartFrame = m_startFrame = 0;
 			m_loopEndFrame = m_endFrame = 1;
```
PhysSong: I think you should set to `m_endFrame` and `m_loopEndFrame` to 0 if the buffer is empty.

src/core/SampleBuffer.cpp
```diff
@@ -214,54 +199,49 @@ void SampleBuffer::update( bool _keep_settings )
...
 	else
 	{
-		// neither an audio-file nor a buffer to copy from, so create
-		// buffer containing one sample-frame
-		m_data = MM_ALLOC( sampleFrame, 1 );
-		memset( m_data, 0, sizeof( *m_data ) );
-		m_frames = 1;
+		m_data.clear ();
 		m_loopStartFrame = m_startFrame = 0;
 		m_loopEndFrame = m_endFrame = 1;
```
PhysSong: Here, too.

include/SampleBuffer.h
```diff
@@ -339,6 +333,25 @@ public slots:
...
+	 * @brief Do some actions necessary after changing m_data.
+	 * @param shoguldLock	The same value you've used on @a beginBufferChange.
+	 * @param shouldKeepSettings	Should we keep playback settings?
+	 * @param bufferSampleRate		The new m_data's sample rate.
+	 * @param shouldNotNotify		Should we not notify anyone about this change?
+	 */
+	void doneBufferChange (bool shouldLock,
+						   bool shouldKeepSettings,
+						   sample_rate_t bufferSampleRate,
+						   bool shouldLockMixer=true);
```
PhysSong: Would `shouldUnlock` and `shouldUnlockMixer` be better?

plugins/audio_file_processor/audio_file_processor.cpp
```diff
@@ -1063,11 +1069,7 @@ void AudioFileProcessorWaveView::updateGraph()
 		m_to = m_sampleBuffer.endFrame();
 	}
 
-	if( m_sampleBuffer.reversed() != m_reversed )
-	{
-		reverse();
-	}
-	else if( m_last_from == m_from && m_last_to == m_to && m_sampleBuffer.amplification() == m_last_amp )
+	if( m_last_from == m_from && m_last_to == m_to && m_sampleBuffer.amplification() == m_last_amp )
```
PhysSong: This won't work correctly even after fixing `reverseModelChanged` if the start point is at the very beginning and the end point is at the very end(`m_from` is 0 and `m_to` is equal to `m_sampleBuffer.frames()`).

src/core/SampleBuffer.cpp
```diff
@@ -293,123 +217,112 @@ void SampleBuffer::update( bool _keep_settings )
...
 	{
-		SampleBuffer * resampled = resample( _src_sr,
-					Engine::mixer()->baseSampleRate() );
-		MM_FREE( m_data );
-		m_frames = resampled->frames();
-		m_data = MM_ALLOC( sampleFrame, m_frames );
-		memcpy( m_data, resampled->data(), m_frames *
-							sizeof( sampleFrame ) );
-		delete resampled;
+		m_data = resampleData (m_data, sampleRate (), sampleRate ());
```
PhysSong: I think the second argument should be `_src_sr`.

Reflexe: good catch! thanks!

src/core/SampleBuffer.cpp
```diff
@@ -1472,3 +1106,93 @@ SampleBuffer::handleState::~handleState()
...
+	// First of all, don't let anyone read.
+	beginBufferChange (true, shouldLockMixer);
+	{
+		if (! newVector.empty()) {
+			// Insert to the end of the resampled vector.
+			m_data.reserve (m_data.size () + newVector.size ());
+			m_data.insert (m_data.end (), newVector.cbegin (), newVector.cend ());
+		} else {
+			// Insert to the end of the vector.
+			m_data.reserve (m_data.size () + (newVector.size ()));
```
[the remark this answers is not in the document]
Reflexe: I have now to write 100 times "Don't copy and paste code"

Reflexe:
- We still have a problem with SDL: It just have a lot of delay until frames are recorded. 
See https://stackoverflow.com/questions/42990071/recording-microphone-with-sdl2-gets-delayed-by-2-seconds
I'm not sure it is the same problem but I don't think it is in LMMS since it does work with other platforms. As long as AudioPort works good on windows, we don't have to worry about it.

- I will look into the problems with `reverse`. I don't think it should be too hard to implement. 
- About the loops: I've managed to solve the problem. I'll push it after cleaning and making sure it does work. 
- About thread-safety: if I understand correctly, `SampleBuffer::addData` could cause data races since it emits `sampleUpdated` on the mixer thread. Which could lead to very strange bugs. I'll try to find a way to work around it.  I still think that we have to somehow separate the gui and the main thread from the mixer.
- About high CPU when recording: I think that the cause it that we emits `sampleUpdated`, that causes all the TCO widget to repaint itself. We could solve it by simply not emitting it and using a QTimer that will repaint it once it 500 milliseconds (or something like that) while we record. 

After doing these things, this pr would be ready the mature enough to be merged. 

Please let me know what you think :)

src/core/SampleBuffer.cpp
```diff
@@ -1482,3 +1106,93 @@ SampleBuffer::handleState::~handleState()
...
+void SampleBuffer::resetData(DataVector &&newData, sample_rate_t dataSampleRate, bool shouldLockMixer) {
+	beginBufferChange (true, shouldLockMixer);
+	{
+		m_audioFile = QString();
+		m_data = std::move (newData);
+	}
+	doneBufferChange (true, /* lock */
+					  false, /* save settings */
+					  dataSampleRate,
+					  true);
```
PhysSong: `true` -> `shouldLockMixer`

plugins/audio_file_processor/audio_file_processor.cpp
```diff
@@ -343,7 +345,12 @@ void audioFileProcessor::setAudioFile( const QString & _audio_file,
 
 void audioFileProcessor::reverseModelChanged( void )
 {
-	m_sampleBuffer.setReversed( m_reverseModel.value() );
+	if (m_reverseModel.value () != m_isCurrentlyReversed) {
+		m_isCurrentlyReversed = m_reverseModel.value ();
+
+		m_sampleBuffer.reverse (true);
```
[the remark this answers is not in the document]
PhysSong: Now looks fine. :+1:

PhysSong:
@Reflexe Could you please update test binaries?

Reflexe:
@PhysSong Of-course! here you go: https://github.com/Reflexe/lmms/releases/tag/recording-2

src/core/SampleBuffer.cpp
```diff
@@ -293,123 +217,112 @@ void SampleBuffer::update( bool _keep_settings )
...
+	if( ( state = src_new( SRC_SINC_MEDIUM_QUALITY,
+					DEFAULT_CHANNELS, &error ) ) != NULL )
+	{
+		SRC_DATA src_data;
+		src_data.end_of_input = 1;
+		src_data.data_in = libSampleRateSrc(inputData.data ())->data ();
+		src_data.data_out = outputData.data ()->data();
+		src_data.input_frames = inputData.size ();
+		src_data.output_frames = dst_frames;
+		src_data.src_ratio = (double) inputSampleRate / requiredSampleRate;
```
PhysSong: This line leads to weird results when resampling. You should swap denominator and numerator.

src/core/SamplePlayHandle.cpp
```diff
@@ -139,7 +139,7 @@ bool SamplePlayHandle::isFromTrack( const Track * _track ) const
 
 f_cnt_t SamplePlayHandle::totalFrames() const
 {
-	return ( m_sampleBuffer->endFrame() - m_sampleBuffer->startFrame() ) * ( Engine::mixer()->processingSampleRate() / Engine::mixer()->baseSampleRate() );
+	return ( m_sampleBuffer->endFrame() - m_sampleBuffer->startFrame() ) * ( Engine::mixer()->processingSampleRate() / m_sampleBuffer->sampleLength ());
```
PhysSong: Shouldn't the denominator be `m_sampleBuffer->sampleRate()`?

PhysSong:
In current implementation, every `SampleBuffer` instance will resample its data when processing sample rate is changed. For recorded data, I wonder if repetitive resampling may affect audio quality.

Reflexe:
First of all, resampling cases are very rare. In most cases, the sample
rate would stay the same. Second, is there something else to do other than
re-sampling (how can we play not-resampled frames)? If we would not
resample that, the audio interface will do that for us.

PhysSong:
Suppose you have recorded samples in 192000Hz and exported it with 44100Hz. There must be irreversible quality loss.

Reflexe:
Oh. You're right. I'll implement that after work. It should not be hard
after the changes I've made to SampleBuffer.

We would need to call resample only when we play the samples.

And a question from someone who have no idea about music theory: is there
any restriction about resampling a part of a buffer? From my understanding,
there would be no problem, just want to make sure.

PhysSong:
> is there any restriction about resampling a part of a buffer?

AFAIK the only thing needed is some additional samples. We do that in `SampleBuffer::play`.

CMakeLists.txt
```diff
@@ -86,7 +86,7 @@ ENDIF(LMMS_BUILD_APPLE)
 
 IF(LMMS_BUILD_WIN32)
 	SET(WANT_ALSA OFF)
-	SET(WANT_JACK OFF)
+        SET(WANT_JACK ON)
```
PhysSong: Please use tabs.

Reflexe:
Ok, we so so close now. 

[ ] Finishing JACK on windows: Installing jack on windows. 
[ ] Finishing loops: Does not work if you're outside of the loop.
[ ] Bugs: Sometimes exporting a file will result an empty file.

After these tasks I'll freeze this PR to properly test and review it without me constantly changing every thing. 

[ ]Another task is to optimise the TCO drawing. Currently, recording for more than 2Mins~ would result very high cpu since `SampleTrack::visualize` does its work every time.

src/gui/EffectSelectDialog.cpp
```diff
@@ -31,6 +31,7 @@
 #include "PluginFactory.h"
 
 #include <QLabel>
+#include <QHeaderView>
```
PhysSong: Please drop this, it is already fixed in  #4374 and synced into `master`.

Reflexe: No problem. I was sure I was the cause somehow (since it just appeared from no where)

Reflexe:
I thunk it should be ready now. I will pull from master to make the CI happy.

PhysSong:
Are there any alternative way for enabling JACK than bundling JACK headers in our repository? It doesn't look good for me.

Reflexe:
@PhysSong  It does not look good for me too. The problem is that weak jack requires jack headers, which would require jack libs, and on win we don't have this package. 
I will try to maybe find another way. Installing the host's version of libjack could be possible.

src/core/SamplePlayHandle.cpp
```diff
@@ -139,7 +137,7 @@ bool SamplePlayHandle::isFromTrack( const Track * _track ) const
 
 f_cnt_t SamplePlayHandle::totalFrames() const
 {
-	return ( m_sampleBuffer->endFrame() - m_sampleBuffer->startFrame() ) * ( Engine::mixer()->processingSampleRate() / Engine::mixer()->baseSampleRate() );
+	return ( m_sampleBuffer->endFrame() - m_sampleBuffer->startFrame() ) * ( Engine::mixer()->processingSampleRate() / m_sampleBuffer->sampleRate ());
```
PhysSong: Seems like a long-standing bug, but the denominator should be:
```cpp
(float)Engine::mixer()->processingSampleRate() / m_sampleBuffer->sampleRate()
```
This will fix empty export result issue.

PhysSong:
Some issues:
- Visualization gets weird if sample track is resized before recording
- Existing visualization doesn't get cleared while recording
- High CPU usage when recording keeps for a long time
- Crash when moving TCOs while recording
- Occasional crash when fallback-TCO recording

I guess there might be some more, but here's what I found so far.

Reflexe:
1. Yes, there is a problem with the beginning offset. 
2.  Should not be hard to fix. 
3. Strange. I will recheck the cause. maybe actually rendering the points could be cpu consuming. I should cache it as `QPixmap` instead of `QPolygonF`.  
4 & 5. There seems to be a race condition between `beforeRecord` (GUI thread) and `play` (worker thread). I will try to use `BlockingQueuedConnection` to make sure we're not playing the track before it has been constructed.

Reflexe:
I've added a smarter implementation of caching. It should be better now for long recordings.

karmux:
When I open some drum sample or open project that contain drum samples, I often get an error "Audio files are limited to 300MB in size and 90 minutes of playing time" and sample fails to open. This happens with all flac and many wav samples but not with ogg samples. I have attached one sample file from Hydrogen drumkits that fails to load.

[dr110kik.zip](https://github.com/LMMS/lmms/files/2087574/dr110kik.zip)

Reflexe:
@karmux Thanks! it should be fixed now.

Reflexe:
Thanks! it also should be ok.

karmux:
Shrinked waveforms are fixed now but w2_invsinehalf.flac still has a furry tail that is flat in master and in Audacity.

Another bug: When sample track inside the pattern is played in the loop and loop is not in the beginning of the song then for first time sample track is silent and starts playing only when
a) playhead moves back to beginning of loop or
b) playhead starts playing sample again in a longer loop and pattern

Reflexe:
Let's say that samples track inside a B&B track is very broken. I don't think it has to do something with this PR.

karmux:
wav and flac samples play 1 octave higher than they should. ogg is fine.

plugins/patman/patman.cpp
```diff
@@ -381,11 +379,7 @@ patmanInstrument::LoadErrors patmanInstrument::loadPatch(
 
 void patmanInstrument::unloadCurrentPatch( void )
 {
-	while( !m_patchSamples.empty() )
-	{
-		sharedObject::unref( m_patchSamples.back() );
-		m_patchSamples.pop_back();
-	}
+	m_patchFile.clear ();
```
PhysSong: You mean `m_patchSamples.clear()`, right?

src/gui/widgets/Graph.cpp
```diff
@@ -582,7 +584,8 @@ void graphModel::setWaveToNoise()
 
 QString graphModel::setWaveToUser()
 {
-	SampleBuffer * sampleBuffer = new SampleBuffer;
+	auto sampleBuffer = std::unique_ptr<SampleBuffer>{new SampleBuffer};
```
PhysSong: I guess you can use `make_unique` from `stdshims.h` which was added in #4235.

PhysSong:
>  wav and flac mono samples play 1 octave higher in Sample Track and AudioFileProcessor.

I think `decodeSampleSF` should convert read data to stereo data. There are other ways, but it will increase complexity.
>  AudioFileProcessor doesn't show loaded sample file name.

`changeAudioFile` doesn't set `m_audioFile`, so it is set nowhere. Something like `m_audioFile = tryToMakeRelative(audioFile);` should work.
>  TripleOscillator preset previews show errors and playing doesn't stop.

Some preset files have `samples/empty.wav` and it doesn't exist. This pull request adds a warning for nonexistent files and it seems to cause the issue.
Here's the list of presets with `samples/empty.wav`:
```
data/presets/TripleOscillator/AmazingBubbles.xpf
data/presets/TripleOscillator/AnalogBell.xpf
data/presets/TripleOscillator/AnalogTimes.xpf
data/presets/TripleOscillator/Analogous.xpf
data/presets/TripleOscillator/Arpeggio.xpf
data/presets/TripleOscillator/ArpeggioPing.xpf
data/presets/TripleOscillator/Bell.xpf
data/presets/TripleOscillator/BellArp.xpf
data/presets/TripleOscillator/DistortedPMBass.xpf
data/presets/TripleOscillator/DullBell.xpf
data/presets/TripleOscillator/E-Organ.xpf
data/presets/TripleOscillator/Freaky-Bass.xpf
data/presets/TripleOscillator/FutureBass.xpf
data/presets/TripleOscillator/FuzzyAnalogBass.xpf
data/presets/TripleOscillator/Harmonium.xpf
data/presets/TripleOscillator/Harp-of-a-Fairy.xpf
data/presets/TripleOscillator/LFO-party.xpf
data/presets/TripleOscillator/LovelyDream.xpf
data/presets/TripleOscillator/MoogArpeggio.xpf
data/presets/TripleOscillator/Play-some-rock.xpf
data/presets/TripleOscillator/PluckArpeggio.xpf
data/presets/TripleOscillator/RaveBass.xpf
data/presets/TripleOscillator/ResonantPad.xpf
data/presets/TripleOscillator/SpaceBass.xpf
data/presets/TripleOscillator/Square.xpf
data/presets/TripleOscillator/TB303.xpf
data/presets/TripleOscillator/TheMaster.xpf
data/presets/TripleOscillator/Whistle.xpf
data/presets/TripleOscillator/Xylophon.xpf
```
> PatMan can't load any pat files.

https://github.com/LMMS/lmms/pull/3947#discussion_r200909827
> w2_invsinehalf.flac has a furry tail in BitInvader.

I guess it's due to resampling. I've dumped values in the tail, and I got values at the scale of 10^-7. Anyway, I think the real issue is that `Graph` uses truncation instead of rounding.

And another issue: `resampleData` tries to resample data even if `inputData` has no data.

src/CMakeLists.txt
```diff
@@ -244,6 +244,18 @@ IF(LMMS_BUILD_WIN32)
 				DESTINATION .)
 	ENDIF()
 
+
+	SET(JACK_INSTALLER_VERSION "1.9.11")
+	# Bundle jack
+	IF (WANT_JACK)
+		ADD_CUSTOM_TARGET(jack_binary_installer ALL
+			COMMAND "curl" -OL "https://github.com/jackaudio/jackaudio.github.com/releases/download/${JACK_INSTALLER_VERSION}/Jack_v${JACK_INSTALLER_VERSION}_32_setup.exe"
```
PhysSong: Instead of using `curl` command, you can use `file(DOWNLOAD ...)` command provided by CMake.

PhysSong:
BTW, which merge strategy should be used for this pull request?

- Ordinary merge will mess up commit history a lot.
- However, it's not a good idea to squash everything.
- We may try manual rebasing, but it will give a ton of conflicts and can be error-prone.

Reflexe:
@PhysSong I'm currently in the middle of a semester and a lot of work. I will get back to this PR when I'll have time. 
I think that we could just add something like "Enable sample track recording #3947" before every commit description. We could also try to re-commit the changes into a newer PR with a better commit history.

lukas-w:
I'll try to rebase this to fix merge conflicts and remove most of the fix-up/trial-and-error commits. I suggest we merge that rebased version using a normal merge commit.

lukas-w:
Rebased and pushed to branch `feature/recording`. I also fixed two bugs (10077c1f2f2f294e56a9d2dd098a8b7beca40db4, reported by @karmux & 2b9d3e26ee5d8dc648cd9f4b9e35b117efaeb1c1) and removed the bundled jack headers (4ed08fce299a8f238ce58bd49dc15d9407af2db3). Should I open a new PR from that branch? Do we want to add a "recording" [project](https://github.com/LMMS/lmms/projects) to track the remaining tasks?

Reflexe:
@lukas-w Great work! I'm so sorry I can't help more. What are the remaining tasks?

PhysSong:
Fixed some more bugs in [`feature/recording`](https://github.com/LMMS/lmms/tree/feature/recording). I also cherry-picked some changes in #4443 to fix MSVC builds.

Remaining bugs:
- TripleOscillator preset previews show errors due to nonexistent `samples/empty.wav`
- w2_invsinehalf.flac has a furry tail in BitInvader due to resampling(maybe an issue with `Graph`?)
- Visualization gets weird when resizing tracks while recording

Possible enhancements:
- Refactor
- Decide when to resample data
- Optimize
- etc.

Reflexe:
The last one has been solved (You just can't resize a track while recording). 
I will check the two other ones in the next few days.

PhysSong:
To be clear, I mean horizontal resizing(Shift + click and drag) of the track.

Reflexe:
Oh. You're right. That does crash lmms for me.

Reflexe:
I'm closing it for the a new feature/recording branch.

PR #3957, opened 2017-11-08
New attempt at fetching submodules

tresf:
The new technique parses the `.gitmodules` file and clones each submodule by path, unless the module was specified in `SKIP_SUBMODULES`.

To test, just clone this branch **without** the `--recursive` option and try to run cmake.

Advanced usage:

```cmake
# Clones everything except plugins/LadspaEffect/swh/swh
SET(SKIP_SUBMODULES swh)
INCLUDE(CheckSubmodules)
```
Or via CLI:

```bash
# Clones everything except zynaddsubfx and rpmalloc
cmake -DSKIP_SUBMODULES=zynaddsubfx;rpmalloc
```

Most of the time, users shouldn't even know this module exists.  It'll just be part of the codebase to improve productivity by reducing the number of build and configure errors directly related to submodule cloning.

* Eventually we may want to bind the `ADD_SUBDIRECTORY(...)` to be cognizant of the submodule status, but that is out of scope for this PR.
* Eventually we may want to check if a submodule is out of date and warn the user.  This may happen if a branch hasn't recently been pulled, or if switching to a branch with a different submodule commit.  This should be rare and is out of scope for this PR.

Once tested and merged to `master`, this PR should be safe to cherry-pick back to `stable-1.2`.

> (Editorial) 
>
> Submodules are annoying.
> 
> As the author of a few submodule PRs, I'm continuously forgetting `--recurse` or I'm switching branches without re-running `git submodule update --init --recursive`.  Sometimes cmake puts files in the directories and makes them non-clean so I have to clean them before running the commmands again.
> 
> To make things worse, I still haven't memorized if it's `init --update` or `update --init` (it's the latter, of course).
> 
> If I'm this frustrated after a week of working on submodules, newcomers will have an even harder time.
> 
> That said, submodules are great!  They allow us to pull directly from upstream without the mess of cluttering our codebase and risk not submitting patches to the right area.
> 
> So we have to live with them.  This is my second attempt to make that happen.  :) -Tres


src/CMakeLists.txt
```diff
@@ -160,10 +160,12 @@ TARGET_LINK_LIBRARIES(lmms
 )
 
 FOREACH(LIB ${LMMS_REQUIRED_LIBS})
-    GET_TARGET_PROPERTY(INCLUDE_DIRS ${LIB} INTERFACE_INCLUDE_DIRECTORIES)
-    if (INCLUDE_DIRS)
-        TARGET_INCLUDE_DIRECTORIES(lmmsobjs PRIVATE ${INCLUDE_DIRS})
-    ENDIF()
+	IF(TARGET ${LIB})
```
lukas-w: I already did this in 0dbdafc1f82ea0f6cb14a07bad4caab8e49643fc (albeit with inconsistent indentation...), it just hasn't been merged back to `master` yet.

tresf: Thanks.  I think I'll keep it here and we can deal with the minor merge conflict when rebasing.

tresf:
It's been 8 days without any testing feedback, but in that time, we've merged quite a bit of submodules making this needed more and more each day.  Calf, for example, is huge and the only way I've found to reduce that size is to reduce the size of the database using the `--depth` parameter, added in 8801ded.

I'm merging this to master, but there's a good chance it will need modifications and if we cherry-pick it to stable, we'll need to be cognizant of syncing them.

PR #3961, opened 2017-11-09
Update patterns in song editor after shifting their notes by semitones in piano roll.

Sawuare:
Fixes that shifting notes by semitones in piano roll doesn't update their pattern in song editor, by adding a call to `m_pattern->dataChanged()` in `PianoRoll::shiftSemiTone()`.

### DIFF
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -961,6 +961,7 @@ void PianoRoll::shiftSemiTone( int amount ) // shift notes by amount semitones
 		}
 	}
 
+	m_pattern->rearrangeAllNotes();
 	m_pattern->dataChanged();
 
 	// we modified the song
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -961,6 +961,8 @@ void PianoRoll::shiftSemiTone( int amount ) // shift notes by amount semitones
 		}
 	}
 
+	m_pattern->dataChanged();
+
 	// we modified the song
 	update();
 	gui->songEditor()->update();


src/gui/editors/PianoRoll.cpp
```diff
@@ -961,6 +961,8 @@ void PianoRoll::shiftSemiTone( int amount ) // shift notes by amount semitones
 		}
 	}
 
+	m_pattern->dataChanged();
```
zonkmachine: I think you need to call ```m_pattern->rearrangeAllNotes();``` before `dataChanged()` as you may have moved the note past another note so they need to be sorted again.

zonkmachine: You need to call ```m_pattern->rearrangeAllNotes();``` before `dataChanged()` as you may have moved the note past another note so they need to be sorted again.

Sawuare: If I understand it correctly, `m_pattern->rearrangeAllNotes()` is for sorting notes by start time, but the main function `shiftSemiTone` doesn't affect the notes' start time (horizontal position), it only affects the notes' keys (vertical position), so `m_pattern->rearrangeAllNotes()` isn't needed, right?

zonkmachine: Wrong. I changed this recently because the notes need to be sorted 'vertically' in order for the arpeggiator to work. Otherwise the order of execution will be randomly. Fixed here: https://github.com/LMMS/lmms/pull/3881

Sawuare: Done in https://github.com/LMMS/lmms/pull/3961/commits/31d32552964f80f0c303a47059f245d397cff287.

PR #3962, opened 2017-11-10
Switch opl2 to submodule

tresf:
Switch the "OpulenZ" (opl2) instrument plugin to pull directly from https://github.com/adplug/adplug.

There are some commits that have been added to the codebase that never made it upstream however upstream has been maintained as well, so some of them are redundant to apply.

One of my concerns is C++11 compatibility, ~~specifically the use of the `__inline` combined with Clang, which caused problems in the past.~~ **Edit:**, upstream removed `__inline`, may be OK.

Here's a full history of **our** (downstream) changes to the `fmopl.c` library.  These will help resolve issues that occur as a result of this merge.
https://github.com/LMMS/lmms/commits/stable-1.2/plugins/opl2/fmopl.c

**Tasks**
* [x] Switch opl2 to use subdirectory
* [x] Rename plugin directory to "OpulenZ" to match convention, write upgrade task.
* [x] Handle edge-case compilation issues

@softrabbit FYI.


tresf:
There doesn't appear to be any issues with the build, even on Apple.  Merging.

PR #3971, opened 2017-11-14
Patch exprtk on Clang

tresf:
Attempts to fix the Travis-CI environment by patching exprtk to workaround a bug with Clang.

More information about the bug is here: https://github.com/LMMS/lmms/pull/3965#issuecomment-344102990


lukas-w:
@tresf I get this error on the `fix/msvc` branch, even though it includes your patch: https://travis-ci.org/LMMS/lmms/jobs/305976591#L7078

> ```
> /Users/travis/build/LMMS/lmms/plugins/Xpressive/exprtk/exprtk.hpp:1965:64: error: default initialization of an object of const type 'const typename numeric::details::number_type<double>::type' (aka 'const exprtk::details::numeric::details::real_type_tag') without a user-provided default constructor
> 
>          const typename numeric::details::number_type<T>::type num_type;
> 
>                                                                ^`
> ```

Any idea what's going wrong here?

tresf:
Just a formal follow-up...

As discussed on Discord, it appears `Clang` is now `AppleClang` and the fix is to use `MATCHES` instead of the current `STREQUALS` in `plugins/Xpressive/CMakeLists.txt`.  This was likely caused by an XCode upgrade on the Travis-CI side.

PR #3982, opened 2017-11-18
Unsolo mixer channels on delete

zonkmachine:
If you remove a channel in the Fx-Mixer while it's soloed, the other channels are left muted. Clear channels before deleting and the remaining channels are restored to their pre-solo values.


zonkmachine:
I've seen no new issues from this PR. I suggest merging this and keeping an eye on #3983. It would be good if someone commented the code and also on the backtrace in #3983.

zonkmachine:
> I suggest merging this and keeping an eye on #3983.

That's a separate issue. I've tested this with the proposed fix in #3983 and they play well together.
Merge?

PR #3986, opened 2017-11-18
Automation Editor - delete automation point

zonkmachine:
Fix regression from b68dc572a3a58ad2187efe3ef5b4848ef7254abd where deleting a point by right clicking without dragging is defunct.

I misunderstood the algorithm. The selection for add/delete was dual.
```
   Left mouse button and above a point, move it.
   Right mouse button and below a point, delete it.
```
The latter appeared to me to be an obvious glitch so I pushed a fix to stable-1.2, messed up and was set for hours of fun with git and submodules. :skull_and_crossbones:  The final bug fixed in this PR removes the dual action altogether.

### DIFF
--- a/src/gui/editors/AutomationEditor.cpp
+++ b/src/gui/editors/AutomationEditor.cpp
@@ -523,7 +523,7 @@ void AutomationEditor::mousePressEvent( QMouseEvent* mouseEvent )
 					( it+1==time_map.end() ||
 						pos_ticks <= (it+1).key() ) &&
 		( pos_ticks<= it.key() + MidiTime::ticksPerTact() *4 / m_ppt ) &&
-					level == it.value() )
+		( level == it.value() || mouseEvent->button() == Qt::RightButton ) )
 				{
 					break;
 				}


zonkmachine:
Here is the essence of the changes from merged commits b68dc572a3a58ad2187efe3ef5b4848ef7254abd, 3de3ea6b16bb78a9559923c1b93dc7996535802b and this PR.

`$ git diff b68dc572a3a58ad2187efe3ef5b4848ef7254abd~1`
```diff
diff --git a/src/gui/editors/AutomationEditor.cpp b/src/gui/editors/AutomationEditor.cpp
index be3fb52..2b6dd2b 100644
--- a/src/gui/editors/AutomationEditor.cpp
+++ b/src/gui/editors/AutomationEditor.cpp
@@ -517,16 +517,14 @@ void AutomationEditor::mousePressEvent( QMouseEvent* mouseEvent )
                                        ( it+1==time_map.end() ||
                                                pos_ticks <= (it+1).key() ) &&
                ( pos_ticks<= it.key() + MidiTime::ticksPerTact() *4 / m_ppt ) &&
-                                       level <= it.value() )
+                                       ( level == it.value() || ( level != it.value() &&
+                                               mouseEvent->button() == Qt::RightButton ) ) )
                                {
                                        break;
                                }
```

I just rolled along with the original code. It feels a bit quirky but it works quite well actually. Trying to straighten it out a bit now but I doubt that I can do it.

src/gui/editors/AutomationEditor.cpp
```diff
@@ -523,7 +523,8 @@ void AutomationEditor::mousePressEvent( QMouseEvent* mouseEvent )
 					( it+1==time_map.end() ||
 						pos_ticks <= (it+1).key() ) &&
 		( pos_ticks<= it.key() + MidiTime::ticksPerTact() *4 / m_ppt ) &&
-					level == it.value() )
+					( level == it.value() || ( level != it.value() &&
+						mouseEvent->button() == Qt::RightButton ) ) )
```
PhysSong: It is equivalent to this: 
```c++
( level == it.value() || mouseEvent->button() == Qt::RightButton ) )
```

zonkmachine: Fixed.

zonkmachine:
The action on delete is not as spot on as for left click to move. It's the same as before though. Fixing this is going to be a lot messier. I think this is good for now.

zonkmachine:
Merge?

PR #3997, opened 2017-11-21
Automation Editor - Straighten out draw line function

### LINKED ISSUE #3995: Automation Editor - Draw line not aligned
drawline

Using the `<Shift>` modifier draws one point at the same level at the beginning and/or the end of the line.

zonkmachine:
Nudge the line (biggest improvement) and quantize x arguments to `AutomationEditor::drawLine(..)`.

Fixes  #3995

### DIFF
--- a/src/gui/editors/AutomationEditor.cpp
+++ b/src/gui/editors/AutomationEditor.cpp
@@ -434,9 +434,11 @@ void AutomationEditor::leaveEvent(QEvent * e )
 }
 
 
-void AutomationEditor::drawLine( int x0, float y0, int x1, float y1 )
+void AutomationEditor::drawLine( int x0In, float y0, int x1In, float y1 )
 {
-	int deltax = qRound( qAbs<float>( x1 - x0 ) );
+	int x0 = Note::quantized( x0In, AutomationPattern::quantization() );
+	int x1 = Note::quantized( x1In, AutomationPattern::quantization() );
+	int deltax = qAbs( x1 - x0 );
 	float deltay = qAbs<float>( y1 - y0 );
 	int x = x0;
 	float y = y0;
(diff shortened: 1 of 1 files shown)


zonkmachine:
This isn't one issue only. I've fixed up `AutomationEditor::drawLine(...)` but there is still things to do.
In `AutomationEditor::mousePressEvent(...)`, if the shift button is held and the left mouse button is pressed, we first draw a line and then carry on with the final point and create a drag value as the last point is selected for dragging. The magic happens [here](https://github.com/LMMS/lmms/blob/55076d0bb05be9144dc0ca2bef892c51621a8e39/src/gui/editors/AutomationEditor.cpp#L544:L570). Many times it's the last or next to last point that is off so I think this is related to a rounding off error and/or quantization in the above mentioned functions and in `AutomationPattern::setDragValue(...)`. If you draw your points close to their quantized values there is little risk of a draw error. If you draw a line with steep angles between two points and aim for slightly to the left of the left point and to the right of the right one (picture), you will get a draw error. The closest points on both sides will snap to the same level. Setting `m_drawLastTick = new_time;` after [it = time_map.find( new_time );](https://github.com/LMMS/lmms/blob/55076d0bb05be9144dc0ca2bef892c51621a8e39/src/gui/editors/AutomationEditor.cpp#L569) avoids this but just pushes the error around.

drawerror


I think it's out of scope for 1.2 to fix this and suggest that we merge this and leave #3995 open.

zonkmachine:
I think it's out of scope for 1.2 to fix this and suggest that we merge #3997 and leave #3995 open.

zonkmachine:
Quantize `drawline()` input. The line is drawn from a mix of unquantized `x0, x1` input values and the quantized `deltax`, and will sometimes not draw correct.

zonkmachine:
Fixed. I'll merge this tomorrow if there are no objections.

Issue #3989, opened 2017-11-19
Some patterns in "Song-Editor" and "Beat+Bassline editor" are displayed larger than they are


vlad0337187:
Hello.

Some patterns in RC4.96 are displayed larger than they are in Song-Editor.
In Piano roll they are displayed normally, also they are played normally.

Here is example:
![screenshot from 2017-11-19 20-02-34](https://user-images.githubusercontent.com/12682937/32993773-6d7e4e40-cd65-11e7-8660-fcc7deb18b6e.png)

Example with beat: https://github.com/LMMS/lmms/issues/3989#issuecomment-345537384

Why is it important ?
Because now it's very annoying resizing all placed beats / patterns in Song-Editor. If not to resize them - they overlay one extra button and to paste new beat / pattern there you need to paste them one button right, than move it.

Hope, it's not big problem and could be corrected to 1.2.0 release.

Best regards,
Vladislav

vlad1777d:
Example with beat:

zonkmachine:
I don't see this on my machine. If you open one of the patterns in the Piano Roll and zoom to 800% you may find that the last note is extending into the following pattern. If not, please upload a project that shows this issue.

tresf:
Sounds similar to https://github.com/LMMS/lmms/issues/2660

Mark-Agent003:
I've had this happen before also, although I could never figure out why.

vlad1777d:
Yes, @zonkmachine is right. But I don't understand why was it happened.
I never changed quanting size of notes - it's 1 / 16, so it must be changed only in such diapason, never making that extra little size.
https://user-images.githubusercontent.com/12682937/32996167-eef8609c-cd87-11e7-908c-98570a115114.png

zonkmachine:
@vlad1777d 

> I didn't thought about this because I didn't changed notes quanting size.

Well. You could have slipped and made a `<Shift> + <Alt>` stretch which overrides the selected note length.
Then again, it could always be a bug in there... :)

Did you import this from 1.1.3 ?

vlad1777d:
@zonkmachine , no, I started making beat from scratch.
Thanks for a tip. Now beats and patterns are needed width =)

zonkmachine:
I think I know what happened here.

To replicate:
1. Add a note, 1/16 is a good length.
2. Hover over the right edge, press the left mouse button and drag it to the far left.
3. Let go of the mouse button.
[...]

PR #3998, opened 2017-11-22
Limit note length to quantization value

zonkmachine:
Draging a note to it's minimum value of 1 will add this new length to
the note if you later choose to stretch it which will not be clearly
visible in the Piano Roll unless you zoom in a bit. Limit the note
length to the quantization value and use <shift> key to set a smaller
value.

PR explained in depth here: https://github.com/LMMS/lmms/issues/3989#issuecomment-626233964

Fixes https://github.com/LMMS/lmms/issues/3989

### DIFF
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -2811,7 +2811,7 @@ void PianoRoll::dragNotes( int x, int y, bool alt, bool shift, bool ctrl )
 				if (note->selected())
 				{
 					int newLength = note->oldLength() + off_ticks;
-					newLength = qMax(alt ? 1 : quantization(), newLength);
+					if (newLength <= 0) { newLength = alt ? 1 : quantization(); }
 					note->setLength(MidiTime(newLength));
 
 					m_lenOfNewNotes = note->length();
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -2810,8 +2810,10 @@ void PianoRoll::dragNotes( int x, int y, bool alt, bool shift, bool ctrl )
 			{
 				if (note->selected())
 				{
-					int newLength = note->oldLength() + off_ticks;
-					if (newLength <= 0) { newLength = alt ? 1 : quantization(); }
+					int oldLength = note->oldLength();
+					int newLength = oldLength + off_ticks;
+					int minLength = qMin(oldLength,quantization());
+					newLength = qMax(alt ? 1 : minLength, newLength);
 					note->setLength(MidiTime(newLength));
 
 					m_lenOfNewNotes = note->length();
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -2811,8 +2811,8 @@ void PianoRoll::dragNotes( int x, int y, bool alt, bool shift, bool ctrl )
 				if (note->selected())
 				{
 					int newLength = note->oldLength() + off_ticks;
-					newLength = qMax(1, newLength);
-					note->setLength( MidiTime(newLength) );
+					newLength = qMax(alt ? 1 : quantization(), newLength);
+					note->setLength(MidiTime(newLength));
 
 					m_lenOfNewNotes = note->length();
 				}


zonkmachine:
Closing this PR for now. It's not clear to me how to improve this and it's not important for 1.2 anyway.

Spekular:
Isn't this made reduntant by #4357?

We also have a quantize button in piano roll. Isn't it better to fix that so it affects lengths as well?

I ask because this reminds me of tresf asking for fine offsets to be removed when clips are moved (until he changed his mind). If I want to extend the length of a weird-length note by 1/4, it's much easier to do that at 1/4 Q than to do it by hand at 1/64th Q.

zonkmachine:
> Isn't this made reduntant by #4357?

Maybe, I'll look into it.



> We also have a quantize button in piano roll. Isn't it better to fix that so it affects lengths as well?

Again maybe, but the issue here are tiny stretches that users aren't aware of that causes the tco to come out the wrong length. If you don't see the note length as being the culprit you're not going to try and change it neither by stretching nor by quantization.

Spekular:
Sorry, it seems I misunderstood what the PR did. I read "limit note length to quantization value" as "do not allow resized notes to be sizes other than n * Q, where Q is the quantization level, n is an integer". I should have looked at the detailed explanation/code instead.

Setting the minimum length to the current Q is better, but still overkill IMO. I think we should only prevent resizing down if the resize would lead to a length of zero. That way a 3/8ths note can be resized to 1/8th even if you're at 1/4th Q, which is a totally valid operation IMO.

src/gui/editors/PianoRoll.cpp
```diff
@@ -2811,8 +2811,8 @@ void PianoRoll::dragNotes( int x, int y, bool alt, bool shift, bool ctrl )
 				if (note->selected())
 				{
 					int newLength = note->oldLength() + off_ticks;
-					newLength = qMax(1, newLength);
-					note->setLength( MidiTime(newLength) );
+					newLength = qMax(alt ? 1 : quantization(), newLength);
```
Spekular, line 2814: Implementation for my preferred behavior, since code might be easier to interpret than my explanation 😅 
```suggestion
					if (newLength <= 0) { newLength = quantization(); }
```
Or if you prefer
```suggestion
					newLength = (newLength > 0) ? newLength : quantization();
```

zonkmachine: That's elegant! I prefer the <Alt> key in there somewhere. You can skip quantization when changing the length of the note with the <Alt> key so I would prefer:

```cpp
if (newLength <= 0){ newLength = alt ? 1 : quantization(); }
```

If you change that I'll commit it.

Spekular: Like so? :)
```suggestion
					if (newLength <= 0) { newLength = alt ? 1 : quantization(); }
```

zonkmachine: Yup!

zonkmachine:
It has some quirks though. Select many different notes that start at different times and change their length. It looks a bit odd when you shorten them and they take turn to first become shorter and then jump up again and become the quantized length.

Spekular:
Oh right, you could end up scaling 3/8->1/8->2/8...

I didn't think it through enough huh 🙃

src/gui/editors/PianoRoll.cpp
Veratil: Reminder that `newLength = alt` is not a boolean operation, but an assignment one. 😉

Spekular:
I wonder if using ``note->oldLength() % quantization()`` as a minimum instead of ``quantization()`` would work, or if oldLength is updated too frequently. That could also become zero, so then you'd need two checks for zero or something clever.

src/gui/editors/PianoRoll.cpp
zonkmachine: Yes. We're assigning either the quantized value or if overriding with Alt key, 1. Then you can still set the note to any length if you want.

Veratil: Okay! Wanted to make sure since tertiary operator can be confusing sometimes.

src/gui/editors/PianoRoll.cpp
```diff
@@ -2811,8 +2811,8 @@ void PianoRoll::dragNotes( int x, int y, bool alt, bool shift, bool ctrl )
 				if (note->selected())
 				{
 					int newLength = note->oldLength() + off_ticks;
-					newLength = qMax(1, newLength);
-					note->setLength( MidiTime(newLength) );
+					if (newLength <= 0) { newLength = alt ? 1 : quantization(); }
+					note->setLength(MidiTime(newLength));
```
Spekular, line 2815: This works the way I wanted, judging from an admittedly quick test.
```suggestion
					int minLength = alt ? 1 : (note->oldLength() % quantization());
					if (minLength == 0) { minLength = quantization(); }
					note->setLength(MidiTime(max(newLength, minLength)));
```

Would love to figure out a shorter way to get 
`x % y` if `x % y > 0`
`y` otherwise

zonkmachine: It works nice but if I set the length of notes of varying length I would want them to end up at the same min value. Example two notes starting at the same time. Lenths, 2/16 and 3/16. Set quantization to 1/8, hit Ctrl + A and set their length to min by dragging. Not the same value.

Spekular: In that case I think we need to iterate over the selction once before getting here to find the smallest minLength of any selected note, then use that for all of them.

zonkmachine:
Back to my original commit. I'm experimenting with letting notes shorter than quantization be remembered and kept as minimum value().
```cpp
int oldLength = note->oldLength();
int newLength = oldLength + off_ticks;
int minLength = qMin(oldLength,quantization());
newLength = qMax(alt ? 1 : minLength, newLength);
...
```

src/gui/editors/PianoRoll.cpp
zonkmachine: Had to test my latest research so PR is for now back at my branch.

zonkmachine:
New idea. Maybe just limit to the shortest note that isn't just a tick? It'll still give a short extra length if you drag it out with quantization but one that is more easy to spot.

```cpp
newLength = qMax(MidiTime::ticksPerTact()/64, newLength);
```

In following order.
A whole note with a 64th note
A whole note with a tick
A whole note

64th

Spekular:
I'm not a fan of this version, it doesn't allow 3/4 -> 1/4 @ 1/2 Q.

Spekular:
Here's a branch that meets both my criteria and yours:
- Notes can be resized to `length % quantization`, or `quantization` if the previous math leads to zero
- All notes can be resized to be as small as the shortest note's minimum length

https://github.com/LMMS/lmms/compare/master...Spekular:noteLen

zonkmachine:
Good solution. I've played around with it for quite a while and can't seem to find any obvious weak spots. I think you should issue a separate PR for your solution and I'll close this one.

On another but related note. If you have a group of notes with beat notes among them, select them all, and proceeds to extend them by dragging, the beat note won't react until after a whole measure has been covered. It looks like it's the full -192 ticks that has been covered first before there is a visible change. If I manually change the beat note length to -10 in the mmp file it will just stop for a brief moment before being extended. I don't know what a preferable behavior would be here.

Spekular:
> I think you should issue a separate PR for your solution

Done!

> On another but related note. If you have a group of notes with beat notes among them, select them all, and proceeds to extend them by dragging, the beat note won't react until after a whole measure has been covered.

Confirmed, but just to clarify this doesn't seem to be a regression as it occurs in my master build and 1.2.1 as well. New issue for that problem once this one is closed?

zonkmachine:
> Confirmed, but just to clarify this doesn't seem to be a regression as it occurs in my master build and 1.2.1 as well.

Yes, this is not new. It's there on 1.1.3. already.

> New issue for that problem once this one is closed?

Yes.

zonkmachine:
Closed in favor of https://github.com/LMMS/lmms/pull/5512

zonkmachine:
Adding this fixes the drag issue I think. At least it demos the issue.
```cpp
if (note->selected())
{
	if(note->oldLength() <= 0){note->setOldLength(1);}
	int newLength = qMax(minLength, note->oldLength() + off_ticks);
	note->setLength(MidiTime(newLength));

	m_lenOfNewNotes = note->length();
}
```

Issue #3999, opened 2017-11-22
MSVC support


lukas-w:
Creating this issue to keep track of tasks required to make LMMS build compatible with MSVC (related to #3492). I'll try to keep this list up to date:

- [ ] LMMS core code & CMake: I have most of the necessary fixes locally, will open a PR soon
- [ ] Plugins that fail building:
  - [ ] LadspaEffect
  - [ ] Monstro
  - [ ] Organic
  - [ ] ReverbSC
  - [ ] SID
  - [ ] VST
  - [ ] Vibed
  - [ ] Xpressive
  - [ ] ZynAddSubFX 

Because Windows provides no package management that we can use for grabbing dependencies, we'll need to find another way of making the overall build process easier. I suggest using [Hunter](https://github.com/ruslo/hunter) for fetching dependencies. Hunter is a package manager written using CMake, so if we leverage that, Windows devs will only need MSVC and CMake to get started. Unfortunately, most of LMMS's dependencies are still missing from Hunter, so we'll need to submit packages upstream. 

Dependencies include (not a complete list):

- [ ] For LMMS core:
  - [x] Qt
  - [ ] FFTW3: Fourier Transformation library
    * Website: http://fftw.org/
    * GitHub: [FFTW/fftw3](https://github.com/FFTW/fftw3)
    * Hunter issue: https://github.com/ruslo/hunter/issues/1160
  - [ ] libsamplerate: Used for resampling
    * Website: http://www.mega-nerd.com/SRC/
    * GitHub: [erikd/libsamplerate](https://github.com/erikd/libsamplerate)
  - [ ] libsndfile: Used for decoding and encoding audio files
    * Website: http://www.mega-nerd.com/libsndfile/
    * GitHub: [erikd/libsndfile](https://github.com/erikd/libsndfile/)
    * Hunter issue: https://github.com/ruslo/hunter/issues/1218
  - [ ] Audio backends:
    - [x] SDL2
    - [ ] libsoundio:
      * Website: http://libsound.io/
      * GitHub: [andrewrk/libsoundio](https://github.com/andrewrk/libsoundio)
    - [ ] PortAudio
      * Website: http://www.portaudio.com/
    - [ ] Jack?
- [ ] For plugins:
  - [ ] ZynAddSubFx
    - [ ] FLTK:
      * Website: http://www.fltk.org/index.php
  - [ ] Carla:
    * GitHub: [falkTX/Carla](https://github.com/falkTX/Carla/)
[...]

PR #4000, opened 2017-11-22
MSVC fixes

lukas-w:
Some fixes for MSVC support. In progress, things left to do:
- [x] Fix Linux/macOS build
- [x] Resolve merge conficts

This changes the minimum CMake version to 3.1.

Summary of the changes that were required to build with MSVC:

* CMake:
  * A lot of compiler flags that are not understood by MSVC
  * If MSVC is detected, disable plugins that are not yet supported
* Variable-Length-Arrays are not supported by MSVC because it's not C99-compliant. Replaced by `QVarLengthArray`
* Some `EXPORT` problems
* Some GCC- or POSIX-specific code: 
  * `MicroTimer` was using `sys/time.h`, ported to standard C++11 [`std::chrono::steady_clock`](http://en.cppreference.com/w/cpp/chrono/steady_clock).
  * `RemotePlugin` was using GCC built-in atomic functions, ported to standard C++11 [`std::atomic`](http://en.cppreference.com/w/cpp/atomic)
* One occurence of inline assembly. MSVC doesn't support the standard [`asm`](http://en.cppreference.com/w/cpp/language/asm) syntax, it has [`__asm`](https://msdn.microsoft.com/en-gb/library/45yd4tzz.aspx) instead, which in turn doesn't work with x64. Replaced `asm("pause")` by the [`_mm_pause()`](https://software.intel.com/en-us/node/524249) intrinsic

Addresses #3999.


Umcaruje:
What a nice addition to be out 4000th issue/pull request :tada:

lukas-w:
Some change introduced on `master` causes an assertion to fail on Windows now whenever I play a note:

```
ASSERT failure in QCoreApplication::sendEvent: "Cannot send events to objects owned by a different thread. Current thread 5c86fd0. Receiver '' (of type 'TrackContentObject') was created in thread 37b220", file kernel\qcoreapplication.cpp, line 541
QObject::~QObject: Timers cannot be stopped from another thread
```

lukas-w:
After recompiling about 5 times, I suddenly get a proper backtrace:

```
1   qt_message_fatal                                qlogging.cpp          1681 0x6669a78b  
2   QMessageLogger::fatal                           qlogging.cpp          795  0x66698fb2  
3   qt_assert_x                                     qglobal.cpp           3078 0x6668c6e1  
4   QCoreApplicationPrivate::checkReceiverThread    qcoreapplication.cpp  533  0x66a6e9be  
5   QApplication::notify                            qapplication.cpp      2970 0x558d1974  
6   QCoreApplication::notifyInternal2               qcoreapplication.cpp  988  0x66a6cd16  
7   QCoreApplication::sendEvent                     qcoreapplication.h    231  0x66a77ca2  
8   QObjectPrivate::setParent_helper                qobject.cpp           1996 0x66acad40  
9   QObject::~QObject                               qobject.cpp           1048 0x66ac4d4a  
10  Model::~Model                                   Model.cpp             30   0x1400bc03a 
11  AutomatableModel::~AutomatableModel             AutomatableModel.cpp  84   0x14001a91b 
12  BoolModel::~BoolModel                           lmms                       0x140005338 
13  TrackContentObject::~TrackContentObject         Track.cpp             130  0x140100ddf 
14  AutomationPattern::~AutomationPattern           AutomationPattern.cpp 106  0x1400230c6 
15  AutomationPattern::`vector deleting destructor' lmms                       0x14002ae68 
16  InlineAutomation::~InlineAutomation             InlineAutomation.h    46   0x1400bdfb7 
17  DetuningHelper::~DetuningHelper                 DetuningHelper.h      44   0x1400bdefa 
18  DetuningHelper::`scalar deleting destructor'    lmms                       0x1400be25c 
19  sharedObject::unref<DetuningHelper>             shared_object.h       64   0x1400bdc42 
20  Note::~Note                                     Note.cpp              89   0x1400bc710 
... <More>                                                                                 
```

Looks like some object is in a different thread than its parent, which is disallowed by Qt.

lukas-w:
I took a look at Qt's source code and found that the reason this only surfaces on Windows for me is that the `checkReceiverThread` is only called when Qt is built in debug mode, which is not the case with packages provided by Linux distros. So this bug is not specific to Windows and therefore not related to this PR. Merging.

PR #1800, opened 2015-03-01
Enabled the use of sample exact controls in Dual Filter

curlymorphic:
The following controls have sample exact enabled

Filter 1 Cutoff
Filter 1 res
Filter 1 gain

Filter 2 Cutoff
Filter 2 res
Filter 2 Gain

Mix

Model.isValueChanged() does not eveluate when recieving sample accurate
so added checks to see if the cutoff and res needed to be recalculated,

PR #4002, opened 2017-11-22
Fix CMake warning

### LINKED ISSUE #3985: CMake warning
Hi,

When using:

```sh
cmake .. -DCMAKE_INSTALL_PREFIX=../target/ -DWANT_QT5=ON
```

I got:

```txt
[...]
CMake Deprecation Warning at CMakeLists.txt:11 (CMAKE_POLICY):
  The OLD behavior for policy CMP0026 will be removed from a future version
  of CMake.

  The cmake-policies(7) manual explains that the OLD behaviors of all
  policies are deprecated and that a policy should be set to OLD only under
  specific short-term circumstances.  Projects should be ported to the NEW
  behavior and not rely on setting a policy to OLD.
[...]
```

zapashcanon:
Hi,

Fix #3985.


CMakeLists.txt
```diff
@@ -8,7 +8,7 @@ IF(COMMAND CMAKE_POLICY)
 	CMAKE_POLICY(SET CMP0005 NEW)
 	CMAKE_POLICY(SET CMP0003 NEW)
 	IF (CMAKE_MAJOR_VERSION GREATER 2)
-		CMAKE_POLICY(SET CMP0026 OLD)
+		CMAKE_POLICY(SET CMP0026 NEW)
```
tresf: Can we just remove it entirely?

zapashcanon: No idea... maybe @lukas-w can tell. I don't know anything about cmake, I just opened this PR because in #3985 @lukas-w said it was safe to set it to `NEW` while remove the warning.

tresf:
The documentation states:

> CMake version 3.0.2 warns when the policy is not set and uses OLD behavior.

So until we mandate 3.0.3, we need to specify it.  Merging.

PhysSong:
https://github.com/LMMS/lmms/blob/55076d0bb05be9144dc0ca2bef892c51621a8e39/CMakeLists.txt#L503

CMP0026 is set to `old` because of this. And this was dropped in `master` via #1800.
Since LMMS requires `CMake>=2.8.7`, we may replace that to a syntax introduced in CMake 2.8 safely(on `stable-1.2`).

Issue #2934, opened 2016-07-23
LMMS is kicked by JACK when exporting


Umcaruje:
Freshly compiled master branch, elementary OS Freya

```
Version 1.1.90-gfa19f00 (Linux/x86_64, Qt 4.8.6, GCC 4.8.4)
```

Steps to reproduce:
1. Load the `Skiessi - Onion` demo project.
2. Press the export button
3. An error message about LMMS getting kicked by jack appeares: 
   ![screenshot from 2016-07-24 01 42 09](https://cloud.githubusercontent.com/assets/6282045/17080843/e7c097a6-513f-11e6-8083-7430c1b3ea8e.png)
4. When you click the export button, it gets to around 8% and then it crashes.

Backtrace: 

```
(gdb) r
Starting program: /home/umcaruje/github/lmms/build/lmms 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Gtk-Message: Failed to load module "pantheon-filechooser-module"
[New Thread 0x7fffdefaa700 (LWP 5789)]
[New Thread 0x7fffde7a9700 (LWP 5790)]
[New Thread 0x7fffddfa8700 (LWP 5791)]
[New Thread 0x7fffdd7a7700 (LWP 5792)]
[Thread 0x7fffdd7a7700 (LWP 5792) exited]
[New Thread 0x7fffdd7a7700 (LWP 5794)]
[New Thread 0x7fffc76e2700 (LWP 5795)]
[New Thread 0x7fffc6ee1700 (LWP 5796)]
VST sync support disabled in your configuration
[New Thread 0x7fffdc0bd700 (LWP 5797)]
[New Thread 0x7fffbc1e2700 (LWP 5798)]
[New Thread 0x7fff864e4700 (LWP 5799)]
[New Thread 0x7fff85ce3700 (LWP 5800)]
[New Thread 0x7fffbc161700 (LWP 5801)]
[Thread 0x7fffddfa8700 (LWP 5791) exited]
[New Thread 0x7fffddfa8700 (LWP 5809)]
SuspendRefNum error
JackClient::Execute error name = lmms
[Thread 0x7fffbc161700 (LWP 5801) exited]
[New Thread 0x7fff84946700 (LWP 5810)]
Jack server was closed but clients are still allocated, cleanup...
Cleanup client ref = 6
Server is not running
[Thread 0x7fffbc1e2700 (LWP 5798) exited]
[...]

Issue #2845, opened 2016-06-13
Exporting audio is broken


unfa:
I'm using version 1.1.3-gcbf6d24 (Linux/x86_64, Qt 4.8.6, GCC 4.8.4) (built from git a few days ago).

Exporting audio produces unusable files - they are sped up, with chopped audio - sounds like a percentage of samples from the audio stream are discarded somehow. Lots of clicks and bad stuff present.

While the rendering is being done, a sped up and messed up version of the track is played back. Sound like JACK freewheeling mode messed up in such a way that the samples computed are played back. But JACK doesn't produce messed up files when this breaks.

Anybody can confirm this issue?

mikobuntu:
Confirmed!!        Linux ubuntu-16.04 64 bit ..... LMMS Master branch from yesterday I think 
Jack and LMMS both at 44100hz
I don't normally use LMMS through Jack as it has/had a history of disconnecting continuously.
@unfa Have you tried with multiple configurations in the LMMS Export settings ? I will test this a bit further when I get more time .

unfa:
Whoa! LMMS crashed now during the export using JACK backend.

What's even weirder - I've seen signal in the mixer channels that are not used in the project. 8O

Exporting with ALSA backend seems to work fine.

unfa:
Okay, new details.

JACK backend produces unusable files, but when using ALSA playback won't work after rendering - I had to restart LMMS to continue work.

PhysSong:
> but when using ALSA playback won't work after rendering

@unfa Can you reproduce this with latest `stable-1.2`?

>JACK backend produces unusable files

JACK issue would be fixed in #4005.

zonkmachine:
No answer, closing. This would have been fixed in #4005.

PR #4005, opened 2017-11-23
Fix various bugs when using JACK

### LINKED ISSUE #3207: LMMS Segfault with JACK when closing
I am building from source. My Linux distro is Fedora 23. JACK version is 1.9.10. 

I've found that when I simply open LMMS and then close it again, it causes a segfault.

i've attached a gdb backtrace of the problem here : [out.txt](https://github.com/LMMS/lmms/files/683936/out.txt)

Any ideas?

### LINKED ISSUE #3396: LMMS crackles when rendering to file
As the issue says, I get crackles when I render anything to file.

The test project I am working with is a default instance of ~~Organic~~ TripleOscillator playing a note for two bars. When I render it with the default options, I get crackles. The resulting audio file and project are zipped below:

[one_note.zip](https://github.com/LMMS/lmms/files/813006/one_note.zip)

Weird thing is, I only get this sort of error on my Fedora E6530 laptop. On my Chromebox running Slackware, things run fine. 

I should also note that when I try to render anything using my [Sporth plugin](http://www.github.com/paulbatchelor/LMMS-Sporth.git), things flat out crash on my laptop.

### LINKED ISSUE #3397: LMMS freezes when you try to render audio twice
Sorry folks, I found another bug which I can reproduce. LMMS really doesn't like this laptop of mine!

When I bounce an audio file, then go to bounce it again, it freezes. The GDB backtrace would seem to suggest that it's some sort of threading race condition.

[gdb.txt](https://github.com/LMMS/lmms/files/814501/gdb.txt)

On a related but unrelated note, is there someone I can talk to off github whose knowledgeable about the part s of the codebase in charge of offline rendering? My email is thisispaulbatchelor at gmail dot com.

PhysSong:
Change the way AudioJack handles processing stop. Fixes #3397, fixes #3396, fixes #3207.

### DIFF
--- a/include/AudioJack.h
+++ b/include/AudioJack.h
@@ -37,6 +37,7 @@
 #include <QtCore/QVector>
 #include <QtCore/QList>
 #include <QtCore/QMap>
+#include <QMutexLocker>
 
 #include "AudioDevice.h"
 #include "AudioDeviceSetupWidget.h"
@@ -107,6 +108,7 @@ private:
 
 	bool m_active;
 	bool m_stopped;
+	QMutex m_processingMutex;
 
 	MidiJack *m_midiClient;
 	QVector<jack_port_t *> m_outputPorts;
--- a/src/core/audio/AudioJack.cpp
+++ b/src/core/audio/AudioJack.cpp
@@ -201,6 +201,7 @@ bool AudioJack::initJackClient()
 
 void AudioJack::startProcessing()
 {
+	QMutexLocker m( &m_processingMutex );
 	m_stopped = false;
 
 	if( m_active || m_client == NULL )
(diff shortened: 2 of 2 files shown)


PhysSong:
@PaulBatchelor Could you test this and let me know if the bug is fixed?

PhysSong:
I confirmed this fixes #3396(JACK part of #2845, too) just now. I guess this will fix #2934, too. @Umcaruje Can you test this?

PhysSong:
I'm afraid it's not thread safe. SDL provides `SDL_LockAudio` function for thread-safety, but I can't find proper JACK equivalent. Maybe adding a custom lock is the best way?

PhysSong:
Is it okay to merge this?

PhysSong:
Merging since there were no objections for 40 hours.

PR #3988, opened 2017-11-19
fixes copy of non selected TCOs in Songeditor by Ctrl + dragging

BaraMGB:
Fixes https://github.com/LMMS/lmms/pull/3487#issuecomment-344235477

copy of non selected TCOs by Ctrl and dragging now works again.

PR #3649, opened 2017-06-20
shortcut refactor for songeditor

BaraMGB:
This cleans up the keyboard short cuts for the song editor to being consist with the Piano Roll editor. I want to have the same behavior as in the Piano Roll editor.

What is implemented so far:

**hold Ctrl - toggle to selection mode**
you can select your Tcos and by release the Ctrl key it turns back into draw mode.

**deselect all selections by:**
click on empty space (new Tco)
click on unselected Tco
[Ctrl] + [shift] + A
[Esc]

**select all Tcos**
[ctrl] + A
*EDIT*:
**[Ctrl] + drag**
copies a selection or an unselected Tco
no matter the mode (draw or selection)

**click + drag**
moves a selection or an unselected Tco
no matter the mode (draw or selection)

**[DEL]**
deletes all selected Tcos

PR #3487, opened 2017-04-05
Sampletrack - resize from left

BaraMGB:
https://www.youtube.com/watch?v=BQfWqTMjxUE&feature=youtu.be

Things to do:
    - save to project doesn't work, yet

This one is the halve way to split (knife tool) a sample track in two.

PR #4008, opened 2017-11-24
Fix song-editor regressions and BB-editor bugs introduced in #3487

PhysSong:
Fix bugs introduced in #3487. Fixes some bugs in #3487, mostly reverting some changes in #3649.
Additional changes: 
- Show copy hint for moving selected TCOs, as well as moving single TCO

Supersedes #3988.

### DIFF
--- a/src/core/Track.cpp
+++ b/src/core/Track.cpp
@@ -717,10 +717,20 @@ void TrackContentObjectView::mousePressEvent( QMouseEvent * me )
 			}
 			else
 			{
-				m_action = ToggleSelected;
+				gui->songEditor()->m_editor->selectAllTcos( false );
+				QVector<TrackContentObjectView *> tcoViews;
+				tcoViews.push_back( this );
+				DataFile dataFile = createTCODataFiles( tcoViews );
+				QPixmap thumbnail = QPixmap::grabWidget( this ).scaled(
+							128, 128,
+							Qt::KeepAspectRatio,
+							Qt::SmoothTransformation );
+				new StringPairDrag( QString( "tco_%1" ).arg(
+										m_tco->getTrack()->type() ),
+									dataFile.toString(), thumbnail, this );
 			}
 		}
-		else if( !me->modifiers() )
+		else
 		{
 			if( isSelected() )
 			{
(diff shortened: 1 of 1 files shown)


src/core/Track.cpp
```diff
@@ -717,19 +717,38 @@ void TrackContentObjectView::mousePressEvent( QMouseEvent * me )
...
+				tcoViews.push_back( this );
+				DataFile dataFile = createTCODataFiles( tcoViews );
+				QPixmap thumbnail = QPixmap::grabWidget( this ).scaled(
+							128, 128,
+							Qt::KeepAspectRatio,
+							Qt::SmoothTransformation );
+				new StringPairDrag( QString( "tco_%1" ).arg(
+										m_tco->getTrack()->type() ),
+									dataFile.toString(), thumbnail, this );
 			}
```
BaraMGB: okay, but now you can't select/deselect a single TCO by [CTRL] + Click. That's why I moved this whole block into the mouseMoveEvent(). There we can test if the mouse moved 2 pixels since click(copy TCO) or not(toggleSelected). See: https://github.com/LMMS/lmms/pull/3988/files

PhysSong: Thanks! I'll look into that.

PhysSong: Done via 8434d5f.

src/core/Track.cpp
```diff
@@ -720,16 +720,25 @@ void TrackContentObjectView::mousePressEvent( QMouseEvent * me )
 				m_action = ToggleSelected;
 			}
 		}
-		else if( !me->modifiers() )
+		else
```
PhysSong: @BaraMGB Should this be `else`, or `else if( !me->modifiers() )`?

BaraMGB: The difference would be you can't move a tco with pressed shift and/or alt key. Perhaps ``` else if(!me->modifier) ``` is more discriptive.

PhysSong: Okay. The latter is consistent with piano roll. Anyway, you changed it to [the former](https://github.com/LMMS/lmms/pull/3649/commits/46e261f738b09d5b36f83ecae44dab0af18d9cd4#diff-cb2a453c979a4777b9330e2b96644f0eR671) in #3649. Was that intentional?

BaraMGB: No, I tried to make the code simpler. But now I think ```else if(! me->modifiers() ``` is more discriptive. If someone wants to add a new modifier key, he/she can see where to add it in the code. Also I think, it's better to have a functionality only in a certain circumstance (no modifier key is pressed)

PhysSong: Okay. Done via 9d4c377.

BaraMGB:
Sorry, next time I ask before I push to your branch.

PhysSong:
@BaraMGB It's okay. That's what I were going to changed, but I've accidentally undone the change before I commit.

Issue #4017, opened 2017-11-28
i18n: one string appears to be non-translatable


trebmuh:
In [plugins/ladspa_browser/ladspa_browser.cpp line 51](https://github.com/LMMS/lmms/blob/stable-1.2/plugins/ladspa_browser/ladspa_browser.cpp#L51), the string "LADSPA Plugin Browser" is not translatable. I don't know how to make it so, then I'm opening this Issue rather than making a Pull Request. If someone tells me how to do so, I'm happy to send a PR.

PhysSong:
The display name isn't translatable for any other plugins. If you still want to translate this, however, you may wrap that with `QT_TRANSLATE_NOOP` as well as the description. Like this: `QT_TRANSLATE_NOOP( "pluginBrowser", "LADSPA Plugin Browser" )`

trebmuh:
@PhysSong : thanks.

> The display name isn't translatable for any other plugins. If you still want to translate this, however, you may wrap that with QT_TRANSLATE_NOOP as well as the description.

OK, see https://github.com/LMMS/lmms/pull/4019 , am I done correct?

> I guess you want to translate `Tools>LADSPA Plugin Browser`. Right?

Yes.

trebmuh:
any news here?

PR #4019, opened 2017-11-29
i18n for #4017

trebmuh:
> In plugins/ladspa_browser/ladspa_browser.cpp line 51, the string "LADSPA Plugin Browser" is not translatable.

See https://github.com/LMMS/lmms/issues/4017 .

*Edited by @tresf, added description*

### DIFF
--- a/plugins/ladspa_browser/ladspa_browser.cpp
+++ b/plugins/ladspa_browser/ladspa_browser.cpp
@@ -48,7 +48,7 @@ extern "C"
 Plugin::Descriptor PLUGIN_EXPORT ladspabrowser_plugin_descriptor =
 {
 	STRINGIFY( PLUGIN_NAME ),
-	QT_TRANSLATE_NOOP( "LADSPA Plugin Browser" ),
+	QT_TRANSLATE_NOOP( "MainWindow", "LADSPA Plugin Browser" ),
 	QT_TRANSLATE_NOOP( "pluginBrowser",
 				"List installed LADSPA plugins" ),
 	"Danny McRae <khjklujn/at/users.sourceforge.net>",
--- a/plugins/ladspa_browser/ladspa_browser.cpp
+++ b/plugins/ladspa_browser/ladspa_browser.cpp
@@ -48,7 +48,7 @@ extern "C"
 Plugin::Descriptor PLUGIN_EXPORT ladspabrowser_plugin_descriptor =
 {
 	STRINGIFY( PLUGIN_NAME ),
-	"LADSPA Plugin Browser",
+	QT_TRANSLATE_NOOP( "LADSPA Plugin Browser" ),
 	QT_TRANSLATE_NOOP( "pluginBrowser",
 				"List installed LADSPA plugins" ),
 	"Danny McRae <khjklujn/at/users.sourceforge.net>",


plugins/ladspa_browser/ladspa_browser.cpp
```diff
@@ -48,7 +48,7 @@ extern "C"
 Plugin::Descriptor PLUGIN_EXPORT ladspabrowser_plugin_descriptor =
 {
 	STRINGIFY( PLUGIN_NAME ),
-	"LADSPA Plugin Browser",
+	QT_TRANSLATE_NOOP( "LADSPA Plugin Browser" ),
```
PhysSong: You need to specify the context here, like `QT_TRANSLATE_NOOP( "pluginBrowser", "LADSPA Plugin Browser" )` or `QT_TRANSLATE_NOOP( "MainWindow", "LADSPA Plugin Browser" )`. Personally I don't like this approach...

trebmuh: @PhysSong : followed you advice, I choose MainWindow rather than pluginBrowser since it's a menu item on the main windows, but I could be wrong. Am I done better now?

trebmuh:
@PhysSong : anything wrong with the fix?

zonkmachine:
@PhysSong Can this be merged? It's against stable-1.2 and that's fine I think.

PhysSong:
Personally I don't like this approach. As @liushuyu said before in https://github.com/liushuyu/lmms-translations/issues/6#issuecomment-282874572, this isn't much good. I don't know whether the problem still exists or not. So I'm not sure now...

tresf:
> Personally I don't like this approach. As @liushuyu said before in liushuyu/lmms-translations#6 (comment), this isn't much good. I don't know whether the problem still exists or not. So I'm not sure now...

Then plugin needs to be extended to have both a descriptor name and a display name.  Currently, the code uses the display name, which is where this should live.  I'd like to see some evidence (in code or unit test) that this causes a bug.  This problem is only going to propagate further as we get more translations to the software.

trebmuh:
Hi guys, I'm not sure to understand well here.

My understanding is that making this menu option translatable would lead to misbehavior/error when opening the LADSPA plugin brower. Am I understanding it right?

liushuyu:
>Hi guys, I'm not sure to understand well here.
>
>My understanding (after reading at this thread and the answer from @liushuyu in the other thread) is that making this menu option translatable would lead to misbehavior/error when opening the LADSPA plugin brower. Am I understanding it right?

I suspect that's the bug in Qt i18n framework or the way we're handling translations (or multi-byte/wide-char code-point sequences). 

And so like I said in Discord channel few months (?) ago, even KDE found the i18n framework of Qt is so buggy that they decide to write a Qt wrapper to wrap around GNU gettext (see https://github.com/KDE/ki18n)

liushuyu:
Okay, I'll close this thread since it's stale now. Feel free to reopen it if needed

trebmuh:
Not too sure to understand the deep ropes here. Is it a "won't fix" ?

PhysSong:
I think this is a "won't fix in this way". If we find a better way to fix this, the issue might be fixed.

trebmuh:
ack

PR #4022, opened 2017-11-29
Enable SC2185 again

### LINKED ISSUE #3684: SC2185 should be enabled when fixed upstream
Hi,

Just a reminder, see [this comment](https://github.com/lmms/lmms/commit/e90bc630fabb3e8c2e815250594204578931066a#diff-cd31ca8078680e1895d11f56d1fe38c2R7)

zapashcanon:
**Edit:** @tresf, repurposed this PR.

Original: 
> **Title:** Enable SC2185 again
> 
> Fix #3684.
> 
> If it doesn't pass now, we will have to wait until a more recent version of ShellCheck is available in the CI.

Reusing this PR to shellcheck all lmms directories up to 3 directories deep.

### DIFF
--- a/.travis/script.sh
+++ b/.travis/script.sh
@@ -9,6 +9,7 @@ if [ "$TYPE" = 'style' ]; then
 	# shellcheck disable=SC2185
 	# shellcheck disable=SC2046
 	shellcheck $(find -O3 . -maxdepth 3 -type f -name '*.sh' -o -name "*.sh.in")
+	shellcheck doc/bash-completion/lmms
 
 else
 
--- a/.travis/script.sh
+++ b/.travis/script.sh
@@ -8,7 +8,7 @@ if [ "$TYPE" = 'style' ]; then
 	# once it's fixed, it should be enabled again
 	# shellcheck disable=SC2185
 	# shellcheck disable=SC2046
-	shellcheck $(find -O3 "$TRAVIS_BUILD_DIR/.travis/" "$TRAVIS_BUILD_DIR/cmake/" -type f -name '*.sh' -o -name "*.sh.in")
+	shellcheck $(find -O3 . -maxdepth 3 -type f -name '*.sh' -o -name "*.sh.in")
 
 else
 
--- a/doc/bash-completion/lmms
+++ b/doc/bash-completion/lmms
@@ -1,3 +1,4 @@
+#/usr/bin/env bash
 # lmms(1) completion                                       -*- shell-script -*-
 # use shellcheck: "shellcheck -e bash <filename>"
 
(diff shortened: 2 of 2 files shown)


PhysSong:
Travis-CI still seems to use ShellCheck 0.46 now. I think we should wait for next Travis-CI image update.

tresf:
shellcheck 0.4.6 is available yet shellcheck 0.3.3 is what Ubuntu 14.04 uses.  We should leverage a more modern version using some other technique.  Perhaps the docker technique explained here? https://github.com/koalaman/shellcheck/wiki/TravisCI

Or is there a modern PPA available?

zapashcanon:
I don't know, currently the `style` build is used only for SC, but if we wan't to add stuff later, it may not be a good idea to use the SC docker image... Maybe using something like a Debian Unstable docker image which would allow us to have recent version of many things could be something to try ? (Currently, SC is 0.4.6 in Debian Unstable too... but, should be updated quickly I hope)

tresf:
@zapashcanon I vote we close this out for now.  We can always follow up with it with it when our build system gets updated.  @lukas-w has a PR open for switching us to Circle-CI/docker images but I don't see a benefit in keeping it open.

zapashcanon:
I don't know, I'd maybe rather leave it open so I don't forget it should be fixed as I often check my opened PR list [here](https://github.com/pulls) ; and I'll try to relaunch the build sometimes to see if SC has been updated.

tresf:
> so I don't forget it should be fixed as I often check my opened PR list here ; and I'll try to relaunch the build sometimes to see if SC has been updated.

Normally I'd agree.... but in this case I think it's smarter to wait until the CI environment changes as Ubuntu 14.04 seems to have no intentions of updating this package and a change to the CI environment requires this PR to be rebased anyway, which seems like overkill for a one-liner.

tresf:
Actually... perhaps we can have the latest using PPA "pinning".  https://help.ubuntu.com/community/PinningHowto

tresf:
@zapashcanon please switch this PR to use a modern mirror (such as through PPA pinning) or close it out.

tresf:
Ugh... `SC2185` has bad code.   I'll quote [the wiki](https://github.com/koalaman/shellcheck/wiki/SC2185#exceptions): 

> ### Exceptions:
>
> You will get a false positive if you concatenate a series of pre-path flags:
>
>    find -XLE .
>
> In such cases, please either use `find -X -L -E .` or [[ignore]] the message.

Since `-O3` is a single command (not a combination of commands), this is a false-positive.  I guess we'll have to explicitly ignore it.

tresf:
Ok, I'm stumbling on the original bug.  Turns out, even Ubuntu 18.04 doesn't ship with `shellcheck 0.4.7` yet (it's at 0.4.6) which is required for this patch.  It's been over a year and no Ubuntu LTS distro supplies this.  I'm going to choose to ignore this.  The code is self-documenting and leaving the PR open until the next LTS doesn't make sense.

As an aside, the bash completion does't pass shellcheck but it's ignored because we only scan two directories.  Changing to scan all directories with a `-maxdepth` of `3` to avoid traversing into plugin directories.

Issue #3979, opened 2017-11-17
Upgrading CAPS LADSPA Plugins


tresf:
The [CAPS LADSPA plugins](http://quitte.de/dsp/caps.html) -- or more properly known in LMMS as "`C*`", are authored by Tim Goetze and they have changed considerably over the years.  The current version as of writing this is 0.9.24.

Current examples:
![image](https://user-images.githubusercontent.com/6345473/32966562-1da6f570-cba8-11e7-89c5-3ff25ebbee79.png)

To remain consistent with our goal to [stop bundling 3rd party source code with LMMS](#3960), we're faced with a decision about CAPS... 

To quote @tobydox about breaking user-space:

> *"[...] We have been shipping CAPS 0.4.x for years now and the available plugins and their behaviour remained stable for the users all the time. If you would depend on https://packages.debian.org/jessie/caps instead, lots of effects would be missing when loading a project as the available plugins inside CAPS have changed significantly between 0.4.x and 0.9.x. If you'd simply rename caps.so to caps-lmms.so no plugins would be loaded at all as the library base name (caps) is saved in the project file (internally we load caps.so, caps.dll or caps.dylib depending on the platform). If we (LMMS developers) decide to switch to 0.9.x we will implement an according upgrade mechanism internally (and probably even still provide CAPS 0.4.x in a separate library) so things don't break for the user. -Toby"*

At some point, we have to look forward and not backward though.  Moddevices has a [clone of CAPS](https://github.com/moddevices/caps-lv2) that we can safely submodule from once LV2 is adopted but that's a little ways out, so we'd just be using it as a reliable and maintained LADSPAv1 mirror for now.

I've done a poll on Discord about migrating over to 0.9.24 and so far, the team seems OK with breaking user-space in lieu of an updated library.

**Missing Plugins** (edit me!):

* `C* Clip`
* `C* Plate2x2` (upgraded to `C* PlateX2` via https://github.com/tresf/lmms/commit/5bee27cc40be8b11349660ed9a5ec1843bca2f61)
* `C* Eq` (upgraded to `C* Eq10` via https://github.com/tresf/lmms/commit/5bee27cc40be8b11349660ed9a5ec1843bca2f61)

The way I see it is we have four options... 

1. 👍   Do nothing, keep CAPS 0.4.x
2. 😄   Ditch 0.4.x, upgrade to 0.9.24, break user-space
[...]

PR #4027, opened 2017-12-01
Switch caps to submodule

tresf:
The [CAPS LADSPA plugins](http://quitte.de/dsp/caps.html) -- or more commonly known in LMMS as "`C*`" -- are authored by Tim Goetze and they have changed considerably over the years.  The current version as of writing this is 0.9.24.

No upstream VCS exists, but moddevices keeps a viable [clone of CAPS](https://github.com/moddevices/caps-lv2) that we can safely use as an LADSPA mirror (and eventually for LV2, when we add support).

Since this involves testing presets and projects, help is needed.  To compare the mapping of a plugin, export a project as `.mmp` (without the z) and compare the ladspa sections.

### Compatibility

0.4 | 0.9
-- | --
AmpIII | --
AmpIV | --
AmpV | --
AmpVTS | AmpVTS
AutoWah | AutoFilter [💡](https://user-images.githubusercontent.com/6345473/33510902-234dd834-d6e0-11e7-8ec2-22d417aae090.png)
CabinetI | --
CabinetII | --
  | CabinetIII
  | CabinetIV
ChorusI | ChorusI
ChorusII | --
Clip | Saturate [🗯Upgrade broken, see notes](https://github.com/LMMS/lmms/commit/0e14f399c42ded79910e716ce4efb3b3dbba7ae2#r25697333) 0e14f39
Compress | Compress
  | CompressX2
Eq | Eq10 ✅ 5bee27c
Eq2x2 | --
  | Eq10x2
  | Eq4p
  | EqFA4p
Narrower | Narrower
  | Noisegate
PhaserI | --
PhaserII | PhaserII
Plate2x2 | PlateX2 ✅ 5bee27c
  | Spice
  | SpiceX2
PreampIII | --
PreampIV | --
SweepVFI | --
SweepVFII | --
ToneStack | ToneStack
ToneStackLT | --

Supercedes #3979


tresf:
FYI,  Travis-CI checks are slightly misleading.  It's passing for LMMS, but failing for tresf.  https://travis-ci.org/LMMS/lmms/builds/366253352

JohannesLorenz:
@tresf The compatibility table tells us which effects are planned to map. What, however, about the following?

* AmpIII, AmpIV, AmpV -> AmpVTS
* CabinetI, CabinetII -> CabinettIV
* ChorusII -> ChorusI
* PhaserI -> PhaserII
* ToneStackLT -> ToneStack

tresf:
@JohannesLorenz we either drop them or do our best to map them.  Sorry I realize re-reading the original description that this is not obvious.

Ideally, we'd try to maintain some form of compatibility if possible.

zonkmachine:
> The current version as of writing this is 0.9.24.

And as of today caps latest release reads` Version 0.9.26  ·  October 2018`

I believe Eq10x2 is to Eq2x2 what Eq10 is to Eq. So...

0.4 | 0.9
-- | --
Eq | Eq10 ✅ 5bee27c
Eq2x2 | Eq10x2

JohannesLorenz:
@tresf I think we have no other chance than finishing this PR :smiley:

I'll try to rebase this to master and try to get the code finished now.

JohannesLorenz:
Marking as TODO: Master commit ea98ba4daefd4055ca98f05d31dbdd3522792a01 needs to go into CAPS.

JohannesLorenz:
@mrkline Can you please open a PR of your changes of ea98ba4 against [moddevices/caps-lv2](https://github.com/moddevices/caps-lv2)?

JohannesLorenz:
sorry, accidentally closed

tresf:
> I'll try to rebase this to master and try to get the code finished now.

Thanks.  Note, at some point we'll need to merge unfinished and just handle the upgrade issues as "I'm sorry" in support/Discord.   Thanks for putting some time to this.  I'm sure it's obvious, but I'm no longer actively maintaining my branches and PRs so please do as you need (push, close, clone, etc).

JohannesLorenz:
@mrkline I submitted the patch for you now.

JohannesLorenz:
I propose to find out the port mappings by testing (as it's too complicated to analyze from the code). Proposal:

1. Open LMMS both in d29066fa83adbd1e5ecec6573c0583cc66d67ea9 (0.4, "old") and 04263d3cf797db8ae1e5c26df7bc791b832bc826 (0.9, "new")
2. Load the same sound in both. You can choose any instrument, but the sound must be useful for the effect. For example, a plain sine wave might not be useful for testing a lowpass filter.
3. Choose an effect from the table above, and apply the "0.4" version to your instrument in 0.4 LMMS, the "0.9" version to your instrument in 0.9 LMMS.
4. For all 0.4 ports, find out if they have an equivalent port in the 0.9 version. Sometimes they have completely new names. Suspicious equivalents may have the same default value and range. Write down if the port has an equivalent, and if yes, if the minimum/maximum are different, or if the mapping function changed between linear/log.
5. For all 0.9 ports that you found are no equivalents to 0.4 ports, write down if their default values must be changed to match the 0.4 effects.
6. Please add all you wrote down into a post that you reply here.

Example (not necessarily correct):

| AmpVTS 0.4 | AmpVTS 0.9 | scale changes |
| --- | --- | --- |
| gain | bright | old scale was linear, new scale is logarithmic |
| bass | bass | old maximum value is now 0.5 |
| mid | mid | - |
| treble | treble | - |
| - | lowcut | must be defaulted to 0 |

Spekular:
> I propose to find out the port mappings by testing (as it's too complicated to analyze from the code).

Could we first ask the developer via the Contact information on [this page](http://quitte.de/dsp/caps.html) if they happen to know how the ports have changed? (Assuming no one has done this yet). If they do, that would save a lot of effort.

The CAPS Version Query section also seems relevant.

JohannesLorenz:
> Could we first ask the developer via the Contact information

OK, I sent and e-mail now. The project seemed pretty dead to me, but let's still try.

tresf:
> The project seemed pretty dead to me, but let's still try.

Which project?  [Tim Goetze](http://quitte.de/dsp/caps.html), the author has always replied to my emails in the past.  Quoting one from 2017:

> > **Tres wrote:** As we prepared our application for the latest CAPS version, we're hitting some upgrade issues from previous versions [...] is "Clip", now "Saturate" ?

> **Tim wrote:** [Yes ...] "clip" mode the Saturate plugin is identical to the old Clip
plugin, save for a small change in output volume.  [...] you'll probably also have to turn off the
DC-blocking filter and set the bias control to zero to achieve identical operation.

He replied in 1 day.

Of course, we don't use his source code, we use a submodule managed by the moddevices team, but that's a separate issue entirely, no?

tresf:
In regards to submitting patches upstream, this is what Tim says:

> [I'm] definitely interested [in patches], and unified diff preferred (although so far,
such changes have been sparse and small enough for manual management;
after all, caps is quite a small project).

His email is tim\<at>quitte.de

JohannesLorenz:
@tresf I was under that impression CAPS was mostly dead because we had submitted CAPS PRs (small ones that were really important) which took around 1 year to get merged.

Anyways, Tim also replied to a mail I wrote to him. He said there's nothing like a mapping, it would be too complicated to find out a good mapping that fits the complex algorithm changes. When I asked how Ardour would have done it, he did not know exactly, but was thinking they don't have such a mapping. Which could mean this PR is more luxury than most users might expect (which does not mean that I dislike the idea of this PR).

tresf:
> this PR is more luxury than most users might expect (which does not mean that I dislike the idea of this PR).

I agree with this statement wholeheartedly but what I think we'll find as a result of merging this is the fact that the CAPS we ship is the CAPS our users use.  This means that projects have no upgrade path to the latest LMMS without performing surgery on the LMMS installation.   As a CAPS user myself, this means that I'll likely downgrade LMMS to open old projects.  This isn't isolated to CAPS (we introduce a lot of breaking changes) but it'll be something we'll be seeing more and more in #support on Discord once the new builds go up.

My vote is to just merge this and watch people complain, tackle as needed however you also have a separate issue you created "Find out CAPS port mappings by testing #5680" to help fix the long list of plugins affected so is it OK if I leave this decision up to you @JohannesLorenz?

JohannesLorenz:
> you also have a separate issue you created "Find out CAPS port mappings by testing #5680" to help fix the long list of plugins affected

I once again asked in `#testing` in discord, and there is no interest to test anything now.

> My vote is to just merge this and watch people complain, tackle as needed

Agreed that we can hope that it works good enough and fix things later if required. If someone complains, at least we can ask them to test the switches.

What should we do now? Should we send this PR into a testing phase (as most PRs do), or just merge now?

tresf:
> What should we do now? Should we send this PR into a testing phase (as most PRs do), or just merge now?

I'm fine with either.  You and I tested this quite a bit so we know it works.  Most of the code is just upgrade routines anyway.  It also appears the `master` branch of the submodule hasn't changed, so we don't have to fast-forward that either.

JohannesLorenz:
If someone discovers issues during testing, they will need to analyze what's wrong. Which was why I opened that "Find out CAPS port mappings by testing #5680" issue, which no one wanted to work on. So if we only discover issues that no one wants to fix, I'm for merging this without testing :smiley: If that makes sense.

tresf:
> I'm for merging this without testing 😃 If that makes sense.

Ok.  I just realized that @zonkmachine had pointed out the following:

> I believe Eq10x2 is to Eq2x2 what Eq10 is to Eq. So...
>
> 0.4 | 0.9
> -- | --
> Eq | Eq10 ✅ 5bee27c
> Eq2x2 | Eq10x2

I've updated the table to reflect this (sorry for the delay), and the mapping is probably very similar to the `Eq` commit, but I'm not surewhetherr or not this task is worth holding the entire PR up. I could probably tackle it in a few hours, but I'm not sure it's worth the time.

PhysSong:
Note that the `register` keyword should be removed in CAPS for C++17 support, as in #6133.

JohannesLorenz:
> Note that the `register` keyword should be removed in CAPS for C++17 support, as in #6133.

Thanks for the info. I mailed the CAPS author (Tim Goetze) and asked him to remove it in the next CAPS version.

zonkmachine:
> The current version as of writing this is 0.9.24.

The current version is now 0.9.26 though the latest release if more of a nudge of 0.9.25
Maybe we just provide upgrade routines where we easily can and then simply leave the user with a dummy effect.

Rossmaxx:
Since this PR contains lots of fixes, would it make sense to merge untested and fix issues as they come by.

Or we can just fix using the demo projects as reference. If everything loads, merge.

src/core/DataFile.cpp
```diff
@@ -1100,14 +1107,15 @@ void DataFile::upgrade_1_3_0()
...
+					/*
+					 * CALF port changes
+					 */
 					if( attribute.attribute( "name" ) == "plugin" &&
 							( attribute.attribute( "value" ) == "MultibandLimiter" ||
 							attribute.attribute( "value" ) == "MultibandCompressor" ||
 							attribute.attribute( "value" ) == "MultibandGate" ) )
 					{
-						auto fn = [&](QDomElement& port, int num, QList<QDomElement>&, QList<QDomElement>& removeList)
+						auto fn = [&](QDomElement& port, int, int num, QList<QDomElement>&, QList<QDomElement>& removeList)
```
Rossmaxx, line 1118: Isn't there a name for the unidentified int value. Nothing major but for readability sake.

Rossmaxx (commented):
Something minor i found at first glance.

zonkmachine:
ToneStackLT
---
0.4 ToneStackLT is pretty much (se release note from caps 0.9.1 below) the same as 0.4 ToneStack with Model value:0 . Same controls. The values are hard coded here:
https://github.com/LMMS/lmms/blob/32fe3e50e7d951a456764dc1692b83f812bb2a87/plugins/LadspaEffect/caps/dsp/ToneStack.h#L214
Same as the first element for ToneStack:
https://github.com/LMMS/lmms/blob/32fe3e50e7d951a456764dc1692b83f812bb2a87/plugins/LadspaEffect/caps/ToneStack.cc#L44-L46

ToneStack wasn't removed until after 0.9 was released.

https://github.com/moddevices/caps-lv2/blob/master/CHANGES
0.9.24 * dropped 44.1 kHz ToneStackLT
0.9.1 * ToneStackLT rolled into ToneStack and eliminated

Finally the modpeople have something to add
https://pedalboards.moddevices.com/plugins/aHR0cDovL21vZGRldmljZXMuY29tL3BsdWdpbnMvY2Fwcy9Ub25lU3RhY2s=
```
The "5F6-A LT" model is using the lattice filter implementation mentioned in [yeh06], operating on
precomputed simulation data for 44.1 kHz.
```
So, apart from the similar data there may also be a different method used for this setting and it seem to have been baked into the 0.9.1 version of ToneStack . Apparently it's hard coded for 44.1 kHz.

zonkmachine:
ToneStackLT upgrades to ToneStack:model 0 in f63ec84d698fad297cfe58c8052d1b13c3e8942c

JohannesLorenz:
Thanks so far @zonkmachine . Just for my understanding, why did you force-push the first commit? ("Turn CAPS into submodule (#4027)")

zonkmachine:
> Thanks so far @zonkmachine . Just for my understanding, why did you force-push the first commit? ("Turn CAPS into submodule (#4027)")

I had rebased again on latest master.

src/core/DataFile.cpp
JohannesLorenz: In my opinion, not useful. This would only be relevant to the caller of that lambda, which is the common routine `iterate_ladspa_ports`. However, at this place, we do not need to know what to pass. So I would prefer to resolve these comments without change, unless you see a strong reason.

Rossmaxx: Ohh ok then.

bratpeki:
@tresf Can your LMMS fork, under the "caps" branch, be used for reliably testing the 0.4-0.9 difference? If so, I'll get to work testing the missing plugins.

tresf:
> @tresf Can your LMMS fork, under the "caps" branch, be used for reliably testing the 0.4-0.9 difference? If so, I'll get to work testing the missing plugins.

I would assume so?

bratpeki:
What I mean is: If I were to make a project in master LMMS and your fork containing only a plugin on the master track, would the only difference be that plugin data? Has nothing else changed in the process?

tresf:
> What I mean is: If I were to make two projects, one in master LMMS and one in your fork, such that they contain only one plugin on the master track, would the only difference be that plugin data? Has nothing else changed in the process?

* The word "plugin" is a bit ambiguous here, CAPS exposes many plugins
* This PR should be fast-forwarded so that the baselines are identical

That said, I think what you're describing is correct.  We haven't done any major plugin data restructuring as a result of this PR and any upgrade routines that this PR ads should only affect upgrades from older LMMS versions, not new projects, assuming no mistakes were made. 🍻

bratpeki:
The word "plugin" is in reference to the specific CAPS plugin placed on the given mixer track, for any mention of the word, you can be sure I meant it as such.

Regarding the answer, that sounds great, I'll download your fork and build it, and get to testing!

bratpeki:
## ChorusI

The input port has been moved from the first to the second-to-last position.

I'll comment more later.

tresf:
> @tresf What do we do about the missing ChorusII? Are we dropping it?

@JohannesLorenz asked a similar question, quoting:

> @tresf The compatibility table tells us which effects are planned to map. What, however, about the following?
> 
> * AmpIII, AmpIV, AmpV -> AmpVTS
> * CabinetI, CabinetII -> CabinettIV
> * ChorusII -> ChorusI
> * PhaserI -> PhaserII
> * ToneStackLT -> ToneStack

I just created a test project to check for audible differences between `ChrorusI` and `ChorusII` and they're quite different.  Since the ports should be a `1:1` map, I'm fine with mapping these over, but it won't sound right, so we would want to issue some type of soft warning if we do (e.g. "C* ChorusII" plugin is no longer available, "C* ChorusI" was used instead).

My best guess is that ChorusII just didn't sound good enough to the author of CAPS plugins to justify keeping around.  That's not to say people didn't take advantage of it in projects and those projects won't sound right after this upgrade.  In my opinion, `ChorusII` makes instruments sound "tinny" with more high-pitched feedback added. `ChorusI` in my opinion sounds more "full".

Replacing a plugin with one that sounds different is likely to attract some strong opinions.  I'm fine with either the decision to remove or replace.  In both cases, the project won't sound right.

bratpeki:
> I'm fine with either the decision to remove or replace.

I feel like replacing *could* be possible if we make our own fork, but shouldn't be done.

My argument is mainly that the maintainer of it felt that it wasn't necessary, and there's little reason to doubt such a decision. Also, it alleviates our headache and makes 1.3 happen sooner.

So I'm all for dropping it, the user is gonna get a message that the plugin has not been found anyway, and we'll list the dropped plugins in the 1.3 changelog.

bratpeki:
> so we would want to issue some type of soft warning if we do (e.g. "C* ChorusII" plugin is no longer available, "C* ChorusI" was used instead)

That also sounds nice, but I wouldn't keep it around for longer than 1.3.x.

tresf:
> That also sounds nice, but I wouldn't keep it around for longer than 1.3.x.

Just note that all upgrade routines live essentially forever.  Unless there's a justification to remove a routine, there's not much cost to keep it around.

bratpeki:
Alright, I'll then move on to the other plugins, to wrap it up.

JohannesLorenz:
There are also some other possibilities:
1. Ask the author why they removed the mentioned effects and especially how they imagine a DAW should handle if a user still has presets for those. I just sent out a mail and will wait for reply.
2. We could port those removed effects into our fork, so they still exist there.
3. Another proposal: We deliver both 0.4 and 0.9 libraries with LMMS (but hide 0.4 for new projects). That way, users can finally use the new libraries, but the old projects will load without change. A disadvantage is the doubled compile time, it would be interesting if someone can give the compile times on a typical desktop for both 0.4 (LMMS master) and a version with 0.4 and 0.9, and compare the compile times.

bratpeki:
Now that we see that this isn't as easy as thought in 2017, maybe a re-vote would be in order?

I think we should just move this to v2.0, since we can then break compat without any worry.

PR #4031, opened 2017-12-04
Remove the "Empty project" message and allow exporting empty projects.

Sawuare:


Sawuare:
Tested Thoroughly. Merging.

Sawuare:
Oh I merged with a merge commit instead of a squash commit. Sorry. :(
What should I do?

tresf:
>  I merged with a merge commit instead of a squash commit

It's only a single commit.  That's fine. 👍

PR #4034, opened 2017-12-05
Use mean based logic for controlling loop points.

Sawuare:
With this, right clicking the time line widget will set the nearest loop point.

Additionally, it only allows setting loop points with right clicking.


Umcaruje:
what If I just want to loop the first bar?

Spekular:
@Umcaruje clicking in a certain region only selects that handle, after that
it can be dragged. I've also suggested Shift+L/R to directly set the
start/endpoint in a manner similar to the current behavior. That way those
who don't like the new behavior can use something similar to the
current/old way.

qnebra:
Looks way more intuitive than current loop points logic.

vlad1777d:
As I understood, you divided distance between loop points on 2 parts.
If somebody click at the left one - than will be moved left loop point, if on right one - right.

If some - than I agree, it's a better behaviour. +1
Because I always confuse on which button to press to move appropriate loop point =)

Sawuare:
> I've also suggested Shift+L/R to directly set the start/endpoint in a manner similar to the current behavior. That way those who don't like the new behavior can use something similar to the current/old way.

@Spekular The new behavior is much better than the old behavior, so I think all users will like it.

Wallacoloo:
Just gave this a try, and I like it. @Spekular, did you have some hang up with this change, or can we merge this?

Spekular:
I think it would be best to allow the old method using a modifier. While
the new method is great, the old method can be faster (any time you wish to
set a loop point within the selection range of another), and in certain
situations even more convenient.

Having one easy way and one fast way to do something isn't unusual. See:
Blender and hotkeys.

Wallacoloo:
@Spekular Okay, if there's a clean way to do that, I won't argue.

I'm pretty sure the existing behavior is normal right-click moves the loop endpoint, and shift+right-click moves the loop start point. Ctrl disables the magnetic behavior. So both the shift and ctrl modifiers are in use. I guess we could somehow throw `Alt` into the mix, but that many modifiers for something seems excessive. We could tuck away an "Enable advanced keybindings" options in the settings somewhere to control which set of keybindings are enabled.

I'm not particularly fond of adding another setting like this, but also I really like the new behavior and would love if there's a way to get it in that keeps older users happy.

Can you think of a good way that the old behavior could still exist alongside this one?

Spekular:
Ah, I forgot about laptop users. I still think my earlier proposal works,
unless I'm missing something. It would shuffle around the controls a bit,
but that can't be avoided if this new method should be default (which I
think it ought to)

RMB: New method, select and drag
Shift+RMB: Set endpoint at cursor
Shift+LMB: Set startpoint at cursor
Ctrl: Still fine adjust
Alt: Still unused

This would unify controls for laptop (/two button) and desktop (/three
button) users, too.

vlad1777d:
I use left-handed mouse configuration, so pressing LMB / CMB always confused me...

Wallacoloo:
@Spekular Ah! Sorry, I misunderstood your original post - thanks for having the patience to reexplain it.

@Sawuare I know it's been ages since you opened this PR, but if you have an opportunity sometime to adjust the behavior to match Spekular's last comment, then I think we could get this merged.

Sawuare:
Shift+LMB is used for `SelectSongTCO`. I think it should be changed to Ctrl+LMB, to leave Shift for the old loop behavior. Thoughts?

Wallacoloo:
It can't be control because that's already used to disable the "magnetic" behavior (i.e. hold control to set a loop point that's not aligned to a bar), IIRC.

I don't know much about "SelectSongTCO": what is that/what exactly is the problem with using shift+LMB?

---- On Thu, 10 May 2018 12:15:19 -0700  notifications@github.com wrote ----

Shift+LMB is used for SelectSongTCO. I think it should be changed to Ctrl+LMB, to leave Shift for the old loop behavior. Thoughts?

Spekular:
I've never used that before, but I agree that ctrl+drag would make sense
for it, considering we use ctrl+drag for box select elsewhere.

Spekular:
It selects every TCO range that you drag over. It overlaps a bit with box
select but could be more convenient because:

1. You don't have to change mode to use it.
2. You don't have to scroll down while selecting if your project is very
tall.

Out of scope for this, but perhaps

-Alt should be the fine adjust modifier in the timeline and song editor
-Shift should be the "clone" modifier in song editor, "old behavior"
modifier in the timeline
-Ctrl should be the box select shortcut in song editor and TCO select
modifier in the timeline

This would be consistent with piano roll amd allow easier box select
(perhaps select mode could be entirely removed).

Sawuare:
> @Spekular Okay, if there's a clean way to do that, I won't argue.

@Wallacoloo I tried and it does not even work, let alone being a clean way.
No users mentioned disliking the new behavior, so I think the old one should be removed.

zonkmachine:
@Sawuare The new behavior is way better... until you zoom in on a part and want to set new loop points. Can the selection be relative to the looped area visible when not both loop points are visible?

Sawuare:
@zonkmachine I think it would be confusing, plus it's just an extra click to move the loop point near the desired position before zooming in and making fine adjustments.

DouglasDGI:
Tested it out.  I like the new behavior.  I did run into the same problem @zonkmachine ran into with not being able to easily set the loop points where I wanted them while zoomed in, but I can't think of a better behavior that would work without confusing some people.
I think it would be cool to (when right clicking on the bar) act like the end of the loop point is at the edge of the screen if it's past the edge of the screen, but I'm not sure how easily people would be able to figure that out automatically, it would probably require a lot of testing to know.
I'm fine with the current behavior, I didn't notice any other problems.  :+1:

eagles051387:
I might have an idea to solve the issue of loop points when zoomed in.

Logic pro you have a bar that when its coloured yellow the loop is on. Now that bar you can drag it regardless if you are zoomed in or not all that adjusts when zoomed in would be the position of the beats in that bar.

Would that potentially help solve the issue?

BaraMGB:
This new behaviour is much more better than the current implementation. I like it. I vote for merge it as is. For any problems with zoom we can add more functionality to this later (e.g. "set loop points to selection", or moving loop range with middle mouse button or something else)

lookitsnicholas:
Hey, longtime user here. I attempted using the alpha builds in hopes of using the new Spectrum Analyser, but ended up having to revert for multiple reasons, the biggest being this PR as a personally breaking workflow regression.

As mentioned in #5505, trying to select a new smaller region while a large loop is currently active is borderline (if not *literally* impossible in some scenarios) with this change merged as-is. I [brought this up](https://discord.com/channels/203559236729438208/250805775646064641/839505792897646642) on Discord earlier today, and it seems like the general sentiment is that "all users" do not like it. The common counterargument seems to primarily consist of three elements:

1. Zooming out as a (partial) mitigation
2. This method is faster
3. This PR's looping mechanism is analogous to other DAWs

My reply to the first point is what I said verbatim in Discord:
> this change in my head is equivalent to an OS update that increases the font size and then the feedback being to back away from your screen if you want the text to appear smaller - doesn’t make any sense

In addition, this method in no way is faster in the aforementioned lengthy loop case. It either requires the previous point's action to be initiated in addition to zooming **back in** to complete modifications of greater resolution, getting extremely lucky with choosing which loop point to change at any given time, or both. 

Finally, this is definitely not how it works within Ableton and FL Studio (which was explicitly mentioned). Both offer an objectively more intuitive/simplistic method of looping, with the latter allowing for multiple loop points (transitively allowing for one-click manipulation).

> Ah, I forgot about laptop users.

This seems to be a common occurrence, as this reminded me of #4522 that @tresf and I ended up resolving about two years ago regarding the inability to add measures on Macintoshes. The approach then taken was to ask the users of Discord what their desired preference was, upon which a purely additive consensus was agreed via a survey (that can be found [here](https://discord.com/channels/203559236729438208/250805775646064641/476108528545693718)) - which I would suggest to gain further input on this change. Maybe this did occur and I missed it, though (not on Discord often due to work, but I would genuinely be surprised based of today's feedback)? Perhaps additional consideration should be taken into account for changes such as this, especially since these devices unfortunately do not seem to be the focus here.

In relation to this PR specifically, however - I, as well as I am sure some other users, would *ideally* like the old controls back by default and this PR available for those who do indeed desire mean-based loop snapping - or, something else... maybe keep this PR with a Shift+LMB invocation and preserve the existing key combination for the legacy looping mechanism, just a solution that is not destructive.

I would be intrigued to hear more input on this, however - it just seems like this change is not ideal if it prompts a multiple of responses such as:
> Is there a way to go back to the original way?

For the time being, though - I'll be staying on version 1.2.2 since I am using an M1 MacBook Air; would be nice to eventually upgrade because running an Apple silicon version of 1.3 would be ideal sometime down the road.

Sawuare:
Some like this change and some hate it, so I will revert it for now and maybe offer it as an alternative in the future.

qnebra:
I had even better idea. Revert every UX change from last 15 years in lmms.

- right mouse button - agree, it can be problematic in device without dedicated RM button, like Macbook or those with touch screen. 
- mean logic - if user is unable to perform basic estimation then something is really wrong with him.

Veratil:
> 
> 
> Some like this change and some hate it, so I will revert it for now and maybe offer it as an alternative in the future.

I think we should get a poll on this before we revert it. How many like the new way? How many like the old? Is this or the previous way more intuitive? Is there a way to make this way better?



> 
> 
> I had even better idea. Revert every UX change from last 15 years in lmms.

This kind of comment is not needed.

rdrpenguin04:
I honestly prefer the old way because, for longer loops, I now have to zoom out or scroll a long ways to move one end of the loop. This is primarily an issue when I'm trying to loop a small section to work on balance or tune a synth.

OTOH, it was inconvenient to remember which button combination moved which side of the loop.

tresf:
I agree, with all comments, especially @Veratil's proposal to make the existing behavior palatable.  Interestingly enough, this regression was predicted by @Spekular and a few others, quoting:

@Spekular wrote:

> @Umcaruje clicking in a certain region only selects that handle, after that it can be dragged. I've also suggested Shift+L/R to directly set the start/endpoint in a manner similar to the current behavior. **That way those who don't like the new behavior can use something similar to the current/old way.**
> ...
> I think it would be best to allow the old method using a modifier. While the new method is great, the old method can be faster (any time you wish to set a loop point within the selection range of another), and in certain situations even more convenient.


@Wallacoloo wrote:

> I'm pretty sure the existing behavior is normal right-click moves the loop endpoint, and shift+right-click moves the loop start point. Ctrl disables the magnetic behavior. So both the shift and ctrl modifiers are in use. I guess we could somehow throw Alt into the mix, but that many modifiers for something seems excessive. We could tuck away an "Enable advanced keybindings" options in the settings somewhere to control which set of keybindings are enabled.
>
> I'm not particularly fond of adding another setting like this, but also I really like the new behavior and would love if there's a way to get it in that keeps older users happy.
>
> Can you think of a good way that the old behavior could still exist alongside this one?

... and then the conversation circles around which keyboards shortcuts are even possible.

@DouglasDGI writes:

> Tested it out. I like the new behavior. **I did run into the same problem @zonkmachine ran into with not being able to easily set the loop points where I wanted them while zoomed in, but I can't think of a better behavior that would work without confusing some people.**

My vote is the following:

@lookitsnicholas, open a new bug report with a screencast of the exact problem.  Tag those people here that seem vested so they're subscribed.

We approach it two ways, these are NOT mutually exclusive:
* Add back some bindings to keep the old behavior<br>
-- AND --
* We look to see if something can be done to fix the zoomed-in behavior

qnebra:
> > I had even better idea. Revert every UX change from last 15 years in lmms.
> 
> This kind of comment is not needed.

Because it is pointless comment. Always someone complains when some changes to UX are made, so best way to avoid that will be simply to not do any changes.

I have no idea if it is ever possible to do, to set loop points based on currently visible section of timeline in Song Editor. I guess it would be really complex in code.

tresf:
> Because it is pointless comment. Always someone complains when some changes to UX are made, so best way to avoid that will be simply to not do any changes.

The person posting -- also a code contributor -- is making a usability argument for the old behavior. I too use a Macbook for production.  Let's be careful how dismissive we are to our users, especially those willing to submit patches. 🍻

RoxasKH:
The principal fact for why this change can be not very well accepted by people is actually because it's a change.
I mean, i struggled with it, too, but only because as an old user i got used to the old logic.

Let's be honest, tho, if we're talking about being intuitive, the new behaviour wins. Yes it can be a bit uncomfortable in doing things, but let's think about new users: move the left point is way more accessible now.
I don't really see this as a regression, sometimes it needs some more tweaking, but it's still very functional.

The main point while this kinda alienate me is because i'm used to the old one to the point is become something i do without even thinking.
It doesn't mean i can't for example get used to the new one, which can also be more comfortable in other cases.

Even if key shortcut are essential in music productioni don't know if the shift-click was better or not.

qnebra:
And I agree with that point about usabiity. Unlike in moment of writing first comment in this PR I currently had laptop without dedicated RMB, and luckily for me I had connected mouse to it. Without it using this scheme of control will be more difficult.

tresf:
> But revert entire PR because someone is used to old way and want to preserve old behaviour forever?

I think we can all agree, reverting the feature in its entirety is a bad idea.

RoxasKH:
Probably we can keep implemented the shift shortcut for moving only the left loop point, can we?

It wouldn't be a complete revert, but it's still something

I dunno how much consistent it'd be tho

akimaze:
Why not add an option in the settings? If each of the approaches has its advantages and disadvantages and it is difficult to decide, it may be worth putting the decision in the hands of the user.

By default, this may be a new behavior because it is easier to discover.

In my opinion, the old solution was much more effective. Now I often have to make a lot more moves to make sure the correct point gets changed. Especially when I don't see the starting and end point. It happens (quite often) that the point that I do not want changes and then I have to set two points again. This is quite annoying.

New solution is better when you see start and end point. When loop is bigger old solution wins.

akimaze:
Another solution could be a graphical representation of which point will change when we move the cursor over the loop bar. Then I would know if I need to move the cursor somewhere else to achieve my goal.

RoxasKH:
> 
> 
> Why not add an option in the settings? If each of the approaches has its advantages and disadvantages and it is difficult to decide, it may be worth putting the decision in the hands of the user.
> 
> By default, this may be a new behavior because it is easier to discover.
> 
> In my opinion, the old solution was much more effective. Now I often have to make a lot more moves to make sure the correct point gets changed. Especially when I don't see the starting and end point. It happens (quite often) that the point that I do not want changes and then I have to set two points again. This is quite annoying.
> 
> New solution is better when you see start and end point. When loop is bigger old solution wins.

The new one requires zooming out, is true, but it's just another passage to do.

If dev were to add an option in the settings for ever change they made the settings would be an infinite list of things, so i don't know if it's the right way to go.

akimaze:
>The new one requires zooming out, is true, but it's just another passage to do.
> If dev were to add an option in the settings for ever change they made the settings would be an infinite list of things, so i don't know if it's the right way to go.

In my opinion, the developer should make this decision (should there be an option?) every time it is necessary. Each case/functionality is different, so there is no precedent law (if we add option once we must add option for every subsequent change). Also, there is no obligation that each task can only be done in one way if there are arguments/other scenarios for it.

Option can be removed when new new solution will work fine in any scenario.

lookitsnicholas:
> Another solution could be a graphical representation of which point will change when we move the cursor over the loop bar. Then I would know if I need to move the cursor somewhere else to achieve my goal. Something like that:
> 
> 
> 
> image

This seems like a Band-Aid (if anything else) because it still requires that the user has to perform additional movements based on the dynamic UI.

My overarching issue with this PR is not necessarily the fact that the previous behavior is changing, it is that it is **breaking** the objectively more efficient route with seemingly no initial reason for it to exist in the first place (which the counterarguments do not seem to directly address).

If discoverability was indeed somehow the main issue, then why not add a tooltip specifying the old control scene upon hover?

Again - if there is a lot of support for this behavior existing, then I would present it as an option instead of destructively modifying the existing behavior without significant feedback.

I will get to creating the new issue in a bit, I just have other things to do in the interim.

Spekular:
> I will get to creating the new issue in a bit, I just have other things to do in the interim.

There's no need, there's already an open issue for it: #5505

lookitsnicholas:
> > I will get to creating the new issue in a bit, I just have other things to do in the interim.
> 
> 
> 
> There's no need, there's already an open issue for it: #5505

This was per suggestion from a previous reply, not sure if any traction was garnered in relation to that particular report.

tresf:
I didn't realized #5505 existed.  @JohannesLorenz actually drew the issue pretty clearly using some ascii characters , so probably no need for a screencast (or we can edit his to include yours).

What if the mean-based logic was changed to be based on the viewable region, not the entire region?

lookitsnicholas:
> What if the mean-based logic was changed to be based on the viewable region, not the entire region?

Maybe I'm missing something, but how would that even work if the loop is located near the extreme of one side of the screen?

tresf:
> Maybe I'm missing something, but how would that even work if the loop is located near the extreme of one side of the screen?

Actual:
```
<----------------------------------->
                    ^------- mean
```

Viewable:
```
<----------------->
          ^------ mean
```

Un-viewable:
```
                          <-------------------->
   ^------- (no mean, like clicking new)
```

lookitsnicholas:
So, zoom actions would be required from the user in certain cases (see: widescreen monitor). Additionally, that does not seem to address the issue that would arise from odd-numbered measure lengths - unless the user zoomed in such that the loop filled the entirety of the screen, multiple inputs would still be necessary in order to move the right marker to the left hand side without issues (depicted below).

1      2      3      4
<------------------>
        ^------ desired loop end point

What I had *thought* was being suggested earlier was somewhat the opposite train of thought, where the mean-based logic was only ever active when the viewable region was less than the entirety of the track (which would be significantly less than ideal, as well).

qnebra:
I feel like best in terms of intuitive use will be "vieweable mean". It requiered only to scroll to desired part in any zoom level, and loop points can be set without further scrolling. I think in my opinion lmms UX should be easy to discover and fun to use.

tresf:
Instead of arguing what will work, I'd recommend opening a new bug report and just describing what you want.   A picture is worth 1,000 words.  Describe the current behavior, the desired behavior and how you'll fix it.

For example, GarageBand just allows you to click and drag your region, then use shift to place left and right at the cursor.  It's intuitive, works for millions of users, doesn't require any complicated memorization of shortcuts or oddball mouse buttons that don't exist on laptops.  GarageBand also allows loop toggling based on click on/off.

I'm not stating Garageband is the holy grail of usabilby, but at some point we should just accept human interface decisions made by others that are good enough for the masses.

tresf:
> How Garageband behave when for example loop 'end point' is far away in timeline?

<kbd>Alt</kbd> modifier combined with <kbd>Shift</kbd> will override mean guessing.

ezgif-6-801bab4ed171

Xeno-Idaltu:
Whoever changed the old Loop Point features did it for its own ego.

I've messeged multiple users online and nobody likes the new Loop Point feature found on 1.3.1-Alpha

Supposedly it was done because laptops dont have MMB. Who made the changes? A GNOME user?

Everyone I know uses the mouse to produce music regardless of DAW.

Please bring the old Loop Point feature and get rid of this annoying feature that truly doesn't make sense.

PR #4044, opened 2017-12-08
Remove GUI related code from the core code

michaelgregorius:
Move GUI related code out of `Song` and into `MainWindow`.

### DIFF
--- a/src/core/Song.cpp
+++ b/src/core/Song.cpp
@@ -51,7 +51,6 @@
 #include "ProjectJournal.h"
 #include "ProjectNotes.h"
 #include "SongEditor.h"
-#include "TextFloat.h"
 #include "TimeLineWidget.h"
 #include "PeakController.h"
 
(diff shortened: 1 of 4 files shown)


michaelgregorius:
Hi @LMMS/developers,

would someone like to review these changes? They remove the following dependencies from Song:
* ExportProjectDialog.h
* ImportFilter.h
* MainWindow.h
* FileDialog.h
* TextFloat.h
* VersionedSaveDialog.h

After application of this PR the last remaining calls that use the `gui` pointer in `Song` can be found in:
* `Song::clearProject`
* `Song::loadProject`
* `Song::saveProjectFile`

I have tried to fix `Song::clearProject` with the same approach as used in some of the commits contained in this PR (add a new signal `projectCleared` in `Song` which is handled by a new slot `onProjectCleared` in `MainWindow` which does the GUI related steps) but this leads to segmentation faults when creating a new project right after the start of the application.

To remove the calls from `Song::loadProject` and `Song::saveProjectFile` we will have to improve the separation of loading and storing the GUI state vs. the "business" state, i.e. the song related data, etc.

So it seems that the other changes are somewhat harder to implement and therefore I'd like to merge these changes made so far first.

src/core/Song.cpp
```diff
@@ -667,17 +667,11 @@ void Song::stop()
...
-							( tl->autoScroll() == TimeLineWidget::AutoScrollEnabled ) )
-					{
-						gui->songEditor()->m_editor->updatePosition( MidiTime(tl->savedPos().getTicks() ) );
-					}
 					tl->savePos( -1 );
 				}
 				break;
 
 			case TimeLineWidget::KeepStopPosition:
-			default:
```
tresf: Generally speaking it should be sane for enums to have a default fall-through to prevents compilation errors if the enum is changed down the road.  I'm not sure why this was removed.  Did your code editor do this automatically?

michaelgregorius: Hi @tresf,

thanks for having a look at the code! I have removed the `default` case because QtCreator highlighted it with a warning that there is a `default` case although all cases are handled.

I currently cannot see how a `default` case that does nothing should help in this place. If you add a new enumeration value then the effect of the current code will be the same like with the `default` case. Nothing happens if the new value is set somewhere. If you rename or remove a value from the enumeration then the compiler will tell you everywhere where it was used.

However, I am not too invested in this. If it's wanted I can put the `default` case back.

tresf: > I have removed the default case because QtCreator highlighted it with a warning that there is a default case although all cases are handled.

Interesting.  Thanks.  I had a feeling this was IDE driven.  More of a query related to the behavior of the IDEs than a criticism of the decision since we're bound to see this happen more.  This enum is likely to stay fully cased due to it's design, but the `default` fall-through is often preferred for frequently changing enums.

src/core/Song.cpp
```diff
@@ -457,6 +451,18 @@ void Song::processAutomations(const TrackList &tracklist, MidiTime timeStart, fp
 	}
 }
 
+void Song::setModified(bool value)
+{
+	if( !m_loadingProject )
```
tresf: Cosmetic, but could be simplified and be more readable.  `value` is ambiguous, something representing the state is a bit more readable.

```cpp
if( !m_loadingProject && m_modified != modified) {
    m_modified = modified;
    emit modified();
}
```

tresf:
The code change seems quite sane.  I haven't hacked at the GUI code in a while, so I'll ask some others to do a proper review.  If it needs unit testing, I can arrange that though. :)

michaelgregorius:
Implemented part of the code review by pulling the condition of the second if clause into the first one. The variable is still called `value` as renaming it to `modified` would have clashed with the signal `modified()` that's emitted in the same method.

I intend to merge these changes soon.

PR #4048, opened 2017-12-10
Fix CPU meter transparency and smoothen the fade transition

Videowelder:
The original CPU meter had a faulty transparency that I fixed and also aligned the bar with the graph so that it looks better. Looks less clunky now. As for the fader, a smoother transition is less harsh and there shouldn't be chunks. I prefer sections that transition smoothly.


Umcaruje:
The fader LED's are purposefully segmented to indicate when a signal is over -6db(yellow) or 0db(red). So I don't think this change should be accepted because it's not an improvement other than personal preference. 

As for the cpu meter I'd need to get back to you on that. To be continued...

Spekular:
I agree with Umcaruje that segmented LEDs are better functionally, and also find them to be nicer aesthetically.

Mark-Agent003:
@Videowelder 's CPU background seems too pixelated and blocky, and it doesn't fit the theme. The normal one seems to be fine except for the alignment issue you mentioned.
slice1
(Videowelder 's on the left, normal one on the right)

The alignment issue can be fixed by shifting the image one pixel to the right.

Umcaruje:
I agree with @Mark-Agent003 I don't see the merit of the change of the cpu meter styling.

Umcaruje:
@Videowelder you should implement the suggested changes or I'll close out this PR.

Umcaruje:
Closed it is.

PR #4061, opened 2017-12-21
Fix some more VST bugs

### LINKED ISSUE #4052: Vestige freezes and crash LMMS with plugins that need verification
This happen to me and other people with every VST that need verification from email or enter key at open, that opens the plugin window + a secondary or primary sub window to input the key or email.

On Windows 7 64.

Happen with MAGIX Vintage Effects Suite and other plugins that need any kind of input into a secondary window.

Event: "AppHangXProcB1" on the RemoteVstPlugin.exe at freeze then crash and the Lmms crash aswell in the event "AppHangB1".

Can be tested with any VST that needs input of any kind.

Crash on 1.1.3 aswell LMMS 1.2.0-rc4

### LINKED ISSUE #4095: iZotope plugins not working in LMMS
I and another LMMS user on the Discord (howellsky) have problems with our iZotope plugins. Once we load our effects in (mine being the iZotope Ozone Imager and howellsky's being the iZotope Vinyl) nothing shows up, and the volumes on our instruments are muted. Turning the wet/dry knob on the effect in the effect list will make the volume come back, but that's about all the effect does.

@SirBothersome also had a problem with his iZotope Neutrino plugin.

DomClark:
Remove the lock synchronising access to the VST plugin's methods. The only methods which are called outside of the UI thread are `process`, `processReplacing` and `dispatcher(effProcessEvents, ...)`, all of which are intended to be called concurrently with other methods in order to facilitate real-time audio processing. This fixes a deadlock which occurs when a plugin waits on window messages sent to LMMS during an idle call.

I've also taken the opportunity to replace `lockShm()` with `tryLockShm()` in `RemoteVstPlugin::process` so as to remove all potentially blocking calls from that method.

The message `sampleRateInformation` now waits for a reply since, now we don't have a lock, audio processing shouldn't start before it's complete.

We now always tell the plugin the sample rate and buffer size on load as some plugins (e.g. Amplitube) crash if not given this information.

Fixes #4052.

The `audioMasterPinConnected` host callback returns `0` now to indicate that all pins are connected. This message is deprecated but some plugins (e.g. iZotope effects) still rely on it and won't produce sound unless told that they are connected.

Fixes #4095.

### DIFF
--- a/include/RemotePlugin.h
+++ b/include/RemotePlugin.h
@@ -419,6 +419,7 @@ enum RemoteMessageIDs
 	IdQuit,
 	IdSampleRateInformation,
 	IdBufferSizeInformation,
+	IdInformationUpdated,
 	IdMidiEvent,
 	IdStartProcessing,
 	IdProcessingDone,
@@ -807,6 +808,7 @@ public:
 	{
 		lock();
 		sendMessage( message( IdSampleRateInformation ).addInt( _sr ) );
+		waitForMessage( IdInformationUpdated, true );
 		unlock();
 	}
 
@@ -1318,9 +1320,14 @@ bool RemotePluginClient::processMessage( const message & _m )
 		case IdSampleRateInformation:
 			m_sampleRate = _m.getInt();
 			updateSampleRate();
+			reply_message.id = IdInformationUpdated;
+			reply = true;
 			break;
 
 		case IdBufferSizeInformation:
+			// Should LMMS gain the ability to change buffer size
+			// without a restart, it must wait for this message to
+			// complete processing or else risk VST crashes
 			m_bufferSize = _m.getInt();
 			updateBufferSize();
 			break;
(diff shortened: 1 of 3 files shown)


tresf:
@DomClark is this ready for merge?  I know you had a question on #devtalk about whether or not to combine a second fix.

DomClark:
Added the second fix now; the plugin concerned wasn't mentioned in the bug report here on GitHub but was mentioned by the reporter on Discord. I'll add to the issue the full list of plugins concerned.

In terms of being ready to merge, there is one potential problem I've spotted; it seems unlikely to occur but I'd like to make sure it won't. Most VST calls can be made concurrently with the audio processing calls, but updating sample rate and buffer size can't. Buffer size is set only upon init, but the sample rate is changed when exporting at a non-default sample rate, so if anyone's particularly unlucky with thread scheduling, the sample rate update message won't make it to the VST before processing begins. The result depends on the plugin; best case scenario nothing happens, but worse case scenario the plugin crashes and LMMS hangs. I'll probably address this by making LMMS wait for a reply to `IdSampleRateInformation` so processing won't begin until everything's ready.

tresf:
@DomClark per request on [Discord](https://lmms.io/chat), here's a build that contains this fix (please pardon the versioning, it's based off of RC5, not RC4)

https://github.com/tresf/lmms/releases/download/v1.2.0-rc4/lmms-1.2.0-rc4.130-win64.exe

PhysSong:
@DomClark Are there anything left before merging this?

DomClark:
I'm not aware of anything else for now, so I'm happy for this to be merged. I can always open another PR if anything new appears.

PR #4063, opened 2017-12-21
Use "Set value" as title for QInputDialog of some widgets.

Sawuare:
These widgets are the Fader, Knob and LcdSpinBox.
The QInputDialog shows when a widget from the above is double-clicked.
I propose this change because the current title `windowTitle()` usually doesn't have a specified value and defaults to `lmms`:
Before
With this change, it will always default to `Set value`:
After

### DIFF
--- a/src/gui/widgets/Fader.cpp
+++ b/src/gui/widgets/Fader.cpp
@@ -218,7 +218,7 @@ void Fader::mouseDoubleClickEvent( QMouseEvent* mouseEvent )
 	// TODO: dbV handling
 	if( m_displayConversion )
 	{
-		newValue = QInputDialog::getDouble( this, windowTitle(),
+		newValue = QInputDialog::getDouble( this, tr( "Set value" ),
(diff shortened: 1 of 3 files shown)


tresf:
This looks good.  In regards to the way we're handling this, I feel we may be able to consolidate the redundant strings.  @liushuyu any thoughts?

Wallacoloo:
1. Given how both `LcdSpinBox` and `Fader` _already_ duplicate the `tr( "Please enter a new value between %1 and %2:" )` string when calling QInputDialog, I think it's not that much worse duplicating the other strings for now.

2. The easiest way to consolidate these calls is to create a new method `getInput` on the `ModelView` class, which creates and populates a `QInputDialog`. I'll try to look into it this week.

Wallacoloo:
I went ahead and merged this as per my point 1 above. I still plan to consolidate these - but when I took a first pass at it today it turned out there's some cleanup around ModelViews/Models required to make the consolidation work well, so it's going to have to happen incrementally.

PR #4065, opened 2017-12-21
Move Zyn GUI to main thread

### LINKED ISSUE #703: ZynAddSubFX "Show GUI" broken (Apple)
The ZynAddSubFX instrument plugin plays audio in OSX, however the "Show GUI" button crashes the software.

The crash seems to be caused within the `initPlugin()` or possibly the `reloadPlugin()` code although this could be a missing library reference as well. ([view initPlugin() code here](https://github.com/LMMS/lmms/blob/stable-1.0/plugins/zynaddsubfx/ZynAddSubFx.cpp#L424))

[Click here for the list of libraries that load when lmms is executed](https://gist.github.com/tresf/136df4fb632283345976)
[Click here for a partial stacktrace of the crash](https://gist.github.com/tresf/96dd4d45acbf6917245d)

Some outstanding `QSharedMemory` bugs:
https://bugreports.qt-project.org/browse/QTBUG-27765
https://bugreports.qt-project.org/browse/QTBUG-5123

I notice fltk isn't in the list when LMMS loads.  I believe that is normal, as only the Remote process should load it.  If I had a way of launching ZynAddSubFX without LMMS it might shed some more light on this issue as even a `cout` debug added to the code won't echo to terminal as it's a separate process.

image

tresf:
Closes #703

We'll need some solid testing on Windows and Linux to make sure no regressions are caused as a result of this change.

### DIFF
--- a/plugins/zynaddsubfx/RemoteZynAddSubFx.cpp
+++ b/plugins/zynaddsubfx/RemoteZynAddSubFx.cpp
@@ -62,18 +62,11 @@ public:
 		waitForMessage( IdInitDone );
 
 		pthread_mutex_init( &m_guiMutex, NULL );
-		pthread_create( &m_guiThreadHandle, NULL, guiThread, this );
+		pthread_create( &m_messageThreadHandle, NULL, messageLoop, this );
 	}
 
 	virtual ~RemoteZynAddSubFx()
 	{
-		m_guiExit = true;
-#ifdef LMMS_BUILD_WIN32
-		Sleep( m_guiSleepTime * 2 );
-#else
-		usleep( m_guiSleepTime * 2 * 1000 );
-#endif
-
 		Nio::stop();
 	}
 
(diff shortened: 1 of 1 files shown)


PhysSong:
As the author of this patch, I'd like to say that it should be fine on every platforms. The reasons are:
1. It is safe to execute the message loop in any thread since it doesn't call UI functions or thread-dependent functions(as far as I know).
2. It is normal to run UI loop in main thread. It's safe, too.
3. I've changed the synchronization logic a little bit. Now the message processing thread always finishes before the GUI loop. So there should'n be any threading issues.

Anyway, testing is always good. :+1:

tresf:
Here's a build for testing this out: https://github.com/tresf/lmms/releases/download/v1.2.0-rc4/lmms-1.2.0-rc5.3-mac10.13.dmg

tresf:
We've extensively tested this on Mac, Linux and Windows.

Here's some evidence that the Zyn UI can remain open and stable with this PR despite several preset changes.  Worth noting... I was able to get it to crash twice, but I cannot reproduce now despite how hard I try.  When it crashed, the stack trace did not point to the UX, so @PhysSong suspects I had found an existing edge case that already exists.


untitled

PR #4067, opened 2017-12-23
Add better Qt5 detection

tresf:
Making a cmake project with Qt5 dependencies can vary in difficulty depending on how Qt5 was obtained.

For example:

* On Linux, Qt5 is usually offered from a package manager, such as `apt` or `dnf`, but it's also quite common to install Qt5 manually into `/opt/`.
* On Mac, Qt5 is usually fulfilled by `homebrew` or `macports` however it is also fairly common to install from the QtCreator desktop installer.
* On Windows, Qt5 is usually installed directly from the QtCreator desktop installer, but may also be obtained from msys2.
    * Windows makes this even harder because QtCreator comes bundled with various C++ compatible versions including mingw, MSVC 2015 32-bit, MSVC 2015 64-bit, etc.  You need to know your exact compiler version and manually add that each time you build a project using cmake.

What this PR aims to do is allow `cmake ..` to find Qt5 without a hassle on the most common build environments.  This means the average developer should never need to type `brew --prefix qt5` or `export CMAKE_PREFIX_PATH=/path/to/qt5` again, although that "normal" technique will still remain to be supported.

**Usage**
```cmake
INCLUDE(FindQt5All)
FIND_PACKAGE(Qt5Core REQUIRED)
```

**Progress**

| OS       | Source | Module | Working |
|---------|--------|-----------|-----------|
| Linux   | from package manager | `FindQt5Linux.cmake`      | |
| Linux   | from 3rd-party ppa |   `FindQt5Linux.cmake`      | |
| Linux   | from qt.io |   `FindQt5Linux.cmake`      | |
| Windows | from QtCreator + MSVC | `FindQt5Windows.cmake`      |✅|
| Windows | from QtCreator + mingw|  `FindQt5Windows.cmake`      | |
| macOS | from QtCreator | `FindQt5Apple.cmake`      |✅|
| macOS | from homebrew |`FindQt5Apple.cmake`      | ✅|
| macOS | from macports | `FindQt5Apple.cmake`      | |

**Additional Modules**

Some detection techniques include parsing of command output (e.g. `EXECUTE_PROCESS(...)`), so an additional module has been added `EnglishLocale.cmake`.

Usage:

```cmake
INCLUDE(EnglishLocale)
USE_ENGLISH_LOCALE()  # SET COMMANDS TO USE ENGLISH STANDARD OUTPUT
EXECUTE_PROCESS(...)
USE_DEFAULT_LOCALE()  # RESET COMMAND ENVIRONMENT BACK TO NATIVE LOCALE
```

This PR is a work in progress and should be merged once the most common platforms have been completed.


PhysSong:
@tresf What's the status of this PR?

tresf:
> @tresf What's the status of this PR?

Stalled. 😄 

Seriously though:

1. There are unrelated changes in here, like the SDL and FFTW include dirs and the AppVeyor stuff.  They can probably be removed.
2. Someone that actually develops on Windows using `master` should help with the various Qt detection environments.  Does vcpkg handle Qt detection?  How do people do it?
3. macOS really needs this PR as it's stupid to keep typing `export CMAKE_PREFIX_PATH=$(brew --prefix qt5)` every time I build.
4. Linux, I have no idea and it's probably a good candidate for the back burner considering Linux often just uses the repo version of Qt (versus Win/Mac which download it manually most of the time).

I don't have the time to dedicate to this right now, but the code is good, even if as a reference point for someone else to run with hit.

tresf:
> @tresf is there anything of use here anymore?

If the question is whether or not this will be merged, the answer is "probably not".
If the question is whether or not we'll use this code, we probably will.

I came back to this last week and decided that it's due for a refactor, especially for Qt6.

tresf:
Superseded by https://github.com/LMMS/lmms/pull/7699.

PR #4080, opened 2017-12-29
3rdparty/cmake: Do not reset CMAKE_C(XX)_FLAGS

schnitzeltony:
Faced this after update rc4 -> rc5. If accepted: Patch applies on master cleanly.

### DIFF
--- a/src/3rdparty/CMakeLists.txt
+++ b/src/3rdparty/CMakeLists.txt
@@ -1,6 +1,7 @@
 # TODO: Move submodules to dedicated directories for flag control
 IF(QT5 AND LMMS_BUILD_LINUX AND WANT_VST)
 	set(BUILD_SHARED_LIBS OFF)
+	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-deprecated")
 	add_subdirectory(qt5-x11embed)
 ENDIF()
 
--- a/src/3rdparty/CMakeLists.txt
+++ b/src/3rdparty/CMakeLists.txt
@@ -1,3 +1,4 @@
+# TODO: Move submodules to dedicated directories for flag control
 IF(QT5 AND LMMS_BUILD_LINUX AND WANT_VST)
 	set(BUILD_SHARED_LIBS OFF)
 	add_subdirectory(qt5-x11embed)
--- a/src/3rdparty/CMakeLists.txt
+++ b/src/3rdparty/CMakeLists.txt
@@ -1,6 +1,3 @@
-set(CMAKE_C_FLAGS "")
-set(CMAKE_CXX_FLAGS "")
-
 IF(QT5 AND LMMS_BUILD_LINUX AND WANT_VST)
 	set(BUILD_SHARED_LIBS OFF)
 	add_subdirectory(qt5-x11embed)
--- a/src/3rdparty/rpmalloc/CMakeLists.txt
+++ b/src/3rdparty/rpmalloc/CMakeLists.txt
@@ -1,4 +1,4 @@
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c11")
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c11 -Wno-unused-variable")
 
 add_library(rpmalloc STATIC
 	rpmalloc/rpmalloc/rpmalloc.c
--- a/src/3rdparty/rpmalloc/CMakeLists.txt
+++ b/src/3rdparty/rpmalloc/CMakeLists.txt
@@ -1,4 +1,4 @@
-set(CMAKE_C_FLAGS "-std=c11")
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c11")
 
 add_library(rpmalloc STATIC
 	rpmalloc/rpmalloc/rpmalloc.c


tresf:
@schnitzeltony thanks.  I think @lukas-w's idea here was to not impose LMMS cmake flags onto modules that we do not directly code or support.  I'll wait for @lukas-w to decide as he's the author of this particular line of code.

lukas-w:
> I think @lukas-w's idea here was to not impose LMMS cmake flags onto modules that we do not directly code or support

Correct. To keep this behaviour, we could move `3rdparty` to the top-level directory and include it in `CMakeLists.txt` before adding LMMS-specific flags.

tresf:
> We could move `3rdparty` to the top-level directory and include it in `CMakeLists.txt` before adding LMMS-specific flags.

I think this is the most logical way to solve this problem.  An alternative is to switch LMMS to use a non-standard variable for `CMAKE_C(XX)_FLAGS` and use that in addition to the built-in ones.  I'm not sure which is more intuitive and organized.

PhysSong:
@lukas-w Do we need to keep this open?

lukas-w:
@PhysSong I think we can merge and safely revert if it causes any troubles. Moving `3rdparty` to top-level should happen in master.

PhysSong:
Travis-CI fails because of `-Werror` passed to `rpmalloc`. I think we should store the passed C/C++ flags before adding LMMS specific ones and use them for 3rd-party sources.

PhysSong:
We now have three solutions:
- Move `3rdparty` to the top level: This looks like the cleanest solution, but @lukas-w suggested doing that in `master`, not in `stable-1.2`.
- Backup original build flags before adding LMMS-specific ones: Looks a bit less clean than the first solution, but it's still small.
- Store LMMS-specific flags in a separate variable: We should manually add those flags wherever required, so it's not that clean.

If the first method is not suitable for `stable-1.2`, I'll fix the problem with the second method.

he29-net:
@PhysSong Since stable-1.2 is now out in the wild, it probably does not make much sense to make any changes to the `3rdparty` stuff, since we would only risk breaking it for no good reason.

So this could now be re-targeted to `master`, meaning that there is also no need to consider the alternative solutions, as you mentioned that moving `3rdparty` to the top level is the best and cleanest one.

PhysSong:
I think the issue is relevant to packagers who need to pass proper flags, but not for ordinary users. If it is considered critical, we need to take one of the alternative methods. If not, we can move to the `master` solution.

tresf:
One idea is to simply add a new variable (i.e.`CMAKE_CXX_FLAGS_ <SUBFOLDER>`, e.g. `CMAKE_CXX_FLAGS_ RPMALLOC`) which a build system would be able to override modularly.

```cmake
// Pseudo-code, untested
MACRO(ADD_SUBDIRECTORY_WITH_FLAGS directory)
  FILE(GLOB children RELATIVE ${directory} ${directory}/*)
  SET(dirlist "")
  FOREACH(child ${children})
    IF(IS_DIRECTORY ${directory}/${child})
      STRING(TOUPPER "${child}" CHILD_UPPER)
      SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS_{${CHILD_UPPER}}")
      SET(CMAKE_C_FLAGS "${CMAKE_CX_FLAGS_{${CHILD_UPPER}}")
      ADD_SUBDIRECTORY(directory)
    ENDIF()
  ENDFOREACH()
ENDMACRO()
```

... and then ... 

```diff
- ADD_SUBDIRECTORY(rpmalloc)
+ // Flags will be honored as CMAKE_CXX_FLAGS_RPMALLOC
+ ADD_SUBDIRECTORY_WITH_FLAGS(rpmalloc)
```

PhysSong:
> Moving `3rdparty` to top-level should happen in master.

How about moving the call to `add_subdirectory()` to the top-level `CMakeLists.txt` while leaving the `3rdparty` directory as-is? In that way, we can make the change minimal. I can add a TODO comment as a reminder.

PhysSong:
@lukas-w @tresf Any opinions on newly proposed ideas?

lukas-w:
> How about moving the call to `add_subdirectory()` to the top-level `CMakeLists.txt` while leaving the `3rdparty` directory as-is?

Sounds reasonable to me :+1:

tresf:
There are a few needs here... 
* One is the possibility that us or downstream want to explicitly control flags on a per-module basis and do not have the ability to.  This is a reasonable scenario to code for, but no one's asked for it.   Since this is a bug-fix only branch, anything which provides more than the bug-fix isn't desired.
* Another need -- likey the motive of the OP -- is to be able to control flags on a whole-build system perspective, but we wipe them out.   The solution which moves `3rdparty` up fixes this problem exactly, but it doesn't really offer any documentation and it's also confusing because other `3rdparty` modules may want these flags.  I fear that someone coming back a year later asking again -- making the code harder to maintain.

I believe the best approach here is to explicitly remove the problematic flags on a per-module basis.  This approach both documents why they shouldn't be there to begin with -- as well as allows them to propagate and be useful  by default.

The argument that we should remove all warning flags on code we don't author is valid, but in practice, we've contributed quite a bit of compiler bug fixes upstream to the likes of `swh/ladspa` and friends.  Keeping 3rdparty flags is beneficial for modules which are historically maintained.  Adding these back in on a per-module basis doesn't seem reasonable.

So my proposed solution is to keep the OP's proposal, while explicitly removing these warning flags. I'll commit directly to the OP's branch as to keep the commits and conversation coupled.

tresf:
`rpmalloc` is passing now, but `qt5-x11embed`  is broken on macOS due to use of deprecated functions.  https://travis-ci.org/github/LMMS/lmms/jobs/682384033

@lukas-w do you have a recommendation?  I believe `-Wall` is the appended (not prepended) to the flags, so we can't modify it downstream, right?

tresf:
From gcc:

>  The combined effect of positive and negative forms is that more specific options have priority over less specific ones, independently of their position in the command-line.

So I can add it in, but it'll pollute `rpmalloc`'s variable.  I'm ok with this, it'll be cleaned up if the `TODO:` is addressed in `master`.

tresf:
Travis is passing.  Merging.

PhysSong:
FYI, this is what I was doing:
https://github.com/PhysSong/lmms/commit/36e65e53fc928ee245715d828ff8117bb5056bf2

tresf:
@PhysSong, I didn't mean to waffle-stomp you.  I was assigned to speed up the closure of 1.2 tasks.  If there's a flag that's unwarranted on a particular platform, I can make amends.

PhysSong:
> but `qt5-x11embed` is broken on macOS due to use of deprecated functions.

Those warnings are from LMMS itself. Re-running the build should have fixed the failure. At the end of the build log:
> The job exceeded the maximum log length, and has been terminated.

PhysSong:
> I was assigned to speed up the closure of 1.2 tasks.

@tresf I understand, but please ask before merging next time. I was unable to react because I was sleeping.

tresf:
The PR has been open for 2 years and it's holding up real work.  If there's a problem point it out.

> Those warnings are from LMMS itself. Re-running the build should have fixed the failure. At the end of the build log:

I had them locally.

PhysSong:
What I meant was your wrong comment:
> `qt5-x11embed` is broken on macOS

It's Linux-only.

tresf:
It broke locally, I'm on MacOS.  I can see that the job log length is what actually broke Travis-CI and I appreciate you pointing that out, but I'm uncertain what restarting the build would accomplish?

Furthermore, my apology was because we were both working on it, I see no evidence you were activity looking at it.  "Please ask" is presumptuous.  If this causes issues, they become my issues.  If you don't believe the code is correct please focus on that.  I was asked to help and did.  The commit is small, addresses the issues preventing merge and is easy to understand and read.  If there's a technical issue with it, please share so it can be improved.  I really don't appreciate my apology of coincidence being answered with a workflow response.  If you want to revert and own this, go ahead.

PhysSong:
> I'm uncertain what restarting the build would accomplish

`ccache` will effectively continue the build from the last failures.
> I see no evidence you were activity looking at it.

https://github.com/LMMS/lmms/pull/4080#issuecomment-612724261, though you may consider it not as an activity.

I was thinking about more submodules in `master` and wanted to handle the underlying issues consistently across branches.
>  If you want to revert and own this, go ahead.

For `stable-1.2`, I won't touch it. For `master`, however, I will move to another fix.

tresf:
> ccache will effectively continue the build from the last failures.

Thanks for that information however stable releases intentionally disable ccache.  This sounds like a timebomb.

tresf:
Ok, I see the mistake, I'll correct it.

tresf:
Corrected, thanks.  You're right, the spamming of messages made me jump to conclusions and assume a library that wasn't being loaded was the culprit.  In hindsight, I'm unsure why my local build failed, but it's fixed.

PR #4088, opened 2017-12-31
Using the cmake variable will insure the proper value

devnexen:
regardless of the platform (already available in 2.0.7 version)


tresf:
Thanks, didn't know this existed.  Can be merged once Travis-CI passes. 👍

Issue #4077, opened 2017-12-28
Plugin embedding - using xembed protocol qt5 crash on linux


Rikislav:
Linux Ubuntu 16.04 - QT 5.9.1
When using VesTige with the xembed protocol option lmms crashes with this message:

`/lmms: symbol lookup error: /plugins/libvstbase.so: undefined symbol: _ZN18QX11EmbedContainerC1EP7QWidget`

PhysSong:
Are you using AppImage, or manually compiled version?
Tagging @lukas-w.

Rikislav:
I'm compiling directly, through QtCreator. Sending a screenshot
![screenshot at 2017-12-29 00 35 19](https://user-images.githubusercontent.com/13100084/34425840-677e0884-ec30-11e7-8419-33c502cdbc5d.png)

PhysSong:
@Rikislav Could you change makefile generator to `Unix Makefiles` or similar?

Z3R0C:
i get the same crashes (same message) with some vst's...if i'm switch to "no embed" option, i get a black empty window and a load average allround 25-30. it's like a freeze. all this testet vst's (zampler/trickster, a.n.a, dune2, nexus2) was working with RC4.  

ubuntu 17.10, qt 5.9.1, compiled with GCC

Rikislav:
Ok, I changed compiler to Unix Makefile, and in LMMS selected the XEmbed protocol, but there is no noticeable change when using a vst dll:

Starting using X11Embed protocol.
RemotePlugin::DebugMessage: inputs: 0  output: 8
RemotePlugin::DebugMessage: creating editor
RemotePlugin::DebugMessage: editor successfully created
/home/riki/git_from/lmms_debug/lmms: symbol lookup error: /home/riki/git_from/lmms_debug/plugins/libvstbase.so: undefined symbol: _ZN18QX11EmbedContainerC1EP7QWidget
/home/riki/git_from/lmms_debug/lmms exited with code 127

The RemoteVstPlugin.exe.so and wineserver remain among running processes and eventually lead to a general crash

Sorry for such a delay in answering, busy period.

qnebra:
I have same issue, but with my own compilled version. VST plugins under official RC5 version works fine.

PhysSong:
@lukas-w I don't know if this is related, but should this be `IF(WANT_QT5 AND LMMS_BUILD_LINUX)`?
[...]

PR #3987, opened 2017-11-19
Upgrade Calf plugins to 0.90

tresf:
The calf plugins we ship are very outdated (~0.18) yet 0.90 is available.  This PR adds the latest.

**Download**

https://github.com/LMMS/lmms/pull/3987#issuecomment-377959431

Or to compile locally...

```bash
git checkout -b tresf-veal master
git pull https://github.com/tresf/lmms.git veal
git submodule deinit -f plugins/LadspaEffect/calf/veal
git submodule update --init --recursive plugins/LadspaEffect/calf/veal
```

**Background**

Calf has removed support for LADSPA, however generous work from @JohannesLorenz, now LADSPA support on 0.90 is finally possible.

The efforts are currently living in `tresf/veal` on a branch called `veal_with_0.90`.  These changes won't ever show in this PR, but will instead be in a separate PR against `lmms/veal` once ready (then we update the submodule here, then we merge), but since few people will read that, I'm posting the high-level changes here.

**What's Changed?**

Below is a list of what's removed and what's added in Calf 0.90.  Note, a removed plugin may be due to a name change.  We'll know more after a proper investigation.

* `WONTFIX` - Input gains `MIN: 0` changed to `MIN: 0.01625`

Plugin<br>(new in bold) | Sound Tested<br>(new premarked) | Notes
----------------------|------------------------------------------|----
**Calf Analyzer** |☑️
Calf Bass Enhancer |   ✅ 
**Calf Compensation Delay Line**  | ☑️
Calf Compressor |  ✅ | New "Mix" knob
**Calf Crusher** | ☑️
Calf Deesser | ✅ 
**Calf Emphasis** | ☑️
Calf Equalizer 12 Band | ✅  | [Display issues](https://user-images.githubusercontent.com/6345473/33529138-51ce4512-d83a-11e7-802d-62f8f5a5fd90.png) (fixed)
**Calf Equalizer 30 Band**  | ☑️
Calf Equalizer 5 Band | ✅ | Same display issues as 12 band (fixed)
Calf Equalizer 8 Band | ✅ | Same display issues as 12 band (fixed)
Calf Exciter | ✅ 
Calf Filter | ✅ | New "Bypass" checkbox, "Input Gain" and "Output Gain" knobs
Calf Filterclavier |  ✅ | New "Bypass" checkbox, "Input Gain" and "Output Gain" knobs
[...]

PR #4089, opened 2018-01-01
added the function randsv to Xpressive

gnudles:
that function let you choose the random seed thus allowing persistent vector upon note plays and upon waveforms transit in the gui.


tresf:
Before merging this, @ArashPartow can you help us decide which commit to use for our submodule?  We can no longer find https://github.com/ArashPartow/exprtk/commit/aad301a6ebbab769620b0bb0e77713c4659117a6.

ArashPartow:
@tresf  Would it be possible to instead of using a particular revision of exprtk module to always use the tip of the master?

tresf:
@ArashPartow thanks kindly for chiming in with a reply.  This is probably a better place than #4077 for this discussion so I'll paste my reply here again.

In our build system, we could coerce this, but no, git submodules do not directly permit this.  Is there a particular reason why the history was erased?  This is generally uncommon for a master branch and it also leaves a bit of ambiguity as to what occurred for a particular update.

Worse, it takes reliable builds completely out of our hands as we wouldn't know if a breaking change existed until after our build system broke.  It's not something I would encourage or want.

For now, our code pulls from [my fork](https://github.com/LMMS/lmms/commit/9547575c502b46857272018d768eae6234a6edd2) which has the history restored .  This fork is only around until we understand if there's a viable and reliable way to use a submodule against upstream.  👍

tresf:
@gnudles now that Travis is fixed there seem to be some code fixes in store, here's one of the errors from Travis-CI:

```
In file included from /home/travis/build/LMMS/lmms/plugins/Xpressive/Xpressive.h:38:0,
                 from /home/travis/build/LMMS/lmms/plugins/Xpressive/Xpressive.cpp:25:
/home/travis/build/LMMS/lmms/plugins/Xpressive/ExprSynth.h:56:45: error: ‘numeric_limits’ was not declared in this scope
```

ArashPartow:
@tresf  with regards to the above numeric_limits issue, look like all that is required is to add `#include <limits>` in the ExprSynth.h header.

tresf:
@ArashPartow done and done.  @gnudles sorry, had to fast-forward your branch, you'll have to wind it back a few commits `git reset --hard HEAD~3` and then `git pull`.

tresf:
Build still failing due to `numeric_limits`.  I'll defer to @gnudles I'd rather leave the coding to him, I'm better off sticking to the build environment. :)

```
/home/travis/build/LMMS/lmms/plugins/Xpressive/ExprSynth.h:57:45: note: suggested alternative:
In file included from /home/travis/build/LMMS/lmms/plugins/Xpressive/ExprSynth.h:30:0,
                 from /home/travis/build/LMMS/lmms/plugins/Xpressive/ExprSynth.cpp:26:
/opt/mingw32/i686-w64-mingw32/include/c++/4.9.4/limits:304:12: note:   'std::numeric_limits'
     struct numeric_limits : public __numeric_limits_base
            ^
In file included from /home/travis/build/LMMS/lmms/plugins/Xpressive/ExprSynth.cpp:26:0:
/home/travis/build/LMMS/lmms/plugins/Xpressive/ExprSynth.h:57:60: error: expected primary-expression before 'float'
  static const int max_float_integer_mod=(1<<numeric_limits<float>::digi
```

ArashPartow:
@tresf fair enough. Perhaps to help @gnudles along, from what the diagnostic is saying it looks like:

- If the build is not in C++11 or after, the **digits** type from numeric_limits will not be available
- If  there's no `using namespace std` then it should be `std::numeric_limits`
- Or it might simply require brackets like so: `(std::numeric_limits<float>::digits)`

gnudles:
@tresf , I hope I didn't messed things up too much, I recreated the fixed commit on the top of lmms:master. I added "std::" before numeric_limits.

gnudles:
BTW, did you know you can use this plugin to play non-western music?
you can set the difference between the keys to be quarter tone like this:
sinew(integrate(440*cent(50*(key-bnote))))

tresf:
@gnudles is this ready to merge?

gnudles:
It looks OK to me, but I didn't tested it yet cause I somewhat can't compile lmms anymore after upgrading ubuntu. it keeps complain on missing XCB in cmake...

PhysSong:
> it keeps complain on missing XCB in cmake...

Can you try `git submodule update --init --recursive? Alternatively, you may install `extra-cmake-modules` instead.

tresf:
@gnudles I think you need:

```bash
sudo apt-get install qtbase5-private-dev libxcb-util0-dev libxcb-keysyms1-dev
```

gnudles:
thanks, now it works. BTW I also get that message:
`Fetched in submodule path 'plugins/LadspaEffect/calf/veal', but it did not contain 816b4d2b7cc20429faf96863d7e8ba1cda48ace8. Direct fetching of that commit failed.
`
so it looks like that commit has been deleted

tresf:
> Direct fetching of that commit failed.

Interesting.  If it compiles, ignore it.  It probably got confused when we renamed the branch from `veal/ladspa` to `veal/ladspa-0.18` in preparation for #3987.

PR #4128, opened 2018-01-23
Remove "What's This?" and update tooltips.

### LINKED ISSUE #896: More texts for "What is this"
The fine new feature "What is this" has a lot of items that need to written. I will do my part, if wanted. We just need to get a plan for doing them, so none is done 'trice'ish'
Just let me know.
an also, -Translations .. woo - How many languages? 
ooooooooooooo
Oki taking some control now
I will devide the objects that need WITs added in 3 categories
- Instruments
- Panels
- Effects 

Look in thread if some has 'taken' an object and are working on it. I will update the categories as WITs are made, and add 'done' WITs in order. starting here:
## 

**WITs ALREADY done:** 

**General Instrument** This is identical on all non-vst presets, and should have own datastructure or class -whatever fits architechture best..
Tab ENV/LFO

**Specified Instruments** (of cause only the PLUGIN tabs need to have specified wit's!)
Monstro

**Panels**
Automation-editor
BB-editor
FX-rack
General instrument button
LFO-controller
LFO-rack
Main-screen
Piano-roll
Song-editor

**Effects**

### LINKED ISSUE #4119: BUG: The "Info" button does not work with the Controller Rack button in the 2nd row
The "i" button does not work with the Controller Rack button in the 2nd row.
(It works with all 6 other buttons. Thus, technically this is a bug, but not worth much.)

In any case the "Info" button seems to be redundant: It does not add any value!
It just shows a tooltip when the mouse is hovered over buttons. 
Well, that same tooltip is shown even without clicking on this button first!

It does not work on the other elements of the screen, either. So I guess that it is yet to be implemented fully (?) If so, a suggestion: When the user clicks on the "i" button first, let LMMS show an **extended tooltip** instead of normal (short) tooltip.

Sawuare:
i
Removes the "What's This?" feature which uses the [QWhatsThis class](https://doc.qt.io/qt-5/qwhatsthis.html), and removes its text.
The removed translatable strings are useful and should be considered when writing the documentation/manual.

After removing the "What's This?" feature, some widgets don't have any help text (tooltips) and some have vague tooltips, so existing tooltips will be updated and new ones will be created.
Tooltips progress:
- [x] Main window
- [x] Editors:
    - [x] Song editor
    - [x] Piano Roll editor
    - [x] Automation editor
- [x] Time line
- [x] Controller Rack (LFO Controller)
- [x] Plug-ins (mostly instruments)
- [x] Track and instrument

Per https://github.com/LMMS/lmms/issues/4119#issuecomment-358241006.
Fixes #4119 and closes #896.


plugins/FreeBoy/FreeBoy.cpp
```diff
@@ -467,39 +467,30 @@ FreeBoyInstrumentView::FreeBoyInstrumentView( Instrument * _instrument,
...
 	m_ch1SweepTimeKnob->move( 5 + 4*32, 106 );
 	ToolTip::add( m_ch1SweepTimeKnob, tr( "Sweep Time" ) );
-	m_ch1SweepTimeKnob->setWhatsThis( tr( "The amount of increase or"
-									" decrease in frequency" ) );
 
 	m_ch1SweepRtShiftKnob = new FreeBoyKnob( this );
 	m_ch1SweepRtShiftKnob->setHintText( tr( "Sweep RtShift amount:" )
 										, "" );
 	m_ch1SweepRtShiftKnob->move( 5 + 3*32, 106 );
 	ToolTip::add( m_ch1SweepRtShiftKnob, tr( "Sweep RtShift amount" ) );
```
Umcaruje: These tooltips are very vague, I think the whatsthis text(or a shorter version of it) should become the tooltip text.

musikBear:
FYI i made *whatsthis* text for all missing 1.0 component, but could not add it, because i had no ..? fork.. in a branch.. of tarballs.. ?.. 🤡 
* Git did not support xp
* Noone was willing to do the actual update
I still have all sentences

tresf:
@musikBear thanks for the information, perhaps you can help review this PR instead to make the tooltip better for users.

musikBear:
@tresf sure But what PR exactly do you mean?  - and @everyone just say if the list of WITs is needed now, i will upload the 10 pages.

tresf:
@musikBear this is a Pull Request a.k.a "pull", it has code changes that are proposed.

You can see the code changes by clicking "Files" tab on the top.  In this case there's only one commit, so you can also comment on each change if needed.  This is done by clicking the `+` next to the change.  Here's an example:

image

plugins/FreeBoy/FreeBoy.cpp
Sawuare: Yes. Many tooltips will be updated, and created.

plugins/monstro/Monstro.cpp
```diff
@@ -1465,29 +1465,12 @@ MonstroView::MonstroView( Instrument * _instrument,
...
-										"audible operator (oscillators 1-3) has 3-4 properties that can be "
-										"modulated by any of the modulators. Using more modulations consumes "
-										"more CPU power. \n\n"
-										"The view is divided to modulation targets, grouped by the target oscillator. "
-										"Available targets are volume, pitch, phase, pulse width and sub-osc ratio. "
-										"Note: some targets are specific to one oscillator only. \n\n"
-										"Each modulation target has 4 knobs, one for each modulator. By default "
-										"the knobs are at 0, which means no modulation. Turning a knob to 1 causes "
-										"that modulator to affect the modulation target as much as possible. Turning "
-										"it to -1 does the same, but the modulation is inversed. " ) );
```
zonkmachine: Monstro makes up some 10% of the lines and it's good instructions. It gives a hint to what it could look like if the WhatsThis function is given some tender loving care instead.

zonkmachine:
I think for 1.3 it's better to:
1. Just remove the info knob but keep the context menu item.
2. Improve the strings.

plugins/monstro/Monstro.cpp
Sawuare: The instruction here are definitely good, but they are a lot (139 lines) and in the source code (not a good place).
The best place for them is a dedicated article on Monstro in the wiki, which already exists and is in a pretty good state: https://lmms.io/wiki/index.php?title=Monstro

musikBear:
@tresf Understood
I must have missed this. Killing the WITs *feature* as such, is imo not good. But tiswhatitis.
That aside i think that in most incidences the WITs are much better than the tool-tips. My thought would then be to do this pull inverted.
That is: 
* preserve WITs text, with a tiny change
* refurbish these changed WITs as new tool-tips
* kill all old tool-tips
An example:
**Current block:**
```
Knob *dk = new sidKnob( this );
 		dk->setHintText( tr("Decay:") , "" );
 		dk->move( 7 + 28, 114 + i*50 );
-		dk->setWhatsThis( tr ( "Decay rate determines how rapidly the output "
-				"falls from the peak amplitude to the selected Sustain level." ) );
```
 
**new block Proposal 1:**
```
Knob *dk = new sidKnob( this ); 		
 		dk->move( 7 + 28, 114 + i*50 );
-		dk->setHintText( tr ( "Decay. -The rate determines how rapidly the output "
-				"falls from the peak amplitude to the selected Sustain level." ) );
```

Proposal one has a serious flaw.
There should newer be string-information in source-code
For that reason, i would propose a new class
public class ToolTips
This class would contain ALL tooltips for ALL features

This leads to final and imo best way: (only pseudo im afraid..)
**new block Proposal 2:**
```
Knob *dk = new sidKnob( this );
 		dk->setHintText( tr(ToolTips.sid.knobDecay.hintTx));
 		dk->move( 7 + 28, 114 + i*50 );
```

Ofcause *ToolTips.sid.knobDecay.hintTx* would be 	
```
 "Decay. -The rate determines how rapidly the output "
				"falls from the peak amplitude to the selected Sustain level."
```

This approach has imo only advantages : 
1) the first word has the functionality of the previous used tool-tip, so it IS a tooltip -The rest of the sentence can be ignored 

2) Reading the whole sentence has deep information value, much like WITs

3) All strings are in one class, and are very easy to maintain and change if/ when needed.
This makes for a cleaner source-code, and wont bloat the classes size.

tresf:
@musikBear thanks.  Architecturally, we'll be keeping ToolTips inside the GUI classes in which they represent.  Content-wise your points are spot on.  It sounds like @Sawuare has a good handle on this.  Perhaps a second look once he's completed his work would be more warranted at this point.

I see a few comments about this being "Removal of functionality" and that's partially correct, but as a GUI developer for other technologies, this is a foreign concept and one that has received varying quality over the years.

I strongly agree @Sawuare that the source code is not a good place for this much text.  Our software should be fairly quick and easy to translate and this is a step in that direction.

I think part of the observation here is that some of our synths are so complicated that we need a novel to use them.  I don't think the CPP code is the place for this novel.  We don't do this for our largest synth -- Zyn, I really don't think we should do it for the others.

musikBear:
> Architecturally, we'll be keeping ToolTips inside the GUI classes in which they represent.
Im really puzzled why you want that?
Bulky strings in source-code is bad!
I feel that my argumentation for a string-class has been presented with enough quality, that at least some real argumentation for preserving current 'in-class' structure, is in place?
Besides that, yes i will do a review asa im asked to :)

@Sawuare I attach 10 pages of WITs. You are welcome to use or dispose as you like.
They are for all **main windows** components
[WITs Main.txt](https://github.com/LMMS/lmms/files/1668581/WITs.Main.txt)

I had a thought about the actual tool-tip.
Is it possible to 'divide' the tool-tip in two tempi?
If so, then the short real tool-tip could be shown first, and then ~2 scd later, the whole WIT-like text. Its just a thought..

Spekular:
I would argue against tined tooltips:

1. If the user wants to read the short text, they're put under time
pressure.
2. If the user wants to read the long text, they have to wait.

If the short tooltips differ in length to a significant degree, a single
timer would work even worse. If we're offering two texts (which as far as I
can tell we don't want to do as we'd rather use the wiki) it makes more
sense to offer a "What's This?" in the context menu.

Speaking of the wiki & the context menu, it might make sense to offer links
to the wiki from within LMMS to replace the What's This text? If this is
implemented as a call to the default browser* then LMMS doesn't need to
access the internet itself.

*I'm not sure what to call this, but for example links on webpages can
launch Spotify (open/play track) or Discord (join server) and IIRC that can
happen in the opposite direction as well.

tresf:
> Bulky strings in source-code is bad!
> I feel that my argumentation for a string-class has been presented with enough quality, that at least some real argumentation for preserving current 'in-class' structure, is in place?

I'm sorry, but please don't back-seat code here.  You're bringing this off topic.

tresf:
> Is it possible to 'divide' the tool-tip in two tempi?

Yes, but if were were to show larger tooltips, it would make more sense to do an active [help-area](https://user-images.githubusercontent.com/6345473/35455614-a07ca66e-02a0-11e8-825b-d793ca1f33d6.png) like Ableton does, which is really out of scope of this PR and would fit better as part of the single-window-interface initiatives.

plugins/sid/sid_instrument.cpp
```diff
@@ -637,10 +622,6 @@ sidInstrumentView::sidInstrumentView( Instrument * _instrument,
 		sync_btn->setInactiveGraphic(
 			PLUGIN_NAME::getIconPixmap( "sync" ) );
 		ToolTip::add( sync_btn, tr( "Sync" ) );
-		sync_btn->setWhatsThis( tr ( "Sync synchronizes the fundamental "
-			"frequency of Oscillator %1 with the fundamental frequency of "
-			"Oscillator %2 producing \"Hard Sync\" effects." ).arg( i+1 )
-			.arg( syncRingWidth[i] ) );
```
tresf: Now that we aren't translating `syncRingWidth`, it's throwing an used variable exception.  It should be removed with this PR.

zonkmachine:
Closes https://github.com/LMMS/lmms/issues/896 - More texts for "What is this"

plugins/sid/sid_instrument.cpp
Sawuare: Dona via e385c11. Sorry for the late reply.

plugins/zynaddsubfx/ZynAddSubFx.cpp
```diff
@@ -507,19 +507,19 @@ ZynAddSubFxView::ZynAddSubFxView( Instrument * _instrument, QWidget * _parent )
 	m_portamento->setLabel( tr( "PORT" ) );
 
 	m_filterFreq = new Knob( knobBright_26, this );
-	m_filterFreq->setHintText( tr( "Filter Frequency:" ), "" );
+	m_filterFreq->setHintText( tr( "Filter frequency:" ), "" );
 	m_filterFreq->setLabel( tr( "FREQ" ) );
 
 	m_filterQ = new Knob( knobBright_26, this );
-	m_filterQ->setHintText( tr( "Filter Resonance:" ), "" );
+	m_filterQ->setHintText( tr( "Filter resonance:" ), "" );
```
Wallacoloo: What's the rationale for writing "Q/Resonance" in some places (e.g. InstrumentSoundShaping) and plain old "resonance" in others?

src/gui/LfoControllerDialog.cpp
```diff
@@ -66,48 +66,25 @@ LfoControllerDialog::LfoControllerDialog( Controller * _model, QWidget * _parent
 	setWindowIcon( embed::getIconPixmap( "controller" ) );
 	setFixedSize( 240, 58 );
 	
-	ToolTip::add( this, tr( "LFO Controller" ) );
-
 	m_baseKnob = new Knob( knobBright_26, this );
 	m_baseKnob->setLabel( tr( "BASE" ) );
 	m_baseKnob->move( CD_LFO_BASE_CD_KNOB_X, CD_LFO_CD_KNOB_Y );
-	m_baseKnob->setHintText( tr( "Base amount:" ), "" );
-	m_baseKnob->setWhatsThis( tr("todo") );
```
Wallacoloo:
```
m_baseKnob->setWhatsThis( tr("todo") );
```
:100:

src/gui/LfoControllerDialog.cpp
```diff
@@ -66,48 +66,25 @@ LfoControllerDialog::LfoControllerDialog( Controller * _model, QWidget * _parent
...
 	m_speedKnob = new TempoSyncKnob( knobBright_26, this );
 	m_speedKnob->setLabel( tr( "SPD" ) );
 	m_speedKnob->move( CD_LFO_SPEED_CD_KNOB_X, CD_LFO_CD_KNOB_Y );
-	m_speedKnob->setHintText( tr( "LFO-speed:" ), "" );
-	m_speedKnob->setWhatsThis(
-		tr( "Use this knob for setting speed of the LFO. The "
-			"bigger this value the faster the LFO oscillates and "
-			"the faster the effect." ) );
-
+	m_speedKnob->setHintText( tr( "Modulation speed:" ), "" );
```
Wallacoloo: I understand trying to preserve some uniformity between "Modulation speed" and "modulation amount", but I think this is better as "LFO speed" or "LFO frequency".

It's sort of a two-step thing: you have an LFO, and you adjust _its_ frequency (i.e. the "LFO frequency" - not "Modulation frequency") and then you modulate some other signal with it ("Modulation amount").

src/gui/LfoControllerDialog.cpp
```diff
@@ -154,7 +131,7 @@ LfoControllerDialog::LfoControllerDialog( Controller * _model, QWidget * _parent
 	moog_saw_wave_btn->setInactiveGraphic(
 		embed::getIconPixmap( "moog_saw_wave_inactive" ) );
 	ToolTip::add( moog_saw_wave_btn,
-			tr( "Click here for a moog saw-wave." ) );
+			tr( "Moog saw wave" ) );
```
Wallacoloo: I think I saw "Moog-like saw wave" somewhere else. We should probably choose one or the other & be consistent throughout.

src/gui/MainWindow.cpp
```diff
@@ -485,81 +475,50 @@ void MainWindow::finalize()
...
-	m_toolBarLayout->addWidget( whatsthis, 0, 7 );
-	m_toolBarLayout->addWidget( m_metronomeToggle, 0, 8 );
+	m_toolBarLayout->addWidget( m_metronomeToggle, 0, 7 );
 
 
 	// window-toolbar
 	ToolButton * song_editor_window = new ToolButton(
 					embed::getIconPixmap( "songeditor" ),
-					tr( "Show/hide Song-Editor" ) + " (F5)",
+					tr( "Song-Editor" ) + " (F5)",
```
Wallacoloo: dehyphenate "Song editor" and "Piano roll" to match "Automation editor" and "FX mixer", while you're at it?

src/gui/SetupDialog.cpp
```diff
@@ -696,17 +688,11 @@ SetupDialog::SetupDialog( ConfigTabs _tab_to_open ) :
 
 	QPushButton * autoSaveResetBtn = new QPushButton(
 			embed::getIconPixmap( "reload" ), "", auto_save_tw );
-	autoSaveResetBtn->setGeometry( 290, 70, 28, 28 );
+	autoSaveResetBtn->setGeometry( 320, 70, 28, 28 );
 	connect( autoSaveResetBtn, SIGNAL( clicked() ), this,
 						SLOT( resetAutoSave() ) );
 	ToolTip::add( autoSaveResetBtn, tr( "Reset to default-value" ) );
```
Wallacoloo: dehyphenate "default value"?

src/gui/widgets/EnvelopeAndLfoView.cpp
```diff
@@ -590,8 +518,7 @@ void EnvelopeAndLfoView::lfoUserWaveChanged()
 		if( m_params->m_userWave.frames() <= 1 )
 		{
 			TextFloat::displayMessage( tr( "Hint" ),
-				tr( "Drag a sample from somewhere and drop "
-					"it in this window." ),
+				tr( "Drag and drop a sample in this window." ),
```
Wallacoloo: How about "Drag and drop a sample **into** this window"? Because, correct me if I'm wrong, but I don't think you can drag a sample _out_ of it.

plugins/OpulenZ/OpulenZ.cpp
```diff
@@ -94,45 +94,45 @@ const unsigned int adlib_opadd[OPL2_VOICES] = {0x00, 0x01, 0x02, 0x08, 0x09, 0x0
...
-	op2_vib_mdl(true, this, tr( "Op 2 Vibrato" )   ),
+	op1_waveform_mdl(0,0,3,this, tr( "Op 1 waveform" ) ),
+
+
+	op2_a_mdl(1.0, 0.0, 15.0, 1.0, this, tr( "Op 2 attack" )   ),
+	op2_d_mdl(3.0, 0.0, 15.0, 1.0, this, tr( "Op 2 decay" )   ),
+	op2_s_mdl(14.0, 0.0, 15.0, 1.0, this, tr( "Op 2 austain" ) ),
+	op2_r_mdl(12.0, 0.0, 15.0, 1.0, this, tr( "Op 2 release" )   ),
+	op2_lvl_mdl(63.0, 0.0, 63.0, 1.0, this, tr( "Op 2 level" )   ),
+	op2_scale_mdl(0.0, 0.0, 3.0, 1.0, this, tr( "Op 2 level Scaling" ) ),
```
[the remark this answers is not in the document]
Sawuare: Yes.

plugins/monstro/Monstro.cpp
```diff
@@ -828,120 +828,120 @@ inline sample_t MonstroSynth::calcSlope( int slope, sample_t s )
...
-		m_env2Rel( 0.0f, 0.0f, 4000.0f, 1.0f, 4000.0f, this, tr( "Env 2 Release" ) ),
-		m_env2Slope( 0.0f, -1.0f, 1.0f, 0.001f, this, tr( "Env 2 Slope" ) ),
-
-		m_o23Mod( 0, 0, NUM_MODS - 1, this, tr( "Osc2-3 modulation" ) ),
+		m_osc1Vol( 33.0, 0.0, 200.0, 0.1, this, tr( "Osc 1 volume" ) ),
+		m_osc1Pan( 0.0, -100.0, 100.0, 0.1, this, tr( "Osc 1 panning" ) ),
+		m_osc1Crs( 0.0, -24.0, 24.0, 1.0, this, tr( "Osc 1 coarse detune" ) ),
+		m_osc1Ftl( 0.0, -100.0, 100.0, 1.0, this, tr( "Osc 1 fine detune left" ) ),
+		m_osc1Ftr( 0.0, -100.0, 100.0, 1.0, this, tr( "Osc 1 fine detune right" ) ),
+		m_osc1Spo( 0.0, -180.0, 180.0, 0.1, this, tr( "Osc 1 stereo phase-offset" ) ),
```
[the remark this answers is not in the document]
Sawuare: I agree.

plugins/peak_controller_effect/peak_controller_effect_control_dialog.cpp
```diff
@@ -76,10 +76,10 @@ PeakControllerEffectControlDialog::PeakControllerEffectControlDialog(
 	m_tresholdKnob->setModel( &_controls->m_tresholdModel );
 	m_tresholdKnob->setHintText( tr( "Treshold:" ) , "" );
 
-	m_muteLed = new LedCheckBox( "Mute Effect", this );
+	m_muteLed = new LedCheckBox( tr( "Mute output audio" ), this );
```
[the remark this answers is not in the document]
Sawuare: Yes, it is `m_muteModel`. I'll change it accordingly.

plugins/sf2_player/sf2_player.cpp
```diff
@@ -99,15 +99,15 @@ sf2Instrument::sf2Instrument( InstrumentTrack * _instrument_track ) :
...
+	m_reverbDamping( FLUID_REVERB_DEFAULT_DAMP, 0, 1.0, 0.01, this, tr( "Damping" ) ),
+	m_reverbWidth( FLUID_REVERB_DEFAULT_WIDTH, 0, 1.0, 0.01f, this, tr( "Width" ) ),
+	m_reverbLevel( FLUID_REVERB_DEFAULT_LEVEL, 0, 1.0, 0.01f, this, tr( "Level" ) ),
 	m_chorusOn( false, this, tr( "Chorus" ) ),
-	m_chorusNum( FLUID_CHORUS_DEFAULT_N, 0, 10.0, 1.0, this, tr( "Chorus Lines" ) ),
-	m_chorusLevel( FLUID_CHORUS_DEFAULT_LEVEL, 0, 10.0, 0.01, this, tr( "Chorus Level" ) ),
-	m_chorusSpeed( FLUID_CHORUS_DEFAULT_SPEED, 0.29, 5.0, 0.01, this, tr( "Chorus Speed" ) ),
-	m_chorusDepth( FLUID_CHORUS_DEFAULT_DEPTH, 0, 46.0, 0.05, this, tr( "Chorus Depth" ) )
+	m_chorusNum( FLUID_CHORUS_DEFAULT_N, 0, 10.0, 1.0, this, tr( "Lines" ) ),
+	m_chorusLevel( FLUID_CHORUS_DEFAULT_LEVEL, 0, 10.0, 0.01, this, tr( "Level" ) ),
```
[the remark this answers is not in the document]
Sawuare: These texts show when the knobs are automated, so you're right, "Reverb" and "Chorus" should be kept for distinction.

src/gui/MainWindow.cpp
Sawuare: Sure. 👍

src/gui/widgets/EnvelopeAndLfoView.cpp
Sawuare: You're correct. I'll change it.

plugins/sf2_player/sf2_player.cpp
```diff
@@ -954,25 +948,21 @@ sf2InstrumentView::sf2InstrumentView( Instrument * _instrument, QWidget * _paren
...
 	m_chorusButton->setInactiveGraphic( PLUGIN_NAME::getIconPixmap( "chorus_off" ) );
 	ToolTip::add( m_chorusButton, tr( "Apply chorus (if supported)" ) );
-	m_chorusButton->setWhatsThis(
-		tr( "This button enables the chorus effect. "
-			"This is useful for cool echo effects, but only works on "
-			"files that support it." ) );
 
 	m_chorusNumKnob = new sf2Knob( this );
-	m_chorusNumKnob->setHintText( tr("Chorus Lines:"), "" );
+	m_chorusNumKnob->setHintText( tr("Lines:"), "" );
```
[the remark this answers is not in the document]
Sawuare: Yes. FWIW, the effects *Calf Multi Chorus* and *Multivoice Chorus* use the term "voices".

plugins/vestige/vestige.cpp
```diff
@@ -432,9 +427,6 @@ VestigeInstrumentView::VestigeInstrumentView( Instrument * _instrument,
 						SLOT( managePlugin() ) );
 	ToolTip::add( m_managePluginButton, tr( "Control VST-plugin from LMMS host" ) );
```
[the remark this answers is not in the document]
Sawuare: I actually missed a bunch of them. 😅

src/gui/LfoControllerDialog.cpp
Sawuare: I think "moog-like" is better than just "moog".

Sawuare: I noticed that there's a "moog" filter though, so I'll keep it like that.

plugins/zynaddsubfx/ZynAddSubFx.cpp
Sawuare: I forgot to change these, however, I'm not sure which term is more appropriate, so I'll revert these changes for now.

plugins/vestige/vestige.cpp
```diff
@@ -263,7 +263,7 @@ void vestigeInstrument::loadFile( const QString & _file )
 	{
 		tf = TextFloat::displayMessage(
 				tr( "Loading plugin" ),
-				tr( "Please wait while loading VST-plugin..." ),
+				tr( "Please wait while loading VST plugin..." ),
```
SecondFlight: Maybe add a "the" before "VST"? In this case it's talking about a specific VST plugin that it's loading and not just any plugin.

Sawuare:
Thanks @Wallacoloo and @SecondFlight for reviewing this. Sorry for squashing all this work into this PR.

Wallacoloo:
@Sawuare Thanks so much for coding these changes! I wish I could have polished the wording on that last comment about the size of the PR, but I was in a slight hurry when I wrote it. Anyway, it's not much of a bother - while smaller PRs can be more manageable, large PRs are infinitely better than no PRs :-)

Issue #3081, opened 2016-10-20
Allow undoing of MIDI entered notes


simonvanderveldt:
Well, the title captures it all :)

When working on a song in LMMS one can undo changes to the notes that were created using the pencil/copy+paste/etc tools, but one cannot undo notes added using MIDI. Would be very nice to add those to the undo stack as well.

Umcaruje:
:+1:

zonkmachine:
I did a quick test and added `addJournalCheckPoint();` [here](https://github.com/LMMS/lmms/blob/27007c258a7b83b8fb89bea753c9b4d767686ef3/src/tracks/Pattern.cpp#L214) which would let me undo one note at a time after recording. Should rather the whole last phrase be undone at the same time?

simonvanderveldt:
> I did a quick test and added addJournalCheckPoint(); here which would let me undo one note at a time after recording. Should rather the whole last phrase be undone at the same time?

I think different people will have different expectations, not sure what to choose. What to the commercial DAWs do?

Also, since it's better to have at least some way to undo vs none, I'd say this is already an improvement worth adding :)

Spekular:
I think it makes a bit more sense to delete all the recorded notes.

Reasoning:

* Box selection might select undesired notes or require scrolling/zooming,
deleting everything with ctrl+z is far more convenient.
* Deleting X individual notes with ctrl+z should be about the same amount
of effort as deleting X individual notes with RMB

musikBear:
> I think it makes a bit more sense to delete all the recorded notes.

Isent *that* just as simple as ctrl+a + del ?
all gone
Stepwise would give two different delete options.. idk...
-But if we are talking *phrase-wise* redo, then its another story, and i guess @Spekular reasoning is sound.
A *phrase* would then be: all notes captured between rec-pressed and stop-pressed -right?

zonkmachine:
> Isent that just as simple as ctrl+a + del ?
> all gone

No. You can record over already recorded notes. If you add a checkpoint to when you start recording you can record something, stop and then record again, stop again and you wouild have two separate recordings in the same pattern. It would take two `<Ctrl>+z` to undo this if you add a checkpoint to the beginning of the recording function.

Umcaruje:
>Should rather the whole last phrase be undone at the same time?

This would be a lot better for UX, if you want to undo specific notes you might as well delete them.

>Maybe check what the commercial DAWs do?

[...]

PR #4144, opened 2018-01-31
Add journaling checkpoint to recording notes

zonkmachine:
Fixes https://github.com/LMMS/lmms/issues/3081


zonkmachine:
This fixes the glitchy behaviour of the undo function with recorded notes in the Piano Roll. You can add notes manually, record phrases with a keyboard or mix those methods. When you now use the undo function the manually added notes will be removed one by one and the recorded ones will be removed together. I'll merge this in a day.

Issue #2380, opened 2015-09-24
Chord delay


DeRobyJ:
In Func menu, Chord section.
Let's add a new knob for "delay", that is a delay between notes.

When I try to play some octaves or fifths manually and I like them, I often activate the Chord function, but they sound different, because all the note start at the same time.
Instead, having a 1ms delay before each note of the chord could be really fun!

It's something like the Tempo knob for Arpeggios, but instead of releasing notes and laying them over and over, it keeps each note until the chord is complete.
Of course it would be nice to select a tempo sync just like in the Tempo knob.

(this knob can be then randomized with an LFO controller, giving a more "human" or natural effect to chords)

If I didn't explain it well enough:
Select Fifths and a range of 2
You play C3:
current behaviour: it plays C3, G3, C4 and G4 simoultaneously.

With the Delay knob, put at 10ms: it plays C3, after 10 ms G3, after another 10 ms C4, and 10ms after G4.

tresf:
Sounds like a "harp" feature, simulate the strumming of strings with a short gap between.

DeRobyJ:
Exactly!
In general every plucked string instrument could make use of that to sound more natural.

musikBear:
Arent we really talking 'general humanizing' -eg ways to make both Quantization and or volume with just a few percent jitter or randomness?
(Then this is a duplicate)

DeRobyJ:
Duplicate of what issue?

tresf:
> Duplicate of what issue?

He's referring to #1165

> general humanizing

Although they both have their roots in adjusting timing (or in the case of #1165, pitch), this isn't the same exact request.  @DeRobyJ is requesting that a chord has a delay between attacks of the same chord.

For example a chord like this:

![image](https://cloud.githubusercontent.com/assets/6345473/10100335/05f72ba6-635f-11e5-8135-3e1564b7ed13.png)

may actually be played like this

![image](https://cloud.githubusercontent.com/assets/6345473/10100344/18fed53c-635f-11e5-8586-d138b17e01f5.png)

But he's requesting it on the `FUNC`, `STACKING` section, e.g.

![image](https://cloud.githubusercontent.com/assets/6345473/10100406/8e460572-635f-11e5-81d4-956fb429658a.png)

DeRobyJ:
Ok, that's right.

(even though I think just one knob could do it)

leeavital:
Would this be for all chords, or just "stacked chords" -- i.e. created through the stacking function.

[...]

PR #4146, opened 2018-01-31
Chord spread

zonkmachine:
chordspread

Implementing delay for the chord stacking. Works pretty well but I'm a bit uncertain what the wanted combined result with arpeggio should be.

Closes https://github.com/LMMS/lmms/issues/2380

_Milestone 1.3.0 - pulled against stable-1.2 because I can't compile master right now_

### DIFF
--- a/include/InstrumentFunctionViews.h
+++ b/include/InstrumentFunctionViews.h
@@ -56,6 +56,7 @@ private:
 	GroupBox * m_chordsGroupBox;
 	ComboBox * m_chordsComboBox;
 	Knob * m_chordRangeKnob;
+	TempoSyncKnob * m_chordTimeKnob;
 
 } ;
 
--- a/include/InstrumentFunctions.h
+++ b/include/InstrumentFunctions.h
@@ -147,7 +147,7 @@ private:
 	BoolModel m_chordsEnabledModel;
 	ComboBoxModel m_chordsModel;
 	FloatModel m_chordRangeModel;
-
+	TempoSyncKnobModel m_chordTimeModel;
 
 	friend class InstrumentFunctionNoteStackingView;
 
--- a/src/core/InstrumentFunctions.cpp
+++ b/src/core/InstrumentFunctions.cpp
@@ -207,7 +207,8 @@ InstrumentFunctionNoteStacking::InstrumentFunctionNoteStacking( Model * _parent 
 	Model( _parent, tr( "Chords" ) ),
 	m_chordsEnabledModel( false, this ),
 	m_chordsModel( this, tr( "Chord type" ) ),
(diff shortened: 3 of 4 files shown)


Mark-Agent003:
Will there be a way to choose which direction the chords spread?

zonkmachine:
That's a bit of a real estate problem. There isn't much room for more controllers in there.

DeRobyJ:
Well the initial thought was made just to add some "humanization" to chords.
So this change applies the idea to have a delay for each note, which can be used as an arpeggio.

If we were to do something more about this, given the space, I'd suggest to have:
- Upward strum (base note -> delay -> second note -> delay -> etc)
- Downward strum (last note -> delay -> secondlast -> delay -> etc)
- Random strum (one random note each time)

As for where to put it, I think we can shrink the drop-down more and include a radio button... 
For example this is how it looks on my localization:
immagine
Look at the Arpeggio chord drop-down, it's already really short, while the Stacking one is quite long.

zonkmachine:
> Well the initial thought was made just to add some "humanization" to chords.

Yeah, this PR isn't precisely 'humanizing' anything. You wan't some randomness in there then.

DeRobyJ:
Yeah but for this PR in particular I'd actually prefer it staying like this, maybe adding the direction options.
On my original issue, I mentioned using an LFO to change the speed of the strum.

As for humanization, I now think it should be done in a different way (and I don't think many DAWs have this kind of feature).
The general idea is to have a general option for the track (in the Misc tab) to add "inaccuracy", that delays any note event of a small random time.

The issue I'm actually trying to address is Phase Relationship between notes or even instruments: when it's fixed, it gets boring, so it needs some randomization to make it more human-like.

But I should write an issue about this, as the feature request is out of the scope of this PR.

lukas-w:
> I'd suggest to have:
> 
>   *  Upward strum (base note -> delay -> second note -> delay -> etc)
>   *  Downward strum (last note -> delay -> secondlast -> delay -> etc)
>   *  Random strum (one random note each time)

This sounds very similar to what the "Direction" option of Arpeggio offers. Can we unify this? The way I understand it, Chord Spread as suggested in #2380 is just Arpeggio with no repetition + increased note length.

DeRobyJ:
As far as I know, the arpeggio funcion has to deal with note looping and timing in a way that might be a lot different to what is needed for this feature. But I may be mistaken about this.
Moreover, putting a toggle on the Arpeggio tab would not allow to have an arpeggio of strums, not that it would be used that much.

zonkmachine:
> This sounds very similar to what the "Direction" option of Arpeggio offers. Can we unify this?

I tried to do this in the arpeggiator first but this is what worked fastest for this proof of concept.


I won't have time to implement this properly so if anyone feels up for the task, please grab this code.

Issue #3973, opened 2017-11-15
Undoing mixer volume change only undoes partially


Mark-Agent003:
How to reproduce:
Open LMMS 1.2.0-rc4
Change one of the mixer's volume knob
Press Ctrl-z a few times

The effect: The volume moves back a fraction of the way, instead of going straight back to 100%

PhysSong:
Confirmed. This seems to affect other faders, too. Note that knobs are not affected.
Technical details: `Knob` disables its model's journaling in `mousePressEvent()` and restores it in `mouseReleaseEvent()`. So, tweaking a knob creates only one journaling point for single action. However, `Fader` doesn't have such logic. In other words, So, adjusting a fader creates journaling points whenever the model's value is changed.

> Change one of the mixer's volume knob

Assuming you mean the fader here...

musikBear:
confirming for win 32 -but Actually if you tab ctrl+z n times, the slider will move to the previous pos, but not 'stay' there, instead it will start going in direction of the next-previous pos, aso

zonkmachine:
> Technical details: Knob disables its model's journaling in mousePressEvent() and restores it in mouseReleaseEvent(). So, tweaking a knob creates only one journaling point for single action. However, Fader doesn't have such logic. In other words, So, adjusting a fader creates journaling points whenever the model's value is changed.

I just adapted the Knob code to the Fader and it worked. I failed at making the Song Editor master sliders, Master volume and Master pitch, to work better with journalling with the same method. https://github.com/LMMS/lmms/pull/4147

zonkmachine:
> Note that knobs are not affected.

Well, scrolling with the mouse wheel is affected the same way.

PR #4147, opened 2018-02-01
Turn journalling off when moving fader

zonkmachine:
Partial fix for https://github.com/LMMS/lmms/issues/3973
This fixes the fader journalling in the FX-Mixer/Equalizer. The sliders in the Song Editor, Master volume and Master pitch, needs to be fixed too. I failed at this...


zonkmachine:
Since the same issue is true with undoing mouse wheel action, maybe we should use [QHoverEvent](http://doc.qt.io/archives/qt-4.8/qhoverevent.html) to detect mouse over/off the fader and use this to set on/off journalling instead of the mouse press event?

zonkmachine:
Review please!

Issue #3936, opened 2017-11-03
ZynAddSubFX Portament not working


superjosh364:
The portamento on the ZASFX front page doesn't work for the range 0-64, but if you turn it above 64 it works fine.

zonkmachine:
> but if you turn it above 64 it works fine.

It is portamento there from 64 and up but it's the same level. It sounds like we're sending an on/off message instead of a level value.

zonkmachine:
Right. What we have implemented in lmms is midi control message 65, portamento on/off. There are three other control messages for portamento. 5, 37 and 84. In the midi standard boolean values are realized with a value of 0-127, the lower half of a byte, where 0-63 is off and 64-127 is on.
Midi standard control messages:
https://www.midi.org/specifications/item/table-3-control-change-messages-data-bytes-2
ZASFX, Enable or disable portamento (scroll down for the rest of the portamento code):
https://github.com/LMMS/lmms/blob/stable-1.2/plugins/zynaddsubfx/zynaddsubfx/src/Params/Controller.h#L55:L58

musikBear:
It has been so ever. There has never been a gradual porta control, so this ticket is a feature request, not bug

zonkmachine:
You're probably right there.

noahb01:
Maybe we could replace the knob with a toggle? I assume there is some technical reason for implementing the portamento control as an on/off instead of a gradual control.

superjosh364:
Well, if you have a portament knob, it usually determines the amount of time the pitch glides from one note to another. So there should be a toggle switch to turn it on, and then a knob for adjusting the length of the portamento.

Spekular:
From what I've seen, there tends to be a single knob w/ no toggle. Setting
the portamento time to 0 naturally turns it off.

noahb01:
@Spekular @superjosh364 What I mean is that if there is no way to access the actual portamento value, we should replace it with an on/off toggle. If we can control that value, by all means do it.

musikBear:
its just because the 'wrong' controller (red circle) were selected for the ui
![zynporta](https://user-images.githubusercontent.com/6662066/36063809-060fbf42-0e82-11e8-9f27-3878b9bea75a.JPG)
It should be *time.t* (green circle).
But as you can see, there are more advanced porta-settings available.
Perhaps the dial could be preserved, but default 'grayed-out' meaning *no-porta*. if enabled with left-click, *time.t* could be changed.
That way the current ui would not have to change drastically.

[...]

PR #4167, opened 2018-02-10
Change portamento controller from toggle to set value

zonkmachine:
In response to https://github.com/LMMS/lmms/issues/3936

This PR turns the portamento off at a portamento setting of 0 and forwards the same setting to the zynaddsubfx controller value `portamento.time`.

 - ZynAddSubFx seem to have separate portamento settings for each partial sound. This I'm not addressing in this PR.
 - Currently no backward compatibility. ~~Easy to fix though as long as the portamento isn't automated. Just set portamento over 64 to 64 and under 64 to 0.~~

_Caveat. I know preciously little about ZynAddSubFX..._

### DIFF
--- a/plugins/zynaddsubfx/zynaddsubfx/src/Params/Controller.cpp
+++ b/plugins/zynaddsubfx/zynaddsubfx/src/Params/Controller.cpp
@@ -56,7 +56,7 @@ void Controller::defaults()
     portamento.propRate     = 80;
     portamento.propDepth    = 90;
     portamento.receive      = 1;
-    portamento.time = 64;
+    portamento.time = 0;
     portamento.updowntimestretch = 64;
     portamento.pitchthresh     = 3;
     portamento.pitchthreshtype = 1;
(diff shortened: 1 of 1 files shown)


Umcaruje:
>Just set portamento over 64 to 64 and under 64 to 0.

Actually that wouldn't work because the zyn preset might have a different portamento time set, it would probably need setting the knob value to the zyn value if over 64. No idea how you would code that though.

zonkmachine:
Yeah, I'm totally not on top of this one. The portamento is weird. The PR works in as much as it turns portomento on/off gradually but I cant assess what it's doing to the presets. @unfa 

The only MIDI controller related to portamento that seem to be implemented in zynaddsubfx seem to be on/off, controller 65. I would have expected them to be represented [here](https://github.com/LMMS/lmms/blob/stable-1.2/plugins/zynaddsubfx/zynaddsubfx/src/globals.h#L171:L181) otherwise.

zonkmachine:
Closing this for now as I won't have time to work on it.

Issue #4155, opened 2018-02-05
Instrument-track measure ticks sometimes off by 1px


noahb01:
In the song-editor, the ticks on each measure for the automation track are sometimes shifted forward by 1px.

**Reproduce:**
Open new project.
Make an instrument-track and an automation-track.
Put some random stuff into the tracks, about 8-16 measures is good.
Look at the ticks on the tracks, some will not line up!

zonkmachine:
It looks like it's the Instrument track that is a bit off. The Automation and BBTracks align with the time line.
![ticksalign](https://user-images.githubusercontent.com/6368949/35791101-124a1d60-0a46-11e8-803e-a0992ab4f43a.png)

noahb01:
![screenshot from 2018-02-05 09-50-16](https://user-images.githubusercontent.com/22487040/35810515-01535350-0a5a-11e8-9a0c-ea634b84d5d2.png)

Interestingly on my system it appears as if the ticks "fall behind" on the Instrument tracks. The first few ticks line up but get worse as the track goes on.

Haven't looked at the code but I assume the ticks are drawn like this: Draw tick, move forward by X pixels, draw tick, ...

If this is the case then it appears that the value "X" being added is off by a very small amount.

noahb01:
BTW, I'm using LMMS 1.2.0-rc5, Qt 5.9.2, GCC 4.8.4 AppImage build.

zonkmachine:
It happened in commit 8c2ebf287d9de57466ffa8b1f4bda184c171dde8

```
8c2ebf287d9de57466ffa8b1f4bda184c171dde8 is the first bad commit
commit 8c2ebf287d9de57466ffa8b1f4bda184c171dde8
Author: Dave French <dave.french3@googlemail.com>
Date:   Fri Dec 26 12:33:00 2014 +0000

    Proposed fix for 1255 Segment/clip not showing whole last note in Song Editor
```

Changing https://github.com/LMMS/lmms/blob/stable-1.2/src/tracks/Pattern.cpp#L887 to
`( width() - TCO_BORDER_WIDTH )` seem to help a bit. The ticks still look a bit off but it's the same for all tracks. @noahb01 If you want to poke this one a bit that's fine with me. I just get the feeling this could be a frustrating job of pushing pixels around. You've been warned.

PhysSong:
@zonkmachine `( width() - 1 )` looks better. See this:
[...]

PR #4171, opened 2018-02-11
Fix ui glitch where track ticks did not line up

noahb01:
This fixes the a bug (see #4155) where the instrument-track ticks did not line up with the timeline. This only applies to `stable-1.2`. A different fix would be required to fix `master`, the code is completely different.

### DIFF
--- a/src/tracks/Pattern.cpp
+++ b/src/tracks/Pattern.cpp
@@ -884,7 +884,7 @@ void PatternView::paintEvent( QPaintEvent * )
 	const float ppt = fixedTCOs() ?
 			( parentWidget()->width() - 2 * TCO_BORDER_WIDTH )
 					/ (float) m_pat->length().getTact() :
-				( width() - 2 * TCO_BORDER_WIDTH )
+				( width() - TCO_BORDER_WIDTH )
 					/ (float) m_pat->length().getTact();
 
 	const int x_base = TCO_BORDER_WIDTH;


noahb01:
The build failure is a `cmake` issue and is not related to this PR. This code compiles correctly on my machine.

BaraMGB:
Can you explain what's the difference between fixing stable and master?

noahb01:
@BaraMGB they're different branches. stable-1.2 is a stable release and is only accepting bug fixes and not new features (to prevent code from breaking), master is the bleeding edge development version and is the most up-to-date (but sometimes has new bugs). Because stable and master had different code for the bug I was fixing, my for only worked in stable and not master.

zonkmachine:
> Because stable and master had different code for the bug I was fixing,

I think this was the question. What has changed between stable-1.2 and master?
[Here](https://gist.github.com/zonkmachine/431e4e47bcf41d581e275d784a799ceb) is the output of `git diff stable-1.2 master -- src/tracks/Pattern.cpp`

The changes in this PR touches code [here](https://gist.github.com/zonkmachine/431e4e47bcf41d581e275d784a799ceb#file-pattern-diff-L93:L97) that needs some 'lovin' in order to be mergeable into master.

zonkmachine:
Merge? _(We merge this now and keep https://github.com/LMMS/lmms/issues/4155 open for a final fix in master.)_

zonkmachine:
I believe this line in master is the equivalent of the line that's changed in this PR.
https://github.com/LMMS/lmms/blob/56b4740146cddf02c76ed0d72063e8eaabdc397f/src/tracks/Pattern.cpp#L904

Merging.

PR #4083, opened 2017-12-30
Don't restore audio device during exporting

PhysSong:
This PR fixes deadlock on multi-track export with SDL.
Fixes #4032.

PR #4189, opened 2018-02-25
Fix crash when cancelling rendering of a project with VSTs

DomClark:
Previously the `RenderManager` was deleted before the `finished()` signal was processed, but since https://github.com/LMMS/lmms/commit/d0b3be7f00e732dd1c5dbb43cb7030d42ff6a346, when VST plugins are present an event loop is run waiting for a reply from a sample rate update message, which processes the `finished()` signal and calls the `renderNextTrack()` slot, resulting in a second call to `~ProjectRenderer()`. Disconnecting the slot manually fixes the problem.

### DIFF
--- a/src/core/RenderManager.cpp
+++ b/src/core/RenderManager.cpp
@@ -52,6 +52,8 @@ RenderManager::~RenderManager()
 void RenderManager::abortProcessing()
 {
 	if ( m_activeRenderer ) {
+		disconnect( m_activeRenderer, SIGNAL( finished() ),
+				this, SLOT( renderNextTrack() ) );
 		m_activeRenderer->abortProcessing();
 	}
 	restoreMutedState();


zonkmachine:
Tested, works!

See comment [here](https://github.com/LMMS/lmms/pull/4083#issuecomment-368321330).

DomClark:
I'd be surprised if this were the cause for the hang there. The only method modified here is `RenderManager::abortProcessing`, which is called when cancelling rendering, but the backtrace there contains
```
#12 0x000000000064ba9b in RenderManager::finished (this=0x189c050) at /home/zonkmachine/builds/lmms/lmms/build/src/moc_RenderManager.cpp:113
#13 0x00000000005521a1 in RenderManager::renderNextTrack (this=0x189c050) at /home/zonkmachine/builds/lmms/lmms/src/core/RenderManager.cpp:77
```

indicating that rendering completed normally.

DomClark:
Merge?

Issue #3922, opened 2017-10-28
Instrument presets saves MIDI settings


zonkmachine:
Instruments are saved with MIDI settings as is the case with some of our factory presets. When previewing these sounds the project is set to modified. Saving instrument presets should not save any MIDI information.
Examples: Monstro - Growl, HorrorLead

 - [x] Fix factory presets. (#4207)
 - [x] Check factory projects. (#4207)
 - [ ] Don't save MIDI settings.

Related: [Presets remember the controller connections](https://github.com/LMMS/lmms/issues/2625)

qnebra:
Maybe this section is preset? Taken from Monstro - Growl preset:

<midiport inputcontroller="0" inports="" fixedoutputvelocity="-1" inputchannel="0" outputcontroller="0" writable="0" outputchannel="1" fixedinputvelocity="-1" fixedoutputnote="-1" outputprogram="1" basevelocity="63" readable="1"/>

PhysSong:
Should we fix factory files in `stable-1.2`, or `master`?

zonkmachine:
I think `stable-1.2`

tresf:
Yes, stable.  1.3.0 is slated to use the assets repo instead.

PhysSong:
In the checklist, first two things can be fixed easily. I can create a pull request addressing them.

zonkmachine:
> In the checklist, first two things can be fixed easily. I can create a pull request addressing them.

Yes. Fix the first two for 1.2 and the saving issue some other time.

zonkmachine:
Possibly related to https://github.com/LMMS/lmms/issues/1644

Rossmaxx:
How hard is it to get this done? I might handle this. Need some pointers to look at tho.

Rossmaxx:
@PhysSong @zonkmachine some help needed.

PR #4207, opened 2018-03-02
Remove MIDI connection from factory files

PhysSong:
Remove MIDI connection from factory presets and factory projects.
Address #3922.


PhysSong:
I'm going to merge this soon if there are no objections.

PhysSong:
It seems like I missed some `.mmpz` files. I'll do this for those files as well.

PR #4211, opened 2018-03-03
Use Canonical Paths for Relative Paths Calculations

### LINKED ISSUE #4173: Trailing slashes can break relative paths
This may be a long standing bug, but there's something in LMMS that appears to put additional trailing slashes in working paths.

e.g. 
```cmd
C:/samples/
```

vs.
```cmd
c:/samples///
```

Although I'm not sure the exact cause of this, it's a benign problem when loading samples.  The issue seems to come into play when saving because the relative path logic is broken.

To reproduce, add an extra slash or two to the `LMMS WORKING DIRECTORY` in the preferences pane and restart LMMS.

screen shot 2018-02-13 at 1 07 00 pm

Next, import some samples into the project from your `LMMS WORKING DIRECTORY PATH` and save the project to an `.mmp` format.  When the files are saved, instead of being relatively, they contain the full paths.

**Expected:**
```xml
<audiofileprocessor ... src="mysample.ogg" />
```

**Actual:**
```xml
<audiofileprocessor ... src="/home/tres/lmms/samples/mysample.ogg" />
```

The underlying bug may actually be something with the preferences dialog, but we need to fix the trailing spaces bug as well incase this is typed by a user on accident.

Disclaimer, this bug was introduced using the Sample Track, not audiofileprocessor.  Example above is hypothetical.

tresf:
Forces strange paths like `/path/.///to/../from/foo/////` to be `/path/from/foo`.

Closes #4173

### DIFF
--- a/tests/src/core/RelativePathsTest.cpp
+++ b/tests/src/core/RelativePathsTest.cpp
@@ -40,8 +40,10 @@ private slots:
 
 		QString absPath = fi.absoluteFilePath();
 		QString relPath = "drums/kick01.ogg";
+		QString fuzPath = absPath.replace(relPath, "drums/.///kick01.ogg");
 		QCOMPARE(SampleBuffer::tryToMakeRelative(absPath), relPath);
 		QCOMPARE(SampleBuffer::tryToMakeAbsolute(relPath), absPath);
+		QCOMPARE(SampleBuffer::tryToMakeRelative(fuzPath), relPath);
 	}
 } RelativePathTests;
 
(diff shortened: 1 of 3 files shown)


tests/src/core/RelativePathsTest.cpp
```diff
@@ -40,8 +40,10 @@ private slots:
 
 		QString absPath = fi.absoluteFilePath();
 		QString relPath = "drums/kick01.ogg";
+		QString fuzPath = absPath.replace(relPath, "drums/.///kick01.ogg");
```
PhysSong: Current one won't work because this line modifies `absPath`. This one will work:
```cpp
QString fuzPath = absPath;
fuzPath.replace(relPath, "drums/.///kick01.ogg");
```

tresf: Ah yes, forgot Strings were mutable in Qt.  Thanks.

PR #4213, opened 2018-03-05
Enable FPE on Mac

tresf:
Adds a small header for MacOS to allow floating point exceptions to be caught.

This header falls back to the stock header.  It should not affect Linux platforms.

Related #3687


tresf:
Merging because this really has no negative effect on a default build.  Note, this FPE will completely desolate a Mac build due to FPE that are fired between Cocoa and the Qt framework that don't seem directly related to LMMS code.

This means in order to effectively use FPE on Mac, the `feenableexcept` block from `main.cpp` needs to be relocated later inside the source code.  Long term, we may need to turn this into a C++ macro so that it can be toggled on for certain areas of code, but that is out of scope.

PR #4227, opened 2018-03-08
Fix some bugs with LFO waveform drag&drop

### LINKED ISSUE #3307: LFO - Glitch on dropping a user wave/sample
There is some glitchy action going on when you drop a sample on the lfo wave graph.

 - If you drop a sample when the user wave is active, you need to first move one of the ENV/LFO knobs to update the graph.
 - If you drop a sample when the user wave is not selected, you need to first select another wave and then the user wave to update.

The new sample is however active/accessible from the instant you drop it on the LFO but it's not indicated.

PhysSong:
Update LFO waveform display after dropping a sample file. Fixes #3307.
This PR also fixes a broken functionality, "drag&drop from sample tracks".

### DIFF
--- a/src/gui/widgets/EnvelopeAndLfoView.cpp
+++ b/src/gui/widgets/EnvelopeAndLfoView.cpp
@@ -394,13 +394,17 @@ void EnvelopeAndLfoView::dropEvent( QDropEvent * _de )
 					StringPairDrag::decodeValue( _de ) );
 		m_userLfoBtn->model()->setValue( true );
 		_de->accept();
+		update();
 	}
 	else if( type == QString( "tco_%1" ).arg( Track::SampleTrack ) )
 	{
 		DataFile dataFile( value.toUtf8() );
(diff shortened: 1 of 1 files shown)


zonkmachine:
> This PR also fixes a broken functionality, "drag&drop from sample tracks".

Nice catch!

PhysSong:
Maybe out of scope, but is it better to activate user-wave mode automatically when a sample file is dropped?

zonkmachine:
> Maybe out of scope, but is it better to activate user-wave mode automatically when a sample file is dropped?

Yes. That would be better.

PhysSong:
> > Maybe out of scope, but is it better to activate user-wave mode automatically when a sample file is dropped?
> 
> Yes. That would be better.

Done via e4f0d254644e5684caaa89ceba1e94756eb2bb45.

zonkmachine:
> Done via e4f0d25.

Better.

PR #4233, opened 2018-03-09
Draw the entire length of a notes pitch bend info

### LINKED ISSUE #4231: In Piano Roll Editor, the last segment of the "Note detuning" curve is missing.
In instruments like Sitar and Santoor, all notes are typically detuned (pitch-bent).
Thus its notation needs to be displayed accurately.

But LMMS does not show the the last segment of the "Note detuning" curve in Piano Roll Editor.

Compare the two cases below:

image

image

**Desired:** Show the curve till the end of the note.

zonkmachine:
The detuning/pitch info is only drawn for the length of the note. The actual detuning is carried out for the duration of the automation and while the note is playing including the decay. This commit removes the note length test from drawDetuningInfo() which will allow for the full length of the detuning to be drawn.

Fixes #4231


zonkmachine:
I've tested this in the Song Editor and BBTrack. It works fine and I've run out of tests to perform so I'm leaving this open until tomorrow and then merge.

PR #4235, opened 2018-03-10
refactor: Use unique_ptr for memory management

Wallacoloo:
Hi!

A huge hurdle when I used to contribute regularly to LMMS was lifetime management. Pointers get passed around throughout the codebase and it's not clear who's supposed to delete them. That leads to double-frees, or memory leaks, and this made it exceedingly difficult to edit any of the core code. That, in turn, raises the barrier of entry to development & drives away good developers, etc. A good path forward is to use owning pointers ("smart pointers"), as they make ownership obvious across interfaces and lesson the likelihood of missing a `delete` or accidentally `delete`ing something that somebody else expected to remain alive.

So, I'll be combing the codebase and switching things over to owning pointers throughout the future. Starting off small with this PR so as to not surprise anyone. This PR changes `AudioPort::m_effects` to be `unique_ptr`, as it's always owned by `AudioPort`. It also changes `ComboBoxModel::addItem` to accept `unique_ptr<PixmapLoader>` instead of `PixmapLoader*`, to indicate that the ownership of the loader is transferred into the class.

I had to update ComboBoxModel to use `std::pair` instead of `QPair`, because the latter doesn't support move semantics, which is required when moving a `unique_ptr`.

Also added a `stdshims.h` header file for access to `make_unique`, since `std::make_unique` is only available in C++14 or greater & we use C++11.

### DIFF
--- a/include/stdshims.h
+++ b/include/stdshims.h
@@ -1,14 +1,17 @@
 //! Shims for std:: functions that aren't available in the current C++ versions
 //! we target.
 
-#pragma once
+#ifndef STDSHIMS_H
+#define STDSHIMS_H
 
 #include <memory>
 #include <utility>
 
 #if (__cplusplus >= 201402L)
 #warning "This file should now be removed! The functions it provides are part of the C++14 standard."
-#endif
+using std::unique_ptr;
+
+#else
 
 /// Shim for http://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique
 template<typename T, typename... Args>
@@ -16,4 +19,7 @@ std::unique_ptr<T> make_unique(Args&&... args)
 {
 	return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
 }
+#endif
+
+#endif // include guard
 
--- a/include/stdshims.h
+++ b/include/stdshims.h
@@ -3,6 +3,9 @@
 
 #pragma once
 
+#include <memory>
+#include <utility>
+
 /// Shim for http://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique
 template<typename T, typename... Args>
 std::unique_ptr<T> make_unique(Args&&... args)
--- a/include/stdshims.h
+++ b/include/stdshims.h
@@ -6,6 +6,10 @@
 #include <memory>
 #include <utility>
 
+#if (__cplusplus >= 201402L)
+#warning "This file should now be removed! The functions it provides are part of the C++14 standard."
+#endif
+
 /// Shim for http://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique
 template<typename T, typename... Args>
 std::unique_ptr<T> make_unique(Args&&... args)
--- a/include/stdshims.h
+++ b/include/stdshims.h
@@ -9,7 +9,7 @@
 
 #if (__cplusplus >= 201402L)
 #warning "This file should now be removed! The functions it provides are part of the C++14 standard."
-using std::unique_ptr;
+using std::make_unique;
 
 #else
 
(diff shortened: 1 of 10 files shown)


include/stdshims.h
```diff
@@ -0,0 +1,19 @@
+//! Shims for std:: functions that aren't available in the current C++ versions
+//! we target.
+
+#pragma once
```
PhysSong: Not critical, but some compilers may not understand this directive. How about using macros for this purpose(you may use macros and pragma directives simultaneously)?

include/stdshims.h
```diff
@@ -0,0 +1,19 @@
...
+
+#include <memory>
+#include <utility>
+
+#if (__cplusplus >= 201402L)
+#warning "This file should now be removed! The functions it provides are part of the C++14 standard."
+#endif
```
PhysSong: Why don't you write this like this?
```c++
#if (__cplusplus >= 201402L)
#warning "This file should now be removed! The functions it provides are part of the C++14 standard."
using std::make_unique;
#else
// Shim
#endif
```
or write the function in the `std` namespace?

src/gui/editors/PianoRoll.cpp
```diff
@@ -65,6 +67,7 @@
 #define MiddleButton MidButton
 #endif
 
+using namespace std;
```
PhysSong: Why did you add `using namespace std;` into this file, while other files don't have this?

Wallacoloo: I needed this to drag `std::move` into the global scope. The other files\* only use `make_unique`, which is dragged into scope by stdshims.h.

\* except for ComboBoxModel.cpp; I just pushed a commit that has it also `using namespace std;` instead of explicitly prefixing all the std types with `std::`.

Also, I avoid `using namespace` directives in any _headers_, as that's usually bad practice (e.g. https://stackoverflow.com/a/14575955/216292 ).

If preferred, I can explicitly specify `using std::move; using std::unique_ptr;` etc, instead of dragging in the whole namespace - just let me know.

include/stdshims.h
Wallacoloo: Never witnessed it in person, but okay - #ifdef-style include guards seem to be the norm for LLMS so I've updated to use those.

include/stdshims.h
Wallacoloo: Pushed an update using the code you listed above. Thanks!

include/stdshims.h
```diff
@@ -1,19 +1,25 @@
...
+#ifndef STDSHIMS_H
+#define STDSHIMS_H
 
 #include <memory>
 #include <utility>
 
 #if (__cplusplus >= 201402L)
 #warning "This file should now be removed! The functions it provides are part of the C++14 standard."
-#endif
+using std::unique_ptr;
```
PhysSong: As I mentioned before, this should be `using std::make_unique;`.

src/gui/editors/PianoRoll.cpp
PhysSong: I prefer `using std::move;`, `using std::unique_ptr;`, etc. However, this would be fine if it doesn't bring any side effects.

include/stdshims.h
Wallacoloo: Ah, my bad - thanks for catching that. Fixed.

src/gui/editors/PianoRoll.cpp
Wallacoloo: Okay, I switched to explicitly writing `using std::move;`, `using std::unique_ptr;`, etc.

tresf:
@Wallacoloo is there a reason you didn't squash this when you merged it?

Wallacoloo:
@tresf Sorry! No reason I didn't squash them - just that I've forgotten some details of the workflow for contributing to lmms. I'll squash groups of related commits like this next time.

PR #4239, opened 2018-03-12
Fix marked semitones in the piano roll

### LINKED ISSUE #3651: Mark current scale in Piano Roll not working - LMMS Beta 1.20-RC3
When selecting any scale and right-clicking on piano roll to mark the scale, the scale doesn't get marked in the Piano Roll editor view after clicking mark current scale. Windows version 64 bit.

Umcaruje:
Fixes #3651 
Went with option `#2` from the issue. @musikBear @RebeccaDeField 
Screenshots: 
image
image

The marked semitones now draw over the whole grid and are a transparent overlay. This will break any existing user themes for 1.2 but it's as easy to fix as adding transparency to your color.

### DIFF
--- a/data/themes/default/style.css
+++ b/data/themes/default/style.css
@@ -146,7 +146,7 @@ PianoRoll {
 	qproperty-noteBorders: false; /* boolean property, set false to have borderless notes */
 	qproperty-selectedNoteColor: #006b65;
 	qproperty-barColor: #078f3a;
-	qproperty-markedSemitoneColor: #06170E;
+	qproperty-markedSemitoneColor: rgba(255, 255, 255, 20);
 	/* Grid colors */
 	qproperty-lineColor: #292929;
 	qproperty-beatLineColor: #2d6b45;
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -2633,23 +2633,6 @@ void PianoRoll::paintEvent(QPaintEvent * pe )
 
 	int key = m_startKey;
 
-	// display note marks before drawing other lines
-	for( int i = 0; i < m_markedSemiTones.size(); i++ )
-	{
-		const int key_num = m_markedSemiTones.at( i );
-		const int y = keyAreaBottom() + 5
-			- KEY_LINE_HEIGHT * ( key_num - m_startKey + 1 );
-
-		if( y > keyAreaBottom() )
-		{
-			break;
-		}
-
-		p.fillRect( WHITE_KEY_WIDTH + 1, y - KEY_LINE_HEIGHT / 2, width() - 10, KEY_LINE_HEIGHT,
-			    markedSemitoneColor() );
-	}
-
-
 	// draw all white keys...
 	for( int y = key_line_y + 1 + y_offset; y > PR_TOP_MARGIN;
 			key_line_y -= KEY_LINE_HEIGHT, ++keys_processed )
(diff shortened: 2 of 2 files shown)


data/themes/default/style.css
```diff
@@ -146,7 +146,7 @@ PianoRoll {
 	qproperty-noteBorders: false; /* boolean property, set false to have borderless notes */
 	qproperty-selectedNoteColor: #006b65;
 	qproperty-barColor: #078f3a;
-	qproperty-markedSemitoneColor: #06170E;
+	qproperty-markedSemitoneColor: rgba(255, 255, 255, 20);
```
zonkmachine: Something like this for classic theme too?

Umcaruje: I always forget about that, will fix

Umcaruje:
image

Updated the classic theme, thanks @zonkmachine

musikBear:
Classic is actually more 'visible' -but this is so easy to customize with css, so anyone can get a personal version.
 So @Umcaruje Great work with the css options!

Umcaruje:
@musikBear I can bump up the visibility in the original theme if you want

musikBear:
@Umcaruje Not for my sake. I have made my own version of CSS, and thats working fine for *me* -but its more a question what other users think. @Mark-Agent003 reacts with 👍, so i think he should chime in, and say if he feel there are visibility issues. But all that super CSS-customizing you have made possible, is a huge help!  Big 👍

tresf:
> @Umcaruje Not for my sake.

Yes, for your sake this is an area you can help with.

But yes @Umcaruje please increase the contrast.  I have new, bright monitors yet still I have to look pretty closely to find the highlighted sections in both classic as well as default themes.

0.4.15 for comparison:

image

musikBear:
> Yes, for your sake this is an area you can help with

well what i did was to make my own idea, with black for marked notes, and a sea-blue for selected notes.
It looks like this:
mycsspianoroll
The sea-blue is not color-blind safe, so its really only a personal choice, thats work in Rebeccas design

tresf:
@musikBear oh you want to influence the note color too?  Sounds reasonable but also sounds like a different request entirely.  I was more inviting feedback on existing feature contrast, so @Umcaruje can decide whether or not to merge this.

Umcaruje:
Well from the discussion in #3651 Rebecca said that black doesn't provide enough contrast, and another thing is that black makes all the other semitones look marked even though it should be the other way around. I'll bump up the transparency the marked notes in the default theme.

As far as the notes go it's the discussion for the other pull request.

musikBear:
> oh you want to influence the note color too?

@tresf No, i just thought i might as well have both my CSS changes for piano-roll in the same picture, here the issue is the marked scale, but @Umcaruje says 
> Rebecca said that black doesn't provide enough contrast

So even though i find it easier to use, it is not for general usage, especially because it just as Umcaruje correctly points out, confuse new users, in respect to what is marked, and what isent marked -Eg *black* for marked, is contra intuitive (and should for that reason not be used :p )

zonkmachine:
I'm trying with some colour on the classic theme. Here's a comparison with lmms-1.1.3
I also think it looks a bit better with the colour marking behind the notes but I'm not particularly invested in anything theme related so I'm fine with merging this as is with the current grey.

lmms-1.1.3/classic theme
theme113

lmms-1.2.0/classic theme with `qproperty-markedSemitoneColor: rgba( 100, 255, 255, 60 );`

theme120

musikBear:
> I also think it looks a bit better with the colour marking behind the notes

I agree, it looks more '3D' like the note in ON the grid
Whit the colour marking *over* the note, the '3D' edges on the notes are eradicated, and  it looks 'flat'

Umcaruje:
@musikBear @zonkmachine  @tresf thanks for the feedback, this is what I came up with: 
screenshot from 2018-03-29 14 51 18
screenshot from 2018-03-29 14 55 04
Enough contrast or you feel it needs more?

zonkmachine:
The classic theme is still a bit on the dim side though. Why the change to an overlay in the first place?

musikBear:
Its better, but its difficult to access the ergonomic effect, without working with real thing. Maby a set of css-files could be included default. Covering subtle to extreme contrast, and then users decide for them self.
But Umcaruje, you have spend seriously long time with this design layout. Isent everything css based? In that case it could be a '*junior-job*' and be done by a non-ccp-coder?

Umcaruje:
>But Umcaruje, you have spend seriously long time with this design layout. Isent everything css based? 

I spent a lot of time with these changes because of life getting in the way. Everything is CSS based, I changed the drawing method to an overlay here though rather than a change in the background.

zonkmachine:
Well, it looks good. I think it's a merge if you're happy with it.

Umcaruje:
image

Here's the classic theme, made it to be closer to 1.1.3 in terms of color contrast.

zonkmachine:
Merge...

PR #4242, opened 2018-03-12
Fix MDI subwindow buttons contrast in the MenuBar

### LINKED ISSUE #4215: When a daughter window (e.g. Piano Roll, Pitch-bend window) is maximized, it cannot be normalized
*Edited by @tresf, corrected description*

MDI Subwindow controls are hard to see when maximized.
image

---

LMMS provides a Mximize button for all its daughter windows (e.g. Piano Roll, Pitch-bend window) (See screenshot  bellow, with red circle.)
image

If this button is pressed, LMMS maximizes that daughter window. Now this daughter window occupies the entire right side of the main window, and cannot be detached from it. The effect is as if LMMS suddenly switches to modeless.

After this, LMMS does not allow us to make this window a floating window, as before. 
Its size control (image) simply vanishes.

Umcaruje:
Fixes #4215

Fixed this by applying the idea @tresf had and changing the `qproperty-button` of our pallete to a light color. Then, I restyled all the other QPushButtons in the software as well as the `QComboBox`es and `QSlider`s which were inheriting the colors. 
image
The QSlider was also restyled.
image

### DIFF
--- a/data/themes/default/style.css
+++ b/data/themes/default/style.css
@@ -33,6 +33,26 @@ QMdiArea {
 	background-color: #111314;
 }
 
+QPushButton, QComboBox {
+	background: #3f4750;
+}
+
+QSlider::handle:horizontal {
+	background-color: #5a6572;
+	border: 1px solid #212020;
+	border-radius: 4px;
+}
+
+QSlider::add-page:horizontal {
+  background-color: #5a626b;
+	border: 1px solid #212020;
+}
+
+QSlider::sub-page:horizontal {
+  background-color: #3a3f45;
+	border: 1px solid #212020;
+}
+
 AutomationEditor {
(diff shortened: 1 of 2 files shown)


PhysSong:
`QHeaderView` and `QTabWidget` stuffs can be fixed easily like `QPushButton` with CSS. However, `QTabBar` stuff looks a little bit tricky.

Umcaruje:
Switched to @PhysSong's concept in https://github.com/LMMS/lmms/issues/4215#issuecomment-373020590

I hard-coded the buttons to a slightly off-white color. 
screenshot from 2018-03-29 15 32 15

Umcaruje:
Tested this out extensively and no objections in 4 days, merging

PR #4243, opened 2018-03-12
Change selected notes color

### LINKED ISSUE #4111: Selected notes in the Piano Roll editor do not look much different from unselected notes
Earlier, the selected bars in the Piano Roll Editor appeared deep blue, and the unselected bars appeared green. Thus they looked distinct from each other, and it was easy to tell them apart.

image

But now the selected bars have a different hue of green. As a result, it is very difficult to tell which bars are selected and which are not.
image

**Desired:** Retain the color difference so that the user can easily tell which bars are selected.
In fact, showing the selected bars in yellow would be a good choice.

Umcaruje:
Fixes #4111
Before:
screenshot from 2018-03-12 21 18 38
After:
screenshot from 2018-03-12 21 43 43

I tested this with http://www.color-blindness.com/coblis-color-blindness-simulator/ and it seems to provide sufficient contrast for color blind people. 

@raindropsfromsky @musikBear

### DIFF
--- a/data/themes/default/style.css
+++ b/data/themes/default/style.css
@@ -144,7 +144,7 @@ PianoRoll {
 	qproperty-noteColor: #0bd556;
 	qproperty-noteOpacity: 165;
 	qproperty-noteBorders: false; /* boolean property, set false to have borderless notes */
-	qproperty-selectedNoteColor: #006b65;
+	qproperty-selectedNoteColor: #064d79;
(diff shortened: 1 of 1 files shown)


musikBear:
I like it. More visible difference Good job!

raindropsfromsky:
How does the bar look when the transparency is changed? (Change velocity
and pan)

zonkmachine:
Merge?

PR #4246, opened 2018-03-13
keeps relative positions of TCOs on group moving

BaraMGB:
fixes https://github.com/LMMS/lmms/issues/2300
This is an alternative approach to: https://github.com/LMMS/lmms/pull/4043
All selected TCOs will be moved with there relatives positions to each other. The selected group snaped with the most left TCO toNearestTact.


zonkmachine:
Thanks for looking into this!

When you drag a group it will snap so the leftmost TCO is at a full tact. Thus you only get the expected result if the leftmost TCO is not shifted to start with.

PR #4249, opened 2018-03-14
Make *ModelView a templated type instead of macro-based class

Wallacoloo:
Among other things, this makes it easier to grep for FloatModelView, BoolModelView, IntModelView in the code base.


tresf:
+1, been bit by this in the past as the macro makes the types a bit harder to find declarations for.  This is much more clear.

Unfortunately, I'm not a strong C++ developer and these techniques are not something I'm yet familiar with, so I'll request the PR is reviewed by someone more qualified. :)

Wallacoloo:
Awesome! Thanks for the review @PhysSong

Issue #4130, opened 2018-01-25
File Browser Font Size


Mark-Agent003:
This was originally included in #4085, but was removed so that that PR could focus on the highlights.
My proposal is that the `font-size` in the file browser be changed to 14px. This is a simple change that goes a long way. The file browser right now uses a 12px font size, and looks too small and tightly packed. Increasing the font size looks cleaner and adds a lot towards visibility.
Here are before and after pictures:
![image](https://user-images.githubusercontent.com/25230510/35404715-923b80e8-01c9-11e8-950b-2c31cff74592.png)
I'd encourage you to edit your style.css file to include ```QTreeView { font-size: 14px; }``` and try it out for a couple days before commenting.

Umcaruje:
l would suggest using a relative unit like em because of hidpi support

On 25 Jan 2018 19:21, "Mark-Agent003" <notifications@github.com> wrote:

This was originally included in #4085
<https://github.com/LMMS/lmms/pull/4085>, but was removed so that that PR
could focus on the highlights.
My proposal is that the font-size in the file browser be changed to 14px.
This is a simple change that goes a long way. The file browser right now
uses a 12px font size, and looks too small and tightly packed. Increasing
the font size looks cleaner and adds a lot towards visibility.
Here are before and after pictures:
[image: image]
<https://user-images.githubusercontent.com/25230510/35404715-923b80e8-01c9-11e8-950b-2c31cff74592.png>
I'd encourage you to edit your style.css file to include QTreeView {
font-size: 14px; } and try it out for a couple days before commenting.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/LMMS/lmms/issues/4130>, or mute the thread

Mark-Agent003:
@Umcaruje, font size only supports px and pt (at least for Qt)

tresf:
Yeah, I don't see a good solution to this from Qt.  Here's an article sort-of explaining it... <http://doc.qt.io/qt-5/scalability.html#calculating-scaling-ratio>

musikBear:
One thing i like to add. 
If the wideness of the list-bar, is too narrow for the *whole* preset-name to be shown, a strange drop in font-size happens:
![fontsizetowidth](https://user-images.githubusercontent.com/6662066/35453149-6bf94c3e-02c9-11e8-80b5-bba879195d5e.JPG)

sufficient width : green
insufficient width : red

PR #4252, opened 2018-03-14
Increase File Browser Spacing

Mark-Agent003:
This is the PR for #4130 . This simply increases the font-size in the file browser, thus increasing visibility and overall spacing.
Here's the before/after screenshot.
image

### DIFF
--- a/data/themes/default/style.css
+++ b/data/themes/default/style.css
@@ -9,6 +9,7 @@ QLabel, QTreeWidget, QListWidget, QGroupBox, QMenuBar {
 
 QTreeView {
 	outline: none;
+	font-size: 14px;
 }
 
 QTreeWidget::item:hover,
--- a/data/themes/default/style.css
+++ b/data/themes/default/style.css
@@ -9,17 +9,23 @@ QLabel, QTreeWidget, QListWidget, QGroupBox, QMenuBar {
 
 QTreeView {
 	outline: none;
-	font-size: 14px;
+	font-size: 12px;
+}
+
+QTreeWidget::item {
+	padding: 1px;
 }
 
 QTreeWidget::item:hover,
 QTreeWidget::branch:hover {
 	background-color: #3C444E;
+	padding-left: 0px;
 }
 
 QTreeWidget::item:selected,
 QTreeWidget::branch:selected {
 	background-color: #17793b;
+	padding-left: 0px;
 }
 
 QTreeView::branch:has-children:open {


tresf:
I really don't understand this PR... 

Here's a side-by-side of 1.2.0-RC5 and Windows Explorer

image

And here's a zoomed font comparison

image

From my perspective, the font size seems sane.  I'm not sure what purpose increasing it does.

Mark-Agent003:
@tresf The current font size is certainly not unreadable, but increasing the font size is also for the sake of spacing. As you can see in your file explorer example, the items have an increased spacing as compared to ours, and our icons are also a bit crowded.
image
Other DAWs and systems have more spacing and a larger font size.
image
(Logic X, Ableton, LMMS (per this PR))
Instead of just increasing the padding, increasing the font size kills two birds in one stone: giving more space to each item while increasing visibility at the same time. Which, in a more practical scenario, is helpful to touch-screen users. 😃

tresf:
I don't agree.  If padding is needed, add padding. I'm not understanding the second bird here.  I think you accidentally killed a squirrel instead. :D

Umcaruje:
I agree, if a user desires a larger text they can easily change the css but I see no justification for a larger default size

Umcaruje:
Though we sure could use some padding in there, @tresf is right

tresf:
To really improve UI, and off-topic is to designate the presets with icons that distinguish them from each other.

This task is a loaded one because it brings the whole concept of recognizable, themed plugin icons into play, more toward the single-window ideas.

The font size isn't set in stone either, I just don't think stable-1.2 is in desparate need of this change.  The font size seems to match that which I've grown to expect.  As the software transforms, it's likely to change, but holistically.

Mark-Agent003:
> I just don't think stable-1.2 is in desparate need of this change.

@tresf  I agree, but I still like the larger font size. I've also noticed that 13px is already the browser's font size on Linux, and I definitely like the way it looks with our current theme:
image
(Images taken from Discord)
Anyways, yes, this is by no means urgent. I'd just like to point out that by switching to 13px, it becomes consistent with what's already default on Linux and removes the need for padding. I hope that I'm not beating a dead horse 😅 .

tresf:
We're really starting to embroach on what's best practice for sane, UI delivery -- what the OS says is sane -- vs -- what LMMS says is sane.

We may want to just ask the OS to begin with so we can split hairs, intentionally.  
https://stackoverflow.com/a/40380684/3196753

The font sizing is a major problem today because some components react better than others (I'll quote #2510)

> - :white_check_mark: `Green` = Scales properly.
> - :no_entry: `Red` = Does not scale properly.
> 
> image

So we seem to be in agreement to match the OS, we're just not sure which OS, and we haven't done a side-by-side with MacOS yet either.

At the end of the day, I have to defer this to the theme creators and with @RebeccaDeField gone, @Umcaruje gets to decide this for the current theme.

PhysSong:
Is it ready for merge?

tresf:
I've already expressed my opinion about this change and the theme creators haven't chimed in.  It's easy enough to revert.

Spekular:
An up-to-date screenshot would be good, as every one I've seen atill has the oncreased font size.

RebeccaDeField:
Missed this one because as you said I was not active at the time and it just got bumped to my inbox.

A screenshot of what the pr looks like in it's current state also seems like a good standard if we are deffering all ui and design choices to a few busy devs and want to fastrack like we did with the plugin redesigns.

I have no problems with the simple spacing change to make it less crowded and from what I see umcaruje commented before it seems like there is already an agreement to move forward with that aspect.

zonkmachine:
Before --> After

before

RebeccaDeField:
Looks like a nice improvement without changing drastically enough to cause problems. Thumbs up from me.

zonkmachine:
Ok. Merging.

zonkmachine:
Oh, I didn't notice the origin and destination didn't match. It was intended for stable-1.2 but ended up in master. The merge went fine though. I'm happy to cherry-picking this into stable-1.2 .

tresf:
>  I'm happy to cherry-picking this into stable-1.2 .

`stable-1.2` is for critical bug-fix only, which this PR is not.

PR #4258, opened 2018-03-17
Remove unused method and extraneous state from AutomatableModelView

Wallacoloo:
- Removed `AutomatableModel::m_dataType`. Either the `AutomatableModel` should handle _all_ functions dependent on its data type (i.e. there are no specialized `FloatModel`, `IntModel`, `BoolModel`), or it should defer datatype-dependent to the derived type via virtual methods - but not both.  
- Removed `m_hasLinkedModels`. Keeping track of it is just extra complexity when `m_linkedModels.empty()` allows the same functionality with essentially the same performance.
- Removed `hasStrictStepSize()` accessor, since nothing uses it.

### DIFF
--- a/include/AutomatableModel.h
+++ b/include/AutomatableModel.h
@@ -118,7 +118,7 @@ public:
 	template<class T>
 	inline T value( int frameOffset = 0 ) const
 	{
-		if( unlikely( m_hasLinkedModels || m_controllerConnection != NULL ) )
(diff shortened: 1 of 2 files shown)


Wallacoloo:
Apple build has been stalled for 5 hours after pushing changes to fix merge conflicts to this branch. But it built fine everywhere else & on my machine, and the changes I pushed since the last verified build are tiny, so going to cross my fingers & hit that merge button. Feel free to shout at me if I break something :|

Issue #3527, opened 2017-05-02
LMMS Filters Pop and Often Cease to Output Audio


UnityParadox:
If putting a filter or equalizer band to a frequency outside of a certain range (usually within lower numbers), the filter will occasionally "pop," becoming so loud as to push most audio outside of the maximum LMMS output. Occasionally, this issue will cause the filter or equalizer to break and halt all signal processing altogether.

Branch: stable-1.2
OS: Windows 10

softrabbit:
Any filter in particular you'd like to single out for punishment? :smiling_imp:

curlymorphic:
@UnityParadox 

Can you give clear instructions on how to reproduce this bug please

the instructions may look like this --

start new project
play a c2 on tipple oscillator
put the duel filter on , and set the lp to 20hz
 

Instructions like this this really makes bug fixing a lot easier.
I am aware these bi-quad filters can be brittle, but we  need access how often this happens as to the solution I propose.

musikBear:
propl. its the Infinite loud sound 'bug'

UnityParadox:
@curlymorphic Any filter will reproduce the issue, however it seems that the bandpass filter plugins have the worst of the issue, with lowpass and highpass coming close second, being easier and more methodical to reproduce.

To reproduce:

Open a synth instance with any waveform.

Should ENV/LFO be enabled on this synth, go to the envelope list and turn the filter on. Alternatively, add a pass filter effect in the FX tab.

Perform a quick shift from generally anywhere above ~100Hz on the filter to 0. The popping/infinitely-loud-silence should occur immediately and take effect on all LMMS audio running at the time.

As for the native equalizer, raise a peak band to a high enough volume (and possibly frequency), and this should reproduce the complete halt of processing, silencing only the intstrument/plugin output.

curlymorphic:
Having spent some time trying to reproduce this, I have failed,

PR #4260, opened 2018-03-17
Remove audio artifacts when automating the Equalizer parameters.

### LINKED ISSUE #4127: Sweeping active Equalizer bands generates audio artifacts.
Sweeping active Equalizer bands generates audio artifact (clicks & pops, DC offset), especially in the range near the minimum frequency.
Sweeping speed affects the intensity of the artifacts, and manual sweeping is more likely to generate them.

[Equalizer audio artifacts.zip](https://github.com/LMMS/lmms/files/1652208/Equalizer.audio.artifacts.zip)

curlymorphic:
The Equalizer plugin uses biquad filters, These do not like having
there parameters updating during processing, and are know to produce
clicks and DC biasing. A twin filter system has been employed with a
cross fade, to interpolate between parameters.

This has removed for the use of sample exactness, as the filter is only
updated once per frame, with interpolation provided by the cross fade.

The same filters are used as previous, ensuring un automated filtering
remains unchanged.

fixes #4127

### DIFF
--- a/plugins/Eq/EqEffect.cpp
+++ b/plugins/Eq/EqEffect.cpp
@@ -70,6 +70,8 @@ EqEffect::~EqEffect()
 
 bool EqEffect::processAudioBuffer( sampleFrame *buf, const fpp_t frames )
 {
+	const int sampleRate = Engine::mixer()->processingSampleRate();
+
 	//wet/dry controls
 	const float dry = dryLevel();
 	const float wet = wetLevel();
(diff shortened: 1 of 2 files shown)


PhysSong:
Could you look into a similar issue #3527, too? Since you said you've failed to reproduce, I'll post an example project.
[3527.zip](https://github.com/LMMS/lmms/files/1823739/3527.zip)

PR #4262, opened 2018-03-19
Keep off-grid position of TCOs on group moving

### LINKED ISSUE #2300: Off-beat samples in song editor snap to grid when moving them with a selection

PhysSong:
This is a simpler alternative of #4043. Instead of calculating relative offsets from snap points, it just quantizes the change amount of offsets.
Fixes #2300, closes #4043.

### DIFF
--- a/src/core/Track.cpp
+++ b/src/core/Track.cpp
@@ -852,10 +852,19 @@ void TrackContentObjectView::mouseMoveEvent( QMouseEvent * me )
 	else if( m_action == MoveSelection )
 	{
 		const int dx = me->x() - m_initialMousePos.x();
+		const bool snap = !(me->modifiers() & Qt::AltModifier) &&
+					me->button() == Qt::NoButton;
 		QVector<selectableObject *> so =
 			m_trackView->trackContainerView()->selectedObjects();
 		QVector<TrackContentObject *> tcos;
-		MidiTime smallest_pos, t;
+		MidiTime dtick;
+		int smallestPos;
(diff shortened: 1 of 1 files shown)


zonkmachine:
This works fine. What about moving a single track in edit mode? Should it snap to grid or keep the offset?

PhysSong:
> What about moving a single track in edit mode? Should it snap to grid or keep the offset?

I think it should snap to grid unless modifier(s) are specified.

zonkmachine:
> I think it should snap to grid unless modifier(s) are specified.

I would probably expect not snapping effect without a modifier as that is what I get with moving a group.

Maybe you can have two snap points. Grid and offset that you step through?
Example. Offset TCO at 1.2 dragging right: 1.2 -> 2.0 -> 2.2 -> 3.0 -> 3.2 etc.
No modifier key needed.

zonkmachine:
Merge?

PhysSong:
> It's already a huge improvement. If the action needs fine tuning later then we can do it in master.

> Merge?

We can easily revert this if there are some issues with it. If there are no objections, I'll merge this.

PR #4265, opened 2018-03-20
Play correct note when dragging left side of virtual keyboard

### LINKED ISSUE #1556: Odd behavior when clicking and dragging out of visible range on the left side of the instrument keyboard
When you click and drag on the keyboard for a synth and you go move your mouse beyond the visible keyboard, it doesn't act the same as when the mouse is on the visible part.

When you have your mouse lined up with the white keys and move it back and forth, it plays a chromatic scale (all notes, including black keys):
image

When you move it back and forth with your mouse lined up with the black keys, it plays only the following notes:
image

What's really strange is that this only works on the left side of the keyboard. If you try it on the right side, it behaves as you would expect.

PhysSong:
Fixes #1556.

### DIFF
--- a/src/gui/PianoView.cpp
+++ b/src/gui/PianoView.cpp
@@ -327,13 +327,22 @@ void PianoView::modelChanged()
  */
 int PianoView::getKeyFromMouse( const QPoint & _p ) const
 {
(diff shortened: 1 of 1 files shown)


tresf:
Works great!

untitled

PhysSong:
Merge?

PR #4279, opened 2018-03-30
Fix stack overflow in RemoteVstPlugin

### LINKED ISSUE #4142: Transient VST causes stack overflow in RemoteVstPlugin
This issue was discovered on Discord, not sure which operating system or version, but I have reproduced it on 1.2.0-RC5 on Windows 10. When opening an existing project containing the [Transient](http://www.vst4free.com/free_vst.php?id=1672) VST plugin, a stack overflow occurs in `RemoteVstPlugin.exe`. This seems to be because the plugin calls `hostCallback(audioMasterAutomate, ...)` in response to `setParameter`, but we call `setParameter` in [the handler for `audioMasterAutomate` ](https://github.com/LMMS/lmms/blob/27007c258a7b83b8fb89bea753c9b4d767686ef3/plugins/vst_base/RemoteVstPlugin.cpp#L1450).

DomClark:
Fixes #4142.

VST plugins use the `audioMasterAutomate` callback to notify the host of parameter changes from within the plugin, and the host uses the `setParameter` function to update the plugin with parameter changes from the host. Neither of these should be called in response to the other as the other party already has the necessary information. The issue arises because both Transient and LMMS both do this, so it could be considered a bug in the plugin as much as one in LMMS, but we can and probably ought to fix the issue on our end at least.

This PR removes the `setParameter` call from the `audioMasterAutomate` handler.

### DIFF
--- a/plugins/vst_base/RemoteVstPlugin.cpp
+++ b/plugins/vst_base/RemoteVstPlugin.cpp
@@ -1450,7 +1450,6 @@ intptr_t RemoteVstPlugin::hostCallback( AEffect * _effect, int32_t _opcode,
 		case audioMasterAutomate:
 			SHOW_CALLBACK( "amc: audioMasterAutomate\n" );
 			// index, value, returns 0
-			_effect->setParameter( _effect, _index, _opt );
 			return 0;
 
 		case audioMasterVersion:


zonkmachine:
Merge?

PhysSong:
This is an one-liner and should bring no side effects as the description says. So I'd say go.
@DomClark FYI, you can merge your pull request yourself when you feel comfortable. Squash and Merge looks like the best merge method for this PR.

Issue #3981, opened 2017-11-18
JSON project format


tresf:
Proposal to add a new JSON-backed, ZIP-encapsulated project format.

Mockup:
![image](https://user-images.githubusercontent.com/6345473/32975599-cc161a4a-cbd6-11e7-8173-002e9b10556a.png)

![image](https://user-images.githubusercontent.com/6345473/32975673-ab4f6496-cbd7-11e7-89f6-c4a3f3436e4d.png)

 * JSON based
 * Standard ZIP format
   * Provides quick access for developers to open/debug projects
   * Provides a scalable method for eventually bundling portable project files (#1257)

Before:
```xml
<?xml version="1.0"?>
<!DOCTYPE lmms-project>
<lmms-project version="1.0" creator="LMMS" creatorversion="1.1.3" type="song">
  <head timesig_numerator="4" mastervol="100" timesig_denominator="4" bpm="140" masterpitch="0"/>
  <song>
    <trackcontainer width="600" x="5" y="5" maximized="0" height="300" visible="1" type="song" minimized="0">
      <track muted="0" type="0" name="TripleOscillator" solo="0">
```

After: (mockup)
```js
{
  lmms-project: {
    version: 1.0,
    creator: "LMMS",
    creatorversion: "1.1.3",
    type: "song",
    head: {
      timesig_numerator: 4,
      mastervol: 100,
      timesig_denominator: 4,
      bpm: 140,
      masterpitch: 0
    },
    song: {
      trackcontainer: {
        width: 600,
        x: 5,
        y: 5,
        maximized: 0,
        height: 300,
        visible: 1,
        type: "song",
        minimized: 0,
        track: [
          {
            muted: 0,
            type: 0,
            name: "TripleOscillator",
            solo: 0,
```

JohannesLorenz:
Did you have plans how to handle external resources? Is this specific point different between XML and json?

tresf:
@JohannesLorenz it would just mimic the existing folder layout (`samples`, etc).  I'd even go as far as to say presets should be stashed as separate files (instead of embedding them) so that they can be easily shared with others.

vlad0337187:
Good idea.
Also I'd suggest to use YAML or CSON for this =)

tresf:
From what I've read YAML is less common for data and more common for configuration.  YAML becomes valuable if we have a lot of self-references.

Quoting j2y:

> **YAML vs JSON**
[...]

PR #4288, opened 2018-04-08
Better default working directory

### LINKED ISSUE #1135: ~/lmms in our build tutorials
Does anyone else have a problem with the fact that our build tutorials all but encourage building in the same directory that the software defaults to?

i.e. `git clone` in the home directory (which is default Terminal landing location for most flavors of Linux/Unix) puts the source code in the same folder that the software uses as its working directory.

I thought of this when @eagles051387 had to do a fresh pull of the github source.  Telling him to blow away `~/lmms` could have unintended consequences. :)

There's not an easy way to prevent this other than changing our default working directory, or making our build instructions specifically move the person from that directory prior to issuing the `git clone` command.

I never use `~/lmms` as the default working directory as I store my samples on a cloud service so I've never blown away any projects or samples, but it seems like it has the potential to mess stuff up.

Thoughts?

-Tres

tresf:
Better behavior of default working directory:

* Fresh installs of LMMS (e.g. no `~/.lmmsrc.xml`) will default to the more browsable, more sane and less conflicting `~/Documents/lmms`.
~~* No longer prompt to create the home directory.~~

**Note:** The Qt4 logic is currently untested.

Closes #1135

### DIFF
--- a/src/core/ConfigManager.cpp
+++ b/src/core/ConfigManager.cpp
@@ -26,7 +26,11 @@
 #include <QDir>
 #include <QMessageBox>
 #include <QApplication>
+#if QT_VERSION >= 0x050000
 #include <QStandardPaths>
+#else
+#include <QDesktopServices>
+#endif
 #include <QtCore/QTextStream>
 
 #include "ConfigManager.h"
@@ -51,7 +55,7 @@ ConfigManager * ConfigManager::s_instanceOfMe = NULL;
 
 ConfigManager::ConfigManager() :
 	m_lmmsRcFile( QDir::home().absolutePath() +"/.lmmsrc.xml" ),
-	#if QT_VERSION >= 0x050100
+	#if QT_VERSION >= 0x050000
 	m_workingDir( QStandardPaths::writableLocation( QStandardPaths::DocumentsLocation ) + "/lmms/"),
 	#else
 	m_workingDir( QDesktopServices::storageLocation( QDesktopServices::DocumentsLocation ) + "/lmms/"),
--- a/src/core/ConfigManager.cpp
+++ b/src/core/ConfigManager.cpp
@@ -26,6 +26,7 @@
 #include <QDir>
 #include <QMessageBox>
 #include <QApplication>
+#include <QStandardPaths>
 #include <QtCore/QTextStream>
 
 #include "ConfigManager.h"
@@ -50,7 +51,11 @@ ConfigManager * ConfigManager::s_instanceOfMe = NULL;
 
 ConfigManager::ConfigManager() :
 	m_lmmsRcFile( QDir::home().absolutePath() +"/.lmmsrc.xml" ),
-	m_workingDir( QDir::home().absolutePath() + "/lmms/"),
+	#if QT_VERSION >= 0x050100
+	m_workingDir( QStandardPaths::writableLocation( QStandardPaths::DocumentsLocation ) + "/lmms/"),
+	#else
+	m_workingDir( QDesktopServices::storageLocation( QDesktopServices::DocumentsLocation ) + "/lmms/"),
+	#endif
 	m_dataDir( "data:/" ),
 	m_artworkDir( defaultArtworkDir() ),
 	m_vstDir( m_workingDir + "vst/" ),
--- a/src/core/ConfigManager.cpp
+++ b/src/core/ConfigManager.cpp
@@ -67,6 +67,10 @@ ConfigManager::ConfigManager() :
 	m_sf2Dir( m_workingDir + SF2_PATH ),
 	m_version( defaultVersion() )
 {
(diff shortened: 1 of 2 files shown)


PhysSong:
> Fresh installs of LMMS (e.g. no `~/.lmmsrc.xml`) will default to the more browsable, more sane and less conflicting `~/Documents/lmms`.

:+1:
> No longer prompt to create the home directory.

Could you explain why did you change that?

zonkmachine:
> Note: The Qt4 logic is currently untested.

Tested, works fine.

tresf:
> No longer prompt to create the home directory.

Because asking newcomers a confusing question on first install is a bad first impression.

PhysSong:
> > No longer prompt to create the home directory.

> Because asking newcomers a confusing question on first install is a bad first impression.

I agree, but I found some situations where that behavior is undesirable. When the working directory is set by user and moved/deleted/temporarily unable, prompt is still valid.
I suggest checking `ConfigManager::inst()->value( "app", "configured" ).toInt()` if you want to detect fresh install.

tresf:
> When the working directory is set by user and moved

A warning is valid and a blocking dialog is the wrong way.

> /deleted/

I'm not sure how this is helpful to anyone in this case.  They can delete the bread crumbs.  It's of zero consequence.

> temporarily un[avail]able

This dialog offers no contextual evidence that it's temporarily unavailable and the mkdir will silently fail.  I'm not sure the use-case here.

> I suggest checking ConfigManager::inst()->value( "app", "configured" ).toInt() if you want to detect fresh install.

That would be create unpredictable behavior.  It's better to just create the directory in 99.9% of use-cases.  We shouldn't be coding to the 1%.

I assume the atypical use-case you're describing is a user-space mountpoint where a driver is mounted in `~` instead of `/media/`, etc?  Apps should create working directories.  Arguably, the `.lmmsrc.xml` should live directly inside the working directory too, it has no place in `~/.lmmsrc.xml`, but I plan to address that as part of DataFile moving from XML to JSON, related: https://github.com/LMMS/lmms/issues/3981.

PhysSong:
I use a NTFS partition as the working directory. If LMMS creates the working directory without prompt when the partition is not mounted, it will create a directory which was the path for the partition. That can break several things(bookmarks, other programs' data directories).

I think it's unusual, but I think it'd be better to cover such cases.

musikBear:
> Arguably, the .lmmsrc.xml should live directly inside the working directory too, it has no place in ~/.lmmsrc.xml

Imo in **Data**, its wrong to force-install things in C: when a user has chosen a different partition as install-point.

tresf:
> Imo in Data

I'm not sure what you're talking about.  There's no such thing as `data`.  Perhaps you're referring to `%APPDATA%`, but that's still explicit.

>  its wrong to force-install things in C:

Nothing's being force-installed anywhere, please don't make blanket statements.

tresf:
Tested this for a day, accidentally clicked the wrong folder and created a bunch of breadcrumbs.  I don't like the dialog and the experience should be better but I'll have to agree with @PhysSong and leave it out of scope for now.

Reverted and ready for review/merge.

musikBear:
> There's no such thing as data

Now you confuse me:

tresf:
@musikBear, that's unrelated to the working directory.  Install LMMS wherever you want. :)

tresf:
```
:..LMMS \data
```

The problem @musikBear is that you are making up terms.  Yes the skeleton directory is created automatically.  That's done as a convenience so that when you save something such as a preset or a sample you don't have to create the directory -- or worse -- you create it in a non-relative path and projects won't load.

> who made the descission to place files here.
> @tresf

The dev team did.  Otherwise the folder structure is unobvious and relative paths break.  This is a stop-gap courtesy until we introduce portable project files.

zonkmachine:
Shouldn't we probe for `~/lmms` or `~/Documents/lmms` when starting with no `.lmmsrc.xml` ? You'd probably want to default to an existing working directory.

tresf:
> Shouldn't we probe for ~/lmms or ~/Documents/lmms when starting with no .lmmsrc.xml ? You'd probably want to default to an existing working directory.

Sure.  `~/Documents/lmms` doesn't need a probe because that's the new default.  #1135 is what inspired this change so we'd need to know what to match on so we don't accidentally select the code repository.

zonkmachine:
Look for`QDir::home().absolutePath() + "/lmms/projects/"` ?

tresf:
> Look `forQDir::home().absolutePath() + "/lmms/projects/"` ?

Done via 052eedc.  Ready for testing/review.

zonkmachine:
Works as intended.

lukas-w:
@tresf Why did you close this?

tresf:
> @tresf Why did you close this?

Branch cleanup, it was a mistake.

PR #4290, opened 2018-04-09
Better behavior when left-clicking a TCO

Sawuare:
Instead of calling `MouseMoveEvent()`, the TCO's "text float" text and position are updated.
This prevents left-clicking the right edge of a resizable TCO from decreasing its size.

Also, removed an unused variable: `m_oldTime`

### DIFF
--- a/include/Track.h
+++ b/include/Track.h
@@ -290,8 +290,6 @@ private:
 
 	TextFloat * m_hint;
 
-	MidiTime m_oldTime;// used for undo/redo while mouse-button is pressed
-
 // qproperty fields
 	QColor m_mutedColor;
 	QColor m_mutedBackgroundColor;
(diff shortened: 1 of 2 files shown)


zonkmachine:
> This prevents left-clicking the right of a TCO from decreasing its size.

Yeah, I don't think I've seen this reported before but I can confirm the issue. You basically end up nudging the TCO a bit to the left sometimes when grabbing the edge. Fix confirmed.
It's not really an enhancement but a ux/bug and I think it's correct to address this in 1.2 .

musikBear:
I feel that it can be a tad difficult to 'grab' the move-handle on notes <= 1/16. I quite often resize the note instead. Its very simple to 'fix'. Setting magnification to 200% let me grab notes easy -Will *this fix* influence 'grabbing' of <= 1/16 notes?

zonkmachine:
> I feel that it can be a tad difficult to 'grab' the move-handle on notes <= 1/16

OT. This PR is about TCO's in the Song Editor and not about grabbing notes.

musikBear:
> OT. This PR is about TCO's in the Song Editor and not about grabbing notes.

ooops.. but Note-TCO cant change size(?) so AUT and B&B.. 
I will write a ticket for the note-grab

Issue #4284, opened 2018-04-07
Weird Behaviour of Delay Effects when Input is shorter than Delay Time


Sauerstoffdioxid:
When a note (or sequence) ends before the delay starts, it will yield weird results. Most of the time it will simply not be delayed (i.e. the wet signal is silent) or it will play the delayed sound once the effect receives another input.

## Steps to reproduce
Start LMMS, add an delay effect (preferably with feedback option), and play a short note. Or take a look at 
[this example project](https://github.com/LMMS/lmms/files/1886525/Bug.zip) showcasing the behaviour for a couple delay effects with delay time set to a quarter note.

## Expected Behaviour
All sounds are delayed/reverbed properly.

## Actual Behaviour
In the example project:

- the first pattern is reverbed/delayed in none of the instruments
- when the second pattern starts, the delay of the first one plays on top of it (at least in the top 3 instruments)
- the third pattern is played correctly by all effects since it's a quarter note long
- if the sound hasn't gone silent before the fourth pattern starts (that should be on all but Echo Delay Line), it will be played correctly
- because of the gap before the fifth pattern, the sound will become silent again. The pattern will play like the first one

Also, if you hold a piano key after the last pattern, you can hear the proper delay for that pattern.

Sauerstoffdioxid:
Another interesting thing becomes noticeable if you copy the short (e.g. the last) pattern between the 4th and the 5th pattern. (like this:)
![grafik](https://user-images.githubusercontent.com/35171164/38465033-38c5a750-3b17-11e8-8da9-4d68c7a6bfe0.png)
It too will echo as intended at first but then suddenly cut off the moment the delay of the longer pattern (3rd) would turn silent (if it weren't followed by other patterns).

Additionally, I noticed that if the sound ends before the note ends (for instance because of an envelope) and the pattern (but not the sounds) is in total longer than the delay time, the echo will play for as long as the notes are long but at least one full delay. 
[Example test file.](https://github.com/LMMS/lmms/files/1887182/Bug02-01.zip)
Note that the delay that was cut off will play once the next input is received.

musikBear:
if you change *Amount* in Vintage UI it normalizes 
Amount must be > 0.25
Maby it would be logical to have Amount default be > 0? F.i. 1.0

Sauerstoffdioxid:
@musikBear Are you referring to Calf Vintage Delay? I agree, a default Amount of 0 is not ideal, but that's not what this issue is about.

[...]

PR #4291, opened 2018-04-10
Minimum framesPerPeriod() effect decay

zonkmachine:
In response to https://github.com/LMMS/lmms/issues/4284
Increase the 1 minimum frame to `Engine::mixer()->framesPerPeriod();`

### DIFF
--- a/include/Effect.h
+++ b/include/Effect.h
@@ -104,7 +104,8 @@ public:
 	inline f_cnt_t timeout() const
 	{
 		const float samples = Engine::mixer()->processingSampleRate() * m_autoQuitModel.value() / 1000.0f;
-		return 1 + ( static_cast<int>( samples ) / Engine::mixer()->framesPerPeriod() );
+		int frames = Engine::mixer()->framesPerPeriod();
+		return frames + static_cast<int>( samples ) / frames;
 	}
 
 	inline float wetLevel() const


DomClark:
The time returned by `timeout()` is in units of buffers, so `framesPerPeriod()` probably isn't the thing to go for here. It seems there is already a one-buffer minimum, although with a buffer size of 256 at 44100Hz this is a mere 6ms, so a longer timeout would still be good. Rather than adding a constant to the value of the "Delay" knob, would it instead be worth increasing the minimum and/or default value of the knob instead?

PhysSong:
> The time returned by `timeout()` is in units of buffers, so `framesPerPeriod()` probably isn't the thing to go for here.

Correct. Simple calculation gives 1.486s for the buffer size of 256. I guess it's too long.
> although with a buffer size of 256 at 44100Hz this is a mere 6ms, so a longer timeout would still be good.

Agreed.
> would it instead be worth increasing the minimum and/or default value of the knob instead?

That's more straightforward. I think it should be at least 50ms \~ 100ms(delay and/or dynamic processors may need longer time though).
Maybe it would be better to allow users to disable the automatic stop feature for effects. That will give a simple and straightforward solution for side effects like #4284.

zonkmachine:
> Maybe it would be better to allow users to disable the automatic stop feature for effects.

Already implemented. That's the setting named **Keep effects running even without input**.

PhysSong:
> Already implemented. That's the setting named Keep effects running even without input.

Thanks! I forgot about that. I don't know if it'd be better to allow changing that per effect.

zonkmachine:
>> would it instead be worth increasing the minimum and/or default value of the knob instead?
> 
> That's more straightforward. I think it should be at least 50ms ~ 100ms(delay and/or dynamic processors may need longer time though).

I tried a minimum of 100ms but could still replicate the issue with some bpm settings. Maybe make '**Keep effects running even without input**' default on?

PhysSong:
> Maybe make **'Keep effects running even without input'** default on?

:+1:

zonkmachine:
So, something like:
1. Close this PR
2. Disable the fx auto quit function for lmms-1.2 (as below)
https://github.com/LMMS/lmms/blob/daa3f535158f321e7a01ea531fc3085f2d26ab37/src/gui/SetupDialog.cpp#L137:L138

```
	m_disableAutoQuit(ConfigManager::inst()->value( "ui",
						   "disableautoquit").toInt() ),
```
to

```
	m_disableAutoQuit(ConfigManager::inst()->value( "ui",
						   "disableautoquit", "1" ).toInt() ),
```

3. Keep #4284 open and add a better framesPerPeriod() fix at a later time?

zonkmachine:
> 2. Disable the fx auto quit function for lmms-1.2

https://github.com/LMMS/lmms/pull/4378

PR #4294, opened 2018-04-14
Fix VST effect bugs

### LINKED ISSUE #1727: VST mda Detune: not able to show controls
LMMS needs default guis for because I have plugins with a lot of controls but there is no way to get to them.
 Ie when I load them in a fx box there is no controls button to click on
capture 1

capture 2

capture3

that's why LMMS needs default guis for fx vsts  when I use these in fl they use a needs default gui.
capture

### LINKED ISSUE #4110: Strange problems when loading a project containing VST effects.
Please change the title to something more descriptive. I don't really know how to summarize this bug briefly.

I think this is best explained by a short video
[Capture_3.zip](https://github.com/LMMS/lmms/files/1629492/Capture_3.zip) (it's in the zip file)

Basically, when I load up a project with an effect plugin and open the controls window of an effect on the master channel (different channels not yet tested), there is a strange box with no (?) content, which keeps whatever was printed last in it. 
I don't know how to describe it. That's why recorded the video. However, the recording software didn't record the cursor correctly, so imagine a resize arrow whenever I'm resizing the box (or am having trouble picking up the effect window).

I think this might be related to the rendering of the VST GUI. In this screenshot I resized it over the actual GUI.
bug

And here is another image, very clearly showing that the black square of the NaN-Detector isn't drawn into the background.
bug2

Also, the affected projects I tested it in had in common, that there was a (almost) empty window of the size of the effect GUI leftover after loading the project (as seen in the video).
I can post a test project if needed.

"Works" with both plugin embedding methods: Win32 API and Qt. Although Qt is showing the detached GUI of the empty window, just not the way it should, I guess:
screenshot 7

Windows 10 64bit, LMMS 1.2.0-rc5

DomClark:
Fixes #4110, and a couple of other VST effect bugs.

Summary of my changes, since I neglected to put any with my commits (Lukas-W's are with his commits):
* __Fix X11 embedding on Qt4__: see https://github.com/LMMS/lmms/issues/4110#issuecomment-375494925.
* __Fix VST effect load crash on non-primary monitor__: LMMS would crash when on a non-primary monitor if the first embedded VST opened was an effect. Upon embedding a VST, the chain of parent windows needs to be made native, and if the main window is not on the primary monitor, this triggers a repaint/relayout before the `VstEffectControlDialog` is added to the `SubWindow`. `SubWindow` then causes a segfault trying to access properties of its `widget()`, which is still null. I've just added null checks where appropriate.
* __Fix toggling UI for non-embedded VST effects__: Fixes a bug from earlier on this branch, behaviour is the same as `stable-1.2`.
* __Fix RemoteVstPlugin not exiting when effect removed__: ~this was a regression on Windows only, caused by a8311a7. On Windows, effect control dialogs are kept around even when the view that opened them is deleted: https://github.com/LMMS/lmms/blob/e554a4c4b093ea4c319cceed98f4495484da37b0/src/gui/widgets/EffectView.cpp#L162-L176
As a result, the `QSharedPointer` in `VstEffectControlDialog` keeps the `VstPlugin` instance alive even when the effect has been removed. I've changed this to a plain `QPointer`, so it maintains the nullify-on-delete behaviour but lets the plugin exit properly.~ Now cbf8cbd8c0c66a79a1fd2965613ad5f4d38a8d99 replaces it.

**Edited** by @PhysSong: Original fix eb560a780f8bb02cc4bbc02ca893eb1b850b9b47 have been dropped and cbf8cbd8c0c66a79a1fd2965613ad5f4d38a8d99 have been added instead. Another new commit d6ae46eaddbd60027ecd8a550d630a7fb8702b59 fixes #1727.

### DIFF
--- a/plugins/VstEffect/VstEffectControlDialog.cpp
+++ b/plugins/VstEffect/VstEffectControlDialog.cpp
@@ -41,7 +41,6 @@
 #include <QToolBar>
 #include <QLabel>
 
-
 VstEffectControlDialog::VstEffectControlDialog( VstEffectControls * _ctl ) :
 	EffectControlDialog( _ctl ),
 	m_pluginWidget( NULL ),
@@ -271,6 +270,17 @@ void VstEffectControlDialog::paintEvent( QPaintEvent * )
 		tbLabel->setText( tr( "Effect by: " ) + m_plugin->vendorString() +
 			tr( "&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;<br />" ) +
 			m_plugin->currentProgramName() );
+	}
+}
+
+void VstEffectControlDialog::showEvent(QShowEvent *_se)
+{
+	EffectControlDialog::showEvent( _se );
+	// Workaround for a (unexplained) bug where on project-load the effect
+	// control window has size 0 and would only restore to the proper size upon
+	// moving the window or interacting with it.
+	if (parentWidget()) {
+		parentWidget()->adjustSize();
 	}
 }
 
--- a/plugins/VstEffect/VstEffectControlDialog.cpp
+++ b/plugins/VstEffect/VstEffectControlDialog.cpp
@@ -45,6 +45,7 @@
 VstEffectControlDialog::VstEffectControlDialog( VstEffectControls * _ctl ) :
 	EffectControlDialog( _ctl ),
 	m_pluginWidget( NULL ),
+
 	m_plugin( NULL ),
 	tbLabel( NULL )
 {
@@ -62,16 +63,10 @@ VstEffectControlDialog::VstEffectControlDialog( VstEffectControls * _ctl ) :
 		embed_vst = m_plugin->embedMethod() != "none";
 
 		if (embed_vst) {
-			m_plugin->createUI( nullptr, true );
-			m_pluginWidget = m_plugin->pluginWidget( false );
-
-#ifdef LMMS_BUILD_WIN32
-			if( !m_pluginWidget )
-			{
-				m_pluginWidget = m_plugin->pluginWidget( false );
+			if (! m_plugin->pluginWidget()) {
+				m_plugin->createUI(nullptr);
 			}
-#endif
-
+			m_pluginWidget = m_plugin->pluginWidget();
 		}
 	}
 
@@ -79,7 +74,7 @@ VstEffectControlDialog::VstEffectControlDialog( VstEffectControls * _ctl ) :
 	{
 		setWindowTitle( m_plugin->name() );
 
-		QPushButton * btn = new QPushButton( tr( "Show/hide" ) );
+		QPushButton * btn = new QPushButton( tr( "Show/hide" ));
 
 		if (embed_vst) {
 			btn->setCheckable( true );
@@ -95,6 +90,7 @@ VstEffectControlDialog::VstEffectControlDialog( VstEffectControls * _ctl ) :
 		btn->setMaximumWidth( 78 );
 		btn->setMinimumHeight( 24 );
 		btn->setMaximumHeight( 24 );
+		m_togglePluginButton = btn;
 
 		m_managePluginButton = new PixmapButton( this, "" );
 		m_managePluginButton->setCheckable( false );
@@ -295,6 +291,14 @@ void VstEffectControlDialog::togglePluginUI( bool checked )
 		return;
 	}
 
-	m_plugin->toggleUI();
+	if ( m_togglePluginButton->isChecked() != checked ) {
+		m_togglePluginButton->setChecked( checked );
+	}
+
+	if ( checked ) {
+		m_plugin->showUI();
+	} else {
+		m_plugin->hideUI();
+	}
 }
 
--- a/plugins/VstEffect/VstEffectControlDialog.cpp
+++ b/plugins/VstEffect/VstEffectControlDialog.cpp
@@ -82,8 +82,8 @@ VstEffectControlDialog::VstEffectControlDialog( VstEffectControls * _ctl ) :
 			connect( btn, SIGNAL( toggled( bool ) ),
 						SLOT( togglePluginUI( bool ) ) );
 		} else {
-			connect( btn, SIGNAL( clicked( bool ) ),
-						SLOT( togglePluginUI( bool ) ) );
+			connect( btn, SIGNAL( clicked() ),
+						m_plugin.data(), SLOT( toggleUI() ) );
 		}
 
 		btn->setMinimumWidth( 78 );
--- a/plugins/VstEffect/VstEffectControlDialog.cpp
+++ b/plugins/VstEffect/VstEffectControlDialog.cpp
@@ -62,14 +62,14 @@ VstEffectControlDialog::VstEffectControlDialog( VstEffectControls * _ctl ) :
 		embed_vst = m_plugin->embedMethod() != "none";
 
 		if (embed_vst) {
-			if (! m_plugin->pluginWidget()) {
+			if (m_plugin->hasEditor() && ! m_plugin->pluginWidget()) {
 				m_plugin->createUI(this);
 			}
 			m_pluginWidget = m_plugin->pluginWidget();
 		}
 	}
 
-	if ( m_plugin && (!embed_vst || m_pluginWidget) )
+	if (m_plugin)
 	{
 		setWindowTitle( m_plugin->name() );
 
@@ -218,7 +218,7 @@ VstEffectControlDialog::VstEffectControlDialog( VstEffectControls * _ctl ) :
 
 		int newSize = 0;
 
-		if (embed_vst) {
+		if (m_pluginWidget) {
 			newSize = m_pluginWidget->width() + 20;
 		}
 		newSize = std::max(newSize, 250);
@@ -234,7 +234,7 @@ VstEffectControlDialog::VstEffectControlDialog( VstEffectControls * _ctl ) :
 		l->addItem( new QSpacerItem( newSize - 20, 30, QSizePolicy::Fixed,
 						QSizePolicy::Fixed ), 1, 0 );
 		l->addWidget( resize, 2, 0, 1, 1, Qt::AlignCenter );
-		if (embed_vst) {
+		if (m_pluginWidget) {
 			l->addWidget( m_pluginWidget, 3, 0, 1, 1, Qt::AlignCenter );
 		}
 		l->setRowStretch( 5, 1 );
--- a/plugins/VstEffect/VstEffectControlDialog.cpp
+++ b/plugins/VstEffect/VstEffectControlDialog.cpp
@@ -64,7 +64,7 @@ VstEffectControlDialog::VstEffectControlDialog( VstEffectControls * _ctl ) :
 
 		if (embed_vst) {
 			if (! m_plugin->pluginWidget()) {
-				m_plugin->createUI(nullptr);
+				m_plugin->createUI(this);
 			}
 			m_pluginWidget = m_plugin->pluginWidget();
 		}
--- a/plugins/VstEffect/VstEffectControlDialog.cpp
+++ b/plugins/VstEffect/VstEffectControlDialog.cpp
@@ -289,7 +289,12 @@ void VstEffectControlDialog::showEvent(QShowEvent *_se)
 
 VstEffectControlDialog::~VstEffectControlDialog()
 {
-	//delete m_pluginWidget;
+#if !(QT_VERSION < 0x050000 && defined(LMMS_BUILD_LINUX))
+	if (m_pluginWidget && layout()) {
+		layout()->removeWidget(m_pluginWidget);
+		m_pluginWidget->setParent(nullptr);
+	}
+#endif
 }
 
 
--- a/plugins/VstEffect/VstEffectControlDialog.h
+++ b/plugins/VstEffect/VstEffectControlDialog.h
@@ -54,6 +54,7 @@ protected:
 private:
 	QWidget * m_pluginWidget;
 
+	QPushButton * m_togglePluginButton;
 	PixmapButton * m_openPresetButton;
 	PixmapButton * m_rolLPresetButton;
 	PixmapButton * m_rolRPresetButton;
@@ -64,7 +65,7 @@ private:
 
 	QLabel * tbLabel;
 
-private slots:
+public slots:
 	void togglePluginUI( bool checked );
 } ;
 
--- a/plugins/VstEffect/VstEffectControlDialog.h
+++ b/plugins/VstEffect/VstEffectControlDialog.h
@@ -50,6 +50,7 @@ public:
 
 protected:
 	virtual void paintEvent( QPaintEvent * _pe );
+	virtual void showEvent( QShowEvent* _se ) override;
 
 private:
 	QWidget * m_pluginWidget;
--- a/plugins/vestige/vestige.cpp
+++ b/plugins/vestige/vestige.cpp
@@ -107,10 +107,12 @@ public:
 	void createUI( QWidget *parent ) override
 	{
 		Q_UNUSED(parent);
-		VstPlugin::createUI( nullptr );
 		if ( embedMethod() != "none" ) {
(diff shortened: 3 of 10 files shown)


zonkmachine:
It tests fine like before but I'm not heavily into VSTs and wouldn't know what to look for.

plugins/VstEffect/VstEffectControlDialog.cpp
```diff
@@ -47,7 +47,8 @@ VstEffectControlDialog::VstEffectControlDialog( VstEffectControls * _ctl ) :
 	m_pluginWidget( NULL ),
 
 	m_plugin( NULL ),
-	tbLabel( NULL )
+	tbLabel( NULL ),
+	m_needsEmbed( false )
```
lukas-w: This appears to be equivalent to `m_plugin && m_plugin->embetMethod() != "none"`. If so, can we remove the variable and make it a method instead in order to reduce redundancy?

plugins/VstEffect/VstEffectControlDialog.cpp
```diff
@@ -81,9 +81,12 @@ VstEffectControlDialog::VstEffectControlDialog( VstEffectControls * _ctl ) :
 		if (embed_vst) {
 			btn->setCheckable( true );
 			btn->setChecked( true );
-		}
-		connect( btn, SIGNAL( toggled( bool ) ),
+			connect( btn, SIGNAL( toggled( bool ) ),
 					SLOT( togglePluginUI( bool ) ) );
+		} else {
+			connect( btn, SIGNAL( clicked() ),
+					m_plugin.data(), SLOT( toggleUI() ) );
```
lukas-w: Why was this change needed? Can we fix it within `togglePluginUI` or `toggleUI` instead?

plugins/VstEffect/VstEffectControlDialog.h
```diff
@@ -67,7 +67,7 @@ class VstEffectControlDialog : public EffectControlDialog
 	PixmapButton * m_managePluginButton;
 	PixmapButton * m_savePresetButton;
 
-	QSharedPointer<VstPlugin> m_plugin;
+	QPointer<VstPlugin> m_plugin;
```
lukas-w: Can we get rid of the Windows-specific behaviour instead? Maybe the bug this workaround was made for has been fixed.

plugins/VstEffect/VstEffectControlDialog.cpp
PhysSong: We can't, because `VstEffectControlDialog::showEvent` may modify the value? I guess it would be safe to replace `embed_vst` with this variable if `showEvent` is never called while construction.

plugins/VstEffect/VstEffectControlDialog.h
PhysSong: The workaround is very old(more than 5 years ago): https://github.com/LMMS/lmms/commit/1c9c76f3994a2054b5f4cb2f6b24c7f9b7fd1a10#diff-e3cda83bb57677b7ecd60d7e5e1fe92cR156
There have been so many changes after that(Qt version change, VST bug fixes, etc.) I think we should test whether the bug still exists.

PhysSong:
Maybe out of scope... I have a fix for #1727. Is it fine to include the fix here?

Umcaruje:
>I have a fix for #1727. Is it fine to include the fix here?

Depends on the size of the patch, though this PR already fixes a lot of things, your fix probably could be more suited as a separate PR, unless the fix is dependent on the changes made in this PR.

PhysSong:
The fix changes a few lines and not directly related to this PR.
FYI, there will be minor merge conflicts when I open a separate pull request.

Umcaruje:
>The fix changes a few lines and not directly related to this PR.
>FYI, there will be minor merge conflicts when I open a separate pull request.

Then I guess you could push here so we avoid the conflicts

plugins/VstEffect/VstEffectControlDialog.h
DomClark: Removing the workaround, I can still reproduce the issue:
1. Enable "One instrument track window mode"
2. Create a project with at least two instruments
3. Add a VST effect to one of them
4. Open the other instrument's window
5. Open the first instrument's window again and show the VST effect dialog

The VST GUI will have disappeared.

I'm sure there's a better way to work around this, but the X11 embedding on Qt4 seems to be especially fragile and I'm not going to have access to a Linux machine again for a couple of months, so I didn't want to touch this any more. If you can produce something better, it would be very welcome.

plugins/VstEffect/VstEffectControlDialog.cpp
DomClark: `m_needsEmbed` is a flag to indicate that the VST GUI needs to be embedded but hasn't been yet. The VST GUI seems to be deleted on Qt4 with X11 embedding if the first thing it's added into isn't currently part of the main window hierarchy, so I deferred the embedding to the first time `showEvent` is called, when the `VstEffectControlDialog` will be safely inside its subwindow. This could be replaced by `m_plugin && m_plugin->embetMethod() != "none"`, but the GUI would be re-added to the dialog every time `showEvent` is called.

plugins/VstEffect/VstEffectControlDialog.h
PhysSong: > Removing the workaround, I can still reproduce the issue:

Could you tell me your OS and Qt version you used? If I can reproduce that, I'll try to figure out what's going on.

DomClark: Windows 10 64-bit, Qt 5.4.2.

PhysSong: I can reproduce the bug on GNU/Linux, too.
I found that the plugin widget is destroyed when `VstEffectControlDialog` is destroyed because of parenting. Simple unparenting seems to fix this.

plugins/VstEffect/VstEffectControlDialog.cpp
PhysSong: > The `toggled(bool)` signal isn't emitted if the button isn't checkable

If we can uncheck the button when VST window is closed, we can always make the button checkable. AFAIK we do something similar for ZynAddSubFX.

src/gui/widgets/EffectView.cpp
```diff
@@ -109,7 +110,9 @@ EffectView::EffectView( Effect * _model, QWidget * _parent ) :
 		{
 			m_subWindow = gui->mainWindow()->addWindowedWidget( m_controlView );
 			m_subWindow->setSizePolicy( QSizePolicy::Fixed, QSizePolicy::Fixed );
-			m_subWindow->setFixedSize( m_subWindow->size() );
+			if (m_subWindow->layout()) {
+				m_subWindow->layout()->setSizeConstraint(QLayout::SetFixedSize);
+			}
```
PhysSong: @lukas-w I don't know if it's needed, but isn't it safe to keep previous behavior(probably with `else`)?

plugins/VstEffect/VstEffectControlDialog.h
PhysSong: @DomClark Is it fine to remove eb560a780f8bb02cc4bbc02ca893eb1b850b9b47 from this PR and add new fix which uses unparenting?

src/gui/widgets/EffectView.cpp
lukas-w: @PhysSong I don't think it's needed, `setSizePolicy( QSizePolicy::Fixed, QSizePolicy::Fixed )` already does that in the `else` case.

plugins/VstEffect/VstEffectControlDialog.h
DomClark: Sure, go ahead 😃

DomClark:
I've mostly implemented Umcaruje's suggestions from https://github.com/LMMS/lmms/issues/4110#issuecomment-381682629, apart from VST windows being always-on-top on Linux. Can anyone help with this? On Windows, if a window is given an owner it will always be displayed on top of it. However, this requires the HWND of the parent, so in order to do this using Win32 code in RemoteVstPlugin, one would have to convert LMMS's native window ID into an HWND for Wine, which I don't know how to do (or if it's possible). Alternatively, this could be done natively, but again I have no idea how.

plugins/VstEffect/VstEffectControlDialog.h
PhysSong: Unparenting approach dosen't work well for XEmbed + Qt4. So I think I need to wrap that with QT version check macro if I do that.

zonkmachine:
@DomClark Is there anything left to do here? I'm testing this and the few vst effects I have open up fine and gui show/hide is remembered.

lukas-w:
@zonkmachine: As far as I know, @PhysSong still wanted to make some adjustments.

PhysSong:
I want to make two changes(one is fixing `RemoteVstPlugin` exiting issue in new way, the other is fixing #1727), but they're optional.

qnebra:
I got small issue, minor one with this PR.

1. Ubuntu 18.04, wine 3.0, latest master with only this PR merged.

2. Add new VST instrument track to project.
3. Add plugin, anything but instrument. In my case sforzando and arminator.
4. Work with them.
5. "Clone this track"

Expected behaviour:
Vestige track is cloned and vst plugin window shows.

Now:
Vestige track is cloned, but vst window is missing. Blank Vestige window.

plugins/vestige/vestige.cpp
```diff
@@ -127,6 +177,13 @@ void vestigeInstrument::loadSettings( const QDomElement & _this )
 	{
 		m_plugin->loadSettings( _this );
 
+		m_plugin->createUI( nullptr );
```
PhysSong: Why do we need this call? At this moment `loadFile` already have created UI, so this call will cause the bug reported by @qnebra.

qnebra: Hmmm, interesing.

plugins/VstEffect/VstEffectControlDialog.h
PhysSong: It seems like destroying parent window also destroys child windows on Windows. I guess it was the reason for the workaround.
https://msdn.microsoft.com/en-us/library/windows/desktop/ms632682(v=vs.85).aspx

Might it be better to use a separate helper class to handle `win32` embedding?

plugins/vestige/vestige.cpp
qnebra: @PhysSong this call make a issue. I commented this out, compile lmms and check. Cloning works correctly.

PhysSong:
Should we do a rebase and squash fixup commits? Squashing everything doesn't look good, but ordinary merge might mess up commit history.

zonkmachine:
There is also the 'Rebase and merge' option.

PhysSong:
I've done https://github.com/LMMS/lmms/pull/4294#issuecomment-389686965 on top of rebased branch. Would it be fine to (force-)push that?

zonkmachine:
Yes. I think so.

PhysSong:
I've done a interactive rebase and cleaned up the commit history. A fix for https://github.com/LMMS/lmms/pull/4294#issuecomment-389819074 is included in 9449c07ac593c7ccc0bb44ae32592cab8ecce38d(the original commit 102345ba57d6c82a26445eb939862ade534d864d was the cause of this bug).
cbf8cbd8c0c66a79a1fd2965613ad5f4d38a8d99 now replaces eb560a780f8bb02cc4bbc02ca893eb1b850b9b47. Since the fix is not needed for Qt4 + XEmbed(it even breaks embedding for situation), I've wrapped it with a pair of macro.
d6ae46eaddbd60027ecd8a550d630a7fb8702b59 is a simple fix for #1727. It was added here per https://github.com/LMMS/lmms/pull/4294#issuecomment-381627977. As of #3786, it's possible to use VST effect dialogs without VST's GUI. So fixing #1727 doesn't require many changes. Also, it will be easy to revert this if it causes any issues.

zonkmachine:
> d6ae46e is a simple fix for #1727

I get a crash when testing this with mda Detune (as used in #1727). Crash occurs in the controls window when opening the control button (control VST-plugin from LMMS host) after pressing the Show/Hide button a bunch of times.


```
VstPlugin::showUI called before VstPlugin::createUI 
VstPlugin::showUI called before VstPlugin::createUI 
VstPlugin::showUI called before VstPlugin::createUI 
VstPlugin::showUI called before VstPlugin::createUI 
Segmentation fault (core dumped)

...

Core was generated by `./lmms'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  QMdiSubWindow::widget (this=0x34ec450) at widgets/qmdisubwindow.cpp:2361
2361	widgets/qmdisubwindow.cpp: No such file or directory.
(gdb) bt
#0  QMdiSubWindow::widget (this=0x34ec450) at widgets/qmdisubwindow.cpp:2361
#1  0x00007f56f4529bc0 in VstEffectControls::managePlugin (this=0x7f57157b0608)
    at /home/zonkmachine/builds/lmms/lmms/plugins/VstEffect/VstEffectControls.cpp:165
#2  0x00007f56f4530282 in VstEffectControls::qt_static_metacall (_o=0x7f57157b0628, _c=QMetaObject::InvokeMetaMethod, _id=1, _a=0x7ffe50d73980)
    at /home/zonkmachine/builds/lmms/lmms/build/plugins/VstEffect/moc_VstEffectControls.cxx:59
#3  0x00007f57141bb87a in QMetaObject::activate (sender=0x33a0a90, m=m@entry=0x7f57153f72a0 <QAbstractButton::staticMetaObject>, 
    local_signal_index=local_signal_index@entry=2, argv=argv@entry=0x7ffe50d73980) at kernel/qobject.cpp:3539
#4  0x00007f5714f72172 in QAbstractButton::clicked (this=<optimized out>, _t1=false) at .moc/release-shared/moc_qabstractbutton.cpp:219
#5  0x00000000005f85fb in AutomatableButton::mouseReleaseEvent (this=0x33a0a90, _me=0x7ffe50d73df0)
    at /home/zonkmachine/builds/lmms/lmms/src/gui/widgets/AutomatableButton.cpp:151
#6  0x0000000000619c03 in PixmapButton::mouseReleaseEvent (this=0x33a0a90, _me=0x7ffe50d73df0)
    at /home/zonkmachine/builds/lmms/lmms/src/gui/widgets/PixmapButton.cpp:94
```

PhysSong:
> I get a crash when testing this with mda Detune

It looks like an old bug. To reproduce:
1. Load any VST effects(an effect dialog will be show up)
2. Click control button in the effect dialog
3. Close the effect dialog and then the control dialog
4. Reopen the effect dialog
5. Click the button in step 2, LMMS will crash

I'll look into this bug.

PhysSong:
> pressing the Show/Hide button

I think we should disable this button when the effect has no GUI. Is it worth to disable it for VSTi too?

PhysSong:
https://github.com/LMMS/lmms/pull/4294#issuecomment-390852701 seems to be introduced in 908591ba7dc186e942305e76463029643aa21593. The problem is `Qt::WA_DeleteOnClose` attribute set by `MainWindow::addWindowedWidget`, so removing that will be sufficient.
I also found that `manageVestigeInstrumentView` and `manageVSTEffectView` are doing almost identical things. I'd like to merge them into one class(something like `VSTPluginManageView`), but it looks like out of scope of this pull request.

zonkmachine:
> I think we should disable this button when the effect has no GUI.

Yes.


> Is it worth to disable it for VSTi too?

For VSTi's without a gui?

PhysSong:
Fixed a crash reported by @zonkmachine. I think Show/Hide button should be disabled for VSTs without GUI, but I think that's an UI/UX enhancement rather than a bug fix. So I think this one could be merged after enough reviews/tests.

zonkmachine:
> Fixed a crash reported by @zonkmachine

Fix confirmed.

> I think this one could be merged after enough reviews/tests.

:+1:

zonkmachine:
I've tested this some more and find nothing else to comment on. The mda Detune issue above I ascribe to mda Detune.

PhysSong:
Should we wait for more testings, or merge this soon?

PR #4297, opened 2018-04-15
Responsive "Effects chain" & "User controller" LEDs

### LINKED ISSUE #3620: "Effects chain" & "User controller" LEDs should be responsive
This button effects chain turns on automatically when an effect is added in the effects chain of a mixer track or in the effects chain of an instrument plug-in. Why doesn't it turn off automatically when there are no plug-ins, too?
Lesser clicks and time, better UX.

Sawuare:
- Turn off the effects chain LED when its chain is empty.
- Turn on the user controller LED when a controller is selected.

Also, removed an unused function: `setEnabled()`

Closes #3620.

### DIFF
--- a/src/gui/widgets/EffectRackView.cpp
+++ b/src/gui/widgets/EffectRackView.cpp
@@ -138,6 +138,10 @@ void EffectRackView::deletePlugin( EffectView* view )
 	Effect * e = view->effect();
 	m_effectViews.erase( qFind( m_effectViews.begin(), m_effectViews.end(), view ) );
 	delete view;
+	if( m_effectViews.size() == 0 )
+	{
+		fxChain()->m_enabledModel.setValue( false );
+	}
 	fxChain()->removeEffect( e );
 	e->deleteLater();
 	update();
--- a/src/gui/widgets/EffectRackView.cpp
+++ b/src/gui/widgets/EffectRackView.cpp
@@ -138,10 +138,6 @@ void EffectRackView::deletePlugin( EffectView* view )
 	Effect * e = view->effect();
 	m_effectViews.erase( qFind( m_effectViews.begin(), m_effectViews.end(), view ) );
 	delete view;
-	if( m_effectViews.size() == 0 )
-	{
-		fxChain()->m_enabledModel.setValue( false );
-	}
 	fxChain()->removeEffect( e );
 	e->deleteLater();
 	update();
@@ -235,7 +231,6 @@ void EffectRackView::addEffect()
 
 	Effect * fx = esd.instantiateSelectedPlugin( fxChain() );
 
-	fxChain()->m_enabledModel.setValue( true );
 	fxChain()->appendEffect( fx );
 	update();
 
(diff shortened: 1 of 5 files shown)


src/gui/widgets/EffectRackView.cpp
```diff
@@ -138,6 +138,10 @@ void EffectRackView::deletePlugin( EffectView* view )
 	Effect * e = view->effect();
 	m_effectViews.erase( qFind( m_effectViews.begin(), m_effectViews.end(), view ) );
 	delete view;
+	if( m_effectViews.size() == 0 )
+	{
+		fxChain()->m_enabledModel.setValue( false );
+	}
```
PhysSong: Since `EffectChain::clear` already have the same functionality, I think you can implement this in `EffectChain::removeEffect` instead.

Sawuare: Done in https://github.com/LMMS/lmms/pull/4297/commits/d3e65cad0bc1aff9df7fa286172ed371f387c295

Wallacoloo:
Using (or even having) a `dataUnchanged()` signal on a model smells a bit fishy. LMMS mostly uses a model-view-controller-like paradigm (MVC), and usually the model emits signals to notify the views that the model's changed.

But in this case, the `dataUnchanged()` signal is being used to notify the view that the _controller_ attempted to do something - not that the model has been updated. If we wanted to hook into something the controller's doing (I think LMMS mostly merges the controller role into the view), the signal should be placed right at the source - on the controller.

I don't think I'm capable of giving a more coherent explanation of why I think `dataUnchanged()` is a bad thing to have or use, but it feels like a leaky abstraction - like we're losing the main advantage of MVC in that the model and view are no longer decoupled. If you can see a way to achieve this work without using `dataUnchanged`, I'd recommend that. I'm unfortunately not familiar enough with LMMS to present such a solution (and if you don't see a better way of doing it, that's okay - I'm still happy to see you implement this feature & I'm not going to hold up the PR over something like that).

PhysSong:
Something more about `dataUnchanged`... We have two use cases: one in `BBTrackContainer`'s constructor, and the other in `Song`'s constructor. However, I guess they doesn't do anything. What I can see about `BBTrackContainer` is 00d2d4a5d3a9ab591f8a25178bc02f0c4cad4cdf, but I don't know why we still have this.

> Using (or even having) a `dataUnchanged()` signal on a model smells a bit fishy.

I think the intention was notify that user interacted with a controller when the value isn't changed(ex. select current item in combo box). However, I think we can cover such cases on controller's side(`ComboBox` in this case).
> If you can see a way to achieve this work without using `dataUnchanged`, I'd recommend that.

> and if you don't see a better way of doing it, that's okay

I think so. :+1:

Sawuare:
Thanks for your input, @Wallacoloo and @PhysSong.

> I think the intention was notify that user interacted with a controller when the value isn't changed(ex. select current item in combo box).

Exactly. I don't know a way to achieve this without using `dataUnchanged()`.

PR #4307, opened 2018-04-21
Revert contributor file removal

### LINKED ISSUE #2914: CONTRIBUTORS is not in source tarball
File `CONTRIBUTORS` should be kept in the tree and a bot should update it. This information is missing when building from a source tarball.

lukas-w:
Re-introduces the code that generates the `CONTRIBUTORS` file using git, removed by via #3016. The intent by @jasp00 was to generate it by an automated task on our servers (see https://github.com/LMMS/lmms.io/pull/198) that would commit it into the source tree, but this was never finished (I don't know the exact status on this). 

image

Closes #2914.


tresf:
@jasp00's arguments were valid... `git` shouldn't be a hard-requirement and builds should be reproducible 100% of the time.  That said... we've been shipping a blank contributor's list with all RCs because that PR never got completed.  Thanks @lukas-w!

Wallacoloo:
Looks good to me.

PR #4310, opened 2018-04-22
Refactoring: Remove duplicate code

lukas-w:
Evidently there's been some serious copy-pasting going on in our code. This PR removes some duplicate code.

### DIFF
--- a/src/core/RenderManager.cpp
+++ b/src/core/RenderManager.cpp
@@ -88,30 +88,7 @@ void RenderManager::renderNextTrack()
 		// for multi-render, prefix each output file with a different number
 		int trackNum = m_tracksToRender.size() + 1;
 
-		// create a renderer for this track
-		m_activeRenderer = make_unique<ProjectRenderer>(
-				m_qualitySettings,
-				m_outputSettings,
-				m_format,
-				pathForTrack(renderTrack, trackNum));
-
-		if ( m_activeRenderer->isReady() )
-		{
-			// pass progress signals through
-			connect( m_activeRenderer.get(), SIGNAL( progressChanged( int ) ),
-					this, SIGNAL( progressChanged( int ) ) );
-
-			// when it is finished, render the next track
-			connect( m_activeRenderer.get(), SIGNAL( finished() ),
-					this, SLOT( renderNextTrack() ) );
-
-			m_activeRenderer->startProcessing();
-		}
-		else
-		{
-			qDebug( "Renderer failed to acquire a file device!" );
-			renderNextTrack();
-		}
+		render( pathForTrack(renderTrack, trackNum) );
 	}
 }
 
@@ -154,11 +131,16 @@ void RenderManager::renderTracks()
 // Render the song into a single track
 void RenderManager::renderProject()
 {
+	render( m_outputPath );
+}
+
+void RenderManager::render(QString outputPath)
+{
 	m_activeRenderer = make_unique<ProjectRenderer>(
 			m_qualitySettings,
 			m_outputSettings,
 			m_format,
-			m_outputPath);
+			outputPath);
 
 	if( m_activeRenderer->isReady() )
 	{
@@ -166,7 +148,8 @@ void RenderManager::renderProject()
 		connect( m_activeRenderer.get(), SIGNAL( progressChanged( int ) ),
 				this, SIGNAL( progressChanged( int ) ) );
 
-		// as we have not queued any tracks, renderNextTrack will just clean up
+		// when it is finished, render the next track.
+		// if we have not queued any tracks, renderNextTrack will just clean up
 		connect( m_activeRenderer.get(), SIGNAL( finished() ),
 				this, SLOT( renderNextTrack() ) );
 
--- a/src/core/RenderManager.cpp
+++ b/src/core/RenderManager.cpp
@@ -158,7 +158,7 @@ void RenderManager::render(QString outputPath)
 	else
 	{
 		qDebug( "Renderer failed to acquire a file device!" );
-		emit finished();
+		renderNextTrack();
 	}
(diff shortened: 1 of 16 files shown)


src/gui/MainWindow.cpp
```diff
@@ -1497,43 +1497,32 @@ void MainWindow::fillTemplatesMenu()
...
-		m_templatesMenu->addAction(
-					embed::getIconPixmap( "project_file" ),
-					( *it ).left( ( *it ).length() - 4 ) );
+		if ( templates.size() && ! m_templatesMenu->actions().isEmpty() )
+		{
+			m_templatesMenu->addSeparator();
+		}
+
+		for( QStringList::iterator it = templates.begin();
+							it != templates.end(); ++it )
```
PhysSong: Maybe out of scope, but I think it's a good chance to replace this with a range-based loop.

src/core/SampleBuffer.cpp
```diff
@@ -75,73 +75,60 @@ SampleBuffer::SampleBuffer( const QString & _audio_file,
...
 		m_origFrames = _frames;
+		update();
 	}
-	connect( Engine::mixer(), SIGNAL( sampleRateChanged() ), this, SLOT( sampleRateChanged() ) );
-	update();
 }
 
 
 
-
```
PhysSong: This looks unintentional and unneeded.

src/core/SampleBuffer.cpp
```diff
@@ -156,6 +143,7 @@ void SampleBuffer::sampleRateChanged()
 }
 
 
+
```
PhysSong: This looks unintentional and unneeded.

src/core/RenderManager.cpp
```diff
@@ -88,30 +88,7 @@ void RenderManager::renderNextTrack()
...
-					this, SLOT( renderNextTrack() ) );
-
-			m_activeRenderer->startProcessing();
-		}
-		else
-		{
-			qDebug( "Renderer failed to acquire a file device!" );
-			renderNextTrack();
-		}
+		render( pathForTrack(renderTrack, trackNum) );
```
Wallacoloo: Previously if one track failed to render (`m_activeRenderer->isReady()` returns false), we would still call `renderNextTrack()` to continue rendering the remaining tracks. The new behavior is that once one track returns false in `m_activeRenderer->isReady()`, we end rendering completely and `emit finished()`.

I don't know when `isReady()` is expected to return false - should we worry about preserving the previous behavior?

src/core/main.cpp
```diff
@@ -223,7 +224,17 @@ void fileCheck( QString &file )
 	}
 }
 
+int usageError(const QString& message)
+{
+	qCritical().noquote() << QString("\n%1.\n\nTry \"%2 --help\" for more information.\n\n")
```
Wallacoloo: I think you should not append a '.' to the `message` (`%1.`) here, otherwise the error messages for most of these messages will have two periods at the end of the error message. You will have to add a period to `noInputFileError` though.

src/core/main.cpp
```diff
@@ -223,7 +224,17 @@ void fileCheck( QString &file )
...
+				   .arg( message ).arg( qApp->arguments()[0] );
+	return EXIT_FAILURE;
+}
 
+int noInputFileError()
+{
+	return usageError( "No input file specified" );
```
PhysSong: This one doesn't finish with '.', while the others do. As @Wallacoloo said, I think you should add a period here and remove one in `usageError`.

src/core/RenderManager.cpp
PhysSong: Can we replace `emit finished()` with `renderNextTrack()`?

lukas-w: @PhysSong I think that should work. Done via daee35ebe86bf91c4c4ba40b1600e5f927ae1d08.

src/core/main.cpp
lukas-w: Removed '.' from the rest of the messages via daee35ebe86bf91c4c4ba40b1600e5f927ae1d08.

PhysSong:
@lukas-w CircleCI build on `master` now fails because of this PR: `noquote()` is introduced in Qt 5.4 and Ubuntu 14.04 ships with [Qt 5.2.1](https://packages.ubuntu.com/trusty/qtbase5-dev).
I guess we should change something in [LMMS/lmms-ci-docker](https://github.com/LMMS/lmms-ci-docker) or find some alternative ways.

lukas-w:
@PhysSong I'll update CircleCI to 17.10 and submit a PR.

PhysSong:
FYI, Ubuntu 18.04 LTS will be released soon.

PhysSong:
@lukas-w Is there any reason to switch to `qCritical().noquote()` instead of plain `printf`? Since some old OS(ex. Ubuntu 14.04) still ships with Qt < 5.4, I guess it would better not to use such a function for compatibility.

src/core/SamplePlayHandle.cpp
```diff
@@ -49,34 +49,20 @@ SamplePlayHandle::SamplePlayHandle( const QString& sampleFile ) :
...
 
 
 
 
 SamplePlayHandle::SamplePlayHandle( SampleTCO* tco ) :
-	PlayHandle( TypeSamplePlayHandle ),
-	m_sampleBuffer( sharedObject::ref( tco->sampleBuffer() ) ),
-	m_doneMayReturnTrue( true ),
-	m_frame( 0 ),
-	m_ownAudioPort( false ),
```
PhysSong: After this PR, `m_ownAudioPort` is always set to true and cause crash when (un)mute sample track while playing. #3947 will fix it.

PR #4321, opened 2018-04-25
Use atomics to count shared_object without locks

mrkline:
C++11 (and subsequent C++ standards) provide portable ways to issue
atomic hardware instructions, which allow multiple threads to load,
store, and modify integers without taking a lock. The standard also
defines a memory model that lets you express the ordering guarantees
around these atomic operations. (x86 is relatively strongly-ordered, but
many other common architectures, such as ARM, are free to reorder loads
and stores unless told not to.)

This patch removes the lock from shared_object and replaces it with the
standard thread-safe reference counting implementation used in
C++'s std::shared_ptr, Rust's std::sync::Arc, and many others.

Additional resources on the topic:
https://assets.bitbashing.io/papers/concurrency-primer.pdf
https://www.youtube.com/watch?v=ZQFzMfHIxng

### DIFF
--- a/include/shared_object.h
+++ b/include/shared_object.h
@@ -26,15 +26,13 @@
 #ifndef SHARED_OBJECT_H
 #define SHARED_OBJECT_H
 
-#include <QtCore/QMutex>
-
+#include <atomic>
 
 class sharedObject
 {
 public:
 	sharedObject() :
-		m_referenceCount( 1 ),
-		m_lock()
+		m_referenceCount(1)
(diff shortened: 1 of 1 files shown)


PhysSong:
> Ultimately, I think `sharedObject` should just be replaced by using `std::shared_ptr` or something similar

Currently it's used for detuning and sample buffers. I guess replacing them won't be difficult.

Anyway, :+1: for this pull request!

PR #4326, opened 2018-04-26
Axe atomic int

mrkline:
This is a series of commits that migrates away from AtomicInt towards the C++11 standard library types.
This allows us to remove Qt version tests (e.g. `#if QT_VERSION > 0x050000`) and simplifies some of the surrounding code.


include/LocklessList.h
```diff
@@ -95,7 +85,7 @@ class LocklessList
 
 
 private:
-	QAtomicPointer<Element> m_first;
+    std::atomic<Element*> m_first;
```
Wallacoloo: Should be tab-indented (not spaces) for consistency.

src/core/LocklessAllocator.cpp
```diff
@@ -101,27 +101,27 @@ static int ffs( int i )
...
 			return NULL;
 		}
 	}
-	while( !m_available.testAndSetOrdered( available, available - 1 ) );
+	while (!m_available.compare_exchange_weak(available, available - 1));
 
-	size_t startIndex = m_startIndex.fetchAndAddOrdered( 1 )
-							% m_freeStateSets;
-	for( size_t set = startIndex;; set = ( set + 1 ) % m_freeStateSets )
+	const size_t startIndex = ++m_startIndex % m_freeStateSets;
```
Wallacoloo: Shouldn't the above be post-increment: `m_startIndex++`?

src/core/NotePlayHandle.cpp
```diff
@@ -598,7 +598,7 @@ void NotePlayHandleManager::release( NotePlayHandle * nph )
 {
 	nph->NotePlayHandle::~NotePlayHandle();
 	s_mutex.lockForRead();
-	s_available[ s_availableIndex.fetchAndAddOrdered( 1 ) + 1 ] = nph;
+	s_available[ s_availableIndex++ + 1 ] = nph;
```
Wallacoloo: Seems like this can now be further simplified to
```
s_available[++s_availableIndex] = nph;
```

src/core/NotePlayHandle.cpp
```diff
@@ -614,7 +614,7 @@ void NotePlayHandleManager::extend( int c )
 
 	for( int i=0; i < c; ++i )
 	{
-		s_available[ s_availableIndex.fetchAndAddOrdered( 1 ) + 1 ] = n;
+		s_available[ s_availableIndex++ + 1 ] = n;
```
Wallacoloo:
```
s_available[++s_availableIndex] = n;
```

Wallacoloo:
It looks like the test cases are sometimes failing. I see `LocklessAllocator: Invalid pointer` in the log, so maybe some of the logic in the LocklessAllocator wasn't quite preserved? https://travis-ci.org/LMMS/lmms/jobs/371385216

include/LocklessList.h
mrkline: Doi. So much for having my vim settings fixed up.

src/core/LocklessAllocator.cpp
mrkline: Yes.

mrkline:
@Wallacoloo - I've pushed fixups per your feedback. Hopefully your comment on the `LocklessAllocator` fixes whatever was causing CI to be unhappy.

Once things are fixed up to everybody's satisfactions, what's LMMS's usual PR process? Should I rebase those changes into the appropriate original commits?

src/core/LocklessAllocator.cpp
```diff
@@ -101,27 +101,27 @@ static int ffs( int i )
...
+	const size_t startIndex = m_startIndex++ % m_freeStateSets;
+	for (size_t set = startIndex;; set = ( set + 1 ) % m_freeStateSets)
 	{
-		for( int freeState = m_freeState[set]; freeState != -1;
-						freeState = m_freeState[set] )
+		for (int freeState = m_freeState[set]; freeState != -1;)
 		{
 			int bit = ffs( ~freeState ) - 1;
-			if( m_freeState[set].testAndSetOrdered( freeState,
+			if (m_freeState[set].compare_exchange_weak(freeState,
```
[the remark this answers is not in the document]
Wallacoloo: `compare_exchange_weak` *looks* fine to me here. As you say, spurious failure means we just try again, and the logic of the loop is such that when we get a spurious failure _no_ state is changed, so functionally it should do the same thing as `compare_exchange_strong`.

include/LocklessList.h
```diff
@@ -53,39 +53,29 @@ class LocklessList
...
+		while (!m_first.compare_exchange_weak(e->next, e,
+				std::memory_order_release,
+				std::memory_order_relaxed))
 		{
-#if QT_VERSION >= 0x050000
-			e->next = m_first.loadAcquire();
-#else
-			e->next = m_first;
-#endif
+			// Empty loop (compare_exchange_weak updates e->next)
```
PhysSong: Why did you change this loop logic?

PhysSong:
`std::atomic` doesn't seem to set initial value to `0`/`nullptr`, while `QAtomicPointer` does. It seems to cause crash.

include/LocklessList.h
mrkline: I didn't; as mentioned in one of the commit messages, and in the comment here, `compare_exchange_*` functions update their first argument to the most recent value if they fail.

src/core/LocklessAllocator.cpp
```diff
@@ -101,27 +101,27 @@ static int ffs( int i )
 
 void * LocklessAllocator::alloc()
 {
-	int available;
+	// Some of these CAS loops could probably use relaxed atomics, as discussed
+	// in http://en.cppreference.com/w/cpp/atomic/atomic/compare_exchange.
+	// Let's use sequentially-consistent ops to be safe for now.
+	int available = m_available.load();
```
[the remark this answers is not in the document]
mrkline: I didn't; as mentioned in one of the commit messages, `compare_exchange_*` functions update their first argument to the most recent value if they fail.

mrkline:
@PhysSong I think you've found the issue I was having. For some reason I misremembered and thought that `atomic_int` initialized itself to zero. Hopefully c9a5435 sorts things out.

mrkline:
FWIW, I rebased the suggested fixups and tweaks into the original commits.

PR #4338, opened 2018-05-01
backport jack transport from https://github.com/gi0e5b06/lmms.git

ycollet:
I tried to backport the jack transport changes from https://github.com/gi0e5b06/lmms.git.
I compiles fine on my repo and it works.
Certainly some more changes will be required.
I performed some tests with xjadeo + lmms (jack slave) both controled by qjackctl and some times, lmms forgot to play some notes. I doesn't checked the processor load during the run of xjadeo + lmms.

### DIFF
--- a/include/AudioJack.h
+++ b/include/AudioJack.h
@@ -41,13 +41,15 @@
 
 #include "AudioDevice.h"
 #include "AudioDeviceSetupWidget.h"
+#include "ITransport.h"
 
+class QComboBox;
 class QLineEdit;
 class LcdSpinBox;
 class MidiJack;
 
 
-class AudioJack : public QObject, public AudioDevice
+class AudioJack : public QObject, public AudioDevice, virtual public ITransport
 {
 	Q_OBJECT
 public:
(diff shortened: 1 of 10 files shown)


PhysSong:
This pull request contains many unnecessary and unwanted changes, and it crashes when I don't use JACK. Since the original branch was significantly diverged from mainstream, it will need many fixes.

ycollet:
I performed a diff betwwen the files which have changes related to jack transport.
On lmms I was on the master branch and on the other repo I was also on the master branch. I know that the other repo is checked out from 1.2.0rc3 but only the required changes are backported. So I don't understand why you say that the branch have diverged.
Could you have some more informations related to the crashes ?

PhysSong:
> So I don't understand why you say that the branch have diverged.

Many commits have been added into mainstream `master` since the branch is created. The downstream branch isn't well-maintained and has a number of unnecessary/unrelated changes.
> but only the required changes are backported.

This pull request reverts large parts of #3711, which is probably undesired. There are some more issues(ex. formatting, redundant changes).
>  Could you have some more informations related to the crashes ?

You need to set audio interface to anything except JACK and press play button. The reason is: `s_transport` is set to `NULL` on initialization and has non-null value if and only if `AudioJack` sets a proper value.

After addressing these points, we still need code reviews and testings.

PhysSong:
I'll close this if it won't be cleaned up. I think clean (re-)implementation can be better than this

Issue #4217, opened 2018-03-05
In the Song Editor, the gear icon should have actions associated with its context menu 


raindropsfromsky:
In the Song Editor tracks, the gear icon is the only place where the context menu is triggered with LMB, rather with RMB. In fact, the RMB has no response at all.

On the same track, all the other elements have context menus triggered with RMB.
(Mute+Solo LEDs, VOL, PAN, and elements on timeline)

**Desired:**  Even the gear icon should have context menu, for the sake of consistency.

tresf:
This is fairly common.  In Firefox, for example, you can right-click the toolbar and have actions -- or -- left click the settings stack and have actions.

![image](https://user-images.githubusercontent.com/6345473/36991331-616f93e4-2075-11e8-8ffb-fdecfa5d1056.png)  ![image](https://user-images.githubusercontent.com/6345473/36991341-684b9654-2075-11e8-8149-1f799f5e2937.png)

When a context menu has a dedicated button, you don't need a right-click trigger.  It's a throwback to File, Edit, View.  I do not consider this a bug.

raindropsfromsky:
Well, in LMMS, there is a distinct difference between a "context menu" action list and the LMB actions.

For example, when we click with LMB, the LEDS turn on/off. The VOL/PAN knobs turn. The label pops up the instrument window. 

Only the gear icon does exactly with LMB what the other things do with RMB.
So why have this exception, if there are no other constraints?

raindropsfromsky:
The only notable exception is Blender, where the object is selected with RMB, instead of LMB.
And this is purposefully [done](https://www.youtube.com/watch?v=rcwk6zvBnVU). Not so in LMMS.

tresf:
> Only the gear icon does exactly with LMB what the other things do with RMB.
> So why have this exception, if there are no other constraints?

Because the gear otherwise has no other action.  The gear is a popup trigger and actions can be popup triggers.  This is exactly how popup triggers work inside a UI.

An example of a popup-trigger with left-click in LMMS is a combo box, as can be observed in the Zoom dropdown.  The popup for zoom is the primary action.  Secondary-actions such as reset, copy, paste will show in a combo box if right-clicked.  It's not often that a menu is used for a primary action, but it's also not uncommon nor nonstandard or even clunky to use.  The gear is supposed to do this.

To switch to right-click would actually be inconsistent.  We plan on moving the File, Edit, View to a similar UI element. 

[...]

PR #4343, opened 2018-05-06
Do not rename track labels when they are double-clicked

### LINKED ISSUE #4985: Song Editor: Rename tracks with left click.
This is mentioned in # 4978. I can rename the tracks with a right click quickly and this is fine; but I can also rename with two clicks with the left button.

ezgif com-video-to-gif

This last option is very annoying because when you open an instrument or an editor quickly, it is unintentionally renamed. I think they should remove the renaming tracks feature with double left click and only rename with the right click.

Sawuare:
When a track-label-button is clicked to open/close an instrument/sample window, it can be easily double-clicked by mistake, which will show its label line-edit, and that is unwanted. This change prevents that.

This has been suggested by a user on our forum: https://lmms.io/forum/viewtopic.php?f=15&t=27239

Closes #4985.

### DIFF
--- a/include/TrackLabelButton.h
+++ b/include/TrackLabelButton.h
@@ -52,7 +52,6 @@ protected:
 	virtual void dragEnterEvent( QDragEnterEvent * _dee );
 	virtual void dropEvent( QDropEvent * _de );
 	virtual void mousePressEvent( QMouseEvent * _me );
-	virtual void mouseDoubleClickEvent( QMouseEvent * _me );
 	virtual void mouseReleaseEvent( QMouseEvent * _me );
 	virtual void paintEvent( QPaintEvent * _pe );
 	virtual void resizeEvent( QResizeEvent * _re );
--- a/src/gui/widgets/TrackLabelButton.cpp
+++ b/src/gui/widgets/TrackLabelButton.cpp
@@ -167,14 +167,6 @@ void TrackLabelButton::mousePressEvent( QMouseEvent * _me )
 
 
 
-void TrackLabelButton::mouseDoubleClickEvent( QMouseEvent * _me )
-{
-	rename();
-}
-
-
-
-
 void TrackLabelButton::mouseReleaseEvent( QMouseEvent *_me )
 {
 	if( m_buttonRect.contains( _me->globalPos(), true ) && m_renameLineEdit->isHidden() )


Umcaruje:
I disagree with this. I think double-clicking to rename is a very useful. Especially when #4217 gets implemented and rename gets tucked away to a context menu.

Wallacoloo:
Right now double-clicking anywhere in the black area triggers a rename:
image

I wonder if a better solution would be to make it so that the rename functionality is only triggered by double clicking on the _text_ area, and not on the _icon_. That way it's possible to quickly toggle the editors without triggering a rename by rapidly clicking on the icon, but it remains possible to quickly rename a track by double-clicking on the text.

SecondFlight:
I like that idea a lot. It gives the user an option to click fast, while still keeping the ability to easily rename. Best of both worlds as I see it.

Sawuare:
From my experience, double-clicking to rename is not a standard behavior.
And the way double-click renaming is done currently is strange. On first click, the track's window is opened, then on second click, it is closed and the lineedit is instantiated. This PR prevents that, too.

SecondFlight:
Right, because single clicking already does something. That didn't cross my mind. I can definitely see the argument against it.

Sawuare:
Merge?

BaraMGB:
I agree with you on the fact that double click for rename is something we should get rid of. But we should also be consist within lmms. So if change the behavior on the track label button we have to change it on every place a rename is possible (e.g. Mixer channel, ect.)

Also in my opinion we don't should use right click for renaming. This should be reserved for a context menu. Bitwig uses [ctrl] +[LMB] for entering the renaming. I vote for this.

PhysSong:
Summarizing some discussions from Discord:
- Renaming on double-click is fine only if it doesn't toggle instrument windows or something like that.
- Add the renaming function into the track operation menu, i.e., the menu which appears when you press the gear icon.
- RMB should trigger a context menu the same as the track operation menu instead of renaming.

Also, quoting https://github.com/LMMS/lmms/pull/3356#issuecomment-281096115 for references:
> I tested to use a QTimer for waiting and check if the click could be a double click. After 400 ms the Instrument Window appears. But this delay makes the whole behavior laggy

If we need to distinguish double- and single-clicks, [`QApplication::doubleClickInterval`](https://doc.qt.io/qt-5/qapplication.html#doubleClickInterval-prop) might be useful.

Sawuare:
I will close this PR and focus on implementing https://github.com/LMMS/lmms/issues/4217#issuecomment-370841321, which will fix what I intended to fix with this PR and other issues as well.

PR #4347, opened 2018-05-10
Turn off noise channels in default patch on FreeBoy and Nescaline

### LINKED ISSUE #4331: Nescaline & Freeboy need to have noise CH disabled pr. default
both instruments has special noise-generator, and it is wrongfully default enabled.
in Nescaline it is CH4 that is the noise-generator.
It should be disabled default

In Freeboy 
CH 1 (oneshot-effects)
and
CH4 (noise channel)
Should be default disabled

Users complain about 'awful distorting instruments' because of this oversight.

SecondFlight:
Fixes #4331. Seems straightforward enough, I've tested it and it works. Not much can really go wrong here.


Wallacoloo:
Looks fine to me. I'll assume you touched the right controls, because those variable names mean nothing to me :/

Would love to hear someone else voice 'yay' or 'nay' towards changing the defaults. But so long as there isn't any objection, this seems reasonable.

Spekular:
I personally never saw any justification for why they should be off by
default.

Users complaining about the noise seems a bit absurd to me. You wouldn't
use a moog filter and then complain about the resonance, so why use an
emulator if all you want is a clean waveform?

SecondFlight:
Yeah, I actually don't care much if this isn't accepted, and I would tend to agree with @Spekular.

PR #4348, opened 2018-05-10
Allow renaming of FX mixer channels with the F2 and enter keys.

### LINKED ISSUE #3157: MIXER - Rename Channel With F2 Hotkey Or Maybe The Enter Key
Hello Everyone,

I was wondering what everyone thought about assigning the F2 button as a hotkey when MIXER is selected to rename the channel label?  Following the Windows standard of using F2 to rename things.
Or maybe even using the "Enter" key (which I would prefer over F2) might work nice because you can use the Enter key to close the "Rename FX Channel" dialog box already as well so it might be a smooth use of it hitting ENTER to enter the rename and then ENTER again to close it.

Just a thought

SecondFlight:
Will resolve #3157. Currently the enter key isn't being hit, and I have no idea why.

```c++
case Qt::Key_F2:
	renameChannel( m_currentFxLine->channelIndex() );
case Qt::Key_Enter:
	renameChannel( m_currentFxLine->channelIndex() );
```

I don't like the code duplication here, but I don't believe there's a way to capture multiple cases in one case statement. It could be reasonable to change to an if/else block just to avoid that.

### DIFF
--- a/include/FxMixerView.h
+++ b/include/FxMixerView.h
@@ -100,6 +100,9 @@ public:
 	void moveChannelLeft(int index, int focusIndex);
 	void moveChannelRight(int index);
 
+	// rename the channel
+	void renameChannel(int index);
+
 	// make sure the display syncs up with the fx mixer.
 	// useful for loading projects
 	void refreshDisplay();
--- a/include/FxMixerView.h
+++ b/include/FxMixerView.h
@@ -100,7 +100,6 @@ public:
 	void moveChannelLeft(int index, int focusIndex);
 	void moveChannelRight(int index);
 
-	// rename the channel
 	void renameChannel(int index);
 
 	// make sure the display syncs up with the fx mixer.
--- a/include/FxLine.h
+++ b/include/FxLine.h
@@ -79,6 +79,8 @@ public:
 
 	static const int FxLineHeight;
 
+	void renameChannel();
+
 private:
 	void drawFxLine( QPainter* p, const FxLine *fxLine, bool isActive, bool sendToThis, bool receiveFromThis );
 	QString elideName( const QString & name );
@@ -98,7 +100,6 @@ private:
 	QGraphicsView * m_view;
 
 private slots:
-	void renameChannel();
 	void renameFinished();
 	void removeChannel();
 	void removeUnusedChannels();
--- a/include/FxLine.h
+++ b/include/FxLine.h
@@ -81,6 +81,8 @@ public:
 
 	void renameChannel();
 
+	bool eventFilter (QObject *dist, QEvent *event);
+
 private:
 	void drawFxLine( QPainter* p, const FxLine *fxLine, bool isActive, bool sendToThis, bool receiveFromThis );
 	QString elideName( const QString & name );
--- a/src/gui/FxMixerView.cpp
+++ b/src/gui/FxMixerView.cpp
@@ -535,11 +535,6 @@ void FxMixerView::keyPressEvent(QKeyEvent * e)
 			break;
 		case Qt::Key_Enter:
 		case Qt::Key_Return:
-		{
-			/*bool accepted = e->isAccepted();
-			if ( e->isAccepted() )
-				break;*/
-		}
 		case Qt::Key_F2:
 			renameChannel( m_currentFxLine->channelIndex() );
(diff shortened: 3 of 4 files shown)


Wallacoloo:
C++ case statements are "fall through", meaning that once you jump into the body of a case, the program runs sequentially until you reach the end brace of the _switch_ block. `break` will exit the switch block. So you can do something like this:

```
switch(e->key()) {
case Qt::Key_F2:
        // does nothing; falls through to the case below.
case Qt::Key_Enter:
	renameChannel( m_currentFxLine->channelIndex() );
        break; // exit the switch statement so as not to run into the cases below.
// [more cases below]
}
```

This is a common enough idiom that you'd usually write it without the comments.

SecondFlight:
Alright, I'll be sure to use that. Figured there must be a way.

SecondFlight:
Now there's an issue where pressing enter to confirm the name doesn't work at all. I'm guessing this is because pressing enter causes the rename to trigger again. I'll look at this more tomorrow.

include/FxMixerView.h
```diff
@@ -100,6 +100,9 @@ class EXPORT FxMixerView : public QWidget, public ModelView,
 	void moveChannelLeft(int index, int focusIndex);
 	void moveChannelRight(int index);
 
+	// rename the channel
```
Sawuare: This comment is useless and should be removed.

SecondFlight: Yeah, I see what you're saying. I was following the rest of the code there, it looks more out of place without the comment IMO. I could remove all these comments as they are all pretty redundant, but for now I'll just remove the one I added.

image

SecondFlight:
I've run into a snag here, and I'd like to reach out for suggestions before I go off and subclass `QLineEdit`.

----------------

#### The short version

Pressing enter after you've finished making your changes to the channel name does save your changes, but the event is not killed and it propagates after the edit is completed. My code that handles the enter key is called, causing it to go right back into edit mode.

----------------

#### The longer version

`editingFinished()` on the `QLineEdit` for the channel name is connected to `renameFinished()` in [FxLine.cpp](https://github.com/SecondFlight/lmms/blob/stable-1.2/src/gui/widgets/FxLine.cpp) via this line:

```c++
connect( m_renameLineEdit, SIGNAL( editingFinished() ), this, SLOT( renameFinished() ) );
```

`editingFinished()` is emitted when the Return or Enter key is pressed or the line edit loses focus. ([source from docs](http://doc.qt.io/qt-5/qlineedit.html#editingFinished))

Based on [this forum post](https://forum.qt.io/topic/31928/pressing-enter-in-my-lineedit-also-presses-enter-on-a-button/8), along with other sources, it would seem that `editingFinished()` does not call `->accept()` on the event, meaning that the event will still propagate through and trigger other event listeners further up the chain. This means that somewhere along the line, `keyPressEvent` in [FxMixerView.cpp](https://github.com/SecondFlight/lmms/blob/fx-mixer-rename-shortcut/src/gui/FxMixerView.cpp) gets called after the edit is finished with `return` as the pressed key, which causes it to go right back into edit mode.

I could be mistaken about something, but assuming all this is correct, it would seem that the only way to fix it is to subclass from `QLineEdit` to change the behavior. This is such a simple change though, and it really feels like there should be a simpler way to do it.

Or I could use a global variable, but that seems like a bad idea :stuck_out_tongue_closed_eyes:

Wallacoloo:
@SecondFlight Are you familiar with [event filters](https://doc.qt.io/archives/qt-4.8/eventsandfilters.html) (scroll down half way)? _I'm not sure they do what you want_ - I haven't really studied Qt much - but maybe they'd provide a way around your problem?

justnope:
Not sure if this will help but remember to check for both Qt::Key_Enter and Qt::Key_Return

SecondFlight:
@Wallacoloo I tried using those, but I think I may have been using them wrong. I have an idea to try with them that I think might work.

SecondFlight:
Yep, I was blocking the event, but I wasn't actually handling it. This is definitely the right way to do it.

SecondFlight:
It appears that the macOS build on Travis is failing. Is there someone with macOS builds set up that might be willing to test this branch?

Issue #4345, opened 2018-05-07
Hard-coded library paths in CMake


lukas-w:
We currently hard-code the paths of libraries we bundle in our packages in `make install` for Windows platforms:

https://github.com/LMMS/lmms/blob/c1b2728d656a41345d76854fa28fcf597a774296/src/CMakeLists.txt#L181-L218

This breaks `make install` for MinGW toolchains other than Toby's Trusty PPA and for MSVC.

Instead of hard-coding, we should use `FindPackage` to locate libraries.

justnope:
About 90% of my time in issuing the latest patch for #4352 went into finding a solution of not hard-coding those paths. I eventually gave up. I know it's possible but my cmake knowledge isn't that advanced and internet searches didn't help.

Not hard-coding paths would also protect us when lib x suddenly uses lib y. Ideally, we don't need change anything, let the (cmake) system figure out what libs are needed (to be installed).

lukas-w:
> About 90% of my time in issuing the latest patch for #4352 went into finding a solution of not hard-coding those paths. I eventually gave up.

Yeah this is not a trivial task and is best done by someone with experience with CMake. Until someone finds the time to work on this, it's probably better to use hard-coded paths for new systems than not supporting them at all.

PhysSong:
For Qt 5, this should work.
```CMake
# Qt libraries
SET(QT_PKG_LIBS Qt5::Core Qt5::Gui Qt5::Widgets Qt5::Xml)
FOREACH (QT_PKG_LIB ${QT_PKG_LIBS})
	GET_TARGET_PROPERTY(Qt5Location ${QT_PKG_LIB} IMPORTED_LOCATION_RELEASE)
	INSTALL(FILES "${Qt5Location}" DESTINATION .)
ENDFOREACH()
# Qt platform plugin
GET_TARGET_PROPERTY(Qt5Location Qt5::QWindowsIntegrationPlugin IMPORTED_LOCATION_RELEASE)
INSTALL(FILES "${Qt5Location}" DESTINATION ./platforms)
```

lukas-w:
[...]

PR #4352, opened 2018-05-11
Msvc/vst-rebase : msvc support

justnope:
Here's some proposed changes to compile using MSVC. I think using vcpkg would be ideal since it has built in support for cmake (find_package where supported). And as an added bonus it's also works on linux (although I don't see a reason for using it there)

It obviously needs more testing, but this isn't the master branch so I think we're ok.

# using vcpkg
## install
git clone https://github.com/microsoft/vcpkg.git
cd vcpkg
.\bootstrap-vcpkg.bat

## installing libs 4 lmms
.\vcpkg.exe install fftw3 libsndfile libsamplerate qt5

## cmake usage
cmake "-DCMAKE_TOOLCHAIN_FILE=\<vcpkgdir\>/scripts/buildsystems/vcpkg.cmake" \<lmms src dir\>


plugins/vst_base/CMakeLists.txt
```diff
@@ -27,33 +27,27 @@ SET(EXTERNALPROJECT_CMAKE_ARGS
 	"-DCMAKE_RUNTIME_OUTPUT_DIRECTORY_DEBUG=${CMAKE_RUNTIME_OUTPUT_DIRECTORY}"
 	"-DCMAKE_MODULE_PATH=${CMAKE_MODULE_PATH}"
 	"-DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}"
+	"-DCMAKE_TOOLCHAIN_FILE=${CMAKE_TOOLCHAIN_FILE}"
```
lukas-w: This needs to be moved to the `MSVC`-specific part. It conflicts with the toolchain file passed in line 70.

src/core/audio/AudioSdl.cpp
```diff
@@ -64,14 +64,22 @@ AudioSdl::AudioSdl( bool & _success_ful, Mixer*  _mixer ) :
 	m_audioHandle.callback = sdlAudioCallback;
 	m_audioHandle.userdata = this;
 
-  	SDL_AudioSpec actual; 
-
+	SDL_AudioSpec actual;
```
lukas-w: FYI, these SDL2 fixes appear to be redundant with those from #3947. I think it's best to drop a414255d9e55648f93e833c571fac8636c5c5fef from this PR to avoid conflicts.

CMakeLists.txt
```diff
@@ -166,10 +169,15 @@ FIND_PACKAGE(Qt5Test)
 SET(QT_QTTEST_LIBRARY Qt5::Test)
 
 # check for libsndfile
-PKG_CHECK_MODULES(SNDFILE REQUIRED sndfile>=1.0.18)
-IF(NOT SNDFILE_FOUND)
-	MESSAGE(FATAL_ERROR "LMMS requires libsndfile1 and libsndfile1-dev >= 1.0.18 - please install, remove CMakeCache.txt and try again!")
-ENDIF()
+IF(USING_VCPKG)
```
lukas-w: Ideally we'd hide these details in a `FindSndFile.cmake` script and only call `FIND_PACKAGE(SndFile)` here, but I guess that's out of scope for this PR so it's fine this way for now.

justnope:
I've added the changes you requested and got most of the tests to pass but I don't know why the travis continuous integration is failing. I can't reproduce that error on my machine.

I removed the install of remotevstplugin64 because appimage was complaining about it. No idea if that's correct or not.

CMakeLists.txt
justnope: I agree. 

I'm a bit relaxed about coding standards because this is only a proof of concept. If it works, I'll implement those changes. If it doesn't, I'll look for other package managers in windows (scoop?) and see if I can implement those.

justnope:
It's now possible to create an NSIS installer package using msvc.
If there are no major issues, I suggest we merge. I can keep on adding to this pull request, but if we merge this pr and also the fix for SDL then others should be able to get msvc running pretty quickly.

lukas-w:
> It's now possible to create an NSIS installer package using msvc.

@justnope Although we're trying to get away from the hard-coded install paths in CMake (see #4345), this is a great addition! Thanks! :+1: 

Thank you for your hard work on this! It's great to see the Windows platform get some attention. I'll go through the changes again and leave some comments about minor code-style issues.

As far as I can see, `RemoteVstPlugin64` and `RemoteVstPlugin64.exe.so` aren't installed or packaged on Linux yet. We can track this in another PR if you like.

plugins/vst_base/CMakeLists.txt
```diff
@@ -24,40 +24,38 @@ SET(EXTERNALPROJECT_CMAKE_ARGS
 	"-DLMMS_SOURCE_DIR=${CMAKE_SOURCE_DIR}"
 	"-DLMMS_BINARY_DIR=${CMAKE_BINARY_DIR}"
 	"-DCMAKE_RUNTIME_OUTPUT_DIRECTORY=${CMAKE_RUNTIME_OUTPUT_DIRECTORY}"
-	"-DCMAKE_RUNTIME_OUTPUT_DIRECTORY_DEBUG=${CMAKE_RUNTIME_OUTPUT_DIRECTORY}"
 	"-DCMAKE_MODULE_PATH=${CMAKE_MODULE_PATH}"
 	"-DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}"
 )
 
+IF(MSVC)
+	LIST(APPEND EXTERNALPROJECT_CMAKE_ARGS "-DCMAKE_TOOLCHAIN_FILE=${CMAKE_TOOLCHAIN_FILE}")
```
lukas-w: Since this is only used in the `ELSEIF(LMMS_BUILD_WIN64 AND MSVC) ` case, you can just move it down there, so that it reads from line 50:

```cmake
ExternalProject_Add(RemoteVstPlugin32
	"${EXTERNALPROJECT_ARGS}"
	CMAKE_GENERATOR ${GENERATOR}
	CMAKE_GENERATOR_TOOLSET ${CMAKE_GENERATOR_TOOLSET}
	CMAKE_ARGS
		"${EXTERNALPROJECT_CMAKE_ARGS}"
		-DCMAKE_TOOLCHAIN_FILE="${CMAKE_TOOLCHAIN_FILE}"
		"-DCMAKE_PREFIX_PATH=${QT_32_PREFIX}"
)
```

justnope:
I've tested msvc 2015 & 2017/32 & 64 bits. It all seems to work after implementing your code review suggestions (thanks btw).

It's definitely not finished as in merge into the master branch, but I think it should be safe enough to merge into msvc/vst-rebase.

TODO:
-Find out why sdl2.dll isn't copied into the build directory
-Create a junction (symbolic link in linux, windows symlinks have a slightly different meaning) in the build directory for the data dir (easier debugging if implemented in cmake).
-fluidsynth is present in vcpkg but not picked up, etc...

justnope:
I've made another branch from this branch msvc/ng. Last time I did this the things in pushed in the new branch also ended up in the PR which wasn't the intention.

I might have screwed up the git commands when that happened, but in any case if you see extra commits they are from the new branch.

lukas-w:
@justnope Only commits you push to this branch will end up in the PR. I'll merge shortly. :+1:

include/stdshims.h
```diff
@@ -7,8 +7,10 @@
 #include <memory>
 #include <utility>
 
-#if (__cplusplus >= 201402L)
+#if (__cplusplus >= 201402L || _MSC_VER)
```
PhysSong: VS 2012(`_MSC_VER == 1700`) doesn't have `std::make_unique` while 2013 (`_MSC_VER == 1800`) has. This will cause compile errors when using VS 2012.

PR #4356, opened 2018-05-14
Give our threads names

Wallacoloo:
Previously all our threads were named `QThread`. The thread name shows up when listing the processes with `ps`, and also when debugging in `gdb`. Providing a real thread name makes stack dumps & such that much easier to read.

Example before/after:
```gdb
(gdb) info threads
  Id   Target Id         Frame 
* 1    Thread 0x7ffff7f90800 (LWP 10519) "lmms" 0x00007ffff367ecd9 in poll () from /usr/lib/libc.so.6
  2    Thread 0x7fffe8d5b700 (LWP 10523) "QXcbEventReader" 0x00007ffff367ecd9 in poll () from /usr/lib/libc.so.6
  3    Thread 0x7fffdec30700 (LWP 10524) "gmain" 0x00007ffff367ecd9 in poll () from /usr/lib/libc.so.6
  4    Thread 0x7fffde42f700 (LWP 10525) "gdbus" 0x00007ffff367ecd9 in poll () from /usr/lib/libc.so.6
  5    Thread 0x7fffdcfd5700 (LWP 10526) "QDBusConnection" 0x00007ffff367ecd9 in poll () from /usr/lib/libc.so.6
  6    Thread 0x7fffcb06a700 (LWP 10528) "QThread" 0x00007ffff7bc707c in pthread_cond_wait@@GLIBC_2.3.2 () from /usr/lib/libpthread.so.0
  7    Thread 0x7fffca869700 (LWP 10529) "QThread" 0x00007ffff7bc707c in pthread_cond_wait@@GLIBC_2.3.2 () from /usr/lib/libpthread.so.0
  8    Thread 0x7fffca068700 (LWP 10530) "QThread" 0x00007ffff7bc707c in pthread_cond_wait@@GLIBC_2.3.2 () from /usr/lib/libpthread.so.0
  9    Thread 0x7ffff7ef0700 (LWP 10531) "PulseHotplug" 0x00007ffff367edd6 in ppoll () from /usr/lib/libc.so.6
  10   Thread 0x7fff9693e700 (LWP 10532) "SDLAudioP1" 0x00007ffff367edd6 in ppoll () from /usr/lib/libc.so.6
  11   Thread 0x7fff9613d700 (LWP 10533) "MidiAlsaSeq" 0x00007ffff367ecd9 in poll () from /usr/lib/libc.so.6
  12   Thread 0x7fff9593c700 (LWP 10534) "QThread" 0x00007ffff7bc707c in pthread_cond_wait@@GLIBC_2.3.2 () from /usr/lib/libpthread.so.0
  17   Thread 0x7fff75850700 (LWP 10540) "dconf worker" 0x00007ffff367ecd9 in poll () from /usr/lib/libc.so.6
```
after:
```gdb
(gdb) info threads
  Id   Target Id         Frame 
* 1    Thread 0x7ffff7f90800 (LWP 12501) "lmms" 0x00007ffff367ecd9 in poll () from /usr/lib/libc.so.6
  2    Thread 0x7fffe8d5b700 (LWP 12505) "QXcbEventReader" 0x00007ffff367ecd9 in poll () from /usr/lib/libc.so.6
  3    Thread 0x7fffdec30700 (LWP 12506) "gmain" 0x00007ffff367ecd9 in poll () from /usr/lib/libc.so.6
  4    Thread 0x7fffde42f700 (LWP 12507) "gdbus" 0x00007ffff367ecd9 in poll () from /usr/lib/libc.so.6
  5    Thread 0x7fffdcfd5700 (LWP 12508) "QDBusConnection" 0x00007ffff367ecd9 in poll () from /usr/lib/libc.so.6
  6    Thread 0x7fffcb06a700 (LWP 12510) "MixerWorkerThre" 0x00007fffb1b93e19 in runDysonCompress () from /home/colin/src/lmms/build/plugins/ladspa/dyson_compress_1403.so
  7    Thread 0x7fffc2869700 (LWP 12511) "MixerWorkerThre" 0x00007ffff7bca3ac in __lll_lock_wait () from /usr/lib/libpthread.so.0
  8    Thread 0x7fffca869700 (LWP 12512) "MixerWorkerThre" 0x00007ffff7bc707c in pthread_cond_wait@@GLIBC_2.3.2 () from /usr/lib/libpthread.so.0
  9    Thread 0x7ffff7ef0700 (LWP 12515) "PulseHotplug" 0x00007ffff367edd6 in ppoll () from /usr/lib/libc.so.6
  10   Thread 0x7fff96794700 (LWP 12516) "SDLAudioP1" 0x00007ffff367edd6 in ppoll () from /usr/lib/libc.so.6
  11   Thread 0x7fff95f93700 (LWP 12517) "MidiAlsaSeq" 0x00007ffff367ecd9 in poll () from /usr/lib/libc.so.6
  12   Thread 0x7fff95792700 (LWP 12518) "QThread" 0x00005555556b8e32 in MixerWorkerThread::JobQueue::wait() ()
```

Part of the text gets cut off for longer thread names, but it still gives one something to grep for, at least.

### DIFF
--- a/include/AudioAlsa.h
+++ b/include/AudioAlsa.h
@@ -38,9 +38,9 @@
 #include "AudioDevice.h"
 
 
-class AudioAlsa : public AudioDevice, public QThread
+class AudioAlsa : public QThread, public AudioDevice
 {
-	// Public classes and enums
+	Q_OBJECT
 public:
 	/**
 	 * @brief Contains the relevant information about available ALSA devices
--- a/include/AudioDummy.h
+++ b/include/AudioDummy.h
@@ -31,8 +31,9 @@
 #include "Mixer.h"
 
 
-class AudioDummy : public AudioDevice, public QThread
+class AudioDummy : public QThread, public AudioDevice
 {
+	Q_OBJECT
 public:
 	AudioDummy( bool & _success_ful, Mixer* mixer ) :
 		AudioDevice( DEFAULT_CHANNELS, mixer )
--- a/include/AudioOss.h
+++ b/include/AudioOss.h
@@ -39,8 +39,9 @@ class LcdSpinBox;
 class QLineEdit;
 
 
-class AudioOss : public AudioDevice, public QThread
+class AudioOss : public QThread, public AudioDevice
 {
+	Q_OBJECT
 public:
 	AudioOss( bool & _success_ful, Mixer* mixer );
 	virtual ~AudioOss();
--- a/include/AudioPulseAudio.h
+++ b/include/AudioPulseAudio.h
@@ -41,8 +41,9 @@ class LcdSpinBox;
 class QLineEdit;
 
 
-class AudioPulseAudio : public AudioDevice, public QThread
+class AudioPulseAudio : public QThread, public AudioDevice
 {
+	Q_OBJECT
 public:
 	AudioPulseAudio( bool & _success_ful, Mixer* mixer );
 	virtual ~AudioPulseAudio();
--- a/include/AudioSndio.h
+++ b/include/AudioSndio.h
@@ -40,8 +40,9 @@ class LcdSpinBox;
 class QLineEdit;
 
 
-class AudioSndio : public AudioDevice, public QThread
+class AudioSndio : public QThread, public AudioDevice
 {
+	Q_OBJECT
 public:
 	AudioSndio( bool & _success_ful, Mixer * _mixer );
 	virtual ~AudioSndio();
--- a/include/MainWindow.h
+++ b/include/MainWindow.h
@@ -29,7 +29,6 @@
 #include <QtCore/QTimer>
 #include <QtCore/QList>
 #include <QMainWindow>
-#include <QThread>
 
 #include "ConfigManager.h"
 #include "SubWindow.h"
--- a/include/MidiAlsaRaw.h
+++ b/include/MidiAlsaRaw.h
@@ -37,11 +37,11 @@
 
 
 struct pollfd;
-class QLineEdit;
 
 
-class MidiAlsaRaw : public MidiClientRaw , public QThread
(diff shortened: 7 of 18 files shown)


Wallacoloo:
I should note that I didn't name the different audio drivers, like `AudioAlsa` and `AudioOss`. Might push another commit here for that if I get a few minutes this week.

Wallacoloo:
Hmm. So it turns out the other solution is to add `Q_OBJECT` at the top of each class that derives from QThread, and Qt's Meta Object Compiler (`moc`) will stamp out implementations for some virtual functions, including one that returns the class name as a string. Then, QThread automatically calls that virtual function and gets the appropriate name to apply to each thread.

I suspect there might be some reasons to not begin sprinkling `Q_OBJECT` around inside `core` classes though. But on the other hand, `MixerWorkerThread` & such _already_ derive from QThread, which means all those virtual functions are already there - just not overriden.

**Advice would be welcome** - which approach should I use: call `setObjectName("NAME_OF_CLASS")` in the constructor of each type that derives from QThread, or add `Q_OBJECT` to the top of each type that derives from QThread?

lukas-w:
@Wallacoloo I think using `Q_OBJECT` is the best approach. It avoids duplicating the class name and I've yet to see a proof for any considerable overhead, especially for classes that already inherit from `QObject`. AFAIK, All it does in this case is add a `QMetaObject` instance to the class.

Wallacoloo:
Okay, I've implemented this for the Midi and Audio drivers. I've also changed to use Q_OBJECT instead of `setObjectName`, except for `Mixer::fifoWriter` because `Q_OBJECT` declarations don't work for nested classes.

I had to reorder some base classes, because the `Q_OBJECT` macro assumes the primary base class derives from `QObject`.

Would like for somebody to give this a quick once-over before merging to make sure I didn't do something dumb. :-)

Issue #4350, opened 2018-05-10
Piano roll : unexpected snap to grid behavior


rghvdberg:
When changing the grid from 1/16 to 1/12 (for example) and dragging a note , the note does not snap to the grid lines.

Recipe:
1) draw some notes with grid set to 1/16
2) change grid to 1/12
3) try dragging a note to a 1/12 grid line

rghvdberg:
I think the problem lies here 
https://github.com/LMMS/lmms/blob/stable-1.2/src/gui/editors/PianoRoll.cpp#L2412-L2417
```
	// if they're not holding alt, quantize the offset
	if( ! alt )
	{
		off_ticks = floor( off_ticks / quantization() )
						* quantization();
}
```
and later at line 2460
 `int pos_ticks = note->oldPos().getTicks() + off_ticks;`

this moves the notes just a number of quantized ticks and not to the grid as one would expect

rghvdberg:
think I solved it 
```
const NoteVector & notes = m_pattern->notes();

	if (m_action == ActionMoveNote)
	{
        bool first_iter = true;
		for (Note *note : notes)
		{
			if( note->selected() )
			{
				if( shift && ! m_startedWithShift )
				{
					// quick resize, toggled by holding shift after starting a note move, but not before
					int ticks_new = note->oldLength().getTicks() + off_ticks;
					if( ticks_new <= 0 )
					{
						ticks_new = 1;
					}
					note->setLength( MidiTime( ticks_new ) );
					m_lenOfNewNotes = note->length();
				}
				else
				{
					// moving note
           
                    if (first_iter){
                        first_iter = false;
                        Note copy (*note);
                        // quantize first note
                        int pos_ticks = copy.oldPos().getTicks() + off_ticks;
                        pos_ticks = qMax(0, pos_ticks);
                        copy.setPos (MidiTime(pos_ticks));
                        copy.quantizePos(quantization());
                        // new off_ticks based on quantized notes
                        off_ticks += copy.pos().getTicks() - (note->oldPos().getTicks() + off_ticks);
                     }
                    int pos_ticks = note->oldPos().getTicks() + off_ticks;
                    std::cout << "new note pos = " << note->oldPos().getTicks()+off_ticks << std::endl;

					int key_num = note->oldKey() + off_key;

					// ticks can't be negative
					pos_ticks = qMax(0, pos_ticks);
					// upper/lower bound checks on key_num
					key_num = qMax(0, key_num);
[...]

PR #4357, opened 2018-05-14
piano roll editor : better snap to grid #4350

rghvdberg:
When moving (dragging) notes, the notes snap to the grid lines.
Same for changing lenght (without shift or alt), note lenghts snap to the grid.


Spekular:
Keeping offsets from lower quantization levels is useful. I think this may
have been discussed earlier in another issue or pull request.

If you want notes on the grid you can easily have that by placing new ones
or using the quantize button.

rghvdberg:
Any other daw let's you drag notes to the grid lines. Current behaviour is imho very unexpected.

SecondFlight:
Are you sure about that @rghvdberg?



@Umcaruje can confirm this in Bitwig as well.

Here's my thinking behind this:
If you make some riff or grace note that uses faster notes than the normal grid division and you want to move it around, you have to go back to the timescale you were using previously to do that. Now, I haven't compiled your branch yet and you didn't put much detail in the PR, but based on your code it looks like it only snaps the first note in a group of notes. Correct me if I'm wrong. That could frustrate the end user in some cases - for example, what if I want to move this over by a quarter note:

image

Currently, I just move it over four 16ths, or switch the snap to quarter if I'm so inclined. With the proposed changes, the only viable option would be to use alt. I used alt to make the lead-in note, so I can't just switch it to a smaller timebase.

Of course, all of this would be nullified if the PR does something considerably different from what I'm saying here, but I can attest that LMMS currently acts very similarly to other DAWs in this regard.

rghvdberg:
No it snaps all notes to the nearest grid.

On the initial drag it calcs the distance from a grid line and overrides
the off_ticks.

I should actually find out which note is grabbed and calc the offset needed
for that note.

Note to self: don't buy bitwig 😋

On 15 May 2018 4:25 a.m., "Joshua Wade" <notifications@github.com> wrote:

Are you sure about that @rghvdberg <https://github.com/rghvdberg>?

<https://camo.githubusercontent.com/951f0ae19162a8482751c4f8fb6275a499c1c199/68747470733a2f2f692e696d6775722e636f6d2f723062666a6e372e676966>

@Umcaruje <https://github.com/Umcaruje> can confirm this in Bitwig as well.

Here's my thinking behind this:
If you make some riff or grace note that uses faster notes than the normal
grid division and you want to move it around, you have to go back to the
timescale you were using previously to do that. Now, I haven't compiled
your branch yet and you didn't put much detail in the PR, but based on your
code it looks like it only snaps the first note in a group of notes.
Correct me if I'm wrong. That could frustrate the end user in some cases -
for example, what if I want to move this over by a quarter note:

[image: image]
<https://user-images.githubusercontent.com/6700184/40033142-0f68b08a-57c5-11e8-8fa8-4be2223b8e1e.png>

Currently, I just move it over four 16ths, or switch the snap to quarter if
I'm so inclined. With the proposed changes, the only viable option would be
to use alt. I used alt to make the lead-in note, so I can't just switch it
to a smaller timebase.

Of course, all of this would be nullified if the PR does something
considerably different from what I'm saying here, but I can attest that
LMMS currently acts very similarly to other DAWs in this regard.

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/LMMS/lmms/pull/4357#issuecomment-389022401>, or mute
the thread

SecondFlight:
That's actually FL Studio. As I understand it, that behavior is standard across DAWs.

I'll grab your code later when I get the time and make sure I'm understanding the change.

rghvdberg:
Lol, I was trying to install FL studio to look how that one works. You just
saved me some time.

Spekular:
Here's the PR I was thinking of, with discussion on why it's valuable to keep fine adjustments when moving at a higher Q: https://github.com/LMMS/lmms/pull/4043

Here's an example in the piano roll of some recorded chords.

First, the chords are on the offbeat:
image

But perhaps I realize that I want them on the 1/2/3/4, or the recording was delayed by 1/8th for some reason. With the current behavior, I simply move them over two 16ths (though not visible in the screenshot, my Q is at 1/16th here):
image

The chords keep their inconsistencies and slight delays. If your suggestion were implemented, they would now all hit exactly at the right time, with no delay between notes in a chord:
image

Note that to get the notes as they are in the third screenshot, I simply hit the quantize button. So it's currently easy to achieve what I want (shifting the notes by a certain offset) and what you want (snapping the notes to certain positions). If moving notes also snapped them, it would be far more difficult to do what I want.

I do not think this PR should be merged. I've explained why the current behavior is desired and SecondFlight has shown that it is common in other DAWs. That said, I do want to apologize for not bringing this up earlier, before you opened the PR.

rghvdberg:
I guess I'm used to this behaviour 

Ardour
ardour grid

Qtractor
qtractor grid

Drumroll .... LMMS (my PR)
new grid lmms

rghvdberg:
improved snap
Improved it a bit.
The snap now depends on which note you started dragging.

SecondFlight:
What if we bind that to a modifier key? Alt is taken, and I think shift is too, but we could put that on ctrl + shift or ctrl + alt or something like that.

I'm still very much against this as the default, though I will say that your most recent modification makes it considerably more usable in the general case IMO.

Another option would be to make a toggle in the settings, something like "quantize dragged notes" (there's probably a better way to word that)

SecondFlight:
Also, and I forgot to mention this before, you can quickly get a single note or the first of a group of notes to snap to the grid by dragging it all the way to the left and then back again. This might not be practical for really long patterns, but it's an intuitive way to go about it based on the interface. This doesn't beat a shortcut key or extra setting though IMO, especially with the new change you added.

Spekular:
I'm fine with this but not as the default. A toggle in the piano roll menu
or somewhere in the LMMS settings would be the best option IMO.

rghvdberg:
I've given it some thought and I think a modifier key and maybe a toggle in settings would be the best option.
I was trying to add another button in the piano roll editor but then you'd have 2 'places' for moving notes and that can only lead to confusion.

SecondFlight:
I agree, the UX on that is not ideal.

I don't see any reason to have both a modifier key and a settings toggle. I think that would make it harder to find. Someone could turn on the setting and reasonably be confused as to why nothing has changed, unless the setting is worded right, but even then I don't see a good reason to make the user go in there if it's already behind a modifier key.

rghvdberg:
Agreed 👍

PhysSong:
@rghvdberg Please use tabs for indentation. It's our coding convention.

rghvdberg:
Yup I noticed. I'll fix it

rghvdberg:
I've added a combo box.
defaults to the old behaviour ofc

nu_snap

Spekular:
This looks very nice!

As a nitpick, I think "nudge" and "snap" should be capitalized
("Nudge"/"Snap"). Our dropdowns tend to be sentence case, from the ones
I've checked.

SecondFlight:
At some point in the future it may be useful to consolidate some of the odds and ends sorts of options somehow, but at this point I don't think the UI is super cluttered. This is a fantastic solution as far as I'm concerned!

Does it remember your choice between sessions? If that's not trivial then don't worry about it.

rghvdberg:
Never crossed my mind to keep it between sessions. I'll have a look how to implement that.

rghvdberg:
Do the other options (quantize, note lenght etc) persist between settings?
If not, it would be odd to have this setting persist and not the others.

SecondFlight:
Yeah, that's fair, they don't persist.

Does it default to the old behavior?

rghvdberg:
Yup. Nudge == old behaviour and the default.

justnope:
Are we talking about persisting in global config of lmms or in the individual project file? Or both?

My two cents is that this is definitely a worth a new issue.

PhysSong:
@Spekular As the author of #4973, do you think this PR is still useful?

Spekular:
@PhysSong yes, in fact I think it's even more warranted now so we have consistent functionality between the editors.

PhysSong:
@rghvdberg Do you want to continue working on this, or let some of us take this over? @Spekular Are you interested in this?

rghvdberg:
Would be fun to work on it again. Gimme a bit of time to set things up again.

PhysSong:
Continued in #5848.

Issue #3689, opened 2017-07-08
Flanger - Render from gui not set up properly


zonkmachine:
Something isn't properly set up/reset when rendering from the gui. I've seen this with the Flanger effect only so far. I'm not sure if this is an issue with Flanger specifically but ReverbSC and Delay renders identically from the gui.

Left side rendered from the command line and to the right, from the gui.
![flangercl](https://user-images.githubusercontent.com/6368949/27983026-61cb9632-63b3-11e7-82b4-d20247f71ae1.png)

Edit: [flange.mmp.zip](https://github.com/LMMS/lmms/files/1132851/flange.mmp.zip)

PhysSong:
@zonkmachine Could you upload the project file? I'll do some tests with it.

zonkmachine:
^^--- Fixed!

zonkmachine:
Oops, wrong project. Fixed!

zonkmachine:
Sorry, earlier confusing message removed. The actual issue isn't fixed... the issue _report_ is.

PhysSong:
Our native flanger has built-in LFO. It isn't reset automatically when the export starts.
I haven't tested yet, but I guess that's the reason.

PhysSong:
We may add something like this in the LFO controller class, in order to reset a flanger's LFO:
https://github.com/LMMS/lmms/blob/e5db27542b3b05fe0b0cbdfe277f1ad187cf48d0/src/core/LfoController.cpp#L61-L62

PhysSong:
This bug seems minor, but flanger is the only plugin with this behavior AFAIK. So I don't know which branch is proper as the target.
@zonkmachine How do you think about this?

zonkmachine:
Bump flanger overlord @curlymorphic

PR #4360, opened 2018-05-16
Ensure flanger LFO starts at the same position on render

SecondFlight:
Fixes https://github.com/LMMS/lmms/issues/3689

Here are two GUI renders:

image

The two renders weren't identical, and that doesn't sit well with me. Regardless, the LFO works exactly the same on both renders.

### DIFF
--- a/plugins/Flanger/FlangerControls.cpp
+++ b/plugins/Flanger/FlangerControls.cpp
@@ -43,6 +43,7 @@ FlangerControls::FlangerControls( FlangerEffect *effect ) :
 
 {
 	connect( Engine::mixer(), SIGNAL( sampleRateChanged() ), this, SLOT( changedSampleRate() ) );
+	connect( Engine::getSong(), SIGNAL( playbackStateChanged() ), this, SLOT( changedPlaybackState() ) );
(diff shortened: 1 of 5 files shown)


PhysSong:
@SecondFlight Good work! I wonder if this could go into `stable-1.2`.

SecondFlight:
I imagine it could. The merge doesn't work by default of course, but I'm sure I can find a way to do get the commits on 1.2.

If they go on 1.2, is there anything I would need to do to get them on master?

PhysSong:
> If they go on 1.2, is there anything I would need to do to get them on master?

We merge `stable-1.2` into `master` at appropriate time, it's a common workflow.

curlymorphic:
@SecondFlight nice work,

SecondFlight:
I'm having trouble building 1.2, based on what I'm seeing it looks like I'd have to build Qt 4 from scratch, and I don't really want to do that just yet. I'll still implement the changes since the code should all be the same, but I can't actually test the functionality.

PhysSong:
Since #4363 is opened, I'll close this one.

PR #4363, opened 2018-05-17
Restart flanger LFO on Song::playbackStateChanged signal

SecondFlight:
https://github.com/LMMS/lmms/pull/4360/, but on stable-1.2.

I'm unable to test this, could someone who has 1.2 builds set up test this and make sure it works as advertised?


SecondFlight:
Thanks for testing, @zonkmachine!

I figured out the build process for 1.2-stable and tested this, and it appears to work just fine on my end. The only reason I could see why the 10th render would fail is if the event I'm listening to doesn't fire every 10th render for some reason. But based on your edit I think this is good to merge. I'll do that tomorrow if nobody objects before then.

PR #4364, opened 2018-05-17
Remove FluidSynth requirement for background pic

Sawuare:
Because it is irrelevant.

As @PhysSong mentioned on Discord, the requirement was introduced in https://github.com/LMMS/lmms/commit/ae03f9f3b16aa8fe799caf8f38df5282583633b2.


zonkmachine:
Nice catch!

Issue #4325, opened 2018-04-26
Limiter No.6 causes deadlock when changing latency mode while processing


PhysSong:
To reproduce:
1. Load Limiter No.6 on master channel
2. Press "Fix. latency" while playing audio

Here's the relevant part of backtrace for `RemoteVstPlugin.exe.so`:
```gdb
Thread 3 (Thread 0x2393b40 (LWP 14783)):
#0  0xf7fd5059 in __kernel_vsyscall ()
#1  0xf7dda5c2 in __lll_lock_wait () from /lib/i386-linux-gnu/libpthread.so.0
#2  0xf7dd2b7c in pthread_mutex_lock ()
   from /lib/i386-linux-gnu/libpthread.so.0
#3  0x7fb0dffd in RemoteVstPlugin::lockShm() ()
   from /home/hjsong/.wine/dosdevices/z:/usr/local/lib/lmms/RemoteVstPlugin.exe.so
#4  0x7fb0ba8c in RemoteVstPlugin::updateInOutCount() ()
   from /home/hjsong/.wine/dosdevices/z:/usr/local/lib/lmms/RemoteVstPlugin.exe.so
#5  0x7fb0c1b0 in RemoteVstPlugin::hostCallback(AEffect*, int, int, int, void*, float) ()
   from /home/hjsong/.wine/dosdevices/z:/usr/local/lib/lmms/RemoteVstPlugin.exe.so
#6  0x1001b6ab in ?? ()
#7  0x10017ee3 in ?? ()
#8  0x1001a63d in ?? ()
#9  0x7fb07946 in RemotePluginClient::doProcessing() ()
```

The problem is the callback tries to lock `m_shmLock` but it's already locked by `doProcessing`. I don't know which solution will be the best, but what I think is: make `m_shmLock` to a recursive mutex.

@DomClark Any ideas/suggestions?

DomClark:
I don't think making `m_shmLock` recursive is a good solution in general - the purpose of the mutex is to prevent `m_inputs`, `m_outputs` and the shared memory from being deleted during a call to `m_plugin->processReplacing`. It might work in this particular case, but for other plugins it may well result in use-after-free crashes within the plugin.
[...]

Issue #3918, opened 2017-10-27
Issue with Vestige + Helm


wtpisaac:
I've noticed that when trying to use Helm as a VST under Windows, running on LMMS v1.2.0rc4, attempting to open a menu in Helm does not work.

**Expected Behavior:** Clicking on Distortion menu in the Helm synth opens the Distortion menu. Right clicking opens the right click menu.

**Actual behavior:** These menus open, but then close immediately, too quickly to interact with anything in it. _Very rarely_, it will work, which I notice causes Helm's animations to stop.

I believe this is an issue with LMMS (specifically vestige?) as the standalone Helm application does not have this issue on Windows, nor does this issue occur in other DAWs.

tresf:
Tagging @DomClark -- our resident expert with VSTs. :)

DomClark:
Successfully reproduced. I've found one other host where this problem occurs: Hermann Seib's [VSTHost](http://www.hermannseib.com/english/vsthost.htm) when the plugin is running with a different number of bits to that of the host, in which case it is loaded in a separate bridge process, like LMMS does. Presumably having the parent window in a different process messes something up somewhere. I'm quite busy with university work currently, but I'll take a look when I have time. In the meantime, the value can be changed by hovering over the menu and scrolling.

DomClark:
Here is a pair of Helm builds with `Process::isForegroundProcess` patched to always return `true`, which should behave slightly better. Build is of master at time of writing, at commit [81a65d0](https://github.com/mtytel/helm/tree/81a65d0e68980f59450c1e46d008a54784ca6674). Standard "this may break stuff" disclaimers apply.

[helmpatched.zip](https://github.com/LMMS/lmms/files/1424561/helmpatched.zip)

zonkmachine:
> somebody should probably just file an issue with JUCE.

Maybe this is a job for the Helm devs?
JUCE issue tracker here: https://github.com/WeAreROLI/JUCE

tresf:
@DomClark excellent investigation.  Can you please open a bug at JUCE describing the issue?

PhysSong:
Can be reproduced on Windows, but not on Linux. FYI one can work around this issue once #3786 is merged, by using separate VST window mode. Note that this isn't a complete fix...

zonkmachine:
@DomClark Can you please open an issue with JUCE?
https://github.com/WeAreROLI/JUCE
I wouldn't know how to word it.

Also, bumping helm @mtytel

[...]

PR #4366, opened 2018-05-18
fix #4306 VST UI Lockup

### LINKED ISSUE #4306: VST Issue: TX16Wx causes complete UI lockup when attempting to launch file picker
This is an issue with [TX16Wx](https://www.tx16wx.com/). 

Steps to reproduce:
image
image

At this point it will lock up. The CPU meter is still active, so it appears to not be a crash or a freeze.

I did a bit of digging, this is what is supposed to come up when you use that option:
image

justnope:
Bug fix for vst UI lockup when a file dialog is opened.
If it's too risky/late to include for stable-1.2, I'll change it to master instead.

### DIFF
--- a/plugins/vst_base/RemoteVstPlugin.cpp
+++ b/plugins/vst_base/RemoteVstPlugin.cpp
@@ -1961,6 +1961,10 @@ LRESULT CALLBACK RemoteVstPlugin::messageWndProc( HWND hwnd, UINT uMsg,
 
 int main( int _argc, char * * _argv )
 {
+#ifdef LMMS_BUILD_WIN32
+	CoInitialize(nullptr);
+#endif
+
 #ifdef SYNC_WITH_SHM_FIFO
 	if( _argc < 4 )
 #else
@@ -2086,6 +2090,9 @@ int main( int _argc, char * * _argv )
 
 	delete __plugin;
 
+#ifdef LMMS_BUILD_WIN32
+	CoUninitialize();
+#endif
 
 #ifdef LMMS_BUILD_WIN32
 #ifndef __WINPTHREADS_VERSION


SecondFlight:
I can test this on a few VSTs if someone is willing to make a quick Windows build. I don't have Windows builds set up at the moment.

justnope:
You can download it at https://gist.github.com/justnope/6c1ea70edaf83cb1acec4a1dc60dc92b

SecondFlight:
Thanks! I'll let you know when I get the chance to test this.

PhysSong:
Tested, works!
This can be applied for Linux build as well. To do that, `-lole32` should be added into `WINE_CXX_ARGS`.
> If it's too risky/late to include for stable-1.2,

I think it's safe if (un)initializing COM library more than once.

justnope:
@PhysSong feel free to add to this PR to enable it for the linux build. I think it should be possible because allow edits from maintainers is on.

zonkmachine:
Merge?

SecondFlight:
I don't see why not. I haven't gotten the chance to test this yet but I can't imagine it breaking anything too badly.

PhysSong:
I'll merge this if I can't find any issues after reading related MSDN articles.

PhysSong:
I just found that MSDN recommends to use `CoInitializeEx` instead of `CoInitialize`. `CoInitializeEx(nullptr, COINIT_APARTMENTTHREADED);` will be behave the same as current one.
I think not calling `CoInitialize`/`CoInitializeEx` where needed is a bug of VST plugin, but handling problematic plugin looks good for me.
FYI, [`OleInitialize`](https://msdn.microsoft.com/ko-kr/library/windows/desktop/ms690134) initializes more functionalities than `CoInitialize` does.

DomClark:
I think we should use `OleInitialize`: there is only additional overhead when loading/unloading plugins, and given that we've found a plugin that doesn't initialise COM when it should, there may well be plugins that don't initialise OLE when they should.

I'm going to put in another PR with VST fixes soon, addressing things like #3918 and #4325; since @justnope has deleted his account, shall I make the changes there instead?

PhysSong:
> since @justnope has deleted his account, shall I make the changes there instead?

That looks fine, but you can also open a new pull request with commits in this PR cherry-picked.

PR #4371, opened 2018-05-22
RemoteVST: process all remaining messages after the process has quit

justnope:
If the process exits, keep processing the remaining messages. Some might be debug messages or a msg that the vst loaded was 32 bits.

### DIFF
--- a/src/core/RemotePlugin.cpp
+++ b/src/core/RemotePlugin.cpp
@@ -54,7 +54,7 @@ ProcessWatcher::ProcessWatcher( RemotePlugin * _p ) :
 
 void ProcessWatcher::run()
 {
-	while( !m_quit && m_plugin->isRunning() )
+	while( !m_quit && (m_plugin->messagesLeft() || m_plugin->isRunning()) )
 	{
 		msleep( 200 );
 	}


src/core/RemotePlugin.cpp
```diff
@@ -54,7 +54,7 @@ ProcessWatcher::ProcessWatcher( RemotePlugin * _p ) :
 
 void ProcessWatcher::run()
 {
-	while( !m_quit && m_plugin->isRunning() )
+	while( !m_quit && (m_plugin->messagesLeft() || m_plugin->isRunning()) )
```
Wallacoloo: It seems to me there's a possibility for a race condition:
1. `m_plugin->messagesLeft()` returns `false`.
2. VST posts new message and then exits
3. `m_plugin->isRunning()` returns `false`.
4. Loop aborts even though there are pending messages.

I think the conditional should just be rearranged so that `m_plugin->messagesLeft()` is only checked _after_ `m_plugin->isRunning()`:
```
while( !m_quit && (m_plugin->isRunning() || m_plugin->messagesLeft()) )
```

I'm not very familiar with the area of the code though, so don't hesitate to point out why I might be wrong about this.

justnope: heh :) good catch, thanks. (I wanted to write something about why you were wrong but then realized you were right)

Wallacoloo:
Thanks for the patch! I went ahead and made the adjustment mentioned above, and committed that to master: https://github.com/LMMS/lmms/commit/4fd8ecd7e4084aedbd39900ffe5cfd107ceba1a0

Credited you by your original username, but if there was any other metadata in the original commit, it was probably lost since I had to reauthor the commit (couldn't pull from the original repo since it's been deleted).

lukas-w:
@Wallacoloo It looks like this change is ineffective because of a second `invalidate()` call here: https://github.com/LMMS/lmms/blob/392c7534d5f18b9921c7cc8e917282afd1c76dc6/src/core/RemotePlugin.cpp#L454-L460

I ran into `IdVstBasDllFormat` messages not being processed when trying to implement 64bit VSTs on Linux. I feel like that `invalidate()` call can just be removed because `ProcessWatcher` will call `invalidate()` eventually, but I'm not actually sure about the reason `@jasp00` added it there, nor do I understand what the `#ifdef` is for.

Do you have any idea how to approach this?

PR #4372, opened 2018-05-22
extra validation saved audio/midi devices

justnope:
It crashes when it can find an audio interface saved in its settings.

Maybe not a feature useful to the users but when I'm developing I sometimes enable and disable interfaces and previously it crashed.

Also is there a unified/recommended way to output errors?

### DIFF
--- a/src/gui/SetupDialog.cpp
+++ b/src/gui/SetupDialog.cpp
@@ -839,8 +839,18 @@ SetupDialog::SetupDialog( ConfigTabs _tab_to_open ) :
 		ConfigManager::inst()->setValue(
 					"mixer", "audiodev", audioDevName );
 	}
-	m_audioInterfaces->
-		setCurrentIndex( m_audioInterfaces->findText( audioDevName ) );
+
+	int index = m_audioInterfaces->findText(audioDevName);
+	//check index, during debugging (dis/en)abling audio devices this can cause a crash (index -1)
+	//TODO: output error msg? 
+	if (index < 0)
+	{
+		audioDevName = Engine::mixer()->audioDevName();
+		ConfigManager::inst()->setValue(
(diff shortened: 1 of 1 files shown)


src/gui/SetupDialog.cpp
```diff
@@ -839,8 +839,18 @@ SetupDialog::SetupDialog( ConfigTabs _tab_to_open ) :
 		ConfigManager::inst()->setValue(
 					"mixer", "audiodev", audioDevName );
 	}
-	m_audioInterfaces->
-		setCurrentIndex( m_audioInterfaces->findText( audioDevName ) );
+
+	int index = m_audioInterfaces->findText(audioDevName);
+	//check index, during debugging (dis/en)abling audio devices this can cause a crash (index -1)
+	//TODO: output error msg? 
+	if (index < 0)
```
PhysSong: I think you can merge the check into line 836 because it does the same thing inside `if` clauses.

src/gui/SetupDialog.cpp
```diff
@@ -942,8 +952,18 @@ SetupDialog::SetupDialog( ConfigTabs _tab_to_open ) :
 		ConfigManager::inst()->setValue(
 					"mixer", "mididev", midiDevName );
 	}
-	m_midiInterfaces->setCurrentIndex( 
-		m_midiInterfaces->findText( midiDevName ) );
+
+	index = m_midiInterfaces->findText(midiDevName);
+	//not found? init to default
+	//TODO: output error msg?
+	if (index < 0)
```
PhysSong: You can do something similar here, too.

PhysSong:
Since the OP deleted the account, I'll cherry-pick and edit this commit to create a new pull request.

PhysSong:
Superseded by #4407.

PR #4373, opened 2018-05-22
"Delaying" unsetting jack headers path so weakjack can benefit it.

devnexen:


PhysSong:
I think this fix should target to `stable-1.2`.

devnexen:
Unless I m mistaken, there is no weakjack module on stable-1.2 ?

PhysSong:
> there is no weakjack module on stable-1.2?

Yes. I didn't notice that.
We don't build it as a standalone library since edc8f43d5b3b4975f4dc998e4eb8d2c30dd8b7b7, but I think the commit isn't synced into `master` well. I guess I should have removed that on `master`. @tresf right?
Anyway, I think we can just drop the line which unsets `JACK_INCLUDE_DIRS` because it is needed for building LMMS.

devnexen:
Yes removing JACK_INCLUDE_DIRS unsetting would be simpler I did not dare to do so believing there might have been some legit reasons behind it.

CMakeLists.txt
```diff
@@ -378,7 +378,6 @@ IF(WANT_JACK)
 			SET(LMMS_HAVE_WEAKJACK TRUE)
 			SET(WEAKJACK_INCLUDE_DIRS ${CMAKE_SOURCE_DIR}/src/3rdparty/weakjack/weakjack)
 			SET(STATUS_JACK "OK (weak linking enabled)")
-			SET(JACK_INCLUDE_DIRS "")
```
tresf: If I recall correctly, this was a throwback to our unorthodox method of including variables that weren't ever set inside [`src/`](https://github.com/LMMS/lmms/blob/e614711d29d5ee5a958da55f4bb3716d942cd5c6/src/CMakeLists.txt#L54-L55) to avoid cmake warnings or failures.  This is undoubtably the wrong approach so as long as `jack.pc` sets this correctly (and from my experience, `pkg-config` is very consistent at this).  Perhaps it should be set to blank if it WASN'T found.

tresf: @devnexen thanks for the attention to detail here. 👍 Travis-CI was hung, kick-started it and it's good.  Merging.  If this causes issues in the future it's easy enough to put an addition check in place.

PR #4376, opened 2018-05-26
Wacom Tablet support

david-polak:
Hi,

I have just started learning LMMS and found it quite difficult to control the Knobs with my Wacom pen tablet. I use my tablet to alleviate RSI strain of using the mouse, and have gotten quite used to it.

Unfortunately, the current implementation of the Knob is incompatible with a tablet set to Absolute mode (pixels on tablet are mapped roughly 1:1 to pixels on screen). Since in Knob::mouseMoveEvent

`QCursor::setPos( mapToGlobal( m_origMousePos ) );`

the cursor position gets reset, which then triggers a new QMouseEvent because now the cursor position is far away from the current tablet position (which is absolute). First I tried to hack the existing code, but my effort has proven fruitless, due to the nature of the Knob mouse handling implementation (though, I couldn't figure out a better way to do it either)

So I have implemented the TabletEvent interface. Any events that aren't handled by the new code get 
`event->ignore()`ed and then Qt spawns a regular QMouseEvent which then propagates as usual.

So far, Knob is the only GUI element which I found that behaved strangely with my tablet, I will add support to any other elements if I find them.

### DIFF
--- a/include/Knob.h
+++ b/include/Knob.h
@@ -28,6 +28,7 @@
 
 #include <QWidget>
 #include <QtCore/QPoint>
+#include <QTabletEvent>
 
 #include "AutomatableModelView.h"
 #include "templates.h"
@@ -134,6 +135,10 @@ protected:
 	virtual void mouseDoubleClickEvent( QMouseEvent * _me );
 	virtual void paintEvent( QPaintEvent * _me );
 	virtual void wheelEvent( QWheelEvent * _me );
+	virtual void tabletEvent( QTabletEvent * _te );
+	virtual void tabletPressEvent( QTabletEvent * _te );
+	virtual void tabletReleaseEvent( QTabletEvent * event );
+	virtual void tabletMoveEvent( QTabletEvent * _te );
 
 	virtual float getValue( const QPoint & _p );
 
@@ -178,6 +183,9 @@ private:
 	QPoint m_origMousePos;
 	float m_leftOver;
 	bool m_buttonPressed;
+
+	QPointF m_prevTabletPos;
+	bool m_tabletPressed;
 
 	float m_totalAngle;
 	int m_angle;
--- a/src/gui/widgets/Knob.cpp
+++ b/src/gui/widgets/Knob.cpp
@@ -61,6 +61,7 @@ TextFloat * Knob::s_textFloat = NULL;
 	m_volumeKnob( false ), \
 	m_volumeRatio( 100.0, 0.0, 1000000.0 ), \
 	m_buttonPressed( false ), \
+	m_tabletPressed( false ), \
 	m_angle( -10 ), \
 	m_lineWidth( 0 ), \
 	m_textColor( 255, 255, 255 )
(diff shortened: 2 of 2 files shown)


david-polak:
Is the Gcc Xcode build broken? This is not the only PR on which it fails and it has no relation to the changes in this PR.

SecondFlight:
I see you're using tablet-specific Qt events. Does the mouse behavior remain the same?

david-polak:
Yes, QMouseEvent is unaffected in any way. 

When QTabletEvent is not handled and not accepted, Qt generates a QMouseEvent.
When the actual mouse is the originator no QTabletEvent is generated and everything is handled the standard way. I haven't touched the mouse handling logic at all.

david-polak:
I just reread my code, another commit incoming. Need to fix the tabletRelease handler

SecondFlight:
Alright, awesome. I will model the vector graph tension handles (https://github.com/LMMS/lmms/pull/4367) after that. Would you be willing to test it if I ping you about it in maybe a week or so?

SecondFlight:
Clone/build the branch, add the plug-in ”Waveshaper 2” to the master effects chain, and try to drag the little circle in the middle of the line. I've implemented that functionality in a similar way to the knob dragging so I expect it to break in a similar way. I can just ping you when I have a fix implemented.

justnope:
What tablet are you using and which driver version?

I agree that there are problems with tablets but I'm not seeing your issue with my particular tablet. I have other issues with my tablet. (intuos, driver: 6.3.29-6)

Try raising the value of the knob with your stylus, hold it, check the value while holding the stylus motionless, then lower it. What I'm seeing on my end is that for a brief moment the value increases when going down, as if the motion was buffered.

What I mean is that it, from first impressions, looks like it's processing `buffer - <some offset>` instead of the actual latest version `buffer`.

So my questions are: 1) Does my explanation make any sense? 2) Do you also have this "delay" with your tablet?

david-polak:
Interesting, what OS are you using, I am using Arch Linux, will check the tablet version when I get to my pc? 

Do you have the tablet in absolute mode? I could see this issue not being present if the driver actually stops giving updates when you are not moving, but the linux driver I am using does not behave like that.

Have you tried building and running with my branch? Does it affect your issue?

justnope:
@david-polak I'm running windows so the issues will be different. Sorry, I keep forgetting people run on it on linux.

I've not noticed any difference when running your patch, so that's good (I don't think I have your mouse reset problems). So the patch is ok by me if it fixes things on your end. 

This is irrelevant to your bug report, but it would be nice to know if the next issues are also present in linux.

In linux, do you notice any "lag/buffering"? ie. Move your stylus up, hold it then look at the value then move it down. Does the value immediately go down, or goes up for a slight moment and then go down.

I'm wondering if it's a qt or a windows issue.

btw: Would it help to adjust the sentivity of the movement? Meaning the movement up/down is over a larger range so you can be more precise?

PhysSong:
There are some situations where `QCursor::setPos` is called:
- Resizing track height (shift + click and drag, called when resizing starts)
- Drawing graph out of the boundary
- Tweaking knob (the original problem)
- Changing value of LCD spin box (e.g. changing pitch range by click and vertical drag)

PhysSong:
Just a question: can we remove unnecessary and/or problematic calls to `setPos` instead?

PhysSong:
Due to the conflicts of #5360, it makes no sense to merge this as-is.
@JohannesLorenz @tresf Any ideas on this PR? I'm asking you because this looks similar to #5194 in some sense.

JohannesLorenz:
I think even if we rewrite this PR correctly concerning `setPos`, it does not make sense to test it without #5360 . This would mean we first have to

1. Merge #5360 to `stable-1.2`
2. Merge `stable-1.2` to `master`
3. Rebase this PR to `master`

PhysSong:
#5360 fixed the issue, so I'm closing this.

PR #2875, opened 2016-06-26
Split journal ID range between new and loaded elements

jasp00:
Sometimes when loading projects like `Greippi - Krem Kaakkuja (Second Flight Remix).mmpz`, there are ID collisions and messages like `JO-ID %d already in use by %s!`. This may affect automation. To solve this, the range of IDs for new elements is different than for loaded ones. Another improvement is the optimization of assigning a random ID.

PR #4377, opened 2018-05-27
Fix journaling system bugs

### LINKED ISSUE #3359: Copies of  shapes in automation are wrong
OS  : winXP sp3
LMMS 1.1.90-RC2.8

**Issue description:**
Graphs/  shapes in automation are wrong, after they have been copied.

**Example          :**
copyautomarionerror
**Recreate         :**
New project
* drag a panning-dial into automation-track
* Open track
* insert automation like in the picture' top part
* Close editor
* ctrl+drag-copy block to other pos.
The result is like the second part of the picture
A similar issue exists for all 3 shapes, but in linear the bug will only apear if a *declining* pattern is drawn.

Bug is logical (no dial control), not just graphical

### LINKED ISSUE #3781: Control links are lost when duplicating automation clips
When using Copy/Paste or Ctrl+Drag to duplicate  an automation clip, the link to a control is lost.

This makes it very tedious to duplicate automation clips, as one needs to restore the links manually for each clip.

LMMS 1.2 RC3 Linux.

PhysSong:
Previously `JournallingObject::changeID` failed if `_id` is being used. This commit adds handling of ID collision and partially reverts #2875. So it will fix regressions in #2875, too.
Fixes #3781, fixes #3359.

### DIFF
--- a/include/ProjectJournal.h
+++ b/include/ProjectJournal.h
@@ -76,7 +76,7 @@ public:
 		reallocID( _id, NULL );
 	}
 
-	static jo_id_t idToSave( jo_id_t id );
+	void changeID(const jo_id_t oldID, const jo_id_t newID);
 
 	void clearJournal();
 	void stopAllJournalling();
--- a/src/core/AutomatableModel.cpp
+++ b/src/core/AutomatableModel.cpp
@@ -29,7 +29,6 @@
 #include "AutomationPattern.h"
 #include "ControllerConnection.h"
 #include "Mixer.h"
-#include "ProjectJournal.h"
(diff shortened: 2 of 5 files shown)


PhysSong:
I guess this way might bring undesired results though it won't happen in noticeable frequency. I have a better approach and will add changes in this PR.

PhysSong:
The simplest way I can think of is to decouple journaling ID and automation/controller connection. I'll reopen this pull request once it's ready.

PhysSong:
Will be superseded with a new fix soon.

PR #4378, opened 2018-05-27
FX autoquit disabled by default

zonkmachine:
Really short notes doesn't work well with delay based effects with the
default decay settings of the FX autoquit system. disable autoquit by
default.

In response to #4284
Supersedes #4291

### DIFF
--- a/src/gui/SetupDialog.cpp
+++ b/src/gui/SetupDialog.cpp
@@ -135,7 +135,7 @@ SetupDialog::SetupDialog( ConfigTabs _tab_to_open ) :
 	m_displayWaveform(ConfigManager::inst()->value( "ui",
 						   "displaywaveform").toInt() ),
 	m_disableAutoQuit(ConfigManager::inst()->value( "ui",
-						   "disableautoquit").toInt() ),
+						   "disableautoquit", "1" ).toInt() ),
 	m_vstEmbedMethod( ConfigManager::inst()->vstEmbedMethod() )
 {
 	setWindowIcon( embed::getIconPixmap( "setup_general" ) );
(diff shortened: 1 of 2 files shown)


zonkmachine:
Testing to disable the `Decay` knob when auto quit is disabled. Just like we do it for FX plugins that are missing on load.

PhysSong:
> Testing to disable the `Decay` knob when autoquit is disabled

If you do so, I think you should also do that for `GATE` knob too because it only affects auto-quit.

Anyway, I think this is a nice idea. :+1:

zonkmachine:
> If you do so, I think you should also do that for GATE knob too because it only affects auto-quit.

Right, added.

PhysSong:
Is it ready for merge?

zonkmachine:
Yes, but as it doesn't inform the user as to why the controls are frozen it could possibly be a bit annoying. I'm OK merging this however (and keep  #4284 open).

PhysSong:
> as it doesn't inform the user as to why the controls are frozen it could possibly be a bit annoying.

If you think so, you may drop the part before merging.

zonkmachine:
@PhysSong Sorry for the delay. I don't have time to deal with this properly right now. Last time I looked at this I felt that it should be dealt with a proper 'minimum decay' fix instead instead of this more temporary fix.

zonkmachine:
What's needed here is a `:disabled` setting for knobs in our stylesheet (and other components that lack this). The tension knob in the Automation Editor would also benefit from this as its disabled when it doesn't have any effect. We do this already for menu items, look at `Edit->Redo` when there are no actions to undo. I won't have time to look into theming unfortunately. @LMMS/artwork-team 

I also looked into temporarily setting the knobs to 0.0f when disabled and even though this looked good it became much too complex to pull off properly.

Conclusion. I think merge this in its current state and someone more into theming can look into greying out the disabled knobs later.

zonkmachine:
> What's needed here is a `:disabled` setting for knobs in our stylesheet

Opened an issue for this here: https://github.com/LMMS/lmms/issues/4749

PR #4382, opened 2018-05-29
Initialize peak controller last sample with base value

DouglasDGI:
This change makes it so that when an LMMS project is loaded, each knob connected to a Peak Controller will be set to the Peak Controller's Base value, rather than its minimum possible value.

### DIFF
--- a/plugins/peak_controller_effect/peak_controller_effect.cpp
+++ b/plugins/peak_controller_effect/peak_controller_effect.cpp
@@ -26,12 +26,11 @@
 
 #include "Controller.h"
 #include "Song.h"
-#include "PresetPreviewPlayHandle.h"
 #include "PeakController.h"
 #include "peak_controller_effect.h"
 #include "lmms_math.h"
 
-#include "embed.cpp"
+#include "embed.h"
 
 extern "C"
 {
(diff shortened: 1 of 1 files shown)


Wallacoloo:
@DouglasDGI Thanks for the bugfix!

I'm going to go ahead and cherry-pick this onto master as well.

DomClark:
This doesn't work because the state of the controls has not been restored at the point where `m_lastSample` is initialised. The effect of this PR is that instead of initialising the value to 0, it is initialised to 0.5, the default value of the base model. This isn't a bad change, since the variable now has the correct initial value upon creation of a new peak controller, but it isn't the desired change either.

A working approach would be either to set `m_lastSample` from `PeakControllerEffectControls::loadSettings` after the settings for `m_baseModel` are loaded:
```cpp
void PeakControllerEffectControls::loadSettings( const QDomElement & _this )
{
	m_baseModel.loadSettings( _this, "base" );
	m_effect->m_lastSample = m_baseModel.value();
...
```
or to override `loadSettings` in `PeakControllerEffect`, something along the lines of
```cpp
void PeakControllerEffect::loadSettings( const QDomElement & _this )
{
	Effect::loadSettings( _this );
	m_lastSample = m_peakControls.m_baseModel.value();
}
```

PhysSong:
@DomClark I believe newly created peak controllers still need the old change.

PR #4391, opened 2018-06-03
Fix crash when loading project with missing peak controller effect

### LINKED ISSUE #4060: Project crash on loading, controller loading error
Some projects saved with 1.2 beta versions will always fail to crash.

Affected projects trigger this assertion:
```
void Song::restoreControllerStates( const QDomElement & element )
{
    QDomNode node = element.firstChild();
    while( !node.isNull() && !isCancelled() )
    {
        Controller * c = Controller::create( node.toElement(), this );
        Q_ASSERT( c != NULL );

        addController( c );

        node = node.nextSibling();
    }
}
```

backtrace
```
QtXmlWrapper::loadXMLfile(): empty data
ASSERT: "c != NULL" in file /home/ubuntu/Desktop/lmms/src/core/Song.cpp, line 1292

Program received signal SIGABRT, Aborted.
0x00007ffff3fb6c37 in __GI_raise (sig=sig@entry=6)
    at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
56    ../nptl/sysdeps/unix/sysv/linux/raise.c: No such file or directory.
```

The bug is OS-independent.
These are two affected projects:
[crashing project.zip](https://github.com/LMMS/lmms/files/1573310/crashing.project.zip)

By unpacking "pentatonica emitonica-01.mmpz" using [this tool](https://gist.github.com/tresf/e5d4f1edb5a93b36f779359eb44b008a) you can see that **a peakcontroller is present but no peakcontrollereffect is found**, as opposed to a good project that will show both the controller and the fx.

Here are the xml of one of the affected project, along with a sample .mmp file with peak controller present
[goodvsbad.zip](https://github.com/LMMS/lmms/files/1573325/goodvsbad.zip)

PhysSong:
Fixes #4060.

### DIFF
--- a/src/core/AutomatableModel.cpp
+++ b/src/core/AutomatableModel.cpp
@@ -110,7 +110,8 @@ void AutomatableModel::saveSettings( QDomDocument& doc, QDomElement& element, co
 		element.setAttribute( name, m_value );
 	}
 
-	if( m_controllerConnection )
+	if( m_controllerConnection && m_controllerConnection->getController()->type()
+				!= Controller::DummyController )
 	{
 		QDomElement controllerElement;
 
--- a/src/core/ControllerConnection.cpp
+++ b/src/core/ControllerConnection.cpp
@@ -162,6 +162,11 @@ void ControllerConnection::finalizeConnections()
 			c->setController( Engine::getSong()->
 					controllers().at( c->m_controllerId ) );
 		}
+		else if (c->getController()->type() == Controller::DummyController)
+		{
+			delete c;
+			--i;
+		}
 	}
 }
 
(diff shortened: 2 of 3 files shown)


PhysSong:
Loading such project will create dummy connections where the broken connection was. Would it be better to remove them or leave them? Leaving will let user easily notice there were connections, but it leads to weird connections when re-saved.
I will add a check for dummy connection in saving code, and removal of them if needed.

zonkmachine:
I'm only able to reproduce #4060 with the AppImages but the test build of this PR worked at a more casual test. I have problems building right now but may test this more later. I don't know why I see #4060 only with the AppImages.

PhysSong:
Issues with dummy connections:
- Lock model value
- Cause wrong connections to be added when saving/reopening

So I think broken connections should be removed when loading project.

zonkmachine:
I've tested this some more and the demo files from #4060 opens up with no problems. I'm also fine with removing broken connections. It's the manly way to go about the issue. Just nuke it. I think you should go ahead and merge this if/when you're happy with it. I bumped #devtalk for a review and there's been no objections.

PR #4396, opened 2018-06-03
Fix one of Equlizer LP48 filter taking HP value

### LINKED ISSUE #4394: Equalizer: Lowpass filter broken at 48dB
-🅱oublas- reported on Discord that the Lowpassfilter at 48dB is broken. I made some pictures to illustrate that:

Edit: I can reproduce this only in the latest master branch. In stable-1.2 it seems to be fine.
----------------------------------------------------------------------------------------------------------------------------

This is the EQ of an earlier version:
eqshould

This is the current EQ behavior:
eqbroken

curlymorphic:
The Equalizer plugin had a bug,where the LP48 filter was having one of its poles incorrectly set by the HP filter model. This resulted, in an unusable LP48 filter.

This fixes the above.

fixes #4394


PhysSong:
Merge?

PR #4397, opened 2018-06-03
Improved Oscillators

curlymorphic:
LMMS used the same oscillators in many places for audio generation. This pull request provides a band limited option, so no aliasing, and a cleaner signal. The Implementation has been via TripleOscillator, will the addition of a new button per oscillator allowing switching between the classic sounds, and the new. Currently the sin, triangle, saw, square, moog saw and the exponential waveforms have a band limited version. The user wave shape is still WIP

The clearness of the waveforms can be clearly heard, and become more apparent when using FM or phase mode, backward compatibility is retained.
 

image

I have a wiki page with explanations images and audio files
https://github.com/curlymorphic/lmms/wiki/LMMS-Oscillator-Class


curlymorphic:
I am aware it appears to fail to link the fftw3 library , and am investigating.

karmux:
screenshot_20180608_232606

It plays silence using these settings. There are many more settings that play silence when wt buttons are active. When I deactivate last wt button it starts to make sound again.

karmux:
One more issue.

When default TripleOscillator/SEGuitar.xpf is played at the same time with simple beat made of kick01.ogg then LMMS always crashes very quickly. I can't crash LMMS when playing SEGuitar.xpf in solo.

SecondFlight:
Thanks for testing this, @karmux!

curlymorphic:
@karmux 
Thank you for your testing, you reveled a bug where the moogsaw and the exp waveforms were not generated. I have updated the code.

Wallacoloo:
Added the `in progress` label since the end of the description ("This code is not being presented as finished, but I would like to present the concept for comments") makes it sound like it's not quite finalized (and there are still some TODOs sprinkled in the code). But please feel free to remove that label once you're happy with it & ready for the code to be reviewed/merged :+1:

qnebra:
This with @SecondFlight 3xOSC randomization patches will be great.

curlymorphic:
I feel this is ready for both testing and code review. 
I shall of course squash once all amendments are finalized, before merging

The adding of the WT oscillators to additional internal plugins would be better handled in additional pull requests, once any issues are ironed out.

src/core/Oscillator.cpp
```diff
@@ -88,6 +118,206 @@ void Oscillator::update( sampleFrame * _ab, const fpp_t _frames,
...
+	for (int i = 0; i < WAVETABLE_LENGTH; i++)
+	{
+		m_generatedWaveTable[i] = sinf(((float)i / (float)WAVETABLE_LENGTH) * F_2PI);
+	}
+}
+
+void Oscillator::generateSawWaveTable(int bands)
+{
+	float max = 0;
+	m_generatedWaveTable = new sample_t[WAVETABLE_LENGTH];
```
Wallacoloo: These `generate<X>WaveTable()` functions all leak the allocation associated with `m_generatedWaveTable` (i.e. there's no `delete` matching this `new` statement). I would advise just returning a `std::array<sample_t, WAVETABLE_LENGTH>` type rather than returning the data through a heap-allocated member variable. i.e.
```
std::array<sample_t, WAVETABLE_LENGTH> Oscillator::generateSawWaveTable(int bands)
{
    std::array<sample_t, WAVETABLE_LENGTH> wave;
    // populate wave[i] for each index i.
    return wave;
}
```

src/core/Oscillator.cpp
```diff
@@ -88,6 +118,206 @@ void Oscillator::update( sampleFrame * _ab, const fpp_t _frames,
...
+	for (int i = 0; i < WAVETABLE_LENGTH; i++)
+	{
+		m_generatedWaveTable[i] = 0.0;
+		for (float n = 1; n <= bands; n++)
+		{
+			m_generatedWaveTable[i] +=powf((float)-1.0, (float)(n + 1)) *
```
Wallacoloo: You can write `(float)-1.0` as `-1.0f`, but instead of `powf((float)-1.0, (float)(n + 1))`, it might be more efficient to write this as `(n%2 ? 1.0f : -1.0f)`, i.e.
```
m_generatedWaveTable[i] += (n%2 ? 1.0f : -1.0f) / n * sinf(F_2PI * (float)(i * n) / (float)WAVETABLE_LENGTH);
```

Wallacoloo: Also, I would provide a comment in each of these functions along the lines of "Compute a band-limited saw wave using its fourier series representation. See https://en.wikipedia.org/wiki/Sawtooth_wave" (or link to some other reference that shows the fourier decomposition of the wave. It seems the Wikipedia page on sawtooth waves conveniently puts that info right at the top).

Otherwise, this just looks like black magic :P

include/Oscillator.h
```diff
@@ -153,6 +165,21 @@ class EXPORT Oscillator
...
+		f_cnt_t f1 = static_cast<f_cnt_t>(frame) % WAVETABLE_LENGTH;
+		if (f1 < 0)
+		{
+			f1 += WAVETABLE_LENGTH;
+		}
+		f_cnt_t f2 = f1 < WAVETABLE_LENGTH - 1 ?
+					f1 + 1 :
+					0;
+		return linearInterpolate(s_waveTables[shape][waveTableBandFromFreq(m_freq)][f1],
+				s_waveTables[shape][waveTableBandFromFreq(m_freq)][f2], fraction(frame));
```
Wallacoloo: Right now, the compiler can't always see into `waveTableBandFromFreq(m_freq)` since that's defined in the source file. So I'll bet it's going to actually call that function twice, since it won't be able to verify it has no side-effects. This _might_ change if you move that function into the header, but I'd feel more comfortable if you assigned that result to a variable, e.g. something like

```
int band = waveTableBandFromFreq(m_freq);
return linearInterpolate(s_waveTables[shape][band][f1], s_waveTables[shape][band][f2], fraction(frame));
```

src/core/Oscillator.cpp
```diff
@@ -28,9 +29,33 @@
...
+void Oscillator::waveTableInit()
+{
+	if (!s_waveTableBandFreqs)
+	{
+		createFFtPlans();
+		s_waveTableBandFreqs = new float[127];
+		s_waveTablesPerWaveformCount = 0;
+		for (int i = 1; i < 127; i+=4)
+		{
+			s_waveTableBandFreqs[s_waveTablesPerWaveformCount] = 440.0 * powf(2.0, (i - 69.0) / 12.0);
```
Wallacoloo: Similar to `waveTableBandFromFreq`, I think it would be simpler to replace `s_waveTableBandFreqs` with a function that computes the frequency on-demand (actually, I'll bet there's a function somewhere in the codebase that already does this). Fewer allocations and state to keep track of.

include/Oscillator.h
```diff
@@ -165,6 +192,34 @@ class EXPORT Oscillator
...
+	void generateSawWaveTable(int bands);
+	void generateTriangleWaveTable(int bands);
+	void generateSquareWaveTable(int bands);
+	void generateMoogSawWaveTable(int bands);
+	void generateExpWaveTable(int bands);
+	void generateFromFFT(int bands, float threshold);
+	int  waveTableBandFromFreq(float freq);
+	void generateWaveTables();
+	void allocWaveTables();
+	void createFFtPlans();
```
PhysSong: Somewhat odd choice of cases, I guess either FFT or Fft would be better.

src/core/Oscillator.cpp
PhysSong: A few more alternatives:
- Pass the address to write results
- Use arrays and/or some smart pointers for `s_waveTables`

PhysSong: Also, I think you may reduce duplications by using passing waveform type/pointer to a function that returns coefficient.

src/core/Oscillator.cpp
```diff
@@ -88,6 +118,206 @@ void Oscillator::update( sampleFrame * _ab, const fpp_t _frames,
...
+	s_waveTables = new sample_t**[WaveShapes::NumWaveShapes];
+	memset(s_waveTables, 0, sizeof(sample_t) * WaveShapes::NumWaveShapes);
+	for (int i = 0; i < WaveShapes::NumWaveShapes; ++i)
+	{
+		s_waveTables[i] = new sample_t*[s_waveTablesPerWaveformCount +1];
+		memset(s_waveTables[i], 0, sizeof(sample_t) * s_waveTablesPerWaveformCount);
+	}
+}
+
+void Oscillator::createFFtPlans()
```
PhysSong: Could you create something like `destroyFFTPlans` as well as this function?

src/core/Oscillator.cpp
```diff
@@ -88,6 +118,206 @@ void Oscillator::update( sampleFrame * _ab, const fpp_t _frames,
...
+		s_waveTables[WaveShapes::SquareWave][i] = m_generatedWaveTable;
+	}
+	//generate triangle tables
+	for (int i = 0; i < s_waveTablesPerWaveformCount; ++i)
+	{
+		generateTriangleWaveTable(MAX_FREQ / s_waveTableBandFreqs[i]);
+		s_waveTables[WaveShapes::TriangleWave][i] = m_generatedWaveTable;
+	}
+	//generate moogSaw tables
+	//generate signal buffer
```
PhysSong: I think it's possible to generate moog saw/exponential(actually parabolic) without FFT. Though their Fourier decompositions are not known well, but I can calculate it if you want. :wink:

curlymorphic:
@Wallacoloo @PhysSong Thanks for the review,  I shall work through the changes over the coming days

src/core/Oscillator.cpp
curlymorphic: I was intending to keep the fft, As I am intending to use the same functionality for the user wave shapes, However, I would prefer to get the current scope implemented correctly first.

SecondFlight:
Also, I remember noticing that the FFT on some notes doesn't extend all the way to the nyquist rate. How far up it would go varied per-note if I'm remembering correctly. It's possible that this isn't realistic to fix, and it doesn't sound audibly bad to me. Still would be nice to get that precision if it's not out of your way or has a huge performance impact.

Or I might be remembering things wrong. I don't have an easy analysis setup on Ubuntu so I can't test it here.

PhysSong:
> This produces a genuinely horrifying noise. I've gotten it to segfault by moving my mouse back and forth, but I can't seem to reproduce this consistently.

I guess it's an out-of-range error. Current implementation seems to handle pitch within MIDI standard correctly, but it doesn't seem like for too high or low notes.

plugins/triple_oscillator/TripleOscillator.cpp
```diff
@@ -727,6 +744,15 @@ TripleOscillatorView::TripleOscillatorView( Instrument * _instrument,
...
 				"waveform for current oscillator." ) );
 
+		PixmapButton * uwt = new PixmapButton( this, NULL );
+		uwt->move( 100, btn_y );
+		uwt->setActiveGraphic( PLUGIN_NAME::getIconPixmap(
+							"wavetable_active" ) );
+		uwt->setInactiveGraphic( PLUGIN_NAME::getIconPixmap(
+							"wavetable_inactive" ) );
+		uwt->setCheckable(true);
+		ToolTip::add( uwt, tr( "Use multiband wave table ocillators." ) );
```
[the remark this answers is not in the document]
curlymorphic: Good call. 
I have revised the tool tip text to read "Use alias-free wavetable oscillators."

Will be included in my next commit

src/core/Oscillator.cpp
```diff
@@ -88,6 +118,206 @@ void Oscillator::update( sampleFrame * _ab, const fpp_t _frames,
 	}
 }
 
+void Oscillator::generateSineWaveTable(int bands)
+{
+	bands = bands;
```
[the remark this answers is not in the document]
curlymorphic: noted,  seeing as a sinewave has no harmonics, as the generation is independant of the number of bands,   '< bands>'    has been removed from the

include/Oscillator.h
```diff
@@ -165,6 +192,34 @@ class EXPORT Oscillator
...
 	const SampleBuffer * m_userWave;
+	bool m_useWaveTable;
+
+	/* Multiband WaveTable */
+	static const int WAVETABLE_LENGTH = 1024;
+	static const int MAX_FREQ = 20000; //limit to the audio spectrum
+	sample_t *m_generatedWaveTable;
+	static float *s_waveTableBandFreqs;
+	static int s_waveTablesPerWaveformCount;
+	static sample_t ***s_waveTables;
```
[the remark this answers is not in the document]
curlymorphic: This has the additional  advantage of  removing the untidy memory allocations for the generated tables, that were never correctly deleted.

src/core/Oscillator.cpp
curlymorphic: The generation functions have been revised to take a pointer to the relevant location in `s_waveTables`. This addresses the issue of the memory leakage, and removes unnecessarily verbose code.

src/core/Oscillator.cpp
curlymorphic: Maths updated as requested.

Short descriptions added and links provided to explain the fourier generation.

src/core/Oscillator.cpp
curlymorphic: refactored as requested. I had spent too much time procrastinating this, had previously been a function,  I guess this solves the issue.

include/Oscillator.h
curlymorphic: revised

src/core/Oscillator.cpp
```diff
@@ -88,6 +118,206 @@ void Oscillator::update( sampleFrame * _ab, const fpp_t _frames,
...
+}
+
+
+int Oscillator::waveTableBandFromFreq(float freq)
+{
+	int i;
+	for (i = 0; i < s_waveTablesPerWaveformCount; ++i)
+	{
+		if (s_waveTableBandFreqs[i] > freq) return i;
+	}
```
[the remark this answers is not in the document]
curlymorphic: Agreed, this approach is better than my iterative method. minor tweeks applied, to account for the number of semitones per band, and now the ideal location for range checking

src/core/Oscillator.cpp
curlymorphic: `s_waveTableBandFreqs` has been replaced with the `freqFromWaveTableBand(i)` function.

Researching the current codebase for similar implementations leads to https://github.com/LMMS/lmms/blob/62d505b2e6af03f3f2d63ed7f23c254fa64b9f95/src/core/NotePlayHandle.cpp#L513

The implementation in NotePlayHandle is unsuitable for use in this context, as it is designed to handle Base frequency and instrument pitching, The requirements of the oscillator is not to convert from midi, as this is already correctly done but from wave table band.

include/Oscillator.h
Wallacoloo: @curlymorphic I don't know any specific resources to point you to regarding compiler optimizations. If you can read assembly code, you can figure out what kinds of things compilers are capable of by typing code into http://godbolt.org and then trying to figure out what the assembly code is doing. You can even select old compiler versions and see how the code they generate and the optimization they can do has improved over time. But that can be a lot of work.

lukas-w: @curlymorphic The optimization @Wallacoloo is describing is called [common subexpression elimination](https://en.wikipedia.org/wiki/Common_subexpression_elimination).

curlymorphic:
> It would appear that playing a note too high has some unintended side effects. If you're going to test this, turn your volume down.
> 
> Steps to reproduce:
> 
>     Set 3osc to the saw or square waveforms
>     Move the tuning control above the piano to the lowest possible setting
>     Click the highest note you can
> 
> This produces a genuinely horrifying noise. I've gotten it to segfault by moving my mouse back and forth, but I can't seem to reproduce this consistently.

@SecondFlight good catch, It was indeed an out of range error, and has been rectified, and will be included on the next push

include/Oscillator.h
```diff
@@ -80,15 +83,22 @@ class EXPORT Oscillator
 	}
 
 
+	void waveTableInit();
```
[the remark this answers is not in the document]
curlymorphic: Done.

src/core/Oscillator.cpp
curlymorphic: The FFT analysis has remained, as the functionality has now been extended to the user waveforms

curlymorphic:
All the above code review comments and bug reports have been addressed. In addition the user wave forms can now band limited, greatly improving the scope of this pull request.

As a stress test, I have had 40 OSC3 with all 3 oscillators using the wave tables, at 30% cpu reading on the LMMS gui, that includes various release times, and a mixture of wave and user waves. resulting in 720 instances of the Oscillator class running simultaneously

Having compared with the band limited waves used by Monstro, this implementation has various advantages

- Tonally different, giving the user a greater pallet of sounds
- Reduced Aliasing
- All wave forms except noise band limited
- User Wave forms band limited

I would not propose to replace the Monstro implementation, or to use in OSC3 based on the increased pallet alone.

I will of course update the wiki artice linked in the OP in due course, once this PR is finalized, As a quick representation the below image is a comparison of the FFT plots, The top image being the Monstro band limited saw, the bottom the proposed OSC3 implementation
image

Wallacoloo:
@curlymorphic I'm sorry I didn't get to this sooner. Everything looks ready to merge, except for the two new comments I just left.

include/Oscillator.h
```diff
@@ -79,6 +83,14 @@ class EXPORT Oscillator
 		delete m_subOsc;
 	}
 
+	static void waveTableInit();
+	static void destroyFFTPlans();
```
PhysSong: Can you move this line after `createFFTPlans` for consistency?

src/CMakeLists.txt
```diff
@@ -141,6 +141,7 @@ SET(LMMS_REQUIRED_LIBS ${LMMS_REQUIRED_LIBS}
 	${SAMPLERATE_LIBRARIES}
 	${SNDFILE_LIBRARIES}
 	${EXTRA_LIBRARIES}
+	${FFTW3F_LIBRARIES}
```
PhysSong: Really minor, but I think it'd be better if this line is before `${EXTRA_LIBRARIES}`.

src/core/SampleBuffer.cpp
```diff
@@ -1482,3 +1484,7 @@ SampleBuffer::handleState::~handleState()
 {
 	src_delete( m_resamplingData );
 }
+
+
+
+//sample_t SampleBuffer::m_userAntiAliasWaveTable[Oscillator::WAVE_TABLES_PER_WAVEFORM_COUNT][Oscillator::WAVETABLE_LENGTH] = {0};
```
PhysSong: I think we don't need to left this comment.

src/core/main.cpp
```diff
@@ -236,7 +237,6 @@ int noInputFileError()
 	return usageError( "No input file specified" );
 }
 
-
```
PhysSong: You may revert meaningless and unnecessary newline changes in some files.

include/Oscillator.h
curlymorphic: Done.

src/CMakeLists.txt
curlymorphic: Done.

src/core/SampleBuffer.cpp
curlymorphic: Done.

include/OscillatorConstants.h
```diff
@@ -0,0 +1,45 @@
...
+{
+public:
+	static const int WAVETABLE_LENGTH = 2446; //minimum size of table to have all bands for midi note 1
+	static const int MAX_FREQ = 20000; //limit to the audio spectrum
+
+	//SEMITONES_PER_TABLE, the smaller the value the smoother the harmonics change on frequency sweeps
+	// with the trade off of increased memory requirements to store the wave tables
+	// require memory = NumberOfWaveShapes*WAVETABLE_LENGTH*(MidiNoteCount/SEMITONES_PER_TABLE)*BytePerSample_t
+	// 7*2446*(128/1)*4 = 8766464 bytes
+	static const int SEMITONES_PER_TABLE = 1;
```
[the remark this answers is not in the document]
curlymorphic: Yes that is the intention, When pitch-bending at lower frequencies, the signal could be heard to become brighter in steps as the thresholds are crossed. The reduction in width of the bands, reduces the magnitude of each step.

src/core/main.cpp
curlymorphic: Done

include/Oscillator.h
PhysSong: Okay.

include/SampleBuffer.h
```diff
@@ -255,6 +257,9 @@ class EXPORT SampleBuffer : public QObject, public sharedObject
 	static QString tryToMakeAbsolute(const QString & file);
 
 
+	sample_t m_userAntiAliasWaveTable[OscillatorConstants::WAVE_TABLES_PER_WAVEFORM_COUNT][OscillatorConstants::WAVETABLE_LENGTH] ;
```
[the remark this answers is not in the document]
curlymorphic: Done, this is much better.

curlymorphic:
Updated the handling of the wavetables for the userWaveShape, the buffer is now dynamically allocated only when used by `Oscillator` greatly reducing memory usage.

@Wallacoloo, @PhysSong  I feel I have addressed all outstanding comments

src/core/Oscillator.cpp
```diff
@@ -88,6 +101,211 @@ void Oscillator::update( sampleFrame * _ab, const fpp_t _frames,
...
+	//copy to generateWaveTable
+	for (int i = 0; i < OscillatorConstants::WAVETABLE_LENGTH; ++i)
+	{
+		table[i] = s_sampleBuffer[i] / max;
+	}
+}
+
+void Oscillator::generateAntiAliasUserWaveTable(SampleBuffer *sampleBuffer)
+{
+	delete sampleBuffer->m_userAntiAliasWaveTable;
```
PhysSong: Possible memory leaks: `sampleBuffer->m_userAntiAliasWaveTable[i]`(`0 <= i < WAVE_TABLES_PER_WAVEFORM_COUNT`) are not deleted.

I also think you can move allocation/deallocation code to `SampleBuffer` instead. If so, you can share the deletion code with `SampleBuffer` destructor.
Alternatively, you may use `std::unique_ptr` as suggested by @Wallacoloo. I can provide an example code if you want.

curlymorphic:
@PhysSong Yes I shall address, I am nearing the end of a busy project, and will have some time to finish this issue.

PhysSong:
@curlymorphic Are you still here?

curlymorphic:
I Shall not be issuing any additional commits to this pull request, please feel free to make any amendments if any developer wishes

he29-net:
Hi; I would be interested in getting this PR finished and merged, so I started to play around with it in some free moments. It seems to work fine, although it had to be rebased to master before it even compiled (some headers missing in Zyn, a missing Qt include somewhere else, ladspa libraries failing dynamic linking.. apparently this old version does not like my up-to-date system?).

One thing I noticed though: it seems that the band-limited waveform is selected purely based on the pressed key, with no regard to detuning. In other words, the harmonic series of any (non-sine) tone played with the default triple-osc preset abruptly ends at 10 kHz for osc. 2 and at 5 kHz for osc. 3. This obviously makes a big difference to the final sound so I assume it was not intended?

Are there any other known issues I missed, apart from this and the memory leak? I also noticed the wave tables take some time to generate when LMMS starts, and only a single CPU core was utilized. So spawning a few threads to take care of it instead could be a good idea as well (although at this point I didn't even see the code yet so I guess I'll decide when it comes to that).

SecondFlight:
Good find. That sounds like a bug to me. I don't remember seeing any other issues, but that doesn't mean there aren't any.

he29-net:
So, I think I have the memory issues resolved (yes, plural -- apart from the leak, valgrind also found some uninitialized values used in conditions -- one in this code and one in my spectrum analyzer, so I fixed that as well).

I tried to use `unique_ptr` as suggested in the discussion, but I could not get it working. One problem is that `std::make_unique` is only available with C++14, but even after trying to work around that, it seems using `unique_ptr` for 2D arrays is very tricky and problematic. At least for me. So I gave up and kept the ordinary pointers.

Now, the question is: how do I get my changes here? The only way I can think of is making a new PR based on my modified branch; all the commits should remain visible, so `curlymorphic` should properly get the credit. Or someone could cherry-pick it the changes from my branch and add them here. https://github.com/he29-net/lmms/tree/oscillator-fix

I also found 2 more issues:
 - if I enable wavetable mode in tripleosc, load a user waveform into osc 1 and then switch osc. 2 or 3 into user waveform mode _without_ loading a waveform into it, osc 1 becomes silent;
 - the threshold used to discard weak spectral bands in the user wave table may be still too high: when loading `fmsine2.flac` (I assume it is one of the defaults?), all harmonic content suddenly disappears when the amplitude drops below approximately -42 dB. I know there is not much to be heard there, but it is bound to raise some questions, since the new spectrum analyzer goes down to -50 dB by default and the band-limited waveforms are so clean that this sharp drop won't be hidden by noise.. :)

he29-net:
The detuning issue should be fixed now as well. All commits are available at:
https://github.com/he29-net/lmms/commits/oscillator-fix

I also parallelized the initialization for platforms that properly support `std` multi-threading (i.e. not `mingw`) and the wave table generation time dropped from around 4 s to around 2 s on my machine. The saw wave is the bottleneck, for some reason it takes 2 seconds (square and triangle takes each around 1 s, sine and FFT are practically instant).

And I also reduced the cutoff threshold for the exponential wave (reasoning same as with the user wave). That should be about it, so far I don't know about any other issues that would block this PR.

curlymorphic:
Thanks for finishing and fixing the bugs in this pull request. I can not speak for LMMS as I am no longer active, but as for the pull request, if it makes life easier and the commit logs cleaner I would not object if this was to be squashed into a single new commit and pull request, no need to worry about keeping my name in the logs.  @he29-net You deserve the credit for this.

@PhysSong would probably be best to ask on how to handle the pull request.

JohannesLorenz:
Can anyone summarize what needs to be done next with this PR?

he29-net:
Someone with write access should merge my branch from:
https://github.com/he29-net/lmms/commits/oscillator-fix
into this PR (or get the patch in here by other means). It fixes the outstanding memory leak and a few other issues. It is a few months old though, so I'm not sure if it resolves all the current conflicts with master.

And a final round of testing may be needed as well, I'm not sure how much testing was done in the original review.

JohannesLorenz:
@he29-net Normally, you are allowed to push on other users' PRs. Did you try? Otherwise I'll merge it and push for you.

JohannesLorenz:
@he29-net Does your branch make sense on its own? If yes, I'd suggest you make a PR against master and then we rebase that PR on master. Otherwise, I would cherry-pick it all from your branch, i.e. just copy the commits.

he29-net:
@JohannesLorenz I was worried it could somehow mess up Dave's attribution, so I did not even try. But perhaps it won't be so bad; from what I saw in the mean time, the squash commits tend to credit multiple authors, so hopefully that includes all the commits in the PR?

Anyway, Dave wrote earlier that we should not worry too much about keeping his name in the logs if it makes things easier, so I'll open a new PR for the fixed version as you suggest, since this seems to be the case that "makes things easier". :)

JohannesLorenz:
This was not my suggestion, but it seems OK, too :smile:

he29-net:
> This was not my suggestion, but it seems OK, too smile

Wait, what? :D Why else would you suggest to make a PR using my branch then?

I'm just running a test build to see if anything broke after I merged current master into the branch. If it works, I'll open the PR and see how it looks..

JohannesLorenz:
Oh, I just thought you could make

1. a PR with only your fix against master, but it sounded like
2. you wanted a PR with dave's commits + your fix?

he29-net:
Well, "PR with only my fix against master" makes zero sense, since the code that I fixed does not exist in master. :) My fixed branch = Dave's branch + a few extra commits (my fixes).

PhysSong:
Continued in #5826, closing.

PR #4399, opened 2018-06-03
Equaliser, Disable analyser when not visible

### LINKED ISSUE #4389: Equalizer: The visualizer takes up most of the processing time of the plugin, regardless of whether or not the plugin is open
Tested on my current working branch (master from a week ago) with debugging symbols.

Steps to reproduce:
1) Add equalizers to the master channel until you see a significant cpu drop when you play a note.
2) Disable visualizers on all equalizers
3) Play a note, notice that the CPU is significantly lower.

With visualizers enabled:
image

Without visualizers enabled:
image

curlymorphic:
Previously the Equalizer plugin, would continue to perform FFT analysis on the audio when the plugin was not viable. This has a negative affect on performance.

This pull request, suspends the FFT while the plugin is not visible
fixes #4389


SecondFlight:
Any reason this can't be merged?

Issue #1191, opened 2014-10-03
(Windows) Allow international characters in file names and paths


mc0re:
When using a WAV / MP3 / ,,, file, and path contains non-ASCII (e.g. Cyrillic) characters, LMMS (AudioProcessor) declines to load the file. Moreover, when adding a file from the built-in browser panel, all non-ASCII characters are replaced with question marks. Moving and renaming the file helps. Please, fix.

_(2014-10-03 @tresf added steps to reproduce, screenshot)_
**Steps to reproduce:**
1. Switch to a Windows computer
2. Download a file, preset or plugin.
3. Rename to `АБВГДЕЖЅZЗИІКЛМНОПҀРСТȢѸФХѾЦЧШЩЪЫЬѢꙖѤЮѦѪѨѬѠѺѮѰѲѴ`
4. Add file to SongEditor
5. File doesn't work

![image](https://cloud.githubusercontent.com/assets/6345473/4507159/01fe77f2-4b0a-11e4-8ab1-3c0406e88d71.png)

tresf:
If I'm reading the bug report correctly, we may be causing this bug in [`SampleBuffer.cpp#L193`](https://github.com/LMMS/lmms/blob/master/src/core/SampleBuffer.cpp#L193) when we convert the file name directly from `QString` to `char` instead of switching to `QFile` first.

Perhaps internally, we should use `QFile` for all file loading to let Qt handle the cross-platform filename encoding issues.

PR #4401, opened 2018-06-04
Fix some encoding issues (mostly on Windows)

### LINKED ISSUE #2662: VST plugins fail to load from folders containing diacritical marks
Problem looks just like with every VST loading problem.

Steps to reproduce:
1) Download any VST plugin into a directory, in which path there are diacritical marks (in Polish we have these: ąćęłńóśżź), for example C:/Users/Jacek Mariański/VST/plugin.dll
2) Try to load them.

Easily fixable by moving into any other folder. Like to C:/VST/

### LINKED ISSUE #3855: ZynAddSubFX resets its sound to default
An issue in LMMS 1.2.0-rc4. When you open ZynAddSubFX and you create a sound and then close ZynAddSubFX, it resets the sound to default.
The sound resets when you click on "Show GUI" to close the GUI.

Running Windows 10/64 bit

I'm adding a video of me showing that issue: https://youtu.be/NJ4zkQKZa74

PhysSong:
This pull request fixes several encoding issue. Every issue except STK one are Windows-specific.
- ZynAddSubFX setting loading/saving
- VST DLL loading
- VST setting loading/saving
- Sample file decoding
- MIDI import
- STK rawwave path (incomplete on Windows due to upstream issue)

Issues remaining on Windows:
- .pat file loading
- .gig file loading
- STK rawwave path (which `toLocal8Bit` fails to handle)

I'm planning to track remaining issues in #1191.

Fixes #3855, Fixes #2662, and almost #1191.

### DIFF
--- a/plugins/MidiImport/MidiImport.cpp
+++ b/plugins/MidiImport/MidiImport.cpp
@@ -26,8 +26,11 @@
 #include <QDomDocument>
 #include <QDir>
 #include <QApplication>
+#include <QFile>
 #include <QMessageBox>
 #include <QProgressDialog>
+
+#include <sstream>
 
 #include "MidiImport.h"
 #include "TrackContainer.h"
@@ -279,8 +282,6 @@ public:
 
 bool MidiImport::readSMF( TrackContainer* tc )
 {
-	QString filename = file().fileName();
-	closeFile();
 
 	const int preTrackSteps = 2;
 	QProgressDialog pd( TrackContainer::tr( "Importing MIDI-file..." ),
@@ -291,7 +292,11 @@ bool MidiImport::readSMF( TrackContainer* tc )
 
 	pd.setValue( 0 );
 
-	Alg_seq_ptr seq = new Alg_seq(filename.toLocal8Bit(), true);
+	std::stringstream stream;
+	QByteArray arr = readAllData();
+	stream.str(std::string(arr.constData(), arr.size()));
+
+	Alg_seq_ptr seq = new Alg_seq(stream, true);
 	seq->convert_to_beats();
 
 	pd.setMaximum( seq->tracks()  + preTrackSteps );
--- a/plugins/stk/mallets/mallets.cpp
+++ b/plugins/stk/mallets/mallets.cpp
@@ -620,7 +620,7 @@ malletsSynth::malletsSynth( const StkFloat _pitch,
 	{
 		Stk::setSampleRate( _sample_rate );
 		Stk::setRawwavePath( QDir( ConfigManager::inst()->stkDir() ).absolutePath()
-						.toLatin1().constData() );
+						.toLocal8Bit().constData() );
 #ifndef LMMS_DEBUG
 		Stk::showWarnings( false );
 #endif
@@ -670,7 +670,7 @@ malletsSynth::malletsSynth( const StkFloat _pitch,
 	{
 		Stk::setSampleRate( _sample_rate );
 		Stk::setRawwavePath( QDir( ConfigManager::inst()->stkDir() ).absolutePath()
-						.toLatin1().constData() );
+						.toLocal8Bit().constData() );
 #ifndef LMMS_DEBUG
 		Stk::showWarnings( false );
 #endif
@@ -718,7 +718,7 @@ malletsSynth::malletsSynth( const StkFloat _pitch,
 	{
 		Stk::setSampleRate( _sample_rate );
 		Stk::setRawwavePath( QDir( ConfigManager::inst()->stkDir() ).absolutePath()
-						.toLatin1().constData() );
+						.toLocal8Bit().constData() );
 #ifndef LMMS_DEBUG
 		Stk::showWarnings( false );
 #endif
--- a/plugins/vst_base/RemoteVstPlugin.cpp
+++ b/plugins/vst_base/RemoteVstPlugin.cpp
@@ -88,6 +88,7 @@ struct ERect
 #include "lmms_basics.h"
 #include "Midi.h"
 #include "communication.h"
+#include "IoHelper.h"
 
 #include "VstSyncData.h"
 
@@ -678,9 +679,9 @@ void RemoteVstPlugin::init( const std::string & _plugin_file )
 
 
 
-static void close_check( int fd )
+static void close_check( FILE* fp )
 {
-	if( close( fd ) )
+	if( fclose( fp ) )
 	{
 		perror( "close" );
 	}
@@ -790,7 +791,7 @@ void RemoteVstPlugin::destroyEditor()
 
 bool RemoteVstPlugin::load( const std::string & _plugin_file )
 {
-	if( ( m_libInst = LoadLibrary( _plugin_file.c_str() ) ) == NULL )
+	if( ( m_libInst = LoadLibraryW( toWString(_plugin_file).c_str() ) ) == NULL )
 	{
 		// give VstPlugin class a chance to start 32 bit version of RemoteVstPlugin
 		if( GetLastError() == ERROR_BAD_EXE_FORMAT )
@@ -1072,13 +1073,13 @@ void RemoteVstPlugin::saveChunkToFile( const std::string & _file )
 		const int len = pluginDispatch( 23, 0, 0, &chunk );
 		if( len > 0 )
 		{
-			int fd = open( _file.c_str(), O_WRONLY | O_BINARY );
-			if ( ::write( fd, chunk, len ) != len )
+			FILE* fp = F_OPEN_UTF8( _file, "wb" );
+			if ( fwrite( chunk, len, 1, fp ) != len )
 			{
 				fprintf( stderr,
 					"Error saving chunk to file.\n" );
 			}
-			close_check( fd );
+			close_check( fp );
 		}
 	}
 }
@@ -1237,7 +1238,7 @@ void RemoteVstPlugin::savePreset( const std::string & _file )
 	if (!isPreset &&!chunky) uIntToFile = (unsigned int) m_plugin->numPrograms;
 	pBank->numPrograms = endian_swap( uIntToFile );
 
-	FILE * stream = fopen( _file.c_str(), "w" );
+	FILE * stream = F_OPEN_UTF8( _file, "w" );
 	fwrite ( pBank, 1, 28, stream );
 	fwrite ( progName, 1, isPreset ? 28 : 128, stream );
 	if ( chunky ) {
@@ -1289,7 +1290,7 @@ void RemoteVstPlugin::loadPresetFile( const std::string & _file )
 	unsigned int * pLen = new unsigned int[ 1 ];
 	unsigned int len = 0;
 	sBank * pBank = (sBank*) new char[ sizeof( sBank ) ];
-	FILE * stream = fopen( _file.c_str(), "r" );
+	FILE * stream = F_OPEN_UTF8( _file, "r" );
 	if ( fread ( pBank, 1, 56, stream ) != 56 )
 	{
 		fprintf( stderr, "Error loading preset file.\n" );
@@ -1390,12 +1391,12 @@ void RemoteVstPlugin::loadChunkFromFile( const std::string & _file, int _len )
 {
 	char * chunk = new char[_len];
 
-	const int fd = open( _file.c_str(), O_RDONLY | O_BINARY );
-	if ( ::read( fd, chunk, _len ) != _len )
+	FILE* fp = F_OPEN_UTF8( _file, "rb" );
+	if ( fread( chunk, 1, _len, fp ) != _len )
 	{
 		fprintf( stderr, "Error loading chunk from file.\n" );
 	}
-	close_check( fd );
+	close_check( fp );
 
 	pluginDispatch( effSetChunk, 0, _len, chunk );
 
--- a/plugins/zynaddsubfx/zynaddsubfx/src/Misc/QtXmlWrapper.cpp
+++ b/plugins/zynaddsubfx/zynaddsubfx/src/Misc/QtXmlWrapper.cpp
@@ -55,9 +55,12 @@
 #include <sstream>
 #include <cstdarg>
 #include <zlib.h>
-#include "lmmsconfig.h"
 #include "../globals.h"
 #include "Util.h"
+
+// Include LMMS headers
+#include "lmmsconfig.h"
+#include "IoHelper.h"
 
 
 struct XmlData
@@ -193,11 +196,12 @@ int QtXmlWrapper::dosavefile(const char *filename,
                            int compression,
                            const char *xmldata) const
 {
+    FILE *file = F_OPEN_UTF8(std::string(filename), "w");
+    if(file == NULL) {
+        return -1;
+    }
+
     if(compression == 0) {
-        FILE *file;
-        file = fopen(filename, "w");
-        if(file == NULL)
-            return -1;
         fputs(xmldata, file);
         fclose(file);
     }
@@ -210,7 +214,7 @@ int QtXmlWrapper::dosavefile(const char *filename,
         snprintf(options, 10, "wb%d", compression);
 
         gzFile gzfile;
-        gzfile = gzopen(filename, options);
+        gzfile = gzdopen(fileToDescriptor(file), options);
         if(gzfile == NULL)
             return -1;
         gzputs(gzfile, xmldata);
@@ -313,7 +317,8 @@ int QtXmlWrapper::loadXMLfile(const std::string &filename)
 char *QtXmlWrapper::doloadfile(const std::string &filename) const
 {
     char  *xmldata = NULL;
-    gzFile gzfile  = gzopen(filename.c_str(), "rb");
+
+    gzFile gzfile  = gzdopen(fileToDescriptor(F_OPEN_UTF8(filename, "rb")), "rb");
 
     if(gzfile != NULL) { //The possibly compressed file opened
         std::stringstream strBuf;             //reading stream
--- a/src/core/DrumSynth.cpp
+++ b/src/core/DrumSynth.cpp
@@ -26,10 +26,12 @@
 
 #include "DrumSynth.h"
 
-#include <fstream>
+#include <sstream>
 #include <cstring>
 
 #include <math.h>     //sin(), exp(), etc.
+
+#include <QFile>
 
 #ifdef LMMS_BUILD_WIN32
 #define powf pow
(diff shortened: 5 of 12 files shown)


zonkmachine:
> Every issue except STK one are Windows-specific

I replicated issue #2662 on LinuxMint/Qt4 . It's indeed fixed by this PR.

Ready to merge?

plugins/zynaddsubfx/zynaddsubfx/src/Misc/QtXmlWrapper.cpp
```diff
@@ -193,9 +202,18 @@ int QtXmlWrapper::dosavefile(const char *filename,
...
                            const char *xmldata) const
 {
+    QFile f(filename);
+    f.open(QIODevice::WriteOnly);
+#ifdef LMMS_BUILD_WIN32
+    int fd = _dup(f.handle());
+#else
+    int fd = dup(f.handle());
+#endif
+    f.close();
```
Wallacoloo: I see you're using this to open files by their descriptor instead of the filename. Is it `gzopen` that's giving you problems, `fopen`, or both? Because if it's only `gzopen` that's broken, it might be possible to just always `fopen`, and then call `gzdopen(fileno(file), options)` if compression is enabled, rather than doing this trickery with duplicating the file handles and such.

plugins/zynaddsubfx/zynaddsubfx/src/Misc/QtXmlWrapper.cpp
```diff
@@ -313,7 +331,16 @@ int QtXmlWrapper::loadXMLfile(const std::string &filename)
...
     char  *xmldata = NULL;
-    gzFile gzfile  = gzopen(filename.c_str(), "rb");
+    QFile f(QString::fromUtf8(filename.data(), filename.size()));
+    f.open(QIODevice::ReadOnly);
+#ifdef LMMS_BUILD_WIN32
+    int fd = _dup(f.handle());
+#else
+    int fd = dup(f.handle());
+#endif
+    f.close();
```
Wallacoloo: I wonder if you could offload this logic to a helper function,
```
static int fdopenFromQString(QString filename, QIODevice::OpenMode mode)
```
to avoid duplicating the body in multiple functins, and then leave a comment near that function explaining why the function is necessary (to most readers, this code will look silly and they may be tempted to simplify it back to the way it was before).

src/core/DrumSynth.cpp
```diff
@@ -162,17 +164,20 @@ float DrumSynth::waveform(float ph, int form)
...
     char *k, *b;
     int len = 0;
 
     line = (char*)malloc(200);
 
-    is.open (file, ifstream::in);
+    QFile f(file);
+    f.open(QIODevice::ReadOnly);
+    QByteArray dat = f.readAll().constData();
+    is.str(string(dat.constData(), dat.size()));
```
Wallacoloo: The casual reader will see this and think "we're reading the entire file into a string, and then operating on it only as a stream? Why not directly open the file with `ifstream`?" and will be tempted to change this back to how it was previously. I think there should be some [short] comment explaining why we _must_ open the file with `QFile`, possibly linking back to this github PR.

src/core/SampleBuffer.cpp
```diff
@@ -416,7 +412,10 @@ f_cnt_t SampleBuffer::decodeSampleSF( const char * _f,
 	f_cnt_t frames = 0;
 	bool sf_rr = false;
 
-	if( ( snd_file = sf_open( _f, SFM_READ, &sf_info ) ) != NULL )
+
+	QFile f(_f);
```
Wallacoloo: I'm thinking these should all be commented with something along the lines of
```
// Open the file with QFile and pass the descriptor to sf_open_fd instead of directly calling sf_open
// for better filename encoding support. See https://github.com/LMMS/lmms/pull/4401
```
so that people know not to revert this change.

plugins/vst_base/RemoteVstPlugin.cpp
```diff
@@ -119,6 +119,29 @@ RemoteVstPlugin * __plugin = NULL;
 HWND __MessageHwnd = NULL;
 
 
+static std::wstring toWString(std::string s)
+{
+	/*
+	QString qstr = QString::fromUtf8(s.data(), int(s.size()));
```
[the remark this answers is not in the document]
PhysSong: Beacause it requires including Qt headers(on Windows). If that don't matter (we already do that), I can switch to QString is needed.

plugins/vst_base/RemoteVstPlugin.cpp
```diff
@@ -119,6 +119,29 @@ RemoteVstPlugin * __plugin = NULL;
...
+	ret.resize(len);
+	MultiByteToWideChar(CP_UTF8, MB_ERR_INVALID_CHARS, s.data(), s.length(), &ret[0], len);
+	return ret;
+}
+
+#ifdef LMMS_BUILD_WIN32
+#define F_OPEN(a, b) _wfopen(toWString(a).c_str(), L##b)
+#else
+#define F_OPEN(a, b) fopen((a).c_str(), b)
+#endif
```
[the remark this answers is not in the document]
PhysSong: Forgot to commit that. Thank you!

plugins/zynaddsubfx/zynaddsubfx/src/Misc/QtXmlWrapper.cpp
PhysSong: Both are broken, because C library only supports characters in local code page.

plugins/zynaddsubfx/zynaddsubfx/src/Misc/QtXmlWrapper.cpp
PhysSong: Good suggestion!

src/core/SampleBuffer.cpp
PhysSong: Good point.

plugins/MidiImport/MidiImport.cpp
```diff
@@ -281,6 +284,11 @@ bool MidiImport::readSMF( TrackContainer* tc )
 {
 	QString filename = file().fileName();
 	closeFile();
+	QFile file(filename);
+	if (!file.open(QIODevice::ReadOnly))
+	{
+		return false;
+	}
```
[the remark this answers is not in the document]
PhysSong: I wrote like this because `m_outputFile` is a private member of `AudioFileDevice`. I'll add a method to work around.

plugins/zynaddsubfx/zynaddsubfx/src/Misc/QtXmlWrapper.cpp
Wallacoloo: Oh, oops. I had typed out that comment and after digging deeper to better understand the problem I meant to remove it before posting the review >.<

Yeah, I agree with your conclusion here. It's not pretty, but I don't see any better solution.

plugins/vst_base/RemoteVstPlugin.cpp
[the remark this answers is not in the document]
Wallacoloo: Ah, I see. @lukas-w has offered me advice on where it's okay to use Qt stuff in the past. I think things like QString, at least, are fine everywhere, but hopefully he'll chime in here to confirm that this is OK if he's not too busy.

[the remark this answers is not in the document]
PhysSong: To be more clear, `RemoteVstPlugin` use Qt libraries for IPC on Windows. So I think that would be fine.

plugins/vst_base/RemoteVstPlugin.cpp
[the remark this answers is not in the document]
PhysSong: Addressed.

Wallacoloo:
@PhysSong Thanks for the improvements!

I'm marking this as "approved" now, since you've addressed any concerns I had. But it looks like Lukas wanted to review this too, so you should probably wait until that before merging :tada:

plugins/stk/mallets/mallets.cpp
```diff
@@ -620,7 +620,12 @@ malletsSynth::malletsSynth( const StkFloat _pitch,
 	{
 		Stk::setSampleRate( _sample_rate );
 		Stk::setRawwavePath( QDir( ConfigManager::inst()->stkDir() ).absolutePath()
-						.toLatin1().constData() );
+#ifdef LMMS_BUILD_WIN32
+						.toLocal8Bit().constData() );
+#else
+						.toUtf8().constData() );
+#endif
+
```
lukas-w: I think we can (should?) use `toLocal8Bit` on Unix as well. It should be equivalent to `toUtf8` on most Linux and macOS systems. Correct me if I'm wrong.

plugins/vst_base/RemoteVstPlugin.cpp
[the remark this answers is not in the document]
lukas-w: @PhysSong is right, this is not a good place to use Qt because that would add MinGW Qt as a dependency on Linux builds. We can however use C++11 functions to convert to UTF-16:
```c++
#include <locale>
#include <codecvt>
```
```c++
std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>> converter;
return converter.from_bytes(s);
```

This does the same thing, just in fewer lines. I think we already use C++11 for `RemoteVstPlugin` on `stable-1.2`, but I didn't check.

plugins/zynaddsubfx/zynaddsubfx/src/Misc/QtXmlWrapper.cpp
```diff
@@ -193,9 +216,11 @@ int QtXmlWrapper::dosavefile(const char *filename,
                            int compression,
                            const char *xmldata) const
 {
+   int fd = fdopenFromUtf8(filename, QIODevice::WriteOnly);
```
lukas-w: Why not use the same method here that's used in `RemoteVstPlugin.cpp`? We can move the `toWString` method and the `F_OPEN` defines to a common header and include that in both files.

PhysSong: I think `F_OPEN` macro is an ugly workaround for `fopen`. I can do that if needed, but I don't like the macro approach.

plugins/vst_base/RemoteVstPlugin.cpp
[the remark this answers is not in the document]
PhysSong: > because that would add MinGW Qt as a dependency on Linux builds.

It seems like we can use host version Qt, but I don't think it's a good idea because it adds 32bit Qt as a dependency on 64bit Linux.

> We can however use C++11 functions to convert to UTF-16:

That doesn't work with our old MinGW because of missing header files(it should work with newest `g++-mingw-w64`). Those functions are deprecated in C++17, too.

plugins/zynaddsubfx/zynaddsubfx/src/Misc/QtXmlWrapper.cpp
lukas-w: @PhysSong You can make it a function instead.

PhysSong: Something tricky about that: `"r"` vs `L"r"` or something like that. That's why I chose macro approach.

lukas-w: @PhysSong Sounds like you need to convert the mode string as well. So something like this:

```c++
std::FILE* openFile(std::string filename, std::string mode)
{
#ifdef LMMS_BUILD_WIN32
	return _wfopen(toWString(filename), toWString(mode));
#else
	return fopen(filename.data(), mode.data());
#endif
}
```

PhysSong: Looks fine for me, except missing `data()` for WIN32.

lukas-w: Can you elaborate on that?

PhysSong: I mean `return _wfopen(toWString(filename).data(), toWString(mode).data());`.

lukas-w: Of course, sorry. I thought you were telling me `data()` isn't available for WIN32 and started getting angry at MinGW. :)

PhysSong: I don't really know which approach will be better here. Both approach seem to have not much overheads. Qt provides a nice API for this, but `_wfopen` looks somewhat straightforward.
Alternatively, `CreateFileW` + [`_open_osfhandle`](https://msdn.microsoft.com/en-us/library/bdts1c9x.aspx)(what `QFile` internally does) may work for getting file descriptor.

lukas-w: Neither is really better than the other, but they do the same thing so I think they should be consolidated into a single function. The Qt solution is not an option for this because it's not available in `RemoteVstPlugin`.

PhysSong: > but they do the same thing
They *almost* do. On Windows, `QFile` holds native file handle and converts it to a C file handle when requested.
`_wfopen` creates a `FILE` which holds a C handle to the file. That's why I used the former here, but I can switch to the latter if needed.

lukas-w: Surely `F_OPEN` and `fdopenFromUtf8` fulfil the same task, which is opening a C `FILE` given an UTF8-encoded file name, right? Even though they don't work the same way internally, they "do the same thing" as in "they solve the same problem". From that point of view, this is duplicate code and there should only be one function for it.

PhysSong: Then where should I move `F_OPEN` and related stuffs? I can't find a good place for it.
I was not using that in Zyn because it(more precisely `zlib`) needs `fdopen`-able handle, not a `FILE` pointer. However, getting(and duplicating) file handle is not hard so I can easily add it.

lukas-w: Good question, I can't find a good place either. Maybe just create a new header? Something like `lmmsio.h` (inspired by `stdio.h`), although that name is very similar to https://lmms.io/.

PhysSong: I also used `QFile` for getting file descriptors which will be used by `libsndfile`. Do you think it's worth unifying everything with `F_OPEN` based macros/functions?

PhysSong: I think either `lmms_io.h` or `IoHelper.h` will be fine. I'll address this point soon.

PhysSong: Or how about putting them in `RemotePlugin.h`? I can't use-cases of UTF-8 string filename except in remote plugins.

PhysSong: Is it fine for our Zyn to include LMMS' headers at all? I think this will be fine in `stable-1.2`, but I'm not sure about `master`'s submodule. It would be the last concern for me...

PhysSong: I know it doesn't make sense to build https://github.com/LMMS/zynaddsubfx as a standalone, of course. However, I think it's possible to check if it's building with LMMS or not, though not really needed.

PhysSong:
@lukas-w I addressed these issues and pushed to a [separate branch](https://github.com/PhysSong/lmms/tree/encoding2). I'll push this change to the original branch when/if you says okay.

PhysSong:
Oops, accidentally pushed them. I can roll back to [original commits](https://github.com/PhysSong/lmms/commit/1f09c05133722a5274782bb0d1d1078f200a3f80) whenever needed.

lukas-w:
@PhysSong Please roll back to the original commits, all the comment/review history on GitHub is lost otherwise. It's okay to have one extra commit on top of the original ones, there's no need to rewrite history.

PhysSong:
> all the comment/review history on GitHub is lost otherwise.

What do you mean? Review comments are still there unless you commented on specific commits, the are just collapsed.

Anyway, I agree that it's not necessary to rewrite history. I just wanted to keep things cleaner, but that was all. If you're still unhappy with this force-push, I can roll back.

lukas-w:
> What do you mean?

Sorry, I was being imprecise. The comments are still there, but the context is lost. Review comments are always on specific commits and you can't view that commits any more, so you can't see the changes the review refers to. Try clicking on the "View changes" button next to a review, this is what you'll get now:

image

> If you're still unhappy with this force-push, I can roll back.

Yes, please do 👍

PhysSong:
I just noticed even a rebasing can cause such a thing. I just accidentally rebased commits because I didn't know that... I'll revert to the revision before rebasing if needed.

lukas-w:
@PhysSong Apart from restoring the pre-force-push commits, nope.

PhysSong:
Then do you mind if I merge this(probably with cleaner history)?

lukas-w:
I don't mind, go ahead ❤️

plugins/vst_base/RemoteVstPlugin.cpp
```diff
@@ -1072,13 +1073,13 @@ void RemoteVstPlugin::saveChunkToFile( const std::string & _file )
 		const int len = pluginDispatch( 23, 0, 0, &chunk );
 		if( len > 0 )
 		{
-			int fd = open( _file.c_str(), O_WRONLY | O_BINARY );
-			if ( ::write( fd, chunk, len ) != len )
+			FILE* fp = F_OPEN_UTF8( _file, "wb" );
+			if ( fwrite( chunk, len, 1, fp ) != len )
```
lukas-w: Only spotted this now: `fwrite`returns the `count` of elements successfully written, so in this case it will return `1` on success, not `len`. Therefore this condition is always leading to a false error message on success (except when coincidentally `len == 1`).

PhysSong: Thank you! Changed to `fwrite( chunk, 1, len, fp )` in 0f3b41f590ed8d573ce3891f6502f211d1be176a.

PhysSong: Also synced into `master` via 170a46e7a3b5bc3eb49999874344b82cda1cc53f.

Issue #4405, opened 2018-06-05
RC6 and RC5 does not start on my windows 10


metalgigio:
Hi,
as subject RC6 and RC5 does not start on my windows 10. It appears splash screen but the program does not load.
1.1.3 version works correctly.
Theres some log i can check to se what happen?

lukas-w:
Unfortunately, our debugging options on Windows are very limited at this time. What happens after the splash screen shows? Does LMMS crash, do you get any error messages?

metalgigio:
On RC6  splash screen appears, then show some loading message on bottom left and then disappear but program does not start. No error messages.
RC1, RC2, RC3, RC4 start correctly, RC5 does not even show splash screen.
I use Windows 10 64bit updated to last version

PhysSong:
I guess you're using PotrAudio. I've done a quick test:
1. Open LMMS 1.1.3 and set audio.backend to PortAudio
2. Open RC5 or later, it will crash
I'm not sure it's the identical issue, but I guess so because PortAudio have been disabled since RC5.

metalgigio:
this is my lmmsrc.xml:

```
<?xml version="1.0"?>
<!DOCTYPE lmms-config-file>
<lmms version="1.2.0-rc4">
  <MidiAlsaRaw device=""/>
  <app nommpz="0" configured="1" displaydbfs="0" nomsgaftersetup="0"/>
  <audioportaudio backend="Windows WASAPI" device="Altoparlanti (Dispositivo High Definition Audio)"/>
  <audiosdl device=""/>
  <mixer framesperaudiobuffer="256" hqaudio="0" mididev="WinMM MIDI" audiodev="PortAudio"/>
  <paths stkdir="data:/stk/rawwaves/" fldir="C:/Users/gigi\" sf2dir="C:/Users/gigi/lmms/samples/sf2/" gigdir="C:/Users/gigi/lmms/samples/gig/" laddir="C:/Program Files/LMMS\plugins\ladspa\" defaultsf2="" artwork="data:/themes/default/" vstdir="C:\Users\gigi\Documents\VST\" backgroundartwork="" workingdir="C:/Users/gigi\lmms\"/>
  <tooltips disabled="0"/>
[...]

PR #4407, opened 2018-06-06
Better handling of invalid audio/MIDI backend

PhysSong:
Fix crash when nonexistent/not supported audio/MIDI backend is written in `.lmmsrc.xml`.
Supersedes #4372.

### DIFF
--- a/include/Mixer.h
+++ b/include/Mixer.h
@@ -308,6 +308,9 @@ public:
 	void requestChangeInModel();
 	void doneChangeInModel();
 
+	static bool isAudioDevNameValid(QString name);
+	static bool isMidiDevNameValid(QString name);
+
 
 signals:
 	void qualitySettingsChanged();
--- a/src/core/Mixer.cpp
+++ b/src/core/Mixer.cpp
@@ -829,14 +829,138 @@ void Mixer::runChangesInModel()
 	}
 }
 
+bool Mixer::isAudioDevNameValid(QString name)
+{
+#ifdef LMMS_HAVE_SDL
+	if (name == AudioSdl::name())
+	{
+		return true;
+	}
+#endif
 
 
+#ifdef LMMS_HAVE_ALSA
+	if (name == AudioAlsa::name())
+	{
+		return true;
+	}
+#endif
+
+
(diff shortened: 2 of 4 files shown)


PhysSong:
Merging this one will stop crash in #4405. However, LMMS will show up a dialog to select audio backend(the default value will be Dummy). Before releasing RC7, we should add an upgrade logic or some alternative way to change PortAudio to SDL. Should I add that in this pull request, or open a separate one?

zonkmachine:
> Should I add that in this pull request, or open a separate one?

Just fix it up here.

PhysSong:
> Before releasing RC7, we should add an upgrade logic or some alternative way to change PortAudio to SDL.

I have some ideas about it:
- Fallback to default audio backend if invalid one is selected. In Windows case, it will automatically fallback to SDL.
- Write a separate upgrade logic for PortAudio -> SDL. Then the fix will be out of scope of this PR.
- In case of invalid backend selections, show default backend instead of Dummy one in setup dialog at launch.

zonkmachine:
> * Fallback to default audio backend if invalid one is selected. In Windows case, it will automatically fallback to SDL.

If this means that you won't have to restart lmms again after doing a selection this is my choice. A separate  upgrade routine sounds like a bit of an over kill. Maybe one of those floating messages to give a hint of the changes to the user would be a good idea.

lukas-w:
Either option 1 or 3 is fine in my eyes.

PhysSong:
Then I'll implement option 1 here. It looks simple to me.

PhysSong:
I found that if the audio backend is set to an empty string, LMMS will use the default one and not show the setup dialog.
In case of invalid settings, the dialog should show up. However, I don't know if this should be the same for empty strings. It depends on how I implement the fallback logic, and the converse is also true.
How should I handle the empty-string case?

PhysSong:
Implemented the fallback logic via c5810d2b5b1a21e8af57d897079fb9ec433ac072.

PhysSong:
I will merge this one soon if no one objects. Further refactorings and UX improvements are also possible, but I'd like to defer them to the `master` branch.

PR #4433, opened 2018-06-17
Qt4 cleanups

lukas-w:
Removes old Qt4 compatibility code from `master` branch.


plugins/VstEffect/VstEffectControlDialog.cpp
```diff
@@ -272,7 +272,7 @@ void VstEffectControlDialog::showEvent(QShowEvent *_se)
 
 VstEffectControlDialog::~VstEffectControlDialog()
 {
-#if !(QT_VERSION < 0x050000 && defined(LMMS_BUILD_LINUX))
+#ifndef LMMS_BUILD_LINUX
```
PhysSong: The condition is now always true(`!(false && something) == !false == true`). So this check should be dropped.

lukas-w: Right! Thanks.

PhysSong:
I think `QPixmap::grabWidget` should be replaced with `QWidget::grab`(a member function introduced in Qt5) in `StringPairDialog.cpp` and `Track.cpp` because the former is depecated in Qt5.

lukas-w:
@PhysSong I bet there's a lot of deprecated Qt API that we still use. I can try to remove them all by finding them with the `QT_DEPRECATED_WARNINGS` compile flag. Should I push them to this branch or open a separate PR for it?

PhysSong:
> I bet there's a lot of deprecated Qt API that we still use.

Well, I think finding them is first. If a function that isn't deprecated in Qt4 is in Qt5, it may be in scope of this pull request. In other cases, I don't know.

lukas-w:
@PhysSong Removed all deprecated Qt API use via 11c6932.

plugins/patman/patman.cpp
```diff
@@ -78,7 +78,7 @@ PLUGIN_EXPORT Plugin * lmms_plugin_main( Model *, void * _data )
 
 patmanInstrument::patmanInstrument( InstrumentTrack * _instrument_track ) :
 	Instrument( _instrument_track, &patman_plugin_descriptor ),
-	m_patchFile( QString::null ),
+	m_patchFile( QString() ),
```
PhysSong: I guess you can just construct `QString` with no argument.

src/core/Track.cpp
```diff
@@ -95,7 +95,7 @@ TextFloat * TrackContentObjectView::s_textFloat = NULL;
 TrackContentObject::TrackContentObject( Track * track ) :
 	Model( track ),
 	m_track( track ),
-	m_name( QString::null ),
+	m_name( QString() ),
```
PhysSong: Here as well.

src/gui/AutomatableModelView.cpp
```diff
@@ -43,8 +43,8 @@ static float floatFromClipboard(bool* ok=nullptr);
 
 AutomatableModelView::AutomatableModelView( ::Model* model, QWidget* _this ) :
 	ModelView( model, _this ),
-	m_description( QString::null ),
-	m_unit( QString::null )
+	m_description( QString() ),
+	m_unit( QString() )
```
PhysSong: Also here.

src/gui/SetupDialog.cpp
```diff
@@ -1273,7 +1269,7 @@ void SetupDialog::openWorkingDir()
 {
 	QString new_dir = FileDialog::getExistingDirectory( this,
 					tr( "Choose LMMS working directory" ), m_workingDir );
-	if( new_dir != QString::null )
+	if( new_dir != QString() )
```
PhysSong: I think it'd be more clear to use `isNull` or `isEmpty`.

lukas-w:
@PhysSong Done via c670d48.

lukas-w:
Merged via ff1c30e35daefceb67118e28ce88ddf454941125, somehow GitHub doesn't detect it.

Issue #1676, opened 2015-01-22
Add an uninstall target for make


unfa:
Currently there's no way to automatically remove a manual installation of LMMS (done with make install command).

lukas-w:
From the [CMake FAQ](https://cmake.org/Wiki/CMake_FAQ#Can_I_do_.22make_uninstall.22_with_CMake.3F):

> By default, CMake does not provide the "make uninstall" target, so you cannot do this. We do not want "make uninstall" to remove useful files from the system.
> 
> If you want an "uninstall" target in your project, then nobody prevents you from providing one. You need to delete the files listed in install_manifest.txt file. Here is how to do it. First create file cmake_uninstall.cmake.in in the top-level directory of the project:
> 
> ```cmake
> if(NOT EXISTS "@CMAKE_BINARY_DIR@/install_manifest.txt")
>   message(FATAL_ERROR "Cannot find install manifest: @CMAKE_BINARY_DIR@/install_manifest.txt")
> endif(NOT EXISTS "@CMAKE_BINARY_DIR@/install_manifest.txt")
> 
> file(READ "@CMAKE_BINARY_DIR@/install_manifest.txt" files)
> string(REGEX REPLACE "\n" ";" files "${files}")
> foreach(file ${files})
>   message(STATUS "Uninstalling $ENV{DESTDIR}${file}")
>   if(IS_SYMLINK "$ENV{DESTDIR}${file}" OR EXISTS "$ENV{DESTDIR}${file}")
>     exec_program(
>       "@CMAKE_COMMAND@" ARGS "-E remove \"$ENV{DESTDIR}${file}\""
>       OUTPUT_VARIABLE rm_out
>       RETURN_VALUE rm_retval
>       )
>     if(NOT "${rm_retval}" STREQUAL 0)
>       message(FATAL_ERROR "Problem when removing $ENV{DESTDIR}${file}")
>     endif(NOT "${rm_retval}" STREQUAL 0)
>   else(IS_SYMLINK "$ENV{DESTDIR}${file}" OR EXISTS "$ENV{DESTDIR}${file}")
>     message(STATUS "File $ENV{DESTDIR}${file} does not exist.")
>   endif(IS_SYMLINK "$ENV{DESTDIR}${file}" OR EXISTS "$ENV{DESTDIR}${file}")
> endforeach(file)
> ```
> 
> Then in the top-level CMakeLists.txt add the following logic:
> 
> ```cmake
> # uninstall target
[...]

PR #4448, opened 2018-06-24
Added an uninstall target for make

PJ-Finlay:
I modified CMakeLists.txt to add an uninstall target for make. If you run "make uninstall" after running "make install" a cmake script at cmake/uninstall.cmake will be run that attempts to delete all of the files in install_manifest.txt. This is to fix issue #1676.

### DIFF
--- a/cmake/uninstall.cmake
+++ b/cmake/uninstall.cmake
@@ -1,0 +1,78 @@
+MESSAGE(STATUS "Attempting to create uninstall target for make")
+
+#Remove all of the files listed in install_manifest.txt
+SET(INSTALL_MANIFEST_PATH "${CMAKE_CURRENT_BINARY_DIR}/install_manifest.txt")
+IF(NOT EXISTS ${INSTALL_MANIFEST_PATH})
+	MESSAGE(FATAL_ERROR "Could not find install manifest at ${CMAKE_CURRENT_BINARY_DIR}/install_manifest.txt\nThis may be because 'make install' has not been run or install_manifest.txt has been deleted")
+ENDIF()
+MESSAGE(STATUS "install_manifest.txt found")
+FILE(STRINGS ${INSTALL_MANIFEST_PATH} FILES_TO_REMOVE)
+FOREACH(FILE_TO_REMOVE ${FILES_TO_REMOVE})
+	IF(EXISTS ${FILE_TO_REMOVE})
+		EXECUTE_PROCESS(
+			COMMAND ${CMAKE_COMMAND} -E remove ${FILE_TO_REMOVE}
+			RESULT_VARIABLE EXIT_CODE
+		)
+		IF(${EXIT_CODE} EQUAL 0)
+			MESSAGE(STATUS "Successfully removed file ${FILE_TO_REMOVE}")
+		ELSE()
+			MESSAGE(FATAL_ERROR "Failed to remove file ${FILE_TO_REMOVE} with error code ${EXIT_CODE}")
+		ENDIF()
+	ELSE()
+		MESSAGE(WARNING "Could not find file ${FILE_TO_REMOVE}")
+	ENDIF()
+ENDFOREACH(FILE_TO_REMOVE)
+
+#Remove empty directories created during installation
+#Required to support IN_LIST operator
+CMAKE_POLICY(SET CMP0057 NEW)
+
+# Checks if a directory is empty and saves the result in out_var
+FUNCTION(IS_EMPTY_DIR OUT_VAR DIR)
+	FILE(GLOB FILES "${DIR}/*")
+	LIST(LENGTH FILES NUM_FILES)
+	IF(NUM_FILES EQUAL 0)
+		SET(${OUT_VAR} TRUE PARENT_SCOPE)
+	ELSE()
+		SET(${OUT_VAR} FALSE PARENT_SCOPE)
+	ENDIF()
+ENDFUNCTION()
+
+# Recursively append all parent directories of path to out_var
+FUNCTION(PARENT_DIRECTORIES OUT_VAR PATH)
+	GET_FILENAME_COMPONENT(PARENT "${PATH}" DIRECTORY)
+	IF(PARENT AND NOT PARENT STREQUAL PATH AND NOT PARENT IN_LIST ${OUT_VAR})
+		LIST(APPEND ${OUT_VAR} ${PARENT})
+		PARENT_DIRECTORIES(${OUT_VAR} "${PARENT}")
+	ENDIF()
+	SET(${OUT_VAR} ${${OUT_VAR}} PARENT_SCOPE)
+ENDFUNCTION()
+
+# Removes all empty parent directories of the given files
+FUNCTION(REMOVE_EMPTY_DIRECTORIES FILES)
+	SET(DIRECTORIES)
+	FOREACH(FILE_TO_REMOVE ${FILES})
+		PARENT_DIRECTORIES(DIRECTORIES "${FILE_TO_REMOVE}")
+	endforeach()
+	LIST(REMOVE_DUPLICATES DIRECTORIES)
+	# Sort and reverse so we remove subdirectories first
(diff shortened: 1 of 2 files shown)


Wallacoloo:
Looks good to me.

cmake/uninstall.cmake
```diff
@@ -0,0 +1,26 @@
...
+			ELSE()
+				MESSAGE(FATAL_ERROR "Failed to remove file ${FILE_TO_REMOVE} with error code ${EXIT_CODE}")
+			ENDIF()
+		ELSE()
+			MESSAGE(WARNING "Could not find file ${FILE_TO_REMOVE}")
+		ENDIF()
+	ENDFOREACH(FILE_TO_REMOVE)
+ELSE()
+	MESSAGE(FATAL_ERROR "Could not find install manifest at ${CMAKE_CURRENT_BINARY_DIR}/install_manifest.txt")
+	MESSAGE(FATAL_ERROR "This may be because 'make install' has non been run or install_manifest.txt has been deleted")
```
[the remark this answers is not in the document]
PJ-Finlay: Good catch, I just amended the commit to combine them into one line with a \n character.

cmake/uninstall.cmake
```diff
@@ -0,0 +1,25 @@
+MESSAGE(STATUS "Attempting to create uninstall target for make")
+SET(INSTALL_MANIFEST_PATH "${CMAKE_CURRENT_BINARY_DIR}/install_manifest.txt")
+IF(EXISTS ${INSTALL_MANIFEST_PATH})
+	MESSAGE(STATUS "install_manifest.txt found")
+	FILE(STRINGS ${INSTALL_MANIFEST_PATH} FILES_TO_REMOVE)
+	FOREACH(FILE_TO_REMOVE ${FILES_TO_REMOVE})
+		IF(EXISTS ${FILE_TO_REMOVE})
+			EXEC_PROGRAM(
```
lukas-w: `exec_program` is deprecated, you should use [`execute_process`](https://cmake.org/cmake/help/latest/command/execute_process.html) instead.

cmake/uninstall.cmake
```diff
@@ -0,0 +1,25 @@
...
+				MESSAGE(STATUS "Successfully removed file ${FILE_TO_REMOVE}")
+			ELSE()
+				MESSAGE(FATAL_ERROR "Failed to remove file ${FILE_TO_REMOVE} with error code ${EXIT_CODE}")
+			ENDIF()
+		ELSE()
+			MESSAGE(WARNING "Could not find file ${FILE_TO_REMOVE}")
+		ENDIF()
+	ENDFOREACH(FILE_TO_REMOVE)
+ELSE()
+	MESSAGE(FATAL_ERROR "Could not find install manifest at ${CMAKE_CURRENT_BINARY_DIR}/install_manifest.txt\nThis may be because 'make install' has non been run or install_manifest.txt has been deleted")
```
lukas-w: Typo in `has non been run`.

PJ-Finlay:
I think I made the requested changes. However, I think CMAKE_CURRENT_BINARY_DIR is the right way to find the install_manifest.txt file.

Changes:
-Are you sure about changing CMAKE_CURRENT_BINARY_DIR to configure file? I realize that CMAKE_CURRENT_BINARY_DIR is set to the current directory. When a user runs make they need to be in the build directory where the the install_manifest.txt file is. make has a -C option which lets you run it from a different directory, but this solution still works while doing that. 

-I changed to the requested control flow.

-I switched exec_program to execute_process.

-I fixed the typo in "has non been run"

lukas-w:
Thanks! 👍 

> Are you sure about changing CMAKE_CURRENT_BINARY_DIR to configure file?

I believe it depends on the generator used, but it works at least with GNU Make and Ninja in my tests, so I guess it's fine.

PhysSong:
BTW, this pull request targets to `stable-1.2` which is on a feature freeze. @lukas-w Is it okay to merge this into `stable-1.2`?

lukas-w:
Running `make uninstall` still leaves all the (empty) folders created by installing. If we can address that in this PR, I guess it's fine to merge into `stable-1.2`. Otherwise, I'd rather merge it into `master`.

PJ-Finlay:
I can try to also delete the folders.

PJ-Finlay:
@lukas-w how should removing folders be done? Currently it's just deleting all of the files in install_manifest.txt. I could just delete the include/lmms & lib/lmms folders, and the things that are in share/ directly but I don't know how portable this would be. Also it seems like most other people I can find online don't do this and just delete from install_manifest.txt: [https://gergap.wordpress.com/2015/08/18/cmake-uninstall/](https://gergap.wordpress.com/2015/08/18/cmake-uninstall/), [https://bitbucket.org/eigen/eigen/pull-requests/91/added-cmake-uninstall-target/diff](https://bitbucket.org/eigen/eigen/pull-requests/91/added-cmake-uninstall-target/diff), [https://gist.github.com/royvandam/3033428](https://gist.github.com/royvandam/3033428).

lukas-w:
I hacked something together, but it requires CMake 3.3 because it uses `IN_LIST` so it would only be good for `master`:

```cmake
# Checks if a directory is empty and saves the result in out_var
function(is_empty_dir out_var dir)
	file(GLOB files "${dir}/*")
	list(LENGTH files num_files)
	if(num_files EQUAL 0)
		set(${out_var} TRUE PARENT_SCOPE)
	else()
		set(${out_var} FALSE PARENT_SCOPE)
	endif()
endfunction()

# Recursively gets all parent directories
function(parent_directories out_var path)
	get_filename_component(parent "${path}" DIRECTORY)
	if(parent AND NOT parent STREQUAL path AND NOT parent IN_LIST ${out_var})
		LIST(APPEND ${out_var} ${parent})
		parent_directories(${out_var} "${parent}")
	endif()
	set(${out_var} ${${out_var}} PARENT_SCOPE)
endfunction()

# Removes all empty parent directories of the given files
function(remove_empty_directories files)
	set(directories)
	foreach(FILE_TO_REMOVE ${FILES_TO_REMOVE})
		parent_directories(directories "${FILE_TO_REMOVE}")
	endforeach()
	list(REMOVE_DUPLICATES directories)
	# Sort and reverse so we remove subdirectories first
	list(SORT directories)
	list(REVERSE directories)

	foreach(dir ${directories})
		# Skip directories not inside the install prefix
		if(NOT (EXISTS "${dir}" AND dir MATCHES "^${CMAKE_INSTALL_PREFIX}/"))
			continue()
		endif()

		is_empty_dir(dir_empty "${dir}")
		if(dir_empty)
			message(STATUS "Removing empty directory ${dir}")
			file(REMOVE_RECURSE "${dir}")
		elseif()
			message(STATUS "Skipping non-empty directory ${dir}")
		endif()
	endforeach()
endfunction()

...

remove_empty_directories("${FILES_TO_REMOVE}")
```
```cmake
ADD_CUSTOM_TARGET(uninstall
	COMMAND ${CMAKE_COMMAND} -DCMAKE_INSTALL_PREFIX="${CMAKE_INSTALL_PREFIX}" -P "${CMAKE_CURRENT_SOURCE_DIR}/cmake/uninstall.cmake"
)
```

PJ-Finlay:
I added @lukas-w's code to remove the empty directories created during installation into my pull request:
- I fixed a typo in the remove_empty_directories function where the ${FILES_TO_REMOVE} variable was being used from the parent scope instead of the ${FILE} variable from the functions arguments.

- I set [Cmake policy CMP0057](https://cmake.org/cmake/help/latest/policy/CMP0057.html) to NEW to use the IN_LIST operator.

- I changed the minimum Cmake version to 3.3 which is required for the IN_LIST operator.

- I switched to upper case because that's what the rest of the project seemed to be doing.

lukas-w:
Thanks @PJ-Finlay. I made some minor changes via 63fd427c5e386fe769d1dde10057a094e68f2fd4 and will merge into `master` soon.

PR #4450, opened 2018-06-25
Fix occasional audio interface deadlock

### LINKED ISSUE #4406: SDL deadlock on exit
Occasionally, when exiting LMMS, the process remains hanging. I don't have a backtrace unfortunately, because gdb takes ages to launch LMMS on my machine, and I can't reliably trigger the issue, but here's what I gleaned from Process Hacker on the 1.2.0-RC6 release:

Main thread is blocked on `SDL_LockAudio()` in `AudioSdl::stopProcessing`:
https://github.com/LMMS/lmms/blob/3bf3fab0e33b7bb199b2c4b32b1ccd0e1da096ca/src/core/audio/AudioSdl.cpp#L106-L115
`AudioSdl::stopProcessing` is called from `Mixer::stopProcessing` is called from `LmmsCore::destroy`.

The SDL audio lock is held by SDL which is in the audio callback, which is blocked on `m_reader_sem.acquire()` in `fifoBuffer::read`:
https://github.com/LMMS/lmms/blob/32f07ac317b42dda42a1d00b0a3b1fa1feef0655/include/fifo_buffer.h#L60-L67
`fifoBuffer::read` is called from `Mixer::nextBuffer` is called from `AudioDevice::getNextBuffer` is called from `AudioSdl::sdlAudioCallback`.

Related: #4032.

PhysSong:
Fixes #4406 by ensuring audio thread has read the `NULL` buffer before stop processing. Dropping them have caused several regressions(#4406, #3397, #3396, etc.) because `stopProcessing` of backend can be called before it receives `NULL` buffer.
Instead of relying on per-backend semaphores(Dropped for SDL in 98346f7a9a7a7c07e91bdad953229f1ba7523c2e, for JACK in 1dcacbb95e5182f784d8a951dee63c96c7487830), `Mixer::fifoWriter` now checks it.

### DIFF
--- a/include/AudioJack.h
+++ b/include/AudioJack.h
@@ -37,7 +37,6 @@
 #include <QtCore/QVector>
 #include <QtCore/QList>
 #include <QtCore/QMap>
-#include <QMutexLocker>
 
 #include "AudioDevice.h"
 #include "AudioDeviceSetupWidget.h"
@@ -108,7 +107,6 @@ private:
 
 	bool m_active;
 	bool m_stopped;
-	QMutex m_processingMutex;
 
 	MidiJack *m_midiClient;
 	QVector<jack_port_t *> m_outputPorts;
--- a/include/AudioPortAudio.h
+++ b/include/AudioPortAudio.h
@@ -145,7 +145,6 @@ private:
 	int m_outBufSize;
 
 	bool m_stopped;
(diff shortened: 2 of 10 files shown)


PhysSong:
Merging since there was no review and I can find any regressions.

PR #4451, opened 2018-06-25
Fix hang on exit when using specific MIDI devices on Windows

### LINKED ISSUE #3920: Hang when closing related to WinMM MIDI driver
Configuration:  Windows 7 SP1 32-bit, LMMS V1.2.0RC4, M-Audio MIDISport 2+2, driver version 6.1.3

When closing LMMS, a "zombie" LMMS process is left in the Task Manager which cannot be killed.  These chew up about 40MB at a time.  If LMMS is started and stopped repeatedly, more copies of the process will show up in the Task Manager.  In testing, I found that changing the audio back-end to "Dummy" did not make a difference in this issue, but changing the MIDI back-end to "Dummy" enabled LMMS to close successfully.  If I changed it back to WinMM, LMMS left a dead process running.  There are no error messages displayed, and it leaves no entries in the Windows system logs.

I have other applications that I use with MIDI (OpenMPT, saVihoST, SimplePiano) which use the MIDISport 2+2 driver directly, and they don't seem to have any difficulty with my setup.  I would be curious about other MIDI USB devices and WinMM, but I don't have any others to test.  Anybody with USB keyboard controllers might want to have a look at your task manager!

lmms_task_manager

PhysSong:
Fixes #3920.

### DIFF
--- a/src/core/midi/MidiWinMM.cpp
+++ b/src/core/midi/MidiWinMM.cpp
@@ -248,9 +248,13 @@ void MidiWinMM::closeDevices()
 	m_outputSubs.clear();
 
 	QMapIterator<HMIDIIN, QString> i( m_inputDevices );
+
+	HMIDIIN hInDev;
 	while( i.hasNext() )
 	{
-		midiInClose( i.next().key() );
+		hInDev = i.next().key();
+		midiInReset( hInDev );
+		midiInClose( hInDev );
 	}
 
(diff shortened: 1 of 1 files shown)


PhysSong:
Since this one was [already tested](https://github.com/LMMS/lmms/issues/3920#issuecomment-399174536), I think it's fine to merge. Any objections?

Issue #1825, opened 2015-03-05
ZynAddSubFX "Show GUI" causes severe audio dropout.


mikobuntu:
https://www.youtube.com/watch?v=IMt_oy-9ljY&feature=youtu.be

This has been a long known issue that using the " Show GUI " button on ZASF will cause a significant dropout of all audio during playback.

Hopefully there is a way to avoid this completely or at least make the transition less noticeable .

I understand the concept that ZASF runs in a Remote process .

thanks Mikobuntu :)

Sti2nd:
So you are proposing to check if the event prevents the audio from playing cause obviously the specs of a computer would also play a role.

mikobuntu:
@Sti2nd  Yeah it's true that lower spec computers could struggle with loading ZASF , this particualer machine i am using for this demo is quite adequate .
![specs](https://cloud.githubusercontent.com/assets/335217/6508491/051b2c9a-c351-11e4-97fe-cf37c01ec329.png)

tresf:
> So you are proposing to check if the event prevents the audio from playing cause obviously the specs of a computer would also play a role.

We should review the Show GUI button and see if we're accidentally starting the ZynGUI thread with the wrong priority.

I've never experienced this issue personally.

-Tres

mikobuntu:
@tresf  Great, hopefully this can be fixed and I have always had this problem as far back as I can remember.

mikobuntu:
@unfa Do you experience this behaviour using ZynAdd ?

tresf:
@mikobuntu,

A good place to start is perhaps here, where the slot is setup:

[ZynAddSubFx.cpp#L538](https://github.com/LMMS/lmms/blob/master/plugins/zynaddsubfx/ZynAddSubFx.cpp#L538)

I would recommend checking the thread priority in `toggleUI()` here: [ZynAddSubFx.cpp#L633](https://github.com/LMMS/lmms/blob/master/plugins/zynaddsubfx/ZynAddSubFx.cpp#L633) to see if it is firing the GUI show event with an abnormally high priority.

If it does, we should experiment switching it to use `QThread::IdlePriority`, `QThread::LowPriority` or `QThread::LowestPriority`.

Umcaruje:
I can confirm that this has happened to me almost every time I open the Zyn GUI.

musikBear:
[...]

PR #4460, opened 2018-07-04
Use tryLock in audio threads for VST/ZynAddSubFX

PhysSong:
Prevent loading VST or toggling ZynAddSubFX GUI from blocking entire audio processing. It still blocks audio output from specific VeSTige/ZynAddSubFX but not whole playback.
Address #1825.

### DIFF
--- a/plugins/VstEffect/VstEffectControls.cpp
+++ b/plugins/VstEffect/VstEffectControls.cpp
@@ -62,7 +62,7 @@ void VstEffectControls::loadSettings( const QDomElement & _this )
 {
 	//m_effect->closePlugin();
 	//m_effect->openPlugin( _this.attribute( "plugin" ) );
-	if (!m_effect->m_pluginMutex.tryLock()) {return;}
+	m_effect->m_pluginMutex.lock();
 	if( m_effect->m_plugin != NULL )
 	{
 		m_vstGuiVisible = _this.attribute( "guivisible" ).toInt();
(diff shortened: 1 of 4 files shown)


plugins/VstEffect/VstEffectControls.cpp
```diff
@@ -62,7 +62,7 @@ void VstEffectControls::loadSettings( const QDomElement & _this )
 {
 	//m_effect->closePlugin();
 	//m_effect->openPlugin( _this.attribute( "plugin" ) );
-	m_effect->m_pluginMutex.lock();
+	if (!m_effect->m_pluginMutex.tryLock()) {return;}
```
DomClark: Won't this cause the plugin settings not to be loaded if the mutex can't be locked? I think this code runs on the main thread anyway so it shouldn't affect audio processing if it has to wait for the mutex.

PhysSong: You're right here, probably a mistake.

PhysSong: Fixed in e79aeb711d1756690ff7ceba4d1ff29e0578c227.

PhysSong:
Merge?

PhysSong:
`PlayHandle::doProcessing` clears the buffer before processing, so I believe it doesn't matter.

falkTX:
If that is the case then all is good :)
Now I wonder if carla plugin is doing the same... I think not..

PhysSong:
> Now I wonder if carla plugin is doing the same... I think not..

I guess the mutex can be dropped with a proper workaround for race conditions(maybe `Mixer::requestChangeInModel` stuff?). However, the mutex is less likely to cause severe blocking than the VST/ZynAddSubFX case...

PhysSong:
I guess this patch might break rendering. Should I restrict this fix to playback only?

PhysSong:
I found that this is insufficient because `RemotePlugin::process()` calls `lock()`. I think https://github.com/LMMS/lmms/issues/5074#issuecomment-516846313 might be the ultimate solution.

PR #4439, opened 2018-06-17
Add MSVC Appveyor CI

lukas-w:
This PR adds automated MSVC testing on Appveyor.

Based on #4438

PR #4468, opened 2018-07-08
Sdl audio fix

devnexen:


Wallacoloo:
After this patch, the left argument to qMin has type `size_t`, and the right is `unsigned long`. So this assumes `size_t` is the same type as `unsigned long`. You probably want to just cast both arguments to `uint`, since that's the type of the variable you're assigning this expression to.

Correct me if I'm wrong - C++ implicit conversions can be confusing.

@lukas-w @Reflexe why are `m_currentBufferFramesCount` and `m_currentBufferFramePos` typed as `uint64_t`? They're being used as indices into an array - one that's certainly not > 2^32 in size - so it looks to me like a machine-native `int` would make sense, or a `size_t` if you want the semantics of an index. Seems like using either of those types would simplify their use.

devnexen:
I caught up into OpenBSD so maybe my assumption was not wide enough. Maybe just changing the types to those to size_t would really fix it and we can close this PR then.

lukas-w:
This fix is already part of #4439 (fixed by changing the type to `size_t`), see https://github.com/LMMS/lmms/pull/4439/files#diff-e74d39cfad524614724dbb924124ffe6. I hope you don't mind me closing this PR in favour of it. Thank you anyway @devnexen and sorry for the doubled work. There's been no objection to #4438 and #4439 and the changes are relatively small so I think I'll merge them soon.

devnexen:
No problem at all as long it s fixed thx !

Issue #4436, opened 2018-06-17
Panning knob in song-editor partially hidden in swedish


saurid:
The panning knob in the song-editor is pushed to the right, making it difficult to see what value it has.
![image](https://user-images.githubusercontent.com/1075365/41509504-2aafb302-7255-11e8-9fea-5c67c8b6bfed.png)
Looking in the sv.ts file on row 5985, I think the translation to swedish should be the shorter PAN. The knob is centered beneath the longer PANORERA, which makes the knob partially dissapear.

Tested in LMMS 1.2.0 rc6

musikBear:
PAN must be enough for my neighbours. Danes would also say 'Panorere' or 'Panorering' but Everyone would understand PAN

saurid:
It's still named PAN in stable version I think. Not sure about other languages, but if it is more than three characters, i guess they also have the same problem as I do. VOL is also a short, so I think it was just a mistake during translation :)

zonkmachine:
This has already been fixed in master but we probably need to change it manually on stable-1.2

> PAN must be enough for my neighbours.

Correct.

> It's still named PAN in stable version I think. Not sure about other languages, but if it is more than three characters, i guess they also have the same problem as I do.

This has indeed been reported for other languages on Transifex and I do see this on stable-1.2

Issue #4285, opened 2018-04-07
Misaligned Knob in localised Delay Effect


Sauerstoffdioxid:
![grafik](https://user-images.githubusercontent.com/35171164/38457727-bc73ac10-3a94-11e8-97eb-3bcc3443b21e.png)

The delay knob of the German localisation ("VERZÖGERUNG") is misaligned, because the German term is much longer then the English one and doesn't fit into the "side-bar".

LMMS 1.2.0 RC5

liushuyu:
>The delay knob of the German localisation ("VERZÖGERUNG") is misaligned, because the German term is much longer then the English one and doesn't fit into the "side-bar".

I don't know if German localization can have a shorter version of that, in the case of impossibility to make a shorter version of translation, we might want to make the UI respond to the variable text length.

Sauerstoffdioxid:
As a German, I don't think the term can be abbreviated. However, since all other knobs use the English terms (well "rate" and German "Rate" are the same, so I don't know which is used here), one possibility would be to stick to the English terms, e.g. "Delay". Afterall the plugin itself is called Delay, too, so it shouldn't be confusing.

zonkmachine:
Same issue in our native flanger. I'll fix this in https://github.com/LMMS/lmms/pull/4482

PR #4482, opened 2018-07-13
Fix some formatting issues in translations

zonkmachine:
Many translations seem to have been done without verifying against a compiled version of lmms. Maybe on the Transifex online editor. Especially the instrument pluggin editor and the native instruments/effects has many short terms and acronyms that must stay short or the layout will go south. I've gone through all translation files and reverted the worst cases. It may be that my machine won't show all problematic translations so a second pair of eyes would be nice here.
Fixes https://github.com/LMMS/lmms/issues/4436
Fixes https://github.com/LMMS/lmms/issues/4285

PS. I've also fixed up some general issues in the Swedish translation.

### DIFF
--- a/data/locale/ja.ts
+++ b/data/locale/ja.ts
@@ -3277,7 +3277,7 @@ You can remove and move FX channels in the context menu, which is accessed by ri
     </message>
     <message>
         <source>PAN</source>
-        <translation>パニング</translation>
+        <translation>PAN</translation>
     </message>
     <message>
         <source>MIDI</source>
@@ -5984,7 +5984,7 @@ Reason: &quot;%2&quot;</source>
     </message>
     <message>
         <source>PAN</source>
-        <translation>パニング</translation>
+        <translation>PAN</translation>
     </message>
 </context>
 <context>
--- a/data/locale/sv.ts
+++ b/data/locale/sv.ts
@@ -1063,7 +1063,7 @@ If you&apos;re interested in translating LMMS in another language or want to imp
     </message>
     <message>
         <source>FREQ</source>
-        <translation>FREKV.</translation>
+        <translation>FREQ</translation>
     </message>
     <message>
         <source>Cutoff frequency</source>
@@ -1443,7 +1443,7 @@ Right clicking will bring up a context menu where you can change the order in wh
     </message>
     <message>
         <source>Freq x 100</source>
-        <translation>Frekv. x 100</translation>
+        <translation>Freq x 100</translation>
     </message>
     <message>
         <source>Modulate Env-Amount</source>
@@ -1454,7 +1454,7 @@ Right clicking will bring up a context menu where you can change the order in wh
     <name>EnvelopeAndLfoView</name>
     <message>
         <source>DEL</source>
-        <translation>RADERA</translation>
+        <translation>DEL</translation>
     </message>
     <message>
         <source>Predelay:</source>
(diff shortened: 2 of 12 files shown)


data/locale/sv.ts
```diff
@@ -1454,7 +1454,7 @@ Right clicking will bring up a context menu where you can change the order in wh
     <name>EnvelopeAndLfoView</name>
     <message>
         <source>DEL</source>
-        <translation>RADERA</translation>
```
zonkmachine: :clown_face:

zonkmachine: This is not only a horribly wrong translation but it fundamentally borks the envelope knob layout. People are busy using the Transifex online interface and don't double check against a compiled real version of lmms.

data/locale/sv.ts
```diff
@@ -1063,7 +1063,7 @@ If you&apos;re interested in translating LMMS in another language or want to imp
     </message>
     <message>
         <source>FREQ</source>
-        <translation>FREKV.</translation>
+        <translation>FREK</translation>
```
zonkmachine: Either this or just revert to no translation here.

data/locale/ja.ts
```diff
@@ -9952,4 +9952,4 @@ The LED in the lower right corner of the waveform editor determines whether the
         <translation>出力ゲイン</translation>
     </message>
 </context>
-</TS>
\ No newline at end of file
+</TS>
```
zonkmachine: One of the Japanese translators is working on a Windows machine...

data/locale/sv.ts
zonkmachine: I reverted those to Freq/FREQ and some other Instrument plugin knob texts to the English original for now.

liushuyu:
Please upload the translation fix to Transifex otherwise they will be overriden in next translation update

PhysSong:
> Please upload the translation fix to Transifex otherwise they will be overriden in next translation update

It requires some permission. @zonkmachine can update Swedish, and @liushuyu can update German.
There are no Japanese translators in our development team, but I guess @Umcaruje can push it because he's the [Team manager](https://docs.transifex.com/teams/understanding-user-roles#team-managers) for all languages.

PhysSong:
Then @liushuyu, could you update translation on Transifex once this PR gets merged?

liushuyu:
>Then @liushuyu, could you update translation on Transifex once this PR gets merged?

Sure thing.

data/locale/ja.ts
zonkmachine: I still don't get what's going on here. I haven't found a setting for formatting in my editor (xed). I think only the Swedish translation didn't see this and I haven't touched it until now I think. Are these format fixes correct or should I revert them?

zonkmachine:
OK. I have reverted the translations that mess up formatting the most, especially in the Instrument Editor and the native pluggins, but have allowed some smaller issues as to keep this PR at a reasonable size.
This is now ready for review.

zonkmachine:
Merge?

zonkmachine:
For reviewing this I don't think there's any point in going through all of it. The changes are pretty much the same across all languages. Reverting to original English as the translated line is too long and has a negative effect on the knob layout. Just comparing a couple of the languages should be enough and then the translators can take it from there.

zonkmachine:
I reverted some automatic formatting changes that my editor made. These edits may have been perfectly fine but I don't have time to look into it right now. Merging.

PhysSong:
@liushuyu Could you please upload these changes to Transifex? It seems like you haven't.

PR #4486, opened 2018-07-14
Allow building with sndio support on more systems than just OpenBSD

t6:
[Sndio](http://www.sndio.org) also supports FreeBSD and Linux.


CMakeLists.txt
```diff
@@ -56,6 +56,7 @@ OPTION(WANT_MP3LAME	"Include MP3/Lame support" ON)
 OPTION(WANT_OGGVORBIS	"Include OGG/Vorbis support" ON)
 OPTION(WANT_PULSEAUDIO	"Include PulseAudio support" ON)
 OPTION(WANT_PORTAUDIO	"Include PortAudio support" ON)
+OPTION(WANT_SNDIO	"Include sndio support" OFF)
```
PhysSong: Why did you turn off Sndio by default? I think you can turn this on and disable on Windows.

t6: > Why did you turn off Sndio by default? I think you can turn this on and disable on Windows.

Done.

I was trying to be conservative and keep the behavior on non-OpenBSD
systems the same as before.

t6:
What does sndio have to do with RoarAudio? The claim in #3596 that sndio is only available on OpenBSD is bogus. If you want to test this on Ubuntu install their [libsndio-dev](https://packages.ubuntu.com/bionic/libsndio-dev) package.

Here's a build log from building LMMS 1.2.0-rc6 on FreeBSD with this patch applied: https://people.freebsd.org/~tobik/logs/lmms-1.2.0.r6,2.log

tresf:
> The claim in #3596 that sndio is only available on OpenBSD is bogus.

It's a description, not a claim.  The message has since been removed from our `stable-1.2` branch.

> What does sndio have to do with RoarAudio

Our CI environment uses Ubuntu 14.04 and for that platform `libsndio-dev` is [only provided as a virtual package though `libroar-dev`](https://packages.ubuntu.com/trusty/libsndio-dev), which would not compile at time of adding that comment.

> Here's a build log from building LMMS 1.2.0-rc6 on FreeBSD with this patch applied: 

Thanks for the log and for the link to the bionic mirror.  Please understand this will break builds on any 14.04-based machines that have `libroar-dev` installed.  Broken builds eventually turn into bug reports and/or support questions, so it's sane behavior to guard against them.

At the time of writing this, 14.04 is old enough that most desktop users have moved away from it, so this just needs to be rebased and merged... 

> changed the base branch from `stable-1.2`  to `master`  17 days ago

@PhysSong can you explain why this PR was removed from our stable branch?  If you have no objections, let's put it back on stable and merge it.

devnexen:
I do not know how it works under Linux but is device naming similar as OpenBSD ? Does not know if it brings a different experience than with also neither.

devnexen:
I forgot to mention is sndiod installed ?

t6:
> If we're going to toggle this on for all distros now, how is this backend used? I've tried rsnd/0 but it doesn't appear to work.

For starters disable all other audio daemons.  Sndio on Linux needs exclusive access to the audio device. AFAIK `pasuspender` should temporarily disable PulseAudio. `rsnd/0` is the first audio device (it translates to the `hw:0` ALSA device) but it might not be what you need on your hardware. So try `rsnd/1`, `rsnd/2`, etc. too.

It might be easier to test this with `aucat -i /dev/urandom -f rsnd/1` first (if this works you'll hear some whitenoise, `aucat` is part of the `sndio-tools` package on Ubuntu) instead of trying it in LMMS.

devnexen:
Ok so I built all under debian and seems to work tried couple of seconds with rsnd/0 device in my side.

tresf:
> I forgot to mention is sndiod installed ?

No, it wasn't but I managed to get it to work.

> Sndio on Linux needs exclusive access to the audio device when used locally. AFAIK `pasuspender` should temporarily disable PulseAudio. 

Thanks and spot-on.  `sndio` works as long as `pasuspender` prefixes the 
LMMS command.

For the installed version:
```bash
pasuspender -- lmms
```

For the compiled version:
```bash
pasuspender -- ./lmms
```

For the AppImage version:
```
pasuspender -- ./lmms-1.2.0-rc6.32-linux-x86_64.AppImage
```

> `rsnd/0` is the first audio device

That's perfect, I'm in a VM, so there's only one back-end.  Fortunately, once PA is suspended, sndio just works.

So I think this is ready for merge to `stable-1.2` if @PhysSong agrees.   Thanks for the support assistance.  The only downside is we can't ship it with our official AppImage because 14.04 doesn't provide the proper library.

From a testing perspective, we ship a very resource intensive demo `unfa - Spoken` and it starts to under-run on virtual hardware about half through where SDL is able to keep up, but it's usable.  We'll have more real-world examples trickle in once this is merged.

PhysSong:
> So I think this is ready for merge to `stable-1.2` if @PhysSong agrees

What I was afraid of was that `stable-1.2` is on a feature freeze. However, I'd like to listen to opinions from other devs. If they agree on merging into `stable-1.2`, then I won't block that.

tresf:
> I think we can add `libsndio-dev` into our CircleCI docker image.

Sounds good.  I still think this belongs on `stable-1.2`, so you can toggle on `libsndio-dev` in Cirlce-CI anytime.

PR #4502, opened 2018-07-26
Fix invisible editors on project load by validating size

### LINKED ISSUE #4270: Piano roll, BB and Automation editors can't be opened (1.2-RC5)
When I was playing around with LMMS, at some point, LMMS couldn't allowed me to open piano roll, BB or automation editors, while other panels such as song editor, effect chain, and others could be opened just fine.

When I tried to open any of those panels, it looked like the current panel unfocused, but the piano roll didn't appear.

Here is a dump file and the project file where it happens. [Project file](https://github.com/LMMS/lmms/files/1835130/test.mmpz.zip) and [a bug report](https://github.com/LMMS/lmms/files/1835132/test.mmp.zip) (exported with `-d`).

lukas-w:
Validate an editor's size when loading and ignore it if it's empty. Fixes #4270. This only fixes the symptom, not the cause. I don't know why a project file with a zero size would be created in the first place, but I'm not sure if it's really worth investigating.


PhysSong:
Merge?

PR #4503, opened 2018-07-26
MIDI import: Reset the seek pointer after parsing magic bytes.

irrenhaus3:
Fixes a silent bug that prevented any and all SMF midi import due to the Alg_midifile_reader's parse() function erroneously reading bytes 4 to 7 instead of 0 to 3.

### DIFF
--- a/include/ImportFilter.h
+++ b/include/ImportFilter.h
@@ -73,6 +73,11 @@ protected:
 		return -1;
 	}
 
+	inline void resetReadPosition()
+	{
+		m_file.seek(0);
+	}
+
 	inline int readBlock( char * _data, int _len )
 	{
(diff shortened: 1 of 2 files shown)


PhysSong:
It seems like a regression in #4401 and thus this pull request should target to `stable-1.2`.

PhysSong:
I'll add the seek function into `readAll` instead. I should have tested properly when fixing the encoding issue.
@irrenhaus3 Anyway, thank you for reporting!

PhysSong:
Fixed via 112d34226ad1ee3f8c53f0160aba4ff27c0fb6d6.

irrenhaus3:
Very nice, thank you for your time.

PR #4505, opened 2018-07-28
MSVC: Fix SID

lukas-w:
Fixes SID for MSVC by using the provided working buffer instead of a local one to avoid the use of VLA. Changed the loop to go backwards to avoid overwriting data in the short-to-float conversion. Hope this logic is correct (it appears to sound right).

### DIFF
--- a/plugins/CMakeLists.txt
+++ b/plugins/CMakeLists.txt
@@ -94,9 +94,7 @@ IF(MSVC)
 		monstro
 		organic
 		ReverbSC
-		sid
 		vibed
-		#VstEffect
 		Xpressive
 		zynaddsubfx
 	)
--- a/plugins/sid/sid_instrument.cpp
+++ b/plugins/sid/sid_instrument.cpp
@@ -324,7 +324,7 @@ void sidInstrument::playNote( NotePlayHandle * _n,
 
 	cSID *sid = static_cast<cSID *>( _n->m_pluginData );
 	int delta_t = clockrate * frames / samplerate + 4;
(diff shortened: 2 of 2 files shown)


PhysSong:
Nice trick! However, I think you should add a comment that describes the reason for your change.

tresf:
> I think you should add a comment that describes the reason for your change.

They were obvious, so I just added them.  Hope you don't mind @lukas-w  :)

tresf:
@lukas-w the error from Circle-CI is:

```
>>>>> linuxdeployqt
linuxdeployqt 5 (commit fc64c50), build 609 built on 2019-01-06 03:55:18 UTC
Please run on a system no newer than the oldest still-supported Ubuntu LTS release.
This is so that the resulting bundle can run on all still-supported releases of Ubuntu.
Exited with code 1
```

... this seems to be caused by https://github.com/probonopd/linuxdeployqt/commit/fc64c50c09119f0175a741f15cffac295a02fdf4

PhysSong:
> ... this seems to be caused by [probonopd/linuxdeployqt@fc64c50](https://github.com/probonopd/linuxdeployqt/commit/fc64c50c09119f0175a741f15cffac295a02fdf4)

More precisely, https://github.com/probonopd/linuxdeployqt/commit/b9b57e266e080aab0f81706199bd0eb30e63f165 is the right commit. It seems like there were some issues with AppImages created with some build environments. However, I think they should have made `linuxdeployqt` print a warning when building on a newer environment.

PR #4509, opened 2018-07-30
Fix #4459: The oscilloscope color can't be changed

### LINKED ISSUE #4459: The oscilloscope color can't be changed
Line 176 below, won't change the color of the background of the oscilloscope and it wont change the color of the sound waves

VisualizationWidget  {
	background: none;
	border: none;
}

michaelgregorius:
Fix #4459  by making the oscilloscope colors available as properties so that they can be changed via style sheets. Adjust the existing styles to use the colors that have been hard coded previously.

Cleanup the `paintEvent` method of `VisualizationWidget`, e.g. by extracting a method to determine the color to use for the line and by pulling variables to the place where they are used. Fix some Clang warnings.

Also adjust `Mixer::getPeakValues` so that clients do not have to allocate the variables for the peak values.

### DIFF
--- a/data/themes/classic/style.css
+++ b/data/themes/classic/style.css
@@ -158,6 +158,9 @@ GroupBox {
 VisualizationWidget  {
 	background: none;
 	border: none;
+	qproperty-normalColor: rgb(71, 253, 133);
+	qproperty-warningColor: rgb(255, 192, 64);
+	qproperty-clippingColor: rgb(255, 64, 64);
 }
 
 /* main toolbar cpu load widget - this can have transparent bg now */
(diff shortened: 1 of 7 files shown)


Umcaruje:
Wow, that was fast!

I think this should go into `stable-1.2` because I consider it a theme enhancement, though the second commit kind of seems more suited for master.

michaelgregorius:
I have split this into #4512 which is intended to go into `stable-1.2` and #4513 which is intended to go into `master`. Closing this one now.

Umcaruje:
Thanks!

PR #4526, opened 2018-08-06
Change universal key shortcut to insert bars

### LINKED ISSUE #4522: Unable to add measure on macOS
the universal mapped key shortcut is Shift+Ins, which isn't present on Mac keyboards and does not work with an external keyboard - an easy fix would just be to implement an on-screen button to add/delete measures

lookitsnicholas:
the Ins key is absent/non-functional on macOS devices, a change is necessary for this functionality to work

**Edit by @tresf** Closes #4522

### DIFF
--- a/src/gui/editors/SongEditor.cpp
+++ b/src/gui/editors/SongEditor.cpp
@@ -318,6 +318,7 @@ void SongEditor::keyPressEvent( QKeyEvent * ke )
 						ke->key() == Qt::Key_Insert )
 	{
 		m_song->insertBar();
+	}
     #ifdef LMMS_BUILD_APPLE
         if( /*_ke->modifiers() & Qt::ShiftModifier*/
            gui->mainWindow()->isShiftPressed() == true &&
--- a/src/gui/editors/SongEditor.cpp
+++ b/src/gui/editors/SongEditor.cpp
@@ -318,7 +318,7 @@ void SongEditor::keyPressEvent( QKeyEvent * ke )
 						ke->key() == Qt::Key_Insert )
 	{
 		m_song->insertBar();
-    #ifdef Q_OS_MACOS
+    #ifdef LMMS_BUILD_APPLE
         if( /*_ke->modifiers() & Qt::ShiftModifier*/
            gui->mainWindow()->isShiftPressed() == true &&
                             ke->key() == Qt::Key_Enter || ke->key() == Qt::Key_Return )
--- a/src/gui/editors/SongEditor.cpp
+++ b/src/gui/editors/SongEditor.cpp
@@ -315,10 +315,17 @@ void SongEditor::keyPressEvent( QKeyEvent * ke )
 {
 	if( /*_ke->modifiers() & Qt::ShiftModifier*/
 		gui->mainWindow()->isShiftPressed() == true &&
-						ke->key() == Qt::Key_Enter )
+						ke->key() == Qt::Key_Insert )
 	{
 		m_song->insertBar();
-	}
+    #ifdef Q_OS_MACOS
+        if( /*_ke->modifiers() & Qt::ShiftModifier*/
+           gui->mainWindow()->isShiftPressed() == true &&
+                            ke->key() == Qt::Key_Enter || ke->key() == Qt::Key_Return )
+        {
+            m_song->insertBar();
+        }
+    #endif
 	else if(/* _ke->modifiers() & Qt::ShiftModifier &&*/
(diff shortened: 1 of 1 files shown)


src/gui/editors/SongEditor.cpp
```diff
@@ -315,7 +315,7 @@ void SongEditor::keyPressEvent( QKeyEvent * ke )
 {
 	if( /*_ke->modifiers() & Qt::ShiftModifier*/
 		gui->mainWindow()->isShiftPressed() == true &&
-						ke->key() == Qt::Key_Insert )
+						ke->key() == Qt::Key_Enter )
```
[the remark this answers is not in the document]
tresf: Or you can make two shortcuts with an `||`.

src/gui/editors/SongEditor.cpp
```diff
@@ -318,7 +318,14 @@ void SongEditor::keyPressEvent( QKeyEvent * ke )
...
 		m_song->insertBar();
-	}
+    #ifdef Q_OS_MACOS
+        if( /*_ke->modifiers() & Qt::ShiftModifier*/
+           gui->mainWindow()->isShiftPressed() == true &&
+                            ke->key() == Qt::Key_Enter || ke->key() == Qt::Key_Return )
+        {
+            m_song->insertBar();
+        }
+    #endif
```
tresf: 1. I think you accidentally removed a curly bracket here.
2. We use [LMMS_BUILD_APPLE](https://github.com/LMMS/lmms/blob/stable-1.2/src/lmmsconfig.h.in#L4) as our precompiler directive for MacOS.
3. You should group the `||` logic with parens to be explicit.
4. @LMMS/developers are there any objections to just removing the precompiler and doing `( ke->key() == Qt::Key_Insert || Qt::Key_Enter || ke->key() == Qt::Key_Return)`  for all platforms?

lookitsnicholas: Yeah, I noticed that I accidentally removed the curly bracket - was planning on just creating a new PR.
We should probably test Return on macOS platforms before universally changing it, tho - if it doesn't work, Shift+D and Shift+I would be good options to fulfill the same purpose (as discussed previously).

lukas-w: > there any objections to just removing the precompiler and doing[…] for all platforms?

No objections, sounds like a good idea. It's simpler and avoids confusion for people who switch platforms.

PhysSong:
Please use tabs for indentations instead of spaces. This is our coding convention.

lookitsnicholas:
tabs were used, Xcode must encode them differently or smth

tresf:
> tabs were used, Xcode must encode them differently or smth

Each editor has its own preference when you hit the `tab` key.  Even Qt Creator -- our preferred editor -- uses spaces by default, so this isn't uncommon.

src/gui/editors/SongEditor.cpp
PhysSong: No objections, too.

zonkmachine: No objections.

lookitsnicholas: Should we proceed with integrating this change into the master branch, then?

zonkmachine: Yes. At this point this will be going into master. @tresf

lookitsnicholas: A thought just crossed my mind... has anyone actually *tested* these particular key combos on macOS? I'm currently running a build that @tresf graciously compiled for me, but it uses Shift+Command to insert bars instead of the designated key combos suggested here.

tresf: I've since started using a Mac keyboard and the lack of delete shortcut is a huge hindrance.  Any small shortcut changes that fix an OS-specific bug are fine in stable.  It's unfortunate that the OP has abandoned the PR so if the best we can do is land it in master, oh well.

lookitsnicholas: haha, I'm still here - also, Delete is Fn+Shift+Delete (pretty sure that we talked about this in the Discord chat); this PR just needs testing, which I can probably later today/tomorrow (wasn't able to setup my dev environment prior to macOS Mojave's GA)

tresf:
@lookitsnicholas I'd love to see this in `stable-1.2`.  I found a problem with compilation on Mojave and patched it here: #4767. Let me know if this makes it easier to test this feature.

PhysSong:
Superseded by #4851.

PR #4531, opened 2018-08-09
Add sample track recording support

PhysSong:
This is the continuation of #3947 which has done by @Reflexe. Due to the merge conflict and some more reason, @lukas-w have rebased the original commits and pushed a new branch [`feature/recording`](https://github.com/LMMS/lmms/tree/feature/recording).
You can contribute to this work by pushing commits to `feature/recording` or opening a pull request against this branch.
Thanks to @Reflexe for the work, @lukas-w for rebasing commits and many reviewers!

## Bugs
- [x] TripleOscillator preset previews show errors and don't stop due to nonexistent `samples/empty.wav`
- [ ] `w2_invsinehalf.flac` has a furry tail in BitInvader due to resampling(maybe an issue with Graph?)
- [ ] Visualization gets weird when resizing tracks while recording

## Enhancements

- [ ] (optional) Add more backends
  - [ ] ALSA
  - [ ] CoreAudio
  - [ ] Sndio
  - [ ] libsoundio
- [x] (optional) Refactor
- [x] Decide when to resample data
- [ ] Optimize


Reflexe:
About the first one, I'm not sure how to do it. I think that the best way would be to just have an `empty.wav`. However, the current way of loading audio files returns empty data as a sign of failure. 

About the second bug (re-sampling), would it be better to just drop automatic resampling (resampling higher to match the current output device) and stick with resampling on-the-fly?

 About the third, I think I should implement an attribute for every track that "locks" it from the user. Track that's being recorded or manipulated in some way should not disappear (removed), resized or moved to another track, copied and so on.

PhysSong:
> would it be better to just drop automatic resampling (resampling higher to match the current output device) and stick with resampling on-the-fly?

That's what we used to do. Resampling also makes some people can think LMMS hangs when exporting.
> I think that focusing on one backend would help us debug and make sure we release a stable code.

Since this pull request targets to the development branch(unusual name `master` though), I think we don't need to concern about stability issues now.

PhysSong:
Reminder: 18a7f2829db1335b536ad15b23d5719acd759204 and 88dca5732e74e89315e553a0b37c9db3841406a3 should be dropped once #4586 gets merged.

Reflexe:
Do we want to merge it? I don't want this work to go to waste.

zonkmachine:
> (All of the bugs has been fixed)

Great! So what's left to do? Adding more backends and optimization?

Reflexe:
> Great! So what's left to do? Adding more backends and optimization?
IIRC, I worked a lot to optimize the recording and adding more backends is not that useful for now (Alsa is very different that Pulse and modern APIs). 
The only thing left to do is maybe make the arming button more friendly (maybe a led, like the mute button).

PhysSong:
I'll rebase commits against current `master` to produce newer test binaries.

PhysSong:
You can find newest test binaries here: https://github.com/PhysSong/lmms/releases/tag/record

Reflexe:
@PhysSong Great job with `SampleBuffer`! Do you think we need more refactoring on other parts of this PR?

include/MemoryManager.h
```diff
@@ -65,6 +68,18 @@ struct MmAllocator
...
+{
+	return true;
+}
+
+template< class T1, class T2 >
+constexpr bool operator!=( const MmAllocator<T1>&, const MmAllocator<T2>&)
+{
+	return false;
+}
+
```
Reflexe: Do we need that?

include/SampleBuffer.h
```diff
@@ -168,19 +184,17 @@ class LMMS_EXPORT SampleBuffer : public QObject, public sharedObject
 
 	inline f_cnt_t frames() const
 	{
-		return m_frames;
+		dataReadLock();
```
Reflexe: I think I should use the RAII pattern in here too.

include/SampleBuffer.h
```diff
@@ -201,102 +215,162 @@ class LMMS_EXPORT SampleBuffer : public QObject, public sharedObject
...
 
 	static QString tryToMakeRelative( const QString & _file );
 	static QString tryToMakeAbsolute(const QString & file);
 
+	/**
+	 * @brief Try to add data to the buffer,
+	 * @param begin	Beginning of an InputIterator.
+	 * @param end	End of an InputIterator.
+	 * @param syncWithMixer	Should we call requestChangeInModel?
+	 * @return
```
Reflexe: What does it returns? (success)

include/SampleBuffer.h
```diff
@@ -201,102 +215,162 @@ class LMMS_EXPORT SampleBuffer : public QObject, public sharedObject
...
+	 */
+	bool tryAddData(const DataVector &vector, sample_rate_t sampleRate, bool syncWithMixer=true);
+
+	/**
+	 * @brief Add data to the buffer,
+	 * @param begin	Beginning of an InputIterator.
+	 * @param end	End of an InputIterator.
+	 * @param syncWithMixer	Should we call requestChangeInModel?
+	 * @return
```
Reflexe: What does it returns? (success)

include/SampleBuffer.h
```diff
@@ -201,102 +215,162 @@ class LMMS_EXPORT SampleBuffer : public QObject, public sharedObject
...
+	 * @param syncWithMixer	Should we call requestChangeInModel?
+	 */
+	void resetData(DataVector &&newData, sample_rate_t dataSampleRate, bool syncWithMixer=true);
+
+	/**
+	 * @brief A non-blocking version of resetData.
+	 * @param newData	mm, that's the new data.
+	 * @param dataSampleRate	Sample rate for @a newData.
+	 * @param syncWithMixer	Should we call requestChangeInModel?
+	 * @return
```
Reflexe: What does it returns? (success)

include/SampleBuffer.h
```diff
@@ -201,102 +215,162 @@ class LMMS_EXPORT SampleBuffer : public QObject, public sharedObject
...
+		m_sampleRate = _rate;
+	}
+
+	static sample_rate_t mixerSampleRate();
+
+	// HACK: libsamplerate < 0.1.8 doesn't get read-only variables
+	//	     as const. It has been fixed in 0.1.9 but has not been
+	//		 shipped for some distributions.
+	//		 This function just returns a variable that should have
+	//		 been `const` as non-const to bypass using 0.1.9.
```
Reflexe: make this comment a little bit more understandable

plugins/audio_file_processor/audio_file_processor.cpp
```diff
@@ -343,7 +343,12 @@ void audioFileProcessor::setAudioFile( const QString & _audio_file,
 
 void audioFileProcessor::reverseModelChanged( void )
 {
-	m_sampleBuffer.setReversed( m_reverseModel.value() );
+	if (m_reverseModel.value () != m_isCurrentlyReversed) {
```
Reflexe: it `loadSettings` calls this functions, `m_isCurrentlyReversed` won't be initialized (at least, not for the loaded buffer; it'll be initialized to `m_reverseModel`'s current value) on line 269.

include/MemoryManager.h
Reflexe: Yes we do https://en.cppreference.com/w/cpp/memory/allocator/operator_cmp

PhysSong:
@Reflexe How is your work going on? Do you need some help?

plugins/audio_file_processor/audio_file_processor.cpp
Reflexe: I still don't know how to solve that. Maybe just a `isLoading` attribute. :\

Reflexe:
Closed in favour of the rebased and refactored version at #4994

PR #4536, opened 2018-08-12
Recording -> SampleBuffer -> TripleOscillator: Support loading empty files and add samples/empty.wav  

Reflexe:
Since a few TripleOscillator presets uses the non-existing `samples/empty.wav`, I've decided to use the simplest way and just add this file. 

However, the current implementation of SampleBuffer treated empty data as an error. So this PR should take care of that too.


karmux:
This PR fixes empty.wav problem with TripleOscillator presets.

Wouldn't it have been simpler to just change `userwavefileX="samples/empty.wav"` to `userwavefileX=""` in project files? :slightly_smiling_face:

Reflexe:
It would. But i think that this is the safest way to ensure backward
compatibility with older presets and avoid changing a big amount of presets
which could lead to more bugs.

PhysSong:
> Wouldn't it have been simpler to just change `userwavefileX="samples/empty.wav"` to `userwavefileX=""` in preset files? 🙂

I think we can use `DataFile::upgrade`. BTW, what if a user has non-empty `samples/empty.wav?` Should we ignore that case?

Reflexe:
@PhysSong  Could you merge it to the main recording pull request?

PhysSong:
@Reflexe I'm not sure how to handle the `empty.wav` issue, but the rest of this PR looks good. Sorry for the late response.

Reflexe:
@PhysSong I think that this is the best and natural way to do it. that would also make sure user's presets would not get invalidated.

PhysSong:
@Reflexe Cherry-picked most commits to #4531. Thank you for your works! :+1:

PR #4538, opened 2018-08-13
Fix Freeboy noise channel playback for srw = 1.

BonsaiDen:
In order for `GB_apu` to correctly generate samples when the LSFR width is
set to `7`, the trigger write to `0xff23` must happen after all other
writes.

### DIFF
--- a/plugins/papu/papu_instrument.cpp
+++ b/plugins/papu/papu_instrument.cpp
@@ -276,9 +276,6 @@ void papuInstrument::playNote( NotePlayHandle * _n,
 		data += m_ch4SweepStepLengthModel.value();
 		papu->write_register( 0xff21, data );
 
-		//channel 4 init
-		papu->write_register( 0xff23, 128 );
-
 		_n->m_pluginData = papu;
 	}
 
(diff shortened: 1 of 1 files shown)


BonsaiDen:
Oops, seems like some debug include got left in there. Will be removing that later today.

SecondFlight:
@PhysSong This looks like a pretty simple fix. Do you want to merge this into 1.2 since it's technically not a feature or defer it to 1.3?

@BonsaiDen Could you give a bit more information at a higher level as to what issue this PR solves?

BonsaiDen:
@SecondFlight Sure thing.

If you set up channel 4 with the `SRW` value of 15 you get the expected noisy output:

noise_srw15

However, if you switch the `SRW` value to 7 you get the following output (basically just two clicks):

noise_srw7

What would be expected instead is a similar waveform but with a "faster" sounding noise.

The issues arises from the way the APU registers on the GameBoy must be accessed. Writing the enable bit (1000_0000) to `FF23` causes the noise oscillator to start outputting samples and writing the other registers (including the one for the `SRW` value) during the time the oscillator is active has undefined behaviour.

This is what causes the "silent" output above. Moving the output enable write so it occurs after the noise channel registers have all been set up (as in the PR) fixes this issue :)

PhysSong:
Read the CPU's manual and looks good. The only concerns are possible minor compatibility issues.
I think we can merge this into `stable-1.2`. Any objections?

Issue #4461, opened 2018-07-05
LMMS shows only a part of the NI Reaktor 6 interface


jibin1573:
I am using LMMS 1.2.0 RC6 (64bit) on Windows 10 version 1803. I loaded the free Native Instruments Realtor 6 (version 6.2.2) into LMMS (through vestige) , and part of the interface was missing. This happens only for the "Lazerbass.ens' patch on Realtor 6.

jibin1573:
Yes! That is what is happening. The VST works fine but chopped-off part is
inaccessible.

DomClark:
Reproduced; sounds like #4087 too. I have a fix in development.

tresf:
> #4087 

Marking as duplicate as the "cutoff" issue is already being addressed there per @DomClark's comments.  @DomClark please feel free to mark as duplicate or consolidate as you find 'em.

Issue #4087, opened 2017-12-31
Kontakt is buggy in LMMS (Explanation in Description)


EtornaZ:
When I open LMMS (latest beta version, RC5,) I can have the Kontakt player running in order to use my sound pack, Lunaris. However, the problem is when I reopen LMMS. Kontakt will load up Lunaris, but all of the samples are turned off, and sometimes (randomly) Kontakt will open with only a small section of the VST UI, like so. 

![image](https://user-images.githubusercontent.com/33297131/34463609-37ca0c68-ee2f-11e7-8015-4574fabb7cae.png)

This bug can happen upon reopening the project or loading the VST into LMMS. It's completely random in this case also.
Some addition information is that Lunaris can run on 4 GB of RAM at the minimum. My computer only has 4 GB of RAM, so it may be related to that.

lukas-w:
@EtornaZ Did it work with LMMS 1.1.3? Can you change the "embed method" in the settings, restart LMMS, then try again and report with your results?

Spekular:
Closing and consolidating into #5433

hibobmaster:
Kontakt will be opened with a small part of UI even adjust the scale of the VeSTige.
Version: 1.22 stable and 1.3.0-alpha.1
![image](https://user-images.githubusercontent.com/32976627/102064834-c50a3700-3e32-11eb-96fe-ec8df0f770dc.png)

musikBear:
We can add:
Using the wrench subpage for _automation - dial-array_, only open an array with _numbered_ dials, and non of them seams to actually connect to anything.
(An issue that also is true for ZynAddSubFX-VST2-version)

PR #4540, opened 2018-08-14
VST fixes again

### LINKED ISSUE #4117: RC5 inWin10x64 retarding system & many CMD in taskmanager
LMMS RC5 doesn't crush unlike the previous RCs but retarding system In Win10x64 on i3/4GbRAM very hard! CPU usage reaches 90%. If expand LMMS branch In taskmanager may to see there are many running CMDLine processes. Is it right?

### LINKED ISSUE #4306: VST Issue: TX16Wx causes complete UI lockup when attempting to launch file picker

### LINKED ISSUE #4325: Limiter No.6 causes deadlock when changing latency mode while processing

### LINKED ISSUE #4479: Detuning at higher sample rates
I'm using Windows 10 (version 1803)
LMMS 1.2.0 RC6

When I export songs at a sample rate higher than 44100hz, there is an increase in the pitch of certain VST instruments. I have found out some VST instruments which get detuned:

1.MaxSynth's Bass Module (Free): Effect: Pitch shifts upwards
Cause: Happened when I was exporting wav @ 24bit with sample rate at 96000hz.
(No problems were found when I exported it at 44100hz @24bit.)

2. Helm:
Effect: Pitch shifts upwards
Cause: Happened when I was exporting wav @ 24bit with sample rate at 96000hz.
(No problems were found when I exported it at 44100hz @24bit.)

3. DSK Overture:
Effect: Pitch shifts upwards
Cause: Happened when I was exporting wav @ 24bit with sample rate at 96000hz.
(No problems were found when I exported it at 44100hz @24bit.)

4. Orchestral Strings One by Sound Magic:
Effect: Pitch shifts upwards
Cause: Happened when I was exporting wav @ 24bit with sample rate at 96000hz.
(No problems were found when I exported it at 44100hz @24bit.)

5.  Piano One by Sound Magic:
Effect: Pitch shifts upwards
Cause: Happened when I was exporting wav @ 24bit with sample rate at 96000hz.
(No problems were found when I exported it at 44100hz @24bit.)

7.  Bitsonic TB-303:
Effect: Pitch goes up one octave
Cause: Happened when I was exporting wav @ 24bit with sample rate at 96000hz.
(No problems were found when I exported it at 44100hz @24bit.)

8.  T-Force Alpha Plus by Mastrcode music:
Effect: Pitch shifts upwards
Cause: Happened when I was exporting wav @ 24bit with sample rate at 96000hz.
(No problems were found when I exported it at 44100hz @24bit.)

I'll report other VSTs, when I get time.

### LINKED ISSUE #4498: VST-host sync doesnt work
When you use the AcidBox-VSTi in "host sync"-mode, it starts and keeps playing since the plugin is loaded. LMMS doesnt control the play/stop (but it should). You can try this with the demo-version:

https://www.audioblast.me/acidbox.html

With other DAWs this works like excepted.

I talked to AudioBlast, the maker of AcidBox and they wrote to me the following lines:

"in fact, LMMS (or Vestige) doesn't transmit the host "transport bar" flags : the play / stop / loop etc.. are never sent to the plugin.
If you have access to the source code of LMMS/Vestige, search for kVstTransportPlaying and kVstTransportChanged ! "

Tested with LMMS 1.2.0 RC6 on Windows 10 und Windows 8.1.

### LINKED ISSUE #4500: Free Distortion VST Temper resets values when project is re-opened
I've recently installed a great distortion VST called Temper. My problem with it is that whenever I close a project with a Temper effect in it and re-open the project, the effect has reset back to its default values. This can be fixed by connecting every value to an automation track (7 values total), but that's rather annoying in cases where you're not automating. 
I'm on a Windows 10 computer using the latest beta version of LMMS (1.2.0 rc6)

You can download Temper here: https://www.pluginboutique.com/product/2-Effects/30-Distortion/4378-Temper

DomClark:
Most remarks are in commit messages this time.

e5cc1a1 fixes #4479 
4c1d617 fixes #4325
a4e1a0b fixes #4498
8e6ee6f fixes #4306 (see PR #4366)
3af9e26 fixes #4500

b9a8186 will fix window resizing when a plugin is not embedded, but won't work for embedded plugins (things like #4087, #4461); the machinery required to fix this is outside the scope of 1.2 I believe.

I think edf2365 fixes #4117, but it's not entirely clear to me what the point of that issue is; slow VSTs aren't necessarily our problem, but the "many CMD in taskmanager" part is addressed by this commit. @lukas-w requested on Discord that this be implemented differently on the `master` branch using `add_executable`.


SecondFlight:
❤️

PhysSong:
Great work! At first glance, I can't find any possible regressions.
> 24c403d, 993803d and 01cf0ee fix #4306 (see PR #4366)

I think you may squash and reauthor three COM/OLE commits before merging if you want.
> @lukas-w requested on Discord that this be implemented differently on the master branch using add_executable.

It will/should be handled when syncing branches. I guess what @lukas-w means is to use `target_link_libraries` like this:
https://github.com/LMMS/lmms/blob/c44bc4eb7f802c7337ee7f272b8cd28cf2eb6c83/plugins/vst_base/RemoteVstPlugin/CMakeLists.txt#L42

PhysSong:
> the machinery required to fix this is outside the scope of 1.2 I believe.

@DomClark Could you explain that briefly?

DomClark:
> Could you explain that briefly?

Updating the size of an embedded plugin requires sending a message from the remote plugin back to the host LMMS instance. LMMS isn't currently well equipped to receive messages back from plugins - messages are only processed when the plugin is first started, or when waiting for reply messages. In the current state, the window resize message would most likely be picked up on the processing thread while waiting for processing to complete, which isn't desirable. Even then it would only be received if processing were being done, so if, say, a VST effect didn't have any input at the time and the "process effects without input" setting were turned off, the GUI might not resize itself until audio input came again, which is also obviously undesirable.
Periodically processing messages on the main thread is insufficient as a fix as it doesn't stop messages from being picked up elsewhere where they aren't wanted. One could periodically poll for size changes, sending an 'update me' message and waiting for a reply with a new size, but this would unnecessarily lock the plugin, potentially reducing performance and introducing another spot from which deadlock could arise.
The ideal solution here would be to redo the messaging system so that LMMS is always listening for messages from the remote plugin and dispatching them to the appropriate places. This way any window size updates could be applied as soon as possible, on the main thread, and without holding up processing of anything else. Other advantages here would be that multiple messages that require replies can be waited on in parallel - currently these lock the plugin from sending the initial message until the reply is received, otherwise two of these running at once could eat each other's reply messages and cause deadlock - meaning less delay on the processing thread. There wouldn't be any more problems with the main thread processing events, triggering a slot and trying to lock the plugin mutex twice as it wouldn't need to hold the mutex while processing events.
Such an implementation I think would take a while to write and perfect, and in the short term may cause more trouble than it's worth.
I realise this wasn't exactly brief as requested, and possibly a little over-optimistic about what could be done, but I hope it's clear enough.

PhysSong:
> The ideal solution here would be to redo the messaging system so that LMMS is always listening for messages from the remote plugin and dispatching them to the appropriate places.

It sounds good. I think we should separate real-time(audio processing) and non-real-time(UI, etc.) messages when/if we do so.

Here's just an idea:
- Create a watcher thread like `ProcessWatcher` which checks for messages from plugin periodically.
- Message consumers interact with the watcher instead of the plugin itself. They should be able to trigger checking for messages manually when needed.
- When received UI messages, post them using queued connections.
- In other cases, store them in queues.

PhysSong:
Passing `WIN32` into `add_executable` specifies the Windows subsystem. CMake will add `-mwindows` for MinGW and `/subsystem:windows` for MSVC.
MSVC looks for `WinMain` by default when `/subsystem:windows` is specified. `/entry:mainCRTStartup` fixes it, but I have no idea which approach is better.
@DomClark Please let me know what you prefer so I can handle it when syncing branches.

PhysSong:
Merging since @DomClark said okay. I want to know @lukas-w's opinion on the MSVC stuff though.

PhysSong:
@lukas-w MinGW allows using `main` with `-mwindows`.

lukas-w:
I didn't know that. Not sure what the right approach is here. Doesn't matter too much I guess.

Issue #1421, opened 2014-12-07
Navigating and inserting notes in piano roll using keyboard only


grzesiek-galezowski:
Hi!

Just wanted to ask for a small feature that is in Jeskola Buzz and that I am very used to. It would be very nice to be able to put specific notes on specific positions using only keyboard in piano roll.

In buzz, there is a feature that lets me "navigate" my position on the track using keyboard arrows and when I press a key on my keyboard, like 'z', it would automatically leave corresponsing note on this position. In lmms, when I want to leave a note on a specific position, it seems I either have to use the mouse (and point each place) in draw mode or recording (forces me to get timing right or fix things later).

Sti2nd:
Information: There are no shortcuts reserved for this feature, as it probably will use a lot of them, and thus should be a mode.

sorath72:
I've been working with tracking starting with fast tracker 2 then mad tracker, then currently using psycle tracker and sunvox for music composition. I just wish the LMMS would turn more into a fully functional and usable DAW for everyone. Not just FL Studio or other DAWs interfaces PROs out there.

tresf:
> I've been working with tracking starting with fast tracker 2 then mad tracker, then currently using psycle tracker and sunvox for music composition. I just wish the LMMS would turn more into a fully functional and usable DAW for everyone.

We all do, which is why we're learning the code and fixing it.  This stuff is a lot of work and we need help! :beer:

Redsandro:
I've been ~~looking into~~ _thinking about_ this a couple of times ever since in 2010 someone on `#lmms` told me how to recompile parts of the software. I wanted to break the LMMS piano roll dialog and turn it into a tracker. For personal use, and because it would attract the entire population that was born in the 80s to LMMS.

This keyboard-based step recording sounds like the next best thing. Having to use the mouse what made me step away from 'modern' music apps.

On an unrelated note; I do think that mouse events make it easier to port to Android though. On a touch screen, click (touch) events become fun again! FL Studio for Android managed to make me create some tunes again. :smile:

Sti2nd:
> I'm gonna subscribe to notifications

Hey, no need to subscribe to notifcations if you code it :-D !! hint, hint

zonkmachine:
Implemented in https://github.com/LMMS/lmms/pull/4544. Closing.

PR #4544, opened 2018-08-19
Step Recording feature

Mister-Lemon:
Step Recording feature adds a new type of recording in which the user record in steps resolution. 
(Addresses #1421) 
e.g.:
step_recording_wip2

### **PR is fully functional! **

**Behaviour description:**
- Toggle step-recording mode using the dedicated icon
- (This mode is mutually exclusive with other recoding modes (record/record accompany))
- Step-Recording while song is playing is allowed (and fun! :) ) 
- When start recording, the start recoding-position will be set where the timeline curser points (quantized backwards using PianoRoll's current quantization)
- Step length is determined by PianoRoll's current note-length (can be changed dynamically during step-recording)
- The record-position can be moved forward/backward using right/left keys
step_recording_wip4

- When notes are pressed on keyboard/midi-device, they will be added temporarily ("recorded") with a length of a step. while still pressed, user can adjust the length by steps resolution using the arrow keys (e.g. moving right once will make the note's length 2-steps, another right press will make the length 3-steps etc.)
step_recording_wip5

- When all pressed-keys are released, the actual recording happen and the notes are added. 
- If the user press multiple notes, and release some of them for some time which indicates it is intentional i.e. he didn't want to do a full release to record the step but rather just change what will be recorded (I set the "intentional release threshold" to 70 milliseconds) - these note will be removed from current step-recording. e.g.
step_recording_wip3
- Added notes are not quantized, making the addition simpler and WYSIWYG
- Similiarly to adding notes using mouse clicks, an undo-checkpoint is added per added step and not for the whole recording as in other record modes.

**Still missing:**
- [ ] Icons (I made temporary ones, waiting for suggestions...)

**Done:**
- [X] Focus on start position when start recording 
- [x] Auto-scroll when record position becomes off-screen
- [x] On start step-recording : show message to user about using right and left arrows to move recording position

Let me know what you think,
Thanks!

**update1**:
I modified the code so that editing is allowed during step-recording.
now, the currently recorded step is not subjected to modifications. e.g. deleting using mouse right-click:
step_recording_wip6

**update2**:
- Now there is no limition with adding new notes using mouse while step-recording. 
In such case, the addition will work but the note will not play during the addition.
This behaviour is aligned with what happens when user add note during other recording modes 
- I added usage hint that will show on the bottom of the main window when toggling step recording on:
hint

**update3**:
**Feature is fully functional**.  I added the last part - auto scroll to current step-recording cursor position.
(known limitation: only works if position is within pattern's current length. this is the most common use case so I think it is fine as-is. working around this limitation would require (ugly) changes in auto-scroll implementation and does not seems justified for such corner case)


Mister-Lemon:
Hi, 
please wait with code reviews for now(in case someone planned to do it) as I am refactoring large portion of StepRecorder.cpp and plan to upload the changes soon.

My aim for the upcoming update is to remove the editing restriction while step-recording.

Feedback regarding how the feature behaves and also gui suggestions(icon, piano-roll appearance etc.) is still welcome :)

src/core/StepRecorder.cpp
```diff
@@ -0,0 +1,465 @@
...
+	m_stepRecorderWidget(stepRecorderWidget)
+{
+ 	m_stepRecorderWidget.hide();	
+
+#if defined(DEBUG_CREATE_WIN32_CONSOLE) && defined(_WIN32)
+	//create win32 console and attach it for output 
+	if (AttachConsole(ATTACH_PARENT_PROCESS) || AllocConsole())
+	{
+		freopen("CONOUT$", "w", stdout);
+		freopen("CONOUT$", "w", stderr);
```
PhysSong: Though it should not be a part of this PR, `AttachConsole` + `freopen` workaround may be useful. :+1:

PhysSong: You can wrap debug prints with `LMMS_DEBUG` preprocessor definition which is defined only in debug builds. I also suggest using [Qt's functions](http://doc.qt.io/qt-5/qdebug.html) instead of plain `printf`.
For the Windows-specific console code, I'd say it should be in a separate pull request. It's totally unrelated to this feature.

Mister-Lemon: 1. how do you use LMMS_DEBUG ? (where it is defined? is it triggered by some build flag?)
2. I tried to use QT's print functions before using printf and it did not work for some reason (did not output to console. I couldn't figure why)
3. I am not sure I understand, are you suggesting I should have two onging PRs on the same file?
how can see debug prints in the meanwhile? I need something I can work with and debug the component I work on... can you please provide some alternative?

PhysSong: > how do you use `LMMS_DEBUG`? (where it is defined? is it triggered by some build flag?)

It's defined in the root `CMakeLists.txt`. The macro is defined in debug builds.
> I tried to use QT's print functions before using printf and it did not work

Could you let me know your OS, terminal and Qt version?
> I am not sure I understand, are you suggesting I should have two onging PRs on the same file?

I think you should move the console code to somewhere else and open a new PR.

Mister-Lemon: >It's defined in the root CMakeLists.txt. The macro is defined in debug builds.

👍

>Could you let me know your OS, terminal and Qt version?

Win10 and QT5. (is terminal question revelant to win?)

>I think you should move the console code to somewhere else and open a new PR.

I will move it, but probably won't open PR for it but rather keep it locally.

PhysSong:
Please make sure you've used only tabs for indentation when you've implemented everything! :smiley:

Mister-Lemon:
Hi, I would like to get feedback on the following:
1. usage hint: 
hint

2. Icons: On: image  Off: image

Mister-Lemon:
### **Feature is now fully functional!** :)

PhysSong:
@Mister-Lemon Thanks for this work! I'll test and review this in a few days.

src/gui/editors/PianoRoll.cpp
```diff
@@ -613,55 +622,56 @@ void PianoRoll::setCurrentPattern( Pattern* newPattern )
...
+	}
+
 	// set new data
 	m_pattern = newPattern;
 	m_currentPosition = 0;
 	m_currentNote = NULL;
 	m_startKey = INITIAL_START_KEY;
 
-	if( ! hasValidPattern() )
+	if(hasValidPattern())
```
PhysSong: We used to use `if`-`return` pattern to reduce indentation. I think it'll be better if you revert changes from here and add `m_stepRecorder.setCurrentPattern(newPattern);` before this `if` statement if the change does not break anything.

src/gui/widgets/StepRecorderWidget.cpp
```diff
@@ -0,0 +1,156 @@
...
+ */
+
+#include "StepRecorderWidget.h"
+#include "TextFloat.h"
+#include "embed.h"
+
+StepRecorderWidget::StepRecorderWidget( 
+		QWidget * parent, 
+		const int ppt,    
+		const int margin_top,
```
PhysSong: It's not an explicit convention, but we prefer `camelCase` to `snake_case`.

src/gui/widgets/StepRecorderWidget.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+void StepRecorderWidget::setEndPosition(MidiTime pos)
+{
+	m_curStepEndPos = pos;	
+	emit positionChanged( m_curStepEndPos );
+}
+
+void StepRecorderWidget::showHint()
+{
+	TextFloat::displayMessage( tr( "Hint" ),
```
PhysSong: According to our coding convention, you shouldn't put whitespaces in parentheses in new files. You can put them in old files for consistency though.

include/PianoRoll.h
```diff
@@ -261,7 +271,7 @@ protected slots:
 	void autoScroll(const MidiTime & t );
 
 	MidiTime newNoteLen() const;
-
+
```
PhysSong: Please remove trailing whitespaces which you've added.

include/StepRecorderWidget.h
```diff
@@ -0,0 +1,92 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this program (see COPYING); if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA 02110-1301 USA.
+ *
+ */
+#ifndef _STEP_RECOREDER_WIDGET
```
PhysSong: You need to remove `_` at the beginning.

src/core/StepRecorder.cpp
```diff
@@ -0,0 +1,366 @@
...
+
+	m_pattern = newPattern;
+}
+
+void StepRecorder::removeNotesReleasedForTooLong()
+{
+	int nextTimout = INT_MAX;
+	bool notesRemoved = false;
+
+	 QMutableVectorIterator<StepNote*> itr(m_curStepNotes);
```
PhysSong: Remove a space after a tab.

include/StepRecorderWidget.h
```diff
@@ -0,0 +1,92 @@
...
+#include <QPainter>
+
+class StepRecorderWidget : public QWidget
+{
+	Q_OBJECT
+	
+public:
+	StepRecorderWidget(
+		QWidget * parent, 
+		const int _ppt,
```
PhysSong: In function parameters as well.

include/StepRecorder.h
```diff
@@ -0,0 +1,143 @@
...
+	QVector<Note*> getCurStepNotes();
+
+	bool isRecording() const
+	{
+		return m_isRecording;
+	}
+
+	QColor curStepNoteColor() const
+	{ 
+		return QColor(245,3,139); // radiant pink
```
PhysSong: I think we can move this to stylesheets later.

src/gui/widgets/StepRecorderWidget.cpp
```diff
@@ -0,0 +1,155 @@
...
+		const int marginBottom,
+		const int marginLeft,
+		const int marginRight) : 
+	QWidget(parent),
+	m_marginTop(marginTop),
+	m_marginBottom(marginBottom),
+	m_marginLeft(marginLeft),
+	m_marginRight(marginRight)
+{
+	const QColor baseColor =  QColor(255, 0, 0);// QColor(204, 163, 0); // Orange
```
PhysSong: Here as well.

include/StepRecorder.h
```diff
@@ -0,0 +1,143 @@
...
+class StepRecorder : public QObject
+{
+	Q_OBJECT
+
+	public:
+	StepRecorder(PianoRoll& pianoRoll, StepRecorderWidget& stepRecorderWidget);
+
+	void initialize();
+	void start(const MidiTime& currentPosition,const MidiTime& stepLength);
+	void stop();
```
PhysSong: Some more hidden trailing whitespaces here. You may want to find some more occurrences before merging this.
Anyway, you don't need to worry about this for now. :smiley:

PhysSong:
```
QObject::connect: No such slot PianoRollWindow::patternRenamed() in /home/travis/build/LMMS/lmms/src/gui/editors/PianoRoll.cpp:4382
QObject::connect: No such slot PianoRollWindow::patternRenamed() in /home/travis/build/LMMS/lmms/src/gui/editors/PianoRoll.cpp:4383
```
You'll also need to change remaining `patternRenamed` to `updateAfterPatternChange`.

include/StepRecorder.h
Mister-Lemon: fixed

PhysSong:
I found that I can step-record patterns while playing it. Is it intended?

Mister-Lemon:
Yes, see bullet 3 in the description above

PhysSong:
I missed that, thanks!
I also found some bugs.
## Bug 1
- Start step recording and hold a key
- Press the right arrow key and then press another key
- Release two keys simultaneously

Then two notes will have different lengths. I'm not sure it's a bug, but it looks like a bug to me.

## Bug 2
- Hold a key and go back a few steps
- Press new keys and go forward

It will behave somewhat weird.

Mister-Lemon:
Thanks for testing! I will look into it

Mister-Lemon:
@PhysSong , thanks for reviewing and testing the feature!
I fixed the bugs you found, can you please verify the fix?

PhysSong:
Will do that in a few days. :)

PhysSong:
> I am dropping support for this PR.

That's okay. If no one takes this PR, I will.
BTW, why did you decide like this?

Mister-Lemon:
Hi, 
It just became a burden having it in 'open' state for so long.
Thank you @PhysSong and good luck!

PhysSong:
Okay. Hope you'll come with new works later. :)

qnebra:
> Hi,
> It just became a burden having it in 'open' state for so long.

Normal in this repository. Good PRs waiting endlessly, their devs lose all hope and enthusiasm and move on to other things.

zonkmachine:
> I fixed the bugs you found, can you please verify the fix?

@PhysSong 
Bug 1. I can verify both the issue and the fix. :+1: 
Bug 2. I don't see the issue in the first place. By 'going back a few steps' I assume you mean by pressing the left arrow? Nothing strikes me as odd here. I keep on testing though.

This PR has conflicts since https://github.com/LMMS/lmms/pull/4575 was merged. I've rebased it against master (b68c5ee5b56c7f0194a21458f879e78fd9dd4225), fixed the conflict, and intend to push this together with some whitespace fixes.

zonkmachine:
> * When start recording, the start recoding-position will be set where the timeline curser points (quantized backwards using PianoRoll's current quantization)

Which can be a bit confusing because turning step recording on while playing is allowed and the start recording position is then a bit random unless you're dead on the beat. Maybe in the case where the pattern/song is playing, the recording position should be relative to the start of the pattern. Position 0.

zonkmachine:
> Maybe in the case where the pattern/song is playing, the recording position should be relative to the start of the pattern. Position 0.

https://github.com/LMMS/lmms/blob/d035e43d4511390b4ce655e9f2f7774043a7a123/src/gui/editors/PianoRoll.cpp#L3674-L3677


Tested. Seem to  work.
```cpp
	if(hasValidPattern())
	{
		if( Engine::getSong()->isPlaying() )
		{
			m_stepRecorder.start( 0, newNoteLen());
		}
		else
		{
			m_stepRecorder.start(
				Engine::getSong()->getPlayPos(
					Song::Mode_PlayPattern), newNoteLen());
		}
	}
```

zonkmachine:
^OK. Pushed that. The original PR is backed up by the way.

CaBachhav:
A very very important function is to be added to this Step sequencer , if it is added the LMMS will become better than FLstudio.that is in old school mod trackers and latest OPENMPT tracker , number of "steps seek" to be assigned before note is put. Ex. If we put 4 step or N step , the step sequencer cursur jumps forward to 4 steps or N steps. this N can be anything from 1 to 32 .

zonkmachine:
I suggest we merge (my fix included) unless anyone else wants to test/review this some more.

src/core/StepRecorder.cpp
```diff
@@ -0,0 +1,366 @@
...
+			}
+		}
+	}
+
+	if(notesRemoved)
+	{
+		m_pianoRoll.update();
+	}
+
+	if(nextTimout != INT_MAX)
```
justnope: here too.

src/core/StepRecorder.cpp
```diff
@@ -0,0 +1,366 @@
...
+	{
+		dismissStep();
+	}
+
+	m_pattern = newPattern;
+}
+
+void StepRecorder::removeNotesReleasedForTooLong()
+{
+	int nextTimout = INT_MAX;
```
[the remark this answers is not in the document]
zonkmachine: Nitpicked...

CaBachhav:
Anybody working on step Curser jump Function ? Which is very useful in old school Amiga type Mod Trackers, fast tracker etc. Denoted as ADD function.

zonkmachine:
> Anybody working on step Curser jump Function ?

I don't think so. Please open an issue for this if you like:
https://github.com/LMMS/lmms/issues/new

TheBricktop:
so where is it ?

Monospace-V:
> so where is it ?


This feature may not be available in stable 1.2.2. You can access it if you download the nightly version of LMMS from the site https://lmms.io/download . It should also be present in the alpha version.

TheBricktop:
Thank You.

PR #4545, opened 2018-08-20
Don't try to clear null buffer in NotePlayHandle::play

### LINKED ISSUE #4537: Rare crash on tempo automation when stacking and/or arpeggio is on
Reported by **`@FairPlay137-TTS#8972`** on Discord. An example project that I made:
[4536.zip](https://github.com/LMMS/lmms/files/2280772/4536.zip)

Here's the backtrace I've got:
```gdb
(gdb) bt
#0  __memset_avx2_unaligned_erms ()
    at ../sysdeps/x86_64/multiarch/memset-vec-unaligned-erms.S:186
#1  0x000000000050e791 in NotePlayHandle::play(float (*) [2]) ()
```
It seems like `NotePlayHandle::play` calls `memset` when `_working_buffer` is `nullptr`.
https://github.com/LMMS/lmms/blob/2ecc8bb4fafd112cbe9205b5c78ba969646400c9/src/core/NotePlayHandle.cpp#L244-L247

## The reason for the crash
- Stacking and/or arpeggio makes `m_usesBuffer` of the master note `false`.
- `NotePlayHandle::resize` changes `m_totalFramesPlayed` on tempo change. SA #3775.
- In some cases, `m_totalFramesPlayed` becomes 0 even though the note is being played.
- Thus, there can be master notes with `m_totalFramesPlayed == 0`.
- For those notes, `NotePlayHandle::play` calls `memset` with `nullptr`.

## Possible solutions
- Drop the `memset` call. It should work fine thanks to 6fc4577f102e5bb60cc246a426fb9ed8b4ac3be9.
- Fixing #3775 may also fix this, see e.g. https://github.com/LMMS/lmms/issues/3775#issuecomment-327450110.
- Don't use `m_totalFramesPlayed` to check if the note was played before.

PhysSong:
This is a workaround for rare crashes when changing tempo while playing notes with stacking and/or arpeggio.
When playing the master note, `_working_buffer` is null. Tempo change causes false positive in the check because `NotePlayHandle::resize` changes `m_totalFramesPlayed`.
Thanks to 6fc4577f102e5bb60cc246a426fb9ed8b4ac3be9, we can safely drop the memset call.

Fixes #4537.

### DIFF
--- a/src/core/NotePlayHandle.cpp
+++ b/src/core/NotePlayHandle.cpp
@@ -238,13 +238,6 @@ void NotePlayHandle::play( sampleFrame * _working_buffer )
 	// decreasing release of an instrument-track while the note is active
 	if( framesLeft() > 0 )
 	{
-		// clear offset frames if we're at the first period
-		// skip for single-streamed instruments, because in their case NPH::play() could be called from an IPH without a buffer argument
-		// ... also, they don't actually render the sound in NPH's, which is an even better reason to skip...
-		if( m_totalFramesPlayed == 0 && ! ( m_instrumentTrack->instrument()->flags() & Instrument::IsSingleStreamed ) )
-		{
-			memset( _working_buffer, 0, sizeof( sampleFrame ) * offset() );
-		}
 		// play note!
 		m_instrumentTrack->playNote( this, _working_buffer );
 	}


PhysSong:
As I mentioned in #4537, we still have #3775 after fixing this crash. #3775 means we're resizing notes in a wrong way, but it's not really critical. So I'll leave it for 1.3 unless it causes serious issues in normal use-cases.

PR #4547, opened 2018-08-21
Fix decimal separator handling when loading real numbers

### LINKED ISSUE #4442: RC6 plays at least one group of zasfx presets different from previous ver.
OS  : winXP sp3
LMMS RC6

**Issue description:**
In zasfx we have the Cormi collection
In that collection we have a group named Trillo<>
These FX-presets replayed as randomised beeps -as in a alien cp environment -i guess..
In RC6 this is changed to a continuous sine-wave. The variation over time is gone.

**Recreate         :**
You need need an earlier RC , to hear the correct output -I used RC4. Would be interesting to see if RC5 does it correct or not.

PhysSong:
- Remove broken `DataFile::LocaleHelper`.
When saving `float` values into `QDomElement`, Qt uses `QString::setNum`. As of Qt 4, the function always uses the C locale, so we don't need to worry about decimal separators when saving files.
The problem is Qt5 doesn't provide a way to handle both period-separated(`1.2`) and comma-separated(`1,2`) real numbers in one function call because `QString::toFloat` always use the C locale only. `QLocale::toFloat` always tries its specific locale only.
- Add new `LocaleHelper.h` and use it
To solve the problem, I created a function which tries both period and comma. Using it will work on any platforms.

Fixes #4442.

### DIFF
--- a/include/DataFile.h
+++ b/include/DataFile.h
@@ -84,25 +84,6 @@ public:
 		return m_type;
 	}
 
-	// small helper class for adjusting application's locale settings
-	// when loading or saving floating point values rendered to strings
-	class LocaleHelper
-	{
-	public:
-		enum Modes
-		{
-			ModeLoad,
-			ModeSave,
-			ModeCount
-		};
-		typedef Modes Mode;
-
-		LocaleHelper( Mode mode );
-		~LocaleHelper();
-
-	};
-
-
 private:
 	static Type type( const QString& typeName );
 	static QString typeName( Type type );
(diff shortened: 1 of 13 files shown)


JohannesLorenz:
I could code-read and test this branch. Let me know when it's ready.

JohannesLorenz:
@PhysSong Sounds good. Also maybe add doxygen-like comments so one can guess what the functions do by just looking at the header.

PhysSong:
There are some functions defined in other header files. I think my helper functions are not too large and it's fine to leave them inline.
> Also maybe add doxygen-like comments so one can guess what the functions do by just looking at the header.

Updated the description at the beginning.

PR #4567, opened 2018-08-29
Restructure help, remove minus-minus from actions

JohannesLorenz:
Please check if the restructuring is OK and if you're OK with removing the minus-minus-signs from the "actions".

No functional differences except removing minus-minus.


PhysSong:
You can update the man page(`doc/lmms.1`) as well. I can't find any reason to remove `--`, but I don't disagree with it. Just need to wait for opinions.

tresf:
It seems quite strange to remove such a common usage -- such as `-d = dump` in the middle of a stable release.  Is there a reason why we don't deprecate these off instead?

JohannesLorenz:
OK, fixed that. If there are no more comments, I'll squash this in two days, add the bash completion and merge it to lmms/stable-1.2.

PR #4569, opened 2018-08-29
Add automation support for Carla.

CYBERDEViLNL:
This will read Carla's slots and add knobs to the plugin widget.

### DIFF
--- a/plugins/carlabase/carla.cpp
+++ b/plugins/carlabase/carla.cpp
@@ -33,6 +33,8 @@
 #include "InstrumentPlayHandle.h"
 #include "InstrumentTrack.h"
 #include "Mixer.h"
+#include "Knob.h"
+#include "EffectControls.h"
 
 #include <QApplication>
 #include <QFileDialog>
@@ -40,6 +42,11 @@
 #include <QPushButton>
 #include <QTimerEvent>
 #include <QVBoxLayout>
+#include <QHBoxLayout>
+#include <QSizePolicy>
+#include <QScrollArea>
+#include <QtDebug>
+#include <QString>
 
 #include <cstring>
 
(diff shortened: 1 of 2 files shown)


SecondFlight:
Hi! Thanks for this PR. We're on feature freeze for 1.2 - sorry this isn't more obvious!

Could you open a new PR with the same commits added to the master branch?

plugins/carlabase/carla.cpp
```diff
@@ -299,14 +306,69 @@ void CarlaInstrument::saveSettings(QDomDocument& doc, QDomElement& parent)
...
+
+				// Create new model for the knob.
+				floatModels.push_back(new FloatModel(param_value,0.0f,1.0f,0.001f,this,name));
+
+				// Load settings into model.
+				QString idStr = CARLA_SETTING_PREFIX + QString::number(i);
+				floatModels[i]->loadSettings( settingsElem, idStr );
+
+				// Connect to signal dataChanged to knobChanged function.
+				connect( floatModels[i], &FloatModel::dataChanged, [=]() { knobModelChanged(i); });
```
PhysSong: Connecting slots to functors is not supported in Qt4. You have two choices:
- Use normal `SIGNAL()`/`SLOT()` and use [`sender()`](http://doc.qt.io/archives/qt-4.8/qobject.html#sender) to detect the sender(what we already do for VSTs)
- Keep this syntax and switch to `master` branch which has dropped the support for Qt4

plugins/carlabase/carla.cpp
```diff
@@ -33,13 +33,20 @@
...
 #include <QApplication>
 #include <QFileDialog>
 #include <QFileInfo>
 #include <QPushButton>
 #include <QTimerEvent>
 #include <QVBoxLayout>
+#include <QHBoxLayout>
+#include <QSizePolicy>
+#include <QScrollArea>
+#include <QtDebug>
```
PhysSong: I recommend you including `<QDebug>` instead which does the same thing.

plugins/carlabase/carla.cpp
```diff
@@ -33,13 +33,20 @@
 #include "InstrumentPlayHandle.h"
 #include "InstrumentTrack.h"
 #include "Mixer.h"
+#include "Knob.h"
+#include "EffectControls.h"
```
PhysSong: I think you will not need to include `EffectControls.h` here as well as in `carla.h`.

plugins/carlabase/carla.cpp
PhysSong: I also think you should use `Qt::DirectConnection` because automated changes will be handled in the UI thread if you use the normal connection type. It may prevent automation working correctly when exporting.

plugins/carlabase/carla.cpp
```diff
@@ -495,10 +570,36 @@ CarlaInstrumentView::CarlaInstrumentView(CarlaInstrument* const instrument, QWid
...
+	m_refreshParamsButton = new QPushButton( tr( "" ), this );
+	m_refreshParamsButton->setIcon( embed::getIconPixmap( "reload" ) );
+	m_refreshParamsButton->setWhatsThis(
+                tr( "Click here to reload the Carla parameter knobs." ) );
+    QSizePolicy sizePolicy(QSizePolicy::Maximum, QSizePolicy::Fixed);
+	sizePolicy.setHorizontalStretch(0);
+	sizePolicy.setVerticalStretch(0);
+	sizePolicy.setHeightForWidth(m_refreshParamsButton->sizePolicy().hasHeightForWidth());
+	m_refreshParamsButton->setSizePolicy(sizePolicy);
+	m_refreshParamsButton->setCheckable( false );
```
PhysSong: Mixing old and new coding styles doesn't look good. I suggest putting whitespaces either anywhere or nowhere.

plugins/carlabase/carla.h
```diff
@@ -96,16 +114,33 @@ class CarlaInstrumentView : public InstrumentView
...
     void uiClosed();
+    void refreshButtons();
+    void onRefreshButton();
 
 private:
     virtual void modelChanged();
     virtual void timerEvent(QTimerEvent*);
 
+    void addKnob(FloatModel& knobModel);
+	void clearKnobs();
```
PhysSong: Our coding convention says you should use tabs, but most part of this file uses spaces. I don't know which one will be better, but please don't mix them like this. You can change indentation after some discussions.

CYBERDEViLNL:
Thanks for the response guys.

@SecondFlight ; Sure thing my fault.
@tresf ; My git knowlage is very limited :-P
@PhysSong ; Thanks for the feedback I made the changes in my new pull request.

New pull request in master here:  https://github.com/LMMS/lmms/pull/4571

PR #4570, opened 2018-08-29
Update UI 60 times per second.

### LINKED ISSUE #4562: Optimize LMMS To 60FPS
Hello, so LMMS currently runs at a cap of 30FPS for work flow but it would be nice if it was Optimize at 60FPS, that way we could experience a buttery smooth feel and look in LMMS.

karmux:
Fixes #4562.

### DIFF
--- a/plugins/Eq/EqFader.h
+++ b/plugins/Eq/EqFader.h
@@ -80,7 +80,7 @@ private slots:
 	{
 		const float opl = getPeak_L();
 		const float opr = getPeak_R();
-		const float fall_off = 1.2;
+		const float fallOff = 1.07;
 		if( *m_lPeak > opl )
 		{
 			setPeak_L( *m_lPeak );
@@ -88,7 +88,7 @@ private slots:
 		}
 		else
 		{
-			setPeak_L( opl/fall_off );
+			setPeak_L( opl/fallOff );
 		}
 
 		if( *m_rPeak > opr )
@@ -98,7 +98,7 @@ private slots:
 		}
 		else
 		{
-			setPeak_R( opr/fall_off );
+			setPeak_R( opr/fallOff );
 		}
 		update();
 	}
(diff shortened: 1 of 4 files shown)


PhysSong:
Merge?

SecondFlight:
I remember doing this on my own fork, and I noticed that the EQ visualizer seemed to behave somewhat differently. Has anybody else noticed this? I didn't see it as a problem, but I might have missed something.

PhysSong:
> the EQ visualizer seemed to behave somewhat differently.

I think the falloff in `EqFader::updateVuMeters` should be adjusted to reflect the refresh rate change.

PhysSong:
Also in `EqSpectrumView::paintEvent`. The exact value should be about `1.06266`, but I suggest slightly larger value around to bring some visual effects.

karmux:
I will look into it in the weekend.

karmux:
Adjusted also falloff of mixer faders (`FxMixerView::updateFaders`). Used value `1.07`.

SecondFlight:
I can't think of any other possible issues with this. The fact that it's not actually 60 fps still feels wrong to me (62.5 fps, assuming the timer is accurate), but I doubt this will cause any actual issues.

Merge?

PhysSong:
I'm not sure if this one is okay for `stable-1.2`, but the change itself looks good to me.

SecondFlight:
I would hate to defer something so simple and beneficial to 1.3, but we are on feature freeze. It's your call, @PhysSong, I won't complain if we have to defer.

karmux:
It's more like a fine-tuning than feature.

PhysSong:
Merging into `stable-1.2` because we can safely revert it if it causes any issues. :tractor:

musikBear:
Any win Test binaries?

PhysSong:
@musikBear This will be part of RC7. So please wait a few days. :)

musikBear:
@PhysSong Just want to let you know, that RC7 runs fine on my old Tube-monitor 👍

DomClark:
Should `TimeLineWidget` be updated too? This looks like a 20 fps timer to me: https://github.com/LMMS/lmms/blob/d8fb07ff52ee87a290ad7e69d0851cfb1c20d261/src/gui/TimeLineWidget.cpp#L96

PhysSong:
@DomClark Nice catch!

PR #4571, opened 2018-08-30
Added Carla automation support.

CYBERDEViLNL:
Thanks PhysSong for the review in the previous PR.
* Converted all indentation to tabs.
* Added whitespaces (coding style).
* Removed EffectControls.h from carla.cpp
* Removed QtDebug from carla.cpp, it wasn't needed anymore.
* Signal/Slots Qt4 support.

### DIFF
--- a/plugins/carlabase/carla.cpp
+++ b/plugins/carlabase/carla.cpp
@@ -33,6 +33,7 @@
 #include "InstrumentPlayHandle.h"
 #include "InstrumentTrack.h"
 #include "Mixer.h"
+#include "Knob.h"
 
 #include <QApplication>
 #include <QFileDialog>
@@ -40,6 +41,10 @@
 #include <QPushButton>
 #include <QTimerEvent>
 #include <QVBoxLayout>
+#include <QHBoxLayout>
+#include <QSizePolicy>
+#include <QScrollArea>
+#include <QString>
 
 #include <cstring>
 
@@ -63,47 +68,47 @@ static const double ticksPerBeat = 48.0;
 
 static uint32_t host_get_buffer_size(NativeHostHandle handle)
 {
-    return handlePtr->handleGetBufferSize();
+	return handlePtr->handleGetBufferSize();
 }
 
 static double host_get_sample_rate(NativeHostHandle handle)
 {
-    return handlePtr->handleGetSampleRate();
+	return handlePtr->handleGetSampleRate();
 }
 
 static bool host_is_offline(NativeHostHandle handle)
 {
-    return handlePtr->handleIsOffline();
+	return handlePtr->handleIsOffline();
 }
 
 static const NativeTimeInfo* host_get_time_info(NativeHostHandle handle)
 {
-    return handlePtr->handleGetTimeInfo();
+	return handlePtr->handleGetTimeInfo();
 }
 
 static bool host_write_midi_event(NativeHostHandle, const NativeMidiEvent*)
 {
-    return false; // unsupported?
+	return false; // unsupported?
 }
 
 static void host_ui_parameter_changed(NativeHostHandle handle, uint32_t index, float value)
 {
-    handlePtr->handleUiParameterChanged(index, value);
+	handlePtr->handleUiParameterChanged(index, value);
 }
 
 static void host_ui_custom_data_changed(NativeHostHandle handle, const char* key, const char* value)
 {
-    // unused
+	// unused
 }
 
 static void host_ui_closed(NativeHostHandle handle)
 {
-    handlePtr->handleUiClosed();
+	handlePtr->handleUiClosed();
 }
 
 static intptr_t host_dispatcher(NativeHostHandle handle, NativeHostDispatcherOpcode opcode, int32_t index, intptr_t value, void* ptr, float opt)
 {
-    return handlePtr->handleDispatcher(opcode, index, value, ptr, opt);
+	return handlePtr->handleDispatcher(opcode, index, value, ptr, opt);
 }
 
 #undef handlePtr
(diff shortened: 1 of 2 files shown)


PhysSong:
> Converted all indentation to tabs.

Sorry, What I mean was keeping consistency in new code, not changing old code.

CYBERDEViLNL:
@PhysSong ; Ok do you want me to change them to spaces again? (I’ll prefer the tabs BTW.)

I'm also working on putting the parameters in a subwindow since some Carla plugins have a lot of parameters. I have working code for this, but it needs to be cleaned before I put it online.

Other thoughts of adding are giving the user the ability to colour a knob so it stands out and it easy recognisable. And adding a search/filter bar that will result in only showing knobs with matching names.

PhysSong:
It seems like you messed up the ZynAddSubFX submodule while merging branches, but it's okay. I can fix that before merging if you want.

PhysSong:
Since `master` branch doesn't support Qt4, you can revert knob connection as you did in https://github.com/LMMS/lmms/pull/4569#pullrequestreview-150840198 . Functor-based connections look cleaner than the current way to me.

CYBERDEViLNL:
@PhysSong

About the messed up ZynAddSubFX submodule: Hmm I didn’t touch anything but `carla.h` and `carla.cpp`, did I have to remove the submodules before merging with the upstream? And yes please fix it before merging if you would.

I made some changes (see commit log) in response to your review, thanks again :-)

plugins/carlabase/carla.cpp
```diff
@@ -286,14 +311,97 @@ void CarlaInstrument::saveSettings(QDomDocument& doc, QDomElement& parent)
...
+			QString idStr = CARLA_SETTING_PREFIX + QString::number(i);
+			floatModels[i]->setObjectName(QString::number(i)); // use this as reference for parameter index.
+			if (init)
+			{
+				floatModels[i]->loadSettings(settingsElem, idStr);
+			}
+
+			// Connect to signal dataChanged to knobChanged function.
+			//connect(floatModels[i], SIGNAL(dataChanged()), this, SLOT(knobModelChanged()));
+			connect(floatModels[i], &FloatModel::dataChanged, [=]() {knobModelChanged(i);});
```
PhysSong: Could you add `Qt::DirectConnection` as the last argument, as well as #4692?
Also, I can't find a reason not to delete the old connection code.

plugins/carlabase/carla.cpp
```diff
@@ -286,14 +311,97 @@ void CarlaInstrument::saveSettings(QDomDocument& doc, QDomElement& parent)
...
+
+			completerData.push_back(name);
+
+			// current_value, min, max, steps
+			floatModels.push_back(new FloatModel(param_value, paramInfo->ranges.min,
+					paramInfo->ranges.max, paramInfo->ranges.step, this, name));
+
+			// Load settings into model.
+			QString idStr = CARLA_SETTING_PREFIX + QString::number(i);
+			floatModels[i]->setObjectName(QString::number(i)); // use this as reference for parameter index.
```
PhysSong: I think this is not needed anymore because you're using functor-based connections.

plugins/carlabase/carla.cpp
```diff
@@ -286,14 +311,97 @@ void CarlaInstrument::saveSettings(QDomDocument& doc, QDomElement& parent)
...
+	//Clear the list
+	floatModels.clear();
+}
+
+void CarlaInstrument::knobModelChanged(uint32_t index)
+{
+	//FloatModel *senderFloatModel = qobject_cast<FloatModel *>(sender());
+	if (fDescriptor->set_parameter_value != nullptr){
+		//fDescriptor->set_parameter_value(fHandle, senderFloatModel->objectName().toInt(), senderFloatModel->value());
+		fDescriptor->set_parameter_value(fHandle, floatModels[index]->objectName().toInt(), floatModels[index]->value());
```
PhysSong: I believe `floatModels[index]->objectName().toInt()` can be changed to `index`. Is it correct?

plugins/carlabase/carla.cpp
CYBERDEViLNL: You are right.

plugins/carlabase/carla.cpp
CYBERDEViLNL: Right again :+1:

plugins/carlabase/carla.cpp
CYBERDEViLNL: Done, thanks for the review again :-)

falkTX:
Just saw this, note that carla will have a way to handle automation in plugin mode "soon".
The base work for it is already done

CYBERDEViLNL:
Hello @falkTX , can you tell what that does mean for this PR? And the done base work, is it on the Carla git repo so we can follow it?

falkTX:
Only the code that allows the actual automation of plugin parameters is in place, so nothing to see yet.
Previously the rack and patchbay engine modes would misbehave on variable process cycles, that hosts use for sample-accurate automation.
WIth that fixed, there is no more blocker to exposing some parameters in carla plugin versions.

My idea is to have a static number of parameters (say, 100), that the user can map to a specific plugin parameter. I have to expose the 100 parameters in the plugin code, and then design a new UI for this.

GingkathFox:
Lemmie just revive for a sec:
I ran this on MacOS Mojave, and i like it. The issues i have with it is that i can only automate the first vst in the list, or the instrument vst. It for some reason cant grab the automatable knobs in a new carla entity(?). It only got them in a existing project.

CYBERDEViLNL:
@GingkathFox 
> i can only automate the first vst in the list, or the instrument vst.

I’m not sure if it’s currently possible with Carla yet, but I will look into it.

> It for some reason cant grab the automatable knobs in a new carla entity(?). It only got them in a existing project.

I don’t quite understand, what do you mean by grab then? Grab and drag to a automation track? And this only happens when you add a new Carla Patchbay/Rack instance? What about after saving and reopening the project, does it work then?

If you can tell me what version of Carla you use, I will try to reproduce on Linux, I don’t have access to a Mac.

CYBERDEViLNL:
I have found a way to get a list of loaded plugins by Carla but this is kinda tricky. Also I haven’t found a way to get dispatched by Carla on the changes of plugins. A lot of questions come to mind on how to handle such events, some examples:

### Unique identifier.
The `CarlaBackend::CarlaPlugin->getUniqueId()` method isn’t reliable since most plugins I tested didn’t supply one. We can work around that by using `getName()` wich should be unique but it isn’t perfect. What about this:

1. User removes a plugin.
2. User adds another plugin.
3. User renames that plugin to the same name as the previously removed plugin.

### Index (Id)
We can access a plugin by their `Id` through `CarlaBackend::CarlaEngine→getPlugin(i)`. But if the order of the plugins changes inside Carla so will their Id’s.


So I think it’s best to leave it this way and wait till the `static number of parameters` feature @falkTX mentioned (https://github.com/LMMS/lmms/pull/4571#issuecomment-438995134) is implemented, then change this code to work with that.

Also I’ve found two bugs in the current code. will fix that asap.
* **Bug 1**: If there are knobs loaded, remove all plugins inside Carla, then reload knobs and the old knobs won’t be removed. It should be empty.
* **Bug 2**: If there are no knobs, pushing `refresh values` will crash LMMS.

Oh and @GingkathFox, just to be sure did you press the `refresh parameter knobs` button after you loaded a plugin inside Carla? If not with current code you should, like I said I haven’t found a way yet to get dispatched on plugin changes inside Carla so till then you have to press the button manually after loading a plugin inside Carla.

GingkathFox:
@CYBERDEViLNL i did, still didnt work.

CYBERDEViLNL:
With Carla 2.0 we need to refresh the params and their values manual. With the `develop` branch of Carla it can be done automaticly, it however only works for the parameters of the first plugin inside Carla, in the future Carla should get all parameters until the max parameters is reached. (This is still TODO for Carla.)

What this means for this PR: It can be used if the behaviour of manual refreshing the parameters is acceptable. If it is not then we have to wait until Carla 2.1 is released. I'm in for waiting until the Carla 2.1 release.

I created another branch on my LMMS fork with code that's working for Carla `develop` branch. https://github.com/CYBERDEViLNL/lmms/tree/master_carla

Please let me know what we want to do here so I can update this PR or create a new one.

PhysSong:
> Anyway, a proper implementation of this will need changing the code in carla.
> Otherwise we just hack around it, and then have to maintain it as well, which is not nice.

@falkTX How is that going on? We have Carla 2.2.0 now.

falkTX:
> > Anyway, a proper implementation of this will need changing the code in carla.
> > Otherwise we just hack around it, and then have to maintain it as well, which is not nice.
> 
> @falkTX How is that going on? We have Carla 2.2.0 now.

This is fully supported now, exposed in Carla internal plugin as well as in the Carla VST2 version.

There is a callback on carla plugin API for when a refreshing of the parameter list needs to be done (basically everytime a plugin is added, removed or reordered).
There parameters are even grouped by plugin.

Here is how it looks for the carla-rack plugin inside of carla:
Screenshot_20201210_121706

PhysSong:
Closing in favor of #5846.

PhysSong:
...And I didn't really close this. :joy:

tinkerlevu:
Wait a minute, speaking of 'getting merged', if this issue is moved to #5846, why is this still in https://github.com/LMMS/lmms/projects/11?

PR #4574, opened 2018-09-01
Divide knob step size by 1000 for LADSPA effects, and minor changes to scroll wheel behavior

### LINKED ISSUE #4565: LMMS step size too large for LADSPA effect values
@tresf, @DomClark and I talked about this a bit on the LMMS Discord.

The problem is that LMMS has a set step size that can be restrictive in some cases.  For example, if you load up the Decimator, the Sample Rate can be set to 55.0699 or 110.14, but nothing in between.

In lmms/src/core/LadspaControl.cpp, you can see the current settings for knob step sizes.  Using FLOATING as an example, the step size is 1/8000th of the range of the knob for logscale, and 1/800th for anything else.  After decreasing the step size from 1/800th to 1/800000th, everything worked quite well.  The only problem that arose was that using the mouse wheel to change knob values does it by individual steps, so changing values with the mouse wheel would take a horribly long time.  Otherwise, there were no negative effects for the UI, and it would be a wonderful change to implement.

What I'm thinking, is that we massively decrease the step size like I did, and then change things so the mouse wheel, instead of changing the value by single steps, changes the value by a specified fraction of the knob's total range.  This would have the same outcome as giving the mouse wheel a larger step size in comparison other methods of changing knob values (click+drag, double click then type value, etc.), which will have a very small step size for extra precision.  This will allow very precise changes to easily be made to knob values, without any significant changes to UI.

LostRobotMusic:
These changes change the knob value step size for certain knobs to 1/1000th of what it was previously. This makes the step sizes extremely small, allowing a significantly larger amount of precision when choosing knob values. This is necessary for a variety of reasons. For example, before this change, when using a Decimator, the Sample Rate can be set to 55.0699 or 110.14, but nothing in between. That only allows precision of 55.0701Hz, which is not nearly as precise as most would want. This change allows you to be precise within 0.00550701 Hz, which is so precise that I highly doubt anybody would want anything more precise. It is a significant improvement in comparison to where it was before. Of course, that was just an example, this change applies to all knobs that don't require a set step size (like the Type for C* Cabinet, which must stay set to 1 for obvious reasons).

When scrolling the mouse wheel over a knob, it increments the value by 1 step. Since this change massively decreases the step size, it renders use of the mouse wheel quite useless. So, another change has been added. The number of steps the mouse wheel changes the knob by has been changed, here is the formula that fit the best:

`Number of Steps = Knob Value Range / 2000 / Knob Step Size`

I spent about an hour doing nothing but testing out different knobs making sure this change didn't have any negative UI effects, and then left it on there for a day while I just normally made music. I didn't notice anything negative, and actually eventually forgot the change was even on there, which is a good sign.

The decrease in step size allows more precise values to be chosen with all methods of setting a knob value, whether it be clicking and dragging, double clicking and typing the value, and via automation and controllers (which, as a consequence, should cause both to sound better and smoother than before)!

Fixes #4565.

### DIFF
--- a/src/core/LadspaControl.cpp
+++ b/src/core/LadspaControl.cpp
@@ -79,7 +79,7 @@ LadspaControl::LadspaControl( Model * _parent, port_desc_t * _port,
 				( m_port->max - m_port->min )
 				/ ( m_port->name.toUpper() == "GAIN"
 					&& m_port->max == 10.0f ? 4000.0f :
-								( m_port->suggests_logscale ? 8000.0f : 800.0f ) ) );
+								( m_port->suggests_logscale ? 8000000.0f : 800000.0f ) ) );
 			m_knobModel.setInitValue( m_port->def );
 			connect( &m_knobModel, SIGNAL( dataChanged() ),
 						 this, SLOT( knobChanged() ) );
--- a/src/gui/widgets/Knob.cpp
+++ b/src/gui/widgets/Knob.cpp
@@ -701,7 +701,8 @@ void Knob::paintEvent( QPaintEvent * _me )
 void Knob::wheelEvent( QWheelEvent * _we )
 {
 	_we->accept();
-	const int inc = ( _we->delta() > 0 ) ? 1 : -1;
+	const float stepMult = model()->range() / 2000 / model()->step<float>();
+	const int inc = ( ( _we->delta() > 0 ) ? 1 : -1 ) * ( ( stepMult < 1 ) ? 1 : stepMult );
 	model()->incValue( inc );
 
 


DouglasDGI:
I've been using this for the past two weeks, I haven't noticed any problems or undesirable behavior at all!  If other people could test it as well and see if there's anything they personally would change, that would be awesome.

zonkmachine:
> January 2019 edit: Still using it, it works so well I forgot I even made the change xD
> It's very useful. Large step sizes are no longer a problem.

OK. I'll review it next.

JohannesLorenz:
@zonkmachine What exactly do you mean by "next"? :smile:

PhysSong:
I think it's ready to merge if the new values are fine.

DomClark:
I think I've already made this comment somewhere, but I can't see it in this PR or the related issue, so I guess it's lost somewhere in Discord chat history. I'll repeat it here:

This PR is undoubtably an improvement over the current behaviour, but it doesn't address what I would consider to be the fundamental issue here. For `FLOATING` type LADSPA knobs, the `FloatModel` should be continuous, accepting any float value in its range. There are many other knobs in the program that could benefit from this as well, without having their scroll-wheel step changed. What's happened here is that some magic numbers in the code have been replaced with new, bigger, magic numbers, and some more magic numbers have been added elsewhere to balance these out. While bigger is better, it still doesn't reflect the concept it's trying to model, and it's still full of arbitrary magic numbers.

Spekular:
+1 to Dominic's comment, this solution reminds me a bit of [this article](https://tomforsyth1000.github.io/blog.wiki.html).

PhysSong:
I agree that the `FloatModel` can have any values in a continuous interval. @DomClark Could you open an issue regarding this?
> While bigger is better,

@DomClark Does that mean you're not against the change in this PR?

src/core/LadspaControl.cpp
```diff
@@ -79,7 +79,7 @@ LadspaControl::LadspaControl( Model * _parent, port_desc_t * _port,
 				( m_port->max - m_port->min )
 				/ ( m_port->name.toUpper() == "GAIN"
 					&& m_port->max == 10.0f ? 4000.0f :
```
PhysSong: This special case isn't handled. However, do we even need to keep this exception?

DouglasDGI:
I do agree with those points, this PR is much better than the current behavior but is more of a workaround than the permanent solution we should be aiming for.
@DomClark So what would the new behavior be?  Would mouse wheel scrolling and mouse click+dragging obey the step size, while double-click ignores it?

DomClark:
> I agree that the FloatModel can have any values in a continuous interval. @DomClark Could you open an issue regarding this?

Will do.

> @DomClark Does that mean you're not against the change in this PR?

I'm not, no. Knowing this project, we might not see a better fix appear for years, so this should do in the meantime. Unless @DouglasDGI wants to start working on that now.

> @DomClark So what would the new behavior be? Would mouse wheel scrolling and mouse click+dragging obey the step size, while double-click (enter a value) ignores it?

That's the idea, yeah.

JohannesLorenz:
When you specify the issue, please differ between the following:

1. The entered values shall not respect the step size anymore (mentioned above)
2. The mouse wheel shall keep respecting them (mentioned above)
3. Dragging on a knob shall, assumably, be changed to now ignore steps? (See `Knob::setPosition`, I think this could be changed easily).
4. Automation should be changed: It should ignore any steps.

Also, since sometimes (at least (2)) the steps are still respected, IMO they shall be smaller on low values and larger on large values if the scale is logscale. Though I'd vote for adding this into a different issue report.

zonkmachine:
It's an improvement. Merge?

zonkmachine (approved):
lgtm

DomClark:
Finally got round to opening an issue - see #5626.

JohannesLorenz:
Was the open comment from PhysSong solved?

Issue #4578, opened 2018-09-04
3 new minor developer features for LMMS


JohannesLorenz:
I'm on my personal LMMS fork (it basically supports a new plugin for zyn) and would like to extract up to 3 minor features and move them over to LMMS. All features are only interesting for developers and have no "visual" effect:
1. Add `doc/automation.md` , explaining how LMMS' automation basically works (classes and widgets)
2. Add a simple `XPMLoader` class to embed.h, inheriting `PixmapLoader` (it only has an xpm array as member, and a `pixmap()` override which returns it as `QPixmap`)
3. Functionality in src/core/AutomatableModel.cpp to load and save models quoted if they do not match `^[A-Za-z0-9._-]+$` (instead of `<modelname ...>`, they are stored as `<quoted-model quoted-name="modelname" ...>`). For example, this allows "save" saving of OSC string models, like "/volume".

Can you please let me know which I can merge to master? I'm not sad if any/all are rejected, but I'd find it cleaner if this code is being reused by master.

JohannesLorenz:
Sorry, I didn't know that this branch was even pushed in that state. Anyways, for reference:
XPMLoader: https://github.com/JohannesLorenz/lmms/blob/osc-plugin/include/embed.h#L94-L104
AutomatableModel: https://github.com/JohannesLorenz/lmms/blob/osc-plugin/src/core/AutomatableModel.cpp#L94-L229

@tresf For the AutomatableModel, I agree, "model" sounds better. I'll add some lines that find out whether the name of the model was "model", or whether it was quoted. Then it should work.

JohannesLorenz:
Proposal from gi0e5b06 on Discord:
```
quoted-name -> name
quoted-model -> automatablemodel
```
I think that fits even better.

tresf:
> XPMLoader: https://github.com/JohannesLorenz/lmms/blob/osc-plugin/include/embed.h#L94-L104

Forgive my lack of understanding and cosmetic critique here but doesn't C++ allow multiple constructors?  Is there a reason why `PluginPixmapLoader( const char** xpm )` is undesired?

> @tresf For the AutomatableModel, I agree, "model" sounds better. I'll add some lines that find out whether the name of the model was "model", or whether it was quoted. Then it should work.

👍 

> You need to look at the osc-plugin branch: https://github.com/JohannesLorenz/lmms/blob/osc-plugin/doc/automation.md The branch has all those mentioned features, too.

[...]

PR #4588, opened 2018-09-10
Add wiki as clonable submodule

tresf:
Per conversation https://github.com/LMMS/lmms/issues/4578#issuecomment-419643823 so that eventually we can ship a tarball with everything needed to build LMMS -- including the build docs.


PhysSong:
It's a nice idea! BTW, the wiki has been updated since you opened this PR.

Issue #4515, opened 2018-08-01
Carla not detected on openSUSE with lmms-1.2.0-rc6-linux-x86_64.AppImage


geckolinux:
Hi there, I'm running openSUSE Tumbleweed and Leap 15.0 with `carla` and `carla-vst` installed at version `1.9.7+git20171111.5b97ec1e-50.14`. The latest `lmms-1.2.0-rc6-linux-x86_64.AppImage` doesn't detect Carla being installed:

`````````
Cannot load library /tmp/.mount_lmms-1bfmTqI/usr/lib/lmms/libcarlabase.so: (libcarla_standalone2.so: cannot open shared object file: No such file or directory)
Cannot load library /tmp/.mount_lmms-1bfmTqI/usr/lib/lmms/libcarlarack.so: (libcarlabase.so: cannot open shared object file: No such file or directory)
Cannot load library /tmp/.mount_lmms-1bfmTqI/usr/lib/lmms/libcarlapatchbay.so: (libcarlabase.so: cannot open shared object file: No such file or directory)
`````````

Thanks a lot for developing LMMS, it's fantastic!

SecondFlight:
Looks like a regression. Tagging @PhysSong, sounds like this should be looked into before the 1.2 release.

geckolinux:
Thanks @SecondFlight for looking into it.

Bleuzen:
Same on Manjaro. AppImage doesn't seem to detect Carla. Self compiled rc6 does.

PhysSong:
It looks like the bug in `package_linux.sh.in`, but the AppImage worked fine because of an upstream bug, https://github.com/probonopd/linuxdeployqt/issues/289.
Upstream fixed the issue via https://github.com/probonopd/linuxdeployqt/commit/c17e3f0e8a61fa2793491f6942fcfe7266871811, and now our bug is an issue.
![apprun](https://user-images.githubusercontent.com/12591074/43673469-a55f7e88-97fe-11e8-9676-898df9ca4d33.png)

`linuxdeployqt` creates `AppRun` symlink to the executable file. Due to the `linuxdeployqt` bug, it was a link to `lmms`, which is the wrapper script. After the bug was fixed, however, it links to `lmms.real` so the script never runs.

@tresf I think we should remove the problematic symbolic link and manually create desired one. I also wonder why we substitute the `Exec` entry back to `lmms` after running `linuxdeployqt`.

[...]

Issue #4197, opened 2018-02-27
Next LMMS Release and Roadmap


Anonymouqs:
I'm going to start making an LMMS roadmap on the wiki. Like in [GIMP](https://wiki.gimp.org/wiki/Roadmap) What features are we going to have in the release? And when are we going to do that release?

I heard that if you know what your outcomes are, your chances of success increase a thousand-fold.

This issue is made to come to a conclusion on what we are doing in the future. Please do not get this off topic.

tresf:
Is this in response to https://github.com/LMMS/lmms/issues/4182#issuecomment-368941338?  Milestoning marketing efforts and trying to milestone the software features aren't the same.  Let's not bite off more than we can chew here. :)

Anonymouqs:
This is just to document stuff

tresf:
@Anonymouqs GitHub has some built-in triaging and project management utilities as well as milestones.  Maintaining a constantly changing section of the wiki seems to be a bit of an overkill when the functionality is built-into the product, don't you think?  https://help.github.com/articles/about-project-boards/.

Furthermore, you may want to take the leap to [Trello](https://blog.trello.com/github-and-trello-integrate-your-commits).  We can sponsor power-ups if needed for proper execution.

fundamental:
@tresf IMO github issue tracking of milestones and a wiki summary can have very different audiences. Consider "Calf is at v1.2.3, please update to v.2.3.4 for more cowbell" vs. "Update builtin plugins to external versions".

One of them is for individuals familiar with the goings on within LMMS and the other can be for users to take a glance at to know the direction of the project without the details. The former is an internal management tool while the latter can be (if targeted) an external marketing tool.

tresf:
> The former is an internal management tool while the latter can be (if targeted) an external marketing tool.

It was just a recommendation, and one in the direction of progress.  Trello can handle high-level milestones, as can GitHub.

My primary concern is I specifically requested that @Anonymouqs milestoned some of his marketing ideas and now he's promising to make a futuremap.  It's not what we discussed and a quick chat on Discord would have avoided yet another futuremap post.

Here's our 4-year old futuremap.  It's probably a good place to start. 
 Some of the items we've made progress on since it was written. 
 Many we have not.  https://github.com/LMMS/lmms/issues/908

Anonymouqs:
Apologies

musikBear:
my ½ bend € on this:
[...]

Issue #4156, opened 2018-02-05
Build failure when linking carlabase.so: undefined references 


GvMariani:
Still preparing a package of lmms 1.2.0 RC5 for Rosa Linux Fresh R10 (a Mandriva derivative) on x86_64: until a few days ago Rosa did not provide a package for Carla and libsoundio, hence their support was disabled in the cmake command line...
Now I managed to make a package for both things and tried to enable the support for Carla and libsoundio... hitting another (little) roadblock, as shown in the log snippet below:
***************************************
```
[ 26%] Linking CXX shared library ../libcarlabase.so
cd /tmp/abf/rpmbuild/BUILD/lmms-1.2.0-rc5/build/plugins/carlabase && /usr/bin/cmake -E cmake_link_script CMakeFiles/carlabase.dir/link.txt --verbose=1
/usr/bin/c++  -fPIC -fno-exceptions -Wall -Werror=unused-function -Wno-sign-compare -Wno-strict-overflow -Wno-array-bounds -O2 -Wa,--compress-debug-sections -gdwarf-4 -fvar-tracking-assignments -frecord-gcc-switches -Wstrict-aliasing=2 -pipe -Wformat -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -ffat-lto-objects -fno-delete-null-pointer-checks -fstack-protector --param=ssp-buffer-size=4 -fPIC -fPIC -DPIC -DNDEBUG -Wl,--as-needed -Wl,--no-undefined -Wl,-z,now -Wl,-z,relro -Wl,-O1 -Wl,--build-id -Wl,--enable-new-dtags -Wl,--hash-style=gnu -shared -Wl,-soname,libcarlabase.so -o ../libcarlabase.so CMakeFiles/carlabase.dir/carla.cpp.o CMakeFiles/carlabase.dir/moc_carla.cpp.o  -L/usr/lib64/carla -Wl,-rpath,/usr/lib64/carla -lcarla_standalone2 /usr/lib64/libQt5Widgets.so.5.9.4 /usr/lib64/libQt5Xml.so.5.9.4 /usr/lib64/libQt5Gui.so.5.9.4 /usr/lib64/libQt5Core.so.5.9.4 
CMakeFiles/carlabase.dir/carla.cpp.o: In function `CarlaInstrument::CarlaInstrument(InstrumentTrack*, Plugin::Descriptor const*, bool)':
[...]

Issue #3452, opened 2017-03-24
Loading certian instruments in Carla Rack causes a segfault


unfa:
I tried to load synthv1 into Carla Rack (also Carla Pathbay) and LMMS immediately crashes when I do that. Same for drumkv1 and Calf Wavetable.

I also tried loading these with no problems:

Calf Organ, Calf Monosynth, MDA Piano, ZynAddSubFX, amsynth, Wolpertinger, the infamous cellular automaton synth, setBfree, WhySynth DrumSynth, and Noize Mak3r.

I'm using LMMS 1.1.3 from KX Studio repositories.

falkTX:
Calf wavetable is an experimental plugin, crashes with it are to be expected.

synthv1 and drumkv1 crashing sounds like a qt4 vs qt5 thing.
Are you using Debian testing or Ubuntu >= 16.04? If yes, that would explain it.

Since LMMS in KXStudio repos uses qt4, you cannot load qt5 stuff in it.

Working around this is possible, but very tricky.
We'd need to know the current process qt version used in the backend (which can be different if carla is run as LMMS plugin, normal plugin or standalone), then check if a plugin to load has a conflicting qt version...

PhysSong:
@unfa Do you still experience the problem with LMMS 1.2.2 or 1.3 alphas?

PR #4593, opened 2018-09-13
Use new carla-native-plugin library

falkTX:
Also fixes build with latest carla (which will be 2.0 soon)

Should be pushed to stable-1.2 branch too, but there will be conflicts to resolve for that.
@tresf can I let you handle that?


PhysSong:
`stable-1.1` is no longer used, so you should open a PR against `stable-1.2`.
I think @tresf can handle conflicts. If not, I can help you resolve them.

falkTX:
stable-1.1 is still the latest stable release of LMMS available, and it is simply not that that is no longer used - I am using it myself and everyone with kxstudio repos.
Without the patch, the build fails against latest carla.
This is a bugfix, not a new feature. Please merge.

PhysSong:
BTW, won't it make the build fail with old version of Carla?

falkTX:
no, as the library it uses was changed. it no longer links against the full carla standalone lib, but only against the exported plugin version (libcarla_native-plugin.so).
it makes a lot more sense to link against this new object file.

as it is now, stable-1.1 finds carla-standalone and then fails to build.
the patch fixes this, among other things.

PhysSong:
I meant there's no fallback logic for old versions. It looks like a regression to me. Can we try `carla-standalone` if the old version of Carla is detected?
> stable-1.1 finds carla-standalone and then fails to build.

I can't reproduce it. Do you have a build log with the failure?

falkTX:
treat it as a regression if you wish, I feel like now it is doing things proper.
the symbols it uses are all defined in a single header file, which is the one also used when exporting the library.
the old behaviour was hacky, experimenting with getting it to work.
now with carla going stable (2.0), it was time to clear up the hacks and make things work properly.

I dont have the logs with the failure, but the error only happens due to -Werror.
You need latest git of carla (from maximum 2 days ago) to get the new header file and library.

tresf:
I have mixed feelings about this landing on the `stable-1.1` branch.  First of all, I really don't care because the branch is stale, unmaintained and doesn't provide any binaries to our website, so the impact is only downstream and I can't speak on behalf of downstream + Carla.  @falkTX perhaps you can shed some insight on this?

So the impact of merging this to `stable-1.1` is:
 * This would only impact distros that ship 1.2 (which most do) that would pull from `HEAD` of the `stable-1.2` branch.   Does this ever happen?  I can't imagine this would be common, instead they'd pull from the `1.1.3` release.  Do we need to ship a `1.1.4` version?  It would be for Carla only as the Windows and Mac builds are not impacted by this change.
 * What I don't truly understand is how the distros get Carla to begin with.  I'm familiar with the Debian/Ubuntu technique but even then I'm not exactly sure how it's provided as I don't think the major distributions have Carla support baked in.  If this is only to satisfy the KXStudio's build process, then it seems like a sane change.
* @PhysSong's concerns about regression handling seem valid but only if the distribution of these libraries happens separately and I can't speak on behalf of those distros.  The ones that come to mind are Fedora and Arch, which we get a lot of bug reports for -- including Carla bug reports (#4156, #4515, #3452, #3218, #2213, #2212, #2549, #2543, #1444, #1221, #1201, #1199).

What I think would be more responsible of a change -- albeit leaving cruft into the Carla build system -- is for @falkTX to leave the old library around and deprecate it off sanely instead of putting this decision on us at LMMS.  The justification for this is we've [worked with package maintainers that require a version bump to receive a change](https://github.com/LMMS/lmms/issues/4197#issuecomment-415896881) so I'd like to be cognizant of that.

@falkTX thoughts welcome. :)

falkTX:
thanks for the reply.

I understand the situation, I was hoping 1.1 would still be maintained in a way.
it is the current stable branch, and the one I am still using. but it is fine that it is not supported anymore.
In such cases, then **I really recommend taking out -Werror, and that is it.**
Without that flag, carla will keep building and working just fine with the current code.

Most distros don't have Carla afaik.
I know Arch does, I know the maintainer for the packager and we meet in Berlin from time to time, so I can request updates in person :P
Some Fedora and Gentoo audio overlay/repos/external-sources contain it too, but I dont know how those work.
Debian does not have it, neither does Ubuntu.

@tresf could you please merge this patch into your lmms-carla-osx PR?
I think pushing to 1.2 branch right now is a bit pointless, if you will have conflicts with it right away, so you will have to resolve them anyway.

Issue #4534, opened 2018-08-11
Add bash-completion


JohannesLorenz:
This is a linux (and similar) commandline only feature request.

Bash has a default completion, which means it completes every argument as a path name. In order to complete stuff like `--parameter` or options that are not path-related (e.g. numbers) or to specify files ending on some pattern, one has to write a bash completion script. Once done, you can type
```
lmm<tab>--s<tab>4<tab>--r<tab> ~/my-p<tab>
```
to get something like
```
lmms --samplerate 44100 --render ~/my-project.mmz
```

As I wrote many of those scripts recently, I'd like to write one for LMMS now. **Any concerns before I start?**

Also, if I write one, **if a project file is expected, should it automatically add project files from the LMMS working directory to the completion list even if you are in another directory?** E.g. should `lmms --render so<tab>` complete to `lmms --render <lmms working dir>/projects/song1.mmp` automatically? The advantage would be quicker navigation to projects without typing your LMMS working directory path; the disadvantage is that you might add a bloat of files even if you don't want to complete from your LMMS working directory...

Concerning the milestone, **this could even be done for 1.2.0**, if wanted, as it does not affect the LMMS binary.

PhysSong:
> if a project file is expected, should it automatically add project files from the LMMS working directory to the completion list even if you are in another directory?

Could be a problem of taste, but I don't think so. The disadvantage looks greater to me.

JohannesLorenz:
The help has been restructured per #4567 . I'll add a commit introducing the completion (tomorrow?), then this should be finished.

JohannesLorenz:
Ok, I just pushed 9e0ee04 (see above). It's white-listing to only use `/usr` and `/usr/local`. I also removed the shell script, as it would sound very strange if users could write to `/usr/local`, but not `/usr/local/share/completions/bash-completion`.

Somehow, I'm OK with the solution. Don't want to complicate it further.

Reviews are appreciated.

@tresf  Can you please test if installation works, and if it really completes after instalation (e.g. in your container)?

tresf:
Full code available here: https://github.com/LMMS/lmms/compare/stable-1.2...JohannesLorenz:stable-1.2

JohannesLorenz:
Please wait with testing, I just want to add another whitelist case (install to `/usr` and bashcomp in `/etc...`)... 10 minutes...

tresf:
[...]

PR #4604, opened 2018-09-20
Add bash completion (#4534)

JohannesLorenz:
This should now work for `/usr`, `/usr/local`, user installs and AppImages.


doc/bash-completion/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
...
+		# all other paths are user paths
+
+		# Add ../target to bash-completion whitelist for appimages
+		FOREACH(_prefix ${BASHCOMP_WHITELIST_PREFIX})
+		IF("${CMAKE_INSTALL_PREFIX}" MATCHES "^${_prefix}/?$")
+			SET(BASHCOMP_PATH_FINAL "${CMAKE_INSTALL_PREFIX}/share/bash-completion/completions")
+		ENDIF()
+		ENDFOREACH()
+		
+	ELSEIF("${BASHCOMP_PKG_PATH}" STREQUAL "/etc/bash_completion.d")
```
tresf: Can you please explain the purpose of this logic?  If it's `/etc/bash_completion.d`, shouldn't we just blindly use it and white-list it?

JohannesLorenz: I wanted it to behave like the other logic, but don't install it if the install is going to `/usr/local` (imagine they're doing an architekture independent install to `/usr/local`, but an architecture depending one for the bash-completion to `/etc`). It's a contradiction to me, so I change the behavior for `/etc/bash_completion.d`.

doc/bash-completion/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
...
+		ENDIF()
+		ENDFOREACH()
+
+	ENDIF()
+
+	IF("${BASHCOMP_PATH_FINAL}" STREQUAL "")
+	
+		# maybe it's a user install?
+	
+		SET(BASHCOMP_WHITELIST_STARTSWITH "/home/;/root/;/Users/")
```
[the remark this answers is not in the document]
JohannesLorenz: Can you please explain why you first suggested to let the script check for those paths?

doc/bash-completion/CMakeLists.txt
tresf: Sorry, I had made some bad logic suggestions on Discord... let's try this instead.  I think you'll find it covers all common scenarios and all edge-cases.

```cmake
# Detects the best place to install bash completion support
# - Windows is not supported
# - TODO: macOS support can be provided through Homebrew
IF(NOT LMMS_BUILD_WIN32 AND NOT LMMS_BUILD_APPLE)

	# standard install location for most distributions
	SET(BASHCOMP_FALLBACK "/usr/share/bash-completion/completions")

	# check for common portable installation (CMAKE_INSTALL_PREFIX=../target)
	GET_FILENAME_COMPONENT(BINARY_ADJ_INSTALL "${CMAKE_BINARY_DIR}/../target" ABSOLUTE)
	IF("${BINARY_ADJ_INSTALL}" STREQUAL "${CMAKE_INSTALL_PREFIX}")
		SET(SKIP_PKG_CHECK TRUE)
	ENDIF()

	# check for other signs of user-level install
	SET(USER_LEVEL_INSTALL "/home/;/root/;/Users/")
	FOREACH(_startswith ${USER_LEVEL_INSTALL})
		IF("${CMAKE_INSTALL_PREFIX}" MATCHES  "^${_startswith}")
			SET(SKIP_PKG_CHECK TRUE)
			BREAK()
		ENDIF()
	ENDFOREACH()

	IF(SKIP_PKG_CHECK)
		SET(BASHCOMP_PKG_PATH "${CMAKE_INSTALL_PREFIX}${BASHCOMP_FALLBACK}")
	ELSE()
		PKG_CHECK_MODULES(BASH_COMPLETION bash-completion)
		PKG_GET_VARIABLE(BASHCOMP_PKG_PATH bash-completion completionsdir)
	ENDIF()

	# finally, fallback for older operating systems if using root-level install
	IF("${BASHCOMP_PKG_PATH}" STREQUAL "")
		SET(ROOT_LEVEL_INSTALL "/usr/;/usr/local/")
		FOREACH(_startswith ${ROOT_LEVEL_INSTALL})
			IF("${CMAKE_INSTALL_PREFIX}" MATCHES  "^${_startswith}")
				SET(BASHCOMP_PKG_PATH "${BASHCOMP_FALLBACK}")
				BREAK()
			ENDIF()
		ENDFOREACH()
	ENDIF()
	
	IF("${BASHCOMP_PKG_PATH}" STREQUAL "")
		MESSAGE("Unable to determine BASHCOMP_PKG_PATH.  Install manually from completions/lmms if needed.")
	ELSE()
		MESSAGE(STATUS "Bash completions will be installed to ${BASHCOMP_PKG_PATH}")
		INSTALL(FILES lmms DESTINATION "${BASHCOMP_PKG_PATH}" PERMISSIONS OWNER_READ GROUP_READ WORLD_READ)
	ENDIF()
ENDIF()
```

tresf: @devnexen so that we can handle any edge-cases, can you tell us if this logic (at a glance) will work on OpenBSD?

doc/bash-completion/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+	FOREACH(_startswith ${USER_LEVEL_INSTALL})
+		IF("${CMAKE_INSTALL_PREFIX}" MATCHES "^${_startswith}")
+			SET(SKIP_PKG_CHECK TRUE)
+			BREAK()
+		ENDIF()
+	ENDFOREACH()
+
+	# determine BASHCOMP_PKG_PATH variable
+	IF(SKIP_PKG_CHECK)
+		SET(BASHCOMP_PKG_PATH "${CMAKE_INSTALL_PREFIX}${BASHCOMP_FALLBACK}")
```
PhysSong: I believe it may give a wrong path because of `/usr`. For example, `~/.local` prefix should result in `~/.local/share/...`, not `~/.local/usr/share/...`.

doc/bash-completion/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+	ELSE()
+		PKG_CHECK_MODULES(BASH_COMPLETION bash-completion)
+		PKG_GET_VARIABLE(BASHCOMP_PKG_PATH bash-completion completionsdir)
+	ENDIF()
+
+	# final fallback for non-user-level installs
+	IF("${BASHCOMP_PKG_PATH}" STREQUAL "")
```
PhysSong: If you determine whether to install the completion script in this way, it will always be installed if `BASHCOMP_PKG_PATH` is defined. For example, `cmake . -DCMAKE_INSTALL_PREFIX=/tmp` gives:
```
-- Bash completions will be installed to /usr/share/bash-completion/completions
```

doc/bash-completion/CMakeLists.txt
tresf: > ~/.local prefix should result in ~/.local/share/

Says who?  What's the difference between installing between these two unrecognized locations?  They're both rubbish to any type of bash completion.  If you want to blindly prefix, then we go back to the "can I write here?" logic and always fallback to a consistent path if we can't.

doc/bash-completion/CMakeLists.txt
tresf: Yeah I discussed this with @JohannesLorenz.  I hadn't considered `/tmp`, we can add that to the user-space whitelist.  There's not too many places that are utilized like this and we forgot about `/tmp`.  Are you ok with simply adding it?

Some may argue that any mountpoint can be valid, but I'm not sure how far down this rabbit hole you want to go.  In the end, I'd like both AppImages and Homebrew builds to be able to get to bash completion, even if they won't be used for some time.

doc/bash-completion/CMakeLists.txt
PhysSong: `~/.local/share/bash-completion/completions` does work by default. `/usr` is usually the install prefix for distributon-provided packages.

doc/bash-completion/CMakeLists.txt
PhysSong: I wasn't talking about a specific path. There are some more spaces that user can write, e.g. `/var/tmp`, some subdirectories of `/media`/`/mnt`(some mount points), but I'm not sure how should we handle them.

doc/bash-completion/CMakeLists.txt
PhysSong: For reference: https://github.com/scop/bash-completion/blob/c2dd6745661d886327bb79591a6c6911f6f5c3f6/bash_completion#L2029-L2033

tresf: @PhysSong it is not our place to take over a user's home directory or default settings via CMake.  CMake `INSTALL` is simply not intended for user-space installs.  It's designed for root-level POSIX installs.  If you think it's common practice to install LMMS to `~/.local/bin/lmms` please provide an example of where people would leverage CMake for deployment.  The validity of this location is irrelevant.

doc/bash-completion/CMakeLists.txt
tresf: Agreed.  @JohannesLorenz' first proposal was to check to see if the directory was writable.  Perhaps that's better overall?

doc/bash-completion/CMakeLists.txt
PhysSong: >  If you think it's common practice to install LMMS to `~/.local/bin/lmms`

I saw someone does it. `pip` also installs some scripts to the same directory: https://www.python.org/dev/peps/pep-0370/
I think it may be useful for users who don't have a root access.

JohannesLorenz: We can add `~/.local` to the whitelist, it's OK for me.

tresf: @PhysSong `pip` is a user-space application by-design.  `pip` is a package management tool.  CMake is not, it's a build tool and the `INSTALL_PREFIX` is not intended to modify the user's environment.  Please don't do this.

tresf: > We can add `~/.local` to the whitelist, it's OK for me.

Why the hell would anyone install LMMS to `~/.local`?

doc/bash-completion/CMakeLists.txt
tresf: > The install paths leading to an install of the completion should be a white-list, e.g. only /usr and /usr/share.

This whitelist is useless if we can't detect a non-root install.

JohannesLorenz: Checking for writable requires the ugly shell script.

@tresf  Why do we need to detect a non-root install? What's wrong with my last commit from yesterday?

doc/bash-completion/CMakeLists.txt
JohannesLorenz: > Why the hell would anyone install LMMS to `~/.local`?

I guess adding this to the whitelist is faster than discussing it :stuck_out_tongue: Let's bring this stupid feature request to an end...

doc/bash-completion/CMakeLists.txt
tresf: > @tresf Why do we need to detect a non-root install? What's wrong with my last commit from yesterday?

Because I want the completions part of the AppImages and eventually available through Homebrew, neither needing `sudo/root`.

doc/bash-completion/CMakeLists.txt
tresf: > I guess adding this to the whitelist is faster than discussing it 😛 Let's bring this stupid feature request to an end...

Funny, but untrue.  It's not faster, it's not practical and it's not wise.  Why would anyone install an application to a hidden directory like `.local`?  The answer: They wouldn't.  Ever.

@PhysSong's talking about a special direction, not unlike the one we're hardcoding for root-level installs.  CMake needs to stay out of the user's `.local`. That's the job for [`appimaged`](https://github.com/AppImage/appimaged), and it's successor [`AppImageLauncher`](https://github.com/TheAssassin/AppImageLauncher).  This is NOT NOT NOT our job, ever with CMake unless CMake starts acting like a user-space deployment helper.  This prefix has no basis with build logic.  It just looks technically plausible.  The only argument that could be made is the manual override, which we can handle without weird whitelisting. :)

JohannesLorenz: Well, I don't mind how you decide here. I'm fine with either decision.

doc/bash-completion/CMakeLists.txt
JohannesLorenz: For AppImages, I think we wanted to use `../target`. For Homebrew, no idea what you are talking about. Probably best if you make another commit we can check...

tresf:
Due to all the whitelisting problems, I propose we go back to a shell script... here's my proposal:

```cmake
# install bash completion support
# - Windows is not supported
# - TODO: macOS support can be provided through Homebrew
IF(NOT LMMS_BUILD_WIN32 AND NOT LMMS_BUILD_APPLE)
	PKG_CHECK_MODULES(BASH_COMPLETION bash-completion)
	PKG_GET_VARIABLE(BASHCOMP_PKG_PATH bash-completion completionsdir)

	IF("${BASHCOMP_PKG_PATH}" STREQUAL "")
		SET(BASHCOMP_PKG_PATH "/usr/share/bash-completion/completions")
	ENDIF()

	# fallback for non-root installs
	SET(BASHCOMP_USER_PATH "${CMAKE_INSTALL_PREFIX}${BASHCOMP_PKG_PATH}")

        # use script to handle permission issues gracefully for non-root installs
	CONFIGURE_FILE("install_bash_completion.sh.in" "install_bash_completion.sh" @ONLY)
	INSTALL(CODE "EXECUTE_PROCESS(COMMAND chmod u+x install_bash_completion.sh WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR} )")
	INSTALL(CODE "EXECUTE_PROCESS(COMMAND install_bash_completion.sh WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR} )")

	MESSAGE(STATUS "Bash completions will be installed to ${BASHCOMP_PKG_PATH} or fallback to ${BASHCOMP_USER_PATH} if unwritable.")
ENDIF()
```

**`install_bash_completion.sh.in`**
#/bin/bash

set -e

BASHCOMP_PKG_PATH="@BASHCOMP_USER_PATH@"
if [ -w "@BASHCOMP_PKG_PATH@" ]; then
	BASHCOMP_PKG_PATH="@BASHCOMP_PKG_PATH@"
fi

echo -e "\nInstalling bash completion..."
mkdir -p "$BASHCOMP_PKG_PATH"
cp "@CMAKE_CURRENT_SOURCE_DIR@/lmms" "$BASHCOMP_PKG_PATH"
chmod ugo+r "$BASHCOMP_PKG_PATH/lmms"
echo -e "Bash completion installed to $BASHCOMP_PKG_PATH"
```

PhysSong:
How about this:
- Add an option to force the install path for the completion script, e.g. `BASHCOMP_INSTALL_PATH`
- If no path specified, install the script into the install prefix for some whitelisted locations

doc/bash-completion/CMakeLists.txt
PhysSong: Back to the leading `/usr`, it shouldn't be there when concatenating paths. I still think `${CMAKE_INSTALL_PREFIX}/share/...` is natural because it becomes `/usr/share/...` and `/usr/local/share/...` for `/usr` and `/usr/local` respectively.

tresf:
> Add an option to force the install path for the completion script, e.g. BASHCOMP_INSTALL_PATH

It's non-standard and a general bad idea, but fine.

> If no path specified, install the script into the install prefix for some whitelisted locations

No more whitelist please, please review the latest proposal.  Less code, more scalable.

doc/bash-completion/CMakeLists.txt
tresf: > Back to the leading /usr, it shouldn't be there when concatenating paths.

This isn't about concatenating paths.  PLease review the latest proposal.  `${CMAKE_INSTALL_PREFIX}` should never be used unless it's a non-root or non-sudo install.  `/usr` is mandatory for the [AppImages to eventually use it](https://github.com/TheAssassin/AppImageLauncher/issues/87#issuecomment-423610462).  As far as I can tell `/usr/share/bash-completion/completions` is a made-up, magic path. Let's keep it as such.  Let's not overthink the prefix because it's non-standard and it's a bad idea.

If you prefer to append `${BASHCOMP_PKG_PATH}` to the end of `${CMAKE_INSTALL_PREFIX}`, I'll take that, but this latest code proposal honors the `pkg-config` value first, hard-coded `/usr/share/bash-completion/completions` second, and if neither of those are writable, it tucks it into whatever strange prefix was used.

Prefix prediction is fine if everything follows that rule, but no evidence has been provided to suggest bash-completion follows these arbitrary prefix rules, please don't propose them, they make this PR much more difficult than it needs to be and it also breaks the (eventual) AppImages.

If you still aren't OK with that, please fix the AppImage path to match the standard as part of this PR, but post-install during the packaging stage.

PhysSong: I'll clarify. We don't install the LMMS binary to `${CMAKE_INSTALL_PREFIX}/bin`, not `${CMAKE_INSTALL_PREFIX}/usr/bin`. I wanted to say that this pass calculation doesn't make sense because the `${BASHCOMP_FALLBACK}` has a leading `/usr` which is needed for absolute paths, not for relative paths.

tresf: I double checked the appimage and the `/usr` is prefixed from the script so that should be fine as `${CMAKE_INSTALL_PREFIX}/share/`.

If that's it, @johaneslorenz can we use the new proposal?

JohannesLorenz: @tresf Without following the conversation, the code below seems to cover most cases I can think of. I can't see any errors, but if you'd make a commit I could run some tests.

Btw, `ugo+r` should be identical to `+r` ?

PhysSong: Yes, `ugo` means everyone.

tresf: > Btw, `ugo+r` should be identical to `+r`?

No, `+r` is identical to `u+r`. `ugo+r` is identical to the cmake's `PERMISSIONS OWNER_READ GROUP_READ WORLD_READ`.

tresf:
Created a new PR against this branch to bring all of this conversation logic into a cmake module. https://github.com/JohannesLorenz/lmms/pull/1.  Feedback appreciated.

tresf:
Awesome.  Now time for testing.  I think Fedora, Ubuntu and Arch should cover our bases.   I can test the first two.

JohannesLorenz:
I made some tests on Manjaro (~Arch). It's mostly OK, I only noticed:
* travis-ci failed
* The shebang line of the install script is just `#!` for me.
* It's a bit weird that a prefix of `/usr/local` says it wants to install to `/usr/share` and only fallback to `/usr/local/share`. Though I know this is part of the design.

tresf:
If pkg-config and cmake doesn't care about the prefix, I'm not sure why we should.  The prefix is just as strange looking when installing to /opt.  It's just a spaghetti code mess if we don't trust the modules provided by the OS.

JohannesLorenz:
Tested and working :+1:
I don't want to propose whitelists again, we'd be spinning in circles. No further changes required from my side.

PhysSong:
Since it's functioning, we can handle some special cases later, e.g. `/usr/local`. :thumbsup:

tresf:
Testing passed on Fedora 27 and Ubuntu 18.04.  Merging.  Thanks @JohannesLorenz and @PhysSong for all of the back-and forth on this.  We've been in PM for about a week trying to tackle this and now we have a cmake module to do it.  Excellent team work!

PR #4605, opened 2018-09-22
Add `AutomatableModel` nodename attribute (#4578)

JohannesLorenz:
* Save `AutomatableModel` nodename in attribute if it must be quoted
* Loading an `AutomatableModel` with name <name> now means it
  - either <name> must be `QDomElement::nodeName()` (as before)
    *and* must not have a `nodename` attribute (new)
  - or must have a `nodename` attribute with value <name>


JohannesLorenz:
Merged on a console, so no need to merge this in the github UI.

PR #4608, opened 2018-09-22
Enable compiled-in XPM pixmaps (#4578)

JohannesLorenz:
Should I merge this to master? It's practical and may help coders in the future, but it's also currently unused code (as it's only inside my osc-plugin branch), so coders have no example about how it must be used.

### DIFF
--- a/include/embed.h
+++ b/include/embed.h
@@ -35,7 +35,17 @@
 namespace embed
 {
 
-QPixmap LMMS_EXPORT getIconPixmap( const QString&  _name, int _w = -1, int _h = -1 );
+/**
+ * Return an image for the icon pixmap cache.
+ *
+ * @param _name Identifier for the pixmap. If it is not in the icon pixmap
+ *   cache, it will be loaded from the artwork QDir search paths (exceptions are
+ *   compiled-in XPMs, you need to provide @p xpm for loading them).
+ * @param xpm Must be XPM data if the source should be raw XPM data instead of
+ *   a file
+ */
+QPixmap LMMS_EXPORT getIconPixmap( const QString&  _name,
+	int _w = -1, int _h = -1 , const char** xpm = nullptr );
 QString LMMS_EXPORT getText( const char * _name );
 
 }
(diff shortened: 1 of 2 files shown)


JohannesLorenz:
Note: CI only failed because of a used `assert()` without an include. I'll fix this soon (it looks like it's `Q_ASSERT` is common style?).

PhysSong:
`Q_ASSERT` crashes the program if not built with `QT_NO_DEBUG`. Currently we don't define it in release builds so `Q_ASSERT` will always crash LMMS.

JohannesLorenz:
@PhysSong  Thanks for the hint. Aborting if a pixmap is not loaded sounds like overkill, even in debug mode. I changed it into a warning now in e0384b8 , and the code will safely run on ...

src/gui/embed.cpp
```diff
@@ -52,6 +52,15 @@ QPixmap getIconPixmap(const QString& pixmapName, int width, int height )
 	}
 	QImageReader reader(QString("artwork:%1").arg(pixmapName));
 
+	// xpm strings are not loaded from file, you must use addXpmToCache()
+	if (pixmapName.startsWith("xpm:"))
```
PhysSong: I think it'd be better if this check were before creating a `QImageReader` instance.

include/embed.h
```diff
@@ -35,6 +36,32 @@
...
+ * Once done, you can use getIconPixmap() or new PixmapLoader() with
+ * "xpm:<unique identifier>".
+ *
+ * @note This function can only be called while a GuiApplication is running.
+ *   Otherwise, expect segmentation faults.
+ *
+ * @param pixmapName Unique identifier for the XPM. Do *not* prepend "xpm:".
+ * @param xpm_data The raw XPM string (*not* a file name).
+ */
+inline void addXpmToCache(const QString& pixmapName, const char** xpm_data) {
```
PhysSong: > This function can only be called while a `GuiApplication` is running.

Can we check it in this function and warn if the condition not met?

Issue #4597, opened 2018-09-16
Minor GUI issues


Nymbius:
There are some minor gui issues (2).
The first one is the LFO Graph.
In the LFO Graph the wave stops before the Graph box ends.
I simply fixed it by making the LFO Graph box a little bit smaller to fit the LFO wave perfectly (as shown in the picture).

The second one is the Master Volume which I fixed(?) by adding the correct spacing.
On the right is the original and on the left is my "Fixed" version.
![image](https://user-images.githubusercontent.com/39532803/45600131-d8ca5700-b9f7-11e8-8143-476a59264092.png)

PhysSong:
How did you fix the first issue? Could you create a pull request with the first fix?
I think the second change needs some discussion. That's not my area, but it might be intentional.

Nymbius:
I simply made the LFO Graph box smaller as shown on this pic : 
![image](https://user-images.githubusercontent.com/39532803/45621966-b4678c80-ba82-11e8-831a-e9db18467846.png)
As for the Master Volume I could make Strawpoll and post it to the discord server to se what people think.

PhysSong:
I think @Umcaruje/@RebeccaDeField may review your ideas and changes.

tresf:
@Nymbius can you please submit a pull request with these changes?  We'd be happy to have them!

RebeccaDeField:
@Nymbius did you edit the .svg source to make this change or the raster .png? If you need some help making a pr there should be support for that on discord, or alternatively, you could tell us the exact pixel difference you made for it to line up right so that someone else can make the pr for it.

tresf:
@RebeccaDeField his PR is available here: #4613

Nymbius:
Where can I find the .svg files?

RebeccaDeField:
@tresf  Ah, looks like I missed the PR, thanks. 

@Nymbius 
Design elements can be found here: https://github.com/LMMS/artwork/blob/master/src/DesignElements2.svg

Icons can be found here: 
https://github.com/LMMS/artwork/blob/master/src/Icons.svg

PR #4613, opened 2018-09-24
Fix minor GUI issues

Nymbius:
This should fix the minor GUI issues described here : https://github.com/LMMS/lmms/issues/4597


Umcaruje:
I agree with the LFO graph change, thanks for that.

Can you explain your reasoning for the CPU LED change? I don't find the original design flawed in any way.

Same goes to the master volume icon, I think it should stay at the original.

Nymbius:
The cpu leds were "cut" on the left side meanwhile on the right they were round. My "fix" made the both sides round : 
image
(I made the latters bold since they looked better to me that way. But it isnt as big of a problem as the "cut".)

As for the master volume I would want a poll to see what people think about my change.

RebeccaDeField:
@Nymbius Making the left side of the CPU graph flush was a stylistic choice, but I don't think making it rounded on both sides is a problem (unless @Umcaruje's opinion differs).

Bold text is also okay but what you posted is fuzzy and less clear in comparison which is caused by aliasing problems. IIRC that text was custom made and pixel perfect (path, not font) so I'm not sure what you did to get the bold effect but it would have to be edited manually with nodes and not stretched.

PhysSong:
I think we may also fix the LFO graph issue by tweaking the drawing code.
https://github.com/LMMS/lmms/blob/dd6d4a552b3e1c51fc39db7c0cfc19d3f33908ef/src/gui/widgets/EnvelopeAndLfoView.cpp#L494-L497
A similar issue exists in the classic theme as well...

tresf:
@RebeccaDeField can you please fix and merge these small changes?  I feel there's a lot of dialog for a very quick change. :)

zonkmachine:
There seem to be a 1 pixel glitch on the left side of the graph too.

before
lfograph3


after
lfograph

after with this tweak of the snippet @PhysSong linked to above.
```diff
-       int LFO_GRAPH_W = s_lfoGraph->width() - 6;      // substract border
+       int LFO_GRAPH_W = s_lfoGraph->width() - 3
;      // substract border
        int LFO_GRAPH_H = s_lfoGraph->height() - 6;     // substract border
-       int graph_x_base = LFO_GRAPH_X + 3;
+       int graph_x_base = LFO_GRAPH_X + 2;
        int graph_y_base = LFO_GRAPH_Y + 3 + LFO_GRAPH_H / 2;
```

RebeccaDeField:
@tresf trying to explain changes when I don't have a computer in condition to make them on tends to end with a lot of words and not so much action. ^^;

This should be a simple fix, I'll take care of it as soon as I have availability to do so.

tresf:
I just commited the recommended changes to move this along.  Please test and merge at your earliest convenience.

tresf:
> I'll take care of it as soon as I have availability to do so.

No worries, we can tackle the small stuff.

zonkmachine:
> I just commited the recommended changes to move this along.

Then you need to also revert `data/themes/default/lfo_graph.png`

tresf:
> Then you need to also revert data/themes/default/lfo_graph.png

Done.

zonkmachine:
I don't see that things are moving along here fast enough that this will reach 1.2 . I suggest pushing 3db03e0 and fb2b0a4 to stable-1.2 directly and then closing this PR. You can continue the discussion of details of the gui/theme in the original issue, #4597 and take the time you need. Then issue a new PR against master instead.

PR #4635, opened 2018-09-29
Add wet/dry panning knob to all LMMS effects

### LINKED ISSUE #4611: Add wet/dry panning knob to all LMMS effects
Please notify me if this has already been requested.

Currently every LMMS effect has a Wet/Dry, Decay, and Gate knob.  A panning knob would be another fantastic thing to add, which works in companionship with the Wet/Dry knob.  It simply changes the Wet/Dry of individual ears.

This would be useful for many things:

1. I'm sure there's already a request out there to change the Equalizer so it allows different gains for each ear, this would take care of that perfectly.
2. When you want to apply an effect to just one ear, this allows that.
3. You can also have strange and unique methods of obtaining more stereo width in your music.
4. Other advantages would depend on the plugin, there are many things you can do with a feature like this!

For anybody who wants to do this with LMMS's current capabilities, here's the process they have to go through for every effect:

1. Send instrument to two different busses
2. Use the Amplifier effect to mute the left/right sides, one side per bus
3. Use two copies of the effect with different Wet/Dry values
4. Send both back into another FX bus

With that method, keeping everything sane would be... not fun.  Not to mention that the requirement of two copies of the same effect may have a very undesirable impact on weaker computers.

I understand that this would be difficult to do since the size of an LMMS effect is made to fit three knobs.  Here's the only idea I can think of (the length of the effect has not been modified, I simply moved the Controls button over):

image

Hopefully somebody can find a neat way to implement this feature, I feel like LMMS has needed this for a long time.

LostRobotMusic:
Resolves #4611 

I named the new knob "PAN".  This is what it looks like: 

screenshot from 2018-09-29 10-31-01

### DIFF
--- a/include/Effect.h
+++ b/include/Effect.h
@@ -117,6 +117,27 @@ public:
 		return 1.0f - m_wetDryModel.value();
 	}
 
+	inline float panAmount() const
+	{
+		return m_panModel.value();
+	}
+	inline float wetLevelL() const
+	{
+		return ( m_panModel.value() < 0 ) ? m_wetDryModel.value() : ( m_wetDryModel.value() * ( ( -m_panModel.value() + 100 ) / 100 ) );
+	}
+	inline float wetLevelR() const
+	{
+		return ( m_panModel.value() > 0 ) ? m_wetDryModel.value() : ( m_wetDryModel.value() * ( ( m_panModel.value() + 100 ) / 100 ) );
+	}
+	inline float dryLevelL() const
+	{
+		return 1.0f - ( ( m_panModel.value() < 0 ) ? m_wetDryModel.value() : ( m_wetDryModel.value() * ( ( -m_panModel.value() + 100 ) / 100 ) ) );
+	}
+	inline float dryLevelR() const
+	{
+		return 1.0f - ( ( m_panModel.value() > 0 ) ? m_wetDryModel.value() : ( m_wetDryModel.value() * ( ( m_panModel.value() + 100 ) / 100 ) ) );
+	}
+
 	inline float gate() const
 	{
 		const float level = m_gateModel.value();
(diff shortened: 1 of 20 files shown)


DouglasDGI:
It may be worth pointing out that this may look sort of ugly when using the classic LMMS theme.  It looks just fine with the default theme though.

JohannesLorenz:
@DouglasDGI Does this require a review, or is it still WIP?

DouglasDGI:
@JohannesLorenz It is fully functional, though there are some GUI concerns some people have.  I can do what Spekular suggested, but back when I made this PR I had no idea how Github worked and now the original repository has been removed, so I'm unsure of what I can do to apply those changes...

curlymorphic (changes requested):
Hi, I have looked at your code from a dsp perspective, we just need to update the panning law used as mentioned above.

I wonder if the panning is acting as intended, should the wet panning adjust the panning of the dry signal? My preference would be to leave the dry signal alone but is worth discussing here.

rdrpenguin04:

LostRobotMusic:
This feature shouldn't be added to LMMS.  Back when I conceptualized it, I was brand new to audio processing.  Simply adjusting wet/dry in each channel to obtain effect panning would introduce major issues in any circumstance in which short delays or phase shifts are present (e.g. any effect with basic IIR filters, especially lowpass and highpass filters).  Panning the processing of an effect has to be handled on a case-by-case basis by the effect itself for this issue to be avoided, not by making wet/dry a stereo parameter.  There's a very good reason no other DAW has this feature.

zonkmachine:
@LostRobotMusic OK. So close https://github.com/LMMS/lmms/issues/4611 ? :wink:

Issue #4624, opened 2018-09-26
[Feature Request] Rendering looped sections multiple times on export


StevenChristy:
I would like the ability to export my song with the looped section repeated 2-3 times. Ideally this would work by allowing one to specify the number of times the looped section is played while rendering (it would default to 1). Increasing this to 2 or 3 would cause the looped section to be repeated. 

Such a feature does not necessarily need to work with other looped options and the upper limit on loops could be capped at a low value if there is a good reason to put a limit on it.

StevenChristy:
I decided to implement the feature in my fork: https://github.com/StevenChristy/lmms/commit/5df14e3a028cfd3ec7f75f6e8a422d102be48c83
The only part I didn't address is the progress bar. I figured it wasn't worth all the hassle unless the feature was deemed useful enough for inclusion in the origin.

tresf:
@StevenChristy thanks!  Can you send us a screenshot?  Probably worth including, I'll ask the users on Discord to give feedback.

StevenChristy:
Here is what the export UI looks like with the feature added:

![lmms_multiloop_export](https://user-images.githubusercontent.com/1801979/46250387-a9581900-c3ff-11e8-8dfe-d8b4bd4e75e5.png)

The scrollbar currently behaves as one would expect, resetting itself to a certain point on each iteration of the loop. I can correct it but the changes are more complex.

PR #4639, opened 2018-09-30
Rendering looped sections multiple times on export (#4624)

StevenChristy:
@tresf suggested in https://github.com/LMMS/lmms/issues/4624 that this feature might be useful for inclusion in LMMS.

This pull request adds the ability to render looped sections multiple times when exporting. A spinbox is added to the export dialog that allows one to specify how many times the looped section would be rendered while exporting. This defaults to 1  but the user can specify a value to allow the section between looping markers to be rendered up to 9 times (change the spinbox maximum in export_project.ui to increase this if a greater number than 9 is desired).

Feature was tested with both Export Project and Export Tracks but is not applicable to Export to MIDI. The method for calculating export progress needed to be changed to allow the progress bar to give a correct estimate for the percentage completion. There are no known limitations or side effects other than what is described above.

### DIFF
--- a/src/core/Song.cpp
+++ b/src/core/Song.cpp
@@ -86,7 +86,8 @@ Song::Song() :
 	m_patternToPlay( NULL ),
 	m_loopPattern( false ),
 	m_elapsedTicks( 0 ),
-	m_elapsedTacts( 0 )
+	m_elapsedTacts( 0 ),
+	m_loopRenderCount ( 1 )
 {
 	for(int i = 0; i < Mode_Count; ++i) m_elapsedMilliSeconds[i] = 0;
 	connect( &m_tempoModel, SIGNAL( dataChanged() ),
@@ -330,7 +331,7 @@ void Song::processNextBuffer()
 			}
 			m_playPos[m_playMode].setTicks( ticks );
 
-			if( checkLoop )
+			if( checkLoop || m_loopRenderCount > 1 )
 			{
 				m_vstSyncController.startCycle(
 					tl->loopBegin().getTicks(), tl->loopEnd().getTicks() );
@@ -340,6 +341,8 @@ void Song::processNextBuffer()
 				// beginning of the range
 				if( m_playPos[m_playMode] >= tl->loopEnd() )
 				{
+					if( m_loopRenderCount > 1 ) 
+						m_loopRenderCount--;
 					ticks = tl->loopBegin().getTicks();
 					m_playPos[m_playMode].setTicks( ticks );
 					setToTime(tl->loopBegin());
@@ -745,6 +748,7 @@ void Song::stopExport()
 	stop();
 	m_exporting = false;
 	m_exportLoop = false;
+	m_loopRenderCount = 1;
 
 	m_vstSyncController.setPlaybackState( m_playing );
 }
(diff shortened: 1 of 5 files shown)


PhysSong:
I did some simple tests. It worked well even with automated time signatures. :tada:
However, I think you should try to address our [coding convention](https://github.com/LMMS/lmms/wiki/Coding-conventions). There are some codes which don't agree with our coding conventions. Please let me know if you want to know what to change.
For GUI changes, I'd ask @Umcaruje and @RebeccaDeField for reviews.

StevenChristy:
@PhysSong, thanks. I haven't had a chance to address this yet, but I noticed that the existing code doesn't seem to follow the coding conventions you stated. At the risk of being a little too snarky, I think its worth mentioning that if the existing code doesn't follow the conventions that the project doesn't really have a convention... However, I will make the changes you've requested.

@Umcaruje / @RebeccaDeField, please review the GUI changes. Thank you.

StevenChristy:
@PhysSong, the whitespace changes you requested were addressed. I also corrected one inconsistency in the .ui file that didn't appear to me to do much, but might have been important for other platforms.

zonkmachine:
Testing the AppImage above. It looks like the code has been changed since so I'll try and compile this later. I think this works really well and looks good.


> up to 9 times (change the spinbox maximum in export_project.ui to increase this if a greater number than 9 is desired).

This is too little. There is room for more there so think big. I saw one user complaining about issues with tracks of the size 2h and longer. One may want to render a long drone for a hippie convention etc. I'd go with 9999 as max. :)

zonkmachine:
This is good to merge. I haven't gone over the code with a magnifying glass but it looks good.

Some thoughts. 

 * 9 times is to little for number of times to loop. At least 999 times is the way to go.
 * You could show time of track to be exported as a function of `'time of loop'` * `'number of times'`. This would be good to know when looping many times.

zonkmachine:
Merge?

zonkmachine:
> change the spinbox maximum in export_project.ui to increase this if a greater number than 9 is desired

Done. Bumped it to 99.

src/core/Song.cpp
```diff
@@ -721,15 +737,41 @@ void Song::stop()
...
+		if (!m_exportLoop) 
+			m_exportSongEnd += MidiTime(1,0);
+        
+		m_exportSongBegin = MidiTime(0,0);
+		m_exportLoopBegin = m_playPos[Mode_PlaySong].m_timeLine->loopBegin() < m_exportSongEnd && 
+			m_playPos[Mode_PlaySong].m_timeLine->loopEnd() <= m_exportSongEnd ?
+			m_playPos[Mode_PlaySong].m_timeLine->loopBegin() : MidiTime(0,0);
+		m_exportLoopEnd = m_playPos[Mode_PlaySong].m_timeLine->loopBegin() < m_exportSongEnd && 
+			m_playPos[Mode_PlaySong].m_timeLine->loopEnd() <= m_exportSongEnd ?
+			m_playPos[Mode_PlaySong].m_timeLine->loopEnd() : MidiTime(0,0);
```
PhysSong: I found that `m_timeLine` is null if LMMS runs without GUI. As a consequence, this breaks the command-line rendering.
I think we can work around this by not setting `m_exportLoopBegin` and `m_exportLoopEnd` when the timeline is null because we don't use that in CLI renders.

FYI, that check is not needed elsewhere because LMMS doesn't support "export between loop markers" feature on the command line(for a similar reason).

Issue #4632, opened 2018-09-28
Automating "mute" is broken on project load


RoxasKH:
Every time you do an automation of the mute control (the green led) either on mixer, tracks on songeditor or tracks in beat bassline editor, after restart lmms and reopen the project, the control links are lost and so you have to redo ctrl every time you load it. 

This is really annoying, and i think it's a serious bug. Hope this helped for RC8 :)

To reproduce it just do like i did here

![immagine](https://user-images.githubusercontent.com/40141826/46228417-4f554600-c362-11e8-83e1-4cab361fc493.png)

The first is an automation on the 3osc track, the second of the kicker in B&B ed and the third one is on master mixer. After reloading the project neither of those will work

Gabrielxd195:
Friend in which version of lmms are you ?, I had never seen that interface.

jibin1573:
@Gabrielxd195, it is the FL studio theme which @RoxasKH has developed. If you are unaware, I'll like to bring to your notice that you can apply different themes on LMMS. Here you can find different type of themes: https://lmms.io/lsp/?action=browse&category=UI%20themes

RoxasKH:
@Gabrielxd195 When lmms 1.2 stable will be released i will release my theme, if you're interested, but the main argument of this was the automation problem, so the next time contact me through DM

musikBear:
Lets stay with the **Issue** -shall we?
I *confirm* the bug on win 32 rc7
It is a real loss of connection. The chosen shape does not apply

tresf:
@PhysSong we may be able to tackle this when #3781 is closed.  Milestoning for 1.2.0 assuming it's a quick fix.  If not, feel free to bump it to 1.3.0.

PhysSong:
> @PhysSong we may be able to tackle this when #3781 is closed.

We can fix this even if #3781 is open.
> assuming it's a quick fix

Yes. Only two lines in `Track::saveSettings` need to be changed to save automation for mute/solo. FX mixer's mute/solo works well, so we need to just adapt the logic.

PhysSong:
A few more things which can be automated but its settings can't be saved:
- The tension of automation patterns
- Whether to enable effect chains

If we address them as well, it'll be enough to change just 8 lines.
Note that saving them may break loading it for old versions. However, I think that'd be fine because upgrading LMMS is enough. Also, we're approaching to the 1.2 stable now. :)

PhysSong:
> Note that saving them may break loading it for old versions.

[...]

Issue #3639, opened 2017-06-14
About track/FX channel mute automation


PhysSong:
In automation of track(or FX channel) mute button, 0 means it isn't muted, and 1 means it is muted. In other LED controls, however, 0 means inactive state,  1 means active state.
Maybe people use track/FX channel mute automation very rarely, I think it can be confusing because the mute buttons of NOT muted tracks look like other LED toggles in ACTIVE state. First one has value 0 while second one has value 1.

How do you guys think about it?

Hussam-Eddin-Alhomsi:
I agree.
The problem here isn't in the automation itself, it's in the values of the button, you can check these values by right clicking on the button.
**The values should be reversed to be 0 when it's inactive, and 1 when it's active.**

I think that the button we are talking about here shouldn't be called the "_Mute_" button but called the "_Mute/Unmute_" button. People think that "_Mute_" button is the right name because the tooltip that the button displays when hovering over it (wether it's inactive or active) says: "_Mute this track_", which is false.
**The tooltip should display "_Mute_" when it's active/unmuted & "_Unmute_" when it's inactive/muted.**

The "_Solo_" button has the same problem as in the paragraph above. The tooltip that the button displays when hovering over it (wether it's inactive or active) says: "_Solo_", which is false.
**The tooltip should display "_Solo_" when it's active/soloed & "_Solo_" when it's inactive/unsoloed.**

tresf:
The text "mute" is fine, but I agree, it's a matter of context.  You're right, the state is inverted in regards to the action.  Making a negative state LED and an upgrade routine for backwards compatibility should be enough.

The solo button is logic and properly implemented/lit.  More text is superfluous.  This is an inverse-state problem, nothing more.

tresf:
I'd even take it a step further and say the mute should be lit up red.  This would make it more obvious when a mute action has occurred, such as when the solo button is pressed on another track.  Making all tracks go from green "audible" to red "muted" seems like better UI to me.

Umcaruje:
setting the mute button's default state to off would make both LED's the same and the mute and solo buttons not distinguishable. I think we could switch to actual `M` and `S` buttons imho

tresf:
> I think we could switch to actual M and S buttons imho

This idea was shot down before by one of our predecessors but this is consistent with other DAWs as well as @budislav's [mockup](https://github.com/LMMS/lmms/issues/1911#issue-65360834).  This would do just fine as well.

[...]

PR #4667, opened 2018-10-20
Fix not saving some automations (#4632)

JohannesLorenz:
OK, tested:
* saving + loading all 3 Models with this patch works
* saving with the patch + loading without will discard the automation... ugly, but we can't get any better here...
* saving without the patch (i.e. doing no automation) + loading with the patch works (the knobs are simply not automated, but keep their values)

I would say let's just merge it in.

### DIFF
--- a/src/core/EffectChain.cpp
+++ b/src/core/EffectChain.cpp
@@ -53,7 +53,7 @@ EffectChain::~EffectChain()
 
 void EffectChain::saveSettings( QDomDocument & _doc, QDomElement & _this )
 {
-	_this.setAttribute( "enabled", m_enabledModel.value() );
+	m_enabledModel.saveSettings( _doc, _this, "enabled" );
 	_this.setAttribute( "numofeffects", m_effects.count() );
 
 	for( Effect* effect : m_effects)
@@ -80,7 +80,7 @@ void EffectChain::loadSettings( const QDomElement & _this )
 
 	// TODO This method should probably also lock the mixer
 
-	m_enabledModel.setValue( _this.attribute( "enabled" ).toInt() );
+	m_enabledModel.loadSettings( _this, "enabled" );
 
 	const int plugin_cnt = _this.attribute( "numofeffects" ).toInt();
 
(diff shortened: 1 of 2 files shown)


zonkmachine:
Saving/loading the effect chain switch works fine.

The Solo/Mute automation seem to have some peculiarities though. These are not from this PR however but I think it should at least be discussed.
 - Mute. The mute switch automation seem to have its behavior inverted. A zero turns the channel on and one turns it off. I expect the action to be the opposite.
- Solo. When the solo switch of an instrument track is turned off by an automation track the automation track is inactivated too so whatever comes next in the automation is invalid.

JohannesLorenz:
> The mute switch automation seem to have its behavior inverted.
> A zero turns the channel on and one turns it off. I expect the action to be the opposite.

The problem is that the model is named "mute" (which means true = no sound), but what you see (from the LED) is "active" or "on" (which means true = sound). Renaming the model to "active" would be doable, but I'm afraid this could cause regression issues in such a late RC.

> When the solo switch of an instrument track is turned off by an automation track the
> automation track is inactivated too so whatever comes next in the automation is invalid.

Imagine the following naive approach: each time when a track is getting solo, when iterating over the other tracks in `Track::toggleSolo`, do not mute those tracks that are `AutomationTrack`s and have an `AutomationPattern` at the current position which has the current Track's solo model in their `AutomationPattern::m_objects`.

But if you do it like this, imagine the solo-connected Automation pattern is also connected to an entirely different boolean model (or even just another pattern in this track). Having the whole AutomationTrack not muted would be wrong, as we're obviously still automating something else. But having it muted would also be wrong, as you noticed.

Conclusion: you need to mark all AutomationTracks as muted, but let LMMS still compute the AutomationPatterns and then decide in `AutomatableModel::setAutomatedValue` for which linked model you really set the automated value (these are exactly the solo models).

The solution here seems to get a deeper core thing and can not be done for 1.2. Maybe it's better if you open an enhancement?

zonkmachine:
> The problem is that the model is named "mute" (which means true = no sound), but what you see (from the LED) is "active" or "on" (which means true = sound). Renaming the model to "active" would be doable, but I'm afraid this could cause regression issues in such a late RC

OK. I think an issue should be opened for this then (label ux).

PhysSong:
> I think an issue should be opened for this then (label ux).

There's #3639

PhysSong:
I think the soloing issue can be fixed by adding an option which prevents a track from being muted when another track gets soloed.

Anyway, I'd say let's merge this unless it causes severe regressions.

PR #2466, opened 2015-11-26
Bezier curves for automation tracks

codythecoder:
Based on [this](https://github.com/LMMS/lmms/issues/2381) issue, I think having bezier curves in LMMS would make automation tracks a lot more powerful, and much faster to do certain types of curves.

PR #4674, opened 2018-10-23
Support Bezier curves for automation patterns

### LINKED ISSUE #2381: Bezier curves for automation tracks
So one problem I have with the automation tracks as they are is that they lack control for the endpoints. It's impossible to get the curve to start or end going strait upwards, which can sometimes be what you want to do, and the only way to do this at the moment is with a lot of points. I think that bezier curves would be able to solve this problem, and make the automation tracks easier for newcomers and seasoned users alike.
Here's a rough idea of what I'd imagine it would look like:
temp
temp2
temp3

I can do the coding, but what do people think? Anything you would add/change to this?

PhysSong:
This pull request continues #2466 which was done by @codythecoder. Note that I haven't tested this feature much yet.
Closes #2381.


qnebra:
No crashes when I worked on lmms with this branch (and a lot others) merged on quite challenging project is a good sign.

PhysSong:
This pull request now have massive merge conflicts due to #5712 and class renames. I will try to resolve them after refactor.

PhysSong:
I think I should wait for #5924, which is very likely to contain a lot of duplicated works.

zonkmachine:
> I think I should wait for #5924, which is very likely to contain a lot of duplicated works.

#5924 has been merged.

PhysSong:
> I believe this can be resurrected

Yes, I just don't have enough time for now(until the end of this month).

PhysSong:
I'll possibly redo the work because of the large changes in the codebase.

Rossmaxx:
I'm moving this out of the 1.3 milestone, as @PhysSong might not have time to complete this.

messmerd:
@PhysSong Do you think it would be possible to upgrade all existing cubic hermite curves to Bezier curves?

Cubic Hermite curves have the undesirable property of affecting the part of the automation curve before and after their start/end points, which is the primary reason #7477 needed to split automation clips using a "soft split" (a workaround which duplicates the clip, keeping all automation nodes, but then simply adjusts their start/end regions to emulate a real, "hard split").

If Bezier curves are introduced, I see no reason to keep Cubic Hermite curves around. Not saying they should be removed in this PR though.

PhysSong:
> @PhysSong Do you think it would be possible to upgrade all existing cubic hermite curves to Bezier curves?

Yes, at least theoretically, by putting two control points at the 1/3 and 2/3 time points of a segment. The generalization will be likely to add some more computations, though.

> Cubic Hermite curves have the undesirable property of affecting the part of the automation curve before and after their start/end points

@messmerd BTW, what does that mean? Were there no reasonable alternative approaches?

messmerd:
> BTW, what does that mean? Were there no reasonable alternative approaches?

Nevermind, I think we could avoid the issue if we wanted to using the ability to edit tangents added in #5924. We would just need to insert a node at the split point and set the tangents so the curve doesn't change after the clip is split.

PR #4678, opened 2018-10-25
Allow building SF2 player with FluidSynth 2.x

### LINKED ISSUE #4640: lmms 1.2.0-rc7 fails to build with fluidsynth 2.0.0
I'm currently rebuilding a bunch of software against fluidsynth 2.0.0 on Arch Linux.
Sadly that introduced an [API change](http://www.fluidsynth.org/api/index.html#NewIn2_0_0).

LMMS currently fails to build with it: 
[lmms-1.2.0rc7-fluidsynth2.0.txt](https://github.com/LMMS/lmms/files/2434148/lmms-1.2.0rc7-fluidsynth2.0.txt)

Any help in integrating/patching the `sf2_player` would be greatly appreciated!
Plus, I think this should probably be fixed before releasing 1.2.0!

PhysSong:
Resolves the incompatibility between FluidSynth 1.x and 2.x due to some API changes by shimming some functions.
Fixes #4640.

### DIFF
--- a/plugins/sf2_player/fluidsynthshims.h
+++ b/plugins/sf2_player/fluidsynthshims.h
@@ -1,0 +1,88 @@
+/*
+ * fluidsynthshims.h - a shim header for FluidSynth 2.0 API changes
+ *
+ * Copyright (c) 2018 Hyunjin Song <tteu.ingog@gmail.com>
+ *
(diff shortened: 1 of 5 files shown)


PhysSong:
I'll merge this in a day because it's tested by some people including myself.

PR #4690, opened 2018-11-01
Add .clang-format and .clang-tidy

JohannesLorenz:
Don't merge yet, needs discussion, see below.

The following patch provides style files for ~2/3 of the [LMMS coding conventions](https://github.com/LMMS/lmms/wiki/Coding-conventions) (~4/5 if clang-tidy will accept a PR I'll need to create).
clang-format and clang-tidy provide functionality to find style issues (the former for whitespace, the latter for code) and to fix them automatically.

The following works as an example (assuming you're in the main LMMS folder and use a build dir `build`). For best results, use what `make VERBOSE=1` spits out.

```
# clang-format -style=file <filename>
clang-format -style=file src/core/AutomatableModel.cpp
# clang-tidy -config= <filename> -- <compile-flags>
clang-tidy -config-file=.clang-tidy src/core/AutomatableModel.cpp -- -Iinclude -Ibuild/{src,} -I/usr/include/qt/{QtCore,QtWidgets,QtGui,QtXml,}
```

QtCreator:
Full support for clang-format and clang-tidy :smile: 
* clang-tidy:
  - Analyze -> Clang-Tidy and Clazy
  - Select custom style, click manage, duplicate existing one, select Clang-Tidy tab, choose to read style from file
  - Select the files that you want to analyze (selecting all takes a long time!), click "Analyze"
* clang-format
  - Enable Help -> About Plugins -> C++ -> Beautifier
  - Restart
  - Tools -> Options -> Beautifier, do your settings, but especially, go to the "Clang Format" tab, use predefined style "file"


clang-format output for Mixer.cpp:

[click here](https://imgur.com/a/8Va0svF) for examples.


clang-tidy output for Mixer.cpp:

```
/home/johannes/cprogs/lmms/master/src/core/Mixer.cpp:109:20: warning: redundant boolean literal supplied to boolean operator [readability-simplify-boolean-expr]
        if( renderOnly == false )
            ~~~~~~~~~~~~~~^~~~~
            !renderOnly                                                                                                                                                                                                                      
/home/johannes/cprogs/lmms/master/src/core/Mixer.cpp:306:64: warning: redundant boolean literal supplied to boolean operator [readability-simplify-boolean-expr]
        return cpuLoad() >= 99 && Engine::getSong()->isExporting() == false;
                                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~
                                  !Engine::getSong()->isExporting()                                                                                                                                                                          
/home/johannes/cprogs/lmms/master/src/core/Mixer.cpp:348:23: warning: invalid case style for static variable 'last_metro_pos' [readability-identifier-naming]
        static Song::PlayPos last_metro_pos = -1;
                             ^~~~~~~~~~~~~~
                             s_last_metro_pos                                                                                                                                                                                                
/home/johannes/cprogs/lmms/master/src/core/Mixer.cpp:374:18: warning: Potential memory leak [clang-analyzer-cplusplus.NewDeleteLeaks]
                last_metro_pos = p;
                               ^
[...]

/home/johannes/cprogs/lmms/master/src/core/Mixer.cpp:405:8: warning: statement should be inside braces [readability-braces-around-statements]
                        else delete *it;
                            ^
                             {
/home/johannes/cprogs/lmms/master/src/core/Mixer.cpp:459:8: warning: statement should be inside braces [readability-braces-around-statements]
                        else delete *it;
                            ^
                             {
/home/johannes/cprogs/lmms/master/src/core/Mixer.cpp:679:25: warning: redundant boolean literal supplied to boolean operator [readability-simplify-boolean-expr]
        if( criticalXRuns() == false )
            ~~~~~~~~~~~~~~~~~~~^~~~~
            !criticalXRuns()
/home/johannes/cprogs/lmms/master/src/core/Mixer.cpp:690:6: warning: statement should be inside braces [readability-braces-around-statements]
        else delete handle;
            ^
             {
/home/johannes/cprogs/lmms/master/src/core/Mixer.cpp:743:8: warning: statement should be inside braces [readability-braces-around-statements]
                        else delete _ph;
                            ^
                             {
/home/johannes/cprogs/lmms/master/src/core/Mixer.cpp:769:8: warning: statement should be inside braces [readability-braces-around-statements]
                        else delete *it;
                            ^
                             {
/home/johannes/cprogs/lmms/master/src/core/Mixer.cpp:785:25: warning: statement should be inside braces [readability-braces-around-statements]
        if( s_renderingThread )
                               ^
                                {
/home/johannes/cprogs/lmms/master/src/core/Mixer.cpp:807:25: warning: statement should be inside braces [readability-braces-around-statements]
        if( s_renderingThread )
                               ^
                                {
Suppressed 13365 warnings (13365 in non-user code).
```

Things to discuss:
* CMake integration: multiple possibilities:
  - No CMake integration
    * Users must use clang-tidy from their IDE, if they use an IDE supporting it,
      otherwise they have to run clang-tidy manually, passing command line args to it
    * No Continous integation!
  - Let CMake only [export compile commands](https://cmake.org/cmake/help/v3.5/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html)
    * No full automation
  - Always run clang-tidy at compile, if a flag is set
    * Compile time overhead!
    * The need to set that CMake flag may confuse users
  - Add a target for CMakeLists to fix (and alternatively just to scan) all source files
    * Disadvantage: Can only accept all fixes, where in an IDE you can select what to fix
* Line length: The coding conventions say "80", but it seems to look better using "100".
  With 80, many `QObject::connect` calls are broken into 4 lines.

For the future, I hope we can fix the complete LMMS sources after stable-1.2 is being merged.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,68 @@
+---
+# Language
+Language:        Cpp
+Standard:        Cpp11  # Cpp14 and Cpp17 are not supported by clang 11
+
+# Indentation
+TabWidth:        4
+UseTab:          Always
+IndentWidth:     4
+ColumnLimit:     0
+
+# Indentation detail
+AlignAfterOpenBracket: DontAlign
+ContinuationIndentWidth: 4
+BreakConstructorInitializers: BeforeComma
+ConstructorInitializerIndentWidth: 4
+ConstructorInitializerAllOnOneLineOrOnePerLine: false
+BinPackParameters: true
+BinPackArguments: true
+AlignOperands: false
+AlignEscapedNewlines: DontAlign
+AccessModifierOffset: -4
+
+# Includes
+IncludeBlocks:   Regroup
+IncludeCategories:
+  # the "main header" implicitly gets priority 0
+  # system headers
+  - Regex:           '^<[^>]+>$'
+    Priority:        1
+  # non-system headers
+  - Regex:           '.*'
+    Priority:        2
+SortIncludes:    true
+
+# Spaces
+SpaceBeforeAssignmentOperators: true
+SpaceBeforeParens: ControlStatements
+SpacesInAngles:  false
+SpacesInCStyleCastParentheses: false
+SpacesInParentheses: false
+
+# Brace wrapping
+# Not directly mentioned in the coding conventions,
+# but required to avoid tons of auto reformatting
+BreakBeforeBraces: Custom
+BraceWrapping:
+  AfterClass: true
+  AfterControlStatement: true
+  AfterEnum: true
+  AfterFunction: true
+  AfterNamespace: true
+  AfterStruct: true
+  AfterUnion: true
+  AfterExternBlock: true
+  BeforeCatch: true
+  BeforeElse: true
+  IndentBraces: false
+  SplitEmptyFunction: true
+  SplitEmptyRecord: true
+  SplitEmptyNamespace: true
+  
+# Pointers
+# Use pointer close to type: `const char* const* function()`
+PointerAlignment: Left
+
+...
+
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,33 @@
+---
+Checks:          'bugprone-macro-parentheses,bugprone-macro-repeated-side-effects,modernize-use-override,readability-identifier-naming,readability-misleading-indentation,readability-simplify-boolean-expr,readability-braces-around-statements'
+WarningsAsErrors: ''
+HeaderFilterRegex: '' # don't show errors from headers
+AnalyzeTemporaryDtors: false
+FormatStyle:     none
+User:            user
+CheckOptions:    
+  - key:             readability-identifier-naming.ClassCase
+    value:           CamelCase
+  - key:             readability-identifier-naming.EnumCase
+    value:           CamelCase
+  - key:             readability-identifier-naming.TypedefCase
+    value:           CamelCase
+  - key:             readability-identifier-naming.UnionCase
+    value:           CamelCase
+  - key:             readability-identifier-naming.StructCase
+    value:           CamelCase
+  - key:             readability-identifier-naming.UnionCase
+    value:           CamelCase
+# not yet working, as it currently applies both for static and object members
+# - key:             readability-identifier-naming.MemberPrefix
+#   value:           'm_'
+  # currently only working for local static variables:
+  - key:             readability-identifier-naming.StaticVariablePrefix
+    value:           's_'
+# not yet working
+# - key:             readability-identifier-naming.VariableCase
+#   value:           camelBack
+  - key:             readability-identifier-naming.FunctionCase
+    value:           camelBack
+...
+


JohannesLorenz:
Per 0b5a7d8, function call arguments are now not separated by newlines, so e.g. calls to `QObject::connect` are more compact now.

JohannesLorenz:
Explanation about the constructor lists: As you can see, now multiple initializers from the CTOR initializer list are put into one line. I find that ugly, but it looks like clang-format does not provide any workaround here.

JohannesLorenz:
A few more examples: JohannesLorenz@bdde4e3 shows code-reading rework I did automized, and JohannesLorenz@1e091ea shows rework that remained to do by hand. As one can see, types are corrected automatically, while variables can currently not being checked (due to clang-format which can not differ between `s_` and `m_` variable prefixes).

.clang-format
```diff
@@ -0,0 +1,65 @@
...
+BinPackParameters: true
+BinPackArguments: true
+AlignOperands: false
+AlignEscapedNewlines: DontAlign
+AccessModifierOffset: -8
+
+# Includes
+IncludeBlocks:   Regroup
+IncludeCategories:
+  # the "main header" iplicitly gets priority 0
```
PhysSong: Typo.

JohannesLorenz:
Side note: The tab width (4 or 8) is not clearly defined within the project. We need to find that out.

Reflexe:
#4810 states that we should change to spaces (4) instead of tabs. clang-format could save me a lot of time instead of trying to guess every project's code style (and then force-push style changes). Also, clang-tidy could help us to modernize the codebase (e.g. foreach instead of the old "iterator for", `auto` in casts, and more)

Spekular:
Looks to me like tabs of length four have most votes in each poll, am I
missing somewhere where spaces are decided on?

Reflexe:
> Looks to me like tabs of length four have most votes in each poll, am I missing somewhere where spaces are decided on?

Im sorry. I got confused between the emojis.

JohannesLorenz:
Tab width fixed to `4`.

Reflexe:
Maybe we should just use Qt's code style with 4-width tabs? It will be a
lot easier to set up with qtcreator and clang-format. Also, it takes me so
much time and frustration to manually format my code every time.

JohannesLorenz:
Some more additional info:

* You can turn lines off that you really do not want to format (e.g. our constructor initializer lists?):
```
// clang-format off
...
// clang-format on
```
* There's a [github clang formatter app](https://github.com/apps/clang-formatter), similar to travis (I think @Reflexe found it out)

JohannesLorenz:
> Maybe we should just use Qt's code style with 4-width tabs? It will be a lot easier to set up with qtcreator and clang-format. Also, it takes me so much time and frustration to manually format my code every time.

Setting up a coding style once for QtCreator is no big deal. Just create an XML file and you can always re-use this file. Setting it up for clang-format won't be easier: With this PR, we already have a setup for clang-format.

I'm not sure if Qt's code style will match ours. At least for some variable names, this is true, but every derivation will cause more changes in our code base, which is why I was for defining our own code style.

JohannesLorenz:
I find it problematic that this PR is almost ignored. If properly set up, it can increase PR review time a lot, but the points that need to be discussed (as mentioned in the first post) are not discussed at all. The example images/outputs have also not been discussed.

1 or 2 persons can't decide anything that will have significant influence for all active coders. So I'm going to tag everyone to find a consent. Please read the posts carefully, check the examples, make proposals, and if in doubt, try it out on your machine.

* @DomClark
* @michaelgregorius
* @Sawuare
* @jasp00
* @Wallacoloo
* @zonkmachine
* @PhysSong
* @tresf
* @lukas-w

Sawuare:
@JohannesLorenz I agree with you on increasing the column limit to 100.

tresf:
> column limit to 100 instead of 80? Why?  Because you picked 100?  Just get rid of it.

Spekular:
At the very least 120 imo, otherwise the increase is marginal.

JohannesLorenz:
Line width is not only about editor for me. E.g. take this line:
```
m_syncData->m_latency = m_syncData->m_bufferSize * m_syncData->m_bpm / ( (float) m_syncData->m_sampleRate * 60 );
```
Do you really want to read that? You could make this more readable by using:
```
<whatever type>& d = m_syncData;
d->m_latency = d->m_bufferSize * d->m_bpm / ( (float) d->m_sampleRate * 60 );
```

tresf:
... but it's misplaced to use arbitrary values to fix unreadable code.  It's a anecdotal fix.  Code reviews will still be required for human readibility.  A computer can't automate all human tasks.

tresf:
Also, some of the advanced editors can help simplify and clean code using hints.  JetBrains has excellent code helpers for this stuff.

JohannesLorenz:
> Why? Because you picked 100? Just get rid of it.

I'm not sure if clang-format works sane without any line length limit (or a limit of, say 200). I guess it would concatenate almost every multi-liner into a very very long line... Maybe this must be tested.

Btw, @DouglasDGI Sorry, I forgot you in the list of "active coders"

winniehell:
@JohannesLorenz Thank you so much for starting this! :heart: I was just about to suggest something like this.

---

Addressing the most recent discussion here: I feel like having a formatter for the project is a higher goal than having a good line length. I understand that there are arguments for sticking with 80 as well as for increasing it. In the end it mostly boils down to personal preference.

Personally, I'm not a fan of setting it to 80—mainly because I think this rule is no longer relevant for modern screens. However I would like to suggest to stick with 80 for this pull request if that means it can be merged earlier. We can still discuss changing the length in a separate issue afterwards.

Throwing in some arbitrary numbers here:

```shell
$ find include/ -name '*.h' -exec awk '{ print length($0) }' {} \; | wc -l
30486

$ find include/ -name '*.h' -exec awk '{ print length($0) }' {} \; | grep -P '(9.|1..|2..)' | wc -l
198

$ find src/ -name '*.cpp' -exec awk '{ print length($0) }' {} \; | wc -l
72010

$ find src/ -name '*.cpp' -exec awk '{ print length($0) }' {} \; | grep -P '(9.|1..|2..)' | wc -l
715
```

So roughly 1000 lines out of 102000 lines exceed the 80 character limit. That is less than 1%. In my opinion it's ok to change them now and potentially change them back later. If only `QObject::connect` are a problem, maybe we can even define an exception for those lines.

JohannesLorenz:
@winniehell Thanks for your research. If it's just 1%, I think we could easily go with a length of 80. On the other hand, I wonder who would be annoyed about 1% of long lines.

To mention a few other C++ coding styles:
* LLVM: 80
* Google: 80
* Chromium: 80
* Mozilla: 80
* Qt: [100](https://wiki.qt.io/Qt_Coding_Style) (?)

Anyways, as there's no clear opinion in this PR, I'll open a poll. I'll take the result, that's it.

JohannesLorenz:
At all developers: Can you please take a minute to complete [this poll about linelength](https://www.surveymonkey.de/r/GTHM2BB)? Thanks on advance.

michaelgregorius:
>At all developers: Can you please take a minute to complete this poll about linelength? Thanks on advance.

Done. I have voted "None, regular" by the way. As already mentioned by @winniehell modern screens are wide and I think this real estate should be used. If a developer decides to use one or more unreasonable long line that cannot be viewed on most screens without scrolling then this will most likely be caught in a code review anyway.

JohannesLorenz:
Survey results:
* don't care: 1
* 80, regular contributions: 1
* 80, occasional contributions: 2
* 100, regular: 1
* 100, occasional: 1
* 120, occasional: 2
* none, regular: 1
* none, occasional: 2
* none, not: 1

The vote is very close. We could interpret it as a close winning for "no line limit". However, since there were a few votings on 100 and 120, we could say it's more 80 or 100.

How should we interpret it? Or should we start a further voting just between 80 and "none"?

tresf:
> How should we interpret it?

.. because the poll was too complex.

* 🎉 : Fixed column length
* 🚀 : Unlimited column length
* 👀 : Stop making me answer polls, I don't care

Then go from there.  If you want to get nitpicking like 80 vs. 100 vs. 120 that already assumes someone wants a max length to begin with, diluting the results.

.clang-format
```diff
@@ -0,0 +1,65 @@
...
+  - Regex:           '^<[^>]+>$'
+    Priority:        1
+  # non-system headers
+  - Regex:           '.*'
+    Priority:        2
+SortIncludes:    true
+
+# Spaces
+SpaceBeforeAssignmentOperators: true
+SpaceBeforeParens: ControlStatements
```
Sawuare: I would change the key here to `Never`, as `ControlStatements` keywords are generally syntax highlighted and easily distinguishable without a space before them.

Spekular:
To quote tresf (from memory): "mobile makes that close button too easy to hit".

Sorry about the accidental close.

.clang-format
JohannesLorenz: I could agree, but our coding conventions enforce this. This PR only tries to follow the coding conventions.

You'd have to ask in discord (or an issue) to change the coding conventions first.

JohannesLorenz:
> My concern with the line length limit is that it can never be broken. A warning of some sort when you pass 80/100 chars seems fine, but disallowing it completely feels like a bit much.

True, but on the other hand, setting the line length to 120 will disallow using 80 completely. `clang-format` overwrites *everything*, so it will convert stuff like a long constructor list to ugly blocks of 120 length, no matter how well you consider formatting it in your PR.

> does anyone feel the need to write unbroken lines over 120 chars? At that point the line is wider than half of my screen in Atom, and i've even lowered the font height from 14 to 13.

Agreed. If I use my favorite font size on a 15,4'' laptop 120 is the absolute maximum that fits in QtCreator. Then you have to scroll through QtCreator on a laptop without a mouse (!). Any further wishes ignored for line lengths > 120. A sane argument outplays any votings, as tresf would say. :smile: 

I'll bash in another voting between
* 80 (:eyes:)
* 100 (:tada:)
* 120 (:rocket:)

The numerical mean might be taken. Or we just find another good argument that will kill all votings.

Spekular:
More and more it sounds like clang-formatting for line length sucks. Can it be disabled completelt?

JohannesLorenz:
@Spekular oh, I just saw [a stackoverflow post](https://stackoverflow.com/questions/47683910/can-you-set-clang-formats-line-length) that `ColumnLimit: 0` preserves the input breaking decisions. That sounds cool, it would reduce the number of changed LOC a lot. I'll try it out.

tresf:
The teamwork is heartwarming. ❤️

No limit is awesome, it allows the devs to decide without the nasty side-effects.

.clang-format
Sawuare: I see this convention follows the Qt coding style, so I'm fine with it as is.

JohannesLorenz:
OK, `ColumnLimit: 0` indeed preserves *most* line break decisions. E.g. the constructor initializer lists are not broken anymore. As this option has the least changes to the code base, and as it allows developers to choose their line breaks, I'd say we're taking this now, unless anyone disagrees.

JohannesLorenz:
While we have it unlimited in `clang-format` now, what should be in the coding conventions? It currently says "try to limit to 80". Should we increase that number?

tresf:
I propose 100 as our new recommended value.  Linus proposed 120 for the Linux kernel, but I think it's a bit too long... 

Here's a very basic example of 120, 100, 80.

```cpp
/**
 * === 120 ===
 */
int mySuperLongFunctionDeclaration(double myFirstParam, double mySecondParam, double myThirdParam, double myFourthParam)
{
    // my function body
}

/**
 * == 100 ==
 */
int mySuperLongFunctionDeclaration(double myFirstParam, double mySecondParam, double myThirdParam,
                                double myFourthParam)
{
    // my function body
}


/**
 * === 80 ===
 */
int mySuperLongFunctionDeclaration(double myFirstParam, double mySecondParam,
                                double myThirdParam, double myFourthParam)
    // my function body
```

he29-net:
I think GitHub is not the best benchmark to judge code readability. If you have the browser maximized, there is more of white space on the sides than text in the middle. Not saying it is bad, it is just made for human languages and these are easier to read when lines are kept short.

But text editors and IDEs should offer far more useful space -- and I have a hard time imagining how many widgets and gadgets would have to be around the text window to squish it under 120 characters in width.

I personally just use gVim for most programming and I can easily fit two 120-character windows side-by-side with around 15 characters to spare (depending on the font size). When limiting line length to 80 or 100 characters, it simply results in a lot of unused blank space that could be used to keep the code more readable instead of forcing line breaks in awkward spots.
(For illustration: left = 122, top-right = 81, bottom-right = 96)
editors

You could of course argue that three windows with 80 characters each would be even better, but then the window management gets a bit messy.

It would be interesting to see how many columns are visible in other commonly used IDEs, as that's where most of the code will be viewed and that's where both readability and line length matters the most.

tresf:
> I think GitHub is not the best benchmark to judge code readability. If you have the browser maximized, there is more of white space on the sides than text in the middle. Not saying it is bad, it is just made for human languages and these are easier to read when lines are kept short.

I actually proposed `100` based on  my zoom level in Visual Studio Code on a 13" display.  GitHub happens to be close-enough that I found it better than posting screenshots.

tresf:
> I personally just use gVim for most programming 

I use `vim` for most of my LMMS coding because other editors muck up formatting without me asking 🤦‍♂.  That said, for larger code-bases like LMMS where multiple files are being modified, I often choose to use an IDE to help find mistakes and make refactoring easier.  In most IDEs the screen real estate is shared with other components (Eclipse, QtCreator, CLion, Visual Studio, etc).  That said, IntellIJ with a sidebar comfortably fits 130 characters on the same screen with the project explorer open.

tresf:
> I think GitHub is not the best benchmark to judge code readability.

Keep in mind that it's currently the preferred method for code review.

he29-net:
> on a 13" display

Hmm, you are right, I did not think about physically smaller displays I almost never use a "bare" laptop without external LCD attached so it did not even cross my mind.

I tried the same window setup as before and even after lowering the font size to the smallest font size that seemed reasonable, two 100 character windows side-by-side can only barely fit. So in that respect the 100 character limit may be indeed better.

tresf:
> > I think GitHub is not the best benchmark to judge code readability.

> Keep in mind that it's currently the preferred method for code review.

If I'm counting correctly, GitHub uses `127`.

he29-net:
> Keep in mind that it's currently the preferred method for code review.

Good point. I just tried to count it and got only 115. Maybe it self-adjusts to some extend based on the longest line in the displayed diff?

tresf:
> Good point. I just tried to count it and got only 115. Maybe it self-adjusts to some extend based on the longest line in the displayed diff?

gist.github.com seems to be in an identity crisis... it uses 123 on my display for typing, but when I save the file it's closer to 127/128 like GitHub proper.

https://gist.github.com/tresf/a3b2c0c82f25afa24b43fbbc020f2065

he29-net:
Oh, I was looking at the diff view, which has one more column for line numbers in the beginning; that's probably why I only count 115 (maybe 116).

tresf:
> Oh, I was looking at the diff view, which has one more column for line numbers in the beginning; that's probably why I only count 115 (maybe 116).

On gist.github.com, there seems to be a 6 character difference with the diff view when it comes to line wrapping.

https://gist.github.com/tresf/a3b2c0c82f25afa24b43fbbc020f2065/revisions#diff-39859e34161f0cb6f5f50974282a2593

tresf:
I'm counting with Visual Studio Code which incorrectly starts the cursor off at character 1 (not zero), making all values off by one.  Fixed.

At this point, I think either 100 or 120 would be good values.  Perhaps we take a poll between the two?

Here's what `vim` looks like on that same 13" display BTW... 
image

JohannesLorenz:
OK, can everyone please participate in this further poll?

While our clang-format will allow line lengths of any size, we ask for a recommendation. This will go into the wiki (coding conventions), but not into clang-format (and you *can* set your editor to use that, but it is voluntary).

What line length shall we recommend?
* :eyes: 100
* :rocket: 120

JohannesLorenz:
The poll is finished. It's also now in our [coding conventions](https://github.com/LMMS/lmms/wiki/Coding-conventions), thanks to @he29-net .

As for the rest of this PR (testing `clang-format` and `clang-tidy`), like mentioned, I won't continue until stable-1.2 comes to an end.

JohannesLorenz:
Due to a bug in clang-format, it will not allow our current style to break CTOR initializer lists after the comma: using `AfterColon` or `AfterComma` for `BreakConstructorInitializers` results in

```
	Model(parent, displayName, defaultConstructed),
	m_scaleType(Linear), m_minValue(min), m_maxValue(max), m_step(step), m_range(max - min), m_centerValue(m_minValue),
	m_valueChanged(false), m_setValueDepth(0), m_hasStrictStepSize(false), m_controllerConnection(NULL),
	m_valueBuffer(static_cast<int>(Engine::mixer()->framesPerPeriod())), m_lastUpdatedPeriod(-1),
```

If we want to keep one member per line in CTOR initializers (currently the case in our code), there are only two ways:

1. We can use `// clang-format off` and `// clang-format on` around the CTOR initializers. Then, they won't be touched. But this also means that useful rules (like no spaces inside parentheses) will not apply to the whole CTOR initializer list.
2. Tell clang-format to break before the comma. Looks a bit weird to me, but still better than stuffing all in one line (or multiple if the line length is reached)
```
	: Model(parent, displayName, defaultConstructed)
	, m_scaleType(Linear)
	, m_minValue(min)
	, m_maxValue(max)
	, m_step(step)
	, m_range(max - min)
	, m_centerValue(m_minValue)
	, m_valueChanged(false)
	, m_setValueDepth(0)
	, m_hasStrictStepSize(false)
	, m_controllerConnection(NULL)
	, m_valueBuffer(static_cast<int>(Engine::mixer()->framesPerPeriod()))
	, m_lastUpdatedPeriod(-1)
	, m_hasSampleExactData(false)
```

Honestly, I'm for solution 2, or at least 1. I think stuffing it all in one line (like in the first snippet) is unreadable: Can you spot quickly e.g. where m_range is initialized there?

What's your opinion?

lukas-w:
> Ignore this whole post, I was wrong

I'd still vote for breaking before the comma.

JohannesLorenz:
One open point here is still how clang-format is actually being applied...

1. The coder itself must run it locally (but then, some will forget it, and some can not even install `clang-format` or `clang-tidy`)
2. The CI will simply reject ill-formatted code (is there a CI plugin?, also, users who can't get `clang-format` or `clang-tidy` run on their systems will spend days to fix everything until the CI is happy)
3. The CI (or a bot?) will accept all code, but auto-format it (is there a CI plugin or a bot?)

JohannesLorenz:
Thanks to @curlymorphic for a recommendation for a [CI action](https://github.com/DoozyX/clang-format-lint-action) that runs on every commit, server side it produces a second copy of the updated files, runs clang-format then compares the commit to the second copy. This might be a good approach for CI.

JohannesLorenz:
I want to start a voting, as this topic had different opinions previously.

Currently, our code style for CTOR initializer lists is:
```
	: Model(parent, displayName, defaultConstructed),
	m_scaleType(Linear),
	m_minValue(min),
	m_maxValue(max),
	m_step(step),
	m_range(max - min),
	m_centerValue(m_minValue),
	m_valueChanged(false),
	m_setValueDepth(0),
	m_hasStrictStepSize(false),
	m_controllerConnection(NULL),
	m_valueBuffer(static_cast<int>(Engine::mixer()->framesPerPeriod())),
	m_lastUpdatedPeriod(-1),
	m_hasSampleExactData(false),
```

Due to a bug in clang-format, it will not allow our current style to break CTOR initializer lists after the comma: using `AfterColon` or `AfterComma` for `BreakConstructorInitializers` results in (almost) only using newlines when the maximum recommended line length (e.g. 120) is reached:

```
	Model(parent, displayName, defaultConstructed),
	m_scaleType(Linear), m_minValue(min), m_maxValue(max), m_step(step), m_range(max - min), m_centerValue(m_minValue),
	m_valueChanged(false), m_setValueDepth(0), m_hasStrictStepSize(false), m_controllerConnection(NULL),
	m_valueBuffer(static_cast<int>(Engine::mixer()->framesPerPeriod())), m_lastUpdatedPeriod(-1),
```

If we want to keep one member per line in CTOR initializers, there are only two reasonable ways:

1. :heart: Use `ColumnLimit: 0`. It tells clang-format to not touch line break decisions of humans. Advantage: most of our code will stay as it is. Disadvantage: If old code/new PRs contain ugly line break decisions, clang-format will not correct them.
2. :rocket: Tell clang-format to break before the comma. May look a bit strange, but it's the only *working* automatic formatting that clang-format currently supports that does not stuff all initializers into one line:
```
	: Model(parent, displayName, defaultConstructed)
	, m_scaleType(Linear)
	, m_minValue(min)
	, m_maxValue(max)
	, m_step(step)
	, m_range(max - min)
	, m_centerValue(m_minValue)
	, m_valueChanged(false)
	, m_setValueDepth(0)
	, m_hasStrictStepSize(false)
	, m_controllerConnection(NULL)
	, m_valueBuffer(static_cast<int>(Engine::mixer()->framesPerPeriod()))
	, m_lastUpdatedPeriod(-1)
	, m_hasSampleExactData(false)
```

Please add a reaction for :heart: or :rocket: for your vote.

.clang-format
```diff
@@ -0,0 +1,65 @@
+---
+# Language
+Language:        Cpp
+Standard:        Cpp11 # newer ones are not supported currently
```
PhysSong, line 4: We now use C++14.

JohannesLorenz: I wrote the comment because all other don't seem to work with current versions of clang-format (11.0.0).

JohannesLorenz:
Watch clang-format [in action](https://github.com/JohannesLorenz/lmms/actions/runs/523249430) :smiley: 

So it's already detecting issues. Next I'll find out how to let this action commit the changes.

If you haven't already, please participate in [last week's survey about CTOR initializers](https://github.com/LMMS/lmms/pull/4690#issuecomment-766058959).

tresf:
>  let this action commit the changes.

I don't know what consensus is, but I believe the developer needs to do this task, or at least accept the proposed change.   I understand there's a sweeping change that needs to occur, but I don't think we should be changing PR code without permission.  The other projects I use just issue warnings, they don't attempt to fix them for you.  We might be able to make this experience a bit better with code recommendations, but I'd strongly discourage abusing the project's access to submitter's PR branches as a default action.

JohannesLorenz:
@tresf I somehow agree. Imagine you push a commit, then want to commit another commit and get push-reject, because the bot pushed in between. You can't even revert the bot's change, because the reverting force-push would trigger the bot again.

What possibilities do we have?

1. As discussed, let the bot push onto the PR (probably not good).
2. Let the bot commit on a branch like "BRANCHNAME-bot-fixes-DATE". The user can choose to merge this later, but this can cause branch pollution.
3. Let the bot just say "Hey, your code is not good, you can let me fix it later".
    - Offer something like a button in the PR "push to add formatting commit" (how?)
    - Maybe reject "not good" PRs (how?)
4. Let the bot only run before merges (how?). This saves energy :deciduous_tree: 
5. More ideas ...?

tresf:
Since GitHub allow code recommendations, that might be good if the actions support it.  That's the most direct way, you can choose to accept the bot's changes, or choose to ignore them.

If actions aren't available, that's still OK, just need to do the change manually, which is how other projects do it.

tresf:
Here's how Homebrew does it (just shows a CI error) https://github.com/Homebrew/homebrew-core/runs/1487869171

JohannesLorenz:
If someone will commit changes which contain 1000 lines of ill-formatted code, single recommendations will be too disturbing.

tresf:
> If someone will commit changes which contain 1000 lines of ill-formatted code, single recommendations will be too disturbing.

Well, a threshold might be warranted but I'm not sure that it's that big of a deal.  These are programmers afterall, they'll see the pattern and fix globally.  These are the same people that are ok with 1,000 errors from the compiler because of a missing semicolon.

IanCaio:
I'm not in favor of committing directly to the PR neither, not only because it can cause that push-reject issue, but we should be able to double check the clang suggestions before applying.

So I'd either add them as requested changes if possible or show it as a failed job with an option to check the details of which lines were at fault. We can add the threshold in number of suggestions as @tresf mentioned, but I think if someone submits a PR with 1000 lines of unformatted code the number of suggestions is not the biggest concern :laughing:

PhysSong:
I suggest uploading the generated diff as an artifact and applying that right before merging, for some cases like the contributor breaks the formatting again and again.

curlymorphic:
I would be in favour of just having the formatting test fail, and to be fixed by the author. I don't feel it is unreasonable to ask a contributor to install clang-format, as they will have already had to install the current toolchain and a reasonable number of dependencies.

he29-net:
Dom mentioned this in one of my PRs; is it something we want to do, and can `clang-*` do something about it?
>Up for debate. Coding conventions say "group #includes by type", but it's not clear exactly what this means. I personally like to keep standard library headers separate from Qt headers.

(https://github.com/LMMS/lmms/pull/5865/files#r562306504)

It seems too specific to do automatically, so I assume `clang-*` may not have a solution for that. But it would be nice to make use of the PR freeze to sort it out either way. After the refactor it will become harder again to touch all files (although merge conflicts in includes would be easy to fix, so it's not a big issue).

JohannesLorenz:
> It seems too specific to do automatically, so I assume `clang-*` may not have a solution for that.

It can do it if we configure it right. Currently it just does this:
```
# Includes
IncludeBlocks:   Regroup
IncludeCategories:
  # the "main header" implicitly gets priority 0
  # system headers
  - Regex:           '^<[^>]+>$'
    Priority:        1
  # non-system headers
  - Regex:           '.*'
    Priority:        2
SortIncludes:    true
```

So we can separate Qt headers from std headers, using regular experssions. However, I wonder what we would do then with headers that don't fall into either category.

Veratil:
> So we can separate Qt headers from std headers, using regular experssions. However, I wonder what we would do then with headers that don't fall into either category.

They would be the least priority, matching the non-system headers.

I think the order we should have is:
1. main header (file.cpp's file.h)
2. first party system headers (cstdlib, etc)
3. third party system headers (Qt, other libs)
4. local headers (lmms headers, etc)

Separating the Qt headers I don't think would conflict with any first party headers since the Q is capitalized.

JohannesLorenz:
> first party system headers (cstdlib, etc)

Any idea how to match those with regular expressions? Think of `<lv2.h>`, `<string.h>`, `<cstring>`, `<samplerate.h>`, `<SDL2/SDL.h>`...

Veratil:
> 
> 
> > 1. first party system headers (cstdlib, etc)
> 
> @Veratil Any idea how to match those with regular expressions? Think of `<lv2.h>`, `<string.h>`, `<cstring>`, `<samplerate.h>`, `<SDL2/SDL.h>`...

lv2, samplerate, sdl2 are third party, string and cstring are first party... unfortunately it's not easy with regex :(

My suggestion is the ideal order though. If we can't get that exactly then it should still be alright. We can get close.

```
^<[^Q>][^>]+>$  // any non-Q leading header name
^<Q[^>]+>$      // all Qt headers start with Q (or should)
```

JohannesLorenz:
I now made a testing branch that shows how it will look like: #5948 . It would be nice if some of you look over the changes that `clang-format made`, to check if you are happy with them. If you have branches that consist of only one commit, you can (hopefully) easily rebase them on that branch (instructions in that PR).

Sawuare:
I think `AlignEscapedNewlines` with `Left` (instead of `DontAlign`) would make it easier to read multiline macro definitions.

Here are examples from modified LMMS code:

`DontAlign`
```cpp
#define MAKE_KNOB(name, x, y, model, label, hint, unit) \
    Knob* name = new Knob(knobBright_26, this); \
    name->move(x, y); \
    name->setModel(&controls-> model); \
    name->setLabel(label); \
    name->setHintText(hint, unit)
```
`Left`
```cpp
#define MAKE_KNOB(name, x, y, model, label, hint, unit) \
    Knob* name = new Knob(knobBright_26, this);         \
    name->move(x, y);                                   \
    name->setModel(&controls-> model);                  \
    name->setLabel(label);                              \
    name->setHintText(hint, unit)
```

I reacted with 👍 to start a vote.

tresf:
> I think `AlignEscapedNewlines` with `Left` (instead of `DontAlign`) would make it easier to read multiline macro definitions.
> 
> Here are examples from modified LMMS code:
> 
> `DontAlign`
> 
> ```c++
> #define MAKE_KNOB(name, x, y, model, label, hint, unit) \
>     Knob* name = new Knob(knobBright_26, this); \
>     name->move(x, y); \
>     name->setModel(&controls-> model); \
>     name->setLabel(label); \
>     name->setHintText(hint, unit)
> ```
> 
> `Left`
> 
> ```c++
> #define MAKE_KNOB(name, x, y, model, label, hint, unit) \
>     Knob* name = new Knob(knobBright_26, this);         \
>     name->move(x, y);                                   \
>     name->setModel(&controls-> model);                  \
>     name->setLabel(label);                              \
>     name->setHintText(hint, unit)
> ```
> 
> I reacted with 👍 to start a vote.

Please don't do this.  Letterboxing with whitespace is a nighmare.  We're not trying to win a code beauty competition here.  Function over fashion.

tresf:
> Letterboxing with whitespace is a nightmare

Case and point.... see the diff this rubbish creates because of `someLongerVariable`


```diff
- #define MAKE_KNOB(name, x, y, model, label, hint, unit) \
-    Knob* name = new Knob(knobBright_26, this);         \
-     name->move(x, y);                                   \
-     name->setModel(&controls-> model);                  \
-     name->setLabel(label);                              \
+ #define MAKE_KNOB(name, x, y, model, label, hint, someLongerVariable) \
+     Knob* name = new Knob(knobBright_26, this);                       \
+     name->move(x, y);                                                 \
+     name->setModel(&controls-> model);                                \
+     name->setLabel(label);                                            \
name->setHintText(hint, unit)

**Comment by Sawuare** [comment]
> see the diff this rubbish creates because of `someLongerVariable`

That's a great point. Thank you for thinking ahead, @tresf.

I saw that the glibc source avoids such scenarios by aligning the backslashes to the right-most column.
The option `Right` does that but I don't think it would work here since there's no column limit. Even if it worked it would use too many spaces.

I think that settles it for the current option `DontAlign`.

**Comment by JohannesLorenz** [comment]
Open question from the OP:

* CMake integration: multiple possibilities:
  - No CMake integration
    * Users must use clang-tidy from their IDE, if they use an IDE supporting it,
      otherwise they have to run clang-tidy manually, passing command line args to it
    * No Continous integation!
  - Let CMake only [export compile commands](https://cmake.org/cmake/help/v3.5/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html)
    * No full automation
  - Always run clang-tidy at compile, if a flag is set
    * Compile time overhead!
    * The need to set that CMake flag may confuse users
  - Add a target for CMakeLists to fix (and alternatively just to scan) all source files
    * Disadvantage: Can only accept all fixes, where in an IDE you can select what to fix

What's your opinion?

**Comment by Veratil** [comment]
Automation is usually the easiest route. We'd need to make it clear that it's running so anyone doesn't wonder why their code (possibly) changed from what they originally wrote though.

**Comment by he29-net** [comment]
Having a `cmake` flag available to set clang-tidy to run at compile time would be nice either way I think. People who do not like the overhead or don't want to set the flag can still run it manually (or set it up in their IDE), so the first and third options should not be mutually exclusive.

In CI, I assume it would do scan-only (by utilizing the `cmake` flag?), and reject the PR if it finds issues. PR authors should read about the coding style before making a contribution, and it would tell them to either set the `cmake` flag or to run clang-tidy / clang-format manually to easily meet the requirements. The CI job could also remind them ("Did you forget to run clang-*?") to make it clear why it failed.

I can't really comment on the "export-only" variant since I don't quite understand what it means (but it seems like a complication compared to a standard `cmake .. && make`?). And the last option seems to have not enough control: what if some user updates their clang-* and it insists on fixing some newly detected issue that is not yet fixed in the LMMS master branch? It would force them to include the fix in their PR, even if it is completely unrelated to their changes. So a way to exclude files from the check / formatting could be quite useful..

**Comment by JohannesLorenz** [comment]
> Automation is usually the easiest route. We'd need to make it clear that it's running so anyone doesn't wonder why their code (possibly) changed from what they originally wrote though.

FWIR, we agreed (in discord) on not modifying anything in the PR.

Personally, I think it would be OK if the bot creates a branch like "<branchname>-bot" (and removes that branch after the merge). This allows the user to fetch the bot changes if they want, but it does not modify the PR. What about that?

**Comment by JohannesLorenz** [comment]
> Having a `cmake` flag available to set clang-tidy to run at compile time would be nice either way I think. People who do not like the overhead or don't want to set the flag can still run it manually (or set it up in their IDE), so the first and third options should not be mutually exclusive.

I would probably not like that too much. Changing the `cmake` config each time seems annoying to me. Maybe I'll compile and wonder why it takes so long, and only after a while realize that I have `clang-tidy` enabled while compiling...

> I can't really comment on the "export-only" variant since I don't quite understand what it means (but it seems like a complication compared to a standard `cmake .. && make`?).

The idea is that `clang-tidy` must know how each source file is compiled (which defines and which includes are used). So instead of running it directly from the `cmake` call, we let `cmake` just output one large file with compile commands for each source file. Then, we let `clang-tidy` use that file in a second step.


> And the last option seems to have not enough control: what if some user updates their clang-* and it insists on fixing some newly detected issue that is not yet fixed in the LMMS master branch? It would force them to include the fix in their PR, even if it is completely unrelated to their changes. So a way to exclude files from the check / formatting could be quite useful..

Agreed. That's probably why I suggested in the previous post to let the CI do all and just pull it: The CI will not update their clang-format version (as long as we don't want that).

**Comment by tresf** [comment]
> Personally, I think it would be OK if the bot creates a branch like "-bot" (and removes that branch after the merge). This allows the user to fetch the bot changes if they want, but it does not modify the PR. What about that?

I'd rather not pollute the branches (we can't create them for forks and this could quickly pollute our own).  If possible, scripting code change recommendations is best, then they can just accept or ignore the recommendation, but without any intermediary branches.

**Comment by JohannesLorenz** [comment]
> I'd rather not pollute the branches (we can't create them for forks and this could quickly pollute our own). If possible, scripting code change recommendations is best, then they can just accept or ignore the recommendation, but without any intermediary branches.

We could simply create a separate LMMS fork, like "LMMS/bot-suggestions". Then our "LMMS/lmms" fork will not be polluted, and like I mentioned, I'd let the bot remove its branches after the original branches have been merged anyways. So I don't think this would end up in pollution.

**Comment by he29-net** [comment]
> Maybe I'll compile and wonder why it takes so long, and only after a while realize that I have clang-tidy enabled while compiling...

How much extra time it takes, roughly? Although I agree that doing repeatedly something that needs to be run only once before the final submission may not be very efficient use of CPU cycles regardless.

> The idea is that clang-tidy must know how each source file is compiled (which defines and which includes are used). [...] Then, we let clang-tidy use that file in a second step.

The export-only option could be OK as well in that case, if it just saves data for a subsequent manual clang-tidy run. Either way seems better to me than making and managing a bunch of new branches "in the cloud" for something that could be easily done locally.

The cloud solution could also make `git blame` less useful: imagine you change three lines in a file made by somebody else and they are all touched by the formatting bot. You pull the changes and now the bot is the author of these three lines. When I see these lines and I want to ask the author about them, I would have to figure out which PR it came from, since the name of the true author was lost. And that again may be hard, since the commit message probably won't contain the PR number, and the bot branch itself may be long deleted.

**Comment by JohannesLorenz** [comment]
> > Maybe I'll compile and wonder why it takes so long, and only after a while realize that I have clang-tidy enabled while compiling...
> 
> How much extra time it takes, roughly? Although I agree that doing repeatedly something that needs to be run only once before the final submission may not be very efficient use of CPU cycles regardless.

Interestingly, now only 9 seconds (on a normal desktop CPU). When I ran it the first times with tons of changes, it took multiple minutes for a whole run (2 or 3 seconds per file). So runtime seems to be no issue anymore (contrary to what I first wrote/thought). So we could integrate it into cmake.

> The cloud solution could also make `git blame` less useful.

I think the question is how it looks on master. Either the whole branch is a "feature; fix; fix; fix", in which case you just squash into one commit, which you author (most cases in LMMS). Or, the branch commit history is good enough for not being squashed - in this case, I'd expect the developer to not keep "fixed format" commits on the branch, but squash them into the commits where they really belong (e.g. using `git rebase -i` and using `f` in the rebase editor). So on master, I'd expect no bot commits in either case.

**Comment by tresf** [comment]
If a developer changes a line, that's the diff they expect to see.  Changing what they changed again sounds like a cat chasing its tail.  Code formatting can be yelled at by the CI without adding pandemonium to an otherwise predictable coding workflow.

**Comment by JohannesLorenz** [comment]
Correcting my own statement:

> Interestingly, now only 9 seconds (on a normal desktop CPU). When I ran it the first times with tons of changes, it took multiple minutes for a whole run (2 or 3 seconds per file). So runtime seems to be no issue anymore (contrary to what I first wrote/thought). So we could integrate it into cmake.

Those 9 seconds were for clang-format. For clang-tidy, it is still very slow (~ as slow as for a normal compile, so it doubles compiling time). So IMO a flag to activate it would be not so good.

**Comment by JohannesLorenz** [comment]
> If a developer changes a line, that's the diff they expect to see. Changing what they changed again sounds like a cat chasing its tail. Code formatting can be yelled at by the CI without adding pandemonium to an otherwise predictable coding workflow.

Sorry, I can't follow what you mean. Of course, if they submit changes that are against our coding conventions, the bot *must* submit changes to their changes.

**Comment by tresf** [comment]
> Sorry, I can't follow what you mean. 

Don't touch other's code without their approval (or abandonment).  Ever.  It's a mistake.

> the bot must submit changes to their changes.

The bot MUST do no such thing.  Good code is more important than formatted code.  The CI can yell at them and they can fix it manually.  If you get creative, make code recommendations using GitHub comments.  Anything that changes a PR without permission is a mistake.  Creating rogue branches is a mistake.  Don't make code formatting more important than the code itself.

**Comment by JohannesLorenz** [comment]
You mean the CI gives the author hints, and they *can* fix it, but are not forced to take it? Do you propose that the reviewer would still check that the relevant bot hints were used (which means manual review again, which we wanted to avoid?). Or should the reviewer then just say "Author, I trust you did the formatting well"?

Another problem with this is: If you do changes A, B and C, deny clang-format's changes and bring it on master. Then the next coder who wants to do change D will be confronted with issues from A, B, C and D. In the end, the whole file will be marked with errors from `clang-*`, and we get more and more away from `clang-*`'s suggestions. Of course, you can still manually pass the line numbers to `clang-*` (this is another parameter you may specify), but if even your changed lines are already full of previous errors... Also, specifying the line numbers you changed can get difficult if you changed lines in 10 places of a file - [there are already scripts that feed `clang-*` with the `git diff`](https://gist.github.com/AtnNn/10d17408516ae408788484418740f18b), but do we want to use scripts for that? (such script should work on Windows, too...)

Summing it up, I'm for either use `clang-*` and always keep all files 100% consistent, or to not use `clang-*` at all. But what do the other coders say?

**Comment by he29-net** [comment]
From my point of view, there are two key "big picture" requirements that out `clang-*` deployment should meet:

1) Make it mandatory.
  Apart from "making code look nice", eliminating manual code style checking from the review process is one of the main goals. If there are any `clang-*` issues left in the LMMS code base, the CI job would be pretty useless, since it would just fail every time (or someone would have to keep managing the exceptions).
  That means it would still require attention from reviewers, and a quick look on the PR tracker shows that their attention would be better focused elsewhere. So I definitely support the "no exceptions"\* idea; either the PR meets the formatting requirements or it does not get merged. I'm sure that having good _and_ formatted code at the same time isn't exclusive. 
\* perhaps excluding `plugins` or `3rd-party` folders, since forcing our style on these may conflict with upstream

2) Make it easy to use.
  In order to make it a mandatory requirement for PR acceptance, it _must_ be also easy to use. Shifting some of the work from reviewers to CI and PR authors is good, but some people have already trouble compiling LMMS for the first time (mainly on Windows), so adding another obstacle for the newcomers would not be helpful. This should also include multiple ways to achieve conformity: simple manual solutions for casual devs who are submitting just a small 5 line patch for the first time, and advanced options for devs that want more automation and are willing to invest some time up front to set things up.

  Going deeper into 2): For me, the "cloud bot approach" (create a branch with suggested changes that the PR author can use) seems complicated and messy, but for some devs it may be preferred. Others like me could prefer an option that lets me easily run `clang-*` locally. So why not support both, so that everyone can find something that works best for them?

Both "cloud" and "local" solutions may be still kind of difficult for some people, though: one requires relatively advanced git knowledge (I did not even know the "interactive rebase" existed before Johannes told me last year), the other requires extra tools that Windows users may need to set up.
  So as a fallback, it should be also easy to fix problems manually by editing the code. That's why I think a simple CI job that says "hey, your code failed the formatting check, here is a log that says what's wrong" may be more useful in this case than a new branch appearing somewhere in a different repo, where the user has to locate the appropriate commit, which then probably contains only the suggested fixes, but no explanation of what was wrong. (But again, the CI output could contain the log _and_ a link to the branch with suggested fixes, these are also not exclusive..)

**Comment by tresf** [comment]
> adding another obstacle for the newcomers would not be helpful.

This is not true.  Automating the "sorry your code doesn't pass our style guidelines" DOES help the user.  Otherwise they're waiting for days, weeks or months for a human to do it.  This notion that early users can't fix formatting or that it's a detriment to notify them by bot is ludicrous.   Our job is NOT to write other people's code and the more we conflate the workflow and take the responsibility off of the coder, the less we expect of our coders.  I'll keep using Homebrew as an example here because they merge hundreds and hundreds of PRs every week and they do NOT make changes for people.  Homebrew is a Ruby project, so we're talking about a lower barrier to entry.  I don't think we should do anything automatically that touches other people's code unless they explicitly request it.  Once code is compiling well, I think developers will be happy to fix formatting errors to get it merged.

> Going deeper into 2): For me, the "cloud bot approach" (create a branch with suggested changes that the PR author can use) seems complicated and messy, but for some devs it may be preferred. Others like me could prefer an option that lets me easily run clang-* locally. So why not support both, so that everyone can find something that works best for them?

Preferred by whom?  I've contributed to several dozen projects, none do code formatting forking nonsense. I really don't understand any use-case where creating forks for formatting is warranted.  Forks should be used for code divergence, not for making sure there's a newline and no spaces after a `{`.

>  the CI output could contain the log and a link to the branch with suggested fixes

Homebrew's errors just cite the rule and I believe sometimes links to examples.

> You mean the CI gives the author hints, and they can fix it, but are not forced to take it?

Yes.

> Do you propose that the reviewer would still check that the relevant bot hints were used

If your workflow requires a passing codestyle, you block merging until it passes.  There should always be a manual override though.

>  the whole file will be marked with errors from clang-*, and we get more and more away from clang-*'s suggestions. 

Ideally, we'd only run checks on stuff that's changed.  We need not make the style check more important than the code itself, no?

> but if even your changed lines are already full of previous errors... Also, specifying the line numbers you changed can get difficult if you changed lines in 10 places of a file - there are already scripts that feed clang-* with the git diff, but do we want to use scripts for that? (such script should work on Windows, too...)

That's the same for any code that's changed and was previously under review.  If a developer chooses to ignore the formatting errors and change lines again, the old code comments should naturally become [outdated](https://github.com/go-gitea/gitea/issues/12523) and collapse by default.

**Comment by tresf** [comment]
> it must be also easy to use

I don't think there's anything "easy" about auto-formatting code unless the IDE's doing it as you type.  Any solution that happens after the code is written (in my opinion) should require human intervention.  This doesn't mean the local scripts wont help, they absolutely will (and the CI can offer some guidance here).  This becomes increasingly helpful for PRs that becomes stale and only formatting it preventing it from being merged.  Any developer can run the format on the code locally and bump the PR with the formatting changes.

>  (such script should work on Windows, too...)

For command-line stuff, Windows may be the oddball out here, but from what I'm reading the tidy tasks [are supported](https://docs.microsoft.com/en-us/cpp/code-quality/clang-tidy?view=msvc-160), at least in the IDEs and will pick up the tidy files automatically.

**Comment by he29-net** [comment]
>>    adding another obstacle for the newcomers would not be helpful.
>
>This is not true. Automating the "sorry your code doesn't pass our style guidelines" DOES help the user. [...]

I am not saying otherwise. As I explained in the post, I agree that having a CI job that rejects non-compliant code is useful, since it saves time of our reviewers. That CI job is not a new obstacle, it just changes who deals with it.

By "adding obstacles" I mean making `clang-*` tools required for compilation (which could be annoying to Windows users) or making the suggestions only available as a diff in some extra bot-generated fork (which could be annoying to anyone who is not a git expert). If I want to submit a short patch that takes 15 minutes to write and test (my first LMMS PR was literally adding a missing `)` to a formula), I don't want to spend two more hours setting up tools that I never used before or learning git dark magic. I just want the CI job to tell me what failed and how can I fix it manually.

>> [fork based approach]
>
>Preferred by whom?

I don't know, I can't speak for others. I do not like it either, but Johannes seemed to prefer it (unless he was just further explaining what the process could look like and I mistakenly understood it as a preference).

> Any solution that happens after the code is written (in my opinion) should require human intervention.

And I agree (and the human should be the PR author, not the reviewer). By "being easy to use" I mean not forcing the user into a specific way to fix the code (which could require complex setup or advanced knowledge), but simply telling them in the CI log what is wrong and how it can be fixed manually, so that literally anyone could do it, even if they just barely managed to even compile LMMS for the first time and fixed a simple bug. I think we actually agree on that point (and most others) and just describe it differently.

**Comment by JohannesLorenz** [comment]
> Preferred by whom?

Me

> nonsense

Can we talk less destructive?

IDEs will help, and IMO we should do checks prior to merging (I think we agreed until here). And if everyone can easily apply `clang-*` on their changes, with an identical result on all machines worldwide, all will be great. However, I fear this will not work:

1. Users will have different versions of the `clang-*` tools
2. `clang-*` will behave different on different systems (due to bugs, CRLF, whatever)
3. Some users will be too unskilled to handle these tools (especially, not every Windows user may easily install the clang-tools, open a cmd and call the `clang-*` tools correctly)

All in all, I still expect regular situations where users tried to submit code and the CI rejects them. I expect this not just 1%, but rather something like 10% or 20%. I think if it can happen regularly, we should be prepared for it. So I, personally, don't want to look at a 100 lines diff, fixing each single indentation issue with my own hands. So you may say that the CI could just put the resulting files zipped somewhere. Now compare: Hitting the download button in your browser for the zip, then unzipping it, carefully copying over all the files into you own git tree, typing a git commit, possibly removing the zip and the folder. Compare this with a simple `git fetch && git merge bot-fixes/mybranchname` (something like this). I think my solution with the bot fixes will be much faster and safer.

Am I too pessimistic about all `clang-*` worldwide will produce the same output and that problems will occur only in exceptional cases?

**Comment by he29-net** [comment]
> Am I too pessimistic about all clang-* worldwide will produce the same output and that problems will occur only in exceptional cases?

I think it is a valid concern, but we would likely not be the only ones hit with such issues, so I doubt that there will be many breaking changes in behavior across versions and platforms that couldn't be fixed by a new config option or something. Big organizations using these tools would probably not be happy if each new version reformatted their whole code-base with not workaround available.

As I noted on Discord, we started using `clang-format` recently at work, and since day zero we had to add an extra option to the config file, which must be commented out on version 7 (since that's the version we mainly use and it does not know the option), and uncommented on version 11-ish and newer, to keep the formatting consistent with the older `clang-format` version. So it does happen, but we could likely just provide an alternative config file for new versions that behave differently.

I don't see any harm in having a bot-fork based option for those who want to use it (currently 33 % of participants in this discussion :)) ), but I would definitely like to see a fully supported option to run `clang-*` locally as well (which is probably also a prerequisite for all the IDE helper things).

**Comment by tresf** [comment]
> CRLF, whatever

CRLF is handled by Git.  We can turn that off project-side using `.gitattributes`.

> Can we talk less destructive?

Rash perhaps, but destructive? I don't understand what's so confusing about having a CI tell someone to fix their code.  These talks of automating this with branches is -- in my opinion -- destructive in a literal sense because it will pollutes and conflate.  Downstream, we lack sufficient access to fork, upstream we'll be making forks for formatting?   I find it a bit hypocritical to think that formatting code is too complex for someone to grasp while merging in an upstream fork is not complex.  One is a matter of following rules, the other requires two remotes and hand-merging or pull requests against forks, neither of which I find to be intuitive.  Sorry if that comes off as "destructive", I just find the workflow to be all around messy.

> I would definitely like to see a fully supported option to run clang-* locally 

I think this is more pragmatically good.

> I don't see any harm in having a bot-fork based option for those who want to use it 

OpenJDK does some pretty creative bot stuff for merging PRs, but it's mostly for project integration,  I guess for stale branches this might help, I just think it'll cause more PRs in the end.  If we have access to push to a stal PR, that's likely the cleaner solution, but I won't fight the bot conversation any longer, I've shared my opinions on it. 🍻 

>  we started using clang-format recently at work, and since day zero we had to add an extra option to the config file, which must be commented out on version 7 (since that's the version we mainly use and it does not know the option), and uncommented on version 11-ish and newer, to keep the formatting consistent with the older clang-format version. So it does happen, but we could likely just provide an alternative config file for new versions that behave differently.

Thanks for sharing.  These are the types of exception scenarios I was trying to discuss on Discord.

**Comment by JohannesLorenz** [comment]
> > I would definitely like to see a fully supported option to run clang-* locally
> 
> I think this is more pragmatically good.

For local support, going through the above options:

```
1. No CMake integration
2. Let CMake only export compile commands
3. Always run clang-tidy at compile, if a flag is set
4. Add a target for CMakeLists to fix (and alternatively just to scan) all source files
```

So I think (1) is not possible, because we will want to do checks at CI side to forbid ill-formed code. (3) and (4) seem to make it impossible to run CMake only on the changed files. However, imagine changing only a header which is included by 100 cpp files. Then there's no way around re-compiling those 100 cpp files with clang-titdy, too (because the definitions may have changed, resulting in different clang-tidy-fixes)? So is this really a problem? (2) seems nice because you can restrict clang-tidy on the files you really changed, but isn't this exactly the danger with the header that requires re-compiling 100 files?

All in all, I think we can even offer 2, 3 and 4 in parallel. We'll probably see what is best in practice and the CI will save us or errors anyways...

.clang-format
PhysSong: From https://releases.llvm.org/11.0.0/tools/clang/docs/ClangFormatStyleOptions.html:
> `Cpp11` is a deprecated alias for `Latest`

This means we should use `c++17` since we again moved to C++17.

Rossmaxx:
isn't this done, why not close it?

PR #4692, opened 2018-11-01
Handle automation on processing thread

### LINKED ISSUE #4543: Lmms crashes often when I automate pitch on vsti's

### LINKED ISSUE #4584: lmms 1.20 r6 crash when lfo controller use with high configaretion ..
lmms 1.20 r6 crash when lfo controller use with high configaretion ..  
i use lfo controller for get some wably or wavy effect... and  its crush ... i think its a bug.

### LINKED ISSUE #4625: The Automation of the Pitch in VeSTige appears broken in RC7
If you try to automate the pitch in vestige with a Automation track, it  makes a nasty unacceptable crackling/static  noise 

this is on a Windows 10 OS PC.

DomClark:
Connections that update values for remote plugins are now all `DirectConnection`s. The result is that changes to these values no longer always come from the main thread, but instead from the thread that changed them, which in the case of automation or controllers will be the processing thread.

This fixes the following issues:
* Hangs occurring when the project is saved while Zyn or VSTs are being automated. Value updates were sometimes sent while LMMS was waiting to receive the save data from the remote plugin, resulting in a deadlock trying to lock the plugin mutex twice on the same thread. Sending the update on a different thread solves this. (Fixes #4543, fixes #4584.)
* Stuttering audio during automation of Zyn or VSTs. Since #4460, the plugin mutex is locked using `tryLock` before getting audio from the remote plugin. If a value was being automated, the main thread was sometimes holding the lock to update the value which caused a buffer of audio to be skipped for the plugin. Sending the automation on the processing thread ensures the updates are finished before audio is rendered. (Fixes #4625.)
* Choppy/laggy/missing VST automation, especially during export. The automation used to get caught up on the main thread (a problem that was much worse on export), but is now delivered immediately on the processing thread. (I can't find an issue for this, but I've seen it mentioned from time to time on the Discord server.)

### DIFF
--- a/include/Model.h
+++ b/include/Model.h
@@ -41,6 +41,10 @@ public:
 		m_displayName( _display_name ),
 		m_defaultConstructed( _default_constructed )
 	{
+#if QT_VERSION < 0x050000
+		connect( this, SIGNAL( dataChanged() ), this,
+			SLOT( thisDataChanged() ), Qt::DirectConnection );
+#endif
 	}
 
 	virtual ~Model()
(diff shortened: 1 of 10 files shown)


DomClark:
I think so, yes.

PR #4696, opened 2018-11-04
Initialize Peak Controller with Base Value fix

DouglasDGI:
This should fix the bug (present in both master and stable 1.2) I accidentally introduced with #4382, the bug being that the Peak Controllers would initialize with a base value of 0.5 after that change, and 0 before that change.  @DomClark luckily pointed out the problem.


PhysSong:
I think this change should be merged into `stable-1.2`. @DouglasDGI Can you create a new branch from `stable-1.2` and redo these commits? If you can't, I can cherry-pick your commits manually.

Issue #4695, opened 2018-11-02
Debian sid integration


jasp00:
I would like to integrate Debian sid support. I would also like to help release version 1.2; I might have some spare time. What is the current workflow? `master` to `stable-1.2`? `stable-1.2` directly?

PhysSong:
@jasp00 Welcome back! We commit stuff for 1.2 into `stable-1.2`, then merge them into `master`.

jasp00:
@lukas-w, what is going on with Qt5? qt5-x11embed is not integrated. Are we going to use it or do we fix Qt directly?

PhysSong:
@jasp00 Can you elaborate it? This repository submodules `qt5-x11embed` which depends on the private component of core module and the X11 extras module.

jasp00:
> the module should be rebuilt on Qt upgrades

This is inconvenient. Either we use this extension or we fix Qt and use `createWindowContainer()`. In any case, the solution should be integrated in Qt.

Also, this submodule structure is troublesome. These sources are not in the tarball and neither are packaged dependencies.

I am speaking from a distro point of view. Let us see how we organize this.

PhysSong:
> or we fix Qt and use `createWindowContainer()`.

IIRC @lukas-w said the `createWindowContainer` implementation worked fine with Qt 5.9.
> Also, this submodule structure is troublesome. These sources are not in the tarball and neither are packaged dependencies.

There was a discussion on it, #4100.

jasp00:
My job is done. I would like to invite @RossGammon, current maintainer in Debian, to join this project and take advantage of this Debian sid integration.

lukas-w:
>  @lukas-w said the `createWindowContainer` implementation worked fine with Qt 5.9.

I think this was false alarm from my side. I works better, but still not as good as `qt5-x11embed` and IIRC it was stated on Qt's bug trakcer that `createWindowContainer` is not intended for our use case.

PR #4519, opened 2018-08-02
Update the LMMS logo and mimetypes

Umcaruje:
This has been a project that was long in the making, started it with Rebecca around the time we worked on the new stable-1.2 theme, but the icons were left unfinished in my dropbox folder, until now. 

The whole logo was redesigned from scratch, to be flat and fit with the new theme better. All the icons were designed to conform to specific pixel sizes: 16x16x, 24x24, 32x32, 48x48, 64x64, 128x128, and I have also rendered them at double the dpi so they stay sharp on hidpi displays.

Here's the application icon in various sizes:
64x64 48x48 32x32 24x24 16x16

Here's the application mimetype in various sizes: 
project-64x64 project-48x48 project-32x32 project-24x24 project-16x16

Also, I reverted back the changes from https://github.com/LMMS/lmms/commit/e84b537a60b603b4987fc9075fce83ae904550b2 to now have proper file association inside windows. We embed two icons inside `lmms.exe` now, one being the app icon, and the other one being the file association icon (mimetype). Also changed the logic inside of the file association code to always rewrite the registry for the new file association entry. A big thank you to @tresf and @PhysSong for helping for this part.

Here are some screenshots on how the mimetype icons look inside of dark and light Arc themes on elementary OS: 

screenshot from 2018-08-02 20 49 36
screenshot from 2018-08-02 20 32 40
screenshot from 2018-08-02 20 33 11
screenshot from 2018-08-02 20 48 23

And some screenshots of the new icons on macOS:

screenshot at aug 01 17-44-21
screenshot at aug 01 17-43-19
screenshot at aug 01 17-42-53

Also, I updated the mainwindow icon to conform to a 16x16 grid, so it's sharp at a small size. a new icon in the theme was created for this called `icon_small`

I can't seem to get my VM to work for building windows artefacts, but Tres offered to create an installer so I'll post the windows screenshots soon enough.

Thank you for reading this lengthy post, let me know if you like the new icons!

PR #4697, opened 2018-11-04
Add initial Debian packaging for stable-1.2

jasp00:
This is the first step towards Debian integration, as stated in #4695.


PhysSong:
Could you consider removing `clang.patch`? We don't add `-Werror` by default anymore. Also, I can't apply the old patch into current `stable-1.2`.
Also, `CONTRIBUTORS` are outdated in your patch.

jasp00:
I did not explain that patches under `debian/patches` are not meant to be applied, since the package format is native.

> Could you consider removing `clang.patch`?

It is only a reminder to add a clang pass on Debian sid.

> We don't add `-Werror` by default anymore.

It is good practice to enable `-Werror` on sid. Specific errors can be disabled.

> Also, `CONTRIBUTORS` are outdated in your patch.

It is just another reminder.

trebmuh:
Suggestion to remove the debian/menu file following [this Debian bug report](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=741573) which have deprecated the menu file.

jasp00:
If there is no more feedback, I will merge this and continue the integration.

PhysSong:
I forgot to mention that `debian/lmms.xpm` should be updated according to #4519.

PR #4699, opened 2018-11-04
Initialize peak controller last sample with base value fix (Stable 1.2)

DouglasDGI:
@PhysSong requested that I make a pull request for this for Stable 1.2 as well, which is probably a good idea.  Sorry if making a separate pull request isn't the best way to do this (I'm still learning Github physics), but I'm assuming it's fine.

This is the same fix as #4696, and fixes the bug introduced in #4382.


DomClark:
Should this be merged?

PR #4707, opened 2018-11-14
Add Debian sid build pass

jasp00:
This request adds one pass to Travis and updates Debian packaging. The extra time is important but less noticeable because of the parallel osx build. I would like to add another pass for i386 and another one for Clang.


tresf:
No objections here.  @lukas-w can you please review as well?

jasp00:
ccache breaks detecting compile features, so it is not an option yet.

PhysSong:
```
ccache: error: Failed to create directory /var/cache/pbuilder/ccache/f/e: Permission denied
```
It turns out the bug is due to the [pbuilder bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=666525). It's fixed in the recent version of pbuilder.

PhysSong:
I have a WIP branch: https://github.com/PhysSong/lmms/commits/debian-pkg

jasp00:
Your changes look fine. Please merge them into this branch.

jasp00:
ccache is enabled. If there are no more issues, I will merge.

PhysSong:
When I tested on my fork, 80776dac3cb0d628386cadf204d779782d72048f seems to make 3cb11db812871aefba5fc13350002441837c6c03 not needed anymore. I also deleted the cache before testing that.

jasp00:
> 80776da seems to make 3cb11db not needed anymore.

I suspected that, but pbuilder 0.229.1 supports build profiles, which may be useful. Since it works, I would rather keep it.

PR #4708, opened 2018-11-15
New pixmap as per #4519

jasp00:
As [noticed](https://github.com/LMMS/lmms/pull/4697#issuecomment-438917419), this is the updated pixmap.


jasp00:
Merging...

Issue #3561, opened 2017-05-18
LMMS  crashes when using VST


Andrew344:
I use LMMS 1.1.3 (64-bit ver.) on Windows 7 64bit, аnd I have problem with VST plugins. 
Neither one doesn't work acceptable.
Watch the video: https://www.youtube.com/watch?v=FxKjCU7Bcao
This happens regardless of instrumemts, and number of track in project.
It can happens even if project is empty. 
I can't see what's error because, the terminal displays nothing.
![default](https://cloud.githubusercontent.com/assets/28777908/26202495/52026ac8-3be0-11e7-84fa-53ed465bc9b7.png)

![default](https://cloud.githubusercontent.com/assets/28777908/26202274/6d64d540-3bdf-11e7-9d21-f880c2f8384a.png)

At first I thought that the reason the memory is full, but judging from the readings of Windows task manager, the amount of used memory doesn't exceed 2.7 GB (4 GB total).
The antivirus program does not affect it too.

Therefore, most probably it's LMMS bug.

Sorry for my broken English.

tresf:
FYI, Unlike Linux, Windows can't show terminal output without recompiling with the `mconsole` flag, which permanently makes `lmms.exe` open a command window on launch.  This is a limitation of the platform.  We will address this in a future release using a logging framework.

In regards to your problem, does it go away with the latest RC build? (At the time of writing this, 1.2.0-rc2)

Andrew344:
Sorry, I forgot to write, in LMMS 1.2 this problem remains.

musikBear:
@Andrew344 
Could you try with a VST that is on the list:
https://lmms.io/wiki/index.php?title=Tested_VSTs#Keys
I recommend MDA-piano, because it is tiny in dl and quite good as a piano emulator.
"Predator" is not tested

Andrew344:
MDA piano works OK. Probably, I downloaded the wrong plug-ins. I'm not going to close the issue now. I will try to use different plugins for two more days, if the problem does not appear, the issue will be closed.

Andrew344:
I will try another plugins.

musikBear:
> MDA-piano works OK

Ok then that *Predator* is not 'liked' by lmms. A few VST instruments and more FX is that way. Its being looked at.
Does the CPU load stays high after loading?
The best VST in lmms is Synth1. You need to install, but both quality and cpu-load is magnificent
[...]

PR #4719, opened 2018-11-30
Allow console output on Windows if available

PhysSong:
As mentioned in https://github.com/LMMS/lmms/issues/3561#issuecomment-302405321, `lmms.exe` doesn't show console outputs by default.
This PR makes the console output work on Windows if launched from the terminal. Qt's logging system like `qWarning` also works with this PR.
This PR doesn't touch redirections. So, one can still use `lmms -d foo.mmpz > foo.mmp`.
The idea is from #4544.

### DIFF
--- a/src/core/main.cpp
+++ b/src/core/main.cpp
@@ -105,6 +105,21 @@ static inline QString baseName( const QString & file )
 }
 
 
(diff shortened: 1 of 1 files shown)


PhysSong:
Merge?

PR #4723, opened 2018-12-09
Fix lost control links in copied automation patterns

### LINKED ISSUE #3359: Copies of  shapes in automation are wrong

### LINKED ISSUE #3781: Control links are lost when duplicating automation clips

PhysSong:
As I said in #3781, I decided to work around the bug temporarily, at least in `stable-1.2`. I believe we can fix underlying issues as well, but the required change is too large for stable releases.
I think the ultimate solution is to decouple journaling and automation(see https://github.com/LMMS/lmms/issues/3781#issuecomment-407301931). I'll create a new issue to keep track on it once this PR gets merged.

Fixes #3781, fixes #3359.

### DIFF
--- a/src/core/AutomationPattern.cpp
+++ b/src/core/AutomationPattern.cpp
@@ -791,7 +791,7 @@ void AutomationPattern::resolveAllIDs()
 						{
 							// FIXME: Remove this block once the automation system gets fixed
 							// This is a temporary fix for https://github.com/LMMS/lmms/issues/3781
-							o = Engine::projectJournal()->journallingObject( *k + ( 1 << 23 ) );
+							o = Engine::projectJournal()->journallingObject(ProjectJournal::idFromSave(*k));
 							if( o && dynamic_cast<AutomatableModel *>( o ) )
 							{
 								a->addObject( dynamic_cast<AutomatableModel *>( o ), false );
--- a/src/core/AutomationPattern.cpp
+++ b/src/core/AutomationPattern.cpp
@@ -787,6 +787,16 @@ void AutomationPattern::resolveAllIDs()
 						{
 							a->addObject( dynamic_cast<AutomatableModel *>( o ), false );
 						}
+						else
+						{
+							// FIXME: Remove this block once the automation system gets fixed
+							// This is a temporary fix for https://github.com/LMMS/lmms/issues/3781
+							o = Engine::projectJournal()->journallingObject( *k + ( 1 << 23 ) );
+							if( o && dynamic_cast<AutomatableModel *>( o ) )
+							{
+								a->addObject( dynamic_cast<AutomatableModel *>( o ), false );
+							}
+						}
 					}
 					a->m_idsToResolve.clear();
 					a->dataChanged();
--- a/include/ProjectJournal.h
+++ b/include/ProjectJournal.h
@@ -77,6 +77,7 @@ public:
 	}
 
 	static jo_id_t idToSave( jo_id_t id );
+	static jo_id_t idFromSave( jo_id_t id );
 
 	void clearJournal();
 	void stopAllJournalling();
--- a/src/core/ProjectJournal.cpp
+++ b/src/core/ProjectJournal.cpp
@@ -164,6 +164,11 @@ jo_id_t ProjectJournal::idToSave( jo_id_t id )
 	return id & ~EO_ID_MSB;
 }
 
+jo_id_t ProjectJournal::idFromSave( jo_id_t id )
+{
+	return id | EO_ID_MSB;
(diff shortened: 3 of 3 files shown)


src/core/AutomationPattern.cpp
```diff
@@ -787,6 +787,16 @@ void AutomationPattern::resolveAllIDs()
 						{
 							a->addObject( dynamic_cast<AutomatableModel *>( o ), false );
 						}
+						else
+						{
+							// FIXME: Remove this block once the automation system gets fixed
+							// This is a temporary fix for https://github.com/LMMS/lmms/issues/3781
+							o = Engine::projectJournal()->journallingObject( *k + ( 1 << 23 ) );
```
jasp00: It is dangerous to mess with ProjectJournal internals like this. You should request new ids and copy data if necessary.

PhysSong: I also think this is a bad design, but this is the minimal fix for the bug which doesn't involve possibly dangerous structure changes for stable releases.
I want to remove this one in `master` as soon as I get my work completed.

jasp00: Please make it less minimal. Do not expose `EO_ID_MSB`.

qnebra:
Possible regression with projects which single kind of parametr from multiple tracks connected to one automation track. For further checking.

src/core/AutomationPattern.cpp
PhysSong: I didn't want to do so. However, changing `ProjectJournal` may affect other stuff than automation. I can address this, but it will make the *temporary* fix much complicated.

jasp00: Please address this.

jasp00: Do you want to become a coder or not?

PhysSong: What do you mean?

jasp00: Let us assume you want to be a better coder. Minimizing structural changes does not mean to make the smallest patch. Minimizing changes certainly does not mean breaking code organization.

I have asked you an easy task: do not expose `EO_ID_MSB`. It will not make your fix that much complicated. Could you give it a try? Do you need a hint?

PhysSong: I agree that the code organization is important, but we're on a branch targetting an upcoming release. Code organization is important in the long term, but bug fixes are more important for now. We consider design changes are out of the scope of `stable-1.2`, though they are important in `master`.
I will remove the exposure of `EO_ID_MSB` in `master`, with new changes in the design.

jasp00: Fine, I will fix your fix.

PhysSong: On which branch, and how?

zonkmachine: @PhysSong I think he intends fixing this in stable-1.2.
Merge this PR and we'll know for sure.

PhysSong: Okay. BTW, @jasp00, please note that you will probably make double work. I already have a fix with new designs and what I need to do is just cleaning them up.

DomClark: I think it makes more sense to argue using comments, rather than to argue using commits. We can merge or close once we've reached a consensus on what the code should look like.
Is the desired change something as simple as, e.g.
```cpp
jo_id_t ProjectJournal::idFromSave( jo_id_t id )
{
	return id | EO_ID_MSB;
}
```
?

PhysSong: I'm not sure if we need that, but that could be a solution.

PhysSong: @jasp00 If you're comfortable with the @DomClark's suggestion, I will use that for now.

jasp00: OK.

PR #4737, opened 2018-12-20
transparent window background, fixes for fedora linux 64bits, and gamepad support

hartsantler:


plugins/CMakeLists.txt
```diff
@@ -76,9 +76,9 @@ IF("${PLUGIN_LIST}" STREQUAL "")
 		stereo_enhancer
 		stereo_matrix
 		stk
-		vst_base
-		vestige
-		VstEffect
+		#vst_base
+		#vestige
+		#VstEffect
```
zonkmachine: It looks like these changes were added by mistake. It will remove anything VST and GigPlayer for everyone.

karmux:
Before and after screenshots would be nice.
These joystick changes should be in the separate pull request.

hartsantler:
I got the basic gamepad stuff working.  Sometimes it crashes on exit, i thought i cleaned up and closed down properly, but maybe i'm missing something.
To test it just plug in a standard dual analog gamepad, and then right click any Knob, and selected one of the gamepad axis (1-6)
Note, game pad buttons do not do anything yet.

plugins/CMakeLists.txt
hartsantler: i didn't see apt-get or dnf instructions how to install GigPlayer.

i think it is fine to comment out the VST stuff, because it depends on fixes to wine that are still upstream and not in the Fedora repos.

The master branch needs to compile on Linux with just a few standard commands.  Editing the CMake is bullshit.
Other operating systems can come second, Linux needs to be first class, after all this project is called:  LinuxMultiMediaStudio

zonkmachine: > i didn't see apt-get or dnf instructions how to install GigPlayer.

The name of the package in Debian is libgig. I don't know why it's not in Fedora but libgig comes from the linuxsampler project and they don't use a license that all projects can work with/around and I've seen Fedora remove stuff before. Our own sample library for instance. There should be a repo for it somewhere out there. To turn gigasampler off hit cmake with this:
```-DWANT_GIG=OFF```

> i think it is fine to comment out the VST stuff, because it depends on fixes to wine that are still upstream and not in the Fedora repos.

The solution for now is to either revert wine to an earlier version or turn VST stuff off with
```-DWANT_VST```

Please revert your changes to plugins/CMakeLists.txt.

**Comment by hartsantler** [comment]
@zonkmachine, i wish i had time to break things apart into multiple pull requests.  Unfortunately i am on a very tight time line, and making updates along the way to all the features i'm adding.
So will not be able to break it apart into multiple pull requests.

File: plugins/CMakeLists.txt (hunk 1 above)
Comment (hartsantler) [reply plugins/CMakeLists.txt]: i didn't want to install wine in the first place, and i'm not going to try to downgrade it.

The CMake script could be changed to look for wine, and if not installed automatically set WANT_VST to off.  CMake could call out to `lsb_release -is` which on my machine returns `Fedora`.  Then we can have simple and standard build commands, that works for all the top linux distros.

Universal build commands for Linux
```bash
cd lmms
mkdir build
cd build
cmake ../.
make
```

.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "src/3rdparty/qt5-x11embed"]
 	path = src/3rdparty/qt5-x11embed
-	url = https://github.com/Lukas-W/qt5-x11embed.git
+	url = https://github.com/hartsantler/qt5-x11embed.git
```
tresf: Please revert, your branches are identical.

plugins/CMakeLists.txt
tresf: > The CMake script could be changed to look for wine, and if not installed automatically set WANT_VST to off. 

This can and should be the case but is out of scope of this PR.  Please adjust it accordingly.

tresf:
> So will not be able to break it apart into multiple pull requests

I'm sure you're familiar with the difficulties of merging larger PRs.  The larger the PR, the longer it takes for review (code quality, feature testing) which in turn means the greater chance of a code conflict (e.g. moving target, etc).

Perhaps when you have more time, you can cherry-pick some of your useful features.

PhysSong:
I closed this PR because it's unmergeable in this state. Some of the changes might be useful though.

Issue #3321, opened 2017-02-03
Tooltip for channel-names in FX-mixer inconsistent 


musikBear:
OS  : winXP sp3
LMMS 1.1.90-RC2.8

**Issue description:**
Displaying FX-channel-names with tool-tip is unpredictable inconsistent 
**Example          :**
In both examples, the mouse is over the red dot.
![tooltipinconsisten01](https://cloud.githubusercontent.com/assets/6662066/22603114/5b7fa192-ea46-11e6-9f0c-12cb01ac1300.JPG)

Here the tool-tip only shows *FX1* and not the channel-name
![tooltipinconsisten02](https://cloud.githubusercontent.com/assets/6662066/22603156/7dba9eba-ea46-11e6-9084-e1899adcdb79.JPG)

Here the channel-name of the next channel is shown in the tool-tip

I have tried with both new and old projects, and some are fine, some has it messed up
I have no explanation for the behaviour 
So far i know:
* It is not connected to the version of creation
* it has not to do with focus
* not related to muted

zonkmachine:
I've also noticed some possible odd behaviour there but as you've noticed too it's intermittent. Just acts up when it feels like.

follower:
Oh, good, I don't have to make any screenshots for this issue. :)

I also noticed the same issue and eventually noticed at least one reliable cause:

 * Track Gear Wheel (e.g. on Triple Oscillator track) > FX 0: Master > Assign to new FX Channel.

This will result in a FX Channel tool tip with "FX 1" rather than "TripleOscillator" as expected.

"Double-clicking on the FX Channel name" and ("hitting enter" or "clicking off the name") will result in the tool tip displaying the correct result. (Which I sometimes do accidentally so hadn't noticed the connection previously.)

I've yet to look into the code to see why this happens as I'm also looking into a bunch of other issues but hopefully being able to reproduce the issue is a start.

I've also noticed (maybe only once?) the actual channel name and tool-tip being the truncated form with an ellipsis ("...") but it may be unrelated.

musikBear:
Need a *'Bug'-label*

zonkmachine:
Fixed in #4740 / dd6d4a552b3e1c51fc39db7c0cfc19d3f33908ef

PR #4740, opened 2018-12-21
Fix broken FxLine tool tips (#3321)

xy124:
Hi, this is my first contribution, so please tell me if I do something wrong.
I hope many will follow ;)

### DIFF
--- a/src/gui/FxMixerView.cpp
+++ b/src/gui/FxMixerView.cpp
@@ -73,7 +73,7 @@ FxMixerView::FxMixerView() :
 
 	// Set margins
 	ml->setContentsMargins( 0, 4, 0, 0 );
-	
+
 	// Channel area
 	m_channelAreaWidget = new QWidget;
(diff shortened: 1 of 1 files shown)


zonkmachine:
Tested, fixes https://github.com/LMMS/lmms/issues/3321, thanks!

It looks fine to me but I can't really review the code. If it's acceptable to the rest of the devs I think this should go into stable-1.2 .

Welcome to the lmms but tracker... :)

PhysSong:
It looks fine to me, though I think we may also fix this on a different side.

zonkmachine:
> It looks fine to me, though I think we may also fix this on a different side.

@PhysSong Can you want to be more specific? I would like to push this to stable-1.2 tomorrow if there are no bigger objections.

zonkmachine:
Pushed to stable-1.2 directly. dd6d4a552b3e1c51fc39db7c0cfc19d3f33908ef
@xy124 thanks for the fix!

Issue #4665, opened 2018-10-14
Forced clipping issues


zonkmachine:
As commented [here](https://github.com/LMMS/lmms/pull/3706#issuecomment-382326210), since https://github.com/LMMS/lmms/pull/3706 was merged old projects may sound differently. The forced clipping is by design and necessary for the rest of that PR to function but may need some tweaking.

We can bump the clipping level to maybe +/-10 or so.

https://github.com/LMMS/lmms/blob/fbdd09dbf56f0912ed8750cc72dda8403f5a01ca/src/core/MixHelpers.cpp#L87

Any other ideas?

@Jousboxx Can you comment on the difference reported on your track, BuzzerBeater in 1.2.0-rc6 or later?

Jousboxx:
There isn't much audible change in Buzzer Beater on 1.2 rc7, but FX4 peaks at +14db, so I'm guessing it isn't audible just because there is also a hard limiter (basically clipping) on the master.

I just read through the thread you linked, and I can see why clipping the FX channels would help remove NaN/inf values. However, I think it should be optional and not on by default. I've seen a whole bunch of projects (including my own) from less experienced producers that heavily rely on the mixers not clipping. Even though it isn't good practice to go over 0db, it's really easy to just turn something up if it sounds too quiet rather than turning everything else down, and that's what a lot of people do. Eventually you end up with red mixers everywhere. To demonstrate this, here's a video of one of my brother's projects after about 2 years of producing: https://youtu.be/jq4A6cWw7o8. This is not a joke project, that's just how he and many others work. He made several songs like that that sounded quite good with the limiter on, and he's only recently started using proper mixing techniques.

If the 10db clipping was on by default, a lot of noobs would probably think that LMMS "just sounds bad", or they might not even notice and end up with clipping in their final mix.

In my entire time using LMMS I've never had any problems with NaN/infs except in projects that were sent to me explicitly to demonstrate them. This is by no means a common issue! I think the option should just be available so that if someone gets really unlucky and has them, they can still use their project.

zonkmachine:
> However, I think it should be optional and not on by default. I've seen a whole bunch of projects (including my own) from less experienced producers that heavily rely on the mixers not clipping. Even though it isn't good practice to go over 0db

If this was made an option I'd rather see it on by default and then the kids can advice each other on Discord how to switch it off.

@JohannesLorenz It's actually by design. I understand understand fully if people see this as a bug though and you can label it as you see fit.

[...]

Issue #3170, opened 2016-12-20
LADSPA vinyl effect is broken


musikBear:
OS  : winXP sp3
LMMS: 1.1.90-g33d8v44

**Issue description:** User report this plugin broken. it will create strange beeb artefacts in various freq.

**Recreate         :** 
new project. 
Insert 3oc default Bell.
insert LADSPA Vinyl-effect in instrument FX-tab
play a note

DeRobyJ:
Confirmed on win7 and win8.1

karmux:
I can reproduce this on Kubuntu Linux 16.10 but only with SDL. Vinyl and some other LADSPA issues vanish with ALSA for some reason.

zonkmachine:
I've got the beeps but it's pretty intermittent.

The plugin generates some warnings of unititialized variable.

```

vynil_1905.xml:78:36: warning: variable 'click_buffer_omega' is uninitialized when used here [-Wuninitialized]

        plugin_data->click_buffer_omega = click_buffer_omega;

                                          ^~~~~~~~~~~~~~~~~~

vynil_1905.xml:237:2: note: variable 'click_buffer_omega' is declared here

        fixp16 click_buffer_omega;

        ^

vynil_1905.xml:79:34: warning: variable 'click_buffer_pos' is uninitialized when used here [-Wuninitialized]

        plugin_data->click_buffer_pos = click_buffer_pos;

                                        ^~~~~~~~~~~~~~~~

vynil_1905.xml:238:2: note: variable 'click_buffer_pos' is declared here

        fixp16 click_buffer_pos;

        ^

2 warnings generated.
```

zonkmachine:
Demo project here: [vynilnoise.mmp.zip](https://github.com/LMMS/lmms/files/2724415/vynilnoise.mmp.zip)
Turn the volume down.

zonkmachine:
The 'beeps' seem to be unaffected by buffer size and sample quality and are square wave bursts at some fixed frequency depending on the effects settings and lasting about **35-50 ms**. I think it could possibly be a filter driven into resonance by a sudden sound spike.

zonkmachine:
It seem to be the `Year` setting that decides the frequency of the square wave burst.
Some settings will generate inf/NaN.

VilhelmEnt:
If your problems are just beeping, it is because the effect is on a certain instrument. I fixed this problem by adding it as an effect for the overall effect bar

lbfalvy:
[...]

PR #4743, opened 2018-12-26
Improve handling of nan/inf

zonkmachine:
If we find nan/inf, we declare the whole buffer bad and set it to 0.0f. I'm still clipping the signal, for now at least, but set the level at ~~+/-100.0f~~ **+/-10.0f**.

Fixes https://github.com/LMMS/lmms/issues/4665

None of the example projects used to test nan/inf works any longer as each of their specific issues has been fixed. To test this PR properly you need to cherry-pick it on top of an older version of lmms like rc2.

### DIFF
--- a/src/core/MixHelpers.cpp
+++ b/src/core/MixHelpers.cpp
@@ -95,7 +95,7 @@ bool sanitize( sampleFrame * src, int frames )
 			}
 			else
 			{
-				src[f][c] = qBound( -100.0f, src[f][c], 100.0f );
+				src[f][c] = qBound( -4.0f, src[f][c], 4.0f );
 			}
 		}
 	}
--- a/src/core/MixHelpers.cpp
+++ b/src/core/MixHelpers.cpp
@@ -79,12 +79,23 @@ bool sanitize( sampleFrame * src, int frames )
 		{
 			if( isinff( src[f][c] ) || isnanf( src[f][c] ) )
 			{
-				src[f][c] = 0.0f;
+				#ifdef LMMS_DEBUG
+					printf("Bad data, clearing buffer. frame: ");
+					printf("%d: value %f\n", f, src[f][c]);
+				#endif
+				for( int f = 0; f < frames; ++f )
+				{
+					for( int c = 0; c < 2; ++c )
+					{
+						src[f][c] = 0.0f;
+					}
+				}
 				found = true;
+				return found;
 			}
 			else
 			{
-				src[f][c] = qBound( -4.0f, src[f][c], 4.0f );
+				src[f][c] = qBound( -100.0f, src[f][c], 100.0f );
 			}
 		}
 	}
--- a/src/core/MixHelpers.cpp
+++ b/src/core/MixHelpers.cpp
@@ -95,7 +95,7 @@ bool sanitize( sampleFrame * src, int frames )
 			}
 			else
 			{
-				src[f][c] = qBound( -4.0f, src[f][c], 4.0f );
+				src[f][c] = qBound( -10.0f, src[f][c], 10.0f );
 			}
 		}
 	}


src/core/MixHelpers.cpp
```diff
@@ -77,7 +77,7 @@ bool sanitize( sampleFrame * src, int frames )
 	{
 		for( int c = 0; c < 2; ++c )
 		{
-			if( isinff( src[f][c] ) || isnanf( src[f][c] ) )
+			if( unlikely( isinff( src[f][c] ) || isnanf( src[f][c] ) ) )
```
PhysSong: Please note that `unlikely` is incompatible with MSVC.

PhysSong:
I wonder if clearing whole buffer will cause some noise like https://github.com/LMMS/lmms/issues/4625#issuecomment-425470843 , though I guess the possibility is low.

src/core/MixHelpers.cpp
zonkmachine: Right. As I understand it this should work in master then where `Q_UNLIKELY(x)` is used instead of `builtin_expect`. I'll revert this for stable-1.2 .

zonkmachine: Fixed!

zonkmachine:
I find it hard to guess with the nan stuff. If you have a situation where the signal is going into/out of forbidden regions (inf/nan) you would probably get a squarish, digital noise kind of signal, clipped at +/-100 . Not an altogether appealing scenario.
What I've seen when testing this PR on top of rc2 is that the nan/inf mostly happens already in the beginning (first or second sample) of the buffer then that run is mostly corrupted so the end result is not different from the earlier way with the whole buffer becoming cleared out. The difference is when the corruption happens somewhere in the middle with extreme level signals in front of it. Here we just sneak back and cut that part out.
I think the best idea going forward would be to somehow bypass a unit once it creates bad data and signal with the LED color like is done for that FX units that aren't found . The noise source is contained and the user informed.

jasp00:
Sanitizing values should not depend on buffer size. Overall, `MixHelpers::sanitize()` should not be executed by default. This behavior should be available as an effect plug-in.

zonkmachine:
> Sanitizing values should not depend on buffer size.

Why not? It is a wee bit aggressive, yes, but the problem area (in my opinion) is before and after the nan/inf samples and this was the easiest way to get to them. I haven't tested this with different buffer size setting yet. Are you OK with clearing out samples around the nan/inf samples if it just isn't the whole buffer? Such as 16 bits before/after. It would no longer depend on the buffer size.

> Overall, `MixHelpers::sanitize()` should not be executed by default.

A switch in `Settings > Performance Settings` ?

Do you know how sanitizing is handled in other DAWs?

jasp00:
Sanitizing values should not depend on buffer size because it does not fix your problem. What if the first sample in the buffer is NaN? To accomplish your goal, you need to apply a window, a filter. This decision should be left to the user through effect plug-ins, not hard coded neither through settings.

Another issue is that NaN and infinite are not the same. Setting infinite to zero is wrong. Clipping should also be left to the user through effect plug-ins.

I have not looked at other DAWs, I do not think it is necessary in this case.

zonkmachine:
> > Do you mean some of this should be fixed for 1.2 already like the option to turn the sanitizing on/off?
> 
> Your call. My choice for 1.2 would be a LADSPA effect.

To fix what? The sanitize() function used to only run when exporting but now it's been turned on for regular use too (see https://github.com/LMMS/lmms/pull/3706). Do you mean this should be reverted and an effect unit used instead? I'm not against it but I can't do it myself.

Spekular:
I still haven't seen any reason provided for why the sanitizing should be
an effect, and it seems absurd to me. I can't imagine any other DAW would
do this. In an ideal world, the user wouldn't have to know or care about
the possibility of a bug existing, so why would you intentionally make this
optional, off by default, and implemented via effect?

Release builds don't provide a "crash instantly on exceptions" toggle in
the settings, so why would they include a "let audio output completely
break" setting?

zonkmachine:
> What if the first sample in the buffer is NaN?

Or infinite... Then the whole buffer gets knocked out. One solution for all cases. It's not about making that channel sound good but about solving an issue that is an extreme case and to try and remove the possibility of sound artifacts that can be annoyingly/dangerously loud.

> As a user, I could use an anti-NaN effect before the bug is fixed.
> 
> >

Or, I could observe an error message that informs me that a pluggin is acting up and that lmms has taken action to protect 1. my sanity 2. my ears 3. my equipment 4. the rest of the project.
I'm still not convinced that this is the wrong way to go about things.

jasp00:
I will add a hidden setting to disable forced sanitizing, okay?

zonkmachine:
> I will add a hidden setting to disable forced sanitizing, okay?

Sure!

softrabbit:
> I think the best idea going forward would be to somehow bypass a unit once it creates bad data and signal with the LED color like is done for that FX units that aren't found . The noise source is contained and the user informed.

That sounds like a good solution, as it could be implemented through checking continuosly for NaN/Inf on the master channel, invoking more thorough checking only when needed to find the offending unit.

zonkmachine:
> to disable forced sanitizing, okay?

Maybe remove it for exporting too then, to make render and playback work the same? It's not very common but there are cases where users have experienced noise in the exported track where there is none in playback or silence in playback but exported track works fine.

zonkmachine:
Reverted the increased clipping level and here is the file that made me take that decision.
[vynilnoise.mmp.zip](https://github.com/LMMS/lmms/files/2724415/vynilnoise.mmp.zip)
https://github.com/LMMS/lmms/issues/3170

It works pretty poor with or without sanitizing. It still works worst without sanitizing though, but not with a ground breaking lead. The Vynil effect is one that receives quite a lot of remarks of making unvanted noise. It simply makes loud beeps spontaneously.

PR #4745, opened 2018-12-27
Allow sub plugins for instruments aswell (LV2 requirement)

JohannesLorenz:
For LADSPA, we only need sub plugins for effects, but for LV2 (and SPA), we will also need them for instruments. At least, assuming we want to keep using the current sub plugin structure. That's what this patch does.

Some notes for reviewers:
* It's 47 files, but 25 of them are only minor changes in the plugin descriptors
* The commit messages should help explaining what was changed and why
* Don't worry too much about whitespace and identation. Sometimes I even broke the coding conventions to not have 2 coding styles in one cpp file. In the end, we'll run clang-format over the whole LMMS code anyways...

As this is an LV2 requirement and has influence on large parts of the core, I'd be really thankful if someone tries to review this. :heart:


JohannesLorenz:
Side-node: It compiles now on most/all CI platforms, and I could test it in a non-committed version of my SPA branch. However, without SPA/LV2, you won't see anything new, as we currently don't have instruments with sub plugins.

davisjacobym:
Not sure if this is necessarily what you're asking, but just wanna clarify--

By "sub plugin", would that mean nested VSTs? (meaning a vst which hosts other vsts from the same origin)

JohannesLorenz:
> By "sub plugin", would that mean nested VSTs? (meaning a vst which hosts other vsts from the same origin)

No idea what nested VSTs are. E.g. LADSPA is a plugin, and CALF Mono Compressor is a LADSPA sub plugin. Or, VST is a plugin, and zynaddsubfx-vst is a sub plugin. However, I have no idea if this PR affects VST. I only did this for Lv2 and SPA.

include/Plugin.h
```diff
@@ -101,12 +147,28 @@ class LMMS_EXPORT Plugin : public Model, public JournallingObject
 
 				inline bool isValid() const
 				{
-					return desc != NULL && name.isNull() == false;
+					return desc != nullptr;
```
Reflexe: Is this change needed for this pr?

src/core/Plugin.cpp
```diff
@@ -74,6 +80,97 @@ Plugin::~Plugin()
...
+
+
+QString Plugin::Descriptor::SubPluginFeatures::Key::additionalFileExtensions() const
+{
+	Q_ASSERT(isValid());
+	return desc->subPluginFeatures
+		// get from sub plugin
+		? desc->subPluginFeatures->additionalFileExtensions(*this)
+		// get from plugin
+		: nullptr;
```
Reflexe: `nullptr` is `get from plugin` (maybe `QString()` instead)?

src/core/PluginFactory.cpp
```diff
@@ -150,42 +157,79 @@ void PluginFactory::discoverPlugins()
...
+		if (library->resolve("lmms_plugin_main"))
 		{
-			descriptorName = descriptorName.mid(3);
+			QString descriptorName = file.baseName() + "_plugin_descriptor";
+			if( descriptorName.left(3) == "lib" )
+			{
+				descriptorName = descriptorName.mid(3);
+			}
+
+			pluginDescriptor = reinterpret_cast<Plugin::Descriptor*>(library->resolve(descriptorName.toUtf8().constData()));
```
Reflexe: you would have OOB read if `sizeof(Plugin::Descriptor)` is greater than the actual returned size.

src/core/PluginFactory.cpp
```diff
@@ -150,42 +157,79 @@ void PluginFactory::discoverPlugins()
...
 		}
-
-		Plugin::Descriptor* pluginDescriptor = reinterpret_cast<Plugin::Descriptor*>(library->resolve(descriptorName.toUtf8().constData()));
-		if(pluginDescriptor == nullptr)
+		else
 		{
-			qWarning() << qApp->translate("PluginFactory", "LMMS plugin %1 does not have a plugin descriptor named %2!").
-						  arg(file.absoluteFilePath()).arg(descriptorName);
-			continue;
+			//qDebug() << "Ignoring" << file.absoluteFilePath() << "(no lmms_plugin_main())";
```
Reflexe: Should you remove that?

src/core/PluginFactory.cpp
```diff
@@ -150,42 +157,79 @@ void PluginFactory::discoverPlugins()
...
+			info.library = library;
+			info.descriptor = pluginDescriptor;
+			pluginInfos << info;
+
+			auto addSupportedFileTypes =
+				[this](QString supportedFileTypes,
+					const PluginInfo& info,
+					const Plugin::Descriptor::SubPluginFeatures::Key* key = nullptr)
+			{
+				if(!supportedFileTypes.isNull())
```
Reflexe: Please add use a new block for this multiline expression (the for loop)

src/gui/PluginBrowser.cpp
```diff
@@ -171,8 +192,9 @@ void PluginDescWidget::mousePressEvent( QMouseEvent * _me )
 {
 	if( _me->button() == Qt::LeftButton )
 	{
-		new StringPairDrag( "instrument", m_pluginDescriptor.name,
```
Reflexe: mmm memory leak? (it has nothing to do with this PR, I just haven't seen anything like it)

include/Plugin.h
JohannesLorenz: The `name` class members can be nullptr while the whole key is valid. `name` is only used in some situations and must be considered temporary. See my comment above the `Key::name` member.

Also, the change should be harmless to old code, as it has only ever be used inside `EffectSelectDialog::acceptSelection`.

src/core/Plugin.cpp
JohannesLorenz: You are right in both cases. I removed "get from plugin", because obviously, you can't get additional file extensions if there is no sub plugin. Also I changed `nullptr` to `QString`.

src/core/PluginFactory.cpp
JohannesLorenz: Yes, I've removed it.

include/Plugin.h
```diff
@@ -125,11 +187,42 @@ class LMMS_EXPORT Plugin : public Model, public JournallingObject
...
+			virtual QString additionalFileExtensions(const Key&) const
+			{
+				return nullptr;
+			}
+
+			virtual QString displayName(const Key& k) const
+			{
+				return k.isValid()
+					? k.desc->displayName
+					: k.name.toUtf8().data();
```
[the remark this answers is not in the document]
JohannesLorenz: Good point! I returned `const char*` in an early version of this PR, and I forgot to fix that.

src/core/PluginFactory.cpp
JohannesLorenz: Can you please elaborate?

src/gui/PluginBrowser.cpp
JohannesLorenz: [Probably not](https://doc.qt.io/archives/qt-4.8/dnd.html): "The QMimeData and QDrag objects created by the source widget should not be deleted - they will be destroyed by Qt".

This makes sense: `StringPairDrag` passes `this` to its parent class `QDrag` which passes it to Qt. `QDrag` is not being deleted in this drag yet, but only after the drop (by Qt).

src/core/PluginFactory.cpp
Reflexe: Since you directly reads from a memory buffer without checking its actual length, you could crash in here if the returned symbol is shorter than Descriptor. (It's minor and should not happen, but also isn't a good practice).

JohannesLorenz: I think this can only happen if plugin and LMMS have different opinions on what `Plugin::Descriptor` is, i.e. if they are linked to different LMMS core versions?

Do you have any suggestions how to improve that?

JohannesLorenz: Resolving, as this can't really be fixed, and it's not really part of this PR.

JohannesLorenz:
Many thanks @Reflexe for reviewing.

LMMS can now handle *masses* of instruments :smiley:

Issue #4746, opened 2018-12-28
Automation tracks in the Beat Bassline are rendered as empty sound files


zonkmachine:
Automation tracks in the Beat/Bassline are rendered as empty sound files. You see this when you render as separate tracks. I've enclosed a simple demo project. Just choose `Export Tracks...` .

I see this already in lmms-1.1.3 .

[exportauto.mmp.zip](https://github.com/LMMS/lmms/files/2715822/exportauto.mmp.zip)

PR #4747, opened 2018-12-29
Don't render Automation tracks in the Beat/Bassline

zonkmachine:
Copy/paste of related code in `RenderManager::renderTracks()`

Fixes https://github.com/LMMS/lmms/issues/4746

### DIFF
--- a/src/core/RenderManager.cpp
+++ b/src/core/RenderManager.cpp
@@ -142,7 +142,10 @@ void RenderManager::renderTracks()
 	for( auto it = t2.begin(); it != t2.end(); ++it )
 	{
 		Track* tk = (*it);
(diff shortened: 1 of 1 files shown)


zonkmachine:
This comment here looks a bit 'off'.
https://github.com/LMMS/lmms/blob/5218c1d1c58313118e9faac2e80296cf6e42fa27/src/core/RenderManager.cpp#L133

I think "Don't render Automation Tracks" is probably what was intended. The function commented selects all tracks that aren't muted and which is either an Instrument Track or a Sample Track for rendering. So an Automation Track is treated the same whether it's muted or not.

zonkmachine:
I've seen nothing problematic with this. Merging this year if no one opposes. :)

PR #4757, opened 2019-01-03
Corrected assignment

shlyakpavel:
There is no need in conversion from float to integer


tresf:
> It could be applied to stable-1.2 too.

Yes, it would be welcome on `stable-1.2` but since it's targeted at master and is likely [optimized out of release builds anyway](https://stackoverflow.com/questions/49225332) we'll just apply it to `master`.

PR #4758, opened 2019-01-03
Prepare source: set project year

jasp00:
Prepare source by processing Git history, which helps reproducible builds. This first patch updates the project year.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ language: cpp
 compiler: gcc
 dist: trusty
 sudo: required
+git:
+    depth: false
 cache:
     directories:
         - apt_mingw_cache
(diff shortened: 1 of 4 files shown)


tresf:
This is redundant with the attempted `maintenance tasks` from https://github.com/LMMS/lmms.io/pull/198.  Just bump the year for now.  Having Travis do this is the wrong place and the maintenance tasks didn't end up being merged due to some failed tests.

jasp00:
This is my preferred option. It was not possible before because we did not build a source artifact. It is more robust than a maintenance task. Which method do you prefer? External job or integrated build step?

jasp00:
For stable-1.2, how do you handle the contributor list? Will you upload `doc/CONTRIBUTORS`?

tresf:
@jasp00 yes, someone should. It could be combined with the `PROJECT_YEAR` update.

jasp00:
Approach for stable-1.2 at #4763.

PR #4760, opened 2019-01-04
Fix JACK crash on exit

### LINKED ISSUE #4688: Crash on close when using JACK for both audio and MIDI
`MidiJack` reuses JACK connection from `AudioJack` if the audio backend is JACK. The `MidiJack` destructor calls `MidiJack::jackClient` which calls `m_jackAudio->jackClient()`, but `m_jackAudio` is already deleted at the point. It causes segmentation faults.
#4384 contains a commit(abb13bd9378c22c0647a7958f8f779d4bd63e273) which addresses this issue, but the PR was closed because @justnope have deleted the account.
I believe this issue is not very hard to fix.

softrabbit:
Fix segfault on exit when using JACK for both audio and MIDI. This is picked from an old commit (https://github.com/LMMS/lmms/commit/abb13bd9378c22c0647a7958f8f779d4bd63e273) by @justnope.

### DIFF
--- a/src/core/midi/MidiJack.cpp
+++ b/src/core/midi/MidiJack.cpp
@@ -75,6 +75,7 @@ MidiJack::MidiJack() :
 	{
 		// if a jack connection has been created for audio we use that
 		m_jackAudio->addMidiClient(this);
+		m_jackClient = NULL;
 	}else{
 		m_jackAudio = NULL;
 		m_jackClient = jack_client_open(probeDevice().toLatin1().data(),
@@ -113,6 +114,9 @@ MidiJack::MidiJack() :
 
 MidiJack::~MidiJack()
 {
+	if( m_jackAudio ) {
+		m_jackAudio->removeMidiClient();
+	}
 	if(jackClient())
 	{
 		if( jack_port_unregister( jackClient(), m_input_port) != 0){
@@ -143,14 +147,15 @@ MidiJack::~MidiJack()
 	}
 }
 
+void MidiJack::removeJackAudio() {
+	m_jackAudio = nullptr;
+}
+
 jack_client_t* MidiJack::jackClient()
 {
-	if( m_jackAudio == NULL && m_jackClient == NULL)
-		return NULL;
-
-	if( m_jackAudio == NULL && m_jackClient )
+	if( m_jackAudio == NULL )
 		return m_jackClient;
-
+	
 	return m_jackAudio->jackClient();
 }
 
--- a/src/core/midi/MidiJack.cpp
+++ b/src/core/midi/MidiJack.cpp
@@ -155,7 +155,7 @@ jack_client_t* MidiJack::jackClient()
 {
 	if( m_jackAudio == NULL )
 		return m_jackClient;
-	
+
 	return m_jackAudio->jackClient();
 }
 
--- a/include/AudioJack.h
+++ b/include/AudioJack.h
@@ -57,6 +57,7 @@ public:
 	// the jack callback is handled here, we call the midi client so that it can read
 	// it's midi data during the callback
 	AudioJack * addMidiClient(MidiJack *midiClient);
+	void removeMidiClient();
 	jack_client_t * jackClient() {return m_client;};
 
 	inline static QString name()
--- a/include/MidiJack.h
+++ b/include/MidiJack.h
@@ -52,7 +52,8 @@ class MidiJack : public QThread, public MidiClientRaw
 public:
 	MidiJack();
 	virtual ~MidiJack();
-
+	void removeJackAudio();
+	
 	jack_client_t* jackClient();
 
 	static QString probeDevice();
--- a/src/core/audio/AudioJack.cpp
+++ b/src/core/audio/AudioJack.cpp
@@ -78,7 +78,9 @@ AudioJack::~AudioJack()
 		unregisterPort( m_portMap.begin().key() );
 	}
 #endif
-
+	if(m_midiClient) {
+		m_midiClient->removeJackAudio();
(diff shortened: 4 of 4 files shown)


PhysSong:
It's the same that we're referencing deleted `AudioJack` instance there, but we don't try to do something with a closed client anymore. The crash seems gone with this patch, this is already an improvement.
However, I believe the wrong reference should be fixed in `master` later.

softrabbit:
> I thought in the original fix I made I just created another instance of jack_client_t for midi. At the time it seemed like a better solution to create separate instances for midi and audio instead of sharing it. I set the pointer to null not to fix anything, but rather to help with debugging and force a crash should something try to use it after it was deleted. Check the rest of the code in the original fix.

OK, I'll look closer. I was shooting for the minimal solution to the problem at hand, and the lines I picked do seem to get rid of the segmentation fault at exit...

justnope:
A minimal solution is usually the best solution.

I've reread my patch and you'll see that at a certain point I switched the delete of the midi and audio device. That was me being desperate before I realised "hey, let's ask jack for another device". And it worked. No more worries about who owns what devices, what's being shared. They can both be independently be shutdown.

The other parts of the patch are trying (and succeeded but needs more tests) to get jack midi working with my synth. In retrospect, maybe I should've split the patch in two.

tresf:
>  the lines I picked do seem to get rid of the segmentation fault at exit...

> > A minimal solution is usually the best solution.

Agreed, especially on our stable branch.  Can we come to a decision and merge?

src/core/midi/MidiJack.cpp
```diff
@@ -143,14 +147,15 @@ MidiJack::~MidiJack()
...
 jack_client_t* MidiJack::jackClient()
 {
-	if( m_jackAudio == NULL && m_jackClient == NULL)
-		return NULL;
-
-	if( m_jackAudio == NULL && m_jackClient )
+	if( m_jackAudio == NULL )
 		return m_jackClient;
-
+
```
PhysSong: Trailing whitespaces added here.

PhysSong:
I think we should use separate connections for audio and MIDI, or shared pointers with a wrapper class for correct cleanups.

PhysSong:
For 1.2, I'd vote for the first approach.

jasp00:
@softrabbit, do you want to merge this or not?

lukas-w:
I agree that 26586825ffafbfd53ccf9a0d6a5ff1e5273fbf2f should be less preferred because it enforces Audio/Midi intertwining as @jasp00 pointed out.

However, acdabdafcf051fb55b85804c48d1e42ef8fdc673 may hide the crash, but it leaves undefined behaviour because `MidiJack` is still accessing the deleted `AudioJack` instance.

As a very minimal fix, I suggest simply swapping the deletion order of `m_audioDev` and `m_midiClient` in `Mixer.cpp` so that `m_midiClient` is deleted first, which corresponds to `m_midiClient` being constructed last.

On `master`, can we cherry-pick abb13bd9378c22c0647a7958f8f779d4bd63e273 or is there an advantage in only opening a single jack client?

PhysSong:
> As a very minimal fix, I suggest simply swapping the deletion order of `m_audioDev` and `m_midiClient` in `Mixer.cpp` so that `m_midiClient` is deleted first, which corresponds to `m_midiClient` being constructed last.

@lukas-w Sounds good. Have you tested that?

lukas-w:
> Have you tested that?

Yes, the crash is gone after fixing another unrelated bug:

```patch
diff --git a/src/core/Mixer.cpp b/src/core/Mixer.cpp
index fae7b35db..c87305641 100644
--- a/src/core/Mixer.cpp
+++ b/src/core/Mixer.cpp
@@ -186,8 +186,8 @@ Mixer::~Mixer()
 	}
 	delete m_fifo;
 
-	delete m_audioDev;
 	delete m_midiClient;
+	delete m_audioDev;
 
 	for( int i = 0; i < 3; i++ )
 	{
diff --git a/src/core/midi/MidiJack.cpp b/src/core/midi/MidiJack.cpp
index cea7f7381..568b6dae1 100644
--- a/src/core/midi/MidiJack.cpp
+++ b/src/core/midi/MidiJack.cpp
@@ -61,6 +61,7 @@ static void JackMidiShutdown(void *arg)
 
 MidiJack::MidiJack() :
 	MidiClientRaw(),
+	m_jackClient( nullptr ),
 	m_input_port( NULL ),
 	m_output_port( NULL ),
 	m_quit( false )
```

I'll push this directly to `stable-1.2` if that's fine.

PhysSong:
> I'll push this directly to `stable-1.2` if that's fine.

I think it'll be fine because only `AudioJack` depends on MIDI clients only while it's processing.

lukas-w:
Done via 17e87c1d689e1548d0c98a8a2fce5b229c799dbd.

justnope:
Sorry to be annoying.

The order of deletion doesn't matter. They store the pointers to eachother, if one of them is destroyed the other will still dereference the memory location of the other.

It works in release but not in debug where you overwrite the memory locations you freed with a byte pattern.

That's why I gave up fixing the intertwined jack audio/midi and went straight to separating them.

lukas-w:
@justnope I can't see [`~AudioJack`](https://github.com/LMMS/lmms/blob/stable-1.2/src/core/audio/AudioJack.cpp#L72-L94) using `m_midiClient`, but `~MidiJack` is using `m_audioClient` through `jackClient()`. So changing the order of deletion does fix this particular crash. I tested this using a debug build.

I do agree though that separating them is the way to go for `master`.

PR #4763, opened 2019-01-06
Do not depend on Git history, use data from 2019-01-07

jasp00:
This approach is meant to be until maintenance tasks are implemented at lmms.io. Contributor list is the output of: `git shortlog -sne | cut -c8-`


jasp00:
Merging...

PR #4770, opened 2019-01-14
Bring back PortAudio support on Windows

### LINKED ISSUE #4465: Bring back Portaudio (WASAPI)
I was using LMMS 1.2.0 RC4 on Windows 10 Home, version:1709, OS build: 16299.402.
I used a USB  MIDI keyboard. And my buffer size was 192 frames and latency was 4ms. I tested both SDL and Port Audio (WASAPI) at 140bpm with Piano One vsti, and found that WASAPI was better than SDL, in terms of latency.
I also found that, while exporting the songs at a sample rate higher than 44100hz, some tracks would get detuned (or their volume would decrease), while using SDL. But, no such problems were found in Port audio (WASAPI), while exporting at sample rates higher than 44100hz.
Then later I tried the SDL on 1.2.0 RC6, (on Windows 10 version 1804) and it's SDL has the same problems.
Please consider bringing back Port Audio option. I think WASAPI is getting better.

Thank you.

DomClark:
Fixes #4465.

Note that this does not fix the PortAudio crashes, but prevents them occurring at startup by not initialising PortAudio until it is chosen in the audio interface menu.


zonkmachine:
Since the issue fixed is on Windows, if this needs testing maybe we should provide binaries?

PhysSong:
> maybe we should provide binaries?

I will.

zonkmachine:
Did a quick test on Linux, switching between Alsa, SDL and PortAudio, and I didn't see any differences in the behavior.

PhysSong:
Binaries are available now.
32bit: https://github.com/PhysSong/lmms/releases/download/vdummy/lmms-1.2.0-rc7+portaudio-win32.exe
64bit: https://github.com/PhysSong/lmms/releases/download/vdummy/lmms-1.2.0-rc7+portaudio-win64.exe

tresf:
PortAudio + Windows 10 x64 Testing.  Hardware is Dell XPS 9333 13" Developer Edition.  Sound card: Onboard Realtek High Definition Audio

* ✅**Windows WASAPI + Speakers / Headphones:** Playback good 6-12% CPU
* ⚠️**MME + Microsoft Sound Mapper:** Playback choppy and unusable
* ⚠️**Windows DirectSound + Speakers / Headphones:** Playback choppy and unusable
* ⛔️**Windows WDM-KS + PC Speaker:** No usable configuration for playback

Baseline comparison against SDL:
* ✅ **SDL:** Playback good 7-20% CPU

DomClark:
I get the same results with choppy playback and inappropriate device options. Note however that this is the same behaviour as in 1.1.3, and this PR doesn't attempt to change that - the idea was just to bring back the existing PortAudio backend for Windows again. It definitely needs some work and a new bug report is the way to go.

T0NIT0RMX:
I tried and on my Windows 64 bits, MSI GP60 2QE Leopard, Realtek Audio Card, there is no way to use any settings of portaudio.

If I use windows WASAPI, when I open again lmms the settings window pop up like there is a problem, and if I click OK and get to lmms window, there is no sound.

I tried in Audacity and WASAPI seems to work properly, so I don't really understand what's going wrong.

PhysSong:
@T0NIT0RMX What if you replace the PortAudio dll with Audacity's one?

T0NIT0RMX:
@PhysSong Audacity doesn't seems to have a portaudio dll, or I can't find it

PhysSong:
@T0NIT0RMX Can you test [libportaudio-2.dll from @tobydox's newer PPA](https://github.com/LMMS/lmms/files/2839161/libportaudio-2.dll.zip)?

T0NIT0RMX:
@PhysSong Same results.
But using this one I only have "MME" and "Windows WDM-KS" backend.
No matter what I chose, when LMMS opens again the config window pop up and if I press OK there is no sound in LMMS

zonkmachine:
@T0NIT0RMX How does older versions of lmms work? Like 1.1.3 or 1.2.0 before we removed portaudio?

zonkmachine:
Since we're not getting much test data back I think we should merge this PR and release rc8.

tresf:
Agreed.

zonkmachine:
OK, merging!

@T0NIT0RMX I'm still curious about how portaudio worked under your machine with older versions of lmms. See my [comment above](https://github.com/LMMS/lmms/pull/4770#issuecomment-461768718).

T0NIT0RMX:
@zonkmachine sorry for the delay I've been busy ! I'll let you know

PR #4772, opened 2019-01-14
[Equalizer] Bright analyzer colors, opacity increased

GingkathFox:
I dont really like how dark and transparent the analyzer colors are. For me, it's hard to actually see the difference between the input and the output. This pull changes the in color and makes both colors a bit more opaque.

### DIFF
--- a/plugins/Eq/EqControlsDialog.cpp
+++ b/plugins/Eq/EqControlsDialog.cpp
@@ -55,10 +55,10 @@ EqControlsDialog::EqControlsDialog( EqControls *controls ) :
 
 	EqSpectrumView * inSpec = new EqSpectrumView( &controls->m_inFftBands, this );
 	inSpec->move( 26, 17 );
-	inSpec->setColor( QColor( 54, 45, 142, 150 ) );
+	inSpec->setColor( QColor( 63, 55, 141, 245 );
 
 	EqSpectrumView * outSpec = new EqSpectrumView( &controls->m_outFftBands, this );
-	outSpec->setColor( QColor( 9, 166, 156, 150 ) );
+	outSpec->setColor( QColor( 10, 88, 156, 220 ) );
 	outSpec->move( 26, 17 );
 
 	m_parameterWidget = new EqParameterWidget( this , controls );
--- a/plugins/Eq/EqControlsDialog.cpp
+++ b/plugins/Eq/EqControlsDialog.cpp
@@ -55,7 +55,7 @@ EqControlsDialog::EqControlsDialog( EqControls *controls ) :
 
 	EqSpectrumView * inSpec = new EqSpectrumView( &controls->m_inFftBands, this );
 	inSpec->move( 26, 17 );
-	inSpec->setColor( QColor( 63, 202, 141, 245 ) );
+	inSpec->setColor( QColor( 109, 169, 96, 245 ) );
 
 	EqSpectrumView * outSpec = new EqSpectrumView( &controls->m_outFftBands, this );
 	outSpec->setColor( QColor( 10, 88, 156, 220 ) );
--- a/plugins/Eq/EqControlsDialog.cpp
+++ b/plugins/Eq/EqControlsDialog.cpp
@@ -55,10 +55,10 @@ EqControlsDialog::EqControlsDialog( EqControls *controls ) :
 
 	EqSpectrumView * inSpec = new EqSpectrumView( &controls->m_inFftBands, this );
 	inSpec->move( 26, 17 );
-	inSpec->setColor( QColor( 0, 36, 255, 150 ) );
+	inSpec->setColor( QColor( 77, 101, 242, 150 ) );
 
 	EqSpectrumView * outSpec = new EqSpectrumView( &controls->m_outFftBands, this );
-	outSpec->setColor( QColor( 0, 236, 255, 150 ) );
+	outSpec->setColor( QColor( 0, 255, 239, 150 ) );
 	outSpec->move( 26, 17 );
 
 	m_parameterWidget = new EqParameterWidget( this , controls );
--- a/plugins/Eq/EqControlsDialog.cpp
+++ b/plugins/Eq/EqControlsDialog.cpp
@@ -55,10 +55,10 @@ EqControlsDialog::EqControlsDialog( EqControls *controls ) :
 
 	EqSpectrumView * inSpec = new EqSpectrumView( &controls->m_inFftBands, this );
 	inSpec->move( 26, 17 );
-	inSpec->setColor( QColor( 109, 169, 96, 245 ) );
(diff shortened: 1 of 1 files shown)


plugins/Eq/EqControlsDialog.cpp
```diff
@@ -55,10 +55,10 @@ EqControlsDialog::EqControlsDialog( EqControls *controls ) :
 
 	EqSpectrumView * inSpec = new EqSpectrumView( &controls->m_inFftBands, this );
 	inSpec->move( 26, 17 );
-	inSpec->setColor( QColor( 54, 45, 142, 150 ) );
+	inSpec->setColor( QColor( 63, 202, 141, 245 );
```
PhysSong: A `(` is missing...

GingkathFox: oof, lemmie edit

GingkathFox: fixed, how the heck did i do that 🤔

GingkathFox: wait a sec
why is it not displaying

GingkathFox: o wait, its under xD
i need to figure out this pull request thing

GingkathFox:
[Uploading EqControlsDialog.cpp…]()

tresf:
With any UI-only changes, a before and after helps.

Before: screen shot 2019-01-14 at 10 34 33 pm
After: screen shot 2019-01-14 at 10 38 10 pm

Spekular:
If you want to make the input and output more distinct, it's an odd choice
to make them both blue.

I tried to get reference on what colors other EQs use. A quick search shows
that Pro-Q, Logic X's EQ, and izotope's EQs (inside neutron/ozone) use a
colorless waveform, instead using color to show how individual bands are
affecting the eq curve. Ableton's EQ forgoes color coding entirely and FL
studio's EQ uses an entirely different visualization.

tresf:
Let's please focus on accepting or denying this PR.  If the OP's concerns are valid, focuson on them.  Enhancements described above are out of scope of our stable branch.

Spekular:
I'm not describing any enhancements, though.


1. I have no trouble seeing the two waveforms in the current version, nor
in this edited version. However, in the edited version the waveforms are
less distinct. Perhaps calling it an "odd choice" was too diplomatic? In
the new version, it's harder to tell which waveform is in and which is out.


2. If the color is to be changed, I thought it would be helpful to see how
it's done elsewhere rather than figuring things out from scratch. This
presents two options:


A: We distinguish the signals in some way other than color, as seems to be
the trend. This would mean this PR should be closed rather than changing a
set of colors that will be scrapped.


B: We distinguish the signals by color. This means point 1 is valid and the
PR may need to be edited before being accepted.


Which part of this isn't relevant to the PR?

tresf:
> Enhancements described above are out of scope of our stable branch.

> > Which part of this isn't relevant to the PR?

Out of scope does not negate relevance, but establishing scope forces decision.

* Do you agree the current colors are too transparent
* Do you agree the current colors are hard to differentiate

The PR touches **two** lines of code and makes no mention of making things colorless, making the bands affect color, etc.  The points are extremely relevant, but they're out of scope, mostly because they lead to "endless" discussion.

Let's focus on @GingkathFox's PR and how he can make it better please.

RoxasKH:
In my personal opinion, i really like the actual visualizer colors. I think the new ones are a bit more confusionary, also they don't fit the UI so much. So i'd maintain the actual ones, but it's not only my decision.

GingkathFox:
@Spekular  Oh crap, i think i accidentally made both blue. The inputs supposed to be a green color, lemmie change that.
@tresf cmake and make on MacOS hate me, absolutely nothing will build without a fatal error of some sort. Ill see if i can build on Ubuntu. 
@RoxasKH see above xp

GingkathFox:
Lemmie try to build on Ubuntu, ill get back to yall in a bit.

GingkathFox:
Wasnt able to build, ill try something tomorrow.

GingkathFox:
I cant get make to install on ubuntu 18.04, sorry. Can someone check the color change for me?

GingkathFox:
In:
In
Out:
Out
Both:
Both

GingkathFox:
@RebeccaDeField i dunno, i didnt really like the blue&green combo. ill try it.

GingkathFox:
screen shot 2019-01-31 at 9 01 19 pm

RebeccaDeField:
@GingkathFox I didn't mean to use the old colors, just the same hues. Blue and green should still be alright as long as it has enough contrast (one is lighter than the other) and is brighter to fulfill the purpose of this pr 👍

RebeccaDeField:
@GingkathFox would be a little more complicated to change the hue using RGB. In inkscape, try switching to HSL in fill and stroke.
dialog_fillstroke_hsl

Then only edit the Lightness value, leave the hue/saturation and you should be good to go.

GingkathFox:
im running MacOS >w<
ive been using[this](http://colorizer.org), maybe i can just change the hue of the HSL value and copy the RGB?

tresf:
> In inkscape, try switching to HSL in fill and stroke.

> > im running MacOS >w<

Translation: macOS can't run Inkscape (they claim it works with X11 server, but it doesn't).  @GingkathFox you have a Linux VM, muscle up. 💪 😄

GingkathFox:
@tresf i dont wanna reboot pc to switch thoooo XDDD

RebeccaDeField:
> ive been using this, maybe i can just change the hue of the HSL value and copy the RGB?

@GingkathFox Any color program using HSL will work fine, only assumed inkscape because that's where the source files for the colors are contained but the program choice is arbitrary as long as you can grab the colors you need.

GingkathFox:
In:
in
Out: out
In&Out:
in out

RebeccaDeField:
@GingkathFox This looks good, but I am thinking that the In color should be brighter given the purpose fo the PR?

GingkathFox:
XD now i gotta mess with the in color again! Time to fire up Colorizer and ill push the changes later today @RebeccaDeField

RebeccaDeField:
Given that this matches our existing colors and it looks more visible than the previous version of the equalizer, I think we've accomplished our goal here and this has my approval from the design dept.

RebeccaDeField:
Some checks not successful. Anything we can do to fix that or is this ok to merge? Hoping to get this cleared out of the pr requests as it is a pretty minor change.

PhysSong:
> Some checks not successful. Anything we can do to fix that or is this ok to merge?

The failure looks unrelated to changes in this PR. I restarted the Travis-CI build and it should pass in 30 minutes.

PR #4775, opened 2019-01-16
Enable jack on windows

justnope:
I added a pr to weakjack for 64bit msvc compilation x42/weakjack#11

Currently the official jack installer has a bug in the 64bit version which causes your program to crash if you weak link to it. Compiling your own jack client dll however works. I'll post the instructions below on how to do it with msvc. 
Maybe someone with more experience can take a look at how to compile/use jack with mingw.

jack2 compilation with msvc:
git clone -b 64b https://github.com/justnope/jack2
cd jack2
$env:Path= "\<python27\>;" + $env:Path
$env:Path= "\<path\>/windows;"  + $env:Path
windows/install_vcpkg_deps.bat
$env:INCLUDE="\<path\>/windows/deps/include"
$env:LIB="\<path\>/windows/deps/lib_x64"
python waf configure [--debug] --msvc_targets="x64"
python waf build

\<path\> is the location of the repo
To use it in lmms set the following cmake vars:
JACK_INCLUDE_DIR \<path\>/common
JACK_LIBRARY \<path\>/build/common/jack.lib

Without weak linking: copy libjack-0.dll to the lmms directory.
With weak linking (in case you're compiling 64bits) replace the installed libjack.dll & libjackserver.dll in the windows directory with the version you compiled.

EDIT: fixed the \< and \> in the text. 2019/01/17


justnope:
Does anyone want to take this over?

PR #4785, opened 2019-01-21
MIDI import/export: map note volume and MIDI velocity with their full range

### LINKED ISSUE #4761: Export MIDI - Note velocity
Our note velocity go from 0 - 200 but midi note velocity is 0 - 127. In our export code we set anything over 127 to 127 so a track using the full range in lmms will export wrong.

https://github.com/LMMS/lmms/blob/7f0593c601a7f48fda193139f2d6bb3d76f10384/plugins/MidiExport/MidiExport.cpp#L277

Example issue reported in the forum: https://lmms.io/forum/viewtopic.php?f=7&t=29497

Can we just divide note velocity in two when exporting?
Example, crude but working:
```cpp
mnote.volume = qMin(qRound(base_volume * 0.635 * LocaleHelper::toDouble(note.attribute("vol", "100"))), 127);
```
@PhysSong

PhysSong:
Fixes #4761.

### DIFF
--- a/plugins/MidiExport/MidiExport.cpp
+++ b/plugins/MidiExport/MidiExport.cpp
@@ -274,7 +274,8 @@ void MidiExport::writePattern(MidiNoteVector &pat, QDomNode n,
 		// TODO interpret pan="0" fxch="0" pitchrange="1"
 		MidiNote mnote;
 		mnote.pitch = qMax(0, qMin(127, note.attribute("key", "0").toInt() + base_pitch));
-		mnote.volume = qMin(qRound(base_volume * LocaleHelper::toDouble(note.attribute("vol", "100"))), 127);
+		 // Map from LMMS volume to MIDI velocity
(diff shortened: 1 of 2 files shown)


zonkmachine:
I'm going to just go ahead and merge.

jasp00:
According to MIDI 1.0 specification, which might be this [one](https://people.carleton.edu/~jellinge/m108f13/pages/00Reference/MIDIspecification.pdf), velocity is not strictly defined, but it should follow a logarithmic scale. I guess velocity = 64 should mean 100% in LMMS.

zonkmachine:
> velocity is not strictly defined

Correct. It's basically just info about how fast a key is struck.



> I guess velocity = 64 should mean 100% in LMMS.

And it is. (had to double check...)

jasp00:
Sure, 100% is 64. I was suggesting to consider a logarithmic scale instead of a linear one.

jasp00:
Should we use the algorithm from our favorite free MIDI player?

PR #4787, opened 2019-01-21
Disable built-in NaN handler through hidden setting

jasp00:
Disable built-in handler for NaN and huge values; this allows the user to control these cases. Default is to use the handler.

### DIFF
--- a/include/MixHelpers.h
+++ b/include/MixHelpers.h
@@ -33,6 +33,10 @@ namespace MixHelpers
 
 bool isSilent( const sampleFrame* src, int frames );
 
+bool useNaNHandler();
+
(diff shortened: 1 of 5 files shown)


zonkmachine:
Are you sure this should be a hidden setting?

zonkmachine:
Sorry, I really don't follow you now.

jasp00:
We agreed that this should be a hidden setting.

zonkmachine:
Right. And now I'm suggesting maybe it should not be hidden. I'm totally cool either way.

jasp00:
I prefer the setting to be hidden until a user asks otherwise because the setup dialog and the code are simpler.

PR #4443, opened 2018-06-21
Refactoring: Memory management

lukas-w:
Copying description from #4311, which for some reason GitHub won't let me reopen. I made these changes in November and April, but didn't have time to fix them on all platforms back then.

-----

Some refactorings regarding memory management and containers.

* `MemoryManager.h` renamed to `Memory.h`
* `MemoryHelper::alignedMalloc` moved to `Memory.h`, ported to an stl-compatible allocator called `AlignedAllocator`, implementation replaced by a single rpmalloc call.
* `LocklessAllocator` implementation replaced by a new `MemoryPool` class, which now also replaces the implementations of `BufferPool` (formerly `BufferManager`) and `NotePlayHandlePool` (formerly `NotePlayHandleManager`). Instead of a custom-written lockless stack, `MemoryPool` uses a 3rdparty lockless queue implementation which yields much better performance.
* Add some tests and naïve micro-benchmarks.

This includes a a new submodule, [`libcds`](https://github.com/khizmax/libcds), which is a collection of concurrent and lock-free data structures.

Starting with this PR, I'd like to license my contributions as MIT, while the project as a whole is still licensed as GPL. Hope that doesn't cause any trouble.

PR #4790, opened 2019-01-24
Prevent lmms freezing when a vst hangs

justnope:
I've added a detection to see if a VST hangs using timed mutexes. When a VST hangs it will block the msg semaphore, but lmms still sends an idleUpdate. That idleUpdate will also block and call invalidate on the plugin.

I've added m_process.kill in case the plugin is embedded. This is necessary because at that time the windows input queues are linked and one can block the other.


plugins/vst_base/RemoteVstPlugin32.cmake
```diff
@@ -63,5 +63,7 @@ ENDIF()
 IF(LMMS_BUILD_LINUX)
 	INSTALL(PROGRAMS "${CMAKE_CURRENT_BINARY_DIR}/../RemoteVstPlugin32" "${CMAKE_CURRENT_BINARY_DIR}/../RemoteVstPlugin32.exe.so" DESTINATION "${PLUGIN_DIR}")
 ELSEIF(LMMS_BUILD_WIN32)
-	INSTALL(PROGRAMS "${CMAKE_CURRENT_BINARY_DIR}/../RemoteVstPlugin32.exe" DESTINATION "${PLUGIN_DIR}")
+	INSTALL(FILES "${CMAKE_PREFIX_PATH_32}/bin/Qt5Core.dll" DESTINATION "${PLUGIN_DIR}/32")
+	INSTALL(FILES "${CMAKE_PREFIX_PATH_32}/bin/zlib1.dll" DESTINATION "${PLUGIN_DIR}/32")
+	INSTALL(PROGRAMS "${CMAKE_CURRENT_BINARY_DIR}/../RemoteVstPlugin32.exe" DESTINATION "${PLUGIN_DIR}/32")
```
PhysSong: I think these changes will conflict with #4797. I think it'd be better reverting them here.

plugins/vst_base/VstPlugin.cpp
```diff
@@ -152,7 +152,7 @@ VstPlugin::VstPlugin( const QString & _plugin ) :
 		tryLoad( "RemoteVstPlugin64" );
 		break;
 	case PE::MachineType::i386:
-		tryLoad( "RemoteVstPlugin32" );
+		tryLoad( "32/RemoteVstPlugin32" );
```
PhysSong: Here as well.

include/RemotePlugin.h
```diff
@@ -857,12 +889,12 @@ public slots:
...
 private:
 	void resizeSharedProcessingMemory();
-
+	bool m_disconnected;
 
 	QProcess m_process;
 	ProcessWatcher m_watcher;
 
-	QMutex m_commMutex;
+	std::recursive_timed_mutex m_commMutex;
```
lukas-w: What's the reason you changed to `std::recursive_timed_mutex`? I think we can keep using `QMutex`, avoiding the MinGW compatibility problems.

plugins/vst_base/VstPlugin.cpp
```diff
@@ -305,45 +306,51 @@ void VstPlugin::toggleUI()
 
 void VstPlugin::setTempo( bpm_t _bpm )
 {
-	lock();
-	sendMessage( message( IdVstSetTempo ).addInt( _bpm ) );
-	unlock();
+	if (try_lock())
```
lukas-w: Can we invert the control flow here and at the other places where `try_lock` is used to this:
```c++
if (! try_lock())
    return;
}
```
I think it's more natural this way because it doesn't have the normal case inside an `if` condition and as a bonus we avoid all the indentation changes, which makes reviewing easier.

plugins/vst_base/VstPlugin.cpp
```diff
@@ -391,9 +400,9 @@ bool VstPlugin::processMessage( const message & _m )
 #ifdef LMMS_BUILD_WIN32
 			// We're changing the owner, not the parent,
 			// so this is legal despite MSDN's warning
-			SetWindowLongPtr( (HWND)(intptr_t) m_pluginWindowID,
-					GWLP_HWNDPARENT,
-					(LONG_PTR) gui->mainWindow()->winId() );
+			//SetWindowLongPtr( (HWND)(intptr_t) m_pluginWindowID,
```
lukas-w: I'm not sure what this is for or why it was removed. @DomClark, you added this code in fcc883f8873974cdb8f3f09e715912a8ec039139. Can you chime in?

include/RemotePlugin.h
justnope: I would like to avoid QT at as much as possible when it's not related to the UI. In the longer term it would, hopefully, make it easier to separate lmms into two parts: a library and what it is now. 

If it's not possible to not use QMutex for this, I'll of course change it.

plugins/vst_base/VstPlugin.cpp
justnope: From my point of view:

It's complicated.

When you do stuff like this, it solves some problems but also creates others. What this does, short version, it links processes together in a single message queue and therefor when one hangs, the other one does too. 

I think this was the first version because something similar happens when the user selects "plugging embedding". I deleted it so nothing happens when they select "no embedding" and normally the same things happen when they select qt or win32 embedding.

plugins/vst_base/VstPlugin.cpp
justnope: Consider it done :)

include/RemotePlugin.h
lukas-w: I understand wanting to avoid Qt for core functionality. I'm just not very happy about the MinGW mutex support detection & workaround and would like to avoid it being copy&pasted around the CMake scripts.

What do you think about reverting this PR to `QMutex` for now until (a) we find a cleaner solution or (b) we drop support for old MinGW versions? An example of how (a) could look like can be seen in 57ca68194069c7cd0697a7f707b17b42bb4b648e which is part of #4443. In this commit, I added `include/mingw-std-threads/mutex` which shadows the system `mutex` using `include_directories(BEFORE include/mingw-std-threads)` when it's found necessary. If done correctly, this could fix MinGW thread support for all targets in the project without the need of `#ifdef USE_MINGW_THREADS_REPLACEMENT` checks at every include.

justnope: I think it's a bad idea, but I won't mind if you change it back to QMutex.

It's a bad idea because if we can't even sever the link between ui and core at this simple level, it will never be done.

plugins/vst_base/VstPlugin.cpp
DomClark: This is here to make plugin UIs always-on-top when not embedded. It does indeed attach the input queues, but that shouldn't cause any extra problems since the queues are attached in embed mode anyway. Also, plugins are generally designed to have the same input queue as the host, since most hosts load the plugin into the host's own process, so this isn't exactly a bad idea.

include/RemotePlugin.h
justnope: Give me a couple of days to think this over... I've deleted my previous msg because I responded to quickly; there may be other options.

Are we using mingw for any other reason than to cross-compile to windows?

This pr won't be merged any time soon, so we have all the time we need to find a solution that's clean and acceptable.

lukas-w: @justnope Don't worry about using QtCore functionality inside our core, this is completely fine and we'll probably always do it. It doesn't create any coupling between core and UI. The only reason I can see to use `std` mutex here is re-usability outside of LMMS. However, If you ever need to use this code outside of LMMS, you can simply swap `QMutex` for `std::recursive_timed_mutex`, their APIs are compatible.

> Are we using mingw for any other reason than to cross-compile to windows?

We use `winegcc` for `RemoteVstPlugin`, and we support MSYS2 on Windows which uses MinGW.

lukas-w:
What's the reason for closing this @justnope? In my eyes this PR is a valuable contribution, I'd very much like to see it merged.

justnope:
It's quicker to start from scratch than to revert everything.

justnope:
@lukas-w I don't know what to do.

Recap: This isn't a fix, this is a work-around. When @DomClark said it isn't a bad idea to have the input queues linked together, it actually made this work-around invalid. I explained it in discord and they agreed. If they already forgot what makes this work then maybe we should rethink the approach.

Long story short: A thread blocks because a vst blocks. Also: lmms transmits using win msg queues keep alives. I've added timeouts to detect blocks (in the msg queue). If it timeouts, we kill the process and whatever. This results in also unblocking the thread waiting for a reply. And presto, lmms doesn't freeze.

If Dom already forgot what the key point of this fix is, then maybe we should rethink it. QCore doesn't provide a timeout for what we use to communicate across processes. Therefor, I used this helter-skelter fix/work-around.

It's easy to reproduce. Attach a debugger to lmms, attach one to the vst process. In the debugger, hang the vst process; lmms should recover.

PR #4792, opened 2019-01-25
Don't let silent instruments wake up sleeping effects

### LINKED ISSUE #4564: Anormal CPU usage by VST's in 1.2.0-RC's
**Steps to reproduce :**
1) Launch LMMS 1.2.0 RC-5 or 6 (audio interface : SDL, plugin embedding : Native Win32 API)
2) Put 10*EasyQ VST on the master channel (for the sake of simplicity, I took EasyQ as an example, it's and I load it 10 times to simulate big projects)
3) Put a Vestige instrument from the instrument plugin tab (just put vestige on the project, no need to load a VSTi)
4) Look at the CPU meter
image

1) Launch LMMS 1.1.3 (audio interface : SDL)
2) Put 10*EasyQ VST on the master channel
3) Put a Vestige instrument from the instrument plugin tab
4) Look at the CPU meter
image

**More Infos :**
I'm not sure where the error could be, I don't even know if it's only me or if it's a real bug, but the result is that I can't open my biggest projects in LMMS 1.2.0-RC's because the CPU overload too quickly compared to LMMS 1.1.3
Seems like a regression to me.

**System config :**
Windows 7 64bits

If you need more informations, please let me know, I will do everything I can to help.

Thanks in advance,

T0NIT0 RMX

DomClark:
Fixes #4564.

Ensures a `PlayHandle`'s buffer is not silent before mixing it in to the audio port. This avoids waking up all downstream effects unless there is some audio present. The test is not performed for `NotePlayHandle`s for performance reasons, in line with similar code in `InstrumentTrack::processAudioBuffer`. https://github.com/LMMS/lmms/blob/c1ae1ed5f4246de40c3cfe5327d7dfbd5cd4470b/src/tracks/InstrumentTrack.cpp#L170-L175

Using a slightly different approach from https://github.com/LMMS/lmms/issues/4564#issuecomment-456479396 as suggested by gi0e5b06.

### DIFF
--- a/src/core/audio/AudioPort.cpp
+++ b/src/core/audio/AudioPort.cpp
@@ -119,7 +119,9 @@ void AudioPort::doProcessing()
 	{
 		if( ph->buffer() )
 		{
-			if( ph->usesBuffer() )
+			if( ph->usesBuffer()
+				&& ( ph->type() == PlayHandle::TypeNotePlayHandle
+					|| !MixHelpers::isSilent( ph->buffer(), fpp ) ) )
 			{
 				m_bufferUsage = true;
 				MixHelpers::add( m_portBuffer, ph->buffer(), fpp );


T0NIT0RMX:
I tried it and the bug seems still here :/

PR #4793, opened 2019-01-25
Ladspa

justnope:
These are the changes to make it compile with msvc.

When I reread the changes I realized I could've made it work without changing ladspa.h. It looks like that file came from somewhere externally so there might be good reasons to avoid changing it. CMake has the option to export all symbols from a library, so we could also do it from cmake instead of changing the header file.


justnope:
forgot to rebase

PR #4794, opened 2019-01-26
LADSPA msvc compatibility

justnope:
* added FindFluidSynth.cmake
* update submodules
* Added _USE_MATH_DEFINES in cmake
* removed hardcode perl in cmake

I've realised that instead of editing ladspa.h, I could've also edited the cmake file so it exports all symbols like gcc & clang does. Since ladspa.h came from outside of lmms, there might be good reasons to not allow any changes to that file.


src/tracks/Pattern.cpp
```diff
@@ -678,7 +678,7 @@ void PatternView::constructContextMenu( QMenu * _cm )
 
 	if( gui->pianoRoll()->currentPattern() &&
 			gui->pianoRoll()->currentPattern() != m_pat &&
-			not m_pat->empty() )
+			!m_pat->empty() )
```
PhysSong: I think now it's time to revert this.

PhysSong:
If the FluidSynth stuff is not necessary for this PR, I suggest making a separate request.

plugins/LadspaEffect/calf/CMakeLists.txt
```diff
@@ -36,16 +36,25 @@ SET_TARGET_PROPERTIES(veal PROPERTIES PREFIX "")
...
 
 # Don't strip if "Debug" or "RelWithDebInfo"
 IF(LMMS_BUILD_WIN32 AND NOT CMAKE_BUILD_TYPE MATCHES "Deb")
 	ADD_CUSTOM_COMMAND(TARGET veal POST_BUILD COMMAND "${STRIP}" "$<TARGET_FILE:veal>")
 ENDIF()
 IF(NOT LMMS_BUILD_APPLE AND NOT LMMS_BUILD_OPENBSD)
-	SET_TARGET_PROPERTIES(veal PROPERTIES LINK_FLAGS "${LINK_FLAGS} -shared -Wl,-no-undefined")
+	IF(MSVC)
+		SET_TARGET_PROPERTIES(veal PROPERTIES LINK_FLAGS "${LINK_FLAGS}")
+		TARGET_LINK_LIBRARIES(veal ${FLUIDSYNTH_LIBRARIES} Ws2_32)
```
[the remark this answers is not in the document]
justnope: No idea why I included those libraries. I'll remove the entire block.
Do I need to link against m in non msvc builds like tap & swh or is it ok?

src/tracks/Pattern.cpp
justnope: I'll rebase on master, then it should be gone in this PR

plugins/LadspaEffect/calf/CMakeLists.txt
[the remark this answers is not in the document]
PhysSong: There were no issues, so fine not to link against `libm`.

src/tracks/Pattern.cpp
PhysSong: Okay.

plugins/LadspaEffect/swh/CMakeLists.txt
```diff
@@ -1,18 +1,26 @@
 # Create blank config.h
 FILE(WRITE ladspa/config.h "")
+INCLUDE(FindPerl)
+
+#TODO: should we do something when perl isn't found? IF(PERL_FOUND) ...
```
PhysSong: I think we should print a warning message and `return()` in that case.

PR #3814, opened 2017-09-19
Enable HiDPI Scaling (Qt 5.6+ only) [stable-1.2]

flynn16:
This pull request is based on the `stable-1.2` branch. Hopefully it will solve the long-standing UI scaling issue of LMMS. Tested on Ubuntu 16.04 via an AppImage.

Issue #1814, opened 2015-03-03
Some settings revert to older when opening settings dialog


softrabbit:
```
$ grep -o 'mididev=".*"' ~/.lmmsrc.xml 
mididev="ALSA-Sequencer (Advanced Linux Sound Architecture)"
```
Start LMMS, change MIDI device in settings dialog, then:
```
$ grep -o 'mididev=".*"' ~/.lmmsrc.xml 
mididev="Dummy (no MIDI support)"
```
Open settings dialog, don't even look at the MIDI tab, just click "OK"
```
$ grep -o 'mididev=".*"' ~/.lmmsrc.xml 
mididev="ALSA-Sequencer (Advanced Linux Sound Architecture)"
```

Looks like this setting reverts to previously loaded data whenever the settings dialog is opened instead of what's been set (and saved!) previously. There might be other settings with similar behavior as well.

M374LX:
The same happens with the audio device selection.

Clicking on the "OK" button on the settings dialog calls SetupDialog::accept(), which sets the value of "mididev" on CongigManager. Nevertheless, the value of "mididev" is never read within SetupDialog. Again, the same happens with "audiodev".

tresf:
@softrabbit @mikobuntu Is this related at all to #1731?

mikobuntu:
@tresf  I will check this soon to see if the bugs are related. Regarding the original posters @softrabbit  bug, I think this is working as it should e.g dummy midi is a fallback and once LMMS picks up ( on restart ) an available midi port it will connect, which usually is the last one that would have been chosen. To test change form Alsa-Seq-Midi to Alsa-Raw-Midi and the settings will stick . I can/will test this further to be absolutely sure, but this is my gut instinct on this one :)

softrabbit:
@tresf, not related to #1731. That one's about MIDI connections, this is about MIDI driver selection in the setup dialog.

tresf:
Closed via #1832

PR #4800, opened 2019-01-28
Display properly on a HiDPI screen

hfiguiere:
With this LMMS display properly on a HiDPI screen

### DIFF
--- a/cmake/linux/lmms.desktop
+++ b/cmake/linux/lmms.desktop
@@ -8,7 +8,7 @@ Comment=Music sequencer and synthesizer
 Comment[ca]=Producció fàcil de música per a tothom!
 Comment[fr]=Production facile de musique pour tout le monde !
 Icon=lmms
-Exec=env QT_X11_NO_NATIVE_MENUBAR=1 QT_AUTO_SCREEN_SCALE_FACTOR=1 lmms %f
+Exec=env QT_X11_NO_NATIVE_MENUBAR=1 lmms %f
 Terminal=false
 Type=Application
 Categories=Qt;AudioVideo;Audio;Midi;
--- a/src/core/main.cpp
+++ b/src/core/main.cpp
@@ -251,7 +251,9 @@ int main( int argc, char * * argv )
 		return EXIT_FAILURE;
 	}	
 #endif
-
+#if QT_VERSION >= QT_VERSION_CHECK(5, 6, 0)
+	QCoreApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
+#endif
 	QCoreApplication * app = coreOnly ?
 			new QCoreApplication( argc, argv ) :
 					new MainApplication( argc, argv );
--- a/src/gui/GuiApplication.cpp
+++ b/src/gui/GuiApplication.cpp
@@ -54,11 +54,6 @@ GuiApplication* GuiApplication::instance()
 
 GuiApplication::GuiApplication()
 {
-	// enable HiDPI scaling before showing anything (Qt 5.6+ only)
-	#if (QT_VERSION >= QT_VERSION_CHECK(5, 6, 0))
-		QApplication::setAttribute(Qt::AA_EnableHighDpiScaling, true);
-	#endif
-	
 	// prompt the user to create the LMMS working directory (e.g. ~/Documents/lmms) if it doesn't exist
 	if ( !ConfigManager::inst()->hasWorkingDir() &&
 		QMessageBox::question( NULL,


PhysSong:
Thank you for the contribution! However, I believe #3814 already addresses what you want to achieve.

hfiguiere:
PR #1814 doesn't work (I actually tested it doesn't work). I suspect setAttribute is done too late. It as to be done before the QCoreApplication constructor.

hfiguiere:
Yes it does. But then it need to be reworked to remove the code that doesn't work. I'll do that later.

cmake/linux/lmms.desktop
```diff
@@ -8,7 +8,7 @@ Comment=Music sequencer and synthesizer
 Comment[ca]=Producció fàcil de música per a tothom!
 Comment[fr]=Production facile de musique pour tout le monde !
 Icon=lmms
-Exec=env QT_X11_NO_NATIVE_MENUBAR=1 QT_AUTO_SCREEN_SCALE_FACTOR=1 lmms %f
```
jasp00: According to [documentation](http://doc.qt.io/qt-5/qt.html), setting `QT_AUTO_SCREEN​_SCALE_FACTOR=1` is enough.

hfiguiere: Yes but not having to set it is better. And this would make thing not work if lmms was launched differently (ie from the command line).

Which is why the previous change might have been considered working.

jasp00: >Yes but not having to set it is better.

That reason does not convince me. We could drop `QT_X11_NO_NATIVE_MENUBAR=1`, but we do not.

hfiguiere: Here is *my* rationale:

HiDPI should work out of the box. Nothing to do for the user. `QCoreApplication::setAttribute()` didn't work before, it does now with the patch, make the environment variable `QT_AUTO_SCREEN​_SCALE_FACTOR=1` is not needed anymore. Adding it in the .desktop is only useful when the .desktop is used, which is not ideal.

FWIW, I submitted similar patches to Muse Sequencer and Rosegarden and they got accepted.

jasp00: You should detect whether the screen is HiDPI.

hfiguiere: Qt does that for you.

PhysSong: The Qt documentation says setting `Qt::AA_EnableHighDpiScaling` before constructing `QGuiApplication` is equivalent to setting `QT_AUTO_SCREEN​_SCALE_FACTOR` to 1, as @jasp00 said. However, I agree more with https://github.com/LMMS/lmms/pull/4800#discussion_r252891838. It'd be definitely better to have one which should always work and drop another fix which works only in some situations.

jasp00: It is not a fix. It is an opt-in feature that may incur some scaling or painting artifacts. See "[Migration of Existing Applications](http://doc.qt.io/qt-5/highdpi.html#migration-of-existing-applications)" for the real fix.

hfiguiere: but you were ready to have it except it wasn't properly activated. And without that you can't start fixing said artifacts.

tresf: I'm moving this conversation to the conversation area, where it really belongs.

tresf:
> The Qt documentation says setting `Qt::AA_EnableHighDpiScaling` before constructing `QGuiApplication` is equivalent to setting `QT_AUTO_SCREEN​_SCALE_FACTOR` to 1, as @jasp00 said. However, I agree more with [#4800 (comment)](https://github.com/LMMS/lmms/pull/4800#discussion_r252891838). It'd be definitely better to have one which should always work and drop another fix which works only in some situations.

Agreed.

> HiDPI should work out of the box. Nothing to do for the user. `QCoreApplication::setAttribute()` didn't work before, it does now with the patch, make the environment variable `QT_AUTO_SCREEN​_SCALE_FACTOR=1` is not needed anymore. Adding it in the .desktop is only useful when the .desktop is used, which is not ideal.

Agreed, this is how it works on Mac.  I'm curious.. will this fix the same issue for Windows clients?

> but you were ready to have it except it wasn't properly activated. And without that you can't start fixing said artifacts.

That change was reverted in the AppImage actually.  Apparently we (I) forgot to follow suit in the desktop icon. https://github.com/LMMS/lmms/pull/4026/commits/22124dd79ae6468cb24fc92e2097e407982f6992

Anyway, I've had to switch to `QT_SCREEN_SCALE_FACTORS="eDP-1=1.5;HDMI-1=1.5;"` to make LMMS usable on my machine.  I'll test out this PR.  Also, we need to know how it behaves on various Qt versions so we have recommendations for people if/when this backfires.

tresf:
I'm testing Ubuntu 18.04 w/ Qt 5.9 on a Dell XPS 13 and I can't get either technique to change anything on this machine.

I'm only attaching one screenshot because all tests were identical.  Note this display requires fractional scaling, it's not 2:1 like a Retina display.

I have a VM in Parallels on a Apple Retina display that suffers from the inconsistent scaling which I've seen people complain about with the AppImage.  I'll test that next with both this PR and the `QT_AUTO_SCREEN​_SCALE_FACTOR` setting.

screenshot from 2019-02-07 21-36-08

hfiguiere:
This is how it works on Fedora 29 with GNOME+Wayland with autoscale, on the XPS 13:

image

Without autoscale:

image

hfiguiere:
About compatibility, if Lmms is *build* on Qt < 5.6, this won't work, even if run on >= 5.6

tresf:
> About compatibility, if Lmms is _built_ on Qt < 5.6, this won't work, even if run on >= 5.6

Understood, I'm talking about the possibility of Qt addressing fixes between Qt 5.6 and now.  No matter what I do, I can't get Ubuntu 18.04 to recognize the `QT_AUTO_SCREEN_SCALE_FACTOR` setting.  I'm on Retina now running 2x scaling and my results are identical to yours regardless of whether or not `QT_AUTO_SCREEN_SCALE_FACTOR` is set.  I had to force scaling in the VM, so this may be hardware problem.  Unfortunately I can't test this PR on any of my machines because it offers no visual difference.  Forcing `QT_SCALE_FACTOR=2` makes most of the text the correct size, but some components are ballooned so large they're unusable.

tresf:
An interesting write-up: https://vicrucann.github.io/tutorials/osg-qt-high-dpi/

tresf:
@hfiguiere what Qt version are you using?

tresf:
So I just stumbled upon this bug report: https://github.com/linuxmint/Cinnamon/issues/4902

The issues are very similar to ours.  One person recommended this:

```
export QT_SCALE_FACTOR=1
export QT_AUTO_SCREEN_SCALE_FACTOR=0
export QT_SCREEN_SCALE_FACTORS=2
```

Strangely, this fixes scaling for my retina display.  I have a feeling that newer (Qt > 5.9) versions behave differently but I'll need to do more testing to confirm this.

hfiguiere:
In the Flatpak it is Qt 5.12.

tresf:
> In the Flatpak it is Qt 5.12.

That may be the difference then.  I'll test against 5.12 now.

hfiguiere:
Fedora 29 has 5.11.3 FWIW. And it works the same.

tresf:
The Qt version does not appear to make a difference on my machines.  The [above workaround](https://github.com/LMMS/lmms/pull/4800#issuecomment-461681138) is still the only setting that fixes the scaling on my retina display in VM.

Same for the XPS 13 (fractional scaling).

The good thing is, this PR doesn't make anything worse.

screen shot 2019-02-07 at 11 31 39 pm

tresf:
@hfiguiere are you on Wayland or X11?

hfiguiere:
Wayland. It's the default on Fedora.

tresf:
> Wayland. It's the default on Fedora.

Thought so.  Can you see if you experience the same improvements using X11?  My XPS requires fractional scaling (e.g. 125%) which I'm starting to believe is not supported in much except Windows.  To switch to X11 usually just requires logging out, clicking the gear and selecting X11 or whatever Fedora calls it. AFAIR the gear will only show if you have a password.

My VM is running Parallels which is current incapable of using Wayland.  I'd like to know if the display server is a factor.

tresf:
I booted Ubuntu 18.04 to X11 on the MacBook pro and can confirm that this patch fixes the scaling issue.  I can also confirm that it's functionally equivalent to the `QT_AUTO_SCREEN_SCALE_FACTOR` environmental variable for machines which honor it.

screenshot from 2019-02-08 05-29-06

In regards to the `QT_X11_NO_NATIVE_MENUBAR` comments above @jasp00 we'd love to remove that if there's a technique to accomplish the same thing using code but at the time of writing that, I could not find another approach.

This PR should be merged.

jasp00:
>In regards to the `QT_X11_NO_NATIVE_MENUBAR` comments above @jasp00 we'd love to remove that if there's a technique to accomplish the same thing using code

Do you still want that?

tresf:
> Do you still want that?

Want what?

tresf:
> Remove QT_X11_NO_NATIVE_MENUBAR if there's a technique to accomplish the same thing using code.


Ideally.  It makes our launcher harder to use our of the box with `linuxdeployqt`.  This is no invite to over-engineer anything.  The flag works fine, but if there's a simple way to do it without polluting our launcher that would be nice.

jasp00:
I will make a request after this one is merged.

tresf:
Per discussions above:

- Having this in code (NOT in our launcher) makes the setting more consistent across platforms 
- @hfiguiere tested with Fedora 29/Wayland/Qt 5.11.3
- @tresf tested with Ubuntu 18.04/X11/Qt 5.12.1
- @tresf tested with Ubuntu 18.04/X11/Qt 5.9
-  `Qt::AA_EnableHighDpiScaling` was set improperly.  This is best explained in the [Qt documentation](https://doc.qt.io/qt-5/qt.html#ApplicationAttribute-enum).
   > ... This attribute must be set **before**` QGuiApplication` is constructed. This value was added in Qt 5.6.
- In addition to the above testing, this was tested on two HiDPI systems that did not take advantage of scaling, but the results were identical before and after this PR.

Despite all of this, there was push back when this was set in the AppImage, so we should expect some pushback for edgecases where this setting makes things worse. We'll do our best to reproduce on similar hardware and handle accordingly.

Merging.

Issue #2556, opened 2016-02-09
Package building on Linux


pbella:
Something that would be very handy would be a means to locally build a Linux package (i.e. a .deb that can be built/installed locally, not THE Linux package) that would make it easier to install/test/remove a given build.  This may just be a documentation issue as the existing build process has a package build product that appears to be the input to the package building process but I haven't seen any docs on how to get to a .deb...

tresf:
If you want a `.deb`, this is a duplicate of #1620.

Not sure where the `ux` tag comes from.

Can you clarify this requirement please?
> not THE Linux package

If you're ok with a `.deb`, our wiki has a skeleton of how @Israel- has done this in the past.  We need help on this effort, so any work towards this goal is greatly appreciated.

pbella:
I'm looking for something similar to, and that would help with #1620, but a bit different: a Debian, rather than Ubuntu package (similar, but not necessarily the same)  To clarify what I mean (I'm spit-balling terms here as best I can given my current knowledge of the details of Linux packaging):

1) THE Linux binary package would be the one that gets uploaded to the Debian package repository (i.e. it presumably has some minimal level of testing, is properly signed, etc.)

2) A Linux binary package (what I was originally looking for) is just the ability to take a local build and put it into a .deb that can be installed locally (i.e. build it in a VM, package it up so that I can install it on my host machine for testing) and I think is needed for 1. (with the addition of the signing steps that are in his cheat sheet)

3) A Linux source package is needed for 2 so that it can be rebuilt from source.

I took a look at #1620 but this appears to be a cheat sheet to get to 1 for the Ubuntu repos assuming that 3 already exists. (unless I'm missing something, the cheat sheet doesn't cover how to actually create 3)  My original request was for how to take the binary build product from make to create #2 but let's change my request to how to take a snapshot of the source tree from git to create #3 since that will address my immediate need and sounds like it will help get to #1.

Once I figure out how to get 3, and then 2, done locally, I would be happy to volunteer to step up as the Debian package manager if that is a job that needs someone to fill.  From reading some of the earlier conversation on this subject, I think what Isreal was saying is that with the Debian package in place, since Ubuntu is based on Debian, the Ubuntu package maintainers life is easier.

If we don't have anyone available right now who knows the details of how to get from 3 to 1, I'll be happy to start doing some research... (one thing I would need help with doing a 64-bit build with support for 32-bit VSTs... that's something I have not yet been able to get working properly)

tresf:
[...]

PR #4813, opened 2019-02-06
Switch to Xenial build environment 

jasp00:
Switching to Xenial will make `stdatomic.h` available, which is part of C11. Then, we can update rpmalloc.

@falkTX, a Carla package for Xenial should be provided, but let us wait some days for feedback.


PhysSong:
> Switching to Xenial will make `stdatomic.h` available, which is part of C11. Then, we can update rpmalloc.

Yes, but won't it break compatibility for old systems? Trusty is still supported, though its lifetime will end soon.

For Qt, BTW, there are PPAs for Xenial as well as Trusty. See https://launchpad.net/~beineri for available versions. We can also work around for missing Carla with the `sed` trick like we did in `linux..before_install.sh` for now if @falkTX doesn't provide one.

falkTX:
> @falkTX, a Carla package for Xenial should be provided, but let us wait some days for feedback.

There is already one, let me know when you need it.

tresf:
> We can also work around for missing Carla with the `sed` trick like we did in `linux..before_install.sh` for now if @falkTX doesn't provide one.

Moving forward, we should use dependency pinning instead of sed replacements. https://help.ubuntu.com/community/PinningHowto#Examples.

.travis/linux..before_install.sh
```diff
@@ -1,9 +1,9 @@
 #!/usr/bin/env bash
 
-sudo add-apt-repository ppa:beineri/opt-qt592-trusty -y
-sudo add-apt-repository ppa:andrewrk/libgroove -y
-sudo sed -e "s/trusty/precise/" -i \
-	/etc/apt/sources.list.d/andrewrk-libgroove-trusty.list
+#sudo add-apt-repository ppa:beineri/opt-qt592-trusty -y
+#sudo add-apt-repository ppa:andrewrk/libgroove -y
+#sudo sed -e "s/trusty/precise/" -i \
+#	/etc/apt/sources.list.d/andrewrk-libgroove-trusty.list
```
tresf: Why is this commented out (and not removed)?

.travis/linux..install.sh
```diff
@@ -3,6 +3,7 @@
 PACKAGES="cmake libsndfile-dev fftw3-dev libvorbis-dev libogg-dev libmp3lame-dev
 	libasound2-dev libjack-jackd2-dev libsdl-dev libsamplerate0-dev libstk0-dev stk
 	libfluidsynth-dev portaudio19-dev g++-multilib libfltk1.3-dev
+	fluid
```
tresf: I assume xenial doesn't provide `fluid` along side `libfluidsynth-dev`?  Can you please keep the formatting consistent and place it on an existing line?

jasp00:
Review comments addressed in https://github.com/LMMS/lmms/pull/4813/commits/ed2c0136f969a1ca4b538cdcf549c17c53c320e6.

.travis/linux..install.sh
```diff
@@ -1,25 +1,25 @@
 #!/usr/bin/env bash
 
-PACKAGES="cmake libsndfile-dev fftw3-dev libvorbis-dev libogg-dev libmp3lame-dev
-	libasound2-dev libjack-jackd2-dev libsdl-dev libsamplerate0-dev libstk0-dev stk
-	libfluidsynth-dev portaudio19-dev g++-multilib libfltk1.3-dev
-	libgig-dev libsoundio-dev"
```
tresf: Please don't rearrange these.  The less lines we change, the easier this is to merge with other PRs, etc.

jasp00: What other PRs? Sorting this list is an improvement. One package per line would be even better.

jasp00: >Perhaps but it is out of scope.

I propose it now. Do you agree?

PhysSong: I'd say let's do that on `master` once we merge `stable-1.2` into it. It will reduce the merge conflicts.

PhysSong:
We've reverted `rpmalloc` via d9b5ee03fea807e70e74c5962b3f903ea598ed1c and the Trusty build works now.
The EOL of Trusty is indeed coming, so are there something that we should do?

JohannesLorenz:
I think testing on Xenial is more useful than on Trusty, because it will target more users (since Trusty has passed EOL). So I'm for merging this now (assuming it works and is reviewed).

Other opinions?

PhysSong:
It's okay to update to Xenial, but the issue with Carla should be resolved first.

JohannesLorenz:
> It's okay to update to Xenial, but the issue with Carla should be resolved first.

@PhysSong Which issue exactly? The OP just says a Carla build (of LMMS?) should be done. Do you mean that?

tresf:
> @PhysSong Which issue exactly? The OP just says a Carla build (of LMMS?) should be done. Do you mean that?

@JohannesLorenz Correct.  Carla still isn't available for Ubuntu 16.04 (I just checked).

falkTX:
why not build carla in 16.04 then?
can be done in any random PPA. the backend has no dependencies at all besides a compiler, and that is the only part needed for LMMS Carla plugin

tresf:
> why not build carla in 16.04 then?

We can, it just saves a considerable amount of CI build-time to grab a precompiled version and we don't do any PPA maintenance currently and we prefer that.

@falkTX I'm surprised it's not yet offered for 16.04.  Does 18.04 have a download for it or are musicians just not using it on those platforms?

tresf:
> we don't do any PPA maintenance currently and we prefer that.

Slight exception, Toby does still maintain the mingw PPAs for our win32 stuff, the rest we submodule (e.g. LADSPA stuff) or try to grab from the authors.

tresf:
> the backend has no dependencies at all besides a compiler

Maybe we'll submodule it then?  I vaguely remember some python-Qt stuff with how we launch it but it's been a while since I've visited the code.

falkTX:
> @falkTX I'm surprised it's not yet offered for 16.04. Does 18.04 have a download for it or are musicians just not using it on those platforms?

There was, for a very long time. Xenial was the distro base for the old "gcc5" packages in the kxstudio repos. Now I moved on to 18.04 as minimum dependency though.

Feel free to use this package https://launchpad.net/~kxstudio-debian/+archive/ubuntu/gcc5/+sourcepub/10337928/+listing-archive-extra

Or preferably copy the binary (without rebuild, no need since it uses xenial) from this link https://launchpad.net/~kxstudio-debian/+archive/ubuntu/gcc5/+copy-packages

PhysSong:
@falkTX Is it safe to use `kxstudio-debian/gcc5` PPA for Xenial?

falkTX:
No, I am not keeping that repo forever since it is now legacy.
Better to copy the package as I said. If you copy without rebuild, there should be no other dependencies since I build those as static libs

PhysSong:
@falkTX Thank you for clarification. Does that mean we should prefer copying binaries from the GCC 5 PPA rather than using that directly or the Trusty PPA?

tresf:
> @falkTX Thank you for clarification. Does that mean we should prefer copying binaries from the GCC 5 PPA rather than using that directly or the Trusty PPA?

It sounds like a simple submodule + build is the most sustainable solution, quoting:

>  the backend has no dependencies at all besides a compiler, and that is the only part needed for LMMS Carla plugin

PhysSong:
> It sounds like a simple submodule + build is the most sustainable solution

This will increase the time takes to build. Is that really better than managing a shared package repository?

tresf:
It's a trade-off.  I guess it depends on the build time.  We already do this for quite a few other plugins.

tresf:
>> @falkTX wrote:
>> the backend has no dependencies at all besides a compiler, and that is the only part needed for LMMS Carla plugin

> @tresf wrote:
> I guess it depends on the build time. We already do this for quite a few other plugins.

Benchmarking the following command on my laptop (i7, Clang compiler)

```
cd ~/Carla
time make -j4 backend

real	1m16.113s
user	3m27.355s
sys	0m12.821s
```

@falkTX is there a minimally viable way to do this that doesn't add ~3 minutes of compile time to user-space?

> @PhysSong wrote:
> Is that really better than managing a shared package repository?

@PhysSong at ~3 minutes, I'm starting to second guess this recommendation.  We technically [do this for Homebrew](https://github.com/Homebrew/homebrew-core/commit/87849b8e47931c5341078bc85e24bdd491229063), so it's not out of the question, but Homebrew has done a great job of automating their build system (e.g. use a robot to check for new versions) which makes the maintenance much lower from our perspective.  Homebrew includes the build recipe as well, so even newer or older systems can build from source quite seamlessly when the binaries aren't available.  I'm not really sure how much work the PPA is.  The only help we've ever gotten on PPAs was @Israel-D [several years ago](https://github.com/LMMS/lmms/issues/2556#issuecomment-183149356).

falkTX:
> @falkTX is there a minimally viable way to do this that doesn't add ~3 minutes of compile time to user-space?

Yes, you can add a few flags to disable a lot of stuff not needed just to build the library for linkage.

On my laptop:
```
$ time make -j2 backend
real    1m47,266s
user    3m15,592s
sys     0m7,431s
```
With a bunch of stuff forcibly disabled and debug on:
```
$ time make -j2 backend DEBUG=true HAVE_ALSA=false HAVE_DGL=false HAVE_FFMPEG=false HAVE_FLUIDSYNTH=false HAVE_HYLIA=false HAVE_LIBLO=false HAVE_LIBMAGIC=false HAVE_PULSEAUDIO=false HAVE_QT4=false HAVE_QT5=false HAVE_SNDFILE=false HAVE_X11=false
real    0m17,270s
user    0m32,308s
sys     0m2,878s
```

tresf:
> With a bunch of stuff forcibly disabled and debug on:

Thanks.  Testing on Mac, it says there's undefined symbols using those flags.  Also before it errors, the build time is at about 1 minute.

<details>
<summary>macOS compile log</summary>

```
Linking libcarla_standalone2.dylib
Undefined symbols for architecture x86_64:
  "_IOCreatePlugInInterfaceForService", referenced from:
      juce::(anonymous namespace)::createAppleRemoteInterface(unsigned int, void**) in juce_gui_extra.a(juce_gui_extra.mm.o)
  "_IOIteratorNext", referenced from:
      juce::(anonymous namespace)::getAppleRemoteDevice() in juce_gui_extra.a(juce_gui_extra.mm.o)
  "_IOObjectGetClass", referenced from:
      juce::(anonymous namespace)::createAppleRemoteInterface(unsigned int, void**) in juce_gui_extra.a(juce_gui_extra.mm.o)
  "_IOObjectRelease", referenced from:
      juce::AppleRemoteDevice::start(bool) in juce_gui_extra.a(juce_gui_extra.mm.o)
      juce::(anonymous namespace)::getAppleRemoteDevice() in juce_gui_extra.a(juce_gui_extra.mm.o)
  "_IOPMAssertionCreateWithName", referenced from:
      juce::ScreenSaverDefeater::PMAssertion::PMAssertion() in juce_gui_basics.a(juce_gui_basics.mm.o)
  "_IOPMAssertionRelease", referenced from:
      juce::ScreenSaverDefeater::PMAssertion::~PMAssertion() in juce_gui_basics.a(juce_gui_basics.mm.o)
  "_IOServiceGetMatchingServices", referenced from:
      juce::(anonymous namespace)::getAppleRemoteDevice() in juce_gui_extra.a(juce_gui_extra.mm.o)
  "_IOServiceMatching", referenced from:
      juce::(anonymous namespace)::getAppleRemoteDevice() in juce_gui_extra.a(juce_gui_extra.mm.o)
  "_kIOMasterPortDefault", referenced from:
      juce::(anonymous namespace)::getAppleRemoteDevice() in juce_gui_extra.a(juce_gui_extra.mm.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[1]: *** [../../bin/libcarla_standalone2.dylib] Error 1
make: *** [backend] Error 2
```
</details>

Since macOS doesn't suffer the same dependency issue as 16.04, I'll switch to Ubuntu 16.04 for now and report my progress. @falkTX should I file a proper bug report on the above macOS build issue?

falkTX:
> @falkTX should I file a proper bug report on the above macOS build issue?

yes please, though for this case please add `USING_JUCE=false` at the end of those commands, the build will be much faster. I did not realize this involved macOS as well.

falkTX:
oh and I forgot to say, with `USING_JUCE=false` you should not get link errors on macOS

tresf:
> I did not realize this involved macOS as well.

It doesn't, we have a Homebrew package that provides Carla for macOS, but it's my main machine.  I tested the command on Ubuntu 16.04 and it works just fine.

```
real	0m20.012s
user	0m27.534s
sys	0m4.962s
```

tresf:
With some help from @falkTX on #kxstudio (freenode) we came to a mutual conclusion that it may be easiest to just use a dummy Carla lib for linking purposes.

I haven't gotten far enough to test this, but the initial code is here: https://github.com/LMMS/lmms/compare/stable-1.2...tresf:carla

Currently, if Carla's not found on the system, it will build `DummyCarla.cpp` and use the submodule's include paths for building the plugin.

tresf:
The `DummyCarla.cpp` approach seems to be in pretty good shape, but it requires some testing.

I can cherry-pick the commits atop this branch to see if Travis-CI is OK, but we'll ultimately need a new `.AppImage` and a machine with Carla installed to know it's working.

tresf:
@jasp00 can you offer any insight as to why Debian is failing?

```
make[3]: *** No rule to make target 'plugins/libcarla_native-plugin.so', needed by 'plugins/libcarlabase.so'.  Stop.
make[3]: *** Waiting for unfinished jobs...
```

https://travis-ci.org/tresf/lmms/jobs/635470250#L3719

tresf:
Building a dummy Carla plugin is working well in my tests.
* Built and link using Xenial 16.04 using the `DummyCarla.cpp`
* Test using Bionic 18.04 with Carla installed from KXStudio

The two remaining issues are:
1. Debian builds are failing (haven't figured out why, asking for help)
2. For some reason the pixmaps for carlapatchbay and carlarack are now missing from LMMS.  I'm not sure if this is an edge-case bug with my build environment or if it's a problem I've created with the patches.<br>

Any help is greatly appreciated.

PhysSong:
@tresf I extracted `ADD_DEPENDENCIES(carlabase carla_native-plugin)` from @jasp00's commit ca7c36d68c3188e0935f1be538f1f364ec774c59 and tested it on top of this branch. That seems to fix Debian builds.

tresf:
I still can't explain why the [Carla plugin images went missing](https://github.com/LMMS/lmms/pull/4813#issuecomment-573233521), but it's fixed and all requests and issues have been addressed.

1. We have a weak-linked version of Carla if it's not provided for a particular platform, allowing `.AppImages` to build with Carla support on Xenial.
2. We've migrated the Linux build system from Ubuntu 14.04 "Trusty" to Ubuntu 16.04 "Xenial" as intended.

Merging.

plugins/carlabase/CMakeLists.txt
```diff
@@ -6,7 +6,29 @@ IF(NOT CMAKE_VERSION VERSION_LESS 3.9)
...
+# If Carla was not provided by the system, make a dummy library instead
+if(LMMS_HAVE_WEAKCARLA)
+  # Mimic the Makefile header
+  FILE(WRITE ${CMAKE_CURRENT_BINARY_DIR}/config.h "")
+  SET(CARLA_INCLUDE_DIRS
+    ${CMAKE_CURRENT_SOURCE_DIR}/carla/source
+    ${CMAKE_CURRENT_SOURCE_DIR}/carla/source/includes
+    ${CMAKE_CURRENT_SOURCE_DIR}/carla/source/utils
+    ${CMAKE_CURRENT_SOURCE_DIR}/carla/source/backend
+    ${CMAKE_CURRENT_BINARY_DIR}
```
PhysSong: @tresf This line will break MinGW builds once merged into `stable-1.2` because it affects the search order of `plugin_export.h`. Can we use some other path for the dummy `config.h`?

PhysSong:
Carla doesn't detect C++11 support in MSVC builds without `HAVE_CPP11_SUPPORT` defined as a truthy value, because [MSVC doesn't report a correct value of `__cplusplus` by default](https://docs.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=vs-2019).
We could simply disable Carla in MSVC builds, or we can find a workaround for this.

plugins/carlabase/CMakeLists.txt
tresf: > once merged into stable-1.2

It is merged... `tresf merged 13 commits into stable-1.2 from xenial 17 days ago`

> because it affects the search order of `plugin_export.h`

Can you help me understand where `plugin_export.h` is created?  I'm just not sure the exact issue.

> Can we use some other path for the dummy config.h?

So I assume then the problem is this line... 

```cmake
INCLUDE_DIRECTORIES(${CARLA_INCLUDE_DIRS})
```

Ideally, [`TARGET_INCLUDE_DIRECTORIES`] should really be favored but the build is farmed off to a plugin, so I wasn't aware of a standard way to do this.

I'm still a bit confused though.  Is the issue that another plugin that Mingw uses has a file also called `config.h`?  What build system is failing, is it `master`, not `stable-1.2`?

To answer the question, the file can be written anywhere, but I'm a bit confused why it would actually help.  I'd expect the `carlabase` build (i.e. `CMAKE_CURRENT_BINARY_DIR`) directory to be about as sane as it gets.

tresf:
> Carla doesn't detect C++11 support in MSVC builds without HAVE_CPP11_SUPPORT defined

Can't we just write it to [`config.h`](https://github.com/LMMS/lmms/pull/4813/files#diff-8e97a3188216f613f844021537ce7b6cR12)?  It would probably be best to detect-and-write it just incase.

plugins/carlabase/CMakeLists.txt
PhysSong: Sorry, I mean `master`.

tresf: @PhysSong if it's a search order problem, I'd expect it to be fixed by changing this... 

```cmake
- INCLUDE_DIRECTORIES(${CARLA_INCLUDE_DIRS})
+ INCLUDE_DIRECTORIES(AFTER ${CARLA_INCLUDE_DIRS})
```

... but cmake claims that the default behavior is already `AFTER`.  I guess I just don't understand how this is happening... Is it that the newer version of mingw is writing the same file in the same place?  If that's the case, can we just wrap a check around it?

tresf:
@PhysSong can you please offer some more context to the issues you've identified in this PR?  I'd love to help!

plugins/carlabase/CMakeLists.txt
PhysSong: I think `${CMAKE_CURRENT_BINARY_DIR}` is causing the problem. If we use something like `${CMAKE_CURRENT_BINARY_DIR}/carla`, that shouldn't interfere with the search order.

PhysSong:
@tresf If you're interested in those issues, you can refer to https://github.com/PhysSong/lmms/commits/master2. See the status of the last commit.

plugins/carlabase/CMakeLists.txt
tresf: Done via a77e592.

tresf:
> @tresf If you're interested in those issues, you can refer to https://github.com/PhysSong/lmms/commits/master2. See the status of the last commit.

Done via a77e592.  It took me a while to figure out the CI environment.  Apparently Circle-CI uses a docker image created by @lukas-w so getting a similar environment setup took me a while.

I made an [initial draft of the dependencies needed for Ubuntu 18.04 and mingw](https://github.com/LMMS/lmms/wiki/dependencies-windows#ubuntu-cross-compile).  Updates are welcome.

Issue #488, opened 2014-03-19
MDI Interface bug with Ubuntu Unity


tresf:
![screen shot 2014-03-18 at 11 40 49 pm](https://f.cloud.github.com/assets/6345473/2456505/640d92c0-af1a-11e3-8a7e-54eb7df97ee6.png)

As seen in the above picture, there is a bug with QMdiArea and Ubuntu's Unity interface which puts the Minimize/Restore/Maximize buttons in the upper left corner of LMMS (instead of the upper right) and erroneously attaches it to a movable toolbar.

I couldn't find anywhere on the internet explaining this QT/Unity combination behavior, but @tobydox suggested it was due to the Unity menu relocation, and eventually stumbled upon this article https://wiki.ubuntu.com/DesktopExperienceTeam/ApplicationMenu#Troubleshooting in the "Kubuntu Desktop Installation" provides the hint, which allows QT applications to disable the Desktop Environment from rendering it's own native menubar:

```
QT_X11_NO_NATIVE_MENUBAR=1 lmms
```

or for the `.desktop` file:

```
[Desktop Entry]
Exec=env QT_X11_NO_NATIVE_MENUBAR=1 lmms
```

> _Ubuntu 14.04 beta x86_
> 
> ![screen shot 2014-03-18 at 11 45 46 pm](https://f.cloud.github.com/assets/6345473/2456503/4c8d50ea-af1a-11e3-91ca-a6f6afec0d61.png)

**NOTE!** The env parameter is **REQUIRED** in the desktop shortcut, but not in terminal.

Before applying this patch, we should hear from other X11 based Desktop Environment users after testing this feature.

Is anyone running another X11 Desktop Environment that is able to try this? Please chime in if you can help test this!

PhysSong:
Recently I investigated this issue again. `QT_X11_NO_NATIVE_MENUBAR` is related to the `appmenu-qt` package which unsets `Qt::AA_DontUseNativeMenuBar` attribute. However, the package only affects Qt4 apps.

tresf:
This bug report suggests Qt5 honors `Qt::AA_DontUseNativeMenuBar` since 5.6.1. https://bugreports.qt.io/browse/QTBUG-28960

@PhysSong are you suggesting we switch to this method?  If so, this bug report should be reopened.

tresf:
Also, this should be a Linux-only (or Unity-only) setting.  MacOS doesn't suffer the same MDI window problems as Unity when the menu is removed..

[...]

PR #4818, opened 2019-02-09
Move QT_X11_NO_NATIVE_MENUBAR to code

jasp00:
Enable `QT_X11_NO_NATIVE_MENUBAR` by default. Clean up the desktop entry.

### DIFF
--- a/src/core/main.cpp
+++ b/src/core/main.cpp
@@ -194,11 +194,6 @@ void fileCheck( QString &file )
 
 int main( int argc, char * * argv )
 {
-#if defined( LMMS_BUILD_LINUX ) && QT_VERSION < 0x050000
-	if( setenv( "QT_X11_NO_NATIVE_MENUBAR", "1", 0 ) )
-		fprintf( stderr, "Error setting default environment.\n" );
-#endif
-
 	// initialize memory managers
 	NotePlayHandleManager::init();
 
@@ -256,6 +251,9 @@ int main( int argc, char * * argv )
 		printf( "LMMS cannot be run as root.\nUse \"--allowroot\" to override.\n\n" );
 		return EXIT_FAILURE;
 	}	
+#endif
+#if defined( LMMS_BUILD_LINUX ) && QT_VERSION < 0x050000
+	QCoreApplication::setAttribute( Qt::AA_DontUseNativeMenuBar );
 #endif
 #if QT_VERSION >= QT_VERSION_CHECK(5, 6, 0)
 	QCoreApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
--- a/src/core/main.cpp
+++ b/src/core/main.cpp
@@ -194,7 +194,7 @@ void fileCheck( QString &file )
 
 int main( int argc, char * * argv )
 {
-#ifdef LMMS_BUILD_LINUX
+#if defined( LMMS_BUILD_LINUX ) && QT_VERSION < 0x050000
 	if( setenv( "QT_X11_NO_NATIVE_MENUBAR", "1", 0 ) )
 		fprintf( stderr, "Error setting default environment.\n" );
 #endif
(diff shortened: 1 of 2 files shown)


PhysSong:
- I don't think the error message is descriptive enough.
- I found that the environment variable is meaningless when using Qt5. See #488 for details.

jasp00:
>* I don't think the error message is descriptive enough.

Make a proposal.

tresf:
@jasp00 thanks for the trick, this will come in useful.  For this particular setting, I recommend we remove the env and use code instead, thanks to @PhysSong's findings here:  https://github.com/LMMS/lmms/issues/488#issuecomment-462057876

tresf:
@jasp00 `Qt::AA_DontUseNativeMenuBar` is [part of the Qt4 documentation](https://doc.qt.io/archives/qt-4.8/qt.html).  Is there a reason why you did not use it?

jasp00:
I do not see a way to override `Qt::AA_DontUseNativeMenuBar`.

PhysSong:
> I do not see a way to override `Qt::AA_DontUseNativeMenuBar`.

I think @tresf meant [`QCoreApplication::setAttribute`](https://doc.qt.io/qt-5/qcoreapplication.html#setAttribute). Both ways work the same to me.
Quoting part of the source code of `appmenu-qt5`:
```cpp
void AppMenuPlatformMenuBar::createMenuBar()
{
    static bool firstCall = true;
    static bool envSaysNo = !qgetenv("QT_X11_NO_NATIVE_MENUBAR").isEmpty();
    static bool envSaysBoth = qgetenv("APPMENU_DISPLAY_BOTH") == "1";

    if (!m_menuBar->parentWidget()) {
        return;
    }

    m_adapter = 0;

    if (!firstCall && !envSaysBoth && QApplication::testAttribute(Qt::AA_DontUseNativeMenuBar)) {
        return;
    }

    if (envSaysNo) {
        if (firstCall) {
            firstCall = false;
            m_nativeMenuBar = NMB_DisabledByEnv;
            QApplication::setAttribute(Qt::AA_DontUseNativeMenuBar, true);
        }
        return;
    }
```

> > * I don't think the error message is descriptive enough.
> 
> Make a proposal.

Something like this?
```
Failed to set QT_X11_NO_NATIVE_MENUBAR environment variable. MDI interface might be buggy.
```

tresf:
... but the error message is irrelevant because we should be using the API, or remove it entirely since it does nothing for Qt4.

PhysSong:
> ... but the error message is irrelevant because we should be using the API, or remove it entirely since it does nothing for Qt5.

Okay. Now I'm fine with keeping the error message as-is.

tresf:
I'm confused.  The error message wouldn't be needed if using `QCoreApplication::setAttribute` which is more consistent with the hidpi PR.  Can we just use that and settle this?  How does an error message fit into that?

PhysSong:
@tresf I was talking about the error message we should use if we stick to `setenv`.

jasp00:
Ready.

src/core/main.cpp
```diff
@@ -252,6 +252,9 @@ int main( int argc, char * * argv )
 		return EXIT_FAILURE;
 	}	
 #endif
+#if defined( LMMS_BUILD_LINUX ) && QT_VERSION < 0x050000
```
tresf: I believe we should set this for all Qt versions with a `TODO` or `FIXME` comment that is currently ineffective for Qt5.

tresf:
Testing with Qt4:

Before:
image

After:
image

src/core/main.cpp
tresf: Done via f693bf6.

tresf:
Travis-CI shows failed, but that's just the Debian jobs.  Appears to be failing, not related to the codebase at all.  Merging.

PR #2798, opened 2016-05-25
fix the maximize behavior of subwindows on Mac

BaraMGB:
https://github.com/LMMS/lmms/issues/2450#issuecomment-217666180 @tresf reported an issue on mac. The subwindow don't maximize proper. The button don't change from maximize to restore. This PR fix this. 

**edit Explanation:**

I added a function for testing if the subwindow is maximized. It's quite simple: it checks if the subwindow size is identically to the size of the MdiArea. But we have to care about the Scrollbars of the MdiArea. If an other Subwindow is in Front of the maximized Subwindow, the MdiArea has Scrollbars (sometimes horizontal, sometimes vertical). Our Subwindow becomes smaller than the MdiArea although it is maximized.

PR #4819, opened 2019-02-10
Ensure the restore button of a subwindow shows up correctly 

PhysSong:
Fixes https://github.com/LMMS/lmms/issues/488#issuecomment-462098257 correctly, as described in https://github.com/LMMS/lmms/issues/488#issuecomment-462101363. Also removes a Mac-specific workaround introduced in #2798.


tresf:
Tested on macOS, Ubuntu 14.04, both pass.

PR #4705, opened 2018-11-13
Allow mid-note automation of BitInvader sample length

DouglasDGI:
This is one of the things requested in #4238.  It simply updates the sample length value while in the middle of playing a note in BitInvader rather than only reading the value at the beginning of the note.  You can get some very cool sounds out of this!

PR #4824, opened 2019-02-13
Fix Bitinvader waveform cutoff behavior, and allow realtime changing of BitInvader's sample length

### LINKED ISSUE #3797: Support various behaviors when increasing sample length after drawing a waveform in BitInvader
To re-produce:
1. Load a default BitInvader.
2. Decrease the sample length (e.g., to `64`).
3. Draw a waveform or click any waveform button (e.g., sine wave button).
4. Increase the sample length (e.g., to `128`).

The odd behavior:
image

The expected behavior:
capture20170908175912396

DouglasDGI:
(This includes #4705 and also fixes #3797, they're combined because I am painfully unaware of how to actually use GitHub.  Sorry.)

This PR restores BitInvader's 1.1.3 behavior when clicking a waveform button, so everything not displayed in the graph will be cut off.  It also makes it so mid-note automation of the sample length knob applies the changes to notes that are already playing, allowing you to do cool things like automate pulse width and other things with BitInvader.

As long as my code isn't horrendous, I don't think there are any problems with this, with the exception of a possible problem @zonkmachine may have found, here: https://github.com/LMMS/lmms/pull/4705#issuecomment-459977039.  

I apologize for my lack of GitHub-fu skills.


zonkmachine:
https://github.com/LMMS/lmms/issues/3797 should be fixed in stable-1.2
And yeah, git and GitHub takes some getting used to.

PhysSong:
@DouglasDGI Do you really need to clear whole samples, not only invisible part? I think you can remove the need for copying in `usrWaveClicked`.

DouglasDGI:
@PhysSong Yeah, you're right, I was being kind of dumb yesterday.  It should be fixed now.

PhysSong:
Do you mind if I cherry-pick the relevant changes to `stable-1.2` while keeping the author(@DouglasDGI you)?

DouglasDGI:
@PhysSong I don't mind, that would be great.

PhysSong:
@DouglasDGI Done via 8d707df171fa89c8426dca2d0d42de744d5db365, thank you for the work!

PR #4825, opened 2019-02-14
lmms_math: Fix build with musl

makepost:
When deciding to polyfill glibc features, check if the standard library claims to be glibc, instead of enumerating platforms in the condition. Unlike master branch which in de3b344 changes math function calls to standard, stable-1.2 keeps their glibc names and fails to build on Linux with another libc such as musl.

### DIFF
--- a/include/lmms_math.h
+++ b/include/lmms_math.h
@@ -34,7 +34,7 @@
 #include <cmath>
 using namespace std;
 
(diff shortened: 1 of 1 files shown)


zonkmachine:
Merge?

jasp00:
A build pass should be proposed if musl is supported.

PR #4840, opened 2019-02-22
Sync Debian version

jasp00:
Set Debian version following rules from `cmake/modules/VersionInfo.cmake`.

Fixes release issue uncovered by https://github.com/LMMS/lmms/commit/55eb831507250b236de80e63d376083dde0b3ae1.

### DIFF
--- a/.travis/linux.debian-sid.script.sh
+++ b/.travis/linux.debian-sid.script.sh
@@ -61,8 +61,6 @@ sync_version() {
 
 	sed "1 s/@VERSION@/$VERSION/" -i debian/changelog
 	echo Set Debian version to $VERSION
-# temporary
-head debian/changelog
 }
 
 sync_version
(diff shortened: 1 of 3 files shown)


jasp00:
Merging...

.travis/linux.debian-sid.script.sh
```diff
@@ -31,6 +31,40 @@ else
 	sudo pbuilder --update --basetgz "$BASETGZ"
 fi
 
+sync_version() {
+	local VERSION
+	local MMR
+	local STAGE
+	local EXTRA
```
PhysSong: `shellcheck` says `local` is undefined in the POSIX shell. Most shells(bash, dash, etc.) support the keyword though.
Since `master` checks shell scripts during CI builds using `shellcheck`, we have three choices(`stable-1.2` isn't affected):
- Specify a shell in the shebang which supports `local`
- Do nothing but silence `shellcheck` warnings
- Use a portable way for declaring local variables, if any

PR #4842, opened 2019-02-22
Move apt_mingw_cache out of build directory

jasp00:
Source tarball contains `apt_mingw_cache`, which is unnecessary. So move this cache directory to a better location.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ dist: trusty
 sudo: required
 cache:
     directories:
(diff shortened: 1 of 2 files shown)


tresf:
Does this need a subsequent commit to test since the cache has moved?

jasp00:
No, restarting one Travis job is enough. You can [see](https://travis-ci.org/LMMS/lmms/builds/497170300) that job #7590.1 uses the cache and takes less time.

PR #4847, opened 2019-02-23
Test deployment preparation in regular builds 

jasp00:
Deployment steps should be tested in regular builds because contributions may break this process inadvertently.

### DIFF
--- a/.travis/linux..script.sh
+++ b/.travis/linux..script.sh
@@ -7,10 +7,11 @@ if [ $QT5 ]; then
 	source /opt/qt59/bin/qt59-env.sh
 fi
 
-cmake -DUSE_WERROR=ON $CMAKE_FLAGS ..
+cmake -DCMAKE_INSTALL_PREFIX=../target/ -DUSE_WERROR=ON $CMAKE_FLAGS ..
 
 make -j4
 make tests
 ./tests/tests
 
-sudo make package
+make install
+make appimage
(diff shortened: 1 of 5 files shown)


jasp00:
Deployment on macOS does not work; fixing this is out of scope.

.travis/linux..script.sh
```diff
@@ -12,3 +12,5 @@ cmake -DUSE_WERROR=ON $CMAKE_FLAGS ..
 make -j4
 make tests
 ./tests/tests
+
+sudo make package
```
tresf: we don't use `make package`, we use `make appimage`.

tresf:
> Deployment on macOS does not work; fixing this is out of scope.

No, it's not.

.travis/osx..script.sh
```diff
@@ -12,3 +12,6 @@ cmake $CMAKE_FLAGS -DUSE_WERROR=OFF ..
 make -j4
 make tests
 ./tests/tests
+
+# TODO
```
tresf: We use `make dmg`. Note, the prefix path must be configured or else bad things will happen.  This is all documented here: https://github.com/LMMS/lmms/wiki/Compiling#macos-packaging

jasp00:
Ready.

tresf:
I wasn't aware that the `linuxdeployqt` was compatible with qt4, but surprisingly the builds pass.  This looks good to merge, thanks.  This will catch bundling issues earlier.

PR #4855, opened 2019-02-26
Load projects without stealing focus

### LINKED ISSUE #3769: LMMS steals focus when loading
Not sure of the proper wording for this. If another window is in focus while LMMS loads, LMMS will regain focus whenever a new part of loading completes. I believe this is undesirable.

Tested on Windows 10 with 1.2.0-rc2

Spekular:
Closes #3769 using PhysSong's suggestion (Setting WA_ShowWithoutActivating).

### DIFF
--- a/plugins/vestige/vestige.cpp
+++ b/plugins/vestige/vestige.cpp
@@ -115,7 +115,28 @@ public:
 		if ( embedMethod() != "none" ) {
 			m_pluginSubWindow.reset(new vstSubWin( gui->mainWindow()->workspace() ));
 			VstPlugin::createUI( m_pluginSubWindow.get() );
-			m_pluginSubWindow->setWidget(pluginWidget());
+
+			QWidget* pWidget = pluginWidget();
+			//WA_ShowWithoutActivating is apparently not implemented for any
+			//platform in Qt 5. In Qt 4 setting this attribute led to passing
+			//SW_SHOWNOACTIVATE to ShowWindow()
+			//The window flag Qt::WindowDoesNotAcceptFocus should be equivalent
+			//to WA_ShowWithoutActivating
+			pWidget->setAttribute(Qt::WA_ShowWithoutActivating);
+			pWidget->setAttribute(Qt::WA_X11DoNotAcceptFocus);
+			pWidget->setWindowFlags(Qt::WindowDoesNotAcceptFocus);
+			pWidget->setWindowFlags(Qt::WindowStaysOnBottomHint);
+			pWidget->setFocusPolicy(Qt::NoFocus);
+			pWidget->setWindowState(Qt::WindowMinimized);
+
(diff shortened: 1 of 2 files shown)


Spekular:
Focus still seems to be stolen when VSTs load.

Spekular:
This version only steals focus once, and then doesn't do it again after that. However, it also completely breaks VST UIs, so they can't be opened. Will revert stuff tomorrow and isolate the change that "fixes" loading. I suspect the commented out

```
m_plugin->createUI(nullptr);
m_plugin->showUI();
```

which is also obviously breaking VSTs. Simply commenting out m_plugin->showUI() doesn't fix the issue, but does prevent VST GUIs from flashing briefly onscreen during loading. Probably also prevents them from showing when a user loads a new VST normally, so would need to fix that somehow.

Spekular:
Some findings:

- Commenting out `m_pluginSubWindow->setWidget(pluginWidget());` in vestige.cpp prevents LMMS from popping into the foreground (except once near the start of loading). It also breaks VST GUIs (still).
- Even when the above "fix" is applied `emit dataChanged()` seems to steal focus without popping the window to the foreground. If you're typing in a text document, for example, it'll stop receiving input.

If there's a way to prevent the issue with `m_pluginSubWindow->setWidget(pluginWidget());`, I think this should be fixable. I'd probably try adding some sort of condition (parameter to loadFile?) to skip firing dataChanged during project load.

Spekular:
The following lines in createUI (inside ``if ( embedMethod != "none" )`` can run without the repeated foreground-pop:
```
m_pluginSubWindow.reset(new vstSubWin( gui->mainWindow()->workspace() ));
VstPlugin::createUI( m_pluginSubWindow.get() );
QWidget* pWidget = pluginWidget();
```

So at the very least setWidget() seems to be the problematic line rather than pluginWidget().

Spekular:
Can't get this to work, so I'm pushing the current "desperate last attempt" state of the code and closing this PR.

- `setWidget` pops LMMS to foreground no matter what I've tried.
- Removing `setWidget` breaks VST UIs as expected, and LMMS still steals focus even though it doesn't pop to foreground.
- Removing `emit dataChanged` *and* `setWidget` prevents both focus stealing and popping to foreground... Except that it still happens once per load. In my experience, this focus stealing happens far enough into the loading progress that it ruins the benefit of LMMS loading in the background. Not to mention `emit dataChanged` still needs to be called after vestigeInstrument.loadFile. This function can be called outside of project loading (user loading new VST) so adding `emit dataChanged` to the end of project load doesn't cover that case.

Something I considered investigating but decided against, doubt it would fix anything and might need a lot of rewriting:

> The most common way to construct a QMdiSubWindow is to call QMdiArea::addSubWindow() with the internal widget as the argument. You can also create a subwindow yourself, and set an internal widget by calling setWidget().

I'm tempted to say this isn't our bug based on what I've seen when investigating focus theft in QT.

PR #4870, opened 2019-03-03
Fix #3926: QCursor in AFP

### LINKED ISSUE #3926: QCursor in AFP
```
Thread 1 "lmms" received signal SIGSEGV, Segmentation fault.
#0  0x00007ffff4ee6536 in QCursor::shape() const () from /usr/lib/x86_64-linux-gnu/libQt5Gui.so.5
#1  0x00007fffc858436d in AudioFileProcessorWaveView::mouseMoveEvent (this=0x555557fe2d10, _me=0x7fffffffcf90)
    at lmms/plugins/audio_file_processor/audio_file_processor.cpp:877
#2  0x00007ffff74f9918 in QWidget::event(QEvent*) () from /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5
```

michaelgregorius:
Fix a crash that occurred on the following steps:
1. Add an AFP track.
2. Open it, and move the waveform display to overlap the track label
button.
3. Close the AFP window and open it again by clicking the track label.
4. Move the mouse pointer.

The problem occurs because the code makes the implicit assumption that
`AudioFileProcessorWaveView::enterEvent` (and hence
`QApplication::setOverrideCursor`) is called before
`AudioFileProcessorWaveView::mouseMoveEvent`. This is not the case when
the waveform display is on top of the track label. In this case the AFP
windows is opened with the mouse being immediately positioned over the
wave form display. There is no enter event and move events are issues
directly. This then leads to a crash in
`AudioFileProcessorWaveView::mouseMoveEvent` when trying to determine the
value for `is_size_cursor` because the override cursor is still `null` but
is dereferenced directly without checking.

Only adding a check would not solve the problem because in that case the
cursor would not change to the hand cursor when being moved inside the
waveform display.

The solution is to remove all calls to the global methods
`setOverrideCursor` and `restoreOverrideCursor` and to only set the cursor
locally.

This fix is based on a patch by @gi0e5b06 which is committed under 8a10c52
in his repository but for which he seems to never have created a pull request.

### DIFF
--- a/plugins/audio_file_processor/audio_file_processor.cpp
+++ b/plugins/audio_file_processor/audio_file_processor.cpp
@@ -795,6 +795,7 @@ AudioFileProcessorWaveView::AudioFileProcessorWaveView( QWidget * _parent, int _
 
 	m_graph.fill( Qt::transparent );
 	update();
+	updateCursor();
 }
 
 
@@ -811,7 +812,7 @@ void AudioFileProcessorWaveView::isPlaying( f_cnt_t _current_frame )
 
 void AudioFileProcessorWaveView::enterEvent( QEvent * _e )
 {
-	QApplication::setOverrideCursor( Qt::OpenHandCursor );
+	updateCursor();
 }
 
 
@@ -819,10 +820,7 @@ void AudioFileProcessorWaveView::enterEvent( QEvent * _e )
 
 void AudioFileProcessorWaveView::leaveEvent( QEvent * _e )
 {
-	while( QApplication::overrideCursor() )
-	{
-		QApplication::restoreOverrideCursor();
-	}
+	updateCursor();
 }
 
 
@@ -850,7 +848,7 @@ void AudioFileProcessorWaveView::mousePressEvent( QMouseEvent * _me )
 	else
 	{
 		m_draggingType = wave;
-		QApplication::setOverrideCursor( Qt::ClosedHandCursor );
+		updateCursor(_me);
 	}
 }
 
@@ -862,7 +860,7 @@ void AudioFileProcessorWaveView::mouseReleaseEvent( QMouseEvent * _me )
 	m_isDragging = false;
 	if( m_draggingType == wave )
 	{
-		QApplication::restoreOverrideCursor();
+		updateCursor(_me);
 	}
 }
 
@@ -873,22 +871,7 @@ void AudioFileProcessorWaveView::mouseMoveEvent( QMouseEvent * _me )
 {
 	if( ! m_isDragging )
 	{
-		const bool is_size_cursor =
-			QApplication::overrideCursor()->shape() == Qt::SizeHorCursor;
-
-		if( isCloseTo( _me->x(), m_startFrameX ) ||
-			isCloseTo( _me->x(), m_endFrameX ) ||
-			isCloseTo( _me->x(), m_loopFrameX ) )
-		{
-			if( ! is_size_cursor )
-			{
-				QApplication::setOverrideCursor( Qt::SizeHorCursor );
-			}
-		}
-		else if( is_size_cursor )
-		{
-			QApplication::restoreOverrideCursor();
-		}
+		updateCursor(_me);
 		return;
 	}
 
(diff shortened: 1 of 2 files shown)


jasp00:
>This fix is based on a patch by @gi0e5b06 which is committed under 8a10c52
in his repository but for which he seems to never have created a pull request.

Please close this request and start from scratch. See [LICENSE.txt](https://github.com/gi0e5b06/lmms/blob/master/LICENSE.txt).

zonkmachine:
> > This fix is based on a patch by @gi0e5b06 which is committed under [8a10c52](https://github.com/LMMS/lmms/commit/8a10c52b506ea828ffea3f8a545ed5857132674a)
> > in his repository but for which he seems to never have created a pull request.
> 
> Please close this request and start from scratch. See [LICENSE.txt](https://github.com/gi0e5b06/lmms/blob/master/LICENSE.txt).

He first commits the AFP fix in on Oct 31, 2017 8a10c52b506ea828ffea3f8a545ed5857132674a .
The license at this point is: https://github.com/LMMS/lmms/blob/8a10c52b506ea828ffea3f8a545ed5857132674a/LICENSE.txt
Then he changed the license in 7cc8efeb9d8ff3613cde81c5de1fba2212aa2c0d committed on Oct 10, 2018. That's almost a year later. I think we're in the clear here.

michaelgregorius:
What kind of GPL license does LMMS use? According to the [compatibility matrix](https://www.gnu.org/licenses/gpl-faq.html#AllCompatibility) there seems to be "GPL v2" and "GPL v2 and later".

In case LMMS uses "GPL v2" (as opposed to "GPL v2 and later") then it seems not to be OK to clone the v2 repository and to put it under a v3 license in the first place.

zonkmachine:
> In case LMMS uses "GPL v2" (as opposed to "GPL v2 and later") then it seems not to be OK to clone the v2 repository and to put it under a v3 license in the first place.

It looks like this is the case. He's actually forked the project, changed name to lsmm...
https://github.com/gi0e5b06/lmms/blob/master/CMakeLists.txt#L3
and with a, possibly, not compliant license. He should fix the license and also change name on his project so there is as little chance of mix-up between the projects as possible.

zonkmachine:
I gave this PR a quick test and couldn't replicate the issue.

tresf:
Note, any borrowing of code from a GPL3+ project needs permission from the author to be downgraded to GPL2+.  @gi0e5b06 has been banned from several communication channels but he hasn't been banned from a few of our Discord channels so you can DM him or tag him in a non-developer channel.

In regards to his codebase, he's legally allowed to upgrade LMMS to GPL3+.

jasp00:
Please do not merge without an explicit permission from gi0e5b06.

lukas-w:
@jasp00 As was already pointed out, the code this is based on is published under GPL2+. We don't need approval.

jasp00:
@michaelgregorius, did you download the patch by gi0e5b06 on a date before he changed the license?

PhysSong:
@jasp00 Per discussion on Discord #techtalk:

> [tresf]
> @gi0e5b06 does the project have permission to use your GPL3 source code in LMMS under the GPL2+ license? https://github.com/LMMS/lmms/pull/4870
> 
> [gi0e5b06]
> Yes. Anything commited before October 3rd, 2018 is GPL2+. Anything commited after should be considered GPL3+ by default. That said, LMMS should be upgraded to GPL3+ (even if it may not be needed until VST3).
>

michaelgregorius:
@jasp00, I did not download the patch. I had a look at his code and based my commit on his changes. In some places I copied his code for convenience (as can for example be seen in the changed header where I missed to replace the `NULL` with `nullptr`). In other places I cleaned up the code a bit, e.g. in the method `updateCursor`. So this pull request is a mix between his code and mine.

This was all done under the assumption that [this comment](https://github.com/LMMS/lmms/issues/3926#issuecomment-340729925) indicates that he wants the problem to be fixed and that the offered commit can be incorporated into LMMS code base. I was not even aware that his cloned repository has a changed license till after creating the pull request.

lukas-w:
Glad that's settled then. Merging. Thanks for the work and the excellent explanation @michaelgregorius!

PR #4872, opened 2019-03-03
VST fixes and improvements

### LINKED ISSUE #4802: Disable "always on top" behavior for VSTs
Request to disable "always on top" behavior of VSTs, or make it configurable.

This decision was made by @Umcaruje in #4110.  I'll quote:

> it would be beneficial if the VST window actually had a flag to always stay on top, since it disappears every time I click over to LMMS, which is bad UX IMO (Out of scope for this Pull Request though)

Although I agreed that switching is bad UX, covering up the software is also bad UX, so this depends on workflow.  The Windows/Linux/MacOS task bar allows quickly switching to a plugin which is arguably more convenient than minimizing it.

For example, a project with 10 VST UIs can cover up all of LMMS requiring a user to minimize many windows just to use the DAW.  If this behavior is kept, I vote that we keep it as a configurable flag, defaulting to off.

Feel free to cast your vote by clicking the "React" that matches below.

**VST Windows - Always On Top**
👍 - Keep current behavior ON, do NOT make configurable
😄  - Keep DEFAULT behavior to ON, make configurable
❤️ - Change DEFAULT behavior to OFF, make configurable
🚀  - Change DEFAULT behavior to OFF, do NOT make configurable

### LINKED ISSUE #4843: Don't show error message when loading empty VeSTige instance
If a VST plugin fails to load when a project is opened, LMMS will display the message "The VST plugin [plugin name] could not be loaded". However, the error check doesn't verify if there *should* be a plugin to load.

Steps to reproduce:
* Create a new project
* Place an instance of VeSTige in the song editor, without loading a VST
* Save the project
* Reload the project

An error message will show, stating "The VST plugin  could not be loaded". (Tested with 1.2 RC7/RC8)

If a VST plugin was never loaded in the first place, I'd expect the error message not to display.

### LINKED ISSUE #4861: VST plugins crash on Linux with VST sync enabled
I've seen a few reports of this on Discord and the forums. Apparently it appeared in rc8, tends to occur when loading projects, and is caused by a division-by-zero exception.

My guess is that this was introduced in 614bca7f04ce1ea35a50b0ac823b784907e4046a, and is due to alignment/padding issues in the `VstSyncData` struct. On Linux, `double`s are 4-byte aligned in 32-bit code and 8-byte aligned in 64-bit code. This gives a different layout of the structure in LMMS and RemoteVstPlugin. This doesn't occur on Windows as `double`s are always 8-byte aligned there.

DomClark:
d2854f674ac20f4c44fcbf83b35b3e9511587520 resolves #4802
b2a30c23cca3026a2c8d092e0fd64d1dbb2b36c6 fixes https://github.com/LMMS/lmms/issues/4802#issuecomment-458624067
fed062a13dcf0727aec1c7e3a5adff9ebaece627 fixes #4843
510d0b9112ce4b6dc91203f2723e355b2b05e5f9 fixes #4861

### DIFF
--- a/include/SetupDialog.h
+++ b/include/SetupDialog.h
@@ -126,6 +126,9 @@ private slots:
 	void toggleDisplayWaveform( bool en );
 	void toggleDisableAutoquit( bool en );
 
+	void vstEmbedMethodChanged();
+	void toggleVSTAlwaysOnTop( bool en );
+
 	void setLanguage( int lang );
 
 
@@ -207,6 +210,8 @@ private:
 
 	QComboBox* m_vstEmbedComboBox;
 	QString m_vstEmbedMethod;
+	LedCheckBox * m_vstAlwaysOnTopCheckBox;
+	bool m_vstAlwaysOnTop;
 } ;
 
 
--- a/include/VstSyncController.h
+++ b/include/VstSyncController.h
@@ -76,10 +76,10 @@ private slots:
 private:
 	struct VstSyncData
 	{
-		bool isPlaying;
 		double ppqPos;
 		int timeSigNumer;
 		int timeSigDenom;
+		bool isPlaying;
 		bool isCycle;
 		bool hasSHM;
 		float cycleStart;
--- a/include/VstSyncData.h
+++ b/include/VstSyncData.h
@@ -41,10 +41,10 @@
 
 struct VstSyncData
 {
-	bool isPlaying;
 	double ppqPos;
 	int timeSigNumer;
 	int timeSigDenom;
+	bool isPlaying;
 	bool isCycle;
 	bool hasSHM;
 	float cycleStart;
--- a/plugins/vestige/vestige.cpp
+++ b/plugins/vestige/vestige.cpp
@@ -181,7 +181,13 @@ vestigeInstrument::~vestigeInstrument()
 
 void vestigeInstrument::loadSettings( const QDomElement & _this )
 {
-	loadFile( _this.attribute( "plugin" ) );
+	QString plugin = _this.attribute( "plugin" );
+	if( plugin.isEmpty() )
+	{
+		return;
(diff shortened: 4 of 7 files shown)


PhysSong:
Merge?

jasp00:
I do not look at requests that fix more than one issue. Changes have been approved, so merge at your discretion.

DomClark:
> I don't really understand why b2a30c2 works, but the rest looks good to me 👍

@lukas-w It ensures the owner change is done before the window is shown. Why that is necessary I'm not sure, but it seems to work. I guess it might just be some quirk of Windows.

> Merge?

@PhysSong I haven't been able to test 510d0b9 because I'm having some issues building on Linux, so I'd like to get that done first. Would you be able to provide an AppImage?

> I do not look at requests that fix more than one issue.

@jasp00 The fixes have been split into separate commits. I'm not sure how splitting them into separate PRs affects reviewability, other than that you can't approve them independently. If you don't want to review, that's fine, since others have approved the changes as you stated. If you do want to review, and you have objections, feel free to reject all the changes and I will happily address any concerns.

PhysSong:
> @PhysSong I haven't been able to test [510d0b9](https://github.com/LMMS/lmms/commit/510d0b9112ce4b6dc91203f2723e355b2b05e5f9) because I'm having some issues building on Linux, so I'd like to get that done first. Would you be able to provide an AppImage?

I can, but it may take a day.

PhysSong:
@DomClark https://github.com/PhysSong/lmms/releases/download/vdummy/lmms-1.2.0-rc8.21-linux-x86_64.AppImage

DomClark:
Thanks for the AppImage @PhysSong. I've tested 510d0b9 and it seems to work, so I'd say this is good to merge now.

Issue #4385, opened 2018-05-31
[request] remove all midi connections from plugin


VirToReal:
Yeah, thats a request, not a real issue:
I'm using a novation Launch Control XL and connect it's knobs to a plugin, for example Triple OSC. After generating a nice sound, I have to disconnect every midi knob in order to save my creation as a preset. Because the midi connection will be saved in the preset-file (which is really great), all knob positions will be set to zero. 

If there would be a button to erase all connections, I can save one preset as midi-preset, using this for creating sounds via launch control, deleting all midi connections and save it as a real preset.

Maybe two file-formats? One for midi-connected presets, and one for non-midi-connected presets?

PhysSong:
Are there more use-cases for "remove all midi connections"? If preset saving is the only one, I can't find a reason for that button.
> Because the midi connection will be saved in the preset-file

Related: #2625

VirToReal:
Not really... just leaving the last midi-value on the knob while removing the connection would be great. Removing every connection would be nice, if there are, for example on the triple oscillator, all knobs connected to a midi-controller. Or maybe just a keyboard shortcut for removing the midi-connection while pressing the connected button/knob.

In each case of course, the midi-value should be placed on the actual value of the knob/button while removing the midi-connection.

Issue #2625, opened 2016-03-02
Presets remember the controller connections


IvanMaldonado:
Steps to reproduce:

1. Open a plugin
2. Connect any knob to a controller (LFO for example)
3. Save it as a preset
4. Open a new project
5. Open your preset
6. Try to move the knob you previously linked to a controller

You have to remove the connection to be able to use the knob.

zonkmachine:
Confirmed

zonkmachine:
Related to #2564

Issue #193, opened 2014-01-29
Saving preset with peak controller: instrument connecting to its controller


SadaleNet:
To reproduce:
- Create any instrument
- In the instrument, create a peak controller fx
- Connect volume knob to its peak controller
- Save the instrument as preset
- Load the preset back
- Result: The volume knob is connected to an undefined controller.

It seems that this bug is difficult to fix. This bug is not damaging tho.

curlymorphic:
> Going to bump this up to 1.2. Don't really want to touch peak controllers for now, since there's going to be lots of changes to controller code in 1.2...

@dizzy is this comment now aimed at 2.0?

DeRobyJ:
Status:
bug still here, but the problem is not really about peak controller, imo.

In general, if you save a preset with a control linked to something (e.g. to a controller created from the Rack), and then you open it on a new project, its options will show it is "Connected to <nothing>".
The difference with peaks controllers is that if you try to "edit connection", the program crashes, while if the controller was external it won't.
Fixing this would also fix this other small issue: #2625 

Second problem, which is more about effect instances in general:
when you preview a preset from the browser, and that instrument had a peak controller, a Peak Controller is created in the Rack.
Then, if you load the same preset as a track, a _second_ Peak controller is created.
In both cases, the connected knob fails to find that controller instance (I don't know how controller instances work, but maybe when you load one from a preset, a new instance ID is created, while the knob connection is asking for another ID? something like that...)
It may also have to do with the order in which the instrument is loaded.
(it should ideally load the peak controller first, then the other knobs that might be connected to it)

This is somewhat linked to #1644 
In general, preset previews tend to load effects (and sometime other things) and keep them there until they are triggered by something.

karmux:
Third problem (or different description of second problem): Peak controllers must be before LFO controllers in controller rack. If they are not then after saving and reloading project they get sorted automatically and some connections get LFO controller instead of Peak.

Order before save/load: Peak 1, LFO, Peak 2. Effect X connects to Peak 2.
Orger after save/load: Peak1 , Peak 2, LFO. Effect X connects to LFO now.

BaraMGB:
@karmux look at https://github.com/LMMS/lmms/issues/2869

MIntrovrt:
Can't replicate in 1.2.0-rc4
Is this still an issue or did no one close it, since the referenced issues here have been closed, I would assume this one can be closed to in order to avoid further confusion?

dan-giddins:
[...]

PR #4874, opened 2019-03-04
Add CLI flag to disconnect all MIDI controllers when loading a project

pboi20:
Hi!

Consider this PR as a request for comment, as this is my first attempt at hacking on LMMS.

This introduces a CLI flag (`--disconnect-midi-controllers`) used to disconnect all MIDI inputs from instrument tracks, as well as all MIDI controllers from instrument, FX and other global/song parameters. Other user controllers should be left untouched.

In and of itself, I find it useful to have a quick way to remove all midi connections from an existing project. For example, consider moving a project that has an extensive MIDI configuration for a specific hardware controller, over to another type of hardware controller. As I understand it, the current workflow would require disconnecting each individual controller mapping, one by one.

Also, it could be thought of as a quick fix for the following issues:
https://github.com/LMMS/lmms/issues/4385
https://github.com/LMMS/lmms/issues/2625
https://github.com/LMMS/lmms/issues/193

Is this idea of any interest to anyone else?

If so, I understand that this is probably a naive approach and would welcome your feedback!

Many thanks!

P.

### DIFF
--- a/include/Song.h
+++ b/include/Song.h
@@ -265,6 +265,16 @@ public:
 		return false;
 	}
 
+	inline void setDisconnectMidiControllersOnLoad( bool val )
+	{
+		m_disconnectMidiControllersOnLoad = val;
+	}
+
+	inline bool getDisconnectMidiControllersOnLoad()
+	{
+		return m_disconnectMidiControllersOnLoad;
+	}
+
 	void addController( Controller * c );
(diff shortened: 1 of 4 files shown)


PhysSong:
Welcome! First, I should say that `stable-1.2` is on a feature freeze. I think you should rebase commits against `master` if you can.

Issue #3605, opened 2017-06-03
Implementing oversampling in plug-ins.


Hussam-Eddin-Alhomsi:
• Instrument plug-ins (synthesizers only):
When playing high notes on a synthesizer using a harmonically rich wave like a saw wave, there are probably some frequencies that will go above the nyquist frequency, so aliasing occurs.

Just play the note C8 on TripleOscillator using a saw wave to hear the aliasing noise.

My idea about the GUI implementation is a bunch of radio buttons in the plugin tab of each synthesizer offering oversampling factors in the power of two that's "1x" (no oversampling), "2x", "4x", "8x", "16x", "32x" and "64x".

• Effect plug-ins: 
There are 2 types of plug-ins that would benefit from using oversampling (as far as I know).

(1) Waveshapers, distorters, overdrivers, bitcrushers and Lo-fi effects (or anything that generates a lot of harmonics).
When using any of the effects above on a sound, extra harmonics are being added to it, and if these harmonics go above the nyquist frequency, aliasing occurs.

A spectrogram of a heavily waveshaped sine wave. (Note the aliasing noise highlighted in yellow before the fundamental frequency of the sine wave.)
![distorted sine wave](https://cloud.githubusercontent.com/assets/25503477/26704782/30472e22-473a-11e7-9854-3ef5930b41cd.png)
A spectrogram of the same waveshaped sine wave but the waveshaper has 16x oversampling active. (Note that most of the aliasing that was present before activating oversampling is gone.)
![distorted sine wave with 16x oversampling](https://cloud.githubusercontent.com/assets/25503477/26704785/34b8acd8-473a-11e7-839e-220765215709.png)

My idea about the GUI implementation is a slider in each of these plug-ins mentioned earlier offering oversampling factors like "1x" (no oversampling), "2x", "4x", "8x", "16x", "32x" and "64x".

(2) Equalizers.
When a band in an equalizer gets close to the nyquist frequency, its shape distorts.
The distortion can happen with any band type.

A normal low-pass filter at 1000 Hz looks like this:
![lpf 1000 hz](https://cloud.githubusercontent.com/assets/25503477/26700675/ea4781d2-4726-11e7-9181-0b1ea01809da.png)
The same low-pass filter with the same resonance but at 16000 Hz looks like this:
![lpf 16000 hz](https://cloud.githubusercontent.com/assets/25503477/26700730/23bff5f2-4727-11e7-97f0-4439165ad109.png)

[...]

PR #4881, opened 2019-03-09
LMMS 1.2 Native Equalizer 2x Oversampling

DouglasDGI:
This adds a button that gives the LMMS 1.2 EQ 2x oversampling, preventing the curve distortion you get when you move the cutoff of any of the filters really high.

Here is a video to show the changes:
https://www.youtube.com/watch?v=9SH8IlLhXWs

Let me know if I messed anything up in the PR (still getting used to Github).


plugins/Eq/EqControlsDialog.cpp
```diff
@@ -208,6 +208,14 @@ EqControlsDialog::EqControlsDialog( EqControls *controls ) :
 	hpBtnGrp->addButton( hp24Button );
 	hpBtnGrp->addButton( hp48Button );
 	hpBtnGrp->setModel( &m_controls->m_hpTypeModel,false);
+
+	//Oversample
+	PixmapButton * oversampleButton = new PixmapButton( this, NULL );
+	oversampleButton->setActiveGraphic( PLUGIN_NAME::getIconPixmap( "oversampleon" ) );
+	oversampleButton->setInactiveGraphic(  PLUGIN_NAME::getIconPixmap( "oversampleoff" ) );
+	oversampleButton->setCheckable(true);
```
PhysSong: You should add `oversampleon.png` and `oversampleoff.png` to make the button usable.

PhysSong:
@DouglasDGI I think you oversampled samples without any interpolation. It causes aliasing and may result in undesired behavior.
[4881.mmp.zip](https://github.com/LMMS/lmms/files/2955541/4881.mmp.zip)
To solve the problem, you should oversample the input signal correctly using e.g. `libsamplerate`. You can find some good examples by searching for `src_process` in this repository.

plugins/Eq/EqControlsDialog.cpp
DouglasDGI: Whoops!  Completely forgot to add those.

PhysSong:
@DouglasDGI Do you think you can fix the aliasing issue? If you don't, can we close this for now and open a feature request for it?

DouglasDGI:
I'll go ahead and close this, I was very new to DSP in general when I made this and didn't even understand how oversampling actually worked.
A feature request doesn't need to be opened, it's already listed in some Equalizer meta issue or something if I remember right...

PhysSong:
Yes, I think #3605 mentions it.

PR #4883, opened 2019-03-09
Add CLI flag to disconnect all MIDI controllers when loading a project

pboi20:
This is a redo of #4874, using master as base branch.

==

Hi!

Consider this PR as a request for comment, as this is my first attempt at hacking on LMMS.

This introduces a CLI flag (--disconnect-midi-controllers) used to disconnect all MIDI inputs from instrument tracks, as well as all MIDI controllers from instrument, FX and other global/song parameters. Other user controllers should be left untouched.

In and of itself, I find it useful to have a quick way to remove all midi connections from an existing project. For example, consider moving a project that has an extensive MIDI configuration for a specific hardware controller, over to another type of hardware controller. As I understand it, the current workflow would require disconnecting each individual controller mapping, one by one.

Also, it could be thought of as a quick fix for the following issues:
#4385
#2625
#193

Is this idea of any interest to anyone else?

If so, I understand that this is probably a naive approach and would welcome your feedback!

Many thanks!

### DIFF
--- a/include/Song.h
+++ b/include/Song.h
@@ -304,6 +304,16 @@ public:
 		return false;
 	}
 
+	inline void setDisconnectMidiControllersOnLoad( bool val )
+	{
+		m_disconnectMidiControllersOnLoad = val;
+	}
+
+	inline bool getDisconnectMidiControllersOnLoad()
+	{
+		return m_disconnectMidiControllersOnLoad;
+	}
+
 	void addController( Controller * c );
 	void removeController( Controller * c );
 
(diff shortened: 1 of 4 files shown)


Reflexe:
The code looks fine. However, maybe we should add an export option that will not save MIDI connections to the file in first place? If that was your intention, I could try to implement  that in the next following days.

pboi20:
Hi, thanks for the feedback!

I also think that an export option to omit the MIDI connections would be a cleaner solution. I hesitated before sending this PR because my solution felt a bit naive and hacky.

If you have some time to look into it, I welcome your help in implementing this export option.

Reflexe:
Update: I have a working POC for a save dialog with a "discard MIDI connections" option. I'll polish and finish it and the next following days (hopefully tomorrow)

Reflexe:
Update: I've opened a PR at #5021; feel free to test it and report back.

pboi20:
Hi!

Just tested the master branch with all of your merged commits. It works very well.

Closing this PR because the new save option makes it irrelevant.

Many thanks,

P.

PR #4888, opened 2019-03-12
Avoid shallow clones in all Debian sid builds

jasp00:
Shallow clone may break version detection. This is fatal in Debian builds, so use full clone.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,8 +24,12 @@ matrix:
           git:
               depth: false
         - env: TARGET_OS=debian-sid TARGET_ARCH=i386
+          git:
+              depth: false
         - compiler: clang
           env: TARGET_OS=debian-sid
+          git:
+              depth: false
 before_install:
    - . ${TRAVIS_BUILD_DIR}/.travis/${TRAVIS_OS_NAME}.${TARGET_OS}.before_install.sh
 install:


jasp00:
Thank you, @PhysSong, for noticing this. Feel free to merge it.

tresf:
I'm fine with this being merged as-is, but shallow cloning saves quite a bit of time in our build system.  @jasp00 is there a value that would be acceptable that doesn't grab the entire history every time?  Travis-CI defaults to 50.    We' have > 6,000 in total.  Would 100 be enough?

PhysSong:
> Would 100 be enough?

I think that may not be enough for `master` once the version detection(#4840) is added into `master`.

tresf:
Ok.  Let's merge as-is then.  I did some local benchmarking (anecdotal) with the `time` command and there's about 5 seconds difference between `--depth 100` and no depth at all.

With Calf's submodule this is pretty important (it can take a very long time to clone without shallow)  but LMMS (currently) seems in good shape without it.

tresf:
> I think that may not be enough for master once the version detection(#4840) is added into master.

Since `master` would be the branch that benefits the most from shallow cloning, we should NOT merge this commit to `master`.  Instead on `master`, I vote we disable [`--recurse-submodules`](https://docs.travis-ci.com/user/common-build-problems/#git-submodules-are-not-updated-correctly) and let [`cmake` handle it at build time](https://github.com/LMMS/lmms/blob/master/cmake/modules/CheckSubmodules.cmake#L24).  Disclaimer, I'm not as familiar with `master`'s CI environment so I'm not sure how this would affect Circle-CI.

PhysSong:
> I'm not sure how this would affect Circle-CI

It seems like CircleCI doesn't clone/initialize submodules automatically by default.

> I vote we disable [`--recurse-submodules`](https://docs.travis-ci.com/user/common-build-problems/#git-submodules-are-not-updated-correctly) and let [`cmake` handle it at build time](https://github.com/LMMS/lmms/blob/master/cmake/modules/CheckSubmodules.cmake#L24) (and then merge this commit).

Sounds good.

PR #4895, opened 2019-03-14
Fix Blackman-Harris window formula, adjust formatting

he29-net:
This fixes the formula used to generate Blackman-Harris window created in the Eq plugin. Cosf() was replaced by cos() to make all 3 lines consistent; parentheses were added where they should have been.

I also changed the formatting so that subsequent lines align better with the first one (when tab spaces = 4 which seems to be quite common value).
Before change (white noise + 3 sine waves at 100 Hz, 1 kHz and 10 kHz at different amplitudes):
eq_before
After change (improvement mainly noticeable for the 100 Hz signal):
ex_fixed

### DIFF
--- a/plugins/Eq/EqSpectrumView.cpp
+++ b/plugins/Eq/EqSpectrumView.cpp
@@ -47,9 +47,9 @@ EqAnalyser::EqAnalyser() :
 
 	for (int i = 0; i < FFT_BUFFER_SIZE; i++)
 	{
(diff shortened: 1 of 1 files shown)


PhysSong:
@he29-net Do you mind if I merge this pull request with some additional formatting changes?

he29-net:
@PhysSong No problem. :)

PR #4902, opened 2019-03-17
Make Model class visitable

JohannesLorenz:
This makes the different `AutomatableModel`s [visitable](https://en.wikipedia.org/wiki/Visitor_pattern) and adds a visitor based cast function for cleaner code and more type safety. Currently only used in the Lv2 branch, but it's a generic feature.

### DIFF
--- a/include/AutomatableModel.h
+++ b/include/AutomatableModel.h
@@ -33,6 +33,7 @@
 #include "MidiTime.h"
 #include "ValueBuffer.h"
 #include "MemoryManager.h"
+#include "ModelVisitor.h"
 
 // simple way to map a property of a view to a model
 #define mapPropertyFromModelPtr(type,getfunc,setfunc,modelname)	\
@@ -68,6 +69,7 @@ class LMMS_EXPORT AutomatableModel : public Model, public JournallingObject
 	Q_OBJECT
 	MM_OPERATORS
 public:
+
 	typedef QVector<AutomatableModel *> AutoModelVector;
 
 	enum ScaleType
@@ -80,6 +82,32 @@ public:
 
 	virtual ~AutomatableModel();
 
+	// Implement those by using the MODEL_IS_VISITABLE macro
+	virtual void accept(ModelVisitor& v) = 0;
+	virtual void accept(ConstModelVisitor& v) const = 0;
+	// use this to make subclasses visitable
+#define MODEL_IS_VISITABLE \
+	void accept(ModelVisitor& v) override { v.visit(*this); } \
+	void accept(ConstModelVisitor& v) const override { v.visit(*this); }
+
+public:
+	//! Return this class casted to Target, or nullptr if impossible
+	template<class Target>
+	Target* dcast(bool doThrow = false)
+	{
+		DCastVisitor<Target> vis; accept(vis);
+		if(doThrow && !vis.result) Q_ASSERT(false);
+		return vis.result;
+	}
+
+	//! Return this class casted to const Target, or nullptr if impossible
+	template<class Target>
+	const Target* dcast(bool doThrow = false) const
+	{
+		ConstDCastVisitor<Target> vis; accept(vis);
+		if(doThrow && !vis.result) Q_ASSERT(false);
+		return vis.result;
+	}
 
 	bool isAutomated() const;
 	bool isAutomatedOrControlled() const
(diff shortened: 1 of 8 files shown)


JohannesLorenz:
This has no impact, so I'll merge without review in 5 days.

include/AutomatableModel.h
```diff
@@ -80,6 +82,32 @@ class LMMS_EXPORT AutomatableModel : public Model, public JournallingObject
...
+	virtual void accept(ConstModelVisitor& v) const = 0;
+	// use this to make subclasses visitable
+#define MODEL_IS_VISITABLE \
+	void accept(ModelVisitor& v) override { v.visit(*this); } \
+	void accept(ConstModelVisitor& v) const override { v.visit(*this); }
+
+public:
+	//! Return this class casted to Target, or nullptr if impossible
+	template<class Target>
+	Target* dcast(bool doThrow = false)
```
PhysSong: Could we have a more descriptive function name?

include/AutomatableModel.h
```diff
@@ -80,6 +82,32 @@ class LMMS_EXPORT AutomatableModel : public Model, public JournallingObject
...
+	{
+		DCastVisitor<Target> vis; accept(vis);
+		if(doThrow && !vis.result) Q_ASSERT(false);
```
PhysSong: Could you provide an example where the cast fails?
Also, please use `if (` instead of `if(`.

src/core/ModelVisitor.cpp
```diff
@@ -0,0 +1,28 @@
...
+ * License along with this program (see COPYING); if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA 02110-1301 USA.
+ *
+ */
+
+#include "ModelVisitor.h"
+
+ModelVisitor::~ModelVisitor() {}
+ConstModelVisitor::~ConstModelVisitor() {}
```
PhysSong: If those constructors do nothing, you can use default constructors and possibly remove this file. However, please feel free to ignore this if you have any other intentions to keep this file.

include/AutomatableModel.h
```diff
@@ -80,6 +82,32 @@ class LMMS_EXPORT AutomatableModel : public Model, public JournallingObject
 
 	virtual ~AutomatableModel();
 
+	// Implement those by using the MODEL_IS_VISITABLE macro
+	virtual void accept(ModelVisitor& v) = 0;
+	virtual void accept(ConstModelVisitor& v) const = 0;
+	// use this to make subclasses visitable
+#define MODEL_IS_VISITABLE \
+	void accept(ModelVisitor& v) override { v.visit(*this); } \
+	void accept(ConstModelVisitor& v) const override { v.visit(*this); }
```
PhysSong: Why did you add the macro definition into the class? I think you may pull out the definition like `mapPropertyFromModel`.

include/AutomatableModel.h
JohannesLorenz: It could be renamed to `dynamic_cast`, I think that explains it best. Any other ideas/preferences?

include/AutomatableModel.h
JohannesLorenz: For example, you have an `AutomatableModel*` pointing to a `ComboBoxModel` or `IntModel`, but try converting the pointer to `FloatModel*` (using `FloatModel` for the `Target` parameter). Should I add this to the comment?

Notes to self: Document `doThrow`, and correct the current statement that `nullptr` is only returned if `!doThrow`. And add something like `see non const version`.

src/core/ModelVisitor.cpp
JohannesLorenz: I'd normally do so, but if I write `~ModelVisitor() {}` into the header, clang's `-Wweak-vtables` warns me that the vtable will be placed into every translation unit unless it has at least one function in a cpp file. For `= default;`, the warning vanished, but this seems to be an error of the clang switch.

include/AutomatableModel.h
JohannesLorenz: I wanted to make them look like class members, but now I think it looks better if they are next to `mapPropertyFromModel` .

include/AutomatableModel.h
PhysSong: `dynamic_cast` itself is a C++ keyword. We can use something like `dynamicCast` if desired. Note that there's also `ModelView::castModel` which uses `dynamic_cast`.

JohannesLorenz:
I just added review commits. The old `dcast` could only cast to exactly the same class. The new `dcast` will cast to exactly the same class or any parent class.

There's also a test now documenting how it works.

PhysSong:
I'm not sure if it's worth handling this, but `InlineAutomation` inherits `FloatModel` as well. There are also `SfxrZeroToOneFloatModel` and `SfxrNegPosOneFloatModel`, but I don't think we should take care of them.

JohannesLorenz:
I don't think it's necessary now. Currently, I only need this to handle widgets of different types, like knobs, LEDs, ComboBoxes, for dialogs with LADSPA or Lv2 ports. `InlineAutomation` seems to only be used for note detuning, so I guess this can not be used for any of those widgets? The sfxr models can't be used in `ModelVisitor.cpp`, because that file can not see classes defined in plugins.

Issue #3640, opened 2017-06-14
Locked Note with VeSTige While Changing Tempo


BiRD4:
When I play a note while changing the tempo of a song, the note I played doesn't end until I have played it again (or stopped the song if it's in the song editor).

Using Windows 8.1 64-bit and LMMS version 1.2.0-rc2. I used Helm (64-bit) in VeSTige.

musikBear:
Specific for one VST, or is this also the case with those that are tested?

BiRD4:
The only VST that I currently have is Helm, so someone else would have to test to be sure, but it appears to be an LMMS issue. (https://github.com/mtytel/helm/issues/138)

DomClark:
It appears that the issue lies in `NotePlayHandle::resize`.
https://github.com/LMMS/lmms/blob/f79c2929a5b9baa81279d016341a01bf160eaa3c/src/core/NotePlayHandle.cpp#L529-L540
The product `m_origFrames * m_origTempo` is a signed 32-bit integer multiplication, which is liable to overflow. This always occurs when playing with a MIDI controller or the instrument window keyboard, since `m_origFrames` is set to `typeInfo<f_cnt_t>::max() / 2` (2^30-1) and `m_origTempo` has a minimum value of 10. While it's unlikely that this would also occur during normal song playback, it's far from impossible: at the maximum tempo of 999bpm, and a sample rate of 44100Hz, a 100-second note will overflow here.

Using doubles for the multiplication will fix the latter case, but not the former; the conversion back to `f_cnt_t` can overflow for sufficiently large tempo changes, which leads to undefined behaviour. (In practise, GCC seems to use the `cvttsd2si` instruction to convert back, which is documented to produce the value 0x80000000=-2^31 in case of overflow.) To fully address this, I would suggest we also either bound the resulting frame count, or have a special case to do nothing if `m_origin == OriginMidiInput`.

The observed issue is caused by the resulting incorrect values of `m_totalFramesPlayed` and `m_frames`, which cause the following condition in `NotePlayHandle::play` to be triggered erroneously:
https://github.com/LMMS/lmms/blob/f79c2929a5b9baa81279d016341a01bf160eaa3c/src/core/NotePlayHandle.cpp#L222-L229
[...]

PR #4908, opened 2019-03-22
Fix notes getting stuck under high CPU conditions

### LINKED ISSUE #4826: Stuck VST notes with high CPU load
When the CPU usage is too high, notes are skipped to save processing time. This results in the `NotePlayHandle`'s destructor being called in the same buffer as its constructor. The constructor sends a note-on event with the offset set to the offset of the note within the buffer, but the destructor sends a note-off with zero offset. The note-off now comes before the note-on chronologically, so the note gets stuck on. This may also affect some external MIDI devices.

DomClark:
Fixes #4826. Also happens to fix a bug where notes get stuck for MIDI-based instruments if the arpeggiator is turned on while the note is playing.

This PR changes `NotePlayHandle`s so they send a note-on message when they are first played, rather than when they are constructed. This means that `NotePlayHandle`s that are immediately discarded due to high CPU usage no longer send note-on messages to their plugins.

### DIFF
--- a/include/NotePlayHandle.h
+++ b/include/NotePlayHandle.h
@@ -302,6 +302,7 @@ private:
 	NotePlayHandleList m_subNotes;			// used for chords and arpeggios
 	volatile bool m_released;				// indicates whether note is released
 	bool m_releaseStarted;
+	bool m_hasMidiNote;
 	bool m_hasParent;						// indicates whether note has parent
 	NotePlayHandle * m_parent;			// parent note
 	bool m_hadChildren;
--- a/src/core/NotePlayHandle.cpp
+++ b/src/core/NotePlayHandle.cpp
@@ -199,7 +199,7 @@ void NotePlayHandle::play( sampleFrame * _working_buffer )
 
 	lock();
 
-	if( m_totalFramesPlayed == 0
+	if( m_totalFramesPlayed == 0 && !m_hasMidiNote
 		&& ( hasParent() || ! m_instrumentTrack->isArpeggioEnabled() ) )
 	{
 		m_hasMidiNote = true;
--- a/src/core/NotePlayHandle.cpp
+++ b/src/core/NotePlayHandle.cpp
@@ -63,6 +63,7 @@ NotePlayHandle::NotePlayHandle( InstrumentTrack* instrumentTrack,
 	m_subNotes(),
 	m_released( false ),
 	m_releaseStarted( false ),
+	m_hasMidiNote( false ),
 	m_hasParent( parent != NULL  ),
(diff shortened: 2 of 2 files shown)


DomClark:
CI failure appears to be unrelated to these changes.

karmux:
Fantastic! I was suffering from notes getting stuck a lot and with these code changes they don't anymore.

DomClark:
Don't merge this yet; it seems like it might have some issues with tempo automation.

qnebra:
If this was what I thinking then this was really amazing fix for me. How many notes are lose due to high CPU usage.

DomClark:
Tempo automation with this pull request can sometimes create stuck notes; this is due to `m_totalFramesPlayed` being zero for two consecutive buffers, resulting in an extra note-on message. This happens often when playing using the keyboard or a MIDI controller, due to the overflow issues mentioned in https://github.com/LMMS/lmms/issues/3640#issuecomment-475906581, and can also occur occasionally during normal playback when `m_totalFramesPlayed` is changed as the `NotePlayHandle` is resized. @PhysSong suggests in https://github.com/LMMS/lmms/issues/3775#issuecomment-327450110 that `m_totalFramesPlayed` should be left alone during `NotePlayHandle` resizing; until that is implemented, I'm just going to add an additional check that the note-on message has not yet been sent.

DomClark:
This should work properly now; it can be reviewed and then hopefully merged.

PhysSong:
> This means that `NotePlayHandle`s that are immediately discarded due to high CPU usage no longer send note-on messages to their plugins.

Does this PR affect MIDI output ports connected to the instrument? Isn't `InstrumentTrack::processOutEvent` supposed to handle MIDI output by calling `m_midiPort.processOutEvent()`?

DomClark:
> Does this PR affect MIDI output ports connected to the instrument?

Yes, it does. However, prior to this change, MIDI output ports would be given a note-on message followed immediately by a note-off message in the same MIDI tick, resulting in either a very short note or no note at all. After this change, they will simply not receive any message, so I don't think this is a serious regression.

The issue here is that MIDI output ports and MIDI-based instruments both receive the same MIDI events, so it's not possible to have the MIDI output port receive a note but the instrument not. Also, MIDI notes are driven by play handles, so these would need to be kept around rather than be discarded in high-CPU conditions. Perhaps in the future we should separate external MIDI output from internal MIDI control and CPU-saving note skipping, but I feel this is out of scope here.

PhysSong:
Then I'd say let's merge this. I can't see any other issues with these changes. :rocket:

Issue #4789, opened 2019-01-22
SIGSEGV on cloning pattern


StinaIsNormallyDistributed:
When ctrl-clicking to drag-clone any pattern (beat, pianoroll or automation), lmms segfaults without further error.
```
lmms 1.2.0-rc7.42
qt5 5.12.0-3
```

Cangle50:
Looks a little like what was reported in #4738 .

@megahertz commented
> Fixed in Arch Linux in qt5-base 5.12.0-2

Looks like qt5 5.12.0-3 has broken it again.

dvzrv:
@StinaIsNormallyDistributed seems this is happening on exit:

```
QPixmap::grabWidget is deprecated, use QWidget::grab() instead
[1]    1168 segmentation fault (core dumped)  lmms
```

dvzrv:
Looking at the [QPixmap documentation](http://doc.qt.io/qt-5/qpixmap-obsolete.html), this is obsolete since 5.12 and needs to be changed within lmms in [these files](https://github.com/LMMS/lmms/search?q=grabwidget&unscoped_q=grabwidget).
However, this should not lead to a crash (at least not intentionally).
FWIW, it's the same problem with 5.12.0-2.

dvzrv:
This seems to trickle down to Qcursor:

```
Thread 1 "lmms" received signal SIGSEGV, Segmentation fault.
QCursor::shape (this=this@entry=0x0) at kernel/qcursor.cpp:539
539     kernel/qcursor.cpp: No such file or directory.
```

Don't have much time for a debug build, but might do that in the coming days

CYBERDEViLNL:
While dragging a `leaveEvent` in `TrackContentObjectView` is triggered that calls `QApplication::restoreOverrideCursor()`, that is what is causing the crash.

It won't crash with this patch:

include/Track.h
```
--- a/include/Track.h
+++ b/include/Track.h
@@ -311,6 +311,7 @@
 	bool m_gradient;
 
  	bool m_needsUpdate;
+ 	bool m_isDragging;
 	inline void setInitialMousePos( QPoint pos )
 	{
 		m_initialMousePos = pos;
```


src/core/Track.cpp
```
--- a/src/core/Track.cpp
+++ b/src/core/Track.cpp
@@ -275,7 +275,8 @@
 	m_textShadowColor( 0, 0, 0 ),
 	m_BBPatternBackground( 0, 0, 0 ),
 	m_gradient( true ),
[...]

PR #4918, opened 2019-03-27
Fix `SIGSEGV on cloning pattern` https://github.com/LMMS/lmms/issues/4789

CYBERDEViLNL:
TrackContentObjectView now sets the cursor locally instead of global.

Fixes https://github.com/LMMS/lmms/issues/4789

### DIFF
--- a/src/core/Track.cpp
+++ b/src/core/Track.cpp
@@ -570,9 +570,9 @@ void TrackContentObjectView::dropEvent( QDropEvent * de )
  */
 void TrackContentObjectView::leaveEvent( QEvent * e )
 {
-	while( QApplication::overrideCursor() != NULL )
+	if( cursor().shape() != Qt::BitmapCursor )
 	{
-		QApplication::restoreOverrideCursor();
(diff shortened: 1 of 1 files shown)


PhysSong:
First of all, sorry for the late response.
I can't reproduce the original issues, maybe due to different environments. However, the code looks good to me.
I'm not sure how to handle remaining usages of global cursors since they seem not to cause severe issues.

CYBERDEViLNL:
Ok np :-)

The issue only occurs in combination with Qt version 5.12 and bigger (which some GNU/Linux distro’s like Arch use by default)

I think it is a good idea to always use local cursor methods where possible. May I suggest to merge this PR if there are no objections of any kind and address other files with global cursor methods at a later stage since they don’t cause issues (yet)?

zonkmachine:
> I'm not sure how to handle remaining usages of global cursors since they seem not to cause severe issues.

Then I think we should merge this if it's good to go and just leave the rest for later.

Issue #2121, opened 2015-06-20
Refreshing the File Browser should not collapse all folders


Wallacoloo:
Currently, clicking the refresh icon in the lower right of a file browser in the left pane of LMMS will cause all of its folders to be collapsed.

Ideal behavior, in my mind, is to preserve the collapsed/expanded state of all folders that existed before the refresh and have only the newly created folders default to being collapsed. Similarly, if a file/folder was selected before the refresh, it should also be selected after the refresh, if it still exists, though that's a minor difference.

musikBear:
I have already once made a topic around this, but it was dismissed. I have to chime in on this one, because the current behavior is a real anointment (imo)
Considder why the *refresh* is used:
The user open a folder, where he expect to find a particular preset
User enters s search-token
The search gives no usable result
The reason must be that the token was wrong OR the folder was wrong
so:
1 Token is wrong
If the user now press *refresh* the token remains, but the browser is empty -eg no further search possible
Current behavior is useless

2 Folder is wrong
If the user now press *refresh*  the browser is empty. There is no folder-tree to access -eg no further search possible
Current behavior is useless -again

In which circumstance is the *current behavior* useful /make sense?

tresf:
As part of a pruning effort, this `enhancement` request is archived into a dedicated "Better Workflow" checklist here #4877.

PR #4930, opened 2019-04-03
FileBrowser: Backup expanded directory states.

CYBERDEViLNL:
Partially fixes https://github.com/LMMS/lmms/issues/2121 (it doesn't store the currently selected item).

### DIFF
--- a/include/FileBrowser.h
+++ b/include/FileBrowser.h
@@ -55,7 +55,7 @@ public:
 
 private slots:
 	void reloadTree( void );
-	void expandItems( QTreeWidgetItem * item=NULL );
+	void expandItems( QTreeWidgetItem * item=NULL, QList<QString> expandedDirs = QList<QString>() );
 	// call with item=NULL to filter the entire tree
 	bool filterItems( const QString & filter, QTreeWidgetItem * item=NULL );
 	void giveFocusToFilter();
@@ -86,6 +86,10 @@ class FileBrowserTreeWidget : public QTreeWidget
 public:
 	FileBrowserTreeWidget( QWidget * parent );
 	virtual ~FileBrowserTreeWidget() = default;
+
+	//! This method returns a QList with paths (QString's) of all directories
+	//! that are expanded in the tree.
+	QList<QString> expandedDirs( QTreeWidgetItem * item = nullptr ) const;
 
 
 protected:
--- a/src/gui/FileBrowser.cpp
+++ b/src/gui/FileBrowser.cpp
@@ -163,6 +163,7 @@ bool FileBrowser::filterItems( const QString & filter, QTreeWidgetItem * item )
 
 void FileBrowser::reloadTree( void )
 {
+	QList<QString> expandedDirs = m_fileBrowserTreeWidget->expandedDirs();
 	const QString text = m_filterEdit->text();
 	m_filterEdit->clear();
 	m_fileBrowserTreeWidget->clear();
@@ -171,17 +172,17 @@ void FileBrowser::reloadTree( void )
 	{
 		addItems( *it );
 	}
-	expandItems();
+	expandItems(NULL, expandedDirs);
 	m_filterEdit->setText( text );
 	filterItems( text );
 }
 
 
 
-void FileBrowser::expandItems( QTreeWidgetItem * item )
+void FileBrowser::expandItems( QTreeWidgetItem * item, QList<QString> expandedDirs )
 {
 	int numChildren = item ? item->childCount() : m_fileBrowserTreeWidget->topLevelItemCount();
-	for( int i = 0; i < numChildren; ++i )
+	for (int i = 0; i < numChildren; ++i)
 	{
 		QTreeWidgetItem * it = item ? item->child( i ) : m_fileBrowserTreeWidget->topLevelItem(i);
 		if ( m_recurse )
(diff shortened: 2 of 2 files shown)


JohannesLorenz:
@CYBERDEViLNL Is this ready for review, or do you plan to add more commits?

CYBERDEViLNL:
@JohannesLorenz it now is ready for review.

src/gui/FileBrowser.cpp
```diff
@@ -326,6 +328,49 @@ FileBrowserTreeWidget::FileBrowserTreeWidget(QWidget * parent ) :
...
+		// Add expanded child directories.
+		expandedDirs.append(
+			expandedChildDirs( item )
+		);
+
+	}
+	return expandedDirs;
+}
+
+QList<QString> FileBrowserTreeWidget::expandedChildDirs( QTreeWidgetItem * item ) const
```
JohannesLorenz: While both functions look correct to me, you could turn both into one, like in `FileBrowser::expandItems` with the `?:`-operator. However, this is just a suggestion.

JohannesLorenz:
Just did some testing. There's an old bug that the arrow left to a folder vanishes on reload. The bug is old: Open a folder, open a sub-folder, reopen the top folder, the arrow of the sub-folder is vanished.

It's an old bug, you don't need to fix it, unless you want.

src/gui/FileBrowser.cpp
CYBERDEViLNL: Yes I like that :-D

src/gui/FileBrowser.cpp
```diff
@@ -326,6 +328,49 @@ FileBrowserTreeWidget::FileBrowserTreeWidget(QWidget * parent ) :
...
+{
+	QList<QString> expandedDirs;
+	for( int i = 0; i < topLevelItemCount(); ++i )
+	{
+		QTreeWidgetItem * item  = topLevelItem( i );
+
+		// Add expanded top level directories.
+		if (item->isExpanded() && (item->type() == TypeDirectoryItem))
+		{
+			Directory *d = static_cast<Directory *> ( item );
```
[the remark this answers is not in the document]
CYBERDEViLNL: Yes it is intended. Since the check `if (item->isExpanded() && (item->type() == TypeDirectoryItem))` we can be very certain that it’s type is `Directory`, so I thought no need to do a `dynamic_cast`. Please do correct me if I’m wrong.

src/gui/FileBrowser.cpp
```diff
@@ -326,6 +328,49 @@ FileBrowserTreeWidget::FileBrowserTreeWidget(QWidget * parent ) :
...
+	QList<QString> expandedDirs;
+	for (int i = 0; i < item->childCount(); ++i)
+	{
+		QTreeWidgetItem * childItem  = item->child( i );
+		if (childItem->isExpanded() && (childItem->type() == TypeDirectoryItem))
+		{
+			Directory *d = static_cast<Directory *> ( childItem );
+			expandedDirs.append( d->fullName() );
+
+			// Recurse.
```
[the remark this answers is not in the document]
CYBERDEViLNL: Haha yeah I had the same thoughts when working on it :-P

CYBERDEViLNL:
@JohannesLorenz 

> There's an old bug that t..

Hmm I can't seem to reproduce, here it only vanishes if the directory is empty. (tested with stable and master)

> I'm finished with review, both functional + style. No serious issues. Feel free to fix what you like.

Thank you! I've made some changes :-)

src/gui/FileBrowser.cpp
```diff
@@ -326,6 +328,32 @@ FileBrowserTreeWidget::FileBrowserTreeWidget(QWidget * parent ) :
 
 }
 
+// This method returns a QList with paths (QString's) of all directories
```
JohannesLorenz: I'd add those two lines into the header (above the function). If you use `//!` there instead of `//`, it shows up in our doxygen documentation.

include/FileBrowser.h
```diff
@@ -87,6 +87,8 @@ class FileBrowserTreeWidget : public QTreeWidget
 	FileBrowserTreeWidget( QWidget * parent );
 	virtual ~FileBrowserTreeWidget() = default;
 
+	QList<QString> expandedDirs( QTreeWidgetItem * item=NULL ) const;
```
JohannesLorenz: Coding conventions: "Infix operators ('=', '+', '-', '*', '/', etc.) should have a space before and after them"
```suggestion
	QList<QString> expandedDirs( QTreeWidgetItem * item = nullptr ) const;
```

CYBERDEViLNL:
> I added two more comments, both voluntary. Tests were OK, so I think there's no further review required.

Thank you I like the tips :-)

JohannesLorenz:
Great work! Do you want to "squash and merge"?

CYBERDEViLNL:
Yeah that will be great!

JohannesLorenz:
Just as a hint: You can use the "squash and merge" button, no need to squash manually and then force-push. But it's not wrong to do so :smile:

CYBERDEViLNL:
Good hint, except I don't have commit access and thus no button here :yum:

JohannesLorenz:
Oh, np! Does this mean you can't merge in general? If that's true, I'll merge it for you.

CYBERDEViLNL:
Yes that's `true`.

Issue #874, opened 2014-06-20
Finer Scaling In The Sequencer


HDDigitizerMusic:
Can we add a drop down list in the sequencer tools area like in the piano roll where it allows finer scaling?

Here's an example:
![example](https://cloud.githubusercontent.com/assets/6778944/3345081/9dd92662-f8b3-11e3-994e-2974b69be624.png)

Of course, it would not have that many scales but I think you get the idea.

Here's what it looks like in another DAW:
![fl-studio-automation-460-85](https://cloud.githubusercontent.com/assets/6778944/3344960/a395a924-f8b1-11e3-8cb1-2f9e478ef6f4.jpg)

See how there are four large sections like in LMMS but in each section you will find four more measures and inside those two? I’ve seen this feature in a few other DAW's and think it is one of high importance.

I know we have smooth scaling, but a finer measure grid would help get an exact placement. Any thoughts?

By the way, I have to say, LMMS is getting pretty good. I was going to get FL Studio but after I found the changes going on in LMMS, I don’t think I’ll want or need to.

Spekular:
+1 for this, almost made a new issue for it before I found this one. I don't know that I find diizy's solution/workaround preferable to an actual quantization selector though, at least not as a replacement.

An additional feature I think would be neat is an "auto" mode. At 100% zoom it would use the default quantization of one bar, at 200% zoom it would snap to half bars, 400% zoom = beats, and so on. Optionally with a multiplier dropdown next to it, so people can define what Q they'd like at 100% zoom. For example Q set to auto with multiplier set to 2x => half bar snapping at 100% zoom, beat snapping at 200%, ...

Spekular:
Better UX for the auto proposal:
One dropdown selecting Q level
One checkbox enabling/disabling Auto/Adaptive Q

Then Auto/Adaptive Q uses the selected Q level at 100% zoom and adapts at other zooms.

PhysSong:
Implemented via #4973.

PR #4933, opened 2019-04-05
Allow sample track TCOs to resize smaller than one bar

Spekular:
Currently, TCOs in sample tracks can't be resized to less than one bar, which is frustrating if short samples are loaded (they'll have useless empty space to the right). This PR allows sample tracks TCOs to be resized to less than one bar's size. Empty TCOs will still be created with a size of one bar.

At 100% zoom, it becomes difficult to see the TCO if it is at minimum size, but this already occurs with BB Track TCOs, so I wouldn't consider it a regression.


Spekular:
Looks like some weird unrelated stuff ended up here, will fix.

src/tracks/SampleTrack.cpp
```diff
@@ -147,8 +144,19 @@ void SampleTCO::setSampleBuffer( SampleBuffer* sb )
...
-	m_sampleBuffer->setAudioFile( _sf );
-	changeLength( (int) ( m_sampleBuffer->frames() / Engine::framesPerTick() ) );
+	int length;
+	if (_sf == "") //When creating an empty sample pattern make it a bar long
+	{
+		float nom = Engine::getSong()->getTimeSigModel().getNumerator();
+		float den = Engine::getSong()->getTimeSigModel().getDenominator();
+		length = DefaultTicksPerTact * ( nom / den );
+	}
+	else //Otherwise set it to the samples length
```
Spekular:
```suggestion
	else //Otherwise set it to the sample's length
```

BaraMGB:
Do you really want to change this for stable?

Spekular:
I think you could argue the original behavior is a bug. I did intentionally target it for stable but if there are objections I could retarget to master.

JohannesLorenz:
The freshly loaded samples (doubleclick on the sample track TCOs) load with a correct length, but when you resize them to 1, you can't resize them back. In this situation, even when you load the same sample file again, the TCO size is still 1. Both can cause confusion and may be worth a fix.

As for the stable-1.2 question: I'm fine with merging it into stable-1.2 after the review will have approved this branch.

Spekular:
Hmm, I definitely tried scaling them up from zero and it seemed to work fine for me. How exactly do I reproduce the issue? Does the drag handle show up? Does it work when zoomed in further?

JohannesLorenz:
1. Click on the sample track to create a yellow, empty TCO
2. Double click on that TCO, open a file
3. Move mouse on border of TCO, resize dialog appears
4. Resize to 1
5. Try to resize to <1 at any zoom, drag handle will shop up, but you can't make it smaller than 1

Spekular:
@JohannesLorenz can't reproduce with those steps. Are you holding ctrl to fine adjust? Since it resizes in bar increments by default, 1 bar is the smallest it can go without fine adjusting.

The clip scales to the correct size when a sample is loaded for me too. However, if you attempt to reload the same sample, this doesn't happen. I think the call to setSampleFile might be skipped when loading the existing sample again? I'll look into that.

Spekular:
Reformatted for brevity, but here's what I think the cause is. Will fix.
```
if( af != "" && af != m_tco->m_sampleBuffer->audioFile() )
	{ m_tco->setSampleFile( af ); }
```

Spekular:
The sample track clip should now resize to the correct length no matter what sample is loaded (a new one or the same that it already contained).

Spekular:
There's currently a bug with cloning. A short (less than one bar) sample clip cannot be cloned to the right of itself unless another sample clip exists between it and the clone's destination. In other words, a short sample at position A can be cloned to position C if B exists, but B cannot be cloned to position C.

A B [c]

Spekular:
A TrackContentWidget::dragEnterEvent is fired when a sample clip is Ctrl+dragged over or off of another sample clip. For a clip of 1 bar or more, we exit the clip's "airspace" directly into another bar than the original clip. For a short clip, we end up in the same bar.

tcoPos is apparently only accurate to one bar, so it looks like we're pasting to the same position when pasting a short clip to the right of itself. If the clip is dragged to another track and back to the original, or on top of anotoer clip in the same track and back, tcoPos is forced to update again, which can allow short clips to be cloned to their right.

Spekular:
Essentially short clips often evaluate true on ``tcoPos == grabbedTCOTact`` because tcoPos is only accurate to one bar and is only updated on some events. This returns false in canPasteSelection.

Spekular:
Previous issue is fixed.

Allowing short sample clips exposes an edge case. Cloning a sample clip to a bar containing a short clip can completely cover the short clip. Whether or not a safeguard should be implemented for this is up for debate.

JohannesLorenz:
@Spekular I did a few tests. While everything worked fine it's a bit sad that you can't snap two short blocks together. Though I don't see that much use, as the sample TCO lengths/positions can't be set to an exact value anyways. I'd say this may be a nice feature, but it's absolutely not stopping this PR. If you want, you can try to add this to this PR, but I see it as totally optional.

Spekular:
@johanneslorenz I'd rather leave it as is and then take a crack at #874 once this is merged.

Spekular:
@JohannesLorenz fixed everything except the tick_t comment. I wouldn't know how to implement that one, the scope seems a bit big.

src/tracks/SampleTrack.cpp
```diff
@@ -147,8 +144,19 @@ void SampleTCO::setSampleBuffer( SampleBuffer* sb )
...
+	if (_sf == "")
+	{	//When creating an empty sample pattern make it a bar long
+		float nom = Engine::getSong()->getTimeSigModel().getNumerator();
+		float den = Engine::getSong()->getTimeSigModel().getDenominator();
+		length = DefaultTicksPerTact * ( nom / den );
+	}
+	else
+	{	//Otherwise set it to the sample's length
+		m_sampleBuffer->setAudioFile( _sf );
+		length = sampleLength();
```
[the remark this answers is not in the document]
Spekular: :)

JohannesLorenz:
I'm OK with not fixing the `tick_t`. Can you please [re-request review](https://github.blog/changelog/2019-02-21-re-request-review-on-a-pull-request/), so I can mark it as OK? I'm curious if that works :smile:

Note, before merging, maybe it's good to squash those 10 commits.

src/tracks/SampleTrack.cpp
```diff
@@ -147,8 +144,19 @@ void SampleTCO::setSampleBuffer( SampleBuffer* sb )
 
 void SampleTCO::setSampleFile( const QString & _sf )
 {
-	m_sampleBuffer->setAudioFile( _sf );
-	changeLength( (int) ( m_sampleBuffer->frames() / Engine::framesPerTick() ) );
+	int length;
```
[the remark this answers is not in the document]
Spekular: Skipping this as scope seems big.

Spekular:
I on't have that option, perhaps because 

> Pull request authors can't request reviews unless they are either a repository owner or collaborator with write access to the repository.

I've done my best to squash as well, no idea if it worked.

JohannesLorenz:
OK, It's accepted now, anyways. I think you did not squash it correctly: You squashed (a091d50) , but then merged (062041f) your unsquashed branch on top? I think you can solve it all by one simple `git rebase -i 97d5529c18bee51d638607a192ab91b131d41086`, marking all `s`, and no further merge.

Spekular:
When I try that, I immediately get a merge conflict :)

Spekular:
I ran into a brand new error on this squash attempt, aside from merge conflicts some commits couldn't be applied because ???, and I couldn't really tell how to proceed. Maybe whoever merges can do a squash and merge?

PhysSong:
@JohannesLorenz Using the `Squash and merge` feature will be perfectly fine in this case. You may want to edit the commit message to remove redundant messages though.

src/tracks/SampleTrack.cpp
```diff
@@ -440,8 +448,15 @@ void SampleTCOView::mouseReleaseEvent(QMouseEvent *_me)
 void SampleTCOView::mouseDoubleClickEvent( QMouseEvent * )
 {
 	QString af = m_tco->m_sampleBuffer->openAudioFile();
-	if( af != "" && af != m_tco->m_sampleBuffer->audioFile() )
-	{
+
+	if ( af.isEmpty() ) {} //Don't do anything if no file is loaded
+	else if ( af == m_tco->m_sampleBuffer->audioFile() )
+	{	//Instead of reloading the existing file, just reset the size
```
DomClark: If a user modifies one of their audio files (not entirely unlikely; could be new vocals, updated bounced track, new export from another DAW, etc.), opening the file dialog and re-selecting the sample will no longer load the modified file. IMO this is unintuitive.

Spekular: I didn't introduce this behavior. Existing version skips reloading entirely, my patch makes it skip reloading but at least reset size

Should i remove the check entirely?

DomClark: Ah, indeed. I didn't read the original carefully enough, just saw the comment added and assumed it was new. In that case, I think the condition ought to go, but I don't know whether it's in scope here. Sorry.

Spekular: I'd rather leave as is then since this targets stable.

Issue #4927, opened 2019-04-02
Preset saving system bug


RoxasKH:
The title really doesn't explain it very well. What i want to make devs notice, is that if you save a preset of an instrument that has one of its knobs linked to a controller in the project you saved it, if you then try to load it in a new project that has no controllers, just clicking on it in the sidebar will make LMMS freeze or crash and close itself.

I think it should be more clever if when you save a preset this way it autoremoves and don't save the information about the connection to a controller.
Or a better way would be to make the preset create a new controller with the same characteristics in the project where you load it and autoconnect the knob to it.

I'm on Windows 10 64-bit and i'm using LMMS 1.2.0 RC8

Here i upload an example, just put it in you presets folder and try to load it from the sidebar
[Glitchy.zip](https://github.com/LMMS/lmms/files/3034989/Glitchy.zip)

Thanks

CYBERDEViLNL:
Confirmed. Automation track connections are also stored in the preset, they are unused but don't seem to crash LMMS.

noahb01:
I've fixed the crashing issue with pull request #4937.

SecondFlight:
Should this be closed via #4939? Or are there still more issues that need addressing?

noahb01:
My pr fixes the crash, but perhaps we should keep this issue open because we probably shouldn’t be saving controller connections in the first place, which my pr does nothing to fix.

On the other hand, this issue is directly about the crash, which has been fixed, so I’m in favor of closing it.

SecondFlight:
Yeah, I agree. This issue is describing a bug, and leaving it open to serve as an enhancement will likely cause confusion.

@noahb01 Feel free to create a separate enhancement request, since you likely know a lot more about what's going on. I'd agree that either 1) controller connections should not be saved or 2) controllers should be added when the preset is added.

PR #4937, opened 2019-04-11
ControllerConnection: Don't connect controllers that do not exist.

noahb01:
This fixes a crash when attempting to load a preset that was saved with controller connections into a project with no controllers (or fewer controllers than the preset's `m_controllerId`)

See issue #4927 for more info.

### DIFF
--- a/src/core/ControllerConnection.cpp
+++ b/src/core/ControllerConnection.cpp
@@ -211,10 +211,12 @@ void ControllerConnection::loadSettings( const QDomElement & _this )
 			m_controllerId = -1;
 		}
 
-		if (!Engine::getSong()->isLoadingProject() && m_controllerId != -1)
+		if (!Engine::getSong()->isLoadingProject()
+			&& m_controllerId != -1
+			&& m_controllerId < Engine::getSong()->controllers().size())
 		{
 			setController( Engine::getSong()->
-					controllers().at( m_controllerId ) );
+				controllers().at( m_controllerId ) );
 		}
 		else
 		{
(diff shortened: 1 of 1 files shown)


PhysSong:
I think this change should target to `stable-1.2` as it's a simple bug fix.

noahb01:
Is there a reason we can’t apply it to both?

PhysSong:
Normally we apply fixes to `stable-1.2` first and then merge the branch into `master`.

noahb01:
Makes sense. Does this mean that b

noahb01:
Sorry didn’t mean to close this.

Anyway, can I simply change the target of this pr or do I have to remake it?

I would like to see the fix in master though, for me at least both branches are equally buggy and master has a lot of new features, so I use master for normal use.

SecondFlight:
You can't just change the target of the PR, as that will attempt to merge all of the new features in master to stable 1.2. The easiest way to target this for 1.2 without git wizardry is to clone 1.2 and re-apply your changes. Your PR looks small so this should be trivial to do.

It's my understanding that bug fixes are regularly put into master from 1.2, so it's only a matter of time before this fix shows up in master.

noahb01:
#4939 is this pr targeted against `stable-1.2`.

Should I close this or leave this open to allow it to be merged to master as well?

PhysSong:
> Should I close this or leave this open to allow it to be merged to master as well?

We'll merge `stable-1.2` into `master` to let `master` get fixes from `stable-1.2`.

PR #4938, opened 2019-04-11
Record chords

### LINKED ISSUE #4807: Record chords
Edited by @tresf 

In the Piano Roll editor, the "chords" drop-down does not affect notes played through a keyboard, only those placed with a mouse, also impacting recording.

image

---

Original

When the piano-roll *chord-selector* has any other setting than *no-chords*, using  *record* ought to insert the selected* chord* in piano-roll. For now only the pressed note is inserted.

alxdttn:
Checks for currently selected chord when recording from keyboard.

Closes #4807

### DIFF
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -3523,7 +3523,7 @@ void PianoRoll::finishRecordNote(const Note & n )
 						it->getPanning() );
 				n1.quantizeLength( quantization() );
 
-				//ADDED CODE
+				//Get selected chordModel
 				const InstrumentFunctionNoteStacking::Chord & chord = InstrumentFunctionNoteStacking::ChordTable::getInstance()
 						.getChordByName( m_chordModel.currentText() );
 
@@ -3538,7 +3538,6 @@ void PianoRoll::finishRecordNote(const Note & n )
 							m_pattern->addNote( new_note );
 						}
 					}
-					//END ADDED CODE
 
 				m_pattern->addNote( n1 );
 				update();
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -3523,21 +3523,20 @@ void PianoRoll::finishRecordNote(const Note & n )
 						it->getPanning() );
 				n1.quantizeLength( quantization() );
 
-				//Get selected chordModel
+				//Get selected chord
 				const InstrumentFunctionNoteStacking::Chord & chord = InstrumentFunctionNoteStacking::ChordTable::getInstance()
-						.getChordByName( m_chordModel.currentText() );
+					.getChordByName( m_chordModel.currentText() );
 
-					if( ! chord.isEmpty() )
+				if( !chord.isEmpty() )
+				{
+					for( int i = 1; i < chord.size(); i++ )
 					{
-						// if a chord is selected, create following notes in chord mode
-						for( int i = 1; i < chord.size(); i++ )
-						{
-							Note new_note( n.length(), it->pos(), it->key() + chord[i] );
-							new_note.setPanning( it->getPanning() );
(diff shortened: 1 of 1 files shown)


src/gui/editors/PianoRoll.cpp
```diff
@@ -3522,6 +3522,23 @@ void PianoRoll::finishRecordNote(const Note & n )
 						it->key(), it->getVolume(),
 						it->getPanning() );
 				n1.quantizeLength( quantization() );
+
+				//Get selected chordModel
+				const InstrumentFunctionNoteStacking::Chord & chord = InstrumentFunctionNoteStacking::ChordTable::getInstance()
+						.getChordByName( m_chordModel.currentText() );
+
+					if( ! chord.isEmpty() )
```
PhysSong: It'd be better to unindent the `if` block one level.

src/gui/editors/PianoRoll.cpp
```diff
@@ -3522,6 +3522,23 @@ void PianoRoll::finishRecordNote(const Note & n )
...
+					{
+						// if a chord is selected, create following notes in chord mode
+						for( int i = 1; i < chord.size(); i++ )
+						{
+							Note new_note( n.length(), it->pos(), it->key() + chord[i] );
```
PhysSong: We prefer `camelCase` to `snake_case`, though it's not in the coding convention.

src/gui/editors/PianoRoll.cpp
```diff
@@ -3522,6 +3522,23 @@ void PianoRoll::finishRecordNote(const Note & n )
 						it->key(), it->getVolume(),
 						it->getPanning() );
 				n1.quantizeLength( quantization() );
+
+				//Get selected chordModel
```
PhysSong: `// get selected chord` is enough, but you can leave this as-is.

src/gui/editors/PianoRoll.cpp
PhysSong: Also, `if( a )` is an old style. According to [our coding convention](https://github.com/LMMS/lmms/wiki/Coding-conventions), `if (a)` is correct. However, you can leave this as-is if the change breaks consistency.

src/gui/editors/PianoRoll.cpp
```diff
@@ -3522,6 +3522,23 @@ void PianoRoll::finishRecordNote(const Note & n )
...
+				const InstrumentFunctionNoteStacking::Chord & chord = InstrumentFunctionNoteStacking::ChordTable::getInstance()
+						.getChordByName( m_chordModel.currentText() );
+
+					if( ! chord.isEmpty() )
+					{
+						// if a chord is selected, create following notes in chord mode
+						for( int i = 1; i < chord.size(); i++ )
```
PhysSong: Here as well.

PhysSong:
Thank you for your contribution! I left some comments regarding coding styles.
Also, please consider creating a new branch when you create a pull request next time.

SecondFlight:
@PhysSong That's on me, they asked in #devtalk if they should create a new branch and I said they didn't need to.

zonkmachine:
Tested. This fixes the issue as reported in #4807. I'm a bit confused as to how the chord function is intended to work though. Even though the chord is recorded only the base note is played back while recording/playing.

alxdttn:
I thought that would be best, as it matches the behavior of using the mouse to place notes when in chord mode, which also seems to only play the bass note.

zonkmachine:
> I thought that would be best, as it matches the behavior of using the mouse to place notes when in chord mode, which also seems to only play the bass note.

Yes. Its enough for this PR. This is good to merge I think and I'll open a new issue for the _'only base note sounds while playing chord notes'_ issue.

zonkmachine:
@PhysSong @alxdttn Merge?

PR #4939, opened 2019-04-12
Don't try to connect to nonexistent controllers

noahb01:
This is the same as #4937 but for `stable-1.2` instead of `master`.

### DIFF
--- a/src/core/ControllerConnection.cpp
+++ b/src/core/ControllerConnection.cpp
@@ -211,9 +211,9 @@ void ControllerConnection::loadSettings( const QDomElement & _this )
 			m_controllerId = -1;
 		}
 
-		if (!Engine::getSong()->isLoadingProject()
+		if ( !Engine::getSong()->isLoadingProject()
 			&& m_controllerId != -1
-			&& m_controllerId < Engine::getSong()->controllers().size())
+			&& m_controllerId < Engine::getSong()->controllers().size() )
 		{
 			setController( Engine::getSong()->
 				controllers().at( m_controllerId ) );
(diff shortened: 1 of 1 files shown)


src/core/ControllerConnection.cpp
```diff
@@ -211,10 +211,12 @@ void ControllerConnection::loadSettings( const QDomElement & _this )
 			m_controllerId = -1;
 		}
 
-		if (!Engine::getSong()->isLoadingProject() && m_controllerId != -1)
+		if ( !Engine::getSong()->isLoadingProject()
+			&& m_controllerId != -1
+			&& m_controllerId < Engine::getSong()->controllers().size() )
```
PhysSong: I think you can move this check to the index validation a few lines above if you want.

noahb01: Moved. It does make more sense to have all the range-checks in one if() instead of spread between the two.

noahb01:
Is this good to merge?

PhysSong:
> Is this good to merge?

Sure. BTW, may I revert some redundant formatting changes before merging to make the history cleaner?

noahb01:
Yes, but those formatting changes were made to keep the formatting consistent across the file. I dont suppose it matters that much though if we are going to implement an automatic formatting tool anyway.

src/core/ControllerConnection.cpp
PhysSong: This check should have been there because the two `if` statements do different things.
Reported by @zonkmachine:

> With commit 91f9f1a from #4939 I get an issue with loading a project with a controller.
> To replicate:
>  1. In a new default project, connect an LFO to the 3osc pan knob in the song editor.
>  2. Save and reload.
> The project loads but the knob isn't moving. Command line output reads controller index invalid.

This was my mistake.

PhysSong: Fixed via 934ea6d0866818833caaacbe16c816b9a94c9b9a.

zonkmachine: @PhysSong Fix confirmed, thanks!

PR #4944, opened 2019-04-14
Fix gig and sf2 directories being unused

### LINKED ISSUE #4180: SF2 Directory Setting/Relative Paths
Spin-off from #3810... 

There are two named settings that pertain to soundfonts... 

1. There's a path called "SF2 DIRECTORY"
2. There's a path called "DEFAULT SOUNDFONT FILE"

screen shot 2018-02-19 at 9 50 51 pm

Unfortunately the software doesn't really make use of either of these when calculating relative paths, and I'm not sure we ever will, but I feel this is a bug.

Soundfonts were decided to be stored inside `samples/sf2`, discussed in https://github.com/LMMS/lmms/issues/1807 and implemented in https://github.com/LMMS/lmms/pull/1908 and then renamed to `samples/soundfonts` in https://github.com/LMMS/lmms/pull/3895.

What seems to be the missing puzzle piece here is that we added an option for a "SF2 DIRECTORY" folder, but we ignore it.  I think it simply needs to be removed.

I'm sure there will be a few people asking to have their SF2 directory outside their `samples` directory, but still relocatable, but it's simply not something we support.  If you can point me to a regression that occured between 1.1.0 and 1.2.0, I'll consider writing a patch for this, but at a glance it appears the bug was adding this location to the UI to begin with.

This is not to be confused with storing samples relative to project files, which is outlined in https://github.com/LMMS/lmms/issues/2982.

Spekular:
Fixes #4180.

### DIFF
--- a/plugins/GigPlayer/GigPlayer.cpp
+++ b/plugins/GigPlayer/GigPlayer.cpp
@@ -1074,7 +1074,7 @@ void GigInstrumentView::showFileDialog()
 	}
 	else
 	{
-		ofd.setDirectory( ConfigManager::inst()->gigDir() );
+		ofd.setDirectory( ConfigManager::inst()->userGigDir() );
 	}
 
 	m_fileDialogButton->setEnabled( false );
--- a/plugins/sf2_player/sf2_player.cpp
+++ b/plugins/sf2_player/sf2_player.cpp
@@ -1121,7 +1121,7 @@ void sf2InstrumentView::showFileDialog()
 	}
 	else
 	{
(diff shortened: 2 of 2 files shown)


Spekular:
Fixes #4180.

zonkmachine:
> Fixes #4180.

This line needs to be in the first post to automatically close the issue on merge.

Spekular:
Fixed "Fixes #4180" ;)

Spekular:
This fixes nothing.

PR #4966, opened 2019-04-29
Fix race condition in NotePlayHandleManager

### LINKED ISSUE #4956: Dragging the position fader back in song editor causes crash
LMMS Version: 1.2.0-rc8

In case you need the project file (does this also need the samples...?) i can upload that of course. 

btw: last time i used lmms was a couple of years ago, really nice progress :)

PhysSong:
Fixes #4956 by using a write lock when acquiring `NotePlayHandle`s.
It should be superseded by #4443 or similar on `master`.

### DIFF
--- a/src/core/NotePlayHandle.cpp
+++ b/src/core/NotePlayHandle.cpp
@@ -577,13 +577,9 @@ NotePlayHandle * NotePlayHandleManager::acquire( InstrumentTrack* instrumentTrac
 				int midiEventChannel,
 				NotePlayHandle::Origin origin )
 {
-	if( s_availableIndex < 0 )
-	{
-		s_mutex.lockForWrite();
-		if( s_availableIndex < 0 ) extend( NPH_CACHE_INCREMENT );
-		s_mutex.unlock();
-	}
-	s_mutex.lockForRead();
+	// TODO: use some lockless data structures
+	s_mutex.lockForWrite();
+	if (s_availableIndex < 0) { extend(NPH_CACHE_INCREMENT); }
 	NotePlayHandle * nph = s_available[ s_availableIndex.fetchAndAddOrdered( -1 ) ];
 	s_mutex.unlock();
 


JohannesLorenz:
I might be interested in reviewing this. Is it already ready to review, or do you plan further commits?

PhysSong:
> I might be interested in reviewing this. Is it already ready to review, or do you plan further commits?

I'm not going to add new commits unless there are some regressions.

JohannesLorenz:
Code is stylistic and functionally OK and should fix the mentioned issues.

The commit description confused me:
* in the mentioned case where two threads are calling the function and `s_availableIndex` is 1, one thread is doing an OOB read and only returns `nullptr` when you're lucky?
* the ABA problem occurs when there's a comparison for something to still be the same (still be "A"). I see the race condition, but wouldn't call it an ABA problem (where is the mentioned comparison?)

Is there any way to also test the changes? Maybe an issue report showing how to reproduce this?

Btw... `s_available[++s_availableIndex] = nph;` in `NotePlayHandleManager::release` can still be called simultaneously by 2 threads. Imagine two threads enter here, and the order is

```
s_available_index is 0
Thread 1: ++s_available_index;
Thread 2: ++s_available_index;
Thread 1: s_available[2] = nph;
Thread 2: s_available[2] = nph;
```
Isn't this leaving a hole at `s_available[1]`? Should this be fixed?

PhysSong:
> Btw... `s_available[++s_availableIndex] = nph;` in `NotePlayHandleManager::release` can still be called simultaneously by 2 threads. Imagine two threads enter here, and the order is
> 
> ```
> s_available_index is 0
> Thread 1: ++s_available_index;
> Thread 2: ++s_available_index;
> Thread 1: s_available[2] = nph;
> Thread 2: s_available[2] = nph;
> ```
> 
> Isn't this leaving a hole at `s_available[1]`? Should this be fixed?

The operations on `s_available_index` are atomic.
https://github.com/LMMS/lmms/blob/32df2d7fbad48c97322fdd1ebdffd0b5fee8fff8/include/NotePlayHandle.h#L352
So, this will happen instead:

```
s_available_index is 0
Thread 1: ++s_available_index; // gets 1 here because the operation is atomic
Thread 2: ++s_available_index; // gets 2 here because the operation is atomic
Thread 1: s_available[1] = nph;
Thread 2: s_available[2] = nph;
```

> * in the mentioned case where two threads are calling the function and `s_availableIndex` is 1, one thread is doing an OOB read and only returns `nullptr` when you're lucky?

That would be a better description.

> * the ABA problem occurs when there's a comparison for something to still be the same (still be "A"). I see the race condition, but wouldn't call it an ABA problem (where is the mentioned comparison?)

My description was wrong, but what I meant was:

1. `s_available_index` was 0
2. Thread 1: `++s_available_index;`, gets 1
3. Thread 2: `s_available_index--;`, gets 1
4. Thread 2: `NotePlayHandle * nph = s_available[1];` before writing
5. Thread 1: ` s_available[1] = nph;`

JohannesLorenz:
> The operations on s_available_index are atomic.

Oh, I thought this was just atomic `add`, but yeah, it's `fetch_add`.

I'm for merging this. If you want, you can edit the commit message to clarify things.

PR #4968, opened 2019-05-05
Nescaline and Freeboy - Better default sound

zonkmachine:
- Nescaline

Turned off channel 4, noise.

 - FreeBoy

Turned off channel 4, noise and turned down channel 1 sweep time.

Fixes https://github.com/LMMS/lmms/issues/4331


CYBERDEViLNL:
Code is good, does what this PR describes. :+1: But why turn down the sweep of FreeBoy?

zonkmachine:
> But why turn down the sweep of FreeBoy?

Yeah. I tried an even more basic sound while I was at the case. I've played around with it some more and I think I'll just change that one back actually.

zonkmachine:
OK, I reverted the FreeBoy sweep change.

PR #4973, opened 2019-05-15
Enhanced snapping in song editor

### LINKED ISSUE #874: Finer Scaling In The Sequencer

Spekular:
- New default behavior: Preserves offsets when moving clips, resizes in fixed increments.
- Adds shift + drag: Snaps moved start position (like current behavior) or end position (new), based on which is closest to real position. Snaps clip's edge when resizing. When moving a selection, the grabbed clip snaps into position and the rest move relative to it.
- Adds alt + drag: Allows fine adjustment of a clip's position or size, as an alternative to drag + ctrl.
- Adds a Q dropdown in the song editor to allow finer or coarser snapping (8 bars to 1/16th bar)
- Adds a proportional snap toggle. When enabled, snapping size/Q adjusts based on zoom, and a label appears showing the current snap size. In other words, onscreen distance between snap points remains the same regardless of zoom level. This is disabled by default.

Will close #874.

Ready to merge.

Enhanced

[Video Preview](https://www.youtube.com/watch?v=6M9gCW35S3A&feature=youtu.be)
(This shows a version where snapping modes are switched via toggle button. Since then I've switched to using modifier keys, but the snapping shown still works)


Spekular:
Issues:
 * I'd like to add an "adaptive" mode where the snapping will get finer relative to the current zoom level, but I can't figure out how to add a checkbox to the song editor
 * Icons/UI adjustments are needed, there's no icon for the new dropdown and no padding between it and the zoom level.
 * Doesn't affect where cloned patterns are dropped.
 * Doesn't preserve offsets. Based on how I implemented this, I don't think this should be a regression, but it would be nice. Especially at coarse Q's, the current behavior is awkward.

Spekular:
Regarding the last bullet point, perhaps a nudge/snap toggle (#4357) should be implemented in the song editor as well?

Spekular:
Icon ideas/mockups. Two far left are the current Quantization and zoom icons.

image

If someone feels like doing a reaction vote:
image

Spekular:
Implemented a snap/shift/snap & shift setting. Single clips can now be moved in snap (quantize final position), shift (quantize offset) or snap & shift (select which of the two is closest to mouse cursor at the current time). It doesn't work for group selections yet, I'd prefer not to implement it there until someone confirms if this setting is something we want to have or not.

JohannesLorenz:
The mockup voting is for a pixmap at the position where the current "Q" is?

What where your reasons to use the "Q" letter?

Spekular:
@JohannesLorenz The Q is there because this dropdown sets the snapping size/quantization used when moving and resizing clips (like in piano roll), so I think it works fine. The icon that needs to be made is for the proportional snap button, which is currently using the "add automation" icon.

Spekular:
Something that could be documented is the reason for the setInitialPositions() function.

The mouseMoveEvent() function is called continuously(?) on mouse move. The old code worked by getting the difference between the latest update and right now. It's possible to preserve offsets with this method, or snap to grid, but you CAN'T toggle between them. Once the pattern is snapped to grid, the latest position is quantized. This means all following positions are based on a quantized position, and as such the user's offset is permanently lost.

setInitialPositions() is called on mouse click, and saves:
- Mouse position
- Start position of the clicked TCO
- End position of the clicked TCO
- Offsets between every selected TCO and the clicked TCO.

Since you can't click again during a move/resize event (you're holding the mouse down to drag things around), calling setInitialPositions() on a click saves the position before move. (Sidebar: can you middle or right click while moving and destroy this logic?). By saving the initial position, you always have access to the initial offset. This means you can toggle between snapping to grid and snapping in increments without losing information. All you have to do is:
- Get the distance the mouse has moved since the last click (start of the move/resize)
- Convert mouse distance to ticks
- Add that to your initial position to get new position
- Quantize according to the current snap mode

Basically, I've rewritten a lot of the code to be relative to the initial position, rather than the latest position.

One thing that could be considered a downside to this is that, currently, move and resize events stick to their initial offset the whole time. If you begin a move event by fine-adjusting the position of a clip and then release ctrl/alt, the clip will go back to snapping to the initial offset. If we consider this to be a problem, I believe there's a solution. When fine moving, we should call setInitialPositions(), this saving the offset. This might need some additional changes, since it's possible that updating the initial positions this often could break other move code. We'd have to use "relative to latest position" code on fine adjustments, and "relative to initial position" on snapped.

src/core/Track.cpp
```diff
@@ -1118,6 +1180,41 @@ bool TrackContentObjectView::mouseMovedDistance( QMouseEvent * me, int distance
 
 
 
+/*! \bried Calculate the new position of a dragged TCO from a mouse event
```
Spekular:
```suggestion
/*! \brief Calculate the new position of a dragged TCO from a mouse event
```

src/gui/editors/SongEditor.cpp
```diff
@@ -728,13 +766,34 @@ SongEditorWindow::SongEditorWindow(Song* song) :
...
+	snap_lbl->setPixmap( embed::getIconPixmap( "quantize" ) );
+
+	//Set up quantization/snapping selector
+	m_snappingComboBox = new ComboBox( m_toolBar );
+	m_snappingComboBox->setFixedSize( 80, 22 );
+	m_snappingComboBox->setModel(m_editor->m_snappingModel);
+	m_snappingComboBox->setToolTip(tr("Clip snapping"));
+
+	m_setProportionalSnapAction = new QAction(embed::getIconPixmap("add_automation"),
+											 tr("Toggle roportional snap on/off"), this);
```
Spekular:
```suggestion
											 tr("Toggle proportional snap on/off"), this);
```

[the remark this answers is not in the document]
Spekular: Fixed, thanks!

Spekular:
BaraMGB reported a bug with this on Discord. Current steps to reproduce are:

- Click to create a clip
- Zoom in to 400%
- Click on the clip

The clip should jump over towards the left.
Current suspicions:

- This might have to do with the start position of the editor changing rather than zoom specifically?
- mapToParent() and setInitialPos() feel relevant for investigation

Spekular:
@BaraMGB I believe I've fixed the bug you discovered now.

BaraMGB:
This seems to work now. But I'm not sure about this button. It is quite misleading and I can not imagine a situation where a user needs this function. The snapping works like a charm, though. :+1:

src/gui/editors/SongEditor.cpp
```diff
@@ -32,6 +32,7 @@
 #include <QMdiArea>
 #include <QMdiSubWindow>
 #include <QPainter>
+#include <QDebug>
```
BaraMGB: ;)

src/gui/editors/SongEditor.cpp
```diff
@@ -261,7 +279,22 @@ void SongEditor::loadSettings( const QDomElement& element )
 	MainWindow::restoreWidgetState(parentWidget(), element);
 }
 
+float SongEditor::getSnapSize() const{
+	// 1 Bar is the third value in the snapping dropdown
+	int val = -m_snappingModel->value() + 3;
+	// If proportional snap is on, we snap to finer values when zoomed in
+	if (m_proportionalSnap){
```
BaraMGB: The curled braces in this block.

Spekular:
- Perhaps they want to keep snapping to one bar even when zoomed out.
- Maybe they're sequencing drums or similar and want to make sure they know
exactly what increment they're snapping to. With adaptive snap you have to
calculate the current snap level in your head vased on snap setting and
zoom.

I don't see what's misleading about it. The icon is a placeholder, if
that's what you mean.

BaraMGB:
The code looks fine except the small issues. About the button: perhaps someone else could test it. I find it misleading, because if I set 1 bar, I expect that the TCO also shifts by 1 bar.

Spekular:
Ah, so you want the adaptive snap to be disabled completely. I thought you wanted it to be permanently on.

src/core/Track.cpp
```diff
@@ -44,6 +44,7 @@
 #include <QMouseEvent>
 #include <QPainter>
 #include <QStyleOption>
+#include <QDebug>
```
Spekular:
```suggestion
```
Fix

src/core/Track.cpp
```diff
@@ -889,76 +897,85 @@ void TrackContentObjectView::mouseMoveEvent( QMouseEvent * me )
...
-									ppt ) );
-		if( ! ( me->modifiers() & Qt::ControlModifier )
-		   && me->button() == Qt::NoButton )
-		{
-			t = t.toNearestTact();
-		}
-		m_tco->movePosition( t );
+		MidiTime newPos = draggedTCOPos( me );
+
+		newPos = max( 0, newPos.getTicks() ); // Don't go left of bar zero
```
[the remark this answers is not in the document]
Spekular: Your call if you think I need to change it, otherwise I'll leave it as-is.
```suggestion
		// Don't go left of bar zero
		newPos = max( 0, newPos.getTicks() );
```

src/gui/editors/SongEditor.cpp
Spekular:
```suggestion
```
Fix

src/gui/editors/SongEditor.cpp
Spekular:
```suggestion
	if (m_proportionalSnap)
	{
```
Good catch

src/gui/editors/SongEditor.cpp
```diff
@@ -240,6 +243,20 @@ SongEditor::SongEditor( Song * song ) :
 	connect( m_zoomingModel, SIGNAL( dataChanged() ),
 					this, SLOT( zoomingChanged() ) );
 
+	//Set up snapping model, 2^i
+	for ( int i = 3; i >= -4; i-- ){
```
Spekular:
```suggestion
	for ( int i = 3; i >= -4; i-- )
	{
```

src/gui/editors/SongEditor.cpp
```diff
@@ -240,6 +243,20 @@ SongEditor::SongEditor( Song * song ) :
...
+		if ( i > 0 ){
```
Spekular:
```suggestion
		if ( i > 0 )
		{
```

src/gui/editors/SongEditor.cpp
```diff
@@ -240,6 +243,20 @@ SongEditor::SongEditor( Song * song ) :
...
+			m_snappingModel->addItem( QString( "%1 Bars").arg( 1 << i ) );
+		}
+		else if ( i == 0 ){
```
Spekular:
```suggestion
		else if ( i == 0 )
		{
```

src/gui/editors/SongEditor.cpp
```diff
@@ -240,6 +243,20 @@ SongEditor::SongEditor( Song * song ) :
...
+			m_snappingModel->addItem( "1 Bar" );
+		}
+		else {
```
Spekular:
```suggestion
		else
		{
```

src/gui/editors/SongEditor.cpp
```diff
@@ -728,13 +770,34 @@ SongEditorWindow::SongEditorWindow(Song* song) :
...
+	QLabel * snap_lbl = new QLabel( m_toolBar );
+	snap_lbl->setPixmap( embed::getIconPixmap( "quantize" ) );
+
+	//Set up quantization/snapping selector
+	m_snappingComboBox = new ComboBox( m_toolBar );
+	m_snappingComboBox->setFixedSize( 80, 22 );
+	m_snappingComboBox->setModel(m_editor->m_snappingModel);
+	m_snappingComboBox->setToolTip(tr("Clip snapping"));
+
+	m_setProportionalSnapAction = new QAction(embed::getIconPixmap("add_automation"),
```
Spekular:
```suggestion
	m_setProportionalSnapAction = new QAction(embed::getIconPixmap("proportional_snap"),
```

src/gui/editors/SongEditor.cpp
```diff
@@ -728,13 +770,34 @@ SongEditorWindow::SongEditorWindow(Song* song) :
...
+	//Set up quantization/snapping selector
+	m_snappingComboBox = new ComboBox( m_toolBar );
+	m_snappingComboBox->setFixedSize( 80, 22 );
+	m_snappingComboBox->setModel(m_editor->m_snappingModel);
+	m_snappingComboBox->setToolTip(tr("Clip snapping"));
+
+	m_setProportionalSnapAction = new QAction(embed::getIconPixmap("proportional_snap"),
+											 tr("Toggle proportional snap on/off"), this);
+	m_setProportionalSnapAction->setCheckable(true);
+	m_setProportionalSnapAction->setChecked(true);
```
Spekular:
```suggestion
	m_setProportionalSnapAction->setChecked(false);
```
Default to proportional snap being disabled.

BaraMGB:
Please, can somebody look at the proportional snapping? @zonkmachine @tresf @PhysSong

src/gui/editors/SongEditor.cpp
```diff
@@ -76,11 +76,14 @@ SongEditor::SongEditor( Song * song ) :
 	TrackContainerView( song ),
 	m_song( song ),
 	m_zoomingModel(new ComboBoxModel()),
+	m_snappingModel(new ComboBoxModel()),
+	m_proportionalSnap( true ),
```
PhysSong: Since you disabled proportional snap by default, this should be `false`.

src/core/Track.cpp
```diff
@@ -889,76 +896,86 @@ void TrackContentObjectView::mouseMoveEvent( QMouseEvent * me )
...
 			}
-			m_tco->changeLength( t );
+			else
+			{	// Otherwise, resize in fixed increments
+				MidiTime initialLength = m_initialTCOEnd - m_initialTCOPos;
+				MidiTime offset = MidiTime( l - initialLength ).quantize( snapSize );
+				MidiTime min = MidiTime( initialLength % (int)(MidiTime::ticksPerTact() * snapSize) );
+				l = qMax<int>( min, initialLength + offset);
+			}
+			l = qMax<int>( minLength, l );
```
PhysSong: In some cases, part of the length which is out of quantization won't be preserved. For example:
1. Suppose there's a TCO whose length is 1 bar.
2. Shrink the TCO when the quantization size is 1 bar. Now the length of the TCO is 1/16 bars.
3. Release the mouse and extend the TCO again, it would be 1 and 1/16 bars instead of 1 bar.

src/core/Track.cpp
```diff
@@ -971,13 +988,38 @@ void TrackContentObjectView::mouseMoveEvent( QMouseEvent * me )
...
-				if( ! ( me->modifiers() & Qt::ControlModifier )
-						&& me->button() == Qt::NoButton )
+				float snapSize = gui->songEditor()->m_editor->getSnapSize();
+				//When snapping, avoid making the new TCO tiny
+				MidiTime minLength = MidiTime( MidiTime::ticksPerTact() / 16 );
+
+				// If the user is holding alt, or pressed ctrl after beginning the drag, don't quantize
+				if(    me->button() != Qt::NoButton
+				   || (me->modifiers() & Qt::ControlModifier)
+				   || (me->modifiers() & Qt::AltModifier)    )
```
PhysSong: Odd usage of multiple whitespaces.

PhysSong:
Would it be better to change the displayed snap size according to the real size? In that case, should we restore previous snap size when disabling the feature?

src/core/Track.cpp
Spekular: The length/position is always based on the initial value, or on the latest fine adjust, so I don't quite see how this is a bug. To me, releasing the mouse = applying the change. What's the alternative?

src/core/Track.cpp
Spekular: Do you mean in the if statement? I aligned the me-> part across conditions, and spaced the final parenthasis to be left of the entire condition. I can change the format if you indicate what would be preferred.

src/gui/editors/SongEditor.cpp
Spekular: Good catch. I'll hold off on changing this until we resolve your comment about displaying the "real" snap size.

Spekular:
@PhysSong I think dynamically changing the values in a drop-down selector is probably overly complicated and prone to bugs. Do we have a nice component for displaying text but not selecting it? If so, the layout could be changed to:

- Dropdown selector: 0.5X, 1X, 2X...
- Proportional Snap Toggle
- Snap size display

Though there are some issues with this as well. Since it would be confusing to have two separate bar lengths showing, I think the drop down would have to change to a factor. However, this might be hard to interpret. I would want 2X to give me twice the resolution, AKA snap to half bars. Someone else might argue 2X should snap to twice the size, AKA snap to two bars.

src/gui/editors/SongEditor.cpp
Spekular:
```suggestion
	m_proportionalSnap( false ),
```

Spekular:
Updated. When proportional snap is off, the quantization dropdown will display the tooltip "Clip snapping size. When proportional snap is on, the dropdown changes it's tooltip to "Base snapping size", and a label appears titled "Snap: X Bars", where X is the current snap level.

src/gui/editors/SongEditor.cpp
```diff
@@ -264,6 +285,44 @@ void SongEditor::loadSettings( const QDomElement& element )
...
+
+QString SongEditor::getSnapSizeString() const
+{
+	int val = -m_snappingModel->value() + 3;
+	val = val - m_zoomingModel->value() + 3;
+	val = max(val, -6); // -6 gives 1/64th bar snapping. Lower values cause crashing.
+
+	if ( val >= 0 ){
+		int bars = 1 << val;
+		return QString( "%1 Bar" ).arg(bars);
```
Spekular:
```suggestion
		if ( bars == 1 ) { return QString("1 Bar");
		else
		{
			QString( "%1 Bars" ).arg(bars);
		}
```

src/gui/editors/SongEditor.cpp
```diff
@@ -264,6 +285,48 @@ void SongEditor::loadSettings( const QDomElement& element )
...
+	int val = -m_snappingModel->value() + 3;
+	val = val - m_zoomingModel->value() + 3;
+	val = max(val, -6); // -6 gives 1/64th bar snapping. Lower values cause crashing.
+
+	if ( val >= 0 ){
+		int bars = 1 << val;
+		if ( bars == 1 ) { return QString("1 Bar");
+		else
+		{
+			QString( "%1 Bars" ).arg(bars);
```
Spekular:
```suggestion
			return QString( "%1 Bars" ).arg(bars);
```
Scala really spoils me.

src/gui/editors/SongEditor.cpp
```diff
@@ -264,6 +285,48 @@ void SongEditor::loadSettings( const QDomElement& element )
...
+
+QString SongEditor::getSnapSizeString() const
+{
+	int val = -m_snappingModel->value() + 3;
+	val = val - m_zoomingModel->value() + 3;
+	val = max(val, -6); // -6 gives 1/64th bar snapping. Lower values cause crashing.
+
+	if ( val >= 0 ){
+		int bars = 1 << val;
+		if ( bars == 1 ) { return QString("1 Bar");
```
PhysSong: `}` is missing.

src/core/Track.cpp
PhysSong: What I meant was something like #3989. Related suggestion in the issue: https://github.com/LMMS/lmms/issues/3989#issuecomment-353770092

Spekular: I suppose I could/should change minLength to be ``note length modulo snap size``. I believe I set it to 1/16ths because at that size a clip can still be resized back up at 100% zoom. 1/8ths and smaller are a bit troublesome because they're very hard or impossible to size back up unless you zoom in.

Spekular: Latest commit ought to address this!

PhysSong: Works great! :+1:

src/core/Track.cpp
```diff
@@ -889,76 +896,87 @@ void TrackContentObjectView::mouseMoveEvent( QMouseEvent * me )
...
 		}
 	}
 	else if( m_action == Resize || m_action == ResizeLeft )
 	{
 		if( m_action == Resize )
 		{
-			MidiTime t = qMax( MidiTime::ticksPerTact() / 16, static_cast<int>( me->x() * MidiTime::ticksPerTact() / ppt ) );
-			if( ! ( me->modifiers() & Qt::ControlModifier ) && me->button() == Qt::NoButton )
+			MidiTime l = static_cast<int>( me->x() * MidiTime::ticksPerTact() / ppt );
+			float snapSize = gui->songEditor()->m_editor->getSnapSize();
```
PhysSong: It seems like many lines from here are used in many places with very small variations. Can we somehow consolidate them?

Spekular: snapSize and perhaps the recently added snapLength could probably be moved up at least one level, so they're defined once for both resize modes rather than for each separately. Moving them up any further risks exposing them to if statements that don't need them though, IIRC. The repeated modifier check could also be in a variable, perhaps? The rest feels tricky to consolidate as it's slightly different for each case.

Spekular: I changed a bit but I'm not sure if it helps much @PhysSong.

Spekular:
I considered one other way to "consolidate" some of the logic, but it doesn't really decrease LOC or improve readability imo. Here's a gist for whatever it's worth: https://gist.github.com/Spekular/f29d722df27d283cba62c12c0d01a341

Also added snapping of clip's end position in shift mode.

Spekular:
Travis and Appveyor are failing on a line that doesn't exist on my machine, nor on my branch on github... Fun. Will try to rebase and fix, PR 5043 introduces a line that uses a function I replaced.

Spekular:
I have no idea what git is doing or why.

PhysSong:
I will merge this in a few days if there are no objections or new reviews.

DouglasDGI:
@Spekular Hey, thank you for the PR!  I've been wanting this feature for a long time now.
I tested it out and I have not really noticed any problems.  I tested it with messed up time signatures, groups of TCOs, individual TCOs, everything.  It all works perfectly, except for one small thing which might not actually be a bug.

If your TCO's edge does not match up with a snap line, you can use Shift+drag to make it match up.  But, if you use Shift+drag to do that, and let go of the Shift without letting go of the mouse, it will revert to the normal behavior, reverting the changes made by holding Shift even after those changes are displayed on the screen.  So if you want to match something up with a snap line and then resize it, you have to Shift+drag it, let go of your mouse button and Shift key, and then resize it normally.  That may or may not be intentional, I'm not sure, but it may be confusing to people.  It's a very minor thing though, and people may prefer either behavior.

Otherwise, everything seems to be working 100% perfectly.  :+1:

Spekular:
The shift behavior is intentional, as it allows you to alternate between
offset snap and grid snap without losing the offset. Otherwise pressing
shift once would lock the whole move to grid snap, leaving no way to go
back.

Thanks a bunch for the testing :)

PhysSong:
@DouglasDGI Are you fine with the new shift behavior?

DouglasDGI:
@PhysSong Yes, it's just fine.  I approve.

Spekular:
> I will merge this in a few days

@PhysSong now?

PhysSong:
Okay. Could you write a suitable commit message for this PR? The default message in `Squash and merge` is too messy.

Spekular:
A full length one, or the short title? A long one could probably be copied from the first post in this thread. As for a short one, I have trouble with those but here's some attempts

- If we should mention the core change in MidiTime: "Replace MidiTime's 'toNearestTact' with 'quantize', allow different snap sizes in song editor"
- Otherwise: "Add snap size selection and new snap behaviors to song editor"

PhysSong:
Thanks. Do you think the summary in the first post is okay for long one?

Spekular:
Yes, I do. Just change "will close" to "closes" and remove everything after that line and it seems good to me.

Spekular:
🎉

musikBear:
Will there be binaries for tests?

Spekular:
Master branch should get nightly builds at some point in the future, considering that I haven't planned on building binaries for this.

PR #4975, opened 2019-05-16
fix sample track out of sync with jack on 48000 hz

### LINKED ISSUE #4786: sample track out of sync with jack
Hi,
my sample tracks are out of sync with the midi tracks when not starting playback from the beginning.

I'm using 1.2.0 rc7 on Linux.

BaraMGB:
fixes  https://github.com/LMMS/lmms/issues/4786

### DIFF
--- a/src/core/SampleBuffer.cpp
+++ b/src/core/SampleBuffer.cpp
@@ -73,7 +73,7 @@ SampleBuffer::SampleBuffer( const QString & _audio_file,
 	m_amplification( 1.0f ),
 	m_reversed( false ),
 	m_frequency( BaseFreq ),
-	m_sampleRate( Engine::mixer()->baseSampleRate() )
+	m_sampleRate( Engine::mixer()->processingSampleRate() )
 {
 	if( _is_base64_data == true )
 	{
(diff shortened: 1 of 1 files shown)


Spekular:
For whatever my code review is worth:

No cosmetic issues since you're just replacing functions, no concern about the actual logic of the fix either. Have you tested if this fix works with oversampling enabled?

Spekular (approved):

src/core/SampleBuffer.cpp
```diff
@@ -378,10 +378,10 @@ void SampleBuffer::normalizeSampleRate( const sample_rate_t _src_sr,
 							bool _keep_settings )
 {
 	// do samplerate-conversion to our default-samplerate
```
Spekular:
```suggestion
	// do samplerate-conversion to our processing-samplerate
```

Spekular (changes requested):

src/core/SampleBuffer.cpp
Spekular: On second thought, this comment should be changed, unless this line was changed by mistake.

BaraMGB:
I close this in favor of a solution of @Reflexe

PR #4976, opened 2019-05-17
playing/recording pianoRoll's chord notes (#4963)

### LINKED ISSUE #4963: Piano Roll - Only base note sounds when recording chords
While recording in the Piano Roll with a chord selected (drop down menu), only the base note is played/sounds even though the rest of the notes is properly recorded.
Related https://github.com/LMMS/lmms/pull/4938

sharpblade4:
This change makes pianoRoll chord settings to generate notes accordingly, and record them (fixes #4963).  It can be triggered by mouse and keyboard, however it does not supported with an external MIDI controller (as current behavior). I'll work on another change that will enable the chord playing (and thus recording) of a base note that was generated by an external MIDI device in future (because it needs different approach, maybe MIDI messages manipulating).

### DIFF
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -1894,7 +1894,7 @@ void PianoRoll::testPlayKey( int key, int velocity, int pan )
 	// turn off old key
 	pianoModel->handleKeyRelease( m_lastKey );
 	// if a chord was set, stop the chords notes as well
-	pauseChordNotes(key);
+	pauseChordNotes(m_lastKey);
 
 	// remember which one we're playing
 	m_lastKey = key;
@@ -2032,14 +2032,15 @@ void PianoRoll::mouseReleaseEvent( QMouseEvent * me )
 			{
 				m_pattern->instrumentTrack()->pianoModel()->
 						handleKeyRelease( note->key() );
-				note->setIsPlaying( false );
 				pauseChordNotes(note->key());
+				note->setIsPlaying( false );
 			}
 		}
 
 		// stop playing keys that we let go of
 		m_pattern->instrumentTrack()->pianoModel()->
 						handleKeyRelease( m_lastKey );
+		pauseChordNotes(m_lastKey);
 	}
 
 	m_currentNote = NULL;
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -1056,9 +1056,11 @@ void PianoRoll::keyPressEvent(QKeyEvent* ke )
 	{
 		const int key_num = PianoView::getKeyFromKeyEvent( ke ) + ( DefaultOctave - 1 ) * KeysPerOctave;
 
-		if(! ke->isAutoRepeat() && key_num > -1)
+		if (!ke->isAutoRepeat() && key_num > -1)
 		{
-			m_pattern->instrumentTrack()->pianoModel()->handleKeyPress( key_num );
+			m_pattern->instrumentTrack()->pianoModel()->handleKeyPress(key_num);
+			//  if a chord is set, play all chord notes (simulate click on all):
+			playChordNotes(key_num);
 			ke->accept();
 		}
 	}
@@ -1256,10 +1258,11 @@ void PianoRoll::keyReleaseEvent(QKeyEvent* ke )
 	if( hasValidPattern() && ke->modifiers() == Qt::NoModifier )
 	{
 		const int key_num = PianoView::getKeyFromKeyEvent( ke ) + ( DefaultOctave - 1 ) * KeysPerOctave;
-
-		if( ! ke->isAutoRepeat() && key_num > -1 )
+		if (!ke->isAutoRepeat() && key_num > -1)
 		{
-			m_pattern->instrumentTrack()->pianoModel()->handleKeyRelease( key_num );
+			m_pattern->instrumentTrack()->pianoModel()->handleKeyRelease(key_num);
+			// if a chord is set, simulate click release on all chord notes
+			pauseChordNotes(key_num);
 			ke->accept();
 		}
 	}
@@ -1705,7 +1708,9 @@ void PianoRoll::mousePressEvent(QMouseEvent * me )
 			{
 				// left click - play the note
 				int v = ( (float) x ) / ( (float) WHITE_KEY_WIDTH ) * MidiDefaultVelocity;
-				m_pattern->instrumentTrack()->pianoModel()->handleKeyPress( key_num, v );
+				m_pattern->instrumentTrack()->pianoModel()->handleKeyPress(key_num, v);
+				// if a chord is set, play the chords notes as well:
+				playChordNotes(key_num, v);
 			}
 		}
 		else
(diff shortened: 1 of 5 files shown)


JohannesLorenz:
I'd like to review this. It's ready for review, right?

Do you want this to go into 1.2.0 or 1.2.1?

sharpblade4:
Yes it is ready. It is working for mouse \ keyboard inputs but not for external MIDI controller inputs (for those, the behavior is as it was until now - only base note sounds and recorded).
I don't know for which version it is better, I think 1.2.0?

JohannesLorenz:
Unfortunatelly, doesn't always work for me. What I do is:

- Open a new project
- Add a note section in the 3osc track by clicking somwhere in it (maybe not at the first bar)
- Open the note section in piano roll
- Select a chord there, e.g. "moll"
- Click record
- Play letters on your PC's keyboard

The notes are often placed at the wrong positions or sometimes missing.

How did you test it?

JohannesLorenz:
Per the question, 1.2.0 vs 1.2.1, that may also depend on the release date of 1.2.0. I think 1.2.1 is OK, but we'll see.

sharpblade4:
Weird - I tested it with quite the same flow you described. Clicking "record" is not a mandatory step though, since it should play the chord live as well. 
The only difference I see is that you choose "moll" chord, which I don't have- and I saw that this is german for "minor". 

Anyway, I'll be able to check this only later. I'll update after I do that.

sharpblade4:
It works for me. Attaching a screenshot. Do you have any idea on what can I do to find why it doesn't for you? 
Screenshot from 2019-05-20 18-32-48

JohannesLorenz:
Heh, I just found out it worked on another computer. The reason was that I had "SDL" selected for audio device in the settings on this computer. With other backends, it works.

I'll go on with the review.

src/gui/editors/PianoRoll.cpp
```diff
@@ -3530,22 +3580,6 @@ void PianoRoll::finishRecordNote(const Note & n )
 						it->key(), it->getVolume(),
 						it->getPanning() );
 				n1.quantizeLength( quantization() );
-
-				//Get selected chord
-				const InstrumentFunctionNoteStacking::Chord & chord = InstrumentFunctionNoteStacking::ChordTable::getInstance()
-					.getChordByName( m_chordModel.currentText() );
-
-				if( !chord.isEmpty() )
```
JohannesLorenz: Can you please explain why you removed this? It looks right, I just can't see it.

src/gui/editors/PianoRoll.cpp
```diff
@@ -1985,6 +2033,7 @@ void PianoRoll::mouseReleaseEvent( QMouseEvent * me )
 				m_pattern->instrumentTrack()->pianoModel()->
 						handleKeyRelease( note->key() );
 				note->setIsPlaying( false );
```
JohannesLorenz: I'd prefer to have all `Note::setIsPlaying` after all chord notes are paused, for consistency, and also to avoid race conditions in the future.

src/gui/editors/PianoRoll.cpp
```diff
@@ -1056,9 +1056,11 @@ void PianoRoll::keyPressEvent(QKeyEvent* ke )
 	{
 		const int key_num = PianoView::getKeyFromKeyEvent( ke ) + ( DefaultOctave - 1 ) * KeysPerOctave;
 
-		if(! ke->isAutoRepeat() && key_num > -1)
+		if (!ke->isAutoRepeat() && key_num > -1)
 		{
-			m_pattern->instrumentTrack()->pianoModel()->handleKeyPress( key_num );
+			m_pattern->instrumentTrack()->pianoModel()->handleKeyPress(key_num);
```
JohannesLorenz: Would it make sense to always handle the base note key press/release in the chord functions aswell? That way, the code would get a bit smaller. Though I'm also OK with this solution.

JohannesLorenz:
I had an issue with both jack and pulse: Open triple oscillator piano roll, select minor, press key "C5" for a short duration (maybe 1/8 note) by simply using the left mouse key. When I did that, only the base key did release.

Can you reproduce this?

sharpblade4:
> I had an issue with both jack and pulse: Open triple oscillator piano roll, select minor, press key "C5" for a short duration (maybe 1/8 note) by simply using the left mouse key. When I did that, only the base key did release.
> 
> Can you reproduce this?

Yes. This is a bug (all clicks directly on the pianoKeys graphics will release the base note only). Sorry. I'll fix that in addition to the other comments asap.

src/gui/editors/PianoRoll.cpp
sharpblade4: I think it would force an additional condition each time a note is played (that is, before playing the chord\note, check if there is a chord - then do what's necessary). In my opinion this is less readable.

src/gui/editors/PianoRoll.cpp
sharpblade4: You're right

src/gui/editors/PianoRoll.cpp
sharpblade4: Until now chords were recorded by this and the actual chord's notes were not real, so this code virtually added them to the m_pattern, knowing that they should be added since there is a real base note. 
Now every base note is played with the rest of the chord notes (notes are actually generated on and turned off when paused). A signal of midiNoteOff is emitted for each of them, triggering the record for them just like the base note.  If I'll keep this code here - the chord will be recorded twice.

sharpblade4:
Thank you for finding the bug, and once again sorry for that. 
It is now fixed, along with the other comments in the initial CR.

JohannesLorenz:
Np for the bug. Good you could fix it.

Btw, this is unrelated to your PR, but: If you take triple oscillator and slide over the keyboard fast (e.g. from C2 to C4), are some notes displayed as still pressed (while they don't sound anymore)?

JohannesLorenz:
Functional and style review passed. Tests passed. You can squash-merge it, or let me know if I should.

JohannesLorenz:
Oh, note, I'm not sure about the merge strategy, I'll ask on discord. Sorry, please wait.

sharpblade4:
Regarding the notes that keep beign displayed when you slide really fast over them - I notice this as well. Planned to open an issue for it later on.
On the subject of the squash-merge, I actually don't know what do you mean. In general, after a PR is approved, should I be the one to do something? because in my previous PR here I didn't.

JohannesLorenz:
Usually I leave it to the author to merge. However, I can't see whether an author has the rights to merge. If you don't see the merge button, then I can do it.

I'll first have to ask if we can merge shortly before the 1.2.0 release.

sharpblade4:
I don't have the 'merge' button here

JohannesLorenz:
It looks like this PR is a feature (as stated in #4989), and stable-1.2 only allows fixes. If you agree, it should be better re-based and re-targetted on/to master, and this PR should be closed instead.

Sorry for the overhead!

sharpblade4:
Alright. Why close this PR? I think I can change it to commit into branch master. 
I'll try that now. If that's not what you meant please elaborate what I should do.

JohannesLorenz:
Yeah, `git cherry-pick` can make the copying easier. *can*...

PR #4977, opened 2019-05-17
Ignore VS Code config files.

enp2s0:
Simple fix to prevent the .vscode folder from being committed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 .*.sw?
 .DS_Store
 *~
+.vscode
 /CMakeLists.txt.user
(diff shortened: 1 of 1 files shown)


PhysSong:
Thanks! There might be some more entries to add, e.g. https://gitignore.io/api/visualstudiocode

noahb01:
I wonder if it would be better to whitelist what we want rather than try to blacklist files created by every IDE/toolchain.

PhysSong:
I'm closing this for now. See #5062 for related discussions.

PR #4982, opened 2019-05-19
SampleTrack: call requestChangesInModel before unref SampleBuffer

### LINKED ISSUE #4980: Removing a SampleTCO while it plays will crash lmms (master and latest rc)
## Steps to reproduce 
 - Open lmms
- load a sampletrack (doubleclick on a sample track and select an audio file)
- play it
- remove it while it plays (middle mouse button) 
- see if lmms crashes

Please verify it so I'll work on a solution.

Reflexe:
To avoid a race condition between the gui thread which destroys the
samplebuffer and the mixer thread which increases the buffer's ref-
count, we'll make sure to touch the ref-count only when whe're synced
with the mixer.

Closes #4980.

### DIFF
--- a/src/tracks/SampleTrack.cpp
+++ b/src/tracks/SampleTrack.cpp
@@ -110,7 +110,9 @@ SampleTCO::~SampleTCO()
 	{
 		sampletrack->updateTcos();
 	}
+	Engine::mixer()->requestChangeInModel();
 	sharedObject::unref( m_sampleBuffer );
+	Engine::mixer()->doneChangeInModel();
 }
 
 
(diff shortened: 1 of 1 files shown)


src/tracks/SampleTrack.cpp
```diff
@@ -133,7 +135,9 @@ const QString & SampleTCO::sampleFile() const
 
 void SampleTCO::setSampleBuffer( SampleBuffer* sb )
 {
-	sharedObject::unref( m_sampleBuffer );
+  Engine::mixer()->requestChangeInModel();
+        sharedObject::unref( m_sampleBuffer );
```
PhysSong: Spaces.

PhysSong: Also, the indentation looks somewhat weird to me.

PR #4988, opened 2019-05-22
Refactoring for the recording branch

Reflexe:
* Use async methods for blocking operations in the gui thread. 
* Clean and refactor a lot of code around SampleBuffer.
* Make it easier to use the mixer's rendering guard.
* Eliminate a few dataraces between the gui thread and the mixer thread. 
* more.


Reflexe:
I think I'll have to revent af24824 and 87f7af5 to get mingw to build. However, I think we're ready to review.

include/SampleTrack.h
```diff
@@ -123,10 +130,14 @@ public slots:
 private:
 	SampleTCO * m_tco;
 
-	SampleBufferVisualizer m_sampleBufferVisualizer;
-} ;
+	std::shared_ptr<std::mutex> m_sampleBufferVisualizerMutex = std::make_shared<std::mutex>();
```
PhysSong: `std::mutex` is unavailable in MinGW with win32 threading model as well as `std::thread`. You can use [mingw-std-threads](https://github.com/meganz/mingw-std-threads) as in 48cdce9a790b6ba7384a36d5b4a6ffadc21a89bd or switch to `QMutex`.
Also, don't forget to include `<mutex>` if you're going to stick to `std::mutex`. MSVC complains about it.

Reflexe: Thread sanitizer doesn't work well with QMutex. However, I'll just stick to it for now and then add that option.

include/SampleBuffer.h
```diff
@@ -26,385 +26,396 @@
...
 
-	void setAllPointFrames( f_cnt_t _start, f_cnt_t _end, f_cnt_t _loopstart, f_cnt_t _loopend )
-	{
-		m_startFrame = _start;
-		m_endFrame = _end;
-		m_loopStartFrame = _loopstart;
-		m_loopEndFrame = _loopend;
+	template<class LaunchType=LaunchType::Async>
+	void setLoopStartFrame(f_cnt_t _start) {
+		runToSetPlayInfo<LaunchType>([=](GuardedPlayInfo &playInfo) {
```
PhysSong: To fix build errors for Windows, you should either move bodies of this kind of functions to the `.cpp` file or add `LMMS_EXPORT` to `internal::SampleBufferPlayInfo` and `internal::SampleBufferData`.

include/internal/SampleBufferData.h
```diff
@@ -0,0 +1,78 @@
...
+
+#include <vector>
+#include <QtCore/QString>
+#include "Mixer.h"
+#include "Engine.h"
+#include "MemoryManager.h"
+#include "lmms_basics.h"
+
+namespace internal {
+	sample_rate_t GetMixerSampleRate();
```
PhysSong: Function names should be `camelCase` starting with the lower case. Also, please consider removing the `get` prefix here if it's not necessary.

Reflexe: Thanks! I got confused with Google's code style.

Reflexe:
Travis fails because of a strange carla problem (I have not touched carla's files); I'll ignore it for now. 
Currently, this PR is ready for reviews and testers. Then, I'll wait one week and finally merge it into master (if it is ok, of cause).

PR #4995, opened 2019-05-27
Removed unnecessary SF2 directory setting

### LINKED ISSUE #4180: SF2 Directory Setting/Relative Paths

necrashter:
Closes #4180 

Removed every occurrence of SF2 directory in SetupDialog as well as ConfigManager. You might want to check my modification in the plugins/sf2_player/sf2_player.cpp though. I didn't understand what original code does; removed it anyway.

### DIFF
--- a/include/ConfigManager.h
+++ b/include/ConfigManager.h
@@ -161,11 +161,6 @@ public:
 		return m_gigDir;
 	}
 
-	const QString & sf2Dir() const
-	{
-		return m_sf2Dir;
-	}
-
 	const QString & vstDir() const
 	{
 		return m_vstDir;
@@ -266,7 +261,6 @@ private:
 	QString m_vstDir;
 	QString m_ladDir;
 	QString m_gigDir;
-	QString m_sf2Dir;
 	QString m_version;
 #ifdef LMMS_HAVE_STK
 	QString m_stkDir;
--- a/include/SetupDialog.h
+++ b/include/SetupDialog.h
@@ -76,7 +76,6 @@ private slots:
 	void setWorkingDir( const QString & _wd );
 	void setVSTDir( const QString & _vd );
 	void setGIGDir( const QString & _gd );
-	void setSF2Dir( const QString & _sfd );
 	void setArtworkDir( const QString & _ad );
 	void setLADSPADir( const QString & _ld );
 	void setSTKDir( const QString & _sd );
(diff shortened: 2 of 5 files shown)


Spekular:
I disagree with tresf on removing this directory. It adds work if someone later wishes to improve relative paths, since this will have to be reverted. It also results in user's sf2 directory configuration being lost. Finally, I believe we found the directory *does* affect where sf2 browsing opens to, which is valuable in itself.

necrashter:
Thanks for feedback, sir.

> I believe we found the directory _does_ affect where sf2 browsing opens to, which is valuable in itself.

This seems to be the only function of this setting. But the soundfonts selected from SF2 directory are not saved relatively, i.e. it saves their absolute path. Whereas soundfonts from samples/soundfonts directory are saved relatively.

To be frank, I like the idea of configurable SF2 directory so that multiple programs can use soundfonts from the same directory; however it's poorly implemented right now.

Should we remove this setting and use samples/soundfonts directory, or should we close this pull request and make soundfont paths relative to the SF2 directory?

Spekular:
I believe the correct solution to this is to improve our handling of
relative paths, hence why I argue for keeping the setting. The "only"
problem with that solution is that it's much more difficult than axing the
setting.

Spekular:
Relevant: #5117

JohannesLorenz:
Since #4180 got re-milestoned to 1.3.0, I'll change the PR base from stable-1.2 to master in 7 days. (Or you can also do it now @necrashter)

Spekular:
I'm closing this since I'm pushing to get #5117 merged before the refactor in #5592

PR #4994, opened 2019-05-26
Recording after rebase to master

Reflexe:
This is a refactored and rebased-to-master version of #4531 . 

## Bugs
- [x] TripleOscillator preset previews show errors and don't stop due to nonexistent `samples/empty.wav`
- [x] `w2_invsinehalf.flac` has a furry tail in BitInvader due to resampling(maybe an issue with Graph?)
- [x] Visualization gets weird when resizing tracks while recording

## Enhancements

- [ ] (optional) Add more backends
  - [ ] ALSA
  - [ ] CoreAudio
  - [ ] Sndio
  - [ ] libsoundio
- [x] (optional) Refactor
- [x] Decide when to resample data
- [x] Optimize

PR #4998, opened 2019-05-29
Knife Tool for Sample Clips

Spekular:
~~This PR builds off of #4973, so that one needs to be reviewed and merged first.~~

~~Needs to be updated with all the changes from #4973 before merge, somehow.~~

Mergeable once again!

Adds a knife tool for splitting sample clips, based on BaraMGB's work.

Preview of feature
Screeshot shows, from left to right: Unquantized cut, quantized cut before releasing mouse, cut quantized to clip boundaries (in this case, two bars left of the clip's end)

TODO:
- Cursor for knife mode
- Undo correctly
- Short hang when cutting playing clip, according to @PhysSong. Still present in latest commit?

Usage:
- Select knife mode.
- Click and hold on a sample clip. A marker appears showing where the split will occur.
- The mouse can be dragged around. Split position is quantized to song editor Q by default
- Holding ctrl or alt disables quantization.
- Holding shift snaps the cut position to a quantized offset from the clip's left or right edge, whichever is closer to the cursor. This means at least one of the new clips after the cut will have a quantized length (both, if the original length is a multiple of the current snap size).
- On release, the clip will be split, unless the mouse was moved off to either side (or is exactly at the start/end position).


Gabrielxd195:
friend, this also serves for tco or piano roll clips ?.  is that this tool looks great.

Spekular:
This currently only works for sample clips.

Spekular:
I'd love it if git didn't completely destroy everything every time I tried to rebase. Fetched and pulled on snap branch, checked out knife branch, fetched and pulled again, "rebase spek SESnap", everything explodes.

PhysSong:
I can help you with rebasing commits.

Spekular:
It seems to have fixed itself once I fixed the merge "conflict" that there was an extra newline somewhere. For a while it was displaying changes as if I'd touched tens of files, even though the "changes" matched master exactly. I may have been trigger happy in my complaining, since last time was a huge timesink.

PhysSong:
I tried rebasing commits. Surprisingly, there were only 2 commits after rebasing. If you want a clean branch, I can push the rebased one.

Spekular:
Feel free to do so!

PhysSong:
Rebased. You can pull the branch with
```
git checkout Knife
git fetch spek Knife
git reset hard spek/Knife
```

Spekular:
Great! Lots of refactoring coming soon, and a shift mode for this as well,
that snaps the length of the new left/right clip.

Just one tiny issue to fix: Cuts end up one pixel to the left of where they
should be at 200% zoom and less. Might have to do with the fact that the
patterns are drawn one pixel to the left of the bar lines.

One fix is to add a pixel to the calculations and see id it helps. Another
is to do the quantization twice, once for drawing and once for cutting, but
this results in essentially duplicate code. Therefore I'm trying the pixel
solution first, even though it feels hacky.

Spekular:
Fixing that "tiny issue" took a while. You'd think that the result from knifeMarkerPos( me ) could be converted back to a MidiTime and used for the actual cut itself, but for some reason the shift/default behavior ends up inverted if you do. Adding a boolean switch to flip the two still left some weirdness, so I moved the code that decides the position back into mouseReleaseEvent. After a ton of trial and error I think it's working, but I can't say I understand why it has to be the way it is.

Spekular:
With the icon added, I don't think there's anything left to add to this. Ready for code review and testing!

DouglasDGI:
@Spekular If you right click while holding left click with the knife tool, the black marker line thing stays behind (even after switching tools).  If you move the TCO with the black line left behind, then the mark will stay there permanently.

Spekular:
Good catch. I can probably just add a call to disable the marker in the RMB
handling, will check when I get the time.

include/SampleTrack.h
```diff
@@ -73,6 +73,15 @@ class SampleTCO : public TrackContentObject
...
 
+	void inline setMarkerPos( int x )
+	{
+		m_markerPos = x;
+	}
+	void inline setMarkerEnabled( bool e )
+	{
+		m_marker = e;
+	}
+
```
BaraMGB: Let's remove the spaces inside the brackets in new code.

src/core/Track.cpp
```diff
@@ -812,6 +840,14 @@ void TrackContentObjectView::mousePressEvent( QMouseEvent * me )
 		{
 			remove();
 		}
+		if (m_action == Split){
```
BaraMGB: the brace belongs to the next line.

src/core/Track.cpp
```diff
@@ -812,6 +840,14 @@ void TrackContentObjectView::mousePressEvent( QMouseEvent * me )
...
+			m_action = NoAction;
+			SampleTCO * sTco = dynamic_cast<SampleTCO*>( m_tco );
+			if (sTco) {
```
BaraMGB: here as well.

src/core/Track.cpp
```diff
@@ -1034,6 +1070,12 @@ void TrackContentObjectView::mouseMoveEvent( QMouseEvent * me )
 						MidiTime::ticksPerTact() ) );
 		s_textFloat->moveGlobal( this, QPoint( width() + 2, height() + 2) );
 	}
+	else if( m_action == Split )
+	{
+		SampleTCO * sTco = dynamic_cast<SampleTCO*>( m_tco );
+		sTco->setMarkerPos( knifeMarkerPos( me ) );
```
BaraMGB: You forgot to test sTco after dynamic_cast.

src/core/Track.cpp
```diff
@@ -1070,12 +1112,56 @@ void TrackContentObjectView::mouseReleaseEvent( QMouseEvent * me )
...
+	else if( m_action == Move || m_action == Resize || m_action == ResizeLeft )
 	{
 		// TODO: Fix m_tco->setJournalling() consistency
 		m_tco->setJournalling( true );
 	}
+	else if( m_action == Split )
+	{
+		SampleTCO * leftTCO = dynamic_cast<SampleTCO*>( m_tco );
+
+		leftTCO->setMarkerEnabled( false );
```
BaraMGB: You forgot to test leftTco after dynamic_cast.

src/core/Track.cpp
```diff
@@ -1070,12 +1112,56 @@ void TrackContentObjectView::mouseReleaseEvent( QMouseEvent * me )
...
+
+		const int relativePixelPos = me->pos().x();
+		const float ppt = m_trackView->trackContainerView()->pixelsPerTact();
+		MidiTime splitPos = relativePixelPos * MidiTime::ticksPerTact() / ppt;
+		const float snapSize = gui->songEditor()->m_editor->getSnapSize();
+
+		//Essentially duplicate logic from knifeMarkerPos, but shift needs to be inverted for some reason
+		if ( me->modifiers() & Qt::ControlModifier
+		  || me->modifiers() & Qt::AltModifier    ){}
```
BaraMGB: This looks a little bit odd to me. Perhaps you find a better way for this logic?

src/core/Track.cpp
```diff
@@ -1070,12 +1112,56 @@ void TrackContentObjectView::mouseReleaseEvent( QMouseEvent * me )
...
+			else splitPos = rightPos;
+		}
+		else
+		{
+			splitPos = MidiTime(splitPos + m_initialTCOPos).quantize( snapSize ) - m_initialTCOPos;
+		}
+		splitPos += m_initialTCOPos;
+
+		//Don't do anything if we slid off the TCO
+		if ( splitPos <= m_initialTCOPos || splitPos >= m_initialTCOEnd ){}
```
BaraMGB: Why you don't ask `if (splitPos >= m_initialTCOPos || splitPos <= m_initialTCOEnd)` ?

src/core/Track.cpp
```diff
@@ -1218,6 +1304,43 @@ MidiTime TrackContentObjectView::draggedTCOPos( QMouseEvent * me )
...
+		MidiTime midiPos = markerPos * MidiTime::ticksPerTact() / ppt;
+		//2: Snap to the correct position, based on modifier keys
+		if ( me->modifiers() & Qt::ShiftModifier )
+		{	//If shift is held we quantize the length of the new left clip...
+			const MidiTime leftPos = midiPos.quantize( snapSize );
+			//...or right clip...
+			const MidiTime rightOff = m_tco->length() - midiPos;
+			const MidiTime rightPos = m_tco->length() - rightOff.quantize( snapSize );
+			//...whichever gives a position closer to the cursor
+			if ( abs(leftPos - midiPos) < abs(rightPos - midiPos) ) midiPos = leftPos;
```
BaraMGB: quote our coding convention:
> If you can comfortably fit the block on one line, then it's acceptable to format it like this:
> 
> 	if (m_sample > 0) { --m_sample; }
> 
> but note that the braces must always be included.

src/core/Track.cpp
```diff
@@ -1218,6 +1304,43 @@ MidiTime TrackContentObjectView::draggedTCOPos( QMouseEvent * me )
...
+			else midiPos = rightPos;
+		}
+		else {
```
BaraMGB: the brace belongs in the next line.

src/core/Track.cpp
```diff
@@ -1218,6 +1304,43 @@ MidiTime TrackContentObjectView::draggedTCOPos( QMouseEvent * me )
...
+			midiPos = MidiTime(midiPos + m_initialTCOPos).quantize( snapSize ) - m_initialTCOPos;
```
BaraMGB: spaces

BaraMGB (commented):
There are a lot of spaces in braces. As far as I know, we decided to use the current coding convention if we write new code. Perhaps you look into it. I think I have not barked at every single point.

include/SampleTrack.h
Spekular: I thought policy was to match surrounding code until an autoformatter fixes this stuff.

Spekular:
Resolved all comments from @BaraMGB except for the spaces in brackets. If someone can confirm that we prefer to mix old/new style over writing "old fashioned" code I can remove the spaces as well.

src/core/Track.cpp
```diff
@@ -1070,12 +1114,47 @@ void TrackContentObjectView::mouseReleaseEvent( QMouseEvent * me )
...
+		{
+			leftTCO->setMarkerEnabled( false );
+
+			const int relativePixelPos = me->pos().x();
+			const float ppt = m_trackView->trackContainerView()->pixelsPerTact();
+			MidiTime splitPos = relativePixelPos * MidiTime::ticksPerTact() / ppt;
+
+			if ( me->modifiers() & Qt::ControlModifier
+			  || me->modifiers() & Qt::AltModifier    ){}
+			else { splitPos = quantizeMarkerPos( splitPos, me->modifiers() & Qt::ShiftModifier ); }
```
BaraMGB: I would suggest:
```
if (!(me->modifiers() & Qt::ControlModifier || me->modifiers() & Qt::AltModifier))
{
    splitPos = quantizeMarkerPos(splitPos, me->modifiers() & Qt::ShiftModifier);
}
```

BaraMGB:
I fear I don't see that comment. Can you point me to that line?

BaraMGB:
Sorry, but I only see my suggestion on this if-statement. Maybe you accidentally started a review and this is still pending? What did you comment to this line?

Spekular:
To me that's harder to parse, and it's the same number of lines.
How do you feel about something like

```
const bool quantize = [ctrl or alt] ? false: true;
If (quantize) { [...] } 
```

Maybe that would be the clearest?

Spekular:
It seems I did start a review, yes. That should be fixed now, and I've also pasted my comment in the message above (before I figured out how to finish the review).

BaraMGB:
If you find a better way for avoiding an empty code block after an if-statement, please go on. But I don't see how a boolean can help here.

Spekular:
It's not there to avoid an empty code block (especially considering the code block is empty on purpose), I proposed the intermediate variable as a way to make the code more readable.

To me,` !(long && compound condition)` takes way too much effort to understand, having to parse the individual conditions, then their union, then the inverse of that.

If you don't like the boolean, what about a function that takes a mouse event and checks for alt or ctrl? For lack of a more concise name, something like `isUnquantizedModifierPressed`?

BaraMGB:
I'm not sure, perhaps I'm not the right person for reviewing this. I just wanted to help because I want to see this feature in lmms. Perhaps someone else can help here.

Spekular:
I think your review is fine, I'm just being picky/stubborn with this line. I don't really know how best to solve it myself.

src/core/Track.cpp
PhysSong: Using the de Morgan's law, @BaraMGB's solution can be written as:
```cpp
if (!(me->modifiers() & Qt::ControlModifier)
 && !(me->modifiers() & Qt::AltModifier))
{
    splitPos = quantizeMarkerPos(splitPos, me->modifiers() & Qt::ShiftModifier);
}
```
I'm not sure if it impropves readability though.

Veratil: Just to throw another idea out there...
```cpp
if (!(me->modifiers() & (Qt::ControlModifier | Qt::AltModifier)))
```

Spekular: In the end I went with creating a function for this, which shortens the condition and makes it clear what's being checked for.

Spekular:
Unless anyone has objections to the new unquantizedModHeld function, the only remaining review comment as of now is a style comment. Should I follow the surrounding code's style, or use our new style and have inconsistencies? Are there any other comments?

I believe at least one of the build failures are unrelated to my changes as circleci: linux.gcc is broken for all recent builds. The Travis fail is from a single job (also gcc) that fails on bundling/linking at the very end.

include/SampleTrack.h
PhysSong: > match surrounding code

`setIsPlaying()` doesn't have spaces in the parentheses, but `setSampleBuffer()` does. Neither way will match the surrounding code perfectly.

src/core/Track.cpp
```diff
@@ -1070,12 +1113,48 @@ void TrackContentObjectView::mouseReleaseEvent( QMouseEvent * me )
...
+				leftTCO->copy();
+				SampleTCO * rightTCO = new SampleTCO ( leftTCO->getTrack() );
+				rightTCO->paste();
+
+				leftTCO->changeLength( splitPos - m_initialTCOPos );
+
+				rightTCO->movePosition( splitPos );
+				rightTCO->changeLength( m_initialTCOEnd - splitPos );
+				rightTCO->setStartTimeOffset( leftTCO->startTimeOffset() - leftTCO->length() );
+			}
```
PhysSong: Can we move the TCO-type-specific splitting logic into its own class for each TCO type? I think it will make it easy to extend this feature to other TCO types.

PhysSong: Also, do you really need to use the `copy()` and `paste()` calls? It will silently change the content of the clipboard.

Spekular: > move the TCO-type-specific splitting logic

Good idea.

> do you really need to use the copy() and paste() calls

Probably not, but I have no idea how else to do it. I tried some things on my own but only ever managed to make either the left or the right clip. Since BaraMGB had solved it I just took that existing code and used it.

Either way I can't work on this again until November (at earliest). If you're able to push changes to this and want to apply the above in the meantime, feel free to do so.

include/SampleTrack.h
Spekular: The vast majority of the file uses spaces, but fair point.

Spekular:
> Can we move the TCO-type-specific splitting logic into its own class for each TCO type?

I gave this a shot today but couldn't really figure out how best to do it. Do you know specifically which logic you'd want me to move and to where? There's a lot of track classes (Track, TrackContentObject, TCOView, and subclasses for each track type, I think).

Spekular:
@PhysSong I meant to tag you in the above comment.

PhysSong:
I resolved merge conflicts and tested it a bit.
- If I split a clip while it's playing, LMMS hangs for a short period.
- I think there should be a custom cursor for the knife mode.

Spekular:
> I resolved merge conflicts and tested it a bit.

Thanks!

> * If I split a clip while it's playing, LMMS hangs for a short period.

I don't have time to investigate this for a while, and I suspect I may need(/want) help if the fix turns out to be non-trivial. I can probably manage to block the cut action for playing clips on my own, but that feels suboptimal.

> * I think there should be a custom cursor for the knife mode.

That would be good. @RebeccaDeField, would you happen to have time and be willing to do this?

RebeccaDeField:
> That would be good. @RebeccaDeField, would you happen to have time and be willing to do this?

I'm on it :) Do you happen to know what size this needs to be in/any other details before I get started? Otherwise, I'll do a bit of investigating.

Spekular:
According to the [Qt docs for QCursor](https://doc.qt.io/qt-5/qcursor.html#details), 32x32px is recommended. That's as much as I could find.

> Valid cursor sizes depend on the display hardware (or the underlying window system). We recommend using 32 x 32 cursors, because this size is supported on all platforms. Some platforms also support 16 x 16, 48 x 48, and 64 x 64 cursors.

Spekular:
Looks like the existing cursors are in the theme directory, just not suffixed (or prefixed) with "cursor" or "mouse". For example, "hand.png" https://github.com/LMMS/lmms/blob/master/data/themes/default/hand.png

RebeccaDeField:
@Spekular oddly enough, the standard cursor in the source files is 18x22. I wonder if this would be improved at that size as well? I really like the precision I get when video editing with adobes cursor, giving the line to indicate exactly where it will cut. 
edits-cover

But this will only work if the cut is made exactly at the left end of where the cursor clicks. If we can't do that and it makes the cut, say, in the center of the cursor, then I will just do a scissors icon.

Spekular:
The position can be configured to be exactly where we want it. I agree that it should cut on the left end, but a few things worth noting:
- The cursor won't always align with the cut (when snapping of some sort is enabled)
- The sample clip itself will always draw a line at the exact cut position

That is to say: I think it makes sense to have an "edge" so to speak on the left side of the cursor, but we don't need a full marker/line embedded in the icon.

src/core/Track.cpp
Spekular: > do you really need to use the copy() and paste() calls?

Fixed in latest commit.

Spekular:
@PhysSong I can't reproduce the hang. Does it still occur on the latest commt?

RebeccaDeField:
I have found that both the default and classic them use odd sizes for the cursor which are different from each other. I will start with the cursor icons tomorrow with 32x32 and we can scale down to 16x16 if it seems too large. Might want to update the default cursor to match so that the size doesn't significantly jump and if it is more widely supported that may be an overall improvement.

RebeccaDeField:
A few options from what I have mocked up so far. The first one needs some final tweaking to get it just right.
Screenshot from 2020-05-07 18-38-17

I lean towards the razor icon because I am used to it from other programs. Scissors might look more obvious. Thoughts?

Spekular:
Between these I prefer the razor as well, but in general I'm partial to a knife/scalpel metaphor. One of my concerns with scissors is their association with cut as in cut/copy/paste. I've added a screenshot to the PR description where you can see what the icon for the button looks like, apologies for not having this up before :(

RebeccaDeField:
@Spekular no problem, my mistake for not checking the current icon.

> One of my concerns with scissors is their association with cut as in cut/copy/paste.

Good point, I agree.

> Between these I prefer the razor as well, but in general, I'm partial to a knife/scalpel metaphor.

I like this concept. I can refine the existing icon as well. Should the cursor match this icon, just in a 32x32 size?

Spekular:
> I can refine the existing icon as well.

Feel free to, [here's the .svg file.](https://drive.google.com/open?id=19GPMtjsgIYzr1133c88Gc8Ng4RC36l6x)

> Should the cursor match this icon, just in a 32x32 size w/ border?

Matching the orientation might look odd; the button icon has to aim 45 degrees down and left to match the pencil, whereas cursors usually range between pointing up/left and pointing straight up. Of course, if it doesn't look weird or out of place with the other cursors, I suppose it's not an issue.

Otherwise I think it's probably a good idea for them to match. I assume you're referring to the black outline/stroke around most cursors when you say "border"? My first thought was a square frame around the icon so I'm double checking, haha.

Zoomed in screenshot highlighting cursor outline

RebeccaDeField:
> Matching the orientation might look odd; the button icon has to aim 45 degrees down and left to match the pencil, whereas cursors usually range between pointing up/left and pointing straight up. Of course, if it doesn't look weird or out of place with the other cursors, I suppose it's not an issue.

I think this is a great idea.

> Otherwise I think it's probably a good idea for them to match. I assume you're referring to the black outline/stroke around most cursors when you say "border"?

Yep :+1: 

Okay, I have created a slightly tweaked version for hopefully added crispness (.5 aligned slanted shapes tend to stairstep better) and a cursor version. For the cursor, I took out the slant on the tip of the blade so that it's a little more precise and matching a regular cursor expectation. I can leave the icon version with the slant as it does help metaphorically. Or we can make this straight as well for consistency.

Screenshot from 2020-05-14 16-05-09

Screenshot from 2020-05-14 16-05-18

@Spekular

Spekular:
Looks awesome! I'm a fan of straight cursor + slanted button, so let's go with that unless you disagree? Once I have .pngs for the cursor and tweaked button I'll try to get the cursor implemented as soon as I can. (I might wait for #5489, though)

RebeccaDeField:
It's a little tricky to make these crisp due to the number of slanted lines. I'm attaching the best results for the icon I could get. Slanted lines should translate better at half pixel but I don't see much of a difference in this case.

[knifeicons.zip](https://github.com/LMMS/lmms/files/4660021/knifeicons.zip)

If the icon master himself (@Umcaruje) has time to look at this and notices any way to crisp it up any further...

But otherwise, let's just make sure to use a cairo export.

PhysSong:
LMMS sometimes crashes on undoing cuts.
https://github.com/LMMS/lmms/blob/5cee7b234a03d327d1c74c016761eb15c9dea6f5/src/tracks/SampleTrack.cpp#L646-L651
It seems like `getTCO(i)` sometimes return null due to some multithread-unsafe code, and we don't check the return value.

The hang was fixed by not using `copy()` and `paste()`, but it caused an issue. `SampleTCO` and `SamplePlayHandle` assumes that no two TCOs share the same `SampleBuffer` instance, but your code breaks the assumption now. I think the best solution is to copy the `SampleBuffer` without reloading the audio file, but it looks quite tricky with the current design of `SampleBuffer`.

Spekular:
> It seems like `getTCO(i)` sometimes return null due to some multithread-unsafe code, and we don't check the return value

Does this multithread unsafeness need to be fixed in the areas I've touched, or elsewhere? Also, that loop never references `i` except to get TCOs, would replacing it with a range-based for loop help at all?

> The hang was fixed by not using `copy()` and `paste()`, but it caused an issue. `SampleTCO` and `SamplePlayHandle` assumes that no two TCOs share the same `SampleBuffer` instance, but your code breaks the assumption now. This makes the first TCO doesn't stop playing when it should.
> I think the best solution is to copy the sample buffer without reloading the audio file, but it looks quite tricky with the current design of it.

Ok, so how do we move forward? I believe I tried using a copy constructor, but the compiler complained about it not existing even though a default one should exist.

PhysSong:
> Does this multithread unsafeness need to be fixed in the areas I've touched, or elsewhere?

I think `Track::loadSettings()` is deleting TCOs in a thread-unsafe manner, I'll fix that on `master`.
> Ok, so how do we move forward? I believe I tried using a copy constructor, but the compiler complained about it not existing even though a default one should exist.

There's no default copy constructor because the copy constructor of `QReadWriteLock` is deleted. Even if there's one, you shouldn't use it because it won't copy the memory block for samples.
You may use `saveSettings()` and `loadSettings()` from `SerializingObject`, but that will reintroduce the short hang. I'm okay with it because it's better than having a weird bug.

Spekular:
@PhysSong how does this look?

https://github.com/LMMS/lmms/compare/master...Spekular:ROT

Still seems to work from a quick test, the first TCO stops playing when it should. If this is good then I should also add a copy assignment to SampleBuffer and SampleTCO.

PhysSong:
That looks fine, but I think this will conflict with #4994. I want to hear @Reflexe's opinion for the final decision, but I'd say go ahead for now.

PR #4999, opened 2019-05-31
Further C++11 inheritance updates.

devnexen:


include/AudioAlsa.h
```diff
@@ -80,10 +80,10 @@ class AudioAlsa : public QThread, public AudioDevice
 	static DeviceInfoCollection getAvailableDevices();
 
 private:
-	virtual void startProcessing();
-	virtual void stopProcessing();
-	virtual void applyQualitySettings();
-	virtual void run();
+	virtual void startProcessing() override;
```
JohannesLorenz: I think the `virtual` keyword is redundant if you use `override`, since `override` implicits `virtual`. I'd vote for removing all `virtual` keywords whenever you `override`, this makes the class way cleaner and faster to read.

include/AudioSdl.h
```diff
@@ -62,7 +62,7 @@ class AudioSdl : public AudioDevice
 		setupWidget( QWidget * _parent );
 		virtual ~setupWidget();
```
JohannesLorenz: If this dtor is overriding, it should be marked `override` as well (some compilers complain if not all functions are marked `override`).

include/AudioAlsa.h
JohannesLorenz: Before you change all this manually, consider using something like `sed 's/virtual \(.*override;\)/\1/' -i include/*`. Let me know if you need help.

JohannesLorenz:
BUMP. Are you planning to fix the comments @devnexen ?

devnexen:
Sorry I did not notice them ... doing.

devnexen:
@JohannesLorenz hopes it s good :-)

JohannesLorenz:
@devnexen There is still a lot missing, I still see

* many functions with both `virtual` and `override` keywords. The `virtual` should be removed in that case
* many virtual destructors that still miss the `override` keyword at all.

JohannesLorenz:
@devnexen Is this ready for the next review? Also, can you please fix the conflicts?

devnexen:
Yes this is now :)

lukas-w:
I added the missing `override` keywords and removed redundant `virtual` keywords via 3b6707ff67fcb6e336dc3b5b4cc77a963f442199. Merging.

Issue #4062, opened 2017-12-21
Windows does not display splash screen


tresf:
As reported on [Discord](https://lmms.io/chat) by @StakeoutPunch
* Reproduced on , 1.2.0-RC5 32-bit, , 1.2.0-RC5 64-bit
* Reproduced on Windows 7 and Windows 10

tresf:
The splash that ships with the classic theme is working, default is not working:

* data/themes/**classic**/splash.png: (681x573, 24-bit RGB, non-interlaced, 71.5%)`
* data/themes/**default**/splash.png:  (681x573, **32-bit RGB+alpha**, non-interlaced, 92.2%)`

PR #5002, opened 2019-06-03
Updated splash screen

RebeccaDeField:
Updated the splash screen to match the current logo/branding/website that @Umcaruje worked on.

**Old**
splash

**New**
splash

It's been a while since I've used git or github so if I've done this incorrectly just let me know. I removed the alpha layer but the splash still needs to be reviewed to make sure we don't encounter the same issue here #4062.


PhysSong:
I confirmed that the image doesn't have an alpha channel. :+1:

RebeccaDeField:
@PhysSong So is this good to merge, then?

PhysSong:
I can't comment on the design itself, so I think it's fine to merge if you think it's okay.

zonkmachine:
With this splash.png I get `libpng warning: iCCP: CRC error` on startup.
I'm on LinuxMint/libpng12.

PhysSong:
Same here with Ubuntu 18.04/`libpng16`.

zonkmachine:
$ pngcheck ../data/themes/default/splash.png 
../data/themes/default/splash.png  CRC error in chunk iCCP (computed a0032057, expected a4b05ab4)

PhysSong:
Possibly related: https://gitlab.gnome.org/GNOME/gimp/issues/2111
If I manually change the checksum in the image file using hex editor, the bug is gone.
Here's the image file with the checksum fixed:
splash
I can commit the change now.

PhysSong:
@RebeccaDeField If you used GIMP to export this image, could you try unchecking `Save Color Profile` on export?

RebeccaDeField:
@PhysSong Yes. Is there a way for you to test this against the file you produced to make sure this is solving the problem? 

splash

PhysSong:
> @PhysSong Yes. Is there a way for you to test this against the file you produced to make sure this is solving the problem?

I checked this file with both `pngcheck` and GIMP, and it looks fine. The error message is also gone. :tada:

Issue #4974, opened 2019-05-15
Song editor: strange behavior when selecting large area


Gabrielxd195:
I have many tracks and I want to select them all to move them, but when I activate the '' select mode '' and try to select all the tco using the mouse wheel to scroll down, the first tracks that you select remain unscreened while the bottom ones . something like:
![ezgif com-video-to-gif](https://user-images.githubusercontent.com/26127693/57801492-d1dc1b80-7721-11e9-90d3-3821b7d72fa9.gif)
As you can see, when I make the selection box and when I move with the mouse wheel, the box moves and the tco that remain outside the box are not selected and I am forced to select by parts. previously mentioned some errors and improvements in #4602 in relation to the select mode but this error had not taken into account.

Spekular:
As a workaround for this, you can use shift+drag on the timeline at the top
of the song editor to select every track's contents between two points.

Gabrielxd195:
> As a workaround for this, you can use shift+drag on the timeline at the top of the song editor to select every track's contents between two points.

This for me would be a temporary solution, since I can put the zoon to 12.5% and select it in this way (I had forgotten it) but the problem still persists, when the tco go out of the selection box when you scroll with the mouse wheel , the tco are left unselected and you have to select by parts or as you say with shift+drag. but anyway, thanks.

musikBear:
The time-line-method unfortunately does not allow sh+drag for a additional selection. but sure safe method to move everything now, is to place playhead infont of the wanted split-position and the 'shove' everything upstream playhead, with sh+INSERT

BaraMGB:
In master branch of lmms you can select all by <CTRL> + <A>.

musikBear:
> In master branch of lmms you can select all by CTRL + A.

Not only select. These works too
* fractured-selectons
* move selected
* copy selected
* delete selected

It would be really good, if these found their way into RC9/ 1.1.2
I have enjoyed them a lot in my RC4-Master: 4.6493 
Afair i have not had one single issue and no crashes at all. Looks absolutely solid

BaraMGB:
Since this is fixed in master this issue can be closed.

SecondFlight:
Is the broken selection behavior fixed in master? I would consider ctrl+a to be a workaround, but the behavior shown in the GIF is definitely a bug.

BaraMGB:
Sorry, my bad. I'll have a deeper look at this.

BaraMGB:
I think, I know, what's going on.

[...]

PR #5003, opened 2019-06-04
Rubberband fix for selecting large area in Songeditor

BaraMGB:
fixes https://github.com/LMMS/lmms/issues/4974

scrolling and zooming while selecting an area works now. But with 180 additions and 70 deletions this is a big bug fix. I don't know if this should be merged so close before releasing 1.2.

### DIFF
--- a/include/SongEditor.h
+++ b/include/SongEditor.h
@@ -142,6 +142,8 @@ private:
 	QPoint m_origin;
 	QPoint m_scrollPos;
 	QPoint m_mousePos;
+	int m_rubberBandStartTrackview;
+	MidiTime m_rubberbandStartMidipos;
 
 	friend class SongEditorWindow;
 
--- a/include/SongEditor.h
+++ b/include/SongEditor.h
@@ -75,6 +75,7 @@ public:
 
 public slots:
 	void scrolled( int new_pos );
+	void updateRubberband();
 
 	void setEditMode( EditMode mode );
 	void setEditModeDraw();
@@ -85,6 +86,9 @@ public slots:
 
 protected:
 	virtual void closeEvent( QCloseEvent * ce );
+	virtual void mousePressEvent( QMouseEvent * _me );
+	virtual void mouseMoveEvent( QMouseEvent * _me );
+	virtual void mouseReleaseEvent( QMouseEvent * _me );
 
 private slots:
 	void setHighQuality( bool );
@@ -135,6 +139,8 @@ private:
 	bool m_smoothScroll;
 
 	EditMode m_mode;
+	QPoint m_origin;
+	QPoint m_mousePos;
 
 	friend class SongEditorWindow;
 
--- a/include/SongEditor.h
+++ b/include/SongEditor.h
@@ -145,6 +145,7 @@ private:
 	int m_rubberBandStartTrackview;
 	MidiTime m_rubberbandStartMidipos;
 	int m_currentZoomingValue;
+	int m_trackHeadWidth;
 
 	friend class SongEditorWindow;
 
--- a/include/SongEditor.h
+++ b/include/SongEditor.h
@@ -75,6 +75,8 @@ public:
 
 public slots:
 	void scrolled( int new_pos );
+	void selectRegionFromPixels(int xStart, int xEnd);
+	void stopSelectRegion();
 	void updateRubberband();
 
 	void setEditMode( EditMode mode );
@@ -149,6 +151,7 @@ private:
 	MidiTime m_rubberbandStartMidipos;
 	int m_currentZoomingValue;
 	int m_trackHeadWidth;
+	bool m_selectRegion;
 
 	friend class SongEditorWindow;
 
--- a/include/SongEditor.h
+++ b/include/SongEditor.h
@@ -113,7 +113,8 @@ private:
 
 	virtual bool allowRubberband() const;
 
-	int trackNumber(int yPos);
+	int trackIndexFromSelectionPoint(int yPos);
+	int indexOfTrackView(const TrackView* tv);
 
 
 	Song * m_song;
--- a/include/SongEditor.h
+++ b/include/SongEditor.h
@@ -140,6 +140,7 @@ private:
 
 	EditMode m_mode;
 	QPoint m_origin;
+	QPoint m_scrollPos;
 	QPoint m_mousePos;
 
 	friend class SongEditorWindow;
--- a/include/SongEditor.h
+++ b/include/SongEditor.h
@@ -113,6 +113,8 @@ private:
 
 	virtual bool allowRubberband() const;
 
+	int trackNumber(int yPos);
+
 
 	Song * m_song;
 
--- a/include/SongEditor.h
+++ b/include/SongEditor.h
@@ -144,6 +144,7 @@ private:
 	QPoint m_mousePos;
 	int m_rubberBandStartTrackview;
 	MidiTime m_rubberbandStartMidipos;
+	int m_currentZoomingValue;
 
 	friend class SongEditorWindow;
 
--- a/include/SongEditor.h
+++ b/include/SongEditor.h
@@ -88,9 +88,9 @@ public slots:
 
 protected:
 	virtual void closeEvent( QCloseEvent * ce );
-	virtual void mousePressEvent( QMouseEvent * _me );
-	virtual void mouseMoveEvent( QMouseEvent * _me );
-	virtual void mouseReleaseEvent( QMouseEvent * _me );
+	virtual void mousePressEvent(QMouseEvent * me);
+	virtual void mouseMoveEvent(QMouseEvent * me);
+	virtual void mouseReleaseEvent(QMouseEvent * me);
 
 private slots:
 	void setHighQuality( bool );
--- a/include/Track.h
+++ b/include/Track.h
@@ -210,6 +210,12 @@ public:
 	{
 		return m_tco;
 	}
+
+	inline TrackView * getTrackView()
+	{
+		return m_trackView;
+	}
+
 	// qproperty access func
 	QColor mutedColor() const;
 	QColor mutedBackgroundColor() const;
@@ -229,11 +235,7 @@ public:
 	// access needsUpdate member variable
 	bool needsUpdate();
 	void setNeedsUpdate( bool b );
-	
-	inline TrackView * getTrackView()
-	{
-		return m_trackView;
-	}
+
 public slots:
 	virtual bool close();
 	void cut();
--- a/include/Track.h
+++ b/include/Track.h
@@ -230,6 +230,10 @@ public:
 	bool needsUpdate();
 	void setNeedsUpdate( bool b );
 	
+	inline TrackView * getTrackView()
+	{
+		return m_trackView;
+	}
 public slots:
 	virtual bool close();
 	void cut();
@@ -256,10 +260,6 @@ protected:
 
 	float pixelsPerTact();
 
-	inline TrackView * getTrackView()
-	{
-		return m_trackView;
-	}
 
 	DataFile createTCODataFiles(const QVector<TrackContentObjectView *> & tcos) const;
 
--- a/src/gui/editors/SongEditor.cpp
+++ b/src/gui/editors/SongEditor.cpp
@@ -511,13 +511,13 @@ void SongEditor::closeEvent( QCloseEvent * ce )
 
 
 
-void SongEditor::mousePressEvent(QMouseEvent *_me)
+void SongEditor::mousePressEvent(QMouseEvent *me)
 {
 	if (allowRubberband())
 	{
 		//we save the position of scrollbars, mouse position and zooming level
 		m_scrollPos = QPoint(m_leftRightScroll->value(), contentWidget()->verticalScrollBar()->value());
-		m_origin = contentWidget()->mapFromParent(QPoint(_me->pos().x(), _me->pos().y()));
+		m_origin = contentWidget()->mapFromParent(QPoint(me->pos().x(), me->pos().y()));
 		m_currentZoomingValue = zoomingModel()->value();
 
 		//paint the rubberband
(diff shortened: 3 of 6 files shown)


PhysSong:
> I don't know if this should be merged so close before releasing 1.2.

If you want to test more, we can include this in 1.2.1 instead of 1.2.0.

src/gui/editors/SongEditor.cpp
```diff
@@ -293,6 +300,80 @@ void SongEditor::scrolled( int new_pos )
...
+		//the miditime the mouse is hover
+		MidiTime rubberbandMidipos = MidiTime((m_mousePos.x() - m_trackHeadWidth) / pixelsPerTact()
+											  * MidiTime::ticksPerTact()) + m_currentPosition ;
+		//collect all Tcos
+		QVector<selectableObject *> so;
+		QList<selectableObject *> l = findChildren<selectableObject *>();
+
+		for (QList<selectableObject *>::iterator it = l.begin(); it != l.end(); ++it)
+		{
+			so.push_back( *it );
```
PhysSong: No need to waste resources here, you can just use the list directly.

src/gui/editors/SongEditor.cpp
```diff
@@ -293,6 +300,80 @@ void SongEditor::scrolled( int new_pos )
...
+		}
+		//are tcos in the rect of selection?
+		for (QVector<selectableObject *>::iterator it = so.begin(); it != so.end(); ++it)
```
PhysSong: It's a good chance to switch to the range-based loop.

src/gui/editors/SongEditor.cpp
```diff
@@ -293,6 +300,80 @@ void SongEditor::scrolled( int new_pos )
...
+		rubberBand()->setGeometry(QRect( origin,
+										 contentWidget()->mapFromParent(QPoint(m_mousePos.x(), m_mousePos.y()))
+										).normalized());
+
+		//the index of the TrackView the mouse is hover
+		const TrackView * tv = trackViewAt(m_mousePos.y() - m_timeLine->height());
+		int rubberBandTrackview = 0;
+		if (tv)
+		{
+			for (QList<TrackView *>::const_iterator it = trackViews().begin(); it != trackViews().end(); ++it)
```
PhysSong: A range-based loop will be better.

PhysSong (commented):
This is the first part of my review. I will review this PR again in a few days.

BaraMGB:
Thank you for your time.

src/gui/editors/SongEditor.cpp
```diff
@@ -293,6 +300,75 @@ void SongEditor::scrolled( int new_pos )
...
+										 contentWidget()->mapFromParent(QPoint(m_mousePos.x(), m_mousePos.y()))
+										).normalized());
+
+		//the index of the TrackView the mouse is hover
+		const TrackView * tv = trackViewAt(m_mousePos.y() - m_timeLine->height());
+		int rubberBandTrackview = 0;
+		if (tv)
+		{
+			for (auto it : trackViews())
+			{
```
Reflexe: Can't understand the reason behind this loop; why won't you use `indexOf`?

src/gui/editors/SongEditor.cpp
```diff
@@ -293,6 +300,75 @@ void SongEditor::scrolled( int new_pos )
...
+											  * MidiTime::ticksPerTact()) + m_currentPosition ;
+		//collect all Tcos
+		auto l = findChildren<selectableObject *>();
+
+		//are tcos in the rect of selection?
+		for (auto &it : l)
+		{
+			TrackContentObjectView * tco = dynamic_cast<TrackContentObjectView*>(it);
+
+			if (trackViews().indexOf(tco->getTrackView())     >= qMin(m_rubberBandStartTrackview, rubberBandTrackview)
```
Reflexe: What's the meaning of this statement? `X >= Y` and then `X <= Y`? Am I missing something? 
Also, check for `nullptr` after `dynamic_cast`.

src/gui/editors/SongEditor.cpp
```diff
@@ -404,7 +480,68 @@ void SongEditor::closeEvent( QCloseEvent * ce )
...
+		m_currentZoomingValue = zoomingModel()->value();
+		//paint the rubberband
+		rubberBand()->setEnabled(true);
+		rubberBand()->setGeometry(QRect( m_origin, QSize()));
+		rubberBand()->show();
+
+		//the trackView(index) and the miditime where the mouse has clicked
+		const TrackView * tv = trackViewAt(_me->pos().y() - m_timeLine->height());
+		m_rubberBandStartTrackview = 0;
+		if (tv)
```
Reflexe: A duplicate of line 350. Please extract it into something like `updateRubberbandSelectionForMousePos(pos)`.

Reflexe (changes requested):

src/gui/editors/SongEditor.cpp
BaraMGB: You can make the selection from top to bottom or from bottom to top. The right side of the if-statement is the exact opposite in the second if-statement. Is it in the first statement `m_rubberBandStartTrackview` so it is mandatory in the second if-statement `rubberbandTrackView` and vice versa. So you can test whether the selection is made from below or from above.

But maybe I missed something? It gave me a real headache. Perhaps there is a simpler way?

include/Track.h
```diff
@@ -210,6 +210,12 @@ class TrackContentObjectView : public selectableObject, public ModelView
 	{
 		return m_tco;
 	}
+
```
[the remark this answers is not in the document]
BaraMGB: It was protected but I need this outside of the class. Can I avoid this move?

src/gui/editors/SongEditor.cpp
Reflexe: I understand. We could add `indexOfTrackView(const TrackView)` with `std::distance(std::find(m_trackViews.begin(), m_trackViews.end(), tv), m_trackViews.begin()` instead.

src/gui/editors/SongEditor.cpp
Reflexe: I don't understand: the condition `(X >= Y && X <= Y)` will be true only if `X==Y`.

Reflexe: My bad; missed something.

Reflexe (approved):

BaraMGB:
@Reflexe Thanks a lot for reviewing this.
@PhysSong Do you have any objection about this PR?

src/gui/editors/SongEditor.cpp
```diff
@@ -239,6 +243,9 @@ SongEditor::SongEditor( Song * song ) :
 					this, SLOT( scrolled( int ) ) );
 	connect( m_song, SIGNAL( lengthChanged( int ) ),
 			this, SLOT( updateScrollBar( int ) ) );
+	connect( m_leftRightScroll, SIGNAL(valueChanged(int)),this, SLOT(updateRubberband()));
+	connect( contentWidget()->verticalScrollBar(), SIGNAL(valueChanged(int)),this, SLOT(updateRubberband()));
```
PhysSong: There's whitespace after the first opening parenthesis, but there isn't one before the last closing parenthesis. It looks a bit inconsistent to me.

src/gui/editors/SongEditor.cpp
```diff
@@ -293,6 +300,57 @@ void SongEditor::scrolled( int new_pos )
...
+
+		//take care of the scrollbar position
+		int hs = (m_leftRightScroll->value() - m_scrollPos.x()) * pixelsPerTact();
+		int vs = contentWidget()->verticalScrollBar()->value() - m_scrollPos.y();
+
+		//the adjusted origin point
+		QPoint origin = QPoint(qMax(originX - hs, m_trackHeadWidth), m_origin.y() - vs);
+
+		//paint the rubber band rect
+		rubberBand()->setGeometry(QRect( origin,
```
PhysSong: Similar here.

src/gui/editors/SongEditor.cpp
```diff
@@ -293,6 +300,57 @@ void SongEditor::scrolled( int new_pos )
...
+
+		//the miditime the mouse is hover
+		MidiTime rubberbandMidipos = MidiTime((m_mousePos.x() - m_trackHeadWidth)
+											  / static_cast<int>(pixelsPerTact())
+											  * MidiTime::ticksPerTact()) + m_currentPosition ;
+		//are tcos in the rect of selection?
+		for (auto &it : findChildren<selectableObject *>())
+		{
+			TrackContentObjectView * tco = dynamic_cast<TrackContentObjectView*>(it);
+			if(tco)
```
PhysSong:
```suggestion
			if (tco)
```

src/gui/editors/SongEditor.cpp
```diff
@@ -404,7 +462,55 @@ void SongEditor::closeEvent( QCloseEvent * ce )
...
+	{
+		//we save the position of both scrollbars into a QPoint
+		m_scrollPos = QPoint(m_leftRightScroll->value(), contentWidget()->verticalScrollBar()->value());
+		//and the mouse position
+		m_origin = contentWidget()->mapFromParent(QPoint(_me->pos().x(), _me->pos().y()));
+		//and the zooming
+		m_currentZoomingValue = zoomingModel()->value();
+		//paint the rubberband
+		rubberBand()->setEnabled(true);
+		rubberBand()->setGeometry(QRect( m_origin, QSize()));
```
PhysSong: Here as well.

src/gui/editors/SongEditor.cpp
```diff
@@ -404,7 +462,55 @@ void SongEditor::closeEvent( QCloseEvent * ce )
...
+		rubberBand()->show();
+
+		//the trackView(index) and the miditime where the mouse has clicked
+		m_rubberBandStartTrackview = trackIndexFromSelectionPoint(_me->y());
+
+		m_rubberbandStartMidipos = MidiTime((_me->x() - m_trackHeadWidth)
+											/ pixelsPerTact() * MidiTime::ticksPerTact())
+											+ m_currentPosition;
+	}
+	QWidget::mousePressEvent( _me );
```
PhysSong: Removing whitespaces in parentheses in the new code will be better.

src/gui/editors/SongEditor.cpp
```diff
@@ -404,7 +462,55 @@ void SongEditor::closeEvent( QCloseEvent * ce )
...
+}
+
+
+
+
+void SongEditor::mouseReleaseEvent(QMouseEvent *_me)
+{
+	rubberBand()->hide();
+	rubberBand()->setEnabled( false );
+	QWidget::mouseReleaseEvent( _me );
```
PhysSong: Here as well.

PhysSong (commented):
Here are minor comments regarding whitespaces. I'm testing the functionality and will let you know if there are any issues or not.

PhysSong:
I found that lassoing with Shift+drag on the timeline is broken. @BaraMGB Could you look into it?

BaraMGB:
@PhysSong Thank you. I'll look into it. :+1:

src/gui/editors/SongEditor.cpp
```diff
@@ -80,15 +79,21 @@ SongEditor::SongEditor( Song * song ) :
 	m_zoomingModel(new ComboBoxModel()),
 	m_scrollBack( false ),
 	m_smoothScroll( ConfigManager::inst()->value( "ui", "smoothscroll" ).toInt() ),
-	m_mode(DrawMode)
+	m_mode(DrawMode),
+	m_origin(),
+	m_scrollPos(),
+	m_mousePos(),
+	m_rubberBandStartTrackview(),
+	m_rubberbandStartMidipos(),
```
PhysSong: I think initializing `int` values with explicit 0s will be better.

src/gui/editors/SongEditor.cpp
```diff
@@ -293,6 +302,98 @@ void SongEditor::scrolled( int new_pos )
...
+
+		//we save the position of scrollbars, mouse position and zooming level
+		m_origin = QPoint(xStart, 0);
+		m_scrollPos = QPoint(m_leftRightScroll->value(), contentWidget()->verticalScrollBar()->value());
+		m_currentZoomingValue = zoomingModel()->value();
+
+		//calculate the song position where the mouse was clicked
+		m_rubberbandStartMidipos = MidiTime((xStart - m_trackHeadWidth)
+											/ pixelsPerTact() * MidiTime::ticksPerTact())
+											+ m_currentPosition;
```
PhysSong: You should set `m_rubberBandStartTrackview` to 0 as well. If I select some TCOs in track views and then from the timeline, some TCOs are not selected although the rubberband displays correctly.

BaraMGB: fixed it.

BaraMGB:
@PhysSong can this be merged or do you want to test it more?

include/SongEditor.h
```diff
@@ -85,6 +88,9 @@ public slots:
 
 protected:
 	virtual void closeEvent( QCloseEvent * ce );
+	virtual void mousePressEvent( QMouseEvent * _me );
+	virtual void mouseMoveEvent( QMouseEvent * _me );
+	virtual void mouseReleaseEvent( QMouseEvent * _me );
```
PhysSong: No underscores anymore...
```
• Function parameters are not prefixed with '_' anymore.
```
https://github.com/LMMS/lmms/wiki/Coding-conventions

src/gui/editors/SongEditor.cpp
```diff
@@ -404,7 +506,53 @@ void SongEditor::closeEvent( QCloseEvent * ce )
 		hide();
 	}
 	ce->ignore();
- }
+}
+
+
+
+
+void SongEditor::mousePressEvent(QMouseEvent *_me)
```
PhysSong: The same here and below.

PhysSong:
The functionality looks okay, but I think we have some time to test this more because this PR targets the stable branch.

BaraMGB:
@PhysSong Do you think, this should go into 1.2.1?

PhysSong:
I think it's okay to merge this now. As far as I tested, this won't bring any serious regressions.

Issue #4996, opened 2019-05-28
Export tracks :Missing metadata bug


T0NIT0RMX:
Hello,
We found an issue when using the "export tracks" option in LMMS (RC8)

**To reproduce :** 
- Remove every existing tracks from a new project
- Load 2 3xOsc and put some notes on them
- Export tracks so you have : 1_TripleOscillator.wav and 2_TripleOscillator.wav

Issue :
![image](https://user-images.githubusercontent.com/12613367/58460386-d3e4a800-812d-11e9-939b-0925a209c5e7.png)
As you can see on the screenshot, only the first track has correct metadata, especially the "length" field ("Longeur" on my screenshot), same for the "bitrate" ("vitesse de transmission" on my screenshot)

**Consequence** : 
Even if some audio players don't care about the missing "length" field (like VLC, Audacity), some other audio players seems to pay a lot of importance to it and block the file (like the native Win10 player, FL Studio, Ableton) if the length field is missing (or maybe = 00:00:00)

This is really blocking when we try to export stems for collaborations with guys using other DAW like FL or Ableton.

Thanks in advance for your help,

T0NIT0 RMX

PhysSong:
What happens if you export to MP3 or OGG?

T0NIT0RMX:
We tried with .wav, .mp3, .ogg

.wav & .mp3
- 1st track : OK
- other tracks : no length field, no bitrate, can't be read by some audio players

.ogg
- 1st track : OK,
- other tracks : size = 0 bytes, but can be read by the Win10 player

Hope it helps, don't hesitate if you need more details, I'll reply ASAP

JohannesLorenz:
@T0NIT0RMX Can you easily check if it's on master, too?

Suspected file is src/core/RenderManager.cpp .

Might be a candidate to get fixed before 1.2.0 ...

JohannesLorenz:
Confirming on Linux on branches master and stable-1.2.

E.g. mpv says for the 2nd track: `[ffmpeg/demuxer] wav: Estimating duration from bitrate, this may be inaccurate` .

JohannesLorenz:
OK, all `m_si.xxx` variables in `AudioFileWave::startEncoding()` have the same values when writing the two tracks. Must be a different error.

JohannesLorenz:
[...]

PR #5005, opened 2019-06-04
Fix metadata when exporting multiple tracks (#4996)

JohannesLorenz:


JohannesLorenz:
I'm adding the 1.2.0 label as a reminder. If you want to merge it without review, that's OK. If there's time for a review before, even better.

src/core/Mixer.cpp
```diff
@@ -575,21 +575,35 @@ void Mixer::changeQuality( const struct qualitySettings & _qs )
 
 
 
-void Mixer::setAudioDevice( AudioDevice * _dev,
-			    bool startNow )
+void Mixer::doSetAudioDevice( AudioDevice * _dev )
 {
-	stopProcessing();
+	// TODO: Use shared_ptr here in the future.
```
PhysSong: Why don't you do this now? Also, `QSharedPointer` will work as well as `std::shared_pointer`.

JohannesLorenz: Fixed in fc5cb9c

JohannesLorenz:
Tested, also tested jack. It all seems to work. Let me know if you want to change something, or if it can be merge.

PhysSong:
* I think you shouldn't include translation updates here.
* Windows builds are failing because C++11 is disabled for LADSPA: https://github.com/LMMS/lmms/blob/47d6388182e9105c0cf7e87d494dd2db593350c5/plugins/LadspaEffect/CMakeLists.txt#L5-L8
You can find the pretty old CI errors in https://travis-ci.org/LMMS/lmms/jobs/253835510#L1277-L1289, I'm not sure if we can re-enable it now. If `std::shared_pointer` doesn't work, we should move to `QSharedPointer` or just revert relevant commits.

JohannesLorenz:
Do you propose to not add translation updates to this branch or not to stable-1.2?

About the Windows error: Has this been fixed meanwhile? Maybe the CI now works with C++11? If not, I'd say let's move my shared_ptr fix aside and merge the version from yesterday?

I did try QSharedPointer, but the Qt4 version did not have a proper reset function, I didn't like the workarounds.

PhysSong:
> Do you propose to not add translation updates to this branch or not to stable-1.2?

Merging that commit into `stable-1.2` is okay, but not by this pull request.

> About the Windows error: Has this been fixed meanwhile?

This patch works(MinGW requires `-std=gnu++0x` for `M_PI`):
```diff
diff --git a/plugins/LadspaEffect/CMakeLists.txt b/plugins/LadspaEffect/CMakeLists.txt
index 029cd9168..1c500673c 100644
--- a/plugins/LadspaEffect/CMakeLists.txt
+++ b/plugins/LadspaEffect/CMakeLists.txt
@@ -2,8 +2,10 @@ INCLUDE(BuildPlugin)
 
 # Disable C++11
 REMOVE_DEFINITIONS(-std=c++0x)
-# Enable C++11 for CXXFLAGS only and not for Windows
-IF(NOT LMMS_BUILD_WIN32)
+# Enable C++11 for CXXFLAGS only
+IF("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+	SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++0x")
+ELSE()
 	SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++0x")
 ENDIF()
```
Anyway, please feel free to defer changes related to shared pointers to `master` if you want to do so.

JohannesLorenz:
Would you prefer to apply that C++11 fix to Ladspa and then release with `shared_ptr`, or only apply `shared_ptr` on `master`? The first option may be a bit cleaner, but it's risky to do that on `stable-1.2`, right?

PhysSong:
If you think that change is unnecessary, you can remove the shared pointer part from here.

JohannesLorenz:
OK, I did an undo of the German translation and and undo of the `shared_ptr`. I'll merge this now and later add the German translation to `stable-1.2` and the `shared_ptr` to `master`.

PR #5019, opened 2019-06-08
Fix color profile error

RebeccaDeField:
Looks like gimp was exporting the image with an ICC profile in the export settings after removing the alpha layer. This should fix problems described here #5002 

@PhysSong @zonkmachine


zonkmachine:
Fix confirmed, thanks!

PR #5040, opened 2019-06-15
 Replace tLimit() with qBound() and remove templates.h

Sawuare:
Because the only use of `templates.h` is its `tLimit()`, which is essentially [qBound()](https://doc.qt.io/qt-5/qtglobal.html#qBound) but with a different order of parameters, and the header is included, but not used, in many files.


JohannesLorenz:
The PR is all OK. It even increases type safety, because `tLimit()` allowed to pass by value, causing implicit casts. `qBound()` only allows to pass by reference.

Time to merge.

JohannesLorenz:
Do you prefer this as 2 commits or as 1 commit on master?

Sawuare:
1 Commit, and thank you for reviewing.

Issue #4810, opened 2019-02-03
Indentation (tabs or spaces)


JohannesLorenz:
Looking at the current code base, indentation is made using tabs, which also conforms to our [coding conventions](https://github.com/LMMS/lmms/wiki/Coding-conventions). The conventions, however, don't name the length for each tab. You can usually check line breaks or constructor initializer lists to see what tab length was used in the past.

* src/core/Instrument.cpp: 4
* plugins/LadspaEffect/LadspaSubPluginFeatures.h: 4
* src/core/ProjectRenderer.cpp: 8
* plugins/GigPlayer/GigPlayer.cpp: 8

The most files seem to use 4, though.

Note: We can do automatic conversions using #4690 .

Please vote (multiple votes allowed):
* :tada: Use tab lenght of 4
* :heart: Use tab length of 8
* :rocket: Change tabs into spaces (length 4)
* :-1: Other option, specify

JohannesLorenz:
I'm for staying with tabs, and use a length of 4:

* this seems to be used in the most files
* this also allows longer code lines then a length of 8 (especially if you have many if/switch/for etc).

zonkmachine:
> The conventions, however, don't name the length for each tab.

I believe the convention used to specify a tab length of 8.

JohannesLorenz:
Note: Github by default uses a tab size of 8. We can fix this adding an `.editorconfig` to LMMS, like [suggested on SO](https://stackoverflow.com/a/33831598).

PhysSong:
There is another way, indenting by tabs and aligning by spaces. It's a bit complicated, but one can change the tab size without breaking alignments.

Spekular:
I assume tabs + spaces for alignment is what's meant by the tab options.

justnope:
Are we still going to use .clang-format? (https://github.com/LMMS/lmms/pull/4690)

To be honest, I don't care whether it's tabs or spaces. I work a lot in upstream libraries and whatever you choose, I'll still going to have to edit my options to match the tabs/space flavour of that lib. I used to have an opinion but now I'm more leaning towards: "f it! let a formatter figure it out!" 

Which bring us back to .clang-format.

noahb01:
8-space tabs.

C++ is hard to read with small indents because of braces etc. So I prefer a width of 8.
I know most people don't agree with me, so I prefer tabs because then everyone can set whatever width they want.

tresf:
[...]

PR #5041, opened 2019-06-16
Add .editorconfig

JohannesLorenz:
This file tells the github online editor to use tabs of width 4 instead of 8 (as [discussed](https://github.com/LMMS/lmms/issues/4810)). Open [this](https://github.com/JohannesLorenz/github_test/blob/master/test.cpp) test file in your browser to see how it works (the width is 5 in the example).

Actually, it also tells many other editors...

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,5 @@
+root = true
+
+[*]
+indent_style = tab
+tab_width = 4


JohannesLorenz:
CI has passed now. I'll merge without review in 5 days.

BaraMGB:
Do you think, there is a risk? I think it would be a great relief.

JohannesLorenz:
Not really. I just want to give everyone time to veto.

JohannesLorenz:
OK, it's working on every branch where it's present.

PR #5043, opened 2019-06-17
Drop sample on sampletracks

### LINKED ISSUE #3228: Drop sample on sampletrack should add new tco
Currently a new audio file processor track will be inserted by dropping a sample on a (empty) sample track. Intuitive I would expect to get a new tco on that sample track.

BaraMGB:
fixes https://github.com/LMMS/lmms/issues/3228

You can drop your samples to Song editor or BB Editor. If you drop on a sample track it will add to it at the dropped position. ~~Else it will create a new sample track.~~ Otherwise it will be added to a new Audio File processor as usual.

### DIFF
--- a/include/SongEditor.h
+++ b/include/SongEditor.h
@@ -31,7 +31,6 @@
 
 #include "ActionGroup.h"
 #include "Editor.h"
-#include "TimeLineWidget.h"
 #include "TrackContainerView.h"
 
 class QLabel;
@@ -75,10 +74,6 @@ public:
 
 	ComboBoxModel *zoomingModel() const;
 
-	int timeLineWidgetHeight() const
-	{
-		return m_timeLine->height();
-	}
 
 public slots:
 	void scrolled( int new_pos );
--- a/include/SongEditor.h
+++ b/include/SongEditor.h
@@ -31,6 +31,7 @@
 
 #include "ActionGroup.h"
 #include "Editor.h"
+#include "TimeLineWidget.h"
 #include "TrackContainerView.h"
 
 class QLabel;
@@ -74,6 +75,11 @@ public:
 
 	ComboBoxModel *zoomingModel() const;
 
+	int timeLineWidgetHeight() const
+	{
+		return m_timeLine->height();
+	}
+
 public slots:
 	void scrolled( int new_pos );
 
@@ -104,7 +110,6 @@ private slots:
 	void updateScrollBar(int len);
 
 	void zoomingChanged();
-	void timeLineHeigthChanged();
 
 private:
 	virtual void keyPressEvent( QKeyEvent * ke );
--- a/include/SongEditor.h
+++ b/include/SongEditor.h
@@ -74,7 +74,6 @@ public:
 
 	ComboBoxModel *zoomingModel() const;
 
-
 public slots:
 	void scrolled( int new_pos );
 
--- a/include/SongEditor.h
+++ b/include/SongEditor.h
@@ -104,6 +104,7 @@ private slots:
 	void updateScrollBar(int len);
 
 	void zoomingChanged();
+	void timeLineHeigthChanged();
 
 private:
 	virtual void keyPressEvent( QKeyEvent * ke );
--- a/include/TrackContainerView.h
+++ b/include/TrackContainerView.h
@@ -32,9 +32,10 @@
 #include <QThread>
 
 #include "Track.h"
+#include "SampleTrack.h"
 #include "JournallingObject.h"
 #include "InstrumentTrack.h"
-
+#include <QDebug>
 
 class QVBoxLayout;
 class TrackContainer;
@@ -187,6 +188,7 @@ private:
 
 	RubberBand * m_rubberBand;
 	QPoint m_origin;
+	int m_trackHeadWidth;
 
 
 signals:
--- a/include/TrackContainerView.h
+++ b/include/TrackContainerView.h
@@ -32,7 +32,6 @@
 #include <QThread>
 
 #include "Track.h"
-#include "SampleTrack.h"
 #include "JournallingObject.h"
 #include "InstrumentTrack.h"
 
@@ -125,6 +124,11 @@ public:
 
 	RubberBand *rubberBand() const;
 
+	virtual int timeLineWidgetHeight() const
+	{
+		return 0;
+	}
+
 public slots:
 	void realignTracks();
 	TrackView * createTrackView( Track * _t );
@@ -153,7 +157,6 @@ protected:
 	virtual void resizeEvent( QResizeEvent * );
 
 	MidiTime m_currentPosition;
-	int m_timeLineWidgetHeight;
 
 
 private:
--- a/include/TrackContainerView.h
+++ b/include/TrackContainerView.h
@@ -153,7 +153,7 @@ protected:
 	virtual void resizeEvent( QResizeEvent * );
 
 	MidiTime m_currentPosition;
-	int m_timeLineWidgetHeigth;
+	int m_timeLineWidgetHeight;
 
 
 private:
--- a/include/TrackContainerView.h
+++ b/include/TrackContainerView.h
@@ -124,7 +124,6 @@ public:
 
 	RubberBand *rubberBand() const;
 
-
 public slots:
 	void realignTracks();
 	TrackView * createTrackView( Track * _t );
--- a/include/TrackContainerView.h
+++ b/include/TrackContainerView.h
@@ -36,7 +36,7 @@
 #include "JournallingObject.h"
 #include "InstrumentTrack.h"
 
-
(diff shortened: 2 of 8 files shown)


Spekular:
Does this remove the ability to drop samples on the left side of the song
editor to create new AFP instances? IMO that's the most convenient way to
add them.

include/TrackContainerView.h
```diff
@@ -152,6 +153,7 @@ public slots:
 	virtual void resizeEvent( QResizeEvent * );
 
 	MidiTime m_currentPosition;
+	int m_timeLineWidgetHeigth;
```
Spekular:
```suggestion
	int m_timeLineWidgetHeight;
```

src/gui/TrackContainerView.cpp
```diff
@@ -51,6 +51,7 @@ TrackContainerView::TrackContainerView( TrackContainer * _tc ) :
 	JournallingObject(),
 	SerializingObjectHook(),
 	m_currentPosition( 0, 0 ),
+	m_timeLineWidgetHeigth(0),
```
Spekular:
```suggestion
	m_timeLineWidgetHeight(0),
```

src/gui/TrackContainerView.cpp
```diff
@@ -391,6 +393,28 @@ void TrackContainerView::dropEvent( QDropEvent * _de )
 		//it->toggledInstrumentTrackButton( true );
 		_de->accept();
 	}
+	else if(type == "samplefile")
+	{
+		const TrackView * trackView = trackViewAt(_de->pos().y() - m_timeLineWidgetHeigth);
```
Spekular:
```suggestion
		const TrackView * trackView = trackViewAt(_de->pos().y() - m_timeLineWidgetHeight);
```

src/gui/editors/SongEditor.cpp
```diff
@@ -555,6 +555,7 @@ static inline void animateScroll( QScrollBar *scrollBar, int newVal, bool smooth
 
 void SongEditor::updatePosition( const MidiTime & t )
 {
+	m_timeLineWidgetHeigth = m_timeLine->height();
```
Spekular:
```suggestion
	m_timeLineWidgetHeight = m_timeLine->height();
```

Spekular (changes requested):
Heigth -> Height

BaraMGB:
@Spekular what would you suggest? Create a AFP - track if the sample is not dropped on a sample track? Could work, too...

Spekular:
@BaraMGB I was thinking about samples dragged from the sidebar, but now I realize this is probably for samples dragged from another program (file explorer) onto LMMS. As long as the sidebar behavior isn't removed I'm fine with the current PR behavior.

BaraMGB:
@Spekular I'm confused. This PR is about samples drag and drop from the sidebar.

Spekular:
Well then my original question stands. Current behavior when a sample is dragged from the sidebar is to add a new AFP instance. Creating a sample TCO when the sample is dropped on top of a sample track in the timeline is great, but dropping elsewhere should still create an AFP, otherwise you're removing the most convenient way to add AFP instances.

BaraMGB:
I'll implement it this way

src/gui/TrackContainerView.cpp
```diff
@@ -377,18 +378,46 @@ void TrackContainerView::dropEvent( QDropEvent * _de )
...
+				{
+					int trackHeadWidth = ConfigManager::inst()->value("ui", "compacttrackbuttons").toInt()==1
+											 ? DEFAULT_SETTINGS_WIDGET_WIDTH_COMPACT + TRACK_OP_WIDTH_COMPACT
+											 : DEFAULT_SETTINGS_WIDGET_WIDTH + TRACK_OP_WIDTH;
+
+					int xPos = _de->pos().x() < trackHeadWidth
+											  ? trackHeadWidth
+											  : _de->pos().x();
+
+					SampleTCO * newTco = new SampleTCO(tv->getTrack());
```
Reflexe: I think you have to use the Track::createTCO function instead.

src/gui/TrackContainerView.cpp
```diff
@@ -377,18 +378,46 @@ void TrackContainerView::dropEvent( QDropEvent * _de )
...
+										 ? MidiTime((xPos - trackHeadWidth) / pixelsPerTact()
+													* MidiTime::ticksPerTact()).toNearestTact()
+										 : MidiTime(0)
+										   );
+				}
+			}
+		}
+		else
+		{
+			InstrumentTrack * it = dynamic_cast<InstrumentTrack *>(
```
Reflexe: correct me if i'm wrong, isn't it another feature?

src/gui/editors/SongEditor.cpp
```diff
@@ -625,6 +626,14 @@ void SongEditor::zoomingChanged()
 
 
 
+void SongEditor::timeLineHeigthChanged()
```
Reflexe: I think we could add a virtual `timeLineWidget` getter instead of this signal, member variable and slot. But that's ok too.

src/gui/TrackContainerView.cpp
BaraMGB: This code handles the file types if we are not dropping on a sampletrack or it's not a sample file:
```
(type == "samplefile" || type == "pluginpresetfile"
			|| type == "soundfontfile" || type == "vstpluginfile"
			|| type == "patchfile")
```

PhysSong:
Question: why don't you use `SampleTrackView` to handle drag-and-drop as like `AutomationTrackView` does? I think that will be much simpler.

BaraMGB:
@PhysSong looks good, I will implement it in SampleTrackView().

src/gui/TrackContainerView.cpp
```diff
@@ -377,18 +377,26 @@ void TrackContainerView::dropEvent( QDropEvent * _de )
...
-		PluginFactory::PluginInfoAndKey piakn =
-			pluginFactory->pluginSupportingExtension(FileItem::extension(value));
-		Instrument * i = it->loadInstrument(piakn.info.name(), &piakn.key);
-		i->loadFile( value );
-		//it->toggledInstrumentTrackButton( true );
+		const TrackView * tv = trackViewAt(_de->pos().y() - timeLineWidgetHeight());
+		if (type == "samplefile" && tv && tv->getTrack()->type() == Track::SampleTrack)
+		{
+			_de->accept();
+		}
```
PhysSong: I think you don't even need this part, `SampleTrackView::dragEnterEvent` will do the job. If this works, you need to change only `SampleTrackView`.

BaraMGB: Seems to work. :+1:

include/SongEditor.h
```diff
@@ -74,6 +75,11 @@ class SongEditor : public TrackContainerView
 
 	ComboBoxModel *zoomingModel() const;
 
+	int timeLineWidgetHeight() const
+	{
+		return m_timeLine->height();
+	}
```
PhysSong: This can be removed now, with the `"TimeLineWidget.h"` include.

include/TrackContainerView.h
```diff
@@ -124,6 +124,11 @@ class TrackContainerView : public QWidget, public ModelView,
 
 	RubberBand *rubberBand() const;
 
+	virtual int timeLineWidgetHeight() const
+	{
+		return 0;
+	}
```
PhysSong: This can be removed now.

src/gui/editors/SongEditor.cpp
```diff
@@ -44,7 +44,6 @@
 #include "MeterDialog.h"
 #include "Mixer.h"
 #include "TextFloat.h"
-#include "TimeLineWidget.h"
```
PhysSong: Can(should?) be re-added now.

src/gui/TrackContainerView.cpp
```diff
@@ -377,7 +377,7 @@ void TrackContainerView::dropEvent( QDropEvent * _de )
 		//it->toggledInstrumentTrackButton( true );
 		_de->accept();
 	}
-	else if( type == "samplefile" || type == "pluginpresetfile" 
+	else if( type == "samplefile" || type == "pluginpresetfile"
```
PhysSong: Can be reverted, but not necessary.

PhysSong:
Looks good. :+1:

src/gui/TrackContainerView.cpp
BaraMGB: I tried it, Qtcreator removes it on save. :) I can force it by using an other editor if it is necessary.

BaraMGB:
@PhysSong do you think it's ready for merging?

PhysSong (approved):

PR #1426, opened 2014-12-08
Editor Refactoring

Lukas-W:
Refactoring the editors' code. Introduced a new superclass for them providing a toolbar.
This is a try to make the code somewhat maintainable, but there's just so much done wrong there, it's still a long way.

PR #5060, opened 2019-06-27
show BBEditor on clicking the TrackLabelButton

BaraMGB:
In the LMMS.io forums a user reported the following:

> I have an issue in new stable release 1.2.0
>
> When you close the beat+bass line editor screen, and if you click to open it again, the beat+bass line editor screen don't open, you have to press the shortcut to open it again.
> 
> I have tried in another PC and it's the same.
> In LMMS 1.2.0 RC8 the behavior is normal and open and close as always.

The user means the TrackLabelButton. This PR fixes the issue.

### DIFF
--- a/src/tracks/BBTrack.cpp
+++ b/src/tracks/BBTrack.cpp
@@ -664,5 +664,5 @@ bool BBTrackView::close()
 void BBTrackView::clickedTrackLabel()
 {
 	Engine::getBBTrackContainer()->setCurrentBB( m_bbTrack->index() );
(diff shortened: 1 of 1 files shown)


PhysSong:
I guess it's introduced in #1426, and the change in this fix is consitent with the PR.

BaraMGB:
Do you think, we shouldn't fix it in stable? @PhysSong

PhysSong:
Yes, it's a regression in 1.2.

PhysSong:
Merge?

PR #5069, opened 2019-07-06
Closes #5050: Add Mixer threading comments

### LINKED ISSUE #5050: Race conditions when deleting effects?
@he29-net and me were wondering if there are race conditions possible when Effects are deleted while being used.

`EffectRackView::deletePlugin( EffectView* view )` does

```
fxChain()->removeEffect( e );
e->deleteLater(); // probably this deletes e after the current callback returned to Qt
```

Imagine meanwhile, the audio thread just runs. Can the following order happen and lead to a crash?

1. User uses UI to delete effect
2. audio thread fetches effect pointer
3. UI thread deletes effect pointer from fx chain (`EffectChain::removeEffect`)
4. UI thread returns to Qt event loop and deletes effect
5. audio thread now uses its fetched pointer, which points to cleaned and deleted memory

JohannesLorenz:
@PhysSong Can you please check if those 2 comments are correct?

### DIFF
--- a/include/Mixer.h
+++ b/include/Mixer.h
@@ -315,6 +315,7 @@ public:
 	inline bool isMetronomeActive() const { return m_metronomeActive; }
 	inline void setMetronomeActive(bool value = true) { m_metronomeActive = value; }
 
+	//! Block until a change in model can be done (i.e. wait for audio thread)
 	void requestChangeInModel();
 	void doneChangeInModel();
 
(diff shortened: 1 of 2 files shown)


include/Mixer.h
```diff
@@ -366,6 +367,8 @@ class LMMS_EXPORT Mixer : public QObject
 
 	void clearInternal();
 
+	//! Function to be called before processing audio
+	//! Blocks if there are currently changes done in model
```
PhysSong: Actually, it's called after processing and before writing the buffer to the audio device or a file.
Also, grammars are a bit odd in the last sentence.

JohannesLorenz:
@PhysSong Thanks. I did a force push with fixes. Can you please check?

PR #5080, opened 2019-07-16
Enable automatic dll collection on MinGW and activate CircleCI artifacts. 

Reflexe:
This PR is blocked by https://github.com/LMMS/lmms-ci-docker/pull/3


CMakeLists.txt
```diff
@@ -3,6 +3,9 @@ CMAKE_MINIMUM_REQUIRED(VERSION 3.3)
 PROJECT(lmms)
 
 SET(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/modules" ${CMAKE_MODULE_PATH})
+set(CMAKE_INSTALL_PREFIX ${CMAKE_BINARY_DIR})
```
PhysSong: @Reflexe It looks weird. Are there any reasons for it?

cmake/win32/CreateFixupBundle.cmake
```diff
@@ -0,0 +1,32 @@
+#
+function(create_fixup_bundle)
+    set(options)
+    set(oneValueArgs BINARY_PATH SCRIPT_OUTPUT_PATH)
+    set(multiValueArgs LIBRARIES_PATHS IGNORE_ITEMS)
+    cmake_parse_arguments(BUNDLE "${options}" "${oneValueArgs}"
+            "${multiValueArgs}" ${ARGN} )
+
+      configure_file(
+        "${LMMS_SOURCE_DIR}/cmake/win32/FixupBundle.gen.cmake"
```
PhysSong: We usually use `.in` extension for files to be configured. So, I'm suggesting `FixupBundle.cmake.in` for the name of the input file and `FixupBundle.cmake` for the output.

cmake/win32/FixupBundle.gen.cmake
```diff
@@ -0,0 +1,27 @@
...
+    string(TOUPPER ${item} item)
+    if (item MATCHES ${WIN_SYSTEM_LIBS})
+        set(${resolved_var} 1 PARENT_SCOPE)
+    endif()
+endfunction()
+
+# Override tool for mingw.
+if (NOT "@MSVC@")
+    set(gp_tool "objdump")
+endif()
```
PhysSong: I think `GetPrerequisites` already do this for MinGW.

src/CMakeLists.txt
```diff
@@ -137,6 +137,16 @@ SET_DIRECTORY_PROPERTIES(PROPERTIES ADDITIONAL_MAKE_CLEAN_FILES "${LMMS_RCC_OUT}
 
 IF(LMMS_BUILD_WIN32)
 	SET(EXTRA_LIBRARIES "winmm")
+
+	# Find Qt's qwindows.dll and install it.
```
PhysSong: The standard way to do that in CMake is:
```cmake
IF(LMMS_BUILD_WIN32)
	INSTALL(FILES $<TARGET_FILE:Qt5::QWindowsIntegrationPlugin> DESTINATION platforms)
ENDIF()
```
(from https://github.com/lukas-w/lmms/commit/4d5cd9a8cc2d170c09b8259436bebc363f397c66?diff=split#diff-e7bc98360bbad3a6a2c524ae5bc94b8fR38)

cmake/win32/FixupBundle.gen.cmake
Reflexe: it uses `ldd` on linux - always. on windows with mingw it would use objdump.

CMakeLists.txt
Reflexe: @PhysSong probably left it by mistake after testing. thanks.

cmake/win32/FixupBundle.gen.cmake
PhysSong: > it uses `ldd` on linux

When cross-compiling, it uses `objdump`.

Reflexe: See for yourself; without those lines:
```
Run CPack packaging tool...
CPack: Create package using NSIS
CPack: Install projects
CPack: - Run preinstall target for: lmms
CPack: - Install project: lmms
CMake Error at /usr/share/cmake-3.10/Modules/GetPrerequisites.cmake:838 (message):
  /usr/bin/ldd failed: 1

Call Stack (most recent call first):
  /usr/share/cmake-3.10/Modules/BundleUtilities.cmake:628 (get_prerequisites)
  /usr/share/cmake-3.10/Modules/BundleUtilities.cmake:903 (get_bundle_keys)
  /lmms/build/FixupBundle.cmake:26 (fixup_bundle)
  /lmms/build/src/cmake_install.cmake:50 (include)
  /lmms/build/cmake_install.cmake:38 (include)


CPack Error: Error when generating package: lmms
Makefile:129: recipe for target 'package' failed
make: *** [package] Error 1
```
Also, from the cmake [docs](https://cmake.org/cmake/help/latest/module/GetPrerequisites.html):
```
It uses various tools to obtain the list of required shared library files:
dumpbin (Windows)
objdump (MinGW on Windows)
ldd (Linux/Unix)
otool (Mac OSX)
```

PhysSong: Oh, it seems like `WIN32` is defined only in the build step, not in the install step.

PhysSong:
Can we test that against AppVeyor as well?

cmake/win32/FixupBundle.gen.cmake
Reflexe: > Can we test that against AppVeyor as well?

Let's try. why not :D

PhysSong:
BTW, why did you use spaces for CMake-related files? We are using tabs in other files.

PhysSong:
It seems like plugin dependencies(ex. FluidSynth, libgig) are not bundled.

FYI, I have outdated [`cmake/install-refactor`](https://github.com/PhysSong/lmms/tree/cmake/install-refactor) branch which is based on [@lukas-w's one](https://github.com/PhysSong/lmms/tree/cmake/install-refactor). It worked for most plugins except VST. Please feel free to reuse our work, or you may reuse the old branch.

Reflexe:
> It seems like plugin dependencies(ex. FluidSynth, libgig) are not bundled.
> 
> FYI, I have outdated [`cmake/install-refactor`](https://github.com/PhysSong/lmms/tree/cmake/install-refactor) branch which is based on [@lukas-w's one](https://github.com/PhysSong/lmms/tree/cmake/install-refactor). It worked for most plugins except VST. Please feel free to reuse our work, or you may reuse the old branch.

That could be a good idea. With all of my work, I was not satisfied with my solution but 
I could not think of a better one. I'll close it and move to work on VST for the other branch.

PR #5083, opened 2019-07-20
Fixed renaming mixer channel from the context menu

freshstudio:
`renameChannel()` wasn't a slot previously, so the menu couldn't access it. This PR fixes that.


DouglasDGI:
Nice catch.  Tested it, everything's good.  I'll go ahead and merge it now.

Issue #323, opened 2014-02-16
SF2 with no "default" patch


tresf:
Some soundfonts have a default bank number a default a patch number that is higher than zero.

1. Download [045_Florestan_Pizzicato.sf2](https://musical-artifacts.com/artifacts/397) (or any SF2 where the default patch is non-zero)
2. Drag it into the Song Editor
3. Try playing a note

This is minor, as the SF2 still works great with LMMS, except when browsing sounds, the SF2 doesn't make a sound.

The request is to allow SF2's to "automagically" choose the first patch that is available for initial preview and playback.

Pictured is an example of a free soundfont that has a "default" patch of 45.
![image](https://f.cloud.github.com/assets/6345473/2180908/d92030cc-9737-11e3-82a2-5f457cda4a0b.png)

SecondFlight:
This appears to be fixed in RC5.

zonkmachine:
> This appears to be fixed in RC5.

I don't see this. @tresf Do you see this issue solved?

tresf:
> This appears to be fixed in RC5.

@SecondFlight on what evidence?

> I don't see this. @tresf Do you see this issue solved?

Thanks for catching this.  Of course not.  Nothing's changed to warrant a close.

![ezgif-1-01be90744630](https://user-images.githubusercontent.com/6345473/55303517-b9839c00-5414-11e9-822e-8b2baeb8e118.gif)

SecondFlight:
Huh. I know I tested this and thought the issue was solved, but I guess I didn't hit this specific edge case? And I certainly didn't document my process. Sorry about that!

tresf:
> default a patch number that is higher than zero.
> > didn't hit this specific edge case? 

The specific edge-case is the only unit test.  Sorry, I wrote this bug report 5 years ago so there's no exact steps to reproduce.   Here they are:

1. Download [045_Florestan_Pizzicato.sf2](https://musical-artifacts.com/artifacts/397) (or any SF2 where the default patch is non-zero)
2. Drag it into the Song Editor
3. Try playing a note

amazing-username:
Is this issue still valid? I want to give it a go as my first contribution.

tresf:
Yes, please!

amazing-username:
I'll give it a go, I appreciate the quick response.

PR #5084, opened 2019-07-20
Fix for the SF2 with no "default" patch

kdeng00:
Using the instructions from #323 to test the fix for this PR, I was able to reproduce the problem and provide a fix. Let me know if there is anything else that is left from the issue.


plugins/sf2_player/sf2_player.cpp
```diff
@@ -267,6 +268,46 @@ void sf2Instrument::loadFile( const QString & _file )
...
+			fluid_preset_t *pCurPreset = &preset;
+#else
+			fluid_preset_t *pCurPreset;
+#endif
+
+			while ( ( pCurPreset = fluid_sfont_iteration_next_wrapper( pSoundFont, pCurPreset ) ) ) {
+#ifdef CONFIG_FLUID_BANK_OFFSET
+				iBank += iBankOff;
+#endif
+				if ( pCurPreset && !foundFirstPreset ) {
```
PhysSong: * Checking for `pCurPreset` looks redundant because the `while` statement already checks for it.
* Can you use an `if` statement and a `break` statement instead of this double-loop? The current implementation looks very inefficient.

plugins/sf2_player/sf2_player.cpp
```diff
@@ -267,6 +268,46 @@ void sf2Instrument::loadFile( const QString & _file )
...
+			fluid_preset_t preset;
+			fluid_preset_t *pCurPreset = &preset;
+#else
+			fluid_preset_t *pCurPreset;
+#endif
+
+			while ( ( pCurPreset = fluid_sfont_iteration_next_wrapper( pSoundFont, pCurPreset ) ) ) {
+#ifdef CONFIG_FLUID_BANK_OFFSET
+				iBank += iBankOff;
+#endif
```
PhysSong: I think `iBankOff` should be added to `iBank` after getting the value of `iBank`.

plugins/sf2_player/sf2_player.cpp
```diff
@@ -184,6 +184,7 @@ sf2Instrument::sf2Instrument( InstrumentTrack * _instrument_track ) :
 
 	InstrumentPlayHandle * iph = new InstrumentPlayHandle( this, _instrument_track );
 	Engine::mixer()->addPlayHandle( iph );
+
```
PhysSong: Could you revert unnecessary whitespace changes(this includes new lines) if you can?

kdeng00:
Hey @PhysSong, 
Thanks for the feedback and I made the changes. However, there is one test that is failing for Mac OS X and I can't seem to understand why it is failing.

plugins/sf2_player/sf2_player.cpp
kdeng00: Made the fix https://github.com/LMMS/lmms/pull/5084/commits/37e151b5fc1e7661d153a782e449e6db91227257

plugins/sf2_player/sf2_player.cpp
kdeng00: Made the fix https://github.com/LMMS/lmms/pull/5084/commits/37e151b5fc1e7661d153a782e449e6db91227257

plugins/sf2_player/sf2_player.cpp
kdeng00: Made the fix https://github.com/LMMS/lmms/pull/5084/commits/37e151b5fc1e7661d153a782e449e6db91227257

zonkmachine:
> However, there is one test that is failing for Mac OS X and I can't seem to understand why it is failing.

I think it's Homebrew acting up a bit. Restarting the Mac build a couple of times worked.

kdeng00:
Sounds good, I would have thought that since Mac OS was POSIX compliant that it would have just been compatible but hombrew tool is something else.

PR #5085, opened 2019-07-20
Add 'Change color' option to mixer channels

v0idzz:
This enables the user to choose whatever color they want for a mixer channel. This is my first... bigger(?) PR, so don't be surprised if something's wrong. :stuck_out_tongue_closed_eyes:

### DIFF
--- a/src/gui/widgets/FxLine.cpp
+++ b/src/gui/widgets/FxLine.cpp
@@ -105,8 +105,8 @@ FxLine::FxLine( QWidget * _parent, FxMixerView * _mv, int _channelIndex ) :
 	setToolTip( name );
 
 	m_useStyleColor = channel->m_useStyleColor;
-	m_customBackgroundActive.setColor( channel->m_customColor );
-	m_customBackground.setColor( channel->m_customColor.darker( 150 ) );
+	m_customBackgroundActive.setColor( channel->m_customColor.lighter( 125 ) );
+	m_customBackground.setColor( channel->m_customColor );
 
 	m_renameLineEdit = new QLineEdit();
 	m_renameLineEdit->setText( name );
@@ -281,8 +281,8 @@ void FxLine::changeColor()
 	targetChannel->m_useStyleColor = false;
 
 	m_useStyleColor = false;
-	m_customBackgroundActive.setColor( color );
-	m_customBackground.setColor( color.darker( 150 ) );
+	m_customBackgroundActive.setColor( color.lighter( 125 ) );
+	m_customBackground.setColor( color );
 
 	update();
 }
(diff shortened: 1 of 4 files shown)


DouglasDGI:
I gave it a test, it seems to work fine for the most part.  I did notice one very minor bug where resetting the color of the selected bus to default doesn't update the color on the screen until you click on the fader or switch to another fader.

Also, how the colors are handled may need some discussion.  From what I can see, any unselected channels are combined with the color black, which can make the channel colors look ugly.  Maybe instead, we should make the default-colored channels the dark color, and then mix the selected channel with the color white?  Any opinions?  Maybe that would break some LMMS themes, I'm not sure.

Otherwise, I did not notice any bugs or problems.  :+1:

src/gui/widgets/FxLine.cpp
```diff
@@ -245,6 +265,40 @@ void FxLine::contextMenuEvent( QContextMenuEvent * )
...
+}
+
+
+
+void FxLine::resetColor()
+{
+	FxChannel * targetChannel = Engine::fxMixer()->effectChannel( m_channelIndex );
+	targetChannel->m_useStyleColor = true;
+	m_useStyleColor = true;
+}
```
PhysSong: I think you should add `update();` at the end of this function to fix the bug discovered by @DouglasDGI.

PhysSong:
The bug fixed. :+1:

noahb01:
I agree with @DouglasDGI, we should lighten the selected channel rather than darkening all the unselected ones. Qt has a function `QColor::lighten()` that could be used for this. I don't think its a big deal if it breaks themes, its not a major change and it will look different but not necessarily bad.

qnebra:
It colors "song editor" too, or only mixer?

freshstudio:
> It colors "song editor" too, or only mixer?

It has no effect on the song editor.

I agree with @DouglasGDI and @noahb01. Pushing changes which make it work the opposite way to what I have implemented.

PhysSong:
To remove the GUI dependency added in this PR, some extra changes are needed to loading/saving FX channel settings.

russiankumar:
Didn't know this existed... made a PR #5589 which does this same thing.

claell:
@russiankumar Too bad. Is there any chance this can get reviewed, tested and merged soon?

russiankumar:
I kind of see no point, considering the newer one can already be merged without any rebasing hassle, while this one will have to wait. The newer one was already tested by some as well, seems to work as expected, so I’d just go with that one. Though I want a mod to confirm this, since technically I’m biased.

claell:
Probably the best, especially since this seems not to be very active at the moment.

Spekular:
Superceded by #5589, which is now merged.

PR #5086, opened 2019-07-21
Don't use deprecated Qt5 functions

noahb01:
This PR removes a number of deprecated QT functions.
`QColor::dark()` becomes `QColor::darker()`
`QColor::light()` becomes `QColor::lighter()`
`QFontMetrics::width()` becomes `QFontMetrics::horizontalAdvance()`
`QDir` objects can no longer have their paths set via `operator=`, use `QDir::setPath()` instead

I still get compiler warnings about obsolete `QSignalMapper` but I don't know enough Qt to remove them.


noahb01:
This PR builds fine on my machine (Arch Linux, x86-64). However, it fails with a compilation error on travis and circleci...

PhysSong:
> `QFontMetrics::width()` becomes `QFontMetrics::horizontalAdvance()`

The change was introduced in Qt 5.11, but the latest Qt version that's officially supported is 5.9. I think you should either revert the change or use `#define horizontalAdvance width` with a Qt version check.

noahb01:
The `#define` seems really ugly to me, so I'll just revert it for now. Once Qt 5.9 support is dropped (or when the latest version of Qt fully removes `width()`, we can do a new PR to remove it.

PhysSong:
> I still get compiler warnings about obsolete `QSignalMapper` but I don't know enough Qt to remove them.

I think it can be replaced with functor-based connections. In other words, you can connect signals to lambda functions in Qt5. See https://doc.qt.io/qt-5/qobject.html#connect-4 and https://doc.qt.io/qt-5/qobject.html#connect-5 for details.
If you're not familiar with the technique, you can leave that part as-is.

noahb01:
OK, `QSignalMapper` has been removed.

noahb01:
Sorry about that, my editor was misconfigured and I didnt notice until after commit+push. Shouldnt happen again.

PhysSong:
I found that `src/tracks/InstrumentTrack.cpp` and `src/tracks/SampleTrack.cpp` also uses deprecated `QSignalMapper`. You can use [an override of `QMenu::addAction`](https://doc.qt.io/qt-5/qmenu.html#addAction-4) which supports lambda functions, but it requires Qt 5.6. To support old Qt(our Windows builds on Travis CI still uses 5.4), you need to create a action without a signal-slot connection and then connect to a lambda function.

lukas-w:
Superseded by #5156.

PR #5101, opened 2019-07-28
Fully hide song editor scrollbar when not needed

Spekular:
Previously, the scrollbar's background area would still be displayed.
To work around this, resize the scrollbar area to 0px when it's not needed.

### DIFF
--- a/src/gui/TrackContainerView.cpp
+++ b/src/gui/TrackContainerView.cpp
@@ -227,8 +227,19 @@ void TrackContainerView::scrollToTrackView( TrackView * _tv )
 void TrackContainerView::realignTracks()
 {
 	QWidget * content = m_scrollArea->widget();
-	content->setFixedWidth( width()
-				- m_scrollArea->verticalScrollBar()->width() );
+
+	//Hide the vertical scrollbar *and* its scrollbar area when not needed
+	auto verticalScrollBar = m_scrollArea->verticalScrollBar();
+	if (verticalScrollBar->maximum() == verticalScrollBar->minimum())
+	{	//min == max implies the whole editor is visible without scrolling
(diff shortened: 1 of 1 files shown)


BaraMGB:
I found a simpler solution:

https://github.com/LMMS/lmms/blob/master/src/gui/TrackContainerView.cpp#L491
becomes to:
`	setVerticalScrollBarPolicy( Qt::ScrollBarAsNeeded );`
and:
https://github.com/LMMS/lmms/blob/master/src/gui/TrackContainerView.cpp#L231
becomes to:
`	content->setFixedWidth(width());`

Spekular:
`setVerticalScrollBarPolicy` is set to `ScrollBarAsNeeded` by default, and I saw no difference by calling `m_scrollArea->setVerticalScrollBarPolicy( Qt::ScrollBarAsNeeded )` in `TrackContainerView::TrackContainerView`. But if your version works then that is indeed a whole lot simpler. I guess you tested it @BaraMGB ?

BaraMGB:
Sorry I reformated the comment above.

BaraMGB:
Yes I've tested it.

Spekular:
Also, I believe there will be a slight visual difference between my solution and always having the full width. The timeline drew above (vertically, not depthwise) the scrollbar when I did that. It doesn't look very buggy though so I suppose that behavior is fine.

BaraMGB:
It looks exactly the same.

Spekular:
It does indeed.

Spekular:
@BaraMGB You may as well commit this yourself, if you don't mind. You can remove this line entirely, I just tested: https://github.com/LMMS/lmms/blob/1c715bc335b70d92b76bbb5d8c8c71dfba710be7/src/gui/TrackContainerView.cpp#L491

PR #5102, opened 2019-07-28
shows the vertical scrollbar of songeditor and bb-editor only if needed

BaraMGB:
As a simpler solution for the PR from @Spekular : https://github.com/LMMS/lmms/pull/5101


BaraMGB:
I will merge this within the next few hours.

Issue #3682, opened 2017-07-03
Sidebar close button


M374LX:
A close button for the sidebar would be instantly recognizable, unlike the need to click on the currently selected option to hide the bar:
![2017-07-03-194642_1920x1080_scrot](https://user-images.githubusercontent.com/6584238/27809888-5b04d34a-602a-11e7-9863-8798ca7aaccc.png)

Another small benefit this would bring is that the mouse cursor, when located on any of the editors (as in most of the time), would have a shorter distance to travel.

Hussam-Eddin-Alhomsi:
I don't like the current way of closing the open tab for this reason:
when the user opens a tab and performs a drag & drop action for a file, he will always do the action on the right side of the GUI, and to close the open tab, he will have to return all the way to the left.

But if we are going to change how tabs are closed, I think we should change how they open.

tresf:
@Hussam-Eddin-Alhomsi, you'll have to be more specific.  The mouse dragging across the screen and back again paradox a common UI annoyance.  I'm not sure how that is specific to LMMS.

To avoid the sidebar is impossible given our current UI.  You should either explicitly stated how this can be improved within scope of this PR, refute this PR, or open and crosslink a new enhancement.  We embrace great ideas but they shouldn't get in the way of good ones.

budislav:
Just add one more button in the main toolbar.
Is it possible to move tabs on the top, just icons?
![untitled](https://user-images.githubusercontent.com/3619927/28028915-7c07cd6c-659e-11e7-827f-46c464c2a931.png)

Hussam-Eddin-Alhomsi:
> A couple of reasons why I propose this positioning of the buttons is that vertical reading is impossible so text there is useless. We have only icons and with only icons we have a lot of empty space. Space is precious so we can just show and hide browser with button on the top just like every other window.

I totally agree. 💯 
I like your mockup because it removes the icons from the far left and also removes the vertical text, and becuase you positioned the text/image of the open tab in a proper position, becuase it's currently positioned like this:
![image](https://user-images.githubusercontent.com/25503477/28088609-e105e5d2-668e-11e7-81d0-6c98b70db12d.png)

budislav:
Yes and there is no need for two same icons, also I would remove title underline. 
[...]

PR #5114, opened 2019-08-05
Add option to move SideBar to right side of window

Veratil:
Reading through #3682, I saw a [suggestion](https://github.com/LMMS/lmms/issues/3682#issuecomment-326762851) about adding an option to move the SideBar to the right side of the window.

This adds a new configuration option to the SetupDialog, but for some reason on my end it would barely overlap the start of the `Plugin Embedding` section. I'm not sure what to modify to get that to work.

### DIFF
--- a/src/gui/MainWindow.cpp
+++ b/src/gui/MainWindow.cpp
@@ -117,6 +117,13 @@ MainWindow::MainWindow() :
 	splitter->setChildrenCollapsible( false );
 
 	ConfigManager* confMgr = ConfigManager::inst();
+	bool sideBarOnRight = confMgr->value("ui", "sidebaronright").toInt();
+
+	// create the QMdiArea with an empty splitter so the workspace is on the left
+	if(sideBarOnRight)
+	{
+		m_workspace = new QMdiArea(splitter);
+	}
 
 	emit initProgress(tr("Preparing plugin browser"));
 	sideBar->appendTab( new PluginBrowser( splitter ) );
@@ -171,7 +178,11 @@ MainWindow::MainWindow() :
 					embed::getIconPixmap( "computer" ).transformed( QTransform().rotate( 90 ) ),
 							splitter, dirs_as_items) );
 
-	m_workspace = new QMdiArea( splitter );
+	// create the QMdiArea with a populated splitter so the workspace is on the right
+	if(!sideBarOnRight)
+	{
+		m_workspace = new QMdiArea(splitter);
+	}
 
 	// Load background
 	emit initProgress(tr("Loading background picture"));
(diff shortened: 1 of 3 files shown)


src/gui/SetupDialog.cpp
```diff
@@ -249,6 +251,15 @@ SetupDialog::SetupDialog( ConfigTabs _tab_to_open ) :
...
 		m_openLastProject, SLOT(toggleOpenLastProject(bool)));
 
+	LedCheckBox * sideBarOnRight = new LedCheckBox(
+				tr( "Sidebar on right" ),
+								misc_tw );
+	labelNumber++;
+	sideBarOnRight->move( XDelta, YDelta*labelNumber );
+	sideBarOnRight->setChecked( m_sideBarOnRight );
+	connect ( sideBarOnRight, SIGNAL( toggled( bool ) ),
+				this, SLOT( toggleSideBarOnRight( bool ) ) );
```
PhysSong: As there's `addLedCheckBox` function, you can do it in two lines:
```cpp
addLedCheckBox("Sidebar on right",
		m_sideBarOnRight, SLOT(toggleSideBarOnRight(bool)));
```
Also, I suggest changing the text a bit. IMO it should be something like `Show sidebar on the right-hand side`.

Veratil: > As there's `addLedCheckBox` function, you can do it in two lines:

I realize now that I'm certain I started this from the `stable-1.2` branch instead of `master`. I must not have realized that when I rebased it.

> Also, I suggest changing the text a bit. IMO it should be something like `Show sidebar on the right-hand side`.

Sure thing! I just put it as a placeholder until I got a good suggestion of what would be preferred.

PhysSong:
There are merge conflicts since #3820 was merged, but the conflict looks quite small.

PhysSong:
You can use [`QSplitter::insertWidget()`](https://doc.qt.io/qt-5/qsplitter.html#insertWidget) to move the MDI area instead of changing the execution order. However, the current way is fine, too.

Veratil:
> You can use [`QSplitter::insertWidget()`](https://doc.qt.io/qt-5/qsplitter.html#insertWidget) to move the MDI area instead of changing the execution order. However, the current way is fine, too.

Nifty! Updated the code to use that now. 😄

src/gui/MainWindow.cpp
```diff
@@ -194,9 +195,13 @@ MainWindow::MainWindow() :
 	m_workspace->setHorizontalScrollBarPolicy( Qt::ScrollBarAsNeeded );
 	m_workspace->setVerticalScrollBarPolicy( Qt::ScrollBarAsNeeded );
 
-	hbox->addWidget( sideBar );
-	hbox->addWidget( splitter );
-
+	hbox->addWidget(sideBar);
+	hbox->addWidget(splitter);
+	if(sideBarOnRight)
```
PhysSong: To match our coding style:
```suggestion
	if (sideBarOnRight)
```

PhysSong:
BTW, the code still works well. :+1:

src/gui/MainWindow.cpp
PhysSong: Also, can you add a comment which explains what this block does?

PhysSong:
I think it's safe to merge this. Any objections?

Issue #5097, opened 2019-07-27
Proposal for a bigger, modern looking time display


michaelgregorius:
I have reactivated some old branch where I have experimented with a more modern looking time display. If you want to check out the current state for yourself you can find the branch at https://github.com/michaelgregorius/lmms/tree/timedisplay.

## New look
Here's how it looks when showing the time:
![TimeDisplayNewTime](https://user-images.githubusercontent.com/9293269/61995901-66ea8180-b08e-11e9-8eaa-36ad1c85f019.png)

Here's how it looks when showing beats, bars and ticks:
![TimeDisplayNewBarBeatTick](https://user-images.githubusercontent.com/9293269/61995904-736eda00-b08e-11e9-9137-1e236eaf3fe0.png)

As you can see the tempo widget was moved to the right which makes the layout less cluttered.

## Current look
For comparison here's the old implementation showing time:
![TimeDisplayOldTime](https://user-images.githubusercontent.com/9293269/61995910-808bc900-b08e-11e9-9a82-382b6b86418b.png)

And the old implementation showing beats, bars and ticks:
![TimeDisplayOldBarBeatTick](https://user-images.githubusercontent.com/9293269/61995915-8ed9e500-b08e-11e9-8534-ced7b08c03a8.png)

## Caveat
However, there's still one caveat with the new implementation. Because it uses several labels the font used by the application must have numbers which are of equal width. Otherwise the layout will jump around when the time is updated. I think this might be fixed by putting a font into the resources of the application and using this font to display the time. That way it should work the same everywhere.

What do you think?

kriskropd:
I don't really like it. Maybe it's because it shares the same color of the icons, but it blends in to the background too much. To be honest though, I prefer the current look already and not sure why it needs to change.

If it's a matter of size and readability, what if we just stretched the existing counters to be larger?

Spekular:
Now thay you've mentioned it, I can see that the current layout might look a bit cluttered. That said, I don't find this particular style to improve it, and i share the readability concerns. In particular, I think it's an odd choice to simply have the numbers on the toolbar without any frame/bg/visual element to delineate it from the rest.

BaraMGB:
@Spekular  Maybe we can approach the design of @budislav ?


![](https://user-images.githubusercontent.com/3619927/28041705-c3864098-65ca-11e7-9d8c-5b8108941c18.png)

michaelgregorius:
[...]

PR #5118, opened 2019-08-10
WIP: Redesign time display

winniehell:
This creates a new design for the time display widget.

It is still work in progress because the code needs some cleanup before merging.

| Before | After |
| --- | --- |
| Bildschirmfoto vom 2019-08-10 23-57-26 | Bildschirmfoto vom 2019-08-10 23-55-35 |
| Bildschirmfoto vom 2019-08-10 23-57-31 | Bildschirmfoto vom 2019-08-10 23-55-40 |

video: 
[new-time-display.zip](https://github.com/LMMS/lmms/files/3489322/new-time-display.zip)

closes https://github.com/LMMS/lmms/issues/5097

### DIFF
--- a/src/gui/widgets/TimeDisplayWidget.cpp
+++ b/src/gui/widgets/TimeDisplayWidget.cpp
@@ -109,28 +109,36 @@ void TimeDisplayWidget::updateTime()
 {
 	Song* s = Engine::getSong();
 
+	int major = 0;
+	int minor = 0;
+	int milliSeconds = 0;
+
 	switch( m_displayMode )
 	{
 		case MinutesSeconds:
 			int msec;
 			msec = s->getMilliseconds();
-			m_majorValue.setNum(msec / 60000);
-			m_minorValue.setNum((msec / 1000 ) % 60);
-			m_milliSecondsValue.setNum(msec % 1000);
+			major = msec / 60000;
+			minor = (msec / 1000) % 60;
+			milliSeconds = msec % 1000;
 			break;
 
 		case BarsTicks:
 			int tick;
 			tick = s->getPlayPos().getTicks();
-			m_majorValue.setNum((int)(tick / s->ticksPerTact() ) + 1);
-			m_minorValue.setNum( ( tick % s->ticksPerTact() ) /
-						 ( s->ticksPerTact() / s->getTimeSigModel().getNumerator() ) +1 );
-			m_milliSecondsValue.setNum((tick % s->ticksPerTact()) %
-							(s->ticksPerTact() / s->getTimeSigModel().getNumerator()));
+			major = (int)(tick / s->ticksPerTact()) + 1;
+			minor = (tick % s->ticksPerTact()) /
+						 (s->ticksPerTact() / s->getTimeSigModel().getNumerator()) +1;
+			milliSeconds = (tick % s->ticksPerTact()) %
+							(s->ticksPerTact() / s->getTimeSigModel().getNumerator());
 			break;
 
 		default: break;
 	}
+
+	m_majorValue.setText( QString::number( major ).rightJustified( 4, '0' ) );
+	m_minorValue.setText( QString::number( minor ).rightJustified( 2, '0' ).prepend( ':' ) );
+	m_milliSecondsValue.setText( QString::number( milliSeconds ).rightJustified( 3, '0' ).prepend( '.' ) );
 }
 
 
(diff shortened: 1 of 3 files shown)


src/gui/widgets/TimeDisplayWidget.cpp
```diff
@@ -34,20 +34,42 @@
...
+	m_majorValue.setAlignment(Qt::AlignRight);
+	m_majorValue.setStyleSheet(valueStyle);
+	m_minorLabel.setAlignment(Qt::AlignRight);
+	m_minorLabel.setStyleSheet(labelStyle);
+	m_minorValue.setAlignment(Qt::AlignRight);
+	m_minorValue.setStyleSheet(valueStyle);
+	m_milliSecondsLabel.setAlignment(Qt::AlignRight);
+	m_milliSecondsLabel.setStyleSheet(labelStyle);
+	m_milliSecondsValue.setAlignment(Qt::AlignRight);
+	m_milliSecondsValue.setStyleSheet(valueStyle);
```
winniehell: This repetition deserves its own widget.

src/gui/widgets/TimeDisplayWidget.cpp
```diff
@@ -34,20 +34,42 @@
...
-
-	setMaximumHeight( 32 );
+	m_spinBoxesLayout.addWidget( &m_majorLabel, 0, 0 );
+	m_spinBoxesLayout.addWidget( &m_majorValue, 1, 0 );
+	m_spinBoxesLayout.addWidget( &m_minorLabel, 0, 1 );
+	m_spinBoxesLayout.addWidget( &m_minorValue, 1, 1 );
+	m_spinBoxesLayout.addWidget( &m_milliSecondsLabel, 0, 2 );
+	m_spinBoxesLayout.addWidget( &m_milliSecondsValue, 1, 2 );
+
+	this->setStyleSheet( "TimeDisplayWidget { border-radius: 4px; background-color: rgba(0,0,0,70); margin: 8px; padding: 4px; }" );
```
winniehell: This should go in the stylesheet.

src/gui/widgets/TimeDisplayWidget.cpp
```diff
@@ -34,20 +34,42 @@
...
+
+	QString labelStyle = "font-size: 10px;";
+	QString valueStyle = "color: #0bd556; font-size: 24px;";
```
winniehell: This should go in the stylesheet.

qnebra:
In my opinion numbers should have same style, now your PR looks quite incoherently.

winniehell:
@qnebra Yes, I agree! :+1: I will create another pull request to TEMPO/BPM and TIME SIG consistent after this design got approved.

winniehell:
@Umcaruje given that you are the UI/UX maintainer, can I get your feedback here?

Umcaruje:
In my opinion, the number size is a bit too large, especially if you compare it to the buttons on the left, so I would make it a bit smaller, something around 20px should be good

> I will create another pull request to TEMPO/BPM and TIME SIG consistent after this design got approved.

I would suggest you do this in this PR.

Also, I really liked the idea of either making the unused zeroes in the minutes darker, that you posted somewhere else.

Also I would return the CPU and waveform meter to its previous place. 

All in all, this looks more modern and I'm all for it, but the tempo and time signature should be integrated into the actual design, as this looks wildly inconsistent at the moment

winniehell:
@Umcaruje Thank you for your feedback! :+1: 

> something around 20px should be good

sure, I will give that a try.

> I would suggest you do this in this PR.

My intention was to keep this one as small as possible code-wise. However I trust your judgement and will update the other two components as well.

> I really liked the idea of either making the unused zeroes in the minutes darker

Would you restrict that to minutes (and bars) only or should I apply that to all segments?

> Also I would return the CPU and waveform meter to its previous place.

Thank you for catching this! :heart: This was definitely not intended but I didn't notice. :see_no_evil: I will revert that change.

Umcaruje:
>My intention was to keep this one as small as possible code-wise.

Yeah but in terms of design it will be much easier to fine tune it if it's in the same PR.

>Would you restrict that to minutes (and bars) only or should I apply that to all segments?

Only minutes and bars, it wouldn't make sense on the other values.

PhysSong:
It seems like there's another work #5261 in progress.

winniehell:
yep, thank you @PhysSong. I'm closing this one.

PR #5121, opened 2019-08-12
Cmake/install refactor

Reflexe:


Reflexe:
I don't get it; for some strange reason cmake tries to write each `filelists.txt` 9 times with appveyor.

cmake/install/InstallTargetDependencies.cmake
```diff
@@ -0,0 +1,74 @@
...
+    FOREACH(LIB_PATH ${DEPS_LIB_DIRS})
+        FOREACH(suffix ${DEPS_LIB_DIRS_SUFFIXES})
+            list(APPEND DEPS_LIB_DIRS "${LIB_PATH}/${suffix}")
+        ENDFOREACH()
+    ENDFOREACH()
+
+    # Create the list of files using file(GENERATE)
+    # generator expressions on install(CODE) are not supported on old
+    # cmake versions.
+    SET(DEPLOY_LIST_FILE "${CMAKE_CURRENT_BINARY_DIR}/${DEPS_NAME}_filelist_${CMAKE_BUILD_TYPE}.txt")
```
PhysSong: > it probably has something to do with those generator expressions.
```suggestion
    SET(DEPLOY_LIST_FILE "${CMAKE_CURRENT_BINARY_DIR}/${DEPS_NAME}_filelist_$<CONFIG>.txt")
```

src/CMakeLists.txt
```diff
@@ -186,210 +186,18 @@ FOREACH(LIB ${LMMS_REQUIRED_LIBS})
...
-					DESTINATION .)
-		ENDIF()
-
-	ELSE(LMMS_BUILD_WIN32)
-		IF(NOT LMMS_BUILD_APPLE)
-			SET_TARGET_PROPERTIES(lmms PROPERTIES LINK_FLAGS "${LINK_FLAGS} -Wl,-E")
-		ENDIF(NOT LMMS_BUILD_APPLE)
-
-		INSTALL(TARGETS lmms RUNTIME DESTINATION "${BIN_DIR}")
-		INSTALL(FILES "${CMAKE_BINARY_DIR}/lmms.1.gz" DESTINATION "${CMAKE_INSTALL_PREFIX}/share/man/man1/" PERMISSIONS OWNER_READ GROUP_READ WORLD_READ)
```
PhysSong: This line should be restored and optionally moved.

cmake/install/InstallTargetDependencies.cmake
PhysSong: > I don't get it; for some strange reason cmake tries to write each `filelists.txt` 9 times with appveyor.

Overwriting the file isn't a problem. The real problem is the content differs in Release and Debug configuration.
https://github.com/LMMS/lmms/blob/ce5c74432e83f0eaa617cc26332f4209bc4efc7f/cmake/modules/BuildPlugin.cmake#L75-L78
The `SET_TARGET_PROPERTIES` shouldn't check for `STRIP`.

cmake/install/CMakeLists.txt
```diff
@@ -41,58 +6,14 @@ ENDIF()
...
-			get_filename_component(lib_dir ${lib} PATH)
-			list(APPEND LIB_DIRS ${lib_dir})
-			if(lib MATCHES ".(lib|dll.a)$")
-				if(IS_DIRECTORY ${lib_dir}/../bin)
-					list(APPEND LIB_DIRS ${lib_dir}/../bin)
-				endif()
-				if(IS_DIRECTORY ${lib_dir}/bin)
-					list(APPEND LIB_DIRS ${lib_dir}/bin)
-				endif()
-			endif()
```
PhysSong: This trick should be moved to `cmake/install/InstallTargetDependencies.cmake`, after line 32. It was here to support non-standard locations, ex. `libwinpthread-1.dll` in Travis-CI builds.

Reflexe: I tried to solve that with the PREFIXES argument. However, I don't know why it doesn't work.

PhysSong: Have you tried adding `../bin` to the `SUFFIXES` argument?

cmake/modules/DefineInstallVar.cmake
```diff
@@ -0,0 +1,24 @@
...
+    set(oneValueArgs NAME )
+    set(multiValueArgs CONTENT)
+    cmake_parse_arguments(VAR "${options}" "${oneValueArgs}"
+            "${multiValueArgs}" ${ARGN} )
+
+    # install(CODE) does not support generator expression in ver<3.14
+    if(VAR_GENERATOR_EXPRESSION AND ${CMAKE_VERSION} VERSION_LESS "3.14.0")
+        include(CreateTempFilePath)
+        CreateTempFilePath(OUTPUT_VAR file_path)
+        file(GENERATE OUTPUT "${file_path}" CONTENT "${VAR_CONTENT}")
```
PhysSong: The counter logic doesn't work because the CMake script is evaluated only once. For the MSVC generator, `file(GENERATE)` itself executed once, but the content is evaluated multiple times at the *generation time*.

PhysSong: NVM, I misunderstood the purpose of it.

cmake/modules/DefineInstallVar.cmake
```diff
@@ -13,12 +13,12 @@ function(DEFINE_INSTALL_VAR)
             "${multiValueArgs}" ${ARGN} )
 
     # install(CODE) does not support generator expression in ver<3.14
```
PhysSong: On a side note, `CMP0087` should be set to use generator expressions in `CMake >= 3.14`.

PhysSong:
Regarding the MSVC failure: I debugged the generation logic and got values from different configurations.
`RemoteVstPlugin`:
```
C:/projects/lmms/build/plugins/32/RemoteVstPlugin32.exe
C:/projects/lmms/build/plugins/32/RemoteVstPlugin32.exe
C:/projects/lmms/build/plugins/32/RemoteVstPlugin32.exe
C:/projects/lmms/build/plugins/32/RemoteVstPlugin32.exe
```
LMMS binary:
```
C:/projects/lmms/build/Debug/lmms.exe
C:/projects/lmms/build/MinSizeRel/lmms.exe
C:/projects/lmms/build/Release/lmms.exe
C:/projects/lmms/build/RelWithDebInfo/lmms.exe
```
Other plugins:
```
C:/projects/lmms/build/plugins/Debug/audiofileprocessor.dll;(omitted)
C:/projects/lmms/build/plugins/MinSizeRel/audiofileprocessor.dll;(omitted)
C:/projects/lmms/build/plugins/Release/audiofileprocessor.dll;(omitted)
C:/projects/lmms/build/plugins/RelWithDebInfo/audiofileprocessor.dll;(omitted)
```

I think we can generate with `$<CONFIG>` and read with `${CMAKE_BUILD_TYPE}` when using multi-config generators like the MSVC generator.

cmake/modules/DefineInstallVar.cmake
```diff
@@ -0,0 +1,25 @@
...
+            "${multiValueArgs}" ${ARGN} )
+
+    # install(CODE) does not support generator expression in ver<3.14
+    if(VAR_GENERATOR_EXPRESSION AND ${CMAKE_VERSION} VERSION_LESS "3.14.0")
+        include(CreateTempFilePath)
+        CreateTempFilePath(OUTPUT_VAR file_path)
+        file(GENERATE OUTPUT "${file_path}" CONTENT "${VAR_CONTENT}")
+        install(CODE "file(READ ${file_path} \"${VAR_NAME}\")")
+    else()
+        cmake_policy(SET CMP<0087> NEW)
```
PhysSong:
```suggestion
        cmake_policy(SET CMP0087 NEW)
```
Also, why did you use spaces instead of tabs in some new files?

PhysSong:
Due to the GitHub outage, this PR is moved to #5142.

cmake/modules/DefineInstallVar.cmake
PhysSong: However, I still wonder: Can we create files with unique names(more readable if possible) without a counter?

PR #5123, opened 2019-08-14
Center vertical scroll position when opening the Automation Editor

Veratil:
When you open the Automation Editor it centers at the exact center of the vertical scroll range. This can be very annoying when you have something that has a large range, such as the tempo (10-999).

This PR makes it so that the Automation Editor centers to the first node in the `timeMap`.

As a small extra, I noticed my console was getting flooded with `not in range` messages when the lines cannot be drawn visibly on the editor. This blocks it out with a preprocessor guard checking `LMMS_DEBUG`.

### DIFF
--- a/src/gui/editors/AutomationEditor.cpp
+++ b/src/gui/editors/AutomationEditor.cpp
@@ -323,7 +323,7 @@ void AutomationEditor::updateAfterPatternChange()
 	m_minLevel = m_pattern->firstObject()->minValue<float>();
 	m_maxLevel = m_pattern->firstObject()->maxValue<float>();
 	m_step = m_pattern->firstObject()->step<float>();
-	m_scrollLevel = ( m_minLevel + m_maxLevel ) / 2;
+	centerTopBottomScroll();
 
 	m_tensionModel->setValue( m_pattern->getTension() );
 
@@ -1587,12 +1587,36 @@ void AutomationEditor::drawLevelTick(QPainter & p, int tick, float value)
 
 		p.fillRect( x, y_start, rect_width, rect_height, currentColor );
 	}
-
+#ifdef LMMS_DEBUG
 	else
 	{
 		printf("not in range\n");
 	}
+#endif
+}
 
+
+
+
+// center the vertical scroll position on the first object's value
+void AutomationEditor::centerTopBottomScroll()
+{
+	// default to the m_scrollLevel position
+	int pos = (int) m_scrollLevel;
+	// If a pattern exists...
+	if (m_pattern) {
+		// get time map of current pattern
+		timeMap & time_map = m_pattern->getTimeMap();
+		// If time_map is not empty...
+		if ( !time_map.empty() ) {
+			// get the first object
+			timeMap::iterator it = time_map.begin();
+			// set the position to the inverted value ((max + min) - value)
+			// If we set just (max - value), we're off by m_pattern's minimum
+			pos = m_pattern->getMax() + m_pattern->getMin() - (int) it.value();
+		}
+	}
+	m_topBottomScroll->setValue( pos );
 }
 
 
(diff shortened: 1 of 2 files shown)


src/gui/editors/AutomationEditor.cpp
```diff
@@ -1587,12 +1587,36 @@ void AutomationEditor::drawLevelTick(QPainter & p, int tick, float value)
...
+}
+
+
+
 
+// center the vertical scroll position on the first object's value
+void AutomationEditor::centerTopBottomScroll()
+{
+	// default to the m_scrollLevel position
+	int pos = (int) m_scrollLevel;
```
BaraMGB: I think we should avoid c-style castings.

src/gui/editors/AutomationEditor.cpp
```diff
@@ -1587,12 +1587,36 @@ void AutomationEditor::drawLevelTick(QPainter & p, int tick, float value)
...
+	// If a pattern exists...
+	if (m_pattern) {
```
BaraMGB: The brace should be in the next line.

src/gui/editors/AutomationEditor.cpp
```diff
@@ -1587,12 +1587,36 @@ void AutomationEditor::drawLevelTick(QPainter & p, int tick, float value)
...
+		// get time map of current pattern
+		timeMap & time_map = m_pattern->getTimeMap();
+		// If time_map is not empty...
+		if ( !time_map.empty() ) {
```
BaraMGB: Same here.

src/gui/editors/AutomationEditor.cpp
```diff
@@ -1587,12 +1587,36 @@ void AutomationEditor::drawLevelTick(QPainter & p, int tick, float value)
...
+			// get the first object
+			timeMap::iterator it = time_map.begin();
+			// set the position to the inverted value ((max + min) - value)
+			// If we set just (max - value), we're off by m_pattern's minimum
+			pos = m_pattern->getMax() + m_pattern->getMin() - (int) it.value();
```
BaraMGB: Couldn't you use timeMap.begin() directly instead of the iterator?

BaraMGB: Also we should avoid spaces in braces for new code. Have a look at our coding convention.

BaraMGB (changes requested):

src/gui/editors/AutomationEditor.cpp
Veratil: I copied the previous usage within the code, but I can certainly update it to C++ style casting.

src/gui/editors/AutomationEditor.cpp
Veratil: Oof, that's on me. Will update and the other.

src/gui/editors/AutomationEditor.cpp
Veratil: > Couldn't you use timeMap.begin() directly instead of the iterator?

Ah, you're right. I was copying other code for this. I'll fix it.

> Also we should avoid spaces in braces for new code. Have a look at our coding convention.

I've been copying the style currently there thinking it's what's preferred, and I see spaces in the code all over. I assume this is a new convention going forward? Is there a need to go through and update style?

BaraMGB: In the past we used spaces. It's common sense that we change it when we touch the code. But don't touch code if it's not necessary.

BaraMGB (approved):

Spekular (commented):
I would rename the function to centerVerticalScroll, but LGTM otherwise.

Veratil:
> I would rename the function to centerVerticalScroll, but LGTM otherwise.

I named it to be easily seen as linked with `m_topBottomScroll`

josh-audio (approved):

Veratil:
3 approvals. Merging! :)

PR #5127, opened 2019-08-16
Fix stuck keys when dragging on piano view

### LINKED ISSUE #5028: Dragging up/down vertical piano results in stuck notes.
image

Steps to reproduce:
1. Create a piano pattern on the song editor or beat/bassline editor.
2. Drag mouse down the piano keys.

Expected behavior: Each key presses and releases as I go up and down the piano.

Actual result: Multiple piano keys graphically remain pressed (the keys do not audibly stick)

DomClark:
Fixes #5028.

It is possible for `NotePlayHandle::noteOff` to be called before `NotePlayHandle::play` is called for the first time, which results in a note-on message being sent without a subsequent note-off message. This PR checks whether the note has already been released before sending the note-on message.

### DIFF
--- a/src/core/NotePlayHandle.cpp
+++ b/src/core/NotePlayHandle.cpp
@@ -199,7 +199,7 @@ void NotePlayHandle::play( sampleFrame * _working_buffer )
 
 	lock();
 
-	if( m_totalFramesPlayed == 0 && !m_hasMidiNote
+	if( m_totalFramesPlayed == 0 && !m_hasMidiNote && !m_released
 		&& ( hasParent() || ! m_instrumentTrack->isArpeggioEnabled() ) )
 	{
 		m_hasMidiNote = true;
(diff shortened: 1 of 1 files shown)


src/core/NotePlayHandle.cpp
```diff
@@ -199,7 +199,7 @@ void NotePlayHandle::play( sampleFrame * _working_buffer )
 
 	lock();
 
-	if( m_totalFramesPlayed == 0 && !m_hasMidiNote
+	if( m_totalFramesPlayed == 0 && !m_hasMidiNote && !m_released
```
Reflexe:
```suggestion
	/* It is possible for NotePlayHandle::noteOff to be called before NotePlayHandle::play.  
	 * which results in a note-on message being sent without a subsequent note-off message. 
	 * Therefore, we check here whether the note has already been released before sending 
	 * the note-on message. */
	if( !m_released 
		&& m_totalFramesPlayed == 0 && !m_hasMidiNote
```

zonkmachine:
Tested, works!

 Merge?

PR #5133, opened 2019-08-19
Add close button to SideBarWidget

Veratil:
My take on #3682.

I feel like there could be a little more refactoring for finding the button/widget, but this works for now.

### DIFF
--- a/src/gui/widgets/SideBar.cpp
+++ b/src/gui/widgets/SideBar.cpp
@@ -121,6 +121,28 @@ void SideBar::appendTab( SideBarWidget *widget )
 	widget->setMinimumWidth( 200 );
 
 	ToolTip::add( button, widget->title() );
+
+	connect(widget, SIGNAL(closeButtonClicked(QWidget *)),
+		this, SLOT(toggleWidget(QWidget *)));
+}
+
+
+
+
+void SideBar::toggleWidget(QWidget * widget)
+{
+	for(auto it = m_widgets.begin(); it != m_widgets.end(); ++it)
+	{
+		QToolButton *curBtn = it.key();
+		QWidget *curWidget = it.value();
+
+		if(curWidget == widget)
+		{
+			curWidget->hide();
+			curBtn->setChecked(false);
+			curBtn->setToolButtonStyle(Qt::ToolButtonIconOnly);
+		}
+	}
 }
 
 
--- a/src/gui/widgets/SideBar.cpp
+++ b/src/gui/widgets/SideBar.cpp
@@ -129,25 +129,6 @@ void SideBar::appendTab( SideBarWidget *widget )
 
 
 
-void SideBar::toggleWidget(QWidget * widget)
-{
-	for(auto it = m_widgets.begin(); it != m_widgets.end(); ++it)
-	{
-		QToolButton *curBtn = it.key();
-		QWidget *curWidget = it.value();
-
-		if(curWidget == widget)
-		{
-			curWidget->hide();
-			curBtn->setChecked(false);
-			curBtn->setToolButtonStyle(Qt::ToolButtonIconOnly);
-		}
-	}
-}
-
-
-
-
 void SideBar::toggleButton( QAbstractButton * button )
 {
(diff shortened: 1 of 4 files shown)


src/gui/widgets/SideBarWidget.cpp
```diff
@@ -39,6 +39,10 @@ SideBarWidget::SideBarWidget( const QString & _title, const QPixmap & _icon,
 	m_layout = new QVBoxLayout( m_contents );
 	m_layout->setSpacing( 5 );
 	m_layout->setMargin( 0 );
+	m_closeBtn = new QToolButton(this);
+	m_closeBtn->setText(QString("X"));
```
PhysSong: An `X`-shaped icon will work better than a literal `X`. See this example(`src/gui/SubWindow.cpp`):
https://github.com/LMMS/lmms/blob/e1ef5fad0d5d8ce9b74b0c970c128cb90d695783/src/gui/SubWindow.cpp#L56-L57

Veratil: Ah! I was curious where to get that from. I couldn't find it for some reason. I'll fix that up.

src/gui/widgets/SideBarWidget.cpp
```diff
@@ -39,6 +39,10 @@ SideBarWidget::SideBarWidget( const QString & _title, const QPixmap & _icon,
...
+	connect(m_closeBtn, SIGNAL(released()),
+		this, SLOT(closeButtonClick()));
```
PhysSong: You can connect a signal directly to another signal without a helper slot.

Veratil: I'm still learning Qt, and saw that it was signal->slot, but that's good to know! 😄

PhysSong:
The best way I can think of is:
- Connect the `released()` signal of the close button to a lambda function
- The lambda function hides the widget and emits the `closeButtonClicked()` signal
- `closeButtonClicked()` is connected to corresponding `SideBarButton` instead of `SideBar` itself if possible

src/gui/widgets/SideBar.cpp
```diff
@@ -121,6 +121,28 @@ void SideBar::appendTab( SideBarWidget *widget )
...
 	ToolTip::add( button, widget->title() );
+
+	connect(widget, &SideBarWidget::closeButtonClicked,
+		[=]() { button->click(); });
+}
+
+
+
+
+void SideBar::toggleWidget(QWidget * widget)
```
PhysSong: Could you remove this function as it's no longer needed?

src/gui/widgets/SideBarWidget.cpp
```diff
@@ -22,23 +22,34 @@
...
 {
 	m_contents = new QWidget( this );
 	m_layout = new QVBoxLayout( m_contents );
 	m_layout->setSpacing( 5 );
 	m_layout->setMargin( 0 );
+	m_closeBtn = new QPushButton(embed::getIconPixmap("close"), QString(), this);
+	m_closeBtn->resize(m_buttonSize);
+	m_closeBtn->setFocusPolicy(Qt::NoFocus);
+	m_closeBtn->setCursor(Qt::ArrowCursor);
+	m_closeBtn->setAttribute(Qt::WA_NoMousePropagation);
```
PhysSong: Are there any reasons for setting `Qt::NoFocus` and `Qt::WA_NoMousePropagation`?

PhysSong: It seems like `Qt::NoFocus` and `Qt::ArrowCursor` are set by default. I think the `SubWindow` needs `Qt::WA_NoMousePropagation` because it's undesirable to propagate mouse events of the button to the parent subwindow. It's fine to either keep or drop the line because mouse events on `SidebarWidget` are no-ops.

Veratil: I'll remove them for simplicity.

PhysSong (approved):
Looks good and works well.

PR #5134, opened 2019-08-19
Allow playing muted pattern in the piano roll

Veratil:
An annoyance to me so far has been wanting to listen to a pattern I just opened from the Song Editor, but I've muted it beforehand for whatever reason. I am then required to click back to the Song Editor to unmute the pattern (I don't know if there's a shortcut key to unmute the pattern from the Piano Roll window) and then I can go back to the Piano Roll to listen to it.

This should allow the Piano Roll (and I believe any other editor which uses `InstrumentTrack`) to play its pattern while being muted on the Song Editor.

My next step I want to be able to do is play while the entire track is muted, since this PR doesn't fix that instance.

### DIFF
--- a/src/tracks/InstrumentTrack.cpp
+++ b/src/tracks/InstrumentTrack.cpp
@@ -653,7 +653,9 @@ bool InstrumentTrack::play( const MidiTime & _start, const fpp_t _frames,
 	{
 		Pattern* p = dynamic_cast<Pattern*>( *it );
 		// everything which is not a pattern or muted won't be played
-		if( p == NULL || ( *it )->isMuted() )
+		if(p == NULL ||
+			(Engine::getSong()->playMode() == Song::Mode_PlaySong
+			&& (*it)->isMuted()))
 		{
 			continue;
 		}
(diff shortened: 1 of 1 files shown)


src/tracks/InstrumentTrack.cpp
```diff
@@ -653,7 +653,9 @@ bool InstrumentTrack::play( const MidiTime & _start, const fpp_t _frames,
 	{
 		Pattern* p = dynamic_cast<Pattern*>( *it );
 		// everything which is not a pattern or muted won't be played
-		if( p == NULL || ( *it )->isMuted() )
+		if(p == NULL ||
+			(Engine::getSong()->playMode() == Song::Mode_PlaySong
```
PhysSong: It will make muted patterns in a Beat/Bassline pattern play sound when playing the BB pattern.
```suggestion
			(Engine::getSong()->playMode() != Song::Mode_PlayPattern
```
Also, could you add a comment describing this behavior?

Veratil: I think allowing a BB Editor pattern to be played as well is a good change, too, then. There is no indication that I can see that tells you the pattern is muted, so behavior of pressing play in their own window should be "I hear sound" rather than "I hear nothing". Ideally I believe the only time you shouldn't hear the pattern play is from the Song Editor, any other place you should allow it to play.

Spekular:
I think PhysSong means that tracks muted *inside the bb editor* will still
play, which is obviously undesireable.

Otherwise you could have a kick/snare loop you're previewing in the BB
editor, mute the snare track, and have the snare continue to play.

Veratil:
> 
> I think PhysSong means that tracks muted *inside the bb editor* will still play, which is obviously undesireable. Otherwise you could have a kick/snare loop you're previewing in the BB editor, mute the snare track, and have the snare continue to play.

Ah! That's right. I just tested and that does indeed ignore the mute in BB Editor. Will fix. 👍

PR #5137, opened 2019-08-21
Fix #4188 - Check for failed std::find

### LINKED ISSUE #4188: Mark/unmark corresponding octave semitones crash (LMMS 1.2.0-RC5)
**Steps to reproduce**

- open piano roll
- mark any note on the piano roll keyboard with RMB "mark/unmark current semitone"
- on the same note use RMB "mark/unmark corresponding octave semitones"
- LMMS crashes

Above doesn't occur while using "mark/unmark corresponding octave semitones" on note selected using "mark/unmark corresponding octave semitones" option.

OS: Windows 7 (64)

Veratil:
I noticed this crashed on `stable-1.2` and then in `master`.

Crash happens when trying to erase a note that doesn't exist in `m_markedSemiTones` when selecting "Mark/unmark all corresponding octave semitones", so the solution is just to check for a failed `std::find`.

### DIFF
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -549,7 +549,10 @@ void PianoRoll::markSemiTone( int i )
 				for (int ix = 0; ix < aok.size(); ++ix)
 				{
 					i = std::find(m_markedSemiTones.begin(), m_markedSemiTones.end(), aok.at(ix));
-					m_markedSemiTones.erase(i);
+					if (i != m_markedSemiTones.end())
+					{
+						m_markedSemiTones.erase(i);
+					}
 				}
 			}
 			else


Reflexe:
@PhysSong I'm not sure, Do we have to target it for stable and sync it later to master? Or just do it manually now?

Reflexe (approved):

PhysSong:
@Reflexe Since #4188 is in the 1.2.1 milestone, the commit should somehow go to `stable-1.2`. The specific method doesn't matter too much, but I think it'd be better to merge into `stable-1.2` first.
This should be tested before we merge this, of course.

Reflexe:
Since we've changed `qFind` to `std::find` between stable-2 and master, I;ve created another pr for stable: https://github.com/LMMS/lmms/pull/5184

PR #5144, opened 2019-08-22
Fix crash opening VST automation window

### LINKED ISSUE #5132: Master hangs when opening VST automation window
Win 10 64 bit, happens regardless of which VST is loaded, but not in an empty VeSTige instance.

image

DomClark:
Fixes #5132.

The VST automation window initially adds a subwindow with a null widget, so we have to check that `w` isn't `nullptr` before getting its size hint.


Reflexe:
I'm just waiting for confirmation that the issue is not present in stable-1.2 too (I'll check it later this day)

PR #5147, opened 2019-08-24
Synchro, a simple phase modulation synth

rubiefawn:
2-oscillator PM synth for creating nasty sounds.

## ⚠️ Todo
- [ ] Add non-automatable UI controls to change oversampling rate
- [ ] Use the HIIR library for proper bandlimiting
- [ ] Proper monophonic behavior. See #6516
  * [ ] Amplitude envelopes
  * [ ] Glide/portamento
- [ ] Style knobs properly

## ❓ Optional
- [ ] Improve performance. Possible pain points:
  * [ ] Heavy use of `exp()` and `pow()` during waveform generation
  * [ ] The halves of the oversampling double buffer are not guaranteed to be contiguously allocated, which might cause cache issues.
- [ ] Use source `.svg` for background instead of `.png`, whenever that becomes possible
- [ ] Pitch envelopes
- [ ] Modulation envelope

## ❌ Issues
- Line width for waveform preview is too thin. Ideally it would be 2px wide, but `gui::Graph` doesn't have any logic for line width
- There's currently a bug where the modulator will reset its own phase every time the carrier completes a cycle, although the shape of the modulator in each of those windows is correct. For example, if the modulator is 1 octave below the carrier, it should complete a cycle once per every two carrier cycles, but instead it just plays the first half of its cycle twice.


BaraMGB:
At first glance you could add an argument in the constructor of the Graph class: 
`Graph::Graph( QWidget * _parent, graphStyle _style, int _width,
		int _height, bool readOnly=false )`

In the mousePressEvent() you could have an if-statement and test this readOnly-bool.

BaraMGB:
I love it! It's so simple but you can make very heavy neuro basses with it!

DomClark:
`Graph` is a subclass of `QWidget`, so you can use [`QWidget`'s `enabled` property](https://doc.qt.io/qt-5/qwidget.html#enabled-prop). Access it with `isEnabled()` and set it with `setEnabled(bool)`.

DomClark:
Since you're no longer modifying the graph code, could you revert the formatting changes there?

plugins/Synchro/SynchroSynth.h
```diff
@@ -0,0 +1,163 @@
...
+	FloatModel m_carrierDrive;
+	FloatModel m_carrierSync;
+	FloatModel m_carrierChop;
+	FloatModel m_modulatorDetune;
+	FloatModel m_modulatorDrive;
+	FloatModel m_modulatorSync;
+	FloatModel m_modulatorChop;
+	graphModel m_carrierView;
+	graphModel m_modulatorView;
+	graphModel m_resultView;
```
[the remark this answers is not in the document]
iansannar: Compliant as of 4bcfe82688b8228dda9401d491fb963471fe75f1. If this is not mandatory I would like to revert this change for readability.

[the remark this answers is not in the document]
PhysSong: What I meant was `graphModel` is for storing and manipulating the data, and `Graph` is the class for GUI interactions.

plugins/Synchro/SynchroSynth.h
```diff
@@ -0,0 +1,169 @@
...
+	float Sync;
+	float Chop;
+} SynchroOscillatorSettings;
+
+class SynchroNote
+{
+	MM_OPERATORS
+public:
+	SynchroNote(NotePlayHandle * nph);
+	virtual ~SynchroNote();
```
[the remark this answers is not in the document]
iansannar: Gotcha 👍. This was built off of Bit Invader as a base, so there are a lot of leftovers from that. I'll make sure to clean it up.

plugins/Synchro/SynchroSynth.h
```diff
@@ -0,0 +1,169 @@
...
+	void nextStringSample(sampleFrame &outputSample, sample_rate_t sample_rate,
+		float modulationStrength, float modulationAmount, float harmonics,
+		SynchroOscillatorSettings carrier,
+		SynchroOscillatorSettings modulator);
+private:
+	NotePlayHandle * nph;
+	float sample_index[2] = {0, 0}; //Index 0 is carrier, index 1 is modulator
```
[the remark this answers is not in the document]
iansannar: Splitting and renaming that array is a great idea. I originally designed it that way in the event of even more oscillators/unison, but I don't think those would be useful features currently. Will fix 👍.

plugins/Synchro/SynchroSynth.h
```diff
@@ -0,0 +1,169 @@
...
+};
+
+class SynchroSynth : public Instrument
+{
+	Q_OBJECT
+public:
+	SynchroSynth(InstrumentTrack * instrument_track);
+	virtual ~SynchroSynth();
+	virtual void playNote(NotePlayHandle * n, sampleFrame * working_buffer);
```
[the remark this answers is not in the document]
iansannar: More Bit Invader leftovers. Will fix.

iansannar:
> I'm not quite sure of the advantage of having both "mod" and "mod max" controls. They're functionally identical, their values just being multiplied together. Is this a common thing to have? Can't you get the same result just by not turning "mod" up as high?

Yes, but it's harder. Compare it to pitch bend vs pitch bend range. It is slightly redundant, but it increases ease of use, especially in automation.

> The "sync" control just seems to be a frequency multiplier. From what I've seen, sync usually means one oscillator resets another oscillator every cycle. Is the current behaviour and wording what is intended?

The current behavior is exactly as intended, but you're right; the wording is a bit off. I'm not sure what to call it exactly, but "sync" seemed closest to the functionality. I am open to suggestions here.

plugins/Synchro/SynchroSynth.h
```diff
@@ -0,0 +1,169 @@
...
+	float m_modulatorNormalize;
+	friend class SynchroSynthView;
+};
+
+class SynchroSynthView : public InstrumentView
+{
+	Q_OBJECT
+public:
+	SynchroSynthView(Instrument * instrument, QWidget * parent);
+	virtual ~SynchroSynthView() {};
```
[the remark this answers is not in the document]
iansannar: 👾

plugins/Synchro/SynchroSynth.cpp
```diff
@@ -0,0 +1,389 @@
...
+void SynchroNote::nextStringSample(sampleFrame &outputSample,
+	sample_rate_t sample_rate, float modulationStrength,
+	float modulationAmount, float harmonics, SynchroOscillatorSettings carrier,
+	SynchroOscillatorSettings modulator)
+{
+	float sampleRatePi = F_2PI / sample_rate; //For oversampling
+	//Find position in modulator waveform
+	sample_index[1] += sampleRatePi * DetuneOctaves(nph->frequency(),
+		modulator.Detune);
+	while (sample_index[1] >= F_2PI) { sample_index[1] -= F_2PI; } //Make sure phase is always between 0 and 2PI
```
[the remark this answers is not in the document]
iansannar: Is `fmod` faster? I'm open to using that instead. I think that would maintain readability while making things more concise.

src/gui/widgets/Graph.cpp
```diff
@@ -747,4 +747,3 @@ void graphModel::drawSampleAt( int x, float val )
 
 
 
-
```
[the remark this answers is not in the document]
iansannar: I can revert this, but are you sure you don't want some extra whitespace trimmed? 😄

plugins/Synchro/SynchroSynth.cpp
```diff
@@ -0,0 +1,389 @@
...
+{
+	//Empty destructor
+}
+
+void SynchroNote::nextStringSample(sampleFrame &outputSample,
+	sample_rate_t sample_rate, float modulationStrength,
+	float modulationAmount, float harmonics, SynchroOscillatorSettings carrier,
+	SynchroOscillatorSettings modulator)
+{
+	float sampleRatePi = F_2PI / sample_rate; //For oversampling
```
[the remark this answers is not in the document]
iansannar: > I feel that the variable name and comment here are confusing.

What do you suggest? I'm not fond of the name, either.

> It's a conversion factor from cycles per second to radians per sample, and not specifically related to oversampling.

True. I added that comment to help me remember what it was being used for as part of my copy-paste of oversampling code provided by @DouglasDGI. I'm sure I can remove the comment while maintaining readability if I rename the variable and comment other things better.

plugins/Synchro/SynchroSynth.h
```diff
@@ -0,0 +1,169 @@
...
+	Knob * m_carrierSyncKnob;
+	Knob * m_carrierChopKnob;
+	Knob * m_modulatorDetuneKnob;
+	Knob * m_modulatorDriveKnob;
+	Knob * m_modulatorSyncKnob;
+	Knob * m_modulatorChopKnob;
+	Graph * m_carrierGraph;
+	Graph * m_modulatorGraph;
+	Graph * m_resultGraph;
+	static QPixmap * s_artwork;
```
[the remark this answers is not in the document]
iansannar: 👾

plugins/Synchro/SynchroSynth.h
```diff
@@ -0,0 +1,169 @@
...
+};
+
+//Triangle waveform generator
+static inline float tri(float x)
```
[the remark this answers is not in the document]
iansannar: They can and will be moved to the cpp. 😄

plugins/Synchro/SynchroSynth.h
```diff
@@ -0,0 +1,169 @@
...
+	Q_OBJECT
+public:
+	SynchroSynth(InstrumentTrack * instrument_track);
+	virtual ~SynchroSynth();
+	virtual void playNote(NotePlayHandle * n, sampleFrame * working_buffer);
+	virtual void deleteNotePluginData(NotePlayHandle * n);
+	virtual void saveSettings(QDomDocument & doc, QDomElement & parent);
+	virtual void loadSettings(const QDomElement & thisElement);
+	virtual QString nodeName() const;
+	virtual f_cnt_t desiredReleaseFrames() const { return(64); } //Not sure what this is used for
```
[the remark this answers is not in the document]
iansannar: I haven't implemented any instrument-specific release functionality, and there shouldn't be any processing that requires this AFAIK. Would it be safe to remove altogether?

> Also, the coding conventions say not to use parentheses around return values.
👾, will fix.

plugins/Synchro/SynchroSynth.cpp
```diff
@@ -0,0 +1,389 @@
...
+#include "embed.h"
+#include "Engine.h"
+#include "InstrumentTrack.h"
+#include "lmms_constants.h"
+#include "Mixer.h"
+#include "NotePlayHandle.h"
+#include "plugin_export.h"
+
+#define SYNCRHO_VERSION "0.4"
+#define SYNCHRO_VOLUME_CONST 0.15f
```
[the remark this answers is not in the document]
iansannar: Yes. 👍

plugins/Synchro/SynchroSynth.cpp
```diff
@@ -0,0 +1,389 @@
...
+
+static const int SYNCHRO_OVERSAMPLE = 4; //Anti-aliasing samples
+static const int SYNCHRO_PM_CONST = 20; //Strength of the phase modulation
+static const int SYNCHRO_GRAPH_SAMPLES = 48; //Resolution of the waveform view GUI elements
```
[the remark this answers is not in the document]
iansannar: Will fix. After a quick google search, `constexpr` seems pretty neat.

plugins/Synchro/SynchroSynth.cpp
```diff
@@ -0,0 +1,389 @@
...
+
+SynchroNote::SynchroNote(NotePlayHandle * nph) :
+	nph(nph)
+{
+	//Empty constructor
+}
+
+SynchroNote::~SynchroNote()
+{
+	//Empty destructor
```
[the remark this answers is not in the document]
iansannar: 👾

plugins/Synchro/SynchroSynth.cpp
```diff
@@ -0,0 +1,389 @@
...
+	}
+	generalChanged();
+}
+
+void SynchroSynth::generalChanged()
+{
+	for (int i = 0; i < SYNCHRO_GRAPH_SAMPLES; ++i)
+	{
+		int octaveDiff = m_carrierDetune.value() - m_modulatorDetune.value();
+		float pitchDifference = powf(2, octaveDiff);
```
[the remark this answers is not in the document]
iansannar: These values can't be `const` at all, since they can (and are) changed by knob input. `octaveDiff` is only used once and can be inlined into `pitchDifference` for readability, though.

I may be able to factor them out of the loop, but only if I were to opt against making the oscillator octave controls sample-exact.

plugins/Synchro/SynchroSynth.cpp
```diff
@@ -0,0 +1,389 @@
...
+
+SynchroNote::~SynchroNote()
+{
+	//Empty destructor
+}
+
+void SynchroNote::nextStringSample(sampleFrame &outputSample,
+	sample_rate_t sample_rate, float modulationStrength,
+	float modulationAmount, float harmonics, SynchroOscillatorSettings carrier,
+	SynchroOscillatorSettings modulator)
```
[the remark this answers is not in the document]
iansannar: I'm not familiar with const reference. What would be the benefit?

plugins/Synchro/SynchroSynth.cpp
```diff
@@ -0,0 +1,389 @@
...
+}
+
+QString SynchroSynth::nodeName() const
+{
+	return synchro_plugin_descriptor.name;
+}
+
+void SynchroSynth::playNote(NotePlayHandle * n, sampleFrame * working_buffer)
+{
+	if (n->totalFramesPlayed() == 0 || n->m_pluginData == NULL)
```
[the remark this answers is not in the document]
iansannar: 👾

plugins/Synchro/SynchroSynth.h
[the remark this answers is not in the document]
DomClark: The `applyRelease` call here will use this value:
https://github.com/LMMS/lmms/blob/f8dfb32f2f6169bf5908b282047f5c8e10a8864b/plugins/Synchro/SynchroSynth.cpp#L240
This gives a 64-frame fade-out to zero once the note is released. You can try changing or removing the function, but setting it too small will probably cause clicks when the note is released. (The default is zero, which is almost certainly too small.)

plugins/Synchro/SynchroSynth.cpp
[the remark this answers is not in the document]
DomClark: There's a famous quote:

> There are only two hard things in Computer Science: cache invalidation and naming things.

Perhaps `freqToPhaseStep`? Or maybe `freqToSampleStep`? I'll let you decide.

plugins/Synchro/SynchroSynth.cpp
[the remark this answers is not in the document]
DomClark: It would be difficult to say without profiling it, but I don't think it will have a significant impact. `fmod` is inherently slower than subtraction, but the while loop may need to do multiple comparisons and subtractions, and is also subject to branch prediction. `fmod` will at least be more consistent in speed. I would be surprised if this were a bottleneck though.

One nice advantage to `fmod` is that if some disaster occurs and `sample_index` becomes infinite (or large enough that subtracting `F_2PI` makes no difference), then `fmod` will return `NaN` (or some value at least), whereas the while loop will hang. It might sound unlikely, but I've seen it happen a few times in otherwise correct code.

My main motivation for this suggestion though is that I think it's easier to read, and clearer that it's not going to misbehave. That might just be personal preference though, so I'll leave it up to you.

(One thing to remember is that `fmod` returns a negative value for negative inputs, so you'll need to write `if (sample_index[0] < 0) { sample_index[0] += F_2PI; }` if you choose to use it in the second instance lower down. This can be an if rather than a while though.)

plugins/Synchro/SynchroSynth.cpp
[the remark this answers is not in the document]
DomClark: Passing by value (`void foo(Bar bar)`)  requires that the original value be copied for the function to use, whereas const reference (`void foo(const Bar &bar)`) passes a reference to the original value instead, which is often faster (and saves a bit of stack space). The reference doesn't technically have to be const, but it's good practice to make things const unless you intend to modify them.

DouglasDGI:
About sync, I have seen things labeled as "sync" work this way (e.g. Serum).  You can think of it as the waveform being "synced" with a waveform with a frequency determined by the Sync knob.  In other words, the Sync knob multiplies the frequency of a fake waveform which is causing the real waveform to reset.

plugins/Synchro/SynchroSynth.cpp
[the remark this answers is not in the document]
DomClark: `const` only applies for the duration of the function's execution, so that wouldn't stop it getting a different value for the knob the next time the function was called. Since the variables referred to aren't modified during the function (once they've been factored out of the loop, that is), `const` is applicable.

Bear in mind that the graphs only show a snapshot of the waveform at each point in time. The code here will run on a scale of nanoseconds, but a sample of audio is on the scale of microseconds and a whole buffer of audio is on the scale of milliseconds. The samples generated here are along an "imaginary" time axis, not real time, and are generated on a different thread to the audio, so the question of how sample-exact automation corresponds to samples on the graph is pretty much meaningless. Your best bet is just to pull the value of the knobs at the start, and go off that.

DomClark:
> Yes, but it's harder. Compare it to pitch bend vs pitch bend range. It is slightly redundant, but it increases ease of use, especially in automation.

👍 

> The current behavior is exactly as intended, but you're right; the wording is a bit off. I'm not sure what to call it exactly, but "sync" seemed closest to the functionality. I am open to suggestions here.

OpulenZ and Watsyn call it "mul", with the tooltip "frequency multiplier". (At least I think those knobs are equivalent to this one, correct me if I'm wrong.)

> About sync, I have seen things labeled as "sync" work this way (e.g. Serum). You can think of it as the waveform being "synced" with a waveform with a frequency determined by the Sync knob. In other words, the Sync knob multiplies the frequency of a fake waveform which is causing the real waveform to reset.

But it's not playing part of the waveform at normal speed then resetting it, it's playing the whole thing but faster. Plus, I recall you mentioning that some of the other labels on Serum are less than accurate. 😉

DouglasDGI:
Unless the code was changed since the last time I looked at it (I have a version of Synchro from a bit before this PR was made), it isn't only multiplying the frequency, it also resets the frequency-multiplied waveform whenever the original waveform would have completed.  So with integer values of Sync, it will be a pure frequency multiplication, but with non-integer values, it won't.  So, for example, a Sync of 1.5 plays the full waveform, and then the first half of the waveform, all in the same time it would have normally taken it to play one wave with a sync of 0.

Likewise, "syncing" a waveform with another waveform one octave up would be identical to multiplying the frequency of that waveform by 2.

DouglasDGI:
Sorry for all the messups in the previous posts, I'm super scatterbrained right now.  I deleted some, now I will speak as if I actually have a brain.  =)

The Sync knob is a frequency multiplier, *but* that waveform is being synced with a hypothetical waveform of the original frequency.  So if you're playing a note at 440 Hz with a Sync value of 1.5, then the result is the waveform at 660 Hz being synced at 440 Hz.  With integer Sync values, this is identical to pure frequency multiplication, but with non-integer values, it creates a drastically different sound.

DomClark:
It doesn't look to me like the modulator gets synced back to the original frequency any more. Here's the only place where the modulator's phase is modified, and it's just moved forward by the adjusted frequency (mod 2pi), without being reset:
https://github.com/LMMS/lmms/blob/f8dfb32f2f6169bf5908b282047f5c8e10a8864b/plugins/Synchro/SynchroSynth.cpp#L86-L89
(I may well be wrong - I'm only getting this from the code. My last build of this PR is from before fractional sync was supported, so I can't test it right now.)

DouglasDGI:
I think the `while (sample_index[1] >= F_2PI) { sample_index[1] -= F_2PI; }` part would cause the phase reset.  It sets the phase to 0 whenever the original waveform (ignoring sync) reaches its end, which would interrupt the synced waveform at fractional frequency multiplications.

plugins/Synchro/SynchroSynth.cpp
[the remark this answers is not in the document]
iansannar: This is my bad; I thought this was in reference the audio synthesis code at first. This makes much more sense now that I see it's the graph.

iansannar:
The frequency multiplication happens in the waveform generation function, so the phase is automatically reset as part of that.

DomClark:
Ah yes, I see. I thought `sample_index[1]` was the phase after sync had been applied, but it's the phase before. Thanks for clearing that up.

PhysSong:
@iansannar I fixed the merge conflict for you. Please pull it before you go further.

src/gui/widgets/Graph.cpp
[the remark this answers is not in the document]
PhysSong: It's for cleaner commit history.

plugins/Synchro/SynchroSynth.h
```diff
@@ -73,6 +73,7 @@ class SynchroSynth : public Instrument
 	virtual QString nodeName() const;
 	virtual f_cnt_t desiredReleaseFrames() const { return(64); } //Not sure what this is used for
 	virtual PluginView * instantiateView(QWidget * parent);
+	virtual Flags flags() const {	return IsSingleStreamed; } //Disables default envelopes/LFOs
```
iansannar, line 76: Added this since I plan to add per-oscillator envelopes (exponential, too). However, since I disabled this, LMMS immediately crashes upon playing a note through Synchro. I would guess this is because there's no envelope to process the audio through.

PhysSong: It's because `working_buffer` is null in `playNote()` if you specify `IsSingleStreamed`. It's intended to make `playNote()` only check for the status for a note and render the sound in `play()`. See LB302 for how single-streamed instruments work.

iansannar: My intention is to disable the default envelopes as I intend to add my own to the instrument itself. Is there a proper way to do this then?

PhysSong: No, at least currently. Do you want to disable LFOs and the filter as well?

iansannar: Yes, though there are a few edge cases where they could be useful. I'll re-enable it for now until I get the custom envelopes worked out.

plugins/Synchro/SynchroSynth.cpp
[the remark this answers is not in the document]
iansannar: This part of the code will require a rewrite to accommodate sample-exact modulation. Commenting to remind myself when I do that.

PhysSong:
It seems like you removed the `SynchroNote` destructor accidentally. You should either restore it on the source file or add `= default` in the header file to fix the undefined reference error.

PhysSong:
To prevent compiler warnings, you should make the order of declaration and initialization consistent.

PhysSong:
@iansannar Do you want to add something more here? Or, can I start review?

iansannar:
> @iansannar Do you want to add something more here? Or, can I start review?
It is usable in its current form, so you can go ahead and review it. I may add more features later, but that won't be for awhile as I am busy with school and work.

plugins/Synchro/SynchroSynth.cpp
```diff
@@ -0,0 +1,565 @@
...
+		sampleFrame outputSample = {0, 0};
+		sampleFrame tempSample = {0, 0};
+		SynchroOscillatorSettings carrier = { m_carrierDetune.value(), m_carrierDrive.value(),
+			m_carrierSync.value(), m_carrierChop.value(),
+			m_carrierAttack.value(), m_carrierDecay.value(), m_carrierSustain.value(), m_carrierRelease.value() };
+		SynchroOscillatorSettings modulator = { m_modulatorDetune.value(), m_modulatorDrive.value(),
+			m_modulatorSync.value(), m_modulatorChop.value(),
+			m_modulatorAttack.value(), m_modulatorDecay.value(), m_modulatorSustain.value(), m_modulatorRelease.value() };
+
+		//Oversampling using linear interpolation
```
PhysSong, line 435: A side-note: this will reduce aliasing, but it will slightly filter the signal.

plugins/Synchro/SynchroSynth.h
```diff
@@ -0,0 +1,154 @@
...
+
+//Renders the sound for each note or something
+class SynchroNote
+{
+	MM_OPERATORS
+public:
+	//Constructor
+	SynchroNote(NotePlayHandle * nph);
+	//Renders a single sample of audio
+	void nextStringSample(sampleFrame &outputSample, sample_rate_t sample_rate,
```
PhysSong, line 62: I'm curious why the name of this function is `nextStringSample`.

plugins/Synchro/SynchroSynth.cpp
```diff
@@ -0,0 +1,565 @@
...
+	//The final sample Synchro uses for phase modulation
+	float pmSample = modulatorSample * modulatorEnvelope * SYNCHRO_PM_CONST;
+
+	//Find position in carrier waveform
+	carrierSampleIndex += freqToSampleStep * DetuneOctaves(nph->frequency(), carrier.Detune);
+	//Make sure phase is always between 0 and 2PI
+	while (carrierSampleIndex >= F_2PI) { carrierSampleIndex -= F_2PI; }
+	//Extra check since phase modulation can push it below 0
+	while (carrierSampleIndex < 0) { carrierSampleIndex += F_2PI; }
+	//Carrier envelope
```
PhysSong, line 183: It looks like the code fragment for the carrier envelope is similar to modulator's one, isn't it possible to reduce duplication?

plugins/Synchro/SynchroSynth.cpp
```diff
@@ -0,0 +1,565 @@
...
+	modulatorSampleIndex += freqToSampleStep * DetuneOctaves(nph->frequency(), modulator.Detune);
+	//Make sure phase is always between 0 and 2PI
+	while (modulatorSampleIndex >= F_2PI) { modulatorSampleIndex -= F_2PI; }
+
+	//Get modulator waveform at position
+	float modulatorSample = SynchroWaveform(modulatorSampleIndex,
+		modulator.Drive, modulator.Sync, modulator.Chop, harmonics) * (modulationStrength * modulationAmount);
+
+	//Modulator envelope
+	//Using sample_rate here gives an incorrect envelope length for some reason
```
PhysSong, line 149: You're passing `Engine::mixer()->processingSampleRate() * SYNCHRO_OVERSAMPLE` here, but you're comparing with `NoteTime` which is based on `Engine::mixer()->processingSampleRate()`.

plugins/Synchro/SynchroSynth.cpp
```diff
@@ -0,0 +1,565 @@
...
+
+}
+
+void SynchroNote::nextStringSample(sampleFrame &outputSample, sample_rate_t sample_rate,
+	const float & modulationStrength, const float & modulationAmount, const float & harmonics,
+	const SynchroOscillatorSettings & carrier, const SynchroOscillatorSettings & modulator)
+{
+	float freqToSampleStep = F_2PI / sample_rate;
+	//How long (in frames) this note has been playing
+	float NoteTime = nph->totalFramesPlayed();
```
PhysSong, line 136: Variable names should start with lowercase characters.

plugins/Synchro/SynchroSynth.h
```diff
@@ -0,0 +1,154 @@
...
+		const float & modulationStrength, const float & modulationAmount, const float & harmonics,
```
DomClark, line 63: No need to pass `float`s by reference - pass by value is fine.

plugins/Synchro/SynchroSynth.h
```diff
@@ -0,0 +1,154 @@
...
+		const SynchroOscillatorSettings & carrier, const SynchroOscillatorSettings & modulator);
+private:
+	NotePlayHandle * nph;
+	float carrierSampleIndex = 0; //The index (or phase) of the carrier oscillator
+	float modulatorSampleIndex = 0; //The index (or phase) of the modulator oscillator
```
DomClark, line 68: `m_` prefix for member variables.

plugins/Synchro/SynchroSynth.h
```diff
@@ -0,0 +1,154 @@
...
+//Synth GUI
+class SynchroSynthView : public InstrumentView
+{
+	Q_OBJECT
+public:
+	SynchroSynthView(Instrument * instrument, QWidget * parent);
+	QSize sizeHint() const override { return QSize(448, 250); } //~3px wider than the artwork
+protected slots:
+private:
+	virtual void modelChanged();
```
DomClark, line 80: When overriding methods, omit `virtual` and add `override` (`void modelChanged() override;`). There are a few other instances to look at in this file.

plugins/Synchro/SynchroSynth.cpp
```diff
@@ -0,0 +1,565 @@
...
+		}
+	}
+	//applyRelease(working_buffer, n);
+	instrumentTrack()->processAudioBuffer(working_buffer, frames + offset, n);
+}
+
+//Tells LMMS how much extra time Synchro needs at the end of a note to finish our envelopes
+f_cnt_t SynchroSynth::desiredReleaseFrames() const
+{
+	return (f_cnt_t)(m_carrierRelease.value() * 0.001 * Engine::mixer()->processingSampleRate());
```
DomClark, line 460: Prefer C++-style casts over C-style casts. (`(f_cnt_t)` -> `static_cast<f_cnt_t>`)

plugins/Synchro/SynchroSynth.cpp
```diff
@@ -0,0 +1,565 @@
...
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this program (see COPYING); if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA 02110-1301 USA.
+ */
+
+#include "SynchroSynth.h"
+//Standard headers
+#include <math.h>
```
DomClark, line 27: C-style `*.h` standard library headers are deprecated in C++ - prefer `cmath` instead.

plugins/Synchro/SynchroSynth.cpp
```diff
@@ -0,0 +1,565 @@
...
+}
+
+QString SynchroSynth::nodeName() const { return synchro_plugin_descriptor.name; };
+
+//Update carrier waveform view
+void SynchroSynth::carrierChanged()
+{
+	for (int i = 0; i < SYNCHRO_GRAPH_SAMPLES; ++i)
+	{
+		float phase = (float)i / (float)SYNCHRO_GRAPH_SAMPLES * F_2PI;
```
DomClark, line 523: More C-style casts to avoid here. However, instead of using `static_cast`, you could just reorder the operands here, e.g.
`F_2PI * i / SYNCHRO_GRAPH_SAMPLES` or `i * F_2PI / SYNCHRO_GRAPH_SAMPLES`.

DomClark (commented):
Sorry for the delay on this review. Code mostly looks good; I've made some suggestions to bring it in line with LMMS's coding conventions and modern C++ style.

plugins/Synchro/SynchroSynth.h
DomClark: I think it's derived from bit_invader's function of the same name, probably via one of Lost Robot's plugin templates.

src/gui/widgets/Graph.cpp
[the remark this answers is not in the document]
DomClark: Sorry to bring it up again, but this change still seems to be here.

plugins/Synchro/SynchroSynth.cpp
iansannar: Fixed the cast you used as an example, making sure there are no other occurrences.

plugins/Synchro/SynchroSynth.h
iansannar: 👾 leftovers. Changed to `nextSample()`. If there's a better/standard name for this kind of function, I will gladly change it to match.

plugins/Synchro/SynchroSynth.cpp
iansannar: Good call! I made a little inline function called `wrapPhase`. If there's an even better way to to this, please let me know!

plugins/Synchro/SynchroSynth.cpp
iansannar: A quick fix is to just make `NoteTime = nph->totalFramesPlayed() * SYNCHRO_OVERSAMPLE`, but I feel like this might be a band-aid on the underlying issue of how I'm handling envelopes in general. I'll come back to this.

plugins/Synchro/SynchroSynth.h
iansannar: lol what's a stack and heap
(i fixed it, thanks)

iansannar:
https://github.com/LMMS/lmms/issues/5592 This PR is almost ready.

- [ ] Style Review: In progress
- [ ] Code Review: In progress
- [x] Testing: Tested by a handful of volunteers from the LMMS Discord server as well as @DouglasDGI's Discord server. No issues found so far.

src/gui/widgets/Graph.cpp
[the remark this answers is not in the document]
iansannar: Whoops. Fixing that now.

plugins/Synchro/SynchroSynth.cpp
```diff
@@ -0,0 +1,570 @@
...
+#include "plugin_export.h" //Required for the plugin to appear in LMMS
+
+//Plugin metadata
+extern "C"
+{
+	Plugin::Descriptor PLUGIN_EXPORT synchro_plugin_descriptor =
+	{
+		STRINGIFY(PLUGIN_NAME),
+		"Synchro",
+		QT_TRANSLATE_NOOP("pluginBrowser", "2-oscillator PM synth"),
```
DomClark, line 46: The `pluginBrowser` translation context has changed to `PluginBrowser` since #5629 (just merged, so this didn't make it into the previous review).

plugins/Synchro/SynchroSynth.h
```diff
@@ -0,0 +1,154 @@
...
+	MM_OPERATORS
+public:
+	//Constructor
+	SynchroNote(NotePlayHandle * nph);
+	//Renders a single sample of audio
+	void nextSample(sampleFrame &outputSample, sample_rate_t sample_rate,
+		const float modulationStrength, const float modulationAmount, const float harmonics,
+		const SynchroOscillatorSettings & carrier, const SynchroOscillatorSettings & modulator);
+private:
+	NotePlayHandle * nph;
```
DomClark, line 66: One more missing `m_`.

plugins/Synchro/SynchroSynth.cpp
iansannar: Changed 👍. Unfortunately it won't build anymore because this PR doesn't have the code from https://github.com/LMMS/lmms/pull/5629.

DomClark: Build error actually seems to be down to a missing open parenthesis on line 176 of `SynchroSynth.cpp`:
https://github.com/LMMS/lmms/blob/3b7b55b1318c0c98342fcfcd8c518caa32f0cc5e/plugins/Synchro/SynchroSynth.cpp#L176

iansannar: Whoops, that would make more sense. Fixed.

DomClark (approved):
Looks good to me.

plugins/Synchro/SynchroSynth.cpp
PhysSong: I wrote this comment to explain why you shouldn't use `sample_rate`(`= Engine::mixer()->processingSampleRate() * SYNCHRO_OVERSAMPLE`), not to say you're doing something wrong. I think you should revert the `noteTime` change to get the correct behavior.

PhysSong: `nph->totalFramesPlayed()` will give you the number of frames based on `processingSampleRate()`, so you can compare with the envelope frames below without multiplying/dividing by `SYNCHRO_OVERSAMPLE`.

plugins/Synchro/SynchroSynth.cpp
PhysSong: I think I meant the whole logic for calculating `modulatorEnvelope` and `carrierEnvelope`.

plugins/Synchro/SynchroSynth.cpp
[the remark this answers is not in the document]
PhysSong: It seems like this one is still not addressed.

PhysSong:
@iansannar How is it going with sample-exactness?
https://github.com/LMMS/lmms/pull/5147#discussion_r329351206

plugins/Synchro/SynchroSynth.cpp
PhysSong: Oh, I even found `noteTime` doesn't change during a rendering period because you're ignoring the frame offset in this function(`frame` in `playNote()`).

iansannar: I don't quite understand that last comment (there's still a lot I don't understand about audio programming). What's frame offset?

plugins/Synchro/SynchroSynth.cpp
iansannar: That logic has now been moved to an `ADSR()` function.

GingkathFox:
this is a pretty fun synth, i like :D

DomClark:
What's the status of this?

iansannar:
Pretty sure it's just waiting for @PhysSong to review both code and style.

plugins/Synchro/SynchroSynth.cpp
```diff
@@ -0,0 +1,553 @@
...
+//sync: hard sync with harmonic multiple
+//chop: how strong the amplitude falloff is per waveform period
+//harmonic: how strong the magic harmonics are
+static inline float SynchroWaveform(float x, float drive, float sync, float chop, float harmonic)
+{
+	return fastTanh(trih(F_PI_2 + x * sync, harmonic) * drive / powf(F_2PI / (F_2PI - x), chop));
+}
+
+//Envelope function
+static inline float ADSR(float attack, float decay, float sustain, float release, float time, bool isReleased, float releaseFramesDone) {
```
PhysSong, line 133: `time` doesn't have a time dimension, so something like `framesPlayed` is more appropriate.
Also, could you add a comment that mentions ADR variables are in ms? You can rename the variables instead if you want.

plugins/Synchro/SynchroSynth.h
```diff
@@ -0,0 +1,153 @@
...
+#include "InstrumentView.h" //Required for instrument GUI
+#include "Knob.h" //Required to use knobs
+#include "NotePlayHandle.h" //Required for audio rendering
+
+constexpr char SYNCHRO_VERSION [] = "0.6";
+constexpr float SYNCHRO_VOLUME_CONST = 0.15f; //Prevents clipping
+constexpr int SYNCHRO_OVERSAMPLE = 4; //Anti-aliasing samples
+constexpr int SYNCHRO_PM_CONST = 20; //Strength of the phase modulation
+constexpr int SYNCHRO_GRAPH_SAMPLES = 48; //Resolution of the waveform view
+constexpr float MAGIC_HARMONICS[2][2] = {{32, 0.5}, {38, 0.025}}; //Yoioioi
```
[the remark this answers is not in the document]
iansannar: I don't remember my exact methodology, but those harmonics (applied to the modulator in particular) give a certain sound that I had envisioned Synchro being able to make since I had the idea years ago.

I think I found those exact values by messing around with Sytrus in the FL Studio demo until I found the desired sound, then looked at exactly what harmonics I had ended up with. There were more than two harmonics, but I found that adding more than the first two had diminishing returns and so it would be easier to only worry about those two.

I'm not attached to that line of code in particular, just the sound it enables. If you've got a way to do it that's more readable/more proper, I'm all ears.

plugins/Synchro/SynchroSynth.cpp
curlymorphic: @PhysSong "@curlymorphic Any ideas how we can reduce aliasing while not losing f < Fs/2 bands significantly?" 

Oversampling could be the best solution and is commonly used, however, I would advise using a third-party library, as the filters will be optimised for this task as it is computationally expensive. The project I am currently on uses   https://github.com/xiph/speexdsp

PhysSong: I'm not sure why I can't add comments to my original review. :thinking:
@curlymorphic Do you mean oversampling and then downsample it using 3rd-party library?

curlymorphic: The signal is being generated at the oversampled rate as expected. It is the downsampling that is the issue, just taking the mean, as the code currently does here is very simplistic.  Schematically the signal is generated at the oversampled rate,  then passed through a filter to remove content above the desired fs/2 then downsampled. This can be very computationally expensive.  A good overview can be found in "Designing audio effects plugin, 2nd, pirkle p624-640."

The filtering and downsampling code, could be written by LMMS. but a lot of development would be needed to make it both efficient and nice sounding, and I feel it would be a lot of effort just to reinvent the wheel.

rubiefawn:
Ah. I have no idea how I did that, but clearly I didn't rebase properly and now the commit history is duplicated 😓

Anyways, after re-examining this, it's clear that it's not even close to finished. I've updated the todo comment at the top.

plugins/Synchro/SynchroSynth.cpp
messmerd: The [HIIR library by Laurent De Soras](http://ldesoras.free.fr/prod.html#src_hiir) was just added to LMMS as a 3rd party library. Maybe it could help here.

michaelgregorius:
I have resolved the merge conflicts by merging `master` and have fixed all compilation problems that have amassed over time with commit d67a3eda6f0.

Hope this might revive development. The synth still works. :slightly_smiling_face:

rubiefawn:
I didn't intend to delete this. I had mistakenly made these commits on my fork's `master` branch, so I decided to rename this branch `feature/synchro`, make a new `master` branch, and switch my fork's default branch to the new `master`. That somehow closed this PR.

I'm working on a full refactor of this synth anyways, so perhaps I'll just open a new PR with that once it's ready for review?

rubiefawn:
?

messmerd:
@rubiefawn Sure, you could open a new PR once it's ready if you'd like

PR #5125, opened 2019-08-14
Extract TemplatesMenu class from MainWindow

winniehell:
Extracts the logic for the project template menu from MainWindow into a new class.

This does not change or add functionality. It's mainly intended to reduce the size of MainWindow and allow extracting the toolbar logic in the long term.

PR #5148, opened 2019-08-24
Extract RecentProjectsMenu class from MainWindow

winniehell:
Extracts the logic for the recent projects menu from MainWindow into a new class.

This does not change or add functionality. It's mainly intended to reduce the size of MainWindow and allow extracting the toolbar logic in the long term.

This refactoring is similar to https://github.com/LMMS/lmms/pull/5125.

### DIFF
--- a/src/gui/MainWindow.cpp
+++ b/src/gui/MainWindow.cpp
@@ -57,6 +57,7 @@
 #include "ProjectJournal.h"
 #include "ProjectNotes.h"
 #include "ProjectRenderer.h"
+#include "RecentProjectsMenu.h"
 #include "RemotePlugin.h"
 #include "SetupDialog.h"
 #include "SideBar.h"
@@ -89,7 +90,6 @@ void disableAutoKeyAccelerators(QWidget* mainWindow)
 
 MainWindow::MainWindow() :
 	m_workspace( NULL ),
-	m_recentlyOpenedProjectsMenu( NULL ),
 	m_toolsMenu( NULL ),
 	m_autoSaveTimer( this ),
 	m_viewMenu( NULL ),
@@ -285,8 +285,7 @@ void MainWindow::finalize()
 					this, SLOT( openProject() ),
 					QKeySequence::Open );
 
-	m_recentlyOpenedProjectsMenu = new RecentProjectsMenu(this);
-	project_menu->addMenu(m_recentlyOpenedProjectsMenu);
+	project_menu->addMenu(new RecentProjectsMenu(this));
 
 	project_menu->addAction( embed::getIconPixmap( "project_save" ),
 					tr( "&Save" ),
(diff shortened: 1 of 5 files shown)


winniehell:
@lukas-w Can you review this one, too, please?

src/gui/menus/RecentProjectsMenu.cpp
```diff
@@ -0,0 +1,74 @@
+#include "RecentProjectsMenu.h"
+#include "embed.h"
+#include "ConfigManager.h"
+#include <QFileInfo>
+#include "Song.h"
+#include "Engine.h"
+#include "MainWindow.h"
+#include "GuiApplication.h"
```
lukas-w: Minor, but we usually pay attention to the order of the included headers. You can take a look at other files for examples. This one could be improved to:

```cpp
#include "RecentProjectsMenu.h"

#include <QFileInfo>

#include "ConfigManager.h"
#include "Engine.h"
#include "Song.h"

#include "embed.h"
#include "GuiApplication.h" 
#include "MainWindow.h"
``` 

Bonus points for grouping LMMS headers by GUI/Core as in the suggestion above, but that's often difficult because it's not always separated clearly.

src/gui/menus/RecentProjectsMenu.cpp
```diff
@@ -0,0 +1,74 @@
...
+	connect( this, SIGNAL( aboutToShow() ),
+			 this, SLOT( updateRecentlyOpenedProjectsMenu() ) );
+	connect( this, SIGNAL( triggered( QAction * ) ),
+			 this, SLOT( openRecentlyOpenedProject( QAction * ) ) );
+}
+
+
+
+
+void RecentProjectsMenu::updateRecentlyOpenedProjectsMenu()
```
lukas-w: Now that the recent projects menu is in its own class, this method name appears overly descriptive. Maybe try something like `fillMenu()`.

src/gui/MainWindow.cpp
```diff
@@ -285,13 +285,8 @@ void MainWindow::finalize()
...
 
-	m_recentlyOpenedProjectsMenu = project_menu->addMenu(
-				embed::getIconPixmap( "project_open_recent" ),
-					tr( "&Recently Opened Projects" ) );
-	connect( m_recentlyOpenedProjectsMenu, SIGNAL( aboutToShow() ),
-			this, SLOT( updateRecentlyOpenedProjectsMenu() ) );
-	connect( m_recentlyOpenedProjectsMenu, SIGNAL( triggered( QAction * ) ),
-			this, SLOT( openRecentlyOpenedProject( QAction * ) ) );
+	m_recentlyOpenedProjectsMenu = new RecentProjectsMenu(this);
+	project_menu->addMenu(m_recentlyOpenedProjectsMenu);
```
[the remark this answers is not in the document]
winniehell: Makes sense! :+1:

src/gui/menus/RecentProjectsMenu.cpp
winniehell: Thanks—I have switched to the correct order now.

src/gui/menus/RecentProjectsMenu.cpp
winniehell: sure! :+1: 

My intention was to keep the name the same so it's easier to review but I agree with your comment.

winniehell:
@lukas-w Thank you! :rocket: I have applied your suggestions.

lukas-w:
Great, thanks!

lukas-w:
:+1: for 52163d02ff01131f8126c832d3bcc7bf7625bdc5 BTW.

PR #5152, opened 2019-08-25
Fix loading state for recent projects menu

winniehell:
This fixes a small bug introduced by https://github.com/LMMS/lmms/pull/5148: When choosing a project from the recent projects menu, no loading indicator is shown because the `setCursor()` call happens on the menu instead of the MainWindow.

This can be reproduced by replacing `loadProject()` with something like `this->thread()->sleep(10);`


winniehell:
@lukas-w Can you please review this small bug fix?

winniehell:
Thank you @PhysSong! :bowing_man:

PR #5155, opened 2019-08-26
Simplify TemplatesMenu class

winniehell:


winniehell:
@lukas-w Can I ask you to take a look at this?

winniehell:
@lukas-w Thank you for the review! :bowing_man: 

I got rid of the additional class. Do you have a rule of thumb in which cases introducing a new class makes sense?

lukas-w:
> Do you have a rule of thumb in which cases introducing a new class makes sense?

Not really, but in this case I'd argue that spanning just a couple of lines, the alternative solution is more concise and more readable.

PR #5142, opened 2019-08-22
Cmake/install refactor

Reflexe:

PR #5156, opened 2019-08-27
Don't use deprecated Qt5 functions

lukas-w:
Continuation of @noahb01's #5086. I rebased on `master`,  removed whitespace changes, squashed related commits and removed the remaining usages of `QSignalMapper`.


PhysSong:
As I mentioned in https://github.com/LMMS/lmms/pull/5086#issuecomment-520282408, e39d6f3d9cc6028d7d5b5ddc981ec38a7ace40f1 requires Qt 5.6, whereas Travis Windows builds use 5.4.

lukas-w:
@PhysSong Should we just remove the Travis Windows builds? We already test with Ubuntu 18.04 Qt 5.9 on CircleCI and I don't see a reason to support Qt versions older than that.

PhysSong:
> Should we just remove the Travis Windows builds?

I think it will be fine once I finish working on #5142.

> We already test with Ubuntu 18.04 Qt 5.9 on CircleCI and I don't see a reason to support Qt versions older than that.

For references, Ubuntu 16.04 ships with Qt 5.5. Upstream dropped support for that version last year. Qt 5.9 LTS is the oldest version which is still supported by upstream. Qt 5.6 LTS has been unsupported since 16.03.2019, but I think it's too early to drop support for Qt 5.6.

lukas-w:
> I think it's too early to drop support for Qt 5.6.

Can you explain why? Linux users don't benefit from this because we bundle Qt in our AppImages anyway and I think developers can be expected to install a recent Qt version through a PPA or the official Qt installer.

PhysSong:
> Linux users don't benefit from this because we bundle Qt in our AppImages anyway

Agreed. BTW, does that mean we're going to deprecate LMMS packages from Linux distributions?
Downstream packagers should take care of the Qt version requirement, but now I think they won't package the new version of LMMS if they don't bundle a recent version of Qt. In other words, I think it's not a big deal for normal users.

> I think developers can be expected to install a recent Qt version through a PPA or the official Qt installer.

If you want to go through the way, you should document this first.

PhysSong:
FYI, the CircleCI Linux build uses Ubuntu 16.04 now because of some issues with AppImages. It means the build uses Qt 5.5 unless upgraded by some packages from the official installer or a PPA.
And we can remove the Travis MinGW builds if you want because packaging works with the newer MinGW toolchains/libraries and even with MSVC.

lukas-w:
> Downstream packagers should take care of the Qt version requirement, but now I think they won't package the new version of LMMS if they don't bundle a recent version of Qt. In other words, I think it's not a big deal for normal users.

Exactly, any distro ship the next LMMS release will also provide a recent Qt version.

> FYI, the CircleCI Linux build uses Ubuntu 16.04 now because of some issues with AppImages

Having to downgrade to an Ubuntu version that has reached end of life because of AppImages is unfortunate. Not that I've been active enough lately to judge but I feel like changes in AppImage and linuxdeployqt are constantly breaking our build. Out of curiosity, can you provide a link for context as to what went wrong? What's the oldest version of Ubuntu we want to support for AppImage users?

> If you want to go through the way, you should document this first.

[Added instructions](https://github.com/LMMS/lmms/wiki/Compiling/_compare/5b52a276ac034b2e92300a3077dea47e20f071e2...5d2642c1b342d935369fa19f58cebc87e1836f6b) to use Qt 5.6+ to the wiki and added the 5.6+ requirement to CMake via 60fc79e5169455259e399c53535cf335ae268042.

PhysSong:
From Discord:
> It seems like linuxdeployqt added a check for libc6 bundling when using `-unsupported-allow-new-glibc`:
https://github.com/probonopd/linuxdeployqt/commit/586aaf7111950047e7654c69f8b604e2afe88976

So, we decided to revert to Ubuntu 16.04 and it's done in #5212.

PhysSong:
@lukas-w FYI, I think we can add a newer Qt from a PPA on our CircleCI Linux build image.

lukas-w:
@PhysSong Thanks for the hints. Working on it right now.

lukas-w:
Fixed via eb5940d348c13b72fc2b3f6f28983e868d1af8be and https://github.com/LMMS/lmms-ci-docker/pull/7, should be good to merge now.

PhysSong:
@lukas-w Since I already pushed commits for CircleCI changes, is it okay to drop the last commit? Sorry for missing this one.

lukas-w:
@PhysSong No worries. I merged 400c8d8105b1ed3cb312d45c2c42f7b9b28da1c2 into master, leaving out the last redundant commit.

Issue #5139, opened 2019-08-21
MIDI Program Change received with incorrect program number


artur-twardowski:
Setup:
- LMMS from current master under Linux (Ubuntu 18.04.3), compiled with VST support
- MIDI keyboard (used Alesis QX49)
- a2jmidid installed

**Preparation for Jack-MIDI:**
- In the LMMS settings window, select MIDI Interface "Jack-MIDI", then restart LMMS
- Execute the command: a2jmidid -e (in a separate terminal or in the background)
- Open the Connections window in Qjackctl and connect the output port of your MIDI keyboard (a2j/QX49) with input port of the LMMS

**Preparation for ALSA-Sequencer:**
- In the LMMS settings window, select MIDI Interface "ALSA-Sequencer), then restart LMMS

**Steps to reproduce:**
- Add an instance of VeSTige to the song editor
- Load into VeSTige any VST instrument that supports Program Change (e. g. DSK Strings, Superwave P8)
- Trigger the Program Change message from the MIDI keyboard

**Expected behavior:**
- The appropriate patch is selected - if we do Program Change with the value of 30, the patch 31 will be selected (MIDI counts programs from 0, LMMS starts numbering them from 1), for Program Change "0" the first patch will be chosen.

**Actual behavior:**
- For Jack-MIDI, the patch number is decreased by 12 - Program Change "12" corresponds to the first patch, Program Change "30" chooses 19th patch. Programs below 12 actually select patches 245 to 256.
- For ALSA-Sequencer, always the first patch is selected

**Further remarks:**
- I have added a few printf()'s inside InstrumentTrack::processInEvent function; the wrong values come directly from the MIDI clients' implementations. I have also checked the actual MIDI messages sent by the keyboard, using amidi in dump mode.
- Jack-MIDI client implementation utilizes MidiClientRaw::parseData function to convert the raw MIDI data into the higher-level data structures. In the place where MIDI Note On, Note Off, Key Pressure, Channel Pressure and Program Change events are handled the number of the key is decreased by KeysPerOctave constant (=12). I assume that the subtraction is needed for the former four events (even though I don't know the purpose), but for Program Change it is unneccessary. Since the function is a member of MidiClientRaw class, I expect that ALSA-Raw MIDI and OSS-Raw MIDI interfaces will also be affected by this bug, but I did not test them.

michaelgregorius:
Hi @artur-twardowski,

[...]

Issue #5138, opened 2019-08-21
MIDI Program Change to change the patches in VeSTige


artur-twardowski:
After loading a VST instrument into VeSTige we are able to walk through all the plugin's patches and try them out instantly. We can choose them using the mouse (clicking on [<], [>], [v] buttons) or the keyboard ([-] and [+] keys).

Many MIDI keyboards are able to send Program Change events (0xC_n_), so that we can change the patches directly from it. However, some VST instruments seem to ignore them while still working correctly with VeSTige's controls (examples are SQ8L and DVS Guitar). Additionally, changing patches from the MIDI keyboard does not update the patch name visible in VeSTige user interface.

My proposal is to embed the handling of Program Change directly into VeSTige. Additionally CCs 0 and 32 (Bank Select MSB & LSB) should be supported as well, so that we are able to access the patches beyond 128th one. To give the user a choice, a toggle button can be added to the VeSTige GUI, allowing to enable or disable the feature.

DomClark:
Thanks for the enhancement suggestion! In order to keep our tracker clean and simple, I'm going to consolidate this issue into our MIDI placeholder issue #1472, which contains many other issues related to general MIDI improvement. If you'd like to have a go at implementing this yourself, we'd love the help.

artur-twardowski:
Similar change to be implemented in Sf2 Player and GIG Player

Issue #1472, opened 2014-12-20
Better MIDI support


tresf:
This is a placeholder to consolidate MIDI related issues.  If a new MIDI enhancement is reported on the tracker, close it out and mark it as a duplicate of this bug (and naturally add to this list as needed) (#1472)

**Note:** As opposed to our other placeholder bug reports, MIDI is slightly different as it is mostly feature-complete.  Version-specific MIDI bugs should remain open and tracked on a per-version basis.  Don't close out valid bug reports simply because they use the word "MIDI" in them, first make sure they aren't attributed to a specific build. :+1: 

---

**Basic Functionality:**
- [x] LMMS should not manipulate the raw incoming MIDI streams #1857 [Milestone: 1.3]
- [ ] Internal routing/support for MIDI-generating plugins/effects (e.g. BlueARP) #3485
- [ ] MIDI import should support pitch bend range #1967
- [ ] MIDI export support #258, #699 (via #1686, reverted via #3114 per #2384)
- [ ] (Windows) empty/assignable MIDI channel #793 (Related #3485)
- [x] (Apple) basic MIDI support #1153

**Enhanced Functionality:**
- [ ] Simultaneous multi-track recording #4603
- [ ] Expose MIDI controls (all 128 CC messages) #1159
  - [ ] MIDI play/pause/stop/record #1071 #4837
  - [ ] MIDI move loop #1071
  - [ ] MIDI bank-selection (MSB) support #62
  - [x] MIDI pitch-bend recording support #1918, #3724 
  - [ ] MIDI modulation wheel, sustain pedal and expression pedal support #1918 #1917 #1890 #1700 #1953
  - [ ] MIDI program change support for plugins #5138
- [ ] Midi timecode output socket (ALSA, JACK #4838) 
- [ ] Allow using the MIDI output of Carla for LMMS' MIDI out #4858
- [x] Faster access to MIDI assignment/auto-assignment #3400 #1189 #1781 #1918 #1942 #3185
- [ ] Better MIDI import support/transpose #815 #148 #1918
- [ ] Game controller MIDI support #695
- [ ] MIDI mapping using custom key ranges #1743 #1381 #1792 #1460 [Milestone: 1.3]
- [ ] Better VST effect support (i.e. Autotune, Vocoder) #1202
- [ ] Multiplexing MIDI channels to a single instrument #1178
- [ ] Microtonality (Scala `.scl` files, retuning, etc) #1387
- [ ] MPE #3439 
[...]

PR #5158, opened 2019-08-28
MIDI Program Change message support in VeSTige, SF2 Player and GIG Player

artur-twardowski:
Adds the native support for the following MIDI events in VeSTige:
- Program Change
- Controller Change for controllers 0 and 32, which are Bank Select MSB and LSB, respectively

The patch adds two toggle switches to VeSTIge GUI:
- Capture Program Change events
- Use Bank Select LSB
which are loadable and saveable. To make the room for new switches it was neccessary to reorganize slightly the existing user interface - "Show/Hide GUI" and "Stop all notes" buttons were reduced to the form of icons. Additionally, the MIDI preset number was moved from preset name line to the "Preset:" caption above it, so that there is more space for the actual preset name.

**Description of new features**

When "Capture Program Change events" switch is on, the events above are captured directly by VeSTige and used to determine the ID of preset to switch to; the VST plugin will not receive these events. When the switch is off the MIDI events will be passed through to the plugin.

When "Use Bank Select LSB" switch is on, both controllers 0 and 32 are used. The values are combined to obtain the bank number between 0 and 16383 (2^14 - 1). When the switch is off, only controller 0 is used and is treated as LSB. This allows to choose banks between 0 and 127 (2^7 - 1). This allows to work with MIDI devices that use either of the approaches.

The target preset ID is determined from the bank number and the program number. Presets 0 to 127 belong to bank 0, presets 128..255 fall into bank 1, presets 256..383 belong to bank 2, and so on.

Reception of the Program Change and/or Bank Select messages will trigger the update of the VeSTige GUI, so the preset name is visible immediately.

**Validation of changes**
_Preparation_
- Connect any MIDI controller capable of sending subject messages (e. g. Alesis QX49)
- Add VeSTige to the song editor and load any VST plugin that has its GUI and at least a few presets. I have used SuperWave P8 and SQ8L.
- Enable the MIDI input. If you use ALSA Sequencer as MIDI interface, pick appropriate MIDI device from the list.
- Open the VST GUI window

_Validating Program Change_
- Having both switches disabled, trigger Program Change event at your controller.
  - if the plugin you loaded accepts Program Change messages (e. g. SuperWave P8), you will see a change in the plugin's GUI, but no change of the preset in VeSTige
  - if the plugin does not handle Program Change (e. g. SQ8L), there will be no change in the GUI
- Enable "Capture Program Change events" switch and trigger Program Change event.
  - the preset name displayed in VeSTige GUI should change according to the program number received, contents of plugin's GUI should be updated as well

_Validating Bank Select_
Note: here I recommend using a VST plugin with more than 128 presets.
- Enable "Use Bank Select LSB" switch
- Send Program Change with the value of 0
  - the first preset should be selected
- Send Bank Select MSB with the value of 0 and Bank Select LSB with the value of 1
  - the 129th preset should be selected (bank 1, program 0, meaning Preset ID=128)
- Send Bank Select LSB with the value of 2
  - the 257th preset should be selected (bank 2, program 0, meaning Preset ID=256)
- Send Bank Select MSB with the value of 1 and Bank Select LSB with the value of 0
  - the last preset should be selected (bank 128, program 0, meaning Preset ID=16384)

- Disable "Use Bank Select LSB" switch
- Send Bank Select MSB with the value of 0 and Bank Select LSB with the value of 1
  - the first preset should be selected (bank 0, program 0, meaning Preset ID=0)
- Send Bank Select LSB with any value
  - there should be no reaction of the plugin
- Send Bank Select MSB with the value of 1
  - the 129th preset should be selected (bank 1, program 0, meaning Preset ID=128)

_Validating save & load_
- Set any state of the switches in VeSTige GUI, then save the project
- Open a new project
- Reopen the previously saved project
- Open the GUI of the instance of VeSTige loaded into the project
  - the state of switches should be the same as when the project was saved

**Notes**
Change in scope of #5138 (already closed) which is actually a part of #1472.
The pull request includes a fix for #5139 which has not yet been merged to the master branch, but is required by the new feature.


Spekular:
I don't see why these changes should affect the "Show/Hide GUI" and "Stop all notes" buttons. As these are MIDI related settings, wouldn't they go in the MIDI settings tab?

artur-twardowski:
MIDI settings tab is part of LMMS. Only the plugin tab displays the controls specific to plugins (in this case VeSTige).

artur-twardowski:
I thought a bit more about the proposal above, it seems that "Use Bank Select LSB" switch could actually be moved to the MIDI tab. I started working on implementation of similar functionality in GIG Player and Sf2 Player, it turned out that I need to repeat there a substantial amount of code. The same problem would appear in other plugins when they get the same functionality as a further step.

Moreover, GIG Player has a plenty of room in its GUI, but fitting the switch into GUI of Sf2 Player will be a hard task.

I think that a new message should be added in the abstraction layer between MIDI clients and consumers. Maybe Instrument or Plugin class could be enhanced with a new virtual function, say, changePresets? The common code would combine Bank Selects and Program Change into a proper call with ready-made preset number.

Still, something has to be done with "Capture Program Change" switch. It is specific to VeSTige. If we could move the "plugin information" box a few pixels up, then I could restore the original appearance of aforementioned buttons, adding the new switch below "Stop all notes".

plugins/vestige/vestige.cpp
```diff
@@ -160,6 +164,10 @@ vestigeInstrument::vestigeInstrument( InstrumentTrack * _instrument_track ) :
 			 Qt::QueuedConnection );
 }
 
+void vestigeInstrument::bindView(VestigeInstrumentView * view)
+{
+	m_pView = view;
```
PhysSong: Can you use Qt's signal-slot connection instead of this approach? It's considered bad to store the reference to the view in the model.

plugins/vestige/vestige.cpp
```diff
@@ -558,33 +628,62 @@ VestigeInstrumentView::VestigeInstrumentView( Instrument * _instrument,
...
+							"stop", 20, 20  ) );
+	connect( m_panicButton, SIGNAL( clicked() ),
+		 this,SLOT( noteOffAll() ) );
+
+	m_capturePgmChangeButton = new PixmapButton(this,
+						  tr("Capture Program Change events"));
+	m_capturePgmChangeButton->setActiveGraphic(embed::getIconPixmap("led_green"));
+	m_capturePgmChangeButton->setInactiveGraphic(embed::getIconPixmap("led_off"));
+	m_capturePgmChangeButton->setCheckable(true);
+	ToolTip::add( m_capturePgmChangeButton, tr(
```
PhysSong: The tooltip text looks too long to me.

plugins/vestige/vestige.cpp
```diff
@@ -558,33 +628,62 @@ VestigeInstrumentView::VestigeInstrumentView( Instrument * _instrument,
...
+			      "simply want VeSTige to do this task."
+			      ));
+	m_capturePgmChangeButton->move(20, 205);
+
+	m_useBankSelectLSBButton = new PixmapButton(this,
+						  tr("Use MIDI Bank Select LSB"));
+	m_useBankSelectLSBButton->setActiveGraphic(embed::getIconPixmap("led_green"));
+	m_useBankSelectLSBButton->setInactiveGraphic(embed::getIconPixmap("led_off"));
+	m_useBankSelectLSBButton->setCheckable(true);
+	ToolTip::add( m_useBankSelectLSBButton, tr(
```
PhysSong: Here as well.

plugins/vestige/vestige.cpp
artur-twardowski: Done

plugins/vestige/vestige.cpp
artur-twardowski: Reworked the solution, so the control does not exist anymore, neither does the tooltip.

plugins/vestige/vestige.cpp
artur-twardowski: same as above

artur-twardowski:
I have finally found a reasonable way to move "Capture Program Change" switch to MIDI tab as well. For this reason I restored the original appearance of VeSTige (but still leaving the preset number in the upper row).

Since the change is now in the common part of the code, I decided to take an opportunity and adapt Sf2 Player and GIG Player to the new functionality. Originally I intended to create a separate pull request for them, but since they allow to choose bank independently from preset I decided to re-design a bit the interface. The instruments do not receive the ready-made preset number anymore, but instead they get a separate "nullable" bank number and program number.

There are 3 policies of handling Bank Select CCs:
- **Ignore**; CCs 0 and 32 are not used, bank switching is entirely disabled, the plugin will get bank=-1 when changing the preset. The impact depends on the plugin:
  - If a plugin required bank number to expand the range of selectable presets, it will treat the bank number as 0, effectively allowing selection of presets from 0 to 127. For a moment I thought about "guessing" the bank number from currently selected preset - if we are at, let's say, preset 300th (2 x 128 + 44), we can shift the selection range to "virtual" bank 2, exposing presets 256 to 383. This is how SQ8L treats these "high" presets - it will identify it as C044. However, I finally came up to the conclusion that this behavior would be really confusing to the user.
  - If a plugin has a separate control for bank selection (Sf2 Player, Gig Player), it may leave the bank untouched and switch only the preset. This way, we can select manually a bank, then browse the presets inside using MIDI keyboard. This is where the reason from change came from - if we sent just a single number, the bank would always reset to 0.
- **Use MSB only**; CC 0 is used and is treated as LSB, CC 32 is ignored. Some MIDI controllers use MSB as the only bank switching message while others (such as Casio CTK-4200) send both Bank Select messages, but the actual bank number is still in MSB.
- **Use both MSB and LSB**; CC 0 and 32 are both used and handled according to their names. To be used with MIDI controllers that actually treat MSB as MSB and LSB as LSB, allowing to choose banks from 0 to 16383. Controllers that use MSB only will cause selection of banks being multiples of 128.

Open points:
- are there any MIDI controllers that treat CC 0 as LSB and CC 32 as MSB? If there were ones that behave this way, a "bank-inversion" policy could be added.
- to fit the new control into MIDI tab I removed (hid) the explanation of Custom Base Velocity. Otherwise the section was truncated, but still it enlarged a bit the window creating an oddly-looking margin below the plugin's GUI. Any comments or suggestions on that?
- the API provides a function allowing each plugin to tell if it actually supports preset switching or not. The function is currently unused - my intention is however to hide the "Enable Preset Selection" section or display some placeholder text in case a plugin does not implement preset switching.

PhysSong:
Some comments:
- Please remove `vestige.orig.cpp`.
- Is it okay not checking if there's a patch with the specific number?
- It seems like `patchChanged` is emitted even if the patch remains the same.

artur-twardowski:
> Some comments:
> 
> * Please remove `vestige.orig.cpp`.
Done.

> * Is it okay not checking if there's a patch with the specific number?
For SF2/Gig players, they allow to select an inexistent patch anyway via LcdSpinBoxes.
For VeSTige, selection of a preset with too high number will result in choosing the last preset available.
I can update the code of all three plugins to ignore the messages containing invalid values, the question is if such behavior will be confusing to the user.

> * It seems like `patchChanged` is emitted even if the patch remains the same.
This behavior is not harmful since the handling of the signal consists of UI update. Nevertheless, for sake of consistency with the signal's name, I updated the code of SF2/Gig players to check if the bank and/or preset number have actually changed. For VeSTige there will be a bit more work to do.

src/tracks/InstrumentTrack.cpp
```diff
@@ -882,6 +903,85 @@ Instrument * InstrumentTrack::loadInstrument(const QString & _plugin_name,
...
+		if (m_programBankMSB != -1)
+		{
+			m_instrument->changePreset(
+				(m_programBankMSB << 7) | m_programBankLSB,
+				m_programNumber);
+		}
+		else
+		{
+			m_instrument->changePreset(-1, m_programNumber);
+	}
```
PhysSong: Weird indentation.
```suggestion
		}
```

artur-twardowski: Right, fixed.

plugins/GigPlayer/GigPlayer.h
```diff
@@ -267,6 +267,8 @@ class GigInstrument : public Instrument
 
 
 	void setParameter( const QString & _param, const QString & _value );
+	virtual bool presetChangeSupported() override;
+	virtual void changePreset(int bank, unsigned int preset) override;
```
PhysSong, line 271: `override` implies `virtual`, so you can omit `virtual`.

plugins/sf2_player/sf2_player.h
```diff
@@ -91,6 +91,8 @@ class sf2Instrument : public Instrument
 
 	void setParameter( const QString & _param, const QString & _value );
 
+	virtual bool presetChangeSupported();
+	virtual void changePreset(int bank, unsigned int preset);
```
PhysSong, line 95: Please add `override` and remove `virtual`.

plugins/vestige/vestige.cpp
```diff
@@ -60,8 +60,6 @@
 
 #include "embed.h"
 
-
-
```
PhysSong, line 64: As a reminder: I recommend you reverting irrelevant changes in new lines and whitespace after finishing this PR.

PhysSong:
The code looks good in general. I think we need some testers for this. Testing PR before merging is not mandatory but strongly recommended.

botwhytho:
Any testers still needed? In Linux as a daily driver so could test SF2, might be able to boot into a Windows environment and test Vestige too.

zonkmachine:
> Any testers still needed?

Yes please!

botwhytho:
@zonkmachine If there are any instructions on how to build feature branches on Linux (on Ubuntu 20.04) or somewhere to get Linux and/or Windows assets that would be great, otherwise any pointers would be of help as I've used LMMS for a couple of years but not really familiar with it's build system or dependencies. Thanks.

SecondFlight:
Hi @botwhytho, there are links to pre-built .AppImage files for this PR. See the comment above by @LmmsBot, which is regularly edited with the latest build links.

src/gui/widgets/InstrumentMidiIOView.cpp
```diff
@@ -145,11 +166,15 @@ InstrumentMidiIOView::InstrumentMidiIOView( QWidget* parent ) :
 	baseVelocityLayout->setContentsMargins( 8, 18, 8, 8 );
 	baseVelocityLayout->setSpacing( 6 );
 
+	/* Base velocity help label hidden to preserve the window size.
+	 * Maybe there is a better solution */
+#if 0
```
PhysSong, line 171: This should be fixed before merging.

artur-twardowski:
Code ported to the new code base. A new pull request was created (#6676), closing this one.

PR #5161, opened 2019-08-30
Add Disintegrator effect to LMMS

DouglasDGI:
(Made this whole thing in like 30 minutes, haha)

I'm not going to try to hide it, this effect is a blatant ripoff of Ableton's Erosion.  You could probably even call it a clone.  That was less because I intended to copy them and more because I couldn't think of anything else I'd want to add.

Disintegrator (and Erosion) is an effect that delays the sound with a modulated delay in super aggressive ways to get some very digital-sounding distortion.  The first mode modulates the delay with white noise, which is filtered by the two filtering knobs before it does any modulation.  The second mode is identical to the first, except the white noise is stereo rather than mono, which sounds much wider.  The third mode uses a sine wave as the modulator, which sounds very similar to FM and can result in some very cool timbres.

Notice that the two filtering knobs are replaced with a frequency knob when the sine wave mode is selected.

This effect is quite basic and doesn't make any changes to other parts of LMMS, so I'm assuming the review-and-merge process will go by fairly quickly.


plugins/Disintegrator/Disintegrator.cpp
```diff
@@ -0,0 +1,279 @@
...
+	Effect(&disintegrator_plugin_descriptor, parent, key),
+	m_disintegratorControls(this)
+{
+	for (int a = 0; a < 200; ++a)
+	{
+		for (int b = 0; b < 2; ++b)
+		{
+			m_inBuf[b].push_back(0);
+		}
+	}
```
PhysSong: This will run slightly faster:
```cpp
		for (int i = 0; i < 2; ++i)
		{
			m_inBuf[i].reserve(200);
			for (int j = 0; j < 200; ++j)
			{
				m_inBuf[i][j] = 0.0f;
			}
		}
```
Also, could you use a constant whose value is 200 instead of using the value directly? This will make things much easier if you change the value. Also, it will make clear what 200 means.

FYI, loop variables *usually* start from `i`, not `a`.

plugins/Disintegrator/Disintegrator.h
```diff
@@ -0,0 +1,84 @@
...
+
+	virtual EffectControls* controls()
+	{
+		return &m_disintegratorControls;
+	}
+
+	inline float realfmod(float k, float n);
+	inline float detuneWithOctaves(float pitchValue, float detuneValue);
+
+	inline void calcLowpassFilter(sample_t &outSamp, sample_t inSamp, int which, float lpCutoff, float resonance, sample_rate_t Fs);
```
PhysSong: `which` means the index of filter, right? I think the name is a little bit unclear. Also, I think `sampleRate` will be better than `Fs`.

plugins/Disintegrator/Disintegrator.cpp
```diff
@@ -0,0 +1,279 @@
...
+	const float d = dryLevel();
+	const float w = wetLevel();
+	
+	const ValueBuffer * lowCutBuf = m_disintegratorControls.m_lowCutModel.valueBuffer();
+	const ValueBuffer * highCutBuf = m_disintegratorControls.m_highCutModel.valueBuffer();
+	const ValueBuffer * amountBuf = m_disintegratorControls.m_amountModel.valueBuffer();
+	const ValueBuffer * typeBuf = m_disintegratorControls.m_typeModel.valueBuffer();
+	const ValueBuffer * freqBuf = m_disintegratorControls.m_lowCutModel.valueBuffer();
+
+	sample_rate_t sample_rate = Engine::mixer()->processingSampleRate();
```
PhysSong: Names with camel cases are preferred because it's consistent with other parts of the code.
```suggestion
	sample_rate_t sampleRate = Engine::mixer()->processingSampleRate();
```

plugins/Disintegrator/Disintegrator.cpp
```diff
@@ -0,0 +1,279 @@
...
+		const float alpha = sin(w0) / 0.3535f;
+
+		m_LPa0 = 1 + alpha;
+		m_LPb0 = (1 - tempCosW0) * 0.5f;
+		m_LPb1 = 1 - tempCosW0;
+		m_LPa1 = -(2 * tempCosW0);
+		m_LPa2 = 1 - alpha;
+	}
+
+	m_filtLPY[which][0] = (m_LPb0 * (inSamp + m_filtLPX[which][2]) + m_LPb1 * m_filtLPX[which][1] - m_LPa1 * m_filtLPY[which][1] - m_LPa2 * m_filtLPY[which][2]) / m_LPa0;
```
PhysSong: This line looks too long. Can you break this to 2 or 3 lines for better readability?

plugins/Disintegrator/Disintegrator.cpp
```diff
@@ -0,0 +1,279 @@
...
+		const float alpha = sin(w0) / 0.3535f;
+
+		m_HPa0 = 1 + alpha;
+		m_HPb0 = (1 + tempCosW0) * 0.5f;
+		m_HPb1 = -(1 + tempCosW0);
+		m_HPa1 = -2 * tempCosW0;
+		m_HPa2 = 1 - alpha;
+	}
+
+	m_filtHPY[which][0] = (m_HPb0 * (inSamp + m_filtHPX[which][2]) + m_HPb1 * m_filtHPX[which][1] - m_HPa1 * m_filtHPY[which][1] - m_HPa2 * m_filtHPY[which][2]) / m_HPa0;
```
PhysSong: Here as well.

DouglasDGI:
I just realized the code is quite lacking in comments.  I'll fix that.

plugins/Disintegrator/Disintegrator.cpp
```diff
@@ -0,0 +1,279 @@
...
+{
+	return pitchValue * std::exp2(detuneValue); 
+}
+
+
+
+// Handles negative values properly, unlike fmod.
+inline float DisintegratorEffect::realfmod(float k, float n)
+{
+	return ((k = fmod(k,n)) < 0) ? k+n : k;
```
PhysSong:
```suggestion
	return ((k = fmod(k, n)) < 0) ? k + n : k;
```
Alternatively,
```suggestion
	float r = fmod(k, n);
	return r < 0 ? r + n : r;
```

DouglasDGI:
Alright, I added comments to the code and also removed the "detuneWithOctaves" function which I accidentally left in there.  Should be ready now.

plugins/Disintegrator/Disintegrator.cpp
```diff
@@ -0,0 +1,279 @@
...
+		m_prevLPCutoff[which] = lpCutoff;
+		const float w0 = D_2PI * lpCutoff / Fs;
+		const float tempCosW0 = cos(w0);
+		const float alpha = sin(w0) / 0.3535f;
+
+		m_LPa0 = 1 + alpha;
+		m_LPb0 = (1 - tempCosW0) * 0.5f;
+		m_LPb1 = 1 - tempCosW0;
+		m_LPa1 = -(2 * tempCosW0);
+		m_LPa2 = 1 - alpha;
```
PhysSong: It seems like you're using a Biquad filter with resonance 0.707(not the `resonance` parameter for some reason). You may be able to use `BasicFilters` with type `LowPass` and `HiPass`.

plugins/Disintegrator/DisintegratorControls.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+#include "Engine.h"
+#include "Song.h"
+
+
+DisintegratorControls::DisintegratorControls(DisintegratorEffect* effect) :
+	EffectControls(effect),
+	m_effect(effect),
+	m_lowCutModel(2.0f, 2.0f, 21050.0f, 0.01f, this, tr("Low Cut")),
+	m_highCutModel(21050.0f, 2.0f, 21050.0f, 0.01f, this, tr("High Cut")),
```
PhysSong: I think a filter with the frequency limit above 20kHz is not quite meaningful.

plugins/Disintegrator/Disintegrator.cpp
DouglasDGI: Whoops, forgot to remove the resonance argument.
I think the filters I'm using run significantly faster than the ones in BasicFilters since they're optimized specifically for those filter types, should I still use BasicFilters anyway?

PhysSong: > I think the filters I'm using run significantly faster than the ones in BasicFilters since they're optimized specifically for those filter types, should I still use BasicFilters anyway?

Try it, and check for the difference. I think using `BasicFilters` will be better if the difference is negligible. If there are significant differences, please feel free to keep using this.

DouglasDGI: The difference isn't noticeable unless you have a very large number of the effect loaded.  I switched over to BasicFilters.h in the last commit and I think it sounds better (I think I recall the code saying those filters had 2x oversampling?).

DouglasDGI:
An absolute galaxy brain (known as Oncill) suggested that I add a new mode that modulates the delay time depending on the audio input.  I tested it out.  The result can work as an aggressive exciter, as well as something that (although is very violent and digital) is unnaturally good at enhancing the general timbre of any instrument you feed it.  The filters are still enabled for this mode, which filter the input audio modulator (but not the "carrier", if that's even the correct term in this case).  The new mode was added in the latest commit.

plugins/CMakeLists.txt
```diff
@@ -49,6 +49,7 @@ IF("${PLUGIN_LIST}" STREQUAL "")
 		carlarack
 		CrossoverEQ
 		Delay
+		Disintegrator
```
zonkmachine: This part of `plugins/CMakeLists.txt` has been moved to `cmake/modules/PluginList.cmake` in 15fe551b1da9863db6b127665dd390473615513d

DouglasDGI:
Don't test this out yet, I found some weird bug with the self modulation mode that isn't appearing in my copy for some reason...

DouglasDGI:
Alright, all merge conflicts and bugs have been removed.  Feel free to test now.
Here's a video of me messing around with this effect: https://youtu.be/nzUV3gDL4Wo

plugins/Disintegrator/Disintegrator.cpp
PhysSong: Oh, I think this will be better: https://stackoverflow.com/a/34683195

DouglasDGI:
I just realized this effect sounds different when the sample rate is changed.  Fixed it in the latest commit. :+1:

plugins/Disintegrator/Disintegrator.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+					// Distance between samples
+					newInBufLocFrac[i] = fmod(newInBufLoc[i], 1);
+				}
+
+				break;
+			}
+			case 2:// Sine Wave
+			{
+				m_sineLoc = fmod(m_sineLoc + (freq / m_sampleRate * F_2PI), F_2PI);
```
PhysSong, line 168: You may precalculate `F_2PI / m_sampleRate` for optimization purpose.

PhysSong:
@DouglasDGI REMINDER. Did you see my reviews?

plugins/Disintegrator/Disintegrator.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+
+DisintegratorEffect::DisintegratorEffect(Model* parent, const Descriptor::SubPluginFeatures::Key* key) :
+	Effect(&disintegrator_plugin_descriptor, parent, key),
+	m_disintegratorControls(this),
+	m_lp(Engine::mixer()->processingSampleRate()),
+	m_hp(Engine::mixer()->processingSampleRate()),
+	m_needsUpdate(true)
+{
+	emit sampleRateChanged();
```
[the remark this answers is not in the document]
PhysSong:
```suggestion
	sampleRateChanged();
```

plugins/Disintegrator/Disintegrator.cpp
```diff
@@ -0,0 +1,259 @@
...
+	
+	const ValueBuffer * amountBuf = m_disintegratorControls.m_amountModel.valueBuffer();
+	const ValueBuffer * typeBuf = m_disintegratorControls.m_typeModel.valueBuffer();
+	const ValueBuffer * freqBuf = m_disintegratorControls.m_lowCutModel.valueBuffer();
+
+	// Update filters
+	if(m_needsUpdate || m_disintegratorControls.m_highCutModel.isValueChanged())
+	{
+		m_lp.setLowpass(m_disintegratorControls.m_highCutModel.value());
+
```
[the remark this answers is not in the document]
PhysSong:
```suggestion
```

PhysSong:
@DouglasDGI Is this ready for review?

DouglasDGI:
@PhysSong I was planning on rewriting this, it's incompatible with sample rate changes and isn't written very well in general.  It shouldn't take more than a few days to rewrite (maybe even just one day, it's a fairly simple plugin), I'll get to that as soon as I'm done with the Phaser changes I've been working on.

DouglasDGI:
@PhysSong I was wrong, turns out I did make it compatible with sample rate changes, just not in a nice or neat way...
Anyway, the whole code was kind of a mess and it turns out it was super buggy as well.  The last commit fixes everything.  The filters should actually work correctly now, and the Self Modulation mode should sound significantly cleaner.

Also, the Amount knob is now in units of milliseconds like it should have been, rather than samples like it was before.

plugins/Disintegrator/Disintegrator.cpp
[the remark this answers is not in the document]
PhysSong: You somehow added this again.

plugins/Disintegrator/DisintegratorControlDialog.cpp
```diff
@@ -0,0 +1,97 @@
...
+	connect(&m_controls->m_typeModel, SIGNAL(dataChanged()), this, SLOT(updateKnobVisibility()));
+	emit m_controls->m_typeModel.dataChanged();// Needed to update knobs when view is opened
+}
+
+
+/* Switches between the lowcut/highcut and
+frequency knobs depending on the modulation type. */
+void DisintegratorControlDialog::updateKnobVisibility()
+{
+	if (m_controls->m_typeModel.value() == 2)
```
PhysSong, line 85: I suggest adding a comment that says 2 is for sine wave.

plugins/Disintegrator/Disintegrator.cpp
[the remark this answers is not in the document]
DouglasDGI: ...yeah, somehow.  How does that even happen?  I didn't do that...
...well, it's fixed now I guess.  Hm.

plugins/Disintegrator/Disintegrator.cpp
```diff
@@ -0,0 +1,240 @@
...
+	for (fpp_t f = 0; f < frames; ++f)
+	{
+		const float amount = (amountBuf ? amountBuf->value(f) : m_disintegratorControls.m_amountModel.value()) * 0.001f * m_sampleRate;
+		const int type = typeBuf ? typeBuf->value(f) : m_disintegratorControls.m_typeModel.value();
+		const float freq = freqBuf ? freqBuf->value(f) : m_disintegratorControls.m_freqModel.value();
+	
+		sample_t s[2] = {buf[f][0], buf[f][1]};
+
+		// Increment buffer read point
```
curlymorphic, line 115: when implementing a circular buffer, it is common to have the buffer size a power of 2, with 512 being the next closest to the 500 you stare in the header.

The index increment and wrap at end of buffer can then be reduced to: 
``` 
++m_inBufLoc;
m_inBufLoc &= (m_bufferSize - 1)
```

This is preferable as conditional branches can be processor expensive, and may also hinder some compiler side optimisations.

plugins/Disintegrator/Disintegrator.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+	// Update filters
+	if(m_needsUpdate || m_disintegratorControls.m_highCutModel.isValueChanged())
+	{
+		m_lp->calcFilterCoeffs(m_disintegratorControls.m_highCutModel.value(), 0.5);
+		
+	}
+	if(m_needsUpdate || m_disintegratorControls.m_lowCutModel.isValueChanged())
+	{
+		m_hp->calcFilterCoeffs(m_disintegratorControls.m_lowCutModel.value(), 0.5);
```
curlymorphic, line 103: The updating of the filter coefficients should be done per frame using the value buffer, not once per buffer.

The current implementation is block size dependant, so the audio will be different depending on the block size. This is not desirable, and the effect should remain independent. 

One of the drives for sample exactness was to remove the buzzing and zipping noises, these are caused by sudden stepped changes in parameters on a per-block basis, and are removed by smoothly updating the parameters on a per-frame basis.

curlymorphic (changes requested):
A nice plugin, seam to get a lot for your money as they say.  It is really simple and effective. I have left some comments in the code, I feel these are minor amendments, leading to a substantial improvement.

My other concern is the name "Disintegrator", Lmms already comes bundled with a LADSPA effect of the same name. Maybe now would be a good time to reconsider the name, as a name change would not be viable once release.

I have kept my review to the audio processing code, I have not reviewed any UI components, or comment on style.

rdrpenguin04:

plugins/Disintegrator/Disintegrator.cpp
```diff
@@ -0,0 +1,240 @@
...
+	sampleRateChanged();
+}
+
+
+
+void DisintegratorEffect::sampleRateChanged()
+{
+	m_sampleRate = Engine::mixer()->processingSampleRate();
+	m_2PiOverSR =  F_2PI / m_sampleRate;
+	m_lp = new BasicFilters<2>(Engine::mixer()->processingSampleRate());
```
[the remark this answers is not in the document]
PhysSong: We recently switched to C++14, FYI.

DouglasDGI:
@rdrpenguin04 I don't actually recall requesting that... I did request something similar for Microwave, but for everything else I'd much rather keep them open to give others the chance to finish them if they're able to.

(And I'm back for the time being by the way, I should be able to finish this one up)

rdrpenguin04:
> @rdrpenguin04 I don't actually recall requesting that... I did request something similar for Microwave, but for everything else I'd much rather keep them open to give others the chance to finish them if they're able to.
> 
> (And I'm back for the time being by the way, I should be able to finish this one up)

Sorry about that; I had assumed you had meant all of your PRs.

plugins/Disintegrator/DisintegratorControls.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+
+DisintegratorControls::DisintegratorControls(DisintegratorEffect* effect) :
+	EffectControls(effect),
+	m_effect(effect),
+	m_lowCutModel(20.0f, 20.0f, 20000.0f, 0.01f, this, tr("Low Cut Frequency")),
+	m_highCutModel(20000.0f, 20.0f, 20000.0f, 0.01f, this, tr("High Cut Frequency")),
+	m_amountModel(1.6f, 0.0f, 10.0f, 0.001f, this, tr("Amount")),
+	m_typeModel(this, tr("Type")),
+	m_freqModel(100.0f, 1.0f, 21050.0f, 0.01f, this, tr("Frequency"))
```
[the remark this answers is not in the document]
DouglasDGI: 100 is the default value, not the lowest value.  The lowest value is 1.  The highest value... I think that's a typo.  Pretty sure I meant 22050.

PR #5168, opened 2019-09-04
Add Phaser effect to LMMS

LostRobotMusic:
(Awesome GUI designed by Roxas)

Introducing the Phaser, the ultimate Flanger companion!  :)

Here are the changes made to the Flanger:
1. Allow negative feedback values (I'm surprised it wasn't an option already)
2. Add a phase knob to change the stereo LFO phase (so you can have a mono flanger, for example)
3. QuadratureLfo was combined into one file, because the CPP file had almost nothing in it and QuadratueLfo needed to be included by the Phaser.

All three changes are 100% backwards-compatible.

Here's a screenshot of the new Phaser: 
image

Here's an overview of its features (EDIT: This is out of date, I make some updates in some comments below):
1. For those who don't know, a phaser is an allpass filter with an LFO at a partial wet/dry, which very commonly has an adjustable number of allpass filters as well as a 1-sample delay feedback option.
2. This Phaser can have as few as 1 and as many as 32 allpass filters, adjustable using the filter order spinbox.
3. The LFO has an adjustable rate (with optional tempo sync), amount, and stereo phase.
4. Phaser is the first-ever animated LMMS effect to my knowledge (unless you want to count the volume meters in the fader in the Delay effect...).  The two dots at the bottom display the allpass filter frequencies in a very intuitive way, with the green dot showing the left ear and the blue dot showing the right ear.
5. Unlike most phasers (surprisingly), the LFO can be disabled, allowing the user to automate the cutoff knob to have full control over the Phaser.  This Phaser also comes with adjustable resonance, which is also a rare feature for reasons I'm unaware of.
6. This Phaser allows positive and negative feedback values, which are an essential feature for most phasers.  Also, unlike any other phaser I've seen, the feedback delay can be adjusted to values other than just one sample (which I've noticed sounds very good with negative feedback).
7. I have also added an input follower, which changes the filter frequencies depending on the input volume.  The Attack and Release knobs change how quickly that value fluctuates.
8. And, as is obvious, you can control the input/output volumes with the faders, and there is a built-in Wet/Dry knob as well with a default value of 50% (mixing the dry with the wet is very important for phasers).

I should also mention that because so many filters need to be used, I very aggressively optimized the allpass filters to the point that they only need an impressive two additions, two subtractions, and two multiplications each(!!).  Most Phaser controls are also sample-exact, which is a fun bonus.

I'm excited for everybody to get their hands on this tool.  It sounds very cool, and can easily add some deliciously vocal characteristics to your sounds.


plugins/Phaser/Phaser.h
```diff
@@ -0,0 +1,85 @@
...
+	inline float detuneWithOctaves(float pitchValue, float detuneValue);
+	void changeSampleRate();
+	void restartLFO();
+	void calcAttack();
+	void calcRelease();
+
+	PhaserControls m_phaserControls;
+
+	float m_filtX[32][2][4] = {{0}};// [filter number][channel][samples back in time]
+	float m_filtY[32][2][4] = {{0}};// [filter number][channel][samples back in time]
```
PhysSong: Needs one more pair of braces respectively.

plugins/Flanger/QuadratureLfo.h
```diff
@@ -30,51 +30,66 @@
...
 
-	void tick( float *s, float *c );
+
+	inline void setOffset( double offsetVal )
+	{
+		m_offset = offsetVal;
+	}
+
+
+	void tick( float *s, float *c )
```
PhysSong: I think the original intention for those names was to represent sine and cosine functions. It might be better to rename them(maybe `l` and `r`?).

plugins/Flanger/QuadratureLfo.h
```diff
@@ -30,51 +30,66 @@
...
+		m_phase += m_increment;
+
+	}
 
 private:
 	double m_frequency;
 	double m_phase;
 	double m_increment;
 	double m_twoPiOverSr;
+	double m_srOverTwo;
```
PhysSong: Do we really need this variable?

PhysSong:
I think the changes in the flanger plugin should be decoupled before merging this. Also, I think `Phaser.h` and `Phaser.cpp` can be renamed to `PhaserEffect.h` and `PhaserEffect.cpp`

plugins/Flanger/QuadratureLfo.h
DouglasDGI: You're right, my brain saw that and saw am unnecessarily small performance enhancement but it failed to realize that it's only when the setFrequency function is called, which probably won't be called every loop anyway... I think I did it because m_twoPiOverSr appears to exist for the same reason.  I'll revert it.
(And while I'm at it, I'll make it so setFrequency isn't called every loop in the Phaser, I didn't catch that...)

DouglasDGI:
@PhysSong Alright, I fixed all of those.
When you say you want the Flanger changes separated from the Phaser changes, does that include the QuadratureLfo changes?  The Phaser relies on the changes to the QuadratureLfo files, which is why I put it all in one PR.

plugins/Phaser/PhaserControls.cpp
```diff
@@ -0,0 +1,138 @@
...
+	connect( &m_phaseModel, SIGNAL( dataChanged() ), this, SLOT( updatePhase() ) );
+
+	m_outPeakL = 0.0;
+	m_outPeakR = 0.0;
+}
+
+
+PhaserControls::~PhaserControls()
+{
+	disconnect( gui->mainWindow(), SIGNAL( periodicUpdate() ), 0, 0 );
```
PhysSong: Not needed anymore.

DouglasDGI: Whoa!  How did that sneak its way into there?  I'm sure I deleted that.  Weird.

DouglasDGI:
All merge conflicts and bugs have been removed. Feel free to test now.  
Here's a video of me messing around with this effect: https://youtu.be/hoD1MKwEH9k

DouglasDGI:
I just added some major enhancements to this Phaser effect, including both QOL tweaks and some new features.  Here are some of the changes, though I'm sure I won't remember them all...

- Changed one of the constants to drastically improve the input followers attack/release curves
- Added an "Invert" button which inverts the wet signal.  This effectively shifts the signal by 180 degrees, inverting the notches and peaks created by the phaser when combined with the dry signal.
- Added a "Wet" button which isolates the wet signal
- Added some gentle waveshaping distortion to the feedback loop to prevent infinite feedback
- Added a distortion knob which controls a full-wave rectifier in the feedback loop.
Why would I put a full-wave rectifier in a feedback loop?
  1. I needed a distortion type that made a noticeable change
to the sound that wouldn't result in infinite feedback loops.
An overdriven symmetrical distortion wouldn't work for that
reason.
  2. I just tried this strange idea as an experiment, and,
surprisingly, it sounded fantastic.
- Commented the code a bit better, including a diagram of what's going on.
- Some style changes I missed
- Rewrote how the filter feedback works in a way that allows 0-delay feedback.  I got the idea from this book: https://www.native-instruments.com/fileadmin/ni_media/downloads/pdf/VAFilterDesign_1.1.1.pdf
- Made the feedback delay be in milliseconds rather than samples, making it smoother to automate, and 100% sample rate change compatible.

This is 100% ready for review, then merge.  I do not intend to make any other changes to it feature-wise, unless you guys have any suggestions.

PhysSong:
@DouglasDGI Your last commit changed the permission of some files from `644` to `755` accidentally.

DouglasDGI:
@PhysSong Oh, that got messed up when I made a backup of my files before reinstalling my OS.  I didn't realize changing permissions mattered with this stuff.  I have a few questions about fixing that, I'll message you about it on Discord (later when I have time).

DouglasDGI:
Welp, that was a mess... somehow all of the files in my backup got set to 755 and I have no idea how.  I managed to copy all of the file permissions over from an older backup, I think that fixed everything.
I also noticed a bug in my code where resetLFO and changeSampleRate weren't labeled as slots so they didn't do anything, I fixed that.
I... think everything's okay now?  Hope everything's okay now?

plugins/Phaser/PhaserEffect.h
```diff
@@ -0,0 +1,104 @@
...
+	PhaserEffect(Model * parent, const Descriptor::SubPluginFeatures::Key * key);
+	virtual ~PhaserEffect();
+	virtual bool processAudioBuffer(sampleFrame * buf, const fpp_t frames);
+
+	virtual EffectControls* controls()
+	{
+		return &m_phaserControls;
+	}
+
+	float m_realCutoff[2] = {0, 0};
```
DomClark: Rather than having `m_realCutoff` be public, it would better to keep it private and expose it through a const getter function.

plugins/Phaser/PhaserEffect.h
```diff
@@ -0,0 +1,104 @@
...
+	void calcFeedback();
+	void calcOutGain();
+	void calcInGain();
+	void calcDelay();
+	void calcPhase();
+	void changeSampleRate();
+	void restartLFO();
+
+private:
+	inline void calcAllpassFilter(sample_t &outSamp, sample_t inSamp, sample_rate_t Fs, int filtNum, int channel, float b0, float b1);
```
DomClark: `calcAllpassFilter` needn't be `inline`, and `outSamp` can be the return value, rather than an output parameter.

plugins/Phaser/PhaserEffect.h
```diff
@@ -0,0 +1,104 @@
...
+	inline float detuneWithOctaves(float pitchValue, float detuneValue);
```
DomClark: `detuneWithOctaves` needn't be `inline`, and can be `static`.

plugins/Phaser/PhaserEffect.h
```diff
@@ -0,0 +1,104 @@
...
+
+	PhaserControls m_phaserControls;
+
+	float m_filtX[32][2][4] = {{{0}}};// [filter number][channel][samples back in time]
+	float m_filtY[32][2][4] = {{{0}}};// [filter number][channel][samples back in time]
```
DomClark: These only need 2 for their third dimension. `m_filtX[...][...][0]`, `m_filtX[...][...][3]` and `m_filtY[...][...][3]` are never touched, and `m_filtY[...][...][0]` is overwritten each time, so doesn't need preserving and can be local to `calcAllpassFilter` instead.

plugins/Phaser/PhaserEffect.h
```diff
@@ -0,0 +1,104 @@
...
+	std::vector<float> m_filtDelayBuf[2];
+	int m_delayBufSize;
+	float m_filtFeedbackLoc = 0;
```
DomClark: `m_filtFeedbackLoc` can be an integer.

plugins/Phaser/PhaserEffect.h
```diff
@@ -0,0 +1,104 @@
...
+
+	QuadratureLfo * m_lfo;
+
+	RmsHelper * m_rms [2];
+
+	float m_twoPiOverSr;
+
+	float m_b0[2];
+	float m_b1[2];
```
DomClark: `m_b0` and `m_b1` can be local to `processAudioBuffer`. Same for `m_firstAdd` and `m_secondAdd` further down.

plugins/Phaser/PhaserEffect.h
```diff
@@ -0,0 +1,104 @@
...
+
+class PhaserEffect : public Effect
+{
+	Q_OBJECT
+public:
+	PhaserEffect(Model * parent, const Descriptor::SubPluginFeatures::Key * key);
+	virtual ~PhaserEffect();
+	virtual bool processAudioBuffer(sampleFrame * buf, const fpp_t frames);
+
+	virtual EffectControls* controls()
```
DomClark: Here `virtual` can be omitted, and `override` should be specified. The same goes for virtual functions in other files (again, don't worry about fixing up old code in the flanger).

DomClark (commented):
Most of the code looks good; I haven't checked that the DSP code does what it ought to - I believe you know what that is better than I do.

I think there ought to be a nicer way to share code between plugins than the ugly `#include "../other_plugin/header.h"`, but that's out of scope here.

plugins/Phaser/PhaserEffect.h
DouglasDGI: Oh, oops, that was me copy+pasting some filter code I used elsewhere.  Nice catch.

plugins/Flanger/FlangerControls.cpp
```diff
@@ -88,3 +93,11 @@ void FlangerControls::changedPlaybackState()
 {
 	m_effect->restartLFO();
 }
+
+
+
+
+void FlangerControls::updatePhase()
+{
+	m_effect->m_lfo->setOffset(m_lfoPhaseModel.value() / 180 * D_PI);
```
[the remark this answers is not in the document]
DouglasDGI: I did that a while back because putting it as FlangerEffect didn't work, I didn't know back then that the cause was FlangerEffect.h not being listed under MOCFILES in CMakeLists.txt.  Will fix.

plugins/Phaser/PhaserEffect.cpp
```diff
@@ -0,0 +1,463 @@
...
+PhaserEffect::PhaserEffect(Model* parent, const Descriptor::SubPluginFeatures::Key* key) :
+	Effect(&Phaser_plugin_descriptor, parent, key),
+	m_phaserControls(this)
+{
+	m_currentPeak[0] = m_currentPeak[1] = PHA_NOISE_FLOOR;
+
+	m_lfo = new QuadratureLfo(Engine::mixer()->processingSampleRate());
+
+	m_rms[0] = new RmsHelper(64 * Engine::mixer()->processingSampleRate() / 44100);
+	m_rms[1] = new RmsHelper(64 * Engine::mixer()->processingSampleRate() / 44100);
```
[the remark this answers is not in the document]
DouglasDGI: I somehow missed the fact RmsHelper::setSize exists.  That'll fix this issue.

plugins/Phaser/PhaserEffect.cpp
```diff
@@ -0,0 +1,463 @@
...
+		return false;
+	}
+
+	sample_rate_t sample_rate =  Engine::mixer()->processingSampleRate();
+
+	double outSum = 0.0;
+	const float d = dryLevel();
+	const float w = wetLevel();
+
+	// The following are sample-exact.
```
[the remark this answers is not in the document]
DouglasDGI: The other values aren't pulled directly from the models.  They're instead stored in variables which are updated whenever the corresponding knobs are moved.  This results in *enormous* performance benefits, especially for the attack and release coefficients which require an `exp10`.

I don't know whether those are sample-exact, they might be.  They use `Qt::DirectConnection`, I'm unsure of how instant that is.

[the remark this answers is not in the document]
PhysSong: > This results in _enormous_ performance benefits

Do you know how much it is?

PhysSong:
@DouglasDGI What's the status of this PR?

DouglasDGI:
Probably ready for merge, unless there's anything you see that needs to be fixed.

plugins/Phaser/PhaserControlDialog.cpp
```diff
@@ -0,0 +1,190 @@
...
+}
+
+
+void PhaserControlDialog::updateSliders()
+{
+	// Magic. Do not touch.
+	m_cutoffDotLeftLabel->move((int((log2(m_controls->m_effect->getCutoff(0)) - log2(PHA_MIN_FREQ)) /
+		log2(PHA_MAX_FREQ / PHA_MIN_FREQ) * PHA_DOT_SLIDER_LENGTH)) + 12, 139);
+	m_cutoffDotRightLabel->move((int((log2(m_controls->m_effect->getCutoff(1)) - log2(PHA_MIN_FREQ)) /
+		log2(PHA_MAX_FREQ / PHA_MIN_FREQ) * PHA_DOT_SLIDER_LENGTH)) + 12, 139);
```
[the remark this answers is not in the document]
DouglasDGI: > Also, I'm curious why you chose it instead of log(base e).

The formula used makes the visualizer exponential based on octaves/semitones/cents rather than just using Euler's number.  It'll make the frequency locations placed the same way as you'd see from any spectrum analyzer.

> You may use one log2 for each numerator or two for each denominator.

What do you mean by this?  Are you saying this can be simplified?

plugins/Flanger/FlangerEffect.h
```diff
@@ -43,7 +44,10 @@ class FlangerEffect : public Effect
 	{
 		return &m_flangerControls;
 	}
+
+private slots:
 	void changeSampleRate();
+	void updatePhase();
 	void restartLFO();
```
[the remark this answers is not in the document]
DouglasDGI: I figured it would be best to do it this way since it's modifying variables in PhaserEffect rather than PhaserControls.  Would it be better for me to put those functions into PhaserControls and modify the values in PhaserEffect from there?  I feel like that wouldn't look as neat.  What's the reason for keeping these things in PhaserControls?

DouglasDGI:
I made the input follower for this plugin shortly after making a compressor, so I made it in a similar way... however, the RMS is completely unnecessary.  I removed it.  The input follower should be a bit snappier now.

DouglasDGI:
I made some major updates.  Video here: [https://www.youtube.com/watch?v=_Zyd7OcZRO8](url)

1. Four Phaser modes, consisting of the Default mode, a Standard mode which is more similar to some generic phasers out there, a Nested mode which is actually a nested allpass filter, and "Inane", where any and all logic in the algorithm is aggressively defenestrated (see code comments for details).
2. "Double", which allows you to double the resulting phase shifting from the allpass filter bank.
3. "Alias", which linearly reflects all frequencies around Nyquist before and after the filters.
4. The new "Analog" phaser setting, which speaks for itself.  A distortion amount knob is also provided.
5. There is now a knob that lets you control the delay amount using the input follower and LFO, which means you can use this plugin as a Flanger as well.  This plugin is now a full-on Phaser/Flanger Frankenstein's monstrosity.

The resulting timbres you can get from this absolute unit have increased in variety substantially.

H4CKTON3 helped me with the art yet again, here is the result:
image

Ready for both review and eventually merge.  I will not be making any more changes to this plugin in the near future.

plugins/Phaser/PhaserControls.cpp
```diff
@@ -0,0 +1,143 @@
...
+	m_delayControlModel.setScaleLogarithmic(true);
+
+	m_modeModel.addItem(tr("Default"));
+	m_modeModel.addItem(tr("Standard"));
+	m_modeModel.addItem(tr("Nested"));
+	m_modeModel.addItem(tr("Inane"));
+}
+
+
+void PhaserControls::saveSettings(QDomDocument& doc, QDomElement& _this)
```
PhysSong, line 85: Please rename `_this`s in all of your plugins, see https://github.com/LMMS/lmms/pull/5731#discussion_r539041773.

plugins/Phaser/PhaserControls.h
```diff
@@ -0,0 +1,101 @@
...
+class PhaserEffect;
+
+
+class PhaserControls : public EffectControls
+{
+	Q_OBJECT
+public:
+	PhaserControls(PhaserEffect* effect);
+
+	void saveSettings(QDomDocument & _doc, QDomElement & _parent) override;
```
PhysSong, line 45: `_docs` should be `docs`, and `_parent` should be something like `parent` or `elem`.

plugins/Phaser/PhaserEffect.cpp
```diff
@@ -0,0 +1,613 @@
...
+	new PluginPixmapLoader("logo"),
+	NULL,
+	NULL
+} ;
+
+}
+
+
+constexpr float PHA_NOISE_FLOOR = 0.00001f;
+constexpr double PHA_LOG = 2.2;
```
PhysSong, line 56: Please add a comment like you explained in the PR description.

plugins/Phaser/PhaserEffect.cpp
```diff
@@ -0,0 +1,613 @@
...
+		// Calculate input follower values
+		const double sAbs[2] = {abs(s[0]), abs(s[1])};
+		for (int i = 0; i < 2; i++)
+		{
+			if (sAbs[i] > m_currentPeak[i])
+			{
+				m_currentPeak[i] = qMin(m_currentPeak[i] * m_attCoeff, sAbs[i]);
+			}
+			else
+			if (sAbs[i] < m_currentPeak[i])
```
PhysSong, line 286: Why two lines for `else if`?

plugins/Phaser/PhaserControls.cpp
```diff
@@ -0,0 +1,143 @@
...
+	m_feedbackModel(0.0f, -100.0f, 100.0f, 0.01f, this, tr("Feedback")),
+	m_orderModel(8, 0, 32, this, tr("Order")),
+	m_delayModel(0.0f, 0.0f, 20.0f, 0.001f, this, tr("Delay")),
+	m_rateModel(9.0f, 0.01f, 60.0f, 0.001f, 60000.0, this, tr("Rate")),
+	m_amountModel(4.5f, 0.f, 5.0f, 0.01f, this, tr("Amount")),
+	m_enableLFOModel(true , this, tr("Enable LFO")),
+	m_phaseModel(180.0f, 0.f, 360.0f, 0.01f, this, tr("Phase")),
+	m_inFollowModel(0.f, -30.f, 30.0f, 0.01f, this, tr("Input Follow")),
+	m_attackModel(200.0f, 0.0f, 2000.0f, 1.0f, this, tr("Attack")),
+	m_releaseModel(500.0f, 0.0f, 2000.0f, 1.0f, this, tr("Release")),
```
zonkmachine, line 56: m_attackModel and m_releaseModel both have a minimum of 0.0f which leads to division with 0 in calcAttack() and calcRelease() respective. Choose a minimum value here or guard against 0 in mentioned functions. I suggest a minimum value here.

zonkmachine (dismissed):

plugins/Phaser/PhaserEffect.cpp
```diff
@@ -0,0 +1,613 @@
...
+
+			Note that the DC offset added here by this distortion is
+			removed later on.
+			*/
+			const float trueDistVal = distortion * 1.5;
+			tempAdd[0] = abs(tempAdd[0] + 1.5 - trueDistVal) - 1.5 + trueDistVal;
+			tempAdd[1] = abs(tempAdd[1] + 1.5 - trueDistVal) - 1.5 + trueDistVal;
+		}
+
+		// Remove DC Offset
```
PhysSong, line 505: https://github.com/LMMS/lmms/pull/5731#discussion_r539043436

plugins/Phaser/PhaserEffect.cpp
```diff
@@ -0,0 +1,613 @@
...
+		
+		/*
+		Precalculate the allpass filter coefficients.
+		Because of very aggressive CPU optimization, some of
+		the coefficient names are inaccurate concerning what
+		they actually do.
+		*/
+		float b0[2];
+		float b1[2];
+		for (int b = 0; b < 2; ++b)
```
PhysSong, line 317: Using `a`, `b`, `c`  for loop variables is probably unexpected. Use `i`, `j`, `k` or something meaningful like `ch`.

plugins/Phaser/PhaserControls.cpp
zonkmachine: Edit. The AppImage doesn't see any problems because of this but we have built-in debug stuff that catches floating point errors. If we don't appease the machine chaos will ensue.

plugins/Phaser/PhaserEffect.cpp
```diff
@@ -0,0 +1,613 @@
...
+				tempAdd[0] = delayOut[0] * feedback;
+				tempAdd[1] = delayOut[1] * feedback;
+				for (int a = 0; a < order; ++a)
+				{
+					for (int b = 0; b < 2; ++b)
+					{
+						for (int c = 0; c < 1 + doubleVal; ++c)// Runs twice if doubleVal is true
+						{
+							tempAdd[b] = calcAllpassFilter(tempAdd[b], sample_rate, a, b, b0[b], b1[b]);
+						}
```
PhysSong, line 376: Will this work as expected? Calling `calcAllpassFilter` updates the state variables in `m_filtX[a][b]` and `m_filtY[a][b]`. If you're going to run the filter twice, shouldn't you use a separate set of state variables to prevent feeding multiple samples for a filter state when only one real input sample is given?

plugins/Phaser/PhaserEffect.cpp
```diff
@@ -0,0 +1,613 @@
...
+			secondAdd[1] *= m_aliasFlip;
+		}
+		for (int a = 0; a < order; ++a)
+		{
+			for (int b = 0; b < 2; ++b)
+			{
+				for (int c = 0; c < 1 + doubleVal; ++c)// Runs twice if doubleVal is true
+				{
+					secondAdd[b] = calcAllpassFilter(secondAdd[b], sample_rate, a, b, b0[b], b1[b]);
+				}
```
PhysSong, line 423: Same here.

zonkmachine:
Testing it and it's brilliant!

I think the Envelope Follower should have an enable button like the LFO.
What does +/-30 octaves mean on the Envelope Follower Amount knob?

plugins/Phaser/PhaserEffect.cpp
PhysSong: Also, I think the way you store state variables can be improved. Here's an example:
```cpp
struct FilterState
{
	float x[2];
	float y[2];
};
FilterState m_filt[32][2];
```
Also, I suggest you reading existing filter classes in `include/BasicFilters.h` and see if any of them fits your needs.

plugins/Phaser/PhaserControls.cpp
DouglasDGI: Oh, all the coding style violations I missed are because I copy+paste LMMS's Amplifier plugin to start off with a plugin template.  I'll fix those.

plugins/Phaser/PhaserEffect.cpp
DouglasDGI: I'm not sure what you're referring to, are you talking about this?
> and "Inane", where any and all logic in the algorithm is aggressively defenestrated (see code comments for details).

If so, scroll down and you'll see a very large comment which shows all the Phaser feedback circuit modes and explains what they do.

plugins/Phaser/PhaserEffect.cpp
DouglasDGI: > Will this work as expected?

Refer to the large comment I made explaining the various feedback circuits.  Both filters using the same buffer is intentional, for the appropriately-named "Inane" mode.  The extremely odd behavior is exactly what I was going for, and can result in some very cool resonant screeches and unexpected timbres.

DouglasDGI: Actually, I may have misunderstood what you were talking about.
A separate set of state variables is not required, since the input `tempAdd[b]` is also the output `tempAdd[b]`.

DouglasDGI: (Also, that all being said, the BasicFilters.h filters can't be used, or else the fan-favorite Inane mode will have to be removed due to its odd behavior, and keeping things ludicrously optimized is fun anyway.)

DouglasDGI: ...wait, I'm wrong.  You're right.  It'll need its own buffer for it to be two separate filters.  I was wondering why the results were more interesting and cool than I expected.
I'll update the tooltip so it's accurate.

plugins/Phaser/PhaserEffect.cpp
PhysSong: I mean what is `PHA_LOG` and why it's `2.2`.

DouglasDGI: Oh, I see.  I'm actually not sure, it's just something I found in many different documentation sources on compressor attack/release coefficient calculation.  I've seen both "2.2" and "1" used.
You can find multiple papers that use the number 2.2 from a quick Google search, e.g. [this one](http://www2.ece.rochester.edu/~zduan/teaching/ece472/projects/2016/White_Dick_paper.pdf), [this one](http://www.cs.tut.fi/~sgn14006/PDF/L05-dynamics.pdf), and [this one.](https://audioordeal.co.uk/how-to-build-a-vst-compressor/)

I know that using "1" instead makes it so the attack time is how long it takes for the signal to reach `1 - 1/e` volume, so maybe 2.2 moves that over to a more meaningful value like 90% or something like that.

In any case, it's definitely a standard value, but I'm not actually entire certain of why it is chosen.

PhysSong: Now I remember I saw something related on your compressor. A simple comment will help future developers to understand what it is.

plugins/Phaser/PhaserEffect.cpp
DouglasDGI: I've decided to leave the Double feature as it is.  From what I can tell, it ends up being equivalent to running 88200 Hz audio through a filter calculated for 44100 Hz audio, and then outputting 44100 Hz audio... but what matters is that it sounds cool.  :D 
My original idea also was not a standard or orthodox phaser feature (like most of the new things I added...), and once I implemented it the way I originally intended, it sounded too close to just doubling the filter order for it to be interesting at all.  
Instead of running the audio through two filters per loop, it runs every input through the filter twice.  "Double" still makes sense here, so I've decided to keep it.

DouglasDGI:
Alright, I think that covers everything.
> What does +/-30 octaves mean on the Envelope Follower Amount knob?

@zonkmachine +1 octave means the input follower detecting an input volume of 0 dbFs (taking attack/release into account) will increase the Phaser frequency by 1 octave.
Also, I don't know how to resolve your "requested change", am I supposed to "dismiss" it or do you have to do something on your end?  I'm still learning Github, heh.
> I think the Envelope Follower should have an enable button like the LFO.

Done.  Good suggestion, I never bothered adding one since setting the Input Follower amount to 0 does the same thing, but that's rather inconvenient.

PhysSong:
I got a ton of uninitialized variable errors while testing with Valgrind because `QuadratureLfo` constructor doesn't initialize `m_phase` and `m_frequency`. I got them from the existing Flanger plugin, too. It seems like the problem existed before you touched anything, though.

DouglasDGI:
@PhysSong Just changed it so m_frequency and m_phase are set to 0 in the constructor, I assume that fixes the issue.

curlymorphic:
This is a really nice update, the old flanger has a new lease of life and sounds much better, it's not often that 3Osc can make me smile on the default patch, but with your upgrades to the flanger I did. I have to be honest, testing this PR has made me somewhat embarrassed at the horrid sound made when adjusting the delay knob, something I shall be looking into.

The Phaser, nice work, the UI look clean and well laid out, sounds good, and is fun to play with. This will be a nice new addition to LMMS. 

> I think the changes in the flanger plugin should be decoupled before merging this. 

I think this is very important, we have 2 very distinct updates in this one pull request. This makes reviewing, testing, maintenance and commits logs very messy.  I don't think it will take too much effort to separate these into individual pull requests, I am happy to provide any assistance you want.

My proposal would be to start a new PR with the Flanger updates, due to the phaser being dependant on these changes. I would take the opportunity to refactor any DPS classes that are used in both plugins and relocate them within the core codebase, in preparation for the Phaser. I feel this would mainly be a copy-paste exercise, so all the previous work put into reviews will remain valid, I may have a few code comments but let's get the code separated first.

Progression onto the Phasor will then be a lot simpler, as I am hoping the number of files changes drops from the current significant 29 to a more manageable level. We can look at the details as soon as the flanger PR is complete.

rdrpenguin04:

PhysSong:
@curlymorphic I extracted the Flanger changes to #5873.

plugins/Phaser/PhaserEffect.cpp
```diff
@@ -0,0 +1,623 @@
...
+
+
+extern "C"
+{
+
+Plugin::Descriptor PLUGIN_EXPORT Phaser_plugin_descriptor =
+{
+	STRINGIFY(PLUGIN_NAME),
+	"Phaser",
+	QT_TRANSLATE_NOOP("pluginBrowser", "A versatile phaser plugin"),
```
PhysSong, line 43: See #5629:
```suggestion
	QT_TRANSLATE_NOOP("PluginBrowser", "A versatile phaser plugin"),
```

plugins/Phaser/PhaserEffect.h
```diff
@@ -0,0 +1,110 @@
...
+ *
+ */
+
+
+#ifndef PHASER_H
+#define PHASER_H
+
+#include "PhaserControls.h"
+
+#include "../Flanger/QuadratureLfo.h"
```
PhysSong, line 31: This should be changed to "QuadratureLfo.h" once #5873 is merged. You should then revert changes in in `Flanger` before merging upstream `master`.

PhysSong:
Due to the changes in #5873, this PR needs to be updated.
@DouglasDGI Do you need some help, or you can handle the conflict? Please see https://github.com/LMMS/lmms/pull/5168#discussion_r563227572 if you want to try it yourself.

PhysSong:
@curlymorphic Do you want to re-review this PR?

plugins/Phaser/PhaserEffect.cpp
```diff
@@ -0,0 +1,623 @@
...
+		}
+
+		/*
+		Some gentle feedback distortion to prevent infinite loops.
+
+		I used tanh(x/2)*2 rather than just tanh(x) because tanh(x)
+		decreased the volume too much for my liking.
+		*/
+		tempAdd[0] = tanh(tempAdd[0] * 0.5) * 2;
+		tempAdd[1] = tanh(tempAdd[1] * 0.5) * 2;
```
curlymorphic, line 491: Has the aliasing been considered? although I feel this``` tanh``` needs to remain, as it is much preferable to the infinite signal rise that could happen if removed

plugins/Phaser/PhaserEffect.cpp
```diff
@@ -0,0 +1,623 @@
...
+
+void PhaserEffect::restartLFO()
+{
+	m_lfo->restart();
+}
+
+
+
+// Takes input of original Hz and the number of cents to detune it by, and returns the detuned result in Hz.
+float PhaserEffect::detuneWithOctaves(float pitchValue, float detuneValue)
```
curlymorphic, line 605: The names are not clear what units ```detuneValue`` is in, the description says cents, the function name says octaves.

curlymorphic (commented):
My previous comment about spitting this pull request, the flanger changes being separated from this new plugin, but more active devs may have a different opinion.

plugins/Phaser/PhaserEffect.cpp
DouglasDGI: Ouch, my bad.  I copy+pasted it from other code I had that used cents instead of octaves.

plugins/Phaser/PhaserEffect.cpp
```diff
@@ -0,0 +1,623 @@
...
+						{
+							tempAdd[j] = calcAllpassFilter(tempAdd[j], sample_rate, i, j, b0[j], b1[j]);
+						}
+					}
+				}
+			}
+		}
+		if (analog)
+		{
+			firstAdd[0] = tanh(s[0] * analogDist) / analogDist + tanh(tempAdd[0] * analogDist) / analogDist;
```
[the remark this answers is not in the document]
DouglasDGI: It isn't called "analog" because the distortion is analog, it's called "analog" because having distortion inside of the feedback loop is a characteristic of analog audio hardware.  Like how Moog filters work by adding nonlinearities (distortion) into the filter feedback loops.

That being said, I do agree that aliasing should be avoided... but I am of the opinion that oversampling should be handled by LMMS itself rather than individual plugins, and that's out-of-scope for this plugin.  I think every plugin should come with oversampling options, just like how every plugin comes with a Wet/Dry knob.

So, unless you disagree, I don't think we should bother with putting oversampling into here.  I didn't put it into any of my other plugins for the same reason.  If we implement oversampling as an LMMS feature rather than a per-plugin feature, we'll never have to worry about it ever again... because, in reality, over 90% of the plugins in the entire catalog could benefit from oversampling, so adding it in on a per-plugin basis seems quite unnecessary.

plugins/Phaser/PhaserEffect.cpp
```diff
@@ -0,0 +1,623 @@
...
+		}
+		else
+		{
+			m_currentPeak[0] = 0;
+			m_currentPeak[1] = 0;
+		}
+
+		// Calculate real allpass filter frequencies
+		float lfoResults[2] = {0, 0};
+		if (enableLFO)
```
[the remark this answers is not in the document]
DouglasDGI: Good catch.  Fixed.

plugins/Phaser/PhaserEffect.cpp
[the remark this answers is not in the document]
curlymorphic: Aliasing is a very important fundamental concept in DSP, and has many reasons to be avoided, harmonic sounds become inharmonic, sound out of tune when played with other notes, fill the audio spectrum with unwanted content that makes final mixdowns muddy, and are an indicator of the quality of the product. I am aware that historically aliasing has not been of a high priority in LMMS, but I am not convinced adding to the problem would be constructive.

It is good practice to only oversample where necessary, due to the increased processing, If LMMS were to have a user option to oversample by plugin, this would be very inefficient due to the whole signal path in the plugin being processed multiple times, and expecting the artist to have an understanding of DSP and the consequences of setting the oversample rate. This is why I am not aware of any reputable DAW that takes this approach. It is always up to the plugin developer to handle anti-aliasing in an efficient manner.

That leaves the issue of how to progress with this plugin, specifically the "Analog distortion", I have a few suggestions, any one of these alone would solve this issue.

1. Oversample and decimate the non linear processing (tanh). 
2. Rename the option from "Analog" to better describe the outcome, I would suggest a name that makes it clear that the aliasing is intended, as you have done elsewhere.
3. Leave it as is, and hope no one notices or cares.

I shall leave the decision up to yourself, and shall not raise this issue again on this PR, unless asked.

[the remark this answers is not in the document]
PhysSong: I agree that we should try to reduce aliasing, but there are some issues.
> Oversample and decimate the non linear processing (tanh).

AFAIK, that add some delays around the non-linear processing. Wouldn't that affect the outcome in a way other than reducing aliasing?
Slightly out of scope, but I think filters may also leverage from oversampling for better frequency/phase response.

> Rename the option from "Analog" to better describe the outcome

That might be a good choice, but I have no ideas on better names.

LostRobotMusic:
This PR has been replaced by #6540.

PR #5170, opened 2019-09-04
Set focus on Automation Editor editor instead of window.

CYBERDEViLNL:
Set focus on Automation Editor editor instead of window so keyPressEvent's can be processed. This makes shortcuts work without manually have to click somewhere in the grey area of the editor itself.

### DIFF
--- a/src/gui/editors/AutomationEditor.cpp
+++ b/src/gui/editors/AutomationEditor.cpp
@@ -198,6 +198,8 @@ AutomationEditor::AutomationEditor() :
 	setCurrentPattern( NULL );
 
 	setMouseTracking( true );
+	setFocusPolicy( Qt::StrongFocus );
+	setFocus();
 }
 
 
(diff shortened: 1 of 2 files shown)


CYBERDEViLNL:
@PhysSong, what about moving `keyPressEvent` to `AutomationEditorWindow`? Even with this change when the editor has focus and you click the window title it loses focus and thus still doesn't process `keyPressEvent`'s.

PhysSong:
It seems like `PianoRollWindow::focusInEvent()` gives the focus to its editor as well. We may add something like this to `AutomationEditorWindow` or even `Editor`.

CYBERDEViLNL:
Yes that did it! It now always processes the `Editor` it's `keyPressEvent`'s :-)

src/gui/editors/AutomationEditor.cpp
```diff
@@ -2502,7 +2504,7 @@ void AutomationEditorWindow::open(AutomationPattern* pattern)
 	setCurrentPattern(pattern);
 	parentWidget()->show();
 	show();
-	setFocus();
+	m_editor->setFocus();
```
PhysSong: Isn't this change redundant now?

CYBERDEViLNL: Yes you are right.

PR #5171, opened 2019-09-04
Removed the excessive margin in instruments' GUI (#5129)

### LINKED ISSUE #5129: Instrument window's MIDI tab is wider than the rest
In master, the MIDI tab has grown in width. The remaining tabs have stayed the same, however, leading to an awkward margin on the right hand side.

image

Tested on Win 10, 2560x1440 screen.

artur-twardowski:
Fixes #5129.

Two changes had to be made:
- the label describing the meaning of "master pitch" was displayed with larger font which expanded the window. After reducing the pitch the MIDI tab is actually the widest one.
- spacing between LcdSpinBoxes had to be slightly reduced. I don't actually understand why, because the spacing was the same as in stable-1.2 branch, where the problem did not exist.

Additionally, "CHANNEL" label under the first LcdSpinBox was actually too wide, it expanded the layout's cell causing uneven spacing between the widgets. For Polish language, the caption of "VELOCITY" was translated to a huge string "GŁOŚNOŚĆ UDERZENIA", which again expanded severily the whole window. Moreover, the original captions were very close to each other, with very little spacing between them.
To fix these problems I shortened "CHANNEL" to "CHAN", "VELOCITY" to "VELOC" and "PROGRAM" to "PROG". I believe the new mnemonics are still explanatory. I also adjusted the locale files, shortening the foreign names where applicable and essential.


Spekular:
I believe the translations will need to be fixed on Transifex, rather than in the repo. Otherwise they'll probably be overriden on the next translation sync.

Ideally there'd also be a note attached to the string explaining that it needs to be short, via a translation comment or disambiguation: https://doc.qt.io/qt-5/i18n-source-translation.html

Spekular:
Tested and working, no longer touches the translation files, and has translator comments. LGTM.

qnebra:
For me this is fine as quick and simple workaround of this issue. In my opinion better solution will be increase width of other tabs, but it can be made in future in another PR.

Spekular:
Thanks for the fix @artur-twardowski, and sorry it took so long to get merged!

PR #5173, opened 2019-09-06
Remove "What's This?" leftovers

Sawuare:
A follow-up to  #4128.


Sawuare:
I got excited when I saw `whatsthis.png` and removed them thoughtlessly, turns out they are used somewhere. 😅

Sawuare:
I believe this is it and there is nothing left.

PhysSong:
> I got excited when I saw `whatsthis.png` and removed them thoughtlessly, turns out they are used somewhere. :sweat_smile:

It's here:
https://github.com/LMMS/lmms/blob/84d1091100d60708ba4d3def92c0b9fd17112d74/src/core/DataFile.cpp#L1498-L1510

We may rename it instead if needed.

Sawuare:
> We may rename it instead if needed.

I think it's fine as-is. Let it be a vestige of the extinct feature. 😄

Sawuare:
Merge?

Issue #5169, opened 2019-09-04
Shift + Scroll Wheel does not scroll space horizontally


AndyThePie:
I don't know how to describe this well, but I'll try:

You can't scroll the LMMS window horizontally with shift+scroll wheel (this works in the song editor, though...)
The only way I can scroll the window horizontally is with the scroll bar at the bottom or with a trackpad.

(and just in case anyone's confused about "Scrolling the LMMS Window", I mean when an embedded window goes off-screen, and the scroll bars appear in the main window.)

(I am running Windows 10 1903 w/ LMMS 1.2.0)

BaraMGB:
I suppose you mean the QMdiArea.

AndyThePie:
I don't know the correct names for the UI elements, but I hope this poorly highlighted images helps:
![image](https://user-images.githubusercontent.com/25379179/64393312-cc7f4580-d01e-11e9-90f4-9a727606fce5.png)

Veratil:
I've been playing with this since I saw it reported just to learn a little more Qt stuff. It's `MainWindow`'s `m_workspace` that needs to be modified for this as far as I can tell. I've been playing around with it right now, but for some reason I only get `wheelEvent`s when the window can't scroll any more. If there's ability to scroll I don't get them.

DomClark:
Alt+scroll works for me with LMMS 1.2.2 on Windows 10.

PR #5174, opened 2019-09-06
WIP: Add horizontal scrolling to MainWindow QMdiArea

Veratil:
WIP Fix for #5169.

My first attempt was to capture `MainWindow`'s `wheelEvent` but there were times when that wouldn't capture the event properly. I took the step to subclass QMdiArea just to capture `wheelEvent`s for this purpose.

Currently scrolling works unless you are hovering over the Song Editor or Piano Roll. I'm looking into how to handle those cases.

Vertical and horizontal scrolling will double speed when using the Control key. I haven't looked into how much more this should be, but it's a modifier to the scroll wheel angle delta. Subtraction is used because of positive/negative angle delta values for scrolling backwards/forwards.

By Qt's documentation, most mouse steps are 15 degrees (what I get when I scroll is +/-15), but some may be smaller. In those cases this will scroll slower, so we could use a constant number or figure out another approach to use.

I'm not certain if this will work on track pad's scrolling, I don't have one to test with.


Veratil:
I've made the `wheelEvent` ignored if scrolling won't do anything in the Song Editor and Piano Roll which properly bubbles up to `MainWindow`.

Only issue right now that I can't see why it's happening is in the Song Editor when horizontal scrolling is at the limits it will scroll up/down even with Shift held. I'm not sure if this is a bug with Qt or LMMS's handling of the event.

Spekular:
Vertical scroll worked before you added this, no? I assume something like this is happening:

- Horizontal scrolling hits it's limit
- Since horizontal scrolling isn't possible, scroll event "bubbles up"
- Whatever handler performed vertical scroll before this PR is still active "above" yours and recieves the bubbled message
- Vertical scroll happens as it used to

I could be wrong, but I don't think you need a new class + scroll handler to allow horizontal scroll. You should be able to override the scroll handling in the relevant class, I think? Maybe you could check for shift, delegate to the super handler if it doesn't exist, and then only handle the shift case...

Veratil:
> Vertical scroll worked before you added this, no?

I actually didn't test if this was before behavior, but it's improper IMO either way.

> I could be wrong, but I don't think you need a new class + scroll handler to allow horizontal scroll. You should be able to override the scroll handling in the relevant class, I think?

I tried this first, you can see in my opening comment about my tests with it.

Veratil:
I believe I've found the cause. `TrackContainerView` handles the event first, then passes it to `SongEditor`.
```cpp
	// always pass wheel-event to parent-widget (song-editor
	// bb-editor etc.) because they might want to use it for zooming
	// or scrolling left/right if a modifier-key is pressed, otherwise
	// they do not accept it and we pass it up to QScrollArea
	m_trackContainerView->wheelEvent( _we );
```
 Once it's ignored in `SongEditor` it returns to `TrackContainerView` which since it's ignored it passes it to its `QScrollArea::wheelEvent` which scrolls it up/down until it can't, which then passes up to the `QMdiArea` of `MainWindow`.

I believe if I add the horizontal scroll to this area it should fix the issue, and I'll have to make sure it doesn't mess anything else up. 😩

Veratil:
So after some debugging, I've found that `QAbstractScrollArea::wheelEvent` is still using the now obsolete `orientation()` which isn't handling the event properly.

```cpp
void QAbstractScrollArea::wheelEvent(QWheelEvent *e)
{
    Q_D(QAbstractScrollArea);
    if (e->orientation() == Qt::Horizontal)
        QApplication::sendEvent(d->hbar, e);
    else
        QApplication::sendEvent(d->vbar, e);
}
```
Getting the `e->orientation()` returns `Qt::Vertical`, but getting `e->modifiers()` returns `0x2000000` which is `Qt::ShiftModifier`. AFAICT this means the orientation for the event isn't getting set correctly anymore. So I've modified `TrackContainerView` to create a new `QWheelEvent` copying everything from the original, only changing the orientation to `Qt::Horizontal` and it handles the scrolling properly.

Veratil:
@PhysSong @Reflexe I believe I have this at a good point enough to remove WIP. I've tested with subwindows as much as I can think that this should touch, and it seems to be working as intended.

src/gui/TrackContainerView.cpp
```diff
@@ -508,7 +508,22 @@ void TrackContainerView::scrollArea::wheelEvent( QWheelEvent * _we )
...
-		QScrollArea::wheelEvent( _we );
+		// Handle Shift+Wheel orientation of QWheelEvent
+		if (_we->modifiers() & Qt::ShiftModifier
+			&& _we->orientation() == Qt::Vertical)
+		{
+			// copy QWheelEvent but change orientation to Horizontal
+			QWheelEvent we(_we->posF(), _we->globalPosF(),
+				_we->pixelDelta(), _we->angleDelta(), _we->delta(),
+				Qt::Horizontal, _we->buttons(), _we->modifiers(),
+				_we->phase(), _we->source(), _we->inverted());
```
PhysSong: It seems like `QWheelEvent::source()` was introduced in 5.5, and `inverted()` was in 5.7. It will make copying the wheel event complicated.

PhysSong: Also, are you sure that you don't need to change the return values of `pixelDelta()`, `angleDelta()` and `delta()`?

Veratil: > It seems like `QWheelEvent::source()` was introduced in 5.5, and `inverted()` was in 5.7. It will make copying the wheel event complicated.

What version of Qt5 should we target? I'm using 5.12.4. I don't mind targeting a lower version.

> Also, are you sure that you don't need to change the return values of `pixelDelta()`, `angleDelta()` and `delta()`?

From all the testing I've done those seem to be okay and handled properly. As a note, `delta()` is obsolete in addition to the `orientation()`.

PhysSong: > What version of Qt5 should we target?

The minimum is 5.6 LTS now, but we may consider bumping to 5.9 if necessary. For now, you may check for `QT_VERSION` to support Qt 5.6.

Veratil: Alright, I'll make sure I use a constructor from at most 5.6

PhysSong: I think you should use that form only for 5.6 while keeping the code for 5.7 and upper.

Veratil: Ah, gotcha. Will do. 👍

PhysSong: And you never did.

PR #5180, opened 2019-09-09
Support compilation with MSVC 2019

DomClark:
* Qt only offers binaries up to MSVC 2017, since MSVC 2019 is binary compatible with 2017. Thus we set `MSVC_YEAR` to 2017 even for MSVC 2019.
* The default platform for MSVC 2019 is no longer Win32, but is now that of the host. Thus it is necessary to specify `CMAKE_GENERATOR_PLATFORM Win32` in `ExternalProject_Add`.


plugins/vst_base/RemoteVstPlugin32.cmake
```diff
@@ -39,6 +42,7 @@ ELSEIF(LMMS_BUILD_WIN64 AND MSVC)
 	ExternalProject_Add(RemoteVstPlugin32
 		"${EXTERNALPROJECT_ARGS}"
 		CMAKE_GENERATOR "${CMAKE_GENERATOR_32}"
+		CMAKE_GENERATOR_PLATFORM Win32
```
Reflexe: Is that needed? I think it is the default.

PhysSong: From the same link:
> For VS 2019 and above this is based on the host platform.

Reflexe: Missed that :D

DomClark: From that link,
> For VS 2019 and above this is based on the host platform.

So for VS 2019 on a 64-bit machine, it will try to build a 64-bit RemoteVstPlugin32 without that line there.

Issue #5032, opened 2019-06-13
FX controllers overlap.


Lusted4ever:
See this: ![Untitledadasdasd223122312](https://user-images.githubusercontent.com/35467995/59460559-11cd1600-8e28-11e9-8be1-aa244ed57fa1.png)
Running win10 x64.
Using current lmms build (LMMS 1.2.0 rc8)

zonkmachine:
Confirmed on lmms-1.2.0

zonkmachine:
This is leftover from https://github.com/LMMS/lmms/pull/4482. I can fix it by going over the translations again and just removing the borked lines but maybe this is best handled on Transifex?

zonkmachine:
In https://github.com/LMMS/lmms/pull/4482/files#diff-463bf9a84ee53b7ed35f64c1a918887bL1035-L1038 I just swapped the translation for the original.
```diff
--- a/data/locale/cs.ts
+++ b/data/locale/cs.ts
@@ -1034,7 +1034,7 @@ Máte-li zájem o překlad LMMS do jiného jazyka, nebo chcete-li vylepšit exis
     </message>
     <message>
         <source>FDBK</source>
-        <translation>ZP. VAZBA</translation>
+        <translation>FDBK</translation>
     </message>
```

I'm not sure whether this is a good way to do it. Some other options that gives the same result (on the user side) would be to:

Delete the translation:
```diff
     <message>
         <source>FDBK</source>
-        <translation>ZP. VAZBA</translation>
+        <translation></translation>
     </message>
```

Using the unfinished tag:
```diff
     <message>
         <source>FDBK</source>
-        <translation>ZP. VAZBA</translation>
+        <translation type="unfinished"/>
     </message>
```

Using the unfinished tag and some text/translation:
```diff
     <message>
         <source>FDBK</source>
-        <translation>ZP. VAZBA</translation>
+        <translation type="unfinished">FIXME</translation>
     </message>
```

I guess it doesn't matter that much if it's not merged back to Transifex but I don't know how we manage that side of the code. I'm thinking of going with `<translation type="unfinished"/>` when just deleting a translation that messes up the gui from now on. Any opinions on this @liushuyu ?

zonkmachine:
Closed in https://github.com/LMMS/lmms/pull/5185

PR #5185, opened 2019-09-14
Translation Fixes (Continuation of #4482)

zonkmachine:
Fixes https://github.com/LMMS/lmms/issues/5032

Fixing some formatting issues in LMMS-1.2 caused by translations.
Continuation of https://github.com/LMMS/lmms/pull/4482

Problematic translation strings replaced with `<translation type="unfinished"/>` as mentioned [here](https://github.com/LMMS/lmms/issues/5032#issuecomment-531496557) except for the [polish translation](https://github.com/LMMS/lmms/compare/stable-1.2...zonkmachine:translationfixes2?expand=1#diff-f644f02e8f755b7cb39a896fa278355fR3005-L3005) where I removed one of the words. I don't speak polish. Not a bit.

### DIFF
--- a/data/locale/pl.ts
+++ b/data/locale/pl.ts
@@ -3002,7 +3002,7 @@ You can remove and move FX channels in the context menu, which is accessed by ri
     </message>
     <message>
         <source>VELOCITY</source>
-        <translation>PRĘDKOŚĆ UDERZENIA</translation>
+        <translation>PRĘDKOŚĆ</translation>
     </message>
     <message>
         <source>ENABLE MIDI OUTPUT</source>
--- a/data/locale/ru.ts
+++ b/data/locale/ru.ts
@@ -1664,7 +1664,7 @@ Oe Ai &lt;oeai/at/symbiants/dot/com&gt;</translation>
     <message>
         <location filename="../../src/gui/widgets/EffectView.cpp" line="66"/>
         <source>W/D</source>
-        <translation>НАСЫЩ</translation>
+        <translation type="unfinished"/>
     </message>
     <message>
         <location filename="../../src/gui/widgets/EffectView.cpp" line="69"/>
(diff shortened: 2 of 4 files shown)


liushuyu (commented):
>This is leftover from #4482. I can fix it by going over the translations again and just removing the borked lines but maybe this is best handled on Transifex?

Yes, these changes will be eventually got overridden by Transifex. However, removing translations is not a decent way to do this since you will always find some languages got extra long strings somewhere.

zonkmachine:
> However, removing translations is not a decent way to do this since you will always find some languages got extra long strings somewhere.

What is the way to go about this then? Report the problematic lines on Transifex instead?
I won't have time to dive into that site. I tried once and we didn't seem to be compatible...
Dow we merge back translations from Transifex for lmms-1.2 ?

qnebra:
For me @liushuyu it was some kind of lazy translation. Translate without caring about avalaible space and line lenght.

liushuyu:
>Even if they will be removed later by transifex it will fix the formatting issues in lmms-1.2.1 that is  being prepared.

You can merge this for now.

>Should these strings be reported on Transifex each one separately and then fixed by the translators?

You can open an issue on Transifex.

zonkmachine:
> You can open an issue on Transifex.

OK. Is this one issue per string? I don't find a general issue tracker there.

liushuyu:
>Is this one issue per string?

Yes

I don't find a general issue tracker there.

>There's a per-language tracker (discussion section). This is where Transifex does not work well.

zonkmachine:
@liushuyu I have fixed the Swedish strings on transifex but for the rest I don't think I have access to even comment on the other languages (and I probably shouldn't have... :) )
I have found the LMMS translatin team however and will start a conversation there. Thanks for your patience!

zonkmachine:
> I have found the LMMS translation team 'discussions page' however and will start a conversation there.

https://www.transifex.com/lmms/teams/61632/discussions/19385/
Links and Pics and editing the post doesn't seem to work though.

PR #5196, opened 2019-09-16
Add auto-highlight scale and key selection

Veratil:
This adds a new key selection combobox to the Piano Roll next to the scale combobox. When a key+scale is selected, it will automatically highlight the scale on the selected key. If either is set back to "No key/scale" the highlights will be removed.

This doesn't remove any functionality already there with marking semitones, but is a QOL improvement saving steps when wanting to mark scales.

lmms-key-scale

lmms-key-selection

Additionally, as another small QOL improvement this links the toolbar icon with the combobox so when resized you don't get this:

lmms-toolbar-clip

### DIFF
--- a/include/PianoRoll.h
+++ b/include/PianoRoll.h
@@ -195,12 +195,13 @@ protected slots:
 	void zoomingYChanged();
 	void quantizeChanged();
 	void noteLengthChanged();
+	void keyChanged();
 	void quantizeNotes();
 
 	void updateSemiToneMarkerMenu();
 
 	void changeNoteEditMode( int i );
-	void markSemiTone( int i );
+	void markSemiTone(int i, bool fromMenu = true);
 
 	void hidePattern( Pattern* pattern );
 
@@ -311,6 +312,7 @@ private:
 	ComboBoxModel m_zoomingYModel;
 	ComboBoxModel m_quantizeModel;
 	ComboBoxModel m_noteLenModel;
+	ComboBoxModel m_keyModel;
 	ComboBoxModel m_scaleModel;
 	ComboBoxModel m_chordModel;
 
@@ -507,6 +509,7 @@ private:
 	ComboBox * m_zoomingYComboBox;
 	ComboBox * m_quantizeComboBox;
 	ComboBox * m_noteLenComboBox;
+	ComboBox * m_keyComboBox;
 	ComboBox * m_scaleComboBox;
 	ComboBox * m_chordComboBox;
 	QPushButton * m_clearGhostButton;
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -374,6 +374,13 @@ PianoRoll::PianoRoll() :
 	connect( &m_noteLenModel, SIGNAL( dataChanged() ),
 					this, SLOT( noteLengthChanged() ) );
 
+	// Set up key selection dropdown
+	m_keyModel.addItem(tr("No key"));
+	// Use piano roll note strings for key dropdown
+	for (int i = 0; i < 12; i++) { m_keyModel.addItem(s_noteStrings[i]); }
+	m_keyModel.setValue(0); // start with "No key"
+	connect(&m_keyModel, &ComboBoxModel::dataChanged, this, &PianoRoll::keyChanged);
+
 	// Set up scale model
 	const InstrumentFunctionNoteStacking::ChordTable& chord_table =
 			InstrumentFunctionNoteStacking::ChordTable::getInstance();
@@ -388,6 +395,8 @@ PianoRoll::PianoRoll() :
 	}
 
 	m_scaleModel.setValue( 0 );
+	// connect scale change to key change so it auto-highlights with scale as well
+	connect(&m_scaleModel, &ComboBoxModel::dataChanged, this, &PianoRoll::keyChanged);
 	// change can update m_semiToneMarkerMenu
 	connect( &m_scaleModel, SIGNAL( dataChanged() ),
 						this, SLOT( updateSemiToneMarkerMenu() ) );
@@ -485,10 +494,16 @@ void PianoRoll::changeNoteEditMode( int i )
 }
 
 
-void PianoRoll::markSemiTone( int i )
+void PianoRoll::markSemiTone(int i, bool fromMenu)
 {
-	const int key = m_pianoKeySelected;
+	const int key = fromMenu
+		? getKey(mapFromGlobal(m_semiToneMarkerMenu->pos()).y())
+		: m_keyModel.value() - 1;
 	const InstrumentFunctionNoteStacking::Chord * chord = nullptr;
+
+	// if "No key" is selected, key is -1, unmark all semitones
+	// or if scale changed from toolbar to "No scale", unmark all semitones
+	if (!fromMenu && (key < 0 || m_scaleModel.value() == 0)) { i = stmaUnmarkAll; }
 
 	switch( static_cast<SemiToneMarkerAction>( i ) )
 	{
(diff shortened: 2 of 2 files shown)


necrashter:
Good idea but I think I found a bug:
1. Open piano roll, mark a note.
2. Right click on the marked note and select "Mark/Unmark all corresponding octave semitones"
Program crashes with segfault. This doesn't happen in master branch.

Also, the combobox doesn't change if you mark a scale by right clicking on a note and selecting "Mark current scale". Is this the expected behavior?

PhysSong:
@necrashter Because this branch doesn't contain the changes from #5137 which is merged into `master`.

src/gui/editors/PianoRoll.cpp
```diff
@@ -404,6 +405,23 @@ PianoRoll::PianoRoll() :
 	connect( &m_noteLenModel, SIGNAL( dataChanged() ),
 					this, SLOT( noteLengthChanged() ) );
 
+	// Set up key selection
+	m_keyModel.addItem(tr("No key"));
+	m_keyModel.addItem(tr("C"));
+	m_keyModel.addItem(tr("C♮/D♭"));
```
DomClark: Shouldn't this be "C♯/D♭"? Same for the others.

Veratil: doh... I couldn't tell from the font. Fixing.

include/PianoRoll.h
```diff
@@ -388,6 +390,7 @@ protected slots:
 	EditModes m_ctrlMode; // mode they were in before they hit ctrl
 
 	bool m_mouseDownRight; //true if right click is being held down
+	bool m_markFromMenu; // true if from right-click menu
```
Spekular, line 393: As far as I can tell this is just an indirect way to pass a parameter to `markSemitone`. I would strongly prefer to avoid this method. Couldn't you just add a `fromMenu` parameter to `markSemitone` that defaults to true? If leaving existing calls unmodified is your concern, this should solve the issue.

Veratil: I can add a default parameter. 👍

src/gui/editors/PianoRoll.cpp
```diff
@@ -4139,6 +4168,12 @@ void PianoRoll::noteLengthChanged()
 	update();
 }
 
+/*! \brief Handler when the scale dropdown changes
```
Veratil, line 4171:
```suggestion
/*! \brief Handler when the key dropdown changes
```

src/gui/editors/PianoRoll.cpp
```diff
@@ -374,6 +374,23 @@ PianoRoll::PianoRoll() :
...
+	m_keyModel.addItem(tr("D"));
+	m_keyModel.addItem(tr("D♯/E♭"));
+	m_keyModel.addItem(tr("E"));
+	m_keyModel.addItem(tr("F"));
+	m_keyModel.addItem(tr("F♯/G♭"));
+	m_keyModel.addItem(tr("G"));
+	m_keyModel.addItem(tr("G♯/A♭"));
+	m_keyModel.addItem(tr("A"));
+	m_keyModel.addItem(tr("A♯/B♭"));
+	m_keyModel.addItem(tr("B"));
```
Spekular, line 390: Perhaps this could be a collection `keys` and then a `for_each` could call `addItem` on them all? I'm fine with leaving it like this though.

Spekular (approved):
No more complaints here, LGTM!

src/gui/editors/PianoRoll.cpp
Veratil: Yeah I was thinking about this myself. When I wrote it I felt it was short enough and never going to change so this was an okay solution. Possible area for cleanup later for sure towards DRY.

Spekular:
Worked fine for me when testing. I find that the key combobox looks a bit cramped when "No key" is displayed. Also, highlights only apply when I move my mouse after clicking, so if I select a different key and don't move my mouse the previous highlights stick around. However, this occurs for me with other builds as well (when manually highlighting notes) so it isn't an issue with your PR.

Spekular:
Actually, I did find one issue that I can't reproduce on builds of my own branch. The vertical zoom combobox is misplaced (on top of the transport controls) and missing it's icon.

Veratil:
> I find that the key combobox looks a bit cramped when "No key" is displayed.

Ah you're right, I never noticed I guess since I know what it says. I'll see about expanding it a few pixels.

> Also, highlights only apply when I move my mouse after clicking, so if I select a different key and don't move my mouse the previous highlights stick around. However, this occurs for me with other builds as well (when manually highlighting notes) so it isn't an issue with your PR.

I see this happens with the right-click menu. It could easily be adding an `update()` call after processing. I'll look into it for another PR.

Veratil:
> 
> 
> Actually, I did find one issue that I can't reproduce on builds of my own branch. The vertical zoom combobox is misplaced (on top of the transport controls) and missing it's icon.

Ah that's because it was added after my PR. Fixing up now.

Veratil:
I added `update();` at the end of `markSemiTone` so it properly redraws now. Squashed all commits into 1 with rebase.

src/gui/editors/PianoRoll.cpp
Veratil, line 390: @Spekular so what if I used the same note strings as the piano keys?
```suggestion
	for (int i = 0; i < 12; ++i) { m_keyModel.addItem(tr(s_noteStrings[i].toStdString().data())); }
```

Spekular: Clever!

Spekular:
Tested again after the recent changes, all my concerns are addressed and the feature still works as expected :+1:

src/gui/editors/PianoRoll.cpp
```diff
@@ -374,6 +374,13 @@ PianoRoll::PianoRoll() :
 	connect( &m_noteLenModel, SIGNAL( dataChanged() ),
 					this, SLOT( noteLengthChanged() ) );
 
+	// Set up key selection dropdown
+	m_keyModel.addItem(tr("No key"));
+	// Use piano roll note strings for key dropdown
+	for (int i = 0; i < 12; i++) { m_keyModel.addItem(tr(s_noteStrings[i].toStdString().data())); }
```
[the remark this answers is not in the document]
Veratil: Ah, good point. I think that was when I learned about the German notes, so I figured English/German would have different notes.

src/gui/editors/PianoRoll.cpp
```diff
@@ -4139,6 +4160,12 @@ void PianoRoll::noteLengthChanged()
 	update();
 }
 
+/*! \brief Handler when the key dropdown changes
```
JohannesLorenz, line 4163: If you want, you can also use one line doxygen comments: `//! ...`

src/gui/editors/PianoRoll.cpp
```diff
@@ -485,11 +494,21 @@ void PianoRoll::changeNoteEditMode( int i )
 }
 
 
-void PianoRoll::markSemiTone( int i )
+/*! \brief highlight/clear piano key row or rows based on action i
```
JohannesLorenz, line 497: As Dom has already noted, these would be better in the header.

JohannesLorenz (commented):
Only doxygen comments reviewed.

There are still open questions about matter of taste:

* Maybe we should set `JAVADOC_AUTOBRIEF = YES` and `QT_AUTOBRIEF = YES` in `doc/Doxyfile.in`. This would simply make the first sentence (until the first dot) `\brief` (without use of the `\brief` command).
* Do we want to use `@param` vs `\param`, `//!` vs `///`, `/**` vs `/*!` , or should we just allow both?

src/gui/editors/PianoRoll.cpp
Veratil: Yeah, I agreed with it. :)

Veratil:
> 
> 
> Only doxygen comments reviewed.
> 
> There are still open questions about matter of taste:
> 
>     * Maybe we should set `JAVADOC_AUTOBRIEF = YES` and `QT_AUTOBRIEF = YES` in `doc/Doxyfile.in`. This would simply make the first sentence (until the first dot) `\brief` (without use of the `\brief` command).
> 
>     * Do we want to use `@param` vs `\param`, `//!` vs `///`, `/**` vs `/*!` , or should we just allow both?

I think this is a good discussion to move to an issue, rather than start it on this PR. I can remove the doc comments for now from this PR in favor of adding them in a future PR?

DomClark (approved):

PR #5197, opened 2019-09-17
Pattern import and export

CYBERDEViLNL:
This makes it possible to export and import patterns. They will be saved to a XML document. Currently with the file-extension `.xpf`, I think this extension needs to change, suggestions?

This PR is in-progress since I like to add compression and I feel to change the extension to something else.

Also feel free to ignore/close this if it is not wanted.


ibuki2003:
I'm pleased with this enhancement.

> file-extension .xpf, I think this extension needs to change, suggestions?

As you know extension .xpf is already used as 'Preset File'.
The extension doesn't have to contain 'f' for 'File', so I suggest .xpt.

> I like to add compression
pattern file should have both formats (without compression) in order to make editing with text editor easier

Finally, I report a problem
when loading failed, succesfully loaded message also shown.

Sorry for my bad English.
Thank you

CYBERDEViLNL:
Thank you for testing this feature.


> The extension doesn't have to contain 'f' for 'File', so I suggest .xpt.

What does the `t` in `.xpt` stand for?

> > pattern file should have both formats (without compression) in order to make editing with text editor easier

:+1: 

> when loading failed, succesfully loaded message also shown.

This should be fixed now :-)

ibuki2003:
Sorry for not explaining enough.
`pt` stands for **p**a**t**tern

CYBERDEViLNL:
Changed the file extentions to `.xpt` and `.xptz` and added compression.

If compression is used depends on if you add `.xptz` or `.xpt` to the filename, if no suffix is given then it depends on the `Compress project files by default` setting.

tecknixia:
Successfully exported and imported patterns over 1000 measures long using .xpt and .xptz, each using both settings... "compress project files by default"  selected and unselected.

Seems to work great.  Linux Mint 64.

Only one question... when "compress project files by default" is selected, I imagine typing .xptz compresses files, and .xpt does not.  When I unselect "compress project files by default", and save with xptz, does it still compress the files?

Sawuare:
@CYBERDEViLNL I guess the advantage of `xpt`/`xptz` over `midi` is having note pan data, right?

CYBERDEViLNL:
@Sawuare for me it's more for exporting a pattern from project A and importing it to project B.

Spekular:
@Cyberdevilnl functioning MIDI import/export would allow the same thing. I believe that's what Sawuare is pointing out, that the only additional data provided by pattern export/import is note pan.

CYBERDEViLNL:
@Sawuare @Spekular ah yes it would, it is possible with midi too but it's messy since each note would require it's own channel in the midi file.

Sawuare:
Anyway, I like this feature, but I think it should be presented to the user through the piano roll, not the song editor, because that is where the pattern editing happens.
If pattern MIDI export/import is ever implemented, it should also reside there.

CYBERDEViLNL:
All right, where should the import/export options/buttons be placed in the Piano-Roll? I can add it to the tools `QToolButton` if this gets merged: https://github.com/LMMS/lmms/pull/5216 if that's fine.

If so then I suggest to move the functionality code (`void exportPattern()`, `void importPattern()` and `int saveXML(QString filepath, bool compress = true)`) to `Pattern` instead of `PatternView` so we can access it through the Piano-Roll.

Sawuare:
How about a "File" `QToolButton`? I think "Tools" should be left for note operations like glue, quantize, reverse, flip, etc.

About moving the functionality code, I say that sounds good.

CYBERDEViLNL:
All right I'm on it. I have a few questions.

1. Where should the "File" `QToolButton` be added? (before, after or in `notesActionsToolBar`)
2. Should import only be enabled for patterns without notes? If not then should it merge or overwrite?

Sawuare:
1. Before.
2. No. Overwrite.

CYBERDEViLNL:
Ok I made the changes, I chose to move the functionality code to PianoRollWindow instead of what I previously suggested to move it to `Pattern`.

Current behaviour:
 * If no valid pattern is set, then the "File" `QToolButton` is disabled.
 * On import of a pattern it warns the user that notes will be overwritten if there are any present.

Will resolve the merge-conflicts asap.

Sawuare:
> On import of a pattern it warns the user that notes will be overwritten if there are any present.

The action should be undoable, so no need for a warning.

PhysSong:
@cyber-bridge Can you create a new pull request from your new account?

cyber-bridge:
@PhysSong: Sorry it took so long to respond, but will do.

PhysSong:
Continued in #5891.

PR #5199, opened 2019-09-19
Implement fade in to prevent Triple Osc from clicking

### LINKED ISSUE #5119: Triple Oscillator clicks if with phase shifting of 90 degrees
Select sine wave only, start at 90 "PO" (oscillator phase shift). Play a note (for lower note, you'll hear it best). Then 3Osc starts with a click.

Should this be fixed? If you have an attack time of > 0, there is no additional fade in algorithm required. But if your attack time is = 0, I think a fade-in algorithm must be applied.

For background info, zynaddsubfx uses [S-formed envelopes for fade-in](https://github.com/zynaddsubfx/zynaddsubfx/blob/4e36e765f677dbc689461889e502d015c22966a5/src/Synth/ADnote.cpp#L1202-L1222). The length depends on the note played, which results in a punchy, but not clicking sound for all frequencies.

necrashter:
Fixes #5119

Before:
image
After:
image

Algorithm is similar to [ZynAddSubFX's fade in function](https://github.com/zynaddsubfx/zynaddsubfx/blob/4e36e765f677dbc689461889e502d015c22966a5/src/Synth/ADnote.cpp#L1202-L1222) but not exactly the same.

Please review the code.

### DIFF
--- a/include/Instrument.h
+++ b/include/Instrument.h
@@ -132,8 +132,8 @@ public:
 
 
 protected:
-    // fade in to prevent clicks
-    void applyFadeIn( sampleFrame * buf, const NotePlayHandle * _n );
+	// fade in to prevent clicks
+	void applyFadeIn( sampleFrame * buf, const NotePlayHandle * _n );
 
 	// instruments may use this to apply a soft fade out at the end of
 	// notes - method does this only if really less or equal
--- a/include/Instrument.h
+++ b/include/Instrument.h
@@ -133,7 +133,7 @@ public:
 
 protected:
 	// fade in to prevent clicks
-	void applyFadeIn(sampleFrame * buf, const NotePlayHandle * _n);
+	void applyFadeIn(sampleFrame * buf, const NotePlayHandle * n);
 
 	// instruments may use this to apply a soft fade out at the end of
 	// notes - method does this only if really less or equal
--- a/include/Instrument.h
+++ b/include/Instrument.h
@@ -133,7 +133,7 @@ public:
 
 protected:
 	// fade in to prevent clicks
-	void applyFadeIn( sampleFrame * buf, const NotePlayHandle * _n );
+	void applyFadeIn(sampleFrame * buf, const NotePlayHandle * _n);
 
 	// instruments may use this to apply a soft fade out at the end of
 	// notes - method does this only if really less or equal
--- a/include/Instrument.h
+++ b/include/Instrument.h
@@ -133,7 +133,7 @@ public:
 
 protected:
 	// fade in to prevent clicks
-	void applyFadeIn(sampleFrame * buf, const NotePlayHandle * n);
+	void applyFadeIn(sampleFrame * buf, NotePlayHandle * n);
 
 	// instruments may use this to apply a soft fade out at the end of
 	// notes - method does this only if really less or equal
--- a/include/Instrument.h
+++ b/include/Instrument.h
@@ -132,6 +132,9 @@ public:
 
 
 protected:
+    // fade in to prevent clicks
+    void applyFadeIn( sampleFrame * buf, const NotePlayHandle * _n );
+
 	// instruments may use this to apply a soft fade out at the end of
 	// notes - method does this only if really less or equal
 	// desiredReleaseFrames() frames are left
--- a/src/core/Instrument.cpp
+++ b/src/core/Instrument.cpp
@@ -129,11 +129,12 @@ void Instrument::applyFadeIn(sampleFrame * buf, NotePlayHandle * n)
 	if (total == 0)
 	{
 		const fpp_t frames = n->framesLeftForCurrentPeriod();
+		const f_cnt_t offset = n->offset();
 
 		// We need to skip the first sample because it almost always
 		// produces a zero crossing; it's not helpful while
 		// determining the fade in length. Hence 1
-		int max_zc = count_zero_crossings(buf, 1, frames);
+		int max_zc = count_zero_crossings(buf, offset+1, offset+frames);
 
 		fpp_t length = getFadeInLength(MAX_FADE_IN_LENGTH,frames,max_zc);
 		n->m_fadeInLength = length;
@@ -144,7 +145,7 @@ void Instrument::applyFadeIn(sampleFrame * buf, NotePlayHandle * n)
 		{
 			for (ch_cnt_t ch = 0; ch < DEFAULT_CHANNELS; ++ch)
 			{
-				buf[f][ch] *= 0.5 - 0.5 * cosf(F_PI * (float) f / (float)n->m_fadeInLength);
+				buf[offset+f][ch] *= 0.5 - 0.5 * cosf(F_PI * (float) f / (float)n->m_fadeInLength);
 			}
 		}
 	}
--- a/src/core/Instrument.cpp
+++ b/src/core/Instrument.cpp
@@ -82,51 +82,79 @@ bool Instrument::isFromTrack( const Track * _track ) const
 	return( m_instrumentTrack == _track );
 }
 
-
-void Instrument::applyFadeIn(sampleFrame * buf, const NotePlayHandle * n)
+int count_zero_crossings(sampleFrame *buf,fpp_t start, fpp_t frames)
 {
-	// apply only if it's the start of the note
-	if (n->totalFramesPlayed() == 0)
+	// zero point crossing counts of all channels
+	int zero_crossings[DEFAULT_CHANNELS] = {0};
+	// maximum zero point crossing of all channels
+	int max_zc = 0;
+
+	// determine the zero point crossing counts
+	for (fpp_t f = start; f < frames; ++f)
 	{
-		const fpp_t frames = n->framesLeftForCurrentPeriod();
-
-		// zero point crossing counts of all channels
-		int zero_crossings[DEFAULT_CHANNELS] = {0};
-		// maximum zero point crossing of all channels
-		int max_zc = 0;
-
-		// determine the zero point crossing counts
-		for (fpp_t f = 0; f < frames; ++f)
+		for (ch_cnt_t ch=0; ch < DEFAULT_CHANNELS; ++ch)
 		{
-			for (ch_cnt_t ch=0; ch < DEFAULT_CHANNELS; ++ch)
+			// we don't want to count [-1, 0, 1] as two crossings
+			if ((buf[f-1][ch] <= 0.0 && buf[f][ch] > 0.0) ||
+					(buf[f-1][ch] >= 0.0 && buf[f][ch] < 0.0))
 			{
-				// we don't want to count [-1, 0, 1] as two crossings
-				if ((buf[f-1][ch] <= 0.0 && buf[f][ch] > 0.0) ||
-						(buf[f-1][ch] >= 0.0 && buf[f][ch] < 0.0))
+				++zero_crossings[ch];
+				if (zero_crossings[ch] > max_zc)
 				{
-					++zero_crossings[ch];
-					if (zero_crossings[ch] > max_zc)
-					{
-						max_zc = zero_crossings[ch];
-					}
+					max_zc = zero_crossings[ch];
 				}
 			}
 		}
+	}
+
+	return max_zc;
+}
+
+
+void Instrument::applyFadeIn(sampleFrame * buf, NotePlayHandle * n)
+{
+	f_cnt_t total = n->totalFramesPlayed();
+	if (total == 0)
+	{
+		const fpp_t frames = n->framesLeftForCurrentPeriod();
+
+		int max_zc = count_zero_crossings(buf, 1, frames);
 
 		// calculate the length of the fade in
 		// Length is inversely proportional to the max of zero_crossings,
 		// because for low frequencies, we need a longer fade in to
 		// prevent clicking.
-		fpp_t length = (fpp_t) (
-				((float)frames - 1)  /
-				((float)max_zc / 2.0f + 1.0f) / 3.0f);
+		fpp_t length = (fpp_t) (85.0  / ((float)max_zc / ((float)frames/128.0f) + 1.0f));
+		n->m_fadeInLength = length;
 
 		// apply fade in
+		length = length < frames ? length : frames;
 		for (fpp_t f = 0; f < length; ++f)
 		{
 			for (ch_cnt_t ch = 0; ch < DEFAULT_CHANNELS; ++ch)
 			{
-				buf[f][ch] *= 0.5 - 0.5 * cosf(F_PI * (float) f / (float) length);
+				buf[f][ch] *= 0.5 - 0.5 * cosf(F_PI * (float) f / (float)n->m_fadeInLength);
+			}
+		}
+	}
+	else if(total < n->m_fadeInLength)
+	{
+		const fpp_t frames = n->framesLeftForCurrentPeriod();
+
+		int new_zc = count_zero_crossings(buf, 0, frames);
+
+		if(new_zc != 0)
+		{
+			n->m_fadeInLength = (fpp_t) ((float)n->m_fadeInLength * 0.875f);
+		}
+
+		fpp_t length = n->m_fadeInLength - total;
+
+		for (fpp_t f = 0; f < length; ++f)
+		{
+			for (ch_cnt_t ch = 0; ch < DEFAULT_CHANNELS; ++ch)
+			{
+				buf[f][ch] *= 0.5 - 0.5 * cosf(F_PI * (float) (total+f) / (float)n->m_fadeInLength);
 			}
 		}
 	}
@@ -160,6 +188,3 @@ QString Instrument::fullDisplayName() const
 {
 	return instrumentTrack()->displayName();
 }
-
-
-
--- a/src/core/Instrument.cpp
+++ b/src/core/Instrument.cpp
@@ -83,7 +83,7 @@ bool Instrument::isFromTrack( const Track * _track ) const
 }
 
 
-void Instrument::applyFadeIn(  sampleFrame * buf, const NotePlayHandle * n )
+void Instrument::applyFadeIn(sampleFrame * buf, const NotePlayHandle * n)
 {
 	// apply only if it's the start of the note
 	if (n->totalFramesPlayed() == 0)
@@ -96,15 +96,15 @@ void Instrument::applyFadeIn(  sampleFrame * buf, const NotePlayHandle * n )
 		int max_zc = 0;
 
 		// determine the zero point crossing counts
-		for ( fpp_t f = 0; f < frames; ++f )
+		for (fpp_t f = 0; f < frames; ++f)
 		{
-			for ( ch_cnt_t ch=0; ch < DEFAULT_CHANNELS; ++ch )
+			for (ch_cnt_t ch=0; ch < DEFAULT_CHANNELS; ++ch)
 			{
-				if ( ( buf[f-1][ch] < 0.0 && buf[f][ch] > 0.0 ) ||
-						( buf[f-1][ch] > 0.0 && buf[f][ch] < 0.0 ) )
+				if ((buf[f-1][ch] < 0.0 && buf[f][ch] > 0.0) ||
+						(buf[f-1][ch] > 0.0 && buf[f][ch] < 0.0))
 				{
 					++zero_crossings[ch];
-					if ( zero_crossings[ch] > max_zc )
+					if (zero_crossings[ch] > max_zc)
 					{
 						max_zc = zero_crossings[ch];
 					}
@@ -114,13 +114,13 @@ void Instrument::applyFadeIn(  sampleFrame * buf, const NotePlayHandle * n )
 
 		// calculate the length of the fade in
 		fpp_t length = (fpp_t) (
-				( (float)frames - 1 )  /
-				( (float)max_zc / 2.0f + 1.0f ) / 3.0f );
+				((float)frames - 1)  /
+				((float)max_zc / 2.0f + 1.0f) / 3.0f);
 
 		// apply fade in
-		for ( fpp_t f = 0; f < length; ++f )
+		for (fpp_t f = 0; f < length; ++f)
 		{
-			for ( ch_cnt_t ch = 0; ch < DEFAULT_CHANNELS; ++ch )
+			for (ch_cnt_t ch = 0; ch < DEFAULT_CHANNELS; ++ch)
 			{
 				buf[f][ch] *= 0.5 - 0.5 * cosf(F_PI * (float) f / (float) length);
 			}
--- a/src/core/Instrument.cpp
+++ b/src/core/Instrument.cpp
@@ -82,6 +82,7 @@ bool Instrument::isFromTrack( const Track * _track ) const
 	return( m_instrumentTrack == _track );
 }
 
+// helper function for Instrument::applyFadeIn
 int count_zero_crossings(sampleFrame *buf,fpp_t start, fpp_t frames)
 {
 	// zero point crossing counts of all channels
@@ -110,9 +111,20 @@ int count_zero_crossings(sampleFrame *buf,fpp_t start, fpp_t frames)
 	return max_zc;
 }
 
+// helper function for Instrument::applyFadeIn
+fpp_t getFadeInLength(float max_length, fpp_t frames, int zero_crossings)
+{
+	// calculate the length of the fade in
+	// Length is inversely proportional to the max of zero_crossings,
+	// because for low frequencies, we need a longer fade in to
+	// prevent clicking.
+	return (fpp_t) (max_length  / ((float)zero_crossings / ((float)frames/128.0f) + 1.0f));
+}
+
 
 void Instrument::applyFadeIn(sampleFrame * buf, NotePlayHandle * n)
 {
+	const static float MAX_FADE_IN_LENGTH = 85.0;
 	f_cnt_t total = n->totalFramesPlayed();
 	if (total == 0)
 	{
@@ -123,11 +135,7 @@ void Instrument::applyFadeIn(sampleFrame * buf, NotePlayHandle * n)
 		// determining the fade in length. Hence 1
 		int max_zc = count_zero_crossings(buf, 1, frames);
 
-		// calculate the length of the fade in
-		// Length is inversely proportional to the max of zero_crossings,
-		// because for low frequencies, we need a longer fade in to
-		// prevent clicking.
-		fpp_t length = (fpp_t) (85.0  / ((float)max_zc / ((float)frames/128.0f) + 1.0f));
+		fpp_t length = getFadeInLength(MAX_FADE_IN_LENGTH,frames,max_zc);
 		n->m_fadeInLength = length;
 
 		// apply fade in
@@ -144,22 +152,23 @@ void Instrument::applyFadeIn(sampleFrame * buf, NotePlayHandle * n)
 	{
 		const fpp_t frames = n->framesLeftForCurrentPeriod();
 
-		int new_zc = count_zero_crossings(buf, 0, frames);
+		int new_zc = count_zero_crossings(buf, 1, frames);
+		fpp_t new_length = getFadeInLength(MAX_FADE_IN_LENGTH, frames, new_zc);
 
-		if(new_zc != 0)
-		{
-			n->m_fadeInLength = (fpp_t) ((float)n->m_fadeInLength * 0.875f);
-		}
-
-		fpp_t length = n->m_fadeInLength - total;
-
-		for (fpp_t f = 0; f < length; ++f)
+		for (fpp_t f = 0; f < frames; ++f)
 		{
 			for (ch_cnt_t ch = 0; ch < DEFAULT_CHANNELS; ++ch)
 			{
-				buf[f][ch] *= 0.5 - 0.5 * cosf(F_PI * (float) (total+f) / (float)n->m_fadeInLength);
+				float currentLength = n->m_fadeInLength*(1.0f-(float)f/frames) + new_length*((float)f/frames);
+				buf[f][ch] *= 0.5 - 0.5 * cosf(F_PI * (float) (total+f) / currentLength);
+				if(total+f >= currentLength)
+				{
+					n->m_fadeInLength = currentLength;
+					return;
+				}
 			}
 		}
+		n->m_fadeInLength = new_length;
 	}
 }
 
--- a/src/core/Instrument.cpp
+++ b/src/core/Instrument.cpp
@@ -100,8 +100,9 @@ void Instrument::applyFadeIn(sampleFrame * buf, const NotePlayHandle * n)
 		{
 			for (ch_cnt_t ch=0; ch < DEFAULT_CHANNELS; ++ch)
 			{
-				if ((buf[f-1][ch] < 0.0 && buf[f][ch] > 0.0) ||
-						(buf[f-1][ch] > 0.0 && buf[f][ch] < 0.0))
+                // we don't want to count [-1, 0, 1] as two crossings
+				if ((buf[f-1][ch] <= 0.0 && buf[f][ch] > 0.0) ||
+						(buf[f-1][ch] >= 0.0 && buf[f][ch] < 0.0))
 				{
 					++zero_crossings[ch];
 					if (zero_crossings[ch] > max_zc)
@@ -113,6 +114,9 @@ void Instrument::applyFadeIn(sampleFrame * buf, const NotePlayHandle * n)
 		}
 
 		// calculate the length of the fade in
+		// Length is inversely proportional to the max of zero_crossings,
+		// because for low frequencies, we need a longer fade in to
+		// prevent clicking.
 		fpp_t length = (fpp_t) (
 				((float)frames - 1)  /
 				((float)max_zc / 2.0f + 1.0f) / 3.0f);
--- a/src/core/Instrument.cpp
+++ b/src/core/Instrument.cpp
@@ -22,11 +22,13 @@
  *
  */
 
+#include "Instrument.h"
+
 #include <cmath>
 
-#include "Instrument.h"
-#include "InstrumentTrack.h"
 #include "DummyInstrument.h"
+#include "InstrumentTrack.h"
+#include "lmms_constants.h"
 
 
 Instrument::Instrument(InstrumentTrack * _instrument_track,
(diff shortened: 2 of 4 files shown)


src/core/Instrument.cpp
```diff
@@ -79,7 +81,52 @@ bool Instrument::isFromTrack( const Track * _track ) const
...
+        }
+
+        // find the biggest one
+        for( ch_cnt_t ch = 0; ch < DEFAULT_CHANNELS; ++ch )
+        {
+            if( zero_crossings[ch] > max_zc )
+            {
+                max_zc = zero_crossings[ch];
+            }
+        }
```
Veratil: This loop is pulled out of the previous, which to me says this `if` check can be combined into the previous for loop to prevent looping through things multiple times.

src/core/Instrument.cpp
```diff
@@ -79,7 +81,52 @@ bool Instrument::isFromTrack( const Track * _track ) const
...
+        
+        // calculate the length of the fade in
+        fpp_t length = (fpp_t) (
+                ( (float)frames - 1 )  /
+                ( (float)max_zc / 2.0f + 1.0f ) / 3.0f );
+
+        for( fpp_t f = 0; f < length; ++f ){
+            for( ch_cnt_t ch = 0; ch < DEFAULT_CHANNELS; ++ch ){
+                // define macro for PI?
+                buf[f][ch] *= 0.5 - 0.5 * cosf(3.1415926536f * (float) f / (float) length);
```
Veratil: Looked this up, looks like in `<QtMath>` there is an `M_PI` that can be used. Unsure if this is preferred or not, will defer to main devs.

Spekular:
I don't know that it's accurate to call the issue clipping. Nitpicking as
an innaccurate term in for example the commit message could be misleading
to someone on the future.

src/core/Instrument.cpp
PhysSong: `lmms_constants.h` also contains the value of the pi with various precisions(`F_PI` for `float`, `D_PI` for `double`, and `LD_PI` for `long double`). You can use those macros by including `lmms_math.h` as well.

PhysSong:
> I don't know that it's accurate to call the issue clipping.

IIRC it's called clicking.

PhysSong:
@necrashter Please use a tab instead of 4 spaces for indentation. According to our coding conventions, you should use tabs instead of spaces.

src/core/Instrument.cpp
necrashter: I considered that but I was unsure whether it will be more efficient or not. Because if we combine these loops, it will check if it's the new maximum for every zero point crossing. Performance difference is probably marginal, but perhaps it can make a meaningful difference if there are too many instruments.

I will merge these two loops anyway in the next commit to improve readability. If someone proves that the current method is faster, I can revert these changes.

src/core/Instrument.cpp
necrashter: Thanks, I searched `lmms_math.h` for pi but didn't see `lmms_constants.h`

necrashter:
Thank you all for your feedback.

Sorry for the misleading title, I said "clipping" but meant "clicking".

Is it okay to put spaces after opening parenthesis and before closing parenthesis? In `Instrument.cpp` there are spaces inside parenthesis but the coding conventions page doesn't state anything about that.

Also this PR is for master branch. Should I have made a PR for stable-1.2 branch instead?

JohannesLorenz:
> Also this PR is for master branch. Should I have made a PR for stable-1.2 branch instead?

I think so. It's a bugfix and it's very unlikely to cause any regression.

PhysSong:
> It's a bugfix and it's very unlikely to cause any regression.

I think it's a bug-fix, too. Depending on the POV, however, some devs may think this is an enhancement rather than a bug-fix. Also, this will introduce some behavioral changes in the stable version. For someone who (ab-)used this behavior, this change might be unwanted.

Anyway, please forget about `stable-1.2` before we complete this.

PhysSong:
> Is it okay to put spaces after opening parenthesis and before closing parenthesis? In `Instrument.cpp` there are spaces inside parenthesis but the coding conventions page doesn't state anything about that.

That's an old convention. According to the new coding convention, you shouldn't put those whitespaces in the new code.

necrashter:
> Depending on the POV, however, some devs may think this is an enhancement rather than a bug-fix. Also, this will introduce some behavioral changes in the stable version. For someone who (ab-)used this behavior, this change might be unwanted.

Good point. As far as I know, starting with a click is not the expected behavior of a synthesizer but some people may have used it.

> That's an old convention. According to the new coding convention, you shouldn't put those whitespaces in the new code.

I'm on it, but I won't change the old code for the sake of this PR's integrity.

JohannesLorenz:
> Good point. As far as I know, starting with a click is not the expected behavior of a synthesizer but some people may have used it.

If we care about backward compatibility, we could also use a boolean flag in the XML that enables the new fade in. The flag would be set for savefiles with this project version and above, but will be absent in earlier versions. Don't know if that makes sense, though.

necrashter:
> If we care about backward compatibility, we could also use a boolean flag in the XML that enables the new fade in. The flag would be set for savefiles with this project version and above, but will be absent in earlier versions. Don't know if that makes sense, though.

What if they want to enable this bugfix/feature for an old project? Not all users can understand XML. Perhaps we should put a on/off button somewhere. Any other ideas?

Veratil:
> 
> 
> > If we care about backward compatibility, we could also use a boolean flag in the XML that enables the new fade in. The flag would be set for savefiles with this project version and above, but will be absent in earlier versions. Don't know if that makes sense, though.
> 
> What if they want to enable this bugfix/feature for an old project? Not all users can understand XML. Perhaps we should put a on/off button somewhere. Any other ideas?

Instead of a boolean in the config to enable this, this should be an option on the plugin itself.

include/Instrument.h
```diff
@@ -133,7 +133,7 @@ class LMMS_EXPORT Instrument : public Plugin
 
 protected:
 	// fade in to prevent clicks
-	void applyFadeIn( sampleFrame * buf, const NotePlayHandle * _n );
+	void applyFadeIn(sampleFrame * buf, const NotePlayHandle * _n);
```
Veratil: You could remove the variable names here entirely, but at least remove the underscore in `_n`.

src/core/Instrument.cpp
```diff
@@ -79,7 +83,50 @@ bool Instrument::isFromTrack( const Track * _track ) const
...
+		int zero_crossings[DEFAULT_CHANNELS] = {0};
+		// maximum zero point crossing of all channels
+		int max_zc = 0;
 
+		// determine the zero point crossing counts
+		for ( fpp_t f = 0; f < frames; ++f )
+		{
+			for ( ch_cnt_t ch=0; ch < DEFAULT_CHANNELS; ++ch )
+			{
+				if ( ( buf[f-1][ch] < 0.0 && buf[f][ch] > 0.0 ) ||
```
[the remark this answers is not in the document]
softrabbit: Not entirely sure that formula is 100% accurate. `abs(1) > abs(1 + -3)` is false AFAIK, so you might miss the occasional crossing this way too?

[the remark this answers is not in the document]
Veratil: Would adding an extra `|| buf[f][ch] == 0.0` make it catch exact zeros?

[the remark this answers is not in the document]
Spekular: Yeah you're right, somehow I assumed the first sample could be the one with the biggest absolute value. You could probaly change it to `max(abs(a), abs(b))` to fix it. There's also `abs(a) + abs(b) > abs(a+b)`.

- `max(abs(a), abs(b)) >= abs(a+b) && a != b` should catch everything in 6 comparisons.
- `abs(a) + abs(b) > abs(a+b) || (a != b && (a = 0 || b = 0))` uses the same number but more AND/ORs. It catches non-exact-zero crossings more elegantly IMO, but is a pain to catch the exact ones.
- `((a >= 0 && b <= 0) || (a <= 0 && b>= 0)) && a != b` is five comparisons, if we modify the original solution.

All of these would catch two separate crossings if the sample value goes { -1, 0, 0, 1 }.

How/why is the zero crossing detection actually used in this PR? I haven't figured it out from my brief review.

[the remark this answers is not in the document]
necrashter: The intuition behind counting zero point crossings is adapting to the frequency of the sound. We need a longer fade in for lower frequencies because the clicking is more audible for lower notes. 

You are absolutely right, current algorithm misses the crossing in [1, 0, -1]. However, the problem with catching [1, 0] and [0, -1] as zero point crossings is that [1, 0, -1] will be count as two zero point crossings. We don't want that because if you plot the amplitude vs time graph of [1, 0, -1], you can see that it crosses the time axis only once. Furthermore, [1, -1, 1] clearly has more frequency than [1, 0, -1], hence these should not have the same zero point crossing counts.

Although using abs function looks shorter, it might be slower (depending on the implementation) if it uses conditional branching. The same goes for max function. Also I don't think that it contributes to the readability of the code.

`(a >= 0 && b < 0) || (a <= 0 && b > 0)` might be a good solution. [1, 0, -1] will be count as one zero point crossing since [1, 0] returns false and [0, -1] returns true.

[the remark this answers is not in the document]
Spekular: >We don't want [duplicate crossings]

Yeah, that's why I mentioned it.

>using abs function [...]

Aside from the first (faulty) check, I wasn't suggesting any of these. I was just presenting the "fully developed" abs solutions for comparison purposes.

`(a >= 0 && b < 0) || (a <= 0 && b > 0)` looks great.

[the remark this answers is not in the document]
softrabbit: Why not ask the NotePlayHandle for its frequency and calculate a suitable fade in from that?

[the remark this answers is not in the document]
necrashter: Because the frequency may change with coarse detuning, frequency modulation etc.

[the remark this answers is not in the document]
JohannesLorenz: > `(a >= 0 && b < 0) || (a <= 0 && b > 0)` might be a good solution. [1, 0, -1] will be count as one zero point crossing since [1, 0] returns false and [0, -1] returns true.

Yes, though `[1, -0.000000001, 1]` or `[-1, 0.000000001, -1]` will count as 2.

I'm not sure why we deviate from zyn's original algorithm which counts only positive crossings. Improvements are always cool, but zyn's algorithm sounds absolutely perfect for >10 years and I wonder if we'll get better than absolutely perfect :grin: .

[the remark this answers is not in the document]
JohannesLorenz: Though zyn's algorithm only works until a buffersize of 256. At 32, it's useless. This algorithm will have the same difficulties. There are different approaches, like using the note height (@softrabbit mentioned it, but as discussed, there is modulation and stuff, and there are also harmonics with a different frequency), or, in case of 32 samples, using this algorithm iteratively to get step-wise to the full volume (i.e. multiple runs).

Maybe, instead of trying, we should ask for a good algorithm on [dsp.stackexchange.com](https://dsp.stackexchange.com) , [linux-audio-developper mailing list](https://lists.linuxaudio.org/listinfo/linux-audio-dev/) or [The KVR audio forum](https://www.kvraudio.com/forum/).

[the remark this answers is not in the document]
necrashter: > [1, -0.000000001, 1] or [-1, 0.000000001, -1] will count as 2.

In that case we can add some margin of error:
`(a >= 0 && b < -0.000000001) || (a <= 0 && b > 0.000000001)`
Perhaps this margin of error should be `std::numeric_limits<float>::epsilon()` or a constant.

> I'm not sure why we deviate from zyn's original algorithm which counts only positive crossings. 

Frankly, I don't understand why it only counts positive crossings (for a little bit of speed, maybe?). So I decided to count both.

> At 32, it's useless. This algorithm will have the same difficulties.

As a solution, I thought about calculating the length in the first iteration and the saving it somewhere. In the next iterations we can just use that length to calculate the proper amplitude. However, there are problems with that. At a buffersize of 32, we can only see a tiny fraction of the wave, so the length calculation may not be accurate. Also, I have no idea where the length should be saved.

Another way to solve this problem might be to calculate the steepness of the wave. If function starts at 0 and gradually increases to 1 like a sine wave, we don't need any fade in. If function starts at 1 and gradually decreases to 0, similarly to the example in the screenshot, we will need to smooth the beginning of the wave, because it actually jumps from 0 to 1 in the beginning. 

I am not an expert on this topic, probably there are many people who are more qualified than me on the sites you have mentioned.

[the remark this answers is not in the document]
JohannesLorenz: I just posted in the [linux audio developer list](http://linux-audio.4202.n7.nabble.com/Click-free-fade-in-algorithm-for-synths-td108758.html) ("LAD"). Let's see if this will help us.

Considering your solutions, the 1st sounds doable. The first 32 samples don't reveal much useful information, but you can still use it and just increase e.g. up to ~ 1/8. If you were much wrong, maybe one can not hear it, because it's just 1/8. And at 2/8, you get more exact, still with a bit error, but it's still not crucial at 2/8. And so on. The more crucial it gets, the more information you have. Anyways, I'd prefer waiting for LAD.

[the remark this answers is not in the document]
JohannesLorenz: The only think applicable I found in the mailing list were the links from Ardour:

* https://github.com/Ardour/ardour/blob/master/libs/ardour/ardour/disk_reader.h#L135
* https://github.com/Ardour/ardour/blob/master/libs/ardour/disk_reader.cc#L1445
* https://github.com/Ardour/ardour/blob/master/libs/ardour/amp.cc#L163

Has anyone time to read them and check what that algorithm actually does?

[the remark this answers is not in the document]
necrashter: >Has anyone time to read them and check what that algorithm actually does?

`DeclickAmp` has a _g and _a member values. _a is dependent on the sample rate, and it affects how fast _g increases. Every sample is multiplied with _g until it reaches the `target`, which is a parameter of `apply_gain`. When _g reaches the target, it uses the `Amp`'s `apply_simple_gain` method.

 `apply_gain` method of `DeclickAmp` iterates through the buffer in groups of maximum 16. `g`, which is the local version of `_g`, gets updated when the iteration of a group ends. So g is not updated after each sample, it only gets updates once every 16 samples. g is updated using `g += a * (target - g);` where `a==_a`. Thus the increase rate of g gets smaller and smaller. When the difference between `g` and `target` becomes smaller than 1e-5, `g` is set to `target`, ending the fade in phase.

So this algorithm doesn't do anything with the shape/frequency/zero point crossings of the wave. It only depends on the sample rate. And it decays exponentialy, which makes sense because the clicking happens at the beginning.

Some people on the mailing list have objected to our approach, saying that a synth should be controllable, it shouldn't try to outsmart its user. I respect the developers of ZynAddSubFX, but their approach might be obsolete/unwanted. Perhaps all we need is a controllable fade in whose length depends on the frequency of the note, so that the low notes will have longer fade in times. If the frequency of the notes can change with modulation, detuning etc., the user can just adjust the parameters of the fade in accordingly.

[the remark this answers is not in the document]
JohannesLorenz: Then let's let the algorithms compete. You already have one for the zero crossings, one dependent on the sample rate ("fixed length") should be easy to implement.

You could do it like this: Add a counter to fade in, which increases by one for every fade in. If it's even, use zero crossings, if it's odd, use "fixed" length (we could even try some more, I'd be interested to experiment there as well). This will make it easy to experiment and find out which fades are best (without recompiling).

I'm sorry that this sounds like more work then you could have expected from the PR. But these are important experiments that may be reused for many other LMMS synths and even zyn, so this is very valuable research.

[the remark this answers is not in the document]
necrashter: I'm on it, I will provide the visualizations and output files for each algorithm when it's done. This review has turned into a long conversation, so I will resolve it here and post the results as a new comment.

necrashter:
After a long conversation in the code review above, we decided to try another fade in algorithm and compare them. Current algorithm counts the zero point crossings of the wave and calculates the length according to that. Other algorithm is fixed length and based on sample rate. Detailed explanations can be found in the resolved conversation above.

Here is E2 note played by a sine wave with a phase offset of 90 degrees: (I'm tempted to call this cosine wave :D )
image
First one is the zero-crossings algorithm, second one is fixed-length and last one doesn't have any fade in for reference.

You can clearly see that second algorithm only updates the multiplier every 16 samples, it progresses step by step. It looks bad when visualized (due to zoom) but you don't notice it while listening; I tried updating every step and couldn't hear any difference.

Another important difference is the length of the fade in. Even at a low frequency, zero crossings algorithm produces a shorter fade in, resulting in a punchier sound. I used a buffer size of 256 samples and 44100 Hz sample rate. The longest possible fade in this algorithm can generate with these settings is about 0.0019 seconds, which is approximately equal to the length of the fixed-length algorithm. But even for low notes like E2, fade in length is half of that.

The most important drawback of the fixed length algorithm is that it smooths the high notes too much, making them lose the punch. Here's the C5 note played by the same instrument as above:
image
This time top one is fixed-length algorithm and bottom one is zero-crossings algorithm. Zero-crossings algorithm completes its fade in within the first oscillation. This is not the case for fixed-length algorithm. You can notice the difference when you listen to it.

[You can access these sound files here.](https://drive.google.com/open?id=1kIzjPRBZjXp_TiVIbLCIR0AKnsjZr2GL)

We can modify the current algorithm to consider the sample rate while calculating the fade in length. This way, we can get consistent fade in times regardless of sample rate. But the buffer size is a problem. For example, at 256 samples, the algorithm detected 2 crossings in the first sound and 7 in the second. It sees the start of the note as a zero point crossing in both cases because the gain jumps from 0 to 1. If buffer size were 32 samples, the algorithm would see 1 crossing and 2 crossings respectively. So if a sound start with a zero crossing, it's not a useful information.

I think we can improve the current algorithm to work with smaller buffer sizes. Fixed-length algorithm doesn't seem to be designed for this purpose. I will try that tomorrow but I need to ask these first:
- How can I get the sample rate? (I hard-coded it to quickly develop for this test.)
- Where to save the length of the fade in for small buffer sizes? `NotePlayHandle`?

lleroy:
has anyone looked at my proposed code in 
https://github.com/LMMS/lmms/issues/3086#issuecomment-522903500
to see if that is viable as a fix for this issue?

necrashter:
> has anyone looked at my proposed code in
> [#3086 (comment)](https://github.com/LMMS/lmms/issues/3086#issuecomment-522903500)
> to see if that is viable as a fix for this issue?

Algorithm is similar but I think it's not wise to implement fade-in effect in the release function. Also your implementation will run into the aforementioned issues while running on small buffer size. But that partial polynomial function looks nice, I used the cosine function in this implementation.

Our main problem is that we want a smooth fade in on low frequencies and a steeper, punchier one on higher frequencies. Another thing to consider is the reproducibility of the results with different bit-rates, buffer sizes etc. In order to fix these issues, we need to vary the length of the fade-in depending on these variables. 

This implementation calculates and saves the fade-in length in NotePlayHandle as fade-in may take longer than one period. But the current implementation doesn't consider the bit-rate while calculating the length. I'm not sure if we should make the fade-in shorter (sample-wise) with higher bit-rates to maintain the same length (time-wise) across different bit-rates.

softrabbit:
> This implementation calculates and saves the fade-in length in NotePlayHandle as fade-in may take longer than one period. But the current implementation doesn't consider the bit-rate while calculating the length. I'm not sure if we should make the fade-in shorter (sample-wise) with higher bit-rates to maintain the same length (time-wise) across different bit-rates.

Uh... in higher bitrates you'd need _more_ samples to maintain the length, right? 

Maybe a low pass filter would be the most elegant solution, both here and for other control values that change too fast.

necrashter:
> Uh... in higher bitrates you'd need _more_ samples to maintain the length, right?

Yeah, I made a fool of myself. Just fixed that now.
 
> Maybe a low pass filter would be the most elegant solution, both here and for other control values that change too fast.

I have been experimenting with this using the ENV/LFO, and it seems to be very similar to S-shaped envelopes. I don't know which one is more elegant though.

src/core/Instrument.cpp
```diff
@@ -78,8 +82,83 @@ bool Instrument::isFromTrack( const Track * _track ) const
...
+
+
+void Instrument::applyFadeIn(sampleFrame * buf, NotePlayHandle * n)
+{
+	f_cnt_t total = n->totalFramesPlayed();
+	if (total == 0)
+	{
+		const fpp_t frames = n->framesLeftForCurrentPeriod();
+
+		int max_zc = count_zero_crossings(buf, 1, frames);
```
JohannesLorenz: Following the PR's comments, I think the `1` is used in order to skip the first sample, which always produced a zero-crossing? If that's the case, I'd add a comment explaining why you used `1`. For us, it's clear, but a future reader might be confused.

src/core/Instrument.cpp
```diff
@@ -78,8 +82,83 @@ bool Instrument::isFromTrack( const Track * _track ) const
...
+
+		// calculate the length of the fade in
+		// Length is inversely proportional to the max of zero_crossings,
+		// because for low frequencies, we need a longer fade in to
+		// prevent clicking.
+		fpp_t length = (fpp_t) (85.0  / ((float)max_zc / ((float)frames/128.0f) + 1.0f));
```
JohannesLorenz: I can't follow why you changed the formula. Where are the constants `85` and `128` from? I imagine like this:

* `85.0` is the maximum fade in that you ever want to allow (IIRC we never had more in zyn)
* `128` is used because at 256, we get a perfect fade with zyn's formula, so this was the old `2.0`, and now you're scaling it down to respect the number of frames

Are my assumptions correct?

src/core/Instrument.cpp
```diff
@@ -78,8 +82,83 @@ bool Instrument::isFromTrack( const Track * _track ) const
...
+	}
+	else if(total < n->m_fadeInLength)
+	{
+		const fpp_t frames = n->framesLeftForCurrentPeriod();
+
+		int new_zc = count_zero_crossings(buf, 0, frames);
+
+		if(new_zc != 0)
+		{
+			n->m_fadeInLength = (fpp_t) ((float)n->m_fadeInLength * 0.875f);
```
JohannesLorenz: So, I guess `7/8` is just a guess for what sounded good? You're just saying: If the wave crosses 0 again, we can increase speed a bit?

It might work in this direction, but imagine the opposite: The note starts with tons of zero crossings, so the algorithm thinks "I can make a quick fade in". In the second buffer it then sees "not many/no zero crossings", so it actually increased too fast. In this case, shouldn't you slow down?

Though I just think there are cases where you have no chance. Imagine you have 32 samples buffersize, and the algorithm calculates a fade of 20 for the first 32 samples. It's finished then. In the next 32 sample, the number of zero crossings is way lower. Did the algorithm increase too fast?

Maybe you could rather do it like this: Calculate the expected fade in from 0 to the last sample (i.e. `total + length`) in each call of `applyFadeIn`. Calculate the increase speed for this buffer depending on that. Interpolate the new speed respective to the old to not have a steep edge. But I'm not sure at all if this is better.

src/core/Instrument.cpp
necrashter: Your assumptions are correct.

The Zyn's formula `((float)frames - 1)  / ((float)max_zc / 2.0f + 1.0f) / 3.0f` calculates the fade-in length pretty well at 256 samples. If we plug `frames=256` into the formula, it becomes `255.0f  / ((float)max_zc / 2.0f + 1.0f) / 3.0f` which simplifies to `85.0f  / ((float)max_zc / 2.0f + 1.0f)`.

This assumes a buffer size of 256 samples, so we need to adapt it to different buffer sizes. Instead of using the number of zero point crossings, we should use the number of zero point crossings per frame. If we divide `max_zc ` by 2 at 256 samples, we need to divide it by 4 at 512 samples, by 1 at 128 samples and so on. This is where 128 comes from.

I think 128 is somewhat understandable, but 85 looks magical and appears out of nowhere. We can define a constant for the max fade in length and replace 85 with that.

src/core/Instrument.cpp
necrashter: > ... In this case, shouldn't you slow down?

Since the purpose of this algorithm is to prevent clicking while producing punchy notes, I think it shouldn't. If a signal starts with lots of zero crossings followed by very few zero crossings, it means that it has a harsh beginning. We don't want to over-smooth the wave, that's the purpose of the regular envelope. We just want to prevent clicks in this context.

At 32 samples, even if the algorithms sees only one zero point crossing, it generates a fade in of length 17 samples. At 64 samples, it fades-in in 28.33 samples for one zero point crossing. So it's always too late to slow down.

> Interpolate the new speed respective to the old to not have a steep edge

If we interpolate using `currentSpeed = (oldSpeed+newSpeed)/2.0`, It might be harsher than the current one. Perhaps something like `currentSpeed = (3*oldSpeed+newSpeed)/4.0` can work better. Interpolation is a great idea, I need to experiment.

JohannesLorenz: You could also try something like `currentSpeed = f * newSpeed + (blockSize-f) * oldSpeed` (where `f` is the frame running from `0` to `blockSize - 1`).

src/core/Instrument.cpp
JohannesLorenz: Do you want to add a further comment for the formula, or should we resolve this comment?

necrashter: Should we also define `128` as a constant? I will make a `const static` variable in function to define `85`, is that all right?

JohannesLorenz: Would be OK for me. You can also make it const static in the class under `private` if you want. Probably both is equally good.

necrashter:
This fade in works while playing notes on the Instrument Window, but it doesn't work when you play a pattern in piano roll. Does anyone know why?

PhysSong:
> This fade in works while playing notes on the Instrument Window, but it doesn't work when you play a pattern in piano roll. Does anyone know why?

I'll look into the issue tomorrow. Sorry for the delay.

PhysSong:
@necrashter When you play a note in the piano-roll, the note may not start at the beginning of the buffer cycle. You can get the offset using `NotePlayHandle::noteOffset()` to apply the fade-in from the beginning of the note, not the beginning of the buffer.

src/core/Instrument.cpp
```diff
@@ -78,8 +82,96 @@ bool Instrument::isFromTrack( const Track * _track ) const
 	return( m_instrumentTrack == _track );
 }
 
+// helper function for Instrument::applyFadeIn
+int count_zero_crossings(sampleFrame *buf,fpp_t start, fpp_t frames)
```
PhysSong, line 86: Can you make it a static function or a static member function and rename it to use camelCase?
Also, please add a whitespace after the first comma.

necrashter: Done!

src/core/Instrument.cpp
```diff
@@ -78,8 +82,96 @@ bool Instrument::isFromTrack( const Track * _track ) const
...
+	f_cnt_t total = n->totalFramesPlayed();
+	if (total == 0)
+	{
+		const fpp_t frames = n->framesLeftForCurrentPeriod();
+		const f_cnt_t offset = n->offset();
 
+		// We need to skip the first sample because it almost always
+		// produces a zero crossing; it's not helpful while
+		// determining the fade in length. Hence 1
+		int maxZeroCrossings = countZeroCrossings(buf, offset+1, offset+frames);
```
Veratil, line 137:
```suggestion
		int maxZeroCrossings = countZeroCrossings(buf, offset + 1, offset + frames);
```

src/core/Instrument.cpp
```diff
@@ -78,8 +82,96 @@ bool Instrument::isFromTrack( const Track * _track ) const
...
+	// Length is inversely proportional to the max of zeroCrossings,
+	// because for low frequencies, we need a longer fade in to
+	// prevent clicking.
+	return (fpp_t) (maxLength  / ((float)zeroCrossings / ((float)frames/128.0f) + 1.0f));
+}
+
 
+void Instrument::applyFadeIn(sampleFrame * buf, NotePlayHandle * n)
+{
+	const static float MAX_FADE_IN_LENGTH = 85.0;
```
Veratil, line 127: Do we want this constant here or in a header?

src/core/Instrument.cpp
```diff
@@ -78,8 +82,96 @@ bool Instrument::isFromTrack( const Track * _track ) const
...
+		fpp_t length = getFadeInLength(MAX_FADE_IN_LENGTH, frames, maxZeroCrossings);
+		n->m_fadeInLength = length;
+
+		// apply fade in
+		length = length < frames ? length : frames;
+		for (fpp_t f = 0; f < length; ++f)
+		{
+			for (ch_cnt_t ch = 0; ch < DEFAULT_CHANNELS; ++ch)
+			{
+				buf[offset+f][ch] *= 0.5 - 0.5 * cosf(F_PI * (float) f / (float)n->m_fadeInLength);
```
Veratil, line 148:
```suggestion
				buf[offset + f][ch] *= 0.5 - 0.5 * cosf(F_PI * (float) f / (float) n->m_fadeInLength);
```

src/core/Instrument.cpp
```diff
@@ -78,8 +82,96 @@ bool Instrument::isFromTrack( const Track * _track ) const
...
+			}
+		}
+	}
+	else if(total < n->m_fadeInLength)
```
Veratil, line 152:
```suggestion
	else if (total < n->m_fadeInLength)
```

src/core/Instrument.cpp
```diff
@@ -78,8 +82,96 @@ bool Instrument::isFromTrack( const Track * _track ) const
...
+		const fpp_t frames = n->framesLeftForCurrentPeriod();
+
+		int new_zc = countZeroCrossings(buf, 1, frames);
+		fpp_t new_length = getFadeInLength(MAX_FADE_IN_LENGTH, frames, new_zc);
+
+		for (fpp_t f = 0; f < frames; ++f)
+		{
+			for (ch_cnt_t ch = 0; ch < DEFAULT_CHANNELS; ++ch)
+			{
+				float currentLength = n->m_fadeInLength*(1.0f-(float)f/frames) + new_length*((float)f/frames);
```
Veratil, line 163:
```suggestion
				float currentLength = n->m_fadeInLength * (1.0f - (float) f / frames) + new_length * ((float) f / frames);
```

src/core/Instrument.cpp
```diff
@@ -78,8 +82,96 @@ bool Instrument::isFromTrack( const Track * _track ) const
...
+				buf[f][ch] *= 0.5 - 0.5 * cosf(F_PI * (float) (total+f) / currentLength);
```
Veratil, line 164:
```suggestion
				buf[f][ch] *= 0.5 - 0.5 * cosf(F_PI * (float) (total + f) / currentLength);
```

src/core/Instrument.cpp
```diff
@@ -78,8 +82,96 @@ bool Instrument::isFromTrack( const Track * _track ) const
...
+				if(total+f >= currentLength)
```
Veratil, line 165:
```suggestion
				if (total + f >= currentLength)
```

src/core/Instrument.cpp
```diff
@@ -78,8 +82,96 @@ bool Instrument::isFromTrack( const Track * _track ) const
...
+}
+
+// helper function for Instrument::applyFadeIn
+fpp_t getFadeInLength(float maxLength, fpp_t frames, int zeroCrossings)
+{
+	// calculate the length of the fade in
+	// Length is inversely proportional to the max of zeroCrossings,
+	// because for low frequencies, we need a longer fade in to
+	// prevent clicking.
+	return (fpp_t) (maxLength  / ((float)zeroCrossings / ((float)frames/128.0f) + 1.0f));
```
Veratil, line 121:
```suggestion
	return (fpp_t) (maxLength  / ((float) zeroCrossings / ((float) frames / 128.0f) + 1.0f));
```

src/core/Instrument.cpp
```diff
@@ -78,8 +82,96 @@ bool Instrument::isFromTrack( const Track * _track ) const
...
+	int maxZeroCrossings = 0;
+
+	// determine the zero point crossing counts
+	for (fpp_t f = start; f < frames; ++f)
+	{
+		for (ch_cnt_t ch=0; ch < DEFAULT_CHANNELS; ++ch)
+		{
+			// we don't want to count [-1, 0, 1] as two crossings
+			if ((buf[f-1][ch] <= 0.0 && buf[f][ch] > 0.0) ||
+					(buf[f-1][ch] >= 0.0 && buf[f][ch] < 0.0))
```
Veratil, line 100:
```suggestion
					(buf[f - 1][ch] >= 0.0 && buf[f][ch] < 0.0))
```

src/core/Instrument.cpp
```diff
@@ -78,8 +82,96 @@ bool Instrument::isFromTrack( const Track * _track ) const
...
+	// maximum zero point crossing of all channels
+	int maxZeroCrossings = 0;
+
+	// determine the zero point crossing counts
+	for (fpp_t f = start; f < frames; ++f)
+	{
+		for (ch_cnt_t ch=0; ch < DEFAULT_CHANNELS; ++ch)
+		{
+			// we don't want to count [-1, 0, 1] as two crossings
+			if ((buf[f-1][ch] <= 0.0 && buf[f][ch] > 0.0) ||
```
Veratil, line 99:
```suggestion
			if ((buf[f - 1][ch] <= 0.0 && buf[f][ch] > 0.0) ||
```

src/core/Instrument.cpp
```diff
@@ -78,8 +82,96 @@ bool Instrument::isFromTrack( const Track * _track ) const
...
+{
+	// zero point crossing counts of all channels
+	int zeroCrossings[DEFAULT_CHANNELS] = {0};
+	// maximum zero point crossing of all channels
+	int maxZeroCrossings = 0;
+
+	// determine the zero point crossing counts
+	for (fpp_t f = start; f < frames; ++f)
+	{
+		for (ch_cnt_t ch=0; ch < DEFAULT_CHANNELS; ++ch)
```
Veratil, line 96:
```suggestion
		for (ch_cnt_t ch = 0; ch < DEFAULT_CHANNELS; ++ch)
```

src/core/Instrument.cpp
Veratil: To further elaborate, do we want this to be customizable per `Instrument`?

necrashter: > Do we want this constant here or in a header?

It's matter of preference. I think it's better to put this here, because this constant is used only in this method.

> To further elaborate, do we want this to be customizable per `Instrument`?

Good idea. But we need to consider the possibility that end users might find it too technical. Most users don't want clicks, however it's better to give options to advanced users.

necrashter:
Is there a tool for checking if the code follows the coding conventions?

For the clicks on the note end, there exists `applyRelease` function. As you said, I don't think it works as it should. But it's beyond the scope of this PR.

PhysSong:
I think we can merge it now. @Veratil Do you think so? If you do, please approve this PR.

Issue #3306, opened 2017-01-27
Piano Roll - Vertical arrow cursor stuck


zonkmachine:
![untitled](https://cloud.githubusercontent.com/assets/6368949/22370657/d00175f4-e492-11e6-8c04-3959e93d4a9e.png)

In the Piano Roll, if you move the cursor over the border to the Note Edit Area (Bottom...) the cursor is changed to an override cursor with vertical arrows. This is true for all the edit modes but the override cursor is only dropped when moving out of the area if you're in draw mode. For the other modes, the only way to drop the vertical arrows is to switch to draw mode.

The switch to vertical arrow happens here:
https://github.com/LMMS/lmms/blob/master/src/gui/editors/PianoRoll.cpp#L1987:L1988

BaraMGB:
this would be fixed in https://github.com/LMMS/lmms/pull/3065

PR #5200, opened 2019-09-21
Set cursor local instead of global in PianoRoll.

CYBERDEViLNL:
Set cursor using `setCursor` instead of the global `QApplication::setOverrideCursor` in the PianoRoll.

Also added extra `if else` (https://github.com/LMMS/lmms/compare/master...CYBERDEViLNL:master_pianoroll_cursor#diff-ecf6965b3ecabc1b5c4cb7ded8c130bfR2530) so the cursor gets restored if in Erase, Select or in EditDetuning mode and did not trigger by something else. That fixes https://github.com/LMMS/lmms/issues/3306

### DIFF
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -1404,7 +1404,7 @@ void PianoRoll::keyPressEvent(QKeyEvent* ke)
 			{
 				m_ctrlMode = m_editMode;
 				m_editMode = ModeSelect;
-				QApplication::changeOverrideCursor( Qt::ArrowCursor );
+				setCursor( Qt::ArrowCursor );
 				ke->accept();
 			}
 			break;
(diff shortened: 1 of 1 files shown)


src/gui/editors/PianoRoll.cpp
```diff
@@ -2553,6 +2527,13 @@ void PianoRoll::mouseMoveEvent( QMouseEvent * me )
 				}
 			}
 		}
+		else if (me->buttons() == Qt::NoButton &&
+			( m_editMode == ModeErase ||
+			m_editMode == ModeSelect ||
+			m_editMode == ModeEditDetuning))
```
PhysSong: I'm fine with this, but I'm curious why you didn't wrote it `m_editMode != ModeDraw`.

CYBERDEViLNL: No reason, I'll change it for simplicity :+1:

zonkmachine:
> That fixes #3306

It does indeed, thanks!

zonkmachine:
@PhysSong You've reviewed this and it looks ready to merge.

knarrff:
I cannot say a lot about the functionality change, but except the one comment about spacing in one line I left, everything else looks clear from a style point of view.

PhysSong:
I think we need to merge this as-is because I can't push more changes into this PR.

zonkmachine:
OK, merge?

Issue #3654, opened 2017-06-22
Opening combo box screens key events 


PhysSong:
How to reproduce:
- Add a new Monstro instrument track
- Click small triangle button next to sine wave pixmap
- Press any key shortcut, it won't work.

Reason: ComboBox uses QMenu to show contents, and it screens key events.

Is there anyone who knows how to fix this?

tresf:
@PhysSong I recommend looking into the [`keyPressEvent`](http://doc.qt.io/qt-5/qwidget.html#keyPressEvent) on the QMenu in general however this drop-down is actually a custom made ComboBox (not a native `QComboBox` I don't think).

I recommend binding a `keyPressEvent` that propagates to the parent widget.  It seems to be trapping the keystrokes inside the widget.

For starters, I'd mimic the [`mousePressEvent`](https://github.com/LMMS/lmms/blob/32f07ac317b42dda42a1d00b0a3b1fa1feef0655/src/gui/widgets/ComboBox.cpp#L119)

tresf:
Ok, you're right, it uses a [QMenu under the covers](https://github.com/LMMS/lmms/blob/32f07ac317b42dda42a1d00b0a3b1fa1feef0655/include/ComboBox.h#L71).  Since AFAIK there's no externally exposed API method to override `QMenu`'s keypress events, you will likely need to subclass the menu to grab the proper virtual function.  Seems overkill if coming from a non-C++ language, but I think it's needed.  I had a similar conversation with `jasp00` about this in [this thread](https://github.com/LMMS/lmms/pull/3219#issuecomment-271134612).

PR #5202, opened 2019-09-26
Fix stuck notes when menu is open

### LINKED ISSUE #4892: Stuck notes when menu is open
Found by @DomClark. Closely related to #1755.

Steps to reproduce:
1. Open TripleOscillator.
2. Open the edit menu.
3. Press some keys on the keyboard to play notes in TripleOscillator.

This causes the notes to stick.

Reproduced on LMMS 1.2.0 RC8 and master (as of [https://github.com/LMMS/lmms/pull/4639#issuecomment-425760993](this build)).

necrashter:
Fixes #4892

I created a subclass of QMenu and overloaded the `keyReleaseEvent` to send the event to the parent widget. Passing the event directly didn't work. So I cloned the event in the stack and sent it to parent. [This seems to be the usual practice.](https://doc.qt.io/archives/qq/qq11-events.html)

This does NOT solve the related bug #3654 because it's about `keyPressEvent`. Also QMenu blocks the CTRL key. Try opening the edit menu and then pressing CTRL+S. QMenu only sends the S key.


PhysSong:
Could you add license information at the beginning of the new files?

necrashter:
> Could you add license information at the beginning of the new files?

I added the license information, but I didn't add any name. Should I add my legal name?

JohannesLorenz:
I just tested exactly what @SecondFlight mentioned in the issue description. The problem still seems to exist. Am I testing right?

necrashter:
@JohannesLorenz I tried this PR again. It seems to fix the problem for me. I am running on Ubuntu 19.04, perhaps the problem is platform-dependent?

Could you please provide more information? I open the edit menu on top and play some notes, and the notes don't get stuck unlike before.

JohannesLorenz:
I just rebuilt at 8a4636a97259da1924e79b24ab857aaf19895e74 to make absolutely sure. It still fails. Like you, I open 3osc, play some notes pressing keyboard keys (works as usual), then click edit, and while the edit menu is open, press the keyboard keys again. Then the 3osc notes get green and play (correct behavior), but releasing the key, they remain green and remain playing (incorrect).

My system is Manjaro, updated regularly.

JohannesLorenz:
As necrashter assumed, the issue is Qt4 vs Qt5: The fix works with `-DWANT_QT5=ON`, but not without. I just confirmed it on the same machine.

zonkmachine:
> As necrashter assumed, the issue is Qt4 vs Qt5: The fix works with `-DWANT_QT5=ON`

So, the easiest fix then is to retarget this to master, which is Qt5 only.

JohannesLorenz:
I wonder if this is a "critical" issue - It's not "normal" to play notes while opening a menu, and even when you do, the stuck notes can be easily resetted by leaving the menu and playing the stuck notes again.

I'm OK with a Qt5-only solution, or no solution at all (though a Qt4+5 solution would be nice).

Opinions?

CYBERDEViLNL:
Can you try this against Qt 5.6 (or lower)? In https://github.com/LMMS/lmms/pull/5328 @he29-net and I tested identical solution for `QColorDialog` instead of `QMenu` and it didn't work on Qt 5.6 but did for Qt 5.11 and higher (didn't test any version between Qt 5.6 and 5.11).

DomClark:
Is there a more general solution for this? As @CYBERDEViLNL mentioned, this also occurs for `QColorDialog` in #5328, and it happens for me with all the current dialogs as well (open, save, settings, about, soundfont patch, etc.). It would be nice if we could avoid needing to add this code to everything that can steal key events. Also, the bug occurs with all menus in LMMS, including context, combo-box, track button, etc., so those will also need to use the new `Menu` class.

JohannesLorenz:
1. Is it a stable-1.2 bug? Just close the menu and you can continue using LMMS as usual.
2. Is it even a bug? It's not wrong behavior to block the key events, it's just one possible behaviour.
3. Is it even worth a feature request? Why shall this be changed?

PhysSong:
> Is it even a bug?

I think yes because it's not normal to make keys stuck in that situation.
> Is it a stable-1.2 bug?

It's a bad UX, but I don't think this is critical.
> didn't test any version between Qt 5.6 and 5.11

I can try with 5.9.

PhysSong:
It works well with Qt 5.9.

JohannesLorenz:
OK, let's treat it as a bug which shall be fixed on master.

@necrashter I'll change the PR base to master if you don't disagree within 7 days..

necrashter:
@JohannesLorenz OK, please go ahead.

tresf:
> Also, the bug occurs with all menus in LMMS, including context, combo-box, track button, etc., so those will also need to use the new Menu class.

I find this question to be very important yet not answered.  Is the intention to fix it holistically?

JohannesLorenz:
@necrashter What are you going to do with this PR?

necrashter:
I checked the Qt documentation for a more general solution, but couldn't find any. In fact, [the documentation](https://doc.qt.io/qt-5/qwidget.html#keyReleaseEvent) says that the default implementation of keyReleaseEvent ignores the event, so that the widget's parent can interpret it. What we are trying to achieve seems to be the expected behavior. We can close this if you think this bug is related to Qt, not LMMS.

Also in the latest version, context menus seem to block all key events. Should we do the same with QMenu for more consistent behavior? If so, I can change this PR accordingly, and change other `QMenu`s to use the new `Menu` class.

yohannd1:
Bumping this PR - I wonder how the state of this PR is, now that years have passed? I think Qt4 is not supported anymore, but the Qt5 version might still be lagging a bit behind on some targets on the CI. I think, however, that shouldn't be a blocker if it will _eventually_ work, and it is not a critical bug.

Also, in response to [this comment](https://github.com/LMMS/lmms/pull/5202#issuecomment-570961387), I believe there might be a better solution, though I don't quite see it yet. I've fixed a bug similar in nature recently (https://github.com/LMMS/lmms/pull/8282) but I don't know if the cause is quite the same, and whether we can apply it to any widget we want. The solution I made goes in a different direction: it detects whether the window has lost focus and then specifically marks the shift modifier as not-pressed.

messmerd:
Currently the minimum Qt version required to build LMMS is 5.15, which is the latest Qt5 version. All CI builds use that version except Windows MSVC which uses Qt 6.

yohannd1:
Ah, I imagine the bugfix might be working everywhere then (can't really test though).

Issue #241, opened 2014-02-02
User added scales.


zonkmachine:
There should be a way for users to add scales. Today the scales are hard coded but they could instead be read from a file and also from user defined templates.
How about adding a scale by actually playing it?

sphenocorona:
Old issue I know, but here's another reason to allow some sort of user-made custom scale/chord 'pack/bank' sorta thing: If someone is using, say, ZynAddSubFx and trying to make microtonal music that has more than 12 notes per octave, it can get disorienting and confusing to remember which notes you're using and where the octaves occur. Since you can't save which notes are marked either (which might be another good option), it would be very helpful to at least have a shortcut for re-marking notes on the piano roll.
Though, saving those marked notes might at least serve as a "quick fix" that would at least remove some hassle.
It doesn't make sense to bundle scales like that into LMMS, so... that isn't a practical solution in my opinion.

musikBear:
in fact user 'scales' can be created since 1.1
Use the context option *'Mark all corresponding....'*
Now you will have marked out the 1. note of your user-scale
Repeat for n-notes
The cravat here, is that these 'marked notes' are not preserved within the project-file, so this tedious task has to be repeated every time you open the project, but it can be done..

sphenocorona:
That's the problem, it's very tedious. I mentioned that method in my last post, saying that another option might just be to make an option to save the marked notes in the project file.

psxws:
here's a great resource, a compedium of all possible scales analyzed mathematically

https://ianring.com/musictheory/scales

this might provide a framework for user added scales. there's certainly many missing, i myself found this topic looking for a way to add the ["Youlan Scale"](https://ianring.com/musictheory/scales/1783) as none of the current scales are nonatonic.... but a custom nonatonic scale would be even better, GUI suggestion would be to represent scales as in that website, & perhaps offer the same navigation between them (modes, prime, chirality, etc)

musikBear:
@psxws Great find!
Adding to **Resources** at next update!

psxws:
@musikBear a proper implementation of this allowing the user to set any mathematically possible scale & have it work natively is a killer feature no other software let alone a DAW let's you do, as far as I'm aware.

It's true innovation & would set LMMS apart & allow users to Craft music that pushes boundaries & I think this should be taken seriously & implemented quickly considering how awesome it'll be..

dan-giddins:
to update this, currently, as of 1.3.0-alpha, we have the following functionality
[...]

PR #5204, opened 2019-09-26
Add loading of user-defined scales/chords

Veratil:
Enhancement for #241.

This will look for a `notes.xml` file in the default working directory. If found, it will load those chords and scales instead of the default ones. I decided to load only one or the other because it gives the option (currently requiring manual moving/deleting/updating/whatever of the file) of changing out XML files in case a user wanted to only see certain chords/scales for a project.

Possible future enhancement: Link projects to custom scale/chord files.

The XML file is laid out as so:
```xml
<notes>
  <chords>
    <chord name="Major">
      <semiTone key="0"/>
      ...
    </chord>
    ...
  </chords>
  <scales>
    <scale name="Major">
      <semiTone key="0"/>
      ...
    </scale>
    ...
  </scales>
</notes>
```

As the current chords/scales are set with the equivalent to [pitch class integer notation](https://en.wikipedia.org/wiki/Pitch_class#Integer_notation) the `key`s follow the same.

There is a little leeway with getting the `<semiTone>` nodes as it doesn't actually look for the node name, so technically as long as the `key` attribute is there it could be called anything.

---

Additionally, this PR "fixes" the `Chord stacking & arpeggio` tab to only list actual chords in the Stacking combobox, instead of including scales as well. I doubt anyone wants to hit all the notes of the scale at the same time. I left the Arpeggio combobox to include them though.

Another slight change is to the `octave` chord to actually play the next octave up (I don't see the point of having it if it's just the note you select). If this is incorrect then I can revert that line.

lmms-user-chords

lmms-user-scales


PhysSong:
> to the `octave` chord to actually play the next octave up (I don't see the point of having it if it's just the note you select)

I think the real problem is the misleading name. IMO it should be `none(octave)` because you can use the chord(?) to play with notes with octave intervals.

zonkmachine:
Completely missed this PR. Sorry!

https://github.com/LMMS/lmms/blob/c755b56a52145a572247303a09d126e764238c11/include/InstrumentFunctions.h#L83-L86

Old way... Yeah, this was maybe a bit too optimistic...

I thought I could just restart the failing Travis build job and an AppImage would magically appear somewhere. That assumption was incorrect.

zonkmachine:
First impressions.

* I'm now seeing chords and scales where they're supposed to be. Thanks!
* I managed to get a closing tag wrong at my first attempt. It would be good with some kind of warning that a failed attempt to load notes.xml had occurred.
* I think it's better to load user scales on top of or after the default ones. I suggest to have the first entry in the chords/scales tab be a directory to user files if any are present.
* Maybe add `~/lmms/notes` directory, either by default or just looking for its presence, where you can arrange your material.

Veratil:
> I managed to get a closing tag wrong at my first attempt. It would be good with some kind of warning that a failed attempt to load notes.xml had occurred.

I might change to JSON with the recent idea for switching over. It would be a lot easier to edit.

For instance
```json
{
  "scales": [
    { "name": "Major", "keys": [0, 2, 4, 5, 7, 9, 11] }
  ],
  "chords": [
    { "name": "Major", "keys": [0, 4, 7] }
  ]
}
```

> I think it's better to load user scales on top of or after the default ones. I suggest to have the first entry in the chords/scales tab be a directory to user files if any are present.

I tossed the idea around, but adding to the end of a *long* list would be annoying. My ideal solution is the idea of attaching a scales/chords file to project files, so they're auto loaded per project. There's a few ways we can handle this really. This PR is a stepping stone to get it in first! 😁

> Maybe add `~/lmms/notes` directory, either by default or just looking for its presence, where you can arrange your material.

Yeah, like above there's a few ideas we can move to from this. I'd like some more feedback from people who want to use it though.

dan-giddins:
@Veratil are you still working on this one? If not, I am happy to take it on...

I agree with the move to JSON. I think its a much more concise way of displaying this information.

I would argue 'keys' should be renamed to 'notes' (i.e. notes of a scale) as a scale's 'key' is something different (the tonic root of a scale). In this case, you are defining a generic set of numerical values that make up the scale.

Veratil:
> 
> 
> @Veratil are you still working on this one? If not, I am happy to take it on...
> 
> I agree with the move to JSON. I think its a much more concise way of displaying this information.
> 
> I would argue 'keys' should be renamed to 'notes' (i.e. notes of a scale) as a scale's 'key' is something different (the tonic root of a scale). In this case, you are defining a generic set of numerical values that make up the scale.

I put it aside for a while to work on other things, this was from when I was early learning the codebase. If you'd like to fork my branch and make a PR against it I'd be happy to merge them in here, too.

luzpaz:
Anyone want to pick up the mantle on this one ?

PR #5217, opened 2019-10-02
Xpressive plugin - updated plugin artwork, updated the formula in the…

gnudles:
… presets to use integrate(f) instead of t*f, as integrate operation is more robust to frequency changes. and renamed the plugin to Xpressive instead of X-Pressive


gnudles:
Screenshot from 2019-10-02 22-09-37
I'm not the greatest artist, but made it look a bit better.

gnudles:
does anyone have a clue why it fails?

PhysSong:
The build error is not related to your code.

plugins/Xpressive/Xpressive.cpp
```diff
@@ -53,7 +53,7 @@
 extern "C" {
 
 Plugin::Descriptor PLUGIN_EXPORT xpressive_plugin_descriptor = { STRINGIFY(
-		PLUGIN_NAME), "X-Pressive", QT_TRANSLATE_NOOP("pluginBrowser",
+        PLUGIN_NAME), "Xpressive", QT_TRANSLATE_NOOP("pluginBrowser",
```
PhysSong: Spaces.

gnudles: right... spaces

gnudles:
let's keep this pull request open for a couple of days to see if anyone may suggest further improvements to this plug-in

RoxasKH:
I really like the fresh feel you give to it with this new artwork, the old was like: "why pink!?"
I especially love the new logo, but imo there's too space left between the 2 letter "S", you should cut it out a bit. Also i'd make the white outcircle of knobs on the right a little bit thinner and give a little bit of more contrast to the waveforms with the bg.
Just another question: there's no way to change also the bar between the graph and the box were you write the formulas? Because those gradients are too much '90s.
Anyway good work

gnudles:
> I really like the fresh feel you give to it with this new artwork, the old was like: "why pink!?"
Thanks. 
> I especially love the new logo, but imo there's too space left between the 2 letter "S", you should cut it out a bit. Also i'd make the white outcircle of knobs on the right a little bit thinner and give a little bit of more contrast to the waveforms with the bg.
Wow, now I notice the distance between the two S. I think the white outline is Ok. what do you mean about the contrast in the waveforms?
> Just another question: there's no way to change also the bar between the graph and the box were you write the formulas? Because those gradients are too much '90s.
I was about to do so.
> Anyway good work
thanks pal.

gnudles:
Maybe someone wants to add sample xpf files?

RoxasKH:
I won't discuss about the outcircle, because they are tastes and it's good either this way. For the logo and the waveforms, you can see all here
xpressive

As you can see there's more contrast with the backround were the waveform icons are, so they are more clearly visible.

gnudles:
Oh I see. ok!

zonkmachine:
Looks good! @gnudles Do you want to see this merged now or are there any more changes coming?

gnudles:
wait, I have some changes

gnudles:
If you think it's ok now, you can merge. I was just thinking about hiding the smooth knob background when it is not active.

zonkmachine:
> I was just thinking about hiding the smooth knob background when it is not active.

I went ahead with a merge as we have a whole lot of open PR's right now.
PS. Great work! :tractor:

PR #5222, opened 2019-10-05
Workaround for submodules failing

tresf:
As discovered by @PhysSong, `git version 2.7.4` -- the default for Ubuntu 16.04 -- has a bug where it doesn't properly honor the `deinit` call, nor the `--depth 2147483647` hack that was previously provided in #5182 as a workaround to the ECM nested submodule failing during shallow clone.

This works around that bug by using a trick @PhysSong found, delete the `.git/modules/<path>`, then call `deinit`, then do the `--depth 2147483647` hack.

The advantage this has over just forcing full clones for older git versions is it fixes the issue reactively for offending submodules (e.g. ECM) without doing full clones for everything (e.g. the huge CalfStudioGear submodule).


PhysSong:
Fix confirmed, but I recommend deleting the directory after deinitialization. I can't deinitialize the submodule without `-f` if I remove the directory before the deinitialization.

tresf:
Missy Elliott has flipped it and reverse it.  Tested on 16.04 and it seems to work.

tresf:
> Good. Could you rebase to squash the first commit and the last commit before merging while leaving the second commit as-is?

I refuse to rebase to such microscopic detail.  If you want to, go ahead.

tresf:
Per Discord:

> I wanted to decouple the bug fix and enhancement, but I'm okay with squashing everything.

PhysSong:
I found that `--depth 2147483647` breaks a nested submodule which points to a non-default branch. I have an idea of how to fix this, but I need to make sure it doesn't break anything that worked previously.

PR #4991, opened 2019-05-25
SampleBuffer -> Samplerate - Fix two sample rate issues:

Reflexe:
1. Resampling was the wrong (src_rate was invalid)
2. SampleBuffer was using baseSampleRate as the default samplerate
instead of the actual processingSampleRate.

Closes #4786

PR #5226, opened 2019-10-06
Make SampleBuffer adjust its members when resampling. Fixes #5218 for stable-1.2

### LINKED ISSUE #5218: Oversampling changes pitch for AFP
See picture, count the waves. 
Upper track: drums in AFP, 1x oversampling
Lower track: the same with 2x oversampling, sounds an octave too low.
oversampling-broken

LMMS version: 1.2.0 in Windows 8.1 pro / 64 bit. Didn't find any fitting reported issue, and I think I've used oversampling successfully in the past, so this might be a recent bug.

softrabbit:
My proposition on how to make oversampling work again in the 1.2 line after the changes in #4991. Feel free to chime in if some use case of SampleBuffer remains broken (or gets a fresh bug) after this PR.

### DIFF
--- a/src/core/SampleBuffer.cpp
+++ b/src/core/SampleBuffer.cpp
@@ -382,6 +382,7 @@ void SampleBuffer::directFloatWrite ( sample_t * & _fbuf, f_cnt_t _frames, int _
 void SampleBuffer::normalizeSampleRate( const sample_rate_t _src_sr,
 							bool _keep_settings )
 {
+	const sample_rate_t old_rate = m_sampleRate;
 	// do samplerate-conversion to our default-samplerate
 	if( _src_sr != mixerSampleRate() )
 	{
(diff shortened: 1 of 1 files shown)


softrabbit:
Oops. This won't work, just realized I must've only tested the latest changes from command line. The loop points will be off when returning to normal sample rate after GUI export. I'll yet be refining this.

Reflexe:
I have heavily refactored this class in the recording branch. Maybe it would be smarter to use that code instead (Will do that hopefully later this day).

softrabbit:
> I have heavily refactored this class in the recording branch. Maybe it would be smarter to use that code instead (Will do that hopefully later this day).

In master, no doubt. It's meant to end up there anyway.

In stable, why not if you already fixed the bug in your branch. If not, I'd be inclined to not backport and just fix. But either way works for me, you are most familiar with your code so you get to decide.

Reflexe:
Looks good, however, maybe it would be better to just avoid normalization altogether? (let `play` handle resampling). 
That will also make sure we won't have to resample again after we finish exporting (which will may reduce quality)

softrabbit:
> Looks good, however, maybe it would be better to just avoid normalization altogether? (let `play` handle resampling).
> That will also make sure we won't have to resample again after we finish exporting (which will may reduce quality)

I believe ```update()``` calls ```normalizeSampleRate()``` only after reloading the original data from file, so there shouldn't be a successive resampling problem.

Reflexe:
> believe `update()` calls `normalizeSampleRate()` only after reloading the original data from file, so there shouldn't be a successive resampling problem.

You're right. I just hate this `update` function so much ;)

I think it just would be better to extract `(float)mixerSampleRate() / old_rate` into one variable and we're done.

PhysSong:
Is it okay if I(or someone else) add a commit addressing https://github.com/LMMS/lmms/pull/5226#issuecomment-540174093 and merge this?

softrabbit:
> Is it okay if I(or someone else) add a commit addressing [#5226 (comment)](https://github.com/LMMS/lmms/pull/5226#issuecomment-540174093) and merge this?

Sure. I will probably not be much around my dev box for the next few days, so go ahead.

PhysSong:
Okay, but I'm not sure how to name the variable.

zonkmachine:
It looks like this is still an issue on drumsynth (.ds) files.
https://lmms.io/forum/viewtopic.php?f=7&t=31122

Issue #403, opened 2014-03-04
Remove global automations


diizy:
So we have this thing called global automations. Not many people even know about them. I've never used them in any serious project. They seem like yet another hardly-discoverable secret feature of LMMS. They work well enough though - you set an automation for a control and it works as if it's an automation track placed at the beginning of the project.

So what's the problem with them? Well... think about it: you add a global automation to a control. You draw the automation curve, close the automation editor... and your control now has a hidden automation track, one which is in no way represented by the UI, one which you can only access from the context menu of the control itself. 

This is not good UI design... it is not WYSIWYG, it is not discoverable, it is not ergonomic. It is a hidden UI feature. There's no visual representation of these automations anywhere apart from when you're actually editing them.

The idea of global automations is good, it's just not implemented in a very user-friendly way. I can see three ways of dealing with this:

1 - remove global automations. For backwards compatibility, load old global automations as regular automation tracks. 

2 - add another editor window (which is hidden by default) that displays all global automations as automation tracks - kind of like the BB-editor, but for automations only.

3 - add a switch to hide/display global automations in the song editor.

Each of these options has their ups and downs. 

I've marked this for 1.1.0.

---

Here are the steps for global automation removal as [outlined by PhysSong](https://github.com/LMMS/lmms/issues/403#issuecomment-506914758):
- [x] 1) Let MIDI import plugin use separate automation tracks instead of global ones (DONE [#5223](https://github.com/LMMS/lmms/pull/5223), [#6605](https://github.com/LMMS/lmms/pull/6605))
- [x] 2) Write an upgrade logic that removes global automation patterns from project files and replaces them with normal automation tracks (DONE [#5229](https://github.com/LMMS/lmms/pull/5229))
- [x] 3) Remove any related GUI menus (DONE [#5230](https://github.com/LMMS/lmms/pull/5230))
- [ ] 4) Remove the feature if needed

badosu:
Can we decide between removing this feature or suggesting a mockup for an UI improvement?

I'd like to take a shot at this issue..

PhysSong:
I found that the feature is confusing for many people, and the confusion results in bug reports. Thus, I'd like to either remove or deprecate this feature. My plan consists of those steps:
1. Let MIDI import plugin use separate automation tracks instead of global ones
2. Write an upgrade logic that removes global automation patterns from project files and replaces them with normal automation tracks
3. Remove any related GUI menus
4. Remove the feature if needed

Opinions and suggestions are welcome.

[...]

PR #5223, opened 2019-10-06
Remove global automation from MidiImport

Veratil:
Completes step 1 from #403

PR #5229, opened 2019-10-07
Remove song import global automation

Veratil:
Completes step 2 from #403 

I've also removed the global automation `<track>` node from the templates. I noticed in `default.mpt` that there were `<object>` nodes, but no related `id` in any other nodes like there would normally be if there was a pattern connected. I believe this might be a bug in the automation editor.


PhysSong:
I think `DataFile` is a better place to handle the upgrade stuff rather than `Song`.

Veratil:
> I think `DataFile` is a better place to handle the upgrade stuff rather than `Song`.

I tried that initially, but it didn't do anything. I'll try playing with it some more.

Veratil:
@PhysSong okay it took some playing around with finding nodes, but I finally managed to get it working in DataFile.

PhysSong:
@Veratil Could you look into factory project files to see if any of them has global automation tracks?

Veratil:
> @Veratil Could you look into factory project files to see if any of them has global automation tracks?

@PhysSong I believe that's already taken care of in the second commit, or are there other ones that I missed?

PhysSong:
It will be okay if you've checked all demo projects and confirmed that there are no global automation tracks that you need to remove.

Veratil:
> It will be okay if you've checked all demo projects and confirmed that there are no global automation tracks that you need to remove.

I've gone through and found that some of the demos are still adding tracks that aren't linked to anything. I've updated the check to look for more than one `<time>` node now which has prevented adding extraneous automation tracks which aren't linked to anything.

There is one demo that has an unlinked automation track which has mulitple time nodes.

@PhysSong Would you like me to go through and update all the demo files in this patch to the "newest" save file version then?

PhysSong:
> @PhysSong Would you like me to go through and update all the demo files in this patch to the "newest" save file version then?

If you want to do so, I'd recommend not touching files without global automations. I'm also okay if you do that in a separate pull request.

Veratil:
> 
> 
> > @PhysSong Would you like me to go through and update all the demo files in this patch to the "newest" save file version then?
> 
> If you want to do so, I'd recommend not touching files without global automations. I'm also okay if you do that in a separate pull request.

Is there an easy way to read the `mmpz` files? That would be the easiest to see if there is any automation patterns that should be imported.

PhysSong:
> Is there an easy way to read the `mmpz` files?

`lmms -d input.mmpz` prints the decompressed version of `input.mmpz`.

Veratil:
> 
> 
> > Is there an easy way to read the `mmpz` files?
> 
> `lmms -d input.mmpz` prints the decompressed version of `input.mmpz`.

Would it be okay to run `lmms upgrade [in] [out]` on all the files? 😄

PhysSong:
That makes sense, but Qt will mix the order of attributes in XML tags. You can write a simple program to compress/decompress `.mmpz` files with `QByteArray::qCompress()` and `QByteArray::qUncompress()`.

PhysSong:
The result of `qCompress()` consists of the size of the uncompressed data(in 4byte big-endian integer) and the Zlib-compressed data. It means you can use some other software to (de)compress the project files.

JohannesLorenz:
@Veratil Are you going to finish this PR?

Spekular:
Seconded. It would be superb if we could get this and step 3 (remove from the GUI) done for 1.3, and step 3 should be a quick and safe fix.

Veratil:
I need to go through all the demo files still.

PhysSong:
@Veratil Bump.

Veratil:
1. I rebased on master 😁 
1. I updated the update function to the new format.
2. I removed the global automation save from Song's `saveSettings`, no more hidden automations will be saved after this at all.
3. I went through each demo file and actually loaded them in lmms. Any that had an unused hidden automation tracks that actually created visible automation tracks I removed and saved over. Anything else I didn't touch. If they do have the block, it's either ignored or will get removed on future updates where we update the demos for whatever reason.

PhysSong:
Regarding global automation clips with only one point: In many cases, they are just holding the model's value at time 0, but it's not guaranteed that they always do.

PhysSong:
The rest looks good to me.

Veratil:
> 
> 
> Regarding global automation clips with only one point: In many cases, they are just holding the model's value at time 0, but it's not guaranteed that they always do.

That's where thing are weird. Every global automation track that had one node was either a leftover that wasn't removed (file bloat because there wasn't a way to delete the global track), or didn't actually affect the model's value anymore (that value was saved as an attribute instead or didn't actually connect anymore).

I can make the change to load it regardless, and I'll go through all the projects again to see if anything else pops up.

PhysSong:
I looked into this PR again. This PR will change some behaviors:
- Global automation patterns have lower priority than ordinary ones, but this PR will reverse it due to the order of XML nodes
- Global automation patterns with single point which "locks" the value of connected models will be removed
- Global automation patterns created by user will never be changed

The first one sounds undesirable, but the rest don't sound too bad once #5230 is merged and https://github.com/LMMS/lmms/pull/5223#issuecomment-759225962 is addressed.

src/core/DataFile.cpp
```diff
@@ -1385,6 +1387,44 @@ void DataFile::upgrade_noHiddenClipNames()
...
+	// convert global automation tracks to non-hidden
+	QDomElement song = firstChildElement("lmms-project")
+		.firstChildElement("song");
+	QDomElement tc = song.firstChildElement("trackcontainer");
+	QDomElement gaTrack = song.firstChildElement("track");
+	if (!gaTrack.isNull()
+		&& gaTrack.attribute("type").toInt() == Track::HiddenAutomationTrack)
+	{
+		// global automationpatterns
+		QDomNodeList aps = gaTrack.elementsByTagName("automationpattern");
```
PhysSong, line 1401: This should be changed to `"automationclip"` now.

Veratil: Done!

Veratil:
Rebased on master and (hopefully) fixed conflicts (correctly). Tested build and it seemed to be working.
I'll go back through all the demos to double check everything, those are the biggest changes for removing the hidden automation tracks.

Veratil:
Looks like everything is still good in the demos. 👍

PhysSong:
> Global automation patterns have lower priority than ordinary ones, but this PR will reverse it due to the order of XML nodes

I think this is now the only one which probably needs to be addressed. @Veratil any opinions?

Veratil:
> > Global automation patterns have lower priority than ordinary ones, but this PR will reverse it due to the order of XML nodes
> 
> I think this is now the only one which probably needs to be addressed. @Veratil any opinions?

I'm honestly not sure what you mean by priority here. If two automation clips are connected to the same object the global wins?

PhysSong:
> If two automation clips are connected to the same object the global wins?

Yes, I mean that case.
https://github.com/LMMS/lmms/blob/1f30ffc5e4dddbd6ffb843bd8c56a8b4de859675/src/core/Song.cpp#L841-L844
Here you can see the global automation track is the first element before removal. You can also see that `TrackContainer::automatedValuesFromTracks` uses the value from the last track connected to an object.

Veratil:
Okay I'll try to see what can be done for this to keep them in order. It never occurred to me that you could connect an object to multiple automations.

Rossmaxx:
Update?

PhysSong:
> Update?

It's ready for review now.

Rossmaxx:
Is the track order thing fixed now?

data/projects/templates/AcousticDrumset.mpt
```diff
@@ -114,8 +114,8 @@
       <automationpattern prog="0" mute="0" name="Master volume" pos="0" tens="1" len="192"/>
       <automationpattern prog="0" mute="0" name="Master pitch" pos="0" tens="1" len="192"/>
     </track>
-    <mixer visible="1" width="561" height="332" x="10" y="314" maximized="0" minimized="0">
```
Rossmaxx, line 117: Why is mixer renamed to fxmixer again? Is this some regression?

Rossmaxx: You gotta look for this at mmpz files too.

data/projects/demos/StrictProduction-DearJonDoe.mmp
```diff
@@ -1,643 +1,670 @@
 <?xml version="1.0"?>
 <!DOCTYPE lmms-project>
-<lmms-project creator="LMMS" version="1.0" type="song" creatorversion="1.2.0">
```
Rossmaxx, line 3: Is this force resave necessary? If so, what are these new tags?

data/projects/demos/StrictProduction-DearJonDoe.mmp
```diff
@@ -1,643 +1,670 @@
...
+              <pattern type="0" muted="0" pos="768" steps="16" name=""/>
+              <pattern type="0" muted="0" pos="960" steps="16" name=""/>
+              <pattern type="0" muted="0" pos="1152" steps="16" name=""/>
+              <pattern type="0" muted="0" pos="1344" steps="16" name=""/>
+              <pattern type="0" muted="0" pos="1536" steps="16" name="">
+                <note vol="100" pan="0" len="-192" pos="0" key="57"/>
               </pattern>
             </track>
           </trackcontainer>
         </bbtrack>
```
Rossmaxx, line 375: I thought we renamed it to pattern

Rossmaxx: The line below

data/projects/templates/CR8000.mpt
```diff
@@ -232,8 +232,8 @@
       <automationpattern prog="0" mute="0" name="Master volume" pos="0" tens="1" len="192"/>
       <automationpattern prog="0" mute="0" name="Master pitch" pos="0" tens="1" len="192"/>
     </track>
-    <mixer visible="1" width="561" height="332" x="10" y="314" maximized="0" minimized="0">
-      <mixerchannel num="0" muted="0" volume="1" name="Master" soloed="0">
+    <fxmixer visible="1" width="561" height="332" x="10" y="314" maximized="0" minimized="0">
```
Rossmaxx, line 235: All these other presets seems to have this change.

Rossmaxx (commented):
Got a few concerns from the diff.

PhysSong:
@Rossmaxx Factory projects probably requires re-saving again, I think

Rossmaxx:
Yeah, or unstaging them for now.

PhysSong:
> Yeah, or unstaging them for now.

I think we can defer re-saving to right before releasing 1.3 or any other sane point. Unstaging for now.

I also manually removed related parts in factory templates and tutorials.

Rossmaxx (approved):
Everything looks good for now.

Rossmaxx:
Removing global automation shouldn't have waited this long.

src/core/Song.cpp
```diff
@@ -1239,7 +1233,6 @@ bool Song::saveProjectFile(const QString & filename, bool withResources)
 
 	saveState( dataFile, dataFile.content() );
 
-	m_globalAutomationTrack->saveState( dataFile, dataFile.content() );
```
michaelgregorius, line 1242: Shouldn't the member `m_globalAutomationTrack` be fully removed as well?

src/core/DataFile.cpp
```diff
@@ -1638,6 +1639,51 @@ void DataFile::upgrade_1_3_0()
...
+	// convert global automation tracks to non-hidden
+	QDomElement song = firstChildElement("lmms-project")
+		.firstChildElement("song");
+	QDomElement tc = song.firstChildElement("trackcontainer");
+	QDomElement gaTrack = song.firstChildElement("track");
+	if (!gaTrack.isNull()
+		&& gaTrack.attribute("type").toInt() == static_cast<int>(Track::Type::HiddenAutomation))
+	{
+		// global automationclips
+		QDomNodeList aps = gaTrack.elementsByTagName("automationclip");
```
michaelgregorius, line 1653: IMO the code would become better readable if that was called `automationClips` instead of `aps`. What does "aps" stand for?

src/core/DataFile.cpp
```diff
@@ -1638,6 +1639,51 @@ void DataFile::upgrade_1_3_0()
...
+		QList<QDomNode> tracksToInsert;
+		for (int i = 0; i < aps.length(); ++i)
+		{
+			QDomElement ap = aps.item(i).toElement();
```
michaelgregorius, line 1657: Similar as above. Would be better to understand if `ap` was called `automationClip`.

src/core/DataFile.cpp
```diff
@@ -1638,6 +1639,51 @@ void DataFile::upgrade_1_3_0()
...
+			// If ap has time nodes, move it to trackcontainer
+			// There are times when an <object> node is present without an
+			// object with the same ID in the file, so we ignore that node
+			if (ap.elementsByTagName("time").length() > 1)
+			{
+				QDomElement aptrack = createElement("track");
+				aptrack.setAttribute("muted", QString::number(false));
```
michaelgregorius, line 1664: In general it's a bit unfortunate that the upgrade code must know the implementation details of how other entities represent their data in the XML file. However, that's a design problem and not something for this PR.

michaelgregorius (changes requested):

src/core/Song.cpp
PhysSong: Yes, but in another PR.

src/core/DataFile.cpp
PhysSong: I think `ap` stands for `AutomationPattern` which is `AutomationClip` now. I agree on renaming.

PhysSong: Renamed variables.

src/core/Song.cpp
PhysSong: To elaborate, That is the step 4 in #403, which should be done after #5230(step 3).

michaelgregorius: Ok, thanks for the heads up! :+1:

michaelgregorius (approved):

Rossmaxx:
Shiit, acted too soon. A test from @bratpeki would have helped. But now, you can test on master too.

bratpeki:
It's cool, I'll take a look at master after the weekend! Can you give me a TLDR of all changes made here?

sakertooth:
Should the global automation test in `AutomationTrackTest` been removed?

Rossmaxx:
I believe we should, but I'll wait for @PhysSong 's opinion too

PhysSong:
Maybe we could do as part of step 4 of #403.

Issue #3312, opened 2017-01-31
Project loading-time in RC2.8 significantly longer than 1.1.3


musikBear:
OS  : winXP sp3
LMMS 1.1.90-RC2.8
2 gb ram

I have a project i use as sketchbook. It is 35 mins long.
This project earlier loaded in ~ 50 secd
In RC2.8 this takes ~4.5mins 
It looks like lmms uses progressively longer time for longer files
I have made a file over *SkiesiOnions* that shows this
The file is expanded to 11 mins
Compare the loading of the default *SkiesiOnions* and the the 11 mins 'version'
On **my system** the loading halts around 50% and from there it takes ~3 mins to load the project.
This is different from earlier versions of lmms, and i am sure RC2.2 loaded files faster
Is this also happening on better pc's?
A proper benchmark for loading one *specific* file would perhaps be a good thing to establish, so this feature could be less 'ad hoc' and instead precisely determined on *one specified* system

https://lmms.io/lsp/?action=show&file=10232

michaelgregorius:
I can confirm that loading got significantly slower between 1.1.3 and the current master (c8e987348defe985253e6ae007cb015abd18f6f1). Here are some manual measurements that have been acquired with a stop watch while loading https://lmms.io/lsp/?action=show&file=10232:
1.1.3: ~4.5s
master: ~9.0s

Some profiling with valgrind shows that most of the delta is spent in `MemoryPool::getChunks(int)`. This method in turn is mainly called by `Note::createDetuning()`.

For a visual overview here is the call graph for 1.1.3:
![valgrind - version 1 1 3](https://cloud.githubusercontent.com/assets/9293269/22525160/76828472-e8c6-11e6-80c0-05e0f280ee3c.png)

And here is the call graph for master:
![valgrind - master](https://cloud.githubusercontent.com/assets/9293269/22525184/87d9a4a8-e8c6-11e6-9bb5-a1c1192ad9a9.png)


The problem was also described in #2029.

michaelgregorius:
Commenting out the macro `MM_OPERATORS` in `DetuningHelper` also does not help. Loading the file still takes ~9s in this case. So the cost is in fact caused by the sheer number of `DetuningHelper` instances that are created.

musikBear:
There is a connection to the new theme
If i choose the *Classic* theme, i can load the file above in 1.05 min (stop-watch)
[...]

PR #5236, opened 2019-10-10
Don't recalculate the song length for every added TCO while loading

softrabbit:
Don't make LMMS calculate the song length for every added TCO when loading project, instead do it only once afterwards. Fix for #3312, solution suggested and investigated by @michaelgregorius ages ago.


zonkmachine:
Tried this and without taking time the projects(Skiessi - Onion/RandomProject). loaded about twice as fast. The provided linux AppImage doesn't appear to work though.

softrabbit:
On my system this change cut loading time down from 26 seconds to 5. (Pathological case: https://lmms.io/lsp/?action=show&file=10232). Compiled and run under identical conditions, of course.

zonkmachine:
Right. With the pathological case and on my puny netbook I get a consistent result of 2 minutes 20 seconds without the patch and 20 seconds with it.

src/core/Song.cpp
```diff
@@ -608,6 +608,9 @@ void Song::playPattern( const Pattern* patternToPlay, bool loop )
 
 void Song::updateLength()
 {
+	if(m_loadingProject) {
```
zonkmachine:
```cpp
if(m_loadingProject)
{
```

qnebra:
Nice improvement in loading time. In my latest project current loading time was around 31:50 seconds in default lmms 1.2.0. Exactly same project was loaded in 24:40 seconds on lmms version avalaible here.

It was quite heavy project with lot of VST instruments, plugins and automation going on. I think this nicely represents 'normal user case'.

src/core/Song.cpp
PhysSong: I think either of them is fine, but notice the space between `if` and `(`.
```suggestion
	if (m_loadingProject) {
```
```suggestion
	if (m_loadingProject)
	{
```

zonkmachine: > I think either of them is fine, but notice the space between `if` and `(`.

It's actually in the [coding convention](https://github.com/LMMS/lmms/wiki/Coding-Conventions#braces) but I'm really happy with either one. You're second suggestion would be the correct one. The rest of the file adheres to the brackets on separate lines. I guess this will be run over by a code formatting script some day anyway.

zonkmachine:
Wouldn't this be a good fix for lmms-1.2.1 ?

src/core/Song.cpp
softrabbit: The coding convention allows it, so I compacted the if clause into one line.

qnebra:
I think yes.

zonkmachine:
Merge? In stable-1.2?

zonkmachine:
I saw no issues with this PR while testing for a couple of hours.
@PhysSong Do you want to test this further? I think we should try and get this merged.

knarrff:
I am no LMMS developer, but to me it seems this change forceably changes UpdateLength() to be a no-op during load. I could imagine some other part of the code actually having good reasons to call UpdateLength() and expect it to work, even when loading a song.

Is it not possible to look for unnecessary calls of UpdateLength() and not call it from there while loading instead of making it a no-op? That would also solve the issue of long loading times, but it would leave UpdateLength() working for those that might have to rely on it.

Also, even if you do not strictly have to rely on it, the current patch could easily lead to hard-to-debug bugs, since a developer, rightfully so, might assume UpdateLength() do what it is supposed to be doing, while in fact, with the patch, it will not longer always do so, and there is no indication to the calling function when it does not.

softrabbit:
> I am no LMMS developer,

Neither am I, TBH. I just throw a PR at the project now and then.


> but to me it seems this change forceably changes UpdateLength() to be a no-op during load. I could imagine some other part of the code actually having good reasons to call UpdateLength() and expect it to work, even when loading a song.

Well, the good reasons would IMO be GUI updates, playback and audio export (which really is a special kind of playback). I expect no playback of any kind to happen while loading a project, which leaves the GUI. I can live without the song editor being up-to-date behind the import dialog. None of the actual data in the song should be affected, as that is what the song length is calculated from, not the other way around.


> Is it not possible to look for unnecessary calls of UpdateLength() and not call it from there while loading instead of making it a no-op? That would also solve the issue of long loading times, but it would leave UpdateLength() working for those that might have to rely on it.

You aren't wrong, and this might be the more correct way. 

 
> Also, even if you do not strictly have to rely on it, the current patch could easily lead to hard-to-debug bugs, since a developer, rightfully so, might assume UpdateLength() do what it is supposed to be doing, while in fact, with the patch, it will not longer always do so, and there is no indication to the calling function when it does not.

Looks like a valid point.

PhysSong:
I think #6741 may help fixing the issue which this PR tries to fix.

LostRobotMusic:
It certainly helps quite a bit, but there can still be noticeable delay if the song has enough stuff in it.  

Currently the song length calculations are only needed for the horizontal scrollbar length in the Song Editor and for export.  I can't think of any reasons for song length info ever being needed during project load, even hypothetically.  Merging a PR like this is still something I'd say we should aim for.

IanCaio:
On a quick review it looks good to me: The update method is called after everything is loaded and m_loadingProject is unset. I'd tend to add a comment, since that adds an exception behavior to the method and it wouldn't hurt to make it more explicit in the code (I'd probably comment above the `return` conditional on the update method, mentioning that load and create project will call the method manually after setting m_loadingProject to false). I often comment codes a lot, so if others disagree it's fine by me.

Rossmaxx:
Seems like this PR was supposed to be merged long ago but was left out. The fix doesn't seem to horrible too. @zonkmachine does this PR still have a scope?

michaelgregorius:
The performance problem is still in master. A debug build of master loads the file in the original description in 20 s whereas with the patch of this PR it only takes 3 s.

It seems to be an open question whether this PR should be merged or #6741.

Rossmaxx:
I asked lost on discord, I'll update once i get the reply. I personally feel this PR looks better from the diff.

Rossmaxx (approved):
I approve.

michaelgregorius:
I have just checked out @LostRobotMusic's branch, merged with master and loaded the file. With #6741 it takes 5 s to load the file. So it's:
* master: 20 s
* #6741: 5 s
* #5236: 3 s

Perhaps both PRs could be combined.

LostRobotMusic:
I left a comment in this thread about this around a year ago.  My PR and this PR are tackling two entirely separate and unrelated issues, they have almost nothing to do with each other and should both be merged.

My PR removes the exponential nature of the song length calculation, speeding it up by hundreds of times, which mainly demolishes the enormous lag that results from moving/deleting/copying groups of clips in large projects (and coincidentally improves project loading speed).  This PR makes it so the song length doesn't even need to be calculated in the first place for project loading specifically, which is a good thing because there's no reason for the song editor length to be calculated here at all.

Merging only my PR would leave behind these unnecessary calculations during project load.  Merging only this PR would leave behind 100% of the current debilitating lag present in all song editor operations.  They're unrelated PRs solving separate issues and shouldn't be compared in any way.

PR #5239, opened 2019-10-13
Set loop points from right-clicking pattern, first draft.

softrabbit:
Doesn't yet work for sample tracks or automation. I'll be amending the PR with support for those in a few days. Critique is most welcome, as I'm not entirely up to speed on the GUI side of things here.


PhysSong:
Why don't you just use `ClipView::contextMenuEvent`?

softrabbit:
> Why don't you just use `ClipView::contextMenuEvent`?

No idea, really... it's been close to 5 years now so who cares? 😁
It was probably because I found the spot where the context menu was constructed and added to that.

PR #5243, opened 2019-10-16
Use '_' instead of '-' and no leading '0' when versioning project files

Sawuare:
Because:
- `_` is a more common separator than `-`
- modern operating systems can sort correctly without a leading `0`

`project-01` -> `project_1`

### DIFF
--- a/src/gui/dialogs/VersionedSaveDialog.cpp
+++ b/src/gui/dialogs/VersionedSaveDialog.cpp
@@ -86,7 +86,7 @@ VersionedSaveDialog::VersionedSaveDialog( QWidget *parent,
 
 bool VersionedSaveDialog::changeFileNameVersion(QString &fileName, bool increment )
 {
-	static QRegExp regexp( "[- ]\\d+(\\.\\w+)?$" );
+	static QRegExp regexp( "[_ ]\\d+(\\.\\w+)?$" );
 
 	int idx = regexp.indexIn( fileName );
 	// For file names without extension (no ".mmpz")
@@ -100,7 +100,7 @@ bool VersionedSaveDialog::changeFileNameVersion(QString &fileName, bool incremen
 		if ( increment == false )
 			return false;
 		else
-			fileName.insert( insertIndex, "-01" );
+			fileName.insert( insertIndex, "_1" );
 	}
 	else
 	{
@@ -115,7 +115,7 @@ bool VersionedSaveDialog::changeFileNameVersion(QString &fileName, bool incremen
 			return false;
 		// Replace version number
 		version = increment ? version + 1 : version - 1;
-		QString newnumber = QString( "%1" ).arg( version, 2, 10, QChar( '0' ) );
+		QString newnumber = QString( "%1" ).arg( version );
 
 		fileName.replace( idx+1, number.length(), newnumber );
 	}


Spekular:
> _ is a more common separator than -

Where? I see dashes in filenames more often than underscores, and to me they look nicer. If we're changing this, why not just use a space? "Project 01"/"Project v01"

"Modern OSes can sort without leading zero" is a funny argument to make when LMMS's own sidebar fails to do this for samples. I won't push back on this too hard I'm probably biased, but I prefer a leading zero. Using a single digit and hoping that whatever program the user uses will sort it "intelligently" feels fragile.

Sawuare:
> Where?

Snake case, hashtags, and where a whitespace character is not permitted.

> why not just use a space?

I considered this, but I thought a visible character would be more suitable for something that was added automatically by the program.

> "Modern OSes can sort without leading zero" is a funny argument to make when LMMS's own sidebar fails to do this

Yes, but I think the solution is to fix the sidebar.

Spekular:
"Where whitespace is not permitted" applies to dashes as well. I can't recall seeing snake case anywhere except variable names, and even there I fond camelCase more common.

Hashtags, in my experience, absolutely don't use underscores. They just smash the words together, either lowercase or camelCase or UpperCamerCase. Here's every compound hashtag I found scrolling my timeline just now. Including tweets from political accounts ans major game accounts:

 b

he29-net:
Btw. what happens to existing projects when somebody decides to upgrade to new LMMS? If they have `Project-06.mmpz`, what comes next? `Project-06_1.mmpz`? Or will it recognize the format, convert it and continue by `Project_7.mmpz`?

This seems a bit like a change for the sake of change, that could break more things than it improves.

Sawuare:
@Spekular Hashtags with cursive scripts like Arabic: لبنان_ينتفض#
The hyphen-*minus* represents 2 things, whereas the underscore is just that.

Sawuare:
> what comes next? Project-06_1.mmpz?

@he29-net Correct, though I wouldn't consider this a breaking change.

Spekular:
No one is going to think projectname-06 means "projectname minus 6", just like I don't think projectname.mmpz is a decimal number (unlike 1.01 or 23.4, which also use periods).

Also, 06_1, 06_2, 06_3 seems like a texbook case of breaking the versioning scheme, considering the unbroken version is 07, 08, 09.

I hope this doesn't come off as hostile, but we shouldn't change the format unless the new one is so much better that it justifies either:

A: Messing up old files' consistency

B: Writing an upgrade routine

Given the fact that this discussion exists, I think it clearly isn't enough of an improvement to warrant the downsides.

tresf:
@Sawuare thanks for the thought that went into this, but this is the type of PR that needs conversation first and that's what Discord is for.

1. There's no upgrade path to previously named files as identified by @he29-net
2. The sorting is presumptuous.  Example:
   ```bash
   ls -al *.mmpz
   -rw-r--r--   1 owner  staff      0 Nov 14 23:33 test-1.mmpz
   -rw-r--r--   1 owner  staff      0 Nov 14 23:34 test-12.mmpz
   -rw-r--r--   1 owner  staff      0 Nov 14 23:34 test-2.mmpz
   ```

3. Underscores aren't really any easier to read, and certainly not easier to type.  Closing.

Please don't take this as discouragement and thanks for the fine attention to detail, the project needs it!

PR #5245, opened 2019-10-17
CircleCI: Suport macOS

### LINKED ISSUE #5228: Use CircleCI for macOS builds
Thanks to @tresf, we have access to macos builds on CircleCI. We should use it :)

Reflexe:
closes #5228

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,6 +12,18 @@ shared:
       key: ccache-{{ arch }}-{{ .Environment.CIRCLE_JOB }}-{{ .Branch }}-{{ .BuildNum }}
       paths:
         - ~/.ccache
+  restore_homebrew_cache: &restore_homebrew_cache
+    restore_cache:
+      keys:
+        - homebrew-{{ arch }}-{{ .Environment.CIRCLE_JOB }}-{{ .Branch }}
+        - homebrew-{{ arch }}-{{ .Environment.CIRCLE_JOB }}
+        - homebrew-{{ arch }}
+  save_homebrew_cache: &save_homebrew_cache
+    save_cache:
+      key: homebrew-{{ arch }}-{{ .Environment.CIRCLE_JOB }}-{{ .Branch }}-{{ .BuildNum }}
+      paths:
+        - ~/Library/Caches/Homebrew
+        - /usr/local/Homebrew
 
   ccache_stats: &ccache_stats
     run:
@@ -143,10 +155,53 @@ jobs:
       - run:
           name: Shellcheck
           command: shellcheck $(find "./cmake/" -type f -name '*.sh' -o -name "*.sh.in")
+  macos:
+    environment:
+      <<: *common_environment
+    macos:
+      xcode: "9.3.1"
+    steps:
+      - checkout
+      - *init
+      - *restore_homebrew_cache
+      - *restore_cache
+      - run:
+          name: Install Homebrew dependencies
+          command: brew update && brew install ccache fftw cmake pkg-config libogg libvorbis lame libsndfile libsamplerate jack sdl libgig libsoundio stk fluid-synth portaudio fltk qt5 carla
+      - run:
+          name: Install nodejs dependencies
+          command: npm install -g appdmg
+      - run:
+          name: Building
+          command: |
+            mkdir build && cd build
+            cmake .. -DCMAKE_INSTALL_PREFIX="../target" -DCMAKE_PREFIX_PATH="$(brew --prefix qt5)" $CMAKE_OPTS -DUSE_WERROR=OFF
+            make
+      - run:
+          name: Build tests
+          command: cd build && make tests
+      - run:
+          name: Run tests
+          command: build/tests/tests
+      - run:
+          name: Build DMG
+          command: |
+            cd build
+            make install
+            make dmg
+            cp ./lmms-*.dmg /tmp/artifacts/
+      - store_artifacts:
+          path: /tmp/artifacts/
+          destination: /
+      - *save_cache
+      - *save_homebrew_cache
+
+
 workflows:
   version: 2
   build-and-test:
     jobs:
+      - macos
       - mingw32
       - mingw64
       - linux.gcc


Reflexe:
After that PR, we need to drop Travis from the commit status.

.circleci/config.yml
```diff
@@ -34,7 +46,7 @@ shared:
   # Commmon environment variables
   common_environment: &common_environment
     QT5: True
-    CMAKE_OPTS: -DUSE_WERROR=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo -DUSE_CCACHE=ON
+    CMAKE_OPTS: -DCMAKE_BUILD_TYPE=RelWithDebInfo -DUSE_CCACHE=ON
```
PhysSong: I think removing `-DUSE_WERROR=ON` from Linux/Windows builds is not good. Can we leave this as-is and override it on the Mac section?

.circleci/config.yml
```diff
@@ -12,6 +12,18 @@ shared:
...
+      keys:
+        - homebrew-{{ arch }}-{{ .Environment.CIRCLE_JOB }}-{{ .Branch }}
+        - homebrew-{{ arch }}-{{ .Environment.CIRCLE_JOB }}
+        - homebrew-{{ arch }}
+  save_homebrew_cache: &save_homebrew_cache
+    save_cache:
+      key: homebrew-{{ arch }}-{{ .Environment.CIRCLE_JOB }}-{{ .Branch }}-{{ .BuildNum }}
+      paths:
+        - ~/Library/Caches/Homebrew
+        - /usr/local/Homebrew
```
PhysSong: Caching `~/Library/Caches/Homebrew` sounds good, but does `/usr/local/Homebrew` ever exist?

Reflexe:
This is where homebrew actually downloads its metadata data to. That
reduces setting up by about 5m.

tresf:
@reflexe, the DMG appears to be in working order.

Screen Shot 2019-10-22 at 1 45 32 PM
Screen Shot 2019-10-22 at 1 46 52 PM

.circleci/config.yml
Reflexe: This is where homebrew actually downloads its metadata data to. That
reduces setting up by about 5m.

.circleci/config.yml
Reflexe: You're right.

CMakeLists.txt
```diff
@@ -491,7 +491,7 @@ FIND_PACKAGE(Samplerate 0.1.8 MODULE REQUIRED)
 
 # set compiler flags
 IF(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
-	SET(WERROR_FLAGS "-Wall -Werror=unused-function -Wno-sign-compare -Wno-strict-overflow")
+	SET(WERROR_FLAGS "-Wall -Werror=unused-function -Wno-sign-compare -Wno-strict-overflow -Wdeprecated-register")
```
PhysSong: Clang emits a ton of errors than this one. Also, did you mean `-Wno-deprecated-register`?

PhysSong: Copying what I said on Discord:
> @Reflexe I'd rather add a `-DUSE_WERROR=OFF` to the macOS build.

CMakeLists.txt
```diff
@@ -493,7 +493,9 @@ FIND_PACKAGE(Samplerate 0.1.8 MODULE REQUIRED)
 IF(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
 	SET(WERROR_FLAGS "-Wall -Werror=unused-function -Wno-sign-compare -Wno-strict-overflow")
 	OPTION(USE_WERROR "Add -werror to the build flags. Stops the build on warnings" OFF)
-	IF(${USE_WERROR})
+
+	# Don't use -Werror with clang since it breaks a lot of stuff.
+	IF(${USE_WERROR} AND NOT CMAKE_CXX_COMPILER_ID EQUAL "Clang")
```
PhysSong: To handle AppleClang:
```suggestion
	IF(${USE_WERROR} AND NOT CMAKE_CXX_COMPILER_ID MATCHES "Clang")
```

Reflexe:
Is there any issue with this PR? can I merge it? 😕

PhysSong:
I can't find any serious issues with this PR. @tresf Could you look at the changes so we can double-check?

tresf:
I'm confused, what part of this build requires CMake changes?  Can't you just set `-DUSE_WERRO=False` on Mac builds?

PhysSong:
> Can't you just set `-DUSE_WERROR=False` on Mac builds?

Do you mean adding the flag after `$CMAKE_OPTS`?

tresf:
I mean setting it in Circle, not CMakeLists.txt

Reflexe:
Our werror options will cause clang to fail. Currently, in circleci, CMAKE_OPTS is a shared for all jobs and specify werror. Instead of changing it, I have decided to fix the root of the problem.

PhysSong:
@Reflexe I think what @tresf mean is:
```diff
-            cmake .. -DCMAKE_INSTALL_PREFIX="../target" -DCMAKE_PREFIX_PATH="$(brew --prefix qt5)" $CMAKE_OPTS
+            cmake .. -DCMAKE_INSTALL_PREFIX="../target" -DCMAKE_PREFIX_PATH="$(brew --prefix qt5)" $CMAKE_OPTS -DUSE_WERROR=OFF
```

Reflexe:
I think that repeating the same argument is a bit strange. Also, we have a problem with clang and werror.

tresf:
> I think that repeating the same argument is a bit strange.

Stranger than changing the CMakeLists (which affects all build environments) for the sole purpose of a CI configuration?

I don't much care as ideally, Clang will some day build with WERROR (@jasp00 actually made it happen, but we got into some disagreements about how to handle upstream build issues).

Anyway, I prefer to handle this environmentally, but won't fight it.

Reflexe:
> > I think that repeating the same argument is a bit strange.
> 
> Stranger than changing the CMakeLists (which affects all build environments) for the sole purpose of a CI configuration?
> 
> I don't much care as ideally, Clang will some day build with WERROR (@jasp00 actually made it happen, but we got into some disagreements about how to handle upstream build issues).
> 
> Anyway, I prefer to handle this environmentally, but won't fight it.

You are right :) I will change it tomorrow.

lukas-w:
What's the benefit of running macOS tests on CircleCI instead of Travis?

Reflexe:
> What's the benefit of running macOS tests on CircleCI instead of Travis?

Reliability and speed. Travis is just hell to work with.

tresf:
> Travis is just hell to work with.

This seems to be modern consensus, but Travis has been a good platform for us for years.

The way I see it is, Circle-CI offers consistency as its primary factor.  For example, the build artifacts would be the influencing factor.  Speed and reliability seem like empty promises.  Each CI will suffer it's own issues on both speed and reliability, both of which fluctuate and are out of our control.

tresf:
> Travis's macOS builds are very inconsistent.

I agree that they do fail to start more often than Linux, but once they start, the usually succeed.  I believe recent failures were due to running an old environment, building qt5 from scratch, because homebrew deliberately breaks older OSs (quite annoying) but would exist on any CI running a dated macOS version.

PR #5247, opened 2019-10-17
Move peak controller out of core

### LINKED ISSUE #5242: Move PeakController out of core
@BaraMGB's https://github.com/LMMS/lmms/pull/2944 got into a problem with that. Therefore, I have implemented a way to load PeakController dynamically from a plugin and completely moved PeakController from the core since core components should not depend on plugins.  

I'll open a PR for that in the next few days. 

closes #2056

Reflexe:
closes #5242


PhysSong:
I suggest something different: can we move the peak controller effect to LMMS itself?

PhysSong:
Closing for now, we should decide which one to move. This can be reopened once we decide this is the way to go.

Issue #2286, opened 2015-08-23
Piano roll toolbar is too wide


michaelgregorius:
The current piano roll toolbar is very wide and create problems on size constraint displays, e.g. 1280x1024 or smaller:
![pianoroll-wide](https://cloud.githubusercontent.com/assets/9293269/9429178/f34d4ce8-49c7-11e5-993a-ea1482cef2c1.png)
The individual groups should be broken down into separate toolbars with a two line default layout.

tresf:
https://github.com/LMMS/lmms/issues/387 :+1:

michaelgregorius:
The patch above changes the piano roll's toolbar to the following look:
![pianorollchanges-movabletoolsbars](https://cloud.githubusercontent.com/assets/9293269/9428756/25249618-49b9-11e5-9f12-4c9638e28cf1.png)
There are now five groups / tool bars: "Transport controls", "Note controls",
"Copy paste controls", "Timeline controls" and "Zoom and note controls"
(this group is a bit mixed). Each group can be turned off and on using
the standard toolbar context menu provided by Qt.

The new default layout of the toolbars saves horizontal space by putting
the "Zoom and note controls" below the other toolbars. The toolbars can
be moved but their positions are not saved. Floating toolbars are
disabled, i.e. the toolbars will always stay inside the window (top,
bottom, left, right).

tresf:
Moving the conversation here... 

> @tresf This is all done using stuff that's already available from Qt! If you look at the code changes you will see that it was a rather low hanging fruit. There is some not so nice code concerning `m_toolBar` from the `Editor` class because I had to override some of the defaults being taken in the `Editor` implementation. I don't think it's a good idea that every `Editor` automatically has some transport controls but for now this does not seem to create any problems (except some headscratching when reading some of the code :wink:).

Wishful thinking... should we just fix the parent Editor once and for all?

-Tres

michaelgregorius:
@tresf I had a closer look at the `Editor` class. It seems that it's only concerned with transport functionality and providing the basis for toolbars by inheriting from `QMainWindow`. I thought that it would provide a bit more functionality which might be useful to other editors that might not need transport capabilities. So if we'd remove the transport toolbar we could more or less remove the whole class.

But whatever the decision is I think all these things should be handled in a new issue anyway because it would not only concern the piano roll.

Umcaruje:
Very cool looking! 

>The toolbars can be moved but their positions are not saved

[...]

Issue #387, opened 2014-02-27
Toolbar vertical real-estate enhancement


tresf:
A proposal for the toolbar to reclaim some unused screen real-estate.

![image](https://f.cloud.github.com/assets/6345473/2278896/a999bda2-9f5f-11e3-9edc-91ca69256b2b.png)

---

**Left aligned tool menu items:**
_Make the file and project toolbar buttons a single row that is capable of overflowing into a second row if needed._
![image](https://f.cloud.github.com/assets/6345473/2278954/f26004dc-9f60-11e3-8fea-8a281d599cd9.png)

---

**Right aligned tool menu items:**
_Make the remaining components a single row that is capable of overflowing into a second row if needed._
1.  All time-related items in a single row (BPM, time sig, clock)
2.  Master volume/master pitch rotated horiz (single row height)
3.  Optionally consolidate min/sec/millisec into a single "clock"
4.  Visualization shorter (single row height)
5.  CPU vertical, shorter (single row height)
![image](https://f.cloud.github.com/assets/6345473/2279079/c57a438a-9f63-11e3-8ed9-eabef78bf79a.png)

---

**File menu hidden**
_All menu items consolidated into a single quick-access menu.  This is_ **no less clicks** _than the current menu and finds a better place for single-item menus such as LADSPA or Settings.  Most items already have dedicated toolbar buttons, encourages their usage.  Chrome menu button used only as example._
![image](https://f.cloud.github.com/assets/6345473/2278948/d8055bfa-9f60-11e3-88d7-450e10629c03.png)

---

**The big picture:**
_All changes combined into a single mock-up.  Notice the master volume is frozen to the top right, for quick accessibility._
![image](https://f.cloud.github.com/assets/6345473/2279081/d484de30-9f63-11e3-97dc-9d7bb099ad1c.png)

---

**Additional Notes**

I would personally recommend removing the CPU and visualization from the default view and instead offer these as FX visualization plugins.  Alternately, a basic CPU/MEM indicator might be a good alternative.

-Tres

RustoMCSpit:
is this still in the works, it's been tagged so many times but hasnt been updated in nine years

tresf:
Its just a feature request, so anyone with basic programming experience can try to tackle it, but as far as I know it was never actually started.

PR #5261, opened 2019-10-20
Modernize the toolbar

Lathigos:
This pull request updates the main toolbar to be much more modern, and moves the looks of LMMS more towards @budislav's design (see #1911).

new_toolbar_design_002

There are some key changes from that design:

- There are sliders instead of small spinboxes, which I deem more suitable for the type of controls they handle (a small spinbox for the master volume is a bit cumbersome).
- The master volume out visualizer and CPU load visualizer are swapped. This is so that the master volume out visualizer is next to the master volume slider, which I think is a more intuitive position rather than next to the pitch slider.

This is what it looks like in a full window. Note that the widget is dynamically positioned to the center of the screen:
new_toolbar_design_001


Reflexe:
Very nice work! I hope to review it in the next week or so. Thank!

Gabrielxd195:
Hi, I just saw this project and it's great! but where would you put the oscilloscope? I made a montage in gimp to give you a suggestion.

Osciloscopio de lmms

I suggest that you place the marker next to the buttons to take advantage of space and to make it more pleasing to the eye, also to place the oscilloscope, while the empty space on the right can be reserved for other future projects. regards

Lathigos:
@Sawuare Ah, you're right! I'll update the file descriptions and copyright information. What are the conventions for adding in developer names and such?

@Gabrielxd195 I'm not sure we should change it back to that layout, as it would ultimately look cluttered again which is exactly what I wanted to avoid. Note that my implementation has been based on budislav's design (see #1911) — I think we should adhere to that unless we have a good reason not to. Any additional widgets shouldn't be added to the right of the widget, but be properly incorporated in an _intuitive_ position in the toolbar. Doing otherwise would be bad user experience. For that reason, I'll probably eventually move those old buttons as well, but that's not for this pull request.

For the same reason, I think we also shouldn't add back the oscilloscope, since it serves no real function in the toolbar other than to take up space. The only reasons to use that oscilloscope is to view the volume output levels (which the 'OUT' widget in the upper right corner does much more efficiently), since the output was way too stuttery (and the widget too small) to provide any real information about the waveform. The toolbar just isn't the right place for this oscilloscope.

@CaBachhav Interesting idea. I'll have to see how well it'd fit in the window, since I would want to avoid wasting too much real-estate when laying it out horizontally. I think this would have to be discussed with the design folks. Perhaps it can be designed as an additional mode for the clock widget in which the beat/bar and min/sec are stacked vertically, but then the numbers would have to be smaller.

Lathigos:
Oh, general question: I'm not too familiar with the travis-ci setup over here. Does anyone know why it's giving an error?

Reflexe:
You can ignore it. We are in the process of getting rid of it.

qnebra:
Works nicely for me, everything was fine as it should be in terms of usability. 

But, but. Those buttons on left side looks quite "old" with new toolbar mechanics and design, they are cramped in small space.

Lathigos:
@qnebra Yeah, I totally agree. I haven't touched the buttons yet because there are too many of them — and in the single-window design at least half of them would disappear anyway — but perhaps I could change some things to make it look better. I'm currently working on a mockup .png to show a possible solution, but it's a bit problematic with the amount of buttons there are.

Perhaps some buttons could be removed to make the toolbar less cluttered? Most of the top buttons have their behaviour mirrored in the menubar anyway, and some of the bottom buttons as well (namely 'show piano roll', 'show BB editor', and 'show automation editor' can all be activated by double-clicking the respective track in the Song Editor, which I think is more intuitive anyway).

So perhaps we should reconsider what buttons we really want to keep as _quickbuttons_ in the toolbar. That is, what actions _really_ warrant being able to do it in one click as opposed to two? Things such as creating a new template-based file is not something I do every five minutes in LMMS, mostly just once per session. Those stuff could be removed as quickbuttons and just remain in the menubar instead.

Sawuare:
@Lathigos The copyright notice here is generally of the form:
Copyright (c) YEAR NAME <EMAIL>

Spekular:
If all the file operations were removed (since they're accessible in the
File menu, and some via shortcuts), the remaining top row buttons would
probably fit fine on the bottom row. This could let us reduce the height of
the toolbar, if desired.

Umcaruje:
Hi @Lathigos, nice work! Here are some of my observations:
- I still feel like the sliders use too much space and don't quite fit. I would suggest actually implementing the original design, which is the text based widgets. I consulted with people on the discord server and they do like the idea.

- > Note that the widget is dynamically positioned to the center of the screen:

  I don't like this because it creates a really large gap between the buttons and the as your screen gets wider: 
image

  I would suggest having a fixed gap between the widget and the buttons (everything left aligned on the screen)

- I agree with @mxmilkb 's observations from the other issue, the zeroes have no place in the time signature, it just looks confusing. e.g. 
  image vs image

  The 4 looks a bit misplaced with the gap, so the ideal behaviour would be to have it like this:
image

  and turn into this once you get into double digits
image

  But if that's too much work just removing zeroes would work fine in my opinion.

- I think these widgets should be edited like we do TrackLabelButtons, that is they behave like a text box on double click rather than launching a dialog, but this might be out of scope for this PR.

- The out meter doesn't behave like the FX meters, if you play a low sine you can see the out meter fluctuating while the FX master stays the same (which imo is the correct behaviour)
record

qnebra:
> @qnebra Yeah, I totally agree. I haven't touched the buttons yet because there are too many of them — and in the single-window design at least half of them would disappear anyway — but perhaps I could change some things to make it look better. I'm currently working on a mockup .png to show a possible solution, but it's a bit problematic with the amount of buttons there are.
> 
> Perhaps some buttons could be removed to make the toolbar less cluttered? Most of the top buttons have their behaviour mirrored in the menubar anyway, and some of the bottom buttons as well (namely 'show piano roll', 'show BB editor', and 'show automation editor' can all be activated by double-clicking the respective track in the Song Editor, which I think is more intuitive anyway).
> 
> So perhaps we should reconsider what buttons we really want to keep as _quickbuttons_ in the toolbar. That is, what actions _really_ warrant being able to do it in one click as opposed to two? Things such as creating a new template-based file is not something I do every five minutes in LMMS, mostly just once per session. Those stuff could be removed as quickbuttons and just remain in the menubar instead.

Create from template was rarely used I think. 

Also I made some small amateur mockup, with how I will see this in lmms. Also, also, I was kinda forgot about making toolbar thinner by 7 px (that was a lot of space ;) )

lmmstoolbarmockup

Lathigos:
@qnebra I was thinking of something like this, perhaps (just a quick mockup):
mockup_new_buttons_004

The file operations and stuff would be on the left hand side of the widget as normal buttons, and the window visibility buttons would be flushed to the right side of the window. This would also help distinguish their different functionalities.

Of course, we could then indeed remove the margin at the top and bottom of the widget to make the entire toolbar more thin! :smiley:

Perhaps in the final design I can move the metronome button into the central widget as opposed to next to it, more like the original design. And perhaps I should make all buttons the same size, I don't know.

@Umcaruje 
> I agree with @mxmilkb 's observations from the other issue, the zeroes have no place in the time signature, it just looks confusing.

> The out meter doesn't behave like the FX meters, if you play a low sine you can see the out meter fluctuating while the FX master stays the same (which imo is the correct behaviour)

> I still feel like the sliders use too much space and don't quite fit.

Alright, many thanks for the feedback, I'll work on them!

>I think these widgets should be edited like we do TrackLabelButtons, that is they behave like a text box on double click rather than launching a dialog, but this might be out of scope for this PR.

That'd be a great addition! If I find the time, I'll work on that as well.

>I don't like this because it creates a really large gap between the buttons and the as your screen gets wider:

This is the only point which I fundamentally disagree about. I understand that everybody is used to having a left-flushed layout in a toolbar (it took me some time to get used to it as well), but it's an outdated design choice. Once we move towards a single-window design, the centered widget looks much more modern, clean, and above all _intuitive_ — as opposed to cluttered and confusing. Besides, the unused space on the toolbar is wasted space either way.

So I'll experiment around with it, but I'm not sure I like the idea unless there's some really compelling reason to do it.

RebeccaDeField:
Great work here! I have a few ideas regarding the colors and contrast. Will comment with some hexes as soon as I get the chance.

qnebra:
> @qnebra I was thinking of something like this, perhaps (just a quick mockup):
> mockup_new_buttons_004
> 
> The file operations and stuff would be on the left hand side of the widget, and the window visibility buttons would be flushed to the right side of the window. This would also help distinguish their different functionalities.

Great idea for me, currently this left side area was really crowded, and right side feels really empty.

> Of course, we could then indeed remove the margin at the top and bottom of the widget to make the entire toolbar more thin! 😃

And maybe flat color? But this gradient in your mockup looks really good.

RebeccaDeField:
I also wanted to chime in on the matter of scaling that I agree on the center alignment will work on large monitors now that the idea is to put the icons on either side.

If you are proposing to remove the oscilloscope, do we have an idea of where it may be placed or accessed instead? Can we say that this is universally unused and pointless or could this be something which users would find missing?

We need to consider this design with the smallest screen scenario as well as the largest, especially if the point is to go in the direction of scalable UI.

Lathigos:
@Umcaruje Alright, I finally got around to coding a little bit again. I started experimenting with removing the zeroes from the time sig widget. It looks good when there's a single digit on both sides, as well as when there's two digits at both sides. But then I noticed this annoying case, a single digit on one side and two digits on the other:

asymmetric_metric

asymmetric_metric2

Now the time sig widget doesn't look centered at all. Of course, one could center the entire widget, but then the "/" would start shaking around as you scroll through the spinboxes and that looks really awful.

So I propose to keep the zeroes in the time sig widget.

Sawuare:
The oscilloscope is useful as it can detect clipping & DC bias and examine timbre & stereo separation, so do not remove it.

RoxasKH:
I think that if it's really needed to remove the oscilloscope, it should become a lmms native plugin

RebeccaDeField:
Is there a reason that the oscilloscope could be added to the centerpiece along with the volume controls now that we have found a better layout?

Otherwise, if it can remain as a separate plugin, that may work. I think it is best to not remove it entirely in any case.

RebeccaDeField:
Here's a rough mockup of what direction I think we should take for the colors which gives some nice contrast improvements and matches the single-window design a bit closer:

Rough mockup:
Toolbar2
Single window:
ToolbarIdea

@umcaruje thoughts?

Note: For whatever reason, my inkscape HSL values are not reflecting the right numbers so I will give hex codes after that is resolved on my end for good measure.

RebeccaDeField:
The time/tempo toolbar from my mockup should have a 220 hue and 10 saturation (inkscape goes up to 255 for all values). Not sure why inkscape will not display this with consistency but in case I'm holding things up, I tried some online apps to get hopefully better hex numbers for the mockup I proposed. I would also like to make the toolbar/menu which this part is inset into have a more subtle gradient but I imagine this is something I can do in another PR to keep things organized.

Top stop: `#202327`
Bottom stop: `#191b1f`
Border: `#151719`

Lathigos:
I updated the images in the description to reflect the new work I have done.

@Umcaruje 
> I still feel like the sliders use too much space and don't quite fit.

I have implemented the design you suggested. I haven't yet pushed the code, because I still feel like the text-based widgets are unintuitive. In fact, I have shown several people both designs (see below) and asked which one they thought was more intuitive (while not showing my own opinion to prevent bias). All of them immediately responded that the sliders are more intuitive, even when I tried to convince them of the other design.

option1
option2

This affirms the ideas of good ergonomics design: consistency is essential. Volume is already controlled using sliders in the FX Channels (as well as in every major operating system), so that's what the user has come to expect. The only reason not to use sliders is if it wouldn't fit, but not much real-estate is saved by changing the sliders anyway.

I could probably still change the icons to be text instead ('VOL' and 'PITCH'), since they are probably the features that stick out most, but I think it's best to let the actual controls be sliders.

Other than that, I think I am mostly done with all requested features (except for the oscilloscope). In both designs, there's no room for the oscilloscope in the toolbar in the smallest-screen scenario, so should I re-add it as a native plugin instead?

Umcaruje:
The design without the faders looks a lot more consistent in my opinion.

> I still feel like the text-based widgets are unintuitive. In fact, I have shown several people both designs (see below) and asked which one they thought was more intuitive

Were those people LMMS users? The thing about these controls is that they are rarely used. The master volume is absolutely redundant, because we already have the master FX slider. And the global pitch is much better represented by a number, because if you're transposing your whole project you'd want to know at all times how many semitones you have transposed.

Before I suggested removing the sliders, I talked with a couple of people on the discord server and they said they wouldn't mind removing the sliders as long as the text widgets are editable.

> Other than that, I think I am mostly done with all requested features (except for the oscilloscope). In both designs, there's no room for the oscilloscope in the toolbar in the smallest-screen scenario, so should I re-add it as a native plugin instead?

Yes, that would be a good idea. 

Also I really like the moving of the buttons to the right and left side, it makes the centered widget really work and look consistent. Great job!

Lathigos:
@Umcaruje
> Were those people LMMS users?

Some of them were, actually, but others used different DAWs.

> The master volume is absolutely redundant, because we already have the master FX slider. And the global pitch is much better represented by a number, because if you're transposing your whole project you'd want to know at all times how many semitones you have transposed.

This is a fair point. I agree the semitones are indeed better represented by a number. I didn't consider the fact that the master FX slider is redundant anyway, so it being intuitive is less of a point. The reason I brought it up again was because of the responses I got from the people I consulted, but I suppose it could go either way. Anyway, thanks for the feedback! I'll push the changes. :smiley:

Umcaruje:
I also had a quick poll with the users over on discord: 
image
There were some concerns on how effective the dragging will be but if the UX was right, all votes were on removing the sliders.

> Anyway, thanks for the feedback! I'll push the changes. 

Thank you for working on this! I'll try to test this all out asap.

Umcaruje:
>Alright, I finally got around to coding a little bit again. I started experimenting with removing the zeroes from the time sig widget. It looks good when there's a single digit on both sides, as well as when there's two digits at both sides. 

>But then I noticed this annoying case, a single digit on one side and two digits on the other. Now the time sig widget doesn't look centered at all. Of course, one could center the entire widget, but then the "/" would start shaking around as you scroll through the spinboxes and that looks really awful. 

>So I propose to keep the zeroes in the time sig widget.

@Lathigos I just noticed this. I disagree on that it would be better to have an uncentered widget than one with zeroes before the numbers. I would ask to also remove the zeroes before numbers in the tempo widget. These look really weird in my opinion and wouldn't fit in such a modern design.

RebeccaDeField:
Just wanted to check in on this. I think even with the 0 to start the time sig, it would be a major enough improvement because of the clarity added to this part of the UI. I wonder if removing the 0 could be another PR improvement for later? If there are any icon improvement needs to finish this, I am also on board to help.

> The oscilloscope could be added to the centerpiece along with the volume controls now that we have found a better layout?

There were some concerns about removing the oscilloscope. I believe now that the design was condensed, it should be okay to keep it on the toolbar until someone has time to separate it out as a native plugin.

Even with the oscilloscope, cleaning up the toolbar to this degree would be a big step in the right direction.

qnebra:
I decided to use lmms with this PR in daily music production, just for testing.

1. Still wasn't clear which window is active. Maybe green tint under icon of active editor?
2. Maybe file butttons and editor buttons should be closer to main "central" toolbar pieces, to avoid unnecesary running across toolbar. Completly optional, and maybe was completly wrong.

Spekular:
I suppose it's late to bring this up, but the placement of the window buttons on the far right is a bit concerning to me. Most important features can be accessed from the left hand side of the UI (menu bar, sidebar, transport controls and other menu items in editors). Having to move the mouse all the way up to the top right every time I want to open an editor seems like a (minor but very common) productivity loss. Maybe even more physically stressing in the long run.

The buttons currently on the left, on the other hand, are buttons I've never once used in years of using LMMS. I've always used the file menu or hotkeys instead. Perhaps others use them, but even so actions like saving/loading projects should be less common than opening/closing editors. Finally, the file menu is already on the left, so even if the buttons are moved these actions can be accessed from that side of the screen. For that reason I believe the editor buttons deserve the prime location more than the file buttons do.

MetaDark:
Would it maybe better to swap the positions of the spacers so all the controls are grouped together in the center?

Screenshot from 2020-05-19 05:53:30

Lathigos:
@MetaDark @Spekular @qnebra I put the buttons closer to the center piece. I also removed the leading zeroes of the tempo and time sig widgets.

LMMS-Toolbar-Redesign-001

Since I will likely not be able to add a custom oscilloscope plugin in this pull request, I was wondering what position would work best for the oscilloscope within the toolbar. Does anyone have an opinion on that? I have to admit there is very little space left for such a large widget in the toolbar, because I would like to keep it small enough so that the window can still snap to half a 1920x1080 screen (so the maximum width of the window should be at most 960 pixels, which it currently already is). Perhaps I can add a toggle switch in 'Settings' to show or hide the oscilloscope in the toolbar.

Spekular:
IMHO this isn't an improvement. The distance to the editor buttons is barely decreased and the layout of [left aligned - center - right aligned] was nicer to me.

The solution to my complaint, IMO, is to swap places so the editor buttons are on the left and file ops on the right. However, I would love to know what others think of this in case I'm in the minority on this. Perhaps a reaction vote is in order

:+1: prefer editor buttons on the left
:-1: prefer editor buttons on the right
:confused: no preference

RoxasKH:
Why the button on the right should be smaller than the ones on the left?

Immagine

RoxasKH:
Anyway, watching other DAWs, a lot of them has most of the things on the left, like Reaper, Ardour, or Bitwig, but other ones has things on the right, too, like Logic.

In my opinion if someone finds uncomfortable to have those buttons on the right is only because he's used to have them on the left, like LMMS is at the moment.

Spekular:
> Anyway, watching other DAWs, a lot of them has most of the things on the left, like Reaper, Ardour, or Bitwig, but other ones has things on the right, too, like Logic.

Not all buttons are equal. You'll notice the most important controls (transport controls e.g. play/pause) are still on the left in Logoc Pro X.

> In my opinion if someone finds uncomfortable to have those buttons on the right is only because he's used to have them on the left, like LMMS is at the moment.

Given that I explained my reasoning, I think it's silly to imply this is "only because" of habit.

RoxasKH:
> 
> 
> > Anyway, watching other DAWs, a lot of them has most of the things on the left, like Reaper, Ardour, or Bitwig, but other ones has things on the right, too, like Logic.
> 
> Not all buttons are equal. You'll notice the most important controls (transport controls e.g. play/pause) are still on the left in Logoc Pro X.
> 
> > In my opinion if someone finds uncomfortable to have those buttons on the right is only because he's used to have them on the left, like LMMS is at the moment.
> 
> Given that I explained my reasoning, I think it's silly to imply this is "only because" of habit.

I know, but i think habit has its importance in this, too.

Yeah, speaking of Logic, you're right, we have to find a compromise on what's the most comfortable to use and good to see at the same time.
Having everything on the left could be good, as i said reaper keep all his commands on the left. But it should also looks good

qnebra:
Or merge toolbars of editors with main toolbar. But this is more as silly and dumb idea, not as thing to do. Also it will possibly make main toolbar really crowded in small screen resolutions.

tresf:
Using garageband as an example, having left-floating and right-floating items is a good design decision.  Any proposals which take this away from @Lathigos' original proposal I find to be a regression (except the re-addition of the master volume slider, which is obviously needed!)

ezgif-2-07ca9b9d756d




In regards to where they should go, @Spekular linked this bug report in our #testing channel on Discord channel (http://lmms.io/chat), but I'm not certain what I'm voting on.  Personally, I'll adapt to wherever things are.  I believe the visibility toggling is a workflow created by MDI windowing which should eventually go away when single-window is completed.

tresf:
Unpopular opinion, but I think master-pitch should be hidden by default.   We can ditch the volume artwork next to the master slider if we ditch the master-pitch from this toolbar.  (I love the master-pitch, so I say this without a clear proposal/solution as to where it should live afterward).

GrimmZane:
lmms_ux-modernize-toolbar
Some feedback regarding UX, generally frowned on making users go an extra distance for any commonly used or associated UI elements.  With a lazy relocation for illustration.


I'd personally recommend going for a slimmer toolbar, ribbon UI is mocked for good reason, and having any wasted/negative space in software that can make full use of pixel real estate, seems borderline criminal.
lmms_ux-modernize-toolbar_slim

Separated the project-notes and tempo buttons as they seem separate enough.  Left 6 are explicitly file operation related, 6 on the right are production UI element invoking.  Metronome being close to time/tempo makes sense.  The project-notes is both UI invoking but not directly production related like the other 6. The spacing and justifying of those icons could probably be improved, i.e. metronome relative to the time/tempo region.

Button separation is less about just organization and more about making it easy to associate areas of the UI with their relevant functions, instead of densely packing them so extra time is needed to dis-cypher icons. 

speaking of, project notes might benefit from being changed from the page icon as it's already heavily associated with file ops, to perhaps an i in a circle, for information? don't feel strongly about that suggestion tho, but putting it in with the file ops icons, does cause identification issues imo.
info

You could always have 2 layouts that can be toggled in settings, like the 'compact track buttons'. if there were any strong objections to the slim toolbar.

tresf:
Removing pitch-slider mockup:

**Before:**
image

**After**
vol

tresf:
@Spekular thanks for clarifying the vote of the buttons.  Most apps I use on the daily don't have dedicated save buttons, so I'd be a fan of just removing the file operations altogether.  Then the windowing controls could go there and no workflow is broken.

Audacity as an example: 
image

GrimmZane:
Edit of the slim toolbar suggestion with removal of the file op buttons.
lmms_ux-modernize-toolbar_slim_goodbye-file-ops

centered also keeps it to the right of the expanded side column panel.  Which is nice.

I don't think anchoring it to the sides, garage band style, makes that much sense.  Due to the segregation of buttons to sub windows, so it's not as cluttered as other tools (which I think is a good thing)..  I think a more centered alignment does make more sense regarding LMMS's existing UI paradigm and workflow.

if you wanted to duplicate core functionality buttons from sub windows that'd need a lot more work and seems unnecessary.

RoxasKH:
> 
> 
> Edit of the slim toolbar suggestion with removal of the file op buttons.
> lmms_ux-modernize-toolbar_slim_goodbye-file-ops
> 
> centered also keeps it to the right of the expanded side column panel. Which is nice.
> 
> I don't think anchoring it to the sides, garage band style, makes that much sense. Due to the segregation of buttons to sub windows, so it's not as cluttered as other tools (which I think is a good thing).. I think a more centered alignment does make more sense regarding LMMS's existing UI paradigm and workflow.
> 
> if you wanted to duplicate core functionality buttons from sub windows that'd need more work and seems unnecessary.

Don't know, this seems so far from modern to me, which is what the pr is intended to do, and really different from budislav mockup we're trying to reach.
Like, why separating everything with a box?

RoxasKH:
> 
> 
> @Spekular thanks for clarifying the vote of the buttons (on Discord). Most apps I use on the daily don't have dedicated save buttons, so I'd be a fan of just removing the file operations altogether. Then the windowing controls could go there and no workflow is broken.
> 
> Audacity as an example (no save/open/buttons)
> image

I personally use the dedicated buttons a lot, but if most of lmms user don't, i think i can get used to using the menu.

tresf:
> Don't know, this seems so far from modern to me, which is what the pr is intended to do, and really different from budislav mockup we're trying to reach.

The theme can always change.  There are several steps to modernizing it and any PR which moves stuff will be subject to a lot of scrutiny and opinion, which we should be careful not to share un-objectively, it can lead to never-ending opinionated arguments, which result in never-merged items.

I'd argue that feedback like @RoxasKH saying "I use them but I can learn to live without them" is a feedback that's valuable.  Modern is subjective and there's a lot of leeway in that definition, but I'm doing my best to provide examples of well-known, commonly used apps and how they do it in 2020.  Most of our layout is identical to the way it was when LMMS was original released.  Moving to a modern interface requires accepting change, and it's not a democracy.  The deciding people will be a combination decisions by LMMS developers with any final decisions in @RebeccaDeField's.  Voting is good for feedback, but design is not a democratic process.

GrimmZane:
Harsh, but okay, my main suggestion was layout and UX oriented. having UI elements on the far edges (especially when the center is free), is objectively bad UX, it was already addressed by multiple suggestions and including the updated layouts by the PR author.  When the original proposed layout is maximized to 1080p or 1440p it's absolutely terrible distance for cursor travel. 

I am however genuinely asking, what justification is there for toolbar height? it doesn't add to usability or readability once you decide to move away from the double height button layout, it's fairly unjustifiable.  Going slim adds the entire height of a song editor track to the workspace.  Even @ 1080p that's a very worthwhile addition of space.  Most modern tools with hefty toolbars, don't also have loads of embedded toolbars in sub windows within the program doing the heavy lifting and taking up space already.  Opting for the former while also having the latter seems a bit crazy honestly.

Here's an edit just to revert the needless experimenting, and an attempt at incorporating the slider more seamlessly, personally I don't mind the text volume/pitch.

lmms_ux-modernize-toolbar_slim_goodbye-file-ops

Also how much is/will be customized/themed and how much hard coded? Guess that modernization is outside of the scope of this PR.  Also If pursuing a slim toolbar is opted against in the scope of this PR, maybe one can be opened after this is merged.  So an optional toggle might be added as I feel there is value in it especially for lower resolution devices :).  Especially the poor folks with those annoying sub 1080p laptop screens. 

Either way, responding to the discussion of design not being a democracy etc, I'm well aware of the potential null impact my edits and suggestions will have, and have and had no intention of trying to hijack the PR, just offering suggestions and feedback.

tresf:
> Harsh, but okay, my main suggestion was layout and UX oriented. having UI elements on the far edges (especially when the center is free), is objectively bad UX

This is your opinion and you've already made this point.  The direction we're driving towards is here: https://github.com/LMMS/lmms/issues/1911.   Assuming we keep this, the window toggles will eventually be replaced with a simpler single-window toggle, but that's probably not doable as long as we're using MDI (floating) windows.

> and including the updated layouts by the PR author.

Perhaps, but the original screenshots look closer to what we're aiming towards, not these ones with toolbar buttons near the playback.

> what justification is there for toolbar height?

Which height are you talking about?  I don't recall this conversation, please provide a link if it's easier.

> how much is/will be customizable/themed and how much hard coded?

We've added some customization in the Piano Roll editor: https://github.com/LMMS/lmms/issues/2286 and it was mostly well-received but too much customization can (and does) lead to some bad experiences (this is what I consider actual UX, not nitpicking mouse-move distances).  Anyone that's worked tech support for an app that allows it to hide and rearrange toolbars knows that it often is a pain point.   The most commonly used components should be easily accessible and in a predictable location.  Where it sits and whether it's aligned here or there is a design choice.  People really don't mind moving their mouse across the screen.  Use Gimp or Photoshop for 5 minutes and watch how often you move from toolbar to layers the full length of the screen.  It's not really worth arguing over if the buttons are accessible and predictably located.

>  pursuing a slim toolbar is opted against

I don't think anyone's against [maximizing real estate](https://github.com/LMMS/lmms/issues/387) 😉 just a matter of adding it in at the right time.

RebeccaDeField:
Just commenting to hopefully help move this forward. So without further ado, a couple of thoughts:

Most apps and software look modern through the results of placement according to what is perceived to give the best user experience for their particular user base. We can polish any idea into looking sleek and nice, but that's only an imitation of "modern-looking" so let's focus on these foundational changes first and worry about perfect aesthetics later.

This PR is intended to reduce the clutter of the toolbar, something Budislav's mockup is a good reference for. A cluttered toolbar doesn't just look bad, it's hard to use. Pretty much every mockup proposed here is a massive improvement from a UI perspective :+1: 

We're just down to the details now, so let's not lose sight of that progress with bikeshedding :bike: :heart: 

There are some claims here about separating the sections on either side as being objectively unusable. We're actually in a pros and cons situation. Good UIX is a balance of a lot of things with no one singular golden rule. I don't see a point in writing up the pros and cons further between these ideas, as it's been pretty much fully expanded on already.

The great thing about working via PR is, we can start with one big step to improve the toolbar, and it doesn't have to end there. If the options we pick doesn't work, we'll just open another PR to improve it. Design by committee doesn't usually produce the best design. Taking a step back after using and testing a merged PR tends to go a whole lot better. 

**Options according to what I've read here:**

1. Remove file operation buttons, align show/hide buttons to the left?
2. Left align all of the sidebar contents
3. Center align all of the sidebar contents
4. Align button sections to the opposite ends of the toolbar

I don't think I've seen any mockups for what it would look like with the file operations buttons removed, and since it seems like most people are open-minded to it, I think we should see how that looks so that we can properly compare and come to a decision on this.

MetaDark:
@RebeccaDeField I completely agree, we should stick as close to Budislav's mockup as much as possible. It looks like @GrimmZane already posted a mockup with the file operations removed, but I think adjusting the toolbar height deviates too far from the original design, even though I can see the practical use for it. are a bunch of combinations of layouts I tried by repositioning the buttons & spacers:

**Left Aligned**
Screenshot from 2020-05-23 10:25:58
**Left Aligned Buttons, Centered Controls**
Screenshot from 2020-05-23 10:45:53
**Left & Right Aligned Buttons, Centered Controls**
Screenshot from 2020-05-23 10:44:49
**Lopsided Buttons, Controls Centered**
Screenshot from 2020-05-23 10:25:33
**Balanced Buttons, Controls Centered**
Screenshot from 2020-05-23 10:26:18

Spekular:
I'm a big fan of left aligned buttons, centered controls. If I had veto power I would use it against "balanced" buttons which I find even worse than the original version, we *absolutely should not* split the editor buttons into two locations.

GrimmZane:
I can't see any evident UX issues in Budislav's redesign. Quite ambitious :) I should point out, pushing actively used icons to opposite edges of the screen is not reflected in his designs at all.  The layout-mode dropdown, is to the far side.  And not something you'd invoke regularly, so is very reasonable. If that was the layout i'd be content.

More or less agree with everything RebeccaDeField said.

> objectively unusable

i'd definitely not go that far, poor and irksome though for sure, especially on Ultra Wides,  For someone who has RSI issues personally i would strongly advocate against putting any relatively frequently used element far away from active workspace area. I'd wager it would bother many users.

if an element is barely used or has all it's related activity in the same region (like layer operations, or brush editing etc in the Image editing examples used above) I agree distance becomes more or less a non issue, but for something frequently used, that's not adjacent to related UI elements, it is a major UX concern.
 
. 
 
The layout edits I've done have mostly been offering mild suggestions, and besides the location of more frequently invoked button distance, have no real strong feelings towards the overall layout, and hopeful it'll be an improvement.

>Which height are you talking about? I don't recall this conversation

Where I do have strong convictions is reducing the height, having a tall toolbar has no real practical arguments, once you remove double stacked icons, there is no practical justification for it. 

This is my mockup suggestion, in a comparison to convey it's merits more explicitly.
lmms_ux-modernize-toolbar_slim_comparison

a reduction of 30+ pixels is a meaningful recovery of space for the working area. 

And it makes some justifiable sense in a single window layout like Budislav's design, but when every floating window has it's own height additions, it just compounds space issues. Matching that element of his design seems a tad silly in that context.

Again I'd be willing to open a PR for a slim toolbar, and if possible make it a 'restart required' toggle option. If keeping the toolbar padding is decided.

A more radical suggestion would be to include an option to hide the menubar and have it hide/show via tapping Alt like it does in some modern browsers & apps. Or via a hamburger dropdown.  That way keeping the save buttons will be more justified, of course you'd default the menubar to show as to not befuddle new users.

After some LMMS sessions paying closer attention to my UI usage. If you really wanted to declutter the toolbar, there are already 2 methods to invoke each removed editor window, from within the song editor. double clicking on a track pattern, or right click 'open in X'. So you could reduce the buttons down to this and only lose redundancy. Once doing this it illustrates that Mixer and Controller are the more critical buttons yo keep until you have any major overhauls i.e. nested mixer.

lmms_ux-modernize-toolbar_slim_declutter

GrimmZane:
Oh nice, def lean towards 
**Left Aligned Buttons, Centered Controls** 
Or 
**Lopsided Buttons, Centered Controls**

qnebra:
Well, @MetaDark you forgot about one thing:

1. Your concepts are very, very empty in terms of usability.
2. MDI must be gone.
3. My idea was to move most common used operations like "play", "pause", "stop", "record", "add track" etc to left, and window switches and less common operations to right. This will of course slightly increase main toolbar height due to avoid being really crowded.
4. Why? Because we read from left to right, and have tendency to search for things on the left side.

MetaDark:
@qnebra I definitely think we should move away from the MDI layout, but that will require significantly more effort than what's done in this pull request. I think right now is a good incremental step, and we can deal with other UI features later. I like the idea of having the play, pause, stop, record buttons in the toolbar, but then you lose the ability to only play what's in the editor.

GrimmZane:
A few more reasons towards why removing some of those hide/show window buttons is a valid consideration;

They only function as show/hide.  Choosing an active pattern explicitly opens them and is the only way to change between patterns so not only do you already need to interact with the song editor regardless, doing so already fulfills the 'show' functionality. and the 'X' in the window corner acts as the hide. 

And users universally already need to know and likely predominantly use the song editor to launch those windows, having a show/hide seems to have limited practicality.

And there is not much if any redundancy in Budislav's redesign. So moving away from them seems like the goal?

> but then you lose the ability to only play what's in the specific editor.
That's covered by the part/song buttons in his design
PART-SONG

MetaDark:
@GrimmZane Oh right good point, is this something that should be done in this pull request though? It would still involve modifying all the editors.

qnebra:
In my mind those "global control buttons" will be by default used by current active editor. This mean if "song editor" is active, which was indicated by green "song editor" button background, then they are used by song editor. Of course editor windows will be had switch between "Part" and "Song" as usually user (I guess) want first to play what was recorded, and then how this sounds in context of overall song.

And I agree, those changes which we all discuss happily in this PR requests much more work in terms of polishing overall concept, and had to include possible user cases with creating nice and smooth workflow. For example, easy mappings parameters to automation editors, RMB on desired parameter and "connect to new automation track"?

GrimmZane:
> @GrimmZane Oh right good point, is this something that should be done in this pull request though? It would still involve modifying all the editors.

The play buttons? yeah that is definitely outside the scope of this PR afaik

Unless you meant the button removal? That wouldn't need any editor modification?
lmms_ux-modernize-toolbar_slim_declutter
It'd just involve removing some buttons from the toolbar. the editors already function in way that they're already a redundancy. As mentioned you literally open piano roll by double clicking or right click > open patterns, which is a mandatory to everyone's workflow without exception right now for instrument and automation tracks.

beat pattern you can technically just use the drop down to select the active beat pattern, but you can still use the double click/RMB method with that too,

MetaDark:
@GrimmZane I'm still not really sold on reducing the height of the toolbar by default, the padding looks off, but it may be a useful as an option. Although I think it would make sense to remove redundancy with the buttons.

GrimmZane:
> @GrimmZane I'm still not really sold on reducing the height of the toolbar by default. The padding looks off, but it may be a useful as an option. Although I think it would make sense to remove redundancy with the buttons.

Set your resolution to 1440×900 or 1366×768 (common laptop/tablet display resolutions even today)
And you'd be sold immediately 😅 even at 1080p I find it cramped.

the pertinent UI and information is presented identically, and still reads well at a glance. I think it's difficult to argue for aesthetic dead space over functionality, until moving away from MDI.

I did unify the button sizes in my edit, I just went with the bigger ones without really thinking much of it, I could have gone with just the smaller variants, that would have made more sense.  In fact, might as well try that.
lmms_ux-modernize-toolbar_slim_small

Any better? Margins around the buttons increase from 2px top/bottom to 4px.

RebeccaDeField:
> @RebeccaDeField I completely agree, we should stick as close to Budislav's mockup as much as possible.

I may have miscommunicated on this point. I think we should follow Budisla's intentions behind his mockup, especially considering a lot of the buttons we're debating won't be applicable if it really was single window as Tres has been trying to point out. The objective at hand is de-cluttering, improving the "at a glance" readability of the toolbar and the usability within what LMMS is right now (not single window). If the design organically changes and improves beyond the mockup we're using as reference, that's okay, there's no reason to match just for the sake of matching.

I like this one, and everyone seems to agree that this is a good middle ground. The spacing isn't different than what LMMS users are already used to. Let's move forward with this one then?

82733576-b287af00-9ce2-11ea-8df2-d709c910f181

What's great about the option above, is that it's essentially the same mockup as the "aligned to corners" version, so if people truly miss the file operations buttons, we can just add to the right corner and call it a day.

> i'd definitely not go that far, poor and irksome though for sure, especially on Ultra Wides, For someone who has RSI issues personally i would strongly advocate against putting any relatively frequently used element far away from active workspace area. I'd wager it would bother many users.

@GrimmZane  I've dealt with serious RSI issues before, so I understand your concern. Doesn't this option fix the issue with mouse distance/travel by aligning the buttons to the left?

Future improvements/PR's that we will want to consider:
- Moving play/stop buttons to the toolbar
- Reducing hight of the toolbar

Out of the scope of this PR, but not out of the scope of possibility. Let's open new PR's to work on these aspects, should someone be available and willing to work on it.

> Again I'd be willing to open a PR for a slim toolbar, and if possible make it a 'restart required' toggle option. If keeping the toolbar padding is decided.

Perfect :+1: 

I think the "redundant" buttons may be still useful to beginner users, no? Seems like this is the last item that needs to be decided to wrap this discussion up :confetti_ball:

RoxasKH:
> 
> 
> > @RebeccaDeField I completely agree, we should stick as close to Budislav's mockup as much as possible.
> 
> I may have miscommunicated on this point. I think we should follow Budisla's intentions behind his mockup, especially considering a lot of the buttons we're debating won't be applicable if it really was single window as Tres has been trying to point out. The objective at hand is de-cluttering, improving the "at a glance" readability of the toolbar and the usability within what LMMS is right now (not single window). If the design organically changes and improves beyond the mockup we're using as reference, that's okay, there's no reason to match just for the sake of matching.
> 
> I like this one, and everyone seems to agree that this is a good middle ground. The toolbar height isn't different than what LMMS users are already used to. Let's move forward with this one then?
> 
> 82733576-b287af00-9ce2-11ea-8df2-d709c910f181
> 
> What's great about the option above, is that it's essentially the same mockup as the "aligned to corners" version, so if people truly miss the file operations buttons, we can just add to the right corner and call it a day.
> 
> > i'd definitely not go that far, poor and irksome though for sure, especially on Ultra Wides, For someone who has RSI issues personally i would strongly advocate against putting any relatively frequently used element far away from active workspace area. I'd wager it would bother many users.
> 
> @GrimmZane I've dealt with serious RSI issues before, so I understand your concern. Doesn't this option fix the issue with mouse distance/travel by aligning the buttons to the left?
> 
> Future improvements/PR's that we will want to consider:
> 
>     * Moving play/stop buttons to the toolbar
> 
>     * Reducing height of the toolbar
> 
> 
> Out of the scope of this PR, but not out of the scope of possibility. Let's open new PR's to work on these aspects, should someone be available and willing to work on it.
> 
> > Again I'd be willing to open a PR for a slim toolbar, and if possible make it a 'restart required' toggle option. If keeping the toolbar padding is decided.
> 
> Perfect +1
> 
> I think the "redundant" buttons may be still useful to beginner users, no? Seems like this is the last item that needs to be decided to wrap this discussion up confetti_ball

I totally agree with this. It's good to see, close to Budislav's one, it doesn't break the workflow, button are on the left just as they were before, and as i said i can get myself used to using the menu for saving and other things.

This is another step straight to single-window, so it's really important, at least for me.

GrimmZane:
> @GrimmZane I've dealt with serious RSI issues before, so I understand your concern. Doesn't this option fix the issue with mouse distance/travel by aligning the buttons to the left?

Yeah that layout is perfectly fine. 👍

Agree, should get this merged without too much stalling, instead of derailing it too much with future proposals. Looks like I'll wait for this to get merged before forking, see how that goes.

Last comment on it, worth noting this makes the current height an understandable decision.
Screenshot_2020-05-24_05-14-32
But isn't the case with these changes.

> I think the "redundant" buttons may be still useful to beginner users, no?

Personally I think getting new users habituated to potentially forming a workflow based on redundancies, ones that will potentially disappear in the next major overhaul anyway, isn't too compelling of a rationale. But I think it becomes far less important if getting rid of file ops is the decision taken. 

And a plethora of buttons is just as likely to slow down beginners as benefit them.

Any removals are likely to effect someone who has become habituated to them including file ops.

I'm starting to lean towards probably not removing anything, and in my fork/proposal creating a "minimal toolbar" that removes buttons as well as the slimming.

RebeccaDeField:
> Personally I think getting new users habituated to potentially forming a workflow based on redundancies, ones that will potentially disappear in the next major overhaul anyway, isn't too compelling of a rationale. But I think it becomes far less important if getting rid of file ops is the decision taken.
> 
> And a plethora of buttons is just as likely to slow down beginners as benefit them.
> 
> Any removals are likely to effect someone who has become habituated to them including file ops.

Yeah, I do believe we're taking out the file ops from the feedback given here to start.

You make some solid points here, I think we are down to minor enough differences that we can leave the additional show/hid button removal up to the op on this. We'd know soon enough from users testing if this is significant enough to need to add back or not.

GrimmZane:
Okay I know I said last comment 😅, but I actually pulled the fork and edited the height variable and tested it out.

And without any editing outside a single variable, I think it's a win/win and pretty hard to justify 64px toolbar

(ignore the button layout, this isn't a layout proposal, it's just building the current snapshot)

Here is 40px
lmms_ux-modernize-toolbar_slimmer_40
and here is 48px
lmms_ux-modernize-toolbar_slimmer_48

Both are bigger than my edit which was 36, and requires editing the toolbar frame, so ditch that.
40 is close enough and feels better to me, but if padding is wanted 48 feels like a good compromise.

Is the menubar height tied to Qt theme? Seems that's uneditable in the source.

I really hope at the very least 48 is considered for this PR, I'm thinking a more helpful feature I could look into, would be a fullscreen mode that makes use of the space taken up by the OS titlebar & taskbar 🙂

Spekular:
Just throwing this out there: If the top right is left empty, it could be used to display hover-sensitive info. Looking somewhere is faster and easier than moving the mouse there, so this would make the space there useful without the distance being a concern.

The actual hover functionality is obviously out of scope for this PR, but I wanted to put the idea out there.

RebeccaDeField:
@GrimmZane 48px seems like a reasonable compromise to me, might raise to 50px even? As long as we keep a balance of readable text sizes and not cramped I don't see why not. If it's easy enough, sure, let's slim down the bar a bit in this PR. If not, let's pursue it in a follow-up. I will help review it from a design perspective either way.

claell:
Regarding shrinking the height of the bar:

I visually like the original height better than any mockups with reduced height. After thinking a bit about it, I think that is not only a thing of looking good (and "wasting" precious space) but also having actual functional benefits. I have often seen people vote to use space as efficient as possible and put as much buttons in as possible. I also used to have this point of view. It turns out however that when doing this, you usually get a pretty cluttered and confusing UI, which is sometimes referred to as "made for power users". However, I think that even power users who are pretty used to the aspects of the complicated UI will have problems with such kind of designs.

The benefit from having more space is a more clean and distinguishable design leading to a better ability to orientate fast and find what you are looking for. The current widget with (mostly) time and tempo information is a visual element that one looks at probably rather often. Having spacing around it removes distraction of that information by other UI elements. But that is just my POV.

Lathigos:
Whoa, lots of discussion I have missed it seems. :P 

I agree with @claell that removing the spacing from the centermost widget will add more visual clutter, and being able to easily orient oneself in the program is very valuable. Reducing the height could be something for a future PR once more of the buttons are removed due to the removal of MDI.

Regarding the other changes requested, I am very much in favor of removing the file operation buttons, since they are redundant anyway.

So, if I understand correctly, the only changes for now would be to remove the file operation buttons and instead move the window buttons there, right? I think for now that's a good middle ground. I'll work on those parts, then.

Just my few cents: I would prefer not to delay merging this PR too long after this, since a lot of the changes requested I consider to be out of scope of this PR. In the end, a lot of the discussion is also about the window buttons that will be ultimately removed once MDI is gone, so I think I would rather start working on removing MDI first (which is a large task) before we design the toolbar into perfection, and for me to be able to begin working on that it would be nice to have this PR merged first so I can continue building from there. I agree with @RebeccaDeField that later changes can go in a subsequent PR.

claell:
I fully agree that having this merged fast and postponing further fine tuning will probably be pretty beneficial.

Also @GrimmZane since UI stuff can be pretty individual there are probably more users like you that would really like a more "space efficient" UI for their workflow. Especially when it comes to smaller resolutions that might get relevant. So as you proposed, that can hopefully be discussed as an option in a future PR or issue. Maybe as an option (one just has to be careful not to "clutter" the settings too much as well 😉) or maybe even introducing some "responsiveness" based on the used resolution.

MetaDark:
There was a lot of discussion on this, but it's been silent for a while now. Is there currently anything blocking this from being merged in?

Spekular:
I'd love to see @Lathiagos update the original post with a screenshot of what this PR currently looks like. That would make it easier to do a review without necessarily building/installing the PR. There are also a few merge conflicts to be fixed, according to github.

qnebra:
Basically, at this moment toolbar with this PR merged is far less usable than in previous version. You had gigantic empty space with few buttons, and all file operations hidden in "File" menu. I will edit this comment to post screenshot in around 15 minutes.

claell:
Afaik it has been reached a consensus what should be done with this PR. Further things were postponed. After that @Lathigos made the changes. Since then it has been waiting to be merged or for a final review. Don't know who is responsible or whether they are aware of the state.

qnebra:
toolbaratthismoment

Current look of toolbar with this PR.

RebeccaDeField:
> Afaik it has been reached a consensus what should be done with this PR. 

Yep, for now, we came to the best compromise and conclusion with the information available. We decided to go ahead with this layout and make possible changes or enhancements as another stage.

> Current look of toolbar with this PR.

This looks good to me, this should be ready to test and merge?

russiankumar:
There seem to be conflicting files. I think the original dev has to merge the latest master?

Also I'm willing to test, will get onto it after a while.

Spekular:
The code needs review as well, and there are surprisingly big changes here. Some of them look suspect to me, but I haven't had time to look into it myself.

russiankumar:
Alright, I'll try some code review as well. I'm not the best coder myself, but I'll look for suspicious changes.

src/core/FxMixer.cpp
```diff
@@ -172,8 +172,9 @@ void FxChannel::doProcessing()
 		m_stillRunning = m_fxChain.processAudioBuffer( m_buffer, fpp, m_hasInput );
 
 		Mixer::StereoSample peakSamples = Engine::mixer()->getPeakValues(m_buffer, fpp);
-		m_peakLeft = qMax( m_peakLeft, peakSamples.left * v );
-		m_peakRight = qMax( m_peakRight, peakSamples.right * v );
+		
+		m_peakLeft = peakSamples.left * v;
+		m_peakRight = peakSamples.right * v;
```
russiankumar, line 177: Source for potential bugs?

src/gui/widgets/CPULoadWidget.cpp
```diff
@@ -62,42 +70,14 @@ CPULoadWidget::~CPULoadWidget()
...
-		int w = ( m_leds.width() * m_currentLoad / 300 ) * 3;
-		if( w > 0 )
-		{
-			p.drawPixmap( 23, 3, m_leds, 0, 0, w,
-							m_leds.height() );
-		}
-	}
-	QPainter p( this );
-	p.drawPixmap( 0, 0, m_temp );
-}
```
russiankumar, line 87: Why's paint event deleted? Didn't exactly understand

src/core/FxMixer.cpp
russiankumar: For reference: there used to be a QMax to ensure it doesn't cross some maximum amount (peakSamples.left or .right respectively I believe)

russiankumar:
@Lathigos, @Spekular

### After testing, here's what I found

- OUT meter works as expected after testing with a sine wave (as discussed here before)
- Toolbar doesn't break upon moving anywhere, works as expected
- Interacting with the toolbar also works as expected
- Double-click to set time signature doesn't work, though I'm not sure if that's a thing in the current toolbar either
- Hints for buttons and toolbar mostly work as expected with following two exceptions
- Tooltips for controller rack and project notes - `show/hide xxx (shortcut)` - inconsistent with remaining tooltips - `xxx (shortcut)`
- Tooltips for volume and pitch activate only over the control, not the `VOL` and `PITCH` text, where I think they are needed more - also inconsistent with the `OUT` and `CPU` text where the tooltip _does_ display

#### Screenshot

image

This is all I found after some testing. I can do some more to see what else is wrong though I think beyond these points there's nothing much else to worry about. @Spekular, I'll leave the code review to you since I don't think I can manage figuring out all of that long code.

@Lathigos, sorry for being so late, and great thanks for your contribution!

Spekular:
I still don't have time for a full review right now, but just as I suspected, IntegerSpinBox is a copy paste of LcdSpinBox. Making a new, mostly duplicated class in order to change how it's drawn is not a good solution. The common functionality should be in a single class (probably a new abstract superclass) and only the difference (draw method) should be separate. This way we're not repeating ourselves (having to make all future changes to this class in two places).

This issue alone makes the PR unmergeable.

RoxasKH:
I spotted a bug: the reset function is completely useless for text-based widgets atm, it just suggests again the current value that was set.

Example with master volume:
immagine

Obviously is the same also with tempo, time signature and pitch.

roshanrules1:
> If you are proposing to remove the oscilloscope, do we have an idea of where it may be placed or accessed instead? 

i think we can use it as an effect (like spectrum visualiser).

roshanrules1:
Im late to this discussion but here's what i found

- removing the file ops buttons.
- removing oscilloscope and putting it within a plugin (similar to spectrum visualiser).
- reducing the size of the toolbar.
- left aligned buttons, centrally aligned widget.
- if i left something out, just comment it out

So here's my take on what's to do with this pr:
Since the code refactoring brought in a lot of changes, fixing this pr can be hella difficult. So i think this or would benefit from splitting up.
- 1st sub pr shall be removing the file ops buttons (because that's the easiest to do and has little to no effect in most people's workflow). This one shall be a part of 1.3 milestone.
- 2nd sub pr shall be removing the oscilloscope and putting it in as an effect and also reducing the size of the toolbar(this one needs a bit more effort but it cleans up the toolbar a bit better than the 1st one). I would prefer it to be a part of 1.4 (cause i believe small incremental changes are better than big ones).
- the 3rd sub pr shall be the widget itself (what is left of this pr after the first 2 sub prs). This can be made a part of lmms 2.0 (big changes require bigger jumps in version number).
- 4th sub pr (perhaps a follow up) can be a hovering tooltips window (or whatever it is called). This too can go into 2.0

This is just my opinion. If you differ, feel free to discuss. The point of this reply is to initiate discussion.

actuallyexeon:
will this ever be pulled in after the requested changes by Sawuare?

Sawuare (approved):

Rossmaxx:
Don't we need to fix the merge conflicts. @Lathigos is it fine if someone else take over this pr?

FyiurAmron:
@Rossmaxx TBH Lathigos doesn't seem to be active on GitHub for more than 4 years now. I might be mistaken, but the data would seem to indicate so I think.

Rossmaxx:
In that case, I'll try taking over.

Rossmaxx:
I did look into this a bit. seems like it's better to start from scratch and refactor the entire toolbar code into it's own file on the way. But I'm postponing the resurrection to after 1.3.

Eveeifyeve:
Idk I might follow this pr modernising the song editor a bit because the mute switches and stuff is a bit confusing to a new user.

PR #5262, opened 2019-10-20
Fix #3005

### LINKED ISSUE #3005: Vertical piano in piano roll not responsive to mouse clicks.
I noticed that in piano roll you can't play melodies on vertical piano unless you play really slow. When clicking quickly then it just don't play some of the notes. I think vertical piano in piano roll should be as responsive as horizontal piano in intrument window. There you can play as fast as your fingers press.

Veratil:
`mouseDoubleClickEvent` wasn't forwarding to `mousePressEvent` like the default implementation does. This adds that check.

I don't believe this should cause any problems double clicking elsewhere (I can't seem to find any problems with a basic test), but just in case we can restrict this to just the piano key area.

### DIFF
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -1965,7 +1965,7 @@ void PianoRoll::mouseDoubleClickEvent(QMouseEvent * me )
 	}
 	else
 	{
-		mousePressEvent(me);
+		QWidget::mouseDoubleClickEvent(me);
 	}
 }
 
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -1963,6 +1963,10 @@ void PianoRoll::mouseDoubleClickEvent(QMouseEvent * me )
 			enterValue( &nv );
 		}
 	}
+	else
+	{
+		mousePressEvent(me);
+	}
 }
 
 


PhysSong:
> `mouseDoubleClickEvent` wasn't forwarding to `mousePressEvent` like the default implementation does.

Then why don't you call the default implementation instead of calling `mousePressEvent` directly? I understand both give the same result, but this will make it easy to see what it does.

Veratil:
> 
> 
> > `mouseDoubleClickEvent` wasn't forwarding to `mousePressEvent` like the default implementation does.
> 
> Then why don't you call the default implementation instead of calling `mousePressEvent` directly? I understand both give the same result, but this will make it easy to see what it does.

I don't understand, is this not easy to see what it does?

src/gui/editors/PianoRoll.cpp
```diff
@@ -1963,6 +1963,10 @@ void PianoRoll::mouseDoubleClickEvent(QMouseEvent * me )
 			enterValue( &nv );
 		}
 	}
+	else
+	{
+		mousePressEvent(me);
```
lukas-w:
```suggestion
		QWidget::mouseDoubleClickEvent(me);
```

Veratil:
> 
> 
> @PhysSong's suggestion is the standard way of doing this, see https://doc.qt.io/qt-5/eventsandfilters.html#event-handlers for reference. We don't act on the event, so we should let the base class handle it.

Ah! I understand now. 😄 Thanks!

lukas-w:
Merged and cherry-picked to `stable-1.2` via a8d91b10e8dd32157d25caeb97597eb018e9405c.

Issue #4312, opened 2018-04-22
Error Report Can Go Offscreen


Spekular:
If a large number of samples & VSTs fail to load, the error report displayed may grow tall enough to go off the screen. 

I believe the prompt should instead stay onscreen (never grow taller than screen size) and display a scrollbar. The ability to resize the prompt horizontally would also allow longer paths to display in fewer lines.

Sawuare:
The error reports use a [QMessageBox](https://doc.qt.io/qt-5/qmessagebox.html), which doesn't support scroll bars. It's supposed to support a size grip, but it doesn't seem to work when I tried it.

It takes _a lot_ of errors for the report to go off-screen. so I don't think it's worth a custom dialog.

Sawuare:
I tried to add a scroll area before and it did not work, but now it does. I will fix this.

Sawuare:
So applying a scroll area does not work as expected.

![image](https://user-images.githubusercontent.com/25503477/40757388-ec571702-6490-11e8-8ba1-6c2610a32054.png)
- The message box is inside a scroll area window, instead of its own window.
- The window is not centered and is non-modal (does not block input to other windows).
- Window title cannot be changed.
- OK button is not always visible.

I might be doing something wrong though, so for comparison, here is my code:
```cpp
QMessageBox * mb = new QMessageBox;
mb->setWindowTitle(tr("LMMS Error report"));
mb->setText(errorSummary());
mb->setIcon(QMessageBox::Warning);

QScrollArea * sa = new QScrollArea;
sa->setHorizontalScrollBarPolicy(Qt::ScrollBarAsNeeded);
sa->setVerticalScrollBarPolicy(Qt::ScrollBarAsNeeded);
sa->setWidget(mb);
sa->show();
```
and the current code:
```cpp
QMessageBox::warning( NULL, tr("LMMS Error report"), errorSummary(), QMessageBox::Ok );
```

PR #5269, opened 2019-10-22
Consolidate error messages while loading project

softrabbit:
Consolidate error messages while loading project and report a count for repeated errors.

Addresses part of #4312.

### DIFF
--- a/src/core/Song.cpp
+++ b/src/core/Song.cpp
@@ -1420,19 +1420,15 @@ void Song::clearErrors()
 
 void Song::collectError( const QString error )
 {
-	if( !m_errors.contains( error ) )
-	{
-		m_errors[ error ] = 1;
-	} else {
-		m_errors[ error ]++;
-	}
+	if (!m_errors.contains(error)) { m_errors[error] = 1; }
+	else { m_errors[ error ]++; }
 }
 
 
 
 bool Song::hasErrors()
 {
-	return ( !m_errors.isEmpty() );
+	return !(m_errors.isEmpty());
 }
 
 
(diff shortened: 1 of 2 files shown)


Spekular:
This is perfect for me to test! Since I use a version with relative paths merged, any version without them fails to load nearly everything.

Looks great, caught both repeated VST errors and repeated sample errors:
example

src/core/Song.cpp
```diff
@@ -1420,21 +1420,37 @@ void Song::clearErrors()
 
 void Song::collectError( const QString error )
 {
-	m_errors.append( error );
+	if( !m_errors.contains( error ) )
+	{
+		m_errors[ error ] = 1;
+	} else {
+		m_errors[ error ]++;
+	}
```
Spekular, line 1428: I believe` } else {` is a style violation. The simplest way to fix this is to move else to a new line. However, assuming it doesn't break the 80 char line length I you could also do:

```suggestion
	if( !m_errors.contains( error ) ) { m_errors[ error ] = 1; }
	else { m_errors[ error ]++; }
```

Which I personally think would be the nicest format.

Spekular: Just checked, the line length in my suggestion is fine (also the max is 120 and not 80, oops).

PhysSong: FYI, `m_errors[error]++` always works because `QHash::operator[]` will automatically insert 0 if the key is not in the hashmap. The same holds for STL containers as well.
I'm okay with using an `if`-`else` block, too.

Also, please change `m_errors[ error ]` to `m_errors[error]`.
https://github.com/LMMS/lmms/wiki/Coding-Conventions#braces

Issue #5263, opened 2019-10-21
Add warning dialogue for "Remove this track" button in song editor.


coderman64:
I have noticed, while using the current release version of LMMS, it is incredibly easy to accidentally hit "Remove this track" when you mean to hit "Clone this track." In other applications, permanent operations like this would require a dialogue to make sure the user actually wants the track to be deleted and didn't click it by mistake. I propose a similar dialogue be implemented in LMMS, somewhat akin to the save/discard dialogue that appears when you try to close a project with unsaved changes:
![image](https://user-images.githubusercontent.com/12971494/67178174-c4e1e900-f39f-11e9-97dc-020b284743b0.png)
Quick-and-dirty GIMP mockup:
![image](https://user-images.githubusercontent.com/12971494/67178332-505b7a00-f3a0-11e9-9cdf-6ed4854a183f.png)

itlives:
If this is still open ill gladly take it. I used to contribute on lmms a while ago but I had to pause because of my job.

zonkmachine:
@fschuelke Please fix this! :whale:

musikBear:
I have inserted a response control through a messagebox: 
![image](https://user-images.githubusercontent.com/6662066/82563847-0f1a8b00-9b78-11ea-9d0e-eda719c6df49.png)
This can be closed.

Spekular:
> This can be closed.

Not until you've opened a PR that's been reviewed, approved, and merged. For starters it should say "Removing a track can't be *undone*." and I would prefer the OP's suggested "Are you sure you want to remove [track name]?" over "Do you want to proceed".

Secondly, I would personally appreciate if this dialogue could be disabled in settings or via a "don't ask me again" checkbox as it's never been an issue for me. (If it really is a common issue then the layout of the menu ought to change to prevent this, but that's a separate issue and I'm fine with this solution if it's optional)

zonkmachine:
Thirdly there is a PR adressing this already and I was just about to bump it.
https://github.com/LMMS/lmms/pull/5274

musikBear:
:p so 
![image](https://user-images.githubusercontent.com/6662066/82620213-343ee600-9bd8-11ea-93ce-91c43851272a.png) Means *'done and in PR'*, i only looked at the date ..nov 2019..
Nwm i thought myself usage of QmessageBox, through input of a user' response 
Time well spent anywitch

coderman64:
[...]

PR #5274, opened 2019-10-26
Added message box prompt when user tries to delete track

fschuelke:
commit for #5263


PhysSong:
You should use tabs for indentation instead of spaces.
https://github.com/LMMS/lmms/wiki/Coding-Conventions#indentation

src/core/Track.cpp
```diff
@@ -3042,3 +3043,30 @@ void TrackView::createTCOView( TrackContentObject * tco )
...
 	tco->selectViewOnCreate( false );
 }
+
+
+/*! \brief Show a message box warning the user that this track is about to be closed */
+bool TrackOperationsWidget::warnRemoveTrack()
+{
+    QString messageRemoveTrack = tr("After removing a track, it can not"
+                                    "be recovered. Are you sure you want to remove track \"");
+    messageRemoveTrack += m_trackView->getTrack()->name() + "\"?";
```
PhysSong: I recommend using [`QString::arg()`](https://doc.qt.io/qt-5/qstring.html#arg) instead of manual concatenation. In some languages, that word order will make no sense when translated.

src/core/Track.cpp
```diff
@@ -3042,3 +3043,30 @@ void TrackView::createTCOView( TrackContentObject * tco )
...
+    QString messageTitleRemoveTrack = tr("Removing this track");
```
PhysSong: I think `Confirm removing track` will be slightly better.

Spekular: I would go with `Confirm removal`. Or, if you really feel like "track" needs to be there, `Confirm track removal`. The latter seems redundant to me though, the user knows they clicked on a track-related action and the message content specifies which track will be removed.

src/core/Track.cpp
```diff
@@ -3042,3 +3043,30 @@ void TrackView::createTCOView( TrackContentObject * tco )
 	}
 	tco->selectViewOnCreate( false );
 }
+
+
+/*! \brief Show a message box warning the user that this track is about to be closed */
+bool TrackOperationsWidget::warnRemoveTrack()
+{
+    QString messageRemoveTrack = tr("After removing a track, it can not"
+                                    "be recovered. Are you sure you want to remove track \"");
```
zonkmachine: You need to insert a space after 'not' or it will be written together with 'be' on the line after.
`tr("After removing a track, it can not "`

fschuelke:
> You should use tabs for indentation instead of spaces.
> https://github.com/LMMS/lmms/wiki/Coding-Conventions#indentation

I did use tabs, could be the issue of the IDE... do we have a clang-tidy format or anything like that? the code shows up normally formatted for me

fschuelke:
@PhysSong i checked the indentation, but for me it looks fine. Also doesnt show any wrong indentation in the git file viewer either

PhysSong:
If there was a tab instead of 4 spaces, this wouldn't be possible.
image

fschuelke:
> If there was a tab instead of 4 spaces, this wouldn't be possible.
> image

im gonna commit it directly in github. dont know why clion shows me correct indentation when there isnt

src/core/Track.cpp
```diff
@@ -2793,13 +2796,11 @@ QMenu * TrackView::createFxMenu(QString title, QString newFxLabel)
...
-/*! \brief Close this track View.
- *
- */
+/*! \brief Close this track View. */
 bool TrackView::close()
 {
-	m_trackContainerView->removeTrackView( this );
-	return QWidget::close();
+    m_trackContainerView->removeTrackView( this );
+    return QWidget::close();
```
PhysSong: Spaces -> tabs

src/core/Track.cpp
PhysSong: Anyway, this title really needs a change.

fschuelke: @PhysSong latest commit has it changed already

PhysSong: I can only see the change in the function name, not the message box title.

Umcaruje:
I think this is not the right approach to tackle this problem. The user
should be simply able to actually undo their track deletion.

This feature makes it a hassle to actually delete a track and could get
pretty annoying in big projects. While it does help with the accidental
deletions it feels like a step back with the regular delete functionality.
Implementing proper undo would resolve both these issues.
-- 
-Uroš|Umcaruje

PhysSong:
> The user should be simply able to actually undo their track deletion.

We should allow that finally. However, I think it needs much more code changes than what are in this PR.

src/core/Track.cpp
fschuelke: > I can only see the change in the function name, not the message box title.

oh you were referring to the message box title. Should it be "Confirm removal" as well?

fschuelke:
/usr/local/bin/appdmg -> /usr/local/lib/node_modules/appdmg/bin/appdmg.js

macos-alias@0.2.11 install /usr/local/lib/node_modules/appdmg/node_modules/macos-alias
node-gyp rebuild

ERROR:root:code for hash md5 was not found.

First lines of the circleci calls; is the error caused by any programming?

PhysSong:
Ignore the error, it's not related to this PR. You can see the same message in many other places.

PhysSong:
> I think this is not the right approach to tackle this problem.

@Umcaruje Then, what do you want for this PR?

zonkmachine:
What's the status of this? Whitespace changes can be done by any of us.

PhysSong:
I resolved merge conflicts and fixed formatting issues.

PhysSong:
There was a suggestion from Discord: there should be a way for users to suppress the dialog until the proper undo/redo function is implemented.

Spekular:
So I'm struggling to push to this branch for whatever reason, but I've gone ahead and added a setting that lets users opt out of this warning. They can either check a box that says "Don't show again" in the dialogue, or go to Edit > Settings and change it there.

<https://github.com/LMMS/lmms/compare/master...Spekular:trackDeletion>

PhysSong:
@Spekular I can, but you need a default `1` in `TrackOperationsWidget::confirmRemoval` for `needConfirm`.

Spekular:
@PhysSong good point! I added a default value now.

Spekular (approved):

Issue #5278, opened 2019-10-27
Replace macros by delegating ctors


JohannesLorenz:
Use delegating constructors instead of macros. For beginners, but requires C++11 knowledge.

Start your PR by branching from `stable-1.2`. Fix the following comments:

```
src/gui/widgets/LedCheckbox.cpp://! @todo: in C++11, we can use delegating ctors
src/gui/widgets/Knob.cpp://! @todo: in C++11, we can use delegating ctors
src/gui/widgets/LcdWidget.cpp://! @todo: in C++11, we can use delegating ctors
```

knittl:
I've created a pull request to replace the macros with delegating constructors: https://github.com/LMMS/lmms/pull/5279

JohannesLorenz:
Thanks for the PR :+1:

PR #5279, opened 2019-10-27
Replace initializer list macros with delegating constructors

knittl:
This replaces macros with delegating constructors as mentioned in #5278

### DIFF
--- a/src/gui/widgets/LcdWidget.cpp
+++ b/src/gui/widgets/LcdWidget.cpp
@@ -39,41 +39,31 @@
 
 
 
-//! @todo: in C++11, we can use delegating ctors
-#define DEFAULT_LCDWIDGET_INITIALIZER_LIST \
-	QWidget( parent ), \
-	m_label(), \
-	m_textColor( 255, 255, 255 ), \
-	m_textShadowColor( 64, 64, 64 )
-
 LcdWidget::LcdWidget( QWidget* parent, const QString& name ) :
-	DEFAULT_LCDWIDGET_INITIALIZER_LIST,
-	m_numDigits( 1 )
+	LcdWidget( 1, parent, name )
 {
-	initUi( name );
 }
 
 
 
 
 LcdWidget::LcdWidget( int numDigits, QWidget* parent, const QString& name ) :
-	DEFAULT_LCDWIDGET_INITIALIZER_LIST,
-	m_numDigits( numDigits )
+	LcdWidget( numDigits, QString("19green"), parent, name )
 {
-	initUi( name );
 }
 
 
 
 
 LcdWidget::LcdWidget( int numDigits, const QString& style, QWidget* parent, const QString& name ) :
-	DEFAULT_LCDWIDGET_INITIALIZER_LIST,
+	QWidget( parent ),
+	m_label(),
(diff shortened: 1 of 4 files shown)


src/gui/widgets/LedCheckbox.cpp
```diff
@@ -56,13 +52,10 @@ LedCheckBox::LedCheckBox( const QString & _text, QWidget * _parent,
 
 LedCheckBox::LedCheckBox( QWidget * _parent,
 				const QString & _name, LedColors _color ) :
-	DEFAULT_LEDCHECKBOX_INITIALIZER_LIST
+	LedCheckBox( QString::null, _parent, _name, _color)
```
knittl: I'm not 100% sure about `QString::null` here. Is that okay or should that be `NULL` or something else? It's been a while since I did some C++ :) the macro left the member uninitialized when calling this constructor.

JohannesLorenz:
Thanks, I'll review this in few hours.

src/gui/widgets/LcdWidget.cpp
```diff
@@ -39,42 +39,32 @@
...
-	initUi( name );
 }
 
 
 
 
 LcdWidget::LcdWidget( int numDigits, QWidget* parent, const QString& name ) :
-	DEFAULT_LCDWIDGET_INITIALIZER_LIST,
-	m_numDigits( numDigits )
+	LcdWidget( numDigits, QString("19green"), parent, name )
```
JohannesLorenz: Since you've added the default color here, would you mind removing it from the default params of `initUi` in the header?

src/gui/widgets/LedCheckbox.cpp
JohannesLorenz: Before this change, `m_text` has been constructed with its default ctor `QString::QString()`, so I'd simply pass `QString()` here:

```suggestion
	LedCheckBox( QString(), _parent, _name, _color )
```

src/gui/widgets/LcdWidget.cpp
```diff
@@ -39,42 +39,32 @@
...
-#define DEFAULT_LCDWIDGET_INITIALIZER_LIST \
-	QWidget( parent ), \
-	m_label(), \
-	m_textColor( 255, 255, 255 ), \
-	m_textShadowColor( 64, 64, 64 )
-
 LcdWidget::LcdWidget( QWidget* parent, const QString& name ) :
-	DEFAULT_LCDWIDGET_INITIALIZER_LIST,
-	m_numDigits( 1 )
+	LcdWidget( 1, parent, name)
```
JohannesLorenz: A space after `name` would be more consistent with the surrounding code.

JohannesLorenz:
Code and style review complete :tada: 

Once you're finished with rework, I'll run some quick checks and then I'd do a squash and merge.

src/gui/widgets/LcdWidget.cpp
knittl: Thanks, I've missed this blank twice. Fixed

src/gui/widgets/LedCheckbox.cpp
knittl: Done, replaced with the default constructor

PR #5280, opened 2019-10-27
 Simplify the oscilloscope

husamalhomsi:
- Rename `VisualizationWidget` to `Oscilloscope`.

- Remove unused code.

- Remove ability to manually deactivate the widget.

- Remove obsolete setting related to the widget.

- `addStretch()` before adding the widget.

- Use 1 color instead of 3 volume dependent colors (normal, warning, clipping).

- Make background a member variable.


DouglasDGI:
I'm fine with all the other changes, but I think it would be best to keep the three volume dependent colors, I reference them quite often.

PhysSong:
> Remove ability to manually deactivate the widget.

Why?

Sawuare:
@DouglasDGI I understand. I actually think a clipping indicator is useful, but a warning indicator based on the arbitrary range [.9, 1.], not really.

How about a simple stereo meter widget that can indicate clipping, put in the toolbar?

Sawuare:
@PhysSong Because there is no reason to deactivate it, really.

PhysSong:
> there is no reason to deactivate it

I always use LMMS with the oscilloscope enabled, but I think someone might want to disable it because of resource consumptions or personal preferences.

Sawuare:
The oscilloscope looks ugly when disabled. Its resource consumption should be negligible, especially with this PR.

Spekular:
I don't think that's a good reason.

> resource consumption should be negligible

If you're going to use this as an argument you should have data to back it up. Otherwise I think we should default to keeping this, especially as some people need every performance boost they can get. `If (m_active)` doesn't make the code complex or hard to understand, so I don't see the need to get rid of it, code wise.

> especially with this PR

The only performance improvement I can see is that you removed the three colors, which I disagree with. If that's reverted the resource consumption will go back to what it was.

> the oscilliscope is ugly when disabled

In that case, a better solution would be to make it less ugly. Personally, I find it ugly and distracting when enabled and thus never use it.

Instead of removing the ability to disable the oscilliscope completely, I suggest you change it to be active by default. That seems like a sane setting to me, and then people won't see the "ugly" mode unless they specifically click it.

The downside is that the click behavior isn't obvious when the oscilliscope is on, so you might want to add a tooltip on hover or something similar to indicate that it can be turned off by clicking.

noahb01:
a FL-style FFT display would also be cool to have as an oscilloscope mode.

PhysSong:
@Sawuare I suggest reverting everyting except:
> - Rename `VisualizationWidget` to `Oscilloscope`.
> - Remove unused code.
> - `addStretch()` before adding the widget.
> - Make background a member variable.

Any opinions?

Sawuare:
All right. I'll open a new PR.

PR #5282, opened 2019-10-28
Only filter out <>:"/\|?* while exporting tracks.

Cyp:
This now preserves most characters in exported track names.

### DIFF
--- a/include/Track.h
+++ b/include/Track.h
@@ -72,6 +72,8 @@ const int DEFAULT_TRACK_HEIGHT = 32;
 
 const int TCO_BORDER_WIDTH = 2;
 
+char const *const FILENAME_FILTER = "[\\0000-\x1f\"*/:<>?\\\\|\x7f]";
+
 
 class TrackContentObject : public Model, public JournallingObject
 {
--- a/src/core/RenderManager.cpp
+++ b/src/core/RenderManager.cpp
@@ -203,7 +203,7 @@ QString RenderManager::pathForTrack(const Track *track, int num)
 {
 	QString extension = ProjectRenderer::getFileExtensionFromFormat( m_format );
 	QString name = track->name();
-	name = name.remove(QRegExp("[^a-zA-Z]"));
+	name = name.remove(QRegExp(FILENAME_FILTER));
 	name = QString( "%1_%2%3" ).arg( num ).arg( name ).arg( extension );
 	return QDir(m_outputPath).filePath(name);
 }
--- a/src/tracks/InstrumentTrack.cpp
+++ b/src/tracks/InstrumentTrack.cpp
@@ -1615,7 +1615,7 @@ void InstrumentTrackWindow::saveSettingsBtnClicked()
 	sfd.setDirectory( presetRoot + m_track->instrumentName() );
 	sfd.setFileMode( FileDialog::AnyFile );
 	QString fname = m_track->name();
-	sfd.selectFile( fname.remove(QRegExp("[^a-zA-Z0-9_\\-\\d\\s]")) );
+	sfd.selectFile(fname.remove(QRegExp(FILENAME_FILTER)));
 	sfd.setDefaultSuffix( "xpf");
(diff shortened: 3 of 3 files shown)


Spekular:
This looks good to me. Technically it misses [a few illegal names on Windows](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file), but they weren't covered before either, aside from ending a filename with a period.

Out of scope for this PR, but it feels like a bad idea to handle this in these classes. Really, all file writing should delegate to a single method that cleans the filename. That would ensure consistent behavior and make sure every fix (like this) applies to every file we save. Plus, it could be a bit larger (to handle Windows's reserved names and OS differences better) without taking up space in the classes that use it.

Spekular:
Oh, but I don't feel like the INSTALL.txt changes  belong in this. I've seen a lot of activity around fixing CI recently so this fix may already be applied or made redundant in some other PR?

Cyp:
I see INSTALL.txt in master does mention submodules. It doesn't mention that cmake can mysteriously fail to generate a Makefile without them, though. Anyway, dropping the INSTALL.txt commit would be fine.

Cyp:
Rebased on latest stable-1.2, and dropped the INSTALL.txt commit…

PhysSong:
Is it safe to be merged?

PhysSong:
It seems like there are some file systems which support limited sets of characters. Should we take care of them?
See https://en.wikipedia.org/wiki/Filename#Comparison_of_filename_limitations for details.

Cyp:
Now also filters 0x00-0x1F and 0x7F. I hope that covers any semi-serious filesystems there that aren't EBCDIC or with a 16-or-less character limit.

JohannesLorenz:
@Cyp what do you think about a global common function for that, that also filters out periods at the end, like @Spekular suggested? I ask because, imagine the next time someone will change the filter in `RenderManager`, they'll surely forget to change it in `InstrumentTrack`.

PhysSong:
> a global common function for that, that also filters out periods at the end

In most cases, we append the file extension at the end. A trailing period is equivalent to the empty file extension(NOT a file without extensions).

Cyp:
It's a global common string (in the last version of the commit), so there shouldn't be anyone changing it in one place but not the other. I don't think there should be any files written without an extension added, so I don't think checking for trailing dots is needed (although it could be added if needed).

JohannesLorenz:
If we save directories, it may be valuable, but it still seems like a seldom case.

JohannesLorenz:
What were your reasons to choose that filter? E.g.:

1. Why allow `!` but filter `?`?
2. Why allow `;` but filter `:`?

PhysSong:
@JohannesLorenz The filter reflects restricted characters in major OSes/file systems.

JohannesLorenz:
OK, then I consider functional and style reviews done. Testing remains.

JohannesLorenz:
I just tested to save a file `aaa<>:"\|?*bbb` from the instrument window's save button. It saved as `\|?*bbb.xpf`, so it

* did not remove the claimed chars `\|?*`
* removed the `aaa` postfix

Can you please check what happens in your local copy?

Cyp:
Same here, but I think it's unrelated to the commit. The `"` seems to have special meaning to the save dialogue box. If trying to save to `/te"st"` in some other program (kwrite), it tries saving to `/st` (with just one `"`, it refuses to even try to save).

If naming the track `aaa<>:"\|?*bbb` and trying to save in the instrument window, it suggests `aaabbb` as the default filename.

PhysSong:
I checked the regular expression, but I couldn't find any issues with it. I could also confirm the issue with the file dialog, using Qt creator.
I think it's ready to merge.

JohannesLorenz (approved):

JohannesLorenz:
Right, if you don't use `"` in the dialog, it works.

I'll merge, thanks for the contribution!

PR #5286, opened 2019-10-29
Fix scrolling direction in SongEditor due to stuck Ctrl/Shift.

Cyp:
As far as I can tell, MainWindow::focusOutEvent never gets called no matter what I do, so clearKeyModifiers fails to be called.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -480,7 +480,7 @@ ENDIF()
 
 # Due to a regression in gcc-4.8.X, we need to disable array-bounds check
 IF (CMAKE_COMPILER_IS_GNUCXX AND ((CMAKE_CXX_COMPILER_VERSION VERSION_EQUAL "4.8.0") OR (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER "4.8.0") OR LMMS_BUILD_WIN32))
-   SET(WERROR_FLAGS "${WERROR_FLAGS} -Wno-array-bounds")
+   SET(WERROR_FLAGS "${WERROR_FLAGS} -Wno-array-bounds -Wno-attributes")
 ENDIF()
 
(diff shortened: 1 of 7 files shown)


JohannesLorenz:
A bit more info please: What is the issue you are fixing? What are the steps to reproduce that issue?

JohannesLorenz:
Not required but it will help future readers if you split stuff into multiple commits, e.g. cleanups, override fixes, ... It's not difficult, just a `git reset --soft HEAD~1`, and then `git add` on the files you want to add (or `git add -p` if a file shall be splitted).

Cyp:
It fixes that if scrolling with the mouse wheel in the Song Editor, it would often unexpectedly zoom or scroll horizontally instead of scrolling vertically, due to a stuck Ctrl/Shift state.

This was reproducible by switching away from LMMS using Ctrl+F[1-10] to change desktops, and switching back again without using Ctrl. Then LMMS would behave as if the Ctrl key was still down. Not sure how to reproduce Shift being stuck down, but it also happened sometimes somehow.

As a side effect, it should also fix a theoretical issue of it thinking Ctrl/Shift was down during the click when it was actually pressed after the click (might happen in the case of very slow computers).

JohannesLorenz:
I also tested the ctrl key issue, it's working. No regressions for shift and ctrl. Testing finished.

JohannesLorenz:
Thanks!

PhysSong:
FYI, I'm getting a deprecation warning since `Knob::getValue()` uses deprecated `MainWindow::isShiftPressed()`. It will break CircleCI builds on `master` branch.

JohannesLorenz:
I removed `[[deprecated]]` now on `stable-1.2`. Hopefully, this will fix the CI issues on `master`.

Issue #5225, opened 2019-10-06
Automation point adjustment mouse wheel and double click


tecknixia:
https://cdn.discordapp.com/attachments/332258319228207114/629248936099708928/AutomationEditor2-2019-10-03_04.21.20.mp4

I have successfully made this usable already, but since I'm new to this, I'm not sure how what I did fits in with everything else, so it can be altered if necessary.  There are also elements of this issue that could use polishing, which are:

1. the values for detecting the point with the mouse on the y axis are hard coded and do not adapt to scaling changes of the y axis

2. when moving the mouse over a point, the value of the tooltip willl reflect the y value of the mouse position, instead of the point's y level, until the mouse wheel moves.  This causes an inaccurate reading of the point level depending on where the cursor is placed on the point, so the value jumps from the cursor y value to the point y level.

Here are instructions to make it work like I had it:

- AutomationEditor.h

// set variable in header
@ line 212 (or wherever appropriate) insert `float m_point_y_level;`

- AutomationEditor.cpp

// initialize variable
@ line 89 insert `m_point_y_level( 0 ),`

// reset variable to zero if mouse moves
@ line 733 (beginning of mouseMoveEvent, after vaid pattern check) insert `m_point_y_level = 0;`

// I originally intended to include a TODO comment for the 2nd point above (optional)
@ line 1102 TODO: when mouse is over a point (ti@me map key), show level of the point on tooltip instead of mouse cursor y position

// the tooltip value is decided, reads the variable that is set when mouse wheel is moved
@ line 1111 (starts with "QToolTip:showText") replace with:

```
        if (m_point_y_level == 0) {
		QToolTip::showText( tt_pos, QString::number( scaledLevel ), this );
	}

	if (m_point_y_level != 0) {
		QToolTip::showText( tt_pos, QString::number( m_point_y_level ), this );
	}

```
// when mouse wheel moves, detects the point on time map and sets
// variable to the point's y level to be read when displaying tool tip
@ line 1721 (within last "else" for wheelEvent after setting value for m_topBottomScroll) insert:

```
if( we->y() > TOP_MARGIN )
{

	float level = getLevel( we->y() );

	int x = we->x();

	if( x >= VALUES_WIDTH )
	{
		
		// set or move value

		x -= VALUES_WIDTH;

		// get tick in which the cursor is posated
		int pos_ticks = x * MidiTime::ticksPerTact() / m_ppt + m_currentPosition;

		// get time map of current pattern
[...]

PR #5291, opened 2019-10-30
Automation Editor point fine tuning with wheel scroll

tecknixia:
Related to issue #5225 
And replaces old PR #5232 

Allows user to place cursor over an automation point in the Automation Editor, and use scroll wheel to adjust the y level for that point.  It also improves mouse cursor interaction with the point, and shows the point's Y level when mouse cursor is over a point.

https://cdn.discordapp.com/attachments/332258319228207114/629248936099708928/AutomationEditor2-2019-10-03_04.21.20.mp4 (example, not current)

This PR involves the creation of a variable, and the use of `mouseMoveEvent`, `drawCross`, and `wheelEvent`.

 - A variable, `m_pointYLevel` is created and used to hold a point's Y level position.
- The value of `m_pointYLevel` is set when the mouse moves over a point.
- The value of `m_pointYLevel` changes when the mouse wheel is scrolled, and over a point.
- The point's Y level changes when the mouse wheel is scrolled, and over a point.
- If the mouse moves away from a point, the value of `m_pointYLevel` returns to `0`.
- When the value of `m_pointYLevel` is `0`, the tool tip in `drawCross` displays the Y level position of the mouse.
- When the value of `m_pointYLevel` is greater than `0`, the tool tip in `drawCross` displays the point's Y level position.

### DIFF
--- a/include/AutomationEditor.h
+++ b/include/AutomationEditor.h
@@ -218,6 +218,8 @@ private:
 
 	Actions m_action;
 
+	float m_pointYLevel;
+
 	tick_t m_selectStartTick;
 	tick_t m_selectedTick;
 	float m_selectStartLevel;
--- a/src/gui/editors/AutomationEditor.cpp
+++ b/src/gui/editors/AutomationEditor.cpp
@@ -88,6 +88,7 @@ AutomationEditor::AutomationEditor() :
 	m_topLevel( 0 ),
 	m_currentPosition(),
 	m_action( NONE ),
+	m_pointYLevel(0),
 	m_moveStartLevel( 0 ),
 	m_moveStartTick( 0 ),
 	m_drawLastLevel( 0.0f ),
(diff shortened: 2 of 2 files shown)


tecknixia:
@russiankumar #5592 
Issue with macos, not sure how to resolve.
Double clicking is good enough. ( #5292 )
This PR is not as important.  In fact, I just realized that it would be better if the automation point was able to be selected before using the scroll wheel so it doesn't interfere with scrolling on the y-axis.
This PR can be closed.

russiankumar:
Ah, I believe we’ve dropped the overhaul for now. If you need to work on something you can do so, there’s no hurry for now.

IanCaio:
> @russiankumar #5592
> Issue with macos, not sure how to resolve.
> Double clicking is good enough. ( #5292 )
> This PR is not as important. In fact, I just realized that it would be better if the automation point was able to be selected before using the scroll wheel so it doesn't interfere with scrolling on the y-axis.
> This PR can be closed.

Closing this PR in favor of the other one as suggested by the author!

PR #5293, opened 2019-10-31
Fix crash due to calling QWidget::move from a non-GUI thread while ex…

Cyp:
…porting tracks.

Using signals/slots instead should be thread safe.

Crash callstack:

QWidget::move
SongEditor::updatePosition
Song::stop
Song::stopExport
ProjectRenderer::run
QThreadPrivate::start

### DIFF
--- a/src/core/Song.cpp
+++ b/src/core/Song.cpp
@@ -685,7 +685,7 @@ void Song::stop()
 				if( gui && gui->songEditor() &&
 						( tl->autoScroll() == TimeLineWidget::AutoScrollEnabled ) )
 				{
-					gui->songEditor()->m_editor->updatePosition(0);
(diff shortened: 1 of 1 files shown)


Cyp:
Bah, apparently signals are protected in old Qt 4…

lukas-w:
@Cyp Thanks for your PR. Can you provide steps to trigger the crash so we can test this?

Cyp:
It's not very reproducible — think it's only crashed 2 times so far for me — got the stack trace the second time.

The steps are to open something with lots of tracks, press Ctrl+Shift+E and export the tracks. Maybe try to mess around with the GUI while it's exporting, not sure whether that's needed. Then repeat maybe a few hundred times until it crashes.

lukas-w:
@Cyp Understood. I can't reproduce a crash, but the change makes sense and should be safe. Merging :+1:

Issue #5289, opened 2019-10-30
Drop Qt4 from Wiki


lukas-w:
Any objections to dropping all Qt4 related instructions from the wiki? Qt5 was released seven years ago, both `master` and `stable-1.2` support Qt5. I don't really see a point in continuing support for Qt4.

PhysSong:
`stable-1.2` still supports Qt4, though using it is not recommended. Since `master` stopped supporting Qt4, I think it's okay to drop Qt4-related instructions from our wiki.

lukas-w:
:+1: I'll go ahead and remove it then.

lukas-w:
Maybe we should think more about what this means for `stable-1.2` first: I'd suggest we make `WANT_QT5` default and keep a section in the wiki that tells you how to turn it off if you wish to use Qt4 instead. Sounds fine?

piece-o-ham:
The compiling page says "Qt 5.6+ is required on the master branch", so does that mean this should be closed? Is this referring to a different wiki?

lukas-w:
Yeah this was done by @tresf, see https://github.com/LMMS/lmms/pull/5294#issuecomment-569522876. Closing

PR #5294, opened 2019-10-31
Build with Qt5 by default

lukas-w:
This makes `stable-1.2` build with Qt5 by default as suggested in https://github.com/LMMS/lmms/issues/5289#issuecomment-547836729.

### DIFF
--- a/cmake/build_mingw32.sh
+++ b/cmake/build_mingw32.sh
@@ -18,6 +18,8 @@ fi
 
 if [ $QT5 ]; then
 	CMAKE_OPTS="-DWANT_QT5=$QT5 -DCMAKE_PREFIX_PATH=$MINGW $CMAKE_OPTS"
+else
+	CMAKE_OPTS="-DWANT_QT5=OFF"
 fi
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,7 +67,7 @@ OPTION(WANT_TAP		"Include Tom's Audio Processing LADSPA plugins" ON)
 OPTION(WANT_VST		"Include VST support" ON)
 OPTION(WANT_VST_NOWINE	"Include partial VST support (without wine)" OFF)
 OPTION(WANT_WINMM	"Include WinMM MIDI support" OFF)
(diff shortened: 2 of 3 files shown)


cmake/build_mingw32.sh
```diff
@@ -18,6 +18,8 @@ fi
 
 if [ $QT5 ]; then
 	CMAKE_OPTS="-DWANT_QT5=$QT5 -DCMAKE_PREFIX_PATH=$MINGW $CMAKE_OPTS"
+else
+	CMAKE_OPTS="-DWANT_QT5=OFF"
```
tresf: What does this `else` achieve?

tresf:
If we're going to toggle it on by default, wouldn't we also be removing the redundant directives from travis, or is the point here to touch as few files as possible?

cmake/build_mingw32.sh
tresf: Ok, I think I understand it... it sets the Cmake flag back to it's old default value... but then we still need "Qt5 instructions in the wiki", since this flag is needed for Windows builds, no?  I guess that's an OK compromise.  Any alternative attempt would break backwards compat on build systems.

JohannesLorenz:
@lukas-w Can you please help tres to bring this stable-1.2 issue to a merge? (or just say if you won't have time to finish this in the next weeks)

Thanks on advance.

cmake/build_mingw32.sh
lukas-w: @tresf Sorry for te very late reply. If I recall correctly, I didn't want to touch the travis scripts which all use the environment variable `QT5=True` for Qt5 and `QT5=` (unset) for Qt4. I didn't consider that `build_mingw32.sh` is not just used internally, so yeah... It'd make sense to change this to be consistent, but it would also require reversing the Travis Qt4/Qt5 env variable logic.

tresf:
I'm closing this.  Feel free to reopen if you disagree.  Reasoning, it will add confusion to the build scripts, break compatibility with existing build systems and I don't really see the value of it.

For now, the platform matrix has been simplified... In addition, I'll remove the Qt4 stuff now on each tutorial to avoid confusion.  In the rare chance someone wants to build with Qt4 still, they can either look at the wiki history or figure it out through trial and error.

>

PR #2665, opened 2016-03-11
File browser. Factory files off by one

zonkmachine:
Since cee287116bfe1fc183addc15d403c2520d72206a the last of the user presets is listed first in the `--- Factory files ---` section.

PR #1505, opened 2014-12-26
Proposed fix for 1150 Preset Browser: division between factory and user ...

curlymorphic:
Proposed fix for  #1150 Preset Browser: division between factory and user files fails if factory has folders

image

Simply added a counter to the Directory class. that is incremented for every child directory added.

Then used this counter to calc the position in tree of --Factory Presets --

Issue #1150, opened 2014-09-10
Preset Browser: division between factory and user files fails if factory has folders


DeRobyJ:
If you try to create a ZynAddSubFX preset, and you place it in the user folder, it will be shown in the browser at the end of the list, and the "--- Factory files ---" line will be after the first folder, wich is actually a factory folder.

This seems to happen because folders are always on top.
If you create a subfolder in another plugin user-preset directory (for example 3-OSC), there will be no problem. But if you create a subfolder in the factory-preset directory (lmms/data/presets/<plugin>), then you will see that folder on top of the list (as it was a user subfolder), and the last of your presets right under "--- Factory files ---"

*(edited 2014-10-03 @tresf to shrink screenshot to standard size)*
![image](https://cloud.githubusercontent.com/assets/6345473/4505732/dc8398fa-4af9-11e4-9ff0-a0a6ab24febd.png)

tresf:
Closed via #1505

PR #5296, opened 2019-11-01
Fix browser 'Factory files' delimiter position

### LINKED ISSUE #3183: Factory files separator incorrectly placed if factory directory has subdirectories.
My Samples/drums, My Samples/drumsynth and My Presets/ZynAddSubFX.
I have my own folders under samples/misc and there is no separator at all.
I'm using up to date master.

zonkmachine:
Fixes an issue where the factory file delimiter in the browser tree is placed wrong where there are user added directories. Lingering issue from #1150. Earlier fixed in #2665 and #1505.

Fixes #3183

### DIFF
--- a/src/gui/FileBrowser.cpp
+++ b/src/gui/FileBrowser.cpp
@@ -827,7 +827,6 @@ bool Directory::addItems(const QString & path )
 			{
 				addChild( new Directory( cur_file, path,
 								m_filter ) );
-				m_dirCount++;
 			}
 
 			added_something = true;


zonkmachine:
More info in comments here: https://github.com/LMMS/lmms/issues/3183#issuecomment-548638933
Review requested from the last to code around this, @curlymorphic

src/gui/FileBrowser.cpp
```diff
@@ -827,7 +827,6 @@ bool Directory::addItems(const QString & path )
 			{
 				addChild( new Directory( cur_file, path,
 								m_filter ) );
-				m_dirCount++;
```
zonkmachine: Even though the `--- Factory files ---` delimiter comes out in the right spot this way I don't find it a satisfying solution. The orphans seem to be counted twice and if this is the case we should make the logic right instead. I don't know under what circumstances it could pose a problem but the `insertChild()` function may not actually run if it finds the child in question already exists and in this case m_dirCount will still be increased. The same is probably true for `addChild()` as it's realised using `insertChild()`. I guess if this would ever happen we would see actual duplicates in the tree and this doesn't seem to be the case.

JohannesLorenz:
I'll unassign this after talking with Curlymorphic.

The reviewer also might want to check that this will not cause regression of #1150 .

src/gui/FileBrowser.cpp
JohannesLorenz: @zonkmachine What do you mean by "the orphans seem to be counted twice"?

JohannesLorenz:
FYI I'll review this now.

src/gui/FileBrowser.cpp
zonkmachine: As I remember it I had some debug statements in there while testing. Unfortunately I may have to repeat this test in order to understand my own comment.

JohannesLorenz: When will you have time for it?

zonkmachine: Maybe a week.

zonkmachine: Sorry! I tried to reproduce the debug session but failed. I retraced my steps with printf() debug statements but failed to arrive at the same conclusion. I'm currently at 'I poked that code and it works.' You deserve a better answer... :disappointed:

JohannesLorenz: Np, but can you explain if you find your solution still unsatisfying, and if yes, why you find so?

zonkmachine: I'm just unsatisfied with not knowing precisely what I've done. That's all. I knew the math was wrong and roughly where the error was occurring. I removed that line to poke the bug and it turned out to fix it.

JohannesLorenz:
After we talked about this, this was the wrong solution. I'll close this, the correct PR is #5378 .

Thanks still for the fix, feel free to reopen.

PR #5298, opened 2019-11-01
CircleCI: Test merged pull request

lukas-w:
In contrast to Travis, CircleCI only tests a PR's branch without merging it first, see https://ideas.circleci.com/ideas/CCI-I-926. With CircleCI's default behaviour, we can't know if a PR breaks tests after merging. This PR adds an init step that checks out the merged PR.

Unfortunately, this only works with PRs from forks.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,6 +42,14 @@ shared:
         mkdir -p /tmp/artifacts
         # Workaround for failing submodule fetching
         git config --global --unset url."ssh://git@github.com".insteadOf || true
+        if [[ -n "${CIRCLE_PR_NUMBER}" ]]
+        then
+            echo "Fetching out merged pull request"
+            git fetch -u origin refs/pull/${CIRCLE_PR_NUMBER}/merge:pr/merge
+            git checkout pr/merge
+        else
+            echo "Not a pull request"
+        fi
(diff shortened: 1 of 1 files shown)


Reflexe:
Mm but what about rebase?

lukas-w:
What about it? I don't understand the question.

Reflexe:
> What about it? I don't understand the question.

I couldn't  find any documentation for `refs/pull/id/merge`, however, if I understand correctly, it assumes a regular merge using a merge commit. However, rebasing instead of merging could have other effects (especially when merge is not possible).

lukas-w:
> However, rebasing instead of merging could have other effects

Shouldn't they have the same result? I'm not sure what the `merge` ref points to when there are merge conflicts though. I've tested a few PRs with conflicts in this repo and the ref pointed to a merge of the the PR and a less recent target branch commit without conflicts.

Reflexe:
Yes, sorry. You are probably right about that.

lukas-w:
I understand your concern though, GitHub's merge ref is undocumented and we can't know exactly how it will behave. I suggest to merge this and just see how it goes. It's possible we have to add a check to skip the merge checkout if the ref doesn't exist.

PR #5300, opened 2019-11-01
Fix remote plugin crash reading parameters from Grooove plugin

DomClark:
Reported by `@Pigroy#0913` on Discord. The crash occurs when performing an action that reads parameters from the plugin, such as loading a project, or opening the control window. [Grooove](https://bedroomproducersblog.com/2016/12/23/grooove-bpb/) writes a 65-byte string in response to `effGetParamName`, but we only provide a 32-byte buffer, resulting in stack corruption.

### DIFF
--- a/plugins/vst_base/RemoteVstPlugin.cpp
+++ b/plugins/vst_base/RemoteVstPlugin.cpp
@@ -1072,7 +1072,7 @@ void RemoteVstPlugin::getParameterDump()
 
 	for( int i = 0; i < m_plugin->numParams; ++i )
 	{
-		char paramName[32];
+		char paramName[65];
 		memset( paramName, 0, sizeof( paramName ) );
 		pluginDispatch( effGetParamName, i, 0, paramName );
 		paramName[sizeof(paramName)-1] = 0;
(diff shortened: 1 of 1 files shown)


noahb01:
is there a better way to do this than just hardcoding a value? Is anything stopping a new plugin from coming along and writing a 66-byte string, causing this issue to reappear?

DomClark:
> is there a better way to do this than just hardcoding a value?

Not really - the caller has to provide raw memory, so this has to be allocated in advance (barring some silly and complicated tricks with guard pages).

> Is anything stopping a new plugin from coming along and writing a 66-byte string, causing this issue to reappear?

When I wrote "writes a 65-byte string", it was inaccurate use of terminology. It writes 65 bytes of memory, no matter how long the string actually is. The extra memory is filled with zeros. 65 bytes isn't entirely arbitrary; it's enough for a 64-byte string (a nice power of 2) followed by a null terminator. The actual limit is specified by the VST API; reading https://www.kvraudio.com/forum/viewtopic.php?t=319195, it looks like it's supposed to be 8 bytes + null terminator, although nobody sticks to that - even Steinberg in their VST3->VST2 wrapper use a 32-byte limit. For now, I think 64+null will suffice; all plugins that used to work still will, and now this new one does too. If someone finds a plugin one day that has an entire essay for a parameter name, we can bump it up again to 129 or 257 or something.

lukas-w:
> If someone finds a plugin one day that has an entire essay for a parameter name, we can bump it up again to 129 or 257 or something.

Why not just do this now so we don't have to in the future? We should be safe with e.g. 256 bytes, it's not like we can't afford the memory.

DomClark:
Done. It's at 256 now.

DomClark:
Merge?

PR #5304, opened 2019-11-04
Add missing ¹⁄₉₆ quantization.

Cyp:


PhysSong:
I think `AutomationEditor::setQuantization()` will be broken with this changes while `PianoRoll::quantization()` doesn't.

Cyp:
Eeek, sorry about that.

PhysSong:
Looks good, but I don't think the stable branch needs these changes.
@Cyp Sorry for the delay.

PhysSong:
BTW, is it okay to keep `Quantizations` in `MidiTime.h`?

Veratil:
> 
> 
> BTW, is it okay to keep `Quantizations` in `MidiTime.h`?

I feel if we need to use quantizations in other areas keeping it in that file will be beneficial for reducing code duplication. Allow a model to use whatever it wants, but have all the options available. Since the Automation Editor also uses the values, we shouldn't have them in both places.

PhysSong:
I also agree with reducing code duplication. What I mean is that `Quantizations` is for the GUI whereas `MidiTime.h` is independent of the UI.

Veratil:
> 
> 
> I also agree with reducing code duplication. What I mean is that `Quantizations` is for the GUI whereas `MidiTime.h` is independent of the UI.

🤦 gotcha! What would be the better choice in your opinion?

PhysSong:
The problem is, I don't know either. `Editor.h` is a candidate, but I'm not sure if it's the best place.

Cyp:
Rebased and fixed merge conflicts. Moved `Quantizations` from `MidiTime.h` to `Editor.h`, and hope it's ok there.

PR #5306, opened 2019-11-05
Fix crash on deleting instrument with controller connections on knobs

### LINKED ISSUE #5106: Crash on deleting instrument tracks with knobs connected to controllers
Reported by @DouglasDGI on Discord:
> When you delete an instrument track with any knob connected to any controller, LMMS crashes (in `Knob::friendlyUpdate`, according to gdb)

It's related to the stupid `InstrumentTrackWindow` caching system.
- The `InstrumentTrackView` destructor calls 
`InstrumentTrackView::freeInstrumentTrackWindow()`, and it calls `InstrumentTrackWindow::updateInstrumentView()`. the function deletes and then *re-instantiates* `m_instrumentView`. It results in deleting the instrument before deleting the view.
- An `InstrumentTrackWindow` instance may still exist after the corresponding track is deleted, making the knob still updated after the model is deleted. Since the mixer thread handles value changes from controllers, the slot might be invoked after the sender(model) is deleted.

It seems like the current caching system is doing unnecessary jobs and causes some bugs like #3623 as well as this one.

Anyway, `Knob::friendlyUpdate` should check for the model and do nothing if it's null or already deleted.

PhysSong:
Fixes #5106.

### DIFF
--- a/src/gui/widgets/Knob.cpp
+++ b/src/gui/widgets/Knob.cpp
@@ -803,9 +803,9 @@ void Knob::enterValue()
 
 void Knob::friendlyUpdate()
 {
-	if( model()->controllerConnection() == NULL ||
+	if (model() && (model()->controllerConnection() == NULL ||
 		model()->controllerConnection()->getController()->frequentUpdates() == false ||
-				Controller::runningFrames() % (256*4) == 0 )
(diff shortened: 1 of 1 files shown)


PhysSong:
Merging it because this is a simple nullity check.

PR #1710, opened 2015-01-28
Proposed fix for #1595 Instrument track activity LED lights when muted.

curlymorphic:
Proposed fix for #1595 Instrument track activity LED lights when muted.

The issue mentioned using blue to show muted, I chose to use red, because it keeps inline with the rest of the lmms gui.

PR #5307, opened 2019-11-05
Sampletrack activity indicator

serdnab:
My proposed implementation for the Sample Track activity indicator led.
When a sample track is playing samples it emits the corresponding signal to `FadeButton` that activates the button animation.
When It ceases playing samples it does the same with the signal `notPlaying()`.
This is a work in progress, the behaviour is missing when the track is muted (maybe the same as `InstrumentTrack`, that is, the button fades in to another color), and the action when the led is pressed.
Suggestions for these?

### DIFF
--- a/include/FadeButton.h
+++ b/include/FadeButton.h
@@ -47,6 +47,7 @@ public:
 public slots:
 	void activate();
 	void noteEnd();
+	void notPlaying();
 
 
 protected:
--- a/include/SampleTrack.h
+++ b/include/SampleTrack.h
@@ -29,6 +29,7 @@
 #include <QLayout>
 
 #include "AudioPort.h"
+#include "FadeButton.h"
 #include "FxMixer.h"
 #include "FxLineLcdSpinBox.h"
 #include "Track.h"
@@ -161,6 +162,10 @@ public:
 		return "sampletrack";
 	}
 
+signals:
+	void playing();
+	void notPlaying();
+
 public slots:
 	void updateTcos();
 	void setPlayingTcos( bool isPlaying );
@@ -171,6 +176,7 @@ private:
 	FloatModel m_panningModel;
 	IntModel m_effectChannelModel;
 	AudioPort m_audioPort;
+	bool m_wasPlaying;
 
 
 
@@ -230,6 +236,7 @@ private:
 	SampleTrackWindow * m_window;
 	Knob * m_volumeKnob;
 	Knob * m_panningKnob;
+	FadeButton * m_activityIndicator;
 
 	TrackLabelButton * m_tlb;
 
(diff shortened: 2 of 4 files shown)


Spekular:
Here's the relevant change for the instrument track indicators: https://github.com/LMMS/lmms/pull/1710/files

I think it'd be ideal if the behavior could be moved into a parent class (Track.cpp?). Maybe move the code up and wrap it in a check to see if the track has an indicator?

serdnab:
> Maybe move the code up and wrap it in a check to see if the track has an indicator?

I'm on it.

serdnab:
Done. Moved the activity indicator mute code from `InstrumentTrackView` to the parent class (`TrackView`) and set the mute code for `SampleTrackView`.

Spekular:
Sorry for the super delayed response. Testing results:
- Doesn't react to empty sample clips (good)
- Flashes fully lit when it enters a sample clip after not being on one, but overlapping clips don't re-trigger the indicator
- With longer clips, spends most of its time half lit (same intensity as muted tracks trigger color) since it only reacts to on/off & not volume
- Activates even on totally silent clips (not too surprising given previous findings)
- If a clip is extended, indicator stays lit even when entering the part with no audio. If the clip is cropped from the left so the audio is excluded, the empty segment doesn't trigger the indicator. If the clip is extended leftwards, the indicator only triggers once it reaches the actual audio.

Since reacting to volume would require more significant changes to FadeButton, I'm fine with this behavior as an improvement over having no indicator at all. The one thing I would want fixed, if possible, is that the indicator stays when entering an empty section of a clip (if you've extended the clip).

As for the code, my gut instinct is that it could be simplified a bit more. I'd like to take a closer look at it before making any suggestions though, and unfortunately that could take me a few days to get around to.

Spekular:
I had a go at refactoring this, after fixing the merge issues. I tried to introduce fewer new members and prefer positive/present (isPlaying) over negative/past (wasPlaying, notPlaying). My version is here: https://github.com/LMMS/lmms/compare/master...Spekular:SampleIndicator

I'm not sure what the best way to handle this PR is, but just in case: @serdnab do you have anything against it if this is merged in a separate/new PR?

Spekular:
I also did some additional testing, because I realized sample tracks in the BB editor were a special case. For a while, my version made the indicators stick on forever after one trigger, which is a problem that the original PR didn't have. However, I managed to fix it, so both PRs are now the same from a user perspective (as far as I can tell).

Spekular:
Closing, replaced by #5477.

serdnab:
@Spekular  
> do you have anything against it if this is merged in a separate/new PR?

ok, no problem

PR #5311, opened 2019-11-06
Simplify FadeButton update

husamalhomsi:
I have no idea why the deleted headers were created in the first place.


PhysSong:
You didn't remove `#include "update_event.h"` in `src/gui/widgets/FadeButton.cpp`.

Sawuare:
Fix commited. Thanks!

DomClark:
> I have no idea why the deleted headers were created in the first place.

`QEvent::User` isn't really a type of event on its own; it's just a constant to indicate where IDs for user events can begin. Using it directly is no different to using a magic number (it's just a magic number offset, equivalent to the implicit magic number `QEvent::User + 0`), and indicates no meaning as to what the event actually does. (What does `QApplication::postEvent(this, new updateEvent());` do? It posts an update event. What does `QApplication::postEvent(this, new QEvent(QEvent::User));` do? I certainly haven't a clue.)

Moreover, can you see where else this event is used, e.g. who's listening for it? The only place showing up for me in GitHub code search is the place modified in this PR, so it might be possible to get rid of it altogether.

Sawuare:
> can you see where else this event is used, e.g. who's listening for it?

`FadeButton` only.

> so it might be possible to get rid of it altogether.

Yes.

DomClark (approved):
It seems that the original reason for the event was to call `update()` from the UI thread, but since `activate()` and `noteEnd()` are now only called via automatic signal/slot connections, they run on the UI thread anyway. Thus I think the removal is safe.

Code looks good, but I haven't tested it. If it still works properly, I think this should be merged. Sorry I took so long to get to this review.

(Not too important, but there's an extra blank line added before `paintEvent`. I realise you're trying to maintain a consistent 4 lines between each function, but there are a couple of other irregular gaps in this file. I think they should either all be fixed, or none fixed, and am slightly in favour of the latter to keep unrelated changes out of this PR.)

PR #5315, opened 2019-11-09
Automation Editor - Prevent dragging last point when drawing line

### LINKED ISSUE #4040: Automation Editor - draw line function selects last point drawn for a drag
A minor nuisance in the draw line function. When you draw a line the last automation point drawn is selected and ready to be dragged. This is a bit unintuitive and can lead to some unexpected results which is especially prone to happen if you work fast. The resulting drag action gives the illusion that it's the line draw action that is buggy with skipped points and/or nudged levels.

drawlineagain

A - Expected result from a line draw.
B - Last point moved.
C - Last point level nudged.

To reproduce:
Method 1 - Draw a line and while keeping the left mouse button down, move the mouse around.
Method 2 - Press <Shift> and draw lines fast moving the mouse around. You will find gaps in the lines from where a point has been moved after the line was drawn.

tecknixia:
Resolves #4040 

> When you draw a line the last automation point drawn is selected and ready to be dragged. This is a bit unintuitive and can lead to some unexpected results which is especially prone to happen if you work fast.
-zonkmachine 

After a line is drawn, when the user still has their finger on the left mouse button, and the mouse is moved, it is considered a drag, since the button is down when in edit mode.  Using `applyDragValue` and setting `m_action=NONE` after the `drawLine` command doesn't work, because as soon as the mouse is moved, the point is dragged.  Bypassing the drag with the shift key doesn't work, because it's set in `mousePressEvent` and as soon as the shift key is let go, a new line is drawn to the the cursor.

It seems that the main problem is the fact that the point gets dragged when the mouse button is down, so this fix disables the drag in `mouseMoveEvent` by replacing `mouseEvent->buttons() & Qt::LeftButton` with a variable called `m_mouseDownLeft`.  This way the variable can be set to `false` right after `drawLine` is called, and will return to `true` when the mouse button is clicked again.

-  create `m_mouseDownLeft`
-  set it to true in `mousePressEvent` when it's pressed
-   set it to false right under the `drawLine` call in `mousePressEvent`
-   set it to false again when released in `mouseReleaseEvent`
-   replace the first `mouseEvent->buttons() & Qt::LeftButton` with `m_mouseDownLeft` in `mouseMoveEvent` where it checks for condition of the left mouse button and `DRAW` mode (before the other `drawLine` call)

### DIFF
--- a/include/AutomationEditor.h
+++ b/include/AutomationEditor.h
@@ -240,6 +240,7 @@ private:
 
 	EditModes m_editMode;
 
+	bool m_mouseDownLeft;
 	bool m_mouseDownRight; //true if right click is being held down
 
 	TimeLineWidget * m_timeLine;
--- a/src/gui/editors/AutomationEditor.cpp
+++ b/src/gui/editors/AutomationEditor.cpp
@@ -96,6 +96,7 @@ AutomationEditor::AutomationEditor() :
 	m_y_delta( DEFAULT_Y_DELTA ),
 	m_y_auto( true ),
 	m_editMode( DRAW ),
+	m_mouseDownLeft(false),
 	m_mouseDownRight( false ),
 	m_scrollBack( false ),
 	m_barLineColor( 0, 0, 0 ),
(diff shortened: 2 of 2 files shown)


zonkmachine:
Taking the discussion initiated in [`#5316` (here](https://github.com/LMMS/lmms/pull/5316#issuecomment-557491691)) into consideration, I still think this PR should be merged.

zonkmachine:
@tecknixia Thanks for fixing this!

PR #5328, opened 2019-11-21
Vectorscope

he29-net:
Hi,

here is a native Vectorscope plugin. It uses the new lockless ring buffer to get out of the way of the audio thread as soon as possible, offers a mode switch for high performance vs. high quality operation and is resizable and HiDPI friendly.

It can be used to determine the stereo width and balance of the signal, to spot DC balance issues,
 

to see a limiter in action, or even to draw images and watch stuff like [this](https://oscilloscopemusic.com/watch.php#album)
 

(and probably for a variety of other things).

This PR should be pretty much finished / ready for review, but if you think it is missing some essential must-have functionality, let me know and I can look into that.


DouglasDGI:
Works perfectly for me, and given it doesn't change any existing files (except CMakeLists.txt of course) I'd say it's ready for merge as well.

Though, one feature request... perhaps you could allow the user to choose the color of the vectorscope?  Qt comes with a color chooser (which is used when coloring B/B tracks) which is quite easy to make use of in code.

plugins/Vectorscope/VectorView.cpp
```diff
@@ -0,0 +1,284 @@
...
+		// Persistence value specifies fraction of light intensity that remains after 10 ms.
+		// → Compensate it based on elapsed time (exponential decay).
+		const float persist = log10(1 + 9 * m_controls->m_persistenceModel.value());
+		const float persistPerFrame = pow(persist, elapsed / 10.f);
+		// Note that for simplicity and performance reasons, this implementation only dims all stored
+		// values by a given factor. A true simulation would also do the inverse of desaturation that
+		// occurs in high-intensity traces in HQ mode.
+		for (std::size_t i = 0; i < useableBuffer; i++)
+		{
+			m_displayBuffer.data()[i] = m_displayBuffer.data()[i] * persistPerFrame;
```
PhysSong: I think using a `*=` will look better.

plugins/Vectorscope/VectorView.cpp
```diff
@@ -0,0 +1,284 @@
...
+	// Draw new points on top
+	float left, right;
+	int x, y;
+
+	const bool logScale = m_controls->m_logarithmicModel.value();
+	const unsigned short activeSize = hq ? m_displaySize : m_displaySize / 2;
+
+	// Helper lambda functions for better readability
+	// Make sure pixel stays within display bounds:
+	auto saturate = [=](short pixelPos) {return fmax(fmin(pixelPos, activeSize - 1), 0);};
```
PhysSong: FYI, you can use `qBound()` for this purpose. Also, remind that `fmax` and `fmin` return `double` in this case. Maybe you'd better using `std::max` and `std::min` instead.

PhysSong: You may take care of the type of integer literals as well.

plugins/Vectorscope/VectorView.cpp
```diff
@@ -0,0 +1,284 @@
...
+	// Take existing pixel and brigthen it. Very bright light should reduce saturation and become
+	// white. This effect is easily approximated by capping elementary colors to 255 individually.
+	auto updatePixel = [&](unsigned int x, unsigned int y, QColor addedColor)
+	{
+		QColor currentColor = ((QRgb*)m_displayBuffer.data())[x + y * activeSize];
+		currentColor.setRed(fmin(currentColor.red() + addedColor.red(), 255));
+		currentColor.setGreen(fmin(currentColor.green() + addedColor.green(), 255));
+		currentColor.setBlue(fmin(currentColor.blue() + addedColor.blue(), 255));
```
PhysSong: Here as well.

plugins/Vectorscope/VectorView.cpp
```diff
@@ -0,0 +1,284 @@
...
+		// High quality mode: check distance between points and draw a line.
+		// The longer the line is, the dimmer, simulating real electron trace on luminescent screen.
+		for (std::size_t frame = 0; frame < frameCount; frame++)
+		{
+			// Scale left and right channel from (-1.0, 1.0) to display range
+			if (logScale)
+			{
+				// To better preserve shapes, the log scale is applied to the distance from origin,
+				// not the individual channels.
+				const float distance = sqrt(pow(inBuffer[frame][0], 2) + pow(inBuffer[frame][1], 2));
```
PhysSong: It seems like manual multiplication is faster than using `pow()` in some cases. It's not the case when using GCC with `-ffast-math`, but I don't know about other compilers.
See: https://baptiste-wicht.com/posts/2017/09/cpp11-performance-tip-when-to-use-std-pow.html

he29-net:
> perhaps you could allow the user to choose the color of the vectorscope?

Hmm, why not. It should be relatively easy, the hardest thing will be probably finding a good place for the color picker control. :) I originally used the same blue as in the Spectrum Analyzer (green is not among my favorites), but in this case there is no technical reason for that specific color so I thought the standard "LMMS green" should be preferred to fit the theme better.

plugins/Vectorscope/VecControls.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+
+// Create the VecControlDialog widget which handles display of GUI elements.
+EffectControlDialog* VecControls::createView()
+{
+	return new VecControlsDialog(this);
+}
+
+
+void VecControls::loadSettings(const QDomElement &_this)
```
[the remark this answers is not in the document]
he29-net: Ah, thanks. This is copied from the spectrum analyzer, which is copied form the equalizer, which was probably copied from somewhere else.. :D

Umcaruje:
This looks great! I would suggest renaming it to something like "Stereo Scope" or "Stereo Tool" so it will be easier for beginners to understand what the plugin is all about.

plugins/Vectorscope/VectorView.cpp
```diff
@@ -0,0 +1,284 @@
...
+	m_oldY(m_displaySize / 2)
+{
+	setMinimumSize(200, 200);
+	setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Expanding);
+
+	connect(gui->mainWindow(), SIGNAL(periodicUpdate()), this, SLOT(periodicUpdate()));
+
+	m_displayBuffer.resize(sizeof qRgb(0,0,0) * m_displaySize * m_displaySize, 0);
+
+	#ifdef VEC_DEBUG
```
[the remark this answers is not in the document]
he29-net: Interesting, I'm not sure if I saw that before. I usually treat preprocessor stuff just as any other code and indent according to its function, so that the indentation a) does not break the flow of surrounding code and b) separates the conditional code so that, well, it is obvious that it runs only under certain conditions.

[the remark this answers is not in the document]
PhysSong: There are several ways of indenting such directives.
- Never indent it
- Indent it except the `#` at the beginning
- Indent it, but don't indent anything in the surrounded block
- Treat it as plain code fragment

plugins/Vectorscope/VectorView.cpp
he29-net: > FYI, you can use qBound() for this purpose

Ah, nice. I was searching for something like that in the `std` library and `cmath` and did not find anything. 

What do you mean about the integers? I notice that some may be too big; did you mean that?

PhysSong: I mean something like `(short)0` vs. `0`.

he29-net:
> I would suggest renaming it to something like "Stereo Scope" or "Stereo Tool"

I am not sure if that is wise; vectorscope is an industry standard name for this "instrument" so this way it would be easier to search for tutorials and explanations about it. It also says "A stereo field visualizer." in the plugin description, so that should help to explain what the plugin does to newcomers.

plugins/Vectorscope/VectorView.cpp
he29-net: You are right; I just checked at https://gcc.godbolt.org/ and while trying a few compilers none of them ended up even remotely close to the few instructions that a simple multiplication produced.

Umcaruje:
>It also says "A stereo field visualizer." in the plugin description, so that should help to explain what the plugin does to newcomers.

Ah, that can work too. Thanks!

plugins/Vectorscope/VectorView.cpp
he29-net: Correction: it seems that with -O3 the results are generally much better. Sometimes even exactly the same. I tried to see what would happen if I used a `square` lambda that implemented the multiplication, but in general it seems to make things worse. So a simple `num * num` would be probably the safest bet for performance (even though it makes the line a bit too long).

he29-net:
I added the color change support, but I did not want to clutter the interface so it is triggered by a double-click on the vector display.

A right-click context menu may have been better, but it would take much more time to implement (I don't even know where to start, or what kind of custom menus LMMS uses) and it would only have one item, which is kind of silly.

plugins/Vectorscope/VectorView.cpp
Sawuare: How about `hypot`?
https://en.cppreference.com/w/cpp/numeric/math/hypot

Sawuare:
I think the vectorscope graph color should be a theme setting, not a project setting.
Like the oscilloscope graph color.

DouglasDGI:
Given it's a common hobby for people to make shapes in vectorscopes with their music (a lot of people are fascinated with this in the LMMS server), I think it would be nice to be able to change the color without changing your theme.
Having the theme simply choose the _default_ color would be a good compromise.

RoxasKH:
And this is awesome, i love it!

he29-net:
I am not a fan of pulling the color from a theme (even as a default), because you can literally get anything. Like black, for example. So it would need to include various sanity checks to make sure the color will produce a good result.

Even many "normal" colors may not be suited: if a color is too saturated, the desaturation that makes overlapping traces appear to "glow hotter" will be barely noticeable or won't occur at all. So some themes could basically break a large part of the HQ mode functionality and the user would not even know.

So I would rather keep it as a sort of Easter egg, so that users have to intentionally break it themselves. :)) A balance between flexibility and good default behavior.

plugins/Vectorscope/VectorView.cpp
he29-net: Hard to say. GCC apparently produces just 8 instructions with -O3 for manual `sqrt(num*num)` and a function call to who knows what when `hypot` is used (insides of the function were not shown). So if anything, it is probably slightly worse performance-wise.
MSVC produced a call either way (to sqrt or hypot) so the basic overhead is the same and then it would depend on what exactly is inside. Overall it is probably safer to just use `sqrt(num*num)`, although the code does not look as nice..

Sawuare:
Most of LMMS is theme-able. It's the themer's responsibility to choose sane values. We could write a note in the theme file.
When there's a color picker, there's a need to distinguish objects (e.g., BB blocks). That's not the case here, though.

plugins/Vectorscope/VectorView.cpp
```diff
@@ -0,0 +1,294 @@
...
+					 QString("Exec avg.: ").append(std::to_string(m_executionAvg).substr(0, 5).c_str()).append(" ms"));
+#endif
+}
+
+
+// Periodically trigger repaint and check if the widget is visible
+void VectorView::periodicUpdate()
+{
+	m_visible = isVisible();
+	if (m_visible) {update();}
```
Sawuare: According to https://doc.qt.io/qt-5/qwidget.html#update, there is no need to check visibility before calling `update()`, or am I missing something?

he29-net: This is supposed to save performance when the widget is not visible -- unlike other effects there is no need to keep processing new samples when the window is closed. I am not 100 % sure how it interacts with the `update()`, I just re-used some code that originally came from the EQ plugin, where it was specifically added to save performance. So I just _assumed_ it is somewhat important, but I did not dig deeper.

he29-net:
> It's the themer's responsibility to choose sane values.

OK, I thought we are talking about a random "signature color" pulled from the theme. If there was to be a specific theme value for the Vectorscope and a proper description for theme makers, it could work.

Although it still brings this from a small, easy to add feature to a fully supported "thing". And I never even _used_ a LMMS theme, let alone developed a feature that uses the theme.. engine or whatever is it called?

CYBERDEViLNL:
> Although it still brings this from a small, easy to add feature to a fully supported "thing". And I never even _used_ a LMMS theme, let alone developed a feature that uses the theme.. engine or whatever is it called?

Theming is done through stylesheets and they can access `Q_PROPERTY`'s.

Example:
## .h

```cpp
Q_OBJECT
Q_PROPERTY( QColor scopeColor READ scopeColor WRITE setScopeColor )

public:
    QColor scopeColor() const;
    void setScopeColor( const QColor & c );

private:
    QColor m_scopeColor;
```

## .cpp

```cpp
QColor YourClass::scopeColor() const
{ return m_scopeColor; }

void YourClass::setScopeColor( const QColor & c )
{ m_scopeColor = c; }
```

here `m_scopeColor` is the color you get from the sylesheet.

plugins/Vectorscope/VectorView.cpp
Sawuare: It is not `sqrt(num*num)`, it is `sqrt(num1*num1 + num2*num2)`.

`hypot(num1, num2)` beats `sqrt(num1*num1 + num2*num2)` on all optimization levels on GCC.

https://godbolt.org/z/wZdQ2-

he29-net: Ouch, damn, I mixed up the code with my previous experiments. Thanks, I'll look at it once more.

he29-net:
@CYBERDEViLNL hmm, that does not look _that_ bad. What happens when the color isn't defined by the theme? (Which is the case for all themes currently in existence.)

he29-net:
> Note, I wonder if an input gain control might be practical. If your instrument is not that loud, the interesting parts in the center are currently small and there is a lot of space wasted.

That's not a bad idea. This was initially supposed to be solved by the log scale, but I sometimes it is not quite enough. I'll try and see if I can simply map it to mouse wheel; that would not need a space for a full extra control and it is common for zooming in many other applications.

plugins/Vectorscope/VectorView.cpp
he29-net: Hm, I still cannot confirm it. Each implementation still has a function call (either for sqrt or for hypot) and since they are inside a library, I can't get it to show the instructions.

he29-net: I decided to leave it as it is; if anything, google searches turn out questions about why is `hypot` _slower_ than doing the same thing by hand, and after implementing the zoom everything fits into one line again, so no issue with readability either.

he29-net:
I tried to get the default color from the theme as suggested, but it did literally nothing. Maybe I just did not add the new style sheet record correctly, or maybe one of a dozen other things went wrong. But it does not even print any errors, so I have no clue about what's happening.

I.e. it turned out exactly as I expected. So if anyone needs a theming support, you can make a follow-up PR. :stuck_out_tongue:

So at this point I think I am done with the extra features and the review can continue.

CYBERDEViLNL:
> @CYBERDEViLNL hmm, that does not look _that_ bad. What happens when the color isn't defined by the theme? (Which is the case for all themes currently in existence.)

It probably will be black. You can use `m_scopeColor.isValid()` to check if it is set (properly) by the currently active theme. A note: `setScopeColor` will be called after your class initiated.

CYBERDEViLNL:
Found 2 little issues:
 1. Play a note (keep playing) and open the color chooser, the note will be stuck.
 2. Open the color chooser and cancel the action, color will now be black.

he29-net:
I reproduced the second issue and have a fix ready, but I can't reproduce the first one. What do you mean by "stuck"? Does the sound keep playing? Or does the color not change?

CYBERDEViLNL:
> I reproduced the second issue and have a fix ready, but I can't reproduce the first one. What do you mean by "stuck"? Does the sound keep playing? Or does the color not change?

Yes the note keeps playing.

1. Play a note (keep playing)
2. Open color chooser (now stop playing the note)
3. The note keeps playing. (You have to play/press the note again to stop it from playing)

he29-net:
@CYBERDEViLNL Ah, I assume you are using a physical MIDI keyboard. I don't have one available right now, but I could test it tomorrow. I just tried to add a note into piano roll and opened the color picker after it started playing and before it ended, with no issues. Same goes for ordinary keyboard, the note stops right when the piano roll or instrument window loses focus, before the color picker appears.

With a MIDI keyboard (or some other external input?) my guess is that the color picker window has focus when you release the note and therefore gets the release event. But I don't have a slightest idea if anything can be done about that.

he29-net:
Hah, I managed to reproduce it with ordinary keyboard after fiddling around a bit more. And the same thing happens when you open the Setting dialog while playing a note. So I'm not sure I can do anything about that, it seems to be an inherent problem of Qt and the way it manages inputs or something..

CYBERDEViLNL:
@he29-net No I'm sorry; I had to be more specific. MIDI keyboard works just fine, normal computer keyboard doesn't. The color dialog indeed steals focus and the instrument window never receives `keyReleaseEvent`. LMMS probably needs a sub-class of the `QColorDialog` to handle this.

CYBERDEViLNL:
It's the same issue as https://github.com/LMMS/lmms/issues/4892 ; I looked at how https://github.com/LMMS/lmms/pull/5202 handles it.

This works for me (it also tackles the black color on cancel issue):

### .h
```cpp
#include <QColorDialog>
#include <QApplication>
#include <QKeyEvent>

class ColorChooser: public QColorDialog
{
public:
	ColorChooser(QWidget * parent): QColorDialog(parent) {};

protected:
	void keyReleaseEvent(QKeyEvent *event) override
	{
		QKeyEvent ke(*event);
		QApplication::sendEvent(parentWidget(), &ke);
	}
};
```

### .cpp
```cpp
void VectorView::mouseDoubleClickEvent(QMouseEvent *event)
{
	ColorChooser *colorDialog = new ColorChooser(this);
	if (colorDialog->exec())
	{
		m_controls->m_colorFG = colorDialog->currentColor();
	}
}
```

he29-net:
@CYBERDEViLNL Thanks, I'll try it out and add the class to the `include/` folder (it's so tiny I don't see much sense in splitting the implementation into a `cpp` file).

he29-net:
Hm, strangely enough it does not work for me, the note is still stuck even when using the "forwarding" class. I even tested it on two machines, because on my main the `QColorDialog` seems completely broken (I'm just getting a blank window with an array of invisible buttons).

he29-net:
Well, since the change makes sense and does not _break_ anything for me, I pushed the commit anyway. Maybe it fixes the issue at least on some configurations.

@CYBERDEViLNL I entered you as the author in the license header of `ColorChooser.h`; if you are all right with that, could you update the info? I just used your nickname and a dummy mail as a placeholder.

CYBERDEViLNL:
> Hm, strangely enough it does not work for me,

It doesn't work with the `::getColor` method, you need to use `exec` as in the example.

> I entered you as the author in the license header of ...

Yes that is awesome :-)


I created a PR (to your fork) with some changes to the ColorChooser.

he29-net:
> It doesn't work with the ::getColor method, you need to use exec as in the example.

Hm, nope.. :D Still broken for me exactly the same way as before. I would be surprised it that made a difference, `getColor` should produce the exact same dialog as `exec`..

Thanks for the PR; I thought you would just "suggest a change" here, using the review workflow (assuming you can, I'm not sure if that option is open to everyone).

CYBERDEViLNL:
> Hm, nope.. :D Still broken for me exactly the same way as before. I would be surprised it that made a difference, getColor should produce the exact same dialog as exec..

I'll guess it doesn't work with `::getColor` because it's a static function. What version of Qt and what OS do you have? I tried with Qt version `5.13.2` and `5.11.3`, both on Linux.

>  I thought you would just "suggest a change" here, using the review workflow (assuming you can, I'm not sure if that option is open to everyone).

All right I could do that in the future if it's preferred.

he29-net:
> I'll guess it doesn't work with ::getColor because it's a static function.

That could make some sense (maybe, I'm not all that sure about C++ stuff that I don't use regularly in weird scenarios).

Both my tests were with Qt 5.11.3 on Debian (which now confuses me even more, considering that one install has a broken color dialog and one is working fine -- I thought it was because of version difference). I now also tried the AppImage build, which uses Qt 5.6.3 -- also does not work (and the dialog used looks different again, but working). It could be a window manager thing; I'm using Xfce 4.12.

> All right I could do that in the future if it's preferred.

I don't mind either way, I just thought it would be faster than making an extra branch, uploading it and opening the PR. :)

CYBERDEViLNL:
All right; The `Qt 5.11.3` test I did was also on Debian with Xfce `4.12`.

Hmm I also tested the AppImage, the QColorDialog is indeed diffrent and can confirm that the instrument doesn't receive the `keyReleaseEvent`.

he29-net:
Apart from the color dialog focus stealing issue, there is another potential problem. A user on Discord wanted to try and test the Vectorscope, so I directed them to the Windows CI builds, but LMMS apparently crashed with:
image

This was the MSVC build. I tried to reproduce it but I was able to open both LMMS and the Vectorscope and they both worked fine. But when I tried to add the Spectrum Analyzer, I got the same error message. A quick google of the source code got me nowhere (there was no assert on that line, so perhaps it is a different version).

I also tried the mingw version and it is similar, but it simply segfaults instead of giving the error message.

Is there a way to instruct the CI system to make a Windows build with debug symbols so I could run it in a debugger? Or does anyone have any other idea what could be the problem / how to debug that without setting up all the billion things needed for building on Windows?

--
As for the focus stealing / stuck notes: I think it is a "dead end" issue -- there is no obvious way to solve it and it happens with various other windows as well, so it not exactly "new".

It also requires the user to change the color while holding a key on the (non-MIDI) keyboard, which could be likely to occur when testing various colors, but not in a normal workflow. So I suggest we keep the current version with CYBERDEViLNL's improvement that fixes it at least in some cases, and leave it at that, otherwise this PR will probably go nowhere..

PhysSong:
> A quick google of the source code got me nowhere (there was no assert on that line, so perhaps it is a different version).

https://github.com/qt/qtbase/blob/de8538966f51cb1589bfc4032be99b266707e26c/src/corelib/thread/qmutex.cpp#L606
The comment seems to say it's trying to unlock a mutex that isn't locked.

he29-net:
>The comment seems to say it's trying to unlock a mutex that isn't locked.

Oh, thanks. I gave up the search when the first version of the file did not have the line and did not even look like it would be useful.

This should help, I think the Analyzer does not use explicit `unlock()` without `QMutexLocker` that often.

he29-net:
I added some debug outputs around the three occurrences so I can test it in the afternoon, but I did not find anything suspicious -- all unlocks should only ever occur after lock has been acquired, there is not even any branching in the way. So there is also a small possibility that it's something else.

he29-net:
Nice, the mingw version does not crash anymore. Turns out the `useless_lock` needed by `QWaitCondition` is not so useless. I was using `QMutexLocker`s in all the places before, so when I needed to use an ordinary `QMutex` for the wait condition, I had a brain fart forgot I have to lock it. And apparently it only caused problem on Windows. o_O

This also goes to show that when reviewing PRs, we should try to get testers for all the "other" platforms, since most devs only use Linux. ^^

PhysSong:
I got a similar crash with the new spectrum analyzer. I also confirmed that 51b01780d33b7d4a5a49406dcf1c2fc9b2e0f3f2 fixes the crash as well.

he29-net:
Yeah, this crash only occurred in the Spectrum Analyzer; the Vectorscope wasn't affected because it does not need notification for new data and therefore avoided the `QWaitCondition` code completely. I just didn't want to make an extra PR for a tiny bugfix like that, so I added it here.

PhysSong:
> I just didn't want to make an extra PR for a tiny bugfix like that

However, the bug seems to crash LMMS on opening some projects containing spectrum analyzers. So, I'm suggesting you do that.

he29-net:
@PhysSong Since the review was finished and all bugs I could find fixed, I did not expect this PR would be stuck for another month.

I will extract the fix when I get home tonight if it is needed.

PhysSong:
@he29-net Do you think it's ready for review, or probably even for merge?
I think we don't need to spend much time here because it won't break anything that worked before.

he29-net:
@PhysSong Yes, it is waiting for a review.
Some people here apparently tested it and I don't remember any reported issues with functionality (apart from the fixed mutex bug and the (unsolvable?) color dialog key release event stealing issue). And you did some code / style review in the beginning and after that there were only minor changes. So, overall, the review is "kinda done, but not really". I'm not sure if it is sufficient for a merge.

plugins/Vectorscope/VecControls.h
```diff
@@ -0,0 +1,66 @@
...
+	Vectorscope *m_effect;
+
+	FloatModel m_persistenceModel;
+	BoolModel m_logarithmicModel;
+	BoolModel m_highQualityModel;
+
+	QColor	m_colorFG;
+	QColor	m_colorGrid;
+	QColor	m_colorLabels;
+	QColor	m_colorOutline;
```
PhysSong: Did you use tabs after `QColor` intentionally?

he29-net: Oh; probably not, I'm used to align signal definitions in VHDL so I may have done this without even realizing it. :))
I will fix it here and also in `SaControls.h` (where it originally came from). Good catch, thanks. :)

Issue #4904, opened 2019-03-17
After linking models, values are swapped instead of being the same


JohannesLorenz:
There's a "Link Channels" LED in LADSPA mono effects. The following does not work:

* Create a LADSPA mono effect with >= 2 knobs
* Unlink knob 1, change its value on one side
* Change the value of knob 2 (it's currently linked)
* Hit the "Link Channels" LED 2n+1 times for any n>=0 (e.g. once), and knob 1 is displayed as linked, but the knobs show different values and the models do have different values

This sounds like a bug to me, because the LED looks like that knob was linked.

The issue also occurs on the Lv2 branch, since the code was copied from LADSPA.

JohannesLorenz:
Can anyone confirm that this occurs for them, and they consider it a bug?

It's hard to believe to have a bug in there for such a long time...

karmux:
I can reproduce it already with a single click to "Link Channels" LED.

But for me there's also some drawing problem. I need to reopen LADSPA modal or cover knob 2 it with other modal window for a moment to visually see different value of knob 2 when they are linked.

Knob values get switched. Left knob gets the value of right knob and vice versa from what I see after redrawing the modal window. I would expect to see both knobs have the value of knob 2.

PhysSong:
It seems like there's a bug with `AutomatableModel::linkModels`. You can trigger a similar bug without any LADSPA plugins:
1. Add two instrument tracks
2. Set the volume of the first track to 100% and the second one to 0%
3. Drag the volume knob of the first track, and drop it to the second one
4. Notice the volume of each instrument is changed

PhysSong:
If we try to link model A and model B. Let's consider some situations:
1. User Ctrl+drags A and drops on B
2. A is for the left channel of a mono LADSPA effect, and B is for the left channel(the original case)
Assuming the two models should have the same value after linking, which value should we take?

JohannesLorenz:
@PhysSong I'm not sure why you changed the title. In the OP, I can't see where values are swapped. They are not changed visibly at all.

Also, what do you mean by "B is for the left channel(the original case)"? Do you mean B is a knob of a left channel of another LADSPA *stereo* effect?

PhysSong:
> In the OP, I can't see where values are swapped. They are not changed visibly at all.

Try this:
1. Add a LADSPA mono amplifier
2. Unlink channels
3. Set the gain of the left channel to 0
4. Re-link the channel

[...]

PR #5336, opened 2019-12-07
core: Fix controls linking misbehaviors (#4904)

JohannesLorenz:
This fixes the two issues on linking controls (e.g. knobs) mentioned in #4904:

* The values seem swapped.
* The changed controls are not immediatelly updated in the UI; you have to force it to redraw by e.g. moving a subwindow.

I tested that it now works in all possible cases, which are:

* Linking knobs using ctrl + drag-drop.
* Linking LADSPA controls using the global link LED (only for LADSPA mono effects)

Reviewers may need to read #4904 .

### DIFF
--- a/tests/src/core/AutomatableModelTest.cpp
+++ b/tests/src/core/AutomatableModelTest.cpp
@@ -61,6 +61,20 @@ private slots: // tests
 		AutomatableModel::linkModels(&m1, &m2);
 		QVERIFY(!m1Changed); // it's already linked
 		QVERIFY(!m2Changed);
+
+		resetChanged();
+		BoolModel m3(false);
+		m1.setValue(1.f);
+		m2.setValue(1.f);
+		AutomatableModel::linkModels(&m1, &m2);
+		QVERIFY(m1.value());
+		QVERIFY(m2.value());
+		QVERIFY(!m3.value());
+		AutomatableModel::linkModels(&m2, &m3); // drag m3, drop on m2
+		// m2 should take m3's (0) value
+		// due to a bug(?), this does not happen
+		QVERIFY(m2.value());
+		QVERIFY(!m3.value());
 	}
 } AutomatableModelTests;
 
--- a/tests/src/core/AutomatableModelTest.cpp
+++ b/tests/src/core/AutomatableModelTest.cpp
@@ -1,0 +1,67 @@
+/*
+ * AutomatableModelTest.cpp
+ *
+ * Copyright (c) 2020 Johannes Lorenz <j.git$$$lorenz-ho.me, $$$=@>
+ *
+ * This file is part of LMMS - https://lmms.io
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this program (see COPYING); if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA 02110-1301 USA.
+ *
+ */
+
+#include "QTestSuite.h"
+
+#include "AutomatableModel.h"
+
+class AutomatableModelTest : QTestSuite
+{
+	Q_OBJECT
+
+	bool m1Changed, m2Changed;
+	void resetChanged() { m1Changed = m2Changed = false; }
+
+private slots: // helper slots
+	void onM1Changed(Model* ) { m1Changed = true; }
+	void onM2Changed(Model* ) { m2Changed = true; }
+
+private slots: // tests
+	void LinkTests()
+	{
+		BoolModel m1(false), m2(false);
+
+		QObject::connect(&m1, SIGNAL(dataChanged(Model*)),
+			this, SLOT(onM1Changed(Model*)));
+		QObject::connect(&m2, SIGNAL(dataChanged(Model*)),
+			this, SLOT(onM2Changed(Model*)));
+
+		resetChanged();
+		AutomatableModel::linkModels(&m1, &m1);
+		QVERIFY(!m1Changed); // cannot link to itself
+		QVERIFY(!m2Changed);
+
+		resetChanged();
+		AutomatableModel::linkModels(&m1, &m2);
+		QVERIFY(m1Changed); // since m1 takes the value of m2
+		QVERIFY(!m2Changed); // the second model is the source
+
+		resetChanged();
+		AutomatableModel::linkModels(&m1, &m2);
+		QVERIFY(!m1Changed); // it's already linked
+		QVERIFY(!m2Changed);
+	}
+} AutomatableModelTests;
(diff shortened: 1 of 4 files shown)


PhysSong:
Looks fine as a temporary fix. Do you want to wait mor for reviews?

JohannesLorenz:
@PhysSong Thanks.

> Do you want to wait more for reviews?

Did you think through the logic? Did you do enough tests?

Spekular:
> The changed controls are not immediately updated

It makes sense to me that this would be fixed by `model1->m_value = model2->m_value;`, is this accurate?

I don't understand what the buffer stuff is doing, or how the swapped value issue is fixed. (To be fair, I also don't see how it would occur in the first place).

Finally, the relevant section of `AutomatableModel::setValue( const float value )` seems to be unchanged. Does this mean that this bug/fix only concerns the effects of creating a link? I was under the (mistaken?) impression that the linking behavior was fundamentally broken in some way and wouldn't properly update values.

JohannesLorenz:
> > The changed controls are not immediately updated
> 
> It makes sense to me that this would be fixed by `model1->m_value = model2->m_value;`, is this accurate?

No, this is about being "updated *in the UI*". This is done in 84bbc4e by adding the signal.

> I don't understand what the buffer stuff is doing, or how the swapped value issue is fixed. (To be fair, I also don't see how it would occur in the first place).

FWIR, some controllers (actually most) can be set to be "sample-exact". This means while audio computes `buffersize` samples, controllers do not just computer one sample, they also compute `buffersize samples` (e.g. controller rate = audio rate). So if they are sample exact, they write their values into an `std::vector`, which is the buffer you are asking about.
 
> Finally, the relevant section of `AutomatableModel::setValue( const float value )` seems to be unchanged. Does this mean that this bug/fix only concerns the effects of creating a link? I was under the (mistaken?) impression that the linking behavior was fundamentally broken in some way and wouldn't properly update values.

Indeed, the issue report was only about situations where you create links. Fixing `setValue` would not have helped as it is not called in the situations. There were two problems: (1) that linking did not update the value of any knob, and (2) that linking did not trigger the UI about value changes.

Spekular:
Many thanks for the explanation. So in essence; both the m_value line and the buffer code are there to correct the initial value (the latter for sample exact knobs) on the model side, and emitChanged fixes the view side. Makes sense!

JohannesLorenz:
One issue I noticed though: The two complementary calls of `AutomatableModel::linkModel` have checks to not link models that are are already linked (and not linking a model to itself). My newly inserted code does not do these checks, leading to inefficient buffer copy, and worse IMO, to sending `dataChanged` when no data has been changed.

I plan to add something like:

```
if (!model1.m_linkedModels.contains( model2 ) && model1 != model2)
{
    // new code
}
```

JohannesLorenz:
There are 2 new commits added which require further review.

Also, one odd thing:
* If you drag any knob A on *unconnected* knob B, B takes the value of A (like wanted)
* If you drag any knob A on *connected* knob B, A takes the value of B

It can be fixed by using the following code, but that introduces more strange issues:

```
if( !hasLinkedModels() )
{
	QObject::connect( model, SIGNAL( dataChanged() ),
		this, SIGNAL( dataChanged() ), Qt::DirectConnection );
}
```

IMO wrong behaviour, but not serious enough for the feature freeze branch.

PhysSong:
I think we should rewrite the linking system to fix the oddness completely. However, it's too much for the stable branch.

JohannesLorenz:
Can please anyone do a review for this PR? It's still missing.

Spekular:
> There are 2 new commits added which require further review.

If [this](https://github.com/LMMS/lmms/pull/5336/files/fd37e54c20b723117af770260a642e2753e7ab71..83395adcaf4aac11bd04f8a1a4dcc3af0de9360d) is the correct diff, then the changes make a lot of sense to me. Approved, basically.

JohannesLorenz:
@Spekular Thanks, but this is not the diff. Click on the "files changed" tab on top.

Spekular:
Right, I did that and then selected the latest two commits. I can check the whole thing again though

JohannesLorenz:
@Spekular that's probably better. Yes, the last 3(?) commits should basically be it, but just to be sure... Many thanks for reviewing.

tests/src/core/AutomatableModelTest.cpp
```diff
@@ -0,0 +1,81 @@
...
+		AutomatableModel::linkModels(&m1, &m2);
+		QVERIFY(!m1Changed); // it's already linked
+		QVERIFY(!m2Changed);
+
+		resetChanged();
+		BoolModel m3(false);
+		m1.setValue(1.f);
+		m2.setValue(1.f);
+		AutomatableModel::linkModels(&m1, &m2);
+		QVERIFY(m1.value());
```
Spekular: I assume this is checking `m1.value() != 0`. I think this would be clearer if it were explicit.

Spekular: Same for following two checks.

tests/src/core/AutomatableModelTest.cpp
```diff
@@ -0,0 +1,81 @@
...
+		QVERIFY(m2.value());
+		QVERIFY(!m3.value());
+		AutomatableModel::linkModels(&m2, &m3); // drag m3, drop on m2
+		// m2 should take m3's (0) value
+		// due to a bug(?), this does not happen
+		QVERIFY(m2.value());
```
Spekular: Same as above for this line and next

Spekular:
Aside from some nitpicks in the test, LGTM.

tests/src/core/AutomatableModelTest.cpp
JohannesLorenz: IIRC `m1.value()` is already `bool` (`BoolModel::value`). So comparing it with `0` would not make sense, it would need to be compared with `false`. But things like `m1.value() == true` will be [removed by clang-tidy](https://clang.llvm.org/extra/clang-tidy/checks/readability-simplify-boolean-expr.html), and we'd end up with `m1.value()` again.

So I think we should leave it as it is.

Spekular: Ah, now I see that they're BoolModels. I was thrown off by the `setValue` call using `1.f`, I assumed that was a float value or something.

PhysSong:
@JohannesLorenz It seems like the `Update UI after linking models` commit is unneeded and breaking the test cases; I don't think we still need it because we have another one in the `if` statement.
BTW, it seems like unit test failures don't make the build fail on `stable-1.2` for some reason.

JohannesLorenz:
@PhysSong good catch with the "Update UI after linking models commit"! Looks like I reordered the commits wrong or something.

I'd like to make a commit that reverts this change, but let's first make use of our failing test. Can you tell where in the CI you found the failing test that did not fail the build?

PhysSong:
> Can you tell where in the CI you found the failing test that did not fail the build?

Here: https://travis-ci.org/github/LMMS/lmms/jobs/655021139#L3811-L3812
FYI, `master` builds will fail as expected when any of the unit tests fails.

JohannesLorenz:
Found it, `stable-1.2` does not `set -e` in `.travis/linux..script.sh`, so it does not abort if `tests/tests` failed. `master` does `set -e`.

@PhysSong I can add the `set -e` easily, but it is missing in ~15 files which look different on master and stable. Should I try this, or do we want to avoid this to avoid merge conflicts? Or should I just try it with one file?

PhysSong:
For the `set -e` stuff, please see #3476 for related discussions. I think we can leave scripts in `stable-1.2` as-is because we're not likely to add many more changes to the branch.

Issue #5290, opened 2019-10-30
convert mmp into mmpz


quatran:
What is the  command to compress a *.mmp file into a *.mmpz file?

Basically I'm looking for the opposite of ` lmms -d inputfile.mmpz > output.mmp`.

AndiEcker:
There is an old python project at github (
https://github.com/AndiEcker/LmmsPrjMngr) that allows to convert mmpz to
mmp and visa versa.

The command line options -u converts mmpz to mmp and -c from mmp to mmpz. I
coded it 7 years ago and not touched it since then ... hope it still works
and helps.

If you have any problems let me know.

Am Mi., 30. Okt. 2019 um 18:20 Uhr schrieb Quang Tran <
notifications@github.com>:

zonkmachine:
> Basically I'm looking for the opposite of ` lmms -d inputfile.mmpz > output.mmp`.

That I don't think we have. I'm pretty sure it's an easy task to implement it though and I guess it would be consistent with the `--dump` argument. Do you have an example of when this would be needed?
Example: `lmms --compress inputfile.mmp > output.mmpz`

quatran:
@AndiEcker 
Your project looks interesting but have'nt tried it out yet.  I will provide feedback then.

@zonkmachine 
> Do you have an example of when this would be needed?
Example: lmms --compress inputfile.mmp > output.mmpz

My use case would be writing an application that do some stuff with the project files and then finally convert them back.

olivia8090:
Make Real Money That,s Real Job Not Fack

like Andrew responded I'm in shock that people able to profit $4494 in four weeks on the internet .my Aunt Kyler George got a very nice BMW X 3 S just by some part-time working online with a computer. take a look at the site here
:) AND GOOD LUCK.:)

HERE →→→→→ www.securework99.com

bahaokten:
Hi, I am a new to LMMS
This issue seems like a good first issue for a newcomer, so if it is still open, I can work on it.

rdrpenguin04:
Wow, I was actually kinda looking for something like this without realizing it! I'm writing a converter from SDAT to MMP, so this would be a nice final step. I was thinking of implementing this, but it seems it has a PR already.

zonkmachine:
Fixed in https://github.com/LMMS/lmms/pull/5341

PR #5341, opened 2019-12-09
Implement compress action for commandline use

bahaokten:
Fixes the issue: #5290 
Added compress commandline action to allow conversion from mmp to mmpz
Usage: `lmms --compress infile.mmp > outfile.mmpz`

### DIFF
--- a/doc/bash-completion/lmms
+++ b/doc/bash-completion/lmms
@@ -89,7 +89,7 @@ _lmms()
     pars_render=(--float --bitrate --format --interpolation)
     pars_render+=(--loop --mode --output --profile)
     pars_render+=(--samplerate --oversampling)
-    actions=(dump render rendertracks upgrade)
+    actions=(dump compress render rendertracks upgrade)
     actions_old=(-d --dump -r --render --rendertracks -u --upgrade)
     shortargs+=(-a -b -c -f -h -i -l -m -o -p -s -v -x)
 
--- a/src/core/main.cpp
+++ b/src/core/main.cpp
@@ -161,7 +161,7 @@ void printHelp()
 		"Actions:\n"
 		"  <no action> [options...] [<project>]  Start LMMS in normal GUI mode\n"
 		"  dump <in>                             Dump XML of compressed file <in>\n"
-		"  compress <in>                         Compress file <n>\n"
+		"  compress <in>                         Compress file <in>\n"
 		"  render <project> [options...]         Render given project file\n"
 		"  rendertracks <project> [options...]   Render each track to a different file\n"
 		"  upgrade <in> [out]                    Upgrade file <in> and save as <out>\n"
--- a/src/core/main.cpp
+++ b/src/core/main.cpp
@@ -161,6 +161,7 @@ void printHelp()
 		"Actions:\n"
 		"  <no action> [options...] [<project>]  Start LMMS in normal GUI mode\n"
 		"  dump <in>                             Dump XML of compressed file <in>\n"
+		"  compress <in>                         Compress file <n>\n"
 		"  render <project> [options...]         Render given project file\n"
 		"  rendertracks <project> [options...]   Render each track to a different file\n"
 		"  upgrade <in> [out]                    Upgrade file <in> and save as <out>\n"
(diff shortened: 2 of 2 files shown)


zonkmachine:
Thank you for working on this!

I've tested it and it works as described.

Some thoughts.

>  Usage: `lmms --compress infile.mmp > outfile.mmpz`

If you type the command wrong or try the command without all arguments in order to try out how it works you'll end up dumping binary data to the terminal and this doesn't look nice. I don't know if you can have any other problems but the purely cosmetic ones. Maybe the output could be directly dumped in a file with the same name as the input file but with the .mmpz suffix, if such a file does not already exist. In the similar way as the **--export** command works (though **--export** doesn't test if the file already exist...).

src/core/main.cpp
```diff
@@ -161,6 +161,7 @@ void printHelp()
 		"Actions:\n"
 		"  <no action> [options...] [<project>]  Start LMMS in normal GUI mode\n"
 		"  dump <in>                             Dump XML of compressed file <in>\n"
+		"  compress <in>                         Compress file <n>\n"
```
zonkmachine: I think this needs improving. `compress <in> > <out>` is more like an ordinary use case. I don't know if that's good syntax too but you get the idea. Also I don't know what it looks like on non linux/unix systems.

`Compress file <n>\n"`

Did you mean `Compress file <in>\n"`

src/core/main.cpp
```diff
@@ -427,6 +428,22 @@ int main( int argc, char * * argv )
 
 			return EXIT_SUCCESS;
 		}
+		else if( arg == "compress" || arg == "--compress")
```
zonkmachine:
```suggestion
		else if( arg == "compress" || arg == "--compress" )
```

src/core/main.cpp
```diff
@@ -427,6 +428,22 @@ int main( int argc, char * * argv )
...
+		{
+			++i;
+
+			if( i == argc )
+			{
+				return noInputFileError();
+			}
+			
+			QFile f( QString::fromLocal8Bit( argv[i] ) );
+			f.open( QIODevice::ReadOnly);
```
zonkmachine:
```suggestion
			f.open( QIODevice::ReadOnly );
```

zonkmachine (changes requested):
Just some style nitpicking.

zonkmachine:
@bahaokten You have one more unanswered question here: https://github.com/LMMS/lmms/pull/5341#discussion_r357937386

Shouldn't it be `Compress file <in>\n"`?

src/core/main.cpp
Veratil, line 164:
```suggestion
		"  compress <in>                         Compress file <in>\n"
```

src/core/main.cpp
```diff
@@ -427,6 +428,22 @@ int main( int argc, char * * argv )
 
 			return EXIT_SUCCESS;
 		}
+		else if( arg == "compress" || arg == "--compress" )
```
Veratil, line 431: Can you update the code formatting to the [Coding Conventions](https://github.com/LMMS/lmms/wiki/Coding-Conventions) please?

Just some whitespace changes is all. 😄

Veratil (changes requested):

src/core/main.cpp
zonkmachine: Merged. It looks like the one who press the merge button on  a suggested change becomes the 'Author'.

src/core/main.cpp
zonkmachine: But it's adopting to the surrounding code so I think it's good like this.

doc/bash-completion/lmms
```diff
@@ -89,7 +89,7 @@ _lmms()
     pars_render=(--float --bitrate --format --interpolation)
     pars_render+=(--loop --mode --output --profile)
     pars_render+=(--samplerate --oversampling)
-    actions=(dump render rendertracks upgrade)
+    actions=(dump compress render rendertracks upgrade)
     actions_old=(-d --dump -r --render --rendertracks -u --upgrade)
```
zonkmachine, line 93: The bash completion for compress works but it would also nice to have the old style --compress work.

zonkmachine: Right. None of the other actions are bash completing with old style longopts so scratch that for this PR.

zonkmachine:
What's left to do here?

I can compress projects from the command line. :+1: 
This looks good to go to me.

 @Veratil There hasn't been any response from OP since January. No need to wait any longer. If you want to fixup any formatting issues to your liking I think you should do so and then squash it.

zonkmachine (approved):

zonkmachine:
@Veratil Merge? Do you want to touch up the formatting first?

zonkmachine:
> @Veratil Merge? Do you want to touch up the formatting first?

OK. No answer, merging.

PR #5343, opened 2019-12-11
MIDI import: fix putting notes before the beginning of a pattern

### LINKED ISSUE #2741: MIDI import misplaces notes
Reported on the forum: http://www.lmms.io/forum/viewtopic.php?f=7&t=5016

> I produce my midi files with musescore 2.0.2 on Ubuntu 15.10 and save them with file>export in musescore. 

What happens is that notes seem  to be created at negative points in time when importing in LMMS, i.e. before the segment starts. The notes can be "resurrected" by opening the segment in the piano roll and moving the notes. Musescore always exports using [running status](http://www.blitter.com/~russtopia/MIDI/~jglatt/tech/midispec/run.htm), which some other software can't handle at all, so that's my prime suspect.

Using the `midi2text` and `text2midi` programs that come with the [midifile](https://github.com/craigsapp/midifile) library I converted the test file provided into one not using running status. The converted one imported correctly. As far as I can see the files are identical when it comes to note data, but there's some controller data in the original that's getting lost in the process, so I'm not quite 100% certain. But it's most likely the `portsmf/allegro` library that's somehow reporting wrong times to the next layer of the import code.

import results comparison

Sample files, the original with running status and the other without:
[testfiles.zip](https://github.com/LMMS/lmms/files/240138/testfiles.zip)

(:bulb: That midifile library looks like a good choice if MIDI import and export is to be rewritten at some point)

PhysSong:
Fixes #2741.

### DIFF
--- a/plugins/MidiImport/MidiImport.cpp
+++ b/plugins/MidiImport/MidiImport.cpp
@@ -215,8 +215,7 @@ public:
 		p( NULL ),
 		it_inst( NULL ),
 		isSF2( false ),
-		hasNotes( false ),
-		lastEnd( 0 )
+		hasNotes( false )
 	{ }
 	
 	InstrumentTrack * it;
@@ -224,7 +223,6 @@ public:
 	Instrument * it_inst;
 	bool isSF2; 
 	bool hasNotes;
-	MidiTime lastEnd;
 	QString trackName;
 	
 	smfMidiChannel * create( TrackContainer* tc, QString tn )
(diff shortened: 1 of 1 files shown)


zonkmachine:
The test files now opens correctly.

One thing I notice though is that the two test files seem different. For instance, the 'Running status' one has a lower volume and just a hint of pan in it while the 'No running status' one has default volume and pan. I assume this is because the test files are differing and not because of other issues with MIDI import.

zonkmachine:
Merge?

PR #5348, opened 2019-12-17
Fix wrong lengths of exported tracks, when tracks have different lengths.

Cyp:
Song::updateLength() was called in ProjectRenderer::run() after Song::startExport(),
updating m_length too late, resulting in it being used as the length of the wrong track.

Doesn't seem to be needed in stable-1.2, as 1.2's Song::startExport doesn't use m_length there.


Cyp:
I think the two test failures may be unrelated to this commit:

```
Error: No available formula with the name "5" 
==> Searching for a previously deleted formula (in the last month)...
Warning: homebrew/core is shallow clone. To get complete history run:
  git -C "$(brew --repo homebrew/core)" fetch --unshallow
Error: No previously deleted formula found.
```

```
ERROR:root:code for hash md5 was not found.
Traceback (most recent call last):
```

PhysSong:
I haven't tested yet, but the code looks good.

PhysSong:
@Cyp Could you explain how to reproduce the relevant bug? I want to test it.

Cyp:
Have at least two tracks of very different lengths, and export tracks separately. As far as I remember, the exported first track would get the length of the second track, being cut off or extended.

PhysSong:
I found that there's a bug with `Export as loop` when rendering multiple tracks, but it's not related to your changes. It's because `Song::stopExport()` unsets `m_exportLoop` after rendering the first track.
After changing that, your PR works as expected. :+1:

Issue #1857, opened 2015-03-11
MIDI-based instruments play an octave too low by default


softrabbit:
Left: TripleOscillator, one sine wave, coarse tuning 0, 440 Hz
Right: ZynAddSubFX, one sine wave, coarse tuning 0 AFAIK, 220 Hz

Same root note, same note played as the image shows. Both do output the correct note 69 through MIDI out.

![off_by_1_octave](https://cloud.githubusercontent.com/assets/6411463/6592131/19e073ee-c7d2-11e4-8cbd-d9086c81849a.png)

(For further evidence, this is how I had to compensate in OpulenZ: https://github.com/LMMS/lmms/blob/master/plugins/opl2/opl2instrument.cpp#L306 )

badosu:
@softrabbit This is weird, A4 should always be 440hz right?

eagles051387:
Yes as per http://www.phy.mtu.edu/~suits/notefreqs.html A4 is 440hz.

On Wed, Mar 11, 2015 at 3:57 PM, Amadeus Folego <notifications@github.com>
wrote:

> @softrabbit <https://github.com/softrabbit> This is weird, A4 should
> always be 440hz right?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/LMMS/lmms/issues/1857#issuecomment-78279353>.
>



-- 
Jonathan Aquilina

Sti2nd:
> A4 should always be 440hz right?

Maybe, but the world doesn't agree. 442 is also highly accepted for orchestras. We can only assume LMMS uses 440? The frequncy on the right image is far lower than 440Hz, though, so it is definitely not an A.

badosu:
Well 220 is half 440, so it's A, just one octave below :-).

Sti2nd:
Lol, I was too fast there, yes, of course it could be a lower A!

badosu:
Yeah, if we had an integration layer before touching Zyn, this would probably be not an issue.

@softrabbit @Sti2nd Can you make a list of intrumenta that you have found that present this problem for when someone takes this task the person already has something to work with?

softrabbit:
Grepping the plugin sources for IsMidiBased finds these possible victims: Carla, Vestige, ZynAddSubFX (confirmed), OpulenZ (fixed internally).

[...]

PR #5357, opened 2019-12-28
Custom note labels in Pianoroll editor

komodor1:
I coded function for custom note labels for Pianoroll editor. Three labels files paths are stored in lmms config file. They can be set in settings -> paths. After setting paths lmms need restart. Then in Pianoroll editor you can select which file you want to use. Function reads labels from .txt files in pattern like attached one. File can't have any empty lines between labels/comments. Empty lines can be only after labels list. Between note number and label must be ONLY ONE space. 
In attached screenshots you can see how modifications looks like in Pianoroll and settings.

TODO:
Icon for labels file combo box in Pianoroll editor (currently using Quantize icon),
Only three files support, maybe add some kind of list,
May need some code clean-up and check,

settings
pianoroll
[labels.txt](https://github.com/LMMS/lmms/files/4007313/labels.txt)

### DIFF
--- a/src/gui/SetupDialog.cpp
+++ b/src/gui/SetupDialog.cpp
@@ -30,6 +30,8 @@
 #include <QLineEdit>
 #include <QMessageBox>
 #include <QScrollArea>
+#include <QTextStream>
+#include <QFile>
 
 #include "debug.h"
 #include "embed.h"
@@ -143,7 +145,10 @@ SetupDialog::SetupDialog(ConfigTabs tab_to_open) :
 	m_sf2File(QDir::toNativeSeparators(ConfigManager::inst()->sf2File())),
 #endif
 	m_themeDir(QDir::toNativeSeparators(ConfigManager::inst()->themeDir())),
-	m_backgroundPicFile(QDir::toNativeSeparators(ConfigManager::inst()->backgroundPicFile()))
+	m_backgroundPicFile(QDir::toNativeSeparators(ConfigManager::inst()->backgroundPicFile())),
+	m_labels1Dir(QDir::toNativeSeparators(ConfigManager::inst()->labels1Dir())),
+	m_labels2Dir(QDir::toNativeSeparators(ConfigManager::inst()->labels2Dir())),
+	m_labels3Dir(QDir::toNativeSeparators(ConfigManager::inst()->labels3Dir()))
 {
 	setWindowIcon(embed::getIconPixmap("setup_general"));
 	setWindowTitle(tr("Settings"));
@@ -766,7 +771,18 @@ SetupDialog::SetupDialog(ConfigTabs tab_to_open) :
 		SLOT(setBackgroundPicFile(const QString &)),
 		SLOT(openBackgroundPicFile()),
 		m_backgroundPicFileLineEdit);
-
+	addPathEntry("Piano roll custom labels 1", m_labels1Dir,
+		SLOT(setLabels1Dir(const QString &)),
+		SLOT(openLabels1Dir()),
+		m_labels1DirLineEdit);
+	addPathEntry("Piano roll custom labels 2", m_labels2Dir,
+		SLOT(setLabels2Dir(const QString &)),
+		SLOT(openLabels2Dir()),
+		m_labels2DirLineEdit);
+	addPathEntry("Piano roll custom labels 3", m_labels3Dir,
+		SLOT(setLabels3Dir(const QString &)),
+		SLOT(openLabels3Dir()),
+		m_labels3DirLineEdit);
 	pathSelectorsLayout->addStretch();
 
 	pathSelectors->setLayout(pathSelectorsLayout);
@@ -923,6 +939,11 @@ void SetupDialog::accept()
 	ConfigManager::inst()->setGIGDir(QDir::fromNativeSeparators(m_gigDir));
 	ConfigManager::inst()->setThemeDir(QDir::fromNativeSeparators(m_themeDir));
 	ConfigManager::inst()->setBackgroundPicFile(m_backgroundPicFile);
+	ConfigManager::inst()->setLabels1Dir(QDir::fromNativeSeparators(m_labels1Dir));
+	ConfigManager::inst()->setLabels2Dir(QDir::fromNativeSeparators(m_labels2Dir));
+	ConfigManager::inst()->setLabels3Dir(QDir::fromNativeSeparators(m_labels3Dir));
+
+
 
 	// Tell all audio-settings-widgets to save their settings.
 	for(AswMap::iterator it = m_audioIfaceSetupWidgets.begin();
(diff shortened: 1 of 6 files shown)


src/gui/editors/PianoRoll.cpp
```diff
@@ -2974,6 +3090,10 @@ void PianoRoll::paintEvent(QPaintEvent * pe )
 				}
 				p.drawText( textStart, noteString );
 			}
+		}
+		if (drawCustomLabels){
```
Veratil: Empty if?

src/gui/editors/PianoRoll.cpp
```diff
@@ -2830,8 +2858,96 @@ int PianoRoll::xCoordOfTick( int tick )
...
+		if(customLabelsFile == 1) labelsPath = labels1Path; 
+		if(customLabelsFile == 2) labelsPath = labels2Path; 
+		if(customLabelsFile == 3) labelsPath = labels3Path; 
+				
+		QFile labelsFile(labelsPath);
+		labelsFile.open(QIODevice::ReadOnly);
+		QTextStream stream(&labelsFile);
+		QString line = stream.readLine();
+		QRegExp num("\\d*");
+		while (!line.isNull()) {
```
Veratil: This happens *every* `paintEvent`, which will block on file I/O which can cause freezing of the UI in worse cases. These labels need to be read on LMMS load rather than here and painted from memory.

src/gui/SetupDialog.cpp
```diff
@@ -1319,7 +1340,176 @@ void SetupDialog::setBackgroundPicFile(const QString & backgroundPicFile)
...
+		}	
+	}
+}
+
+void SetupDialog::setLabels3Dir(const QString & labels3Dir)
+{
+	m_labels3Dir = labels3Dir;
+}
+
+void SetupDialog::openLabels3Dir()
```
Veratil: All three of these functions are essentially duplicates, I feel they could be combined into one with `if`s depending on which file loaded. Or if changed to an XML file per my review comment this wouldn't be necessary. ;)

src/gui/SetupDialog.cpp
```diff
@@ -1319,7 +1340,176 @@ void SetupDialog::setBackgroundPicFile(const QString & backgroundPicFile)
...
+}
+
+void SetupDialog::openLabels2Dir()
+{
+	QString new_file = FileDialog::getOpenFileName(this,
+		tr("Choose your labels file"), m_labels2Dir, "Text files (*.txt)");
+
+	if (!new_file.isEmpty())
+	{
+		bool fileCorrect = 1;
```
Veratil: This should be `true`, not 1.

src/gui/SetupDialog.cpp
```diff
@@ -1319,7 +1340,176 @@ void SetupDialog::setBackgroundPicFile(const QString & backgroundPicFile)
...
+				break;
+			}
+			if(line.mid(0, 2) == "//"){
+				//Comment line
+			} else if (num.exactMatch(line.mid(0,1))){
+				if (num.exactMatch(line.mid(1,1))){
+					if (num.exactMatch(line.mid(2,1))){
+						int number = line.mid(0,3).toInt();
+						if (number > 127){
+							fileCorrect = 0;
```
Veratil: These should be `false`, not 0.

src/gui/SetupDialog.cpp
```diff
@@ -1319,7 +1340,176 @@ void SetupDialog::setBackgroundPicFile(const QString & backgroundPicFile)
...
+	{
+		bool fileCorrect = 1;
+		QFile labelsFile(new_file);
+		labelsFile.open(QIODevice::ReadOnly);
+		if (!labelsFile.isOpen()) return ;
+		
+		QTextStream stream(&labelsFile);
+		QString line = stream.readLine();
+		QRegExp num("\\d*");
+		while (!line.isNull()) {
```
Veratil: These loops and ifs within these blocks should follow the [Coding Conventions](https://github.com/LMMS/lmms/wiki/Coding-Conventions).

src/gui/editors/PianoRoll.cpp
komodor1: Removed in last commit.

src/gui/SetupDialog.cpp
komodor1: Corrected.

src/gui/SetupDialog.cpp
komodor1: Corrected.

src/gui/SetupDialog.cpp
komodor1: Reviewed all changes, and made them follow Coding Conventions.

src/gui/editors/PianoRoll.cpp
komodor1: Yes, I've noticed that. I know that is not good way, I'm thinking how to change that.

src/gui/SetupDialog.cpp
komodor1: You made me think about XML file, so I won't change that for now ;)

komodor1:
> I like the idea!
> 
> But I don't like that it's new settings options rather than a specific file you read or not, and you are limited to 3 files instead of the ability to add any number to a single file.
> 
> I think this would be much more extensible if it was an XML file that can be read on load (if present in working dir or not). If like this, you can create a tree structure in XML that allows an unlimited number of labels. Granted the file is a little more difficult to write out, but it would be able to be parsed by Qt's loadSettings and not have to be parsed manually.

Thanks for feedback! 
XML file is really good idea. But it will be kinda difficult to write maps manually by user, so that's why I used TXT files - easy to write, and very similar to Reaper files. But I see another way. Parse .TXT file, and make tree structure like you proposed. It will be easier to maintain multiple labels maps. Then just add simple list of loaded maps in settings, where you can add or remove maps.
About parsing file in paintEvent. I don't see way how change that. Need some time to think.

qnebra:
I think better this will work as another option in pianoroll bar. As this will be more user-friendly UX.

Veratil:
> XML file is really good idea. But it will be kinda difficult to write maps manually by user, so that's why I used TXT files - easy to write, and very similar to Reaper files. But I see another way. Parse .TXT file, and make tree structure like you proposed. It will be easier to maintain multiple labels maps. Then just add simple list of loaded maps in settings, where you can add or remove maps.

The reason I say XML is twofold. 1) You leverage Qt's XML parser, and don't need to do all this extra manual parsing. 2) Everything in LMMS is saved as an XML file, this being a plaintext file breaks from that and doesn't fit the mold.

> About parsing file in paintEvent. I don't see way how change that. Need some time to think.

Read on startup from the file and save in memory.

src/gui/SetupDialog.cpp
```diff
@@ -1319,7 +1340,225 @@ void SetupDialog::setBackgroundPicFile(const QString & backgroundPicFile)
...
+{
+	QString new_file = FileDialog::getOpenFileName(this,
+		tr("Choose your labels file"), m_labels3Dir, "Text files (*.txt)");
+
+	if (!new_file.isEmpty())
+	{
+		bool fileCorrect = true;
+		QFile labelsFile(new_file);
+		labelsFile.open(QIODevice::ReadOnly);
+		if (!labelsFile.isOpen()) return ;
```
Veratil: Missing `{}`

src/gui/editors/PianoRoll.cpp
```diff
@@ -128,7 +128,9 @@ static QString getNoteString( int key )
 	return s_noteStrings[key % 12] + QString::number( static_cast<int>( key / KeysPerOctave ) );
 }
 
-
+bool labelsFile1Correct = 0;
```
Veratil: `false` instead of `0`.

src/gui/editors/PianoRoll.cpp
```diff
@@ -366,6 +369,32 @@ PianoRoll::PianoRoll() :
 	connect( &m_quantizeModel, SIGNAL( dataChanged() ),
 					this, SLOT( quantizeChanged() ) );
 
+	// Set up labels model
```
Veratil: Formatting for this entire section to line 397.

src/gui/editors/PianoRoll.cpp
```diff
@@ -3114,6 +3267,7 @@ void PianoRoll::paintEvent(QPaintEvent * pe )
 			}
 			p.drawLine( WHITE_KEY_WIDTH, y, width(), y );
 			++key;
+
```
Veratil: Extra line added on accident?

src/gui/editors/PianoRoll.cpp
```diff
@@ -3456,10 +3610,30 @@ void PianoRoll::paintEvent(QPaintEvent * pe )
 		case ModeEditDetuning: cursor = s_toolOpen; break;
 	}
 	QPoint mousePosition = mapFromGlobal( QCursor::pos() );
-	if( cursor != NULL && mousePosition.y() > keyAreaTop() && mousePosition.x() > noteEditLeft())
+	if ( cursor != NULL && mousePosition.y() > keyAreaTop() && mousePosition.x() > noteEditLeft())
 	{
 		p.drawPixmap( mousePosition + QPoint( 8, 8 ), *cursor );
 	}
+	if (drawCustomLabels)
```
Veratil: Since this is the last drawn thing in `paintEvent`, this will draw over notes at the start of the pattern. Should it?

src/gui/editors/PianoRoll.cpp
```diff
@@ -4421,6 +4600,50 @@ Note * PianoRoll::noteUnderMouse()
 	return NULL;
 }
 
+bool PianoRoll::labelsPathCorrect( QString labelsPath ){
+	bool fileCorrect = 1;
```
Veratil: `true` instead of `1`.

src/gui/editors/PianoRoll.cpp
```diff
@@ -4421,6 +4600,50 @@ Note * PianoRoll::noteUnderMouse()
 	return NULL;
 }
 
+bool PianoRoll::labelsPathCorrect( QString labelsPath ){
```
Veratil: Formatting for this function.

src/gui/editors/PianoRoll.cpp
```diff
@@ -4515,6 +4738,15 @@ PianoRollWindow::PianoRollWindow() :
 	m_quantizeComboBox->setFixedSize( 64, 22 );
 	m_quantizeComboBox->setToolTip( tr( "Quantization") );
 
+	// setup labels-stuff
```
Veratil: Fix whitespace in this area.

PhysSong:
> > XML file is really good idea. But it will be kinda difficult to write maps manually by user, so that's why I used TXT files - easy to write, and very similar to Reaper files. But I see another way. Parse .TXT file, and make tree structure like you proposed. It will be easier to maintain multiple labels maps. Then just add simple list of loaded maps in settings, where you can add or remove maps.
> 
> The reason I say XML is twofold. 1) You leverage Qt's XML parser, and don't need to do all this extra manual parsing. 2) Everything in LMMS is saved as an XML file, this being a plaintext file breaks from that and doesn't fit the mold.

We can support both Reaper-compatible plaintext file and standard XML/JSON files. BTW, we need to think about #1857 first.


> > About parsing file in paintEvent. I don't see way how change that. Need some time to think.
> 
> Read on startup from the file and save in memory.

I can see the parsing logic is repeated a few times in many places. We can probably remove the file lists from the settings and let users load them manually(optionally with histories).

PR #5360, opened 2019-12-30
Fix #5194: Fix knobs moving too fast

### LINKED ISSUE #5194: Knobs are erratic on macOS (and other systems)
As the title suggests, all knobs behave "wonky" on macOS.

Originally [posted on Reddit](https://www.reddit.com/r/lmms/comments/d4qe3b/knobs_acting_strangely/), the OP (`u/Coelacanth2`) was kind enough to share a screen record.  Holding <kbd>Shift</kbd> helps, but they're still way off.
A9tKsji

I can reproduce this on my machine, macOS Mojave 10.14.6 using the latest tip of `stable-1.2` and all dependencies updated.  Note, the change in behavior from older versions may be Qt5 related.

JohannesLorenz:
For systems where QCursor::setPos() has no effect, this fix adds a
workaround for the situation where a knob value is changed via mouse
click and move. The fix simulates that the cursor is moved back to the
original click position.

### DIFF
--- a/src/gui/widgets/Knob.cpp
+++ b/src/gui/widgets/Knob.cpp
@@ -28,6 +28,7 @@
 #include <QInputDialog>
 #include <QMouseEvent>
 #include <QPainter>
+#include <QScreen>
 #include <QWhatsThis>
 
 #ifndef __USE_XOPEN
@@ -50,6 +51,7 @@
 #include "TextFloat.h"
 
 TextFloat * Knob::s_textFloat = NULL;
+Knob::SetPosStatusType Knob::s_setPosStatus = Knob::SetPosStatusType::NotChecked;
 
 
 
@@ -586,14 +588,39 @@ void Knob::mousePressEvent( QMouseEvent * _me )
 			thisModel->saveJournallingState( false );
 		}
 
+		// on apple, we know that setPos will not work in many cases
+		// (and that the below check won't work either)
+		// to not confuse the user, we won't hide the cursor on apple,
+		// and we won't call setPos on apple either
+#ifdef LMMS_BUILD_APPLE
+		(void)s_setPosStatus;
+#else
+		if(s_setPosStatus == SetPosStatusType::NotChecked)
+		{
+			QPoint oldPos = QCursor::pos();
+			QCursor::setPos(oldPos - QPoint(1,1));
+			s_setPosStatus = (QCursor::pos() == oldPos)
+				? SetPosStatusType::NoEffect
+				: SetPosStatusType::Works;
+			QCursor::setPos(oldPos);
+		}
+#endif
+
 		const QPoint & p = _me->pos();
-		m_origMousePos = p;
-		m_mouseOffset = QPoint(0, 0);
+		m_origMousePos = m_lastMousePos = p;
 		m_leftOver = 0.0f;
 
 		emit sliderPressed();
 
-		QApplication::setOverrideCursor( Qt::BlankCursor );
+#ifndef LMMS_BUILD_APPLE
+		if(s_setPosStatus == SetPosStatusType::Works)
+		{
+			// if overriding of cursor does not work, the user can
+			// be confused when a hidden cursor reappears somewhere
+			// outside of the knob
+			QApplication::setOverrideCursor( Qt::BlankCursor );
+		}
+#endif
 		s_textFloat->setText( displayValue() );
 		s_textFloat->moveGlobal( this,
 				QPoint( width() + 2, 0 ) );
@@ -618,12 +645,55 @@ void Knob::mousePressEvent( QMouseEvent * _me )
 
 void Knob::mouseMoveEvent( QMouseEvent * _me )
 {
-	if( m_buttonPressed && _me->pos() != m_origMousePos )
+	if( m_buttonPressed && _me->pos() != m_lastMousePos )
 	{
-		m_mouseOffset = _me->pos() - m_origMousePos;
-		setPosition( m_mouseOffset );
+		// knob position is changed depending on last mouse position
+		setPosition( _me->pos() - m_lastMousePos );
 		emit sliderMoved( model()->value() );
-		QCursor::setPos( mapToGlobal( m_origMousePos ) );
+
+		// the rest of the code updates the cursor and/or m_origMousePos
+
+		// get screen min/max values where a flip is not required yet
+		const QRect& rec =
+#if QT_VERSION >= 0x050A00
+			QApplication::screenAt(QCursor::pos())->geometry();
+#else
+			QApplication::desktop()->availableGeometry();
+#endif
+		int ymax = qMax(rec.height() - 2, 0), xmax = qMax(rec.width() - 2, 0);
+		int ymin = 1, xmin = 1;
+
+		// calculate new coordinates
+		int x = QCursor::pos().x(), y = QCursor::pos().y();
+		QPoint newPos(
+			(x<xmin) ? xmax : (x>xmax) ? xmin : x,
+			(y<ymin) ? ymax : (y>ymax) ? ymin : y);
+
+		auto setCursor = [&](const QPoint& p) -> bool
+		{
+#ifdef LMMS_BUILD_APPLE
+			(void)p;
+			return false;
+#else
+			QCursor::setPos(newPos);
+			return QCursor::pos() == newPos;
+#endif
+		};
+
+		// no flip, or calling QCursor::setPos() for flip has no effect?
+		// (the latter occurs on some systems without accessability)
+		if(newPos == QCursor::pos() || !setCursor(newPos))
+		{
+			// original position for next time is current position
+			m_lastMousePos = _me->pos();
+		}
+		else
+		{
+			// successful flip of cursor
+			// bash original position to current position
+			// to avoid the knob snapping back
+			m_lastMousePos = mapFromGlobal(QCursor::pos());
+		}
 	}
 	s_textFloat->setText( displayValue() );
 }
@@ -642,7 +712,13 @@ void Knob::mouseReleaseEvent( QMouseEvent* event )
 		}
 	}
 
-	m_buttonPressed = false;
+	if( m_buttonPressed )
+	{
+		m_buttonPressed = false;
+#ifndef LMMS_BUILD_APPLE
+		QCursor::setPos( mapToGlobal( m_origMousePos ) );
+#endif
+	}
 
 	emit sliderReleased();
 
--- a/src/gui/widgets/Knob.cpp
+++ b/src/gui/widgets/Knob.cpp
@@ -499,8 +499,8 @@ float Knob::getValue( const QPoint & _p )
 {
 	float value;
 
-	// arcane mathemagicks for calculating knob movement
-	value = ( ( _p.y() + _p.y() * qMin( qAbs( _p.y() / 2.5f ), 6.0f ) ) ) / 12.0f;
+	// knob value increase is linear to mouse movement
+	value = .09f * _p.y();
 
 	// if shift pressed we want slower movement
 	if( gui->mainWindow()->isShiftPressed() )
--- a/src/gui/widgets/Knob.cpp
+++ b/src/gui/widgets/Knob.cpp
@@ -498,7 +498,7 @@ float Knob::getValue( const QPoint & _p )
 	float value;
 
 	// knob value increase is linear to mouse movement
-	value = .09f * _p.y();
+	value = .4f * _p.y();
 
 	// if shift pressed we want slower movement
 	if( gui->mainWindow()->isShiftPressed() )
--- a/src/gui/widgets/Knob.cpp
+++ b/src/gui/widgets/Knob.cpp
@@ -28,7 +28,6 @@
 #include <QInputDialog>
 #include <QMouseEvent>
 #include <QPainter>
-#include <QScreen>
 #include <QWhatsThis>
 
 #ifndef __USE_XOPEN
@@ -51,7 +50,6 @@
 #include "TextFloat.h"
 
 TextFloat * Knob::s_textFloat = NULL;
-Knob::SetPosStatusType Knob::s_setPosStatus = Knob::SetPosStatusType::NotChecked;
 
 
 
@@ -588,39 +586,12 @@ void Knob::mousePressEvent( QMouseEvent * _me )
 			thisModel->saveJournallingState( false );
 		}
 
-		// on apple, we know that setPos will not work in many cases
-		// (and that the below check won't work either)
-		// to not confuse the user, we won't hide the cursor on apple,
-		// and we won't call setPos on apple either
-#ifdef LMMS_BUILD_APPLE
-		(void)s_setPosStatus;
-#else
-		if(s_setPosStatus == SetPosStatusType::NotChecked)
-		{
-			QPoint oldPos = QCursor::pos();
-			QCursor::setPos(oldPos - QPoint(1,1));
-			s_setPosStatus = (QCursor::pos() == oldPos)
-				? SetPosStatusType::NoEffect
-				: SetPosStatusType::Works;
-			QCursor::setPos(oldPos);
-		}
-#endif
-
 		const QPoint & p = _me->pos();
 		m_origMousePos = m_lastMousePos = p;
 		m_leftOver = 0.0f;
 
 		emit sliderPressed();
 
-#ifndef LMMS_BUILD_APPLE
-		if(s_setPosStatus == SetPosStatusType::Works)
-		{
-			// if overriding of cursor does not work, the user can
-			// be confused when a hidden cursor reappears somewhere
-			// outside of the knob
-			QApplication::setOverrideCursor( Qt::BlankCursor );
-		}
-#endif
 		s_textFloat->setText( displayValue() );
 		s_textFloat->moveGlobal( this,
 				QPoint( width() + 2, 0 ) );
@@ -650,50 +621,8 @@ void Knob::mouseMoveEvent( QMouseEvent * _me )
 		// knob position is changed depending on last mouse position
 		setPosition( _me->pos() - m_lastMousePos );
 		emit sliderMoved( model()->value() );
-
-		// the rest of the code updates the cursor and/or m_origMousePos
-
-		// get screen min/max values where a flip is not required yet
-		const QRect& rec =
-#if QT_VERSION >= 0x050A00
-			QApplication::screenAt(QCursor::pos())->geometry();
-#else
-			QApplication::desktop()->availableGeometry();
-#endif
-		int ymax = qMax(rec.height() - 2, 0), xmax = qMax(rec.width() - 2, 0);
-		int ymin = 1, xmin = 1;
-
-		// calculate new coordinates
-		int x = QCursor::pos().x(), y = QCursor::pos().y();
-		QPoint newPos(
-			(x<xmin) ? xmax : (x>xmax) ? xmin : x,
-			(y<ymin) ? ymax : (y>ymax) ? ymin : y);
-
-		auto setCursor = [&](const QPoint& p) -> bool
-		{
-#ifdef LMMS_BUILD_APPLE
-			(void)p;
-			return false;
-#else
-			QCursor::setPos(newPos);
-			return QCursor::pos() == newPos;
-#endif
-		};
-
-		// no flip, or calling QCursor::setPos() for flip has no effect?
-		// (the latter occurs on some systems without accessability)
-		if(newPos == QCursor::pos() || !setCursor(newPos))
-		{
-			// original position for next time is current position
-			m_lastMousePos = _me->pos();
-		}
-		else
-		{
-			// successful flip of cursor
-			// bash original position to current position
-			// to avoid the knob snapping back
-			m_lastMousePos = mapFromGlobal(QCursor::pos());
-		}
+		// original position for next time is current position
+		m_lastMousePos = _me->pos();
 	}
 	s_textFloat->setText( displayValue() );
 }
@@ -712,13 +641,7 @@ void Knob::mouseReleaseEvent( QMouseEvent* event )
 		}
 	}
 
-	if( m_buttonPressed )
-	{
-		m_buttonPressed = false;
-#ifndef LMMS_BUILD_APPLE
-		QCursor::setPos( mapToGlobal( m_origMousePos ) );
-#endif
-	}
+	m_buttonPressed = false;
 
 	emit sliderReleased();
 
--- a/src/gui/widgets/Knob.cpp
+++ b/src/gui/widgets/Knob.cpp
@@ -587,7 +587,7 @@ void Knob::mousePressEvent( QMouseEvent * _me )
 		}
 
 		const QPoint & p = _me->pos();
-		m_origMousePos = m_lastMousePos = p;
+		m_lastMousePos = p;
 		m_leftOver = 0.0f;
 
 		emit sliderPressed();
--- a/src/gui/widgets/LcdSpinBox.cpp
+++ b/src/gui/widgets/LcdSpinBox.cpp
@@ -41,7 +41,7 @@ LcdSpinBox::LcdSpinBox( int numDigits, QWidget* parent, const QString& name ) :
 	LcdWidget( numDigits, parent, name ),
 	IntModelView( new IntModel( 0, 0, 0, NULL, name, true ), this ),
 	m_mouseMoving( false ),
-	m_origMousePos(),
+	m_lastMousePos(),
 	m_displayOffset( 0 )
 {
 }
@@ -53,7 +53,7 @@ LcdSpinBox::LcdSpinBox( int numDigits, const QString& style, QWidget* parent, co
 	LcdWidget( numDigits, parent, name ),
 	IntModelView( new IntModel( 0, 0, 0, NULL, name, true ), this ),
 	m_mouseMoving( false ),
-	m_origMousePos(),
+	m_lastMousePos(),
 	m_displayOffset( 0 )
 {
 }
@@ -98,8 +98,7 @@ void LcdSpinBox::mousePressEvent( QMouseEvent* event )
 						event->y() < cellHeight() + 2  )
 	{
 		m_mouseMoving = true;
-		m_origMousePos = event->globalPos();
-		QApplication::setOverrideCursor( Qt::BlankCursor );
+		m_lastMousePos = event->globalPos();
 
 		AutomatableModel *thisModel = model();
 		if( thisModel )
@@ -121,7 +120,7 @@ void LcdSpinBox::mouseMoveEvent( QMouseEvent* event )
 {
 	if( m_mouseMoving )
 	{
-		int dy = event->globalY() - m_origMousePos.y();
+		int dy = 3 * (event->globalY() - m_lastMousePos.y());
 		if( event->modifiers() & Qt::ShiftModifier )
 			dy = qBound( -4, dy/4, 4 );
 		if( dy > 1 || dy < -1 )
@@ -129,8 +128,8 @@ void LcdSpinBox::mouseMoveEvent( QMouseEvent* event )
 			model()->setInitValue( model()->value() -
 						dy / 2 * model()->step<int>() );
 			emit manualChange();
-			QCursor::setPos( m_origMousePos );
 		}
+		m_lastMousePos = event->globalPos();
 	}
 }
 
@@ -142,10 +141,7 @@ void LcdSpinBox::mouseReleaseEvent( QMouseEvent* )
 	if( m_mouseMoving )
 	{
 		model()->restoreJournallingState();
-
-		QCursor::setPos( m_origMousePos );
 		QApplication::restoreOverrideCursor();
-
 		m_mouseMoving = false;
 	}
 }
--- a/src/gui/widgets/LcdSpinBox.cpp
+++ b/src/gui/widgets/LcdSpinBox.cpp
@@ -129,8 +129,8 @@ void LcdSpinBox::mouseMoveEvent( QMouseEvent* event )
 			float fdy = static_cast<float>(dy);
 			if( event->modifiers() & Qt::ShiftModifier )
 				fdy = qBound( -4.f, fdy/4.f, 4.f );
-			float floatValNotRounded = model()->value() + m_remainder -
-										fdy / 2.f * model()->step<int>();
+			float floatValNotRounded =
+				model()->value() + m_remainder - fdy / 2.f * model()->step<int>();
 			float floatValRounded = roundf( floatValNotRounded );
 			m_remainder = floatValNotRounded - floatValRounded;
 			model()->setInitValue( floatValRounded );
@@ -189,6 +189,5 @@ void LcdSpinBox::enterValue()
 		model()->setValue( new_val );
 	}
 }
-
 
 
--- a/src/gui/widgets/LcdSpinBox.cpp
+++ b/src/gui/widgets/LcdSpinBox.cpp
@@ -41,6 +41,7 @@
 LcdSpinBox::LcdSpinBox( int numDigits, QWidget* parent, const QString& name ) :
 	LcdWidget( numDigits, parent, name ),
 	IntModelView( new IntModel( 0, 0, 0, NULL, name, true ), this ),
+	m_remainder( 0.f ),
 	m_mouseMoving( false ),
 	m_lastMousePos(),
 	m_displayOffset( 0 )
@@ -53,6 +54,7 @@ LcdSpinBox::LcdSpinBox( int numDigits, QWidget* parent, const QString& name ) :
 LcdSpinBox::LcdSpinBox( int numDigits, const QString& style, QWidget* parent, const QString& name ) :
 	LcdWidget( numDigits, parent, name ),
 	IntModelView( new IntModel( 0, 0, 0, NULL, name, true ), this ),
+	m_remainder( 0.f ),
 	m_mouseMoving( false ),
 	m_lastMousePos(),
 	m_displayOffset( 0 )
@@ -121,17 +123,20 @@ void LcdSpinBox::mouseMoveEvent( QMouseEvent* event )
 {
 	if( m_mouseMoving )
 	{
-		float dy = static_cast<float>(event->globalY() - m_lastMousePos.y());
-		if( event->modifiers() & Qt::ShiftModifier )
-			dy = qBound( -4.f, dy/4.f, 4.f );
-		if(dy > .5f || dy < -.5f) // that means "if dy != 0"
+		int dy = event->globalY() - m_lastMousePos.y();
+		if( dy )
 		{
-			float initValueNotRounded = model()->value() -
-						dy / 2 * model()->step<int>();
-			model()->setInitValue( roundf(initValueNotRounded) );
+			float fdy = static_cast<float>(dy);
+			if( event->modifiers() & Qt::ShiftModifier )
+				fdy = qBound( -4.f, fdy/4.f, 4.f );
+			float floatValNotRounded = model()->value() + m_remainder -
+										fdy / 2.f * model()->step<int>();
+			float floatValRounded = roundf( floatValNotRounded );
+			m_remainder = floatValNotRounded - floatValRounded;
+			model()->setInitValue( floatValRounded );
 			emit manualChange();
+			m_lastMousePos = event->globalPos();
 		}
-		m_lastMousePos = event->globalPos();
 	}
 }
 
--- a/src/gui/widgets/LcdSpinBox.cpp
+++ b/src/gui/widgets/LcdSpinBox.cpp
@@ -23,6 +23,7 @@
  *
  */
 
+#include <cmath>
 #include <QApplication>
 #include <QLabel>
 #include <QMouseEvent>
@@ -120,13 +121,14 @@ void LcdSpinBox::mouseMoveEvent( QMouseEvent* event )
 {
 	if( m_mouseMoving )
 	{
-		int dy = 2 * (event->globalY() - m_lastMousePos.y());
+		float dy = static_cast<float>(event->globalY() - m_lastMousePos.y());
 		if( event->modifiers() & Qt::ShiftModifier )
-			dy = qBound( -4, dy/4, 4 );
-		if( dy > 1 || dy < -1 )
+			dy = qBound( -4.f, dy/4.f, 4.f );
+		if(dy > .5f || dy < -.5f) // that means "if dy != 0"
 		{
-			model()->setInitValue( model()->value() -
-						dy / 2 * model()->step<int>() );
+			float initValueNotRounded = model()->value() -
+						dy / 2 * model()->step<int>();
+			model()->setInitValue( roundf(initValueNotRounded) );
 			emit manualChange();
 		}
 		m_lastMousePos = event->globalPos();
--- a/src/gui/widgets/LcdSpinBox.cpp
+++ b/src/gui/widgets/LcdSpinBox.cpp
@@ -127,8 +127,9 @@ void LcdSpinBox::mouseMoveEvent( QMouseEvent* event )
 		if( dy )
 		{
 			float fdy = static_cast<float>(dy);
-			if( event->modifiers() & Qt::ShiftModifier )
+			if( event->modifiers() & Qt::ShiftModifier ) {
 				fdy = qBound( -4.f, fdy/4.f, 4.f );
+			}
 			float floatValNotRounded =
 				model()->value() + m_remainder - fdy / 2.f * model()->step<int>();
 			float floatValRounded = roundf( floatValNotRounded );
@@ -189,5 +190,4 @@ void LcdSpinBox::enterValue()
 		model()->setValue( new_val );
 	}
 }
-
 
--- a/src/gui/widgets/LcdSpinBox.cpp
+++ b/src/gui/widgets/LcdSpinBox.cpp
@@ -120,7 +120,7 @@ void LcdSpinBox::mouseMoveEvent( QMouseEvent* event )
 {
 	if( m_mouseMoving )
 	{
-		int dy = 3 * (event->globalY() - m_lastMousePos.y());
+		int dy = 2 * (event->globalY() - m_lastMousePos.y());
 		if( event->modifiers() & Qt::ShiftModifier )
 			dy = qBound( -4, dy/4, 4 );
 		if( dy > 1 || dy < -1 )
--- a/include/Knob.h
+++ b/include/Knob.h
@@ -174,8 +174,8 @@ private:
 	BoolModel m_volumeKnob;
 	FloatModel m_volumeRatio;
 
-	QPoint m_mouseOffset;
-	QPoint m_origMousePos;
+	QPoint m_origMousePos; //!< position where user clicked to turn the knob
+	QPoint m_lastMousePos; //!< mouse position in last mouseMoveEvent
 	float m_leftOver;
 	bool m_buttonPressed;
 
@@ -196,6 +196,13 @@ private:
 
 	knobTypes m_knobNum;
 
+	enum class SetPosStatusType
+	{
+		NotChecked,
+		Works,
+		NoEffect
+	};
+	static SetPosStatusType s_setPosStatus;
 } ;
 
 #endif
--- a/include/Knob.h
+++ b/include/Knob.h
@@ -174,7 +174,6 @@ private:
 	BoolModel m_volumeKnob;
 	FloatModel m_volumeRatio;
 
-	QPoint m_origMousePos; //!< position where user clicked to turn the knob
 	QPoint m_lastMousePos; //!< mouse position in last mouseMoveEvent
 	float m_leftOver;
 	bool m_buttonPressed;
--- a/include/Knob.h
+++ b/include/Knob.h
@@ -196,13 +196,6 @@ private:
 
 	knobTypes m_knobNum;
 
-	enum class SetPosStatusType
-	{
-		NotChecked,
-		Works,
-		NoEffect
-	};
-	static SetPosStatusType s_setPosStatus;
 } ;
 
 #endif
--- a/include/LcdSpinBox.h
+++ b/include/LcdSpinBox.h
@@ -74,7 +74,7 @@ protected:
 
 private:
 	bool m_mouseMoving;
-	QPoint m_origMousePos;
+	QPoint m_lastMousePos; //!< mouse position in last mouseMoveEvent
 	int m_displayOffset;
 	void enterValue();
 
--- a/include/LcdSpinBox.h
+++ b/include/LcdSpinBox.h
@@ -73,6 +73,7 @@ protected:
 	virtual void mouseDoubleClickEvent( QMouseEvent * _me );
 
 private:
+	float m_remainder; //!< floating offset of spinbox in [-0.5, 0.5]
 	bool m_mouseMoving;
 	QPoint m_lastMousePos; //!< mouse position in last mouseMoveEvent
 	int m_displayOffset;


JohannesLorenz:
Review scope: The following should be done:

1. A functional code review
2. A test on Mojave (macOS): Does the fix has an effect?

I already tested that the "if" is not token on Linux, so a regression test on Linux is not required IMO.

@tresf Can you please try a review?

Spekular:
I think a regression test on Windows would be good as well, but it seems the build bot doesn't work on stable branch. Code looks good to me though, FWIW.

PhysSong:
I tested this on my MacBook with macOS 10.15.
- This works as expected If I turn off the accessibility setting.
- However, I can't move the knob well if I turn on it. I guess it takes some time to move the cursor after `QCursor::setPos()` calls.

JohannesLorenz:
> This works as expected If I turn off the accessibility setting.

Without accessibility setting is where `QCursor::setPos()` has no effect? E.g. this is where the bug was reported?

> I can't move the knob well if I turn on it

I assume that "with accessibility" means that `QCursor::setPos()` has an effect, but is too slow on macOS 10.15? If so, what about disabling `QCursor::setPos()` on macOS completely by using an `#ifdef`?

Spekular:
I know some parts of the code have a check in place that ignores mouse movements below a certain distance. If this check uses m_initialMousePos, could it be that it affects the workaround and leads to choppy movement?

JohannesLorenz:
@Spekular I can find `m_initialMousePos` only in `Track.h` (which is not related)

tresf:
> what about disabling `QCursor::setPos()` on macOS completely by using an `#ifdef`?

I feel strongly that this should be consistent across platforms.  macOS isn't the only platform that has this issue.  To that point:

1. We should aim to keep the user experience identical across similar platforms (e.g. mouse & keyboard)
1. We can't move the mouse position without accessibility settings; there should be no calls to  `QCursor::setPos()` anywhere in the code that are critical to UI operation (I consider knobs critical).
1. macOS isn't the only platform that this bug occurs on #5194's title has the words **"... and other systems"** because when this was being tracked on Discord other users reported this same issue on Linux suggesting that ANY Desktop environment that disallows the unauthorized movement of the cursor is going to suffer this bug.  The fix should take other systems into account.

The fix should be usable and testable on other platforms.  I'd be happy to test any patches on macOS, but it shouldn't be required for proof of concept.

JohannesLorenz:
I did know that the whole issue is not only on mac, but I thought the mouse jittering (as PhysSong described) was only on mac.

If that's not true, I'd say let's disable `QCursor::setPos` on all platforms. Advantage is a consistent, working solution. Disadvantage is that now even users with platforms where `QCursor::setPos` does work can't use it, so they may in some cases need to click multiple times to spin a knob.

JohannesLorenz:
Can we please continue the discussion to bring the PR to an end/to a merge?

tresf:
> Can we please continue the discussion to bring the PR to an end/to a merge?

The PR still uses `QCursor::setPos`, it's not a patch at all.

@Spekular what's up with the 👎?  If discussion is what we want, we should do it with words, not emojis.

Spekular:
Wantes to avoid lengthening the convo, but I disagree with causing a regression for Windows users to fix this. Having to torn a knob several times because you hit the "edge" of the screen is silly, to put it nicely.

tresf:
>  I disagree with causing a regression for Windows users to fix this. Having to turn a knob several times because you hit the "edge" of the screen is silly, to put it nicely.

Ok, but every other app behaves this way, it's consistent and it's more secure, as the control of one's mouse cursor should be limited to accessibility or automation software.

Have you considered:
1. When Apple makes these Desktop decisions, the other Desktops are likely to follow suit
2. The fact that this issue affects other platforms means we don't have to spend future time hunting down other workarounds

I like to use Carla as an example.  It's a well-designed UI, uses Qt just like we do and doesn't do this, which means a plugin that we integrate with has a more consistent knob behavior than our own.

I too thought that the loss of range was going to be a game changer, until I actually opened up and tried this on various apps that have knobs.  To my surprise, they all behaved the same and I was just accustomed to using the knobs that way.

SecondFlight:
FL Studio has an interesting solution to the issues with accessibility and touch screens, one it shares with Blender: instead of resetting the cursor on every mouse move, it simply jumps the cursor to the opposite side of the screen when it reaches the edge. FL still makes the cursor disappear, but blender does not.

I have not tried implementing this in anything yet, so I don't know if there are any pitfalls. In theory, this would also allow an easy fallback for environments that don't support mouse control, and without a significant difference in UX.

tresf:
> In theory, this would also allow an easy fallback for environments that don't support mouse control, and without a significant difference in UX.

Agreed, a simple cursor-jump would be easy to toggle off, as well as cursor visibility.

JohannesLorenz:
So you two propose that we reset the pointer to the other side of the screen if it reaches a border, and if e.g. MacOS does not support resetting the pointer, it will automatically stay at the border, leading to the discussed behaviour?

Is that correct? If yes, I could implement that.

It would sound like a fair compromise because it works on MacOS and does not limit Linux users.

tresf:
@JohannesLorenz I propose that the behavior is written to behave like Zyn, Carla.

@SecondFlight & @Spekular make the point that IF adding the additional mouse range (e.g. through the repositioning of the cursor) is possible using a technique similar to how FL Studio does it, it shouldn't hurt anything and will be easy enough to comment out if it causes issues with other platforms.

I'm not sure how you'd know if `setPos` is working, but hopefully there's a quick way to test this so that the feature can be turned on programatically instead of turned off at compilation time.

tresf:
> and does not limit Linux users

Or MacOS users that have granted LMMS accessibility access to permit this.

Keep in mind, some Linux distributions had issues with this `setPos` call too.  If we can handle it for any platform which restricts this call, we'll be in better shape regardless.  Another future example where this may cause strange issues is a draggable touch-interface.

JohannesLorenz:
Zyn (both old and new) just hit the border and then they stop.

> I'm not sure how you'd know if setPos is working, but hopefully there's a quick way to test this so that the feature can be turned on programatically instead of turned off at compilation time.

Why do we need to know if it works? If you're at the border, do `setPos` to the opposite side. If it suceeds, we have FL behaviour. If not, we have zyn behaviour.

tresf:
> > you'd know if setPos is working, 

>  if it works? [...] if [...] `setPos` suceeds,

Well that would mean it works. :D

JohannesLorenz:
OK, I'm about to write a solution, will submit in the next days.

JohannesLorenz:
This should work now as desired:

* When you pressed a knob, the hidden cursor now follows mouse movement.
* If it hits the border, it tries to flip on the opposite screen side. If that worked, you can continue moving the knob. If not, you'll need to click again.
* On releasing, the mouse pointer is always visible back on the knob.

Can we please test the following points?

* Linux: :heavy_check_mark: 
* Mac
  - with accessibility
  - without
* Windows
* Testing with two monitors, maybe even with two with different sizes

tresf:
@JohannesLorenz this works pretty well.  Some feedback... 

1. When LMMS is granted the following permissions, the "stationary" cursor works OK, but there's some issues... 
   a. When the cursor reached the edge of the screen, [LMMS segfaults](https://gist.github.com/tresf/67fe905c58e70fcf3d764f7ed2d26d47).
   b. The cursor occasionally shows up and disappears, making the experience
2.  Can we use an `#ifdef` to always show the cursor on MacOS?  This disappearing cursor is ok on a platform which successfully moves the mouse, but is very confusing for a platform that doesn't, it just makes the cursor disappear and reappear somewhere else, showing it the whole time is more organic in this instance.
3. Due to the points mentioned above, can we simply disable `setPos(...)` entirely on MacOS?

JohannesLorenz:
@tresf Thanks for the quick reply.

>        a. When the cursor reached the edge of the screen, [LMMS segfaults](https://gist.github.com/tresf/67fe905c58e70fcf3d764f7ed2d26d47).

I think it's the line 

https://github.com/JohannesLorenz/lmms/blob/c7c3e6f1e30173f7fbee3c4e5e4cac137802ac04/src/gui/widgets/Knob.cpp#L630

Probably the screen number is out of range, and it accesses a screen which is not in the screen array. Can you please add a debug printf or use gdb to confirm this?

>        b. The cursor occasionally shows up and disappears, making the experience cosmetically worse over existing knobs in other software.

I think this will be fixed by the `#ifdef` from point 2.

>     2. Can we use an `#ifdef` to always show the cursor on MacOS?  This disappearing cursor is ok on a platform which successfully moves the mouse AND successfully hides the cursor, but is very confusing for a platform that doesn't.  Testing with Accessibility turned off, it just makes the cursor disappear and reappear somewhere else, showing it the whole time is more organic in this instance.

The reappearing somewhere else would not be limited to MacOS. I think we should limit it to every platform that can not `setPos()`? (Because those can not set the cursor back at the end)

>     3. Due to the points mentioned above, can we ALSO simply disable `setPos(...)` entirely on MacOS?

I was always thinking it had no effect at all, so disabling or calling it would make no difference. What effect on MacOS do you suspect when calling `setPos()`?

JohannesLorenz:
About the segfault, this line I showed seems to not compile on Windows at all, so I'll probably have to change it completely.

tresf:
> What effect on MacOS do you suspect when calling setPos()?

Consistency.  :D

> About the segfault, this line I showed seems to not compile on Windows at all, so I'll probably have to change it completely.

Ok, I'll wait for that before running the debugger.

tresf:
> > What effect on MacOS do you suspect when calling setPos()?

> Consistency. :D

I'll elaborate... envision the cursor showing but NOT moving (when Accessibility is toggled on).

JohannesLorenz:
> envision the cursor showing but NOT moving (when Accessibility is toggled on).

You mean setPos() does not set the position, but show the previously hidden cursor? That would be a very buggy Qt implementation... :smile: If that's really true, I agree we may not call it at all.

tresf:
> You mean setPos() does not set the position, but show the previously hidden cursor? That would be a very buggy Qt implementation... 😄 If that's really true, I agree we may not call it at all.

No, I mean envision a system which shows the cursor but (after explicitly enabling the accessibilityy setting) allows it to be moved.  You get a very bad combination of show and non-moving cursor.

If you can detect whether or not the `setPos` works, we can hide the cursor only when `setPos` isn't available, but I spent 20 minutes researching how to do this today to no avail.

JohannesLorenz:
> No, I mean envision a system which shows the cursor but (after explicitly enabling the accessibilityy setting) allows it to be moved. You get a very bad combination of show and non-moving cursor.

Why do you have a non-moving cursor if the systen allows it to be moved? oO Maybe post a video?

> If you can detect whether or not the `setPos` works, we can hide the cursor only when `setPos` isn't available, but I spent 20 minutes researching how to do this today to no avail.

You'll see this if you call it and the cursor did not change :D

tresf:
> Why do you have a non-moving cursor if the systen allows it to be moved?

By current design, the "movement" brings it back to its origin, thus resulting in a non-moving yet visible cursor.

> You'll see this if you call it and the cursor did not change :D

This isn't a human observation it's a boolean flag.  If you think you can detect it, knock yourself out, but I have a feeling it's best to just abolish this nonsense.  In my opinion, the behavior just needs to go away on all platforms, but it's not worth my breath to argue something that's so easy to just toggle off two years from now when Linux and Windows also decide to lock down the cursor to just a list of approved accessibility programs.  When that time comes, I hard-link this comment, make the Mac behavior default, get a few thumbs up (to counter the thumbs down that are above) and humbly move on. :D

>  oO Maybe post a video?

Just do it yourself.  Make it visible and watch the ridiculous experience.  It really doesn't need a video.

JohannesLorenz:
I just commented the line

```
QApplication::setOverrideCursor( Qt::BlankCursor );
```

out in the mousePressEvent and it looks as I expected (on Linux): The mouse glides with the cursor, does flipping, and snaps back to the knob when releasing the key. No glitches or anything.

> By current design, the "movement" brings it back to its origin, thus resulting in a non-moving (or worse, jittering and wiggling) yet visible cursor.

On c7c3e6f, the only time where the cursor is brought back to its origin is when the mouse is released. I can easily remove that part, but I guess you don't mean that. Are you maybe not on the right commit (or forgot make/make install) @tresf?

tresf:
> On c7c3e6f, the only time where the cursor is brought back to its origin is when the mouse is released.

Ok, fine, but it's still bad UX, just not as jittery as I had painted in my head.  Regardless, I simply ask to always show cursor on Mac, never move.  That's all I'm asking.

JohannesLorenz:
OK, so your concern is, if I get it right, that the mouse cursor should be visible on Mac since it reappears somewhere else on the screen, correct? In this case, I think it should be that way for all platforms that can hide the cursor but not move it.

I wish we could auto detect that "can `setPos`" property...

JohannesLorenz:
If I don't find anything, I'll open an SO question. In the worst case, hack-solution or mac-ifdef. But I'm optimistic we'll find a solution.

tresf:
> I think it should be that way for all platforms that can hide the cursor but not move it.

Agreed 💯

tresf:
... and link the S.O. article so I can upvote it. :D :D :D

JohannesLorenz:
@tresf [Here is the link](https://stackoverflow.com/questions/60064492/find-out-if-qcursorsetpos-has-an-effect) to the SO question.

JohannesLorenz:
@tresf If there will be no useful answer from SO, what would be your preferred solution? `#ifdef LMMS_BUILD_APPLE` around the cursor hiding? Plus maybe that `setPos()`-check that does not work on mac (but maybe works right on any other system)? I think that's the best we can get...

(I'm meanwhile trying to get the CI errors fixed)

tresf:
I strongly feel that both `setPos()` and cursor hiding need to be disabled for MacOS.  No one's going to know to toggle this option on since Qt doesn't raise the standard security dialog.

If other platforms start to complain down the road, we can discuss how best to handle it.

JohannesLorenz:
@tresf Can you please test again with the `#5194-setpos` branch? It should work there on Mac as you expected.

tresf:
@JohannesLorenz this works great on MacOS tests pass no crashes.

A side-note.... I strongly discourage adding any untested code to `stable-1.2`, such as the logic which checks to see if the mouse movement was successful.  Without a positive failure, it's speculative code which may lead to undefined behavior.

If someone has a Linux/Unix machine which experiences this issue (perhaps some DE's or some combination of Display server) then I'd change my mind, but currently, I think it's best to leave it exactly as-is until we have evidence otherwise.  Furthermore, it gives us some time to do proper detection, such as in the even that the StackOverflow question is answered (or Qt adds this feature to the API, etc).

PhysSong:
> Qt doesn't raise the standard security dialog.

AFAIK, the dialog will appear only if you run the LMMS binary directly from the terminal, which is not normal for ordinary users.

PhysSong:
> If someone has a Linux/Unix machine which experiences this issue

I could confirm that `QCursor::setPos()` doesn't work on Wayland. It seems like [Wayland doesn't allow it](
https://github.com/qt/qtwayland/blob/143367312e0a6b1389c44cf62ca70f133b99c7b1/src/client/qwaylandcursor.cpp#L284).

tresf:
> I could confirm that QCursor::setPos() doesn't work on Wayland. It seems like Wayland doesn't allow it.

Well, there's your second instance of the problem and a test case @JohannesLorenz.

JohannesLorenz:
@PhysSong Wayland on what OS/distro? What happened with that branch `#5194-setpos`, did it hide the cursor while moving? Where was the mouse when you released the key?

PhysSong:
> What happened with that branch `#5194-setpos`, did it hide the cursor while moving? Where was the mouse when you released the key?

I tested the branch on GNOME on Wayland. When using XWayland(`QT_QPA_PLATFORM=xcb`), cursor works as if it's running on a normal X11 environment. With `wayland` plugin, however, `setCursor()` fails and I got some rare jitters(I don't know why).

JohannesLorenz:
@PhysSong Can you please answer all the questions?

PhysSong:
@JohannesLorenz On the `wayland` case, I can see the cursor while dragging. The cursor stays where it is when I finish dragging. The cursor moves smoothly, but the knob sometimes doesn't for some unknown reason.

PhysSong:
The jitter I experienced might be an issue with my system, so I will test with another machine and OS.

PhysSong:
> The jitter I experienced might be an issue with my system, so I will test with another machine and OS.

It turned out that libraries on my system were too old, so ignore it.

JohannesLorenz:
OK, so what do we do now with this branch?

1. Just merge now?
2. Further testing (what platforms?), then merge?
3. Reject?

@tresf @PhysSong

PhysSong:
How about new changes in the `#5194-setpos` branch?

JohannesLorenz:
> How about new changes in the `#5194-setpos` branch?

Thanks for the reminder! I almost forgot this. I now merge the `#5194-setpos` into this PR.

JohannesLorenz:
Still, the questions remain. What shall we do now?

1. Just merge now?
2. Further testing (what platforms?), then merge?
3. Reject?

Please give me both your opinion @PhysSong @tresf

tresf:
The mac behavior isn't ideal, but it's much, much better.

For some reason the knob moves more depending on how fast you move the mouse.  If you move the mouse slowly, it moves a much smaller range, which isn't consistent with apps like GarageBand which move the same amount regardless of speed/acceleration of the cursor.

Here's an illustration about the problem.  It's workable though.  It's better than the alternative.

First segment: New behavior
Second segment: Old Behavior

ezgif-2-68ce442375f9

tresf:
Garageband for comparison:
ezgif-2-f65b9b07ef74

JohannesLorenz:
Strange, on Linux I can't confirm this - the knob angle increases independent of mouse movement speed. I thought the Linux and apple code would be the same, except for the visuals of the cursor and the edge flipping.

I can't figure it out with just looking at the code. We can let someone with a MAC debug it, or we can say it's "stable enough" to close this PR.

tresf:
Just a guess, but it may be related to https://github.com/LMMS/lmms/issues/5194#issuecomment-594910346.

Anyway, I'm fine with good enough.  It's better than not working. :D

tresf:
Disregard the above, that's about moving the mouse, not tracking its movement.  I have no idea why acceleration is bugged, but I'm fine with it as-is for now.

PhysSong:
> The knob moves a different range depending on the speed of the cursor

I explained it at https://github.com/LMMS/lmms/issues/5194#issuecomment-535359322 a while ago.

BTW, it seems like LCD spin boxes suffer from a similar issue. I think it might be better to add a helper class for `QCursor::setPos()` and use it where we used `QCursor::setPos()` before.
Alternatively, we can remove some `QCursor::setPos()` calls for consistency and better UX for tablet inputs.

tresf:
Ah yes, thanks for the link.  Can we get rid of that?  It's undefined behavior.  Most desktops have accelerations in the position already so adding to it seems like it'll just make usiblity harder for eternity.  Ok if the answer is "yes, but on `master`" :D

JohannesLorenz:
So merge this and open another issue/PR? Do you two agree?

Is the other issue/PR to be fixed within 1.2?

tresf:
Yes merge, no don't hold up stable for an age-old mouse acceleration bug.

PhysSong:
Can we fix LCD spin boxes as well? They suffer from a similar issue because they use `QCursor::setPos()` as well.

JohannesLorenz:
> no don't hold up stable for an age-old mouse acceleration bug

Is it age-old, or introduced by this PR? I'm not sure.

> Can we fix LCD spin boxes as well?

If it really does not work for them, too, I think we should do this in this PR.

PhysSong:
> Is it age-old, or introduced by this PR? I'm not sure.

I think it's a hidden issue which this PR has revealed.

JohannesLorenz:
> Can we fix LCD spin boxes as well? They suffer from a similar issue because they use `QCursor::setPos()` as well.

```
$ grep QCursor::setPos -r include/ src/ | awk -F: '{print $1}' | tr -d ':' | uniq
src/core/Track.cpp
src/gui/widgets/Graph.cpp
src/gui/widgets/Knob.cpp
src/gui/widgets/LcdSpinBox.cpp
src/gui/editors/PianoRoll.cpp
src/gui/editors/AutomationEditor.cpp
```

:sob: 

Shouldn't we repair them all?

PhysSong:
I checked all of them, but I experienced problems only with knobs and LCDs.

JohannesLorenz:
> I checked all of them, but I experienced problems only with knobs and LCDs.

Thanks for checking.

Before I'll look at the LCDs I still want to fix the knob angle depending on the mouse speed. As I asked @tresf , he said a nervous hand could lead to a sudden increase of knobs like volume. Imagine this e.g. for the CALF LADSPA Compressor, which can increase its input volume by factor 20. If the whole UI is full of dangerous knobs, I want to investigate this.

JohannesLorenz:
Can please someone help? Before the PR, it was

```
setPosition( _me->pos() - m_origMousePos );
```

and after the PR:

```
setPosition( _me->pos() - m_lastMousePos );
```

This *should* result in the same `setPosition` calls, because before the PR, the cursor was constantly moved back to `m_origMousePos`, e.g. what was the original mouse position is now the last mouse position.

So if these are the same `setPosition` calls, why is the behavior now different for Mac users? The Mac changes only affect `setPos`, which is only used to flip at the borders.

PhysSong:
Maybe because `setPos()` is asynchronous on Mac? That's the only possibility I could find.

tresf:
> So if these are the same setPosition calls, why is the behavior now different for Mac users? The Mac changes only affect setPos, which is only used to flip at the borders.

I didn't think it is behaving differently.  What do you mean?

tresf:
Oh, I think you're inquiring why this patch helps Apple at all over the old logic?  I don't know the answer.

JohannesLorenz:
> Oh, I think you're inquiring why this patch helps Apple at all over the old logic? I don't know the answer.

No, I don't mean that. The knob angle increase is dependent on speed on Apple, but not on my Arch Linux.

> Maybe because `setPos()` is asynchronous on Mac? That's the only possibility I could find.

FWIR I removed all `setPos()` calls in Mac. Maybe you meant the mouse move callbacks (`Knob::mouseMoveEvent`) are asynchronous?

JohannesLorenz:
Imagine the following: On the computer from @tresf , somehow, the `Knob::mouseMoveEvent` gets called less often. If he makes a fast mouse movement, it's maybe just 1 event for him, but 10 on my PC. Now, from https://github.com/LMMS/lmms/issues/5194#issuecomment-535359322 we know that the transition functions is partially exponential. This turns 1 mouse event with a large jump to a way larger angle increase than 10 mouse movements with 1/10 jumps each.

Wouldn't that explain it? Would that mean we should make the transition function linear?

PhysSong:
@JohannesLorenz You said you need some help. What can I do?

JohannesLorenz:
> You said you need some help. What can I do?

Try to answer the questions from my last post :smiley:

PhysSong:
> > You said you need some help. What can I do?
> 
> Try to answer the questions from my last post
Do you mean https://github.com/LMMS/lmms/pull/5360#issuecomment-620835854? If so, I think that may make sense.

JohannesLorenz:
@tresf I pushed a new commit 35eaa0a to this PR. Can you please check if this fixes the knob increase depending on mouse speed?

tresf:
1. It's linear (much better!) now
2. It segfaults when I hit the bottom of the screen
3. It's too insensitive now (it's not fast enough).  To go from 0 to MAX on the default tripleoscillator track, I have to go to the top screen edge 6x.  I would consider 1x, 2x or a max of 3x knob-to-screen edge travel necessary for this action.

Veratil:
>     3. It's too insensitive now (it's not fast enough).  To go from 0 to MAX on the default tripleoscillator track, I have to go to the top screen edge 6x.  I would consider 1x, 2x or a max of 3x knob-to-screen edge travel necessary for this action.

Keyboard modifier for movement divider/multipler?

tresf:
> Keyboard modifier for movement divider/multipler?

It behaves a bit like shift is pressed, but no, sane defaults.  Always. 🙂

src/gui/widgets/Knob.cpp
```diff
@@ -618,12 +645,55 @@ void Knob::mousePressEvent( QMouseEvent * _me )
...
+		int ymax = qMax(rec.height() - 2, 0), xmax = qMax(rec.width() - 2, 0);
+		int ymin = 1, xmin = 1;
+
+		// calculate new coordinates
+		int x = QCursor::pos().x(), y = QCursor::pos().y();
+		QPoint newPos(
+			(x<xmin) ? xmax : (x>xmax) ? xmin : x,
+			(y<ymin) ? ymax : (y>ymax) ? ymin : y);
+
+		auto setCursor = [&](const QPoint& p) -> bool
```
PhysSong, line 672: Can you use `s_setPosStatus` or cache the checked result, if that works?

src/gui/widgets/Knob.cpp
```diff
@@ -618,12 +645,55 @@ void Knob::mousePressEvent( QMouseEvent * _me )
...
+		setPosition( _me->pos() - m_lastMousePos );
 		emit sliderMoved( model()->value() );
-		QCursor::setPos( mapToGlobal( m_origMousePos ) );
+
+		// the rest of the code updates the cursor and/or m_origMousePos
+
+		// get screen min/max values where a flip is not required yet
+		const QRect& rec =
+#if QT_VERSION >= 0x050A00
+			QApplication::screenAt(QCursor::pos())->geometry();
```
PhysSong, line 659: It seems like `QApplication::screenAt(QCursor::pos())` returns a null pointer on macOS if the cursor hits the bottom of the screen.

JohannesLorenz:
> It's too insensitive now (it's not fast enough).  To go from 0 to MAX on the default tripleoscillator track, I have to go to the top screen edge 6x.  I would consider 1x, 2x or a max of 3x knob-to-screen edge travel necessary for this action.

Why is it not too insensitive on my Linux then? The behaviour should not be any different.

I'm out of ideas. We could rewrite the whole code, such that not the difference to the last mouse point, but only the difference to the original mouse point is counted. However, with linear scales, the behaviour should be excatly like it is now, so I fear this won't solve the mouse being to slow on @tresf 's computer.

PhysSong:
I tested it both on a MacBook Pro(macOS, 2880x1600) and another laptop(Linux, 1366x768). In both cases, knobs moved significantly slower than before. As I mentioned before, I think the real issue is the multiplier.

tresf:
> Why is it not too insensitive on my Linux then? The behaviour should not be any different.

I'd guess trackpad vs. mouse.  Mice have the luxury of having a virtually unlimited glide surface.  Trackpads are limited to their edges.

Remember, the <kbd>Shift</kbd> modifier is available to help with this.  Also, if Windows and Linux hate the Mac sensitivity, we can add an `#ifdef` around the Mac multiplier.

JohannesLorenz:
I give up this PR. I'm tired trying to provide something which again and again does not work on MAC (already for 6 months now). This PR must be finished from someone who has a MAC.

tresf:
@JohannesLorenz the PR's fine.  @PhysSong said it happens on Linux too.  Mac just exposed a problem with our code.  To call this a Mac problem is a bit short-sighted.  If you're asking for @PhysSong or myself to tweak the multiplier to a sane value, one of us can.  I'm not a C++ coder (and don't proclaim to be) so I've only assumed the role of testing on this one.  That said, I can roll up my sleeves and find a sane value, but it will be a few days before I can tackle it.

Veratil:
Interestingly enough if run in a VM and your mouse goes outside of the screen (guest additions to not capture mouse), it instantly jumps to min/max. 😂

In testing though for me the movement is smooth on Linux.

tresf:
> Interestingly enough if run in a VM and your mouse goes outside of the screen (guest additions to not capture mouse), it instantly jumps to min/max. 😂

I use Parallels for my VMs, never had an issue with the cursor if the guest tools are installed properly.

> In testing though for me the movement is smooth on Linux.

It's fantastic for me on Mac too, but the default sensitivity is just too low.  I can USB+persistance boot Ubuntu 20.04 on my HP Spectre 360 and see how it is on the HP trackpad, but it'll take some time to setup, so a proper evaluation is going to be at least a few days.  If someone beats me to this, LMK.

JohannesLorenz:
> If you're asking for @PhysSong or myself to tweak the multiplier to a sane value, one of us can

Why change it? It's working perfect on my Arch Linux PC now. If you change it, it will be worse for me. Should this PR really decrease usability on some platforms?

Veratil:
> 
> 
> > Interestingly enough if run in a VM and your mouse goes outside of the screen (guest additions to not capture mouse), it instantly jumps to min/max. 😂
> 
> I use Parallels for my VMs, never had an issue with the cursor if the guest tools are installed properly.

I'm not saying it doesn't work properly, it does work properly, it's just interesting that the sudden jump happens. I have it printf'ing out the mouse position data, it looks like the last mouse position data gets set incorrectly somehow as soon as the mouse goes off screen (VM window is 1600x900 within my 1920x1080).

```
omp (10, 13)  lmp (32, 65)  pos (32, 66)
omp (10, 13)  lmp (32, -832)  pos (33, 66)

omp = m_oldMousePos
lmp = m_lastMousePos
pos = _me->pos()
```

This jump is what happens which launches the knob to min/max.

> 
> > In testing though for me the movement is smooth on Linux.
> 
> It's fantastic for me on Mac too, but the default sensitivity is just too low.

For me here are the step values:
Volume and panning it's increasing by 0.2% at a time, with shift 0.1%.
Floats are 0.01 at a time, with shift 0.0025.
Integers are 1 at a time, with shift 0.5 (which keeps it at previous value until two "pixels" instead of one).

What steps would feel more comfortable to you? For me, I think this is fine for what it does. I like the smaller movement precision myself.

tresf:
> Why change it? It's working perfect on my Arch Linux PC now. If you change it, it will be worse for me. Should this PR really decrease usability on some platforms?

This seems unnecessarily hostile attitude.  I've offered to test on Ubuntu 20.04 and work on a sane value.  It's MacOS which exposes the mouse problem to begin with.  Windows, Linux and Mac should be tested/evaluated and coded for.  That's how we do things.

Platform by popularity (1.2.1 using Downloads page as a metric):
* Linux (AppImage): 71,413 (8%)
* Windows: 772,143 (86%)
* MacOS: 53,041 (6%)

Linux results are slightly askew due to package managers, but those are still real human beings using the software.

JohannesLorenz:
> > Why change it? It's working perfect on my Arch Linux PC now. If you change it, it will be worse for me. Should this PR really decrease usability on some platforms?
> 
> This seems unnecessarily hostile attitude.

I'm just saying that I think this isn't just fixed by changing a constant. Apologies for my hostility, I found the sarcasm necessary here to draw more attention on it.

Right now, I think the best solution would be to let the user change the mouse sensitivity in the settings. That's already common for many computer games.

So I see 3 solutions:
1. Use some linear factor in between "Linux" and "Mac/some Linuxes" that may be sub-optimal for both
2. Spend months to find out why "Linux" and "Mac/some Linuxes" behave so differently and writing a very complicated algorithm that works on all
3. Let the user simply adjust the factor in the settings. This has the additional advantage that users who have fast/slow hands/input devices can adjust the settings to their personal wishes

tresf:
It doesn't take months to test sensitivity values.  If the Mac sensitivity sucks, we fix it.   The <kbd>Shift</kbd> key already lessens the sensitivity.  Comparing this to video games is making a mountain out of a molehill.  A new sensitivity slider shouldn't land on a bug-fix only branch.

Veratil:
4. Use constants per operating system for now until we figure out a better solution.

We already use blocks for Apple and the cursor, shouldn't be any different to change the constant the same way.

src/gui/widgets/Knob.cpp
```diff
@@ -497,8 +499,8 @@ float Knob::getValue( const QPoint & _p )
 {
 	float value;
 
-	// arcane mathemagicks for calculating knob movement
-	value = ( ( _p.y() + _p.y() * qMin( qAbs( _p.y() / 2.5f ), 6.0f ) ) ) / 12.0f;
+	// knob value increase is linear to mouse movement
+	value = .09f * _p.y();
```
tresf, line 503: On Mac using the trackpad, a `0.4x` multiplier is pretty good.  Will test on Ubuntu 20.04 next.

```cpp
value = .4f * _p.y();
```

Veratil: Fedora 30 VM, `.4f` works well for me!

tresf: On Ubuntu 20.04 running in a VM, the existing `.09x` multiplier seems just fine.  Will continue investigating why there's a disparity.

tresf: Forcing `SetPosStatusType::NoEffect` on Ubuntu 20.04 reproduces the sensitivity issue with macOS.  My theory is that this sensitivity is a result of the `SetPosStatusType::Works` logic.  Will post in the main thread.

tresf:
> Fedora 30 VM, .4f works well for me!

Same if I force `SetPosStatusType::NoEffect` on Ubuntu.  @JohannesLorenz I believe the issue is related to the `SetPosStatusType::Works` code.  I'd theorize that Fedora (an OS which ships with Wayland enabled by default) is having issues moving the cursor and experiencing the same behavior as Mac.

Ubuntu 20.04 by default ships with X11 enabled by default.  If @Veratil switches Fedora 30 to Gnome with X11, the existing sensitivity might work.

Veratil:
> 
> 
> > Fedora 30 VM, .4f works well for me!
> 
> Same if I force `SetPosStatusType::NoEffect` on Ubuntu. @JohannesLorenz I believe the issue is related to the `SetPosStatusType::Works` code. I'd theorize that Fedora (an OS which ships with Wayland enabled by default) is having issues moving the cursor and experiencing the same behavior as Mac.
> 
> Ubuntu 20.04 by default ships with X11 enabled by default. If @Veratil switches Fedora 30 to Gnome with X11, the existing sensitivity value might work.

I'm using X11, not Wayland.

tresf:
So this is even stranger, but it appears the logic which toggles `QCursor::setPos(...)` somehow breaks my VM's mouse pointer integration and I have to reboot between tests.  This may be specific to the hypervisor, Qt version or my machine, but it worries me.

tresf:
> I'm using X11, not Wayland.

Thanks.  Is the cursor shown or hidden?

Veratil:
> 
> 
> > I'm using X11, not Wayland.
> 
> Thanks. Is the cursor shown or hidden?

Hidden, but doesn't hold position (probably because VM and host OS has control).

tresf:
> Hidden, but doesn't hold position (probably because VM and host OS has control).

These types of nuances really worry me.  Especially considering quite a few people test in VMs for various reasons.

Unpopular opinion, but I still feel this comment is valid and will yield cleanest code: https://github.com/LMMS/lmms/pull/5360#issuecomment-570630645.  We now have quite a disparity of behavior.

PhysSong:
I agree with dropping the cursor magic completely. I think it's nearly impossible to write a code without that which works cosistently.

src/gui/widgets/Knob.cpp
JohannesLorenz: I [tried](https://github.com/JohannesLorenz/lmms/pull/4/commits/33dd543f7f07f667efc29a3fa8e4c8ecfcc39ed5), but somehow this causes the knob resetting to the opposite site even on native Arch Linux where `setPos` works.

JohannesLorenz:
> I agree with dropping the cursor magic completely. I think it's nearly impossible to write a code without that which works cosistently.

If you mean the border flip logic that I implemented in this PR, I agree, it's too buggy and too risky for `stable-1.2`.

JohannesLorenz:
Right now I have a different thought. What if OS/distro are not the relevant factor, but screen size and resolution are? Did you consider this during all the tests?

My idea is this: Normally, if you configure your window manager, you'll make sure that moving from bottom to top with the mouse is doable quick enough, but not to quick. So "bottom to top" is a good measure. However, bottom to top is not measured in pixel, and our code currently measures pixel. So imagine you have a 4K and a HD screen. In both, you'll make sure moving the mouse from bottom to top takes about the same hand movements. But when you use our knob code, the 4K user will have the knob increase with double speed (if I recall our code correctly).

Veratil:
> ... our code currently measures pixel.

> But when you use our knob code, the 4K user will have the knob increase with double speed (if I recall our code correctly).

Isn't this contradictory though? Unless with the second you mean current code, in which case I don't know the details of that.

The question of knob value by mouse movement I believe comes down to this: per pixel distance from the origin of the event, do we want the value to increase linearly or non-linear?

Linear I think will make all knobs move the "same", but feel different based on their min/max. This different feel might have been where the original magic formula came from (or not, I'm making a guess about it), to try and smooth all knobs to feel the same.

Non-linear we can play with and make movement feel the same regardless of min/max value, but this takes time to figure out to work with any min/max.

JohannesLorenz:
> > ... our code currently measures pixel.
> 
> > But when you use our knob code, the 4K user will have the knob increase with double speed (if I recall our code correctly).
> 
> Isn't this contradictory though? Unless with the second you mean current code, in which case I don't know the details of that.

Both code bases (before and in this PR) take the y pixel distance of two cursor points, subtract them (the subtraction result is `p` below), and take only that as input for their formulae:

```
void Knob::setPosition ( const QPoint & _p )
{
    // ...
    // old
    value = ( ( _p.y() + _p.y() * qMin( qAbs( _p.y() / 2.5f ), 6.0f ) ) ) / 12.0f;
    // new
    value = .09f * _p.y();
    // ...
}

void Knob::mouseMoveEvent( QMouseEvent * _me )
{
    // ...
    m_mouseOffset = _me->pos() - m_origMousePos; // differs between HD and 4K
    setPosition( m_mouseOffset );
    // ...
}

```

So if you change from HD to 4K and adjust the mouse in a way that it takes the same hand movements to move from bottom to top, the number of pixels double, which doubles the knob increasement in the linear formula.

PhysSong:
It seems like Carla divides the pixels with the size of the knob:
https://github.com/falkTX/Carla/blob/20cc5244d08de13668a36749b7b165507b4805bf/source/frontend/widgets/pixmapdial.py#L328-L329

We may somehow scale the pixels according to DPI, if possible. One issue is that Qt < 5.6 doesn't support HiDPI well.

JohannesLorenz:
For information, I pushed an alternative PR #5491 for testing with the following differences:

* No setPos/cursor hiding/border flip at all (all operating systems equal)
* Knob increase is now depending on screen height (DPI might be better, as @PhysSong noted, but let's just try this)

JohannesLorenz:
I pushed 28e7848 now which is @tresf 's proposal:

* no `setPos`/cursor-hide/border-flip at all (on no system)
* linear transition function, factor 0.4

It's almost the same as the alternative PR, except that this PR does not take the screen height into account. (I.e. tests with the same platform with 4K vs HD would be interesting)

zonkmachine:
> * linear transition function, factor 0.4

I don't see this change in there.

JohannesLorenz:
> > * linear transition function, factor 0.4
> 
> I don't see this change in there.

Sorry, I forgot to change the factor. It should be fixed now.

For my system (Manjaro native, 1080 height), it's faster than on master. It's still controllable, but 0.2 feels more like "master" to me than 0.4. How is it working on your 600 height notebook @zonkmachine ? Does 0.4 give you a master-like feeling?

tresf:
I still receive a segfault when the mouse hits the bottom of the screen.

tresf:
> I still receive a segfault when the mouse hits the bottom of the screen.

Scratch that, I had made a mistake fast-forwarding, it's fixed.

zonkmachine:
>  Does 0.4 give you a master-like feeling?

Yes. I think master is perhaps slightly faster. 0.2 is a bit on the slow side.

zonkmachine:
If there is trouble finding a good value for this, can't we just add a setting so the user can fix this on their own?

tresf:
Testing on Ubuntu 20.04 on a physical machine (switching between stable-1.2 and this PR) there's a noticeable speed-up in sensitivity.  Ubuntu 20.04 ships with Qt 5.12 so the HiDPI seems fine (minus the hand cursor being unusually small).

That said, I find the new range to be closer to other apps.  Comparing it to Audacity -- which uses almost a pixel-for-pixel control, it still offers more granularity than Audacity.  Shift works fine to make it even more granular.

I generally compose on a laptop making left/right swiping and up/down swiping the norm (superior in some ways to using the mouse to locate a scroll bar).  When using these scroll gestures, the DAW responds quickly and navigates fast.  To bring the knobs slower by default to match the old behaviour seems to be a change in the wrong direction in my opinion but I'd be happy to hear other people's thoughts.

I find 0.4f to be good on Ubuntu 20.04 and macOS (both HiDPI).  Standard DPI and Windows should be tested and we should come to a consensus on sensitivity.

> If there is trouble finding a good value for this, can't we just add a setting so the user can fix this on their own?

Since we won't be able to influence sensitivity on things like scrollbars, my instinct is to NOT add more configuration options to the software.  I'm a firm believer that user interfaces should be useable out of the box with sane defaults.  This type of granularity just requires more changes to a bug-fix only branch and will be overlooked by most users.  The sensitivity is something we'll all have to adjust to.

If deciding a good value is too hard, I would advocate for copying Carla's sensitivity.

JohannesLorenz:
0.4 works on my native Arch Linux. Interestingly, it makes sense both with HD and 4K resolution. I was assuming when you use 4K, you'll double the mouse speed to faster navigate through the screen. However, in the end, I kept the mouse speed in order to easily hit small widgets. This seems to indicate that the resolution does not have much influence (and my alternative PR is wrong).

I agree it's slightly faster than on master now, but I have no opinion about it.

Maybe this solution is not super for everyone, but OK for all, and fixes the original issue. So if there are no more complaints, and it has been tested on Windows, I'd say let's do code and style review and then merge it ASAP.

tresf:
> Maybe this solution is not super for everyone, but OK for all

💯 

I know some really like the hidden/wrap-around proposal (for good reason) but this new patch is smoother and consistent on all platforms I've tested so far.

JohannesLorenz:
I fixed the forgotten `LcdSpinBox`:

* Like with `Knob`, I removed the code that resets the cursor using `setPos`
* The scale type seemed already linear
* I added a factor "3" now to get a sane feeling

src/gui/widgets/LcdSpinBox.cpp
```diff
@@ -121,16 +120,16 @@ void LcdSpinBox::mouseMoveEvent( QMouseEvent* event )
...
 	if( m_mouseMoving )
 	{
-		int dy = event->globalY() - m_origMousePos.y();
+		int dy = 3 * (event->globalY() - m_lastMousePos.y());
 		if( event->modifiers() & Qt::ShiftModifier )
 			dy = qBound( -4, dy/4, 4 );
 		if( dy > 1 || dy < -1 )
 		{
 			model()->setInitValue( model()->value() -
 						dy / 2 * model()->step<int>() );
```
PhysSong, line 129: I think yo may multiply the factor here instead of changing `dy`.

JohannesLorenz: I saw that, too, but then, the ratio between non-shift- and shift-movement would have changed. The way it is now, the `dy` values should be close to those of `master`, and so the `qBound( -4, dy/4, 4 )` should have the same effect.

JohannesLorenz:
@tresf Would you mind doing some tests for `LcdSpinBox`, too?

tresf:
> @tresf Would you mind doing some tests for LcdSpinBox, too?



Sure.

The short: 

It feels fine.  From a click+drag perspective, it could probably afford to be a bit less sensitive, but when comparing it to the current non-trackpad-friendly mouse-wheel emulation, it's far superior.

The long:

When comparing it to the mouse-wheel emulation there are quite a few of nuances.  First (and I consider this to be correct, consistent behavior, but others may disagree) is that it scrolls the opposing direction as the mouse wheel simulation on a default trackpad (two finger scroll).  This is due to Apple's decision to "scrub" like a touch screen by default (they call this "natural scrolling", but it's more accurately describing as "scrubbing" or "pushing" where you shove content into a direction instead of emulate a wheel).  This means scrolling down using the wheel-gesture increases the number (try this gesture on a smartphone if descripting is confusing).

Anyway, the mouse-wheel emulation (which is NOT the topic of this PR) is completely bazaar on Mac.  Disregarding the reverseness, the scrolling emulates a list/document moving at an animated pace.  This isn't how it behaves in apps like GarageBand.  Infact, GarageBand LOWERS the master volume like one would expect when scrolling down.  This can't be easy to code/assume since it's opposite of the "natural" or "scrubbing" direction.   Since most UI elements in GarageBand don't respond to the wheel, (e.g. master volume works, track volume does not work)  this can probably be treated as an edge-case behavior which will be reported at a future date, feel free to link to this TL;DR blurb. but even more oddly, the LMMS MASTER VOLUME SLIDER works properly with this gesture whereas the knobs and spinboxes do not.  This may mean we have a bug somewhere?  Again, it's not directly related to this PR but someone will report it as a bug down the road so it's nice to have documented.

Back on topic... In regards to the click spinbox sensitivity, it feels though as if a single pixel is the incrementing unit on the Mac trackpad.  this sensitivity is a bit too aggressive in my opinion and is not consistent with the range imposed by the knobs.  I'd ask that the sensitivity is reduced by at least 50%.

I have not tested on any other platforms.

JohannesLorenz:
> This may mean we have a bug somewhere? 

OK, but you don't consider this critical for stable-1.2, right?

> this sensitivity is a bit too aggressive in my opinion and is not consistent with the range imposed by the knobs. I'd ask that the sensitivity is reduced by at least 50%.

I decreased from 3 to 2 now (exactly 50%). For me, it's still way to fast for small numbers, but the PR does not change the speed significantly. Using any number < 2 (e.g. 1.9f) requires to change more logic, so can you please test 2 for now?

tresf:
> OK, but you don't consider this critical for stable-1.2, right?

No, it's probably been around since 2011 when "natural scrolling" was added to MacOS 10.7.  Filed a separate bug report here: #5507

> I decreased from 3 to 2 now (exactly 50%). For me, it's still way to fast for small numbers,

Agreed, it's still not ideal.  I tried using a float but every number I changed just made it worse.  For the sake of merging, we can leave it as-is (no changes) and merge.  If we do that it should be fixed on master.  If it's a quick fix, I'd be happy to do more testing.  The knob fixes are fantastic.

JohannesLorenz:
> If it's a quick fix, I'd be happy to do more testing

Thanks for the offer. I'll look at how quick it fixes at the end of the week.

JohannesLorenz:
@tresf I pushed a commit which sets the factor down to `1` (and fixes a few other issues in the `LcdSpinBox` calculation).

Can you please re-test? We can now easily change the factor to any value.

The song editor has the TEMPO/BPM spinbox (10-999), where this PR is a bit slow (but you can enter "999" by double click), and it has the TIME SIG spinbox (1-32) where I find this PR a bit too agressive. IMO the speed should be dependent on the range of the spin box, but I'm not sure if this should be fixed on stable-1.2. I think the PR is usable (though not perfect).

JohannesLorenz:
Sorry, just had to push a minor fixup. Please test **now** :smiley:

tresf:
There's still something that feels off about the sensitivity.  Using the PITCH RANGE knob on the instrument window, it's extremely hard to move it from `1` -> `2`.  It usually jumps to `3`.  Worse, it doesn't really go back down properly.  If I move the mouse slow enough it never changes numbers at all.  This means when I drag slowly to correct the large jump in numbers, I'm rewarded with it never moving.  This results in a very sporadic use.  It's not usable.

Meanwhile, the knobs are stellar.  They don't seem to suffer from slowly moving cursor syndrom.

tresf:
Here's a video, which I think explains it much better....

JohannesLorenz:
Sorry, I somehow missed that issue. The "punishment" was due to the `LcdSpinBox` doing some divisions and not recalling the floating offset (after the comma). So if you were too slow (or your system sent mouse events too often), the numbers got stuck.

Should be fixed now, can you please re-test @tresf ?

tresf:
@JohannesLorenz as you've already identified, it's a bit sensitive, but it works perfectly now, exactly as expected. 👍 👍 👍

JohannesLorenz:
@tresf Great. I can still change the speed easily. Should I try something like 0.75 or 0.5, or do you prefer that sensitivity?

tresf:
> @tresf Great. I can still change the speed easily. Should I try something like 0.75 or 0.5, or do you prefer that sensitivity?

I think you made a very good point earlier about granularity based on range.  For now, I think it's workable.  If you can make the pitch go from 1 to 2 on an instrument with a mouse without any struggle, I feel the trackpad users can deal with the hypersensitivity until the range is handled better down the road.  It fixes the issue and makes it better, which is more than we can ask for on the stable branch.

JohannesLorenz:
OK then it's FINALLY time for reviews:

1. functional code review
2. style review
3. test review

About 3:

* @tresf which systems did you test? I guess at least Mojave (macOS)?
* I tested Manjaro Linux (native)
* We probably need a windows test again.

src/gui/widgets/LcdSpinBox.cpp
```diff
@@ -121,15 +123,19 @@ void LcdSpinBox::mouseMoveEvent( QMouseEvent* event )
...
+		int dy = event->globalY() - m_lastMousePos.y();
+		if( dy )
 		{
-			model()->setInitValue( model()->value() -
-						dy / 2 * model()->step<int>() );
+			float fdy = static_cast<float>(dy);
+			if( event->modifiers() & Qt::ShiftModifier )
+				fdy = qBound( -4.f, fdy/4.f, 4.f );
+			float floatValNotRounded = model()->value() + m_remainder -
+										fdy / 2.f * model()->step<int>();
```
Veratil, line 133:
```suggestion
			float floatValNotRounded =
				model()->value() + m_remainder - fdy / 2.f * model()->step<int>();
```

Or something similar. All those extra tabs aren't needed IMO.

src/gui/widgets/LcdSpinBox.cpp
```diff
@@ -121,15 +123,19 @@ void LcdSpinBox::mouseMoveEvent( QMouseEvent* event )
...
-			dy = qBound( -4, dy/4, 4 );
-		if( dy > 1 || dy < -1 )
+		int dy = event->globalY() - m_lastMousePos.y();
+		if( dy )
 		{
-			model()->setInitValue( model()->value() -
-						dy / 2 * model()->step<int>() );
+			float fdy = static_cast<float>(dy);
+			if( event->modifiers() & Qt::ShiftModifier )
+				fdy = qBound( -4.f, fdy/4.f, 4.f );
```
Veratil, line 131:
```suggestion
			if( event->modifiers() & Qt::ShiftModifier ) {
				fdy = qBound( -4.f, fdy/4.f, 4.f );
			}
```

Explicit braces is all.

src/gui/widgets/LcdSpinBox.cpp
JohannesLorenz: Indeed, it looks better when you break at the equal sign :+1:

zonkmachine:
> IMO the speed should be dependent on the range of the spin box

:+1:  It's a bit nervous on the short range values but that was already the case. Fixing this on master is fine by me.

Testing on LinuxMint, Samsung netbook. The knob is a bit too slow for me. I'm testing `value = .8f * _p.y();` and that's perfect on my machine.

JohannesLorenz:
@zonkmachine Thanks for testing. We could change the factor, but now that it works on all operating sytems, I don't dare to touch this code :smiley:

JohannesLorenz:
Tested on all OS. Functional code review left out for now. I'll leave it still open if anyone wants to comment, but we should avoid to change the code.

Ready to merge, this one is not blocking 1.2.

Spekular:
I want to preface this by saying that I'm not opposed to this being merged into 1.2.

With that out of the way: I still consider the new behavior a regression. When testing I find that the mouse moving around is distracting and creates a bad "feeling", for lack of a better explanation. I also just noticed this comment from Tresf earlier in the convo:

> Every other app behaves this way

Which I want to point out is false. For one thing, games have been "hijacking" the mouse for ages (think every FPS ever), and no one takes issue with this as some sort of terrible violation of their mouse input nor a security issue. In more closely related examples:

- Blender locks + hides the mouse when adjusting numeric input fields with the mouse
- Mixxx locks + hides the mouse when adjusting knobs
- Serato locks + hides the mouse when adjusting BPM of a track with the mouse

zonkmachine:
> I just commented the line
> 
> ```
> QApplication::setOverrideCursor( Qt::BlankCursor );
> ```
> 
> out in the mousePressEvent and it looks as I expected (on Linux): The mouse glides with the cursor ...

@JohannesLorenz 
Can we have that line back? It doesn't look like it was part of the issue in the first place.

PhysSong:
> Can we have that line back?

We decided to remove it because it may look odd if the cursor moves while it's hidden, especially when the 'hidden' cursor hits the screen boundary.
We're not locking the cursor anymore because of these two reasons:
- We can't check if `QCursor::setPos()` works reliably
- There were some inconsistencies from (not) using `QCursor::setPos()`

JohannesLorenz:
I couldn't summarize it better than PhysSong did.

Also, this issue is often cited as Mac-only, but we had at least 2 Linux users where `setPos` also did not work (IIRC Fedora was affected, for example). This makes it very difficult to check if `QCursor::setPos` works. I don't know how Blender/... solve that problem. If you want to dive into their code, maybe you'll find out (or end up writing an issue report for their code :smiley: ).

zonkmachine:
OK. As I suggested on Discord, platforms that do not see this issue could still have the fix wrapped in a simple if/endif statement. Windows for instance.

tresf:
> OK. As I suggested on Discord, platforms that do not see this issue could still have the fix wrapped in a simple if/endif statement. Windows for instance.

Sure but that's only Windows (unless they eventually modify this mouse hijacking behavior too).  Linux has this issue with some desktops.  Mac has this issue with some desktops.  @JohannesLorenz tried really hard to make this conditionally grandfathered to the old behavior when available but detecting the presence of this ability to take over the cursor didn't work universally.

JohannesLorenz:
> Windows (unless they eventually modify this mouse hijacking behavior too)

That was also my thought when not using `#ifdef`s. I wonder if it's a trend of operating systems to forbid setting the cursor positions.

But even if Windows will not change their behavior, we had issues with things like how often mouse events happen (which leaded us to make the transition function linear, IIRC). Maybe they change the event rate in the future, and then it's unclear if the transition function will still match the old `setPos` behavior.

I somehow think if we want this really "stable" on Windows, we better don't touch this code anymore. :smiley:

PR #5362, opened 2020-01-02
fix locking of the useless_lock in LocklessRingBuffer.h

he29-net:
Cherry-picked bugfix from #5328.

### DIFF
--- a/include/LocklessRingBuffer.h
+++ b/include/LocklessRingBuffer.h
@@ -122,6 +122,7 @@ public:
 	void waitForData()
 	{
 		QMutex useless_lock;
+		useless_lock.lock();
 		m_notifier->wait(&useless_lock);
 		useless_lock.unlock();
 	}


JohannesLorenz:
Looks good, and is confirming to docs.

Code and style review finished. No testing done yet.

JohannesLorenz:
IIRC you say in the #5328 that it had only been crashing on Windows? If yes, then we need a Windows tester. As it's such a simple change though, for me it's just OK without testing, what do you think?

he29-net:
Yes, it crashed only on Windows (that's why we missed it in the spectrum analyzer update review). I tested the Vectorscope build on Windows and this commit fixed the issue. PhysSong also mentioned it fixed "a similar crash" for him, so that should be hopefully a good enough indication that the fix works as intended.

JohannesLorenz:
Thanks for the fix!

PR #5365, opened 2020-01-05
Fix stuck notes with Helm VSTi

### LINKED ISSUE #3640: Locked Note with VeSTige While Changing Tempo

DomClark:
Fixes #3640.

As mentioned in that issue, resizing notes from MIDI input during tempo changes causes overflows, which result in incorrect MIDI messages sent to the plugin. This is avoided by not resizing these notes.

### DIFF
--- a/src/core/NotePlayHandle.cpp
+++ b/src/core/NotePlayHandle.cpp
@@ -540,6 +540,15 @@ void NotePlayHandle::processMidiTime( const MidiTime& time )
 
 void NotePlayHandle::resize( const bpm_t _new_tempo )
 {
+	if (origin() == OriginMidiInput ||
+		(origin() == OriginNoteStacking && m_parent->origin() == OriginMidiInput))
+	{
+		// Don't resize notes from MIDI input - they should continue to play
+		// until the key is released, and their large duration can cause
+		// overflows in this method.
+		return;
+	}
+
 	double completed = m_totalFramesPlayed / (double) m_frames;
 	double new_frames = m_origFrames * m_origTempo / (double) _new_tempo;
 	m_frames = (f_cnt_t)new_frames;


JohannesLorenz:
Initial tests looked/sounded good. However, when using arpeggio with long notes (e.g. 1/1), changing tempo did reset the arpeggio. Is this intended?

DomClark:
> However, when using arpeggio with long notes (e.g. 1/1), changing tempo did reset the arpeggio. Is this intended?

Is this a regression in this PR? It's quite possibly unintended, but may be out of scope here. The arpeggiator is rather buggy anyway.

JohannesLorenz:
> Is this a regression in this PR? It's quite possibly unintended, but may be out of scope here. The arpeggiator is rather buggy anyway.

Totally agree. And it's no regression.

I just wondered that the Arpeggiator case is not in the code, but somehow the original bug does not occur in Arpeggio (for whatever strange reason...). Was there a reason why you did not catch that case? Would it hurt to catch it?

DomClark:
The root cause of the issue is that notes from MIDI/keyboard input are implemented as normal notes but with a length of `std::numeric_limits<f_cnt_t>::max() / 2`. This causes `m_origFrames * m_origTempo` to overflow in `NotePlayHandle::resize`. Stacked notes inherit their length from their parent, so suffer the same problem, but arpeggio notes are always a sane length (a maximum of two seconds), so don't need special handling.

JohannesLorenz:
OK, makes sense.

All reviews passed, I'll merge this.

PR #5387, opened 2020-01-28
DesignerCodeBrowser

tecan:
Qt designer IDE for developers to reach code faster / easier.


tresf:
@tecan Qt Creator should have cmake support built-in meaning this file is redundant with our existing build system/definitions.

That combined with the fact that this `.pro` file has info specific to only your computer mean we won't be accepting this PR, but we do kindly appreciate the gesture and we hope it does not discourage future contributions.

tecan:
its not part of the build system, its for IDE features. also it can be fixed up to compile eventually.

tresf:
The list of cpp and headers will be obsoleted as soon as a those files change or new ones are added.  Since these files are already defined in CMake, I'm a bit confused why they're in this file to begin with and it's not a list we're interested in maintaining twice.  I'll give someone with more experience with QtCreator a chance to chime in before closing, but I don't see much value in its current state.  Furthermore, in regards to "features" being toggled on with this file, I've read it twice and I do not see evidence of it.

tecan:
i had picked the name DesignerCodeBrowser to lessen confusion for its purpose. its to allow designers to use IDE. it can be set to use the makefile with the IDE. keeping up with new files will be no issue if devs are actually using the file.

tresf:
Thanks for the explanation.  Since the aforementioned issues aren't addressed, I'm closing this for now.  Again, please do not take this as discouragement, we appreciate all PRs even if they aren't accepted into base.

PR #5401, opened 2020-02-23
Bias dropped clip position backwards

### LINKED ISSUE #5393: Loop drops onto next bar over (macOS Mojave 10.14, LMMS v1.2.1.581)
2/5 D:
### Bug Summary
- copying and pasting a loop is offset? hovering over the bar you want it to be and dropping it drops it on the next bar

#### Steps to reproduce
Open a project, make a loop, and try to copy it to the next bar.

#### Expected behavior
Loop drops onto the bar.
#### Actual behavior
Loop drops onto the next bar.

#### LMMS version used
1.2.1.581

Spekular:
This matches the drop behavior from before #4973, as users expect clips to start on the bar they hover over, not on the bar whose start is closer to the cursor. Scales with snap size as well, so it drops onto the half-bar hovered over rather than the half-bar whose start is closer to the cursor, etc.
Closes #5393

### DIFF
--- a/src/core/Track.cpp
+++ b/src/core/Track.cpp
@@ -1620,9 +1620,13 @@ bool TrackContentWidget::pasteSelection( MidiTime tcoPos, QDropEvent * de )
 	// TODO -- Need to draw the hovericon either way, or ghost the TCOs
 	// onto their final position.
 
+	float snapSize = gui->songEditor()->m_editor->getSnapSize();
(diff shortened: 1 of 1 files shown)


PhysSong:
The code looks good. Please feel free to merge it if you think we should revert to the old snapping behavior(no objections from me, FYI).

Spekular:
Thanks for the review! :)

PR #5407, opened 2020-02-26
Decrease Bitcrusher quantization step size

DouglasDGI:
The LMMS Bitcrusher's bit depth reduction step size is at 1.  However, there's no reason for it to be locked to only integers.  Simply decreasing the step size will allow finer adjusting and smoother automation of this value.  This is particularly useful at very low values.

Q: Backwards compatibility?
A: This is technically a breaking change, but it would only make a difference if the user was automating this knob using one of the sloped interpolation modes, and would only be noticeable with automation like that at low values (which they likely wouldn't do anyway because of the small step size).  And the change would likely be closer to what they were trying to do in the first place anyway, in the small chance that they do encounter a noticeable difference.

(I know the commit history is a mess, that won't be a problem with Squash + Merge)


JohannesLorenz:
I'll look at this the next days, especially testing and backwards comp.

JohannesLorenz:
Tested and works. Style and code OK.

Per backwards compatibility, I could imagine that a user wants to use this knob in steps, but probably not in 256 small steps. In the worst case, such a user needs to draw the relevant parts of the interpolated curve by hand, which is also doable. But I think this is just a problem in theory.

I'm OK with the PR. If you don't want to do more commits, I'll squash and merge.

DouglasDGI:
I'll go ahead and merge it then, nothing else needs to be done.  It's just a single number change.

Issue #5389, opened 2020-02-08
SDL2 is not found using mingw


tresf:
Installing `sdl2-mingw-w64` [from @tobydox's `mingw-w64` PPA](https://launchpad.net/~tobydox/+archive/ubuntu/mingw-w64), SDL2 isn't picked up.  SDL1 works just fine.

OS: Ubuntu 18.04

Using the [Ubuntu 18.04 "Bionic" dependencies for `mingw`](https://github.com/LMMS/lmms/wiki/dependencies-windows#ubuntu-cross-compile), replace `sdl-mingw-w64` with `sdl2-mingw-w64` (uninstall `sdl-mingw-w64` if needed).

```bash
../cmake/build_win64.sh
```

```diff
  Supported audio interfaces
  --------------------------
  * ALSA                        : <not supported on this platform>
  * JACK                        : <not supported on this platform>
  * OSS                         : <not found or not supported on this platform>
  * Sndio                       : 
  * PortAudio                   : not found, please install portaudio19-dev (or similar, version >= 1.9) ;if you require PortAudio support
  * libsoundio                  : <disabled in this release>
  * PulseAudio                  : <not supported on this platform>
+ * SDL                         : not found, please install libsdl2-dev (or similar) ;if you require SDL support

PhysSong:
Related: https://github.com/Kitware/CMake/commit/c5c217c6b53b035303747a2971b1b80be5391566
This diff seems to fix it:
```diff
diff --git a/cmake/modules/FindSDL2.cmake b/cmake/modules/FindSDL2.cmake
index 89aea7a8a..7b9e5b149 100644
--- a/cmake/modules/FindSDL2.cmake
+++ b/cmake/modules/FindSDL2.cmake
@@ -82,7 +82,7 @@ SET(SDL2_SEARCH_PATHS
 FIND_PATH(SDL2_INCLUDE_DIR SDL.h
 	HINTS
 	$ENV{SDL2DIR}
-	PATH_SUFFIXES include/SDL2 include
+	PATH_SUFFIXES SDL2 include/SDL2 include
 	PATHS ${SDL2_SEARCH_PATHS}
 )
 ```

tresf:
@PhysSong added per 0196f31

PR #5417, opened 2020-03-10
Used pkg-config to find SDL2 fixes issue #5389

DragonEagle:


PhysSong:
Thanks, but it breaks SDL2 detection for MSVC builds.
I found a proper fix that doesn't rely on `pkg-config`, so I'm closing this.
Please note that I'm *not* discouraging your contributions. Thanks anyway!

PR #5427, opened 2020-03-23
Keyboard shortcuts to preview/add sounds from sidebar

Spekular:
Now ready for testing and review

The file browser now behaves as follows:
- Pressing arrow keys, space, or enter/return will cancel any running previews
- Pressing up/down moves to the previous/next item
  - If the new item is a sample, it will automatically preview
  - If the key is held down and auto-repeats, previews are suppressed
- Holding space previews the current item. The preview ends when space is released.
- Enter/return sends the current item to the song editor as an instrument
  - Samples can be sent to a new sample track instead by holding shift
  - Holding control sends to the BB editor instead
  - Samples can't be sent as sample tracks to the BB editor (wontfix), so ctrl+shift does nothing
- RMB context menu gets fake headers, however legibility isn't ideal

Preset Menu
Sample Menu

~Todo:~
- ~Sustained presets shouldn't play indefinitely. Only play presets while left arrow held?~ Done
- ~Shift + right arrow to add a sample to a new (empty) sample track in song editor~ Done
- ~Ctrl + shift + right arrow to add a sample to a new (empty) sample track in BB editor~ Not feasible, nor particularly useful.

~Code quality is iffy at best right now, among other things I did some questionable refactoring in the context menu code.~ The commit "DRY up code and ..." vastly improves this

### DIFF
--- a/include/FileBrowser.h
+++ b/include/FileBrowser.h
@@ -99,6 +99,10 @@ public:
 	//! that are expanded in the tree.
 	QList<QString> expandedDirs( QTreeWidgetItem * item = nullptr ) const;
 
+	void tryAddSEInstrumentTrack(FileItem* file);
+	void tryAddBBInstrumentTrack(FileItem* file);
+	void previewFileItem(FileItem* file);
+
 
 protected:
 	void contextMenuEvent( QContextMenuEvent * e ) override;
@@ -108,6 +112,8 @@ protected:
 
 
 private:
+	void keyPressEvent( QKeyEvent * ke ) override;
+
 	void handleFile( FileItem * fi, InstrumentTrack * it );
 	void openInNewInstrumentTrack( TrackContainer* tc, FileItem* item );
 
@@ -118,8 +124,8 @@ private:
 	PlayHandle* m_previewPlayHandle;
 	QMutex m_pphMutex;
 
-	void populateSampleMenu(QMenu& contextMenu, FileItem* item);
-	void populatePluginMenu(QMenu& contextMenu, FileItem* item);
+	QList<QAction*> getContextActionsSE(FileItem* item);
+	QList<QAction*> getContextActionsBBE(FileItem* item);
 
 
 private slots:
@@ -234,6 +240,11 @@ public:
 	inline FileHandling handling( void ) const
 	{
 		return( m_handling );
+	}
+
+	inline bool isTrack( void ) const
+	{
+		return m_handling == LoadAsPreset || m_handling == LoadByPlugin;
 	}
 
 	QString extension( void );
--- a/include/FileBrowser.h
+++ b/include/FileBrowser.h
@@ -135,7 +135,7 @@ private:
 private slots:
 	void activateListItem( QTreeWidgetItem * item, int column );
 	void openInNewInstrumentTrack( FileItem* item, bool songEditor );
-	bool openInNewSampleTrack( FileItem* item, bool songEditor );
+	bool openInNewSampleTrack( FileItem* item );
 	void sendToActiveInstrumentTrack( FileItem* item );
 	void updateDirectory( QTreeWidgetItem * item );
 	void openContainingFolder( FileItem* item );
--- a/include/FileBrowser.h
+++ b/include/FileBrowser.h
@@ -115,6 +115,7 @@ private:
 
 	void keyPressEvent( QKeyEvent * ke ) override;
 	void keyReleaseEvent( QKeyEvent * ke ) override;
+	void hideEvent( QHideEvent * he ) override;
 
 	void handleFile( FileItem * fi, InstrumentTrack * it );
 	void openInNewInstrumentTrack( TrackContainer* tc, FileItem* item );
@@ -123,9 +124,9 @@ private:
 	bool m_mousePressed;
 	QPoint m_pressPos;
 
+	//! This should only be accessed or modified when m_pphMutex is held
 	PlayHandle* m_previewPlayHandle;
 	QMutex m_pphMutex;
-	FileItem* m_previewFileItem;
 
 	QList<QAction*> getContextActions(FileItem* item, bool songEditor);
 
--- a/include/FileBrowser.h
+++ b/include/FileBrowser.h
@@ -113,7 +113,7 @@ protected:
 private:
 	//! Start a preview of a file item
 	void previewFileItem(FileItem* file);
-	//! If a preview is playing, stop it. Returns false if nothing was playing
+	//! If a preview is playing, stop it.
 	void stopPreview();
 
 	void handleFile( FileItem * fi, InstrumentTrack * it );
--- a/include/FileBrowser.h
+++ b/include/FileBrowser.h
@@ -105,6 +105,9 @@ protected:
 	void mousePressEvent( QMouseEvent * me ) override;
 	void mouseMoveEvent( QMouseEvent * me ) override;
 	void mouseReleaseEvent( QMouseEvent * me ) override;
+	void keyPressEvent( QKeyEvent * ke ) override;
+	void keyReleaseEvent( QKeyEvent * ke ) override;
+	void hideEvent( QHideEvent * he ) override;
 
 
 private:
@@ -112,10 +115,6 @@ private:
 	void previewFileItem(FileItem* file);
 	//! If a preview is playing, stop it. Returns false if nothing was playing
 	void stopPreview();
-
-	void keyPressEvent( QKeyEvent * ke ) override;
-	void keyReleaseEvent( QKeyEvent * ke ) override;
-	void hideEvent( QHideEvent * he ) override;
 
 	void handleFile( FileItem * fi, InstrumentTrack * it );
 	void openInNewInstrumentTrack( TrackContainer* tc, FileItem* item );
--- a/include/FileBrowser.h
+++ b/include/FileBrowser.h
@@ -101,7 +101,6 @@ public:
 
 	void tryAddSEInstrumentTrack(FileItem* file);
 	void tryAddBBInstrumentTrack(FileItem* file);
-	void previewFileItem(FileItem* file);
 
 
 protected:
@@ -112,7 +111,13 @@ protected:
 
 
 private:
+	//Preview a file
+	void previewFileItem(FileItem* file);
+	//If a preview is playing, stop it and return true. Otherwise return false
+	bool stopPreview();
+	
 	void keyPressEvent( QKeyEvent * ke ) override;
+	void keyReleaseEvent( QKeyEvent * ke ) override;
 
 	void handleFile( FileItem * fi, InstrumentTrack * it );
 	void openInNewInstrumentTrack( TrackContainer* tc, FileItem* item );
@@ -124,14 +129,12 @@ private:
 	PlayHandle* m_previewPlayHandle;
 	QMutex m_pphMutex;
 
-	QList<QAction*> getContextActionsSE(FileItem* item);
-	QList<QAction*> getContextActionsBBE(FileItem* item);
+	QList<QAction*> getContextActions(FileItem* item, bool songEditor);
 
 
 private slots:
 	void activateListItem( QTreeWidgetItem * item, int column );
-	void openInNewInstrumentTrackBBE( FileItem* item );
-	void openInNewInstrumentTrackSE( FileItem* item );
+	void openInNewInstrumentTrack( FileItem* item, bool songEditor );
 	void openInNewSampleTrack( FileItem* item );
 	void sendToActiveInstrumentTrack( FileItem* item );
 	void updateDirectory( QTreeWidgetItem * item );
--- a/include/FileBrowser.h
+++ b/include/FileBrowser.h
@@ -63,9 +63,9 @@ public:
 
 private slots:
 	void reloadTree( void );
-	void expandItems( QTreeWidgetItem * item=NULL, QList<QString> expandedDirs = QList<QString>() );
+	void expandItems( QTreeWidgetItem * item=nullptr, QList<QString> expandedDirs = QList<QString>() );
 	// call with item=NULL to filter the entire tree
-	bool filterItems( const QString & filter, QTreeWidgetItem * item=NULL );
+	bool filterItems( const QString & filter, QTreeWidgetItem * item=nullptr );
 	void giveFocusToFilter();
 
 private:
@@ -112,7 +112,7 @@ private:
 	void previewFileItem(FileItem* file);
 	//! If a preview is playing, stop it. Returns false if nothing was playing
 	bool stopPreview();
-	
+
 	void keyPressEvent( QKeyEvent * ke ) override;
 	void keyReleaseEvent( QKeyEvent * ke ) override;
 
--- a/include/FileBrowser.h
+++ b/include/FileBrowser.h
@@ -109,7 +109,7 @@ protected:
 
 private:
 	void handleFile( FileItem * fi, InstrumentTrack * it );
-	void openInNewInstrumentTrack( TrackContainer* tc );
+	void openInNewInstrumentTrack( TrackContainer* tc, FileItem* item );
 
 
 	bool m_mousePressed;
@@ -118,14 +118,16 @@ private:
 	PlayHandle* m_previewPlayHandle;
 	QMutex m_pphMutex;
 
-	FileItem * m_contextMenuItem;
+	void populateSampleMenu(QMenu& contextMenu, FileItem* item);
+	void populatePluginMenu(QMenu& contextMenu, FileItem* item);
 
 
 private slots:
 	void activateListItem( QTreeWidgetItem * item, int column );
-	void openInNewInstrumentTrackBBE( void );
-	void openInNewInstrumentTrackSE( void );
-	void sendToActiveInstrumentTrack( void );
+	void openInNewInstrumentTrackBBE( FileItem* item );
+	void openInNewInstrumentTrackSE( FileItem* item );
+	void openInNewSampleTrack( FileItem* item );
+	void sendToActiveInstrumentTrack( FileItem* item );
 	void updateDirectory( QTreeWidgetItem * item );
 	void openContainingFolder();
 
--- a/include/FileBrowser.h
+++ b/include/FileBrowser.h
@@ -125,6 +125,7 @@ private:
 
 	PlayHandle* m_previewPlayHandle;
 	QMutex m_pphMutex;
+	FileItem* m_previewFileItem;
 
 	QList<QAction*> getContextActions(FileItem* item, bool songEditor);
 
--- a/include/FileBrowser.h
+++ b/include/FileBrowser.h
@@ -111,7 +111,7 @@ private:
 	//! Start a preview of a file item
 	void previewFileItem(FileItem* file);
 	//! If a preview is playing, stop it. Returns false if nothing was playing
-	bool stopPreview();
+	void stopPreview();
 
 	void keyPressEvent( QKeyEvent * ke ) override;
 	void keyReleaseEvent( QKeyEvent * ke ) override;
--- a/include/FileBrowser.h
+++ b/include/FileBrowser.h
@@ -99,9 +99,6 @@ public:
 	//! that are expanded in the tree.
 	QList<QString> expandedDirs( QTreeWidgetItem * item = nullptr ) const;
 
-	void tryAddSEInstrumentTrack(FileItem* file);
-	void tryAddBBInstrumentTrack(FileItem* file);
-
 
 protected:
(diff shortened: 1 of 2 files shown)


src/gui/FileBrowser.cpp
```diff
@@ -405,62 +510,71 @@ void FileBrowserTreeWidget::mousePressEvent(QMouseEvent * me )
...
+
+
+
+void FileBrowserTreeWidget::previewFileItem(FileItem* file)
+{
+	m_pphMutex.lock();
+	if( m_previewPlayHandle != NULL )
+	{
+		Engine::mixer()->removePlayHandle(
+						m_previewPlayHandle );
```
IanCaio, line 526: Wouldn't it be better to just have `Engine::mixer()->removePlayHandle( m_previewPlayHandle );` in a single line? Since the indentation required for readability makes the width of the second line close to the width of a single line statement?

src/gui/FileBrowser.cpp
```diff
@@ -405,62 +510,71 @@ void FileBrowserTreeWidget::mousePressEvent(QMouseEvent * me )
...
+	if( file->type() == FileItem::SampleFile )
+	{
+		TextFloat * tf = TextFloat::displayMessage(
+				tr( "Loading sample" ),
+				tr( "Please wait, loading sample for "
+							"preview..." ),
+				embed::getIconPixmap( "sample_file",
+							24, 24 ), 0 );
+		qApp->processEvents(
+				QEventLoop::ExcludeUserInputEvents );
```
IanCaio, line 541: Same as the last comment

src/gui/FileBrowser.cpp
```diff
@@ -405,62 +510,71 @@ void FileBrowserTreeWidget::mousePressEvent(QMouseEvent * me )
...
+		SamplePlayHandle * s = new SamplePlayHandle( file->fullName() );
+		s->setDoneMayReturnTrue( false );
+		m_previewPlayHandle = s;
+		delete tf;
+	}
+	else if(
+		( file->extension ()== "xiz" || file->extension() == "sf2" ||
+		  file->extension() == "sf3" || file->extension() == "gig" ||
+		  file ->extension() == "pat" ) &&
+		! pluginFactory->pluginSupportingExtension(file->extension()).info.isNull() )
```
IanCaio, line 551: I think that can be shortened to `! pluginFactory->pluginSupportingExtension(file->extension()).isNull() )` since the `PluginInfoAndKey` struct has a `isNull()` function returning `info.isNull()`.

src/gui/FileBrowser.cpp
Spekular: Resolving these since formatting has changed so much since this review. Most of them were the result of copying existing code.

Spekular:
IMHO this should close #5671. The proposal there isn't an improvement as far as I'm concerned, and with this merged it makes even less sense.

Veratil (commented):
The `nullptr` comment scopes across the whole review.

src/gui/FileBrowser.cpp
```diff
@@ -362,59 +364,145 @@ QList<QString> FileBrowserTreeWidget::expandedDirs( QTreeWidgetItem * item ) con
...
+	QTreeWidget::keyPressEvent(ke);
+	// Then, ignore all autorepeats (they would spam new tracks or previews)
+	if (ke->isAutoRepeat()){ return; }
+	// We should stop any running previews before we do anything new
+	else if (vertical || horizontal || preview || insert){ stopPreview(); }
+
+	// Try to get the currently selected item as a FileItem
+	FileItem * file = dynamic_cast<FileItem *>(currentItem());
+	// If it's null (folder, separator, etc.), there's nothing left for us to do
+	if (file == NULL){ return; }
```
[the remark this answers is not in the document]
Spekular: I ended up fixing all the uses of NULL for pointers in the file, it felt weird to only fix some.

src/gui/FileBrowser.cpp
```diff
@@ -362,59 +364,145 @@ QList<QString> FileBrowserTreeWidget::expandedDirs( QTreeWidgetItem * item ) con
...
+void FileBrowserTreeWidget::keyPressEvent(QKeyEvent * ke )
 {
-	FileItem * f = dynamic_cast<FileItem *>(itemAt(e->pos()));
-	if (f == nullptr)
+	// Shorter names for some commonly used properties of the event
+	const auto key = ke->key();
+	const bool vertical = key == Qt::Key_Up || key == Qt::Key_Down;
+	const bool horizontal = key == Qt::Key_Left || key == Qt::Key_Right;
+	const bool preview = key == Qt::Key_Space;
+	const bool insert = key == Qt::Key_Enter || key == Qt::Key_Return;
```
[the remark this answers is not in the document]
Spekular: Fair point, I've cleaned this up now!

Spekular:
FYI, CircleCI mac failure looks unrelated (it fails to fetch homebrew).

Spekular:
The latest commit is a bit of a shot in the dark, the overhead from locking/unlocking more often and changing `m_previewFileItem` might cancel out the benefit of not hogging the lock as long. That being said, as far as I can tell it shouldn't introduce threading issues.
- Previously: The code that I extracted into `previewFileItem` locked the preview mutex immediately, and only unlocked before returning. The code that I extracted into `stopPreview` was also completely wrapped in lock/unlock. Thus, whatever function entered the lock most recently would be the one that took effect.
- Now: `previewFileItem` and `stopPreview` immediately lock and then set `m_previewFileItem` before unlocking. Whatever function entered the lock most recently decides the state of `m_previewFileItem`, which in turn decides what action will take effect.

If this commit works, I'll add comments clarifying that access to `m_previewFileItem` and `m_previewPlayHandle` should be protected by `m_pphMutex` (which should maybe be renamed to `m_previewMutex`?). If it fails or someone has objections, I'll revert the changes affecting when locks are aquired/released, but leave the use of QMutexLocker.

Spekular:
Testing doesn't reveal any apparent difference, which isn't surprising. In hindsight it's pretty obvious that making the preview lock in FileBrowser available slightly faster wouldn't help, but it was an easy thing to check.

src/gui/FileBrowser.cpp
```diff
@@ -362,59 +364,155 @@ QList<QString> FileBrowserTreeWidget::expandedDirs( QTreeWidgetItem * item ) con
...
+	const bool fileIsSample = file->type() == FileItem::SampleFile;
+
+	QString destination = fileIsSample ? "AFP Instance" : "Instrument Track";
+	QString shortcutMod = songEditor ? "" : "Ctrl +";
+
+	QAction* toInstrument = new QAction(
+		tr("Send to new %1 (%2 Enter)")
+		.arg(destination, shortcutMod),
+		nullptr
+	);
```
DomClark, line 486: A couple of things here:
* I think "AFP" would be clearer written out in full as "AudioFileProcessor" - some users may not be familiar with the acronym.
* The substitutions here may not work in every language (e.g. "new" may need to inflect differently for each destination; "to" may be an inflection of the noun itself, not a separate word; etc.). I think the the destination should be included directly in the string to translate, rather than substituted in later.
* "Ctrl" is "⌘" on Mac - use the `UI_CTRL_KEY` macro instead. (Also, ideally "Ctrl" should be translated too, but we don't do that anywhere else, so feel free not to.)

src/gui/FileBrowser.cpp
```diff
@@ -556,36 +671,30 @@ void FileBrowserTreeWidget::mouseReleaseEvent(QMouseEvent * me )
...
-				s->setDoneMayReturnTrue( true );
-				m_previewPlayHandle = NULL;
-				m_pphMutex.unlock();
-				return;
+				s->setDoneMayReturnTrue(true);
 			}
 		}
-		Engine::mixer()->removePlayHandle( m_previewPlayHandle );
-		m_previewPlayHandle = NULL;
+		else { stopPreview(); }
```
DomClark, line 689: This no longer stops the preview of samples, even if there are more than three seconds remaining.

DomClark (commented):
Regarding `m_pphMutex`, couldn't it be removed altogether? From what I can tell, it's only ever locked from the UI thread, so there's nothing for it to protect against.

Spekular:
> Regarding `m_pphMutex`, couldn't it be removed altogether? From what I can tell, it's only ever locked from the UI thread, so there's nothing for it to protect against.

My understanding is that it prevents several calls created via UI input from interfering with each other. Is this not necessary?

For example, if you very quickly tap the spacebar, couldn't this happen without locks?
- previewFileItem gets as far as checking `m_previewPlayHandle != nullptr`
- stopPreview sets the play handle to `nullptr`
- We end up calling call `Engine::getMixer()->addPlayHandle()` with a `nullptr` argument

Also, as far as I can tell C++ makes no guarantees that a variable is in a valid state when read in unsynchronized code. Isn't this alone a good reason to avoid concurrent access?

Spekular:
Ah wait, you mean that all the modifications to the play handle originate from UI events and thus run on the UI thread, so there's no concurrency happening. That makes sense, but to me it raises the issue that perhaps some of this work should actually happen outside the UI thread.

DomClark:
> ... it raises the issue that perhaps some of this work should actually happen outside the UI thread.

True - it would definitely be nice to load samples (and presets too, I guess) asynchronously. Let's leave it in for now.

src/gui/FileBrowser.cpp
```diff
@@ -663,53 +772,66 @@ void FileBrowserTreeWidget::activateListItem(QTreeWidgetItem * item,
...
 
-void FileBrowserTreeWidget::openInNewInstrumentTrackSE( void )
+bool FileBrowserTreeWidget::openInNewSampleTrack(FileItem* item)
 {
-	openInNewInstrumentTrack( Engine::getSong() );
+	// Can't add non-samples to a sample track
+	if (item->type() != FileItem::SampleFile){ return false; }
+
+	// Create a new sample track for this sample
+	SampleTrack* sampleTrack = dynamic_cast<SampleTrack*>(
```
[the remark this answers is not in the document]
Spekular: Done in latest commit.

include/FileBrowser.h
```diff
@@ -108,26 +108,36 @@ class FileBrowserTreeWidget : public QTreeWidget
 
 
 private:
+	//! Start a preview of a file item
+	void previewFileItem(FileItem* file);
+	//! If a preview is playing, stop it. Returns false if nothing was playing
+	bool stopPreview();
```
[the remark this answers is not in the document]
Spekular: Done.

src/gui/FileBrowser.cpp
Spekular: Should be satisfactory as of latest commit, unresolve and let me know if I missed anything.

src/gui/FileBrowser.cpp
Spekular: Fixed in latest commit.

Spekular:
Thanks for the review, @DomClark! Recent commits should have addressed those issues.

I also tried replacing the preview mutex with a QThreadPool, but `QRunnable::create(... lambda expression...)` is only in Qt 5.15 and later and we're quite a bit behind. Anyway, the idea was as follows:
- Create a thread pool with a maximum of one thread. As long as all access to `m_previewPlayHandle` happens by starting `QRunnable`s on the thread pool, we achieve thread confinement for the play handle and also get out of the event thread's way.
- `previewFileItem` and `stopPreview` would initially call `clear()` on the thread pool, clearing any queued-but-not-started tasks (since they're outdated). Then they would call `start(QRunnable::create(... do all the work in a lambda...))`. This should let the thread pool take care of their work ASAP unless another, newer, event comes along first.
- `mouseReleaseEvent` would just queue itself. It wouldn't call `clear()`, since it may or may not need to stop the preview.

include/FileBrowser.h
```diff
@@ -108,26 +108,36 @@ class FileBrowserTreeWidget : public QTreeWidget
...
 
 private:
+	//! Start a preview of a file item
+	void previewFileItem(FileItem* file);
+	//! If a preview is playing, stop it. Returns false if nothing was playing
+	void stopPreview();
+
+	void keyPressEvent( QKeyEvent * ke ) override;
+	void keyReleaseEvent( QKeyEvent * ke ) override;
+	void hideEvent( QHideEvent * he ) override;
```
Veratil, line 118: What's our policy on overrides class visibility? These are under `private` whereas Qt has them under `protected`. I don't think we put any event handlers under `private` in any class.

src/gui/FileBrowser.cpp
```diff
@@ -338,6 +337,9 @@ FileBrowserTreeWidget::FileBrowserTreeWidget(QWidget * parent ) :
 
 }
 
+
+
+
```
Veratil, line 342: I recall somewhere about having 4 newlines between functions, but I don't know if we care to keep doing that? I say because this hunk is unrelated to PR, and more a formatting update.

src/gui/FileBrowser.cpp
```diff
@@ -749,9 +871,9 @@ void FileBrowserTreeWidget::updateDirectory(QTreeWidgetItem * item )
 
 
 
-QPixmap * Directory::s_folderPixmap = NULL;
-QPixmap * Directory::s_folderOpenedPixmap = NULL;
-QPixmap * Directory::s_folderLockedPixmap = NULL;
+QPixmap * Directory::s_folderPixmap = nullptr;
+QPixmap * Directory::s_folderOpenedPixmap = nullptr;
+QPixmap * Directory::s_folderLockedPixmap = nullptr;
```
Veratil, line 876: I'd rather see these at the top of the file (kinda standard place to put variable initializers).

src/gui/FileBrowser.cpp
```diff
@@ -928,13 +1050,13 @@ bool Directory::addItems(const QString & path )
...
-QPixmap * FileItem::s_vstPluginFilePixmap = NULL;
-QPixmap * FileItem::s_midiFilePixmap = NULL;
-QPixmap * FileItem::s_unknownFilePixmap = NULL;
+QPixmap * FileItem::s_projectFilePixmap = nullptr;
+QPixmap * FileItem::s_presetFilePixmap = nullptr;
+QPixmap * FileItem::s_sampleFilePixmap = nullptr;
+QPixmap * FileItem::s_soundfontFilePixmap = nullptr;
+QPixmap * FileItem::s_vstPluginFilePixmap = nullptr;
+QPixmap * FileItem::s_midiFilePixmap = nullptr;
+QPixmap * FileItem::s_unknownFilePixmap = nullptr;
```
Veratil, line 1059: Same with these.

Veratil (commented):
The only other thing I see is no whitespace between parenthesis and curly brace in `if (){ .. }`.

src/gui/FileBrowser.cpp
Spekular: I wanted to keep FileBrowserTreeWidget's spacing a bit more consistent since it varied before, but I can undo this if you'd like.

src/gui/FileBrowser.cpp
Spekular: They're not added in this PR, so I haven't moved them.

src/gui/FileBrowser.cpp
Spekular: Same as the other.

include/FileBrowser.h
DomClark: It depends on the situation. The general rule is to give things as little visibility as possible, but of course there are exceptions. You shouldn't reduce the visibility of `public` functions - these are a part of the public interface of the base class, so if they're not a part of the public interface of your derived class, you're probably using inheritance badly. It's okay to override `protected` methods as `private` if you want to stop further derived classes from calling your overridden implementation.

In this case though, I think they should stay `protected`. The way you get default event handling in Qt is to call the base class's implementation of the handler, so making them `private` makes this class very difficult to derive from. If this is the intention (although I doubt it), they should probably be declared `final` - it doesn't really make sense to allow a derived class to handle events, but force it to handle each event fully itself, and never pass it back up the hierarchy for default handling.

src/gui/FileBrowser.cpp
DomClark: FWIW, I don't like the four newlines. (I'm generally in favour of anything that uses less vertical space.)

include/FileBrowser.h
Spekular: Good point, fixed in latest commit!

Spekular:
Thanks for the review @Veratil! I've fixed the event handler visibility and spacing in if statements, and also fixed some inconsistent spacing between FileBrowserTreeWidget's methods. I can revert the spacing changes if you'd like, but figured if they're going to be there they should be consistent.

src/gui/FileBrowser.cpp
Spekular: I would also prefer a single newline between functions of the same class, but I've generalmy stuck with four in order to have some semblance of consistency.

DomClark: Yeah - I don't think it should be changed here, but it might be a candidate for the reformatting if others agree.

Veratil: I think it would be a thing for the clang-tidy/format PR.

src/gui/FileBrowser.cpp
Veratil: That doesn't mean you can't move them. :)

Spekular: Yeah, but that's an even more extreme formatting change than adding whitespace, so I'd rather not

Spekular: Also, the way it is now they're defined above the class that uses them. If each class had it's own file, they would end up at the top of the file.

Veratil: That's fair!

Spekular:
Any objections to me merging this? There are technically no approving reviews, I suppose, but I've fixed the issues brought up in review and no one has mentioned new ones so far.

src/gui/FileBrowser.cpp
```diff
@@ -556,36 +672,29 @@ void FileBrowserTreeWidget::mouseReleaseEvent(QMouseEvent * me )
...
-				m_previewPlayHandle = NULL;
-				m_pphMutex.unlock();
-				return;
+				s->setDoneMayReturnTrue(true);
 			}
+			else { stopPreview(); }
 		}
-		Engine::mixer()->removePlayHandle( m_previewPlayHandle );
-		m_previewPlayHandle = NULL;
 	}
```
DomClark, line 691: This now stops sample previews with more than three seconds left correctly, but preset previews now go on forever.

src/gui/FileBrowser.cpp
```diff
@@ -663,53 +772,66 @@ void FileBrowserTreeWidget::activateListItem(QTreeWidgetItem * item,
...
+	// Can't add non-samples to a sample track
+	if (item->type() != FileItem::SampleFile) { return false; }
+
+	// Create a new sample track for this sample
+	SampleTrack* sampleTrack = static_cast<SampleTrack*>(
+		Track::create(Track::SampleTrack, Engine::getSong()));
+
+	// Add the sample clip to the track
+	Engine::mixer()->requestChangeInModel();
+	SampleTCO* clip = dynamic_cast<SampleTCO*>(sampleTrack->createTCO(0));
```
DomClark, line 810: This can be a `static_cast` too.

src/gui/FileBrowser.cpp
Spekular: Good catch, that was poor testing on my part! Fixed and tested, works properly again for me.

src/gui/FileBrowser.cpp
Spekular: Fixed in latest commit.

DomClark (approved):

src/gui/FileBrowser.cpp
```diff
@@ -556,36 +672,30 @@ void FileBrowserTreeWidget::mouseReleaseEvent(QMouseEvent * me )
...
-							m_previewPlayHandle );
-			if( s && s->totalFrames() - s->framesDone() <=
-				static_cast<f_cnt_t>( Engine::mixer()->
-						processingSampleRate() * 3 ) )
+			SamplePlayHandle* s = dynamic_cast<SamplePlayHandle*>(m_previewPlayHandle);
+			auto second = static_cast<f_cnt_t>(Engine::mixer()->processingSampleRate());
+			if (s && s->totalFrames() - s->framesDone() <= second * 3)
 			{
-				s->setDoneMayReturnTrue( true );
-				m_previewPlayHandle = NULL;
```
PhysSong, line 573: I think this removed line is the key of #5736. `Mixer` will cleanup the play handle later, and we won't stop previewing in this case.

PR #5429, opened 2020-03-24
Applied PhysSong's Hydrogen import fix to master

DragonEagle:
Applied PhysSong's Hydrogen import fix to master branch.  Build tested under both windows (cross compiled) and Linux.  Built and imported successfully.

### DIFF
--- a/plugins/HydrogenImport/HydrogenImport.cpp
+++ b/plugins/HydrogenImport/HydrogenImport.cpp
@@ -202,13 +202,19 @@ bool HydrogenImport::readSong()
 			else 
 			{
 				unsigned nLayer = 0;
-				QDomNode layerNode = instrumentNode.firstChildElement( "layer" );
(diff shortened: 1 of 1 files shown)


zonkmachine:
@DragonEagle We usually merge from the stable branch to the master and all at once. This is usually done after a release and 1.2.2 is not far away so this fix should probably be merged with the rest.

zonkmachine:
Also, it doesn't look like you used `git cherry-pick` to add the commit to master. The proper way to do this would have been to execute `git cherry-pick 2367a62a51ec833e08d6dd342b65b9b1829391d5` on the master branch. That way @PhysSong's original commit message would have been retained.
If you're not already active on the discord chat I think you should look into it. This is were the developers hang out. https://discord.gg/3sc5su7

I'm closing this now and it will be merged into master in maybe a week from stable-1.2 .

PR #5437, opened 2020-04-04
Include QtGlobal for qAsConst

Reflexe:


Reflexe:
Oh I see that @PhysSong already fixed that. Thanks :D

Issue #5405, opened 2020-02-25
Incorrect note velocity on change


BluuDreamz:
Selecting a group of notes on a piano roll in order to change the velocity of said selected notes to a new value altogether through the dialog box only updates the first selected/highlighted note to the designated value.

#### Step 1:
Group-select a series of notes on a piano roll.

#### Step 2:
Double-click on any one of the velocity nodes within the note velocity editor on the piano roll to prompt the dialog box.

#### Step 3: 
Enter in any value you wish to change the group-selected notes to.

#### Expected behavior
Changing the note velocity of a selected group of notes through the dialog box to x value should set all selected notes to that value.

#### Actual behavior
Despite having them all selected, changing the note velocity of a selected group of notes through the dialog box to x value only sets the first highlighted/selected velocity note value to the designated value.

#### Screenshots
![image](https://user-images.githubusercontent.com/39863151/75263093-816b1180-57bb-11ea-8090-ddcdd51fc8ec.png)

![image](https://user-images.githubusercontent.com/39863151/75263140-95167800-57bb-11ea-93f5-0bb41d60733d.png)

![image](https://user-images.githubusercontent.com/39863151/75263155-9a73c280-57bb-11ea-9cab-8843a85147dd.png)

#### LMMS version used: 1.2.1

The practicality to this function is to allow the user to reset a select group of notes to a default value of their choice.

JohannesLorenz:
@BluuDreamz There's a fixing PR #5438 that needs testing. Would you mind to do some testing for your issue?

zonkmachine:
Fixed in https://github.com/LMMS/lmms/pull/5438

@BluuDreamz Thanks for reporting! The fix will be available in lmms-1.2.2, coming soon-ish...

PR #5438, opened 2020-04-04
Note selection: Editing values works in dialog

zonkmachine:
Allow setting volume/pan for selected notes in pop up dialog.
I think this is safe for stable-1.2 but the PR works on master too.

Fixes https://github.com/LMMS/lmms/issues/5405

### DIFF
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -1772,8 +1772,9 @@ void PianoRoll::mouseDoubleClickEvent(QMouseEvent * me )
 		{
 			const Note * closest = NULL;
 			int closest_dist = 9999999;
-			// if we caught multiple notes, find the closest...
-			if( nv.size() > 1 )
+			// if we caught multiple notes and we're not editing a
+			// selection, find the closest...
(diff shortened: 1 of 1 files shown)


Spekular:
Style and code LGTM.

JohannesLorenz:
What kind of testing is required @zonkmachine ? Blackbox testing (tester has no code knowledge) or whitebox testing (tester looks at the code)?

PhysSong:
In this case, trying the steps in the original issue will be enough. That's the only case this PR affects.

JohannesLorenz:
OK, I contacted the OP of the issue, I think they can test it best.

zonkmachine:
I think just comparing the issue before and after the PR is enough. I've compared editing a pattern in the Piano Roll, with or without the fix side by side and I've seen no problem. Apart from editing all notes via the dialog the behaviour is the same. If no one has time to test it I'm happy to see it merged anyway as it's a one-liner and the code has been approved.

zonkmachine:
I tested it some more but I don't have any more cases to throw at it. I think its good to go.
Merge at will or I'll merge in a couple of days if there are no objections.

PR #5441, opened 2020-04-06
VST preset preview

### LINKED ISSUE #5039: Idea for re-establishing VST-based LMMS-preset 'pre-hear'
As all know, pre-hear from side-bar| myPresets with saved VSTs-presets had to be removed, because it was crash-prone.
I have an idea how to bring it back for user-saved presets with VSTs, and presets where VSTs effects has been inserted.

When a user creates a new preset, either based on VST, or has a VST-effect embedded, and save it as a named preset, LMMS would need to create a render for ~ 3 secd in ogg-forbis /wave. This soundfile should now be associated with that preset!
When the user then, in side-bar,  press mouse on that specific preset, the sound-file is played by LMMS. This would give the user a pre-hear from side-bar of the preset, without any VST-related crashing.

Special case: Old presets.
a method for creating and associating old presets in the myPresets| VST folder, that for natural reasons do not have any sound-file associations, is also needed
This is a one-off, and does not even have to be part of LMMS, but could be a utility that users only uses once.

akimaze:
Hello all.

Recently I started using LMMS, I like it very much. I started creating presets and it turned out that presets created in Vestige are not playing. In addition, when you click any VST preset, zynaddsubfx presets also stops playing. So I created a small patch that activates VST presets. To minimize the waiting time between preset switching, I changed the code to not delete the vestige instrument and plugin dll (when next preset uses the same plugin) every time preset is changed. Plugin UI is always hidden.

This is my first PR for LMMS so maybe I didn't understand something in the sources (spent only few hours). Maybe this patch will be helpful for someone who also wants to hear their presets. Tested on Linux.

Regards.

### DIFF
--- a/include/InstrumentTrack.h
+++ b/include/InstrumentTrack.h
@@ -214,6 +214,10 @@ public:
 
 	void setPreviewMode( const bool );
 
+	bool isPreviewMode() const
+	{
+		return m_previewMode;
+	}
 
 signals:
 	void instrumentChanged();
--- a/include/InstrumentTrack.h
+++ b/include/InstrumentTrack.h
@@ -230,6 +230,8 @@ protected:
 		return "instrumenttrack";
 	}
 
+	QString getInstrumentName(const QDomElement & thisElement) const;
+
 
 protected slots:
 	void updateBaseNote();
--- a/include/InstrumentTrack.h
+++ b/include/InstrumentTrack.h
@@ -234,7 +234,8 @@ protected:
 		return "instrumenttrack";
 	}
 
-	QString getInstrumentName(const QDomElement & thisElement) const;
+	// get the name of the instrument in the saved data
+	QString getSavedInstrumentName(const QDomElement & thisElement) const;
 
 
 protected slots:
--- a/src/tracks/InstrumentTrack.cpp
+++ b/src/tracks/InstrumentTrack.cpp
@@ -771,7 +771,8 @@ void InstrumentTrack::saveTrackSpecificSettings( QDomDocument& doc, QDomElement 
 
 void InstrumentTrack::loadTrackSpecificSettings( const QDomElement & thisElement )
 {
-	silenceAllNotes( true );
+	silenceAllNotes(!(m_previewMode && m_instrument && m_instrument->nodeName() == "vestige"
+					&& getInstrumentName(thisElement) == "vestige"));
 
 	lock();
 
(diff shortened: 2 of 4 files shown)


PhysSong:
Hi from the LMMS team! Thanks for your contribution.

> To minimize the waiting time between preset switching, I changed the code to not delete the vestige instrument and plugin dll (when next preset uses the same plugin) every time preset is changed.

Is that the only reason? If it's the case, I think you may apply your logic to more general cases.

akimaze:
Hi @PhysSong 

> Is that the only reason? If it's the case, I think you may apply your logic to more general cases.

I tested this approach in about 20 VST plugins (I have about 500 working VST presets and I love that I can hear them in preset tree). However, I didn't want to change the way LMMS handles VST plug-ins outside the preview mode because it would significantly affect the way LMMS works. And I don't know enough about LMMS sources to be sure that it won't have a negative impact.
Therefore, I think it is better to enter it in the preview mode itself and if there are no bug reports that the preview mode does not work for a significant number of VST plugins, we can change the behavior of LMMS more widely.

The second reason is that sometimes VST plugins/wine stop working (crashes) and can be easily reloaded by loading the plug again. If we introduced this behavior that I did for preview. The user would have to load another plugin (to change the dll file path in Vestige) and then load the plugin back. Or maybe we should then add the "Reload" button to the Vestige window. But this is a bigger task that I did not want to address in this PR.

To sum up, I think it would be best to test this approach in preview mode in next release (1.3.0?) and if there are no bug reports about vst preset preview then we can consider changing how Vesitge works in other cases.

src/tracks/InstrumentTrack.cpp
```diff
@@ -771,7 +771,8 @@ void InstrumentTrack::saveTrackSpecificSettings( QDomDocument& doc, QDomElement
 
 void InstrumentTrack::loadTrackSpecificSettings( const QDomElement & thisElement )
 {
-	silenceAllNotes( true );
+	silenceAllNotes(!(m_previewMode && m_instrument && m_instrument->nodeName() == "vestige"
+					&& getInstrumentName(thisElement) == "vestige"));
```
PhysSong, line 775: Any reason for this change? I don't understand why did you add this exception.

src/tracks/InstrumentTrack.cpp
```diff
@@ -815,33 +816,44 @@ void InstrumentTrack::loadTrackSpecificSettings( const QDomElement & thisElement
...
 
-				emit instrumentChanged();
+				// don't delete vestige instrument in preview mode because loading vst
+				// can take long time, so better try change settings without recreating
+				// everything (this speeds up the loading of subsequent presets from one plugin)
+				if (m_previewMode && m_instrument && m_instrument->nodeName() == "vestige" &&
+						node.toElement().attribute("name") == "vestige")
+				{
+					m_instrument->restoreState(node.firstChildElement());
+					emit instrumentChanged();
```
PhysSong, line 831: You don't need this in this case because `m_instrument` doesn't change.

src/tracks/InstrumentTrack.cpp
```diff
@@ -815,33 +816,44 @@ void InstrumentTrack::loadTrackSpecificSettings( const QDomElement & thisElement
...
-					node.toElement().attribute( "name" ), this, &key);
-				m_instrument->restoreState( node.firstChildElement() );
+				PluginKey key(node.toElement().elementsByTagName("key").item(0).toElement());
 
-				emit instrumentChanged();
+				// don't delete vestige instrument in preview mode because loading vst
+				// can take long time, so better try change settings without recreating
+				// everything (this speeds up the loading of subsequent presets from one plugin)
+				if (m_previewMode && m_instrument && m_instrument->nodeName() == "vestige" &&
+						node.toElement().attribute("name") == "vestige")
```
PhysSong, line 828: I think you can compare two names directly rather than comparing it to `"vestige"` because this will help saving time for other instruments as well. I think we don't need to re-instantiate instrument plugins in general cases unless something errored on the previous load.

include/InstrumentTrack.h
```diff
@@ -230,6 +234,8 @@ class LMMS_EXPORT InstrumentTrack : public Track, public MidiEventProcessor
 		return "instrumenttrack";
 	}
 
+	QString getInstrumentName(const QDomElement & thisElement) const;
```
PhysSong, line 237: I think the name is misleading, but I don't have a good idea right now.

akimaze:
@PhysSong  Now I think I understand what you meant. I was constantly thinking about VeSTige because it is the only one instrument that loads extremely long. And you were thinking about other instruments in preview mode. Other instruments load/run a lot faster so I don't think about them before :)

src/tracks/InstrumentTrack.cpp
akimaze: That's because for `m_instrument` deleting there was `silenceAllNotes(true)`. But when you want not delete `m_instrument` you need `silenceAllNotes(false)` otherwise instrument is not reusable. I don't quite understand how `silenceAllNotes()` works, maybe it should be done in a different way.

src/tracks/InstrumentTrack.cpp
akimaze: Yeah, that can be changed if we want save time for other instruments.

PhysSong:
@akimaze Could you please read my reviews and tell me what do you want to do?

akimaze:
@PhysSong I read everything, I just focused on piano roll vert zoom and did not have enough time to make changes in this PR (I have a lot of work). I will try to fix it in a few days.

src/tracks/InstrumentTrack.cpp
```diff
@@ -771,7 +771,7 @@ void InstrumentTrack::saveTrackSpecificSettings( QDomDocument& doc, QDomElement
 
 void InstrumentTrack::loadTrackSpecificSettings( const QDomElement & thisElement )
 {
-	silenceAllNotes( true );
+	silenceAllNotes(!(m_previewMode && m_instrument && m_instrument->nodeName() == getInstrumentName(thisElement)));
```
PhysSong, line 774: I think we may extract the conditions in parentheses to a variable like `reuseInstrument` to make the meaning clear and reuse it below.

PhysSong:
@akimaze Can I add some minor formatting changes and merge this?

akimaze:
@PhysSong OK, Thanks!

Issue #2294, opened 2015-08-26
Vertical Zoom Support in Piano Roll


LYF610400210:
LMMS can have a "vertical zoom" button in the piano roll window. It seems that the blocks of notes are too small.

![Vertical Zoom](https://cloud.githubusercontent.com/assets/6359921/9488514/2865e9ca-4c0d-11e5-8e8d-fbcb38788a0a.PNG)

musikBear:
Yes the room for those labels are in-sufficient, but that will be changed. #2285

Spekular:
@LYF610400210 Can you explain whst this would be used for? Do you want to
be able to zoom out, in, or both?
@musikBear I'm sure if we really want to implement this we'll be able to
make room for it.

LYF610400210:
@Spekular The labels can be wider, so I will not add a new note by mistake when I want to click on the note to play it.
![Wider Labels](https://cloud.githubusercontent.com/assets/6359921/9494368/24c5afd2-4c38-11e5-96d4-e495f2f3c76b.png)
But I think the labels are too small to zoom out vertically.

tresf:
>  The labels can be wider, so I will not add a new note by mistake

Valid reasoning.  I'll categorize appropriately and leave this open for feedback.

musikBear:
> able to zoom out, in, or both?
IF something would be useful, i would say IN. That would aid selection of larger segments (but if time-spanned-selections comes, then that wont even be really interesting.)
The thing is, that expanding the notes in height, will first and foremost lead to less of the composition on screen, i really cant see the usecase for that?

@LYF610400210 (or anyone else)  -where will **that** be a benefit? 

> (**More** on screen -eg a result of 'thinner' notes *could*, at rare situations, benefit -like marking a few notes in or out of a existing selectionl.. )

tresf:
Probably the most sensible way of implementing this is to make the height of a [white key resizable](https://github.com/LMMS/lmms/blob/master/src/gui/editors/PianoRoll.cpp#L81-L82) and let the note heights follow suit.  After reviewing this enhancement again, I feel a full-fledged vertical zoom feature is overkill for the piano roll.  It's only useful up to about 200% and it would take up unnecessary real estate on the toolbar.

In some ways, I feel this qualifies more as an accessibility feature.

tresf:
As part of a pruning effort, this `enhancement` request is archived into a dedicated "Better Workflow" checklist here #4877.

Issue #4877, opened 2019-03-06
Better Workflow


tresf:
I'm creating a meta issue for better UI/UX workflow.  This list should be updated frequently because it's the source of many of our new bug reports.  If a new bug report arrives that seems workflow related **AND won't be closed by a quick patch** it can be marked as a duplicate and placed here.

Since workflow is so important to users -- something they can directly relate with -- this can also help serve as a futuremap if the developer team is to start milestoning certain enhancements (something we don't do a great job of currently)

Lastly, it can serve as a birds-eye-view (so to speak) of what our users really want. ❤️ 

💥 = Bug
🚀 = Good first issue for new developers

See also:

## Tracks
* #### Song Editor/Beat+Bassline Editor
   * [ ] Better transport/playback/playhead behavior
      * [ ]  Position indicator and line behavior inconsistent between Song-Editor, Piano-Roll, Beat/Bassline and Automation Editor #857, #2123, #3052
      * [ ] Global playback/transport #1357
      * [ ] Project should remember the last-used playhead behavior per #4192, #4556
      * [ ] Better keyboard shortcut behavior when mixer is in focus #1140 (related: #1488)
      * [ ] Better keyboard shorts in general #1144 (related: #1488)
      * [ ] Proposal to use `>>` as default playhead behavior #3740
      * [ ] Stop and return `|<<` versus "pause" default behavior #3740
      * [ ] Link JACK transport with LMMS transport #4412
      * [ ] Stop button should stop ALL sounds #1535
      * [ ] Playback should stop at the end of a track #4643
   * [ ] Better track color support #1006 #1160 #1176 #2332
   * [ ] Rename Song Editor/Beat+Bassline Editors #121
   * [ ] Add Beat/Bassline pattern previews #4845
   * [ ] Ability to cut/join patterns in the sequencer #3829 #3633
   * [ ] Show mixer channel on each instrument #2563
   * [ ] Volume control on individual B&B tracks #2218
   * [ ] Disabling or changing the Song Editor's Bar Grid #2171 
   * [ ] 🚀💥 Wrong cursor when moving blocks in song editor #879
   * [ ] Stretch/Shrink/Shorten Patterns via SongEditor #1658 
   * [ ] Inconsistent/confusing toolbar layout #4829
   * [ ] Better workflow when adding BB #2014
   * [ ] Annotations/markers/bookmarks #156 #380
   * [ ] 🚀 Open pattern to specific bar #3305
[...]

PR #5253, opened 2019-10-19
Refactor PianoRoll

Veratil:
Kind of a bigger PR, but fixes a few things: #4098, #5008 (partial, but easy to add to editors now).

In addition to those fixes, it reworks how the piano roll was painted before (and makes it somewhat easier to follow). Local tests of ~1000 `paintEvent` frames using the demo `Skiessi-C64.mmpz` put `master`'s `paintEvent` at ~6-6.1ms average time, this PR runs at ~4-4.1ms (improvement of ~33%) on `DEBUG` build.

General summary of updates:
* Split out SongEditor's `positionLine` class into an updated separate class file `PositionLine`
* No longer can the key area be resized larger than `NumKeys * KEY_LINE_HEIGHT` (108 * 12) pixels
* If window is sized larger than max key area size, note edit area is resized instead
* Key area snaps to `KEY_LINE_HEIGHT` to prevent partial key draws (except if top key is black key, the previous key is drawn in that case)
* Position marker added to timeline
* Simplifies some math, some if branches, and draw code (old code went through white keys then through black keys, now only one for loop)

EDIT: I've updated the title to be a little more general since I'm touching a little more than just `painteEvent` with this PR. It still focuses on `paintEvent`, but there's a lot more to it at this point.

PR #5442, opened 2020-04-07
Piano roll vertical zoom

akimaze:
Hi,

After several hours of work / study at Piano Roll, I noticed that too small notes cause eye pain. So I implemented vertical zoom to relieve my eyes. This significantly improves the comfort of work.
This fixes https://github.com/LMMS/lmms/issues/2294 from https://github.com/LMMS/lmms/issues/4877.

### Screenshots
Default value is the same:
pianoroll_100_100

Personally, I prefer 150% in both directions, which is similar to FL Studio:
pianoroll_150_150

Of course, the size can be specified in each direction separately:
pianoroll_100_200

### Notes
I did not change the names of the variables related to horizontal zoom (like `zoomingChanged()`, `m_zoomingComboBox` etc.) to facilitate the code review.

Regards

### DIFF
--- a/include/PianoRoll.h
+++ b/include/PianoRoll.h
@@ -192,6 +192,8 @@ protected:
 	// for entering values with dblclick in the vol/pan bars
 	void enterValue( NoteVector* nv );
 
+	void updateYScroll();
+
 protected slots:
 	void play();
 	void record();
@@ -217,6 +219,7 @@ protected slots:
 	void updatePositionStepRecording(const MidiTime & t );
 
 	void zoomingChanged();
+	void zoomingYChanged();
 	void quantizeChanged();
 	void noteLengthChanged();
 	void quantizeNotes();
@@ -330,12 +333,14 @@ private:
 	static TextFloat * s_textFloat;
 
 	ComboBoxModel m_zoomingModel;
+	ComboBoxModel m_zoomingYModel;
 	ComboBoxModel m_quantizeModel;
 	ComboBoxModel m_noteLenModel;
 	ComboBoxModel m_scaleModel;
 	ComboBoxModel m_chordModel;
 
 	static const QVector<double> m_zoomLevels;
+	static const QVector<double> m_zoomYLevels;
 
 	Pattern* m_pattern;
 	NoteVector m_ghostNotes;
@@ -384,6 +389,12 @@ private:
 	int m_notesEditHeight;
 	int m_ppb;  // pixels per bar
 	int m_totalKeysToScroll;
+
+	static int s_keyLineHeight;
+	int m_octaveHeight;
+	int m_whiteKeySmallHeight;
+	int m_whiteKeyBigHeight;
+	int m_blackKeyHeight;
 
 	// remember these values to use them
 	// for the next note that is set
@@ -501,6 +512,7 @@ private:
 	PianoRoll* m_editor;
 
 	ComboBox * m_zoomingComboBox;
+	ComboBox * m_zoomingYComboBox;
 	ComboBox * m_quantizeComboBox;
 	ComboBox * m_noteLenComboBox;
 	ComboBox * m_scaleComboBox;
--- a/include/PianoRoll.h
+++ b/include/PianoRoll.h
@@ -180,7 +180,7 @@ protected:
 	void focusOutEvent( QFocusEvent * ) override;
 
 	int getKey( int y ) const;
-	static void drawNoteRect( QPainter & p, int x, int y,
+	void drawNoteRect( QPainter & p, int x, int y,
 					int  width, const Note * n, const QColor & noteCol, const QColor & noteTextColor,
 					const QColor & selCol, const int noteOpc, const bool borderless, bool drawNoteName );
 	void removeSelection();
@@ -390,7 +390,7 @@ private:
 	int m_ppb;  // pixels per bar
 	int m_totalKeysToScroll;
 
-	static int s_keyLineHeight;
+	int m_keyLineHeight;
 	int m_octaveHeight;
 	int m_whiteKeySmallHeight;
 	int m_whiteKeyBigHeight;
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -79,26 +79,9 @@ const int PIANO_X = 0;
 
 const int WHITE_KEY_WIDTH = 64;
 const int BLACK_KEY_WIDTH = 41;
-/*const int WHITE_KEY_SMALL_HEIGHT = 18;
-const int WHITE_KEY_BIG_HEIGHT = 24;
-const int BLACK_KEY_HEIGHT = 16;
-const int KEY_LINE_HEIGHT = 12;
-const int OCTAVE_HEIGHT = KEY_LINE_HEIGHT * KeysPerOctave;	// = 12 * 12;*/
 
-// double size
-/*const int WHITE_KEY_SMALL_HEIGHT = 27; //18*2; // key line * 1.5
-const int WHITE_KEY_BIG_HEIGHT = 36; //24*2; // key line * 2
-const int BLACK_KEY_HEIGHT = 24; //16*2; // keyline * 1.3333
-const int KEY_LINE_HEIGHT = 18; //12*2;
-const int OCTAVE_HEIGHT = KEY_LINE_HEIGHT * KeysPerOctave;	// = 12 * 12;*/
-
-const int KEY_LINE_HEIGHT = 24;
-const int DEFAULT_KEY_LINE_WIDTH = 12;
-const int OCTAVE_HEIGHT = KEY_LINE_HEIGHT * KeysPerOctave;
-const int WHITE_KEY_SMALL_HEIGHT = KEY_LINE_HEIGHT * 1.5;
-const int WHITE_KEY_BIG_HEIGHT = KEY_LINE_HEIGHT * 2;
-const int BLACK_KEY_HEIGHT = round(KEY_LINE_HEIGHT * 1.3333);
-
+const int DEFAULT_KEY_LINE_HEIGHT = 12;
+const int DEFAULT_CELL_WIDTH = 12;
 
 
 const int NOTE_EDIT_RESIZE_BAR = 6;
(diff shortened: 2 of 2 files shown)


PhysSong:
@Veratil I'm requesting your review since you also opened a PR(#5253) which touches the piano roll a lot. I think you may need to pay attention to this one because this one can affect your one.

include/PianoRoll.h
```diff
@@ -385,6 +390,12 @@ protected slots:
 	int m_ppb;  // pixels per bar
 	int m_totalKeysToScroll;
 
+	static int s_keyLineHeight;
```
Veratil, line 393: Is there a reason this needs to be `static`?

Veratil:
> @Veratil I'm requesting your review since you also opened a PR(#5253) which touches the piano roll a lot. I think you may need to pay attention to this one because this one can affect your one.

It does affect my PR, but I'm okay with this merging before mine and rebasing.

include/PianoRoll.h
akimaze: That's because `PianoRoll::drawNoteRect()` is `static`.

Veratil: I'm looking through and I don't see a need for `drawNoteRect` to be static either.
@PhysSong @tresf any insight on this?

PhysSong: I think it should be a member variable because it's a property of the piano-roll instance. Now `PianoRoll::drawNoteRect()` should know the value of `s_keyLineHeight` because it's the height of the note to draw. We can either make `PianoRoll::drawNoteRect()` nonstatic or add a height parameter.

src/gui/editors/PianoRoll.cpp
```diff
@@ -171,6 +175,10 @@ PianoRoll::PianoRoll() :
 	m_oldNotesEditHeight( 100 ),
 	m_notesEditHeight( 100 ),
 	m_ppb( DEFAULT_PR_PPB ),
+	m_octaveHeight(s_keyLineHeight * KeysPerOctave),
+	m_whiteKeySmallHeight(round(s_keyLineHeight * 1.5)),
+	m_whiteKeyBigHeight(round(s_keyLineHeight * 2)),
+	m_blackKeyHeight(round(s_keyLineHeight * 1.3333)),
```
PhysSong, line 181: I think it's also possible to use integer multiplications and divisions to save the computation power, but the difference will be negligible.

src/gui/editors/PianoRoll.cpp
```diff
@@ -4279,6 +4306,17 @@ void PianoRoll::zoomingChanged()
 }
 
 
+void PianoRoll::zoomingYChanged()
+{
+	s_keyLineHeight = m_zoomYLevels[m_zoomingYModel.value()] * DEFAULT_KEY_LINE_HEIGHT;
+	m_octaveHeight = s_keyLineHeight * KeysPerOctave;
+	m_whiteKeySmallHeight = round(s_keyLineHeight * 1.5);
+	m_whiteKeyBigHeight = round(s_keyLineHeight * 2);
+	m_blackKeyHeight = round(s_keyLineHeight * 1.3333);
```
PhysSong, line 4315: The same for here.

src/gui/editors/PianoRoll.cpp
akimaze: Without `round()` I sometimes saw 1 pixel position errors - especially for black keys.

include/PianoRoll.h
akimaze: I would prefer a solution with removing `static` from the `PianoRoll::drawNoteRect()` function (if using `static` in it is not intentional). This function already has a lot of params. 

I didn't know why `PianoRoll::drawNoteRect() is `static` so I did `keyLineHeight` static too.

src/gui/editors/PianoRoll.cpp
akimaze: See comment above, this is recalculated only once when zoom changes so the overhead is small.

src/gui/editors/PianoRoll.cpp
PhysSong: I mean you can do it like `(s_keyLineHeight * 8 + 4) / 6`, though it looks unclear a bit.

include/PianoRoll.h
akimaze: So I will remove the static modifier from `PianoRoll::drawNoteRect()` if no one protests.

src/gui/editors/PianoRoll.cpp
akimaze: @PhysSong  So, should I change that? It's hard to decide (for me) is it a small optimization or readability of the code more important. Usually in such cases I choose the readability of the code. I removed unnecessary `round()` when multiplying by 2.

include/PianoRoll.h
akimaze: Fixed by https://github.com/LMMS/lmms/pull/5442/commits/c69f02677a280ab50b12914ff9747e1741dca4a6

SecondFlight:
Visually, it's bugging me that the note text is not vertically centered at 150% and 200% zoom. Is this fixable?

image

akimaze:
@SecondFlight, @PhysSong I fixed note text to be centered, but this solution needs Qt5.8 (because it uses `capHeight()`). Other solution is to use something like: 
```
int const yOffset = ((noteHeight + noteTextHeight - fontMetrics.descent() + 1) * 8 + 8) / 16;
```
But with this there will be slight text position errors (1 px too high or too low - what looks not so good). 

Or maybe someone knows better way to get height of the font in Qt?

src/gui/editors/PianoRoll.cpp
```diff
@@ -1009,7 +1009,7 @@ void PianoRoll::drawNoteRect( QPainter & p, int x, int y,
 
 			int const distanceToBorder = 2;
 			int const xOffset = borderWidth + distanceToBorder;
-			int const yOffset = (noteHeight + noteTextHeight) / 2;
+			int const yOffset = ((noteHeight + fontMetrics.capHeight()) * 8 + 8) / 16;
```
zonkmachine, line 1012: I think this is equal to

```cpp
int const yOffset = ((noteHeight + fontMetrics.capHeight()) + 1) / 2;
```

Veratil:
> @SecondFlight, @PhysSong I fixed note text to be centered, but this solution needs Qt5.8 (because it uses `capHeight()`).

Qt5.8 has been released since Jan 2017, so I *think* we should be okay with this.

> Or maybe someone knows better way to get height of the font in Qt?

We already get the `textSize` here:
```cpp
QSize textSize = fontMetrics.size(Qt::TextSingleLine, noteKeyString);
```

Why not use `textSize.height()` ?

tresf:
> Qt5.8 has been released since Jan 2017, so I think we should be okay with this.

Pretty sure Travis + Win32 still builds with 5.6 tho. 🤯 🔫

akimaze:
> Why not use textSize.height() ?

Hi @Veratil, Sorry for modifying the same files.
I was surprised myself. For a 23px note, setting the text height to 18 pixels, the text drawn is 13px (I made a screenshot and measured in Gimp). And `textSize.height()` returns 26 px - probably Qt adds some margins here and space for font descent, overline etc. I also tried use `ascent()` without success.

My simple test:
```
qDebug() << "noteHeight " << noteHeight;
qDebug() << "noteTextHeight " << noteTextHeight;
qDebug() << "textSize.height() " << textSize.height();
qDebug() << "fontMetrics.capHeight() " << fontMetrics.capHeight();
```

Results:
```
noteHeight  23
noteTextHeight  18
textSize.height()  26
fontMetrics.capHeight()  13
```
On Linux Mint.

tresf:
As a reminder, any Qt5.8-only features are perfectly welcome, but if one of the CI systems fail as a result, just guard with a precompiler check.

PhysSong:
I tried many different combinations and found that `(noteHeight + fontMetrics.ascent() - fontMetrics.descent()) / 2` is quite usable for most(but not all) fonts.
Also, I think the label on the piano keys should translate(and probably scale) according to the vertical zoom as well.

PhysSong:
This looks better to me:
```cpp
QRect const boundingRect = fontMetrics.boundingRect(QChar::fromLatin1('H'));
int const yOffset = (noteHeight - boundingRect.top() - boundingRect.bottom()) / 2;
```
I chose 'H' because it's a flat character.

akimaze:
For me, it generates the same spacing as `capHeight()`. But it's better because it doesn't require newer Qt. Thanks!

akimaze:
> Also, I think the label on the piano keys should translate(and probably scale) according to the vertical zoom as well.

I'll try to fix it tonight.

akimaze:
I fixed the positions of the key names, but I did not scale the names because I don't want to change the width of the keyboard. A wider keyboard would only take up more space (unnecessarily).

akimaze:
Don't know why latest commit (https://github.com/akimaze/lmms/commit/101c0d290d24dad43ab4b31cfb23c473c5711208) is not visible on PR :confused:

src/gui/editors/PianoRoll.cpp
```diff
@@ -2867,6 +2867,8 @@ void PianoRoll::paintEvent(QPaintEvent * pe )
 
 	// set font-size to 8
 	p.setFont( pointSize<8>( p.font() ) );
+	QFontMetrics fontMetrics(p.font());
+	int const fontHeight = fontMetrics.boundingRect(QChar::fromLatin1('H')).height();
```
PhysSong, line 2871: The reason I used `- boundingRect.top() - boundingRect.bottom()` is that a font can span both the area above and the area below the baseline. It means using `height()` may give a different value for some fonts while the desired value is the center of the area.

src/gui/editors/PianoRoll.cpp
```diff
@@ -2868,7 +2868,8 @@ void PianoRoll::paintEvent(QPaintEvent * pe )
 	// set font-size to 8
 	p.setFont( pointSize<8>( p.font() ) );
 	QFontMetrics fontMetrics(p.font());
-	int const fontHeight = fontMetrics.boundingRect(QChar::fromLatin1('H')).height();
+	QRect const boundingRect = fontMetrics.boundingRect(QChar::fromLatin1('H'));
+	int const fontHeight = - boundingRect.top() - boundingRect.bottom();
```
PhysSong, line 2872: Now `fontHeight` isn't the height anymore, it's two times of the y coordinate of the center of the bounding rectangle.

PhysSong:
I think it's almost complete now. :rocket:

src/gui/editors/PianoRoll.cpp
akimaze: I can't find any intuitive name, maybe `boundingRectY`, `fontDeltaY` any suggestions?

Spekular: Is that not, per definition, the height? The bounding box bottom should align with the bottom of the text, or it's not a true bounding box. Same for the top. Finally, top - bottom = height, no? It includes ascenders and descenders, but that's not incorrect since we don't call it cap height ot x-height.

PhysSong: @Spekular it's `-(top+bottom)=-2(center)`.

akimaze: @PhysSong @Spekular We think about space for note label so maybe change from `font` to `label` `noteLabelHeight`, `labelHeight`, `labelArea` or `labelSpace`?

PhysSong: For making the meaning clear, I think we need something like `twoTimesBaselineYOffset`, but I think it's too long. We can use some other name and add a comment which describes what the value means.

PR #5449, opened 2020-04-13
Fix #4098

### LINKED ISSUE #4098: Piano roll crashes when window is scaled too large
Tested on 1.2.0-rc5

LMMS will crash when the piano roll window is scaled too large. I found that it seems to be when it's scaled to around 1600-ish pixels in height.

Veratil:
This should fix #4098 for stable-1.2. I've tested resizing and placing notes, etc. I'd like another to test playing around with resizing the window as much as they can to make sure nothing else gets out of place though.

### DIFF
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -2662,7 +2662,12 @@ void PianoRoll::paintEvent(QPaintEvent * pe )
 			break;
 	}
 	// start drawing at the bottom
-	int key_line_y = keyAreaBottom() - 1;
+	int key_line_y = qMin(keyAreaBottom() - 1, OCTAVE_HEIGHT * 9);
+	// we need to set m_notesEditHeight here because it needs to fill in the
+	// rest of the window if key_line_y is bound to OCTAVE_HEIGHT*9
+	if (key_line_y == OCTAVE_HEIGHT * 9) {
+		m_notesEditHeight = height() - (PR_TOP_MARGIN + OCTAVE_HEIGHT * 9);
+	}
 	// used for aligning black-keys later
 	int first_white_key_height = WHITE_KEY_SMALL_HEIGHT;
 	// key-counter - only needed for finding out whether the processed
@@ -2782,9 +2787,10 @@ void PianoRoll::paintEvent(QPaintEvent * pe )
 	key = m_startKey;
 	keys_processed = 0;
 	int white_cnt = 0;
+	key_line_y = qMin(keyAreaBottom() - 1, OCTAVE_HEIGHT * 9);
 
 	// and go!
-	for( int y = keyAreaBottom() + y_offset;
+	for( int y = key_line_y + y_offset;
 					y > PR_TOP_MARGIN; ++keys_processed )
 	{
 		// check for black key that is only half visible on the bottom
@@ -2850,16 +2856,16 @@ void PianoRoll::paintEvent(QPaintEvent * pe )
 
 	// erase the area below the piano, because there might be keys that
 	// should be only half-visible
-	p.fillRect( QRect( 0, keyAreaBottom(),
-			WHITE_KEY_WIDTH, noteEditBottom() - keyAreaBottom() ), bgColor );
+	p.fillRect( QRect( 0, key_line_y,
+			WHITE_KEY_WIDTH, noteEditBottom() - key_line_y ), bgColor );
 
 	// display note editing info
 	QFont f = p.font();
 	f.setBold( false );
 	p.setFont( pointSize<10>( f ) );
 	p.setPen( noteModeColor() );
-	p.drawText( QRect( 0, keyAreaBottom(),
-					  WHITE_KEY_WIDTH, noteEditBottom() - keyAreaBottom() ),
+	p.drawText( QRect( 0, key_line_y,
+					  WHITE_KEY_WIDTH, noteEditBottom() - key_line_y ),
 			   Qt::AlignCenter | Qt::TextWordWrap,
 			   m_nemStr.at( m_noteEditMode ) + ":" );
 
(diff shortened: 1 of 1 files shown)


PhysSong:
I think you should replace `OCTAVE_HEIGHT * 9` with either `OCTAVE_HEIGHT * NumOctaves` or `KEY_LINE_HEIGHT * NumKeys` for various reason. I prefer the latter to the former because we might change the number of keys to not a multiple of 12.

**Note**: This and #5442 might affect each other like #5253.

Veratil:
> I think you should replace `OCTAVE_HEIGHT * 9` with either `OCTAVE_HEIGHT * NumOctaves` or `KEY_LINE_HEIGHT * NumKeys` for various reason. I prefer the latter to the former because we might change the number of keys to not a multiple of 12.

Ah, good catch. I've adjusted to use `KEY_LINE_HEIGHT` now.

JohannesLorenz:
There's one issue still, though I'm not sure if we need to fix it: Open a piano roll, resize piano roll so it's getting big. Move it down to the bottom and resize it to the upper limit (that way, you enlarge the PianoRoll, thanks to @PhysSong ). Do that 3 or 4 times. Now, the notes don't start at C0 anymore, and when you play some keys, the green key markers don't apply to those notes that you play. (It's not the same bug that I report for the master PR)

Veratil:
> 
> 
> There's one issue still, though I'm not sure if we need to fix it: Open a piano roll, resize piano roll so it's getting big. Move it down to the bottom and resize it to the upper limit (that way, you enlarge the PianoRoll, thanks to @PhysSong ). Do that 3 or 4 times. Now, the notes don't start at C0 anymore, and when you play some keys, the green key markers don't apply to those notes that you play. (It's not the same bug that I report for the master PR)

I can't reproduce this on my build?

Veratil:
> 
> 
> > There's one issue still, though I'm not sure if we need to fix it: Open a piano roll, resize piano roll so it's getting big. Move it down to the bottom and resize it to the upper limit (that way, you enlarge the PianoRoll, thanks to @PhysSong ). Do that 3 or 4 times. Now, the notes don't start at C0 anymore, and when you play some keys, the green key markers don't apply to those notes that you play. (It's not the same bug that I report for the master PR)
> 
> I can't reproduce this on my build?

Okay I've managed to reproduce it now, you have to resize *fast*.

JohannesLorenz:
> Okay I've managed to reproduce it now, you have to resize _fast_.

If you can find a quick fix, that would be good, but if not...

I wonder if it's an important issue, as long as the maximize button works (can't test it). If it even happens to anyone, is it a workaround to close the editor and open again, or something like that?

Veratil:
> 
> 
> > Okay I've managed to reproduce it now, you have to resize _fast_.
> 
> If you can find a quick fix, that would be good, but if not...
> 
> I wonder if it's an important issue, as long as the maximize button works (can't test it). If it even happens to anyone, is it a workaround to close the editor and open again, or something like that?

Workaround is to resize again, but slow, by any amount, even a pixel.

JohannesLorenz:
OK, I'll continue the review tomorrow then. Is the issue on your master-fix of the same type?

Veratil:
> 
> 
> OK, I'll continue the review tomorrow then. Is the issue on your master-fix of the same type?

Trying the same method I cannot reproduce on master's rework.

JohannesLorenz:
> Trying the same method I cannot reproduce on master's rework.

I swear I saw something not working, but I may have missed up. I confirm the master-fix is now working.

JohannesLorenz:
Functionally OK. Testing OK.

Ready for squash & merge?

Veratil:
If we're going to merge #5442 into `stable-1.2` then we'll want to wait. If not, then we can.

@PhysSong your call for that

Spekular:
Not the person you asked, but vertical zoom is a feature and as such I don't believe we should target 1.2.2 with it. Especially considering it would delay a bugfix.

Next, the only thing I notice in a review of the changes is that ``key_line_y`` is set to ``keyAreaBottom() - 1``, but then you replace instances of ``keyAreaBottom()`` directly with ``key_line_y``. Doesn't this shift those things down a pixel?

Veratil:
@Spekular @PhysSong if we're good with this, shall we merge?

Phys, I did notice you had an idea in #4098 changing the functions rather than paintEvent. Do you feel that's a better solution or is this okay since this is just a bugfix for 1.2 and my other PR changes the paintEvent function entirely for 1.3+?

Spekular:
For my understanding, the key change in that diff is the same qMin call, just placed directly in keyAreaBottom(). keyAreaTop() is defined to be equal to PR_TOP_MARGIN, so those swaps don't matter, and the rest of the changes are also equivalent (replacing what used to be duplicate code with calls to keyAreaBottom()). All in all it seems like a cleaner fix, so I'd rather we merge that if it's tested and working.

PhysSong:
As it turned out that either solution has some issues. I think we should consider potential conflicts with #5253.
So, I'd defer the decision to @Veratil.

PhysSong:
For information, this is what @Veratil said on Discord:
> I want to solve the missing key issue for my solution
> Once I do that, then I'm okay with merging it

See also: https://github.com/LMMS/lmms/issues/4098#issuecomment-618805283

Veratil:
> 
> 
> For information, this is what @Veratil said on Discord:
> 
> > I want to solve the missing key issue for my solution
> > Once I do that, then I'm okay with merging it
> 
> See also: [#4098 (comment)](https://github.com/LMMS/lmms/issues/4098#issuecomment-618805283)

I'm still not sure what to do to fix this missing key problem. I've tried a few things which I thought would work by adjusting the calculated height and all, but none of it worked. I don't want to hold up 1.2 on this issue any longer honestly. Feel free to merge this.

**The workaround for this issue though is:** do not have the window size maxed revealing all piano keys, scroll to the top then drag the notes edit area down. This will reveal `B8` along with however much more you drag (`C9` and up).

This being a known issue is okay for me, since for 1.3 we have #5253 which will overwrite this fix for cleaner and faster code.

JohannesLorenz:
@Veratil Thanks for the analysis

I'll merge in 3 days if there will be no further comments about the missing key, mentioned in https://github.com/LMMS/lmms/issues/4098#issuecomment-618805283

PR #5451, opened 2020-04-13
Remove plugins/LadspaEffect/swh/ladspa-util.c

JohannesLorenz:
This file contains no used code and it caused build problems, so we
remove it. Thanks to @plater for the issue report.


JohannesLorenz:
No review required at all.

But if anyone has an veto, let me know. Otherwise, I'll just merge in 3 days.

Issue #5453, opened 2020-04-14
Add "Open File Location" option to right-click menu in My Samples browser


ttepatti:
### Enhancement Summary

An additional button should be added when right-clicking a sample or folder containing samples in the "My Samples" browser, allowing users to open the file or folder's location in their system's file browser.

#### Justification

This makes it easy to locate sample files on your system from within LMMS if you need to edit, delete, rename, or otherwise modify your samples.

#### Mockup

This screenshot is from FL Studio, but displays the functionality.
![image](https://user-images.githubusercontent.com/7002801/79172989-47ce8400-7dc4-11ea-8b18-8d53a3628d56.png)

michaelgregorius:
#5454 implements a first solution which opens the folder that contains the selected file. It looks like this:
![5453-OpenContainingFolder-Cropped](https://user-images.githubusercontent.com/9293269/79227152-c5939d80-7e5f-11ea-9307-cec9d005279e.png)

PR #5454, opened 2020-04-14
Add "Open containing folder" (#5453)

michaelgregorius:
Add functionality to open the containing folder of a file that's selected
in LMMS' file browser.

Technical details
------------------
Add a new private method openContainingFolder to FileBrowser. Add a new
action to the context menu of a selected file. This action in turn calls
the added method.

The current implementation of openContainingFolder delegates to
QDesktopServices::openUrl with the directory of the selected file. Please
note that this will only open the directory but not select the file as
this is much more complicated due to different implementations that are
needed for the different platforms (Linux/Windows/MacOS).

Using QDesktopServices::openUrl seems to be the most simple cross
platform way which uses functionality that's already available in Qt.

### DIFF
--- a/include/FileBrowser.h
+++ b/include/FileBrowser.h
@@ -125,9 +125,9 @@ private slots:
 	void activateListItem( QTreeWidgetItem * item, int column );
 	void openInNewInstrumentTrackBBE( void );
 	void openInNewInstrumentTrackSE( void );
-	void openContainingFolder();
 	void sendToActiveInstrumentTrack( void );
 	void updateDirectory( QTreeWidgetItem * item );
+	void openContainingFolder();
 
 } ;
 
--- a/include/FileBrowser.h
+++ b/include/FileBrowser.h
@@ -125,6 +125,7 @@ private slots:
 	void activateListItem( QTreeWidgetItem * item, int column );
 	void openInNewInstrumentTrackBBE( void );
 	void openInNewInstrumentTrackSE( void );
+	void openContainingFolder();
 	void sendToActiveInstrumentTrack( void );
 	void updateDirectory( QTreeWidgetItem * item );
 
--- a/src/gui/FileBrowser.cpp
+++ b/src/gui/FileBrowser.cpp
@@ -364,30 +364,41 @@ QList<QString> FileBrowserTreeWidget::expandedDirs( QTreeWidgetItem * item ) con
 
 void FileBrowserTreeWidget::contextMenuEvent(QContextMenuEvent * e )
 {
-	FileItem * f = dynamic_cast<FileItem *>( itemAt( e->pos() ) );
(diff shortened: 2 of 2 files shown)


michaelgregorius:
Implements #5453.

I propose to merge it like this which for now only opens the directory of the file as this is already an improvement for the end use.

In case someone wants to enhance it by also selecting the file in the opened file browser some hints on how to do this can be found here: https://stackoverflow.com/questions/3490336/how-to-reveal-in-finder-or-show-in-explorer-with-qt.

src/gui/FileBrowser.cpp
```diff
@@ -380,6 +381,11 @@ void FileBrowserTreeWidget::contextMenuEvent(QContextMenuEvent * e )
 								"B+B Editor" ),
 						this,
 					SLOT( openInNewInstrumentTrackBBE() ) );
+		contextMenu.addSeparator();
+		contextMenu.addAction( QIcon(embed::getIconPixmap( "folder" )),
```
Veratil, line 385: Clean these four lines up following [Coding Conventions](https://github.com/LMMS/lmms/wiki/Coding-Conventions)

include/FileBrowser.h
```diff
@@ -125,6 +125,7 @@ private slots:
 	void activateListItem( QTreeWidgetItem * item, int column );
 	void openInNewInstrumentTrackBBE( void );
 	void openInNewInstrumentTrackSE( void );
+	void openContainingFolder();
```
Veratil, line 128: I think putting this at the bottom of this block would be better, so after the `updateDirectory` line.

src/gui/FileBrowser.cpp
michaelgregorius: Done with commit b85aef2f33794eb6bd7fa1153f7c411ef14c5b7c. I have cleaned up the whole method.

include/FileBrowser.h
michaelgregorius: Done with commit b85aef2f33794eb6bd7fa1153f7c411ef14c5b7c.

src/gui/FileBrowser.cpp
Veratil: Usually we try not to touch other lines unless there is a need, as eventually we'll have an autoformatter whenever that PR gets merged, but I think this is okay for now.

PhysSong:
> Please note that this will only open the directory but not select the file

Could you add a comment to the code itself as well, to make the intention clear?

michaelgregorius:
> > Please note that this will only open the directory but not select the file
> 
> Could you add a comment to the code itself as well, to make the intention clear?

Hi @PhysSong , done with commit c37fdd005529f1fbc19d974c62d617fe89dfe988.

Issue #3642, opened 2017-06-15
Native EQ Out Slider Ignores Effect Wet/Dry Knob


UnityParadox:
At full dry, the EQ will continue to output the equalized signal at the volume specified on the out slider, alongside the original dry audio.

System Details: Windows 10, LMMS 1.2.0 rc2

BaraMGB:
Is this fixed by https://github.com/LMMS/lmms/pull/3520?

UnityParadox:
@BaraMGB I believe I discovered the issue after the fact of that pull request.

PhysSong:
Confirmed with latest master. Input gain and output gain are applied to dry signal.
I think it is not hard to fix, @curlymorphic?

curlymorphic:
> System Details: Windows 10, LMMS 1.2.0 rc2

The pull request to fix this bug was merged on 02 may 2017, so would not have been present in the build under test.

Would someone please test this in RC3.

PhysSong:
I changed the title because #3505 is partially fixed by #3520. However, Dry signals are **still** affected by IN/OUT gain sliders.

zonkmachine:
Isn't the reason that the gain is done by letting a separate function accessing the buffer?
https://github.com/LMMS/lmms/blob/3745bfbd7fb7b90f682eecd7cd9c7c264122f95f/plugins/Eq/EqEffect.h#L45

https://github.com/LMMS/lmms/blob/master/plugins/Eq/EqEffect.cpp#L176
https://github.com/LMMS/lmms/blob/master/plugins/Eq/EqEffect.cpp#L272

DomClark:
Shouldn't we handle all the dry/wet stuff in `Effect`, outside the call to `processAudioBuffer`, rather then leaving it to each individual effect to implement? The noise control on the flanger also doesn't interact properly with the dry/wet control, and these mistakes could easily be avoided while making the code a bit cleaner.

he29-net:
>Shouldn't we handle all the dry/wet stuff in Effect, outside the call to processAudioBuffer, rather then leaving it to each individual effect to implement?

How would you deal with unknown signal latency of the effect? Say you feed in a 1 kHz sine wave, the effect does something with the signal and then produces output 0.5 ms later. When you mix the dry and wet signals (which are now exactly opposite phase), I imagine they would mostly cancel each other out.

DomClark:
> How would you deal with unknown signal latency of the effect?

[...]

PR #5458, opened 2020-04-19
Add Compressor effect to LMMS

LostRobotMusic:
(Simple video demo: https://www.youtube.com/watch?v=j-i0gY_KmVk)

This took a few months to make, the visualizer in the back taking the bulk of the development time.  I wanted to make this the best audio compressor in the world... and while it's entirely subjective which compressor is the best, I'm very satisfied with how this turned out.

This is a purely digital compressor, completely void of any coloring of any kind so you can focus on exactly what a compressor is meant for, being compression.  Even with 2572985782 db of input gain, unless your parameter settings are asking for it, you'll hear absolutely no distortion whatsoever.

It comes with some fairly unique features, like the toggleable Feedback mode, the Hold control (which is usually only for Gates), the blending between stereo linking modes, the mid/side compression, and the 100% adjustable automatic compression and release, which is significantly more complex than automatic attack/release in most compressors.

The automatic attack/release in this compressor analyzes the crest factor (peak divided by RMS) and adjusts those parameters automatically depending on that value.  The knobs control how much this has an impact on those parameters, giving you ultimate flexibility which I've never seen in any other compressor.

It also features fully-automatable lookahead, which applies a constant 20 ms of latency when enabled regardless of the lookahead time.  Along with L/R or M/S balancing for either the sidechain input or the audio itself, a sidechain tilt filter which can transform into a highpass, lowpass, or a hybrid between the two, adjustable RMS sizes that can go all the way up to 21.5 ms, a Mix knob which takes lookahead delay into account, and an audition button that can be used to listen to the delta signal of the compressor to know exactly how the compressor is changing your sound, this is by far LMMS's most complex audio effect and I'm very proud to be the author of it.

And well, nobody likes a compressor without a fancy visualizer to assist you.  Or maybe they do.  But those people are weird.  The visualizer has a flowing, moving graph showing the input volume, output volume, and gain reduction.  It also contains a visualizer of the compressor's dynamics curve, so you can see exactly how your dynamics are being impacted at any time depending on your compressor settings.

The compressor window is fully resizeable.  You can just drag the corner of the window, and the contents will scale to fit the new size.  Also, the visualizer supports zooming in all the way to 3 db or all the way out to 96 db, so you can see what the compressor is doing to your audio with whatever amount of detail best suits your needs.  All in all, a very cool plugin.  :)

This'll definitely need a code/style review.  I abandoned this project three times or so, so the code may be kind of a mess.  I tried to clean it up the best I could.

(Special thank you to H4CKTON3 (aka RoxasKH on Github) for helping me immensely with the artwork!)


qnebra:
At this moment compressor work really nice. No crashes, no bugs at this moment.

PhysSong:
Why is the unit of RMS size 1/44100 second? Isn't it more natural to use something like milliseconds?

plugins/Compressor/CompressorControlDialog.h
```diff
@@ -0,0 +1,234 @@
...
+constexpr int KNEE_SCREEN_Y = COMP_SCREEN_Y;
+constexpr int COMP_KNEE_LINES = 20;
+constexpr int COMP_BOX_X = 720;
+constexpr int COMP_BOX_Y = 280;
+
+constexpr float COMP_NOISE_FLOOR = 0.000001;// -120 dbFs
+
+// -2.2 seems to be the best value for this.
+// I've seen some compressors choose -1, however.
+constexpr float COMP_LOG = -2.2;
```
PhysSong, line 57: FYI, some articles say there's no industry standard for the definition of attack and release values.

PhysSong: And I think it should be in `Compressor.h`(probably with `COMP_NOISE_FLOOR`).

plugins/Compressor/Compressor.cpp
```diff
@@ -0,0 +1,630 @@
...
+{
+	delete m_rms[0];
+	delete m_rms[1];
+}
+
+
+float CompressorEffect::timeToCoeff(float time)
+{
+	// Convert time in milliseconds to applicable lowpass coefficient
+	return expf(COMP_LOG / (time * m_sampleRate * 0.001f));
```
PhysSong, line 107: You may precalculate `COMP_LOG / (m_sampleRate * 0.001f)` for optimization.
Also, it's hard to know that the unit of `time` without looking at the implementation. You may add `ms` in the function name or the parameter name.

plugins/Compressor/Compressor.cpp
```diff
@@ -0,0 +1,630 @@
...
+void CompressorEffect::calcRatio()
+{
+	m_ratioVal = 1.f / m_compressorControls.m_ratioModel.value();
+	m_redrawKnee = true;
+}
+
+void CompressorEffect::calcRange()
+{
+	// Range is inactive when turned all the way down
+	m_rangeVal = (m_compressorControls.m_rangeModel.value() > m_compressorControls.m_rangeModel.minValue()) ? dbfsToAmp(m_compressorControls.m_rangeModel.value()) : 0;
```
PhysSong, line 177: This line and some other lines look too long.

plugins/Compressor/Compressor.cpp
```diff
@@ -0,0 +1,630 @@
...
+					m_maxLookaheadTimer[i] = m_lookaheadLength;
+				}
+
+				// Decrement timer.  When the timer reaches 0, that means the
+				// stored maximum value has left the buffer and a new
+				// maximum value must be found.
+				if (--m_maxLookaheadTimer[i] <= 0)
+				{
+					m_maxLookaheadTimer[i] = std::distance(std::begin(m_lookaheadBuf[i]),
+						std::max_element(std::begin(m_lookaheadBuf[i]), std::begin(m_lookaheadBuf[i]) + m_lookaheadLength));
```
PhysSong, line 363: Note, it's possible to optimize finding the maximum value in a ring buffer using a double-ended queue. I can provide the detail if you want.
BTW, why do you want to use the maximum value in the lookahead buffer?

plugins/Compressor/Compressor.cpp
```diff
@@ -0,0 +1,630 @@
...
+						gainResult[0] = linearInterpolate(temp1, temp2, blend - 2);
+						gainResult[1] = gainResult[0];
+					}
+				}
+				break;
+			}
+		}
+
+		// Bias compression to the left or right (or mid or side)
+		if (stereoBalance)
```
PhysSong, line 475: I think it's better to explicitly compare this to 0.

plugins/Compressor/Compressor.cpp
PhysSong: It seems like this doesn't consume CPU much, so I think the existing comment is enough which says about the possibility of further optimization.

softrabbit: > Note, it's possible to optimize finding the maximum value in a ring buffer using a double-ended queue. I can provide the detail if you want.

Would that be faster than O(n)? 😮

PhysSong: It's amortized O(1) for fixed lookahead value, and additional O(log(n)) if it's automated.

DouglasDGI:
Rather than using true RMS, I'm now instead using a buffer-free approximation which sounds better in a compression context.  I also made it so RMS is in units of milliseconds rather than samples.

On top of that, I fixed a bug with auto gain which made it act... incorrectly... as well as a bug making it so the input fader displayed the output volume.

I was also notified of some bug involving a crash every time the lookahead knob was moved up all the way.  I never experienced this bug on my system, but their error message suggested an out-of-range index, so I bumped up the lookahead buffer size to 21 milliseconds rather than 20.  That'll probably fix the issue.

plugins/Compressor/Compressor.cpp
```diff
@@ -0,0 +1,636 @@
...
+	connect(&m_compressorControls.m_autoAttackModel, SIGNAL(dataChanged()), this, SLOT(calcAutoAttack()));
+	connect(&m_compressorControls.m_autoReleaseModel, SIGNAL(dataChanged()), this, SLOT(calcAutoRelease()));
+
+	connect(&m_compressorControls.m_thresholdModel, SIGNAL(dataChanged()), this, SLOT(calcAutoMakeup()));
+	connect(&m_compressorControls.m_ratioModel, SIGNAL(dataChanged()), this, SLOT(calcAutoMakeup()));
+	connect(&m_compressorControls.m_kneeModel, SIGNAL(dataChanged()), this, SLOT(calcAutoMakeup()));
+	connect(&m_compressorControls.m_autoMakeupModel, SIGNAL(dataChanged()), this, SLOT(calcAutoMakeup()));
+
+	connect(Engine::mixer(), SIGNAL(sampleRateChanged()), this, SLOT(changeSampleRate()));
+	emit changeSampleRate();
```
PhysSong, line 88: Here as well.

DouglasDGI:
Everything is ready now as far as I know.  Probably needs some more testing though, to be sure.

plugins/Compressor/Compressor.cpp
```diff
@@ -0,0 +1,642 @@
...
+
+CompressorEffect::~CompressorEffect()
+{
+}
+
+
+float CompressorEffect::msToCoeff(float ms)
+{
+	// Convert time in milliseconds to applicable lowpass coefficient
+	return expf(m_coeffPrecalc / ms);
```
Sawuare, line 102: The single `expf()` here is inconsistent with the overloaded `exp()`s elsewhere.

plugins/Compressor/Compressor.cpp
```diff
@@ -0,0 +1,642 @@
...
+	m_compressorControls.m_inPeakR = rInPeak;
+
+	return isRunning();
+}
+
+
+// Regular modulo doesn't handle negative numbers correctly.  This does.
+inline int CompressorEffect::realmod(int k, int n)
+{
+	return ((k %= n) < 0) ? k+n : k;
```
Sawuare, line 575: Excess parantheses around the condition of `?:`.

plugins/Compressor/Compressor.cpp
```diff
@@ -0,0 +1,642 @@
...
+}
+
+// Regular fmod doesn't handle negative numbers correctly.  This does.
+inline float CompressorEffect::realfmod(float k, float n)
+{
+	return ((k = fmod(k, n)) < 0) ? k+n : k;
```
Sawuare, line 581: Excess parantheses around the condition of ?:.

plugins/Compressor/Compressor.cpp
```diff
@@ -0,0 +1,642 @@
...
+						gainResult[0] = linearInterpolate(temp1, temp2, blend - 2);
+						gainResult[1] = gainResult[0];
+					}
+				}
+				break;
+			}
+		}
+
+		// Bias compression to the left or right (or mid or side)
+		if (stereoBalance != 0)
```
Sawuare, line 479: You avoided comparing to 0 elsewhere. I think you should do the same here.

plugins/Compressor/Compressor.cpp
```diff
@@ -0,0 +1,642 @@
...
+	else if (abs(1 - m_thresholdVal) < m_kneeVal)// Within knee
+	{
+		const float temp = -m_thresholdVal + m_kneeVal;
+		tempGainResult = ((m_compressorControls.m_limiterModel.value() ? 0 : m_ratioVal) - 1) * temp * temp / (4 * m_kneeVal);
+	}
+	else// Above knee
+	{
+		tempGainResult = m_compressorControls.m_limiterModel.value()
+			? m_thresholdVal
+			: (m_thresholdVal - m_thresholdVal * m_ratioVal);
```
Sawuare, line 125: Excess parantheses around the third operand of `?:`.

plugins/Compressor/Compressor.cpp
```diff
@@ -0,0 +1,642 @@
...
+			else if (abs(currentPeakDbfs - m_thresholdVal) < m_kneeVal)// Within knee
+			{
+				const float temp = currentPeakDbfs - m_thresholdVal + m_kneeVal;
+				gainResult[i] = currentPeakDbfs + ((limiter ? 0 : m_ratioVal) - 1) * temp * temp / (4 * m_kneeVal);
+			}
+			else// Above knee
+			{
+				gainResult[i] = limiter
+					? m_thresholdVal
+					: (m_thresholdVal + (currentPeakDbfs - m_thresholdVal) * m_ratioVal);
```
Sawuare, line 425: Excess parantheses around the third operand of `?:`.

plugins/Compressor/Compressor.cpp
DouglasDGI: The other areas are boolean.  `stereoBalance` is a float (from -1 to 1), so I think it would be best to explicitly compare it to zero.

plugins/Compressor/Compressor.cpp
Sawuare: You did fix the inconsistency in 57e7e01, but you did it backwards, and now there's an inconsistency with the "f-less" `log()` and `abs()`. In my previous comment I meant `expf()` -> `exp()`.

firewall1110:
I tested this branch by compiling from source in Debian Buster using clang compiler. 

It seems all work fine. 

P.S.

Nice interface, but I would like some check box to hide controls |& background. 

[I am new in testing ...]

plugins/Compressor/Compressor.cpp
```diff
@@ -374,19 +373,25 @@ bool CompressorEffect::processAudioBuffer(sampleFrame* buf, const fpp_t frames)
 
 			if (t > m_yL[i])// Attack phase
 			{
+				// Lowest possible value of m_crestFactorVal[i] is 2, given by a sine wave.
```
PhysSong, line 376: Isn't it 1 from a square wave?

DouglasDGI: Yes.  Sleep is important.

plugins/Compressor/Compressor.cpp
```diff
@@ -0,0 +1,651 @@
...
+		}
+		else if (autoMakeup)
+		{
+			s[0] *= m_autoMakeupVal;
+			s[1] *= m_autoMakeupVal;
+		}
+
+		// Calculate wet/dry value results
+		const float temp1 = trueDrySignal[0] / m_inGainVal;
+		const float temp2 = trueDrySignal[1] / m_inGainVal;
```
PhysSong, line 557: This may not be the REAL dry signal because `m_inGainVal` may differ at the moment queued into `m_inputBuf` and at this point. If you set the W/D ratio to 0 and move the input gain slider very fast, you will get some volume artifacts.
The simplest trick will be saving the dry signal both w/ and w/o input gain. The ultimate solution will be implementing W/D handling with proper latency compensation, as @DomClark said in https://github.com/LMMS/lmms/issues/3642#issuecomment-554559218.

plugins/Compressor/Compressor.cpp
```diff
@@ -0,0 +1,651 @@
...
+
+				inputValue = m_maxLookaheadVal[i];
+			}
+
+			float t = inputValue;
+
+			if (t > m_yL[i])// Attack phase
+			{
+				// We want the "resting value" of our crest factor to be with a sine wave,
+				// which with this variable has a value of 2.
```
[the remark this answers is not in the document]
DouglasDGI: I did modify the comment.  I forgot my reasoning behind doing things when I programming this a few months ago, and the comment was brand new, which is why it was incorrect at first.  I wanted the attack and release to remain unmodified when the input is a sine wave.  Then the values will be decreased with higher crest factor values, or increased with lower crest factor values.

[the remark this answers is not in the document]
PhysSong: Okay, got it.

plugins/Compressor/CompressorControls.h
```diff
@@ -0,0 +1,101 @@
...
+	void saveSettings(QDomDocument & _doc, QDomElement & _parent) override;
+	void loadSettings(const QDomElement & _this) override;
+	inline QString nodeName() const override
+	{
+		return "CompressorControls";
+	}
+
+	int controlCount() override
+	{
+		return 31;
```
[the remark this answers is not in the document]
DouglasDGI: Not sure what was going through my head back then.  Maybe I was counting GUI elements instead of models for some reason.  I'm still not sure what that function is used for, I'll check.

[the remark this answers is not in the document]
PhysSong: For now, it doesn't matter unless you put 0 here. LMMS uses this value only to check if a plugin has any controls.

plugins/Compressor/Compressor.cpp
DouglasDGI: Oh, nice catch.  For some reason I used `trueDrySignal[0] / m_inGainVal` instead of just `buf[f][0]`.

PhysSong: I think you used that to use delayed dry signal for lookahead mode. However, `buf[f][0]` is never delayed.

DouglasDGI: Yeah.  Maybe I should rename the variable to `delayedDrySignal`.

PhysSong: Detailed explanation: The previous code supported W/D mixing with proper latency compensation, but without proper gain correction. Now you get the correct gain for the dry signal, but you now have phase differences between W/D signal even if the compressor does nothing.

plugins/Compressor/Compressor.cpp
```diff
@@ -0,0 +1,651 @@
...
+}
+
+
+
+void CompressorEffect::calcAutoMakeup()
+{
+	// Formulas using the compressor's Threshold, Ratio, and Knee values to estimate a good makeup gain value
+
+	float tempGainResult;
+	if (-m_thresholdVal < -m_kneeVal)// Below knee
```
PhysSong, line 114: If I read the code correctly, `m_thresholdVal` is in `[-60, 0]` and `m_kneeVal` is in `[0, 48]`(both in dBFS). So I think you're doing something wrong with this conditionals.

plugins/Compressor/Compressor.cpp
```diff
@@ -0,0 +1,651 @@
...
+	{
+		tempGainResult = 0;
+	}
+	else if (abs(1 - m_thresholdVal) < m_kneeVal)// Within knee
```
PhysSong, line 118: Where does the 1 come from? isn't it just `-m_thresholdVal`?

plugins/Compressor/Compressor.cpp
```diff
@@ -0,0 +1,651 @@
...
+			s[0] = -s[0] + delayedDrySignal[0];
+			s[1] = -s[1] + delayedDrySignal[1];
+		}
+		else if (autoMakeup)
+		{
+			s[0] *= m_autoMakeupVal;
+			s[1] *= m_autoMakeupVal;
+		}
+
+		// Calculate wet/dry value results
```
PhysSong, line 555: As a reminder: proper latency compensation between W/D signal is required when lookahead is enabled. It's even possible to provide both fixed and minimum latency mode, but that's not that important until we implement general latency compensation feature.

plugins/Compressor/Compressor.cpp
DouglasDGI: The problem here is that the tooltips and such say knee is in terms of dBFS rather than dB.  `m_kneeVal` is the distance between the threshold and the edge of the knee (note that variable is half the value of the knob), so the code is entirely correct.  The labels mix up dBFS and dB though, so I'll fix those.

PhysSong: Yes, the unit of `m_kneeVal` is dB, indeed. The formula itself is correct, but this block will never execute.

PhysSong: One more minor comment: you're using `db` and `dbFs` instead of `dB` and `dBFS`.

DouglasDGI: > One more minor comment: you're using db and dbFs instead of dB and dBFS.

Yeah, just barely fixed that a couple minutes ago when I noticed it.  :+1: 

> Yes, the unit of m_kneeVal is dB, indeed. The formula itself is correct, but this block will never execute.

Ah, you're right.  In an older version of this plugin I allowed the Threshold to go up to +12 dbFs, but that's not longer allowed.

plugins/Compressor/Compressor.cpp
PhysSong: If I understood correctly, the separation of region is given by:
- `level < threshold - knee`: below the knee
- `threshold - knee < level < threshold + knee`: in the knee
- `threshold + knee < level`: above the knee

`level` is 0(dBFS) here.

DouglasDGI: Oh, nice catch, I was mixing up amplitude and dBFS, which is why I used 1 instead of 0.

PhysSong: Just noticed you have a similar code in `CompressorEffect::processAudioBuffer`, but you made a mistake when you assumed `currentPeakDbfs` is equivalent to 1 here.

PhysSong: Also I should mention you don't need a `abs` here and Line 425.

plugins/Compressor/Compressor.cpp
PhysSong: - If you use `delayedDrySignal[0]`, the dry signal will be affected by the input gain
- If you use `delayedDrySignal[0] / m_inGainVal`, you'll get weird gain artifacts on automating the input gain
- if you use `buf[f][0]`, you will get an artifact like flanger when lookahead is enabled

The solution is storing the dry signal **before** applying input gain into a ring buffer(`delayedDrySignal` is the signal after applying input gain). I'm not sure if the values for other controls should be evaluated before or after the delay.

rdrpenguin04:

superpaik:
Some thoughs about the plugin:
- To me it works well, though I'm not an audio engineer, so I can't tested it in that way. I tested it for audio spill/noise and it seems to not have any. The only thing I found that I don't know if it's correct or not, is that when having the mix knob on (over 0%) the input slider works weird to me: the higher the input the lower the volumen produced. It's that correct?
- Though it's somehow transparent, I think the knob's section hide a lot of information (only the last 15db are clearly visible). I'd be great to have the possibility to hide that section or have a slide to make it more transparent.
- As for the knobs, some visible scale would be great as well, because you have to click to see the value.
- The pluggin has a lot of possibilities regarding compression (more than most of commercial one's). I think users would need a manual to understand the possibilites of the pluggin, otherwise it's use seem overwhelming.
- A request: having a SideChain option like the Calf SideChain CompressorLDSPA, if it's not very complicated to implement. Having the visual aid to set the sidechain effect would be great ;-)

Very good work! Thanks for the effort!

DouglasDGI:
> The only thing I found that I don't know if it's correct or not, is that when having the mix knob on (over 0%) the input slider works weird to me: the higher the input the lower the volumen produced. It's that correct?

Oh gosh.  I guess I was so focused on fixing input gain bugs for the Autogain and Lookahead and such that I didn't realize I broke it for literally everything else.  Fixed in the latest commit, it should work properly now.

> Though it's somehow transparent, I think the knob's section hide a lot of information (only the last 15db are clearly visible). I'd be great to have the possibility to hide that section or have a slide to make it more transparent.

If you scroll on the background, it will change its scale so you can see more.  Also, I made the Compressor fully resizable, so you can resize the Compressor window to open up more room for you to see the visualizer.

> A request: having a SideChain option like the Calf SideChain CompressorLDSPA, if it's not very complicated to implement. Having the visual aid to set the sidechain effect would be great ;-)

I agree that that would be a great feature to have, but I am waiting for LMMS to support multiple audio inputs to effects so that I can allow the user to directly send their audio into the compressor's sidechain signal, rather than using a knob and peak controller as messy workaround that we'll have to worry about backwards compatibility for later.

> I think users would need a manual to understand the possibilites of the pluggin, otherwise it's use seem overwhelming.

I put a lot of information in the knob tooltips that you can check out, but yes, maybe an in-plugin manual is called for here since many of the features may seem a bit hidden (e.g. scrolling to resize the visualizer)...

superpaik:
Great! It work ok now!
Sorry, I didn't get the scroll on the background. It's very good!
I think it's a great stock pluggin!

plugins/Compressor/Compressor.cpp
```diff
@@ -0,0 +1,648 @@
...
+	calcRange();
+	calcLookaheadLength();
+	calcHold();
+	resizeRMS();
+	calcThreshold();
+	calcKnee();
+	calcOutGain();
+	calcInGain();
+	calcTiltCoeffs();
+	calcMix();
```
Veratil, line 633: Is this order *THE ORDER* it should be done in? If so then a comment saying this order shouldn't change should be added.

plugins/Compressor/CompressorControlDialog.cpp
```diff
@@ -0,0 +1,753 @@
...
+	m_lastGainPoint = yGainPoint;
+
+	update();
+}
+
+
+
+void CompressorControlDialog::paintEvent(QPaintEvent *event)
+{
+	if (isVisible())
```
Veratil, line 565: If we invert this guard to return early we can save unnecessary indentation.

plugins/Compressor/CompressorControlDialog.cpp
```diff
@@ -0,0 +1,753 @@
...
+void CompressorControlDialog::resizeEvent(QResizeEvent *event)
+{
+	resetCompressorView();
+}
+
+
+void CompressorControlDialog::wheelEvent(QWheelEvent * event)
+{
+	const float temp = m_dbRange;
+	m_dbRange = round(qBound(3.f, m_dbRange - event->delta() / 20.f, 96.f) / 3.f) * 3.f;
```
Veratil, line 609: A comment about where the constants come from can help in case wheel event deltas change or cause issues on other systems. What defines the bounds? Why 3? Are any of these a target for `SOME_CONSTANT_VARIABLE_NAME`? etc.

plugins/Compressor/CompressorControlDialog.cpp
```diff
@@ -0,0 +1,753 @@
...
+	p.begin(&m_graphPixmap);
+
+	p.setRenderHint(QPainter::Antialiasing, false);
+
+	p.setCompositionMode(QPainter::CompositionMode_Source);
+	p.fillRect(0, 0, m_windowSizeX, m_windowSizeY, QColor("transparent"));
+	p.setCompositionMode(QPainter::CompositionMode_SourceOver);
+
+	p.setPen(QPen(m_textColor, 1));
+	p.setFont(QFont("Arial", qMax(int(m_windowSizeY / 1080.f * 24), 12)));
```
Veratil, line 634: Where does this font's `pointSize` calculation come from and will it work on all systems?

plugins/Compressor/CompressorControlDialog.cpp
```diff
@@ -0,0 +1,753 @@
...
+
+	// Redraw graph
+	p.setPen(QPen(m_graphColor, 1));
+	for (int i = 1; i < m_dbRange / 3.f + 1; ++i)
+	{
+		p.drawLine(0, dbfsToYPoint(-3 * i), m_windowSizeX, dbfsToYPoint(-3 * i));
+		p.drawLine(dbfsToXPoint(-3 * i), 0, dbfsToXPoint(-3 * i), m_kneeWindowSizeY);
+		p.drawText(QRectF(m_windowSizeX - 50, dbfsToYPoint(-3 * i), 50, 50), Qt::AlignRight | Qt::AlignTop, QString::number(i * -3));
```
Veratil, line 642: There's a lot of constants here that could use defining what they mean. I'm seeing `3` a lot.

plugins/Compressor/CompressorControlDialog.cpp
```diff
@@ -0,0 +1,753 @@
...
+void CompressorControlDialog::setInVolColor(const QColor & c){m_inVolColor = c;}
+void CompressorControlDialog::setOutVolAreaColor(const QColor & c){m_outVolAreaColor = c;}
+void CompressorControlDialog::setOutVolColor(const QColor & c){m_outVolColor = c;}
+void CompressorControlDialog::setGainReductionColor(const QColor & c){m_gainReductionColor = c;}
+void CompressorControlDialog::setKneeColor(const QColor & c){m_kneeColor = c;}
+void CompressorControlDialog::setKneeColor2(const QColor & c){m_kneeColor2 = c;}
+void CompressorControlDialog::setThreshColor(const QColor & c){m_threshColor = c;}
+void CompressorControlDialog::setTextColor(const QColor & c){m_textColor = c;}
+void CompressorControlDialog::setGraphColor(const QColor & c){m_graphColor = c;}
+void CompressorControlDialog::setResetColor(const QColor & c){m_resetColor = c;}
```
Veratil, line 753: If we use `Q_PROPERTY`'s `MEMBER` we can remove all this. These can be auto generated by moc during compile time.

plugins/Compressor/CompressorControlDialog.h
```diff
@@ -0,0 +1,230 @@
...
+	void setThreshColor( const QColor & c );
+
+	QColor const & textColor() const;
+	void setTextColor( const QColor & c );
+
+	QColor const & graphColor() const;
+	void setGraphColor( const QColor & c );
+
+	QColor const & resetColor() const;
+	void setResetColor( const QColor & c );
```
Veratil, line 112: Instead of `READ ... WRITE ...` we can use `MEMBER m_variable` to auto-generate the getter/setter functions. Saves a lot of boilerplate repetition that just sets or returns values.

plugins/Compressor/CompressorControlDialog.h
```diff
@@ -0,0 +1,230 @@
...
+	QColor m_inVolColor = QColor(209, 216, 228, 100);
+	QColor m_outVolAreaColor = QColor(209, 216, 228, 30);
+	QColor m_outVolColor = QColor(209, 216, 228, 240);
+	QColor m_gainReductionColor = QColor(180, 100, 100, 210);
+	QColor m_kneeColor = QColor(39, 171, 95, 255);
+	QColor m_kneeColor2 = QColor(9, 171, 160, 255);
+	QColor m_threshColor = QColor(39, 171, 95, 100);
+	QColor m_textColor = QColor(209, 216, 228, 50);
+	QColor m_graphColor = QColor(209, 216, 228, 50);
+	QColor m_resetColor = QColor(200, 100, 15, 200);
```
Veratil, line 168: I'm fine with this, but I think we normally set these on the constructor.

plugins/Compressor/CompressorControlDialog.cpp
```diff
@@ -0,0 +1,753 @@
...
+	}
+
+	if (!m_controls->m_effect->isEnabled() || !m_controls->m_effect->isRunning())
+	{
+		m_controls->m_effect->m_displayPeak[0] = COMP_NOISE_FLOOR;
+		m_controls->m_effect->m_displayPeak[1] = COMP_NOISE_FLOOR;
+		m_controls->m_effect->m_displayGain[0] = 1;
+		m_controls->m_effect->m_displayGain[1] = 1;
+		//m_lastPoint = dbfsToYPoint(-9999);
+		//m_lastGainPoint = dbfsToYPoint(0);
```
Veratil, line 384: Debugging leftover?

RoxasKH:
Ok, so, i don't wanna be the stop-merging guy, but here's what i think about it.

The default size of the plugin is a bit big for HD screens, but the plugin is resizable, so this is not a big problem.

Now let's talk about what i still think it's the real issue in this plugin. The autogain feature.
I'm not an expert in dsp, neither i have still taken a glance at LMMS code. But as a user of LMMS and as i produce music, i feel like it's not functional at all.

For how it works at the moment it makes the sound incredibly clips, and this obviously forces you to do a manual decrease in volume after activating it. But this feels nosense. What's the point in having an autogain button that forces you to do a manual adjusting in gain after? You could just do a manual increase in gain from the start, in 1 move instead of 2. And this is actually how most compressor works out there, they just have a postgain knob, just to make an example i could mention the LADSPA version of Calf Compressor which is native in LMMS. It's way more rare to find compressor with an auto gain function, because it's probably difficult to code i guess.

But according to my experience, the autogain button should do something like detecting the peak of the sound after the compression and bring it up to 0db. To make an example, i had my hands on the WA Production SphereComp as i bought the whole Sphere bundle on sale. This compressor has an autogain function, which isn't always able to bringup the sound to 0db, but for sure doesn't make the sound clip so much, if most of the times not at all.

To mention another plugin, the MCompressor by Melda Production have an autogain function even if it's not called this way, "Maximize to 0db"; the name itself it's pretty explicit, even if haven't tried the vst myself.

I feel like bringing this up because if it won't be fixed now, in my opinion it will come up multiple times as an issue in the future, too. With the little differences that it won't be fixable without breaking retrocompatibility.

The plugin is great, LMMS needs it, and every other feature seems fine to me, but i'd rather have a working auto gain button.
I don't know if this is fixable (maybe considering it brings up the sound the the limit, an auto decrease after the increase could be programmed), but if there's any possibility to make it works i'd say properly, it'd be better to do it, in my opinion.

As i said i'm not a coder, neither a dsp expert, but this is what i feel from my user experience.

DouglasDGI:
I think your concerns are due to confusion around what the auto gain function is actually for.

The auto gain function is *not* meant to choose a decent output volume for you; that depends on the context and is up to the musician.  The auto gain function is used to maximize the compression shape to 0 dbFs, so a 0 dbFs input will result in a 0 dbFs target output.  This allows the musician to change the compression amount with minimal loudness change, so they can better hear what the compression is doing to the sound.

The clipping you're encountering is not an inherent issue.  Again, this function is not intended to choose an output volume for you (and isn't a mastering tool either), so you should fix the clipping that results due to your settings by adjusting the output gain.

> But according to my experience, the autogain button should do something like detecting the peak of the sound after the compression and bring it up to 0db.

This isn't possible with realtime audio.  And if we were to make use of the lookahead to try to make it possible, it would result in unwanted upward compression when the Auto Gain isn't supposed to change over time at all.

> To mention another plugin, the MCompressor by Melda Production have an autogain function even if it's not called this way, "Maximize to 0db"; the name itself it's pretty explicit, even if haven't tried the vst myself.

You mentioned the "Maximize to 0 db" function in Melda Audio's MCompressor.  The Auto Gain button in this compressor is actually identical to that, they do exactly the same thing.  In fact, Fabfilter's Pro-C 2 also does the same thing (except they also lower the volume depending on the attack time, which I decided to not bother with).  The feature is working exactly as intended.

plugins/Compressor/CompressorControlDialog.cpp
DouglasDGI: This is for changing the font size when the window is resized.  The font size calculation is entirely arbitrary, I just changed the formula until it looked good.  It just increases the font size by 1 for every 45 pixels in the screen's height, while not allowing the font size to go below 12.
It... should work on all systems.  In what circumstances could it potentially not work?

plugins/Compressor/Compressor.cpp
DouglasDGI: I didn't think the order mattered at first, but I just noticed there are some incredibly rare niche edgecases in which calcAutoMakeup being before Threshold/Ratio/Knee could potentially cause an issue.  I'll fix that.

plugins/Compressor/CompressorControlDialog.cpp
DouglasDGI: It's 3 db per grid line.  I'll add a const variable for that.

plugins/Compressor/CompressorControlDialog.cpp
DouglasDGI: It's 3 db per grid line.  I'll add a const variable for that.

RoxasKH:
> 
> 
> I think your concerns are due to confusion around what the auto gain function is actually for.
> 
> The auto gain function is _not_ meant to choose a decent output volume for you; that depends on the context and is up to the musician. The auto gain function is used to maximize the compression shape to 0 dbFs, so a 0 dbFs input will result in a 0 dbFs target output. This allows the musician to change the compression amount with minimal loudness change, so they can better hear what the compression is doing to the sound.
> 
> The clipping you're encountering is not an inherent issue. Again, this function is not intended to choose an output volume for you (and isn't a mastering tool either), so you should fix the clipping that results due to your settings by adjusting the output gain.
> 
> > But according to my experience, the autogain button should do something like detecting the peak of the sound after the compression and bring it up to 0db.
> 
> This isn't possible with realtime audio. And if we were to make use of the lookahead to try to make it possible, it would result in unwanted upward compression when the Auto Gain isn't supposed to change over time at all.
> 
> > To mention another plugin, the MCompressor by Melda Production have an autogain function even if it's not called this way, "Maximize to 0db"; the name itself it's pretty explicit, even if haven't tried the vst myself.
> 
> You mentioned the "Maximize to 0 db" function in Melda Audio's MCompressor. The Auto Gain button in this compressor is actually identical to that, they do exactly the same thing. In fact, Fabfilter's Pro-C 2 also does the same thing (except they also lower the volume depending on the attack time, which I decided to not bother with). The feature is working exactly as intended.

Ok, thanks again for the clarifying, i was misunderstanding something.
As i said i hadn't tried MCompressor myself, so i wasn't aware of the real job of the feature.
Just to talk about the fabfilter addition, i don't even get why the decreasing should be done basing on the attack time (unless it's not talking about the attack time of the compressor but of some other attack time).

DouglasDGI:
Fabfilter Pro-C 2 does that because increasing the attack time results in an overall increase in volume (and therefore perceived loudness as well), but when I analyzed the scale at which attack time became auto gain volume reduction... it was really added in a fairly lazy way from what I could see, basically just turning the attack knob into a linear volume knob without really taking the scale of the attack time into account (...from what I can remember from over a year ago when I tested this).  I decided to not bother with that aspect, it seemed rather unnecessary anyway.

plugins/Compressor/CompressorControlDialog.cpp
Veratil: > It... should work on all systems. In what circumstances could it potentially not work?

We never know. Just wanted to clarify!

plugins/Compressor/CompressorControlDialog.cpp
DouglasDGI: Oh, thanks for catching that issue.  The formula was a temporary lazy formula that just worked, not something I intended to keep there.  Something as simple as somebody using a different mouse or something could potentially mess with this formula in undesirable ways I think.  I'll fix that.

RoxasKH:
> 
> 
> Fabfilter Pro-C 2 does that because increasing the attack time results in an overall increase in volume (and therefore perceived loudness as well), but when I analyzed the scale at which attack time became auto gain volume reduction... it was really added in a fairly lazy way from what I could see, basically just turning the attack knob into a linear volume knob without really taking the scale of the attack time into account (...from what I can remember from over a year ago when I tested this). I decided to not bother with that aspect, it seemed rather unnecessary anyway.

Oh got it, well the volume increase would be only at the start of the sound tho, so

DouglasDGI:
No, it's for the attack time, as in the value that the attack knob is set to.
Automatic gain compensation will *never* change over time.  It'll only change when a knob's value is changed (in this case, either the threshold, the knee, or the ratio).  Otherwise, it would change the dynamics of the sound in its own ways, which would obviously not fall in line with what we want the effect to do...

RoxasKH:
> 
> 
> No, it's for the attack time, as in the value that the attack knob is set to.
> Automatic gain compensation will _never_ change over time. It'll only change when a knob's value is changed (in this case, either the threshold, the knee, or the ratio). Otherwise, it would change the dynamics of the sound in its own ways, which would obviously not fall in line with what we want the effect to do...

Nono, i was just saying that increase the value of the attack knob increase the volume only of the start of the sound, as the compressor follows it to compress the sound. Just stating that the increase wouldn't be on the whole sound, nothing more.

DouglasDGI:
The volume change would be on the entire sound though.

plugins/Compressor/Compressor.cpp
```diff
@@ -0,0 +1,648 @@
...
+					: m_thresholdVal + (currentPeakDbfs - m_thresholdVal) * m_ratioVal;
+			}
+
+			gainResult[i] = dbfsToAmp(gainResult[i]) / m_yL[i];
+			gainResult[i] = qMax(m_rangeVal, gainResult[i]);
+		}
+
+		switch (stereoLink)
+		{
+			case 1:// Maximum
```
Veratil, line 439: We should probably make these a constant or enum.

plugins/Compressor/CompressorControlDialog.cpp
```diff
@@ -0,0 +1,757 @@
...
+	ToolTip::add(m_autoAttackKnob, tr("Automatically control attack value depending on crest factor"));
+
+	m_autoReleaseKnob = new Knob(knobStyled, this);
+	makeSmallKnob(m_autoReleaseKnob, tr("Auto Release:") , "%");
+	m_autoReleaseKnob->setModel(&controls->m_autoReleaseModel);
+	ToolTip::add(m_autoReleaseKnob, tr("Automatically control release value depending on crest factor"));
+
+
+
+
```
Veratil, line 173: Noticed a lot of empty lines here is all

plugins/Compressor/CompressorControlDialog.cpp
```diff
@@ -0,0 +1,757 @@
...
+	const float gainAvg = (m_controls->m_effect->m_displayGain[0] + m_controls->m_effect->m_displayGain[1]) * 0.5f;
+
+	QPainter p;
+	float yPoint = dbfsToYPoint(ampToDbfs(peakAvg));
+	float yGainPoint = dbfsToYPoint(ampToDbfs(gainAvg));
+
+	int threshYPoint = dbfsToYPoint(m_controls->m_effect->m_thresholdVal);
+	int threshXPoint = m_kneeWindowSizeY - threshYPoint;
+
+	p.begin(&m_visPixmap);
```
Veratil, line 395: I think this (and every new `p.begin()` "block") is a good candidate for a separate helper function, as to not make this `updateDisplay` function too complex. Then we just call the function `drawX(p)`.

plugins/Compressor/CompressorControlDialog.cpp
```diff
@@ -0,0 +1,757 @@
...
+	resetCompressorView();
+}
+
+
+void CompressorControlDialog::wheelEvent(QWheelEvent * event)
+{
+	const float temp = m_dbRange;
+	m_dbRange = round(
+		qBound(COMP_GRID_SPACING, m_dbRange - copysignf(COMP_GRID_SPACING, event->delta()), COMP_GRID_MAX)
+		/ COMP_GRID_SPACING) * COMP_GRID_SPACING;
```
Veratil, line 611: Maybe pull the `qBound` out to a temp variable in order to make this format better. As it sits this is sort of hard to read.

plugins/Compressor/CompressorControlDialog.cpp
```diff
@@ -0,0 +1,757 @@
...
+	if (m_dbRange != temp)
+	{
+		resetGraph();
+		m_controls->m_effect->m_redrawKnee = true;
+		m_controls->m_effect->m_redrawThreshold = true;
+	}
+}
+
+
+void CompressorControlDialog::resetGraph()
```
Veratil, line 623: Maybe along with pulling the other draw blocks out from `updateDisplay`, this could be renamed to `drawGraph` instead?

RoxasKH:
Just wanna state that any misunderstanding has been cleared here and on discord.

DouglasDGI:
Everything's finished here for now.  Ready for merge as long as there aren't other issues with it.

plugins/Compressor/CompressorControlDialog.cpp
```diff
@@ -0,0 +1,758 @@
...
+#include "Compressor.h"
+#include "CompressorControlDialog.h"
+#include "CompressorControls.h"
+
+#include "embed.h"
+#include "GuiApplication.h"
+#include "gui_templates.h"
+#include "interpolation.h"
+#include "MainWindow.h"
+#include <QGraphicsOpacityEffect>
```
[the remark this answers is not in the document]
DouglasDGI: Ah, right.  I used it in an older version, forgot to remove it.

plugins/Compressor/Compressor.cpp
```diff
@@ -0,0 +1,652 @@
...
+
+		// Calculate wet/dry value results
+		const float temp1 = delayedDrySignal[0];
+		const float temp2 = delayedDrySignal[1];
+		buf[f][0] = d * temp1 + w * s[0];
+		buf[f][1] = d * temp2 + w * s[1];
+		buf[f][0] = (1 - m_mixVal) * temp1 + m_mixVal * buf[f][0];
+		buf[f][1] = (1 - m_mixVal) * temp2 + m_mixVal * buf[f][1];
+
+		outSum += s[0] + s[1];
```
DomClark, line 563: `outSum` should be computed with the squares of the wet/dry mixed values. See the amplifier code: https://github.com/LMMS/lmms/blob/00ac4f58f50d67b79ad805ec11f1fb507962b2b6/plugins/Amplifier/Amplifier.cpp#L123-L125

plugins/Compressor/CompressorControlDialog.cpp
```diff
@@ -0,0 +1,757 @@
...
+ * License along with this program (see COPYING); if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA 02110-1301 USA.
+ *
+ */
+
+#include <QLayout>
+
+#include "Compressor.h"
+#include "CompressorControlDialog.h"
```
DomClark, line 28:
```suggestion
#include "CompressorControlDialog.h"

#include <QLayout>

#include "Compressor.h"
```
Related header file must be included first. See the [coding conventions](https://github.com/LMMS/lmms/wiki/Coding-Conventions#include-order).

plugins/Compressor/CompressorControlDialog.h
```diff
@@ -0,0 +1,218 @@
...
+	PixmapButton * minimumButton;
+	PixmapButton * blendButton;
+	automatableButtonGroup * stereoLinkGroup;
+
+	PixmapButton * autoMakeupButton;
+	PixmapButton * auditionButton;
+	PixmapButton * feedbackButton;
+	PixmapButton * lookaheadButton;
+
+	QTime m_timeElapsed;
```
DomClark, line 212: Consider using [`QElapsedTimer`](https://doc.qt.io/qt-5/qelapsedtimer.html) instead.

plugins/Compressor/CompressorControls.cpp
```diff
@@ -0,0 +1,147 @@
...
+ * Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA 02110-1301 USA.
+ *
+ */
+
+
+#include <QDomElement>
+
+#include "CompressorControls.h"
+#include "Compressor.h"
```
DomClark, line 29:
```suggestion
#include "CompressorControls.h"

#include <QDomElement>

#include "Compressor.h"
```
Include order again.

plugins/Compressor/Compressor.cpp
```diff
@@ -0,0 +1,652 @@
...
+	connect(&m_compressorControls.m_limiterModel, SIGNAL(dataChanged()), this, SLOT(redrawKnee()));
+	connect(&m_compressorControls.m_mixModel, SIGNAL(dataChanged()), this, SLOT(calcMix()));
+
+	connect(&m_compressorControls.m_autoAttackModel, SIGNAL(dataChanged()), this, SLOT(calcAutoAttack()));
+	connect(&m_compressorControls.m_autoReleaseModel, SIGNAL(dataChanged()), this, SLOT(calcAutoRelease()));
+
+	connect(&m_compressorControls.m_thresholdModel, SIGNAL(dataChanged()), this, SLOT(calcAutoMakeup()));
+	connect(&m_compressorControls.m_ratioModel, SIGNAL(dataChanged()), this, SLOT(calcAutoMakeup()));
+	connect(&m_compressorControls.m_kneeModel, SIGNAL(dataChanged()), this, SLOT(calcAutoMakeup()));
+	connect(&m_compressorControls.m_autoMakeupModel, SIGNAL(dataChanged()), this, SLOT(calcAutoMakeup()));
```
DomClark, line 87: Some of these connections should be direct (any of them which change variables that are used in the processing code).

DomClark (commented):
Not a comprehensive review by any means, but I saw you intended to merge soon, so I wrote up some notes I made when I started looking at this a while ago.

plugins/Compressor/CompressorControlDialog.cpp
DouglasDGI: Oh, I fixed that on my local branch but completely forgot to push the changes.

DouglasDGI:
Merged.  Here's a demo video I made:
https://www.youtube.com/watch?v=4uaCVkrF2u4

Issue #4181, opened 2018-02-21
User cannot see the FX Mixer connections at a glance: Some solutions


raindropsfromsky:
LMMS does not have an easy visualization that shows which instrument is connected with which FX channel; and also which channels are connected with other FX Channels.

I found a couple of implementations (see below)

Personally, I like the **patch-panel matrix** of Ardour

![image](https://user-images.githubusercontent.com/9047168/36465934-9aacbc94-16fd-11e8-920c-89522640a1cf.png)

This implementation is not only clean, but it can convey a lot of information by color-coding the dots and the labels.

The second category is where the connection is shown with a line or curve.

![image](https://user-images.githubusercontent.com/9047168/36466014-f81a0d6e-16fd-11e8-80d8-67eec770ec5a.png)
IMHO this cannot convey the information like the patch matrix can. Probably the matrix is easier to code also!

![image](https://user-images.githubusercontent.com/9047168/36467480-f6a86956-1704-11e8-8225-83e660114a45.png)
This is from OpenOctave Midi.

At minimum, LMMS should display a list of all connections as extended tooltip. 
(When the user hovers his mouse on a channel, show all connections in an extended tooltip.)

musikBear:
> At minimum, LMMS should display a list of all connections as extended tooltip.

Or the user should use *Project-Notes* and write down what he connect, and maby more important, *why*. Im not sure a visualisation is worth that much.

Anonymouqs:
>Or the user should use Project-Notes and write down what he connect, and maby more important, why.

Some artists just *feel*, they like to try stuff out; sometimes, the user might want to see a "summary" instead of reading the actual book. Also, hand-documenting connections is beyond tedious, it is soul-destroying in a complicated project. 

Finally, such visualization has clearly become a standard. If we want to set a standard(by going beyond), we must first *meet* it. 

And every (good) standard has a reason, this is for the ease of the user and allows them to keep their music more organized. I feel Ardour's version would be more intuitive and allows ease of refactoring.

Spekular:
Just as an FYI, FL has a display with "cables" connecting channels:
https://www.image-line.com/support/FLHelp/html/mixer_index.htm

FWIW I'm more of a fan of the line connections (2 lists or FL style) than a
matrix display.

Anonymouqs:
>FWIW I'm more of a fan of the line connections (2 lists or FL style) than a
matrix display.

[...]

Issue #1214, opened 2014-10-15
FX-mixer strip color indicating it's type


unfa:
![mockup](https://cloud.githubusercontent.com/assets/491704/4647361/f75fe52c-547a-11e4-9a2c-3ecb94cb2cd6.png)

The idea is to make the mixer aware of some strip categories, and colour the strips accordingly to make it easier for the user to see what is what.

Rules for colouring:

1. Color as submix (yellow) if the strip has more than 1 instruments or sends assigned;
2. Color as B+B instrument (violet) if has an instrument assigned that belongs to B+B space;
3. Color as Song instrument (green) if has an instrument assigned that belongs to Song space;
4. Color with default if none above criteria is met (unassigned strips, master strip)

Spekular:
@tresf to be consolidated into #1665.

tresf:
This bug report has been consolidated (at least for now) into a placeholder #1665 for better color handling in general.  Feel free to update this bug report or the parent bug as needed.

PR #5465, opened 2020-04-24
Add FX-Mixer channel color differentiation

Veratil:
An idea that popped up in Discord's #production channel, wishing for a similar effect to FL Studio's "wires" that show connected channels.

lmms-mixer-colors

The active (selected) channel is still highlighted "white", but all channels that it sends to are highlighted different than the other channels (in this case a green tint). All channels that it receives from are highlighted different as well (in this case a red tint).

This does slightly change the default view for the master channel though:
lmms-mixer-colors-master

When you're just sending to other channels:
lmms-send-to-highlight

I leave it up to the color deciders to figure out the proper fitting colors to use for this. 😄

### DIFF
--- a/data/themes/default/style.css
+++ b/data/themes/default/style.css
@@ -594,6 +594,8 @@ FxLine {
 	background: #14161A;
 	color: #d1d8e4;
 	qproperty-backgroundActive: #3B424A;
+	qproperty-backgroundSendTo: #304040;
+	qproperty-backgroundReceiveFrom: #3C3237;
 	qproperty-strokeOuterActive: #262B30;
 	qproperty-strokeOuterInactive: #262B30;
 	qproperty-strokeInnerActive: #0C0D0F;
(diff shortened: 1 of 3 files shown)


Spekular:
Relevant: #4181

Potential clash: #1214, user configurable channel colors.

Personally I think the UI is already rather clear about what sends where. Assuming channel x is selected:
- Upwards arrow on a channel means it sends to x
- Enabled SEND button and amount knob means it receives from x

If we want to make this more evident I prefer the cable solution.

Umcaruje:
I have to agree with @Spekular on this one. The colors make it even more confusing to me. We also should leave the space for custom colors on the FX mixer, rather than adding more non-changeable colors. 

What I see as a problem is more clearly distincting the currently selected channel so you know where the sends are going. This is elegantly solved in FL with the virtual cables, which work because their send controls are at the bottom of the mixer. I don't know how this would be implemented in LMMS unless we actually moved the knobs to the bottom.

Veratil:
Sounds good to me! I was just playing around and figured it might be a good first step. 😄

Issue #4201, opened 2018-02-28
Beat editor: clicking on a beat toggles the previous beat on/off


tukkek:
I'm using a fairly large amount of beats on a small window (48 beat/12 bars) and the actual click area of each beat does not correspond to the beat itself as-draw. Instead, it often will toggle the beat to the left of it, even when I'm clicking it *very obviously* on the right beat.

OK just as I'm writing this I think I notice what's wrong. When I move the cursor on to the beat, if goes from the normal system cursor to a "pointing hand" and the actual cursor click does not register on the position of the pointing finger but on the thumb instead, which is a few pixels to the left.

Here's what I mean: even though the index finger is clearly on the right beat, the click registers on the thumb instead (left beat):

![screenshot_20180228_152152](https://user-images.githubusercontent.com/3819826/36805110-48a50e44-1c9b-11e8-86ca-6e541a931fc2.png)

I'm not sure if that's my system's cursor too or something that comes with LMMS. Either way, it gets pretty confusing, especially when working with more beats in less screen space so an alternative would be great.

tukkek:
I have tried resizing the window to get bigger beats and the problem still persists (this is how I took the screenshot above). I can't recall ever having a problem like this in previous versions of LMMS and I've used the beat editor a fair amount of times, as you'd imagine.

tresf:
Yes, this is a mouse position bug but it appears worse on your computer, possibly due to the way the theme is rendering.  Here's the code: https://github.com/LMMS/lmms/blob/stable-1.2/src/tracks/Pattern.cpp#L723

It appears we do some magic math to determine which step was clicked.  I would recommend beginning debugging there.  Any additional information about your computer is useful so that a developer can understand what UI differences would cause this, i.e. Anything that influences the spacing or positioning of items.

PhysSong:
Reproduced. I guess this is related with cursor shapes. If the coordinate information in `QMouseEvent` doesn't point to the end of index finger, that could be a issue.

zonkmachine:
> If the coordinate information in `QMouseEvent` doesn't point to the end of index finger, that could be an issue.

[...]

PR #5466, opened 2020-04-25
BBEditor - Select step more precisely

zonkmachine:
Fix https://github.com/LMMS/lmms/issues/4201

Tweak the algorithm for selecting out which beat step is checked with the mouse cursor. This attempt adds both a fixed value and a value depending on the number of steps in the pattern.

### DIFF
--- a/src/tracks/Pattern.cpp
+++ b/src/tracks/Pattern.cpp
@@ -722,7 +722,8 @@ void PatternView::mousePressEvent( QMouseEvent * _me )
 //	get the step number that was clicked on and
 //	do calculations in floats to prevent rounding errors...
 		float tmp = ( ( float(_me->x()) - TCO_BORDER_WIDTH ) *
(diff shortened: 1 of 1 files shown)


zonkmachine:
On master, stable-1.2 and this branch if I click on the right margin, just when you get a pointing hand icon to indicate you're over the last step, I get this message:

`Something went wrong in pattern.cpp: step 16 doesn't exist in pattern!`

1. It should probably indicate step 17 for a human reader as it's counting from 0.
2. It shouldn't show in the first place so I'll try and tweak the code some more

zonkmachine:
Currently trying to fix the algorithm instead of tweaking it after computation.

zonkmachine:
Fix wrong as described here: https://github.com/LMMS/lmms/issues/4201#issuecomment-620224098 , closing.

PR #5467, opened 2020-04-26
DrumSynth: ensure correct envelope length in any sample rate

### LINKED ISSUE #5342: .ds files envelopes shortened by oversampling
> It looks like this is still an issue on drumsynth (.ds) files.
> https://lmms.io/forum/viewtopic.php?f=7&t=31122

_Originally posted by @zonkmachine in https://github.com/LMMS/lmms/pull/5226#issuecomment-555184114_

So, what happens is that the pitch is correct but the envelopes are too short. Looks like the unit for envelope times in the .ds file is samples, which breaks when a sample rate != 44100 is passed in. 

The fix is probably as simple as a one-liner adjusting the ```timestretch``` variable ( https://github.com/LMMS/lmms/blob/master/src/core/DrumSynth.cpp#L65 ) when loading the .ds file.

PhysSong:
Fixes #5342.


zonkmachine:
Code reviewed and tested. Quick merge.

PR #5469, opened 2020-04-26
Set AutomationPattern length to 1 bar if the length is 0

### LINKED ISSUE #5254: Automation scroll limited by song editor
If you create a one block segment for an automation track in the song editor, then open and edit that segment.  Then you zoom out along the x axis, draw a long pattern, then zoom back in... you are prevented from scrolling to the right to see the rest of the pattern you drew.

It seems that the range of view along the x axis is determined by the length of the segment in the song editor.

The left side of the window seems to always stop at the 2nd measure, and the right side depends on the amount of zoom.  In the song editor, if you change the size of the segment to take up 4 blocks, then in Automation Editor, the scroll will stop at the 5th measure.  So, you can still edit beyond the range of the segment, but the left side of the window won't go past the end of the segment's range.

automationScrollIssue

May possibly affect:
#4804 
#4563

Veratil:
Fixes #5254 

~~AutomationPattern's length is now always updated (not only if it's a HiddenAutomationTrack which we're removing eventually). Just changing that though resulted in the length of the block in the SongEditor to be 0 (or a single pixel), so this also makes the minimum length of the block 1 bar.~~

~~Tested playing around with adding and removing blocks from BB tracks, Automation tracks, and Piano tracks, all seemed to still behave properly.~~

~~EDIT: I totally forgot that there were actually tests for the automation track length. I'm not sure how we want to change these.~~

If the last bar and last tick is 0, then set the length to 1 bar.

### DIFF
--- a/src/core/AutomationPattern.cpp
+++ b/src/core/AutomationPattern.cpp
@@ -179,18 +179,15 @@ const AutomationPattern::objectVector& AutomationPattern::objects() const
 MidiTime AutomationPattern::timeMapLength() const
 {
 	MidiTime one_bar = MidiTime(1, 0);
-	if (m_timeMap.isEmpty())
-	{
-		return one_bar;
-	}
+	if (m_timeMap.isEmpty()) { return one_bar; }
+
 	timeMap::const_iterator it = m_timeMap.end();
 	tick_t last_tick = static_cast<tick_t>((it-1).key());
-	bar_t last_bar = qMax(0, MidiTime(last_tick).nextFullBar() - 1);
-	if (last_bar == 0 && last_tick == 0)
-	{
-		return one_bar;
-	}
-	return MidiTime(last_bar, last_tick);
+	// if last_tick is 0 (single item at tick 0)
+	// return length as a whole bar to prevent disappearing TCO
+	if (last_tick == 0) { return one_bar; }
+
+	return MidiTime(last_tick);
 }
 
 
@@ -198,7 +195,8 @@ MidiTime AutomationPattern::timeMapLength() const
 
 void AutomationPattern::updateLength()
 {
-	changeLength( timeMapLength() );
+	// Do not resize down in case user manually extended up
+	changeLength(qMax(length(), timeMapLength()));
 }
 
 
--- a/src/core/AutomationPattern.cpp
+++ b/src/core/AutomationPattern.cpp
@@ -178,9 +178,19 @@ const AutomationPattern::objectVector& AutomationPattern::objects() const
 
 MidiTime AutomationPattern::timeMapLength() const
 {
-	if( m_timeMap.isEmpty() ) return 0;
+	MidiTime one_bar = MidiTime(1, 0);
+	if (m_timeMap.isEmpty())
+	{
+		return one_bar;
+	}
 	timeMap::const_iterator it = m_timeMap.end();
-	return MidiTime( MidiTime( (it-1).key() ).nextFullBar(), 0 );
+	tick_t last_tick = static_cast<tick_t>((it-1).key());
+	bar_t last_bar = qMax(0, MidiTime(last_tick).nextFullBar() - 1);
+	if (last_bar == 0 && last_tick == 0)
+	{
+		return one_bar;
+	}
+	return MidiTime(last_bar, last_tick);
 }
 
 
@@ -223,12 +233,7 @@ MidiTime AutomationPattern::putValue( const MidiTime & time,
 	}
 	generateTangents( it, 3 );
 
-	// we need to maximize our length in case we're part of a hidden
-	// automation track as the user can't resize this pattern
-	if( getTrack() && getTrack()->type() == Track::HiddenAutomationTrack )
-	{
-		updateLength();
-	}
+	updateLength();
 
 	emit dataChanged();
 
@@ -251,10 +256,7 @@ void AutomationPattern::removeValue( const MidiTime & time )
 	}
 	generateTangents(it, 3);
 
-	if( getTrack() && getTrack()->type() == Track::HiddenAutomationTrack )
-	{
-		updateLength();
-	}
+	updateLength();
 
 	emit dataChanged();
 }
--- a/src/core/Track.cpp
+++ b/src/core/Track.cpp
@@ -159,7 +159,7 @@ void TrackContentObject::movePosition( const MidiTime & pos )
 
 /*! \brief Change the length of this TrackContentObject
  *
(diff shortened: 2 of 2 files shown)


Spekular:
> this also makes the minimum length of the block 1 bar.

This breaks some of the functionality #4973 introduced. Patterns shorter than a bar are valid. They're especially useful for sample tracks, but I think they should be available in all track types.

Proposed alternatives:
- Only update when the pattern would grow
- Only update when adding a node to the right of all other nodes
- Only update if at least one node with position > 0 exists
- Apply snapping to the automatic length change? I know I had to add some checks to avoid zero length clips, so perhaps the snapping code would handle this case.

src/core/Track.cpp
```diff
@@ -159,14 +159,16 @@ void TrackContentObject::movePosition( const MidiTime & pos )
...
  *  also add a journal entry for undo and update the display.
  *
  * \param _length The new length of the track content object.
  */
 void TrackContentObject::changeLength( const MidiTime & length )
 {
-	m_length = length;
+	// use a minimum of 1 bar always, otherwise it's possible to get 0 length TCOs
+	// which will make you think it has disappeared
+	m_length = qMax(MidiTime(1, 0), length);
```
Spekular, line 171: As mentioned in comment, clips shorter than a bar are valid and shouldn't be prevented.

Veratil:
I didn't realize it wasn't restricted!

I'll rework and figure out a better solution. 👍

Veratil:
Update OP.

This should fix the issue, while still allowing any length TCO as @Spekular pointed out.


> This breaks some of the functionality #4973 introduced. Patterns shorter than a bar are valid. They're especially useful for sample tracks, but I think they should be available in all track types.

I do notice you can't resize PianoRoll TCOs, but all the others can resize the same (even Automation). 🙂

src/core/AutomationPattern.cpp
```diff
@@ -178,9 +178,19 @@ const AutomationPattern::objectVector& AutomationPattern::objects() const
...
+		return one_bar;
+	}
 	timeMap::const_iterator it = m_timeMap.end();
-	return MidiTime( MidiTime( (it-1).key() ).nextFullBar(), 0 );
+	tick_t last_tick = static_cast<tick_t>((it-1).key());
+	bar_t last_bar = qMax(0, MidiTime(last_tick).nextFullBar() - 1);
+	if (last_bar == 0 && last_tick == 0)
+	{
+		return one_bar;
+	}
```
Spekular, line 192: I think you could write:
```suggestion
	tick_t last_tick = static_cast<tick_t>((it-1).key());
	if (last_tick == 0)
	{
		return one_bar;
	}
	bar_t last_bar = MidiTime(last_tick).nextFullBar() - 1;
```
Ticks must be absolute for `MidiTime(last_tick).nextFullBar()` to return the right bar, so you may as well just check if we're at zero ticks and skip checking `last_bar == 0`.

This lets us short circuit a little faster and makes the qMax call redundant (`nextFullBar` can't be less than zero if `last_tick >= 1`

src/core/AutomationPattern.cpp
```diff
@@ -178,9 +178,19 @@ const AutomationPattern::objectVector& AutomationPattern::objects() const
 
 MidiTime AutomationPattern::timeMapLength() const
 {
-	if( m_timeMap.isEmpty() ) return 0;
+	MidiTime one_bar = MidiTime(1, 0);
+	if (m_timeMap.isEmpty())
+	{
+		return one_bar;
+	}
```
Spekular, line 185: I think this is a great candidate to be on one line:
```suggestion
	if (m_timeMap.isEmpty()) { return one_bar; }
```
Same for the other if. I would also drop the `one_bar` variable and directly return `MidiTime(1, 0)` in both cases. Reasoning:

- In some calls the value is never used.
- It's pretty reasonable to guess that MidiTime(1, 0) is one bar, or check if you really want to know. Even without knowing the specific length, it's clear that a placeholder value is returned for empty maps.
- Dropping the line keeps `timeMapLength` a little shorter. Alternatively, replacing it with a comment keeps the same length but could provide more insight. Hence you could argue the variable is the "worst of both worlds", haha.
  - `//Return one bar if length would otherwise be zero, to avoid invisible TCOs`

Veratil: Ah yeah, those can be one line!

As for `one_bar`, I was thinking more along DRY standard there.

Spekular:
I was hoping that leaving a "comment" review would get rid of the merge-blocking "changes requested" state without explicitly approving the changes, but that didn't work. I've approved them now because I think you should be able to merge this if you want to, but I do have another round of proposed changes :)

src/core/AutomationPattern.cpp
Veratil: Sometimes one overthinks the solution. Good catch. 👍

Veratil:
Actually, I do have a question for the class now: this changes the behavior of Automation tracks.

Let's say I do this:
1. Resize the track to 15 bars
2. Put a (last) tick on bar 6

This resizes the length down to 6 bars. Do we want this behavior? Should we never resize _down_?

src/core/AutomationPattern.cpp
Spekular: Yeah, I'm aware of DRY, I just don't think this is enough repetition to count. It's fine with me if you leave it though, this was my least important comment.

Spekular:
> Should we never resize _down_?

I think this is a pretty sane solution. Does it work if you change line 201 in AutomationPattern.cpp as below?
```diff
-changeLength( timeMapLength() );
+changeLength( qMax(timeMapLength(), length()) );
```

Veratil:
> 
> 
> > Should we never resize _down_?
> 
> I think this is a pretty sane solution. Does it work if you change line 201 in AutomationPattern.cpp as below?
> 
> ```diff
> -changeLength( timeMapLength() );
> +changeLength( qMax(timeMapLength(), length()) );
> ```

It doesn't seem to work. It looks like if you extend the visual track, the internal length is still only the default `192` (or whatever the latest change through `updateLength` is).

Spekular:
That's odd, resizing should call changeLength...

https://github.com/LMMS/lmms/blob/24194334e65b7fb508fad8e03149866bde5d732f/src/core/Track.cpp#L948

Veratil:
Disregard previous comment, apparently had old compiled version running. That does indeed work!

Spekular:
Tested and works perfectly for me:
- Doesn't create zero length clips
- Never shrinks clips

Issue #5462, opened 2020-04-22
Possibility of processing input MIDI messages with their original channel


IanCaio:
First of all, I would like to thank all the effort put into LMMS!

_I wasn't completely sure whether to post this as a bug report or feature request, because I don't know if the current behavior is intentional. In doubt I'm posting it as a feature request._

Suppose you have a track "A" and enable MIDI Input on that track without selecting a channel ("--", meaning all channels are accepted). Then you have two other tracks, "B" and "C". "B" has MIDI Output enabled with channel 1 and "C" has MIDI Output enabled with channel 2. Both route to "A". When "A" sends the MIDI events to its plugin (i.e.: Carla-Patchbay), those channel informations from the MIDI events of "B" and "C" are lost, because all messages from "A" will have a single channel, which is the one selected on the MIDI Output Channel from the Instrument window MIDI section. That happens even if the MIDI Output from "A" is not enabled.

A LMMS forum post that might give some context to the issue:
https://lmms.io/forum/viewtopic.php?f=7&t=32622

### Enhancement Summary

- Add the possibility of forwarding the MIDI messages received from the input MIDI ports keeping their original channel.

#### Justification

It would open lots of possibilities if we were able to maintain the channel information and send those messages received on the input ports with their original channels. On Carla for example, you could use the channels to select different patches on a synthesizer. Or even use one channel for automation and another one for sending the notes. At first glance it sounds like a possibly simple change that would only contribute positively for the functionality of LMMS.

I wish I had experience with C++ and Qt to be of more help, but I'll try to hunt through the code and, who knows, even try some changes here anyways. From my research today I've seen that `src/core/NotePlayHandle.cpp`, `src/tracks/InstrumentTrack.cpp` and `src/core/midi/MidiPort.cpp` are key to the forwarding of those MIDI messages, particularly the `processOutEvent` method.

This issue could be added to placeholder #1472 if people agree with it!

PR #5470, opened 2020-04-26
Allows instruments to keep the midi channel information when forwarding

IanCaio:
Implemented the feature discussed on issue #5462 .

### Without this PR, LMMS handles midi I/O the following way:

- Input can be enabled to allow an instrument to receive midi events. A channel can be selected (so only that channel is received) or none can be selected ("--", which allows all channels to be received).
- Output can be enabled to forward the midi events processed on the track to other channels. A channel from 1-16 can be selected, so only the events with that channel are forwarded.

There's another behavior related to the midi output channel that is less known:

- The midi events from the track itself (and all others received from other tracks) are sent to the instrument plugin using the midi output channel, being it enabled or not.

That behavior makes impossible to send midi events with different channels to an instrument plugin, since they will be all converted to have the selected output channel. An example of a situation where this could be an issue is if we decide to use a different channel to signal the instrument to use another preset. Another example: Suppose we make two tracks with different channels, where the first have the notes to be played and the second has notes that are going to be interpreted as automation values. We could forward those to a Carla-patchbay and automate parameters using the key values from the second track, differentiating each one using the channel. That would be a nice workaround for the lack of Carla automation we currently have (I managed to do it here with a build from my code and can share projects showing how to do it). Finally, some MIDI keyboards might support channel mapping and this feature could make use of that.

### With this PR the behavior is now the following:

- Midi input behavior stays the same
- We can now select a midi output channel from 1 to 16 (to filter the channels forwarded, exactly like before) or select no channel ("--"). If no channel is selected, all received midi events are forwarded keeping their original channels. The events generated by the instrument itself will be sent with the default channel (I've chosen 1). There will be no filtering, all events will be forwarded.
- Since the midi output channel is now relevant even when we have the midi output disabled (since the midi events will be sent to the instrument plugin and the channel used depends on it) the spin box widget that selects the midi output channel is enabled even when you have midi output disabled so you can change channels.

### Changes to the code:

- The `outputChannel()` from the class _MidiPort_ could only have values ranging from 1 to 16. Now it can have values from 0 to 16, 0 being a value representing "no channel".
- The `realOutputChannel()` will return `outputChannel() - 1` like before if the value of the channel is inside the old range, or it will return `1` if the value is 0. That would guarantee backward compatibility because the values will be inside the expected range and from my research `outputChannel()` is not used to retrieve the channel by any plugin, only `realOutputChannel()`.
- Created local variable `handleEventOutputChannel` in method `InstrumentTrack::processOutEvent` that will hold the channel used to handle midi events (channel sent to instrument plugins). Its value is decided based on the behavior described before (1-16 if output channel is selected, or `event.channel()` if not).
- Changed method `MidiPort::processOutEvent` to forward midi events from all channels when midi output channel "--" is selected.
- Changed code on the `InstrumentMidiIOView` widget to make the midi output channel spin box display "--" when the output channel value is 0 and to keep it to becoming disabled when midi output is disabled.

### DIFF
--- a/src/core/midi/MidiPort.cpp
+++ b/src/core/midi/MidiPort.cpp
@@ -151,36 +151,19 @@ void MidiPort::processInEvent( const MidiEvent& event, const MidiTime& time )
 
 void MidiPort::processOutEvent( const MidiEvent& event, const MidiTime& time )
 {
-	// If we selected a channel from 1-16, we will only route the selected channel.
-	// But if we selected channel 0 ("--") we will route all channels out.
-	if( isOutputEnabled() )
+	// When output is enabled, route midi events if the selected channel matches
+	// the event channel or if there's no selected channel (value 0, represented by "--")
+	if( isOutputEnabled() && ( outputChannel() == 0 || realOutputChannel() == event.channel() ) )
 	{
-		// Selected channel 0 ("--"). Route all channels
-		if( outputChannel() == 0 )
+		MidiEvent outEvent = event;
+
+		if( fixedOutputVelocity() >= 0 && event.velocity() > 0 &&
+			( event.type() == MidiNoteOn || event.type() == MidiKeyPressure ) )
 		{
-			MidiEvent outEvent = event;
-
-			if( fixedOutputVelocity() >= 0 && event.velocity() > 0 &&
-				( event.type() == MidiNoteOn || event.type() == MidiKeyPressure ) )
-			{
-				outEvent.setVelocity( fixedOutputVelocity() );
-			}
-
-			m_midiClient->processOutEvent( outEvent, time, this );
+			outEvent.setVelocity( fixedOutputVelocity() );
 		}
-		// Selected channel between 1-16. Mask event.
-		else if( realOutputChannel() == event.channel() )
-		{
-			MidiEvent outEvent = event;
 
-			if( fixedOutputVelocity() >= 0 && event.velocity() > 0 &&
-				( event.type() == MidiNoteOn || event.type() == MidiKeyPressure ) )
-			{
-				outEvent.setVelocity( fixedOutputVelocity() );
-			}
-
-			m_midiClient->processOutEvent( outEvent, time, this );
-		}
+		m_midiClient->processOutEvent( outEvent, time, this );
 	}
 }
 
--- a/src/core/midi/MidiPort.cpp
+++ b/src/core/midi/MidiPort.cpp
@@ -47,7 +47,7 @@ MidiPort::MidiPort( const QString& name,
 	m_midiEventProcessor( eventProcessor ),
 	m_mode( mode ),
 	m_inputChannelModel( 0, 0, MidiChannelCount, this, tr( "Input channel" ) ),
-	m_outputChannelModel( 1, 1, MidiChannelCount, this, tr( "Output channel" ) ),
+	m_outputChannelModel( 1, 0, MidiChannelCount, this, tr( "Output channel" ) ),
 	m_inputControllerModel( 0, 0, MidiControllerCount, this, tr( "Input controller" ) ),
 	m_outputControllerModel( 0, 0, MidiControllerCount, this, tr( "Output controller" ) ),
 	m_fixedInputVelocityModel( -1, -1, MidiMaxVelocity, this, tr( "Fixed input velocity" ) ),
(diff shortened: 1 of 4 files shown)


src/core/midi/MidiPort.cpp
```diff
@@ -151,18 +151,36 @@ void MidiPort::processInEvent( const MidiEvent& event, const MidiTime& time )
...
+			if( fixedOutputVelocity() >= 0 && event.velocity() > 0 &&
+				( event.type() == MidiNoteOn || event.type() == MidiKeyPressure ) )
+			{
+				outEvent.setVelocity( fixedOutputVelocity() );
+			}
+
+			m_midiClient->processOutEvent( outEvent, time, this );
 		}
+		// Selected channel between 1-16. Mask event.
+		else if( realOutputChannel() == event.channel() )
```
Veratil, line 172: I don't see a difference between this block and the `outputChannel() == 0` one.

src/gui/widgets/InstrumentMidiIOView.cpp
```diff
@@ -84,8 +84,9 @@ InstrumentMidiIOView::InstrumentMidiIOView( QWidget* parent ) :
 	midiOutputLayout->setSpacing( 6 );
 
 	m_outputChannelSpinBox = new LcdSpinBox( 2, m_midiOutputGroupBox );
+	m_outputChannelSpinBox->addTextForValue( 0, "--" );
 	m_outputChannelSpinBox->setLabel( tr( "CHANNEL" ) );
-	m_outputChannelSpinBox->setEnabled( false );
+	m_outputChannelSpinBox->setEnabled( true );
```
Veratil, line 89: What if we don't want to enable midi output? This will always be enabled then.

src/gui/widgets/InstrumentMidiIOView.cpp
```diff
@@ -108,8 +109,6 @@ InstrumentMidiIOView::InstrumentMidiIOView( QWidget* parent ) :
 	midiOutputLayout->addWidget( m_fixedOutputNoteSpinBox );
 	midiOutputLayout->addStretch();
 
-	connect( m_midiOutputGroupBox->ledButton(), SIGNAL( toggled( bool ) ),
-			m_outputChannelSpinBox, SLOT( setEnabled( bool ) ) );
```
Veratil, line 112: As above, what if we don't want to enable midi output?

src/core/midi/MidiPort.cpp
IanCaio: That's true, should have written a single conditional `if( outputChannel() == 0 || realOutputChannel() == event.channel() )` instead, my bad. I'll change that!

src/gui/widgets/InstrumentMidiIOView.cpp
IanCaio: That only enables the midi output channel spin box widget, not the midi output itself. That's so you're able to choose the channel that the track will use to send the midi events to the instrument plugin without needing to enable midi output.

src/gui/widgets/InstrumentMidiIOView.cpp
IanCaio: Same as above, this makes the midi output channel spin box widget independent from the led button that enables/disables the midi output. That's just so you can more conveniently change the channel without needing to enable the midi output.

qnebra:
It changes in any way lmms workflow with midi? Because in some cases, like a few midi controlled parameters assigned to few midi inputs, even from same controller device, it get really annoying. 

I asked, because there are situations where using lmms can be pure nightmare, and one of them is when user record automation using midi controller and wants to preview it as it was done.

IanCaio:
> It changes in any way lmms workflow with midi? Because in some cases, like a few midi controlled parameters assigned to few midi inputs, even from same controller device, it get really annoying.
> 
> I asked, because there are situations where using lmms can be pure nightmare, and one of them is when user record automation using midi controller and wants to preview it as it was done.

It doesn't, the Midi I/O behavior stays the same when the previously allowed channels are selected. It only changes when Midi Output channel 0 (represented by `--`) is selected, which was not allowed before, so it should be backwards compatible. I tried to write this PR adding the feature while messing the least I could with the existent functionality. There are more details regarding the functions that were changed at the bottom of the PR description, if you are curious about what was changed internally!

src/gui/widgets/InstrumentMidiIOView.cpp
DomClark: Aren't `LcdSpinBox`es enabled by default? In which case, this line could be removed altogether.

src/tracks/InstrumentTrack.cpp
```diff
@@ -393,6 +393,14 @@ void InstrumentTrack::processOutEvent( const MidiEvent& event, const MidiTime& t
 	const MidiEvent transposedEvent = applyMasterKey( event );
 	const int key = transposedEvent.key();
 
+	// If we have a selected output midi channel between 1-16, we will use that channel to handle the midi event.
+	// But if our selected midi output channel is 0 ("--"), we will use the event channel instead.
+	auto handleEventOutputChannel = midiPort()->realOutputChannel();
+	if( midiPort()->outputChannel() == 0 )
+	{
+		handleEventOutputChannel = event.channel();
+	}
```
DomClark, line 402: Can you use a ternary operator here instead? I think it would be easier to read, and it means you can make `handleEventOutputChannel` const.

src/gui/widgets/InstrumentMidiIOView.cpp
IanCaio: You're right @DomClark , I was making the requested changes and I'm about to update the PR.

src/tracks/InstrumentTrack.cpp
IanCaio: Sure! I agree, I think it improved readability. Just rebasing my repo and I'll update the PR!

PhysSong:
The code looks good. Are there anyone who wants to review this more?

ToruTheRedFox:
as of 1.2.2 I still can't select 0 as the output channel, only input

DomClark:
This feature isn't available in the 1.2 series. You can download an 1.3 alpha version from our website if you want to try it out.

ToruTheRedFox:
ah alright

PR #5472, opened 2020-04-27
AFP: fix loading start and loop points if the loop point is automated

### LINKED ISSUE #5063: Automation issues with Audio File Processor's start and loop points
When automating the starting point and the loop point of a sample in the Audio File Processor at the same time, only the loop point connection will be saved. If both are linked to the same automation pattern, only the loop point connection will be preserved. If they are linked to distinct patterns, the whole situation becomes even weirder. The pattern linked to the starting point of the sample turns its connection into the loop point, and the loop point pattern loses its connection completely.

## Example Projects
bug.mmp - both were orginally linked to the same pattern
bug-01.mmp - only the start point was linked to the pattern (works as expected)
bug-02.mmp - start, loop and end point were linked. The start point's connection is lost
bug-03.mmp - only start and end, but not the loop point were automated. Works as expected.
bug-04.mmp - start and loop point were originally connected to the automation tracks of the same name. As described above.
[bug.zip](https://github.com/LMMS/lmms/files/3341715/bug.zip)

Also, on an unrelated note, the automation itself is a little wonky it seems. If you listen to bug-01 a little longer you'll notice there will sometimes be multiple snares in a row. It looks like most of the time the automation is applied before the sample is played, but also sometimes shortly after, so the sample starts at the old position. The chances for either one to happen seem to be random. Please tell me if this is worth its own bug report or not.

System Info: Win10 10.0.18362.175 LMMS 1.2.0 x64

PhysSong:
Fixes #5063.

### DIFF
--- a/plugins/audio_file_processor/audio_file_processor.cpp
+++ b/plugins/audio_file_processor/audio_file_processor.cpp
@@ -253,17 +253,16 @@ void audioFileProcessor::loadSettings( const QDomElement & _this )
 	m_loopModel.loadSettings( _this, "looped" );
 	m_ampModel.loadSettings( _this, "amp" );
 	m_endPointModel.loadSettings( _this, "eframe" );
+	m_startPointModel.loadSettings( _this, "sframe" );
 
 	// compat code for not having a separate loopback point
-	if( _this.hasAttribute( "lframe" ) )
+	if (_this.hasAttribute("lframe") || !(_this.firstChildElement("lframe").isNull()))
 	{
 		m_loopPointModel.loadSettings( _this, "lframe" );
-		m_startPointModel.loadSettings( _this, "sframe" );
 	}
 	else
 	{
 		m_loopPointModel.loadSettings( _this, "sframe" );
-		m_startPointModel.setValue( m_loopPointModel.value() );
 	}
 
 	m_reverseModel.loadSettings( _this, "reversed" );


plugins/audio_file_processor/audio_file_processor.cpp
```diff
@@ -253,17 +253,16 @@ void audioFileProcessor::loadSettings( const QDomElement & _this )
 	m_loopModel.loadSettings( _this, "looped" );
 	m_ampModel.loadSettings( _this, "amp" );
 	m_endPointModel.loadSettings( _this, "eframe" );
+	m_startPointModel.loadSettings( _this, "sframe" );
 
 	// compat code for not having a separate loopback point
-	if( _this.hasAttribute( "lframe" ) )
+	if (_this.hasAttribute("lframe") || !(_this.firstChildElement("lframe").isNull()))
```
Spekular, line 259: If only loadSettings returned errors this could be something like:
```
if ( m_loopPointModel.loadSettings( _this, "lframe" ) {}
else { m_loopPointModel.loadSettings( _this, "sframe" ); }
```

PhysSong: You can simplify it more in that case.
```cpp
if (!m_loopPointModel.loadSettings(_this, "lframe"))
{
	m_loopPointModel.loadSettings(_this, "sframe");
}
```

Spekular:
Bug reproduced on stable-1.2 build + fix confirmed. Merge?

PR #5475, opened 2020-04-29
Fix #5461 and ensure consistent use of check gate

### LINKED ISSUE #5461: CrossoverEQ: Invalid checkGate function
### Bug Summary

From plugins/CrossoverEQ/CrossoverEQ.cpp:
```
const float d = dryLevel();
const float w = wetLevel();
double outSum = 0.0;
for( int f = 0; f < frames; ++f )
{
	outSum = buf[f][0] * buf[f][0] + buf[f][1] * buf[f][1];
	buf[f][0] = d * buf[f][0] + w * m_work[f][0];
	buf[f][1] = d * buf[f][1] + w * m_work[f][1];
}
	
checkGate( outSum );
```

There are at least 2 bugs in this calculation:

1. It seems to calculate outSum incorrectly - it's assigned rather than incremented in the loop, so only the last sample is actually checked. (Thanks to @DomClark for reporting)
2. `outSum` must be divided by `frames` in the end. (caused by #5460)

#### Steps to reproduce

Unclear

#### Expected behavior

Unclear

#### Actual behavior

Unclear. It could lead to too early or to late stopping of effect processing, and thus to clipping.

#### Affected LMMS versions

all

thmueller64:
This PR addresses #5461 and ensures that `checkGate` is called on the processed buffer for several other builtin plugins. In the file "plugins/Eq/EqEffect.cpp", `checkGate` is also used incorrectly, however, fixing the problem is not as straightforward as for the other plugins and will be addressed in a separate PR.

### DIFF
--- a/plugins/Amplifier/Amplifier.cpp
+++ b/plugins/Amplifier/Amplifier.cpp
@@ -83,7 +83,6 @@ bool AmplifierEffect::processAudioBuffer( sampleFrame* buf, const fpp_t frames )
 	for( fpp_t f = 0; f < frames; ++f )
 	{
 //		qDebug( "offset %d, value %f", f, m_ampControls.m_volumeModel.value( f ) );
-		outSum += buf[f][0]*buf[f][0] + buf[f][1]*buf[f][1];
 	
 		sample_t s[2] = { buf[f][0], buf[f][1] };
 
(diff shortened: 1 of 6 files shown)


JohannesLorenz:
Code and style review finished.

I'll still do some simple tests.

JohannesLorenz:
Tests passed, all effects still work.

The last question is if this should be fixed in stable-1.2 . I think it won't damage anything. @DomClark what's your opinion?

DomClark:
I think this is certainly safe enough to add to `stable-1.2`. The only question is backwards compatibility with existing projects. However, this is something we're typically not very strict about, and I think the impact would be minor (there's also the workaround of just using the previous effect's gate and decay if you want to keep the input gate behaviour), so I'm fine for it to go in `stable-1.2` if there aren't any other objections.

JohannesLorenz:
OK, rebasing this automatically with the PR's "edit" button did not work. Or at least, it looked really bad :smiley: 

@thmueller64 Can you please run a `git rebase stable-1.2` on your PR and do a force push? Then we can merge it.

PR #5478, opened 2020-04-30
Select piano key from mouse position

zonkmachine:
When selecting a Piano Key to mark semi tones in the Piano Roll we
select key from the y position of the pop-up menu and not the mouse.
Incidentally these two are most often the same as the menu builds
from the mouse y positon and down. If there is room for it. If there
is no room downwords it will create the menu so the lower part of the
frame aligns with the mouse y position.

Fixed by creating a variable to hold the pressed key before creating
the menu.

keySelect

In the case given above the pressed key will not be C4 but C#5

Bug replicated on lmms-1.0.3

### DIFF
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -507,7 +507,7 @@ void PianoRoll::changeNoteEditMode( int i )
 
 void PianoRoll::markSemiTone( int i )
 {
-	const int key = getKey( mapFromGlobal( m_semiToneMarkerMenu->pos() ).y() );
+	const int key = m_pianoKeySelected;
 	const InstrumentFunctionNoteStacking::Chord * chord = nullptr;
 
 	switch( static_cast<SemiToneMarkerAction>( i ) )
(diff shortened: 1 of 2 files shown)


src/gui/editors/PianoRoll.cpp
```diff
@@ -1704,6 +1704,7 @@ void PianoRoll::mousePressEvent(QMouseEvent * me )
 			if( me->buttons() == Qt::RightButton )
 			{
 				// right click, tone marker contextual menu
+				m_pianoKeySelected = getKey( me->y() );
```
Veratil, line 1707:
```suggestion
				m_pianoKeySelected = getKey(me->y());
```

zonkmachine: Yes but the surrounding code is the older style. It doesn't look good when we mix the styles. I'm waiting for some automatic code formatting to come true. Cool?

zonkmachine:
Merge?

Must remember to change 'downwords' to 'downward' in the commit message.

Veratil:
> Must remember to change 'downwords' to 'downward' in the commit message.

Just do an amend commit and update the message, then push with `--force-with-lease`.

zonkmachine:
Done did it, thanks!

Issue #4602, opened 2018-09-19
The selection tool does not delete clips in groups with the delete key


Gabrielxd195:
Hi guys, how are you ?. as I said in other threads. I have been using lmms lately and while I have used it, I have found some minimal errors. (apparently irrelevant) but that make it more difficult to use in lmms.

Some of these errors have the selection tools.
![selector no borra en grupo modificada](https://user-images.githubusercontent.com/26127693/45774046-fe36ab00-bc19-11e8-85f3-624bf6b63ab0.png)

Both the selection tool, as the selection in all lmms and the selection by pressing '' Ctrl ''. They have errors such as:

1- Multiple clips can not be deleted together with the '' delete '' key. then I have to delete them individually with the mouse wheel.
2- When selecting a piano roll, bb editor or sample track clip, you can not select the desired parts of the clip, which is to select the entire clip.
3- Does not allow individual or group selection in the file browser, I mean, you can not select anything there.
4- You can not select the tracks in the song editor or in the bb editor, which means that you have to delete them from the options menu.
5- A free selection tool is required. The current selection is square or rectangular. the selection is not accurate.
6- In the bb editor it is impossible to select notes in groups, the free selection would be very useful.
7- The piano roll requires a multiple selection tool. here would be quite useful for piano notes
8- The option of inverted selection in all the editors is needed.
9- In the mixer, in '' add effect '', in the chain of effects and in the tray of controllers, there is no selection tool, nor can they be selected in groups by pressing '' Ctrl '', that is, it is impossible the selection. selection in those parts is very necessary, since in some cases it is necessary to eliminate or use many effects.
10- In the automation editor there is no type of selection tool, nor normal, inverted or multiple. Here would be enough to move, copy or delete automation points.

So far are the problems I've seen both the selection tool and the selection in general, maybe they are minimal errors, but they are very annoying and waste time. it does not always work by pressing '' Ctrl ''. Greetings.

musikBear:
@Gabrielxd195  First - Are you on 1.1.3?
Please update to **1.2.0 RC6** (orange buttons) https://lmms.io/download/#windows
Then retry your fail-collection. block-deletes are fixed in RC, so many of your fails are fixed already

Gabrielxd195:
no friend, I am in version 1.2.0 rc6, what happens is that you can not see the title bar. the selection errors are still valid This photo is from lmms is from the latest version, and that is why I put my suggestions.

musikBear:
[...]

Issue #4507, opened 2018-07-29
'Play selection only' / invert-selected


musikBear:
Suggetion
It could be useful, if it was possible to replay only selected notes, in piano roll. Meaning that un-selected notes passed by the playhead, would be mute.
This feature would also be possible, if lmms had an 'invert-selected' feature, meaning that selected notes would turn un-selected, and un-selected would turn selected. Maby as easy as to switch a boolean(?)

sTiKyt:
Yea, i though same xD

PR #5480, opened 2020-05-01
Piano Roll - invert selection

zonkmachine:
Proof of concept (working).

In response to: https://github.com/LMMS/lmms/issues/4507 https://github.com/LMMS/lmms/issues/4602

With the key command `<Ctrl> + <Shift> + i` you invert the selection. If no notes are selected you select all notes.

Things to do:

* Need to study the code in [void PianoRoll::selectAll()](https://github.com/LMMS/lmms/blob/c755b56a52145a572247303a09d126e764238c11/src/gui/editors/PianoRoll.cpp#L3930). It's vastly different from this PR which is a bit worrying. This is a dumb first attempt but it's working,
* Better key combination? Just `<Ctrl> + i`?
* Should there be a gui knob to activate this?
  * Icon needed
* I think I have a redundant `update()` in there.

### DIFF
--- a/include/PianoRoll.h
+++ b/include/PianoRoll.h
@@ -296,6 +296,7 @@ private:
 	void shiftSemiTone(int amount);
 	bool isSelection() const;
 	int selectionCount() const;
+	void invertSelection() const;
 	void testPlayNote( Note * n );
 	void testPlayKey( int _key, int _vol, int _pan );
 	void pauseTestNotes(bool pause = true );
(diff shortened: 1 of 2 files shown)


Veratil:
Ideally this invert should be a part of the same selection function. We should probably look into doing that.

zonkmachine:
> Ideally this invert should be a part of the same selection function. We should probably look into doing that.

Unfortunately I won't have time to look into this any further for now so I'm closing this PR.

Please grab the code... :)

Issue #4749, opened 2018-12-31
Theming of disabled components


zonkmachine:
Some components that are disabled don't show this without theming. An example being knobs, see the tension knob in the Automation Editor. It's only enabled with the cubic hermite progression and disabled for the other progressions but there is no visual indication that it's disabled other than that it's immobile.

https://github.com/LMMS/lmms/blob/386c471ed7fc1cc565be761927c2d7058512f4b9/src/gui/editors/AutomationEditor.cpp#L2506

I think all changes needed may be in `style.css`. We have some theming of disabled components already. See the `Edit->Redo` menu entry when there are no undone actions.

https://github.com/LMMS/lmms/blob/386c471ed7fc1cc565be761927c2d7058512f4b9/data/themes/default/style.css#L109

@LMMS/artwork-team 

_Edit: New commit with disabled knobs (https://github.com/LMMS/lmms/pull/4378)..._ https://github.com/LMMS/lmms/blob/a3ede058cdabb9254f50e69e5133fe5807036f39/src/gui/widgets/EffectView.cpp#L78

thmueller64:
I've opened a draft PR on this one. Unfortunately, I could not get it to work with pseudo-states. See the PR for the reasoning. Maybe someone can test it and provide feedback. I still have to look for resource leaks etc., so it is a draft PR for now.

![knob_theming](https://user-images.githubusercontent.com/64359888/80872254-20772280-8cb1-11ea-8059-0e8465f501bb.gif)

zonkmachine:
Tested. This works lovely, thanks!



> I could not get it to work with pseudo-states

And what's a pseudo state?  :)

thmueller64:
Thank you for the testing! I had a discussion on Discord with @Veratil about that feature and refactored the PR quite a bit. I'll open a new one, but this time not as a draft.

With a pseudo state you can easily control the theming based on the state of a Widget, like:
```
QPushButton:enabled {color: green}
QPushButton:disabled {color: red}
```

zonkmachine:
Closed in https://github.com/LMMS/lmms/pull/5549

PR #5481, opened 2020-05-02
Theming of disabled knobs

thmueller64:
This PR partly addresses #4749 and introduces the possibility to theme disabled knobs. 
I could not get it to work using the pseudo state ":disabled" in the style sheets, as suggested in #4749. As the knob icons are png Images, we somehow have to catch the `enabledChange` event and convert the image to grayscale. Moreover, the pseudo-states do not seem to work with a custom `QPROPERTY`, in contrast to the standard ones like `background-color` etc. . So redrawing the middle line and arc does not seem possible with pseudo-states.

### DIFF
--- a/data/themes/default/style.css
+++ b/data/themes/default/style.css
@@ -40,6 +40,11 @@ QMdiArea {
 	background-color: #111314;
 }
 
+Knob[enabled=false] {
+	qproperty-lineColor: rgb(120, 120, 120);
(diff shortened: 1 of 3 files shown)


PhysSong:
@thmueller64 Can we close this in favor of #5549?

thmueller64:
Yes, I've forgot to close this one, sorry!

PR #5482, opened 2020-05-02
Feature: Glue notes in Piano-Roll

zonkmachine:
https://github.com/LMMS/lmms/pull/5216 by CYBERDEViLNL rebased on c755b56 and merge conflict solved.


zonkmachine:
I have tested the glue feature and it works well. I think it's good to merge. The original developer has left GitHub and wont be able to respond so If someone wants to fix anything, either push to this branch or use GitHubs online editor.

zonkmachine:
Merge?

src/gui/editors/PianoRoll.cpp
```diff
@@ -662,6 +663,73 @@ void PianoRoll::clearGhostPattern()
...
+		{
+			// key and position match for glue
+			if ((*note)->key() == (*nextNote)->key()
+				&& (*nextNote)->pos() >= (*note)->pos()
+				&& (*nextNote)->pos() <= (*note)->pos()
+				+ (*note)->length())
+			{
+				(*note)->setLength((*note)->length() +
+						((*nextNote)->pos() + (*nextNote)->length()) -
+						((*note)->pos() + (*note)->length()));
```
Spekular, line 710:
```suggestion
				(*note)->setLength((*nextNote)->endPos() - (*note)->length());
```

Spekular: This should work... right?

Spekular:
Does this work for the following cases? I fail to see how this logic could merge more than two consecutive notes with the loop it currently uses:

image

zonkmachine:
> Does this work for the following cases?

Yes.



> Something about the buildbot build makes it crash when I open LMMS or I'd test myself.

I have seen crashes and I've had to launch lmms with a file known to work as an argument to make it launch. Not with this build though.

zonkmachine:
Oh. It doesn't seem to work well with beat notes though. The result is really inconsistent there.

zonkmachine:
Yeah, look at the length I managed to create on the first beat note when trying to glue it to a normal one.

```
<pattern steps="16" type="0" name="Kicker" pos="0" muted="0">
  <note len="-168" key="57" vol="100" pos="12" pan="0"/>
  <note len="-192" key="57" vol="100" pos="108" pan="0"/>
</pattern>
```

Spekular:
Did you merge with an eight note? Beat notes have a length of -192, IIRC, so the math checks out if so (not that it makes this a good result, haha).

I wonder how much effort it would take to make beat notes a subclass of Note so length and endPos could be overriden. Or even simpler, add a boolean (isBeatNote, for example) so they can have positive lengths. Might be out of scope for this PR but they really cause a lot of trouble with the current hacky implementation.

zonkmachine:
> Did you merge with an eight note? Beat notes have a length of -192, IIRC, so the math checks out if so (not that it makes this a good result, haha).

No, more like a whole note.

src/gui/editors/PianoRoll.cpp
```diff
@@ -662,6 +663,73 @@ void PianoRoll::clearGhostPattern()
...
+		auto nextNote = note+1;
+		NoteVector::iterator end = selectedNotes.end();
+
+		while (note != end && nextNote != end)
+		{
+			// key and position match for glue
+			if ((*note)->key() == (*nextNote)->key()
+				&& (*nextNote)->pos() >= (*note)->pos()
+				&& (*nextNote)->pos() <= (*note)->pos()
+				+ (*note)->length())
```
zonkmachine, line 706: `(*note)->length()` Maybe change to something like: `(*note)->length() < 0 ? 0 : (*note)->length()`

src/gui/editors/PianoRoll.cpp
zonkmachine: Did you try it?

Spekular: No, but the logic makes sense since we want our new note to cover the distance from the start of this note to the end of the next one. In addition, endPos is defined as startPos + length in Note.cpp, so you can substitute that in to prove equivalence:

len + ((nPos + nLen) - (pos + len))
len + (nEnd - (pos + len)
len + nEnd - pos - len
nEnd - pos

Where n prefix is appended to properties of the next note.

src/gui/editors/PianoRoll.cpp
Spekular: Maybe add a `(*note)->length() > 0` in the if statement to reject beat notes entirely from glue? Otherwise you'd also need similar guard logic when getting the end of the next note. Also, setting beat note length to zero for this calculation could look weird since they're drawn as four ticks long. Might lead to two notes not being merged although they look connected.

zonkmachine: Yes, that sounds better.

zonkmachine:
Oh, dear!

This project
[gluetest-02.mmp.zip](https://github.com/LMMS/lmms/files/4775340/gluetest-02.mmp.zip)
gluebug1

`<Ctrl> + A`
`<Shift> + G`
gluebug2

src/gui/editors/PianoRoll.cpp
PhysSong: @Spekular I think you meant this:
```suggestion
				(*note)->setLength((*nextNote)->endPos() - (*note)->pos());
```

src/gui/editors/PianoRoll.cpp
PhysSong: In FL studio, normal notes absorb infinite length notes(similar to our BB notes).

PhysSong:
I changed the `if` statement a little bit and was able to get the behavior identical to FL.

I also suggest gluing all notes if no notes are selected.

src/gui/editors/PianoRoll.cpp
Spekular: Yes, definitely. Weird slip up on my part in the earlier suggested change 😬.

PhysSong:
I've updated this PR with the changes I suggested.

zonkmachine:
Fixed a merge conflict.

zonkmachine:
I've tested this PR again and the only problem I saw earlier was fixed by @PhysSong in https://github.com/LMMS/lmms/pull/5482/commits/587c559047a7b1c47571b02063b646e18d67b2ae

zonkmachine:
Merge?

Spekular:
Sounds good to me!

src/gui/editors/PianoRoll.cpp
```diff
@@ -627,6 +628,72 @@ void PianoRoll::clearGhostPattern()
...
+
+		NoteVector::iterator note = selectedNotes.begin();
+		auto nextNote = note+1;
+		NoteVector::iterator end = selectedNotes.end();
+
+		while (note != end && nextNote != end)
+		{
+			// key and position match for glue
+			if ((*note)->key() == (*nextNote)->key()
+				&& (*nextNote)->pos() >= (*note)->pos()
```
IanCaio, line 669: Will that conditional ever be `false`? Since the notes were previously ordered by key and by pos, if two notes have the same key isn't it always `true` that the next note will have a position equal or greater than the previous note?

IanCaio: The comment highlighted all 4 lines, so to make it clearer the conditional I meant is `(*nextNote)->pos() >= (*note)->pos()` which will be evaluated if the `key`s are the same on the two notes.

PhysSong: It is redundant, but I think there should be a comment which explains what you said.

PhysSong:
Any ideas how we should deal with per-note pitch automation? Should we ignore it for now?

IanCaio:
> Any ideas how we should deal with per-note pitch automation? Should we ignore it for now?

IMHO it could be just removed from any glued note. It's to be expected that pitch automations would be lost in a note glue operation, and removing it we avoid the inconsistency of it being kept if it's in the first note or being removed if it's in the second one. So just removing it altogether sounds like a good idea to me.

Quickly looking at the code I think it has to be done like this?

```
// While glueing the note
sharedObject::unref( (*note)->detuning() );
```

Personally, I even think that this `sharedObject` call should be added to `Note.h` as a helper method, something like `removeDetune()`, to make the code less obscure to readers.

zonkmachine:
> Any ideas how we should deal with per-note pitch automation? Should we ignore it for now?

I think we can ignore it for now. The effect of notes with pitch automation in a glue operation isn't hard to predict.

src/gui/editors/PianoRoll.cpp
zonkmachine: So, what needs to be done here? Remove `&& (*nextNote)->pos() >= (*note)->pos()` and improve the comment?

IanCaio: Yeah, I think just removing that part of the conditional and having the comment explaining that you don't need to check if the next note is after the first because they are ordered already. Removes an unnecessary comparison but leaves an explanation why.

zonkmachine: OK. Hopefully fixed in https://github.com/LMMS/lmms/pull/5482/commits/d86ba8c17d461922e082b5a93c894b6c63506627

zonkmachine:
Now merge? :)

Spekular (approved):
LGTM.

IanCaio (approved):
LGTM too!

zonkmachine:
Rebased in a separate PR https://github.com/LMMS/lmms/pull/5721. Closing this one.

PR #5489, opened 2020-05-09
Fix #4201: BB editor: adjust cursor position

### LINKED ISSUE #4201: Beat editor: clicking on a beat toggles the previous beat on/off

JohannesLorenz:
This fixes an offset for cursors whose pointer position varies between
different themes.

~~TODO: Offset is not 100% accurate yet. (16, 16) is wrong, but it's
good for testing.~~ -> Fixed. Un-drafting PR.

### DIFF
--- a/data/themes/classic/style.css
+++ b/data/themes/classic/style.css
@@ -619,8 +619,7 @@ TrackContentObjectView   {
 	qproperty-textShadowColor: rgb( 0, 0, 0 );
 	qproperty-gradient: true; /* boolean property, set true to have a gradient */
 	/* finger tip offset of cursor */
-	qproperty-mouseHotspotX: 3;
-	qproperty-mouseHotspotY: 3;
+	qproperty-mouseHotspotHand: 3px 3px;
 }
 
 /* instrument pattern */
--- a/data/themes/classic/style.css
+++ b/data/themes/classic/style.css
@@ -618,6 +618,9 @@ TrackContentObjectView   {
 	qproperty-textColor: rgb( 255, 255, 255 );
 	qproperty-textShadowColor: rgb( 0, 0, 0 );
 	qproperty-gradient: true; /* boolean property, set true to have a gradient */
+	/* finger tip offset of cursor */
+	qproperty-mouseHotspotX: 3;
+	qproperty-mouseHotspotY: 3;
 }
 
 /* instrument pattern */
--- a/include/Track.h
+++ b/include/Track.h
@@ -201,6 +201,8 @@ class TrackContentObjectView : public selectableObject, public ModelView
 	Q_PROPERTY( QColor textShadowColor READ textShadowColor WRITE setTextShadowColor )
 	Q_PROPERTY( QColor BBPatternBackground READ BBPatternBackground WRITE setBBPatternBackground )
 	Q_PROPERTY( bool gradient READ gradient WRITE setGradient )
+	Q_PROPERTY( int mouseHotspotX READ getMouseHotspotX WRITE setMouseHotspotX )
+	Q_PROPERTY( int mouseHotspotY READ getMouseHotspotY WRITE setMouseHotspotY )
 
 public:
 	TrackContentObjectView( TrackContentObject * tco, TrackView * tv );
@@ -226,6 +228,8 @@ public:
 	QColor textShadowColor() const;
 	QColor BBPatternBackground() const;
 	bool gradient() const;
+	int getMouseHotspotX() const { return m_mouseHotspotX; }
+	int getMouseHotspotY() const { return m_mouseHotspotY; }
 	void setMutedColor( const QColor & c );
 	void setMutedBackgroundColor( const QColor & c );
 	void setSelectedColor( const QColor & c );
@@ -233,6 +237,9 @@ public:
 	void setTextShadowColor( const QColor & c );
 	void setBBPatternBackground( const QColor & c );
 	void setGradient( const bool & b );
+	void setMouseHotspotX(int x) { m_mouseHotspotX = x; }
+	void setMouseHotspotY(int y) { m_mouseHotspotY = y; }
+
 
 	// access needsUpdate member variable
 	bool needsUpdate();
@@ -302,8 +309,11 @@ private:
 	QColor m_textShadowColor;
 	QColor m_BBPatternBackground;
 	bool m_gradient;
+	int m_mouseHotspotX;
+	int m_mouseHotspotY;
+	bool m_cursorSetYet = false;
 
- 	bool m_needsUpdate;
+	bool m_needsUpdate;
 	inline void setInitialMousePos( QPoint pos )
 	{
 		m_initialMousePos = pos;
--- a/include/Track.h
+++ b/include/Track.h
@@ -30,6 +30,7 @@
 #include <QtCore/QList>
 #include <QWidget>
 #include <QSignalMapper>
+#include <QSize>
 #include <QColor>
 #include <QMimeData>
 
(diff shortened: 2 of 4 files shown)


JohannesLorenz:
I recommend Veratil should check the code, as he designed a proposal for it and is involved in the issue.

src/core/Track.cpp
```diff
@@ -572,7 +580,7 @@ void TrackContentObjectView::leaveEvent( QEvent * e )
 {
 	if( cursor().shape() != Qt::BitmapCursor )
 	{
-		setCursor( QCursor( embed::getIconPixmap( "hand" ), 3, 3 ) );
+		setCursor( QCursor( embed::getIconPixmap( "hand" ), m_mouseHotspotX, m_mouseHotspotY ) );
```
JohannesLorenz, line 583: Looking at this now, I think we need to add `3` and `3` in the "classic" theme CSS?

src/core/Track.cpp
```diff
@@ -317,6 +319,12 @@ TrackContentObjectView::~TrackContentObjectView()
  */
 void TrackContentObjectView::update()
 {
+	if( !m_cursorSetYet )
```
JohannesLorenz, line 322: I only set this once to avoid setting the cursor all the time. I hope this is enough, but the mouse hotspot should actually never change during runtime, right?

Spekular: It can be set when a cursor is constructed, isn't that enough?

Spekular: Nevermind, I missed your comment in the issue post.

JohannesLorenz: What do you mean? The cursor has already been constructed in `TrackContentObjectView`'s CTOR, and at that time, the hotspot is not yet read from the CSS.

src/core/Track.cpp
zonkmachine: Correct, and I believe the new cursor was using something like 3, 7 last I tried it.

JohannesLorenz: How do you find out the right coordinates? Opening both images in gimp/etc and just use where the finger points? If yes, I think it's (2,7) for the default, and (3,3) for the classic.

zonkmachine: Right, 2,7. I've added the settings for classic too and this works fine.

JohannesLorenz: Thanks, added this in a new commit.

JohannesLorenz:
I tested it and now the mouse selects exactly the correct patterns (though it might be subjective where exactly the finger points). IMO there's no testing required, only the code review.

JohannesLorenz:
CI (Windows) just complained:
```
/home/travis/build/LMMS/lmms/include/Track.h:314:24: error: non-static data member initializers only available with -std=c++11 or -std=gnu++11 [-Werror]

  bool m_cursorSetYet = false;
```
Can't we set -std=c++11? Otherwise, I'll change this code ofc.

PhysSong:
We disabled C++11 for some plugins on `stable-1.2`, so I think you need to change the code.

data/themes/classic/style.css
```diff
@@ -618,6 +618,9 @@ TrackContentObjectView   {
 	qproperty-textColor: rgb( 255, 255, 255 );
 	qproperty-textShadowColor: rgb( 0, 0, 0 );
 	qproperty-gradient: true; /* boolean property, set true to have a gradient */
+	/* finger tip offset of cursor */
+	qproperty-mouseHotspotX: 3;
```
Spekular, line 622: Can we name this something more specific, since we have several different cursors? `cursorHotspotX-hand`?

For example, when the Knife tool is merged that will have it's own cursor which should ideally be themeable as well. If this is merged as-is and then renamed, we'd end up breaking themes between 1.2 and 1.3 unnecessarily.

include/Track.h
```diff
@@ -226,13 +228,18 @@ class TrackContentObjectView : public selectableObject, public ModelView
 	QColor textShadowColor() const;
 	QColor BBPatternBackground() const;
 	bool gradient() const;
+	int getMouseHotspotX() const { return m_mouseHotspotX; }
+	int getMouseHotspotY() const { return m_mouseHotspotY; }
```
Spekular, line 232: Are these functions necessary? I removed all references to `getMouseHotspot_`, and the PR still compiled and worked as expected. Once again, I'm concerned about adding more cursors, so it'd be ideal if each one took as few lines as possible.

Spekular:
> IMO there's no testing required, only the code review.

I tested anyway, works as expected! I left a review, but technically the comments aren't relevant for 1.2. Looks good otherwise.

data/themes/classic/style.css
JohannesLorenz: Wouldn't that knife symbol need to be treated the same way in order to hit the right B&B blocks? With your proposition, we'd then name it `cursorHotspotX-knife`?

Spekular: Yes, that's what I'm thinking.

include/Track.h
JohannesLorenz: Indeed, I removed those here and after the `READ` in the properties. Works fine.

data/themes/classic/style.css
Spekular: The exact naming scheme is up to you of course, the essence of the suggestion is just to add the specific cursor type somewhere in the property name.

JohannesLorenz: Oh I think I get what you mean. We now have only one cursor (per theme), so one "qproperty" is always enough to describe the offset. But if we have two cursors (per theme), we need to qproperties, because the offsets can differ (between knife and hand). Do you mean that?

Spekular: Yes!

Veratil: What if we make a new class which contains all the cursor data?

Also, I'm looking at how to make `QPoint` properties which will allow us to do something like `qproperty-hotspot-hand: 3 3;`

JohannesLorenz: @Veratil I like the `QPoint` idea. About the new class, sounds good, but how would our `TrackContentObjectView` access that class? Do you plan that each class that needs these cursor data should inherit that special class? If such a class must inherit `QObject`, we must be careful with diamond inheritance.

zonkmachine: This PR looks good to merge to me. Do you intend the proposed new class to be implemented here? (_this PR, stable-1.2_)

JohannesLorenz: @zonkmachine That was the idea.

JohannesLorenz:
I just pushed rework which now uses `QSize` instead of 2 integers. I agreed with Veratil that this solves his comment.

Can someone please code-review this small commit ( 943062a ) ? Maybe @Spekular (you did the initial review) or @Veratil ? Then we can merge.

Spekular (approved):
LGTM! 🎉

PR #5490, opened 2020-05-10
addStretch() before adding the oscilloscope

Sawuare:
Before:
before

After:
after

I think it looks better.


JohannesLorenz:
Nice. Code and style review passed. Merging.

PR #5491, opened 2020-05-10
Fix #5194: Fix knobs moving too fast (alternative PR)

### LINKED ISSUE #5194: Knobs are erratic on macOS (and other systems)

JohannesLorenz:
This PR does

* Remove all `setPos()` from Knob.cpp on all OS
* Therefore use the previous position, not the original position to
  compute the cursor delta
* Use the cursor delta now linearly
* Use the cursor delta in relation to the screen height

### DIFF
--- a/src/gui/widgets/Knob.cpp
+++ b/src/gui/widgets/Knob.cpp
@@ -511,7 +511,7 @@ float Knob::getValue( const QPoint & _p )
 
 	// knob value increase is linear to mouse movement
 	// for larger screens, change the knob slower
-	value = .09f * (_p.y()*(1080/s_desktopSize.y()));
+	value = .09f * (((float)_p.y())*(1080.f/(float)s_desktopSize.y()));
 
 	// if shift pressed we want slower movement
 	if( gui->mainWindow()->isShiftPressed() )
--- a/src/gui/widgets/Knob.cpp
+++ b/src/gui/widgets/Knob.cpp
@@ -511,7 +511,7 @@ float Knob::getValue( const QPoint & _p )
 
 	// knob value increase is linear to mouse movement
 	// for larger screens, change the knob slower
-	value = .09f * (_p.y()/(1080/s_desktopSize.y()));
+	value = .09f * (_p.y()*(1080/s_desktopSize.y()));
 
 	// if shift pressed we want slower movement
 	if( gui->mainWindow()->isShiftPressed() )
--- a/src/gui/widgets/Knob.cpp
+++ b/src/gui/widgets/Knob.cpp
@@ -24,10 +24,12 @@
 
 #include <QApplication>
 #include <QBitmap>
+#include <QDebug>
 #include <QFontMetrics>
 #include <QInputDialog>
 #include <QMouseEvent>
 #include <QPainter>
+#include <QScreen>
 #include <QWhatsThis>
 
 #ifndef __USE_XOPEN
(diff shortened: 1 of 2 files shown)


JohannesLorenz:
This alternative to the other PR ( #5360 ) has some differences:

* No `setPos`/cursor hiding/border flip at all (all operating systems equal)
* Knob increase is now depending on **screen height** (DPI might be better, as @PhysSong noted, but let's just try this)

I'm not deceived if this won't work :smiley: But let's just try it out.

zonkmachine:
Tested. Turning a knob is now much too slow on my machine (linuxmint, 1024x600). If I tweak value `= .09f * (_p.y( ...` to something like `= .5f * (_p.y( ...` I get a better action.

JohannesLorenz:
@zonkmachine Sorry, I divided the screen factor instead of multiplying. It should now be better. Can you please test again?

src/gui/widgets/Knob.cpp
```diff
@@ -497,8 +509,9 @@ float Knob::getValue( const QPoint & _p )
 {
 	float value;
 
-	// arcane mathemagicks for calculating knob movement
-	value = ( ( _p.y() + _p.y() * qMin( qAbs( _p.y() / 2.5f ), 6.0f ) ) ) / 12.0f;
+	// knob value increase is linear to mouse movement
+	// for larger screens, change the knob slower
+	value = .09f * (_p.y()*(1080/s_desktopSize.y()));
```
zonkmachine, line 515: ` (_p.y()*(1080/s_desktopSize.y()))` These are all int values right?
Maybe force floating point computation to avoid integer math with truncation?
`value = .09f * (_p.y()*(1080.f/s_desktopSize.y()));`

tresf:
> Use the cursor delta in relation to the screen height

I'm still a bit unclear as to why we'd do this.  This means portrait mode is a different sensitivity than landscape.  A sane value is and will always be my vote.  Any black magic is going to make inconsistent results.

src/gui/widgets/Knob.cpp
JohannesLorenz: Oh how could I not have seen this. Of course, on my screen with 1080 height this always worked, but that explains a lot.

I pushed a fix now, sorry for that.

zonkmachine:
`23aa4ec` OK, now we're getting closer. With the mousepad on 1.1.3 I would be able to rotate a knob from it's extreme positions without repositioning my finger to the middle of the pad. Now I need to do it. Changing .09f to .2f gives the right action.

JohannesLorenz:
Thanks for testing.

> Changing .09f to .2f gives the right action.

Interesting. 0.2f * (1080/600) ~ 0.4f. The other PR uses 0.4f as well now, so that one should be fine, too.

JohannesLorenz:
Not used. Closed by request.

PR #5492, opened 2020-05-10
Remove warning color from oscilloscope

Sawuare:
Because there is nothing to warn about when the audio level is in the arbitrary range of [90%, 100%].


Sawuare:
@DouglasDGI Requesting your review since you expressed concerns about the change earlier.

> I think it would be best to keep the three volume dependent colors, I reference them quite often and other people probably do as well.

DouglasDGI:
I think I'm fine with this.  The clipping indicator is essential but the warning indicator isn't really needed.

Sawuare:
Thanks for the prompt reply. Merging. :rocket:

Issue #3400, opened 2017-03-04
There should be the option to universally assign a Midi input, and then have that Midi input applied to just the current armed track.


tigerprawn:
At the moment if you are playing a Midi keyboard and you want to use it on a different tracks, you need to connect it every time to each new track and arm the track, and even then it is still connected to the other unarmed tracks. I think it would be better if it functioned like Ableton where you setup a Midi device in Preferences and then it is automatically connected just to the current armed track.

tresf:
We're aware.  Mentioned here generically (Faster access to MIDI assignment/auto-assignment): #1472

Which is also mentioned here #1189 #1781 #1918 #1942 #3185

Issue #3185, opened 2016-12-28
Midi keyboard working in style of the normal keyboard


Devix167:
BEFORE YOU READ: Im new to github and all this so dont be mad if i screw up 😄 

I hate having to re assign the midi keyboard just to play a different instrument. can midi keyboards work like a normal keyboard? like a settings in the menu or something? 🎹

tresf:
@Devix167 is there a chance you're request is similar to these?

#1189 #1781 #1918 #1942 (parent bug report: #1472)

Devix167:
@tresf 
Sorry about that. I did numerous searches and none of these came up.

tresf:
> Sorry about that. 

No apology necessary; needle in a haystack.  Closing as duplicate of #1472.

Issue #1942, opened 2015-04-10
Fast midi switcher


Achifaifa:
Hi

I use a small MIDI keyboard with lmms. Works fine, but when I have to assign it to a track I have to click in "actions'>'MIDI'>'input'>controller, and when I want to assign it to another track I have to do that twice, once to disable it in the current track and another one to enable it in the other one.

Could it be possible po have an extra check mark per instrument with the 'mute track'/'solo' buttons? A column of buttons per instrument you can click to enable/disable it in that track. That would make switching instruments so much faster. Another pop-up window with a table showing instruments and tracks would also work

tresf:
@Achifaifa thanks for filing a bug.  This is a duplicate of #1189 #1781 #1918, which is already on our radar.  Thanks!

-Tres

oeai:
you can assign midi channel #1 to first instrument
and midi channel #2 to a second
then you can change channels on your keyboard to switch between instruments

Achifaifa:
@tresf Thanks :)
@oeai It's a 25 key midi controller (Akai LPK25) so I don't think I can do that, but thanks for the tip.

Issue #1918, opened 2015-04-03
Capture rehaul and MIDI handling


unfa:
# Preface
It started when I got a few ideas where LMMS could better use so called "master keyboards" or musical keyboards in general. The idea is as always - to save user's time and do stuff for him, to keep him focused on the music, not breaking world records for mouse clicks per minute.
However it turned out I have much more on my mind, and the issue goes into rethinking the process of capturing MIDI and audio (not yet implemented) data into LMMS and making use of MIDI data in a smarter way.

# Issues
### Problem
LMMS allows to use any MIDI device as input and/or output to each instrument. That's nice for live operation I'd say, where I can route different controllers to different instruments or even filter that via MIDI channels (16 instruments driven by one keyboard etc.).

What I lack is a "master MIDI controller" option, that'd allow the user to select a MIDI device he wants to use as a"fallback" or default MIDI input for Piano Roll and Automation data recording.
Why do I think we need this? Time of course.

The current workflow for recording notes for two (let's call them A and B) instruments is as follows:
1. Assign the MIDI keyboard as input for the instrument A,
2. Record notes for instrument A,
3. Disable the MIDI input for instrument A,
4. Assign the MIDI input for instrument B,
5. Record notes for instrument B,
6. Disable MIDI input for instrument B

And with a master MIDI controller option it'd look like this:
1. Select master MIDI input (once and for all)
2. Record notes for instrument A
3. Record notes for instrument B

Also, the Piano Roll and Automation editor workflows are different - you don't have "REC" button in Automation editor, and you need to arm/disarm automation patterns for recording - instrument clips don't have that. Why instrument and automation regions work differently? Can we make this consistent? I fell there's even better way of handling all this recording thing.

### Solution
My idea is to remove recording controls from Piano Roll window and place them in the main transport pane. The REC would be a toggle button, also each track: instrument, automation, audio would have an additional "REC" toggle LED button similar to the "ON" and "SOLO" controls we have there. No per-region record-arming for automation, instrument or audio clips.

In the settings window there could be unified "Capture" tab with panels for: instrument, automation and audio capture. The user could select a default device for each capture type (notes, auto, audio). 
This could also cover capturing multiple tracks of the same type at once.

For audio: for example JACK ports system:capture1,2 for 1st audio track, ports 3,4 for 2nd, 5,6 for 3rd etc.
For automation: device(s), MIDI channel(s),  CC1, CC2, CC3... or CC1-8 or CC4,6,8,10 whatever.
[...]

PR #5499, opened 2020-05-15
Automatically assign a midi input device to the selected track

LNooteboom:
Hello everyone, this is my first pullrequest here!

This pullrequest implements "Faster access to MIDI assignment/auto-assignment" in meta-issue #1472 (as well as #3400 #1189 #1781 #1918 #1942 #3185)

With this pullrequest, it is now possible to automatically assign/connect a "master" midi input device (eg. a keyboard) to the currently selected instrument track and piano roll, this makes it behave similarly to the QWERTY keyboard for note input.
Whereas previously you needed to manually assign your MIDI keyboard via the menu every time you wanted to play a different instrument, now you can just select your main midi keyboard in the settings menu (MIDI Tab), and LMMS will automatically assign it. This speeds up the workflow for people who record using a MIDI device.

The use of this feature is completely optional, if you leave the auto-assign setting at 'none', your workflow will not be impacted at all.

### DIFF
--- a/src/gui/SetupDialog.cpp
+++ b/src/gui/SetupDialog.cpp
@@ -677,9 +677,25 @@ SetupDialog::SetupDialog(ConfigTabs tab_to_open) :
 			this, SLOT(midiInterfaceChanged(const QString &)));
 
 
+	// MIDI autoassign tab.
+	TabWidget * midiAutoAssign_tw = new TabWidget(
+			tr("Automatically assign MIDI controller to selected track"), midi_w);
+	midiAutoAssign_tw->setFixedHeight(56);
+
+	m_assignableMidiDevices = new QComboBox(midiAutoAssign_tw);
+	m_assignableMidiDevices->setGeometry(10, 20, 240, 28);
+	m_assignableMidiDevices->addItem("none");
+	m_assignableMidiDevices->addItems(Engine::mixer()->midiClient()->readablePorts());
+	int current = m_assignableMidiDevices->findText(ConfigManager::inst()->value("midi", "midiautoassign"));
+	if (current >= 0)
+	{
+		m_assignableMidiDevices->setCurrentIndex(current);
+	}
+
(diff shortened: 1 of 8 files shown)


PhysSong:
Thanks for the great work!
Some thoughts:
- The MIDI input gets unassigned if I open a dialog or a context menu. It'd be great if we could improve that behavior.
- This is useful with sequenced MIDI clients, but not with raw MIDI clients, i.e., MIDI backends which don't separate inputs from different MIDI input devices.

src/gui/SetupDialog.cpp
```diff
@@ -685,7 +685,14 @@ SetupDialog::SetupDialog(ConfigTabs tab_to_open) :
...
 	m_assignableMidiDevices->setGeometry(10, 20, 240, 28);
 	m_assignableMidiDevices->addItem("none");
-	m_assignableMidiDevices->addItems(Engine::mixer()->midiClient()->readablePorts());
+	if ( !Engine::mixer()->midiClient()->isRaw() )
+	{
+		m_assignableMidiDevices->addItems(Engine::mixer()->midiClient()->readablePorts());
+	}
+	else
+	{
+		m_assignableMidiDevices->addItem("yes");
```
PhysSong, line 694: "none" vs. "yes" look a bit odd. I think "all" might be better, any ideas?

qnebra:
Works fine for casual preview and recording with midi devices. Of course, automation was bit tricker and still requiered more steps, but overall this is fine.

PhysSong (approved):

qnebra:
You are in 1.2.2 and this feature is in at least 1.3-alpha

Saitama697:
> You are in 1.2.2 and this feature is in at least 1.3-alpha

Okay, thank you! I’m just curious— the post I quoted was made on May 15, 2020. How long does it usually take for an update to reach the stable version?

Just curious.

Thank you!

PR #5501, opened 2020-05-16
24 bit FLAC export. Clip negative side of wave

### LINKED ISSUE #5128: FLAC export not identical to WAV
When exporting to FLAC with a recent master build, I noticed that the output had noticeable (sometimes severe) crackling not present in previous exports. Here's a comparison of the output waveform, with WAV on top and FLAC on the bottom:

image

Examining the top left of both waveforms, the FLAC export appears to have a downwards DC offset. In addition, when this offset wave reaches -1.0, it wraps around and "sticks" at 1.0.

Export options: 44.1kHz, 24 Bit, Sinc best, No oversampling

Also, a minor filename bug:

- When choosing output location, leave extension on the default "WAV"
- Select FLAC as output format
- File extension of exported file is `.wav.flac`

zonkmachine:
Fixes #5128

### DIFF
--- a/src/core/audio/AudioFileFlac.cpp
+++ b/src/core/audio/AudioFileFlac.cpp
@@ -24,6 +24,7 @@
 
 #include <QtGlobal>
 
+#include <cmath>
 #include <memory>
 
 #include "AudioFileFlac.h"
@@ -88,6 +89,7 @@ bool AudioFileFlac::startEncoding()
 void AudioFileFlac::writeBuffer(surroundSampleFrame const* _ab, fpp_t const frames, float master_gain)
 {
 	OutputSettings::BitDepth depth = getOutputSettings().getBitDepth();
+	float clipvalue = std::nextafter( -1.0f, 0.0f );
 
 	if (depth == OutputSettings::Depth_24Bit || depth == OutputSettings::Depth_32Bit) // Float encoding
 	{
@@ -99,7 +101,7 @@ void AudioFileFlac::writeBuffer(surroundSampleFrame const* _ab, fpp_t const fram
 				// Clip the negative side to -0.999999 in order to prevent it from changing sign
 				// Upstream issue: https://github.com/erikd/libsndfile/issues/309
 				// When this commit is reverted libsndfile-1.0.29 must be made a recuirement for FLAC
-				buf[frame*channels() + channel] = qMax(-0.999999f, _ab[frame][channel] * master_gain);
+				buf[frame*channels() + channel] = qMax( clipvalue, _ab[frame][channel] * master_gain);
 			}
 		}
 		sf_writef_float(m_sf,static_cast<float*>(buf.get()),frames);
--- a/src/core/audio/AudioFileFlac.cpp
+++ b/src/core/audio/AudioFileFlac.cpp
@@ -98,7 +98,7 @@ void AudioFileFlac::writeBuffer(surroundSampleFrame const* _ab, fpp_t const fram
 		{
 			for(ch_cnt_t channel=0; channel<channels(); ++channel)
 			{
-				// Clip the negative side to -0.999999 in order to prevent it from changing sign
+				// Clip the negative side to just above -1.0 in order to prevent it from changing sign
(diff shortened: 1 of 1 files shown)


src/core/audio/AudioFileFlac.cpp
```diff
@@ -94,7 +96,10 @@ void AudioFileFlac::writeBuffer(surroundSampleFrame const* _ab, fpp_t const fram
 		{
 			for(ch_cnt_t channel=0; channel<channels(); ++channel)
 			{
-				buf[frame*channels() + channel] = _ab[frame][channel] * master_gain;
+				// Clip the negative side to -0.999999 in order to prevent it from changing sign
+				// Upstream issue: https://github.com/erikd/libsndfile/issues/309
+				// When this commit is reverted libsndfile-1.0.29 must be made a recuirement for FLAC
+				buf[frame*channels() + channel] = qMax(-0.999999f, _ab[frame][channel] * master_gain);
```
PhysSong, line 102: `std::nextafterf(-1.0f, 0.0f)` from `<cmath>` is a portable alternative of the constant.

zonkmachine: Cool! I see that I used nextafter() instead of nextafterf(). They seem to be the same though.

zonkmachine: Same-ish. Changing to nextafterf().

zonkmachine:
@PhysSong I'm happy with this now. It would probably be a good idea if someone else tested it. Just render a short note, that is overdriven and clipping in lmms, in 24 bit flac and make sure it's not wonky.

src/core/audio/AudioFileFlac.cpp
```diff
@@ -94,7 +98,10 @@ void AudioFileFlac::writeBuffer(surroundSampleFrame const* _ab, fpp_t const fram
 		{
 			for(ch_cnt_t channel=0; channel<channels(); ++channel)
 			{
-				buf[frame*channels() + channel] = _ab[frame][channel] * master_gain;
+				// Clip the negative side to -0.999999 in order to prevent it from changing sign
```
zonkmachine, line 101: Should this now be `// Clip the negative side to just above -1.0 ...` ?

Spekular: I think so.

src/core/audio/AudioFileFlac.cpp
```diff
@@ -94,7 +98,10 @@ void AudioFileFlac::writeBuffer(surroundSampleFrame const* _ab, fpp_t const fram
...
+				// Upstream issue: https://github.com/erikd/libsndfile/issues/309
+				// When this commit is reverted libsndfile-1.0.29 must be made a recuirement for FLAC
```
Spekular, line 103:
```suggestion
				// When this commit is reverted libsndfile-1.0.29 must be made a requirement for FLAC
```

src/core/audio/AudioFileFlac.cpp
zonkmachine: Fixed!

zonkmachine:
@PhysSong and @Spekular Thanks for the review and testing!

I believe this is ready to merge. We found a noise issue, a small difference between exporting wav and flac (both at 24 bit/44.1kHz) that seemed to be not related to this PR. It looks like a one bit rounding off error that is distributed between the two waves except where clipping occurs. Where there is clipping there is no difference between the sounds. It should be documented though.

zonkmachine:
> It looks like a one bit rounding off error that is distributed between the two waves except where clipping occurs.

Note to self. Audacity performs dither by default on some sample conversions. Must turn it off and redo the tests. In this case specifically the dither could have occurred on loading samples into Audacity.

zonkmachine:
I can't replicate the noise with 16 bit export but the noise is there on master on 24 bit. Just export some file that isn't clipping and it's there already. Turning off dither on audacity didn't do anything and I also tried only bitwidth in the sound chain that wouldn't dither and that didn't do it either. Probably not dither then.

zonkmachine:
I think we're done here. Merge?

Spekular:
Agreed!

Issue #5340, opened 2019-12-07
zero-length-notes added to normal-notes does only play the bar with the normal-note


musikBear:
Zero-length notes added to normal notes only play when in the bar with the normal note.

### Bug Summary
When deciding where to loop, the piano roll does not account for zero-length notes when there are non-zero-length notes present (see gif below). Related: #5339 

#### Steps to reproduce
* Open a block in piano-roll. Insert one normal note. 
* Copy a several bar long pattern made in the B&B-editor. Insert it in newly made block.
* Play.

#### Expected behavior
* All notes should play normally both in the piano roll and the song editor.

#### Actual behavior
* The piano roll will not continue playing to the additional bar.

#### Screenshot
![dwhiCGmdB0](https://user-images.githubusercontent.com/6700184/70382466-b5583880-192a-11ea-8fdb-873a0fd75831.gif)

#### LMMS version used
1.2.1 on Windows 10

Demo file: 
[zero-length-notesApart121.zip](https://github.com/LMMS/lmms/files/3935676/zero-length-notesApart121.zip)

*SecondFlight: move info to bug description, clarify, and add screen recording*

Spekular:
Please provide steps to reproduce and reopen.

musikBear:
1.2.1 on win10
zero-length-notes added to normal-notes does only play the bar with the normal-note
Recreate:
* open a block in piano-roll. Insert one *normal*-note. 
* Copy a several bar long pattern made in B&B-editor. Insert it in newly made block upstram the single normal-note
* Play
Behaviour:
* Piano-roll The replay will only happen in the bar with the normal note. Other notes are ignored. The playhead loops in the recognizable bar.
* Song-editor Replay only of normal-note bar. Rest bar invisible and not relayed
Expected:
* All notes should play normally both in PR and SE.

Demofile: 
[zero-length-notesApart121.zip](https://github.com/LMMS/lmms/files/3935676/zero-length-notesApart121.zip)

Issue #5339, opened 2019-12-07
Zero-length notes copied from the B&B editor cause incorrect behavior in the song editor


musikBear:
### Bug Summary
Notes copied from the B&B editor do not get registered as notes in the song editor. The block with the notes remains empty and grey.
In the picture on the left, you can see the grey block. Its length is extended, but there is no visible content. On the right, you can see the notes actually in the block.

![zeroLengthNotes121](https://user-images.githubusercontent.com/6662066/70381737-0d556600-1950-11ea-89e3-99be708edae7.JPG)
(The block will *appear* to have no content, but content is played anyway)

#### Steps to reproduce
Recreate:
* Add some content to the default kicker patch in the B&B editor
* Open the pattern content in the piano roll
* Select all and copy
* Open a new clip on an instrument in the song editor and paste the notes

#### Expected behavior
The clip turns green and extends in length.

#### Actual behavior
The clip does not turn green, but the length is still extended.

#### LMMS version used
LMMS 1.2.1 on Windows 10

*SecondFlight: migrated info to issue description and reworded some things*

musikBear:
ver 1.2.1 on win10
Notes copied from B&B-editor does not get registered as notes in song-editor. The block with the notes remain empty and grey. 
In picture left, you see the grey *length-extended* block with no visible content. Right you see the notes actually in the block.

![zeroLengthNotes121](https://user-images.githubusercontent.com/6662066/70381737-0d556600-1950-11ea-89e3-99be708edae7.JPG)
(The block will *appear* to have no content, bu content *is played* in replay)

Recreate:
* Make a pattern in BB-editor
* From note-context-menu, select copy or open in piano-roll, then copy
* Insert the copy in an empty new piano-roll-block

PR #5502, opened 2020-05-16
#5339, #5340: Fix pasting notes from bb editor

thmueller64:
Addresses #5339 and #5340: When pasting notes from the bb editor to the piano roll, the length of the notes copied is now set to the envelope length of the underlying instrument, dependent on the BPM and envelope settings at the moment of pasting. Previously, copying the notes led to unexpected behavior including them not being registered in the song editor or being ignored by the loop. Thanks @Veratil for the discussion on discord and the change requests.

### DIFF
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -4126,6 +4126,15 @@ void PianoRoll::pasteNotes()
 			cur_note.restoreState( list.item( i ).toElement() );
 			cur_note.setPos( cur_note.pos() + Note::quantized( m_timeLine->pos(), quantization() ) );
 
+			if ((cur_note.length().getTicks() == -DefaultTicksPerBar) && (m_pattern->type() != Pattern::BeatPattern)) 
+			{
+				// the length of copied notes is set to the length of the instruments envelope
+				f_cnt_t envFrames = m_pattern->instrumentTrack()->envFrames();
+				// if notes have zero length due to rounding set the length to a small value
+				tick_t noteLength = qMax(static_cast<float>(1), envFrames / Engine::framesPerTick());
+				cur_note.setLength(MidiTime(noteLength));
+			}
+
 			// select it
 			cur_note.setSelected( true );
 
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -4131,7 +4131,7 @@ void PianoRoll::pasteNotes()
 				// the length of copied notes is set to the length of the instruments envelope
 				f_cnt_t envFrames = m_pattern->instrumentTrack()->envFrames();
 				// if notes have zero length due to rounding set the length to a small value
-				tick_t noteLength = qMax(static_cast<float>(1), envFrames / Engine::framesPerTick());
+				tick_t noteLength = static_cast<tick_t>(qMax(1.0f, ceil(envFrames / Engine::framesPerTick())));
 				cur_note.setLength(MidiTime(noteLength));
 			}
 
--- a/src/tracks/Pattern.cpp
+++ b/src/tracks/Pattern.cpp
@@ -247,7 +247,6 @@ void Pattern::removeNote( Note * _note_to_del )
 	}
 	instrumentTrack()->unlock();
 	
-	adjustSteps();
 	checkType();
 	updateLength();
 
--- a/src/tracks/Pattern.cpp
+++ b/src/tracks/Pattern.cpp
@@ -55,13 +55,14 @@ QPixmap * PatternView::s_stepBtnOffLight = NULL;
 Pattern::Pattern( InstrumentTrack * _instrument_track ) :
 	TrackContentObject( _instrument_track ),
 	m_instrumentTrack( _instrument_track ),
-	m_patternType( BeatPattern ),
+	m_patternType( MelodyPattern ),
 	m_steps( MidiTime::stepsPerBar() )
 {
 	setName( _instrument_track->name() );
 	if( _instrument_track->trackContainer()
 					== Engine::getBBTrackContainer() )
 	{
+		m_patternType = Pattern::BeatPattern;
 		resizeToFirstTrack();
 	}
 	init();
@@ -218,6 +219,7 @@ Note * Pattern::addNote( const Note & _new_note, const bool _quant_pos )
 	m_notes.insert(std::upper_bound(m_notes.begin(), m_notes.end(), new_note, Note::lessThan), new_note);
 	instrumentTrack()->unlock();
 
+	adjustSteps();
 	checkType();
 	updateLength();
 
@@ -244,11 +246,18 @@ void Pattern::removeNote( Note * _note_to_del )
 		++it;
 	}
 	instrumentTrack()->unlock();
-
+	
+	adjustSteps();
 	checkType();
 	updateLength();
 
 	emit dataChanged();
+}
+
+
+void Pattern::adjustSteps()
+{
+	m_steps = qMax(1, MidiTime(m_notes.last()->pos()).nextFullBar()) * MidiTime::stepsPerBar();
 }
 
 
--- a/src/tracks/Pattern.cpp
+++ b/src/tracks/Pattern.cpp
@@ -55,14 +55,13 @@ QPixmap * PatternView::s_stepBtnOffLight = NULL;
 Pattern::Pattern( InstrumentTrack * _instrument_track ) :
 	TrackContentObject( _instrument_track ),
 	m_instrumentTrack( _instrument_track ),
-	m_patternType( MelodyPattern ),
+	m_patternType( BeatPattern ),
 	m_steps( MidiTime::stepsPerBar() )
 {
 	setName( _instrument_track->name() );
 	if( _instrument_track->trackContainer()
 					== Engine::getBBTrackContainer() )
 	{
-		m_patternType = Pattern::BeatPattern;
 		resizeToFirstTrack();
 	}
 	init();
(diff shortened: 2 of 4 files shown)


src/gui/editors/PianoRoll.cpp
```diff
@@ -4165,6 +4165,13 @@ void PianoRoll::pasteNotes()
 			cur_note.restoreState( list.item( i ).toElement() );
 			cur_note.setPos( cur_note.pos() + Note::quantized( m_timeLine->pos(), quantization() ) );
 
+			// handle case where notes are copied from BB editor
+			if (cur_note.length().getTicks() == -192)
```
Veratil, line 4169: Just in case [this ever changes](https://github.com/LMMS/lmms/blob/master/src/tracks/Pattern.cpp#L301)...
```suggestion
			if (cur_note.length().getTicks() == -DefaultTicksPerBar)
```

src/gui/editors/PianoRoll.cpp
```diff
@@ -4165,6 +4165,13 @@ void PianoRoll::pasteNotes()
...
+			{
+				// magic for addNote to indicate that paste came from BB editor
+				cur_note.setLength(MidiTime( -193 ));
```
Veratil, line 4172: We should make the `-193` a magic number constant, so something like `BB_NOTE_COPYPASTE_LENGTH`.

src/tracks/Pattern.cpp
```diff
@@ -209,6 +209,21 @@ MidiTime Pattern::beatPatternLength() const
...
+	if (new_note->length().getTicks() == -193)
+	{
+		// the length of copied notes is set to the length of the instruments envelope
+		f_cnt_t envFrames = instrumentTrack()->envFrames();
+		tick_t noteLength = envFrames / Engine::framesPerTick();
+		// if notes have zero length due to rounding set the length to a small value
+		if (noteLength == 0)
+		{
+			noteLength = 1;
+		}
```
Veratil, line 223: Can simplify the if below:
```suggestion
		// if notes have zero length due to rounding set the length to a small value
		tick_t noteLength = qMax(1, envFrames / Engine::framesPerTick());
```

src/gui/editors/PianoRoll.cpp
thmueller64: I've added the magic number in Pattern.h

Spekular:
Instead of defining a new constant, why don't you just check if the pasted note has a length of -DefaultTicksPerBar?

thmueller64:
Because `Pattern::addNote` is also called by `Pattern::addStepNote` with a note length of `-DefaultTicksPerBar`. You would not be able to add any steps, because it would create a note equal to the envelope length of the underlying instrument. The instrument track would immediately get replaced by the piano roll when clicking on a step. Thus, I think there's no obvious way around indicating where the notes to be added come from.

**What i noticed while writing the above**: This PR fixes the problem where notes are copied from the BB Editors Piano Roll and pasted to an Instruments Piano Roll as described in the issues. I have not thought of the implications of the PR when copying notes from a BB Editor to itself or to another BB Editor. @Veratil, please remove the approval for now.

Spekular:
But, instead of setting the note length to a magic number and then resizing it when it's added, couldn't you just set the note to the correct length before adding it?

thmueller64:
You're totally right, thanks! I missed that the pattern is accessible from the `PianoRoll` class through the `m_pattern` member. Update is coming.

src/tracks/Pattern.cpp
```diff
@@ -244,14 +246,20 @@ void Pattern::removeNote( Note * _note_to_del )
 		++it;
 	}
 	instrumentTrack()->unlock();
-
+
```
PhysSong, line 249: Remove the trailing tab.
```suggestion

```

src/tracks/Pattern.cpp
```diff
@@ -218,6 +219,7 @@ Note * Pattern::addNote( const Note & _new_note, const bool _quant_pos )
 	m_notes.insert(std::upper_bound(m_notes.begin(), m_notes.end(), new_note, Note::lessThan), new_note);
 	instrumentTrack()->unlock();
 
+	adjustSteps();
```
PhysSong, line 222: Could you explain why you update `m_steps` only in this case, not whenever the length changes?

src/tracks/Pattern.cpp
```diff
@@ -55,13 +55,14 @@ QPixmap * PatternView::s_stepBtnOffLight = NULL;
 Pattern::Pattern( InstrumentTrack * _instrument_track ) :
 	TrackContentObject( _instrument_track ),
 	m_instrumentTrack( _instrument_track ),
-	m_patternType( BeatPattern ),
+	m_patternType( MelodyPattern ),
```
Veratil, line 58: Just to clarify, what is this change for?

src/gui/editors/PianoRoll.cpp
```diff
@@ -4126,6 +4126,15 @@ void PianoRoll::pasteNotes()
 			cur_note.restoreState( list.item( i ).toElement() );
 			cur_note.setPos( cur_note.pos() + Note::quantized( m_timeLine->pos(), quantization() ) );
 
+			if ((cur_note.length().getTicks() == -DefaultTicksPerBar) && (m_pattern->type() != Pattern::BeatPattern)) 
+			{
+				// the length of copied notes is set to the length of the instruments envelope
+				f_cnt_t envFrames = m_pattern->instrumentTrack()->envFrames();
+				// if notes have zero length due to rounding set the length to a small value
+				tick_t noteLength = qMax(static_cast<float>(1), envFrames / Engine::framesPerTick());
```
Veratil, line 4134: Would `1.0f` work here instead of `static_cast<float>(1)`?

Also, might be good to note that `tick_t` isn't a float/double type, but an integer type.

src/tracks/Pattern.cpp
thmueller64: This is because by default an empty pattern is a `BeatPattern`, see also `Pattern::checkType()`. When I create an instance of 3osc for example and copy notes from a BB Editor to the Piano Roll of the 3osc, it is assumed to be a beat pattern. The notes are not set to the envelope length then. 
I thought this would be more consistent with the behavior introduced, but after thinking twice I'm not sure about that... What do you think? Is it intended that every instrument gets a BB Pattern on the timeline by default?

src/gui/editors/PianoRoll.cpp
thmueller64: Thanks for the review, maybe something like
`tick_t noteLength = static_cast<tick_t>(qMax(1, round(envFrames / Engine::framesPerTick())))`;
would be better?

src/tracks/Pattern.cpp
thmueller64: When a user clicks on "Add steps" and creates a pattern on the added steps, if he then deselects all of the steps in the extended range for a moment, the range is shrunk immediately. I thought this would be a bit annoying, though it looks nicer when steps are moved inside the piano roll or removed and the beat pattern adapts. This would be a one line change essentialy.

src/gui/editors/PianoRoll.cpp
Veratil: I was thinking something like this:
```suggestion
				tick_t noteLength = static_cast<tick_t>(qMax(1.0f, envFrames / Engine::framesPerTick()));
```
With the static cast, we drop the remainder completely. Thinking about this though, we may always want to round up otherwise the note would be cut off from a complete playthrough.
If we always want to round up:
```suggestion
				tick_t noteLength = static_cast<tick_t>(qMax(1.0f, ceil(envFrames / Engine::framesPerTick())));
```

src/tracks/Pattern.cpp
Veratil: With `checkType` always making sure the pattern is set to either `BeatPattern` or `MelodyPattern`, it seems like something we don't have to change since it will change itself.

thmueller64: Alright, I agree. I'll revert this change.

src/tracks/Pattern.cpp
IanCaio: I could be wrong, that's my first glance on the PR. But from what I read, on master `m_steps` is only changed by methods related to the BBTrack. When you have an `MelodyPattern` (adding notes through the piano roll), it seems that the length of the pattern changes but `m_steps` remains the same value. So when you delete the notes and it goes back to being a `BeatPattern`, the number of steps is the one you set earlier on the BBEditor. I didn't test yet but I think with this line you'd end up updating the number of steps as you add notes and once you go back to a `BeatPattern` you'd have the last value you had, not the one you set on the BBEditor. Changing that behavior is not really necessary and IMHO it makes sense to keep `BeatPattern` number of steps separated from the `MelodyPattern` length.

IanCaio:
At first sight, the code looks good, though I'd personally remove `adjustSteps` as I believe it's unnecessary and changes the current behavior of the BBEditor handling of number of steps. I've one concern though: This makes it impossible to paste "step notes" on the Piano Roll. I understand this fixes the mentioned issues, because they are related to the rendering of TCOs and the PianoRoll loop not handling negative length notes, but that also removes the possibility of adding a "step note" on the Song Editor to play a sample until its end, instead of having to adjust the length manually. That's probably rarely used at the moment, but I looked forward for us to handle this kind of notes differently so they could be used in the future outside the BBEditor. Maybe adding a `m_type` variable to the `Note` class which would define whether the note is played until the end of the sample or until the end of the note. That would probably require much more changes to the code and we also are running towards the refactor so it's probably something that will have to wait, thus if everyone agrees that this is a suitable solution for now I'm good with it. I'd like that we add a `TODO` comment though, so we know that if we change the way we handle this kind of note we will have to remove that block of code from `PianoRoll::pasteNotes`.

thmueller64:
Close in favor of https://github.com/LMMS/lmms/pull/5902 .

PR #5509, opened 2020-05-20
Fix #5504: invalid warning

### LINKED ISSUE #5504: Invalid warning about "dummy-MIDI-client"
### Bug Summary

LMMS is a digital audio workstation ([DAW](https://en.wikipedia.org/wiki/Digital_audio_workstation)). In order to control how audio is being played (e.g. to send notes to LMMS), the [MIDI](https://en.wikipedia.org/wiki/MIDI) protocol can be used. This enables you to connect a virtual or real keyboard to LMMS and play music.

There are different libraries that allow LMMS to use MIDI. LMMS uses these libraries to start "MIDI clients". Of those, LMMS currently supports ALSA, OSS and sndio.

You can also select a dummy client if you don't need MIDI. A dummy client is also created if a desired MIDI client could not be started.

Now, starting LMMS with dummy MIDI client (explicitly selected) warns you:

```
Couldn't create MIDI-client, neither with ALSA nor with OSS. Will use dummy-MIDI-client.
```

This is wrong/misleading:

1. There is also sndio MIDI, not only ALSA and OSS.
2. LMMS did not try to create a MIDI client.

#### Steps to reproduce

Enable dummy MIDI client in the settings (MainWindow Menu Bar -> Edit -> Settings -> MIDI). Close LMMS. Start LMMS. You'll see the warning in the command line.

#### Expected behavior

No output at all in case dummy client was explicitly selected. Warning if a "real" MIDI client was selected, but could not be created.

No mentioning of ALSA or OSS.

#### Actual behavior

Output in all cases, mentioning only ALSA and OSS (but not sndio).

#### Affected LMMS versions

All.

### Solution hints

* Look at `src/core/Mixer.cpp`.
* The same problem is fixed correctly in `Mixer::tryAudioDevices`.
* Submit a pull request (PR) against `master`. (we may change the base, but don't worry about that for now)

thmueller64:
Fixes the invalid warning about "dummy-MIDI-client" mentioned in #5504. The implementation is similar to `Mixer::tryAudioDevices`, like mentioned in the issue, but prints the precise device name that failed.

### DIFF
--- a/src/core/Mixer.cpp
+++ b/src/core/Mixer.cpp
@@ -1203,8 +1203,18 @@ MidiClient * Mixer::tryMidiClients()
     printf( "midi apple didn't work: client_name=%s\n", client_name.toUtf8().constData());
 #endif
 
-	printf( "Couldn't create MIDI-client, neither with ALSA nor with "
-		"OSS. Will use dummy-MIDI-client.\n" );
+	if(client_name != MidiDummy::name())
+	{
+		if (client_name.isEmpty())
+		{
+			printf("Unknown MIDI-client. ");
+		}
+		else
+		{
+			printf("Couldn't create %s MIDI-client. ", client_name.toUtf8().constData());
+		}
+		printf("Will use dummy-MIDI-client.\n");
+	}
 
 	m_midiClientName = MidiDummy::name();
 


Spekular:
Tested:

- ALSA Sequencer (working): Neither version prints any warnings
- ALSA Raw (failing): Both versions print three lines of warnings, but this PR is more specific on the third
  - Master: `Couldn't create MIDI-client, neither with ALSA nor with OSS. Will use dummy-MIDI-client.`
  - PR: `Couldn't create ALSA Raw-MIDI (Advanced Linux Sound Architecture) MIDI-client. Will use dummy-MIDI-client.`
- Dummy: Master prints a generic warning, this PR prints no warning

So, it makes the error message more specific, and gets rid of it when irrelevant. Seems to fix the issue from what I can tell. Code LGTM as well.

JohannesLorenz:
Thanks for fixing and reviewing. As this is just a few `printfs` and some `if`/`else`, is everyone OK if we rebase this to `stable-1.2`?

JohannesLorenz:
I think you force-pushed right, but we probably still need to tell github that we want to merge into `stable-1.2` (by clicking the "edit" button on top right of your PR).

JohannesLorenz:
CI was passed, see draft PR #5514 . Merging now.

PR #5512, opened 2020-05-21
Better minimum length when resizing notes

Spekular:
- If the next step down would lead to a length of <= 0, prevent the resize instead of creating a 1 tick long note.
- If multiple notes are selected, the smallest minimum length applies to all of them, so note lengths can be made equal by scaling down to the minimum, releasing, then scaling up.
- Alt overrides the minimum length (sets it to one tick) so any length can be set when fine adjusting length.

Based on conversation in #3998, should make #3989 less common by not creating 1 tick notes as easily. Should probably be squashed on merge?


src/gui/editors/PianoRoll.cpp
```diff
@@ -2806,13 +2816,14 @@ void PianoRoll::dragNotes( int x, int y, bool alt, bool shift, bool ctrl )
 		else
 		{
 			// shift is not pressed; stretch length of selected notes but not their position
+			int minLength = alt ? 1 : m_minResizeLen.getTicks();
```
Veratil, line 2819: How does a length of 1 handle move and resize actions?

Spekular: If alt is held we want to allow notes to be scaled to any length, so 1 tick is the minimum.

zonkmachine: > How does a length of 1 handle move 

I don't know if this varies with screen but on my small one it's not possible to get a hand icon when hovering over a 1 tick note.

> and resize actions?

It readily resizes like other notes but you can't do it on the smallest zoom levels.

Veratil: > 
> 
> > How does a length of 1 handle move
> 
> I don't know if this varies with screen but on my small one it's not possible to get a hand icon when hovering over a 1 tick note.

I recall the minimum draw size should be like 4 pixels... Hm.


> 
> > and resize actions?
> 
> It readily resizes like other notes but you can't do it on the smallest zoom levels.

Good to know.

zonkmachine: > I recall the minimum draw size should be like 4 pixels... Hm.

Top: beat note
Bottom: 1 tick note
beatnotevs1tick

zonkmachine: https://github.com/LMMS/lmms/blob/c3e056a21df4601f1d537b0719c42eff523c9935/src/gui/editors/PianoRoll.cpp#L2126-L2129

Set  `else if( len_ticks < 0 )` to  `else if( len_ticks < 4 )` to make 1 tick notes same length as beat notes. It won't improve ability to grab and stretch them however. It's also not related to this PR so </OT>  ;)

Spekular:
Note: need to integrate #5515 into this.

zonkmachine:
@Spekular My commits in here have their own story in #3998. I think you should manually squash the PR down to one commit and if the last one belongs to me, change it to you. `git commit --amend --author="Author Name <email@address.com>"`. My contributions in this one is maybe naming a variable.

Spekular:
@zonkmachine ok! Anything else to do once that's done, or do you think this is good to merge?

src/gui/editors/PianoRoll.cpp
```diff
@@ -1772,6 +1773,15 @@ void PianoRoll::mousePressEvent(QMouseEvent * me )
 					// then resize the note
 					m_action = ActionResizeNote;
 
+					m_minResizeLen = quantization();
+					for (Note *note : getSelectedNotes())
+					{
+						int thisMin = note->oldLength() % quantization();
```
Spekular, line 1779:
```suggestion
						if (note->oldLength() <= 0) { note->setOldLength(1); }
						int thisMin = note->oldLength() % quantization();
```
Integrates #5515. Since we already iterate through the selection here we may as well fix this length instead of checking on every drag event.

zonkmachine:
OK. stable-1.2 is merged into master and now there are conflicts to sort out.

Spekular:
I'll try to take care of them today, otherwise tomorrow.

Spekular:
For anyone who wants to do a code review, this diff shows the actual changes: https://github.com/LMMS/lmms/pull/5512/commits/171288a03d59bdf45fa5f83c5beeafc3d45ecdd7

Spekular:
Merge? The conflicts are fixed since a week+ ago.

zonkmachine:
I've been testing this on and off and I'm a bit unsure what is best here. As soon as you have a 1 tick component in there the whole selection will have a default minimum of 1 tick. Like Its been touched by evil. It's still better behavior than before. I'm not against merging though but I still like my way more.

Spekular:
> As soon as you have a 1 tick component in there the whole selection will have a default minimum of 1 tick.

This was introduced to my solution by your request to have the same minimum length for all notes. When I had the minimum calculated per note this wouldn't occur, but dragging as far left as possible could create notes with different lengths unless alt is held (which I don't mind, personally). I'd gladly revert that aspect if you prefer?

Spekular:
Or perhaps we need opinions from more people than just us two, haha.

zonkmachine:
> same minimum length for all notes

Yes, but apart from 1 tick. As you may have noticed I'm a bit sensitive when it comes to those ones.


> Or perhaps we need opinions from more people than just us two, haha.

The thought has crossed my mind.

Spekular:
> Yes, but apart from 1 tick. As you may have noticed I'm a bit sensitive when it comes to those ones.

Ah. I think this is only really solveable if we (1) disallow beat notes from being resized or (2) reimplement them in a less hacky way. The moment we resize a beat note, we have to give it a positive length so it can properly resize. Once we've done that, we can't distinguish it from a regular note, and as such it would have to be accounted for when calculating a minimum.

Also worth noting that longer note with a few ticks tail (ex. 1/4 and 4 ticks) will lead to the same scenario of the minimum length being a few ticks long.

Smallest change in beat note implementation I can see solving the problem is an "isBeatNote" boolean set to true for beat notes, and a "convertBeatNote" (questionable name) function. Then we let them have a length of 4 ticks instead of a negative length, replacing all checks for length < 0 with isBeatNote. Resizing happens as usual, but on mouse release convertBeatNote is called. If the beat note has a different length now, it's converted to a regular note and isBeatNote becomes false. If it's still 4 ticks long we treat it as a beat note still. This would also allow us to, for example, snap beat notes to Q length when they first get sized up, to avoid a weird length of Q + 4 ticks.

Spekular:
Before this goes completely stale, I'll merge, citing the following.

> It's still better behavior than before. I'm not against merging though but I still like my way more.

Issue #5287, opened 2019-10-29
MIDI export exports everything as piano, making things out of tune


musikBear:
The attached project was given to me, because it could not be exported.
I have tried to export with 1.2.1 
It exports 51kb, and cuts off almost exactly one mins.
I have shaved it down to minimum, and added a control beeper at project end.
I also had to add an envelope to watsyn, due to clicks and pops
[crash.zip](https://github.com/LMMS/lmms/files/3784388/crash.zip)

( I have NO idea what this music is called...

I then took the demo 'Kaakkuja' and midi-exported that
[Greippi - Krem Kaakkuja (Second Flight Remix).zip](https://github.com/LMMS/lmms/files/3784622/Greippi.-.Krem.Kaakkuja.Second.Flight.Remix.zip)
Its is fully exported, but completely broken

PhysSong:
I can't see any issues with the MIDI export. @musikBear Could you provide more details?

musikBear:
> I can't see any issues with the MIDI export.

Kaakkuja?!
Cant you hear that it is complete out of synch and tuning? What midi-player do you have? Mine is Anvil on Windows64.
If that project sound fine on your midi.player it would be great, because then Anvil have the problem, not LMMS!
This is what i can hear
[Kaakkuja.zip](https://github.com/LMMS/lmms/files/3799019/Kaakkuja.zip)
Pretty much unbearable :[ @PhysSong 
**Also** -I cant **reimport** the MIDI-file *into* LMMS at all! Try that too!

How much of the project crash.zip could you export?

SecondFlight:
Hi, I'm the author of the second song, and I can confirm everything there is working as expected.

The issue is that everything is exporting as piano, which includes drums. I'm using AFP for most of the drums in that song, and converting the drum notes directly to piano creates a bunch of awful off-tune stuff. It sounds to me like the non-drum notes are coming out correctly.

TL;DR: if the job of the MIDI export is to convert all piano roll stuff to actual piano, then it's working just fine.

Here's the song for reference:
https://soundcloud.com/second_flight/krem-kaakkuja-remix

coderman64:
[...]

Issue #5198, opened 2019-09-19
More control over MIDI export?


atirutw:
I am making a cover of a game music in all-sf2 player but I also wanted to export the song to MIDI. Unfortunately this feature put everything on one MIDI channel which is not good for a music that's not exclusively piano. Is there any plan to improve the MIDI export feature?

musikBear:
"Duplicate" but i cant find it. Its not an enhancement. It is a serious export-bug!
Confirming as bug! (LMMS1.2.0 win10-64) 
LMMS MIDI export throws every exported MIDI channel into one track when re-imported
Import this: 
[Greippi - Krem Kaakkuja (Second Flight Remix).zip]
(https://github.com/LMMS/lmms/files/3636857/Greippi.-.Krem.Kaakkuja.Second.Flight.Remix.zip)
Everything ends up in Track0
It btw also sounds HORRID in **Anvil**
Can anyone try any other MIDI programs, like MUSEscore?

WattanaGaming:
> is a serious export-bug!

I thought the devs doesn't bother to make the option to control the exporting process

SecondFlight:
LMMS doesn't only work with real instruments, so what would be the expected behavior in that case (exporting the remix)? Adding the ability for someone to set a MIDI output channel and instrument would resolve the issue, but it would be adding functionality, not fixing existing functionality. I think that's the reasoning for calling it an enhancement instead of a bug.

Also, @WattanaGaming: in case you didn't know or intend it this way, the wording here comes across as negative, as if we should have done it right but were just lazy:

> I thought the devs doesn't bother to make the option

Please keep in mind that we're all volunteers and are developing and maintaining free software.

musikBear:
@SecondFlight 
> expected behavior in that case (exporting the remix)

Definitively not jamming all tracks into track0

> I think that's the reasoning for calling it an enhancement instead of a bug.

But surely if you try and import that MIDI-file, you see the bug?!

SecondFlight:
It's a semantic difference - the difference between *desired* behavior and *expected* behavior. This feature works as the original implementer intended, so it's expected behavior. But it doesn't work as well as is optimal, so it's not desired behavior. Either way, it's still a problem worth fixing.

WattanaGaming:
Why is this issue dead tho

SecondFlight:
[...]

PR #5516, opened 2020-05-24
Add MidiExport support for different track channels, default SF2 patches and drum mapping

EmoonX:
Hello, folks! This is my first pull request to LMMS and probably the first serious one to GitHub as a whole. Hope it could be the start of a safe journey.

This should solve issues #5198 and #5287 (which seem like duplicates?).

## Issues

**MidiExport** plugin currently exports every instrument track (be it one from song editor or BB editor) to a single channel (*0*) with no program (patch) change event. In other words, everything is always merged to a single track/channel with GM Grand Piano (*0*) patch by default when exporting to the .mid file format.

The loss of information is a barrier to collaborate with non-LMMS using people, or even to work with MIDI between different programs or electronic keyboards (since, even though LMMS correctly imports to different tracks and patches using the default .sf2, all goes wrong if you simply try next to MIDI export everything; which, ultimately, led me to work on this PR).

## Enhancement

The PR introduces the following modifications:
+ Each instrument track from song editor is written to a different incremental MIDI channel (skipping channel *10*; see below)
+ In case of the current instrument track being an Sf2 Player one, a program change event valued to the current SF2 patch (*0* to *127*) is added to the MIDI file track, signaling the GM instrument to be imported further
+ BB editor instrument tracks are saved exclusively to channel *10* with patch *0*, using the repeated pattern currently generated from the BB editor. MIDI note pitch (which corresponds to a different drum sound following GM percussion map) is extracted from the BB instrument tracks and will only make sense if using Sf2 Player for these (which will also need some code tweaks in MidiImport to get a decent result)

I also did some general refactoring and documentation on the MidiFile/MidiExport modules, both to help myself personally understanding the code and to improve readability and reusing for everyone.

## What more?

General ideas to further improve results and workflow:
+ Add a "default instrument / preset" custom mapping for each patch number. This is an interesting standalone idea as many times we end up working with another bank, another .sf2 file or even a preset from another plugin altogether. Every default preset could have a custom patch number, which would then be used both for exporting and importing. For example, suppose a cool TripleOscillator saw wave custom preset could be saved as the default for GM Saw Wave (*81*): MidiExport would pick up that info and write a *81* program change to the file, and MidiImport would read that *81* and load the designated plugin and preset (instead of default Sf2 Player and patch number). In other words, this work wouldn't be restricted anymore to only soundfonts.
+ Smartly import the channel 10 track as a pure BB track. MidiImport currently just imports the drums track as a normal (repeating) instrument one. It would be nice to implement some algorithm to build a BB track from this repeated info; that way, it could be instantly exported with no loss of drum information (the way it is with the PR, everything would be lost as only the BB editor ones are written to channel 10; another imported drum one would be just exported as piano...)
+ Do a workaround when there's more than *16* (or *15* normal) channels

---

Comments and suggestions are more than welcome!

General MIDI reference: http://www.music.mcgill.ca/~ich/classes/mumt306/StandardMIDIfileformat.html


Spekular:
Based on the PR description (which is pretty well written!) this looks like a nice improvement.

However, I don't believe BB tracks should export to channel 10, at least not unconditionally. For starters, even if we assume that the BB editor only contains simple patterns and drums, we can't know what each instrument in the editor is supposed to be (which track is the snare? Which is the kick?). Secondly the BB editor can contain melodies, just like any other track. Channel 10 isn't suitable to export patterns in that case.

If the BB -> Channel 10 conversion is kept I think it ought to be as an option set at export time (similar to how render settings are set now).

EmoonX:
@Spekular Thanks! And indeed, you've got a point.

The PR *does* make it possible to retain drum info if you use solely Sf2 Player instruments in BB editor (and by changing the center note in each one so it matches the desired drum sound to be repeated). The good part is that MIDI file specification allows saving more than one track per channel, so MidiImport could be tweaked in a further PR to retain all the original separated BB tracks info from channel 10.

Yet, I totally agree that drums shouldn't be solely attributed to BB tracks and vice versa. An easy fix would be simply picking up all *bank=128 / patch=0* tracks (be them from song or BB editor), merging them to channel 10, and leaving the other BB tracks untouched. That way the MIDI import/export cycle would lose no info, and this PR would be somewhat complete even without the extra MidiImport tweaks I mentioned earlier.

I think this way should be straightforward and intuitive enough to be implemented without an explicit export option, but do feel free to discuss further.

plugins/MidiExport/MidiExport.cpp
```diff
@@ -23,338 +23,338 @@
...
-		mnote.duration = note.attribute("len", "0").toInt();
-		pat.push_back(mnote);
+		class Instrument *inst = instTrack->instrument();
+		uint8_t bank = inst->childModel("bank")->value<uint8_t>();
+		if (bank == 128)
+		{
+			// Drum Sf2 track, so set its channel to 10
+			// (and reverse counter increment)
+			midiTrack.m_channel = 9;
+			m_channel--;
```
PhysSong, line 230: If `m_channel` was 9 before calling this function, this line will change it to 10 instead of going back to 9.

plugins/MidiExport/MidiExport.cpp
```diff
@@ -23,338 +23,338 @@
...
+	midiTrack.addName(track->name().toStdString(), 0);
+
+	// If the current track is a Sf2 Player one, set the current
+	// patch to the exporting track. Note that this only works
+	// decently if the current bank is a GM 1~128 one (which would be
+	// needed as the default either way for successful import).
+	// BB tracks are always bank 128 (see MidiImport), patch 0.
+	uint8_t patch = 0;
+	QString instName = instTrack->instrumentName();
+	if (instName == "Sf2 Player")
```
PhysSong, line 221: I think this check should belong to `Instrument` or `InstrumentTrack` in the future. Could you add a comment which addresses it?

PhysSong:
@EmoonX Pinging you in case you've missed this one.

firewall1110:
I have compiled this branch on Debian Buster using clang (without VST and Carla). 
I have tested it in simple way (without drums and BB and with <10 channels). 
Exported mid-file understands MuseScore2 and CubaseSX 2.0 (see channels and Program Changes; Muse Score provides correct instrument names - it is good improvement). 

P.S.

Program Change inserting sometimes is good (MuseScore importing), but mostly after importing mid-file with Cubase I have to manually delete Program Changes. Some "configuration check box" would be needed in this case ...

[I am new in testing ...]

dacanizares:
Any news?

JakeThe28th:
Any news?

RiedleroD:
I believe this PR is dead, not only bc the original author isn't responding, but also because, with the recent cascade of internally breaking changes, this would be a nightmare to merge. I think completely redoing this PR wouldn't be a crazy idea.

xiiixvv:
I apologize for necroposting, but I was curious on if there as been any more work done towards improving the MIDI stuff? I was wanting to try using LMMS for making custom music for DOOM Mods, but am extremely disappointed to see that everything is exported to just sound like a grand piano. If there hasn't then that's understandable to an extent, I just am interested on any progress or attempts made after the last few posts.

Rossmaxx:
If you haven't already, try nightly. Midi support has improved quite a bit. Any further improvements are unpredictable as we are having a dev crunch and most active devs including myself are clueless about the midi standard.

EmoonX:
Got taken by surprise, seeing a notif for this after all these years. Long story short, I opened this PR back then as a way to improve my `LMMS <-> musical keyboard` workflow at the time, but... soon disinterest (and some personal issues) kicked in, and I moved to Bitwig Studio a few months later anyway. I deeply apologize to everyone who didn't get to witness this large amount of new code seeing the light of the day.

As it appears to be a requested feature still, I'd love to tackle it again at some point, yet most of the code and concepts involved look like magic runes to me at this point, not to mention the obscene number of merge conflicts that are bound to happen. It's worth a shot just for the sake of it, but no guarantees are given.

xiiixvv:
> If you haven't already, try nightly. Midi support has improved quite a bit. Any further improvements are unpredictable as we are having a dev crunch and most active devs including myself are clueless about the midi standard.

I remember getting a nightly build for the slicer plugin but havent updated in a bit. What should I expect from the midi support now?

> Got taken by surprise, seeing a notif for this after all these years. Long story short, I opened this PR back then as a way to improve my `LMMS <-> musical keyboard` workflow at the time, but... soon disinterest (and some personal issues) kicked in, and I moved to Bitwig Studio a few months later anyway. I deeply apologize to everyone who didn't get to witness this large amount of new code seeing the light of the day.
> 
> As it appears to be a requested feature still, I'd love to tackle it again at some point, yet most of the code and concepts involved look like magic runes to me at this point, not to mention the obscene number of merge conflicts that are bound to happen. It's worth a shot just for the sake of it, but no guarantees are given.

Completely up to you if you want to do it or not, programming is a really hard thing at times. Would love to see more improvements to LMMS in general, since it's development seems extremely slow (if I knew a thing in CPP or two I would try and chip in myself). In the mean time, I transferred some of my work to MidiEditor.

Rossmaxx:
>> If you haven't already, try nightly. Midi support has improved quite a bit. Any further improvements are unpredictable as we are having a dev crunch and most active devs including myself are clueless about the midi standard.

> I remember getting a nightly build for the slicer plugin but havent updated in a bit. What should I expect from the midi support now?

The improvements stopped before SlicerT came along, so whatever is there for you rn, that's it. I was comparing nightly against stable.

bratpeki:
This looks cool, and a user and dev showed interest in this, so I could test it as well!

> Do a workaround when there's more than 16 (or 15 normal) channels

Could you explain the workaround?

messmerd:
This PR was incredibly out-of-date and missing its original fork/branch, so I've rebased it and opened a new PR: #8048

Issue #2080, opened 2015-06-03
New arpeggiator functions


zonkmachine:
Yo!

I'm updating the source from pull request https://github.com/LMMS/lmms/pull/687 and am preparing a brand new pull request. It still lacks a proper GUI though and I'm looking into it.
The functionality we need here is similar to what's going on in the plug-in 'Mallets', which switches layout depending on which instrument you select.

![arpeggiator](https://cloud.githubusercontent.com/assets/6368949/7944715/b34c4bda-096c-11e5-85f9-ac9558112307.png)

>REP / Repeat
Note repeats this number of times.

>CYCLE
Will skip over this number of notes, spreading them out over the range. It might get stuck and bounce between just a couple of notes if the length of the range / jump length favors this. Not a bug. :-)

>SCR / Scramble
Not implemented. I was thinking of adding the function to dial in the start key.

>SKIP
Dial in the probability that the note will be skipped altogether.

>MISS
Dedides the probability that the note will be a random one instead.

>FLOOR
All notes under this will be assigned the floor / lowest note instead.

>CEIL / Ceiling
Same as floor but sets upper limit. In the event you set FLOOR higher than CEIL, the later function, CEIL has precedence and sets the note.

About waiting to 2.0 
The algorithms are pretty straight forward and will be easy to reimplement.

Thoughts?

Wallacoloo:
Huh. This is an interesting idea. I don't actually use arpeggiators when sequencing. But I do *occasionally* use them when doing improv stuff in private.

I really can't judge how useful these features will be before trying them (and I would definitely fiddle around for them a bit once they're ready). It seems that they could be really great for certain niches though. My only concern is that it will make the interface look more complex to anyone not yet familiar with these features. Therefore it might be worth adding an "Advanced mode" checkbox (default disabled) that shows or hides these controls, or something along those lines.

To my knowledge, we are indeed in a feature freeze. But this feature seems totally reasonable for the next minor version (1.3).

tresf:
> Will skip over this number of notes, spreading them out over the range. It might get stuck and bounce between just a couple of notes if the length of the range / jump length favors this. Not a bug. :-)

This one is the big win in my opinion per #1836.

Related #1298 #340 #687 (previously linked)

> My only concern is that it will make the interface look more complex to anyone not yet familiar with these features. Therefore it might be worth adding an "Advanced mode" checkbox (default disabled) that shows or hides these controls, or something along those lines.

[...]

PR #5521, opened 2020-05-30
New arpeggiator functions - Floor and Ceiling

zonkmachine:
Some six years later and I'm still revisiting this one from time to time. First proposed in https://github.com/LMMS/lmms/pull/687
Implements functions 'Floor' and 'Ceiling' proposed in https://github.com/LMMS/lmms/issues/2080

I made the maxValue of floor and ceil update when the arpeggio model or the range is changed. I also made the ceil value stick to maxValue, if it is at maxValue and the maxValue change, to keep up with the maxValue changing all the time. Those were the biggest issues to solve the last time I touched this code. The last issue is concerning a better gui solution and this is still unsolved.

I have also upgrade the Cycle knob to follow the range.

To do

 - [ ] Fix upgrade path for cycle function

---

arplayout

### DIFF
--- a/include/InstrumentFunctionViews.h
+++ b/include/InstrumentFunctionViews.h
@@ -81,6 +81,7 @@ private:
 	Knob * m_arpCycleKnob;
 	Knob * m_arpSkipKnob;
 	Knob * m_arpMissKnob;
+	Knob * m_arpFloorKnob;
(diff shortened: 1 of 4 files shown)


zonkmachine:
I'm closing this for the time being.

PR #5523, opened 2020-05-31
Arpeggiator - Cut off trailing short notes

zonkmachine:
Prevent triggering of extra notes at the end by not counting trailing notes shorter than one fifth of the notes arp_frames. This value being arbitrarily found by trial and error.

Addresses one of the tickets (Extra notes) in https://github.com/LMMS/lmms/issues/3880
The picture in 3880 also demonstrates this fix in the lower stereo track where the duplicated notes (the bigger notes) are gone but the missing notes (another, separate issue) remains.

arpagain

### DIFF
--- a/src/core/InstrumentFunctions.cpp
+++ b/src/core/InstrumentFunctions.cpp
@@ -389,8 +389,10 @@ void InstrumentFunctionArpeggio::processNote( NotePlayHandle * _n )
 	while( frames_processed < Engine::mixer()->framesPerPeriod() )
 	{
 		const f_cnt_t remaining_frames_for_cur_arp = arp_frames - ( cur_frame % arp_frames );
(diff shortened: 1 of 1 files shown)


tresf:
@zonkmachine this is a one-liner, is there a reason why it hasn't been merged?

zonkmachine:
@tresf I've been busy at work. I'll have to look into this again.

zonkmachine:
I think it's fine to merge but theoretically it could make your old project skip a note at the end if you have already tweaked the note length to avoid this issue but have set it to cut off precisely after the last arp note goes on. I think this will happen in precisely 0 cases but you never know. I don't know if there's a better way to do this than an arbitrary small number like this.

zonkmachine:
I'll test it some more and then merge in a couple of days if there are no objection.

PR #5524, opened 2020-06-02
Knife Tool for Sample Clips (Again)

### LINKED ISSUE #5394: Song editor mode sticks if modal is opened while holding control
3/5, more of a issue than a bug.
### Issue Summary
- clicking command in the Song Editor toggles draw and edit mode, really annoying when trying to export and the piano roll swaps to edit mode.

#### Steps to reproduce
Focus on the Song Editor and hit command. prolly works with control too on windows

#### Expected behavior
nothing happens.
#### Actual behavior
song editor toggles to edit mode until command key is released. or sometimes, when exporting for example, it locks to edit mode until command is hit again.
#### LMMS version used
1.2.1.581

Spekular:
New version of #4998, with copy constructors for SampleBuffer and SampleTCO and a scalpel cursor when hovering over sample tracks in Knife/Split mode. Also adds copy assignment operator for SampleBuffer and a comment about this being missing in SampleTCO.h.

It's worth noting that this PR currently creates a new sample buffer with each cut, because apparently sample clips rely on having a unique buffer. This means that memory use will increase by the samplebuffer size with each cut. Cloning clips currently behaves the same way, so this isn't a regression, but I believe it eventually needs to be fixed as this leads to performance issues.

TODO:
- [x] Add a TODO comment about the sample buffer issue


thmueller64:
Tested, works great!

src/gui/TrackContainerView.cpp
```diff
@@ -305,6 +305,14 @@ bool TrackContainerView::allowRubberband() const
 
 
 
+bool TrackContainerView::knifeMode() const
+{
+	return false;
+}
```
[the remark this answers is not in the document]
Spekular: This is inside the generic TrackContainerView class, so all tracks return false by default. Then, any track that *can* be split overrides the function.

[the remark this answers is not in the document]
russiankumar: Oh, alright then. I'll resolve this.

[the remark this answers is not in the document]
russiankumar: Except.... I can't

Spekular:
This should be ready for review now. Generally, it seems we're okay with merging this first and fixing the SampleBuffer issue in a later PR (preferably before 1.3!).

superpaik:
Some issues I have found:
- sometimes, after using the knive and coming back to draw mode (clicking the pencil icon) the cursor still holds the knive. Then you move the cursor around and you get the hand cursor.
- import and audio. Cut it in two. Delete one of the pieces. Hit play and when the play-line arrives at the end of the remaining piece, it crashes (tested on windows 10 with the 64bit mingw generated by the bot). 
-

he29-net:
In my case, I found the crash when I deleted the first part and the second was about to start playing. I re-tested it now with different samples and it seems to be length-dependent: shorter samples sometimes fail the way superpaik described, and sometimes they do not fail at all.

Looking briefly at the code, the only idea I have is that when calling
`SampleTCO * rightTCO = new SampleTCO ( *leftTCO );`
the constructor does not copy the sample buffer buried inside (that may be intentional), but just assigns it. And when one TCO is deleted, it deletes the sample buffer as well. And it crashes when the remaining TCO tries to touch its sample buffer, since it points to deleted memory.

Spekular:
> Looking briefly at the code, the only idea I have is that when calling
> `SampleTCO * rightTCO = new SampleTCO ( *leftTCO );`
> the constructor does not copy the sample buffer buried inside (that may be intentional), but just assigns a pointer. And when one TCO is deleted, it deletes the sample buffer as well. And it crashes when the remaining TCO tries to touch its sample buffer, since it points to deleted memory.

This shouldn't be the case, as I introduced a copy constructor to SampleBuffer (that's where the branch name ROT comes from, Rule of Three). I'll see if I can reproduce these issues with latest master merged in, because so far in my testing everything has worked fine. Either way I might leave this PR stale for a bit as I intend to break it into smaller components that can be reviewed and merged one at a time.

he29-net:
I meant the `SampleTCO::SampleTCO()` constructor; it basically just does:
```
	m_sampleBuffer = orig.m_sampleBuffer;
	m_isPlaying = orig.m_isPlaying;
```
and since `m_sampleBuffer` is just a pointer, all it does is copy one number (the pointer value), there is no `SampleBuffer` constructor involved. Or am I missing something?

Spekular:
> I meant the `SampleTCO::SampleTCO()` constructor; it basically just does:
> 
> ```
> 	m_sampleBuffer = orig.m_sampleBuffer;
> 	m_isPlaying = orig.m_isPlaying;
> ```
> and since `m_sampleBuffer` is just a pointer, all it does is copy one number (the pointer value), there is no `SampleBuffer` constructor involved.

Ah, you're right, that could be it! Thanks for the lead (or basically the whole fix) :)

> Or am I missing something?

No, on the contrary I was the one missing things. I guess I assumed this would be fixed by the copy assignment function (which IIRC I implemented as delegating to the copy constructor, or vice versa) but I suppose that won't apply to pointer assignments?

PhysSong:
@Spekular I think you can resume this because #5727 is now merged.

Spekular:
Yes! I'm planning on fixing the merge conflicts soon, not sure if there's much to do after that.

Spekular:
Merged, maybe. Had to manually copy a bunch of stuff since the split of Track.cpp got merged before this, and handle that + MidiTime -> TimePos + my own SampleBuffer changes in one go, so I wouldn't be surprised if there are some lingering issues :upside_down_face:

src/gui/TrackContentObjectView.cpp
```diff
@@ -106,6 +109,7 @@ TrackContentObjectView::TrackContentObjectView( TrackContentObject * tco,
 	setAttribute( Qt::WA_OpaquePaintEvent, true );
 	setAttribute( Qt::WA_DeleteOnClose, true );
 	setFocusPolicy( Qt::StrongFocus );
+	setCursor( m_cursorHand );
 	setCursor( QCursor( embed::getIconPixmap( "hand" ), m_mouseHotspotHand.width(), m_mouseHotspotHand.height() ) );
```
PhysSong, line 113: Wrong merge:
```suggestion
	setCursor( m_cursorHand );
```

PhysSong:
You can also optionally revert some whitespace changes if you want.

src/gui/TrackContentObjectView.cpp
```diff
@@ -918,12 +980,51 @@ void TrackContentObjectView::mouseReleaseEvent( QMouseEvent * me )
...
+
+				rightTCO->movePosition( splitPos );
+				rightTCO->changeLength( m_initialTCOEnd - splitPos );
+				rightTCO->setStartTimeOffset( leftTCO->startTimeOffset() - leftTCO->length() );
+
+				leftTCO->getTrack()->restoreJournallingState();
+			}
+		}
+	}
+
```
IanCaio, line 1027:
```suggestion
	else if (m_action == Split)
	{
		SampleTCO * leftTCO = dynamic_cast<SampleTCO*>(m_tco);

		if (leftTCO)
		{
			leftTCO->setMarkerEnabled(false);

			const int relativePixelPos = me->pos().x();
			const float ppb = m_trackView->trackContainerView()->pixelsPerBar();
			TimePos splitPos = relativePixelPos * TimePos::ticksPerBar() / ppb;

			if (!unquantizedModHeld(me))
			{
				splitPos = quantizeMarkerPos(splitPos, me->modifiers() & Qt::ShiftModifier);
			}

			splitPos += m_initialTCOPos;

			// Don't split if we slid off the TCO or if we're on the clip's start/end
			// Cutting at exactly the start/end position would create a zero length
			// clip (bad), and a clip the same length as the original one (pointless).
			if (splitPos > m_initialTCOPos && splitPos < m_initialTCOEnd)
			{
				leftTCO->getTrack()->addJournalCheckPoint();
				leftTCO->getTrack()->saveJournallingState(false);

				SampleTCO * rightTCO = new SampleTCO (*leftTCO);

				leftTCO->changeLength(splitPos - m_initialTCOPos);

				rightTCO->movePosition(splitPos);
				rightTCO->changeLength(m_initialTCOEnd - splitPos);
				rightTCO->setStartTimeOffset(leftTCO->startTimeOffset() - leftTCO->length());

				leftTCO->getTrack()->restoreJournallingState();
			}
		}
	}

```

Code style, but more importantly, I think that part of the code would be better extracted to a method inside `SampleTCO`. Then the mouse event would just do something like:
```
SampleTCO * stco = dynamic_cast<SampleTCO*>(m_tco);
if (stco) { stco->splitTCO(me->pos().x()); }
```

src/gui/TrackContentObjectView.cpp
```diff
@@ -1271,6 +1372,15 @@ bool TrackContentObjectView::mouseMovedDistance( QMouseEvent * me, int distance
 
 
 
+
+bool TrackContentObjectView::unquantizedModHeld( QMouseEvent * me )
```
IanCaio, line 1376:
```suggestion
bool TrackContentObjectView::unquantizedModHeld(QMouseEvent * me)
```

src/gui/TrackContentObjectView.cpp
```diff
@@ -1285,12 +1395,8 @@ TimePos TrackContentObjectView::draggedTCOPos( QMouseEvent * me )
 	TimePos newPos = m_initialTCOPos + mouseOff * TimePos::ticksPerBar() / ppb;
 	TimePos offset = newPos - m_initialTCOPos;
 	// If the user is holding alt, or pressed ctrl after beginning the drag, don't quantize
-	if (    me->button() != Qt::NoButton
-		|| (me->modifiers() & Qt::ControlModifier)
-		|| (me->modifiers() & Qt::AltModifier)    )
-	{
-		// We want to preserve this adjusted offset,
-		// even if the user switches to snapping
+	if ( me->button() != Qt::NoButton || unquantizedModHeld(me) )
```
IanCaio, line 1398:
```suggestion
	if (me->button() != Qt::NoButton || unquantizedModHeld(me))
```

src/gui/TrackContentObjectView.cpp
```diff
@@ -1313,6 +1419,50 @@ TimePos TrackContentObjectView::draggedTCOPos( QMouseEvent * me )
...
+	}
+	else
+	{
+		return (TimePos(midiPos + m_initialTCOPos).quantize( snapSize ) - m_initialTCOPos);
+	}
+}
+
+
+
+
```
IanCaio, line 1465:
```suggestion
int TrackContentObjectView::knifeMarkerPos(QMouseEvent * me)
{
	// Position relative to start of clip
	const int markerPos = me->pos().x();

	// In unquantized mode, we don't have to mess with the position at all
	if (unquantizedModHeld(me)) { return markerPos; }
	else
	{	// Otherwise we...
		// 1: Convert the position to a TimePos
		const float ppb = m_trackView->trackContainerView()->pixelsPerBar();
		TimePos midiPos = markerPos * TimePos::ticksPerBar() / ppb;
		// 2: Snap to the correct position, based on modifier keys
		midiPos = quantizeMarkerPos(midiPos, me->modifiers() & Qt::ShiftModifier);
		// 3: Convert back to a pixel position
		return midiPos * ppb / TimePos::ticksPerBar();
	}
}




TimePos TrackContentObjectView::quantizeMarkerPos(TimePos midiPos, bool shiftMode)
{
	const float snapSize = gui->songEditor()->m_editor->getSnapSize();
	if (shiftMode)
	{	// If shift is held we quantize the length of the new left clip...
		const TimePos leftPos = midiPos.quantize(snapSize);
		// ...or right clip...
		const TimePos rightOff = m_tco->length() - midiPos;
		const TimePos rightPos = m_tco->length() - rightOff.quantize(snapSize);
		// ...whichever gives a position closer to the cursor
		if (abs(leftPos - midiPos) < abs(rightPos - midiPos)) { return leftPos; }
		else { return rightPos; }
	}
	else
	{
		return TimePos(midiPos + m_initialTCOPos).quantize( snapSize ) - m_initialTCOPos;
	}
}




```

Parenthesis spacing and `return` statements must not have parenthesis.

src/gui/editors/SongEditor.cpp
```diff
@@ -944,13 +957,16 @@ SongEditorWindow::SongEditorWindow(Song* song) :
...
+	m_knifeModeAction = m_editModeGroup->addAction(embed::getIconPixmap("edit_knife"), tr("Knife mode (split sample clips)"));
 	m_selectModeAction = m_editModeGroup->addAction(embed::getIconPixmap("edit_select"), tr("Edit mode (select and move)"));
 	m_drawModeAction->setChecked(true);
 
 	connect(m_drawModeAction, SIGNAL(triggered()), m_editor, SLOT(setEditModeDraw()));
+	connect(m_knifeModeAction, SIGNAL(triggered()), m_editor, SLOT(setEditModeKnife()));
 	connect(m_selectModeAction, SIGNAL(triggered()), m_editor, SLOT(setEditModeSelect()));
 
 	editActionsToolBar->addAction( m_drawModeAction );
+	editActionsToolBar->addAction( m_knifeModeAction );
```
IanCaio, line 969:
```suggestion
	editActionsToolBar->addAction(m_knifeModeAction);
```

src/tracks/SampleTrack.cpp
```diff
@@ -634,6 +631,10 @@ void SampleTCOView::paintEvent( QPaintEvent * pe )
 			embed::getIconPixmap( "muted", size, size ) );
 	}
 
+	if ( m_tco->m_marker )
+	{
+		p.drawLine(m_tco->m_markerPos, rect().bottom(), m_tco->m_markerPos, rect().top());
+	}
```
IanCaio, line 637:
```suggestion
	if (m_tco->m_marker)
	{
		p.drawLine(m_tco->m_markerPos, rect().bottom(), m_tco->m_markerPos, rect().top());
	}
```

src/gui/TrackContentObjectView.cpp
```diff
@@ -882,9 +930,23 @@ void TrackContentObjectView::mouseMoveEvent( QMouseEvent * me )
...
 	}
+	else if( m_action == Split )
+	{
+		SampleTCO * sTco = dynamic_cast<SampleTCO*>( m_tco );
+		if (sTco) {
+			setCursor( m_cursorKnife );
+			sTco->setMarkerPos( knifeMarkerPos( me ) );
+		}
+		update();
+	}
```
IanCaio, line 941:
```suggestion
	else if (m_action == Split)
	{
		SampleTCO * sTco = dynamic_cast<SampleTCO*>(m_tco);
		if (sTco) {
			setCursor(m_cursorKnife);
			sTco->setMarkerPos(knifeMarkerPos(me));
		}
		update();
	}
```

IanCaio (changes requested):
Initial review, most of it is related to code style but there are maybe a couple points related to the code itself:

1) As I commented on the suggestion, I think the SampleTCO `split` logic would be better extracted to a method inside the `SampleTCO` class, which would just be called from the mouse event.
2) Not really important, I suggested an alternative name for `unquantizedModHeld` that I think it's slightly more legible.

At first sight I didn't notice anything standing out on the code except for that, but I'd like to take a better look at the `knifeMarkerPos` and `quantizeMarkerPos` later. PhysSong also already mentioned a small mistake left from the merge conflicts resolving.

src/gui/TrackContentObjectView.cpp
Spekular: If it's fine with you I'll extract it, but leave it in `TrackContentObjectView.cpp`. Splitting should really be available for all clips, so eventually all of this code should be shared.

src/gui/TrackContentObjectView.cpp
Spekular: I'll fix the parenthases around return.

Spekular:
I won't do style changes that mismatch surrounding code when the autoformat is going to run soon and fix all that anyway. With those resolved I see that there aren't that many real issues remaining, so hopefully I can get those fixed later today.

src/gui/TrackContentObjectView.cpp
```diff
@@ -662,6 +701,16 @@ void TrackContentObjectView::mousePressEvent( QMouseEvent * me )
 		{
 			remove( active );
 		}
+		if (m_action == Split)
+		{
+			m_action = NoAction;
+			SampleTCO * sTco = dynamic_cast<SampleTCO*>( m_tco );
+			if (sTco)
+			{
+				sTco->setMarkerEnabled( false );
```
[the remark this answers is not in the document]
Spekular: IIRC this is there because there was a bug where RMB would leave the marker pos stuck.

src/gui/TrackContentObjectView.cpp
Spekular: Resolved as described in latest commit.

src/gui/TrackContentObjectView.cpp
Spekular: Parens fixed in latest commit.

src/gui/TrackContentObjectView.cpp
```diff
@@ -882,9 +930,23 @@ void TrackContentObjectView::mouseMoveEvent( QMouseEvent * me )
...
+			sTco->setMarkerPos( knifeMarkerPos( me ) );
+		}
+		update();
+	}
 	else
 	{
 		SampleTCO * sTco = dynamic_cast<SampleTCO*>( m_tco );
+
+		if (sTco && m_trackView->trackContainerView()->knifeMode())
+		{ setCursor( m_cursorKnife ); }
```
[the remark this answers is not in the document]
Spekular: Changed to oneliners in latest commit.

include/SampleTrack.h
```diff
@@ -79,6 +79,15 @@ class SampleTCO : public TrackContentObject
...
 	void setIsPlaying(bool isPlaying);
 
+	void inline setMarkerPos( int x )
+	{
+		m_markerPos = x;
+	}
+	void inline setMarkerEnabled( bool e )
+	{
+		m_marker = e;
+	}
```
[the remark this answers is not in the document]
Spekular: Changed to oneliners in latest commit.

src/gui/TrackContentObjectView.cpp
Spekular: Fixed in latest commit!

src/gui/TrackContentObjectView.cpp
```diff
@@ -1456,8 +1417,54 @@ TimePos TrackContentObjectView::quantizeMarkerPos( TimePos midiPos, bool shiftMo
...
+		return TimePos(midiPos + m_initialTCOPos).quantize( snapSize ) - m_initialTCOPos;
+	}
+}
+
+
+
+
+// This should eventually be available for all clip types, hence why it's not
+// located inside SampleTCOView
+bool TrackContentObjectView::splitTCO( QMouseEvent * me )
```
IanCaio, line 1429: What about making `TrackContentObjectView::splitTCO` a virtual method on `TrackContentObjectView` and override it on `SampleTCOView`? Then if the split action is added to other types of TCOs we can just override them on the respective ones (i.e.: `PatternView`).

I also think it could be useful in the future if we had the relative X position as the parameter of the method call instead of the mouse event. Then we could call `splitTCO` in contexts other than a mouse event (just examples: a keyboard shortcut for splitting, or a built in splicer). The split call would then be `splitTCO(me->pos().x())`.

include/TrackContentObjectView.h
```diff
@@ -226,6 +226,8 @@ protected slots:
 	TimePos draggedTCOPos( QMouseEvent * me );
 	int knifeMarkerPos( QMouseEvent * me );
 	TimePos quantizeMarkerPos( TimePos, bool shiftMode );
+	//! Return true iff TCO could be split
```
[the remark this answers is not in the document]
Spekular: iff = "if and only if". I can change to a regular "if" if you want though, people will probably infer the "and only if" part regardless of spelling.

src/gui/TrackContentObjectView.cpp
Spekular: Virtual is a good idea! I also agree that passing a mouse event feels excessive, but as it stands the method also uses other event properties, like modifers.

include/TrackContentObjectView.h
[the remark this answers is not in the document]
IanCaio: No need! Wasn't familiar with that abbreviation

src/gui/TrackContentObjectView.cpp
IanCaio: True, I missed that the `unquantizationModHeld` and `quantizeMarkerPos` were also using it. You could still move away from using the event directly by having something like `splitTCO(const int xPos, const bool unquantize, const bool quantizeClosest)` then calling it as `splitTCO(me->pos().x(), unquantizeModHeld(me), me->modifiers() & Qt::ShiftModifier)`.

Spekular:
Latest commit makes splitTCO virtual, which turned out to be a pretty invasive change. Pushed anyway for feedback from @IanCaio and @PhysSong, will revert if necessary.

IanCaio:
From a quick review of the last commit the changes look good! It wasn't that invasive, except from moving some properties from `private` to `protected`, which still makes sense for them. The alternative solution would be to have getter methods for them, that's something I'll wait for other devs to give an opinion on. For me having them as `protected` is fine.

IMHO the only change left would be to change `splitTCO` to use the X position instead of the mouse event. If you would like I could try to suggest the necessary changes on a review to make that work.

Spekular:
> IMHO the only change left would be to change `splitTCO` to use the X position instead of the mouse event.

A fully generalized split function would have to take the final split position after quantization*, and at that point we're almost back to having the logic directly in the mouse event. I vote we leave this as is until the functionality is actually expanded, in favor of merging sooner. I might very well be the person to make the change later, because I'm planning to add more split modes if I get to that point in my dev-queue before someone else does.

IanCaio:
> A fully generalized split function would have to take the final split position after quantization*, and at that point we're almost back to having the logic directly in the mouse event. I vote we leave this as is until the functionality is actually expanded, in favor of merging sooner. I might very well be the person to make the change later, because I'm planning to add more split modes if I get to that point in my dev-queue before someone else does.

Some small changes would have to be done on `knifeMarkerPos` and `unquantizeModHeld` so they don't use the mouse event neither, then some changes on `splitTCO`. I thought it would be easier to let it ready for the expansion, but if you have plans already of being the one doing that I'm cool with leaving it for later.

I'll give the code a second review later, but from what I saw from the last commits it should be good.

src/gui/TrackContentObjectView.cpp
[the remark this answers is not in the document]
IanCaio: This still shows as 2 lines on the diff for some reason, did the change get commited?

src/gui/TrackContentObjectView.cpp
```diff
@@ -1313,6 +1380,50 @@ TimePos TrackContentObjectView::draggedTCOPos( QMouseEvent * me )
...
+		//...or right clip...
+		const TimePos rightOff = m_tco->length() - midiPos;
+		const TimePos rightPos = m_tco->length() - rightOff.quantize( snapSize );
+		//...whichever gives a position closer to the cursor
+		if ( abs(leftPos - midiPos) < abs(rightPos - midiPos) ) { return leftPos; }
+		else { return rightPos; }
+	}
+	else
+	{
+		return TimePos(midiPos + m_initialTCOPos).quantize( snapSize ) - m_initialTCOPos;
```
IanCaio, line 1420: Why is it necessary to add `m_initialTCOPos` here and subtract it later? It seems it doesn't get added when `shiftMode` is true but when testing it the quantization worked on that mode too.

src/tracks/SampleTrack.cpp
```diff
@@ -667,6 +668,47 @@ void SampleTCOView::reverseSample()
...
+	m_tco->setMarkerEnabled( false );
+
+	const int relativePixelPos = me->pos().x();
+	const float ppb = m_trackView->trackContainerView()->pixelsPerBar();
+	TimePos splitPos = relativePixelPos * TimePos::ticksPerBar() / ppb;
+
+	if ( !unquantizedModHeld(me) )
+	{
+		splitPos = quantizeMarkerPos( splitPos, me->modifiers() & Qt::ShiftModifier );
+	}
```
IanCaio, line 683: I was thinking here, since we have the position of the marker stored on `m_marker`, couldn't we use it on this method to avoid calling the `unquantizedModHeld` and `quantizeMarkerPos` methods? Since the marker position visually represents the place where the splitting should happen, it should be accurate on the moment this method is called.

That would save a couple method calls and maybe even make it easier to move away from using the mouse event, since that conditional was the trickier part to get around without using the event.

IanCaio (changes requested):

src/gui/TrackContentObjectView.cpp
Spekular: Did you test with an offset clip? `midiPos` is relative to the clip's start position, so `m_midiPos + m_initialTCOPos` gives the absolute position of the cursor in the song. In other words:
- Shift mode quantizes cut position to a grid based on the start and end positions of the clip, choosing whichever is closest to the mouse. This  results in a left and/or right clip with a quantized length.
- Unmodified mode quantizes the cut position to the underlying song grid.

With apologies for the terrible colors, this image shows the unmodified snap positions in red, and shift snap in blue (start point based) and green (end point based)
image

src/gui/TrackContentObjectView.cpp
```diff
@@ -159,7 +162,9 @@ void TrackContentObjectView::update()
 {
 	if( !m_cursorSetYet )
 	{
-		setCursor( QCursor( embed::getIconPixmap( "hand" ), m_mouseHotspotHand.width(), m_mouseHotspotHand.height() ) );
+		m_cursorHand = QCursor( embed::getIconPixmap( "hand" ), m_mouseHotspotHand.width(), m_mouseHotspotHand.height() );
+		m_cursorKnife = QCursor( embed::getIconPixmap( "cursor_knife" ), m_mouseHotspotKnife.width(), m_mouseHotspotKnife.height() );
+		setCursor( m_cursorHand );
```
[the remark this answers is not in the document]
Spekular: > Calling setCursor in the CTOR does not work. The properties are set after the CTOR, so setCursor in the CTOR still uses 0 and 0. So I added a further setCursor into the update function, which is only ever executed once.

https://github.com/LMMS/lmms/issues/4201#issuecomment-626128550

src/tracks/SampleTrack.cpp
Spekular: See comments about this same idea on previous version:

> You'd think that the result from knifeMarkerPos( me ) could be converted back to a MidiTime and used for the actual cut itself, but for some reason the shift/default behavior ends up inverted if you do. Adding a boolean switch to flip the two still left some weirdness, so I moved the code that decides the position back into mouseReleaseEvent.

https://github.com/LMMS/lmms/pull/4998#issuecomment-515694258

https://github.com/LMMS/lmms/pull/4998#issuecomment-515707007

src/gui/TrackContentObjectView.cpp
IanCaio: Oh, makes sense, thanks for explaining!

src/gui/TrackContentObjectView.cpp
```diff
@@ -568,7 +578,35 @@ void TrackContentObjectView::mousePressEvent( QMouseEvent * me )
...
+				m_action = Split;
+				setCursor( m_cursorKnife );
+				sTco->setMarkerPos( knifeMarkerPos( me ) );
+				sTco->setMarkerEnabled( true );
+				update();
+			}
+			// We can't split anything except samples right now, so disable the
+			// action to avoid entering if statements we don't need to. This
+			// also saves us a few 'if(sTco)' checks
+			else { m_action = NoAction; }
```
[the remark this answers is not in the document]
IanCaio: I updated the suggestion, I think it's better to surround the whole block with `if (sTco && !m_tco->getAutoResize())` because it might be a problem to split a resizable TCO (since the splitting itself involves resizing of clips).

src/gui/TrackContentObjectView.cpp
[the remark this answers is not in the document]
IanCaio: Didn't know about that, at first sight it looked redundant. Thanks for clarifying, resolving this!

src/tracks/SampleTrack.cpp
IanCaio: That's odd, I wonder why that didn't behave properly. I know we are in a bit of a schedule, but you mind if I investigate this a little? If I don't figure out soon enough why using `m_marker` didn't work we can go with the current method.

src/gui/TrackContentObjectView.cpp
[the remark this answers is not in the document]
Spekular: I forgot to mention this earlier, but I can't reproduce this issue. Regardless of whether I have knife mode enabled in the song editor or not I can't cut in the BB editor.

src/gui/TrackContentObjectView.cpp
```diff
@@ -568,7 +578,34 @@ void TrackContentObjectView::mousePressEvent( QMouseEvent * me )
...
+			{
+				m_action = ResizeLeft;
+				setCursor( Qt::SizeHorCursor );
+			}
+			else if( me->x() >= width() - RESIZE_GRIP_WIDTH )
+			{
+				m_action = Resize;
+				setCursor( Qt::SizeHorCursor );
+			}
+			else if (sTco)
```
PhysSong, line 595: Can we add a virtual method to check if a TCO supports splitting? Eventually for left resizing as well, but it's out of scope.

Spekular:
Issues with move and resize should be fixed. While I was at it I ended up making box select behave in an (IMO) better and (IME) more standard way. Rather than toggling over to box select mode when control is held, it just starts a box selection when you control + click an empty space, and doesn't visually affect the selected edit mode. On release the edit mode is restored by re-triggering the selected edit button, which seems to be less buggy than the previous implementation based on my quick tests (no more getting stuck in select mode).

PhysSong:
Is the stuck <kbd>Ctrl</kbd> issue #5394?

Spekular:
> Is the stuck Ctrl issue #5394?

Yes. I can still reproduce that issue in 1.3-alpha.1, but not in the current state of this PR.

IanCaio:
On `TrackContentObjectView::mouseMoveEvent()` if there's no action being done the following code runs:

```
		SampleTCO * sTco = dynamic_cast<SampleTCO*>( m_tco );

		if (sTco && m_trackView->trackContainerView()->knifeMode())
		{ setCursor( m_cursorKnife ); }


		if( ( me->x() > width() - RESIZE_GRIP_WIDTH && !me->buttons() && !m_tco->getAutoResize() )
		||  ( me->x() < RESIZE_GRIP_WIDTH && !me->buttons() && sTco && !m_tco->getAutoResize() ) )
		{
			setCursor( Qt::SizeHorCursor );
		}
		else
		{
			leaveEvent( NULL );
		}
```

There are two issues I noticed with this block:
1) If you are on the Knife mode there's some "flickering" of cursor changes when hovering the edges, because both `if (sTco && m_trackView->trackContainerView()->knifeMode())` and `if( ( me->x() > width() - RESIZE_GRIP_WIDTH && !me->buttons() && !m_tco->getAutoResize() ) ||  ( me->x() < RESIZE_GRIP_WIDTH && !me->buttons() && sTco && !m_tco->getAutoResize() ) )` are true, so you set the cursor to the `m_cursorKnife` and right after you set the cursor to `Qt::SizeHorCursor`.

2) If you are on the Knife mode, hovers the mouse on the middle of the sTCO and leave it, switch to the Draw Mode and hover again in the middle of the sTCO you'll end up with the knife cursor. That's because when `leaveEvent` is called, the cursor of the widget is still `m_cursorKnife`, which is of type `Qt::BitmapCursor`. So the conditional that switches the cursor back to the `m_cursorHand` isn't met and you end up with a knife on the draw mode, until you move to one of the edges and then back to the middle. That's because when you move to the edges the cursor is changed to `Qt::SizeHorCursor`, then when you move back to the middle the `if( cursor()->shape() != Qt::BitmapCursor )` conditional is true and `leaveEvent` changes the cursor back to the hand one.

To be honest I'm not sure why `leaveEvent` is being called in that particular part of the code instead of dealing with the cursor changes inside `TrackContentObjectView::mouseMoveEvent` alone. Maybe someone can clarify if there's another meaning for that method call there. If not, we can get rid of it and just rewrite the logic so it accounts for those situations more appropriately, checking for all 3 situations (mouse in the middle in knife mode, mouse in the middle in draw mode and mouse on the edges).

IanCaio:
And as mentioned in Discord, there is a small bug I still didn't manage to track down, could use some help on this one. The steps to reproduce it:

1) Add a SampleTCO to a Sample Track
2) Split that SampleTCO
3) Try to Copy and Paste (either through the context menu or by dragging and dropping) the left-most SampleTCO from the two splitted ones

It will fail to paste the TCO, pasting a infinitely small SampleTCO instead. If you move or resize the left-most SampleTCO before copying and pasting however it does work. Which makes me think that it has something to do with some update routine that is not being called and is causing the TCO data to be corrupt or something. I still didn't find out what it is, I only know that copying the right-most SampleTCO works fine.

I also unresolved a change request from a one-liner `if` that ended up not being changed in the last commits. Besides those issues I don't see anything else. I like the changes on the way `Ctrl` is being handled on the Song Editor. If I find out anything more about this bug I'll update you.

IanCaio:
Already messaged on Discord, but just for reference, apparently the cause of the bug mentioned above are the following lines:

https://github.com/LMMS/lmms/blob/245160dd196dc51bfc01ffceac6d92953c582aeb/src/gui/TrackContentObjectView.cpp#L606-L607

Even though the comment says `move or resize`, it's being called for Move, Resize, ResizeLeft and Split. However, the journalling is only being set back to true for the first 3:

https://github.com/LMMS/lmms/blob/245160dd196dc51bfc01ffceac6d92953c582aeb/src/gui/TrackContentObjectView.cpp#L966-L970

Because `setJournalling` is not set back to true on the SampleTCO being split, when it's being copied `saveState` will not save the data from the TCO because it only does so when `isJournalling` is true. Moving or Resizing the TCO fixes the problem because then `mouseReleaseEvent` sets the journalling back to true.

src/gui/TrackContentObjectView.cpp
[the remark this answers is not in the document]
Spekular: Whatever line this is referring to doesn't exist anymore.

src/gui/widgets/TrackContentWidget.cpp
```diff
@@ -573,6 +581,15 @@ void TrackContentWidget::mousePressEvent( QMouseEvent * me )
 
 
 
+void TrackContentWidget::mouseReleaseEvent( QMouseEvent * me )
+{
+	gui->songEditor()->syncEditMode();
+	QWidget::mousePressEvent(me);
```
IanCaio, line 587:
```suggestion
	QWidget::mouseReleaseEvent(me);
```

I just noticed this, was it intentional or a typo?

src/gui/TrackContentObjectView.cpp
```diff
@@ -592,28 +613,39 @@ void TrackContentObjectView::mousePressEvent( QMouseEvent * me )
 				gui->songEditor()->m_editor->selectAllTcos( false );
 				m_tco->addJournalCheckPoint();
 
-				// move or resize
+				// Move, Resize, ResizeLeft and Split
 				m_tco->setJournalling( false );
```
IanCaio, line 617:
```suggestion
				// Move, Resize and ResizeLeft
				if (m_action == Move || m_action == Resize || m_action == ResizeLeft)
				{
					m_tco->setJournalling( false );
				}
```

I apologize for another change request, because I missed it myself on the diff that I sent you. It doesn't make much sense for us to disable the TCO journalling just to re-enable it immediately before splitting the TCO. So instead we could just disable it for the actions that require it. `if (m_action != Split)` would work too, but I think checking for the actions that need disabling is better in case we add more actions in the future, so we don't forget about this conditional and it end up disabling it for them too.

src/gui/TrackContentObjectView.cpp
```diff
@@ -918,17 +957,27 @@ void TrackContentObjectView::mouseReleaseEvent( QMouseEvent * me )
...
-
-	if( m_action == Move || m_action == Resize || m_action == ResizeLeft )
+	else if( m_action == Move || m_action == Resize || m_action == ResizeLeft )
 	{
 		// TODO: Fix m_tco->setJournalling() consistency
 		m_tco->setJournalling( true );
 	}
+	else if( m_action == Split )
+	{
+		m_tco->setJournalling( true );
```
IanCaio, line 967:
```suggestion
```

This suggestion is coupled with the one before (Line 616)

IanCaio (approved):
As discussed on Discord, I pushed the last review's suggestions to the PR, gave the code another broad look and it all looks fine.

Style guidelines kept aside, the code LGTM

PhysSong:
@DomClark Do you still want to review this?

DomClark (approved):
Looks good to me. `include/Track.h` and `src/core/Track.cpp` only have formatting changes; can you revert these before merging?

Spekular:
Reverted, and merging now! Huge thanks to everyone who reviewed and helped out, including of course BaraMGB for making the PR this is rooted in.

JohannesLorenz:
Causes #5949 .

PR #5526, opened 2020-06-04
Make better use of getSelectedNotes() in PianoRoll.cpp

Spekular:
Plus some other refactoring.

In terms of objective measures
- Reduces LOC
- Uses fewer levels of indentation in many places

Subjectively speaking
- Communicates the intent of code blocks more clearly. "For each selected note, do ..." rather than "For every note, check if it is selected. If it is, do ..."
- Easier to read in general

Could be slower in some cases, since we iterate once to get selected notes and once to use them, rather than one iteration with an if statement. However, should also be faster in other cases where we can do things like reuse the vector of selected notes or avoid repeated checks like ``if (first)`` (replaced by handling the first element of the vector once, before the loop).

### DIFF
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -1010,8 +1010,8 @@ void PianoRoll::drawNoteRect( QPainter & p, int x, int y,
 			int const distanceToBorder = 2;
 			int const xOffset = borderWidth + distanceToBorder;
 
-			// noteTextHeight, textSize are not suitable for determining vertical spacing, 
-			// capHeight() can be used for this, but requires Qt 5.8. 
+			// noteTextHeight, textSize are not suitable for determining vertical spacing,
+			// capHeight() can be used for this, but requires Qt 5.8.
 			// We use boundingRect() with QChar (the QString version returns wrong value).
 			QRect const boundingRect = fontMetrics.boundingRect(QChar::fromLatin1('H'));
 			int const yOffset = (noteHeight - boundingRect.top() - boundingRect.bottom()) / 2;
@@ -1112,27 +1112,24 @@ void PianoRoll::clearSelectedNotes()
 
 
 
-
-void PianoRoll::shiftSemiTone( int amount ) // shift notes by amount semitones
+void PianoRoll::shiftSemiTone(int amount) //Shift notes by amount semitones
 {
-	if (!hasValidPattern()) {return;}
+	if (!hasValidPattern()) { return; }
 
+	auto selectedNotes = getSelectedNotes();
+	//If no notes are selected, shift all of them, otherwise shift selection
+	if (selectedNotes.empty()) { return shiftSemiTone(m_pattern->notes(), amount); }
+	else { return shiftSemiTone(selectedNotes, amount); }
+}
+
+void PianoRoll::shiftSemiTone(NoteVector notes, int amount)
+{
 	m_pattern->addJournalCheckPoint();
-	bool useAllNotes = ! isSelection();
-	for( Note *note : m_pattern->notes() )
-	{
-		// if none are selected, move all notes, otherwise
-		// only move selected notes
-		if( useAllNotes || note->selected() )
-		{
-			note->setKey( note->key() + amount );
-		}
-	}
+	for (Note *note : notes) { note->setKey( note->key() + amount ); }
 
 	m_pattern->rearrangeAllNotes();
 	m_pattern->dataChanged();
-
-	// we modified the song
+	//We modified the song
 	update();
 	gui->songEditor()->update();
 }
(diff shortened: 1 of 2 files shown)


Spekular:
Oh, and from what I can tell this doesn't touch any of the stuff in #5253.

src/gui/editors/PianoRoll.cpp
```diff
@@ -1712,57 +1689,39 @@ void PianoRoll::mousePressEvent(QMouseEvent * me )
...
-					}
+					m_moveBoundaryLeft = qMin(
+										note->pos().getTicks(),
+										(tick_t) m_moveBoundaryLeft );
+					m_moveBoundaryRight = qMax( (int) note->endPos(),
+											m_moveBoundaryRight );
+					m_moveBoundaryBottom = qMin( note->key(),
+									   m_moveBoundaryBottom );
+					m_moveBoundaryTop = qMax( note->key(),
+												m_moveBoundaryTop );
```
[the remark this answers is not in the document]
Spekular: I left it as is, aside from outdenting. I can probably try to clean it up though.

src/gui/editors/PianoRoll.cpp
```diff
@@ -2813,16 +2753,13 @@ void PianoRoll::dragNotes( int x, int y, bool alt, bool shift, bool ctrl )
 		else
 		{
 			// shift is not pressed; stretch length of selected notes but not their position
-			for (Note *note : notes)
+			for (Note *note : getSelectedNotes())
```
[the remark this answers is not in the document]
Spekular: I noticed this as well but let it be since getSelectedNotes would only be called once either way, and I was wary of accidentally reusing selectedNotes for too long. However, I think you're right about this since the selection isn't modified in this function.

src/gui/editors/PianoRoll.cpp
```diff
@@ -1112,66 +1112,53 @@ void PianoRoll::clearSelectedNotes()
...
 
 
 
-void PianoRoll::shiftPos( int amount ) //shift notes pos by amount
+void PianoRoll::shiftPos(int amount) //Shift notes pos by amount
 {
-	if (!hasValidPattern()) {return;}
+	if (!hasValidPattern()) { return; }
 
+	auto selectedNotes = getSelectedNotes();
```
Spekular, line 1144:
```suggestion
	auto selectedNotes = getSelectedNotes();
	//If no notes are selected, shift all of them, otherwise shift selection
```

src/gui/editors/PianoRoll.cpp
```diff
@@ -4188,47 +4121,22 @@ void PianoRoll::pasteNotes()
...
-		}
-	}
-
-	if( update_after_delete )
-	{
-		Engine::getSong()->setModified();
-		update();
-		gui->songEditor()->update();
-	}
+	for(Note* note: selectedNotes) { m_pattern->removeNote( note ); }
```
Spekular, line 4134:
```suggestion
	for (Note* note: selectedNotes) { m_pattern->removeNote( note ); }
```

src/gui/editors/PianoRoll.cpp
```diff
@@ -4188,47 +4121,22 @@ void PianoRoll::pasteNotes()
...
 
-void PianoRoll::deleteSelectedNotes()
+//Return false if no notes are deleted
+bool PianoRoll::deleteSelectedNotes()
 {
-	if( ! hasValidPattern() )
-	{
-		return;
-	}
+	if(!hasValidPattern()) { return false; }
```
Spekular, line 4127:
```suggestion
	if (!hasValidPattern()) { return false; }
```

src/gui/editors/PianoRoll.cpp
```diff
@@ -2762,40 +2701,37 @@ void PianoRoll::dragNotes( int x, int y, bool alt, bool shift, bool ctrl )
...
+			for (Note *note : selectedNotes)
 			{
-				if(note->selected())
+				// scale relative start and end positions by scaleFactor
+				int newStart = stretchStartTick + scaleFactor *
+					(note->oldPos().getTicks() - stretchStartTick);
+				int newEnd = stretchStartTick + scaleFactor *
+					(note->oldPos().getTicks()+note->oldLength().getTicks() - stretchStartTick);
+				// if  not holding alt, quantize the offsets
+				if(!alt)
```
Spekular, line 2712:
```suggestion
				if (!alt)
```

src/gui/editors/PianoRoll.cpp
```diff
@@ -2686,36 +2626,33 @@ void PianoRoll::dragNotes( int x, int y, bool alt, bool shift, bool ctrl )
 
 	if (m_action == ActionMoveNote)
 	{
-		for (Note *note : notes)
+		for (Note *note : getSelectedNotes())
 		{
-			if( note->selected() )
+			if( shift && ! m_startedWithShift )
```
Spekular, line 2631:
```suggestion
			if (shift && ! m_startedWithShift)
```

src/gui/editors/PianoRoll.cpp
```diff
@@ -1796,28 +1750,14 @@ void PianoRoll::mousePressEvent(QMouseEvent * me )
 					// if they're holding shift, copy all selected notes
 					if( ! is_new_note && me->modifiers() & Qt::ShiftModifier )
 					{
-						// vector to hold new notes until we're through the loop
-						QVector<Note> newNotes;
-						for( Note* const& note : notes )
+						for (Note* note: selectedNotes)
```
Spekular, line 1753:
```suggestion
						for (Note *note: selectedNotes)
```

src/gui/editors/PianoRoll.cpp
```diff
@@ -1796,28 +1750,14 @@ void PianoRoll::mousePressEvent(QMouseEvent * me )
...
 						{
-							if( note->selected() )
-							{
-								// copy this note
-								Note noteCopy( *note );
-								newNotes.push_back( noteCopy );
-							}
-							++it;
+							Note * newNote = m_pattern->addNote(*note, false);
```
Spekular, line 1755:
```suggestion
							Note *newNote = m_pattern->addNote(*note, false);
```

src/gui/editors/PianoRoll.cpp
```diff
@@ -1712,57 +1689,34 @@ void PianoRoll::mousePressEvent(QMouseEvent * me )
...
+
+				auto selectedNotes = getSelectedNotes();
+
+				m_moveBoundaryLeft = selectedNotes.first()->pos().getTicks();
+				m_moveBoundaryRight = selectedNotes.first()->endPos();
+				m_moveBoundaryBottom = selectedNotes.first()->key();
+				m_moveBoundaryTop = m_moveBoundaryBottom;
 
+				//Figure out the bounding box of all the selected notes
+				for (Note* note: selectedNotes)
```
Spekular, line 1707:
```suggestion
				for (Note *note: selectedNotes)
```

src/gui/editors/PianoRoll.cpp
```diff
@@ -1712,57 +1689,34 @@ void PianoRoll::mousePressEvent(QMouseEvent * me )
 				m_mouseDownKey = m_startKey;
 				m_mouseDownTick = m_currentPosition;
 
-				bool first = true;
-				for( it = notes.begin(); it != notes.end(); ++it )
+				//If clicked on an unselected note, remove selection and select that new note
+				if(!m_currentNote->selected())
```
Spekular, line 1693:
```suggestion
				if (!m_currentNote->selected())
```

src/gui/editors/PianoRoll.cpp
```diff
@@ -1112,66 +1112,53 @@ void PianoRoll::clearSelectedNotes()
...
-				if( m_moveBoundaryLeft + amount < 0 )
-				{
-					amount += 0 - (amount + m_moveBoundaryLeft);
-				}
-				first = false;
-			}
-			note->setPos( note->pos() + amount );
-		}
-	}
+	for( Note *note : notes ) { note->setPos( note->pos() + shiftAmount ); }
```
Spekular, line 1157:
```suggestion
	for (Note *note : notes) { note->setPos( note->pos() + shiftAmount ); }
```

Spekular:
Any objections if I merge this? There's no functional difference*, so I'll merge on Monday if no one complains or does it before me.

*One tiny change, that I consider to be an improvement, due to the removal of this check `useAllNotes && note->length() > 0`. Previously, shifting note position (horizontal shift) with no notes selected would move all notes except beat notes. This is inconsistent with the behavior when all notes are selected (they all shift), and with shifting semitones without a selection (all notes shift).

PR #5529, opened 2020-06-07
Modifying factory presets to sound more friendly.

### LINKED ISSUE #5527: Almost all presets has no envelopes
# Bug Summary
I have searched, but no issue seams to address this, but Almost all presets in LMMS has no envelopes. The zasfx presets does, but they are of cause internal, and not related to LMMS own ENV/LFO tab.
Most problematic are the *samples*, where many clicks and pops, but i have not been through all preset, so i cant say how many we got with artefacts.

#### Argumentation 
On youTube LMMS is ridiculed and dissed as being of low quality. Popping and clicking presets is part of this. It is reasonably easy to fix. No coding is needed, just edit and resave.
Any user with good understanding of the presets, can do this.

#### Steps to reproduce
Open any default preset

#### Expected behavior
LMMS presets ought to have an envelope, that prevents clicks and pops from attack ao release. A default All-pass-filter could also be set. 
This  would not inflict old projects or any saved presets. All presets has envelopes hardcoded in both project-file, and preset-files.

#### Actual behavior
Most presets does not have either envelope or default filter

#### Affected LMMS versions
All

#### Suggested method
All presets are renamed with an _ infront of current name with a batch-script, then they can be saved with new envelopes, without the leading underscore, and nothing would have changed. Total backward comparability is preserved.

dj-pixus:
fixes #5527.
- Added a very short attack and release to prevent click
- Adjusted some levels to prevent clipping
- Removed DC offset.


qnebra:
For me, from what I tested, in terms of sound this is really nice. I know, this is subjective and without scientific scrutiny, but I think this made presets better.

data/presets/TripleOscillator/Drums_HardKick.xpf
```diff
@@ -3,7 +3,7 @@
 <lmms-project version="1.0" creator="LMMS" creatorversion="1.0.2" type="instrumenttracksettings">
   <head/>
   <instrumenttracksettings muted="0" type="0" name="Hard Kick" solo="0">
-    <instrumenttrack pan="0" fxch="0" pitchrange="1" pitch="0" basenote="45" vol="100">
+    <instrumenttrack pan="0" fxch="0" pitchrange="1" pitch="0" basenote="45" vol="1000">
```
tresf, line 6: Is this `1000` a typo?

tresf:
> fixes #5527 

I'm not sure this statement is true.  Think it **partially** fixes #5527.  To that point, I'd like @musikbear to review this PR as the author of the linked bug report.

Furthemore, I'd love someone to go through these presets (before and after) and make sure they are in fact better than before.  I think @qnebra has done this, but I'm not sure.

Preset changes are generally safe because they don't affect existing projects, only new ones.   Perhaps to speed this up, @dj-pixus could create a zip file attachment with all affected presets.  That way, a non-developer can jump in and help review.  I'm also going to cross-post in #testing on Discord.

superpaik:
Sound is subjective, but I found some issues if you wanna doublecheck:
- bitinvader - subbass, has less click at the beginning than the original but still (I think is due to the waveform)
- bitinvader - toypiano, too low volume around -30dB (also in the original)
- kicker - trapkick, still has a click, increasing attack to 0,02 will do
- organic - rubberband, to much volume (also in the original) it clips
- SID - bass, the AMT is not activated, so the enveloped is not working. If activated, the sound is weird for a bass, maybe a increase the HOLD and shorten the SUST. 
- TripleOsc - RaveBass, still has click, maybe double the attack.
- TripleOsc - Space Bass, totally different sound. Maybe is due to the 'userwavefile0="samples/empty.wav"' param, I don't find it
- TripleOsc - Square, also different sound. The same reason ¿?
- TripleOsc - TranceLead, attack=0 ¿?
- Xpressive - can't test because I don't have it installed (windows lmms-v1.2.2)

dj-pixus:
Made some fine tuning on the presets @superpaik mentioned, and also some new ones. And corrected the 1000% typo.
However, there are cases where the "problems" are intended, e.g. some Kicker presets have the Click knob turned up, because that's part of the concept. So I left them in peace.
Also there were some hyper-super stereo effects with serious phase problems, tried to regulate them a bit.
In BitInvader preset Subbass there was an audio file loaded in the cutoff frequency LFO, no clue why. I removed it because it does not do anything good in any way.

superpaik:
To me is ok! In anycase, presets are a starting point to get the sound the user wants, so is ok if some of them are not "perfect" because they have to be tailored by the user to the song/sound produced.

tresf:
> To me is ok! In anycase, presets are a starting point to get the sound the user wants, so is ok if some of them are not "perfect" because they have to be tailored by the user to the song/sound produced.

Thanks kindly for the review, it's greatly appreciated.  @dj-pixus thanks for the contribution, we couldn't do this without the help of people like you!

These changes will only apply to new projects created with `master` from this point forward.  Old projects will remain unmodified.  I assume any preset-specific disagreements can be fixed on a new PR/issue.

Merging.

musikBear:
Sorry i have not been active here for some days. I never saw 
https://github.com/LMMS/lmms/pull/5529#issuecomment-710263846
I will look at the release.

musikBear:
I have listened to those in the zipPack.
All are not to my personal preferences, but thats beside the scope. Clicks and Clipping has bee dealt with, and thats fine.
However there are many more -All *samples* should have envelopes. It would also remove a really stupid coding-issue with AFP, if all samples had *saved sample-specific* envelopes
That task is on the other hand a unrewarding task, because there is work being done on a brand new Sample-collection for lmms. However the old ones are still in use in all projects.
drtl...
Back on review
These i did not like -Otherwise good job @dj-pixus 👍 
`
BitInvaderd
Bell:                         Should ring out
Toy-piano:                    Cant play to fast notes without blend. REL too long
wah_synth:                    A bit of LFO should be added to show full potential

Organic
Rubberband:                   Not organ envelope and too loud

3oc
RaveBass:                     Better before. Filter not invoked
Space Bass:                   Better before `

tresf:
@musikBear thanks for the feedback, I guess it's up to @dj-pixus to issue another PR with your changes if in agreement.

PR #5530, opened 2020-06-09
Remove macros likely() and unlikely()

Sawuare:
Because they're merely aliases for the Qt macros `Q_LIKELY()` and `Q_UNLIKELY()`, and that's a bad idea.

### DIFF
--- a/include/AutomatableModel.h
+++ b/include/AutomatableModel.h
@@ -148,7 +148,7 @@ public:
 	template<class T>
 	inline T value( int frameOffset = 0 ) const
 	{
-		if( unlikely( hasLinkedModels() || m_controllerConnection != NULL ) )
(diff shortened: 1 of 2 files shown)


Spekular:
> Because they're merely aliases for the Qt macros Q_LIKELY() and Q_UNLIKELY()

That is a fairly useless alias. +1 for removal.

> and that's a bad idea.

Why? It'd be great if you clarify the downsides you see, aside from the fact that the renaming is unnecessary.

Sawuare:
@Spekular Because it hides the fact that they're provided by Qt. :oncoming_police_car:

Sawuare:
Oops! I accidentally closed this. I'll wait for the CI checks to complete again then merge.

DomClark:
Can we remove the only use of the macro as well? Unless it's been profiled, it looks like premature optimisation to me.

PR #5531, opened 2020-06-11
Adds a button to clone the current BB track pattern inside the BB Editor

IanCaio:
## Issue

I noticed a common situation during the music production workflow on LMMS. Sometimes I was working drumlines inside the BBEditor and for a small section I wanted to add a variation (i.e.: some drum fill). Currently there are two options:

1. Cloning the track on the song editor to keep the current pattern, editing the pattern on the BBEditor to add the variation wanted and then clearing the track on the song editor, adding just the bars where this variation should play.
2. Adding a new empty BB track, copying all the pattern by hand adding the variations wanted and then adding the bars on the song editor.

I realized it would make the process a little quicker if there was a button inside the BBEditor that just cloned the pattern, but not the song editor content (which is usually the desirable behavior on that case).

## Proposed PR solution

This PR adds a small button on the BBEditor, next to the "_Add Beat/Bassline_", that clones the current BBTrack pattern, but also clears the content of the song editor. For now the icon being used is a copy of the _edit_copy.png_ icon, but it can be changed to something more distinct.


IanCaio:
> Instead of copying the innards of the clone and clear functions, why not just call the existing clone function, then the existing "clear track" function? That should be DRYer.

@Spekular The `cloneTrack` and `clearTrack` methods are from the `TrackOperationsWidget` class, but I haven't figured out yet how to access the `TrackOperationsWidget` from a particular `Track` object (even my failed attempts looked a little verbose). Is there an easy way I have missed?

Spekular:
Having taken a closer look, no there probably isn't without moving things around, since clone and clear are private slots. Even if you get the widget with

```
BBTrackContainer *bbtc = static_cast<BBTrackContainer*>(model());
BBTrack *bbt = BBTrack::findBBTrack(bbtc->currentBB());
auto *widget = bbt.getTrackView()->getTrackOperationsWidget();
```

you wouldn't be able to call the methods. I guess I'm okay with this version then, although I think it'd be nice if you added a call to `bbt->addJournalCheckPoint();` before the clone so it can be undone.

include/BBEditor.h
```diff
@@ -86,6 +86,7 @@ public slots:
 	void removeSteps();
 	void addSampleTrack();
 	void addAutomationTrack();
+	void cloneBBTrackPattern();
```
Spekular, line 89:
```suggestion
	void clonePattern();
```
Because we're in BBEditor, you can drop the "BBTrack" part of this function name.

src/gui/editors/BBEditor.cpp
```diff
@@ -311,3 +314,21 @@ void BBTrackContainerView::makeSteps( bool clone )
 		}
 	}
 }
+
+// Creates a clone of the current BB track with the same pattern, but with no TCOs on the song editor
```
Spekular, line 318:
```suggestion
// Creates a clone of the current BB track with the same pattern, but no TCOs in the song editor
// TODO: Avoid repeated code from cloneTrack and clearTrack in TrackOperationsWidget somehow
```
1. Removed second `with`. This is super nitpicky, but I think it's shorter and more natural.
2. `no TCOs on the song editor` -> `no TCOs in the song editor`. You could even drop `in the song editor` entirely.
3. I think it'd be nice to leave a TODO here for future reference.

Spekular:
From conversation on Discord, it seems the journal system doesn't handle tracks being added/deleted. With that in mind I don't have any big issues with this PR anymore, but I left a review with some nitpicks. I'm willing to merge once they're addressed if no one else disagrees.

IanCaio:
> From conversation on Discord, it seems the journal system doesn't handle tracks being added/deleted. With that in mind I don't have any big issues with this PR anymore, but I left a review with some nitpicks. I'm willing to merge once they're addressed if no one else disagrees.

Thanks for the review @Spekular , I made a new commit with the requested changes!

Spekular:
Awesome, thanks! I'll wait a day or two with merging so others can chime in if they want. Feel free to ping me on Monday if I forget :)

PR #5536, opened 2020-06-13
basics: Change sampleFrame to use `std::array`

JohannesLorenz:
... in order to make standard containers be able to store it.

### DIFF
--- a/include/LocklessRingBuffer.h
+++ b/include/LocklessRingBuffer.h
@@ -33,79 +33,33 @@
 #include "../src/3rdparty/ringbuffer/include/ringbuffer/ringbuffer.h"
 
 
-//! A convenience layer for a realtime-safe and thread-safe multi-reader ring buffer library.
+//! A convenience layer for a realtime-safe and thread-safe multi-reader ringbuffer
 template <class T>
-class LocklessRingBufferBase
+class LocklessRingBuffer
 {
 	template<class _T>
 	friend class LocklessRingBufferReader;
 public:
-	LocklessRingBufferBase(std::size_t sz) : m_buffer(sz)
+	LocklessRingBuffer(std::size_t sz) : m_buffer(sz)
 	{
 		m_buffer.touch();	// reserve storage space before realtime operation starts
 	}
-	~LocklessRingBufferBase() {};
+	~LocklessRingBuffer() {};
 
 	std::size_t capacity() const {return m_buffer.maximum_eventual_write_space();}
 	std::size_t free() const {return m_buffer.write_space();}
 	void wakeAll() {m_notifier.wakeAll();}
+	std::size_t write(const sampleFrame *src, std::size_t cnt, bool notify = false)
+	{
+		std::size_t written = LocklessRingBuffer<T>::m_buffer.write(src, cnt);
+		// Let all waiting readers know new data are available.
+		if (notify) {LocklessRingBuffer<T>::m_notifier.wakeAll();}
+		return written;
+	}
 
 protected:
 	ringbuffer_t<T> m_buffer;
 	QWaitCondition m_notifier;
-};
-
-
-// The SampleFrameCopier is required because sampleFrame is just a two-element
-// array and therefore does not have a copy constructor needed by std::copy.
-class SampleFrameCopier
-{
-	const sampleFrame* m_src;
-public:
-	SampleFrameCopier(const sampleFrame* src) : m_src(src) {}
-	void operator()(std::size_t src_offset, std::size_t count, sampleFrame* dest)
-	{
-		for (std::size_t i = src_offset; i < src_offset + count; i++, dest++)
-		{
-			(*dest)[0] = m_src[i][0];
-			(*dest)[1] = m_src[i][1];
-		}
-	}
-};
-
-
-//! Standard ring buffer template for data types with copy constructor.
-template <class T>
-class LocklessRingBuffer : public LocklessRingBufferBase<T>
-{
-public:
-	LocklessRingBuffer(std::size_t sz) : LocklessRingBufferBase<T>(sz) {};
-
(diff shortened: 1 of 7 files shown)


JohannesLorenz:
This PR does 3 things:

* Change `sampleFrame` and `surroundSampleFrame` to use `std::array`
* Remove definition of `sampleFrameA` (seems to be unused since 2008)
* Change code around libsamplerate use (I haven't checked that code myself yet)

JohannesLorenz:
Sorry, I had to do a force-push, it did not compile before.
@Reflexe This branch will have merge conflicts in 2 lines, I marked them with "@recording".

JohannesLorenz:
Ready to review now. @DomClark I added you as reviewer. Of course, feel free to remove yourself if you don't want to review.

IMO there are 2 questions:

1. Should we keep the code like it is in the PR around libsamplerate? @DomClark mentioned he was not sure "what the rules are about passing C++ objects to C code, especially when this is going to access `_src[0].data()` out-of-bounds". I think the only alternative would be 2 slow `memcpy`s for input and output for each resampling (or not use libsamplerate at all). Maybe it's best to leave it as it is and only change it when anyone complains?
2. In general, there are tons of code possibly affected by this. Most stuff would be array access, but imagine anyone casts `sampleFrame*` to `void*`, or uses type traits, or ... It's so much to check, so better don't check it at all?

PhysSong:
> when this is going to access `_src[0].data()` out-of-bounds

This is a tricky problem, but I think this won't make stuff worse because OOB accesses are problematic when accessing to C objects as well.

> imagine anyone casts `sampleFrame*` to `void*`, or uses type traits

`std::array<T, N>` mimics `T[N]` quite well, but there are some differences including:
- You can cast `T[N]` to `T*` implicitly, but you can't do that with `std::array`
- `std::is_array` will return `false` for `std::array`

However, I don't think that will cause issues, at least for now.

@DomClark Can we merge this as-is?

qnebra:
@DomClark To add more context for investigation of this bug, I use a lot of VST in projects. Propably completely irrelevant at all, but maybe will help.

PhysSong:
@qnebra The freezing might be related to the embedding method(the default is XEmbed on Linux). If you run LMMS from the build directory, it won't use the default LMMS setting file. Can you test this PR with the "none" embedding method?

qnebra:
Well, I use Windows at this moment, also not touched embedding settings in lmms. I will test this further after fixing broken build system on my PC.

qnebra:
> @qnebra The freezing might be related to the embedding method(the default is XEmbed on Linux). If you run LMMS from the build directory, it won't use the default LMMS setting file. Can you test this PR with the "none" embedding method? Please make sure you check the setting before you open your project files.

Export crashes even if only track in project is single 3xOSC, completly default settings.
Here is Win installer of current master with only this PR merged, you will figure out by name as I add "stdarray":

https://drive.google.com/file/d/1caGagpw_IqobJ99E-Nvi9v7_2s4YCBvy/view?usp=sharing

JohannesLorenz:
@DomClark do you still plan to investigate this?

DomClark:
I investigate the crashes and freezing, and they appear to be unrelated to this PR. I will get on with checking the code for the potential issue I mentioned above.

DomClark (changes requested):
I only found one instance which will break with the new pass-by-value behaviour: https://github.com/LMMS/lmms/blob/1dab416c6cb9dd437829693bdabad5533b37271e/plugins/Delay/StereoDelay.h#L48
It should be sufficient to change it to a reference.

JohannesLorenz:
> I only found one instance which will break with the new pass-by-value behaviour

I just cross-checked your analysis (`grep`ping for `'([^(]*sampleFrame[[:space:]]*[^&*].*)'`), and found only this occurrence, but no other ones. So I fixed and pushed it. I think that issue is solved now.

I also pushed a commit removing the `LocklessRingBuffer` specialization for `sampleFrame` (@he29-net sorry for removing your code! do you want to check my commit 52a8b94?)

@DomClark should we analyze for any other issues? Do you recommend testing? Or can this be merged?

DomClark (approved):
LGTM. I can't think of any other potential issues, although obviously that doesn't mean there aren't any. I did my search in a similar way; the only way something might have been missed is if a `sampleFrame` were used under a different name (`using`/`typedef`/template parameter), but I couldn't find any such cases (I only found two template uses - the vector in the LV2 code, and the ringbuffer, which are both fine). I'm not sure what we can do to test this further, other than to cover every bit of code that handles audio, and I think the best way to do that is to have people actually use this. So I think it can be merged.

Code considerations:
* Should the `typedef`s and C-style float pointer casts be replaced by the more modern `using` and `const_cast`? Maybe out of scope, but since those lines are touched anyway they could be updated.
* Are the merge conflict warnings worth keeping? The recording PR already has lots of merge conflicts, including in that same file.

PhysSong:
> The recording PR already has lots of merge conflicts

I think there is a plan to split that into several small parts. I'm not sure if we will do that, BTW.

JohannesLorenz:
> Should the `typedef`s and C-style float pointer casts be replaced by the more modern `using` and `const_cast`?

`using` sounds good.

The cast to non-`const` is because libsamplerate used `float* data_in` (without `const`) in the past, but they fixed it in January 2013. If our CI accepts it, can I remove this const cast, or should I keep it for compatibility (and use `const_cast`)?

> Are the merge conflict warnings worth keeping? The recording PR already has lots of merge conflicts, including in that same file.

Somehow I think they can be removed. I'll do that soon.

JohannesLorenz:
OK, I just tested, our CI fails without the const cast :-1: So I'm now adding `const_cast`.

JohannesLorenz:
OK, CI passes, all is reviewed, @he29-net also replied he is ok with 21ee7f1. I'll squash and merge this in the next hours.

PR #5537, opened 2020-06-14
Fix some TCOs not saving properly

### LINKED ISSUE #3964: Random notes deleted upon project reopen
I can't find a way to consistently reproduce this bug, it always happens randomly

Whenever I close and reopen LMMS to the project I was once working on, a segment of notes would be deleted randomly from the Song editor. I always have to type in the notes again after that and risk them being deleted again.

I'm using LMMS 1.2.0 RC4, is this an already existent bug?

DomClark:
Fixes #3964. See https://github.com/LMMS/lmms/issues/3964#issuecomment-643685445 for details on this patch.

### DIFF
--- a/src/core/Track.cpp
+++ b/src/core/Track.cpp
@@ -717,7 +717,7 @@ void TrackContentObjectView::mousePressEvent( QMouseEvent * me )
 
 		setInitialMousePos( me->pos() );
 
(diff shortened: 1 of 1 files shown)


zonkmachine:
Tested, works like a charm!

zonkmachine:
Merge?

PR #5538, opened 2020-06-15
June2020m b

musikBear:
I have added 7 features
General:
    New piano-like envelope for general instrument (Part of LMMS#2092
Pianoroll
    Humanize note position (speedbar-button) (Part of LMMS#1165
    Humanize note velocity (speedbar-button) (Part of LMMS#1165
    Strum from top (speedbar-button) (LMMS#2380
    Strum from bottom (speedbar-button) (LMMS#2380
    Shorten notes by 1/192 ( key - ) (LMMS#3138
    lengthen notes by 1/192 ( key + ) (LMMS#3138
    Methods demo: https://youtu.be/OlBtU0JYvyY


src/core/EnvelopeAndLfoParameters.cpp
```diff
@@ -94,12 +94,12 @@ EnvelopeAndLfoParameters::EnvelopeAndLfoParameters(
...
-	m_sustainModel( 0.5, 0.0, 1.0, 0.001, this, tr( "Env sustain" ) ),
-	m_releaseModel( 0.1, 0.0, 2.0, 0.001, this, tr( "Env release" ) ),
-	m_amountModel( 0.0, -1.0, 1.0, 0.005, this, tr( "Env mod amount" ) ),
+	m_attackModel( 0.025, 0.0, 2.0, 0.001, this, tr( "Env attack" ) ),
+	m_holdModel( 0.0, 0.0, 2.0, 0.001,     this, tr( "Env hold" ) ),
+	m_decayModel( 0.6, 0.0, 2.0, 0.001,    this, tr( "Env decay" ) ),
+	m_sustainModel( 0.0, 0.0, 1.0, 0.001,  this, tr( "Env sustain" ) ),
+	m_releaseModel( 0.2, 0.0, 2.0, 0.001,  this, tr( "Env release" ) ),
+	m_amountModel( 1.0, -1.0, 1.0, 0.005,  this, tr( "Env mod amount" ) ),
 	m_valueForZeroAmount( _value_for_zero_amount ),
```
zonkmachine, line 103: Here is an earlier, similar, fix that had to be reverted: https://github.com/LMMS/lmms/pull/3090

PianoRoll.h
```diff
@@ -0,0 +1,516 @@
+/*
```
zonkmachine, line 1: This commit adds PianoRoll.h and PianoRoll.cpp to the top of the lmms src. Why?

src/gui/editors/PianoRoll.cpp
```diff
@@ -64,6 +64,11 @@
 #include "TimeLineWidget.h"
 #include "StepRecorderWidget.h"
 
+//for random
```
zonkmachine, line 67: We already perform srand() in main.cpp and that should be enough for all of LMMS.

PianoRoll.h
musikBear: I cant understand what you mean? Are the files in different order?
I have only added code.

src/core/EnvelopeAndLfoParameters.cpp
musikBear: The fix only influence new made instruments. Not even instruments in current default template, has the envelope. Non of the existing saved presets should be changed, but i must admit, that i did not test **samples**. I will try and see if they act different from other items in *Instruments*
About default template. How about having a nice zasfx preset, instead of the sine-beeping 3oC? Afaiu even MAC now supports zasfx?

PianoRoll.h
zonkmachine: This example is from your master branch. These two files are where they're supposed to be:
https://github.com/musikBear/lmms/blob/master/src/gui/editors/PianoRoll.cpp
https://github.com/musikBear/lmms/blob/master/include/PianoRoll.h

These are the duplicate ones that were added by you in 4492e2cb9932204c600333e0e219849f8cb735f3 :
https://github.com/musikBear/lmms/blob/master/PianoRoll.cpp
https://github.com/musikBear/lmms/blob/master/PianoRoll.h

This PR adds 5,734 lines of code. Does that sound like what you intended?

musikBear: > These are the duplicate ones that were added by you in 4492e2c :
https://github.com/musikBear/lmms/blob/master/PianoRoll.cpp
https://github.com/musikBear/lmms/blob/master/PianoRoll.h

oH! NOW i understand that weird thing!
I made a complete hickup, (spelled with an f) inside my *own* master-repo.
For some odd github reason i 'suddenly' had a whole new repo, instead of the commit i was trying to make. I had no idea that it could result in duplicate files, but yes i did see that weird message about 5734 lines of code.
I thought it was because i had used an external editor, and then deleted everything in the .h file, and *replaced* it with the content from the external editor, so i 'kind of' had changed everything. I asked on Discord what i should do. The reply was that 'it could happen'. 
I used the method 'Add-files', and they did not appear, but what i have done is to add the files to *root* of my master, instead of adding them inside the *sub-directories*!
The correct ones are :
* https://github.com/musikBear/lmms/blob/master/**src/gui/editors**/PianoRoll.cpp
* https://github.com/musikBear/lmms/blob/master/**include**/PianoRoll.h
The other are is garbage.
Github gives me some issues, because part of the ui is written in grey text, i then only see the black text, and then mistakes happens. 
Can you or can (should) I remove the two additional files?
-One question: What is 'Blob' ?

zonkmachine: As for what 'blob' is in this context. Dude, I don't know.
https://developer.github.com/v3/git/blobs/

>  Can you or can (should) I remove the two additional files?

Yes, those files must be removed. Since they were added in one distinct commit I would remove them by removing that commit. I would do this with the rebase command.
`git rebase -i HEAD~5` Will open the commits in your default text editor. Just delete the 
I would revert the commit with the rebase command. Just delete the line with commit 4492e2c and save/close the file.
Stage two is to `push -f`

src/core/EnvelopeAndLfoParameters.cpp
musikBear: > Me said : must admit, that i did not test samples.

and i should have done
You are correct, samples are different from presets. They do not have their envelope 'baked in' They uses current default
That is a terrible descission, but that does not count. It is so. It cant be revoked, that will crush backward compatibility -the holy graal.. 
But i wonder
Several part of the code differentiates between being for MAC / Linux / win
Could we make a similar differentiation between an envelope in *Instruments* and one in AFP -eg **samples**
Then we could have nice default envelopes in instruments, but preserve the no-envelope-state for Samples. I will take a new look at envelope again -eg
`src/core/EnvelopeAndLfoParameters.cpp`
And see if i can make a distinction between Instruments and Samples in AFP

PianoRoll.h
musikBear: I will see if i can swing that. I am *not* strong in GIT-cmds, but as long as i am only messing with my own repo, i guess nothing serious bad can happen(?)

zonkmachine: The suggested action above removes a commit so it's destructive by design. Definitely seriously bad things can happen from this. Always make a backup of the branch you're working on if you're doing something you've never done before.

musikBear: Ok! Thanks for the heads-up!

DomClark:
Is there a particular reason for the choice to implement humanise and strum as modifications to notes in the piano roll, instead of so-called "instrument functions", like note stacking and arpeggios? The latter option strikes me as more flexible and easier to use; for example, I could humanise an arpeggio, or if I wanted to add more notes to my humanised pattern, I wouldn't have to re-align the notes, reset their velocities, and then re-humanise them.

musikBear:
@DomClark particular reason no, but it is one form for implementation. 
I also think its a logical to have methods that 'mess' with note placement inside piano-roll. It could be confusing if notes placed themself apparently out of alignment, because a selection was set 'on' inside the instrument UI.
Comes down to doing something active (clicking a button), versus something being caused by a passive, maybe even accidental setting, away from the component, but i would say that the strongest benefit would be, that you can **redo** in piano-roll, so the placements are also on different levels

DomClark:
Sorry, I may have been unclear before. I meant that the feature, if it went in the instrument functions tab, would not change the positions of any notes in the piano roll; it would only affect playback. This is the same as the arpeggio function - the arpeggio itself doesn't appear in the piano roll; it is only applied to the notes during playback. There would be no need for undo or redo as the notes themselves don't move - all that would be required is turning the effect on or off. I wouldn't expect this feature to be any more confusing than the arpeggiator.

musikBear:
@DomClark I understand your idea, but i still would prefer to have the notes rearranged, because it happens where the notes are visible, and because it is same expression in every replay, so it is *known* what is being played/ exported.

DomClark:
> it is same expression in every replay, so it is known what is being played/ exported.

True - deterministic behaviour is a convincing argument in favour of having humanising in the piano roll. It doesn't apply to strumming though, and I don't know whether it outweighs my other concerns. What do others think?

russiankumar:
Is this one the same as #5544? Could this be closed if it is?

Spekular:
Yes, this seems to be obsolete, considering #5544 mentions "Cleared repo-root for 2 duplicate files" (PianoRoll.cpp and .h)

Issue #5008, opened 2019-06-05
Playhead: Show header line in the editors


Gabrielxd195:
Currently the white line next to the playhead is only displayed in the song editor. however, the white line is not seen within the piano roll editors and the automation editor. I suggest you add the head line to have a better view of the playback head.

That way I would be inside the piano roll editor.
![Cabezal en piano roll](https://user-images.githubusercontent.com/26127693/58923364-64fae680-870d-11e9-8e5e-12a103ce06f4.png)

And so it would look inside the automation editor.
![Cabezal en ed de automatizacion](https://user-images.githubusercontent.com/26127693/58923400-97a4df00-870d-11e9-8892-c8ede00014cb.png)

qnebra:
Great UX improvement idea.

Veratil:
I've added this in my PR for the Piano Roll #5253

PR #5543, opened 2020-06-22
Adding a trail (gradient) behind the position bar

russiankumar:
Hey all, this is my first commit to LMMS.

This commit adds a little aesthetic feature similar to one in FL studio, where there's a little gradient right before the bar. It looks nice, and it makes the bar a little easier to see on large-res screens (without having to look at the triangle at the top).

Here's what it looks like:

Screenshot from 2020-06-22 19-05-36

Currently there's a bug (or an intended feature) which doesn't allow selecting anything under the bar, and thus under the gradient too. I'm trying to figure out a fix, though guidance would really help as I'm new to contributing to LMMS and don't know its ins-and-outs quite well yet.

Thanks!

### DIFF
--- a/include/SongEditor.h
+++ b/include/SongEditor.h
@@ -52,7 +52,7 @@ class positionLine : public QWidget
 	Q_PROPERTY ( bool tailGradient READ hasTailGradient WRITE setHasTailGradient )
 	Q_PROPERTY ( QColor lineColor READ lineColor WRITE setLineColor )
 public:
-	positionLine ( QWidget* parent, ComboBoxModel* zoom );
+	positionLine ( QWidget* parent );
 	
 	// qproperty access functions
 	bool hasTailGradient () const;
@@ -60,15 +60,14 @@ public:
 	QColor lineColor () const;
 	void setLineColor ( const QColor & c );
 
+public slots:
+	void zoomChange (double zoom);
+
 private:
 	void paintEvent( QPaintEvent* pe ) override;
 	
 	bool m_hasTailGradient;
 	QColor m_lineColor;
-	
-	// to accomodate the change in size by zoom
-	ComboBoxModel* m_currentZoom;
-	static const QVector<double> s_zoomLevels;
 
 };
 
@@ -181,6 +180,9 @@ private:
 	bool m_selectRegion;
 
 	friend class SongEditorWindow;
+
+signals:
+	void zoomingValueChanged( double );
 } ;
 
 
--- a/include/SongEditor.h
+++ b/include/SongEditor.h
@@ -51,21 +51,21 @@ Q_DECLARE_METATYPE(QLinearGradient)
 class positionLine : public QWidget
 {
 	Q_OBJECT
-	Q_PROPERTY( bool tailGradient READ tailGradient WRITE setTailGradient )
-	Q_PROPERTY( QColor lineColor READ lineColor WRITE setLineColor )
+	Q_PROPERTY ( bool tailGradient READ tailGradient WRITE setTailGradient )
+	Q_PROPERTY ( QColor lineColor READ lineColor WRITE setLineColor )
 public:
-	positionLine( QWidget * parent, ComboBoxModel*, Song* );
+	positionLine ( QWidget * parent, ComboBoxModel * zoom, Song * song );
 	
 	int width ();
-	void go (int, int);
+	void go ( int x, int y );	// NOTE: Check SongEditor.cpp for usage
 	void zoomUpdate ();
 	void playStateChanged ();
 	
 	// qproperty access functions
-	bool tailGradient() const;
-	void setTailGradient( const bool & g );
-	QColor lineColor() const;
-	void setLineColor( const QColor & c );
+	bool tailGradient () const;
+	void setTailGradient ( const bool & g );
+	QColor lineColor () const;
+	void setLineColor ( const QColor & c );
 
 private:
 	void paintEvent( QPaintEvent * pe ) override;
@@ -77,13 +77,13 @@ private:
 	QColor m_lineColor;
 	
 	// to accomodate the change in size by zoom
-	ComboBoxModel* currentZoom;
+	ComboBoxModel * currentZoom;
 	static const QVector<double> m_zoomLevels;
 	
 	// to remove gradient when the line is not moving
-	Song * p_song;
+	Song * p_song;	// NOTE: p here stands for pointer
 
-} ;
+};
 
 
 class SongEditor : public TrackContainerView
--- a/include/SongEditor.h
+++ b/include/SongEditor.h
@@ -54,7 +54,12 @@ class positionLine : public QWidget
 	Q_PROPERTY( bool tailGradient READ tailGradient WRITE setTailGradient )
 	Q_PROPERTY( QColor lineColor READ lineColor WRITE setLineColor )
 public:
-	positionLine( QWidget * parent );
+	positionLine( QWidget * parent, ComboBoxModel*, Song* );
+	
+	int width ();
+	void go (int, int);
+	void zoomUpdate ();
+	void playStateChanged ();
 	
 	// qproperty access functions
 	bool tailGradient() const;
@@ -65,8 +70,18 @@ public:
 private:
 	void paintEvent( QPaintEvent * pe ) override;
 	
+	int m_width;
+	int m_x, m_y;	// NOTE: m_x is the playback position, not the position where the tail starts
+	
 	bool m_tailGradient;
 	QColor m_lineColor;
+	
+	// to accomodate the change in size by zoom
+	ComboBoxModel* currentZoom;
+	static const QVector<double> m_zoomLevels;
+	
+	// to remove gradient when the line is not moving
+	Song * p_song;
 
 } ;
 
@@ -129,6 +144,8 @@ private slots:
 	void updateScrollBar(int len);
 
 	void zoomingChanged();
+	
+	void playbackStateChanged();
 
 private:
 	void keyPressEvent( QKeyEvent * ke ) override;
--- a/include/SongEditor.h
+++ b/include/SongEditor.h
@@ -49,21 +49,21 @@ class TimeLineWidget;
 class positionLine : public QWidget
 {
 	Q_OBJECT
-	Q_PROPERTY ( bool tailGradient READ tailGradient WRITE setTailGradient )
+	Q_PROPERTY ( bool tailGradient READ hasTailGradient WRITE setHasTailGradient )
 	Q_PROPERTY ( QColor lineColor READ lineColor WRITE setLineColor )
 public:
 	positionLine ( QWidget* parent, ComboBoxModel* zoom );
 	
 	// qproperty access functions
-	bool tailGradient () const;
-	void setTailGradient ( const bool & g );
+	bool hasTailGradient () const;
+	void setHasTailGradient ( const bool g );
 	QColor lineColor () const;
 	void setLineColor ( const QColor & c );
 
 private:
 	void paintEvent( QPaintEvent* pe ) override;
 	
-	bool m_tailGradient;
+	bool m_hasTailGradient;
 	QColor m_lineColor;
 	
 	// to accomodate the change in size by zoom
--- a/include/SongEditor.h
+++ b/include/SongEditor.h
@@ -46,20 +46,13 @@ class Song;
 class TextFloat;
 class TimeLineWidget;
 
-Q_DECLARE_METATYPE(QLinearGradient)
-
 class positionLine : public QWidget
 {
 	Q_OBJECT
 	Q_PROPERTY ( bool tailGradient READ tailGradient WRITE setTailGradient )
 	Q_PROPERTY ( QColor lineColor READ lineColor WRITE setLineColor )
 public:
-	positionLine ( QWidget * parent, ComboBoxModel * zoom, Song * song );
-	
-	int width ();
-	void go ( int x, int y );	// NOTE: Check SongEditor.cpp for usage
-	void zoomUpdate ();
-	void playStateChanged ();
+	positionLine ( QWidget* parent, ComboBoxModel* zoom );
 	
 	// qproperty access functions
 	bool tailGradient () const;
@@ -68,20 +61,14 @@ public:
 	void setLineColor ( const QColor & c );
 
 private:
-	void paintEvent( QPaintEvent * pe ) override;
-	
-	int m_width;
-	int m_x, m_y;	// NOTE: m_x is the playback position, not the position where the tail starts
+	void paintEvent( QPaintEvent* pe ) override;
 	
 	bool m_tailGradient;
 	QColor m_lineColor;
 	
 	// to accomodate the change in size by zoom
-	ComboBoxModel * currentZoom;
+	ComboBoxModel* m_currentZoom;
 	static const QVector<double> m_zoomLevels;
-	
-	// to remove gradient when the line is not moving
-	Song * p_song;	// NOTE: p here stands for pointer
 
 };
 
@@ -144,8 +131,6 @@ private slots:
 	void updateScrollBar(int len);
 
 	void zoomingChanged();
-	
-	void playbackStateChanged();
 
 private:
 	void keyPressEvent( QKeyEvent * ke ) override;
--- a/include/SongEditor.h
+++ b/include/SongEditor.h
@@ -28,6 +28,7 @@
 #define SONG_EDITOR_H
 
 #include <QVector>
+#include <QLinearGradient>
 
 #include "ActionGroup.h"
 #include "Editor.h"
@@ -45,13 +46,27 @@ class Song;
 class TextFloat;
 class TimeLineWidget;
 
+Q_DECLARE_METATYPE(QLinearGradient)
+
 class positionLine : public QWidget
 {
+	Q_OBJECT
+	Q_PROPERTY( bool tailGradient READ tailGradient WRITE setTailGradient )
+	Q_PROPERTY( QColor lineColor READ lineColor WRITE setLineColor )
 public:
 	positionLine( QWidget * parent );
+	
+	// qproperty access functions
+	bool tailGradient() const;
+	void setTailGradient( const bool & g );
+	QColor lineColor() const;
+	void setLineColor( const QColor & c );
 
 private:
 	void paintEvent( QPaintEvent * pe ) override;
+	
+	bool m_tailGradient;
+	QColor m_lineColor;
 
 } ;
 
--- a/include/SongEditor.h
+++ b/include/SongEditor.h
@@ -68,7 +68,7 @@ private:
 	
 	// to accomodate the change in size by zoom
 	ComboBoxModel* m_currentZoom;
-	static const QVector<double> m_zoomLevels;
+	static const QVector<double> s_zoomLevels;
 
 };
 
--- a/src/gui/editors/SongEditor.cpp
+++ b/src/gui/editors/SongEditor.cpp
@@ -52,12 +52,21 @@
 #include "PianoRoll.h"
 #include "Track.h"
 
-positionLine::positionLine( QWidget * parent ) :
+const QVector<double> positionLine::m_zoomLevels =
+		{ 0.125f, 0.25f, 0.5f, 1.0f, 2.0f, 4.0f, 8.0f, 16.0f };
+
+positionLine::positionLine( QWidget * parent, ComboBoxModel* zoom, Song* song ) :
 	QWidget( parent ),
 	m_tailGradient ( false ),
 	m_lineColor (0, 0, 0, 0)
 {
-	setFixedWidth( 8 );
+	m_x = 0;
+	m_y = 0;
+	m_width = 8;
+	currentZoom = zoom;
+	p_song = song;
+	resize( m_width, height() );
+	
 	setAttribute( Qt::WA_NoSystemBackground, true );
 	setAttribute( Qt::WA_TransparentForMouseEvents );
 }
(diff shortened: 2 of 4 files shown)


Spekular:
I suspect the cursor is recieving click events and not forwarding them to the song editor, hence why it blocks you from interacting with things beneath it.

I'd also request that you don't hard code the gradient for this. Instead, add a property to the theme and read from there. Perhaps the classic (old/legacy) theme can have a fully transparent gradient, leaving it with no "tail".

russiankumar:
> I suspect the cursor is recieving click events and not forwarding them to the song editor, hence why it blocks you from interacting with things beneath it.

I'll try to fix up on this. I take it this was always the case with the position bar, so I may need some help figuring it out.

> I'd also request that you don't hard code the gradient for this. Instead, add a property to the theme and read from there. Perhaps the classic (old/legacy) theme can have a fully transparent gradient, leaving it with no "tail".

That seems like a valid idea, and I'll definitely make a property like that. I could either have a true/false property or a setting for alpha value on the right side of the gradient. I suspect the alpha setting will be more popular.

I'm very new to contributing, and the source code of LMMS in general, so is it okay if I pester you with some questions (on this thread) if I don't get something? Pardon my incompetence, but I know C++ fairly well and really want to contribute something to LMMS :D

Spekular:
> That seems like a valid idea, and I'll definitely make a property like that. I could either have a true/false property or a setting for alpha value on the right side of the gradient. I suspect the alpha setting will be more popular.

Gradients are supported as theme properties, so just use that. It gives full flexibility to theme authors to use different colors, slopes, etc.

> I'm very new to contributing, and the source code of LMMS in general, so is it okay if I pester you with some questions (on this thread) if I don't get something?

I'm not the most familiar with this part of the code either, but yeah post any questions you have here (or in #techtalk/#devonly on Discord) and hopefully someone can help out.

russiankumar:
I think having a gradient property would be hard, because as far as I understand, the theme is loaded using LmmsStyle.cpp, and the gradient it returns is a vertical one:

```
QLinearGradient getGradient( const QColor & _col, const QRectF & _rect )
{
	QLinearGradient g( _rect.topLeft(), _rect.bottomLeft() );
```

I think changing that would break the entire GUI.

So unless I'm wrong and this is not the function to get my gradient, either I write a `getHorizontalGradient ()` function, or make it an on/off thing.

russiankumar:
Okay, three updates:

1. `getGradient` was NOT the function to use.
2. When I did find the correct function to use, QT would not read the gradient from the stylesheet, but when changed to an on/off boolean arrangement, it works.
3. As a result, the style sheet will not have gradient control but an on/off (`tailGradient`) and a color for the line + gradient (`lineColor`)

Note, I forgot to change the valur from RGBA to RGB in the classic theme, so that will come in a further commit. Also the code setting the color is VERY messy, and a style review would really help.

russiankumar:
> Note, I forgot to change the value from RGBA to RGB in the classic theme, so that will come in a further commit.

This has been fixed.

> I suspect the cursor is recieving click events and not forwarding them to the song editor, hence why it blocks you from interacting with things beneath it.

This has also been fixed (with surprisingly less effort than it took to make a tail toggle).

vlad0337187:
It looks great =)

src/gui/editors/SongEditor.cpp
```diff
@@ -808,7 +829,7 @@ void SongEditor::updatePosition( const MidiTime & t )
 	if( x >= trackOpWidth + widgetWidth -1 )
 	{
 		m_positionLine->show();
-		m_positionLine->move( x, m_timeLine->height() );
+		m_positionLine->move( x-7, m_timeLine->height() );
```
SecondFlight, line 832: A bit of a nitpick, but could the 7 (and the 8 above) be either 1) sourced from the theme or 2) made to use the same constant variable above? This is the kind of magic number that can cause bugs, i.e. where two separate magic numbers need to have a certain relationship to be valid.

src/gui/editors/SongEditor.cpp
```diff
@@ -53,19 +53,40 @@
...
+	if (m_tailGradient) 
+	{ 
+		gradient.setColorAt(0.875, 
+			QColor (m_lineColor.red(), m_lineColor.green(), m_lineColor.blue(), 60) );
+	}
+	else 
+	{ 
+		gradient.setColorAt(0.875, 
+			QColor (m_lineColor.red(), m_lineColor.green(), m_lineColor.blue(), 0) );
+	}
```
SecondFlight, line 86: Again, a total nitpick, but might I suggest something like:

```cpp
	int opacity;
	if (m_tailGradient) 
	{ 
		opacity = 60;
	}
	else 
	{ 
		opacity = 0;
	}
	gradient.setColorAt(0.875, 
		QColor (m_lineColor.red(), m_lineColor.green(), m_lineColor.blue(), opacity) );
```

SecondFlight:
Just an idea: have you thought about making the gradient disappear when not playing? FL does this, and it makes the gradient give a feel of motion. FL also resizes the gradient depending on the current zoom, but it also has floating point zoom levels, so /shrug

src/gui/editors/SongEditor.cpp
DomClark: I would favour a ternary expression here; it's always nice if you can initialise variables at point of declaration, and mark them const too. (And it uses far less vertical space too.)
```cpp
const auto opacity = m_tailGradient ? 60 : 0;
```

Alternatively, to avoid manually copying the red, green, and blue components, which might be a bit easier to read:
```cpp
auto lineColor = m_lineColor;
lineColor.setAlpha(m_tailGradient ? 60 : 0);
gradient.setColorAt(0.875, lineColor);
```

SecondFlight: Oh of course, that didn't even cross my mind. Good suggestion!

russiankumar: > to avoid manually copying the red, green, and blue components, which might be a bit easier to read

Something I was looking to do as well. I'll implement the last one, that seems the most straightforward and short solution. Thanks for the suggestion!

russiankumar:
@SecondFlight , 

> A bit of a nitpick, but could the 7 (and the 8 above) be either 1) sourced from the theme or 2) made to use the same constant variable above?

Option 1 over here and

> FL also resizes the gradient depending on the current zoom

are kind of a bit conflicting, though definitely both of them could work together, would make the setting in the theme not straightforward at all. So what I'm suggesting is to not let the thickness be theme-based, but to instead have it depend on the current zoom level.

> Just an idea: have you thought about making the gradient disappear when not playing?

Not thought of this, but that's what the community is for. I suppose I'll have to retrieve whether the song is playing or not, and make opacity statement a bit more like `if (m_tailGradient && isPlaying) ...`

If you're okay with this, I'll take some more time, and work on it.

> Probably overkill unless you think it would be worth the effort.

Well, if I'm contributing, I'd rather not leave the feature half-complete.

Thanks for your review!

SecondFlight:
> a bit conflicting

Agreed, the two aren't compatible. Just spit-balling ideas. Do what seems best to you :)

> If you're okay with this

That's what I had in mind.

> I'd rather not leave the feature half-complete

I'd say the feature is complete when you're happy with it. My last comment wasn't intended as a review.

... Also, turns out my memory is completely wrong. FL doesn't resize the cursor based on the speed:

5aYQkVlhGA
ac350btJ95

Sorry about that.

russiankumar:
> FL doesn't resize the cursor based on the speed

That makes my life significantly easier :D

> Do what seems best to you :)

I guess I'll implement the arrangement I discussed before.

Commits coming soon!

russiankumar:
> have it depend on the current zoom level.

Implemented this.

> making the gradient disappear when not playing

This as well.

I'll ask for a re-review, and if there's no other things left for me to change/add/remove, we're done here :)

src/gui/editors/SongEditor.cpp
russiankumar: > the number here must always be one less than the number above.

Yes, that's right. Fixed it in the new commit, along with other things.

SecondFlight:
Tail length doesn't depend on anything in FL, zoom included. When I said speed, I was thinking that the tail would move faster when zoomed in. Should have said it more clearly.

That said, I think having the tail length depend on zoom will look quite nice :)

I'll have some time later to look at the code.

russiankumar:
Oh well, we don’t need an exact replica of FL anyways, do we? In my opinion it’s a nice little feature, so I’d vouch for keeping it how it is right now.

SecondFlight:
Just tested it and it looks good, haven't found any major issues so far.

I did find something minor: the gradient shows up in the song editor if the piano roll or B&B editors are playing, even though the cursor isn't moving in this case.

russiankumar:
Oh, that must be because it still thinks the song is playing. Any idea how we can know if the current window is active? Especially the song editor since that’s what we need here.

PhysSong:
@russiankumar Do you want to check if LMMS is playing the song or something else such as a pattern? If it's the case, `Song::playMode()` is what you need.

russiankumar:
> If it's the case, Song::playMode() is what you need.

Exactly what I need. Thanks @PhysSong !

russiankumar:
> If it's the case, Song::playMode() is what you need.

Exactly what I need. Thanks @PhysSong!

russiankumar:
Phew... I think that's all the fixes done. You can review the code and if there's nothing else (there shouldn't be, I think the major bugs are patched), this is completed.

SecondFlight:
I'll do a bit more testing this evening and merge it tomorrow if there are no objections.

russiankumar:
Cool! Let me know if there’s any more bugs, otherwise I think it’s ready for merging.

DomClark:
I'd like to suggest some changes before this gets merged.

src/gui/editors/SongEditor.cpp
```diff
@@ -108,8 +193,10 @@ SongEditor::SongEditor( Song * song ) :
 			this, SLOT( selectRegionFromPixels( int, int ) ) );
 	connect( m_timeLine, SIGNAL( selectionFinished() ),
 			 this, SLOT( stopRubberBand() ) );
+	connect( m_song, SIGNAL( playbackStateChanged() ),
+			 this, SLOT( playbackStateChanged() ) );
 
-	m_positionLine = new positionLine( this );
+	m_positionLine = new positionLine( this, m_zoomingModel, m_song );
```
DomClark, line 199: Instead of passing `m_zoomingModel` to the position line, consider listening for its `dataChanged` signal and then updating the position line's width.

src/gui/editors/SongEditor.cpp
```diff
@@ -52,26 +52,111 @@
...
+	m_y = y;
+}
+
+// update width if zoom changes
+void positionLine::zoomUpdate()
+{ move ( m_x - width() + 1, m_y ); }
+
+// Essentially repaints so that gradient does not persist when stopped (this is a bug fix)
+void positionLine::playStateChanged()
+{ repaint(); }
```
DomClark, line 139: Prefer `update` to `repaint`: https://doc.qt.io/qt-5/qwidget.html#repaint.

include/SongEditor.h
```diff
@@ -45,15 +46,44 @@ class Song;
...
+	bool tailGradient () const;
+	void setTailGradient ( const bool & g );
+	QColor lineColor () const;
+	void setLineColor ( const QColor & c );
 
 private:
 	void paintEvent( QPaintEvent * pe ) override;
+	
+	int m_width;
+	int m_x, m_y;	// NOTE: m_x is the playback position, not the position where the tail starts
```
DomClark, line 74: Is it possible to get rid of `m_width`, `m_x`, and `m_y` and just use the position and size properties inherited from `QWidget`? Redundant information is just a place for bugs to hide should it ever get out of sync, and the code is easier to read if it sticks to standard methods.

SecondFlight:
Did a quick once-over bug test and it seems okay.

src/gui/editors/SongEditor.cpp
```diff
@@ -837,9 +924,14 @@ void SongEditor::zoomingChanged()
...
 	updateRubberband();
+	
+	m_positionLine->zoomUpdate();
 }
 
-
+void SongEditor::playbackStateChanged()
+{
+	m_positionLine->playStateChanged();
+}
```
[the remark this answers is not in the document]
PhysSong: Since Qt doesn't support metaobject features for nested classes, you can't use the old `SLOT(playStateChanged())` syntax. It's possible to use the new `&positionLine::playStateChanged()` syntax or a lambda function, though.
@DomClark's second suggestion also makes sense if we're not going to add something else to `positionLine::playStateChanged()`.

include/SongEditor.h
```diff
@@ -45,15 +46,44 @@ class Song;
...
+	
+	bool m_tailGradient;
+	QColor m_lineColor;
+	
+	// to accomodate the change in size by zoom
+	ComboBoxModel * currentZoom;
```
PhysSong, line 80: The coding convention says you should add the `m_` prefix.

include/SongEditor.h
```diff
@@ -45,15 +46,44 @@ class Song;
...
+	static const QVector<double> m_zoomLevels;
```
PhysSong, line 81: For static members, you need to use `s_` instead of `m_`.

include/SongEditor.h
```diff
@@ -45,15 +46,44 @@ class Song;
 class TextFloat;
 class TimeLineWidget;
 
+Q_DECLARE_METATYPE(QLinearGradient)
```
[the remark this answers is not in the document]
russiankumar: Oh, I left it here from a previous test. Will be removed.

russiankumar:
@DomClark & @SecondFlight,

In short, the following changes **were** made:

- Remove `Q_DECLARE_METATYPE(QLinearGradient)`
- Remove `m_x`, `m_y` and `m_width`
- Replace `p_song` with `Engine::getSong()`
- Replace all `repaint()`s with `update()`s
- Replace `currentZoom` with `m_currentZoom`
- Connect `Song::playbackStateChanged` to `positionLine::update()`
- Use `update()` instead of `positionLine::zoomingUpdate()`

I have **not** removed the zooming model as an attribute for the constructor as I haven't figured out a way to point to `m_zoomingModel->value()`. If there is a function similar to `Engine::getSong()` for the zooming model, I could use that instead. If not, I think these commits should should cover all of your requested changes.

I've requested a re-review from @DomClark since the review was yours. @SecondFlight, I've done a tiny bit of bug testing and the code works more or less the same as it did prior to the changes. If you want to, you could test with these changes, though I don't think anything should have changed.

include/SongEditor.h
russiankumar: It's possible, I've figured out a way to get around without them. See commit.

src/gui/editors/SongEditor.cpp
russiankumar: All `repaint()`s were changed to `update()`s

src/gui/editors/SongEditor.cpp
russiankumar: I'm already listening for the signal, I use the attribute as a way to get the zooming value.

src/gui/editors/SongEditor.cpp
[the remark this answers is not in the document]
russiankumar: I connected `Song::playbackStateChanged()` to `positionLine::update()`

src/gui/editors/SongEditor.cpp
```diff
@@ -52,26 +52,111 @@
...
+	
+	// If width > 1, we need the gradient
+	else
+	{
+		// Create the gradient trail behind the line
+		QLinearGradient gradient( rect().bottomLeft(), rect().bottomRight() );
+		
+		// If gradient is enabled, we're in focus and we're playing, enable gradient
+		if (p_song->isPlaying() && m_tailGradient && 
+			p_song->playMode() == Song::Mode_PlaySong)
```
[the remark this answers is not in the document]
russiankumar: Noted and implemented.

src/gui/editors/SongEditor.cpp
```diff
@@ -52,26 +52,111 @@
...
+	QWidget( parent ),
+	m_tailGradient ( false ),
+	m_lineColor (0, 0, 0, 0)
 {
-	setFixedWidth( 1 );
+	m_x = 0;
+	m_y = 0;
+	m_width = 8;
+	currentZoom = zoom;
+	p_song = song;
```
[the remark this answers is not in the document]
russiankumar: Most of these have been removed.

include/SongEditor.h
```diff
@@ -45,15 +46,44 @@ class Song;
...
+	positionLine ( QWidget * parent, ComboBoxModel * zoom, Song * song );
+	
+	int width ();
+	void go ( int x, int y );	// NOTE: Check SongEditor.cpp for usage
+	void zoomUpdate ();
+	void playStateChanged ();
+	
+	// qproperty access functions
+	bool tailGradient () const;
+	void setTailGradient ( const bool & g );
```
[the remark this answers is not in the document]
russiankumar: Didn't exactly get what you mean by this...?

include/SongEditor.h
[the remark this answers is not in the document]
russiankumar: > Will be removed.

Has been removed.

include/SongEditor.h
russiankumar: Renamed `currentZoom` to `m_currentZoom`

include/SongEditor.h
russiankumar: Renamed `m_zoomLevels` to `s_zoomLevels`. I'd like to point out that it is named as `m_zoomLevels` in the Song Editor class as well (since the implementation was a copy-paste).

include/SongEditor.h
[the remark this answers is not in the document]
DomClark: The `&` in `const bool & g` means that rather than passing the boolean value directly, a reference to a value stored elsewhere is passed. Typically references are used if the original object passed to the function needs to modified, or if the object is expensive to copy. Neither of these are the case here, so `bool g` is sufficient for the parameter.

(Minor nitpick - perhaps `hasTailGradient` and `setHasTailGradient` would be better names? The current ones sound like they access the gradient itself, rather than whether it is present or not.)

src/gui/editors/SongEditor.cpp
DomClark: You could pass the zooming value as a parameter to `positionLine::zoomUpdate`. In fact, this would allow you to pass `m_zoomLevels[m_zoomingModel->value()]`, which means you don't need to duplicate `m_zoomLevels` in `positionLine`.

src/gui/editors/SongEditor.cpp
```diff
@@ -52,26 +52,84 @@
...
+		{
+			gradient.setColorAt((double)( ( width() - 1.0f )/width() ),
+				QColor( m_lineColor.red(), m_lineColor.green(), m_lineColor.blue(), 0) );
+		}
+		
+		// Fill in the remaining parts
+		gradient.setColorAt(0,
+			QColor( m_lineColor.red(), m_lineColor.green(), m_lineColor.blue(), 0) );
+		gradient.setColorAt(1,
+			QColor( m_lineColor.red(), m_lineColor.green(), m_lineColor.blue(), 153) );
```
DomClark, line 107: Sorry, I forgot to mention this in the previous review. Did you intend to switch to using the suggestion I made before?
```cpp
		// Create the gradient trail behind the line
		QLinearGradient gradient(rect().bottomLeft(), rect().bottomRight());
		auto lineColor = m_lineColor;

		// If gradient is enabled, we're in focus and we're playing, enable gradient
		const auto enableGradient = Engine::getSong()->isPlaying() && m_tailGradient
			&& Engine::getSong()->playMode() == Song::Mode_PlaySong;
		lineColor.setAlpha(enableGradient ? 60 : 0);
		gradient.setColorAt((width() - 1.0) / width(), lineColor);

		// Fill in the remaining parts
		lineColor.setAlpha(0);
		gradient.setColorAt(0, lineColor);
		lineColor.setAlpha(153);
		gradient.setColorAt(1, lineColor);
```

russiankumar: I actually decided not to do this because although my code is larger, it’s easily understood and doesn’t involve creating new variables. Your code works just as well but I just thought the original code is more friendly to people new to the code. If you do want me to implement this however, I’d be glad to.

include/SongEditor.h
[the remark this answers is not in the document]
russiankumar: Well, I wrote that function based off of the other functions for reading qproprties, and all of them had the & symbol there. Maybe it’s something QT requires? Either ways, I’d rather be on the safe side here especially when the rest of the code uses references for bool.

About the tail gradient rename, I’ll commit with the modified name soon.

src/gui/editors/SongEditor.cpp
DomClark: I don't mind either way - I just wondered whether you'd forgotten since you said you were going to change it. If you prefer the old way, let's go with that.

That said, would you mind removing the `double` casts? C-style casts are usually avoided in C++, and since `float` can implicitly convert to `double`, no cast is needed anyway. You could even change `1.0f` to `1.0`, so the calculation only uses `double`s in the first place.

include/SongEditor.h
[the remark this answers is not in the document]
DomClark: Most of the other `Q_PROPERTY` setters in LMMS work with `QColor`, which comes under "expensive to copy". It's not actually that expensive, but the rule of thumb is generally "has a user-defined copy constructor, or is larger than a pointer", both of which `QColor` satisfies. Other `Q_PROPERTY(bool)`s in LMMS don't use references, for example:
https://github.com/LMMS/lmms/blob/5d7e6720e1ba0c154041fd663801f7a7a0e16cb5/include/PianoRoll.h#L73
https://github.com/LMMS/lmms/blob/5d7e6720e1ba0c154041fd663801f7a7a0e16cb5/include/PianoRoll.h#L156

src/gui/editors/SongEditor.cpp
russiankumar: Well, I've removed `zoomUpdate()` because it was redundant (I connected the signal to `update()`). I'll rewrite it though.

include/SongEditor.h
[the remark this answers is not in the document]
russiankumar: Okay then, I'll make it a `bool`.

src/gui/editors/SongEditor.cpp
russiankumar: Is it acceptable to keep it this way or do I need a rewrite?

src/gui/editors/SongEditor.cpp
russiankumar: Yep, I've removed the `double` casts.

russiankumar:
I hope that it's finally complete now xD.

I've made some final changes as requested by @DomClark, and hopefully, _hopefully_ the code is ready for merge.

As always, I've asked for a re-review.

To everyone: thanks for helping me out!

russiankumar:
Umm... any updates...?

SecondFlight:
I'll merge this in a day or two if there are no objections 👍

DomClark:
Sorry for getting back to you late - I had some other stuff on.

I don't mean to go on about it, but you're still passing the `SongEditor`'s `m_zoomingModel` to the `positionLine` to get the value. The reason that bothers me is that it couples the position line to the internals of the song editor. Suppose we changed `SongEditor` to store its zoom in a `FloatModel`; we would then have to update `positionLine` too. Suppose we refactored all the editors' zoom models into one base class; that base class would then have to expose its zoom model more widely in order for `positionLine` to get access to it. Suppose we reused `positionLine` with multiple editors that stored their zoom in different ways (or couldn't be zoomed at all); then we'd have to think of something completely different.

Generally when writing software, you aim to keep components as independent as possible, communicating through public interfaces rather than sharing internals. This makes it easier to change individual parts of the software without requiring subsequent changes throughout the codebase, and makes it easier to reuse components elsewhere. (Admittedly, this isn't something LMMS does very well at the moment.) `positionLine` shouldn't need to know how its parent editor handles zooming - all it needs to know is how much it should zoom by. If you can get `SongEditor` to tell it the new zoom value whenever the zoom changes, I think that would be good, and it would make it easier for later improvents like #5253 and #5008.

russiankumar:
Alright man, I'm convinced.

I've added a `zoomingValueChanged()` signal in the song editor class, and connected it to a slot in `positionLine`. This made redundant the `s_zoomingValues` array so that has been removed too. I discovered a new bug in the meantime and fixed that too.

Do review this new code, though I hope I've done everything you've asked for here.

Also, I think #5253 and the likes will have to rewrite some code to accomodate for these changes (the entire PR in general has to be accomodated for), though I hope that will be easily done.

src/gui/editors/SongEditor.cpp
```diff
@@ -124,6 +120,17 @@ QColor positionLine::lineColor() const
 void positionLine::setLineColor( const QColor & c )
 { m_lineColor = c; }
 
+// NOTE: the move() implementation fixes a bug where the position line would appear
+// in an unexpected location when positioned at the start of the track
+void positionLine::zoomChange( double zoom )
+{
+	move( x() + width() - 1, y() );
+	resize( 8.0 * zoom, height() );
+	move( x() - width() + 1, y() );
```
russiankumar, line 129: Actually, I think I could change this up a little by using a temporary variable. I think that would make it quicker too without actually moving anything.

russiankumar: I'll commit by today evening with a fix.

russiankumar: Done.

src/gui/editors/SongEditor.cpp
```diff
@@ -124,9 +124,10 @@ void positionLine::setLineColor( const QColor & c )
 // in an unexpected location when positioned at the start of the track
 void positionLine::zoomChange( double zoom )
 {
-	move( x() + width() - 1, y() );
+	int m_x = x() + width() - 1;
+	
 	resize( 8.0 * zoom, height() );
-	move( x() - width() + 1, y() );
+	move( m_x - width() + 1, y() );
```
SecondFlight: The `m_` prefix is reserved for mutable class fields. Since this is just a function-scope variable, it shouldn't have the prefix.

SecondFlight:
That's all I got. I'll let @DomClark comment his approval on the latest change before merging.

src/gui/editors/SongEditor.cpp
russiankumar: Renamed to `X` (would let me name it `x` as there is a function with that name).

russiankumar:
It appears I've made the same commit twice, but I removed a commented out piece of code in the second commit and wasn't sure if both the commits would upload. Well, now I know my answer :D

I think we're done here?

russiankumar:
> something more descriptive as to which x position this is

@Spekular per your request, it's now called `playHeadPos`.

PR #5547, opened 2020-06-26
Changes the behavior of "solo" so it doesn't mute Automation Tracks

IanCaio:
This PR changes the `Track::toggleSolo` method so it doesn't mute Automation Tracks, keeping their original state (either already muted or not muted). I believe this behavior is more desirable than the current behavior of muting every track including Automation Tracks. Usually, when a user toggle solo on a track, the idea is to hear how that particular track sounds alone and the automations that affect it are part of how it will sound on the final mix.

It was mentioned on the discord server that Automation Tracks can also automate the mute and solo states. I've included a project file with a very simple automation of solo and mute states as a proof of concept of how it behaves with the current solo method and how it would behave with the changes included on this PR. The way the solo works now, if there's an Automation Track that automates the solo state, as soon as the solo gets toggled on another track, the Automation Track itself will be muted and it won't be able to turn the solo state off. With the suggested change, since Automation Tracks won't have the mute state changed when another track's solo is enabled, they will be able to control the solo state even after turning on the solo from another track. In my opinion that's also a more desirable behavior.

[solotest.zip](https://github.com/LMMS/lmms/files/4839566/solotest.zip)

# Update

After chatting with other LMMS users and devs, we came to an agreement that an option under LMMS settings could be added to return the behavior of the "solo" button to the old one. This setting was added on the latest commit (Settings -> General -> "Use solo legacy behavior") and the code was changed to take this value under consideration. An upgrade method was added to `ConfigManager` called `upgrade_1_3_0` which upgrades the configuration file to have this new attribute. The default value is to use the new behavior.


src/core/Track.cpp
```diff
@@ -2653,7 +2653,14 @@ void Track::toggleSolo()
 			{
 				( *it )->m_mutedBeforeSolo = ( *it )->isMuted();
 			}
-			( *it )->setMuted( *it == this ? false : true );
+			// Don't mute AutomationTracks (keep their original state)
+			if( *it == this ){
+				( *it )->setMuted( false );
+			} else {
+				if( ( *it )->type() != AutomationTrack ){
```
DomClark, line 2660: `else if`?

DomClark:
Currently, if you solo a track, change the mute state of some other tracks, and unsolo the first track again, then the mute state of the other tracks is reset to how it was before soloing. If this PR makes it so soloing tracks doesn't affect automation tracks, should it also make it so unsoloing tracks doesn't affect them either?

src/core/Track.cpp
IanCaio: My bad! I'll update it with the `else if`

IanCaio:
> Currently, if you solo a track, change the mute state of some other tracks, and unsolo the first track again, then the mute state of the other tracks is reset to how it was before soloing. If this PR makes it so soloing tracks doesn't affect automation tracks, should it also make it so unsoloing tracks doesn't affect them either?

Makes sense, you mean only for the Automation Tracks right? I can add a conditional before restoring the mute state to check if it's an Automation Track and only restore it if it isn't.

IanCaio:
Added the possibility to use the legacy behavior of the solo button (muting every track including the Automation Tracks) for anyone that prefers the old behavior. It can be changed under `Settings > General > "Use solo legacy behavior"`. It's the first time I work with the configuration file so a review regarding this change is appreciated! I added an upgrade method for the configuration file (`upgrade_1_3_0`), if it should be another version just let me know!

src/gui/SetupDialog.cpp
```diff
@@ -233,6 +235,8 @@ SetupDialog::SetupDialog(ConfigTabs tab_to_open) :
 		m_oneInstrumentTrackWindow, SLOT(toggleOneInstrumentTrackWindow(bool)), true);
 	addLedCheckBox("Show sidebar on the right-hand side", gui_tw, counter,
 		m_sideBarOnRight, SLOT(toggleSideBarOnRight(bool)), true);
+	addLedCheckBox("Use solo legacy behavior", gui_tw, counter,
```
Veratil, line 238: Maybe `Solo mutes Automation Tracks` or something along these lines? Someone coming into LMMS in the future would have to go look up what legacy behavior is (or test it out which they may not have automation tracks in and be confused). Changing it to explain what it will do I think is better than just saying `legacy behavior`.

IanCaio: You're right, I'll change it to something more descriptive!

IanCaio: I used `Mute automation tracks during solo` as a description of the setting now!

src/core/Track.cpp
```diff
@@ -2653,15 +2653,29 @@ void Track::toggleSolo()
...
 		{
-			( *it )->setMuted( ( *it )->m_mutedBeforeSolo );
+			// Unless we are on the sololegacybehavior mode, only restores the
+			// mute state if the track isn't an Automation Track
+			if( ConfigManager::inst()->value("app","sololegacybehavior").toInt() )
+			{
+				( *it )->setMuted( ( *it )->m_mutedBeforeSolo );
+			} else if( ( *it )->type() != AutomationTrack ) {
+				( *it )->setMuted( ( *it )->m_mutedBeforeSolo );
+			}
```
Veratil, line 2678: It looks like both branches do the same thing, so I don't see why we couldn't merge them with an `||`.
```suggestion
			if( ConfigManager::inst()->value("app","sololegacybehavior").toInt()
				|| ( *it )->type() != AutomationTrack ) {
				( *it )->setMuted( ( *it )->m_mutedBeforeSolo );
			}
```

src/core/Track.cpp
```diff
@@ -2653,15 +2653,29 @@ void Track::toggleSolo()
...
 			}
-			( *it )->setMuted( *it == this ? false : true );
+			// Don't mute AutomationTracks (keep their original state) unless we are on the sololegacybehavior mode
+			if( *it == this ){
+				( *it )->setMuted( false );
+			} else {
+				if( ConfigManager::inst()->value("app","sololegacybehavior").toInt() || ( *it )->type() != AutomationTrack ){
+					( *it )->setMuted( true );
+				}
+			}
```
Veratil, line 2663: The `else` is guarded with an `if`, so we can merge those blocks. Then since both `if` and `else if` are doing the same thing we should be able to merge with an `||`.
```suggestion
			if( *it == this || ConfigManager::inst()->value("app","sololegacybehavior").toInt() || ( *it )->type() != AutomationTrack ){
				( *it )->setMuted( true );
			}
```

Veratil (commented):
Maybe I'm misreading the if blocks, but it looks like we can merge some checks to reduce duplicated code.

src/core/Track.cpp
IanCaio: True! Sometimes I think twice before nesting `if`s with `||`s for the sake of keeping the logic clearer, but in this case I think it's clear enough. I'll merge `Master`, make the change and push it!

src/core/Track.cpp
IanCaio: This one however can't be merged because the `if` and the `else if` are doing different things (one is setting muted to `false` on the track being soloed and the other setting muted to `true` on the other tracks). I can merge the `else` with the `if` though, it will save two lines at least.

Veratil: > This one however can't be merged because the `if` and the `else if` are doing different things (one is setting muted to `false` on the track being soloed and the other setting muted to `true` on the other tracks). I can merge the `else` with the `if` though, it will save two lines at least.

Oops! You're right. I overlooked the true/false.

Veratil (commented):
Code LGTM! 👍

src/core/ConfigManager.cpp
```diff
@@ -158,6 +158,16 @@ void ConfigManager::upgrade_1_1_91()
...
 
 
+void ConfigManager::upgrade_1_3_0()
+{
+	// Adds the option of using the solo legacy behavior (muting Automation Tracks as well) if
+	// it isn't on the configuration file (safety check + upgrade)
+	if(value("app", "sololegacybehavior").isNull()) {
+		setValue("app", "sololegacybehavior", "0");
+	}
+}
```
DomClark, line 133: I don't think this upgrade function is necessary - we haven't written one when adding any other new values in the past. Instead we prefer to specify a default value, which covers more scenarios (e.g. when the configuration file doesn't yet exist). Upgrades are generally used instead to transform existing values.

IanCaio: Sure, I can delete the `upgrade` method and just add a default value to every call to `ConfigManager::value` that uses this entry. I'm writing the changes atm and will push them!

DomClark:
Currently the code re-reads the configuration value for every track in the loop - can it be pulled out into a local variable?

src/core/ConfigManager.cpp
```diff
@@ -123,6 +123,7 @@ void ConfigManager::upgrade_1_1_91()
 }
 
 
+
```
Veratil, line 126: Since we're not touching ConfigManager anymore, let's make this not touch the file either.

IanCaio: No problem, I didn't realize I added a blank line until I pushed the changes. On line 146 though I actually deleted a blank space that was probably left in the empty line. Should I leave it there too?

IanCaio:
> Currently the code re-reads the configuration value for every track in the loop - can it be pulled out into a local variable?

True, more efficient that way! I'll make the changes :+1:

src/core/ConfigManager.cpp
IanCaio: I left the Tab where it was to remove the file from the "Files changed" list. I believe it will probably be removed later when the code base is refactored through clang.

DomClark (approved):
Looks good to me.

Veratil (approved):

Spekular (approved):

Spekular:
Merged since there were 2 approvals and no arguments against it (code LGTM as well).

Issue #5548, opened 2020-06-27
Garbage in audio when exporting tracks


softrabbit:
### Bug Summary
Exporting tracks leaves a few samples of garbage at the start of the audio files. Might be some buffer that should be cleared after each render but isn't?

#### Steps to reproduce
Export tracks as 16-bit WAV @ 44100 Hz.

#### Expected behavior
No garbage at start of audio files.

#### Actual behavior
Garbage at start of audio files, except for the first one rendered. 

#### Screenshot
Ardour showing a couple of imported tracks:
![ardour-shot](https://user-images.githubusercontent.com/6411463/85920154-02293f80-b87a-11ea-867f-a752815eeb71.png)

Hachoir showing the data for one of the tracks, audio data highlighted in blue:
![hachoir](https://user-images.githubusercontent.com/6411463/85920161-13724c00-b87a-11ea-853f-21d9adad25e0.png)

#### Affected LMMS versions
Master pulled and compiled on 2020-06-27, i.e. LMMS 1.2.1.702 (Linux x86_64, Qt 5.12.4, GCC 9.2.1 20191008)
Libsndfile 1.0.28 could be relevant, too.

thmueller64:
I could not reproduce this behavior with the current master and libsndfile 1.0.28-7. Could you maybe share a project file?

softrabbit:
[issue-5548-shorter.mmpz.zip](https://github.com/LMMS/lmms/files/4859477/issue-5548-shorter.mmpz.zip)

A few more things I noticed:
- the garbage seems identical between files from the same export.
- looks like some degree of oversampling is needed for the bug to pop up, 1x seems to work fine.

Did a rebuild just in case something wasn't right, blowing away and recreating my `build` and `target` directories just to be sure before pulling and initing submodules and whatever the wiki says. Same bug still.

thmueller64:
I can reproduce the behavior now with oversampling > 1. From what I can tell, the issue is that libsamplerate sometimes generates less samples than lmms expects. Could you please try out the draft PR above and check if the problem persists?

softrabbit:
Cool! I'll check the PR tomorrow or possibly later today.

softrabbit:
Looks like closing time, the fix is merged.

PR #5552, opened 2020-07-02
Fix garbage in exported audio caused by resampling

thmueller64:
This PR attempts to fix the problem of garbage in exported audio as reported by #5548. In the case of oversampling > 1, it is possible that libsamplerate returns fewer samples than assumed. This is exposed by libsamplerate via `output_frames_gen`.


softrabbit:
Looks like this fixes the bug as intended.

PR #5332, opened 2019-12-01
add pause shortcut shift + sapce (#5301)

tjrdnjs1609:
add shortcut shift + space for pause on SongEditor.cpp
it only work on playing state.
when it paused, shift + space does not work.

Issue #5301, opened 2019-11-03
Different hotkey for pause and stop


MrcSnm:
When playing on piano roll, there is an option on the tool bar to stop the current playing and to pause it.

The difference between them is that stop will go back to the 0:00 position, while pause will maintain the current track position, what I want is to correct a bug that Stop and Pause button have the same hotkey (Spacebar), and that is boring me to death, I don't want to be going back to 0 every time I press this key.

What I recommend is having Spacebar for Pause and Shift+Spacebar for Stop, I would program this myself, but I really don't know where is the controller for that.

There is a screenshot I made showing that the hotkey are the same for these two options

![Error](https://user-images.githubusercontent.com/10136262/68080804-8c9bcb00-fde1-11e9-8be5-bf8da960213c.png)

tjrdnjs1609:
i try to solve this problem. but i couldn't find 'pause' in SongEditor.cpp .  anyone who explain about how to add or change 'pause' shotcut???

zonkmachine:
Have a look at [`void SongEditorWindow::play()`](https://github.com/LMMS/lmms/blob/a9e3e70ae3d28e7aa846904edae71863a9734a71/src/gui/editors/SongEditor.cpp#L1045:L1056)

MrcSnm:
Please remember that this problem persists on piano roll, song editor and much probably in beat editor

musikBear:
#728 orr. post

tjrdnjs1609:
#musikBear

> #728 orr. post ..and _from_ 2014...!
Thanks for the good info. but i think same shortcut is still problem. 

Does it make sense if I solve this problem?

MrcSnm:
Absolutely

> #musikBear
> 
> > #728 orr. post ..and _from_ 2014...!
> > Thanks for the good info. but i think same shortcut is still problem.
> 
> Does it make sense if I solve this problem?

Absolutely, having this bug on pause/stop really breaks the workflow

zonkmachine:
> Does it make sense if I solve this problem?

As you have volunteered to fix this and have a PR open I have assigned you to this issue.

russiankumar:
Hey, is this being worked on? The PR seems inactive and I'd be glad to work on this if I am allowed to.

DomClark:
Fixed in #5554.

PR #5554, opened 2020-07-03
Make Pause hotkey Shift+Space

russiankumar:
This PR targets issue #5301 (and similar issues). This fix is similar to that of #5332, however that PR is stagnant and I want to commit changes to all editors so pausing can be done in all editors.

I've had some issues with git commiting recently, so you may see a whole load of random commits, but the final result works so far on `SongEditor.cpp`. Will commit soon to implement the same in Piano Roll, B+B Editor, etc.

It works as expected:
- While playing, Shift+Space stops the song in its place (like the `>>` option would for playback start position)
- While paused, Shift+Space (or Space) continues from the place it is stopped at.

### DIFF
--- a/data/themes/classic/style.css
+++ b/data/themes/classic/style.css
@@ -118,6 +118,11 @@ QMenu::indicator:selected {
 	background-color: #747474;
 }
 
+positionLine {
+	qproperty-tailGradient: false;
+	qproperty-lineColor: rgba(255, 255, 255, 153);
+}
+
 PianoRoll {
 	background-color: rgb(0, 0, 0);
 	qproperty-backgroundShade: rgba( 255, 255, 255, 10 );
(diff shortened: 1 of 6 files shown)


russiankumar:
Ok, I underestimated how messed up my git has become.
image
Though the final file has the expected code, so not an issue?

russiankumar:
Discovered a very useful `togglePause()` function. Changed to use that function as it emits `playbackStateChanged()` as well.

russiankumar:
Moved the shortcut to `Editor.cpp`. Every `Editor` will now pause on Shift+Space.

---
#### Noted bug
When having the song editor in pause state, a pattern is played, the song editor is stopped. This may be intended behaviour, so please let me know if I should make changes to that as well.

russiankumar:
Uhh... could someone review this...? @PhysSong? @SecondFlight? @DomClark? Tagging because you reviewed my previous PR too.

musikBear:
👍 * 💯 Good stuff! 🥇

PR #5555, opened 2020-07-04
Fix bug where LcdSpinBoxes always use the default style

DouglasDGI:
Take a look at the change.  Turns out we forgot an argument there.

There will be no visible results of this fix to my knowledge, other than making the SF2Player spinboxes purple like they were programmed to be in the first place.


Sawuare:
GIG Player uses that same constructor, so its spinboxes will be affected. too.

https://github.com/LMMS/lmms/blob/984fd3a935cb7b85ead204dc81ff32293aa556cf/plugins/GigPlayer/GigPlayer.cpp#L955-L960
https://github.com/LMMS/lmms/blob/984fd3a935cb7b85ead204dc81ff32293aa556cf/plugins/sf2_player/sf2_player.cpp#L937-L946

PR #5559, opened 2020-07-05
Fix a bug on the "Remove unused channels" command inside the FX mixer

IanCaio:
Inside the FX mixer, there's a command to remove the channels that are currently unused. The method that processes that command (`void FxMixerView::deleteUnusedChannels()`) only checked the channels for InstrumentTracks linked to them, but ignored SampleTracks. Because of that, if there was a channel where only SampleTracks are forwarded to and we clicked on "Remove unused channels", that channel was going to be removed.

This commit fixes that bug.


Spekular:
LGTM.

Issue #5201, opened 2019-09-25
[Feature Request] Portable Mode


MrTeeXD:
Some of us are not lucky enough to have our own PC and thats where this Feature Request comes in. Am example of this Request is OBS Studio. When a .txt file is created at the base of the installation, OBS will immediately go into Portable Mode, meaning all the Settings and important things will be stored in the Installation location such as a External Harddrive which can then be used to move from one PC to another.

PhysSong:
When running LMMS from the build directory, LMMS detects `CMakeCache.txt` and use the value of `lmms_BINARY_DIR` from that file to locate the config file. It should be easy to implement the requested functionality with a similar trick.
The relevant code fragment is in the `ConfigManager` constructor.

MrTeeXD:
I wonder if Samples can also follow this.

SeleDreams:
I'll look into implementing it

Issue #5062, opened 2019-06-28
Should we add IDE project files to .gitignore?


JohannesLorenz:
Question targets developers only. Issue opened because of #4977 .

I'd rather say this should go into your global gitignore. Otherwise, we could also add `.o`, `.bak`, ...

* :heart: IDE project files should go into `.gitignore`
* :eyes: IDE project files should go into your global `.gitignore`
* :rocket: Other opinion, please specify

noahb01:
Wait, there's a global gitignore, as in "these files will be ignored for all repos on this computer"? If so, then I'd close #4977 and do that instead.

JohannesLorenz:
On the Linuxes I used, it's `~/.config/git/ignore` . Does that fit your purpose?

noahb01:
Yes that works fine, git allows you to create the file anywhere and set it with `git config --global core.excludesfile <file>`

JohannesLorenz:
We had that discussion at work today with someone who wanted to add IDE project files to some repos, so they did not want to add it into their global `ignore`.

Maybe one should use the global ignore only for files where it's absolutely clear that you would not add them into any repo (e.g. `.o` files)?

PhysSong:
I found that `.git/info/exclude` also works.

SeleDreams:
> I found that `.git/info/exclude` also works.

does it mean I should create this file in my local repo and use it like a gitignore ?

SecondFlight:
I'm not completely sure what PhysSong means there (sorry!), but this is what is meant by a global `.gitignore`:

https://stackoverflow.com/questions/7335420

@JohannesLorenz I'll close this in a few days if I don't hear otherwise. I switched my vote as I apparently now disagree with past me.

DomClark:
Changing my vote too. I wasn't aware of `.git/info/exclude`, and so was against forcing LMMS's decision about project files onto users' global settings. In fact, the [Git documentation](https://git-scm.com/docs/gitignore) recommends the use of `.git/info/exclude` exactly for this purpose:
> Patterns which are specific to a particular repository but which do not need to be shared with other related repositories (e.g., auxiliary files that live inside the repository but are specific to one user’s workflow) should go into the `$GIT_DIR/info/exclude` file.

JohannesLorenz:
I'm OK with closing this, but just to understand it:

[...]

PR #5561, opened 2020-07-06
Implemented a portable mode to lmms

SeleDreams:
I've implemented this portable mode based on this issue https://github.com/LMMS/lmms/issues/5201

this is my first contribution to the project I tried to respect the programming style while refactoring a bit.

the overall modifications are : 
- I have added an environment variable PORTABLE_MODE, when LMMS is built with this environment variable, it will default its directory to the directory where it is executed (in portable mode, the lmms directory has been renamed to lmms-workspace in order to avoid name conflict on linux between the executable and the directory)

- I have fixed the typo from the function availabeVstEmbedMethods to availableVstEmbedMethods

- I have refactored the constructor of ConfigManager to be more readable

- I have added the CLion files autogenerated when opening a cmake project to the gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,5 @@
 /build
 /target
-/.idea
-/cmake-build-debug
 .*.sw?
 .DS_Store
 *~
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 /build
 /target
+/.idea
+/cmake-build-debug
 .*.sw?
 .DS_Store
 *~
--- a/src/core/ConfigManager.cpp
+++ b/src/core/ConfigManager.cpp
@@ -66,8 +66,10 @@ ConfigManager::ConfigManager() : m_version(defaultVersion())
 	m_sf2Dir = m_workingDir + SF2_PATH;
 	m_gigDir = m_workingDir + GIG_PATH;
 	m_themeDir = defaultThemeDir();
-	if (! qgetenv("LMMS_DATA_DIR").isEmpty())
+	if (!qgetenv("LMMS_DATA_DIR").isEmpty())
+	{
 		QDir::addSearchPath("data", QString::fromLocal8Bit(qgetenv("LMMS_DATA_DIR")));
+	}
 	initDevelopmentWorkingDir();
 
 #ifdef LMMS_BUILD_WIN32
--- a/src/core/ConfigManager.cpp
+++ b/src/core/ConfigManager.cpp
@@ -51,67 +51,30 @@ static inline QString ensureTrailingSlash(const QString & s )
 ConfigManager * ConfigManager::s_instanceOfMe = NULL;
 
 
-ConfigManager::ConfigManager() :
-	m_workingDir(QStandardPaths::writableLocation(QStandardPaths::DocumentsLocation) + "/lmms/"),
-	m_dataDir("data:/"),
-	m_vstDir(m_workingDir + "vst/"),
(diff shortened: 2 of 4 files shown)


PhysSong:
First of all, thank you for your contribution! It seems like your editor is using 4 spaces for indentation, but it should be a tab according to our coding convention.

I also suggest adding a runtime check because people don't need to rebuild LMMS to make existing builds portable. For example, OBS checks for `portable_mode.txt` and runs in portable mode if it exists.

SeleDreams:
> First of all, thank you for your contribution! It seems like your editor is using 4 spaces for indentation, but it should be a tab according to our coding convention.
> 
> I also suggest adding a runtime check because people don't need to rebuild LMMS to make existing builds portable. For example, OBS checks for `portable_mode.txt` and runs in portable mode if it exists.

I just got CLion as my editor and as I didn't know it enough I didn't know the type of identation it used, I'll have to see if I can make it switch to tabs instead

about runtime check, that's something I thought about and I asked it on the discord when I started working on it but I didn't get responses so I went with my first idea, it shouldn't be that hard to implement though, just checking if the file exists instead of checking for the define

SeleDreams:
Screenshot from 2020-07-07 04-22-40
found the setting, it should make future contributions easier

SeleDreams:
I made the changes

PhysSong:
This time, you(or the IDE) changed formatting too much. This is not about conventions, it's about irrelevant changes. Maybe it's better to reset to clean `master` and redo your works.

SeleDreams:
as it's a bit late in this case I would probably more do it tomorrow (it's 5 AM in france)

SeleDreams:
I just re-did it and it now only affects the changes I made

src/core/ConfigManager.cpp
```diff
@@ -51,68 +51,31 @@ static inline QString ensureTrailingSlash(const QString & s )
...
+	{
+		initInstalledWorkingDir();
 	}
+	m_dataDir = "data:/";
+	m_vstDir = m_workingDir + "vst/";
+	m_sf2Dir = m_workingDir + SF2_PATH;
+	m_gigDir = m_workingDir + GIG_PATH;
+	m_themeDir = defaultThemeDir();
+	if (! qgetenv("LMMS_DATA_DIR").isEmpty())
+		QDir::addSearchPath("data", QString::fromLocal8Bit(qgetenv("LMMS_DATA_DIR")));
```
PhysSong, line 70: According to the [coding convention](https://github.com/LMMS/lmms/wiki/Coding-Conventions#braces), you should use braces here.

.gitignore
```diff
@@ -1,5 +1,7 @@
 /build
 /target
+/.idea
+/cmake-build-debug
```
PhysSong, line 4: Please see https://github.com/LMMS/lmms/issues/5062#issuecomment-656440516, a discussion about `.gitignore`.

src/core/ConfigManager.cpp
SeleDreams: I usually always do, but as it was already like this I kept it as it was, was I supposed to change it ?

PhysSong: I didn't know that. I'll leave it up to you.

SeleDreams:
It should be good now

russiankumar:
Is this PR complete?

SeleDreams:
Yes

PR #5563, opened 2020-07-07
Enable LMMS fullscreen and... (long title, read first line of description)

russiankumar:
## Enable LMMS to go fullscreen with hotkey Ctrl+F11, and toggle internal window's maximise state with Alt+F11

The title says exactly which features this commit will add.

#### Features Added:
- [x] LMMS can go fullscreen via <kbd>Ctrl</kbd>+<kbd>F11</kbd> (later just <kbd>F11</kbd>)
- [x] Subwindows (Song Editor, etc.) can be maximised/returned to normal size by <kbd>Shift</kbd>+<kbd>F11</kbd>
- [x] Fullscreen menu button under View menu
- [x] Remapping <kbd>F5</kbd>-<kbd>F11</kbd> to <kbd>Ctrl</kbd>+<kbd>1</kbd>-<kbd>7</kbd> to allow just <kbd>F11</kbd> for fullscreen
- [x] Custom pixmap for fullscreen (possibly)

#### Bugs fixed:
- [x] Linux: LMMS window seems to not restore maximised state after exiting fullscreen
- [x] Windows: Alt+F11 does not work (fix: change shortcut to Shift+F11)
- [x] Variable `maximise` not consistent with project spelling (maximize)

#### Progress
progress 
All done!

---

This isn't a two-day feature and will probably take a while to get past everyone's approval (especially the fourth feature which may not even be included). Either ways, in its present state, it is open for review as it is a functional feature.

I feel fullscreen mode is a sort-of useful feature for when you have just that one track that goes beyond the window that you need to see, or when you're bored of the taskbar, or... well you get the point. Also, this is contributing to the efforts to make LMMS more hotkey-friendly by adding the <kbd>Shift</kbd>+<kbd>F11</kbd> hotkey for an already existing function.

---

If there's any new bugs discovered or fixed, or features outvoted or fixed, I'll edit this comment so you can get the status at a glance.


src/gui/MainWindow.cpp
```diff
@@ -245,6 +245,9 @@ MainWindow::MainWindow() :
 
 	connect(Engine::getSong(), SIGNAL(modified()), SLOT(onSongModified()));
 	connect(Engine::getSong(), SIGNAL(projectFileNameChanged()), SLOT(onProjectFileNameChanged()));
+
+	maximised = ( isMaximized() ? true : false );
```
russiankumar, line 249: This line will be replaced with just `maximised = isMaximised()`, I don't know what got over me here xD

src/gui/MainWindow.cpp
```diff
@@ -782,6 +786,9 @@ void MainWindow::restoreWidgetState( QWidget * _w, const QDomElement & _de )
 		flags = _de.attribute( "maximized" ).toInt() ?
 				( flags | Qt::WindowMaximized ) :
 				( flags & ~Qt::WindowMaximized );
+		flags = _de.attribute( "fullscreen" ).toInt() ?
+				( flags | Qt::WindowFullScreen ) :
+				( flags & ~Qt::WindowFullScreen );
```
russiankumar, line 791: Is this necessary? Not exactly sure what `restoreWidgetState()` does.

src/gui/MainWindow.cpp
russiankumar: Fixed this up.

russiankumar:
Fullscreen can now be found in the View menu:

image

I have not been able to find the menu for internal windows, so I have no idea how to add the entry for Alt+F11 for maximise. The shortcut does work though.

Also, I'm currently using the maximise pixmap for fullscreen. Getting a new pixmap for fullscreen (if possible) would be a lot more user-friendly in my opinion, though I'm not placing bets on that happening soon.

russiankumar:
Found a bug where Alt+F11 does not work on Windows. Pending change to Shift+F11.

russiankumar:
> Possibly remapping F5-F11 to F4-F10 to allow just F11 for fullscreen

Do I cancel this? I don't think many would like this workflow change.

Spekular:
> > Possibly remapping F5-F11 to F4-F10 to allow just F11 for fullscreen
> 
> Do I cancel this? I don't think many would like this workflow change.

Is this something a lot of people use? I had to open LMMS and test to see what these buttons did. Either way, an alternative would be to remap them to Ctrl+F5-11. Then fullscreen can be an unmodified F11, but the editors aren't offset from their location (in case someone has muscle memory for that).

russiankumar:
> Is this something a lot of people use?

I don't know, but I'd rather not make assumptions. Maybe I'm the only one who doesn't use them xD.

> an alternative would be to remap them to Ctrl+F5-11. Then fullscreen can be an unmodified F11, but the editors aren't offset from their location.

This seems legit. That would open up F5 through F10 for other more often used shortcuts (if any are required, like say closing non-focused or all windows like FL has).

russiankumar:
@Spekular I've committed such that it's now Ctrl+F5-F10, and it fixed a bug I had before as well (apparently Ctrl+F11 was the shortcut for restoring to normal size in Lubuntu). I'd vouch for keeping it this way.

image

russiankumar:
If you were wondering: Ctrl+F11 doesn't minimise the window as of the latest commit, so I think there's some bug with Qt on Linux using Ctrl+F11 for fullscreen. Not sure what it is, but it's fixed.

src/gui/MainWindow.cpp
Spekular: I'm pretty sure this is used to put widgets in the correct (saved) position when loading a file. Considering that and the fact that individual widgets don't have a fullscreen attribute, I would remove these lines. Unless, of course, your testing shows that it breaks something to do so.

Spekular: In fact, I don't think fullscreen state should be restored on file load at all, so there's another argument for not saving or loading this attribute anywhere.

src/gui/MainWindow.cpp
```diff
@@ -738,6 +741,7 @@ void MainWindow::saveWidgetState( QWidget * _w, QDomElement & _de )
 	_de.setAttribute( "visible", visible );
 	_de.setAttribute( "minimized", _w->isMinimized() );
 	_de.setAttribute( "maximized", _w->isMaximized() );
+	_de.setAttribute( "fullscreen", _w->isFullScreen() );
```
Spekular, line 744: This should also be removed, according to logic in previous comments.

russiankumar: Could you explain why? I'm not 100% sure what it does but I saw maximise & minimise there so I kept this as well. Does it save the "fullscreen" argument in the project file by chance?

include/MainWindow.h
```diff
@@ -224,6 +225,8 @@ private slots:
 	ToolButton * m_metronomeToggle;
 
 	SessionState m_session;
+	
+	bool maximised;
```
Spekular, line 229: This variable is only ever assigned, never read?

russiankumar: Oh no, I left the reading out from my debug phase xD
Let me fix that real quick

russiankumar: Yep, now its being read to restore it to its previous state (maximised or normal).

russiankumar:
The macOS build seems to fail every once in a while due to some python issue for recent commits and PRs, any idea why? (and is it something we can fix?)

src/gui/MainWindow.cpp
Spekular: Essentially yes, I believe that's what it does. QDomElement is "[an] element in the DOM tree", and "Document Object Model (DOM) [...] treats an XML or HTML document as a tree structure". In other words, anything involving a QDomElement has to do with de/serialization, as far as I'm aware.

Since we have an MDI interface, individual windows inside the main LMMS window can have minimize/maximize states, which is what's saved here. Fullscreen is a property of the top level LMMS window though, so even if it were saved it shouldn't be once per widget.

Hope this isn't too terrible of an explanation, haha.

include/MainWindow.h
Spekular: Ok. Since I thought it would get removed I didn't mention this, but for consistency's sake it should probably be renamed to maximized 😅. Sorry for making a bunch of separate comments instead of a single review pass 😬.

russiankumar: > renamed to maximized

Lol sure I'll do that

Spekular:
> Also, I'm currently using the maximise pixmap for fullscreen. Getting a new pixmap for fullscreen (if possible) would be a lot more user-friendly in my opinion, though I'm not placing bets on that happening soon.

@RebeccaDeField is this something you could handle? From what I can tell, these icons should be 16x16 in order to be crisp.

src/gui/MainWindow.cpp
russiankumar: Ohh, I see. Basically I was saving the fullscreen attribute for widgets which didn't even have the fullscreen option xD

Thanks for the explanation, this will probably help in future PRs.

include/MainWindow.h
russiankumar: Did that

Spekular:
Everything LGTM, but I'll wait and see about the fullscreen icon before approving. In addition, I'd like confirmation from at least one more person that it's ok to change the editor shortcuts. Personally, I'm on board with it. Adding a modifier to the shortcut isn't that big of a change, and preliminary results from a poll (via @lmmsio on twitter) show that most users didn't even know what they did, and <10% use them more than "sometimes".

include/Editor.h
```diff
@@ -62,6 +62,8 @@ private slots:
 	/// Called by pressing shift+space. Toggles pause state.
 	void togglePause();
 
+	void toggleMaximise();
```
Spekular, line 65: Dang it, missed this one earlier. Once again, I'd prefer if this was consistent with the existing isMaximized function's name.

russiankumar:
> I'll wait and see about the fullscreen icon before approving.

Yeah, I'd rather wait for at least some kind of response before the merge. If it gets too late though we can always add the pixmap in a later PR (eg. your issue #5569 will probably create a PR for pixmap corrections, which can also add a custom pixmap for fullscreen)

include/Editor.h
russiankumar: Ughhh I'll fix it in a commit in sometime

he29-net:
I think there may be compatibility issues with some of the proposed shortcuts.

Notably, re-mapping anything to `Ctrl` + `F5-F11` will break some Linux desktops, as `Ctrl` + `Fx` is commonly used to switch virtual desktops. I currently use 6 desktops, so I would be unable to use two of these shortcuts. I doubt many people commonly use more than 10 desktops, so `Ctrl + F11` should be _relatively_ safe to use.

Another problem is with `(left)Alt + F11`, which, at least in Xfce, is already used to toggle fullscreen. This is done at the window manager level, so the application probably does not get to decide whether it wants to go fullscreen or not. So maximizing an internal window using `Alt + F11` would toggle LMMS fullscreen state as well.

RebeccaDeField:
> @RebeccaDeField is this something you could handle? From what I can tell, these icons should be 16x16 in order to be crisp.

Sure I'm on it, just wanted to check since you said "icons" do you need more than one? I only see a request for a fullscreen icon.

Spekular:
@RebeccaDeField ah, by "these icons" I meant "the icons in the file menu". Sorry for the confusing wording.

Re linux desktops inconveniently reserving a bunch of shotcuts: damn it. Are there any issues/collisions with shift + F#? That would just leave the shortcut for maximizing windows, which could perhaps be dropped, or achieved by changing the toggle hotkey to a cycle. In other words, Shift+F5 opens the song editor if it's closed, maximizes it if it's open, and closes it if it's maximized.

he29-net:
I don't know about any system-wide use for the `Shift + Fx` shortcuts, these should work fine. Unless they are used for something on another platform. I.e. does even Mac have the Fx keys? I just remember it uses a "flower" thing instead of a `Ctrl`, but Apple tends to "innovate" quite a lot, so I wouldn't be that surprised if they dropped the Fx keys. Also, on some laptops you have to press the `Fn` modifier to access `F1-12`, since by default the keys are mapped to stuff like volume or brightness. So yeah, unless it is a pure `Ctrl + letter`, having _anything_ work well across all platforms seems pretty challenging..

I'm also not a fan of remapping the show/hide shortcuts in general: you will probably toggle fullscreen with `Ctrl + F11` once, at the beginning. But if you use the `F5-F11` keys to show/hide windows, you are probably using them far more often than the fullscreen toggle. So requiring the user to combine them with `Shift` would only make their use more tedious and undesirable..

Spekular:
I'm a proponent of using the unmodified key for fullscreen because it's more consistent with other software. F11 for fullscreen is so extremely standard that if pressing F11 doesn't fullscreen a software, my assumption will be that fullscreen isn't supported at all. I think it's a bad idea to subvert that.

russiankumar:
Alright, since ctrl+F5-F11 are likely to cause issues, what would you say of ctrl+4-0? I don’t think there’s currently any such shortcut... Either that, or numpad keys 1-7 since I think most users would have a numpad (most laptops ship with numpads too). 

In the worst case we could have F12 as the shortcut if the F5-F11 mapping needs to stay. Just to remind that Ctrl+F11 has bugs on Linux which are again at the window manager level, so we need to settle for either remapping to ctrl+4-0 or to the numpad keys.

RebeccaDeField:
@Spekular The first one is the more standard icon, but the second may be clearer metaphorically. Any preference?

Screenshot from 2020-07-09 22-27-55

he29-net:
I like the idea of using the keypad; some laptops won't have it, but as I said earlier, it's probably impossible to find something that works 100 % everywhere.

Using F12 does not make sense to me; Spekular has a good point, F11 is the "industry standard" for fullscreen toggle so it should get the priority. Also, I'm already using F12 in the microtonality PR to toggle the microtuner settings window, so there would be a collision in the future anyway.

I'm not sure why F1-F4 aren't used at all. I thought it's because F1 is often used to open the manual, so it would, again, be unexpected? But I like the idea of using F1-F8 about the same as the numpad keys, and it may even be more acceptable and familiar for users and their muscle memory. :) Starting at F4 seems "misaligned" to me, and would prevent future expansion (e.g. for the microtuner window).

russiankumar:
@he29-net sounds cool then, I'll move the keys to numpad 1-7.

russiankumar:
@Spekular and @RebeccaDeField, just placing my two cents out here: since we're so dedicated to making F11 the fullscreen shortcut as it's industry standard, I guess having an industry standard icon would make sense as well. Though if we're going for a vibe of "not like the others", the second one would suit perfectly.

include/Editor.h
russiankumar: Just updated, now fixed.

Spekular:
@RebeccaDeField I lean towards the more standard icon.

Spekular:
The results of the [poll on twitter](https://twitter.com/lmmsio/status/1281159776555151362?s=19) regarding usage of the hotkeys are in, with 65 votes:
- 63% of users don't even know what the hotkeys do
- 14% know, but don't use them
- 12% use them sometimes
- 11% always use them

So 77% don't use the hotkeys, and of those that do only half use them more than "sometimes". I don't think we have too much to worry about in changing them.

RebeccaDeField:
Looks like we all lean towards the more standard one!

Icon source: [Fullscreen.svg.zip](https://github.com/LMMS/lmms/files/4904684/Fullscreen.svg.zip)

russiankumar:
Thanks for the icon! The latest commit uses it instead of the maximise icon, however I have placed the .svg in the data folder. Is that what is supposed to be done or am I supposed to convert to .png and then use it? It works fine and looks clear as .svg.

RebeccaDeField:
@russiankumar Sorry for any confusion, all icons are in .png for now. I do believe an effort should be made in another PR to see if we can convert all to .svg which will be aided by my efforts to separate and track all the original sources (almost complete :tada:  )

russiankumar:
@RebeccaDeField so do I leave it as a .svg or do I convert it for now?

RebeccaDeField:
@russiankumar I think it's best to export as png for now to stay consistent.

russiankumar:
It's now in .png format (in 16x16 resolution for max clarity). Looking nice so far:

image

Not ready to merge yet as I have to move Ctrl + F5-F11 to numpad keys. Nonetheless, thanks a lot @RebeccaDeField!

russiankumar:
No amount of Googling showed me how to detect numpad keys (and _only_ numpad keys, not normal number keys). What if we make it Ctrl+1-7? It seems okay that way... @he29-net, your call

Spekular:
You can try checking for Qt::KeypadModifier, I think this might let you detect numpad events. Otherwise I think Ctrl+1-7 is fine if those combos are unused right now.

russiankumar:
I've tried checking for it but to no avail. Besides Ctrl+1-7 can be used by people without numpads as well, so I think we can set it as that.

russiankumar:
Commited with a working build using Ctrl+1-7. Also cleared up the menu to have tab spaces before shortcuts to be more consistent with the rest of the menus:

image

This menu had the shortcuts in brackets prior to this (the only menu with this style). Also added a separator between the editors and fullscreen because I don't really think they belong to the same submenu. It looks nice anyways :)

Spekular:
Tested, working, merging.

PhysSong:
I found that the new <kbd>Ctrl</kbd>+<kbd>1</kbd>-<kbd>7</kbd> shortcuts conflicts with some of the shortcuts for changing the note length in piano-roll. I'm not sure if there are some more.

charbelnicolas:
I haven't tried the nightly 1.3 versions yet, but does this fix https://github.com/LMMS/lmms/issues/6015?

charbelnicolas:
Will this also fix LMMS not remembering its maximized window state on close/open? LMMS always opens its main window maximized no matter what.

ryuukumar:
Nope, that would have to be a separate PR/bug fix. I'm not actively developing, ~but feel free to register this as a feature request if it isn't already.~ (I've noticed that you've already registered #6015, so maybe you could try a ping on Discord in the dev channels?)

PR #5572, opened 2020-07-10
Color coding for entire track

russiankumar:
This PR extends BB track's color dialog implementation into a track setting.

I will provide a more accurate description in an edit later (as I am busy trying to fix my already messed up Git). I apologize in advance for the commit mess; I tried to solve it before but all it did was somehow break the build instead of doing anything useful. 

I had coded extra features but had lost them due to the Git errors, therefore, this PR is the earliest working build. It's not rebased, which I will have to do manually because I'm a noob and can't figure out how to do it without messing up the PR all over again.

Nonetheless, here's the features I plan to add (or added until Git messed them all up):
- Track-wide "Change Color" and "Reset Color" settings
- Custom colors in the QColorDialog

You may expect me to close this PR and add another one later if my Git skills fail me again.

Here's a screenshot:

image


russiankumar:
Most of the above commits are my older ones, you can safely ignore them as they do nothing with respect to this PR.

russiankumar:
It appears to be broken. I'll refactor and re-PR when it's fixed.

PR #5573, opened 2020-07-10
Enable track-wide color coding

russiankumar:
## What does this PR do?
This PR adds the following features:

- [x] BB pattern-like color coding, except on all tracks
- [x] Saves the color for each track in the project
- [x] Resets the color of the track to the default color
- [x] Enables custom colors in the color dialog for 48 better-looking colors
- [x] Allows picking random colors from these custom colors
- [x] Adds a band in the track grip where the track color is displayed
- [x] Changes a demo in "demos" to demonstrate color-coding (in both track and mixer)

### Updated (current state)

image

image

image

---

I believe this feature will truly lift LMMS to a whole new level, as this is something most commercial DAWs have. And it's not mandatory; without using it, LMMS still appears with the same default colors like everyone is used to.


russiankumar:
The melody pattern appears a little bit off (on the right side) compared to the rest, any idea why?

image

Spekular:
I don't think this is your PR's fault, the inner border for melody clips is drawn one pixel to the left of where it should be (at least at high zooms/track heights).

russiankumar:
@Spekular yeah I've noticed this in other builds too, shall I fix it or is it not a big deal?

russiankumar:
#### Noted bug

The background of beat+bassline tracks is also colored, this may be related to the bug where initialising colors are ignored for some reason.

image

_I made the background colorful to emphasise the bug, it doesn't actually appear like this xD_

Spekular:
> @Spekular yeah I've noticed this in other builds too, shall I fix it or is it not a big deal?

I definitely think it should be fixed, so the question is if it's in this PR or another. I think I'd prefer a separate PR. As a pure bugfix, that one could get merged rather quickly and not be tied to however long this takes to merge. It'd also make the diff for this PR cleaner. It does make more work for you with potential merge conflicts, and I prefer a fix in this PR over no fix at all, so I leave the decision to you.

russiankumar:
#### Noted bug

New TCOs do not take up the track color when a custom color has been assigned.

russiankumar:
@Spekular yeah I think another PR would be a better choice since it's (hopefully, probably) a really tiny fix. It'll take some time setting it up but I think it would save everyone time not having to wait for this PR to merge.

russiankumar:
Fixed all the bugs I found related to my changes, and a basic color coding implementation is now complete. I have also changed the styling on the sample track TCO because some colors look awkward with the old settings.

**Looking for people willing to test this PR for potentially undetected bugs.**

russiankumar:
By the latest commit, `QColorDialog`'s "Basic Colors" are overwritten to

1. look good as they are sorted
2. be relatively dark colors to not cause problems with seeing tiny notes

#### Demo

image

_These colors are obtained algorthmically using HSL values_

russiankumar:
I've now added a feature where the side of the track colors itself based on the color of the entire track. This feature is purely subjective and if people take this as an unwelcome feature I'll remove it. I just think this looks good, and I'd vouch for it being implemented in its current state.

I have pushed an edit updating a demo, which I didn't realise I did. I'll replace it with the original and use another file for my testing (or we could leave it like that and color-code all the other projects to make them look better. I got the time to do that.)

#### Demo

image

image

_Track colors were placed manually, the fading of the color as it goes down is not an implemented option_

russiankumar:
Things I still have to complete:

- [ ] Make selections also depend on color
- [ ] Completely migrate individual TCO colors to one master track color setting
- [ ] Disable the gradient on the side when track is muted (or make it a different gray color)

russiankumar:
### Noted bug

TCOs do not take up track color.

#### Steps to Reproduce

1. Use a track with a set color and TCOs
2. Reset track color
3. Select some TCOs
4. Set a track color

The TCOs show with the right color only while selected.

The build is broken; I will patch it with a commit later.

russiankumar:
Suggestion: shall I create a short track with color-coding to demo color coding in LMMS for new users? I think it would be nice.

russiankumar:
Random colors can now be picked

wujekbrezniew:
Is this work with mixer to? I mean, if i choose color for one instrument, connected fx channel in mixer is colored same?

russiankumar:
No, it doesn’t. That’s something to do in a later PR, and I’ll be working on it soon. This is a major change enough and I think mixer color coding deserves its own PR.

russiankumar:
@wujekbrezniew you may want to see #5589, that's the PR I made for mixer color coding.

russiankumar:
### Update

By the request of others, colors are now displayed in a different way, replacing the original gradient with a tiny color bar.

image

RoxasKH:
Maybe we should suggest to the user some quick default colors that fit LMMS ui well as FL Studio does, leaving obviously the option to choose the color you want.

Love the little bars, btw

Spekular:
@RoxasKH we do already, and this PR has improved them: https://github.com/LMMS/lmms/pull/5573#issuecomment-658084465

RoxasKH:
> 
> 
> @RoxasKH we do already, and this PR has improved them: [#5573 (comment)](https://github.com/LMMS/lmms/pull/5573#issuecomment-658084465)

Oh, i misread that, perfect

russiankumar:
Pushed commit implementing per-pattern colors with following behaviour/hierarchy:

- Pattern/clip stores only the color being displayed (_i.e._ does not store both the track color and local color, track color is the track's job)
- While Ctrl+dragging, when dragged to another track,
  - if the clip uses the track color, it changes to use the destination track color.
  - if it has its own color, it keeps using its own color.
- Track color options override local clip options

I was personally against this, but it's in high demand so here it is.

This is a basic implementation, @Spekular and @DomClark has suggested a better implementation which I will be figuring out on my local branch. In case I fail to, the feature would still be ready for implementation as is right now, and can be edited in a later PR post-overhaul.

claell:
> @Spekular yeah I think another PR would be a better choice since it's (hopefully, probably) a really tiny fix. It'll take some time setting it up but I think it would save everyone time not having to wait for this PR to merge.

I created #5593 to track this issue now.

russiankumar:
Thanks for the help! Guess I don't have to fix it now xD

russiankumar:
@Spekular could you glance over the code to see if I have to fix anything before I start experimenting with the vector idea? Maybe @DomClark, would you have something to say about this? Tagging because you reviewed my previous PRs.

Also regarding the demo I wanted to add, I'm probably not going to be able to make one using just stock and have it sound good (which is required for a demo). If this will be merged after #5589 (I hope it will), I would like to take any one particular demo and edit that one to use color features. I've been using @SecondFlight's track to demo this entire thing, so maybe he would grant permission to update his track? If not, I guess I'll have to make do with something I make myself.

include/Pattern.h
```diff
@@ -181,6 +183,12 @@ class PatternView : public TrackContentObjectView
 	QColor const & getMutedNoteBorderColor() const { return m_mutedNoteBorderColor; }
 	void setMutedNoteBorderColor(QColor const & color) { m_mutedNoteBorderColor = color; }
 
+	QColor color() const
+	{
+		return( m_pat->color() );
+	}
+	void setColor( QColor _new_color );
```
Spekular, line 190: Please extract these functions into TrackContentObjectView, since they're identical for each subclass.

include/Pattern.h
```diff
@@ -191,6 +199,11 @@ protected slots:
 
 	void resetName();
 	void changeName();
+	
+	void changeClipColor();
+	void useTrackColor();
+	void trackColorChanged( QColor & );
+	void trackColorReset();
```
Spekular, line 206: Extract into TrackContentObjectView.

include/Track.h
```diff
@@ -133,6 +133,43 @@ class LMMS_EXPORT TrackContentObject : public Model, public JournallingObject
...
+
+	void setColor( const QColor & c )
+	{
+		m_color = QColor( c );
+	}
+	
+	void setColorRgb( const unsigned int & c )
+	{
+		m_color.setRgb( c );
+	}
```
Spekular, line 155:
```suggestion
	unsigned int colorRgb() const
	{
		return( m_color.rgb() );
	}
	
	QColor color() const
	{
		return m_color;
	}

	void setColor( const QColor & c )
	{
		m_color = QColor( c );
	}
	
	void setColorRgb( const unsigned int & c )
	{
		m_color.setRgb( c );
	}
```

More consistent naming?

include/Track.h
```diff
@@ -133,6 +133,43 @@ class LMMS_EXPORT TrackContentObject : public Model, public JournallingObject
...
+	{
+		return m_useStyleColor;
+	}
+	
+	bool useCustomClipColor()
+	{
+		return m_useCustomClipColor;
+	}
+	
+	void setUseCustomClipColor( bool b );
```
Spekular, line 172:
```suggestion
	void useStyleColor( bool b )
	{
		m_useStyleColor = b;
	}
	
	bool usesStyleColor()
	{
		return m_useStyleColor;
	}
	
	void useCustomClipColor( bool b );
	
	bool usesCustomClipColor()
	{
		return m_useCustomClipColor;
	}

```

I would prefer this, or just "set/get" prefixes on everything, `use__` doesn't sound like a getter to me.

include/Track.h
russiankumar: Sure! Was a copy paste from BB color implementation, guess that needs a change.

include/Track.h
russiankumar: IMO setX and X are more compliant with the rest of the code, but I suppose I could change these if you wish.

Spekular: Dropping the "get" from a getter works a lot better when the the variable name is clearly a property of some sort (ex. children() / height() / active()).

useStyleColor() sounds like it would set m_useStyleColor to true, or take a boolean argument and set m_useStyleColor to it.

If you were to rename the variable to m_hasCustomColor, you suddenly get a more intuitive getter name, hasCustomColor(). I think this might be the best solution.

(A quick categorization of my opinion on getter names (that hopefully doesn't have huge flaws):
- Good: Nouns, adjectives, third person verbs
- Bad: First + second person verbs
)

include/Pattern.h
russiankumar: Done.

include/Track.h
russiankumar: I changed it to the suggested naming scheme, but renaming to m_hasCustomColor is something I shall look over (m_useCustomClipColor may be confused with this, and then there's all the logic I'll have to rewrite).

include/Track.h
russiankumar: Done.

include/Pattern.h
russiankumar: Extracted.

russiankumar:
Alright, #5589 is merged, which means I now have to rework a bunch of this code. I'm shifting active development efforts back to this PR, expect changes soon.

include/Track.h
```diff
@@ -294,6 +294,11 @@ class TrackContentObjectView : public selectableObject, public ModelView
 	void setBBPatternBackground( const QColor & c );
 	void setGradient( const bool & b );
 	void setMouseHotspotHand(const QSize & s);
+	
+	QColor color() const
+	{
+		return( m_tco->colorRgb() );
```
Spekular, line 300: Should this be:
```suggestion
		return( m_tco->color() );
```
? Based on the name and return type.

src/core/AutomationPattern.cpp
```diff
@@ -597,11 +600,14 @@ void AutomationPattern::loadSettings( const QDomElement & _this )
...
-		setUseCustomClipColor( _this.attribute( "clipcolor" ).toInt() );
-		setColor( _this.attribute( "color" ).toUInt() );
+	if( _this.hasAttribute( "color" ) )
+	{
+		useStyleColor( false );
+		useCustomClipColor( _this.attribute( "isclipcolor" ).toInt() );
+		
+		QColor loadedColor;
+		loadedColor.setNamedColor( _this.attribute( "color" ) );
+		setColor( loadedColor );
```
Spekular, line 610: You can use [this constructor](https://doc.qt.io/qt-5/qcolor.html#QColor-5) directly.
```suggestion
		setColor( _this.attribute( "color" ) );
```

src/core/Track.cpp
```diff
@@ -329,7 +329,7 @@ TrackContentObjectView::TrackContentObjectView( TrackContentObject * tco,
 	m_gradient( true ),
 	m_mouseHotspotHand( 0, 0 ),
 	m_cursorSetYet( false ),
-	m_usesCustomSelectedColor( ! m_tco->useStyleColor() | m_tco->useCustomClipColor() ),
+	m_usesCustomSelectedColor( ! m_tco->usesStyleColor() | m_tco->usesCustomClipColor() ),
```
Spekular, line 332: Is there any reason you're not using a short-circuiting or `||`?

src/core/Track.cpp
```diff
@@ -614,6 +615,63 @@ void TrackContentObjectView::updatePosition()
...
+	{
+		m_tco->useStyleColor( true );
+		Engine::getSong()->setModified();
+		update();
+	}
+}
+
+
+
+void TrackContentObjectView::setColor( QColor & new_color )
```
Spekular, line 662: Doesn't this (in combination with the `trackColorChanged` signal-slot connection above) mean that custom clip colors will all be cleared when a track color is changed?

src/core/Track.cpp
russiankumar: Oh, I think I intended `||` but typed just one `|` somehow. It works as expected, but I'll change it.

src/core/AutomationPattern.cpp
russiankumar: Oh, didn't know it exists. Will update

include/Track.h
russiankumar: Ugg... I probably didn't notice this from when I was renaming stuff... thanks for pointing it out, I'll update it

src/core/Track.cpp
russiankumar: The clip color boolean won't change (I think I have to set that to `false` in that function, would be a nice idea) but since style color is `true` after the color is reset, then the painter ignores the clip color boolean and moves straight to rendering in default color.

include/Track.h
Veratil: Just to note, `return` doesn't need parenthesis around it.

include/Pattern.h
```diff
@@ -110,7 +110,9 @@ class LMMS_EXPORT Pattern : public TrackContentObject
 
 
 	TrackContentObjectView * createView( TrackView * _tv ) override;
-
+	
+	void setBGColor( QColor & c );
```
Spekular, line 114: Are these supposed to be here? I can't see these methods defined anywhere in Pattern.cpp, + they should be inherited from Track.h/Track.cpp.

src/core/Track.cpp
Spekular: I would expect custom colored clips to preserve their color when the track color is changed. Resetting all clips to use the track color should be a separate action to changing track color, IMO. So you'd have three color related track actions, with the third one being the only one to clear per-clip colors:

- Change track color
- Reset track color
- Reset all clips to track color / Clear per-clip colors

It might be worth asking other people what behavior they'd prefer, but here's the reasoning behind my suggestion:

<table>
<thead>
  <tr>
    <th rowspan="2">Desired behavior:<br>Change track color and...</th>
    <th colspan="2">Actions required if per-clip colors...</th>
  </tr>
  <tr>
    <td>...auto reset</td>
    <td>...don't reset</td>
  </tr>
</thead>
<tbody>
  <tr>
    <td>Reset per-clip colors</td>
    <td>1</td>
    <td>2</td>
  </tr>
  <tr>
    <td>Preserve per-clip colors</td>
    <td>1 + n</td>
    <td>1</td>
  </tr>
</tbody>
</table>

Where n is the number of clips with custom colors.

Spekular:
All the changes since my last review look good to me, and I didn't notice any outstanding issues when going over the full diff, code wise. As for behavior, I would prefer if per-clip colors weren't reset when track color changes, but it's up for debate if that's better or not.

Aside from that, I think it'd be a good idea to add comments above the new functions describing what they do (and perhaps some for the functions moved from BBTrack, too). Stuff like explaining that `styleColor` refers to the default color for a specific track type, for example (unless I've misunderstood).

src/core/Track.cpp
russiankumar: > Resetting all clips to use the track color should be a separate action to changing track color, IMO.

IMHO, that would make the menu a little bloated. I'll implement it, and see if it is okay.

Oh, and that would make four options, since random color is an option too.

include/Pattern.h
russiankumar: Probably not, I think I just forgot to move them to Track.h

russiankumar:
> I would prefer if per-clip colors weren't reset when track color changes

I assume with this you also mean changing the track color also doesn't affect clips? If we're going to add a "make clips follow track color" option, then I can make it so track color changes don't at all affect custom clips until the follow track color option is used.

> I think it'd be a good idea to add comments above the new functions describing what they do

I tried to make the naming as self-explanatory as I could (if there's something I misnamed please do point it out), but yeah, I shall add comments to accompany it as well.

src/core/Track.cpp
Veratil: > IMHO, that would make the menu a little bloated. I'll implement it, and see if it is okay.

One extra menu item shouldn't be that bad right?

russiankumar: I added a separator in the menu and it seems okay. I guess I'll use it like that then. I'll post a screenshot of the menu when I commit.

russiankumar:
@Veratil, @Spekular,

By the latest commit, the menu now looks like this:

image

The clip colors' behavior has been changed too to be as so:
- Clip colors are given preference when displayed
- If a clip has a custom color, when moved to a different-colored track, it retains that color. When the clip is following the color of the track, it takes up the color of the destination track
- Track color options (change color, reset etc) don't affect clip colors
- All clip colors in a track can be disabled by the "Use track color in clips" option in the track menu

--------------

Also I would like to request to make more distinguishable color icons for setting, resetting and etc, because IMO using just one icon is a little outdated considering colors are getting an overhaul. Besides the same icon slammed onto four options seems kinda monotonous. This is just my opinion, though.

src/core/Track.cpp
russiankumar: @Spekular I've changed the behavior in the latest commit as per your review. Is the new behavior what you had in mind?

src/tracks/SampleTrack.cpp
```diff
@@ -309,11 +309,8 @@ void SampleTCO::loadSettings( const QDomElement & _this )
 	if( _this.hasAttribute( "color" ) )
 	{
 		useStyleColor( false );
-		useCustomClipColor( _this.attribute( "isclipcolor" ).toInt() );
-		
-		QColor loadedColor;
-		loadedColor.setNamedColor( _this.attribute( "color" ) );
-		setColor( loadedColor );
+		useCustomClipColor( _this.attribute( "isclipcolor" ) == "1" ? true : false );
```
Spekular, line 312: `condition ? true : false` is equivalent to `condition`. Same applies to three other instances of the same line (in AutomationPattern.cpp, Pattern.cpp, and BBTrack.cpp)

```suggestion
		useCustomClipColor( _this.attribute( "isclipcolor" ) == "1");
```

src/core/Track.cpp
Spekular: Looks good, based on the code. I would change the name of the action to something like "Reset clip colors" though as I find it a bit unclear what "Use track color in clips" would do (ignoring the fact that I've looked at the implementation, haha).

Will do some testing now to make sure I haven't misunderstood how the PR works in any ways and make sure things seem to work right.

Spekular:
### Testing results: ###
✔️ Set color, random color, and reset color all work as expected. None of them override nor clear custom clip colors.
✔️ Custom clip colors can be cleared for an entire track or individually.
✔️ Cloning a track preserves all colors.
✔️ Clips with no custom color adopt their new track's color when copied to a different track.
➖  Duplicating a colored clip onto the same track...
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;✔️ ...preserves the clip color if the track is colored.
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;❌ ...loses the clip color if the track isn't colored.
❌ Custom clip colors are lost when copying them to a different track.

russiankumar:
@Spekular, if you say that the improvements are beyond the scope of this PR, maybe file issues for them after this is merged so anyone could take up on those?

About the testing, I'm patching up stuff, so expect a commit pretty soon.

russiankumar:
Commited. Should now fix these two bugs:

> ❌ Duplicating a colored clip onto the same track loses the clip color if the track isn't colored.
> ❌ Custom clip colors are lost when copying them to a different track.

src/core/Track.cpp
russiankumar: Great! Is "Disable clip colors" a better alternative?

Spekular: It doesn't prevent you from setting custom clip colors in the future, so disable isn't accurate IMO ("(of an action or circumstance) prevent or discourage (someone) from doing something").

This is similar to the issue I have with "use track colors in clips", that description sounds like it enables the behavior where clips inhereit the track's color. Obviously it doesn't since they do that by default, it just removes/resets/clears/unsets the per-clip color of any clip on the track that has one.

Spekular:
I can confirm both bugs I found in my previous testing are now fixed. IMO, this is ready to merge, so I'll do that tomorrow unless someone objects or merges before then.

russiankumar:
IMO it's still early to merge. I was/am waiting for a reply from @SecondFlight on whether they would allow me to modify their demo to have color-coding, and I was also thinking of new icons as described in [this comment](https://github.com/LMMS/lmms/pull/5573#issuecomment-674477177).

Also, no offence, but it would be great to have two people review just to make sure that everything's OK. @DomClark, you've reviewed the mixer colors, so if you have the time, maybe you could review this too?

IanCaio:
I think new icons can always be added through a new PR. It's probably going to be done by one of the artist collaborators so it's even fair that they get to submit the PR of the icon changes themselves.

As for reviewing, I can try to review the PR later, but I'm quite far behind on it and would take me a while to get familiar with everything. I'm also not sure my approval really counts for merging.

Spekular:
Good point, I mistakenly thought Dom had already done a review of this somewhat recently (probably got it mixed up with the mixer color PR.

As for the icons, I think it'd be fine to leave them as is and add new ones in a later PR. That way, people can use the feature in the meanwhile, so there's less stress on the person designing the new icons.

SecondFlight:
@russiankumar Sorry, I missed this earlier. Feel free to update my track and reorganize it, as long as you don't change the song itself. I'm not a very organized producer so don't stress the details too much. The screenshot you posted looks really pretty :)

russiankumar:
Two of you have pointed out that the icons can come in a later PR, so sure, that won't be a concern.

SecondFlight, thanks for the permission! I'll make sure not to change any song elements.

src/core/Track.cpp
russiankumar: Changed the option name based on what we discussed on Discord. Resolving this.

src/core/Track.cpp
```diff
@@ -2190,7 +2371,18 @@ void TrackOperationsWidget::mousePressEvent( QMouseEvent * me )
 void TrackOperationsWidget::paintEvent( QPaintEvent * pe )
 {
 	QPainter p( this );
-	p.fillRect( rect(), palette().brush(QPalette::Background) );
+	
+	if( hasColor && ! m_trackView->getTrack()->getMutedModel()->value() ) 
+	{
+		QRect coloredRect( 0, 0, 10, m_trackView->getTrack()->getHeight() );
+		
+		p.fillRect( rect(), palette().brush(QPalette::Background) );
```
DomClark, line 2379: This line can be factored out above the if statement, then the else block can be removed.

src/core/Track.cpp
```diff
@@ -2200,6 +2392,12 @@ void TrackOperationsWidget::paintEvent( QPaintEvent * pe )
 	{
 		p.drawPixmap( 2, 2, embed::getIconPixmap("track_op_grip_c"));
 	}
+		
+	if( colorBarNeedsUpdate )
+	{
+		colorBarNeedsUpdate = false;
+		update();
+	}
```
DomClark, line 2400: It's not clear to me what this is for. The colour bar is already painted earlier in `paintEvent`; this will just cause the `TrackOperationsWidget` to be painted a second time.

include/Track.h
```diff
@@ -498,11 +582,17 @@ private slots:
 	PixmapButton * m_muteBtn;
 	PixmapButton * m_soloBtn;
 
+	QColor m_backgroundColor;
+	bool hasColor;
```
DomClark, line 586: Similarly to before, these just seem to duplicate the corresponding values from `Track`. Instead of having both, and trying to keep them in sync, consider just using the ones from `Track` instead.

src/core/Track.cpp
```diff
@@ -609,6 +746,21 @@ void TrackContentObjectView::dropEvent( QDropEvent * de )
 	QDomElement tcos = dataFile.content().firstChildElement( "tcos" );
 	m_tco->restoreState( tcos.firstChildElement().firstChildElement() );
 	m_tco->movePosition( pos );
+	auto old_tco = dynamic_cast<TrackContentObjectView *>( qwSource )->m_tco;
```
DomClark, line 749: Is it always the case that `qwSource` is an instance of `TrackContentObjectView`? If so, use `static_cast` instead. If not, ensure that `dynamic_cast` doesn't result in `nullptr` before accessing the result. (This should always be considered when using `dynamic_cast` with pointer types.)

src/gui/AutomationPatternView.cpp
```diff
@@ -218,6 +219,10 @@ void AutomationPatternView::constructContextMenu( QMenu * _cm )
 	}
 
 	_cm->addSeparator();
+	_cm->addAction( embed::getIconPixmap( "colorize" ),
+			tr( "Set clip color" ), this, SLOT( changeClipColor() ) );
+	_cm->addAction( embed::getIconPixmap( "colorize" ),
+			tr( "Use track color" ), this, SLOT( useTrackColor() ) );
```
DomClark, line 225: These actions are repeated in each different clip view. It might be a good idea to move them to `TrackContentObjectView::constructContextMenu`, and then call the base `TrackContentObjectView::constructContextMenu` from each overridden version.

src/core/Track.cpp
```diff
@@ -386,7 +466,7 @@ QColor TrackContentObjectView::mutedBackgroundColor() const
 { return m_mutedBackgroundColor; }
 
 QColor TrackContentObjectView::selectedColor() const
-{ return m_selectedColor; }
+{ return ( m_usesCustomSelectedColor ? m_customSelectedColor : m_selectedColor ); }
```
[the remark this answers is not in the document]
russiankumar: I agree that it is a function to retrieve the `selectedColor` property, however I found it to be more convenient to check whether there is a custom selection color (i.e. the TCO has a custom color) here, in this function, rather than have to implement logic for checking whether the track has a color every time `selectedColor()` is called (implement logic once vs. atleast 4 times). Using another function to get the custom selection color leaves space for some possible bugs, and still has the problem that there now has to be logic every time `selectedColor()` is called.

I'm against changing this back to original state, and vouch for allowing an exception for this, but if you think that there's no point or there's a better way, please let me know.

src/core/Track.cpp
russiankumar: There is a bug where the `TrackOPWidget` has to be painted twice, as it doesn't update with the color changes the first time it's painted. This results in the track displaying a different color than it actually is, and causing an update (eg. moving the track) changes it to the real color. I don't know what causes this, but repainting it fixes the issue, so I stuck with it.

include/Track.h
```diff
@@ -348,6 +411,8 @@ protected slots:
 	QSize m_mouseHotspotHand; // QSize must be used because QPoint isn't supported by property system
 	bool m_cursorSetYet;
 
+	bool m_usesCustomSelectedColor;
```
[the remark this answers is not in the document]
russiankumar: I removed m_customSelectedColor. Forgot to remove the bool, will commit to update that too.

src/core/Track.cpp
russiankumar: A dynamic cast is used some five or six lines above to check whether it's pasting the same TCO into itself. I'm going to stick to the dynamic cast to stay safe, but I'll add a check for a `nullptr`.

src/gui/AutomationPatternView.cpp
russiankumar: I've migrated these to `TCOView`'s defintion, however I had to leave it in `SampleTCO` because it has its own construct menu definition (which overrides `TCOView`'s).

src/core/Track.cpp
DomClark: I tested this locally and it seems sufficient to call `update()` instead in the places where you set `colorBarNeedsUpdate` to true. Then the `colorBarNeedsUpdate` variable can go.

When the menu closes, it triggers a repaint of the `TrackOperationsWidget`, but only the region where the cog button is. You're asking for a full repaint when you get this partial repaint event, but you can get the same effect by asking for a full repaint in the first place.

src/core/Track.cpp
DomClark: Sorry, I didn't explain the problem very well. `dynamic_cast` checks at runtime whether its argument really is an instance of the type you asked it to cast to. If it isn't, and the target type is a pointer type, the result is `nullptr`. (If it isn't, and the target type is a reference type, it throws `std::bad_cast`.)

There are two possibilities:
* The argument is always an instance of the target type. In this case, you don't need the extra overhead of `dynamic_cast`'s runtime checks, so use `static_cast` instead.
* The argument isn't always an instance of the target type. In this case, `dynamic_cast` may result in `nullptr`, even if its parameter is never null. You should check whether the result is `nullptr` before you try to dereference it (in this case, access `->m_tco`). It doesn't make sense to ask `dynamic_cast` to check whether the cast is successful or not, only to ignore it and dereference the result anyway.

The `dynamic_cast` above is fine, because the result is never dereferenced. It's just compared to `this`, so if the result is `nullptr`, the comparison fails, which is the intended behaviour.

I would write something like
```cpp
const auto tcovSource = dynamic_cast<TrackContentObjectView *>(qwSource);
if (tcovSource && tcovSource->m_tco->usesCustomClipColor())
{
    // ...
```

You could declare `tcovSource` a little further up and reuse it in place of both `dynamic_cast`s, which would be a little less repetitive and a little more efficient.

src/core/Track.cpp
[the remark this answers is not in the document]
DomClark: My concern is that this leads to unintuitive behaviour for this function, and possibly bugs as a consequence. Consider a few cases:
```cpp
TrackContentObjectView *tcoView = getMeATcoView();
tcoView->setSelectedColor(QColor{11, 22, 33});
tcoView->selectedColor(); // <- != QColor(11, 22, 33) - where did my colour go!?
```
```cpp
TrackContentObjectView *tcoView = getMeATcoView();
tcoView->setSelectedColor(tcoView->selectedColor().darker(120)); // Dim the selected colour a bit
// Later on, user removes custom colour from clip
// Clip now has a broken selected colour, based on an old custom colour
```
What used to be a public accessor function - and is still explicitly declared to be a public accessor function in a `Q_PROPERTY` in `Track.h` - has been repurposed as an internal utility function. This can lead to bugs when this function is used by someone who thinks it will behave sanely, and isn't aware of its actual purpose. Furthermore, given that this is a property getter, one of those users is Qt, and we don't have control over that code. I would rather you kept the property getter as a property getter, and if you need a utility function to get a colour to paint with, add a new one.

russiankumar:
Sorry for my inactivity, was caught up in some stuff. I'll be addressing the changes requested in Dom's review soon, I pretty much agree with all of them by now. I'll push some commits within the next few days with the changes.

src/core/Track.cpp
russiankumar: I've now done what you've shown in the code. I decided to leave the other one as is as I think it's a check that has to be done, and the second is an initialisation, and they are also handled differently on true/false scenarios. You can resolve this if you feel this is satisfactory.

src/core/Track.cpp
russiankumar: Yep, can confirm that is the case.

I'll be committing soon to remove this variable.

russiankumar:
MacOS builds are failing again, or is it something exclusive to me?

src/core/Track.cpp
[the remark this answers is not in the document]
russiankumar: Alright, so I rewrote logic in all the `paintEvent`s to make them darken the TCO color appropriately while selected. Due to that, I've got some pros and cons.

Pros:
- Your concern is solved, and the `selectedColor()` function does as expected.
- I've noticed that there's no point in storing a custom selection color anyways, as it is always equal to `color.darker(200)`.

Cons:
- There's a lot more logic involved in the `paintEvent`s, turning what's essentially 4-5 lines of code into ~20 just in the name of readability.
- There's a lot of magic numbers (i.e. constants) which aren't synced across `paintEvent`s, thus if a dev is unaware of this, selected colors and muted colors will be inconsistent across different types of TCOs.

And as a side note: I also changed code such that TCOs when muted also get colored in a darker variant of their assigned custom color (non-colored TCOs work as always).

image
_Muted tracks_

image
_Unmuted tracks_

I think I should post this as a comment of its own xD

russiankumar:
By the latest commit, TCOs when muted get colored in a darker variant of their assigned custom color (non-colored TCOs work as normal).

image
_Muted tracks_

image
_Unmuted tracks_

Spekular:
> By the latest commit, TCOs when muted get colored in a darker variant of their assigned custom color (non-colored TCOs work as normal).

I think you should either 
- Revert this
- Apply it to "non-colored" clips as well. Rather than not having a color, I would think of them as defaulting to a color when none is specified. I think it'd look awkward if only some clips turned gray when muted

I prefer the first option, because I find it a bit hard to see what's muted in the example screenshot. If you go with the second one, I would also desaturate the muted clips. Saturated dark colors are very finicky IME and don't look great in this situation.

src/core/Track.cpp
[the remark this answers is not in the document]
Spekular: In response to the downsides, I would quote Dom from the comment immediately above yours:

> if you need a utility function to get a colour to paint with, add a new one.

This should solve the issues mentioned.

russiankumar:
I'd rather go with the second option, because I've been using this PR for the past month or so for production, and I noticed that it's a bit weird not having the colors when they're muted. I agree they look weird as they are right now, I was going to code for improving how they look soon, so once that's done if you still don't like it, I'll remove the feature.

qnebra:
Desaturate muted tracks, and it will be a lot clearer.

russiankumar:
@Spekular muted colors now look like this:

image
image

If this is okay, then I'll keep it as is (unless you have some suggestions). If you still are against this, I'll revert the changes.

Tagging @qnebra too, please let me know if this is satisfactory.

Spekular:
That's a lot clearer 👍

src/core/Track.cpp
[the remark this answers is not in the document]
russiankumar: Yeah, and I think since I changed the muted color logic too, I should add a function for that as well in the superclass. I'll be doing that soon.

Veratil:
@russiankumar can you try darkening the muted color by 10-20%? When I glanced through the tracks I didn't even see the unmuted green one in the center at first.

russiankumar:
@Veratil I'm already darkening them by `.darker(200)`.

Here's what 250 looks like:

image

Is this alright?

Veratil:
Much better in my eyes! The unmuted green track pops out a lot more now.

russiankumar:
Great, updated!

DomClark (commented):
There's one more piece of state duplication here, whose removal I think would simplify things further. Currently, when a clip doesn't have a custom colour, but its track does, the track colour is duplicated into the clip. This means that the clip colour may have to be updated not just when the clip colour is changed, but when:
* The track colour is changed
* The track colour is reset
* The clip colour is reset
* The clip is copied to a new track (both drag/drop and copy/paste)

Also, it means the clip colour is saved to the project file even when it's just derived from the track colour, and the clip has to keep track of where its colour comes from.

A lot of the new code in `Track.cpp` appears to be there to keep these in sync; if the clip is only responsible for its own colour, and not that of the track too, it would probably clean things up a bit.

This may slightly further complicate the UI code for choosing what colour to paint with (now's probably the time to move it into its own function) but it shouldn't be too bad - something like this (simplified a bit, of course):
```cpp
auto paintColor =
    tco->hasColor()
        ? tco->color()
    : track->hasColor()
        ? track->color()
        : defaultColor;
```

russiankumar:
For some reason I didn't get a notification for the review you left... so I'm sorry for a late response and my inactivity.

I'll be looking into improving this code, to just have the clip take the track color if `useStyleColor` is unset and `usesCustomColor` is unset too. That'll probably take some more logic as you pointed out, but I think it's doable.

I see I'm also getting merge conflicts, so I think I should finish this up quickly before this gets lost in the countless other stale PRs.

russiankumar:
So far the beat+bassline window hasn't had it's share of coloring, so I was thinking of ways to do that. One way may be similar to how I did it in [here](https://github.com/LMMS/lmms/pull/5573/#issuecomment-657189166) (albeit unintentionally, but I can prune and prod it to get it to work properly).

Another option may be this idea I was prodding around with:

image

It's already in the latest commit, although it's commented out so that it doesn't affect the present commit.

The third approach would be just to not have colors, although that kind of defeats the purpose of this PR.

src/core/Track.cpp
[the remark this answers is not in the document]
russiankumar: I've added a `getColorForDisplay()` function by the latest commit, if that is satisfactory, you may resolve this comment.

russiankumar:
@DomClark,

I've changed the code around to do as you said. Most notably, the function which is responsible for getting which color to paint with now uses the track color when there's no clip color, and while saving colors, clips which use the track color only save one attribute - `usescustomcolor = false`. If it uses the custom color it also stores the color itself.

I believe much of the logic had been unchanged, which is why I'm worried that I missed changing the logic somewhere. If there are any more efficient ways to do what I'm doing, or any logic that I missed out, please do point it out.

DomClark:
What I was thinking is that now that clips don't store the track colour, they don't need both `m_useStyleColor` and `m_useCustomClipColor`. `m_useStyleColor` is equivalent to `!(m_useCustomClipColor || getTrack()->useColor())`, and the clip doesn't need to store information about the track any more. This is the approach I'm trying to go for:
* Clips and tracks both have a single member `std::optional<QColor> m_color;`. Since we're still using C++11, this instead has to be emulated with two members each: `QColor m_color; bool m_hasColor;`.
* These function completely independently - clips don't need to store whether their track has a colour set or not. The only time clip colours need to be updated from track code is when the "clear clip colours" menu item is clicked; this only updates `TrackContentObject::m_hasColor` to `false` - it doesn't store any information about the track's colour in the clip.
* The actual colour to paint a clip can be determined as follows: If the clip has a custom colour, use that. Otherwise, if the track has a custom colour, use that. Otherwise, use the default colour.

I think this should significantly reduce the interaction between the clip and track code. Sorry for any confusion - I hope this is clearer now.

russiankumar:
So I assume by what you said, more or less the only major change I have to make is to remove `m_useStyleColor`, right? In that case, I'll get working on it today and try to push a commit soon.

russiankumar:
I've gotten rid of `m_useStyleColor`, but I've made a new function `hasColor()` for convenience (typing out `hasColor()` is a lot easier than `!( m_useCustomClipColor || getTrack()->useColor() )`).

The paintEvent functions already don't use `m_useStyleColor` iirc, so that made removing it easier.

I'd love it if someone bug tested it, with the new painting logic.

Also any opinions on the B&B coloring proposal?

Spekular:
I would skip the BB coloring. I don't think it's as useful as track coloring is, and it looks worse right now. I have ideas on how it could be improved visually, but I think they'd require significant changes and it's not worth it IMHO.

As for testing, I'll try doing my tests from earlier again!

RoxasKH:
For B&B I'd just leave only the little colored bar on the left at the moment.

russiankumar:
Alright, I'll leave BB coloring for someone else to tackle if they want to. And thanks for offering to test, @Spekular!

Spekular:
### Testing results:
*As of commit "Remove m_useStyleColor" (a601b50)*
✔️ Set color, random color, and reset color all work as expected. None of them override nor clear custom clip colors.
✔️ Custom clip colors can be cleared for an entire track or individually.
✔️ Cloning a track preserves all colors.
✔️ Clips copied onto a different track...
      ✔️ ...preserve their custom color if they have one.
      ✔️ ...adopt the new track's color otherwise.
✔️ Duplicating a colored clip onto the same track...
      ✔️ ...preserves the clip color if the track is colored.
      ✔️ ...preserves the clip color if the track isn't colored.

russiankumar:
Alright, removed the BB code.

All the tests passed (thanks for testing, again), so I assume that for the most part, if @DomClark gives approval, the code part of this is done!

DomClark (commented):
This is definitely getting there now. Now that the other changes have been made, quite a bit of code can be removed or simplified, which I've mostly pointed out in this review. Hopefully it'll become clearer where I was coming from with the earlier suggestions. Sorry for dragging this on further.

src/core/Track.cpp
```diff
@@ -247,7 +260,32 @@ void TrackContentObject::setStartTimeOffset( const MidiTime &startTimeOffset )
...
+	if( ! m_useCustomClipColor )
+	{
+		emit trackColorChanged();
+	}
+}
+
+void TrackContentObject::resetColor()
+{
+	if( ! m_useCustomClipColor ) emit trackColorReset();
+}
```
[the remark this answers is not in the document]
russiankumar: I've kept updateColor() and removed resetColor(). Is that alright?

src/core/AutomationPattern.cpp
```diff
@@ -538,6 +538,19 @@ void AutomationPattern::saveSettings( QDomDocument & _doc, QDomElement & _this )
...
+		if( usesCustomClipColor() )
+		{
+			_this.setAttribute( "color", color().name() );
+			_this.setAttribute( "usesclipcolor", true );
+		}
+		else
+		{
+			_this.setAttribute( "usesclipcolor", false );
+		}
+	}
```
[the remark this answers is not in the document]
russiankumar: Right, implemented this.

src/core/AutomationPattern.cpp
```diff
@@ -538,6 +538,11 @@ void AutomationPattern::saveSettings( QDomDocument & _doc, QDomElement & _this )
 	_this.setAttribute( "prog", QString::number( progressionType() ) );
 	_this.setAttribute( "tens", QString::number( getTension() ) );
 	_this.setAttribute( "mute", QString::number( isMuted() ) );
+	
+	if( hasColor() && usesCustomClipColor() )
```
DomClark, line 542: `usesCustomClipColor()` implies `hasColor()`, so this can be simplified to just `usesCustomClipColor()`.

include/Track.h
```diff
@@ -165,6 +186,8 @@ public slots:
 	void lengthChanged();
 	void positionChanged();
 	void destroyedTCO();
+	void trackColorChanged();
+	void trackColorReset();
```
DomClark, line 190: The `trackColorReset` signal is never emitted - remove it?

include/Track.h
```diff
@@ -245,6 +271,11 @@ class TrackContentObjectView : public selectableObject, public ModelView
 	void setBBPatternBackground( const QColor & c );
 	void setGradient( const bool & b );
 	void setMouseHotspotHand(const QSize & s);
+	
+	QColor color() const
+	{
+		return m_tco->color();
+	};
```
DomClark, line 278: It looks like this function isn't used - remove it?

DomClark (commented):
Okay, found a few more things - a bit more code that looks to be unused, and one thing that can be simplified slightly. Also, GitHub seems to have hidden a couple of comments from my last review behind a "see more" button, so if you could look for those, that would be great.

I think that's just about everything now - thank you for putting up with my nitpicking and I look forward to seeing this merged soon!

src/core/Track.cpp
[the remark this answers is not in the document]
DomClark: Looks good to me. 👍

include/Track.h
russiankumar: I got a build error when I removed it, with a function in Track.cpp is using it, so keep it around?

src/core/AutomationPattern.cpp
russiankumar: Yepp that was a blunder.... fixed it xD

include/Track.h
russiankumar: It was linked to a whole bunch of functions, but those functions were made redundant by changing how colors were displayed, so I removed the links too.

russiankumar: Also tracked down one more function - `trackColorReset()` (a different function in another class) - which has been made redundant, got rid of it as well.

src/core/Track.cpp
```diff
@@ -660,6 +759,9 @@ DataFile TrackContentObjectView::createTCODataFiles(
 		tcoElement.setAttribute( "trackIndex", trackIndex );
 		tcoElement.setAttribute( "trackType", tcoTrack->type() );
 		tcoElement.setAttribute( "trackName", tcoTrack->name() );
+		tcoElement.setAttribute( "color", ( *it )->m_tco->color().name() );
+		tcoElement.setAttribute( "styleColor", ! ( *it )->m_tco->hasColor() );
+		tcoElement.setAttribute( "clipColor", ( *it )->m_tco->usesCustomClipColor() );
```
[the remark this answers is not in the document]
russiankumar: While testing, I found out that removing this breaks the copying of clip colors for some reason. I've removed the other two, but I had to leave this because the color copying wouldn't work otherwise.

russiankumar:
Thanks for all of your efforts @DomClark! I'll be the first to admit - I think a lot of the code that was just leftover from me trying to set up wacky links between colors should have been cleaned up by me, because a lot of your review was just that. I hope that the code part is more or less done. With the final few issues left to sort out, I think this will be good to go in, worst case, a few weeks, given that I plan to comment on the code and update the demo too.

src/core/Track.cpp
```diff
@@ -1864,7 +1964,7 @@ bool TrackContentWidget::pasteSelection( MidiTime tcoPos, const QMimeData * md,
 		if( wasSelection )
 		{
 			tco->selectViewOnCreate( true );
-		}
+    	}
```
DomClark, line 1967: Some spaces have sneaked in at the start of the line here - change back to tabs?

DomClark (approved):
Looks good to me.

include/Track.h
DomClark: If removing it causes an error, then yes, keep it around. I don't get an error removing it on my machine though - what error do you get?

src/core/Track.cpp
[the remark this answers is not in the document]
DomClark: I can't reproduce this locally - can you explain what to do?

DomClark:
Also, there's some code left that's added but commented out - should it be removed instead?

src/core/Track.cpp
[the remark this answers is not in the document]
russiankumar: I tested again and it works now... I guess it had something to do with a removed piece of code. Either ways, it's now removed.

include/Track.h
russiankumar: Ah, I'm sorry, it builds now because the function that used it was removed xD. Removed the function.

src/core/Track.cpp
russiankumar: Yup, done

src/tracks/BBTrack.cpp
```diff
@@ -370,7 +286,6 @@ void BBTCOView::update()
 }
 
 
-
 QColor * BBTrack::s_lastTCOColor = NULL;
```
russiankumar, line 289: Do you think I should remove this? I don't use this in the new color implementation, and the BB track doesn't have its own color implementation anymore.

russiankumar:
Yeah, I'd left that code from testing if removing a function was safe, but forgot to remove the comments.

Also got rid of some of those inserted/removed spaces that were left as a result of functions or code that was added and then removed. I think I missed some of them, I'll fix them soon.

src/tracks/BBTrack.cpp
DomClark: Sounds good to me, yes.

russiankumar: Great, I removed it.

russiankumar:
@SecondFlight,

I think this will be the final version for the demo coloring, please do confirm:

image

I've colored it to make it aesthetically pleasing, not to serve any color-coding purpose (red for drums, blue for leads etc.)

If you're alright with this, I'll commit with the update.

russiankumar:
Alright, I've update the demo. If some changes are requested, I'll commit with another copy.

russiankumar:
I've added comments for anything that isn't obvious by the naming. If the current state of the PR is satisfactory, then I can finally call it a day on this.

IanCaio:
@russiankumar I think you added some submodules files by accident. They were updated lately, so you might have merged master with the updated submodules without running `git submodule update` on the branch PR. This happened to me before, I don't remember exactly, but I think I had to go back to the commit where master was merged and then update the submodules before moving to the next commit (so the files from the submodules are not staged).

russiankumar:
Oh no.... I think I messed up more than I fixed

russiankumar:
@IanCaio 
Umm... so I updated the submodules and now I have even more of these submodule files. Did I do something wrong?

russiankumar:
Phew... seems like it's fixed. Now I need to re-add all those comments xD

IanCaio:
I think force pushing is the only way to do it, because you have to go back in history and make new commits (at least I had to do it too when it happened to me). After you stage and commit the submodule files even if you update them they are still in the change log, so you have to go back in history like you did afaik. Maybe someone more experienced with git knows if there's a better way, because I imagine it would be a big mess to have to do that after you made lots of commits over the merge that caused that issue.

Anyways, important thing is that you got it fixed :smiley:

russiankumar:
I think this is done. @Spekular, if you approve the latest changes, could you merge this?

Oh, and also an out of context question: is there an automated system for credits, or do the PR authors have to add it in themselves?

DomClark:
The contributor list is automatically generated from git history. (At least in theory - it can be incorrect for CI builds due to shallow cloning, but we try to get it right for releases...)

SecondFlight:
Sorry for missing this for so long. The changes are good with me 👍

russiankumar:
image

Yeah, it does mention me, but it seems it pulled my nickname out of somewhere xD

Anyways, this PR is done. Thanks for all the help everyone offered! Can't wait to see this enter mainstream LMMS :)

superpaik:
The color system works really fine and I think It is something that will be greatly appreciated by the community. Thanks for the effort!

I found just one issue that I don't know if it's related to this PR or not (maybe is due to QT framework)
Windows (10 at least) has a functionality for changing the screen resolution called "Scale and layout" where you don't change the screen resolution but how the apps display its content (applying a % of increment to text and UI items). It helps in laptops with high resolution screen in order to make texts readable. Long story short. When this parameter is set to, say 125%, the "Pick Screen Color" functionality that it is inside the "Select color" popup window doesn't return the color where the cursor is. The cursor position showed in the popup window shows a value like the display has a lower resolution. When is set to 100% it works fine.

Spekular:
> I think this is done. @Spekular, if you approve the latest changes, could you merge this?

Yes :)

russiankumar:
@superpaik thanks for the test! I believe that's a QT issue, since from what I understood it has to do with the box with the color gradient. I'll look into it and if there is a fix I can think of I'll make another PR.

Glad this one is finally merged!

Issue #4554, opened 2018-08-24
Very, very old external controllers automation bug.


qnebra:
1. You need external controller for reproduce.
2. Connect for example "volume" to "mod wheel", it must be on external controller
3. Make automation from "volume"
4. Create/add/play some notes in this instrument from you get "volume" (just for listen what you automate).
5. Check "set/clear record" on automation track.
6. Record automation.
7. Disable recording on automation track, just repeat 5.
8. Play track.
9. This recorded automation is completly ignored, just doesn't work.
10. Wait, it works, only if you disconnect "volume" from "mod wheel".

Copied and pasted from discord discussion. Behaviour discovered even in 0.4.15, so that means it's a very old one. But why it is something wrong?

Because you need do many things for get listening your track with all automations, in heavy automated tracks (like 20+) connect and disconnect all many times is very, very irritating.

qnebra:
It look something like this:

- recording - here is ok
- playing - lmms still listen signal from external controller, while soft should get signal only from automation. Now lmms try get signal both from controller and automation, so this create conflict.

qnebra:
I guess it can be closed, due to fix being merged.

PR #5574, opened 2020-07-11
Automation now responds when connected to a midi controller

serdnab:
Fix for #4554.
I added a flag when the value of the midi controller changes and also when the automation value changes. So `AutomatableModel.h:value()` checks the one that last changed and remains returning this one until it changes to the other.

### DIFF
--- a/include/AutomatableModel.h
+++ b/include/AutomatableModel.h
@@ -34,6 +34,7 @@
 #include "ValueBuffer.h"
 #include "MemoryManager.h"
 #include "ModelVisitor.h"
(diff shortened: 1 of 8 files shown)


serdnab:
It is not working for now, I will reopen it later.

qnebra:
From what I can tell you and remember, issue happens when both automation and controller were connected to to the same value, VOL for example. To controller of course to change it, and into automation to record this changes. 

But I guess you go into exactly the same issue, and you know how atrocious it get when you had multiple controller connections to manually disconnect only to preview what you just recorded.

PR #5577, opened 2020-07-12
Small refactor on FxMixerView.cpp and FxMixer.cpp

IanCaio:
The codes on FxMixerView.cpp and FxMixer.cpp were using the types `TrackContainer::TrackList` and `QVector<Track *>` unconsistently. `TrackContainer::TrackList` is just a typedef for `QVector<Track *>` so it makes sense that we use it, specially in terms of readability.

Places where `QVector<Track *>` were used are now replaced with `TrackContainer::TrackList`.

Also, we were not including "TrackContainer.h" directly (the typedef was likely being included indirectly through one of the other include files), so we also include this header on both source codes.

### DIFF
--- a/src/core/FxMixer.cpp
+++ b/src/core/FxMixer.cpp
@@ -34,6 +34,7 @@
 #include "InstrumentTrack.h"
 #include "SampleTrack.h"
 #include "BBTrackContainer.h"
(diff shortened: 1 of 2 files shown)


Spekular:
Since FxMixerView already used TrackContainer::TrackList, I don't see any arguments against being consistent with it. Nice catch!

PR #5579, opened 2020-07-13
Fix crash on CLI rendering

### LINKED ISSUE #5107: Segfault at begin of CLI rendering
### version
LMMS 1.2.0.495
(Linux x86_64, Qt 5.11.3, GCC 9.1.0)
from git.

### gdb session
```
(gdb) r  render /home/ps/lmms/projects/1-10.mmpz -o /home/ps/lmms/projects/export/1-10.wav -f wav -s 192000 -b 320 -i sincbest -x 1
...

Thread 15 "ProjectRenderer" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7faa2a5b6700 (LWP 18310)]
0x00000000005f9910 in MidiTime::operator int() const () at /usr/src/lmms/src/core/midi/MidiTime.cpp:121
(gdb) bt
#0  0x00000000005f9910 in MidiTime::operator int() const () at /usr/src/lmms/src/core/midi/MidiTime.cpp:121
#1  0x00000000005c28be in TimeLineWidget::loopBegin (this=0x0) at /usr/src/lmms/include/TimeLineWidget.h:128
#2  Song::startExport (this=0x1725780) at /usr/src/lmms/src/core/Song.cpp:766
#3  0x00000000005b70ec in ProjectRenderer::run (this=0x68b5040) at /usr/src/lmms/include/Engine.h:77
#4  0x00007faa726488dc in ?? () from /m/qt/lib/libQt5Core.so.5
#5  0x00007faa74c51fc5 in start_thread (arg=<optimized out>) at pthread_create.c:486
#6  0x00007faa722bb7af in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
(gdb) 
```
No problems with GUI export.

PhysSong:
Fixes #5107.
This is a temporary workaround. To make all export options available in CLI, some properties of `TimeLineWidget` should be moved to a new core class.

### DIFF
--- a/src/core/Song.cpp
+++ b/src/core/Song.cpp
@@ -767,12 +767,16 @@ void Song::startExport()
 			m_exportSongEnd += MidiTime(1,0);
         
 		m_exportSongBegin = MidiTime(0,0);
-		m_exportLoopBegin = m_playPos[Mode_PlaySong].m_timeLine->loopBegin() < m_exportSongEnd && 
-			m_playPos[Mode_PlaySong].m_timeLine->loopEnd() <= m_exportSongEnd ?
-			m_playPos[Mode_PlaySong].m_timeLine->loopBegin() : MidiTime(0,0);
-		m_exportLoopEnd = m_playPos[Mode_PlaySong].m_timeLine->loopBegin() < m_exportSongEnd && 
-			m_playPos[Mode_PlaySong].m_timeLine->loopEnd() <= m_exportSongEnd ?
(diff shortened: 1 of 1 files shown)


russiankumar:
Tested, does not crash.

Got some "Not enough samples" errors:

image

image

I think it's a problem with the project file though, and this PR fixed what it was supposed to fix.

GUI export works as expected as well.

PR #5581, opened 2020-07-14
Adds support for MIDI CC events inside LMMS

IanCaio:
This PR implements a feature that I personally consider very useful for LMMS and hope others share the thought!

MidiCC

## Feature Description

A MIDI CC rack is added to LMMS where there are Knobs for every single CC controller (0-127) linked to each Instrument Track. When the rack is enabled (LED button) any change to the values of those Knobs will trigger MIDI CC events to that particular track with the current value of the controller, very similar to what happens when we connect a hardware MIDI keyboard with controllers to the track.

Various plugins, like Carla-patchbay (and hence Lv2 plugins as Surge, TAL-Filter, TAL-Reverb, etc) and ZynAddSubFX accept mapping parameters to CC controllers. Since those Knobs can be automated, this feature basically makes automating any mapped parameters possible! This opens the doors to lots of fancy automation possibilities that before were either not possible, required live recording or required fancy hacking to do.

The way it works now, those events are only generated when a Knob's value is changed. If the MIDI CC rack is disabled on a track, the value of the Knob is changed and it's reenabled though, it won't generate a event until the value is once again changed. That behavior resembles MIDI keyboards as well: if you turn on a keyboard with a slider on value 50 for example, it will only trigger a MIDI CC event when you change the value, so I believe it should be kept that way.

## Technical Stuff:

For implementing that feature, a new class called `MidiCCRackView` was created. It basically creates the whole GUI for the MIDI CC rack (ComboBox for tracks and Knobs for controllers). Lots of connections are made to keep the ComboBox up to date with the current Instrument Tracks available and there's also a `TrackContainer::TrackList` member on the class holding pointers to the Instrument Tracks objects themselves.

The `InstrumentTrack` class now also have one `BoolModel` called `m_midiCCEnable` (which tells if the CC controllers are enabled/disabled) and also a `FloatModel` for each CC controller. The `dataChanged()` event of each CC controller `FloatModel` is connected to a SLOT method called `processCCEvent(int controller)`, which will check if `m_midiCCEnable` is true, and if it is it will process the MIDI CC events to the output channel of the track.

Every time a track is selected on the MIDI CC rack ComboBox, the LED button and Knobs are pointed to the models of that particular track. The proper code for adding the rack on the `MainWindow` and `GuiApplication` classes was also added.


src/tracks/InstrumentTrack.cpp
```diff
@@ -126,6 +128,25 @@ InstrumentTrack::InstrumentTrack( TrackContainer* tc ) :
...
+
+	// Initialize the MIDI CC controller models and connect them to the method that processes
+	// the midi cc events
+	QSignalMapper *midiCCSignalMapper = new QSignalMapper( this );
+
+	for( int i = 0; i < MIDI_CC_MAX_CONTROLLERS; ++i )
+	{
+		m_midiCCModel[i] = new FloatModel(0.0f, 0.0f, 127.0f, 1.0f, NULL, QString("CC Controller ") + QString::number(i));
+		connect( m_midiCCModel[i], SIGNAL( dataChanged() ),
+			midiCCSignalMapper, SLOT( map() ) );
```
PhysSong, line 143: You can use a lambda function here like this:
 https://github.com/LMMS/lmms/blob/619b1d67cd6f6361d2546ce7b8e157e7598d702d/plugins/VstEffect/VstEffectControls.cpp#L93-L94

PhysSong: I think you need to add `Qt::DirectConnection`. Without this, the UI thread will handle automated CC changes, which is not desired.
There's a problem, `processOutEvent` is realtime unsafe. However, LMMS is already calling it on realtime threads, so I think it's not a big problem here.

IanCaio: My first attempt was using lambda functions, but I was running into compilation errors and just now reading your comment I realized why. I was still using the `SIGNAL()` macro, so `connect` expected a `const char *` where I used a lambda function (probably has to do with the `SIGNAL/SLOT` shenanigans). I changed the second parameter to `&FloatModel::dataChanged` and it compiled fine. I'll soon update the PR with the change!

Btw, I got a little curious, `processOutEvent` currently opens space for some race-conditions? I'd guess it being called so often would make us see these type of bugs at least some times. So in the future we might have to implement a `mutex` on it?

IanCaio: Just updated the code to use lambda functions instead, I'll ping you later on Discord to ask more about the thread safeness of `processOutEvent`!

src/tracks/InstrumentTrack.cpp
```diff
@@ -35,6 +35,7 @@
 #include <QMessageBox>
 #include <QMdiSubWindow>
 #include <QPainter>
+#include <QSignalMapper>
```
PhysSong, line 38: You can remove this as well.

IanCaio: Thanks! Forgot to remove it too :facepalm:

russiankumar:
Hey, I think you've forgotten to add a MIDI CC entry in the view menu. That can be found in MainWindow.cpp I believe, and it's probably some two lines of code.

IanCaio:
Took me a while, but I just added the item in the View Menu for the MIDI CC Rack!

russiankumar:
Amazing!

include/InstrumentTrack.h
```diff
@@ -229,6 +233,8 @@ class LMMS_EXPORT InstrumentTrack : public Track, public MidiEventProcessor
 	void newNote();
 	void endNote();
 
+public slots:
+	void processCCEvent(int controller);
```
russiankumar, line 237: Picky, but I believe LMMS uses `( element )` format instead of `(element)`, so request you to change this for consistency

russiankumar: Checking other code slices though `(element)` is used, so if this file uses `(element)` you can leave it as is

PhysSong: We used to use `( foo )`, but `(foo)` is the standard now.
https://github.com/LMMS/lmms/wiki/Coding-Conventions#braces

russiankumar: Yep, but most of the files are still on using `( element )`. I was suggesting to stick to that until a project-wide reformat takes place

IanCaio: Since automatically reformatting is going to be a thing soon, when it comes to braces and small code format details I tend to use what is being used in the file I'm changing to keep it at least consistent (unless it's some terrible formatting mistake). On this file both styles are used, but the method right below mine uses `(element)`, so that's probably why I chose it. As @PhysSong pointed out, this is actually the standard now (not that I had that in mind when I wrote it), so should I keep it as it is?

russiankumar: Yeah, I have no issues. I couldn't see the line below, so I assumed it uses the spaced brackets.

IanCaio: Alright, thanks! I'll mark this as resolved then :)

include/InstrumentTrack.h
```diff
@@ -221,6 +222,9 @@ class LMMS_EXPORT InstrumentTrack : public Track, public MidiEventProcessor
 
 	void autoAssignMidiDevice( bool );
 
+	BoolModel *m_midiCCEnable;
+	FloatModel *m_midiCCModel[MIDI_CC_MAX_CONTROLLERS];
```
PhysSong, line 226: I think you can use `MidiControllerCount` from `Midi.h` and remove the `MidiCCRackView.h` include.

include/MidiCCRackView.h
```diff
@@ -0,0 +1,49 @@
...
+	void updateKnobsModels();
+
+private:
+	ComboBox *m_trackComboBox;
+	ComboBoxModel *m_trackComboBoxModel;
+	TrackContainer::TrackList m_tracks; // List with pointers to the ComboBox tracks themselfs
+
+	GroupBox *m_midiCCGroupBox; // MIDI CC GroupBox (used to enable disable MIDI CC)
+
+	Knob *m_controllerKnob[MIDI_CC_MAX_CONTROLLERS]; // Holds the knob widgets for each controller
```
PhysSong, line 45: Here as well.

src/gui/widgets/MidiCCRackView.cpp
```diff
@@ -0,0 +1,197 @@
+#include <Qt> // For Qt::WindowFlags
```
PhysSong, line 1: Please add the copyright info at the beginning like existing files.

PhysSong: Also, this line is unnecessary because many headers include `qnamespace.h`.

src/tracks/InstrumentTrack.cpp
```diff
@@ -249,6 +263,24 @@ MidiEvent InstrumentTrack::applyMasterKey( const MidiEvent& event )
...
+	if( !m_midiCCEnable->value() )
+	{
+		return;
+	}
+
+	uint8_t channel = static_cast<uint8_t>( midiPort()->realOutputChannel() );
+	uint16_t cc = static_cast<uint16_t>( controller );
+	uint16_t value = static_cast<uint16_t>( m_midiCCModel[ controller ]->value() );
+
+	processOutEvent( MidiEvent( MidiControlChange, channel, cc, value ) );
```
PhysSong, line 278: Both `processInEvent` and `processOutEvent` passes the MIDI events to the instruments, but each of them has their unique functionality:
- `processInEvent`: handle some MIDI events which aren't handled anywhere else
- `processOutEvent`: send MIDI events to external MIDI devices

I don't know your intention, but make sure you're using the correct one.

IanCaio: I was reading `processInEvent` and `processOutEvents` again and I think I wasn't that off in understanding what they do. It can get a little bit confusing at some points though. Here are my thoughts:

Apparently, `processInEvent` processes events that are received through the MIDI Input port. That includes everything external sending events to the instrument track (either other tracks, MIDI keyboards/controllers or the piano widget itself). Some of those events will be processed in a particular way (i.e.: `MidiNoteOn` will make the instrument track create a `NotePlayHandle` for the note, etc), and most of them (but not all) will also trigger a `processOutEvent`, which will send the event further down the MIDI ports chain. MIDI Control Change events are not among those that trigger a `processOutEvent`. What that means is that the instrument plugin will receive the MIDI CC event through the `handleMidiEvent` method, but if you select another track on the MIDI Output dialog it will not receive those messages.

`processOutEvent` basically processes events generated by the instrument track that will be sent out through the MIDI Out ports. All tracks that are connected to this track through MIDI Input/Output ports will receive messages processed through `processOutEvent` (that are not blocked by the MIDI output channel filter of course).

So the decision between both of them resides on how do we want to see those MIDI CC events being interpreted as: Are they supposed to be an emulation of a hardware MIDI controller and behave **exactly** like we were using a physical one or should they be thought as events generated by the instrument track itself? I didn't have enough time to weight pros/cons of each approach so I left the code as is and added a TODO comment so I come back to this. Honestly, as I write this I think emulating a physical hardware makes sense (using `processInEvent`), though then some knobs are interpreted in a way that might not be obvious to the user (Knob 64 will enable/disable sustain and Knobs 120/123/124/125/126/127 will silence all notes). Still sounds like a good idea, since the same would happen with a physical MIDI controller.

IanCaio:
Decided to process MIDI CC rack events exactly the same way as hardware MIDI controller events. More details on the reasoning in the last commit message.

Be aware that because MIDI CC events can't be forwarded through MIDI Output ports when they are processed through `processInEvent`. So now it's not possible to debug them through a MIDI monitor like `kmidimon`. The Instrument Plugins will still receive them and process accordingly though.

src/tracks/InstrumentTrack.cpp
```diff
@@ -249,6 +263,25 @@ MidiEvent InstrumentTrack::applyMasterKey( const MidiEvent& event )
...
 
+void InstrumentTrack::processCCEvent(int controller)
+{
+	// Does nothing if the LED is disabled
+	if( !m_midiCCEnable->value() )
+	{
+		return;
+	}
+
+	uint8_t channel = static_cast<uint8_t>( midiPort()->realOutputChannel() );
```
PhysSong, line 274: I missed this, but does it make sense to use the output channel with `processInEvent`?

IanCaio: I think it does, because the event isn't actually being received by the input MIDI ports, but being generated by the track and then forwarded to the instrument plugin. The reason I used `processInEvent` was for it to be processed just as a MIDI event received through a physical controller would be (there are some CC events that are handled by LMMS like sustain events for example). Since the notes generated by the track are forwarded to the plugin through the output channel, I thought generated CC messages should be too.

You think it makes more sense to use the input channel instead?

PhysSong: Hard to decide. Can we compare with other DAWs?

IanCaio: Of course! If other DAWs use the input channel instead or if most people think it's a better alternative we can change it. Personally I thought using the output channel would be more intuitive. Whichever we decide for, we should document it on the LMMS wikia since it's something the regular user can't easily find out (since with `processInEvent` processing the event we can't use a MIDI monitor to see the channel being used afaik, so the user has to know it before hand).

wujekbrezniew:
I found a bug in 702 build with automation tracks. 
To replicate create automation track with CC knob connected to it and start playing the track. Then, while playing, do "save as" and save project. After reopening LMMS, daw crashes when you try to open saved project with CC automation track.

PhysSong:
I got a backtrace for the crash with a MinGW build.
bt
It's somewhere in `InstrumentTrack::processInEvent()`.

PhysSong:
I found that `InstrumentTrack::processCCEvent()` is called when loading `m_midiCCModel`, and `m_instrument` is null at that moment. In other words, it's a matter of loading order.

PhysSong:
A simple workaround is saving MIDI CC stuff after saving the instrument. It will allow sending CC events to the instrument on loading.
As a double safety check, you may check the nullity of `instrument()` in `InstrumentTrack::processInEvent()` before forwarding MIDI events.

IanCaio:
@PhysSong Thanks for the help with figuring this out! I added a fix which takes both approaches we discussed:

- Having a sanity check on the handleMidiEvent call inside processInEvent
- Having MIDI CC disabled during the track being loaded so it doesn't trigger processInEvent calls.

I thought about adding a conditional similar to the one in `processOutEvent` and just returning the method if `m_instrument` was NULL, but I was unsure if the method is supposed to be called in certain situations where there isn't a loaded instrument yet, so I kept it contained to the only `handleMidiEvent` call on the method.

wujekbrezniew:
Tested. Works like a charm. C:

src/tracks/InstrumentTrack.cpp
```diff
@@ -375,7 +413,9 @@ void InstrumentTrack::processInEvent( const MidiEvent& event, const MidiTime& ti
 			break;
 	}
 
-	if( eventHandled == false && instrument()->handleMidiEvent( event, time, offset ) == false )
+	// If the event wasn't handled, check if there's a loaded instrument and if so send the
+	// event to it. If it returns false means the instrument didn't handle the event, so we trigger a warning.
+	if( eventHandled == false && instrument() && instrument()->handleMidiEvent( event, time, offset ) == false )
```
PhysSong, line 418: The event will not be handled if `instrument()` is null, so you should change the conditional.

PhysSong:
I have a question: should we trigger the CC event right after loading the instrument, or not?

src/tracks/InstrumentTrack.cpp
IanCaio: Nice catch! If I have `eventHandled == false && ( instrument() == nullptr || instrument()->handleMidiEvent( event, time, offset ) == false )` the call to handleMidiEvent will be skipped when instrument is null right? And just a readability question, should I write `instrument() == nullptr` or `!instrument()` there? I'm thinking the former is more readable even though a little more verbose.

IanCaio:
> I have a question: should we trigger the CC event right after loading the instrument, or not?

I don't think that's necessary. They will be triggered on the first CC value change and most likely the value of whatever model they are controlling will have been saved too on the project file, so upon loading they should have a value that matches the position of the CC knob afaik. I'm open ears if someone thinks it should, would probably be just a small change on the end of the InstrumentTrack constructor anyways.

src/tracks/InstrumentTrack.cpp
PhysSong: You're right. I prefer `!foo` to `foo == nullptr`, but there are no rules for comparing values with `nullptr`.
Also, there are some alternatives like `eventHandled == false && !( instrument() && instrument()->handleMidiEvent( event, time, offset ) )`.

IanCaio:
Mistake on the conditional of `processInEvent` that triggers a "not-handled MIDI event" warning fixed on the last commit!

IanCaio:
Alright, brace yourselves for some long text as this last commit is a little bit dense (not in number of lines, but in the reasons behind them) :smile: 

I noticed a bug when I had a CC model automated (hence assigned an ID) and then loaded another project that had this same ID assigned to something. It would break the loaded projects automation and trigger a warning of "ID already in use". Hunting the core issue, I figured out I was not deleting the model. When I loaded another project, the CC model was hanging (the instrument that pointer to it was destroyed) but still existent, so it didn't release the ID and I couldn't assign it to the model from the loaded project.

The solution was then simple: Delete the models, so the IDs are released and there are no more memory leaks. However there was a small complication. The MIDI CC rack has 1 LED and 128 Knobs (model views) that are connected to the selected track's CC models. If I deleted the models from the track selected on the MIDI CC rack, I'd trigger a SegFault because the Knobs and LED would be pointing to unexistent models. So before deleting the models, I needed to make sure the model views weren't pointing to them anymore. There wasn't a `unsetModel` method to Model Views, but I noticed ModelView.cpp handled NULL models. So I thought I could just call `setModel( NULL )`. Problem is, ModelView.cpp handles NULL models but classes that inherit from it (like AutomatableModelView) don't. That IMHO is a design flaw, but that's outside the scope of this PR. I'll further discuss this issue on Discord, but I'm writing it here just as a reference and justification for the approach I had.

Since I can't call `setModel( NULL )`, my solution was to create a `unsetModel()` method that creates a dummy model and calls `setModel` to it. This dummy model is created as a _defaultConstructed_ model and because of that it's destroyed when another call to `setModel` is made. I also created a method on MidiCCRackView.cpp called `unsetModels()`, which is just a convenience method to unsetting all the model views from MIDI CC rack.

With those methods in hand, all I needed to do was have the following sequence on the Instrument Track's destructor:

1. Call `MidiCCRackView::unsetModels()` if the MIDI CC Rack is accessible. That's because the destructor might be called when LMMS is closed, and in that case the MIDI CC Rack might be already destroyed and not accessible anymore.
2. Delete the models related to CC events. There's no need to call `updateKnobsModels` explicitly, because after the track is destroyed a `trackRemoved` signal is going to be triggered, which will call the update knobs method.

That seemed to fix the bug and the memory leak associated with it. I also added code to the `GuiApplication::childDestroyed` method to assign `nullptr` to `m_midiCCRackView` when it was destroyed, same way it's done to all other children of GuiApplication, since I apparently forgot it before.

wujekbrezniew:
Tested again. Problems with loading gone. Works as stable.

Spekular:
Did some grepping, `InstrumentTrackWindow::modelChanged()` calling `m_pitchKnob->setModel( NULL );` is the only instance of setModel being explicitly called with NULL, and the instrument window caching is something we should remove. With that exception aside, it seems to me that ModelViews are never unset in the existing codebase.

Spekular:
Ok, thoughts after reading the diff as well:
I still feel a little weird about this PR introducing the concept of un-setting a view's model, since it was never used before. If no one else takes issue with it I won't complain though. Not having seen the alternative I can't say for sure, but reusing the models doesn't seem to convolute the code all that much.

I do wonder though, would it make sense to want to view CC racks for several tracks at once? If you would ever want to link two CC signals, it seems like you'd want to have several racks open at once, with different track's controls in each one.

IanCaio:
> Did some grepping, `InstrumentTrackWindow::modelChanged()` calling `m_pitchKnob->setModel( NULL );` is the only instance of setModel being explicitly called with NULL, and the instrument window caching is something we should remove. With that exception aside, it seems to me that ModelViews are never unset in the existing codebase.

I see, thanks for searching further about that, because what I intended with bringing that up was exactly making us review this behavior in the code and discuss it. If you look at `Knob.cpp` you'll see that some methods are "NULL-safe" and others are not (i.e.: `updateAngle` is safe to be called with a NULL model, `drawKnob` is not). I think the ones related to drawing or that are supposed to be called when a widget is visible (i.e.: context menu, mouse events) are the ones unsafe, so we get away with setting it to NULL there because a couple lines before `m_pitchKnob->hide()` is called, so those unsafe methods are probably not called while model is NULL.

I personally still think it's somewhat a bad design to have this kind of behavior not well defined. We either accept NULL or we don't, but having that depend on a runtime sort of obscure condition like the widget being hidden makes it less clear to whoever is coding when it's alright to set it to NULL. That's obviously just a discussion I wanted to bring up that don't necessarily requires a immediate change, but it's good that we have that in mind for a possible refactor in the future where we either make NULL completely unacceptable or make it acceptable on the classes that inherit from `ModelView` as well in a more obvious way. That is, of course, if others agree this is an issue.

> Ok, thoughts after reading the diff as well:
> I still feel a little weird about this PR introducing the concept of un-setting a view's model, since it was never used before. If no one else takes issue with it I won't complain though. Not having seen the alternative I can't say for sure, but reusing the models doesn't seem to convolute the code all that much.
> 
> I do wonder though, would it make sense to want to view CC racks for several tracks at once? If you would ever want to link two CC signals, it seems like you'd want to have several racks open at once, with different track's controls in each one.

Thanks for the review! I haven't thought of that possibility but that's definitely something to consider. In that case the GUI would have to be rethought though, to give the user the possibility of having different racks opened at once. Maybe a dropdown menu where the track could be selected from on the MainWindow MIDI CC Rack button which would open that track's rack (but then the shortcut wouldn't work anymore, or it would have to open the dropdown menu instead of the rack). Another possibility I thought was maybe having the rack as another tab on the Instrument Widget. But then, if we are soon getting rid of the Instrument Widget Cache that would probably have to wait or there will be some heavy conflicts to solve once that is merged.

I didn't think long on the alternatives so maybe there's a better one I'm missing, but for sure that's something that could be changed if having more than one rack opened at once makes sense in the UX point of view!

PhysSong:
> I do wonder though, would it make sense to want to view CC racks for several tracks at once?

I think yes.

Spekular:
> I think yes.

In that cas it makes a lot of sense to me to launch the rack via the instrument track's cog menu, and let each rack have it's own set of knobs.

wujekbrezniew:
Maybe even something like this where CC rack is an additional pop-up menu sticked to instrument widget.
image

IanCaio:
I'm still very newbie when it comes to making GUIs with Qt, so I might need some help (the current GUI is a result of much googling and frustration :rofl: ). Good side is I might be able to reuse the GUI setup code, except for making the knobs and led individual and getting rid of the dropdown menu.

But can I consider it decided then, that the MIDI Rack should be opened from the Instrument Track? @Spekular the _cog_ menu you meant is the context menu on the song editor (the little gear)? Can't promise, but I might be able to start working on it today.

Spekular:
> But can I consider it decided then, that the MIDI Rack should be opened from the Instrument Track?

If you don't want it in the instrument window (which I agree with*) then it seems to be the most suitable place IMO.  The track options already have a "MIDI" submenu where you could put the entry, and it would make it very obvious which track's rack is being opened.

*The instrument MIDI tab is already cramped, and putting it in the Miscellaneous tab would be poor categorization.

> @Spekular the _cog_ menu you meant is the context menu on the song editor (the little gear)?

Yup, the cog/cogwheel/gear on the left side of the track label with clone/remove/clear this track options.

IanCaio:
> If you don't want it in the instrument window (which I agree with*) then it seems to be the most suitable place IMO. The track options already have a "MIDI" submenu where you could put the entry, and it would make it very obvious which track's rack is being opened.

Looks like a good place to have it. I should be working on the change soon!

Just a reminder to myself:
1) Change MIDI CC events to be processed through `processOutEvent` so they are not ignored during the song export.
2) Change the MIDI CC Rack to be individual for each track and be opened through the context menu of the TrackView.

PhysSong:
> Change MIDI CC events to be processed through `processOutEvent` so they are not ignored during the song export.

If we can differentiate MIDI events from external controllers and CC knobs, you don't need this.

IanCaio:
> If we can differentiate MIDI events from external controllers and CC knobs, you don't need this.

Afaik there are not many advantages to treat those CC events like `Input` events instead of `Output`, so I'm not sure it's worth the trouble of changing MIDI headers to be able to differentiate between events from hardware and events from MIDI CC Rack. I think using `processOutEvent` would keep it simpler. Any drawbacks with using it instead?

PhysSong:
> I think using `processOutEvent` would keep it simpler. Any drawbacks with using it instead?

It will make some differences with CC 64(sustain).

IanCaio:
> It will make some differences with CC 64(sustain).

True, I think there are a few other CC events that silence notes that also would behave slightly different. They are not very useful in the context of exporting a track but I get it that having the behavior closer to the hardware controller would be better (that's why I chose `processInEvent` the first time). Do you have any suggestions to identify the CC events as originated from the MIDI CC Rack? For a second I thought of using MidiMetaEvent but I realize this is part of the MIDI standard and I probably should use something else. Maybe add a method called `isHardware()` on the MidiEvent class and add `m_fromHardware` to all the constructors as defaulting to `true` and set it to `false` when creating a CC event from the rack?

By the way, I made a commit yesterday starting the GUI change so every InstrumentTrack has its own rack. As I said on discord, I'm kind of learning about Qt's GUI as I go, so it could benefit from some careful review. There are a couple of things I still want to change on it like creating the window on a slightly different position each time so I don't create a window right over another, and maybe using an action with a checkbox on the MIDI menu instead of a default action.

IanCaio:
Fixed bug where MIDI CC rack events were ignored during the track export. For that I created a variable inside `class MidiEvent` called `m_fromHardware` to signal whether the event should be ignored during export as if it were generated by hardware.

As I wrote on the commit message, I'm still not completely happy with the parameter name. It kind of implies that the event was generated by hardware controllers which could not be true (they would just be ignored on `processInEvent` as if they were). For lack of a better name for now I kept it that way, but I'm open for ideas. Was thinking something along the lines of `ignoreOnExport`.

I also would like to know other's opinions on that approach, since I'm not sure if there's anything inside `processInEvent` that shouldn't be called during an export. There are calls for `Engine::mixer()->doneChangeInModel()`, `Engine::mixer()->addPlayHandle()` and `NotePlayHandleManager::acquire()` in the case of MidiNoteOn and MidiNoteOff events. I know I'm using it for MidiControlChange, which is apparently safe, but opening that exception on the conditional could be a bad idea if it opens the possibility of a MidiEvent triggering one method that shouldn't be called during an export.

include/MidiEvent.h
```diff
@@ -37,34 +37,38 @@ class MidiEvent
...
 		m_sysExData( sysExData ),
-		m_sourcePort( NULL )
+		m_sourcePort( NULL ),
+		m_fromHardware( fromHardware )
 	{
 		m_data.m_sysExDataLen = dataLen;
 	}
 
-	MidiEvent( const MidiEvent& other ) :
+	MidiEvent( const MidiEvent& other, bool fromHardware = true ) :
```
PhysSong, line 64: Do you really need `fromHardware` for the copy constructor?

IanCaio: Sorry about that, I'm changing it so it just copies the value from the other event!

wujekbrezniew:
Tested latest build. Now CC automations are exported. 
I found one bug with CC 0 knob. Surge does not recognize it as CC signal and LABS recognize it as CC 32. Idk if it is connected straight this CC Rack or maybe it is something in MIDI CC specification.

I think that green LED button in rack should be removed. For now it has no purpose for me. And i think that possibility of choosing MIDI CC channel should be added.

Picture of green LED button thing:
image

IanCaio:
Thanks again for testing man!

Some Control Change numbers have specified functions, so the plugin might reserve that for that particular use. CC0 is reserved for "Bank Select", so maybe Surge uses it to cycle through patches or something. CC120 for example is "All sounds off" (LMMS itself will respond to that CC event by stopping current played notes). I could add a debug message here and test just to make sure that the CC event being sent is the right one, but different plugins might respond differently to some messages.

I'm for keeping the LED button: it doesn't clutter the GUI and allows the user to quickly disable the CC events for that track. But I could change it so it defaults to "On" if that's preferred!

I remember reading that PR. If I recall right falkTX mentioned that he was working on exposing some parameters inside Carla, then the PR could use those exposed parameters instead of the workaround it was using at the time. I'm not sure if the exposed parameters were introduced already, but when they do I think that PR could coexist with this one. They don't conflict and it could even be more practical on some situations. Currently the PR seems to be stale though.

PhysSong:
Is this ready? I can't find any major issues, so I wonder if our first nightly can include this one.

IanCaio:
It has been a while I didn't merge master to this branch, but I think it's all good, no conflicts apparently! All the issues reported were fixed. I couldn't think of a more appropriate name for the MidiEvent variable than `m_fromHardware` yet, and that definitely shouldn't hold this PR. If something more accurate comes to mind it can be easily changed.

IanCaio:
Added 3 commits + a master merge. The 3 commits apply the following changes:

[b8081a4](https://github.com/LMMS/lmms/pull/5581/commits/b8081a4b890f333acfe7f351a94d99ada2bd7eac):
- Changes the name of the variable used in MidiEvent from `m_fromHardware` to `m_ignoreOnExport`, since the latter is more accurate with the purpose of the variable
- Fixes code style issues throughout the PR (mostly spacing and some NULL being changed to nullptr)

[19018c3](https://github.com/LMMS/lmms/pull/5581/commits/19018c3ded877bdfbb5d6b9c7505f5577157134c):
- Remove some signals that were previously added to be used when all tracks shared the same MidiCCRackView, which are not necessary anymore since each track now has its own MidiCCRack (removed from `TrackContainer` and `TrackContainerView` classes).

[23911d2](https://github.com/LMMS/lmms/pull/5581/commits/23911d222487051dbc620050ce278fd9fe628649):
- Adds 3 lines of code to randomize the position of the newly created MidiCCRack. That's so they don't overlap each other all the time.

Last commit merges current master with no conflicts.

I believe the PR is ready and as @PhysSong mentioned we could include it on the first release of our alpha builds. What do you all think of waiting until the weekend for a review of the last commits (which besides the code style fixing were small changes) and then merging if there are no objections?

PhysSong:
I'd say let's merge this and make it available for more users.

include/MidiEvent.h
```diff
@@ -204,7 +218,9 @@ class MidiEvent
 
 	const char* m_sysExData;
 	const void* m_sourcePort;
-
+	bool m_ignoreOnExport;	// Should we ignore this MIDI event when exporting a project?
+				// This helps us ignoring MIDI events that shouldn't be processed
+				// during a project export, like physical controller events.
```
Veratil, line 223: Two minor problems: alignment with tabs and grammar. I think flipping the comment to above would be better, and some IDEs like vscode (maybe the C/C++ extension though) will use comments above as description of variable if no doctext is found.
```suggestion
	// This helps us ignore MIDI events that shouldn't be processed
	// during a project export, like physical controller events.
	bool m_ignoreOnExport; // Should we ignore this MIDI event when exporting a project?
```

src/gui/AutomatableModelView.cpp
```diff
@@ -142,6 +142,31 @@ void AutomatableModelView::setModel( Model* model, bool isOldModelValid )
 
 
 
+// Unsets the current model by setting a dummy empty model. The dummy model is marked as
+// "defaultConstructed", so the next call to setModel will delete it.
+void AutomatableModelView::unsetModel()
+{
+	if (dynamic_cast<FloatModelView*>(this))
+	{
+		setModel(new FloatModel( 0, 0, 0, 1, nullptr, QString(), true));
```
Veratil, line 151:
```suggestion
		setModel(new FloatModel(0, 0, 0, 1, nullptr, QString(), true));
```

src/gui/widgets/MidiCCRackView.cpp
```diff
@@ -0,0 +1,169 @@
...
+	// Remove maximize button
+	Qt::WindowFlags flags = subWin->windowFlags();
+	flags &= ~Qt::WindowMaximizeButtonHint;
+	subWin->setWindowFlags(flags);
+
+	// Adjust window attributes, sizing and position
+	subWin->setAttribute(Qt::WA_DeleteOnClose, false);
+	subWin->move(
+		rand() % 731 + 50, // X is choosen randomly from 50 to 780
+		rand() % 331 + 50 // Y is choosen randomly from 50 to 380
```
Veratil, line 70: Why random?

src/gui/widgets/MidiCCRackView.cpp
```diff
@@ -0,0 +1,169 @@
...
+	QGridLayout *knobsAreaLayout = new QGridLayout();
+
+	knobsArea->setLayout(knobsAreaLayout);
+	knobsScrollArea->setWidget(knobsArea);
+	knobsScrollArea->setWidgetResizable(true);
+
+	knobsGroupBoxLayout->addWidget(knobsScrollArea);
+
+	// Adds the controller knobs
+	for (int i = 0; i < MidiControllerCount; ++i){
```
Veratil, line 110:
```suggestion
	for (int i = 0; i < MidiControllerCount; ++i)
	{
```

src/gui/widgets/MidiCCRackView.cpp
```diff
@@ -0,0 +1,169 @@
...
+		m_controllerKnob[i]->setLabel(QString("CC %1").arg(QString::number(i)));
+		knobsAreaLayout->addWidget(m_controllerKnob[i], i/3, i%3);
+	}
+
+	// Set all the models
+	// Set the LED button to enable/disable the track midi cc
+	m_midiCCGroupBox->setModel(m_track->m_midiCCEnable);
+
+	// Set the model for each Knob
+	for (int i = 0; i < MidiControllerCount; ++i){
```
Veratil, line 121:
```suggestion
	for (int i = 0; i < MidiControllerCount; ++i)
	{
```

src/tracks/InstrumentTrack.cpp
```diff
@@ -121,6 +122,19 @@ InstrumentTrack::InstrumentTrack( TrackContainer* tc ) :
...
 
+	// Initialize the m_midiCCEnabled variable, but it's actually going to be connected
+	// to a LedButton
+	m_midiCCEnable = new BoolModel(false, nullptr, "Enable/Disable MIDI CC");
+
+	// Initialize the MIDI CC controller models and connect them to the method that processes
+	// the midi cc events
+	for (int i = 0; i < MidiControllerCount; ++i)
+	{
+		m_midiCCModel[i] = new FloatModel(0.0f, 0.0f, 127.0f, 1.0f, NULL, QString("CC Controller ") + QString::number(i));
```
Veratil, line 133: The text alone is 115 characters wide. Add 2 tabs at 4 spaces that's 123 characters, at 8 spaces (like GitHub shows for this file) it's 131 characters. Should split it somehow.
```suggestion
		m_midiCCModel[i] = new FloatModel(
			0.0f, 0.0f, 127.0f, 1.0f, NULL, QString("CC Controller ") + QString::number(i));
```

src/tracks/InstrumentTrack.cpp
```diff
@@ -244,9 +266,27 @@ MidiEvent InstrumentTrack::applyMasterKey( const MidiEvent& event )
...
+	processInEvent(MidiEvent(MidiControlChange, channel, cc, value, NULL, false));
+}
+
+
+
+
 void InstrumentTrack::processInEvent( const MidiEvent& event, const MidiTime& time, f_cnt_t offset )
 {
-	if( Engine::getSong()->isExporting() )
+	if(Engine::getSong()->isExporting() && event.ignoreOnExport())
```
Veratil, line 289:
```suggestion
	if (Engine::getSong()->isExporting() && event.ignoreOnExport())
```

src/tracks/InstrumentTrack.cpp
```diff
@@ -375,9 +415,11 @@ void InstrumentTrack::processInEvent( const MidiEvent& event, const MidiTime& ti
 			break;
 	}
 
-	if( eventHandled == false && instrument()->handleMidiEvent( event, time, offset ) == false )
+	// If the event wasn't handled, check if there's a loaded instrument and if so send the
+	// event to it. If it returns false means the instrument didn't handle the event, so we trigger a warning.
+	if(eventHandled == false && !(instrument() && instrument()->handleMidiEvent(event, time, offset)))
```
Veratil, line 420:
```suggestion
	if (eventHandled == false && !(instrument() && instrument()->handleMidiEvent(event, time, offset)))
```

src/tracks/InstrumentTrack.cpp
```diff
@@ -999,11 +1065,13 @@ InstrumentTrackView::InstrumentTrackView( InstrumentTrack * _it, TrackContainerV
 	m_panningKnob = new Knob( knobSmall_17, getTrackSettingsWidget(),
 							tr( "Panning" ) );
 	m_panningKnob->setModel( &_it->m_panningModel );
-    m_panningKnob->setHintText( tr( "Panning:" ), "%" );
+	m_panningKnob->setHintText( tr( "Panning:" ), "%" );
```
Veratil, line 1068:
```suggestion
	m_panningKnob->setHintText(tr("Panning:"), "%");
```

Veratil (changes requested):
Just style stuff

src/gui/widgets/MidiCCRackView.cpp
IanCaio: Though it was explained in Discord, just to keep it documented here: Position is set in a random range just so windows from multiple instrument's MIDI CC racks don't get created on top of each other.

src/tracks/InstrumentTrack.cpp
IanCaio: I know we usually keep one argument per line, but on this case seemed like an overkill, so I divided the arguments in two lines (also changed NULL to nullptr)

Veratil (approved):

PhysSong:
@DomClark It seems like you want to review this. Should I wait for you or merge this before releasing alpha binaries?

DomClark:
@PhysSong I'll try to review this today.

DomClark (commented):
Review points that didn't fit anywhere in the code:
* The icons use the letters "CC". Is this universal, or is someone who doesn't speak English, and maybe doesn't even use the Latin alphabet, going to be confused as to what this is?
* I think `MidiCCRackView` should go in `src/gui` rather than `src/gui/widgets` - the latter is generally for reusable things like buttons and knobs, whereas entire windows go in the former.
* Not particularly important, but could the rack be four knobs wide? The number of knobs is a multiple of four, but not a multiple of three, and it could line up nicely at the end.
* The rack view isn't destroyed when a track is removed, so they pile up over time. Delete `m_midiCCRackView` in `InstrumentTrackView`'s destructor (or, even better, keep it in a `std::unique_ptr`), and delete the subwindow in `MidiCCRackView`'s destructor. ([See how `InstrumentTrackWindow` does it](https://github.com/LMMS/lmms/blob/d797df5/src/tracks/InstrumentTrack.cpp#L1493-L1497).)
* Now that a `MidiCCRackView` doesn't need to outlive any `InstrumentTrack`, can the model unsetting code go?
* I recall some discussion about this before, but is `processInEvent`, rather than `processOutEvent`, really the right place to send the CC events? Both send them all to the instrument, but `processOutEvent` means it's possible to control external devices with them too. The only behaviours exclusive to `processInEvent` are sustain using CC64, and silencing all notes using CC120/CC123-127.

PhysSong:
> The icons use the letters "CC". Is this universal, or is someone who doesn't speak English, and maybe doesn't even use the Latin alphabet, going to be confused as to what this is?

I think this term is common for who doesn't speak English.

> I recall some discussion about this before, but is `processInEvent`, rather than `processOutEvent`, really the right place to send the CC events? Both send them all to the instrument, but `processOutEvent` means it's possible to control external devices with them too. The only behaviours exclusive to `processInEvent` are sustain using CC64, and silencing all notes using CC120/CC123-127.

It's a very tricky question. Maybe we call `processInEvent()` first and then call `m_midiPort.processOutEvent()` explicitly?

IanCaio:
@DomClark 

I finished making the requested changes. I did everything on a separate commit to make it easier to review the changes, would appreciate if you to double checked them, specially the last couple ones: I come from a C background so I haven't used smart pointers much. I was also a little bit in doubt whether the lazy creation of the MidiCCRackView should be done on the InstrumentTrackView toggle method or inside the MidiCCRackView class, but it didn't make the method much verbose so I think it's alright to keep it there.

As for the `unsetModels` code, I removed the method from MidiCCRackView since it's not needed there anymore, however I kept the `unsetModel` method on the ModelView class if everyone is alright with that. I elaborate a little on why in this comment (https://github.com/LMMS/lmms/pull/5581#issuecomment-695334291). We have a undefined/unclear behavior on our ModelView, where sometimes it's acceptable to pass a `nullptr` model and sometimes it isn't. This could lead to hard to track bugs (like one I had in the process of working on this PR), so having a `unsetModel` method gives the developer a more clear way of doing that when necessary. I'd also change the only occurrence where `nullptr` is passed to a model view to use this method instead, but that is a small change that can be done on a separate PR to avoid going even more out of scope here.

wujekbrezniew:
> > The icons use the letters "CC". Is this universal, or is someone who doesn't speak English, and maybe doesn't even use the Latin alphabet, going to be confused as to what this is?

What do you think about icon simmilar to DIN connector? They are widely used in MIDI hardware setups.
image

zonkmachine:
> What do you think about icon simmilar to DIN connector?

Great! There are din style icons around after the work with the instrument tabs: https://github.com/LMMS/lmms/pull/3569#issuecomment-302964939

IanCaio:
Forgot to mention, regarding the `processInEvent` vs `processOutEvent`, it was indeed discussed before and it was hard to get to a definite decision, each having a different pro/con. I went for the `processInEvent` (after initially using `processOutEvent`) because it is the closest we would be to an actual hardware controller behavior, with LMMS handling some special CC as if they had originated from one. The drawback as you mentioned is that we can't forward those. We could try what @PhysSong suggested and manually forward CC events if they aren't marked for being ignored on export (which means they originated from the CC rack), adding this line on the end of `case MidiControlChange` of `processInEvent`:

```
if (!event.ignoreOnExport())
{ // Ignore space indenting
   m_midiPort.processOutEvent(event, time);
}
```

@wujekbrezniew 
I like the idea, think it would look nice! I tried to get the icons from the PR linked by @zonkmachine , but they looked a bit odd in the small resolution we have for icons (20x20), which I think was why Umcaruje moved away from them throughout the PR. I'm not sure if the CC icon would confuse people that speak languages without latin characters (PhysSong being Korean would probably have better insight on this), but if someone manages to make an icon that suits this I can change it.

wujekbrezniew:
@IanCaio I created two icons in res 256x256. They should match LMMS style:
[DIN.zip](https://github.com/LMMS/lmms/files/5613086/DIN.zip)

DomClark (commented):
Made a few suggestions that I missed the first time.

* Include ordering should be updated to follow the [coding conventions](https://github.com/LMMS/lmms/wiki/Coding-Conventions#include-order), and I think some includes are no longer used and could be removed.
* How about adding the CC rack icon to the menu item used to open it?
* What's the purpose of disabling the CC knobs? It's not like disabling an effect, where it no longer changes the sound - it just freezes them at their current positions instead.

> As for the unsetModels code, I removed the method from MidiCCRackView since it's not needed there anymore, however I kept the unsetModel method on the ModelView class if everyone is alright with that.

That's fine - I still think we should disallow having a view without a model, but that's out of scope here, so if this change improves the existing code, I think it's good to have it for now.

> Forgot to mention, regarding the `processInEvent` vs `processOutEvent`, it was indeed discussed before and it was hard to get to a definite decision, each having a different pro/con. I went for the `processInEvent` (after initially using `processOutEvent`) because it is the closest we would be to an actual hardware controller behavior, with LMMS handling some special CC as if they had originated from one. The drawback as you mentioned is that we can't forward those.

Yeah - CCs from actual hardware controllers don't get forwarded to external devices either, so there's room for more general improvement here. This is out of scope I feel, so again what you've got here is fine for now.

src/tracks/InstrumentTrack.cpp
```diff
@@ -121,6 +122,27 @@ InstrumentTrack::InstrumentTrack( TrackContainer* tc ) :
 	}
 
 
+	// Initialize the m_midiCCEnabled variable, but it's actually going to be connected
+	// to a LedButton
+	m_midiCCEnable = std::unique_ptr<BoolModel>(
+		new BoolModel(false, nullptr, tr("Enable/Disable MIDI CC"))
+	);
```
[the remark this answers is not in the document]
IanCaio: I read briefly about those. I'm not sure I fully understood their advantage (I read something about unspecified parameter evaluation order resulting in a possible memory leak when using `new` that could be avoided with `std::make_unique`), but I also read they were C++14, so I ended up using the constructor that way.

IanCaio:
> Made a few suggestions that I missed the first time.
> 
>     * Include ordering should be updated to follow the [coding conventions](https://github.com/LMMS/lmms/wiki/Coding-Conventions#include-order), and I think some includes are no longer used and could be removed.

Reorganized them according to the coding conventions and removed ones that weren't being used anymore (leftovers from changing the code quite a bit from the original concept).

>     * How about adding the CC rack icon to the menu item used to open it?

Done! Using the icon provided by @wujekbrezniew on both the window and menu action.

>     * What's the purpose of disabling the CC knobs? It's not like disabling an effect, where it no longer changes the sound - it just freezes them at their current positions instead.

Initially it was thought as a convenience for the user to be able to disable the triggering of CC messages (any change in values trigger them, so if there are automations connected to the CC knob you can just disable the MIDI CC rack if you wish to play the track without the CC automations for example). But later it was also useful when loading the instrument settings, to avoid a SegFault being caused by MIDI CC events being triggered while the instrument was still loading (https://github.com/LMMS/lmms/pull/5581/commits/6abd13062a50d49a247dbfaf4baa6bed4d123eba).

src/tracks/InstrumentTrack.cpp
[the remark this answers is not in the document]
PhysSong: @IanCaio You can use C++14 features now on `master`.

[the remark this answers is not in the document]
IanCaio: @PhysSong My bad! I thought we were still at C++11. Just updated the code to use `std::make_unique`!

DomClark (approved):

zonkmachine:
@IanCaio This somehow breaks the track preview button (LED left of volume knob).

IanCaio:
Just quoting @DomClark regarding the `m_activityIndicator` issue on Discord:

> `MidiEvent( MidiNoteOn, 0, DefaultKey, MidiDefaultVelocity )` in `InstrumentTrackView::activityIndicatorPressed` is now calling the wrong constructor of `MidiEvent`.
> It used to call this one:
> 
> ```
> MidiEvent(MidiEventTypes type = MidiActiveSensing,
>             int8_t channel = 0,
>             int16_t param1 = 0,
>             int16_t param2 = 0,
>             const void* sourcePort = nullptr,
>             bool ignoreOnExport = true)
> ```
> 
> but now it calls this one:
> 
> ```
> MidiEvent(MidiEventTypes type, const char* sysExData, int dataLen, bool ignoreOnExport = true)
> ```

We are still discussing the best way to fix this problem. It wasn't a direct result of this PR, but an issue that got uncovered by the fact that now that the other constructor has 4 arguments the ADL results in the wrong constructor being picked.

PR #5590, opened 2020-07-19
Avoid leaking the track grip QPixmap at each paintEvent

Smjert:
Given that s_grip is not used anywhere else we remove it
and instead of allocating every paintEvent a new QPixmap on the heap,
which is constructed from the cached pixmap,
to then leak it, we just pass the cached pixmap directly.

ASAN report:
```
=================================================================
==222949==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 8480 byte(s) in 265 object(s) allocated from:
    #0 0x7fcfe353ef41 in operator new(unsigned long) /build/gcc/src/gcc/libsanitizer/asan/asan_new_delete.cpp:99
    #1 0x55651cb436f2 in TrackOperationsWidget::paintEvent(QPaintEvent*) /home/smjert/Development/lmms/src/src/core/Track.cpp:1964

Direct leak of 2912 byte(s) in 91 object(s) allocated from:
    #0 0x7fcfe353ef41 in operator new(unsigned long) /build/gcc/src/gcc/libsanitizer/asan/asan_new_delete.cpp:99
    #1 0x55651cb43872 in TrackOperationsWidget::paintEvent(QPaintEvent*) /home/smjert/Development/lmms/src/src/core/Track.cpp:1971
```

### DIFF
--- a/include/Track.h
+++ b/include/Track.h
@@ -459,8 +459,6 @@ private slots:
 	void clearTrack();
 
 private:
-	static QPixmap * s_grip;
-
 	TrackView * m_trackView;
 
 	QPushButton * m_trackOps;
--- a/src/core/Track.cpp
+++ b/src/core/Track.cpp
@@ -1830,10 +1830,6 @@ void TrackContentWidget::setEmbossColor( const QBrush & c )
 // trackOperationsWidget
 // ===========================================================================
 
-
-QPixmap * TrackOperationsWidget::s_grip = NULL;     /*!< grip pixmap */
-
-
 /*! \brief Create a new trackOperationsWidget
  *
  * The trackOperationsWidget is the grip and the mute button of a track.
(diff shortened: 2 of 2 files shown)


PhysSong:
Hi, thank you for your contribution!
The purpose of using a static variable is to cache the pixmap, but the original code does something wrong. I think `std::unique_ptr` may be useful in this case.

Smjert:
> Hi, thank you for your contribution!
> The purpose of using a static variable is to cache the pixmap, but the original code does something wrong. I think `std::unique_ptr` may be useful in this case.

Thanks!
I figured there was an attempt to; if there's a performance concern, I would change though the pixmaps to be both created on `TrackOperationsWidget` construction and then a pointer changed to point the correct one, not through paintEvent, but based on what actually changes the state (so there's less unnecessary logic running via paintEvent).

But my reasoning was that embed::getIconPixmap already internally caches the Pixmap via QPixmapCache and that although it's queried in a paintEvent, it seemed an operation light enough to be done in this way.

Let me know what you think!.

PhysSong:
That makes sense.

Smjert:
As requested by https://github.com/LMMS/lmms/issues/5592, my PR is not stale, just waiting for approvals.

russiankumar:
Okay, marked as ready

Issue #3740, opened 2017-07-31
Better transport/playback/playhead behavior


musikBear:
**Edited by @tresf**

* [ ] Global playback/transport #1357
* [ ] Project should remember the last-used playhead behavior per #4192, #4556 
* [ ] Better keyboard shortcut behavior when mixer is in focus #1140 (related: #1488)
* [ ] Better keyboard shorts in general #1144  (related: #1488)
* [ ] Proposal to use `>>` as default playhead behavior #3740
* [ ] Stop and return `|<<` versus "pause" default behavior #3740
* [ ] Link JACK transport with LMMS transport #4412
* [ ] Stop button should stop ALL sounds #1535 
* [ ] Playback should stop at the end of a track #4643
* [ ] Position indicator and line behavior inconsistent between Song-Editor, Piano-Roll, Beat/Bassline and Automation Editor #857, #2123, #3052

See also:
* Better MIDI support: #1472
* Better JACK support: #1467 

--- 

**Original bug report, modified by @tresf for readiblity**

LMMS has 3 play-head behaviours `|<<` ,  `<<` , and `>>`.  Currently `|<<` is selected as default butI think it will be best if it is `>>` that is default.

Really often i start playing a track, then i hear a [part that needs correction] and hit [space]. I should have [clicked pause] but that's a mouse-event, and intuitively space is pressed.  [Unfortunately] the play-head jumps to [the beginning of the track]. If `>>` was default the play-head would [behave like pause by default].  [The `Home` key can still be used to quickly jump to the beginning of the track.]

This is why i believe that `>>` is the best option for default behaviour, it is simply inducing fewer **arghh** situations :)

tresf:
I too find this default behavior intolerable when editing a track.

Spekular:
Isn't the real solution here to add a keyboard shortcut for pause? If the
stop button doesn't move the cursor back to start there's no real reason to
have it there instead of just having a pause button.

zonkmachine:
> Isn't the real solution here to add a keyboard shortcut for pause?

Yes. There was a PR/patch submitted for this some six months ago. I'll do some digging.

zonkmachine:
I remembered it wrong. It was this one https://github.com/teknopaul/lmms/commit/1f9a610bbf29a1f963655b0f6a467f841db00e55
No pause key in there but some improvements nonetheless.

Umcaruje:
[...]

Issue #4192, opened 2018-02-25
Saving condition of playback management button


NETMANSKY:
Condition of playback management button in song editor is not saving in project. Every time after launch of project need to press this button to setup playback as it should.

musikBear:
"playback management button"(?)
I bet you mean |<< / << / >> aka playhead-behaviour  
Then this is a sort of a *doublet* of 'mine' #3740

NETMANSKY:
@musikBear Yes, I'm sure you're right, but plus its state is not saved in project after closing. It's not just uncomfortable but little horrorable. I didn't know how this button mean in English and wrote as it may be translated in Russian ;-) But now I know how it's naming: поведение при проигрывании :) 👍

NETMANSKY:
But trouble is not until solved.

NETMANSKY:
This problem has not been solved in RC7 so far.

tresf:
Consolidated into #3740.

PR #5591, opened 2020-07-20
Set default behaviour of playhead to << instead of |<<

russiankumar:
Solves part of issue #3740. Title is self-explanatory, and I also fixed some wrong grammar in the tooltips. Minimal changes, and I did as decided in the issue related to this topic (although pause is now more convinient with a Shift+Pause hotkey).

### DIFF
--- a/src/gui/TimeLineWidget.cpp
+++ b/src/gui/TimeLineWidget.cpp
@@ -142,7 +142,7 @@ void TimeLineWidget::addToolButtons( QToolBar * _tool_bar )
 					SLOT( toggleBehaviourAtStop( int ) ) );
 	connect( this, SIGNAL( loadBehaviourAtStop( int ) ), behaviourAtStop,
 					SLOT( changeState( int ) ) );
-	behaviourAtStop->changeState( 1 );
+	behaviourAtStop->changeState( BackToStart );
 
 	_tool_bar->addWidget( autoScroll );
 	_tool_bar->addWidget( loopPoints );
--- a/src/gui/TimeLineWidget.cpp
+++ b/src/gui/TimeLineWidget.cpp
@@ -130,7 +130,7 @@ void TimeLineWidget::addToolButtons( QToolBar * _tool_bar )
 
 	NStateButton * behaviourAtStop = new NStateButton( _tool_bar );
 	behaviourAtStop->addState( embed::getIconPixmap( "back_to_zero" ),
-					tr( "After stopping go back to begin" )
+					tr( "After stopping go back to beginning" )
 									);
 	behaviourAtStop->addState( embed::getIconPixmap( "back_to_start" ),
 					tr( "After stopping go back to "
@@ -140,6 +140,7 @@ void TimeLineWidget::addToolButtons( QToolBar * _tool_bar )
 					tr( "After stopping keep position" ) );
 	connect( behaviourAtStop, SIGNAL( changedState( int ) ), this,
 					SLOT( toggleBehaviourAtStop( int ) ) );
+	behaviourAtStop->changeState( 2 );
 
 	_tool_bar->addWidget( autoScroll );
 	_tool_bar->addWidget( loopPoints );
--- a/src/gui/TimeLineWidget.cpp
+++ b/src/gui/TimeLineWidget.cpp
@@ -140,9 +140,9 @@ void TimeLineWidget::addToolButtons( QToolBar * _tool_bar )
 					tr( "After stopping keep position" ) );
 	connect( behaviourAtStop, SIGNAL( changedState( int ) ), this,
 					SLOT( toggleBehaviourAtStop( int ) ) );
+	connect( this, SIGNAL( loadBehaviourAtStop( int ) ), behaviourAtStop,
+					SLOT( changeState( int ) ) );
 	behaviourAtStop->changeState( 1 );
-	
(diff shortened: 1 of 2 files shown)


Spekular:
I'm still opposed to this, as I mentioned in the linked issue.

"If the stop button doesn't move the cursor back to start there's no real
reason to have it there instead of just having a pause button".

There's already a PR in progress for a pause shortcut, I believe. That
means users who prefer this behavior can simply use the pause shortcut.

russiankumar:
> There's already a PR in progress for a pause shortcut, I believe.

It's been merged (#5554). I find myself using << a lot more, so what say we use that instead (tbh |<< is kinda useless unless you're finalising the song)

DouglasDGI:
Huh.  People seem to be debating between `>>` and `|<<`, but I honestly really think `<<` should be the default behavior... the one where the cursor moves back to where playback started from.

Veratil:
> Huh. People seem to be debating between `>>` and `|<<`, but I honestly really think `<<` should be the default behavior... the one where the cursor moves back to where playback started from.

> I find myself using << a lot more, so what say we use that instead (tbh |<< is kinda useless unless you're finalising the song)

> I'm still opposed to this, as I mentioned in the linked issue. "If the stop button doesn't move the cursor back to start there's no real reason to have it there instead of just having a pause button".

So the solution to all this is a user defined default behavior. Right?

DouglasDGI:
We could simply save this value in project files.  Then people could use `File -> Set as default template` if they want to change their default.

DouglasDGI:
Seems the saving of this value is asked for in #3740 (and #4192).

russiankumar:
> I honestly really think << should be the default behavior...

+1 to this.

> We could simply save this value in project files.

Topic of age-old discussion. Can NStateButton states be saved? If so that's an easy thing to do probably.

russiankumar:
> So the solution to all this is a user defined default behavior.

That comes after customization; this PR here is related to the options set when you open LMMS for the first time (and later on as well if the option isn't customised)

DouglasDGI:
If you think the default behavior should be `<<`, why does this PR set it to `>>`?

russiankumar:
AFAIK that was the final decision in the issue this is meant to close, that's why I made it so.

DouglasDGI:
I asked about this on the LMMS Discord via Strawpoll: https://www.strawpoll.me/20632234

After leaving it for about an hour, we got nine votes.  One vote was for `|<<`, two votes were for `>>`, and six votes were for `<<`.  It looks like the general consensus favors the `<<` behavior.

russiankumar:
Well, I guess it's clear what needs to be changed then.

russiankumar:
Accidentally kept the same commit name, but big deal I guess. Default behaviour changed to `<<`.

russiankumar:
Just saying, saving may be beyond the scope of this PR, but given the overhaul coming in maybe two weeks it would be nice to get a saving feature merged too in this PR. Shall I work on that or not?

russiankumar:
Pushed it anyways xD

Storing is accomplished by storing a pointer to the button which is used to set/get state at load/save time. Not the best implementation, so if there's a better idea, let me know.

Spekular:
Can't you use the existing behaviourAtStop() function (defined in TimelineWidget.h) instead of adding a duplicate button?

russiankumar:
That's okay for reading, but when it comes to writing, updating `m_behaviourAtStop` doesn't actually change anything. That's why I needed a pointer to the actual button itself which will consequently update `m_behaviourAtStop`.

Spekular:
Fair point, but I think the solution to that is to add a setter method as well. That should be a smaller change, and avoids exposing all of the button's other functions.

russiankumar:
> add a setter method

Alright, probably not a huge thing. Will have to figure out how to but I'll probably get this done.

russiankumar:
@Spekular sorry for the delay, updated to use a signal-slot connection instead of a pointer.

src/gui/TimeLineWidget.cpp
```diff
@@ -140,9 +140,9 @@ void TimeLineWidget::addToolButtons( QToolBar * _tool_bar )
 					tr( "After stopping keep position" ) );
 	connect( behaviourAtStop, SIGNAL( changedState( int ) ), this,
 					SLOT( toggleBehaviourAtStop( int ) ) );
+	connect( this, SIGNAL( loadBehaviourAtStop( int ) ), behaviourAtStop,
+					SLOT( changeState( int ) ) );
 	behaviourAtStop->changeState( 1 );
```
Spekular, line 145: Please change this to use the the enum name (BackToZero, BackToStart, or KeepStopPosition) rather than just an int.

Spekular:
> @Spekular sorry for the delay, updated to use a signal-slot connection instead of a pointer.

Looks good. Left one review comment, once that's fixed and the PR title is updated to be accurate I'm willing to merge :)

src/gui/TimeLineWidget.cpp
russiankumar: I assume you meant `behaviourAtStop->changeState( 1 );`? I've changed that now.

Passing an int in those functions is the same as the line above, which wasn't written by me, and I left it as int for the sake of consistency.

Spekular: Ah yeah, clicked the wrong line haha. Thanks for fixing it!

PR #5594, opened 2020-07-21
Fix the outer border for the instrument track

### LINKED ISSUE #5593: Outer border for melody clips is drawn one pixel too much to the right and bottom
The melody pattern appears a little bit off (on the right side) compared to the rest.

image

_Originally posted by @russiankumar in https://github.com/LMMS/lmms/pull/5573#issuecomment-657010886_

Creating this issue to keep track of this bug.

claell:
Closes #5593 

Before:
Screenshot_20200721_162618

After:
Screenshot_20200721_162403

### DIFF
--- a/src/tracks/InstrumentTrack.cpp
+++ b/src/tracks/InstrumentTrack.cpp
@@ -22,6 +22,7 @@
  * Boston, MA 02110-1301 USA.
  *
  */
+#include "InstrumentTrack.h"
 
 #include <QDir>
 #include <QQueue>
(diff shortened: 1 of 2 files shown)


claell:
Additionally, I moved a line to make the relevant header include visible more prominently.

russiankumar:
Is it meant to bulge out when selected? Something tells me leaving it inside would be more neat. Just my opinion though.

Spekular:
> Is it meant to bulge out when selected? Something tells me leaving it inside would be more neat. Just my opinion though.

It doesn't actually expand, it's just that the outer stroke is drawn lighter. I double checked that it's consistent with the automation clip's behavior too.

PR #5601, opened 2020-07-29
Improve the context menu actions when multiple TCOs are selected

### LINKED ISSUE #1212: Song editor group operations
mockup

While more than 1 clip is selected, a special menu item would indicate that you're about to do something on a group of clips. Opening B+B editor or Piano roll editor or Automation editor is unapplicable here, so these items should be hidden (as the mockup shows).

Now you can delete, mute, copy, cut, paste (replace with clipboard contents), change color or rename all at once.

Pasting a group of clips would need to know what tracks these clips belong to (just like with notes in piano roll).
And just like with Piano roll, I think the pasted items should start at the playhead, so the user has precise control where to paste stuff.

Other possibility is special "duplicate" function (instead of copy+paste).
Duplicate could create duplicates on the tracks the clips are already on, offseting them by teh selection length.

Also just like in the Piano roll, the Ctrl+drag should deploy rectangular selection IMHO. It's not assigned to anything yet, and keeping this consistent among different places is very important.

So duplicating this:

```
 xxx 
    x
```

Would create this:

```
 xxx xxx
    x   x
```

That'd be very useful for duplicating song parts. A hotkey for this would also be great (Ctrl+D?)

Also - extending the length of one clip while others are selected should work just like in Piano roll - extend all selected by the same length. Same with moving deleting with Delete key and duplicating with Ctrl+drag when the drag is started over a clip (unlike the selecting drag that starts over an empty space).

IanCaio:
Right clicking a TCO on the Song Editor, we have access to actions like _Cut_, _Copy_, _Paste_, _Remove_ and _Mute_. When the user has multiple TCOs selected, it's expected that the actions will affect all TCOs, but currently they only affect the TCO that was right-clicked on. This draft starts working on a solution that will enable those actions to affect all selected TCOs.

In its current state, _Remove_ and _Mute_ are fully functional (they will affect the individual TCO if there are no TCOs selected, or all selected TCOs if there are). _Cut_ and _Copy_ are partially functional (they will create a Datafile with the TCO information and copy it to the Clipboard when multiple TCOs are selected, or do the _Cut_/_Copy_ the previous way if no TCOs are selected). _Paste_ is only functional for single TCOs (when multiple TCOs were previously copied/cut, it will print the copied data in the terminal using `qWarning()` and when a single TCO was copied/cut it will paste normally).

Reason I opened a draft is that it will be easier to allow others to test what is written so far with the LMMSBot builds and also because it will make it more visible so others can share their ideas on how to finish it. I'm still figuring out the _Paste_ operation.

Closes #1212


IanCaio:
Methods used for the Drag&Drop paste were overloaded so they can be used with a QMimeData object instead of a QDropEvent one. That makes it possible for us to reuse those methods for our Context Menu paste actions. There are some details that are not optimal (`canPasteSelection` is called twice, once on every version of the method, if the `QDropEvent` `pasteSelection` method is called), but they work the way they are supposed to and could be fixed in a somewhat simple way.

IanCaio:
**Note to self:**

Apparently `src/tracks/SampleTrack.cpp` overrides the `contextMenuEvent` method. My changes end up not affecting the Sample tracks and even misbehaving. That file will also need to be changed appropriately.

IanCaio:
The basic functionality is implemented, all the Context Menu actions (Remove, Cut, Copy, Paste and Mute) will now affect all the TCOs selected. The Paste action follow the same rules as the Drag&Drop copy/paste:
- TCO being pasted to has to be of the same type of the TCO from where the context menu was opened
- All copied tracks need to land in a track of the same type
- You can't paste on the same bar where you copied

The current state of the PR is still a draft because there are probably things that should be changed to make the code better (one example is the issue with `canPasteSelection` being called twice on the `pasteSelection` that receives a `QDropEvent`). I also wish to create a context menu on the TrackContentWidget so the user can paste anywhere in the Track, instead of just over another Track. I can make the PR "Ready to review" though if the Devs would like to give input already on how things could be improved, just let me know!

IanCaio:
Marking the PR as Ready for review. All proposed functionality was implemented!

In my opinion, we should soon move away from using the `Clipboard` class and use the system's clipboard instead for all TCO copy/paste operations (and even others that currently use `Clipboard.h`, like `PianoRoll.cpp`). My PR uses the system's clipboard just like Drag&Drop copy/paste does. However, individual TCO copy/paste, which was already implemented, still uses the `Clipboard` class. It's possible to swift everything to the use the regular clipboard and I think we should do it soon, but IMHO in a different PR. Not only it's outside the scope of this one, but it would make it easier to review.

The functionality added should help a lot with people's workflow, since copying/pasting during production is a very constant thing and currently besides the Drag&Drop those operations are very limited.

russiankumar:
Alright, so I tested this, here's what I have to say:
- This works more or less as expected when using the menu
- While using shortucts (eg. for muting - Ctrl+middle click) it still affects only the targeted TCO and not all the selected TCOs
- Right-click menu for paste works a bit weird. Say I want to paste somewhere but I hit the wrong box by mistake. In that case I would move my mouse and right click on the correct box. However I have to right click twice for this as the first right click only disables the previous menu. _(On later testing, this is actually observed with every menu, so this is probably a fix beyond the scope of the PR)_.
- Copy-paste works with different quantization settings (as expected)

Hope this helps!

IanCaio:
Thanks for testing @russiankumar !

> - While using shortucts (eg. for muting - Ctrl+middle click) it still affects only the targeted TCO and not all the selected TCOs

True, I only changed the context menu actions, so the shortcuts were not affected. But I think this is an easy fix. I can change `mousePressEvent` to instead of calling `remove()` call `contextMenuAction( Remove )` for example. I personally think this is better than the other alternative I thought of, which would be moving the multiple/single remove logic to the `remove()` method, since I think that one should be reserved for individual removes. There's a third alternative which maybe would even make the `contextMenuAction` method much smaller, which is moving the logic from multiple/single remove/copy/cut/paste/mute to a new method which would check for selections and call the individual methods on each one. I would need to think of good naming for those though, maybe `removeSelected`, etc?

> - Right-click menu for paste works a bit weird. Say I want to paste somewhere but I hit the wrong box by mistake. In that case I would move my mouse and right click on the correct box. However I have to right click twice for this as the first right click only disables the previous menu. (On later testing, this is actually observed with every menu, so this is probably a fix beyond the scope of the PR).

Yeah, that's more related to how the Qt context menu works AFAIK, not sure the PR can really do anything to make it less annoying but if someone knows a way just let me know.

As for the pasting works, I understand it's not ideal because people probably expected to be able to paste anywhere on any tracks and everything would land on their appropriate tracks. However that's a deeper issue, because it involves how all other copy/paste operations works at LMMS at the moment: Currently, LMMS allows copying a track TCO and pasting it on another track, or dragging a whole selection of TCOs and dropping them on another track (as long as those rules mentioned before are followed). I even thought of making "Paste" just paste everything on its original track, but then I create a third behavior on the code base and contribute to the confusion. I thought following the guidelines of Drag&Drop was the ideal, making the individual TCO copy/paste the only operation that has a different behavior (which as mentioned could be fixed later by moving away from `Clipboard.cpp`). That's just my take on it, but I'd like to hear everyone's opinions as well!

src/core/Clipboard.cpp
```diff
@@ -35,6 +38,10 @@ void Clipboard::copy( JournallingObject * _obj )
 	QDomElement parent = doc.createElement( "Clipboard" );
 	_obj->saveState( doc, parent );
 	content[_obj->nodeName()] = parent.firstChild().toElement();
+
+	// Clear the QApplication clipboard, so we don't have any conflicts when LMMS has to
+	// decide between the QApplication clipboard and the internal clipboard data
+	QApplication::clipboard()->clear( QClipboard::Clipboard );
```
Spekular, line 44: This feels like an unnecessary/dangerous change. Have you checked all uses of the QApplication clipboard to ensure that this doesn't change copy/paste behavior at all?

src/core/Track.cpp
```diff
@@ -826,22 +830,50 @@ void TrackContentObjectView::mousePressEvent( QMouseEvent * me )
 	{
 		if( me->modifiers() & Qt::ControlModifier )
 		{
-			m_tco->toggleMute();
+			if( so.size() > 0 )
+			{
+				toggleMuteSelection( so );
+			}
```
Spekular, line 836: I think a cleaner implementation than doing all these checks is to use one function for both selections and single clips. Simply pass a collection containing a single clip if you only want to affect that clip.

I'm also not sure that this logic holds. What if I select a bunch of clips, and the press mute on an unselected clip? `QVector so` will still contain the selected objects, so `so.size > 0` should be true, and the selection will be muted instead of the clip that was clicked on.

Spekular:
I think the wording on the context menu items should change to reflect what action will be taken. Especially with the delete action it's good if users are made aware that the entire selection will be affected.

src/core/Track.cpp
IanCaio: Nice catch, you're right! I have to check both if there are selected TCO views and if the one where the event happened is among them. I'll fix that on all actions.

I thought about having `so.size > 0` inside the `toggleMuteSelection` method, but I thought the name wouldn't describe the behavior pretty well anymore, because `toggleMuteSelection` would both mute selections or individual tracks, and I don't know, it could be useful in the future to have those methods that take selections and modify them. What do you think about making a third method that would check if `so.size > 0` and if the event TCO view is selected them and then call either `toggleMuteSelection` or `toggleMute`? What could I call it?

src/core/Clipboard.cpp
IanCaio: I didn't check every use of `QApplication::clipboard()`, but I'm not sure what's the danger in it. I mean, the clipboard is volatile, when you copy you add data to it but when you paste you can't trust the data will still be there (i.e.: if you copy something else before pasting). Maybe I'm missing something?

I had to do that because currently LMMS mixes the `Clipboard` class and the system's clipboard on copy/paste operations. The individual TCO copy uses `class Clipboard` but the multiple TCO copy uses the system's clipboard. So it's possible I'll have data on both simultaneously. When pasting I wouldn't be able to know which one should be pasted, so `Clipboard::copy` now clears the system's clipboard as well. Then I can know I should use it because the system's clipboard will be empty. Ideal would be if I cleared the `Clipboard` data as well when I copied multiple TCOs (which I don't currently do).

Another reason I advocate for changing everything to the system's clipboard, things can get confusing mixing both but I tried to come up with the cleanest solution possible, at least until the change happens (if everyone agrees). From a quick search it seems `Clipboard::copy` is only used on `src/core/Track.cpp` anyways, but other files uses `Clipboard::mimeType()` too.

IanCaio:
> I think the wording on the context menu items should change to reflect what action will be taken. Especially with the delete action it's good if users are made aware that the entire selection will be affected.

Sure, could be done! Personally I think it's expected that "Delete" will remove everything selected. I understand some long time users could have gotten used to the current behavior though and since the `Undo` system is not flawless right now this could lead to headaches, so I'm open minded about changing it. Any ideas on what should be written instead? `Delete TCO/Selection` maybe?

src/core/Track.cpp
IanCaio: I fixed the bug and added the conditionals inside the `removeSelection()`,`copySelection()`,`cutSelection()`,`toggleMuteSelection()` methods. Still unsure about the naming, but had no better ideas until now :thinking:

IanCaio: I fixed the bug and added the conditionals inside the `removeSelection()`,`copySelection()`,`cutSelection()` and `toggleMuteSelection()` methods. Still unsure about the naming, but had no better ideas until now :thinking:

Spekular:
> Any ideas on what should be written instead?

I think "[action] selection" is great for selections. For single items I think we can probably leave the text as is. If you want to change it I'd prefer "[action] clip" over [action] TCO" as TCO is kind of a meaningless term to users (or should be, at least).

> I [...] added the conditionals inside the [action] methods.

Ah, that's not what I was proposing 😅. I meant more like this: 

```cpp
void TrackContentObjectView::mousePressEvent( QMouseEvent * me )
{
	//Get a list of selected selectableObjects
	QVector<selectableObject*> sos = gui->songEditor()->m_editor->selectedObjects();

	//Convert to a list of selected TCOVs
	QVector<TrackContentObjectView*> selection;
	selection.reserve(sos.size());
	for (auto so: sos) {
		TrackContentObjectView* tcov = dynamic_cast<TrackContentObjectView*>(*so);
		if (tcov != nullptr) { selection.append(tcov); }
	}

	//If we clicked part of the selection, affect all selected clips. Otherwise affect the clip we clicked
	QVector<TrackContentObjectView *> active = selection.contains(this) ? selection: QVector(this);

	//Now inside the switch we call
	removeSelection(active);
	togleMute(active);
	//etc...
}

The removeSelection, toggleMute and so on look a bit like this
...
void TrackContentObjectView::removeSelection(QVector<TrackContentObjectView *> tcovs)
{
	for (auto tcov: tcovs) { if (tcov != null) { tocv->remove(); } }
}
```

IanCaio:
@Spekular I updated the code to use the logic structure you suggested. I changed the name of the methods to `removeActive`, `copyActive`, `cutActive`, `toggleMuteActive`, which I think better describe the behavior now. The pasting method is still called `pasteSelection` (because that one is only called when we have a selection in the clipboard, otherwise `paste` is called).

The labels on the context menu were also updated so they change if the actions will affect selections. I also made `removeActive` and `toggleMuteActive` static as we discussed, so they can be called elsewhere in the code if future features require those actions to be called from outside a TCO view.

src/core/Track.cpp
```diff
@@ -1125,25 +1165,170 @@ void TrackContentObjectView::contextMenuEvent( QContextMenuEvent * cme )
...
+
+			// If we have TCO data on the clipboard paste it. If not, do our regular TCO paste.
+			if( QApplication::clipboard()->mimeData( QClipboard::Clipboard )->hasFormat( StringPairDrag::mimeType() ) )
+			{
+				pasteSelection();
+			}
+			else
+			{
+				getTrackContentObject()->paste();
+			}
```
IanCaio, line 1242: What are your opinions on this logic staying here? Should I move it to the `pasteSelection` method and change its name to something like `pasteContent`? Or maybe keep it that way and just move everything to the `paste` method once we (possibly) move away from using `Clipboard::copy`?

IanCaio:
Renamed `pasteSelection` to just `paste`, since I realized the TrackContentObjectView class doesn't have a paste method, just the TrackContentObject class does, hence it doesn't conflict with anything and is an appropriate name for the action.

Moved the logic for deciding between the multiple TCOV paste and individual TCOV paste to inside `TrackContentObjectView::paste()`

src/core/Track.cpp
```diff
@@ -735,8 +735,24 @@ void TrackContentObjectView::paintTextLabel(QString const & text, QPainter & pai
...
+	{
+		TrackContentObjectView *tcov = dynamic_cast<TrackContentObjectView *> ( so );
+		if( tcov != nullptr )
+		{
+			selection.append( tcov );
+		}
+	}
+
+	// If we clicked part of the selection, affect all selected clips. Otherwise affect the clip we clicked
+	QVector<TrackContentObjectView *> active = selection.contains( this ) ? selection : QVector<TrackContentObjectView *>( 1, this );
```
Spekular, line 754: Let's extract this into a new method (`getClickedTCOs`?) since you reuse the same code later.

src/core/Track.cpp
```diff
@@ -1148,6 +1136,26 @@ void TrackContentObjectView::mouseReleaseEvent( QMouseEvent * me )
...
+	{
+		TrackContentObjectView *tcov = dynamic_cast<TrackContentObjectView *> ( so );
+		if( tcov != nullptr )
+		{
+			selection.append( tcov );
+		}
+	}
+
+	// Individual TCO or selection being right-clicked?
+	bool individualTCO = selection.contains( this ) ? false : true;
```
Spekular, line 1157: First of all, you don't need a ternary operator here, you can just use `!selection.contains(this)`.

Secondly, once the duplicate code is extracted, you can write this instead:
`bool individualTCO = getClickedTCOs().size <= 1;`
(Replacing `getClickedTCOs` with whatever we name the new method.)

src/core/Track.cpp
```diff
@@ -1179,177 +1197,134 @@ void TrackContentObjectView::contextMenuEvent( QContextMenuEvent * cme )
 // This method processes the actions from the context menu of the TCO View.
 void TrackContentObjectView::contextMenuAction( ContextMenuAction action )
 {
-	// List of selected TCOs
-	QVector<selectableObject *> so = gui->songEditor()->m_editor->selectedObjects();
+	// Get a list of selected selectableObjects
+	QVector<selectableObject *> sos = gui->songEditor()->m_editor->selectedObjects();
```
Spekular, line 1201: Another place where the newly extracted method should be used.

src/tracks/SampleTrack.cpp
```diff
@@ -343,6 +343,26 @@ void SampleTCOView::updateSample()
 
 void SampleTCOView::contextMenuEvent( QContextMenuEvent * _cme )
 {
+	// Depending on whether we right-clicked a selection or an individual TCO we will have
```
Spekular, line 346: another extract

include/Track.h
```diff
@@ -250,12 +250,16 @@ class TrackContentObjectView : public selectableObject, public ModelView
 	bool needsUpdate();
 	void setNeedsUpdate( bool b );
 
-	// Methods to remove, copy, cut, paste and mute selections of TCO views
-	void removeSelection( QVector<selectableObject *> so );
-	void copySelection( QVector<selectableObject *> so );
-	void cutSelection( QVector<selectableObject *> so );
-	void pasteSelection(); // TODO: Rethink naming?
-	void toggleMuteSelection( QVector<selectableObject *> so );
+	// Methods to remove, copy, cut, paste and mute a QVector of TCO views
```
Spekular, line 253: I think you should drop the `Active` from these names. Technically you can pass whatever QVector of TCOs you want and the function will copy/cut/remove/mute them, so I think it's more accurate that way.

Spekular (changes requested):
Renaming the functions isn't strictly necessary (though I feel pretty strongly that the "Active" is unnecessary/inaccurate) but the repeated code definitely needs to be extracted into a new method.

src/core/Track.cpp
IanCaio: Good idea! I updated the code to use this new method, which I'm calling `getClickedTCOs` as suggested for now

include/Track.h
IanCaio: I also removed the `Active` from the method names, so they are now overloads of the individual TCOs methods. At first I thought they should be differentiated by name to make it less confusing to have the same name as other methods that works differently, but I agree the `QVector` argument should be enough for someone reading the code to understand what they are there for!

Spekular (approved):

src/core/Track.cpp
```diff
@@ -734,6 +735,9 @@ void TrackContentObjectView::paintTextLabel(QString const & text, QPainter & pai
  */
 void TrackContentObjectView::mousePressEvent( QMouseEvent * me )
 {
+	QVector<TrackContentObjectView *> active = getClickedTCOs();
```
Veratil, line 738: I wish there were a way to only use this in the `else if` blocks since this adds overhead to the left click event as well.

src/tracks/SampleTrack.cpp
```diff
@@ -352,20 +356,30 @@ void SampleTCOView::contextMenuEvent( QContextMenuEvent * _cme )
 	if( fixedTCOs() == false )
 	{
 		contextMenu.addAction( embed::getIconPixmap( "cancel" ),
-					tr( "Delete (middle mousebutton)" ),
-						this, SLOT( remove() ) );
+					tr( individualTCO
+						? "Delete (middle mousebutton)"
+						: "Delete selection (middle mousebutton)" ),
+						[this](){ contextMenuAction( Remove ); } );
```
Veratil, line 362: This and all the other lambdas are indented as if they are part of the `tr()` function. This is confusing and should be de-dented.

src/tracks/SampleTrack.cpp
```diff
@@ -352,20 +356,30 @@ void SampleTCOView::contextMenuEvent( QContextMenuEvent * _cme )
 	if( fixedTCOs() == false )
 	{
 		contextMenu.addAction( embed::getIconPixmap( "cancel" ),
-					tr( "Delete (middle mousebutton)" ),
-						this, SLOT( remove() ) );
+					tr( individualTCO
```
Veratil, line 359: Can we fix the indention of all the `addAction` calls? All the indention is crazy with unnecessary tabs.

src/core/Track.cpp
```diff
@@ -1125,25 +1133,176 @@ void TrackContentObjectView::contextMenuEvent( QContextMenuEvent * cme )
 	if( fixedTCOs() == false )
 	{
 		contextMenu.addAction( embed::getIconPixmap( "cancel" ),
-					tr( "Delete (middle mousebutton)" ),
-						this, SLOT( remove() ) );
+					tr( individualTCO
```
Veratil, line 1136: Can we fix the indention of all the addAction calls? All the indention is crazy with unnecessary tabs.

src/core/Track.cpp
```diff
@@ -1125,25 +1133,176 @@ void TrackContentObjectView::contextMenuEvent( QContextMenuEvent * cme )
...
+						? "Delete (middle mousebutton)"
+						: "Delete selection (middle mousebutton)" ),
+						[this](){ contextMenuAction( Remove ); } );
```
Veratil, line 1139: This and all the other lambdas are indented as if they are part of the tr() function. This is confusing and should be de-dented.

src/core/Track.cpp
Spekular: How about this?

`auto active = me->button() == Qt::RightButton ? getClickedTCOs(): QVector();`

IanCaio: I'm not sure how this could be done without repeating code, because both `Qt::LeftButton` and `Qt::MidButton` use the `active` QVector at the moment.

It could be moved to inside the `if` blocks but we would repeat the `getClickedTCOs()` line 4 times. Or doing something similar to what @Spekular suggested and having `auto active = ( me->button() == Qt::LeftButton || Qt::MidButton ) ? QVector() : getClickedTCOs();`. Personally, between the two I think I prefer having `getClickedTCOs` repeated on each `if` block just before the action call, something like:

```
if( me->modifiers() & Qt::ControlModifier )
{
    active = getClickedTCOs();
    remove( active );
}
```

It doesn't look bad (even though the code is repeated a few times), would be more clear and avoids calling the `QVector()` constructor unnecessarily. Should I do it that way?

src/core/Track.cpp
IanCaio: Sure, I tried to keep the indentation as it was originally, but it can be fixed. Should it be this instead?

```
    contextMenu.addAction( embed::getIconPixmap( "cancel" ),
        tr( individualTCO ? "Delete (middle mousebutton) " : "Delete selection (middle mousebutton)" ),
        [this](){ contextMenuAction( remove ); } );
```

src/core/Track.cpp
Spekular: In case you saw the original comment and missed the edit, note that you don't need to check several modifiers. The options are left/mid/right, and active is used in mid and right, so checking for left is enough.

I also doubt the empty QVector() constructor takes much time at all.

If you *do* want to move the declaration into the if statements, please do it only once for right and once for middle click (i.e. two repetitions). There's no reason to do it once per action when all actions  in the branchrequire it.

IanCaio: You're right, I misunderstood the logic when I commented. Your solution is drier and I'm up to using it! The idea of having the `getClickedTCOs` before the action was kind of getting ahead of myself thinking about if later we added actions that didn't use the `active` list but were activated by the Right/Mid buttons. Probably not worth it to get those extra lines thinking of something that might not even get added.

Veratil: I'm cool with Spekular's way. 👍

src/core/Track.cpp
Veratil: Second option is cleaner to me, just pull the `embed` part down as well.
```
contextMenu.addAction(
	embed,
	tr,
	[]()
);
```

IanCaio:
Fixed the indentation on `contextMenuEvent` methods and added blank lines for readability where convenient.

Avoided unnecessary calls to `getClickedTCOs` using @Spekular 's suggestion.

src/core/Track.cpp
```diff
@@ -735,8 +735,11 @@ void TrackContentObjectView::paintTextLabel(QString const & text, QPainter & pai
  */
 void TrackContentObjectView::mousePressEvent( QMouseEvent * me )
 {
-	QVector<TrackContentObjectView *> active = getClickedTCOs();
-	// active will be later used for the remove, copy, cut or toggleMute methods
+	// Right now, active is only used on right/mid clicks actions, so we use a ternary operator
+	// to avoid the overhead of calling getClickedTCOs when it's not used
+	auto active = me->button() == Qt::LeftButton
+							? QVector<TrackContentObjectView *>()
+							: getClickedTCOs();
```
Veratil, line 742: I think two tabs over ternary options would be best, not 7 tabs.

Spekular: It looks like you're using tabs for alignment, which isn't good. If someone displays tabs as 2 spaces, these lines wouldn't start to the right of the above. If they use 8 spaces, they would be *extremely* far out.

If you want to align the cases with the end of the row above, you'd want to tab until the start of the lines line up, and then use spaces from there. This preserves alignment between lines at all tab sizes.

However, I agree with Veratil that we could stick to just indenting one extra tab.

IanCaio: My bad, even to the alignment I wanted I was off. I thought about aligning it with the conditional but I went further. I kept it as an extra tab as suggested.

Spekular:
@Veratil do you have any remaining complaints? If not, shall we merge this?

(Note: this will properly close #1212, IMO, so the relevant checkbox in the Better Workflow issue should be checked when it's merged)

Veratil:
> @Veratil do you have any remaining complaints? If not, shall we merge this?

LGTM 👍

Issue #5592, opened 2020-07-21
Reorganising LMMS, renaming editors and classes and everything in between (cont. #120)


ryuukumar:
**If you are here because your PR is mentioned** then check the instructions at the bottom of the comment.

This is an issue continued on from #120. We're mostly following on from what was decided there, though if you have something to say, feel free to.

These are the draft steps that we will have done one-by-one in order to achieve this massive task:

- [x] Freeze code on December 5th, 2020
- [x] Clear PR backlog by January 25th, 2021
  - [x] Ping devs of older PRs via the table in this issue
  - [x] Review and test PRs to be merged before reorganization
    - [x] #5868 MIDI Range
    - [x] #5522 Microtuner
    - [x] #5657 Play automation pattern with controller
    - [x] #5846 Carla Automation
    - [x] #5826 Improved Oscillators
    - [x] #5735 Local paths and project bundles
    - [x] #5873 Update flanger effect
    - [x] #5923 Automation fine tuning with double click
- [x] Format code...
  - [x] with format PRs (#5791, _[request to not push any more such PRs](https://github.com/LMMS/lmms/issues/5592#issuecomment-740187644)_)
  - [x] with dos2unix (#6198, must be done prior to clang-format)
  - [x] with "include-what-you-use" (#5860)
  - [x] with Clang + CI (#4690, [comment](https://github.com/LMMS/lmms/issues/5592#issuecomment-738513836), latest approach at #6323) - It has been agreed on not doing an automatic format, but leave the decision to the authors of individual PRs
  - [x] However, we do agreed on a few special clang-tidy fixes:
    - [x] modernize-redundant-void-arg (https://clang.llvm.org/extra/clang-tidy/checks/modernize/redundant-void-arg.html)
    - [x] modernize-use-bool-literals (https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-bool-literals.html)
    - [x] modernize-use-emplace (https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-emplace.html)
    - [x] modernize-use-equals-default (#6450) (https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-equals-default.html)
    - [x] ~~modernize-use-equals-delete (https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-equals-delete.html)~~ no occurrences in our code
[...]

PR #5605, opened 2020-07-30
Remove excess parentheses from ternary operations

husamalhomsi:
Thanks to the very low precedence of the ternary operator and its right-to-left associativity.

### DIFF
--- a/include/ComboBoxModel.h
+++ b/include/ComboBoxModel.h
@@ -58,7 +58,7 @@ public:
 
 	QString currentText() const
 	{
-		return ( size() > 0 && value() < size() ) ? m_items[value()].first : QString();
+		return size() > 0 && value() < size() ? m_items[value()].first : QString();
 	}
 
 	const PixmapLoader* currentData() const
--- a/include/EnvelopeAndLfoParameters.h
+++ b/include/EnvelopeAndLfoParameters.h
@@ -73,7 +73,7 @@ public:
 
 	static inline float expKnobVal( float _val )
 	{
-		return ( ( _val < 0 ) ? -_val : _val ) * _val;
+		return ( _val < 0 ? -_val : _val ) * _val;
 	}
 
 	static LfoInstances * instances()
--- a/include/LinkedModelGroups.h
+++ b/include/LinkedModelGroups.h
@@ -116,7 +116,7 @@ public:
 	AutomatableModel* getModel(const std::string& s)
 	{
 		auto itr = m_models.find(s);
-		return (itr == m_models.end()) ? nullptr : itr->second.m_model;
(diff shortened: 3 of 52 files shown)


Veratil:
I feel like this is something that can/will be handled with #4690?

Spekular:
What is the benefit to this? Doesn't the huge number of lines with the same format* imply that people prefer the parenthases to be there?

*(condition) ? true val: false val

IanCaio:
You definitely put some work on this PR by hunting all ternary operators in the code base, but the code style guideline is kind of open when it comes to the use of parenthesis on the ternary operator. So it's something to be discussed and decided, whichever is thought to be the best way.

Personally, I think the parenthesis looks better when the conditional is an expression a little bit bigger than usual. When it's a small expression, no parenthesis look fine to me, maybe even better. So if I was to create a rule for that I'd probably use the amount of operators and size of the evaluation expression as a factor to decide for/against using the parenthesis. Though there's already a rule that says if the expression is big, multiple lines should be used. Maybe if that rule was more strictly followed there would be no need for the parenthesis (that would require quite some more changes on the PR though, for dividing the ternary operators that have bigger expressions in multiple lines).

Anyways, think more voices should be heard on the matter. As for me, I would vote for the PR if the ternary operators with larger evaluation expressions were also divided in multiple lines for the sake of readability. I would vote against if they are kept as one-liners, because in that case I'd prefer the parenthesis to be there.

Sawuare:
@veratil No clang-format or clang-tidy checks affect parentheses in ternary expressions.

Sawuare:
@Spekular Actually, ()-less expressions outnumber those with ().

DomClark:
I don't see why ternary operators either all must have parentheses, or all must not. There's no such requirement in our coding conventions currently, and I would be against adding one - we shouldn't restrict style so much that there's only a single way to format a piece of code, as that removes the flexibility to choose what is most readable in any particular situation. We could equally well go around and remove parentheses from around `&&` whenever it's next to `||`, since it has precedence there, but I imagine this would seriously reduce readability.

As for whether this gets merged or not, I'm indifferent. It doesn't make anything worse, but it doesn't really make anything better either.

Sawuare:
I think excess parentheses obfuscate code, but I realize this is me trying to enforce my personal preference...
If this PR is deemed unneeded/unwanted, I will close it.

Veratil:
> I think excess parentheses obfuscate code

It *can*, but doesn't always. Sometimes it helps readability. There certainly are some areas where we can remove some of those parenthesis, but a global removal is probably not the best idea. I think after we reorg with #5592 and merge the clang-tidy/format #4690 this can be readdressed for sure. 👍

Until then, we can close this one out.

PR #5616, opened 2020-08-08
Enable fading of samples in the Sample Track

thmueller64:
This PR introduces a way to conveniently apply fades to samples like in other DAWs. This is a work in progress and if there is interest in the feature, I will continue to work on this to get it ready to use. I can also think of some possible extensions:

- [ ] Crossfading

- [ ] More flexible fading function

sample_fading_demo


RoxasKH:
Seems great.

I feel like there are some graphic glitches on the left, also imo it needs a solid line in addition to the lightup thing imo.

Like in bitwig

immagine

Or reaper, ecc.

thmueller64:
True, there are glitches. You can also see those at the top. I'd have to move the lines a slight bit inside the widget I suppose. The widget should not be considered final, though.

BaraMGB:
Big up!

src/core/Track.cpp
```diff
@@ -708,7 +708,6 @@ void TrackContentObjectView::paintTextLabel(QString const & text, QPainter & pai
 		elidedPatternName = text.trimmed();
 	}
 
-	painter.fillRect(QRect(0, 0, width(), fontMetrics.height() + 2 * textTop), textBackgroundColor());
```
Spekular, line 711: I disagree with this change. We need this to ensure text stays legible, especially since we allow clips to be colored. I think you should either let the rectangle display on top of the fade curve, or adjust the fade curve's height so that it doesn't share any space with the rectangle.

russiankumar:
Love this feature. I would like to request one thing though, and that is to make the color darker instead of lighter where the curve is cutting off the clip. It makes it seem like the cut off volume is highlighted (thus in focus), but it would make more sense to have the active area be the highlighted/in focus one (thus the non-active area would be darker).

RoxasKH:
> 
> 
> Love this feature. I would like to request one thing though, and that is to make the color darker instead of lighter where the curve is cutting off the clip. It makes it seem like the cut off volume is highlighted (thus in focus), but it would make more sense to have the active area be the highlighted/in focus one (thus the non-active area would be darker).

I completely agree. Every other DAW, darken the color instead of brightening. For example bitwig, you can see it in the image i posted before

src/core/Track.cpp
thmueller64: I changed it now. With the text, it looks a bit strange when not zoomed in. Maybe the fade handles should only be displayed when zoomed in close enough?

Spekular: I think it looks great now while zoomed in! I'll have to test it out before I can give an opinion on different zoom levels, I'll try to do that tomorrow.

RoxasKH:
Great job. I'd make it a little less dark, and also would be great if in the future there will be the possibility to choose the curve of the fading. If you watch bitwig one you can get what i mean, we only have 1 possibility of fading, while would be cool to have the possibility to choose between a straight line, or a curve in both direction. FL achieves that using a point in the middle of the line. Moving that point up or down will choose the shape of the line.

immagine

wujekbrezniew:
I tested this PR and when sample is not long enough fading marker cannot be grabbed. Instead sample moves through timeline. 

https://drive.google.com/file/d/1A5LRGgYVgs7lDgJypMpWpCqGGRh0r0zV/view?usp=sharing

thmueller64:
> Great job. I'd make it a little less dark, and also would be great if in the future there will be the possibility to choose the curve of the fading. If you watch bitwig one you can get what i mean, we only have 1 possibility of fading, while would be cool to have the possibility to choose between a straight line, or a curve in both direction. FL achieves that using a point in the middle of the line. Moving that point up or down will choose the shape of the line.
> 
> immagine

I'm working on that. Will most likely be something like a bezier curve.

thmueller64:
> I tested this PR and when sample is not long enough fading marker cannot be grabbed. Instead sample moves through timeline.
> 
> https://drive.google.com/file/d/1A5LRGgYVgs7lDgJypMpWpCqGGRh0r0zV/view?usp=sharing

I cannot open this, but I know what problem you are referring to. The areas for grabbing are 1/32 of the widget width and will thus vanish when the widget is too small. I will fix that.

src/core/Track.cpp
Spekular: I think it looks fine zoomed out, it's easy enough to see the duration of the fades. That being said, the boxes representing the handles seem to be drawn too low, with a regular height clip you have to press above the box in order to adjust the fade.

superpaik:
I find the feature very interesting! Is it finished? I tested it and works but it's not saving the fading in the project file.

thmueller64:
Unfortunately, this feature is far from being finished. The fading positions are not saved in a project file yet.

src/core/SampleBuffer.cpp
```diff
@@ -792,6 +792,11 @@ bool SampleBuffer::play( sampleFrame * _ab, handleState * _state,
 	{
 		_ab[i][0] *= m_amplification;
 		_ab[i][1] *= m_amplification;
+        if (m_fading)
+        {
+            _ab[i][0] *= fadefunc(play_frame+i);
+            _ab[i][1] *= fadefunc(play_frame+i);
+		}
```
IanCaio, line 799:
```suggestion
	if (m_fading)
	{
		_ab[i][0] *= fadefunc(play_frame+i);
		_ab[i][1] *= fadefunc(play_frame+i);
	}
```
Converts spaces to tabs and fixes indentation on last bracket.

src/core/SampleBuffer.cpp
```diff
@@ -957,10 +962,15 @@ void SampleBuffer::visualize( QPainter & _p, const QRect & _dr,
...
 	for( int frame = first; frame < last; frame += fpp )
 	{
+		float fade = 1.0f;
+		if (m_fading)
+		{
+			fade = fadefunc(frame);
+		}
 		l[n] = QPointF( xb + ( (frame - first) * double( w ) / nb_frames ),
-			( yb - ( m_data[frame][0] * y_space * m_amplification ) ) );
+			( yb - ( m_data[frame][0] * y_space * m_amplification * fade) ) );
```
IanCaio, line 971: This will conflict with changes from a recent merge (#5765) but it's a simple fix. Not sure why Github didn't accuse a conflict though.

src/core/SampleBuffer.cpp
```diff
@@ -1447,3 +1457,167 @@ SampleBuffer::handleState::~handleState()
...
+
+
+
+// Fading related functions
+
+float SampleBuffer::fadingVal(f_cnt_t pos, bool is_left)
+{
+    float relativePos = static_cast<float>(pos)/static_cast<float>(m_frames);
+    return (this->*fadefuncPtr)(relativePos, is_left);
+}
```
IanCaio, line 1469:
```suggestion
float SampleBuffer::fadingVal(f_cnt_t pos, bool isLeft)
{
    float relativePos = static_cast<float>(pos)/static_cast<float>(m_frames);
    return (this->*fadefuncPtr)(relativePos, isLeft);
}
```

src/core/SampleBuffer.cpp
```diff
@@ -1447,3 +1457,167 @@ SampleBuffer::handleState::~handleState()
...
+    if (is_left)
+    {
+        amp *= m_leftBezierFade.evaluate(relativePos/m_leftFader);
+    }
+    else
+    {
+        amp *= m_rightBezierFade.evaluate(1-(relativePos-m_rightFader)/(1-m_rightFader));
+    }
+    return amp;
+}
```
IanCaio, line 1511:
```suggestion
float SampleBuffer::bezierFade(float relativePos, bool isLeft)
{
    float amp = 1.0f;
    if (isLeft)
    {
        amp *= m_leftBezierFade.evaluate(relativePos / m_leftFader);
    }
    else
    {
        amp *= m_rightBezierFade.evaluate(1 - (relativePos - m_rightFader) / (1 - m_rightFader));
    }
    return amp;
}
```

camelCase naming and space around the arithmetic operators.

src/core/SampleBuffer.cpp
```diff
@@ -1447,3 +1457,167 @@ SampleBuffer::handleState::~handleState()
...
+
+
+float CubicBezier::cubicBezierLastSegmentP1(float t, float p1, float p2)
+{
+    return 2*p1*t*(1-t) + p2*pow(t, 2);
```
IanCaio, line 1516:
```suggestion
    return 2 * p1 * t * (1 - t) + p2 * pow(t, 2);
```

I'm not meaning to nitpick, tbh on the context of the Bezier formula I even think the way you written it might be more readable. But our code style guidelines says we need spaces around operators. I'm wondering if there are/should be exceptions. This repeats on other lines but I'm not marking them all because I want to hear from others if we should open exceptions where the formula is more readable without spaces on every operator (like here).

src/core/SampleBuffer.cpp
```diff
@@ -1447,3 +1457,167 @@ SampleBuffer::handleState::~handleState()
...
+float CubicBezier::cubicBezierDerivative(float t, float p1, float p2)
+{
+    return 3*(3*pow(t, 2)-4*t+1)*p1 + 3*t*(2-3*t)*p2 + 3*pow(t, 2);
+}
+
+
+float CubicBezier::solveXforTNewton(float x, bool useCache=true)
+{
+    int max_iter = 100;
+    float t, t_up, f, f_prime;
```
IanCaio, line 1542:
```suggestion
    int maxIter = 100;
    float t, tUp, f, fPrime;
```

src/core/SampleBuffer.cpp
```diff
@@ -1447,3 +1457,167 @@ SampleBuffer::handleState::~handleState()
...
+    if (useCache)
+    {
+        t = m_tCache;
+    }
+    else
+    {
+        t = 0.5f;
+    }
+    for (int i=0; i<max_iter; ++i)
```
IanCaio, line 1551:
```suggestion
    for (int i = 0; i < maxIter; ++i)
```

src/core/SampleBuffer.cpp
```diff
@@ -1447,3 +1457,167 @@ SampleBuffer::handleState::~handleState()
...
+}
+
+
+float CubicBezier::solveXforTBisection(float x)
+{
+    float a = 0.0f;
+    float b = 1.0f;
+    float t, f_a, f_t;
+
+    int max_iter = 10;
```
IanCaio, line 1577:
```suggestion
    float t, fA, fT;

    int maxIter = 10;
```

src/core/SampleBuffer.cpp
```diff
@@ -1447,3 +1457,167 @@ SampleBuffer::handleState::~handleState()
...
+    for (int i = 0; i < max_iter; ++i)
+    {
+        t = (a + b)/2;
+        f_t = cubicBezier(t, m_P1.x(), m_P2.x()) - x;
+        f_a = cubicBezier(a, m_P1.x(), m_P2.x()) - x;
+        bool a_sign = (f_a >= 0) ? true: false;
+        bool t_sign = (f_t >= 0) ? true: false;
```
IanCaio, line 1584:
```suggestion
        bool aSign = (f_a >= 0) ? true: false;
        bool tSign = (f_t >= 0) ? true: false;
```

src/core/SampleBuffer.cpp
```diff
@@ -1447,3 +1457,167 @@ SampleBuffer::handleState::~handleState()
...
+float CubicBezier::evaluate(float x)
+{
+    float t = solveXforT(x);
+    return cubicBezier(t, m_P1.y(), m_P2.y());
+}
+
+
+float SampleBuffer::fadefunc(f_cnt_t pos)
+{
+    // This function should be cached in the future
```
IanCaio, line 1608:
```suggestion
    // TODO: This function should be cached in the future
```

Makes it easier to find places where enhancements are needed.

include/SampleTrack.h
```diff
@@ -109,22 +116,52 @@ class SampleTCOView : public TrackContentObjectView
...
+	void leaveEvent(QEvent *e) override;
 
 
 private:
 	SampleTCO * m_tco;
 	QPixmap m_paintPixmap;
+
+	// Members related to fading
+	bool m_leftCorner = false;
+	bool m_rightCorner = false;
```
IanCaio, line 143:
```suggestion
	bool m_moveLeftCorner = false;
	bool m_moveRightCorner = false;
```

As far as I understood from this first review, those are used on the mouse events for dragging the corners right? Maybe call them `m_moveLeftCorner` to make it consistent with the other members like `m_moveLeftP1` for example, also making their function clearer.

IanCaio (changes requested):
I really like this PRs work so far! I gave it an initial review, still not a complete one. There are code style issues to be fixed, specially about indentation (you probably used space-indenting on your editor). Some other smaller ones like variable naming and spacing. As far as the logic goes I still didn't find any issues, but I didn't dig deep enough yet.

When you merge master/rebase, double check one method mentioned in the review (`SampleBuffer::visualize`), since Github isn't accusing a conflict but there were changes to this method recently. I've seen it happen before that one PR overwrites changes from another one by accident, so be watchful with that one.

I'll be coming back to this PR to another more thoughtfully review!

src/core/SampleBuffer.cpp
IanCaio: So bringing it up on the discord chat an alternative has been brought up by @Veratil which can make the polynomial equation more readable while still conforming to our code style:
```
// Term with power of 1
auto e1 = 2 * p1 * t * (1 - t);
// Term with power of 2
auto e2 = p2 * pow(t, 2);
// Polynomial equation
return e1 + e2;
```

Compiler optimization makes sure the code is compiled to the same as if we had no variables holding the equation terms.

Another alternative that is simpler and might be good enough here is:
```
return (p2) * pow(t, 2) + (2 * p1 * (1 - t)) * t;
```
Which just puts the coefficients on parenthesis and leave the independent variable outside. Just some ideas, you can just add the spaces if you think it's good enough.

include/SampleBuffer.h
```diff
@@ -48,6 +48,86 @@ class QRect;
...
+    float m_x;
+    float m_y;
+    float adjustPoint(float val)
+    {
+        /*
+        Constrain the control points to lie in the interval [0.0, 1.0].
+        This ensures that the fade function has a unique value at each sample
+        and that the fade function can not go above 1.0 and below 0.0.
+        */
+        return qMin(qMax(val, 0.0f), 1.0f);
```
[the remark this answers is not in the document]
Veratil: Also should `inline` this function.

thmueller64:
Thanks a lot for the initial review @IanCaio and @Veratil. I'll look into it tomorrow. This PR is far from being finished however, it is just a draft for the basic concept I've put together. I think it won't be possible to get this  PR ready before the refactor deadline as many things are still missing (crossfading, saving the fades in the project, performance issues for very long samples).

Veratil:
> ... many things are still missing (crossfading, saving the fades in the project, performance issues for very long samples).

Add in saving and then we can add in the other stuff? 😁

IanCaio:
> Thanks a lot for the initial review @IanCaio and @Veratil. I'll look into it tomorrow. This PR is far from being finished however, it is just a draft for the basic concept I've put together. I think it won't be possible to get this PR ready before the refactor deadline as many things are still missing (crossfading, saving the fades in the project, performance issues for very long samples).

Not a thing! We can still try to merge it before the refactor begins: there are other things in motion before we can start it, so there's a little time. Even if it doesn't have all the features you want to include, if it's usable and stable we could review, merge it and continue working on it after the refactor to add the things missing.

As for the performance, I'll have to look better into it. The easiest and simplest solution is probably a look-up table for the values of the curve. There are other alternatives that will probably require some more rewriting, i.e.: actually applying the fading to the samplebuffer data instead of multiplying during playtime. That would require us to have a copy of the original frames that are faded for when we drag the handles again. If the performance load isn't huge we can maybe leave this for after the refactor as well.

include/SampleBuffer.h
```diff
@@ -48,6 +48,86 @@ class QRect;
 // may need to be higher - conversely, to optimize, some may work with lower values
 const f_cnt_t MARGIN[] = { 64, 64, 64, 4, 4 };
 
+
+class ControlPoint
```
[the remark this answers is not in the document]
thmueller64: Good point. In the meantime I've noticed that @PhysSong has been working on Bezier curves for automation tracks in #4674. Maybe we can work on the Bezier header together?

src/core/SampleBuffer.cpp
```diff
@@ -792,11 +792,11 @@ bool SampleBuffer::play( sampleFrame * _ab, handleState * _state,
...
-        if (m_fading)
-        {
-            _ab[i][0] *= fadefunc(play_frame+i);
-            _ab[i][1] *= fadefunc(play_frame+i);
-		}
+	if (m_fading)
+	{
+		_ab[i][0] *= fadefunc(play_frame+i);
+		_ab[i][1] *= fadefunc(play_frame+i);
+	}
```
Veratil, line 799: Just needs 1 extra tab.

thmueller64:
> 
> As for the performance, I'll have to look better into it. The easiest and simplest solution is probably a look-up table for the values of the curve. There are other alternatives that will probably require some more rewriting, i.e.: actually applying the fading to the samplebuffer data instead of multiplying during playtime. That would require us to have a copy of the original frames that are faded for when we drag the handles again. If the performance load isn't huge we can maybe leave this for after the refactor as well.

The problem is that we have to find the value of the Bezier curve for a specific sample. The Bezier curves are parametrized with a value `t` that is in [0, 1]. So for every sample `x` we have to find the `t` that corresponds to it. Using Newton's method for finding the roots is very fast, because we can essentially warm start the algorithm with a cached value. When benchmarking this I've noticed that no more than a few iterations are needed. The problem is that i don't know how to constrain the solution to lie in [0, 1], so for now I've set (the very inefficient) bisection as the default for root finding. 

@PhysSong: As you have already worked on Bezier curves, do you know a better solution?

You notice the performance impact when you apply the fades to a whole song loaded as a sample. It is only impacted when moving the fades though. There is no issue while playing.

src/tracks/SampleTrack.cpp
```diff
@@ -606,11 +942,158 @@ void SampleTCOView::paintEvent( QPaintEvent * pe )
...
+			m_tco->m_sampleBuffer->setRightFader(faderPos);
+		}
+		updateSample();
+	}
+	else if ((e->buttons() & Qt::LeftButton) && (m_moveLeftP1 || m_moveLeftP2))
+	{
+		float xpos = e->localPos().x();
+		float ypos = qMin(qMax(e->localPos().y(),
+							   static_cast<qreal>(m_textLabelHeight)),
+						  static_cast<qreal>(this->height()));
```
Veratil, line 990:
```suggestion
		float ypos = qMin(
			qMax(e->localPos().y(), static_cast<qreal>(m_textLabelHeight)),
			static_cast<qreal>(this->height())
		);
```
Also `qreal` by default is a `double` so this will lose precision on assigning to `float`.

src/tracks/SampleTrack.cpp
```diff
@@ -606,11 +942,158 @@ void SampleTCOView::paintEvent( QPaintEvent * pe )
...
+		float leftFaderAbsolutePos = m_tco->m_sampleBuffer->leftFaderPos() * this->width();
+		float Px = xpos/leftFaderAbsolutePos;
```
Veratil, line 992:
```suggestion
		float Px = xpos / leftFaderAbsolutePos;
```

src/tracks/SampleTrack.cpp
```diff
@@ -606,11 +942,158 @@ void SampleTCOView::paintEvent( QPaintEvent * pe )
...
+		float Py = 1 - (ypos-m_textLabelHeight)/(this->height()-m_textLabelHeight);
```
Veratil, line 993:
```suggestion
		float Py = 1 - (ypos - m_textLabelHeight) / (this->height() - m_textLabelHeight);
```

src/tracks/SampleTrack.cpp
```diff
@@ -606,11 +942,158 @@ void SampleTCOView::paintEvent( QPaintEvent * pe )
...
+		}
+		updateSample();
+	}
+	else if ((e->buttons() & Qt::LeftButton) && (m_moveRightP1 || m_moveRightP2))
+	{
+		float xpos = e->localPos().x();
+		float ypos = e->localPos().y();
+		float rightFaderAbsolutePos = m_tco->m_sampleBuffer->rightFaderPos() * this->width();
+		float Px = (xpos - rightFaderAbsolutePos) / (this->width() - rightFaderAbsolutePos);
+		float Py = 1 - (ypos-m_textLabelHeight)/(this->height()-m_textLabelHeight);
```
Veratil, line 1010:
```suggestion
		float Py = 1 - (ypos - m_textLabelHeight) / (this->height() - m_textLabelHeight);
```

src/tracks/SampleTrack.cpp
```diff
@@ -606,11 +942,158 @@ void SampleTCOView::paintEvent( QPaintEvent * pe )
...
+void SampleTCOView::setManualFade()
+{
+	if (m_leftFadeClicked)
+	{
+		m_leftFadeManual = true;
+	}
+	else if (m_rightFadeClicked)
+	{
+		m_rightFadeManual = true;
+	}
```
Veratil, line 1094:
```suggestion
	if (m_leftFadeClicked) { m_leftFadeManual = true; }
	else if (m_rightFadeClicked) { m_rightFadeManual = true; }
```

Veratil (changes requested):
Mostly style review, one or two comments about other issues.

include/SampleBuffer.h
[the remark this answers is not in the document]
IanCaio: I'm not sure if he is actively working on it at the moment, but as long as the Bezier header is done in a way that makes it usable by any other class that need to calculate Bezier curves it should be fine!

include/Track.h
```diff
@@ -288,9 +288,9 @@ public slots:
 	void contextMenuAction( ContextMenuAction action );
 	void dragEnterEvent( QDragEnterEvent * dee ) override;
 	void dropEvent( QDropEvent * de ) override;
-	void leaveEvent( QEvent * e ) override;
+	virtual void leaveEvent( QEvent * e ) override;
 	void mousePressEvent( QMouseEvent * me ) override;
-	void mouseMoveEvent( QMouseEvent * me ) override;
+	virtual void mouseMoveEvent( QMouseEvent * me ) override;
```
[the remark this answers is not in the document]
DomClark: They definitely don't need `virtual`. Whether they should have `virtual` anyway is the subject of a few arguments, but the general consensus, and what we do here, is to omit it.

IanCaio:
> The problem is that we have to find the value of the Bezier curve for a specific sample. The Bezier curves are parametrized with a value `t` that is in [0, 1]. So for every sample `x` we have to find the `t` that corresponds to it. Using Newton's method for finding the roots is very fast, because we can essentially warm start the algorithm with a cached value. When benchmarking this I've noticed that no more than a few iterations are needed. The problem is that i don't know how to constrain the solution to lie in [0, 1], so for now I've set (the very inefficient) bisection as the default for root finding.

I still need to read a little bit more on Bezier curves and on how it's implemented in the code. When you mentioned performance issues I thought those were related to the Bezier curve points being calculated during play time (`SampleBuffer::play`) instead of previously stored in a lookup table. When I get the time I'll finish reading that article you linked about Bezier and get deeper on the logic of the calculations here to see if I can help figuring this one out.

> You notice the performance impact when you apply the fades to a whole song loaded as a sample. It is only impacted when moving the fades though. There is no issue while playing.

Well, it's still an issue but much less critical if the performance is only impacted when setting the fades instead of when playing the samples. If we don't figure it out until the refactor we can merge it as is, documenting the need to optimize it, and figure it out after the refactor (since it won't affect the user as negatively as playtime performance issues).

Veratil (commented):
Other than the unneeded `virtual` keywords as Dom said in `Track.h` for `leaveEvent` and `mouseMoveEvent`, code style looks good. 👍

include/Track.h
```diff
@@ -299,7 +299,7 @@ public slots:
 	}
 
 	float pixelsPerBar();
-
+    int m_textLabelHeight = 0;
```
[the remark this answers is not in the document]
PhysSong:
```suggestion
	int m_textLabelHeight = 0;
```

src/core/SampleBuffer.cpp
PhysSong:
```suggestion
		if (m_fading)
		{
			_ab[i][0] *= fadefunc(play_frame+i);
			_ab[i][1] *= fadefunc(play_frame+i);
		}
```

src/core/SampleBuffer.cpp
```diff
@@ -1447,3 +1457,153 @@ SampleBuffer::handleState::~handleState()
...
+	int maxIter = 10;
+	for (int i = 0; i < maxIter; ++i)
+	{
+		t = (a + b)/2;
+		fT = cubicBezier(t, m_p1.x(), m_p2.x()) - x;
+		fA = cubicBezier(a, m_p1.x(), m_p2.x()) - x;
+		bool aSign = (fA >= 0) ? true: false;
+		bool tSign = (fT >= 0) ? true: false;
+		if (aSign == tSign) { a = t; }
+		else { b = t; }
```
PhysSong, line 1580: If both `m_p1.x()` and `m_p2.x()` are in `[0, 1]`, It can be proved that the cubic polynomial is non-decreasing. Using that, you can reduce one call to `cubicBezier()`:
```suggestion
		if (cubicBezier(t, m_p1.x(), m_p2.x()) < x) { a = t; }
		else { b = t; }
```

PhysSong:
Regarding the Bezier curve: there are several method to calculate/approximate the value.
- Subdividing the Bezier polygon: suggested in https://github.com/LMMS/lmms/issues/2381#issuecomment-143464234
- Linear interpolation method: converges faster than bisection in most cases, but converges slowly when the derivative is very small at the target point
- Solving the cubic polynomial exactly: involving cubic root and might be computationally expensive

Also, I think Newton's method is not good for some cases where the derivative goes to zero.

src/tracks/SampleTrack.cpp
```diff
@@ -321,6 +336,36 @@ void SampleTCO::loadSettings( const QDomElement & _this )
 	if (_this.hasAttribute("sample_rate")) {
 		m_sampleBuffer->setSampleRate(_this.attribute("sample_rate").toInt());
 	}
+
+	if (_this.hasAttribute("leftP1x") && _this.hasAttribute("leftP1y") &&
+	    _this.hasAttribute("leftP2x") && _this.hasAttribute("leftP2y"))
```
Veratil, line 341: I think this looks and reads better.
```suggestion
	if (_this.hasAttribute("leftP1x") && _this.hasAttribute("leftP1y")
	    && _this.hasAttribute("leftP2x") && _this.hasAttribute("leftP2y"))
```

src/tracks/SampleTrack.cpp
```diff
@@ -321,6 +336,36 @@ void SampleTCO::loadSettings( const QDomElement & _this )
...
+	{
+		m_sampleBuffer->setLeftP1(_this.attribute("leftP1x").toFloat(),
+		                          _this.attribute("leftP1y").toFloat());
+		m_sampleBuffer->setLeftP2(_this.attribute("leftP2x").toFloat(),
+		                          _this.attribute("leftP2y").toFloat());
```
Veratil, line 346: Do we want to align with spaces? Or just split parameters out with a single tab? I feel like a single tab is a lot easier to do than holding space for 26 characters.
```suggestion
		m_sampleBuffer->setLeftP1(
			_this.attribute("leftP1x").toFloat(),
			_this.attribute("leftP1y").toFloat()
		);
		m_sampleBuffer->setLeftP2(
			_this.attribute("leftP2x").toFloat(),
			_this.attribute("leftP2y").toFloat()
		);
```

Veratil (commented):
Curious about opinions on these

src/tracks/SampleTrack.cpp
thmueller64: Agreed. Could you please mark the one for "right" too, so i can batch commit?

src/tracks/SampleTrack.cpp
```diff
@@ -321,6 +336,36 @@ void SampleTCO::loadSettings( const QDomElement & _this )
...
+	}
+
+	if (_this.hasAttribute("rightP1x") && _this.hasAttribute("rightP1y") &&
+	    _this.hasAttribute("rightP2x") && _this.hasAttribute("rightP2y"))
```
Veratil, line 350:
```suggestion
	if (_this.hasAttribute("rightP1x") && _this.hasAttribute("rightP1y")
	    && _this.hasAttribute("rightP2x") && _this.hasAttribute("rightP2y"))
```

src/tracks/SampleTrack.cpp
thmueller64: I also think the second one is better here. For function declarations though I think the first one is easier on the eyes.

PhysSong:
@thmueller64 Are you still working on this?

thmueller64:
@PhysSong thanks for the reminder. I have implemented linear interpolation in the meantime which didn't noticeably improve the performance. I found that the real bottleneck lies in redrawing the waveform. When I turn off drawing, moving the fades works seemlessly without any delay. Performant drawing of the audio waveforms like in other DAWs has to be addressed seperately in my opinion.

Veratil:
> I found that the real bottleneck lies in redrawing the waveform. When I turn off drawing, moving the fades works seemlessly without any delay. Performant drawing of the audio waveforms like in other DAWs has to be addressed seperately in my opinion.

Agreed. We also need some comments about where things are done, and not being familiar with that side of the codebase, it took me a while to realize that the drawing was done in `src/tracks/SampleBuffer.cpp` in the `visualize` function.

I think the best way to speed up drawing would be to cache the base waveform, then calculate modified data if a fade touches it. Right now it looks like there's no caching and it's recalculated each time. This is a lot of overhead waste.

IanCaio:
> Agreed. We also need some comments about where things are done, and not being familiar with that side of the codebase, it took me a while to realize that the drawing was done in `src/tracks/SampleBuffer.cpp` in the `visualize` function.

Just a heads up, that's one of the things I think we should fix on the refactor. `SampleBuffer::visualize` go against the decoupling GUI from core paradigm we've been wanting to follow. I think we should figure out a way to remove it from that class and add to another one (not sure if it would be an overkill to have a `SampleBufferView` class)

thmueller64:
I think it has to work even without caching. For example, when you turn the amplification knob in the audio file processor on a sample of like 2min+, you have to update every value with a simple multiplication. On my hardware, the waveform plot also struggles to catch up in time and is very unresponsive.

Veratil:
> I think it has to work even without caching. For example, when you turn the amplification knob in the audio file processor on a sample of like 2min+, you have to update every value with a simple multiplication.

You're right, we definitely need to figure out a better way to speed up the waveform visualization. My cache idea I think is better overall for UI responsiveness when not changing any dependent variables, rather than real-time updating. With many sample buffers and constantly calling the visualize function to calculate the same thing over and over again just makes things slow.

> On my hardware, the waveform plot also struggles to catch up in time and is very unresponsive.

With my cache idea this will still slow down for you unfortunately. We can still look to find some calculation savings though for this loop. Figuring out what is constant during the loop and what can change could speed up some parts.

src/tracks/SampleTrack.cpp
```diff
@@ -652,9 +1088,21 @@ void SampleTCOView::paintEvent( QPaintEvent * pe )
 	p.end();
 
 	painter.drawPixmap( 0, 0, m_paintPixmap );
+	drawFades();
 }
 
 
+void SampleTCOView::enterEvent(QEvent *e)
+{
+	if (e != nullptr) { m_inside = true; }
```
Veratil, line 1097: Can `e` be a `nullptr`? I guess if called manually, but I don't think it will ever be `nullptr` when Qt generates it.

src/tracks/SampleTrack.cpp
```diff
@@ -665,8 +1113,148 @@ void SampleTCOView::reverseSample()
...
+	{
+		m_tco->m_sampleBuffer->setRightP2(Px, Py);
+	}
+	updateSample();
+}
 
 
+void SampleTCOView::mouseMoveEvent(QMouseEvent* e)
+{
+	if ((e->buttons() & Qt::LeftButton) && (m_leftCorner || m_rightCorner))
```
Veratil, line 1176:
```suggestion
	auto leftButtonDown = e->buttons() & Qt::LeftButton;
	if (leftButtonDown && (m_leftCorner || m_rightCorner))
```

and so forth with the others...

src/tracks/SampleTrack.cpp
```diff
@@ -665,8 +1113,148 @@ void SampleTCOView::reverseSample()
...
+	{
+		// This event is only forwarded, when the fades are not currently adjusted, in order
+		// to prevent the SampleTCO from beeing moved around in the timeline when adjusting
+		// the fades.
+		TrackContentObjectView::mouseMoveEvent(e);
+	}
+}
+
+
+void SampleTCOView::setSFade()
```
Veratil, line 1198: I think for these `set*Fade` functions it wouldn't hurt to explain why those constants are used in the `set(Left|Right)P[12]` calls.

include/SampleBuffer.h
```diff
@@ -48,6 +48,85 @@ class QRect;
...
+	{
+		m_x = adjustPoint(x);
+	}
+
+	void setY(float y)
+	{
+		m_y = adjustPoint(y);
+	}
+
+	const float x() const
```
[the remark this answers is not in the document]
thmueller64: Yes, by defining the functions inside the class, they are implicitly inline.

include/SampleBuffer.h
```diff
@@ -48,6 +48,85 @@ class QRect;
...
+private:
+	float cubicBezier(float t, float p1, float p2);
+	float cubicBezierDerivative(float t, float p1, float p2);
+	float cubicBezierLastSegmentP1(float t, float p1, float p2);
+	float cubicBezierLastSegmentP2(float t, float p1, float p2);
+	// Solve with Newton's method: Very fast, but t might lie outside [0.0, 1.0]
+	float solveXforTNewton(float x, bool useCache);
+	// Solve with Bisection method: Slow, but t is guaranteed to lie inside [0.0, 1.0]
+	float solveXforTBisection(float x);
+	float solveXforT(float x)
```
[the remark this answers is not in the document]
thmueller64: Yes, it is inlined already.

include/SampleBuffer.h
```diff
@@ -273,6 +352,75 @@ class LMMS_EXPORT SampleBuffer : public QObject, public sharedObject
...
+	{
+		m_TcoStartTime = startTime;
+	}
+
+	void setTcoFrameLength(float len)
+	{
+		m_TcoFrameLength = len;
+	}
+
+	float leftFaderPos()
```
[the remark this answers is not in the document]
thmueller64: Same as above.

include/SampleTrack.h
thmueller64: Yes they are used for that. I'll rename those.

src/tracks/SampleTrack.cpp
thmueller64: Probably not. I think it is good to check anyway, to be on the safe side.

src/tracks/SampleTrack.cpp
thmueller64: Done.

thmueller64:
@Veratil I did some crude profiling of the visualize function to see where we need to optimize. It turned out that most of the time is spent in drawing the waveform with `drawPolyline`.
Those are the results I got with some simple std::chrono time measurements:

Before for loop: 0,017852
In for loop: 0,078132
After for loop: 0,904016

So we have to figure out a better concept there. Reducing the number of points drawn for example. One standard approach would be to take the maximum of the samples per pixel. This is done in Audacity for example: https://manual.audacityteam.org/man/audacity_waveform.html

include/SampleBuffer.h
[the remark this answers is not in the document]
Veratil: I was never aware of this. TIL!

src/tracks/SampleTrack.cpp
Veratil: Well the reason I bring it up is because the event is called only if raised, so `e` will never be `nullptr`. The only time it will be is if we were to call it manually without.

Veratil:
Good find! We can do that in another PR. 👍

src/core/SampleBuffer.cpp
```diff
@@ -1027,9 +1032,14 @@ void SampleBuffer::visualize(
...
 	{
+		float fade = 1.0f;
+		if (m_fading)
+		{
+			fade = fadefunc(frame);
+		}
 		auto x = xb + ((frame - first) * double(w) / nbFrames);
 		// Partial Y calculation
-		auto py = ySpace * m_amplification;
+		auto py = ySpace * m_amplification * fade;
```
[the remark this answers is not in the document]
thmueller64: I created a separate issue for the whole drawing concept https://github.com/LMMS/lmms/issues/5908 . I hope it is OK if I resolve this for now and defer the discussion to the issue.

LostRobotMusic:
I drastically improved sample drawing speeds in #5927, so this PR should be viable now.

However, the fade shapes provided are incorrect for the purpose of crossfading between uncorrelated audio signals.  Using them will result in a 3 dB dip in volume partway through the fade.  Nearly every DAW gets this wrong and LMMS shouldn't be one of them.  I explain this in a video here: https://youtu.be/-5cB3rec2T0

Taking the square root of any constant-gain crossfade will result in one that is constant-power.  I recommend doing this with the raised cosine fade so there's no slope discontinuity at the highest point in the fade.  I also heavily recommend making this the default shape, as is currently the case in REAPER and should be the case in every DAW.

You can verify it was done correctly by fading between two white noise sources and verifying that there isn't a 3 dB dip in RMS in the middle of the fade.  You can also verify this by ensuring the power (square) of each fade function adds up to a flat line, as shown:
image

regulus79:
Wow this fading feature looks really cool! Are there any plans to finish it?

Issue #5494, opened 2020-05-12
Icons and comboboxes mismatch in arpeggiator in Instrument Editor


qnebra:
### Bug Summary
Misaligned icons and combobox in Arpeggio editor inside Instrument Editor

#### Steps to reproduce
Open Instrument Editor, Go to "Chord stacking & arpeggio" and look down.

#### Expected behavior
Evenly aligned icons and combobox.

#### Actual behavior
Misaligned icons and combobox, which create feeling of amateur project, regarding of it's actual impressive quality.

#### Screenshot

![Adnotacja 2020-05-12 195908](https://user-images.githubusercontent.com/11295229/81728725-304dee00-948b-11ea-9524-fd03b8d1353d.jpg)

#### Affected LMMS versions

1.2.1 and further. I don't have previous versions installed to confirm that other versions than 1.2.1 were affected.

zonkmachine:
Confirmed. The combo boxes are more narrow so their content doesn't fit.

zonkmachine:
![new](https://user-images.githubusercontent.com/6368949/81746279-7e241f80-94a6-11ea-88a7-3358f03bb5df.png)

1.2.1 to the left and some older version to the right. The arpeggiator area has been pushed upward a bit so the components has less vertical space to share.

zonkmachine:
The issue is there already in 1.2.0-RC5

qnebra:
Theoretically, simplest solution will be to increase combobox height to have arrow icon centered in it. But, this was global or local thing?

zonkmachine:
Yes. I think this could be a global thing that just happened to be troublesome in the arpeggiator control section.

musikBear:
ehhhr
is this 
![image](https://user-images.githubusercontent.com/6662066/81850613-83da3d80-9558-11ea-8c3e-714a5e0bdbe8.png)
the subject..
The triangle ..not centred?

... Wont keep me awake ....

qnebra:
> Yes. I think this could be a global thing that just happened to be troublesome in the arpeggiator control section.

Well, nope. There were at least two different comboboxes height in lmms.

Also Bear, I see your comment before deleting. This is kind of details which create overall feeling of quality in software.

zonkmachine:
> Well, nope. There were at least two different comboboxes height in lmms.

Probably because there are two combobox classes being used. QComboBox and our own [Combobox](https://github.com/LMMS/lmms/blob/master/include/ComboBox.h) class.

[...]

PR #5617, opened 2020-08-09
Fix for Icons and comboboxes mismatch in arpeggiator in Instrument Ed…

DigArtRoks:
ComboBox
…itor #5494

(https://github.com/LMMS/lmms/issues/5494)

Introduce a static const int variable for the default height of a ComboBox.
Set this height already in the constructor of the ComboBox object.
Update all modules setting the height of a ComboBox object to make use of the new constant.

### DIFF
--- a/data/themes/classic/style.css
+++ b/data/themes/classic/style.css
@@ -118,7 +118,7 @@ QMenu::indicator:selected {
 	background-color: #747474;
 }
 
-positionLine {
+PositionLine {
 	qproperty-tailGradient: false;
 	qproperty-lineColor: rgb(255, 255, 255);
 }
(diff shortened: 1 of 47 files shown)


qnebra:
Rebase (or something like that) this against latest master, as recently bunch of big patches landed in repository and your PR looks very large at this moment, being in fact quite small.

- How this will behave in HiDPI screens with hardcoded widht? It was my main concern.

DigArtRoks:
I just rebased. I'm new to GitHub, not sure I use the right workflow though.

Concerning your initial remark on HiDPI screens. It should be fine I think. In fact I only replaced the magic value 22 by a named static constant and made sure that if a module is not changing the geometry of the combobox, its height will also be set to 22 by default.

Issue #5615, opened 2020-08-08
LB302 filter doesn't adjust to sample rate correctly


dj-pixus:
When exporting with a sample rate other than 44.1 kHz, LB302's filter frequency goes higher than the set value.

Steps to reproduce:
- Drag LB302 to the song editor and make some pattern.
- Set the filter so that its effect is hearable
- Export the project with oversampling or higher sample rate

Expected behavior:
The sound sounds the same as in the editor.

Actual behavior:
The filter frequency is higher than expected.

Tested on current master and 1.2.2 AppImage.

thmueller64:
There was a similar issue with exports and oversampling that is now fixed on master. Could you try to reproduce this on a master build?

dj-pixus:
Oh yes, master also has the issue. Investigating further, it is not a problem with oversampling, but with rendering at higher sample rates generally.
The real problem appears to be that LB302 filter does not adjust itself to the sampling freq.

thmueller64:
I followed your steps to reproduce and compared the 44.1kHz export to a 96kHz and 192kHz export. The filter frequency appears to be the same for the exports, at least there is no noticeable difference. Could you please share a project file where the problem appears? Maybe it is related to more specific settings.

dj-pixus:
Sure.
[lb302-bug.zip](https://github.com/LMMS/lmms/files/5047156/lb302-bug.zip)

dj-pixus:
Sure.
[lb302-bug.zip](https://github.com/LMMS/lmms/files/5047165/lb302-bug.zip)

thmueller64:
Could you please check out the PR above and see if this fixes the problem? For me the export now sounds the same for different sampling rates. The problem turned out to be related to the 24dB filter.

dj-pixus:
I tested the PR and yes, the problem went away. LB302 works well now, but turns out there might be more similar bugs with other instruments. For example .ds presets behave strangely too, but I believe that's a different issue.
Anyway, the PR is right.

PR #5618, opened 2020-08-09
LB302: Prevent cutoff frequency modification when the sampling rate changes.

thmueller64:
Fixes issue #5615. It turned out that the problem appears when the 24dB Filter is switched on and Env Mod > 0. 

Findings:
The term where I made the change gets normalized a few lines down by `1/(Engine::mixer()->processingSampleRate()/2)`, so the term is most likely the cutoff frequency and should therefore not depend on the sampling rate.
Yet, the calculation of the cutoff frequency involves a term with the current sampling rate times the value of the knob `k`: `(k)*(w)*(k)*(Engine::mixer()->processingSampleRate()/2.f)`. The `w` should be set proportional to `1/Engine::mixer()->processingSampleRate()` by the base class, but the author of the code explicitly set his own static values along with the comment "// do not call base class". When the sampling rate changes during the export, this term also changes and consequently the cutoff frequency.
A simple fix is to calculate the cutoff frequency with 44.1kHz fixed and maybe also check if the cutoff is above the nyquist frequency which might happen at a lower samplerate than 44.1kHz. Cutoffs up to 22.05kHz are sufficient in all cases.

### DIFF
--- a/plugins/lb302/lb302.cpp
+++ b/plugins/lb302/lb302.cpp
@@ -74,7 +74,7 @@
 
 
 //#define engine::mixer()->processingSampleRate() 44100.0f
-
+const float sampleRateCutoff = 44100.0f;
 
 extern "C"
 {
@@ -228,8 +228,11 @@ void lb302Filter3Pole::envRecalc()
 	// e0 is adjusted for Hz and doesn't need ENVINC
 	w = vcf_e0 + vcf_c0;
 	k = (fs->cutoff > 0.975)?0.975:fs->cutoff;
+    // sampleRateCutoff should not be changed to anything dynamic that is outside the
+    // scope of lb302 (like e.g. the mixers sample rate) as this changes the filters cutoff
+    // behavior without any modification to its controls.
 	kfco = 50.f + (k)*((2300.f-1600.f*(fs->envmod))+(w) *
-	                   (700.f+1500.f*(k)+(1500.f+(k)*(44100.f/2.f-6000.f)) *
+	                   (700.f+1500.f*(k)+(1500.f+(k)*(sampleRateCutoff/2.f-6000.f)) *
 	                   (fs->envmod)) );
 	//+iacc*(.3+.7*kfco*kenvmod)*kaccent*kaccurve*2000
 
--- a/plugins/lb302/lb302.cpp
+++ b/plugins/lb302/lb302.cpp
@@ -229,7 +229,7 @@ void lb302Filter3Pole::envRecalc()
 	w = vcf_e0 + vcf_c0;
 	k = (fs->cutoff > 0.975)?0.975:fs->cutoff;
 	kfco = 50.f + (k)*((2300.f-1600.f*(fs->envmod))+(w) *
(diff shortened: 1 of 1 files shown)


SecondFlight:
If this is and really should be a constant `44100.0f`, could it be added as a named constant? It looks like this bug was introduced by mistake in 910daa97 and I can see it happening again, as `44100.0f` looks like it shouldn't be a constant.

In fact, that commit replaces the original author's `LB_HZ` constant. The line affected by this PR is only one of a few lines where the constant was switched with a call to `engine::getMixer()->getSampleRate()`, which has since been superseded by `engine::mixer()->processingSampleRate()`.

plugins/lb302/lb302.cpp
```diff
@@ -229,7 +229,7 @@ void lb302Filter3Pole::envRecalc()
 	w = vcf_e0 + vcf_c0;
 	k = (fs->cutoff > 0.975)?0.975:fs->cutoff;
 	kfco = 50.f + (k)*((2300.f-1600.f*(fs->envmod))+(w) *
-	                   (700.f+1500.f*(k)+(1500.f+(k)*(Engine::mixer()->processingSampleRate()/2.f-6000.f)) *
+	                   (700.f+1500.f*(k)+(1500.f+(k)*(44100.f/2.f-6000.f)) *
```
SecondFlight, line 232: See comment above.

thmueller64: Yes, a named constant would be better.
It most likely should be a constant. What happens when the sampling rate changes is that the filters cutoff frequency changes for the same settings otherwise, compare the figures below. The `w` is constant btw, see line 212.

CutoffFreq

SecondFlight: Thanks for the explanation there. It seems really odd at face value, but DSP still doesn't come naturally to me, especially filter coefficients and such. I think it's fine as long as it's easy to see that `44100.0f` is intentional.

thmueller64: That is just the `kfco` in the code, for which I think it is at least approximately the cutoff frequency at the start of a note, plotted against the knob position. I just noticed that if the plots above are valid then only for short notes. The `vcf_c0` in `w = vcf_e0 + vcf_c0` is set to `1` at the beginning of a note (for which I generated the plots) and then gets successively multiplied with the decay `vcf_c0 *= fs->envdecay` in `lb302filter::envRecalc()`. After a few moments `w` is nearly zero and the `kfco` reduces to `50.f + (k)*((2300.f-1600.f*(fs->envmod))` so the Env Mod knob is decreasing the cutoff now instead of increasing it like at the beginning of a note. You can hear this very well when you adjust the Env Mod knob for a pattern of many short notes versus a single long note. 

Anyways,  this does not change my opinion that there shouldn't be a dependency of `kfco` on the dynamic sampling rate, which changes the filters behavior completely for the same knob positions. What do you think?

SecondFlight:
Looks good to me. I'll merge this in a day or two if there are no objections.

Issue #5598, opened 2020-07-27
Fix annoying Qt5 deprecated warnings


JohannesLorenz:
### Bug Summary

LMMS brings tons of Qt5 deprecated warnings (e.g. `QFlag`) which can make it hard to see relevant compiler issues. The warnings are justified, they should be fixed (suppressing them would not be the solution).

#### Steps to reproduce

Compile LMMS (master) with Qt5.

#### Expected behavior

No Qt5 warnings.

#### Actual behavior

Tons of Qt5 warnings (e.g. `QFlag`, see *Logs*; has the `QString` warning already been fixed?)

#### Affected LMMS versions

At least on master.

#### Logs
<details>
  <summary>Click to expand</summary>
<pre>
[  0%] Building CXX object src/CMakeFiles/lmmsobjs.dir/gui/Lv2ViewBase.cpp.o
In file included from lv2/include/MainWindow.h:34,
                 from lv2/src/gui/Lv2ViewBase.cpp:46:
lv2/include/SubWindow.h:58:67: warning: 'constexpr QFlags<T>::QFlags(QFlags<T>::Zero) [with Enum = Qt::WindowType; QFlags<T>::Zero = int QFlags<Qt::WindowType>::Private::*]' is deprecated: Use default constructor instead [-Wdeprecated-declarations]
   58 |  SubWindow( QWidget *parent = NULL, Qt::WindowFlags windowFlags = 0 );
      |                                                                   ^
In file included from /usr/include/qt/QtCore/qglobal.h:1302,
                 from /usr/include/qt/QtCore/qchar.h:43,
                 from /usr/include/qt/QtCore/qstring.h:49,
                 from /usr/include/qt/QtCore/QString:1,
                 from lv2/include/Lv2ViewBase.h:33,
                 from lv2/src/gui/Lv2ViewBase.cpp:25:
/usr/include/qt/QtCore/qflags.h:123:80: note: declared here
  123 |     QT_DEPRECATED_X("Use default constructor instead") Q_DECL_CONSTEXPR inline QFlags(Zero) noexcept : i(0) {}
      |                                                                                ^~~~~~
In file included from lv2/src/gui/Lv2ViewBase.cpp:46:
[...]

PR #5619, opened 2020-08-10
Qt deprecation fix

uthidata:
Fix Qt deprecation warnings

- Replace QWheelEvent::delta with angleDelta
- Replace QWheelEvent::orientation with comparison of angleDelta.x and y

- Remove Qt::WindowFlags parameters when adding SubWindows
- Remove QFileDialog::Options parameters when opening files
These parameters are currently not used anywhere.

- Use QFlag(0) when initializing items in ladspaPortDialog instead of 0
- Use explicit enum when specifying QStyle state in ComboBox's paintEvent

- Use QElapsedTimer instead of QTime where appropriate
- Use QAtomicPointer::loadRelaxed() instead of load()
- Use QSet<T>(list.begin(), list.end()) instead of list.toSet()
- Use Qt::endl in QTextStream
- Use QString::isEmpty() instead of comparing to QString::null
- Use QFontMetrics::horizontalAdvance instead of width
- Use QWheelEvent::position() instead of pos()

### DIFF
--- a/plugins/lb302/lb302.cpp
+++ b/plugins/lb302/lb302.cpp
@@ -433,8 +433,11 @@ QString lb302Synth::nodeName() const
 // OBSOLETE. Break apart once we get Q_OBJECT to work. >:[
 void lb302Synth::recalcFilter()
 {
+#if (QT_VERSION >= QT_VERSION_CHECK(5,14,0))
+	vcf.loadRelaxed()->recalc();
+#else
 	vcf.load()->recalc();
-
+#endif
 	// THIS IS OLD 3pole/24dB code, I may reintegrate it.  Don't need it
 	// right now.   Should be toggled by LB_24_RES_TRICK at the moment.
 
@@ -683,7 +686,11 @@ void lb302Synth::initNote( lb302Note *n)
 
 	if(n->dead ==0){
 		// Swap next two blocks??
+#if (QT_VERSION >= QT_VERSION_CHECK(5,14,0))
+		vcf.loadRelaxed()->playNote();
+#else
 		vcf.load()->playNote();
+#endif
 		// Ensure envelope is recalculated
 		vcf_envpos = ENVINC;
 
--- a/plugins/sf2_player/patches_dialog.h
+++ b/plugins/sf2_player/patches_dialog.h
@@ -43,7 +43,7 @@ class patchesDialog : public QDialog, private Ui::patchesDialog
 public:
 
 	// Constructor.
-	patchesDialog(QWidget *pParent = 0, Qt::WindowFlags wflags = 0);
+	patchesDialog(QWidget *pParent = 0, Qt::WindowFlags wflags = QFlag(0));
 
 	// Destructor.
 	virtual ~patchesDialog();
--- a/src/core/audio/AudioFileOgg.cpp
+++ b/src/core/audio/AudioFileOgg.cpp
@@ -30,7 +30,9 @@
 
 #ifdef LMMS_HAVE_OGGVORBIS
 
-
+#if (QT_VERSION >= QT_VERSION_CHECK(5,10,0))
+#include <QRandomGenerator>
+#endif
 #include <string>
 #include <vorbis/vorbisenc.h>
 
@@ -136,8 +138,13 @@ bool AudioFileOgg::startEncoding()
 
 	// We give our ogg file a random serial number and avoid
 	// 0 and UINT32_MAX which can get you into trouble.
-	qsrand( time( 0 ) );
+#if (QT_VERSION >= QT_VERSION_CHECK(5,10,0))
+	QRandomGenerator::global()->seed(time(0));
+	m_serialNo = 0xD0000000 + QRandomGenerator::global()->generate() % 0x0FFFFFFF;
+#else
+	qsrand(time(0));
 	m_serialNo = 0xD0000000 + qrand() % 0x0FFFFFFF;
+#endif
 	ogg_stream_init( &m_os, m_serialNo );
 
 	// Now, build the three header packets and send through to the stream
--- a/include/DeprecationHelper.h
+++ b/include/DeprecationHelper.h
@@ -1,0 +1,63 @@
+/*
+ * DeprecationHelper.h - This file contains the declarations of helper functions
+ * which helps centralize the #ifdefs preprocessors regarding deprecation based on Qt versions.
+ * The functions are defined differently based on the callers' Qt versions.
+ *
+ * Copyright (c) 2020 Tien Dat Nguyen <ntd.bk.k56/at/gmail.com>
+ *
+ * This file is part of LMMS - https://lmms.io
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this program (see COPYING); if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA 02110-1301 USA.
+ *
+ */
+
+#ifndef DEPRECATIONHELPER_H
+#define DEPRECATIONHELPER_H
+
+#include <QFontMetrics>
+#include <QWheelEvent>
+
+/**
+ * @brief horizontalAdvance is a backwards-compatible adapter for
+ * QFontMetrics::horizontalAdvance and width functions.
+ * @param metrics
+ * @param text
+ * @return text's horizontal advance based on metrics.
+ */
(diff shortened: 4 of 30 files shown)


PhysSong:
I recommend you looking at the Qt documentation to check the version compatibility of those functions.

uthidata:
>I recommend you looking at the Qt documentation to check the version compatibility of those functions.

I am on it, got a bit sidetracked but I'm finishing up now.

plugins/sf2_player/patches_dialog.h
```diff
@@ -65,10 +65,10 @@ protected slots:
 
 protected:
 
-	void setBankProg(int iBank, int iProg);
+	void setBankProg( int iBank, int iProg );
 
-	QTreeWidgetItem *findBankItem(int iBank);
-	QTreeWidgetItem *findProgItem(int iProg);
+	QTreeWidgetItem *findBankItem( int iBank );
+	QTreeWidgetItem *findProgItem( int iProg );
```
DomClark, line 71: The formatting here should be left as-is - we're moving over to this style, without the spaces between the parentheses.

uthidata: I will amend the commits about these.

Btw do you have a .clang-format file I can use to format automatically? I hate to turn off format on save just for LMMS and turn it back on for every other project.

DomClark: There's #4690, but I have no idea whether it's up to date with our current [coding conventions](https://github.com/LMMS/lmms/wiki/Coding-Conventions). It might be a nice opportunity to test it though.

uthidata: The file in #4690 seems to tick all the boxes, however I cannot use it right now because the old files are not conform to such standards in the first place. Good to see it moving along though.

uthidata:
@DomClark The pragma warnings and style violations are gone. I think it's ready.

plugins/lb302/lb302.cpp
```diff
@@ -433,8 +433,11 @@ QString lb302Synth::nodeName() const
 // OBSOLETE. Break apart once we get Q_OBJECT to work. >:[
 void lb302Synth::recalcFilter()
 {
-	vcf.load()->recalc();
-
+#if (QT_VERSION >= QT_VERSION_CHECK(5,14,0))
+	vcf.loadRelaxed()->recalc();
+#else
+	vcf.load()->playNote();
```
DomClark, line 439: Shouldn't this line be `vcf.load()->recalc();`?

uthidata: Oh yes, my bad.

JohannesLorenz:
@DomClark as you commented this as lgtm, what kind of review do you expect? At least, from testing, I'm satisfied, as this removed all the warnings mentioned in #5598 .

There are still open comments in this PR btw.

DomClark:
> What kind of review do you expect?

Just after some confirmation that this fixed the warnings you were seeing, since you opened the issue. Sounds like it does.
> There are still open comments in this PR btw.

Resolved.

Will merge tomorrow if there are no objections.

PhysSong:
Please wait, I have to check something about wheel events.

PhysSong:
> Replace QWheelEvent::orientation with comparison of angleDelta.x and y

It *does* work for Qt 5, but I think it's not a good idea.
- If it's possible to handle both vertical and horizontal scroll events without that check, why not?
- Qt 5 generates two events(vertical one and horizontal one) for arbitrary scrolling to be compatible with Qt 4. That's not true in Qt 6 anymore.

`SongEditor::wheelEvent` looks somewhat tricky to get rid of the trick, so let me check first. `AutomationEditor` and `PianoRoll` look fine.

PhysSong:
To be clear, `SongEditor` handles horizontal scroll events and `TrackContainerView::scrollArea` handles vertical ones. This makes handling wheel events more complicated than other editors.

Veratil:
> 
> 
> To be clear, `SongEditor` handles horizontal scroll events and `TrackContainerView::scrollArea` handles vertical ones. This makes handling wheel events more complicated than other editors.

It doesn't handle horizontal well, but I made this #5174 which still needs to handle Qt version differences.

uthidata:
> * If it's possible to handle both vertical and horizontal scroll events without that check, why not?

The check was there before. I merely translated it into non-deprecated form.


> * Qt 5 generates two events(vertical one and horizontal one) for arbitrary scrolling to be compatible with Qt 4. That's not true in Qt 6 anymore.

I'm not sure what you mean by this. Qt 6 will remove the orientation function, and the angleDelta with x and y will be there. Comparing their absolute values is the fastest way to determine the orientation. Unless you're saying the orientation is not important, but there should be a design document on how you expect it to work. Should the modifiers work regardless of orientation? Which orientation would be prefered?... I think it is not in the scope of this PR. I would gladly change them if you post an issue detailing the problem.

---

Now that I think of it, maybe we can put the desired behaviors in a unit test?

> It doesn't handle horizontal well, but I made this #5174 which still needs to handle Qt version differences.

Unit tests could theoretically solve this across the board. No more manual testings.

PhysSong:
I mean, wheel events can have arbitrary orientation in Qt6(not restricted to horizontal or vertical).

uthidata:
What's the desired behavior of wheel scroll on those windows? So I can write tests.

PhysSong:
A note on `QWheelEvent::position()`: There's [`QWheelEvent::posF()`](https://doc.qt.io/archives/qt-5.6/qwheelevent.html#posF) with slightly different signature(the former returns a value whereas the latter returns a constant reference).

PhysSong:
Ideally, A wheel event with `(dx, dy)` should behave the same as the combination of `(dx, 0)` and `(0, dy)`. If that's tricky for any reason, you should add a TODO comment for Qt 6.

uthidata:
> Ideally, A wheel event with `(dx, dy)` should behave the same as the combination of `(dx, 0)` and `(0, dy)`. If that's tricky for any reason, you should add a TODO comment for Qt 6.

The comparison is not there to differentiate between (dx, 0) and (0, dy), but to determine the right direction to scroll based on a single (dx, dy).

> A note on `QWheelEvent::position()`: There's [`QWheelEvent::posF()`](https://doc.qt.io/archives/qt-5.6/qwheelevent.html#posF) with slightly different signature(the former returns a value whereas the latter returns a constant reference).

QWheelEvent::posF is deprecated.

PhysSong:
> > Ideally, A wheel event with `(dx, dy)` should behave the same as the combination of `(dx, 0)` and `(0, dy)`. If that's tricky for any reason, you should add a TODO comment for Qt 6.
> 
> The comparison is not there to differentiate between (dx, 0) and (0, dy), but to determine the right direction to scroll based on a single (dx, dy).

IMHO, `(dx, dy)` should scroll `(dx, dy)`, neither `(dx, 0)` nor `(0, dy)`.

> > A note on `QWheelEvent::position()`: There's [`QWheelEvent::posF()`](https://doc.qt.io/archives/qt-5.6/qwheelevent.html#posF) with slightly different signature(the former returns a value whereas the latter returns a constant reference).
> 
> QWheelEvent::posF is deprecated.

I know, but I mentioned it for `position().toPoint()` vs. `pos()` vs. `posF.toPoint()`.

uthidata:
> IMHO, `(dx, dy)` should scroll `(dx, dy)`, neither `(dx, 0)` nor `(0, dy)`.

How can one scroll (dx, dy) while holding Ctrl + Shift? There is only one scrolling dimension, but two inputs. They will have to combine, or selected. My code currently selects dy.

uthidata:
Hi, what can I change to get this merged?

PhysSong:
I think the best way for now is reverting `QWheelEvent::orientation()` changes or adding a FIXME comment(for Qt 6).

uthidata:
I just added FIXME comments where orientation was replaced.

JohannesLorenz:
Can anyone please summarize which points are still open in this PR?

uthidata:
All deprecation warnings are fixed. The point about horizontal/vertical scrolling distinction is dealt with as suggested. I don't think there's anything left to do.

JohannesLorenz:
@PhysSong Can you please check whether your suggestion was coded correctly, and if yes merge this PR?

PhysSong:
I'll check this again tomorrow and merge if looks okay.

JohannesLorenz (approved):
approved from my side - no compiler warnings anymore

PhysSong:
I'll merge it in a day unless there are any objections.

src/core/audio/AudioFileOgg.cpp
```diff
@@ -136,8 +138,13 @@ bool AudioFileOgg::startEncoding()
 
 	// We give our ogg file a random serial number and avoid
 	// 0 and UINT32_MAX which can get you into trouble.
-	qsrand( time( 0 ) );
+#if (QT_VERSION >= QT_VERSION_CHECK(5,10,0))
+	QRandomGenerator::global()->seed(time(0));
+	m_serialNo = 0xD0000000 + QRandomGenerator::global()->generate() % 0x0FFFFFFF;
```
PhysSong, line 143: Reported on Discord: an user got the following error on OGG export.
> Attempted to overwrite a QRandomGenerator to system() or global().

The documentation says `QRandomGenerator::global()` is already seeded and can't be overwritten by `seed()`.

JohannesLorenz: Open a new issue with this information so it gets attention and a PR can reference it?

Issue #3594, opened 2017-05-31
Fix semantic versioning


tresf:
We've been using the `1.0.0-rc1` style releases, however this is not proper.

According to semantic versioning, it should be "rc.1" instead of "rc1". (e.g. `2.0.0-rc.1`)

http://semver.org/spec/v2.0.0-rc.1.html

Umcaruje:
The RC4 release is now on a `v1.2.0-rc.4` tag. I suggest we close this issue out and have correct versioning for all consecutive releases, starting from RC4.

tresf:
Have we unit tested this against our internal comparators?

Lukas-W:
The way I understand the spec, "rc1" is not a violation. Pre-release versions may be dot-separated, but don't have to be. Dot-separated identifiers that contain letters are compared lexicographically. So `rc1` < `rc2` holds true. We're fine as long as we don't reach `rc10`.

tresf:
Correct, we'd be betting on the chance we never hit 10, which isn't a violation, but also is not proper.

Lukas-W:
> which isn't a violation, but also is not proper

Agreed.

I just wanted to share the way pre-release versions are sorted according to the specification (see [point 11](http://semver.org/spec/v2.0.0.html#spec-item-11)), because of what switching mid-release will mean: `rc.4` < `rc3` because `rc` < `rc3`! Better not start with RC4.

Umcaruje:
>because of what switching mid-release will mean: rc.4 < rc3 because rc < rc3! Better not start with RC4.

Thanks for the insight, I reverted the change for RC4 on the draft release

Hussam-Eddin-Alhomsi:
I think we should omit the 0s to the right of the version number. They're as useless as having 0s to the right of a decimal number.
For example: `1.2.0` should be `1.2`.

tresf:
@Hussam-Eddin-Alhomsi, we increment this zero to a one when we release a patch to a stable version.  This has happened every single release.  Truncating the zero would be more work than what it's worth.

Spekular:
Now that #5636 is merged I think we can definitely consider this closed. If anyone disagrees please feel free to reopen w/ justification!

PR #5621, opened 2020-08-11
Don't give clips a hidden default name (Fix #5528)

### LINKED ISSUE #5528: Can't give a segment the same name as the track
#### Affected LMMS versions

1.2.1

Spekular:
Fixes #5528 and simplifies/removes lots of name related logic by not giving clips a default name. Adds an upgrade routine to remove the default names in old projects, to avoid previously invisible clip names becoming visible.

Known issues:
- ~Initial testing implies the upgrade routine isn't called for .mmp files~
  - This was likely a coincidence related to the version issue described below, further testing shows .mmp and .mmpz files behave the same way
- ~Version bump doesn't work with LMMSBot builds nor builds from a git clone, so the upgrade logic isn't guaranteed to run once. Depending on your LMMS version and project version it could run multiple times or never run. However, I have confirmed that versions like "1.2.2.686" are correctly detected as less than "1.3.0-alpha-1".~
  - ~Fix: make a 1.3.0-alpha-1 tag when this PR is merged?~

For reviewers: The commit "Automatic formatting changes" can be ignored while reviewing.

A future improvement would be to stop abusing `name()` to give hints in automation clips, but IMO this is out of scope for this PR. There is a `displayName()` function that might be appropriate, but it's used inconsistently.


Spekular:
#### For testers

First and foremost, test that your old projects are correctly upgraded (I don't recommend saving over old projects when doing this!). Opening an old project should *not* lead to clip names becoming visible that were previously invisible. Then, for more thorough testing, you can test:
- Creating clips of these types:
  - Unnamed: Created, name not modified
  - Named: Created and given a name (other than the track name)
  - Same-named: Created and renamed to match track name
  - Reset: Name has been reset
- Renaming a track containing each type of clip
- Copying each type of clip to another track
- Resetting a clip of each type

Spekular:
@PhysSong requesting review from you because I've seen you handle version bumps in the past, so I thought you might know why the bump to 1.3.0-alpha-1 didn't seem to work.

PhysSong:
If you build from a Git repository(not using the tarball), LMMS uses the output of `git describe --tags`. It depends on tags of the repository.

Spekular:
@PhysSong in other words, once this PR is merged we need to make a 1.3.0-alpha-1 tag if we want the version to update? If so, do you mind if I do that? (assuming I have the correct permissions and can figure it out, haha)

Spekular:
Also, I suppose now would be a good time to confirm if the version numbering I used is actually correct. It looks like the current state of the PR would generate `1.3.0-alpha.1`, but will the `.1` be overridden by an automatic build number? Or will it prevent automatic build numbers from showing? Perhaps I should set `VERSION_STAGE` to `alpha-1` and leave `VERSION_BUILD` unset.

```
SET(VERSION             "${VERSION_MAJOR}.${VERSION_MINOR}.${VERSION_RELEASE}")
IF(VERSION_STAGE)
	SET(VERSION     "${VERSION}-${VERSION_STAGE}")
ENDIF()
IF(VERSION_BUILD)
	SET(VERSION     "${VERSION}.${VERSION_BUILD}")
ENDIF()
```

PhysSong:
You should use either `alpha1`(perhaps `alpha01` in case there will be `alpha10`) or `alpha.1` for `VERSION_STAGE` and leave `VERSION_BUILD` to `"0"`. If you use `alpha.1`, `ProjectVersion` should be updated to allow something like `1.3.0-alpha.1.789`.

PhysSong:
See also: #3594

Spekular:
Note to self (emphasis mine): 

> Precedence for two pre-release versions with the same major, minor, and patch version MUST be determined by comparing each dot separated identifier from left to right until a difference is found as follows:
> 
> Identifiers consisting of only digits are compared numerically.
> 
> **Identifiers with letters or hyphens are compared lexically in ASCII sort order.**

Longer strings are bigger than smaller strings if the first part is the same. Testing a few strings with [this site](https://www.textfixer.com/tools/alphabetical-order.php) gives, in ascending order:
- a
- aa
- alpha
- alpha01
- beta

This means that `1.3.0-alpha01 > 1.3.0-alpha.2`, because `alpha01 > alpha` is true. In other words, if we use `alpha01` now, we commit to using `alpha##` format until we reach beta or rc versions.

Veratil:
https://semvercompare.azurewebsites.net/?version=1.3.0-beta&version=1.3.0-aa&version=1.3.0-alpha&version=1.3.0-alpha01&version=1.3.0-a&version=1.3.0-alpha.2&version=1.3.0&version=1.3.0-alpha.2%252Bdeadbeef

semver ordering of examples and a few others, lower in the list is > higher

Spekular:
@PhysSong done in #5636. Should we close #3594 once that's merged?

PhysSong:
It seems like new conflicts are not resolved properly.

PhysSong:
It seems like the merge commit wasn't marked as a merge commit for some reason. I added a commit to fix it.

Spekular:
@PhysSong thank you, I was going to manually transfer the changes to a new branch if the diff didn't resolve to something sensible (as it's occasionally done before).

This is ready for review now that it's no longer blocked by any other PRs.

src/core/DataFile.cpp
```diff
@@ -1395,36 +1395,8 @@ void DataFile::upgrade()
...
-	{
-		upgrade_1_2_0_rc3();
-		upgrade_1_2_0_rc2_42();
-	}
-	if (version < "1.3.0-alpha-1"){ upgrade_1_3_0_alpha_1(); }
-	if( version < "1.3.0" )
-	{
-		upgrade_1_3_0();
-	}
+	// Bump the file version (which should be the size of the upgrade methods vector)
```
Spekular: This change isn't actually part of my PR, it's from merging with master.

include/DataFile.h
```diff
@@ -113,9 +113,8 @@ class LMMS_EXPORT DataFile : public QDomDocument
 	void upgrade_1_1_0();
 	void upgrade_1_1_91();
 	void upgrade_1_2_0_rc3();
-	void upgrade_1_2_0_rc2_42();
-	void upgrade_1_3_0_alpha_1();
 	void upgrade_1_3_0();
+	void upgrade_noHiddenClipNames();
```
Spekular, line 117: The removal of `rc2_42` isn't part of this PR, it's from the merge with master. `alpha_1` (added in this PR) was renamed to `noHiddenClipNames` and moved to the bottom for consistency. Since it's the newest upgrade routine it needs to be last in the upgrade routine list, so it makes sense for it to be last here as well.

Spekular:
@Veratil I'm re-requesting review from you since you checked this out before. The changes are pretty minor, the relevant diff is [here](https://github.com/LMMS/lmms/pull/5621/files/d956154f40f1ab9ec202e546e397188fd40bd1db..d2039fb170deb7931ba7ab903f967924e6303714).

For anyone reviewing for the first time, [this](https://github.com/LMMS/lmms/pull/5621/files/04a4eaf06478f4d754c52073887dd03a4dd8ff86..d2039fb170deb7931ba7ab903f967924e6303714) is the relevant diff (it excludes most of the formatting changes).

src/tracks/BBTrack.cpp
```diff
@@ -212,12 +212,12 @@ void BBTCOView::paintEvent( QPaintEvent * )
...
 	QColor c;
 	bool muted = m_bbTCO->getTrack()->isMuted() || m_bbTCO->isMuted();
-	
+
 	// state: selected, muted, default, user selected
-	c = isSelected() ? selectedColor() : ( muted ? mutedBackgroundColor() 
-		: ( m_bbTCO->m_useStyleColor ? painter.background().color() 
+	c = isSelected() ? selectedColor() : ( muted ? mutedBackgroundColor()
+		: ( m_bbTCO->m_useStyleColor ? painter.background().color()
 		: m_bbTCO->colorObj() ) );
```
Veratil, line 219: I think this is a good candidate for style reformat to make it a bit easier to read. Just an example to show it's 3 layers deep, but I'm not opposed to changing it to this. 😄
```suggestion
	c = isSelected()
		? selectedColor()
		:	(muted
			? mutedBackgroundColor()
			:	(m_bbTCO->m_useStyleColor
				? painter.background().color()
				: m_bbTCO->colorObj()
				)
			)
		;
```

src/core/DataFile.cpp
```diff
@@ -1476,11 +1505,9 @@ void DataFile::loadData( const QByteArray & _data, const QString & _sourceFile )
 			}
 
 			// the upgrade needs to happen after the warning as it updates the project version.
-			if( createdWith.setCompareType( ProjectVersion::Build )
-								< openedWith )
-			{
-				upgrade();
-			}
+			if( createdWith.setCompareType( ProjectVersion::All ) <
+				openedWith.setCompareType( ProjectVersion::All ) )
```
Veratil, line 1509: I find this is easier to read using a leading operator rather than trailing.
```suggestion
			if( createdWith.setCompareType( ProjectVersion::All )
				< openedWith.setCompareType( ProjectVersion::All ) )
```

src/core/DataFile.cpp
```diff
@@ -1355,6 +1355,35 @@ void DataFile::upgrade_1_3_0()
...
+{
+	QDomNodeList tracks = elementsByTagName("track");
+
+	auto clearDefaultNames = [](QDomNodeList clips, QString trackName)
+	{
+		for (int j = 0; !clips.item(j).isNull(); ++j)
+		{
+			QDomElement clip = clips.item(j).toElement();
+			QString clipName = clip.attribute("name", "");
+			if (clipName == trackName){ clip.setAttribute("name", ""); }
```
Veratil, line 1368:
```suggestion
			if (clipName == trackName) { clip.setAttribute("name", ""); }
```

src/tracks/BBTrack.cpp
Veratil: I know it's not part of the PR changes, but since it popped up I'm commenting on it.

src/core/DataFile.cpp
```diff
@@ -1355,6 +1355,35 @@ void DataFile::upgrade_1_3_0()
 	}
 }
 
+void DataFile::upgrade_noHiddenClipNames()
+{
+	QDomNodeList tracks = elementsByTagName("track");
+
+	auto clearDefaultNames = [](QDomNodeList clips, QString trackName)
+	{
+		for (int j = 0; !clips.item(j).isNull(); ++j)
```
DomClark, line 1364: Wouldn't `j < clips.length()` be a more natural choice of condition?

DomClark (commented):
Looks good to me mostly. Can you revert the formatting change in `AutomationPattern.h` so it isn't touched unnecessarily? `BBTrack.cpp` also has quite a few formatting changes, which I worry may cause conflicts with #5573.

src/core/DataFile.cpp
Spekular: I believe this iteration code is nicked directly from another upgrade routine, but I can swap it for a comparison to `clips.size()`. I'm tempted to add `if (item(j).isNull) { continue(); }` if so (to avoid introducing additional assumptions), but maybe that's paranoid?

PhysSong: How about using `clips.size()` and checking with `clips.item(j).isElement()` before converting to `QDomElement`?

src/tracks/BBTrack.cpp
Spekular: I'll leave it alone in order to minimize unrelated changes.

src/core/DataFile.cpp
Spekular: Ok, after further investigation it looks like the check is unnecessary. I added the following at the start of the upgrade routine:

```
qDebug() << "About to test null stuff, if we crash after this you know why";
QDomNode nullNode = QDomNode();
QDomElement nullElement = nullNode.toElement();
QString nullString = nullElement.attribute("name", "null");
qDebug() << "nullString = " << nullString;
```

Upon opening an older project, console output is: 

```
About to test null stuff, if we crash after this you know why
nullString =  "null"
```

So if the node is null we'll receive a null element, and calling `attribute` on the null element will fall back to the default string. In this case `clipName` will be set to an empty string, which is harmless.

src/core/DataFile.cpp
Spekular: Fixed in latest commit.

src/core/DataFile.cpp
Spekular: I made a somewhat drastic change to fix this that avoids these specific `setCompareType` calls completely, thus fitting the condition on one line. For the other long comparison I moved the comparator to the start of the second line as you suggested.

src/core/DataFile.cpp
```diff
@@ -1355,6 +1355,35 @@ void DataFile::upgrade_1_3_0()
...
+	};
+
+	for (int i = 0; i < tracks.size(); ++i)
+	{
+		QDomElement track = tracks.item(i).toElement();
+		QString trackName = track.attribute("name", "");
+
+		QDomNodeList instClips = elementsByTagName("pattern");
+		QDomNodeList autoClips = elementsByTagName("automationpattern");
+		QDomNodeList bbClips = elementsByTagName("bbtco");
```
DomClark, line 1379: Shouldn't these be `track.elementsByTagName`? Otherwise a clip's name will be cleared if it matches the name of any track in the project, not just its parent track.

Spekular: Oh dear, good catch. Will update, probably later today.

Spekular:
Latest commit addresses Dom's most recent comment, and I used a new test file to confirm that the behavior was wrong before and isn't now.

I also touched some code that should arguably be in a separate PR: Previously, `legacyFileVersion` would only be called if the attribute `version` existed and was set to exactly `1.0`. If `version` didn't exist, nothing would happen and the file version would be set to the latest (18 in this case) meaning no upgrades run. I changed it so that if the version is missing OR set to `1.0`, `legacyFileVersion` is used, and otherwise the normal parsing happens. If we want to be strict I can revert that and make a separate PR, but I was going crazy trying to figure out why my upgrade wasn't working with my new test file. It took a few `qDebug`s to realize that stripping out the version number wouldn't trigger `legacyFileVersion`.

src/core/DataFile.cpp
Spekular: Addressed in latest commit. I also went a bit out of scope with said commit, so feel free to tell me if I should revert the out-of-scope parts.

DomClark (approved):

PR #5623, opened 2020-08-11
Fix for Icons and comboboxes mismatch in arpeggiator in Instrument Editor #5494

DigArtRoks:
ComboBox
(https://github.com/LMMS/lmms/issues/5494)

Introduce a static const int variable for the default height of a ComboBox.
Set this height already in the constructor of the ComboBox object.
Update all modules setting the height of a ComboBox object to make use of the new constant.

### DIFF
--- a/include/ComboBox.h
+++ b/include/ComboBox.h
@@ -49,7 +49,7 @@ public:
 		return castModel<ComboBoxModel>();
 	}
 
-	static const int DEFAULT_HEIGHT = 22;
+	static constexpr int DEFAULT_HEIGHT = 22;
 
 public slots:
 	void selectNext();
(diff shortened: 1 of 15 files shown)


SecondFlight:
That looks much better. Thanks for fixing it!

SecondFlight:
I'll merge this soon ™️ if there are no objections.

DigArtRoks:
Not sure what went wrong after my 2nd commit at the travis-ci check... It looks like it could not retrieve the dependencies for the Debian I386 build.

zonkmachine:
> It looks like it could not retrieve the dependencies for the Debian I386 build.

Yes. I restarted the failing build and now it passed.

PR #5627, opened 2020-08-12
Refactor Clipboard methods

IanCaio:
This draft's goal is to refactor some of LMMS's clipboard functionality. Currently the system is a little bit unorganized because it mixes the OS's clipboard (accessed through `QClipboard`) and an internal LMMS's `Clipboard` class (only accessible on the current running instance of LMMS). It would be interesting to move completely to the OS's clipboard for the sake of consistency and also possibility of multiple LMMS instances interacting with themselves when it comes to copying and pasting.

The first commit is just a little reorganizing of some Clipboard related methods. First, the mime type string used to be obtained on both the Clipboard class and the StringPairDrag class (`Clipboard::mimeType()` and `StringPairDrag::mimeType()`). Since it's related to the clipboard, I moved it completely to the Clipboard class, the equivalents of the previous methods now being `Clipboard::mimeType( Clipboard::Default )` and `Clipboard::mimeType( Clipboard::StringPair )`. Any new mime type can be easily added to the `Clipboard` class. The idea is that the Clipboard class will be a class with helper methods for dealing with the clipboard (not a clipboard itself as it is now). Another change is that StringPairDrag class had methods for decoding a string pair key and value from mime data and from a Drop event. I moved the ones that decode the key and value from mime data to the Clipboard class, since they can and are used outside the context of a Drag&Drop event, and kept the ones that extract the key and value from a Drop event on the StringPairDrag class (which by itself calls the `Clipboard::decodeKey` and `Clipboard::decodeValue` respectively).

I know the draft is at its very beginning, but being a structural change I think it's worth having it here so changes are discussed before I dive too deep into it. Suggestions can be made and I take a smaller risk of refactoring in a way that is not agreed upon by other Devs.


IanCaio:
Last few commits continue to move clipboard logic to the Clipboard class (`Track.cpp` and `PianoRoll.cpp` now have fewer lines wasted with setting up the mime data and other things that are now made easier by helper methods from `Clipboard`). Also, saved a few lines on `TrackContentObjectView::cut()` by using `TrackContentObjectView::copy()` inside it to avoid duplicating code.

Next step is to figure out how to copy a `QDomElement` to the clipboard and retrieve it later, so I can convert the current `Clipboard::copy` and `Clipboard::getContent` methods to use the system's clipboard instead. Still not sure whether I should find a way to convert the `QDomElement` to a string and later rebuild the node using the copied text, or if I should find a way to copy the raw data to the clipboard and later retrieve it to build the node.

IanCaio:
I'm changing this PR from __Draft__ to __Ready to review__ because the goal of the PR was achieved on a preliminary stage, which was refactoring the Clipboard class and clipboard operations so they are more consistent and all use the system's clipboard.

- `class Clipboard` is now a helper class for dealing with the system's clipboard.
- The TCO view copy and paste operations now use the same logic for either individual or multiple TCOs.
- `TCO::copy()` and `TCO::paste()` methods were merged in a single `TCO::copyStateTo( src, dst )` method, since the only place they were used after refactoring had this as a goal (`src/tracks/BBTrack.cpp`). Since it's an internal operation, not related to the userland copy/paste, for that method the clipboard is not used (if it were we would have unwanted behavior, as cleaning the clipboard when cloning a BBTrack).
- `TCO::cut` was removed since it was not used anywhere after refactoring.
- Tried to move a lot of the Clipboard logic to the `class Clipboard` to make the code on other files less polluted and more readable.

Right now, `TCO::copyStateTo()` converts `src` and `dst` from `TrackContentObject *` to `JournallingObject *` because that's how they were treated back when `Clipboard::copy()` and `Clipboard::getContent()` were used. I'm not sure if this is necessary so that's a point that could be addressed during review (not sure because I don't know if `nodeName` and `saveState`/`restoreState` could have different unwanted behaviors on their overridden versions).

__For Testers:__

Things worth testing:

- Copy, Cut, Paste, Remove and Mute operations on individual TCOs, multiple TCOs and individual TCOs when there are other TCOs selected.
- Cloning BBTracks with and without TCO content in their tracks.
- Copy, Cut and Paste operations on other elements besides TCOs, like PianoRoll notes, automation values, text, etc.

PhysSong:
> Right now, `TCO::copyStateTo()` converts `src` and `dst` from `TrackContentObject *` to `JournallingObject *` because that's how they were treated back when `Clipboard::copy()` and `Clipboard::getContent()` were used. I'm not sure if this is necessary so that's a point that could be addressed during review (not sure because I don't know if `nodeName` and `saveState`/`restoreState` could have different unwanted behaviors on their overridden versions).

Child classes are supposed to override `nodeName()`. No classes which inherit `JournallingObject` override `saveState`/`restoreState`. If you REALLY want to achieve that behavior, you should use `src->JournallingObject::saveState()` instead of casting.

IanCaio:
> Child classes are supposed to override `nodeName()`. No classes which inherit `JournallingObject` override `saveState`/`restoreState`. If you REALLY want to achieve that behavior, you should use `src->JournallingObject::saveState()` instead of casting.

To be honest, I wasn't sure if calling the base class `nodeName()` was necessary. I just did it that way because the original method operated over `JournallingObject`s. But taking another look at the code base I think it's not even a requirement to do the casting. Maybe someone can confirm, but I'll rewrite the method tomorrow morning to use `TrackContentObject *` instead and see how it behaves. I'm more confident right now that it will likely behave just the same.

IanCaio:
It was a quick fix, so I just updated the PR. Now the `TrackContentObject::copyStateTo` method doesn't cast the pointers to `JournallingObject *`. I made a quick test cloning some BBTracks that had tracks with TCOs and it seems to be working just the same. `JournallingObject *` was probably being originally used because `Clipboard::copy` and `Clipboard::getContent` were meant to be more generic and available to be used on other Journalling Objects besides TCOs.

include/Clipboard.h
```diff
@@ -29,25 +29,40 @@
...
+	static void copyStringPair( const QString & key, const QString & value );
+	static QString decodeKey( const QMimeData * mimeData );
+	static QString decodeValue( const QMimeData * mimeData );
+
+	static const char * mimeType( MimeType type )
+	{
+		switch( type )
+		{
+			case Clipboard::StringPair:
+				return( "application/x-lmms-stringpair" );
```
PhysSong, line 58:
```suggestion
				return "application/x-lmms-stringpair";
```
Return Without parentheses.

include/Clipboard.h
```diff
@@ -29,25 +29,40 @@
...
+			break;
+			case Clipboard::Default:
+			default:
+				return( "application/x-lmms-clipboard" );
```
PhysSong, line 62:
```suggestion
				return "application/x-lmms-clipboard";
```
Return Without parentheses.

PhysSong (commented):
There are minor formatting issues from copy-pasting. The code looks fine.
If someone can also review this, it will be great.

src/core/Track.cpp
```diff
@@ -184,32 +184,24 @@ bool TrackContentObject::comparePosition(const TrackContentObject *a, const Trac
...
- *
- *  Pastes this track content object into a track.
- *
- * \param _je The journal entry to undo
- */
-void TrackContentObject::paste()
-{
-	if( Clipboard::getContent( nodeName() ) != NULL )
+	// If the node names match we copy the state
+	if( src->nodeName() == dst->nodeName() )
```
DomClark, line 198: I think this if statement can surround the whole of this method. There's no need to serialise the state if it's not going to be deserialised, and there's no need to resolve and update things if nothing has changed.

src/core/Track.cpp
```diff
@@ -1239,75 +1217,35 @@ void TrackContentObjectView::remove( QVector<TrackContentObjectView *> tcovs )
...
-		QMimeData *tco_content = new QMimeData;
-		tco_content->setData( StringPairDrag::mimeType(), finalString.toUtf8() );
-		QApplication::clipboard()->setMimeData( tco_content, QClipboard::Clipboard );
-	}
-	else
-	{
-		tcovs.at(0)->cut();
-	}
+	// Now that the TCOs are copied we can delete them, since we are cutting
+	// TODO: Is it safe to call tcov->remove(); on the current TCOV instance?
```
DomClark, line 1234: It is safe to call `TrackContentObjectView::remove` on the current TCOV - it results in a call to `QWidget::close`, which is safe to call on the current widget.

src/core/Track.cpp
```diff
@@ -1239,75 +1217,35 @@ void TrackContentObjectView::remove( QVector<TrackContentObjectView *> tcovs )
...
+	TrackContentWidget *tcw = getTrackView()->getTrackContentWidget();
 
-		if( tcw->pasteSelection( tcoPos, md ) == true )
-		{
-			// If we succeed on the paste we delete the TCO we pasted on
-			remove();
-		}
-	}
-	else
+	if( tcw->pasteSelection( tcoPos, Clipboard::getMimeData() ) == true )
```
DomClark, line 1245: Since you're modifying this code anyway, get rid of `== true`?

DomClark (commented):
I think `Clipboard.cpp` should be moved from `src/core` to `src/gui`. I don't know whether this should be done in this PR, or left for the refactoring effort.

Veratil:
> I think `Clipboard.cpp` should be moved from `src/core` to `src/gui`. I don't know whether this should be done in this PR, or left for the refactoring effort.

💯 agree.

include/Clipboard.h
```diff
@@ -29,25 +29,40 @@
...
 class Clipboard
 {
 public:
-	typedef QMap<QString, QDomElement> Map;
-
-	static void copy( JournallingObject * _object );
-	static const QDomElement * getContent( const QString & _node_name );
-
-	static const char * mimeType()
+	enum MimeType
```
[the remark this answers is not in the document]
IanCaio: I thought about that, but wouldn't that result in a bit more verbose code when we need to access a MimeType? With `enum` we just need to type `Clipboard::StringPair` for example, while with `enum class` we would need `Clipboard::MimeType::StringPair` afaik. Unless we had the MimeType enum outside the class, but then there's the problem of using another name from the global namespace. I don't know, if you still think it's worth it to change it I can do it.

IanCaio:
I merged master and added 2 new commits. First one addresses the requested changes from the reviews listed below:
- Removes parenthesis from return statements
- Adds a `getString` method to Clipboard and change PianoRoll.cpp to use it instead of directly accessing the mime data.
- Moves a conditional from `TrackContentObject::copyStateTo` to surround the whole method, since nothing has to be run if the conditional is false.
- Removes a comment that had a question about `TCOV->remove()`, already answered by Dom.
- Removes unnecessary `== true` in a conditional.

Second commit was a small change to `AutomatableModelView.cpp`. It used the Qt clipboard directly to copy float values to the clipboard and paste them back to an Automatable Model. It now uses the Clipboard class and the `getString` and `copyString` methods for consistency.

PhysSong:
There's a merge conflict due to #5619, but it looks easy to resolve that.

IanCaio:
> There's a merge conflict due to #5619, but it looks easy to resolve that.

Yeah, from what I've seem the only thing I changed on this file was removing `#include "StringPairDrag.h"`, which #5619 did too. I'll merge master here and update the PR to resolve the conflict.

IanCaio:
Merged master and conflict on `src/tracks/Pattern.cpp` resolved!

include/Clipboard.h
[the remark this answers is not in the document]
DomClark: It would indeed be more verbose, but that's not necessarily a bad thing, and `enum class`es have a couple of advantages over `enum`s.
Firstly, they're more strongly typed - they aren't implicitly convertible to `int` - and the C++ Core Guidelines recommend them over `enum`s primarily for this reason ([Enum.3](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Renum-class)).
Secondly, they offer better organisation and scoping of their enumerators. Plain `enum`s inject their enumerators, a group of closely related constants, into the containing scope, where they sit equally among less related things. This can make it hard to tell what's related to what without reading the header itself. (Does `Clipboard::Default` mean the default clipboard? The default key? `Clipboard::MimeType::Default` is unambiguous.) It can also lead to name collisions if another `enum` is defined in the future, e.g. `enum ClipboardType { Default, Selection, FindBuffer };`.

An `enum` works fine, so I won't require it to be changed, but I do prefer `enum class`.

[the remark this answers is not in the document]
IanCaio: They are all valid and important points, I think the extra verbosity is worth all the benefits you mentioned. I just used `sed` to replace the enum values and should be committing later. On the process I might have found a small bug which I'm trying to find more about (a small misbehavior when cloning BB Tracks in an specific situation), but I should commit anyways since it is almost surely unrelated to the enum changes.

IanCaio:
Changed the Clipboard class to use `enum class` instead of `enum` for the MimeType.

During the process a small bug was found. Steps for reproducing:

1. Open the default project (TripleOsc, SampleTrack, BBTrack and AutomationTrack).
2. Create 2 TCOs on the TripleOsc track on Bars 1 and 2
3. Ctrl+Drag the TripleOsc track to the BBEditor
4. Add some notes to the Kicker track (not necessary for the bug to happen)
5. Clone the BBTrack

After those steps the notes added to the `Beat/Bassline 0` Kicker track are gone and when you switch to the `Clone of Beat/Bassline 0` track on the BB Editor the place where you add the notes to the Kicker are missing.

I'm still investigating what is causing it. Before, cloning a BBTrack would trigger calls to `TCO::copy()` and `TCO::paste()`. On this PR the new method `TCO::copyStateTo()` is called instead. I still haven't found reasons for that method to cause this misbehavior but I'll keep debugging here.

IanCaio:
About the bug mentioned before: Apparently it was already present before my PR, I'm sorry about that. I'll still try to figure out what is causing it, but disregard that bug when reviewing the code.

DomClark (approved):
Looks good to me. One more thing: how about making `Clipboard` a namespace instead of a class, since everything in it is now public and static?

IanCaio:
Sure, it could be done! I'd basically have to replace `class` by `namespace` and remove `public:` and `static` keywords right? I tried quickly here but ran into some linker issues with multiple declarations, will have to take a look with more attention later.

PhysSong:
If you run into multiple definitions issue, you can add `inline`/`static inline` or move the definition to the source file depending on the function body.
`inline` without `static` means "select any of multiple copies", and `static` means "use the local copy, and don't care about other copies".

IanCaio:
`Clipboard` was changed from a `class` to a `namespace`, and `using namespace Clipboard;`  statements were added where needed (a comment above the statement with the used methods/variables from the namespace were also added). There were two methods where I didn't add the `using namespace`, which were `StringPairDrag::decodeKey` and `StringPairDrag::decodeValue` since both were one-liners.

Thanks @PhysSong for the hint about using `inline` to avoid the multiple definition errors!

IanCaio:
Resolved the conflict introduced by #5661 . Does it look good to go?

On the "Files Changed" there appeared some stuff about `src/3rdparty/ringbuffer`, even though I ran `git submodule update` on that branch too. Is that normal?

PhysSong:
I think you've commited the merge commit before running `git submodule update`. If you have troubles with submodules, I or some other devs may fix it.

IanCaio:
> I think you've commited the merge commit before running `git submodule update`. If you have troubles with submodules, I or some other devs may fix it.

Thanks! I thought I ran it before the `git merge`. I'm unsure now because of the merge conflict, whether I ran it before `git merge` or before `git commit`. Anyways, even running it again now it doesn't show any changes to be committed, only on the `Files Changed` here on Github. Do I have to do anything to fix that?

PhysSong:
In that case, ping me when it's ready to merge.

IanCaio:
I think I fixed it. Did a hard reset to revert the merge and did it again. Even running `git submodule update` before the merge, the _src/3rdparty/ringbuffer_ folder appeared as "modified but not staged". Last time I think I accidentally added it to the commit. This time I left it out, finished the merge and ran `git submodule update` again after it. The ringbuffer folder was then updated correctly. So maybe `git submodule update` takes into account the current commit timestamp and the first `git submodule update` kept ringbuffer on the state it was on my last commit in this branch. Because master had updated files for ringbuffer, it caused them being shown as "modified but not staged" on the merge. Keeping them outside the merge and then running `git submodule update` again probably updated ring buffer to the current timestamp of master that was just merged to this branch making them match again.

Well, that's just what I think happened. Anyways, the ringbuffer files are out of the "Files Changed" preview now!

PhysSong (approved):

PhysSong:
Merge?

Issue #5484, opened 2020-05-06
theme problem when renaming a track


trebmuh:
### Bug Summary

When renaming an instrument, selecting a part of the string isn't very readable because the background color is pretty much the same than the normal color.

#### Steps to reproduce
Rename a track (right-click left-double-click).
Select a part of it.

#### Expected behavior
You can easily see the selection because of a background colors difference

#### Actual behavior
You can't easily see the differences because the difference isn't strong enough. 

#### Screenshot
("_space + tra_" selected):

![lmms-rename-background-color](https://user-images.githubusercontent.com/8705846/81237376-c56e6400-8fff-11ea-941f-1e784c83f0a3.png)

#### Affected LMMS versions
1.2.1

#### Logs
no logs needed

zonkmachine:
Yes. It's the default theme that is a bit on the dark side.

trebmuh:
Still affecting 1.2.2.

DigArtRoks:
See my pull request for fixing this. You can also straight away edit data/themes/default/style.css in your lmms installation directory. No need to rebuild lmms.

trebmuh:
Awesome, it fixes it!
Thank you very much.

PR #5628, opened 2020-08-12
Change the background color of the selected text in a text box.

DigArtRoks:
The new background color matches the green background of selected items in a tree view.
TextBoxSelection

Fix for https://github.com/LMMS/lmms/issues/5484
The fix is not limited to track renaming, but to all texboxes that can be edited.

### DIFF
--- a/data/themes/classic/style.css
+++ b/data/themes/classic/style.css
@@ -35,6 +35,7 @@ QLineEdit {
 	border: 2px inset rgba(91,101,113,128);
 	background: #49515b;
 	color: #e0e0e0;
+	selection-background-color: #202020;
 }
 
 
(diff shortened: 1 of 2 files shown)


Veratil:
Does classic theme have this problem? If so we should update that as well.

DigArtRoks:
The classic theme does not suffer from this problem. The selected text is highlighted in black, but the background color of non-selected text is grey. So they stand quite well apart.
Anyway for the sake of consistency between both themes and as reference for theme developers, I added the same modification to the style.css file of the classic theme. Though I kept the color as it was (near black).

TextBoxSelectionClassic

Veratil:
Tested by adding it to my stable installed style.css, works perfect. 👍

PR #5629, opened 2020-08-12
Fix use of translation functions

DomClark:
Fixes some outdated or incorrect usage of translation functions.
* The `pluginBrowser` class has since been renamed to `PluginBrowser`, so translations using that context have been updated to use the new name.
* The context for audio backend name translations has been changed from `setupWidget` to `AudioDeviceSetupWidget`. This is consistent with the use of `MidiSetupWidget` for MIDI backends.
* VeSTige and VstEffect no longer pass strings consisting of a single integer id through `tr`.
* The line `presName.replace(tr("\""), tr("'")); // QFileDialog unable to handle double quotes properly` from vst_base has been changed not to use `tr` - it doesn't make sense to allow translators to override this and break `QFileDialog`.
* Static strings in `InstrumentSoundShaping` now use `QT_TRANSLATE_NOOP` rather than `tr`; `tr` has no effect there since the translations have not been loaded yet at that point.
* Some places in `TrackContentObjectView`, `SetupDialog`, and `SampleTCOView` had calls to `tr` factored out of ternary expressions, or into lambdas. This breaks `lupdate`, which can't recognise anything but string literals within `tr` calls. Calls to `tr` have been moved to wrap the literals.
* In cases where it is necessary to pass the contents of a `QString` to `tr`, `toLatin1` has been changed to `toUtf8`, since this is the encoding Qt expects.
* `PluginBrowser` now translates descriptions in its tooltips for plugins that aren't subplugins.
* An unused static translation has been removed from `InstrumentTrack`.

### DIFF
--- a/src/gui/PluginBrowser.cpp
+++ b/src/gui/PluginBrowser.cpp
@@ -205,7 +205,9 @@ PluginDescWidget::PluginDescWidget(const PluginKey &_pk,
 	setFixedHeight( DEFAULT_HEIGHT );
 	setMouseTracking( true );
 	setCursor( Qt::PointingHandCursor );
-	setToolTip(_pk.description());
+	setToolTip(_pk.desc->subPluginFeatures
+		? _pk.description()
+		: tr(_pk.desc->description));
 }
 
 
--- a/include/AudioAlsa.h
+++ b/include/AudioAlsa.h
@@ -71,7 +71,7 @@ public:
 
 	inline static QString name()
 	{
-		return QT_TRANSLATE_NOOP( "setupWidget",
+		return QT_TRANSLATE_NOOP( "AudioDeviceSetupWidget",
 			"ALSA (Advanced Linux Sound Architecture)" );
 	}
 
--- a/include/AudioDummy.h
+++ b/include/AudioDummy.h
@@ -48,7 +48,7 @@ public:
 
 	inline static QString name()
 	{
-		return QT_TRANSLATE_NOOP( "setupWidget", "Dummy (no sound output)" );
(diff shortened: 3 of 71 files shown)


src/gui/PluginBrowser.cpp
```diff
@@ -205,7 +205,9 @@ PluginDescWidget::PluginDescWidget(const PluginKey &_pk,
 	setFixedHeight( DEFAULT_HEIGHT );
 	setMouseTracking( true );
 	setCursor( Qt::PointingHandCursor );
-	setToolTip(_pk.description());
+	setToolTip(_pk.desc->subPluginFeatures
+		? _pk.description()
+		: tr(_pk.desc->description));
```
liushuyu, line 210: Although this will certainly not break anything, I am not sure whether this will be effective since `_pk.desc->description` is dynamic.

DomClark: In the case that `_pk.desc->subPluginFeatures` is null, `_pk.desc` is a static descriptor from a plugin binary, e.g.: https://github.com/LMMS/lmms/blob/f34f894a84f9885b91da295f8ee676fb96b3c1e1/plugins/Amplifier/Amplifier.cpp#L34-L45

I'm not aware of any code that modifies these (we should probably make them const), and the description is clearly intended to be translated. This also matches the code in `EffectSelectDialog`: https://github.com/LMMS/lmms/blob/f34f894a84f9885b91da295f8ee676fb96b3c1e1/src/gui/EffectSelectDialog.cpp#L237

DomClark:
Merge?

liushuyu:
> Merge?

I am very sorry for the late response. I think it's good to merge now.

PR #5636, opened 2020-08-16
Use valid Semver versions for pre-releases

Spekular:
Reworks the way optional identifiers are handled to be more in line with Semver spec, adds a few new tests.

----

Our current versioning of `MAJOR.MINOR.PATCH.COMMITS` for builds from source is not a valid semantic version. I've added a workaround for versions up to 1.2.2.x to avoid breaking older comparisons, that converts them to 1.2.3-x. This is intentionally disabled for later versions as don't believe we should continue to use this versioning.

In order for everything to work, this PR (or another PR merged first) needs to update the versioning used for builds.
- Builds with no pre-release tag based on version `x.y.z` with `c` extra commits must be versioned `x.y.z+1-c`
  - Example: `1.2.2.42` becomes `1.2.3-42`
- Builds with a pre-release tag must separate the iteration from the stage, and can have commits appended
  - Example: `1.2.0-rc1` should have been `1.2.0-rc.1`, which could then have commits appended, `1.2.0-rc.1.42` 

For reference ([source](https://semvercompare.azurewebsites.net/?version=1.2.2&version=1.2.2-42&version=1.2.3&version=1.2.3-42&version=1.2.3-rc&version=1.2.3-rc.21&version=1.2.3-rc.42)): `1.2.2-42 < 1.2.2 < 1.2.3-42 < 1.2.3-rc < 1.2.3-rc.9 < 1.2.3-rc.42 < 1.2.3`


src/core/ProjectVersion.cpp
```diff
@@ -27,123 +27,82 @@
...
-	if(c == Minor)
-	{
-		return 0;
+	// Use the value of c to zero out identifiers we don't care about
+	int aMaj = 0, bMaj = 0, aMin = 0, bMin = 0, aPat = 0, bPat = 0;
+	if (c >= 1){ aMaj = a.getMajor(); bMaj = b.getMajor(); }
+	if (c >= 2){ aMin = a.getMinor(); bMin = b.getMinor(); }
+	if (c >= 3){ aPat = a.getPatch(); bPat = b.getPatch(); }
+
+	// Then compare as if we care about every identifiers
```
Spekular, line 73:
```suggestion
	// Then compare as if we care about every identifier
```

tests/src/core/ProjectVersionTest.cpp
```diff
@@ -23,6 +23,7 @@
  */
 
 #include "QTestSuite.h"
+#include <QtCore/QDebug>
```
Spekular, line 26:
```suggestion
```

tests/src/core/ProjectVersionTest.cpp
```diff
@@ -39,9 +40,18 @@ private slots:
...
+		//Identifiers with letters or hyphens are compare lexically in ASCII sort order
 		QVERIFY(ProjectVersion("1.1.0-alpha") < "1.1.0-beta");
 		QVERIFY(ProjectVersion("1.2.0-rc1") < "1.2.0-rc2");
+		//Build metadata MUST be ignored when determining version precedence
+		QVERIFY(ProjectVersion("1.2.2") == "1.2.2+metadata");
+		QVERIFY(ProjectVersion("1.0.0-alpha") < "1.0.0-alpha.1");
+		QVERIFY(ProjectVersion("1.0.0-alpha.1") < "1.0.0-alpha.beta");
+		QVERIFY(ProjectVersion("1.0.0-alpha.beta") < "1.0.0-beta");
+		QVERIFY(ProjectVersion("1.0.0-beta.2") < "1.0.0-beta.11");
+
```
Spekular, line 54:
```suggestion
```

include/ProjectVersion.h
```diff
@@ -36,29 +37,27 @@
...
-	int getRelease() const { return m_release; }
-	QString getStage() const { return m_stage; }
-	int getBuild() const { return m_build; }
+	int getPatch() const { return m_patch; }
+	QStringList getLabels() const { return m_labels;}
 	CompareType getCompareType() const { return m_compareType; }
 	ProjectVersion setCompareType(CompareType compareType) { m_compareType = compareType; return * this; }
 
-	static int compare(const ProjectVersion& a, const ProjectVersion& b, CompareType c);
+	static int compare(const ProjectVersion& a, const ProjectVersion& b, int c);
```
Veratil, line 52: Couldn't we still leave this as `CompareType`? Might be more verbose using `CompareType::Release`.

Spekular: I wanted to ensure that you could pass in ints other than those in the enum (in other words, >5). From what I can tell, this would require a cast if the parameter type were CompareType, and I'm not sure if it would allow you to pass an int that doesn't correspond to an enum. On the other hand, passing CompareType::Release to a function that expects an int works just fine.

If it's possible to pass ints outside the enum with a cast, I can revert it to CompareType. Otherwise I prefer to leave that option open. (It also occurs to me that I should allow passing -1 to compare the entire version regardless of length)

PhysSong:
It seems like it can't compare some non-standard version we use properly, for example, `1.2.2` vs. `1.2.2.555`.

Spekular:
I think we should change the formatting of those versions to be either  `1.2.3-555` or `1.3.0-555`. This is a valid semantic version and compares in a suitable manner:
  - They're bigger than `1.2.2`
  - They're smaller than `1.2.3-alpha`/`1.3.0-alpha`
  - `1.2.3-555` is less than `1.2.3-600`

I think we can write an upgrade routine for older versions in the constructor. If the first three identifiers are "1.2.2" or less, we convert `x y.z.aa` to `x.y.(z+1)-aa`, and prepend `w` to the prerelease labels.

As for which of the two options to choose, it'd be great if we could have different versions on master and stable. Assuming the latest release is `x.y.z`, it makes a lot of sense to me that:
- Stable builds would be of the form `x.y.(z+1)-commits`
- Master builds would be of the form `x.(y+1).0-commits`

Veratil:
I agree with Spekular. We should start sticking to semver, and we can write a preroutine to capture older formats and convert to what it should be in semver.

Spekular:
Did some testing of the new CMake versioning logic:

| Latest Tag  | Additional Commits | Output Version           |
|-------------|--------------------|--------------------------|
| v1.2.2      | 689                | 1.2.3-689+gc860b7852     |
| v1.2.3      | 0                  | 1.2.3                    |
| v1.3.0-rc.1 | 0                  | 1.3.0-rc.1               |
| v1.3.0-rc.1 | 10                 | 1.3.0-rc.1.10+ga4a6b273a |

FORCE_VERSION flag works as I'd expect, whatever you enter becomes the version. Is there anything else to test? Otherwise this seems ready to merge as far as I can tell.

Spekular:
@PhysSong re-requesting review since I've fixed the issue you mentioned
@tresf requesting review since you're familiar with CMake and had opinions on how we do versioning ;)

include/ProjectVersion.h
DomClark: It is possible to cast an `int` (or other numeric type) to an enum type, even when the value don't correspond to any of the enumerators, but some care is required. Converting a value with more bits than the enum requires results in undefined behaviour. The safe approach is to use the type you want to convert from as the underlying type of the enum (e.g. `enum class CompareType : int`), then any value of that type can be safely converted. Otherwise, you need to make sure that any value you convert requires no more bits to store than the largest enumerator.

DomClark (commented):
I've only reviewed the C++ here - I'll leave the CMake to someone more familiar with it.

DomClark:
Probably out of scope here, but I'll mention it since you're rewriting the comparison logic:

I don't really like part of the current comparison system, where a `CompareType` is set on a `ProjectVersion` prior to comparison. The `CompareType` is a property of the comparison operation, not of the `ProjectVersion`s being compared, and the code should reflect that. (You could argue to the contrary, that a `ProjectVersion` with a `CompareType` represents a range of versions, but I think this is a bad abstraction. If this is the intention, then there should be a `ProjectVersionRange` type, or a conversion from any `ProjectVersion` to another `ProjectVersion` representing the base of its range for a given `CompareType`.)
Furthermore, for `ProjectVersion`s holding `CompareType`s, `operator<` does not define a strict weak order, which means that arbitrary `ProjectVersion`s cannot be used successfully with many standard library containers and algorithms.

src/core/ProjectVersion.cpp
```diff
@@ -27,123 +27,93 @@
...
+	QString metadataStripped = version.split("+").first();
+	// They must have an obligatory initial segement, and may have
+	// optional identifiers prefaced by a '-'. Both parts affect precedence
+	QStringList mainAndLabels = metadataStripped.split("-");
+
+	// The obligatory segment consists of three identifiers: MAJOR.MINOR.PATCH
+	QStringList mainVersion = mainAndLabels.first().split(".");
+	m_major = mainVersion.at(0).toInt();
+	m_minor = mainVersion.at(1).toInt();
+	m_patch = mainVersion.at(2).toInt();
```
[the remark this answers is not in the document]
Spekular: True, but is there a sane way to handle a version like that? I considered letting them default to zero, but really I don't think it's correct to return a valid version for an invalid parameter.

src/core/ProjectVersion.cpp
```diff
@@ -27,123 +27,93 @@
...
+		QString labelB = bLabels.at(i);
+		// If both labels are the same, skip
+		if (labelA == labelB){ continue; }
+		// Else if both labels are numeric, compare them numerically
+		bool aIsNumeric = false, bIsNumeric = false;
+		int numA = labelA.toInt(&aIsNumeric);
+		int numB = labelB.toInt(&bIsNumeric);
+		if (aIsNumeric && bIsNumeric){ return numA - numB; }
+		// Otherwise, compare lexically
+		return labelA.compare(labelB);
```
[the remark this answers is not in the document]
Spekular: Good catch!

src/core/ProjectVersion.cpp
```diff
@@ -27,123 +27,93 @@
...
+
+	// Otherwise, compare as many labels as we can
+	for (int i = 0; i < commonLabels; i++){
+		QString labelA = aLabels.at(i);
+		QString labelB = bLabels.at(i);
+		// If both labels are the same, skip
+		if (labelA == labelB){ continue; }
+		// Else if both labels are numeric, compare them numerically
+		bool aIsNumeric = false, bIsNumeric = false;
+		int numA = labelA.toInt(&aIsNumeric);
```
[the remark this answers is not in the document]
Spekular: Ah, thanks for pointing that out, I totally would have missed it. I suppose I should set `aIsNumeric &= !labelA.startsWith("-")` after the toInt call (and likewise for B).

src/core/ProjectVersion.cpp
```diff
@@ -27,123 +27,93 @@
...
-	m_stage(parseStage(m_version)),
-	m_build(parseBuild(m_version)),
 	m_compareType(c)
 {
+	// Version numbers may have build data, prefixed with a '+',
+	// but this mustn't affect version precedence in comparisons
+	QString metadataStripped = version.split("+").first();
+	// They must have an obligatory initial segement, and may have
+	// optional identifiers prefaced by a '-'. Both parts affect precedence
+	QStringList mainAndLabels = metadataStripped.split("-");
```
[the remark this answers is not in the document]
Spekular: Will fix!

include/ProjectVersion.h
Spekular: Unresolving. I'll set int as the underlying type for CompareType and change the parameter back. With that done I'll also add `CompareType::All = -1`.

Spekular:
@DomClark Great review, thanks! I'll make sure to extend the tests as well when I fix this. (Note to self: every unresolved conversation is an actionable TODO)

> Probably out of scope here, but [...]
> I don't really like part of the current comparison system, where a `CompareType` is set on a `ProjectVersion` prior to comparison. 

Agreed on both counts.

include/ProjectVersion.h
Spekular: Fixed in latest commit. I ended up setting All to `std::numeric_limits<int>::max()`. With `-1`, CompareTypes weren't correctly ordered, which broke the "use the smaller CompareType when comparing" logic. I also made All the new default comparison, since I think that's the desired behavior.

src/core/ProjectVersion.cpp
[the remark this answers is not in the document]
Spekular: Fixed!

src/core/ProjectVersion.cpp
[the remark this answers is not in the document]
Spekular: Solved by padding the QStringList with zeroes if it's shorter than 3 elements.

src/core/ProjectVersion.cpp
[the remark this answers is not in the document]
Spekular: Fixed!

src/core/ProjectVersion.cpp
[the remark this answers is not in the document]
Spekular: Fixed!

Spekular:
All outstanding comments should be fixed in the latest commit. @DomClark I'll request re-review from you to see if you approve of the fixes.

cmake/CMakeLists.txt
```diff
@@ -9,7 +9,8 @@ IF(VERSION_STAGE)
 	SET(CPACK_PACKAGE_VERSION_PATCH         "${CPACK_PACKAGE_VERSION_PATCH}-${VERSION_STAGE}")
 ENDIF()
 IF(VERSION_BUILD)
-	SET(CPACK_PACKAGE_VERSION_PATCH         "${CPACK_PACKAGE_VERSION_PATCH}.${VERSION_BUILD}")
+	math(EXPR CPACK_PACKAGE_VERSION_PATH_BUMP "${CPACK_PACKAGE_VERSION_PATH}+1")
```
tresf, line 12: Minor: `math` versus `MATH` above is inconsistent formatting.

Why bump this again for cpack?  Note, IIRC cpack is only used for Windows when `make package` is invoked; whereas Linux and macOS use shell scripts to name the binaries so this may result in inconsistently numbered binaries.   I'm still confused by the `+1` and then a subsequent `+1`.  I also thought the `+N` was for indicating commits and was already being accounted for, but I'm probably remembering the conversation on Discord incorrectly.

Spekular: The double bump is entirely because I have a poor understanding of how all the different CMake files interact. I'll make sure to fix the inconsistent case in my next commit, good point.

Can you elaborate on this part? I'm not sure I understand:

> I also thought the +N was for indicating commits and was already being accounted for, but I'm probably remembering the conversation on Discord incorrectly.

tresf:
Out of curiosity, I tested [some older upgrade routines](https://github.com/LMMS/lmms/blob/2a66e83f536ed71984054392ed469991de67874e/src/core/DataFile.cpp#L1390-L1398) and it appears they'd fail using proper semver:

https://jubianchi.github.io/semver-check/#/%3C0.4.0-20080622/0.4.0-beta1

Since most of the popular + older LSP tracks seem to be from the 0.4.15 era, the chances of this upgrade routine being need and not triggered are tremendously small/rare, but it's a risk worth identifying.

I was hoping to sort the LSP on project version, but we've added that feature [retroactively](https://github.com/LMMS/lmms.io/issues/59#issuecomment-63764136) so it's not indexed (and we never went back and crawled them).  I'm sure a bash script on the webserver could find this out, but for the purposes of this PR I simply wanted to state the fall throughs as a future reference point if the edge-case is exposed via issue or support.

tresf:
Whoops, I used that online tool wrong.  My points are still valid, but irrelevant as the example I chose passes the online tool.

cmake/CMakeLists.txt
tresf: > Can you elaborate on this part? I'm not sure I understand:
> > I also thought the +N was for indicating commits and was already being accounted for, but I'm probably remembering the conversation on Discord incorrectly.

Put simply, what was the point of bumping this? For example, if `stable-1.2` in july was technically `stable-1.2.1+57` by commit count calculation, why `+1`?  What is this `+1` for?  Shouldn't we let the VCS determine that (not some arbitrary `+1` logic?)

Spekular: Because prerelease versions are less than release versions*, and adding a commit count can only be done with prerelease identifiers**. If the latest tag is 1.2.2, any commits after that have to be bigger than 1.2.2, hence they must at least be a 1.2.3 prerelease.

*Clarification: according to semver, 1.2.2-500 < 1.2.2
**1.2.2.500 is an invalid semver version. It's unreasonable to increment the patch version for every commit, so we have to use prerelease labels

tresf: > The double bump is entirely because I have a poor understanding of how all the different CMake files interact. I'll make sure to fix the inconsistent case in my next commit, good point.

Ah.. well, if it was needed for Windows, it should be warranted.  IIRC, Linux and Mac will just use what's set in the main CMakeLists.txt so they should be OK, I'm just surprised to see that CPack needed this.

Spekular: > Ah.. well, if it was needed for Windows, it should be warranted. IIRC, Linux and Mac will just use what's set in the main CMakeLists.txt so they should be OK, I'm just surprised to see that CPack needed this.

Pardon my ignorance, but just to clarify: The goal is for windows, mac, and linux to all have the same version number. Will the current logic accomplish this? If not, what should I change?

CMakeLists.txt
```diff
@@ -44,7 +44,9 @@ IF(VERSION_STAGE)
 	SET(VERSION     "${VERSION}-${VERSION_STAGE}")
 ENDIF()
 IF(VERSION_BUILD)
-	SET(VERSION     "${VERSION}.${VERSION_BUILD}")
+	MATH(EXPR VERSION_RELEASE_BUMP "${VERSION_RELEASE}+1")
+	SET(VERSION     "${VERSION_MAJOR}.${VERSION_MINOR}.${VERSION_RELEASE_BUMP"})
```
tresf, line 48:
```suggestion
	SET(VERSION     "${VERSION_MAJOR}.${VERSION_MINOR}.${VERSION_RELEASE_BUMP}")
```

cmake/CMakeLists.txt
Spekular, line 12:
```suggestion
	MATH(EXPR CPACK_PACKAGE_VERSION_PATH_BUMP "${CPACK_PACKAGE_VERSION_PATH}+1")
```

tresf: @Spekular order of operations shows that `ADD_SUBDIRECTORY(cmake)` (~L616) comes after `VERSION_RELEASE_BUMP` (~L47) however, the value appears to be set on `VERSION` variable, not `VERSION_PATCH`, so I believe it's a consistency issue with the PR, not necessarily a logical issue with CMake.

https://github.com/LMMS/lmms/blob/6a73c281648c31d171bd9f7fc56f53bf85e84917/CMakeLists.txt#L47

https://github.com/LMMS/lmms/blob/6a73c281648c31d171bd9f7fc56f53bf85e84917/CMakeLists.txt#L616 

CPack uses its own variable for this and some unobvious NSIS integration uses it in the Windows installers (logic I fully plan to remove at some future time) but so as long as we're using it, we should keep it consistent.

https://github.com/LMMS/lmms/blob/6a73c281648c31d171bd9f7fc56f53bf85e84917/cmake/CMakeLists.txt#L7

My recommendation is to do one of the following:

1. Fix `VERSION_RELEASE` directly, get rid of `VERSION_RELEASE_BUMP` variable so it flows downstream from there.<br>
-- OR --
2. Tell CPack to use `VERSION_RELEASE_BUMP` and remove the redundant calculation.


> Will the current logic accomplish this

I haven't tested it, but at a glance, the redundant code will probably work at the expense of duplicated logic.

src/core/ProjectVersion.cpp
```diff
@@ -27,123 +27,105 @@
...
+
+	// Then we can compare as if we care about every identifier
+	if(aMaj != bMaj){ return aMaj - bMaj; }
+	if(aMin != bMin){ return aMin - bMin; }
+	if(aPat != bPat){ return aPat - bPat; }
+
+	// Decide how many optional identifiers we care about
+	int numLabels = qMax(0, limit - 3);
+	auto aLabels = a.getLabels().mid(0, numLabels);
+	auto bLabels = b.getLabels().mid(0, numLabels);
```
DomClark, line 98: `QList::mid` allocates a new list - this could be avoided by taking `numLabels` into account later on instead, where the number of labels to compare is calculated.

DomClark (approved):
C++ looks good to me. I like the attention given to tests.

There a couple of places I've pointed out that make unnecessary copies - this isn't much of a problem with Qt containers, since their data is shared and refcounted behind the scenes, but is something to bear in mind, and would have more of an impact with standard library containers. If you consider this premature optimisation, feel free to ignore it - it's just as functional as it is.

src/core/ProjectVersion.cpp
Spekular: Notes to self just in case I change this, this line would need adjustment:
```
// If everything else matches, the version with more labels is bigger
return aLabels.size() - bLabels.size();
```

cmake/CMakeLists.txt
Spekular: > My recommendation is to do one of the following:
> 
> 1. Fix `VERSION_RELEASE` directly, get rid of `VERSION_RELEASE_BUMP` variable so it flows downstream from there.
>    -- OR --
> 2. Tell CPack to use `VERSION_RELEASE_BUMP` and remove the redundant calculation.

So `lmms/cmake/CMakeLists` can access variables that are set inside `lmms/CMakeLists`? (And it (`lmms/cmake/CMakeLists`) executes after `lmms/CMakeLists`?) If that's the case, I think I'll go with option one. Thanks for you patience in explaining this :)

tresf: > So `lmms/cmake/CMakeLists` can access variables that are set inside `lmms/CMakeLists`? 

Correct, that's how it gets access to `VERSION_RELEASE`.  There's an order of operations that must occur, but that's been happening for a long time and is unlikely to change, quoting:

> `ADD_SUBDIRECTORY(cmake)` (~L616) comes after `VERSION_RELEASE_BUMP` (~L47) (or `VERSION_RELEASE`, etc)

If `ADD_SUBDIRECTORY(cmake)` where done prior to setting that variable, the variable would never propagate.  This is  why defining version information so early in the CMakeLists.txt is so important.

src/core/ProjectVersion.cpp
```diff
@@ -27,123 +27,105 @@
...
+	// We can only compare identifiers if both versions have them
+	int commonLabels = qMin(aLabels.size(), bLabels.size());
+	// If one version has optional labels and the other doesn't,
+	// the one without them is bigger
+	if (commonLabels == 0){ return bLabels.size() - aLabels.size(); }
+
+	// Otherwise, compare as many labels as we can
+	for (int i = 0; i < commonLabels; i++){
+		QString labelA = aLabels.at(i);
+		QString labelB = bLabels.at(i);
```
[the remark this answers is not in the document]
Spekular: Note to self: if you do this you need to use `*labelA == *labelB`

Spekular:
Testing again with changes that I haven't pushed yet:

| Latest Tag | Additional Commits | Output Version | Valid? |
|-------------|--------------------|--------------------------|----|
| v1.2.2 | 692 | 1.2.3-692+g30a03465c | :heavy_check_mark: |
| **v1.2.2.300** | 0 | 1.2.2.300 (Changed since last test) | :x: |
| **v1.2.2.300** | 2 | 1.2.3-2+g30a03465c | :x: |
| v1.2.3 | 0 | 1.2.3 | :heavy_check_mark: |
| *v1.3.0-rc1* | 0 | 1.3.0-rc1 | :o: |
| *v1.3.0-rc1* | 5 | 1.3.0-rc1.5+g30a03465c | :o: |
| v1.3.0-rc.1 | 0 | 1.3.0-rc.1 | :heavy_check_mark: |
| v1.3.0-rc.1 | 10 | 1.3.0-rc.1.10+g30a03465c | :heavy_check_mark: |

src/core/ProjectVersion.cpp
Spekular: I skipped this change because I felt that it convoluted the logic a bit. If it turns out that version comparisons are slowing things down significantly there are definitely optimizations to be had!

tresf:
> Merge?

I haven't tested it locally, but all of my concerns were addressed, so 👍 from me.

include/ProjectVersion.h
```diff
@@ -36,16 +39,16 @@
...
+	ProjectVersion(QString version, CompareType c = All);
+	ProjectVersion(const char * version, CompareType c = All);
 
 	int getMajor() const { return m_major; }
 	int getMinor() const { return m_minor; }
-	int getRelease() const { return m_release; }
-	QString getStage() const { return m_stage; }
-	int getBuild() const { return m_build; }
+	int getPatch() const { return m_patch; }
+	const QStringList& getLabels() const { return m_labels;}
```
PhysSong, line 51: Inconsistent whitespace style inside braces.

include/ProjectVersion.h
```diff
@@ -54,11 +57,10 @@ class ProjectVersion
...
 	QString m_version;
-	int m_major;
-	int m_minor;
-	int m_release;
-	QString m_stage;
-	int m_build;
+	int m_major = 0;
+	int m_minor = 0;
+	int m_patch = 0;
+	QStringList m_labels = QStringList();
```
PhysSong, line 63: You don't need ` = QStringList()` because it will be default-initialized if you do nothing.

PhysSong:
LGTM

include/ProjectVersion.h
Spekular: Wait, C++ handles uninitialized objects nicely but not primitives? Completely opposite to java, weird.

include/ProjectVersion.h
Spekular, line 51:
```suggestion
	const QStringList& getLabels() const { return m_labels; }
```

include/ProjectVersion.h
Spekular, line 63:
```suggestion
	QStringList m_labels;
```

Spekular (commented):
Changes suggested by @PhysSong

Issue #3816, opened 2017-09-21
FM or heavy PM in TripleOscillator makes outputs odd for some target waveforms


PhysSong:
Step to reproduce:
1. Add a TripleOscillator.
2. Set Osc 1 waveform to one of triangle/saw/moog saw/exponential.
3. Set Osc1 + Osc2 to FM.
4. Peak amplitude is larger than one without FM(Osc2 and Osc3 volume is zero).

Example: [oscbug.mmp.zip](https://github.com/LMMS/lmms/files/1320928/oscbug.mmp.zip)

PhysSong:
I think it is an issue of `Oscillator` class.
Most of other instrument don't behave like our 3osc does.

musikBear:
Could this be related to the 3oc problem here?
#2047
As written here https://github.com/LMMS/lmms/issues/2047#issuecomment-277440706
The fix had to be omitted, because of #3292.
A follow up question would be: Should 3oc be changed? What about backward issues? 
I think this issue could be compared to FLs 3oc versus the Wasp. Wasp is in essence a 3oc too, so imho it would not be completely crazy to have current 3oc preserved, but add a new *'3oc-2000'*, where all oddities in current 3oc are omitted. Opinions?

SecondFlight:
I think this should be closed or at least revisited per https://github.com/LMMS/lmms/pull/4397.

Issue #2047, opened 2015-05-12
TripleOscillator: Oscillators are getting out of sync


musikBear:
I have spend some time trying to figure out, why lmms sounds less 'punchy/ agressive' than other daw's
during that i have seen something strange
To reproduce
insert a std 3oc
set all 3 osc to saw
insert a 18 bar long note in D3
play
The sound variates in a sweep-like fashion:
![paste](https://cloud.githubusercontent.com/assets/6662066/7587156/79f8c944-f8b4-11e4-8856-e39e8fb08444.PNG)

Why?
There are no 
- envelope settings
- filters
- reso

why is this note not completely uniform in spektrum?
If i do exactly the same with 3oc in fls-demo, i get a completely uniform spectrum

![paste](https://cloud.githubusercontent.com/assets/6662066/7587173/a4c4afee-f8b4-11e4-8bfa-75d414a263d3.png)

Sti2nd:
> why lmms sounds less 'punchy/ agressive' than other daw's

Maybe you haven't removed the default limiter from FL? I haven't got a clue how synthesis really work, but I can easily imagine that there are different algorithms for making a sine wave... TripleOscillator for LMMS is not the same instrument as 3OSC for FL!

michaelgregorius:
@Sti2nd I think it's really a bug (if it is not an intended feature).

To get an idea of what might be causing this behaviour and to simplify the conditions I have set all three oscillators to play a sine wave instead of a saw. The same effect can still be observed which means that we can likely eliminate the aliasing in the saw oscillators as a possible cause. My guess is that the three oscillators go out of phase over time and that the observed lowering of amplitude is the effect of some destructive interference due to the phasing.

Here is the sum of all three sines (fundamental + octave + 2. octave) at the beginning of the note (also the first samples of the test render):
![2047-start](https://cloud.githubusercontent.com/assets/9293269/7595596/41cc95ee-f8e8-11e4-95ff-8736135a6c35.png)

Please note that the waveform is symmetric at this point in time and that it looks like expected. And now here is the waveform towards the end of my rendered wav file:
![2047-end](https://cloud.githubusercontent.com/assets/9293269/7595612/5d37a300-f8e8-11e4-8e86-79007c38075f.png)

Under normal conditions it should look like in the first screenshot.

[...]

PR #5651, opened 2020-08-22
Fix for issue #3816 - FM or heavy PM in TripleOscillator makes output…

DigArtRoks:
…s odd for some target waveforms.

The internal waveforms of the class Oscillator produces the wrong amplitude when the input is a
negative phase. When doing PM or FM, negative phases may occur. When a negative phase is e.g. passed
to the the saw sample, it produces values less than -1.0, hence going out of range.

Converted all fraction calls to absFraction calls.

Removed the +2 in the function Oscillator::recalcPhase. The comment here was that it was needed to avoid
negative phases in case of PM. But by converting fraction to absFraction in the waveforms, negative phases
are not an issue anymore. On top of that the m_phase variable gains about 2 extra bits in precision.
As side effect of that, it improves the behaviour of the issue #2047 - TripleOscillator: Oscillators are getting out of sync.
Though I did not investigate it thoroughly over different notes and samplerates.

Add documentation to the fraction and absFraction functions in lmms_math.h as it was not immediately clear by the name what the
functions do. Correct the implementation of the functions in case the flag __INTEL_COMPILER is set. (floorf rounds always down).

### DIFF
--- a/include/Oscillator.h
+++ b/include/Oscillator.h
@@ -97,7 +97,7 @@ public:
 
 	static inline sample_t triangleSample( const float _sample )
 	{
-		const float ph = fraction( _sample );
+		const float ph = absFraction( _sample );
 		if( ph <= 0.25f )
 		{
 			return ph * 4.0f;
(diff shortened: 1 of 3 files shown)


PhysSong (approved):

Spekular (approved):

DomClark:
Code looks good, and I approve of the changes in principle, but what effect does this have on existing presets and projects? TripleOscillator is one of the more commonly used synths in LMMS, so we don't want to change its behaviour too much.

DigArtRoks:
FYI - The fix will affect only the sound when:
- FM is used, as this may generate negative phases for the modulated oscillator
- or PM is used, but the impact is less, because in the function recalcPhase() they already did some compensation (+2) to avoid negative phases. (But this is no longer needed as the fix now uses absFraction for all the primitive waveforms.)
- the volume of the modulating oscillator is turned high
- the waveform of the modulated oscillator is different from sine wave and custom wave (these 2 handled negative phases already correctly before the fix)
The resulting sound will be less distorted compared to before.

DomClark:
It's probably fine then, since turning the volume of the modulating oscillator up high tends to produce objectionable noise anyway. I checked all the built-in presets and couldn't notice any difference between this PR and 1.2.2, so I'm happy with this.

DomClark (approved):

PR #5653, opened 2020-08-23
Update portsmf + some MIDI loading updates

Veratil:
This updates portsmf to the latest commit from sourceforge, r234. I downloaded the snapshot and copied all the files we have over. A lot of changes here are function parameters adding `const`.

With the portsmf update brought an error when trying to use our `Alg_seq` constructor with `stringstream`. The constructor takes a `std::istream` which I tried to convert the `stringstream` to `stringistream`, but everything kept yelling that a constructor couldn't be found (during runtime). So I converted it to use `ifstream`, and everything calmed down and started working again. I've tested loading about 10 MIDIs I have saved for testing purposes, and all seemed to load without crashing or missing tracks (maybe) after the next change below.

I converted the static 256 count array of `smfMidiChannel` to an `unordered_map`. This is pretty much a drop in replacement as its `operator[]` will create a new element if none exists. This change has let me load previously unloadable MIDI files (found in Discord from others reporting it won't load). There are still some fixes to do, such as making sure all tracks are actually loaded. I find some MIDIs will load and have extra tracks created, but no data. I'm not certain if this is an error or a result of looping through a track that doesn't exist in the MIDI.

Finally, I added the `seqnames` attribute check since I've noticed before that track names won't load on some that use that attribute. This check is pretty much the same that's used in the allegro code.

### DIFF
--- a/plugins/MidiImport/MidiImport.cpp
+++ b/plugins/MidiImport/MidiImport.cpp
@@ -31,6 +31,7 @@
 #include <QProgressDialog>
 
 #include <sstream>
+#include <unordered_map>
 
 #include "MidiImport.h"
 #include "TrackContainer.h"
@@ -305,7 +306,7 @@ public:
 
 bool MidiImport::readSMF( TrackContainer* tc )
 {
-
+	const int MIDI_CC_COUNT = 128 + 1; // 0-127 (128) + pitch bend
 	const int preTrackSteps = 2;
 	QProgressDialog pd( TrackContainer::tr( "Importing MIDI-file..." ),
 	TrackContainer::tr( "Cancel" ), 0, preTrackSteps, gui->mainWindow() );
@@ -315,10 +316,7 @@ bool MidiImport::readSMF( TrackContainer* tc )
 
 	pd.setValue( 0 );
 
-	std::stringstream stream;
-	QByteArray arr = readAllData();
-	stream.str(std::string(arr.constData(), arr.size()));
-
+	std::istringstream stream(readAllData().toStdString());
 	Alg_seq_ptr seq = new Alg_seq(stream, true);
 	seq->convert_to_beats();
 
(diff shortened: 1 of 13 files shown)


plugins/MidiImport/MidiImport.cpp
```diff
@@ -315,19 +316,21 @@ bool MidiImport::readSMF( TrackContainer* tc )
...
 	pd.setValue( 0 );
 
-	std::stringstream stream;
-	QByteArray arr = readAllData();
-	stream.str(std::string(arr.constData(), arr.size()));
-
-	Alg_seq_ptr seq = new Alg_seq(stream, true);
+	std::ifstream iFile(file().fileName().toStdString().c_str(), std::ios::in | std::ios::binary);
+	Alg_seq_ptr seq = new Alg_seq(iFile, true);
+	iFile.close();
```
PhysSong, line 321: Please don't change this unless the updated portsmf complains. I changed it in #4401 due to some issues on Windows related to file name encoding.

PhysSong:
You missed a `for` loop at the end which still uses `256`.

plugins/MidiImport/MidiImport.cpp
PhysSong: I missed the comment above. I think you can still replace `std::stringstream` with `std::istringstream` without further changes.

Veratil: I tried, and during runtime it kept complaining that it couldn't find a valid constructor. I'll give it a shot again though and see if I can get it working.

Veratil: Somehow it works now. I don't know what happened before. 😩

While looking at it I figured out the `ifstream` way actually didn't need to add the `.c_str()` part for the filename. Do you think using a `std::string` for the filename would handle the encoding issues you found? I think using an `ifstream` is a little cleaner code, and less converting from `QByteArray` to `std::string` when we can just pass the file directly.

PhysSong: On Windows, the standard C++ I/O can't handle file names outside the local code page, that's why I used `QFile`. If we can wrap it with a better way, it will be great.

Veratil: Okay I've cleaned up the code to be simpler and more readable, and still use the QFile method.

PhysSong:
I saw that `MidiImport::tryImport()` references `chs[9]` at the end. This should probably be changed to take mod 16 for all existing channels.
The other part of `MidiImport` change looks good.

PhysSong (approved):

PR #5660, opened 2020-08-29
Improvements to the upgrade routines

IanCaio:
As previously discussed on the Discord server, both DataFile and ConfigManager should use a different version number than the release versions to make it easier to add new upgrade routines and bump new files. This draft PR is a work in progress to implement those version numbers and work on the backwards compatibility with the old upgrading methods.


src/core/ConfigManager.cpp
```diff
@@ -131,17 +133,17 @@ void ConfigManager::upgrade()
 		return;
 	}
 
-	ProjectVersion createdWith = m_version;
-	
-	if (createdWith.setCompareType(ProjectVersion::Build) < "1.1.90")
+	if( configVersion() < 1 )
```
Spekular, line 136: As suggested by @Veratil on Discord, we can use a switch fallthrough here instead:
```suggestion
	switch (configVersion()) {
		case 0: upgrade_0();
		case 1: upgrade_1();
	}
```

src/core/ConfigManager.cpp
```diff
@@ -400,11 +403,21 @@ void ConfigManager::loadConfigFile(const QString & configFile)
...
-			// Cache the config version for upgrade()
+			// Cache LMMS version
 			if (!root.attribute("version").isNull()) {
 				m_version = root.attribute("version");
 			}
 
+			// Get the version of the configuration file (for upgrade purposes)
+			if( root.attribute("configversion").isNull() )
+			{
+				m_configVersion = -1; // No configversion attribute found
```
Spekular, line 414: I think you should set `m_configVersion = configVersion();` here instead, and rework `configVersion()` as described in my comment there.

src/core/ConfigManager.cpp
```diff
@@ -565,6 +578,7 @@ void ConfigManager::saveConfigFile()
 
 	QDomElement lmms_config = doc.createElement("lmms");
 	lmms_config.setAttribute("version", m_version);
+	lmms_config.setAttribute("configversion", configVersion());
```
Spekular, line 581: If you apply the other changes mentioned, this needs to be changed to:
```suggestion
	lmms_config.setAttribute("configversion", m_configVersion);
```

src/core/ConfigManager.cpp
```diff
@@ -673,3 +687,28 @@ void ConfigManager::initDevelopmentWorkingDir()
...
+			m_configVersion = 1;
+		}
+		else
+		{
+			m_configVersion = 2;
+		}
+	}
+
+	return m_configVersion;
+}
```
Spekular, line 714:
```suggestion
// If configversion is not present, we will convert the LMMS version to the appropriate
// configuration file version for backwards compatibility.
const int ConfigManager::legacyConfigVersion()
{
	ProjectVersion createdWith = m_version;
	createdWith.setCompareType(ProjectVersion::Build);
	if (createdWith < "1.1.90") { return 0; }
	else if (createdWith < "1.1.91") { return 1; }
	else { return 2; }
}
```

src/core/ConfigManager.cpp
```diff
@@ -151,6 +153,7 @@ void ConfigManager::upgrade()
 
 	// Bump the version, now that we are upgraded
 	m_version = LMMS_VERSION;
+	// No need to bump the m_configVersion, because it gets bumped in the configVersion() method
```
Spekular, line 156:
```suggestion
	m_configVersion = CONFIG_VERSION;
```

Spekular (changes requested):
I would prefer to handle m_configVersion a little differently:
- Explicitly bump the version by setting it to a named constant. This makes it easy to know how one should bump the version in the future.
- Make the function currently known as configVersion cleaner. By (1) removing all side effects and (2) narrowing it's responsibility. We can use m_version directly throughout the code as long as we ensure that it's set correctly on load and upgrade.

src/core/ConfigManager.cpp
Spekular: To be clear, you would need to add `CONFIG_VERSION = 2;` somewhere in ConfigManager. Probably makes sense to put it in the header?

Spekular:
`switch` and `if/else` statements work fine for ConfigManager (for now) since there are so few upgrade routines, but when updating DataFile I believe it's better to use sequences. I believe it should look something like [this](https://gist.github.com/Spekular/2f8229947371f7490173083bfd1a975a) (untested for now, sorry).

IanCaio:
I tried to cover everything you mentioned on this last commit. I'll be starting `DataFile.cpp` as soon as I get some more time to code, but if I missed something on the configuration file let me know. I was thinking that this will be a somewhat annoying PR to test because changes are not very visible. Maybe I can add some warnings on what upgrade routines are running for the testing phase and then remove them when we see that everything is working as supposed.

I like the function vector solution for the DataFile upgrade. I'll try to implement it there. After I'm done I can do the same on ConfigManager for consistency between both, and also to leave it ready for when we have more upgrade routines.

src/core/DataFile.cpp
```diff
@@ -49,6 +49,15 @@
...
 
 
+// Vector with all the upgrade methods
+const std::vector<void(DataFile::*)()> DataFile::m_upgradeMethods = {
+	&DataFile::upgrade_0_2_1_20070501, &DataFile::upgrade_0_2_1_20070508, &DataFile::upgrade_0_3_0_rc2,
+	&DataFile::upgrade_0_3_0, &DataFile::upgrade_0_4_0_20080104, &DataFile::upgrade_0_4_0_20080118,
+	&DataFile::upgrade_0_4_0_20080129, &DataFile::upgrade_0_4_0_20080409, &DataFile::upgrade_0_4_0_20080607,
+	&DataFile::upgrade_0_4_0_20080622, &DataFile::upgrade_0_4_0_beta1, &DataFile::upgrade_0_4_0_rc2,
+	&DataFile::upgrade_1_0_99, &DataFile::upgrade_1_1_0, &DataFile::upgrade_1_1_91, &DataFile::upgrade_1_2_0_rc3,
+	&DataFile::upgrade_1_3_0
```
Veratil, line 59: I feel like this should be formatted as a long list to be more readable.
```suggestion
	&DataFile::upgrade_0_2_1_20070501,
	&DataFile::upgrade_0_2_1_20070508,
	&DataFile::upgrade_0_3_0_rc2,
	&DataFile::upgrade_0_3_0,
	&DataFile::upgrade_0_4_0_20080104,
	&DataFile::upgrade_0_4_0_20080118,
	&DataFile::upgrade_0_4_0_20080129,
	&DataFile::upgrade_0_4_0_20080409,
	&DataFile::upgrade_0_4_0_20080607,
	&DataFile::upgrade_0_4_0_20080622,
	&DataFile::upgrade_0_4_0_beta1,
	&DataFile::upgrade_0_4_0_rc2,
	&DataFile::upgrade_1_0_99,
	&DataFile::upgrade_1_1_0,
	&DataFile::upgrade_1_1_91,
	&DataFile::upgrade_1_2_0_rc3,
	&DataFile::upgrade_1_3_0
```

include/ConfigManager.h
```diff
@@ -266,8 +272,8 @@ class LMMS_EXPORT ConfigManager : public QObject
 	ConfigManager(const ConfigManager & _c);
 	~ConfigManager();
 
-	void upgrade_1_1_90();
-	void upgrade_1_1_91();
+	void upgrade_0();
+	void upgrade_1();
```
Veratil, line 276: I don't think we need to rename these.

src/core/DataFile.cpp
IanCaio: True! I'll update it

include/ConfigManager.h
IanCaio: From now on I believe this new naming will be used (`upgrade_` + config version, or something similar) because it makes it more explicit the methods being called per version. I thought changing the older ones to match this naming as well would make the code more organized, but I can revert the change if you think I should keep the old names.

IanCaio:
Reverted the changes on the upgrade method names from `ConfigManager.cpp` and formatted vector initializers from the upgrade methods and version strings from `DataFile.cpp` so they are more readable.

I'm still in favor of changing the upgrade method names on both `ConfigManager` and `DataFile`, to make them consistent with the naming we will use from now on, but I reverted it for now so it can discussed first.

src/core/DataFile.cpp
Spekular: Could we at least list, say, two per row? We can leave this as is until we're sure that the function names are final, but personally I'd like for it to be more compact. I think the comment, variable name, and the way the vector is used later make it pretty clear that it just lists all upgrade routines in order. I'm not sure it's particularly necessary to actually read the whole list for the sake of comprehending the code. Most modifications to it should just be appending another method at the end.

That said, if we want it to be a bit neater despite the compactness, perhaps we could align the columns with spaces. So

```
std::vector ... {
[Tab]upgradeRoutine,[Spaces       ] upgradeRoutine3,
[Tab]anotherUpgradeRoutine,[Spaces] upgradeRoutine4
}
```

Spekular:
I'm in favor of naming the upgrade routines more descriptively than as simple number, at least going forward. Some of the existing ones do a lot, or are tricky to decipher, so I think I'd prefer leaving them alone unless you can think of a good name.

`upgrade_1_1_0`, for example, could be named something like `upgrade_addFxSends`.

Spekular:
Code LGTM. I lean towards slightly more compact lists of upgrades, and we could rename some of the old routines to be more descriptive, but I don't think that should hold up a merge.

As for things that should hold up a merge: testing. Have you tested yet? Do you want assistance with testing this, and if so what kind?

Veratil:
> 
> 
> I'm in favor of naming the upgrade routines more descriptively than as simple number, at least going forward. Some of the existing ones do a lot, or are tricky to decipher, so I think I'd prefer leaving them alone unless you can think of a good name.
> 
> `upgrade_1_1_0`, for example, could be named something like `upgrade_addFxSends`.

I agree. This could be the same with the `ConfigManager` functions as well.



> 
> 
> I lean towards slightly more compact lists of upgrades, and we could rename some of the old routines to be more descriptive, but I don't think that should hold up a merge.

I'd be okay with two per line with descriptive rename. 👍 The previous way was just a huge near unreadable eyesore.

IanCaio:
> As for things that should hold up a merge: testing. Have you tested yet? Do you want assistance with testing this, and if so what kind?

Last week time for coding has been a bit scarce, so I haven't tested it yet, if you happen to be available it would be great! I had the following idea for testing:

- I'd Add a `qWarning`s on each upgrade routine to give us visual cue to what is being called.
- We'd need to create one project file for each range of upgrades (preferentially inside the range and also on the boundaries of the ranges). Same for configuration files.
- Open the projects/Replace the configuration file. Check which upgrade routines are being called.

I could add the warnings to each upgrade routine tonight and update the PR to generate the testing builds if that sounds good, then I'd just remove the warnings when everything is tested. When the project files and configuration files are done they could be posted here to make it easier for others to test too. Not sure I'll be able to do them tonight before going to sleep though.

> I'd be okay with two per line with descriptive rename. :+1: The previous way was just a huge near unreadable eyesore.

I can also format the lists that way while I'm on it!

IanCaio:
I got ahead and updated the PR with the mentioned changes:

- Added `qWarning`s on every upgrade method to help with testing.
- Reformatted the upgrade methods list and versions list.
- Removed some unnecessary white space I found by accident on a line.

Spekular:
I've created a set of test files by saving an .mmp and replacing the creatorversion:

[TestFiles.zip](https://github.com/LMMS/lmms/files/5179791/TestFiles.zip)

I'll try testing with the buildbot AppImage and see how it goes.

Spekular:
From my testing, it appears that the upgrade routines are being called correctly.

@IanCaio now seems like a good time to update ConfigManager to use the same vector-of-function-pointers method, if you still want to do that:

> I like the function vector solution for the DataFile upgrade. I'll try to implement it there. After I'm done *I can do the same on ConfigManager for consistency between both*, and also to leave it ready for when we have more upgrade routines.

After that I'd vote that we merge this.

IanCaio:
@Spekular Thanks for uploading the test files! I've been on a bit of a rush lately, but I'll try to change `ConfigManager.cpp` to use vector pointers too as soon as possible to get it ready for testing and merge, either still today or tomorrow!

IanCaio:
Changed `ConfigManager.cpp` to use a vector of upgrade methods too. Debug messages are still present in the code. It was tested with 3 different files here (versions `1.1.89`, `1.1.90` and `1.1.91`) and it behaved as expected. I'll later test the `DataFile` loading too with Spekular's files.

[ConfigFiles.zip](https://github.com/LMMS/lmms/files/5184506/ConfigFiles.zip)

Spekular:
Since both of our tests indicate that this is working and i don't see any outstanding issues, I'll merge this on Friday (UTC+2) if no one objects before then. If someone else does a passing review, maybe we can merge earlier.

Spekular (approved):
LGTM, thanks for getting this done! :)

IanCaio:
No problem! :smile: 
Just remember me to make a last commit removing the debug messages before the merge

include/ConfigManager.h
```diff
@@ -219,6 +218,10 @@ class LMMS_EXPORT ConfigManager : public QObject
 		return m_version;
 	}
 
+	// Used when the configversion attribute is not present in a configuration file.
+	// Returns the appropriate config file version based on the LMMS version.
+	const unsigned int legacyConfigVersion();
```
DomClark, line 223: Did you mean `unsigned int legacyConfigVersion() const;`? (Same in `DataFile.h`.)

src/core/ConfigManager.cpp
```diff
@@ -131,17 +140,17 @@ void ConfigManager::upgrade()
 		return;
 	}
 
-	ProjectVersion createdWith = m_version;
-	
-	if (createdWith.setCompareType(ProjectVersion::Build) < "1.1.90")
-	{
-		upgrade_1_1_90();
-	}
+	using upgradeMethod = void(ConfigManager::*)();
```
DomClark, line 143: According to our coding conventions, types must start with an uppercase letter. Also, this line can be moved into the header, so it can be used in the declaration of `m_upgradeMethods` too.

src/core/DataFile.cpp
```diff
@@ -1542,3 +1514,29 @@ void findIds(const QDomElement& elem, QList<jo_id_t>& idList)
...
+
+	// Vector of all versions that have upgrade routines.
+	std::vector<QString> upgradeVersions = {
+		"0.2.1-20070501"   ,   "0.2.1-20070508"   ,   "0.3.0-rc2",
+		"0.3.0"            ,   "0.4.0-20080104"   ,   "0.4.0-20080118",
+		"0.4.0-20080129"   ,   "0.4.0-20080409"   ,   "0.4.0-20080607",
+		"0.4.0-20080622"   ,   "0.4.0-beta1"      ,   "0.4.0-rc2",
+		"1.0.99-0"         ,   "1.1.0-0"          ,   "1.1.91-0",
+		"1.2.0-rc3"        ,   "1.3.0"
+	};
```
DomClark, line 1533: This can be const and static, since it isn't modified and is the same for each call. Then it could also contain `ProjectVersion`s instead, so they don't have to be re-parsed for every comparison.

src/core/DataFile.cpp
```diff
@@ -1542,3 +1514,29 @@ void findIds(const QDomElement& elem, QList<jo_id_t>& idList)
...
+
+	// Get an iterator pointing at the first upgrade we need to run (or at upgrades.end() if there is no such upgrade)
+	auto firstRequiredUpgrade = std::find_if( upgradeVersions.begin(), upgradeVersions.end(),
+		[creator](QString version){ return creator < version; }
+	);
```
DomClark, line 1538: Nice use of standard library algorithms. However, there's a specific one for this purpose: [`std::upper_bound`](https://en.cppreference.com/w/cpp/algorithm/upper_bound).

src/core/ConfigManager.cpp
```diff
@@ -673,3 +697,25 @@ void ConfigManager::initDevelopmentWorkingDir()
...
+	}
+	else if( createdWith < "1.1.91" )
+	{
+		return 1;
+	}
+	else
+	{
+		return 2;
+	}
+}
```
DomClark, line 721: I'm not sure whether this should be changed to match the logic in `DataFile::legacyConfigVersion` or not. What do you think?

IanCaio:
Thanks for the review @DomClark ! Requests were addressed on this last commit. I've done a quick test for the configuration files upgrades and loaded a couple of the test files Spekular posted, it seems like it's still working fine. The debugging warnings are still there in case someone else wish to test.

src/core/ConfigManager.cpp
DomClark: We discussed this on Discord and the decision was to leave it as is.

DomClark (approved):

src/core/ConfigManager.cpp
```diff
@@ -38,6 +38,11 @@
 #include "lmmsversion.h"
 
 
+// Vector with all the upgrade methods
+const std::vector<void(ConfigManager::*)()> ConfigManager::UPGRADE_METHODS = {
```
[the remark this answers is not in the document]
IanCaio: Done! I think when I tried to use the qualified alias before I made a typo and wrote `ConfigManager::UpgradeMethods` instead, then I got an `is not a member of ConfigManager` error. Without the typo it compiled fine.

Spekular:
Remove debug calls and merge?

DomClark:
Is it worth testing this with #5636 merged first?

Spekular:
> Is it worth testing this with #5636 merged first?

I suppose so. I'll merge #5636 as soon as tests pass for the latest commit, so it does make sense to test again after that.

Spekular:
> Is it worth testing this with #5636 merged first?

Tested! Everything still works :)

PR #5661, opened 2020-08-29
 Fix for Mixer volume percentage labels are off by a factor of 100

DigArtRoks:
(LMMS#2340)

Add m_conversionFactor to the AutomatableModelView. This factor will be applied to the model->value when displaying
it in the contextmenu of the control for the reset and copy actions. The factor will be applied when copying the value to
the clipboard. When pasting from the clipboard, the value will be divided by the factor.

Remove the model->displayValue() calls when updating the reset/copy/paste action text labels as this gives e.g. in the
Equalizer the wrong action text for the Frequency knobs.

In the Fader class, remove the m_displayConversion variable but rather use the new m_conversionFactor variable.
Rewire the setDisplayConversion() function to set the m_conversionFactor to 1.0 or 100.0.

Faders in FxMixer show now the correct context menu. Copying and pasting values between faders or even volume knobs
in tracks shows consistent behavior. Other faders (like in Eq) show the old behavior.

Replace PR #5654

### DIFF
--- a/include/AutomatableModelView.h
+++ b/include/AutomatableModelView.h
@@ -78,7 +78,7 @@ protected:
 
 	QString m_description;
 	QString m_unit;
-	float   m_conversionFactor;  // Factor to be applied when the m_model->value is displayed
+	float m_conversionFactor; // Factor to be applied when the m_model->value is displayed
 } ;
 
 
--- a/include/AutomatableModelView.h
+++ b/include/AutomatableModelView.h
@@ -69,12 +69,16 @@ public:
 
 	void addDefaultActions( QMenu* menu );
 
+	void setConversionFactor( float factor );
+	float getConversionFactor();
+
 
 protected:
 	virtual void mousePressEvent( QMouseEvent* event );
 
 	QString m_description;
 	QString m_unit;
+	float   m_conversionFactor;  // Factor to be applied when the m_model->value is displayed
 } ;
 
 
--- a/src/gui/widgets/Fader.cpp
+++ b/src/gui/widgets/Fader.cpp
@@ -72,7 +72,6 @@ Fader::Fader( FloatModel * _model, const QString & _name, QWidget * _parent ) :
 	m_persistentPeak_R( 0.0 ),
 	m_fMinPeak( 0.01f ),
 	m_fMaxPeak( 1.1 ),
-	m_displayConversion( true ),
 	m_levelsDisplayedInDBFS(false),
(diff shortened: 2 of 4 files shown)


include/AutomatableModelView.h
```diff
@@ -69,12 +69,16 @@ class LMMS_EXPORT AutomatableModelView : public ModelView
...
+	void setConversionFactor( float factor );
+	float getConversionFactor();
+
 
 protected:
 	virtual void mousePressEvent( QMouseEvent* event );
 
 	QString m_description;
 	QString m_unit;
+	float   m_conversionFactor;  // Factor to be applied when the m_model->value is displayed
```
PhysSong, line 81: I think you might remove some whitespaces, though not necessary.
```suggestion
	float m_conversionFactor; // Factor to be applied when the m_model->value is displayed
```

src/gui/widgets/Fader.cpp
```diff
@@ -330,14 +317,14 @@ void Fader::setPeak_R( float fPeak )
 // update tooltip showing value and adjust position while changing fader value
 void Fader::updateTextFloat()
 {
-	if( ConfigManager::inst()->value( "app", "displaydbfs" ).toInt() && m_displayConversion )
+	if( ConfigManager::inst()->value( "app", "displaydbfs" ).toInt() && ( m_conversionFactor == 100.0 ) )
```
PhysSong, line 320: The parentheses around the comparison is not needed.

PhysSong: `m_conversionFactor == 100.0` doesn't always imply it's a volume fader, but this look out of scope of this PR.

include/AutomatableModelView.h
PhysSong: I'm also not sure which class should hold this property(model, or view?). This question is equivalent to: "Does 1.0 for the mixer volume always mean 100 for end-users?". I'm guessing it's preferable to keep stuff as-is, but I'm asking for other opinions.

PhysSong:
It generally looks good. Let's wait for other developer's feedback.

Spekular (approved):
Seconding PhysSong's comments.

DigArtRoks:
Additional commit to take into account the review comments concerning coding style of PhysSong.

DomClark (approved):
LGTM. FYI, the changes to `AutomatableModelView.cpp` will conflict with #5627.

PR #5677, opened 2020-09-16
LadspaManager: Skip empty entries when looking for plugins

### LINKED ISSUE #2569: Opening mmpz from within zip file crashes LMMS
Double clicking a zipped project results in this:

edit: To be more specific: this is about double clicking the file inside a zip package (so, first open the zip file and then double click the containing project to reproduce).

image

PhysSong:
Fixes #2569.

### DIFF
--- a/src/core/LadspaManager.cpp
+++ b/src/core/LadspaManager.cpp
@@ -59,6 +59,8 @@ LadspaManager::LadspaManager()
 	for( QStringList::iterator it = ladspaDirectories.begin(); 
 			 		   it != ladspaDirectories.end(); ++it )
 	{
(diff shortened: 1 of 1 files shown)


Spekular:
Could we pass `Qt::SkipEmptyParts` as an argument to the `split` calls that populate `ladspaDirectories` instead? If there are other places where an empty path could be added, I suppose the check against empty paths needs to be kept.

PhysSong:
That will also work. I simply copied the logic from `PluginFactory` because it also does this.

Spekular:
Then I suppose it's more consistent this way.

Spekular (approved):

PR #5678, opened 2020-09-18
Don't define NOMINMAX if it's already defined

Spekular:
Co-authored-by: Dominic Clark <mrdomclark@gmail.com>


Spekular:
Being eager to fix this CI error via direct commit just led to more CI errors, so let's go with a PR this time. Bypassing CI when attempting to fix CI was probably a dumb move...

PR #5295, opened 2019-10-31
Rename MidiTime to TimePos

lukas-w:
I redid #4897 because of the conflicts after merging #4878. Also includes `TimePos.h`'s include guard, `NotePlayHandle::processMidiTime` and some parameter names.

PR #4897, opened 2019-03-16
Rename MidiTime to TimePos

M374LX:
Closes #4866.

PR #5684, opened 2020-09-25
Rename MidiTime to TimePos

### LINKED ISSUE #4866: Rename MidiTime class
The class ```MidiTime``` is not used exclusively for MIDI. So, it could be renamed to something like ```TimePos```. For the same reason, the CPP file should not be located in the ```src/core/midi``` directory.

M374LX:
Redoing  #4897 and #5295 after both were unsuccessfully closed.

Closes #4866.

### DIFF
--- a/include/TimePos.h
+++ b/include/TimePos.h
@@ -1,6 +1,6 @@
 /*
- * MidiTime.h - declaration of class MidiTime which provides data type for
- *              position- and length-variables
+ * TimePos.h - declaration of class TimePos which provides data type for
+ *             position- and length-variables
  *
  * Copyright (c) 2004-2014 Tobias Doerffel <tobydox/at/users.sourceforge.net
  *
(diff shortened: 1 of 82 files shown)


Spekular:
This looks like a good candidate to be part of #5592 which is targeted around December.

include/TimePos.h
```diff
@@ -57,17 +57,17 @@ class LMMS_EXPORT TimeSig
 };
 
 
-class LMMS_EXPORT MidiTime
```
JohannesLorenz, line 60: Only in case you learned what this class does: Can you please add a doxygen comment above it? (Same for `TimeSig`, also voluntarily).

JohannesLorenz (changes requested):
Only 3 minor comments.

If you make changes, please add new commits (instead of squashing), so I don't need to review this large diff again ;) I'll squash it together when merging then.

JohannesLorenz (approved):
All good, thanks for the PR!

I'll merge this right now.

PR #5685, opened 2020-09-25
Remove unused stuff

M374LX:
This PR removes unused stuff: ```DummyTrackContainer``` and one of the overloaded methods in ```Mixer``` named ```setAudioDevice()```.

### DIFF
--- a/include/Engine.h
+++ b/include/Engine.h
@@ -35,7 +35,6 @@
 #include "lmms_basics.h"
 
 class BBTrackContainer;
-class DummyTrackContainer;
 class FxMixer;
 class ProjectJournal;
 class Mixer;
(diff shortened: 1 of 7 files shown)


russiankumar:
Please correct me if I'm wrong, but I believe `DummyTrackContainer` is kept around as a template to get a headstart if a new track type were to be added? If that's the case I think it would be better to keep it around than remove it and depend on other track types.

DomClark:
I think it's more likely that it got left behind accidentally; it looks like it was intended solely for holding a dummy instrument track, that could be used as a placeholder model for an `InstrumentTrackWindow` without any corresponding track. It doesn't look particularly useful to me as a template for new track containers. I think it's fine to remove it. (Unless @PhysSong disagrees, as someone who's working on `InstrumentTrackWindow`s.)

DomClark (approved):

PhysSong:
No objections from me, too.

DomClark:
Merge?

russiankumar (approved):
No objections from me.

PR #5117, opened 2019-08-10
Improved relative paths

Spekular:
Moves the toRelative/toAbsolute functions out of SampleBuffer and allow paths to be relative to locations other than the sample directory. This is a WIP, partially opened so I can get Windows binaries ;)

Essentially just @PhysSong 's solution [here](https://github.com/LMMS/lmms/issues/4180#issuecomment-407307675).

PR #5689, opened 2020-09-27
Fix bug with baseQDir in PathUtils.cpp

Spekular:
_Originally posted by @PhysSong in https://github.com/LMMS/lmms/pull/5117#discussion_r495384656:_

If `base` is `Absolute`, `baseQDir(base)` will point to the working directory. It will result in undefined behaviors.

---

Opening for easier discussion, even this solution is flawed:

> Better than before, but that may result in getting foo/bar for /foo/bar which will be interpreted as old-style relative path.

### DIFF
--- a/src/core/PathUtil.cpp
+++ b/src/core/PathUtil.cpp
@@ -36,7 +36,11 @@ namespace PathUtil
 		return QDir::cleanPath(loc) + "/";
 	}
 
-	QDir baseQDir (const Base base) { return QDir(baseLocation(base)); }
+	QDir baseQDir (const Base base)
+	{
+		if (base == Base::Absolute){ return QDir::root(); }
+		return QDir(baseLocation(base));
+	}
 
 	QString basePrefix(const Base base)
 	{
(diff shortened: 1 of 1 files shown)


PhysSong:
I think `relativeOrAbsolute()` shouldn't use `relativePath` if `base == Absolute` in order to fix the issue I mentioned.

src/core/PathUtil.cpp
```diff
@@ -36,7 +36,11 @@ namespace PathUtil
 		return QDir::cleanPath(loc) + "/";
 	}
 
-	QDir baseQDir (const Base base) { return QDir(baseLocation(base)); }
+	QDir baseQDir (const Base base)
+	{
+		if (base == Base::Absolute){ return QDir::root(); }
```
PhysSong, line 41:
```suggestion
		if (base == Base::Absolute) { return QDir::root(); }
```

PhysSong:
LGTM

Spekular:
Just realized the previous fix didn't do what I intended it to, whoops.

PhysSong (approved):

PR #5691, opened 2020-09-29
Enable Lv2 Atom ports

JohannesLorenz:
This commit enables atom ports and feeds them with MIDI events, if the atom ports accepts them. This enables Lv2 instruments with piano roll, e.g. the MDA Piano.

Reviewer hints:
* ~~Please review~~

Tester hints:
* ~~Wait with final testing until review is done~~
* Set up a wide range of packages (my current list of packages at Arch Linux is: "calf lsp-plugins mda.lv2 noise-repellent surge x42-plugins", and from the AUR: "lv2-plugins-aur-meta")
* Any Lv2 instrument is worth testing
* New effects are also available: the x42-eq and x42-dpl
* This PR is actually just about passing piano roll key presses to Lv2 instruments (which it does via MIDI internally: when you press a key in piano roll or play it in the editor, LMMS passes MIDI to the plugin, and Lv2 now recognizes these MIDI events).
* Things to check can include
  - play notes using
    * piano roll
    * keyboard keys
    * the Instrument view's keyboard
  - playing tons of notes at the same or nearly the same time

### DIFF
--- a/src/core/lv2/Lv2Ports.cpp
+++ b/src/core/lv2/Lv2Ports.cpp
@@ -27,9 +27,12 @@
 
 #ifdef LMMS_HAVE_LV2
 
+#include <lv2/lv2plug.in/ns/ext/atom/atom.h>
+
 #include "Engine.h"
 #include "Lv2Basics.h"
 #include "Lv2Manager.h"
+#include "Lv2Evbuf.h"
 
 namespace Lv2Ports {
 
@@ -57,7 +60,7 @@ const char *toStr(Type pt)
 		case Type::Unknown: return "unknown";
 		case Type::Control: return "control";
 		case Type::Audio: return "audio";
-		case Type::Event: return "event";
+		case Type::AtomSeq: return "atom-sequence";
 		case Type::Cv: return "cv";
 	}
 	return "";
@@ -168,6 +171,23 @@ std::vector<PluginIssue> Meta::get(const LilvPlugin *plugin,
 		issue(badPortType, "cvPort");
 		m_type = Type::Cv;
 	}
+	else if (isA(LV2_ATOM__AtomPort))
+	{
+		AutoLilvNode uriAtomSequence(Engine::getLv2Manager()->uri(LV2_ATOM__Sequence));
+		AutoLilvNode uriAtomBufferType(Engine::getLv2Manager()->uri(LV2_ATOM__bufferType));
+		AutoLilvNodes bufferTypes(lilv_port_get_value(plugin, lilvPort, uriAtomBufferType.get()));
+
+		if (lilv_nodes_contains(bufferTypes.get(), uriAtomSequence.get()))
+		{
+			// we accept all kinds of atom sequence ports, even if they take or
+			// offer atom types that we do not support:
+			// * atom input ports only say what *can* be input, but not what is
+			//   required as input
+			// * atom output ports only say what *can* be output, but not what must
+			//   be evaluated
+			m_type = Type::AtomSeq;
+		}
+	}
 
 	if(m_type == Type::Unknown)
 	{
@@ -241,6 +261,11 @@ void Audio::copyBuffersToCore(sampleFrame *lmmsBuf,
 		lmmsBuf[f][channel] = m_buffer[f];
 	}
 }
+
+
+
+
+void AtomSeq::Lv2EvbufDeleter::operator()(LV2_Evbuf *n) { lv2_evbuf_free(n); }
 
 
 
--- a/src/core/lv2/Lv2Proc.cpp
+++ b/src/core/lv2/Lv2Proc.cpp
@@ -27,6 +27,11 @@
 #ifdef LMMS_HAVE_LV2
 
 #include <cmath>
+#include <lv2/lv2plug.in/ns/ext/midi/midi.h>
+#include <lv2/lv2plug.in/ns/ext/atom/atom.h>
+#include <lv2/lv2plug.in/ns/ext/resize-port/resize-port.h>
+#include <QDebug>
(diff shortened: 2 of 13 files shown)


JohannesLorenz:
CI finally passed :smiley:  Ready for review.

src/core/lv2/Lv2Proc.cpp
```diff
@@ -447,6 +689,27 @@ void Lv2Proc::createPorts()
...
 		}
+
+		void visit(Lv2Ports::AtomSeq& atomPort) override
+		{
+			if(atomPort.m_flow == Lv2Ports::Flow::Input)
+			{
+				if(atomPort.flags & Lv2Ports::AtomSeq::FlagType::Midi)
+				{
+					Q_ASSERT(!m_proc->m_midiIn);
+					m_proc->m_midiIn = &atomPort;
```
DomClark, line 700: Is it worth checking in `Lv2Proc::check` that at most one MIDI in (and out) port is present? Otherwise plugins with multiple MIDI in/out ports can appear in the instrument list, and LMMS will exit with an assertion failure if the user tries to add one to a project.

src/core/lv2/Lv2Ports.cpp
```diff
@@ -168,6 +171,7 @@ std::vector<PluginIssue> Meta::get(const LilvPlugin *plugin,
 		issue(badPortType, "cvPort");
 		m_type = Type::Cv;
 	}
+	else if (isA(LV2_ATOM__AtomPort)) { m_type = Type::AtomSeq; }
```
DomClark, line 174: Should this first check that the atom port supports MIDI, and add an "unknown port type" issue if not?

DomClark (commented):
Generally looks good to me.

src/core/lv2/Lv2Proc.cpp
JohannesLorenz: Good point. I fixed that now.

src/core/lv2/Lv2Ports.cpp
JohannesLorenz: I talked with the lv2-devs in the lv2 IRC channel, and they said it's OK to have atom sequence ports, even if you don't support any of its types. A malfunction is not expected in this case. I added a justification comment into the code though.

One other issue I found here is that the line you marked assumes that every atom port is an atom sequence port. I fixed this in the same commit (7e059b3).

JohannesLorenz: I'll mark this as resolved, feel free to re-open.

JohannesLorenz:
All comments re-worked. Let's see what CI says...

JohannesLorenz:
CI passed :tada: @DomClark can you please check my rework?

DomClark (commented):
Although `m_midiIn` is set to the required MIDI in port if one is present, this doesn't seem to be taken into account when actually sending events to the plugin - they all go to the first visited MIDI in port instead.

JohannesLorenz:
@DomClark I fixed it, to use `m_midiIn` now. Can you please re-check? The commit has a large diff, but I only refactored, nothing else.

I wonder if the MIDI to buffer code (`midiInputEventToData`) should be moved into a common function, so that  `plugins/carlabase/carla.cpp` (`CarlaInstrument::handleMidiEvent`) can re-use the code. What do you think?

JohannesLorenz:
@DomClark I only added 2 more commits:
1. I fixed your comment about MIDI input events that might have not been sent to `m_midiIn` (7ada8c0).
2. Refactoring: Carla and Lv2 share the code now which writes a MIDI event into a raw buffer (3a79181).

Can you please check those two, so we can finish the review?

src/core/lv2/Lv2Proc.cpp
```diff
@@ -147,31 +182,89 @@ void Lv2Proc::copyModelsFromCore()
 
 	struct Copy : public Lv2Ports::Visitor
 	{
+		ringbuffer_reader_t<MidiInputEvent>* m_midiInputReader;
```
DomClark, line 185: `m_midiInputReader` isn't used any more now MIDI input is handled separately.

src/core/lv2/Lv2Proc.cpp
```diff
@@ -147,31 +182,89 @@ void Lv2Proc::copyModelsFromCore()
...
+		{
+			port->accept(copy);
+		}
+	}
+
+	// send pending MIDI events to atom port
+	if(m_midiIn)
+	{
+		LV2_Evbuf_Iterator iter = lv2_evbuf_begin(m_midiIn->m_buf.get());
+		if(m_midiIn->flags & Lv2Ports::AtomSeq::FlagType::Midi)
```
DomClark, line 224: This condition is checked when initialising `m_midiIn`, so doesn't need repeating here.

DomClark (approved):
Everything looks correct now, as far as I can tell. I've suggested a couple of small improvements.

IanCaio (approved):
Not a code review (not very familiar with the Lv2 core right now), but a tester review.

Everything seems to be working fine. I tried testing other plugins besides Surge (`drumkv1`, `ZynAddSubFX`, `samplv1`, `Black Pearl Drumkit` and `Red Zeppelin Drumkit`) but those couldn't be loaded due to missing features. If requested I could test with another plugin known to not require those.

Surge has worked perfectly though, the Instrument Widget's, Piano Roll's and QWERTY MIDI keyboards played the notes smoothly. Tried adding lots of notes simultaneously (above 20 I guess) and it worked fine. Also tested using a hardware MIDI keyboard and it behaved properly.

No misbehavior or instabilities noticed. Very excited to see how Lv2 is growing!

Issue #4752, opened 2019-01-03
Update rpmalloc


shlyakpavel:
Hello. I have found some problems related to rpmalloc_thread_collect working improperly. However, it is fixed on both rpmalloc develop and master branches
It would be nice if you merged the up to date version of the code :)
Have a nice day!

shlyakpavel:
Calf seems to be outdated as well...

tresf:
> Calf seems to be outdated as well...

Calf submodule is `master` branch only but it's a special LADSPA fork and won't merge directly.  The fork is located here: https://github.com/LMMS/veal.  Feel free to submit a PR.  @JohannesLorenz is the maintainer of that fork.

> It would be nice if you merged the up to date version of the code ... [to fix] ... some problems related to rpmalloc_thread_collect working improperly

👍 We'll take a look, thank you!

zonkmachine:
OK. I assigned you to the issue and since no one has said anything I just assume you can go ahead and work on it right away.

JohannesLorenz:
Just to explain why we re-opened that. I made an update, but the CI for 14.04 failed. rpmalloc 1.3.1 uses features which gcc 4.8 (the 14.04 compiler) does not support due to a compiler bug (4.7 and 4.9 are not affected).

I see multiple options:
* :rocket: Discard 14.04 for 1.2 completely (maybe use 16.04 or 18.04), as any support for 14.04 runs out on April 2019.
* :eyes: Somehow patch 14.04 to user gcc 4.9 or higher
* :tada: Refuse to update rpmalloc for 1.2

Voting time!

tresf:
This is a tough topic to vote on because each decision has its own ramifications.   Sorry, I have more questions than answers... 

* What are the major ramifications of leaving `stable-1.2` on 14.04 and keeping our current version of `rpmalloc`?  We need less changes to the `stable-1.2` branch, not more and @lukas-w has already [started refactoring the memory management in `master`](https://github.com/LMMS/lmms/pull/4443). 
* I've used gcc 4.9 on 14.04 without any major issues, are there any anticipated problems (e.g. such as with the AppImage?)  If not, this seems like a completely reasonable approach, assuming there **are** major ramifications (depends on bullet 1)
* Again, not updating depends on the severity of not updating `rpmalloc`.

tresf:
[...]

PR #5694, opened 2020-10-02
Haiku build fix.

devnexen:
related to ringbuffer, matching cmake settings to disable mlock for this platform.
Haiku does not support tls model as well.


devnexen:
Little question. If there any plan to update rpmalloc (ie its current git hash is old and sets wrongly tls model, thus the runtime_loader fails to resolve when it starts but it had been solved since) ?

JohannesLorenz:
About rpmalloc, we still have #4752 open. We should get that done, but this is blocked as long as we still support Ubuntu 14.04.

@devnexen Do you require an rpmalloc update to get LMMS run on Haiku?

devnexen:
Right but isn't Ubuntu 14.04 EOL ? Anyway yes in that cas the compiler toolchain might be old for newer rpmalloc. Yes indeed I changed locally to make it work :-)

JohannesLorenz (approved):
Looks all good. I'll merge this now. The rpmalloc fix will remain delayed until #5696 is done.

JohannesLorenz:
@devnexen for your info, rpmalloc is updated on master now.

PR #5697, opened 2020-10-03
Compile LMMS with C++14

JohannesLorenz:
Compiles the LMMS core (and most plugins) with C++14.

Should work now, since we use Xenial (Ubuntu 16).

~~More commits will be added here if it works...~~

~~**No reviews yet, this is still WIP (CI failures)**~~


plugins/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ SET(CMAKE_LIBRARY_OUTPUT_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}")
 SET(CMAKE_DEBUG_POSTFIX "")
 
 # Enable C++11
```
Spekular, line 5: Update this comment as part of whatever you commit next?

src/CMakeLists.txt
```diff
@@ -10,7 +10,7 @@ SET(CMAKE_AUTOMOC ON)
 SET(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR})
 
 # Enable C++11
```
Spekular, line 12: Same here I guess

Spekular:
Ah whoops, you specifically asked for no reviews... Well, since I checked the diff anyway, LGTM so far 😅

JohannesLorenz:
@Spekular No problem, thanks for checking anyways.

JohannesLorenz:
@Spekular Can you please re-check? I only changed `plugins/vst_base/RemoteVstPlugin/CMakeLists.txt` , everything else is like from the last review.

Spekular:
That looks good too as far as I can tell. Seems like `CMAKE_REQUIRED_FLAGS` is undocumented and it's preferable to use something else, but it was there before this PR and CI seems to pass so /shrug.

Also, I'm assuming `plugins/LadspaEffect/calf/CMakeLists.txt` will inherit the `CMAKE_CXX_STANDARD` flag from `plugins/CMakeLists.txt`, or something like that? Otherwise it seems weird to remove the `SET` instead of updating it.

JohannesLorenz:
> Also, I'm assuming `plugins/LadspaEffect/calf/CMakeLists.txt` will inherit the `CMAKE_CXX_STANDARD` flag from `plugins/CMakeLists.txt`, or something like that?

Yes, that's what I understood (which was also why I removed the `SET(CMAKE_CXX_STANDARD 11)` from all the `plugins/.../CMakeLists.txt` instead of updating them.

DomClark (commented):
CI has passed, so hopefully reviews are alright now.
* `#include <memory>` can now be removed from `stdshims.h`
* I don't think including `algorithm` is necessary in the files that used to use shims - `std::make_unique` is in the `memory` header, which is probably already included since the code is using `std::unique_ptr`s in the first place. (Sorry if the comment on the LV2 atom PR was confusing - I was referring to `std::any_of` there.)

JohannesLorenz:
@DomClark Thanks for double checking. I somehow confused `<algorithm>` and `<memory>`. Both fixed.

JohannesLorenz:
@devnexen Can you please check if this PR compiles on Haiku before I merge it to master?

devnexen:
I can confirm it passes.

DomClark (approved):

JohannesLorenz:
@devnexen Great to hear. Thanks!

Issue #5673, opened 2020-09-08
Beat and bassline bug


AbaddonBent:
There's a bug where a instrument in beat and bassline has automation and the tracks above that instrument become hidden and un-editable.
![image](https://user-images.githubusercontent.com/69567825/92424486-5e8dc280-f139-11ea-82a2-52292339ffe0.png)

I was making pitch adjustments when I noticed this

thmueller64:
Looking at your screenshot I assume you are using the 1.2.2 Appimage.
I unsuccessfully tried to reproduce the bug using the information in your screenshot with the following steps:
  1. Create a beat/bassline instance
  2. Create patterns inside the track with audio samples only (I created 16 as above)
  3. Automate the pitch of one of the last patterns

Can you reliably reproduce the bug? If so, could you please detail the steps to reproduce it? You could also share a project file.

zonkmachine:
I think this is mentioned somewhere else. A method to make empty Beat/Bassline tracks is to delete all beat bassline instances in the Song Editor. Existing tracks in the Beat/Bassline will then go blank. Sounds like you managed to do it in some other way.

AbaddonBent:
I've been trying to duplicate it but nothing has been showing....
The file I had has changed and when I opened it, it has reset the beat and bassline, the screenshot is the only proof I've had, but I'll keep trying to see if it happens.
My theory is that it was a one time thing

AbaddonBent:
What I remembered what I did when this happened, I accidentally pressed the add sample track button. Then removed it and pressed the automation track and figured I'd mess with the pitch

IanCaio:
I have experienced something similar by doing the following steps (I described it on my PR because at first I thought it could be related to it but then realized it was not):

1. Open the default project (TripleOsc, SampleTrack, BBTrack and AutomationTrack).
2. Create 2 TCOs on the TripleOsc track on Bars 1 and 2
3. Ctrl+Drag the TripleOsc track to the BBEditor
4. Add some notes to the Kicker track (not necessary for the bug to happen)
5. Clone the BBTrack

Could it be you did something similar, like dragging an instrument that already had TCOs to the BB Editor before cloning? I'm still not sure what causes this bug, but could anyone check if they can reproduce it by following those steps?

IanCaio:
Afaik, all BB Tracks share the same tracks, and what differs a BB pattern from another is the TCO being played/visualized. Something like this:

```
          BBTrack1        BBTrack2       BBTrack3
[...]

Issue #5683, opened 2020-09-25
MSVC RelWithDebInfo configuration uses mixture of debug and release libraries


DomClark:
### Bug Summary

When building with MSVC using the `RelWithDebInfo` configuration, a mixture of debug and release libraries is used.

#### Steps to reproduce

Build the `master` branch with MSVC using `RelWithDebInfo` configuration, or download an MSVC build from CI, since that's the configuration we use there.

#### Expected behavior

All libraries used should belong to the same configuration (preferably `Debug`).

#### Actual behavior

LMMS uses the debug CRT, and debug Qt libraries, but release versions of the libraries from vcpkg.

#### Affected LMMS versions

MSVC `RelWithDebInfo` builds of `master` branch.

#### Further information

This leads to crashes when, e.g., trying to load samples. LMMS and Qt use the debug CRT to open files, but pass these file handles to libsndfile, which is using the release CRT. This causes a crash when the release CRT is asked to read a file using a handle it doesn't recognise.

This could be addressed by setting the [`CMAKE_MAP_IMPORTED_CONFIG_RELWITHDEBINFO`](https://cmake.org/cmake/help/latest/variable/CMAKE_MAP_IMPORTED_CONFIG_CONFIG.html) variable to `RelWithDebInfo Debug`, to instruct CMake to use `Debug` libraries for the `RelWithDebInfo` configuration if `RelWithDebInfo` versions don't exist, instead of `Release` libraries. This won't work for SDL2 however - we have a custom [`FindSDL2.cmake`](https://github.com/LMMS/lmms/blob/0196f31509f52b1ccbf2eb8edd6f0a117f752bf6/cmake/modules/FindSDL2.cmake) module, which takes precedence over SDL2's own `SDL2Config.cmake` script, and doesn't know how to find vcpkg's SDL2 debug build.

DomClark:
Actually tried building `RelWithDebInfo` locally, and it's backwards on my machine. I get the release version of the CRT and Qt, but debug libraries from vcpkg. (SDL2 is still release for the aforementioned reason.) I'm not sure why - perhaps newer versions of Qt and vcpkg? Either way, the inconsistency is a problem.

tresf:
> I get the release version of the CRT and Qt, but debug libraries from vcpkg.

CRT seems to be an odd duck...  I searched and found solutions involving forced flags and generator expressions.  Almost smells like a CMake bug.
[...]

PR #5699, opened 2020-10-04
Fixes createTCO method on some classes

IanCaio:
Classes that inherit from `Track`, also inherit the `createTCO` method. That method takes a MidiTime position as an argument, but except on the class `SampleTrack` that argument was ignored. That lead to unnecessary calls to `TCO->movePosition()` after creating a TCO in many parts of the code base (making the argument completely redundant) and even to a bug on the BBEditor, caused by a call to `createTCO` that expected the position to be set on the constructor (see issue #5673 ).

That PR adds code to move the TCO to the appropriate position inside the constructor of the classes that didn't have it, fixes the code style on the SampleTrack `createTCO` method and removes the now unneeded calls to `movePosition` from source files on _src/_ and _plugins/_. On `Track::loadSettings` there was a call to `saveJournallingState(false)` followed immediately by `restoreJournallingState()` which was deleted because it's redundant (probably a left over from copying the code from the `pasteSelection` method?).


src/tracks/SampleTrack.cpp
```diff
@@ -746,10 +746,10 @@ TrackView * SampleTrack::createView( TrackContainerView* tcv )
 
 
 
-TrackContentObject * SampleTrack::createTCO(const MidiTime & pos)
+TrackContentObject * SampleTrack::createTCO(const MidiTime & _pos)
 {
-	SampleTCO * sTco = new SampleTCO(this);
-	sTco->movePosition(pos);
+	SampleTCO * sTco = new SampleTCO( this );
+	sTco->movePosition( _pos );
```
Spekular, line 752: I'd prefer not to reintroduce outdated formatting
```suggestion
	SampleTCO * sTco = new SampleTCO(this);
	sTco->movePosition(_pos);
```

I know the rest of the createTCO calls have this spacing, but if you want consistency I'd change those instead, since you're replacing most of the code in them anyway. Otherwise I'd leave them as is.

Spekular:
Code LGTM, one minor formatting concern. Do you have any test cases in mind for testers?

src/tracks/SampleTrack.cpp
IanCaio: Cool, I'll revert that change and fix the style on the new code. I actually made a mistake about which was the right formatting, since I was so used to adding space on the arguments from other PRs

IanCaio:
As for testing cases, I'd say all operations that involves creating TCOs, but more specifically the ones which causes calls to `createTCO`. From grepping the source some I could find:

- Drag & Dropping Sample Tracks
- Adding a sample using Shift+Enter on the File Browser
- Copy and Paste (and Drag & Drop) track selections
- Adding TCOs on the song editor
- Drag & Dropping tracks with TCOs to the BB Editor
- And particularly the bug described on my comment in #5673 , to see if it is fixed.

src/tracks/InstrumentTrack.cpp
```diff
@@ -713,9 +713,11 @@ bool InstrumentTrack::play( const MidiTime & _start, const fpp_t _frames,
 
 
 
-TrackContentObject * InstrumentTrack::createTCO( const MidiTime & )
+TrackContentObject * InstrumentTrack::createTCO( const MidiTime & _pos )
```
Spekular, line 716: Sorry that I didn't highlight it earlier, but the same (lack of) spacing should apply to the parameter list. Feel free to disregard, I don't think something this minor should hold back the PR.

IanCaio: I fixed the spacing on the parameters list and on `BBTrack::createTCO` even on the statements that weren't added by this PR

src/tracks/AutomationTrack.cpp
```diff
@@ -57,9 +57,11 @@ TrackView * AutomationTrack::createView( TrackContainerView* tcv )
 
 
 
-TrackContentObject * AutomationTrack::createTCO( const MidiTime & )
+TrackContentObject * AutomationTrack::createTCO(const MidiTime & _pos)
```
PhysSong, line 60: FYI, `_pos` is old style. Please use `pos` in new code, or at least don't revert `pos` to `_pos`.

IanCaio: Thanks, I'll fix it. I had seem some existing files with `_pos` and ended up thinking I was fixing the others, when it was the other way around :laughing: 

What was the underscore for in the old style?

IanCaio:
Fixed the merge conflict and the code style on the changed lines of the code.

russiankumar:
I'll run some tests later in the day. I'll edit this comment with the results.

IanCaio:
Thanks for the detailed testing man!

I'm not sure about that Shift+Enter crash you experienced, couldn't replicate it here. But I'd say it's very likely unrelated. This code only changes the `createTCO` method and updates places that call it.

As for the dragging selection behavior, this was present before this bugfix. I looked briefly at the code and I think it could be fixed by adding `tco->positionChanged()` in the end of the `TrackContentWidget::pasteSelection()` method so it can update the TCOView. I'd have to double check and test it though.

The `called Track::getTCO( 1 ), but TCO 1 doesn't exist` is a warning triggered by `Track::getTCO` when it's given a TCO number where there aren't any. In that case it creates one anyways, I'm not sure this warning was previously used for debugging but it has been there for a while. I think if the first TCO has notes they should be kept on the first BBTrack though, even if there are other TCOs after it. This wasn't the behavior there?

superpaik:
Some issues:
- when "Ctrl+LMB" over a TCO one expects to drag the selected TCO to where the mouse pointer is, but sometimes it goes to another point. It happens with all type of TCO but with samples track TCO happens always, with automation and instruments seems more random
test
- I don't know if it is related to this PR. You LMB + crtl a TCO ("Edit mode" becomes active), if you let go "Ctrl" before LMB "Draw mode" returns to active. But If you let go LMB before "Crtl", "Edit mode" keeps active.
- When you drag&drop a track from the BB Editor to Song Editor, the created TCO is gray (like it's empty). If you add a note (from the piano roll) it becomes green and all notes are shown. I think this behaviour is already present in 1.2.2, but maybe it's realated to the modified code.
- I had to tried the mingw version, because MSVC version crashes a lot ¿?

Spekular:
The MSVC issue is unrelated to this PR, see #5683.

IanCaio:
@superpaik Thanks for testing!

First issue is actually an intended behavior of the drag&drop operation, you can't drag&drop a TCO on itself, so it snaps it to the closest neighbor position. Dragging to any other position should work as intended though. Second issue (with LMB+Ctrl) is probably related to the SongEditor handling of mouse and keyboard events, would have to check the code to be sure. But it's unrelated to this PR.

The third one, is a bit trickier to explain. Notes on the BBEditor are different from the notes we add at the piano roll. They have a "negative" length if I'm not mistaken. It's a way of LMMS knowing that it should play the sample until it ends (instead of cutting it when the note ends). So when you drag a track from the BBEditor back to the SongEditor, a grey clip shows up because the paint event of the clip probably doesn't handle those special notes. They are there, but you can't see them on the song editor (if you open the piano roll they will show as very tiny notes). I don't know if there are plans to moving towards another way of representing those special notes, but for now it can get a bit confusing to the user at first when we start mixing BBEditor notes with SongEditor ones.

IanCaio:
@russiankumar I read the code a little further, and it seems to be purposeful behavior as seen on `src/gui/editor/BBEditor.cpp` line 256:

```
		// Ensure BB TCOs exist
		bool hasValidBBTCOs = false;
		if (t->getTCOs().size() == m_bbtc->numOfBBs())
		{
			hasValidBBTCOs = true;
			for (int i = 0; i < t->getTCOs().size(); ++i)
			{
				if (t->getTCOs()[i]->startPosition() != MidiTime(i, 0))
				{
					hasValidBBTCOs = false;
					break;
				}
			}
		}
		if (!hasValidBBTCOs)
		{
			t->deleteTCOs();
			t->createTCOsForBB(m_bbtc->numOfBBs() - 1);
		}
```

The TCOs from a track are only kept if they all fit perfectly within the BBTrack layout of TCOs. That means that the number of TCOs has to match the number of BBTracks and they must be perfectly lined. When you have a second TCO, because you only have one BBTrack it considers the last TCO to be a deal breaker and just removes both. If you have two BBTracks on the other hand (and make sure you have both TCOs aligned in the position necessary for them to fit the BB layout) this drag and drop operation should keep them.

russiankumar:
Alright, since it is intended, I think we should leave it as is. Didn't even know this was a thing before xD.

src/core/Track.cpp
```diff
@@ -1913,7 +1913,7 @@ bool TrackContentWidget::pasteSelection( MidiTime tcoPos, const QMimeData * md,
 
 		TrackContentObject * tco = t->createTCO( pos );
 		tco->restoreState( tcoElement );
-		tco->movePosition( pos );
+		tco->movePosition(pos); // Because we restored the state, we need to move the TCO again.
```
russiankumar, line 1916: Just being a nitpick, but shouldn't it be `( pos )` instead of `(pos)`? (Based on the way it is written two lines above)

src/core/Track.cpp
```diff
@@ -1967,11 +1967,7 @@ void TrackContentWidget::mousePressEvent( QMouseEvent * me )
 		getTrack()->addJournalCheckPoint();
 		const MidiTime pos = getPosition( me->x() ).getBar() *
 						MidiTime::ticksPerBar();
-		TrackContentObject * tco = getTrack()->createTCO( pos );
-
-		tco->saveJournallingState( false );
-		tco->movePosition( pos );
-		tco->restoreJournallingState();
+		getTrack()->createTCO(pos);
```
russiankumar, line 1970: Same here.

russiankumar: There's also a few more places where it's written as `(pos)`.

src/core/Track.cpp
IanCaio: From what I understood, `(pos)` is actually the way endorsed by our code style guideline (I'm guilt of using spaces inside parenthesis before though, personally I thought they were more readable). On my PRs I've been trying to fix the code style on the lines that are affected by the PR only (that's why the lines above still have `( pos )`), that's so reviewing doesn't get cluttered by all the code style fixes

superpaik:
> 
> 
> The MSVC issue is unrelated to this PR, see #5683.

Ok. I tend to use the MSVC version because of the size of the .exe (mingw exe is three times bigger than MSVC). I will take it into account for future tests. Thanks.

src/core/Track.cpp
russiankumar: Oh, if that is the case, then no issue. I thought it would be better to keep it consistent, but you've already been doing this, so no objections from me.

IanCaio: Soon we should have the clang-format to fix the remaining code style issues, at least those parenthesis spaces should be easily fixable with it :smile:

russiankumar (approved):

src/core/Track.cpp
```diff
@@ -1896,6 +1895,22 @@ bool TrackContentWidget::pasteSelection( MidiTime tcoPos, const QMimeData * md,
...
+
+		MidiTime pos = tcoElement.attributeNode("pos").value().toInt();
+
+		if(pos < leftmostPos) { leftmostPos = pos; }
+	}
+	if((leftmostPos + offset) < 0)
+	{
+		offset = -leftmostPos;
+	}
+
```
Spekular, line 1913:
```suggestion
	offset = max(offset, -leftmostPos)
```

src/tracks/BBTrack.cpp
```diff
@@ -382,17 +382,10 @@ TrackView * BBTrack::createView( TrackContainerView* tcv )
...
-	BBTCO * bbtco = new BBTCO( this );
-	if( s_lastTCOColor )
-	{
-		bbtco->setColor( *s_lastTCOColor );
-		bbtco->setUseStyleColor( false );
-	}
-
-	bbtco->movePosition( _pos );
-
+	BBTCO* bbtco = new BBTCO(this);
```
Spekular, line 387: What's with this removal?

Spekular (commented):
Recent changes look good, except for one that I don't really understand what it's for. One if statement that can be a max() call instead.

Spekular:
Tested this because I was a bit concerned that it might block individual clips rather than the group (i.e. lose offsets if you move a selection left after it hits the start), but both pasting and dragging still seem to respect offsets. Left a new code review as well.

src/tracks/BBTrack.cpp
Spekular: Clarification since I marked the wrong line (oops): this is in reference to the color stuff that was removed.

src/core/Track.cpp
IanCaio: True, that works! It makes the logic a tiny bit less explicit, but works just the same. Should I use `max` or `qMax`?

src/tracks/BBTrack.cpp
IanCaio: Not sure what removal do you mean, some lines from this method were removed on master due to the Track color PR. I think in the current state of this one there's only the `new BBTCO`, `movePosition` and `return` statements on `BBTrack::createTCO`.

src/core/Track.cpp
Spekular: No preference!

src/tracks/BBTrack.cpp
Spekular: Looks like the github diff was acting up, so disregard this 😅

src/core/Track.cpp
IanCaio: Used `std::max`. `qMax` is more often used, but from now on I think I'll give preference to `std` unless it's a requirement that we have some Qt feature. I didn't use max on the MoveSelection action because IMHO it needs a refactor for better readability, so this could be changed when that happen.

I just wanted to explain why I explicitly called `getTicks` on both terms of `std::max`:
If I used `std::max(offset, -leftmostPos);` I'd be comparing a `MidiTime` with an `int` type (compiler error), because since there's an arithmetic operation on `leftmostPos` it calls the `operator int()` method returning `m_ticks` which happens to be an `int` type. If I have `std::max(offset.getTicks(), -leftmostPos);` it would work fine, but I feel like it's bad practice to compare `ticks_t` with `int`, even though they are matching types, without explicitly showing what types are being compared. If it sounds like an overkill I can replace `-leftmostPos.getTicks()` with just `-leftmostPos`.

Veratil: I think the only issue with `std::max` is Windows will throw errors sometimes, at least that's what always seems to happen to me. lol

russiankumar:
I realise that I'm already late 😅 but I've done a test, the functional part of it works.

capture4
*Here, a very useful GIF*

I've already approved the code. LGTM to me after you settle the review Spekular left.

You can close #5768 when this is merged, or I can do it myself.

src/core/Track.cpp
Spekular: Since CI is still passing, I guess it's fine in this case.

Spekular (approved):

PR #5700, opened 2020-10-05
Adds feature to merge Instrument Track patterns

IanCaio:
This PR implements a new action on the context menu of TrackContentObjectViews for merging note patterns. It's only applicable for InstrumentTrack's TCOs. In terms of changes on the code those were the main ones:

- Create a `getPattern` method on the `PatternView` so we can add the notes to a pattern obtained from a pattern view.
- Adds a context menu action called Merge and the appropriate context menu entry if the selection can be merged.
- Create two methods on TrackContentObjectView: `canMergeSelection` and `mergeTCOs`. The first check if a vector of TCOVs can be merged (the conditions: more than one TCO being selected, they all belonging to an Instrument Track and they all belonging to the same track). The second does the actual merging, by creating a new TCO and adding all the notes to it while removing the merged TCOs.
- Small note: There's a `TODO:` comment regarding a statement that should be removed once #5699 is merged.

There are a few warnings that might be unnecessary (they cover some situations that should not normally happen). A placeholder icon was created for the action, which could be later substituted by a better one.

Screenshot from 2020-10-04 20-59-16
Screenshot from 2020-10-04 20-58-09


superpaik:
[Note: this is my first review, so let me know if you need more detail]

The averall functionality works fine with me. Some issues, though:
- when you merge two patterns and one is empty, if it's the first the empty one, it merge the two of them, but if the empty is the second, it doesn't merge. I don't know if this behaviour is wanted or not, but seems odd to me.
- the functionality of undo (ctrl-z) is weird. If you merge two patterns, you need two undos to undo the merge. If you merge the patterns again (after the undo), then you need three undos. If you do it again, then four, and so on. If you merge more than two, say three, and the pattern to be merge are created in order (e.g. first on beat 5, then on beat 9, another on beat 13) and merge them, it works well (aside from the double undo), but if you create them and then move the pattern from 13 to 1 and merge and undo, the behaviour is weird (first the notes on the pattern on beat1 desappear, and then the pattern) ending with only patterns on beats 5 and 9.

IanCaio:
Thanks for testing and giving feedback!

> * when you merge two patterns and one is empty, if it's the first the empty one, it merge the two of them, but if the empty is the second, it doesn't merge. I don't know if this behaviour is wanted or not, but seems odd to me.

That is expected because the clip's size is recalculated so it is the shorted possible on the song editor (that also happens on the piano roll, if you delete notes from the last bar the clip will be shortened to have the fewest bars it can). So when you merge an empty clip with a non-empty one on that order it doesn't get shortened, because there are notes at the end of the clip. But if you merge a non-empty clip with an empty one the empty bars at the end are ignored and the clip is shortened to be the smallest size possible. They are still being merged though.

> * the functionality of undo (ctrl-z) is weird. If you merge two patterns, you need two undos to undo the merge. If you merge the patterns again (after the undo), then you need three undos. If you do it again, then four, and so on. If you merge more than two, say three, and the pattern to be merge are created in order (e.g. first on beat 5, then on beat 9, another on beat 13) and merge them, it works well (aside from the double undo), but if you create them and then move the pattern from 13 to 1 and merge and undo, the behaviour is weird (first the notes on the pattern on beat1 desappear, and then the pattern) ending with only patterns on beats 5 and 9.

Oh yeah, I actually didn't write a history system for merging, so the undo behavior you're seeing is being caused by some calls that are made inside the merge operation (`createTCO()`, `tcov->remove()` and I guess `addNotes()`). Merging works this way:

1) A new empty clip is created
2) There's a loop going through each merged clip, writing their notes on the new clip and deleting them

When you merge 2 clips and hit CTRL+Z, it first undo the clips being removed and then the new empty clip being created. You can confirm it by hitting CTRL+Z only once and then moving the clip: you'll see the two clips you had before below it.

I'm not sure it's worth it trying to fix it, specially because the history system is currently in a very unstable state already (I'd advise against using it at all and instead constantly saving the project as a new version) and is among the plans of things to be changed in the future. But I'll take a better look at it!

Spekular:
You should get most of the way to a correct undo by calling `addChecpoint` and `saveJournallingState(false)` at the start of the merge function, then `restoreJournallingState` at the end. I think it'll be a while until the undo/redo system is overhauled, and until it is we can't just let undo support devolve into a useless state.

IanCaio:
You're right, I'll try to make it work here!

IanCaio:
The history system was fixed for the merge operation, undo and redo should work fine now. I've a small doubt whether it's a good idea to restore the journalling state of a TCO after removing it, when I disabled it before removing. For now, I'm restoring the state but I left a TODO comment to remind me of removing the method call if it's unnecessary.

superpaik:
It works fine now, both undo as redo. I've tried with like 100 different patterns and it worked!

IanCaio:
Awesome, glad it's working now! Thanks for testing man!

IanCaio:
Merged master and removed line that is now unnecessary since the bug fix from #5699 .
Also fixed the code style.

IanCaio:
Fixed another conflict introduced by #5806 , changed the pointer declarations so they adhere to `Type* var;` code style and removed a line that I believe was unnecessary.

src/gui/TrackContentObjectView.cpp
```diff
@@ -948,6 +955,8 @@ void TrackContentObjectView::contextMenuEvent( QContextMenuEvent * cme )
 
 	QMenu contextMenu( this );
 
+	bool canMergeTCOs = individualTCO ? false : canMergeSelection(selectedTCOs);
```
PhysSong, line 958:
```suggestion
	bool canMergeTCOs = !individualTCO && canMergeSelection(selectedTCOs);
```

src/gui/TrackContentObjectView.cpp
```diff
@@ -1106,6 +1127,122 @@ void TrackContentObjectView::toggleMute( QVector<TrackContentObjectView *> tcovs
...
+		// Set the previousOwnerTrackView to the first TrackView
+		if (!previousOwnerTrackView)
+		{
+			previousOwnerTrackView = ownerTrackView;
+		}
+
+		// If there are TCOs from different tracks or TCOs from tracks
+		// other than an InstrumentTrack, can't merge them
+		if (ownerTrackView != previousOwnerTrackView
+			|| !dynamic_cast<InstrumentTrackView*>(ownerTrackView))
```
PhysSong, line 1156: Instead of trying `dynamic_cast` for every `TrackView`s, you can check the type of the first `ownerTrackView` and then compare with the other `ownerTrackView`s without type checking.

IanCaio: True! I'll move the `dynamic_cast` to the `if (!previousOwnerTrackView)` conditional and return false if it fails

src/gui/TrackContentObjectView.cpp
```diff
@@ -1106,6 +1127,122 @@ void TrackContentObjectView::toggleMute( QVector<TrackContentObjectView *> tcovs
...
+		{
+			canMerge = false;
+			break;
+		}
+	}
+
+	return canMerge;
+}
```
Spekular, line 1164: Simplify the logic by outsourcing it to a set data structure? Probably no noticeable performance impact.
```suggestion
bool TrackContentObjectView::canMergeSelection(QVector<TrackContentObjectView*> tcovs)
{
	bool canMerge = tcovs.size() > 1 &&
		!dynamic_cast<InstrumentTrackView*>(tcovs.at(0)->getTrackView());

	QSet<TrackView*> parentTracks;
	for (auto tcov: tcovs){ parentTracks.insert(tcov); }

	return canMerge && parentTracks.size() == 1;
}
```

src/gui/TrackContentObjectView.cpp
```diff
@@ -1106,6 +1127,122 @@ void TrackContentObjectView::toggleMute( QVector<TrackContentObjectView *> tcovs
...
+	TimePos currentPos = earliestPos;
+
+	for (auto tcov: tcovs)
+	{
+		currentPos = tcov->getTrackContentObject()->startPosition();
+		if (currentPos < earliestPos)
+		{
+			earliestPos = currentPos;
+		}
+	}
```
Spekular, line 1193: Something like this should work?
```suggestion
	const auto earliestTCOV = std::min_element(tcovs.begin(), tcovs.end(),
		[](TrackContentObjectView* a, TrackContentObjectView* b){
			return a->getTrackContentObject()->startPosition() <
			b->getTrackContentObject()->startPosition;
		}
	);
	const TimePos earliestPos = earliestTCOV->getTrackContentObject()->startPosition();
```

Spekular (commented):
LGTM. Suggested a few potential changes to make things shorter, but I haven't tested them to ensure they're totally correct so there might be syntax issues and such.

src/gui/TrackContentObjectView.cpp
IanCaio: Are Qt iterators guaranteed to be compatible with methods that expect `std::iterator`s?

Spekular: I think so? https://doc.qt.io/qt-5/containers.html#stl-style-iterators

These can probably be constBegin and constEnd by the way.

IanCaio: Cool! I thought they should be, but was asking to make sure. I'm trying to use `std::set` on the other suggestion btw, just to follow that guideline of avoiding Qt where it isn't needed (even though `QVector` is being used for the selection, but that's something for another PR :laughing: )

src/gui/TrackContentObjectView.cpp
IanCaio: Done! Used `std::set` instead and removed the `tcovs.size() > 1` since that was checked for on the `contextMenuAction`. Thinking about moving it here though, probably belongs better to this method.

Spekular: Yeah, that's why I included it, feels incorrect for canMerge to return true for a single clip.

IanCaio: Agreed, it wasn't a problem before because canMergeSelection was only called with multiple TCOs, but this check suits the method better. Changed on the last commit!

src/gui/TrackContentObjectView.cpp
```diff
@@ -1106,6 +1127,100 @@ void TrackContentObjectView::toggleMute( QVector<TrackContentObjectView *> tcovs
...
+
+	// Update length since we might have moved notes beyond the end of the pattern length
+	newPattern->updateLength();
+	// Rearrange notes because we might have moved them
+	newPattern->rearrangeAllNotes();
+	// Restore journalling states now that the operation is finished
+	track->restoreJournallingState();
+	newPattern->restoreJournallingState();
+	// Update engine
```
PhysSong, line 1219:
```suggestion
	// Update song
```

src/gui/TrackContentObjectView.cpp
```diff
@@ -1106,6 +1127,100 @@ void TrackContentObjectView::toggleMute( QVector<TrackContentObjectView *> tcovs
...
+		tcov->remove();
+	}
+
+	// Update length since we might have moved notes beyond the end of the pattern length
+	newPattern->updateLength();
+	// Rearrange notes because we might have moved them
+	newPattern->rearrangeAllNotes();
+	// Restore journalling states now that the operation is finished
+	track->restoreJournallingState();
+	newPattern->restoreJournallingState();
```
PhysSong, line 1218: Not important, but I want to change the order of two calls because `saveJournallingState`/`restoreJournallingState` works like push/pop.
```suggestion
	newPattern->restoreJournallingState();
	track->restoreJournallingState();
```

IanCaio: Done! Does that mean that when I called `track->restoreJournallingState();` I was actually restoring `newPattern`s journalling state?

PhysSong: It's rather a stylistic/semantic point, nothing was working wrong.

IanCaio: Oh, got it now! Thanks!

PhysSong (approved):

Spekular (approved):

IanCaio:
Sure, I think it's good to go! And it has 2 approvals now :smile:

PR #5695, opened 2020-10-03
Specify DirectConnection for a few plugins

Firepal:
Fixes #5665.
Continues #4942. (somewhat, at least)

(If i understand this problem right,) some knob connections were automatically set to QueuedConnection, which made them update poorly and thus caused audible "snapping" with knob values in knobs that were automated. Manually setting those connections to DirectConnection fixes that.

The following plugins were changed:

- TripleOscillator
- BitInvader
- AudioFileProcessor
- **Monstro**
- **Nescaline** (haven't tested these last two, but for good measure...)

PR #5701, opened 2020-10-05
Don't reload sample from disk when reversing

DomClark:
Currently, when reversing a sample, `SampleBuffer` reloads it from disk. This is unnecessary and makes it impossible to do so in realtime code. This change still isn't realtime safe, since it locks a mutex, but a proper fix requires much wider changes, and it doesn't read from the disk any more at least.

Blocks #5695.

### DIFF
--- a/src/core/SampleBuffer.cpp
+++ b/src/core/SampleBuffer.cpp
@@ -24,6 +24,8 @@
 
 #include "SampleBuffer.h"
 
+#include <algorithm>
+
 #include <QBuffer>
 #include <QFile>
 #include <QFileInfo>
@@ -1412,7 +1414,12 @@ void SampleBuffer::setAmplification( float _a )
 void SampleBuffer::setReversed( bool _on )
 {
 	m_reversed = _on;
-	update( true );
+	Engine::mixer()->requestChangeInModel();
+	m_varLock.lockForWrite();
+	std::reverse(m_data, m_data + m_frames);
+	m_varLock.unlock();
+	Engine::mixer()->doneChangeInModel();
+	emit sampleUpdated();
 }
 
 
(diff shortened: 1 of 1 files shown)


Spekular:
Code looks reasonable to me! Do you want any testing on this? Anything specific?

Spekular:
Wait, actually, doesn't this assome that every call to setReversed toggles the state? It appears to me that `_on` is meant to define if the sample should be reversed or not, so repeated calls to `setReversed` with the same argument should do nothing, right? Whereas here the sample would be reversed each time.

DomClark:
Good catch! I was looking at similar code by Reflexe, but didn't quite take in that he changed the function from `setReversed(bool)` to `reverse()`. Should be fixed now. (Also, `m_reversed` is now set while the lock is held, which should be safer.)

In terms of testing, I don't think an awful lot is required. Flip some samples back and forth in AFP, make sure they save and load the same way round.

JohannesLorenz (approved):
LGTM

Still not RT safe, but a great performance increase.

Spekular:
Did some very quick testing, everything worked as it should. I would probably have wrapped the whole function inside `if (m_reversed != on) { ... }` to avoid taking the locks if they're not needed, but in practice the `if` should always be true so it won't make any difference.

Spekular (approved):

DomClark:
> I would probably have wrapped the whole function inside if (m_reversed != on) { ... } to avoid taking the locks if they're not needed

If this function is going to be called from multiple threads, the variable should be read and written under the same lock. Suppose two threads tried to reverse the sample simultaneously - both see that it's not reversed, so both reverse it and it ends up unreversed, even though `m_reversed` is now `true`.

Anyway, this has two approvals now, so I will merge it tomorrow if there are no objections.

Spekular:
Ah, great point.

PR #5544, opened 2020-06-23
mB_prJune23_2020

musikBear:
I have 
* Cleared repo-root for 2 duplicate files.
* Changed logic for Instrument-envelopes, in order to preserve AFP-sample backward comparability.
* Added 7 features:

General: 
1. New piano-like envelope for general instrument (Part of LMMS#2092

Piano-roll:

2. Humanize note position (speedbar-button) (Part of LMMS#1165
3. Humanize note velocity (speedbar-button) (Part of LMMS#1165
4. Strum from top (speedbar-button) (LMMS#2380
5. Strum from bottom (speedbar-button) (LMMS#2380
6. Shorten notes by 1/192 ( key - ) (LMMS#3138
7. lengthen notes by 1/192 ( key + ) (LMMS#3138

Methods demo: https://youtu.be/OlBtU0JYvyY

PR #2483, opened 2015-12-12
Add Quantize button to Piano Roll (WIP)

grejppi:
# WORK IN PROGRESS

Inspired by a conversation on #lmms:

```
14:03 < klopsi-u3> is there a way to force an already recorded track to a lower 
                   quantisation?
...
14:42 < grejppi> klopsi-u3: none in the interface
14:42 < grejppi> that would be a useful feature though
14:44 < klopsi-u3> ok
14:45  * grejppi might hack that in
14:45 < klopsi-u3> :)
14:45 < grejppi> or at least try
14:46 < grejppi> I've already been sidetracked by so many things today c:
14:50 < klopsi-u3> i think there should be a seperate button for 'quantise 
                   piano roll'
14:50 < klopsi-u3> iow, not automatically shift quantisation if the 
                   quantisation dropdown is changed
14:51 < grejppi> yeah
14:51 < klopsi-u3> maybe even only quantise selected notes
```

There is now a new button in the Piano Roll editor toolbar:

toolbr

If there is a selection, it will quantize the selected area to the current quantization level (1/4 in these screenshots for clarity):

quantize_sel

Without a selection the entire pattern is quantized.

quantizeproper

## Known issues:
* ~~As you can see from the [last screenshot](https://cloud.githubusercontent.com/assets/772336/11762384/484eefb8-a0ed-11e5-9e53-b93a866bd1db.png), there are still cases where some notes are left as-is. I will investigate this.~~
* This has absolutely not been tested yet. At least it builds for now.

PR #5702, opened 2020-10-05
Icon changes in piano roll and song editor

allejok96:
## Piano roll
Is it just me, or is [quantize](https://www.thefreedictionary.com/quantize) a really weird word? Could we use a simpler to understand term? Since the quantize button only aligns the start of the note to the grid, it doesn't seem like "true" quantization to me.

bild

I know the renaming of quantize could be controversial since the word is used in other DAWs. But that doesn't make it less obscure.

The "scale" icon with ascending boxes is not helpful. Change it to something music related.

bild

Rename "Note lock" to "Same as note" for the quantize/grid size setting.

## Song editor
Clip snapping is not quantization, thus should not have a Q icon.

bild


Spekular:
Disagree with all points relating to quantization (1 & 3).
- The fact that the quantize button doesn't quantize note lengths is a bug. It should be fixed, not renamed.
- If you're working with 3D, you'll know the words vertex, edge, face, geometry, mesh, etc. If you're working with audio, you should know the word quantize.
- The "clip snapping" is indeed quantization. Quantizing the position of clips is no different from quantizing the positions of notes.

That being said, I do think the icons could be changed to something more iconographic. I would defer to @RebeccaDeField and see if they have opinions on/suggestions for replacing the Q in at least some of these places.

allejok96:
Alright, I understand that quantize is a thing.

> Quantizing the position of clips is no different from quantizing the positions of notes.

I have to disagree with this, from a users perspective.

In piano roll quantization is something you can apply, but in song editor quantization is just a part of the workflow of using a grid. Maybe the computer does the same thing in its math, but as a user I don't need to know that. In the end of the day, every digital representation of something analog is quantized.

Spekular:
> in song editor quantization is just a part of the workflow of using a grid

This is identical to the piano roll. Changing the quantization dropdown in piano roll affects how far notes move when shifted, just as changing the quantization dropdown in the song editor affects how far clips move when shifted.

allejok96:
@Spekular you are technically right.

Here is what I think:
- Change the scale icon in piano roll. (Since it sits next to the Key dropdown, I thought that a symbol from a musical key signature was a good choice, but we'll see if Rebecca has another idea.)
- Keep the Q button in piano roll. Quantize is an accepted term in music editing.
- The Q button should alter note length. This needs to be fixed.
- Keep the Q dropdown in piano roll. While it does not strictly enforce quantization (you can still write shorter note if you want), it is enough related to the quantizing that having a Q icon for it is justified. There is also visual feedback when changing this option, so the user will learn quickly what the Q means.
- Change the icon of the Q dropdown in song editor. Here is why:

  - Yes, it does the same as Q in piano roll, but there is no visual feedback. The user won't understand what it does.
  - I believe the concept of bulk-quantizing TCOs may be more uncommon than quantizing notes. So there is no need to make the connection to quantizing in this context.
  - You yourself spontaneously named this button "clip snapping size", both in UI and in code.
  - In #4973 you call it quantizing a few times but everyone else says snapping.
  - @JohannesLorenz (hi there) even asked what the Q had to do with snapping. I am not alone :grinning:

So you see, you nailed it when you called it snapping in the first place. It is a good term, because others instantly understand what you mean. Unless you're writing a doctoral thesis this is the case: Comprehensible > Academically correct.

... I'll look into that Q button and note lengths...

allejok96:
Looks like the note length Q was implemented in #2483 but never made it to the UI. Maybe both position/length Q are useful. Question is how do we add it to the UI? I dislike clutter.

Spekular:
> Question is how do we add it to the UI? I dislike clutter.

Something like a combobox would work IMO.

tresf:
I find the artwork proposed in this PR to be much more intuitive versus what we use today.  Pinging @RebeccaDeField since she's done up most of the artwork we use in our current theme.

> > I know the renaming of quantize could be controversial since the word is used in other DAWs. But that doesn't make it less obscure.
> Disagree with all points relating to quantization (1 & 3).

Why not both?  Call it "Quantize", as it is, but also explain the action in which it performs.

The only criticism I have with this PR is it should have a clear before/after.  Also, the "Q" is used in multiple areas, the PR should reflect them both.

tresf:
> **Title:** Different terminology for quantize and more

I would say that the bigger chance is the improved artwork.  Please consider updating the title to reflect. :)

allejok96:
> The only criticism I have with this PR is it should have a clear before/after.

I'm sorry, I don't quite follow. You mean like a comparison with how it was originally?

> the before/after should reflect them all/both.

... guess I've covered most of the "after" right?

> the "Q" is used in multiple areas

yes, so btw this means changing the code to use different icons in all of these places (thus breaking custom themes)

allejok96:
So both @Spekular and @tresf likes to call it Quantize whether or not the note length changes... How about

bild
**Quantize positions only**
...if you get it

allejok96:
> Something like a combobox would work IMO.

c2

Before going into more details, I wanna know should combo button widget be a separate PR?

tresf:
> > before/after
> You mean like a comparison with how it was originally?

Correct.

> ... guess I've covered most of the "after" right?

The mockups are very nice, yes, but just looking at the current UI, I spot at least 3 places that will be affected.  If the Q goes unchanged, pointing these out probably isn't necessary but if the Q is changed to the new (and in my opinion, improved) artwork, a before/after is warranted.



> yes, so btw this means changing the code to use different icons in all of these places (thus breaking custom themes)

No, I'm sorry what I said was confusing.  I was under the assumption it would be changed everywhere.  As an aside, custom themes will always fall mercy to UI changes but I don't believe this to fall into that category.

Spekular:
> Before going into more details, I wanna know should combo button widget be a separate PR?

If you're adding a new widget type then my personal and current inclination is yes, make that it's own PR for (hopefully) faster and easier review. However, what I was thinking of was basically just JavaFX's SplitMenuButton, so I'd be a bit surprised if Qt didn't have anything like that built in. The idea is that the combo box would select an action, and the button would perform the latest selected action.

tresf:
Adding @RebeccaDeField and @Spekular as reviewers.
* @Spekular's request to keep "Quantize" needs decisions, which I feel he can make final decision on if no one objects
* @RebeccaDeField would need to approve the artwork before merging.

Spekular:
I would like to see "quantize positions", "quantize lengths" and "quantize all" as options in the quantization combo box, and for most of the verbiage changes to be reverted (I'm fine with "same as note" replacing "note lock" if no one else has objections). I'll shamelessly leave the icon review to Rebecca, but I will say I'm not sure "flat" is any more appropriate than "series of notes" for the scale box.

Spekular (changes requested):
See latest comment in thread.

allejok96:
@Spekular, your combo button is here.

bild

Now I can't think of any good reason why you'd want a _quantize length only_ button. I don't think we should add functionality without motivation...

Spekular:
> Now I can't think of any good reason why you'd want a _quantize length only_ button. I don't think we should add functionality without motivation...

I can, that's why I suggested it. Regardless, in this case there's really no downside to including it IMO, even if it were to be "useless". Anyway, lots of sounds are very responsive to note length, so quantizing length allows you to quickly ensure all your notes are at the same length or multiples of a length. Ex. you have an LFO on your sound and want it to stop at the right time.

tresf:
@Spekular / @RebeccaDeField i sort of liked the proposed Quantize artwork.  My opinion shouldn't hold a bunch of weight, but do you both feel that it should go back to the 'Q' logo as the most recent screenshots illustrate?  Again, I'm not the decider here, I just fine the original artwork to a bit more favorable personally and I didn't know if it was requested to be reverted.

RebeccaDeField:
I will review the options for this and give feedback at my soonest availability. Things are very busy on my end right now but it's on my radar for when I get a free moment.

RebeccaDeField:
Thank you all for your patience :pray: 

First off, I'm in favor of removing any letters and replacing them with real visual cues for both translation purposes and to remove redundancy from the title already being shown on hover. The point of our icons is to hopefully give more context than what the text can provide.

The magnet icon was quickly recognizable to me :+1: 

For the scale box, we have these visual comparisons currently:

Screenshot from 2020-12-06 17-42-35

I'm not opposed to changing these to be less box-like if it seems confusing, but I think the chord and scale icons should remain coherent with each other.

Due to the discussion covering many icon changes, am I following that these are the only two in question, or are we still considering separate "grid-snapping" and "grid size" updates?

allejok96:
Just as a recap, this is the current status of the PR... yes, it's a bit messed up.

bild

We have 2 votes for switch to the originally suggested magnet icon, and 1 vote to keep the Q icon. If we would switch to magnet icon we would have to change the icon for "grid size" in Piano roll (as shown in the pics)...

> I think the chord and scale icons should remain coherent with each other.

I agree from an aesthetic perspective, that the icons should have a common theme. But logically they are not so closely related. Key/Scale sets a *highlight* in the view, while Chord creates a chord when *clicking*.

> I'm not opposed to changing these to be less box-like if it seems confusing

Boxes as a Chord icon is perfect, it creates a stack of (box-like) notes. WYSIWYG. The confusing one is the Scale icon, because to me it means "Strum from top" (see also #5544)...

tresf:
> We have 2 votes 

For the record, it's not a democratic process, @RebeccaDeField decides.  She can weigh opinions as she chooses.

allejok96:
Alright guys, I know you don't want to mess around with these discussions over non-issues, so I'll try to make it as smooth as possible so we can get rid of this PR.

- The magnet got thumbs up so I brought that back.
- No special icon for Grid size (just plain magnet).
- Revert back to the old boxlike Scale icon.
- Grid snap size icon remains changed (still needing input from @RebeccaDeField on that one)

Here are the visuals before and after (compared to master branch):

bild

bild

RebeccaDeField:
@allejok96 thanks for your work to make the edits clear and proposed changes clarified. I'm going to re-create the new icons as SVGs so I will update this thread with some new versions based on what we discussed here soon.

allejok96:
Currently - the master branch has added a wrench button:

bild

Suggestion - make them look the same:

bild

Spekular:
@allejok96 Could you extract the enhanced quantize feature to a separate PR? That can be merged regardless of any icon discussion and should really be a separate change.

RebeccaDeField:
Thanks for the suggestion, all of the icons need to be created in SVG and I can do that in a separate PR. Let's move forward with the other updates and I will take care of the icon improvements when I can.

allejok96:
Closing this. The quantize functionality is fixed in #5946, devs have other things to think about than choice of icons and users will figure things out either way.

Issue #3963, opened 2017-11-10
Submodule candidates


tresf:
#### Candidates

| PR | Plugin | Upstream | Comments |
|-------|-------|------------|-----------|
| #3873 | core/rpmalloc | github.com/rpmalloc/rpmalloc | ✅  Done |
| #3931 | ladspa/swh | github.com/swh/ladspa | ✅ Done |
| #3960  | ladspa/tap | github.com/tomszilagyi/tap-plugins | ✅ Done |
| #3940 #3987  | ladspa/calf | github.com/lmms/veal  |  ✅ 🍴 Done.  Upstream API changed, need to use fork. |
| #4027 (WIP) | ladspa/caps | quitte.de/dsp/ | ⌛ In progress using moddevices Caps LV2 mirror  |
| #5755  | ladspa/cmt | github.com/lmms/cmt | ✅ 🍴 Done (our own fork, no upstream VCS)  |
| #3946   | plugins/papu | bitbucket.org/mpyne/game-music-emu | ✅  Done |
| #3962  | plugins/opl2 | github.com/adplug/adplug | ✅ Done  |
| #5724  | plugins/sid | github.com/simonowen/resid | ✅ Done.  |
| #3962  | plugins/zynaddsubfx | sourceforge.net/p/zynaddsubfx | ✅ 🍴 Done.  Upstream API changed.  Need to use fork.  |
| #3965 | plugins/xpressive | github.com/ArashPartow/exprtk | ✅  Done.  ~~Requires patch file for MacOS.~~ Not anymore, upstream patched.  |
| #5713 | plugins/carlabase | github.com/falktx/carla | ✅  Done (weak linking, like weakjack) |
| #3970  | audio/weakjack | github.com/x42/weakjack | ✅ Done.  |
| #5716 | 3rdparty/jackaudio/jack | github.com/jackaudio/headers | ✅ Done. |
| #4588 | wiki | github.com/lmms/lmms.wiki.git | ✅ Done. | 

#### Potential candidates

| Plugin | Upstream | Comments |
|--------|----------|------------|
| plugins/ReverbSC| github.com/csound/csound | ⚠️  42MB library, better candidate for `subtree`? |

Are there any I'm missing?

Related #3957

JohannesLorenz:
While I welcome a clonable wiki, I'm not sure if a submodule is justified. LMMS is already full of submodules, making cloning a hell if your internet is limited. Though, some users may user the shallow clones.

[...]

PR #5716, opened 2020-10-16
Add Windows JACK support

### LINKED ISSUE #2136: Jack audio support on windows
I don't know if this is the proper place for a feature request like this, but eventually it would be cool to have [jack](http://jackaudio.org/) support on windows. There are obviously much more pressing things to work on though.

tresf:
JACK is surprisingly plug-and-play with regards to LMMS on Windows, thanks to the hard work of @falkTX and the rest of the JACK2 team combined with our already existing `weakjack` (thanks @x42) module (preventing the need for windows dev libs).

* Adds JACK build support for Windows systems
* Removes libjack-dev dependency unless specified
* Adds jack2 as a submodule, related #3963 
* Bumps weakjack submodule to add MSVC support.

To test:

* Install JACK for Windows from the website
* Launch JACK with ASIO support per JACK's website instructions
* Switch LMMS backend to JACK
* Configure using `qjackctl`, found in JACK install directory

image

Closes #2136


PhysSong:
The `weakjack` submodule should be updated to support MSVC, see https://github.com/x42/weakjack/pull/9.

src/3rdparty/weakjack/CMakeLists.txt
```diff
@@ -7,6 +7,6 @@ IF(LMMS_HAVE_WEAKJACK)
 	
 	# Library stub for AppImages running on systems without jack
 	ADD_LIBRARY(weakjack MODULE weakjack/weak_libjack.c weakjack/weak_libjack.h)
-	TARGET_INCLUDE_DIRECTORIES(weakjack PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/weakjack ${JACK_INCLUDE_DIRS})
+	TARGET_INCLUDE_DIRECTORIES(weakjack PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/weakjack ${CMAKE_CURRENT_SOURCE_DIR}/../jack2/common ${JACK_INCLUDE_DIRS})
```
PhysSong, line 10: Could you explain why did you add this?

falkTX: because using weakjack the real jack is not needed, only the header files.

but in that case the right submodule is actually https://github.com/jackaudio/headers

x42: For Ardour windows builds, we use http://ardour.org/files/deps/jack_win3264.tar.xz

It's a re-zip of jack headers, .def, .lib, and  32/64bit .dll  from jackd 1.9.10 with a pkg-config file added for convenience,
The files were made using sletz' last release of jack for windows back in 2012:
* https://dl.dropboxusercontent.com/u/28869550/Jack_v1.9.10_32_setup.exe
* https://dl.dropboxusercontent.com/u/28869550/Jack_v1.9.10_64_setup.exe

With weakjack one only needs the headers, but the other files come in handy for unit-tests.

tresf: > Could you explain why did you add this?

This part seems answered.

> the right submodule is actually https://github.com/jackaudio/headers

Done via b852863, but for the record, now the submodule folder is mismatched from the project name and this is inconsistent with the [other submodules we clone](https://github.com/LMMS/lmms/issues/3963) due to needing a `jack/` prefix which `jack2` was already providing but I digress. :)

> With `weakjack` one only needs the headers, but the other files come in handy for unit-tests.

We're moving most of our Windows system to use `vcpkg` bootstrapping and building on-the-fly.  I don't love this strategy from a time-to-build-perspective, but it certainly avoids the pre-built fetching on Windows, which in 2020 still seems to lack any consistent dependency system.  For the scope of this PR this is all moot since we don't have JACK unit tests (yet) but I'm sure this will come back up in the future.  Anyway, thanks kindly for sharing.  I had initially planned on making a [viable dependency system using precompiled binaries and scoop.sh](https://github.com/lukesampson/scoop-extras/pulls?q=is%3Apr+author%3Atresf+is%3Aclosed), but never came to fruition.  For now, `vcpkg` seems to be the least cluster**** solution.  I mention all of this because as a project, we're always interested in how others do it.  LMMS has had a rocky history of building on Windows and we greatly appreciate the info.

tresf: @falkTX the recommended header repo throws the following error in AppVeyor:

```
C:\projects\lmms\src\3rdparty\jackaudio\jack/types.h(25): fatal error C1083: Cannot open include file: 'pthread.h':
```

Which seems to be referenced here:
https://github.com/jackaudio/headers/blob/master/types.h

... but not here:
https://github.com/jackaudio/jack2/blob/develop/common/jack/types.h

Is this just a matter of the headers repo needing updating?

tresf: > > Could you explain why did you add this?

> This part seems answered.

As usually @PhysSong's right 🤦 , I had forgotten to delete the redundant include, the variable is already populated earlier.  Patching now...

tresf: Done via 9c1cf61.  Moving other conversation points to another line-number for context.

.gitmodules
```diff
@@ -40,3 +40,6 @@
 [submodule "plugins/carlabase/carla"]
 	path = plugins/carlabase/carla
 	url = https://github.com/falktx/carla
+[submodule "src/3rdparty/jackaudio/jack"]
```
tresf, line 43: @falkTX wrote:
> the right submodule is actually https://github.com/jackaudio/headers

Done via b852863, however now AppVeyor throws the following:

```
C:\projects\lmms\src\3rdparty\jackaudio\jack/types.h(25): fatal error C1083: Cannot open include file: 'pthread.h':
```

Which seems to be referenced here:
https://github.com/jackaudio/headers/blob/master/types.h

... but not here:
https://github.com/jackaudio/jack2/blob/develop/common/jack/types.h

Is this just a matter of the headers repo needing updating?  I'll also bump weakjack for MSVC support incase that's related.

src/3rdparty/weakjack/CMakeLists.txt
falkTX: > [](https://github.com/jackaudio/jack2/blob/develop/common/jack/types.h)
> 
> Is this just a matter of the headers repo needing updating?

yeah, the headers at the moment target jack1.

tresf:
> The `weakjack` submodule should be updated to support MSVC, see x42/weakjack#9.

Done via b1ab5ae.

src/3rdparty/weakjack/CMakeLists.txt
tresf: > yeah, the headers at the moment target jack1.

Should I revert then?

falkTX: yes sorry.

.gitmodules
tresf: Per @falkTX, this needs to be reverted as the `headers` repo is based off of jack1, not jack2 and has issues compiling on MSVC.  Done via
f55f42d.

tresf:
All notes have been addressed, but there's a lingering issue from #5636 preventing AppVeyor to succeed.  I'm fine waiting for that to be fixed to make sure AppVeyor is happy or merging as-is.  Thoughts welcome.   Pinging @Spekular since he authored  #5636.

Spekular:
Huh. Is this related to the issue fixed by https://github.com/LMMS/lmms/pull/5678 somehow? Other builds since then have been fine, so I'm frankly not sure why an issue would pop up now.

tresf:
> Huh. Is this related to the issue fixed by #5678 somehow? Other builds since then have been fine, so I'm frankly not sure why an issue would pop up now.

Ok, sorry for the ping, I did search beforehand. I'll fast-forward and it should fix it, thanks.

tresf:
Ok, I fast-forwarded my branch didn't fix this @Spekular, can you comment on the AppVeyor error?  It's the same as before.

PhysSong:
I think [`(std::numeric_limits<int>::max)()` may work](https://stackoverflow.com/a/2789509). I'll test it out and fix `lmms_basic.h` as well if that works.

tresf:
> I'm frankly not sure why an issue would pop up now.

@Spekular as discussed in #devtalk on [Discord](https://lmms.io/chat), it appears this is as a result of adding `jack.h` into AppVeyor, which is new with this PR.  So, although the symptom is the same, the cause is likley a 3rdparty header we don't manage.  I've made my first attempt at patching this in d56db57.  If that doesn't work, I might fire up a MSVC environment instead.

> I think `(std::numeric_limits<int>::max)()` may work.

@DomClark shared his sentiments on Discord, and I tend to agree with them that we shouldn't have to worry about adding parens to each call to `max`.  

>  I'll test it out and fix `lmms_basic.h` as well if that works.

I started writing this on a [separate branch, but was largely unsuccessful](https://github.com/LMMS/lmms/compare/master...tresf:nominmax).  I'll revisit if needed.

plugins/vst_base/RemoteVstPlugin.cpp
```diff
@@ -40,10 +40,6 @@
 
 #ifdef LMMS_BUILD_LINUX
 
-#ifndef NOMINMAX
-#define NOMINMAX
-#endif
-
```
PhysSong, line 46: Please keep this or modify relevant CMake files, we don't propagate all C++ flags when building `RemoteVstPlugin` using `ExternalProject`.

tresf:
> I'll test it out and fix `lmms_basic.h` as well if that works.

I decided to opt for a top-level... 
```cmake
ADD_DEFINITIONS(-DNOMINMAX)
```
... which I believe solves the problem proactively and holistically.

On the contrary, `lmms_basics.h` can't extend to headers which we don't manage, like `weak_jack.h`, so I believe this to be ideal.

Components which build using `ExternalProject_Add` may not import all flags, so `RemoteVstPlugin` is left as-is.

tresf:
Something's odd with Travis, it's not even queuing my builds (it's been hours and I've tried restarting it once).  I'm not holding up the PR for that specifically.  There's also a failure after rebase, but it's unrelated and explained here: https://github.com/LMMS/lmms/commit/435dbc5e00fa06cd789693a56e6b340f4c7722ab#r43742718.  Merging.

Issue #3028, opened 2016-09-09
Issues with enabling loop-points while playing and outside of the loop.


zonkmachine:
There is a bit of a glitch when you:
a. toggle loop-points to enabled, while...
b. we're already playing, and...
c. we're outside of the loop.

We will skip to tick 0 of the loop but will only play from the next tick, 1, for the first loop cycle. This doesn't happen when you toggle loop-points when playback is stopped. Also, notes playing aren't stopped which would probably be a good idea.

PR #5723, opened 2020-10-19
Rework Song::processNextBuffer

DomClark:
This was originally an attempt to fix a VST bug reported on Discord. The issue turned out to be in `Song::processNextBuffer`, which was fairly scary and convoluted, so I tried to clean it up a bit. Testing and review would be particularly appreciated, as this bit of code is rather integral to playback, and is rather complex.

Original report, from `@Ambi#0808` on Discord:
>Error report:
VST by the name of "Church Organ 2nd" by C.Hackl consistently stops playing all notes starting at bar 8 of all scheduled midi notes, when its internal reverb is enabled.
Unsure whether this problem is with the VST itself or with the LMMS program. Testing on the latest Windows 10 update, with version 1.2.1 of LMMS.
The plugin is bridged from 32-bit to 64-bit, and it happens at any tempo.
For some reason after repeatedly enabling and disabling the internal reverb, the problem is no longer apparent.
The error also occurs when adding voices on top of each other over time, such that trying to write a CmM7 chord where a voice appears each bar, playing the last and highest note B cancels all voices.
Signs point to a poorly-coded VST, but if anyone would like to test with a more recent version of LMMS as well as other DAWs it would be appreciated.
>
>Downloadable here: https://vst4free.com/plugin/622/
>
>another odd part of this bug is that it seems to behave entirely differently when played in the piano roll vs the song editor
>
>update: i've discovered that when played in the piano roll, it's not merely bar 8 of all scheduled midi notes, but simply the last bar

It turns out that the `kVstTransportChanged` was incorrectly set for the whole last bar in the song editor, due to faulty loop handling in `Song::processNextBuffer`.

The first error is here: https://github.com/LMMS/lmms/blob/0d89d64f499cab7cdbef47bf2325306a6cb47098/src/core/Song.cpp#L288-L292
`MidiTime::getTicks()` returns the total number of ticks, not the number of ticks within the current bar, so this condition is true all the time after the first bar, not just at the beginning of a new bar.

This leads to problems here: https://github.com/LMMS/lmms/blob/0d89d64f499cab7cdbef47bf2325306a6cb47098/src/core/Song.cpp#L316-L318
This is intended to run when `m_playPos[m_playMode]` is at the end of the previous bar, but `ticks` is at the beginning of the next, hence the `+ 1` correction. However, due to the previous issue, the condition is checked every tick after the first bar, and is true throughout the last bar. This leads to erroneous calls to `m_vstSyncController.setPlaybackJumped(true)`, resulting in the observed issue.

### DIFF
--- a/src/core/Song.cpp
+++ b/src/core/Song.cpp
@@ -240,7 +240,7 @@ void Song::processNextBuffer()
 	// If the playback position is outside of the range [begin, end), move it to
 	// begin and inform interested parties.
 	// Returns true if the playback position was moved, else false.
-	const auto enforceLoop = [=](const MidiTime& begin, const MidiTime& end)
+	const auto enforceLoop = [this](const MidiTime& begin, const MidiTime& end)
 	{
 		if (getPlayPos() < begin || getPlayPos() >= end)
 		{
@@ -268,20 +268,20 @@ void Song::processNextBuffer()
 	const auto framesPerTick = Engine::framesPerTick();
 	const auto framesPerPeriod = Engine::mixer()->framesPerPeriod();
 
-	f_cnt_t frameOfPeriod = 0;
+	f_cnt_t frameOffsetInPeriod = 0;
 
-	while (frameOfPeriod < framesPerPeriod)
+	while (frameOffsetInPeriod < framesPerPeriod)
 	{
-		auto frameOfTick = getPlayPos().currentFrame();
+		auto frameOffsetInTick = getPlayPos().currentFrame();
 
 		// If a whole tick has elapsed, update the frame and tick count, and check any loops
-		if (frameOfTick >= framesPerTick)
+		if (frameOffsetInTick >= framesPerTick)
 		{
 			// Transfer any whole ticks from the frame count to the tick count
-			const auto elapsedTicks = static_cast<int>(frameOfTick / framesPerTick);
+			const auto elapsedTicks = static_cast<int>(frameOffsetInTick / framesPerTick);
 			getPlayPos().setTicks(getPlayPos().getTicks() + elapsedTicks);
-			frameOfTick -= elapsedTicks * framesPerTick;
-			getPlayPos().setCurrentFrame(frameOfTick);
+			frameOffsetInTick -= elapsedTicks * framesPerTick;
+			getPlayPos().setCurrentFrame(frameOffsetInTick);
 
 			// If we are playing a BB track, or a pattern with no loop enabled,
 			// loop back to the beginning when we reach the end
@@ -313,13 +313,13 @@ void Song::processNextBuffer()
 			}
 		}
 
-		const f_cnt_t framesUntilNextPeriod = framesPerPeriod - frameOfPeriod;
-		const f_cnt_t framesUntilNextTick = static_cast<f_cnt_t>(std::ceil(framesPerTick - frameOfTick));
+		const f_cnt_t framesUntilNextPeriod = framesPerPeriod - frameOffsetInPeriod;
+		const f_cnt_t framesUntilNextTick = static_cast<f_cnt_t>(std::ceil(framesPerTick - frameOffsetInTick));
 
 		// We want to proceed to the next buffer or tick, whichever is closer
 		const auto framesToPlay = std::min(framesUntilNextPeriod, framesUntilNextTick);
 
-		if (frameOfPeriod == 0)
+		if (frameOffsetInPeriod == 0)
 		{
 			// First frame of buffer: update VST sync position.
 			// This must be done after we've corrected the frame/tick count,
@@ -329,20 +329,20 @@ void Song::processNextBuffer()
 			m_vstSyncController.update();
 		}
 
-		if (static_cast<f_cnt_t>(frameOfTick) == 0)
+		if (static_cast<f_cnt_t>(frameOffsetInTick) == 0)
 		{
 			// First frame of tick: process automation and play tracks
 			processAutomations(trackList, getPlayPos(), framesToPlay);
 			for (const auto track : trackList)
 			{
-				track->play(getPlayPos(), framesToPlay, frameOfPeriod, clipNum);
+				track->play(getPlayPos(), framesToPlay, frameOffsetInPeriod, clipNum);
 			}
 		}
 
 		// Update frame counters
-		frameOfPeriod += framesToPlay;
-		frameOfTick += framesToPlay;
-		getPlayPos().setCurrentFrame(frameOfTick);
+		frameOffsetInPeriod += framesToPlay;
+		frameOffsetInTick += framesToPlay;
+		getPlayPos().setCurrentFrame(frameOffsetInTick);
 		m_elapsedMilliSeconds[m_playMode] += MidiTime::ticksToMilliseconds(framesToPlay / framesPerTick, getTempo());
 		m_elapsedBars = m_playPos[Mode_PlaySong].getBar();
 		m_elapsedTicks = (m_playPos[Mode_PlaySong].getTicks() % ticksPerBar()) / 48;
--- a/src/core/Song.cpp
+++ b/src/core/Song.cpp
@@ -30,6 +30,8 @@
 #include <QFileInfo>
 #include <QMessageBox>
 
+#include <algorithm>
+#include <cmath>
 #include <functional>
 
 #include "AutomationTrack.h"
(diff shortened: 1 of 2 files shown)


src/core/Song.cpp
```diff
@@ -191,226 +193,159 @@ void Song::savePos()
...
 
-	// check for looping-mode and act if necessary
-	TimeLineWidget * tl = m_playPos[m_playMode].m_timeLine;
-	bool checkLoop =
-		tl != NULL && m_exporting == false && tl->loopPointsEnabled();
+	// If we have no tracks to play, there is nothing to do
+	if (trackList.empty()) { return; }
 
-	if( checkLoop )
+	// If the playback position is outside of the range [begin, end), move it to
```
M374LX, line 240: Bracket mismatch in the comment. There is an opening square bracket and a closing parenthesis: [begin, end).

Veratil: That might be showing that it's `[inclusive, exclusive)`

DomClark: Yeah, that's what I meant. If people aren't familiar with this notation, however, I'm happy to change it.

Veratil: I think it's fine since it's prefixed with "range", but that's me.

Hey-Holokin:
Original reporter of the issue here. Can confirm this pull request fixes the issue completely. No random note stops during the last bar in the piano roll!

superpaik:
Tested on 1.2.2 and found the misbehaviour (though only happend with this VST, I tested with other VST and they worked well ¿?). Reproduced with this PR and it works fine.

PauseForAffliction:
Tested with the organ plugin on 1.2.2, 32bit Windows (so no usage of a bit bridge.) Was only able to reproduce the last bar not playing within the piano roll. Was not able to reproduce >bar 8 not playing in the song editor or piano roll.

Tested again using the 32bit MSVC build from this PR. The fix seems to have worked. Last bar in piano roll plays successfully with the organ plugin. Other plugins tested perform as expected on both 1.2.2 and this build.

Spekular:
I'm hesitant to leave an approving review since I haven't thoroughly checked out the old code to see that everything except the fix is (functionally) identical, but I don't see any obviously suspicious changes either. If you want, I can probably give it a more thorough look some time. Quality wise the new code looks good, and much better than the old.

src/core/Song.cpp
```diff
@@ -191,226 +193,159 @@ void Song::savePos()
...
+		m_vstSyncController.setPlaybackJumped(true);
+		getPlayPos().setJumped(false);
 	}
 
-	f_cnt_t framesPlayed = 0;
-	const float framesPerTick = Engine::framesPerTick();
+	const auto framesPerTick = Engine::framesPerTick();
+	const auto framesPerPeriod = Engine::mixer()->framesPerPeriod();
+
+	f_cnt_t frameOfPeriod = 0;
```
PhysSong, line 271: This is "frames played in this period"(or equivalently, "frame offset in period"), so I think this name is a bit unclear. The original name is unclear in somewhat different sense, though.
How about something like `frameOffsetInPeriod`?

src/core/Song.cpp
```diff
@@ -191,226 +193,159 @@ void Song::savePos()
...
 
-	while( framesPlayed < Engine::mixer()->framesPerPeriod() )
+	while (frameOfPeriod < framesPerPeriod)
 	{
-		m_vstSyncController.update();
+		auto frameOfTick = getPlayPos().currentFrame();
```
PhysSong, line 275: Similar to above.

IanCaio (approved):
Code LGTM!

I had to learn the logic of the method as I reviewed, and I've small doubts about some details (would you mind me asking about them in Discord for the sake of learning?), but as far as the behavior of the method goes, the only difference I noticed was the way the loop points are enforced and how the time elapsed on the processed buffer is counted (aka the bugfix itself).

Great job on making the method more organized and readable!

I like the variables renaming proposed by PhysSong: I don't think their meaning is too obscure right now, but his suggestion could improve their understanding even more! There's also that small change on the lambda function we talked about earlier, but since you already said you'll look into it I'm not leaving it as a requested change (it's also not critical, just something to keep in mind in terms of code maintenance if we at some point reach C++20).

DomClark:
Merge?

Veratil (approved):

zonkmachine:
Tested and works fine. Merge.

On a related note, please see https://github.com/LMMS/lmms/issues/3028.
https://github.com/LMMS/lmms/blob/e0f50e80d3c429cc1175ff22d22d1432fd41b7b1/src/core/Song.cpp#L258-L259

 This doesn't work if the loop is activated when the song is playing and outside of the loop. The song will skip to the loop but miss the first events. Off by one thing...

DomClark:
I think it's because the frame count isn't reset when looping. The loop code only sets the ticks, so if we're not in the first frame of a tick when we jump to the beginning of the loop, the first tick's events will be skipped. The other loop checking code only ever runs at the first frame of a tick, so isn't affected.

src/core/Song.cpp
```diff
@@ -191,226 +193,159 @@ void Song::savePos()
...
+			}
+			else if (m_playMode == Mode_PlayPattern && m_loopPattern && !loopEnabled)
+			{
+				enforceLoop(MidiTime{0}, m_patternToPlay->length());
 			}
-			m_playPos[m_playMode].setTicks( ticks );
 
-			if (checkLoop || m_loopRenderRemaining > 1)
+			// Handle loop points, and inform VST plugins of the loop status
+			if (loopEnabled || m_loopRenderRemaining > 1)
```
PhysSong, line 298:
```suggestion
			if (loopEnabled || (m_loopRenderRemaining > 1 && getPlayPos() >= timeline->loopBegin()))
```
This will be explained in the comment below.

src/core/Song.cpp
```diff
@@ -191,226 +193,159 @@ void Song::savePos()
...
 				m_vstSyncController.startCycle(
-					tl->loopBegin().getTicks(), tl->loopEnd().getTicks() );
+					timeline->loopBegin().getTicks(), timeline->loopEnd().getTicks());
 
-				// if looping-mode is enabled and we have got
-				// past the looping range, return to the
-				// beginning of the range
-				if( m_playPos[m_playMode] >= tl->loopEnd() )
+				// Loop if necessary, and decrement the remaining loops if we did
+				if (enforceLoop(timeline->loopBegin(), timeline->loopEnd())
```
PhysSong, line 304: This adds `m_playPos[m_playMode] < timeline->loopBegin()` condition for jumping back which is undesired in case `m_loopRenderRemaining > 1`, and make the area before the loop point be skipped or play wrongly. A relevant bug was reported [on Discord](https://discord.com/channels/203559236729438208/207159523616489473/782842838063775765).
Also, the `startCycle()` call should be checked by `m_playPos[m_playMode] >= timeline->loopBegin()`, though it was already wrong before this PR.
The proposed fix is in the comment above.

src/core/Song.cpp
```diff
@@ -191,226 +193,159 @@ void Song::savePos()
...
-		{
-			// update VST sync position after we've corrected frame/
-			// tick count but before actually playing any frames
-			m_vstSyncController.setAbsolutePosition(
-				m_playPos[m_playMode].getTicks()
-				+ m_playPos[m_playMode].currentFrame()
-				/ (double) framesPerTick );
-		}
+		const f_cnt_t framesUntilNextPeriod = framesPerPeriod - frameOffsetInPeriod;
+		const f_cnt_t framesUntilNextTick = static_cast<f_cnt_t>(std::ceil(framesPerTick - frameOffsetInTick));
```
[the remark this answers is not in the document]
PhysSong: A situation where it matters is:
1. Add a note pattern starting at bar 3
2. Add a note starting at the beginning of that pattern
3. Locate the loop begin point at bar 3, but don't enable it yet
4. Start playing song at bar 1
5. Now enable looping
6. The play head jumps to bar 3, but the note is not played

As I mentioned above, it's not a regression, but an old bug. You can ignore this.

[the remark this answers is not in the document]
DomClark: `frameOffsetInTick` is correct, in that it still corresponds to the ticks of the current `PlayPos`. The ticks of the `PlayPos` should be reset to zero/fmod'd with one though (can't decide which is correct). The issue is #3028. (Zonkmachine mentioned this here: https://github.com/LMMS/lmms/pull/5723#issuecomment-734456621.)

src/core/Song.cpp
DomClark: Fix looks good to me and seems to work properly; PR open at #5814.

src/core/Song.cpp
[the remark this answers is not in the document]
PhysSong: Oh, I see. Thanks for mentioning the relevant issue.

PR #5724, opened 2020-10-21
Switch SID to submodule

tresf:
Switches SID to a resid submodule kindly maintained by @simonowen.
* Rename code occurrences of `cSID()` to `SID()`
* Parse `VERSION` from `resid/CMakeLists.txt`

Related #3963


plugins/Sid/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
+INCLUDE(BuildPlugin)
+
+INCLUDE_DIRECTORIES(resid)
+
+BUILD_PLUGIN(sid
+	SidInstrument.cpp
```
tresf, line 6: The intent was to use `Sid.cpp` and name Class=Plugin like we've done for the other submodules, but I was afraid that Windows machines would confuse `sid.h` and `Sid.h`, so I kept the `Instrument` prefix.

plugins/Sid/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
+INCLUDE(BuildPlugin)
+
+INCLUDE_DIRECTORIES(resid)
```
PhysSong, line 3: If you parse `resid/CMakeLists.txt` below, why did you include that here?

tresf: I don't understand the question.  Can you please rephrase?

tresf: Are you confusing `INCLUDE_DIRECTORY` with `ADD_SUBDIRECTORY` perhaps?

PhysSong: Oh, yes. My mistake.

PR #5727, opened 2020-10-22
Fix rule of three for SampleBuffer, SampleTCO

Spekular:
Custom destruction, copy constructors, and copy assignment should usually all be implemented for a class that implements any one of them. These changes are needed for #5524, but I decided to move them to a separate PR to split the review burden, especially since it's important (and tricky, IMO) to get these correct.

Status as of latest commit
Operator | SampleBuffer | SampleTCO
--|--|--
Copy Constructor | Added | Added
Copy Assignment | Added | Deleted


DomClark:
The rule of three (or five these days, now we have move constructors and assignment operators) doesn't say that objects have to be copyable or moveable; rather it says that if any of these functions are user-defined, the others should be user-defined too, because the compiler-generated ones are likely to be incorrect. If you write a destructor, you are probably trying to dispose of a resource when the object is destructed. The compiler-generated constructors and assignment operators will copy the resource handle, meaning it will be disposed of twice, and the compiler-generated assignment operators will overwrite the existing handle without disposing of it. If a particular constructor or assigment operator doesn't make sense, or you're not sure what to do at the moment, `= delete` is a perfectly acceptable definition. The goal is primarily to get rid of the compiler's own version.

`SampleBuffer::operator=` leaks the original data. You might like to look into the copy-and-swap idiom as a way to implement this.

Spekular:
Will fix merge conflicts as soon as I know that there aren't bigger issues remaining!

src/core/SampleBuffer.cpp
```diff
@@ -131,6 +131,79 @@ SampleBuffer::SampleBuffer( const f_cnt_t _frames )
...
+
+
+
+
+void swap(SampleBuffer& first, SampleBuffer& second) noexcept
+{
+	using std::swap;
+
+	// Lock both buffers for writing
+	// TODO: No guaranteed ordering, risk of deadlock?
```
IanCaio, line 164: I'm not sure I see the possibility of a deadlock being caused by the order those are called (them being different objects). But I do see a possibility of a deadlock being caused by the swap method being called from inside another method that already locked the object for writing, since right now the lock is not recursive. I see two options to handle this:

1) Adding an explicit initialization of the QReadWriteLock making it recursive on the SampleBuffer constructor. That way if we call the copy assignment (or swap method) from a method that locked for write, we wouldn't deadlock. We also would have to be careful to not call the copy assignment from inside a method that locked for reading without unlocking it first (because the recursive lock only works for locks of the same type: write->write->write or read->read->read, __not__ read->write).

2) Keeping the lock non-recursive and moving the responsibility of unlocking before calling the copy assignment to the method that calls it.

I think @DomClark and @PhysSong might have better insight than me on that one though.

IanCaio (changes requested):

src/core/SampleBuffer.cpp
Spekular: Thread A calls `swap(x, y)` while Thread B calls `swap (y, x)`, each thread grabs one lock and waits for the other. I thought I remembered these locks already being recursive, but you're right that I should make sure they are if they aren't.

src/core/SampleBuffer.cpp
```diff
@@ -131,6 +131,79 @@ SampleBuffer::SampleBuffer( const f_cnt_t _frames )
...
+	m_reversed(orig.m_reversed),
+	m_frequency(orig.m_frequency),
+	m_sampleRate(orig.m_sampleRate)
+{
+	orig.m_varLock.lockForRead();
+	//Deep copy m_origData and m_data from original
+	memcpy(m_origData, orig.m_origData, m_origFrames* BYTES_PER_FRAME);
+	memcpy(m_data, orig.m_data, m_frames* BYTES_PER_FRAME);
+	orig.m_varLock.unlock();
+}
```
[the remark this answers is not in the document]
Spekular: IIRC, doesn't the lock only protect m_data? And the argument's m_data is neither read nor written in the initializers.

src/core/SampleBuffer.cpp
IanCaio: Oh true, hadn't thought of that. I'm not sure what would be the best way to handle this situation, maybe a static lock to prevent two `swaps` from happening at the same time?

I think they aren't recursive because I haven't found an initialization of `m_varLock`, so it's probably using the default constructor which is non-recursive.

src/core/SampleBuffer.cpp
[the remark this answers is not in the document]
IanCaio: I thought it was protecting the whole object, since other variables are related to `m_data` as well (like frame counts), but I'm not 100% sure to be honest, will try to investigate further!

src/core/SampleBuffer.cpp
PhysSong: Not a best solution, but you can lock one with lower memory address and then higher. And do nothing if you're swapping two identical objects.

PhysSong: Or you lock one and unlock it if you can't lock another one immediately.

PhysSong: I found that there are [`std::lock()`](https://en.cppreference.com/w/cpp/thread/lock) and [`std::scoped_lock`](https://en.cppreference.com/w/cpp/thread/scoped_lock)(available in C++17), but `QReadWriteLock` isn't compatible with them. The algorithm looks like:
- Lock 1st mutex
- Try locking 2, 3, ..., Nth mutex without blocking
  - Success: good
  - Failure: Unlock 1st, shuffle the locking order, and retry

You can try implementing this.

IanCaio: Always learning something new about locks! But yeah, I think you should be able to reimplement the algorithm used by `std::lock`, need to use `tryLockForWrite` instead then. I'm just confused as to what to do if it fails, since `swap` is `noexcept`. Just keep trying constantly? I wonder if that could in a very specific situation cause a deadlock anyways by the failures being synchronized:

- Thread A tries swap(x,y). Successfully locks X, fails to lock Y. Simultaneously thread B tries swap(y,x). Successfully locks Y, fails to lock X.
- They both unlock their locked objects and try again, but because they will try again in sync they end up in the same situation again.

I know introducing another lock might be unwanted, but maybe having a static lock called `SampleBuffer::s_swapLock`, and whichever thread calls `swap` needs to lock it, so we can be sure two swaps won't happen simultaneously.

PhysSong: It's possible to successfully lock both of them thanks to some pseudo-randomness in OS scheduling.

Spekular: I was a concerbed about noexcept as well,and also came to the conclusion that a loop is the only way to solve this, but reintroduces the risk. It feels overkill, but I'm thinking of implementing exponential backoff to handle that.

IanCaio: Interesting, didn't know about that! Is it done exclusively to avoid that kind of situation?

So @Spekular, I think the changes would be:
- Making `m_varLock` recursive by explicitly calling the QReadWriteLock constructor.
- Having this logic instead there:
```
bool successfulLock = false;
while(!succesfulLock)
{
    if (first.m_varLock.tryLockForWrite())
    {
        if (second.m_varLock.tryLockForWrite())
        {
            successfulLock = true;
        }
        else
        {
            first.m_varLock.unlock();
        }
    }
}
```

That's my first thought on how to approach, maybe there's a better way?

DomClark: > Thread A calls `swap(x, y)` while Thread B calls `swap (y, x)`

How about we just don't do this? There's no requirement for this to be possible; most other types don't support it.

Spekular: It's not like I was planning on doing it, I just can't say for sure that it will never happen.

DomClark: Other types can't say for sure that it won't happen either. C++ isn't idiot-proof - if you write silly code, you get silly results. `std::unique_ptr` is supposed to safely manage pointers, but that doesn't mean it stops you doing something like `delete p.get();` and getting hit in the face later with a double-free. I don't think it's unreasonable to say "swapping two of these objects twice simultaneously will bring pain and misery (just like most objects in C++)," and let people experience pain and misery if they try it.

This situation isn't a concern with the code as it is at the moment anyway - `swap` is currently called from `SampleBuffer`'s assignment operator, where one of the parameters is always a variable local to that method. I also feel that it's a bit out of scope - this PR is to make `SampleBuffer` and `SampleTCO` follow the rule of three, but you're trying to make a function - one that wasn't part of the original plan - do something that such functions don't typically do anyway.

src/core/SampleBuffer.cpp
```diff
@@ -131,6 +131,79 @@ SampleBuffer::SampleBuffer( const f_cnt_t _frames )
...
+	second.m_varLock.unlock();
+}
+
+
+
+
+SampleBuffer& SampleBuffer::operator=(SampleBuffer& that)
+{
+	SampleBuffer* temp = new SampleBuffer(that);
+	swap(*this, *temp);
```
DomClark, line 199: This allocates a `SampleBuffer` on the heap and leaks it. It would be better to allocate it on the stack.
```suggestion
	auto temp = that;
	swap(*this, temp);
```

include/SampleBuffer.h
```diff
@@ -109,6 +109,10 @@ class LMMS_EXPORT SampleBuffer : public QObject, public sharedObject
 	SampleBuffer( const QString & _audio_file, bool _is_base64_data = false );
 	SampleBuffer( const sampleFrame * _data, const f_cnt_t _frames );
 	explicit SampleBuffer( const f_cnt_t _frames );
+	SampleBuffer( SampleBuffer& orig );
+
+	friend void swap( SampleBuffer& first, SampleBuffer& second ) noexcept;
+	SampleBuffer& operator= ( SampleBuffer& that );
```
DomClark, line 115: If you make `m_varLock` mutable, then the copy constructor and assignment operator can take const references again.

src/core/SampleBuffer.cpp
```diff
@@ -131,6 +131,79 @@ SampleBuffer::SampleBuffer( const f_cnt_t _frames )
...
+	swap(first.m_endFrame, second.m_endFrame);
+	swap(first.m_loopStartFrame, second.m_loopStartFrame);
+	swap(first.m_loopEndFrame, second.m_loopEndFrame);
+	// float: std::swap
+	swap(first.m_amplification, second.m_amplification);
+	swap(first.m_frequency, second.m_frequency);
+	// bool: std::swap
+	swap(first.m_reversed, second.m_reversed);
+	// sample_rate_t: std::swap
+	swap(first.m_sampleRate, second.m_sampleRate);
```
DomClark, line 186: I'm not convinced of the usefulness of these comments. The point of writing `using std::swap;` is that the compiler determines the correct `swap` function for each type without needing to manually specify it. If you write `swap` with a comment along the lines of "this particular swap is `std::swap`", you might as well just write `std::swap` in the first place.

src/core/SampleBuffer.cpp
IanCaio: It took me a while to understand your point @DomClark , because somehow I was thinking that `swap(a,c)` running simultaneously with `swap(b,c)` could be an issue (which is not an unlikely situation if we have `bufferA = bufferC;` and `bufferB = bufferC;`). But in this case there's no deadlock, because even if the lock from `C` is disputed by both swaps, the other one is not, so we don't end up in a deadlock. The only situation where it would be an issue would be if we had `bufferA = bufferB;` and `bufferB = bufferA;` which indeed looks like bad code already if they are running concurrently.

It could be some sort of Poka Yoke to have the `tryLockForWrite` logic there though, but I see that it shouldn't be necessary as you said. You think we should keep the lock calls as they are and just make the lock recursive?

src/core/SampleBuffer.cpp
Spekular: Isn't `auto temp = that` recursive, since this occurs inside the definition of assignment?

src/core/SampleBuffer.cpp
Spekular: I'll remove them in a formatting commit, I added them for my own sake since I wanted to keep track of the types and didn't expect so many of the swaps to work with just `swap(a, b)`.

src/core/SampleBuffer.cpp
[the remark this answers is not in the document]
Spekular: From my check it doesn't protect anything else currently, but that doesn't necessarily mean it shouldn't. I think it's fine to leave this as is though, unless anyone disagrees?

src/core/SampleBuffer.cpp
Spekular: If swap is thread unsafe by convention then let's just leave this code as is. I just defaulted to trying to be defensive, because preventing hold and wait here would be simpler than making a resource allocation graph for all of LMMS.

Spekular: The address solution is pretty simple though, so unless it's broken or has negative consequences I'll just use that.

include/SampleBuffer.h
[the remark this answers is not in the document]
Spekular: I left it that way to match the surrounding code. Otherwise I would either inflate the diff with unrelated changes or have `name(args)` immediately below `name( args )`.

src/core/SampleBuffer.cpp
DomClark: `auto temp = that;` is an initialisation, not an assignment. Assignments put a new value into an existing object, but here a new object is being created.

Now I think about it though, a better option might be to take `that` by value. Then the copy is handled by the caller, and the existing copy constructor can take advantage of move semantics automatically as soon as a move constructor is defined.
```cpp
SampleBuffer& SampleBuffer::operator=(SampleBuffer that)
{
	swap(*this, that);
	return *this;
}
```

include/SampleTrack.h
```diff
@@ -48,8 +48,11 @@ class SampleTCO : public TrackContentObject
 	mapPropertyFromModel(bool,isRecord,setRecord,m_recordModel);
 public:
 	SampleTCO( Track * _track );
+	SampleTCO( const SampleTCO& orig );
```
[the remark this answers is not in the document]
Spekular: Matches surroundings.

include/SampleTrack.h
```diff
@@ -48,8 +48,11 @@ class SampleTCO : public TrackContentObject
...
 	virtual ~SampleTCO();
 
+	SampleTCO& operator= ( const SampleTCO& that ) = delete;
```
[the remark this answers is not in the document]
Spekular: Fixed in recent commit.

src/tracks/SampleTrack.cpp
```diff
@@ -108,6 +108,19 @@ SampleTCO::SampleTCO( Track * _track ) :
 
 
 
+SampleTCO::SampleTCO( const SampleTCO& orig ) :
+	SampleTCO( orig.getTrack() )
```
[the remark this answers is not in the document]
Spekular: Fixed in recent commit.

src/tracks/SampleTrack.cpp
```diff
@@ -108,6 +108,19 @@ SampleTCO::SampleTCO( Track * _track ) :
...
+{
+	// TODO: This creates a new SampleBuffer for the new TCO, eating up memory
+	// & eventually causing performance issues. Letting tracks share buffers
+	// when they're identical would fix this, but isn't possible right now.
+	m_sampleBuffer = new SampleBuffer( *orig.m_sampleBuffer) ;
```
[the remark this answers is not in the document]
Spekular: Fixed in a recent commit.

src/core/SampleBuffer.cpp
Spekular: Implemented in recent commit. Feel free to unresolve if you think I should revert this and just do them in argument order, @DomClark.

include/SampleBuffer.h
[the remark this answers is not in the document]
IanCaio: We always hit that dead end with code style. Wasn't the agreement that every line touched by the diff (either added or edited) should be compliant to the guidelines, independent of the surrounding code? Whatever the rule is, that's something we have to make clear, as in "included in the guidelines" clear :grimacing:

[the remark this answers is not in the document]
Spekular: It won't be an issue once the auto reformat is done, because at that point matching the surrounding code and matching the style guidelines won't conflict. Therefore I don't think it's something we need to put in the style guideline.

src/core/SampleBuffer.cpp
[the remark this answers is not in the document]
IanCaio: Sorry I couldn't look into it further, this week I haven't had much spare time. If it doesn't protect all the other members, then I think it's fine to leave it as is and change it when we change `SampleBuffer` to protect them (if we need to)

src/core/SampleBuffer.cpp
```diff
@@ -131,6 +131,79 @@ SampleBuffer::SampleBuffer( const f_cnt_t _frames )
...
+	m_loopStartFrame(orig.m_loopStartFrame),
+	m_loopEndFrame(orig.m_loopEndFrame),
+	m_amplification(orig.m_amplification),
+	m_reversed(orig.m_reversed),
+	m_frequency(orig.m_frequency),
+	m_sampleRate(orig.m_sampleRate)
+{
+	orig.m_varLock.lockForRead();
+	//Deep copy m_origData and m_data from original
+	memcpy(m_origData, orig.m_origData, m_origFrames* BYTES_PER_FRAME);
```
M374LX, line 151: The coding conventions require a space before and after infix operators.

src/core/SampleBuffer.cpp
```diff
@@ -131,6 +131,79 @@ SampleBuffer::SampleBuffer( const f_cnt_t _frames )
...
+	memcpy(m_data, orig.m_data, m_frames* BYTES_PER_FRAME);
```
M374LX, line 152: Same here.

M374LX (changes requested):

src/core/SampleBuffer.cpp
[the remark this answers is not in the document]
Spekular: Unresolving because Ian stated on Discord that
> the lock actually protects members like m_frames. It's just that they depend on m_data and are only changed when it is changed as well

Will fix this and the style issues M374LX brought up in the next commit.

src/core/SampleBuffer.cpp
```diff
@@ -131,6 +131,79 @@ SampleBuffer::SampleBuffer( const f_cnt_t _frames )
...
+	orig.m_varLock.unlock();
+}
```
Spekular, line 154:
```suggestion
SampleBuffer::SampleBuffer(const SampleBuffer& orig)
{
	orig.m_varLock.lockForRead();

	m_audioFile = orig.m_audioFile;
	m_origData = MM_ALLOC(sampleFrame, orig.m_origFrames);
	m_origFrames = orig.m_origFrames;
	m_data = MM_ALLOC(sampleFrame, orig.m_frames);
	m_frames = orig.m_frames;
	m_startFrame = orig.m_startFrame;
	m_endFrame = orig.m_endFrame;
	m_loopStartFrame = orig.m_loopStartFrame;
	m_loopEndFrame = orig.m_loopEndFrame;
	m_amplification = orig.m_amplification;
	m_reversed = orig.m_reversed;
	m_frequency = orig.m_frequency;
	m_sampleRate = orig.m_sampleRate;

	//Deep copy m_origData and m_data from original
	memcpy(m_origData, orig.m_origData, m_origFrames * BYTES_PER_FRAME);
	memcpy(m_data, orig.m_data, m_frames * BYTES_PER_FRAME);
	orig.m_varLock.unlock();
}
```

Spekular:
@DomClark @IanCaio any remaining objections?

src/core/SampleBuffer.cpp
```diff
@@ -131,6 +131,81 @@ SampleBuffer::SampleBuffer( const f_cnt_t _frames )
...
+	m_loopStartFrame = orig.m_loopStartFrame;
+	m_loopEndFrame = orig.m_loopEndFrame;
+	m_amplification = orig.m_amplification;
+	m_reversed = orig.m_reversed;
+	m_frequency = orig.m_frequency;
+	m_sampleRate = orig.m_sampleRate;
+
+	//Deep copy m_origData and m_data from original
+	memcpy(m_origData, orig.m_origData, m_origFrames * BYTES_PER_FRAME);
+	memcpy(m_data, orig.m_data, m_frames * BYTES_PER_FRAME);
```
DomClark, line 154: If `orig.m_origData` or `orig.m_data` can be null, such as for a default-constructed `SampleBuffer`. Calling `memcpy` with a null pointer is undefined behaviour, even if the size is zero.

src/tracks/SampleTrack.cpp
```diff
@@ -109,6 +109,19 @@ SampleTCO::SampleTCO( Track * _track ) :
...
 
+SampleTCO::SampleTCO(const SampleTCO& orig) :
+	SampleTCO(orig.getTrack())
+{
+	// TODO: This creates a new SampleBuffer for the new TCO, eating up memory
+	// & eventually causing performance issues. Letting tracks share buffers
+	// when they're identical would fix this, but isn't possible right now.
+	m_sampleBuffer = new SampleBuffer(*orig.m_sampleBuffer);
+	m_isPlaying = orig.m_isPlaying;
+}
```
DomClark, line 120: The constructor delegated to here initialises `m_sampleBuffer` with a newly allocated, default-constructed `SampleBuffer`. It is then leaked when this constructor overwrites it.

src/core/SampleBuffer.cpp
PhysSong: In addition to the comment above, MOST implementations of `memcpy` does nothing if the length is 0, but it's not guaranteed that ALL implementations do nothing.

src/tracks/SampleTrack.cpp
Spekular: Hopefully fixed in latest commit

src/core/SampleBuffer.cpp
Spekular: Hopefully fixed in latest commit

IanCaio:
`m_varLock` is still not recursive. I'm having second thoughts whether it should though. If we call the copy assignment from a method that has locked the sample buffer we would deadlock, but we should only be locking the sample buffer when accessing it (which means we shouldn't be copy assigning to it until we are done and unlock). So I think we are good having it non-recursive.

I personally think it's more easily understandable to use the `tryLock` approach on the deadlock prevention: the address approach works just as well, but for a reader it might be confusing at first to know why it's done that way. But it's no big deal and it can be kept that way.

I'll approve it, just want to review it with more time later.

Spekular:
> `m_varLock` is still not recursive.

Wow, I totally forgot about that. As you say though, it doesn't seem necessary.

> I personally think it's more easily understandable to use the `tryLock` approach on the deadlock prevention: the address approach works just as well, but for a reader it might be confusing at first to know why it's done that way.

Copying the std::swap() method would require:
- While both not locked
  - Try to lock first, then second. Release any acquired locks if we fail.
  - Try to lock second, then first. Release any acquired locks if we fail.

With our bracket standards this would be a lot of lines.

> But it's no big deal and it can be kept that way.

👍

src/core/SampleBuffer.cpp
```diff
@@ -130,6 +130,86 @@ SampleBuffer::SampleBuffer(const f_cnt_t frames)
 
 
 
+SampleBuffer::SampleBuffer(const SampleBuffer& orig)
+{
+	orig.m_varLock.lockForRead();
+
+	m_audioFile = orig.m_audioFile;
+	m_origData = MM_ALLOC(sampleFrame, orig.m_origFrames);
```
IanCaio, line 138: `malloc` return value for a size 0 allocation is implementation defined, meaning that it could be `nullptr` or a bogus address that can be freed. Do we care that the pointer might not be `nullptr` even though the original is?

It’s a small change anyways if we want them both to be null in that case (just a ternary operator). The same applies to `m_data`.

IanCaio: Just a note, that was the last detail I found while reviewing. The rest of the code LGTM. This is actually not even critical: both `SampleBuffer`s will have an empty `m_origData`. The only thing is that one might have a `m_origData == nullptr` and another one might have a `m_origData == 0xDEADBEEF` or whatever value is used to represent that empty buffer.

I still think it's worth fixing because it's simple to fix it and that could cause hard to track bugs if we compare that pointer with `nullptr` or compare the members two `SampleBuffer`s to check if they are equal. Something like `m_origData = orig.m_origData ? MM_ALOC(sampleFrame, orig.m_origFrames) : nullptr;` should do it.

Spekular: Ought to be fixed in latest commit.

Spekular:
@DomClark and @IanCaio, are you happy with the fixes in the recent commits? If so, merge?

IanCaio (approved):
LGTM! :rocket:

PhysSong (approved):

IanCaio:
Any objections to merging this?

Veratil:
> 
> 
> Any objections to merging this?

I'm going to ignore style review this time and say no objections. 👍

src/tracks/SampleTrack.cpp
```diff
@@ -109,6 +109,20 @@ SampleTCO::SampleTCO( Track * _track ) :
...
 
 
+SampleTCO::SampleTCO(const SampleTCO& orig) :
+	SampleTCO(orig.getTrack())
+{
+	// TODO: This creates a new SampleBuffer for the new TCO, eating up memory
+	// & eventually causing performance issues. Letting tracks share buffers
+	// when they're identical would fix this, but isn't possible right now.
+	delete m_sampleBuffer;
+	m_sampleBuffer = new SampleBuffer(*orig.m_sampleBuffer);
```
DomClark, line 119:
```suggestion
	*m_sampleBuffer = *orig.m_sampleBuffer;
```
This saves an unnecessary deallocation and allocation. Not too important though - this bit of code isn't performance-critical, and will probably change when `SampleBuffer`s can be shared.

DomClark (approved):
LGTM now. I made one comment, but it can be ignored if you want.

src/tracks/SampleTrack.cpp
Spekular: Wait... doesn't this go back to sharing the sample buffers? Or is this some weird syntax for delete, copy, assign?

Spekular: Never mind, dereference both buffers => assignment from buffer to buffer => copy assignment called.

Spekular:
Master CI fails in the same way as this, so I'll merge now 🎉

Issue #5728, opened 2020-10-23
Creating a new virtual midi port crashes Lmms


TheKnarf:
### Bug Summary

Lmms crashes when I create a new virtual midi port on MacOS Catalina.

#### Steps to reproduce

Using the [following node library](https://github.com/justinlatimer/node-midi) I run this example script:

```
const midi = require('midi');

// Set up a new input.
const input = new midi.Input();

// Configure a callback.
input.on('message', (deltaTime, message) => {
    console.log(`m: ${message} d: ${deltaTime}`);
});

// Create a virtual input port.
input.openVirtualPort("Test Input");

// A midi device "Test Input" is now available for other
// software to send messages to.

// ... receive MIDI messages ...

// Close the port when done.
input.closePort();
```

When running this in node (`node src/index.js`) Lmms crashes.

#### Expected behavior

I expect Lmms not to crash just because it detects a new virtual Midi device.

#### Affected LMMS versions

12.2

#### Logs
<details>
  <summary>Click to expand</summary>
<pre>

</pre>
</details>
How do I find crash logs from Lmms on MacOS?

PR #5729, opened 2020-10-23
When deviceName or endPointName are null strlen crashed

TheKnarf:
This fixes issue #5728.

The problem was that the deviceName was undefined when I tried to create a virtual midi device. That made `getName` return nil, which in turn crashed lmms when it tried to `strlen(nil)`.

### DIFF
--- a/src/core/midi/MidiApple.cpp
+++ b/src/core/midi/MidiApple.cpp
@@ -401,9 +401,9 @@ void MidiApple::midiInClose( MIDIEndpointRef reference )
 char *getName( MIDIObjectRef &object )
 {
 	// Returns the name of a given MIDIObjectRef as char *
-	CFStringRef name = nil;
+	CFStringRef name = nullptr;
 	if (noErr != MIDIObjectGetStringProperty(object, kMIDIPropertyName, &name))
-		return nil;
+		return nullptr;
 	int len = CFStringGetLength(name)+1;
 	char *value = (char *) malloc(len);
 	
@@ -615,8 +615,12 @@ char * MidiApple::getFullName(MIDIEndpointRef &endpoint_ref)
 	char * deviceName = getName(device);
 	char * endPointName = getName(endpoint_ref);
 	qDebug("device name='%s' endpoint name='%s'",deviceName,endPointName);
-	char * fullName = (char *)malloc(strlen(deviceName) + strlen(":") + strlen(endPointName)+1);
+	size_t deviceNameLen = deviceName == nullptr ? 0 : strlen(deviceName);
+	size_t endPointNameLen = endPointName == nullptr ? 0 : strlen(endPointName);
+	char * fullName = (char *)malloc(deviceNameLen + endPointNameLen + 2);
 	sprintf(fullName, "%s:%s", deviceName,endPointName);
+	if (deviceName != nullptr) { free(deviceName); }
+	if (endPointName != nullptr) { free(endPointName); }
 	return fullName;
 }
 


TheKnarf:
Seems that circleci failed on the macos build because there's something wrong with the homebrew install? That's not really related to my PR at all. Maybe a maintainer can try triggering a rerun of the build?

src/core/midi/MidiApple.cpp
```diff
@@ -615,7 +615,11 @@ char * MidiApple::getFullName(MIDIEndpointRef &endpoint_ref)
 	char * deviceName = getName(device);
 	char * endPointName = getName(endpoint_ref);
 	qDebug("device name='%s' endpoint name='%s'",deviceName,endPointName);
-	char * fullName = (char *)malloc(strlen(deviceName) + strlen(":") + strlen(endPointName)+1);
+	size_t deviceName_len = 0;
+	if(deviceName != nil) deviceName_len = strlen(deviceName);
+	size_t endPointName_len = 0;
+	if(endPointName != nil) endPointName_len = strlen(endPointName);
+	char * fullName = (char *)malloc(deviceName_len + endPointName_len + 2);
```
IanCaio, line 622: I'm not very familiar with this part of the code, but it looks good to me except for some code style issues (camelCase naming and pointer declaration). I think it could be shortened using ternary expressions too:

```
size_t deviceNameLen = deviceName == nullptr ? 0 : strlen(deviceName);
size_t endPointNameLen = endPointName == nullptr ? 0 : strlen(endPointName);
char* fullName = (char *)malloc(deviceNameLen + endPointNameLen + 2)
```

I also changed `nil` to `nullptr` because I think it's what should be used there. On that last line I'll wait on someone else's opinion whether obscuring the `strlen(":")` is fine, but bottom line it does the same thing so maybe just adding a comment there explaining the `+2` would be fine.

IanCaio (changes requested):

src/core/midi/MidiApple.cpp
TheKnarf: getName uses `nil` and not `nullptr`. Should that be updated as well?

https://github.com/LMMS/lmms/blob/f06f46fc2fec771607f5ebe282d19270c3b60f45/src/core/midi/MidiApple.cpp#L401-L413

IanCaio: Sure! Some of the older code might have code style issues or little issues, feel free to update it as well :)
If you end up fixing too much code that is unrelated to the main issue of the PR, then maybe it's better to submit those as a new refactoring PR for easier review though

Veratil: So, fun facts, at least with gcc both the if and tertiary format compile to near the same assembly. The `if` version is at 11 lines, and the `?:` version is 12.

IanCaio (approved):
LGTM!

src/core/midi/MidiApple.cpp
TheKnarf: I'm surprised that there's any difference at all.

IanCaio: A bit ironic the shorter code compiles to a slightly longer ASM :laughing:

Veratil:
I think we have a memory leak (not directly related to this PR's changes). In `getName` we use `malloc` to allocate space for both `deviceName` and `endPointName`, but then combine them into `fullname`, leaving both allocations behind without freeing.

Can you add after the `sprintf` and before the `return`:
```cpp
	sprintf(fullName, "%s:%s", deviceName,endPointName);
	if (deviceName != nullptr) { free(deviceName); }
	if (endPointName != nullptr) { free(endPointName); }
	return fullName;
```

TheKnarf:
@Veratil the code seems to leak everyplace getName is used in the file...

Veratil:
> 
> 
> @Veratil the code seems to leak everyplace getName is used in the file...

Let's fix them all then while you're there! 😁

TheKnarf:
I'm not about to do that in this PR.

Veratil:
> 
> 
> I'm not about to do that in this PR. Perhaps file an issue for later?

There are only 4 other locations in that file that use `getName`. Why not?

TheKnarf:
@Veratil It's unrelated to the bug I'm trying to fix. It would probably be better to refactor getName to return a std::string or QString. And I was hoping to go to bed.

Veratil:
Fair enough.

PR #5731, opened 2020-10-25
Add new Stereo Control effect

DouglasDGI:
Bada bing bada boom, LMMS's poorly-named utility effect is here.

Here's a basic demo: https://www.youtube.com/watch?v=sKlnuNIGd3k

This'll function as LMMS's equivalent to Ableton's Utility.  However, it does have a strong focus on stereo processing in general, hence the name.  Here's a run-down of its features:
1. Some features that will rarely be used but probably have some use somewhere maybe.  Mainly only added because Utility had it, so I assumed (perhaps incorrectly) that they had a good reason for it.  This includes, the Mute button, the Mono button, the single-channel Invert buttons, and the channel soloing dropdown box.
2. A DC Offset Remover, so that LMMS can have a non-LADSPA native one.
3. A Gain knob, powered by love, positivity, and multiplication.
4. The stereo enhancement section!  There are two types of stereo enhancers here.
i. The bottom one, simply labelled as "Width", is just a volume knob for the side bands.  This is the purest form of stereo enhancement, which is 100% mono-safe and applies no other effects to the audio.  The downside is that if the dry signal has no stereo information, this knob won't do anything.
ii. The top one, labelled "Stereoizer", is a special stereo enhancer which works on mono signals as well.  This is basically a clone (possibly even slightly better) of the famous Polyverse Wider VST, which is famous for making your audio wide while still being "completely mono-compatible" (their VST isn't actually completely mono-safe but they claim it is, this one is the same way).  With mono source material, collapsing the results of this stereo enhancement will result in the original signal without any modifications whatsoever.  This makes it a very healthy stereo enhancer to use, in comparison to some other methods (e.g. Haas).  It also comes with a highpass and lowpass, which filter the delayed signal so you can control which frequencies are widened.  I wrote a blog post about this here: https://r94231.wixsite.com/lostrobot/post/recreating-polyverse-s-wider-making-a-fully-mono-compatible-stereo-enhancer-for-any-audio
5. The bass mono section.  This simply removes the side bands below the set frequency, so you can make the bass frequencies of your audio mono.  This also comes with an Audition button, so you can hear which frequencies are being impacted.
6. And finally, the panning section.  Regular panning is lame, it just changes the volumes of each channel.  This, however, comes with two new types of panning.
i. The Dual mode pans each channel individually, which means you can choose which channel has which content.  Panning both all the way to the left will play both channels in the left channel, while with regular panning you'd only hear from the original left channel.
ii. The Binaural mode is a special Haas-based panning mode that I designed myself.  It applies a fractional delay to the audio being panned away from.  The delay increases as you pan further away.  This makes the panning sound impressively realistic, while having a very minimal impact on the sound, unlike HRTF which oftentimes filters the sound to the point that you wouldn't even want to use it anymore.  This mode also comes with a Delay knob, which lets you have some extra control over that fractional delay length, and a Spectral knob, which vaguely estimates the spectral response of the sound as it's panned around your head (usually best when not turned up all the way).

Try taking a song that has an easily-heard main instrument, like this one: https://soundcloud.com/quok/moonchildren
With the regular panning, the audio will sound panned, but the guitar still sounds like it's coming from in front of you.
However, with binaural panning... try it out.  It really will sound like the guitar is being played in a real 3D space from a different direction.  When it's not being used for mixing and when mono-compatibility isn't a concern, binaural panning is simply superior.  It's very cool.

All in all, it's important for basic functions to be covered in a DAW natively, so I went for that but added in some cool new goodies as well.  This functions as a full replacement for many VSTs and effects, like Ableton's Utility, Polyverse's Wider, Auburn's Psypan, Alex Hilton's A1StereoControl, and many more.

Remember to stay hydrated!


plugins/StereoControl/StereoControl.cpp
```diff
@@ -0,0 +1,443 @@
...
+	m_filtX[filtNum][channel][1] = m_filtX[filtNum][channel][0];
+	m_filtX[filtNum][channel][0] = inSamp;
+	m_filtY[filtNum][channel][1] = m_filtY[filtNum][channel][0];
+	m_filtY[filtNum][channel][0] = filterOutput;
+
+	return filterOutput;
+}
+
+
+void StereoControlEffect::multimodeFilter(sample_t in, float g, sample_t &lp, sample_t &hp, sample_t &filtBuf)
```
DomClark, line 417: Rather than passing different member variables of `StereoControlEffect` into here as references, why not make them members of a new `MultimodeFilter` class? Also, you can return multiple values using something like `std::pair<sample_t, sample_t>` or `std::array<sample_t, 2>`, meaning you don't have to use out parameters. And `multimodeCoeff` is only and always used as a parameter to this function, so could be computed inside of it.

Code like this:
```cpp
float lp;
float hp;
float hp2;
multimodeFilter(allpassValue, multimodeCoeff(stereoizerHP), lp, hp, m_stereoizerFilter[0][i]);
multimodeFilter(hp, multimodeCoeff(stereoizerLP), lp, hp2, m_stereoizerFilter[1][i]);
s[i] += lp * qMin(stereoizer, 1.f);
```
could look like this:
```cpp
const auto stereoHigh = m_stereoizerFilter[0][i].update(allpassValue, stereoizerHP);
const auto stereoLow = m_stereoizerFilter[1][i].update(stereoHigh[1], stereoizerLP);
s[i] += stereoLow[0] * qMin(stereoizer, 1.f);
```

This takes fewer lines since the output parameters don't need declaring, is clearer as to what is input and what is output, doesn't require a separate call to pre-process the frequency, and better associates the member variables with the code that uses them.

plugins/StereoControl/StereoControl.h
```diff
@@ -0,0 +1,77 @@
...
+	float m_sampleRate = 0;
+
+	float m_dcInfo[2] = {0};
+
+	float m_monoBassFilter[2][2] = {{0}};
+	float m_stereoizerFilter[2][2] = {{0}};
+
+	float m_haasSpectralPanFilter[2] = {0};
+
+	friend class StereoControlControls;
```
DomClark, line 73: I think this friend can be removed - `StereoControlControls` doesn't appear to use any private members of `StereoControlEffect`.

DomClark (commented):
As usual, I don't know a lot of DSP or music production, so I'm assuming that bit of code is correct and useful and just reviewing it as C++. Your code quality is noticably improving - I haven't really got much to say here.

plugins/StereoControl/StereoControl.cpp
```diff
@@ -0,0 +1,443 @@
...
+{
+	lp = in * g + filtBuf;
+	hp = in - lp;
+	filtBuf = hp * g * 2.f + filtBuf;
+}
+
+float StereoControlEffect::multimodeCoeff(float freq)
+{
+	const float g = tan(F_PI * freq / m_sampleRate);
+	return g / (1.f + g);
```
PhysSong, line 427: FYI, you can simplify it as `1.f / (1.f + cot(F_PI * freq / m_sampleRate))`.

plugins/StereoControl/StereoControlControlDialog.h
```diff
@@ -0,0 +1,80 @@
...
+
+
+class StereoControlControlDialog : public EffectControlDialog
+{
+	Q_OBJECT
+public:
+	StereoControlControlDialog(StereoControlControls* controls);
+	~StereoControlControlDialog() override
+	{
+	}
```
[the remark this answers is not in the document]
PhysSong: Or if you want to keep this:
```suggestion
	~StereoControlControlDialog() override = default;
```

plugins/StereoControl/StereoControlControls.cpp
```diff
@@ -0,0 +1,118 @@
...
+	m_stereoizerLPModel.setScaleLogarithmic(true);
+	m_stereoizerHPModel.setScaleLogarithmic(true);
+
+	m_soloChannelModel.addItem(tr("Stereo"));
+	m_soloChannelModel.addItem(tr("Left"));
+	m_soloChannelModel.addItem(tr("Right"));
+}
+
+
+void StereoControlControls::saveSettings(QDomDocument& doc, QDomElement& _this)
```
PhysSong, line 68: I suggest renaming `_this` because it doesn't fit the coding convention. Of course, you can't use `this` instead of `this`.
Something like `elem` or `element` will be better, I think.

plugins/StereoControl/StereoControlControls.cpp
```diff
@@ -0,0 +1,118 @@
...
+	m_stereoizerHPModel.saveSettings(doc, _this, "stereoizerHP");
+	m_panSpectralModel.saveSettings(doc, _this, "panSpectral");
+	m_panDelayModel.saveSettings(doc, _this, "panDelay");
+	m_panDualLModel.saveSettings(doc, _this, "panDualL");
+	m_panDualRModel.saveSettings(doc, _this, "panDualR");
+}
+
+
+
+void StereoControlControls::loadSettings(const QDomElement& _this)
```
PhysSong, line 94: The same.

plugins/StereoControl/StereoControl.cpp
```diff
@@ -0,0 +1,443 @@
...
+
+	for(fpp_t f = 0; f < frames; ++f)
+	{
+		const float gain = dbfsToAmp(gainBuf ? gainBuf->value(f) : m_stereocontrolControls.m_gainModel.value());
+		const float stereoizer = (stereoizerBuf ? stereoizerBuf->value(f) : m_stereocontrolControls.m_stereoizerModel.value()) * 0.01f;
+		const float width = (widthBuf ? widthBuf->value(f) : m_stereocontrolControls.m_widthModel.value()) * 0.01f;
+		const int panMode = panModeBuf ? panModeBuf->value(f) : m_stereocontrolControls.m_panModeModel.value();
+		const float pan = (panBuf ? panBuf->value(f) : m_stereocontrolControls.m_panModel.value()) * 0.01f;
+		const bool mono = monoBuf ? monoBuf->value(f) : m_stereocontrolControls.m_monoModel.value();
+		const bool dc = dcBuf ? dcBuf->value(f) : m_stereocontrolControls.m_dcModel.value();
```
PhysSong, line 117: It's hard to guess what `dc` means. Renaming to something like `dcRemoval` will make the meaning clear.
This also applies to related variables, too.

plugins/StereoControl/StereoControl.cpp
```diff
@@ -0,0 +1,443 @@
...
+		}
+
+		if (dc)
+		{
+			for (int i = 0; i < 2; ++i)
+			{
+				m_dcInfo[i] = m_dcInfo[i] * 0.999f + s[i] * 0.001f;
+				s[i] -= m_dcInfo[i];
+			}
+		}
```
PhysSong, line 379: It seems like you're using a one-pole LP filter for DC removal. However, this won't make the output signal consistent for various sample rates.

DouglasDGI:
Sorry for the mess, I guess that's what happens when you start a plugin and forget about it for a year.
Everything brought up has been addressed and/or fixed (except for one which may take a bit more time).

plugins/StereoControl/StereoControl.h
```diff
@@ -0,0 +1,78 @@
...
+	float m_dcCoeff;
+	float m_dcRemovalInfo[2] = {0};
+
+	float m_monoBassFilter[2][2] = {{0}};
+	float m_stereoizerFilter[2][2] = {{0}};
+
+	float m_haasSpectralPanFilter[2] = {0};
+};
+
+#endif
```
PhysSong, line 78: A trailing whitespace here.
```suggestion
#endif
```

plugins/StereoControl/StereoControlControlDialog.h
```diff
@@ -0,0 +1,76 @@
...
+	LedCheckBox * m_monoBassButton;
+	PixmapButton * m_auditionButton;
+	PixmapButton * m_invertLButton;
+	PixmapButton * m_invertRButton;
+	ComboBox * m_soloChannelBox;
+
+private slots:
+	void updateKnobVisibility();
+
+} ;
```
PhysSong, line 74: Unnecessary whitespace between `}` and `;`

plugins/StereoControl/StereoControl.cpp
```diff
@@ -0,0 +1,444 @@
...
+				float haasDelayVal[2] = {0};
+
+				if (pan >= 0)
+				{
+					haasDelayVal[0] = linearInterpolate(0, 0.0008f * m_sampleRate * panDelay, abs(pan));
+				}
+				else
+				{
+					haasDelayVal[1] = linearInterpolate(0, 0.0008f * m_sampleRate * panDelay, abs(pan));
+				}
```
PhysSong, line 289: - Why did you use linear interpolation instead of direct multiplication?
- I guess 0.8ms(`0.0008f` here) is approximately the distance between two ears divided by the speed of sound wave. Theoretically, the time difference should equal that value multiplied by `sin(angle)` where `angle` is the deviation of the source to left/right. This means `arcsin(panDelay * abs(pan))` will work like the deviation angle, right?

plugins/StereoControl/StereoControl.cpp
```diff
@@ -0,0 +1,444 @@
...
+					{
+						m_haasIndex[i] -= m_haasBufSize;
+					}
+					// Send new value to delay line
+					m_haasBuf[i][m_haasIndex[i]] = s[i];
+				}
+
+				if (pan != 0)
+				{
+					const float minGain = linearInterpolate(1.f, 0.15f, abs(pan));
```
PhysSong, line 321: Is `0.15f` arbitrary, or do you have some sources of this? Also, is it worth making the value configurable?

plugins/StereoControl/StereoControl.cpp
```diff
@@ -0,0 +1,444 @@
...
+					}
+				}
+
+				const int lowChnl = (pan <= 0);
+				const int highChnl = !lowChnl;
+
+				float lp;
+				float hp;
+
+				// Filter parameters are semi-arbitrary, but are vaguely based on some random graphs I saw during research
```
PhysSong, line 340: If you remember the sources, please mention them. Also, how did you get the gain formula?

plugins/StereoControl/StereoControl.cpp
PhysSong: Forgot to mention this: please consider extracting `0.0008f` to a named variable to make the code more readable.

PhysSong:
> except for one which may take a bit more time

Do you mean https://github.com/LMMS/lmms/pull/5731#discussion_r515677659? I can explain what you should do in detail, if you want.

plugins/StereoControl/StereoControl.cpp
DouglasDGI: It's mostly arbitrary, but it's also the same value Auburn's Psypan plugin uses.  Probably not worth making the value configurable, given the entire binaural illusion would be lost if the audio is panned too far.

plugins/StereoControl/StereoControl.cpp
DouglasDGI: "Why did you use linear interpolation instead of direct multiplication?"
Because sometimes I program past midnight when I really should be sleeping.
"[Other stuff]"
Oh wow.  I'll trust you then.  The value I chose was mostly chosen semi-arbitrarily (which probably means I mercilessly stole it from another plugin, most likely Auburn's Psypan which this was heavily inspired by).

plugins/StereoControl/StereoControl.cpp
DouglasDGI: Took me a while to scrape up the sites I read through while making this, but I think this part was mostly inspired by the graphs on this page: https://goodhertz.co/panpot/

PhysSong:
I checked the behavior of two plugins you mentioned(Psypan and Panpot). Here are some findings:
- Both plugins are using gain compensation when applying volume: while decreasing volume of one channel, they amplify the other channel slightly(less than 3dB)
- The time difference isn't not linear to the panning value, it's more like a short part of sine wave(approximately `0.0085 * sin(1.25 * pan)` for )
- I don't know the exact formula used by the spectral panning of Panpot, but it looks like the separation frequency is fixed around 1200Hz. Also, your `hpGain` for lower volume channel is way too low(whereas linear will give too high value). `hpGain` for high volume channel looks not bad, BTW.

Also, I'll explain some of my previous review comments:
- Making `0.15f` configurable: if a sound is coming from somewhere near from one of your ears, the level difference between two years will be large, whereas it will be small if the sound source is far from you. It slightly affects the spectral differences as well, but let's ignore this for now.

rdrpenguin04:

plugins/StereoControl/StereoControlControls.cpp
```diff
@@ -0,0 +1,118 @@
...
+	m_soloChannelModel.addItem(tr("Stereo"));
+	m_soloChannelModel.addItem(tr("Left"));
+	m_soloChannelModel.addItem(tr("Right"));
+}
+
+
+void StereoControlControls::saveSettings(QDomDocument& doc, QDomElement& elem)
+{
+	m_gainModel.saveSettings(doc, elem, "gain"); 
+	m_stereoizerModel.saveSettings(doc, elem, "stereoizer");
```
PhysSong, line 71: Trailing whitespaces:
```suggestion
	m_gainModel.saveSettings(doc, elem, "gain");
	m_stereoizerModel.saveSettings(doc, elem, "stereoizer");
```

plugins/StereoControl/StereoControl.cpp
```diff
@@ -0,0 +1,444 @@
...
+#include "plugin_export.h"
+
+extern "C"
+{
+
+Plugin::Descriptor PLUGIN_EXPORT stereocontrol_plugin_descriptor =
+{
+	STRINGIFY(PLUGIN_NAME),
+	"Stereo Control",
+	QT_TRANSLATE_NOOP("pluginBrowser", "General utility and stereo processing, including multiple types of panning and stereo enhancement."),
```
PhysSong, line 39: `pluginBrowser` -> `PluginBrowser`, the same for your plugins. Existing plugins are already fixed in `master`, FYI.

plugins/StereoControl/StereoControl.cpp
```diff
@@ -281,11 +281,11 @@ bool StereoControlEffect::processAudioBuffer(sampleFrame* buf, const fpp_t frame
 
 				if (pan >= 0)
 				{
-					haasDelayVal[0] = linearInterpolate(0, 0.0008f * m_sampleRate * panDelay, abs(pan));
+					haasDelayVal[0] = 0.0008f * m_sampleRate * panDelay * asin(pan);
```
PhysSong, line 284: The delay value should increase slower than linear, but it's faster than linear.

Issue #5595, opened 2020-07-24
Cloning an automation track does not copy pattern contents if pattern is connected to a control


messmerd:
### Summary

When I clone an automation track, the contents of a pattern in that automation track are not copied if that pattern is connected to a control, nor is the connection to the control copied. If a pattern is not connected to a control, however, the contents of that pattern will copy over successfully as expected. 

### My Thoughts

I am not sure if this is a bug or is intentional. 
 
If it is intentional, it seems like a poor design. If there is a concern about there being two patterns connected to the same control at the same time and sending two conflicting values to the control, this conflict could be resolved in other ways that don't involve deleting the pattern contents or removing the connection to the control. For example, you could only use the pattern in the last automation track and ignore any conflicting ones in automation tracks before it. This already seems to be the behavior if you copy an automation track's pattern into a new automation track without using the "Clone this track" feature. 

### Steps to Reproduce

- Create an automation track

- Create empty pattern in automation track and connect a control to it (shouldn't matter which one) 

- Create a waveform of your choice inside the pattern 

- Clone automation track by clicking the gear icon and clicking "Clone this track"

### Expected Behavior 

Cloned automation track contains pattern with the same waveform as the original automation track's pattern and is connected to the same control as the original pattern 

### Actual Behavior

Cloned automation track contains pattern with a range of 0 to 1 and is set at a constant value of 1. The pattern also lacks any connections to controls. 

For example: 
![lmms_automation_track_issue](https://user-images.githubusercontent.com/33463986/88350194-5cee7380-cd20-11ea-996c-acdbf05c5e80.PNG)

In the screenshot above, the first automation track is the original and the second is the clone. As you can see, the first pattern which had a control connected to it was not copied correctly and also lacks its connection to a control. However, the empty pattern and pattern without a control connected to it were copied successfully. 

### Affected LMMS versions

I am using LMMS 1.2.2 (64-bit) on Windows 10.

IanCaio:
[...]

PR #5732, opened 2020-10-25
Fixes bug with cloning Automation Tracks

IanCaio:
Fixes bug from issue #5595. When cloning an automation track, the IDs from the recently created AutomationPatterns weren't being resolved, causing them to show as disconnected automations.

This PR fixes the issue by adding a call to `AutomationPattern::resolveAllIDs()` on the `Track::clone()` method. It also fixes the code style on that method.

### DIFF
--- a/src/core/Track.cpp
+++ b/src/core/Track.cpp
@@ -2554,12 +2554,15 @@ Track * Track::create( const QDomElement & element, TrackContainer * tc )
 /*! \brief Clone a track from this track
  *
  */
-Track * Track::clone()
+Track* Track::clone()
 {
 	QDomDocument doc;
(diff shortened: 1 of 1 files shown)


superpaik:
It works ok. 
When track is cloned, values are copied ok. XML (mmp file) is correctly build (id and values)

IanCaio:
Awesome! Thanks once again for testing @superpaik !
The code review on this one should be an easy one, something short for a change :laughing:

russiankumar (approved):
Code LGTM to me.

Spekular (approved):

IanCaio:
If there are no objections I'll merge this one tomorrow. It's a very small fix and there are 2 approvals.

Issue #5703, opened 2020-10-05
Setting to "maximize" all sample/preset/instrument/project tabs


ttepatti:
## Enhancement Summary

I'd like to be able to "maximize" all of the tabs that appear on the left side of the screen. My Projects, My Samples, My Presets, etc. 

#### Justification

The icons only show up as icons until clicked, and I constantly find myself clicking through them to remember which is which. Having an option to have the titles always displayed would be useful.

#### Mockup

Current Design vs. Mockup

![image](https://user-images.githubusercontent.com/7002801/95135361-473ef680-0732-11eb-9826-55ee4e383172.png) ![image](https://user-images.githubusercontent.com/7002801/95135581-a56bd980-0732-11eb-8aff-e474dd1adbe1.png)

Thanks!! :smile:

yellowcrash10:
I gave this a try because it sounded like an easy fix for someone getting started with C++. Here is what I have so far.
![image](https://user-images.githubusercontent.com/1989498/97507152-9996c000-194a-11eb-9634-22f63d2e1432.png)
The labels are permanently visible, but they are closer to their neighboring button's icon than their own icon. I intend to fix this as well as add an effect when the mouse moves over them to hint that they are clickable.

As a side note, this change may prevent people who use LMMS on smaller screens or in a window from using the last item or two in the sidebar. Currently, a button will disappear if it can't be displayed completely. Some ideas I have for preventing this issue are renaming the tabs to remove "My", eliding the names if the window is too small, or making the sidebar scrollable. As an example of eliding, here is the Kate text editor with the window resized.
![image](https://user-images.githubusercontent.com/1989498/97510557-fb5b2800-1952-11eb-9df9-7e5a0d0dd8a1.png)

Spekular:
@yellowcrash10 awesome! Feel free to open a PR right away. You can mark it as a draft to indicate that it's not finished, and get feedback while you're working on it. That way we can catch any potential issues with the change sooner rather than later.

adam-eldredge:
I see this hasn't been moving for a while and wanted to see if I could give it a try. I made the expanded sidebars an option in the settings tab, and it updates upon restart. Is this still something that we want?

adam-eldredge:
Here is what the expanded sidebar looks like, as well as the new option in settings:

![expanded sidebar](https://user-images.githubusercontent.com/112440613/192375726-c82e018e-b2a5-416d-9eb5-c8c0bfd94537.png)
[...]

PR #5756, opened 2020-10-31
Sidebar Improvements

yellowcrash10:
This PR updates the sidebar to make it more modern and usable. The first change addresses #5703 by making the sidebar button labels visible permanently. This will be helpful for new users so that they will know what each button does immediately without needing to click on each button to discover what it does, and it will help longtime users so they won't have to remember what functions the icons represent.

The second change fixes an issue with the first change--the button labels are closer to the neighboring button's icon than their own button's icon. My solution to this was to move the position of the icon from the top of the buttons to the bottom.

As part of this PR, I have also taken the liberty of capitalizing the Root Directory button to match the other buttons.

Shown below is how this PR looks compared to the default.
image

Before merging this PR, there are two additional changes I would like to make.

- [ ] Add an effect to the buttons to hint they are clickable when the mouse moves over them
- [ ] Keep the buttons visible when the window is too small to show them all

Some ideas for preventing this issue are renaming the tabs to remove "My", eliding the names if the window is too small, or making the sidebar scrollable with a hint at the bottom indicating there are additional items. As an example of eliding, here is the Kate text editor with the window resized.

image

Any feedback is appreciated! 👍

### DIFF
--- a/src/gui/MainWindow.cpp
+++ b/src/gui/MainWindow.cpp
@@ -149,7 +149,7 @@ MainWindow::MainWindow() :
 
 
 	QStringList root_paths;
-	QString title = tr( "Root directory" );
+	QString title = tr( "Root Directory" );
 	bool dirs_as_items = false;
 
 #ifdef LMMS_BUILD_APPLE
--- a/src/gui/widgets/SideBar.cpp
+++ b/src/gui/widgets/SideBar.cpp
@@ -111,7 +111,7 @@ void SideBar::appendTab( SideBarWidget *widget )
 	SideBarButton *button = new SideBarButton( orientation(), this );
 	button->setText( " " + widget->title() );
 	button->setIcon( widget->icon() );
-	button->setLayoutDirection( Qt::RightToLeft );
+	button->setLayoutDirection( Qt::LeftToRight );
 	button->setToolButtonStyle( Qt::ToolButtonTextBesideIcon );
 	button->setCheckable( true );
 	m_widgets[button] = widget;
--- a/src/gui/widgets/SideBar.cpp
+++ b/src/gui/widgets/SideBar.cpp
@@ -112,6 +112,7 @@ void SideBar::appendTab( SideBarWidget *widget )
 	button->setText( " " + widget->title() );
 	button->setIcon( widget->icon() );
 	button->setLayoutDirection( Qt::RightToLeft );
+	button->setToolButtonStyle( Qt::ToolButtonTextBesideIcon );
 	button->setCheckable( true );
 	m_widgets[button] = widget;
 	m_btnGroup.addButton( button );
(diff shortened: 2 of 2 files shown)


zonkmachine:
>  renaming the tabs to remove "My",

The first thing that went through my mind when I saw those pictures side by side was that "My" should go.

Spekular:
Can this be made optional, as suggested in the original enhancement request? I would suggest a toggle at the bottom of the sidebar pane, or one in the settings menu. Otherwise this would be a regression for me and presumably others, who neither need nor want the full text filling up the sidebar at all times and spreading out the buttons.

+1 for removing the "My" prefix, it's not even accurate since factory files are displayed in several tabs.

superpaik:
Though it works well (waiting for fixing the problems caused by low resolution screens or when reducing the size of the windows) I agree with @Spekular that this should be an optional feature. As for the two solutions he propossed, I'd go for the setting menu, because I think it's something that you need or that you don't, as a user preference (similar to setting "Enable note labels on piano roll", I do not change it on regular basis)

+1 as well for removing the "My" prefix.

As for additional ideas to improve the sidebar, i'd go for an icon to the active project folder, where I normally store local resources of the project (samples, sample track and others). Now, you have to navigate a fair amount of folders when you want to add a new sample that you have in the project folder. Only a suggestion, maybe it's out of the scope of this PR.

Spekular:
I would consider adding new tabs to be firmly out of scope, but I agree that it's a good future enhancement.

yellowcrash10:
I will rename the labels and look into adding an option to the settings menu. Would it be reasonable to have it turned on by default? New users will most likely use the default settings and would benefit the most from having the labels turned on, and experienced users could simply turn them off.

superpaik:
> 
> 
> I will rename the labels and look into adding an option to the settings menu. Would it be reasonable to have it turned on by default? New users will most likely use the default settings and would benefit the most from having the labels turned on, and experienced users could simply turn them off.

Yes, I think it is ok to have it turned on by default. If the value is saved in the configuration file, when reinstall or upgrade version it would be set with the user preferences, wouldn't it?

mkljczk:
+1 for removing *My*, as there are also default/demo presets/samples/projects etc. so the word *my* means literally nothing and could be considered confusing (i know it's kinda standard)

DomClark:
Is this ready for review?

Spekular:
The option to enable/disable it is still missing. @yellowcrash10, you can reference PRs 5114 and 5787 to see how options are added.

yellowcrash10:
@DomClark Not yet. I haven't had time to work on this due to college, but I should be able to work on it later this week. I still want to add a mouseover effect to the buttons and fix the issue of the buttons disappearing at small window sizes.

Thank you for the tip, @Spekular. I'll refer to those when creating the option.

PR #5762, opened 2020-11-01
Apply minor CodeFactor code quality fixes

dan-giddins:
Minor code quality changes recommended and auto-generated by https://www.codefactor.io/

### DIFF
--- a/data/locale/ar.ts
+++ b/data/locale/ar.ts
@@ -4628,7 +4628,7 @@ Please visit http://lmms.sf.net/wiki for documentation on LMMS.</source>
         <translation type="unfinished"/>
     </message>
     <message>
-        <source>The Operators view contains all the operators. These include both audible operators (oscillators) and inaudible operators, or modulators: Low-frequency oscillators and Envelopes.
+        <source>The Operators view contains all the operators. These include both audible operators (oscillators) and inaudible operators, or modulators: Low-frequency oscillators and Envelopes. 
 
 Knobs and other widgets in the Operators view have their own what&apos;s this -texts, so you can get more specific help for them that way. </source>
         <translation type="unfinished"/>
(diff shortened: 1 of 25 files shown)


dan-giddins:
Could we add https://www.codefactor.io/ to LMMS as a build hook? It really is rather good

dan-giddins:
Tested locally, and everything seems fine. It might be worth someone with a bit more experience checking through things, but from my limited knowledge of XML and CSS, the changes codefactor has recommended seem semantically equivalent.

PhysSong:
Changes in `.css` files looks generally good. However, I think `.ts` files shouldn't be touched because that might break translations. If a source string contains trailing whitespaces, that should be fixed in the corresponding C++ code. @liushuyu Is this right?

IanCaio:
LGTM too (been a while I don't use CSS, so I didn't remember the three values option, thought it was either 1, 2 or 4 :smile:). As for what PhysSong said, if the source checks are sensible to trailing whitespaces the changes on those files could be an issue. Have to check that first, and if it breaks translations the only way to fix those is to manually fix all the source strings in the source code (could use some sed script to make it easier, still would be a bit workful).

PhysSong:
It seems like #4128 has removed relevant translation strings which existed in `plugins/monstro/Monstro.cpp`.
@liushuyu What should we do with those strings in `.ts` files?

liushuyu:
Sorry for the late reply!

> Changes in `.css` files looks generally good. However, I think `.ts` files shouldn't be touched because that might break translations. If a source string contains trailing whitespaces, that should be fixed in the corresponding C++ code. @liushuyu Is this right?

You are absolutely right! The translation strings should be verbatim versions of the strings in the source code files. Otherwise, the translation will be missing. To resolve this issue, corresponding C/C++ source files should be edited and then you can regenerate a set of new translation files.

> It seems like #4128 has removed relevant translation strings which existed in `plugins/monstro/Monstro.cpp`.
> @liushuyu What should we do with those strings in `.ts` files?

If they've been removed in the source code, they should be removed in the `.ts` files as well. But don't worry, they will not disappear completely, the translation memory in Transifex will still remember them.

dan-giddins:
I have reverted the commits regarding the translations

tresf:
@dan-giddins can you please assist us toggling this tool on for just the items we won't be covering with #4690?  For example, this PR makes CSS more consistent and we'd certainly like to have a CI check that handles this and the PR is small enough to be able to turn that on immediately.

Sorry for the delay.  I have some reservations about using shorthand syntax over readable syntax, but in the end, consistency wins.  I've un-archived the codefactor project.  Happy to add you as a maintainer if it's needed to get the settings right.  For now I've diabled `.ts` files, but I don't think that's enough.

PhysSong:
@tresf So What should we do with this PR?

tresf:
> @tresf So What should we do with this PR?

Closing until/unless the OP comes back to help with [our implementation](https://github.com/LMMS/lmms/pull/5762#issuecomment-740915817).

dan-giddins:
@tresf Sorry for the delay.

Yes I think I will need to be a maintainer to add it to the project. You could of course do it yourself if you want to limit the number of maintainers on the project. Either way, what we would need to do would be to use [GitHub checks](https://support.codefactor.io/i34-using-github-checks).

I've installed it for my personal profile:
image
You could try adding me as a maintainer, and it might just automatically work, but it might need to be the owner who installs it...

tresf:
Why does it need to administer our project?  That seems rather extreme to be granting to a 3rd party service.

tresf:
Invite sent.  Hopefully write access is enough (the codefactor portal suggests it is).

dan-giddins:
@tresf  is this what you are talking about?
image
If we want to use this we need to allow codefactor to have the following access:
image
I didn't want to press 'Approve & Request' incase this just adds it without anyone else approving. Can I do this? Does anyone know what pressing 'Approve & Request' does?
I think that Codefactor is now installed as a 'Check', but if we want annotations, we need to install the github app

tresf:
You can do it, I believe it's read-only and your permissions don't have access to the dangerous stuff.  Oddly, I've already done all of this using the service account (there was a bit of ambiguity as to WHO's access I was granting), so I think it's actually asking about this on your own profile.  Either way, go ahead and proceed.  What I don't like about read-only access to the administrative functions is it may be granting access to some private keys we use for webhooks, something I wouldn't want codefactor to have access to (and I'd expect them to share the same sentiments).

PR #5765, opened 2020-11-04
Allow SampleTCOs/Sample Clips to be reversed

russiankumar:
There was a reverse option in `SampleBuffer`, so I decided to implement it so that it can actually be used.

capture3

Progress tracker:
- [x] Have the samples actually reverse
- [x] Update the GUI
- [x] Save the reversed state
- [x] Pixmap for reversing (Fix: use `flip_x` pixmap found in automation editor)

Pretty minor feature, ~so I think it won't be much of a nuisance, considering it's practically complete at the time of writing.~ *I should not have said that.*


src/tracks/SampleTrack.cpp
```diff
@@ -310,11 +314,18 @@ void SampleTCO::loadSettings( const QDomElement & _this )
 		useCustomClipColor( true );
 		setColor( _this.attribute( "color" ) );
 	}
+
+	if( _this.hasAttribute( "reversed" ) )
+	{
+		m_sampleBuffer->setReversed( true );
+		emit wasReversed();		// tell SampleTCOView to update the view
+	}
```
russiankumar, line 322: I'd like to point out why I didn't just call `reverseSample()` over here. For one, it had bugs and was unable to consistently load the sample reversed, and `reverseSample()` also sets the modified status, which we don't want during loading ideally.

src/core/SampleBuffer.cpp
```diff
@@ -957,13 +957,27 @@ void SampleBuffer::visualize( QPainter & _p, const QRect & _dr,
...
+		for( int frame = last; frame > first; frame -= fpp )
+		{
+			l[n] = QPointF( xb + ( (frame - first) * double( w ) / nb_frames ),
+				( yb - ( m_data[frame][0] * y_space * m_amplification ) ) );
+			r[n] = QPointF( xb + ( (frame - first) * double( w ) / nb_frames ),
+				( yb - ( m_data[frame][1] * y_space * m_amplification ) ) );
+			++n;
+		}
+	}
+	else
```
IanCaio, line 971: Since the drawing statements are equal on both `for` loops, I wanted to know what you and others think of actually just manipulating the for statement, something like:

```
for (int frame = m_reversed ? last : first;
    m_reversed ? frame > first : frame < last;
    m_reversed ? frame -= fpp : frame += fpp)
```

I'm not sure if the cost on readability is really worth it for saving around 8 lines though, that is if the ternary operator can be used that way.

IanCaio: Something I'm also struggling to get used to, but according to our code style guideline it should be `for(statements)` and the expressions inside `QPointF` shouldn't have spaces after `(` and before `)`

src/tracks/SampleTrack.cpp
```diff
@@ -280,6 +280,10 @@ void SampleTCO::saveSettings( QDomDocument & _doc, QDomElement & _this )
 	{
 		_this.setAttribute( "color", color().name() );
 	}
+	if( m_sampleBuffer->reversed() )
+	{
+		_this.setAttribute( "reversed", "true" );
+	}
```
IanCaio, line 286: Same note about code style as above:
```suggestion
	if (m_sampleBuffer->reversed())
	{
		_this.setAttribute("reversed", "true");
	}
```

src/tracks/SampleTrack.cpp
IanCaio: I'd have to check it further, but if I'm not mistaken there's a variable that is set while the project is being loaded. Maybe the `reverseSample()` method could check for that variable and not call `Engine::getSong()->setModified();` if it's true. It sounds totally fine to do it the way you're doing here though, just something to maybe consider!

IanCaio (commented):
Overall it LGTM. I've wrote a suggestion regarding the drawing `for` loop, an idea to fix the issue with calling `reverseSample` while loading the project (which has to be studied first) and some code style convention notes. Didn't explicitly request the changes because they are mostly suggestions not requests and I'd like to hear others opinions on them.

src/tracks/SampleTrack.cpp
russiankumar: Yep, but the bug issue still does stand. I tried to work around it and see what caused it but I wasn't able to come to a conclusion, so I decided to just set the variable here.

In case you want to try and replicate: even though `reverseSample()` is called, it still unreverses it to its original way (or doesn't reverse it at all). This is at load time.

src/tracks/SampleTrack.cpp
russiankumar: Well I've left my opinion about these on your PR a while back xD

I could change these if you want, but I would rather keep them consistent for now (considering I used this style in the track colors PR as well). If you still want me to change these however, will do that.

src/core/SampleBuffer.cpp
russiankumar: I'll test if that works, and edit this comment with the result.

src/tracks/SampleTrack.cpp
IanCaio: Oh, I just noticed `reverseSample` is a TCO view method. So I think that's right, you'll have to call the SampleBuffer method `setReversed` instead because the TCO view might not even have been created yet I think. Just a guess, I haven't looked much further, but seems like the way it's done now is the right way!

src/tracks/SampleTrack.cpp
IanCaio: Sure, it was just a note. In the current state of the code base it's a little complicated enforcing the code style (unless it's something outrageous, which is not the case here :smile: ). I tend to write new code according to the code style though, to avoid adding to the amount of wrongly formatted code, but I'm not leaving this as a request here!

src/core/SampleBuffer.cpp
IanCaio: Awesome! Well, as I said I'm not sure what others think of this approach in terms of readability, but if they agree and you're willing to change the other parts of the code that have similar loops sure! It would save work from a possible refactor in the future :smiley:

russiankumar: Sure, I'll have to look into those snippets to see if that's possible, and if it is, I'll get on it tomorrow!

russiankumar: Sorry for the massive delay, but I've now changed the three snippets. I used @PhysSong's idea that he proposed on Discord for my snippet, and the other two were just assigning a variable based on a ternary operator.

russiankumar:
Hey, since I have changed some other code in `SampleBuffer.cpp` in the functions `convertIntToFloat` and `directFloatWrite` because it had reversing snippets which could be made more shorter and readable. However I'm not sure what these really do, so it would be great if someone could test and see if samples reliably work after this.

In general, if anybody wants to test, you can try these and check for stability:
- [ ] Reversing samples in general and seeing if the GUI waveform and the audio are in sync
- [ ] Reversing samples while they're being played, and checking for GUI-audio sync
- [ ] Copying around sample tracks which are reversed and seeing if they're still reversed
- [ ] Saving reversed samples and seeing if they're still reversed
- [ ] Any other sample-related stuff that you can think of

Also want to point out that I was preferably looking for either a new icon, however if that's not possible, I can probably use the "loop" icon.

src/core/SampleBuffer.cpp
```diff
@@ -957,7 +929,11 @@ void SampleBuffer::visualize( QPainter & _p, const QRect & _dr,
 	const int xb = _dr.x();
 	const int first = focus_on_range ? _from_frame : 0;
 	const int last = focus_on_range ? _to_frame : m_frames;
-	for( int frame = first; frame < last; frame += fpp )
+
+	auto start = m_reversed ? last : first;
+	auto offset = m_reversed ? first : last;
+	for ( int frame = start; m_reversed ? frame > offset : frame < offset;
+		  m_reversed ? frame -= fpp : frame += fpp )
```
IanCaio, line 936: Don't mean to nitpick, everything else looks good to me, but I guess `end` would convey the idea better than `offset`. It feels like in this `for` loop the offset is actually `fpp`. Also, just a suggestion, maybe break the `for` in 3 lines for even better readability?

```suggestion
	auto end = m_reversed ? first : last;
	for( int frame = start;
		m_reversed ? frame > end : frame < end;
		m_reversed ? frame -= fpp : frame += fpp )
```

src/core/SampleBuffer.cpp
```diff
@@ -306,27 +306,13 @@ void SampleBuffer::convertIntToFloat ( int_sample_t * & _ibuf, f_cnt_t _frames,
 
 	// if reversing is on, we also reverse when
 	// scaling
-	if( m_reversed )
+	int idx = m_reversed ? ( _frames - 1 ) * _channels : 0;
+	for( f_cnt_t frame = 0; frame < _frames;
+					++frame )
```
Veratil, line 311:
```suggestion
	for (f_cnt_t frame = 0; frame < _frames; ++frame)
```

IanCaio:
I also didn't get to read the whole methods you refactored (`convertIntToFloat` and `directFloatWrite`) but as far as the logic goes the changes looks good!

I'll try to test the build soon

src/core/SampleBuffer.cpp
```diff
@@ -306,27 +306,13 @@ void SampleBuffer::convertIntToFloat ( int_sample_t * & _ibuf, f_cnt_t _frames,
...
-		for( f_cnt_t frame = 0; frame < _frames;
-						++frame )
-		{
-			m_data[frame][0] = _ibuf[idx+0] * fac;
-			m_data[frame][1] = _ibuf[idx+ch] * fac;
-			idx += _channels;
-		}
+		m_data[frame][0] = _ibuf[idx+0] * fac;
+		m_data[frame][1] = _ibuf[idx+ch] * fac;
+		m_reversed ? idx -= _channels : idx += _channels;
```
Veratil, line 315: I'm not a fan of this style honestly. I don't think it's as easy to read as say:
```suggestion
		idx = m_revsered ? idx - _channels : idx + _channels;
```
or even
```suggestion
		idx += m_reversed ? -_channels : _channels;
```
Curious what everyone else thinks.

src/core/SampleBuffer.cpp
```diff
@@ -306,27 +306,13 @@ void SampleBuffer::convertIntToFloat ( int_sample_t * & _ibuf, f_cnt_t _frames,
 
 	// if reversing is on, we also reverse when
 	// scaling
-	if( m_reversed )
+	int idx = m_reversed ? ( _frames - 1 ) * _channels : 0;
```
Veratil, line 309: Just style update, can be ignored if wanted...
```suggestion
	int idx = m_reversed ? (_frames - 1) * _channels : 0;
```

src/core/SampleBuffer.cpp
```diff
@@ -341,27 +327,13 @@ void SampleBuffer::directFloatWrite ( sample_t * & _fbuf, f_cnt_t _frames, int _
...
```
Veratil, line 330: Just style...
```suggestion
	int idx = m_reversed ? (_frames - 1) * _channels : 0;
```

src/core/SampleBuffer.cpp
```diff
@@ -341,27 +327,13 @@ void SampleBuffer::directFloatWrite ( sample_t * & _fbuf, f_cnt_t _frames, int _
...
+	for( f_cnt_t frame = 0; frame < _frames;
+					++frame )
```
Veratil, line 332:
```suggestion
	for (f_cnt_t frame = 0; frame < _frames; ++frame)
```

src/core/SampleBuffer.cpp
```diff
@@ -341,27 +327,13 @@ void SampleBuffer::directFloatWrite ( sample_t * & _fbuf, f_cnt_t _frames, int _
...
-		for( f_cnt_t frame = 0; frame < _frames;
-						++frame )
-		{
-			m_data[frame][0] = _fbuf[idx+0];
-			m_data[frame][1] = _fbuf[idx+ch];
-			idx += _channels;
-		}
+		m_data[frame][0] = _fbuf[idx+0];
+		m_data[frame][1] = _fbuf[idx+ch];
+		m_reversed ? idx -= _channels : idx += _channels;
```
Veratil, line 336: Same suggestion as above.

src/core/SampleBuffer.cpp
Veratil: Honestly, I would want this to be completely changed. I don't like the expression within the ternary operator, it's not easy to read at a glance. I think setting up variables first and having a single comparison would be easier to read.

What would happen if *somehow* `m_reversed` was changed mid-loop?

src/tracks/SampleTrack.cpp
Veratil: I'd rather you start coding in the new format to get used to it now rather than later.

src/core/SampleBuffer.cpp
IanCaio: The variables are set before, but the comparisons are different on each case (`m_reversed` being true or false, will result in a _higher than_ or _lower than_ comparison) so not sure how to replace that middle statement. Maybe that last one (`m_reversed ? frame -= fpp : frame += fpp`) could be replaced if an `offset` variable was used, something like: `auto offset = m_reversed ? -fpp : fpp;` and having `frame += offset;`

As for `m_reversed` being changed mid-loop, I guess we would need to protect the methods with mutexes if they aren't already, I'd expect they were (because drawing the sample while it changes could be an issue) but I haven't looked to be sure. Very nice point bringing it up! Right now I think the consequences of changing the sample while it's being draw is just that it might be draw wrongly, but with `m_reversed` being used as a conditional on the `for` statement we can end up with out of bound accesses and infinite loops, much worse.

src/core/SampleBuffer.cpp
IanCaio: I think I prefer the first suggestion too: `idx = m_reversed ? idx - _channels : idx + _channels;`

src/core/SampleBuffer.cpp
Veratil: > As for m_reversed being changed mid-loop, I guess we would need to protect the methods with mutexes if they aren't already
If you don't compare during the loop we don't have to worry about that risk, and don't have to protect with mutex (as much) 😁 

> so not sure how to replace that middle statement
```cpp
auto startFrame = m_reversed ? last : first;
auto endFrame = m_reversed ? first : last;
auto nextFrame = m_reversed ? -1 : 1;
for (auto frame = startFrame; startFrame != lastFrame; frame += nextFrame)
{ ... }
```

IanCaio: True, that should work! I think we still need mutexes (to prevent `m_reversed` changing between the variables being set) but in terms of readability it does looks better!

russiankumar: > What would happen if _somehow_ `m_reversed` was changed mid-loop?

Well, according to testing, nothing! It just reverses and continues playing from the new positions.

Veratil seems to have provided a better alternative for this loop, so I'll go with it :)

src/core/SampleBuffer.cpp
russiankumar: Apart from being a tad bit longer, this isn't an issue. I'll change this.

src/tracks/SampleTrack.cpp
russiankumar: I already code like this in any project I do myself, so I'm already used to it. The only reason I'm doing this is for consistency, however since there's two of you who want me to use the new style, I'll go for it over here.

src/core/SampleBuffer.cpp
IanCaio: @russiankumar
Race conditions are hard to test, for the issue to be noticed you'd have to reverse the sample **exactly** when a thread was going through the drawing loop, so the issue might be there but the state that triggers it hasn't happened when you tested. If I'm not wrong, the race condition is still there even with the improvement that @Veratil suggested (if a thread changes `m_reversed` after setting `startFrame` and `endFrame` but before setting `nextFrame` for example). I'm not sure how multithreading works on LMMS, but I think there should be a mutex on every method that accesses (either reading or writing) members of an object. That's of course outside the scope of your PR, though having that `for` checking for `m_reversed` on each loop makes the issue more apparent (there are much more opportunities for the race condition to happen).

As for ways to not have the ternary on the `for` conditional, I can't think of any that wouldn't make it worse to read. The readability is not that bad IMHO. The biger problem is, as mentioned before, it kind of creates more room for the race condition to happen. I'm not sure what's the best approach here, should we add mutexes to the `SampleBuffer` class before merging this?

Veratil: From Dom in Discord:
`A solution for your particular problem would probably be to read m_reversed once and keep it in a local variable, so it keeps the same value throughout the function.`

IanCaio: @Veratil You've probably seen it on Discord already, but copying it here: Dom suggested accessing the member variable just once, storing it on a local variable and using it instead to reduce the space where the race condition can happen. I think that could be a good call for being able to merge this PR before fixing the multi-threading issues. I'd add a "TODO" comment on the lines that store the member variable to a local one explaining this shouldn't be needed when the mutexes are added.

Gabrielxd195:
I think the best icon for the '' reverse sample '' option is this.
Boton de Reverse sample

Which is to '' flip horizontally '' the automation tcos. By reversing an audio sample, you are technically "flipping it horizontally."

And if in the future they implemented the possibility of '' inverting the phase '' to an audio sample, then that would be '' flipping vertically '', and we would use the icon that is already in the automation tco. but I guess that would be discussed in another thread.

src/core/SampleBuffer.cpp
russiankumar: Alright, that seems like a good idea, I'll implement that. Still have no idea what to do about the segfault: do I leave the ternary in loop or should I see if I can figure out a way to extract it and have it work?

PhysSong: > I didn't look to see what `fpp` is

It means frames per pixel.

Here are some more alternatives:
```cpp
auto startFrame = m_reversed ? last : first;
auto count = (last-first) / fpp + 1;
auto step = m_reversed ? -fpp : fpp;
for (auto frame = startFrame, i = 0; i < count; frame += step, ++i)
{ /**/ }
for (size_t i = 0; i < count; ++i)
{ auto frame = startFrame + i * step; /**/ }
```

russiankumar: Thanks! I'll test this out and see if this works, and I'll commit this if it does.

src/core/SampleBuffer.cpp
```diff
@@ -296,72 +296,41 @@ void SampleBuffer::update( bool _keep_settings )
 }
 
 
-void SampleBuffer::convertIntToFloat ( int_sample_t * & _ibuf, f_cnt_t _frames, int _channels)
+void SampleBuffer::convertIntToFloat (int_sample_t * & _ibuf, f_cnt_t _frames, int _channels)
```
Veratil, line 299: Not necessary to change, but since we touched this we should remove the `_` prefix as per new coding conventions.
```suggestion
void SampleBuffer::convertIntToFloat(int_sample_t * & ibuf, f_cnt_t frames, int channels)
```

src/core/SampleBuffer.cpp
```diff
@@ -296,72 +296,41 @@ void SampleBuffer::update( bool _keep_settings )
...
 {
 	// following code transforms int-samples into
 	// float-samples and does amplifying & reversing
 	const float fac = 1 / OUTPUT_SAMPLE_MULTIPLIER;
-	m_data = MM_ALLOC( sampleFrame, _frames );
-	const int ch = ( _channels > 1 ) ? 1 : 0;
+	m_data = MM_ALLOC (sampleFrame, _frames);
```
Veratil, line 304:
```suggestion
	m_data = MM_ALLOC(sampleFrame, _frames);
```

src/core/SampleBuffer.cpp
```diff
@@ -296,72 +296,41 @@ void SampleBuffer::update( bool _keep_settings )
...
+	const int ch = (_channels > 1) ? 1 : 0;
 
 	// if reversing is on, we also reverse when
 	// scaling
-	if( m_reversed )
+	int idx = m_reversed ? (_frames - 1) * _channels : 0;
```
Veratil, line 309: I think we should do Dom's suggestion here.
```suggestion
	const bool is_reversed = m_reversed;
	int idx = is_reversed ? (_frames - 1) * _channels : 0;
```
and use `is_reversed` for the other, too.

src/core/SampleBuffer.cpp
```diff
@@ -296,72 +296,41 @@ void SampleBuffer::update( bool _keep_settings )
...
 }
 
-void SampleBuffer::directFloatWrite ( sample_t * & _fbuf, f_cnt_t _frames, int _channels)
-
+void SampleBuffer::directFloatWrite (sample_t * & _fbuf, f_cnt_t _frames, int _channels)
 {
 
-	m_data = MM_ALLOC( sampleFrame, _frames );
-	const int ch = ( _channels > 1 ) ? 1 : 0;
+	m_data = MM_ALLOC (sampleFrame, _frames);
```
Veratil, line 323:
```suggestion
	m_data = MM_ALLOC(sampleFrame, _frames);
```

src/core/SampleBuffer.cpp
```diff
@@ -296,72 +296,41 @@ void SampleBuffer::update( bool _keep_settings )
...
+		m_data[frame][1] = _ibuf[idx+ch] * fac;
+		idx += m_reversed ? -_channels : _channels;
 	}
 
 	delete[] _ibuf;
 }
 
-void SampleBuffer::directFloatWrite ( sample_t * & _fbuf, f_cnt_t _frames, int _channels)
-
+void SampleBuffer::directFloatWrite (sample_t * & _fbuf, f_cnt_t _frames, int _channels)
```
Veratil, line 320: Same as above.
```suggestion
void SampleBuffer::directFloatWrite(sample_t * & fbuf, f_cnt_t frames, int channels)
```

src/core/SampleBuffer.cpp
```diff
@@ -296,72 +296,41 @@ void SampleBuffer::update( bool _keep_settings )
...
 
-	m_data = MM_ALLOC( sampleFrame, _frames );
-	const int ch = ( _channels > 1 ) ? 1 : 0;
+	m_data = MM_ALLOC (sampleFrame, _frames);
+	const int ch = (_channels > 1) ? 1 : 0;
 
 	// if reversing is on, we also reverse when
 	// scaling
-	if( m_reversed )
+	int idx = m_reversed ? (_frames - 1) * _channels : 0;
```
Veratil, line 328: Same as above with `is_reversed`.

src/tracks/SampleTrack.cpp
```diff
@@ -280,6 +280,10 @@ void SampleTCO::saveSettings( QDomDocument & _doc, QDomElement & _this )
 	{
 		_this.setAttribute( "color", color().name() );
 	}
+	if( m_sampleBuffer->reversed() )
```
Veratil, line 283:
```suggestion
	if (m_sampleBuffer->reversed())
```

src/tracks/SampleTrack.cpp
```diff
@@ -280,6 +280,10 @@ void SampleTCO::saveSettings( QDomDocument & _doc, QDomElement & _this )
...
+	{
+		_this.setAttribute( "reversed", "true" );
```
Veratil, line 285:
```suggestion
		_this.setAttribute("reversed", "true");
```

src/tracks/SampleTrack.cpp
```diff
@@ -310,11 +314,18 @@ void SampleTCO::loadSettings( const QDomElement & _this )
 		useCustomClipColor( true );
 		setColor( _this.attribute( "color" ) );
 	}
+
+	if( _this.hasAttribute( "reversed" ) )
```
Veratil, line 318:
```suggestion
	if (_this.hasAttribute( "reversed" ))
```

src/tracks/SampleTrack.cpp
```diff
@@ -310,11 +314,18 @@ void SampleTCO::loadSettings( const QDomElement & _this )
...
+	{
+		m_sampleBuffer->setReversed( true );
```
Veratil, line 320:
```suggestion
		m_sampleBuffer->setReversed(true);
```

src/tracks/SampleTrack.cpp
```diff
@@ -335,6 +346,9 @@ SampleTCOView::SampleTCOView( SampleTCO * _tco, TrackView * _tv ) :
 	connect( m_tco, SIGNAL( sampleChanged() ),
 			this, SLOT( updateSample() ) );
 
+	connect( m_tco, SIGNAL( wasReversed() ),
+			this, SLOT( update() ) );
```
Veratil, line 350:
```suggestion
	connect(m_tco, SIGNAL(wasReversed()),
			this, SLOT(update()));
```

src/tracks/SampleTrack.cpp
```diff
@@ -408,6 +422,9 @@ void SampleTCOView::contextMenuEvent( QContextMenuEvent * _cme )
 				tr( "Set/clear record" ),
 						m_tco, SLOT( toggleRecord() ) );*/
 
+	contextMenu.addAction( embed::getIconPixmap( "flip_x" ),
+			tr( "Reverse sample" ), this, SLOT( reverseSample() ) );
```
Veratil, line 426:
```suggestion
	contextMenu.addAction(embed::getIconPixmap("flip_x"),
			tr("Reverse sample"), this, SLOT(reverseSample()));
```

src/tracks/SampleTrack.cpp
```diff
@@ -625,6 +642,16 @@ void SampleTCOView::paintEvent( QPaintEvent * pe )
 
 
 
+void SampleTCOView::reverseSample()
+{
+	m_tco->sampleBuffer()->setReversed( !m_tco->sampleBuffer()->reversed() );
```
Veratil, line 647:
```suggestion
	m_tco->sampleBuffer()->setReversed(!m_tco->sampleBuffer()->reversed());
```

src/core/SampleBuffer.cpp
russiankumar: Yup, missed this one, thanks.

src/core/SampleBuffer.cpp
russiankumar: There's a bunch of these underscored variables in `visualise()` too, do I get rid of those as well?

src/core/SampleBuffer.cpp
```diff
@@ -957,12 +925,18 @@ void SampleBuffer::visualize( QPainter & _p, const QRect & _dr,
...
-	{
-		l[n] = QPointF( xb + ( (frame - first) * double( w ) / nb_frames ),
-			( yb - ( m_data[frame][0] * y_space * m_amplification ) ) );
-		r[n] = QPointF( xb + ( (frame - first) * double( w ) / nb_frames ),
-			( yb - ( m_data[frame][1] * y_space * m_amplification ) ) );
+	
+	const bool isReversed = m_reversed;
+
+	auto startFrame = isReversed ? last : first;
+	auto count = (last - first) / fpp + 1;
```
IanCaio, line 932: I'm not sure if maybe this is a scenario that will never happen, but before the `for` loop ran starting on `first` while the current frame was below `last`. So if first was equal to last, the loop would never run.

With the current logic, if last is equal to first, count will be 1, and with `i` beginning at 0 it will run the loop once. I think I understood the `+1` on `count` (so if `last - first` is smaller than `fpp` it will still draw the frames), but is there a chance that `last` will be equal to `first`?

PhysSong: Oh, I thought it was `[first, last]`. If we need `[first, last)` or `(first, last]`, it should be `(last - first + fpp - 1) / fpp`.

IanCaio: I don't know, to be honest the previous loop could be the wrong one, I just compared both. If `first` is equal to `last` do we have a buffer with length of 0 or 1? From the looks of it it seems that last is non-inclusive, since when `focus_on_range == false` it can be set to `m_frames`, which I think is the total number of frames. So `last` would only be equal to `first` if there are no frames. We just need to make sure and use your suggestion if it's non-inclusive,

src/core/SampleBuffer.cpp
IanCaio: I think it's up to you, it can be a bit of a rabbit hole. It's always welcome that those code style issues are fixed, but since it's outside the scope of the PR it's alright if you want to leave it for another one as far as getting this one approved goes! :smile:

Veratil: If you touched the line, that's where I suggested it.

src/core/SampleBuffer.cpp
IanCaio: From what I've been reading, `m_frames` is the length of the sample in frames. Even when a sample fails to load, a 1-frame sample is created (`m_frames = 1`). If `first` starts at 0 and `last` is equal to the total number of frames then `last` is non-inclusive. Apparently `last` being equal to `first` would mean we have a sample buffer of length 0, and this doesn't seem to happen.

The weird part is that when `focus_on_range` is true, `last` doesn't seem to be a length anymore, but instead a frame index (`_to_frame`). If we are using `last` as an index, then maybe we should have `const int last = focus_on_range ? _to_frame : m_frames - 1;` right?

I'm still reading some of the code to make sure, if someone could help me check this out. Sorry if I'm making a bit of a confusion here, I just want to make sure we have a proper loop.

src/core/SampleBuffer.cpp
```diff
@@ -957,12 +925,18 @@ void SampleBuffer::visualize( QPainter & _p, const QRect & _dr,
...
+	auto step = isReversed ? -fpp : fpp;
+	for (int frame = startFrame, i = 0; i < count; frame += step, i++)
+	{
+		l[n] = QPointF (xb + ((frame - first) * double(w) / nb_frames),
+			(yb - (m_data[frame][0] * y_space * m_amplification)));
+		r[n] = QPointF (xb + ((frame - first) * double(w) / nb_frames),
+			(yb - (m_data[frame][1] * y_space * m_amplification)));
 		++n;
 	}
```
IanCaio, line 941: This was already discussed on Discord, but since there were lots of messages where I was still figuring some things out I'm trying to make a more organized post here. One of the things that makes this section of the code confusing is that the naming of variables is very misleading. `_to_frame` for example isn't the last frame to be drawn, but instead points to the frame after the last frame to be drawn (like `std::end`, same applies to `m_frame` since it's the length of the buffer).

Another thing I noticed that completely went over our heads is that we didn't need to rewrite this loop at all. When the sample is reversed, the frames inside `m_data` are already reversed by using a call to `std::reverse`. So everything we had to do was draw them in the order they are laid in the buffer. The only reason the updated loop is actually drawing properly (and not drawing the reverse of an already reversed buffer) is the calculation of the X position of the point: Because we had `frame - first` it was actually not making any difference that we were drawing from the last to the first frame, because we were placing them in the same X position as before.

This section of the code doesn't need to be changed, we don't even need to check if the sample is reversed. But I'd propose the following changes:

```suggestion
	const int first = focus_on_range ? _from_frame : 0;
	const int last = focus_on_range ? _to_frame - 1 : m_frames - 1;

	for (int frame = first; frame <= last; frame += fpp)
	{
		auto x = xb + ((frame - first) * double(w) / nb_frames);
		// Partial Y calculation
		auto py = y_space * m_amplification;
		l[n] = QPointF(x, (yb - (m_data[frame][0] * py)));
		r[n] = QPointF(x, (yb - (m_data[frame][1] * py)));
		++n;
	}
```

The only changes to the loop we had originally is to make `last` point to the last frame instead of the frame after the last and @Veratil 's suggestion of moving some repeating calculation to outside the method call. I haven't tested it yet, but I believe it will draw reversed samples properly. Could you make the changes and test it @russiankumar ?

There are other changes I wish to do, like renaming `_to_frame` to something more accurate (or making it actually point to the last frame to be drawn), but that's outside the scope of this PR and would probably fit better in the refactor draft I've going locally.

russiankumar: Alright, I'll be making these changes, and see if these work. I'll edit this comment if it does, and if it doesn't, I'll implement the two loops idea we were discussing on Discord.

IanCaio: Awesome, I wouldn't worry too much about the name. I'm kind of doing a small refactor of `SampleBuffer.cpp` that I plan to submit after you merge this one and I fix the conflicts, I could change it there. Also, changing the name might span a few more files (the header and I think some AFP files).

IanCaio (approved):
Code LGTM!

Just a note for other reviewers, I know that some small details of the code style isn't 100% fixed (@russiankumar did a great job improving it, but `SampleBuffer.cpp` has quite a lot of stuff to be fixed), but I'm working [locally](https://github.com/LMMS/lmms/compare/master...IanCaio:feature/sampleBufferRefactor) on a small refactor for this file which should fix the remaining issues. It's basically ready (need to double check), I'm just waiting for this one to be merged so I can fix the conflicts.

russiankumar:
Great! If @PhysSong and/or @Veratil approve, you can merge this.

@IanCaio, the refactor you’re doing seems like a little sneak peak of what the LMMS refactor is going to be like, IMO xD

src/core/SampleBuffer.cpp
```diff
@@ -296,75 +296,43 @@ void SampleBuffer::update( bool _keep_settings )
 }
 
 
-void SampleBuffer::convertIntToFloat ( int_sample_t * & _ibuf, f_cnt_t _frames, int _channels)
+void SampleBuffer::convertIntToFloat (int_sample_t * & ibuf, f_cnt_t frames, int channels)
```
Veratil, line 299:
```suggestion
void SampleBuffer::convertIntToFloat(int_sample_t * & ibuf, f_cnt_t frames, int channels)
```

src/core/SampleBuffer.cpp
```diff
@@ -296,75 +296,43 @@ void SampleBuffer::update( bool _keep_settings )
...
+		idx += isReversed ? -channels : channels;
 	}
 
-	delete[] _ibuf;
+	delete[] ibuf;
 }
 
-void SampleBuffer::directFloatWrite ( sample_t * & _fbuf, f_cnt_t _frames, int _channels)
-
+void SampleBuffer::directFloatWrite (sample_t * & fbuf, f_cnt_t frames, int channels)
```
Veratil, line 319:
```suggestion
void SampleBuffer::directFloatWrite(sample_t * & fbuf, f_cnt_t frames, int channels)
```

src/core/SampleBuffer.cpp
```diff
@@ -935,39 +903,41 @@ f_cnt_t SampleBuffer::getPingPongIndex( f_cnt_t _index, f_cnt_t _startf, f_cnt_t
 }
 
 
-void SampleBuffer::visualize( QPainter & _p, const QRect & _dr,
-							const QRect & _clip, f_cnt_t _from_frame, f_cnt_t _to_frame )
+void SampleBuffer::visualize (QPainter & p, const QRect & dr,
+							  const QRect & clip, f_cnt_t from_frame, f_cnt_t to_frame)
```
Veratil, line 907: Curious on thoughts about long parameter lists that go past line length:
1. just leave it a long line (ignore line length)
```suggestion
void SampleBuffer::visualize(QPainter & p, const QRect & dr, const QRect & clip, f_cnt_t from_frame, f_cnt_t to_frame)
```
2. line per parameter
```suggestion
void SampleBuffer::visualize(
	QPainter & p,
	const QRect & dr,
	const QRect & clip,
	f_cnt_t from_frame,
	f_cnt_t to_frame
)
// end parenthesis could be combine with the last parameter, alternatively combine the opening brace with the close parenthesis, e.g.
// ) {
```
3. the oddball, but I've seen some libraries that do this, separate the return type, function name, and then parameters
```suggestion
void
SampleBuffer::visualize(
 /* parameter list same as 2 */
) {
```

src/core/SampleBuffer.cpp
```diff
@@ -935,39 +903,41 @@ f_cnt_t SampleBuffer::getPingPongIndex( f_cnt_t _index, f_cnt_t _startf, f_cnt_t
...
 {
-	if( m_frames == 0 ) return;
+	if(m_frames == 0) return;
```
Veratil, line 909:
```suggestion
	if (m_frames == 0) { return; }
```

src/core/SampleBuffer.cpp
```diff
@@ -935,39 +903,41 @@ f_cnt_t SampleBuffer::getPingPongIndex( f_cnt_t _index, f_cnt_t _startf, f_cnt_t
...
+		r[n] = QPointF(x, (yb - (m_data[frame][1] * py)));
 		++n;
 	}
-	_p.setRenderHint( QPainter::Antialiasing );
-	_p.drawPolyline( l, nb_frames / fpp );
-	_p.drawPolyline( r, nb_frames / fpp );
+
+	p.setRenderHint( QPainter::Antialiasing );
+	p.drawPolyline( l, nb_frames / fpp );
+	p.drawPolyline( r, nb_frames / fpp );
```
Veratil, line 940:
```suggestion
	p.setRenderHint(QPainter::Antialiasing);
	p.drawPolyline(l, nb_frames / fpp);
	p.drawPolyline(r, nb_frames / fpp);
```

src/tracks/SampleTrack.cpp
```diff
@@ -280,6 +280,10 @@ void SampleTCO::saveSettings( QDomDocument & _doc, QDomElement & _this )
 	{
 		_this.setAttribute( "color", color().name() );
 	}
+	if(m_sampleBuffer->reversed())
```
Veratil, line 283:
```suggestion
	if (m_sampleBuffer->reversed())
```

src/tracks/SampleTrack.cpp
```diff
@@ -310,11 +314,18 @@ void SampleTCO::loadSettings( const QDomElement & _this )
 		useCustomClipColor( true );
 		setColor( _this.attribute( "color" ) );
 	}
+
+	if(_this.hasAttribute("reversed"))
```
Veratil, line 318:
```suggestion
	if (_this.hasAttribute("reversed"))
```

src/tracks/SampleTrack.cpp
```diff
@@ -310,11 +314,18 @@ void SampleTCO::loadSettings( const QDomElement & _this )
...
+	{
+		m_sampleBuffer->setReversed(true);
+		emit wasReversed();		// tell SampleTCOView to update the view
```
Veratil, line 321: I don't think we need double tabs here
```suggestion
		emit wasReversed(); // tell SampleTCOView to update the view
```

src/tracks/SampleTrack.cpp
```diff
@@ -310,11 +314,18 @@ void SampleTCO::loadSettings( const QDomElement & _this )
...
+	}
 }
 
 
 
 
+
```
Veratil, line 328:
```suggestion
```

src/tracks/SampleTrack.cpp
```diff
@@ -335,6 +346,9 @@ SampleTCOView::SampleTCOView( SampleTCO * _tco, TrackView * _tv ) :
 	connect( m_tco, SIGNAL( sampleChanged() ),
 			this, SLOT( updateSample() ) );
 
+	connect(m_tco, SIGNAL(wasReversed()),
+			this, SLOT(update()));
```
Veratil, line 350: This doesn't line up unless tabs are 4 spaces, might as well just make it a single line.
```suggestion
	connect(m_tco, SIGNAL(wasReversed()), this, SLOT(update()));
```

src/tracks/SampleTrack.cpp
```diff
@@ -408,6 +422,9 @@ void SampleTCOView::contextMenuEvent( QContextMenuEvent * _cme )
 				tr( "Set/clear record" ),
 						m_tco, SLOT( toggleRecord() ) );*/
 
+	contextMenu.addAction(embed::getIconPixmap("flip_x"),
+			tr("Reverse sample"), this, SLOT(reverseSample()));
```
Veratil, line 426:
```suggestion
	contextMenu.addAction(
		embed::getIconPixmap("flip_x"),
		tr("Reverse sample"),
		this,
		SLOT(reverseSample()));
```
alternatively (my preference)
```suggestion
	contextMenu.addAction(
		embed::getIconPixmap("flip_x"),
		tr("Reverse sample"),
		this,
		SLOT(reverseSample())
	);
```

Veratil (changes requested):

src/core/SampleBuffer.cpp
IanCaio: I prefer 2! About the code style review, I think it would be easier for @russiankumar (and myself when dealing with the conflicts :laughing: ) if we leave less critical style fixes for #5791 . `SampleBuffer.cpp` has tons of them and I had to go through the whole file so I tried to fix most. The parameter list of `visualize` is one I think I fixed on the cpp file but not on the header. Will update it later!

Veratil: The main thing here is removing the space between the function name and the opening parenthesis. If we touch the line it needs to be updated to the new convention, period. I was a huge stickler about it before, and then everyone complained that it didn't matter and code should look the same as the surrounding, and I backed down since everyone was so adamant about it, but I'm tired of seeing people ignore it and I'm going to request fixes to conform to the new style. 😛

As for the ideas there, just curious what people thought. 😄

IanCaio: I didn't mean to imply you were nitpicking, hope it didn't sound like that. I was just working in parallel in a code style revamping on that file that could fix that too. But you're totally right, it should be fixed on the PR that touches the line and we should be strict about it!

russiankumar:
@Veratil I've made the requested changes.

src/core/SampleBuffer.cpp
russiankumar: > As for the ideas there, just curious what people thought. 😄

Leaving this open in case anyone wants to drop their thoughts.

include/SampleBuffer.h
```diff
@@ -117,10 +117,10 @@ class LMMS_EXPORT SampleBuffer : public QObject, public sharedObject
 				const float _freq,
 				const LoopMode _loopmode = LoopOff );
 
-	void visualize( QPainter & _p, const QRect & _dr, const QRect & _clip, f_cnt_t _from_frame = 0, f_cnt_t _to_frame = 0 );
-	inline void visualize( QPainter & _p, const QRect & _dr, f_cnt_t _from_frame = 0, f_cnt_t _to_frame = 0 )
+	void visualize (QPainter & p, const QRect & dr, const QRect & clip, f_cnt_t from_frame = 0, f_cnt_t to_frame = 0);
+	inline void visualize (QPainter & p, const QRect & dr, f_cnt_t from_frame = 0, f_cnt_t to_frame = 0)
```
Veratil, line 121:
```suggestion
	void visualize(QPainter & p, const QRect & dr, const QRect & clip, f_cnt_t from_frame = 0, f_cnt_t to_frame = 0);
	inline void visualize(QPainter & p, const QRect & dr, f_cnt_t from_frame = 0, f_cnt_t to_frame = 0)
```

include/SampleBuffer.h
```diff
@@ -266,8 +266,8 @@ public slots:
 
 	void update( bool _keep_settings = false );
 
-	void convertIntToFloat ( int_sample_t * & _ibuf, f_cnt_t _frames, int _channels);
-	void directFloatWrite ( sample_t * & _fbuf, f_cnt_t _frames, int _channels);
+	void convertIntToFloat (int_sample_t * & ibuf, f_cnt_t frames, int channels);
+	void directFloatWrite (sample_t * & fbuf, f_cnt_t frames, int channels);
```
Veratil, line 270:
```suggestion
	void convertIntToFloat(int_sample_t * & ibuf, f_cnt_t frames, int channels);
	void directFloatWrite(sample_t * & fbuf, f_cnt_t frames, int channels);
```

src/tracks/SampleTrack.cpp
```diff
@@ -280,9 +280,9 @@ void SampleTCO::saveSettings( QDomDocument & _doc, QDomElement & _this )
 	{
 		_this.setAttribute( "color", color().name() );
 	}
-	if(m_sampleBuffer->reversed())
+	if (m_sampleBuffer->reversed())
 	{
-		_this.setAttribute("reversed", "true");
+		_this.setAttribute ("reversed", "true");
```
Veratil, line 285:
```suggestion
		_this.setAttribute("reversed", "true");
```

Veratil (changes requested):

src/tracks/SampleTrack.cpp
russiankumar: Can you clarify on whether the opening brackets should have a space before them or not? You seem to have used different formats in different spaces; I'd like to keep things uniform now that we're changing the style.

Veratil: This is a function call, there should be no space between the function name and the opening parenthesis. The only time there is a space before one is when it's preceded by a keyword (`if`, `for`, etc.).

What snippet did I make that added a space?

russiankumar: > The only time there is a space before one is when it's preceded by a keyword (if, for, etc.).

Ah, didn't know that was the case, I thought it was supposed to be uniform across all. Thanks for the clarification!

russiankumar:
@Veratil now maybe? 😂

Veratil (approved):

russiankumar:
Build error seems time-related, and all other are passing. Merging.

Issue #5757, opened 2020-10-31
Effect Module rectangle edge does not extend to veritcal scroll bar


sound8:
Effect Module rectangle edges on right side do not extend to veritcal scroll bar.
If you load more than 4 effects so that the vertical scroll bar shows up you will see there is an odd spacing compared to the left side when the vertical scroll bar is visible.

Kind of goes along with my issue post #5752 as well as @Spekular response to this post about the vertical scrollbar space when scroll bar is not visible.
Probably would all be worked on at the same time.

EDIT:
Would also have to recenter "controls" button and could also go with #5750 if "GATE" knob gets moved over to the right

Issue #5752, opened 2020-10-29
Effect module names get cut off abrubtly if too long by modules edge


sound8:
Effect module names get cut off abrubtly if names are too long.
The names go literally right to the effect modules rectangle vertical edge and stop, does not look good graphically.

When this happens with other labels/names that are too long the suffix "..." is usually added to the end and a little space after so names don't get chopped off by vertical graphical edges.  Effect module names should get this to keep things consistant.

For example type a really long label name in Song Editor track and watch how it adds the "..." and space after.

Spekular:
For demonstrations sake I loaded a VSTi into an effects rack:
![image](https://user-images.githubusercontent.com/8918426/97627220-56f8e480-1a2b-11eb-821e-f0c85cf43666.png)

Sidenote, since effects in the rack can't make use of the additional space (at least not now), we may as well show the scrollbar area all the time, right? Otherwise it looks a bit odd when the space is unused.

Issue #5750, opened 2020-10-29
Abbreviate effect module decay knob label


sound8:
Effect modules have three knobs
W/D, DECAY, GATE

The "DECAY" label kind of runs into the GATE label, to me it just doesn't look right.
Maybe consider abbreviating the "DECAY" label to "DEC" so the labels are not so close together.

Obviously not a big deal but would clean up the look of things.

Spekular:
The real issue is that the space between decay and gate is shorter than the space between W/D and decay. If this were fixed the labels would be fine as is.
![image](https://user-images.githubusercontent.com/8918426/97625551-f36db780-1a28-11eb-929b-32c71148b149.png)

Measuring pixels in that screenshot, the first margin is 9 pixels and the second is 3.

sound8:
If you are referring to the spacing between the knobs then I would agree.
The only reason I didn't mention this is because I figured I would have a better chance of someone abbreviating a text label, over moving a knob and label a little to the right.
So yeah, two ways to fix this.
1.) Abbreviate "DECAY" label
2.) Move Gate knobw and "GATE" label over to the right.

PR #5766, opened 2020-11-04
Fix EffectRackView appearance (GUI).

DigArtRoks:
* Elide the name of the effect when it tends to be too big. (#5752)
* Evenly space the controls (W/D, Decay Gate). (#5750)
* Show the scrollbar in the default theme to close the gap. (#5752)
* Reduce the gap between the effect and the scrollbar. (#5757)
* Use always the same width for the EffectRackview (InstrumentTrack and SampleTrack) to avoid gaps or cutoffs of the background.
* Widen the background in the default theme.
* Widen the embossed space in the background in the classic theme to fit the controls.

afbeelding afbeelding


superpaik:
It seems perfect to me. Tested on W10/64 bits (both MSVC and Mingw) and with different display resolutions.
Just one comment. You said (point 5): Use always the same width for the EffectRackview (InstrumentTrack and SampleTrack) to avoid gaps or cutoffs of the background.
I think that applies to Effects Chain in FX mixer as well. Or at least the width seems to be the same, which I find correct.

Thanks.

DigArtRoks:
Thansk for trying it out.

Yes the width of the EffectRackView is now the same for FxMixer, InstrumentTrack and SampleTrack.

I initially thought the issue was flagged for the FxMixer alone. After fixing that one, I still saw similar issues in the InstrumentTrack and SampleTrack. Hence my comment in point 5 only referring to the latter 2.

It turned out that for the 3 locations where the EffectRackView is used different widths (and heights) are used. So I fixed the width for all 3 to the same value. The widget does not really allow horizontal resizing due to the fixed background that is used for an EffectView. Either more space between the EffectView and the scrollbar appears or the background is cropped.

The width remains as was for the FxMixer, gets 5 pixels wider for the SampleTrack (which still fits) and is reduced about 5 pixels in the InstrumentTrack.

Spekular:
I've put together a before and after (Ignore the fact that my before image doesn't gray out disabled knobs):
image

I think the classic theme effects now end up a bit too close to the scrollbar, I reckon around halfway between the old width and the new one would look good.

Similarly, I perceive the right side padding on the default theme effects to be narrower than the left, even though they're both 4 pixels. Perhaps 5 pixels on the left would give the scrollbar a little more space to breathe (since it stands out more than the outline). Here's a mockup + comparison of how that would look:
image
(PR on the left, mockup on the right, space in the middle so you can focus on one at a time)

DomClark (approved):
Code looks good.

I do wonder, however, whether the knobs are a little too far apart now. Also, it looks like a dark line has appeared at the bottom right of the classic theme effect background, although it's not very noticeable.

DigArtRoks:
Thanks for the reviews.

I did an additional commit.
* Reduce the background for the default theme by 1 pixel.
* Reduce the background for the classic theme by 2 pixels and remove the darker line at the bottom right.
* Reduce the width of long names of the plugin also by 2 pixels.
* Put the controls 2 pixels closer to each other.

This is how it looks now:
afbeelding afbeelding

superpaik:
Yeah, much better now!

Spekular (approved):
LGTM both code and design wise.

sound8:
I was freaking out because I found a whole bunch of other little things to go with my posts, I created all these graphics, I rushed here to tell you guys....  and then I found this page where you not only found them all but applied them! :)
Wow, excellent job guys, everything looks great
Thank You for all this work.

Spekular:
> Should I close my original posts?

We usually only close issues when the fix is merged, so leave it open :)

Spekular:
Actually though, Dom gave the changes after his latest review a thumbs up reaction, so I suppose this has two up to date + passing reviews and no CI failures. Merging!

Issue #5771, opened 2020-11-07
Plugin blacklist


JohannesLorenz:
### Enhancement Summary

A file answering the Question
```
Q(T, P, PV, H, HV):
Are there known issues with
* Plugin P
* of Type T (LADSPA, Lv2, VST2, VST3, ...)
* in Plugin-Version PV
* in Host H in Host-version HV?
```

#### Justification

Some plugins may be defective, causing ear or speaker damage. E.g. #5767 or test reports from #5761 .

#### Open Questions:

1. Should the blacklist be a separate repo (submodule)? In my opinion: Yes. It can be useful for other DAWs.
2. What file format would be suited? Maybe a simple config file, like
```
# lv2 plugins with the URI specified are buggy in versions 2.4.1 - 2.4.3 in LMMS in versions 1.0.0 to 1.2.0
lv2 http://lsp-plug.in/plugins/lv2/crossover_ms 2.4.1-2.4.3 LMMS 1.0.0-1.2.0
# ladspa plugins with ladspa ID 1902 from version 1.0.0 to current are buggy in all hosts in all versions
ladspa 1902 1.0.0- all all
```
3. ...?

tresf:
Reminds me of https://github.com/LMMS/lmms/blob/5efb3a19cb13e4f3ad23c5d75dcd07a140a7ee86/plugins/LadspaEffect/LadspaEffect.cpp#L574-L590

Issue #5767, opened 2020-11-04
GLAME Butterworth Lowpass/Highpass (Lv2 versions) generate damaged audio


JohannesLorenz:
### Bug Summary

GLAME Butterwort Lowpass and Highpass (Lv2 versions) generate damaged audio in many cases. LADSPA versions work. Lv2 versions work in Audio. Conclusion: LMMS Lv2 implementation is broken.

Thanks to @IanCaio for reporting!

#### Steps to reproduce

Turn off speakers. Then use the effect and play any audio through them.

#### Expected behavior

Normal filter output.

#### Actual behavior

Corrupted audio. Can cause ear or speaker damage.

#### Affected LMMS versions

Master only.

zonkmachine:
Got a backtrace



#### Logs
<details>
  <summary>Click to expand</summary>
<pre>


Thread 13 "Mixer::fifoWrit" received signal SIGFPE, Arithmetic exception.
[Switching to Thread 0x7fffc55bc700 (LWP 18867)]
iir_process_buffer_1s_5 (iirf=<optimized out>, iirf=<optimized out>, gt=<optimized out>, numSampsToProcess=32, outdata=0x555559f752b0, indata=0x555559f751c0)
    at ./util/iir.h:243
243	./util/iir.h: No such file or directory.
(gdb) bt full
#0  iir_process_buffer_1s_5
    (iirf=<optimized out>, iirf=<optimized out>, gt=<optimized out>, numSampsToProcess=32, outdata=0x555559f752b0, indata=0x555559f751c0) at ./util/iir.h:243
        pos = 0
        plugin_data = <optimized out>
        cutoff = 0.112575002
        resonance = <optimized out>
        input = 0x555559f751c0
        output = 0x555559f752b0
        iirf = <optimized out>
        gt = 0x555559f46f20
        sample_rate = <optimized out>
#1  runButtlow_iir (instance=<optimized out>, sample_count=32) at plugins/butterworth-swh.lv2/plugin.c:210
        plugin_data = <optimized out>
        cutoff = 0.112575002
        resonance = <optimized out>
        input = 0x555559f751c0
        output = 0x555559f752b0
        iirf = <optimized out>
        gt = 0x555559f46f20
        sample_rate = <optimized out>
#2  0x0000555555837f99 in lilv_instance_run(LilvInstance*, uint32_t) (instance=0x555559ecc4b0, sample_count=32) at /usr/include/lilv-0/lilv/lilv.h:1704
[...]

PR #5774, opened 2020-11-08
Blacklist SWH GLAME filters for Lv2

JohannesLorenz:
Hotfix for #5767. Implement a blacklist and add the SWH GLAME Lv2 filters to it.

This is just temporary, but I think it's important to take those filters out before more users hear that noise.

Reviewer hints:
1. The implementation may not be 100% perfect:
  a. it should be in the core, not restricted to Lv2
  b. a blacklist should be more general, as suggested in #5771 
2. ~~Wait for CI before review~~

Tester hints:
1. This PR does not require testing IMO, I tested the few bits myself. However, if you really want to test anything: Make sure you use "dummy audio output" in the LMMS settings and have restarted before. The audio produced by those effects is horrible.

### DIFF
--- a/include/Engine.h
+++ b/include/Engine.h
@@ -88,6 +88,8 @@ public:
 		return s_projectJournal;
 	}
 
+	static bool ignorePluginBlacklist();
+
 #ifdef LMMS_HAVE_LV2
 	static class Lv2Manager * getLv2Manager()
 	{
--- a/include/Lv2Manager.h
+++ b/include/Lv2Manager.h
@@ -131,6 +131,11 @@ public:
 	}
 	bool isFeatureSupported(const char* featName) const;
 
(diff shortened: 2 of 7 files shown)


JohannesLorenz:
As this is a fix for a critical issue, I'll merge this without review in 3 days unless anyone has objections.

PhysSong:
Please don't merge this now. I figured out the cause of it.

zonkmachine:
I have a hunch the blacklist function will come in handy again later anyway. It's quite handy to have the option to temporarily remove plugins that cause trouble while we develop this.

JohannesLorenz:
Agreed to both of you. It should not be merged like this. However, e.g. Calf Analyzer and Calf Bass Enhancer cause crashes, so I'lll remove the GLAME filters (handled by #5783) and add those Calf effects.

PR #5775, opened 2020-11-09
Fix duplicate message in de.ts

PauseForAffliction:
Fixes a duplicate translation message in the German translation file. This addresses the (warranted but vague) error passed by Qt Creator.

ts_error

The removed message is an unfinished translation while the message it is a duplicate of (on line 3984) is a finished translation.

### DIFF
--- a/data/locale/de.ts
+++ b/data/locale/de.ts
@@ -4137,10 +4137,6 @@ Bitte besuchen Sie http://lmms.sf.net/wiki für Dokumentationen über LMMS.</tra
 Please make sure you have write permission to the file and the directory containing the file and try again!</source>
         <translation type="unfinished"/>
     </message>
-    <message>
-        <source>Export &amp;MIDI...</source>
-        <translation type="unfinished"/>
-    </message>
 </context>
 <context>
     <name>MeterDialog</name>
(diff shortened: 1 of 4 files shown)


zonkmachine:
Duplicate translation may have been introduced here: https://github.com/LMMS/lmms/commit/457f2c67126bf070bb3fdd7fcb84514f0ba7a530

JohannesLorenz:
> Duplicate translation may have been introduced here: [457f2c6](https://github.com/LMMS/lmms/commit/457f2c67126bf070bb3fdd7fcb84514f0ba7a530)

Actually, it was added there because it was missing:
```
$ git show 457f2c6:data/locale/de.ts | grep  'Export.*MIDI' | wc -l
1
$ git show 457f2c6~1:data/locale/de.ts | grep  'Export.*MIDI' | wc -l
0
```

The other occurrence is from @liushuyu from 2017, but it's definitely not in 457f2c6.

Aside from that, translations should always be fixed via Transifex, since changes will be overwritten by Transifex. In this case, however, we may just merge, as it includes a duplicate?

PauseForAffliction:
Reading through \#devtalk I realize this PR wasn't the way of going about this 😅

In its stead, what would be the correct way of notifying the Transifex team about this issue?

PauseForAffliction:
Whatever the case, I'm going to close this pull request. Not only would it be overwritten anyway, but I botched it up a little because I'm bad at git. Thanks for the consideration and feedback.

Issue #5733, opened 2020-10-25
rpmalloc causes crash on exiting LMMS


JohannesLorenz:
### Bug Summary

On exiting LMMS, LMMS crashes with the following terminal output (see *Logs* for the backtrace):
```
lmms: /data/cprogs/lmms/master/src/3rdparty/rpmalloc/rpmalloc/rpmalloc/rpmalloc.c:1248: _rpmalloc_span_finalize: Assertion `span->list_size == span->used_count' failed.
Aborted (core dumped)
```

Caused by rpmalloc update in c58c781f9331a646a1cdc4f4fc15b67c98214a4f. Might be not our fault.

#### Steps to reproduce

Compile current master. Start LMMS. Close LMMS (File->Quit).

#### Expected behavior

No crash.

#### Actual behavior

Crash.

#### Affected LMMS versions

Master is affected since c58c781f9331a646a1cdc4f4fc15b67c98214a4f.
stable-1.2 is not affected.

#### Logs
<details>
  <summary>Click to expand</summary>
<pre>
(gdb) bt
#0  0x00007ffff60f2615 in raise () at /usr/lib/libc.so.6
#1  0x00007ffff60db862 in abort () at /usr/lib/libc.so.6
#2  0x00007ffff60db747 in _nl_load_domain.cold () at /usr/lib/libc.so.6
#3  0x00007ffff60eabf6 in  () at /usr/lib/libc.so.6
#4  0x0000555555933ffe in _rpmalloc_span_finalize (heap=0x7ffff3190058, iclass=66, span=0x7ffff2d80000, list_head=0x7ffff3190698) at /data/cprogs/lmms/master/src/3rdparty/rpmalloc/rpmalloc/rpmalloc/rpmalloc.c:1248
#5  0x0000555555935843 in _rpmalloc_heap_finalize (heap=0x7ffff3190058) at /data/cprogs/lmms/master/src/3rdparty/rpmalloc/rpmalloc/rpmalloc/rpmalloc.c:1833
#6  0x00005555559348b9 in _rpmalloc_heap_global_finalize (heap=0x7ffff3190058) at /data/cprogs/lmms/master/src/3rdparty/rpmalloc/rpmalloc/rpmalloc/rpmalloc.c:1430
#7  0x0000555555937b35 in rpmalloc_finalize () at /data/cprogs/lmms/master/src/3rdparty/rpmalloc/rpmalloc/rpmalloc/rpmalloc.c:2699
[...]

PR #5776, opened 2020-11-10
Address the memory leak that the rpmalloc assert warns of on program exit

PauseForAffliction:
Hack to take care of the error dialogue sent by the rpmalloc memory manager (#5733 ). Creates a static "```free```" function for ```NotePlayHandleManager``` and then shoves it right before the program ends in ```main.cpp```.

image (The error dialogue)


JohannesLorenz (approved):
Review passed. Thanks for the fix, I'll merge.

Issue #5714, opened 2020-10-16
Truncated sidebar items' font size behaves incorrectly


Spekular:
Tested:
- Windows, master: Truncated items display as bigger(!) than non-truncated ones.
- Windows, 1.2.2: Truncated items display at a smaller size, but the truncation is still calculated for the normal size. This leaves unused space.
- Ubuntu 20, master: Font size is the same for truncated/non-truncated items
On master + windows, when a sidebar item's full name can't be displayed, its font size increases. In 1.2.2, the font size is reduced instead, and on a master build + linux the size is unchanged.

Screenshot of Windows master behavior:
![image](https://user-images.githubusercontent.com/8918426/96288554-e346e880-0fe3-11eb-853b-7f5b03b2fc18.png)

Screenshot of Windows 1.2.2 behavior:
![image](https://user-images.githubusercontent.com/8918426/96290350-bd6f1300-0fe6-11eb-9b36-e5b99242c584.png)

I propose this be fixed by drawing all items at the same font size, regardless of truncation. *If* we want to keep the changing font size, I think the behavior should be as on windows + 1.2.2 but with truncation based on the actual font size.

Spekular:
Tentatively milestoning for 1.3 because this should be a simple fix (removing logic rather than adding it). I haven't investigated yet though, so if it turns out to be a Qt bug or something else tricky the milestone can be dropped.

Spekular:
Fixed by #5777, thanks @DigArtRoks :)

PR #5777, opened 2020-11-10
Fix for the font of truncated sidebar items (#5714).

DigArtRoks:
For windows platforms, retrieve the system font and set it for the FileBrowserTreeWidget. This makes sure that truncated items will use the font as non-truncated items.

### DIFF
--- a/include/GuiApplication.h
+++ b/include/GuiApplication.h
@@ -28,6 +28,7 @@
 #include <QtCore/QObject>
 
 #include "lmms_export.h"
+#include "lmmsconfig.h"
 
 class QLabel;
 
@@ -48,6 +49,9 @@ public:
 	~GuiApplication();
 
 	static GuiApplication* instance();
+#ifdef LMMS_BUILD_WIN32
+	static QFont getWin32SystemFont();
+#endif
 
 	MainWindow* mainWindow() { return m_mainWindow; }
 	FxMixerView* fxMixerView() { return m_fxMixerView; }
--- a/src/gui/FileBrowser.cpp
+++ b/src/gui/FileBrowser.cpp
@@ -55,8 +55,6 @@
 #include "StringPairDrag.h"
 #include "TextFloat.h"
 
-
-
 enum TreeWidgetItemTypes
 {
 	TypeFileItem = QTreeWidgetItem::UserType,
@@ -335,6 +333,11 @@ FileBrowserTreeWidget::FileBrowserTreeWidget(QWidget * parent ) :
 	connect( this, SIGNAL( itemExpanded( QTreeWidgetItem * ) ),
 				SLOT( updateDirectory( QTreeWidgetItem * ) ) );
 
+#ifdef LMMS_BUILD_WIN32
+	// Set the font for the QTreeWidget to the Windows System font to make sure that
+	// truncated (elided) items use the same font as non-truncated items.
+	setFont( GuiApplication::getWin32SystemFont() );
+#endif
 }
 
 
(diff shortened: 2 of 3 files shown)


DigArtRoks:
It looks like there is a shortcoming/bug in QT. When it truncates the text of an item because the sidebar is made smaller, it uses another font (in particular MS Shell Dlg 2 vs. Segoe UI for non-truncated items on my Win10 box).

The issue is flagged here: [https://bugreports.qt.io/browse/QTBUG-29232](url).
Also the QTableWiget seems to have a similar issue: [https://bugreports.qt.io/browse/QTBUG-22572](url).

Both bugreports do a suggestion to fix it or workaround it.
* Set font-family in the stylesheet. This works also fine for LMMS, but in the stylesheet there is no font at all defined. Also a font may not be available or not the same as the system font that is used for other QT Widgets. So I rejected this.
* Retrieve the font that is used by Windows for Non-Client area elements and set it as font for the QApplication. This also works fine for LMMS, but for some reason it changes quite a lot of other text labels in the UI as well.

Taking it further I implemented the second proposal and apply the retrieved font only on the FileBrowserTreeWidget. Resizing the sidebar now looks clean when text items get truncated.
I was not sure where to put the function (getWin32SystemFont) to retrieve the system font. Please advice if there is a better location. The fix is only applicable for Windows builds, but I understood from the issue that it is anyway only Windows related.

@Spekular, unfortunately it's adding logic instead of removing logic.

Spekular:
Tested and working for me. I don't see any issues with the code itself but I'd love to hear from someone else, especially regarding this: 

> I was not sure where to put the function (getWin32SystemFont) to retrieve the system font. Please advice if there is a better location.

As for this:

> @Spekular, unfortunately it's adding logic instead of removing logic.

That is a bit unfortunate, but worth it IMO. That being said, I would add a link to the qt bug report and a searchable keyword to the comments. That should make it easier to remove this code when it becomes obsolete. Something like

```
// Workaround this qt bug, fixed in 5.12.2: https://bugreports.qt.io/browse/QTBUG-29232
// TODO: remove this when all builds use a recent enough version of qt
```

superpaik:
It works perfect. Tested on both 32 and 64bits, on Windows10

DigArtRoks:
Thanks!
I added the suggested comments in an additional commit.

Spekular (approved):

Spekular:
Merging ~24h from now if there are no objections. Shouldn't be hard to move the font function if it does belong elsewhere :P

DomClark:
How about adding the condition `QT_VERSION < QT_VERSION_CHECK(5, 12, 2)` in addition to `LMMS_BUILD_WIN32`? Quoting myself from #5598:

> Knowing which parts can be removed later is easy - just search the codebase for `QT_VERSION`. This is something we've already done before, to remove Qt 4 compatibility code.

As for where to put `getWin32SystemFont`, I think `GuiApplication` is fine if it could be useful elsewhere, otherwise I would put it in an anonymous namespace in `src/gui/FileBrowser.cpp`. Since the Qt bug report mentions other cases where the bug can occur, let's keep the function where it is for now in case the bug crops up somewhere else before we update Qt.

DigArtRoks:
Thanks. Good suggestion! I added the check and commited again.

Spekular:
Good suggestion indeed, I'll have to remember that for the future. My approving review stands, @DomClark if the newly added check is to your satisfaction I'm still in favor of a merge.

DomClark (approved):

Spekular:
🎉

PR #5783, opened 2020-11-12
Fix missing support for lv2core#sampleRate (Fixes #5767)

### LINKED ISSUE #5767: GLAME Butterworth Lowpass/Highpass (Lv2 versions) generate damaged audio

JohannesLorenz:
This multiplies port's min/max value with the processing sample rate
that is used for the plugin. This fixes damaged audio in GLAME
Butterworth High-/Lowpass from #5767.

### DIFF
--- a/include/Lv2Ports.h
+++ b/include/Lv2Ports.h
@@ -108,11 +108,17 @@ struct Meta
 	Flow m_flow = Flow::Unknown;
 	Vis m_vis = Vis::None;
 
-	float m_def = .0f, m_min = .0f, m_max = .0f;
 	bool m_optional = false;
 	bool m_used = true;
 
 	std::vector<PluginIssue> get(const LilvPlugin* plugin, std::size_t portNum);
+
+	float def() const { return m_def; }
+	float min(sample_rate_t sr) const { return m_sampleRate ? sr * m_min : m_min; }
+	float max(sample_rate_t sr) const { return m_sampleRate ? sr * m_max : m_max; }
+private:
+	float m_def = .0f, m_min = .0f, m_max = .0f;
+	bool m_sampleRate = false;
 };
 
 struct PortBase : public Meta
--- a/include/PluginIssue.h
+++ b/include/PluginIssue.h
@@ -42,6 +42,7 @@ enum PluginIssueType
 	portHasNoDef,
 	portHasNoMin,
 	portHasNoMax,
+	defaultValueNotInRange,
 	featureNotSupported, //!< plugin requires functionality LMMS can't offer
 	badPortType, //!< port type not supported
 	noIssue
--- a/src/core/PluginIssue.cpp
+++ b/src/core/PluginIssue.cpp
@@ -50,6 +50,8 @@ const char *PluginIssue::msgFor(const PluginIssueType &it)
 			return "port is missing min value";
 		case portHasNoMax:
 			return "port is missing max value";
+		case defaultValueNotInRange:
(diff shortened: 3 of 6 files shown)


JohannesLorenz:
@zonkmachine I added you as reviewer, but I already checked GLAME Lowpass/Highpass are working. It's just for information.

PhysSong:
Just asking, isn't the default value supposed to be multiplied by sample rate as well?

JohannesLorenz:
@PhysSong Good point. It shall not be multiplied, `sampleRate` is only for multiplying the bounds in order to be confirming to the Nyquist frequency, but having different default values with different sample rates is not wanted. I asked in the lv2 chat to be sure.

However, the GLAME filters got this wrong, too (they pre-divided the default value), so thanks for helping me opening swh/lv2#17:

```
 :port [
     a :InputPort, :ControlPort ;
     :name "Cutoff Frequency (Hz)" ;
     :index 0 ;
     :symbol "cutoff" ;
     :minimum 0.0001 ;
     :maximum 0.45 ;
     
     :default 0.112575 ;
     :portProperty pprops:logarithmic ;
     :portProperty :sampleRate ;
   ] ;
```

In this case, `default` is smaller than `minimum * sampleRate`. Should I write automatic code, that, like in this case, does multiply the default value with the samplerate? Or should we reject such plugins?

PhysSong:
It seems like LADSPA supports both sr-dependent and constant default values, but LV2 don't.
BTW, Have you looked into what other LV2 hosts do in such a case?

JohannesLorenz:
@PhysSong From what I can see in the source code, both Ardour and jalv only multiply min and max, and I can't see that they would do anything to fix pre-divided default values like in the GLAME filters. Also, opening the GLAME filters in ardour will bash the default value to the minimum, which means they don't have any special logic.

Proposal: If "sampleRate" is given and the default value is smaller than "minimum * sampleRate", then it should be multiplied with the sampleRate, too? Or maybe with a fixed sampleRate, like 44100?

JohannesLorenz:
No matter if we are going to add more commits, I'll merge this in 24 hours if no one is opposed. This is a critical fix.

IanCaio:
@JohannesLorenz What about setting the default to `m_min` if it's out of range? Or you can first try to multiply it by the `SampleRate` and if the value is still invalid set it to the minimum.

PhysSong:
Audacity does multiply the default value by sample rate, but GLAME Butterworth filter is broken in Audacity, meaning that we can't use it as a reference.

> What about setting the default to `m_min` if it's out of range?

I think clamping with warning is better, since some plugins may have default value larger than `m_max`. I think it's also better than blacklisting such plugins.

> first try to multiply it by the `SampleRate`

We may open a new issue and keep the discussion there.

JohannesLorenz:
It's clamped to `[min, max]` now (this was also supported in devtalk, mostly because it exposes the issue to the user). I also added a warning to the command line if the default value is not in range.

Should be enough now. I'll merge after CI passes.

Issue #5715, opened 2020-10-16
Allow sidebar filtering by factory/user distinction


Spekular:
#### Enhancement Summary

Allow filtering the sample and preset browsers to show only user content, only factory content, or both. The setting should be independent for each tab. UI-wise I imagine this would either be two separate checkboxes (my preference) or a single combo box, displayed in the browser tab near the search bar. Save the state of this setting per-project.

Alternatively, let this be a global setting, in which case it could be configured in the settings menu.

#### Justification

Users may find factory content to be uninteresting because it is, for example:
- Not suitable for their uses/not to their taste
- Overused by virtue of being factory content
- Not organized according to their preferred system

In this case the items amount to visual clutter. Deleting them from the install folder is suboptimal:
- It prevents loading projects that do use them (older projects, other users' projects)
- It's less user friendly and requires additional file permissions
- It leaves the user with a modified install directory

PR #5786, opened 2020-11-14
Add checkboxes for selecting user and factory content

thmueller64:
Addresses #5715. It is now possible to separate builtin Samples/Projects/Presets from those created by the user.

### DIFF
--- a/src/gui/FileBrowser.cpp
+++ b/src/gui/FileBrowser.cpp
@@ -76,7 +76,6 @@ void FileBrowser::addContentCheckBox()
 
 	auto configCheckBox = [this, &filterWidgetLayout](QCheckBox* box)
 	{
-		box->setObjectName("FileBrowserFilterCheckBox");
 		box->setCheckState(Qt::Checked);
 		connect(box, SIGNAL(stateChanged(int)), this, SLOT(reloadTree()));
 		filterWidgetLayout->addWidget(box);
--- a/src/gui/FileBrowser.cpp
+++ b/src/gui/FileBrowser.cpp
@@ -76,7 +76,7 @@ void FileBrowser::addContentCheckBox()
 
 	auto configCheckBox = [this, &filterWidgetLayout](QCheckBox* box)
 	{
-		box->setAccessibleName("FileBrowserFilterCheckBox");
+		box->setObjectName("FileBrowserFilterCheckBox");
 		box->setCheckState(Qt::Checked);
 		connect(box, SIGNAL(stateChanged(int)), this, SLOT(reloadTree()));
 		filterWidgetLayout->addWidget(box);
--- a/src/gui/FileBrowser.cpp
+++ b/src/gui/FileBrowser.cpp
@@ -65,16 +65,50 @@ enum TreeWidgetItemTypes
 
 
 
+void FileBrowser::addContentCheckBox(void)
+{
+	auto filterWidget = new QWidget(contentParent());
+	filterWidget->setFixedHeight(24);
+	auto filterWidgetLayout = new QHBoxLayout(filterWidget);
+	filterWidgetLayout->setMargin(0);
+	filterWidgetLayout->setSpacing(0);
+
+	auto configCheckBox = [this, &filterWidgetLayout](QCheckBox* box)
+	{
+		box->setStyleSheet("QCheckBox {color: white; font-size: 10px}");
+		box->setCheckState(Qt::Checked);
+		connect(box, SIGNAL(stateChanged(int)), this, SLOT(reloadTree()));
+		filterWidgetLayout->addWidget(box);
+	};
+
+	m_showUserContent = new QCheckBox("User content");
+	configCheckBox(m_showUserContent);
+	m_showFactoryContent = new QCheckBox("Factory content");
+	configCheckBox(m_showFactoryContent);
+
+	addContentWidget(filterWidget);
+};
+
+
 FileBrowser::FileBrowser(const QString & directories, const QString & filter,
 			const QString & title, const QPixmap & pm,
-			QWidget * parent, bool dirs_as_items, bool recurse ) :
+			QWidget * parent, bool dirs_as_items, bool recurse,
+			const QString& userDir,
+			const QString& factoryDir):
 	SideBarWidget( title, pm, parent ),
 	m_directories( directories ),
 	m_filter( filter ),
 	m_dirsAsItems( dirs_as_items ),
-	m_recurse( recurse )
+	m_recurse( recurse ),
+	m_userDir(userDir),
+	m_factoryDir(factoryDir)
 {
 	setWindowTitle( tr( "Browser" ) );
+
+	if ((userDir != "") && (factoryDir != ""))
+	{
+		addContentCheckBox();
+	}
 
 	QWidget * searchWidget = new QWidget( contentParent() );
 	searchWidget->setFixedHeight( 24 );
@@ -162,6 +196,30 @@ bool FileBrowser::filterItems( const QString & filter, QTreeWidgetItem * item )
 }
 
 
+void filterDirectories(QString& dirstring, const QString& filterstring, bool checked)
+{
+	if (!checked)
+	{
+		dirstring = dirstring.replace(filterstring, "");
+	}
+	else
+	{
+		if (!dirstring.contains(filterstring))
+		{
+			dirstring += "*" + filterstring;
+		}
+	}
+	dirstring = dirstring.replace("**", "*");
+	if ((dirstring.length() > 0) && (*dirstring.begin() == '*'))
+	{
+		dirstring = dirstring.remove(0, 1);
+	}
+	if ((dirstring.length() > 0) && (*(dirstring.end()-1) == '*'))
+	{
+		dirstring = dirstring.remove(dirstring.length()-1, 1);
+	}
+}
+
 
 void FileBrowser::reloadTree( void )
 {
@@ -169,10 +227,21 @@ void FileBrowser::reloadTree( void )
 	const QString text = m_filterEdit->text();
 	m_filterEdit->clear();
 	m_fileBrowserTreeWidget->clear();
-	QStringList paths = m_directories.split( '*' );
-	for( QStringList::iterator it = paths.begin(); it != paths.end(); ++it )
+	if ((m_showUserContent != nullptr) && (m_showFactoryContent != nullptr))
 	{
-		addItems( *it );
+		filterDirectories(m_directories, m_userDir,
+						  m_showUserContent->isChecked());
+		filterDirectories(m_directories, m_factoryDir,
+						  m_showFactoryContent->isChecked());
+	}
+
+	if (m_directories.length() > 0)
+	{
+		QStringList paths = m_directories.split( '*' );
+		for( QStringList::iterator it = paths.begin(); it != paths.end(); ++it )
+		{
+			addItems( *it );
+		}
 	}
 	expandItems(nullptr, expandedDirs);
 	m_filterEdit->setText( text );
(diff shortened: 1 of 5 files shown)


src/gui/FileBrowser.cpp
```diff
@@ -162,17 +196,52 @@ bool FileBrowser::filterItems( const QString & filter, QTreeWidgetItem * item )
...
+	if ((m_showUserContent != nullptr) && (m_showFactoryContent != nullptr))
 	{
-		addItems( *it );
+		filterDirectories(m_directories, m_userDir,
+						  m_showUserContent->isChecked());
+		filterDirectories(m_directories, m_factoryDir,
+						  m_showFactoryContent->isChecked());
+	}
+
+	if (m_directories.length() > 0)
```
IanCaio, line 238: More of a curiosity question, we didn't have this check before, so if `m_directories` was an empty string we would have an empty entry being added right? We didn't have it before because `m_directories` would never be empty or was it a bug?

src/gui/FileBrowser.cpp
```diff
@@ -162,17 +196,52 @@ bool FileBrowser::filterItems( const QString & filter, QTreeWidgetItem * item )
...
+	{
+		QStringList paths = m_directories.split( '*' );
+		for( QStringList::iterator it = paths.begin(); it != paths.end(); ++it )
+		{
+			addItems( *it );
```
IanCaio, line 243: I still didn't get to test the build (will download it later), but as far as the code goes, it all looks good to me. Both the logic and the code style of new additions (except for this little section here). I'll update my review to an approval as soon as I test it!

```suggestion
		QStringList paths = m_directories.split('*');
		for (QStringList::iterator it = paths.begin(); it != paths.end(); ++it)
		{
			addItems(*it);
```

src/gui/FileBrowser.cpp
thmueller64: I think the check wasn't there before because `m_directories` would never be empty. If `m_directories` is empty, the files in the folder where lmms was started from are listed. In my opinion, this is not the intended behavior.

Spekular:
Screenshot of the PR:

image

Testing results:

➕  The setting is independent per tab, filters correctly.
➖  Toggling is noticeably delayed on my configuration, setting doesn't persist between sessions or save with projects.

I don't think the short delay on toggle is a big issue, and it would matter even less if the setting was saved in LMMS's config or each project file. I would definitely prefer for the setting to persist between sessions, but this is an improvement either way so I'm not opposed to merging without it.

thmueller64:
I agree, the delay would not be so much of an issue if the state of the checkboxes would be saved for each project. I'll look into that, I assume this would not be that complicated. 
What do you think about stacking the checkboxes vertically instead of horizontally?

superpaik:
I don't have a delay issue in my end (windows 10), maybe because it's a "clean" installation without many "user" files. I agree it would be better to save it as an global setting (better than as a project setting)
Although the font used is in line with some other text in the app (like in "settings") It's not in line with the fonts used in that particular widget (I don't see why they have to be so small, a space problem?). I don't know if that's an issue at all, or worth solving.
On the other hand, having the two checkboxes uncheck doesn't have sense, doesn't it?, so I don't know if that's the best solution for that particular case. I've been thinking about an alternative but I only come up with a dropdown list, but I don't like it either. Maybe you guys can come up with something better (or let it be as it is, It's not that bad) It's just me being picky ;-)

Spekular:
> What do you think about stacking the checkboxes vertically instead of horizontally?

I prefer the way they are now as it leaves more vertical room for items in the sidebar. I do agree with superpaik that the font size could be bumped up a bit, and to me it looks like there's room for a few points increase without running out of space. I might also pull the search bar up a little closer, so the spacing matches the on between the search bar and list. (Note that I mean the perceived spacing, so the exact pixel amount might be different). Quick mockup:

image


As for alternatives to checkboxes, I suppose a radio button list with "user", "factory", and "both" is one. That gets pretty cramped though, at least if it's staying on one line: (pretend the checkboxes are radio buttons instead :P)

image

Personally though I think the existing checkboxes are fine. I don't imagine anyone would manage to uncheck both boxes without realizing what they're doing.

IanCaio (approved):
Code LGTM! I also tested and it works fine.

Some notes:
- I noticed the delay on the reload too, but here it was as short as 1s, expected since the whole tree is being rebuilt.
- I'd be alright with the checkboxes being kept as they are (both being toggled individually). I'm not sure we have an already made radio selector widget. It could also be done by having a callback connected to the checkbox toggle that checks the other checkbox's state. Not holding the approval because of that though, since even with the possibility of unchecking both checkboxes the behavior is very clear to the user.
- I agree that the state of the checkboxes could be saved per session, so I'm also leaving this as a suggestion. Think that could be done through `ConfigManager.cpp`, shouldn't be too difficult.

DomClark (commented):
I think the filtering could be somewhat simplified by working with an actual list, rather than juggling strings with separator characters. Make `FileBrowser::m_directories` a `QStringList`, and work with a copy of it, rather than the original, in `reloadTree`. Then the filtering can be as simple as
```cpp
if (m_showUserContent && !m_showUserContent->isChecked()) { directories.removeAll(m_userDir); }
if (m_showFactoryContent && !m_showFactoryContent->isChecked()) { directories.removeAll(m_factoryDir); }
```
Edit: or even just split `m_directories` into a `QStringList` before, not after, filtering.

src/gui/FileBrowser.cpp
```diff
@@ -65,17 +66,51 @@ enum TreeWidgetItemTypes
...
+{
+	auto filterWidget = new QWidget(contentParent());
+	filterWidget->setFixedHeight(15);
+	auto filterWidgetLayout = new QHBoxLayout(filterWidget);
+	filterWidgetLayout->setMargin(0);
+	filterWidgetLayout->setSpacing(0);
+
+	auto configCheckBox = [this, &filterWidgetLayout](QCheckBox* box)
+	{
+		box->setAccessibleName("FileBrowserFilterCheckBox");
```
DomClark, line 79: Is there a better way to identify the checkboxes for the stylesheet? `setAccessibleName` is designed to provide text for screenreaders, so I don't think it's the right choice here.

thmueller64: `QObject` has `setObjectName()` which would be more suitable. In the style sheet you could then write `QCheckBox[objectName="FileBrowserFilterCheckBox"] {...}`. Alternatively, `QCheckBox` could be subclassed, but this seems a bit overkill. What do you think?

IanCaio:
Just letting you know I just reviewed the recent changes that Dom proposed and they all look good.

I haven't tested the build but from what I've seen at the code you can still uncheck both checkboxes but if you do all the files will show up? If so, is that preferred over having the previous behavior of just filtering everything? I just mean it in the UX perspective, I know it's probably not the intention of the user to ever filter all folders, but if the checkboxes are allowed to be both unchecked I think it would be a bit counter-intuitive that they show all the files again. I'd rather keep it as previously until a better solution is thought of (could be in another PR as far as I'm concerned, since IMHO it's not that big deal that unchecking both removes all folders).

src/gui/FileBrowser.cpp
```diff
@@ -65,17 +66,51 @@ enum TreeWidgetItemTypes
...
+	filterWidget->setFixedHeight(15);
+	auto filterWidgetLayout = new QHBoxLayout(filterWidget);
+	filterWidgetLayout->setMargin(0);
+	filterWidgetLayout->setSpacing(0);
+
+	auto configCheckBox = [this, &filterWidgetLayout](QCheckBox* box)
+	{
+		box->setObjectName("FileBrowserFilterCheckBox");
+		box->setCheckState(Qt::Checked);
+		connect(box, SIGNAL(stateChanged(int)), this, SLOT(reloadTree()));
```
IanCaio, line 81: A suggestion for the unchecking both checkboxes issue. Code hasn't been tested but it should give an idea of what I mean.

```suggestion
		connect(box, SIGNAL(stateChanged(int)),
			this, [this](){
				if (!(m_showUserContent->isChecked() || m_showFactoryContent->isChecked()))
				{
					this->setChecked(true);
				}
				reloadTree();
			},
			Qt::DirectConnection
			);
```

thmueller64: Thanks for the review.
> I haven't tested the build but from what I've seen at the code you can still uncheck both checkboxes but if you do all the files will show up? If so, is that preferred over having the previous behavior of just filtering everything?

No it doesn't show all files, it shows an empty directory. The following check prevents that:
```
if (!paths.isEmpty())
{
	for (QStringList::iterator it = paths.begin(); it != paths.end(); ++it)
	{
		addItems(*it);
	}
}
```

IanCaio: But when you uncheck the last CheckBox, won't it call `reloadTree` again? Then I think since neither conditionals will be true (both checkboxes being unchecked) there will be no filtering of the string list.

IanCaio: Oh, completely ignore the last comments, I misread the conditional :man_facepalming: 
It does behave as before hiding both user and factory folders. That snippet from before could still be used if you want to force at least one to be shown. I'm alright with the current behavior though!

thmueller64: Thanks for the suggestion. I think it would be a bit surprising for the user for the following reason: Let's say that i have factory checked and user unchecked. Then if I want to have only user checked, I (could) uncheck factory first, just to be surprised that it is instantly checked again by the code above. So the only way would be to check user first and then to uncheck factory.

src/gui/FileBrowser.cpp
DomClark: I have to admit I've never written CSS for Qt. Does the standard descendant combinator work, i.e. `FileBrowser QCheckBox { ... }`? If so, I would go for that, otherwise `objectName` sounds good for now.

Veratil: > 
> 
> I have to admit I've never written CSS for Qt. Does the standard descendant combinator work, i.e. `FileBrowser QCheckBox { ... }`? If so, I would go for that, otherwise `objectName` sounds good for now.

Qt CSS is a near complete equivalent to normal CSS, so the descendant selector works like normal. 👍

thmueller64: Thanks @DomClark and @Veratil , I did not know about that. I added another commit, the descendant selector works perfectly fine.

DomClark (approved):

DomClark:
Merge?

src/gui/MainWindow.cpp
```diff
@@ -128,20 +128,26 @@ MainWindow::MainWindow() :
 					"*.mmp *.mmpz *.xml *.mid",
 							tr( "My Projects" ),
 					embed::getIconPixmap( "project_file" ).transformed( QTransform().rotate( 90 ) ),
-							splitter, false, true ) );
+							splitter, false, true,
+				confMgr->userProjectsDir(),
+				confMgr->factoryProjectsDir() ) );
```
Veratil, line 133:
```suggestion
				confMgr->factoryProjectsDir()));
```

IanCaio:
All good IMHO! Code looks good :+1:

src/gui/MainWindow.cpp
```diff
@@ -128,20 +128,26 @@ MainWindow::MainWindow() :
...
 	sideBar->appendTab( new FileBrowser(
 				confMgr->userPresetsDir() + "*" +
 				confMgr->factoryPresetsDir(),
 					"*.xpf *.cs.xml *.xiz *.lv2",
 					tr( "My Presets" ),
 					embed::getIconPixmap( "preset_file" ).transformed( QTransform().rotate( 90 ) ),
-							splitter , false, true  ) );
+							splitter , false, true,
+				confMgr->userPresetsDir(),
+				confMgr->factoryPresetsDir()) );
```
Veratil, line 150:
```suggestion
				confMgr->factoryPresetsDir()));
```

Veratil (commented):
Just some extra whitespace that should get fixed.

DomClark:
Whitespace concerns addressed; will merge tomorrow.

PR #5764, opened 2020-11-03
Stop sample previews on mouse release or browser focus loss

Spekular:
This fix/workaround closes #5736.

I'll make a PR once this is merged that allows samples to keep playing after mouse release (depending on the users settings), because it seems that stopping samples on focus loss is enough to prevent the crash*. In other words, all that remains is some work to add/save/read the option. However, since this would increase the scope of the PR and review times, I'd like to get this merged first so the crash is fixed and nightly builds can go up.

*If there turns out to be a race condition I think we can either add a stopPreview call to the stop button code or fix it in the mixer.

Issue #5736, opened 2020-10-26
Crash when previewing sample


zonkmachine:
Got a crash when previewing a sample when running lmms in gdb.
`-DCMAKE_BUILD_TYPE=Debug -DWANT_DEBUG_FPE=True`
lmms version - master, 7808e0b92fdb7f84180553fa534d66b2f2f041c6 with one unrelated commit on top.

#### Method to replicate
 * Start master, debug build in gdb.
 * Open  demo project loop.mmp [loop.mmp.zip](https://github.com/LMMS/lmms/files/5440854/loop.mmp.zip)
 * Press play in the Song Editor
 * Demo a sample from the side menu
 * Press paus in the Song Editor
 * Demo a sample from the side menu
 * Crash
#### Logs
<details>
  <summary>Click to expand</summary>
<pre>
Thread 1 "lmms" received signal SIGSEGV, Segmentation fault.
0x00005555557b305f in Mixer::removePlayHandle (this=0x555555c920c0, _ph=0x55555a4d34f0) at /home/zonkmachine/builds/lmms/src/core/Mixer.cpp:705
705		if( _ph->affinityMatters() &&
(gdb) bt
#0  0x00005555557b305f in Mixer::removePlayHandle(PlayHandle*) (this=0x555555c920c0, _ph=0x55555a4d34f0)
    at /home/zonkmachine/builds/lmms/src/core/Mixer.cpp:705
#1  0x000055555586a7f9 in FileBrowserTreeWidget::stopPreview() (this=0x5555598e4650) at /home/zonkmachine/builds/lmms/src/gui/FileBrowser.cpp:607
#2  0x000055555586a13b in FileBrowserTreeWidget::previewFileItem(FileItem*) (this=0x5555598e4650, file=0x555559911170)
    at /home/zonkmachine/builds/lmms/src/gui/FileBrowser.cpp:543
#3  0x000055555586a0c2 in FileBrowserTreeWidget::mousePressEvent(QMouseEvent*) (this=0x5555598e4650, me=0x7fffffffd3e0)
    at /home/zonkmachine/builds/lmms/src/gui/FileBrowser.cpp:532
#4  0x00007ffff7a972d1 in QWidget::event(QEvent*) (this=this@entry=0x5555598e4650, event=event@entry=0x7fffffffd3e0) at kernel/qwidget.cpp:8959
[...]

PR #5787, opened 2020-11-14
Add option to continue sidebar previews when mouse released

Spekular:
Adds a new option to the (incredibly crowded) General>UI section of our preferences that allows clicked samples in the sidebar to play to completion/interruption. With this setting enabled, sample previews will not stop on mouse release, but will still stop when:
- Another sidebar item gains focus (another preview starts, cursor moves to folder)
- The sidebar loses focus (another area is clicked on)
- The sample finishes playing

Thanks to changes in #5764 (stopping previews on focus loss, specifically), this should no longer cause the crash described in #5736. Please test this! Somewhat related: #5427

Milestoning for 1.3 since 1.2 had this option... ish.

### DIFF
--- a/include/SetupDialog.h
+++ b/include/SetupDialog.h
@@ -75,6 +75,7 @@ private slots:
 	void toggleCompactTrackButtons(bool enabled);
 	void toggleOneInstrumentTrackWindow(bool enabled);
 	void toggleSideBarOnRight(bool enabled);
+	void toggleLetPreviewsFinish(bool enabled);
 	void toggleSoloLegacyBehavior(bool enabled);
 	void toggleMMPZ(bool enabled);
 	void toggleDisableBackup(bool enabled);
@@ -133,6 +134,7 @@ private:
 	bool m_compactTrackButtons;
 	bool m_oneInstrumentTrackWindow;
 	bool m_sideBarOnRight;
+	bool m_letPreviewsFinish;
 	bool m_soloLegacyBehavior;
 	bool m_MMPZ;
 	bool m_disableBackup;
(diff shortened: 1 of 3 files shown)


superpaik:
It works perfectly as described. Only one minor thing. When you said (regarding scenarios to stop sample from playing): "Another sidebar item gains focus (another preview starts, **cursor moves to folder**)", does it means that if I click (or open/close) on a folder (not a sample) the current sample should stop? It doesn't. Personally I think that's a better behaviour than stop it, but I don't know if this is intended or a side-effect.
No crashes found (related to #5736 )
It saves and loads the configuration setting.

Spekular:
> When you said (regarding scenarios to stop sample from playing): "Another sidebar item gains focus (another preview starts, **cursor moves to folder**)", does it means that if I click (or open/close) on a folder (not a sample) the current sample should stop? It doesn't.

I was thinking about the arrow keys when I said that, so I forgot about the mouse. Thanks for pointing it out!

> Personally I think that's a better behaviour than stop it, but I don't know if this is intended or a side-effect.

It wasn't intended, but I'm not sure if I should change it or not. I'll test a bit in LMMS and other DAWs and see what feels natural.

Also something I realized yesterday and was planning to fix as soon as I got the time: preset previews will likely play forever with this setting on, so I need to restore the previously existing logic that ensured they always turn off on mouse-up.

Spekular:
Latest commit restores proper behavior for preset previews, so they stop when the mouse is released as they should. Nothing else should be different. 

Both Ableton and FL let previews continue when folders are opened/closed, and it doesn't feel bad or unintuitive to me, so I'm leaving that as is.

Veratil (approved):
Nothing stands out to me, it looks good. I haven't built and tested, but the logic is sound.

DomClark (approved):

IanCaio (approved):
Will later try to download the build and test, but as far as the code goes LGTM!

Spekular:
Quoting myself and IanCaio from Discord:
> do you still intend to test #5787 or should I go ahead and merge it?

> [..] I didn’t test it yet but I did review the code and it all looked good, since others have tested and had no issues it’s alright by me if you want to merge it now

Issue #5722, opened 2020-10-19
Undo/Redo of note edits via the menu is glitchy


zonkmachine:
When undoing/redoing note edits via the menu (under Edit) the action doesn't happen until you move the mouse back over the PianoRoll note window.

#### Steps to reproduce

* Add some notes in the Piano Roll
* Menu -> Edit -> undo
* wait...
* Move mouse over the notes and see the last one disappear

This affects the master branch and bisecting suggest that https://github.com/LMMS/lmms/pull/5499, 5d7e6720e1ba0c154041fd663801f7a7a0e16cb5 is where the problem started.

commit 5d7e6720e1ba0c154041fd663801f7a7a0e16cb5
Author: Luna Nooteboom <l.nooteboom@gmail.com>
Date:   Sun Jun 21 04:17:05 2020 +0200

    Automatically assign a midi input device to the selected track (#5499)

Bump @LNooteboom   :)

thmueller64:
Thank you for sorting out the relevant commit. Could you try this simple fix?

PR #5789, opened 2020-11-16
Fix glitch in undo/redo of note edits via the menu

thmueller64:
Addresses #5722.

### DIFF
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -3570,7 +3570,7 @@ void PianoRoll::focusInEvent( QFocusEvent * ev )
 		// Assign midi device
 		m_pattern->instrumentTrack()->autoAssignMidiDevice(true);
 	}
-    QWidget::focusInEvent(ev);
+	QWidget::focusInEvent(ev);
 }
 
 
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -3563,13 +3563,14 @@ void PianoRoll::focusOutEvent( QFocusEvent * )
 	update();
 }
 
-void PianoRoll::focusInEvent( QFocusEvent * )
+void PianoRoll::focusInEvent( QFocusEvent * ev )
 {
 	if ( hasValidPattern() )
 	{
 		// Assign midi device
 		m_pattern->instrumentTrack()->autoAssignMidiDevice(true);
 	}
+    QWidget::focusInEvent(ev);
 }
 
 


src/gui/editors/PianoRoll.cpp
```diff
@@ -3563,13 +3563,14 @@ void PianoRoll::focusOutEvent( QFocusEvent * )
...
 
-void PianoRoll::focusInEvent( QFocusEvent * )
+void PianoRoll::focusInEvent( QFocusEvent * ev )
 {
 	if ( hasValidPattern() )
 	{
 		// Assign midi device
 		m_pattern->instrumentTrack()->autoAssignMidiDevice(true);
 	}
+    QWidget::focusInEvent(ev);
```
zonkmachine, line 3573: You have 4 spaces in front of this line. Please change that to 1 tab.

zonkmachine:
I've tested this and it fixes #5722

Veratil:
@zonkmachine can you make sure this doesn't interfere with any other normal operation?

zonkmachine:
> @zonkmachine can you make sure this doesn't interfere with any other normal operation?

Yes, I need to test if the commit that caused this issue still works.

superpaik:
Undo via menu works well. Via hotkey too.
But there is something weird with redo. Via menu works. But via hotkeys there is a strange behaviour. Steps:
- put three notes on the piano roll.
- undo via menu (last note is deleted) OK
- redo via hotkey doesn't work. KO
- undo again via menu (second note is deleted) OK
- redo via hotkey works well (second note appears) redo via hotkey again (last note appears) OK

DomClark (approved):

thmueller64:
@superpaik : That's strange indeed. I can also reproduce this with the current master, so it appears not to be related to this PR. Maybe file another issue if you can also confirm this on master?

superpaik:
@thmueller64 Yes, it happens in the master as well, so it's not related to this PR.
I won't create another issue because there are few already related to undo/redo and the piano roll, and I've been told that this functionality (undo/redo) needs some kind of refactoring because of it's glitches.
thanks!

zonkmachine:
> > @zonkmachine can you make sure this doesn't interfere with any other normal operation?
> 
> Yes, I need to test if the commit that caused this issue still works.

Some testing. At a quick workthrough everything looks fine to me. I'm not sure I tested everything related to 5d7e6720e1ba0c154041fd663801f7a7a0e16cb5 . I don't have time to do a deeper check right now but I don't think that's necessary. I think this is fine to merge.

zonkmachine (approved):

zonkmachine:
I could replicate the issue that @superpaik describes but only once. It's definitely intermittent for me. 
Merging!

PR #5792, opened 2020-11-18
Fix missing icon of Stereo Matrix plugin

dj-pixus:
Most probably intruduced by 9f905bc, the default lmms logo was replaced by a garbage-coloured pixel on the Stereo Matrix effect. The problem was caused by the image not being in the relevant CMakeLists.txt file.


zonkmachine:
Nice, thanks! Issue and fix confirmed.

PR #5796, opened 2020-11-20
Fixes bug on SampleBuffer::decodeSampleSF

IanCaio:
As requested, this PR extracts the bug fix from #5791. More details about the bug on the mentioned PR page, but basically the
variable `sf_rr` was declared with the wrong type causing the conditional that checks for errors on the loading to return true even when there were no errors (only noticeable with DEBUG builds).

This renames the variable and uses the correct type.


PhysSong (approved):

IanCaio:
Being a very small bug fix PR (3 lines) and having one approval review, I'm thinking of merging this one tomorrow if there are no objections!

Spekular (approved):

PR #5801, opened 2020-11-21
Support multiple instrument subplugin categories

DomClark:
Currently, the instrument sidebar only supports two categories: LMMS and LV2. This pull request supports a category for any instrument exposing subplugins. This is primarily in preparation for VST3, but should also be useful for other formats in the future.

### DIFF
--- a/include/PluginBrowser.h
+++ b/include/PluginBrowser.h
@@ -33,7 +33,6 @@
 
 class QLineEdit;
 class QTreeWidget;
-class QTreeWidgetItem;
 
 
 class PluginBrowser : public SideBarWidget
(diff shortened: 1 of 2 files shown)


JohannesLorenz (approved):
Code reviewed. LGTM.

I also tested all I could, but I currently only have Lv2 and LMMS. Should be enough IMO.

DomClark:
This PR doesn't add any new instrument subplugins yet; it only supports adding them in the future. If it works with LMMS and LV2 still, then it's working properly.

Will merge in a day if there are no objections.

superpaik:
I understand LV2 are not available under windows. It works well. It doesn't crash or malfunction under windows (10)

PR #5803, opened 2020-11-24
FifoBuffer adjustments

M374LX:
The template in ``fifoBuffer`` is unnecessary because only one type is used (``surroundSampleFrame *``). The PR also fixes the coding conventions for that header file.

Additionally, it fixes the typoed method name ``Song::setLoadOnLauch()``.

### DIFF
--- a/include/FifoBuffer.h
+++ b/include/FifoBuffer.h
@@ -27,9 +27,8 @@
 
 #include <QtCore/QSemaphore>
 
-#include "lmms_basics.h"
 
-
+template<typename T>
 class FifoBuffer
 {
 public:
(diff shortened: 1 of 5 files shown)


Veratil:
I like the updates, but I don't think removing the template is necessary even if we're only using it with a single type.

JohannesLorenz:
@M374LX Can you please think about the fix from @Veratil (and fix the conflict by a merge or rebase)?

M374LX:
Rebase done.

Should I keep the removed template? If so, this will become just another reformatting PR.

Veratil:
> Should I keep the removed template? If so, this will become just another reformatting PR.

I think keeping is better for now.

PhysSong:
> Additionally, it fixes the typoed method name `Song::setLoadOnLauch()`.

I think that shouldn't be in a PR which changes `Mixer`-related stuff?

M374LX:
@PhysSong It is just a typo I noticed while preparing this PR.

JohannesLorenz:
This should be merged before reorg, so I'm reviewing this now.

JohannesLorenz (approved):
LGTM

Issue #5799, opened 2020-11-21
Bit Invader - Corrupt data when saving with connected controller


zonkmachine:
The Bit Invader wave gets corrupted on saving with a controller connected to the 'Length' knob.
Probably same issue as first noted here: https://github.com/LMMS/lmms/pull/4705#issuecomment-459977039
Comment from @PhysSong in #4705 
>I dumped the saved data. For some reason the decoded data contain Base64 data. I think the data were corrupted while converting to Base64 but not sure why.

The Bit Invader has also produces bad data (nan/inf). Unsure if that is directly connected to this issue.

#### Steps to reproduce

 * Add a Bit Invader and an LFO Controller.
 * Connect the `Length` knob to the controller.
 * Save and Reload multiple times and watch the graph of the Bit Invader get more and more destroyed in steps.

It also looks like this is related to when in the controller cycle you do a save. If the Length knob is at it's minimum you get the issue. If it's at it's maximum then I haven't seen any issues.

#### Screenshot
Created on lmms-1.2.1
1. Original correct 2. One save/load (similar to issue described in https://github.com/LMMS/lmms/issues/3797) 3. wave partly zeroed. 4. partly zeroed with extra glitches.
![bitinvader](https://user-images.githubusercontent.com/6368949/99875405-ba6fcf80-2bef-11eb-98d7-4b720202643d.png)

#### Affected LMMS versions
Shown on lmms-1.2 and master (87875a18e321ffe4278fd693d3eadcd5802b86ec).

zonkmachine:
Added to the description that it also looks like this is related to when in the controller cycle you do a save. If the Length knob is at it's minimum you get the issue. If it's at it's maximum then I haven't seen any issues so far.

zonkmachine:
BitInvader is created with a default sinewave, length 200, then whatever saved waveform is loaded and writes over the default one. If the saved wavefile is less than 200 samples long there will be a trailing part of the sine left in there. This is seen in graph 2-4 in the issue description. A quick fix for this could be to not load a default sinewave or the rest of the sample shape be written  as 0 on load.

https://github.com/LMMS/lmms/blob/8d4bcd71054f15addb06fcc8cfea7acc4d48d17f/plugins/bit_invader/bit_invader.cpp#L149


The other glitches seem to be fixed by changing:

https://github.com/LMMS/lmms/blob/892aec318d0a78442dbf62d219e1137418d72b45/plugins/bit_invader/bit_invader.cpp#L181

to:
[...]

PR #5805, opened 2020-11-24
Bitinvader - Fix saving with automation and division by 0

zonkmachine:
Fixes https://github.com/LMMS/lmms/issues/5799 and division by 0 detected with fpe debug flags.

Save and Load complete wavetable or you will face issues when the project is automating the Length knob. Hard coded wavetable length is defined in bit_invader.cpp. Maybe it should be in bit_invader.h but this looks fine and other plugins did it similarly.

The normalize function may divide with zero so we solve this the same way as the normalize function in Graph.cpp by starting off with a small value for `max`. 

#### Testing
Use the steps to reproduce the issue from https://github.com/LMMS/lmms/issues/5799 and especially try saving the project with `Ctrl+s` while the Length knob is at it's smallest setting when it's being controlled with an LFO.
 * Add a Bit Invader and an LFO Controller.
 * Connect the `Length` knob to the controller.
 * Save and Reload multiple times and watch the graph of the Bit Invader get more and more destroyed in steps.

### DIFF
--- a/plugins/bit_invader/bit_invader.cpp
+++ b/plugins/bit_invader/bit_invader.cpp
@@ -44,6 +44,8 @@
 
 #include "plugin_export.h"
 
+#define WAVETABLE_SIZE 200
+
 extern "C"
 {
 
@@ -72,8 +74,8 @@ bSynth::bSynth( float * _shape, NotePlayHandle * _nph, bool _interpolation,
 	sample_rate( _sample_rate ),
 	interpolation( _interpolation)
 {
-	sample_shape = new float[200];
-	for (int i=0; i < 200; ++i)
+	sample_shape = new float[WAVETABLE_SIZE];
+	for (int i=0; i < WAVETABLE_SIZE; ++i)
 	{
 		sample_shape[i] = _shape[i] * _factor;
 	}
@@ -138,8 +140,8 @@ sample_t bSynth::nextStringSample( float sample_length )
 
 bitInvader::bitInvader( InstrumentTrack * _instrument_track ) :
 	Instrument( _instrument_track, &bitinvader_plugin_descriptor ),
-	m_sampleLength( 128, 4, 200, 1, this, tr( "Sample length" ) ),
-	m_graph( -1.0f, 1.0f, 200, this ),
+	m_sampleLength( 128, 4, WAVETABLE_SIZE, 1, this, tr( "Sample length" ) ),
+	m_graph( -1.0f, 1.0f, WAVETABLE_SIZE, this ),
 	m_interpolation( false, this ),
 	m_normalize( false, this )
 {
--- a/plugins/bit_invader/bit_invader.cpp
+++ b/plugins/bit_invader/bit_invader.cpp
@@ -197,10 +197,7 @@ void bitInvader::saveSettings( QDomDocument & _doc, QDomElement & _this )
 void bitInvader::loadSettings( const QDomElement & _this )
 {
 	// Clear wavetable before loading a new
-	for (int i = 0; i < WAVETABLE_SIZE; i++)
-	{
-		m_graph.setSampleAt(i, 0.f);
-	}
+	m_graph.clear();
 
 	// Load sample length
 	m_sampleLength.loadSettings( _this, "sampleLength" );
--- a/plugins/bit_invader/bit_invader.cpp
+++ b/plugins/bit_invader/bit_invader.cpp
@@ -44,7 +44,7 @@
 
 #include "plugin_export.h"
 
-#define WAVETABLE_SIZE 200
+static const size_t wavetableSize = 200;
 
 extern "C"
 {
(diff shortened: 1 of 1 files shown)


zonkmachine:
Alternatively start `loadSettings(` with clearing out the wavetable to do away with the multiple calls to `m_graph.setLength()` which maybe isn't very clear.

```
	// Clear wavetable before loading a new
	for (int i = 0; i < WAVETABLE_SIZE; i++)
	{
		m_graph.setSampleAt( i, 0.f );
	}
```

plugins/bit_invader/bit_invader.cpp
```diff
@@ -194,6 +196,12 @@ void bitInvader::saveSettings( QDomDocument & _doc, QDomElement & _this )
 
 void bitInvader::loadSettings( const QDomElement & _this )
 {
+	// Clear wavetable before loading a new
```
Spekular, line 199: From what I can tell, this loop be replaced by `m_graph->model()->clear();`. Either that or `clearInvisible()` just before the LED initialization, at which point the graph length should be updated. The first option seems simpler to me, but all the `xWaveClicked()` functions use `clearInvisible()` so maybe there's a reason to do it that way.

Spekular (commented):
LGTM aside from one potential simplification.

plugins/bit_invader/bit_invader.cpp
zonkmachine: Cool, I'll look into it! Thanks!

plugins/bit_invader/bit_invader.cpp
```diff
@@ -240,7 +248,7 @@ void bitInvader::samplesChanged( int _begin, int _end )
 void bitInvader::normalize()
 {
 	// analyze
-	float max = 0;
+	float max = 0.0001f;
```
thmueller64, line 248: This is more of a general question: Why don't we use the machine epsilon `std::numeric_limits<float>::eps()` in places where we want to prevent division by zero? In my opinion this would be a cleaner solution to hardcoding a small value like in `Graph.cpp`s `normalize()` that you mention.

plugins/bit_invader/bit_invader.cpp
zonkmachine: Fixed!

thmueller64 (commented):
Added some small remarks.

Spekular (approved):

plugins/bit_invader/bit_invader.cpp
```diff
@@ -44,6 +44,8 @@
 
 #include "plugin_export.h"
 
+#define WAVETABLE_SIZE 200
```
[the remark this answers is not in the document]
zonkmachine: Fixed!

plugins/bit_invader/bit_invader.cpp
zonkmachine: Ok. Fixed here with	`std::numeric_limits<float>::epsilon();`

plugins/bit_invader/bit_invader.cpp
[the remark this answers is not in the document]
thmueller64: Looking at this again, it should be safer and better to use `static const int`. Sorry!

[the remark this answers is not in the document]
zonkmachine: OK, fixed... :)

zonkmachine:
Reviewed and tested. I think we're done here.  ;)

zonkmachine:
Merge?

DomClark:
There's something I'd like to check first if that's alright, but I'm a little busy right now. Will try to get round to it before the end of the day.

zonkmachine:
I'm a bit unsure in what order things happen when you save. for instance, I think changing `m_sampleLength` in the middle of a save can not lead to any unvanted behaviours but I'm not completely sure.

DomClark:
While this now saves the whole wavetable, it still doesn't load the whole wavetable. `graphModel::setSamples` only sets samples up to the current length of the wavetable, which is not necessarily the maximum length at the time of loading.
https://github.com/LMMS/lmms/blob/8cc9d8e30397c6c8d791abecd837dade192dd56c/plugins/bit_invader/bit_invader.cpp#L212-L213
`sampleLength` is the current value of the length model, which may have been smaller when saved, or may be connected to a controller and forced to a lower value. I think the `size` parameter from `base64::decode` should be used to set the length before loading. Something like this perhaps?
```cpp
int size = 0;
char* dst = 0;
base64::decode(_this.attribute( "sampleShape"), &dst, &size);

m_graph.setLength(size / sizeof(float));
m_graph.setSamples(reinterpret_cast<float*>(dst));
m_graph.setLength(sampleLength);
```

zonkmachine:
Why not just set it to wavetableSize then? I think one of my first iterations of this fix looked something like this.

```
m_graph.setLength( wavetableSize );
m_graph.setSamples( (float*) dst );
m_graph.setLength( sampleLength );
```

DomClark:
> Why not just set it to `wavetableSize` then?

The base64 data may be smaller than the wavetable size, if loaded from an older project or a tampered project. This would read `dst` out of bounds.

zonkmachine:
> The base64 data may be smaller than the wavetable size

Yes, this is cool! I tried it with bumping the wavetable size to 256 and it worked fine loading older projects. I may open an issue about that. Wavetable synths on the market have sample sizes that are multiples of 2 and with a sample length of 256 we could load for instance Doepfer A112 samples into the Bit Invader.

zonkmachine:
Now merge?

DomClark (approved):

PR #5808, opened 2020-11-28
Remove instrument track window caching to fix related bugs

### LINKED ISSUE #3623: Bugs with toggling One Instrument Track Window Mode
I, along with one other person, have found bugs when toggling this feature.

To reproduce my own issue regarding the feature, perform the following steps (may be dependent on other fundamental data which i was unable to gather at the time):
Open LMMS with the "One instrument track window mode" setting on (located in the first tab of the Settings window)
Open an instrument track window
Turn the aforementioned setting off
Attempt to open a different instrument track window
LMMS will then crash

@tresf has also observed bugs with this feature.

Operating system: Windows 10
LMMS Version: 1.2.0 rc3

### LINKED ISSUE #5743: Crash when Carla Instrument is removed
Removing a Carla track may cause LMMS to segfault.  This was reproducible on Windows 10 using the mingw 64-bit build as well as the MSVC build.

Some notes from #5713 which originally identified this bug.

@PhysSong wrote:

> Reproduced with Carla 2.2 and MSVC build of LMMS. It seems like Carla crashed in `CarlaEngine::ProtectedData::deletePluginsAsNeeded()` which is indirectly called from `CarlaInstrument::timerEvent()`. Inspecting memory and registers suggests there might be some use after free conditions.
2.1.1 Also crashes in `CarlaBackend::CarlaEngineNative::uiIdle()`, but less frequently.
> ...
>  I found that it's related to the (improperly implemented) `InstrumentTrackWindow` caching which causes a temporary `CarlaInstrumentView` to be instantiated after the existing one is deleted. That will be deleted after `CarlaInstrument`, and timer events may kick in between those two events. That results in `CarlaInstrumentView::timerEvent()` using a deleted handle.

Tagging for `windows`, however the bug may exist on other platforms.

PhysSong:
Fixes #3623, fixes #5743.

### DIFF
--- a/include/InstrumentTrack.h
+++ b/include/InstrumentTrack.h
@@ -324,10 +324,6 @@ public:
 		return m_midiMenu;
 	}
 
-	void freeInstrumentTrackWindow();
-
-	static void cleanupWindowCache();
-
 	// Create a menu for assigning/creating channels for this track
 	QMenu * createFxMenu( QString title, QString newFxLabel ) override;
 
@@ -349,11 +345,11 @@ private slots:
 	void assignFxLine( int channelIndex );
 	void createFxLine();
 
+	void handleConfigChange(QString cls, QString attr, QString value);
+
 
 private:
 	InstrumentTrackWindow * m_window;
-
-	static QQueue<InstrumentTrackWindow *> s_windowCache;
 
 	// widgets in track-settings-widget
 	TrackLabelButton * m_tlb;
@@ -482,6 +478,7 @@ private:
 	PianoView * m_pianoView;
 
 	friend class InstrumentView;
+	friend class InstrumentTrackView;
 
 } ;
 
--- a/include/InstrumentTrack.h
+++ b/include/InstrumentTrack.h
@@ -42,7 +42,6 @@
 
 
 class QLineEdit;
-template<class T> class QQueue;
 class InstrumentFunctionArpeggioView;
 class InstrumentFunctionNoteStackingView;
 class EffectRackView;
--- a/src/tracks/InstrumentTrack.cpp
+++ b/src/tracks/InstrumentTrack.cpp
@@ -970,9 +970,8 @@ InstrumentTrackView::InstrumentTrackView( InstrumentTrack * _it, TrackContainerV
 	connect( _it, SIGNAL( nameChanged() ),
 			m_tlb, SLOT( update() ) );
 
-	connect( ConfigManager::inst(), SIGNAL( valueChanged( QString, QString, QString ) ),
-			this, SLOT( handleConfigChange( QString, QString, QString ) ),
-			Qt::QueuedConnection );
+	connect(ConfigManager::inst(), SIGNAL(valueChanged(QString, QString, QString)),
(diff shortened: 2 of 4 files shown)


src/tracks/InstrumentTrack.cpp
```diff
@@ -1547,11 +1488,9 @@ InstrumentTrackWindow::InstrumentTrackWindow( InstrumentTrackView * _itv ) :
 
 InstrumentTrackWindow::~InstrumentTrackWindow()
 {
-	InstrumentTrackView::s_windowCache.removeAll( this );
-
 	delete m_instrumentView;
 
-	if( gui->mainWindow()->workspace() )
+	if (parentWidget())
```
PhysSong, line 1493: This is to prevent crashes on closing.

JohannesLorenz:
Give me some time and I'll review this PR.

src/tracks/InstrumentTrack.cpp
```diff
@@ -975,6 +971,10 @@ InstrumentTrackView::InstrumentTrackView( InstrumentTrack * _it, TrackContainerV
 	connect( _it, SIGNAL( nameChanged() ),
 			m_tlb, SLOT( update() ) );
 
+	connect( ConfigManager::inst(), SIGNAL( valueChanged( QString, QString, QString ) ),
+			this, SLOT( handleConfigChange( QString, QString, QString ) ),
+			Qt::QueuedConnection );
```
JohannesLorenz, line 976: Just curious: Why is this better than using `Qt::AutoConnection`?

src/tracks/InstrumentTrack.cpp
```diff
@@ -1066,7 +1066,8 @@ InstrumentTrackView::InstrumentTrackView( InstrumentTrack * _it, TrackContainerV
 
 InstrumentTrackView::~InstrumentTrackView()
 {
-	freeInstrumentTrackWindow();
+	delete m_window;
```
JohannesLorenz, line 1069: What about using `unique_ptr` instead?

JohannesLorenz (approved):
Only 3 very minor comments.

I also tested everything except the Carla issue #5743 .

LGTM.

src/tracks/InstrumentTrack.cpp
PhysSong: I think I copied this from some commented VST code. Will check this tomorrow.

src/tracks/InstrumentTrack.cpp
PhysSong: I think it's out of scope, but it's doable.

include/InstrumentTrack.h
```diff
@@ -349,12 +345,12 @@ private slots:
 	void assignFxLine( int channelIndex );
 	void createFxLine();
 
+	void handleConfigChange(QString cls, QString attr, QString value);
+
 
 private:
 	InstrumentTrackWindow * m_window;
 
-	static QQueue<InstrumentTrackWindow *> s_windowCache;
```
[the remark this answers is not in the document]
PhysSong: Nice catch.

PR #5813, opened 2020-11-30
Fix crash on OGG export with Qt >= 5.10

PhysSong:
Fix https://github.com/LMMS/lmms/pull/5619#discussion_r532348195.
Thanks to `@Rdrpenguin#0256` on Discord for reporting.


rdrpenguin04:
Tested locally, seems to be working!

zonkmachine (approved):
Rock solid. I thought qrandomgenerator would be seeded with a default 1 if you didn't specify a seed and therefore spit out the same values when you restart lmms. QRandomGenerator::global(), not so much. I've tested this and confirmed the randomness of the exported files serial number and it works just fine. Merge.

PR #5814, opened 2020-11-30
Fix export when rendering looped section multiple times

DomClark:
Apparently I don't understand how the "export looped section n times" feature is meant to work. Fortunately PhysSong does.

Applies PhysSong's fix from https://github.com/LMMS/lmms/pull/5723#discussion_r532386623. See https://github.com/LMMS/lmms/pull/5723#discussion_r532386747 for details.

Thanks to `@Rdrpenguin#0256` on Discord for reporting this bug.

### DIFF
--- a/src/core/Song.cpp
+++ b/src/core/Song.cpp
@@ -295,7 +295,7 @@ void Song::processNextBuffer()
 			}
 
 			// Handle loop points, and inform VST plugins of the loop status
-			if (loopEnabled || m_loopRenderRemaining > 1)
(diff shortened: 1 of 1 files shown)


rdrpenguin04:
This seems to fix one part of the problem: the lead-out now exists. However, anything before the loop is still not played, and the looped section still isn't looped.

rdrpenguin04:
NVM, forgot to `ninja install`.

It works!

PhysSong (approved):

Issue #5809, opened 2020-11-28
SDL startup segfault


M374LX:
Sometimes, LMMS (master branch) segfaults at startup.

**Update:** after building SDL from source, the backtrace gave more useful information.

```SDLAudioC3``` suggests it is related to SDL. The "C" means it is a capture thread: https://github.com/SDL-mirror/SDL/blob/81117f4052b8ae50a1e49fe5c221e05433ac77ac/src/audio/SDL_audio.c#L1449

```
Thread 17 "SDLAudioC3" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffbd69d700 (LWP 26426)]
0x0000000000000000 in ?? ()
(gdb) bt full
#0  0x0000000000000000 in ?? ()
No symbol table info available.
#1  0x00007ffff7741cfa in SDL_DBus_Init () at /home/ax/src/SDL2-2.0.12/src/core/linux/SDL_dbus.c:129
        err = {name = 0x0, message = 0x0, dummy1 = 1, dummy2 = 0, dummy3 = 0, dummy4 = 0, dummy5 = 1, padding1 = 0xa8087600}
        is_dbus_available = SDL_FALSE
#2  0x00007ffff77422e1 in SDL_DBus_Init () at /home/ax/src/SDL2-2.0.12/src/core/linux/SDL_dbus.c:168
        is_dbus_available = SDL_FALSE
        err = {name = <optimized out>, message = <optimized out>, dummy1 = <optimized out>, dummy2 = <optimized out>, dummy3 = <optimized out>, dummy4 = <optimized out>, dummy5 = <optimized out>, padding1 = <optimized out>}
#3  SDL_DBus_GetContext () at /home/ax/src/SDL2-2.0.12/src/core/linux/SDL_dbus.c:170
No locals.
#4  0x00007ffff774c069 in rtkit_setpriority (nice_level=-10, thread=26426) at /home/ax/src/SDL2-2.0.12/src/core/linux/SDL_threadprio.c:62
        ui64 = 26426
        si32 = -10
        dbus = <optimized out>
#5  SDL_LinuxSetThreadPriority_REAL (threadID=<optimized out>, priority=-10) at /home/ax/src/SDL2-2.0.12/src/core/linux/SDL_threadprio.c:105
No locals.
[...]

PR #5815, opened 2020-11-30
Workaround for SDL race condition

M374LX:
Workaround for #5809.

### DIFF
--- a/src/core/audio/AudioSdl.cpp
+++ b/src/core/audio/AudioSdl.cpp
@@ -99,6 +99,9 @@ AudioSdl::AudioSdl( bool & _success_ful, Mixer*  _mixer ) :
 	_success_ful = true;
 
 #ifdef LMMS_HAVE_SDL2
+	// Workaround for a race condition that causes SDL to segfault
+	SDL_Delay(250);
+
 	m_inputAudioHandle = m_audioHandle;
 	m_inputAudioHandle.callback = sdlInputAudioCallback;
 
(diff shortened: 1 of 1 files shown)


src/core/audio/AudioSdl.cpp
```diff
@@ -99,6 +99,9 @@ AudioSdl::AudioSdl( bool & _success_ful, Mixer*  _mixer ) :
 	_success_ful = true;
 
 #ifdef LMMS_HAVE_SDL2
+	// Workaround for a race condition that causes SDL to segfault
+	SDL_Delay(250);
```
PhysSong, line 103: Is 250ms a really good time interval? I'm not sure, but you can try lowering this and try to reproduce the crash.

src/core/audio/AudioSdl.cpp
```diff
@@ -99,6 +99,9 @@ AudioSdl::AudioSdl( bool & _success_ful, Mixer*  _mixer ) :
 	_success_ful = true;
 
 #ifdef LMMS_HAVE_SDL2
+	// Workaround for a race condition that causes SDL to segfault
```
JohannesLorenz, line 102: Using a delay to fix a race condition sounds generally strange. But if it's the only option...

Can you give any source where you found the solution? Or is this just fixed out of your mind?

M374LX: The idea of adding a delay came out of my mind.

Audio devices are implemented as threads in SDL. Consequently, inserting a delay in theory gives the the playback device's thread time to "finish starting" before the capture device (and thread) is created, which prevents a race condition.

JohannesLorenz: Well, if it improves things, why not...

JohannesLorenz:
I'm going to run some tests now, and then we can get this merged.

JohannesLorenz:
Thanks for the PR! The patch seems to cause no regression, I merged it.

PR #5816, opened 2020-12-01
Ensure file opened successfully when loading sample

### LINKED ISSUE #5213: Assertion failure loading missing sample in MSVC debug build
When loading a sample from a file, the file descriptor is passed to `sf_open_fd` without checking whether the file was opened successfully. On Windows, `sf_open_fd` then passes the file descriptor to `_get_osfhandle` to get the native file handle for the file descriptor. However, `_get_osfhandle` has an assertion to check the validity of the provided file descriptor, which is active in debug versions of the CRT (only used by MSVC debug builds, AFAIK) and is triggered when, for example, the sample doesn't exist. To avoid this, I think we should check that `QFile::open` returns `true` before calling `sf_open_fd`.

@Reflexe It looks like, as part of the recording work, you're refactoring `SampleBuffer`, which is where this function is used. Could you take care of this while you're at it?

DomClark:
Fixes #5213; see that issue for an explanation. Testing to verify that the bug is fixed requires building locally using MSVC, but it may be worth checking on other platforms that the changes don't break anything else.

### DIFF
--- a/src/core/SampleBuffer.cpp
+++ b/src/core/SampleBuffer.cpp
@@ -196,11 +196,10 @@ void SampleBuffer::update( bool _keep_settings )
 		{
 			// Use QFile to handle unicode file names on Windows
 			QFile f(file);
-			f.open(QIODevice::ReadOnly);
 			SNDFILE * snd_file;
 			SF_INFO sf_info;
 			sf_info.format = 0;
-			if( ( snd_file = sf_open_fd( f.handle(), SFM_READ, &sf_info, false ) ) != NULL )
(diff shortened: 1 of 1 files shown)


JohannesLorenz:
Code is OK!

PhysSong (approved):

DomClark:
This change is probably difficult for most people to test usefully, so since the code has been reviewed and approved, I will merge it later today.

IanCaio (approved):
Code LGTM too! I was wondering whether the file open error and the libsndfile open error should be handled separately, but I don't think it's needed. It's unlikely we will need that information and if we do need I think knowing it's between `f.open` and `sf_open_fd` is good enough.

PR #5819, opened 2020-12-03
rpmalloc - no asserts with fpe debug

zonkmachine:
The rpmalloc asserts are too noisy when using the floating point
exception debugging flags. Turning them off with fpe but they are
still on with normal debug builds.

Edit: You can't test this PR with the demo builds as it only concerns debug builds.

### DIFF
--- a/src/3rdparty/rpmalloc/CMakeLists.txt
+++ b/src/3rdparty/rpmalloc/CMakeLists.txt
@@ -35,14 +35,8 @@ if (CMAKE_BUILD_TYPE STREQUAL "Debug")
 	else ()
 		set(ENABLE_VALIDATE_ARGS ON)
 	endif()
-	# The rpmalloc asserts are too noisy when debugging with option WANT_DEBUG_FPE.
-	if(WANT_DEBUG_FPE)
-		set(ENABLE_ASSERTS 0)
-	else()
-		set(ENABLE_ASSERTS 1)
-	endif()
 	target_compile_definitions(rpmalloc
-		PRIVATE -DENABLE_ASSERTS=${ENABLE_ASSERTS} -DENABLE_VALIDATE_ARGS=${ENABLE_VALIDATE_ARGS}
+		PRIVATE -DENABLE_ASSERTS=0 -DENABLE_VALIDATE_ARGS=${ENABLE_VALIDATE_ARGS}
 	)
 endif()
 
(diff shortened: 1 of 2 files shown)


JohannesLorenz:
This raises multiple questions:

1. What has malloc to do with floats/doubles? A grep in rpmalloc for both of them returns nothing. So what does FPE report here? That probably means that FPE is also used for integers?
2. Even then, if we have integer overflows or division by zero, shouldn't that be reported to the rpmalloc team?
3. Why do more asserts happen if FPE is enabled? I thought FPE would not change the program logic, it would only react on SIGFPE?

Any ideas?

zonkmachine:
> Any ideas?

We were left with a message here. Maybe we can clean up something and that would make rpmalloc happy?

https://github.com/LMMS/lmms/blob/d3cd704396d9d16417c8ec18e3c35a63c9a3c699/src/core/main.cpp#L96


.


I'm not really happy with this PR. I have this urge to just instead turn off rpmalloc by setting it to 0 and also commenting out FP_UNDERFLOW at the same time. That calms things down. The underflow catches are correct but they pretty much just tells us that the cpu is hit with denormals and is going to work a bit slower. It's not critical and slows down development.

zonkmachine:
@JohannesLorenz Right. I just confused things with the last two commits. I think I'll revert those here and push this to master directly.

```diff
diff --git a/src/3rdparty/rpmalloc/CMakeLists.txt b/src/3rdparty/rpmalloc/CMakeLists.txt
index 047c32678..601d2f1eb 100644
--- a/src/3rdparty/rpmalloc/CMakeLists.txt
+++ b/src/3rdparty/rpmalloc/CMakeLists.txt
@@ -36,7 +36,7 @@ if (CMAKE_BUILD_TYPE STREQUAL "Debug")
                set(ENABLE_VALIDATE_ARGS ON)
        endif()
        target_compile_definitions(rpmalloc
-               PRIVATE -DENABLE_ASSERTS=1 -DENABLE_VALIDATE_ARGS=${ENABLE_VALIDATE_ARGS}
+               PRIVATE -DENABLE_ASSERTS=0 -DENABLE_VALIDATE_ARGS=${ENABLE_VALIDATE_ARGS}
        )
 endif()
 
diff --git a/src/core/main.cpp b/src/core/main.cpp
index 08599362e..e602bd482 100644
--- a/src/core/main.cpp
+++ b/src/core/main.cpp
@@ -254,8 +254,8 @@ int main( int argc, char * * argv )
        // Enable exceptions for certain floating point results
        feenableexcept( FE_INVALID   |
                        FE_DIVBYZERO |
-                       FE_OVERFLOW  |
-                       FE_UNDERFLOW);
+                       FE_OVERFLOW  /*|
+                       FE_UNDERFLOW*/);
 

```


Cool with that? Then I'll probably close this one and open an issue instead.

JohannesLorenz:
@zonkmachine I'll look a bit closer on it this weekend.

zonkmachine:
Closing this. Will open an issue instead.

PR #5824, opened 2020-12-05
Fixes bug where clicking on the Activity Indicator doesn't play a note

IanCaio:
After merging of #5581 a bug was unveiled, where an ambiguity on the arguments types of calls to the `MidiEvent` constructor would result in ADL picking the wrong option.

Discussing with others in Discord about approaches to fix the issue, @DomClark suggested using an `enum class` instead of a boolean on the last argument, replacing it from `ignoreOnExport` to `source` (representing the source of a MidiEvent). One of the arguments type from the second constructor was also changed from `int` to `std::size_t` to better represent it.

Those changes alone fixes the Activity Indicator. It's possible though that adding `static_casts` to the constructor calls could still be needed. For now I'm submitting the changes for further review and discussion.


DomClark (approved):
Looks good, and fixes the bug. There are a couple more things that could be done, but I'm not sure if it's worth doing them here:
* `m_data.m_sysExDataLen` should be changed to a `std::size_t`, to match `dataLen` which is stored in it. However, it seems to be unused, so it doesn't matter too much right now.
* Should all `MidiEvent`s be constructed with the correct source? Currently everything is marked as external, except controller changes. Again, this is fine for now, since most messages go to `processOutEvent` where the source doesn't matter, but it might be worth correcting at some point. (I would probably make the default source `Internal`, and specify `External` in the MIDI backends, to keep the core code cleaner.)

Issue #5827, opened 2020-12-05
Carla instruments - Hard to grab/drag


zonkmachine:
### Bug Summary

It can be hard to grab an open carla instrument in lmms and it feels sometimes 'frozen'.

#### Steps to reproduce

Open a Carla instrument window and a Triple Oscillator, and move them next to each other on a free space with no other windows. Place the cursor, now a hand icon, above the Triple Oscillator and move it slowly down until it is above the instrument. The hand cursor changes to an arrow and you can now grab the instrument by left clicking it and dragging. Now do the same with the Carla instrument. It will not hand over to the arrow and let you grab it. You can only do this when moving over the top border from another direction.

![grabcarla](https://user-images.githubusercontent.com/6368949/101267029-df6a5380-3754-11eb-94e1-e9e4bbef1dc6.png)

#### Affected LMMS versions

stable-1.2 and master

zonkmachine:
Same issue as https://github.com/LMMS/lmms/issues/5828

PR #5829, opened 2020-12-06
Fix too small height of the carla instrument window

### LINKED ISSUE #5828: Carla instruments - Add effect button missing and related size issues
### Bug Summary

Carla Rack and Carla Patchbay miss the "Add effect" button in the FX rack. They also different in size to other plugins and the lower parts of the other tab windows are cut off.

#### Steps to reproduce

Open one of the Carla instruments and drag them above another instrument. Now go through the tabs one by one.

#### Affected LMMS versions

master probably since https://github.com/LMMS/lmms/pull/4906

carla2

PhysSong:
Fixes #5828.

### DIFF
--- a/src/tracks/InstrumentTrack.cpp
+++ b/src/tracks/InstrumentTrack.cpp
@@ -1551,13 +1551,13 @@ InstrumentTrackWindow::InstrumentTrackWindow( InstrumentTrackView * _itv ) :
 	vlayout->addWidget( m_pianoView );
 	setModel( _itv->model() );
 
-	updateInstrumentView();
-
 	QMdiSubWindow* subWin = gui->mainWindow()->addWindowedWidget( this );
 	Qt::WindowFlags flags = subWin->windowFlags();
 	flags |= Qt::MSWindowsFixedSizeDialogHint;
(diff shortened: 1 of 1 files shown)


zonkmachine:
It also affects the behavior of https://github.com/LMMS/lmms/issues/5827 so they seem related after all.

PhysSong:
> It also affects the behavior of #5827

How?

zonkmachine:
> How?

You can now easily grab/drag the plugin, so that's a fix. The cursor behavior has changed a bit though. For instance, earlier you had the hand when you were over the canvas and now it's the arrow. That sounds about right. In the FX Mixer you used to have the hand over the "Add channel" button and now it's the arrow. In the Song Editor you now create a track content objects with the arrow and it used to be the hand cursor. Now when the tco exists it's once more the hand cursor when you move over it. Things like that.

zonkmachine:
OK. I get it now. I saw changes in the rest of lmms but this is because once you open one of the Carla instruments, they won't give you back the arrow. It looks like this persists for the rest of the session.

Merge?

zonkmachine (approved):

PR #5841, opened 2020-12-11
fix compilation with 'AUDIO_PORT_SUPPORT'

zserik:
As this only affects the JACK audio backend, I think this PR won't be affected by #5592.
I added a check for `jack_port_set_name` (deprecated) vs `jack_port_rename`.
see also https://github.com/hydrogen-music/hydrogen/pull/365 (2016 !)


PhysSong:
Thanks for the note! It seems like `jack_port_rename` was introduced in JACK1 0.125.0 and JACK2 1.9.11, but Ubuntu 16.04 still ships with the version of JACK1 without the function while JACK2 is okay.
When using `weakjack`(enabled by default), it dispatches calls to `jack_port_rename` using `jack_port_set_name` internally on runtime, so it's always okay when using `weakjack`.

If you're interested in adding the checks to `CMakeLists.txt`, I can help you. Otherwise, I will add relevant changes to this.

zserik:
I think I should add the appropriate `check_library_exists` check to the `IF(WANT_JACK) { IF(WANT_WEAKCHECK) ... ELSE ... [here] }` section in the `/CMakeLists.txt` file. The lines added would be
```cmake
SET(CMAKE_REQUIRED_LIBRARIES_BACKUP "${CMAKE_REQUIRED_LIBRARIES}")
SET(CMAKE_REQUIRED_LIBRARIES "${JACK_LIBRARIES}")
CHECK_LIBRARY_EXISTS(jack jack_port_rename "" HAVE_JACK_PORT_RENAME)
SET(CMAKE_REQUIRED_LIBRARIES "${CMAKE_REQUIRED_LIBRARIES_BACKUP}")
UNSET(CMAKE_REQUIRED_LIBRARIES_BACKUP)
```
I'm unsure in which header file or such I should place the following line, if at all (I haven't found an appropriate `configure_file` statement in any `CMakeLists.txt`)
```c
#cmakedefine HAVE_JACK_PORT_RENAME
```

zserik:
Oh, the macos CI is broken...

PhysSong (approved):

Issue #5840, opened 2020-12-11
Song Editor - Pasting tco on first position


zonkmachine:
If you try and paste something to the first bar in the Song Editor it ends up on the second bar. This is a glitch that has been introduced after lmms-1.2 .

![paste](https://user-images.githubusercontent.com/6368949/101866766-f4dae580-3b79-11eb-9d14-6f2bb74f8990.png)

#### Affected LMMS versions

master

PhysSong:
Seems like a duplicate of #5393. If it's still not fixed on the latest `master` in #5401, @Spekular will have to investigate.

zonkmachine:
Thanks! A duplicate it is. Closing.

zonkmachine:
Sorry... Those were closed already. Sticking to this one then.

Spekular:
Can't reproduce on my development copy on linux, should I pull an even newer copy to test?

Spekular:
Pulled latest master, still can't reproduce. Steps? OS?

zonkmachine:
He, he. A bit of a mystery then. I think I will have to do a bisect down to where it happens for me on master then.

> OS?

I'm now on Ubuntu/Mate 20.10
Version 1.3.0-alpha.1.33+g4f74151 (Linux/x86_64, Qt 5.6.3, GCC 5.4.0 20160609).

> Steps?

I notice now that this bug is only if you copy a pattern that starts on the first bar to a second instruments first bar so the steps to test replicate this is:

 * Default project, clone the triple oscillator
 * Add a note to the first instrument on the first bar
 * Copy that tco and paste it on the second instrument, first bar

IanCaio:
@zonkmachine That's a weird one. If you create the TCO on the second bar of the first track and copy/paste on the first bar of the second track it works fine. So it seems to be a requirement for the bug to happen that the TCO being copied is at the first bar..

Spekular:
Ok, I know what's happening. Any time you copy from one bar to the same bar, the clip will be offset to avoid landing on itself. Looks like the logic doesn't account for the fact that you're pasting on another track.

Spekular:
Unfortunately I'm lost in the new copy/paste code and can't find where the offset happens anymore. @IanCaio you did the clipboard updates, right?

IanCaio:
> Unfortunately I'm lost in the new copy/paste code and can't find where the offset happens anymore. @IanCaio you did the clipboard updates, right?

[...]

PR #5847, opened 2020-12-13
Fixes bug with pasting of TCOs (#5840)

IanCaio:
Fixes second issue of #5840 . I'd like feedback on the approach, I didn't want to create an exception on the behavior of `TimePos::quantize` since it's consistent, it just happens that we are working with negative values and expected them to be quantized the same way as positive ones. This makes a minor change to the `quantize` and moves most of the fix to the `pasteSelection` method.

`TimePos::quantize` works for negative values, but ends up snapping the TCO to the opposite direction. This is because the snapping down happens in the direction of the origin, which is left for positive values and right for negative values. The snapping up (when we are halfway through the interval) also happens in the opposite direction, away from the origin or right for positive values and left for negative ones:

image

That wasn't accounted for in the pasteSelection method and we ended up with wrong positions when pasting before the TCO(s) we copied. This PR attempts a fix by doing the following:

- Changing TimePos::quantize to have a boolean parameter that will define whether it snaps the position up if we are halfway through the interval or not (this removes the need to use that trick of subtracting `TimePos::ticksPerBar() * snapSize / 2` from the offset to try to "bias" it.
- Checking if we need quantizing at all by checking if the offset falls outsize the snap grid.
- If we do need quantizing and the offset is negative we subtract one snap, since the quantization will move the TCO to the opposite end (to the right). Then we call quantize.

### DIFF
--- a/src/core/TimePos.cpp
+++ b/src/core/TimePos.cpp
@@ -75,9 +75,9 @@ TimePos TimePos::quantize(float bars) const
 	// Ternary expression is making sure that the snap happens in the direction to
 	// the right even if m_ticks is negative and the offset is exactly half-way
 	// More details on issue #5840 and PR #5847
-	int snapUp = (offset == -(interval / 2))
+	int snapUp = ((2 * offset) == -interval)
 		? 0
-		: offset / (interval / 2);
+		: (2 * offset) / interval;
 
 	return (lowPos + snapUp) * interval;
 }
--- a/src/core/TimePos.cpp
+++ b/src/core/TimePos.cpp
@@ -63,7 +63,7 @@ TimePos::TimePos( const tick_t ticks ) :
 {
 }
 
-TimePos TimePos::quantize(float bars, bool snap) const
+TimePos TimePos::quantize(float bars) const
 {
 	//The intervals we should snap to, our new position should be a factor of this
 	int interval = s_ticksPerBar * bars;
@@ -72,7 +72,12 @@ TimePos TimePos::quantize(float bars, bool snap) const
 	//Offset from the lower position
 	int offset = m_ticks % interval;
 	//1 if we should snap up, 0 if we shouldn't
-	int snapUp = snap ? offset / (interval / 2) : 0;
+	// Ternary expression is making sure that the snap happens in the direction to
+	// the right even if m_ticks is negative and the offset is exactly half-way
+	// More details on issue #5840 and PR #5847
+	int snapUp = (offset == -(interval / 2))
+		? 0
+		: offset / (interval / 2);
 
 	return (lowPos + snapUp) * interval;
 }
--- a/src/core/TimePos.cpp
+++ b/src/core/TimePos.cpp
@@ -63,7 +63,7 @@ TimePos::TimePos( const tick_t ticks ) :
 {
 }
 
-TimePos TimePos::quantize(float bars) const
+TimePos TimePos::quantize(float bars, bool snap) const
 {
 	//The intervals we should snap to, our new position should be a factor of this
 	int interval = s_ticksPerBar * bars;
@@ -72,7 +72,7 @@ TimePos TimePos::quantize(float bars) const
 	//Offset from the lower position
 	int offset = m_ticks % interval;
 	//1 if we should snap up, 0 if we shouldn't
-	int snapUp = offset / (interval / 2);
+	int snapUp = snap ? offset / (interval / 2) : 0;
 
 	return (lowPos + snapUp) * interval;
 }
--- a/include/TimePos.h
+++ b/include/TimePos.h
@@ -68,9 +68,7 @@ public:
 	TimePos( const bar_t bar, const tick_t ticks );
 	TimePos( const tick_t ticks = 0 );
 
-	// Snap defines whether we should snap to the closest bar. If it's
-	// false we'll just stick to the lowest bar.
-	TimePos quantize(float bars, bool snap = true) const;
+	TimePos quantize(float) const;
 	TimePos toAbsoluteBar() const;
 
 	TimePos& operator+=( const TimePos& time );
--- a/include/TimePos.h
+++ b/include/TimePos.h
@@ -68,7 +68,9 @@ public:
 	TimePos( const bar_t bar, const tick_t ticks );
 	TimePos( const tick_t ticks = 0 );
 
-	TimePos quantize(float) const;
+	// Snap defines whether we should snap to the closest bar. If it's
+	// false we'll just stick to the lowest bar.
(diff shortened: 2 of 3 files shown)


Spekular:
As the original author: If `(-0.9).quantize(1)` doesn't give `-1` then the quantize function isn't working as intended, and I think that should be fixed rather than introducing workarounds or a `snap` parameter.

When the latter is needed one could just write `pos -= (pos % interval)`, which is no less readable than `pos quantize(1, false)`.

IanCaio:
> As the original author: If `(-0.9).quantize(1)` doesn't give `-1` then the quantize function isn't working as intended, and I think that should be fixed rather than introducing workarounds or a `snap` parameter.
> 
> When the latter is needed one could just write `pos -= (pos % interval)`, which is no less readable than `pos quantize(1, false)`.

`(-0.9).quantize(1)` will give `-1`. The issue is that `(-0.5).quantize(1)` will give `-1` while `(0.5).quantize(1)` will give `1`, meaning that one is snapping down and the other one is snapping up. This breaks the biasing code because it assumes that an offset of 1/2 the interval will snap up, but the quantization of negative values works like a mirror from the positive quantization.

I tried to fix it without changing the logic of `TimePos::quantize` because I don't think it's necessarily wrong, and added the snap parameter because I thought it could be useful to avoid the biasing trick. But as I mentioned on the issue another solution could be to account for that `-1/2 interval` situation and forcing it to snap up instead of down for negative values too.

Spekular:
Huh. Assuming `(-0.1).quantize(1)` also works as intended and gives `0`, I'm not sure I see how the logic breaks. Either way perhaps it's better to just replace the bias + quantize code with a floor towards -1 then, since really that's the intended behavior.

IanCaio:
##### Situation 1:
snapSize is 1 bar (192 ticks). `offset` is 3 bars (576 ticks). Bias subtracts half a bar from it, now `offset` is 480 ticks. `TimePos::quantize` will realize that the position is 2 bars and a half, `snapUp` will be one, so bar 3 will be returned. All fine.

##### Situation 2:
snapSize is 1 bar (192 ticks). `offset` is -3 bars (-576 ticks). Bias subtracts half a bar from it, now `offset` is -672 ticks. `TimePos::quantize` will realize that the position is -3 bars and a half, `snapUp` will be minus one, so bar -4 will be returned. The `snapUp` is working as intended but in a mirrored way.

The illustration on the PR message illustrates it well, the quantization works fine. It's just that when we are __exactly__ in the middle of the interval, the logic is that we snap to the next one. Problem is the next one on the positive side is to the right and the next one on the negative side is to the left. The biasing code depends on that behavior and assumes that it will snap to the right. That's why I wanted to keep the `quantize` logic kind of intact (the snap parameter is more of an convenience, but I can remove it, specially because if you don't look from the illustration perspective the `snap = false` behavior for negative values will look weird).

The alternative I see is having `int snapUp = offset == - (interval / 2) ? 0 : offset / (interval / 2);` on `TimePos::quantize` which will just change the behavior from `(-1/2x).quantize(x)` to snap to the right, independent of whether it's negative or positive.

IanCaio:
@Spekular I used the alternative approach I mentioned in the last commit. I think it might be better since it's requires less change to the code. The only drawback, which in practice is not a problem at all, is that `quantize` behavior is not entirely symmetrical (when we are halfway through the interval we always snap to the right). After writing it I think I prefer this over the old one.

PhysSong:
What should the following calls return in an ideal implementation?
- `(-0.9).quantize(1)`
- `(-0.5).quantize(1)`
- `(-0.1).quantize(1)`
- `(+0.1).quantize(1)`
- `(+0.5).quantize(1)`
- `(+0.9).quantize(1)`

IanCaio:
> What should the following calls return in an ideal implementation?

I think it always should snap to the closes quantized distance even on negative values (like the timeline extended backwards). It actually already does that on master. The last commit just changes the snapping direction when we have `(-x/2).quantize(x)` so it snaps to the right.

 * `(-0.9).quantize(1)` -> -1 (On `master`: -1)
 * `(-0.5).quantize(1)` ->  0 (On `master`: -1) *
 * `(-0.1).quantize(1)` ->  0 (On `master`:  0)
 * `(+0.1).quantize(1)` ->  0 (On `master`:  0)
 * `(+0.5).quantize(1)` ->  1 (On `master`:  1)
 * `(+0.9).quantize(1)` ->  1 (On `master`:  1)

_*The paste method assumed this would snap to the right instead of to the "next" quantized position, which walking to the negative side is to the left_

src/core/TimePos.cpp
```diff
@@ -72,7 +72,12 @@ TimePos TimePos::quantize(float bars) const
 	//Offset from the lower position
 	int offset = m_ticks % interval;
 	//1 if we should snap up, 0 if we shouldn't
-	int snapUp = offset / (interval / 2);
+	// Ternary expression is making sure that the snap happens in the direction to
+	// the right even if m_ticks is negative and the offset is exactly half-way
+	// More details on issue #5840 and PR #5847
+	int snapUp = (offset == -(interval / 2))
```
PhysSong, line 78: This won't wotk properly if `interval` is odd, please multiply both sides by 2.

IanCaio: Oh true, integer division, my bad 😬 
Will fix once I get home

Spekular, line 78: I wonder if this could be condensed by using `std::remainder` to get the offset instead of `%`? It feels like there must be a way to do this w/o such a specific condition, but this works if not.

Spekular: It was there from the start as well, so blame me instead! 😅

Spekular:
> What should the following calls return in an ideal implementation? [...]

I think the behavior as of the latest commit is good. Negative inputs don't make any sense for positions, so this should only affect deltas, like moving or resizing. In those cases I can't see any downsides to the new behavior, and it should fix the issue found in #5840.

src/core/TimePos.cpp
IanCaio: What would be the advantage of using `std::remainder`? Wouldn't the currently used modulus operator be easier and faster since we are working with integers?

IanCaio: Fixed on last commit! The equation might look less obvious, think I should've added a comment?

Spekular: I believe it behaves differently from `%` in what kind of remainder it gives, since there's different ways to do modulo. I was thinking maybe it would somehow let us avoid the ternary. Let's go with the ternary instead though, it's fine that way too.

PhysSong:
Merge?

Spekular (approved):

PR #5848, opened 2020-12-13
pianoroll: nudge/snap while dragging notes

rghvdberg:
Adds combobox for switching behavior of the snapping to grid when dragging.
Old behavior (nudge) is default.
Notice that snapping is relative to the dragged note and not the 1st note in the selected range.

It's essentially the same as #4357 but I lost that fork and just started over.
lmms_snap


Spekular:
Would you consider adding snapping to the end of the note as well? (Choosing whichever is closest to the unquantized position, as with shift drag in the song editor). It would be convenient for reversed samples and such, and shouldn't interfere much with the existing functionality in most situations.

I think this is worth adding regardless, but it's a nice-to-have if you're not opposed.

rghvdberg:
Didn't think of that. Will do.

Op ma 14 dec. 2020 11:25 schreef Spekular <notifications@github.com>:

include/PianoRoll.h
```diff
@@ -294,7 +303,8 @@ protected slots:
 	int noteEditRight() const;
 	int noteEditLeft() const;
 
-	void dragNotes( int x, int y, bool alt, bool shift, bool ctrl );
+	void dragNotes( int x, int y, bool alt, bool shift, bool ctrl, Note * nt );
+	Note * draggedNote;
```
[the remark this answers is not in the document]
rghvdberg: fixed

src/gui/editors/PianoRoll.cpp
```diff
@@ -2595,6 +2612,41 @@ void PianoRoll::dragNotes( int x, int y, bool alt, bool shift, bool ctrl )
...
+					copy->quantizePos(quantization());
+
+					int gap_end = qAbs(end_pos_ticks - copy->pos().getTicks());
+					
+					if (gap_start < gap_end)
+					{
+						copy->setPos(tmp_pos); // restore start position
+						off_ticks += copy->pos().getTicks() - (m_draggedNote->oldPos().getTicks() + off_ticks);
+					}
+					else
```
PhysSong, line 2641: Trailing whitespaces:
```suggestion
					else
```

src/gui/editors/PianoRoll.cpp
```diff
@@ -4360,8 +4421,22 @@ Note * PianoRoll::noteUnderMouse()
...
 
+void PianoRoll::changeSnapMode()
+{
+	//	gridNudge,
+	//	gridSnap,
+	//	gridFree - to be implemented
 
-
+	QString qs_snapMode = m_snapModel.currentText();
+	if (qs_snapMode == "Nudge")
```
PhysSong, line 4431: Translations may break this.

superpaik:
Suggestion: to me it's a bit odd that when snap activated notes are snapping both from the start and the end of the note, and make it seem that the movement it's kind of random (though I see it is as intended). Maybe it would be better if snap has in mind the direction of the movement. If going left, snap with the start of the note. If going right, snap with the end of the note.
I don't know it is correct or not, but when in pianoroll inside beat+bass editor, "no duration" notes do not seem to follow the same behavior as "normal" notes.

rghvdberg:
Snapping that depends on the direction of drag seems like a bad idea. What if I just want to move 1/4 note to the right? I'd first have to drag beyond that note and then drag left to finally land there.
We could use a modifier key to enable snap to end and the default being snap to start.
I don't know if there are any modifier keys left to implement in pianoroll though.

rghvdberg:
I only tested (very shortly) in pianoroll in the song editor. I'll look at the bb editor case.

src/gui/editors/PianoRoll.cpp
rghvdberg: Would using `m_snapModel.getValue()` be better ?
It returns 0 (Nudge) and 1 (Snap).
I am uneasy using 'magic numbers' in code. Maybe add an `enum` for these values?

rghvdberg:
found a bug :-(
when you drag a note over a second note which the same pitch, the second note is moved a bar ...

https://user-images.githubusercontent.com/10710663/104838584-1ab38280-58bc-11eb-87a6-370b25a23e62.mp4

src/gui/editors/PianoRoll.cpp
Veratil: I'd make an enum and compare to that.

include/PianoRoll.h
```diff
@@ -294,7 +303,8 @@ protected slots:
 	int noteEditRight() const;
 	int noteEditLeft() const;
 
-	void dragNotes( int x, int y, bool alt, bool shift, bool ctrl );
+	void dragNotes( int x, int y, bool alt, bool shift, bool ctrl, Note * nt );
+	Note * m_draggedNote;
```
IanCaio, line 307:
```suggestion
	void dragNotes(int x, int y, bool alt, bool shift, bool ctrl);
```

From what I read in the code it seems that `m_draggedNote` is being used with the same goal as `m_currentNote`. Using the latter would probably fix the bug with dragging a note on top of another one (since `m_currentNote` is set on the `mousePressEvent` and will not change during `mouseMoveEvent`) and will prevent adding unnecessary variables to the method and class.

src/gui/editors/PianoRoll.cpp
IanCaio: There's already an enum being used for `m_gridMode`, can't we just reuse that? I think just like the enum, the ComboBox model will go from 0 up, then you can cast and compare them.

include/PianoRoll.h
```diff
@@ -253,6 +255,13 @@ protected slots:
 		PR_BLACK_KEY
 	};
 
+	enum GridMode
+	{
+		gridNudge,
+		gridSnap,
+		gridFree
```
IanCaio, line 262:
```suggestion
		//gridFree
```
I'd remove or comment this out while it's not implemented

src/gui/editors/PianoRoll.cpp
```diff
@@ -2595,6 +2612,41 @@ void PianoRoll::dragNotes( int x, int y, bool alt, bool shift, bool ctrl )
 			else
 			{
 				// moving note
+				if (m_gridMode == gridSnap && firstIteration)
+				{
+					firstIteration = false;
+					Note* copy(m_draggedNote);
+					// guantize first note
```
IanCaio, line 2619:
```suggestion
					// Quantize first note
```

IanCaio (changes requested):
Just an initial review. As said in Discord, I think the bug that you found is being caused by calling `noteUnderMouse` repeatedly to get the note being dragged on the `mouseMoveEvent` method. When two notes are in the same position you might end up with the wrong note on the `dragNotes` call. `m_draggedNote` seems redundant when you have `m_currentNote` and I think you could get rid of it and use the latter instead, fixing the bug and also removing an unnecessary variable.

I wrote a few other comments, but before going further with the review I was thinking about suggesting a different approach on the logic that snaps the note to the grid. Why not using the mouse position, quantizing it and then just calculating the offset necessary to get the note from where it is to that quantized mouse position? It sounds to me like it would make the logic simpler (wouldn't require copying the note, quantizing it and moving it around) and it seems like it would work as intended. The only difference in behavior that I can think of is that the position of the mouse in relation to the grid would matter more than the position of the mouse against the initial position of the note. Would have to check in terms of UX if that behavior is annoying or not, but if it is it shouldn't be too complicated to add the offset against the initial position of the note to the math.

I can help you out writing this alternative approach if you like!

superpaik:
It woks good to me! Tested on regular and "no-duration" (beats) notes.
I think that using the mouse position instead the first note was a great idea! 
Thanks!

IanCaio (approved):

src/gui/editors/PianoRoll.cpp
```diff
@@ -2550,10 +2566,9 @@ void PianoRoll::dragNotes( int x, int y, bool alt, bool shift, bool ctrl )
 
 
 	// if they're not holding alt, quantize the offset
-	if( ! alt )
+	if (!alt)
```
Spekular, line 2569: Why not move this so off_ticks is only quantized if you're in nudge mode? Then you should be able to use off_ticks for the snap calculations as well, instead of recalculating it from mouse positions and stuff below.

Spekular (commented):
The snapping code looks unnecessarily long to me, and reasonably it seems that the logic for snap vs nudge should be on the same level, rather than snap being calculated "inside" the nudge logic and overriding it.

IanCaio:
> The snapping code looks unnecessarily long to me, and reasonably it seems that the logic for snap vs nudge should be on the same level, rather than snap being calculated "inside" the nudge logic and overriding it.

Ideally the whole `dragNotes` method could use some refactoring that would make it easier to have nudge and snap more explicitly separated. I'm not sure how I would do it without rewriting a lot of the method. Since it doesn't make it worse, I'd go with that approach for now and leave that method overhaul for when we are out of that tight refactoring schedule.

src/gui/editors/PianoRoll.cpp
IanCaio: I'm not sure we can really reuse `off_ticks` for the snap mode: `off_ticks` is the offset from the initial position the mouse clicked in ticks, while the position we calculate in the snap mode is the absolute mouse position in ticks on the time line. Maybe if we added `m_moveStartX * TimePos::ticksPerBar() / m_ppb;` to it, but then we are having to do the conversion to ticks again anyways, right?

Spekular: `note->oldPos().getTicks() + off_ticks` should give the absolute position, no?

Spekular: In fact, this is exactly what the existing code already does at line 2598 (2645 in the PR version at time of writing).

```
int pos_ticks = note->oldPos().getTicks() + off_ticks;
...
note->setPos(TimePos(pos_ticks));
```

IanCaio: Sure, I'm giving it a shot. Though I still think this will become a rabbit hole and I'll end up rewriting more than expected.

IanCaio: Last two commits separate the nudge and snap logic more explicitly and reuse off_ticks as you suggested. It required some rearrangement of code. Method still needs refactoring and I tried to rewrite the least possible. A bug that is present on `master` was fixed on the calculation of the boundaries limits for the note keys.

src/gui/editors/PianoRoll.cpp
```diff
@@ -1182,10 +1189,13 @@ void PianoRoll::keyPressEvent(QKeyEvent* ke)
...
-						dragNotes( m_lastMouseX, m_lastMouseY,
-									ke->modifiers() & Qt::AltModifier,
-									ke->modifiers() & Qt::ShiftModifier,
-									ke->modifiers() & Qt::ControlModifier );
+						dragNotes(
+							m_lastMouseX,
+							m_lastMouseY,
+							ke->modifiers() & Qt::AltModifier,
+							ke->modifiers() & Qt::ShiftModifier,
+							ke->modifiers() & Qt::ControlModifier
```
Veratil, line 1197: Since we repeat this function call a few times, can we at least fix this? Just pass the event or modifiers in and test in the function for what we need.

Spekular: I'm not strongly for or against this change, but wouldn't that couple dragNotes more closely to (1) the Qt UI and (2) these specific modifiers? If for example the argument "alt" were renamed to "unquantized" (or inverted and renamed to "quantized") then the dragNotes function is independent of which shortcut is used to perform unquantized actions.

Spekular (approved):
LGTM after Ian's latest changes.

src/gui/editors/PianoRoll.cpp
Veratil: Dragging notes is inherently a UI thing, not a core thing.

I do like the idea of changing the names from `ctrl` `alt` and `shift`, and I would rather just pass in a `modifiers` value to the function.

Spekular: > Dragging notes is inherently a UI thing, not a core thing.

Sure, but at the moment this function contains a lot of logic that doesn't rely on the UI. The same quantization strategies are relevant regardless of how the move is input.

IanCaio: I think the renaming of the arguments and possibly compressing them in a single one seems to be the middle ground: We keep the method decoupled from the Qt event and make the modifiers more generic (allowing us to change them later).

Can I suggest that we do it separately though? As I mentioned on a previous comment, this method needs a bit of an overhaul. Just renaming the arguments would also require some thought because some modifiers are used for multiple actions depending on context. I think both the method refactoring and changing of arguments could be done in a single separate PR, they would both benefit on each other.

Veratil: > Can I suggest that we do it separately though?

Sure. 👍

IanCaio:
Merging tomorrow if there are no objections. There's one conflict but it can be quickly fixed in the web editor before merging.

PR #5850, opened 2020-12-15
Update knob graphic no matter how small change is

### LINKED ISSUE #5747: Knob parameter value graphic (green circle) shows even when value is 0
> **Edit:** @tresf / @Spekular added screenshot of issue
> 

### Bug Summary
Each knob has a green circle graphic indicating the value of the parameter which is greyed out when the value is at 0.
In certain situations the knob graphic can show a little green piece on the left side in the beginning of this circle even though the value is 0.  
#### Steps to reproduce
The easiest way to reproduce this is if you click your mouse on the knob and instead of dragging straight up or down to change the parameter like you normally would, drag your mouse pretty far out to the left and then slowly down.
You will see the value of the knob is going down (decreasing) and if done successfully you will replicate this issue.
I have experienced this in other ways, but I think this is the best way to generate an example of this issue.

#### Expected behavior
Graphic knob indicator is supposed to be greyed out when value is 0
#### Actual behavior
At times graphic knob indicator can show a little value graphically even though the actual parameter value is at 0

Not a big deal because only really happens when using the knob improperly, but if easy to fix may be worth.

#### Affected LMMS versions 
1.2.2

Spekular:
This one liner fixes #5747. See https://github.com/LMMS/lmms/issues/5747#issuecomment-719689944

No evident issues, but maybe someone should test on an incredibly weak machine to be extra sure. On the off chance that performance issues arise I'll switch to this alternative solution: https://github.com/LMMS/lmms/issues/5747#issuecomment-719711424

### DIFF
--- a/src/gui/widgets/Knob.cpp
+++ b/src/gui/widgets/Knob.cpp
@@ -340,7 +340,7 @@ bool Knob::updateAngle()
 	{
 		angle = angleFromValue( model()->inverseScaledValue( model()->value() ), model()->minValue(), model()->maxValue(), m_totalAngle );
 	}
-	if( qAbs( angle - m_angle ) > 3 )
+	if( qAbs( angle - m_angle ) > 0 )
 	{
 		m_angle = angle;
 		return true;


Spekular:
This is such a tiny change that I'll merge it without reviews, assuming there are no concerns raised by Friday.

zonkmachine:
LGTM! I'm going to go ahead and just merge this.

PR #5857, opened 2020-12-23
4 new note modification tools

allejok96:
Add 4 new tools to the Piano roll note tools menu:
* Fill ([requested by Avicii](https://youtu.be/-dIcuU58Oy8?t=1410))
* Cut overlaps
* Min length
* Max length

bild


Iron-Squid:
These are pretty good tools, but I suggest `latest` be changed to `Last note` to make it more consistent with the `Note Length` drop-down menu.

allejok96:
Thanks, I wanted an English speaker's input on that one. I was afraid "last" could be interpreted as "rightmost".

src/gui/editors/PianoRoll.cpp
```diff
@@ -712,6 +712,89 @@ void PianoRoll::glueNotes()
...
+		m_pattern->addJournalCheckPoint();
+
+		NoteVector notes = getSelectedNotes();
+		if (notes.empty())
+		{
+			notes = m_pattern->notes();
+		}
+
+		// Sort notes by position
+		std::sort(notes.begin(), notes.end(), [](const Note *n1, const Note *n2){ return n1->pos() < n2->pos(); });
```
IanCaio, line 728:
```suggestion
		std::sort(
			notes.begin(),
			notes.end(),
			[](const Note *n1, const Note *n2){ return n1->pos() < n2->pos(); }
		);
```

Just breaking the line to shorter ones.

IanCaio (approved):
Code LGTM! Minor code style suggestion.

We also need some testing to check this approach fits all case scenarios, from reading the logic it seems it does.

src/gui/editors/PianoRoll.cpp
```diff
@@ -712,6 +712,93 @@ void PianoRoll::glueNotes()
...
+		if (notes.empty())
+		{
+			notes = m_pattern->notes();
+		}
+
+		// Sort notes by position
+		std::sort(
+			notes.begin(),
+			notes.end(),
+			[](const Note *n1, const Note *n2){ return n1->pos() < n2->pos(); }
```
PhysSong, line 731: FYI, there is `Note::lessThan` which takes care of the pitch as well. You're encouraged to use that to prevent breaking the existing ordering.

allejok96: Thanks, good find

PhysSong:
Looks good. Merge?

IanCaio:
My approval still stands, can merge it IMO!

superpaik:
Can you explain what is the expected behaviour for each new tool so I can test them based on that?
Suggestion: add a short-cut to the last two new funcionalities, as the other have

allejok96:
@superpaik

- Fill: grow notes/chords to the next note/chord
- Cut overlaps: reduce notes/chords to stop where the next note/chord begins
- Min length: grow notes to be no shorter than the last used (last clicked) note
- Max length: reduce note lengths to be no longer than last used

I don't think we should add shortcuts to the last two since they are more of a "use once" action, and we shouldn't waste shortcuts.

Iron-Squid:
What if instead of having separate shortcuts for different tools, we had one shortcut for the current selected tool, and could switch between them with `shortcut + up/down arrow keys`?
Just a suggestion.

allejok96:
Interesting suggestion @Iron-Squid ... speaking for myself, I'd rather keep the shortcuts for faster access to the more common tools

superpaik:
Thanks for explaining how it should work.
Some issues found:
Fill: sometimes notes/chord grow to end of next beat, sometimes no (test 1)
Cut overlaps: somehow have the same problem (test 2)
If the videos aren't clear, please don't hesitate to ask. In the videos, actions are executed via short-cut.
I tested it with the "no duration" notes (the one that are created inside the "beat+bassline editor") and they work well. It's a perfect feature to transform those notes into regular notes, something that right now it's almost impossible. Well done!

As for the short-cut for the other options, I don't know how internally short-cut works, but if they are only local to the piano-roll (so they can be used on other windows with other features) I don't see them as a waste. To me the two options aren't a "one shot" because they can be usefull when you are writing a melody and you want to test different note durations (like 1/4, 1/8, ...) so you change one and apply it to the others. I know it can be done via selecting all notes and changing them all at once, but I find that with this option it would be quicker and easier.
test1
test2

allejok96:
Good find @superpaik. I designed Fill to grow the notes to the end of the bar. But it's a bug that Cut cuts the last note spanning a bar line...

allejok96:
Test 2 bug is fixed.

Test 1 bug remains. It happens when the last note ends before the bar line, but a previous note spans across the line... I don't think it's worth adding a check for that to the algorithm. The computer still does 99% of it right. Or am I wrong?

@superpaik I added min/max as a lazy way to fix extremely short/long notes created by fill/cut. I understand it can be used in many different ways, but when editing and trying different note lengths, wouldn't it work just as good using `ctrl+A` + `resize` + `shift+C` like shown below?

(And to be honest I think the design of min/max is backwards when the user has to click a note of the desired length *first*, then click the tool... But it was the easiest solution.)

https://user-images.githubusercontent.com/7693838/104821516-b7364000-583c-11eb-9ee3-b87dd5611dbe.mp4

superpaik:
Sorry. I couldn't retest it because somehow there isn't the functionality on the installables generated by the bot (checked both mingw and MSVC)

IanCaio:
As you mentioned, the bug 1 is actually the behavior of the algorithm, but it might be a little surprising to the user. What if the behavior for the last chord was: Stretch to the next full bar OR to end position of the note that reaches further, whichever is bigger? Then in that case the small note would fill up to the end of the previous one. The changes to the code would look a bit like this:

```cpp
			TimePos lastChordEndPos(0);
// ...
			// Update the end position for the last chord, which will stretch
			// to the position where the furthest chord ends or to the
			// next full bar, whichever is bigger
			lastChordEndPos = qMax(notes[i].endPos(), lastChordEndPos);

			// Stretch last chord to end of bar or to the calculated position
			// notes[i] will always be the longest note in the chord
			if (i+1 == notes.count())
			{
				TimePos lastBarPos(notes[i]->endPos().nextFullBar() * TimePos::ticksPerBar())
				lastChordEndPos = qMax(lastBarPos, lastChordEndPos);
				chordLength = lastChordEndPos - notes[i]->pos();
			}
```

I'm not sure that behavior is more user friendly, but at least you wouldn't unexpectedly have the last chord ending before the previous chord. Let me know if I didn't explain it well enough, can throw some images to give some examples.

allejok96:
Thanks @IanCaio, now both bugs @superpaik mentioned are fixed, and your approach made the code simpler too.

Current implementation: Fill stretches the last chord to the end of the last bar (the end of the clip/tco in song editor).

superpaik:
HI! For the fill, I don't know if the behaviour attached is working as expected or no. To me is weird.
As for the cutOverlaps, I think still there are some issues.
test-fill
test-cutOverlaps

superpaik:
Okey!
I understand there are so many different scenarios so it's very difficult to make it work as the user wants/expect. If the functionality is clear and is working as expected, then is ok. The only thing I'd do is write the current behavior in the help so people know what to do with those tools, because that concept of "chords" could be misleading from a user point of view (not always all the notes of a chord start at the same position -for example, if you are adding swing or want to sound more "human")
Anyway, I'm aware of the refactoring phase and all the work behind it. So to me, these tools are ok.
Thanks for the quick and detalied explanation!

allejok96:
@superpaik you've brought good feedback. There will always be edge cases. The more weird scenarios we try to account for, the more complex the logic becomes and the more unpredictable the tool is. So we need to come up with a simple and straightforward rule for how the tool should work.

**GIF 1:**
Maybe Fill should extend the END position of notes to the closest start position to the right? The more I think about it, the better it seems.

**GIF 2:**
I think this is *exactly* what the user can expect, because...
1. It shouldn't cut the end of B4 because that part *doesn't overlap anything*.
2. It shouldn't cut the beginning of B4 because that would significantly alter the melody.
3. It shouldn't remove D#4 because that would remove all harmony in a song.
4. It shouldn't extend D#4 to the length of B4, because the tool is called "cut".

If a user is writing swingy chords, that user won't use Cut overlaps, because it will destroy everything, and that's it.

IanCaio:
> Maybe Fill should extend the END position of notes to the closest start position to the right? The more I think about it, the better it seems.

I don't think that would be a bad idea. It would work for both the chord situations you were originally making it for (from Avicii video) and on the situations superpaik was expecting. It would be slightly less efficient because for each end position you'd have to go through all notes following it until you found one that has a beginning position higher or equal to the end position and then stretch the single note. I could sketch some code if you want help with it, might have some free time tomorrow to work on something.

I agree the cut actions seems to be good as it is, I can't think of a better behavior right now at least.

superpaik:
Yes, I agree with you both. Maybe I'm overthinking things because I'm in "testing" mode ;-) Maybe we should follow the Pareto Principle and focus on the 80% of most used and normal scenarios. Thanks for the effort.

IanCaio:
> Yes, I agree with you both. Maybe I'm overthinking things because I'm in "testing" mode ;-) Maybe we should follow the Pareto Principle and focus on the 80% of most used and normal scenarios. Thanks for the effort.

The UX perspective is very important too, so keep bringing those concerns up! I think both the author and myself agree with you on the first case, I think the fill behavior could be improved to fill up to the closest note beginning (even if it's 2 or 3 notes away).

src/gui/editors/PianoRoll.cpp
```diff
@@ -712,6 +712,85 @@ void PianoRoll::glueNotes()
...
+		std::sort(notes.begin(), notes.end(), Note::lessThan);
+	}
+	if (fill)
+	{
+		std::sort(notes.begin(), notes.end(), [](Note *n1, Note *n2) { return n1->endPos() < n2->endPos(); });
+	}
+
+	int length;
+	NoteVector::iterator ref = refNotes.begin();
+	for (Note *note : notes)
```
IanCaio, line 741:
```suggestion
	for (Note* note : notes)
```

src/gui/editors/PianoRoll.cpp
```diff
@@ -712,6 +712,85 @@ void PianoRoll::glueNotes()
...
+	{
+		// Fast forward to next reference note
+		while (ref != refNotes.end() && (fill ? (*ref)->pos() < note->endPos() : (*ref)->pos() <= note->pos()))
+		{
+			ref++;
+		}
+		if (ref == refNotes.end())
+		{
+			// Last notes stretch to end of last bar
+			length = notes.last()->endPos().nextFullBar() * TimePos::ticksPerBar() - note->pos();
```
IanCaio, line 751:
```suggestion
			// Last notes stretch to end of last bar
			length = note->endPos().nextFullBar() * TimePos::ticksPerBar() - note->pos();
```

I'm still trying to think if there's a possibility that `ref` reaches `refNotes.end()` before `note` reaches `notes.last()`. But that change doesn't create an issue if it doesn't, but prevents an issue if it does.

src/gui/editors/PianoRoll.cpp
```diff
@@ -712,6 +712,85 @@ void PianoRoll::glueNotes()
...
+	m_pattern->addJournalCheckPoint();
+
+	NoteVector notes = getSelectedNotes();
+	if (notes.empty())
+	{
+		notes = m_pattern->notes();
+	}
+
+	TimePos bound = m_lenOfNewNotes;  // will be length of last note
+	for (Note *note : notes)
```
IanCaio, line 781:
```suggestion
	for (Note* note : notes)
```

src/gui/editors/PianoRoll.cpp
```diff
@@ -4432,18 +4511,34 @@ PianoRollWindow::PianoRollWindow() :
...
 	noteToolsButton->setIcon(embed::getIconPixmap("tool"));
 	noteToolsButton->setPopupMode(QToolButton::InstantPopup);
 
-	// Glue
 	QAction * glueAction = new QAction(embed::getIconPixmap("glue"),
 				tr("Glue"), noteToolsButton);
 	connect(glueAction, SIGNAL(triggered()), m_editor, SLOT(glueNotes()));
 	glueAction->setShortcut( Qt::SHIFT | Qt::Key_G );
 
+	QAction *fillAction = new QAction(embed::getIconPixmap("fill"), tr("Fill"), noteToolsButton);
```
IanCaio, line 4523:
```suggestion
	QAction* fillAction = new QAction(embed::getIconPixmap("fill"), tr("Fill"), noteToolsButton);
```

src/gui/editors/PianoRoll.cpp
```diff
@@ -4432,18 +4511,34 @@ PianoRollWindow::PianoRollWindow() :
...
+	connect(fillAction, &QAction::triggered, [this](){ m_editor->fitNoteLengths(true); });
+	fillAction->setShortcut( Qt::SHIFT | Qt::Key_F );
+
+	QAction *cutOverlapsAction = new QAction(embed::getIconPixmap("cut_overlaps"), tr("Cut overlaps"), noteToolsButton);
```
IanCaio, line 4527:
```suggestion
	QAction* cutOverlapsAction = new QAction(embed::getIconPixmap("cut_overlaps"), tr("Cut overlaps"), noteToolsButton);
```

src/gui/editors/PianoRoll.cpp
```diff
@@ -4432,18 +4511,34 @@ PianoRollWindow::PianoRollWindow() :
...
+	connect(cutOverlapsAction, &QAction::triggered, [this](){ m_editor->fitNoteLengths(false); });
+	cutOverlapsAction->setShortcut( Qt::SHIFT | Qt::Key_C );
+
+	QAction *minLengthAction = new QAction(embed::getIconPixmap("min_length"), tr("Min length as last"), noteToolsButton);
```
IanCaio, line 4531:
```suggestion
	QAction* minLengthAction = new QAction(embed::getIconPixmap("min_length"), tr("Min length as last"), noteToolsButton);
```

src/gui/editors/PianoRoll.cpp
```diff
@@ -4432,18 +4511,34 @@ PianoRollWindow::PianoRollWindow() :
...
+	connect(minLengthAction, &QAction::triggered, [this](){ m_editor->constrainNoteLengths(false); });
+
+	QAction *maxLengthAction = new QAction(embed::getIconPixmap("max_length"), tr("Max length as last"), noteToolsButton);
```
IanCaio, line 4534:
```suggestion
	QAction* maxLengthAction = new QAction(embed::getIconPixmap("max_length"), tr("Max length as last"), noteToolsButton);
```

src/gui/editors/PianoRoll.cpp
```diff
@@ -4432,18 +4511,34 @@ PianoRollWindow::PianoRollWindow() :
...
 
-	// Glue
 	QAction * glueAction = new QAction(embed::getIconPixmap("glue"),
 				tr("Glue"), noteToolsButton);
 	connect(glueAction, SIGNAL(triggered()), m_editor, SLOT(glueNotes()));
 	glueAction->setShortcut( Qt::SHIFT | Qt::Key_G );
 
+	QAction *fillAction = new QAction(embed::getIconPixmap("fill"), tr("Fill"), noteToolsButton);
+	connect(fillAction, &QAction::triggered, [this](){ m_editor->fitNoteLengths(true); });
+	fillAction->setShortcut( Qt::SHIFT | Qt::Key_F );
```
IanCaio, line 4525:
```suggestion
	fillAction->setShortcut(Qt::SHIFT | Qt::Key_F);
```

src/gui/editors/PianoRoll.cpp
```diff
@@ -4432,18 +4511,34 @@ PianoRollWindow::PianoRollWindow() :
...
+
+	QAction *cutOverlapsAction = new QAction(embed::getIconPixmap("cut_overlaps"), tr("Cut overlaps"), noteToolsButton);
+	connect(cutOverlapsAction, &QAction::triggered, [this](){ m_editor->fitNoteLengths(false); });
+	cutOverlapsAction->setShortcut( Qt::SHIFT | Qt::Key_C );
```
IanCaio, line 4529:
```suggestion
	cutOverlapsAction->setShortcut(Qt::SHIFT | Qt::Key_C);
```

IanCaio (changes requested):
Minor code style fixes and one small possible fix on the logic. Besides that the new logic looks good, without testing (just reading) it seems to fit what @superpaik was expecting as the behavior.

src/gui/editors/PianoRoll.cpp
PhysSong: I think it's the desired behavior in fill mode. In cut mode, this `if` clause is causing a bug:
1. Create a 1/2 bar note
2. Create another 3/2 bars note at the same position and higher key
3. `Cut overlaps` may truncate the 3/2 bars note to 1 bar(I'm not sure if the behavior is well-defined, but at least Linux+GCC builds allow reproducing this bug)

PhysSong: I think `length` should be `note->length()` in cut mode.

IanCaio: My bad, for some reason I thought `notes.last()` was the one being resized, but it was only the one being used as a reference to get the last bar. It should work fine for the fill action then (using `note->endPos()` instead would fill every note that doesn't have a following reference note to the next bar, which might be different for each one. Not sure if that behavior would be desired).

But it ended up making you find that bug you mentioned, exactly because `notes.last()` is being used. Since having the 3/2 bars note in a higher key makes the 1/2 bar note the last, its next full bar is used instead.

IanCaio: I wouldn't mind `length` being used instead. But for the user, wouldn't it be preferred that notes that "overlap" the last bar are cut to fit in?

IanCaio: For the sake of simplicity, maybe for now we should go with the `length` solution. The idea of cutting the chord to the last bar would require us to go through all notes of the chord to see which is shortest before calculating the length.

That should do it then:
```cpp
length = fill
? note.last()->endPos().nextFullBar() * TimePos::ticksPerBar() - note->pos()
: note->length()
```

src/gui/editors/PianoRoll.cpp
allejok96: Is this LMMS' preferred style? Then it should be added to the [Coding Conventions](https://github.com/LMMS/lmms/wiki/Coding-conventions). It currently says if a convention is not mentioned there it may fall under the Qt coding style, which in this case is asterisk to the right.

src/gui/editors/PianoRoll.cpp
allejok96: Good find, both of you. I don't think Cut should cut notes where the last bar line is, that makes little sense. I'll just add

```c++
if (!fill) { continue; }
```

src/gui/editors/PianoRoll.cpp
IanCaio: I think it wasn't added yet, only discussed on Discord. Maybe I should bring it up about adding it to the wiki page as well.

src/gui/editors/PianoRoll.cpp
IanCaio: It's good by me, better than my suggestion!

src/gui/editors/PianoRoll.cpp
```diff
@@ -712,6 +712,86 @@ void PianoRoll::glueNotes()
...
+	for (Note* note : notes)
+	{
+		// Fast forward to next reference note
+		while (ref != refNotes.end() && (fill ? (*ref)->pos() < note->endPos() : (*ref)->pos() <= note->pos()))
+		{
+			ref++;
+		}
+		if (ref == refNotes.end())
+		{
+			if (!fill) { continue; }
```
PhysSong, line 750: It's better to `break` here because you'll keep hitting this `if` clause.

IanCaio (approved):
Sorry about the delay, after the last fixes it LGTM!

PhysSong:
@allejok96 Could you resolve the conflict with the piano-roll knife PR? I think it can be merged after conflict resolution.

PhysSong (approved):

IanCaio:
Also merging this one tomorrow if no one objects!

Issue #5843, opened 2020-12-12
LMMS 1.3.0 Alpha PianoRoll Minor Graphic Adjustments


RoxasKH:
A couple of things feel a bit off to me from a design standpoint in the Alpha build.

This is how the pianoroll looks like at the moment

![pr1](https://user-images.githubusercontent.com/40141826/101986142-715de900-3c8c-11eb-95a6-fa24ca3ea9b6.PNG)

In my opinion, the contrast of the gradient of the new pianoroll keyboard is way too high compared to the rest of LMMS, and this makes it feels a bit annoying and unprofessional. An analog things counts for the new implemented ghost notes. They are a bit too little visible, and on the contrary the black link at the end feels too much saturated and uncomfortable.

Something like this would probably fit better
![pr2](https://user-images.githubusercontent.com/40141826/101986317-943ccd00-3c8d-11eb-8c4e-92426f9a5257.png)

The color adjustements aren't big, but they make the difference. Speking of the keys i kept the gradient but i reduced the contrast, instead for ghost notes, i made them a little brighter so that they stand out more and used a dark grey for the end of the note, which imo is better than the old black.

Obviously this are just suggestions based on my personal opinion, let me know what you think about it.

he29-net:
IIRC, when Veratil made the Piano Roll refactor, he mentioned in the PR that the color of keys is more of a "placeholder" and should be replaced if needed.

Subjectively, I also felt that the contrast is too high. But I would only make the black keys lighter and keep the white keys pure white. In your mockup they seem kind of "bland" and "dirty" to me, and it would also make them more difficult to distinguish from _disabled_ keys, which will be added in #5522 :
![screen](https://user-images.githubusercontent.com/20932751/101993676-bdc11d00-3cbc-11eb-8710-ac1075b6d400.png)

One more consideration for the black keys is that they should be light enough to allow a noticeable color hue to be added. I want to add a way to assign user-defined colors to the keys in a future microtuner update, to help with orientation in scales that are longer or shorter than the repeating 12-key pattern. For example, you may know that your scale in 24-EDO tuning starts at C, but is it C2 / C4 / C6 /.., or an odd-numbered C? With a color assigned to the first degree it would be clear at a glance and you would not need to even read the labels. For non-repeating scales this would be even more important as the traditional visual cues lose all meaning.

[...]

PR #5862, opened 2020-12-25
Pianoroll graphic adjustments 1.3

RoxasKH:
Refers to #5843 .

This is actually nothing code related but only a couple of color changes in the css style files of classic and default themes.
It fixes the key gradient in pianoroll, and made ghost notes stand out a bit more withouth having an high contrast.

Before:
immagine

After:
immagine

This is my first PR, and i hope i didn't get anything wrong, but i shouldn't as it's pretty simple.
If not, let me know

### DIFF
--- a/data/themes/classic/style.css
+++ b/data/themes/classic/style.css
@@ -149,9 +149,9 @@ PianoRoll {
 	qproperty-noteOpacity: 128;
 	qproperty-noteBorders: true; /* boolean property, set false to have borderless notes */
 	qproperty-selectedNoteColor: rgb( 0, 125, 255 );
-	qproperty-ghostNoteColor: #000000;
+	qproperty-ghostNoteColor: #1e1e1e;
 	qproperty-ghostNoteTextColor: #ffffff;
-	qproperty-ghostNoteOpacity: 50;
+	qproperty-ghostNoteOpacity: 60;
 	qproperty-ghostNoteBorders: true;
 	qproperty-barColor: #4afd85;
 	qproperty-markedSemitoneColor: rgba( 0, 255, 200, 60 );
@@ -159,13 +159,13 @@ PianoRoll {
 	qproperty-whiteKeyWidth: 64;
 	qproperty-whiteKeyActiveTextColor: #000;
 	qproperty-whiteKeyActiveTextShadow: rgb( 240, 240, 240 );
-	qproperty-whiteKeyActiveBackground: qlineargradient(x1:0, y1:0, x2:1, y2:0, stop:0 #43e97b, stop:1 #3bcd6c);
+	qproperty-whiteKeyActiveBackground: qlineargradient(x1:0, y1:0, x2:1, y2:0, stop:0 #3bcd6c, stop:1 #43e97b);
 	qproperty-whiteKeyInactiveTextColor: rgb( 128, 128, 128);
 	qproperty-whiteKeyInactiveTextShadow: rgb( 240, 240, 240 );
-	qproperty-whiteKeyInactiveBackground: qlineargradient(x1:0, y1:0, x2:1, y2:0, stop:0 #eeeeee, stop:1 #ffffff);
+	qproperty-whiteKeyInactiveBackground: qlineargradient(x1:0, y1:0, x2:1, y2:0, stop:0 #d6d6d6, stop:1 #e3e3e3);
 	qproperty-blackKeyWidth: 48;
-	qproperty-blackKeyActiveBackground: qlineargradient(x1:0, y1:0, x2:1, y2:0, stop:0 #43e97b, stop:1 #3bcd6c);
-	qproperty-blackKeyInactiveBackground: qlineargradient(x1:0, y1:0, x2:1, y2:0, stop:0 #333, stop:1 #000);
+	qproperty-blackKeyActiveBackground: qlineargradient(x1:0, y1:0, x2:1, y2:0, stop:0 #3bcd6c, stop:1 #43e97b);
+	qproperty-blackKeyInactiveBackground: qlineargradient(x1:0, y1:0, x2:1, y2:0, stop:0 #262626, stop:1 #313131);
 	/* Grid colors */
 	qproperty-lineColor: rgba( 128, 128, 128, 80 );
 	qproperty-beatLineColor: rgba( 128, 128, 128, 160 );
--- a/data/themes/classic/style.css
+++ b/data/themes/classic/style.css
@@ -153,19 +153,19 @@ PianoRoll {
 	qproperty-ghostNoteTextColor: #ffffff;
 	qproperty-ghostNoteOpacity: 60;
 	qproperty-ghostNoteBorders: true;
-	qproperty-barColor: #4afd85;
+	qproperty-barColor: rgb( 119, 199, 216 );
 	qproperty-markedSemitoneColor: rgba( 0, 255, 200, 60 );
 	/* Piano keys */
 	qproperty-whiteKeyWidth: 64;
 	qproperty-whiteKeyActiveTextColor: #000;
 	qproperty-whiteKeyActiveTextShadow: rgb( 240, 240, 240 );
-	qproperty-whiteKeyActiveBackground: qlineargradient(x1:0, y1:0, x2:1, y2:0, stop:0 #3bcd6c, stop:1 #43e97b);
+	qproperty-whiteKeyActiveBackground: qlineargradient(x1:0, y1:0, x2:1, y2:0, stop:0 #50b8c3, stop:1 #54c4d2);
 	qproperty-whiteKeyInactiveTextColor: rgb( 128, 128, 128);
 	qproperty-whiteKeyInactiveTextShadow: rgb( 240, 240, 240 );
-	qproperty-whiteKeyInactiveBackground: qlineargradient(x1:0, y1:0, x2:1, y2:0, stop:0 #d6d6d6, stop:1 #e3e3e3);
+	qproperty-whiteKeyInactiveBackground: qlineargradient(x1:0, y1:0, x2:1, y2:0, stop:0 #c5c5c5, stop:1 #f9f9f9);
 	qproperty-blackKeyWidth: 48;
-	qproperty-blackKeyActiveBackground: qlineargradient(x1:0, y1:0, x2:1, y2:0, stop:0 #3bcd6c, stop:1 #43e97b);
-	qproperty-blackKeyInactiveBackground: qlineargradient(x1:0, y1:0, x2:1, y2:0, stop:0 #262626, stop:1 #313131);
+	qproperty-blackKeyActiveBackground: qlineargradient(x1:0, y1:0, x2:1, y2:0, stop:0 #50b8c3, stop:1 #54c4d2);
+	qproperty-blackKeyInactiveBackground: qlineargradient(x1:0, y1:0, x2:1, y2:0, stop:0 #101010, stop:1 #3a3a3a);
 	/* Grid colors */
 	qproperty-lineColor: rgba( 128, 128, 128, 80 );
 	qproperty-beatLineColor: rgba( 128, 128, 128, 160 );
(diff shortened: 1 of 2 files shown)


russiankumar:
Pinging @RebeccaDeField for this PR. From what I know we aren't supposed to make style changes in the classic style, right? Apart from that, this looks good to me 👍

RoxasKH:
> 
> 
> Pinging @RebeccaDeField for this PR. From what I know we aren't supposed to make style changes in the classic style, right? Apart from that, this looks good to me 👍

Well my changes were mostly intended for the default theme, but as the keys are now dynamically drawn, that counted for classic, too, so i did the same changes to the classic css style file.

Here's how it changed

Before:
immagine

After:
immagine

Imo feels better for classic, too.

he29-net:
Basically, the Classic theme was already changed by the Piano Roll refactor. So this PR just brings the values closer to what it used before, right?

Or maybe we could simply pull the values from Classic theme bitmaps and plug them into the CSS, getting as close to the original style as possible, if needed.

RoxasKH:
> 
> 
> Basically, the Classic theme was already changed by the Piano Roll refactor. So this PR just brings the values closer to what it used before, right?
> 
> Or maybe we could simply pull the values from Classic theme bitmaps and plug them into the CSS, getting as close to the original style as possible, if needed.

The original image-based style of the keys in classic theme would be really difficult to reproduce in css, i'm not even sure if it's possible.

I've just changed gradients to be more eye wise and relaxing, using pure white and black tones was overkill ofr the eyes.

These are the original classic theme keys.

immagine

RebeccaDeField:
This looks good to me. Is there any way you can make sure all of your colors that you are editing within the pianoroll do not have any blue tints to the hues for both white and black keys (ie: a saturation of 0)? Otherwise, it looks good from a design perspective.

RoxasKH:
> 
> 
> This looks good to me. Is there any way you can make sure all of your colors that you are editing within the pianoroll do not have any blue tints to the hues for both white and black keys (ie: a saturation of 0)? Otherwise, it looks good from a design perspective.

I've just checked the colors in Gimp to be sure, and yeah, there's no saturation or blue tints, are just pure shade of greys.

Spekular:
The colors are in format #RRGGBB, so you can just check that each set of two letters is the same to ensure that there's no saturation. Checking the diff, the only changed colors with saturation are `qproperty-[black/white]KeyActiveBackground`. Both are mostly green, so I assume they're the colors used for playing notes.

RoxasKH:
> 
> 
> The colors are in format #RRGGBB, so you can just check that each set of two letters is the same to ensure that there's no saturation. Checking the diff, the only changed colors with saturation are `qproperty-[black/white]KeyActiveBackground`. Both are mostly green, so I assume they're the colors used for playing notes.

You're right i didn't think about it. Speaking of the thing you mentioned, yeah, those are the gradients used for active keys, so the greenlighted keys when they're played. I haven't touched those colors at all, i've just inverted the gradient as i did for the inactive keys, but those green values are the original ones.

RebeccaDeField:
My bad, the green is fine, I was only referring to the grays, whites and blacks :+1: 

As for the classic theme change, I have no objections to the improvement, and it seems arbitrary in the grand scheme of things, but it may be worth noting that "easy on the eyes" VS "good contrast" has been an ongoing battle and considering the very high contrast of some elements from the classic theme, this might be a reason people would switch to that instead of the default.

I understand the effort to convert it to being drawn instead of raster images, and it may be too much effort to recreate the previous version. For the sake of keeping things simple and consistent, I'd go ahead with it using higher contrast (if not just simply black and white) to hopefully cover most users' needs.

RoxasKH:
> 
> 
> My bad, the green is fine, I was only referring to the grays, whites and blacks 👍
> 
> As for the classic theme change, I have no objections to the improvement, and it seems arbitrary in the grand scheme of things, but it may be worth noting that "easy on the eyes" VS "good contrast" has been an ongoing battle and considering the very high contrast of some elements from the classic theme, this might be a reason people would switch to that instead of the default.
> 
> I understand the effort to convert it to being drawn instead of raster images, and it may be too much effort to recreate the previous version. For the sake of keeping things simple and consistent, I'd go ahead with it using higher contrast for the classic theme and as-is for the default to hopefully cover most users' needs.

Ok, so i should leave the contrast of classic theme higher?

I get your points, but high contrast is imo so not-design wise, that i still don't understand why some people like it.
The new default theme is way better, and i think the main reason for which people use the classic one is mostly because it's nearer to a light theme.

Maybe adding a light version of the actual default theme would be better, but this is probably something that should be discussed in the future.

For the moment, if you think higher contrast is better for the classic theme, i'm just gonna use again pure white and black for the cassic theme, but i'm still gonna reverse the gradient. It's not a problem, also considering my change was done mostly for the default one.

Spekular:
> > [...] it may be worth noting that "easy on the eyes" VS "good contrast" has been an ongoing battle and considering the very high contrast of some elements from the classic theme, this might be a reason people would switch to that instead of the default.

> I get your points, but high contrast is imo so not-design wise, that i still don't understand why some people like it.

@RoxasKH high contrast modes is an accessibility thing. The lower the contrast, the better eyesight required to tell things apart. Good design depends on what your goal is, and for the sake of maximum accessibility a high contrast theme is good. I believe René is saying that the classic theme currently serves as a stand in for an intentional/official "high contrast" mode.

RoxasKH:
> 
> 
> > > [...] it may be worth noting that "easy on the eyes" VS "good contrast" has been an ongoing battle and considering the very high contrast of some elements from the classic theme, this might be a reason people would switch to that instead of the default.
> 
> > I get your points, but high contrast is imo so not-design wise, that i still don't understand why some people like it.
> 
> @RoxasKH high contrast modes is an accessibility thing. The lower the contrast, the better eyesight required to tell things apart. Good design depends on what your goal is, and for the sake of maximum accessibility a high contrast theme is good. I believe René is saying that the classic theme currently serves as a stand in for an intentional/official "high contrast" mode.

Yeah i got it. Even if most pro DAWs doesn't have one and mostly everything that look modern atm has low contrast, i got it.
I wasn't against it infact, i was mostly asking if so i need to leave the higher contrast in the classic theme.

RebeccaDeField:
> Ok, so i should leave the contrast of classic theme higher?

Yeah, since it would be too complex to recreate the old piano, I figure it's good to capture the distinguishing part of it.

> I get your points, but high contrast is imo so not-design wise, that i still don't understand why some people like it.

Contrast is actually a great element to keep in mind and strive for, so long as we balance avoiding eye-fatigue too. I understand that part of your effort to tone it down a touch. I've seen that some users still use the classic theme, so while I try my best to balance contrast and eye fatigue in the default theme, the classic theme should be more or less maintained as-is.

> Maybe adding a light version of the actual default theme would be better, but this is probably something that should be discussed in the future.

I started the default theme as an alt theme option in the sharing platform. If anyone wants to mod the current theme to be lighter and picks up some steam on it, there can certainly be a conversation about bundling that then. Same goes for a modified super-high contrast version of what we currently have.

> For the moment, if you think higher contrast is better for the classic theme, i'm just gonna use again pure white and black for the cassic theme, but i'm still gonna reverse the gradient. It's not a problem, also considering my change was done mostly for the default one.

Sounds good :+1: 

> Yeah i got it. Even if most pro DAWs doesn't have one and mostly everything that look modern atm has low contrast, i got it.

My goal is to have the theme to be the most usable to the most people possible. Even if it becomes trendy to use low-contrast, I don't mind going against the norm for the sake of being usable to the widest range of people. That being said, people still use the classic theme so I try to keep in mind that there are things there benefiting some niche users and I don't want to override that theme with elements from the current one or it defeats the purpose of bundling both.

It did dawn on me now that by using a darker shade for white keys, we lose a bit of distinction between the active keys and non-active keys based purely on the value of the color. Let's bring the green down in value a bit to achieve something like so when desaturated.

**Before**
103139794-2c03d780-46e0-11eb-9394-1290a07bf7da

**After**
103139794-2c03d780-46e0-11eb-9394-1290a07bf7da1

In conclusion: high contrast version in classic theme, slightly darker green = good to go.

superpaik:
Just one thought regarding the ghost notes. To me the first impression was, because they end with a darker gray, that the note duration wasn't the whole beat, but a 1/64 (or some) less. I had to zoom in to see if I was right or wrong. Maybe, instead of going darker it should go to a whiter gray, like the "no" ghost notes.
Not estrictly related to ghost notes, but because this PR is about "pianoroll graphic adjustments" I notice that there has been a change in the notes name style in the piano roll. In v1.2 the Cs are different from the other notes, which I found useful. In this version, every note has the same color and the note names are slightly aligned differently (a bit odd, to me). You can see it in the attach images. I don't know if this have been done in purpose or it is a side effect.
imatge
imatge

RoxasKH:
> 
> 
> Just one thought regarding the ghost notes. To me the first impression was, because they end with a darker gray, that the note duration wasn't the whole beat, but a 1/64 (or some) less. I had to zoom in to see if I was right or wrong. Maybe, instead of going darker it should go to a whiter gray, like the "no" ghost notes.
> Not estrictly related to ghost notes, but because this PR is about "pianoroll graphic adjustments" I notice that there has been a change in the notes name style in the piano roll. In v1.2 the Cs are different from the other notes, which I found useful. In this version, every note has the same color and the note names are slightly aligned differently (a bit odd, to me). You can see it in the attach images. I don't know if this have been done in purpose or it is a side effect.
> imatge
> imatge

That's why i made it brighter than it was. Making it brighter than the color with transparency i'm not sure if it'd be so easy, considering how is setup the style file atm.

I just need to make the more contrat version for the classic theme and the green adjustments, but atm i have no time due to 3 exams.

allejok96:
Just a thought, but perhaps this is tied to the C++ code... It would be nice if the ghost notes and semitone markers where different shades of gray instead of being transparent, because now it gets confusing when combining them.

bild

Veratil:
> Just a thought, but perhaps this is tied to the C++ code... It would be nice if the ghost notes and semitone markers where different shades of gray instead of being transparent, because now it gets confusing when combining them.

These should be customizable through the `qproperty-ghostNoteColor` and `qproperty-markedSemitoneColor` properties in the theme's stylesheet. Their defaults are `#1e1e1e` and `rgba(255,255,255,30)`, respectively.

RoxasKH:
> 
> 
> Just a thought, but perhaps this is tied to the C++ code... It would be nice if the ghost notes and semitone markers where different shades of gray instead of being transparent, because now it gets confusing when combining them.

I personally disagree for a consistency reason. If it was decided that the new LMMS Theme for 1.2 had notes, with transparency, having solid ghost notes would be weird. Even if i agree that highlighting a scale sure make them more confusing.


> 
> 
> > Ok, so i should leave the contrast of classic theme higher?
> 
> Yeah, since it would be too complex to recreate the old piano, I figure it's good to capture the distinguishing part of it.
> 
> It did dawn on me now that by using a darker shade for white keys, we lose a bit of distinction between the active keys and non-active keys based purely on the value of the color. Let's bring the green down in value a bit to achieve something like so when desaturated.
> 
> In conclusion: high contrast version in classic theme, slightly darker green = good to go.

Ok, so this is my final result

Default modern theme (darker shade of green):
modern

Desaturated
ModernDesaturated

Classic theme (higher contrast, i also decided to change the shade of the active keys to light blue to match the note color)
Classic

If everything's ok, i'll commit the new changes


Also i found a big graphical bug again: if a project with ghost notes set it's loaded, ghost notes aren't displayed correctly

ModernBug

ClassicBug

RoxasKH:
I decided to make the velocity bar color of the same color of the notes in the classic theme, for consistency in the pianoroll, as that green didn't fit at all.

immagine

Now i'm gonna make a commit withe new css files, waiting for review.

RoxasKH:
I ask you to close this, because i used the wrong css file. I'm gonna remake it, using the proper file with the proper changes.

Veratil:
> 
> 
> I ask you to close this, because i used the wrong css file. I'm gonna remake it, using the proper file with the proper changes.

You always have the ability to close your own PRs, but I got it. 😃

RoxasKH:
> 
> 
> > I ask you to close this, because i used the wrong css file. I'm gonna remake it, using the proper file with the proper changes.
> 
> You always have the ability to close your own PRs, but I got it. 😃

Didn't know, sorry, i'm pretty new to github

PR #5865, opened 2020-12-29
Add `LcdFloatSpinBox` (extracted from microtonal PR)

he29-net:
This PR contains code for a new `LcdFloatSpinBox` widget, originally included in #5522.

As the name says, it is an LCD Spin Box that can be used for decimal numbers. It is a wrapper class, which instantiates two `LcdSpinBox`es, positions them next to each other and adds a decimal point in between. The whole and fractional part can be individually adjusted by a mouse, allowing easy coarse and fine adjustments.

The goal was to have only a "thin" wrapper and instantiate `LcdSpinBox`es to avoid redundancy and large changes to existing code, but in the end some parts had to be re-used in some form from the `LcdSpinBox` class. So there is a good chance it could be designed better, with even less redundancy, but it would likely require a substantial rewrite of the `LcdSpinBox` class.

### DIFF
--- a/include/LcdFloatSpinBox.h
+++ b/include/LcdFloatSpinBox.h
@@ -37,9 +37,8 @@ class LMMS_EXPORT LcdFloatSpinBox : public QWidget, public FloatModelView
 {
 	Q_OBJECT
 public:
-	LcdFloatSpinBox(int numWhole, int numFrac, QWidget* parent = nullptr, const QString& name = QString());
-	LcdFloatSpinBox(int numWhole, int numFrac, const QString& style, QWidget* parent = nullptr,
-		const QString& name = QString());
+	LcdFloatSpinBox(int numWhole, int numFrac, const QString& name = QString(), QWidget* parent = nullptr);
+	LcdFloatSpinBox(int numWhole, int numFrac, const QString& style, const QString& name, QWidget* parent = nullptr);
 
 	void modelChanged() override
 	{
--- a/include/LcdFloatSpinBox.h
+++ b/include/LcdFloatSpinBox.h
@@ -92,6 +92,6 @@ signals:
 
 };
 
-typedef FloatModel LcdFloatSpinBoxModel;
+using LcdFloatSpinBoxModel = FloatModel;
 
 #endif
--- a/include/LcdFloatSpinBox.h
+++ b/include/LcdFloatSpinBox.h
@@ -1,0 +1,97 @@
+/*
+ * LcdFloatSpinBox.h - class LcdFloatSpinBox (LcdSpinBox for floats)
+ *
+ * Copyright (c) 2005-2014 Tobias Doerffel <tobydox/at/users.sourceforge.net>
+ * Copyright (c) 2020 Martin Pavelek <he29.HS/at/gmail.com>
+ *
+ * This file is part of LMMS - https://lmms.io
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this program (see COPYING); if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA 02110-1301 USA.
+ *
+ */
+
+
+#ifndef LCD_FLOATSPINBOX_H
+#define LCD_FLOATSPINBOX_H
+
+#include <QString>
+
+#include "LcdWidget.h"
+#include "AutomatableModelView.h"
+
+
+class LMMS_EXPORT LcdFloatSpinBox : public QWidget, public FloatModelView
+{
+	Q_OBJECT
+public:
+	LcdFloatSpinBox(int numWhole, int numFrac, QWidget* parent, const QString& name = QString());
+	LcdFloatSpinBox(int numWhole, int numFrac, const QString& style, QWidget* parent, const QString& name = QString());
+	virtual ~LcdFloatSpinBox() = default;
+
+	void modelChanged() override
+	{
+		ModelView::modelChanged();
+		update();
+	}
+
+	/*! Sets an offset which is always added to value of model so we can
+	    display values in a user-friendly way if they internally start at 0 */
+	void setDisplayOffset(int offset)
+	{
+		m_displayOffset = offset;
+	}
+
+	/*! \brief Returns internal offset for displaying values */
+	int displayOffset() const
+	{
+		return m_displayOffset;
+	}
+
+	void setLabel(const QString &label) {m_label = label;}
+
+public slots:
+	virtual void update();
+
+protected:
+	void contextMenuEvent(QContextMenuEvent *me) override;
+	void mousePressEvent(QMouseEvent *me) override;
+	void mouseMoveEvent(QMouseEvent *me) override;
+	void mouseReleaseEvent(QMouseEvent *me) override;
+	void wheelEvent(QWheelEvent *we) override;
+	void mouseDoubleClickEvent(QMouseEvent *me) override;
+	void paintEvent(QPaintEvent *pe) override;
+
+private:
+	void layoutSetup(const QString &style = QString("19green"));
+	void enterValue();
+	float getStep();
+
+	LcdWidget m_wholeDisplay;
+	LcdWidget m_fractionDisplay;
+	bool m_mouseMoving;
+	bool m_intStep;
+	QPoint m_origMousePos;
+	int m_displayOffset;
+	QString m_label;
+
+signals:
+	void manualChange();
+
+};
+
+typedef FloatModel LcdFloatSpinBoxModel;
+
+#endif
--- a/include/LcdFloatSpinBox.h
+++ b/include/LcdFloatSpinBox.h
@@ -37,27 +37,14 @@ class LMMS_EXPORT LcdFloatSpinBox : public QWidget, public FloatModelView
 {
 	Q_OBJECT
 public:
-	LcdFloatSpinBox(int numWhole, int numFrac, QWidget* parent, const QString& name = QString());
-	LcdFloatSpinBox(int numWhole, int numFrac, const QString& style, QWidget* parent, const QString& name = QString());
-	virtual ~LcdFloatSpinBox() = default;
+	LcdFloatSpinBox(int numWhole, int numFrac, QWidget* parent = nullptr, const QString& name = QString());
+	LcdFloatSpinBox(int numWhole, int numFrac, const QString& style, QWidget* parent = nullptr,
+		const QString& name = QString());
 
 	void modelChanged() override
 	{
 		ModelView::modelChanged();
 		update();
-	}
-
-	/*! Sets an offset which is always added to value of model so we can
-	    display values in a user-friendly way if they internally start at 0 */
-	void setDisplayOffset(int offset)
-	{
-		m_displayOffset = offset;
-	}
-
-	/*! \brief Returns internal offset for displaying values */
-	int displayOffset() const
-	{
-		return m_displayOffset;
 	}
 
 	void setLabel(const QString &label) { m_label = label; }
@@ -77,7 +64,7 @@ protected:
 private:
 	void layoutSetup(const QString &style = QString("19green"));
 	void enterValue();
-	float getStep();
+	float getStep() const;
 
 	LcdWidget m_wholeDisplay;
 	LcdWidget m_fractionDisplay;
--- a/include/LcdFloatSpinBox.h
+++ b/include/LcdFloatSpinBox.h
@@ -60,7 +60,7 @@ public:
 		return m_displayOffset;
 	}
 
-	void setLabel(const QString &label) {m_label = label;}
+	void setLabel(const QString &label) { m_label = label; }
 
 public slots:
 	virtual void update();
--- a/include/LcdWidget.h
+++ b/include/LcdWidget.h
@@ -40,7 +40,7 @@ class LMMS_EXPORT LcdWidget : public QWidget
 	Q_PROPERTY( QColor textShadowColor READ textShadowColor WRITE setTextShadowColor )
 	
 public:
-	LcdWidget(QWidget* parent, const QString& name = QString(), bool leadingZero = false);
+	explicit LcdWidget(QWidget* parent, const QString& name = QString(), bool leadingZero = false);
 	LcdWidget(int numDigits, QWidget* parent, const QString& name = QString(), bool leadingZero = false);
 	LcdWidget(int numDigits, const QString& style, QWidget* parent, const QString& name = QString(),
 		bool leadingZero = false);
@@ -104,8 +104,8 @@ private:
 	int m_cellHeight;
 	int m_numDigits;
 	int m_marginWidth;
-	int m_seamlessLeft;
-	int m_seamlessRight;
+	bool m_seamlessLeft;
+	bool m_seamlessRight;
 	bool m_leadingZero;
 
 	void initUi( const QString& name, const QString &style ); //!< to be called by ctors
--- a/include/LcdWidget.h
+++ b/include/LcdWidget.h
@@ -40,9 +40,10 @@ class LMMS_EXPORT LcdWidget : public QWidget
 	Q_PROPERTY( QColor textShadowColor READ textShadowColor WRITE setTextShadowColor )
 	
 public:
-	LcdWidget( QWidget* parent, const QString& name = QString() );
-	LcdWidget( int numDigits, QWidget* parent, const QString& name = QString() );
-	LcdWidget( int numDigits, const QString& style, QWidget* parent, const QString& name = QString() );
+	LcdWidget( QWidget* parent, const QString& name = QString(), bool leadingZero = false );
+	LcdWidget( int numDigits, QWidget* parent, const QString& name = QString(), bool leadingZero = false );
+	LcdWidget( int numDigits, const QString& style, QWidget* parent, const QString& name = QString(),
+		bool leadingZero = false );
 
 	virtual ~LcdWidget();
 
@@ -66,6 +67,10 @@ public:
 	QColor textShadowColor() const;
 	void setTextShadowColor( const QColor & c );
 
+	int cellHeight() const { return m_cellHeight; }
+
+	void setSeamless(bool left, bool right) {m_seamlessLeft = left; m_seamlessRight = right; updateSize();}
+
 public slots:
 	virtual void setMarginWidth( int width );
 
@@ -74,11 +79,6 @@ protected:
 	void paintEvent( QPaintEvent * pe ) override;
 
 	virtual void updateSize();
-
-	int cellHeight() const
-	{
-		return m_cellHeight;
-	}
 
 
 private:
@@ -99,9 +99,12 @@ private:
 	int m_cellHeight;
 	int m_numDigits;
 	int m_marginWidth;
+	int m_seamlessLeft;
+	int m_seamlessRight;
+	bool m_leadingZero;
 
 	void initUi( const QString& name, const QString &style ); //!< to be called by ctors
 
-} ;
+};
 
 #endif
--- a/include/LcdWidget.h
+++ b/include/LcdWidget.h
@@ -40,10 +40,10 @@ class LMMS_EXPORT LcdWidget : public QWidget
 	Q_PROPERTY( QColor textShadowColor READ textShadowColor WRITE setTextShadowColor )
 	
 public:
-	LcdWidget( QWidget* parent, const QString& name = QString(), bool leadingZero = false );
-	LcdWidget( int numDigits, QWidget* parent, const QString& name = QString(), bool leadingZero = false );
-	LcdWidget( int numDigits, const QString& style, QWidget* parent, const QString& name = QString(),
-		bool leadingZero = false );
+	LcdWidget(QWidget* parent, const QString& name = QString(), bool leadingZero = false);
+	LcdWidget(int numDigits, QWidget* parent, const QString& name = QString(), bool leadingZero = false);
+	LcdWidget(int numDigits, const QString& style, QWidget* parent, const QString& name = QString(),
+		bool leadingZero = false);
 
 	virtual ~LcdWidget();
 
@@ -69,7 +69,12 @@ public:
 
 	int cellHeight() const { return m_cellHeight; }
 
-	void setSeamless(bool left, bool right) {m_seamlessLeft = left; m_seamlessRight = right; updateSize();}
+	void setSeamless(bool left, bool right)
+	{
+		m_seamlessLeft = left;
+		m_seamlessRight = right;
+		updateSize();
+	}
 
 public slots:
 	virtual void setMarginWidth( int width );
--- a/src/gui/widgets/LcdFloatSpinBox.cpp
+++ b/src/gui/widgets/LcdFloatSpinBox.cpp
@@ -78,7 +78,7 @@ void LcdFloatSpinBox::layoutSetup(const QString &style)
 	m_wholeDisplay.setSeamless(false, true);
 	m_fractionDisplay.setSeamless(true, false);
 
-	lcdLayout->addWidget(&m_wholeDisplay);  
+	lcdLayout->addWidget(&m_wholeDisplay);
 
 	QLabel *dotLabel = new QLabel("", this);
 	QPixmap *dotPixmap = new QPixmap(embed::getIconPixmap(QString("lcd_" + style + "_dot").toUtf8().constData()));
--- a/src/gui/widgets/LcdFloatSpinBox.cpp
+++ b/src/gui/widgets/LcdFloatSpinBox.cpp
@@ -27,6 +27,7 @@
 #include "LcdFloatSpinBox.h"
 
 #include <cmath>
+
 #include <QApplication>
 #include <QHBoxLayout>
 #include <QInputDialog>
@@ -81,8 +82,8 @@ void LcdFloatSpinBox::layoutSetup(const QString &style)
 	lcdLayout->addWidget(&m_wholeDisplay);
 
 	QLabel *dotLabel = new QLabel("", this);
-	QPixmap *dotPixmap = new QPixmap(embed::getIconPixmap(QString("lcd_" + style + "_dot").toUtf8().constData()));
-	dotLabel->setPixmap(dotPixmap->copy(0, 0, dotPixmap->size().width(), dotPixmap->size().height() / 2));
+	QPixmap dotPixmap(embed::getIconPixmap(QString("lcd_" + style + "_dot").toUtf8().constData()));
+	dotLabel->setPixmap(dotPixmap.copy(0, 0, dotPixmap.size().width(), dotPixmap.size().height() / 2));
 	lcdLayout->addWidget(dotLabel);
 
 	lcdLayout->addWidget(&m_fractionDisplay);
@@ -162,7 +163,7 @@ void LcdFloatSpinBox::mouseMoveEvent(QMouseEvent* event)
 		if (gui->mainWindow()->isShiftPressed()) { dy = qBound(-4, dy/4, 4); }
 		if (dy > 1 || dy < -1)
 		{
-			model()->setInitValue(model()->value() - dy / 2 * getStep());
+			model()->setValue(model()->value() - dy / 2 * getStep());
 			emit manualChange();
 			QCursor::setPos(m_origMousePos);
 		}
@@ -191,7 +192,7 @@ void LcdFloatSpinBox::wheelEvent(QWheelEvent *event)
 	else { m_intStep = false; }
 
 	event->accept();
-	model()->setInitValue(model()->value() + ((event->delta() > 0) ? 1 : -1) * getStep());
+	model()->setValue(model()->value() + ((event->delta() > 0) ? 1 : -1) * getStep());
 	emit manualChange();
 }
 
--- a/src/gui/widgets/LcdFloatSpinBox.cpp
+++ b/src/gui/widgets/LcdFloatSpinBox.cpp
@@ -103,7 +103,7 @@ void LcdFloatSpinBox::layoutSetup(const QString &style)
 void LcdFloatSpinBox::update()
 {
 	const int whole = static_cast<int>(model()->value());
-	const float fraction = model()->value() - static_cast<int>(model()->value());
+	const float fraction = model()->value() - whole;
 	const int intFraction = fraction * std::pow(10.f, m_fractionDisplay.numDigits());
 	m_wholeDisplay.setValue(whole + m_displayOffset);
 	m_fractionDisplay.setValue(intFraction + m_displayOffset);
@@ -159,8 +159,7 @@ void LcdFloatSpinBox::mouseMoveEvent(QMouseEvent* event)
 	if (m_mouseMoving)
 	{
 		int dy = event->globalY() - m_origMousePos.y();
-		if (gui->mainWindow()->isShiftPressed())
-			dy = qBound(-4, dy/4, 4);
+		if (gui->mainWindow()->isShiftPressed()) { dy = qBound(-4, dy/4, 4); }
 		if (dy > 1 || dy < -1)
 		{
 			model()->setInitValue(model()->value() - dy / 2 * getStep());
@@ -211,8 +210,8 @@ void LcdFloatSpinBox::enterValue()
 	new_val = QInputDialog::getDouble(
 			this, tr("Set value"),
 			tr("Please enter a new value between %1 and %2:").
-			arg(model()->minValue()).
-			arg(model()->maxValue()),
+				arg(model()->minValue()).
+				arg(model()->maxValue()),
 			model()->value(),
 			model()->minValue(),
 			model()->maxValue(),
--- a/src/gui/widgets/LcdFloatSpinBox.cpp
+++ b/src/gui/widgets/LcdFloatSpinBox.cpp
@@ -1,0 +1,255 @@
+/*
+ * LcdFloatSpinBox.cpp - class LcdFloatSpinBox (LcdSpinBox for floats)
+ *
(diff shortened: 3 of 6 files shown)


include/LcdFloatSpinBox.h
```diff
@@ -0,0 +1,97 @@
...
+		m_displayOffset = offset;
+	}
+
+	/*! \brief Returns internal offset for displaying values */
+	int displayOffset() const
+	{
+		return m_displayOffset;
+	}
+
+	void setLabel(const QString &label) {m_label = label;}
```
IanCaio, line 63:
```suggestion
	void setLabel(const QString &label) { m_label = label; }
```

include/LcdWidget.h
```diff
@@ -40,9 +40,10 @@ class LMMS_EXPORT LcdWidget : public QWidget
 	Q_PROPERTY( QColor textShadowColor READ textShadowColor WRITE setTextShadowColor )
 	
 public:
-	LcdWidget( QWidget* parent, const QString& name = QString() );
-	LcdWidget( int numDigits, QWidget* parent, const QString& name = QString() );
-	LcdWidget( int numDigits, const QString& style, QWidget* parent, const QString& name = QString() );
+	LcdWidget( QWidget* parent, const QString& name = QString(), bool leadingZero = false );
+	LcdWidget( int numDigits, QWidget* parent, const QString& name = QString(), bool leadingZero = false );
+	LcdWidget( int numDigits, const QString& style, QWidget* parent, const QString& name = QString(),
+		bool leadingZero = false );
```
IanCaio, line 46:
```suggestion
	LcdWidget(QWidget* parent, const QString& name = QString(), bool leadingZero = false);
	LcdWidget(int numDigits, QWidget* parent, const QString& name = QString(), bool leadingZero = false);
	LcdWidget(int numDigits, const QString& style, QWidget* parent, const QString& name = QString(),
		bool leadingZero = false);
```

include/LcdWidget.h
```diff
@@ -66,6 +67,10 @@ class LMMS_EXPORT LcdWidget : public QWidget
 	QColor textShadowColor() const;
 	void setTextShadowColor( const QColor & c );
 
+	int cellHeight() const { return m_cellHeight; }
+
+	void setSeamless(bool left, bool right) {m_seamlessLeft = left; m_seamlessRight = right; updateSize();}
```
IanCaio, line 72:
```suggestion
	void setSeamless(bool left, bool right)
	{
		m_seamlessLeft = left;
		m_seamlessRight = right;
		updateSize();
	}
```

src/gui/widgets/LcdFloatSpinBox.cpp
```diff
@@ -0,0 +1,255 @@
...
+	outerLayout->setContentsMargins(0, 0, 0, 0);
+	outerLayout->setSizeConstraint(QLayout::SetFixedSize);
+	this->setLayout(outerLayout);
+}
+
+
+void LcdFloatSpinBox::update()
+{
+	const int whole = static_cast<int>(model()->value());
+	const float fraction = model()->value() - static_cast<int>(model()->value());
```
IanCaio, line 106:
```suggestion
	const float fraction = model()->value() - whole;
```

Compiler might optimize it the way it's written but if not this saves a cast and a method call.

src/gui/widgets/LcdFloatSpinBox.cpp
```diff
@@ -0,0 +1,255 @@
...
+		FloatModelView::mousePressEvent(event);
+	}
+}
+
+
+void LcdFloatSpinBox::mouseMoveEvent(QMouseEvent* event)
+{
+	// switch between integer and fractional step based on cursor position
+	if (event->x() < m_wholeDisplay.width()) {m_intStep = true;}
+	else {m_intStep = false;}
```
IanCaio, line 157:
```suggestion
	if (event->x() < m_wholeDisplay.width()) { m_intStep = true; }
	else { m_intStep = false; }
```

src/gui/widgets/LcdFloatSpinBox.cpp
```diff
@@ -0,0 +1,255 @@
...
+
+	if (m_mouseMoving)
+	{
+		int dy = event->globalY() - m_origMousePos.y();
+		if (gui->mainWindow()->isShiftPressed())
+			dy = qBound(-4, dy/4, 4);
```
IanCaio, line 163:
```suggestion
		if (gui->mainWindow()->isShiftPressed()) { dy = qBound(-4, dy/4, 4); }
```

Even if multiline `if` is preferred, explicit blocking must be used.

src/gui/widgets/LcdFloatSpinBox.cpp
```diff
@@ -0,0 +1,255 @@
...
+		m_mouseMoving = false;
+	}
+}
+
+
+void LcdFloatSpinBox::wheelEvent(QWheelEvent *event)
+{
+	// switch between integer and fractional step based on cursor position
+	if (event->x() < m_wholeDisplay.width()) {m_intStep = true;}
+	else {m_intStep = false;}
```
IanCaio, line 192:
```suggestion
	if (event->x() < m_wholeDisplay.width()) { m_intStep = true; }
	else { m_intStep = false; }
```

src/gui/widgets/LcdFloatSpinBox.cpp
```diff
@@ -0,0 +1,255 @@
...
+void LcdFloatSpinBox::enterValue()
+{
+	bool ok;
+	float new_val;
+
+	new_val = QInputDialog::getDouble(
+			this, tr("Set value"),
+			tr("Please enter a new value between %1 and %2:").
+			arg(model()->minValue()).
+			arg(model()->maxValue()),
```
IanCaio, line 215:
```suggestion
			tr("Please enter a new value between %1 and %2:").
				arg(model()->minValue()).
				arg(model()->maxValue()),
```

Just a suggestion to make it clearer that those are part of `tr` and not arguments to the `getDouble` call.

src/gui/widgets/LcdFloatSpinBox.cpp
```diff
@@ -0,0 +1,255 @@
...
+	{
+		model()->setValue(new_val);
+	}
+}
+
+
+float LcdFloatSpinBox::getStep()
+{
+	if (m_intStep) {return 1;}
+	else {return model()->step<float>();}
```
IanCaio, line 231:
```suggestion
	if (m_intStep) { return 1; }
	else { return model()->step<float>(); }
```

src/gui/widgets/LcdWidget.cpp
```diff
@@ -40,28 +40,31 @@
 
 
 
-LcdWidget::LcdWidget( QWidget* parent, const QString& name ) :
-	LcdWidget( 1, parent, name )
+LcdWidget::LcdWidget( QWidget* parent, const QString& name, bool leadingZero ) :
+	LcdWidget( 1, parent, name, leadingZero )
```
IanCaio, line 44:
```suggestion
LcdWidget::LcdWidget(QWidget* parent, const QString& name, bool leadingZero) :
	LcdWidget(1, parent, name, leadingZero)
```

src/gui/widgets/LcdWidget.cpp
```diff
@@ -40,28 +40,31 @@
...
 {
 }
 
 
 
 
-LcdWidget::LcdWidget( int numDigits, QWidget* parent, const QString& name ) :
-	LcdWidget( numDigits, QString("19green"), parent, name )
+LcdWidget::LcdWidget( int numDigits, QWidget* parent, const QString& name, bool leadingZero ) :
+	LcdWidget( numDigits, QString("19green"), parent, name, leadingZero )
```
IanCaio, line 52:
```suggestion
LcdWidget::LcdWidget(int numDigits, QWidget* parent, const QString& name, bool leadingZero) :
	LcdWidget(numDigits, QString("19green"), parent, name, leadingZero)
```

src/gui/widgets/LcdWidget.cpp
```diff
@@ -40,28 +40,31 @@
...
 {
 }
 
 
 
 
-LcdWidget::LcdWidget( int numDigits, const QString& style, QWidget* parent, const QString& name ) :
+LcdWidget::LcdWidget( int numDigits, const QString& style, QWidget* parent, const QString& name, bool leadingZero ) :
```
IanCaio, line 59:
```suggestion
LcdWidget::LcdWidget(int numDigits, const QString& style, QWidget* parent, const QString& name, bool leadingZero) :
```

src/gui/widgets/LcdWidget.cpp
```diff
@@ -40,28 +40,31 @@
...
 	QWidget( parent ),
 	m_label(),
 	m_textColor( 255, 255, 255 ),
 	m_textShadowColor( 64, 64, 64 ),
-	m_numDigits( numDigits )
+	m_numDigits( numDigits ),
+	m_seamlessLeft( false ),
+	m_seamlessRight( false ),
+	m_leadingZero( leadingZero )
```
IanCaio, line 67:
```suggestion
	m_numDigits(numDigits),
	m_seamlessLeft(false),
	m_seamlessRight(false),
	m_leadingZero(leadingZero)
```

src/gui/widgets/LcdWidget.cpp
```diff
@@ -77,19 +80,15 @@ LcdWidget::~LcdWidget()
...
 	{
-		s = QString::number( value );
-		// TODO: if pad == true
-		/*
-		while( (int) s.length() < m_numDigits )
-		{
-			s = "0" + s;
+		s = QString::number(value);
+		if (m_leadingZero) {
+			s = s.rightJustified(m_numDigits, '0');
```
IanCaio, line 90:
```suggestion
		if (m_leadingZero)
		{
			s = s.rightJustified(m_numDigits, '0');
```

IanCaio (changes requested):
Except for some code style to be fixed (mostly code block spacing, parenthesis spacing and a `if` without explicit blocks) and one line that could reuse a variable it looks good to me. I've to look at the paint events with more care, since I don't yet understand them completely. No obvious problems were noticed on a first review of them though.

I left the changes as suggestions so they can be batch applied if you want!

he29-net:
Thanks for the cleaning pass. A lot of the code was copied from the original `LcdSpinBox`, and since the clang-format was always going to happen "almost tomorrow" for the last year, I did not bother re-formatting the copied code. :))

As for the remaining suggestions, I posted a question on Discord and I will apply or ignore them depending on what happens with the weird one-liner spacing rule.

include/LcdFloatSpinBox.h
```diff
@@ -0,0 +1,97 @@
...
+#include "AutomatableModelView.h"
+
+
+class LMMS_EXPORT LcdFloatSpinBox : public QWidget, public FloatModelView
+{
+	Q_OBJECT
+public:
+	LcdFloatSpinBox(int numWhole, int numFrac, QWidget* parent, const QString& name = QString());
+	LcdFloatSpinBox(int numWhole, int numFrac, const QString& style, QWidget* parent, const QString& name = QString());
+	virtual ~LcdFloatSpinBox() = default;
```
DomClark, line 42: This destructor can be omitted.

PhysSong: I think it's okay to leave this explicitly.
@DomClark BTW, where are the rest of your reviews?

src/gui/widgets/LcdFloatSpinBox.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+void LcdFloatSpinBox::layoutSetup(const QString &style)
+{
+	// Assemble the LCD parts
+	QHBoxLayout *lcdLayout = new QHBoxLayout();
+
+	m_wholeDisplay.setSeamless(false, true);
+	m_fractionDisplay.setSeamless(true, false);
+
+	lcdLayout->addWidget(&m_wholeDisplay);
```
PhysSong, line 81: Trailing whitespaces:
```suggestion
	lcdLayout->addWidget(&m_wholeDisplay);
```

include/LcdFloatSpinBox.h
```diff
@@ -0,0 +1,97 @@
...
+	/*! \brief Returns internal offset for displaying values */
+	int displayOffset() const
+	{
+		return m_displayOffset;
+	}
+
+	void setLabel(const QString &label) { m_label = label; }
+
+public slots:
+	virtual void update();
```
DomClark, line 66: Can this have a name other than `update`? The base class function isn't virtual, so this function shadows it, not overrides it. I don't know how this interacts with the signal-slot mechanism, but it may not behave as the implementation/callers intended. Also, it needn't be marked virtual, as it doesn't look like it's intended to be overridden.

DomClark (commented):
From what I can tell, `LcdFloatSpinBox` is mostly copied from `LcdSpinBox`, so a lot of these comments apply to that class too. Those comment could be considered out of scope, since the issues referred to aren't introduced by this PR, and could be fixed at the same time as the original class.

include/LcdFloatSpinBox.h
DomClark: I agree it's fine to leave it explicitly - it will behave exactly the same as if omitted - but why add code that does nothing?

> where are the rest of your reviews?

Out now, only two and a half weeks later!

include/LcdFloatSpinBox.h
```diff
@@ -0,0 +1,97 @@
...
+	QPoint m_origMousePos;
+	int m_displayOffset;
+	QString m_label;
+
+signals:
+	void manualChange();
+
+};
+
+typedef FloatModel LcdFloatSpinBoxModel;
```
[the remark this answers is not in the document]
he29-net: Oh, I wasn't even aware that `using` can be (or is supposed to be) used for types like this. I only ever saw it used to avoid repeatedly typing the namespace in front of something. I will probably forget about it soon since `typedef` is more self-explanatory and thus easier to keep in mind, but thanks for the info. :)) LMMS development is now probably the only place where I keep learning about new features and trends in C++.

include/LcdFloatSpinBox.h
he29-net: I will remove it, it only confuses people like me, who are not completely sure if it can be safely removed or it serves some unknown arcane purpose. :D

include/LcdFloatSpinBox.h
```diff
@@ -0,0 +1,97 @@
...
+#include "LcdWidget.h"
+#include "AutomatableModelView.h"
+
+
+class LMMS_EXPORT LcdFloatSpinBox : public QWidget, public FloatModelView
+{
+	Q_OBJECT
+public:
+	LcdFloatSpinBox(int numWhole, int numFrac, QWidget* parent, const QString& name = QString());
+	LcdFloatSpinBox(int numWhole, int numFrac, const QString& style, QWidget* parent, const QString& name = QString());
```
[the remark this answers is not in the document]
he29-net: Seems like a good idea to me. I was first a little worried about making the order of parameters inconsistent with `LcdSpinBox` (to avoid possible mistakes when converting `LcdSpinBox` to `LcdFloatSpinBox`), but since there is already one extra parameter and it is not something that is going to be done often, I think it makes sense to do the change.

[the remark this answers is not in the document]
he29-net: Hm, never mind, it probably wouldn't work well with the overloads. If changed to:
```cpp
LcdFloatSpinBox(int numWhole, int numFrac, const QString& name = QString(), QWidget* parent = nullptr);
LcdFloatSpinBox(int numWhole, int numFrac, const QString& style, const QString& name = QString(), QWidget* parent = nullptr);
```
there would be no way to tell whether `LcdFloatSpinBox(2, 3, "abcd")` is a call to create an instance with the name "abcd" and default parent, or an instance with style = "abcd" and default name and parent.

`grep`ing through the code, I don't actually see any call to `LcdSpinBox()` that would use the style definition, and `initUI()` in `LcdWidget` looks like the implementation is kind of half unfinished:
```cpp
 // We should make a factory for these or something.
 //m_lcdPixmap = new QPixmap( embed::getIconPixmap( QString( "lcd_" + style ).toUtf8().constData() ) );
 //m_lcdPixmap = new QPixmap( embed::getIconPixmap( "lcd_19green" ) ); // TODO!!
 m_lcdPixmap = new QPixmap( embed::getIconPixmap( QString( "lcd_" + style ).toUtf8().constData() ) );
```
(I'm not quite sure what the poet meant to say by these comments..)

So the question is whether we should even bother keeping the constructor with custom style support. I would hope that LMMS eventually moves away from bitmaps and the LCD will be drawn using a style defined in the CSS, so it would not be needed anyway.

include/LcdFloatSpinBox.h
```diff
@@ -0,0 +1,97 @@
...
+	void setDisplayOffset(int offset)
+	{
+		m_displayOffset = offset;
+	}
+
+	/*! \brief Returns internal offset for displaying values */
+	int displayOffset() const
+	{
+		return m_displayOffset;
+	}
```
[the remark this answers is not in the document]
he29-net: Probably not. I can only see it used for instrument MIDI settings with offset = 1, to hide the fact that channels and notes are indexed form zero. Since floats are not used for indexing, it probably makes not sense to keep this, so I'll remove it.

src/gui/widgets/LcdFloatSpinBox.cpp
```diff
@@ -0,0 +1,254 @@
...
+	FloatModelView(new FloatModel(0, 0, 0, 0, nullptr, name, true), this),
+	m_wholeDisplay(numWhole, parent, name, false),
+	m_fractionDisplay(numFrac, parent, name, true),
+	m_mouseMoving(false),
+	m_intStep(false),
+	m_origMousePos(),
+	m_displayOffset(0)
+{
+	layoutSetup();
+}
```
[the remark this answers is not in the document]
he29-net: Again running into problems with the `style` parameter: if I delegate it to the other constructor, I have to pass something to the `style`, meaning that the default value defined in `LcdWidget` won't be used. I could copy it here, but then it would be in two different places that have to be kept in mind when the default changes.

include/LcdFloatSpinBox.h
he29-net: I did not even think of any of this, I just copied it from `LcdSpinBox` and assumed it will work the same. I would guess it still needs to be called `update()` since that is what Qt calls when it plans to redraw a widget?

src/gui/widgets/LcdFloatSpinBox.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+
+void LcdFloatSpinBox::contextMenuEvent(QContextMenuEvent* event)
+{
+	// for the case, the user clicked right while pressing left mouse-
+	// button, the context-menu appears while mouse-cursor is still hidden
+	// and it isn't shown again until user does something which causes
+	// an QApplication::restoreOverrideCursor()-call...
+	mouseReleaseEvent(nullptr);
```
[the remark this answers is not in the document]
he29-net: I'm not even sure this is needed anymore. I just tested several LMMS versions and since 1.2.2 the cursor is not being hidden anymore when changing spinbox or knob values by dragging. IIRC, this may have something to do with changes required for MacOS, which severely limited access to mouse cursor in some recent version?

If that is the case, I would simply remove this, and also the `QApplication::setOverrideCursor(Qt::BlankCursor)` call in `mousePressEvent()`, since it does not appear to be doing anything anyway (at least not on my machine).

include/LcdFloatSpinBox.h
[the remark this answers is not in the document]
PhysSong: One possible solution to resolve ambiguity:
```cpp
LcdFloatSpinBox(int numWhole, int numFrac, const QString& name = QString(), QWidget* parent = nullptr);
LcdFloatSpinBox(int numWhole, int numFrac, const QString& style, const QString& name, QWidget* parent = nullptr);
```

src/gui/widgets/LcdFloatSpinBox.cpp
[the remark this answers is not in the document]
PhysSong: > since 1.2.2 the cursor is not being hidden anymore when changing spinbox or knob values by dragging. IIRC, this may have something to do with changes required for MacOS, which severely limited access to mouse cursor in some recent version?

Yes, you may remove cursor stuff from spin boxes.

include/LcdFloatSpinBox.h
PhysSong: Since `QWidget::update()` is non-virtual, the name of this function doesn't matter.

DomClark: I know it doesn't matter, but I think the name is misleading, shadowing `QWidget::update` with something different. Also, slot lookups are done using strings, not vtables, so it's still not clear how this interacts with the signal-slot mechanism. I'll investigate it.

PhysSong: To clarify, I was answering to @he29-net's question.

include/LcdFloatSpinBox.h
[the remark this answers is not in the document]
DomClark: I guess this always tends to happen when you have a lot of default arguments. But I don't think the current approach of using the position of the parent argument to disambiguate these is particularly intuitive. I like PhysSong's solution for now, but it does look like we're moving away from pixmaps over time.

include/LcdFloatSpinBox.h
DomClark: Ah, that makes sense 😄. It's late and I'm not reading properly. I still want to investigate how slot name shadowing behaves though.

PhysSong: If I remember correctly, Qt looks at the meta object of the target class for string-based(`SIGNAL()`, `SLOT()` macros) connections. You can find something like `src/lmmsobjs_autogen/DGKZTCOIDE/moc_LcdSpinBox.cpp` in your build directory which contains signal/slot information.

src/gui/widgets/LcdFloatSpinBox.cpp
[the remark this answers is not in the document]
he29-net: I removed the overrides (appearance and position) from both `LcdFloatSpinBox` and `LcdSpinBox`. Seems to be working the same as before, but would not bet my life on my desktop environment (before the 1.2.2 change, everything I grabbed in LMMS only moved _up_, regardless of the actual mouse movement) so it probably should be tested by someone else as well.

include/LcdFloatSpinBox.h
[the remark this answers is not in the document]
he29-net: Changed to the format suggested by PhysSong.

src/gui/widgets/LcdWidget.cpp
```diff
@@ -177,21 +185,24 @@ void LcdWidget::paintEvent( QPaintEvent* )
 	}
 
 	// Right Margin
-	p.drawPixmap( QRect( 0, 0, m_marginWidth-1, m_cellHeight ), *m_lcdPixmap, 
-			QRect( charsPerPixmap*m_cellWidth, isEnabled()?0:m_cellHeight,
-				m_cellWidth / 2, m_cellHeight ) );
+	p.drawPixmap(QRect(0, 0, m_seamlessRight ? m_marginWidth : m_marginWidth - 1, m_cellHeight),
```
IanCaio, line 188: Just a question, why the `m_seamlessRight` draws a rect with width of `m_marginWidth` and the one that isn't seamless draws a rect with width `m_marginWidth - 1`?

I was expecting it to be width `0` for seamless and `m_marginWidth - 1` for not seamless.

IanCaio (changes requested):
Most of the code looks good, but there are some things that I think are worth changing, besides the `update` method name that @DomClark was concerned about.

I also left two comments about things I had doubt about (use of `explicit` and right margin drawing).

include/LcdFloatSpinBox.h
he29-net: I just tried to rename it and the display GUI stopped updating. So Qt (or LMMS, not sure who calls the global `update()` on all widgets) definitely expects the slot to be named `update()`.

include/LcdWidget.h
```diff
@@ -40,9 +40,10 @@ class LMMS_EXPORT LcdWidget : public QWidget
 	Q_PROPERTY( QColor textShadowColor READ textShadowColor WRITE setTextShadowColor )
 	
 public:
-	LcdWidget( QWidget* parent, const QString& name = QString() );
-	LcdWidget( int numDigits, QWidget* parent, const QString& name = QString() );
-	LcdWidget( int numDigits, const QString& style, QWidget* parent, const QString& name = QString() );
+	explicit LcdWidget(QWidget* parent, const QString& name = QString(), bool leadingZero = false);
```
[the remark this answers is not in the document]
he29-net: https://github.com/LMMS/lmms/pull/5865#discussion_r551637580

src/gui/widgets/LcdFloatSpinBox.cpp
```diff
@@ -0,0 +1,243 @@
...
+	{
+		int dy = event->globalY() - m_origMousePos.y();
+		if (gui->mainWindow()->isShiftPressed()) { dy = qBound(-4, dy/4, 4); }
+		if (dy > 1 || dy < -1)
+		{
+			model()->setValue(model()->value() - dy / 2 * getStep());
+			emit manualChange();
+		}
+	}
+}
```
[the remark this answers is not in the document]
he29-net: Thanks, that makes sense. I follow the development quite loosely, so I did not notice the `LcdSpinBox` received changes since the Microtuner PR was opened. And since the float version was based on a copy, the recent merges did not really bring any of the updates.

src/gui/widgets/LcdSpinBox.cpp
```diff
@@ -70,17 +70,11 @@ void LcdSpinBox::update()
...
-	// and it isn't shown again until user does something which causes
-	// an QApplication::restoreOverrideCursor()-call...
-	mouseReleaseEvent( NULL );
-
-	CaptionMenu contextMenu( model()->displayName() );
-	addDefaultActions( &contextMenu );
-	contextMenu.exec( QCursor::pos() );
+	CaptionMenu contextMenu(model()->displayName());
+	addDefaultActions(&contextMenu);
+	contextMenu.exec(QCursor::pos());
```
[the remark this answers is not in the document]
he29-net: https://github.com/LMMS/lmms/pull/5865#discussion_r567437183

src/gui/widgets/LcdWidget.cpp
he29-net: Frankly I have no idea. I probably tweaked it empirically back at the time, i.e. tried different values until it worked correctly. I can try to build the original Microtuner PR and try different values to see what happens if you want.

src/gui/widgets/LcdWidget.cpp
```diff
@@ -177,21 +185,24 @@ void LcdWidget::paintEvent( QPaintEvent* )
...
-	opt.state = QStyle::State_Sunken;
-	opt.rect = QRect( 0, 0, m_cellWidth * m_numDigits + (margin+m_marginWidth)*2 - 1,
-			m_cellHeight + (margin*2) );
+	if (!m_seamlessLeft && !m_seamlessRight)
+	{
+		QStyleOptionFrame opt;
+		opt.initFrom(this);
+		opt.state = QStyle::State_Sunken;
+		opt.rect = QRect(0, 0, m_cellWidth * m_numDigits + (margin + m_marginWidth) * 2 - 1,
+			m_cellHeight + (margin * 2));
```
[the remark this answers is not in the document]
he29-net: I'm not sure if I read it correctly, but wouldn't that mean that all "vanilla" `LcdWidget`s would have no borders?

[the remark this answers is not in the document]
IanCaio: No, I just meant that on the regular LCD Widget, the borders are being drawn on the `LcdWidget` class (`LcdWidget::paintEvent`, where this comment was made) but on the Float LCD Widget they are being drawn on the `LcdFloatSpinBox` class (`LcdFloatSpinBox::paintEvent`). As far as I understood, LcdWidget is the LCD number while the LcdSpinBox is the widget itself, that contains the LCD number. LcdFloatSpinBox for example has 2 LcdWidgets.

I was just pointing out it feels inconsistent to draw borders on the LcdWidgets on the regular spin box and on the LcdFloatSpinBox for the float spin box. We probably should do the border drawing on the spin box classes (both `LcdSpinBox` and `LcdFloatSpinBox`).

src/gui/widgets/LcdWidget.cpp
IanCaio: Cool, could you try using 0 for the seamless? I think that would make sense. Having `m_marginWidth` is probably not causing any issues now because nothing is being drawn on this extra portion of the pixmap? I'm not sure tbh

src/gui/widgets/LcdWidget.cpp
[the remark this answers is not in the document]
he29-net: > We probably should do the border drawing on the spin box classes (both LcdSpinBox and LcdFloatSpinBox).

But that would mean `LcdWidget` is left with no border-drawing code, hence my worry that `LcdWidgets` will not have borders if we do that. They are currently used in the FX mixer (where a nuber is needed, but you cannot change it), so there would be no borders around the green numbers on the FX mixer. Either that, or I'm getting really confused. :D

`LcdFloatSpinBox` does border drawing simply because it needs to visually join two widgets that were not originally designed to be joined. That's also why there are the new "seamless" parameters in `LcdWidget`. It all feels a bit like a hack, all the three LCD widgets could be probably rewritten in some way to make it cleaner, but I felt that is already way out of scope of the original Microtuner (I generally try to avoid touching existing code when not necessary, not to mention big rewrites like this).

src/gui/widgets/LcdWidget.cpp
he29-net: I just tested both versions and as far as I can tell they look exactly the same, so replaced it with zero as you suggested.

src/gui/widgets/LcdWidget.cpp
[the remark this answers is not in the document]
IanCaio: Oh, I didn't realize that `LcdWidget` was used by other widgets besides the Spin Box. Well, in that case it's probably a big rewrite to move the border to the outer widgets: Even though it makes sense it's out of the scope of this PR. I'm not sure it's worth it adding a `TODO` comment for the future, but you can ignore this comment then!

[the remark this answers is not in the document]
IanCaio: Could a comment be added at the beginning about the border drawing, just for future reference? Something like:
```
// When either the right or left is seamless we are in a LcdFloatSpinBox
// and the border drawing will be done by the float spin box class instead
// TODO: Possibly move the frame drawing to the outer widget classes
// (i.e.: Spin boxes) for consistency
```

I know this sounds unimportant, but if in the future someone decides to change the frame drawing and isn't aware of that, the frames of regular spin boxes and float spin boxes won't match. The `TODO` also serves as a reference for possibly coming up with an alternative in the future.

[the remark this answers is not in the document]
he29-net: I added a comment explaining why is the border drawing skipped. I left out the "TODO", because I still do not understand where do you propose to "move" the code. If it is moved, it is no longer there, which breaks `LcdWidget` borders in the mixer. So we can only _copy_ it to `LcdSpinBox`, which does not make sense, it only adds more duplicate code that needs to be maintained.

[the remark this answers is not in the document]
IanCaio: I don't know, maybe someone else can step in to give an opinion too, I could be wrong after all. For me it feels like bad design that a widget contains another one (SpinBox contains LcdWidgets) but in some cases the inner widget (LcdWidget) draws the frame and in other cases the outer widget (Float Spinbox) draws the frame. In a ideal scenario either the inner OR the outer widget should be responsible by the frame. That's all I was suggesting, that in the future we move the drawing of the frame to the outer widgets (LcdFloatSpinBox, LcdSpinBox, FxLineLcdSpinBox, etc) or that we come up with a design where we don't have it being drawn in different places.

I understand this is out of the scope of this PR (and obviously not its fault by any means), I just thought about adding a `TODO` comment so we could remember about maybe fixing it in the future. Just to let you know, this is by no means holding my approval (as I mentioned in Discord, I'm mostly waiting on @DomClark regarding the `update` method name, the aside the PR LGTM).

include/LcdFloatSpinBox.h
DomClark: I tested this, see the code at https://gist.github.com/DomClark/fc976514c6e1b083d9bce0ab777eefad. It appears slots have virtual behaviour when connected using `SIGNAL`/`SLOT` macros, and non-virtual behaviour when using member function pointer connections. This inconsistency, along with the fact that it doesn't work with the modern connection style, suggests to me that it's not a good idea to override slots by name. I'm fine with leaving it as is for now, since the existing `LcdSpinBox` code does the same thing, but I think we should clean this up in the future.

DomClark (approved):
I'm happy with this if @IanCaio is.

IanCaio (approved):
LGTM too!

include/LcdFloatSpinBox.h
he29-net: That inconsistency certainly looks like an invitation for some confusing bugs, interesting..

I wonder what is the intended workflow then? I would expect that a derived widget implements its own `update()`, so that it can react to any changes specific to the modified version. And as a part of its `update()` it would call `update()` of the base class to finish the rest. So the non-virtual behavior should be almost always unwanted, right?

But if `update()`s are called by Qt on some occasions, wouldn't choosing a different "update" slot name on a derived class mean that it simply won't be called when needed? Or do we only call / signal `update()`s explicitly from our code?

DomClark: `update()` isn't supposed actually update anything itself. Its purpose is to schedule a repaint of the widget in the near future, once control returns to the event loop. If virtual behaviour were intended, it would be a virtual function.

It looks like the code there is to set the LCD display to the current value of the model, in which case it should run in response to changes in the model. The appropriate places for this are in `LcdFloatSpinBox::modelChanged` (called when a new model is viewed by the spin box) and in response to the `Model::dataChanged` signal (emitted when the model's value changes).

PhysSong: In addition to @DomClark's comment, `ModelView::doConnections` connects `Model::dataChanged` and `Model::propertiesChanged` to `SLOT(update())` of `ModelView::m_widget`. This means at least one of them should be satisfied:
- `paintEvent()` of the model reads the value of the model,
- The view monitors changes in its model
- The view implements its own `update()`

`LcdFloatSpinBox`(and `LcdSpinBox`, too) is using the third approach. If this approach is not recommended, we should move to other approaches.
However, this looks out of scope of this PR.

PR #5868, opened 2020-12-31
MIDI range MKII (extracted from microtonal PR)

### LINKED ISSUE #1857: MIDI-based instruments play an octave too low by default

he29-net:
This PR largely covers what was originally in #5349, plus some other bells and whistles added later in #5522.

It should have all the code to fix #1857 and on the first glance it seems to be working correctly on its own. Parts that depend on the Microtuner code are commented out and / or replaced by constants, they should be easy to resolve as conflicts after this is merged.

For now marking as a draft, since there was a mention (I think on Discord) that some of the presets have a bad base note value, likely caused by a wrong conflict resolution in the recent months.


src/tracks/InstrumentTrack.cpp
```diff
@@ -138,24 +145,32 @@ InstrumentTrack::InstrumentTrack( TrackContainer* tc ) :
...
 }
 
 
 int InstrumentTrack::baseNote() const
 {
 	int mp = m_useMasterPitchModel.value() ? Engine::getSong()->masterPitch() : 0;
 
-	return m_baseNoteModel.value() - mp;
+//	return m_microtuner.baseKey() - mp;
+	return DefaultBaseKey - mp;
```
PhysSong, line 160: I think this is a regression, you should return `m_baseNoteModel.value() - mp` for now.

src/tracks/InstrumentTrack.cpp
```diff
@@ -138,24 +145,32 @@ InstrumentTrack::InstrumentTrack( TrackContainer* tc ) :
...
 }
 
+int InstrumentTrack::firstKey() const
+{
+//	return m_microtuner.firstKey();
+	return 0;
```
PhysSong, line 166: Why don't you return `m_firstKeyModel.value()`? If you're not going to use this for now, you can also remove related code from this PR.

src/tracks/InstrumentTrack.cpp
```diff
@@ -138,24 +145,32 @@ InstrumentTrack::InstrumentTrack( TrackContainer* tc ) :
...
+}
+
+int InstrumentTrack::lastKey() const
+{
+//	return m_microtuner.lastKey();
+	return 127;
```
PhysSong, line 172: Similarly, `m_lastKeyModel.value()`.

src/tracks/InstrumentTrack.cpp
```diff
@@ -1638,12 +1655,26 @@ void InstrumentTrackWindow::modelChanged()
 		m_pitchRangeLabel->hide();
 	}
 
+	if (m_track->instrument() && m_track->instrument()->flags().testFlag(Instrument::IsMidiBased))
+	{
+/*		m_miscView->microtunerGroupBox()->hide();
```
PhysSong, line 1660: I don't want those comments in this PR, but if you still want to include them, you can comment out the whole `if`\~`else` block.

PhysSong (commented):
My initial review for this PR.

src/tracks/InstrumentTrack.cpp
he29-net: Thanks, I did not see the diff when I made the change and apparently I guessed incorrectly. :)

src/tracks/InstrumentTrack.cpp
he29-net: The first and last key models are needed only for the microtuner, so I tried to avoid using them while keeping most of the other code in place to avoid possible spaghetti moments (one removed function → another change required in a different file → another one elsewhere → before you know it it's a mess that does not work).

I did not realize that the first / last key models actually exist in this code, so returning a constant was a fast and safe thing to do. But looking at it now, when I return the model value, the range limits in PianoView actually work (the only thing missing is the disabled key highlighting). So I will fix that and keep it, since it actually provides a useful feature on its own.

src/tracks/InstrumentTrack.cpp
he29-net: I can delete it, I was just trying to make my life easier by making the future conflicts more obvious / less likely to miss.

he29-net:
I went through the presets and found only two that did not have the correct 12 semitone base note offset (AFP/orion.xpf and TripleOsc/PowerStrings.xpf).

I also fixed some other issues after going through the code once again; I believe this PR should be ready for review now.

PhysSong:
If you're going to keep the key range feature, why don't you add `white_key_disabled` and `black_key_disabled` pixmaps instead?

he29-net:
> If you're going to keep the key range feature, why don't you add `white_key_disabled` and `black_key_disabled` pixmaps instead?

The drawing code depends on the microtuner (check if mapped, the range check is included in that), so they won't be drawed anyway. If you want I can include them, it does not really make much difference. Or I could even slap together some condition that check the range to allow drawing them. It just seemed like a wasted effort on something that is going to be (hopefully) replaced in a few days by the proper original code.

he29-net:
Okay, both PianoRoll and PianoView should now draw the out-of-range keys.

data/projects/templates/default.mpt
```diff
@@ -81,5 +81,168 @@
     <projectnotes width="640" x="700" y="10" maximized="0" height="400" visible="0" minimized="0"></projectnotes>
     <timeline lp1pos="192" lp0pos="0" lpstate="0"/>
     <controllers/>
+    <scales>
```
PhysSong, line 84: The scale changes shouldn't be here.

data/projects/templates/default.mpt
```diff
@@ -29,7 +29,7 @@
         <bbtrack>
           <trackcontainer width="640" x="610" y="5" maximized="0" height="400" visible="0" type="bbtrackcontainer" minimized="0">
             <track muted="0" type="0" name="Kicker" solo="0">
-              <instrumenttrack pan="0" fxch="0" usemasterpitch="1" pitchrange="1" pitch="0" basenote="57" vol="100">
+              <instrumenttrack pan="0" fxch="0" usemasterpitch="1" pitchrange="1" pitch="0" basenote="69" vol="100">
```
PhysSong, line 32: Why did you change this while leaving the TripleOscillator track as is? Isn't everything supposed to be handled by the upgrade routine

he29-net: I wish I could tell you, but I hardly remember what I had for lunch 2 days ago, let alone:
```7e55bfb124 (Martin Pavelek  2019-12-17 21:37:34 +0100  32)               <instrumenttrack pan="0" fxch="0" usemasterpitch="1" pitchrange="1" pitch="0" basenote="69" vol="100">```
:D

Looking at the file in the original microtuner PR, the TripleOsc track is also adjusted to 69. For some reason, the change on line 8 did not make it here. o_O

In any case, the last version of the upgrade routine should indeed take care of it, so this change is no longer needed. I will test it and push an update. Thanks.

**Comment by PhysSong** [comment]
I think it will be nice if user presets can be upgraded automatically. Have you tried it, or not?

**Comment by he29-net** [comment]
I tried (https://github.com/LMMS/lmms/pull/5522#issuecomment-743916968), but found that the upgrade routine does not run when loading a preset, so there is currently no mechanism that I know of that could be used to perform such upgrade.

**Comment by PhysSong** [comment]
> the upgrade routine does not run when loading a preset

It does. The real difference is: in preset files, `instrumenttrack` nodes are in `instrumenttracksettings` while it's `track` in project files.

**Comment by he29-net** [comment]
Oh? I'll have to look at it again then. Thanks for the info.

**Comment by he29-net** [comment]
I did not look at the presets yet, but meanwhile @IanCaio reported a potential problem on Discord and it turned to be a pretty big can of worms. ^^;

The reported problem was that when a note is placed on Piano Roll and it is out of the range defined by firstKey and lastKey markers, it would still play. This was meant to be solved by the microtuner keyToFreq() function, that would report a frequency of 0 Hz in case the requested key was invalid, so the result should be silence (0 changes per second). I forgot to add the range check to the legacy frequency computation code in NotePlayHandle, so the notes still played when microtuner was disabled. That was an easy fix.

BUT, it lead me to test another thing I missed before: how do actually instruments _behave_ when they get a 0 Hz note. And the results pretty much range from "as expected" to "disastrous". TripleOsc and one other native synth were pretty much the only ones handling it correctly and producing no sound. Some _almost_ worked but produced clicks when trying to play the "non-note". Most other native synths produced some sort of infra-sound + higher harmonics. Zyn and LV2 instruments ignored the range limit, because the MIDI messages also do not have any key range filter (and MIDI based instruments are not handled by the microtuner either). The worst two offenders were Vibed which caused LMMS to crash, and sfxr which on the first try went completely berserk (could not replicate it again though).

So overall, it seems clear that:
 a) playing 0 Hz tones is pretty dangerous, because even if we fix the native plugins, it could still cause some badly coded external plugins to explode, and
 b) it would be best to avoid playing the out-of-range notes at all, thus fixing LV2 and other MIDI-based synths and also preventing other problems like some playing infrasound or the upper harmonics as if it was a non-zero frequency.

The second point should pretty much take care of everything, but I'll also change the code in #5522 so that if for any reason frequency is requested for a key that is not mapped, 1 Hz is returned instead of 0 to prevent possible crashes and speaker explosions.

**Comment by PhysSong** [comment]
Can we make `NotePlayHandle` somehow notice the range limit without microtuner?

**Comment by he29-net** [comment]
> Can we make `NotePlayHandle` somehow notice the range limit without microtuner?

For `NotePlayHandle` it would be just a small change in `updateFrequency()` (simply look at the `firstKeyModel` and `lastKeyModel` values in instrument track), but since playing 0 Hz notes turned out to be dangerous I'll have to instead figure out how to prevent the note from playing at all. I did not look into that yet, but I'll try to make some progress on my PRs tonight and tomorrow.

**Comment by he29-net** [comment]
@IanCaio The issue you reported on Discord should be now fixed (and review suggestions applied).

It took me a while to figure out the right conditions, actions and their order to use (as I had no idea how `NotePlayHandle` actually works), but hopefully the notes should now not only stay silent if they are out of range, but they should also go through the release phase end emit `MidiOff` if the user changes the key range in a way so that a note comes out of range while it is already playing.

I also fixed (read: "finally remembered to add") saving / loading of `firstKeyModel` and `lastKeyModel`, so that should now only leave the presets upgrade routine to deal with.

**Comment by he29-net** [comment]
@PhysSong The upgrade routine really turned out to be working for presets as you said; it should be now handling VST, Carla and LV2 presets. But I did not realize that user presets for instruments not affected by the "MIDI note number bug" would also sound an octave higher than they should, because the `basenote` definition changed.

Since there does not seem to be anything in the XML that would indicate if the preset is "factory" or "user", I guess the best solution would be now to nuke all the preset file changes in this branch, and let the upgrade routine handle _all_ presets.

That is, all except presets designed and saved internally in Zyn. But even then the end result can be better than I originally imagined possible: the native `.xiz` presets won't by touched by the upgrade routine (which is good: the native factory presets will sound an octave higher, _as they should_) and the user `.xpf` Zyn presets will be upgraded, so they should keep sounding same as before. So exactly what we want, I think.

Having a working upgrade routine for presets really changes everything.. :))

File: src/core/DataFile.cpp [hunk 7]
```diff
@@ -1386,6 +1387,85 @@ void DataFile::upgrade_noHiddenClipNames()
...
+		QDomNodeList presets = elementsByTagName("instrumenttrack");
+		if (presets.item(0).isNull()) { return; }
+		QDomElement preset = presets.item(0).toElement();
+		// Common correction for all instrument presets (make base notes match the new MIDI range).
+		// NOTE: Many older presets do not have any basenote defined, assume they were "made for 57".
+		// (Specifying a default like this also happens to fix a FileBrowser bug where previews of presets
+		// with undefined basenote always play with the basenote inherited from previously played preview.)
+		bool ok = false;
+		int oldBase = preset.attribute("basenote").toInt(&ok);
+		if (!ok) { oldBase = 57; }
```
IanCaio, line 1451:
```suggestion
		int oldBase = preset.attribute("basenote", "57").toInt();
```

I think that's slightly cleaner

IanCaio (approved):
I had done a review previously, which had its requests merged, so this time I looked on the recent changes only.

We talked a lot on Discord about alternatives for dealing with the note ranges, but I think the one being used is currently the best we can do for now without adding too much complexity to the code.

There's a small suggestion for the Upgrade routine. The review on this method was mostly looking for obvious problems because I'm not very familiar with the instruments that require the note changes and the ones which doesn't. In terms of code it looks good, and if there's something being wrongly changed testing will show us pretty quick and it will be something easy to fix anyways. I'd also suggest adding a `TODO` comment on top of this upgrade routine to remind us to fix the presets instead at some point.

Nice work!

src/core/DataFile.cpp
he29-net: Thanks, I wasn't aware that you can pass a default value to the `attribute()`.

he29-net:
> I'd also suggest adding a `TODO` comment on top of this upgrade routine to remind us to fix the presets instead at some point.

I'm not sure what you are referring to; I assume the fact that many old presets are missing the `basenote` value? That should be now solved by the upgrade routine, so there is no real need to touch the individual preset files.

IanCaio:
> I'm not sure what you are referring to; I assume the fact that many old presets are missing the `basenote` value? That should be now solved by the upgrade routine, so there is no real need to touch the individual preset files.

Not only that, but also that most presets have a wrong `basenote` now that the MIDI range is fixed. Even though the upgrade routine solves it, I'm guessing we will want to fix the presets themselves at some point (that's what the TODO comment would be about).

I just realized something we would have to deal with when fixing the preset though: The new upgrade system makes it easy to add an upgrade routine, but not really easy to remove one. When we fixed the presets we would have a bogus upgrade routine (we can't remove it or the version drops). That's something that we can figure out when we come to fixing the presets though, maybe sync the presets fixing with some PR that has an upgrade routine so we can remove the bogus one with no issues.

Anyways, that was what I was referring to :smile:

PhysSong:
@Veratil could you double-check the piano-roll part and check if there are any conflicts between this and #5881?

he29-net:
> When we fixed the presets we would have a bogus upgrade routine (we can't remove it or the version drops).

@IanCaio When we upgrade out presets, their `creator version` will be (should be) updated as well, so the upgrade routine won't run for them. We wouldn't even _want_ to remove the "bogus" upgrade routine, because it would be still needed for all the old user presets that have the same problem but that we cannot fix.

IanCaio:
> @IanCaio When we upgrade out presets, their `creator version` will be (should be) updated as well, so the upgrade routine won't run for them. We wouldn't even _want_ to remove the "bogus" upgrade routine, because it would be still needed for all the old user presets that have the same problem but that we cannot fix.

Maybe I'm making some confusion here, but the way I thought it:

The upgrade routine is called when the project file creator version is below this PRs version. It will increase the base note on the project file, leaving the presets untouched. If we later fix the presets, those projects that had the base note increased will sound off from the point we fixed the presets forward because the projects will be relative to the previous base note. The older projects will also sound off because the upgrade routine will run for them and they will increase the base note twice. Unless we add another upgrade routine that undoes what the first one did for when we fix the presets.

BTW, ignore the idea I mentioned earlier of syncing the fixing of presets with another PR that adds an upgrade routine, it wouldn't work the way I thought.

src/gui/editors/PianoRoll.cpp
```diff
@@ -3954,16 +3978,14 @@ void PianoRoll::updateYScroll()
 						height() - PR_TOP_MARGIN -
 						SCROLLBAR_SIZE);
 
-	int total_pixels = m_octaveHeight * NumOctaves - (height() -
-					PR_TOP_MARGIN - PR_BOTTOM_MARGIN -
-							m_notesEditHeight);
-	m_totalKeysToScroll = qMax(0, total_pixels * KeysPerOctave / m_octaveHeight);
+	const int visible_space = height() - PR_TOP_MARGIN - PR_BOTTOM_MARGIN - m_notesEditHeight;
```
Veratil, line 3981:
```suggestion
	const int visible_space = keyAreaBottom() - keyAreaTop();
```

Veratil:
> 
> 
> @Veratil could you double-check the piano-roll part and check if there are any conflicts between this and #5881?

This PR doesn't touch the 3 lines of `drawNoteRect`. 👍

he29-net:
@IanCaio Uh, I think I'm getting pretty lost here. :D The words enter my brain, bounce around a little and fizzle out.. I'll try to break it up into pieces:
> The upgrade routine is called when the project file creator version is below this PRs version. It will increase the base note on the project file, leaving the presets untouched.

Yes.
> If we later fix the presets, those projects that had the base note increased will sound off from the point we fixed the presets forward

Why? AFAIK the project file does not store a link to the preset, the instrument configuration is stored within the project file the same way as other instruments, no? So it does not matter what happens to presets stored on the disk.
> because the projects will be relative to the previous base note.

I don't understand what is meant by "a project being relative to the previous base note".
>  The older projects will also sound off because the upgrade routine will run for them and they will increase the base note twice. Unless we add another upgrade routine that undoes what the first one did for when we fix the presets.

I think this hinges again on the fact that presets are not stored as links but same as other instrument instances. Running the upgrade routine on older project is always intended and needed, to adjust it to the new MIDI range.

IanCaio:
Sorry, as I suspected I think I made a lot of confusion out of it. I thought the presets were somewhat linked, so I was thinking if you didn't change the base note on the project it would use the preset's default base note (from the preset file). But if it's all stored on the project's XML then you're right, it doesn't matter if we update the preset files later.

So just completely ignore those comments :laughing:

src/core/DataFile.cpp
```diff
@@ -1386,6 +1387,83 @@ void DataFile::upgrade_noHiddenClipNames()
...
+		preset.setAttribute("basenote", oldBase + 12);
+		// Extra correction for Zyn, VeSTige, LV2 and Carla (to preserve the original buggy behavior).
+		QDomNodeList instruments = presets.item(0).toElement().elementsByTagName("instrument");
+		if (instruments.isEmpty()) { return; }
+		QDomElement instrument = instruments.item(0).toElement();
+		if (instrument.attribute("name") == "zynaddsubfx" ||
+			instrument.attribute("name") == "vestige" ||
+			instrument.attribute("name") == "lv2instrument" ||
+			instrument.attribute("name") == "carlapatchbay" ||
+			instrument.attribute("name") == "carlarack")
```
PhysSong, line 1459: This check is long and repeated twice, so it should be simplified or split out if possible.

he29-net: Good idea, fixed. (Although, ironically, by eliminating the repetition the number of lines actually _increased_ by one due to the overhead of our coding conventions.. :smile: )

src/core/DataFile.cpp
```diff
@@ -1386,6 +1387,84 @@ void DataFile::upgrade_noHiddenClipNames()
...
 
 
+/** \brief Note range has been extended to match MIDI specification
+ *
+ * The non-standard note range previously affected all MIDI-based instruments
+ * except OpulenZ, and made them sound an octave lower than they should (#1857).
+ */
+void DataFile::upgrade_extendedNoteRange()
+{
+	auto affected = [](QDomElement instrument)
```
PhysSong, line 1397:
```suggestion
	auto affected = [](const QDomElement& instrument)
```
Or alternatively, you can pass the name instead of the full element.

he29-net: Oh, did not occur to me to make it const, thanks.

include/Piano.h
```diff
@@ -65,6 +65,10 @@ class Piano final : public Model
 	static bool isWhiteKey(int key);
 	static bool isBlackKey(int key);
 
+	static const unsigned int WhiteKeysPerOctave = 7;
+	static const unsigned int BlackKeysPerOctave = 5;
+	static const unsigned int NumWhiteKeys = 75;
+	static const unsigned int NumBlackKeys = 53;
```
Veratil, line 71: Is there a way to gain this through math and our constants?

he29-net: I don't think they could be derived "mathematically"; the first two are arbitrary constants that describe how many notes from one 12-EDO octave are deemed "special" when making a piano keyboard, the other two count how many of these special keys are available in total. And since the black-white order is also arbitrary, there is no easy mathematical formula that says how many special keys are in the last, incomplete octave.

It could be determined by an algorithm though, by simply checking `Piano::KeyTypes KEY_ORDER[]` (defined in `Piano.cpp` for some reason, not the header file), i.e. running `isBlackKey()` and `isWhiteKey()` for all keys and, well, counting them. Or, more efficiently, count all complete octaves using a simple formula, and finishing the count by looping through the last, incomplete octave.

Either way seems pretty convoluted and probably would have to run in the constructor (can you initialize a constant in the constructor? Actually, can you initialize static _anything_ in a constructor, since there is only one "instance"?)

Veratil: All good then. Just wanted to check. 👍

PhysSong:
@he29-net Could you resolve the conflicts?

src/gui/PianoView.cpp
```diff
@@ -83,39 +88,49 @@ const int LABEL_TEXT_SIZE = 7;      /*!< The height of the key label text */
...
+	{
+		s_blackKeyDisabledPm = new QPixmap(embed::getIconPixmap("black_key_disabled"));
 	}
 
-	setAttribute( Qt::WA_OpaquePaintEvent, true );
-	setFocusPolicy( Qt::StrongFocus );
-	setMaximumWidth( WhiteKeysPerOctave * NumOctaves * PW_WHITE_KEY_WIDTH );
+	setAttribute(Qt::WA_OpaquePaintEvent, true);
+	setFocusPolicy(Qt::StrongFocus);
+	setMaximumWidth(Piano::NumWhiteKeys * PW_WHITE_KEY_WIDTH +
```
Spekular, line 126: Maybe leave a comment explaining the reason for this? I can't think of a concise way to do so though.

Black keys are drawn halfway between the preceding and succeeding white note, and contained within their width. Therefore if the first and last keys are white, the maximum required width is the width of all the white notes. However, if the last key is black, this width would cut off the half of that key that would overlap the next white note.

Also, shouldn't this apply if the first note is black as well? Why is the first note guaranteed to be white but not the last one?

Spekular (commented):
Looks good (in the sense than nothing looks bad) to me.

src/gui/PianoView.cpp
he29-net: > Also, shouldn't this apply if the first note is black as well?

I don't think it is currently possible to have the first note black (unless someone changes the black / white definitions in `src/core/Piano.cpp`), since there is no option to make the first octave incomplete, i.e. it always starts from C.

The `PianoView` code is not ready for that possibility either: it always starts drawing from a white key. So if the first key was to be black, it would have to be cut in half, since there is no preceding white key to set as `m_firstKey`.

Anyway, I added a comment, hopefully it makes it a bit clearer.

he29-net:
> @he29-net Could you resolve the conflicts?

Done. Thanks for the ping, I did not notice there was a conflict.

PhysSong:
@gnudles Sorry for pinging, but are there any chances that some Xpressive formulas are affected by this PR? I think `key` and `bnote` parameters may be affected, but I'm not sure how should we handle that cases.

gnudles:
It might be affected. let me check.

gnudles:
@PhysSong, my original intention when I exported / introduced these values (bnote and key) in formula, was indeed to allow customized key to frequency mapping and other scales. From what I understand, this pull request now allows to make all linear key to freq mapping. So Xpressive has now only the advantage of non-linear mappings. since non linear mappings are very rare, and since the plugin released only under alpha version, I think we should consider to remove these variables from the formula?

gnudles:
Btw, looks like somebody really liked my plugin XD
https://m.youtube.com/watch?v=G6TfCTm-wmo

IanCaio:
Hey man! Apparently there are 2 new conflicts, simple ones to solve (just on the header includes), could you fix those?

he29-net:
@IanCaio Thanks, done.

---
I'm still not sure what to do about the key variables in Xpressive formulas (if anything). I checked the factory presets and the only one using them seems to be "Faded Colors - notes test.xpf", so breaking one "test preset" by removing the variables should not be a major problem. It's hard to say how many users used them, but as `gnudles` said, it's a plugin only available in an alpha version, so that should not cause much pain either.

@gnudles As for the "feature-completeness": if I understand what you meant by non-linear mapping, it should also be possible to do with the Microtuner. You would simply make a keymap covering all 127 keys, and a scale with 127 scale degrees, creating a "lookup table" for an arbitrary function. Making it would not be as comfortable as writing a single formula, but definitely possible.

Btw. I also saw that video; the "get out of my left ear!" was very close to my first experience with Xpressive, the stereo imbalance in some presets is a bit hard for my brain to accept as a valid input.. :D

gnudles:
Ok. I will remove the "note test" from the presets and the stereo balance, and will remove the "key" and "bnote" variables.

he29-net:
@PhysSong With the Xpressive issue resolved, is there anything else you wanted to look into, or is this ready to be merged?

src/core/NotePlayHandle.cpp
```diff
@@ -196,6 +196,25 @@ void NotePlayHandle::play( sampleFrame * _working_buffer )
...
+	// TODO: handle the range check by Microtuner, and if the key becomes "not mapped", save the current frequency
+	// so that the note release can finish playing using a valid frequency instead of a 1 Hz placeholder
+	if (key() < m_instrumentTrack->m_firstKeyModel.value() ||
+		key() > m_instrumentTrack->m_lastKeyModel.value())
+	{
+		// Release the note if it started playing before going out of range
+		if (!m_released || m_hasMidiNote)
+		{
+			noteOff(0);
+		}
```
DomClark, line 209: If `m_released` is `true`, `noteOff` returns immediately, so this test can be removed and `noteOff` called unconditionally.

DomClark (commented):
Generally looks good.

* The Xpressive documentation changes in 80775479d9dbd9102a4eee89d09ab105054a53a4 probably belong in this PR.
* How important is it to keep the commented-out microtuner code here? It's fine as it is, but the history would be cleaner if it weren't there.

he29-net:
>     * The Xpressive documentation changes in [8077547](https://github.com/LMMS/lmms/commit/80775479d9dbd9102a4eee89d09ab105054a53a4) probably belong in this PR.
> 
>     * How important is it to keep the commented-out microtuner code here? It's fine as it is, but the history would be cleaner if it weren't there.

Thanks for taking a look. I see that `gnudles` probably plans to only have the `key` and `bnote` disabled and not completely removed in #5979, so keeping the documentation update and including it here makes sense. When I made this PR, I assumed it will be just a quick and temporary PR to lighten to load of the Microtuner PR, so I may have missed a few things like this.

The commented-out code is there for the same reason; with the pace things were going before Christmas, I assumed I will be rebasing #5522 a weak or two later, so I left a lot of the code in place so that it generates more obvious merge conflicts. I don't trust git to always magically do all changes correctly, so this would guide me to places that need to be reviewed manually. But if we want this to be a "proper" commit, I can do a cleanup pass, it's not that important (I can just do a bit more testing after the rebase to make sure nothing got broken).

DomClark:
> The commented-out code is there for the same reason; with the pace things were going before Christmas, I assumed I will be rebasing #5522 a weak or two later, so I left a lot of the code in place so that it generates more obvious merge conflicts. I don't trust git to always magically do all changes correctly, so this would guide me to places that need to be reviewed manually. But if we want this to be a "proper" commit, I can do a cleanup pass, it's not that important (I can just do a bit more testing after the rebase to make sure nothing got broken).

Sorry about the change in pace. This particular change is only aesthetic, so I'll let you weigh up whether it's worth the risks with the merge conflicts.

he29-net:
I pushed the other small changes for now, I'll look at the comments tomorrow-ish. Some of them may be helpful to explain some of the byproducts of the rough extraction from the main PR (i.e. stuff like `if (true)` placeholders for tests performed by the microtuner, not sure if any of them remained), but if there are others without such purpose I'll try to get rid of them, to make a better impression on the future archeologists digging through our commit history. ^^

he29-net:
The comment purge is complete; I left only two of them (explaining some  `if (true)` and `if (false)` placeholders) and noted the line numbers for the others to check manually after the Microtuner rebase. Maybe I'm just too paranoid and it won't be needed, I'm curious to see how git handles it. :)

DomClark (approved):

Issue #5817, opened 2020-12-02
Knob tooltip inconsistencies


dj-pixus:
There are some formatting issues with some tooltips that appear when you change the value of a knob/slider, which makes things look a bit unprofessional.

These are the ones I found so far:
- In most places e.g. the native filter in the instrument window there is a space between the number and "Hz", while for example in the Dual Filter plugin there isn't. Same for "s" and "ms" on some time controls. (Which one is correct?)
- Native Delay's controls don't have a colon between the parameter name and value. Same for Vectorscope's persistence knob.
- Crossover Equalizer's sliders have double spaces between the number and "dBFS"
- Resonance knobs in Native EQ have a space before the colon
- Flanger plugin uses "s" and "Sec" mixed

Even tho Transifex shows the context, it's easy to overlook it so I believe a lot of these errors are present in translations as well.

Also, if you find further issues like this, write it here.

M4rotte:
Hi,

Maybe I even already opened an issue, I can’t remember and I’m shamefully not inclined to check right now… 

In the LV2 plugin dialog, the drop-down lists should be larger so it can show what’s the selected value my weak brain forgot it was. I find it particularly annoying with the CALF Filter plugin: 

![image](https://github.com/LMMS/lmms/assets/6515724/a58bc93f-317c-4e2c-85c6-098c552db10a)

Rossmaxx:
> In the LV2 plugin dialog, the drop-down lists should be larger so it can show what’s the selected value my weak brain forgot it was. I find it particularly annoying with the CALF Filter plugin:

LV2 is experimental and such issues are to be expected. They will be fixed once LV2 GUI is done.

PR #5872, opened 2021-01-02
Fix some problems with knob labels

dj-pixus:
This PR addresses some of the stuff in #5817, making tooltips a bit more consistent. Also made the Crossover Eq controls translatable.
It's just a quick aesthetic change which shouldn't conflict with the reorganization.

### DIFF
--- a/plugins/BassBooster/BassBoosterControlDialog.cpp
+++ b/plugins/BassBooster/BassBoosterControlDialog.cpp
@@ -47,7 +47,7 @@ BassBoosterControlDialog::BassBoosterControlDialog( BassBoosterControls* control
 	Knob * freqKnob = new Knob( knobBright_26, this);
 	freqKnob->setModel( &controls->m_freqModel );
 	freqKnob->setLabel( tr( "FREQ" ) );
-	freqKnob->setHintText( tr( "Frequency:" ) , "Hz" );
+	freqKnob->setHintText( tr( "Frequency:" ) , " Hz" );
 
 	Knob * gainKnob = new Knob( knobBright_26, this );
 	gainKnob->setModel( &controls->m_gainModel );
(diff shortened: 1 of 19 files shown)


qnebra:
For me its fine. Only some text changes, few typos fixed. Nothing technical was touched from what I see.

IanCaio (approved):
Code style ignored on review. Changes LGTM!

superpaik:
EQ: A question: why IN and OUT gain faders in EQ are labeled **dBv** and gain faders in filters are **dB**? Is there a reason for that?
Flanger: Delay has the label "Delay time" without space before "s" (0.XXs) but Rate has space (0.XX s)

Suggestion: since you are checking knobs labels in general, the instrument's envelop knobs (in volume, cutoff, resonance) doesn't have a unit. I don't know if its always the same unit for every instrument and can be added to the label? Maybe you can check that out.
imatge

dj-pixus:
> EQ: A question: why IN and OUT gain faders in EQ are labeled **dBv** and gain faders in filters are **dB**? Is there a reason for that?

Don't know what the original concept was, but **dBv** is certainly not the right unit here. It is used on analog gear to specify the voltage of a signal, which is uninterpretable in a digital environment.
Apparently these faders use normal **dB** values, so I'm going to change them to **dB** like it is everywhere else in the program.

Some weird thing about spaces: `setHintText( tr( "Delay time:" ) + " ", " s" )`
Looks like the leading space in `" s"` gets trimmed here, but in other places there is no space written in the quotes, but the final displayed string still contains one. So I guess the space is somehow part of the value itself(?) But it's not consistent.

The envelope parameters:
This is where my brain gave up. It appears to be some really weird exponential-like scale. `0.5` is 1.25 second, `1.0 is` 5 seconds, `2.0` is 20 seconds. It doesn't look like any regular function. Also, they seem to be adjusting the slope rather than the time, because the sustain knob also affects the decay time with the slope remaining the same.

superpaik:
I agree with you on de dB dBv matter.
As for the delay, maybe the correct code should be delayKnob->setHintText( tr( "Delay time:" ) + , " s" ) with the space next to the s, as it is in the rest of the code.
And about the envelope parameters, now I see why it was so difficult to use ¿? Let's see if someone else can shed some light on the matter, if not let it be as it is.

Veratil:
Hey @dj-pixus if you could rebase and update to the current style conventions we can work on getting this merged. :)

dj-pixus:
Rebased in #6651, closing this one.

Issue #4925, opened 2019-04-01
ignore capitals in Token-search


musikBear:
Sometimes i find it difficult to find a resource, even though i can remember its name. The reason is that LMMS make distinktion between capital letters and small letters.
So  where it would be logical that 'Horn01' and 'horn02' was listet directly after each other, they wont be if we have a series of 'Hot xyz' named resources. 
I suggest that capitals in names/ labels are ignored in Token-search-*listings*.
(May bee an accessibility related thing as well)

PhysSong:
Currently, it looks like this(case-sensitive):
```
AB.wav
Ba.wav
aa.wav
bb.wav
```
The suggestion is(case insensitive):
```
aa.wav
AB.wav
Ba.wav
bb.wav
```
I think it should be easy to change the behavior because `QDir::entryList` have an option to ignore cases when sorting. However, I'm not sure if all people think the latter behavior is desired.

There are several different options:
* :laugh:: Keep the current behavior
* :eyes:: Ignore case when sorting(change to the latter behavior)
* :heart:: Add an option to allow users to change the behavior

Please let me know user opinion!

musikBear:
yes, an *option* for *user* to decide, is a brilliant idea!

musikBear:
:| cant 'vote' -The smileys are not click-able in this browser, so you can add one for "3"
> Add an option to allow users to change the behavior

SecondFlight:
@musikBear I'll add a vote for you for :heart:. For the record though, my vote is :eyes: :)

musikBear:
@SecondFlight  Thanks! :+1

noahb01:
Definitely change it. When you have a ton of presets, it ends up looking like its been alphabetized twice (first all the caps, then the lowercase), which makes finding the preset/sample you want harder than it needs to be.

the-bebop:
hey all I'm fairly new to LMMS coding, but I'd I think I can start on this one.

zonkmachine:
> hey all I'm fairly new to LMMS coding, but I'd I think I can start on this one.

OK. Assigning you to the issue.   ;)

the-bebop:
Thanks, just to make sure we're on the same spot we're talking about list elements like the _my presets_ browser i.e. in the fixed version _orion.xpf_ should be put before _SString.xpf_ see example screenshot from a freshly built lmms:

[...]

PR #5875, opened 2021-01-04
sort items in tree-browsers caseinsensitively in alphabetical order

the-bebop:
This PR adjusts the item order alphabetically independent from naming style. Hence items will be sorted with the following manner:

1. aa
2. Ab
3. Ba
4. bb

Affects amongst others _Instrument Plugins, My Projects, My Samples, My Presets, My Home, Root Directory_ and their sub items. See [the issue](https://github.com/LMMS/lmms/issues/4925) for details

### DIFF
--- a/src/gui/FileBrowser.cpp
+++ b/src/gui/FileBrowser.cpp
@@ -198,6 +198,8 @@ bool FileBrowser::filterItems( const QString & filter, QTreeWidgetItem * item )
 void FileBrowser::reloadTree( void )
 {
 	QList<QString> expandedDirs = m_fileBrowserTreeWidget->expandedDirs();
+	expandedDirs.sort(Qt::CaseSensitive);
+
 	const QString text = m_filterEdit->text();
 	m_filterEdit->clear();
 	m_fileBrowserTreeWidget->clear();
@@ -262,7 +264,6 @@ void FileBrowser::giveFocusToFilter()
 }
 
 
-
 void FileBrowser::addItems(const QString & path )
 {
 	if( m_dirsAsItems )
@@ -274,6 +275,7 @@ void FileBrowser::addItems(const QString & path )
 	// try to add all directories from file system alphabetically into the tree
 	QDir cdir( path );
 	QStringList files = cdir.entryList( QDir::Dirs, QDir::Name );
+	files.sort(Qt::CaseInsensitive);
 	for( QStringList::const_iterator it = files.constBegin();
 						it != files.constEnd(); ++it )
 	{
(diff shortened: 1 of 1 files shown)


tresf:
Can you explain why the directories are case sensitive whereas the files are case insensitive?   Is this desired?

the-bebop:
Hi @tresf can you please point out where this happened? looks good so far on my side see directories

grafik

or files

grafik

tresf:
> Hi @tresf can you please point out where this happened? looks good so far on my side see directories

It was in regards to this code change here: https://github.com/LMMS/lmms/blob/673df262590b45392d854a55ca9dcccab738ba3e/src/gui/FileBrowser.cpp#L201

src/gui/FileBrowser.cpp
```diff
@@ -198,6 +198,8 @@ bool FileBrowser::filterItems( const QString & filter, QTreeWidgetItem * item )
 void FileBrowser::reloadTree( void )
 {
 	QList<QString> expandedDirs = m_fileBrowserTreeWidget->expandedDirs();
+	expandedDirs.sort(Qt::CaseSensitive);
```
tresf, line 201: Why `CaseSensitive` here?

tresf:
> Hi @tresf can you please point out where this happened? looks good so far on my side see directories

Moved to in-line comment to avoid confusion. :)

src/gui/FileBrowser.cpp
the-bebop: right, let me fix that

src/gui/FileBrowser.cpp
```diff
@@ -262,7 +263,6 @@ void FileBrowser::giveFocusToFilter()
 }
 
 
-
```
tresf, line 265: This whitespace seems superfluous. We wouldn't normally fault for this but there's a second instance now introduced from a reverted commit, might as well fix them both.

src/gui/FileBrowser.cpp
```diff
@@ -198,6 +198,7 @@ bool FileBrowser::filterItems( const QString & filter, QTreeWidgetItem * item )
 void FileBrowser::reloadTree( void )
 {
 	QList<QString> expandedDirs = m_fileBrowserTreeWidget->expandedDirs();
+
```
tresf, line 201: This whitespace needn't be added now that you've reverted this change.

tresf:
Thanks!  Just some whitespace formatting (hope you don't mind) and we'll merge!

src/gui/FileBrowser.cpp
the-bebop: alright, btw. are we planing to set up an auto formater (clang-format for instance) or is there another way to enforce the code styles? In the Wiki I could only find which rules to follow.

src/gui/FileBrowser.cpp
```diff
@@ -262,7 +262,6 @@ void FileBrowser::giveFocusToFilter()
 }
```
Spekular, line 264:
```suggestion


```
Reverts whitespace change.

Spekular (approved):
LGTM.

Spekular:
Counting tresf's earlier comment as approval and adding my own => 2 approvals, merging. CI passed for previous commit (ignoring unrelated errors) and the only change since is an empty line.

> Just some whitespace formatting (hope you don't mind) and we'll merge!

src/gui/FileBrowser.cpp
tresf: Yes, we'll be doing just that (clang-tidy) in the future.  Thanks for the concerns, consideration and patience.  The codebase is undergoing a lot of refactoring which will greatly reduced this type of back-and-forth.

Issue #5024, opened 2019-06-09
lmms 1.2.0: text relocation in /usr/lib/lmms/RemoteVstPlugin.exe.so


dvzrv:
When building lmms 1.2.0 for Arch Linux, our package check tool namcap reports text relocations in `/usr/lib/lmms/RemoteVstPlugin.exe.so`: 

```
lmms W: ELF file ('usr/lib/lmms/RemoteVstPlugin.exe.so') has text relocations.
```

Is that library build with `-fPIC`?

tresf:
Related #4806.

No, I don't think so.  Looking back at the codebase, I believe the whole project [used to be in 2006](https://github.com/LMMS/lmms/commit/f9063f938356565fedf2f9d93dbd722e22ee58d3), but must have gotten refactored out at some point.

From what I'm reading, all plugins should build with this flag.  Do you get more of these?  If not, it may be due to the fact that we [compile this particular plugin by hand](https://github.com/LMMS/lmms/blob/stable-1.2/plugins/vst_base/CMakeLists.txt#L78-L89) instead of using whatever defaults that `cmake` uses.  Long term, we should convert this to `ExternalProject_Add` instead.

PhysSong:
>  Long term, we should convert this to `ExternalProject_Add` instead.

IIRC, @lukas-w refactored `RemoteVstPlugin` build logic to use that in #4421.

PR #5878, opened 2021-01-09
Fix 32-bit winegcc call on Linux when using LLD

rdrpenguin04:


PhysSong:
Possibly related: #5024
See also: https://github.com/llvm/llvm-project/commit/0a7412f00ba4a2ccca069c55a1e21cdafb3bade1

rdrpenguin04:
That other issue could be related; since `lld` blocks text relocations by default, it was a hard error for me.

The commit you referenced seems to be implementing the `-z notext` flag, not setting it as the default. I run `main` branch LLVM, so if it was made the default, I would have it by now.

This commit simply adds the flag so that RemotePluginVST32 will build and work properly, and, as I have stated, I have tested it on a 32-bit VST that I have.

PhysSong:
LGTM. Please revert the last commit, BTW. The `CONTRIBUTORS` file is a fallback for automatically generated file, so it should probably be regenerated properly if it needs updating.

rdrpenguin04:
K, will do.

PR #5881, opened 2021-01-17
Fix drawNoteRect misalignments

### LINKED ISSUE #5863: Pianoroll resizing is graphically broken in LMMS 1.3
Resizing pianoroll vertically in LMMS 1.3.0 Alpha nightly, is often graphically bugged.

Steps to reproduce
1. Open a pianoroll pattern.
2. Resize it vertically and take an eye on the keys that are now dinamically drawn.

Here's an example
a

Veratil:
Partially Fixes #5863 (Notes misaligning, not keys visual glitch)

Tested with ghost notes, detuning, scrolling vertical and horizontal, and resizing the note edit area. Everything seems to be working as normal now for notes.

### DIFF
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -3099,6 +3099,12 @@ void PianoRoll::paintEvent(QPaintEvent * pe )
 
 		QPolygonF editHandles;
 
+		// Return a note's Y position on the grid
+		auto noteYPos = [&](const int key)
+		{
+			return (topKey - key) * m_keyLineHeight + keyAreaTop() - 1;
+		};
+
 		// -- Begin ghost pattern
 		if( !m_ghostNotes.empty() )
 		{
@@ -3132,7 +3138,7 @@ void PianoRoll::paintEvent(QPaintEvent * pe )
 
 					// we've done and checked all, let's draw the note
 					drawNoteRect(
-						p, x + m_whiteKeyWidth, (topKey - note->key()) * m_keyLineHeight + keyAreaTop() - 1, note_width,
+						p, x + m_whiteKeyWidth, noteYPos(note->key()), note_width,
 						note, m_ghostNoteColor, m_ghostNoteTextColor, m_selectedNoteColor,
 						m_ghostNoteOpacity, m_ghostNoteBorders, drawNoteNames);
 				}
@@ -3154,8 +3160,6 @@ void PianoRoll::paintEvent(QPaintEvent * pe )
 				len_ticks = 4;
 			}
 
-			const int key = note->key() - m_startKey + 1;
-
 			int pos_ticks = note->pos();
 
 			int note_width = len_ticks * m_ppb / TimePos::ticksPerBar();
@@ -3172,7 +3176,7 @@ void PianoRoll::paintEvent(QPaintEvent * pe )
 			{
 				// we've done and checked all, let's draw the note
 				drawNoteRect(
-					p, x + m_whiteKeyWidth, (topKey - note->key()) * m_keyLineHeight + keyAreaTop() - 1, note_width,
+					p, x + m_whiteKeyWidth, noteYPos(note->key()), note_width,
 					note, m_noteColor, m_noteTextColor, m_selectedNoteColor,
 					m_noteOpacity, m_noteBorders, drawNoteNames);
 			}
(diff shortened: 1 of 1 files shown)


PhysSong:
Code looks good. @RoxasKH Could you test this and confirm the fix?

RoxasKH:
> 
> 
> Code looks good. @RoxasKH Could you test this and confirm the fix?

Just tested it and it seems it's still buggy to me. It'd be better if someone else could test it, too, tho.

https://streamable.com/uvtzh6

kalinushkin:
> 
> 
> > Code looks good. @RoxasKH Could you test this and confirm the fix?
> 
> Just tested it and it seems it's still buggy to me. It'd be better if someone else could test it, too, tho.
> 
> https://streamable.com/uvtzh6

Well, i tested it, and i find it correct working. "As by design" 👍

RoxasKH:
> 
> 
> > > Code looks good. @RoxasKH Could you test this and confirm the fix?
> > 
> > 
> > Just tested it and it seems it's still buggy to me. It'd be better if someone else could test it, too, tho.
> > https://streamable.com/uvtzh6
> 
> Well, i tested it, and i find it correct working. "As by design" 👍

No idea why it still happens to me, but i gotta admit i don't install properly LMMS versions to test them.
Maybe could also be related to resolution as i'm still on a 720p screen, but probably not.

Veratil:
> 
> 
> > Code looks good. @RoxasKH Could you test this and confirm the fix?
> 
> Just tested it and it seems it's still buggy to me. It'd be better if someone else could test it, too, tho.
> 
> https://streamable.com/uvtzh6

This is for the notes, the artifacting on the piano keys is another issue I noticed that I'll look into later.

RoxasKH:
> 
> 
> > > Code looks good. @RoxasKH Could you test this and confirm the fix?
> > 
> > 
> > Just tested it and it seems it's still buggy to me. It'd be better if someone else could test it, too, tho.
> > https://streamable.com/uvtzh6
> 
> This is for the notes, the artifacting on the piano keys is another issue I noticed that I'll look into later.

Oh, thanks to point this out, i actually created the issue for the artifacts of the pianoroll keys so i never really noticed the notes bug while resizing.

Seems still a little buggy, but mostly unnoticeable.

https://streamable.com/d625ie

superpaik:
In windows it doesn't work properly. When resizing the keys movement aren't smooth and sometimes they are keep unaligned . They get aligned again when the cursor enters the piano roll window.
imatge

Veratil:
> 
> 
> In windows it doesn't work properly. When resizing the keys movement aren't smooth and sometimes they are keep unaligned . They get aligned again when the cursor enters the piano roll window.
> imatge

This PR is for the notes themselves, the piano keys I have to look into separately.

PhysSong:
@Veratil Do you want to fix piano keys in this PR or in a separate PR?

Veratil:
> 
> 
> @Veratil Do you want to fix piano keys in this PR or in a separate PR?

I haven't figured out the reason for the piano keys glitch yet, but that will be a separate PR.

PhysSong:
@Veratil Isn't the `drawDetuningInfo` call at line 3224 supposed to be fixed as well?

Veratil:
> @Veratil Isn't the `drawDetuningInfo` call at line 3224 supposed to be fixed as well?

I didn't seem to have any issues with detuning draw, but I can double check it again.

> Also, could you extract `(topKey - note->key()) * m_keyLineHeight + keyAreaTop() - 1` to a variable to avoid repetitions and potential mistakes?

Yeah that's reasonable. It'd have to be a lambda though since `note->key()` changes within the for loop.

PhysSong (approved):

PhysSong:
This can potentially conflict with #5928, FYI.

Veratil:
> 
> 
> This can potentially conflict with #5928, FYI.

Double checking the diff for 5928, I don't see it touching these lines. It also still calculates the midpoint of the note based on the Y position given, so afaict we're safe on both. 👍

Spekular:
I can confirm the issue Roxas encountered, but I think it may be the octave lines that are bugged, since they stay aligned with the buggy keyboard. Dragging the bottom of the piano roll window down exposes more and more of the next lowest key until a full key is visible, at which point the view jumps to expose the next higher key instead. Regardless of whether this is contributing to the octave line/note desync, it's rather unpleasant.

Spekular (approved):
Code wise changes LGTM. As for testing, piano roll resizing is still buggy, but this is an improvement over the existing behavior on master.

Veratil:
> I can confirm the issue Roxas encountered, but I think it may be the octave lines that are bugged, since they stay aligned with the buggy keyboard. Dragging the bottom of the piano roll window down exposes more and more of the next lowest key until a full key is visible, at which point the view jumps to expose the next higher key instead. Regardless of whether this is contributing to the octave line/note desync, it's rather unpleasant. The same occurs when dragging the top of the piano roll up.

I'll see what I can debug to prevent this state.


> In 1.2, the new notes at the top are smoothly revealed.

I agree smooth reveal is better, and I'll aim to get back to that. 👍 For now we're slowly just fixing and cleaning the old, slow, and poorly maintained code. One step back, two steps forward. 😄

Issue #5876, opened 2021-01-06
MP3 stereo mode doesn't match to the selection in export dialog


PhysSong:
#3826 reordered the stereo mode selections in the export dialog, but it didn't update `mapToStereoMode` function in `ExportProjectDialog.cpp`. The order should now be changed to mono, stereo, joint stereo.
https://github.com/LMMS/lmms/blob/24761a4f7772df61315957d13ab562e91f5694cc/src/gui/ExportProjectDialog.cpp#L140-L153

deenadayalancs:
@PhysSong is this the expected changes need to be updated? please let me know. will push the changes.

OutputSettings::StereoMode mapToStereoMode(int index) 
{
switch (index) 
 	{
 	case 0: 
 		return OutputSettings::StereoMode_Mono;
 	case 1: 
                return OutputSettings::StereoMode_Stereo;
       case 2:
               return OutputSettings::StereoMode_JointStereo;
      default:
 	      return OutputSettings::StereoMode_Stereo; 
 	} 
}

PhysSong:
@deenadayalancs You're right.

deenadayalancs:
@PhysSong Please assign this to me. I am not able to checkout and update the code. Not able to create a pull request by myself.

Please create a pull request and assign this to me. Thanks again.

zonkmachine:
> I am not able to merge and update the code. Not able to create a pull request by myself.

You have already made a PR earlier. What's the problem this time?
The developers are active on Discord https://discord.com/invite/3sc5su7
I suggest you use the channel `#dev-support` for help with PR's.

deenadayalancs:
@zonkmachine thanks for the details. I tried to edit from my desktop client. guess, something is missing there.  updated and merged the code.

Have a nice day.

DomClark:
Fixed in #5890.

PR #5890, opened 2021-01-21
Fix MP3 stereo mode doesn't match to the selection in export dialog

deenadayalancs:
Fix for MP3 stereo mode doesn't match to the selection in export dialog #5876

Updated the code to match the #5876.

### DIFF
--- a/src/gui/ExportProjectDialog.cpp
+++ b/src/gui/ExportProjectDialog.cpp
@@ -144,7 +144,7 @@ OutputSettings::StereoMode mapToStereoMode(int index)
 	case 0:
 		return OutputSettings::StereoMode_Mono;
 	case 1:
- 		return OutputSettings::StereoMode_Stereo;
+		return OutputSettings::StereoMode_Stereo;
 	case 2:
 		return OutputSettings::StereoMode_JointStereo;
 	default:
--- a/src/gui/ExportProjectDialog.cpp
+++ b/src/gui/ExportProjectDialog.cpp
@@ -142,11 +142,11 @@ OutputSettings::StereoMode mapToStereoMode(int index)
 	switch (index)
 	{
 	case 0:
-		return OutputSettings::StereoMode_Stereo;
-	case 1:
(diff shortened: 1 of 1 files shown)


src/gui/ExportProjectDialog.cpp
```diff
@@ -142,11 +142,11 @@ OutputSettings::StereoMode mapToStereoMode(int index)
...
-		return OutputSettings::StereoMode_Stereo;
+		return OutputSettings::StereoMode_Mono;
 	case 1:
-		return OutputSettings::StereoMode_JointStereo;
+ 		return OutputSettings::StereoMode_Stereo;
 	case 2:
-		return OutputSettings::StereoMode_Mono;
+		return OutputSettings::StereoMode_JointStereo;
 	default:
 		return OutputSettings::StereoMode_Stereo;
```
zonkmachine, line 151: Please add an extra tab to everything withing the brackets
```
{
    like
    this
}
```

zonkmachine (commented):
I've tested the PR and it fixes the issue. I suggest some style changes while you're at it. See comment below.

src/gui/ExportProjectDialog.cpp
PhysSong: We don't have a rule for `switch` statement, so it's fine.

src/gui/ExportProjectDialog.cpp
```diff
@@ -142,11 +142,11 @@ OutputSettings::StereoMode mapToStereoMode(int index)
 	switch (index)
 	{
 	case 0:
-		return OutputSettings::StereoMode_Stereo;
+		return OutputSettings::StereoMode_Mono;
 	case 1:
-		return OutputSettings::StereoMode_JointStereo;
+ 		return OutputSettings::StereoMode_Stereo;
```
PhysSong, line 147: Please remove a space at the beginning of this line.
```suggestion
		return OutputSettings::StereoMode_Stereo;
```

Issue #3393, opened 2017-03-01
import/export native pattern data


PaulBatchelor:
What is the feasibility with regards to importing/exporting LMMS pattern data?

I would imagine it would save the data in a similar way that it is represented inside the mmpz format (last I checked, this was some sort of XML, but I could be wrong here). It could then be loaded similarly to how you load a project. 

My thinking here is that this could be a good step forward towards using external programs to generate pattern data, which would be a huge win in my opinion.

jasp00:
>What is the feasibility with regards to importing/exporting LMMS pattern data?

It depends on what you want to do. LMMS needs to export and import data to copy between program instances.

PaulBatchelor:
I want to write programs to generate patterns for me.

musikBear:
> programs to generate pattern data, which would be a huge win in my opinion.

With the term *pattern data* you @PaulBatchelor mean notes, right?
Eg program generated music.
I had an exchanges of view-points on this, with an user that wanted that, some weeks ago. I must say, im against it. Computer generated midi-patterns are.. idk ..'wrong' 
I like that programs can aid creativity, **just** like lmms, for whitc without, i personally never would have made any music -ever!
But having the program making the creative parts as well..?

jasp00:
>I want to write programs to generate patterns for me.

If your program is GPL-compatible (MIT, BSD, etc.) or only used by you, it can be integrated as a tool plug-in. For other cases, you may import data through the file browser, drag and drop and the like. Fixes may be needed, but the feature is certainly feasible.

PR #5891, opened 2021-01-21
Feature: Pattern import/export

cyber-bridge:
Continuation of https://github.com/LMMS/lmms/pull/5197

This makes it possible to export and import patterns. They will be saved to a XML document. Currently with the file-extension `.xpt` and `.xtpz` for compressed files.

Closes https://github.com/LMMS/lmms/issues/3393

### DIFF
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -4408,7 +4408,7 @@ PianoRollWindow::PianoRollWindow() :
 	notesActionsToolBar->addAction( quantizeAction );
 
 	// -- File actions
-	DropToolBar *fileActionsToolBar = addDropToolBarToTop( tr( "File actions" ) );
+	DropToolBar* fileActionsToolBar = addDropToolBarToTop(tr("File actions"));
 
 	// -- File ToolButton
 	m_fileToolsButton = new QToolButton(m_toolBar);
@@ -4416,11 +4416,11 @@ PianoRollWindow::PianoRollWindow() :
 	m_fileToolsButton->setPopupMode(QToolButton::InstantPopup);
 
 	// Import / export
-	QAction * importAction = new QAction(embed::getIconPixmap("project_import"),
-				tr("Import pattern"), m_fileToolsButton);
+	QAction* importAction = new QAction(embed::getIconPixmap("project_import"),
+		tr("Import pattern"), m_fileToolsButton);
 
-	QAction * exportAction = new QAction(embed::getIconPixmap("project_export"),
-				tr("Export pattern"), m_fileToolsButton);
+	QAction* exportAction = new QAction(embed::getIconPixmap("project_export"),
+		tr("Export pattern"), m_fileToolsButton);
 
 	m_fileToolsButton->addAction(importAction);
 	m_fileToolsButton->addAction(exportAction);
@@ -4845,10 +4845,10 @@ void PianoRollWindow::exportPattern()
 		!exportDialog.selectedFiles().isEmpty() &&
 		!exportDialog.selectedFiles().first().isEmpty())
 	{
-		QString suffix = ConfigManager::inst()->value( "app",
-								"nommpz" ).toInt() == 0
-							? "xptz"
-							: "xpt" ;
+		QString suffix =
+			ConfigManager::inst()->value("app", "nommpz").toInt() == 0
+				? "xptz"
+				: "xpt";
 		exportDialog.setDefaultSuffix(suffix);
 
 		QString fullPath = exportDialog.selectedFiles()[0];
@@ -4856,7 +4856,7 @@ void PianoRollWindow::exportPattern()
 
 		bool compress = (chosenSuffix == "xpt") ? false : true;
 
-		if (savePatternXML(fullPath, compress) != 0)
+		if (!savePatternXML(fullPath, compress))
 		{
 			TextFloat::displayMessage(tr("Export pattern failed"),
 				tr("No permission to write to %1").arg(fullPath),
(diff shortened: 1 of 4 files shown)


src/gui/editors/PianoRoll.cpp
```diff
@@ -4803,6 +4834,116 @@ void PianoRollWindow::ghostPatternSet( bool state )
...
+	exportDialog.setAcceptMode(FileDialog::AcceptSave);
+
+	if (exportDialog.exec() == QDialog::Accepted &&
+		!exportDialog.selectedFiles().isEmpty() &&
+		!exportDialog.selectedFiles().first().isEmpty())
+	{
+		QString suffix = ConfigManager::inst()->value( "app",
+								"nommpz" ).toInt() == 0
+							? "xptz"
+							: "xpt" ;
```
IanCaio, line 4851:
```suggestion
		QString suffix =
			ConfigManager::inst()->value("app", "nommpz").toInt() == 0
				? "xptz"
				: "xpt";
```

src/gui/editors/PianoRoll.cpp
```diff
@@ -4803,6 +4834,116 @@ void PianoRollWindow::ghostPatternSet( bool state )
...
+
+	FileDialog importDialog(this, tr("Open Pattern"), "",
+			tr("XML pattern file (*.xpt *.xptz)"));
+	importDialog.setFileMode(FileDialog::ExistingFiles);
+
+	if (importDialog.exec () == QDialog::Accepted &&
+		!importDialog.selectedFiles().isEmpty())
+	{
+		QString fullPath = importDialog.selectedFiles()[0];
+		QString suffix = fullPath.section( '.', -1 );
```
IanCaio, line 4897:
```suggestion
		QString suffix = fullPath.section('.', -1);
```

src/gui/editors/PianoRoll.cpp
```diff
@@ -4803,6 +4834,116 @@ void PianoRollWindow::ghostPatternSet( bool state )
...
+	{
+		return;
+	}
+
+	FileDialog importDialog(this, tr("Open Pattern"), "",
+			tr("XML pattern file (*.xpt *.xptz)"));
+	importDialog.setFileMode(FileDialog::ExistingFiles);
+
+	if (importDialog.exec () == QDialog::Accepted &&
+		!importDialog.selectedFiles().isEmpty())
```
IanCaio, line 4894:
```suggestion
	if (importDialog.exec() == QDialog::Accepted &&
		!importDialog.selectedFiles().isEmpty())
```

Extra space before parenthesis.

I also got curious, why does the export requires a `!selectedFiles().first().isEmpty()` check and the import doesn't? Is it because the file mode is `ExistingFiles`, so it will only accept the dialog with an existing file name?

IanCaio (changes requested):
Code looks good, this review mostly points some code style or formatting minor issues. There's also one suggestion on using `bool` instead of `int` as a return type of `savePatternXML`. I think it makes more sense as a return type, even though what you have now is POSIX compliant :smile:

cyber-bridge:
@IanCaio thanks for the review, I've made the suggested changes.

> I also got curious, why does the export requires a `!selectedFiles().first().isEmpty()` check and the import doesn't? Is it because the file mode is `ExistingFiles`, so it will only accept the dialog with an existing file name?

Yes it is.

However this question did make me look at that code again and noticed that `FileDialog::ExistingFiles` was used instead of `FileDialog::ExistingFile`; so also changed that.

src/gui/editors/PianoRoll.cpp
```diff
@@ -4803,6 +4834,118 @@ void PianoRollWindow::ghostPatternSet( bool state )
...
+		{
+			TextFloat::displayMessage(tr("Import pattern failed"),
+				tr("Pattern file corrupt %1").arg(fullPath),
+				embed::getIconPixmap("error"), 4000);
+			f.close();
+			return;
+		}
+		f.close();
+
+		int pos = m_editor->m_pattern->startPosition(); // Backup position in timeline.
```
IanCaio, line 4935:
```suggestion
		TimePos pos = m_editor->m_pattern->startPosition(); // Backup position in timeline.
```

I know it works because `TimePos` is converted to `int` without issues, but I think it's better to use the former.

IanCaio (approved):
Thanks for the quick response!

Code LGTM! I'm leaving a last change request: It's a minor thing, code works fine the way it's written but I think it's a better habit for us to use `TimePos` instead of `int` where it's expected, even though they can be interchangeable.

I'm approving the PR despite that last change suggestion. Nice work!

cyber-bridge:
@IanCaio

> I think it's a better habit for us to use TimePos instead of int where it's expected, even though they can be interchangeable.

Agreed :-)

PhysSong:
Looks good. Could you look into `DataFile` class so it's possible to use it for data serialization?

superpaik:
It works well to me.
Only one unexpected behaviour. The folder that is opened to both import/export is the last folder used (being that by import/export, audiofileprocessor -select sample-, sampletrack -select track-, save project, etc) even if cancel action on selecting the folder is been pressed. I don't know if that's the best option from the user point of view. Maybe the default behaviour should be: if no import/export have been executed, then project folder (if exists, if not current folder whatever that is); if import/export have been executed, then that should be the folder used for the next import/export. Just a suggestion.

cyber-bridge:
@PhysSong that's a cool class, adopted the code to use it.

@superpaik I get the issue; from what I've tested all other LMMS file dialogs have pre-defined paths (example: {LMMS_HOME}/projects or {LMMS_HOME}/samples).

Maybe create a default `pattern` directory and use that as default?

Also would it be nice to be able to drag and drop from the sidebar filebrowser into a `InstrumentTrack`?

IanCaio:
Recent changes look good to me!

I agree that it would be nice to have a pattern folder, maybe accessible from the sidebar, but could we leave this for a future enhancement? We're running a little bit against time to make the code base refactoring happen, so maybe we can keep the feature in its state, which is perfectly functional, and leave the improvements for after the refactor?

cyber-bridge:
Awesome :-)

Yeah leaving it for what it is sounds OK to me. Once it's merged I will create another PR for drag-drop from the sidebar.

On the default directory idea: I thought there was a issue on exporting automation tracks, now I can't find it :') But if that's the case it will probably better to use `{LMMS_HOME}/patterns/notes` so potential future (other) patterns can have their own sub-directories. But yeah.. that's future stuff then.

PhysSong:
A remark on names: `Pattern` can be renamed to something like `MidiClip` during the reorganization(see #5592).

src/gui/editors/PianoRoll.cpp
```diff
@@ -4803,6 +4832,84 @@ void PianoRollWindow::ghostPatternSet( bool state )
 
 
 
+void PianoRollWindow::exportPattern()
+{
+	QString extFilter("XML Pattern (*.xpt *.xptz)");
+	FileDialog exportDialog(this, tr("Export pattern"), "", extFilter);
```
PhysSong, line 4838: There are some minor inconsistencies between `exportPattern()` and `importPattern()`:
- `XML Pattern` vs. `XML pattern file`. I'm not sure if `XML pattern` describe well what the file is.
- This file extension filter in this function is not translatable while the one in `importPattern()` is. Also, only this function uses an explicit temporary `QString`.

PhysSong:
@cyber-bridge Have you read my latest comment?

cyber-bridge:
Yes thanks for reminding me! Would "notes pattern" be more descriptive?

IanCaio:
@PhysSong You think this PR is ready for merge after those latest changes?

PhysSong:
> @PhysSong You think this PR is ready for merge after those latest changes?

Yes.

PR #5895, opened 2021-01-30
Toggle ghost notes

allejok96:
Toggle, set and paste ghost notes in Piano roll.

Before:
bild

After:
bild

### Toggle ghost notes
The button should only view/hide ghost notes. Setting ghost notes will toggle the button to be on. Clear ghost notes button is removed from lower toolbar. Ghost notes are only cleared when resetting the project, or setting an empty pattern as ghost. Included style change for buttons with menu (copied from #5857).
 
### Current pattern as ghost
Should do exactly the same as right clicking a clip in Song editor and it as ghost.

### Paste as ghost
Should work exactly as Paste in Piano Roll (with regards to positioning and quantizing) but create ghost notes instead of normal notes. Only difference is that previous ghost notes are cleared before pasting the new ones.

### ~Guess scale~
~This selects the most likely key in the *Major* scale in Piano Roll based on the notes in the pattern. The selection is based on how many notes are "in scale" and, if more than one scale matches, [how common the notes are in pop song chords](https://www.hooktheory.com/blog/i-analyzed-the-chords-of-1300-popular-songs-for-patterns-this-is-what-i-found/) (yeah, don't ask how I crunched the numbers). I tossed this feature in this PR because it touched code in the same places...~


src/gui/editors/PianoRoll.cpp
```diff
@@ -4086,6 +4140,42 @@ void PianoRoll::pasteNotes()
 }
 
 
+void PianoRoll::pasteGhostNotes()
+{
+	if(!hasValidPattern()) { return; }
+
+	QString value = Clipboard::getString(Clipboard::MimeType::Default);
+
+	if(!value.isEmpty())
```
Veratil, line 4149: I'd rather see the inverse of this. Put the failure state inside the `if` instead of the success state.

Veratil (commented):
Other than many style fixes, I like the toggle and paste functions.

allejok96:
I might revert the Guess scale feature and make it a separate PR... It's not very accurate right now given only a single pattern. A lot of times it cannot tell a difference between C major and G major for example. Music is more complicated than counting the *amount* of notes on "important" keys. Much better if it could look at the whole project when doing the analysis. That's why I'm thinking of tucking it away in the global Tools menu.

But the key/scale settings are inside the Piano roll window, so placing the button there makes sense. Question is: would anyone find a Guess scale button in Piano roll useful?

IanCaio:
Just explaining the `after-refactor` tag: We've been planning a code refactor for a while and that depends on us clearing the PR log. For that reason, a while ago we started sort of a "PR-freeze": The PRs opened after that wouldn't be discarded, but they would have to be rebased after the refactor happened to fix any conflicts.

It's still possible that this PR will be merged before the refactor, but it's being tagged because I think it's only fair to merge a PR that was opened after the PR-freeze if we can merge **all** PRs that were opened after the PR-freeze. So it will all depend on how many PRs will be opened and how many reviewers and time left we have until the refactor begins.

Nothing to be majorly concerned about because even if this doesn't get merged before the refactor it will still be reviewed after it. Small PRs like this one shouldn't be too difficult to rebase, specially if the author accompanies the refactoring and does it incrementally as we advance with it.

src/gui/editors/PianoRoll.cpp
```diff
@@ -4145,36 +4145,34 @@ void PianoRoll::pasteGhostNotes()
...
+	}
 
-		QDomNodeList list = dataFile.elementsByTagName(Note::classNodeName());
-
-		if(!list.isEmpty())
-		{
-			m_ghostNotes.clear();
+	if(list.isEmpty())
+	{
+		TextFloat::displayMessage(tr("Paste as ghost failed"),
```
Veratil, line 4158: Should this return after this or continue clearing the ghost notes and all that?

allejok96: Oh sorry I forgot to add return there

allejok96:
Removed Guess scale, it does not belong in this PR. I'm looking into some more sophisticated key finding algorithms.

allejok96:
Superseded by 6527

Issue #2074, opened 2015-05-27
Zero-length beat notes end too soon


musikBear:
a Q in forum made me do some experiments with B&B and audio-loops over 1-2 bars (typical commercial percussion loops, f.i) 
They are actual quite difficult to get to play right _and_ loop right
Only way i can get this somewhat working, is to use very long notes, but then the _'only-play-from-note-start'_ cravat kicks in
Could the green no-length-notes in BB have a logic, where they always played _exactly_ for the duration of the setting in AFP, then perfect looping could be easy for the user

tresf:
Zero-length notes play a sample until completion.  Why isn't that working for you?

-Tres

musikBear:
> Why isn't that working for you?
Good question - I also thought so , but after i tested, it only plays for one bar.
Try latin_guitar01.ogg. That samples length is ~3.2 bar in length
A no-length-note play this sample for 1 bar, then cuts off
To reproduce you need piano-roll loop-points 0..4 bar, in order to force the playhead past the one bar vertical, but the no-length-note stops playing
It surprised me too 8|

musikBear:
Have anyone re-tested this behavior, as described?
If no-length-notes are supposed always to play the whole sample, there is a bug (1.1.3 x32)

tresf:
No, no one has re-tested this.  Perhaps the warm weather in the northern hemisphere combined with academia has the helpers with other endeavors. :sun_with_face:

Wallacoloo:
@musikBear I just tested the behavior. By the way, for anyone reading this not knowing what AFP stands for (me, previously), it's the "Audio File Processor" - the sampler that is implicitly instantiated when you drag a sample into the instrument track list.

I opened the Beat/Bassline editor, dragged latin/latin_guitar01.ogg into it and created a note on the first beat. Playing the pattern resulted in the sample being played for just UNDER a full bar before being silenced.

Lowering the tempo after doing this results in the already instantiated notes still playing for the same actual duration (i.e. same number of seconds, but now for fewer beats). New instruments and notes are ALSO affected, playing for exactly the same duration as the previous notes and not synced with the tempo. This suggests to me that the duration of zero-length notes are **not based on the tempo** - it's probably a fixed number of samples or seconds.

musikBear:
@Wallacoloo Thanks for confirming
 its a bug, so il post a new ticket

tresf:
Thanks Colin.  Sounds like it bares some resemblance to #1592.

-Tres

[...]

PR #5902, opened 2021-02-03
Implement Note Types

### LINKED ISSUE #5339: Zero-length notes copied from the B&B editor cause incorrect behavior in the song editor

### LINKED ISSUE #5340: zero-length-notes added to normal-notes does only play the bar with the normal-note

IanCaio:
This PR moves away from using negative length for notes that are supposed to be played for the length of the sample. Instead, we now use Note Types, which describe the type of note that is being played (for now, either `RegularNote` or `StepNote`). The length of a step note is now 1/16th by default, but it doesn't matter if you resize it because `InstrumentTrack::play` will create a NotePlayHandle with 0 frames if the type of the note is `StepNote`.

The StepNotes are drawn now as red notes on the PianoRoll. BeatPattern TCOs are drawn in the Song Editor with the assigned BBPatternBackground color (which is grey on default theme).

On the image, TripleOsc with a MelodyPattern and AFP with a BeatPattern and a MelodyPattern that has both step notes and regular notes (opened on the PianoRoll):
NoteType

There's a `qWarning` left on `TimePos::frames` that is supposed to be removed. Keeping it just to check if there's any code left creating negative length notes. Didn't rush to remove it since I'm not even sure this will be reviewed before the refactor.

Fixes #5339
Fixes #5340


zonkmachine:
This also closes #1681. There is an issue brought up somewhere around https://github.com/LMMS/lmms/issues/1681#issuecomment-277537185 concerning representation of beat notes in tco's that perhaps should have it's own ticket. It's fixed in this PR.

IanCaio:
> This also closes #1681. There is an issue brought up somewhere around [#1681 (comment)](https://github.com/LMMS/lmms/issues/1681#issuecomment-277537185) concerning representation of beat notes in tco's that perhaps should have it's own ticket. It's fixed in this PR.

True, it partially solves it. The notes that are off-beat will still not show up on the BB Step Note editor and neither will it be displayed as a TCO on that case. I think it's better to leave it that way until we can implement allowing step notes to be added from the Piano Roll.

I'm about to push the upgrade routine which I totally forgot. Can't load older projects properly without it.

IanCaio:
Done. Upgrade method was added, older projects should be loaded properly now.

PhysSong:
Note: MIDI export may require some additional changes.

thmueller64:
I tested this PR and it works as intended. The PR https://github.com/LMMS/lmms/pull/5502 should be closed in favor of this one.
There is one thing that #5502 handles, which should also be taken care of here:

1. Draw pattern in BBEditor
2. Right click on the pattern --> Open in PianoRoll
3. Copy the notes and paste them at the end

Expected Result:

4. Length of the pattern shown in the BBEditor changes accordingly

Actual Result:

4. Length of the BB pattern stays the same (the pasted notes are hidden essentially)

In the linked PR, this is handled with the `adjustSteps` function.

IanCaio:
> Note: MIDI export may require some additional changes.

True, it might not though because it uses `saveSettings` I guess. So if I update it on the note class I guess it should save and load properly on the XML pattern file.

> I tested this PR and it works as intended. The PR #5502 should be closed in favor of this one.
> There is one thing that #5502 handles, which should also be taken care of here:

You're right. I'm not sure there's an open issue for that (it seems to be the behavior on master) but we can fix it too. I was looking how you did it, you added an `adjustStep` call on `addNote`, there should be one on `removeNote` too right? I thought about connecting it to `dataChanged` instead but quickly looking I see there might be issues, as other methods like `addStep` also emit it and it might recalculate the steps unnecessarily more than once. I just worry that `addNote` and `removeNote` doesn't cover every situation where the pattern changes (was thinking undo/redo for example, or copy/paste, even import pattern in the future).

PhysSong:
> Note: MIDI export may require some additional changes.

I wrote this warning because it depends on the negative length of BB notes.

thmueller64:
> there should be one on removeNote too right?

I deliberately didn't add it there:
Say the user clicks on "Add steps" and creates a pattern on the added steps. If the user deselects all of the steps in the extended range for a moment, the range is shrunk immediately. I thought this would be annoying.

> I just worry that addNote and removeNote doesn't cover every situation where the pattern changes (was thinking undo/redo for example, or copy/paste, even import pattern in the future).

undo/redo, copy/paste works fine, I testet it with #5502. We can open a separate issue for this one.

IanCaio:
> I deliberately didn't add it there:
> Say the user clicks on "Add steps" and creates a pattern on the added steps. If the user deselects all of the steps in the extended range for a moment, the range is shrunk immediately. I thought this would be annoying.

Oh, that makes sense, so it would only grow automatically but you'd need to remove the steps manually to shrink it again.

> undo/redo, copy/paste works fine, I testet it with #5502. We can open a separate issue for this one.

Awesome, I thought there might be some issues because some of them use the XML state and doesn't make calls to `addNote` (`loadSettings`), but then, they also copy the steps value, so it should be fine. `PianoRoll::pasteNotes` seems to use `addNote` to paste the notes so it seems to behave well too.

I think a separate PR would be better in the testing/bug-hunting aspect, but that's definitely something worth adding. Or I could just add it to this one, I don't mind it neither :grin:

src/tracks/Pattern.cpp
```diff
@@ -295,10 +290,21 @@ void Pattern::clearNotes()
...
+			Note(TimePos(DefaultTicksPerBar / 16),
+			TimePos::stepPosition(step)), false
+		);
+	n->setType(Note::StepNote);
+	// Have to check the type again now that we changed
+	// the type of the note
+	// TODO: This means checkType is called twice:
+	// Once on addNote and once here. Rethink it so we
+	// can call it only once.
+	checkType();
```
Veratil, line 306: If `m_patternType` was `MelodyPattern` before `addNote`, then it will stay `MelodyPattern` afterwards.
If `m_patternType` was `BeatPattern` before `addNote`, then it will stay `BeatPattern` afterwards.

Therefore, if we grab `m_patternType` before calling `addNote`, we should be able to simply know if we need to keep the same type or change.

```suggestion
	auto prev_patternType = m_patternType;
	Note * n =
		addNote(
			Note(TimePos(DefaultTicksPerBar / 16),
			TimePos::stepPosition(step)), false
		);
	n->setType(Note::StepNote);
	// addNote will change to MelodyPattern with addNote (calls checkType),
	// we need to keep it BeatPattern if it was beforehand
	if (prev_patternType == BeatPattern) { setType(BeatPattern); }
```

src/tracks/Pattern.cpp
```diff
@@ -938,9 +955,74 @@ void PatternView::paintEvent( QPaintEvent * )
 
 	const int x_base = TCO_BORDER_WIDTH;
 
-	// melody pattern paint event
 	NoteVector const & noteCollection = m_pat->m_notes;
-	if( m_pat->m_patternType == Pattern::MelodyPattern && !noteCollection.empty() )
+
+	// Beat pattern paint event
```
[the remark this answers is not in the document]
IanCaio: Sure, I ended up not doing it because I moved a big block but didn't change it and wasn't sure how it would show in the diff. Can I leave this style fix for after it's decided whether this will be merged before or after refactor? Because if it's merged after I'm probably going to have to rebase against the fixed-style master anyways

src/tracks/Pattern.cpp
IanCaio: That's a good idea! I'll change the code to use it

IanCaio:
Done! Except for the code style formatting of the big block of code that was moved.

I was thinking, another alternative was to change `addNote` so it had a type as a parameter and having RegularNote as default. But the current solution is good and I'm keeping it for now.

src/tracks/Pattern.cpp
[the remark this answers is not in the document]
Veratil: I'm okay with leaving for clang PR.

src/tracks/Pattern.cpp
```diff
@@ -339,16 +345,27 @@ void Pattern::setType( PatternTypes _new_pattern_type )
...
-	setType( BeatPattern );
+
+	if (beatPattern)
+	{
+		setType(BeatPattern);
+	}
+	else
+	{
+		setType(MelodyPattern);
+	}
```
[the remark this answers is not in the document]
PhysSong: `setType(beatPattern ? BeatPattern : MelodyPattern);` is shorter, BTW.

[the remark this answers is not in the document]
IanCaio: Changed to your suggestion. I went for the one line `if` at first because I'm not a fan of using the ternary to run a statement instead of returning a value (as in `beatPattern ? setType(BeatPattern) : setType(MelodyPattern);`), but inside the method call as a way to choose the argument looks cleaner to me and saves a line.

superpaik:
It works fine to me, except in one case.
If the sample duration is longer than the beat, it seems that ends before the beat ends. Check with house_loop01.ogg at 140bpm, it sounds for like 3 and 3/4 beats instead of 4 beats. There is a silence between beats. It doesn't happen if you play the sample as an audiofile directly for 4 beats.

Not strictly related to the functionality but more of UX. I think it would be great to have the possibility to add new Step Notes directly from the pianoroll (now it's only possible through copy-and-paste). Maybe a tool/option to change between type notes would be great.

IanCaio:
> It works fine to me, except in one case.
> If the sample duration is longer than the beat, it seems that ends before the beat ends. Check with house_loop01.ogg at 140bpm, it sounds for like 3 and 3/4 beats instead of 4 beats. There is a silence between beats. It doesn't happen if you play the sample as an audiofile directly for 4 beats.

I think the issue you experienced is likely this one #2074 . It's probably more noticeable in the piano roll than it is in the BB editor though.

> Not strictly related to the functionality but more of UX. I think it would be great to have the possibility to add new Step Notes directly from the pianoroll (now it's only possible through copy-and-paste). Maybe a tool/option to change between type notes would be great.

That's the idea! The PR currently only implements the base for allowing that, but following enhancements should allow you to add this type of note directly from the PianoRoll. I even envision other types of notes in the future, like "Arpeggio" notes, so we can move away from the currently buggy Arpeggiator and have a more flexible and visual one.

For now, since this will possibly only be merged after the refactor I'd rather just add the base functionality for now and leave those enhancements for later. We'd have to decide how it would work though: Would the Draw mode icon have a dropdown menu where you chose the type of note? Or a context menu for notes on the piano roll? Maybe even both, thought it would require UX changes since the right click is currently used for deleting notes.

superpaik:
>> I think the issue you experienced is likely this one #2074 . It's probably more noticeable in the piano roll than it is in the BB editor though.

Yes, it seems the same bug.

As for your other comment, I agree with you. Let's wait for refactoring.

Some ideas as how I see they could work (for the future): I think that, in general, you'd use Regular or Step Notes, but I think it's kind of weird to mix them up (I'm not saying that it shouldn't be addressed, but I see it as an exception). So, if the piano roll is open form inside the BBEditor, I'd set the draw to Step notes by default, and if it's open  from the song editor, the Regular Notes should by the default. Maybe, to change the draw mode, just by clicking on the draw icon it could go from one type to the other. Also, the "last note" draw option should have into account the different type notes, so it's easier for the user to change note type.

IanCaio:
Reminder to self:

https://github.com/LMMS/lmms/blob/e880e3cb2a02c6d2906bb0303699ab3c7e32f331/src/gui/editors/PianoRoll.cpp#L1591-L1596

This piece of code on the `PianoRoll::mousePressEvent` is there because of negative length notes and with this PR will not be necessary anymore.

Out of the scope of this PR but `mousePressEvent` needs some serious refactoring. Right after that code block there's this conditional:

https://github.com/LMMS/lmms/blob/e880e3cb2a02c6d2906bb0303699ab3c7e32f331/src/gui/editors/PianoRoll.cpp#L1599-L1600

`len > 0` will never be false since it's set to 4 when it's negative and as far as I know there aren't notes with length equal to 0. I'll hold any attempt on refactoring this method for after the PRs that affect it are merged though (IIRR only #5848 and #5857)

Veratil:
> Out of the scope of this PR but mousePressEvent needs some serious refactoring.

I've been slowly hacking at all mouse events in PianoRoll. I have a branch locally which has rewritten a lot to use switch's and organizing the code to be a lot more readable. It's still a big WIP though.

IanCaio:
> I've been slowly hacking at all mouse events in PianoRoll. I have a branch locally which has rewritten a lot to use switch's and organizing the code to be a lot more readable. It's still a big WIP though.

I was actually thinking of asking if you wanted to team up fixing those mouse event methods, maybe even coming up with controls that were more up-to-date to the current new features: One thing I was thinking of was implementing a context-menu, we could move the 4 note tools and the glue tool to that context menu, opening space for many others without filling the window toolbar.

It's awesome that you have this WIP going already, just moving to a more organized `switch` will improve it greatly :smile:

IanCaio:
Conflicts with master fixed

IanCaio:
Conflicts from recent file splits are resolved

src/gui/editors/PianoRoll.cpp
```diff
@@ -3499,11 +3499,15 @@ void PianoRoll::paintEvent(QPaintEvent * pe )
 			// is the note in visible area?
 			if (note->key() > bottomKey && note->key() <= topKey)
 			{
-				// we've done and checked all, let's draw the note
+				// We've done and checked all, let's draw the note with
+				// the appropriate color
+				QColor fillColor = note->type() == Note::RegularNote ? m_noteColor : m_stepNoteColor;
```
sakertooth, line 3504: Use ``const auto``. Once again, mostly subjective, but knowing the type here is not too important (we actually do know it already by intuition, lol).
```suggestion
				const auto fillColor = note->type() == Note::RegularNote ? m_noteColor : m_stepNoteColor;
```

src/tracks/MidiClip.cpp
```diff
@@ -174,19 +174,18 @@ TimePos MidiClip::beatClipLength() const
...
-	if( m_steps != TimePos::stepsPerBar() )
+	if (m_steps != TimePos::stepsPerBar())
 	{
-		max_length = m_steps * TimePos::ticksPerBar() /
-						TimePos::stepsPerBar();
+		max_length = m_steps * TimePos::ticksPerBar() / TimePos::stepsPerBar();
 	}
 
-	return TimePos( max_length ).nextFullBar() * TimePos::ticksPerBar();
+	return TimePos(max_length).nextFullBar() * TimePos::ticksPerBar();
```
sakertooth, line 188: Guidelines suggest to use ``{}`` rather than ``()`` to [avoid parsing ambiguities](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es23-prefer-the--initializer-syntax) in the event that they may occur. Just a guideline that may not have any real impact in practice depending on the code, so make the switch if you so choose.

src/tracks/MidiClip.cpp
```diff
@@ -281,8 +280,19 @@ void MidiClip::clearNotes()
 
 Note * MidiClip::addStepNote( int step )
 {
-	return addNote( Note( TimePos( -DefaultTicksPerBar ),
-				TimePos::stepPosition( step ) ), false );
+	auto prevPatternType = m_clipType;
```
sakertooth, line 283:
```suggestion
	const auto prevPatternType = m_clipType;
```

src/tracks/MidiClip.cpp
```diff
@@ -355,16 +365,20 @@ void MidiClip::setType( MidiClipTypes _new_clip_type )
 void MidiClip::checkType()
 {
 	NoteVector::Iterator it = m_notes.begin();
-	while( it != m_notes.end() )
+
+	// If all notes are StepNotes, we have a BeatClip
+	bool beatClip = true;
+	while (it != m_notes.end())
```
sakertooth, line 371: Since ``QVector`` provides ``begin()`` and ``end()`` members, we could use a ranged-based for loop (Or even better, use ``std::find_if`` or ``std::count_if`` to make this code far more compact. Any one works, although ``std::find_if`` expresses intent more clearly IMO).

src/tracks/MidiClip.cpp
```diff
@@ -238,13 +237,13 @@ void MidiClip::removeNote( Note * _note_to_del )
 }
 
 
-// returns a pointer to the note at specified step, or NULL if note doesn't exist
-
-Note * MidiClip::noteAtStep( int _step )
+// Returns a pointer to the note at specified step, or nullptr if note doesn't exist
+Note * MidiClip::noteAtStep( int step )
```
sakertooth, line 241:
```suggestion
Note * MidiClip::noteAtStep(int step)
```

include/Note.h
```diff
@@ -247,6 +261,9 @@ class LMMS_EXPORT Note : public SerializingObject
 	TimePos m_length;
 	TimePos m_pos;
 	DetuningHelper * m_detuning;
+
+	// The type of this note
+	Type m_type;
```
sakertooth, line 266: I would suggest to also use an in-class initializer here.
```suggestion
	Type m_type = Types::RegularNote;
```

src/tracks/InstrumentTrack.cpp
```diff
@@ -758,8 +758,11 @@ bool InstrumentTrack::play( const TimePos & _start, const fpp_t _frames,
 		while( nit != notes.end() &&
 					( cur_note = *nit )->pos() == cur_start )
 		{
-			const f_cnt_t note_frames =
-				cur_note->length().frames( frames_per_tick );
+			// If the note is a Step Note, frames will be 0 so the NotePlayHandle
+			// plays for the whole length of the sample
+			const f_cnt_t note_frames = cur_note->type() == Note::StepNote
```
sakertooth, line 763:
```suggestion
			const auto note_frames = cur_note->type() == Note::StepNote
```

src/tracks/MidiClip.cpp
```diff
@@ -281,8 +280,19 @@ void MidiClip::clearNotes()
...
 {
-	return addNote( Note( TimePos( -DefaultTicksPerBar ),
-				TimePos::stepPosition( step ) ), false );
+	auto prevPatternType = m_clipType;
+	Note * n =
+		addNote(
+			Note(TimePos(DefaultTicksPerBar / 16),
+			TimePos::stepPosition(step)), false
+		);
+	n->setType(Note::StepNote);
```
sakertooth, line 289: Scoped enums.
```suggestion
	n->setType(Note::Types::StepNote);
```

src/core/TimePos.cpp
```diff
@@ -161,11 +161,11 @@ tick_t TimePos::getTickWithinBeat( const TimeSig &sig ) const
...
 {
-	if( m_ticks >= 0 )
-	{
-		return static_cast<f_cnt_t>( m_ticks * framesPerTick );
-	}
-	return 0;
+	//TODO: Remove next line, it's just a warning to see
+	// if any leftover code is still creating notes with
+	// negative lengths
+	if (m_ticks < 0) { qWarning("TimePos with negative length"); }
```
sakertooth, line 167: Was going to comment on this in the first review, but I wasn't sure if having it would be necessary or not given the surrounding code. I just noticed you said it can be removed, although I think it could be replaced with an ``assert`` (assuming that negative lengths should never happen)?

include/Note.h
```diff
@@ -101,6 +101,20 @@ class LMMS_EXPORT Note : public SerializingObject
 	Note( const Note & note );
 	~Note() override;
 
+	// Note types
+	enum Types
```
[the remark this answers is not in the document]
sakertooth: I would actually rename ``Types`` to ``Type`` as it seems that other ``enums`` around the codebase with the same name dont have the ending ``s``. Not too sure if theres an actual preference or not though.

src/core/Note.cpp
```diff
@@ -47,7 +47,8 @@ Note::Note( const TimePos & length, const TimePos & pos,
 	m_panning( qBound( PanningLeft, panning, PanningRight ) ),
 	m_length( length ),
 	m_pos( pos ),
-	m_detuning( nullptr )
+	m_detuning( nullptr ),
```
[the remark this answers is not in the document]
IanCaio: I ended up keeping this change out, since the PR doesn't involve the detuning member. The line is only changed because I needed to add a comma, maybe since it's a member from another feature the change should be done separately, just to make sure the it doesn't affect other code?

src/core/TimePos.cpp
IanCaio: The idea was to remove this completely before merging, negative length notes should not happen anymore after this change. I can also replace it with an assert, if there's interest in keeping the test here.

src/tracks/MidiClip.cpp
IanCaio: From what I read, since the conditional is a "different-than" I would have to use `std::not1` with a conditional bool method, which I think kills a bit the purpose of simplifying this code. But I changed it to use a range-based loop!

IanCaio:
Fixed some of the issues pointed out on @sakertooth 's review, others are still to be worked on.

src/tracks/MidiClip.cpp
sakertooth: > From what I read, since the conditional is a "different-than" I would have to use `std::not1` with a conditional bool method, which I think kills a bit the purpose of simplifying this code. But I changed it to use a range-based loop!

It should be as simple as writing (not tested):
```cpp
auto beatClip = std::find_if(m_notes.begin(), m_notes.end(), [](auto note) { return note->type() != Note::StepNote }) == m_notes.end();
```

So, if a ``Note`` is found that is not a ``StepNote``, then the iterator returned would not be equal to ``m_notes.end()`` and thus the whole expression would return false. Otherwise, if every note is a ``StepNote``, then the iterator returned is ``m_notes.end()``, and the expression would return true.

src/core/Note.cpp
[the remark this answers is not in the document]
sakertooth: > I ended up keeping this change out, since the PR doesn't involve the detuning member. The line is only changed because I needed to add a comma, maybe since it's a member from another feature the change should be done separately, just to make sure the it doesn't affect other code?

Oh, I thought you had initialized ``m_detuning`` to be ``nullptr`` in the member wise initialization list. In that case, I would say just fix the formatting on that line. This doesn't need to be an in-class initializer at least for this PR, even though it can be.

src/core/Note.cpp
```diff
@@ -74,7 +75,8 @@ Note::Note( const Note & note ) :
 	m_panning( note.m_panning ),
 	m_length( note.m_length ),
 	m_pos( note.m_pos ),
-	m_detuning( nullptr )
+	m_detuning( nullptr ),
+	m_type( note.m_type )
```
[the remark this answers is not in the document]
sakertooth: Ignore the change relating to ``m_detuning``. I thought you initialized it to ``nullptr``, lol.

src/gui/clips/MidiClipView.cpp
```diff
@@ -458,9 +458,74 @@ void MidiClipView::paintEvent( QPaintEvent * )
...
+		stepoff = s_stepBtnOff->scaled( w / steps,
+						s_stepBtnOff->height(),
+						Qt::IgnoreAspectRatio,
+						Qt::SmoothTransformation );
+		stepoffl = s_stepBtnOffLight->scaled( w / steps,
+						s_stepBtnOffLight->height(),
+						Qt::IgnoreAspectRatio,
+						Qt::SmoothTransformation );
+
+		for( int it = 0; it < steps; it++ )	// go through all the steps in the beat clip
```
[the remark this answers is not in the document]
IanCaio: Fixed formatting but ended up keeping the comment: It was already there since this is legacy code, it showed in the diffs because I moved places since I needed a different conditional flow. Thought the comments might help with refactoring of this code in the future.

include/Note.h
```diff
@@ -246,7 +259,10 @@ class LMMS_EXPORT Note : public SerializingObject
 	panning_t m_panning;
 	TimePos m_length;
 	TimePos m_pos;
-	DetuningHelper * m_detuning;
+	DetuningHelper * m_detuning = nullptr;
```
sakertooth, line 262: Since I remember you saying you did not intend to directly change ``m_detuning`` (which I mistakenly understood you did), you can omit the in-class initialization and undo the changes you made to this line. This is so that the PR is more focused.

include/Note.h
```diff
@@ -246,7 +259,10 @@ class LMMS_EXPORT Note : public SerializingObject
...
+
+	// The type of this note
+	Type m_type = Type::RegularNote;
```
sakertooth, line 265: This comment is fairly unnecessary as it is mostly self explanatory. I would suggest removing it and the spaces between ``m_type`` and the preceding ``m_detuning`` variable. For reference, [NL.1](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#nl1-dont-say-in-comments-what-can-be-clearly-stated-in-code) from the guidelines.

src/core/Note.cpp
```diff
@@ -74,7 +74,8 @@ Note::Note( const Note & note ) :
 	m_panning( note.m_panning ),
 	m_length( note.m_length ),
 	m_pos( note.m_pos ),
-	m_detuning( nullptr )
+	m_detuning(nullptr),
```
sakertooth, line 77: The formatting change of this line can be undone and the memberwise initialization of the following ``m_type(note.m_type)`` can stay.

IanCaio: I kept the formatting since I'd have to change this line anyways on the PR (by adding the comma)

sakertooth (approved):
Can't test the PR at the moment, but in terms of the code, LGTM.

src/tracks/MidiClip.cpp
```diff
@@ -281,8 +281,19 @@ void MidiClip::clearNotes()
...
+		addNote(
+			Note(TimePos(DefaultTicksPerBar / 16),
+			TimePos::stepPosition(step)), false
+		);
+	n->setType(Note::Type::StepNote);
+	// addNote will change a BeatClip to a MelodyClip
+	// because it calls checkType after adding a regular note.
+	// We need to revert it back to a BeatClip if it was one
+	// before we added the note.
+	if (prevPatternType == BeatClip) { setType(BeatClip); }
```
DomClark, line 295: This feels a bit dirty. Could you set the note type prior to adding it instead?

PhysSong:
> > Note: MIDI export may require some additional changes.
> 
> I wrote this warning because it depends on the negative length of BB notes.

Reminder.

IanCaio:
> > > Note: MIDI export may require some additional changes.
> > 
> > 
> > I wrote this warning because it depends on the negative length of BB notes.
> 
> Reminder.

Ops, thanks for the reminder!

I took a quick look at MidiExport.cpp, it seems like what I'd have to do is:

1. Add a member on the MidiNote struct that holds the note type (on MidiExport.h)
2. Change MidiExport::writeMidiClip so it also writes the note type to the MidiNoteVector
3. Change MidiExport::processPatternNotes to use the note type on the conditional

From:
```cpp
		if (it->duration < 0)
		{
			it->duration = qMin(qMin(-it->duration, next - cur), cutPos - it->time);
		}
```
To:
```cpp
		if (it->noteType == Note::Type::Step)
		{
			it->duration = qMin(qMin(it->duration, next - cur), cutPos - it->time);
		}
```

Am I missing something?

IanCaio:
*MidiExport* upgraded to use Note Types

IanCaio:
Fixed conflict. @DomClark @PhysSong @Veratil You have started reviewing this one in the past, does it look good now? Sakertooth approved so there's just one approval pending.

include/DataFile.h
```diff
@@ -128,6 +128,7 @@ class LMMS_EXPORT DataFile : public QDomDocument
 	void upgrade_bbTcoRename();
 	void upgrade_sampleAndHold();
 	void upgrade_midiCCIndexing();
+  void upgrade_noteTypes();
```
PhysSong, line 131: Fix indentation using tabs

include/Note.h
```diff
@@ -107,6 +107,19 @@ class LMMS_EXPORT Note : public SerializingObject
...
+	{
+		Regular = 0,
+		Step
+	};
+
+	Type type() const
+	{
+		return m_type;
+	}
+	inline void setType(Type t) { m_type = t; }
```
PhysSong, line 121: Why don't you make both multi-line or single line for consistency?

src/core/TimePos.cpp
```diff
@@ -161,11 +161,7 @@ tick_t TimePos::getTickWithinBeat( const TimeSig &sig ) const
 
 f_cnt_t TimePos::frames( const float framesPerTick ) const
 {
-	if( m_ticks >= 0 )
-	{
-		return static_cast<f_cnt_t>( m_ticks * framesPerTick );
-	}
-	return 0;
+	return static_cast<f_cnt_t>(m_ticks * framesPerTick);
```
PhysSong, line 164: Have you checked for any potential issues with the removal of `m_ticks >= 0` check? I agree the check is not always desired, though.

src/gui/clips/MidiClipView.cpp
```diff
@@ -458,9 +459,74 @@ void MidiClipView::paintEvent( QPaintEvent * )
...
+						Qt::IgnoreAspectRatio,
+						Qt::SmoothTransformation);
+		stepoffl = s_stepBtnOffLight->scaled(w / steps,
+						s_stepBtnOffLight->height(),
+						Qt::IgnoreAspectRatio,
+						Qt::SmoothTransformation);
+
+		for (int it = 0; it < steps; it++)	// go through all the steps in the beat clip
+		{
+			Note * n = m_clip->noteAtStep(it);
```
messmerd, line 495:
```suggestion
			Note* n = m_clip->noteAtStep(it);
```
style

messmerd (commented):
Just a couple minor things

include/DataFile.h
IanCaio: My bad, I used Github editor for this fix and I think it converts tabs to spaces. I even copied the indenting from the line above to make sure but it still left spaces :sweat_smile:

src/core/TimePos.cpp
IanCaio: I had an assert there before, it wasn't triggered back then in any of the tests but I can add it again. Even if this PR doesn't leave any `m_ticks < 0` condition, who knows, another PR might introduce it again by accident, so I added it again!

messmerd:
@IanCaio It looks like `<cassert>` needs to be included in TimePos.cpp to fix the build.

IanCaio:
> @IanCaio It looks like `<cassert>` needs to be included in TimePos.cpp to fix the build.

Weird, for some reason it compiled fine here, but I just pushed a new commit with the header. Let's see if the builds work now!

IanCaio:
@PhysSong @messmerd Does the PR looks good to you after the requested changes?

@DomClark I noticed you self-requested a review, in case there's a second approval would you like for us to wait some time so you can do it?

plugins/MidiExport/MidiExport.cpp
```diff
@@ -113,6 +114,7 @@ bool MidiExport::tryExport(const TrackContainer::TrackList &tracks,
 			int base_pitch = 0;
 			double base_volume = 1.0;
 			int base_time = 0;
+			int base_beatLen = 16;
```
messmerd, line 117: Looks like there are a few places where the default step note length of 16 is hard-coded in. You should probably create a constexpr constant for it and use that instead.

messmerd (commented):
Here's a few things I must have missed last time.

I'll test this PR soon and if all goes well, I think it should be about ready to merge

plugins/MidiExport/MidiExport.cpp
IanCaio: I just followed the example of the other variables, that value is more like a default in case there's no stored value on the XML. Should I change it to be `constexpr int base_beatLen = 16;` here and in the other method?

messmerd: No, I meant if you want the default step note length to be a 1/16th note, you shouldn't have "16" hard-coded in 3 or 4 different places. You should probably use a static constexpr member in the `Note` class called `DefaultStepNoteLength` or something like that:

```c++
// Inside Note class in Note.h
static constexpr int DefaultStepNoteLength = 16; // 16th note
```

Then replace every "16" which you're using to represent the default length of step notes with `Note::DefaultStepNoteLength`.

The point of this is to avoid magic numbers and have the default step note length defined in a single place so it's easy to change later if we ever decide to do that.

IanCaio: Oh, got it! I wrote the changes here, I'm just double checking because I think `base_beatLen` is actually not related to the default step note length of 1/16th (the one that shows on the piano roll):

> Updates MidiExport to use Note Types
> 
> 	- Now MidiExport is updated to use note types instead of relying
> on negative length notes.
> 	- For that change it was necessary to find a way of letting
> MidiExport know how long step notes should be. **The solution found was to
> add an attribute to the Instrument XML called "beatlen", which would
> hold the number of frames of the instrument's beat. That would be
> converted to ticks, so we could calculate how long the MIDI notes would
> have to be to play the whole step note. If the attribute was not found,
> the default value of 16 ticks would be used as a length of step notes,
> as a fallback.**

I think I'd need another constexpr for the beatLen fallback value. Should I have it on "InstrumentTrack.h"?

IanCaio:
I reverted one of the changes in MidiExport. Turns out `beatLen` is a per note attribute, not per instrument (it's just that some instruments will use the same beatLen for every note, but for AFP for example calling `::beatLen()` with a nullptr will result in a SegFault). So instead we would need to store the attribute on every note on the XML, but it also requires a NotePlayHandle to be calculated. Since I didn't figure out a practical and not very complex way of storing the beat length on a member variable of the Note class, it wouldn't be easily accessible on the MidiExport processing.

So now a magic number is being used as a default beat length, that in practice just represents the upper limit of the length of a step note on the MIDI in ticks. I left a TODO comment so this is clear, and in no way this worsens the MidiExport logic, since the previous code didn't care about the beat length at all. It's probably even improved because before the step notes were cut short because before the negative length of step notes was used as a limit instead.

That should be the last change hopefully. @DomClark do you still wish to review this? And the other reviewers can check whether everything they suggested was covered?

IanCaio:
Does anyone still wishes to review this PR?

zonkmachine (approved):
Tested and my projects work as they should. Two issues are solved by this and I can confirm the issues and the fixes.

> There's a qWarning left on TimePos::frames that is supposed to be removed. Keeping it just to check if there's any code left creating negative length notes. Didn't rush to remove it since I'm not even sure this will be reviewed before the refactor.

This is in the top post. It looks like this has been fixed, right?

zonkmachine:
@IanCaio I introduced a new merge conflict in the latest merge. Just a final 'touch up' and then merge, I say.

include/Note.h
```diff
@@ -107,6 +107,16 @@ class LMMS_EXPORT Note : public SerializingObject
 	Note( const Note & note );
 	~Note() override;
 
+	// Note types
+	enum class Type
+	{
+		Regular = 0,
+		Step
+	};
```
messmerd, line 115:
```suggestion
	// DO NOT modify existing values - would break project file compatibility
	enum class Type
	{
		Regular = 0,
		Step    = 1
	};
```
You could add a warning to future devs, just in case

DomClark: I don't think this needs a special comment - the same goes for any enum whose numerical value is saved to the project file, of which there are quite a few. (I'd like to switch to string values for this reason, but that's a discussion for another time.)

src/core/TimePos.cpp
```diff
@@ -23,6 +23,7 @@
  *
  */
 
+#include <cassert>
```
DomClark, line 26: This should be included after `TimePos.h`. (The first inclusion in any source file should be the corresponding header file.)

IanCaio:
Just fixed the conflicts and the order of included files

DomClark (approved):

Issue #5571, opened 2020-07-09
Display hover-sensitive info in top right corner of toolbar


Spekular:
I think this would be a nice addition, but I want to gauge interest in/resistance to the idea* before investigating more, since it'd probably be a fairly big undertaking. The concept, as seen from the user's view, is:

- No more "what's this"/explanatory/documentation-esqe tooltips show up
- All info along those lines shows up the dedicated hover-text area
- More hover text can be added now that it has it's own space
- Tooltips are still fine for readouts, like showing dB in the mixer or a control's value

Benefits of a dedicated region over tooltips:

- The top right of the toolbar can display more text than a tooltip without obstructing the user's view
- A dedicated hover-info area can update instantly on hover, whereas a tooltip needs to be delayed
- The top right area of the header is currently unused, and recent mockups in #5261 leave it that way
  - Top right is far away from most of the interface, so it's reasonable to avoid placing interactive components there. However, looking is faster than moving the mouse, so it wouldn't be an issue to put something read-only there
- FL Studio and Ableton both do it, and based on my experience with that I think it works well

The concept, from a dev perspective:

- Move the actual text so it's not in a .cpp/.h file, so there's no resistance to adding/lengthening tips because it takes up space in between actual logic.
  - A subfolder inside LMMS's install directory should contain files with the contents of each hint. For example, the `Hover Texts` directory might contain `Xpressive.htx`, which contains a hint labeled `PN Knob`.
- Components reference hints by filename and specifier
  - Straightforward solution: components manually call a `setHoverText` function when hovered.
    - Fairly easy to imagine how this is implemented, some helper class defines the function and updates the hover text area when called
    - Might need to manually call a `clearHoverText` so the text goes away when exiting the component's area
    - Might need way more event handling in a lot of components that didn't need it before
    - Less work for the helper class leaves more work (and potential mistakes to make) for each added hint
  - Better (?) solution: components just provide an ID. Another class observes which ID is currently hovered and sets the appropriate text
    - Probably a oneliner for each added hint
    - Conflicts are resolved in the observing class
    - Can't forget to call `clearHoverText` or anything like that

So, as mentioned at the start, are there any objections to something like this? From a UI/UX perspective? What about from a back end perspective? Does it sound like a good/bad idea?

DomClark:
It's important to make sure the hover text is translatable, and I'm not sure how easily we can do that with external hover text files, especially in conjunction with standard Qt translation.

An alternative approach on how to display these:
[...]

PR #5911, opened 2021-02-17
Knob value hover

salilra:


PhysSong:
I can see the "stuck" text float even after the mouse pointer moves out of the knob. Also, [indentations should be done with tabs](https://github.com/LMMS/lmms/wiki/Coding-Conventions#indentation), not spaces.

superpaik:
In addition to what PhysSong said, I just notice that the text appears almost instantly. It's common UX practice to wait some time (about a second) before the text popup to avoid the text to constantly appear when the user moves the mouse.

Spekular:
> In addition to what PhysSong said, I just notice that the text appears almost instantly. It's common UX practice to wait some time (about a second) before the text popup to avoid the text to constantly appear when the user moves the mouse.

A second is rather long. Off the top of my head I recall 300ms being "standard", but searching again I find a range of 300-500ms. Google claims the average human reaction time (for visual stimuli) is 250ms, which lines up nicely with the 300ms delay.

If we had used a QToolTip for this, there's an [existing method](https://doc.qt.io/qt-5/qtooltip.html#showText-2) to show a tooltip with delay. [TextFloat](https://github.com/LMMS/lmms/blob/89b13280de93ad2131c12dc234fa65cd10855996/include/TextFloat.h) doesn't have such a method at the moment though, so it would need to be improved if we want to use it for this. (I don't think it's wise to implement the delay logic inside Knob).

In regards to the stuck tooltip, I think you should use the [enterEvent](https://doc.qt.io/qt-5/qwidget.html#enterEvent) and [leaveEvent](https://doc.qt.io/qt-5/qwidget.html#leaveEvent) rather than mouseMoveEvent to trigger the tooltip.

michaelgregorius:
The cleanest way would be if Qt provided some mechanism to be notified when a (custom) tool tip should be shown or hidden. That way the custom tool tip would be consistent with the other ones with regards to the time that it takes until it is triggered.

The documentation of [QToolTip](https://doc.qt.io/qt-6/qtooltip.html#details) has some details on how to intercept a `QHelpEvent` that indicates that a tool tip would be about to show. Unfortunately, this only works half the way, i.e. there does not seem to be a way to get informed that a tool tip should be hidden again.

An alternative option might be to have a fixed place in the GUI where detail information can be shown, e.g. when the mouse pointer is above certain elements/widgets. So as soon as you'd move the mouse over a knob it would show some information about the current value in that area.

michaelgregorius:
> > An alternative option might be to have a fixed place in the GUI where detail information can be shown, e.g. when the mouse pointer is above certain elements/widgets. So as soon as you'd move the mouse over a knob it would show some information about the current value in that area.
> 
> #5571

Ah, there's always already an issue. 😅

zonkmachine:
> Ah, there's always already an issue. 😅

And now there's a PR. I bumped you to review it. My coding is weak. https://github.com/LMMS/lmms/pull/6835

zonkmachine:
Closing as superseded by https://github.com/LMMS/lmms/pull/6835
@salilra Thanks for your efforts!

michaelgregorius:
> > Ah, there's always already an issue. 😅
> 
> And now there's a PR. I bumped you to review it. My coding is weak. #6835

Done, @zonkmachine!

Issue #5885, opened 2021-01-18
Undo in Song Editor Fails After Duplicating Multiple Midi Reigions


tinkerlevu:
### Bug Summary
LMMS fails to to properly undo an operation where midi regions from two separate instrument tracks are duplicated.
This has also randomly caused crashes or instability.

#### Steps to reproduce
1. Create two instrument tracks next to each other in the song editor.
2. Create a Midi regions in each track and populate them with notes.
3. Select the multiple Midi regions simultaneously.
4. Use Ctrl+drag to duplicate the selected reigions to other positions in the tracks
5. Use Ctrl + Z to attempt an Undo

#### Expected behavior
The two duplicates are deleted.

#### Actual behavior
Only the duplicates in one of the instrument tracks are deleted.
A crash may occur depending on the instrument and project.

#### Screenshot
![bug1](https://user-images.githubusercontent.com/7609579/104942887-da6a1680-5a19-11eb-8eb7-c046998f8474.png)
![bug2](https://user-images.githubusercontent.com/7609579/104942886-d9d18000-5a19-11eb-8cc2-26f24a05ecf8.png)
![bug3](https://user-images.githubusercontent.com/7609579/104942884-d938e980-5a19-11eb-9e68-30c1a91379c3.png)
![bug4](https://user-images.githubusercontent.com/7609579/104942879-d807bc80-5a19-11eb-9160-345cc8d4ff95.png)

#### Affected LMMS versions
1.3.0 Alpha

#### Logs
<details>
  <summary>Click to expand</summary>
<pre>
Carla appears to be installed on this system at /usr/lib[64]/carla so we'll use it.
Jack appears to be installed on this system, so we'll use it.
Lv2 plugin SUMMARY: 95 of 192  loaded in 1550 msecs.
For details about not loaded plugins, please set
  environment variable "LMMS_LV2_DEBUG" to nonempty.
Lv2 Plugins blacklisted: 2 of 192 
  If you want to ignore the blacklist (dangerous!), please set
  environment variable "LMMS_IGNORE_BLACKLIST" to nonempty.
</pre>
</details>

PR #5920, opened 2021-02-24
Fix Undo in Song Editor Fails After Duplicating Multiple Midi Reigions

cyber-bridge:
Fixes https://github.com/LMMS/lmms/issues/5885

### DIFF
--- a/src/gui/widgets/TrackContentWidget.cpp
+++ b/src/gui/widgets/TrackContentWidget.cpp
@@ -438,8 +438,6 @@ bool TrackContentWidget::pasteSelection( TimePos tcoPos, const QMimeData * md, b
 	QString type = decodeKey( md );
 	QString value = decodeValue( md );
 
-	getTrack()->addJournalCheckPoint();
-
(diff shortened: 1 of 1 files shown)


superpaik:
It works Ok on windows.
Just one comment about something minor I just found. If it's intended, don't worry about it. If the two regions are in separate tracks, the first undo deletes one of them, the second the other as you said. But if the regions are in the same track (or they are mixed, like two in the same track and a third in another one) the undo, deletes all the regions duplicated in the same track together. If we are consistent it should delete just one region, or maybe all of them altogether.

cyber-bridge:
@superpaik thanks for testing!

Yes from my perspective it is intended for now, checkpoints in LMMS are done per Track. From what I could tell a overhaul of the way journaling works would be required if we want to undo multiple tracks at once. What could be done however is to add a check-point per TCO (region) instead of the per Track what I've done now.

cyber-bridge:
Closing in favor of https://github.com/LMMS/lmms/pull/6410

PR #5292, opened 2019-10-31
Automation Editor point fine tuning with double click

tecknixia:
Related to issue #5225 and old PR #5232 (closed)

Allows user to double click an automation point in the Automation Editor, and enter new number to adjust the y level for that point.  It also improves mouse cursor interaction with the point, and shows the point's Y level when mouse cursor is over a point.  

This PR involves the creation of `doubleClickEvent`, creation of variable `m_pointYLevel`, and the use of `mouseMoveEvent` and `drawCross`.

- A variable, `m_pointYLevel` is created and used to hold a point's Y level position.
- The value of `m_pointYLevel` is set when the mouse moves over a point.
- The value of `m_pointYLevel` is used when opening an input dialog box, for the current value.
- If the mouse moves away from a point, the value of `m_pointYLevel` returns to `0`.
- When the value of `m_pointYLevel` is `0`, the tool tip in `drawCross` displays the Y level position of the mouse.
- When the value of `m_pointYLevel` is greater than `0`, the tool tip in `drawCross` displays the point's Y level position.

PR #5712, opened 2020-10-13
Changes AutomationPattern to use nodes instead of raw float values

IanCaio:
The original goal of this PR was to introduce the use of Automation Nodes instead of Automation Values to give more flexibility in terms of adjusting the automation shapes. During the process of writing some issues were noticed in affected files (mostly on `AutomationEditor.cpp`), and since they would have to be changed to the introduction of this new feature they were fixed as those issues were found, adding up to the final size of the PR. I believe the automation editor code was improved in comparison to what we had (there were conditions that caused UB, obsolete legacy code and repeated routines for example) and refactoring it created a clear ground for introducing the automation nodes.

### Current behavior

lmms1

At the moment, each automation pattern is basically a sequence of values in a timeline. Every point on the automation editor represents the value of the automated parameter at that particular time. The progression type will tell LMMS how it should transition between each of those values.The problem with that approach is that we can't store any more information regarding that point in time, so we are very limited in terms of adjustments that can be done to the shape of the automation pattern.

### New behavior

lmms2

With the changes proposed by this PR, instead of values in a timeline we have Automation Nodes. Nodes can store basically any information we want it to, not just a value. Right now, it stores an input value, an output value, the input and output tangents for curve progressions, the position in the timeline and the pattern that owns it.

The input value (`inValue`) represents the value of the node for the progression on the left side of it. It can be considered the real value of that node, since it represents the value that the parameter should have exactly at that time. It's represented by a blue circle.

The output value (`outValue`) represents the value of the node immediately after the time it's situated in. That value will be used for the progression to the right side of the node. It allows the user to create a discrete jump on the exact position of that node. By default, the output value matches the input value.

The input tangent of the curve represents the tangent of the curve (for Cubic Hermite progressions) to the left of the node and the output tangent the same, but for the right side of the curve. They are currently automatically calculated using the same logic as before, but having them stored at the automation node opens the possibility of we allowing them to be manually changed in the future.

Another idea that was discussed (and was the original motivation for this transition) is having the nodes storing a progression type, so we can have different transition types on the same automation pattern. This PR creates the foundation for that to be implemented, but considering the current size of it already I believe that should be left for another PR, after this one is merged.

## Automation Editor controls

Currently, the automation editor have 2 edit modes: Draw mode and Erase mode. With the changes on this PR, a third mode is added ("Draw outValue mode"). The icon representing it is to be changed later, right now we just reuse the one from Draw mode. The controls for each mode are:

### Draw mode (Shift + D):

[...]

PR #5923, opened 2021-02-28
Automation Node fine tuning with double-click (Rewrites #5292)

IanCaio:
This PR rewrites #5292 under the new framework of Automation Patterns (where Automation Nodes are used).

Double clicking the node's inValue will open a dialog to edit the inValue and double clicking the node's outValue will open a dialog to edit the outValue.


IanCaio:
The massive number of commits is due to the fact I had started worked on this feature before #5712 was merged, so Github is thinking the earlier commits are from this PR when in fact they are from the original Automation Nodes PR. The only relevant commits are from 2226108 forward. The diff is small enough that it shouldn't matter though.

IanCaio:
Fixed the mentioned bug, should be good for testing and reviewing.

Spekular:
Why not open a dialogue that exposes both the in and out values (potentially highlighting the one that was clicked)? This would be convenient if one wants to adjust both, or struggles to click the correct one (in and out values are the same or very close).

IanCaio:
> Why not open a dialogue that exposes both the in and out values (potentially highlighting the one that was clicked)? This would be convenient if one wants to adjust both, or struggles to click the correct one (in and out values are the same or very close).

That sounds good in terms of UX, but I don't think Qt has a convenient input dialog for 2 floats, so I think a custom one would have to be made. I'm not very experienced with Qt, so if that custom dialog isn't too complex to write and someone want to give it a go it's cool.

But as for the issue you mentioned, what defines whether the inValue or outValue will be adjusted is the mode you're in, so it isn't a problem if they are in the same level as the Draw mode will always edit the inValue and the Draw OutValues mode will always edit the outValue. One thing that I probably should change is that if the `inValue` is equal to the `outValue`, setting the former should adjust the latter as well (kind of like how dragging works now).

IanCaio:
Now fine tuning the `inValue` of a node that has `inValue` equal to the `outValue` will change both, the same way as dragging works.

superpaik:
It works ok on windows.
Some thoughts. 
-In Erase Mode it's only possible to delete "normal" automation points, not outValues. Maybe it'd be ok to erase also outValues, I don't see the point of adding a new tool option to do that. And thinking about that, why do not have just one tool for creating automation points? Normal points are like they are now, outValues are created with a modifier key (like "Ctrl" or "Shift" and mouse action). Also right click would delete the point you are over, regardless of the selected tool.
-As for the project xml I notice a strange value "<automationtrack/>" inside the automation track. For sure it's not related to this PR, but just in case.
Capture

IanCaio:
Thanks for testing man!

In Erase mode you should be able to reset outValues right clicking, while left clicking removes the node completely. You can also reset the outValues on the Draw OutValue mode with the right mouse button as well.

The Draw OutValue mode and mouse behavior was set on #5712 . Originally, the idea was to be able to edit the outValues through the Draw edit mode using `alt` as a modifier to drag them, as `ctrl` and `shift` are already used ("Keep surrounding nodes" and "Draw line" respectively). The problem is that in some OSs it's problematic to use `alt` as a modifier key as @Spekular pointed out [here](https://github.com/LMMS/lmms/pull/5712#issuecomment-710383396), so among the available alternatives I thought the extra mode was the better one. It's possible that we review the controls and change them in the future, but for now it would be complicated because it involves changing the other shortcuts as well.

As for the removing of nodes and reseting of outValues with the same mouse click, it could be done. But I think the workflow with LMB to remove nodes and RMB to reset outValues is faster because you don't have to worry about clicking on the top of the right sphere, and you can also quickly click and drag to remove/reset multiple nodes. I rarely use the Erase mode tbh though.

I didn't notice which strange value you mean on the XML. The nodes are stored as `<time>`, and the inValue is stored on `value` (for backwards compatibility I used `value` instead of `inValue`) and the outValue is stored on `outValue`. The time position is stored under `pos`. So `<time outValue="115.3" value="152.6" pos="120">` means that you have a node 120 ticks from the beginning of the pattern, with the inValue equal to 152.6 and the outValue equal to 115.3.

superpaik:
Oh, sorry. I didn't get the right-click to delete outValues while on Erase Mode. That's perfect! Can't we have the same behaviour for creating nodes? I mean left click to create normal nodes, right-click for outValue Nodes? I don't know if that it's problematic with OSs other than Windows, but it's kind of mimicking the current Erase behaviour.

As for the XML file, sorry there's a misspelling in my comment. The thing it's related to what it's highlighted on the image, the "<automationtrack/>" element without content. As I said, I doubt it's related to this PR and maybe it's there for future implementations. The information regarding points/time it is correct.

superpaik:
Sorry again. Don't worry about the XML file. It seems that this element is on version 1.2 as well. Maybe it is to indicate that the track is an automation track, but I believe that "type=5" is for automation tracks.
Note: again the '<automationtrack/>' is been deleted by github when saving ¿?

IanCaio:
> Oh, sorry. I didn't get the right-click to delete outValues while on Erase Mode. That's perfect! Can't we have the same behaviour for creating nodes? I mean left click to create normal nodes, right-click for outValue Nodes? I don't know if that it's problematic with OSs other than Windows, but it's kind of mimicking the current Erase behaviour.

That would work, would just need to have a more broad opinion survey, as I think many people might be used to the convenience of removing with the right click (I think I only remove nodes that way haha). But if it's compensated by a faster workflow with outValues it could be changed with no problems!

> Sorry again. Don't worry about the XML file. It seems that this element is on version 1.2 as well. Maybe it is to indicate that the track is an automation track, but I believe that "type=5" is for automation tracks.
Note: again the xml "automationtrack" tag has been deleted by github when saving ¿?

No worries, I realized what you were referring to now. It's weird indeed there's a tag with no information or children. The track type is marked on the attributes of `<track>`, maybe it's just some visual aid so people reading the XML can quickly see it's an Automation Track, without having to know the meaning of type = 5? Or maybe there's some information that could go there but it just happens your project didn't have it? Idk to be honest

src/gui/editors/AutomationEditor.cpp
```diff
@@ -615,6 +615,91 @@ void AutomationEditor::mousePressEvent( QMouseEvent* mouseEvent )
 
 
 
+void AutomationEditor::mouseDoubleClickEvent(QMouseEvent * mouseEvent)
+{
+	if (!validPattern()) { return; }
+
+	// Helper lambda function to fine tune a node's in/out value
+	auto fineTuneValue = [this](timeMap::iterator node, bool editingInValue)
```
Spekular, line 623: Can you extract this to a regular function? It could make sense to call outside of double click events (like via a context menu, keyboard shortcut, or new edit mode).

src/gui/editors/AutomationEditor.cpp
```diff
@@ -615,6 +615,91 @@ void AutomationEditor::mousePressEvent( QMouseEvent* mouseEvent )
...
+				return true;
+			}
+		}
+
+		// If nothing was changed (invalid node, dialog closed, etc) return false
+		return false;
+	};
+
+	// Do we need to call update()?
+	bool needUpdate = false;
```
Spekular, line 673: Any reason you don't just call ``update()`` in the places where you set ``needUpdate = true``? Otherwise, you could at least move this to be inside the if statement below since it's only used there.

Spekular (changes requested):

src/gui/editors/AutomationEditor.cpp
IanCaio: Sure! I'll move it to `AutomationEditor::fineTuneValue` then

src/gui/editors/AutomationEditor.cpp
IanCaio: To be honest I don't remember, maybe back then I thought it was more readable or something. You prefer to replace it with a straight call to `update` or moving it inside the `if` block?

Spekular: The way the method looks now I prefer a straight call to update.

IanCaio:
Moved `fineTuneValue` to a separate method and removed that `needsUpdate` variable.

Spekular (approved):
LGTM

src/gui/editors/AutomationEditor.cpp
```diff
@@ -398,6 +398,58 @@ void AutomationEditor::drawLine( int x0In, float y0, int x1In, float y1 )
 
 
 
+bool AutomationEditor::fineTuneValue(timeMap::iterator node, bool editingInValue)
+{
+	if (node != m_pattern->getTimeMap().end())
```
Veratil, line 403: Style suggestion:

We could save indention by using the inverse here:
```suggestion
	if (node == m_pattern->getTimeMap().end()) { return false; }
```

Then the rest of the block is the normal function indented one less.

src/gui/editors/AutomationEditor.cpp
```diff
@@ -615,6 +667,36 @@ void AutomationEditor::mousePressEvent( QMouseEvent* mouseEvent )
 
 
 
+void AutomationEditor::mouseDoubleClickEvent(QMouseEvent * mouseEvent)
+{
+	if (!validPattern()) { return; }
+
+	// If we double clicked inside the AutomationEditor viewport
+	if (mouseEvent->y() > TOP_MARGIN && mouseEvent->x() >= VALUES_WIDTH)
```
Veratil, line 675: Style suggestion:

Same inverse guard as above.
```suggestion
	if (mouseEvent->y() <= TOP_MARGIN || mouseEvent->x() < VALUES_WIDTH) { return; }
```
Then indent the function left once.

src/gui/editors/AutomationEditor.cpp
```diff
@@ -615,6 +667,36 @@ void AutomationEditor::mousePressEvent( QMouseEvent* mouseEvent )
...
+	{
+		timeMap::iterator clickedNode;
```
Veratil, line 677: Can clear two assignments by doing this (and reusing the bool later if change `fineTuneValue` argument to `outValue` instead).
```suggestion
		const bool isOutVal = m_editMode == DRAW_OUTVALUES;
		timeMap::iterator clickedNode = getNodeAt(mouseEvent->x(), mouseEvent->y(), isOutVal);
```

src/gui/editors/AutomationEditor.cpp
```diff
@@ -398,6 +398,58 @@ void AutomationEditor::drawLine( int x0In, float y0, int x1In, float y1 )
 
 
 
+bool AutomationEditor::fineTuneValue(timeMap::iterator node, bool editingInValue)
```
Veratil, line 401: If we change `editingInValue` to `editingOutValue` then we can simplify the switch statement in `mouseDoubleClickEvent`.

src/gui/editors/AutomationEditor.cpp
```diff
@@ -615,6 +667,36 @@ void AutomationEditor::mousePressEvent( QMouseEvent* mouseEvent )
...
+			case DRAW:
+				// We will edit the inValue
+				clickedNode = getNodeAt(mouseEvent->x(), mouseEvent->y());
+				if (fineTuneValue(clickedNode, true)) { update(); }
+				break;
+			case DRAW_OUTVALUES:
+				// We will edit the outValue
+				clickedNode = getNodeAt(mouseEvent->x(), mouseEvent->y(), true);
+				if (fineTuneValue(clickedNode, false)) { update(); }
+				break;
```
Veratil, line 690: By using the suggestions for `fineTuneValue` and using the `bool` above we can simplify this switch to:
```suggestion
			case DRAW:
			case DRAW_OUTVALUES:
				if (fineTuneValue(clickedNode, isOutVal)) { update(); }
				break;
```

src/gui/editors/AutomationEditor.cpp
```diff
@@ -398,6 +398,58 @@ void AutomationEditor::drawLine( int x0In, float y0, int x1In, float y1 )
...
+			if (editingInValue)
+			{
+				// If the outValue is equal to the inValue we
+				// set both to the given value
+				if (OFFSET(node) == 0)
+				{
+					node.value().setOutValue(value);
+				}
+				node.value().setInValue(value);
+				Engine::getSong()->setModified();
```
Veratil, line 434: We `setModified` in both branches of the `if`, we can move this line to after the if/else.

IanCaio:
Done! Did a quick test, everything looks fine

Veratil:
> 
> 
> Done! Did a quick test, everything looks fine

If all is working the same as before, LGTM. 👍

Veratil (approved):

PR #5924, opened 2021-02-28
Adds feature to edit tangents of Cubic Hermite progressions

IanCaio:
This PR enables the editing of a node's tangents on the Cubic Hermite progression. It works the following way:

- A node can have its tangents locked or unlocked. A node with unlocked tangents will have its own tangents recalculated when new nodes are added or nodes are moved around
- On the Edit Tangents mode, the user can edit the tangents of a particular node. To do so, just click on the pattern and the tangents of the closest node will start being edited.
- When a node has had it's tangents manually edited, they will become locked. That means they won't be recalculated when nodes are added/removed/moved.
- Right clicking on the Edit Tangents mode resets the tangents of the closest node. If you hold the RMB and drag it will reset the tangents of all the nodes on the path.
- When a node's tangent is reset, it will become unlocked again and will be recalculated according to the neighbor nodes.
- The tangent information of a node (and whether it's locked or unlocked) is now saved on the project file. It's not necessary to generate the tangents during load up anymore. For older projects compatibility it's checked whether any node doesn't have the tangent values saved, and if so the tangents are generated during load up.
- Color of the tangent line can be edited on the CSS (the tangent circle is just the same color but darkened).
- The tangents are only drawn on the Cubic Hermite mode. They are also only drawn for the nodes that had their tangents edited. That way the user can quickly know which nodes have the tangents locked and which doesn't.

Again, sorry for the commit mess. This was being worked in my fork before #5712 getting merged and Github is making confusion on the commits that belong to this branch and the ones which belongs to the other.

AutomationTangents


superpaik:
Tested on Windows 10. Generally it works ok.
The only weird thing that I found is that when a point have tangents and you are in "draw mode" and click over that point because you wanna move it, tangents disappear. I think that happens because in reality the action deletes the current point and creates a new one. I don't know if this can be solved, but otherwise I don't know how to move a point with tangents.

Other minor issues:
- The hotkey (Shitf-T) activates the Draw Outmodes tool. It's keyword (Shift-C) does nothing.
- Since edit tangents is only possible while in "Cubic Hermite progression", wouldn't be better to only allow to active the tool if you are using that mode? I mean, if you are on "linear" or "discrete" progression you can activate the edit tangents tool but do nothing. This may confuse the user.
- Tangent reacts to vertical mouse movements (ok) but also to horizontal moves and if you are moving, say, from left to right when you reach the point the tangent goes from totally down to totally up. I don't know if it would be better to only react to vertical movements.

Positive remarks:
- I don't know if that's intended, but if you edit tangents and then change to, say, linear progression, tangents disappear (ok). But if you return to "Cubit Hermite progression" the tangents appear again (super ok). And also that happens if you save and load the project again. Tangents are saved in the project file even if they are not visible in the automation editor. To me is ok!

IanCaio:
Thanks for testing! Fixed a few of the points, I'll soon work on the remaining.

> Tested on Windows 10. Generally it works ok.
> The only weird thing that I found is that when a point have tangents and you are in "draw mode" and click over that point because you wanna move it, tangents disappear. I think that happens because in reality the action deletes the current point and creates a new one. I don't know if this can be solved, but otherwise I don't know how to move a point with tangents.

The dragging of nodes involves creating new nodes objects, but I'll try to figure out a way to keep the tangents if we are dragging and not creating a new node. I thought I had done it but apparently forgot :grimacing: 

> * The hotkey (Shitf-T) activates the Draw Outmodes tool. It's keyword (Shift-C) does nothing.

That was a bug, I did a typo writing a variable's name and was setting the Edit Tangent mode on the Draw Out mode. It should be fixed now!
 
> * Since edit tangents is only possible while in "Cubic Hermite progression", wouldn't be better to only allow to active the tool if you are using that mode? I mean, if you are on "linear" or "discrete" progression you can activate the edit tangents tool but do nothing. This may confuse the user.

Added logic for disabling the Edit Tangent mode button if you're in Discrete or Linear progressions.

> * Tangent reacts to vertical mouse movements (ok) but also to horizontal moves and if you are moving, say, from left to right when you reach the point the tangent goes from totally down to totally up. I don't know if it would be better to only react to vertical movements.

The tangents are calculated by the tangent between the mouse position and the node position. When you go from one side to the other it ends up flipping because the X delta shifts sign. If you stop to think of it as a straight line that continues from one side to the other, the tangent is still accurate (the mouse sits on top of this imaginary line), however I can see how this can be confusing for the user. I'll try to add some logic later to account for switching sides on a node while editing the tangent so it mirrors it to compensate.

IanCaio:
@superpaik 
The last 2 points were addressed. Dragging a node now keeps its tangents if they were locked (aka they are visible, meaning they were edited).

I also slightly changed the tangent calculation behavior, but I don't know if it improved over the previous one (I actually prefer the previous) because crossing the mouse through the node still causes changes, it just happens that the tangent is still pointing in the direction of the mouse but mirrored. I can revert it to the previous behavior. It was good that I had to tweak with it anyways because I fixed a bug with a possible division by 0 on the process.

superpaik:
Everything tested before works ok now.

I just notice that undo-redo do not work as expected. When undo all tangent points disappear with just one undo (and appear with redo) That do not happens with normal points. I know undo-redo needs a general refactoring but maybe you can double-check the code to see is something is amiss.

As for the new behaviour when moving the mouse, to me is more appealing , but I agree that is a matter of preference. Maybe some other can say what's best.

Good work!

IanCaio:
Thanks man!
 
> I just notice that undo-redo do not work as expected. When undo all tangent points disappear with just one undo (and appear with redo) That do not happens with normal points. I know undo-redo needs a general refactoring but maybe you can double-check the code to see is something is amiss.

My bad, I forgot to add a Journalling checkpoint so it was using whatever checkpoint was available before (which was before you added all the tangents, so they all disappeared). I just added the missing code, let me know if it works fine now!

> As for the new behaviour when moving the mouse, to me is more appealing , but I agree that is a matter of preference. Maybe some other can say what's best.

Cool! I'm fine with either one to be honest, so if this new behavior feels better for you I'll keep it!

superpaik:
It works perfectly, now!

IanCaio:
Conflict introduced by #5968 was resolved

src/gui/editors/AutomationEditor.cpp
```diff
@@ -654,6 +671,52 @@ void AutomationEditor::mousePressEvent( QMouseEvent* mouseEvent )
 				}
 			break;
 			}
+			case EDIT_TANGENTS:
+			{
+				if (m_clip->progressionType() == AutomationClip::CubicHermiteProgression)
```
PhysSong, line 676: I think it should be better to have a function in `AutomationClip` which returns if the progression type supports manual tangent edit, instead of hard-coding `AutomationClip::CubicHermiteProgression`.

luzpaz:
Anyone testing this ?

zonkmachine:
> Anyone testing this ?

Just go ahead!

zonkmachine:
@IanCaio Needs update once more!

Testing now and it works great :+1:

One issue:

* After reopening a project with edited nodes I've managed to get into a state where the node edit button becomes unresponsive. I'm still trouble shooting this to try and find a clear case when this happens but basically, edit a node, save, reload, try and edit again, should show this bug.

Some quick observations:

* I think the tangent editing feels a bit 'nervous'. The mouse action results in big changes and I've even tried to edit my mouse settings to make more minute changes possible. Maybe an accelerator button that increases the resolution of the mouse actions would be motivated?
* After editing the tangent on one side, I often find myself moving the other one to the opposite position so they are aligned. Maybe an option to move the other node automagically?

IanCaio:
> @IanCaio Needs update once more!
> 
> Testing now and it works great 👍
> 
> One issue:
> 
>    * After reopening a project with edited nodes I've managed to get into a state where the node edit button becomes unresponsive.
>      Method to replicate issue:
>      Create a project with an automation track and edit tangents in it. Save and reload lmms. Open the projet and press the 'Edit tangents' button. The button is now unresponsive. It will come to life again by selecting one of the other progression types and then switching back to Cubic Hermite.
>      I tried [3016a29](https://github.com/LMMS/lmms/commit/3016a295684c43fa1d86cac841a4c6af7fbafe91) which is the last change that is not 'merge master' and this behavior is already in that version.

Thanks man! I'd have to get a better look, but that's probably due to the button being enabled/disabled on `setProgressionType` and me forgetting to account for when a clip is loaded either by drag&drop or by project-loading, where the progression type is changed manually. If you drop an automation clip with a different progression type I imagine a similar bug would happen. That's just a guess looking quickly at the code, would have to confirm. There are a couple ways to fix it if that's the case, either having a *progressionTypeChange* signal that triggers the button change, or if I add the conditional inside the methods that will set a different clip.

> 
> 
> Some quick observations:
> 
>    * I think the tangent editing feels a bit 'nervous'. The mouse action results in big changes and I've even tried to edit my mouse settings to make more minute changes possible. Maybe an accelerator key that increases the resolution of the mouse actions would be motivated?
> 
>    * After editing the tangent on one side, I often find myself moving the other one to the opposite position so they are aligned. Maybe an option to move the other node automagically?

Those are relatively simple improvements I think, could be done. IMO would be better in another PR, since the more I add to this one the less likely I believe this will actually get reviewed and merged.

If I'm being transparent though, I got a bit demotivated with my PRs. I was not coding for a long time, got pinged and came back to fix issues on them, get them ready for merge and review, but they simply got left aside. All of them are over 2 years old and after I got back I saw lots of PRs being reviewed and merged (all way more recent than mine), more conflicts coming up and me having to keep fixing them while I was getting no reviewers to get them out of the way. Don't take me wrong, you're one of the few that showed interest in the PRs which I appreciate, also others that helped keeping them up to date after the major refactor such as @PhysSong , but it got a bit tiring to just be working on them while being left down in the queue. I'll still see if I find some time to take a look at this bug though.

zonkmachine:
> Those are relatively simple improvements I think, could be done. IMO would be better in another PR, since the more I add to this one the less likely I believe this will actually get reviewed and merged.

I agree.

> I saw lots of PRs being reviewed and merged (all way more recent than mine),

Yeah, I know how that feels. There's unfortunately no easy way to deal with this. I was more or less away from the project while the code were being rearranged and we now have an insane backlog of 150 PR's. I just grabbed the ones that were the easiest to get started with to get them out of the way but there aren't that many PR's I can review as my coding skills are somewhat limited. I'll keep testing the rest of your stuff though. It looks pretty solid all of it. Next up is the ~~BBTrack~~ Pattern Editor features. Update this one when you feel up to it.

IanCaio:
@zonkmachine I think I managed to fix the conflict, now I have to think about how to approach the bugfix. Just another way to reproduce it more easily: you can have two automation clips, have one with linear progression and the other with cubic hermite progression. Then start editing the tangents of the latter, and double click the linear progression clip. You'll have the editor on the Tangent Edit mode with the linear progression automation clip (it won't work but it will be selected). The opposite happens too, if you are in the linear progression clip with the Tangent Edit button disabled, then double click the cubic hermite automation clip it will switch to that but with the button still disabled, you have to change the progression type for it to enable again.

I'll see if I can work on that on the weekend and also check the review that was left on the BBTrack PR!

src/core/AutomationClip.cpp
```diff
@@ -422,6 +422,37 @@ void AutomationClip::resetNodes(const int tick0, const int tick1)
...
+		if (it != m_timeMap.end())
+		{
+			it.value().setLockedTangents(false);
+			generateTangents(it, 1);
+		}
+		return;
+	}
+
+	TimePos start = TimePos(qMin(tick0, tick1));
+	TimePos end = TimePos(qMax(tick0, tick1));
```
messmerd, line 444:
```suggestion
	TimePos start = TimePos(std::min(tick0, tick1));
	TimePos end = TimePos(std::max(tick0, tick1));
```
We should keep `qMin`, `qMax`, and other such Qt functions out of the core. You might need to include `<algorithm>` if it isn't already.

src/gui/editors/AutomationEditor.cpp
```diff
@@ -180,6 +182,10 @@ AutomationEditor::AutomationEditor() :
 	{
 		s_toolDrawOut = new QPixmap(embed::getIconPixmap("edit_draw_outvalue"));
 	}
+	if (s_toolEditTangents == nullptr)
+	{
+		s_toolEditTangents = new QPixmap(embed::getIconPixmap("edit_tangent"));
```
messmerd, line 187: I think this line introduces a memory leak.

@sakertooth recently opened a PR (#6879) which will change each instance of `new QPixmap` throughout the codebase to allocate on the stack instead.

src/gui/editors/AutomationEditor.cpp
```diff
@@ -862,6 +925,56 @@ void AutomationEditor::mouseMoveEvent(QMouseEvent * mouseEvent )
...
+			case EditMode::EditTangents:
+			{
+				// If we moved the mouse past the beginning correct the position in ticks
+				posTicks = qMax(posTicks, 0);
+
+				if (m_mouseDownLeft)
+				{
+					if (m_action == Action::MoveTangent)
+					{
+						timeMap & tm = m_clip->getTimeMap();
```
messmerd, line 937:
```suggestion
						timeMap& tm = m_clip->getTimeMap();
```
style

src/gui/editors/AutomationEditor.cpp
```diff
@@ -862,6 +925,56 @@ void AutomationEditor::mouseMoveEvent(QMouseEvent * mouseEvent )
 				}
 			break;
 			}
+			case EditMode::EditTangents:
+			{
+				// If we moved the mouse past the beginning correct the position in ticks
+				posTicks = qMax(posTicks, 0);
```
messmerd, line 931:
```suggestion
				posTicks = std::max(posTicks, 0);
```
This isn't a necessary change since this is gui code, so it's up to you.

src/gui/editors/AutomationEditor.cpp
```diff
@@ -862,6 +925,56 @@ void AutomationEditor::mouseMoveEvent(QMouseEvent * mouseEvent )
...
+
+				if (m_mouseDownLeft)
+				{
+					if (m_action == Action::MoveTangent)
+					{
+						timeMap & tm = m_clip->getTimeMap();
+						timeMap::iterator it = tm.find(m_draggedTangentTick);
```
messmerd, line 938:
```suggestion
						auto it = tm.find(m_draggedTangentTick);
```
I'd modernize this with `auto`, though this is just me nitpicking

src/gui/editors/AutomationEditor.cpp
```diff
@@ -937,6 +1050,43 @@ inline void AutomationEditor::drawAutomationPoint(QPainter & p, timeMap::iterato
 
 
 
+inline void AutomationEditor::drawAutomationTangents(QPainter & p, timeMap::iterator it)
```
messmerd, line 1053:
```suggestion
inline void AutomationEditor::drawAutomationTangents(QPainter& p, timeMap::iterator it)
```
style

src/gui/editors/AutomationEditor.cpp
```diff
@@ -1822,6 +1988,63 @@ AutomationEditor::timeMap::iterator AutomationEditor::getNodeAt(int x, int y, bo
...
+ */
+AutomationEditor::timeMap::iterator AutomationEditor::getClosestNode(int x)
+{
+	// Remove the VALUES_WIDTH from the x position, so we have the actual viewport x
+	x -= VALUES_WIDTH;
+	// Convert the x position to the position in ticks
+	int posTicks = (x * TimePos::ticksPerBar() / m_ppb) + m_currentPosition;
+
+	// Get our pattern timeMap and create a iterator so we can check the nodes
+	timeMap & tm = m_clip->getTimeMap();
```
messmerd, line 2005:
```suggestion
	timeMap& tm = m_clip->getTimeMap();
```
style

src/gui/editors/AutomationEditor.cpp
```diff
@@ -1822,6 +1988,63 @@ AutomationEditor::timeMap::iterator AutomationEditor::getNodeAt(int x, int y, bo
...
+
+	if (tm.isEmpty()) { return tm.end(); }
+
+	// Get the node with an equal or higher position
+	timeMap::iterator it = tm.lowerBound(posTicks);
```
messmerd, line 2010:
```suggestion
	auto it = tm.lowerBound(posTicks);
```
This is me nitpicking again

include/AutomationEditor.h
```diff
@@ -225,6 +237,7 @@ protected slots:
 
 	void drawCross(QPainter & p );
 	void drawAutomationPoint( QPainter & p, timeMap::iterator it );
+	void drawAutomationTangents(QPainter & p, timeMap::iterator it);
```
messmerd, line 240:
```suggestion
	void drawAutomationTangents(QPainter& p, timeMap::iterator it);
```
style

messmerd (commented):
This is mostly a style review. Overall it looks very good.
Next I'll test it.

messmerd:
I spent some time testing it with an ASan build.
It seems to work fine, and I encountered no issues.

I really like the functionality this PR provides, and it is integrated well.

If I had to find something to complain about, it would be the way tangent lines are edited. It seems to work by setting the tangent's slope based on the mouse Y position, and then the mouse X distance from the node sets the amount that changes in the mouse Y position affect the tangent slope. This works and I'm personally fine if we keep it this way, but it may be a bit unintuitive to users - at least at first.

IanCaio:
I believe I fixed the reported bug @zonkmachine , let me know if it worked, a quick test here seems to show it's gone. I made the button updating a separate method and now I call it both when the progression type is changed or when the clip is changed.

I also addressed @messmerd review requests! The tangent lines are edited the way you described, looked like the simplest way back then. It could be changed in the future, zonkmachine also suggested something with speed sensitivity, but if possible I'd prefer to leave it for a separate PR.

src/gui/editors/AutomationEditor.cpp
IanCaio: That one will depend on which PR will be merged first. If his, then this one will have another merge conflict and that will have to be fixed anyways, and vice-versa. So I think it's safe to leave it as it is for now.

messmerd (approved):
The bug reported by zonkmachine seems to be fixed now.
I can't think of anything else that needs to be done, so it has my approval. :+1:

zonkmachine (approved):

src/gui/editors/AutomationEditor.cpp
```diff
@@ -654,6 +671,52 @@ void AutomationEditor::mousePressEvent( QMouseEvent* mouseEvent )
...
+
+							m_draggedTangentTick = POS(node);
+							if (posTicks >= m_draggedTangentTick)
+							{
+								m_draggedOutTangent = true;
+							}
+							else
+							{
+								m_draggedOutTangent = false;
+							}
```
sakertooth, line 700: Could be less verbose:
```suggestion
m_draggedOutTangent = posTicks >= m_draggedTangentTick;
```

src/gui/editors/AutomationEditor.cpp
```diff
@@ -654,6 +671,52 @@ void AutomationEditor::mousePressEvent( QMouseEvent* mouseEvent )
 				}
 			break;
 			}
+			case EditMode::EditTangents:
+			{
```
sakertooth, line 675: The amount of nesting starting from here is a bit much (which is actually why I think I had to zoom out to view this). You could make some early returns to reduce the nesting a bit, but I won't block the PR because of this.

src/gui/editors/AutomationEditor.cpp
```diff
@@ -862,6 +925,56 @@ void AutomationEditor::mouseMoveEvent(QMouseEvent * mouseEvent )
...
+							{
+								it.value().setInTangent(newTangent);
+							}
+						}
+					}
+				}
+				else if (m_mouseDownRight)
+				{
+					if (m_action == Action::ResetTangents)
+					{
```
sakertooth, line 967: You could just say:
```suggestion
				else if (m_mouseDownRight && m_action == Action::ResetTangents)
				{
```

src/core/AutomationClip.cpp
```diff
@@ -804,6 +872,11 @@ void AutomationClip::loadSettings( const QDomElement & _this )
 {
 	QMutexLocker m(&m_clipMutex);
 
+	// Legacy compatibility: Previously tangents were not stored in
+	// the project file. So if any node doesn't have tangent information
+	// we will generate the tangents
+	bool shouldGenerateTangents = false;
```
sakertooth, line 878: Why not just write ``bool shouldGenerateTangents = !element.hasAttribute("inTan") && !element.hasAttribute("outTan");`` and use ``!shouldGenerateTangents`` for line 906? Then you could just remove the else that comes after it.

src/gui/editors/AutomationEditor.cpp
```diff
@@ -862,6 +925,56 @@ void AutomationEditor::mouseMoveEvent(QMouseEvent * mouseEvent )
 				}
 			break;
 			}
+			case EditMode::EditTangents:
```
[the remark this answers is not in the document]
IanCaio: It's close in size to the `EditMode::Draw` case I think. If you don't mind I'd rather leave this for a PR that would rework all the cases, since most are large.

src/gui/editors/AutomationEditor.cpp
```diff
@@ -1267,6 +1427,12 @@ void AutomationEditor::paintEvent(QPaintEvent * pe )
 			else { cursor = s_toolDrawOut; }
 			break;
 		}
+		case EditMode::EditTangents:
+		{
+			if (m_action == Action::MoveTangent) { cursor = s_toolMove; }
+			else { cursor = s_toolEditTangents; }
```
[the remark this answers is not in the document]
IanCaio: I mostly followed the format of the others, but now I realize this is the only one that doesn't have an `else-if` clause so the ternary only fits here and not the others. So I changed it!

src/gui/editors/AutomationEditor.cpp
```diff
@@ -2102,5 +2331,37 @@ void AutomationEditorWindow::updateWindowTitle()
 	setWindowTitle( tr( "Automation Editor - %1" ).arg( m_editor->m_clip->name() ) );
 }
 
+/**
+ * @brief This method handles the AutomationEditorWindow event of changing
+ * progression types. After that, it calls updateEditTanButton so the edit
+ * tangents button is updated accordingly
+ * @param Int New progression type
+ */
```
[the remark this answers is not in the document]
IanCaio: I've mostly seen those being written in the .cpp files, but a quick search is showing them on both headers and source files. Should I still change it or leave it for when we move every method doc to headers?

src/gui/editors/AutomationEditor.cpp
IanCaio: As with the other case, this one is even smaller than `EditMode::Draw`. I agree the code blocks are kinda big, but maybe we should change them all at once later? It would take a little bit of rethinking for me to do it here, and the others would still be big.

src/gui/editors/AutomationEditor.cpp
IanCaio: I thought separating the mouse conditional from the action one would make it more readable and easier to change in the future if new actions are added/changed, though it does make it more verbose and nested. I changed to the single conditional on this last commit and can keep it if it's preferred.

src/core/AutomationClip.cpp
IanCaio: I think in this case it would behave slightly different. The idea there is that if there's a single node on the project XML that doesn't have tangent information, we should run `generateTangents()`. That would work even if some nodes have the info and some don't (that's an edge case that we won't run into unless someone manually modify a project file I think, but we end up covering for that possibility with not much cost in the code). If I just use `bool shouldGenerateTangents = !element.hasAttribute("inTan") && !element.hasAttribute("outTan");` then `shouldGenerateTangents` will only depend on the last node of the automation clip, ignoring all the others that were cycled (so if there are 5 nodes with no tangent information and the last has, it won't generate tangents for the first 5). What do you think? For now I didn't change that, but if it's still preferred I can change it quickly in a new commit.

src/gui/editors/AutomationEditor.cpp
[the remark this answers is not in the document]
sakertooth: > I've mostly seen those being written in the .cpp files, but a quick search is showing them on both headers and source files. Should I still change it or leave it for when we move every method doc to headers?

I would move any documentation you have into the headers. We mustn't move all the documentation into headers at once in a separate PR. It can always happen gradually.

src/gui/editors/AutomationEditor.cpp
sakertooth: You could combine the "chain of single if-conditions" with ``&&``, and I don't think that would hurt readability or require much thinking. Other parts of the code may still be large in terms of nesting/length, but we don't have to go along with that in the interest of being consistent with the current code (I would argue that this is exactly what we shouldn't being doing as the current code doesn't set a great example of doing things most of the time).

I said I wouldn't block the PR because of this since I do realize doing extraction/early returns can sometimes just cause unintended bugs, so I wouldn't worry about it too much. I just don't like it when the code is unnecessarily verbose and nested.

src/core/AutomationClip.cpp
sakertooth: > What do you think? For now I didn't change that, but if it's still preferred I can change it quickly in a new commit.

Sorry if my suggestion changed the behavior somehow. I was under the assumption that ``shouldGenerateTangents`` would be set true if both ``inTan`` or ``outTan`` was not found as an attribute in ``element``. My main goal here was just to reduce verbosity. I will leave to you to do what is best for the PR and the expected behavior.

src/gui/editors/AutomationEditor.cpp
Veratil: If nesting get too deep, it means you are **_highly_** motivated to separate things out into a different functions.

> I do realize doing extraction/early returns can sometimes just cause unintended bugs

Early returns don't cause bugs, they show you bugs in your thought process. :P

sakertooth: > Early returns don't cause bugs, they show you bugs in your thought process.

Just didn't want to be too hard on him for this. Although if he does go ahead with early returns and extraction, I will review the code again anyways. To me, I would think it should really be a straightforward process.

src/gui/editors/AutomationEditor.cpp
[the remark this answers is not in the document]
IanCaio: Cool, I moved the documentation added on this PR to the headers then!

src/gui/editors/AutomationEditor.cpp
IanCaio: I saw a conditional on each case that could be changed to an early return, it removed one level of nesting and the conditionals were combined on the previous PR. Do you see any other changes that could improve it? Looking broadly I didn't notice anything else but I could be missing something.

src/core/AutomationClip.cpp
IanCaio: It's been a while since I wrote the code, I had to read it to see if there was any difference so I could be wrong. But the `shouldGenerateTangents` is inside a `for` loop that goes through each node. Setting it as `bool shouldGenerateTangents = !element.hasAttribute("inTan") && !element.hasAttribute("outTan");` will set it as true/false based on the last iteration of the loop. In most normal cases that won't matter because if one node has tangent information, all others would normally have. But the behavior is slightly different as in how it's written now: if any of the nodes don't have tangent information, the tangents will be generated, even if some do have the information because inside the loop `shouldGenerateTangents` will have been set as true once. The only situation I could think of that happening is someone editing a project file XML (say, copying some automation from an older project to a new one), which is a very edge case scenario, so if it's preferred that I change it I'm fine with it.

One alternative I thought of right now would be to have: `shouldGenerateTangents = shouldGenerateTangents || (!element.hasAttribute("inTan") && !element.hasAttribute("outTan"));`. Do you think this one looks confusing for someone reading the code?

sakertooth: Sorry, I see the for-loop now (should've probably reviewed this in my editor). You obviously can't access ``element`` out of it, so my suggestion doesn't have much merit. I wish we could use ``std::all_of`` or ``std::any_of``, but the Qt Xml API doesn't really have iterators as far as I am aware.

I think it's fine then. I don't really see a better solution. Another thing is though, is that if there were 5 nodes, and one of them was missing the ``inTan`` attribute but had ``outTan``, we would still call ``generateTangents()`` even if that node did have tangent information. I'm not sure if this is possible in practice, but in theory, it might be. I think you would want to check for ``outTan`` as well on line 901.

src/gui/editors/AutomationEditor.cpp
```diff
@@ -2102,5 +2306,25 @@ void AutomationEditorWindow::updateWindowTitle()
...
+
+	if (!AutomationClip::supportsTangentEditing(progType))
+	{
+		if (m_editTanAction->isChecked()) { m_drawAction->trigger(); }
+		if (m_editTanAction->isEnabled()) { m_editTanAction->setEnabled(false); }
+	}
+	else
+	{
+		if (!m_editTanAction->isEnabled()) { m_editTanAction->setEnabled(true); }
+	}
```
sakertooth, line 2327: Maybe this might be simpler (not tested):
```suggestion
	auto progType = currentClip()->progressionType();
	m_editTanAction->setEnabled(AutomationClip::supportsTangentEditing(progType));
	if (!m_editTanAction->isEnabled() && m_editTanAction->isChecked()) { m_drawAction->trigger(); }
```

src/core/AutomationClip.cpp
```diff
@@ -489,12 +530,31 @@ TimePos AutomationClip::setDragValue(
...
+	TimePos returnedPos;
+
 	if (m_dragKeepOutValue)
 	{
-		return this->putValues(time, value, m_dragOutValue, quantPos, controlKey);
+		returnedPos = this->putValues(time, value, m_dragOutValue, quantPos, controlKey);
+	}
+	else
+	{
+		returnedPos = this->putValue(time, value, quantPos, controlKey);
```
sakertooth, line 541: The ``outValue`` parameter for this function would now be the argument ``quantPos``. Is this intended?

src/gui/editors/AutomationEditor.cpp
sakertooth: I think it's a lot better now :+1:.

src/core/AutomationClip.cpp
IanCaio: Right now, when you edit either the `inTan` or the `outTan` both tangents become "locked" and are both saved in the XML (they're also both saved even if they are "unlocked"). So in practice there shouldn't be a node with one attribute but not the other. Unless someone decides to delete one of them but not the other on the XML, in which case I'm not sure but it could cause a bug. I could set a default value on the `element.attribute("inTan")` call, though I think by default it's an empty string, so `LocaleHelper::toFloat` would convert it to 0? I could use `element.attribute("inTan", "0");` instead

src/gui/editors/AutomationEditor.cpp
IanCaio: It seems to work fine after a quick test here! Pushing the change (:

src/core/AutomationClip.cpp
IanCaio: `AutomationClip::putValue` takes a single value and uses it for both the inValue and outValue and `AutomationClip::putValues` takes a value for each of them

sakertooth: My bad, didn't see that it was a completely different function.

src/core/AutomationClip.cpp
sakertooth: > I could set a default value on the element.attribute("inTan") call, though I think by default it's an empty string, so LocaleHelper::toFloat would convert it to 0? I could use element.attribute("inTan", "0"); instead

I guess it's fine then. I also realized that ``element.attribute`` would return an empty string if the attribute is not found, and ``QLocale::setDouble`` would return ``"0"`` on failure.

sakertooth (approved):
LGTM. We should probably refactor the big switch cases in the future though (e.g., putting them into functions).

src/gui/editors/AutomationEditor.cpp
sakertooth: I'm predicting that this will get merged first, so I will make note of this :+1:.

IanCaio:
> LGTM. We should probably refactor the big switch cases in the future though (e.g., putting them into functions).

Thanks man! Agreed, we could extend that to other `mouse(Press/Move)Event` methods as well, i.e.: PianoRoll has kind of a very big method for those too afair. I'm not sure how exactly to refactor them but I'm sure there's a way to improve those.

Now that there's 3 approvals, maybe we can merge in 24h if nobody comments wishing to review it further?

PR #5926, opened 2021-03-01
Added a method to get a note as a string

SeleDreams:
This is necessary for my UTAU integration, this allows to retrieve the name out of a note such as C#2 D2 etc.

### DIFF
--- a/include/Note.h
+++ b/include/Note.h
@@ -205,6 +205,12 @@ public:
 		return classNodeName();
 	}
 
+	inline QString getNoteName() const
+	{
+        return m_nameList[m_key % 12] +
+               QString::number(static_cast<int>(m_key / KeysPerOctave));
+	}
+
 	static TimePos quantized( const TimePos & m, const int qGrid );
 
 	DetuningHelper * detuning() const
@@ -236,6 +242,21 @@ private:
 	TimePos m_length;
 	TimePos m_pos;
 	DetuningHelper * m_detuning;
+	static constexpr std::array<const char*,12> m_nameList =
+		{
+		"C",
+		"C#",
+		"D",
+		"D#",
+		"E",
+		"F",
+		"F#",
+		"G",
+		"G#",
+		"A",
+		"A#",
+		"B"
+		};
 };
 
 


include/Note.h
```diff
@@ -205,6 +205,12 @@ class LMMS_EXPORT Note : public SerializingObject
 		return classNodeName();
 	}
 
+	inline QString getNoteName() const
+	{
+        return m_nameList[m_key % 12] +
+               QString::number(static_cast<int>(m_key / KeysPerOctave));
+	}
```
Veratil, line 212: Let's have this return a `const char*` and then convert that to QString where needed.

SeleDreams: Done

Veratil: Sorry I wasn't more explicit on this. I meant entirely remove Qt references from it. So the `+ QString::number` too.

Veratil: Ideally we shouldn't have any Qt in this header.

SeleDreams: What about std:: ? because in C++ i'd more return an std::string outside of the context of Qt than const char*

Veratil: That's fine! Just trying to make sure we don't include Qt stuff in core, and Note is definitely a core thing. :)

SeleDreams: looking at the code from Note.h a lot of stuff seems to actually depend on Qt, the notevector is q QVector, the saveSettings and loadSettings depend on QDomDocument and QDomElement, classNodeName uses QString etc

it's because Note inherits from SerializingObject

Veratil: Ah crap, I didn't notice the save/load-Settings. I saw NoteVector, but that can easily be changed to a std::vector. We'll have to figure out another way later then. For now let's still try to keep as much Qt out as we can.

IanCaio: I agree we should move away from Qt where possible, it can maybe hold when it makes things too complex (for example, on the AutomationPattern I still used a QMap because it's iterator is random access while `std::map` would require some extra code to simulate random access). But even if the class makes unnecessary use of Qt, when we can we should avoid it in new code, and possibly slowly remove those existing unnecessary uses through other PRs.

This is a case where not using Qt is a small change thankfully :smile:

SeleDreams: I changed it to use a more standard c++ way now

Veratil:
Can you go through anything that uses the note names and update there now? Such as piano roll and wherever else.

include/Note.h
```diff
@@ -205,6 +205,12 @@ class LMMS_EXPORT Note : public SerializingObject
 		return classNodeName();
 	}
 
+	inline std::string getNoteName() const
```
Sawuare, line 208: I'd rename the function to `getNoteString`, because down below you're using "name" to refer to just a letter and # without an octave number.

include/Note.h
```diff
@@ -205,6 +205,12 @@ class LMMS_EXPORT Note : public SerializingObject
...
+	{
+        return m_nameList[m_key % 12] +
```
Sawuare, line 210: `12` to `KeysPerOctave`.

include/Note.h
```diff
@@ -236,6 +242,21 @@ class LMMS_EXPORT Note : public SerializingObject
 	TimePos m_length;
 	TimePos m_pos;
 	DetuningHelper * m_detuning;
+	static constexpr std::array<const char*,12> m_nameList =
```
Sawuare, line 245: While it is a "list", I'd rename to `m_nameArray` to avoid confusion with `std::list`.

Sawuare: And the prefix should be `s_`, because `static`.

Spekular: Funnily enough this suggestion illustrates why you shouldn't apply it, if the container is changed the name will become outdated. I would suggest something container-independent like `m_noteNames`.

Issue #5908, opened 2021-02-13
Rework waveform plotting


thmueller64:
### Enhancement Summary
Plotting the audio waveform is currently done in `SampleBuffer::visualize`. Creating and updating the waveform plot is very unresponsive. You can observe this for example when you load a whole track in the AudioFileProcessor and turn the amplification knob. 
Furthermore, every `n`th sample is currently plotted. In my opinion, this method of "compressing" the information in the actual waveform makes limited sense only. The maximum or RMS of the samples per pixel should be used instead, like in audacity.
#### Justification
This performance issue can be observed especially for longer samples (>2min) in the following situations:
1. Loading a sample: You have to keep the mouse pressed for 15s+ until you can place the sample in the timeline
2. Moving the amplification knob in the AudioFileProcessor

If this feature is wanted I can try to put something together after the refactor.

IanCaio:
That's definitely something worth working in. I'd add that we need to decouple `visualize` from the `SampleBuffer` class (GUI vs Core). I'm not sure which class should it be in though, since it's used on both the AFP and the SampleTCOs. Feels like an overkill to create a `SampleBufferView` class just for that, but because of it being reused it might be the way now.

DouglasDGI:
Issue resolved via #5927.

PR #5927, opened 2021-03-02
Improve performance and quality of Sample Track display, and add RMS graph.

DouglasDGI:
Before:
Screenshot from 2021-03-01 17-24-24

After:
Screenshot from 2021-04-13 22-40-21

The top row shows the default coloring.

Alright, here's what changed.
Previously, the sample track display code simply skipped a certain number of frames and snagged a random one in the middle of the sample in order to get the next point in the line.  This, of course, cannot possibly look any good without a super high point resolution.  LMMS did have a super high point resolution due to that issue... and because of that, drawing the sample tracks took a ludicrously long amount of time, resulting in absolutely massive amounts of lag every time the user zooms in/out or resizes a sample track vertically.  And, at the same time, quality would still be lower than what would be preferred when the user zooms in quite far.  The waveform would also oftentimes fail to show very short peaks in the audio.

These new formulas I put together take *every* frame of audio into account.  Each point on the line displays the maximum or minimum value of all the frames within that area.  This'll accurately display all peaks, and allows us to have a significantly smaller point resolution while still having a significantly higher quality in general.  These new changes probably run around 5x-10x as fast or so, so users should run into much less lag than before when using sample tracks.

On top of all this, I also added an Audacity-style RMS graph in the middle for bonus points, which should make LMMS look quite a bit fancier, and also give the user a decent general idea of the audio's loudness in that area relative to the peaks.  

~~This PR is a WORK IN PROGRESS, do not merge.  There are still a few bugs (e.g. selected and muted TCOs aren't shown with the right colors when a custom color isn't chosen), and the default colors might change a bit as well.~~  It's ready now.


Veratil:
Out of curiosity... what's with the 11 commits about merging some of the first PRs?

DouglasDGI:
@Veratil Me not knowing how to actually use Github properly, and making PRs from the LMMS master branch to the master branch on my fork in order to update it.
It shouldn't be a problem since we choose Squash and Merge anyway.

thmueller64:
Closes #5908 . Works very well and improves the performance a lot. Also takes into account panning, as opposed to before where the left and right channels simply have been plotted on top of each other. What do you think about drawing the peak/rms envelope, so the hull of the peak/rms values, instead of drawing lines connecting successive min-max-min-max-min values?

src/core/SampleBuffer.cpp
```diff
@@ -1017,30 +1017,63 @@ void SampleBuffer::visualize(
 	const float ySpace = h * 0.5f;
 	const int nbFrames = focusOnRange ? toFrame - fromFrame : m_frames;
 
-	const int fpp = qBound<int>(1, nbFrames / w, 20);
-	QPointF * l = new QPointF[nbFrames / fpp + 1];
-	QPointF * r = new QPointF[nbFrames / fpp + 1];
+	const int fpp = qBound<int>(1, nbFrames / w, 1000);
+	QPointF * fMax = new QPointF[nbFrames / fpp * 2 + 2];
+	QPointF * fRms = new QPointF[nbFrames / fpp * 2 + 2];
```
thmueller64, line 1022: You forgot to delete fRms at the end of the function. I'd suggested to use vectors or add the deleter.
```suggestion
	std::vector<QPointF> fMax(nbFrames / fpp * 2 + 2);
	std::vector<QPointF> fRms(nbFrames / fpp * 2 + 2);
```

src/core/SampleBuffer.cpp
```diff
@@ -1017,30 +1017,63 @@ void SampleBuffer::visualize(
...
+		fRms[n*2] = QPointF(x, (yb - (maxRmsData * py)));
+		fRms[n*2+1] = QPointF(x, (yb - (minRmsData * py)));
 		++n;
 	}
 
-	p.setRenderHint(QPainter::Antialiasing);
-	p.drawPolyline(l, nbFrames / fpp);
-	p.drawPolyline(r, nbFrames / fpp);
+	//p.setRenderHint(QPainter::Antialiasing);
+	p.drawPolyline(fMax, nbFrames * 2 / fpp);
```
thmueller64, line 1070:
```suggestion
	p.drawPolyline(fMax.data(), nbFrames * 2 / fpp);
```

src/core/SampleBuffer.cpp
```diff
@@ -1017,30 +1017,63 @@ void SampleBuffer::visualize(
...
+
+	p.setPen(p.pen().color().lighter(123));//
+
+	p.drawPolyline(fRms, nbFrames * 2 / fpp);
```
thmueller64, line 1074:
```suggestion
	p.drawPolyline(fRms.data(), nbFrames * 2 / fpp);
```

src/core/SampleBuffer.cpp
```diff
@@ -1017,30 +1017,63 @@ void SampleBuffer::visualize(
...
 
-	delete[] l;
-	delete[] r;
+	delete[] fMax;
```
thmueller64, line 1076:
```suggestion
```

src/core/SampleBuffer.cpp
```diff
@@ -1017,30 +1017,63 @@ void SampleBuffer::visualize(
 	const float ySpace = h * 0.5f;
 	const int nbFrames = focusOnRange ? toFrame - fromFrame : m_frames;
 
-	const int fpp = qBound<int>(1, nbFrames / w, 20);
-	QPointF * l = new QPointF[nbFrames / fpp + 1];
-	QPointF * r = new QPointF[nbFrames / fpp + 1];
+	const int fpp = qBound<int>(1, nbFrames / w, 1000);
```
thmueller64, line 1020: Why limit the frames per pixel to 1000 frames?

DouglasDGI:
Oh hey, I didn't realize 5908 exists.  Cool.
Also, I completely forgot that the changes I'm making also impact AudioFileProcessor.  I'm happy things worked out well for it as well, hahaha.

> What do you think about drawing the peak/rms envelope, so the hull of the peak/rms values, instead of drawing lines connecting successive min-max-min-max-min values?

I feel like that isn't needed.  In the future, LMMS will support significantly larger amounts of zoom, allowing us to see the waveform with very high detail... in which case filling between two lines would be quite odd (either wouldn't do anything at all or display incorrectly altogether), and will result in a rather significant increase in resources that would be required.  We'd likely need to have different drawing logic for different amounts of zoom.  We'll also need to fill in the area between the two lines drawn, which will require that we either draw a new line at every pixel in order to fill the space between (which would use far more resources than the current method), or maybe some Qt magic would be possible to fill in the area in between (?).  In the end, as far as I know, the result would likely be visually identical, with not much of a benefit elsewhere.

Of course, if there are good reasons to switch to that method that I'm not taking into account, definitely let me know.

thmueller64:
>  We'd likely need to have different drawing logic for different amounts of zoom

That's how audacity does it. It switches from peak/rms to single samples connected directly when zoomed close enough.

> We'll also need to fill in the area between the two lines drawn, which will require that we either draw a new line at every pixel in order to fill the space between (which would use far more resources than the current method), or maybe some Qt magic would be possible to fill in the area in between (?)

Yes, a closed path (which can be filled) consisting of all the max values from left to right plus all the min values in the opposite direction. But overall I think the drawing strategy of the PR is efficient enough and looks really good visually.

src/core/SampleBuffer.cpp
DouglasDGI: I tried removing the limit entirely, but for some reason I get a SEGFAULT when I do that.

DouglasDGI: I also tried a value of 10,000, which worked until I zoomed out.  There's something that's causing extremely large values of `fpp` to crash.

DouglasDGI: Oh, that was a bug.  Sometimes `frame + i` wound up being larger than `last`, forgot to check for that.  I can remove that limit now.

DouglasDGI:
Alright, everything is finished now.  This is no longer a WIP, it'll be ready for merge as soon as you all deem it as such.

src/core/SampleBuffer.cpp
```diff
@@ -1017,30 +1017,61 @@ void SampleBuffer::visualize(
...
+		fRms[n*2] = QPointF(x, (yb - (maxRmsData * py)));
+		fRms[n*2+1] = QPointF(x, (yb - (minRmsData * py)));
 		++n;
 	}
 
-	p.setRenderHint(QPainter::Antialiasing);
-	p.drawPolyline(l, nbFrames / fpp);
-	p.drawPolyline(r, nbFrames / fpp);
+	//p.setRenderHint(QPainter::Antialiasing);
+	p.drawPolyline(fMax.data(), nbFrames * 2 / fpp);
```
Veratil, line 1070: Since we reuse `nbFrames * 2 / fpp` we could pull that out to a variable. Compiler should be smart enough to see this, but doesn't hurt to help.

src/core/SampleBuffer.cpp
```diff
@@ -1017,30 +1017,61 @@ void SampleBuffer::visualize(
...
 	{
+		float maxData = -1;
+		float minData = 1;
+
+		float rmsData[2] = {0, 0};
+
+		// Find maximum and minimum samples within range
+		for (int i = 0; i < fpp && frame + i < last; ++i)
+		{
+			for (int j = 0; j < 2; ++j)
```
Veratil, line 1038: Is this loop here for future more than left-right channel?

src/core/SampleBuffer.cpp
```diff
@@ -1017,30 +1017,61 @@ void SampleBuffer::visualize(
...
+
 		auto x = xb + ((frame - first) * double(w) / nbFrames);
 		// Partial Y calculation
 		auto py = ySpace * m_amplification;
-		l[n] = QPointF(x, (yb - (m_data[frame][0] * py)));
-		r[n] = QPointF(x, (yb - (m_data[frame][1] * py)));
+		fMax[n*2] = QPointF(x, (yb - (maxData * py)));
+		fMax[n*2+1] = QPointF(x, (yb - (minData * py)));
+		fRms[n*2] = QPointF(x, (yb - (maxRmsData * py)));
+		fRms[n*2+1] = QPointF(x, (yb - (minRmsData * py)));
```
Veratil, line 1065: `n` isn't very descriptive in this context. I realize after scrolling up above the loop it's the current loop counter, but we could rename it to something like `curFrame` or whatnot.

src/tracks/SampleTrack.cpp
```diff
@@ -574,9 +574,19 @@ void SampleTCOView::paintEvent( QPaintEvent * pe )
...
+	QLinearGradient lingrad( 0, 0, 0, height() );
+	QColor c = painter.background().color();
+	if (muted)
+	{
+		c = c.darker(150);
+	}
+	if (selected)
+	{
+		c = c.darker(150);
+	}
```
Veratil, line 589: Short ifs.
```suggestion
	if (muted) { c = c.darker(150); }
	if (selected) { c = c.darker(150); }
```

src/tracks/SampleTrack.cpp
```diff
@@ -593,7 +603,24 @@ void SampleTCOView::paintEvent( QPaintEvent * pe )
 		p.fillRect( rect(), c );
 	}
 
-	p.setPen( !muted ? painter.pen().brush().color() : mutedColor() );
+	auto tcoColor = m_tco->hasColor()
+				? m_tco->usesCustomClipColor()
+					? m_tco->color()
+					: m_tco->getTrack()->color()
+				: painter.pen().brush().color();
```
Veratil, line 610: I think this might be more readable while using less tabs. I'm aware the extra parenthesis are unnecessary, but to my eyes it groups it better.
```suggestion
	auto tcoColor =
		m_tco->hasColor()
		? (m_tco->usesCustomClipColor()
			? m_tco->color()
			: m_tco->getTrack()->color())
		: painter.pen().brush().color();
```

Veratil: Alternatively
```suggestion
	auto clipColor = m_tco->usesCustomClipColor() ? m_tco->color() : m_tco->getTrack()->color();
	auto tcoColor = m_tco->hasColor() ? clipColor : painter.pen().brush().color();
```

src/core/SampleBuffer.cpp
```diff
@@ -1017,30 +1017,61 @@ void SampleBuffer::visualize(
...
+				{
+					maxData = m_data[frame + i][j];
+				}
+
+				if (m_data[frame + i][j] < minData)
+				{
+					minData = m_data[frame + i][j];
+				}
+
+				rmsData[j] += m_data[frame + i][j] * m_data[frame + i][j];
```
Veratil, line 1050: Lots of repeated same data access, again compiler should be smart to reuse, but wouldn't hurt to help and then make it a little more readable.
```suggestion
				auto curData = m_data[frame + i][j];
				
				if (curData > maxData) { maxData = curData; }
				if (curData < minData) { minData = curData; }

				rmsData[j] += curData * curData;
```

src/core/SampleBuffer.cpp
```diff
@@ -1017,30 +1017,57 @@ void SampleBuffer::visualize(
...
+
+		// Find maximum and minimum samples within range
+		for (int i = 0; i < fpp && frame + i < last; ++i)
+		{
+			for (int j = 0; j < 2; ++j)
+			{
+				auto curData = m_data[frame + i][j];
+
+				if (curData > maxData) { maxData = m_data[frame + i][j]; }
+				if (curData < minData) { minData = m_data[frame + i][j]; }
```
Veratil, line 1044:
```suggestion
				if (curData > maxData) { maxData = curData; }
				if (curData < minData) { minData = curData; }
```

src/core/SampleBuffer.cpp
```diff
@@ -1017,30 +1017,57 @@ void SampleBuffer::visualize(
...
+		fRms[curFrame*2] = QPointF(x, (yb - (maxRmsData * py)));
+		fRms[curFrame*2+1] = QPointF(x, (yb - (minRmsData * py)));
+		++curFrame;
 	}
 
-	p.setRenderHint(QPainter::Antialiasing);
-	p.drawPolyline(l, nbFrames / fpp);
-	p.drawPolyline(r, nbFrames / fpp);
+	//p.setRenderHint(QPainter::Antialiasing);
+	p.drawPolyline(fMax.data(), totalPoints);
```
Veratil, line 1066: Is the extra `+ 2` in `totalPoints` going to change from what it was before (`nbFrames / fpp * 2`)?

Veratil (approved):

src/core/SampleBuffer.cpp
```diff
@@ -1017,30 +1017,57 @@ void SampleBuffer::visualize(
...
+		fMax[curFrame*2+1] = QPointF(x, (yb - (minData * py)));
+		fRms[curFrame*2] = QPointF(x, (yb - (maxRmsData * py)));
+		fRms[curFrame*2+1] = QPointF(x, (yb - (minRmsData * py)));
+		++curFrame;
 	}
 
-	p.setRenderHint(QPainter::Antialiasing);
-	p.drawPolyline(l, nbFrames / fpp);
-	p.drawPolyline(r, nbFrames / fpp);
+	//p.setRenderHint(QPainter::Antialiasing);
```
Sawuare, line 1065: Looks like a testing leftover.

src/core/SampleBuffer.cpp
```diff
@@ -1017,30 +1017,57 @@ void SampleBuffer::visualize(
...
+	p.drawPolyline(fMax.data(), totalPoints);
+
+	p.setPen(p.pen().color().lighter(123));//
```
Sawuare, line 1068: Empty comment at the end.

src/tracks/SampleTrack.cpp
```diff
@@ -574,9 +574,13 @@ void SampleTCOView::paintEvent( QPaintEvent * pe )
 
 	QPainter p( &m_paintPixmap );
 
-	QLinearGradient lingrad( 0, 0, 0, height() );
-	QColor c = getColorForDisplay( painter.background().color() );
 	bool muted = m_tco->getTrack()->isMuted() || m_tco->isMuted();
+	bool selected = isSelected();
+
+	QLinearGradient lingrad( 0, 0, 0, height() );
```
IanCaio, line 580:
```suggestion
	QLinearGradient lingrad(0, 0, 0, height());
```

Minor style fix

src/tracks/SampleTrack.cpp
```diff
@@ -593,7 +597,24 @@ void SampleTCOView::paintEvent( QPaintEvent * pe )
...
 	}
 
-	p.setPen( !muted ? painter.pen().brush().color() : mutedColor() );
+	auto tcoColor = m_tco->hasColor()
+			? (m_tco->usesCustomClipColor()
+				? m_tco->color()
+				: m_tco->getTrack()->color())
+			: painter.pen().brush().color();
+
+	p.setPen( tcoColor );
```
IanCaio, line 606:
```suggestion
	p.setPen(tcoColor);
```

src/tracks/SampleTrack.cpp
```diff
@@ -593,7 +597,24 @@ void SampleTCOView::paintEvent( QPaintEvent * pe )
...
+
+	if (muted)
+	{
+		QColor penColor = p.pen().brush().color();
+		penColor.setHsv( penColor.hsvHue(), penColor.hsvSaturation() / 4, penColor.value() );
```
IanCaio, line 611:
```suggestion
		penColor.setHsv(penColor.hsvHue(), penColor.hsvSaturation() / 4, penColor.value());
```

src/core/SampleBuffer.cpp
```diff
@@ -1017,30 +1017,56 @@ void SampleBuffer::visualize(
 	const float ySpace = h * 0.5f;
 	const int nbFrames = focusOnRange ? toFrame - fromFrame : m_frames;
 
-	const int fpp = qBound<int>(1, nbFrames / w, 20);
-	QPointF * l = new QPointF[nbFrames / fpp + 1];
-	QPointF * r = new QPointF[nbFrames / fpp + 1];
-	int n = 0;
+	const int fpp = qMax<int>(1, nbFrames / w);
+	const int totalPoints = nbFrames / fpp * 2 + 2;
```
IanCaio, line 1021:
```suggestion
	const int totalPoints = nbFrames / fpp * 2;
```

After reading the method, I still can't see why those extra 2 points are necessary. It seems it has been tested without it with no issues. Just for safety maybe you could remove it and add an statement on the `for` loop, something like:
`if((curFrame + 1) >= totalPoints) { qWarning("Out of bounds"); }`

Just temporarily, and test it with lots of samples to see whether any of them would trigger it. But from reading the logic it doesn't seem to me that it would ever reach that index.

IanCaio (changes requested):
Everything looks good, except for some small changes and minor code style issues.

src/core/SampleBuffer.cpp
DouglasDGI: I just tested it out without the +2 and the program would abort entirely every time I tried doing anything with a sample track.  I assume it's needed for some reason.

IanCaio: That's odd. I tried to use the Peaks/Troughs approach on that Oscillator plugin but I used a slightly different approach, so it's hard to compare but I didn't need extra points. Did you manage to get a backtrace on the crashes?

I'm trying to debug it here to figure it out, I'm just a bit averse to leave it there without at least knowing what it's for, even though it was already there :laughing: . If it's necessary, I'll try to figure out why so we can at least add a comment.

IanCaio: Small test indicates the logic is has a small flaw: We are drawing past the width of the widget:

```
nbFrames:  166623
Width:  35
fpp:  4760
totalPoints:  72
first:  0
last:  166622

Out of bounds
curFrame:  35
index:  71
x:  35.9953


Out of bounds
curFrame:  36
index:  73
x:  36.9952
```

As can be seen, both "out of bound access" (that reach those extra two points) have an X value higher then the widget's width. 166623 frames is not perfectly divisible by 35 pixels (results in 4760.66), so the `fpp` is floored to 4760. By fitting the 35 pixels we have consumed `4760 * 35` frames, which is 166600. Those extra 23 frames are being drawn too when they should be discarded apparently. If not discarded, they should be used only for the peak/trough calculation of pixel 35.

I'll later try to come up with some fix, but at least the mystery seems to be solved :smile:

IanCaio: Also, somewhat unrelated, but can `curFrame` be renamed to `curPixel`? Think it's a more accurate name

IanCaio (approved):
Pushed some changes that fixes some raised issues. The code LGTM now.

There's one small behavior I didn't figure out the reason why (present on both `master`, Douglas version and mine) where zooming in a lot will make the wave being drawn slightly before the end, even with the proportion calculation. I'm not sure if this has to do with float approximation. This PR doesn't introduce it so I'm approving anyways.

src/core/SampleBuffer.cpp
```diff
@@ -1017,30 +1025,75 @@ void SampleBuffer::visualize(
...
+				auto curData = m_data[frame + i][j];
+
+				if (curData > maxData) { maxData = curData; }
+				if (curData < minData) { minData = curData; }
+
+				rmsData[j] += curData * curData;
+			}
+		}
+
+		const float trueRmsData = (rmsData[0] + rmsData[1]) * 0.5 / fpp;
```
Veratil, line 1072: Checking in godbolt, it looks like we can make this save some cycles by using `/ 2` instead of `* 0.5`.

src/core/SampleBuffer.cpp
```diff
@@ -1017,30 +1025,75 @@ void SampleBuffer::visualize(
...
+		const float sqrtRmsData = sqrt(trueRmsData);
+		const float maxRmsData = qBound(minData, sqrtRmsData, maxData);
+		const float minRmsData = qBound(minData, -sqrtRmsData, maxData);
+
+		// If nbFrames >= w, we can use curPixel to calculate X
+		// but if nbFrames < w, we need to calculate it proportionally
+		// to the total number of points
+		auto x = nbFrames >= w
+			? xb + curPixel
+			: xb + ((double(curPixel) / nbFrames) * w);
```
Veratil, line 1082:
```suggestion
			: xb + ((static_cast<double>(curPixel) / nbFrames) * w);
```

src/core/SampleBuffer.cpp
DomClark: I'm not convinced by this. The number of cycles doesn't correspond exactly to the number of instructions - some are more expensive than others. While the division version has one less instruction in total, it uses a division instruction instead of a multiplication instruction, which is considerably more expensive. Besides, at optimisation levels `-O1` and above, identical code is generated for each (the compiler converts the division to a multiplication, because it's faster). Guessing at microoptimisations like this is nearly useless; actually measuring the execution time is the only sensible thing to do.

(That said, I prefer `/ 2` here anyway - it stands out more clearly to me as an average that way.)

Veratil: Yeah, I did say *might* save some cycles. 😁 You're right though it's not proper to try and optimize like that. I was curious and wanted to see what both turned into.

Veratil:
Merging tomorrow unless anyone objects.

DouglasDGI:
My original method of drawing the sample tracks left a noticeable line thickness difference between the top and bottom half of the waveforms, the latest two commits fix this.

DomClark:
The `fpp` variable is an integer, so isn't the exact number of frames per pixel. Previously this wasn't an issue, since the x coordinate of each point was calculated from the frame offset, but now the x coordinate is advanced by exactly one pixel each iteration, this introduces an error of up to one frame per pixel. I would suggest using a `float` or `double` for `fpp` and `frame`.

Veratil:
Giving one more day before merge after Dom's review just in case anything else pops up.

PR #5928, opened 2021-03-02
Improves the PianoRoll detuning drawing

IanCaio:
This PR rewrites the detuning drawing method of the PianoRoll to behave similar to the AutomationEditor. It now draws Cubic Hermite progressions as well, and the detuning is displayed as a shape instead of a line.

The color of the detuning shape is the same as `m_noteColor`, except it's lighter and has some alpha.

Detuning

There's one visual glitch left to fix, which is also present in `master`: Changes to the automation node's outValues don't immediately trigger a repaint on the PianoRoll, because they are done directly on the `AutomationNode` class which doesn't emit the `dataChanged()` signal on the Automation Pattern. This is an easy fix, once it's decided how to approach it. Waiting for some opinions on the Discord channel before working on the fix.

### DIFF
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -976,82 +976,88 @@ void PianoRoll::drawNoteRect( QPainter & p, int x, int y,
 
 
 
-void PianoRoll::drawDetuningInfo( QPainter & _p, const Note * _n, int _x,
-								int _y ) const
+void PianoRoll::drawDetuningInfo(QPainter & p, const Note * n, int x, int y) const
 {
-	int middle_y = _y + m_keyLineHeight / 2;
-	_p.setPen(m_noteColor);
-	_p.setClipRect(
+	// Get a timeMap of the detuning pattern
+	AutomationPattern* detuningPattern = n->detuning()->automationPattern();
+	timeMap & detuningMap = detuningPattern->getTimeMap();
+
+	// Return if it's empty
+	if (detuningMap.isEmpty()) { return; }
+
+	// Set the detuning color to the same color as the notes,
+	// but lighter and with some alpha
+	QColor detuningColor = QColor(m_noteColor.lighter(150));
+	detuningColor.setAlpha(70);
+
+	// Get the first node
+	timeMap::iterator it = detuningMap.begin();
+
+	// Set the clip area to the PianoRoll area
+	p.setClipRect(
 		m_whiteKeyWidth,
 		PR_TOP_MARGIN,
 		width() - m_whiteKeyWidth,
-		keyAreaBottom() - PR_TOP_MARGIN);
+		keyAreaBottom() - PR_TOP_MARGIN
+	);
 
-	// Draw lines for the detuning automation, treating cubic hermit curves
-	// as straight lines for now. Also draw discrete jumps.
-	int old_x = 0;
-	int old_y = 0;
+	// Reference pixel Y (that will be equivalent to
+	// zero in the automation pattern)
+	int baseY = y + m_keyLineHeight / 2;
 
-	timeMap & map = _n->detuning()->automationPattern()->getTimeMap();
-	for (timeMap::const_iterator it = map.begin(); it != map.end(); ++it)
+	// Two lambda functions to calculate the X and Y position
+	// of nodes/values
+	auto getX = [this, x](int tick)
 	{
-		// Current node values
-		int cur_ticks = POS(it);
-		int cur_x = _x + cur_ticks * m_ppb / TimePos::ticksPerBar();
-		const float cur_level = INVAL(it);
-		int cur_y = middle_y - cur_level * m_keyLineHeight;
+		return x + tick * m_ppb / TimePos::ticksPerBar();
+	};
+	auto getY = [this, baseY](float value)
+	{
+		return baseY - value * m_keyLineHeight;
+	};
 
-		// First line to represent the inValue of the first node
-		if (it == map.begin())
+	while (it + 1 != detuningMap.end())
+	{
+		// Get the current node's x position in ticks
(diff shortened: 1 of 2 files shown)


src/gui/editors/PianoRoll.cpp
```diff
@@ -976,82 +976,88 @@ void PianoRoll::drawNoteRect( QPainter & p, int x, int y,
...
+	// Get the first node
+	timeMap::iterator it = detuningMap.begin();
+
+	// Set the clip area to the PianoRoll area
+	p.setClipRect(
 		m_whiteKeyWidth,
 		PR_TOP_MARGIN,
 		width() - m_whiteKeyWidth,
-		keyAreaBottom() - PR_TOP_MARGIN);
+		keyAreaBottom() - PR_TOP_MARGIN
```
Veratil, line 1001:
```suggestion
		keyAreaBottom() - keyAreaTop()
```

src/gui/editors/PianoRoll.cpp
```diff
@@ -976,82 +976,88 @@ void PianoRoll::drawNoteRect( QPainter & p, int x, int y,
...
+	// Return if it's empty
+	if (detuningMap.isEmpty()) { return; }
+
+	// Set the detuning color to the same color as the notes,
+	// but lighter and with some alpha
+	QColor detuningColor = QColor(m_noteColor.lighter(150));
+	detuningColor.setAlpha(70);
+
+	// Get the first node
+	timeMap::iterator it = detuningMap.begin();
```
Veratil, line 994: Let's move this to before the loop instead of before the lambdas, or more the lambdas up.

cyber-bridge:
I've tried to compile it but it failed (Qt version 5.15.2):

```
/home/unknown/lmms/detuningDrawing/src/gui/editors/PianoRoll.cpp:1046:16: error: aggregate 'QPainterPath path' has incomplete type and cannot be defined
 1046 |   QPainterPath path;
      |                ^~~~
```

After adding `#include <QPainterPath>` to `src/gui/editors/PianoRoll.cpp` it did compile.

Looks great! :-D

src/gui/editors/PianoRoll.cpp
IanCaio: Done! We should soon open a PR and run some `sed` to replace it. PR_TOP_MARGIN is still being used in lots of places (including the original drawDetuningInfo method).

IanCaio:
> I've tried to compile it but it failed (Qt version 5.15.2):
> 
> ```
> /home/unknown/lmms/detuningDrawing/src/gui/editors/PianoRoll.cpp:1046:16: error: aggregate 'QPainterPath path' has incomplete type and cannot be defined
>  1046 |   QPainterPath path;
>       |                ^~~~
> ```
> 
> After adding `#include <QPainterPath>` to `src/gui/editors/PianoRoll.cpp` it did compile.
> 
> Looks great! :-D

Added the missing header! I've no idea how it compiled without complaints here

src/gui/editors/PianoRoll.cpp
Veratil: We can update it as it comes along. I'd rather not worry about that level of detail just yet.

cyber-bridge:
> Added the missing header! I've no idea how it compiled without complaints here

I can only guess that your version of Qt is below 5.15. Probably this commit is responsible: https://code.qt.io/cgit/qt/qtbase.git/commit/?id=50d2acdc93b4de2ba56eb67787e2bdcb21dd4bea

LightningAA:
I personally like lines better than the shapes but other than that it looks good

superpaik:
A part from the little glitch thing you commented, I think that if we use shapes (I like it better than lines, because is in line with the automation detuning editor), we have to modify how it is drawing because it can lead to user confusion. If I'm not wrong, both detuning automations (see attached images) have the same shape in the piano roll editor. One solution would be having in the piano roll the same shape as the automation editor or another, to highlight the shape contour according the automation (second image. Don't take into account the bad quality of the mock-up, it's just an idea ;-)
imatge
imatge

IanCaio:
> A part from the little glitch thing you commented, I think that if we use shapes (I like it better than lines, because is in line with the automation detuning editor), we have to modify how it is drawing because it can lead to user confusion. If I'm not wrong, both detuning automations (see attached images) have the same shape in the piano roll editor. One solution would be having in the piano roll the same shape as the automation editor or another, to highlight the shape contour according the automation (second image. Don't take into account the bad quality of the mock-up, it's just an idea ;-)

True, well noticed! Highlighting the contour might be a little tricky because even though it looks like one single shape, the drawings are actually many shapes close together, but I'd have to look into it to be sure. Maybe there's a way to add that contour that is not really complicated.

But I thought of an alternative too, which would be extending the shape until the end of the note. Then the first one would continue at C5 until the end and the second one would have that discrete jump and continue on top of the note. What do you think?

superpaik:
Yes, that'd be great! But have in mind that user can change the note duration once automation has been established, so it would have to change accordingly, and maybe that is not that easy. But if possible, I think it's a good solution.

IanCaio:
True, once I get some time to work on it I'll check which one will add less complexity: Reworking the path generation so I can have a single path and then use the outline or drawing the detuning all the way to the end of note. The latter might not even require that I worry about the note resizing because the Piano Roll redrawing is connected to the Pattern `dataChanged` signal, so every time a note is resized the Piano Roll will redraw anyways, and the detuning shape will be redrawn too (now accounting to the resized note)

zonkmachine:
detune

Bloody brilliant, want this! Glitches that are already in LMMS doesn't have to be fixed here for a merge. That's my opinion anyway.

zonkmachine:
> or drawing the detuning all the way to the end of note.

Maybe extend the detuning to the end of the note release and let the drawing fade too?

PR #5932, opened 2021-03-04
Added a null check before checking the flag of an instrument

SeleDreams:
I added this check because else LMMS would crash if a track had no instrument, when we want to add midi notes, this is unlikely to happen for standard InstrumentTracks but is still important, as it caused crashes during my development of a new type of track for vocal instruments.

### DIFF
--- a/src/core/NotePlayHandle.cpp
+++ b/src/core/NotePlayHandle.cpp
@@ -106,7 +106,7 @@ NotePlayHandle::NotePlayHandle( InstrumentTrack* instrumentTrack,
 		m_instrumentTrack->midiNoteOn( *this );
 	}
 
-	if( m_instrumentTrack->instrument()->flags() & Instrument::IsSingleStreamed )
+	if(m_instrumentTrack->instrument() && m_instrumentTrack->instrument()->flags() & Instrument::IsSingleStreamed )
 	{
 		setUsesBuffer( false );
 	}


JohannesLorenz (approved):
LGTM

IanCaio:
@DomClark No rush to answer as it's something simple to reverse if necessary, but just double checking: I remember you mentioning null-checks for invalid states could potentially hide bugs (i.e.: if an instrument track without an instrument is an invalid state, this hides the issue until it shows itself somewhere else). Is that true here? If so, should we replace the sanity check with an ASSERT?

DomClark:
Yes, this is one of the places I think an assertion would be better. I'm inclined to leave this as is for the moment, however, and address null safety more generally across the codebase in the future. I think we should annotate when pointers can/cannot be null, either with code or in doxygen comments, and back these up with well-placed assertions. Assuming any pointer can be null and testing before each use is definitely not the way to go.

JohannesLorenz:
Could we write our own class template, like `unique_ptr_not_null<T>`?

DomClark:
That's certainly an option. There are a couple of libraries out there that offer this sort of thing too, such as the [GSL](https://github.com/microsoft/GSL/blob/main/include/gsl/pointers) and Dropbox's [nn](https://github.com/dropbox/nn/blob/master/nn.hpp). Alternatively we could just use a transparent type alias, such as `template<typename T> using NotNull = T;`. These all have slightly different semantics, with various advantages and disadvantages, but I think it would be better to discuss these in a dedicated issue or on Discord, rather than at the bottom of an already-merged PR.

Issue #5518, opened 2020-05-27
Misaligned knob in the BitCrush effect


dj-pixus:
In the Hungarian translation, the Gain knob is misaligned due to the longer word:
![erosites](https://user-images.githubusercontent.com/65627470/83018950-72983300-a026-11ea-9ef8-580295dcbf1e.png)
LMMS 1.2.1, 64 bit AppImage from lmms.io.
I don't know however, what would be a better name for this. The closest meaning of "erősítés" is more like amplification rather than gain.

PR #5934, opened 2021-03-06
Fix misalignment of knobs caused by wide labels

zynskeywolf:
This PR fixes the main cause of #5518. The positions are now recalculated when setting the label so that the knobs stay in place more or less. Also corrected the dirty compensations in builtin effects.


rubiefawn:
Knob internals were modified quite a bit in #7525. Mentioning this since it's relevant for future work on this PR.

PR #5939, opened 2021-03-07
Allow smooth horizontal scrolling in Song Editor

LostRobotMusic:
Currently, the Song Editor's horizontal scroll bar has a precision of 1 tick per bar.  This PR vastly increases that precision, allowing much smoother and nicer-looking horizontal scrolling.

(Like always, sorry for commit hell.  They'll all be fixed once Squash and Merge is chosen.)


superpaik:
Tested on windows 10. I have found no errors.

DouglasDGI:
I made the scroll precision smarter, but there's currently a bug that makes the scroll position move weirdly when resizing the window.  Not too major of a bug, but probably something that should be fixed anyway.

LostRobotMusic:
Taking a look at this PR in the far future, coding mistakes aside, I'm finding many more bugs than I was initially aware of.  An implementation of this feature would likely have to take an entirely different approach, so I see no value in keeping this open.  I may attempt to tackle this again in the future in a separate PR.

PR #5942, opened 2021-03-10
Save Piano Roll's behavior at stop state

### LINKED ISSUE #5938: Playhead stop-behavior isn't saved for piano roll
Same issue as #4192 but for piano roll rather than the song editor. We have quite a lot of issues under the 1.3 milestone that probably won't make it in, but I think this one should.

#### Affected LMMS versions

1.3.0-alpha.1

ryuukumar:
Fixes #5938.

For some reason this seems to affect only the behavior at stop state, and not the loop points. My workaround was to save the state as an element in the piano roll, however suggestions are welcome.

### DIFF
--- a/include/TimeLineWidget.h
+++ b/include/TimeLineWidget.h
@@ -118,6 +118,11 @@ public:
 		return m_behaviourAtStop;
 	}
 
+	void setBehaviourAtStop (int state)
+	{
+		emit loadBehaviourAtStop (state);
+	}
+
 	bool loopPointsEnabled() const
 	{
 		return m_loopPoints == LoopPointsEnabled;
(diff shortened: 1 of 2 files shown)


Veratil (approved):

Spekular:
> For some reason this seems to affect only the behavior at stop state, and not the loop points.

Can you clarify what you mean by this? Loop points in piano roll aren't saved either based on my testing (on master built today).

Looking into things I found that TimeLineWidget has a restoreState, since it's a SerializingObject, but this is never called when the piano roll's state is being restored. What's even weirder is that I can't figure out how the timeline widget state is restored for the song editor (I was thinking we should do it the same way for piano roll). I'm kind of losing it trying to figure out what's going on here, as all references to TimeLineWidget in Song.cpp are via `m_playPos[m_playMode].m_timeLine`... Why does a class inheriting from TimePos store a timeline :upside_down_face:

PR #5951, opened 2021-03-15
Allow assigning all tracks from a BBTrack to a FX channel

IanCaio:
I'm not sure people would actually find this useful, but at the beginning of production when I'm still working the drums inside the BBEditor I tend to change samples a lot and having them all grouped in a single FX channel (unless an individual sample needs some specific processing).

I wrote this so I could save some time assigning all the tracks from BBEditor to a particular FX channel. It works the same as the "Assign to FX channel" of regular tracks but it affects all tracks inside the BB. Opening a PR in case people find this would be a nice addition.

### DIFF
--- a/include/BBTrack.h
+++ b/include/BBTrack.h
@@ -168,10 +168,16 @@ public:
 		return( m_bbTrack );
 	}
 
+	// Create menu to assign all tracks from the BBTrack to a FX channel
+	QMenu* createFxMenu(QString title, QString newFxLabel) override;
 
 public slots:
 	void clickedTrackLabel();
 
+
+private slots:
+	void createFxLine();
+	void assignFxLine(int channelIndex);
 
 private:
 	BBTrack * m_bbTrack;
--- a/src/tracks/BBTrack.cpp
+++ b/src/tracks/BBTrack.cpp
@@ -551,7 +551,7 @@ void BBTrackView::clickedTrackLabel()
 QMenu* BBTrackView::createFxMenu(QString title, QString newFxLabel)
 {
 	// We have a different title for BB Tracks
-	QMenu* fxMenu = new QMenu("Assign BBTrack to FX channel");
+	QMenu* fxMenu = new QMenu(tr("Assign all to FX channel"));
 
 	fxMenu->addAction(newFxLabel, this, SLOT(createFxLine()));
 	fxMenu->addSeparator();
--- a/src/tracks/BBTrack.cpp
+++ b/src/tracks/BBTrack.cpp
@@ -31,11 +31,15 @@
 #include "BBTrackContainer.h"
 #include "embed.h"
 #include "Engine.h"
+#include "FxMixer.h"
+#include "FxMixerView.h"
 #include "gui_templates.h"
 #include "GuiApplication.h"
+#include "InstrumentTrack.h"
 #include "MainWindow.h"
 #include "Mixer.h"
 #include "RenameDialog.h"
+#include "SampleTrack.h"
 #include "Song.h"
 #include "SongEditor.h"
 #include "ToolTip.h"
@@ -542,4 +546,64 @@ void BBTrackView::clickedTrackLabel()
 	Engine::getBBTrackContainer()->setCurrentBB( m_bbTrack->index() );
 	gui->getBBEditor()->parentWidget()->show();
 	gui->getBBEditor()->setFocus( Qt::ActiveWindowFocusReason );
+}
+
+QMenu* BBTrackView::createFxMenu(QString title, QString newFxLabel)
+{
+	// We have a different title for BB Tracks
+	QMenu* fxMenu = new QMenu("Assign BBTrack to FX channel");
+
+	fxMenu->addAction(newFxLabel, this, SLOT(createFxLine()));
+	fxMenu->addSeparator();
+
+	for (int i = 0; i < Engine::fxMixer()->numChannels(); ++i)
+	{
+		FxChannel* curFX = Engine::fxMixer()->effectChannel(i);
+
+		auto index = curFX->m_channelIndex;
+
+		QString label = tr("FX: %1: %2").arg(index).arg(curFX->m_name);
+
+		fxMenu->addAction(label, [this, index](){
+			assignFxLine(index);
+		});
+	}
+
+	return fxMenu;
+}
+
+void BBTrackView::createFxLine()
+{
+	int channelIndex = gui->fxMixerView()->addNewChannel();
+	auto channel = Engine::fxMixer()->effectChannel(channelIndex);
+
+	channel->m_name = getBBTrack()->name();
+	if (getTrack()->useColor()) { channel->setColor(getTrack()->color()); }
+
+	assignFxLine(channelIndex);
+}
+
+void BBTrackView::assignFxLine(int channelIndex)
+{
+	// Assign all tracks to channel:
+	TrackContainer::TrackList tl = Engine::getBBTrackContainer()->tracks();
+
+	for
+	(
+		TrackContainer::TrackList::iterator it = tl.begin();
+		it != tl.end();
+		++it
+	)
+	{
+		if((*it)->type() == Track::TrackTypes::InstrumentTrack)
+		{
+			InstrumentTrack* t = dynamic_cast<InstrumentTrack*>(*it);
+			if (t) { t->effectChannelModel()->setValue(channelIndex); }
+		}
+		else if ((*it)->type() == Track::TrackTypes::SampleTrack)
+		{
+			SampleTrack* t = dynamic_cast<SampleTrack*>(*it);
+			if (t) { t->effectChannelModel()->setValue(channelIndex); }
+		}
+	}
 }


IanCaio:
Updated PR with 1 new feature. I figured that they are somewhat related (both affect BBTracks) and that they will be easier to rebase as a single PR once the refactoring is finished.

IanCaio:
Just a reminder:

I found an issue, where I think automatable model IDs are being reused. I think that's due to the fact I'm using `TrackContentObject::copyStateTo` to copy one TCO from one track to another, when it's probably designed to copy the state of TCOs within the same track. Maybe `saveState` is also saving information about the owner track and by pasting that information on another one the automatable model IDs are being screwed up. I also didn't call `AutomationPattern::resolveAllIDs` at the end of the method.

I'm going to fix it later, could either use the `TCOView` copy/paste methods (downside is it would use the clipboard) or change the `copyStateTo` method so it's safe to use between TCOs of separate tracks (more workful, but probably better)

allejok96:
What if you only wanted to convert one BB track to song editor? I know this would create lots of duplicate instrument tracks if this action is repeated... But still, the current implementation creates multiple TCOs on top of each other in the same track if an instrument is used in multiple BB tracks in parallell. I don't know what's worse, but at least having duplicate instruments are more obvious than duplicate TCOs (oh if only we had grouping/layering of tracks!) Maybe we could have both features? Convert this and convert all. And put them in the gear menu of the track?

IanCaio:
> What if you only wanted to convert one BB track to song editor? I know this would create lots of duplicate instrument tracks if this action is repeated... But still, the current implementation creates multiple TCOs on top of each other in the same track if an instrument is used in multiple BB tracks in parallell. I don't know what's worse, but at least having duplicate instruments are more obvious than duplicate TCOs (oh if only we had grouping/layering of tracks!) Maybe we could have both features? Convert this and convert all. And put them in the gear menu of the track?

That's possible and probably simple to add: The gear menu could just call `Song::convertBBtoSE` but giving a BB ID as reference.

I initially thought about making it per BBTrack, but I changed to affect all BBTracks exactly because of the track duplication issue. I even thought of identifying the tracks that were originated from a BB but that adds extra complexity that I think is unworthy for a feature that would be there just to make do while we don't have linked patterns and the note types (when we wouldn't need BB anymore).

Adding both alternatives of converting all and converting a single BBTrack sounds like a good middle ground though, I'll just wait until I fix this TCO copying issue and I can include it.

IanCaio:
@allejok96 I just implemented the single BBTrack conversion on the Gear menu. Now user can both convert all or convert a single one.

As for the bug I found earlier, apparently this is not related to this PR (or possibly not a bug at all): When we clone a track that has effects whose parameters have assigned IDs LMMS triggers some messages like `JO-ID <number> already in use by automatablemodel:<model>`, which probably means that the effects from both tracks (cloned and original) are using the same automation ID. I'm not sure how problematic that is, but this is already an issue when a track is cloned through the Gear menu.

Since converting BBTracks to Song editor involves cloning the tracks contained in the BBTrack, this kind of situation can happen.

allejok96:
Good job. Think it would be better if it skipped adding empty TCOs

Spekular:
If/when this becomes the basis for an upgrade routine that removes BB tracks, I think we should remember that even empty tracks can contain user data. If there are tracks in the shared BB trackset that aren't actually used, it'd be tempting to remove them in the conversion. However, that would risk deleting an instrument that the user has configured but not yet used (sample with effects on it, synth with a specific sound designed, etc.).

IanCaio:
> Good job. Think it would be better if it skipped adding empty TCOs

Thanks! That's true, I think the way the BB editor works, even tracks with no step notes have an empty TCO so they occupy the right position, haven't thought of skipping them. Should be as easy as adding `Pattern::empty()` to the conditional.

> If/when this becomes the basis for an upgrade routine that removes BB tracks, I think we should remember that even empty tracks can contain user data. If there are tracks in the shared BB trackset that aren't actually used, it'd be tempting to remove them in the conversion. However, that would risk deleting an instrument that the user has configured but not yet used (sample with effects on it, synth with a specific sound designed, etc.).

For sure, tracks should be kept even if not used. I'll just check for the empty TCOs, as those don't hold any information afaik.

IanCaio:
Done! Empty patterns are now skipped during conversion.

IanCaio:
Conflicts from file splitting also fixed on this one!

src/gui/tracks/TrackOperationsWidget.cpp
```diff
@@ -348,6 +350,16 @@ void TrackOperationsWidget::updateMenu()
 		toMenu->addAction( tr( "Turn all recording off" ), this, SLOT(recordingOff()));
 	}
 
+	PatternTrackView* patternTrackView = dynamic_cast<PatternTrackView *>(m_trackView);
+	if (patternTrackView)
+	{
+		int patternIndex = patternTrackView->getPatternTrack()->patternIndex();
+		toMenu->addSeparator();
+		toMenu->addAction(tr("Convert PatternTrack to Song Editor"), [patternIndex](){
```
zonkmachine, line 358:
```
](){
```
Block braces on separate line.
```
]()
{
```

src/core/Song.cpp
```diff
@@ -813,6 +813,83 @@ void Song::addPatternTrack()
...
+	TrackList tl = patternStore->tracks();
+
+	// For each track, go through all clips and copy them in every place
+	// the corresponding PatternTrack has PatternClips in.
+	for
+	(
+		TrackList::iterator it = tl.begin();
+		it != tl.end();
+		++it
+	)
```
messmerd, line 835:
```suggestion
	for (Track* track : tl)
```
`TrackList` is a type alias for `std::vector<Track*>`, so you can use a range-based for loop here.

Each instance of `(*it)->` below would then need to be changed to `track->`.

src/core/Song.cpp
```diff
@@ -813,6 +813,83 @@ void Song::addPatternTrack()
...
+
+			// Get the Clip from the track inside the PatternTrack
+			Clip* src = (*it)->getClip(i);
+
+			// If Clip is a MidiClip and it's empty, skip it
+			MidiClip* mc = dynamic_cast<MidiClip *>(src);
+			if (mc && mc->empty()) { continue; }
+
+			// Go through the PatternTrack's Clips
+			Track::clipVector patternclips = pt->getClips();
```
messmerd, line 871:
```suggestion
			const auto& patternclips = pt->getClips();
```
`getClips()` returns a const reference, so you'll want to store it as a const reference to avoid an unnecessary copy of the entire clips vector. This is due to the `QVector` to `std::vector` refactor which happened recently.

src/gui/tracks/PatternTrackView.cpp
```diff
@@ -78,5 +84,65 @@ void PatternTrackView::clickedTrackLabel()
...
+{
+	// Assign all tracks to channel:
+	TrackContainer::TrackList tl = Engine::patternStore()->tracks();
+
+	for
+	(
+		TrackContainer::TrackList::iterator it = tl.begin();
+		it != tl.end();
+		++it
+	)
```
messmerd, line 132:
```suggestion
	for (Track* track : tl)
```
Instances of `(*it)->` would need to be changed to `track->`, and `*it` changed to `track`.

src/gui/tracks/PatternTrackView.cpp
```diff
@@ -78,5 +84,65 @@ void PatternTrackView::clickedTrackLabel()
...
+	channel->m_name = getPatternTrack()->name();
+	if (getTrack()->useColor()) { channel->setColor(getTrack()->color()); }
+
+	assignMixerLine(channelIndex);
+}
+
+void PatternTrackView::assignMixerLine(int channelIndex)
+{
+	// Assign all tracks to channel:
+	TrackContainer::TrackList tl = Engine::patternStore()->tracks();
```
messmerd, line 125:
```suggestion
	const auto& tl = Engine::patternStore()->tracks();
```

messmerd (commented):
I'll test this PR later this week

src/gui/tracks/TrackOperationsWidget.cpp
DomClark: I personally find this style doesn't work too well with lambdas. If you don't indent the braces, you get misleading indentation, where it's not obvious that the block is part of the statement on the preceding line:
```cpp
toMenu->addAction(tr("Convert PatternTrack to Song Editor"), [patternIndex]()
{
	Engine::getSong()->convertPatterntoSE(true, patternIndex);
}
);
```
If you do indent them, then the lambda body ends up double-indented, which starts to eat up horizontal space and compounds quite quickly when you have nested lambdas:
```cpp
toMenu->addAction(tr("Convert PatternTrack to Song Editor"), [patternIndex]()
	{
		Engine::getSong()->convertPatterntoSE(true, patternIndex);
	}
);
```
Also, it feels odd to have such separation between the lambda introduction/header (I'm not sure what the technical term is) and the opening brace, which is something you don't get with a normal function definition. That could be addressed with something like the following:
```cpp
toMenu->addAction(tr("Convert PatternTrack to Song Editor"),
	[patternIndex]()
	{
		Engine::getSong()->convertPatterntoSE(true, patternIndex);
	}
);
```
(It's a discussion better suited for elsewhere, but I don't like putting braces on their own lines in general within function bodies. I think it wastes vertical space for little increase in clarity, and having code actually visible on the screen is the number-one contribution to readability. But this is one place where I think it doesn't work in particular, for the additional reasons listed above.)

messmerd:
The 1st feature - assigning all tracks in a BB Track to a FX channel - seems to work fine. I didn't see any issues while testing it.

But for the 2nd feature - converting BB Tracks to regular tracks - it looks like it doesn't convert automation tracks. And unfortunately, the zero length notes from the BB tracks aren't displayed in the Song Editor clips, so all the clips are grayed out. Fixing that is probably out of the scope of this PR though.

DomClark:
> unfortunately, the zero length notes from the BB tracks aren't displayed in the Song Editor clips, so all the clips are grayed out.

This is #5339, which will be fixed by #5902.

IanCaio:
> But for the 2nd feature - converting BB Tracks to regular tracks - it looks like it doesn't convert automation tracks. And unfortunately, the zero length notes from the BB tracks aren't displayed in the Song Editor clips, so all the clips are grayed out. Fixing that is probably out of the scope of this PR though.

I was going to upload the requested changes tonight, but I noticed the feature of converting BBTracks to the song editor is broken (it copies the tracks, but fails to copy the clips). That seems to be the case even before I made your changes, so I need to figure out why. Could it be you tested a build from an earlier commit?

IanCaio:
Haven't forgotten about your review @messmerd , time has been very limited so I focused what I had on #5924 but it seems to be finished. Now I have to figure out what is causing this bug I noticed, so I can fix it and then apply your review suggestions!

IanCaio:
After all there was no bug, it was just me being dumb and not adding BB clips on the song editor before converting, so the converted tracks would have no pattern clips. My bad 😅

Only thing I found was that on this line the Track type enum was being accessed without specifying the enum name (seemed to be working, but I changed it to use the full enum name `Track::TrackTypes::`).
```cpp
if ((*it)->type() == Track::InstrumentTrack || (*it)->type() == Track::SampleTrack)
```

Also applied @messmerd review suggestions!

zonkmachine:
@IanCaio Maybe the Pattern Editior should be muted after the patterns have been converted to tracks in the Song Editor? I see no use case for me where I would not mute the Pattern Editor manually after a conversion.

zonkmachine:
Automation tracks don't seem to convert from Pattern Track at all currently.

IanCaio:
> @IanCaio Maybe the Pattern Editior should be muted after the patterns have been converted to tracks in the Song Editor? I see no use case for me where I would not mute the Pattern Editor manually after a conversion.

That can be done, I'm just trying to figure out the details of how, forgot lots of methods so I'm having to hunt the code base to find what I need. I have this so far:

```cpp
void Song::muteAllPatternTracks()
{
	// Get a list of all tracks
	const TrackList& tracks = this->tracks();
	// Go through them and mute the pattern tracks only
	for (Track* t : tracks)
	{
		if (t->type() == Track::TrackTypes::PatternTrack)
		{
			t->getMutedModel()->setValue(false);
			t->dataChanged();
		}
	}
}
```

But it seems to be cycling through more tracks than there is in the song editor (maybe it's going through the global automation track and the tracks inside the pattern editor as well?) and it's also not muting the pattern tracks. I'm not sure the right way to mute it is to set the model manually like `t->getMutedModel()->setValue(false);`, but I didn't find a method that toggles mute, just toggleSolo.

> Automation tracks don't seem to convert from Pattern Track at all currently.

Yeah, for now it only converts sample tracks and midi tracks. I never used automation clips inside the BBEditor so I forgot about it. I'll try to change it so it copies them too.

zonkmachine:
> Yeah, for now it only converts sample tracks and midi tracks. I never used automation clips inside the BBEditor so I forgot about it. I'll try to change it so it copies them too.

OK. I think it's better to get it merged and then we (and with 'we' I mean you) can have a look at it at a later point. ;)

IanCaio:
> > Yeah, for now it only converts sample tracks and midi tracks. I never used automation clips inside the BBEditor so I forgot about it. I'll try to change it so it copies them too.
> 
> OK. I think it's better to get it merged and then we (and with 'we' I mean you) can have a look at it at a later point. ;)

That one might be as simple as changing this line:

```cpp
        if (track->type() == Track::TrackTypes::InstrumentTrack || track->type() == Track::TrackTypes::SampleTrack)
```

To this:

```cpp
        if (track->type() == Track::TrackTypes::InstrumentTrack || track->type() == Track::TrackTypes::SampleTrack || track->type() == Track::TrackTypes::AutomationTrack)
```

I'd just have to test it to make sure. I was just trying to figure out that mute method, it doesn't work and I'm not sure why. If anyone has a clue, should be something simple, I just been a bit away from coding for a while so I forgot a lot about the codebase.

zonkmachine:
> I'm not sure the right way to mute it is to set the model manually like `t->getMutedModel()->setValue(false);`...

Are you sure it isn't `t->getMutedModel()->setValue(true);` ? I can't test that code without a complete implementation.




> To this:

 ```c++
         if (track->type() == Track::TrackTypes::InstrumentTrack || track->type() == Track::TrackTypes::SampleTrack || track->type() == Track::TrackTypes::AutomationTrack)
 ```
> 
> I'd just have to test it to make sure. I was just trying to figure out that mute method, it doesn't work and I'm not sure why. If anyone has a clue, should be something simple, I just been a bit away from coding for a while so I forgot a lot about the codebase.

Tested. Works like a charm with automation hooked up correctly to the converted tracks.

IanCaio:
> > I'm not sure the right way to mute it is to set the model manually like `t->getMutedModel()->setValue(false);`...
> 
> Are you sure it isn't `t->getMutedModel()->setValue(true);` ? I can't test that code without a complete implementation.
> 

Thanks! That was exactly the issue, I had changed it and made a commit yesterday but I ran into trouble when I merged master. After merging master, when I try to run `git submodule update --recursive` I get this:
```bash
error: submodule git dir '/home/lmms/.git/modules/plugins/Sid/resid/resid' is inside git dir '/home/lmms/.git/modules/plugins/Sid/resid'
fatal: refusing to create/use '/home/lmms/.git/modules/plugins/Sid/resid/resid' in another submodule's git dir
Failed to clone 'plugins/Sid/resid/resid'. Retry scheduled
error: submodule git dir '/home/lmms/.git/modules/plugins/Sid/resid/resid' is inside git dir '/home/lmms/.git/modules/plugins/Sid/resid'
fatal: refusing to create/use '/home/lmms/.git/modules/plugins/Sid/resid/resid' in another submodule's git dir
Failed to clone 'plugins/Sid/resid/resid' a second time, aborting
```

So now I'm trying to figure that out 😅

> > To this:
> 
> ```c++
>         if (track->type() == Track::TrackTypes::InstrumentTrack || track->type() == Track::TrackTypes::SampleTrack || track->type() == Track::TrackTypes::AutomationTrack)
> ```
> 
> > I'd just have to test it to make sure. I was just trying to figure out that mute method, it doesn't work and I'm not sure why. If anyone has a clue, should be something simple, I just been a bit away from coding for a while so I forgot a lot about the codebase.
> 
> Tested. Works like a charm with automation hooked up correctly to the converted tracks.

Awesome! Then I'll change it on the next commit as well :smile:

zonkmachine:
To update past the Sid module changes you do the following.

Remove the directories:
```
/lmms/plugins/Sid/resid
/lmms/.git/modules/plugins/Sid
```
Then you run the update.

zonkmachine:
> one behavior that people should be aware of is that if they have a BB clip that is smaller than the BB track length, it will ignore that clip when converting. I.e.: if inside the BB editor they have a instrument patterns being 1 beat long and a automation track that is 4 beats long, the BB track length is 4 beats. If you then add a BB clip on the song editor that is 1 or 2 beats long, it will be smaller than the BB track length and will be ignored during conversion. When you click on the song editor to add BB clips it seems to automatically add clips with the minimum length so that will only happen if someone manually cuts the size of the BB clip down and the behavior is kind of logical (only converting clips with the minimum size of a BB clip), just thought I should mention it.

OK with this behavior. Is implementing exporting these shorter tracks depending on instrument tracks not being resizable?

IanCaio:
> OK with this behavior. Is implementing exporting these shorter tracks depending on instrument tracks not being resizable?

I've to think, right now this is how the number of clips copied is calculated:

```cpp
				auto pos = patternclip->startPosition();
				auto patternTcoLen = patternclip->length().getTicks();
				auto patternTrackLen = patternStore->lengthOfPattern(i) * TimePos::ticksPerBar();
				auto numOfCopies = patternTcoLen / patternTrackLen;
```

Meaning it simply get the length of the BB TCO, divides by the length of the BB pattern, and that's how many times it will copy the clips from each instrument inside the pattern. The problem is that if the user decides to use an automation that is longer than the beat length of the instruments on the pattern (say they are 1 beat long and the user adds an automation 4 beats long, who the hell knows why someone would do that but..) then the pattern length is the highest length, hence the automation length. Then it will only copy the clips on the smallest round number of that division. This behaves a bit differently than the song editor because it will still play a BB TCO that is smaller than that pattern length, so I might actually need to change it :/

Maybe if instead of dividing the `patternTcoLen` by the `patternTrackLen` I can divide each individual clips length by the `patternTcoLen` instead. Something like:

```cpp
				auto pos = patternclip->startPosition();
				auto clipLen = src->length().getTicks();
				auto patternClipLen = patternclip->length().getTicks();
				auto numOfCopies = clipLen / patternClipLen;
```

Still have to test that though

IanCaio:
I think I fixed the pasting behavior for when the pattern clip isn't a perfect multiple of the pattern length, by checking if the current copied clip would fit the remainder:

```cpp
				auto pos = patternclip->startPosition();
				auto clipLen = src->length().getTicks();
				auto patternClipLen = patternclip->length().getTicks();
				auto patternTrackLen = patternStore->lengthOfPattern(i) * TimePos::ticksPerBar();
				auto numOfCopies = patternClipLen / patternTrackLen;
				// If we have an pattern clip that isn't a perfect multiple of the pattern size
				// we check if the clip that will be copied would fit the remainder of the pattern clip
				if (patternClipLen % patternTrackLen >= clipLen)
				{
					++numOfCopies;
				}
```

If another pair of eyes can check that logic and test it further (seemed to work fine here) it would be appreciated!

I did run into another issue though (down the rabbit hole..): When using automation tracks to automate the tracks inside the pattern editor, when they are converted to the song editor they will still be pointing to the parameters of the tracks inside the pattern editor and not the recently converted ones. I'm not sure how to fix that one though, cause afair automation clips are linked to parameters through an ID, and when the tracks are cloned to the song editor, their parameters IDs will be changed cause they'd conflict with the ones from the pattern editor. Right now I can't think of a solution for that (or at least a simple one).

IanCaio:
Well, I didn't come up with a solution for the automation track connection issue. I thought about deleting the tracks from the BBEditor before cloning them to the Song Editor, so the cloned track have the connection IDs available, but so far I didn't manage to make it work, plus it sounds a bit hacky.

The issue itself is not a bug, simply a behavior that could be improved (the automation tracks are cloned perfectly, but if they are connected to tracks on the BBEditor, they'll still be connected to them, so the user will have to manually connect them to the parameters of the cloned track), so personally I think this can still be reviewed and merged. The behavior can be improved in the future if we can figure out a way to do so. If you all agree, then this is ready to a final review.

PR #5953, opened 2021-03-16
Better colors for pianoroll (CSS)

RoxasKH:
Well, new version of #5862 with proper changes.

This is how default and classic theme now looks with the new colors.

Default:
default

Classic:
classic


RebeccaDeField:
@RoxasKH design perspective thumbs up from me :+1:

Umcaruje:
Thumbs up from me too 👍

RoxasKH:
Thanks guys, it's just a color change but it makes the difference compared to the plain black and white of the original code.

If there hasn't been any addition to the css from when i made this new pr, imo you can merge it, it's literally a couple of hex changes in the style files.

PR #5956, opened 2021-03-19
Small bug fix to the StereoEnhancer effect and Gain knob

IanCaio:
Labeling it as `after-refactor`, though it has a small bug fix and should be a really quick one to review. Only affects the StereoEnhancer plugin.

- Fixes a small bug where the buffer was never cleared because the conditional to clearing it was never reached.

- Improves the clearing conditional so it only clears the buffer if it was actually affected in the previous process cycle.

- Adds a Gain knob to be able to adjust the level of the output signal so a more appropriate comparison of the before/after sound is done.


Spekular:
Since `clear` is both a verb and an adjective, the variable name is a bit confusing. Can we think of something unambiguous instead?
```
bufferClear, bufferCleared, bufferIsClear
shouldClearBuffer, mustClearBuffer
filledBuffer, fillBuffer
```

Veratil:
> 
> 
> Since `clear` is both a verb and an adjective, the variable name is a bit confusing. Can we think of something unambiguous instead?
> 
> ```
> bufferClear, bufferCleared, bufferIsClear
> shouldClearBuffer, mustClearBuffer
> filledBuffer, fillBuffer
> ```

Would `empty` be clearer? (no pun intended lol)

IanCaio:
Sure, now that I read it again name is a bit ambiguous. Everyone fine with `m_emptyBuffer`?

Spekular:
I skipped that one intentionally because empty is another example of a verb and adjective, so it's no different from clear.

IanCaio:
On second thought, personally I think since the prefix and lack of parenthesis makes clear it's a variable it isn't that confusing. But I went and changed it to `m_isBufferClear`, which looks good too IMO. I also forgot to set it to true on the clear method, it's fixed now.

Veratil:
> 
> 
> On second thought, personally I think since the prefix and lack of parenthesis makes clear it's a variable it isn't that confusing. But I went and changed it to `m_isBufferClear`, which looks good too IMO. I also forgot to set it to true on the clear method, it's fixed now.

I think `bufferIsClear` reads better than `isBufferClear`. `if (m_isBufferClear)` reads as `if is buffer clear` while `if (m_bufferIsClear)` reads as `if buffer is clear`.

Spekular:
> the prefix and lack of parenthesis makes clear it's a variable

I didn't mean that it would be confused for a method, I meant that the purpose of the variable could be misunderstood (inverted)
- clearBuffer = true, as in "the buffer is clear"
- clearBuffer = true, as in "we should/will clear the buffer (because it isn't clear right now)"

+1 for Veratil's point about the naming in if-statements.

IanCaio:
> I didn't mean that it would be confused for a method, I meant that the purpose of the variable could be misunderstood (inverted)
> 
>     * clearBuffer = true, as in "the buffer is clear"
> 
>     * clearBuffer = true, as in "we should/will clear the buffer (because it isn't clear right now)"

Oh, I see it now! You're right, it can be very misleading.

I changed it again to Veratil's suggestion!

Veratil:
To the same point of readability, is buffer `clear` or is buffer `empty` a better read?

IanCaio:
I think either is fine, but `clear` has the advantage of being consistent with the method name.

PhysSong:
Why you didn't implement loading/saving `m_outputGain`?

IanCaio:
> Why you didn't implement loading/saving `m_outputGain`?

Oh, I think I forgot it :sweat_smile: 

I'll add it and push it later today

IanCaio:
Now outputGain is loaded and saved on the project file!

zonkmachine:
Looks like it's pretty much done from judging by the comments. A quick rebase/merge?
@IanCaio Are you still around? :whale:

zonkmachine:
I resolved a merge conflict via the web-editor. It seem to have failed to merge correctly anyway and the last manual change should fix that.

messmerd:
The buffer clear bug was fixed in #8070, so this PR could focus on just the gain knob now.

JohannesLorenz:
As @IanCaio left LMMS, I propose to take over the PR and care about the knob (obviously this is "after feature freeze", but can yet be prepared). Everyone OK with this?

PR #5957, opened 2021-03-19
Preserve FX channel when replacing track

### LINKED ISSUE #5399: Preserve the FX-Track when an instrument gets replaced
### Enhancement Summary

When an instrument gets replaced (through drag and dropping), it'd be great if the FX-Track that was assigned to the previous instrument would not switch back to the master track (as soon as the instrument gets replaced with a drag and drop).

#### Justification

This would probably make a lot more sense, considering that when you replace an instrument, certain other properties are not affected (e.g. the piano roll and the MIDI configurations).

##### Potential use case

Say that a user decides to try out different musical instruments one-by-one and wants to use the same mixing settings every time.

Spending time with opening the instruments every single time the instrument is changed can be very annoying and does hinder productivity. This particularly applies to genres where mixing settings that essentially change the core aspects of a specific sound. (e.g. dream pop)

Spekular:
Closes #5399

(As far as my testing has shown)

### DIFF
--- a/src/tracks/InstrumentTrack.cpp
+++ b/src/tracks/InstrumentTrack.cpp
@@ -938,6 +938,21 @@ void InstrumentTrack::setPreviewMode( const bool value )
 
 
 
+void InstrumentTrack::replaceInstrument(DataFile dataFile)
+{
+	int effectChannel = effectChannelModel()->value();
+
+	InstrumentTrack::removeMidiPortNode(dataFile);
+	setSimpleSerializing();
+	loadSettings(dataFile.content().toElement());
+	
+	m_effectChannelModel.setValue(effectChannel);
+	Engine::getSong()->setModified();
+}
+
+
+
+
 QString InstrumentTrack::getSavedInstrumentName(const QDomElement &thisElement) const
 {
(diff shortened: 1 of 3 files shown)


src/tracks/InstrumentTrack.cpp
```diff
@@ -938,6 +938,21 @@ void InstrumentTrack::setPreviewMode( const bool value )
 
 
 
+void InstrumentTrack::replaceInstrument(DataFile dataFile)
+{
+	int effectChannel = effectChannelModel()->value();
```
Veratil, line 943: Would suggest adding a small comment explaining why we're doing this. Maybe something like:
`// save FX channel so it doesn't get overwritten by loadSettings`

Spekular:
```suggestion
	// loadSettings clears the FX channel, so we save it here and set it back later
	int effectChannel = effectChannelModel()->value();
```

Spekular: Good call! Fixed :)

Veratil (approved):

PR #5965, opened 2021-03-26
Fix PianoRoll m_positionLine misalignment on zoom

### LINKED ISSUE #5962: 1.3 Alpha playhead behavior bug in pianoroll
When I pan the pianoroll window horizonally, the playhead line will move too, and then it will not align with the triangle cursor on top. I am using lmms-1.3.0-alpha.1.33+g4f74151f0-mingw-win64 version in win10

Veratil:
Fixes #5962 

Adds comments to elaborate on what is being calculated.

### DIFF
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -4444,11 +4444,16 @@ void PianoRoll::updatePosition( const TimePos & t )
 	{
 		autoScroll( t );
 	}
-	const int pos = m_timeLine->pos() * m_ppb / TimePos::ticksPerBar();
-	if (pos >= m_currentPosition && pos <= m_currentPosition + width() - m_whiteKeyWidth)
+	// ticks relative to m_currentPosition
+	// < 0 = outside viewport left
+	// > width = outside viewport right
+	const int pos = (static_cast<int>(m_timeLine->pos()) - m_currentPosition) * m_ppb / TimePos::ticksPerBar();
+	// if pos is within visible range, show it
+	if (pos >= 0 && pos <= width() - m_whiteKeyWidth)
 	{
 		m_positionLine->show();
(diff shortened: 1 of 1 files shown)


src/gui/editors/PianoRoll.cpp
```diff
@@ -4444,11 +4444,16 @@ void PianoRoll::updatePosition( const TimePos & t )
 	{
 		autoScroll( t );
 	}
-	const int pos = m_timeLine->pos() * m_ppb / TimePos::ticksPerBar();
-	if (pos >= m_currentPosition && pos <= m_currentPosition + width() - m_whiteKeyWidth)
+	// ticks relative to m_currentPosition
+	// < 0 = outside viewport left
+	// > width = outside viewport right
```
IanCaio, line 4449:
```suggestion
	// > (width - piano keys width) = outside viewport right
```

src/gui/editors/PianoRoll.cpp
```diff
@@ -4444,11 +4444,16 @@ void PianoRoll::updatePosition( const TimePos & t )
...
+	const int pos = (static_cast<int>(m_timeLine->pos()) - m_currentPosition) * m_ppb / TimePos::ticksPerBar();
```
IanCaio, line 4450: Not related to your PR, but I just stopped to think we really need a `pixelsPerTick` method on the editors that currently have `m_ppb`. We use `m_ppb / TimePos::ticksPerBar()` way too often :smile:

IanCaio (approved):
LGTM! Suggested a small fix on the comment about the width

superpaik:
It look good to me (windows 10)
I just noticed (not strictly related to this PR) that when playhead is not at pos 0 resizing or increasing vertical zoom does not break pianoroll.
Maybe the solution to this PR helps with #5863 (I'll write this on that PR, though)

Feihei:
There's no problem in the win64 build now, I am using win10. Thanks.

Veratil:
Merging tomorrow unless anyone objects.

PR #5970, opened 2021-04-03
Show detailed CPU load information in a tool-tip

he29-net:
Hi,

this PR makes a few changes to `MixerProfiler` and adds a tool-tip with more detailed CPU load analysis to the `CPUGraphWidget`.

profiler2

I was quite surprised to find out how much load comes just from mixing if a lot of buses are used (try it on Jousboxx - Buzzer Beater, or Greippi - Krem Kaakkuja (on the screen above)).

Apart from the detailed load analysis, the less visible changes are:
 - add comments and rewrite the computation of CPU utilization to be easier to understand;
 - replace floats with integers where high precision is not necessary;
 - change `m_cpuLoad` from integer to float to prevent rounding errors in averaging, e.g.:
	- you start the playback and load goes from 0 to 5 %,
	- the new load is computed as `(5*0.1f + 0*0.9f) = 0.5`,
	- saving the updated load as `int` means it stays zero and the next frame the same situation repeats,
	- so the CPU graph first moves only after load reaches 5 or 10 % (not sure if `qBound<int>` did rounding when converting from `float`);
 - remove Qt dependencies from the profiler (since Qt has no business being in the core):
	- `QFile` replaced by `ofstream`,
	- `qBound` replaced by `std::min` (`std::clamp` needs C++17 and was not really needed).


he29-net:
Seems like that one weird error in `linux.gcc` test is a compiler bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=65815

It was fixed 6 years ago, which probably means we should _really_ upgrade the CI images as soon as it becomes possible.. :))

JohannesLorenz:
FWIR, we will upgrade the CI in May, and will also switch to C++17 then.

However, the new version where our CI will run will be 18.04, which just causes your issue with brace-enclosed initializers.

src/core/MixerProfiler.cpp
```diff
@@ -36,28 +40,38 @@ MixerProfiler::MixerProfiler() :
...
 
 
 
-void MixerProfiler::setOutputFile( const QString& outputFile )
+void MixerProfiler::setOutputFile(const std::string &outputFile)
 {
 	m_outputFile.close();
-	m_outputFile.setFileName( outputFile );
-	m_outputFile.open( QFile::WriteOnly | QFile::Truncate );
+	m_outputFile.open(outputFile, std::ios::trunc);
```
PhysSong, line 75: This might break writing to non-ASCII paths on Windows, FYI.

he29-net: Oh, thanks for letting me know. I thought everything is UTF-8 these days and did not anticipate any problems. If I can't figure out what to do about it, I will just roll back the change keep it on `QFile`.

he29-net:
> However, the new version where our CI will run will be 18.04, which just causes your issue with brace-enclosed initializers.

@JohannesLorenz From what I can see, Ubuntu 18.04 uses GCC 7.4.0 as a default, and the bug was milestoned to be fixed in GCC 6.x, so that means it should be fine. Or am I missing something?

JohannesLorenz:
> Ubuntu 18.04 uses GCC 7.4.0 as a default

That might be true, but looking into the CMake output of the Circle CI, I see Ubuntu 18.04 with GCC 5.4.0 :joy: 

The script runs cmake:
```
#!/bin/bash -eo pipefail
source /opt/qt5*/bin/qt5*-env.sh || true
mkdir build && cd build
cmake .. $CMAKE_OPTS -DCMAKE_INSTALL_PREFIX=./install
```
and CMake outputs:
```
-- The C compiler identification is GNU 5.4.0
-- The CXX compiler identification is GNU 5.4.0
```

he29-net:
> That might be true, but looking into the CMake output of the Circle CI, I see Ubuntu 18.04 with GCC 5.4.0 :joy:

:facepalm: That is sad.. I guess that could also interfere with the planed switch to C++17, since based on https://gcc.gnu.org/projects/cxx-status.html#cxx17 most of the C++17 features are only supported from GCC versions 6 and 7.

JohannesLorenz:
I think we should update the GCC to what is default on Ubuntu 18. This line:

https://github.com/LMMS/lmms/blob/2f17c0da2b20ecb9b00f8dcb227d02731325796b/.circleci/config.yml#L124

should be changed to `18.04` (the [docker image is already available](https://hub.docker.com/r/lmmsci/linux.gcc/tags?page=1&ordering=last_updated)), since linux.gcc is running on 18.04 now. @PhysSong do you agree with that change?

PhysSong:
If I remember correctly, `linuxdeployqt` refuses to create AppImages on 18.04(https://github.com/probonopd/linuxdeployqt/issues/340), and then we downgraded to 16.04.

JohannesLorenz:
> If I remember correctly, `linuxdeployqt` refuses to create AppImages on 18.04([probonopd/linuxdeployqt#340](https://github.com/probonopd/linuxdeployqt/issues/340)), and then we downgraded to 16.04.

Almost everything from C++17 is not supported in GCC 5. So does this mean we can not use C++17 until the end of 18.04 (of yes, when will we stop supporting 18.04)? (sorry for OT @he29-net)

src/core/MixerProfiler.cpp
he29-net: It turns out the issue is even more painful than I imagined, so I returned the old `QFile` code.

TL;RD, if I understand it correctly, Windows uses UTF-16 for file paths, but
 a) the only standard C++ way of character encoding conversion is somewhat hard to understand and already deprecated in C++17, and 
 b) even if it was easy to do, it is probably not safe to assume the conversion will always be UTF-8 to UTF-16 (what if the command line uses SHIFT-JIS or something?), and I'm not even sure if it wouldn't break _other_ platforms then.

So letting Qt handle all that file path encoding stuff seems like a better idea, at least Qt authors probably have an idea what are they doing, unlike me. :)

include/MixerProfiler.h
```diff
@@ -41,21 +42,34 @@ class MixerProfiler
...
 
 private:
 	MicroTimer m_periodTimer;
-	int m_cpuLoad;
+	float m_cpuLoad;
 	QFile m_outputFile;
 
+	static const int s_detailCount = 4;                     // set to the actual number of used probes in Mixer.cpp
+	std::array<MicroTimer, s_detailCount> m_detailTimer;    // use arrays to avoid dynamic allocations in realtime code
+#ifdef __GNUC__
```
Reflexe, line 66: Maybe we better to use this workaround on newer GCC  too? I don't find it that harmful and having the same code may have negative consequences on the long term when compiling on different platforms (e.g. VC or clang)

include/MixerProfiler.h
```diff
@@ -41,21 +42,34 @@ class MixerProfiler
...
+
+	int detailLoad(const unsigned int index) const { return m_detailLoad[index]; }
 
 private:
 	MicroTimer m_periodTimer;
-	int m_cpuLoad;
+	float m_cpuLoad;
 	QFile m_outputFile;
 
+	static const int s_detailCount = 4;                     // set to the actual number of used probes in Mixer.cpp
```
Reflexe, line 64: If I understand correctly, this number, including the meaning of each index is tightly coupled with the implementation on the producer Mixer and consumer CPULoadWidget. 

In order to avoid confusions when only changing one part, I would suggest to be a bit more explicit. For example: 
* use a struct with getters:
```
struct CpuStatistics {
 uint32_t getTotalCpu();
 struct Stat {
     MicroTimer timer;
     detailLoad load; 
  } ;
  Stat mixer
  Stat xyz;
};

```

* have explicit constants for each index:


```
enum class CpuLoadTimer {
    Mixer,
    NumberOfTimers
};

std::array<Stat, NumberOfTimers> stats;

// Producer 
stats[CpuLoadTimer::Mixer] = xyz;

// Consumer
AddStat("Mixer", stats[CpuLoadTimer::Mixer] );
```

include/MixerProfiler.h
he29-net: I don't quite understand what are trying to say in the second sentence; do you suggest to keep the workaround even after update to a newer GCC version in our CI (to preserve compatibility with order compilers over long term), or to use the same workaround code for all compilers, not just GCC (since, assuming it works, it would make the code simpler)?

include/MixerProfiler.h
he29-net: I originally used a slightly more complex system, where you would first initialize each probe in the Mixer constructor with a string that describes what it measures, but eventually I removed that since I was worried about accidentally introducing dynamic allocations to the core when working with `QString` (which is what inevitably comes out of the `tr()` translation thing if I'm not mistaken). So I moved the strings to the GUI portions, since it wouldn't be changed often (if ever).

Using an enum instead of a magic constant is a good idea, thanks. I also thought about separate variables or a structure before, but I wanted to keep it in array so that I can process all detail stats in a single loop, avoiding lots of redundancy.

include/MixerProfiler.h
Reflexe: I suggest to use the same code everywhere and avoid code duplication. In this case, I would suggest to use this workaround even if the compiler is not gcc. 
Consider the case where one will only update for gcc and then other compilers will break.

he29-net: OK, makes sense. I will keep a comment that explains the workaround, though. Otherwise someone could one day come around and "make it prettier" by removing the extra braces. :)

Reflexe (approved):
LGTM

he29-net:
Updated to master branch after refactor. (Fun stuff, git did not understand the `Mixer.cpp` → `AudioEngine.cpp` rename and decided to mix the changes from old `Mixer.cpp` into random places of the completely unrelated _new_ `Mixer.cpp`...)

src/gui/widgets/CPULoadWidget.cpp
```diff
@@ -91,15 +90,29 @@ void CPULoadWidget::paintEvent( QPaintEvent *  )
...
-	if( new_load != m_currentLoad )
+	// Additional display smoothing for the main load-value. Stronger averaging
+	// cannot be used directly in the profiler: cpuLoad() must react fast enough
+	// to be useful as overload indicator in AudioEngine::criticalXRuns().
+	const int new_load = (m_currentLoad + Engine::audioEngine()->cpuLoad()) / 2;
+
+	if (new_load != m_currentLoad)
 	{
+		setToolTip(
+			tr("DSP total: ") + QString::number(new_load) + " %\n"
```
PhysSong, line 101: You can also do like this to avoid unnecessary manual concatenations.
```suggestion
			tr("DSP total: %1").arg(new_load) + " \n"
```

include/AudioEngineProfiler.h
```diff
@@ -44,20 +45,39 @@ class AudioEngineProfiler
...
 
 private:
 	MicroTimer m_periodTimer;
-	int m_cpuLoad;
+	float m_cpuLoad;
 	QFile m_outputFile;
+
+	// Use arrays to avoid dynamic allocations in realtime code
+	// NOTE: The double {} is a workaround for https://gcc.gnu.org/bugzilla/show_bug.cgi?id=65815
+	// (Can be removed when LMMS CI updates to GCC 6 or higher.)
```
PhysSong, line 77: We now use GCC 7, so this can probably be removed.

src/core/BandLimitedWave.cpp
```diff
@@ -26,6 +26,7 @@
 #include "BandLimitedWave.h"
 
 #include <QDataStream>
+#include <QFile>
```
PhysSong, line 29: Is this unrelated change necessary?

he29-net: I'll try to remove it; I think something changed in the build order after my changes and it did not compile due to the missing `QFile` include. But that was a long time ago and I don't know if it is or should be automatically included from the other header files.

src/core/AudioEngine.cpp
```diff
@@ -337,6 +337,7 @@ void AudioEngine::pushInputFrames( sampleFrame * _ab, const f_cnt_t _frames )
 const surroundSampleFrame * AudioEngine::renderNextBuffer()
 {
 	m_profiler.startPeriod();
+	m_profiler.startDetail(AudioEngineProfiler::DetailType::NoteSetup);
```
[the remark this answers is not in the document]
he29-net: I read what RAII pattern is and wikipedia describes it in the context of classes, not functions. I suppose you meant to simply call the `startDetail` / `finishDetail` functions at the start and end of the separated stage functions?

It could make `AudioEngine::renderNextBuffer()` easier to navigate, although I'm not sure if it's even worth splitting off stage 2 (which is just 2 + 2 lines); that would have the opposite effect and just make the code a bit harder to find. :)

src/core/BandLimitedWave.cpp
he29-net: So, while it compiled, I don't see any other obvious place from which `QFile` may be included. Maybe it is included by `QDataStream`, but I did not see anything about that in documentation, so it probably should not be assumed to be the case. So I think the new include should stay.

src/core/AudioEngine.cpp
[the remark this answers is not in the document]
he29-net: I made new functions for the four stages, it probably does look a bit better this way. If that's not what you meant and / or this is an unnecessary change, the last commit can be simply reverted.

Veratil:
@he29-net could you rebase and we can work on getting this merged? :)

he29-net:
> @he29-net could you rebase and we can work on getting this merged? :)

Are there any conflicts not shown by github? It seems fine here. But I suppose it won't hurt checking if everything is all right after the refactor.

Veratil:
> > @he29-net could you rebase and we can work on getting this merged? :)
> 
> Are there any conflicts not shown by github? It seems fine here. But I suppose it won't hurt checking if everything is all right after the refactor.

Oop, you're right. I had it on rebase from a previous merge and it was giving me the "cant do this" error. Nevermind on rebase then! 😁

he29-net:
I tested it after merging upstream/master locally and it does what it's supposed to. So I reckon it is still ready for review / merge.
@PhysSong Do you have any more concerns regarding the open comments, or should I consider them resolved?

src/core/BandLimitedWave.cpp
sakertooth: I would insist on removing it. The builds compile fine without it, and it can only add to our build times.

include/AudioEngineProfiler.h
```diff
@@ -44,20 +45,37 @@ class AudioEngineProfiler
...
 
 	int cpuLoad() const
 	{
 		return m_cpuLoad;
 	}
 
-	void setOutputFile( const QString& outputFile );
+	void setOutputFile(const QString& outputFile);
 
+	enum DetailType {
```
sakertooth, line 57: We could use ``enum class`` instead, in light of efforts made in f102777.

src/core/AudioEngine.cpp
```diff
@@ -380,53 +380,74 @@ const surroundSampleFrame * AudioEngine::renderNextBuffer()
...
+	m_profiler.finishDetail(AudioEngineProfiler::DetailType::NoteSetup);
+}
+
+
+
+void AudioEngine::renderStageInstruments()
+{
+	m_profiler.startDetail(AudioEngineProfiler::DetailType::Instruments);
+
+	AudioEngineWorkerThread::fillJobQueue<PlayHandleList>(m_playHandles);
```
sakertooth, line 400: I think we can remove the redundant template parameter ``PlayHandleList``.

sakertooth (commented):
You made quite a few out of scope format changes. I don't really mind them, but you could revert them to make the PR more focused.

src/core/BandLimitedWave.cpp
he29-net: OK, I'll get rid of it then. I suppose someone else could fix it if / when the compile order changes again and the issue returns.

he29-net:
> You made quite a few out of scope format changes. I don't really mind them, but you could revert them to make the PR more focused.

Yeah, I usually updated formatting of any code in the same block with the code that I changed, so that it does not become a wild mix of the new and old styles. But I understand that this practice is now not recommended (which is probably a good thing, as it makes reviews easier). I'll try to get rid of the extra changes.

include/AudioEngineProfiler.h
he29-net: Sure, why not. I did not even know `enum class` existed. :)

src/core/AudioEngine.cpp
sakertooth: > I assume you suggest it should derive the template type based on m_playHandles type?

:+1:

include/AudioEngineProfiler.h
he29-net: Hm, can `enum class` be still used for array indexing etc.? Because after the change I'm getting:
> error: could not convert ‘lmms::AudioEngineProfiler::DetailType::DetailCount’ from ‘lmms::AudioEngineProfiler::DetailType’ to ‘long unsigned int’

which suggests that use-case is completely unsupported..

sakertooth: > Hm, can `enum class` be still used for array indexing etc.? Because after the change I'm getting:
> 
> > error: could not convert ‘lmms::AudioEngineProfiler::DetailType::DetailCount’ from ‘lmms::AudioEngineProfiler::DetailType’ to ‘long unsigned int’
> 
> which suggests that use-case is completely unsupported..

``enum class`` does not allow implicit conversions to integral types. Try making the conversion explicit using ``static_cast``.

src/core/BandLimitedWave.cpp
DomClark: I'm happy for this change to be omitted from this PR since it is indeed out of scope, but I don't think we should remove headers like this in general. This file does use the `QFile` class, so should explicitly include it. It will have a negligible effect on compile times, since modern compilers handle include guards efficiently, and it avoids fragile header dependencies where changes to header files can cause compilation errors in apparently unrelated files.

include/AudioEngineProfiler.h
```diff
@@ -53,11 +54,31 @@ class AudioEngineProfiler
...
 private:
 	MicroTimer m_periodTimer;
-	int m_cpuLoad;
+	float m_cpuLoad;
 	QFile m_outputFile;
+
+	// Use arrays to avoid dynamic allocations in realtime code
+	std::array<MicroTimer, static_cast<int>(DetailType::DetailCount)> m_detailTimer;
+	std::array<int, static_cast<int>(DetailType::DetailCount)> m_detailTime{0};
+	std::array<float, static_cast<int>(DetailType::DetailCount)> m_detailLoad{0};
```
DomClark, line 81: It would be more appropriate to cast the size to `std::size_t`, since that is the actual template parameter type here. Also, since the `DetailCount` member is only ever going to be used as an integer, and not as an enumerator, I think it makes sense to define it as such outside the enum definition. In #6760, I standardised this pattern as the following:
> Final enumerators representing the total number of enumerators have all been renamed to `Count`. Any that were used numerically have had a `std::size_t` constant with the original enumerator name added alongside the enum.

In this case, you'd have:
```cpp
enum class DetailType {
	// ...
	Mixing,
	Count
};
constexpr static auto DetailCount = static_cast<std::size_t>(DetailType::Count);

// ...

std::array<MicroTimer, DetailCount> m_detailTimer;
```

DomClark, line 81: The elements of `m_detailLoad` are written from the audio thread and read from the UI thread, so should use atomics to avoid race conditions. You can use relaxed loads and stores for efficiency if you wish, since there are no data dependencies involved.

he29-net: Thanks, that will be much cleaner.

he29-net: Good point; I never heard of relaxed loads / stores, so I'll read up on that and make the changes tomorrow-ish.

src/core/AudioEngine.cpp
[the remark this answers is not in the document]
he29-net: I see, thanks for the explanation. I know about the concept from smart pointers and lock guards, but I never realized that this is what RAII was referring to. I'll think about it and try to apply it here, since half of the work (splitting stages into separate functions) is already done.

include/AudioEngineProfiler.h
```diff
@@ -53,11 +55,36 @@ class AudioEngineProfiler
...
+		Mixing,
+		Count
+	};
+
+	constexpr static auto DetailCount = static_cast<std::size_t>(DetailType::Count);
+
+	void startDetail(const DetailType type) { m_detailTimer[static_cast<std::size_t>(type)].reset(); }
+	void finishDetail(const DetailType type)
+	{
+		m_detailTime[static_cast<int>(type)] = m_detailTimer[static_cast<std::size_t>(type)].elapsed();
```
DomClark, line 71: Another `int` that could be `std::size_t`:
```suggestion
		m_detailTime[static_cast<std::size_t>(type)] = m_detailTimer[static_cast<std::size_t>(type)].elapsed();
```

src/core/AudioEngineProfiler.cpp
```diff
@@ -38,10 +41,23 @@ AudioEngineProfiler::AudioEngineProfiler() :
...
+	// Compute new overall CPU load and apply exponential averaging.
+	// The result is used for overload detection in AudioEngine::CriticalXRuns()
+	// → the weight of a new sample must be high enough to allow relatively fast changes!
+	const int newCpuLoad = 100 * periodElapsed / timeLimit;
+	m_cpuLoad = std::min(newCpuLoad * 0.1f + m_cpuLoad * 0.9f, 100.f);
 
-	const float newCpuLoad = periodElapsed / 10000.0f * sampleRate / framesPerPeriod;
-    m_cpuLoad = std::clamp<int>((newCpuLoad * 0.1f + m_cpuLoad * 0.9f), 0, 100);
+	// Compute detailed load analysis. Can use stronger averaging to get more stable readout.
+	for (int i = 0; i < DetailCount; i++)
```
DomClark, line 56: One more:
```suggestion
	for (std::size_t i = 0; i < DetailCount; i++)
```

include/AudioEngineProfiler.h
```diff
@@ -53,11 +55,36 @@ class AudioEngineProfiler
...
+
+	int detailLoad(const DetailType type) const
+	{
+		return m_detailLoad[static_cast<std::size_t>(type)].load(std::memory_order_relaxed);
+	}
 
 private:
 	MicroTimer m_periodTimer;
-	int m_cpuLoad;
+	float m_cpuLoad;
```
DomClark, line 81: This could also do with being atomic (also changed in #6839).

src/core/AudioEngineProfiler.cpp
```diff
@@ -38,10 +41,23 @@ AudioEngineProfiler::AudioEngineProfiler() :
...
+	{
+		const int newLoad = 100 * m_detailTime[i] / timeLimit;
+		m_detailLoad[i].store(std::min(newLoad * 0.05f + m_detailLoad[i] * 0.95f, 100.f), std::memory_order_relaxed);
```
DomClark, line 59: The old load could also be loaded with relaxed ordering:
```suggestion
		const auto oldLoad = m_detailLoad[i].load(std::memory_order_relaxed);
		m_detailLoad[i].store(std::min(newLoad * 0.05f + oldLoad * 0.95f, 100.f), std::memory_order_relaxed);
```

src/core/AudioEngineProfiler.cpp
```diff
@@ -38,10 +41,23 @@ AudioEngineProfiler::AudioEngineProfiler() :
...
-	int periodElapsed = m_periodTimer.elapsed();
+	// Time taken to process all data and fill the audio buffer.
+	const unsigned int periodElapsed = m_periodTimer.elapsed();
+	// Maximum time the processing can take before causing buffer underflow. Convert to us.
+	const uint64_t timeLimit = static_cast<uint64_t>(1000000) * framesPerPeriod / sampleRate;
+
+	// Compute new overall CPU load and apply exponential averaging.
+	// The result is used for overload detection in AudioEngine::CriticalXRuns()
+	// → the weight of a new sample must be high enough to allow relatively fast changes!
+	const int newCpuLoad = 100 * periodElapsed / timeLimit;
```
DomClark, line 52: This, and the detailed loads, may as well be computed using floating point arithmetic for extra precision, since the result is stored as a float too.

src/gui/widgets/CPULoadWidget.cpp
```diff
@@ -91,10 +91,21 @@ void CPULoadWidget::paintEvent( QPaintEvent *  )
...
+
+	if (new_load != m_currentLoad)
 	{
+		auto engine = Engine::audioEngine();
+		setToolTip(
+			tr("DSP total: %1").arg(new_load) + " %\n"
+			+ tr(" - Notes and setup: %1").arg(engine->detailLoad(AudioEngineProfiler::DetailType::NoteSetup)) + " %\n"
+			+ tr(" - Instruments: %1").arg(engine->detailLoad(AudioEngineProfiler::DetailType::Instruments)) + " %\n"
+			+ tr(" - Effects: %1").arg(engine->detailLoad(AudioEngineProfiler::DetailType::Effects)) + " %\n"
+			+ tr(" - Mixing: %1").arg(engine->detailLoad(AudioEngineProfiler::DetailType::Mixing)) + " %"
```
DomClark, line 107: I think I prefer the percent signs without a space before them, and it's what we currently do elsewhere in the program. Also, it's better to keep text in translatable strings where possible.
```suggestion
			tr("DSP total: %1%").arg(new_load) + "\n"
			+ tr(" - Notes and setup: %1%").arg(engine->detailLoad(AudioEngineProfiler::DetailType::NoteSetup)) + "\n"
			+ tr(" - Instruments: %1%").arg(engine->detailLoad(AudioEngineProfiler::DetailType::Instruments)) + "\n"
			+ tr(" - Effects: %1%").arg(engine->detailLoad(AudioEngineProfiler::DetailType::Effects)) + "\n"
			+ tr(" - Mixing: %1%").arg(engine->detailLoad(AudioEngineProfiler::DetailType::Mixing))
```

src/gui/widgets/CPULoadWidget.cpp
```diff
@@ -91,10 +91,21 @@ void CPULoadWidget::paintEvent( QPaintEvent *  )
...
 {
-	// smooth load-values a bit
-	int new_load = ( m_currentLoad + Engine::audioEngine()->cpuLoad() ) / 2;
-	if( new_load != m_currentLoad )
+	// Additional display smoothing for the main load-value. Stronger averaging
+	// cannot be used directly in the profiler: cpuLoad() must react fast enough
+	// to be useful as overload indicator in AudioEngine::criticalXRuns().
+	const int new_load = (m_currentLoad + Engine::audioEngine()->cpuLoad()) / 2;
+
+	if (new_load != m_currentLoad)
```
DomClark, line 99: This won't update the tooltip unless the overall CPU load changes, even if the detailed loads do change. This is probably okay, but I wanted to check whether this was intended (and see what others think too).

src/core/AudioEngine.cpp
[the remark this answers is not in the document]
he29-net: OK, so after a few attempts it builds and actually works, so I hope my implementation is adequate. :)

I though about also using the RAII probes for the main timer (started and stopped by `startPeriod()` and `finishPeriod()`), but since it is using a different averaging weights and it can directly influence playback[^1], I would rather not touch it and keep it as a separate special case.

[^1]: By determining the cutoff, when LMMS considers performance insufficient and stops playing new notes. I personally don't think the current cutoff detection works particularly well, but that's for another issue / PR.

src/gui/widgets/CPULoadWidget.cpp
he29-net: Oh, that's interesting. In Czech, "10 %" and "10%" are pronounced differently and mean different things. The first one just states "ten percent", as a standalone number, while the second form is used as an adjective. I thought it works the same way in English so I used the same spacing.

So good idea with moving the symbol into the translatable string, since Czech translation should definitely keep the space. :)

src/gui/widgets/CPULoadWidget.cpp
he29-net: It's not quite intentional, but probably not a big issue either, because currently Qt refuses to update the tooltip unless the mouse cursor moves. So combined with the fact that the numbers jump all around the place even with relatively high averaging, it's only good for rough estimates anyway.

src/core/AudioEngineProfiler.cpp
he29-net: It does not seem to make any difference; the bar graph still only moves in steps of 4 pixels:
https://github.com/LMMS/lmms/assets/20932751/7ddb3549-3413-49db-8024-360b966124a3

So the (admittedly tiny) performance hit would probably not serve any purpose, unless the CPU graph widget is updated to show to better resolution as well. And even after that, the `newCpuLoad` is multiplied by 0.1 on the next line, further reducing any impact the loss of precision brings.

But crunching a few floats per buffer instead of integers probably has no measurable performance impact, so I don't mind either way.

he29-net: Oh, a comment in `CPULoadWidget` reminded me: the low drawing resolution is probably intentional backward compatibility with the classic theme. But as far as I can tell, it's not working particularly well anyway:
screen

Maybe it's just because of my high DPI screen, but the drawing steps do not seem to properly align with the "LEDs", which was the sole purpose of the "rounding code". So if nobody objects, I would just get rid of it and let the load graph move smoothly.

DomClark: Even if the CPU graph precision is limited to some number of pixels, it could still be worth tracking the load more accurately behind the scene. And since we have a tooltip now, it's displayed to the nearest percentage point anyway.

As for a smoother resolution, I very much agree. I though we already had a pull request for that (possibly by @LostRobotMusic), but I can't seem to find one. If you really want to push the boat out, you could add a CSS property for the bar step size, so themes can set it appropriately for their own image.

src/gui/widgets/CPULoadWidget.cpp
DomClark: I never knew that about Czech - it's always interesting to find these things out. Looking at https://en.wikipedia.org/wiki/Percent_sign#Form_and_spacing, it appears the styling varies greatly between languages - some even put the percent sign before, not after, the number. As you say, all the more reason to make it translatable.

src/core/AudioEngineProfiler.cpp
he29-net: > could still be worth tracking the load more accurately behind the scene.

That's true; if someone decides to fix / improve things related to `AudioEngine::criticalXRuns()`, it's probably better to have numbers without unnecessary rounding as the input.

So, pushing a change to floats then. (And the smoother stepping; I also remember some old change made by Lost, but maybe it addressed something else rather than the drawing step.)

he29-net:
One more issue I noticed is that when the CPU load is high, sometimes the sum of "details" grows over 100%. It probably also happens at other times, but isn't as noticeable. My guess is that while the GUI code gathers data for the tooltip, the values change, and samples from different time are mixed together.

I'm not sure how to elegantly address that, other than by adding some buffer for the GUI side, that would be atomically updated before GUI reads the individual values. But somehow I don't like the idea of `std::atomic<std::vector>`... (I'm not even sure if that's allowed, but I feel like it shouldn't be :) ).

DomClark:
> sometimes the sum of "details" grows over 100%

It's probably just rounding error. It's a common enough occurrence with percentages that I don't think it's worth addressing (doing so could even be considered incorrect).

> `std::atomic<std::vector>`... (I'm not even sure if that's allowed, but I feel like it shouldn't be :) )

You're right, it's not. `std::atomic` only works with trivially copyable types.

he29-net:
> It's probably just rounding error.

I mean _significantly_ over 100%. Right now I tried it with a demo project and "yes" running in a terminal to load the CPU, and I got total 96%, instruments 83%, effects 14% and mixing 20%. The sum is 117 %, which does not look like just a simple rounding error. :)

src/core/AudioEngineProfiler.cpp
he29-net: > If you really want to push the boat out, you could add a CSS property for the bar step size

I took it as an opportunity to finally learn how styling and custom attributes in Qt work, and implemented the custom step size support.

While at it, I found out that the misalignment in classic theme was not caused by scaling. It persisted even with `QT_AUTO_SCREEN_SCALE_FACTOR=0`, because the old fixed code assumed LED = 2 pixels + 1 pixel space, while the classic theme uses 3+1.

So now the default theme moves smoother and the classic theme looks nicer. :)

include/CPULoadWidget.h
```diff
@@ -64,6 +65,9 @@ protected slots:
 
 	QTimer m_updateTimer;
 
+	int m_stepSize;
+	int stepSize() const { return m_stepSize > 0 ? m_stepSize : 1; }
```
DomClark, line 69: I'd go for `std::max` instead here. Also, I don't think we have a rule for it in our coding conventions, but we prefer to put all member functions before all member variables.

DomClark (commented):
Good job with the RAII wrapper and the step size property!

DomClark:
> I mean significantly over 100%. Right now I tried it with a demo project and "yes" running in a terminal to load the CPU, and I got total 96%, instruments 83%, effects 14% and mixing 20%. The sum is 117 %, which does not look like just a simple rounding error. :)

Yeah, that's not a rounding error. Perhaps it's an issue with the clamping? The total (and each detail) is limited to a maximum of 100%, so it can dip slightly below 100% even if it spends most of its time well above. The 117% sum may well be accurate in the long run. You could try limiting to 100% in the `cpuLoad` and `detailLoad` functions, while maintaining the internal average without a limit.

he29-net:
Thanks for the followup review. I see I don't write enough object-oriented code lately, since I'm missing constructors, not realizing a class declaration can be nested within another class, etc... It's a good learning experience, but I have to apologize for taking your time. :sweat_smile:

include/AudioEngineProfilerProbe.h
```diff
@@ -0,0 +1,46 @@
...
+
+class AudioEngineProfilerProbe
+{
+public:
+	AudioEngineProfilerProbe(AudioEngineProfiler& profiler, AudioEngineProfiler::DetailType type);
+	~AudioEngineProfilerProbe();
+
+private:
+	AudioEngineProfiler &m_profiler;
+	const AudioEngineProfiler::DetailType m_type;
```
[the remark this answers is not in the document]
he29-net: I did not realize the reference would affect copying etc., I just tend to avoid pointers when they are not necessary. I can change it to a pointer if you think it's a better practice.

The constant `m_type` is intentional: since the timer of a given type is started in the constructor, it should not be possible to modify the type afterwards -- to ensure that the destructor is always called on the same timer. The `const` seemed like the right way how to formalize this intention.

he29-net:
Ah, about the >100% sums: I did not realize that there is no magic watchdog that aborts the processing of current buffer when the time limit is reached. The render stages always run to completion, so it's either the OS or the audio driver / HW that determines when it really needs to stop waiting and put some signal (or silence) on the wire. So it makes sense that the timers may go over 100% during high load: the limit is not a hard wall and LMMS will start dropping notes only a few buffers later, when it notices the high load.

include/AudioEngineProfilerProbe.h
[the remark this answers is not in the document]
DomClark: If you're not changing the const, there's no point changing the reference, so it's probably best left as is, since references are indeed preferable to pointers in general where possible.

The issue is that assigning to an object requires overwriting its member variables with new values, and references and const objects cannot be written. Even if member variables should stay constant between assignments to an object, they need to be mutable to support any eventual assignment. Another impacted scenario is move construction - even if assignment is not supported, this still typically requires clearing out the source object to prevent its destructor doing what is now the responsibility of the new object.

As mentioned in the comment above, this is fine in this class as it is specifically designed to be immovable and unassignable, but usually ought to be avoided even if it would better express invariants, as it significantly impacts the usability of the containing class.

DomClark:
I think it's still worth maintaining the load averages without a limit, and only limiting the display. Otherwise the displayed average will tend to sit below its actual value, which is less useful both for the user and for internal CPU-limiting measures such as dropping notes. You could even remove the display limit (at least for the tooltip - I don't think the bar needs to start running off the edge of the screen) so users can see how far over 100% they are, if you think that sounds useful.

he29-net:
I'm a bit worried about changing anything about `m_cpuLoad`, since it will change how LMMS behaves when reaching full CPU load. But what you say makes sense; it could even be the reason why the note dropping during high CPU load does not behave all that well? 

`criticalXRuns()` currently contains:
`return cpuLoad() >= 99 && Engine::getSong()->isExporting() == false;`

The 99 seems like a pretty arbitrary constant. If the clamping is removed, maybe it'll make sense to change it to a 100, since `m_cpuLoad` won't be artificially lowered. Or, on the contrary, it may be better to start dropping notes even sooner, so as to prevent buffer underflows. Who knows. I'll try to play with that and see what happens; my laptop is sufficiently underpowered so hitting CPU limits does not take much effort. :)

qnebra:
I think dropping notes looks bad from user perspective. In my opinion much better would be to find and fix perfomance bottlenecks, thread locks and other things that would affect perfomance.

DomClark:
> I'm a bit worried about changing anything about `m_cpuLoad`

Don't worry about it - as you've observed, the note dropping doesn't seem to be particularly finely tuned, nor especially reliable. Changing the CPU limit might be out of scope here, and I would leave it slightly below 100, so we start dropping notes before, rather than after, any underruns occur.

> I think dropping notes looks bad from user perspective. In my opinion much better would be to find and fix perfomance bottlenecks, thread locks and other things that would affect perfomance.

No matter how fast everything else in LMMS gets, there will be a limit to how much it can process, since notes take non-zero time to play, and CPUs aren't infinitely fast. Your suggestions are of course good ideas, but they're not a substitute for dropping notes; they can only increase the amount that can be played before dropping is required.

qnebra:
And this is my point, to minimise note dropping.

he29-net:
I tried to compare `criticalXRuns` with 90, 99 and 110% limit, and the only one that seems significantly different to me is 90%. Both the original 99 and the 110% limit tend to result in buffer underflows first, followed by mostly silence (as the dropping code catches up and drops almost all notes). The 90% limit, on the other hand, favors dropping random notes before underflow occurs.

It's hard to decide what is objectively worse; one inserts pops, random delays and silences, the other mangles the song by dropping random notes while wasting up to 10% of performance by dropping notes too early. The load peaks are also semi-random, making it hard to a proper comparison. So keeping it at 99 is probably best for now, although something like 95 could be also a good, balanced choice.

---

As for the 100% clamping: I removed it and only added back the limit to the drawing code (just to be sure; the bitmap does not extend beyond 100% and / or the widget does not allow drawing beyond its borders). I feel like the numbers make much more sense now, and provide additional information that would be otherwise hidden. So thanks @DomClark for noticing the issue; I completely missed it even after you first mentioned it. :) And since there seems to be nothing left to address, thanks everyone involved for the reviews.

DomClark:
> So keeping it at 99 is probably best for now

Yeah, it sounds like it needs a little more tweaking and thought (and maybe even an option in the settings, much as I dislike resorting to that solution), and is probably out of scope here.

It's good to hear that the percentages work properly now, too.

DomClark (approved):
Code looks good to me now. There are some style changes that could be made, but I'll leave that up to someone else if they want to go over them before merging it.

Veratil (commented):
Style LGTM

sakertooth (approved):
The overall review by @DomClark lasted for 5 days, good on you Dom :+1:. As for the code, it LGTM. We could also make ``AudioEngineProfiler`` use RAII, but that doesn't have to be done here.

PR #5971, opened 2021-04-03
Fix/improve includes/fwd decls

JohannesLorenz:
Preparation for clang-format. It will shuffle the includes, and this
preparation will prevent errors from that step.

### DIFF
--- a/include/MidiWinMM.h
+++ b/include/MidiWinMM.h
@@ -28,8 +28,11 @@
 #include "lmmsconfig.h"
 
 #ifdef LMMS_BUILD_WIN32
-#include <windows.h>
+
+// clang-format off
+#include <windows.h> // must be included before mmsystem.h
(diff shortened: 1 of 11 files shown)


include/MidiWinMM.h
```diff
@@ -28,8 +28,11 @@
 #include "lmmsconfig.h"
 
 #ifdef LMMS_BUILD_WIN32
-#include <windows.h>
+
+// clang-format off
```
JohannesLorenz, line 32: I wonder if we should tell clang-format here: If you have windows.h, put it into a separate category above the other headers. It would work, but I wonder if it looks strange to have windows.h in a separate header group...

JohannesLorenz:
Re-working this PR, because of the above comment.

JohannesLorenz:
Closing this PR, making a new one.

PR #5979, opened 2021-04-10
Xpressive plugin fixes

gnudles:
Balanced most of the presets.
Made last function accept floating point input.
Removed "key" and "bnote" for meanwhile.
response for #5868 
-- do not merge yet --


gnudles:
This PR awaits discussion about whether to enable "key" variable in expressions or not.

gnudles:
Before merging, we should upgrade exprtk to the latest version, because expressions such "1 in t" cuases the program to crash...
Latest version seems to accept such formula without crashing.

gnudles:
@PhysSong, do you know why isn't it compiling on windows? I added syntax highlighting in Xpressive plug-in.
Could it be that the following define is true on windows (QT_NO_SYNTAXHIGHLIGHTER)?
Screenshot from 2021-12-15 00-31-47
,

PhysSong:
> Could it be that the following define is true on windows (QT_NO_SYNTAXHIGHLIGHTER)?

It isn't in MSVC builds, but it is in MinGW builds. Looks like @tobydox [disabled the feature](https://launchpadlibrarian.net/420647900/qt5base-mingw-w64_5.9.8-1.diff.gz) when building the MinGW PPA.

qnebra:
Well, majority of packages in this PPA needed upgrade. Some get a major additions and bugfixes in upstream versions since they are built in PPA. Additions and fixes which would be really useful for lmms.

gnudles:
Today I read that MSVC++ treat 'long double' the same as double. It is bit problematic, because the integrate function relies on long double, as it sums frequencies 48000 times, and my tests with plain simple double had distortion after few seconds... any ideas how to overcome this?

PhysSong:
> the integrate function relies on long double

I can't find any `long double`s in `IntegrateFunction`. I'd say the round-off error in integration of `double` is not a concern compared to the precision of `float`.
You can find some useful resources in #2047.

gnudles:
@PhysSong Yeah, you are right, my bad. somehow my memory convinced me that I was working with 'long double' there, but truth is I used double... saw #2047  and it reminded me my experiments with integrate.
By the way, I added another variable called 'prev' which stores the frequency of the previous note, so now in the formula you can make portamento effect in the normal way, or any other weird way...
Also, I enabled variables in the formula, so you can reuse code, instead of copying code all around the place.
I am considering adding a mechanism to store and get computed values between frames. 
One guy on youtube had some suggestions for the plugin which we should consider but I not sure I have the time for this right now (https://www.youtube.com/watch?v=5pJI1JpXNqM&t=25s in comments):

I do have a few suggestions : 
Add more wave tables and general-purpose knobs, maybe 32 or up to 256 of each so you practically have limitless control over the sound design.
.....(here comes some feature request that already existed)....
Finally, A UI feature to detach the wave function editor panel to its own window and save/load your functions to a txt file.

PhysSong:
I think your editor is configured to use spaces instead of tabs.

gnudles:
You know, I am really not pressing 4 times on the spacebar when I want to indent, rather I just press the tab bar. I checked QtCreator options and it showed me that it is already set on 'use tab only'. Last commit I detected all these hidden little spaces and replaced 4 of them with tab

gnudles:
@tobydox can we drop the QT_NO_SYNTAXHIGHLIGHTER from the qt build? It can really help my plugin to look more welcoming

PhysSong:
@gnudles BTW, could you resolve the conflicts?

tobydox:
> @tobydox can we drop the QT_NO_SYNTAXHIGHLIGHTER from the qt build? It can really help my plugin to look more welcoming

Done - also updated to Qt 5.15.6.

gnudles:
@PhysSong yes I should.

gnudles:
@PhysSong Tell me if something is wrong.

gnudles:
Why is it failing?

zonkmachine:
> Why is it failing?

Quite a few PR's are failing on shellcheck. Probably not on your side.
@PhysSong knows more.

PhysSong:
Yes, you can ignore this.
If you're still concerned, you can merge the latest `master` to your branch.

Rossmaxx:
@gnudles If you are still here, we can help you get this merged on master. You will have to work on any regressions coming out of this PR tho.

gnudles:
@Rossmaxx oh I forgot about that, I thought it was merged

PR #5982, opened 2021-04-12
Fix bug introduced by #5657

IanCaio:
There was a bug introduced by #5657 where reloading a project and playing it could cause a Segmentation Fault crash. After some debugging, @DomClark tracked the issue to be likely a use-after-free being caused by m_oldAutomatedValues not being cleared when the project was loaded again.

This commit adds a line to clear the `m_oldAutomatedValues` map on `Song::clearProject()`, which is called from `Song::loadProject()`.

### DIFF
--- a/src/core/Song.cpp
+++ b/src/core/Song.cpp
@@ -886,6 +886,9 @@ void Song::clearProject()
 	m_masterPitchModel.reset();
 	m_timeSigModel.reset();
 
+	// Clear the m_oldAutomatedValues AutomatedValueMap
+	m_oldAutomatedValues.clear();
+
 	AutomationPattern::globalAutomationPattern( &m_tempoModel )->clear();
 	AutomationPattern::globalAutomationPattern( &m_masterVolumeModel )->
 									clear();
--- a/src/core/Song.cpp
+++ b/src/core/Song.cpp
@@ -405,17 +405,15 @@ void Song::processAutomations(const TrackList &tracklist, TimePos timeStart, fpp
 		}
 	}
 
-	// Checks if an automated model stopped being automated by automation patterns
-	// so we can move the control back to any connected controller again
+	// Moves the control of the models that were processed earlier to their controllers.
+	// If they get processed again, setAutomatedValue() will move the control back
+	// to the automation.
 	if (!m_oldAutomatedValues.isEmpty())
 	{
 		for (auto it = m_oldAutomatedValues.begin(); it != m_oldAutomatedValues.end(); it++)
 		{
 			AutomatableModel * am = it.key();
-			if (am->controllerConnection() && !values.contains(am))
-			{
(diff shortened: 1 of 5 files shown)


Veratil (approved):

qnebra:
One liner, fixes what is supposed to fix. Test project is reloaded and played this time without crash.

IanCaio:
Last commit introduces some changes to the behavior of #5657 that remove the need to hold a pointer to the controlled model on the controller object.

qnebra:
Looks like those changes broke behaviour of 5657.

IanCaio:
> Looks like those changes broke behaviour of 5657. Be more corrent, it doesn't work like previously. User needs to mute automation track to regain control with MIDI controller.

Can you ellaborate? I tested here with a LFO controller and it behaved as previously (LFO has control when automation is not affecting model, when it's recording and when music is stopped). Could you describe step by step what you did and how it behaved differently?

qnebra:
Looks like it could be a quickest revocation of the sentence ever. I did more testing, and thinking about it. It wasn't broken, and I was just an idiot.

src/core/Song.cpp
```diff
@@ -687,6 +681,18 @@ void Song::stop()
...
+	// back to their controllers.
+	if (!m_oldAutomatedValues.isEmpty())
+	{
+		for (auto it = m_oldAutomatedValues.begin(); it != m_oldAutomatedValues.end(); it++)
+		{
+			AutomatableModel * am = it.key();
+			am->setUseControllerValue(true);
+		}
+	}
+	m_oldAutomatedValues.clear();
```
DomClark, line 694: This runs on the UI thread, so races with the audio thread. The same goes for the rest of `Song::stop`, though, so I think it should all be wrapped with `Engine::mixer()->requestChangeInModel();` and `Engine::mixer()->doneChangeInModel();` (apart from the quick return condition at the top).

src/core/Song.cpp
```diff
@@ -687,8 +690,22 @@ void Song::stop()
...
+	// Moves the control of the models that were processed on the last frame
+	// back to their controllers.
+	if (!m_oldAutomatedValues.isEmpty())
+	{
+		for (auto it = m_oldAutomatedValues.begin(); it != m_oldAutomatedValues.end(); it++)
+		{
+			AutomatableModel * am = it.key();
+			am->setUseControllerValue(true);
+		}
+	}
```
DomClark, line 702: The if statement here isn't necessary.

DomClark (approved):

Veratil:
Random pipelines seem to be failing, reran once and now linux.gcc is failing instead of mingw32. Travis failure seems to be the same problem. Assuming there's a hiccup in the system somewhere on github side, so merging since good.

Issue #3060, opened 2016-10-02
UI/UX inconsistency: Beat and bassline pattern shown in instrument track


simonvanderveldt:
For an unknown reason (can't reproduce) I managed to get the Song Editor in a state where every click on a normal instrument track would show the Beat and Bassline pattern in the respective bar I clicked on. This didn't happen for a Sample Track.
Double clicking would still bring up the Piano Roll so this seems to be only a visual bug.

![image](https://cloud.githubusercontent.com/assets/204286/19021366/08c67080-88c0-11e6-90fb-536305bb6592.png)

[edit] Actually, it's even possible to enable steps in the pattern which then actually trigger the instrument :o
![image](https://cloud.githubusercontent.com/assets/204286/19021393/99c72232-88c0-11e6-98ca-3f07c678920a.png)

Umcaruje:
This is not actually a bug, all empty patterns are BB patterns.

simonvanderveldt:
> This is not actually a bug, all empty patterns are BB patterns.

Now I'm confused :P
This is in the Song Editor window, that should never show a BB pattern, right?

Also, this is not the normal behaviour I get. Normally when I click somewhere in the Song Editor a new pattern (colored rectangle) gets created for that specific bar, not an entire BB pattern.
![image](https://cloud.githubusercontent.com/assets/204286/19022177/18a6aefe-88d2-11e6-87a3-0f4d54a88b20.png)

grejppi:
When you zoom close enough, all instrument patterns have steps, by design. This is not a bug, it's always been like this.

simonvanderveldt:
> When you zoom close enough, all instrument patterns have steps, by design. This is not a bug, it's always been like this.

:o So that's what's causing it.

Would it be OK to get rid of this and just show the rectangular color for all the zoom levels?

DeRobyJ:
Actually, thanks to this feature you can create a beat right on the song editor. This is perfect when you don't need to repeat the same thing all bars.
I feel no need to get rid of that

simonvanderveldt:
> Actually, thanks to this feature you can create a beat right on the song editor. This is perfect when you don't need to repeat the same thing all bars.
> I feel no need to get rid of that

So you use this to put some patterns with a single note down in the Song Editor? Does that mean you don't use the Beat and Bassline editor?

grejppi:
Why should they be mutually exclusive?

tresf:
[...]

PR #5988, opened 2021-04-16
Don't draw BB editor inside Song Editor

Spekular:
Currently, a small BB editor is drawn inside the Song Editor at high zoom levels. As discussed in #3060 this is unintuitive and appears broken (I've seen several other reports of this as a bug). This PR removes this behavior.

A quick Discord poll implies that 4 other developers agree with this change, but I'll leave this open a bit in case anyone wants to oppose. I'll probably merge on Sunday or so if no one objects.

### DIFF
--- a/include/PatternView.h
+++ b/include/PatternView.h
@@ -92,6 +92,8 @@ private:
 	QColor m_mutedNoteBorderColor;
 
 	QStaticText m_staticTextName;
+	
+	bool m_legacySEBB;
 } ;
  
  
--- a/src/gui/PatternView.cpp
+++ b/src/gui/PatternView.cpp
@@ -28,6 +28,7 @@
 #include <QApplication>
 #include <QMenu>
 
+#include "ConfigManager.h"
 #include "DeprecationHelper.h"
 #include "GuiApplication.h"
 #include "InstrumentTrack.h"
@@ -41,7 +42,8 @@ PatternView::PatternView( Pattern* pattern, TrackView* parent ) :
 	m_noteFillColor(255, 255, 255, 220),
 	m_noteBorderColor(255, 255, 255, 220),
 	m_mutedNoteFillColor(100, 100, 100, 220),
-	m_mutedNoteBorderColor(100, 100, 100, 220)
+	m_mutedNoteBorderColor(100, 100, 100, 220),
+	m_legacySEBB(ConfigManager::inst()->value("ui","legacysebb","0").toInt())
 {
 	connect( gui->pianoRoll(), SIGNAL( currentPatternChanged() ),
 			this, SLOT( update() ) );
@@ -180,9 +182,11 @@ void PatternView::constructContextMenu( QMenu * _cm )
 
 void PatternView::mousePressEvent( QMouseEvent * _me )
 {
+	bool displayBB = fixedTCOs() || (pixelsPerBar() >= 96 && m_legacySEBB);
 	if( _me->button() == Qt::LeftButton &&
 		m_pat->m_patternType == Pattern::BeatPattern &&
-		fixedTCOs() && _me->y() > height() - s_stepBtnOff->height() )
+		displayBB && _me->y() > height() - s_stepBtnOff->height() )
+		
 
 	// when mouse button is pressed in beat/bassline -mode
 
@@ -385,6 +389,7 @@ void PatternView::paintEvent( QPaintEvent * )
 
 	const int x_base = TCO_BORDER_WIDTH;
 
+	bool displayBB = fixedTCOs() || (pixelsPerBar >= 96 && m_legacySEBB);
 	// melody pattern paint event
 	NoteVector const & noteCollection = m_pat->m_notes;
 	if( m_pat->m_patternType == Pattern::MelodyPattern && !noteCollection.empty() )
@@ -500,9 +505,8 @@ void PatternView::paintEvent( QPaintEvent * )
 
 		p.restore();
 	}
-
 	// beat pattern paint event
-	else if( beatPattern &&	fixedTCOs() )
+	else if( beatPattern &&	displayBB )
 	{
 		QPixmap stepon0;
 		QPixmap stepon200;
--- a/src/gui/PatternView.cpp
+++ b/src/gui/PatternView.cpp
@@ -181,9 +181,8 @@ void PatternView::constructContextMenu( QMenu * _cm )
 void PatternView::mousePressEvent( QMouseEvent * _me )
 {
 	if( _me->button() == Qt::LeftButton &&
-				m_pat->m_patternType == Pattern::BeatPattern &&
-				( fixedTCOs() || pixelsPerBar() >= 96 ) &&
-				_me->y() > height() - s_stepBtnOff->height() )
+		m_pat->m_patternType == Pattern::BeatPattern &&
+		fixedTCOs() && _me->y() > height() - s_stepBtnOff->height() )
 
 	// when mouse button is pressed in beat/bassline -mode
 
@@ -503,7 +502,7 @@ void PatternView::paintEvent( QPaintEvent * )
 	}
 
 	// beat pattern paint event
-	else if( beatPattern &&	( fixedTCOs() || pixelsPerBar >= 96 ) )
+	else if( beatPattern &&	fixedTCOs() )
 	{
 		QPixmap stepon0;
 		QPixmap stepon200;


Veratil (approved):
LGTM

kamkamkamkamkamkamkamkam:
couldn't we have an option to toggle this feature on or off instead and just have it off by default? I actually use this feature and I'd seriously mess up my workflow if it were removed

Spekular:
> I actually use this feature abd it would seriously mess up my workflow if it were removed

Honest question: how? You could simply place notes in the song editor instead if you want non-looping clips. The default quantization there is 1/16 so positioning them correctly is exactly the same as doing it in the BB Editor (except of course you have more control). If you want looping clips you need to use the fully fledged BB Editor either way.

qnebra:
Or he use it for placing kick and other percussions in track. But also this create 0-tick notes, and patterns created with this appear as empty. Use actual pianoroll with ghost notes, as they are actual thing in 1.3

kamkamkamkamkamkamkamkam:
> > I actually use this feature abd it would seriously mess up my workflow if it were removed
> 
> Honest question: how? You could simply place notes in the song editor instead if you want non-looping clips. The default quantization there is 1/16 so positioning them correctly is exactly the same as doing it in the BB Editor (except of course you have more control). If you want looping clips you need to use the fully fledged BB Editor either way.

because I program percussion relative to other percussion, and seeing everything all at once is kinda essential to that. If this feature were removed I'd have to go in and out of piano rolls for different TCOs on different tracks to program rhythms which is far slower and could possibly kill any inspiration and momentum I have in the moment, I prefer using the B+B editor in the Song Editor for the flexibility compared to how restrictive arranging B+B blocks are, having to duplicate a pattern just to make any slight variation. Having the B+B editor in the Song Editor gives the best of both worlds.

kamkamkamkamkamkamkamkam:
> Or he use it for placing kick and other percussions in track. But also this create 0-tick notes, and patterns created with this appear as empty. Use actual pianoroll with ghost notes, as they are actual thing in 1.3

read the above reply

Gabrielxd195:
I don't mean to sound nosy guys, but removing the bb patterns from the song editor would be a complete setback in lmms development. because apart from ruining the workflow and the user experience, you would be removing an important function that neither FL Studio and other daws do not have. Most of the free and commercial daws (the ones I have seen) only have a '' drum editor '', a kind of piano roll but with drum patterns that work only with vst or external plugins; While in lmms you can currently add afp samples and create bb editor beats in the song editor with those samples without the need for any external vst or plugin. taking this function away from lmms would be taking away its essence, which is mostly for creating beats. I propose that this change be optional and not final.

Currently you can do some things with the bb editor patterns in the song editor, as in the bb editor itself. such as:

1- **Move and reposition the bb patterns in the song editor.** if you have a lot of bb pattern tracks in the song editor, you can move the patterns without affecting the others; however when you do this inside a track of the same bb editor, you end up moving all the other patterns that are inside, and to do this you would have to duplicate the track.

2- **You have a timeline and playback head in the song editor.** you can move the playhead to where you want and this affects the playback of it. currently the bb editor has no head and you can't go anywhere in the rhythm, just wait for it to start or finish.

3- **You can create variations without the need to duplicate the track.**
there are musical genres that have many variations and rolls like merengue, dubstep, rock. This is the biggest problem of the bb editor, that to make a small variation you must duplicate the track.

4- **You can use the selection tool on the patterns.**
Currently you can select the bb patterns you want within the song editor, while you cannot in the bb editor.

5- **You can loop the bb patterns in the song editor in any position you want**; you can make loops in the bb editor but not in the position you want since it does not have.

6- **It is easier to compose when you have voices.**
When creating a beat for a voice, it is better to use the bb patterns in the song editor than to do it within the editor itself, because it is more flexible and less cumbersome.

7- **You can easily create rhythms with samples without the need for external vst or plugin**. It is basically a '' drum sampler '' but more flexible.

**I think that removing a very important function for visual reasons is something quite absurd,** since it is not the same as composing within the bb editor and the piano roll. Instead of removing the bb patterns from the song editor, it is best to put it as a configuration option, and add more functions to it that you need, such as: resize the pattern, change color and cut. 

There I leave you a beat that I made using this wonderful function. Without this function it would be impossible for me to do.

https://www.youtube.com/watch?v=KjCCvgP3Vjw

Good afternoon.

Spekular:
I strongly disagree with the notion that LMMS's "essence" is beats. Even your own demonstration shows how weird this behavior is, there's zero visual feedback as to where notes are and aren't placed. Everything you've listed as a benefit can be done with regular MIDI clips. If opening the piano roll is one step too much, you could place your audio clips directly in a sample track.

There may be slight adjustments required for anyone relying on this behavior, but I would suggest they rip that bandaid off sooner rather than later. As it stands we have at least two reports of the behavior as a bug, and [two users saying they rely on it (xkcd link)](https://xkcd.com/1172/), so I'm not sure how to proceed.

qnebra:
Song editor for playing with patterns made in pianoroll and B&B editor. But also visual feedback on B&B patterns in song editor suck.

IanCaio:
I do agree this isn't a good or intuitive feature and will either have to go or be reworked (that latter unlikely since apparently there are no devs that actually like this feature for time to be spent reworking it).

However, there are valid points being brought up by the people who currently use it. I know that some of the workflow that this obscure feature covers could be futurely covered through regular patterns with linked patterns, improving ghost notes so we can have multiple ghost patterns, #5902 followed by adding a way to draw step notes on the piano roll, etc. But those haven't been implemented yet, so for the sake of not breaking those people's workflow I suggest we either keep it as an option (even if hidden from the configuration window, maybe only activated manually through the config XML) or keep it around until we add those missing features.

That said, I'd still suggest users that rely on it to get used to the idea of letting it go in the future and work together with us to making using the Song Editor patterns as convenient for their workflow or more than this feature it.

JGHFunRun:
I personally use this feature, and think we should make it toggle-able, default off?

Umcaruje:
I always liked this feature, although it currently more of a gimmick. I could see it being useful, especially for beginners. I think the visual problem stems from the patterns with the BB notes being the same as empty patterns when zoomed out, but #5902 is on the way to fix that and make it more useable.

Real world example: I actually used this feature for my BoL entry, I dropped in some beatboxing I did and then zoomed all the way in to figure out the pattern of my beatboxing. I know I could've done that in the BB editor, but I just don't like using the BB editor for that.

Veratil:
So I think consensus is we need to keep it, and that means rework it.

Can everyone that commented that they use it give some suggestions for how they'd like to swap to the BB view? Context menu? Keep it at zoom (but this would need to change with future zoom updates)?

Gabrielxd195:
I honestly do not know what definition of "intuitive" you have, since what you propose in this pr is the opposite of what you intend. Without offending the developers (which I admit they do a great job) but I don't know what concept of "user experience" they have, since the perception of a common user is not the same as that of a developer. If for a moment they were put in "artist mode" or "common user" by logic they would realize how useful this function is in lmms. because normally people look for possibilities and lmms is the only daw that I have seen so far that has this function, and this is a huge plus. lmms has four ways of making beats: with the bb editor, with the piano roll, with sample track samples, and finally with bb editor patterns in the song editor, which in my experience has been the most comfortable and intuitive method. to make percussion rhythms, but ironically it is the one they plan to remove.

Some do not use it: but it is there; which for others does not seem intuitive: but it is there; that others see it as a mistake, but at least it is there available for those who want to use it. We must take into account that if we have many users who oppose this change it is because we use this function in a great way, and removing it would be counterproductive and irresponsible. If no one wants or can improve it, that's fine, but at least it will be available and it won't hurt anyone.

to clarify what I say, I will give some examples even if they seem exaggerated: fl studio has the option to record audio; however many users complain about how uncomfortable it is and use other daws for that, but at least that function is available; Armor has a piano roll that leaves much to be desired, but it is there for anyone who wants to use it; qtractor gives picason in the eyes of how ugly it is, but it is full of useful functions; in the case of lmms, a few years ago the sample track was just a simple '' sample track '' for playing samples, it was ugly and buggy, but it was available. The same goes for the bb pattern function in the song editor, it is buggy and incomplete, but it has been there for years and those who have discovered it have valued it.

Now my question is: What benefit do we get if we leave this function? make beats with a few clicks, just by going to your browser, taking your sample, zooming in and quickly marking the bb pattern notes, and taking advantage of the song editor features that the bb editor does not have; But what do we gain by removing this function? just make it look more beautiful and whim granted !!!.

It is absurd to remove something that works simply for aesthetic or philosophical reasons, or because someone does not like it, or it seems ugly, or because in some uncertain future it occurs to someone to add another better function. The problem is not that you don't want to use the piano roll or the bb editor, the problem here is to remove a function and a possibility simply on a whim; a possibility that has lmms that not all daws have, and that is lack of vision friend. If you want to make this a little more democratic, go to YouTube and Facebook groups and ask the channels and users if they agree with removing this function. Greetings

qnebra:
You aren't common user, and devs use lmms in production.

Gabrielxd195:
> You aren't common user, and devs use lmms in production. You just found it useful for your own workflow.

As I am not a common user ?, I am not a developer, what would you call me then ?. Also I know that developers do not use lmms in production, when I have said otherwise?

qnebra:
Well, you imply your personal workflow as the same as other users. I think it needs to be removed, but also see benefits when precission is needed, especially in relation to elements in Song Editor.

DouglasDGI:
It's essential that this feature is removed in the future.  As LMMS moves more and more towards supporting a sample-based workflow as sample editing is supported and sample tracks are enhanced, zooming in LMMS will become significantly smoother and, most importantly to this issue, the user will be working at extremely high levels of zoom incredibly often.  

Regardless of anybody's opinions, this feature must go.  In the end it's simply a sloppy workaround for having an actually decent implementation of these same features, and keeping it will block other features that are absolutely essential for LMMS's survival.

Because of this, I suggest that instead of fighting for this feature to be kept, people start brainstorming how efficient sequencing in this manner can be kept via a separate feature which actually makes sense to keep.  In the end, sample editing and sample-based workflows take very high priority over this.

kamkamkamkamkamkamkamkam:
> So I think consensus is we need to keep it, and that means rework it.
> 
> Can everyone that commented that they use it give some suggestions for how they'd like to swap to the BB view? Context menu? Keep it at zoom (but this would need to change with future zoom updates)?

either via context menu or a keybinding, that'd be great and frankly even better than it being activated by zooming in (which is pretty janky). 


> Because of this, I suggest that instead of fighting for this feature to be kept, people start brainstorming how efficient sequencing in this manner can be kept via a separate feature which actually makes sense to keep. In the end, sample editing and sample-based workflows take very high priority over this.

If we were actually to go this route, I think it would be nice to have a separate B+B editor-like window that is basically a mirror of the Song Editor except with B+B editors, and when you program rhythms in it, it is mirrored in a regular TCO in the Song Editor. (maybe this feature could even completely replace the B+B editor?)

qnebra:
Dumb idea:

B&B as sort of "group instrument" in Song Editor.
- collapsed form - user sees it as current B&B pattern, can quickly clone pattern and they are more visible than current B&B mainly to give better idea of what is going on
- expanded form - user can made pattern, assign it to be "default pattern - default pattern is cloned in collapsed form", or made some changes in individual pattern when needed, as grid and timings are the same as in Song Editor. Also it could act as drum machine, maybe with ability to assign midi note to each instrument in it.

Spekular:
The behavior is now toggleable. By default it's off, and in order to enable it you need to manually add `legacysebb="1"` inside the `<ui>` tag in `.lmmsrc.xml`. This avoids cluttering the already dense settings menu, and ensures that no new users will accidentally become accustomed to a behavior that's destined for removal.

IanCaio:
> but I don't know what concept of "user experience" they have, since the perception of a common user is not the same as that of a developer.

I can't say for every other developer, but I develop and constantly produce under LMMS. So does @DouglasDGI and probably others. Sure, we might all use different production workflows, but assuming we don't have a notion of user experience was a wrong assumption you made.

> But what do we gain by removing this function?

Every piece of code has to be maintained. Cleaning a code base from unnecessary parts of code is part of making a project sustainable. Keeping every piece of code forever even when the project is changing will soon bite you in the back.

That feature having confused many users has shown to not be intuitive. We need a good way of sequencing drums on the song editor, but that one isn't it. We are listening to the users that rely on that feature and intend on working on alternatives that makes their workflow possible without it.

IanCaio (approved):
LGTM! Now that users that rely on this feature have a way of enabling it I'm cool with merging this. As soon as we finish working the lacking features we can remove it completely!

Spekular:
> I can't say for every other developer, but I develop and constantly produce under LMMS.

For the record, I also use LMMS regularly, and that's the entire reason I got into developing it.

> > But what do we gain by removing this function?

(1) Principle of least astonishment, (2) less obscure behavior that could conflict with other features, and (3) simplified code. (1) is just good UX, something that we need if LMMS is going to reach it's full potential. For proof that (2) is valid, look no further than zoom. Since the BB editor uses scrolling to adjust step velocity, it's not possible to zoom in or out while hovering over an empty clip at 8 or 16 times zoom. As for (3), the original PR accomplished this, but the newly added toggle actually makes the code more complex.

Issue #5978, opened 2021-04-09
Stuck VU meter on muted mixer channel


softrabbit:
### Bug Summary
Stuck VU meter on muted mixer channel, pretty sure nobody has filed an issue on this.

#### Steps to reproduce
Play sound, mute mixer channel while sound is playing.

#### Expected behavior
The meter on the muted channel should drop to 0.

#### Actual behavior
The meter on the muted channel stays up.

#### Screenshot
![screenshot](https://user-images.githubusercontent.com/6411463/114138599-897e0600-9916-11eb-9ecf-6b1fd317afbe.png)

#### Affected LMMS versions
Version 1.3.0-alpha.1.33+g4f74151 (Linux/x86_64, Qt 5.6.3, GCC 5.4.0 20160609). AppImage from lmms.io
Master pulled from git on 2021-03-08 and compiled on Ubuntu 20.04

#### Logs
<details>
  <summary>Click to expand</summary>
<pre>
Carla does not appear to be installed.  That's OK, please ignore any related library errors.
Jack appears to be installed on this system, so we'll use it.
Gtk-Message: 09:22:25.628: Failed to load module "overlay-scrollbar"
Gtk-Message: 09:22:25.647: Failed to load module "atk-bridge"
Fontconfig warning: "/etc/fonts/conf.avail/65-ttf-droid-sans-fonts.conf", line 61: Having multiple values in <test> isn't supported and may not work as expected
Fontconfig warning: "/etc/fonts/conf.avail/65-ttf-droid-sans-fonts.conf", line 96: Having multiple values in <test> isn't supported and may not work as expected
Fontconfig warning: "/etc/fonts/conf.avail/65-ttf-droid-sans-fonts.conf", line 61: Having multiple values in <test> isn't supported and may not work as expected
Fontconfig warning: "/etc/fonts/conf.avail/65-ttf-droid-sans-fonts.conf", line 96: Having multiple values in <test> isn't supported and may not work as expected
VST sync support disabled in your configuration
Lv2 plugin SUMMARY: 0 of 5  loaded in 4 msecs.
For details about not loaded plugins, please set
  environment variable "LMMS_LV2_DEBUG" to nonempty.
"Cannot load library /usr/lib/ladspa/autotalent.so: (/usr/lib/ladspa/autotalent.so: undefined symbol: __exp_finite)"
[...]

PR #5991, opened 2021-04-17
Fix for issue #5978, VU meters in FX mixer stuck when channel is muted.

softrabbit:
Throwing this to the wolves now, eventhough I know it'll need rebasing after the Great Reorganising. 

Tried my code in one channel and the old in another, with slow falloff to see the difference and I'd say it was minimal. Also tried with a larger buffer size and a faster falloff and didn't see any ill effects. I'm undoing a bugfix here, but I'm not sure about the bug...

### DIFF
--- a/src/gui/FxMixerView.cpp
+++ b/src/gui/FxMixerView.cpp
@@ -598,29 +598,19 @@ void FxMixerView::updateFaders()
 
 	for( int i = 0; i < m_fxChannelViews.size(); ++i )
 	{
-		const float opl = m_fxChannelViews[i]->m_fader->getPeak_L();
-		const float opr = m_fxChannelViews[i]->m_fader->getPeak_R();
 		const float fallOff = 1.25;
-		if( m->effectChannel(i)->m_peakLeft >= opl/fallOff )
-		{
-			m_fxChannelViews[i]->m_fader->setPeak_L( m->effectChannel(i)->m_peakLeft );
-			// Set to -1 so later we'll know if this value has been refreshed yet.
-			m->effectChannel(i)->m_peakLeft = -1;
-		}
-		else if( m->effectChannel(i)->m_peakLeft != -1 )
-		{
-			m_fxChannelViews[i]->m_fader->setPeak_L( opl/fallOff );
-		}
+		
+		m_fxChannelViews[i]->m_fader->setPeak_L(
+			qMax( m->effectChannel(i)->m_peakLeft,
+			      m_fxChannelViews[i]->m_fader->getPeak_L()/fallOff)
+			);
 
-		if( m->effectChannel(i)->m_peakRight >= opr/fallOff )
-		{
-			m_fxChannelViews[i]->m_fader->setPeak_R( m->effectChannel(i)->m_peakRight );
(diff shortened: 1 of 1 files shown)


Rossmaxx:
@softrabbit if you are still around, mind coming back and finishing? 

Also @michaelgregorius did some work with the mixer stuff. He might be able to help.

src/gui/FxMixerView.cpp
```diff
@@ -598,29 +598,19 @@ void FxMixerView::updateFaders()
...
-			m->effectChannel(i)->m_peakRight = -1;
-		}
-		else if( m->effectChannel(i)->m_peakRight != -1 )
-		{
-			m_fxChannelViews[i]->m_fader->setPeak_R( opr/fallOff );
-		}
+		
+		m_fxChannelViews[i]->m_fader->setPeak_L(
+			qMax( m->effectChannel(i)->m_peakLeft,
+			      m_fxChannelViews[i]->m_fader->getPeak_L()/fallOff)
```
michaelgregorius, line 605: I am aware that it was not really introduced by this PR but IMO this is the wrong place for this logic. Why should the whole FX mixer view decide how the Faders behave with regards to the falloff?

The functionality should be moved into the `Fader`  class. The `Fader` should only be fed with the raw peak values and it's then up to the presentation code of the fader to decide how the display is rendered and updated, e.g. how much lag there is.

softrabbit: - In principle, you're right.
- In practice, would I ever want a mixer with VU meters not having identical falloff? Don't think so. 
- Decentralizing the falloff would mean the Fader class probably needs some concept of time - a Qt signal or whatever way the FxMixerView gets its updates. Or then some contract of the _"don't refresh this more than once for every refresh period"_ kind, which I have a bad feeling about. Something will violate that sooner or later.
- I like making my PRs minimal, maybe too minimal sometimes. Possibly hoping it would speed up the process or something.

Feel free to put in your own PR and supersede mine or whatever, I don't mind.

michaelgregorius: The problem is that with the current code it's not even possible to have meters with different falloffs. If someone wanted to implement an analog meter with a different falloff behavior then it would not be possible.

With regards to the concept of time: I wonder if it would not be better in the first place if the faders/meters would get the full contents of the sample buffers and then process them as they see fit, i.e. not even some pre-computation should be done outside. They should also know the current sample rate so that they could use it to compute the falloff factors or coefficients of low pass filters so that the falloff looks the same for all sample rates. Looking at the code it seems that currently they might behave and look differently for different sample rates, e.g. much more fidgety for high sample rates like 192 kHz.

I think minimal PRs are good because they are much better to review. :+1:

softrabbit: I don't think sample rates really come into play here, as the faders are concerned with peaks, which shouldn't differ that much from one rate to another? IIRC, the fader falloff calculation is triggered from some QTimer, and that having a shorter interval would make the faders more fidgety (and probably change some other things visually as well).

softrabbit:
Closing this in favor of a more thorough refactoring.

PR #5994, opened 2021-04-22
Fix JACK MIDI dropping events when two events have the same time stamp.

softrabbit:
As reported and debugged on Discord: JACK MIDI input has a tendency to drop events. 
Explanation: 
- ```jack_midi_event_get``` is called to get a new event at the end of the for loop, if ```in_event_time == i``` 
- ```i``` gets incremented for the next iteration. 
- If the newly fetched event has the same ```time``` as the previous one, the comparison will be false and remain so for the rest of the buffer. Any further events will be ignored.

I suggest turning the ```if``` into a ```while``` to handle all events with the same ```time```value.

### DIFF
--- a/src/core/midi/MidiJack.cpp
+++ b/src/core/midi/MidiJack.cpp
@@ -188,7 +188,7 @@ void MidiJack::JackMidiRead(jack_nframes_t nframes)
 	{
 		for(i=0; i<nframes; i++)
 		{
-			if((in_event.time == i) && (event_index < event_count))
+			while((in_event.time == i) && (event_index < event_count))
 			{
 				// lmms is setup to parse bytes coming from a device
 				// parse it byte by byte as it expects


VOID404:
As the person reporting the bug, I cannot reproduce it on this version

Veratil:
@JohannesLorenz you're the previous committer for this loop fixing another bug, so pinging you for attention (would hate to regress). I see no issue changing `if` to `while` though. So approval from me. Would like your thumbs up, too.

softrabbit:
Looking at the function around the fix, I wonder if it could be straightened into a ```while(jack_midi_event_get(&in_event, port_buf, event_index) )```.

Veratil:
Tangential topic: I did some looking around, it's possible we may have another bug in this loop related to SysEx messages.
https://github.com/jackaudio/jack2/commit/b744332c5d71d825d237fc8ed72fcb8a60dcf5e6

softrabbit:
> Tangential topic: I did some looking around, it's possible we may have another bug in this loop related to SysEx messages.
> [jackaudio/jack2@b744332](https://github.com/jackaudio/jack2/commit/b744332c5d71d825d237fc8ed72fcb8a60dcf5e6)

```MidiClientRaw::parseData``` which handles the bytes will most likely just ignore the SysEx, so I think we're safe on that point.

JohannesLorenz:
Sorry, I missed this, but it looks basically good :+1:

Issue #879, opened 2014-06-22
Wrong cursor when moving blocks in song-editor


musikBear:
1.0.2   32b
Song-editor 
State 'select & move' 
cursor changes to 'extend' at block-ends, but marked blocks moves
(already repd ??)

Umcaruje:
Confirmed on Ubuntu 15.04 x64.

Have a gif:
![gifrecord_2015-06-30_191124](https://cloud.githubusercontent.com/assets/6282045/8436960/37148450-1f5c-11e5-8879-9b1199bc81fc.gif)

SecondFlight:
Leaving a comment to remind myself to reproduce when I get home.

SecondFlight:
Reproduced in 1.2.0 RC8 and master (as of [this build](https://github.com/LMMS/lmms/pull/4639#issuecomment-425760993)).

SecondFlight:
(that was actually me)

As part of a pruning effort, this bug report is archived into a dedicated "Better Workflow" checklist here #4877.

claell:
The question is, what is the expected behaviour? Not having a resize cursor when in selection mode? This also happens when selecting only one "block". Instead of resizing it gets moved.

SecondFlight:
I think either would be acceptable behavior. Either attempt to resize all selected TCOs (am I using that term right?), or don't and disable the cursor.

craslaw:
About to pick this up as a first issue 🚀. I think there's valid cases for both behaviors.

If you want to resize multiple BB blocks at once, it makes sense to select each block then drag out the edge of one block to resize them all. They would all be resized by the same amount of bars. Other track types that allow block resizing like the sample and automation tracks could also be compatible with BB resizing.

However, track types that don't allow resizing (instruments, presets) fit the other behavior better. In the case were a BB block and an instrument block are selected together, the resize cursor shouldn't appear on the edge of the BB block.

I know this was moved to #4877, but it seemed easier to discuss here. Just let me know if it's better to move the discussion to #4877 or Discord.

claell:
I think discussing the details here is indeed better. You can also ask to reopen this issue in #4877 if you plan to work on it.

PR #5996, opened 2021-04-26
Fix for issue #879, wrong cursor for selected resizeable blocks in song editor

craslaw:
This PR checks if TCOs are selected by the box selector before switching the cursor to the resize cursor at the edge of a selected block. 

Before (Issue #879):
The cursor updates to the resize cursor for one or more box-selected TCOs, but clicking and dragging results in moving all the selected TCOs.

After:
Same behavior as before, except the cursor does not change from the hand icon when hovering over the edge of a box-selected TCO.

From a UI stance, it makes sense to be able to resize one or more resizable blocks that are selected with the box tool. However, adding this code to the box select tool seems out of scope for this bug, and may be better off in a separate feature request - post refactor too.


superpaik:
It works ok to me (windows 10-64bits)
Just one minor issue that I think it's not been introduced by this PR but maybe just double check just in case  Samples tracks now can be extended to the left, I don't know if that is intended or not (it doesn't happen in 1.2)

PhysSong:
@Spekular While reviewing this PR, I found that knife mode is not working with selections. Is it intended?
If it is, this looks good to me. There are some merge conflicts due to renaming from `TrackContentObject`(TCO) to `Clip`, though.

Spekular:
> @Spekular While reviewing this PR, I found that knife mode is not working with selections. Is it intended?

I didn't really have selections in mind when I wrote it. I'm not sure what the best UX would be for cutting multiple clips so for now I think it's fine to only cut one at a time?

src/gui/TrackContentObjectView.cpp
```diff
@@ -448,13 +448,13 @@ void TrackContentObjectView::updateCursor(QMouseEvent * me)
 	SampleTCO * sTco = dynamic_cast<SampleTCO*>(m_tco);
 
 	// If we are at the edges, use the resize cursor
-	if ((me->x() > width() - RESIZE_GRIP_WIDTH && !me->buttons() && !m_tco->getAutoResize())
-		|| (me->x() < RESIZE_GRIP_WIDTH && !me->buttons() && sTco && !m_tco->getAutoResize()))
+	if ((!me->buttons() && !m_tco->getAutoResize() && !isSelected())
```
allejok96, line 451: Redundant parenthesis for the first 3

allejok96 (approved):
Looks good, if you can resolve the merge conflicts I think we can merge this and close the bug. It's such a simple change. Thanks for your contribution, even if the response has been slow.

Issue #5998, opened 2021-04-27
Muted tracks' activity indicator is out of sync with notes.


softrabbit:
### Bug Summary
Muted tracks' activity indicator is out of sync with notes.

#### Steps to reproduce
Add instruments to B&B editor, make them play the same rhythm and mute some.

#### Expected behavior
Muted channels' activity indicators should stay unlit, or if they're now intended to light up, they should do it in sync with what is playing.

#### Actual behavior
Muted channels light up, later than unmuted channels. This happens in both B&B and Song editors and can be observed across the editors, too.

#### Screenshot
![eaLNeiH5VN](https://user-images.githubusercontent.com/6411463/116225564-aca01680-a75a-11eb-826c-fb73d665f975.gif)

#### Affected LMMS versions
Version 1.3.0-alpha.1.102+g89fc6c960 (win32/x86_64, Qt 5.9.8, GCC 7.3-win32 20180312).

Spekular:
It almost looks like delay is caused by a ramp up in intensity on muted LEDs, could it be that the intensity animation is inverted/reversed?

enp2s0:
It looks like when a track is muted, it disables the "white flash" part of the animation, but not the "gray fade" part.

enp2s0:
nevermind, that's incorrect. It seems like the offending code is in TrackView.cpp:
```void TrackView::muteChanged()
{
	FadeButton * indicator = getActivityIndicator();
	if (indicator) { setIndicatorMute(indicator, m_track->m_mutedModel.value()); }
}




void TrackView::setIndicatorMute(FadeButton* indicator, bool muted)
{
	QPalette::ColorRole role = muted ? QPalette::Highlight : QPalette::BrightText;
	indicator->setActiveColor(QApplication::palette().color(QPalette::Active, role));
}
```
when a track is muted, it overrides its fadebutton "activated" color, but not its fadebutton "note held" color. I'll try to get a PR fixing this.

enp2s0:
Just created PR #6000 that should fix this.

PR #6000, opened 2021-04-28
Properly disable FadeButtons when tracks are muted

enp2s0:
FadeButtons on muted tracks weren't properly disabled (they wouldn't pulse when they activate, but they would still glow dimly when the note was being held). This fixes that. 

It also changes the way FadeButtons are disabled, as the old way specified colors in two unique places, meaning that changing the colors of an active Fadebutton might not affect a muted one, and vice versa. This PR adds a new method to FadeButtons that allows them to be enabled/disabled, and simply keeps them at their base color when disabled.


enp2s0:
Fixes #5998

enp2s0:
Also, ignore all the bogus PR commits, they are from me making PRs from lmms/lmms to enp2s0/lmms to update my forked master branch.

IanCaio (changes requested):
Minor code style guidelines fixes.

At first glance code looks good. Did you test muting the track in the middle of the fading animation? Does it stop in the animation before it finishes?

enp2s0:
When muting a track that's currently playing a note, the animation immediately stops (which I think is the best way to handle this, since the sound immediately stops as well).

I'm hesitant to apply those style changes -- it seems better to follow the style of the files for such a small change, it would be confusing imo to have a file with everything in one style and a few lines in a different one. When we do a project-wide formatting the styles will be corrected everywhere anyway, and having, for example, two different styles in the *same function definition* is a bad practice imo.

softrabbit:
Looks good. The compiled LMMS, that is. I'll leave the code formatting nuances to someone else, but I know I'm firmly on the side of "make the code fit in with the surrounding code", e.g.:

> I'm hesitant to apply those style changes -- it seems better to follow the style of the files for such a small change, it would be confusing imo to have a file with everything in one style and a few lines in a different one. When we do a project-wide formatting the styles will be corrected everywhere anyway, and having, for example, two different styles in the _same function definition_ is a bad practice imo.

IanCaio:
This argument of the automatic changes have been used before and I've even been flexible with style change requests on recent previous reviews but I'm back to being a little more strict, and here's why:

The argument against a strict code style review was that there would be an automatic code style change on the whole code base (the clang PR) and that it would fix everything anyways. The Clang PR is being worked on right now and as I have warned before it isn't that simple to automatically fix everything. The code style isn't going to magically be 100% fixed in the code base and there will probably be stuff left to be fixed manually. The more uncompliant code is added, the more work is also added to the refactoring. We have very few active devs that would actually work on this and I personally, as one of the maybe 2 devs that was more strict with style, will feel very uncomfortable if I've to fix it on a PR that I had already reviewed for style before.

If we are not being strict with code style we should remove the code style guidelines page even if temporarily, it feels tiring to have to defend something that is explicitly written in our contribution wiki. I can even notice how some contributors dislike having this kind of review on their PRs like somehow I'm nit-picking them.

### That all aside:

There's no surrounding code on `muteChanged` that uses one line `if` with no explicit code blocks, so at the very least that one I'll have to be firm about.

IanCaio:
Oh, and I'm cool with the animation stopping midways. I was just double checking because I haven't tested the build yet and from the code it seemed that if you muted in the middle of the animation, the LED might be frozen in the animation frame it was at the moment of the muting. All good by me if we use this behavior.

Spekular:
The LEDs blinking when muted is intentional, it's just the color gradient that needs adjustment. Even if you intend to have them not blink, you could simply make use of the existing muted property instead of introducing a new `enabled` property.

PR #6003, opened 2021-04-30
Improve spectrum analyzer performance by caching most used computations

he29-net:
As previously discussed on Discord, my first attempt at improving the spectrum view improved the performance by around 4 % and needed 16 extra lines, while making the code less readable. So it did not seem worthwhile to merge that.

But. People say there is no such thing as a free lunch, but after thinking about it a bit more, I was able to bump that to 22 lines and a 100 % performance improvement. If that's not close to a free lunch, I don't know what is.

Compared to the first attempt, I now moved the caching closer to the point of use and extended it to a full LUT of complete results, instead of just pre-computed parts of a formula. Before the optimization, the `QPainter` path building took 9.5 ms out of 13.5 ms for the whole redraw (stereo and peak hold enabled, default TripleOsc tone playing, default analyzer window size on a 4K screen). After the update, path build takes only 1.8 ms and and the full redraw is 6.5 ms (there is also an extra 0.7 ms difference, not sure where that came from). That is, the maximum FPS limit increased from 74 to 154, leaving much more headroom for other plugins and GUI elements.

### DIFF
--- a/plugins/SpectrumAnalyzer/SaSpectrumView.cpp
+++ b/plugins/SpectrumAnalyzer/SaSpectrumView.cpp
@@ -25,7 +25,7 @@
  */
 
 #include "SaSpectrumView.h"
-#include <iostream>
+
 #include <algorithm>
 #include <cmath>
 #include <QMouseEvent>
--- a/plugins/SpectrumAnalyzer/SaSpectrumView.cpp
+++ b/plugins/SpectrumAnalyzer/SaSpectrumView.cpp
@@ -52,7 +52,8 @@ SaSpectrumView::SaSpectrumView(SaControls *controls, SaProcessor *processor, QWi
 	m_cachedRangeMax(-1),
 	m_cachedLogX(true),
 	m_cachedDisplayWidth(0),
-	m_cachedBinCount(0)
+	m_cachedBinCount(0),
+	m_cachedSampleRate(0)
 {
 	setMinimumSize(360, 170);
 	setSizePolicy(QSizePolicy::Minimum, QSizePolicy::Minimum);
@@ -362,13 +363,15 @@ QPainterPath SaSpectrumView::makePath(std::vector<float> &displayBuffer, float r
 	float rangeMin = m_processor->getFreqRangeMin(m_controls->m_logXModel.value());
 	float rangeMax = m_processor->getFreqRangeMax();
 	if (rangeMin != m_cachedRangeMin || rangeMax != m_cachedRangeMax || m_displayWidth != m_cachedDisplayWidth ||
-		m_controls->m_logXModel.value() != m_cachedLogX || m_processor->binCount() + 1 != m_cachedBinCount)
+		m_controls->m_logXModel.value() != m_cachedLogX || m_processor->binCount() + 1 != m_cachedBinCount ||
+		m_processor->getSampleRate() != m_cachedSampleRate)
 	{
 		m_cachedRangeMin = rangeMin;
 		m_cachedRangeMax = rangeMax;
 		m_cachedDisplayWidth = m_displayWidth;
 		m_cachedLogX = m_controls->m_logXModel.value();
 		m_cachedBinCount = m_processor->binCount() + 1;
+		m_cachedSampleRate = m_processor->getSampleRate();
 		for (unsigned int n = 0; n < m_cachedBinCount; n++)
 		{
 			m_cachedBinToX[n] = freqToXPixel(binToFreq(n), m_displayWidth);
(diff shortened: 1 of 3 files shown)


Spekular (approved):

plugins/SpectrumAnalyzer/SaSpectrumView.cpp
```diff
@@ -348,10 +351,28 @@ QPainterPath SaSpectrumView::makePath(std::vector<float> &displayBuffer, float r
...
 	float x_start = -1;		// lower bound of currently constructed point
+
+	// Speed up bin → x position translation by building a LUT cache.
+	// Update the cache only when range or display width are changed.
+	float rangeMin = m_processor->getFreqRangeMin(m_controls->m_logXModel.value());
+	float rangeMax = m_processor->getFreqRangeMax();
+	if (rangeMin != m_cachedRangeMin || rangeMax != m_cachedRangeMax || m_displayWidth != m_cachedDisplayWidth)
+	{
+		m_cachedBinToX.resize(m_processor->binCount() + 1);
+		m_cachedBinToX.clear();
```
PhysSong, line 362: Clearing the vector before overwriting everything looks redundant. Also, isn't it possible to reserve the memory space on construction? it's only about 32k per an instance.

he29-net: Ah, you are right, I did not even realize the clear is completely pointless.

As for the resize itself: first I was worried that the constant could be overlooked when someone decides to increase the maximum FFT block size a decade later (since everything else follows the `processor->binCount()`). But I guess I can simply check the last element of `FFT_BLOCK_SIZES` and allocate the LUT based on that. It could slightly improve performance while the window is being resized (since each pixel-sized step of the resize triggers cache update), so why not. I wonder if it even makes sense to use `array` instead of `vector`, although I doubt it will bring any performance benefit.

In other news, your comment made me realize that there is one more variable that can invalidate the cache: the `binCount`. When the number of bins changes, the mapping of bin to frequency also changes, so the LUT must be rebuilt.

he29-net: Updated.
I kept the cache a `vector`, since the size of `array` could not be defined by the last element from the list of allowed FFT block sizes (the compiler does not consider it constant enough at compile time).

PhysSong (dismissed):

he29-net:
OK, done. The Bass range problem was really unrelated, so maybe the alpha just runs better due to better compile-time optimization or the lack of profiling probes. (The problem was that the Bass range is the most "zoomed in" in respect to individual bins, so when considering the pixel range to draw, the total image width at that zoom level turns out to be around 160 000 pixels. Even though only like 2000 pixels out of it will be drawn, each pixel was tested to see if it belongs to the range useful for drawing. So all that was needed to fix this was to move the test directly to the for cycle init / end condition, to make sure it does not look outside of the useful range at all.)

PhysSong:
Looks good now. BTW, can you guarantee the cache miss condition never changes, or not?

he29-net:
> Looks good now. BTW, can you guarantee the cache miss condition never changes, or not?

I don't think there is any way to guarantee it, but I can't currently think of any future improvement that could affect it. It currently triggers on all inputs of `SaProcessor::freqToXPixel()`, so it should be pretty safe. Just to be sure, I added a comment to that function, explaining that the results may be cached and any new inputs added to the function in the future must also trigger cache update.

Looking deeper, there is also `SaProcessor::binToFreq()`, which could in theory affect the result if sample rate / Nyquist frequency changes. But since LMMS does not support runtime changes of sample rate, this should not be a problem.

--- OK, wait: some time ago I was able to observe sample rate changes visible in the analyzer when exporting at a higher sample rate than LMMS currently uses. First I wasn't able to reproduce it (i.e. the "auto" X range did not adjust to the higher Nyquist limit during export), but I just found out the problem is that OGG export silently ignores the selected sample rate setting and exports at 48 kHz (it would make sense for Opus, which uses fixed 48 kHz, but I thought OGG in LMMS refers to OGG Vorbis.. either way a bit confusing).

When I do a high sample rate export in FLAC, I can still see the change in X range. But because of the cache, the actual displayed spectrum matches the old range instead. I'm not sure if I should add sample rate as another trigger to rebuild the cache. It would be "more correct" and future-proof, but it would be also one more condition that needs to be checked every time and basically never changes. Still, I'm inclined to do it, considering it may save some trouble in the future and compared to the cache savings the impact of one more check is insignificant.

he29-net:
Sample rate check added. Should be pretty bullet-proof now. I think.

PhysSong (approved):
LGTM. Sorry for the very late review.

he29-net:
@PhysSong Thanks. Don't worry about the delay, we are technically still in the refactor-freeze, so I expected it to take a while.

There was another review from Spekular, so unless anyone else wants to look through the code, I suppose it is ready to merge?

PR #6005, opened 2021-05-04
Use standard configuration directories

### LINKED ISSUE #5869: XDG Base Directory
On Linux, in order to reduce the clutter in the user's home directory, LMMS could use the XDG Base Directory specification (https://wiki.archlinux.org/index.php/XDG_Base_Directory) for its configuration file (.lmmsrc.xml).

For backwards compatibility, the configuration would be read initially from the current file location (~/.lmmsrc.xml). If not found, it would be read from ```$XDG_CONFIG_HOME/lmms/lmmsrc.xml```, with ```$XDG_CONFIG_HOME``` defaulting to ```$HOME/.config``` if not set.

M374LX:
With this PR, the "lmmsrc.xml" file is stored in the platform's standard directory for configuration files, but the file at the legacy location is checked first for backwards compatibility.

According to the documentation at https://doc.qt.io/qt-5/qstandardpaths.html#StandardLocation-enum, the location is ``~/.config/<APPNAME>`` for Linux, ``C:/Users/<USER>/AppData/Local/<APPNAME>`` for Windows, and ``~/Library/Preferences/<APPNAME>` for macOS.

Closes #5869.

**Note:** tested only on Linux.

### DIFF
--- a/src/core/ConfigManager.cpp
+++ b/src/core/ConfigManager.cpp
@@ -645,7 +645,20 @@ void ConfigManager::initPortableWorkingDir()
 void ConfigManager::initInstalledWorkingDir()
 {
 	m_workingDir = QStandardPaths::writableLocation(QStandardPaths::DocumentsLocation) + "/lmms/";
-	m_lmmsRcFile = QDir::home().absolutePath() +"/.lmmsrc.xml";
+
+	// Try the legacy RC file location first
+	m_lmmsRcFile = QDir::home().absolutePath() + "/.lmmsrc.xml";
+
+	// If the RC file is not found, try the new location
+	if (!QFileInfo(m_lmmsRcFile).exists())
+	{
+		QString cfgDir = QStandardPaths::standardLocations(QStandardPaths::AppConfigLocation).at(0);
+		QDir dir;
+
+		dir.mkpath(cfgDir);
+		m_lmmsRcFile = cfgDir + "/lmmsrc.xml";
+	}
+
 	// Detect < 1.2.0 working directory as a courtesy
 	if ( QFileInfo( QDir::home().absolutePath() + "/lmms/projects/" ).exists() )
 		m_workingDir = QDir::home().absolutePath() + "/lmms/";


superpaik:
It works ok with windows (64) as well. My user folder is in D: (not as usual in C:) and it works ok.
Some things I've noticed. 
- Before, the file stored at drive:\users\<username> the filename begun with a dot ".lmmsrc.xml" so it was a hide file, now it's a normal file (without dot at the beginning) It's not a problem to me and I think that's the usual behaviour for other apps, but I don't know if this is something wanted or a sideeffect.
- As you say in the PR description, there is backwards compatibility so the legacy location is checked. That is ok, but it is stored there as well. I don't know if the config file should be stored in the new location, otherwise, if a person has the old version and change to the new, the lmmsrc.xml location won't change. I don't have a solid opinion on what the correct behaviour should be.

M374LX:
@superpaik 
The absence of the dot at the beginning of the file name is intentional, as most Linux applications that use the .config directory have no dot in their files.

MRAAGH:
This would be great to see happen! Putting configuration files in the right places seems more professional. I don't know about those additional commits though, does that really belong in the same PR?

M374LX:
@MRAAGH 
The additional commits in the same PR are helpful in preventing code rot and bugs.

RiedleroD:
what happens if files in both locations exist? I haven't checked, but imo LMMS should either give a notice or parse the newest config.

MRAAGH:
> what happens if files in both locations exist?

Usually you would have a predefined order in which candidate configuration files are checked. When a configuration file is found, no additional paths are checked. So, if both locations exist, one overrides the other. Don't rely on modification time with config files, that's just asking for weird accidents

M374LX:
> 

> what happens if files in both locations exist? I haven't checked, but imo LMMS should either give a notice or parse the newest config.

The legacy location (the user's home directory) is checked first and used if the ```.lmmsrc.xml``` file is found there. Otherwise, the new location is checked.

Spekular:
Wouldn't it be better to prefer the new location, and only use the old one as a fallback?

M374LX:
492fcac makes the new location preferable, as suggested by @Spekular.

Spekular:
Is there a reason for all the manual inlining? It makes the diff a lot harder to follow and I can't tell if it's necessary somehow.

M374LX:
@Spekular No specific reason for the manual inlining. Should methods like ``initPortableWorkingDir()`` and ``initInstalledWorkingDir()`` be kept?

Spekular:
> Should methods like `initPortableWorkingDir()` and `initInstalledWorkingDir()` be kept?

At a glance I'd say it seemed easier to tell what was being done. I would keep them, but since it's a matter of taste I think the second reviewer can decide (serving as tiebreaker).

Spekular (approved):

M374LX:
b49e64d restores the directory init methods in case the maintainers prefer this way.

M374LX:
@Spekular I forgot, but the actual reason for the manual inlining is in the description of this PR. It is because ``initPortableWorkingDir()`` and ``initInstalledWorkingDir()`` are not good names due to the fact that the methods also set the RC file. For this reason, I think b49e64d should be reverted.

Spekular:
I think `initPortablePaths` and `initInstalledPaths` would be fairly descriptive alternatives, perhaps?

src/core/ConfigManager.cpp
```diff
@@ -699,12 +687,12 @@ void ConfigManager::initDevelopmentWorkingDir()
...
 			}
 			if (line.startsWith("lmms_BINARY_DIR:")) {
-				m_lmmsRcFile = line.section('=', -1).trimmed() +  QDir::separator() +
-							   ".lmmsrc.xml";
+				m_lmmsRcFile = line.section('=', -1).trimmed() +  QDir::separator() + ".lmmsrc.xml";
 				done++;
 			}
 			if (done == 2)
 			{
+				m_isDevelopment = true;
```
DomClark, line 695: This will only set `m_isDevelopment` if there are further lines after the two that this code looks for. I expect that this will usually be the case, but we shouldn't rely on it. I would instead set it in the sme block as `m_lmmsRcFile`, as that's all we care about for the config manager. The data directory code is irrelevant to this class, and only here so we don't need to duplicate the CMake cache scanning elsewhere.

src/core/ConfigManager.cpp
```diff
@@ -713,6 +701,55 @@ void ConfigManager::initDevelopmentWorkingDir()
...
+	// Detect < 1.2.0 working directory as a courtesy
+	if (QFileInfo(QDir::home().absolutePath() + "/lmms/projects/").exists())
+	{
+		m_workingDir = QDir::home().absolutePath() + "/lmms/";
+	}
+
+	// Determine RC file location to use if not in development
+	if (!m_isDevelopment)
+	{
+		QString cfgDir = QStandardPaths::standardLocations(QStandardPaths::AppConfigLocation).at(0);
```
DomClark, line 728: There are a couple of issues here:
* `standardLocations` isn't guaranteed to return any locations at all, so this may index the result out of bounds.
* We use the same directory both to read and to write the config file. The paths in `standardLocations` aren't guaranteed to be writable: they may include system-wide config locations, for example.
* In the case of XDG directories, relative paths are considered invalid. However, Qt makes no effort to omit these paths from the result.

I think the easiest way to address these is to use the writable location for configuration files instead. (It's possible for the writable location to be an empty string, or a relative path, in which case I'd just fall back to the legacy location.)

If you want to support reading configuration files from all possible locations, you could consider copying any such file found to the writable location first. You could also include the legacy location as one of these possible locations, which would offer a migration mechanism instead of a fallback mechanism.

M374LX: So, should we just fall back to the legacy path if ``standardLocations`` returns an empty or unwritable path?

M374LX:
Knowing of the plans to reduce the dependency on Qt and eventually removing it (as suggested by #6426 and #6472), a possible better idea is to use ``stat()`` and similar C functions and possibly also the Windows equivalents.

Does everyone agree?

BeatLink:
Are there any updates with this?



> Knowing of the plans to reduce the dependency of the core on Qt and eventually removing it (as suggested by #6426 and #6472), a possible better idea is to hard code each platform's standard path and use `stat()` and similar C functions, as well as the Windows equivalents.
> 
> Does everyone agree?

I think it would be a safer bet to use platform agnostic means to determine the configuration folder. Less code maintenance for you guys and better compatibility with non standard OSes, environments and distros. There must be a library out there that handles this for you already.

Rossmaxx:
Drafted because this is no longer being worked on.

PR #6008, opened 2021-05-09
editor couple of build warning

devnexen:


JohannesLorenz:
I just wonder if we should not make the vector contain floats instead of casting?

@Veratil what do you think? It's your code area :smiley:

Veratil:
> I just wonder if we should not make the vector contain floats instead of casting?
> 
> @Veratil what do you think? It's your code area 😃

I don't think we need the precision of doubles, as none of our zoom levels use greater than 7 digits of precision, so making them floats would be easier (should only be two lines in the header).

JohannesLorenz:
LGTM. Are the compiler warnings all gone now?

devnexen:
the remaining are mostly QT deprecations better not touching them I think.

Veratil:
Why does the compare have changes to rpmalloc and weakjack?

devnexen:
was unintended .

Veratil (approved):

JohannesLorenz (approved):

Issue #6014, opened 2021-05-11
Counterintuitive looping controls


lookitsnicholas:
relates to #4034 and #5505 - not really sure why this needs to be separate, tbh

### Bug Summary

As explained [here](https://github.com/LMMS/lmms/pull/4034#issuecomment-833164078), many people have expressed issues with the new control scheme for creating loops because it implicitly discourages changing parameters for larger projects due to the new additional amount of input required in order to do so.

#### Steps to reproduce

Create a loop of 100 measures, and try to move the loop end point to measure 50. Additionally, attempt to create a four-bar loop and move the end of the loop to the start of measure two.

#### Expected behavior

Something analogous to the original behavior, with one right-click/two-finger tap to end the loop and Shift+(right-click/two-finger tap) to start the loop.

#### Actual behavior

Weirdness due how the mean calculations are implemented... the video below shows the inability to end the loop at measure 50, and how incremental decrements of the loop length are currently required to shrink a four-bar loop into one. Zooming out is not an acceptable workaround because the resolution of said changes drops significantly, resulting in potentially more needed clicks.

#### Recording

https://user-images.githubusercontent.com/31457277/117765091-743d1580-b1fb-11eb-8594-7a2c58082df1.mov

#### Affected LMMS versions

Anything recent on 1.3-alpha

#### Feedback

My input on this is that this change is breaking mine and others' workflows because of the unnecessary amount of time clicking around and zooming in order to achieve the same result. A suggestion would be to either revert this behavior, add an option to toggle to eliminate extra button presses to accommodate the original behavior, or at the very least the addition of Shift modifiers (Shift+LMB for the start of a loop and Shift+RMB for the end of a loop) to get rid of the movement now required to quickly change loop parameters.

oldcastlehq:
I'm second to this. It's not only unintuitive, but it makes your work on LMMS harder.

Spekular:
> incremental decrements of the loop length are currently required to shrink a four-bar loop into one.

For the record this is false. Simply click anywhere on the right half to select the endpoint and then drag it left until the loop is one bar long.

lookitsnicholas:
Technically...? On a MacBook, that would consist of zooming out by scrolling on the zoom list box + Control+click + dragging the loop end point to the desired destination - that does not necessarily address the left point in that instance, which would also require zooming out. The point remains that that extra step of zooming out is now suddenly required when the previous implementation/other DAWs would allow for at most two-click operation.

[...]

Issue #5505, opened 2020-05-17
Setting new loop markers is complicated


JohannesLorenz:
### Bug Summary

We have loop markers, e.g. in the song editor, to only play a part of a song in a loop. That part will be colored green. Originally, you could use middle and left click to comfortably select the markers.

There must have been some PR that changed the behavior on master. Now, only the right click works, and it always sets that marker that is closest to your mouse.

I find this not practical. I often have large/all parts of my song selected in the loop (projects saved like that) and then want to reduce the loop to the first or last few bars. Typically because the last bars are where I'm currently continuing the song. This gets difficult:

```
<------------------------ currently selected ------------------------------>
                                                wanted selection <-----> 
```

Setting the left marker of "wanted selection" requires me to scroll through half the song! (The right marker is easy, ofc)

Aside from that, there is no tooltip explaining how the new behavior is to be used. But I'm for switching back to the old behavior.

#### Steps to reproduce

See sketch above.

#### Expected behavior

Not being forced to scroll through half of your song.

#### Actual behavior

Scrolling through half of your song.

#### Affected LMMS versions

Current master. not on `stable-1.2`

Spekular:
The old controls are unusable on devices without a MMB (every laptop I've
used), and surprisingly many hotkeys are already used for this area.

LMB: Set playhead position
Shift + LMB: Group select everything in this range
Ctrl + LMB: Nothing (Same as LMB)
RMB: Set loop pos
Ctrl + RMB: Set loop pos (fine adjust)
Shift + RMB: Nothing (Same as RMB)

Alternatively, perhaps we could rework these to be more consistent instead:

LMB: Playhead pos
RMB: Unused/reserved for menu

Ctrl + LMB: Group select
Ctrl + RMB: Unused/same as RMB

Shift + LMB: Loop Start (snapping)
Shift + RMB: Loop End (snapping)

Ctrl + Shift + LMB: Loop Start (fine)
Ctrl + Shift + RMB: Loop End (fine)

Another proposal was to grab based on.the visible area, so both start and
end position are always grabbable regardless of zoom/pan. However, I think
I might prefer reworking the shortcuts.

JohannesLorenz:
We could at least keep allowing the middle mouse key next to the new behavior, unless we plan to use it otherwise. That would at least help with the left marker.

> Shift + LMB: Loop Start (snapping)
> Shift + RMB: Loop End (snapping)

I like this even more. Tooltips would help to find out what the right shortcuts are.

[...]

PR #6021, opened 2021-05-18
Switch loop marker shortcuts, add context menu

Spekular:
See #5505 and #6014

This PR now halfway-implements remappable loop marker controls, with limitations:
- Unmodified LMB and RMB are reserved (playhead, context menu)
- All combinations of Shift/Control and LMB/RMB can be assigned an action. For applicable actions (everything except selecting TCOs), this will be quantized
- Shift + Control + click is used for the unquantized version of Shift + Click actions

#### Future enhancements

- Bindings are currently assigned in code, but should be exposed via the UI. I'm not sure it's wise to add this to the current settings menu, so this might involve rewriting it or adding a new menu for keybindings.
- Allow the loop section to be shifted as a whole. Could be part of the drag loop mode.
- Populate the context menu, it currently only contains "select between loop markers".
  - Fit loop markers to selection
  - Zoom to loop/selection/song
  - Quantize loop marker length, start, end, start + end
  - Double/halve loop length

#### Proposed default shortcuts:
LMB: Move playhead
RMB: Open context menu

Ctrl + LMB: Group select
Ctrl + RMB: Mean select

Shift + LMB: Loop Start
Shift + RMB: Loop End

Holding both modifiers disables snapping. If both are held at the start of the action, the shift action is selected.

### DIFF
--- a/src/gui/TimeLineWidget.cpp
+++ b/src/gui/TimeLineWidget.cpp
@@ -388,7 +388,7 @@ void TimeLineWidget::mousePressEvent( QMouseEvent* event )
 		const bool quant = !(mods == (Qt::ControlModifier | Qt::ShiftModifier));
 		
 		m_loopPos[0] = quant ? t.quantize(m_snapSize) : t;
-		m_loopPos[1] = t + (quant ? TimePos::ticksPerBar() : 1);
+		m_loopPos[1] = m_loopPos[0] + (quant ? m_snapSize : 1);
 	}
 	
 	// Ensure that the loops beginning and end are stored in the right place
@@ -423,7 +423,7 @@ void TimeLineWidget::mouseMoveEvent( QMouseEvent* event )
 	if (unquantized)
 	{
 		delete m_hint;
-		m_hint = NULL;
+		m_hint = nullptr;
 	}
 
 	switch (m_action)
--- a/src/gui/TimeLineWidget.cpp
+++ b/src/gui/TimeLineWidget.cpp
@@ -312,6 +312,17 @@ void TimeLineWidget::paintEvent( QPaintEvent * )
 void TimeLineWidget::contextMenuEvent(QContextMenuEvent*)
 {
 	QMenu contextMenu(tr("Timeline"), this);
+
+	// TODO: Shortcut hints should be read from a config
+
+	QAction setStartPoint(tr("Loop start (shift + left click)"), this);
+	connect(&setStartPoint, &QAction::triggered, this, [this](){
+		setLoopPoint(0, m_initalXSelect);
+	});
+	QAction setEndPoint(tr("Loop end (shift + right click)"), this);
+	connect(&setEndPoint, &QAction::triggered, this, [this](){
+		setLoopPoint(1, m_initalXSelect);
+	});
 	QAction selectLoopPoints(tr("Select between loop points"), this);
 	connect(&selectLoopPoints, &QAction::triggered, this, [this](){
 		emit regionSelectedFromPixels(
@@ -320,6 +331,8 @@ void TimeLineWidget::contextMenuEvent(QContextMenuEvent*)
 		);
 		emit selectionFinished();
 	});
+	contextMenu.addAction(&setStartPoint);
+	contextMenu.addAction(&setEndPoint);
 	contextMenu.addAction(&selectLoopPoints);
 	contextMenu.exec(QCursor::pos());
 }
@@ -455,7 +468,7 @@ void TimeLineWidget::chooseMouseAction(QMouseEvent* event)
 
 	// TODO: Read these from a config
 	auto leftCtrlAction = SelectSongTCO;
-	auto rightCtrlAction = MoveLoopClosest;
+	auto rightCtrlAction = NoAction;
 	auto leftShiftAction = MoveLoopBegin;
 	auto rightShiftAction = MoveLoopEnd;
 
--- a/src/gui/TimeLineWidget.cpp
+++ b/src/gui/TimeLineWidget.cpp
@@ -387,7 +387,7 @@ void TimeLineWidget::mousePressEvent( QMouseEvent* event )
 	{
 		const bool quant = !(mods == (Qt::ControlModifier | Qt::ShiftModifier));
 		
-		m_loopPos[0] = quant ? t.quantize(1.0) : t;
+		m_loopPos[0] = quant ? t.quantize(m_snapSize) : t;
 		m_loopPos[1] = t + (quant ? TimePos::ticksPerBar() : 1);
 	}
 	
@@ -447,7 +447,7 @@ void TimeLineWidget::mouseMoveEvent( QMouseEvent* event )
 		{
 			const int i = m_action == MoveLoopBegin ? 0 : 1;
 			if (unquantized) { m_loopPos[i] = t; }
-			else { m_loopPos[i] = t.quantize(1.0); }
+			else { m_loopPos[i] = t.quantize(m_snapSize); }
 			// Catch begin == end
 			if (m_loopPos[0] == m_loopPos[1])
 			{
@@ -468,7 +468,7 @@ void TimeLineWidget::mouseMoveEvent( QMouseEvent* event )
 		case DragLoop:
 		{
 			if (unquantized) { m_loopPos[1] = t; }
-			else { m_loopPos[1] = t.quantize(1.0); }
+			else { m_loopPos[1] = t.quantize(m_snapSize); }
 			update();
 			break;
 		}
--- a/src/gui/TimeLineWidget.cpp
+++ b/src/gui/TimeLineWidget.cpp
@@ -63,6 +63,7 @@ TimeLineWidget::TimeLineWidget( const int xoff, const int yoff, const float ppb,
 	m_xOffset( xoff ),
 	m_posMarkerX( 0 ),
 	m_ppb( ppb ),
+	m_snapSize( 1.0 ),
 	m_pos( pos ),
 	m_begin( begin ),
 	m_mode( mode ),
@@ -94,7 +95,9 @@ TimeLineWidget::TimeLineWidget( const int xoff, const int yoff, const float ppb,
 	updateTimer->start( 1000 / 60 );  // 60 fps
 	connect( Engine::getSong(), SIGNAL( timeSignatureChanged( int,int ) ),
 					this, SLOT( update() ) );
-					
+	
+	// Required to prevent modified right click from triggering the context menu
+	// Instead, manually trigger the context menu on unmodified right clicks
 	setContextMenuPolicy(Qt::PreventContextMenu);
 }
 
@@ -319,7 +322,7 @@ void TimeLineWidget::contextMenuEvent(QContextMenuEvent*)
 		emit selectionFinished();
 	});
 	contextMenu.addAction(&selectLoopPoints);
-	contextMenu.exec( QCursor::pos() );
+	contextMenu.exec(QCursor::pos());
 }
 
 
@@ -343,7 +346,8 @@ void TimeLineWidget::mousePressEvent( QMouseEvent* event )
 	if (event->x() < m_xOffset) { return; }
 	
 	// Choose an action based on input event and user's bindings
-	else if (button == Qt::LeftButton) {
+	else if (button == Qt::LeftButton)
+	{
 		if (mods & Qt::ShiftModifier){ m_action = leftShiftAction; }
 		else if (mods == Qt::ControlModifier){ m_action = leftCtrlAction; }
 		// Handles moving playhead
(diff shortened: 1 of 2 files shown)


Spekular:
I experimented a bit with rebinding support, but you'd either need to introduce severe restrictions or drastically rework the mouse event functions. In particular, the `m_action` system is an obstacle.

It's very easy to handle dragging the loop start, loop end, or playhead without it, just check mouse button and modifiers in the mouse move handler. Dragging with mean based logic should work without it if you always recalculate the closest point in the move event, but this could get hairy when the points approach each other. Selection needs state to remember where the selection started, and needs to emit selectionFinished when done.

In the end I decided it wasn't worth it, since there's not much difference between e.g. using ctrl or shift for an operation.

superpaik:
It works just fine, as stated in the PR description.

Some thoughts about it.
- Would it be possible to have an option to move the whole loop marker? I mean, if you have a loop marker of, say, 8 beats, move/drag the whole loop marker (the use case would be move the loop marker from one drop/break/built to the next) As a key/mouse option I'd say LMB inside the loop marker and drag it.
-An as for the LMB option (move playhead), I wonder why the default option it isn't snap to the grid but fine snapping (it's something not introduced by this PR, though, it happens in 1.2.2 as well)? To me it's kind of odd behaviour because you normally want to move the playhead to the start of a beat. I'm not aware right now if there is an alternative to do that. Maybe it's out of the scope of this PR and needs a debate about it.

Spekular:
I'd be very surprised to see the playhead quantized when moving. Even with mostly programmed stuff there are many elements that comes in slightly before grid lines (e.g. preshifted claps) that you can easily account for with an unquantized playhead. If you start working with recorded material things get even less perfectly aligned.

As for moving the loop, I agree that it'd be useful but I don't think unmodified LMB should do anything except playhead movement. If that's desireable then IMHO we should make a vertical division, so there's a playhead lane and a loop marker lane. Of course, that takes up more vertical space since you want the lanes to remain a comfortable height.

Maybe I need to give remapping some more thought.

qnebra:
I think playhead position shouldn't be automatically quantized, especially when manually moved.

Maybe Ctrl+Shift+LMB inside loop to move entire loop section?

superpaik:
I see your point regarding not quantizing the playhead, but I think that in any other use case the default behaviour is quantizing and you have an option for not quantizing it (like in the loop marker, midi/beat/sample track end, etc.) so thats why I said it, but it is ok to me, because I haven't notice it until I tested this PR and thought about how it should behave. As for context, Ableton quantize the playhead by default and with Alt+LMB you can move it freely. 

As for moving the whole loop the option Ctrl+Shift+LMB, as qnebra said, it's already taken for Loop Start (fine). The option Spekular propose would be perfect, but I guess it would require a major code change for just this functionality (unless is used for future functionalities like song markers). I don't know if ALT+LMB is an option because of cross platform issues, but another suggestion would be "doble-click" and move it until next click. Another option, which I think doesn't require a new lane, is that when inside (and only when inside) the loop marker and the cursor is at the top of the lane (above the playhead arrow) change the cursor to an double-arrow so you can move the loop with single LMB (see image). Perhaps with low resolution screens this may be an issue. 
imagen

Anyway, they where just thoughts that came to my mind while testing. Don't be too worried.

src/gui/TimeLineWidget.cpp
```diff
@@ -304,30 +307,43 @@ void TimeLineWidget::paintEvent( QPaintEvent * )
...
-		if( event->x() - m_xOffset < s_posMarkerPixmap->width() )
-		{
-			m_moveXOff = event->x() - m_xOffset;
-		}
-		else
-		{
-			m_moveXOff = s_posMarkerPixmap->width() / 2;
+		if (button == Qt::LeftButton) { m_action = MoveLoopBegin; }
+		if (button == Qt::RightButton) {
+			m_action = MoveLoopEnd;
```
IanCaio, line 338: Second conditional can be one-lined too

IanCaio (commented):
Still need to give a more thoughtful review and test it, but at first glance it looks good. There are some formatting changes I can later put on a review for a batch commit if you want. In this one I only addressed this conditional that can be one-lined to match the previous one.

Spekular:
Ok, I came up with a binding system that I think strikes a nice balance between flexibility and amount of code. I've reworked the PR quite a bit and as such edited the main post. There's at least one bug that needs squashing before merge (assistance here would be awesome), and ideally the re-binding support would be full fledged as well.

src/gui/TimeLineWidget.cpp
```diff
@@ -304,60 +307,103 @@ void TimeLineWidget::paintEvent( QPaintEvent * )
...
-	if( m_action == MoveLoopBegin || m_action == MoveLoopEnd )
+	// Notify the user if they can disable quantization
+	if (mods & Qt::ShiftModifier)
 	{
 		delete m_hint;
 		m_hint = TextFloat::displayMessage( tr( "Hint" ),
-					tr( "Press <%1> to disable magnetic loop points." ).arg(UI_CTRL_KEY),
-					embed::getIconPixmap( "hint" ), 0 );
+			tr( "Hold <%1> to disable quantization." ).arg(UI_CTRL_KEY),
+			embed::getIconPixmap( "hint" ), 0 );
```
IanCaio, line 404:
```suggestion
			tr("Hold <%1> to disable quantization.").arg(UI_CTRL_KEY),
			embed::getIconPixmap("hint"), 0);
```

Plus maybe having that hint box for Ctrl actions as well as suggested in the other comment

IanCaio (changes requested):
Mostly style changes and some doubts/suggestions. The code looks good, behavior is nice given some time to get used to it.

I found the "DragLoop" action name a bit misleading at first. Until I downloaded the build and tested it I thought that action was supposed to drag both loop points simultaneously (thus dragging the loop). I'd maybe rename it to "DragSelectLoop", which sounds more like what it does.

src/gui/TimeLineWidget.cpp
```diff
@@ -304,60 +307,103 @@ void TimeLineWidget::paintEvent( QPaintEvent * )
...
 	}
-	else if( event->button() == Qt::LeftButton  && (event->modifiers() & Qt::ShiftModifier) )
+	else if (button == Qt::RightButton) {
+		if (mods & Qt::ShiftModifier){ m_action = rightShiftAction; }
+		else if (mods == Qt::ControlModifier){ m_action = rightCtrlAction; }
+		else { contextMenuEvent(nullptr); }
+	}
+	
+	// Calculate the clicked position as a TimePos, several actions need this
+	m_moveXOff = s_posMarkerPixmap->width() / 2;
```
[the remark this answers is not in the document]
Spekular: It does, but as far as I can tell moving the playhead still moves perfectly fine. Frankly I have no idea what `m_moveXOff` is supposed to do or in what edge case it ends up not being half the pixmap width. I've tried clicking within the playhead triangle, and clicking so that it's by the start of the song, and everything works.

src/gui/TimeLineWidget.cpp
```diff
@@ -304,60 +307,103 @@ void TimeLineWidget::paintEvent( QPaintEvent * )
...
-			m_action = MoveLoopEnd;
-		}
-
-		if( m_loopPos[0] > m_loopPos[1]  )
-		{
-			qSwap( m_loopPos[0], m_loopPos[1] );
-		}
-
-		m_loopPos[( m_action == MoveLoopBegin ) ? 0 : 1] = t;
+		const bool quant = !(mods == (Qt::ControlModifier | Qt::ShiftModifier));
```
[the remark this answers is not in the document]
Spekular: That wasn't the intended behavior, but yeah, that's how it turned out. That seems fine to me so I'll update the tooltip instead of trying to prevent it.

src/gui/TimeLineWidget.cpp
```diff
@@ -367,13 +418,20 @@ void TimeLineWidget::mouseMoveEvent( QMouseEvent* event )
 {
 	parentWidget()->update(); // essential for widgets that this timeline had taken their mouse move event from.
 	const TimePos t = m_begin + static_cast<int>( qMax( event->x() - m_xOffset - m_moveXOff, 0 ) * TimePos::ticksPerBar() / m_ppb );
+	// Fine adjust when both ctrl and shift are held, hide ctrl+shift hint
+	bool unquantized = event->modifiers() == (Qt::ControlModifier | Qt::ShiftModifier);
+	if (unquantized)
+	{
+		delete m_hint;
+		m_hint = NULL;
```
IanCaio, line 426:
```suggestion
		m_hint = nullptr;
```

src/gui/TimeLineWidget.cpp
```diff
@@ -304,60 +310,105 @@ void TimeLineWidget::paintEvent( QPaintEvent * )
...
-	if( m_action == MoveLoopBegin || m_action == MoveLoopEnd )
+	// Notify the user if they can disable quantization
+	if (m_action != MovePositionMarker && m_action != NoAction)
 	{
 		delete m_hint;
-		m_hint = TextFloat::displayMessage( tr( "Hint" ),
-					tr( "Press <%1> to disable magnetic loop points." ).arg(UI_CTRL_KEY),
-					embed::getIconPixmap( "hint" ), 0 );
+		m_hint = TextFloat::displayMessage(tr("Hint"),
+			tr("Hold <%1> and <Shift> to disable quantization.").arg(UI_CTRL_KEY),
```
IanCaio, line 408: For consistency, is there a `UI_SHIFT_KEY` constant that should be used here too? I'm not sure, just asking, so if there's not just ignore this comment.

src/gui/TimeLineWidget.cpp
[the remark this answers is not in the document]
IanCaio: I see, it's probably not noticeable because the width is very small. We should probably decide between both ways though. If it's being overridden without apparent issue, maybe we should just delete the lines that calculate the offset when it's smaller than the width of the marker.

IanCaio (commented):
Sorry it took a while. On a second review nothing stood out besides those small comments!

src/gui/TimeLineWidget.cpp
Spekular: I'm pretty sure the constant exists from ctrl because macs call it command instead.

src/gui/TimeLineWidget.cpp
```diff
@@ -304,60 +310,105 @@ void TimeLineWidget::paintEvent( QPaintEvent * )
...
+	}
+	// If we're moving loop markers or dragging loop markers, update them
+	if (m_action == MoveLoopBegin){ m_loopPos[0] = t; }
+	else if (m_action == MoveLoopEnd){ m_loopPos[1] = t; }
+	else if (m_action == DragLoop)
+	{
+		const bool quant = !(mods == (Qt::ControlModifier | Qt::ShiftModifier));
+		
+		m_loopPos[0] = quant ? t.quantize(m_snapSize) : t;
+		m_loopPos[1] = t + (quant ? TimePos::ticksPerBar() : 1);
```
Spekular, line 391:
```suggestion
		m_loopPos[1] = m_loopPos[0] + (quant ? m_snapSize : 1);
```

IanCaio (approved):
I'd still rather we remove the irrelevant line (compiler might even be optimizing it out), but since it wasn't added by this PR particularly and everything else was fixed, I'm giving my approval!

src/gui/TimeLineWidget.cpp
```diff
@@ -304,60 +310,105 @@ void TimeLineWidget::paintEvent( QPaintEvent * )
...
+
 void TimeLineWidget::mousePressEvent( QMouseEvent* event )
 {
-	if( event->x() < m_xOffset )
-	{
-		return;
-	}
-	if( event->button() == Qt::LeftButton  && !(event->modifiers() & Qt::ShiftModifier) )
+	// TODO: Read these from a config
+	auto leftCtrlAction = DragLoop;
```
Spekular, line 334:
```suggestion
	auto leftCtrlAction = SelectSongTCO;
```

allejok96:
(To everyone but Spekular)

I've worked out a way to map RMB *click* to context menu, and RMB *drag* to something else. Now what would that something be?

- Moving the loop, like superpaik mentioned
- Drawing a loop, like you'd draw a selection with Ctrl+LMB

**The most desired of these two should be mapped to RMB drag**. The other one could be mapped to mid button. Some people have no mid button, but no big deal, you can still set left/right manually.

Personally I'd love the ability to quickly draw a loop. But I have a mouse, so any case is fine by me.

I also suggest we drop the "mean select" since it made redundant by having "select left" and "select right".

lookitsnicholas:
Personally, I think that moving the loop would make a bit of sense - as the multitude of modifiers used in this area performing the same behaviour could potentially be a bit confusing; am curious to hear other opinions, however.

I agree on the drop due to redundancy - for some reason, I was under the assumption that that was already completed within this PR due to its creation.

lookitsnicholas:
is there a way to test this release with a macOS 10.14+ package? haven't seen @LmmsBot get invoked at any time after the initial commit, and can't compile these changes locally atm

lookitsnicholas:
Can someone rerun this build? I don't currently seem to be able to do so haha

Spekular:
The "remapping" part of this is a half measure, so I'm closing it in favor of a later PR that won't try to future-proof.

PR #6035, opened 2021-05-30
Small independent changes from the Microtuner PR

he29-net:
Some small bits and pieces from #5522. Should be easy / trivial to review.


src/core/SampleBuffer.cpp
```diff
@@ -73,7 +74,7 @@ SampleBuffer::SampleBuffer() :
 	m_loopEndFrame(0),
 	m_amplification(1.0f),
 	m_reversed(false),
-	m_frequency(BaseFreq),
+	m_frequency(DefaultBaseFreq),
```
PhysSong, line 77: I found that there are both `BaseFreq` from `include/Mixer.h` and `DefaultBaseFreq` from `include/Note.h`. Should we consider consolidating them?

he29-net: `BaseFreq` is removed in the main PR, here I just included some independent files so the change is a bit out of context.

PR #6046, opened 2021-06-09
Simplify an if-else

Sawuare:
To avoid an unnecessary comparison with `false` and some code duplication.


JohannesLorenz (approved):
Functional and style OK.

Issue #416, opened 2014-03-05
LinuxVST support


diizy:
Again, another thing I'm just putting on the issue tracker, since it's been discussed on the mailing lists.

LinuxVST support should be relatively easy to implement considering we already have the VST code, and it would be beneficial, seeing as there's lots of interesting developments happening on the LinuxVST front right now (many new plugins being made and such).

And it's also a feature that has been requested by many people.

tresf:
Bumping out to 1.3 (likely after 2.0)

hallogallo:
Are there any news regarding this topic? Google just brought me here :laughing:

jasp00:
>Are there any news regarding this topic?

A volunteer to code would be appreciated. Also, we should merge some fixes for Windows VST first.

RichardJECooke:
I'm trying LMMS for the first time on Ubuntu and am unable to get VSTs to work - either Wine crashes trying to install them, or once installed LMMS' Vestige crashes trying to open them.

I hoped returning to Linux audio after seven years things would be much easier, but it still seems unusable. I really hope the team manages to get round to this 2014 issue some time.

tresf:
This topic is about LinuxVST, so the Wine crashes seem to be misplaced.  As for the state of Linux audio in general, this tracker isn't the place to discuss that despite many people sharing your feelings on the topic.

That said, although the frustrations you are having are real, we really don't tolerate people coming on this tracker pointing out the shortcomings of a few volunteers.

Please keep posts on-topic to the original bug report. 👍

vlad1777d:
Would be VST effects included into this issue ?
(I mean plugins, which are loaded after "add effect" button pressed)

davidhedlund:
I've spent 5 hours to update https://lmms.io/wiki/index.php?title=Tested_VSTs

I started https://lmms.io/wiki/index.php?title=Tested_VSTs#Free_software and added three VST's to under that section.

davidhedlund:
http://linux-sound.org/linux-vst-plugins.html is the most complete list of VST's for GNU/Linux that I've seen.

mrshroomy:
Why is this issue still open??

tresf:
> Why is this issue still open??

Because LinuxVST support has not yet been added.

mrshroomy:
> > Why is this issue still open??
> 
> Because LinuxVST support has not yet been added.

Does the issue mean that Linux should be able to fully support every single VST?

lukas-w:
[...]

Issue #5108, opened 2019-08-01
Support audio processing with more than 2 channels


PhysSong:
Some plugins require more than 2 channels for its audio input/output. Here are some common use-cases:
- Stereo sidechain compressor: requires 2 stereo inputs
- Drum kits: some drum kit VSTs can use multiple outputs for multiple samples, for examples, stereo out 1 for kick, stereo out 2 for snare

However, there are some difficulties which come from the internal structure of the LMMS core:
- LMMS uses interleaved stereo formats while external plugins like VST, LADSPA, Zyn, and LV2 expect non-interleaved format.
For audio outputs and file outputs, the interleaved form might be beneficial. However, I think it's not for internal processing.

Features that can be implemented:
- Multi-channel matrix effect, as an extension of the current stereo matrix plugin
- Better support for sidechain compressor or vocoder plugin, see #3817 and #4337
- Native drum sampler(#4075) with per-sample effect mixing

JohannesLorenz:
The question is also how to visualize that. I find the classical mixer window not suited for more than 2 channels, but I'd prefer a graph where each plugin is a node and you can connect to the inputs by using arrows ("edges"). What do the others think about the visualization?

> LMMS uses interleaved stereo formats while external plugins like VST, LADSPA, Zyn, and LV2 expect non-interleaved format.

True, but why is this more difficult with 6 channels than with 2?

DomClark:
Another issue I can see is fitting this around the instrument's built-in effect chain, which is only able to process a single stream of audio. Should we offer the option of multiple effect chains, one for each stereo channel pair, or should we disable the effect chain for multiple-output instruments, in the same way we disable envelopes for single-streamed instruments?

> What do the others think about the visualization?

I like the idea of a graph, and this seems to be a feature that other DAWs offer (FL Studio's Patcher comes to mind, but I'm sure there are others). I don't think this should be a replacement for the current mixer though - it could easily get disorganised and unmanageable for large projects.

Spekular:
A sane default would be to apply the same chain to each stream, no? Additional flexibility would be great too, of course, but being unable to use effects on a plugin at all seems rather limiting.

DomClark:
[...]

Issue #3975, opened 2017-11-15
Add native VST support for Mac OS X (WIP)


follower:
### Background

I had originally assumed that the reason for LMMS using WINE for VST (non-)support on Mac was because there weren't sufficient native Mac VSTs to justify adding native support. After playing with [Helio Workstation](https://github.com/peterrudenko/helio-workstation/) I discovered that there's plenty of native Mac VSTs so I decided to investigate adding support for them to LMMS.

### Progress so far

I've reached a point[1] where I think I've made sufficient progress that I wanted to keep track of development here--meaning that if I don't get to finishing it hopefully someone else can build on it without having to re-learn all of my lessons. :) (I've discovered a lesson I just learned was already in an unmerged WIP PR from someone else. :) )

The two parts of progress so far are:

 *  Proof-of-concept for opening Mac native VST within Qt 5.5 application: https://github.com/follower/qt-vst-mac-native-poc (which might also be suitable for a Qt-based mostly cross-platform implementation).

 * Compilation of sufficiently hacked up versions of `vst_base`, `vestige` and `RemoteVstPlugin` that LMMS will startup, show the Vestige plugin, allow it to be dragged onto a track, display a file open dialog box, allow selection of an actual binary of a VST renamed with a `.dll` extension and then display the following error messages (without actually crashing):
```
    RemotePluginClient::shmget: No such file or directory
    RemoteVstPlugin.cpp::shmget: No such file or directory
    RemoteVstPlugin.cpp: Failed to initialize shared memory for VST synchronization.
```

The final lesson I needed I eventually discovered had already been mentioned here: https://github.com/LMMS/lmms/pull/2393/commits/d7a63c5adda5d3878f033569453fed44e699e7e9#diff-0cfe54b7b7513b8854b14b3fb8411b78R115 (but, hey, at least I learned something along the way :) )

[[1] *Context: I started writing up this issue maybe a couple of weeks ago and wanted to make it more "complete" before posting it but since that approach doesn't seem to be working I thought I'd just dump it in here now as is and then add to it as I get the opportunity.*]

follower:
Today's additional progress was what led me to decide to actually post what I've got so far...

### Displaying a VST's GUI

I've managed to get multiple VST plugins to display their functional GUI within a standalone Qt application. What's particularly nice is that the approach I'm using doesn't require any Objective-C code to be written, we just need to pass around an opaque value retrieved via `QWidget::winId()`.

[...]

PR #6048, opened 2021-06-09
Native linux VST support

akimaze:
Hello,
With this changes finally we can use Linux VST(2) directly from Vestige. And yes ZynAddSubFX 3.0 works great also :)

Fixes https://github.com/LMMS/lmms/issues/416

I know LMMS is in reorganization but that was just experiment to answer question  "Can I do that?" And goes so good so I made PR ;) I know there is Carla Rack but I think a lot of users will like that feature (In Carla Rack you need open Carla UI and next plugin UI to change something what is a little inefficient (IMO)).

## Limitations
- only instrument support, but code for effects can be easily done (but in other PR)
- currently only no embed plugin window support

## How to use
- set VST plugin embedding to "No embeding"
- in vestige open window change file type to "SO-files" .so
- select plugin .so file

open plugin

## What works?
All needed features like show plugin UI, save/load presets (and preview in lmms browser), automation. 
Some mono synth like Dexed works only in left channel but that is out of scope of this PR.

I tested a lot of synths (most of them from kxstudio repos (usr/lib/vst folder after installation) and some commercial) and works OK:
- Vital
- ZynAddSubFX
- Surge
- Helm
- Tal Noise Maker
- Obxd
- Biotek 2
- CollaB3
- Collective
- Carla Rack

Synths with issues:
- Dexed - has only one output (mono)

## The code
All native linux related code is in `NATIVE_LINUX_VST`.  I'm a busy man so feel free to fix small issues, fork this code or do whatever is needed to merge/make better solution. I didn't noticed any thread issues but maybe it should be done better ;)

Some inspirations how to do X11 code from: https://github.com/ekenberg/vstminihost

## Screenshots

synths_lmms_native_vst

synths2


wujekbrezniew:
Does Carla Patchbay work with this?

PhysSong:
I'm still reading the code, but I think:
- The `#ifdef`s are making the code hard to read. If we decide to support more platforms(such as macOS), that might become even worse.
Ideally, platform-dependent code should be decoupled from the core code(and probably IPC stuff as well, but it's not relevant here).
- For native VSTs, Qt is also available(and it's a cross platform framework). You may find relevant discussions in #3975.

akimaze:
> Does Carla Patchbay work with this? And how automation works?

I never used Carla Patchbay so I can't test it properly but it loads. Below first message you can download app image and test it :)

akimaze:
Sorry for delay, 

> * The `#ifdef`s are making the code hard to read. If we decide to support more platforms(such as macOS), that might become even worse.
>   Ideally, platform-dependent code should be decoupled from the core code(and probably IPC stuff as well, but it's not relevant here).
> * For native VSTs, Qt is also available(and it's a cross platform framework). You may find relevant discussions in #3975.

I can agree with you. However, at this point I am not able to do such deep RemotePlugin refactoring. Designing this API is rather a task for someone who knows LMMS code better (and I don't think it has to be done in the first implementation). Maybe when someone implement MacOS we will know better how the API should look like. And then we just can refactor the code. 

Doing code refactor at this point can also introduce errors in windows VST handling. And a good testing can turn this PR (and adding other platform support) into a really big task that will last for years. Therefore, one of my assumptions was to not change the Windows code as much as possible. And you can check that nothing changes - only one calling from class to object.

Another advantage is that you can easily compare implementations for different platforms (just scroll up/below). And the skeleton remains common.

plugins/vst_base/VstPlugin.cpp
```diff
@@ -129,12 +132,20 @@ VstPlugin::VstPlugin( const QString & _plugin ) :
...
-	try {
-		PE::FileInfo peInfo(m_plugin);
-		machineType = peInfo.machineType();
-	} catch (std::runtime_error& e) {
-		qCritical() << "Error while determining PE file's machine type: " << e.what();
-		machineType = PE::MachineType::unknown;
+	QFileInfo fi(m_plugin);
+#ifdef LMMS_BUILD_LINUX
+	if (fi.completeSuffix() == "so")
+		machineType =PE::MachineType::linux64;
```
PhysSong, line 138: PE means Portable Executable format for Windows binaries such as `.dll`, `.exe`. Having `linux` in `PE` namespace doesn't sound good, but it's not a big issue until this pull request is ready to merge.

plugins/vst_base/RemoteVstPlugin.cpp
```diff
@@ -122,10 +149,14 @@ class RemoteVstPlugin;
...
 
+#ifndef NATIVE_LINUX_VST
 HWND __MessageHwnd = NULL;
 DWORD __processingThreadId = 0;
+#else
+pthread_t __processingThreadId = 0;
+#endif
 
-
+#ifndef NATIVE_LINUX_VST
```
PhysSong, line 159:
```suggestion
#ifdef _WIN32
```

plugins/vst_base/RemoteVstPlugin.cpp
```diff
@@ -402,6 +463,10 @@ class RemoteVstPlugin : public RemotePluginClient
 
 	bool m_processing;
 
+#ifdef NATIVE_LINUX_VST
+	pthread_mutex_t message_mutex = PTHREAD_MUTEX_INITIALIZER;
```
PhysSong, line 467: Why not using `std::mutex`?

plugins/vst_base/RemoteVstPlugin.cpp
```diff
@@ -577,27 +646,39 @@ bool RemoteVstPlugin::processMessage( const message & _m )
...
 			}
 			else
 			{
 				showEditor();
 			}
+
 			return true;
 
 		case IdIsUIVisible:
+			debugMessage("is visible?");
```
PhysSong, line 665: Don't forget to remove some unnecessary `debugMessage` calls once you're done.

plugins/vst_base/RemoteVstPlugin.cpp
```diff
@@ -876,12 +1048,20 @@ bool RemoteVstPlugin::load( const std::string & _plugin_file )
...
+		mainEntry = (mainEntryPointer) dlsym(m_libInst, "main");
+#endif
+	
 	if( mainEntry == NULL )
 	{
 		debugMessage( "could not find entry point\n" );
 		return false;
 	}
-
+
```
PhysSong, line 1064: Remove trailing tabulator.
```suggestion

```

plugins/vst_base/RemoteVstPlugin.cpp
```diff
@@ -2105,10 +2367,52 @@ DWORD WINAPI RemoteVstPlugin::guiEventLoop()
...
+			__plugin->idle();
+		
+		if(__plugin->isInitialized() && !__plugin->isProcessing() )
+		{
+			__plugin->processUIThreadMessages();
+		}
+		
+		struct timespec tim, tim2;
+		tim.tv_sec = 0;
+		tim.tv_nsec = 5000000;
```
PhysSong, line 2402: You can pull out this part out of the loop.

plugins/vst_base/RemoteVstPlugin.cpp
```diff
@@ -2251,6 +2559,13 @@ int main( int _argc, char * * _argv )
 		}
 	}
 
+#ifdef NATIVE_LINUX_VST
+	if (EMBED)
+	{
+		cerr << "Native linux VST works only without embeding." << endl;
```
PhysSong, line 2565: Have you tested this?

plugins/vst_base/RemoteVstPlugin.cpp
PhysSong: If I understood correctly, this mutex is required because `m_messageList` is directly used for transferring messages from processing thread to UI thread. Is this correct?
@DomClark Compared to the `PostMessage` approach we're using on Windows, does this method have any (dis)advantages? If this way works better on Windows, can we consider having a consistent implementation on all platforms?

akimaze: > Why not using `std::mutex`?

I just used pthreads to create thread, so I used pthread everywhere. No special meaning. 

> If I understood correctly, this mutex is required because `m_messageList` is directly used for transferring messages from processing thread to UI thread. Is this correct?

Yes, `RemoteVstPlugin::processUIThreadMessages()` can be called from main thread (ui loop) and `queueMessage()` can be used in main or worker thread (`processingThread()`). `std::queue` is not thread save.

include/aeffectx.h
```diff
@@ -29,7 +29,11 @@
 #include <stdint.h>
 
 // Calling convention
+#ifndef NATIVE_LINUX_VST
```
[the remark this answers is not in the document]
akimaze: @PhysSong  Should I decide/commit or this is for second reviewer? Looks OK for me ;)

plugins/vst_base/VstPlugin.cpp
akimaze: Yeah, I forgot about it, it was a hack done at the beginning to quickly see if it works :) Maybe should be changed to  `ExecutableFormat`?

plugins/vst_base/RemoteVstPlugin.cpp
akimaze: Yes, you can't open or close plugin window then.

plugins/vst_base/VstPlugin.cpp
PhysSong: Let's leave this for now.
(Don't resolve this conversation now!)

plugins/vst_base/RemoteVstPlugin.cpp
akimaze: Removed

DomClark (commented):
Here's my review of this. I don't have any experience with X11, so I'm assuming that those bits of code are correct (or at least tested and working). I've got some things I'd like to work on regarding VST support, so it would be great if we could get this merged soonish.

include/aeffectx.h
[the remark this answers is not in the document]
DomClark: Looks good to me too. 👍

plugins/vst_base/VstPlugin.cpp
DomClark: The `PE::MachineType` enum corresponds to these values here: https://docs.microsoft.com/en-gb/windows/win32/debug/pe-format#machine-types. I think it should be kept for the purpose of reading PE files, and we should introduce a separate enum to indicate which remote plugin host should be used.

For example:
```cpp
enum class ExecutableFormat
{
	Unknown, Win32, Win64, Linux64, // Maybe one day Linux32, Mac64, MacArm, etc.
};

// ...

auto format = ExecutableFormat::Unknown;

// ...

switch (peInfo.machineType())
{
	case PE::MachineType::amd64: format = ExecutableFormat::Win64;
	case PE::MachineType::i386: format = ExecutableFormat::Win32;
}

// ...

if (fi.suffix() == "so") { format = ExecutableFormat::Linux64; }
```

plugins/vst_base/RemoteVstPlugin.cpp
DomClark: This seems a reasonable alternative to `PostMessage` for now. Ideally we wouldn't use either as we aim for realtime safety - the pthreads version here has an explicit mutex, and `PostMessage` takes a lock internally. I plan to start working on this and other VST reworking/refactoring once this PR has been merged. Since we already have working Windows code, and this PR already has working Linux code, I don't think it's worth the effort to design a cross-platform lock-based system now, just to replace it with another design a little later.

DomClark:
@akimaze Are you able to work on this soon, or if not, do you mind if we finish it off?

akimaze:
> @akimaze Are you able to work on this soon, or if not, do you mind if we finish it off?

Sorry, I haven't free time currently to do that, I'll be glad if you finish this PR :)

akimaze:
> Here's my review of this. I don't have any experience with X11, so I'm assuming that those bits of code are correct (or at least tested and working). I've got some things I'd like to work on regarding VST support, so it would be great if we could get this merged soonish.

@DomClark I'm not X11 programmer also. X11 code is inspired by https://github.com/ekenberg/vstminihost . I wrote it after reading this sources (license "use it as you wish"). 
As @PhysSong mentioned it can be replaced by Qt code.

plugins/vst_base/RemoteVstPlugin.cpp
akimaze: From my tests, `PostMessage` uses much less resources. Try adding out of 40 tracks with linux VST (the nanosleep message loop consumes the most of CPU time). I think in the future there should be only one thread for all VST plugins but that requires lots of changes in VST support (or better solution than nanosleep in guiEventLoop).

guglielmofratticioli:
nice one ! 
when I compiled myself on Debian make gave the error in /src/core/MemoryManager.cpp  line: 53
 ```
MemoryManager::ThreadGuard::~ThreadGuard()
{
	if (--thread_guard_depth == 0) {
		rpmalloc_thread_finalize();
	}
}
```

not enaught input arguments for function rmpalloc_thread_finalize();

in the main lmms repo that is called with 'true' arg ->  `rpmalloc_thread_finalize(true);`

File: plugins/vst_base/RemoteVstPlugin/CMakeLists.txt [hunk 9]
```diff
@@ -10,9 +10,9 @@ CHECK_CXX_DEFINE(IS_WIN "_WIN32")
 CHECK_CXX_DEFINE(IS_WIN64 "_WIN64")
 CHECK_CXX_DEFINE(IS_MINGW "__MINGW32__")
 
-if(NOT IS_WIN)
-	message(FATAL_ERROR "Toolchain used does not target windows.")
-ENDIF()
+#if(NOT IS_WIN)
+#	message(FATAL_ERROR "Toolchain used does not target windows.")
+#ENDIF()
```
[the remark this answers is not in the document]
DomClark: I couldn't get this to work so I just got rid of it instead.

include/RemotePlugin.h
```diff
@@ -825,6 +825,16 @@ class LMMS_EXPORT RemotePlugin : public QObject, public RemotePluginBase
...
 
+#ifndef NATIVE_LINUX_VST
+	void setQuit()
+	{
+
+		lock();
+		sendMessage( IdQuit );
+		unlock();
+	}
+#endif
```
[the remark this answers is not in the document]
DomClark: Everything still seems to work fine without this, so I got rid of it.

plugins/vst_base/NativeLinuxRemoteVstPlugin64.cmake
```diff
@@ -0,0 +1,11 @@
+IF(LMMS_BUILD_LINUX)
+	ExternalProject_Add(NativeLinuxRemoteVstPlugin64
+		"${EXTERNALPROJECT_ARGS}"
+		CMAKE_ARGS
+			"${EXTERNALPROJECT_CMAKE_ARGS}"
+			"-DCMAKE_CXX_FLAGS=-m64 -DNATIVE_LINUX_VST"
+            "-DCMAKE_C_FLAGS=-DNATIVE_LINUX_VST"
```
Veratil, line 7: This line looks to be spaces and not tabs like the others

plugins/vst_base/RemoteVstPlugin.cpp
```diff
@@ -52,12 +52,39 @@
 #include <sched.h>
 #endif
 
+#ifndef NATIVE_LINUX_VST
 #include <wine/exception.h>
+#endif
 
 #endif
 
+#ifndef NATIVE_LINUX_VST
```
Veratil, line 61: Throughout this file are `#ifndef NATIVE_LINUX_VST` which as far as I can tell are just `#ifdef _WIN32` areas. Is there a reason it was done with `ifndef`?

Not saying we need to change it, just that it's odd to look a guard saying `LINUX` and seeing Windows stuff, then seeing it's a `ifndef` instead.

plugins/vst_base/RemoteVstPlugin.cpp
```diff
@@ -812,6 +909,46 @@ void RemoteVstPlugin::initEditor()
 	// 64-bit versions of Windows use 32-bit handles for interoperability
 	m_windowID = (intptr_t) m_window;
 #endif
+
```
Veratil, line 912: The above `#endif` is a wrapper for Linux/Window build. Since this is now part of a `#ifndef NATIVE_LINUX_VST` section is that needed anymore? Assuming `NATIVE_LINUX_VST` already means we have `LMMS_BUILD_LINUX` defined.

plugins/vst_base/RemoteVstPlugin.cpp
```diff
@@ -812,6 +909,46 @@ void RemoteVstPlugin::initEditor()
...
+#else
+	// native linux support
+	// inspired by https://github.com/ekenberg/vstminihost
+	XEvent e;
+	Atom prop_atom, val_atom;
+	
+	if (m_display == nullptr)
+		m_display = XOpenDisplay(nullptr);
```
Veratil, line 920: Please wrap the if in `{}`

plugins/vst_base/RemoteVstPlugin.cpp
```diff
@@ -839,31 +994,47 @@ void RemoteVstPlugin::hideEditor() {
...
 	}
 
 	pluginDispatch( effEditClose );
+#ifndef NATIVE_LINUX_VST
 	// Destroying the window takes some time in Wine 1.8.5
 	DestroyWindow( m_window );
 	m_window = nullptr;
+#else
+	if (m_display)
+		XCloseDisplay(m_display);
```
Veratil, line 1009: Please wrap the if in `{}`

plugins/vst_base/RemoteVstPlugin.cpp
```diff
@@ -2056,24 +2304,34 @@ DWORD WINAPI RemoteVstPlugin::processingThread( LPVOID _param )
...
 	// notify GUI thread about shutdown
+#ifndef NATIVE_LINUX_VST
 	PostMessage( __MessageHwnd, WM_USER, ClosePlugin, 0 );
 
 	return 0;
-}
-
+#else
+	if (m.id == IdQuit)
+		_this->queueMessage( m );
```
Veratil, line 2325: Please wrap the if in `{}`

plugins/vst_base/RemoteVstPlugin.cpp
```diff
@@ -2105,10 +2364,45 @@ DWORD WINAPI RemoteVstPlugin::guiEventLoop()
...
+	XEvent e;
+	while(true)
+	{
+		//if (XQLength(m_display) > 0)
+		if (m_display && XPending(m_display) > 0)
+		{
+			XNextEvent(m_display, &e);
+		
+			if (e.type == ClientMessage && e.xclient.data.l[0] == m_wmDeleteMessage)
+				hideEditor();
```
Veratil, line 2383: Please wrap the if in `{}`

plugins/vst_base/RemoteVstPlugin.cpp
```diff
@@ -2105,10 +2364,45 @@ DWORD WINAPI RemoteVstPlugin::guiEventLoop()
...
+		}
 
+		// needed by zynaAddsubfx ui
+		if (__plugin->isInitialized())
+			__plugin->idle();
+		
+		if(__plugin->isInitialized() && !__plugin->isProcessing() )
+			__plugin->processUIThreadMessages();
```
Veratil, line 2391: Please wrap the ifs in `{}`

plugins/vst_base/VstPlugin.cpp
```diff
@@ -128,23 +132,49 @@ VstPlugin::VstPlugin( const QString & _plugin ) :
...
+			case PE::MachineType::amd64:
+				pluginType = ExecutableType::Win64;
+				break;
+			case PE::MachineType::i386:
+				pluginType = ExecutableType::Win32;
+				break;
+			default:
+				qWarning() << "Unknown PE machine type"
+					<< QString::number(static_cast<uint16_t>(peInfo.machineType()), 16);
+				break;
```
Veratil, line 158: Not sure about others, but I prefer wrapping case blocks in `{}`

Veratil (changes requested):
Along with these comments I'd say following the style conventions (such as no spaces before/after the parens in things like `if (...)`) unless other devs are okay with ignoring them in favor of auto styling later.

plugins/vst_base/RemoteVstPlugin.cpp
DomClark: Yes - we support Windows VSTs on Linux using Wine, and the remote plugin host for these is built with `LMMS_BUILD_LINUX` defined (since it's for a Linux system), but not `NATIVE_LINUX_VST` (since it's for hosting Windows VSTs). The different behaviour is needed so that on Windows we get an `HWND`, but on Linux we get Wine's underlying X11 window handle.

plugins/vst_base/RemoteVstPlugin.cpp
DomClark: PhysSong suggested changing a few of them to `_WIN32` in an earlier review, which I have done, but that was for code where a Windows/non-Windows distinction made sense. In a lot of other cases, the distinction is Windows/Linux/future platform X/etc., where there's less clarity to be gained by singling out Windows. At any rate, I'm hoping to factor platform-specific code out of this file after this PR is merged, which should hopefully make the choice redundant, so I'm inclined to leave this unless you feel strongly otherwise.

Veratil: Don't feel strongly about it, mainly strong about easy readability. The `ifndef` makes you have to do an extra negation step in your head for the code in the block it's surrounding. If you'll be refactoring later then it's all good. Just trying to be better readable. 😁

plugins/vst_base/VstPlugin.cpp
DomClark: I personally prefer to leave them, unless the case block contains a variable declaration, in which case braces are often necessary. Shall we make this part of the clang-format discussion, and let the auto-formatter tidy it up later if we decide to use braces?

Veratil: That sounds good to me.

Veratil (approved):

plugins/vst_base/RemoteVstPlugin.cpp
```diff
@@ -2105,10 +2370,51 @@ DWORD WINAPI RemoteVstPlugin::guiEventLoop()
...
+		{
+			XNextEvent(m_display, &e);
+		
+			if (e.type == ClientMessage && e.xclient.data.l[0] == m_wmDeleteMessage)
+			{
+				hideEditor();
+			}
+		}
 
+		// needed by zynaAddsubfx ui
```
PhysSong, line 2394:
```suggestion
		// needed by ZynAddSubFX UI
```

plugins/vst_base/RemoteVstPlugin.cpp
```diff
@@ -812,6 +909,48 @@ void RemoteVstPlugin::initEditor()
 	// 64-bit versions of Windows use 32-bit handles for interoperability
 	m_windowID = (intptr_t) m_window;
 #endif
+
+#else
+	// native linux support
+	// inspired by https://github.com/ekenberg/vstminihost
```
PhysSong, line 915: I'm not sure what the license of the original project is, but shouldn't we mention the author information?

allejok96:
Compiled fine, seems to work fine, only complaint I have as a user is that I must select *.so instead of *.dll in the browser every time, which I probably wouldn't notice if I wasn't told to look for it and so I'd just assume that only dll files can be loaded.

probonopd:
> Some mono synth like Dexed works only in left channel but that is out of scope of this PR.

Is there be a follow-up ticket regarding this issue?

DomClark:
> Only complaint I have as a user is that I must select *.so instead of *.dll in the browser every time

I will address this in a future PR (also removing `*.exe` from that list because it makes no sense).

> > Some mono synth like Dexed works only in left channel but that is out of scope of this PR.
> 
> Is there be a follow-up ticket regarding this issue?

I would hope a solution to #5108 would include better handling of instruments with a single output channel, but we can also open an issue specifically for those instruments. In the meantime, it's easy to work around with a stereo matrix effect.

Issue #5734, opened 2020-10-25
FreeBoy - Division by zero


zonkmachine:
Division by zero. Debugging with the fpe flag `-DCMAKE_BUILD_TYPE=Debug -DWANT_DEBUG_FPE=True` gives a division by zero error in FreeBoy.

This is on stable-1.2 and master as of 7808e0b92fdb7f84180553fa534d66b2f2f041c6. It's probably introduced with the original FreeBoy(papu) commit.

#### Logs
<details>
  <summary>Click to expand</summary>
<pre>
Thread 13 "Mixer::fifoWrit" received signal SIGFPE, Arithmetic exception.
[Switching to Thread 0x7fffbd5bc700 (LWP 93992)]
0x00007fffbcd69f73 in FreeBoyInstrument::playNote (this=0x7ffff2bb0060, _n=0x7ffff2ae0e40, _working_buffer=0x7fffa2ca0860)
    at /home/zonkmachine/builds/lmms/plugins/FreeBoy/FreeBoy.cpp:374
374				f = 524288.0 / ( r * pow( 2.0, s + 1.0 ) );
(gdb) bt full
#0  0x00007fffbcd69f73 in FreeBoyInstrument::playNote(NotePlayHandle*, std::array<float, 2ul>*)
    (this=0x7ffff2bb0060, _n=0x7ffff2ae0e40, _working_buffer=0x7fffa2ca0860) at /home/zonkmachine/builds/lmms/plugins/FreeBoy/FreeBoy.cpp:374
        r = 0 '\000'
        s = 0 '\000'
        sopt = 0 '\000'
        ropt = 1 '\001'
        fopt = 262144
        f = 0
        tfp = 0
        samplerate = 44100
        frames = 255
        offset = 1
</pre>
</details>

https://github.com/LMMS/lmms/blob/7808e0b92fdb7f84180553fa534d66b2f2f041c6/plugins/FreeBoy/FreeBoy.cpp#L370-L377

We set 'r' to 0 in the second for loop and this gives `f = 524288.0 / 0`on line 374.

Something like this for line 374 could work:
`f = 524288.0 / std::max( r * pow( 2.0, s + 1.0 ), 0.000001 );`

It plays back similar on this demo project.
[freeboy.mmp.zip](https://github.com/LMMS/lmms/files/5435683/freeboy.mmp.zip)

zonkmachine:
It's possible to turn the floating point exceptions off temporarily.

```
#include <fenv.h>

...

[...]

PR #6053, opened 2021-06-13
Fixed issue #5734 (FreeBoy Division by zero)

aidan-mueller:
Added comments and used more descriptive variable names for noise
channel initialization block.

Also indented the nested for loop to improve code clarity.
The reasons for doing this can be found in this answer:
https://softwareengineering.stackexchange.com/a/362796


DomClark:
@PhysSong noted this in the associated issue:

> Also, `(r, s) = (0, n)` is equivalent to `(r, s) = (1, n - 1)`, so we don't have to check for `r = 0` unless `s = 0`.

Perhaps it would be simpler to use `r = 0` and `s = 0` for the initial guess outside the loops, then start the inner loop from `r = 1`?

piece-o-ham:
@DomClark You are correct. I actually saw that comment, but for some reason I didn't quite understand what he was saying until now (*facepalm*). I'm thinking of how we could fix this, and my first thought was to just leave it out, since `r = 0, s = 15` produces ~B10 (not even on the keyboard). However, I actually see that there is a free way of including even `r = 0, s = 0`. Look at line 368:

https://github.com/LMMS/lmms/blob/f8d7fa3b871ae79b11858d9a6226bdd38a2bee5b/plugins/FreeBoy/FreeBoy.cpp#L368

The loop starts with `s = 0, r = 1` as the default, which in then computes again (redundantly) inside the loop. If `r = 0, s = 0` was the starting choice, we could skip `r = 0` in the loop and would still cover all possible output values.

I'll add a commit to change this, and perhaps while I'm at it I'll also spend a bit of time trying to optimize further if I can.

PR #6059, opened 2021-06-18
Fixed issue 5364

ifndefJOSH:
Changed position of `QTreeWidget::mousePressEvent(me);` and encased rest of function in `if` statement

PR #6060, opened 2021-06-18
Fixed issue 5364

ifndefJOSH:
Changed position of `QTreeWidget::mousePressEvent(me);` and encased rest of function in `if` statement. Use this pull request instead of https://github.com/LMMS/lmms/pull/6059


Spekular:
There's no need to invert the if, that change is equivalent to the existing code. If you update this to just moving the event forwarding I'd approve it.

Spekular:
Nevermind, I see the issue. However, there's still a cleaner fix. Just place `QTreeWidgetItem * i = itemAt(me->pos());` at the start of the function (before the event is forwarded).

ifndefJOSH:
I've added a new commit which is a bit cleaner of a solution and reflects your suggestion (actually two commits since I missed moving a comment)

Spekular (approved):

PR #6065, opened 2021-06-22
sf2_player: Support the full 14-bit MIDI bank range

xTibor:
A small step towards #1472. This change is based on things described in: MIDI 1.0 Detailed Specification 4.2.1 - Channel Voice Messages - Bank Select.

I had to slightly move the widgets around to be able to fit in those two new bank selector digits. Tried to keep the GUI changes as minimal as possible.

**Screenshot:**


softrabbit:
> A small step towards #1472. This change is based on things described in: MIDI 1.0 Detailed Specification 4.2.1 - Channel Voice Messages - Bank Select.

Based on things in the Soundfont 2.04 specification, section 7.2, I'm not entirely sure this is needed. 
>If the value in either field is not a valid MIDI value of
zero through 127, or 128 for wBank, the preset cannot be played but should be maintained. 

https://soundprogramming.net/manuals/fileformats/SoundFont_2.04_Specification.pdf

xTibor:
Ok, I didn't notice that restriction. Disregard this PR. Regardless that bank value still needs adjustment from 0-999 to 0-128 then.

PR #6075, opened 2021-07-11
Speed up SF2 loading

allejok96:
Improves loading speed of the sf2 plugin + small bugfix

#### Bug
Select a large sf2 as the default soundfont. Create a project with lots of sf2 tracks with a different (small or missing) sf2. Try loading the project. Loading time is slow.

#### Fix
When loading a project file (or previewing a sf2) the default soundfont will not be loaded by the sf2 plugin. But manually adding a new sf2 track loads the default sf2. This may cut project load time by 90%.

#### Bug

Open a project with a missing sf2 file. Select the correct sf2. Instrument keeps playing patch 0 even if another patch is selected. Open and close the patch select dialog. The correct sound is now played.

#### Fix

`openFile()` now calls `updatePath()`

## Removal of redundant constructor code

I've tried my best to be careful here, but this needs double checking by some smart person. This is just a brief overview of some duplicate calls in the constructor.

#### Before

```py
Sf2Player()
  m_synth = new_fluid_synth()
  load_file()
     openFile()
        updateSampleRate()  # [1]
           m_synth = new_fluid_synth()
           updateReverbEtc()
  updateSampleRate()
     m_synth = new_fluid_synth()
     updateReverbEtc()
  updateReverbEtc()

Sf2Player::loadSettings()
  m_gain.loadSettings()      # connected to updateGain()
  m_patchNum.loadSettings()  # connected to updatePatch()
  updateGain()
  updatePatch()
```

[1] Comment states that "preview of a soundfont for the first time fails" if this is removed, but calling it before `openFile()` seems to work fine.

#### After
```py
Sf2Player()
  updateSampleRate()
    m_synth = new_fluid_synth()
    updateReverbEtc()
  load_file()
     openFile()

Sf2Player::loadSettings()
  m_gain.loadSettings()
  m_patchNum.loadSettings()
```


RiedleroD:
I waited so long for both of these fixes… Thank you! Once this gets merged, I'll absolutely use lmms from master.

Spekular:
> When loading a project file (or previewing a sf2) the default soundfont will not be loaded by the sf2 plugin.

This feels like a roundabout check. Doesn't it make more sense to only load the default soundfont if there's no other soundfont assigned? I assume the default soundfont is currently being loaded before that information is available, but that just seems like a good reason to move the default sf load to a later point.

allejok96:
It would be a good thing, but I don't know how. [TrackContainerView](https://github.com/LMMS/lmms/blob/master/src/gui/TrackContainerView.cpp#L379) calls InstrumentTrack::loadInstrument() which calls Instrument::instantiate(). In case of preset/project it calls loadFile().

How should Sf2Player detect *no* call to loadFile? Maybe a new method called Instrument::loadDefault() and calling that from TrackContainerView?

Spekular:
> How should Sf2Player detect _no_ call to `loadFile()`? Maybe a new method called `Instrument::loadDefault()` and calling that from `TrackContainerView::dropEvent()`?

I guess it's not as simple as I thought it'd be, since instrument loading differs from what I expected. I like that `loadDefault` is a relatively unintrusive alternative, but it feels like a weird change to make unless there are other instruments that would benefit.

Might be best to leave that line as is, and potentially expand on why it's necessary in the comment. Idea:
```
// We rely on instantiate() to load the default preset when adding instruments,
// but don't want to load the default soundfont otherwise
```

plugins/sf2_player/sf2_player.cpp
```diff
@@ -153,14 +153,12 @@ sf2Instrument::sf2Instrument( InstrumentTrack * _instrument_track ) :
...
-	updateReverb();
-	updateChorusOn();
-	updateChorus();
-	updateGain();
+	// FIXME: there's no good way to tell if we're loading a preset or an empty instrument
+	// We rely on instantiate() to load the default soundfont for new instruments,
+	// but we don't need that when loading a project/preset/preview
+	loadFile(Engine::getSong()->isLoadingProject() or instrumentTrack()->isPreviewMode()
+				? ""
+				: ConfigManager::inst()->sf2File());
```
DomClark, line 161: We prefer to use `||` over the `or` keyword. Also, it looks like `loadFile` won't do anything if passed an empty string, so perhaps it makes more sense not to call it at all? (It won't call `openFile` if the string is empty, and if no soundfont is loaded, the bank/patch selection loop won't run.)
```suggestion
	if (!Engine::getSong()->isLoadingProject() && !instrumentTrack()->isPreviewMode())
	{
		loadFile(ConfigManager::inst()->sf2File());
	}
```

DomClark (commented):
The constructor changes look good to me. I think the changes to `loadSettings` are fine too, but the justification is a little more subtle:
The update slots are connected to the `dataChanged` signal of the models, so if the loaded value is the same as the model's default value, there is no change and they won't be called. This is fine if fluidsynth's default value matches the model's default value, but ideally we wouldn't rely on that. Fortunately, `updateGain` is called from `updateSampleRate`, which is called from the constructor, and `updatePatch` is called from `openFile`, which is called earlier in `loadSettings`. Thus the default values will be set already, so there is no problem if they aren't set again.

Finally, I think we should rename `updateSampleRate` at some point, as it does far more than its name suggests (i.e. it reloads the whole plugin). This is out of scope here though, I think - your comment that you are relying on it for its side effects should be sufficient for now.

plugins/sf2_player/sf2_player.cpp
allejok96: > This is fine if fluidsynth's default value matches the model's default value, but ideally we wouldn't rely on that.

Good observation. This seems to be the case with `updateChorus` and `updateReverb` too, [when running Fluid version 1](https://github.com/LMMS/lmms/blob/ddff2eba9f430ea0cdde087ba4f08fdfa329d936/plugins/sf2_player/sf2_player.cpp#L133). But as you say, as long as everything gets updated inside `updateSampleRate` this is perfectly fine.

> I think we should rename updateSampleRate at some point

Why not do it right away? It's only used in two places. And this PR makes things more confusing by relying on it for everything. What would be a good name? `reloadSynth` ?

allejok96:
> This is fine if fluidsynth's default value matches the model's default value, but ideally we wouldn't rely on that.

Good observation. This seems to be the case with `updateChorus` and `updateReverb` too, [when running Fluid version 1](https://github.com/LMMS/lmms/blob/ddff2eba9f430ea0cdde087ba4f08fdfa329d936/plugins/sf2_player/sf2_player.cpp#L133). But as you say, as long as everything gets updated inside `updateSampleRate` this is perfectly fine.

> I think we should rename updateSampleRate at some point

Why not do it right away? It's only used in two places. And this PR makes things more confusing by relying on it for everything. What would be a good name? `reloadSynth` ?

allejok96:
@DomClark `updateSampleRate` -> `reloadSynth` plz

DomClark:
Sounds good to me. 👍

DomClark (approved):

PR #6076, opened 2021-07-12
Remove 'using namespace std;' from LMMS headers

irrenhaus3:
Removes "using namespace std" from global scope, which is a common source of symbol clashes. Adds std:: prefix to affected symbols in the header and some sources which included it transitively.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -6,7 +6,7 @@
 	url = https://github.com/mjansson/rpmalloc.git
 [submodule "plugins/zynaddsubfx/zynaddsubfx"]
 	path = plugins/zynaddsubfx/zynaddsubfx
-	url = https://github.com/lmms/zynaddsubfx.git
+	url = https://github.com/irrenhaus3/zynaddsubfx.git
 [submodule "plugins/FreeBoy/game-music-emu"]
 	path = plugins/FreeBoy/game-music-emu
 	url = https://bitbucket.org/mpyne/game-music-emu.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -6,7 +6,7 @@
 	url = https://github.com/mjansson/rpmalloc.git
 [submodule "plugins/zynaddsubfx/zynaddsubfx"]
 	path = plugins/zynaddsubfx/zynaddsubfx
-	url = https://github.com/irrenhaus3/zynaddsubfx.git
+	url = https://github.com/lmms/zynaddsubfx.git
 [submodule "plugins/FreeBoy/game-music-emu"]
 	path = plugins/FreeBoy/game-music-emu
 	url = https://bitbucket.org/mpyne/game-music-emu.git
--- a/include/lmms_math.h
+++ b/include/lmms_math.h
@@ -32,7 +32,6 @@
 #include <QtCore/QtGlobal>
 
 #include <cmath>
-using namespace std;
 
 #ifndef exp10
 #define exp10(x) std::pow( 10.0, x )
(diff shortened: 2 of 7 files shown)


Veratil:
LGTM, double check for any other locations!

PhysSong:
Isn't it possible to add `using std::foo;`s in place of `using namespace std;` instead of adding `std::` everywhere we need?

rdrpenguin04:
That can lead to just as many name conflicts, with the only difference being us opting into every one individually.

The people here have many things to say about this: [StackOverflow link](https://stackoverflow.com/questions/1452721/why-is-using-namespace-std-considered-bad-practice). While this often doesn't address `using std::*`, making use of any `using` declaration at header scope is nearly universally a bad idea for the same reasons. At implementation scope, it's fine, but it's generally clearer to specifically name when something is coming from `std` and when we have our own implementation.

irrenhaus3:
> That can lead to just as many name conflicts, with the only difference being us opting into every one individually.
> 
> The people here have many things to say about this: [StackOverflow link](https://stackoverflow.com/questions/1452721/why-is-using-namespace-std-considered-bad-practice). While this often doesn't address `using std::*`, making use of any `using` declaration at header scope is nearly universally a bad idea for the same reasons. At implementation scope, it's fine, but it's generally clearer to specifically name when something is coming from `std` and when we have our own implementation.

Agreed. `using [namespace]::[symbol] ` is OK inside a lexical scope of some function or class, but at global scope it's more trouble than it's worth. IMO just writing `std::foo` instead of `foo` isn't too big a hassle either, so my personal usage of `using std::foo` is pretty low anyways - but that's more of a matter of personal style.

As for other locations, `using namspace std;` appears at global scope in the following header files:

3rdparty/jack2:
- common/JackNetTool.h
- macosx/coreaudio/JackCoreAudioAdapter.h
- macosx/coreaudio/JackCoreAudioDriver.h   

plugins:
- zynaddsubfx/src/Effects/Alienwah.h
- zynaddsubfx/src/Tests/[several files]
- carlabase/carla/source/native-plugins/external/zynaddsubfx/UI/TipWin.h
- LadspaEffect/calf/veal/src/calf/orfanidis_eq.h

Most of these occurences might be contained to their respective projects - at least the Zyn Test header instances definitely are.

JohannesLorenz:
> plugins:
> 
>     * zynaddsubfx/src/Effects/Alienwah.h
>     * zynaddsubfx/src/Tests/[several files]

We could fix the Alienwah in our fork (though note that this has been fixed on zyn's upstream). The Test files are only a source for auto-generating cpp files using "cxxtest", and they will only be ever included by their autogenerated cpp files. So I'm OK with leaving those as they are.

This PR LGTM, but feel free to fix the occurences in the submodules mentioned - at least in those which we own (e.g. LMMS/zynaddsubfx).

PhysSong:
Note: The changes should be tested against GCC 6 to make sure it doesn't break builds. See #5831 for an example.

JohannesLorenz:
@PhysSong How do we test GCC6? Is this part of our CI, or must we ask one of our devs who has GCC6? I think @Veratil posted with GCC6 on discord?

irrenhaus3:
Update on veal: orfanides_eq.h does in fact not contain a `using namespace std` anymore, I was mistaken because my submodule was not up to date. Fun!
So with the Zyn patch, this removes `using namespace std` from global scope in all LMMS-owned header files except the self-contained Zyn tests.
Do not merge before the PR in the Zyn repo has been merged and the temp. URL for the submodule has been reverted.

JohannesLorenz:
@irrenhaus3 I merged the zyn submodule to its master. If you revert the remote, I think we are ready to go.

JohannesLorenz (approved):

JohannesLorenz (approved):

PR #6077, opened 2021-07-12
Fix cloning of B/B tracks

### LINKED ISSUE #6062: Crash when trying to clone B/B pattern when there is not any
Steps to reproduce:
- Open an empty project with no B/B tracks
- Open the B+B editor
- Click on Clone Pattern button

Expected behavior:
Nothing happens

Actual behavior:
Instant crash.

dj-pixus:
Fixes #6062.
Also the editor changes to the new track correctly now.

### DIFF
--- a/src/gui/editors/BBEditor.cpp
+++ b/src/gui/editors/BBEditor.cpp
@@ -330,7 +330,7 @@ void BBTrackContainerView::clonePattern()
 	{
 		// Clone the track
 		Track *newTrack = bbt->clone();
-		bbtc->setCurrentBB( dynamic_cast<BBTrack *>( newTrack )->index() );
+		bbtc->setCurrentBB( static_cast<BBTrack *>( newTrack )->index() );
 
 		// Track still have the TCOs which is undesirable in this case, clear the track
 		newTrack->lock();
--- a/src/gui/editors/BBEditor.cpp
+++ b/src/gui/editors/BBEditor.cpp
@@ -326,11 +326,15 @@ void BBTrackContainerView::clonePattern()
 
 	BBTrack *bbt = BBTrack::findBBTrack(cur_bb);
 
-	// Clone the track
-	Track *newTrack = bbt->clone();
+	if( bbt )
+	{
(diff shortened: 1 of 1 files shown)


src/gui/editors/BBEditor.cpp
```diff
@@ -326,11 +326,15 @@ void BBTrackContainerView::clonePattern()
...
-	Track *newTrack = bbt->clone();
-
-	// Track still have the TCOs which is undesirable in this case, clear the track
-	newTrack->lock();
-	newTrack->deleteTCOs();
-	newTrack->unlock();
+	if( bbt )
+	{
+		// Clone the track
+		Track *newTrack = bbt->clone();
```
Spekular, line 332:
```suggestion
		BBTrack *newTrack = bbt->clone();
```
I don't have my dev machine running right now, do you mind checking if this works? It probably won't, but if it does you can remove the cast on the next line.

src/gui/editors/BBEditor.cpp
```diff
@@ -326,11 +326,15 @@ void BBTrackContainerView::clonePattern()
...
+		bbtc->setCurrentBB( dynamic_cast<BBTrack *>( newTrack )->index() );
```
Spekular, line 333:
```suggestion
		bbtc->setCurrentBB( static_cast<BBTrack *>( newTrack )->index() );
```
I think you can use a static cast in this case. We know that `bbt` is a BBTrack, and since `newTrack` is cloned from `bbt` it should have the same type.

Spekular (commented):
Good fix, I left a few minor suggestions.

src/gui/editors/BBEditor.cpp
dj-pixus: Checked. Doesn't work because `clone()` always returns a `Track*`.
The static cast is fine though.

Spekular: Figures. Thanks for double checking.

Spekular (approved):
LGTM.

PR #6078, opened 2021-07-13
DrumSynth changes

softrabbit:
- Handle the config file in a slightly prettier way using QSettings and only reading the file once.
- Fix rare envelope bug that affected 4 out of 764 bundled sounds. Too long envelopes in .ds file were being cut off. This will affect any projects using those sounds. 
- Some code clean up.

This code has been tested against the bundled .ds files in my [DrumSynthLive branch](https://github.com/softrabbit/lmms/tree/DrumSynthLive/plugins/DrumSynthLive) so it should be pretty close to production ready.


softrabbit:
This has gone stale enough to abandon. Not going to even try to merge in the last 4 years of changes in master into this, and probably not going to retrace my steps from the current master in a new PR either.

PR #6080, opened 2021-07-13
Fixes incorrect BB tracks being shown in the BB Editor (#6079)

### LINKED ISSUE #6079: Changing time signature causes incorrect BB track to be displayed in the BBEditor
### Bug Summary
When time signature isn't n/n, BBEditor displays incorrect tracks. I tracked the bug to the TimePos::getBar() function call, which gives inaccurate results. Unfortunately fixing this problem will require some class restructuring, and given I am new to the project, I wasn't able to do so.

#### Steps to reproduce
Create multiple BB tracks, enter steps for clarity, change time signature.

#### Expected behavior
I would hope correct track would be displayed, which is not the case.

#### Actual behavior
Incorrect track is displayed. This is because the track number is calculated as m_ticks divided by m_ticksPerBar, where m_ticks doesn't get updated as time signature changes. That means that instead of track 0, track 1 is displayed when the time signature is 5/4 (because 192/240 is 0), etc..

#### Affected LMMS versions
Both stable and nightly.

#### Logs
<details>
<pre>

</pre>
</details>

sdasda7777:
Fixes issue 6079, that is incorrect BB tracks being shown in the BB Editor.
The fix achieves this by not using getBar for calculating the number of the track (as that doesn't work with different signatures than n/n), and instead gets track number by dividing by DefaultTicksPerBar.

### DIFF
--- a/src/gui/widgets/TrackContentWidget.cpp
+++ b/src/gui/widgets/TrackContentWidget.cpp
@@ -209,7 +209,7 @@ void TrackContentWidget::changePosition( const TimePos & newPos )
 						it != m_tcoViews.end(); ++it )
 		{
 		if( ( *it )->getTrackContentObject()->
-						startPosition().getBar() == curBB )
+						startPosition().getTicks() / DefaultTicksPerBar == curBB )
 			{
 				( *it )->move( 0, ( *it )->y() );
 				( *it )->raise();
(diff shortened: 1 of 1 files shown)


sdasda7777:
#6096 is in the same file, so please do not pull this one.

Veratil:
Why didn't you just add the commit to this PR?

sdasda7777:
To be perfectly honest, I'm bit of a noob and didn't know that was a thing.

Veratil:
> 
> 
> To be perfectly honest, I'm bit of a noob and didn't know that was a thing.

All good! Next time you know you can just add to the same branch and push and it updates automagically here. :)

sdasda7777:
I see, thanks

PR #6096, opened 2021-07-17
Fixes multiple bugs related to BBEditor and time signatures

sdasda7777:
This pull request fixes multiple bugs related to BBEditor and time signatures, stemming from incorrect or outdated usage of method getBar():

- Lines 212 and 228: Fix the issue of incorrect BB track being displayed when time signature is not n/n (#6079)
- ~~Line 575: Fixes the issue that if new BB subtrack is added when time signature isn't n/n, due to the way getBar() works the result wouldn't be the one the formula using it was hoping for, and, for example, subtrack added at time signature 8/4 would only appear on every other BB track (subtrack for track 1 would appear on track 2, for 2 on 4, etc.).~~

### DIFF
--- a/src/core/Track.cpp
+++ b/src/core/Track.cpp
@@ -41,6 +41,7 @@
 #include "PatternTrack.h"
 #include "SampleTrack.h"
 #include "Song.h"
+#include "TimePos.h"
 
 
 namespace lmms
@@ -498,9 +499,9 @@ void Track::createClipsForPattern(int pattern)
 {
 	while( numOfClips() < pattern + 1 )
 	{
-		TimePos position = TimePos( numOfClips(), 0 );
+		auto position = TimePos(numOfClips() * DefaultTicksPerBar);
 		Clip * clip = createClip( position );
-		clip->changeLength( TimePos( 1, 0 ) );
+		clip->changeLength(TimePos(DefaultTicksPerBar));
 	}
 }
 
--- a/src/gui/tracks/TrackContentWidget.cpp
+++ b/src/gui/tracks/TrackContentWidget.cpp
@@ -239,7 +239,7 @@ void TrackContentWidget::changePosition( const TimePos & newPos )
 		// first show clip for current pattern...
 		for (const auto& clipView : m_clipViews)
 		{
-			if (clipView->getClip()->startPosition().getBar() == curPattern)
+			if (clipView->getClip()->startPosition().getTicks() / DefaultTicksPerBar == curPattern)
 			{
 				clipView->move(0, clipView->y());
 				clipView->raise();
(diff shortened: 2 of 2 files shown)


PhysSong:
It seems like `Track::createTCOsForBB` needs modification as well to make the behavior consistent.

sdasda7777:
You're right, that seems to do the trick

src/gui/widgets/TrackContentWidget.cpp
```diff
@@ -572,8 +572,7 @@ void TrackContentWidget::mousePressEvent( QMouseEvent * me )
 			so.at( i )->setSelected( false);
 		}
 		getTrack()->addJournalCheckPoint();
-		const TimePos pos = getPosition( me->x() ).getBar() *
-						TimePos::ticksPerBar();
+		const TimePos pos = getPosition( me->x() ).getTicks();
```
PhysSong, line 575: To preserve old behavior(not noticeable unless cloning tracks inside BB to song editor, though):
```suggestion
		const TimePos pos = getPosition( me->x() ).getTicks() / DefaultTicksPerBar * DefaultTicksPerBar;
```

PhysSong, line 575: Please revert this as this doesn't apply to BB. I deleted my old comment because it was wrong.

sdasda7777:
I'm not sure I follow: the bug line 575 fixes is described above and I believe the solution is correct.

src/gui/widgets/TrackContentWidget.cpp
sdasda7777: I'm not sure I follow: the bug line 575 fixes is described above and I believe the solution is correct.

PhysSong: This only brings a regression: clips are not created at the bar boundary now.

sdasda7777: Again, I'm not sure I'm understanding your point correctly, but in my testing the line does indeed fix the bug, so I believe it should be kept.

PhysSong: I can't find any reason this should work. Could you try reverting this and reproducing the bug?

PhysSong: This `if` statement covers the case where you add new `TrackContentObject`s(clips) in the song editor.

sdasda7777: > I can't find any reason this should work. Could you try reverting this and reproducing the bug?

I see, can't seem to be able to trigger the bug either, so that means you're right.

rubiefawn:
I fixed the comically absurd diff† by cherry-picking‡ the two commits that actually matter for this PR onto current `master` branch.

- 1c8aff19fa48012076d4a58b0def73fee4b68710 -> 79c3c54eb5d2daa39fe68b925b27f2eba26665a7
- 5e4040580705a42cbed35bf2a6e3ddcf39ca8061 -> Removed (see https://github.com/LMMS/lmms/pull/6096#discussion_r676092078)
- 66d2886be1d878f0bc1c1b39141573832b1e4f7c -> 3f0c3bab0692e051c12a59f5827e41193cab55b0

Please note that this does not count as code review. I solely focused on fixing the git history.

† image image

‡ Rewrote the equivalent, since files and classes have all been renamed since the original commits were made lol

PR #6107, opened 2021-07-22
Use QSaveFile on file write

### LINKED ISSUE #4514: LMMS file save fail
LMMS occasionally fails to save. I was bumped about this issue on Discord (@DouglasDGI) but I've seen cases on the old forum and also noticed it myself. I saw it once with a **recover.mmp** and I may have mentioned this at the occasion in a thread on GitHub. In the case presented on discord the file was verified to have a size of 0 bytes.

Examples:
[Can't save projects on Mac osx](https://lmms.io/forum/viewtopic.php?f=7&t=28530)
[Trouble with saving projects on Mac OSX](https://lmms.io/forum/viewtopic.php?f=7&t=8446)

When writing a file, we do test for it being larger than 0 bytes [here](https://github.com/LMMS/lmms/blob/e07f9a3b648d8ce168f247217509677ecd38811d/src/core/DataFile.cpp#L307): 

`	if( QFileInfo( outfile.fileName() ).size() > 0 )`

If going by size like this, maybe we could bump the zero to 100 or so, as the minimum file size with header seem to be at least a hundred bytes.

zonkmachine:
Fixes #4514 by replacing QFile with QSaveFile on save action.

### DIFF
--- a/src/core/DataFile.cpp
+++ b/src/core/DataFile.cpp
@@ -412,8 +412,8 @@ bool DataFile::writeFile(const QString& filename, bool withResources)
 		{
 			QMessageBox::critical(NULL,
 				SongEditor::tr("Could not write file"),
-				SongEditor::tr("An unknown error has occured and the file could not "
-						"be saved.").arg(fullName));
+				SongEditor::tr("An unknown error has occured and the file "
+						"could not be saved."));
 		}
 
 		return false;
--- a/src/core/DataFile.cpp
+++ b/src/core/DataFile.cpp
@@ -380,7 +380,7 @@ bool DataFile::writeFile(const QString& filename, bool withResources)
 		}
 	}
 
-	QSaveFile outfile (fullNameTemp);
+	QSaveFile outfile(fullNameTemp);
 
 	if (!outfile.open(QIODevice::WriteOnly | QIODevice::Truncate))
 	{
@@ -413,7 +413,7 @@ bool DataFile::writeFile(const QString& filename, bool withResources)
 		return false;
 	}
 
-	if (ConfigManager::inst()->value("app", "disablebackup" ).toInt())
+	if (ConfigManager::inst()->value("app", "disablebackup").toInt())
 	{
 		// remove current file
 		QFile::remove(fullName);
--- a/src/core/DataFile.cpp
+++ b/src/core/DataFile.cpp
@@ -408,14 +408,8 @@ bool DataFile::writeFile(const QString& filename, bool withResources)
 
 	if (!outfile.commit())
 	{
-		if (gui)
-		{
-			QMessageBox::critical(NULL,
-				SongEditor::tr("Could not write file"),
-				SongEditor::tr("An unknown error has occured and the file "
-						"could not be saved."));
-		}
-
+		showError(SongEditor::tr("Could not write file"),
+			SongEditor::tr("An unknown error has occured and the file could not be saved."));
 		return false;
 	}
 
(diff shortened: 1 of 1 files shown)


Spekular (approved):
LGTM. Merge?

Spekular:
Actually though, aren't we doing duplicate work? QSaveFile already uses a temporary file, so we can probably write directly to the final filename now, no?

Spekular (changes requested):
Approved too fast (only looked at diff). Upon closer inspection, it seems like we should remove our homebrewed temp file since QSaveFile already uses the same method.

DomClark:
Prior to this PR, the backup file was the old version of the project file, before being saved. Now, it is identical to the newly-saved version. Is this intended? It doesn't strike me as particularly useful.

zonkmachine:
Right. Looking into it.

zonkmachine:
After discussion on Discord `#dev-only` fullNameTemp is now back. I just dropped the file size check which is now redundant.
Saving with and without backup file works.

zonkmachine:
Now after the QSaveFile commit() is performed we still do a bunch of QFile actions and the commited file could end up either in fullName + ".new" or being deleted if one of the QFile::remove(...)/QFile::rename(...), for whatever reason, fails.

zonkmachine:
Also, if a save operation fail it returns `false`, but it doesn't look like we act on that everytime we save. We do when using `MainWindow::mayChangeProject()` such as in `MainWindow::closeEvent()`.

src/core/DataFile.cpp
```diff
@@ -387,30 +388,29 @@ bool DataFile::writeFile(const QString& filename, bool withResources)
...
-		{
-			// remove old backup file
-			QFile::remove( fullNameBak );
-			// move current file to backup file
-			QFile::rename( fullName, fullNameBak );
-		}
-		// move temporary file to current file
-		QFile::rename( fullNameTemp, fullName );
+		showError(SongEditor::tr("Could not write file"),
+			SongEditor::tr("An unknown error has occured and the file could not be saved."));
```
zonkmachine, line 394: outfile.errorString() goes in here...

zonkmachine:
I would be jolly happy if someone whould like to take over this issue/PR. @Spekular?
I don't feel confident enough with the coding here and I'm to rusty anyway. Just grab the code.. :) Plz...

src/core/DataFile.cpp
zonkmachine: ^2 years later... What is that man on about?!

src/core/DataFile.cpp
```diff
@@ -379,12 +380,12 @@ bool DataFile::writeFile(const QString& filename, bool withResources)
 		}
 	}
 
-	QFile outfile (fullNameTemp);
+	QSaveFile outfile (fullNameTemp);
```
Veratil, line 383:
```suggestion
	QSaveFile outfile(fullNameTemp);
```

src/core/DataFile.cpp
```diff
@@ -405,30 +406,29 @@ bool DataFile::writeFile(const QString& filename, bool withResources)
...
-		}
-		// move temporary file to current file
-		QFile::rename( fullNameTemp, fullName );
+		showError(SongEditor::tr("Could not write file"),
+			SongEditor::tr("An unknown error has occured and the file could not be saved."));
+		return false;
+	}
 
-		return true;
+	if (ConfigManager::inst()->value("app", "disablebackup" ).toInt())
```
Veratil, line 416:
```suggestion
	if (ConfigManager::inst()->value("app", "disablebackup").toInt())
```

Veratil (approved):
Two quick style fixes and good to go

zonkmachine:
@Veratil Thanks!

Issue #5779, opened 2020-11-11
Show knob value without clicking.


Gabrielxd195:
### Summary of improvements
Show the value of the knob without clicking on it

#### Justification
When I want to know the value of a knob I have to click on it, but in some cases when I click, sometimes by accident I move the mouse and change the value of the knob.

I know that we have the option of '' reset '' to restore the default value, but it would be more practical if the knob will show its value just by hovering over it like the other icons and buttons of the interface.

#### Sketch

Here is an image of the problem.
![Mostrar el valor de la perilla al pasar el puntero](https://user-images.githubusercontent.com/26127693/98775509-a8a65500-23c3-11eb-8714-6aa66fe70358.png)

musikBear:
Agree! And it ought to be a simple fix, if the super-class has an *'on-left-click-event'*. Change that to *on-mouse-over-event*
'Off' could be an issue -though, unless there is a *on-mouse-leave-event* implemented.
Only cravat i can see is a confusing flicker of values, as the mouse is moved over controllers, and that could be annoying. A short 'delay' before the tool-tip is displayed, would sort that(?)

PhysSong:
To implement this, one need to show and place `Knob::s_textFloat` with a `QTimer` to prevent flickers.

salilra:
I can take this issue, if no one is working on it.

justlog:
Thought about trying my hand at contributing to the project and came across this. It seems as though @salilra implemented the fix but it's not merged as of yet. Is there any work to be done regarding this issue?

Rossmaxx:
@salilra update? 
@justlog feel free to make your version.

Monospace-V:
> update?

https://github.com/LMMS/lmms/pull/5911

AngeloSorte:
I like this! I will try to give some help!

chaitanyak175:
Is anyone working on this issue I can give it a try as this will be my first time contributing to an open source project.

MrTopom:
I could work on it...

zonkmachine:
@MrTopom There seem to be an open PR for this issue. https://github.com/LMMS/lmms/pull/5911
I don't know what state it's in though.

zonkmachine:
New comments in the PR https://github.com/LMMS/lmms/pull/5911

PR #6117, opened 2021-07-27
Show knob value on hover.

DarkoiV:
My take on [issue 5778](https://github.com/LMMS/lmms/issues/5779)

### DIFF
--- a/include/Fader.h
+++ b/include/Fader.h
@@ -114,6 +114,8 @@ private:
 	void mousePressEvent( QMouseEvent *ev ) override;
 	void mouseDoubleClickEvent( QMouseEvent* mouseEvent ) override;
 	void mouseMoveEvent( QMouseEvent *ev ) override;
+	void enterEvent( QEvent *ev) override;
+	void leaveEvent( QEvent *ev) override;
 	void mouseReleaseEvent( QMouseEvent * _me ) override;
 	void wheelEvent( QWheelEvent *ev ) override;
 	void paintEvent( QPaintEvent *ev ) override;
--- a/include/Knob.h
+++ b/include/Knob.h
@@ -135,6 +135,8 @@ protected:
 	void mousePressEvent( QMouseEvent * _me ) override;
 	void mouseReleaseEvent( QMouseEvent * _me ) override;
 	void mouseMoveEvent( QMouseEvent * _me ) override;
+	void enterEvent( QEvent * _e) override;
+	void leaveEvent( QEvent * _e) override;
 	void mouseDoubleClickEvent( QMouseEvent * _me ) override;
 	void paintEvent( QPaintEvent * _me ) override;
 	void wheelEvent( QWheelEvent * _me ) override;
--- a/include/TextFloat.h
+++ b/include/TextFloat.h
@@ -47,6 +47,9 @@ public:
 
 	void setVisibilityTimeOut( int _msecs );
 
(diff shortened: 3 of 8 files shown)


superpaik:
Tested on Windows 10 64bits.
It works as expected, but in the song editor volume label for an instrument is "volume" but for a sample track is "channel volume". I supose that's because of the "internal" label of the knob.

Also, it would be great to have this behaviour on the volume slider on the FX-Mixer and also on the piano roll on "note velocity" and "note panning". They are not strictly knobs but I think would be useful. Maybe on other PR, though.

DarkoiV:
I added it to sliders in FX-Mixer too, since it seems like pretty much the same change. 
Though it now shows both name of slider and volume on separate floats, idk if this is desired behavior, but they don't overlap and are readable. 
FXMixer

superpaik:
Yeah, it's kind of weird. Idk why the FX name is shown when hover over the channel (maybe for long FX names). But, for exemple, when you hover a track on the Song Editor you don't get the track name (even if the track name is not full visible). 
I would remove the FX name, but I know that removing existing functionality may not be a good idea.
So, let it be, unless other people says otherwise.

wujekbrezniew:
> I added it to sliders in FX-Mixer too, since it seems like pretty much the same change.
> Though it now shows both name of slider and volume on separate floats, idk if this is desired behavior, but they don't overlap and are readable.
> FXMixer

Does it shows two separate boxes with names/values like one pic? Should not be better to show those things in one bigger box?

DarkoiV:
It probably would be better, but I have no idea how it is exactly shown. Fader class only have one static FloatText, so it has to be something else. 
Greping "TextFloat" only returns "TrackContentObjectView" that could be responsible for it. And I am quite unsure about messing with it's logic. Especially since it has hide() method on mouse release, but from what I see "FX" float disappears on mouse click, not release so I am not certain it is even that. 

Someone who has better grasp on this codebase needs to look at this I think.

Spekular:
I'm think this is where the tooltip is set: 
https://github.com/LMMS/lmms/blob/2f17c0da2b20ecb9b00f8dcb227d02731325796b/src/gui/widgets/FxLine.cpp#L101

If you remove it and add `m_renameLineEdit.setToolTip( name );` instead, you might get a tooltip that only shows when hovering over the name. Then the issue should be solved, as each tooltip has it's own area.

I also found this line, not sure what it does though:
https://github.com/LMMS/lmms/blob/2f17c0da2b20ecb9b00f8dcb227d02731325796b/src/gui/FxMixerView.cpp#L365

DarkoiV:
Yeah, they both are responsible for it.  
I couldn't make it work on `m_renameLineEdit`, but it can work on m_lcd. Though I dunno if it is even worth to keep this tool tip.
Name is clearly visible on line itself anyway. 

I think proper implementation of choosing between line tool tip and fader value should be made with `eventFilter`. 

And I am no sure how to do it right now. FxLine is not parent of Fader. Nor is FxMixerView parent of them. They are both contained in inner class FxChannelView which is not Widget.

For now I just deleted it.

Spekular:
> Though I dunno if it is even worth to keep this tool tip.
> Name is clearly visible on line itself anyway.

Except when the name is too long. It's also easier to read horizontal text than vertical.

DarkoiV:
That's good point, I will try to implement `eventFilter` then.

Issue #4038, opened 2017-12-06
Metronome starts skipping at fast tempo.


EtornaZ:
I am using LMMS 1.2.0 RC4. 
The metronome in LMMS will start skipping if the tempo is faster than 218 BPM, in about 15-20 seconds. If the tempo is sped up to around 230 BPM, it starts skipping sooner, more like 5-7 seconds. When it's at 250 the metronome will immediately start skipping. I restarted LMMS just in case it was a temporary thing, but the bug still persisted. 
This is all of the information I have, but if I find out more about this bug I'll let you all know.

PhysSong:
Confirmed. Metronome is handled in `Mixer::renderNextBuffer()`.
https://github.com/LMMS/lmms/blob/dd4a73eb4bb31870ff4397539710b6d3f0bbf1c6/src/core/Mixer.cpp#L359-L375

With 4/4 time signature, one tick is 1.25/(tempo in BPM) seconds. It's about 5.79 ms if the tempo is 216BPM. And the buffer is about 5.80 ms long with 44100Hz, 256 frames. When one tick is shorter than one mixer period, this logic breaks down.

There are some possible solutions:
1. Do a better calculation in the same function. Might be a little bit tricky?
2. Move the metronome logic to `Song::processNextBuffer()`. It could be a better alternative because this function adds play handles for ordinary tracks. 
3. Introduce a new sort of track, named "Metronome track". This is simple because metronome could be treated as a normal track. It means `Mixer::renderNextBuffer()` and/or `Song::processNextBuffer()` don't need any hacks for metronome.

gi0e5b06:
@PhysSong The best design would be the third one but that would involve too many changes. If the core got redesigned one day, that should be the best choice. You have a better understanding than me on how things work internally so just chose what works best among the two first ones.
Until then, I fixed the problem. The code for the metronome is now frame-exact and 100 times faster. It works fine up to 999 bpm (I haven't tested over).
https://github.com/gi0e5b06/lmms/blob/master/src/core/Mixer.cpp#L344

PhysSong:
@gi0e5b06 Thank you for the suggestion. That makes sense, but looks inefficient. I will try to find a better way.

gi0e5b06:
[...]

Issue #475, opened 2014-03-17
Better metronome


musikBear:
I think the metronome is of low quality. Several has asked for ways to shut it of, all together.. (whitch easy btw) -to me that idicate an area where something new is needed. 
At least a two-effect metronome (eg tttToc ttt Toc) with a clear difference on-beat, but perhaps more. Why not let the user set volume and perhaps a selection of sounds?

In the more exotic department (craaazy:P) I wonder if metronome could be a usefull tool
in live record -How about a time-delayed feature. -Eg a metronome that could be set to a precise off-value. this would let the user listen for the toc, and then have able time to do a note event. Depending of Q-setting and perhaps even HW reaction-delay, this could aid the use, in a creative fashion? -Also a 'beat-counting' feature could be implemented with a variable metronome. The user then turn the dial untill the ticks fits the played music - a display shows current BPM. -just some thoughts in respet to a somewhat disliked feature.

---

 - [x] Toggle metronome on/off. (#2278)
 - [ ] Metronome volume control.

zonkmachine:
I'm looking into this. At a bare minimum I'll fix an on/off button and a setting to switch it default on/off.

Moth-Tolias:
Being able to set an offset would be of valuable assistance when dealing with latency.

the-bebop:
I just had a look into the aspect of the metronome's volume control. I find it a little too much to open up a new (maybe hidden) channel in the mixer just for the metronome, however I couldn't think of anything else less code invasive, then again I just have started with LMMS'Code. Do you think going over the mixer would be the right way?

musikBear:
@the-bebop metronome-volume could be regarded as a similar issue as prehear-volume
![image](https://user-images.githubusercontent.com/6662066/127010024-51aa743a-0241-49d7-9c88-ec16bd5ec060.png)
Both are atm not controllable.
I think something like right click could make a slider appear.
Is there a _property_ for the volume at all? -And _if so_ -in witch class?

Spekular:
#### Reference
**Ableton** uses a combobox/button combo (we have this in piano roll for quantization) to expose additional metronome options.
![image](https://user-images.githubusercontent.com/8918426/127019687-242e46f8-44d0-4863-9fc1-a68ed8f49f14.png)

**FL Studio** lets you choose metronome type by right clicking the metronome enable button.
![image](https://user-images.githubusercontent.com/8918426/127019909-541d018f-b07e-44bb-a540-93cba49ed1cd.png)

[...]

PR #6119, opened 2021-07-30
Added Metronome context menu

the-bebop:
This PR adds additional settings to the Metronome, previously discussed [here](https://github.com/LMMS/lmms/issues/475#issuecomment-887648135) 

This PR does not fix the timing issue [#4038](https://github.com/LMMS/lmms/issues/4038)

grafik

### DIFF
--- a/src/core/SamplePlayHandle.cpp
+++ b/src/core/SamplePlayHandle.cpp
@@ -32,28 +32,29 @@
 
 
 
-SamplePlayHandle::SamplePlayHandle( SampleBuffer* sampleBuffer , bool ownAudioPort ) :
+SamplePlayHandle::SamplePlayHandle(SampleBuffer* sampleBuffer , bool ownAudioPort, float sample_volume) :
 	PlayHandle( TypeSamplePlayHandle ),
 	m_sampleBuffer( sharedObject::ref( sampleBuffer ) ),
 	m_doneMayReturnTrue( true ),
 	m_frame( 0 ),
 	m_ownAudioPort( ownAudioPort ),
-	m_defaultVolumeModel( DefaultVolume, MinVolume, MaxVolume, 1 ),
+    m_defaultVolumeModel( DefaultVolume, MinVolume, MaxVolume, 1 ),
 	m_volumeModel( &m_defaultVolumeModel ),
 	m_track( NULL ),
 	m_bbTrack( NULL )
 {
+    m_sampleBuffer->setAmplification(sample_volume);
 	if (ownAudioPort)
 	{
 		setAudioPort( new AudioPort( "SamplePlayHandle", false ) );
-	}
+    }
 }
 
 
 
 
-SamplePlayHandle::SamplePlayHandle( const QString& sampleFile ) :
-	SamplePlayHandle( new SampleBuffer( sampleFile ) , true)
+SamplePlayHandle::SamplePlayHandle(const QString& sampleFile , float sample_volume) :
+    SamplePlayHandle( new SampleBuffer( sampleFile ) , true, sample_volume)
 {
 	sharedObject::unref( m_sampleBuffer );
 }
--- a/include/MainWindow.h
+++ b/include/MainWindow.h
@@ -32,6 +32,7 @@
 
 #include "ConfigManager.h"
 #include "SubWindow.h"
+#include "MetronomeSettingsMenu.h"
 
 class QAction;
 class QDomElement;
@@ -223,6 +224,7 @@ private:
 	QMenu * m_viewMenu;
 
 	ToolButton * m_metronomeToggle;
+    MetronomeSettingsMenu * m_metronomeSettingsMenu;
 
 	SessionState m_session;
 	
@@ -234,6 +236,9 @@ private slots:
 	void updateViewMenu( void );
 	void updateConfig( QAction * _who );
 	void onToggleMetronome();
+    void onShowMetronomeSettings();
+    void onMetronomeSettingsChanged(std::pair<QString, QString>);
+    void onMetronomeVolumeChanged(float);
 	void onExportProject();
 	void onExportProjectTracks();
 	void onImportProject();
--- a/include/MetronomeSettingsMenu.h
+++ b/include/MetronomeSettingsMenu.h
@@ -1,0 +1,81 @@
+#ifndef METRONOMESETTINGSMENU_H
+#define METRONOMESETTINGSMENU_H
+
+#include <QtWidgets/QWidget>
+#include <QtWidgets/QVBoxLayout>
+#include <QtWidgets/QPushButton>
+#include <QtWidgets/QSlider>
+#include <QtWidgets/QLabel>
+#include <QtCore/QString>
+#include <QMouseEvent>
+#include <QFont>
+
+
+struct OptionElements{
+        std::vector<std::pair<QString, QPushButton*> > rythm;
+        QSlider* volume;
+};
+
+/**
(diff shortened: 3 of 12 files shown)


src/core/SamplePlayHandle.cpp
```diff
@@ -32,28 +32,29 @@
...
-SamplePlayHandle::SamplePlayHandle( SampleBuffer* sampleBuffer , bool ownAudioPort ) :
+SamplePlayHandle::SamplePlayHandle(SampleBuffer* sampleBuffer , bool ownAudioPort, float sample_volume) :
 	PlayHandle( TypeSamplePlayHandle ),
 	m_sampleBuffer( sharedObject::ref( sampleBuffer ) ),
 	m_doneMayReturnTrue( true ),
 	m_frame( 0 ),
 	m_ownAudioPort( ownAudioPort ),
-	m_defaultVolumeModel( DefaultVolume, MinVolume, MaxVolume, 1 ),
+    m_defaultVolumeModel( DefaultVolume, MinVolume, MaxVolume, 1 ),
 	m_volumeModel( &m_defaultVolumeModel ),
```
the-bebop, line 42: cannot find usage of m_volumeModel, has this become obsolete?

the-bebop:
- [ ] take over LMMS appearance for new slider

Monospace-V:
*Rhythm is typoed.
image

Monospace-V:
I found a bug in the metronome behaviour. Do I report it here or on the main tracker?

zonkmachine:
@Monospace-V What problem is it? Have you searched the issues? We have quite a few bugs reported for the metronome.

Monospace-V:
The problem is specifically with the behaviour of metronome context menu, and I've copied out the format of a bug report and filled it in:

### Metronome context menu remaining open in some cases

### Bug Summary
When triggered by certain user behaviour, the metronome context menu can remain open and move to the background as a separate window. The context menu cannot be opened again until it is returned to the foreground and interacted with. Sometimes, the window **can even remain open after LMMS is closed**.

#### Steps to reproduce
1. Right click metronome to open context menu
2. Do not interact with context menu or allow it to close, instead clicking directly on the metronome or something.
3. Try opening metronome context menu again. (does not open)
4. View LMMS' open windows in taskbar. Metronome context menu is still there.

#### Expected behavior
Metronome context menu should close when not in foreground.

#### Actual behavior
Context menu remains open as another separate window. In my case, it was possible for me to close LMMS and the context menu remains open even after closing unless interacted with.

#### Screenshot
(Video)

https://user-images.githubusercontent.com/76674645/130239836-213764c7-6277-4a8e-8861-80bd0a4f5a10.mp4


#### Affected LMMS versions

(The one downloaded from this pull request)

zonkmachine:
> The problem is specifically with the behaviour of metronome context menu

Ok, got it. That belongs here.

frank-ihle:
> My VM, an Ubuntu 18.04, won't even run the AppImage, so I can't tell.
> I cannot tell why this is happening. Perhaps if other Windows users tried reproducing it we would be able to confirm.
> What I did was very quickly get my mouse out and enable metronome without giving the menu time to close.
> I also find that the menu closes **only** if you hover over it while it's still open, and then move your mouse away. If you do not let your mouse touch the menu, it just stays open. I believe this was how I reproduced it staying open in the first place.

Ok I managed to reproduce it, let me see

frank-ihle:
Fixed on your side too @Monospace-V  ?

Monospace-V:
Where do I download it? I do not have the necessary tools or skillset to compile.

zonkmachine:
@Monospace-V He meant the new binaries found here: https://github.com/LMMS/lmms/pull/6119#issuecomment-890150610
No compilation needed.

allejok96:
Why not do it using QMenu and QWidgetAction, instead of opening a new window? Then you wouldn't need to manually close it when focus is lost. And maybe use Knob and LcdSpinBox to make it look more LMMS themed?

PS we need 3/4.

Monospace-V:
image
It doesn't seem to have updated today. Downloaded and checked: gives the bug, this is probably because the binary was not updated.

allejok96:
@Monospace-V you can get it by clicking "Show all checks" > "Details" below this message.
Or direct link: https://14690-15778896-gh.circle-artifacts.com/0/lmms-1.3.0-alpha.1.138%2Bg3234182-linux-x86_64.AppImage

Monospace-V:
Thanks, Alex!
Tested it, Frank, and I notice you've fixed the issue user-side by moving the menu to where there is no way you can not hover over it if you right click the metronome, making it likely impossible that the user ever trigger it.
I then found, after a couple of tries, that you can still reproduce it by moving your mouse fast enough.
However, it does close when LMMS is closed.

frank-ihle:
I updated the design to something that fits a little more into the concept, taking into account @allejok96 's QMenu hint

image

I generally would be happy to finalize the frontend, so I'm open for any remarks.

frank-ihle:
Also it seems like we're lacking a common color definition, as the implementations I saw always come with a local numeric assignment, correct or where can I find the right color definition?

frank-ihle:
> Thanks, Alex!
> Tested it, Frank, and I notice you've fixed the issue user-side by moving the menu to where there is no way you can not hover over it if you right click the metronome, making it likely impossible that the user ever trigger it.
> I then found, after a couple of tries, that you can still reproduce it by moving your mouse fast enough.
> See:
> https://user-images.githubusercontent.com/76674645/131223116-78fac0b1-5db7-4df9-bf00-b175f9aca4bd.mp4
> 
> However, it does close when LMMS is closed.

Yep thanks for pointing out, my opinion is, this is acceptable. As it either gets closed on main application exit or the user is getting _forced_ to make a decision if he doesn't want to get annoyed by an opened menu

Monospace-V:
Agreed.
If possible, though, I'd suggest automatically closing the window if user doesn't interact with it for 30 seconds, or if it is not in foreground. Just to ensure that if a user triggers it, it will not render the menu unusable for the rest of the session.

frank-ihle:
grafik

- [ ] adapt QAction

allejok96:
Although I like the design so far, it does not match LMMS' ugly style :smile_cat: 
bild
Use themed Knob / Slider for volume?

Also we must have spinboxes for a beats like 5/4. Or a checkbox that says "same as time signature" (which is the default behavior right now). 

PS if you use QToolButton->setMenu you won't need to position the window and check for mouse pointer leaving it and so on, guess you've figured it out already...

frank-ihle:
FYI Possible fix for the flaky metronome sounds #6149

frank-ihle:
> Although I like the design so far, it does not match LMMS' ugly style 😸
> bild
> Use themed Knob / Slider for volume?

Good point, how to change the appearance? I was greping for the already implemented sliders but I couldn't find the design settings

allejok96:
I'm just guessing here but using Knob.h or AutomatableSlider.h (vertical only?) would give it the correct style by default. Style is defined in data/themes/default/style.css.

superpaik:
I've notice some issues:

- the popup window goes outside LMMS window boundaries, tested on W10 (see attached image)
- it's kind of weird that when you click on a rhythm the popup closes, so you have to open it again to make sure you have selected the one you wanted. Maybe it would be better to leave it opened and close it when the cursor leaves the window (as it works now for volume slider)
- it malfunctions when time signature is other than 4/4. In 1.2.2 metronome goes acordingly to time signature. For exemple, on 3/4 one expects three beats per bar -when using 1/4 as rhythm-, but you get four.

imatge

PR #6124, opened 2021-08-08
Disable tooltips in QStyle instead of wrapper class

allejok96:
The way the "Enable tooltips" setting is implemented currently, using a wrapper class, is a bit obscure. New developers (myself included) haven't realized they have to use the wrapper instead of doing it the normal Qt way, so may GUI elements does not respect the setting.

With this PR, unchecking the "Enable tooltips" will turn tooltips fully transparent using the QStyle. This affects all instances of `QToolTip`. The tooltips for knobs, sliders and clips are `TextFloat`, so they always remain visible.

The old wrapper class `ToolTip` is kept as a dummy. It's used in about 40+ other source files - keep it or leave it, idk...


Spekular:
I think a decorated class is a pretty reasonable way to implement disable-able tooltips since QToolTip doesn't support it, and this feels a bit hacky. I don't want to oppose this PR just based on a vague feeling though, since it does make it harder to accidentally ignore the setting. One possible difference that came to mind was screen readers, so I did some testing with Windows Narrator and the current version of LMMS.

- When tooltips are enabled and a tooltip has displayed, pressing Narrator (Caps) + 0 reads its content.
- When tooltips are disabled, pressing the same combo reads "no additional information available".

I imagine (haven't tested) that with this PR merged disabled tooltips would still be fully visible to Narrator. That might be fine for users manually examining them, but probably less so if one had tooltips set to automatically read and wanted to disable the non-essential ones? I couldn't get Narrator to auto-read tooltips when hovering stuff, and don't know how it's typically used, but if someone actively disables tooltips I think it's reasonable that they be disabled for real. (As a whole I doubt LMMS is currently useable if one relies on a screen reader, but we probably shouldn't make it worse w/o good reason)

DomClark:
Why not install a global event filter for `QEvent::ToolTip`? It's not hacky, means we can create tooltips the normal Qt way, and allows for easy toggling without a restart.

DomClark:
> The old wrapper class ToolTip is kept as a dummy. It's used in about 40+ other source files - keep it or leave it, idk...

I'd say now is the time to remove it, since we're currently making several large changes across the codebase.

allejok96:
> Why not install a global event filter for QEvent::ToolTip?

Absolutely brilliant, @DomClark! I'm not so good at Qt but I've been experimenting a little. As I understad it, the filter needs to be installed in `QApplication`. Creating `MainApplication::eventFilter` does not do anything per-se. To make that work I'd have to run `mainApp->installEventFilter(mainApp)` and, idk, that makes me feel dirty :stuck_out_tongue: 

Should I make a new QObject class for the filter?

Issue #1643, opened 2015-01-16
PianoRoll Shift+Insert/Shift+Delete Shortcut


musikBear:
would be intuitive if sh+ins / del, worked exactly as in song-editor, eg simply moved all notes right of playhead.
Distance should be current note-lenght
The functionality should be independent of selections

Sti2nd:
Please check the new shortcut key table. It is already listed there, I hope that would've be enough attention needed.

tresf:
@Sti2nd, Link the issue and make your point without the side comments please.

Sti2nd:
Yes, will link, I am on phone now.

jibin1573:
IMO, Shift+P should be the Pencil Tool (Draw Tool) while Shift+ D should be Detuning Tool.

Issue #1569, opened 2015-01-07
Context menu for notes in Piano Roll


Sti2nd:
Placeholder for ideas, feel free to comment and edit.

Actions
* Mute
* Detune
* Razor / Glue https://github.com/LMMS/lmms/issues/746
* Put to the background/place behind (when notes overlapping)
* Delete all notes behind (all overlapping notes on the same note label)
* Show information?

Information?
* Note label (C4, G#4/Ab4)
* Note duration/lenght
* Note velocity/volume
* Note panning value

ShurikMur:
Agreed.
BTW can't find a feature to change color of notes, it also can be done in a context menu.

badosu:
Today we use the right mouse button to delete notes, what should we do to open the context menu?

musikBear:
would it be an option to let the context-menu be opend, not on the selected notes, but on the hot-spot in the *volume-pan* section eg
![pastenot](https://cloud.githubusercontent.com/assets/6662066/5938209/084d815c-a6f9-11e4-8f68-791f7ec860a0.PNG)
instead of or as an extending of current menu?

Sti2nd:
> Today we use the right mouse button to delete notes, what should we do to open the context menu?

Already addressed https://github.com/LMMS/lmms/issues/1488

badosu:
@Sti2nd What was addressed? Do you have a suggestion?

Sti2nd:
> Do you have a suggestion?

Yes, even some shortcuts that don't yet exist (but are issues) are in the table. Delete in Piano Roll will be probably Shift + right click and middle click.

badosu:
@tresf @Sti2nd Can you confirm that this is what everyone has agreed? I can tell for myself: I hate having to use the middle or a composed right click to delete stuff, but if this change is consensus then it's fine.

tresf:
I have to agree with @Sti2nd, we can't have our cake and eat it too.  If we want a context-menu, we have to free up the shortcut that is stealing it.  Middle-click is what we use for song editor.  @Sti2nd is helping us manage these conflicts so I will defer to him in this case as far as our long term goal -- he's done the most research on it.

badosu:
Find by me :-)

tresf:
> Shift + right click and middle click

Isn't `Shift + Left click` what we use for middle?

Sti2nd:
> Isn't Shift + Left click what we use for middle?

[...]

Issue #1165, opened 2014-09-18
Per-instrument Timing, Volume and Pitch Humanization/Randomization


unfa:
I think that an arbitrary way to randomize the instrument output a bit could be a very nice function.

I propose adding a special "Humanize" panel with Volume, Timing and Pitch humanize knobs.

The idea is to process the notes before they are sent to the instruments. It'd need some read-ahead for time humanization so some notes can be played earlier.

The pitch humanization would detune every note randomly in some range - prefarably with a gaussian distribution function. This would simulate non-perfect note fretting on guitars.

The ideas could be ectended to create a "randomization" pane for instrument as there are many things that can be set for this.

What do you think?

rubiefawn:
+1. Great for drum sequences too, makes it seem a little less robotic.

gi0e5b06:
https://github.com/gi0e5b06/lmms/commit/1e3eb1b1b85bdefe9ee30e229c05a3374420dc22
![humanizer_20171121](https://user-images.githubusercontent.com/30031949/33099683-f2ff49b8-cf11-11e7-80b2-7a8fdc968f9f.png)

tresf:
As part of a pruning effort, this `enhancement` request is archived into a dedicated "Better Workflow" checklist here #4877.

tresf:
@musikBear claims this is `in progress`, marking and assigning as such.

allejok96:
Some good ideas suggested by @spechtstatt in #6140. What about something like this:

![bild](https://user-images.githubusercontent.com/7693838/149630372-287aab1e-e628-4aa7-80e0-ee28f507764a.png)

You pull the knob, the notes are humanized. You release the knob, it bounces back to zero and changes are applied.

Good:
- solves note position problem
- enables manual edit of humanization
- live visual feedback

Bad:
- humanization cannot be "turned off" later

musikBear:
i made humanization strumming and some move things, but could not get it on git from a VB
Atm i am trying to get vs2022 to run on windows
This is what i have:
https://www.youtube.com/watch?v=OlBtU0JYvyY

spechtstatt:
@allejok96: I don't think that the missing "turn off" functionality is really much of an issue as long as you have the possibility to reduce the effect again and/or get back to an initial state. 

Best is maybe to show (again) a picture :-)

![image](https://user-images.githubusercontent.com/93736385/149663966-4ff99fde-476b-46d2-9d2f-e1051b5a52d9.png)

[...]

PR #6140, opened 2021-08-26
Demo: Piano-Roll chop, stamp, strum, bulldozer and mode switcher

allejok96:
This is only meant as a showcase. Everything will have to be redone when #6063 get merged. So don't look at the code right now. It's horrible. Let's focus the discussion on the UX aspect.

https://user-images.githubusercontent.com/7693838/130973381-b409fe3c-fe3c-44ff-8aec-bd0cd22a2dd4.mp4

## Motivation
There are many cool features in Pianoroll that is hidden behind modifier keys. And sometimes the order of pressing down keys matter. Like shift-press-move vs press-shift-move and shift-ctrl-resize vs ctrl-shift-resize. This draft suggests splitting out all these features into dedicated edit modes.

## Quick edit mode button

bild

- All extra edit modes are gathered in a ComboButton.
- Last selected action is the one displayed by the button.
- Scrolling the button changes mode.
- Mid-click in the editor is a shortcut to whatever last used on this button.
- Right click *always* takes us back to draw mode.

## Knife improvements

- If cutting on a selection, the cut spans over all notes.
- When holding shift, the cut is repeated to the end of the note/selection.

## Stamp tool

- Copies selected notes to the position of mouse click
- Shift-click in draw mode does this also.

## Strum tool
Related #2380

- Clicking on a stack of notes selects the whole stack
- Dragging a note in the stack create an arp.
- The dragged note is quantized, but not the notes in between.
- Dragging on a selection scales it proportionally (like shift-resize) but without altering note length.

## Bulldozer
Related #1643

- When moving notes, all notes starting from that position and right are moved.
- Resizing a note: all notes ending on the same tick are resized. Posterior notes are moved.
- Resizing a selection: resize proportionally (like shift-resize). Posterior notes are moved.
- Resizing while holding shift: scale everything within start and end mark proportionally. Posterior notes are moved.

I can't seem to find a good logic when it comes to resizing in bulldozer mode...

## Stuff removed from Draw mode
In favor of the dedicated edit modes.

- mousepress-shift-move to quick resize.
- shift-click to create an arp of selected chord.
- shift-ctrl-resize to resize and move posterior notes.

### DIFF
--- a/include/HexMenu.h
+++ b/include/HexMenu.h
@@ -1,0 +1,56 @@
+/*
+ * HexMenu.h - hexagonal menu widget
+ *
+ * Copyright (c) 2021 Alex <allejok96/gmail>
+ *
+ * This file is part of LMMS - https://lmms.io
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this program (see COPYING); if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA 02110-1301 USA.
+ *
+ */
+
+
+#ifndef CIRCULAR_MENU_H
+#define CIRCULAR_MENU_H
+
+#include <QWidget>
+#include <QMenu>
+#include "lmms_export.h"
+
+class LMMS_EXPORT HexMenu : public QWidget
+{
+	Q_OBJECT
+public:
+	HexMenu(QWidget* parent = nullptr);
+	virtual ~HexMenu();
+	void addAction(QAction* action);
+	void addMenu(QMenu* menu);
+
+protected slots:
+	void mouseMoveEvent(QMouseEvent* event) override;
+	void paintEvent(QPaintEvent* event) override;
+	void mouseReleaseEvent(QMouseEvent* event) override;
+
+private:
+	bool m_mouseHasMoved = false;
+
+	QMenu* m_contextMenu = nullptr;
+
+	std::vector<QAction*> m_actions;
+	int m_hoveredAction;
+};
+
+#endif
--- a/include/PianoRoll.h
+++ b/include/PianoRoll.h
@@ -32,6 +32,7 @@
 #include <QInputDialog>
 
 #include "Editor.h"
+#include "HexMenu.h"
 #include "ComboBoxModel.h"
 #include "SerializingObject.h"
 #include "Note.h"
@@ -162,6 +163,7 @@ protected:
 	void mouseDoubleClickEvent( QMouseEvent * me ) override;
 	void mouseReleaseEvent( QMouseEvent * me ) override;
 	void mouseMoveEvent( QMouseEvent * me ) override;
+	void contextMenuEvent(QContextMenuEvent* cme) override;
 	void paintEvent( QPaintEvent * pe ) override;
 	void resizeEvent( QResizeEvent * re ) override;
 	void wheelEvent( QWheelEvent * we ) override;
@@ -531,6 +533,8 @@ public:
 signals:
 	void currentPatternChanged();
 
+protected slots:
+	void contextMenuEvent(QContextMenuEvent* event) override;
 
 private slots:
 	void updateAfterPatternChange();
@@ -545,6 +549,7 @@ private:
 	void updateStepRecordingIcon();
 
 	PianoRoll* m_editor;
+	HexMenu* m_editModeSelector;
 
 	ComboBox * m_zoomingComboBox;
 	ComboBox * m_zoomingYComboBox;
@@ -555,7 +560,6 @@ private:
 	ComboBox * m_chordComboBox;
 	ComboBox* m_snapComboBox;
 	QPushButton * m_clearGhostButton;
-
 };
 
 
--- a/src/gui/CMakeLists.txt
+++ b/src/gui/CMakeLists.txt
@@ -77,6 +77,7 @@ SET(LMMS_SRCS
 	gui/widgets/FxLineLcdSpinBox.cpp
 	gui/widgets/Graph.cpp
 	gui/widgets/GroupBox.cpp
+	gui/widgets/HexMenu.cpp
 	gui/widgets/InstrumentFunctionViews.cpp
 	gui/widgets/InstrumentMidiIOView.cpp
 	gui/widgets/InstrumentSoundShapingView.cpp
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -1539,9 +1539,24 @@ int PianoRoll::keyAreaBottom() const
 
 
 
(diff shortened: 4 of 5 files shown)


Spekular:
Don't have time to deep dive but I'm a bit concerned about

> Right click always takes us back to draw mode.

Personally I think it would be good if right click was available for context menus when possible, so dedicating it as an "reset mode" shortcut would be a conflict. Of course RMB currently deletes notes so there's already a conflict, but that could be solved by e.g. allowing right click on an empty space to launch a menu (requiring a note to be clicked to enter erase mode). Does this PR create additional conflicts with RMB as a context menu?

Another concern is that (without having used it, to be fair) it seems clunky to go to the toolbar every time you want to access these tools. While they save time, there's an initial investment of moving the mouse up and back that makes them feel less convenient and immediate than they could be.

I think you could kill two birds with one stone if the modes and tools were available with a right click. Then you could e.g.
- Ctrl+Drag select > RMB > Fill
- RMB > Mode > Knife
- Ctrl+Drag select > RMB > Duplicate (new)
- Cut/paste selection
- Export selection

In combination with this I would probably collapse erase and select mode into the mode selection dropdown. They seem just as niche to me as some of the other modes introduced here, because I'm not sure who would click a box select icon instead of ctrl dragging, or click erase mode instead of RMB-ing or select + deleting.

allejok96:
Good suggestion, although a delicate topic. Related #1569 

> Does this PR create additional conflicts with RMB as a context menu?

Previously Knife used RMB to exit. The new thing here is that Select and Erase does that too. And the new modes of course. RMB in draw mode is still erase.

> requiring a note to be clicked to enter erase mode

RMB on empty space and dragging over notes should also erase IMO.

allejok96:
My reptile brain wants erase on right mouse button, no matter how hard I try to relearn. Better not change that (unless it becomes a user configurable option).

> there's an initial investment of moving the mouse up and back that makes them feel less convenient and immediate than they could be

Here's some more experimentation: **ctrl right click** to open a context menu. This is where all your actions be put. I just stuffed some stuff there at the moment. Menu content could perhaps depend on if user clicked on a note vs empty space.

And to switch between all them lovely new tools, use **ctrl right drag** to access this NEW FANCY HEXAGONAL GESTURE THINGY :confetti_ball:

https://user-images.githubusercontent.com/7693838/133894555-7e2f0ce3-1f64-48b7-b53a-01397b7e55c5.mp4

allejok96:
Strum tool is currently half broken btw

allejok96:
@spechtstatt thanks for the feedback. #1165 is what you want. Also it sounds like you want strum to be applied automatically to all chords? This strum tool is manual use only, and in hindsight I think that might be useless. You could achieve something strummy right now with the arpeggio, but it's also severely broken.

I've been running lmms with this patch for a while and this is my evaluation so far:
- bulldozer: *must have* for melody writing, should be developed further
- knife chop: time saver when you need it
- clone: just old man's shift-drag at the moment, needs something more
- strum: never used it
- fancy gesture thingy: works alright but no big hit. Having to press ctrl when doing the gesture is just messy on the brain. In that case I'd rather just use the keyboard to switch between modes. Have to think of something better...

spechtstatt:
No, I would just keep the approach with applying the operations for the selected notes only because the strumming may not be applied for all chords at once but varied along the progression - e.g. like it is often done in lofi.

I am not sure if I understand the approaches of @musikBear correctly,  but it seems to me that they are more track related (like the arpeggio) and it sounds to me that the chord selection seems to contain some "magic"?

I think it would be much more flexible to just have a relatively simple editing helper which is just applied to the selected notes. This allows also for an easy mixing of manual and "automatic" editing - like humanize the notes (velocity/start/length/padding and so on) and adapt it afterwards.


(But as I said: maybe my interpretation is incorrect)

allejok96:
For reference, the strumming feature created by musikBear: https://www.youtube.com/watch?v=OlBtU0JYvyY

spechtstatt:
Thank you - this is very impressive. 

If I am interpreting this correctly it is based on the quantization setting. But I am not sure if this limitation is necessary. Maybe this can be done in a more free way by using a kind of knob. Like switching into a specific mode with a button and controlling what is applied with a knob. This could basically be done for all the humanization stuff. 

The only difficulty is the initial value of this knob. But maybe this could be overcome in an easy way by resetting the knob value if the selection changes and also when the mode is selected/changed?

spechtstatt:
And I also think that all the nice editing features are sadly too hidden for the users. We could use an additional toolbar as many graphic programs do. Something like this: 

image

The bar would contain some of the existing functionality (to be honest I didn't really understand what "min length is last" and "max length as last" is doing so I left it out for now) and also the new humanization stuff grouped at the left with the humanizer knob.

Every humanizer button would be a disjunctive state for the knob. I am imagine that pressing one of the modes would allow for a flexible and smooth increase or decrease of the selected humaziation effect. 

image

allejok96:
> all the nice editing features are sadly too hidden for the users

I wouldn't say that things in a drop down menu are hidden from users. They menu is in plain sight and if you find it too hard to click twice you can learn the key combo instead, which may be even faster than moving to the toolbar and clicking a button. But that's just my opinion. We could do some form of tear-off menu toolbar thing like most word processors do, and then we'd have the best of both worlds... (very nice concept art btw!)

> I didn't really understand what "min length is last" and "max length as last" is doing

"Last" as in "Note length: Last note". Yeah I know, it's terrible, and it's my fault. It was a hack mainly for the Fill tool, because it may create undesirably long notes. Click the note you want as reference and then click the min/max button. The whole feature is illogical and should probably be reworked.

> I imagine that pressing one of the modes would allow for a flexible and smooth increase or decrease of the selected humanization effect.

Good idea, how about instead of button we have a knob for each humanization property, and as you increase the knob, the notes are humanized live in the editor. When the knob is released it bounces back to zero and humanization is applied. You could undo of course but it wouldn't be like humanization discussed in #1165 where it could be toggled on and off. The good thing with this approach is that you could manually post-edit the humanization, you'd get immediate visual feedback when turning the knobs, and it would solve the position humanization problem.

spechtstatt:
(Damn - I have really to take care what I criticize :-))

musikBear:
> I am not sure if I understand the approaches of @musikBear correctly, but it seems to me that they are more track related (like the arpeggio) and it sounds to me that the chord selection seems to contain some "magic"?

Mine is note-stack dependent, and only for selected notes.
But am i to understand this as both strumming and humanisation now is done, and my code is not necessary at all?

allejok96:
There is no humanization code in this PR... The images by @spechtstatt are just prototypes right? @musikBear what happened to your humanization code?

spechtstatt:
Yes - its (unfortunately) always easier to just make an UI concept art than to finally do the really hard work and implement it :-)

musikBear:
> There is no humanization code in this PR... The images by @spechtstatt are just prototypes right? @musikBear what happened to your humanization code?
That is also in that PR.
The issue was that i could not get my code out of the VirtualBox with Linux-mint where i made it, without destroying my repo. I fought that github-crap for month and made several messed up PRs and failed every time. The result was that my PR showed 1000'th of edits where non was. 
All the code for strumming humanizing and a little thing the helps to place notes precisely _is made_ and works as shown in the video.
I 'just' need to get it flawless into gitHub, without destroying my home-repo.. 'Just'🙄
But in essence, there is no reason to code this. It is done already
If shjuf comes to push, i can create a second new VirtualBox, and only use that for gitHub, then my home-directory should be untouched in the process 
I am actually simultaneously attempting to build in VS2022 in windows!
That is very much a WIP..
Latest next week, i will make a PR and ask for review, unless... 🥴

regulus79:
This is really cool!
Do you plan on finishing this soon? If not, I could try to maybe help implement these features if you want (especially strumming, that sounds great, and I know users have been asking for that.)

PR #6149, opened 2021-09-06
Fixed partially skipping metronome tick sounds; tidied up related var…

frank-a-i:
This PR comes with a possible fix for master's currently, flaky metronome.

See description, discussion and details in the issue thread https://github.com/LMMS/lmms/issues/4038#issuecomment-913057103


frank-ihle:
> Ok. Now the first tick plays. But not after you've stopped it once. Start lmms with the default project. Activate the metronome. tick, tock, tock, tock, tick... etc
> Stop and start again. The first tick is now omitted. This is also the case with the first tick in a loop.

@zonkmachine Could reproduce, after the latest commit not anymore.

zonkmachine:
Works like a charm... almost there. :clown_face: 
[1tick.mmp.txt](https://github.com/LMMS/lmms/files/7131501/1tick.mmp.txt)

Fist tick of a loop when playing looping around... when the start of the loop isn't the first bar in Song Editor.

zonkmachine:
@frank-ihle And while you're at it there's some white-space issues, a tab and three spaces at the end of lines. The Clang police see what we're doing here! Every step...

frank-ihle:
> @frank-ihle And while you're at it there's some white-space issues, a tab and three spaces at the end of lines. The Clang police see what we're doing here! Every step...

I'm not sure how to do that:
- LMMS is seemingly working with other clang formatting rules than the standard set, because almost every line is affected when i analyse
- Couldn't find a top level clang-format file, nor any non-manual applicable formatting actions
- CI/CD pipeline seems alright with the commits except appveyor but I can't see whats the matter with that, since `mixer.*` files are not affected and I assume I'm not getting punished for errors introduced by other devs (i.e. occcurences in other files). However I'm not even sure how to fix the red light on this one at all

superpaik:
Tested on W10 up to 480 bpm, 2 minutes long, and it works on time! Also tested with higher bpms (720) but I can't say if the glitches (minors) came from the computer itself or LMMS (they weren't consistent, the glitches I mean). To me is more than OK.

src/core/AudioEngine.cpp
```diff
@@ -481,24 +479,41 @@ void AudioEngine::handleMetronome()
...
+	if (ticks == m_prevMetronomeStates.second)
 	{
 		return;
 	}
 
-	if (ticks % (ticksPerBar / 1) == 0)
+	/* Metronome sound generation - explanation:
+	 * The modulo operation below figures out the exact point in time when a sound should be played. As this operation
+	 * relies on ".getTicks()" we have to add a certain tick tolerance, since it is not guaranted that between each call
+	 * the tick gets a +1-increment and thus the exact point in time can be missed.*/
```
zonkmachine, line 490: There's a trailing tab on this line. Please remove.

zonkmachine:
I've tested the latest version and it's solid. Ace! Thanks for fixing this!
Approved by me but I'd like someone more experienced to look over the code.


>     * Couldn't find a top level clang-format file, nor any non-manual applicable formatting actions

No, we're in the middle of rearranging the code and people are talking about clang. I'm not into that and I'm not a seasoned coder. What I meant was, just remove the trailing whitespace.  ;)

zonkmachine (approved):

include/AudioEngine.h
```diff
@@ -398,25 +398,27 @@ class LMMS_EXPORT AudioEngine : public QObject
...
 	void doSetAudioDevice( AudioDevice *_dev );
 	AudioDevice * m_audioDev;
 	AudioDevice * m_oldAudioDev;
 	QString m_audioDevName;
 	bool m_audioDevStartFailed;
 
-	// MIDI device stuff
+	// Metronome
+	bool m_metronomeActive;
+	std::pair<tick_t, tick_t> m_prevMetronomeStates; // .first: number of ticks from last play sound, .second: number of ticks from last call
```
Veratil, line 410: Could we place the comment on the line above so it doesn't extend so far out like this.

src/core/AudioEngine.cpp
```diff
@@ -481,24 +479,41 @@ void AudioEngine::handleMetronome()
...
 	}
 
-	if (ticks % (ticksPerBar / 1) == 0)
+	/* Metronome sound generation - explanation:
+	 * The modulo operation below figures out the exact point in time when a sound should be played. As this operation
+	 * relies on ".getTicks()" we have to add a certain tick tolerance, since it is not guaranted that between each call
+	 * the tick gets a +1-increment and thus the exact point in time can be missed.*/	
+
+	// increase this parameter if with complex projects or poor HW resources the metronome rings out of rythm
+	int nToleranceTicks = 5;
```
Veratil, line 493: How do we increase this value outside of modifying the code directly.

src/core/AudioEngine.cpp
```diff
@@ -481,24 +479,41 @@ void AudioEngine::handleMetronome()
...
+	// LMMS allows abrupt jumps in song positions (e.g. song-loops, when marker was put of that loop), this iscached below
+	bool noDoubleTick = m_prevMetronomeStates.first > nToleranceTicks;
+	// identify which sound should be played below
+	bool highTickEvent = ticks % int(TimePos::ticksPerBar() / 1) <= nToleranceTicks;
```
Veratil, line 497: Why `/ 1`? This is just `TimePos::ticksPerBar()`.

src/core/AudioEngine.cpp
```diff
@@ -481,24 +479,41 @@ void AudioEngine::handleMetronome()
...
+	bool lowTickEvent = ticks % int(TimePos::ticksPerBar() / song->getTimeSigModel().getNumerator()) <= nToleranceTicks;
+
+	// handle sound-'state'-machine
```
Veratil, line 500: If this is a 'state' machine we could probably change this to a `switch` instead. I think it would allow easier future additions, possibly adding in more types of ticks to play.

src/core/AudioEngine.cpp
zonkmachine: That's the lingo in master: https://github.com/LMMS/lmms/blob/e3044575bce939cb00989a4cce6163c2111d3f8c/src/core/AudioEngine.cpp#L492

Veratil: It's literally fluff. Even at `-O0` it's removed. Let's not copy this.

include/AudioEngine.h
frank-ihle: - [ ] move comments

src/core/AudioEngine.cpp
frank-ihle: Yep valid point. From what I saw, the assigned tolerance should be alright, as on my machine I hardly exceeded 1 skipping tick. It does increase with rising BPM that I had 2-3 ticks around ~600 BPM but then again, who works at that speed.
I would recommend not to confuse the user with a further settings option and rather handle it, if it ever occurs, via an issue report.

src/core/AudioEngine.cpp
frank-ihle: Are there any comparable benefits for a `switch` ? I'm not sure if the necessary extra lines that enable this structure in that context are an advantage here.

Also, on another PR ( #6119 ) I'm picking this up actually, but it is currently on hold on until this one is clarified

src/core/AudioEngine.cpp
frank-ihle: Yep compiler kills it, as @zonkmachine  mentioned I just took it over from current master. I considered as a reminder for the dev how the timeslots are choosen. I don't have any hard feelings for it.

src/core/AudioEngine.cpp
frank-ihle: - [ ] cleanup

src/core/AudioEngine.cpp
DomClark: This parameter worries me a bit. There's only a narrow possible range: with the maximum time signature denominator of 32, there is a metronome sound every 6 ticks, so this parameter must be 5 or below to avoid skipping. However, with the maximum tempo of 999bpm at a sample rate of 44100Hz and a large buffer size (LMMS has a maximum of 256 internally), there can be up to 4.64 ticks per buffer, so it must be 4 or above to avoid skipping. That there is a working range here is entirely a coincidence, and depends on a lot of other parameters (tempo, buffer size, sample rate, ppq, time signature). I don't like the idea that this could break again if we ever changed the allowed range of these, and would prefer a solution that works universally.

It's worth noticing that we manage to play notes reliably no matter how those parameters might be configured, so a similar technique could be used for the metronome. If you look in `Song::processNextBuffer`, there's a block of code that is guaranteed to run exactly once per tick - no skips nor duplicates:
https://github.com/LMMS/lmms/blob/8a9a2fae622693d1f8d2e976a28c44da59166ff0/src/core/Song.cpp#L337-L345

Handling the metronome there would make it possible to take advantage of the existing code for tick-based timing, simplifying the code for the metronome itself. Indeed, this is @PhysSong's second suggestion in #4038. This also fixes the off-beat problem you noted, since the metronome will be played at the exact same time as the notes themselves. (You can even get a sample-exact metronome by calling `setOffset` on the `SamplePlayHandle`s.)

Finally, I don't believe that hardware performance should have any effect on the metronome - LMMS doesn't skip buffers if it gets overloaded; instead you get an effect where time effectively runs more slowly within LMMS.

PhysSong: @DomClark Thanks for mentioning the code block in `Song.cpp` and other details!

frank-ihle: @DomClark (@PhysSong )
I finally had some more time again and looked into the Song() structure. Basically shifting shouldn't be a problem, however I couldn't figure out how the sounds are orchestrated here. 

I assume this is done around `Track` and I searched its functions and the ones from `TrackContainer` too, but never found the spot where you assign the audio (file, for this case). Is there a nice example?

frank-ihle: > @DomClark (@PhysSong ) I finally had some more time again and looked into the Song() structure. Basically shifting shouldn't be a problem, however I couldn't figure out how the sounds are orchestrated here.
> 
> I assume this is done around `Track` and I searched its functions and the ones from `TrackContainer` too, but never found the spot where you assign the audio (file, for this case). Is there a nice example?


If the committed approach cannot be accepted, we could tweak it in that way, that the next tick event time slot is stored and the sound gets played whenever this timeslot is passed. This of course would introduce new qt-signaling for situations where the user is looping or generally is jumping in time, which doesn't seem much of an issue to me.

If it must be solution II from @PhysSong I need some support as mentioned, I already lost hours in finding a solution without largely rearranging the code.

sakertooth:
Closing in favor of #7483.

PR #6152, opened 2021-09-07
Fix compilation on macOS aarch64

tresf:
This fixes compilation for newer Apple Silicon hardware (e.g. "M1", "ARM64", "aarch64").

- Bumps rpmalloc submodule to fix assembly compilation
   - Fixes `rpmalloc_thread_finalize()` API change
- Re-use `${APPLE_PREFIX}` to fix aarch64 using separate `/opt/homebrew` Homebrew prefix (Intel used `/usr/local`)
- Uses `__SSE__` precompiler flag (instead of hard-coded architectures) for detecting Intel SSE instructions
- Adds additional architectures (ARM, RISCV) to build system, headers and About screen
   - Shim early ARM64 identification for MSVC compiler, don't assume Windows is Intel.

I've also [updated the wiki](https://github.com/LMMS/lmms/wiki/Dependencies-MacOS/_compare/129635dd2d3109758f5a17d69dfe9bd0c098a24a...d48430b1eac3244071c14ff7d9077feed1d36157) to explain a newer qt6 pitfall, specifically, to use `qt5` or `qt@5` and purge any qt6 instances.


cmake/modules/DetectMachine.cmake
```diff
@@ -20,22 +20,40 @@ ENDIF()
...
 ENDIF()
 
 IF(WIN32)
-	IF(WIN64)
-		SET(IS_X86_64 TRUE)
-		SET(LMMS_BUILD_WIN64 TRUE)
-	ELSE(WIN64)
-		SET(IS_X86 TRUE)
-	ENDIF(WIN64)
-
```
tresf, line 35: Explanation: This block seems to be intended for cross-compilers only (e.g. mingw).  MSVC supports ARM64 targets, we'll set these values manually if on MSVC.  Note, `llvm-mingw` offers ARM64 Windows targets, so it's technically possible that an `mingw` identified compiler can fall-through this use-case.

cmake/modules/DetectMachine.cmake
```diff
@@ -20,22 +20,40 @@ ENDIF()
...
+		ENDIF()
+
+		IF(MSVC_TARGET_PLATFORM MATCHES "x64")
+			SET(IS_X86_64 TRUE)
+			SET(LMMS_BUILD_WIN64 TRUE)
+		ELSEIF(MSVC_TARGET_PLATFORM MATCHES "win32")
+			SET(IS_X86 TRUE)
+		ELSEIF(MSVC_TARGET_PLATFORM MATCHES "arm64")
+			SET(IS_ARM64 TRUE)
+		ELSEIF(MSVC_TARGET_PLATFORM MATCHES "arm")
```
tresf, line 51: Explanation: https://cmake.org/cmake/help/latest/generator/Visual%20Studio%2015%202017.html#platform-selection

src/core/MixerWorkerThread.cpp
```diff
@@ -32,7 +32,7 @@
 #include "ThreadableJob.h"
 #include "Mixer.h"
 
-#if defined(LMMS_HOST_X86) || defined(LMMS_HOST_X86_64)
+#if __SSE__
```
tresf, line 35: Explanation: Consistency: https://github.com/LMMS/lmms/blob/2f17c0da2b20ecb9b00f8dcb227d02731325796b/include/denormals.h#L8-L10

cmake/modules/DetectMachine.cmake
```diff
@@ -50,23 +68,57 @@ IF(WIN32)
 		ENDIF()
 
 		unset(MSVC_VER)
+	else()
+		# Cross-compiled
```
PhysSong, line 72: I think you should add a TODO/FIXME comment for ARM here at least.

cmake/modules/DetectMachine.cmake
```diff
@@ -20,22 +20,40 @@ ENDIF()
...
+		# Detect target architecture
+		IF(CMAKE_GENERATOR_PLATFORM)
+			STRING(TOLOWER "${CMAKE_GENERATOR_PLATFORM}" MSVC_TARGET_PLATFORM)
+		ELSE()
+			STRING(TOLOWER "${CMAKE_VS_PLATFORM_NAME_DEFAULT}" MSVC_TARGET_PLATFORM)
+		ENDIF()
+
+		IF(MSVC_TARGET_PLATFORM MATCHES "x64")
+			SET(IS_X86_64 TRUE)
+			SET(LMMS_BUILD_WIN64 TRUE)
```
PhysSong, line 46: Is `LMMS_BUILD_WIN64` only for x86-64 builds, not for ARM64?

cmake/modules/DetectMachine.cmake
```diff
@@ -20,22 +20,40 @@ ENDIF()
...
 SET(LMMS_HOST_X86_64 FALSE)
+SET(LMMS_HOST_ARM32 FALSE)
+SET(LMMS_HOST_ARM64 FALSE)
+SET(LMMS_HOST_RISCV32 FALSE)
+SET(LMMS_HOST_RISCV64 FALSE)
 
 IF(NOT DEFINED WIN64 AND CMAKE_SIZEOF_VOID_P EQUAL 8)
-    SET(WIN64 ON)
+	# TODO: This seems a bit presumptous
+	SET(WIN64 ON)
```
PhysSong, line 30: Do you want this to be defined on 64bit ARM?

PhysSong:
This might be useful for architecture detection: https://www.boost.org/doc/libs/1_77_0/libs/predef/doc/index.html#_boost_arch_architecture_macros

cmake/modules/DetectMachine.cmake
tresf: It's already defined on all 64-bit platforms that resolve to true in this statement.  This is just odd, presumptuous code, one in which my PR does not aim to fix.  The flag seems like it could be refactored out rather easily, hence the `TODO:`.

cmake/modules/DetectMachine.cmake
tresf: Until the build system has been tested against `llvm-mingw` (and all mingw dependencies are cross-compiled, hosted at a PPA) or `vcpkg` is tested using Windows ARM64, this is up in the air.  Personally, I'd discourage obfuscating the two.  Historically `WIN64` and `LMMS_BUILD_WIN64 ` were synonymous with "x86_64".  Moving forward, this is up in the air for the following reasons:
* Windows ARM64 can execute x86_64 binaries
* Windows ARM64 can execute x86 binaries
* Windows ARM64 can execute ARM64 binaries

For these reasons, RemoteVstPlugin64, RemoveVstPlugin32 may need adjustments so that Windows 11 on ARM64 can run 32-bit and 64-bit Intel binaries.

Perhaps a better question is, "Will these changes break our existing MSVC builds?" and I think they will not, and they are preparing us in part for a day when we can add support for ARM64 on Windows as well.  The `MESSAGE()` alone helps considerably when debugging why a MSVC build isn't firing.

tresf:
> This might be useful for architecture detection: https://www.boost.org/doc/libs/1_77_0/libs/predef/doc/index.html#_boost_arch_architecture_macros

Thanks however the next archs to hit mainstream (`aarch64`, `riscv64`) are obfuscated from their 32-bit counterparts.  For this reason, I think these macros are not very helpful for us.

cmake/modules/DetectMachine.cmake
tresf: Note, I can offer Windows 11 ARM64 hardware if someone is interested in improving the above.

PhysSong:
If you consider ARM Windows out of scope, this looks fine to me.

tresf:
> If you consider ARM Windows out of scope, this looks fine to me.

I think it is out of scope for now.  At a future time, I will create a new PR which fixes the `WIN64`, `LMMS_BUILD_WIN64` so that it's easier to understand, read and process on Windows ARM64.

cmake/modules/DetectMachine.cmake
```diff
@@ -20,22 +20,40 @@ ENDIF()
...
+			SET(LMMS_BUILD_WIN64 TRUE)
+		ELSEIF(MSVC_TARGET_PLATFORM MATCHES "win32")
+			SET(IS_X86 TRUE)
+		ELSEIF(MSVC_TARGET_PLATFORM MATCHES "arm64")
+			SET(IS_ARM64 TRUE)
+		ELSEIF(MSVC_TARGET_PLATFORM MATCHES "arm")
+			SET(IS_ARM32 TRUE)
+		ELSE()
+			MESSAGE(WARNING "Unknown target architecture: ${MSVC_TARGET_PLATFORM}")
+		ENDIF()
```
DomClark, line 55: This only works for the Visual Studio generators, so fails to detect the target architecture when using MSVC with another generator, e.g. Ninja. It may be worth taking a look at [how vcpkg tackles this](https://github.com/microsoft/vcpkg/blob/14c814bfc71a8a98cdb1739a09325deca58b98bd/scripts/buildsystems/vcpkg.cmake#L239-L339) - it looks at the path to the compiler, since the MSVC compilers are installed in directories corresponding to their target architecture.

tresf: Thanks, I supposed this could be amended with a check against`VCPKG_TARGET_TRIPLET` to use vcpkg's technique however that assumes the build system uses vcpkg, which I'm not sure we're OK with project-wide.

The vcpkg technique is quite comprehensive, parsing `CMAKE_GENERATOR`, then finding the first `cl.exe`, an even supporting MSVC on Mac with `CMAKE_OSX_ARCHITECTURES` (grabbing the first in the list).  This is nice, however it seems at a glance like a workaround to a ninja bug.  How does one discover which CMake variables are exposed to the Ninja system?

Another question is... . where do we draw the line?  Do we intend to support ninja project-wide?  If so, why hasn't ninja been added as part of the CI so we're addressing these issues before PRs are merged?

The above code was tested on Windows 11 ARM64 running MSVC 2019 and CMake and it works.  I fear we'll be reinventing the wheel if we borrow too much from vcpkg.  Any advice forward is welcome.

DomClark: The Ninja generator doesn't have a concept of target platform like the VS generators do - it simply uses the first C++ compiler it can find in `PATH`. I don't think this counts as a bug - other generators don't do it either, and we have to ask the compiler itself what it targets using `--dumpmachine`. CL doesn't have such a switch as far as I am aware, but it does have a standard installation path.

I don't think it makes sense to add a new CI job just for testing CMake code, if the C++ code will be built with the same compiler and the same target as another job. However, I also don't think we should rule out using any generator other than Visual Studio with MSVC. I would rather we had some generic code that should, at least in principle, work with any generator, rather than relying solely on generator-specific variables such as `CMAKE_GENERATOR_PLATFORM` and `CMAKE_VS_PLATFORM_NAME_DEFAULT`. If certain combinations not tested by CI happen to break, then it should be up to developers with an interest in them to fix them, but we shouldn't rule out such combinations entirely by not providing code for them. I disagree with a "no CI, not interested" approach - we accept fixes for platforms such as Haiku which aren't tested by CI, so why not CMake generators too?

With that said, here's a diff that works for me, copying just the cl part from vcpkg:
```diff
diff --git a/cmake/modules/DetectMachine.cmake b/cmake/modules/DetectMachine.cmake
index e3daa99f3..c35997811 100644
--- a/cmake/modules/DetectMachine.cmake
+++ b/cmake/modules/DetectMachine.cmake
@@ -50,6 +50,15 @@ IF(WIN32)
 			SET(IS_ARM64 TRUE)
 		ELSEIF(MSVC_TARGET_PLATFORM MATCHES "arm")
 			SET(IS_ARM32 TRUE)
+		ELSEIF(CMAKE_CXX_COMPILER MATCHES "amd64/cl.exe$" OR CMAKE_CXX_COMPILER MATCHES "x64/cl.exe$")
+			SET(IS_X86_64 TRUE)
+			SET(LMMS_BUILD_WIN64 TRUE)
+		ELSEIF(CMAKE_CXX_COMPILER MATCHES "arm/cl.exe$")
+			SET(IS_ARM32 TRUE)
+		ELSEIF(CMAKE_CXX_COMPILER MATCHES "arm64/cl.exe$")
+			SET(IS_ARM64 TRUE)
+		ELSEIF(CMAKE_CXX_COMPILER MATCHES "bin/cl.exe$" OR CMAKE_CXX_COMPILER MATCHES "x86/cl.exe$")
+			SET(IS_X86 TRUE)
 		ELSE()
 			MESSAGE(WARNING "Unknown target architecture: ${MSVC_TARGET_PLATFORM}")
 		ENDIF()
```

tresf: > > Another question is... . where do we draw the line? Do we intend to support ninja project-wide? If so, why hasn't ninja been added as part of the CI so we're addressing these issues before PRs are merged?

> I don't think it makes sense to add a new CI job just for testing CMake code, if the C++ code will be built with the same compiler and the same target as another job.

I mean more generically.  It's quite hard to detect these things otherwise and Ninja seems to have remained popular enough to warrant catching these things in advance.

> The Ninja generator doesn't have a concept of target platform like the VS generators do - it simply uses the first C++ compiler it can find in PATH. I don't think this counts as a bug 

I'm sorry, I had misunderstood ninja.  I thought it was a drop-in replacement for CMake, I didn't realize it was a generator replacement.  So yeah, any cmake-generator defined variables would need to be carefully guarded.  As an aside, I find any solutions utilizing `--dumpmachine` to be rube-goldberg, especially considering how tricky it can be to parse the triplet.  It's nice that projects like vkpkg handle this gracefully, but CMake's very purpose is to "manage the build process in an operating system and in a compiler-independent manner".  I think the main mistake here was that I assumed `MSVC` == `VS_GENERATOR`, which until your comment, I thought was a safe assumption.


>  at least in principle, work with any generator, rather than relying solely on generator-specific variables such as CMAKE_GENERATOR_PLATFORM and CMAKE_VS_PLATFORM_NAME_DEFAULT

Yeah, I would agree that this is the goal for any project, but last I checked, we [used it elsewhere](https://github.com/LMMS/lmms/commit/91cf5bd88d3cb0878fa7af50f75a15a51316542c#diff-a2f241a6c284a3a4a0a65605d0752218122615a25930350d35b73bcc563e046dR45). :wink:

>  I disagree with a "no CI, not interested" approach - we accept fixes for platforms such as Haiku which aren't tested by CI, so why not CMake generators too?

The CI recommendation is influenced by how often I see Ninja (and thus the issues it creates) mentioned on Discord, something I failed to contextualize, sorry.

> then it should be up to developers with an interest in them to fix them

Ok, but please understand, this conversation started with a message targeted (seemingly) at the PR author to investigate, quoting:

> This only works for the Visual Studio generators, so fails to detect the target architecture when using MSVC with another generator, e.g. Ninja. **It may be worth taking a look at how vcpkg tackles this** - it looks at the path to the compiler, since the MSVC compilers are installed in directories corresponding to their target architecture.

So in the above context, "it may be worth taking a look" was taken as passing onus onto the PR author, hence asking for passing onus for a CI job, etc.

I realize this ship has sailed (thank you @DomClark for the patch), but I had left this notification marked as unread until I had some time to defend the CI-comments and investigate the generator remarks.

PR #6155, opened 2021-09-10
Update CPU to Pentium 4 for win32 builds

softrabbit:
The CPU feature requirements for any currently supported 32-bit version of Windows (8.1 and 10) are PAE, NX and SSE2. That should mean a green light for bumping the CPU we build for to the minimum one with SSE2, and what better time for this than the release of 1.3?

### DIFF
--- a/cmake/build_win32.sh
+++ b/cmake/build_win32.sh
@@ -31,7 +31,7 @@ fi
 export PATH=$MINGW/bin:$PATH
 export CXXFLAGS="$CFLAGS"
 if [ "$ARCH" == "32" ]; then
-	export CFLAGS="-march=pentium3 -mtune=generic -mpreferred-stack-boundary=5 -mfpmath=sse"
+	export CFLAGS="-march=pentium4 -mtune=generic -mpreferred-stack-boundary=5 -mfpmath=sse"
 fi
 
 CMAKE_OPTS="-DCMAKE_PREFIX_PATH=$MINGW $CMAKE_OPTS"


he29-net:
What about Windows 7? According to https://gs.statcounter.com/os-version-market-share/windows/desktop/worldwide it has still around 15 % market share, and the fact that it is not supported by MS does not mean it won't zombie around for years to come. Its users probably a) won't or can't upgrade to newer Windows for whatever reason (e.g. said CPU incompatibility), b) can't use Linux for whatever reason (e.g. need a proprietary SW that does not work in wine) and c) don't want to toss perfectly working old hardware (e.g. old computer that parents gave to kids to play with).

I'm not saying to stay limited to obsolete CPU features forever, just to consider whether the performance gains etc. are worth it at this time.

softrabbit:
> What about Windows 7? According to https://gs.statcounter.com/os-version-market-share/windows/desktop/worldwide it has still around 15 % market share, and the fact that it is not supported by MS does not mean it won't zombie around for years to come. 

[That statistic is from web analytics](https://gs.statcounter.com/faq#methodology), and AFAIK there are no mainstream web browsers that work without SSE2. So those in the 15% should mostly be OK. Then there's the unknown number of users that actually would be affected, which number I haven't been able to find much to go on. But let's try.

[In 2010, nearly half of Windows 7 installs were 64-bit,](https://techingiteasy.wordpress.com/2010/07/08/microsoft-nearly-half-of-windows-7-installations-are-64-bit/) let's use that as a starting pint. I don't think that number went down with time, but let's be generous and say half still are 32-bit. I'll throw in a guesstimate that half of those then are running on 64-bit hardware, for some reason or another (I've come across systems built like that). Of the ones with 32-bit CPUs maybe half would be using SSE2 systems (P4, Atom, whatever...). That would leave 1 in 8 Windows 7 users affected, and assuming the measured 15% is the unaffected part we'll end up with something around **2% of all Windows systems** to worry about.

Spekular:
For context the windows 32 version of 1.2.2 has 291,187 downloads. If 1% of them are on incompatible systems that's nearly 3,000 users.

What kind of performance boost would be expected from this change? Is it possible to test for SSE2 (or 3 or 4) support and use it if it's there?

softrabbit:
> What kind of performance boost would be expected from this change? Is it possible to test for SSE2 (or 3 or 4) support and use it if it's there?

First of all, this wasn't about performance. I just thought it wouldn't hurt much to bump the baseline to a 20 year old technology that's been a requirement for probably a lot of software for years already. 

From what I can tell, the boost (if any) is probably smaller than what you'd get from swapping GCC for Clang. Not that my tests on one CPU necessarily mean much, but I did some compiles on 32-bit Ubuntu 18.04(*) in Virtualbox and this is what I got:

| | Default (GCC) | O3 native (GCC) | O3 pentium3 (GCC) |  O3 pentium4 (GCC) | O3 pentium3 (Clang) | O3 pentium4 (Clang) | 
|---|---|---|---|---|---|---|
| **Time** | 163.3 | 162 | 155.7 | 158.7 | 140.7 | 140 |
| **Speedup** | 0% | 1% | 5% | 3% | 16% | 17% |

These are user times in seconds for rendering EsoXLB-CPU.mmpz, average of 3 runs. All builds in release mode. GCC 7.5.0, Clang 6.0.0, AMD Phenom II X6 (i.e. SSE, SSE2, SSE3, SSE4a). 

Testing for features should be possible, there's a thing called [function multiversioning](https://gcc.gnu.org/onlinedocs/gcc/Function-Multiversioning.html) that could be useful for that. There are probably other solutions, too.

Spekular:
> > What kind of performance boost would be expected from this change? Is it possible to test for SSE2 (or 3 or 4) support and use it if it's there?
> 
> First of all, this wasn't about performance. I just thought it wouldn't hurt [...]

Are there other benefits than performance? I assumed that was the main point of enabling this stuff, but maybe I've misunderstood.

> Testing for features should be possible, there's a thing called [function multiversioning](https://gcc.gnu.org/onlinedocs/gcc/Function-Multiversioning.html) that could be useful for that. There are probably other solutions, too.

👍

Rossmaxx:
Actually, we are considering dropping first class 32 bit support altogether. So this PR might end up conflicting the project's direction. @tresf merge or close?

tresf:
The title of this PR is a bit confusing since we use the term `WIN32` for all Windows builds (both 32-bit and 64-bit).  With regards to building for Pentium III vs Pentium IV, I think this statement has even less relevance:

> we are considering dropping first class 32 bit support altogether.

... because **if** we DO plan to drop 32-bit support from our automated build system, this change has even less benefit to users, as they would have to compile the software themselves to take advantage of the newer CPU features.

In the spirit of closing out bugs, I think the team is fine with this being merged and closed out, however future hobbyists may find themselves editing this back, so we'd welcome a flag if it appeases future 32-bit users.

PR #6165, opened 2021-09-18
Color submenu, undoable colors, renaming signals

allejok96:
### Visual changes

- Color actions in sub-menu (clips/tracks/channels)
- Consistent wording (clips/tracks/channels)
- Color change is undoable (clips/tracks)
- Random color action for clips
- [Edit]: Clip color applied to all selected clips

image

### Remove `SampleTCOView::contextMenuEvent`

This was just a copy-pasta of `TrackContentObject::contextMenuEvent`. Use `constructContextMenu` instead, like all other TCO sub-classes does, and let the parent class handle the rest.

### Rename some signals and slots

During my work on this PR I was extremely confused by the naming of some signals and slots related to setting colors. This is my motivation for renaming them.

- `Track::`
  - `trackColorChanged` renamed to `changeColor` (it's a SLOT)
  - `trackColorReset` renamed to `resetColor` (it's a SLOT)
- `TrackOperationsWidget::`
  - `useTrackColor` renamed to `resetTCOColors` to clarify the effect it has on TCOs and not on the Track (as this is the TrackOp widget)
- `TrackContentObjectView::`
  - `changeClipColor` (called to display color picker) renamed to `selectColor` to not be confused with....
  - `setColor(color)` a new function to handle color change and undo stuff
  - `useTrackColor` renamed to `resetColor` for sake of consitancy :confused: 
- `TrackContentObject::`
  - `trackColorChanged` SIGNAL renamed to `colorChanged` and emitted on *both* track and TCO color changes (see Communicating "Clear clip colors").

### Communicating a `Track` color change

This applies to `changeTrackColor`, `resetTrackColor` and `randomTrackColor`. The pseudo code illustrates the complicated way a track color change was communicated to a TCOView:

```py
# Slot triggered by menu action emits another signal
TrackOperationsWidget::resetTrackColor()
    emit colorReset()

# Signal proxied via TrackView
TrackView::TrackView()
    connect(trackOperationsWidget, SIGNAL(colorReset), track, SLOT(trackColorReset))

Track::trackColorReset()
    ... do stuff ...
    for (tco: tcos)
        tco->updateColor()

# Conditional check inside TCO, emits yet another signal if true
TrackContentObject::updateColor()
    if (usesTrackColor)
        emit trackColorChanged()

TrackContentObjectView::TrackContentObjectView()
    connect(tco, SIGNAL(trackColorChanged), this, SLOT(update))
```     

The new way:

```py
# Slot triggered by menu action calls Track directly
TrackOperationsWidget::resetTrackColor()
    trackView->track->resetColor()

Track::resetColor()
    ... do stuff ...
    emit colorChanged()

# Signal picked up by TCOView, conditional check inside lambda
TrackContentObjectView::TrackContentObjectView()
    connect(tco->track, Track::colorChanged, this, [this]{ if (tco->usesTrackColor) update() })
```

### Communicating "Clear clip colors"

The old way did not redraw the TCOView if the TCO got changed from anything else than the view, e.g. by `loadSettings`:

```py
# Slot triggered by menu action emits another signal
TrackOperationsWidget::useTrackColor()
    emit colorParented()

# Signal picked up by each TCOView
TrackContentObjectView::TrackContentObjectView()
    connect(trackOpWidget, SIGNAL(colorParented), this, SLOT(useTrackColor));

# update() called manually by TCOView
TrackContentObjectView::useTrackColor()
    tco->useCustomClipColor(false);
    update()

# This does not trigger an update
TrackContentObject::useCustomClipColor()
    ... do stuff ...
```

To make it work with undo/redo, this is the new way:

```py
# Slot triggered by menu action calls each TCO directly
TrackOperationsWidget::resetTCOColors()
    for (tco: trackView->track->tcos)
        tco->useCustomClipColor(false)

TrackContentObject::useCustomClipColor()
    ... do stuff ...
    emit colorChanged()

# update() called as a result of TCO change
TrackContentObjectView::TrackContentObjectView()
    connect(tco, SIGNAL(colorChanged), this, SLOT(update))
```

### DIFF
--- a/include/TrackContentObjectView.h
+++ b/include/TrackContentObjectView.h
@@ -126,8 +126,10 @@ public slots:
 	void remove();
 	void update() override;
 
-	void changeClipColor();
-	void useTrackColor();
+	void selectColor();
+	void randomColor();
+	void resetColor();
+	void updateTrackColor();
 
 protected:
 	enum ContextMenuAction
@@ -231,6 +233,7 @@ private:
 	bool mouseMovedDistance( QMouseEvent * me, int distance );
 	TimePos draggedTCOPos( QMouseEvent * me );
 	int knifeMarkerPos( QMouseEvent * me );
+	void setColor(const QColor color);
 	//! Return true iff TCO could be split. Currently only implemented for samples
 	virtual bool splitTCO( const TimePos pos ){ return false; };
 	void updateCursor(QMouseEvent * me);
--- a/include/TrackContentObjectView.h
+++ b/include/TrackContentObjectView.h
@@ -127,7 +127,7 @@ public slots:
 	void update() override;
 
 	void selectColor();
-	void randomColor();
+	void randomizeColor();
 	void resetColor();
 
 protected:
@@ -232,7 +232,7 @@ private:
 	bool mouseMovedDistance( QMouseEvent * me, int distance );
 	TimePos draggedTCOPos( QMouseEvent * me );
 	int knifeMarkerPos( QMouseEvent * me );
-	void setColor(const QColor color);
+	void setColor(const QColor* color);
 	//! Return true iff TCO could be split. Currently only implemented for samples
 	virtual bool splitTCO( const TimePos pos ){ return false; };
 	void updateCursor(QMouseEvent * me);
--- a/include/TrackContentObjectView.h
+++ b/include/TrackContentObjectView.h
@@ -129,7 +129,6 @@ public slots:
 	void selectColor();
 	void randomColor();
 	void resetColor();
-	void updateTrackColor();
 
 protected:
 	enum ContextMenuAction
--- a/src/gui/TrackContentObjectView.cpp
+++ b/src/gui/TrackContentObjectView.cpp
@@ -125,8 +125,7 @@ TrackContentObjectView::TrackContentObjectView( TrackContentObject * tco,
 	connect( m_tco, SIGNAL( destroyedTCO() ), this, SLOT( close() ) );
 	setModel( m_tco );
 	connect(m_tco, SIGNAL(colorChanged()), this, SLOT(update()));
-	connect( m_tco, SIGNAL( trackColorChanged() ), this, SLOT( update() ) );
-	connect( m_trackView->getTrackOperationsWidget(), SIGNAL( colorParented() ), this, SLOT( useTrackColor() ) );
+	connect(m_trackView->getTrack(), SIGNAL(colorChanged()), this, SLOT(updateTrackColor()));
 
 	m_trackView->getTrackContentWidget()->addTCOView( this );
 	updateLength();
--- a/src/gui/TrackContentObjectView.cpp
+++ b/src/gui/TrackContentObjectView.cpp
@@ -124,6 +124,7 @@ TrackContentObjectView::TrackContentObjectView( TrackContentObject * tco,
 			this, SLOT( updatePosition() ) );
 	connect( m_tco, SIGNAL( destroyedTCO() ), this, SLOT( close() ) );
 	setModel( m_tco );
+	connect(m_tco, SIGNAL(colorChanged()), this, SLOT(update()));
 	connect( m_tco, SIGNAL( trackColorChanged() ), this, SLOT( update() ) );
 	connect( m_trackView->getTrackOperationsWidget(), SIGNAL( colorParented() ), this, SLOT( useTrackColor() ) );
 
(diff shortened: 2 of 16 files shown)


Spekular:
Passing an invalid color to `TrackContentObjectView::setColor` in order to reset feels a bit hacky to me. That aside this looks great to me, especially the UI changes. @russiankumar do you have time to peek at this, or are you busy?

superpaik:
Minor issue. You said "Consistent wording (clips/tracks/channels)", but FX channels still have just "Color" and not "Channel color", as "Clip color" and "Track color".

allejok96:
Yes this was a conscious decision. I realized that the Track color action and Clip color action produces a visually similar result, so I thought it would be better to be explicit in that case, so the user realizes that these are two different menus.

allejok96:
Oh yes, I forgot to mention one important thing in the PR description... Clip color actions are applied for all selected clips! :facepalm:

russiankumar:
I've taken a look, as far as the UI changes go, I think this was much needed and something I overlooked while making the colors feature. However, I'm not sure if the context menu changes you made are related to color features, so I'd suggest pointing that out in the title too, or making it a separate (albeit not very large) PR.

Thanks for your contribution!

Spekular:
> @russiankumar by context menu changes, you mean changing the words?

I'm guessing it's the change from contextMenuEvent to constructContextMenu. As a separate PR that could be merged instantly and not risk being delayed by discussion on the other changes, but I think we can keep it here.

I intend to merge this on Friday (GMT) if no one objects or merges before then.

Spekular (approved):

include/TrackContentObjectView.h
```diff
@@ -231,6 +232,7 @@ protected slots:
 	bool mouseMovedDistance( QMouseEvent * me, int distance );
 	TimePos draggedTCOPos( QMouseEvent * me );
 	int knifeMarkerPos( QMouseEvent * me );
+	void setColor(const QColor color);
```
DomClark, line 235: Top-level const qualifications on parameters have no effect on the function prototype, so are generally omitted for brevity. Or did you mean to pass by const reference instead?

src/gui/TrackContentObjectView.cpp
```diff
@@ -124,8 +124,13 @@ TrackContentObjectView::TrackContentObjectView( TrackContentObject * tco,
...
 	connect( m_tco, SIGNAL( destroyedTCO() ), this, SLOT( close() ) );
 	setModel( m_tco );
-	connect( m_tco, SIGNAL( trackColorChanged() ), this, SLOT( update() ) );
-	connect( m_trackView->getTrackOperationsWidget(), SIGNAL( colorParented() ), this, SLOT( useTrackColor() ) );
+	connect(m_tco, SIGNAL(colorChanged()), this, SLOT(update()));
+
+	connect(m_trackView->getTrack(), &Track::colorChanged, this, [this]
+	{
+		// redraw if TCO uses track color
+		if (m_tco && !m_tco->usesCustomClipColor()) { update(); }
```
DomClark, line 132: Is it possible for `m_tco` to be null here? We don't test for it anywhere else in this file, and it doesn't really make sense to have a `TrackContentObjectView` that doesn't have a `TrackContentObject` to view.
```suggestion
		if (!m_tco->usesCustomClipColor()) { update(); }
```

src/gui/widgets/TrackOperationsWidget.cpp
```diff
@@ -345,15 +354,15 @@ void TrackOperationsWidget::updateMenu()
...
-						tr( "Clear clip colors" ), this, SLOT( useTrackColor() ) );
+
+	QMenu* colorMenu = new QMenu(tr("Track color"), this);
+	colorMenu->setIcon(embed::getIconPixmap("colorize"));
+	colorMenu->addAction(tr("Change"), this, SLOT(changeTrackColor()));
+	colorMenu->addAction(tr("Reset"), this, SLOT(resetTrackColor()));
+	colorMenu->addAction(tr("Pick random"), this, SLOT(randomTrackColor()));
+	colorMenu->addSeparator();
+	colorMenu->addAction(tr("Reset clip colors"), this, SLOT(resetTCOColors()));
+	toMenu->addMenu(colorMenu);
```
DomClark, line 365: Creating the menu like this means it will last until the `TrackOperationsWidget` itself is deleted, so menus will gradually accumulate in memory until the project is changed or the track deleted. Creating it using `QMenu::addMenu` means it will be a child of the other menu, not the `TrackOperationsWidget`, so it will be deleted along with that menu.
```suggestion
	QMenu* colorMenu = toMenu->addMenu(embed::getIconPixmap("colorize"), tr("Track color"));
	colorMenu->addAction(tr("Change"), this, SLOT(changeTrackColor()));
	colorMenu->addAction(tr("Reset"), this, SLOT(resetTrackColor()));
	colorMenu->addAction(tr("Pick random"), this, SLOT(randomTrackColor()));
	colorMenu->addSeparator();
	colorMenu->addAction(tr("Reset clip colors"), this, SLOT(resetTCOColors()));
```

DomClark (commented):
Here's a review. I've made a few naming suggestions, since that's one of the things going on here. I particularly like the refactoring of the update paths - as well as simplifying them, it moves code only used for UI updates out of the core, which helps towards our long-term goal of separating the two.

allejok96:
> Passing an invalid color to TrackContentObjectView::setColor in order to reset feels a bit hacky to me

@Spekular would `QColor*` and `nullptr` feel less hacky?

allejok96:
I fixed the hacky color, implemented the suggestions by @DomClark plus renamed these slots along the same lines:
- `TrackOperationsWidget`
  - `changeTrackColor()` -> `selectTrackColor()`
  - `randomTrackColor()` -> `randomizeTrackColor()`
- `FxLine`
  - `changeColor()` -> `selectColor()`
  - `randomColor()` -> `randomizeColor()`

src/gui/widgets/TrackOperationsWidget.cpp
allejok96: Phew good you found that!

src/gui/TrackContentObjectView.cpp
allejok96: You're right, I forgot this after debugging what turned out to be a [three arg lambda](https://github.com/insilications/clazy-clr/blob/master/docs/checks/README-connect-3arg-lambda.md) segfault

include/TrackContentObjectView.h
allejok96: Guess I meant const reference, but I've changed to const pointer now

Spekular:
> > Passing an invalid color to TrackContentObjectView::setColor in order to reset feels a bit hacky to me
> 
> @Spekular would `QColor*` and `nullptr` feel less hacky?

A bit I guess, but it still feels roundabout for reset to rely on setColor the way it does right now. Most of the code in setColor is just there to handle journaling, which is also what reset wants to get at. Group deletion could benefit from this as well.

Idea with very little respect for syntax:

```
void TrackContentObjectView::journaledTCOVAction(lambda, tcovs)
{
	std::set<Track*> journaledTracks;

	for (auto tcov: tcovs)
	{
		auto tco = tcov->getTrackContentObject();
		auto track = tco->getTrack();

		// TODO journal whole Song or group of TCOs instead of one journal entry for each track
		
		// If only one TCO changed, store that in the journal
		if (tcovs.length() == 1)
		{
			tco->addJournalCheckPoint();
		}
		// If multiple TCOs changed, store whole Track in the journal
		// Check if track has been journaled already by trying to add it to the set
		else if (journaledTracks.insert(track).second)
		{
			track->addJournalCheckPoint();
		}

		lambda(tcov);
		
		tcov->update();
	}

	Engine::getSong()->setModified();
}
```

DomClark:
The ideal choice would be `std::optional<QColor>`, but that needs a rebase on master for C++17. It's also something we plan to use for the custom colour system in general, so could be implemented along with that instead.

Group deletion can't benefit quite the same way from that algorithm - you can't journal deletion of an object on that object itself, since there's no longer an object to restore the state of when you try to undo. Instead, you always need to journal the parent. Also, if we do switch to `QUndoCommand` for history in the future, colour and deletion will use different commands anyway. I think this might be one of those cases where it's best just to copy and paste some code, since trying to generalise too much will complicate things.

I think this PR is good as it is for now.

DomClark (approved):

Spekular:
> Group deletion can't benefit quite the same way from that algorithm - you can't journal deletion of an object on that object itself, since there's no longer an object to restore the state of when you try to undo.

Fair point. @allejok96 I can merge this now unless you want to change anything?

allejok96:
Go ahead

Spekular:
🎉

PR #6167, opened 2021-09-19
Extend denormal protection through DAZ flag to all capable CPUs.

### LINKED ISSUE #6166: High CPU usage
### Bug Summary
As [mentioned on Discord](https://discord.com/channels/203559236729438208/784594576223633478/885624491105980466), the Equalizer makes the CPU usage go crazy with no input. 

#### Steps to reproduce
Load project [owmycpu.mmp.zip](https://github.com/LMMS/lmms/files/7192477/owmycpu.mmp.zip) in a 32-bit build of LMMS, play. Either turn the decay for Equalizer to max (8 seconds) or have the "keep effects running without input" option turned on.

#### Expected behavior
Low CPU usage.

#### Actual behavior
High CPU usage.

#### Screenshot
Screenshot from 2021-09-19 12-57-03

#### Affected LMMS versions
32-bit Windows "nightly" build 1.3.0-alpha.1 on a not quite current Celeron.
Master on Ubuntu 18.04, e.g. 1.3.0-alpha.1.132+gfd18305fb (Linux/i386, Qt5.9., GCC 7.5.0), in Virtualbox on AMD Phenom II

softrabbit:
A long, long time ago denormal protection was added to LMMS, back then it was only fully enabled for builds at a level of SSE3 and higher, because some SSE2 systems could not handle the setting of the DAZ flag. 

A better way would be probing whether the CPU supports the DAZ flag, but back then nobody made the effort, on my part probably because it looked like it would involve a not too small bit of assembly.

This code corrects that and adds the probing, using the info in [Intel® 64 and IA-32 Architectures Software Developer’s Manual, Volume 1: Basic Architecture](https://www.intel.com/content/dam/support/us/en/documents/processors/pentium4/sb/25366521.pdf), chapter 11.6.3, page 325.

### DIFF
--- a/include/denormals.h
+++ b/include/denormals.h
@@ -5,9 +5,8 @@
 #define DENORMALS_H
 
 #ifdef __SSE__
-#if defined(_MSC_VER)
 #include <immintrin.h>
-#elif defined(__GNUC__)
+#ifdef __GNUC__
 #include <x86intrin.h>
 #endif
 
--- a/include/denormals.h
+++ b/include/denormals.h
@@ -26,10 +26,7 @@ int inline can_we_daz() {
 #endif
 
 // Set denormal protection for this thread. 
-// To be on the safe side, don't set the DAZ flag for SSE2 builds, 
-// even if most SSE2 CPUs can handle it. 
 void inline disable_denormals() {
-
 #ifdef __SSE__
   /* Setting DAZ might freeze systems not supporting it */
   if(can_we_daz()) {
--- a/include/denormals.h
+++ b/include/denormals.h
@@ -5,20 +5,22 @@
 #define DENORMALS_H
 
 #ifdef __SSE__
+#if defined(_MSC_VER)
 #include <immintrin.h>
-#include <fxsrintrin.h>
+#elif defined(__GNUC__)
+#include <x86intrin.h>
+#endif
 
 // Intel® 64 and IA-32 Architectures Software Developer’s Manual,
 // Volume 1: Basic Architecture,
 // 11.6.3 Checking for the DAZ Flag in the MXCSR Register
 int inline can_we_daz() {
   alignas(16) unsigned char buffer[512] = {0};
-  #ifdef LMMS_HOST_X86
+#if defined(LMMS_HOST_X86)
   _fxsave(buffer);
-  #endif
-  #ifdef LMMS_HOST_X86_64
+#elif defined(LMMS_HOST_X86_64)
   _fxsave64(buffer);
-  #endif
+#endif
   // Bit 6 of the MXCSR_MASK, i.e. in the lowest byte,
   // tells if we can use the DAZ flag.
   return ((buffer[28] & (1 << 6)) != 0);
--- a/include/denormals.h
+++ b/include/denormals.h
@@ -4,15 +4,15 @@
 #ifndef DENORMALS_H
 #define DENORMALS_H
 
+#ifdef __SSE__
 #include <immintrin.h>
 #include <fxsrintrin.h>
 
-#ifdef __SSE__
 // Intel® 64 and IA-32 Architectures Software Developer’s Manual,
 // Volume 1: Basic Architecture,
 // 11.6.3 Checking for the DAZ Flag in the MXCSR Register
 int inline can_we_daz() {
-  unsigned char buffer[512] __attribute__ ((aligned(16))) = {0};
+  alignas(16) unsigned char buffer[512] = {0};
   #ifdef LMMS_HOST_X86
   _fxsave(buffer);
   #endif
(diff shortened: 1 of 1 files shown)


include/denormals.h
```diff
@@ -4,27 +4,39 @@
...
+  #endif
+  // Bit 6 of the MXCSR_MASK, i.e. in the lowest byte,
+  // tells if we can use the DAZ flag.
+  return( (buffer[28] & (1 << 6)) != 0);
+}
 #endif
 
-
 // Set denormal protection for this thread. 
 // To be on the safe side, don't set the DAZ flag for SSE2 builds,
```
Spekular, line 29: Outdated comment

include/denormals.h
```diff
@@ -4,27 +4,36 @@
...
 #ifdef __SSE__
-#include <xmmintrin.h>
-#endif
-#ifdef __SSE3__
-#include <pmmintrin.h>
+// Intel® 64 and IA-32 Architectures Software Developer’s Manual,
+// Volume 1: Basic Architecture,
+// 11.6.3 Checking for the DAZ Flag in the MXCSR Register
+int inline can_we_daz() {
+  unsigned char buffer[512] __attribute__ ((aligned(16))) = {0};
```
PhysSong, line 15: `__attribute__` doesn't work on MSVC, consider using [`alignas`](https://en.cppreference.com/w/cpp/language/alignas) instead.

include/denormals.h
```diff
@@ -4,27 +4,36 @@
 #ifndef DENORMALS_H
 #define DENORMALS_H
 
+#include <immintrin.h>
+#include <fxsrintrin.h>
```
PhysSong, line 8: This will break builds for non-x86 architectures. Also, including `fxsrintrin.h` directly is not required and not recommended.

include/denormals.h
```diff
@@ -4,27 +4,36 @@
...
+  unsigned char buffer[512] __attribute__ ((aligned(16))) = {0};
+  #ifdef LMMS_HOST_X86
+  _fxsave(buffer);
+  #endif
+  #ifdef LMMS_HOST_X86_64
+  _fxsave64(buffer);
+  #endif
+  // Bit 6 of the MXCSR_MASK, i.e. in the lowest byte,
+  // tells if we can use the DAZ flag.
+  return( (buffer[28] & (1 << 6)) != 0);
```
PhysSong, line 24:
```suggestion
  return ((buffer[28] & (1 << 6)) != 0);
```
or
```suggestion
  return (buffer[28] & (1 << 6)) != 0;
```

include/denormals.h
```diff
@@ -4,27 +4,36 @@
...
-#ifdef __SSE3__
-	/* DAZ flag */
-	_MM_SET_DENORMALS_ZERO_MODE( _MM_DENORMALS_ZERO_ON );
-#endif
-
 #ifdef __SSE__
-	/* FTZ flag */
-	_MM_SET_FLUSH_ZERO_MODE( _MM_FLUSH_ZERO_ON );
+  /* Setting DAZ might freeze systems not supporting it */
+  if(can_we_daz()) {
```
PhysSong, line 32:
```suggestion
  if (can_we_daz()) {
```

include/denormals.h
softrabbit: I think `fxsrintrin.h` was needed for getting the build working on Ubuntu 18.04, so I'm leaving it in for now.

PhysSong: You're right, only `x86intrin.h` includes `fxsrintrin.h` in that case.

softrabbit: I'm going with the recipe at https://newbedev.com/header-files-for-x86-simd-intrinsics, hope it covers all compilers we care about.

include/denormals.h
```diff
@@ -4,27 +4,38 @@
...
 
-
 #ifdef __SSE__
-#include <xmmintrin.h>
-#endif
-#ifdef __SSE3__
-#include <pmmintrin.h>
+#if defined(_MSC_VER)
+#include <immintrin.h>
+#elif defined(__GNUC__)
```
PhysSong, line 10: How about other unusual compilers?

softrabbit: OK. Let's assume any compiler missing from [Wikipedias compiler list](https://en.wikipedia.org/wiki/List_of_compilers#C++_compilers) is irrelevant enough to ignore. That is, until someone using it complains and gets to be the designated dev for that oddity.

Filtering the ones with a green C++17 column, it looks to me that all should be covered. (C++ Builder is LLVM, i.e. clang, EDG makes a front end from what I can tell, not a compiler, and then there's the GCC/Clang/Visual C++ trio). 

Out of the four partially supporting ones, two are for embedded systems (IAR and SEGGER) and Intel is LLVM. Which leaves us with the PGI compiler, now part of the NVidia HPC SDK, which I'm not sure of. Will have to return to that later.

PhysSong: How about always including `immintrin.h` and additionally including `x86intrin.h` for GCC for extra safety?

PhysSong (approved):

PR #6170, opened 2021-09-26
Clean up PianoRoll selection

allejok96:
In order of appearance:

- Convenience functions: `getTick` `yCoordOfKey` `xCoordOfTick` (taken from `paintEvent`)
- `m_select*` members store start/end instead of start/length
- Timeline selection made to work like Song editor - first draw box, then apply (see video).
- If code wants to check for ongoing selection, check `m_action`, don't care about `m_editMode`. This makes `removeSelection()`, unnecessary.
- Bugfix: Do not end selection if Ctrl is released. This fixes bug when ctrl is released, then pressed again.
- Bugfix: Box selection always includes the key where selection started (see video).
- Bugfix: `getKey()` was 2 pixels off, and keys below the bottom line was a whole key off.
- `selectAll()` simply selects all notes, instead of drawing a box around them (see video)

https://user-images.githubusercontent.com/7693838/134797508-6a35399e-1a0a-4c67-8343-1807c29aa180.mp4


Spekular:
> * Timeline selection made to work like Song editor - first draw box, then apply (see video).

Is there any benefit to this change? The current behavior provides clearer feedback on what's selected.

> * Bugfix: Do not end selection if Ctrl is released. This fixes bug when ctrl is released, then pressed again.

I would consider this expected behavior.

> * Bugfix: Box selection always includes the key where selection started (see video).

I tend to click near the horizontal lines when box selecting and expect the box to select notes either above or below the line depending on my selection endpoint. If there's any bug here I'd say it's the fact that selecting near the bottom line of a note includes that note in the selection, rather than snapping the selection edge to the closest grid line.

Rest looks good to me.

allejok96:
> Is there any benefit to this change?

- Consistency, above all.
- Discoverability. Shows the user what this action does, even if there are no notes.
- If we ever implement Esc as a way to cancel a box select, this way will leave the old selection unaltered.

> The current behavior provides clearer feedback on what's selected.

Yes, I like the visual feedback too. But either way it should be consistent with box select (and with song editor)

> I would consider this expected behavior.

I disagree. Actions where a modifier key needs to be pressed *before* the mouse button should *only* finish when the mouse button that started it is released. (In contrast to things like unqunatize, which can be toggled on/off during drag). Pressing Esc, another mouse button or key should, if anything, cancel the action. Just look at file managers, GIMP, Blender and LMMS knob automation.

> I tend to click near the horizontal lines when box selecting

This is a interesting and good point. The only exception would be when clicking to select a single note. Maybe there is some way to combine these two? I will give it a look.

Spekular:
> > Is there any benefit to this change? 
> * Consistency, above all.

Piano roll box select is the one that should be changed then. Both selection modes in the song editor (box and timeline) select during the action, as does (currently) the timeline select in piano roll.

> * Discoverability. Shows the user what this action does, even if there are no notes.

Can be accomplished without removing the selection.

> * If we ever implement Esc as a way to cancel a box select, this way will leave the old selection unaltered.

Ought to be possible either way.

> > I would consider this expected behavior.
> 
> I disagree[,] look at file managers, GIMP, Blender and LMMS knob automation.

I can't say I've ever thought to release ctrl when doing a ctrl+drag action, but you seem to be correct. Carry on then :)

> This is a interesting and good point. The only exception would be when clicking to select a single note.

I would do this by dragging from slightly above it to slightly below. If the selection top & bottom always snaps to the nearest note edge this works and can be rationalized as consistent w/ the song editor. There's also a "select all notes on this key" option when right clicking piano keys.

src/gui/editors/PianoRoll.cpp
```diff
@@ -4153,51 +4071,12 @@ void PianoRoll::setEditMode(int mode)
...
-	if( ! hasValidPattern() )
-	{
-		return;
-	}
-
-	// if first_time = true, we HAVE to set the vars for select
-	bool first_time = true;
-
-	for( const Note *note : m_pattern->notes() )
+	if (hasValidPattern())
```
Veratil, line 4074: I'd rather keep the return early guard so we don't increase indentation needlessly.
```cpp
if (!hasValidPattern()) { return; }
for (Note* note: m_pattern->notes())
{
  note->setSelected(true);
}
```

src/gui/editors/PianoRoll.cpp
```diff
@@ -926,40 +927,17 @@ void PianoRoll::hidePattern( Pattern* pattern )
...
-					m_selectStartTick );
-	}
-	m_selectedKeys = keyNum - m_selectStartKey;
-	if( keyNum <= m_selectStartKey )
-	{
-		--m_selectedKeys;
-	}
-
-	computeSelectedNotes( false );
+void PianoRoll::selectRegionFromPixels( int xStart, int xEnd )
```
Veratil, line 933:
```suggestion
void PianoRoll::selectRegionFromPixels(int xStart, int xEnd)
```

src/gui/editors/PianoRoll.cpp
```diff
@@ -2216,13 +2161,11 @@ void PianoRoll::computeSelectedNotes(bool shift)
...
 
-			const int key = note->key() - m_startKey + 1;
-
 			int pos_ticks = note->pos();
 
 			// if the selection even barely overlaps the note
-			if( key > sel_key_start &&
-				key <= sel_key_end &&
+			if( note->key() >= sel_key_start &&
+				note->key() <= sel_key_end &&
```
Veratil, line 2168:
```suggestion
			if (note->key() >= sel_key_start &&
				note->key() <= sel_key_end &&
```

src/gui/editors/PianoRoll.cpp
```diff
@@ -3616,15 +3505,16 @@ void PianoRoll::paintEvent(QPaintEvent * pe )
...
-	p.setBrush( Qt::NoBrush );
-	p.drawRect(x + m_whiteKeyWidth, y, w, h);
+	if (m_action == ActionSelectNotes)
+	{
+		int x = xCoordOfTick(std::min(m_selectStartTick, m_selectEndTick));
+		int w = xCoordOfTick(std::max(m_selectStartTick, m_selectEndTick)) - x;
+		int y = yCoordOfKey(std::max(m_selectStartKey, m_selectEndKey)) - 1; // high keys are higher up
+		int h = yCoordOfKey(std::min(m_selectStartKey, m_selectEndKey)) - y - 1 + m_keyLineHeight;
+		p.setPen(m_selectedNoteColor);
+		p.setBrush( Qt::NoBrush );
```
Veratil, line 3515:
```suggestion
		p.setBrush(Qt::NoBrush);
```

Veratil (commented):
Not *needed*, but preferred.

Veratil:
I've sat on replying to this until now since I was thinking about the changes (since I've been focused on the piano roll more than any other dev lately).

* I don't think we need to break out the `xCoordOfTick` and `yCoordOfKey` since they are used only in `paintEvent`, ***but*** there are similar class functions in `AutomationEditor` and `StepRecorderWidget` so I'm okay with it
* I think some of the `getTick` calculations aren't using the same formula, but pulling it into a function is a good update
* I like the timeline selection showing the lines 👍 
* I was with Spekular on the Ctrl key release because I was in the same thought that you held it down all the time, but if it doesn't actually do that in the other programs then I'm cool with the enhancement/bugfix
* Not showing the selection box with `selectAll` is probably better anyway since it's suppose to *select all* 😂 
* "bugfix" of selecting by whole key instead of "top" of key is better IMO
* Did you test *all* the different ways to select notes after removing the `m_editMode` check? There are some weird code paths so just want to make sure

Other than that, looks okay to me.

allejok96:
Good thoughts @Veratil.

> `xCoordOfTick` and `yCoordOfKey` ... are used only in `paintEvent`

After realizing this I was about to revert them. But then again they might prove useful in the future for `drawNoteRect` or other stuff extracted from pain[t]Event. The equation is easy to get wrong (which has proven to be the case). So since you say it's ok, I'll keep them.

> Did you test all the different ways to select notes after removing the `m_editMode` check?

No, I will have to double-check the code paths.

allejok96:
> Piano roll box select is the one that should be changed then.

@Spekular, this was more complicated than expected, due to toggle-selection with shift. We'd have to do something like `note->oldSelected()`. Better make that another PR. Then we could add Esc support to that one.

allejok96:
- [x] Code paths checked, no problem. Although discovered that a whole section in `mouseMoveEvent` is unreachable (see FIXME).

- [x] Went loco on `computeSelectedNotes`. Did not change the logic, appart from removing special treatment of 0 length notes, as they don't exist afaik.

- [x] Selection snaps from grid line to grid line, but if it is just one key, it always picks the key where the mouse is. So holding ctrl and clicking a note will always select that note.
  - I'm not a big fan of the logic behind this, but this was the third and best solution I came up with:
  - `mousePressEvent` stores the initial click position in `m_selectStartTick` and `m_selectStartKeyLine`
  - `mouseMoveEvent` handles all the dirty logic, rounding and stuff
  - The `m_selection*` members store the nicely computed and ready-to-use selection bounds

Spekular:
Code wise things LGTM, but I want to test drive this before approving so it doesn't do anything unexpected. O build bot where art thou?

allejok96:
> I would do this by dragging from slightly above it to slightly below. If the selection top & bottom always snaps to the nearest note edge this works and can be rationalized as consistent w/ the song editor. 

Just tested this code again and I disagree. If I start a selection with the mouse hovering a note, that note should *always* be selected, no matter which grid line I was closest to. See problem:

https://user-images.githubusercontent.com/7693838/162499611-28df06b0-6401-4536-9c38-de7e650f42c1.mp4

This actually matches the behavior of Song editor, because it selects whatever the selection box touches. The major difference is that the selection box does not snap to the grid in Song editor.

src/gui/editors/PianoRoll.cpp
```diff
@@ -926,40 +926,18 @@ void PianoRoll::hidePattern( Pattern* pattern )
...
-	m_selectedKeys = keyNum - m_selectStartKey;
-	if( keyNum <= m_selectStartKey )
-	{
-		--m_selectedKeys;
-	}
-
-	computeSelectedNotes( false );
+void PianoRoll::selectRegionFromPixels(int xStart, int xEnd)
+{
+	if (xStart > xEnd) { qSwap(xStart, xEnd); }
```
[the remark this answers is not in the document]
allejok96: Thanks! I realized it's not event needed since `computeSelectedNotes` takes care of that.

allejok96:
I don't care to dig into this again, there are too many changes at once. Let future generations solve this.

PR #6175, opened 2021-09-30
Fix MSVC architecture detection with non-VS generators

DomClark:
See discussion at https://github.com/LMMS/lmms/pull/6152#discussion_r715232318.

Architecture detection code copied from [vcpkg](https://github.com/microsoft/vcpkg/blob/14c814bfc71a8a98cdb1739a09325deca58b98bd/scripts/buildsystems/vcpkg.cmake#L275-L282). We can use `CMAKE_CXX_COMPILER` instead of `find_program(cl)` since we're not in a toolchain file, are guaranteed to be using C++, and want to ensure the detected architecture matches the particular compiler that is in use.


PhysSong (approved):

DomClark:
Merging. This fixes my workflow for now, and seems highly unlikely to break anything else, since it only runs in cases where detection would already be failing with the previous code. Further discussion and PRs are always welcome, of course.

PR #6176, opened 2021-09-30
Split InstrumentTrack

M374LX:
This PR aims to split InstrumentTrack.cpp and InstrumentTrack.h into one pair of files per class, as per #5592.


JohannesLorenz (approved):

JohannesLorenz:
Thanks for the contribution.

DomClark:
This PR downgrades rpmalloc from the latest `develop` to release 1.4.3. Is there a reason for this? Also, CircleCI didn't run for this PR for some reason, but fortunately all passed when merged into master.

M374LX:
No, the rpmalloc downgrade was unintentional.

DomClark:
Ah, okay. I don't think it's a problem - looking at the rpmalloc changes, it should only affect MSYS2, which we don't officially support any more. It's probably something to watch out for in the future though (not just authors, but reviewers too) - submodules seem to have a habit of doing things like this.

PR #6127, opened 2021-08-13
Rename Mixer to AudioEngine

M374LX:
This PR aims to rename Mixer to AudioEngine, as decided in #6089 and #5592.

At present, it leaves the "mixer" element of the configuration file as is, since backwards compatibility needs to be taken into account.

PR #6181, opened 2021-10-02
Rename "mixer" to "audioengine" on lmmsrc.xml

M374LX:
Although #6127 is now merged, the old name "mixer" persists in the lmmsrc.xml configuration file.

### DIFF
--- a/src/core/ConfigManager.cpp
+++ b/src/core/ConfigManager.cpp
@@ -40,9 +40,8 @@
 
 // Vector with all the upgrade methods
 const std::vector<ConfigManager::UpgradeMethod> ConfigManager::UPGRADE_METHODS = {
-	&ConfigManager::upgrade_1_1_90,
-	&ConfigManager::upgrade_1_1_91,
-	&ConfigManager::upgrade_1_2_2,
+	&ConfigManager::upgrade_1_1_90    ,    &ConfigManager::upgrade_1_1_91,
+	&ConfigManager::upgrade_1_2_2
 };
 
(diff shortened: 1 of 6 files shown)


JohannesLorenz (approved):

JohannesLorenz:
Thanks for your contribution. Will merge in 3 days if no one screams.

src/core/ConfigManager.cpp
```diff
@@ -40,7 +40,9 @@
 
 // Vector with all the upgrade methods
 const std::vector<ConfigManager::UpgradeMethod> ConfigManager::UPGRADE_METHODS = {
-	&ConfigManager::upgrade_1_1_90    ,    &ConfigManager::upgrade_1_1_91
```
Spekular, line 43: I think these should stay on one line since that's what we do in the project file upgrader.

Spekular: To be clear, not all on one line, but 2 per line.

src/core/ConfigManager.cpp
```diff
@@ -119,16 +121,46 @@ void ConfigManager::upgrade_1_1_90()
...
+	{
 		setValue("app", "displaydbfs", value("app", "displaydbv"));
 		deleteValue("app", "displaydbv");
 	}
 }
 
+void ConfigManager::upgrade_1_2_2()
+{
+	// Rename mixer to audioengine
+	if (!value("mixer", "audiodev").isNull())
```
Spekular, line 137: What about using a loop or lambda here?

PR #6182, opened 2021-10-03
Fix clip rendering inconsistency

### LINKED ISSUE #6131: Inconsistent clip rendering
### Bug Summary

Notes, waveforms, and automation nodes are not consistently rendered on the X axis.

#### Steps to reproduce

Create a clip containing alternating notes and rests of the same length. In the same way, create a sample clip with alternating sine/silence sections and a flipflopping automation clip. In the example this has been done with third and quarter notes. Sample clips were generated in Audacity with lengths of 3s and 4s, then imported at 20 BPM.

#### Expected behavior

Notes, samples, and automation clips all display their content at the same position and with the same same length, as they are synchronized.

Our bars are 16 pixels wide and 32 pixels tall at 100% zoom, with each bar encompassing the border to it's left. Thus, quarter notes should be 4 pixels wide, and their start points should be 0, 4, 8, and 12 pixels from the leftmost pixel of the bar. Note that borders must be ignored for positions to be correct relative to the timeline.

#### Actual behavior

Notes in note clips are rendered with correct starting points, but are 1 pixel too long (quarter notes are 5 rather than 4 pixels wide).

~~Conversely, lengths but not positions are correct in sample and automation clips. Contents are shifted to the right by two pixels (width of the clip border) compared to note clips and the playhead's position.~~ Fixed in #6182.

These offsets appear to be absolute, as they are unchanged between 100% and 400% zoom.

#### Screenshot

100% Zoom

100% Zoom

400% Zoom
image

Note playhead is precisely on bar start, but sample and automation clips' contents appear to start after current position.
image

#### Affected LMMS versions

Reproduced in 1.3-Alpha.1, more thorough testing in 1.3-Alpha.2. Tested on Windows 10 with scale set to 100%, and DPI scaling handled by LMMS (Properties > Compatibility > Change high DPI settings > Override, scaling performed by application.)

M374LX:
Fixes #6131.

### DIFF
--- a/src/gui/AutomationPatternView.cpp
+++ b/src/gui/AutomationPatternView.cpp
@@ -58,10 +58,10 @@ AutomationPatternView::AutomationPatternView( AutomationPattern * _pattern,
 
 	ToolTip::add(this, m_pat->name());
 	setStyle( QApplication::style() );
-	
+
 	if( s_pat_rec == nullptr ) { s_pat_rec = new QPixmap( embed::getIconPixmap(
 							"pat_rec" ) ); }
-							
+
 	update();
 }
 
@@ -292,7 +292,7 @@ void AutomationPatternView::paintEvent( QPaintEvent * )
 
 	QLinearGradient lin2grad( 0, min, 0, max );
 	QColor col;
-	
+
 	col = !muted ? painter.pen().brush().color() : mutedColor();
 
 	lin2grad.setColorAt( 1, col.lighter( 150 ) );
@@ -369,7 +369,7 @@ void AutomationPatternView::paintEvent( QPaintEvent * )
 
 	p.setRenderHints( QPainter::Antialiasing, false );
 	p.resetTransform();
-	
+
 	// bar lines
 	const int lineSize = 3;
 	p.setPen( c.darker( 300 ) );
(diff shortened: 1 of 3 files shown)


src/gui/AutomationPatternView.cpp
```diff
@@ -375,19 +374,21 @@ void AutomationPatternView::paintEvent( QPaintEvent * )
 	const int lineSize = 3;
 	p.setPen( c.darker( 300 ) );
 
-	for( bar_t t = 1; t < width() - TCO_BORDER_WIDTH; ++t )
```
M374LX, line 378: The variable names "t" and "tx" are leftovers from when "tact" was used instead of "bar", which was fixed in #4878.

Spekular:
Changes so far approved 👍 . Unless I'm mistaken, the current non-format changes simply remove TCO_BORDER_WIDTH as an addend in several calculations (in automation clips x_base was set to border width and then used).

superpaik:
Maybe it's unrelated to this PR, but...
When tested with 4/4 time signature it works well. 
When changed to 3/4 (or anyother time signature diferent than 4/4) and try to draw an automation, it crashes (W10-64bits)

Spekular:
Here's what this looks like for me.
image

- Border width seems unchanged.
- Automation clips appear to be fixed.
- I'm unable to test sample tracks as the CI build crashes when trying to load a sample. I think this is an MSVC build issue rather than anything to do with this PR.

If the border width changes are reverted this can be merged IMO, the sample and automation clip changes should be safe and I trust that the sample track change will work.

M374LX:
@Spekular So, should the suggestion to reduce the borders in https://github.com/LMMS/lmms/issues/6131#issuecomment-899095955 be left out of this PR?

As there is an outer and an inner border, there are two borders in total. Should we keep only one border?

Spekular:
> @Spekular So, should the suggestion to reduce the borders in [#6131 (comment)](https://github.com/LMMS/lmms/issues/6131#issuecomment-899095955) be left out of this PR?

I think so. The offset is a bug, so a fix for that can be merged right away. Changing border width would require some discussion first since it's a matter of taste

> As there is an outer and an inner border, there are two borders in total. Should we keep only one border?

Practically speaking this would be my proposal, since both borders are already 1px wide they can't be reduced without removing one. But again, I think this would require some discussion before we change it.

M374LX:
Clip border changes reverted.

Spekular:
Proper test confirms that sample tracks are also fixed:

image

image

PR #6192, opened 2021-10-08
Disable tooltips through event filter

allejok96:
Continued from #6124. Makes Tooltips option available in the View menu, and remove all usages of `ToolTip.h`.


superpaik:
It works ok on W10.

One thought. I'd say that this option is more of a preference than an option on the "View Menu". I don't see why someone may want to change back and forth this option. But that's a personal point of view.

Not an error "per se". But, as for the code style, I've noticed that you've deleted the trailing and leading spaces in the lines you've modified.  I suppose it's been done with some kind of script. When the code regarding the tooltip is spread in two (or more) lines it still has the spaces. Like in the like 524 in audio_file_processor.cpp
imatge

Spekular:
> > I don't see why someone may want to change back and forth this option.
> 
> Honestly I never reflected about that. There was this commented code block in the view menu that I just uncommented. But thinking about it, I agree with you @superpaik.

I third this opinion, so maybe let's just remove the commented-out code? Other than that this looks good.

allejok96:
> let's just remove the commented-out code

Done, and rebased.

Spekular (approved):

plugins/Sid/SidInstrument.cpp
```diff
@@ -706,15 +705,15 @@ void SidInstrumentView::updateKnobToolTip()
...
-		ToolTip::add( m_voiceKnobs[i].m_crsKnob,
+		 m_voiceKnobs[i].m_crsKnob->setToolTip(
 				QString::number( (int)k->m_voice[i]->m_coarseModel.value() ) +
 				" semitones" );
 	}
-	ToolTip::add( m_volKnob,
+	 m_volKnob->setToolTip(
 					QString::number( (int)k->m_volumeModel.value() ) );
-	ToolTip::add( m_resKnob,
+	 m_resKnob->setToolTip(
```
DomClark, line 716: The space after the opening parenthesis wasn't deleted for each of these lines, so they are indented an extra space.

DomClark (approved):

Spekular:
Looks ready to merge once the conflicts are resolved!

PR #6208, opened 2021-11-02
Fixes #6186: No segfaults with future lmmsrc.xml version

### LINKED ISSUE #6186: Segfault when loading a future lmmsrc.xml version
### Bug Summary
At startup, if the version of the lmmsrc.xml is newer than what the current LMMS version supports, it segfaults.

#### Steps to reproduce
1. Open the ``.lmmsrc.xml`` configuration file on a text editor.
2. Go to the root ``<lmms>`` element (usually the third line).
3. Change the value of the ``configversion`` property to a high number (9 for example) and the value of ``version`` to anything other than the current value.
4. Save the file.
5. Start LMMS.

#### Expected behavior
No segfault

#### Actual behavior
Segfault

#### Affected LMMS versions
Version 1.3.0-alpha.1.141+ge1d5391f2 (Linux/x86_64, Qt 5.15.2, GCC 10.2.1 20201203).

#### Backtrace
```
#0  0x000055555575172a in operator() (__closure=0x7fffffffe088, um=(void (ConfigManager::*)(ConfigManager * const)) 0x6500740061006c, this adjustment 3080307) at /home/<user>/contrib/lmms/src/core/ConfigManager.cpp:145
        this = 0x555556340e90
#1  0x0000555555756159 in std::for_each<__gnu_cxx::__normal_iterator<void (ConfigManager::* const*)(), std::vector<void (ConfigManager::*)()> >, ConfigManager::upgrade()::<lambda(ConfigManager::UpgradeMethod)> >(__gnu_cxx::__normal_iterator<void (ConfigManager::* const*)(), std::vector<void (ConfigManager::*)(), std::allocator<void (ConfigManager::*)()> > >, __gnu_cxx::__normal_iterator<void (ConfigManager::* const*)(), std::vector<void (ConfigManager::*)(), std::allocator<void (ConfigManager::*)()> > >, struct {...}) (__first=..., __last=..., __f=...) at /usr/include/c++/10.2/bits/stl_algo.h:3839
No locals.
#2  0x00005555557517d0 in ConfigManager::upgrade (this=0x555556340e90) at /home/<user>/contrib/lmms/src/core/ConfigManager.cpp:147
        createdWith = {m_version = {static null = {<No data fields>}, d = 0x7fffffffe0e0}, m_major = 1446252240, m_minor = 21845, m_patch = -7920, m_labels = {<QList<QString>> = {<QListSpecialMethods<QString>> = {<No data fields>}, {p = {
                  static shared_null = {ref = {atomic = {_q_value = {<std::__atomic_base<int>> = {static _S_alignment = 4, _M_i = -1}, static is_always_lock_free = true}}}, alloc = 0, begin = 0, end = 0, array = {0x0}}, d = 0x1},
                d = 0x1}}, <No data fields>}, m_compareType = ProjectVersion::Major}
#3  0x000055555575422a in ConfigManager::loadConfigFile (this=0x555556340e90, configFile=...) at /home/<user>/contrib/lmms/src/core/ConfigManager.cpp:547
        cfg_file = <incomplete type>
        dom_tree = {<QDomNode> = {impl = 0x555556341c30}, <No data fields>}
        __PRETTY_FUNCTION__ = "void ConfigManager::loadConfigFile(const QString&)"
        searchPaths = {<QList<QString>> = {<QListSpecialMethods<QString>> = {<No data fields>}, {p = {static shared_null = {ref = {atomic = {_q_value = {<std::__atomic_base<int>> = {static _S_alignment = 4, _M_i = -1},
                        static is_always_lock_free = true}}}, alloc = 0, begin = 0, end = 0, array = {0x0}}, d = 0x55555633e2c0}, d = 0x55555633e2c0}}, <No data fields>}
#4  0x0000555555705ec1 in main (argc=1, argv=0x7fffffffe588) at /home/<user>/contrib/lmms/src/core/main.cpp:741
        coreOnly = false
        fullscreen = true
        exitAfterImport = false
        allowRoot = false
        renderLoop = false
        renderTracks = false
        fileToLoad = {static null = {<No data fields>}, d = 0x7ffff668e920 <QArrayData::shared_null>}
        fileToImport = {static null = {<No data fields>}, d = 0x7ffff668e920 <QArrayData::shared_null>}
        renderOut = {static null = {<No data fields>}, d = 0x7ffff668e920 <QArrayData::shared_null>}
        profilerOutputFile = {static null = {<No data fields>}, d = 0x7ffff668e920 <QArrayData::shared_null>}
        configFile = {static null = {<No data fields>}, d = 0x7ffff668e920 <QArrayData::shared_null>}
        app = 0x5555562f9da0
        qs = {interpolation = AudioEngine::qualitySettings::Interpolation_SincFastest, oversampling = AudioEngine::qualitySettings::Oversampling_2x}
        os = {m_sampleRate = 44100, m_bitRateSettings = {m_bitRate = 160, m_isVariableBitRate = false}, m_bitDepth = OutputSettings::Depth_16Bit, m_stereoMode = OutputSettings::StereoMode_JointStereo, m_compressionLevel = 0.625}
        eff = ProjectRenderer::WaveFile
        pos = {static null = {<No data fields>}, d = 0x28}
        sparam = {sched_priority = 0}
        sa = {__sigaction_handler = {sa_handler = 0x7fffffffe400, sa_sigaction = 0x7fffffffe400}, sa_mask = {__val = {1, 140737488348552, 93824996303150, 93824998019848, 140737318128242, 97, 140737488348224, 1, 140737488348552,
              140737488348568, 93824996303237, 1, 4295032831, 140737488348240, 93824996303259, 98}}, sa_flags = 1435748629, sa_restorer = 0x0}
        destroyEngine = false
        ret = 0
```

JohannesLorenz:


src/core/ConfigManager.cpp
```diff
@@ -139,7 +139,8 @@ void ConfigManager::upgrade()
 	}
 
 	// Runs all necessary upgrade methods
-	std::for_each( UPGRADE_METHODS.begin() + m_configVersion, UPGRADE_METHODS.end(),
+    std::size_t max = std::min((std::size_t)m_configVersion, UPGRADE_METHODS.size());
```
Veratil, line 142: spaces indention here instead of tabs :)

JohannesLorenz: Solved.

Spekular:
Logic looks good. Should we use static_cast instead of a C style cast?

Veratil:
> 
> 
> Logic looks good. Should we use static_cast instead of a C style cast?

Yes

JohannesLorenz:
> > Logic looks good. Should we use static_cast instead of a C style cast?
> 
> Yes

Solved.

JohannesLorenz:
> I suggest also adding a check to determine if the version of the lmmsrc.xml being loaded is supported by the current LMMS version. If it is too new, it should be treated as in invalid file.

Wouldn't this mean that devs like me, who have that problem (because they switched between branches) can't use LMMS anymore?

M374LX:
The check might cause problems for developers when switching branches, but the absence of the check might cause older LMMS versions to interpret newer lmmsrc.xml versions wrongly.

RiedleroD:
Maybe a setting to disable the check? I know it's weird, but a config to enable the config would be a better solution than the other two imo.

Spekular:
> I suggest also adding a check to determine if the version of the lmmsrc.xml being loaded is supported by the current LMMS version. If it is too new, it should be treated as in invalid file.

How do we define "too new" when we can't see the future? I don't really see a solution other than releasing a patch version of 1.2 if/when we find a config version that breaks compatibility, but that doesn't help if users are on existing 1.2 versions.

Should we pop a "config version newer than LMMS's" notification like we do with project file versions? We could even choose to only show it if config parsing encounters an issue.

Either way, I think potentially mis-parsing the file is an improvement over segfaulting, epecially since this issue only happens if you've downgraded. I would say that users who downgrade should be prepared to encounter some weirdness, and the typical user wouldn't do so.

TL;DR: no segfault > segfault, can we merge and handle any other concerns in another PR?

M374LX:
That is right. Fixing the segfault is more important.

The notification about the newer config version is a nice idea, but it is a subject for another PR.

JohannesLorenz:
Agreed that it should be only a notification, and agreed that this PR should be merged separately to fix the segfault.

I will merge if there will be no complains within 24 hours.

src/core/ConfigManager.cpp
```diff
@@ -139,7 +139,8 @@ void ConfigManager::upgrade()
 	}
 
 	// Runs all necessary upgrade methods
-	std::for_each( UPGRADE_METHODS.begin() + m_configVersion, UPGRADE_METHODS.end(),
+	std::size_t max = std::min(static_cast<std::size_t>(m_configVersion), UPGRADE_METHODS.size());
```
PhysSong, line 142: The variable name looks very odd to me. If I were fixing this bug, I would use a `if` clause instead, but that's just a matter of taste.

Issue #6073, opened 2021-07-04
Preset retains solo/muted state


Monospace-V:
### Bug Summary
It is possible for me to solo a track, save a preset, and later drag that preset into a project, and the preset still has solo on.
In other words, this:

https://user-images.githubusercontent.com/76674645/124378914-c0737600-dcd1-11eb-9388-d1db34f3dceb.mp4

#### Steps to reproduce
Step 1: Make a solo instrument track
Step 2: Save track as a preset without unsoloing it
Step 3: Create a new project and drag it into the project

#### Expected behavior
Preset should not retain information like whether track is muted or not.

#### Actual behavior
Preset saves as solo on.

#### Screenshot
![image](https://user-images.githubusercontent.com/76674645/124378941-e13bcb80-dcd1-11eb-9f5a-de1be2c115d8.png)

#### Affected LMMS versions
1.3 alpha, 1.2.0, (Presumably everything in between and before those)

Note: also extends to muted presets.

Demomaker:
Shouldn't this issue be closed, because the fix was merged?

Monospace-V:
Yes. Done.

PR #6214, opened 2021-11-08
Making preset saving oblivious to mute/solo/muteBeforeSolo states.

justlog:
Addresses issue #6073. 

After saving the track settings into dataFile, we should be able to just set the muted/solo/mutedBeforeSolo attributes to 0 as we don't use them for preset saving.


PhysSong:
There might be some better ways to do this, but this looks fine to me, except indentations. Please use tabs instead of spaces in `src/tracks/InstrumentTrack.cpp`.

justlog:
> There might be some better ways to do this, but this looks fine to me, except indentations. Please use tabs instead of spaces in `src/tracks/InstrumentTrack.cpp`.

Fixed.

Spekular:
I think you should probably implement mutedBeforeSolo get/setters for consistency, leaving m_mutedBeforeSolo private. That would mean putting this in the public section
```
bool isMutedBeforeSolo() const
{
	return m_mutedBeforeSolo();
}
```
and this in the public slots:
```
void setMutedBeforeSolo(const bool muted)
{
	m_mutedBeforeSolo = muted;
}
```

These functions are generated automatically for solo and mute via the lines linkes below, but because mutedBeforeSolo isn't an AutomatableModel you can't do the same for it
https://github.com/LMMS/lmms/blob/6a716fa5fd114525b0562cd1788534820eb5d16d/include/Track.h#L60-L61
	
What you should _definitely_ do is put m_soloModel back in the private section.

That aside this looks good to me.

justlog:
> 

Fixed.

Spekular (approved):
LGTM!

PR #6215, opened 2021-11-09
Fixed bug where on MIDI exported the BPM was modulo'd by 256. 

JGHFunRun:
Also fixed formatting of relevant function. (`addTempo` in plugins/MidiExport/MidiFile.hpp)

### DIFF
--- a/plugins/MidiExport/MidiFile.hpp
+++ b/plugins/MidiExport/MidiFile.hpp
@@ -239,10 +239,15 @@ class MIDITrack
 		addEvent(event);
 	}
 	
-	inline void addTempo(uint8_t tempo, uint32_t time)
+	inline void addTempo(uint32_t tempo, uint32_t time)
 	{
-		Event event; event.channel = channel;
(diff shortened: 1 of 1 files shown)


plugins/MidiExport/MidiFile.hpp
```diff
@@ -239,10 +239,15 @@ class MIDITrack
...
-	inline void addTempo(uint8_t tempo, uint32_t time)
+	inline void addTempo(uint32_t tempo, uint32_t time)
 	{
-		Event event; event.channel = channel;
-		event.type = Event::TEMPO; event.time=time; event.tempo = tempo;
+		Event event;
+		event.channel = channel;
+
+		event.type = Event::TEMPO;
+		event.time=time;
```
PhysSong, line 248: If you're going to fix the style, why don't you fix this?
```suggestion
		event.time = time;
```

JGHFunRun: I missed it, I guess I only focused on the obvious of "Hey these are all on one line and they shouldn't be"

Veratil (approved):

Issue #5511, opened 2020-05-21
Trackpad sensitivity for knobs and spinboxes


tresf:
See also:
* #5360 (knobs moving too fast)
* #5507 (natural scrolling is backwards)

When using a trackpad to "scroll" (e.g. a two-finger gesture -- supported by Windows, Linux and Mac), LCD Spinboxes are too sensitive.  Knobs aren't sensitive enough.

Note, the master volume slider feels OK.  According to @PhysSong this is because it uses `QAbstractSlider`.  Perhaps we could borrow some sensitivy logic from there.

Issue #3370, opened 2017-02-19
LCD Widgets reduce value by 2 when finishing trackpad "wheel scroll" action


follower:
Here is a demonstration using the tempo control:
![lmms-extra-change-scroll-wheel-bug](https://cloud.githubusercontent.com/assets/189962/23100116/f6ce2716-f6dc-11e6-8536-63ee529cf8d5.gif)

The following events happen in the gif:

1. When two fingers are placed first on the trackpad (the beginning of the wheel scroll motion) the tempo drops from 140 to 138;
2. After a pause, the normal scrolling action changes the tempo to 128;
3. The fingers are then lifted off the trackpad and the tempo mistakenly changes to 126.

More details on cause and suggested solution to follow.

Umcaruje:
@follower I'm assuming this is a mac issue, cause I can't reproduce on linux, neither with my touchpad not my mouse.

follower:
Yes, it is Mac-only caused by the addition of [`Qt::ScrollPhase` / `QWheelEvent::phase()`](http://doc.qt.io/qt-5/qwheelevent.html#phase) in Qt 5.2.

I've got more details and plan to put together a PR for this specific case but also want to get feedback on implementing a solution for all other occurrences.

(TL;DW: `.delta()` can now return `0` (when scroll phase is begin/end) in addition to positive or negative values which means that all ternary conditionals fail to work correctly.)

I recommend ignoring this ticket until I add the additional information. :)

zonkmachine:
This sounds like a Qt bug.

@follower What version of macOS/Qt do you see this issue with?

follower:
It's not specifically a Qt *bug* but a change in behaviour coupled with a coding idiom/incorrect coding assumption in LMMS. (It could be argued that the non-backward compatible behaviour is a Qt bug but...)

#### Issue: Delta events should be ignored when in phase start/end

Specifically, the idiom is the use of the ternary operator to evaluate mouse wheel delta changes (e.g. [see](https://github.com/LMMS/lmms/blob/28a522806bb1a670b460f0bfbf8d6147910dea72/src/gui/widgets/LcdSpinBox.cpp#L156)):

```
( _we->delta() > 0 ) ? 1 : -1
```

Although there are also places where `if` occurrences use the same assumption.

The problem being that any action should only happen if the `ScrollPhase` value *also* has the correct phase value.

#### Number of occurrences

[...]

Issue #3154, opened 2016-12-15
Mixer Effects Chain Vertical Scroll With Mouse Scrolls Entire Screen


miketurn:
Hello Everyone,

(Kind of a hard thing to search for if anyone has posted about this, so I apologize if this has already been mentioned)

This appears to be a bug to me but maybe not to others?
To replicated my findings
1.) First drag an item off your visible LMMS screen so that it calls up the vertical scroll bar on the right side of your screen.

2.) Next open up your MIXER and put enough effects in your EFFECTS CHAIN to where a veritcal scroll bar is needed in the EFFECTS CHAIN window.

3.) Next use your mouse scroll wheel to scroll down in the EFFECTS CHAIN window

What you will notice is that once you get to the end of the EFFECTS CHAIN windows vertical scrollbar
it appears to jump over to the LMMS windows veritcal scrollbar and the screen will scroll down.

I would imagine it should just stop in the EFFECTS CHAIN and not scroll the entire screen down.
Not a huge issue, but thought I would mention what I found, if this is normal to others, then I guess ignore this post.

Hope this helps,
Thank You

SirBothersome:
Replicated in 1.1.3. I believe you need to fix your mouse setting/sensitivity though, as this is not a bug. You are probably scrolling past the end of the scrollbar. I've included some footage to show you what I mean
[<iframe src="//giphy.com/embed/zQSKITPMMEbo4" width="480" height="244" frameBorder="0" class="giphy-embed" allowFullScreen></iframe><p><a href="https://giphy.com/gifs/zQSKITPMMEbo4">via GIPHY</a></p>](url)

tresf:
@SirBothersome is correct, this is the default behavior on most UI's and canvases which means the underlying question is is this ideal/desired or not.  By default in Qt "ignored" mouse events get throw up the stack to the higher container until one of them responds.

![untitled](https://cloud.githubusercontent.com/assets/6345473/21266570/d5a4b0f0-c374-11e6-8aeb-be86f274e2a6.gif)

From [blog.qt.io](http://blog.qt.io/blog/2006/05/27/mouse-event-propagation/):

[...]

PR #6220, opened 2021-11-19
High resolution scrolling

allejok96:
Handle really sensitive scroll wheels (or touch pads) gracefully.

### Problem

Most widgets only check if a scroll event is positive or negative, and scroll exactly one step. For mice like the MX master this may result in 7 steps per mouse wheel tick, because it has a wheel resolution of 2° (compared to standard 15°).

### Solution

This PR implements a concise way to calculate scrolled steps, while keeping the remainder value, [as suggested by the Qt docs](https://doc.qt.io/qt-5/qwheelevent.html#angleDelta). The remainder is discarded when the widget loses focus.

Widgets are updated to use this new counter logic. This means that editors, knobs and faders may be scrolled with a much finer precision using a hi-res mouse.

This also sets a new minimum scroll speed for knobs and faders (at most 200 steps form start to end). Scrolling really really fast on a standard mouse is also ~faster~ more accurate now.

### Excuses
I've tried to build the counter in a way that using it would require as little code as possible, and no need to remember increasing, decreasing or resetting it. In other words, make it so you can't get it wrong.

- Using `ScrollCounter` is cleaner than having an `int` member in each widget and doing the same calculation in 20 places.
- `ScrollCounter` is a [singleton](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Ri-singleton) because there is only one scroll wheel, so why would each widget have its own instance of it?
- If a widget doesn't check the counter during event handling, the counter won't be increased. This prevents it from silently growing when scroll events are being ignored or propagated to the parent.
- Also, thanks to the event filter, widgets never have to pass `QWheelEvent` to the counter.

As a side note, I've tried a way to intercept `QWheelEvent` and solve this problem higher up in the event chain. But this does not work due to how Qt does (undocumented) event propagation.

### Bugs fixed in this PR
- #3370 (not tested)
- #5511 (a matter of taste)
- Horizontal scroll on most automatable controls decrease their value
- Alt+scroll in piano roll note edit area only decreases the velocity
- Scrolling on empty pattern in song editor while zoomed in creates step notes even if they are invisible

### Other questionable changes
- Song editor's horizontal scroll step size is now based on zoom level
- Allow horizontal scroll in Piano roll's note edit area
- Scrolling inside Song editor never scrolls the MDI window
  - #3154

### DIFF
--- a/include/ScrollCounter.h
+++ b/include/ScrollCounter.h
@@ -28,8 +28,10 @@
 #include <QObject>
 #include <QPoint>
 
+#include "lmms_export.h"
 
-class ScrollCounter: public QObject
+
+class LMMS_EXPORT ScrollCounter: public QObject
 {
 public:
 	static int getStepsX(float stepSize = 120);
--- a/include/ScrollCounter.h
+++ b/include/ScrollCounter.h
@@ -1,0 +1,50 @@
+/*
+ * ScrollCounter.h - helper to handle smooth scroll on widgets
+ *
+ * Copyright (c) 2021 Alex <allejok96/gmail>
+ *
+ * This file is part of LMMS - https://lmms.io
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this program (see COPYING); if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA 02110-1301 USA.
+ *
+ */
+
+#ifndef SCROLL_COUNTER_H
+#define SCROLL_COUNTER_H
+
+#include <QObject>
+#include <QPoint>
+
+
+class ScrollCounter: public QObject
+{
+public:
+	static int getStepsX(float stepSize = 120);
+	static int getStepsY(float stepSize = 120);
+	static void registerWidget(QWidget* widget);
+
+	bool eventFilter(QObject *watched, QEvent *event) override;
+
+private:
+	ScrollCounter();
+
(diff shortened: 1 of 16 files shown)


qnebra:
For some reason MinGW build failed, but I guess you already notice it.

#3154 is annoying as hell when someone workflow had Song Editor at full screen. I am glad it gets fixed.

allejok96:
I tried building it locally with the `lmmsci/linux.mingw32:18.04` container, and it worked just fine :confused:

qnebra:
Trying to build it locally, on Ubuntu 20.04 using MinGW it throws those errors:

```
[ 64%] Linking CXX shared module ../compressor.dll
/usr/bin/x86_64-w64-mingw32-ld: CMakeFiles/compressor.dir/objects.a(CompressorControlDialog.cpp.obj):CompressorControlDialog.cpp:(.text+0x247e): undefined reference to `ScrollCounter::getStepsY(float)'
/usr/bin/x86_64-w64-mingw32-ld: CMakeFiles/compressor.dir/objects.a(CompressorControlDialog.cpp.obj):CompressorControlDialog.cpp:(.text+0x7843): undefined reference to `ScrollCounter::registerWidget(QWidget*)'
make[2]: *** [plugins/Compressor/CMakeFiles/compressor.dir/build.make:241: plugins/compressor.dll] Error 1
make[1]: *** [CMakeFiles/Makefile2:2712: plugins/Compressor/CMakeFiles/compressor.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
[ 65%] Building CXX object plugins/Delay/CMakeFiles/delay.dir/__/Eq/moc_EqFader.cpp.obj
[ 65%] Linking CXX shared module ../crossovereq.dll
[ 65%] Built target crossovereq
[ 65%] Building CXX object plugins/Delay/CMakeFiles/delay.dir/qrc_delay.cpp.obj
[ 65%] Linking CXX shared module ../delay.dll
[ 65%] Built target delay
make: *** [Makefile:152: all] Error 2

```

Clean master was compiled correctly, without issues.

allejok96:
It's a long shot, but what if you change line 32 in `include/ScrollCounter.h` to

    #include "lmms_export.h"
    class LMMS_EXPORT ScrollCounter: public QObject

qnebra:
> It's a long shot, but what if you change line 32 in `include/ScrollCounter.h` to
> 
> ```
> #include "lmms_export.h"
> class LMMS_EXPORT ScrollCounter: public QObject
> ```

It fixes compilation

PR #6235, opened 2021-12-12
Quick fix for spinboxes in MIDI tab

### LINKED ISSUE #6019: MIDI output settings displayed incorrectly
The value of Velocity and Program spinboxes on the MIDI tab are off by 1.
MIDI programs are numbered from 0 to 127 according to the standard, so one would expect to see these values in the ui. But LMMS uses a range of 1-128 where 1 is 0 and 128 is 127. This really can cause headache for users who don't know about it.
Similar thing happens with the velocity too, the difference here is that the actual value is correct and it's the view that shows manipulated values.
2021-05-15-171143_250x70_scrot
Tested on 1.2.2 and master branch.

dj-pixus:
Closes #6019.
Just a stupidly simple improvement.

### DIFF
--- a/src/gui/widgets/InstrumentMidiIOView.cpp
+++ b/src/gui/widgets/InstrumentMidiIOView.cpp
@@ -109,8 +109,7 @@ InstrumentMidiIOView::InstrumentMidiIOView( QWidget* parent ) :
 	midiOutputLayout->addWidget( m_outputProgramSpinBox );
 
 	m_fixedOutputNoteSpinBox = new LcdSpinBox( 3, m_midiOutputGroupBox );
-	m_fixedOutputNoteSpinBox->setDisplayOffset( 1 );
-	m_fixedOutputNoteSpinBox->addTextForValue( 0, "---" );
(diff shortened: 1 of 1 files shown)


PhysSong:
Good catch! I think the fix should be applied to `m_fixedOutputNoteSpinBox` as well which also have the range of 0-127.

superpaik:
Maybe it doesn't have sense, but Base Velocity display (on Midi Tab->Custom Base Velocity) goes from 1 to 127. I'm unsure if it should go from 0 to 127 as well. If it's intended to go from 1 to 127 is ok, because I don't see the point of having a 0 base velocity.
Other displays ok. Automation works ok as well.
Tested on W10

dj-pixus:
The base velocity is ok in my opinion, also because some keyboards don't even use note off messages but send note on with 0 velocity instead, so using 0 as base velocity would result in notes never releasing in that case.
About the fixed output note, well it always sends whatever note was pressed so this feature doesn't even seem to work. That should be fixed first.

PR #6236, opened 2021-12-13
Implement MIDI output note setting

dj-pixus:
Now the fixed output note setting in the MIDI tab actually works and all notes output to the same key if specified.

### DIFF
--- a/src/core/midi/MidiPort.cpp
+++ b/src/core/midi/MidiPort.cpp
@@ -163,6 +163,12 @@ void MidiPort::processOutEvent( const MidiEvent& event, const TimePos& time )
 			outEvent.setVelocity( fixedOutputVelocity() );
 		}
 
+		if( fixedOutputNote() >= 0 &&
+			( event.type() == MidiNoteOn || event.type() == MidiNoteOff || event.type() == MidiKeyPressure ) )
+		{
+			outEvent.setKey( fixedOutputNote() );
+		}
+
 		m_midiClient->processOutEvent( outEvent, time, this );
 	}
 }
@@ -232,6 +238,7 @@ void MidiPort::loadSettings( const QDomElement& thisElement )
 	m_outputControllerModel.loadSettings( thisElement, "outputcontroller" );
 	m_fixedInputVelocityModel.loadSettings( thisElement, "fixedinputvelocity" );
 	m_fixedOutputVelocityModel.loadSettings( thisElement, "fixedoutputvelocity" );
+	m_fixedOutputNoteModel.loadSettings( thisElement, "fixedoutputnote" );
 	m_outputProgramModel.loadSettings( thisElement, "outputprogram" );
 	m_baseVelocityModel.loadSettings( thisElement, "basevelocity" );
 	m_readableModel.loadSettings( thisElement, "readable" );


superpaik:
When output note is set to a value, It randomly skips some notes. It doesn't happen when output note is set to "---".
It happens at any bpm, but if set to a high value it's easier to test.
When the value is entered by keyboard (double click) the displayed value is plus one. That happened before, but I notice that on #6235 is fixed to velocity and maybe the fix is similar.
Values are saved and restored in project file ok.
Tested on W10-64b

dj-pixus:
@superpaik Thanks for testing. I tried it on several different systems, both Linux and Windows, but couldn't reproduce the problem you had. When I set the tempo high enough for notes to start skipping, they also do when "---" is set, but only at like 600 bpm and 16th notes. All that could waste more time with it enabled is just a simple setter function so not sure what is going on. Gonna dive into this further.
Regarding the ui display, that is fixed now along with the others.

superpaik:
It could be a problem of my device (though it's a W10 i5 - 16Gb) but can also be related to some other part of the code. Midi arpegios also skip notes notably. Anyway, it's not a big problem and as you said and it's more common with high rate bpm. Thanks, though, to look into it.

JohannesLorenz (approved):
Code looks all correct.

PhysSong (approved):

he29-net:
I gave it a quick test in SW only (one instrument set for MIDI output, one for MIDI input). Before the fix, the Note spinner had no effect, the receiving instrument got whatever note was triggered on the sender. After the fix, any key pressed on the sender results in the selected note being sent to the receiver, so I would say that works as intended.

As for the skipped notes: I saw some skipped notes if the PC was under heavy load (compiling mesa). Under not as heavy load (but still not idle), I did not see any skipped notes, but once I saw a note that arrived at the output about 30 ms late. When only LMMS was using CPU, I did not see any problems. So in my case it probably works as expected; i.e. notes are skipped to avoid buffer underruns.

I discovered one issue, though: if I change the Note value during playback, there is a high probability that the change occurs while a note is playing. This means the destination never releases the note, since the related "note off" event comes with a new note number.

On the surface, the solution seems to be either a) to send a note-off event for the old note when the spinner value changes, or b) to remember the note used for last note-on event and use it for the note-off event, regardless of the spinner value. But I'm not sure what happens if somebody tries to automate it and somehow ends up having several notes that all need to be released...

Grissess:
A silly idea: maintain a map of (current note pitch) -> (pitch sent to MIDI)--consult that during the NoteOff, and then clear it. (The map could be a 128-element array if performance is more crucial than space.) That should be invulnerable to most of the issues of automating that control, even with changes across held notes, and theoretically with overlapping notes (if the input pitch is in the map, just send the NoteOn to the MIDI pitch currently registered again). This _might_ be obnoxious to someone expecting the automation to take effect _immediately_, but it seems like a bad idea to send NoteOn's because an automation value is changing (especially when it can change _every period_). More importantly, it should work as one expects in the simpler cases.

Should I make a prototype off of this tip to this effect?

PhysSong:
@Veratil To be sure, have you checked #6571 before merging this?

Veratil:
> @Veratil To be sure, have you checked #6571 before merging this?

I completely overlooked it, I apologize. Shall we revert?

PR #6241, opened 2021-12-24
Fix changing LcdSpinBox value changing their initial value

### LINKED ISSUE #6213: Mouse operations on LcdSpinBox modify "reset" value
### Bug Summary

Mouse operations on a LcdSpinBox for changing the value do also change the initial value so the "Reset" operation in the context menu is always showing the actual value. Selecting the "Reset" operation has therefore no effect because it just sets the actual value again.

#### Steps to reproduce

- focus a LcdSpinBox 
- click the left mouse button and move up or down to change the value (alternatively you could also use the mouse wheel)
- the "Reset" menu entry inside the context menu shows always the current value

#### Expected behavior

Changing the value of a LcdSpinBox using a mouse operation should behave like the manual entering of the value in the "Set Value" modal dialog which does not alter the "Reset" value.

#### Actual behavior

Changing the value of a LcdSpinBox using a mouse operation does set the "Reset" value for every change.

#### Screenshot

image

image

#### Affected LMMS versions

Version 1.3.0-alpha - actual master source state 

#### Logs
<details>
<pre>

</pre>
</details>

merrickclay:
This PR aims to fix #6213.
Very simple fix.

### DIFF
--- a/src/gui/widgets/LcdSpinBox.cpp
+++ b/src/gui/widgets/LcdSpinBox.cpp
@@ -120,7 +120,7 @@ void LcdSpinBox::mouseMoveEvent( QMouseEvent* event )
 				model()->value() + m_remainder - fdy / 2.f * model()->step<int>();
 			float floatValRounded = roundf( floatValNotRounded );
 			m_remainder = floatValNotRounded - floatValRounded;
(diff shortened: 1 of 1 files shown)


spechtstatt:
Great that this is fixed!

PR #6249, opened 2021-12-30
Fix ghost notes are too short in PianoRoll

### LINKED ISSUE #6247: Ghost Notes - Grey Note Boxes Come Up A Little Short
When you add Ghost Notes into the Piano Roll if zoom in to the right side of any grey note box you will see that the notes don't go all the way to the vertical beat / bar lines.
I think the reason for this is that active notes have a bright green grip on the right side of each note to adjust the note length and the Ghost Note graphics do not account for space those grips fill.

merrickclay:
Fixes #6247
See diff.


superpaik:
Works ok on W10-64b

PR #6254, opened 2022-01-01
More LV2 buf-size features

JohannesLorenz:
Implement `LV2_BUF_SIZE__boundedBlockLength` and `LV2_BUF_SIZE__fixedBlockLength`.

It was an oversight when we implemented buf-size to not implement these two. This PR implements them and will give access to a couple of plugins. Highlights:

* Helm
* Some new Guitarix effects

The full list:

```
ADLplug
Dexed
dRowAudio: Distortion
dRowAudio: Distortion Shaper
dRowAudio: Flanger
dRowAudio: Reverb
dRowAudio Tremolo
DrumSynth
EQinox
Gxjcm800pre
Gxjcm800preST
GxMetalAmp
GxMetalHead
GxRedeye Big Chump
GxRedeye Chump
GxRedeye Vibro Chump
Helm
HiReSam
JuceOPL
KlangFalter
LUFS Meter
Luftikus
Noize Mak3r
Obxd
OPNplug
PitchedDelay
ReFine
StereoSourceSeparation
Swanky Amp
Tal-Dub-3
Tal-Filter
Tal-Filter-2
Tal-Reverb
Tal-Reverb-II
Tal-Reverb-III
Tal-Vocoder-II
Temper
The Function
The Pilgrim
Vex
Vitalium
Wolpertinger
```

Reviews:

* Code-review would be nice. If no one has time for code-review, will be merged in 7 days without.
* Testing can be done: Check if some of those plugins must be blacklisted (at least, for me, "DrumSynth" causes high CPU and crashes when I open the instrument window).

### DIFF
--- a/src/core/lv2/Lv2Features.cpp
+++ b/src/core/lv2/Lv2Features.cpp
@@ -69,7 +69,15 @@ void Lv2Features::createFeatureVectors()
 	// create vector of features
 	for(std::pair<const char* const, void*>& pr : m_featureByUri)
 	{
-		Q_ASSERT(pr.second != nullptr);
+		/*
+			If pr.second is nullptr here, this means that the LV2_feature
+			has no "data". If this happens here, this means
+			* either that this feature is static
+			  (e.g. LV2_BUF_SIZE__boundedBlockLength)
+			* or that the programmer forgot to use operator[] before feature
+			  vector creation (This can be done in
+			  Lv2Proc::initPluginSpecificFeatures or in Lv2Features::initCommon)
+		*/
 		m_features.push_back(LV2_Feature { pr.first, pr.second });
 	}
 
--- a/src/core/lv2/Lv2Manager.cpp
+++ b/src/core/lv2/Lv2Manager.cpp
@@ -31,6 +31,7 @@
 #include <cstring>
 #include <lilv/lilv.h>
 #include <lv2.h>
+#include <lv2/lv2plug.in/ns/ext/buf-size/buf-size.h>
 #include <lv2/lv2plug.in/ns/ext/options/options.h>
 #include <QDebug>
 #include <QDir>
@@ -78,6 +79,10 @@ Lv2Manager::Lv2Manager() :
 	m_supportedFeatureURIs.insert(LV2_URID__map);
 	m_supportedFeatureURIs.insert(LV2_URID__unmap);
 	m_supportedFeatureURIs.insert(LV2_OPTIONS__options);
+	// min/max is always passed in the options
+	m_supportedFeatureURIs.insert(LV2_BUF_SIZE__boundedBlockLength);
+	// block length is only changed initially in AudioEngine CTOR
+	m_supportedFeatureURIs.insert(LV2_BUF_SIZE__fixedBlockLength);
 
 	auto supportOpt = [this](Lv2UridCache::Id id)
 	{


zonkmachine:
Many of the plugins from the list above are available as binaries here: https://distrho.sourceforge.io/ports.php

zonkmachine:
Largely it seem to work well. Some issues found.

* DrumSynth hangs when you click the instrument track to open the gui.
* Some plugins don't have all variables accessible in our custom gui.
     Example: TAL-Filter-2, there is a window where you set the filter curve. The effect is pretty useless without it. It doesn't look like something that can be fixed easily.


talfilter2

talfilter2

JohannesLorenz:
I'd be for blacklisting them all. In the blacklist, we can add comment for why they are banned, and then when we will enable the LV2 GUI feature later, we will see "this effect is only blacklisted because it's useless without a GUI" and thus remove such effects from the blacklist.

Can you agree? If yes, can you please give me the URIs or push them yourself to the blacklist (with some comments)?

zonkmachine:
> Can you agree? If yes, can you please give me the URIs or push them yourself to the blacklist (with some comments)?

Yes. I'll go through the list and compare with the plugins running in Carla.

JohannesLorenz:
I did not understand the question. I don't use VST, but in LV2, all controls are visible in all effects (like you say), and in all controls, I see an effect of the knobs "Gain", "Tone" and "Volume". For the others knobs, I can't hear what they do (they seem to not affect the sound), but I'm not an expert of how they work.

zonkmachine:
Right, I think these pictures tells it better than I could ever do. Same effects loaded in an lmms instrument and in Carla.

Chump
---
chumplv2

chumpcarla

Big Chump
---
chumpbiglv2

chumpbigcarla

Vibro Chump
---
chumpvibrolv2

chumpvibrocarla

zonkmachine:
Blacklist
----------------------------
* ADLplug - Is hard to recognize as the same plugin as the one with a gui. Makes sound but it lacks good controll.
https://github.com/jpcima/ADLplug
* DrumSynth - Hangs.
urn:juced:DrumSynth
* HiReSam - GUI, No effect.
https://github.com/klangfreund/SpectrumAnalyser
* KlangFalter - Depends on impulse response files (soundfiles) which cannot be loaded.
https://github.com/HiFi-LoFi/KlangFalter
* LUFS Meter - GUI, No effect.
http://www.klangfreund.com/lufsmeter
* StereoSourceSeparation - Unclear. It looks like neither native LV2 nor Carla when editing manually (raw parameter data), manage to control what Carla with native GUI does. Suggest blacklist.
https://github.com/laixinyuan/StereoSourceSepartion
* Tal-Filter-2 - Not enough controllers to be usable. The cutoff control points are missing without the gui.
urn:juce:TalFilter2
* Vex - Pointless without capability to load waves. Only default sine. Suggest blacklist.
urn:juce:Vex

Keep, but not without a comment
----------------------------------------
* Dexed. Works. A DX7 is notoriously hard to program and with this interface it's almost impossible but the controlls are there. Keep.
* Vitalium - Way to advanced to control without the gui but seem to work. Keepish...
* GxMetalHead - A bit odd but it's the same in Carla. Had to put Tone on either extreme and Volume(Gain1) at max. Upstream...
* Tal-Dub-3 - Works. One control is named Unused. I think that one is the 'tab' function (tap misspelled) from the dub 3.
* Tal-Filter - Without gui, controllers missing but can still produce some effect.

Untested or incomplete tesing
------------------------------------
* OPNplug - Can't find, not tested. Is it supposed to install with ADLplug?
* Helm - Untested. Don't have time to compile it. As I recall it I failed last time... ;)

These following ones started and I played around with them and they could be good to go but they're pretty complex. A proper test would take more time than I can afford right now. I'm pretty sure someone will raise an eyebrow at the ui but I'd let them in for now.
* Noize Mak3r.
* Obxd
* Tal-Vocoder-II - Not tested fully. Seems to be working.

See issue report above: https://github.com/LMMS/lmms/pull/6254#issuecomment-1004676003
* GxRedeye Big Chump
* GxRedeye Chump
* GxRedeye Vibro Chump

JohannesLorenz:
> Right, I think these pictures tells it better than I could ever do. Same effects loaded in an lmms instrument and in Carla.

I get it. For example, sinewave seems not available in all 3 LV2 UIs. So I guess that these might/should be disabled in the TTL file. Though I'm not sure what the intention from the Guitarix devs was. Can you please file an issue for guitarix?

zonkmachine:
> I get it. For example, sinewave seems not available in all 3 LV2 UIs. So I guess that these might/should be disabled in the TTL file. Though I'm not sure what the intention from the Guitarix devs was. Can you please file an issue for guitarix?

I can't file an issue for something I don't understand. Are you sure the issue is in Guitarix. It seem to work under Carla?

JohannesLorenz:
> I can't file an issue for something I don't understand. Are you sure the issue is in Guitarix. It seem to work under Carla?

OK, but Carla is the graphical thing, right? You can't see LMMS-style knobs in Carla, correct?

zonkmachine:
> You can't see LMMS-style knobs in Carla, correct?

Correct.

JohannesLorenz:
OK I'll create the issue report there soon.

zonkmachine:
In Carla you can also edit the parameters via the menu instead of the gui. I found that one of the effects isn't consistent. Look at the 'Big Chump' pictures above. There is a feedback switch via the menu but not on the gui. I don't get any action from the feedback switch from any of the Chump effects.

JohannesLorenz:
@zonkmachine Is every 3rd of your images from a carla UI (i.e. non-Gx effects have this UI, too), or is this sill from a Gx UI?

JohannesLorenz:
Found it. They use `pprop:notOnGUI`, and both the native UI and carla UI exploit that, while LMMS does currently not. The questions are:

1. Why do they actually offer the controls if they have no effect?
2. Should LMMS follow the "not on GUI" advice?

I'll ask.

zonkmachine:
> 1. Why do they actually offer the controls if they have no effect?

Isn't this because they provide three slightly different effects from one base set of files. The functions that are unwanted in the simpler versions are simply turned down/bypassed.



> 2. Should LMMS follow the "not on GUI" advice?

I think so, yes. If the function is provided and, as we see in this case, used, then we do it too if it isn't to much of a job.

---

Here is the disappearing feedback knob under Ardour. It works the same in Ardour as in LMMS under Carla. I wonder if this is simply a bug in Guitarix then.
chumpardour

I think the reason for this is that the feedback switch is not in the gui file here:
https://github.com/brummer10/guitarix/blob/master/trunk/src/LV2/gx_redeye.lv2/gxredeye_ui.cpp#L206-#L219

But it isn't given the `lv2:portProperty pprop:notOnGUI`, which probably should have been after this line:
https://github.com/brummer10/guitarix/blob/master/trunk/src/LV2/gx_redeye.lv2/gx_redeye.ttl#L209

In any case I don't see any effect with the feedback switch engaged.

JohannesLorenz:
> I think the reason for this is that the feedback switch is not in the gui file here: https://github.com/brummer10/guitarix/blob/master/trunk/src/LV2/gx_redeye.lv2/gxredeye_ui.cpp#L206-#L219
> 
> But it isn't given the `lv2:portProperty pprop:notOnGUI`, which probably should have been after this line: https://github.com/brummer10/guitarix/blob/master/trunk/src/LV2/gx_redeye.lv2/gx_redeye.ttl#L209

Yeah, this seems inconsistent. Probably a bug that notOnGUI is not in the ttl file for this port.

I'll keep investigating.

JohannesLorenz:
Issue for "Big Chump" submitted, but IMO it should not stop us from merging this PR.

The other open point is to let LMMS use `pprop:notOnGUI`. It's easy to implement. I somehow think this should go into a separate PR, so it can be properly reviewed there. Is this OK?

zonkmachine:
> Is this OK?

👍

JohannesLorenz:
@zonkmachine Sorry, I forgot that we wanted to blacklist some plugins. Do you want to try to add the effects to `src/core/lv2/Lv2Manager.cpp` in a new PR?

zonkmachine:
> Do you want to try to add the effects to `src/core/lv2/Lv2Manager.cpp` in a new PR?

I'm currently in the middle of a move and everthing is boxed up. If you want to have a go at it I think the list I made above is pretty much complete. Otherwise I can have a go at it in a couple of weeks. It may be longer.
In any case I think you can guess by now what my follow up comment is going to be...
Let's push a new nightly!

JohannesLorenz:
@zonkmachine I don't have many of the plugins you listed. Can you please try to write the blacklist PR yourself?

zonkmachine:
> @zonkmachine I don't have many of the plugins you listed. Can you please try to write the blacklist PR yourself?

OK. I'll look into it.

zonkmachine:
> Can you please try to write the blacklist PR yourself?

Fixed in https://github.com/LMMS/lmms/pull/6486

Issue #6224, opened 2021-11-21
Add a simple label for "root" note and selected key/note range


spechtstatt:
### Enhancement Summary

#5868 added also a feature to select a key/note range by dragging an arrow from the left and/or the right side of the piano inside the instrument window. 

The idea of this enhancement is to add a simple informational label for displaying the "root" note and the selected key/note range. Somehow similar as seen here in the piano roll limiter of FL Studio: 

![image](https://user-images.githubusercontent.com/93736385/142763581-84aae045-964b-4f7c-aa2b-ae71f0508716.png)

Taken from here: https://www.image-line.com/fl-studio-learning/fl-studio-online-manual/html/pianoroll_limit.htm

#### Justification

The key/note range is not really visible in the first place and can only be discovered by scrolling to the begin or the end of the piano view which many users may not do without already knowing about the feature.  At least this is the state of the master branch some weeks ago. 

![image](https://user-images.githubusercontent.com/93736385/142763912-f07f7acb-bb2e-4616-ac82-3f2ab999e561.png)

The discussion in #6210 also showed that this feature may not be as obvious as it should be, because this already implemented feature would e.g. allow the assignment of different notes on channel 10 to use a drum pad. Of course this would not be a really convenient way to set up such a midi channel and key/note assignment - for this a different UI solution may be needed - but it would give an indirect hint to users that a possibility to modify this key/note range is at least likely. 

And the additional "root" note display may allow a better visual indication if it was set to a black or white key which is sometimes difficult to see just by the rectangle marker as seen here:

![image](https://user-images.githubusercontent.com/93736385/142764160-d2521576-05ed-4081-967f-0f603a70d2c9.png)

#### Mockup

![image](https://user-images.githubusercontent.com/93736385/142763700-6dda06e8-d0c0-498e-b419-6b1ca0347005.png)

A slight increase in height of the instrument window may be needed though.

Monospace-V:
Would reducing the width of the rectangle be a viable alternative to the root note confusion (assuming we don't add an extra informative bar)?

he29-net:
> Would reducing the width of the rectangle be a viable alternative to the root note confusion (assuming we don't add an extra informative bar)?

Or maybe even turn it to an arrow pointing down? That would also help in situations when the range starts or ends at the base note position (so the icons overlap).

[...]

PR #6259, opened 2022-01-05
added pianoview context menu for first/last/base, fixed small reset v…

spechtstatt:
This implements a context menu for the instrument piano view for setting the base note and the first/last key as discussed in the issue LMMS/lmms#6224. The context menu is only available in the key section of the PianoView. The automation context menu behavior in the black stripe above is not changed. 

image

"Set single key" sets the first and last key so just a single key is left:

image

The menu entries are disabled accordingly if the operation would just set the same value as already set.

Additionally the shape of the base note marker is changed from an rectangle to an arrow:

image

A small bugfix is also applied - the value of the base note an the first/last key was set by a setInitValue call instead of setValue (seems to be the same bug as described in LMMS/lmms#6213). So the reset value reflects now the correct initial value. 

image

The getNoteString method was moved to the PianoView an renamed to getNoteStringByKey to make it possible to use it from both the PianoRoll and the PianoView.

### DIFF
--- a/include/PianoView.h
+++ b/include/PianoView.h
@@ -30,6 +30,7 @@
 
 #include "AutomatableModel.h"
 #include "ModelView.h"
+#include "ModelView.h"
 
 class Piano;
 
@@ -40,9 +41,9 @@ class PianoView : public QWidget, public ModelView
 public:
 	PianoView( QWidget * _parent );
 	virtual ~PianoView() = default;
-
 	static int getKeyFromKeyEvent( QKeyEvent * _ke );
-
+	static QString getNoteStringByKey(int key);
+	const static QString noteStrings[];
 
 public:
 	void keyPressEvent( QKeyEvent * ke ) override;
@@ -67,6 +68,7 @@ private:
 	int getKeyWidth(int key_num) const;
 	int getKeyHeight(int key_num) const;
 	IntModel *getNearestMarker(int key, QString* title = nullptr);
+	void setMarkerKeyValue(IntModel *noteModel, int key_num);
 
 	static QPixmap * s_whiteKeyPm;
 	static QPixmap * s_blackKeyPm;
@@ -81,11 +83,15 @@ private:
 	int m_startKey;					//!< first key when drawing
 	int m_lastKey;					//!< previously pressed key
 	IntModel *m_movedNoteModel;		//!< note marker which is being moved
-
+	int m_lastContextMenuKey;		//!< previous key selected by the context menu
 
 
 private slots:
 	void pianoScrolled( int _new_pos );
+	void setBaseNote();
+	void setFirstKey();
+	void setLastKey();
+	void setSingleKey();
 
 signals:
 	void keyPressed( int );
--- a/src/gui/PianoView.cpp
+++ b/src/gui/PianoView.cpp
@@ -152,6 +152,21 @@ PianoView::PianoView(QWidget *parent) :
 	connect(Engine::getSong(), SIGNAL(keymapListChanged(int)), this, SLOT(update()));
 }
 
+
+/*! \brief list of note strings
+ */
+const QString PianoView::noteStrings[12] = {"C", "C#", "D", "D#", "E", "F", "F#", "G", "G#", "A", "A#", "B"};
+
+
+/*! \brief return a note string (e.g. "C4") based on the key number
+ *  \param key The midi key number
+ */
+QString PianoView::getNoteStringByKey(int key)
+{
+	return noteStrings[key % 12] + QString::number(static_cast<int>(FirstOctave + key / KeysPerOctave));
+}
+
+
 /*! \brief Map a keyboard key being pressed to a note in our keyboard view
  *
  *  \param _k The keyboard scan code of the key being pressed.
@@ -402,33 +417,59 @@ void PianoView::pianoScrolled(int new_pos)
 }
 
 
-
-
-/*! \brief Handle a context menu selection on the piano display view
+/*! \brief Handle a context menu selection on the piano view
  *
  *  \param me the ContextMenuEvent to handle.
  *  \todo Is this right, or does this create the context menu?
  */
 void PianoView::contextMenuEvent(QContextMenuEvent *me)
 {
-	if (me->pos().y() > PIANO_BASE || m_piano == nullptr ||
+	if (m_piano == nullptr ||
 		m_piano->instrumentTrack()->keyRangeImport())
 	{
 		QWidget::contextMenuEvent(me);
 		return;
 	}
 
-	// check which control element is closest to the mouse and open the appropriate menu
 	QString title;
-	IntModel *noteModel = getNearestMarker(getKeyFromMouse(me->pos()), &title);
+	int key_num = getKeyFromMouse(me->pos());
 
-	CaptionMenu contextMenu(title);
-	AutomatableModelView amv(noteModel, &contextMenu);
-	amv.addDefaultActions(&contextMenu);
-	contextMenu.exec(QCursor::pos());
+	if (me->pos().y() > PIANO_BASE)
+	{
+		// context menu for the key area
+		m_lastContextMenuKey = key_num;
+
+		title = QString("Key %1 [%2]").arg(getNoteStringByKey(key_num)).arg(key_num);
+
+		CaptionMenu contextMenu(title);
+		QAction *actionBase = contextMenu.addAction(tr( "Set &base note" ),
+								this, SLOT( setBaseNote() ) );
+		contextMenu.addSeparator();
+
+		QAction *actionFirst = contextMenu.addAction(tr( "Set &first key" ),
+								this, SLOT( setFirstKey() ) );
+		QAction *actionLast = contextMenu.addAction(tr( "Set &last key" ),
+								this, SLOT( setLastKey() ) );
+		contextMenu.addAction(tr( "Set single key" ),
+								this, SLOT( setSingleKey() ) );
+
+		actionFirst->setEnabled(key_num != m_piano->instrumentTrack()->firstKeyModel()->value());
+		actionLast->setEnabled(key_num != m_piano->instrumentTrack()->lastKeyModel()->value());
+		actionBase->setEnabled(key_num != m_piano->instrumentTrack()->baseNoteModel()->value());
+
+		contextMenu.exec(QCursor::pos());
+	}
+	else
+	{
+		// context menu for the black stripe above the keys containing the first/last/base markers
+		IntModel *noteModel = getNearestMarker(key_num, &title);
+		CaptionMenu contextMenu(title);
+		AutomatableModelView amv(noteModel, &contextMenu);
+		amv.addDefaultActions(&contextMenu);
+
+		contextMenu.exec(QCursor::pos());
+	}
 }
-
-
 
 
 // handler for mouse-click-event
@@ -488,8 +529,7 @@ void PianoView::mousePressEvent(QMouseEvent *me)
 			}
 			else
 			{
-				m_movedNoteModel->setInitValue(static_cast<float>(key_num));
-				if (m_movedNoteModel == m_piano->instrumentTrack()->baseNoteModel()) { emit baseNoteChanged(); }	// TODO: not actually used by anything?
+				setMarkerKeyValue(m_movedNoteModel,key_num);
 			}
 		}
 		else
@@ -500,6 +540,39 @@ void PianoView::mousePressEvent(QMouseEvent *me)
 		// and let the user see that he pressed a key... :)
 		update();
 	}
+}
+
+
+void PianoView::setMarkerKeyValue(IntModel *noteModel, int key_num)
+{
+	noteModel->setValue(static_cast<float>(key_num));
+	if (noteModel == m_piano->instrumentTrack()->baseNoteModel()) { emit baseNoteChanged(); }	// TODO: not actually used by anything?
+}
+
+
+void PianoView::setBaseNote()
+{
+	setMarkerKeyValue(m_piano->instrumentTrack()->baseNoteModel(), m_lastContextMenuKey);
+	update();
+}
+
+void PianoView::setFirstKey()
+{
+	setMarkerKeyValue(m_piano->instrumentTrack()->firstKeyModel(), m_lastContextMenuKey);
+	update();
+}
+
+void PianoView::setLastKey()
+{
+	setMarkerKeyValue(m_piano->instrumentTrack()->lastKeyModel(), m_lastContextMenuKey);
+	update();
+}
+
+void PianoView::setSingleKey()
+{
+	setMarkerKeyValue(m_piano->instrumentTrack()->firstKeyModel(), m_lastContextMenuKey);
+	setMarkerKeyValue(m_piano->instrumentTrack()->lastKeyModel(), m_lastContextMenuKey);
+	update();
 }
 
 
@@ -596,7 +669,7 @@ void PianoView::mouseMoveEvent( QMouseEvent * _me )
 			else if (m_movedNoteModel != nullptr)
 			{
 				// upper section, move the base / first / last note marker
-				m_movedNoteModel->setInitValue(static_cast<float>(key_num));
+				setMarkerKeyValue(m_movedNoteModel,key_num);
 			}
 		}
 		// and let the user see that he pressed a key... :)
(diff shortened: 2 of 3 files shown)


superpaik:
It works ok on W10-64b.
One behaviour that I don't know if it's intended. Start and end arrows can cross resulting in a "no" active keys on the keyboard. I don't see the use case of that behaviour.
On the GUI aspect, it's difficult to see the start or end arrow when it's over a note name. I don't know if it could be solved by adding a black (background colour) border to the arrow. It's not a big deal, though.
imatge

allejok96:
I can understand why you didn't change the context menu of the black stripe, but let's brainstorm for a sec...

The black stripe is where you click to move the arrow. So context menu that moves the arrow should intuitively be there too (besides on the piano key).

There are other situations where a automatable control could benefit from special options in its context menu (for example the tempo box or time signature). I don't if there is a clean way to do this...

spechtstatt:
Yes - it's where I started with the context menu first, but there is this functionality of getting the nearest marker for displaying the automation context menu.


This gives us two possibilities:

1) mix two different positions (marker and current key) in one context menu.

2) remove the "nearest" functionality and display the marker menu exactly (including key menu) and the key menu only in all other cases.


I also wanted to display the key which gets to be selected in the header - I did this for the marker menu header so you see the marker key.


So I tried 2) first, but I found it not very practical to use because it is relatively difficult to target the marker exactly with the right menu button. 1) on the other hand is difficult for displaying what is selected (marker) and what gets to be selected (key) and I wanted to avoid writing the key to each of the new entries (maybe a subheader would solve this)


Another thing is that the selection of a key on the piano roll is easier than on the black bar. Of course we could remove the black bar restriction and show a combined context menu everywhere.


So I get your argument, but I am not sure how to merge this two approaches (nearest marker and exact key including an intuitive selection) in one context menu in a clean way.

spechtstatt:
@superpaik: thank you for the feedback.

Yes - I also discovered the "cross key" behaviour, but it was already present so I didn't touch it. But I can fix it in the course of this PR.

Regarding the black border: good idea. I will try if this improves the visibility.

superpaik:
I understand the logic of having the context menu where the icon is displayed, but to me is better to have the context menu on the piano keys, because the action is "Set this exact key as... (base note, first, last key)". If context menu is on the black stripe that is no so clear.

allejok96:
I think this PR is a big improvement. My concern is not that it is illogical with the new piano menu. Instead I think it may become so logical that users never realize that the black bar contains some extra features, because no one will ever click on it.

 Now people will argue that if you want to automate or MIDI-control base note you will know about this feature already. Still, could we "force" new users to discover these features, and not turn it into just another dev only feature? I'm trying to think of ways to find a middle ground.

Problem that spawned this PR: Users can set base note but does not discover the note range setting. How about a TextFloat similar to the master volume? Print base note and note range. Whenever you click the black bar you'll see that there is a range setting possible.

Problem created by this PR: Users spoiled with piano menu, will never click the black bar and discover all of its interactivity. What about forcing users to click it by removing the base note option from the menu? A bit of a regression yes, but is it really that different? With the new arrow it will still be an improvement, more visually clear that the bar is some form of control.

spechtstatt:
@allejok96 : couldn't it be that the context menu on the piano roll would motivate people to try if there is also a context menu on the black bar? Or that they just click there because they don't expect that there would be a different context menu?

And the question is generally if the new context menu can even solve the underlying problem that the first/last key including the automation menu is not very discoverable in the first place.

For the automation menu the markers itself may need a different style so it would be more obvious that they are automatable controls - more 3D - like the usual knobs.

spechtstatt:
Some proposals:

- move the text to the keys to avoid the  marker/text "collision" at all
- use a thin "slider" line
- use green to have a stronger indication that the marker is a kind of (automatable) control

The "full" version:

image


Or keep the white but with the thin "slider" line:

image


Or simply classical but keep the text in the keys:

image

superpaik:
The first proposal from @spechtstatt it's the best to me, though I know it may be not easy to move note names to keys.
But if that is not possible, having the markers in green it's a good solution to show users that there is "something" in there.

spechtstatt:
Funny thing: the classic theme has green markers already.

image

spechtstatt:
I have applied the changes so far - looks like this for now:

image

I think it is more obvious now that these markers are a kind of control thing. 

The text rotation was a nightmare though :-)

I also addressed the "crossing key" bug. It is not possible anymore to cross the first/last key with the mouse and also the key context menu disables the entry accordingly.

And I added also the missing shortcut identifier for the "Set single key" menu.

spechtstatt:
By the way: not sure if another master rebase was a good idea?

spechtstatt:
Should I change this PR to "ready for review" or are there any change requests regarding the implementation?

allejok96:
For some reason this PR includes the mixer rename commit. As I understand it, the only way you can push after a rebase is with `push --force`, but it doesn't look like you've done that here.

spechtstatt:
So is it bad to do a master rebase from the lmms repo on the personal feature branch?

I mean I could create a new branch and just sync my changes again (and add a new PR) if this is easier.

How do you normally handle it if you want to update a branch?

Because I am still not a git expert - far from it :-)

allejok96:
I have no idea what's going on. It must be some github issue... `git diff` locally doesn't show all these changes, and merging went just fine. I have done rebases before without this happening. Still, were you required to `git push --force` ?

allejok96:
Tested this
```sh
git checkout master
git pull upstream master
git checkout -b messyPR
git rebase spechstatt/feature/pianoview-context-menu
git push --set-upstream origin/messyPR
```
Tried making a PR. It included your commits followed by Broken datafile commit and the Clipboard rename commit (which shouldn't show as a part of the PR). So I removed those two and tried again:
```
git reset --hard 'HEAD^^'  # careful with this
git push --force
```
Now your commit was the last one. Tried making a PR and everything seemed fine.

PhysSong:
> So is it bad to do a master rebase from the lmms repo on the personal feature branch?

Not in this situation.

spechtstatt:
I hope that I have fixed the commit issues - "files changed" looks now normal here. 

I repaired it like @allejok96 was suggesting. I used "git reflog" to identify the head xx number before the whole mess started. Then I used this number for the git "reset --hard HEAD@{xx}" command followed by a "git push -f" call. 

So this leads back to my question if I should  set this PR "ready for review" now or if there are any further issues regarding my implementation?

src/gui/PianoView.cpp
```diff
@@ -502,7 +544,52 @@ void PianoView::mousePressEvent(QMouseEvent *me)
...
+
+void PianoView::setBaseNote()
+{
+	setMarkerKeyValue(m_piano->instrumentTrack()->baseNoteModel(), m_lastContextMenuKey);
+	update();
+}
+
+void PianoView::setFirstKey()
+{
+	if (setMarkerKeyValue(m_piano->instrumentTrack()->firstKeyModel(), m_lastContextMenuKey)) update();
```
allejok96, line 578: The updates here are not necessary as they will be triggered by

    connect(m_piano->instrumentTrack()->firstKeyModel(), SIGNAL(dataChanged()), this, SLOT(update()));

src/gui/PianoView.cpp
```diff
@@ -502,7 +544,52 @@ void PianoView::mousePressEvent(QMouseEvent *me)
...
+	{
+		if (m_piano->instrumentTrack()->firstKeyModel() == noteModel && !firstMarkerAllowed(key_num)) return false;
+		if (m_piano->instrumentTrack()->lastKeyModel() == noteModel && !lastMarkerAllowed(key_num)) return false;
+	}
+	noteModel->setValue(static_cast<float>(key_num));
+	if (noteModel == m_piano->instrumentTrack()->baseNoteModel()) { emit baseNoteChanged(); }	// TODO: not actually used by anything?
+	return true;
+}
 
+bool PianoView::firstMarkerAllowed(int key_num)
```
allejok96, line 559: Totally IMO, but I find these functions overkill

include/PianoView.h
```diff
@@ -30,6 +30,7 @@
 
 #include "AutomatableModel.h"
 #include "ModelView.h"
+#include "ModelView.h"
```
allejok96, line 33: Double

src/gui/PianoView.cpp
```diff
@@ -502,7 +544,52 @@ void PianoView::mousePressEvent(QMouseEvent *me)
...
 
+bool PianoView::setMarkerKeyValue(IntModel *noteModel, int key_num, bool ignoreConstraints)
+{
+	if (!ignoreConstraints)
+	{
+		if (m_piano->instrumentTrack()->firstKeyModel() == noteModel && !firstMarkerAllowed(key_num)) return false;
+		if (m_piano->instrumentTrack()->lastKeyModel() == noteModel && !lastMarkerAllowed(key_num)) return false;
+	}
+	noteModel->setValue(static_cast<float>(key_num));
+	if (noteModel == m_piano->instrumentTrack()->baseNoteModel()) { emit baseNoteChanged(); }	// TODO: not actually used by anything?
```
allejok96, line 555: I'd say remove it if it's not used anywhere, when we're redoing this anyway.

src/gui/PianoView.cpp
```diff
@@ -402,33 +419,59 @@ void PianoView::pianoScrolled(int new_pos)
...
-}
+	if (me->pos().y() > PIANO_BASE)
+	{
+		// context menu for the key area
+		m_lastContextMenuKey = key_num;
+
+		title = QString("Key %1 [%2]").arg(getNoteStringByKey(key_num)).arg(key_num);
 
+		CaptionMenu contextMenu(title);
+		QAction *actionBase = contextMenu.addAction(tr( "Set &base note" ),
```
allejok96, line 447: Code formatting: remove whitespace

include/PianoView.h
```diff
@@ -67,6 +68,9 @@ class PianoView : public QWidget, public ModelView
 	int getKeyWidth(int key_num) const;
 	int getKeyHeight(int key_num) const;
 	IntModel *getNearestMarker(int key, QString* title = nullptr);
+	bool setMarkerKeyValue(IntModel *noteModel, int key_num, bool ignoreConstraints = false);
```
allejok96, line 71: Make `setMarkerKeyValue` private because it's only used internally?

allejok96 (changes requested):
Generally very good, just a few things that can be cleaned out.

allejok96:
> should set this PR "ready for review" 

Looks good. Go for it! Wait. I reviewed it anyway. :smile:

src/gui/PianoView.cpp
```diff
@@ -43,6 +43,8 @@
 #include <QPainter>
 #include <QPainterPath>
 #include <QVBoxLayout>
+#include <QtWidgets>
```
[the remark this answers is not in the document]
spechtstatt: ok (I am just so used to this automatic using/reference management in C# IDEs)

src/gui/PianoView.cpp
spechtstatt: It is because the functionality is used from two different places - for the context menu itself and also when setting the value with constraints.

include/PianoView.h
allejok96: Oh sorry :sweat_smile:

allejok96 (commented):
I think we could trim it down just a bit more... I had this test code lying around.

src/gui/PianoView.cpp
```diff
@@ -502,7 +542,50 @@ void PianoView::mousePressEvent(QMouseEvent *me)
...
+{
+	setMarkerKeyValue(m_piano->instrumentTrack()->lastKeyModel(), m_lastContextMenuKey);
+}
+
+
+void PianoView::setSingleKey()
+{
+	setMarkerKeyValue(m_piano->instrumentTrack()->firstKeyModel(), m_lastContextMenuKey, true);
+	setMarkerKeyValue(m_piano->instrumentTrack()->lastKeyModel(), m_lastContextMenuKey, true);
+	update();
```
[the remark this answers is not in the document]
spechtstatt: This works really quite nice (the ignoreConstraints was a little bit hacky)

allejok96:
Everything looks good, for what it's worth

src/gui/PianoView.cpp
```diff
@@ -402,33 +417,59 @@ void PianoView::pianoScrolled(int new_pos)
...
 
-	CaptionMenu contextMenu(title);
-	AutomatableModelView amv(noteModel, &contextMenu);
-	amv.addDefaultActions(&contextMenu);
-	contextMenu.exec(QCursor::pos());
-}
+	if (me->pos().y() > PIANO_BASE)
+	{
+		// context menu for the key area
+		m_lastContextMenuKey = key_num;
```
DomClark, line 440: Rather than passing this to the various menu handlers via a member variable, I think it would be better to pass it as a normal parameter. You can do this by connecting the menu item to a lambda and capturing `key_num`:
```cpp
contextMenu.addAction(tr("Foo"), this, [this, key_num]() { setFoo(key_num); });
```

spechtstatt: So it would be possible to write the following without the SLOT macro or would it still be needed somehow?

```c++
QAction *actionFirst = contextMenu.addAction(tr("Set &first key"), this, [this, key_num]() { setFirstKey(key_num); });
```

DomClark: That should work, no `SLOT` macro required. It's even possible to move `setFirstKey` and the others out of `private slots:` and into a normal `private:` section, since they're only called as normal functions now.

(The `SLOT` macro just converts its argument into a string, which `addAction` passes to `connect` internally, where it is used to look up the actual slot at runtime. This is a legacy style from Qt4; Qt5 supports passing a slot directly as a member function pointer, or passing an object implementing `operator()` such as a lambda.)

spechtstatt:
@allejok96: this PR still shows that a change is requested from your side but I am not able to find the change request and I am also not able to resolve it. Is there anything you can do about it?

allejok96 (approved):

allejok96:
Ah sorry, LGTM

include/PianoView.h
```diff
@@ -86,8 +89,7 @@ class PianoView : public QWidget, public ModelView
 	int m_startKey;					//!< first key when drawing
 	int m_lastKey;					//!< previously pressed key
 	IntModel *m_movedNoteModel;		//!< note marker which is being moved
-
-
+	int m_lastContextMenuKey;		//!< previous key selected by the context menu
```
DomClark, line 92: This is no longer used - remove it?

include/PianoView.h
```diff
@@ -71,7 +71,10 @@ class PianoView : public QWidget, public ModelView
 	int getKeyX( int _key_num ) const;
 	int getKeyWidth(int key_num) const;
 	int getKeyHeight(int key_num) const;
-	IntModel *getNearestMarker(int key, QString* title = nullptr);
+	IntModel *getNearestMarker(int key);
+	void setMarkerKeyValue(IntModel *noteModel, int key_num);
+	bool firstMarkerAllowed(int key_num);
+	bool lastMarkerAllowed(int key_num);
```
DomClark, line 77: These could all be private.

include/PianoView.h
```diff
@@ -45,9 +45,9 @@ class PianoView : public QWidget, public ModelView
 public:
 	PianoView( QWidget * _parent );
 	~PianoView() override = default;
-
```
DomClark, line 48: I personally prefer having a space between constructors/destructors and the rest of the members, but I don't think we have a strict rule about this.

src/gui/instrument/PianoView.cpp
```diff
@@ -97,7 +97,7 @@ PianoView::PianoView(QWidget *parent) :
 	m_piano(nullptr),                /*!< Our piano Model */
 	m_startKey(Key_C + Octave_3*KeysPerOctave), /*!< The first key displayed? */
 	m_lastKey(-1),                   /*!< The last key displayed? */
-	m_movedNoteModel(nullptr)        /*!< Key marker which is being moved */
```
DomClark, line 100: I think this comment is fine to leave in place.

DomClark (commented):
Sorry for the massive delay in getting round to reviewing this again; allejok96 reminded me earlier today about it.

This basically all looks good to me now. I've left a few comments on things that could be cleaned up or tweaked, but nothing major.

include/PianoView.h
allejok96: Btw shouldn't the destructor be virtual here (even if that wasn't part of this PR)?

DomClark: It already is, hence the `override`. It's preferred not to specify `virtual` along with `override`: see [the entry in the Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-override).

spechtstatt:
Does anyone have an idea how I can fix the mvsc errors?

DomClark:
The MSVC builds were failing due to an upstream problem. It has now been fixed, so I've restarted them. They should hopefully pass without further changes.

messmerd:
I tested this PR on Linux Mint 20.3, and everything seems to be working fine.

However, I have one suggestion:
In the `PianoRoll::drawNoteRect` method at line 1010, if you change it from this
```c++
if( n->selected() )
{
    col = QColor( selCol );
}
const int borderWidth = borders ? 1 : 0;
```
to this:
```c++
if (n->selected())
{
    col = QColor( selCol );
}
else if (!m_midiClip->instrumentTrack()->isKeyMapped(n->key()))
{
    const int gray = qGray(col.rgb());
    col.setRgb(gray, gray, gray, col.alpha());
}

const int borderWidth = borders ? 1 : 0;
```
It will draw disabled notes as grayscale when they are not selected.
I think it looks a lot better that way. It makes it clear to the user when notes are outside the first/last range.

Issue #5016, opened 2019-06-07
Adjusting the Master Pitch range to above/below 12 semitones


jshamble:
Can there be a way to increase/decrease the range of the master pitch slider to above 12 semitones and below 12 semitones to +/- N semitones (where N is user-defined, similar to how adjusting the pitch-bend slider works).

thmueller64:
I would like to work on this, if it is okay.

thmueller64:
Here is my suggestion for the basic layout. I wanted to follow the design of the Sf2 Player mentioned in the issue. What do you think about the layout? I did some basic testing by adjusting the range and the master pitch during playback and, from what I can tell, it is behaving as expected. Next, I'll look at the behavior when automating the master pitch and adjusting the range simultaneously. When this is done, a way to save and load the new parameters has to be implemented.

![master_pitch_demo](https://user-images.githubusercontent.com/64359888/80748778-c26a0400-8b25-11ea-89fa-5c6bceb65d30.gif)

I tried to structure/track the requirements and possible issues. Please feel free to discuss the approach:
- [ ] Basic layout of the range adjustment field 
- [x] Test manual adjustment of master pitch and -range during midi playback
- [ ] Automation range must be able to adjust dynamically (--> what should happen when range is reduced, but automation exceeds the new threshold?)
- [ ] Ability to save and load projects using this option, considering possible implications (e.g., what happens when a project using the feature is imported with an older version of lmms?)

jshamble:
@thmueller64 Looks good so far, thanks for picking this up!

In regards to the layout is seems fine to me, although it kind of stands out as it's on its own, perhaps to rectify this and balance the scales we can a similar range selector to the master volume slider to the left of it as well, I've opened a PR for that as well: https://github.com/LMMS/lmms/issues/5479

------

At a deeper level I would also think why is there no way for the user to manually edit the range of values in the automation editor? I would think that would be a bit more involved though, as perhaps it would require some sanitation of the user input values, I see there is a similar PR https://github.com/LMMS/lmms/issues/4563 that has been opened for that already though.

------

In regards to the functionality: when the range is reduced, I would follow the suit of how automation functions for the pitch bend slider in the soundfont player as shown in the original post. 

1) I add the pitch bend to the automation editor by [ctrl + drag and dropping the slider] set the range to 10, add in a random automation value

![1](https://user-images.githubusercontent.com/6364311/80764850-705dba00-8af6-11ea-8561-f854dded0064.PNG)

[...]

PR #6262, opened 2022-01-07
Add spinbox for adjusting master pitch range. (#5016)

thmueller64:


superpaik:
It works correctly on W10-64b
The only thing that is not clear to me is why the range goes from 4 to 64 (when you double-click to set a value) and no from 1 to 64.

Issue #6271, opened 2022-01-13
VeSTige seems to save the original path of the assigned VST plugin


JimmyKenMerchant:
### Bug Summary
VeSTige seems to save the original path of the assigned VST plugin in the system
#### Steps to reproduce
1. Make a project using VeSTige with loading any VST instrument (VSTi) plugin.
2. Save the project.
3. Transfer the project (a .mmpz file) to another computer or system, .e.g, Linux to Windows OS. No matter whether another one has the VSTi plugin or not.
4. On another computer or system, open the project.
5. Read error messages and you can see the displayed path is your old one.
#### Expected behavior
VeSTige should search VST plugins using the path you set in "VST-PLUGIN DIRECTORY" on the "Setup LMMS" menu of LMMS.
#### Actual behavior
VeSTige searches VST plugins using the original path in a .mmpz file. The original path may include a user name of the old computer or system.
#### Screenshot
N/A
#### Affected LMMS versions
1.2.2 on Linux x86_64
#### Logs
N/A

Monospace-V:
VST-Plugin directory is for VST-FX. Even with FX, I think it saves exact path. Caching VST plugin locations etc so as to run them from any path is a feature that has been suggested before. This is a virtual duplicate of other suggestions, that have mentioned saving relative VST path (issue 4646) (can be done, I think, if VST is saved in Samples folder of lmms), all the way to being able to run a VST if detected wherever in the system it is (issue 4907).
Requesting you to look for others asking the same thing before filing bug reports.

JimmyKenMerchant:
> VST-Plugin directory is for VST-FX. Even with FX, I think it saves exact path. Caching VST plugin locations etc so as to run them from any path is a feature that has been suggested before. This is a virtual duplicate of other suggestions, that have mentioned saving relative VST path (issue 4646) (can be done, I think, if VST is saved in Samples folder of lmms), all the way to being able to run a VST if detected wherever in the system it is (issue 4907). Requesting you to look for others asking the same thing before filing bug reports.

this is an issue specified on VeSTige. Saving an absolute path of the old path in a project file may expose a user name of Linux. This issue may cause any privacy breach.

JimmyKenMerchant:
[...]

PR #6275, opened 2022-01-14
VeSTige Searches Original Path of Assigned VST Plugin in Former Syste…

JimmyKenMerchant:
…m When Opening Project File #6271

A Patch to fix this issues minimally.

I think 1.3.0-a has good modification. However, we would also consider forward compatibility. In this patch, the path for VSTi will be rewritten to the path you configured. Besides, the path in the former system won't be saved in a project file. Running test is on Ubuntu 20.04.3 LTS.

Note that no change in vestigeInstrument::saveSettings makes forward compatibility between 1.2.2 and this patch. However, it should be changed if your community thinks a saved path causes a privacy breach.

### DIFF
--- a/plugins/vestige/vestige.cpp
+++ b/plugins/vestige/vestige.cpp
@@ -181,13 +181,13 @@ vestigeInstrument::~vestigeInstrument()
 
 void vestigeInstrument::loadSettings( const QDomElement & _this )
 {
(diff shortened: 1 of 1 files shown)


PhysSong:
Closing since we're no longer updating 1.2 branches.

PR #6276, opened 2022-01-14
midioss: Use unbuffered I/O to access the device node.

alarixnia:
This fixes problems reading from MIDI devices on NetBSD - the
input blocked until a certain number of notes have been read
rather than returning immediately when a single note is received.

### DIFF
--- a/src/core/midi/MidiOss.cpp
+++ b/src/core/midi/MidiOss.cpp
@@ -40,9 +40,9 @@ MidiOss::MidiOss() :
 	// only start thread, if opening of MIDI-device is successful,
 	// otherwise isRunning()==false indicates error
 	if( m_midiDev.open( QIODevice::ReadWrite |
-		QIODevice::Unbuffered | QIODevice::ExistingOnly ) ||
+		QIODevice::Unbuffered ) ) ||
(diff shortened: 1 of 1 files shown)


hselasky:
Looks good to me.

src/core/midi/MidiOss.cpp
```diff
@@ -39,8 +39,10 @@ MidiOss::MidiOss() :
 {
 	// only start thread, if opening of MIDI-device is successful,
 	// otherwise isRunning()==false indicates error
-	if( m_midiDev.open( QIODevice::ReadWrite ) ||
-					m_midiDev.open( QIODevice::ReadOnly ) )
+	if( m_midiDev.open( QIODevice::ReadWrite |
+		QIODevice::Unbuffered ) ) ||
```
PhysSong, line 43: Oops, I missed a typo which breaks build.
```suggestion
		QIODevice::Unbuffered ) ||
```

PhysSong: Fixed in 877a93198f0793fb6e236c3694c7a02fc2527e57.

alarixnia: Thanks!

PR #6278, opened 2022-01-14
Rename TCO to Clip in the clipboard data

allejok96:
Should be safe to change this, unless somebody wants to drag-and-drop between different LMMS versions...


JohannesLorenz (approved):
Code review finished.

Yes, this is safe, since all the changes are only drag-drop related, and the strings you replaced are not used in any other files.

Testing is to be done, I will try that.

JohannesLorenz (approved):
Tested. Works.

JohannesLorenz:
Thanks for the submission! Merging...

PR #6226, opened 2021-11-28
Rename TCO and related to Clip

M374LX:
Rebasing #6044.

Renaming "pattern" to something else on the data file is a task for another PR. Thus, this one does not touch the data file.

As in #6044, this PR does not include renaming BB to Pattern.

PR #6279, opened 2022-01-14
Fix broken data file upgrade

M374LX:
Addresses the data file upgrade breakage introduced in #6226.

### DIFF
--- a/include/AutomationClip.h
+++ b/include/AutomationClip.h
@@ -149,7 +149,7 @@ public:
 	void saveSettings( QDomDocument & _doc, QDomElement & _parent ) override;
 	void loadSettings( const QDomElement & _this ) override;
 
-	static const QString classNodeName() { return "automationclip"; }
+	static const QString classNodeName() { return "automationpattern"; }
 	QString nodeName() const override { return classNodeName(); }
 
 	ClipView * createView( TrackView * _tv ) override;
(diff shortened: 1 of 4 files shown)


JohannesLorenz (changes requested):
Code review finished. Issues:

* In `DataFile.cpp`, `ELEMENTS_WITH_RESOURCES` still uses `sampleclip`.

Testing is required after that.

JohannesLorenz (approved):
Sorry, last time I have forgotten to check the headers. But you seem to have them complete now, too. LGTM. Now, testing review required.

JohannesLorenz:
Test review minimum:

The following should work with all kinds of clips (Automation Clip, BB Clip, MidiClip (these are the clips with MIDI notes), SampleClip):

* drag/drop a clip
* save a file with that clip type
* load a file (which you created with this PR) with that clip type
* load a file (which you did not create with this PR) with that clip type

M374LX:
Another problem found: when dragging and dropping a sample clip from the Song Editor to the Beat+Bassline Editor, it does not work.

JohannesLorenz:
Oh, OK. Then I'd say let's merge this? It can't get much worse :smiling_face_with_tear:

M374LX:
For me, it is ready to be merged.

PR #6284, opened 2022-01-15
Rename BB to Pattern

allejok96:
Here goes nothing (updated Jan 22)

- "Beat+Bassline" -> "Pattern"
- `BB*` -> `Pattern*`
- `BBTrackContainer` -> `PatternStore`
- Explain what `PatternStore` is with some ASCII art

Notable other changes:
- A bit of sane code formatting, including a few `static_cast`
- README: "Pattern-Editor for creating **beats and patterns**"
- `PatternEditor.cpp` line 89: I took the liberty of changing the tooltip "Add Beat/Bassline" to "**New** pattern".
- Removed redundant `PatternEditor::removePatternView()`.
- `PatternStoreView::removePatternView()`-> `removeViewsForPattern()`

TODO:
- `DataFile.cpp` ;)
- wiki

For an easier view of the diff: `git diff -M40 --color-words='\w'`


include/MidiClip.h
```diff
@@ -42,7 +42,7 @@ class LMMS_EXPORT MidiClip : public Clip
 public:
 	enum MidiClipTypes
 	{
-		BeatClip,
```
M374LX, line 45: I would suggest keeping BeatClip, as a distinction between a beat clip and a melody clip in a pattern (previously beat+bassline) makes sense and a BeatClip is the default for a new instrument track added to a pattern. Also, a BeatClip, as the name suggests, contains beats and no melodies.

include/MidiClip.h
```diff
@@ -122,7 +122,7 @@ protected slots:
 
 
 private:
-	TimePos beatClipLength() const;
```
M374LX, line 125: Same as in https://github.com/LMMS/lmms/pull/6284/files#r785357595.

src/core/PatternClip.cpp
```diff
@@ -1,5 +1,5 @@
 /*
- * BBClip.cpp - implementation of class bbClip
+ * PatternClip.cpp - implementation of class patternClip
```
M374LX, line 2: It should be "implementation of class **P**atternClip", with an uppercase P.

include/MidiClip.h
allejok96: Good thinking, and I just realized that name conflict between `MidiClipTypes::PatternClip` and `PatternClip`. Reverted.

allejok96:
Is there a better name we could give to `BBTrackContainer` aka `PatternTrackContainer` ?

It's confusing since it can never contain a `PatternTrack`. And it's even worse how the GUI makes it seem like there's a `PatternTrackContainer` inside each `PatternTrack`.

If possible the name should indicate that there is only one `PatternTrackContainer` in a project, the same way there is only one `Song` (which is also a form of `TrackContainer`).

`PatternEditorTrackContainer` is way too long and boring :)

M374LX:
@allejok96 Why not ``PatternContainer``?

JohannesLorenz:
Or just `Pattern`. No idea if that makes sense.

allejok96:
`Pattern` sounds like it's a single clip while in reality it's a collection. `PatternContainer` sounds like `TrackContainer` which implies that it contains instances of `Pattern`, but there are no such thing, it's just an abstract concept.

This is how patterns are stored. We want the programmer to visualize this when they hear the name.
bild

It's is like a two-dimensional array, so perhaps `PatternMatrix`? Still not perfect because it sounds like each cell would be a pattern, when in reality a pattern is a column, but it gives you a clue of the data structure.

JohannesLorenz:
I find `PatternMatrix` sounds really misleading, because, as you say, this might make readers think that the red things are the Patterns. `ClipMatrixForBBTrack` would be better.

But I think, to make the inheritation more clear, it should still be named `...TrackContainer...`. So, `TrackContainerOfBBTrack` / `TrackContainerInsideBBTrack` / `TrackContainerForBBTrack`?

Btw, can you please draw that image as ASCII art into the header of the track container?

cyber-bridge:
Looks like a table to me, maybe `BBClipTable`?

allejok96:
We are removing the term `BB`. Some more suggestions:

Song Editor edits the `Song` and Pattern Editor edits the `PatternSet` ...?

Or `PatternTrack`s get their data from the  `PatternSourceContainer` ...?

allejok96:
Forgive my endless naming issue. I think I have made up my mind now. Let's hear what you say.

I loved `PatternSet` but it gets confusing like: `PatternSet::setCurrentPattern()`

So what about **`PatternCollection`**? It has the same vibe but little longer...
- There is just one collection in a project
- The collection contains all patterns
- It does not insinuate that pattern is a class as much as `PatternContainer` would imo
- "Pattern Editor deals with the Pattern Collection"

For reasons why this is a issue in the first place, see this terrible infographic:

img

#### Why not make inheritance more clear in the name?
Well, since we are thinking in terms of "patterns" and pattern isn't even an object, I think putting the focus on tracks makes it more confusing. Makes you wonder: where is the Pattern? With a nice fuzzy name it looks good in the code and then we put that ASCII art in there so when you look it up you'll get a better grip. Besides, `PatternEditorTrackContainerView` would be terrible :)

JohannesLorenz:
I, personally, agree to all you said in the last post. One minor question: If there would be a second PatternTrack in the Song (left side in your image), it would link to the same PatternCollection, correct?

M374LX:
Another possibility is ``PatternStore``.

allejok96:
> If there would be a second PatternTrack ... it would link to the same PatternCollection?
Yes, same PatternCollection, different index/pattern/column.

src/core/Track.cpp
```diff
@@ -122,10 +116,9 @@ Track * Track::create( TrackTypes tt, TrackContainer * tc )
 		default: break;
 	}
 
-	if( tc == Engine::getBBTrackContainer() && t )
+	if (tc == Engine::getPatternStore() && t)
 	{
-		t->createClipsForBB( Engine::getBBTrackContainer()->numOfBBs()
-									- 1 );
+		t->createClipsForPattern(Engine::getPatternStore()->numOfPatterns()	- 1);
```
Spekular, line 121:
```suggestion
		t->createClipsForPattern(Engine::getPatternStore()->numOfPatterns() - 1);
```

src/gui/editors/PatternEditor.cpp
```diff
@@ -316,21 +310,21 @@ void BBTrackContainerView::makeSteps( bool clone )
 	}
 }
 
-// Creates a clone of the current BB track with the same clip, but no Clips in the song editor
+// Creates a clone of the current pattern track with the same clip, but no Clips in the song editor
```
Spekular, line 313:
```suggestion
// Creates a clone of the current pattern track with the same content, but no clips in the song editor
```

Spekular:
Did a super rudimentary review (only cpp, h, and txt files; only side-by-side comparisons). Haven't reviewed BBTrackContainer changes yet since GitHub's diff failed to show them side to side, but didn't find anything obviously wrong elsewhere (aside from the TODO'd data file upgrade).

JohannesLorenz:
Thanks for the fixup. How should we advance? 2nd review? Merge? Extend?

allejok96:
At least someone should glance over the changes of `BBTrackContainer` with `git diff -M40` (since github doesn't detect the rename) before merging

JohannesLorenz:
> At least someone should glance over the changes of `BBTrackContainer` with `git diff -M40` (since github doesn't detect the rename) before merging

@allejok96 `git diff 7fe2152938db49c8b8a2d04a5ab1eccfd38c821a -M40 include/PatternStore.h` just shows the whole file as new. How do I use it correctly?

JohannesLorenz:
Another thing I'm just noticing: If "pattern" == "bb track", then "pattern tracks" means "bb track tracks", right? So "pattern track" should nowhere be used?

Spekular:
> Another thing I'm just noticing: If "pattern" == "bb track", then "pattern tracks" means "bb track tracks", right? So "pattern track" should nowhere be used?

I don't think pattern = bb track though. Aren't the contents of the BB editor a pattern, and then pattern clips are placed on pattern tracks in order to loop that pattern as part of the song?

allejok96:
> shows the whole file as new. How do I use it correctly?

`-M` sets the threshold for file renaming, so you'd have to specify both the old and new file name

    git diff bbRename -M40 -- include/PatternStore.h include/BBTrackContainer.h

> I don't think pattern = bb track though. Aren't the contents of the BB editor a pattern, and then pattern clips are placed on pattern tracks in order to loop that pattern as part of the song?

This is the correct understanding. @JohannesLorenz I may have created confusing with the drawings in this discussion, but I hope there is no misleading comments in the code?

JohannesLorenz:
Oh, right, there are patterns and pattern tracks...

```
* Do not confuse the Tracks in the PatternStore with PatternTracks.
* PatternTracks are used in the Song Editor to reference a pattern.
* They do not contain any MIDI data.
```

Thanks for your explanations. Will continue to review.

include/PatternClip.h
```diff
@@ -42,14 +42,14 @@ class BBClip : public Clip
 		return( "bbtco" );
```
JohannesLorenz, line 42: Can you add a "TODO" here, please?

src/tracks/PatternTrack.cpp
```diff
@@ -175,12 +172,12 @@ void BBTrack::loadTrackSpecificSettings( const QDomElement & _this )
 
 	if( _this.hasAttribute( "clonebbt" ) )
```
JohannesLorenz, line 173: Can you add a similar "TODO" comment like 2 lines below, please?

include/PatternStore.h
```diff
@@ -0,0 +1,106 @@
...
+ * | | Track 2   [Clip C]  [Clip D] | |
+ * | +------------------------------+ |
+ * +---------------------------------+
+ *
+ * Pattern Editor displays one pattern at a time, but they all belong to the same PatternStore.
+ * A "pattern" is not an object, it's just a word we use to describe a column of clips.
+ * In the illustration above, Clip A and Clip C would be "Pattern 1".
+ *
+ * Do not confuse the Tracks in the PatternStore with PatternTracks.
+ * PatternTracks are used in the Song Editor to reference a pattern.
```
JohannesLorenz, line 53: I still miss `PatternClip` in this diagram. Possible change:

```suggestion
 * PatternTracks are used in the Song Editor and contain PatternClips that each reference a pattern.
```

include/PatternClip.h
```diff
@@ -23,17 +23,17 @@
...
-#ifndef BB_CLIP_H
-#define BB_CLIP_H
+#ifndef PATTERN_CLIP_H
+#define PATTERN_CLIP_H
 
 #include "ClipView.h"
 
 
-class BBClip : public Clip
+class PatternClip : public Clip
```
JohannesLorenz, line 32: A diagram here would be cool, so that you know that this is inside the song editor, not in the PatternStore.

JohannesLorenz (commented):
Just a few minor/easy comments. Aside from that, very clear PR :+1:

src/tracks/PatternTrack.cpp
```diff
@@ -118,55 +117,53 @@ bool BBTrack::play( const TimePos & _start, const fpp_t _frames,
...
 
 
 
-Clip* BBTrack::createClip(const TimePos & pos)
+Clip* PatternTrack::createClip(const TimePos & pos)
 {
-	BBClip* bbclip = new BBClip(this);
-	bbclip->movePosition(pos);
-	return bbclip;
+	PatternClip* pattern = new PatternClip(this);
```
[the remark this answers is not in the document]
allejok96: Presumptuous me renamed it to `pc` instead

allejok96:
- Added brief descriptions to `PatternTrack` and `PatternClip`.
- Renamed `PatternTrack::index` to `patternIndex` to avoid confusing code like `track->index()`
- Renamed `PatternClip::patternTrackIndex` to `patternIndex` (it's just a call to the one above)
- Renamed "Beat/Bassline" to "Pattern" in the template files. In a future PR we'll add a regex to do the same for all other old files when they are loaded.

JohannesLorenz:
@allejok96 You describe `PatternTrack` as a "Track type used in the Song (Editor) to reference a pattern in the `PatternStore`". However, `PatternStore`'s comment says that a "pattern" is a column. Do you really reference a column with a `PatternTrack`?

allejok96:
@JohannesLorenz yes. The "reference" is returned by (renamed) `PatternTrack::patternIndex`

Afaik it comes from a map of `PatternTrack*` and `int`, with is essentially a time position (bar) in the PatternStore.

JohannesLorenz (approved):
Rework is OK.

The only thing that might (optionally) be worth adding in PatternStore.h is what you mean with columns. I didn't understand that you can select the column by using that combo box from the editor.

Next step is testing before we merge.

JohannesLorenz:
Thinking about what needs to be tested...

* The whole pattern editor functionality (should still work as before)
* Load/save and DnD with patterns (PR should have no impact)

Is that everything?

@allejok96 Also, in case you don't have merge access rights, feel free to formulate a commit message that I can put into the squash-merge.

allejok96:
That's about it. DnD untouched. Test usability of pattern-things. Check that no icons are missing or that anything looks weird (CSS).

```
Rename Beat/Bassline to Pattern

- BB* -> Pattern*
- PatternTrackContainer -> PatternStore
- BBTrack::index -> PatternTrack::patternIndex

Does not touch save files
```

allejok96:
Oh man I just realized something

| Editor | Central widget |
| --- | --- |
| SongEditorWindow | SongEditor |
| AutomationEditorWindow | AutomationEditor |
| PianoRollWindow | PianoRoll | 
| PatternEditor | PatternStoreView |

They should be PatternEditorWindow and PatternEditor

JohannesLorenz:
> They should be PatternEditorWindow and PatternEditor

Agreeing. Shouldn't be too difficult to just rename two classes.

@Spekular , any objections?

allejok96:
`PatternEditor.cpp` will have the `PatternEditorWindow` definition at the top of the file, followed by the `PatternEditor` definition. Should this order be changed to match the one found in `SongEditor.cpp`, `PianoRoll.cpp` etc? git blame me for everything :cry: I guess the diff is all over the place anyhow...

JohannesLorenz:
> Should this order be changed to match the one found in `SongEditor.cpp`, `PianoRoll.cpp` etc?

I'd say yes (and in `PatternEditor.h`, too). These files are so small that the diff is probably still comprehensive. But if you want to make it clear: Complete the whole renaming stuff without swapping the definitions, give it to review again, then squash it all into 1 commit, and then add a second commit for swapping. If we do not squash-merge the result, then users can see this as 2 commits. (This could even be done in 2 PRs aswell)

allejok96:
I think it's ready, finally.

```
Rename Beat/Bassline to Pattern

- BB* -> Pattern*
- BBTrackContainer -> PatternStore
- BBTrackContainerView -> PatternEditor
- BBEditor -> PatternEditorWindow

Does not touch save files
```

JohannesLorenz (approved):
LGTM

JohannesLorenz:
Branch is opened for testing. The following should be tested:

* The whole pattern editor functionality (should still work as before)
* Everything else with patterns
* Check that no icons are missing or that anything looks weird (CSS)
* Load/save and DnD with patterns (PR did not touch those)

MessyBookshelf (commented):
Tested a few time and it seems to work great on my windows64 build, patterns save and load, extend, everything functions and looks good as far as I can tell 🎉

PR #6286, opened 2022-01-16
Show error message when loading an invalid sample file

### LINKED ISSUE #6283: No error with invalid sample file (sample track)
### Bug Summary
When trying to add an invalid file to a sample track, no error dialog is displayed and the reference to the invalid file is saved to the project.

#### Steps to reproduce
1. Create an empty file with the ".wav" extension.
2. Start LMMS.
3. Add a sample track.
4. Add the empty file created on step 1 to the created sample track.
5. Save the project in uncompressed form (.mmp).

#### Expected behavior
An error dialog and no reference to the empty sample file on the saved project.

#### Actual behavior
No error dialog is displayed and, when the uncompressed project is opened on a text editor (or ``less``), the reference to the invalid sample file is there.

#### Affected LMMS versions
1.3.0-alpha.1.162+g8b2a30914 (Linux/x86_64, Qt 5.15.2, GCC 10.2.1 20201203).

M374LX:
Fixes #6283.

~It also fixes a bug that causes the sample clip view to disappear when trying to load an invalid file, although the clip still exists and it gets saved into the project. What actually happens is that the length of the clip is set to one bar when it is just created but it is set to zero when the file is invalid~ (undone due to the side effect of the path getting lost if the file is valid but cannot be loaded).

There is still an undesirable behavior: when trying to replace a valid sample file with an invalid one, the previous (valid) file disappears. It should be kept intact in this case. It seems quite complex to fix and is better left for another PR (issue: #6287).

### DIFF
--- a/src/core/SampleBuffer.cpp
+++ b/src/core/SampleBuffer.cpp
@@ -408,7 +408,7 @@ void SampleBuffer::update(bool keepSettings)
 
 		if (getGUI() != nullptr)
 		{
-			QMessageBox::information(nullptr, title, message,	QMessageBox::Ok);
+			QMessageBox::information(nullptr, title, message, QMessageBox::Ok);
 		}
 		else
 		{
--- a/src/core/SampleBuffer.cpp
+++ b/src/core/SampleBuffer.cpp
@@ -279,7 +279,11 @@ void SampleBuffer::update(bool keepSettings)
 		m_frames = 0;
 
 		const QFileInfo fileInfo(file);
-		if (fileInfo.size() > fileSizeMax * 1024 * 1024)
+		if (!fileInfo.isReadable())
+		{
+			fileLoadError = ReadPermissionDenied;
+		}
+		else if (fileInfo.size() > fileSizeMax * 1024 * 1024)
 		{
 			fileLoadError = TooLarge;
 		}
@@ -287,16 +291,11 @@ void SampleBuffer::update(bool keepSettings)
 		{
 			// Use QFile to handle unicode file names on Windows
 			QFile f(file);
-			QFileInfo fileInfo(f);
 			SNDFILE * sndFile = NULL;
 			SF_INFO sfInfo;
 			sfInfo.format = 0;
 
-			if (!fileInfo.isReadable())
-			{
-				fileLoadError = ReadPermissionDenied;
-			}
-			else if (f.open(QIODevice::ReadOnly) && (sndFile = sf_open_fd(f.handle(), SFM_READ, &sfInfo, false)))
+			if (f.open(QIODevice::ReadOnly) && (sndFile = sf_open_fd(f.handle(), SFM_READ, &sfInfo, false)))
 			{
 				f_cnt_t frames = sfInfo.frames;
 				int rate = sfInfo.samplerate;
--- a/src/core/SampleBuffer.cpp
+++ b/src/core/SampleBuffer.cpp
@@ -305,7 +305,7 @@ void SampleBuffer::update(bool keepSettings)
 				}
 				sf_close(sndFile);
 			}
-			else if (sndFile == NULL)
+			else if (sndFile == nullptr)
 			{
 				fileLoadError = Invalid;
 			}
--- a/src/core/SampleBuffer.cpp
+++ b/src/core/SampleBuffer.cpp
@@ -379,6 +379,8 @@ void SampleBuffer::update(bool keepSettings)
 		QString title = tr("Fail to open file");
 		QString message;
 
+		m_audioFile = "";
+
 		switch (fileLoadError)
 		{
 			case None:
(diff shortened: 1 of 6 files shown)


src/core/SampleBuffer.cpp
```diff
@@ -287,14 +294,18 @@ void SampleBuffer::update(bool keepSettings)
 				int rate = sfInfo.samplerate;
 				if (frames / rate > sampleLengthMax * 60)
 				{
-					fileLoadError = true;
+					fileLoadError = TooLarge;
 				}
 				sf_close(sndFile);
 			}
+			else if (sndFile == NULL)
```
M374LX, line 301: Knowing that libsndfile is written in C, rather than C++, does anyone know if using ``nullptr`` here is safe?

src/core/SampleBuffer.cpp
```diff
@@ -362,16 +380,37 @@ void SampleBuffer::update(bool keepSettings)
...
+				break;
+
+			case TooLarge:
+				message = tr("Audio files are limited to %1 MB "
+					"in size and %2 minutes of playing time"
+					).arg(fileSizeMax).arg(sampleLengthMax);
+				break;
+
+			case Invalid:
+				message = tr("Invalid audio file");
```
M374LX, line 407: New locale message for translators!

src/core/SampleBuffer.cpp
```diff
@@ -362,16 +380,37 @@ void SampleBuffer::update(bool keepSettings)
...
+		m_audioFile = "";
+
+		switch (fileLoadError)
+		{
+			case None:
+				// present just to avoid a compiler warning
+				break;
+
+			case ReadPermissionDenied:
+				message = tr("Read permission denied");
```
M374LX, line 397: New locale message for translators!

superpaik:
The build generated by the boot crashes when loading a sample either in a sample track or as sample (tested on W10-64b)

M374LX:
@superpaik I have not experienced any crash on Linux. Have you tried the same sample on other PR builds?

Has any other user experienced a crash?

superpaik:
Yes, they work Ok on other PR. Can be due to the LmmsBot? It's weird that there is only the Win-MSVC version available.

src/core/SampleBuffer.cpp
PhysSong: Looks like it does:

https://stackoverflow.com/questions/36484473/is-c-null-equal-to-c11-nullptr
https://en.cppreference.com/w/cpp/types/NULL

src/core/SampleBuffer.cpp
```diff
@@ -362,16 +379,37 @@ void SampleBuffer::update(bool keepSettings)
...
-	if (fileLoadError)
+	if (fileLoadError != None)
 	{
 		QString title = tr("Fail to open file");
-		QString message = tr("Audio files are limited to %1 MB "
-				"in size and %2 minutes of playing time"
-				).arg(fileSizeMax).arg(sampleLengthMax);
+		QString message;
+
+		m_audioFile = "";
```
PhysSong, line 387: Could you explain what do you want to change with this additional line?

M374LX: The path to the invalid file gets saved into the project if it is not cleared.

PhysSong: While this fixes the mentioned issue, it has an undesirable side effect. If the path to a valid file was saved, but the file failed to load for some reason, the path information will be lost.

M374LX: Removing it for now. Remaining problems with file loading from ``SampleBuffer`` can be discussed in #6287.

PhysSong (approved):

src/core/SampleBuffer.cpp
```diff
@@ -247,7 +247,15 @@ void SampleBuffer::update(bool keepSettings)
 	const int fileSizeMax = 300; // MB
 	const int sampleLengthMax = 90; // Minutes
 
-	bool fileLoadError = false;
+	enum FileLoadError
```
DomClark, line 250: This would be better as an `enum class`. (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Renum-class)

src/core/SampleBuffer.cpp
```diff
@@ -271,31 +279,40 @@ void SampleBuffer::update(bool keepSettings)
...
 				f_cnt_t frames = sfInfo.frames;
 				int rate = sfInfo.samplerate;
 				if (frames / rate > sampleLengthMax * 60)
 				{
-					fileLoadError = true;
+					fileLoadError = TooLarge;
 				}
 				sf_close(sndFile);
 			}
+			else if (sndFile == nullptr)
```
DomClark, line 308: This condition will always evaluate as true - the only way `sndFile` can be assigned something other than `nullptr` is in the first condition, but if that happens, then that condition will be true and the first block will run instead.

Veratil:
@DomClark I assume with the changes requested fixed you'll approve now.
@M374LX can you rebase on master and we can work on getting this merged finally. :)

DomClark (approved):

PR #6293, opened 2022-01-28
Update PianoRoll.h

musikBear:
This addresses part of  #2380 
In pianoroll.h Added one prototype 
void strumNotesDn();
This done so new method can be added to pianoroll.cpp in next PR.
In pianoroll.cpp
Update PianoRoll.cpp
I have added one method
void PianoRoll::strumNotesDn()
This method will strum a stack of notes with the size of user-chosen value of Q
I have added one button for this action next to the Quantize button.
This button has tooltip, but it does not have an icon!
I hope our artists will create one in the style of current default theme. (if my patch is accepted.

### DIFF
--- a/include/PianoRoll.h
+++ b/include/PianoRoll.h
@@ -209,6 +209,7 @@ protected slots:
 	void noteLengthChanged();
 	void keyChanged();
 	void quantizeNotes(QuantizeActions mode = QuantizeBoth);
+	void strumNotesDn();
 
 	void updateSemiToneMarkerMenu();
 
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -4590,56 +4590,6 @@ int PianoRoll::quantization() const
 	return DefaultTicksPerBar / Quantizations[m_quantizeModel.value() - 1];
 }
 
-int returnStrum(int st)//input > read value from comboBox Q. Output size of strum
-{
-int strumSize = 0;
-	switch (st)
-	{   
-	//------sys-notes------------
-	case 1:
-        strumSize = 196;//Wholenote
-	break;
-        case 2:
-        strumSize = 98; //halfnote
-	break;
-        case 3:
-        strumSize = 48; //quoter
-	break;
-        case 4:
-        strumSize = 24; // 1/8
-	break;
-        case 5:
-        strumSize = 12; //1/16
-	break;
-        case 6:
-        strumSize = 6; //1/32
-	break;
-        case 7:
-        strumSize = 3; //1/64
-	break;
-	//---------triplets------------
-        case 8:
-        strumSize = 64; //1/2
-	break;
-        case 9:
-        strumSize = 32; //1/4
-	break;
-        case 10:
-        strumSize = 16; //1/8
-	break;
-        case 11:
-        strumSize = 8; //1/16
-	break;
-        case 12:
-        strumSize = 4; //1/32
-	break;
-        default:
-        strumSize = 1; //1/192 //smallest tick-value Availble if option 0 #notelock# is chosen
-	break;
-    }
-    return strumSize;
-}
-
 void PianoRoll::strumNotesDn() //down-top
 //Method for chord-strumming.
 //The strum-distance is the chosen Q-value
@@ -4656,7 +4606,7 @@ void PianoRoll::strumNotesDn() //down-top
   index = m_quantizeModel.value();	//uses UserSet Q-Value in combobox
   									//This value is position, so it need to be qualified
 									//to real piano/roll values
-  int strumSz = returnStrum(index); //size of note movement 
+  int strumSz = quantization(); //size of note movement	determined by users chosen Q-value in dropdown
   int strum = strumSz;//keep orr. value of selected strumming
   bool firstDone = false;//first note should not be moved
   for( Note* n : notes ) //for each selected
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -4594,13 +4594,13 @@ void PianoRoll::strumNotesDn() //down-top
 //Method for chord-strumming.
 //The strum-distance is the chosen Q-value
 {
-  if( ! hasValidPattern() )
+  if(! hasValidMidiClip())
   {
      return;
   }
   int notesInChord = 0; //initializations
   int index = 0;
-  m_pattern->addJournalCheckPoint();
+  m_midiClip->addJournalCheckPoint();
   NoteVector notes = getSelectedNotes();//NoteVector of class QVector
   notesInChord = notes.count();//how many notes are in this chord
   index = m_quantizeModel.value();	//uses UserSet Q-Value in combobox
@@ -4609,19 +4609,19 @@ void PianoRoll::strumNotesDn() //down-top
   int strumSz = quantization(); //size of note movement	determined by users chosen Q-value in dropdown
   int strum = strumSz;//keep orr. value of selected strumming
   bool firstDone = false;//first note should not be moved
-  for( Note* n : notes ) //for each selected
+  for(Note* n : notes) //for each selected
   {
     n = notes.at(notesInChord-1);//Use QVector-method .at
     notesInChord-=1;
-    if( firstDone ){
+    if(firstDone){
         n->setPos( n->pos() + strumSz );
         strumSz += strum; }
     firstDone = true;
         if (notesInChord<=1) continue;      
   }    
-m_pattern->rearrangeAllNotes();
-m_pattern->updateLength();
-m_pattern->dataChanged();
+m_midiClip->rearrangeAllNotes();
+m_midiClip->updateLength();
+m_midiClip->dataChanged();
 update();//Project is changed. Update
 gui->songEditor()->update();
 Engine::getSong()->setModified();
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -4590,7 +4590,6 @@ int PianoRoll::quantization() const
 	return DefaultTicksPerBar / Quantizations[m_quantizeModel.value() - 1];
 }
 
-øøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøø
 int returnStrum(int st)//input > read value from comboBox Q. Output size of strum
 {
 int strumSize = 0;
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -4590,6 +4590,94 @@ int PianoRoll::quantization() const
 	return DefaultTicksPerBar / Quantizations[m_quantizeModel.value() - 1];
 }
 
+øøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøøø
+int returnStrum(int st)//input > read value from comboBox Q. Output size of strum
+{
+int strumSize = 0;
+	switch (st)
+	{   
+	//------sys-notes------------
+	case 1:
+        strumSize = 196;//Wholenote
+	break;
+        case 2:
+        strumSize = 98; //halfnote
+	break;
+        case 3:
+        strumSize = 48; //quoter
+	break;
+        case 4:
+        strumSize = 24; // 1/8
+	break;
+        case 5:
+        strumSize = 12; //1/16
+	break;
+        case 6:
+        strumSize = 6; //1/32
+	break;
+        case 7:
+        strumSize = 3; //1/64
+	break;
+	//---------triplets------------
+        case 8:
+        strumSize = 64; //1/2
+	break;
+        case 9:
+        strumSize = 32; //1/4
+	break;
+        case 10:
+        strumSize = 16; //1/8
+	break;
+        case 11:
+        strumSize = 8; //1/16
+	break;
+        case 12:
+        strumSize = 4; //1/32
+	break;
+        default:
+        strumSize = 1; //1/192 //smallest tick-value Availble if option 0 #notelock# is chosen
+	break;
+    }
+    return strumSize;
+}
+
+void PianoRoll::strumNotesDn() //down-top
+//Method for chord-strumming.
+//The strum-distance is the chosen Q-value
+{
+  if( ! hasValidPattern() )
+  {
+     return;
+  }
+  int notesInChord = 0; //initializations
+  int index = 0;
+  m_pattern->addJournalCheckPoint();
+  NoteVector notes = getSelectedNotes();//NoteVector of class QVector
+  notesInChord = notes.count();//how many notes are in this chord
+  index = m_quantizeModel.value();	//uses UserSet Q-Value in combobox
+  									//This value is position, so it need to be qualified
+									//to real piano/roll values
+  int strumSz = returnStrum(index); //size of note movement 
+  int strum = strumSz;//keep orr. value of selected strumming
+  bool firstDone = false;//first note should not be moved
+  for( Note* n : notes ) //for each selected
+  {
+    n = notes.at(notesInChord-1);//Use QVector-method .at
+    notesInChord-=1;
+    if( firstDone ){
+        n->setPos( n->pos() + strumSz );
+        strumSz += strum; }
+    firstDone = true;
+        if (notesInChord<=1) continue;      
+  }    
+m_pattern->rearrangeAllNotes();
+m_pattern->updateLength();
+m_pattern->dataChanged();
+update();//Project is changed. Update
+gui->songEditor()->update();
+Engine::getSong()->setModified();
+notes.clear();//delete content -if not repeated strums on same selection fails
+}
 
 void PianoRoll::quantizeNotes(QuantizeActions mode)
 {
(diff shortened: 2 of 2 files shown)


src/gui/editors/PianoRoll.cpp
```diff
@@ -4590,6 +4590,93 @@ int PianoRoll::quantization() const
 	return DefaultTicksPerBar / Quantizations[m_quantizeModel.value() - 1];
 }
 
+int returnStrum(int st)//input > read value from comboBox Q. Output size of strum
```
Spekular, line 4593: From what I can tell this just reimplements `PianoRoll::quantization()`

musikBear: If the method `PianoRoll::quantization()` can be used to add _each_ of the values in the Q-dropdown to a position-value, then we should ofcause use that method in my strumming-method. I cant see how  `PianoRoll::quantization()` would do that, but i will look at it again.
One question :
What is the procedure for making an edit, commit & PR that new edit.
Do I
* Edit the _current PR_ (I am not sure how to do that)
Or..
* Do I _close_ the current PR, and make a _new_ one with the _new_ and edited code?

Spekular: To update your PR, all you have to do is make changes to the source branch. So in this case, make any changes you want on the musikBear/master branch and they'll show up here.

musikBear: Woooo _That_ is actually clever!
A 👍 to gitHub for implementing that!
and i looked at `PianoRoll::quantization()` and ofcause you are correct. That method does almost the same as my switch. Its only `default case` where user choice is _"note-length"_ that isent directly covered. I will look at that.

musikBear:
I have changed strumming so it uses the existing method `quantization();` instead of the Switch in `int returnStrum(int st)`
I have also removed the method `returnStrum`.
I ask for review for those changes, and for merging with LMMS/lmms (Master), so i can commit my next new method in a clean way.

include/PianoRoll.h
```diff
@@ -209,6 +209,7 @@ protected slots:
 	void noteLengthChanged();
 	void keyChanged();
 	void quantizeNotes(QuantizeActions mode = QuantizeBoth);
+	void strumNotesDn();
```
allejok96, line 212: At the moment this function strums like this /  ... I would call that up. And down like this \

The best thing would be if the function was defined like:

    void strumNotes(bool down);

or something like that. It would make it possible to strum up and down using the same function, like:

    strumNotes(true);  // down
    strumNotes(false);  // up

Instead of making another function for `strumNotesUp()`

src/gui/editors/PianoRoll.cpp
```diff
@@ -4590,6 +4590,43 @@ int PianoRoll::quantization() const
 	return DefaultTicksPerBar / Quantizations[m_quantizeModel.value() - 1];
 }
 
+void PianoRoll::strumNotesDn() //down-top
+//Method for chord-strumming.
+//The strum-distance is the chosen Q-value
+{
+  if( ! hasValidPattern() )
```
allejok96, line 4597: `hasValidPattern` has been renamed to `hasValidMidiClip`

src/gui/editors/PianoRoll.cpp
```diff
@@ -4590,6 +4590,43 @@ int PianoRoll::quantization() const
...
+  {
+     return;
+  }
+  int notesInChord = 0; //initializations
+  int index = 0;
+  m_pattern->addJournalCheckPoint();
```
allejok96, line 4603: `m_pattern` has been renamed to `m_midiClip`

src/gui/editors/PianoRoll.cpp
```diff
@@ -4745,9 +4782,10 @@ PianoRollWindow::PianoRollWindow() :
 	eraseAction->setShortcut( Qt::SHIFT | Qt::Key_E );
 	selectAction->setShortcut( Qt::SHIFT | Qt::Key_S );
 	pitchBendAction->setShortcut( Qt::SHIFT | Qt::Key_T );
-
-	connect( editModeGroup, SIGNAL( triggered( int ) ), m_editor, SLOT( setEditMode( int ) ) );
-
+	connect( editModeGroup, SIGNAL( triggered( int ) ), m_editor, SLOT( setEditMode( int ) ) );
```
allejok96, line 4785: When you add or change code, try to use the [new code style](https://github.com/LMMS/lmms/wiki/Coding-Conventions) ... that means no spaces around ( )

src/gui/editors/PianoRoll.cpp
musikBear: Ok i will fix that

src/gui/editors/PianoRoll.cpp
musikBear: Ok i will fix that too

src/gui/editors/PianoRoll.cpp
musikBear: Oh.. a new style. Ok.
I will go over all functions for that too.
Does that mean that all code need to be overhauled as well? 
Where has this decision been taken. I try to keep track of things like that, because it is annoying work to make code-style-changes, but i havent seen this new conformity-style change in any posts. We (properly) should have a special page in dev-wiki for things like that, but i will fix it.

allejok96: The white space rule has existed since 2019 when the Coding-Conventions page was created, so that's probably why you never heard any discussion about changing it.

In general, only change the coding style on the lines you edit. That makes it easier to see which lines has been changed when people are reviewing your PR, and it creates less conflicts.

musikBear: > In general, only change the coding style on the lines you edit. That makes it easier to see which lines has been changed when people are reviewing your PR, and it creates less conflicts.

👍Got it! Will do.

include/PianoRoll.h
musikBear: Up/Down
I saw it as from where the change _came_. In strumNotesDn() we start low and end higher.
In strumNotesUp() we start UP and end lower
That was the logic, I will add a headsup for that.

src/gui/editors/PianoRoll.cpp
musikBear: Fixed

src/gui/editors/PianoRoll.cpp
musikBear: Fixed

musikBear:
Has:
* Changed code according to Coding-Conventions.
Changed
* hasValidPattern to hasValidMidiClip
* m_pattern to m_midiClip
Added comment for Dn-logic
___
Not changed
2 method-structure. Reason explained above
Ask for merge.

Spekular:
> Not changed 2 method-structure. Reason explained above

> I saw it as from where the change came. In strumNotesDn() we start low and end higher.

Up and down are directions though, not locations (that would be top or bottom). Thus, per definition, an upwards strum starts at the bottom, and a downwards strum starts at the top.

Regardless, the function should _at least_ accept a boolean to set the direction. IMHO, it would be even better if it accepted a signed int representing the offset (in ticks) of the top note from the bottom note. Passing strumNotes(48) would result in an upwards strum where the top note is moved right by a quarter note and the bottom note unchanged. Passing (-24) would result in a downwards strum, with the bottom note shifting right by an eight note.

(Whether a positive value gives an upwards or downwards strum is technically arbitrary, but as the base note of a chord is often the root and thus "most important" it may make sense to use as a reference. Additionally, the way a guitar is usually strummed results in the lowest notes playing first, so it makes some sense that a positive input to the strum function would mimic this.)

allejok96:
Things that must be done before this can be merged:
- [ ]  Try compiling your code _before_ you upload it - there is a error on line 4626!
- [ ] `strumNotes` should be able to strum up ↗ and down ↘ as Spekular explained
- [ ] Strumming multiple chords or already strummed chords should either work flawlessly or refuse to do anything
- [ ] All buttons must have icons
- [ ] Some kind of menu (where more humanization can be added later)
   bild

musikBear:
with these demands i am not able to make this.
Especially:
> Strumming multiple chords or already strummed chords should either work flawlessly or refuse to do anything

Atm the method does this:
image
The yellow line indicates that there will be `strum-size` between each chord.
The way i read the critique is that the result is expected to be this
image
-Eg the space _between_ chord are preserved.
But that is impossible in this case:
image
How should the number of notes in the chord be known?
That number is needed to set the distance between chords after strumming.
Not possible.
Position cant be used, because n notes in a chord can have any position, but still be one chord
image
-And, Number of notes in a chord is not known.
Hence -not possible.

allejok96:
> with these demands i am not able to make this.

If you feel this is beyond your current programming knowledge, it's ok to quit. I just want to leave you with a little food for thought.

Expected behavior:

bild

Code mockup:

```cpp
strumCount = 0
previousStart = -1  // start with an invalid position
for (note: notes) {
    if (note->pos() != previousStart) {
         // begin new chord
        strumCount = 0
    } else {
        strumCount += 1
        // move note depending on how many notes was before it (in the same chord)
        newStart = note->pos() + (quantization() * strumCount)
        note->setPos(newStart)
    }
    previousStart = note->pos()
}

musikBear:
image
NO they are NOT!
I already explained that!
This is _still_ a chord:
image
And this is a chord
image

So how would you be able to know where one chord changes to another?
There are no indication -Right? How would you make that distinction.

Veratil:
Your example of notes not being on the same tick still being a chord is not the point being made here. The point is notes on the same tick will be detected as a chord and shift appropriately. At this point we wouldn't be able to tell that your example is a chord, and detecting that would take an absolute mess of code to get working, and probably wouldn't ever be able to work perfectly.

Alternatively, since your example is shown as a selection of notes, the strum method could also work on only the selected notes and treat them as a chord. This shifts accidental note movement to the user if they selected notes not meant to be strummed.

And to mention here, we've talked about implementing different note types (to differentiate our BB notes and normal notes). With that we could implement a chord type which would make it much easier to tell that your example is a chord. Strumming up would start at the top-most note in the chord "note", and strumming down would start at the bottom-most.

midi-pascal:
@allejok96 I disagree with your conclusion. In my own musical software, I detect if notes make a chord if notes overlap at a given time. If there is no notes overlap, the strumming function is disabled.
If a chord is detected, the notes are put in a list that is then sorted by note - up or down depending on the strum direction selected by the user - and all the notes that follow the first one are moved left or right by the chosen quantization based on the first note of the list.
Then the "chop" function truncate the notes based on the longest note of the list.
This is somewhat simple to achieve and works as expected.
However, my implementation of strumming is based on selected notes, not on a full track.
PS: Nothing is impossible to a computer :-)

Veratil:
> PS: Nothing is impossible to a computer :-)

Well, we tend to say anything that will take more time to compute than the age of the universe to be impossible. 😂

Anyway, detecting overlapping notes isn't a bad idea. What happens if you have notes from a previous strum that you want to continue to decay and those partially overlap with the next strum?

musikBear:
How about this:
Rule:
Strumming is a semi-automate event. It is a compromise between the user _and_ the code. 
It is the users obligation to define what he consider to be _one_ chord 
That user-defined chord -or rather _stack-of-selected-notes_, and **only that** will be strummed.
If the user want to strum **another** _stack-of-selected-notes_ HE must define that other _stack-of-selected-notes_ AFTER the first _stack-of-selected-notes_ has been strummed.
Strumming is therefore a one-stack-of-selected-notes event. The method does NOT cater for blocks of selected-notes
This is possible:
image

This however is NOT:
image

Would _that_ kind of limitation rule be acceptable?
Then no problem -It works like that already..

allejok96:
I did a test (by hand) on a melody that contains both a melodic line and "perfectly lined up" chords. The result is very interesting.

https://user-images.githubusercontent.com/7693838/156887280-4c6c0cab-a195-4f71-a78c-d95a3e47abb7.mp4

@musikBear your approach has the side effect of slowing down the whole piece of music, but if you speed it up the result is almost identical to mine.

Question is which is more useful:
- Button that strums a single, selected, already somewhat humanized chord
- Button that strums all selected, perfectly lined up chords

Maybe @midi-pascal's solution can combine these two somehow... I didn't fully understand how it would handle a melody like the one in the video.

musikBear:
> 

Yes, that is how my method would strum a BLOCK of selected chordS, but what are your feelings on the approach i made here:
https://github.com/LMMS/lmms/pull/6293#issuecomment-1059431096
Because that implementation would make any type of strumming possible as long as it is ONE chord
It would allow all of the examples in your fine video, because it would be completely up to the user to define what he want to regard as his chord. So nomatter if notes are close, spaced, overlapping, even same notes on one key, they would still be strummed by a user-controlled distance.

But is the rule:

 _Anything selected before pressing the strum-button, will be treated as a chord, and strummed_

_That rule_ -Is that _acceptable_? 
Then it is easy to implement.
What do you think?
@allejok96 @Spekular @Veratil  @midi-pascal 
?

midi-pascal:
Personally I don't see any valid use case in treating "Anything selected" as a chord.
In my software (MIDI only) the user may select a single note (or more than one or a full track).
I solved this by considering selected notes as a chord when they overlap at some time in the selected notes duration. If not, the strum function is disabled.
How a single note can be treated as a chord? How to strum it?
How non-overlapping notes can be considered (musically) as a chord? How to strum them?
There are many strumming edge cases.
Among them: what to do if a strummed note overlap the same note following the strummed one? Chop? Cancel?
As the strumming may look pretty easy to implement, it is not as simple as it seems.
I do not know if you intend to implement a "Chop the chord at end" function but if so you will face other edge cases like if you chop based on the longest note you will lengthen the chord or loose notes because a note end comes before the note start if the quantization is longer than the note length.
Just to say all this looks as a very interesting challenge I am facing too :joy:

midi-pascal:
> @musikBear your approach has the side effect of slowing down the whole piece of music, but if you speed it up the result is almost identical to mine.

This looks like if moving notes moves also all the following notes. IMHO this is not the right way to do it but I may be wrong :joy:

midi-pascal:
> Would _that_ kind of limitation rule be acceptable?

What happens if the first chord is strummed and some of its notes overlap the second chord?
You the will not be able to strum it.
One of these edge cases I think of...

allejok96:
@midi-pascal would your implementation treat a song where all notes overlap each other like one big chord?

midi-pascal:
@allejok96 As long as all the selected notes overlap somewhere in time, they are considered a chord and can be strummed. If there is a silent part between them, this is not a chord (from my point of view) and the strum is disabled.
This is how I implemented the strumming but this does not mean I got the right answers.
As I said previously strumming chords is not as simple as it seems: many edge cases, many pitfalls.
I am simply trying to add some valid points to this discussion, not pushing my views on this topic into my beloved LMMS.
I am facing these same challenges in my own musical MIDI software.

musikBear:
> How a single note can be treated as a chord? How to strum it?

No that i have explained badly, if you got that impression.
What i meant was:
image
Strummed they look like this
image
> How non-overlapping notes can be considered (musically) as a chord? How to strum them?

It may not be useable, but possible
Here is an example
image
Strummed:
image
> what to do if a strummed note overlap the same note following the strummed one? Chop? Cancel?

Same heuristic. If selected it will be strummed. If not it remains untouched:
image
Strummed
image
> As the strumming may look pretty easy to implement, it is not as simple as it seems.

I know, thats why i made that rule. The rule _makes_ it simple.
>I do not know if you intend to implement a "Chop the chord at end" function 

No.
>This looks like if moving notes moves also all the following notes.

No, only notes that was marked is moving
> What happens if the first chord is strummed and some of its notes overlap the second chord?
You will not be able to strum the second one.

Strumming the second one would be a second task, where only the notes that the user wants in second chord is selected by the user, or unwanted notes are de-selected by the user. One chord-strumming is completely unrelated to any previous or anyone done after-
The are autonomous event.

midi-pascal:
@musikBear
First point: I do not see what this use case may be used for but as usual I may be wrong.
Second point: This does not refer to a chord at all. Strumming is destined to chords. I think you may be drifting a lot the notion of "Chord" there :joy:
Third point: Perhaps I was not clear enough. What happens if strumming a chord make some some notes collapse other non-selected same notes? i.e: If after strumming an E4 note collapses a following E4 that was not selected.
Fourth point: In my point of view your "Rule" has many holes considering the previous points. Only my opinion though. I would not apply it to my software to say.
Fifth point: You are better staying away from the "Chop" function. It is kind of nightmare I am currently fighting with.
Sixth: Ok, I did not know. It was simply a speculation based on the fact that @alleJok96 says that the song was slowed down.
Seventh point: This means you have to select the notes individually (not as a group)? This is not as my software works because the notes are selected using a zone delimited by a rubberband  so this does not apply in your case.
LMMS and me are just facing the same challenge at trying to implement a good strumming in our respective software so no pun intended there :joy:

allejok96:
@musikBear let me be blunt:

This is not strumming.


This is pointless.


We are discussing ways that strumming *could* or *should* work. If I were to ask you before you started coding: "what would a strum function do with these notes?" your answer would probably not be like the pictures. The two examples above are not logical, they are just a side-effect from the code.

> Anything selected before pressing the strum-button, will be treated as a chord, and strummed. Is that acceptable?

No, at least not to @midi-pascal and me. We want a strum function that can strum many chords at once. We don't want to sit and select chord by chord and strum it.

> I know, thats why i made that rule. The rule makes it simple.

This is the problem. The rule should be base on what is the most logical and useful thing for a human, not what is the easiest solution to the computer. As we said before, we can never make it perfect, but at least we can try to get it as good as possible.

musikBear:
> We want a strum function that can strum many chords at once

And you can design that?
I cant.

PR #6295, opened 2022-02-01
Only show Piano Roll when editable

### LINKED ISSUE #6234: Don't allow opening piano roll with invalid pattern
User: I opened the piano, but I cannot write notes
Solution: tell user he is doing it wrong
Problem: user is clearly illiterate, and this is bad design*

Developer: piano-roll crashes when pattern is `nullptr`
Solution: add `hasValidPattern()`
Problem: now there are safety checks everywhere

What if we simply prevented piano roll window from opening when there is no valid pattern?

### Solution

1. Disable/gray out piano roll button when there is no valid pattern.

The way the main window toolbar and view menu are set up currently makes things harder for us. It would be easier if we could use a common `QAction` for toggling piano roll that simply could be disabled.

2. Close the piano-roll window when pattern is deleted.

`MainWindow::toggleWindow` can be used to force open windows, but not to close them. This whole implementation could need some cleaning up. There is even a [bugfix](https://github.com/LMMS/lmms/commit/cc1cdfa664e7fd548917ae4d9f879770015b34db) in there, but since some editors code doesn't rely on this function and decide to implement it themselves, the bug remains.

### * Opening a useless window is bad

Most modern software does not open a window if you cannot do anything in it. E.g. a text editor always has a document on the screen. In some cases the main window will be blank, but in those cases there is a Open button or something in the toolbar. There are a few programs out there that uses toolbox-like windows that are unusable until the correct action is taken in the main window (image editors etc), but it is fairly obvious that those small windows are helpers to the main window where 95% of the work will happen.

This is not the case for piano-roll. It actually opens a window that remains unusable until the user does the correct thing *in another window*. The design essentially leads the user into a dead end.

allejok96:
Fixes #6234. This PR brings two changes:

### Close Piano Roll when there is no valid clip
...instead of turning the window black and displaying a text. This eliminates the need for all `hasValidMidiClip()` checks, which may be removed in a future PR (I have tested that).

When the user opens a clip, piano roll will re-appear in the same place, as usual. If the user tries to open piano roll when no clip is selected it will say:

> Please open a clip by double clicking on it in the Song Editor

### Create a clip when opening Piano Roll in an empty project
This will no doubt be helpful to new users. Since it only does this for *empty* projects, it won't disturb our workflow.

The clip will be created on the first instrument track. If there are no instrument tracks, it will say:

> Please add an instrument by dragging it from the sidebar to the Song Editor

### DIFF
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -5381,37 +5381,41 @@ bool PianoRollWindow::hasFocus() const
 
 void PianoRollWindow::showEvent(QShowEvent*)
 {
-	// Don't allow piano roll to be opened if there's no clip to display.
-	// To help new users who don't know how to open a clip:
-	// - if there are no clips, create one in the first track and open it
-	// - if there are multiple empty clips, open the first one
-	// - if there's only one clip with notes, open that
-	// - if there are no tracks, or multiple clips, show a help text
+	// PianoRoll can ONLY be shown if hasValidMidiClip is true
+	// TODO remove hasValidMidiClip checks throughout the code
 	if (m_editor->hasValidMidiClip()) { return; }
+
+	// A new user might try to open PianoRoll in an empty project unaware that they first need to create a clip.
+	// To make life easier for them we create and/or open the first clip in an empty project.
+	// If there are multiple non-empty clips, we tell the user to double click one of them instead.
 
 	InstrumentTrack* firstTrack = nullptr;
 	MidiClip* firstEmptyClip = nullptr;
 	MidiClip* firstMelodyClip = nullptr;
 
+	// Iterate through all instrument tracks in the Song
 	for (auto track: Engine::getSong()->tracks())
 	{
 		if (track->type() != Track::Type::Instrument) { continue; }
 		auto instrumentTrack = static_cast<InstrumentTrack*>(track);
-
 		if (!firstTrack) { firstTrack = instrumentTrack; }
 
+		// Iterate through all midi clips in each track
 		for (auto clip: instrumentTrack->getClips())
 		{
 			auto midiClip = static_cast<MidiClip*>(clip);
 
+			// Remember the first empty clip in the Song (ignore subsequent empty clips)
 			if (midiClip->notes().empty())
 			{
 				if (!firstEmptyClip) { firstEmptyClip = midiClip; }
 			}
+			// Remember the first non-empty clip in the Song
 			else if (!firstMelodyClip)
 			{
 				firstMelodyClip = midiClip;
 			}
+			// If there are multiple non-empty clips in the Song, show a hint
 			else
 			{
 				TextFloat::displayMessage(tr("No clip selected"),
@@ -5422,14 +5426,17 @@ void PianoRollWindow::showEvent(QShowEvent*)
 			}
 		}
 	}
+	// If we found a clip, open it
 	if (firstMelodyClip || firstEmptyClip)
 	{
 		m_editor->setCurrentMidiClip(firstMelodyClip ? firstMelodyClip : firstEmptyClip);
 	}
+	// If we found a track with no clips, create a clip and open it
 	else if (firstTrack)
 	{
 		m_editor->setCurrentMidiClip(new MidiClip(firstTrack));
 	}
+	// If we found no instrument tracks, show a hint
 	else
 	{
 		TextFloat::displayMessage(tr("No instrument tracks"),


superpaik:
It works as explained (tested on win10-64b)
The only thing that has a weird behaviour (imo) is in an empty project you open the beat+bassline editor (either you enter notes or not on the bb editor), opening the piano-roll by the icon menu acts as an empty project (creates a clip for the first instrument). Once you open the pianoroll from the bb editor using the popup menu, the icon menu opens the pianoroll for the bb track. Maybe that should be the default behaviour. I mean, if the users opens the bb editor then using the icon menu should open the pianoroll for the instrument/sampler on the bb editor.

allejok96:
> if the users opens the bb editor then using the icon menu should open the pianoroll for the instrument/sampler on the bb editor

@superpaik we cannot read the user's mind. This is only meant for people who have no idea what they are doing. Imagine a first time user opening the BB editor and making an amazing beat with Kicker. The user now clicks the piano roll button to write a melody. It opens an empty TripleOscillator clip. It does not open the Kicker beat. That would be confusing. (Plus I dislike teaching new users to write the entire song in the BB editor)

> opening the piano-roll by the icon menu acts as an empty project [even if there are notes in the bb editor]

This could be an annoying behavior for an experienced user, but it is very easy to get rid of. If you don't want it to create an empty clip on the TripleOcillator track, simply delete the TripleOscillator track.

Spekular:
After some testing I think this is probably better than our current behavior. Better still would be graying out the button and view action, but a bird in the hand [...]

My biggest concern is that the info/warning box is a lot easier to miss than the (admittedly terrible) piano-roll-with-text, but I lean towards merging anyway, once some other devs have chimed in.

Spekular (approved):

allejok96:
For the record: a disabled button/action cannot get triggered() or clicked(), so it can't give an error message. If people wanted we could fake a grey button by changing the pixmap.

src/gui/editors/PianoRoll.cpp
```diff
@@ -5209,10 +5209,54 @@ bool PianoRollWindow::hasFocus() const
 
 
 
+void PianoRollWindow::showEvent(QShowEvent *se)
```
rubiefawn:
```suggestion
void PianoRollWindow::showEvent(QShowEvent*)
```

The parameter is not used in the function body, so it should not be bound to a name.

src/gui/editors/PianoRoll.cpp
```diff
@@ -5209,10 +5209,54 @@ bool PianoRollWindow::hasFocus() const
...
+					parentWidget()->hide();
+					return;
+				}
+			}
+		}
+
+		if (firstTrack)
+		{
+			MidiClip* midiClip = new MidiClip(dynamic_cast<InstrumentTrack*>(firstTrack));
+			m_editor->setCurrentMidiClip(midiClip);
```
rubiefawn:
```suggestion
		InstrumentTrack* firstTrack = nullptr;
		for (Track* track: Engine::getSong()->tracks())
		{
			if (track->type() != Track::InstrumentTrack) { continue; }
			if (!firstTrack) { firstTrack = static_cast<InstrumentTrack*>(track); }

			// Don't create a new clip if the track has clips
			if (track->numOfClips() != 0)
			{
				TextFloat::displayMessage(tr("No active clip"),
					tr("Please open a clip by double clicking on it in the Song Editor."),
					embed::getIconPixmap("error"), 5000);
				parentWidget()->hide();
				return;
			}
		}

		if (firstTrack)
		{
			m_editor->setCurrentMidiClip(new MidiClip(firstTrack));
```

1. Negating the track type check to skip non-instrument tracks instead makes this a bit more readable by reducing nesting.
2. `firstTrack` is only ever assigned an `InstrumentTrack*`, so making its type `InstrumentTrack*` and performing the cast during assignment makes this more clear. Additionally, it can be a `static_cast`, since at that point `track` will never not be an `InstrumentTrack*`.
3. Given the two above changes, there's no need for the temporary `midiClip` variable, it all fits in one line.

src/gui/editors/PianoRoll.cpp
```diff
@@ -5209,10 +5209,54 @@ bool PianoRollWindow::hasFocus() const
...
+
+
 void PianoRollWindow::updateAfterMidiClipChange()
 {
 	clipRenamed();
 	updateStepRecordingIcon(); //MIDI clip change turn step recording OFF - update icon accordingly
+	if (!m_editor->hasValidMidiClip())
+	{
+		parentWidget()->hide();
+	}
```
rubiefawn, line 5456: (Non-blocking) This can be [one line](https://github.com/LMMS/lmms/wiki/Coding-Conventions#whitespace).

```suggestion
	if (!m_editor->hasValidMidiClip()) { parentWidget()->hide(); }
```

allejok96: Thinking about it we don't want to continue updating stuff like title bar etc if there's no midi clip, so I moved it up and added an early return instead

src/gui/editors/PianoRoll.cpp
allejok96: Thanks, good suggestions

allejok96:
I realized that a new user might click around and create multiple empty clips in an attempt to open piano roll, so I extended the logic:

- if there are multiple empty clips, open the first one
- if there's only one clip with notes, open that
- if there are multiple non-empty clips, show the help text

I also updated the wording of the help texts a little

src/gui/editors/PianoRoll.cpp
```diff
@@ -5379,8 +5379,76 @@ bool PianoRollWindow::hasFocus() const
...
+			m_editor->setCurrentMidiClip(new MidiClip(firstTrack));
+		}
+		else
+		{
+			TextFloat::displayMessage(tr("No instrument tracks"),
+				tr("Drag an instrument plugin or preset from the sidebar to the Song Editor."),
+				embed::getIconPixmap("error"), 5000);
+			parentWidget()->hide();
+		}
+	}
```
rubiefawn: Could this be an early return instead?

```suggestion
	if (m_editor->hasValidMidiClip()) { return; }

	InstrumentTrack* firstTrack = nullptr;
	MidiClip* firstEmptyClip = nullptr;
	MidiClip* firstMelodyClip = nullptr;

	for (auto track: Engine::getSong()->tracks())
	{
		if (track->type() != Track::Type::Instrument) { continue; }
		auto instrumentTrack = static_cast<InstrumentTrack*>(track);

		if (!firstTrack) { firstTrack = instrumentTrack; }

		for (auto clip: instrumentTrack->getClips())
		{
			auto midiClip = static_cast<MidiClip*>(clip);

			if (midiClip->notes().empty())
			{
				if (!firstEmptyClip) { firstEmptyClip = midiClip; }
			}
			else if (!firstMelodyClip)
			{
				firstMelodyClip = midiClip;
			}
			else
			{
				TextFloat::displayMessage(tr("No clip selected"),
					tr("Double click a melody clip in the Song Editor to open it."),
					embed::getIconPixmap("error"), 5000);
				parentWidget()->hide();
				return;
			}
		}
	}
	if (firstMelodyClip || firstEmptyClip)
	{
		m_editor->setCurrentMidiClip(firstMelodyClip ? firstMelodyClip : firstEmptyClip);
	}
	else if (firstTrack)
	{
		m_editor->setCurrentMidiClip(new MidiClip(firstTrack));
	}
	else
	{
		TextFloat::displayMessage(tr("No instrument tracks"),
			tr("Drag an instrument plugin or preset from the sidebar to the Song Editor."),
			embed::getIconPixmap("error"), 5000);
		parentWidget()->hide();
	}
```

regulus79:
I believe we are good to merge?

PR #6297, opened 2022-02-02
pitch wheel recording in the detune pattern

### LINKED ISSUE #3724: Record pitch automation directly into piano roll
Somewhat a duplicating what was requested in #702 (closed, merged with #1472), LMMS needs a way to record pitch directly into the Piano Roll.

It turns out, pitch can be recorded quite effectively now, and this is how... 

1. Connect instrument to MIDI
   * Instrument window
   * Midi tab
   * Enable MIDI input
   * Select device
2. Ctrl + Drag Instrument Pitch Wheel to Automation Track
3. Right Click Automation Pattern, Set/Clear Record (CAREFUL with this, playing will wipe out your automation pattern!)

   screen shot 2017-07-23 at 3 50 09 pm

If your midi device has a pitch wheel, this is very effective in capturing the pitch events that make it to the instrument.

However, the pitch support has now been added to the piano roll through "Detuning mode", so this should (eventually) be recorded directly into the piano roll's pattern-specific automation area instead, without having to setup an additional automation track.

Note, this would require the detuning mode to be scaled to the instrument pitch wheel range (e.g. 1x, 2x, etc)

screen shot 2017-07-23 at 3 51 41 pm

serdnab:
Closes  #3724

When creating a note from midi origin, it now sets the position as the current playing position.
If piano roll is recording, it records the value of the instrument pitch wheel in the detune pattern of the current recording midi note.
When finishing recording of a note, it sets the detune pattern of the new note with the detune pattern of the midi note.
I extended the detune pattern range to 60 semitones, since that is the pitch wheel max range.


RiedleroD:
Could this be expanded to take automation from midi import as well? Would drastically decrease the amount of automation tracks I'd have to carry over, so that'd be very helpful to me.

allejok96:
@serdnab this is awesome! Works like a charm. Just a few things: a detune pattern with a single point is now created for every note even if the pitch wheel is untouched. This will eat memory. And the detune "sample rate" is far too low if you want to do vibrato. (One point per beat?) No reason to hold back on that one, since it will only be used occasionally in a song.

allejok96:
(I think this is another problem, but I'm curious) @RiedleroD can you explain the process you are currently going through when you are carrying over automation for an imported midi file?

RiedleroD:
@allejok96 well if there's lots of pitch automation I leave it as it is (too much work), but sometimes there's just one or two pitchbends which I just approximately recreate using note bending to declutter the project. Most of the time though, there's lots of pitch automation tracks with a single pitchbend to 0 - which are really annoying, but also really easy to remove in a separate PR I guess.

If you'd integrate pitch automation from midi import into note bending, all of that work would just disappear. Plus I'd have even cleaner files than before. Though yeah, it's probably better in a separate PR. Should I create a feature request for this?

serdnab:
> a detune pattern with a single point is now created for every note even if the pitch wheel is untouched

This is a separate bug, as it happens always you create a note, even if you don't record.
> the detune "sample rate" is far too low

If you open the detune editor and you change the quantization before recording any notes, you can increase the "sample rate".
> Should I create a feature request for this?

Yes, please, as that would require some work.

allejok96:
Adding this in `InstrumentTrack::InstrumentTrack()` will solve my problem. It will restrict the pitch knob to whole cents.

    m_pitchModel.setStrictStepSize(true);

Since the problem arises with this PR I think it makes sense to fix it here.

src/core/NotePlayHandle.cpp
```diff
@@ -29,9 +29,11 @@
 #include "AudioEngine.h"
 #include "BasicFilters.h"
 #include "DetuningHelper.h"
+#include "GuiApplication.h"
 #include "InstrumentSoundShaping.h"
 #include "InstrumentTrack.h"
 #include "Instrument.h"
+#include "PianoRoll.h"
```
Veratil, line 36: Since this is a core file, we will not want to add GuiApplication.h or PianoRoll.h here.

src/core/NotePlayHandle.cpp
```diff
@@ -555,15 +557,22 @@ void NotePlayHandle::updateFrequency()
 
 
 
-void NotePlayHandle::processTimePos( const TimePos& time )
+void NotePlayHandle::processTimePos(const TimePos& time, float pvalue)
 {
-	if( detuning() && time >= songGlobalParentOffset()+pos() )
+	if (detuning() && time >= songGlobalParentOffset() + pos())
```
Veratil, line 562: Since this is a guard for the entire function, perhaps we can invert it so we don't indent the entire function?
```suggestion
	if (!detuning() || time < songGlobalParentOffset() + pos()) { return; }
```

src/core/NotePlayHandle.cpp
```diff
@@ -555,15 +557,22 @@ void NotePlayHandle::updateFrequency()
 
 
 
-void NotePlayHandle::processTimePos( const TimePos& time )
+void NotePlayHandle::processTimePos(const TimePos& time, float pvalue)
```
Veratil, line 560: Perhaps a little more descriptive name for `pvalue`?

src/core/NotePlayHandle.cpp
```diff
@@ -555,15 +557,22 @@ void NotePlayHandle::updateFrequency()
...
 {
-	if( detuning() && time >= songGlobalParentOffset()+pos() )
+	if (detuning() && time >= songGlobalParentOffset() + pos())
 	{
-		const float v = detuning()->automationClip()->valueAt( time - songGlobalParentOffset() - pos() );
-		if( !typeInfo<float>::isEqual( v, m_baseDetuning->value() ) )
+		if (GuiApplication::instance()->pianoRoll()->isRecording() && m_origin == Origin::OriginMidiInput)
```
Veratil, line 564: To remove `GuiApplication` here we can probably add a default boolean to the function for `isRecording`.

Veratil (changes requested):

src/core/NotePlayHandle.cpp
serdnab: Is it acceptable to add them to InstrumentTrack.cpp?

serdnab:
> Since the problem arises with this PR I think it makes sense to fix it here.

I will commit it.

src/core/NotePlayHandle.cpp
Veratil: For now it might be, but if we can find a way not doing it we should. I know it'll make it more complicated, but we want to try to keep core and GUI as separate as we can.

serdnab:
Requested changes addressed.
I looked for another way to check if the piano roll is recording, but I didn't find it.
If you have any ideas?

Veratil:
> Requested changes addressed. I looked for another way to check if the piano roll is recording, but I didn't find it. If you have any ideas?

It'll require some re-architecting later, but for now this is fine. It's a good enhancement. 👍

Veratil (approved):

Veratil:
@serdnab can you rebase and fix the conflicts? We can merge after that.

zonkmachine:
@serdnab @Veratil I think I've managed to update the PR and you can always revert to [2813ee2](https://github.com/LMMS/lmms/pull/6297/commits/2813ee29f49a909c31a0d415e060a06c197bc72b)
The build no longer works and fails on a GuiApplication related issue. It looks to me like that is due to changes on master since the last commit before I touched it.

```make
[  0%] Building CXX object src/CMakeFiles/lmmsobjs.dir/tracks/InstrumentTrack.cpp.o
/builds/lmms/src/tracks/InstrumentTrack.cpp: In member function ‘virtual bool lmms::InstrumentTrack::play(const lmms::TimePos&, lmms::fpp_t, lmms::f_cnt_t, int)’:
/builds/lmms/src/tracks/InstrumentTrack.cpp:717:77: error: ‘GuiApplication’ has not been declared
  717 |                 processHandle->processTimePos(_start, m_pitchModel.value(), GuiApplication::instance()->pianoRoll()->isRecording());
      |                                                                             ^~~~~~~~~~~~~~
make[2]: *** [src/CMakeFiles/lmmsobjs.dir/build.make:3361: src/CMakeFiles/lmmsobjs.dir/tracks/InstrumentTrack.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:2136: src/CMakeFiles/lmmsobjs.dir/all] Error 2
make: *** [Makefile:156: all] Error 2
```

zonkmachine:
Above issue fixed. We all bow deeply to @sakertooth

Tested: We got pitch bend on record. Damn!
All hail @serdnab!

zonkmachine (approved):

messmerd (approved):

PR #6300, opened 2022-02-06
Fix small memory leaks in EQ dialog

### LINKED ISSUE #6026: Crash on exit from rpmalloc assert in debug builds, equalizer plugin related
### Bug Summary
`rpmalloc` aborts from an assertion on program exit if the equalizer plugin is added to an FX track.
#5733 might be related, which was thought to be fixed by #5776.

Note: The few other plugins I've tested don't seem to cause this, only equalizer. I haven't tested many other plugins though.

#### Steps to reproduce
1. Run a debug build of LMMS.
2. Add the equalizer plugin to either the master FX track or a created FX track.
3. Close LMMS via the 'X' button or File -> Quit.

#### Expected behavior
LMMS closes without aborting preemptively.

#### Actual behavior
LMMS "crashes" from the assert calling an abort.

#### Screenshot
image

#### Affected LMMS versions
Version 1.3.0-alpha.1.105
Debug configuration.

Unknown if on stable because I don't have the know-how to build and test 1.2.2

#### Logs
(From console)

`Assertion failed: span->list_size == span->used_count, file C:\LMMSDev\lmms\src\3rdparty\rpmalloc\rpmalloc\rpmalloc\rpmalloc.c, line 1248`

#### Note
Probably not a big deal since it happens on program exit, so the OS will (hopefully) clean up anyway, but still something to consider.

PhysSong:
Fixes #6026.

### DIFF
--- a/plugins/Eq/EqControlsDialog.cpp
+++ b/plugins/Eq/EqControlsDialog.cpp
@@ -201,13 +201,13 @@ EqControlsDialog::EqControlsDialog( EqControls *controls ) :
 	lpBtnGrp->addButton( lp12Button );
 	lpBtnGrp->addButton( lp24Button );
 	lpBtnGrp->addButton( lp48Button );
-	lpBtnGrp->setModel( &m_controls->m_lpTypeModel, false);
+	lpBtnGrp->setModel(&m_controls->m_lpTypeModel);
 
 	automatableButtonGroup *hpBtnGrp = new automatableButtonGroup( this, tr( "HP group" ) );
 	hpBtnGrp->addButton( hp12Button );
 	hpBtnGrp->addButton( hp24Button );
 	hpBtnGrp->addButton( hp48Button );
-	hpBtnGrp->setModel( &m_controls->m_hpTypeModel,false);
+	hpBtnGrp->setModel(&m_controls->m_hpTypeModel);
 }
 
 


Spekular:
Breaking this down for myself:

`lpBtnGrp` and `hpBtnGrp` are instances of `automatableButtonGroup`, which is defined in `AutomatableButton`. They inherit from `IntModelView`, which is a `TypedModelView`, which is an `AutomatableModelView`, which has the `setModel` method that then calls `ModelView::setModel`. The parameter being implicitly changed to `true` here is `isOldModelValid`, which allows us to enter this if statement:

https://github.com/LMMS/lmms/blob/f74271075874c482e4c6af0bd0bb0febe1042314/src/gui/ModelView.cpp#L53

Because `automatableButtonGroup` creates an `IntModel` with `defaultConstructed` set to `true`...

https://github.com/LMMS/lmms/blob/f74271075874c482e4c6af0bd0bb0febe1042314/src/gui/widgets/AutomatableButton.cpp#L183
https://github.com/LMMS/lmms/blob/f74271075874c482e4c6af0bd0bb0febe1042314/include/AutomatableModel.h#L477-L480

... we enter the next if statement, where this useless default model is deleted

https://github.com/LMMS/lmms/blob/f74271075874c482e4c6af0bd0bb0febe1042314/src/gui/ModelView.cpp#L57

Whew.

PR #6308, opened 2022-02-13
Use automation track for MIDI program change events

### LINKED ISSUE #6305: MIDI import doesn't import program change automation
### Bug Summary
When importing a MIDI file, LMMS doesn't create an automation track when there are multiple program change events and uses the last program number found for the whole track instead. The problem doesn't come from the MIDI files I used, as they do contain multiple program change events and VLC and Cubase react correctly to them.

#### Steps to reproduce
1. Import a MIDI file with program change events (I can provide one if needed)
2. Look at the automation tracks: you may see a bank change track but no program change track

#### Expected behavior
LMMS should create an automation track when multiple program change events are found in a MIDI file similarly to other events

#### Actual behavior
The last seen program number is used on the entire track

#### Affected LMMS versions
1.2.2

Veratil:
Fixes #6305 

Creates an automation track similar to the Pitch/Panning/Volume update events. Tested with a few MIDIs I've saved for testing breaking LMMS MIDI import, along with both MIDIs provided by Rams in 6305.

Used an unordered map to dynamically create channel CC objects similar to the channels one. Reusing the `smfMidiCC` since it fits perfectly with the program change updates as well.

### DIFF
--- a/plugins/MidiImport/MidiImport.cpp
+++ b/plugins/MidiImport/MidiImport.cpp
@@ -324,6 +324,9 @@ bool MidiImport::readSMF( TrackContainer* tc )
 	// 128 CC + Pitch Bend
 	smfMidiCC ccs[MIDI_CC_COUNT];
 
+	// channel to CC object for program changes
+	std::unordered_map<long, smfMidiCC> pcs;
+
 	// channels can be set out of 256 range
 	// using unordered_map should fix most invalid loads and crashes while loading
 	std::unordered_map<long, smfMidiChannel> chs;
@@ -472,8 +475,12 @@ bool MidiImport::readSMF( TrackContainer* tc )
 					long prog = evt->get_integer_value();
 					if( ch->isSF2 )
 					{
-						ch->it_inst->childModel( "bank" )->setValue( 0 );
-						ch->it_inst->childModel( "patch" )->setValue( prog );
+						auto& pc = pcs[evt->chan];
+						AutomatableModel* objModel = ch->it_inst->childModel("patch");
+						if (pc.at == nullptr) {
+							pc.create(tc, trackName + " > " + objModel->displayName());
+						}
+						pc.putValue(time, objModel, prog);
 					}
 					else {
 						const QString num = QString::number( prog );


superpaik:
It crashes when playing, on Win10-64b
Test: import the linked midi (it's not mine, but downloaded from internet). Hit play. Crash.
It works ok on 1.2. The new version creates the automation patches for the tracks, which are not visible on v1.2

https://mega.nz/file/8UAGDJRZ#OR4Ies3IXoEk1lxrVtGRipSehZ_3P9oVUu4jpu0YlPs

Veratil:
> It crashes when playing, on Win10-64b Test: import the linked midi (it's not mine, but downloaded from internet). Hit play. Crash. It works ok on 1.2. The new version creates the automation patches for the tracks, which are not visible on v1.2
> 
> https://mega.nz/file/8UAGDJRZ#OR4Ies3IXoEk1lxrVtGRipSehZ_3P9oVUu4jpu0YlPs

This looks to be a playback bug, not an import bug.

PhysSong:
I could reproduce the crash, but current `master` also crashed. It happens with a note with pitch C-1(=0). `stable-1.2` is fine because the note pitch range is restricted to C0-C8, unless you transpose down by an octave or more using master pitch and base note.
https://github.com/LMMS/lmms/blob/f56fc68b669b59f525ad133a0376feaba2a1a4ec/plugins/sf2_player/sf2_player.cpp#L634-L637
So I conclude the crash is not related to this PR. I will open a separate issue and PR for the crash
@superpaik BTW, you can upload a `.zip` file directly while writing a comment.

superpaik:
Oh, I see. I tried to attach the midi file directly, but I couldn't. I didn't think about make it a zip. Thanks!

Maybe the problem was introduced here https://github.com/LMMS/lmms/pull/5349

ScolderCreations (commented):
Wait, what happens to the bank data?

Veratil:
> Wait, what happens to the bank data?

bank select is handled through control change event, patch select is a program change event.

bank select handler: https://github.com/LMMS/lmms/blob/master/plugins/MidiImport/MidiImport.cpp#L506

SDwarfs:
> `if( midiNote <= 0 || midiNote >= 128 )`
>  { 
>  	return; 
>  }

The above code actually seems to cause the crash. I dived into the code and found #6406 to be fixed, when I change the if condition to:
> if (midiNote < 0 || midiNote >= 128)

The actual crash happens at Sf2Player.cpp:658 when trying to set  "pluginData->offset =", where pluginData is a cast version of the pointer `n->pluginData, being unitialized.` This usual initialization seems to happen after the above if statement, which does not happen for the case midiNote == 0 due to the "<= 0". As midiNote == 0 seems to be a valid value, this change seems reasonable to me.

Suggested new code (plugins/Sf2Player/Sf2Player.cpp:633):
> `if( midiNote < 0 || midiNote >= 128 )`
>  { 
>  	return; 
>  }

I've no idea how to send you a git "push" request or whatever needed for this change. I'd be glad If someone can send me info how to do it... or just adapt the code accordingly. Thanks!

DomClark:
> I dived into the code and found #6406 to be fixed, when I change the if condition to: `if (midiNote < 0 || midiNote >= 128)`.

See my review comment on #6358, which makes this exact change. The if block is effectively
```cpp
if (midiNote out of range)
{
    crash;
}
```
The solution to that is not to reduce the range of arguments that send control flow down the crashing path; it's to remove the crashing path altogether.

The range is still worth changing, since SF2 player should be able to play C-1 notes, but it is not a fix for the crash.

Veratil:
So... Just to let y'all know that the midi note crash is off topic to this PR... :)

PR #6313, opened 2022-02-16
Make loop markers snap to same increments as clips

### LINKED ISSUE #6007: The new quantization in pianoroll should affect the looping points, too
As the title said, quantization in pianoroll should affect also the looping points.

For example setting this
immagine

will allow the user to move pattern by a half bar, but the looping points can still only be moved by a full bar or completely precisely holding ctrl.

In my opinion setting 1/2 bar in pianoroll should make the looping points move by 1/2 bar, too and so on.

I could be useful if someone want to do some more precise looping.

Spekular:
A bit of an oversight from the enhanced song editor snapping PR, now fixed. Tested and working on my end, unsure if there's a cleaner way to make sure that the snap size is updated when proportional snap is toggled.

Closes #6007

### DIFF
--- a/include/TimeLineWidget.h
+++ b/include/TimeLineWidget.h
@@ -180,7 +180,7 @@ public slots:
 	{
 		updatePosition( TimePos() );
 	}
-	void updateSnapSize( const float snapSize )
+	void setSnapSize( const float snapSize )
 	{
 		m_snapSize = snapSize;
 	}
--- a/include/TimeLineWidget.h
+++ b/include/TimeLineWidget.h
@@ -180,6 +180,10 @@ public slots:
 	{
 		updatePosition( TimePos() );
 	}
+	void updateSnapSize( const float snapSize )
+	{
+		m_snapSize = snapSize;
+	}
 	void toggleAutoScroll( int _n );
 	void toggleLoopPoints( int _n );
 	void toggleBehaviourAtStop( int _n );
@@ -217,6 +221,7 @@ private:
 	int m_xOffset;
 	int m_posMarkerX;
 	float m_ppb;
+	float m_snapSize;
 	Song::PlayPos & m_pos;
 	const TimePos & m_begin;
 	const Song::PlayModes m_mode;
--- a/src/gui/TimeLineWidget.cpp
+++ b/src/gui/TimeLineWidget.cpp
@@ -413,11 +413,11 @@ void TimeLineWidget::mouseMoveEvent( QMouseEvent* event )
 				// marking instead of pushing it.
 				if( m_action == MoveLoopBegin ) 
 				{
-					m_loopPos[0] -= TimePos::ticksPerBar();
+					m_loopPos[0] -= m_snapSize * TimePos::ticksPerBar();
 				}
 				else
 				{
-					m_loopPos[1] += TimePos::ticksPerBar();
+					m_loopPos[1] += m_snapSize * TimePos::ticksPerBar();
 				}
 			}
 			update();
--- a/src/gui/TimeLineWidget.cpp
+++ b/src/gui/TimeLineWidget.cpp
@@ -344,7 +344,7 @@ void TimeLineWidget::mousePressEvent( QMouseEvent* event )
 		{
 			m_action = MoveLoopBegin;
 		}
-		else if( t > loopMid )
+		else
 		{
 			m_action = MoveLoopEnd;
 		}
--- a/src/gui/TimeLineWidget.cpp
+++ b/src/gui/TimeLineWidget.cpp
@@ -62,6 +62,7 @@ TimeLineWidget::TimeLineWidget( const int xoff, const int yoff, const float ppb,
 	m_xOffset( xoff ),
 	m_posMarkerX( 0 ),
 	m_ppb( ppb ),
+	m_snapSize( 1.0 ),
 	m_pos( pos ),
 	m_begin( begin ),
 	m_mode( mode ),
@@ -403,7 +404,7 @@ void TimeLineWidget::mouseMoveEvent( QMouseEvent* event )
 			}
 			else
 			{
-				m_loopPos[i] = t.quantize(1.0);
+				m_loopPos[i] = t.quantize(m_snapSize);
 			}
 			// Catch begin == end
 			if( m_loopPos[0] == m_loopPos[1] )
--- a/src/gui/TimeLineWidget.cpp
+++ b/src/gui/TimeLineWidget.cpp
@@ -341,7 +341,7 @@ void TimeLineWidget::mousePressEvent( QMouseEvent* event )
 		const TimePos loopMid = ( m_loopPos[0] + m_loopPos[1] ) / 2;
 
 		m_action = t < loopMid ? MoveLoopBegin : MoveLoopEnd;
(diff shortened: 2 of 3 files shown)


superpaik:
There is a weird behaviour, unsure if it's due to this PR. Anyway, check the attached video
snap

Spekular:
I'll put it in the same PR, but try to leave it as a separate commit on merge (squashing the first three to one).

include/TimeLineWidget.h
```diff
@@ -180,6 +180,10 @@ public slots:
 	{
 		updatePosition( TimePos() );
 	}
+	void updateSnapSize( const float snapSize )
```
DomClark, line 183: Wouldn't `setSnapSize` be a more conventional name?

src/gui/editors/SongEditor.cpp
```diff
@@ -101,6 +101,12 @@ SongEditor::SongEditor( Song * song ) :
 			 m_positionLine, SLOT( update() ) );
 	connect( this, SIGNAL( zoomingValueChanged( float ) ),
 			 m_positionLine, SLOT( zoomChange( float ) ) );
+			 
+	// ensure loop markers snap to correct size
+	connect(m_snappingModel, &ComboBoxModel::dataChanged,
+		[=]() { m_timeLine->updateSnapSize(getSnapSize()); });
```
DomClark, line 107: Implicit capture of `this` using `[=]` is deprecated in C++20. Not really a problem since we're only on C++17, but explicit capture would be better for the future.

DomClark (approved):
Looks good, only some minor niggles.

include/TimeLineWidget.h
Spekular: Good point! Fixed.

src/gui/editors/SongEditor.cpp
Spekular: Fixed!

src/gui/editors/SongEditor.cpp
```diff
@@ -101,6 +101,12 @@ SongEditor::SongEditor( Song * song ) :
...
+	connect(m_zoomingModel, &ComboBoxModel::dataChanged,
+		[=]() { m_timeLine->updateSnapSize(getSnapSize()); });
```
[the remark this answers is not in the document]
Spekular: I figured doing all the updates via connections would give a good overview of which actions update the snap size, but it does make a lot of sense to use the existing functions for zoom and proportional snap toggle. I've updated the comment to provide an overview instead.

src/gui/TimeLineWidget.cpp
```diff
@@ -412,11 +413,11 @@ void TimeLineWidget::mouseMoveEvent( QMouseEvent* event )
 				// marking instead of pushing it.
 				if( m_action == MoveLoopBegin ) 
 				{
-					m_loopPos[0] -= TimePos::ticksPerBar();
+					m_loopPos[0] -= m_snapSize * TimePos::ticksPerBar();
```
JohannesLorenz, line 416: I tried my best to get negative values here (e.g. by changing bar sizes), but without success :+1: Likely because you call `t.quantize()` right before.

JohannesLorenz (approved):
I did a code review and some testing - LGTM.

This could be applied to PianoRoll in a future PR aswell.

JohannesLorenz:
Sorry, @Spekular , now I see something. Use ctrl+right-mouse to adjust both markers close between 0 and 1 (e.g. 0.3 and 0.4). Change snap size to 8 bars. Then press ctrl again, and drag the lower marker a bit up and down, also sometimes above the upper marker. It sometimes flickers then. Not sure if this is even related to your PR.

src/gui/TimeLineWidget.cpp
```diff
@@ -403,7 +404,7 @@ void TimeLineWidget::mouseMoveEvent( QMouseEvent* event )
 			}
 			else
 			{
-				m_loopPos[i] = t.quantize(1.0);
+				m_loopPos[i] = t.quantize(m_snapSize);
 			}
 			// Catch begin == end
 			if( m_loopPos[0] == m_loopPos[1] )
```
JohannesLorenz, line 410: I guess this should not be done when Ctrl is being pressed (see my other comment).

Spekular:
> Sorry, Spekular, now I see something. Use ctrl+right-mouse to adjust both markers close between 0 and 1 (e.g. 0.3 and 0.4). Change snap size to 8 bars. Then press ctrl again, and drag the lower marker a bit up and down, also sometimes above the upper marker. It sometimes flickers then. Not sure if this is even related to your PR.

Thanks for the review @JohannesLorenz ! Latest commit should fix this, unless I did the repro steps wrong. Mind checking again?

JohannesLorenz:
> Latest commit should fix this, unless I did the repro steps wrong. Mind checking again?

Thanks it did fix it. Indeed, this failed before this PR, so it makes this PR a bugfix, additionally.

I just don't understand why it does not work if you do not subtract anything in the ctrl case. I guess it's just required somewhere else in the code that both markers are different, and that's why you delay one of the markers by one pixel?

It would be nice if one of the previous reviewers (@DomClark or @Veratil) could approve the new changes.

Spekular:
> I just don't understand why it does not work if you do not subtract anything in the ctrl case. I guess it's just required somewhere else in the code that both markers are different, and that's why you delay one of the markers by one pixel?

The nudge to avoid start == end was present before this PR, it's just the distance that now depends on snap size and ctrl being held. I don't know if it's strictly required, but I think it seems sound from a UX perspective to not allow a zero length loop.

JohannesLorenz:
By the way, the fact that line 340 subtracts `m_moveXOff` seems wrong to me. This offset is only relevant for the play position marker (triangle above the play pos line), but we care about the loop positions, which have no offset. I guess if you remove that subtraction, the markers will adjust the mouse pointer exactly, while currently, they have an offset of ~5-10.

But this is not the PR's fault.

src/gui/TimeLineWidget.cpp
```diff
@@ -339,20 +340,8 @@ void TimeLineWidget::mousePressEvent( QMouseEvent* event )
...
-			m_action = MoveLoopEnd;
-		}
-
-		if( m_loopPos[0] > m_loopPos[1]  )
-		{
-			qSwap( m_loopPos[0], m_loopPos[1] );
-		}
-
+		m_action = t < loopMid ? MoveLoopBegin : MoveLoopEnd;
+		std::sort(m_loopPos, std::end(m_loopPos));
```
DomClark, line 344: I think `std::begin(m_loopPos)` would be better here - `m_loopPos` alone only works in this case because an array decays to a pointer to its first element, which happens to be the same type as `std::end(m_loopPos)`. Most iterable types aren't compatible with their iterator types (the only other ones that come to mind are the directory iterators in the filesystem library), so this wouldn't work if we changed from a raw C-style array to `std::array`, for instance.

Spekular: Good point! Fixed locally.

Spekular:
@DomClark thanks for re-review. Any other concerns before I push the `std::begin` fix? Otherwise I might merge after the new commit passes CI.

DomClark (approved):

PR #6320, opened 2022-02-28
Improve includes

JohannesLorenz:
As requested in #5592 .

The last commit has been done using iwyu (see commit for details) and a lot of manual refactoring. The commits before it are preparation.


src/core/midi/MidiSndio.cpp
```diff
@@ -27,17 +27,16 @@
 
 #ifdef LMMS_HAVE_SNDIO
 
-#include <QLabel>
-#include <QLineEdit>
 
 #ifdef LMMS_HAVE_STDLIB_H
-#include <stdlib.h>
+#include <cstdlib>
 #endif
```
DomClark, line 33: Should the CMake header detection be updated to check for the new header?

include/Oscillator.h
```diff
@@ -28,7 +28,6 @@
 
 #include <cassert>
 #include <fftw3.h>
-#include <math.h>
 
 #ifdef LMMS_HAVE_STDLIB_H
 #include <stdlib.h>
```
DomClark, line 33: There still seem to be a few of these C headers lying around.

src/core/ComboBoxModel.cpp
```diff
@@ -37,7 +36,7 @@ void ComboBoxModel::addItem( QString item, unique_ptr<PixmapLoader> loader )
 
 void ComboBoxModel::replaceItem(std::size_t index, QString item, unique_ptr<PixmapLoader> loader)
 {
-	assert(index < m_items.size());
+	Q_ASSERT(index < m_items.size());
```
DomClark, line 39: I believe we're slowly trying to remove Qt from the core, so I think a standard `assert` is actually better here.

PhysSong: One more difference: `Q_ASSERT` doesn't get removed in release builds unless `QT_NO_DEBUG` is defined.

src/core/midi/MidiSndio.cpp
PhysSong: Can we even remove some checks? According to https://en.cppreference.com/w/cpp/freestanding, it looks `<cstdlib>` and `<cstdint>` can be included without any checks.

src/core/VstSyncController.cpp
```diff
@@ -25,16 +25,14 @@
 
 #include <QDebug>
 
+#include "VstSyncController.h"
+
 #include "AudioEngine.h"
 #include "ConfigManager.h"
 #include "Engine.h"
-#include "VstSyncController.h"
 #include "RemotePlugin.h"
```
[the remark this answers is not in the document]
JohannesLorenz: IWYU might have done it correct. It's me who added that `#include`. I did this, because by default, IWYU would have included headers for everything that is being used, even if transitively included. So it's human fault :smiley: 

Anyways, will fix in next commit.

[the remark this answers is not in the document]
JohannesLorenz: (Note that clang-format would have fixed the order anyways)

src/core/midi/MidiSndio.cpp
JohannesLorenz: Agreed to completely remove, especially since we often use that header without that macro and it still works.

include/Oscillator.h
JohannesLorenz: Indeed. I did it only for the CPP files. Will be fixed with next push.

src/core/ComboBoxModel.cpp
JohannesLorenz: Agreed, switching back to `assert` then.

src/core/midi/MidiSndio.cpp
JohannesLorenz: I removed those 2 now completely from LMMS. From your link, there seem to be no more such headers, so that should be enough.

JohannesLorenz:
I reworked your comments and pushed.

For some headers, I think this PR can damage more than it should. For example, if you remove "lmmsconfig.h" accidentally, the code might still compile, but be configured wrong. Though, I didn't notice more headers with that issue, and at least for this one, the worst case seems to be "LMMS compiles more than I wanted".

Another point (already mentioned in one comment): The include order in this PR is not too crucial, as clang-format will reorder all of this - except if there are `#if...` or (FWIR) comments in between.

DomClark (approved):
> For some headers, I think this PR can damage more than it should. For example, if you remove "lmmsconfig.h" accidentally, the code might still compile, but be configured wrong. Though, I didn't notice more headers with that issue, and at least for this one, the worst case seems to be "LMMS compiles more than I wanted".

I'm not sure if there's an easy way to check for this. CI is passing, which is the main thing, so I'd say we merge this and fix up any issues later.

zonkmachine:
@JohannesLorenz I'm on a brand new laptop now and Ubuntu 21.10, fresh install.
7db3fa94a103e2561de397c796b3b28ebd1112b5 generates this error. The commit prior to this builds just fine.

```cpp
Building CXX object src/CMakeFiles/lmmsobjs.dir/core/MixHelpers.cpp.o
/home/zonkmachine/builds/lmms/src/core/MixHelpers.cpp: In function ‘bool MixHelpers::sanitize(sampleFrame*, int)’:
/home/zonkmachine/builds/lmms/src/core/MixHelpers.cpp:105:41: error: ‘printf’ was not declared in this scope
  105 |                                         printf("Bad data, clearing buffer. frame: ");
      |                                         ^~~~~~
/home/zonkmachine/builds/lmms/src/core/MixHelpers.cpp:32:1: note: ‘printf’ is defined in header ‘<cstdio>’; did you forget to ‘#include <cstdio>’?
   31 | #include "ValueBuffer.h"
  +++ |+#include <cstdio>
   32 | 
make[2]: *** [src/CMakeFiles/lmmsobjs.dir/build.make:652: src/CMakeFiles/lmmsobjs.dir/core/MixHelpers.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:2319: src/CMakeFiles/lmmsobjs.dir/all] Error 2
make: *** [Makefile:171: all] Error 2
```

PhysSong:
It seems like debug builds(with `LMMS_DEBUG`) wasn't checked properly. Also, there might be some chances where the code won't build without some optional dependencies due to missing headers.

PhysSong:
Confirmed that it's the only build error, at least on Linux.

DomClark:
When added back, that include ought to be wrapped in `#ifdef LMMS_DEBUG` so it isn't erroneously removed again. Also, we shouldn't be calling `printf` from a real-time thread, even in debug mode.

JohannesLorenz:
Sorry, I really have problems with my notifications lately, but luckily, @PhysSong already fixed it in 80a6672da46a58dd6865a3dda8a7a4f05776393e .

PR #6322, opened 2022-03-05
Split RemotePlugin.h into separate files

DomClark:
This PR splits `include/RemotePlugin.h` into smaller parts, since it contained code serving several different purposes. The resulting files are as follows:
* `include/RemotePlugin.h`:
  This now only contains code that is exclusive to the main LMMS process. It is the header corresponding to `src/core/RemotePlugin.cpp`. It contains the code that was wrapped in `#ifndef BUILD_REMOTE_PLUGIN_CLIENT` at the top level: original lines 748-900 and 1449-1453.
* `include/RemotePluginBase.h`:
  This contains declarations and inline definitions common to the main LMMS process and the remote plugin processes. It is the header corresponding to `src/common/RemotePluginBase.cpp`. It contains the code that was not wrapped by `#ifdef`s at the top level: original lines 1-743.
* `include/RemotePluginClient.h`:
  This contains code that is exclusive to the remote plugin processes. It contains the code that was wrapped in `#ifdef BUILD_REMOTE_PLUGIN_CLIENT` at the top level: original lines 905-1004 and 1186-1444.
* `src/common/RemotePluginBase.cpp`:
  This contains out-of-line definitions common to the main LMMS process and the remote plugin processes. It contains the code that was wrapped in `#ifdef COMPILE_REMOTE_PLUGIN_BASE` at the top level: original lines 1012-1175.

Only CMake and preprocessor code is modified in this PR - all actual C++ code is cut and pasted unmodified (including no style fixes).

There is still some entangling of remote and non-remote code in the common files. This would require some refactoring to fix beyond a simple cut-and-paste job. This should get better as I work on the IPC system, but I don't want to put effort into refactoring something now that I might end up removing later.


plugins/vst_base/RemoteVstPlugin/CMakeLists.txt
```diff
@@ -22,6 +22,8 @@ FOREACH( OUTPUTCONFIG ${CMAKE_CONFIGURATION_TYPES} )
 	SET("CMAKE_RUNTIME_OUTPUT_DIRECTORY_${OUTPUTCONFIG}" "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}")
 ENDFOREACH()
 
+ADD_SUBDIRECTORY(../../../src/common common)
```
PhysSong, line 25: Any advantages of using relative path rather than `LMMS_SOURCE_DIR`?

DomClark: I think `LMMS_SOURCE_DIR` is better. Fixed here and in `plugins/zynaddsubfx/CMakeLists.txt`.

PhysSong (approved):
LGTM

PR #5948, opened 2021-03-13
Clang tools testing

JohannesLorenz:
This is not intended to be merged (I'll do a new branch for the official). The intention is to

1. See if you're happy with clang's changes (comment the code/conversation if unsatisfied, please)
2. Find out how difficult it gets to rebase other branches on this branch

Rebasing your branch on this *might* work like this for **single-commit** branches:

1. `clang-format -style=file -i $(git ls-files '*.cpp' '*.h')`
2. `git commit -am 'fixup'`
3. `git rebase -i HEAD~2`, Chooese `f` for the second commit, accept.
4. `git rebase clang-formatted`

This is a testing brach for #4690 .

PR #6323, opened 2022-03-05
Apply clang-format

JohannesLorenz:
This code here should be the same as in #5948 . Asking for final review.

A good "playground" might be src/gui/editors/PianoRoll.cpp (>5000 lines).

Note: The review can also end with "Let's better not take clang-format", that is also OK.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,84 @@
+---
+# Language
+Language:        Cpp
+Standard:        Cpp11  # Cpp14 and Cpp17 are not supported by clang 11
+
+# Indentation
+TabWidth:        4
+UseTab:          Always
+IndentWidth:     4
+ColumnLimit:     120
+
+# Indentation detail
+AlignAfterOpenBracket: DontAlign
+ContinuationIndentWidth: 4
+BreakConstructorInitializers: BeforeComma
+ConstructorInitializerIndentWidth: 4
+ConstructorInitializerAllOnOneLineOrOnePerLine: false
+BinPackParameters: true
+BinPackArguments: true
+AlignOperands: false
+
+# Alignment
+AlignEscapedNewlines: DontAlign
+AccessModifierOffset: -4
+AllowShortBlocksOnASingleLine: Always
+AllowShortIfStatementsOnASingleLine: Always
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: InlineOnly
+BreakBeforeBinaryOperators: All
+
+# Includes
+IncludeBlocks:   Regroup
+IncludeCategories:
+  # windows.h must go before everything else
+  # otherwise, you will get errors
+  - Regex:           '^<windows.h>$'
+    Priority:        -99
+  # the "main header" implicitly gets priority 0
+  # system headers
+  - Regex:           '^<[^>]+>$'
+    Priority:        1
+  # non-system headers
+  - Regex:           '.*'
+    Priority:        2
+SortIncludes:    true
+
+# Spaces
+SpaceBeforeAssignmentOperators: true
+SpaceBeforeParens: ControlStatements
+SpacesInAngles:  false
+SpacesInCStyleCastParentheses: false
+SpacesInParentheses: false
+
+# Brace wrapping
+# Not directly mentioned in the coding conventions,
+# but required to avoid tons of auto reformatting
+BreakBeforeBraces: Custom
+BraceWrapping:
+  AfterClass: true
+  AfterControlStatement: Always
+  AfterEnum: true
+  AfterFunction: true
+  AfterNamespace: false
+  AfterStruct: true
+  AfterUnion: true
+  AfterExternBlock: false
+  BeforeCatch: true
+  BeforeElse: true
+  IndentBraces: false
+  SplitEmptyFunction: true
+  SplitEmptyRecord: true
+  SplitEmptyNamespace: true
+  BeforeWhile: false
+  BeforeLambdaBody: false
+
+# Do not break doxygen comments
+CommentPragmas:  '^[[:space:]]*\\.+'
+
+# Pointers
+# Use pointer close to type: `const char* const* function()`
+PointerAlignment: Left
+
+...
+
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,33 @@
+---
+Checks:          'bugprone-macro-parentheses,bugprone-macro-repeated-side-effects,modernize-use-override,readability-identifier-naming,readability-misleading-indentation,readability-simplify-boolean-expr,readability-braces-around-statements'
+WarningsAsErrors: ''
+HeaderFilterRegex: '' # don't show errors from headers
+AnalyzeTemporaryDtors: false
+FormatStyle:     none
+User:            user
+CheckOptions:    
+  - key:             readability-identifier-naming.ClassCase
+    value:           CamelCase
+  - key:             readability-identifier-naming.EnumCase
+    value:           CamelCase
+  - key:             readability-identifier-naming.TypedefCase
+    value:           CamelCase
+  - key:             readability-identifier-naming.UnionCase
+    value:           CamelCase
+  - key:             readability-identifier-naming.StructCase
+    value:           CamelCase
+  - key:             readability-identifier-naming.UnionCase
+    value:           CamelCase
+# not yet working, as it currently applies both for static and object members
+# - key:             readability-identifier-naming.MemberPrefix
+#   value:           'm_'
+  # currently only working for local static variables:
+  - key:             readability-identifier-naming.StaticVariablePrefix
+    value:           's_'
+# not yet working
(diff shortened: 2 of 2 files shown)


JohannesLorenz:
I added a [github action](https://github.com/JohannesLorenz/lmms/actions) (see build number 15).

include/ActionGroup.h
```diff
@@ -33,15 +32,14 @@
 /// This class provides the same functionality as QActionGroup, but in addition
 /// has the actionTriggered(int) signal.
 /// It also sets every added action's checkable property to true.
-class ActionGroup : public QActionGroup
-{
+class ActionGroup : public QActionGroup {
```
M374LX, line 35: According to the coding conventions, the opening brace should be on its own line.

plugins/stk/mallets/mallets.cpp
```diff
@@ -487,7 +499,8 @@ void malletsInstrumentView::changePreset() {
 malletsSynth::malletsSynth(const StkFloat _pitch, const StkFloat _velocity, const StkFloat _control1,
 	const StkFloat _control2, const StkFloat _control4, const StkFloat _control8, const StkFloat _control11,
 	const int _control16, const uint8_t _delay, const sample_rate_t _sample_rate)
-	: m_presetIndex(0) {
+	: m_presetIndex(0)
+{
 	try {
```
M374LX, line 504: Again, the opening brace should be on its own line.

include/ActionGroup.h
JohannesLorenz: Thanks, I fixed it now.

plugins/stk/mallets/mallets.cpp
```diff
@@ -506,7 +519,9 @@ malletsSynth::malletsSynth(const StkFloat _pitch, const StkFloat _velocity, cons
 		m_voice->controlChange(128, 128.0f);
 
 		m_voice->noteOn(_pitch, _velocity);
-	} catch (...) { m_voice = nullptr; }
+	} catch (...) {
```
M374LX, line 522: Same here.

plugins/stk/mallets/mallets.cpp
JohannesLorenz: Fixed, too

plugins/stk/mallets/mallets.cpp
JohannesLorenz: You mean like

```
try
{
    // ...
}
catch
{
    // ...
}
```
?

And the same for if-else?

M374LX: Exactly.

JohannesLorenz: Should be fixed now.

include/BandLimitedWave.h
```diff
@@ -52,17 +52,17 @@ typedef struct
 public:
 	inline sample_t sampleAt(int table, int ph)
 	{
-		if (table % 2 == 0) {
-			return m_data[TLENS[table] + ph];
-		} else {
+		if (table % 2 == 0) { return m_data[TLENS[table] + ph]; }
```
M374LX, line 55: The ``if`` is on a single line, but the ``else`` is not. It looks inconsistent.

JohannesLorenz: Putting the else completely into one line [seems impossible](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) (possibly a bug or limitation to `AllowShortIfStatementsOnASingleLine`). As soon as I turn this `else` into an `if`, it works.

I can only change it to

```
                if (table % 2 == 0) { return m_data[TLENS[table] + ph]; }
                else {
                        return m_data3[TLENS[table] + ph];
                }
```

Is that better or worse?

M374LX: This looks better:
``
                if (table % 2 == 0)
                { 
                        return m_data[TLENS[table] + ph]; 
                }
                else 
                {
                        return m_data3[TLENS[table] + ph];
                }
``

A ternary if is even better:
``
                return (table % 2 == 0) ? m_data[TLENS[table] + ph] : m_data3[TLENS[table] + ph];
``

JohannesLorenz: I know, but the coding rules said `If the block can fit on one line, it is acceptable`. I think we must live with the bad else.

Using a ternary is nice, but clang-format can't do that.

include/BandLimitedWave.h
```diff
@@ -94,79 +89,72 @@ class LMMS_EXPORT BandLimitedWave
...
-		return 1.0f / pd;
-	}
+	static inline float pdToLen(float pd) { return 1.0f / pd; }
 
 	/*! \brief This method provides interpolated samples of bandlimited waveforms.
 	 *  \param _ph The phase of the sample.
-	 *  \param _wavelen The wavelength (length of one cycle, ie. the inverse of frequency) of the wanted oscillation, measured in sample frames
-	 *  \param _wave The wanted waveform. Options currently are saw, triangle, square and moog saw.
+	 *  \param _wavelen The wavelength (length of one cycle, ie. the inverse of frequency) of the wanted oscillation,
+	 * measured in sample frames \param _wave The wanted waveform. Options currently are saw, triangle, square and moog
```
JohannesLorenz, line 110: :vomiting_face:

Seeing this, I'm more and more against clang-format...

JohannesLorenz: Can be fixed by `CommentPragmas:  '^\\.+'`, but I need to revert all the code files again because now, the code is already broken.

src/gui/editors/PianoRoll.cpp
```diff
@@ -777,45 +666,34 @@ void PianoRoll::fitNoteLengths(bool fill)
...
 
-	TimePos bound = m_lenOfNewNotes;  // will be length of last note
+	TimePos bound = m_lenOfNewNotes; // will be length of last note
 	for (Note* note : notes)
 	{
-		if (constrainMax ? note->length() > bound : note->length() < bound)
-		{
-			note->setLength(bound);
-		}
+		if (constrainMax ? note->length() > bound : note->length() < bound) { note->setLength(bound); }
```
JohannesLorenz, line 688: It's in 1 line now and fits the 120 columns, so the tool does what we requested, but I'm not sure if it got more readable.

allejok96:
General impressions:
- Long lines are alright, they are still readable because the formatting is very consistent.
- Is it possible to prevent single line function definitions? See end of piano roll.
- Agreed that the if else inconsistency is annoying
- Long strings that have been split across multiple lines by not using comma seems to remain untouched
- Some switch statements are slightly obscured (more than usual) when a multi-line case is preceded by a one-line case, perhaps just a break. See piano roll paint event.
- Otherwise LGTM

JohannesLorenz:
@allejok96 Thanks for your review.

> * Is it possible to prevent single line function definitions? See end of piano roll.

Yes, I added `AllowShortFunctionsOnASingleLine: InlineOnly` to `.clang-format`. It fixes then mentioned definitions in `PianoRoll.cpp`.

> * Some switch statements are slightly obscured (more than usual) when a multi-line case is preceded by a one-line case, perhaps just a break. See piano roll paint event.

Obscured in what way? Can you please point out which exact line you mean, or paste a code snipped?

allejok96:
It might be very rare but in this case it feels like the break gets hidden (from PianoRoll's paint event)

```
switch (prKeyOrder[topNote])
{
    case PR_WHITE_KEY_SMALL:
    case PR_WHITE_KEY_BIG: break;
    case PR_BLACK_KEY:
        // draw extra white key
        drawKey(topKey + 1, grid_line_y - m_keyLineHeight);
}
```

JohannesLorenz:
> It might be very rare but in this case it feels like the break gets hidden

Indeed, I did not even see this when scrolling through it. Patching clang-format like

```patch
-AllowShortCaseLabelsOnASingleLine: true
+AllowShortCaseLabelsOnASingleLine: false
```

fixed it. Of course then, some `switch` statements get larger. Imagine one with 50 or 100 `case`s where every `case` only `return`s. On the other hand, the hidden `break` is really ugly and could be bug prone.

I'm rather for disabling one-line-cases, any opinions?

JohannesLorenz:
> any opinions?

@allejok96 , what do you think?

allejok96:
Glancing over the use of switch statements there are some places where you've got lot of things crammed into one line:
```cpp
case 1: w = (float)fabs(2.0f*(float)sin(fmod(0.5f*ph,TwoPi)))-1.f; break; //sine^2
```

Or you do this horrible whitespace thing:
```cpp
case Base::UserVST          : loc = ConfigManager::inst()->userVstDir(); break;
```

I'd rather see each statement on a separate line.

>  Imagine one with 50 or 100 cases

In theory this would be a problem, but currently the only such switch statement (>20 cases) in non-third-party code is in `PianoView.cpp`. And it will still be possible to do manual formatting, right?

So my opinion is :+1: @JohannesLorenz

JohannesLorenz:
Thanks @allejok96 !

PR #6331, opened 2022-03-15
Reload stylesheet when CSS file changes

DomClark:
This PR is mostly just to get a CI build of this feature for a user on Discord, but maybe it could be useful? Does what it says on the tin. Will close if this isn't considered desirable/sufficiently useful.


Spekular:
This is cool and IMHO definitely worth having, but for whatever reason when I change clip colors (in css) more than once they become gray. Newly created clips adopt the theme color, but any time the clip color is changed in css existing clips become gray again. The same doesn't happen for other elements as far as I can tell (tested mixer channel bg, mixer peak indicator, song editor background).

Rossmaxx:
A user on discord asked for a live theme editor and I got reminded about this PR existing (this PR is the closest thing to live theme editing). @DomClark is it possible to resolve the issue mentioned by spekular and bring this back? 

If you are busy, feel free to avoid it. I felt like bumping because I don't want to see such small PRs gathering dust while it was supposed to be merged easily.

DomClark:
> ... when I change clip colors (in css) more than once they become gray. Newly created clips adopt the theme color, but any time the clip color is changed in css existing clips become gray again.

I can't reproduce this, but I'm not sure whether it's due to something that's changed in master over the past two years, or whether I'm doing something wrong.

Rossmaxx:
@regulus79 this might be related to what you showed in the discord chat recently

regulus79:
This is really neat! Not having to restart a thousand times would absolutely help with theme development. It would be awesome to see more cool themes made by individuals, and I think this PR would really help with that, making it easier for people to get into theming. Especially if you want to change every single color in LMMS, that would take ages without this sort of feature.

messmerd (approved):
The code looks good to me. Haven't tested it though.

regulus79:
I have noticed that existing clip views do not automatically update. If you change their default colors in the css, save, and go back to lmms, they turn grey and sometimes get their name font size messed up. However, if you copy the clips, the new clips have the new correct colors and everything.

Also knob arc colors appear to not update immediately.

tresf:
There seems to be enough demand to make this a proper feature.

@regulus79 asked for a visual notification to appear indicating that a reload has occurred.  I'll add that as well as a small code comment about the `if (!m_styleReloader.files().contains(path))` portion.

tresf:
style css

bratpeki:
Asking out of curiosity, does this impact performance in any way?

tresf:
> Asking out of curiosity, does this impact performance in any way?

I posed the same question on Discord.  The general consensus was that it would be extremely negligible.  If people find this statement to be untrue, we can always toggle it off and offer a preference for it.

Issue #6330, opened 2022-03-13
Issues with track pointer movement w/ keyboard


YohananDiamond:
### Bug Summary

In the song editor, the track pointer movement with the arrow keys seems to not be working properly.
To make matters worse, sometimes when trying to move the track, the "dotted bar" on the left of the instrument tracks disappear, and it even offsets the track pointer.

![the dotted bar in question](https://user-images.githubusercontent.com/23298480/158044353-b6b820bb-c014-48d4-830f-13c85a6eedb4.png)

### Steps for issue 1 (limited track pointer movement)

1. Open the song editor
2. Position the track pointer relatively far (e.g. 5 bars) from the start
![image](https://user-images.githubusercontent.com/23298480/158044814-e4dbbab3-1d2e-4e7d-9a22-63fdbe9ab645.png)
3. Try to move with the arrow keys (in any direction) twice
For both sides, the first time works, but the second time doesn't and, if you change direction, it goes not back to the original place, but to the place it would be if you first pressed the opposite direction.
With `right`, or `right, left`:
![image](https://user-images.githubusercontent.com/23298480/158044838-6c1935ff-dd6b-4a07-9f86-f15cf271b72a.png)
With `left`, or `left, right`:
![image](https://user-images.githubusercontent.com/23298480/158044850-190097c5-3bb3-4a5c-8639-ce02ef0d49ed.png)
It seems that, instead of moving the pointer around, it's offsetting it by either 1 leap left or 1 leap right from the place it was positioned to start at. Playing it doesn't cause any pointer desync issues.

### Steps for issue 2 (the dotted bar disappearing)

1. Put song editor in full screen (max it out in the MDI)
![image](https://user-images.githubusercontent.com/23298480/158044536-78b8517c-9698-46e1-a9b9-8a5ace48e9c1.png)
![image](https://user-images.githubusercontent.com/23298480/158044528-ec6ac54f-bb2f-4255-85d1-d5022de2525a.png)
Note: don't re-position the track pointer or the bug stops, and keep the song editor window focused.
2. Press the right key on the keyboard
Before pressing:
![image](https://user-images.githubusercontent.com/23298480/158044999-9cc6efd4-b42d-443b-aeea-c69d94782ac9.png)
After pressing:
![image](https://user-images.githubusercontent.com/23298480/158044995-e179fc77-10af-4c0b-bc13-7d3c17655368.png)
[...]

PR #6338, opened 2022-03-22
Fix track handles dissapearing

### LINKED ISSUE #6334: Drag handle disappearance
Closest similar issue: https://github.com/LMMS/lmms/issues/6330 where it is mentioned the drag handle disappears sometimes.

### Bug Summary
Drag handle disappears sometimes, on some specific projects. One of the triggers, in some cases, is renaming a track.

#### Steps to reproduce
[1603bugcheck.zip](https://github.com/LMMS/lmms/files/8283308/1603bugcheck.zip)
Download the above zipped folder, and extract the file. It works only in 1.3 alpha.
1. Open the Beat/Bassline editor.
2. Rename any track there.

#### Expected behavior
Track gets renamed and track drag handles (dotted part of track for dragging) continue to stay visible

#### Actual behavior
Track gets renamed and track drag handles cease to exist

#### Screenshot
image
Video:

https://user-images.githubusercontent.com/76674645/158777729-2d0a67b1-77a5-406a-9b8c-488764a753d1.mp4

#### Affected LMMS versions

1.3 alpha

This bug is tricky to reproduce in general, and easier to reproduce with the file.

allejok96:
Fixes #6334, by letting Qt handle resizing of Song Editor's content.

I don't exactly understand what triggered the original bug, so testing is needed :smile:

Also this PR removes unnecessary **redrawing of ALL clips** when:

- Moving a track
- Adding or deleting a track
- Changing height of a track
- Changing height of Song Editor
- Opening/closing Song Editor

### DIFF
--- a/include/TrackContainerView.h
+++ b/include/TrackContainerView.h
@@ -168,8 +168,6 @@ public slots:
 protected:
 	static const int DEFAULT_PIXELS_PER_BAR = 16;
 
-	void resizeEvent( QResizeEvent * ) override;
-
 	TimePos m_currentPosition;
 
 
--- a/src/gui/editors/TrackContainerView.cpp
+++ b/src/gui/editors/TrackContainerView.cpp
@@ -89,11 +89,15 @@ TrackContainerView::TrackContainerView( TrackContainer * _tc ) :
 	m_tc->setHook( this );
 	//keeps the direction of the widget, undepended on the locale
 	setLayoutDirection( Qt::LeftToRight );
+
+	// The main layout - by default it only contains the scroll area,
+	// but SongEditor uses the layout to add a TimeLineWidget on top
 	auto layout = new QVBoxLayout(this);
 	layout->setContentsMargins(0, 0, 0, 0);
 	layout->setSpacing( 0 );
 	layout->addWidget( m_scrollArea );
 
+	// The widget that will contain all TrackViews
 	auto scrollContent = new QWidget;
 	m_scrollLayout = new QVBoxLayout( scrollContent );
 	m_scrollLayout->setContentsMargins(0, 0, 0, 0);
@@ -101,6 +105,7 @@ TrackContainerView::TrackContainerView( TrackContainer * _tc ) :
 	m_scrollLayout->setSizeConstraint( QLayout::SetMinAndMaxSize );
 
 	m_scrollArea->setWidget( scrollContent );
+	m_scrollArea->setWidgetResizable(true);
 
 	m_scrollArea->show();
 	m_rubberBand->hide();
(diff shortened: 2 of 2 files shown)


superpaik:
Regarding the bug it works ok.
Some weird behaviour. If the track is close to the end of the Song-Editor or Pattern-Editor, there is an unwanted scroll.
1603bug

Monospace-V:
> Regarding the bug it works ok. Some weird behaviour. If the track is close to the end of the Song-Editor or Pattern-Editor, there is an unwanted scroll. 1603bug

Interestingly, I can reproduce this even on the LMMS 1.3 before this fix, but not 1.2.0. Intended, or bug? You might need to open a new bug on the tracker for this one.

superpaik:
Ok.
New issue created #6343

Spekular (approved):
Neither the steps in #6334 nor #6330 reproduce the issue for me when testing this, code changes look sane.

allejok96:
I must investigate closer what happens when you update a TrackView and when it is needed. Perhaps renaming the method is stupid. At least there should be better comments.

allejok96:
I've reverted the rename since I've discovered how much cleaning up is needed for the `update()` methods. That's for another time.

Spekular (approved):

Spekular:
> I've reverted the rename since I've discovered how much cleaning up is needed for the `update()` methods. That's for another time.

Just to double check, this was the only change made in the force push, right?

allejok96:
Correct

allejok96:
Hey, I've reverted to the bare minimum needed to fix the bug. Ready to merge when tested @Veratil

I'll make another PR with the redraw related fixes because I have found more of them...

Veratil:
@Monospace-V @PhysSong can you also test that this fixes the issue and doesn't cause any regression of behavior by removing the resizeEvent function?

Rossmaxx:
Update about the fixes/any regressions? Is this ready for merge?

Rossmaxx:
Looking at the code, there doesn't seem to be an issue from removing `resizeEvent` overload, other than that there is a call to `realignTracks` (but why?). I'll hit merge soon after someone confirms.

M4rotte:
> Looking at the code, there doesn't seem to be an issue from removing `resizeEvent` overload, other than that there is a call to `realignTracks` (but why?). I'll hit merge soon after someone confirms.

I merged this commit on the very last lmms/master branch (bda1a9c37ed29cd7c80f08d7405dd4dbfad8947d) and noticed no problem either.

There is still the fact that having a sample track with a sample in it, makes the whole song editor very slow to response, it makes it stutters if I may say (the display, not the audio hopefully!). But I noticed that some time ago, it’s not aggravated by the commit anyhow.

I use sample tracks in the song editor nearly often myself, not even for this slowness, just a matter of taste I guess.

Rossmaxx (approved):

Issue #6189, opened 2021-10-06
Allow opening BAK files from sidebar


Monospace-V:
### Enhancement Summary
The sidebar has a section called "My projects" that lets you browse the projects directory of LMMS and open projects.
![image](https://user-images.githubusercontent.com/76674645/136145487-47deec6a-dda0-4c9d-a159-08ec88e51955.png)
Double clicking a file will open it.
However, LMMS can only open MMP (LMMS project) and MMPZ (compressed LMMS project) files. It cannot run BAK (backup) files.
LMMS should be able to open (mmp and mmpz) BAK files when opened from the sidebar. (Note that some other programs also use bak, so it will need to be ensured that the file in question is a .mmp.bak or a .mmpz.bak and not any other bak.)
(Not restricted to "My projects," also extends to "Home" and essentially any sidebar window that displays directories.)

#### Justification
Convenience of opening a BAK file as easily as opening an MMP file without having to use open dialogue, rename the file, or manually opening through file explorer.

RiedleroD:
maybe just adding an extra icon to an existing file to show that a backup exists would make the sidebar less cluttered by this proposition. The icon should be click- & draggable then.

allejok96:
Good idea! Or add "open backup version" in the context menu for mmp files, and for bak files without matching mmp just list them like normal?

> The icon should be click- & draggable then

Why would you drag a project instead of double clicking it? Idk maybe it's so simple to implement it doesn't matter

RiedleroD:
> Why would you drag a project?

I was talking about the instruments, but I now realize that the main topic was project files 😅 Still, I think it's important that backup-ed instruments can be dragged into view (usability and such)

Monospace-V:
Sometimes I rename BAK files to preserve them when modifying the mmp(z). 
The idea is good, but it would be still easier to be able to open BAK files from sidebar. A separate icon for bak files, perhaps, and a notice saying "This is a backup file" or something along those lines.
While an "Open backup version" would work, for my needs at least sidebar ability makes more sense.

Spekular:
Something like the "Show factory/user content" checkboxes in the sample browser would work here, no? Default to "Show projects" enabled and "Show backups" disabled and we keep an uncluttered view by default, and then if someone wants to view backups as well they check "Show backups".

Issue #5772, opened 2020-11-07
Browser - Sort Order


sound8:
From what I understand each OS has its own standards of how file lists naturally sort from top to bottom.
All (well I hope) file sorting approaches use A-Z but then other sorting approaches vary.
I am on a Windows machine so I have followed their methods of sorting files, and I have noticed a couple of conflicts between the Windows OS sorting approach and the LMMS Browser sorting approach.
Two main things I have noticed are....

1.) Character Prefixes That Make Files Appear At Top of List
Windows OS
characters such as "_"  (as well as others) make files appear at top
LMMS Browser
takes "_" items and puts them at the bottom

2.) Numbering Issues
Windows OS
uses the suffix " (#)" to number its files and sorts by natural counting
LMMS Browser
views this numbering system differently putting all the 1's togther for example

EXAMPLE:
Windows
> snare (1)
> snare (2)
> ....
> snare (9)
> snare (10)
> snare (11)

LMMS Browser
> snare (1)
> snare (10)
> snare (11)
> snare (2)
> snare (3)
> etc.

LMMS Browser confuses down for up, it can't count, what the hell kind of educational system was he exposed to as kid? :)

If others agree would it possible to set the LMMS Browsers coding to use the sorting methods employed by each specific OS it is being installed and used with? Or simply adding a setting to be able to "Use Windows File Sorting Methods"?

Hopefully something like this can be fixed soon, but.... probably not I would imagine.
It is just that all my files are setup this way so can get confusing at times, losing track of things.
Not sure if anyone else has run into any problems with this and would like a solution as well?

tresf:
1, 10, 11, 2 is a natural alphanumeric sorting system.  Windows XP introduced a feature where File Explorer was capable of using numerical sorting instead.  LMMS can leverage this too through Qt, with some modifications.

See also: https://stackoverflow.com/q/11933883/3196753

Note, LMMS is a volunteer led project, many features are implemented by the requestors.  If you're interested in getting involved with the source code, let us know over on Discord at https://lmms.io/chat.

abhishektiwari47:
hey i want to solve this problem.

zonkmachine:
> hey i want to solve this problem.

Have a go! :)

Monospace-V:
Adding to this, LMMS has A->Z then a->z. It would make more sense to have the two together, because people don't usually expect B and b in different places.

AbhiramL:
Can I help?

Coderayush13:
[...]

PR #6339, opened 2022-03-22
File browser improvements

allejok96:
image

- Proper sorting (numbers and special characters) #5772
- Check box for project file backups #6189
- Check box for hidden/unsupported files in the home/root tabs
- Expanded directories stays expanded when toggling the check boxes
- Searching reveals all matches when searching projects or samples
- Multi word search

Files are not longer split into two sections for user/factory, for this you may use the check boxes.

I have some more improvements related to saving backup files and general cleanup of the FileBrowser code but the diff got so messy I had to break it off.


superpaik:
In "Check box for hidden/unsupported files in the home/root tabs", when no option is selected some hidden files/folders are shown (see image)

Not errors, but some thoughts about this functionality.
It's called "My Projects", so by default, maybe "Factory content" should be unchecked. Or that the user selection is stored on config file. IMO Factory content is only interesting for new users.
When you select "Backups" you have to also click on "Refresh list" because if not backups files are not showed (I was expecting them to be shown). Maybe, when you change the selection a "Refresh list" is executed as well (It's like in the  hidden/unsupported files in the home/root tabs, you don't need to click on"Refresh" for them to show) . So "Refresh list" is only needed if new files are added while LMMS is opened.


Captura

allejok96:
Huh, it would be interesting to know what's the bottleneck in your case. The deleting and clearing feels weird.

With this PR this should be the case:
- Directories are loaded when expanded
- Directories for projects and samples are loaded recursively when searching
- Filtering by hidden/factory/unsupported etc should not cause any I/O since the file tree is loaded in memory
- Doing the same search twice or deleting a character should not cause I/O for the same reason

It used to be the case that the samples and projects folder was loaded recursively to memory when you open lmms.

allejok96:
@superpaik .xml is actually supported by LMMS in some cases. And on Windows, files beginning with dot aren't hidden (right?)

superpaik:
@allejok96 Yes, on Windows by default hidden files begin with dots (that is why I cercled .lmmsrc.xml)

allejok96:
No that's Unix behavior. [Windows does it using file attributes.](https://superuser.com/questions/364406/hiding-files-folders-which-begin-with-a-full-stop-period) You should be able to verify that in explorer.exe under Properties > Hidden.

On the other hand, lmmsrc.xml should probably be marked as Hidden when it is created, but that's another issue.

allejok96:
> it also makes searching a lot less responsive

@Spekular what OS are you on? I wanna do some profiling but I can only test for Linux.

Spekular:
> > it also makes searching a lot less responsive
> 
> @Spekular what OS are you on? I wanna do some profiling but I can only test for Linux.

Win 10

superpaik:
@allejok96 Sorry, my bad. I wasn't aware of that property. I've always thought that they were because of the dot.

allejok96:
@Spekular I've done some changes to speed up search. Now the only delay should be the first time you type a letter in the search box. If it works, I'll clean up the code before review, otherwise it's back to the drawing board.

Also note that LMMS with this patch probably starts faster because it doesn't load the whole the file tree during startup.

consolegrl:
@allejok96 Hey so I was trying to add a way to show/hide the .bak and other files with some filter buttons when I noticed the code was copy pasta that had lava flowed. Found this PR as well as the Sort Order issue, and so I'm here to help. Ready to pick this up if you don't have time or work with you on it, w/e works.

DomClark:
They wrote the following in [another PR of theirs](https://github.com/LMMS/lmms/pull/6700#issuecomment-1546989134):
> Fellow devs, I have decided to take a break from coding to pursue other things because it's been getting too much lately and apparently my life balancing skills sucks. I'll just leave the PRs open and if anyone feels like it they can pick it up from there. Sorry, I know it's abrupt but sometimes you have to be ❤️

If you don't get a response soon, feel free to pick it up.

PR #6340, opened 2022-03-22
Knife tool for Midi Clips

### LINKED ISSUE #3829: Ability to cut/join patterns in the sequencer
It would be very useful.

allejok96:
I've had this lying around for a while now, so I don't remember what it does. Let's test it.

Also moved the split function from `ClipView` to the `Clip` because it feels better.

Edit: I better say that it fixes #3829 while I remember it

### DIFF
--- a/include/AutomationClip.h
+++ b/include/AutomationClip.h
@@ -174,6 +174,8 @@ public:
 	bool isRecording() const { return m_isRecording; }
 	void setRecording( const bool b ) { m_isRecording = b; }
 
+	bool supportsStartTimeOffset() const override { return false; }
+
 	static int quantization() { return s_quantization; }
 	static void setQuantization(int q) { s_quantization = q; }
 
--- a/include/Clip.h
+++ b/include/Clip.h
@@ -93,22 +93,6 @@ public:
 		return m_length;
 	}
 
-	/*! \brief Specify whether or not a TCO automatically resizes.
-	 *
-	 *  If a TCO does automatically resize, it cannot be manually
-	 *  resized by clicking and dragging its edge.
-	 *
-	 */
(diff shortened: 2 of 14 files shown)


Spekular:
Since I'm not sure if I mentioned it before: I was thinking this would be best solved by allowing non-destructive clip resize from both ends for MIDI and automation clips, hopefully allowing for some deduplicated clip code and a unified split method. I don't oppose this being merged as a stop-gap if it passes reviews, but at least for now I won't review it.

allejok96:
That would be a cool feature. I could give it a try, although I it raises some questions like what if you edit a shrunken midi clip and make it longer - will it expand in the song editor or will it seem like your edit never happened?

This is a problem with automation atm (or is it?) You have no visual indication in the automation editor where the clip ends. The same goes for using piano roll to edit notes in a bb pattern.

Another solution could be to cut midi clips when you drag the left/right side and then you could simply delete the part you don't want. But I guess that wouldn't result in more unified code. It would just be a different workflow.

allejok96:
Ready for review... assuming it builds.

I made it more generalized and added some convenience functions to tell whether a clip type supports splitting/resizing/etc.

----


> allowing non-destructive clip resize from both ends for MIDI clips

I've been testing this but it's a can of worms :cold_sweat:  Something I realized during testing is that splitting has a different purpose than non-destructive resizing (at least for MIDI clip). Resizing is best in the case where you split and delete. But if you want to keep both parts as two separate shorter clips, **actual splitting** makes more sense.

> allowing for some deduplicated clip code and a unified split method

@Spekular now you have a common `Clip::split()` with a virtual `Clip::truncate()` to handle the dirty parts. It's not a perfect solution, but it's closer to your ambition.

PR #6345, opened 2022-03-26
Rename LedCheckbox to LedCheckBox in filenames

JohannesLorenz:
Justification: The class is named `LedCheckBox`, and other files follow
the casing of the classes, too.


JohannesLorenz:
Let's check if this builds. Reviews are not required.

PR #6347, opened 2022-03-27
Undoable add/remove bar

allejok96:
This is a temporary fix for something that would require big changes to the undo system.

### DIFF
--- a/src/core/Song.cpp
+++ b/src/core/Song.cpp
@@ -778,7 +778,7 @@ void Song::insertBar()
 	for (Track* track: tracks())
 	{
 		// FIXME journal batch of tracks instead of each track individually
-		track->addJournalCheckPoint();
+		if (track->numOfClips() > 0) { track->addJournalCheckPoint(); }
 		track->insertBar(m_playPos[Mode_PlaySong]);
 	}
 	m_tracksMutex.unlock();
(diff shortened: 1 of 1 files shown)


allejok96:
To anyone testing this: this will create an undo checkpoint for *every track* (I guess even some hidden tracks). Copying multiple clips and changing track colors work the same way right now. And it is due to a limitation in our undo system.

src/core/Song.cpp
```diff
@@ -775,10 +775,11 @@ void Song::stopExport()
 void Song::insertBar()
 {
 	m_tracksMutex.lockForRead();
-	for( TrackList::const_iterator it = tracks().begin();
-					it != tracks().end(); ++it )
+	for (Track* track: tracks())
 	{
-		( *it )->insertBar( m_playPos[Mode_PlaySong] );
+		// FIXME journal batch of tracks instead of each track individually
+		track->addJournalCheckPoint();
```
Spekular, line 781:
```suggestion
		if (track->numOfClips() > 0){ track->addJournalCheckPoint(); }
```
And same for removal

Spekular (commented):
I think it'd be nice to exclude empty tracks, so that every ctrl+z has a visible effect.

src/core/Song.cpp
allejok96: Good idea. Added.

Still this is no guarantee that all undo will have a visible effect here. Instead of checking for that we should implement batching of checkpoints...

Spekular (approved):

DomClark (approved):

PR #6349, opened 2022-03-27
Fix QRC dependency issues

### LINKED ISSUE #6177: Qrc dependency issues
### Bug Summary

There are two different issues here, but both are dependency problems with our `GenQrc` module. One is an extra dependency, causing a dependency cycle, and the other is a missing dependency on the qrc inputs.

#### Steps to reproduce

To reproduce the dependency cycle:
* Run CMake using the Ninja (or Ninja Multi-Config) generator.
* Build LMMS.
* Re-run CMake.
* Trigger an automoc re-run for the `lmmsobjs` target - modifications to classes deriving from `QObject` are how this would happen naturally, but you can get the same effect by deleting `build/src/lmmsobjs_autogen/mocs_compilation.cpp` (or a config-specific version if you're using Ninja Multi-Config).
* Build LMMS again.

To reproduce the missing dependency on QRC inputs:
* Run CMake in a fresh build directory. (`build/src/lmms.qrc` must not be left over from a previous build.)
* Build LMMS.
* Modify `LICENSE.txt` or `doc/AUTHORS`.
* Rebuild LMMS (do not re-run CMake).
* Run LMMS and open the about dialog.

#### Expected behavior

Dependency cycle: Successful compilation.
Missing dependency: Updated licence and author list should show in the about dialog.

#### Actual behavior

Dependency cycle: Ninja complains about a dependency cycle.
Missing dependency: About dialog shows original, unmodified licence and author list.

#### Logs
<pre>
ninja: error: dependency cycle: src/lmmsobjs_autogen/timestamp -> src/lmms.qrc -> src/lmmsobjs_autogen:Debug -> src/CMakeFiles/lmmsobjs_autogen-Debug -> src/lmmsobjs_autogen/timestamp
</pre>

#### Further details

Dependency cycle:

The qrc files shouldn't have anything to do with moc, so there are two undesirable dependencies here - `src/lmms.qrc` should not depend on `src/lmmsobjs_autogen`, and `src/lmmsobjs_autogen/timestamp` should not depend on `src/lmms.qrc`. These arise through different routes:
* The `qrc -> autogen` dependency is due to how the Ninja generator associates custom commands with targets. Each custom command is associated with all targets that consume its output (perhaps transitively via other custom commands, as is the case here with the rcc output file `qrc_lmms.cpp`, which does not show up in the cycle). The dependencies for the custom command [are computed](https://github.com/Kitware/CMake/blob/v3.21.3/Source/cmLocalNinjaGenerator.cxx#L617-L634) by taking the intersection of all associated targets' dependencies, under the assumption that any true dependency for the command will be present in all associated targets. `lmms.qrc` is only associated with `lmmsobjs`, thus `lmmsobjs_autogen`, which is a dependency of `lmmsobjs`, is added as a dependency for `lmms.qrc`.
* The `autogen -> qrc` dependency is due to the `qt5_add_resources` function, which behaves differently depending on whether its input files exist at the time it is called. This is why the cycle only arises after building and re-running CMake - it means that the qrc file exists during the second CMake run when calling the function. If the file exists, Qt [uses it as a dummy config file](https://github.com/qt/qtbase/blob/5.15.2/src/corelib/Qt5CoreMacros.cmake#L253-L257) to force CMake to re-run if it changes, in order to recompute dependencies. Separately, CMake adds all CMake source files involved in generating automoc input as dependencies to the autogen target, to force moc to re-run if its inputs may have changed. Unfortunately, these two features share the same list of files internally, which means the autogen target depends on the qrc file.

Missing dependency:

The rcc output file `build/src/qrc_lmms.cpp` should have a dependency on each file incorporated into it. However, Qt is unable to add these dependencies if the qrc file does not exist at configure time, and we don't attempt to add these dependencies manually.

#### Is this our bug?

The missing dependency is, but the dependency cycle I don't think so. CMake's Ninja generator adds an unnecessary dependency on automoc due to a quirk of how it handles custom commands. Also, I think Qt should check not only that the qrc file exists, but that no rule exists to generate it, before it reads dependencies from it, or uses it as a dependency for CMake to re-run. It doesn't make sense to check at configuration time the contents of a file that can change at build time, nor for a CMake re-run to depend on changes to a file that CMake is already aware of the dependencies of. Regardless, we have to work around it.

#### Possible solutions:

* Write the qrc file at configure time. This is probably the simplest option. The qrc file is no longer generated at build time, so will not depend on anything and thus cannot cause a dependency cycle. Qt is able to read the qrc file in `qt5_add_resources`, so the rcc output file will correctly depend on its inputs. This has the disadvantage that we can't use generator expressions for the content, but we don't seem to do that at the moment anyway.
* Don't call `qt5_add_resources` - add the rcc command ourselves. This is more flexible and conceptually cleaner than the first option, but is a little more complex. We retain the ability to use generator expressions, as we can create the qrc file whenever we choose, be that at generation time or at build time. While we need to add the rcc input files manually as dependencies to rcc, it means that we can add them directly, instead of writing them to a file only to have Qt parse them again.
* Ensure that the qrc file does not exist during the call to `qt5_add_resources`. This requires fewer changes to our existing code than the previous solution - we still have to add the rcc dependencies ourselves, but can just delete the qrc file before we call Qt instead of dealing with rcc. However, this would needlessly recreate the file, causing unnecessary recompilations.

Additionally, we could eliminate the qrc/rcc dependency on automoc by creating custom targets for the generation commands, so they don't pick up dependencies from the target their output is used in. However, this is not necessary if we eliminate the other dependency using one of the above solutions.

Opinions?

DomClark:
Fixes #6177.

We now add the `rcc` command manually, rather than using `qt5_add_resources`, and make sure to include all the referenced files in the list of dependencies. See the associated issue for further information.

### DIFF
--- a/cmake/modules/GenQrc.cmake
+++ b/cmake/modules/GenQrc.cmake
@@ -1,48 +1,73 @@
 # GenQrc.cmake - Copyright (c) 2015 Lukas W <lukaswhl/at/gmail.com>
 
-# Generates a simple qrc file containing the given resource files ${ARGN}:
-#     GEN_QRC(resources.qrc artwork.png icon.png PREFIX /icons)
-# Files may also be added using a pattern with the GLOB keyword, e.g.:
-#     GEN_QRC(resources.qrc GLOB *.png)
-FUNCTION(GEN_QRC OUT_FILE)
-	CMAKE_PARSE_ARGUMENTS(RC "" "PREFIX;GLOB" "" ${ARGN})
-
-	IF(DEFINED RC_GLOB)
-		FILE(GLOB GLOB_FILES ${RC_GLOB})
-	ENDIF()
-
-	# Set the standard prefix to "/" if none is given
-	IF(NOT DEFINED RC_PREFIX)
-		SET(RC_PREFIX "/")
-	ENDIF()
-
-	# We need to convert our list to a string in order to pass it to the script
-	# on the command line.
-	STRING(REPLACE ";" "\;" FILES "${RC_UNPARSED_ARGUMENTS};${GLOB_FILES}")
-
-	SET(GENQRC_SCRIPT "${CMAKE_SOURCE_DIR}/cmake/scripts/GenQrc.cmake")
-	ADD_CUSTOM_COMMAND(
-		OUTPUT ${OUT_FILE}
-		COMMAND ${CMAKE_COMMAND} -D OUT_FILE=${OUT_FILE} -D RC_PREFIX=${RC_PREFIX} -D FILES:list=${FILES} -D DIR=${CMAKE_CURRENT_SOURCE_DIR} -P "${GENQRC_SCRIPT}"
-		DEPENDS ${GENQRC_SCRIPT}
-		WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
-		VERBATIM
-	)
-ENDFUNCTION()
-
-# Generates a qrc file named ${QRC_OUT} from ${ARGN}, rccs it and returns Qt's
-# output file.
+# Generates a simple qrc file named ${QRC_NAME} containing the given resource
+# files ${ARGN}, rccs it, and returns Qt's output file.
 # Must only be run once per CMakeLists.txt.
 # Usage example:
-#     ADD_GEN_QRC(RCC_OUTPUT resources.qrc icon.png manual.pdf)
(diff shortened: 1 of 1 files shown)


PhysSong (approved):
Looks good.
I think it's good to have some comments in the module mentioning why we don't simply use `qt5_add_resources` or `AUTOQRC`.

PR #6358, opened 2022-04-01
Fix SF2 Player segfault when releasing C-1 notes

xTibor:
The SF2 Player had an off-by-one error when checking the range of MIDI note values. This caused an early return and missing initialization of `NotePlayHandle->m_pluginData` fields when playing C-1 notes, leading to a subsequent segmentation fault when releasing the note.

### DIFF
--- a/plugins/sf2_player/sf2_player.cpp
+++ b/plugins/sf2_player/sf2_player.cpp
@@ -631,7 +631,7 @@ void sf2Instrument::playNote( NotePlayHandle * _n, sampleFrame * )
 		int midiNote = (int)floor( 12.0 * ( log2( _n->unpitchedFrequency() ) - LOG440 ) - 4.0 );
 
 		// out of range?
-		if( midiNote <= 0 || midiNote >= 128 )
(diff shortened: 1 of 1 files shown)


DomClark (commented):
This is certainly an improvement, but it's even possible to make `midiNote` negative (or greater than 128) using note detuning automation, which leads to the same segfault. If that early return is ever taken, then the segfault will occur upon note-off, so there is some faulty logic to be fixed here beyond correcting the range check. Either the early return code should be moved earlier in the method, so both note-on and note-off are ignored, or a null check should be added before the note-off code.

Of course, the off-by-one fix should remain, since the full MIDI range should be available, but I would consider it to be a separate problem to the segfault.

xTibor:
Oh, it's more busted than I thought. Then I have no idea how to fix it properly.

RiedleroD:
just a little something I found out: this only happens if the note is played by pressing a 'play' button. It doesn't happen in the instrument window, nor when the note is played while being placed or moved. Not sure if that was obvious from the code, I just thought that might be some help in determining the cause of the bug.

PR #6362, opened 2022-04-04
Move editor and dialog classes to appropriate folders

Spekular:
Targets two of the remaining items under `Reorganise all files into better places & folders` in #5592

Current state:
- Moved into `gui/editors`: ControllerRackView, MixerView, ProjectNotes
- Moved into `gui/dialogs`: AboutDialog, ControllerConnectionDialog, ExportProjectDialog, RenameDialog, SetupDialog, EffectSelectDialog.ui
- Renamed (formerly `XYZDialog`): ControllerWindow, LfoControllerWindow, PeakControllerWindow

Current problems:
- MicrotunerConfig: Does this belong together with the editors?
- EffectControlDialog: Does not open in external window. However, all our effect windows inherit from this, so renaming it would be rather invasive
- Is "Dialog" actually inaccurate for the renamed classes? Would it be more accurate to rename the `dialogs` folder to `modals` instead?


allejok96:
I slightly disagree with the editors move. The motivation seems to come from the placement of those buttons in the toolbar. With that logic, microtuner config should go into editors too.

IMO the term "editor" doesn't fit Mixer or Controller Rack. And adding to that there is actually an `Editor.cpp` which should probably go into that folder, no? And it defines an editor as a window with a toolbar and playback controls.

Spekular:
I think the song editor, mixer, controller rack, and project notes all belong together, if they're going to be categorized and not just live in `gui`. Their content is global in the context of a project, and they have no parents in the UI. The microtuner fits those criteria too (I think?) but since it wasn't mentioned in the previous discussion I chose to not touch it in the initial commit. BB Editor is a bit of an edge case since it can be spawned from the song editor, but it does directly expose one of the song's two track containers.

Where exactly those classes go or what to call them I'm not sure of, now that you point out the "has a play button" criteria. I thought `editor` made sense despite the piano roll and automation editor being there, because we still only have one of each and they're evidently related enough to be next to each other in the toolbar.

src/gui/CMakeLists.txt
```diff
@@ -47,14 +42,21 @@ SET(LMMS_SRCS
 	gui/ClipView.cpp
 	gui/TrackView.cpp
 
+	gui/dialogs/AboutDialog.cpp
+	gui/dialogs/ColorChooser.cpp
```
Veratil, line 46: I don't see a move for this.

src/gui/CMakeLists.txt
```diff
@@ -124,8 +124,7 @@ set(LMMS_UIS
 	${LMMS_UIS}
 	gui/dialogs/about_dialog.ui
 	gui/dialogs/export_project.ui
-
-	gui/Forms/EffectSelectDialog.ui
+	gui/dialogs/EffectSelectDialog.ui
```
Veratil, line 127: I don't see a move for this.

src/gui/CMakeLists.txt
Spekular: Opens in external window as mentioned on Discord

Spekular (commented):
Wait... this is ugly. This is a bigger can of worms than I thought.
```
virtual ControllerWindow * createDialog( QWidget * _parent );
```

Spekular:
Closed in favor of #6367

PR #6363, opened 2022-04-04
Sample caching

JGHFunRun:
The code is very basic at the moment. For now I'm just getting this up so that it's visible and possibly someone can contribute code

TODOs:

- [ ] ~~Add a method to generate new audio file names. This is for recording and audio editing mainly, when they are incorporated. It's also why `SampleCache` is it's own class and not just a `typedef`~~
  - ~~My idea for this is to have a counter that counts the number of audio files made for the current project file and then have a user settable format string for it. The output format would also be selectable in settings. My idea for the default for this string is `{project_path}.dir/{n}.{format}`. Wanted feedback on how to do it before implementing it~~
  - I have been recommended by @DomClark to make this a separate thing and will be taking that advice
- [ ] Convert everything to using it
  - [ ] Sample tracks
  - [ ] AFP
  - [ ] 3xOsc's user wave
  - [ ] SF2Player
  - [ ] GIG Player
  - [ ] PatMan
  - [ ] TODO: find other plugins that should use the sample cache

It may be just best to get the ~~first bullet point,~~ AFP, and sample tracks done so that new plugins and features that need it can use it immediately instead of waiting to convert all of the plugins or need to be converted retroactively. ~~Possibly even to~~ Wait on the first bullet point since it's not needed until sample recording or destructive sample editing is implemented

### DIFF
--- a/include/SampleCache.h
+++ b/include/SampleCache.h
@@ -1,0 +1,37 @@
+/*
+ * SampleCache.h - A class for the sample cache
+ *
+ * Copyright (C) 2022 JGHFunRun <JGHFunRun@gmail.com>
+ *
+ * This file is part of LMMS - https://lmms.io
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this program (see COPYING); if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA 02110-1301 USA.
+ *
+ */
+
+#ifndef SAMPLE_CACHE_H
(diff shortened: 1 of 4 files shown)


JGHFunRun:
Rework to the class structure on the way

include/SampleCache.h
```diff
@@ -0,0 +1,37 @@
...
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this program (see COPYING); if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA 02110-1301 USA.
+ *
+ */
+
+#ifndef SAMPLE_CACHE_H
```
roshanrules1, line 25: perhaps use CamelCase here for better readability

include/SampleCache.h
```diff
@@ -0,0 +1,37 @@
...
+#define SAMPLE_CACHE_H
```
roshanrules1, line 26: perhaps you can use CamelCase instead of all caps for better readability

Spekular: That would be inconsistent with the rest of the codebase.

sakertooth:
I have made some contributions to sample caching [https://github.com/JGHFunRun/lmms/pull/1](here). It begins with a separate rewrite of ``SampleBuffer`` in ``SampleBufferV2``, which can be added back into the codebase once all the necessary changes are made. I also attempted to implemented the ``SampleCache``. This is of course mostly a draft so nothing is really finalized.

Rossmaxx:
Closing in favour of #7058

PR #6366, opened 2022-04-07
Kill orphaned subprocesses on crash

### LINKED ISSUE #4408: (Suggestion) Background Process for Checking Crashes / Other LMMS Background Processes
I am certain the topic of how to deal with VST processes after an LMMS crash through programming has gone through the motions in this GitHub repository before. Even if not, I'd like to bring it to the developers' attention with a potential solution of my own.

The idea is to create a much smaller, background process (maybe a small .dll on Windows, or something very similar, either way it should run as a process) contained within the LMMS directory started up by LMMS.exe (with a check in LMMS to see if the process is running, if so, the process will not start up again). Once this process is started, it should check the running processes to see if LMMS.exe is running first.

**If LMMS (any instance of LMMS.exe) is running:**
A value should be set to true to ignore the checks on things like VST processes, etc. that are being used
The program should continue the loop of checking LMMS, preferably on an interval between 5 to 20 seconds so it's not hogging RAM/CPU but also not allowing stray VSTs to hog RAM/CPU either

_BUT_ **if LMMS is NOT running:**
This value should be set to false immediately and all pertinent processes should be shut down before the process itself shuts itself down.

There's two reasons I say this should be an external process:
1. If some sort of fatal error occurs that cannot be handled without extremely convoluted code, LMMS will crash, likely to desktop. In this scenario it makes sense to keep this suggested process external, so whatever is going on with LMMS **does not affect the process itself.** It also saves a lot of room within LMMS' code.
2. A smaller process will take up relatively little RAM/CPU compared to a bigger program like LMMS, and if it is separate from the instances of LMMS the RAM/CPU usage remains constant rather than increasing per instance (since only one instance is allowed per the suggestion).

This is a fairly simple solution to the problem, considering the nature of the problem and the scope of potential scenarios which the solution covers. It is also a preemptive strike against possible crash-errors in the future of LMMS' development that may not be able to be handled properly, leading to stray VSTs. It may seem like overkill, and ultimately is a suggestion, but I urge the developers to look into this as a potential solve.

Thank you for your time.

EDIT: I didn't do such a great job implying this, so I'll just clarify: when the process shuts down the VSTs, it should be telling either the VST processes OR the OS to close the processes down. The latter is preferable if VST processes have a chance of becoming unresponsive after LMMS crashes.

DomClark:
Fixes #4408. Uses the techniques in the comments there, with a few tweaks:
* The Windows code no longer uses `QProcess::pid()`, since it is now deprecated. Replaced with `QProcess::processId()` and `OpenProcess()`.
* The compile guards for the POSIX code now check for "not Windows", rather than just "Apple or Linux", since LMMS is used on a number of other POSIX-compliant systems.
* The POSIX code now uses RAII to cleanly handle the multiple exit points from the remote plugins' `main()` methods.
* The POSIX code now uses a timed wait on a condition variable, so normal exit can happen immediately rather than waiting up to 500ms for each process.

I forgot to add Lukas-W as a co-author on the POSIX commit - will try to remember when I merge this.

### DIFF
--- a/include/RemotePluginClient.h
+++ b/include/RemotePluginClient.h
@@ -135,6 +135,7 @@ private:
 	fpp_t m_bufferSize;
 } ;
 
+#ifndef LMMS_BUILD_WIN32
 class PollParentThread
 {
 public:
@@ -171,6 +172,7 @@ private:
 	std::condition_variable m_cv;
 	std::thread m_thread;
 };
+#endif
 
 #ifdef SYNC_WITH_SHM_FIFO
 RemotePluginClient::RemotePluginClient( key_t _shm_in, key_t _shm_out ) :
--- a/include/RemotePluginClient.h
+++ b/include/RemotePluginClient.h
@@ -32,7 +32,7 @@
 #	include <mutex>
 #	include <thread>
 
-#	include <sys/signal.h>
+#	include <signal.h>
 #	include <unistd.h>
 #endif
 
--- a/include/RemotePluginClient.h
+++ b/include/RemotePluginClient.h
@@ -27,6 +27,15 @@
 
 #include "RemotePluginBase.h"
 
+#ifndef LMMS_BUILD_WIN32
+#	include <condition_variable>
+#	include <mutex>
+#	include <thread>
+
+#	include <sys/signal.h>
+#	include <unistd.h>
+#endif
+
 class RemotePluginClient : public RemotePluginBase
 {
 public:
@@ -125,6 +134,43 @@ private:
 	sample_rate_t m_sampleRate;
 	fpp_t m_bufferSize;
 } ;
+
+class PollParentThread
+{
+public:
+	PollParentThread() :
+		m_stop{false},
+		m_thread{
+			[this] {
+				auto lock = std::unique_lock{m_mutex};
+				while (!m_cv.wait_for(lock, std::chrono::milliseconds(500), [this] { return m_stop; }))
+				{
+					if (getppid() == 1)
+					{
+						kill(getpid(), SIGHUP);
+						break;
+					}
+				}
+			}
+		}
+	{ }
+
+	~PollParentThread()
+	{
+		{
+			const auto lock = std::unique_lock{m_mutex};
+			m_stop = true;
+		}
+		m_cv.notify_all();
+		m_thread.join();
+	}
+
+private:
+	bool m_stop;
+	std::mutex m_mutex;
+	std::condition_variable m_cv;
+	std::thread m_thread;
+};
 
 #ifdef SYNC_WITH_SHM_FIFO
 RemotePluginClient::RemotePluginClient( key_t _shm_in, key_t _shm_out ) :
--- a/include/RemotePluginClient.h
+++ b/include/RemotePluginClient.h
@@ -142,9 +142,11 @@ public:
 	PollParentThread() :
 		m_stop{false},
 		m_thread{
-			[this] {
+			[this]
+			{
+				using namespace std::literals::chrono_literals;
 				auto lock = std::unique_lock{m_mutex};
-				while (!m_cv.wait_for(lock, std::chrono::milliseconds(500), [this] { return m_stop; }))
+				while (!m_cv.wait_for(lock, 500ms, [this] { return m_stop; }))
 				{
 					if (getppid() == 1)
 					{
--- a/plugins/VstBase/RemoteVstPlugin/CMakeLists.txt
+++ b/plugins/VstBase/RemoteVstPlugin/CMakeLists.txt
@@ -35,9 +35,9 @@ add_executable(${EXE_NAME} WIN32
 	${LMMS_COMMON_SRCS}
 )
 if(IS_WIN)
-    target_link_libraries(${EXE_NAME} ole32)
+	target_link_libraries(${EXE_NAME} ole32)
 else()
-    target_link_libraries(${EXE_NAME} dl X11)
+	target_link_libraries(${EXE_NAME} dl X11)
 endif()
 if(NOT WIN32)
 	target_link_libraries(${EXE_NAME} pthread)
--- a/plugins/VstBase/RemoteVstPlugin/CMakeLists.txt
+++ b/plugins/VstBase/RemoteVstPlugin/CMakeLists.txt
@@ -37,8 +37,11 @@ add_executable(${EXE_NAME} WIN32
 if(IS_WIN)
     target_link_libraries(${EXE_NAME} ole32)
 else()
-    target_link_libraries(${EXE_NAME} pthread dl X11)
-ENDIF()
+    target_link_libraries(${EXE_NAME} dl X11)
+endif()
+if(NOT WIN32)
+	target_link_libraries(${EXE_NAME} pthread)
+endif()
 
 target_include_directories(${EXE_NAME}
 	PRIVATE 
--- a/plugins/VstBase/RemoteVstPlugin.cpp
+++ b/plugins/VstBase/RemoteVstPlugin.cpp
@@ -2475,7 +2475,7 @@ int main( int _argc, char * * _argv )
 	}
 
 #ifndef LMMS_BUILD_WIN32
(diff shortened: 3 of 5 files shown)


Spekular:
Not sure if this needs testing or not, but works for me both when hanging/freezing and when crashing. (Win 10, Serum 64)

plugins/VstBase/RemoteVstPlugin/CMakeLists.txt
```diff
@@ -37,8 +37,11 @@ add_executable(${EXE_NAME} WIN32
 if(IS_WIN)
     target_link_libraries(${EXE_NAME} ole32)
 else()
-    target_link_libraries(${EXE_NAME} pthread dl X11)
-ENDIF()
+    target_link_libraries(${EXE_NAME} dl X11)
+endif()
+if(NOT WIN32)
+	target_link_libraries(${EXE_NAME} pthread)
```
Veratil, line 43: spaces or tabs? The eternal conflict. 😁

Veratil (commented):
Other than two long lines (>100 characters) which we can either ignore or split up function arguments... we must answer the debate here and now. Spaces or tabs? 😂

plugins/VstBase/RemoteVstPlugin/CMakeLists.txt
DomClark: The majority of the file uses tabs, but there are a couple of lines that use spaces. I left the existing indentation alone, and used tabs for my additional code. I can fix it up so it's all tabs if you'd like.

Another thing to discuss is casing - all the commands visible in the diff are lowercase, but there are also a lot of uppercase commands in our CMake code. I prefer lowercase as I think it's easier to read, and it's also the recommended style now.

Veratil: Yeah I haven't touched anything in CMake styling yet. I think tabs are probably the better way to go here since they're indentations rather than alignment.
As for casing, CMake casing is all over the place. Just follow their recommendations I think.

DeathSuperMario:
Switched from ubuntu to linux lite. I have 7.5GB ram. Ubuntu showed 7GB ram and linux lite shows 8gb ram. 
In Linux lite lmms don't crash. Still shows high cpu use. Still vst host crashing after I close lmms. Going to try out kill bot.

DomClark:
@DeathSuperMario Do you have VST host processes left running after LMMS has closed (crashed or otherwise)? That's the bug this change is intended to fix.

DeathSuperMario:
> 

In ubuntu VST host was running even when I turned off LMMS. In linux lite LMMS is working better but .still VST host crashing after I close LMMS.

DeathSuperMario:
I can't run LMMS appimage what is uploaded in this thread.  Right now I can't test out kill bot.

DeathSuperMario:
[DomClark](https://github.com/DomClark) Can you specify what you mean and saying?

DomClark:
@DeathSuperMario This pull request thread is specifically for discussing the associated changes, and isn't the right place for support with running the alpha or AppImage. The best idea is to join our Discord server at https://lmms.io/chat and ask in the `#support` channel.

DeathSuperMario:
> @DeathSuperMario This pull request thread is specifically for discussing the associated changes, and isn't the right place for support with running the alpha or AppImage. The best idea is to join our Discord server at https://lmms.io/chat and ask in the `#support` channel.

If I can't run lmms alpha 1.3 I I can't test killing orphaned subprocess on crash. 😉

include/RemotePluginClient.h
```diff
@@ -126,6 +135,47 @@ class RemotePluginClient : public RemotePluginBase
...
+	PollParentThread() :
+		m_stop{false},
+		m_thread{
+			[this]
+			{
+				using namespace std::literals::chrono_literals;
+				auto lock = std::unique_lock{m_mutex};
+				while (!m_cv.wait_for(lock, 500ms, [this] { return m_stop; }))
+				{
+					if (getppid() == 1)
```
Veratil, line 151: I can't really find any documentation on how `getppid` is supposed to work on Windows. Maybe a comment on why this should `== 1`?

PhysSong: Because this code isn't for Windows.

Veratil: ah crap, I always miss that stupid `n` in the `ifndef`. Ignore me. 😩

PhysSong:
I found that this PR prevents loading 64bit Windows VSTs on Linux(tested with Wine 7.6). Running with `WINELOADER=/usr/bin/wine64` fixes the issues with 64bit VSTs but breaks 32bit ones.

`winegcc` generates a wrapper script for launching a `.exe.so` file, and this part of the script launches Wine:
```shell
# determine the WINELOADER
if [ ! -x "$WINELOADER" ]; then WINELOADER="wine"; fi

# and try to start the app
exec "$WINELOADER" "$apppath" "$@"
```

`wine` can't load 64bit binaries directly, because it's a 32bit executable. In this case, `wine` launches `wine64` and then exits, making `getppid()` calls return 1.

DomClark:
Thanks for investigating! I didn't test with 64-bit Windows VSTs on Linux because I had trouble loading them prior to this PR, and naively thought that testing with a 32-bit Windows VST should be sufficient.

Would you be able to work on this? I don't have the Linux experience to easily solve this myself (indeed, I used Lukas-W's code for the current solution), and can't test any fix easily due to the aforementioned problem. Or we can disable it only for the Wine VST hosts, which would still be an improvement from before this PR.

PhysSong:
There are several possible approaches:
- setting `WINELOADER` according to VST architecture before loading: This can ensure the `getppid()` approach working in most cases, but that will likely to conflict with the manually set value of `WINELOADER`
- Disable this approach if the first call to `getppid()` returns 1: It's slightly better than disabling for all Wine VST hosts, but have to make sure the LMMS process is alive on the first call to `getppid()`
- Add a new process for detecting crash: That would make stuff much more complicated, I think

@lukas-w Do you have any ideas and/or opinions?

lukas-w:
Haven't thought these through entirely, but here are some ideas:

* Support reading a `WINELOADER64` variable to still allow manually setting a loader in case we go with the `WINELOADER` fix
* Supply the child VST process with the ID of the main LMMS process so it can look up its state by ID, not by parent relationship
* If I understand the flag correctly, we can use [`prctl(PR_SET_CHILD_SUBREAPER)`](https://linux.die.net/man/2/prctl) in the main process to have `wine64` automatically re-parent to `lmms` instead of to `init` (PID 1) when `wine` dies. Probably the easiest fix by far, but only available on Linux.

PhysSong:
> Supply the child VST process with the ID of the main LMMS process so it can look up its state by ID, not by parent relationship

How are PID-based methods reliable, and which one do you recommend?


> we can use [`prctl(PR_SET_CHILD_SUBREAPER)`](https://linux.die.net/man/2/prctl) in the main process to have `wine64` automatically re-parent to `lmms` instead of to `init` (PID 1) when `wine` dies.

That might work, but what if the `prctl(PR_SET_CHILD_SUBREAPER, 1)` call fails?

lukas-w:
> How are PID-based methods reliable, and which one do you recommend?

I didn't claim they were more reliable, I don't know really. You asked for ideas, not an essay :smile: 

> That might work, but what if the `prctl(PR_SET_CHILD_SUBREAPER, 1)` call fails?

I'm not sure this call can ever fail as long as using Linux >= 3.4. Maybe we should use this flag regardless of whether we employ an additional mechanism for this particular issue. It makes sense to me that grand-children processes are re-parented to `lmms` when their parent dies, instead of being orphaned and daemonized.

This really looks like a wine bug though, doesn't it? `wine` should exec `wine64` instead of creating and orphaning a child IMO. Anyway, `PR_SET_CHILD_SUBREAPER` should be enough to work around that.

PhysSong:
> You asked for ideas, not an essay

You're right. :sweat_smile:
> Maybe we should use this flag regardless of whether we employ an additional mechanism for this particular issue.

That also sounds fine.

PR #6372, opened 2022-04-13
Add Tap Tempo Feature

sakertooth:
This PR adds a Tap Tempo Feature in the Tools menu. Users can use either the mouse or keyboard to tap.
image

All it does is get the periods between each beat in seconds and uses that to incrementally update the average BPM. 
The BPM resets after 2 seconds of inactivity.

- There still needs to be a logo however, so hopefully someone can resolve that. I'm not too good on design, and I left the logo as pitch black. I was considering using the metronome logo, but I ran into difficultes trying to find it.
- For some reason, I have had discrepancies using the in-built metronome, but had rather okay results with other metronome's I found online from YouTube videos and websites. My PC may have had unusual latency though. If someone could test this that would be great.


superpaik:
I tested on W10. 
Tested against the inside metronome, I wasn't able to sinc with it. It always gives less bpms than the metronome.
Maybe the latency it's because it's build as an effect, and not inside the core ¿?
As for the code, I'm not a developer but I being following some threats about it and I don't think that you should be using QT functions unless it's strictly necessary (for the GUI). You should use c++ functions when possible.

sakertooth:
> I tested on W10. Tested against the inside metronome, I wasn't able to sinc with it. It always gives less bpms than the metronome. Maybe the latency it's because it's build as an effect, and not inside the core ¿? As for the code, I'm not a developer but I being following some threats about it and I don't think that you should be using QT functions unless it's strictly necessary (for the GUI). You should use c++ functions when possible.

I will try to see if the latency can be eliminated using the C++ standard library instead of QT. Perhaps will it be more accurate and thus sync with the metronome. Especially given the fact that small differences between periods for each beat mean rather large changes to the resulting BPM.

allejok96:
This feature is a good addition!

I think the problem comes down to storing `m_bpmAverate` as `int`. Even if `QTime::currentTime` was slightly delayed, the time between events should be correct. It's probably a rounding issue. To make it ridiculously correct you could count time difference from the first click always, instead of adding it up.

Superpaik has a point in using as little Qt as possible to make LMMS easier to port to other toolkits in the future, but in this case it's 90% Qt already and contained in the same file, so if it makes it simpler, I'd say go for `QTime`.

The button is slightly huge tho.

allejok96:
You should probably connect to [`pressed`](https://doc.qt.io/qt-5/qabstractbutton.html#pressed) instead of `clicked` although in the end the average will be the same I guess.

sakertooth:
> This feature is a good addition!
> 
> I think the problem comes down to storing `m_bpmAverate` as `int`. Even if `QTime::currentTime` was slightly delayed, the time between events should be correct. It's probably a rounding issue. To make it ridiculously correct you could count time difference from the first click always, instead of adding it up.
> 
> Superpaik has a point in using as little Qt as possible to make LMMS easier to port to other toolkits in the future, but in this case it's 90% Qt already and contained in the same file, so if it makes it simpler, I'd say go for `QTime`.
> 
> The button is slightly huge tho.

Funny you say that, because just as I was switching to use <chrono> instead of QTime, I also noticed that I was taking the average of the whole BPM, rather than taking the average of the double and just presenting it as an int. That was probably the central issue.

My upcoming commit uses the steady_clock and std::chrono::duration instead of QTime and, to be fair, it did not really introduce much complexity. The QTime implementation would surely be simpler but I see little reason in reverting at the moment.

And connecting to pressed instead of clicked does make sense, as there might be latency after the release rather than just calculating it as soon as the button is pressed. Thank you for noticing that. Will apply the changes soon.

Personal preference, but I am in favor of the large button. I will probably make it just a tad bit smaller though.

qnebra:
What about integrating it with "Tempo Set Value" window. Also, shouldn't be it able to set tempo value? Instead of being reliable on user to manually enter tempo value into "Tempo Set Value" window?

allejok96:
See also #6119

sakertooth:
This branch has been renamed to `tap-tempo`. See #6375.

sakertooth:
The branch has been renamed from `sakertooth:master` to `sakertooth:tap-tempo`. See #6375. Sorry for the inconvience.

PR #6374, opened 2022-04-14
Move widget files, add folder for track and instrument

allejok96:
Moving files in `src/gui`, no change to their source. Updated translations accordingly.

- `gui/clips` for `*ClipView`
- `gui/editors` for editors *and* their child widgets
- `gui/instrument` for `InstrumentTrackWindow` and child widgets
- `gui/tracks` for `*TrackView` and child widgets
- `gui/widgets` for generic widgets only


allejok96:
I didn't touch `LinkedModelGroupViews`, `Controls` and `ControlLayout`. I wonder if @JohannesLorenz thinks these should be moved out of `widgets` ?

Spekular (commented):
A few thoughts
- PositionLine is only used in editors, right? Maybe it should live in the editors folder.
- I'm not sure how I feel about ClipViews being in the track folder. I understand the reasoning, but that's not where I would expect them to be if I were searching for them.

Looks good otherwise

allejok96:
> I'm not sure how I feel about ClipViews being in the track folder.

I share that feeling :sweat_smile: but it would be nice if they lived together in a directory, even if it's only 5 of them.

> PositionLine is only used in editors, right? Maybe it should live in the editors folder.

I could see that. Same goes for TimeLineWidget, StepRecorderWidget and Rubberband.

JohannesLorenz:
> I didn't touch `LinkedModelGroupViews`, `Controls` and `ControlLayout`.

I designed `Controls` and `ControlLayout` as generic-purpose widgets, so I think they can stay in `widgets`.

For `LinkedModelGroupViews`, this is only useful if you have a dynamic set of controls (e.g. TripleOscillator is static, but e.g. Lv2, LADSPA etc. are dynamic, because the controls depend on the loaded plugin). Due to their complexity and their rare use case (Lv2 currently, possibly LADSPA, VST), I'd rather say not `widgets`?

allejok96:
Okay so LinkedModelGroupViews goes.

I'd like to move out `ControlLayout` too since it's not a QWidget even if it's general purpose (similarly to ActionGroup).

JohannesLorenz:
> I'd like to move out `ControlLayout` too since it's not a QWidget even if it's general purpose (similarly to ActionGroup).

Sounds good!

allejok96:
I don't know what the consensus is for more small widgets in the `editors` folder and a separate `clips` folder (like @Spekular insinuated), but I've made those changes. Now let's see what the people think.

PhysSong:
Why is `InstrumentTrackWindow.cpp` in `gui/instrument` while `SampleTrackWindow.cpp` is in `gui` and `InstrumentTrackView.cpp` is in `gui/tracks`?
I think `InstrumentTrackWindow.cpp` should stay in `gui` unless we move `SampleTrackWindow.cpp` to `gui/tracks`.

allejok96:
I understand how it can get confusing with many classes named "instrument". The `instrument` directory should really be called `instrumentwindow` to avoid all confusion...

> Why is `InstrumentTrackWindow.cpp` in `gui/instrument` while `SampleTrackWindow.cpp` is in `gui`?

I thought it made sense to put the window in the same directory as all its children, but I'll reverse it if I get a second opinion.

> Why is `InstrumentTrackView.cpp` is in `gui/tracks`?

`InstrumentTrackView` is a track (view)

> ... unless we move `SampleTrackWindow.cpp` to `gui/tracks`

 `SampleTrackWindow` is not a track, and not a child to a track (view)

allejok96:
I need some opinions before I can continue (will have to fix a lot of strings)

- [ ] Is `instrument` a good name for the directory that holds the instrument window and it's stuff?
- [ ] Is it ok to put `TrackContainerView` inside `tracks`? Motivation: where else would you expect to find it?
- [ ] Is it ok to put editor specific widgets, like time line, into `editors`?

Spekular:
Makes more sense to put `TrackContainerView` in `editors `than `tracks`, IMO, since it's the parent to both the song and pattern editors. I lean towards yes for the other two questions.

JohannesLorenz:
My opinion would be a copy of Spekular's:

1. Yes. What other name would fit better if it has instrument specific GUI stuff?
2. In `editors`. Because if our philosophy is that everything in `tracks` is a widget used inside tracks, then `TrackContainerView` would not fit `SongEditor` (which *contains* tracks). Also, it's more like an editor to me.
3. Yes. I think we said everything inside "editors" should be a widget used inside editors.

allejok96:
Thank goodness for the Unix toolbox. This oneliner finds all renamed files since a given commit and updates all strings to use the new path (works only for paths relative to base of repo). Maybe you might find it useful.
```sh
git ls-tree -rz --name-only HEAD | xargs -0 sed -i -f <(git diff --raw ENTER_COMMIT_ID_HERE | sed -En 's/.*R[0-9]+\t(.*)\t(.*)/s^\1^\2^g/p')
```

allejok96:
I've squashed the commits, ready for review.

JohannesLorenz:
@allejok96 While users are reviewing, if you rebase a local copy of this branch on #6379 (or wait until it's in master), you can check your locale updates.

allejok96:
@JohannesLorenz Strings look alright, except for `PatternClipView` which will cause a conflict with #6379 if I rename. Either I wait for check-strings to be merged, or you resolve the conflict later.

JohannesLorenz:
> @JohannesLorenz Strings look alright, except for `PatternClipView` which will cause a conflict with #6379 if I rename. Either I wait for check-strings to be merged, or you resolve the conflict later.

Both is OK to me. I'd say first come first serve?

allejok96:
I wait until the renames of this PR are approved.

JohannesLorenz:
If it's just for moving forward, I'd help you by reviewing this PR, but I'm rarely involved in GUI stuff, so I'm not a good candidate. I might try a review if no one else does it in the next days... :smiling_face_with_tear:

allejok96:
@JohannesLorenz and @Spekular's suggestions has been added. @PhysSong had some concerns - maybe we can get a second opinion? Or else we're good to go, as long as it builds :upside_down_face:

JohannesLorenz (approved):
Most widgets reside where I'd expect them. A few points:

1. `MixerLineLcdSpinbox` sounds a bit non-generic. Should it be in `src/gui`?
2. I wonder if `SubWindow` is so generic that it should go down into `widgets`?
3. Should we add a folder `src/gui/base-classes`? (`AutomatableButton`, `SideBarWidget`, `ControllerDialog`...)
4. We have a lot of `...View`s. Extra folder `src/gui/views`?
5. I wonder why we have an extra directory for `editors`, but leave the non-modal dialogs (`EffectControlDialog`) in `src/gui`. After all, aren't they both some kind of windows inside the MainWindow?

Anyways, these are all just suggestions. The PR is OK.

allejok96:
> I wonder why we have an extra directory for editors, but leave the non-modal dialogs (EffectControlDialog) in src/gui. 

I aimed to group things by concept rather than if it's a base/derived class, if it contains a specific word, or if it's located at a specific level in the GUI. I find this approach more helpful when looking for where a certain feature is implemented. So `editors` are for the main editors of LMMS and their specialized widgets. You'd never call `EffectControlDialog` an editor, so it's excluded.

> MixerLineLcdSpinbox sounds a bit non-generic. Should it be in src/gui?

Perhaps. I just put it where all of the other Lcd* siblings were, since there are a few.

> I wonder if SubWindow is so generic that it should go down into widgets?

On the other hand it's mostly used by `MainWindow`. Personally I wouldn't look for it in `widgets`.

JohannesLorenz:
I'd say let's merge it, then?

PhysSong:
I'm still not sure where `InstrumentTrackWindow.cpp` should be, but I'm okay with the others.

JohannesLorenz:
Note: As I merged the `check-strings` branch prior to this PR, it might be good if you rebase on master and see if CI still passes.

allejok96:
> PatternClipView which will cause a conflict

Lol I forgot to update my own remark. Well, now we know that CI works :smile:

PR #6375, opened 2022-04-14
Add Tap Tempo

### LINKED ISSUE #5181: Tap Tempo
One thing I would love to see in LMMS is the Ability to Tap your own Tempo. It can either be tapped by your spacebar or by your Left Mouse Click. This way you will definitly assign the Tempo you want.

sakertooth:
This is the #6372 PR, but with the branch renamed for organization.
Resolves #5181.


plugins/TapTempo/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+INCLUDE(BuildPlugin)
+
+BUILD_PLUGIN(taptempo TapTempo.cpp MOCFILES TapTempo.h EMBEDDED_RESOURCES artwork.png logo.png)
```
allejok96, line 3: `artwork.png` is usually the background image in plugin windows, you could remove it completely.

src/gui/widgets/TempoLcdSpinBox.cpp
```diff
@@ -0,0 +1,67 @@
...
+TempoLcdSpinBox::TempoLcdSpinBox(int numDigits, QWidget* parent, const QString& name) : 
+	LcdSpinBox(numDigits, parent, name) {}
+
+TempoLcdSpinBox::TempoLcdSpinBox(int numDigits, const QString& style, QWidget* parent, const QString& name) :
+	LcdSpinBox(numDigits, style, parent, name) {}
+
+void TempoLcdSpinBox::contextMenuEvent(QContextMenuEvent* _me) 
+{
+	CaptionMenu contextMenu(model()->displayName());
+	contextMenu.addAction("Tap Tempo", std::bind(&TempoLcdSpinBox::showTapTempoTool, this));
```
allejok96, line 44: Strings in UI should be `tr("Tap Tempo")` to enable localization

allejok96, line 44: To match up with other LMMS code use `[this]{ showTapTempoTool(); }` instead of `std::bind`

allejok96, line 44: Checking if the Tap tempo plugin is installed should happen before it is added to the menu.

allejok96 (changes requested):
Before you go ahead and fix any of my comments, I want to say I liked the old code better. It was simpler and more standalone. This line is the biggest pain point:

    return tool->model()->displayName() == "Tap Tempo";

Checking the display name isn't such a great idea as it could (potentially, not right now) be translated. But I see there is no other way. So maybe this integration is overkill.

Besides, opening the tap tempo tool from the tempo spinbox makes you think they are integrated is some way (like you could set the tempo etc) but they are actually not. And now we have two ways to open the same tap tempo window which feels odd.

Having it integrated in the "Set tempo" dialog is no better, because that window locks up the LMMS interface when opened. And maybe you just want to check the tempo of something and write that value to an automation track...

In the end of the day, writing a number in the tempo spinbox isn't that hard. Tap tempo is a good, simple and versatile tool, even without all that integration.

sakertooth:
I figured that adding a Tap Tempo option in the caption menu would be user-friendly but looking back it did feel rather problematic to add such a small change. It is true that finding the Tap Tempo plugin from the MainWindow in such a manner using `displayName` and providing access to `m_tools` without considering potential translations is a bit unusual. I also dont like how it attempts to find the plugin each time, which feels mostly wasteful and unncecessary, but it felt like the safest way.

Either way, given the ongoing major rework of the codebase, I dont think that such a small addition like Tap Tempo should make much, if any, change to the existing code, but rather add on to it in a modular way. Reverting.

plugins/TapTempo/TapTempo.cpp
```diff
@@ -0,0 +1,142 @@
...
+		m_bpmAverage += (60.0/period - m_bpmAverage)/(m_numTaps - 1);
+		m_bpmButton->setText(QString::number(std::round(m_bpmAverage)));
+		m_previousTime = currentTime;
+	}
+}
+
+void TapTempoView::keyPressEvent(QKeyEvent * event) 
+{
+	QWidget::keyPressEvent(event);
+	if (!m_keyDown)
```
Spekular, line 131: I think you can use `!event->isAutoRepeat() ` instead, and remove `m_keyDown` and `keyReleaseEvent`.

plugins/TapTempo/TapTempo.cpp
```diff
@@ -0,0 +1,142 @@
...
+		parentWidget()->layout()->setSizeConstraint(QLayout::SetFixedSize);
+
+		Qt::WindowFlags flags = parentWidget()->windowFlags();
+		flags |= Qt::MSWindowsFixedSizeDialogHint;
+		flags &= ~Qt::WindowMaximizeButtonHint;
+		parentWidget()->setWindowFlags(flags);
+	}
+}
+
+void TapTempoView::onBpmClick()
```
Spekular, line 102: Why not
- Grab `thisTime` (`currentTime`) immediately at the top of the function, then use that later
- Store `firstTime` and `latestTime`, use (latest - first) / (taps - 1) instead of m_bpmAverage

Something like this should work, right? (latestTime should be initialized to a long time ago)
```
thisTime = now
if (thisTime - latestTime > 2 seconds)
    // 2 seconds passed, reset calculation
    taps = 1
    firstTime = thisTime
    latestTime = thisTime
else
    // register latest tap
    taps++
    latestTime = thisTime
    
average = 60 * (taps - 1) / max(DBL_MIN, latestTime - thisTime)
setText( average )
```

sakertooth: My interpretation may be wrong, but I am not too sure I fully agree with this implementation.

The reason why the current implementation uses an average BPM and wave period rather than just the most current one is because very small changes in wave periods make the BPM less stable and shift more frequently. So, it is, or at least how I see it, more favorable to consider the BPM over each tap as time passes.

Say, for example, the difference in time between 2 taps was 0.2587 seconds and another was 0.2892 seconds. Knowing that the number of beats per second (Hz) is equal to 1 divided by the wave period in seconds:
- Beats per minute given that the distance between each beat is 0.2587 seconds: 60/0.2587 = ~231.929
- Beats per minute given that the distance between each beat is 0.2892 seconds: 60/0.2892 = ~207.469

The difference in seconds between each beat was 0.0305 (30.5 ms, basically no difference in time to the average person) but the BPM drastically changed by 24.46. If instead we were to take an average BPM over time, we would get a better understanding as to what the BPM is.

Seeing `average = 60 * (taps - 1) / max(DBL_MIN, latestTime - thisTime)`, what would happen if `taps` was > 2?

Assuming `latestTime - thisTime` was, say, 0.4893 and the number of taps was 5, wouldn't this expression get the number of beats for every 4 minutes instead of 1 minute?

I may be wrong however.

Spekular: Sorry, I misstyped. The average should use `firstTime - thisTime` as the period. This ensures that floating point inaccuracies don't accumulate & the tempo gets more accurate with each click. 

> Say, for example, the difference in time between 2 taps was 0.2587 seconds and another was 0.2892 seconds.

With the fixed equation, the first 2 taps would give 231.93 as you say, and then the third would update to 219.02

`60 * (0.2587 + 0.2892)/2`

sakertooth: I understand now. This is counting the number of taps that happen in the increasing time duration, rather than taking an average of each period. The code is more compact and it looks like the tempo becomes accurate faster. Thanks.

plugins/TapTempo/TapTempo.cpp
```diff
@@ -0,0 +1,140 @@
...
+{
+	auto currentTime = std::chrono::steady_clock::now();
+	auto distanceFromPreviousTime = std::chrono::duration_cast<std::chrono::duration<double>>(currentTime - m_previousTime).count();
+
+	if (distanceFromPreviousTime > 2.0)
+	{
+		m_numTaps = 1;
+		m_firstTime = currentTime;
+		m_previousTime = currentTime;
+		m_bpmButton->setText("0");
```
Spekular, line 114: This and the return can be removed. Since `numTaps` is set to 1, `numTaps - 1` will be 0, and the bpm.calculation will return 0. The fact that `currentTime == previousTime` would give 0 in the denominator is prevented by `std::max(DBL_MIN, ...)`

Spekular: Ah, nevermind. With the other changes you've made this wouldn't work.

allejok96:
I have a little feature request: displaying BPM as ms and Hz somewhere below the button perhaps. Could be good if you want to set a delay or rate knob that isn't a tempo sync knob for some reason.

sakertooth:
> I have a little feature request: displaying BPM as ms and Hz somewhere below the button perhaps. Could be good if you want to set a delay or rate knob that isn't a tempo sync knob for some reason.

I have added this feature, but I must ask: wouldn't knowing the Hz be enough? I dont mind adding BPM in ms however, I just currently think it will be simpler to have just the Hz.

Spekular:
There are a lot of plugins where e.g. delays or envelopes can be specified in milliseconds, in which case it's valuable to know how long each beat is.

sakertooth:
Seems like it's rather complete. Hopefully someone can verify if everything is in place for good measure.

allejok96:
[I made some touchups](https://github.com/sakertooth/lmms/pull/1)

sakertooth:
> [I made some touchups](https://github.com/sakertooth/lmms/pull/1)

Thank you for this, although I think it would be nice if the tooltip was more descriptive. Very minor, but to indicate a kind of action, maybe "Display high precision" would work better. Do you agree with this change before I add it in?

RoxasKH:
So wait, we're making the tempo tap displaying decimals BPMs while LMMS itself doesn't allow decimals BPMs?

I'm asking because imo it should be the opposite like in FL Studio: you have the opportunity to slightly change your BPM with decimals in the BPM normal box, but when using the tap tempo feature, i think it's more useful and reasonable for it to show and set only integer values, considering that obviously human tapping isn't accurate and you usually wanna get as a result the nearest integer BPM more than the actual weird value you get with your tapping.

Spekular:
Anyone who wants an integer value can easily round to it from a more specific value, whereas the opposite is not true. I would default to displaying two decimals, personally, so it's easier to tell your BPM has actually stabilized to an accurate value. 

You can't always assume that the song you're working with has a fixed integer BPM in the first place, and an integer display is misleading for songs with a non-integer or slightly varying bpm. With just whole numbers one might assume that e.g. the following sequence of 8 taps is stable: 84.05, 84.08, 84.14, 84.20, 84.22, 84.25, 84.30, 84.35.

sakertooth:
Rebased. I think this is complete.

allejok96:
Everything looks good, now I'll only add that [the metronome icon](https://raw.githubusercontent.com/LMMS/lmms/master/data/themes/default/metronome.png) would look nicer.

plugins/TapTempo/TapTempo.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+	if (m_numTaps == 0)
+	{
+		m_startTime = currentTime;
+		m_prevTime = currentTime;
+		m_lastPrevTime = currentTime;
+	}
+	else if (m_numTaps == 1) { m_prevTime = currentTime; }
+	else if (m_numTaps >= 2)
+	{
```
irrenhaus3, line 110: Using the `std::chrono_literals` namespace, you can convert to chrono::milliseconds in a more readable way by writing `auto interval = (time2 - time1) / 1.0ms;`. Saves you the cumbersome duration_cast syntax.

plugins/TapTempo/TapTempo.h
```diff
@@ -0,0 +1,98 @@
...
+	void onBpmClick();
+
+	void keyPressEvent(QKeyEvent*) override;
+	void closeEvent(QCloseEvent*) override;
+
+private:
+	void reset();
+	void updateLabels();
+
+private:
```
irrenhaus3, line 66: I suggest `using clock = std::chrono::steady_clock;` at class scope, and then using the available `clock::time_point` and `clock::duration` aliases; this keeps the type names in our code more readable - e.g. the types of the 3 variables below would change from `std::chrono::time_point<std::chrono::steady_clock>` to a simple `clock::time_point`.

irrenhaus3 (approved):

irrenhaus3 (approved):

plugins/TapTempo/TapTempo.h
```diff
@@ -0,0 +1,99 @@
...
+	static constexpr int s_numRecentTaps = 3;
+	static constexpr int s_bpmDifferenceThreshold = 30;
+	double m_bpm = 0.0;
+	bool m_showDecimal = false;
+};
+} // namespace gui
+
+class TapTempo : public ToolPlugin
+{
+	Q_OBJECT;
```
PhysSong, line 80:
```suggestion
	Q_OBJECT
```

plugins/TapTempo/TapTempo.h
```diff
@@ -0,0 +1,99 @@
...
+ * If this threshold is passed, the counter gets reset.
+ **/
+constexpr double TAP_INTERVAL_THRESHOLD_MS = 500.0;
+
+class TapTempoView : public ToolPluginView
+{
+	Q_OBJECT
+public:
+	using clock = std::chrono::steady_clock;
+	TapTempoView(ToolPlugin*);
```
PhysSong, line 56: Separating them by a newline looks better to me.

plugins/TapTempo/TapTempoUi.h
```diff
@@ -0,0 +1,195 @@
+/********************************************************************************
+** Form generated from reading UI file 'taptempo.ui'
```
PhysSong, line 2: I guess you initially desinged the UI using Qt designer. You should probably cleanup UI class stuff.

PhysSong: I mean, why this class should even exist apart from `TapTempoView`?

sakertooth: When reading up on Qt's UI designer file documentation, this was one of the approaches to get it in code, which was using the ``setupUi`` function.

It's nice that I can separate the stylistic part of the UI into a separate UI file, and focus on only what I want to focus on in the plugin view file.

If I moved the UI into ``TapTempoView``, the whole file would be cluttered, making things harder to review and code.

plugins/TapTempo/TapTempo.h
```diff
@@ -0,0 +1,90 @@
...
+
+class TapTempo : public ToolPlugin
+{
+	Q_OBJECT
+public:
+	TapTempo();
+
+	virtual gui::PluginView* instantiateView(QWidget*) { return new gui::TapTempoView(this); }
+
+	virtual QString nodeName() const;
```
messmerd, line 78:
```suggestion
	gui::PluginView* instantiateView(QWidget*) override { return new gui::TapTempoView(this); }

	QString nodeName() const override;
```
Removed `virtual` and added `override`

messmerd (commented):
Just a few suggestions

plugins/TapTempo/TapTempoUi.h
messmerd: I think I agree with PhysSong on this one. `TapTempoView` and `TapTempoUi` should probably be combined into a single `TapTempoView` class in a `TapTempoView.h` file.

If the concern is over the clutter, it wouldn't be very cluttered if you created a corresponding source file `TapTempoView.cpp` to implement the methods currently in `TapTempoUi` rather than doing it all in the header.

sakertooth: > I think I agree with PhysSong on this one. `TapTempoView` and `TapTempoUi` should probably be combined into a single `TapTempoView` class in a `TapTempoView.h` file.
> 
> If the concern is over the clutter, it wouldn't be very cluttered if you created a corresponding source file `TapTempoView.cpp` to implement the methods currently in `TapTempoUi` rather than doing it all in the header.

Just went ahead and merged the two files. Let me know if there is any further action needed or it's alright :+1:.

plugins/TapTempo/TapTempoView.h
```diff
@@ -0,0 +1,70 @@
...
+class TapTempo;
+}
+
+namespace lmms::gui {
+
+class TapTempoView : public ToolPluginView
+{
+	Q_OBJECT
+public:
+	TapTempoView(ToolPlugin*);
```
messmerd, line 50:
```suggestion
	TapTempoView(TapTempo* plugin);
```
You could probably pass the `TapTempo` pointer instead - then you won't have to dynamic_cast it in the source file.

plugins/TapTempo/TapTempoView.h
```diff
@@ -0,0 +1,70 @@
...
+	QLabel* m_hzLabel;
+	QVBoxLayout* m_verticalLayout;
+	QCheckBox* m_precisionCheckBox;
+	QCheckBox* m_muteCheckBox;
+	TapTempo* m_plugin;
+	friend class TapTempo;
+};
+} // namespace lmms::gui
+
+#endif
```
messmerd, line 70:
```suggestion
#endif // LMMS_GUI_TAP_TEMPO_VIEW_H

```
style

plugins/TapTempo/TapTempoView.h
```diff
@@ -0,0 +1,70 @@
...
+namespace lmms::gui {
+
+class TapTempoView : public ToolPluginView
+{
+	Q_OBJECT
+public:
+	TapTempoView(ToolPlugin*);
+	void closeEvent(QCloseEvent*) override;
+	void keyPressEvent(QKeyEvent*) override;
+	void updateLabels();
```
messmerd, line 53:
```suggestion
	void updateLabels();
	void keyPressEvent(QKeyEvent* event) override;
	void closeEvent(QCloseEvent*) override;
```
`keyPressEvent(...)` uses its parameter, so I gave it a name to suppress IntelliSense warning. Also reordered the functions to match the order in the *.cpp file because why not

plugins/TapTempo/TapTempoView.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+#include "Engine.h"
+#include "SamplePlayHandle.h"
+#include "Song.h"
+#include "TapTempo.h"
+
+namespace lmms::gui {
+TapTempoView::TapTempoView(ToolPlugin* plugin)
+	: ToolPluginView(plugin)
+	, m_plugin(dynamic_cast<TapTempo*>(plugin))
```
messmerd, line 45:
```suggestion
TapTempoView::TapTempoView(TapTempo* plugin)
	: ToolPluginView(plugin)
	, m_plugin(plugin)
```
Should be able to do this - no more dynamic_cast

messmerd (commented):
A few more suggestions

plugins/TapTempo/TapTempoUi.h
messmerd: I think it looks a lot better now! Other than the couple small issues I just pointed out in a review, everything looks fine to me coding-wise. I haven't looked into the logic or tested the plugin yet, but I may be able to do so tomorrow

plugins/TapTempo/TapTempo.cpp
```diff
@@ -0,0 +1,104 @@
...
+#include "plugin_export.h"
+
+namespace lmms {
+
+extern "C" {
+Plugin::Descriptor PLUGIN_EXPORT taptempo_plugin_descriptor
+	= {LMMS_STRINGIFY(PLUGIN_NAME), "Tap Tempo", QT_TRANSLATE_NOOP("PluginBrowser", "Tap to the beat"),
+		"saker <sakertooth@gmail.com>", 0x0100, Plugin::Tool, new PluginPixmapLoader("logo"), nullptr, nullptr};
+
+TapTempo::TapTempo()
```
messmerd, line 53: It looks like `TapTempo::TapTempo()` and `void TapTempo::onBpmClick()` are defined within the `extern "C"` block when they should be outside it. You could probably just move `lmms_plugin_main(...)` up to the top so that it can be grouped together with the descriptor, then the `TapTempo` methods could come after that. However you want to do it

plugins/TapTempo/TapTempoView.h
sakertooth: Odd how I didn't get a warning. Not sure if you're using VSCode or something different, so maybe that's why (or mine just isn't working for some reason). I'll still commit though, thanks :+1:

plugins/TapTempo/TapTempo.cpp
sakertooth: Yeah, I must've put everything into ``extern "C"`` by accident. Nice catch :+1:

plugins/TapTempo/TapTempoView.h
messmerd: I'm at home now and using VSCode, so I didn't actually test it, but when I'm at work and using Visual Studio, things like this give me warnings due to the method signature in the header having a different parameter name than the implementation in the .cpp file.

messmerd:
I just tested it. Here are some things I noticed:
- Clicking "Precision" or "Mute" checkboxes puts the focus on those controls, so using the spacebar will select/deselect those boxes instead of working as tap tempo input.
- If the rate that I use tap Spacebar or click changes too much, it will briefly say the tempo is zero, which is a little strange
- I wonder if Tap Tempo could be integrated with the global tempo control somehow, and whether that would be worthwhile?
- Tap Tempo should have a better icon
- Mute Tap Tempo, set the LMMS global tempo to some value, then click play in the Song-Editor to start the metronome. Then click Tap Tempo at a rate significantly lower than the metronome's rate for a while before trying to match the rate of the metronome. If you do this, the BPM that Tap Tempo gives should be a noticeably lower BPM than the global tempo, and slowly increase to approach the value that it should be. Is this something you're able to replicate? Play around with it a little - for me, the BPM often doesn't quite match what it should be - it often settles on an equilibrium which isn't very accurate.

sakertooth:
> Clicking "Precision" or "Mute" checkboxes puts the focus on those controls, so using the spacebar will select/deselect those boxes instead of working as tap tempo input.

Yeah, I've noticed that too. A bit annoying. Will fix.

> If the rate that I use tap Spacebar or click changes too much, it will briefly say the tempo is zero, which is a little strange

This is because the counter gets reset when the BPM on the display does not closely reflect the current BPM. This gives a chance for the counter to get more accurate readings in the following taps. When the user taps, there are two intervals: the one from the current time and the time when the user first tapped the button, and the other is from the current time and the last 3 taps. The latter is more accurate, while the former stabilizes better at a certain number but does not play well to rapid changes in BPM. If the difference between these two BPM's is greater than 30 (an arbitrary threshold I set), then the counter gets reset. Instead of resetting however, I could just set the counter to use the BPM from the more accurate interval (``currentTime`` - ``prevTime``), and then do further calculations using the stable interval (``currentTime`` - ``startTime``).

I'm open to suggestion if you have a better algorithm that not only stabilizes but also responds well to rapid changes in BPM.

> Tap Tempo should have a better icon

Not the best artist :sweat_smile:. Hopefully someone can come by and improve this somewhat?

> Mute Tap Tempo, set the LMMS global tempo to some value, then click play in the Song-Editor to start the metronome. Then click Tap Tempo at a rate significantly lower than the metronome's rate for a while before trying to match the rate of the metronome. If you do this, the BPM that Tap Tempo gives should be a noticeably lower BPM than the global tempo, and slowly increase to approach the value that it should be. Is this something you're able to replicate? Play around with it a little - for me, the BPM often doesn't quite match what it should be - it often settles on an equilibrium which isn't very accurate.

As previously mentioned, when the BPM from the stable interval and the accurate interval diverge and pass a threshold of a difference of 30 BPM, the counter gets reset. So, if you get the counter to reset and then start playing the tempo of the project, it should converge to it. I could change this behavior to use the assign the display BPM to the BPM from the more accurate interval when the threshold is passed instead of resetting, but it wouldn't be a slow ramp up, so I would have to make the more accurate interval a bit more stable.

The whole algorithm is a bit confusing, so once again I am open to any adjustments to it. The algorithm starts of simple, and we just calculate the BPM from the interval ``currentTime`` - ``startTime``, but if user where to change their pace, it wouldn't accurately reflect that, which is were it gets a bit more intricate.

zonkmachine:
> I wonder if Tap Tempo could be integrated with the global tempo control somehow, and whether that would be worthwhile?

I was thinking this too. You could have a context menu option to launch the tap tempo app.

zonkmachine (approved):
I think this is good to be merged. There are things that could be added as discussed above but I think it's better to merge now and add stuff later.

sakertooth:
> I think this is good to be merged. There are things that could be added as discussed above but I think it's better to merge now and add stuff later.

I'm just going to add a "Sync" button that syncs the displayed BPM with the project tempo, but after that I think that'll be it. The other features/changes mentioned might be beneficial, but could probably be looked at in another PR if the need arises.

plugins/TapTempo/TapTempoView.cpp
```diff
@@ -0,0 +1,169 @@
...
+		m_plugin->onBpmClick();
+		updateLabels();
+
+		if (!muteCheckBox->isChecked())
+		{
+			const auto timeSigNumerator = Engine::getSong()->getTimeSigModel().getNumerator();
+			m_plugin->m_numTaps % timeSigNumerator == 0
+				? Engine::audioEngine()->addPlayHandle(new SamplePlayHandle("misc/metronome02.ogg"))
+				: Engine::audioEngine()->addPlayHandle(new SamplePlayHandle("misc/metronome01.ogg"));
+		}
```
messmerd, line 116:
```suggestion
		if (!muteCheckBox->isChecked())
		{
			const auto timeSigNumerator = Engine::getSong()->getTimeSigModel().getNumerator();
			Engine::audioEngine()->addPlayHandle(new SamplePlayHandle(
				m_plugin->m_numTaps % timeSigNumerator == 0 ? "misc/metronome02.ogg" : "misc/metronome01.ogg"));
		}

		m_plugin->onBpmClick();
		updateLabels();
```
This makes the metronome's down beat the 1st beat instead of the last. I also made the call to `addPlayHandle` a little simpler.

plugins/TapTempo/TapTempoView.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+	connect(precisionCheckBox, &QCheckBox::toggled, [this](bool checked) {
+		m_plugin->m_showDecimal = checked;
+		updateLabels();
+	});
+
+	connect(syncButton, &QPushButton::clicked, this, [this]() {
+		const auto& tempoModel = Engine::getSong()->tempoModel();
+		if (m_plugin->m_bpm < tempoModel.minValue() || m_plugin->m_bpm > tempoModel.maxValue()) { return; }
+		Engine::getSong()->setTempo(m_plugin->m_bpm);
```
messmerd, line 129:
```suggestion
		Engine::getSong()->setTempo(std::round(m_plugin->m_bpm));
```
I think the bpm should be rounded first, so that values like 119.9 aren't truncated to 119.

messmerd (commented):
I made a couple small fixes.

There's also a memory leak detected after closing LMMS. It only happens when the Tap Tempo metronome sound plays, but I haven't figured out why it happens yet. Here's the error message:

```
lmms: /home/dalton/Documents/lmms_sakertooth/src/3rdparty/rpmalloc/rpmalloc/rpmalloc/rpmalloc.c:1328: _rpmalloc_span_finalize: Assertion `(span->list_size == span->used_count) && "Memory leak detected"' failed.
Aborted (core dumped)
```

sakertooth:
> I made a couple small fixes.
> 
> There's also a memory leak detected after closing LMMS. It only happens when the Tap Tempo metronome sound plays, but I haven't figured out why it happens yet. Here's the error message:
> 
> ```
> lmms: /home/dalton/Documents/lmms_sakertooth/src/3rdparty/rpmalloc/rpmalloc/rpmalloc/rpmalloc.c:1328: _rpmalloc_span_finalize: Assertion `(span->list_size == span->used_count) && "Memory leak detected"' failed.
> Aborted (core dumped)
> ```

That's unrealted to this PR. This has been an outstanding issue that was looked at previously with #5733, although I'm not sure why it still happens.

messmerd:
@sakertooth I still wonder why that sets off the rpmalloc leak detection but using the metronome doesn't, nor previewing samples.

messmerd (approved):
Works pretty well when I tested it, and it behaves similarly to other Tap Tempo tools that I tried online.

I think I'm fine with merging it now, provided the rpmalloc issue is not a problem with this PR.

PR #6379, opened 2022-04-18
Add check-strings

JohannesLorenz:


Veratil:
Are we only searching `.h` and `.cpp` files?

tests/check-strings/check-strings
```diff
@@ -0,0 +1,152 @@
...
+if not os.path.exists(".gitmodules"):
+    print('You need to call this script from the LMMS top directory')
+    exit(1)
+
+result = subprocess.run(['git', 'ls-files', 'src/', 'include/', 'plugins/'], capture_output=True, text=True, check=True)
+files = result.stdout.splitlines()
+classes = set()
+class_pat = re.compile(r'class(\s+LMMS_EXPORT)?\s+([a-zA-Z0-9_]+)')
+for cur_file in files:
+    if os.path.isfile(cur_file):
```
PhysSong, line 46: It's matter of style, but you can reduce indentation by negating the condition and using `continue`.

tests/check-strings/check-strings
```diff
@@ -0,0 +1,152 @@
...
+        with open(cur_file, errors='replace') as fp:
+            lines = fp.readlines()
+            for line in lines:
```
PhysSong, line 49: you can either use `for line in fp:` or unindent the `for` loop by one level.

tests/check-strings/check-strings
```diff
@@ -0,0 +1,152 @@
...
+    root = tree.getroot()
+    for location in root.findall('./context/message/location'):
+        filenames.add(location.attrib['filename'])
+    for location in root.findall('./context/name'):
+        classes_found.add(location.text)
+for f in sorted(filenames):
+    if not os.path.isfile(f) and not os.path.isfile("data/locale/" + f):
+        error('data/locale', 'Source file does not exist: ' + f)
+for c in sorted(classes_found):
+    if c[0] != 'Q' and not '::' in c and not c in classes:
```
PhysSong, line 81: Why don't you pull out the checks for classes which are excluded in this check? The condition may change later and is used in one more place.

tests/check-strings/check-strings
```diff
@@ -0,0 +1,152 @@
...
+        error('data/locale', 'Class does not exist in source code: ' + c)
+
+
+caption('themes')
+
+for cur_file in sorted(os.listdir(os.fsencode('data/themes'))):
+    if os.path.isdir('data/themes/' + cur_file.decode("latin-1")):
+        filename = 'data/themes/' + os.fsdecode(cur_file) + '/style.css'
```
PhysSong, line 89: - Why `latin-1`?
- Why don't you check existence of `filename` instead?

tests/check-strings/check-strings
```diff
@@ -0,0 +1,152 @@
...
+        classes_in_sheet = set()
+        with open(filename) as fp:
+            css = fp.read()
+            rules = tinycss2.parse_stylesheet(str(css))
```
PhysSong, line 93: You don't need a `str` because `fp` uses text IO.

PhysSong (commented):
Part of my review. Some comments which belongs nowhere:
- For formatting strings, f-string looks better than simple concatenation
- `"foo"` and `'foo'` are mixed without clear rules

JohannesLorenz:
> Are we only searching `.h` and `.cpp` files (for classes)?

@Veratil Good point. I had to add `.hpp` (for some headers in Carla) and `.ui`. I pushed again and now I think they are sufficient.

tests/check-strings/check-strings
JohannesLorenz: For your first question: Because our filenames are all in `latin-1`, right? `data/themes/default/` and `data/themes/classic/` are all latin chars. Or what do you suggest?

JohannesLorenz:
The github actions work now, but the PR does not display it, probably because it did not work in the beginning. But you can use e.g. [this link](https://github.com/JohannesLorenz/lmms/runs/6084234733) to see it in action.

For some strange reason it still complains about 5 "swh" files, saying they were not there, but they are there. It works on my machine(TM), can anyone check if they get the error on their machine?

tests/check-strings/check-strings
Veratil: ASCII would be the most basic, and able to convert to UTF-8 without problem. latin-1 defines characters in the range 128-255 which isn't directly compatible with UTF-8.

We should try either ASCII or utf-8 for all.

PhysSong:
> For some strange reason it still complains about 5 "swh" files, saying they were not there, but they are there.

On `master`, those `.c` files are generated from `.xml` files on build time.

PhysSong:
> The github actions work now, but the PR does not display it

It's only running on your fork, indeed.

> > For some strange reason it still complains about 5 "swh" files, saying they were not there, but they are there.

> On master, those .c files are generated from .xml files on build time.

You may have to make affected files to point `.xml` instead.

tests/check-strings/check-strings
```diff
@@ -0,0 +1,183 @@
...
+files = result.stdout.splitlines()
+if os.path.isdir('plugins/carlabase/carla'):
+    carlabase = 'carlabase'
+else:
+    carlabase = 'CarlaBase'
+result = subprocess.run(['git',
+                         '--git-dir',   'plugins/'+carlabase+'/carla/.git',
+                         '--work-tree', 'plugins/'+carlabase+'/carla/',
+                         'ls-files', 'resources/ui', 'source/frontend'], capture_output=True, text=True, check=True)
+files_carla = ['plugins/CarlaBase/carla/' + i for i in result.stdout.splitlines()]
```
PhysSong, line 51: - Why don't you put `'plugins/'+carlabase+'/carla` in a variable?
- Why you used `'plugins/CarlaBase/carla/'` instead of ``'plugins/'+carlabase+'/carla`` when collecting paths?

JohannesLorenz:
> You may have to make affected files to point `.xml` instead.

Thanks for the hint. Finally fixed that last issues with the SWH files. Will fix the style comments the next days.

tests/check-strings/check-strings
JohannesLorenz: We now uses `re.findall()` (Veratil's idea)

tests/check-strings/check-strings
JohannesLorenz: At least, for this case, I could remove `c[0] != 'Q'` completely. However, I can write a function for the other one.

JohannesLorenz:
I cleaned up the whole code with the help of your comments. Ready for review!

data/themes/classic/style.css
```diff
@@ -585,10 +585,6 @@ SideBar QToolButton {
 
 /* Instrument plugin list */
 
-PluginDescList {
-	background-color: #5b6571;
-}
-
```
Veratil, line 591: Just making sure, this is unused now?

tests/check-strings/check-strings
```diff
@@ -0,0 +1,158 @@
...
+
+errors = 0
+
+
+# functions
+
+def caption(my_str):
+    print('')
+    print(f'# {my_str}')
+    print('')
```
Veratil, line 27:
```suggestion
    print(f'\n# {my_str}\n')
```

tests/check-strings/check-strings
```diff
@@ -0,0 +1,158 @@
...
+            if mpath.parent == Path('plugins/LadspaEffect/swh/ladspa/'):
+                mpath = mpath.with_suffix('.xml')
+            if not mpath.is_file():
+                error(str(cur_file), f'Source file does not exist: {str(mpath)}')
+
+
+caption('debian docs (only one string)')
+
+# Checks for caps.html. This gets relevant when #4027 will be merged
+for line in Path('debian/lmms-common.docs').read_text(errors='replace'):
```
Veratil, line 134: This doesn't give you line by line. This gives you the entire file contexts and then go through it character by character.
```suggestion
for line in Path('debian/lmms-common.docs').read_text(errors='replace').splintlines():
```

data/themes/classic/style.css
JohannesLorenz: The class has been removed in 3ef33dbbe994141ce3d7c5da7b9d44bcb1fdffb2, and the color scheme after that commit has been accepted (See description of 8e2c586ea87806b10edad385e257fd85f52341a0). So it is unused. Also, executing LMMS with this PR, it looks unchanged.

JohannesLorenz:
@Veratil Thanks for your review. I resolved all of your comments.

Veratil (approved):
Alright, LGTM. :)

tests/check-strings/check-strings
```diff
@@ -0,0 +1,156 @@
...
+carlapath = f'plugins/{carlabase}/carla/'
+result = subprocess.run(['git',
+                         '--git-dir',   f'{carlapath}/.git',
+                         '--work-tree', f'{carlapath}',
+                         'ls-files', 'resources/ui', 'source/frontend'], capture_output=True, text=True, check=True)
+files.extend([Path(f'{carlapath}/{f}') for f in result.stdout.splitlines()])
+
+classes = set()
+class_pat = re.compile(r'^\s*class(?:\s+LMMS_EXPORT)?\s+([a-zA-Z0-9_]+)', re.MULTILINE)
+class_pat_ui = re.compile(r'<class>([a-zA-Z0-9_]+)</class>')
```
PhysSong, line 57: Class names can't start with numbers, so `[a-zA-Z0-9_]+` should be `[a-zA-Z_][a-zA-Z0-9_]*` instead.

JohannesLorenz:
Thanks for the reviews. Fixed everything. Should I merge now?

Rossmaxx (changes requested):
reported some inconsistencies i saw

data/locale/sr.ts
```diff
@@ -637,7 +637,7 @@ If you&apos;re interested in translating LMMS in another language or want to imp
     </message>
 </context>
 <context>
-    <name>BBTrack</name>
+    <name>PatternTrack</name>
     <message>
         <source>Beat/Bassline %1</source>
```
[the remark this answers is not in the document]
Rossmaxx: ie line 642

data/locale/sr.ts
```diff
@@ -614,7 +614,7 @@ If you&apos;re interested in translating LMMS in another language or want to imp
     </message>
 </context>
 <context>
-    <name>BBClipView</name>
+    <name>PatternClipView</name>
     <message>
         <source>Open in Beat+Bassline-Editor</source>
```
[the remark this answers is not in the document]
Rossmaxx: line 619

data/locale/sr.ts
```diff
@@ -563,7 +563,7 @@ If you&apos;re interested in translating LMMS in another language or want to imp
     </message>
 </context>
 <context>
-    <name>BBEditor</name>
+    <name>PatternEditor</name>
     <message>
         <source>Beat+Bassline Editor</source>
```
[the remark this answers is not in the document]
Rossmaxx: line 568

allejok96:
Regarding the css parser errors I played around with some 100% fool proof future proof regex
```py
# Remove /*comments*/
nocomment = re.sub(r'/\*.*?\*/', rawcss, re.DOTALL)

# Remove { blocks }
noblocks = re.sub(r'{.*?}', nocomments, re.DOTALL)

# Remove [properties="like this"]
noprops = re.sub(r'\[.*?\]', noblocks, re.DOTALL)

# Extract ClassName out of .ClassName::sub-control:state#objectName
classes = re.findall(r'\.?(\w+)[:#\w-]*')
```

data/locale/sr.ts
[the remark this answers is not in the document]
allejok96: If we would replace the English string here, Serbian LMMS would (in theory) keep calling it BB Editor. All translators need to manually provide a new translation for "Pattern Editor". The incorrect line will probably be removed automatically once translations are updated.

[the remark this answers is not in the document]
JohannesLorenz: > Serbian LMMS would (in theory) keep calling it BB Editor

Why would they keep calling it BB Editor? I can't follow.

allejok96:
Using tinycss:
```py
import tinycss2

classes_in_sheet = set()
with open('data/themes/default/style.css') as fp:
    rules = tinycss2.parse_stylesheet(str(fp.read()))
    
    for rule in rules:
        if rule.type == 'qualified-rule':
            class_found = False
            for c in rule.prelude:
                if c.type == 'ident' and not class_found:
                    classes_in_sheet.add(c.value)
                    class_found = True
                # After whitespace or comma comes a new class
                elif c.type == 'whitespace' or (c.type == 'literal' and c.value == ','):
                    class_found = False

print(classes_in_sheet)
```

data/locale/sr.ts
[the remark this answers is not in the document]
allejok96: It was a weird example. I we changed the English strings in the `.ts` file but left the Serbian strings untouched I guess that would translate occurrences of "Pattern Editor" in English to "BB editor" in Serbian.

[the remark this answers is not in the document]
JohannesLorenz: OK, then let's ignore these `<name>` strings in the translations for now? :smiley:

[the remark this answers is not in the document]
allejok96: Yeah, I never thought of that. Are these files are generated by Transifex? If so, will changes to them will be overwritten when new translations are pulled in? And will Transifex detect the renaming or will it treat it like one class/context being deleted and another context full of strings being added?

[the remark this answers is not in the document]
JohannesLorenz: I'll tag @liushuyu , as they are our Transifex expert.

JohannesLorenz:
@Veratil what do you think of allejok's proposal?

Veratil:
I didn't realize that cssutils threw out some stuff, so we can totally go back to using tinycss instead. Not a fan of the 6 levels of indention there though. 😁

data/locale/sr.ts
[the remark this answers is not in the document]
liushuyu: > Yeah, I never thought of that. Are these files are generated by Transifex? If so, will changes to them will be overwritten when new translations are pulled in? And will Transifex detect the renaming or will it treat it like one class/context being deleted and another context full of strings being added?

It doesn't matter, as long as the original text is unchanged, the translations would be applied correctly.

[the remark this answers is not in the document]
JohannesLorenz: @Rossmaxx I'll resolve your comments then, as disccussed with @allejok96 in this comment. I hope this is OK.

JohannesLorenz:
I pushed a switch to `tinycss` again. Was that correct?

I didn't know how to save more indention, feel free to comment.

JohannesLorenz:
I suggest if no one starts a review, I'll just merge this. It's not a harmful PR anyways.

data/locale/sr.ts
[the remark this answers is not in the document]
Rossmaxx: Yeah its ok. Just reported what i saw.

Rossmaxx:
> I suggest if no one starts a review, I'll just merge this. It's not a harmful PR anyways.

Yeah it would be better to merge it.

allejok96:
Just to make sure I understand: `expectation.txt` is just old errors used by the verify script to test against the old code base. After this PR is merged you can expect check-strings to output no errors.

I think `verify` is a good tool for the check-strings dev, but probably something that person can check manually on the occasion he/she is altering check-strings. IMO a bit overkill to have in CI.

`verify` will also return failure if check-strings is improved to find more errors, and in that case `expectation.txt` will need to be updated to match it.

JohannesLorenz:
> Just to make sure I understand: `expectation.txt` is just old errors used by the verify script to test against the old code base. After this PR is merged you can expect check-strings to output no errors.

Correct.

> I think `verify` is a good tool for the check-strings dev, but probably something that person can check manually on the occasion he/she is altering check-strings. IMO a bit overkill to have in CI.

The sole reason I can see to exclude it from CI is to save energy. Aside from that, my experience is that if you don't run tests regularly, everyone forgets to use them and after years you'll end up with a "this has not been working for years" situation. :smiley:

JohannesLorenz:
Merge time?

Rossmaxx:
yup

PR #6381, opened 2022-04-18
Loop marker quick fix

allejok96:
I know @Spekular has something coming up (hope you don't mind).

Make it possible to grab the end of the loop even if it goes outside the visible area.


Spekular:
This would effectively be reverted by my upcoming PR, FWIW.

allejok96:
It was worth a shot :grin: diffs are cheap

PR #6382, opened 2022-04-18
New loop marker shortcuts, attempt 2

Spekular:
Changes:
- Move all loop marker adjustments from RMB to Shift + click actions
- Offer users a few options for loop marker adjustment via the settings menu
  - "Move closest" emulates FL Studio (current behavior on master)
  - "Handles" emulates Ableton: clicking the handle near the loop's edge (or the screen's edge if the full loop isn't visible) moves that point. Clicking the region between handles slides the entire loop.
  - "Dual-button" is custom, LMB sets start and RMB sets end.
- Display a custom cursor when shift is held in "Move closest" or "Handles" mode, indicating which handle will be moved on click
- Refactor position calculations in TimeLineWidget.
  - Instead of changing `m_xOffset` based on the playhead icon's width, and then undoing this change in tons of places (sometimes with magic numbers), let it always represent the width of the unused timeline region. The few calculations that need to adjust for pixmap width do so on their own instead.
  - Remove `m_posMarkerX`, because it's equivalent to `markerX(m_pos)`
  - Remove `m_moveXOff`, which is unnecessary now that `m_xOffset` isn't being modified
- More comments

TODO:
- Better handle drawing
  - Align with actual loop (but how? Too high when offset 0, too low when offset 1, 0.5 behaves like 0...)
- Basic RMB menu
- Proper settings layout


allejok96:
Yay! This was fast.

I suggest putting the Mode option in a radio list in a context menu of the timeline.

My guess is that "Closest mode people" will complain most about having to press shift and not so much about having to alternate between left and right mouse button. So since RMB is already killed by this design I think we can safely drop Closest mode in favor of Speedy mode.

To calm down the anti-shifters (and to educate people about speedy mode) we put in the context menu:
- Set start (shift+left)
- Set end (shift+right)

Handle mode TODO:
- Draw handles on the time line
- Use native cursors
- Middle handle for moving

Spekular:
Welp, I can't get a handle on this submodule situation. Will probably open yet another PR later this week instead of trying to fix this branch.

Spekular:
> I suggest putting the Mode option in a radio list in a context menu of the timeline.

I feel like this is more of a "set once and get used to it" setting.

> safely drop Closest mode

I'd rather keep it, to facilitate users moving to/from LMMS with one fewer workflow shock.

> Use native cursors

All the non-custom cursors look native to me on ubuntu & windows, and I don't think there are any native cursors to indicate "grab left" and "grab right".

allejok96:
> set once and get used to it

Not for the sake of it being quick to change, but for the sake of being discovered. You can almost expect users to fiddle around with the timeline when trying to set the loop marker, eventually opening the context menu. In there they would see the key combos for how to set the marker and also that there is a setting to select the mode they are most comfortable with. If you place it it settings dialog most users wouldn't even look for it and if they did come across it it's not immediately clear what the setting does. You'd also have to close the window to test the setting. 

> All the non-custom cursors look native to me

Hm, for me they are 3x the size of normal (on a regular DPI monitor). Yeah sadly Qt doesn't include a |< and >| cursor.

allejok96:
bild

Please forgive my great exaggeration. It's 50% bigger. This is all fine, just needs to be [aligned](https://doc.qt.io/qt-5/qcursor.html#hotSpot) right.

DomClark:
@Spekular I'd love to have this PR finished up and merged - are you still interested in working on it, or could someone else take it over?

Spekular:
> @Spekular I'd love to have this PR finished up and merged - are you still interested in working on it, or could someone else take it over?

Since I've been a bit busy and several conflicting PRs were merged while this stayed open, I unfortunately don't think I'll have time to get it back in a mergeable state myself. I'd be very thankful if someone wants to get it fixed up! (Then, maybe a follow-up PR could bring back the RMB menu from #6021? 😉 )

DomClark:
I've merged master into this branch (one of the more finicky merges I've done), and it appears to be working still after a quick test. I hope to be able to add the finishing touches soon. (I'm tempted to add the context menu in this PR, in order to implement allejok96's suggestion. It doesn't have to be mutually exclusive with a config dialog entry, and it would be a useful thing to have in place for future enhancements.)

DomClark:
Aside from resolving conflicts and some tidying up and formatting fixes, I have made the following changes:
* Fixed handle drawing so they align with the loop at all scaling factors
* Changed the left and right selection cursor hotspots to be vertically centred within the cursor image
* Changed the handle hitboxes to match the drawn handle width, rather than a fixed 5px width
* Fixed the shift modifier getting stuck down
* Added a label to the config dialog entry
* Made the config dialog options translatable and changed code not to rely on their English values
* Set the default edit mode to be dual-button, instead of having the loop be uneditable for unset/unknown modes
* Added a context menu with options to set loop points and change the loop edit mode

I believe that addresses all the TODOs in the PR description, allejok96's feedback, and a few other bugs. I chose "dual-button" as the default mode because it will be more familiar for users coming from 1.2, "closest" mode was generally unpopular, and both "closest" and "handle" mode can make it difficult to move loop points that are off-screen, but I am open to suggestions.

src/gui/editors/TimeLineWidget.cpp
```diff
@@ -60,22 +63,18 @@ TimeLineWidget::TimeLineWidget(const int xoff, const int yoff, const float ppb,
...
 	m_ppb( ppb ),
 	m_snapSize( 1.0 ),
 	m_pos( pos ),
 	m_timeline{&timeline},
 	m_begin( begin ),
 	m_mode( mode ),
 	m_hint( nullptr ),
-	m_action( Action::NoAction ),
-	m_moveXOff( 0 )
+	m_action( Action::NoAction )
```
Veratil, line 73:
```suggestion
	m_action(Action::NoAction)
```

src/gui/editors/TimeLineWidget.cpp
```diff
@@ -225,83 +212,138 @@ void TimeLineWidget::paintEvent( QPaintEvent * )
...
 	p.drawRect( innerRectangle );
+	
+	// Draw loop handles if necessary
+	const auto handleMode = ConfigManager::inst()->value("app", "loopmarkermode") == "handles";
+	const auto hw = std::min(m_loopHandleWidth, loopRectWidth / 2 - 1);
+	const auto leftHandle = QRectF(loopStart - .5, loopRectMargin - .5, hw, loopRectHeight);
+	const auto rightHandle = QRectF(loopEndR - hw - .5, loopRectMargin - .5, hw, loopRectHeight);
+	if (handleMode && underMouse() && QGuiApplication::keyboardModifiers().testFlag(Qt::ShiftModifier))
+	{
+		auto color = loopPointsActive ? m_activeLoopHandleColor : m_inactiveLoopHandleColor;
```
Veratil, line 233: Minor since it's a short lived variable, but would it be good practice to `const` this, too?

src/gui/editors/TimeLineWidget.cpp
```diff
@@ -225,83 +212,138 @@ void TimeLineWidget::paintEvent( QPaintEvent * )
...
 {
-	if( event->x() < m_xOffset )
-	{
-		return;
+	switch (action) {
+		case Action::MoveLoop: return Qt::SizeHorCursor;
+		case Action::MoveLoopBegin: return m_cursorSelectLeft;
+		case Action::MoveLoopEnd: return m_cursorSelectRight;
+		// Fall back to normal cursor if no action or action cursor not specified
+		default: return Qt::ArrowCursor;
```
Veratil, line 296: What do we think better practice would be: using `default:`, or no default with `return x` outside the switch?

src/gui/editors/TimeLineWidget.cpp
```diff
@@ -349,17 +390,34 @@ void TimeLineWidget::mouseMoveEvent( QMouseEvent* event )
 			update();
 			break;
 		}
-	case Action::SelectSongClip:
+		case Action::MoveLoop:
+		{
+			TimePos dragDelta = timeAtCursor - m_dragStartPos;
```
Veratil, line 395: Is this a `const`?

src/gui/editors/TimeLineWidget.cpp
DomClark: I'm not aware of a strong argument either way. I prefer using `default` here, as then it's consistent with the other cases, and has a reading that more closely matches the intent (it literally marks the default value with "default").

A couple of arguments I can think of in favour of putting it outside the switch are:
* Compilers can emit warnings if there is no default and a missing case, which helps catch errors introduced when extending the enum. However, this then requires all possible enumerators to have their own case, which is unnecessarily verbose. Also, I don't want to create a precedent for writing code to be read by tools over humans.
* It ensures that a value is still returned if a case breaks instead of returning. However, compilers will typically warn about this anyway, and in simple cases like this, it's difficult to miss such a mistake.

src/gui/editors/TimeLineWidget.cpp
DomClark: Absolutely - `const` all the things.

src/gui/editors/TimeLineWidget.cpp
DomClark: I considered making this change, but if I were to "fix" this line, I'd rather do so by moving the initialisation to the member declaration in the header, at which point that changes the line above (removing the comma), and then that needs fixing, and so on...

I can go ahead with this if you'd like, though.

Veratil: I'd be fine with removing the above line and initializing in the header.

src/gui/editors/TimeLineWidget.cpp
Veratil: Alright, then we can leave this to code readability preference.

src/gui/editors/TimeLineWidget.cpp
```diff
@@ -225,83 +197,138 @@ void TimeLineWidget::paintEvent( QPaintEvent * )
...
 	QRect innerRectangle = outerRectangle.adjusted( 1, 1, -1, -1 );
 	p.setPen( loopPointsActive ? getActiveLoopInnerColor() : getInactiveLoopInnerColor() );
 	p.setBrush( Qt::NoBrush );
 	p.drawRect( innerRectangle );
+	
+	// Draw loop handles if necessary
+	const auto handleMode = ConfigManager::inst()->value("app", "loopmarkermode") == "handles";
+	const auto hw = std::min(m_loopHandleWidth, loopRectWidth / 2 - 1);
+	const auto leftHandle = QRectF(loopStart - .5, loopRectMargin - .5, hw, loopRectHeight);
+	const auto rightHandle = QRectF(loopEndR - hw - .5, loopRectMargin - .5, hw, loopRectHeight);
```
Veratil, line 215: I think these can be moved into the if?

Veratil (approved):
LGTM

Veratil:
Tested MSVC build, all seems to be working 👍

src/gui/editors/TimeLineWidget.cpp
```diff
@@ -225,83 +197,139 @@ void TimeLineWidget::paintEvent( QPaintEvent * )
...
 		p.setOpacity(0.6);
-		p.drawPixmap(m_posMarkerX, height() - m_posMarkerPixmap.height(), m_posMarkerPixmap);
+		p.drawPixmap(markerX(m_pos) - (m_posMarkerPixmap.width() / 2),
+			height() - m_posMarkerPixmap.height(), m_posMarkerPixmap);
 	}
 }
 
+auto TimeLineWidget::getClickedTime(const int xPosition) const -> TimePos
+{
+	// How far into the timeline we clicked, measuring pixels from the leftmost part of the editor
```
sakertooth, line 237: Seems like this describes the overall function, so it would probably serve a better purpose as documentation instead of an internal comment.

src/gui/editors/TimeLineWidget.cpp
```diff
@@ -225,83 +197,139 @@ void TimeLineWidget::paintEvent( QPaintEvent * )
...
 		// Let the position marker extrude to the left
 		p.setClipping(false);
 		p.setOpacity(0.6);
-		p.drawPixmap(m_posMarkerX, height() - m_posMarkerPixmap.height(), m_posMarkerPixmap);
+		p.drawPixmap(markerX(m_pos) - (m_posMarkerPixmap.width() / 2),
+			height() - m_posMarkerPixmap.height(), m_posMarkerPixmap);
 	}
 }
 
+auto TimeLineWidget::getClickedTime(const int xPosition) const -> TimePos
```
sakertooth, line 235: Having the `const` in `const int xPosition` here is redundant.

src/gui/editors/TimeLineWidget.cpp
```diff
@@ -225,83 +197,139 @@ void TimeLineWidget::paintEvent( QPaintEvent * )
...
+	}
+}
 
-void TimeLineWidget::mousePressEvent( QMouseEvent* event )
+auto TimeLineWidget::actionCursor(Action action) const -> QCursor
 {
-	if( event->x() < m_xOffset )
-	{
-		return;
+	switch (action) {
```
sakertooth, line 277:
```suggestion
	switch (action)
	{
```

sakertooth (approved):
LGTM, tested and everything seems fine. Awesome feature. Just got a few suggestions code wise, but they are not blocking by any means.

src/gui/editors/TimeLineWidget.cpp
Spekular: It describes an intermediate value in pixels. Note thay the returned value is scaled by ppb and offset by the start position of the editor viewport.

Spekular:
Thanks for getting this fixed up and merged! 🎉

src/gui/editors/TimeLineWidget.cpp
sakertooth: Okay, so I happened to be right here. It's not that its inherently wrong to make `xPosition` `const`, [but its overly pedantic since function arguments that are passed by value are rarely ever mutated](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#S-const).

I didn't mention it because I actually didn't know about this for sure (just went off a feeling that this was off but couldn't tell why), so I had to dig through the guidelines. No action needed, but good to know in the future.

DomClark: My understanding of their comment is that code analysis tools should not emit a warning if const is omitted from function parameters, since this will likely produce a large number of warnings for little benefit. I don't think they mean that const is actually better omitted in those cases.

sakertooth: Yeah, "redundant" wasn't the best way to put it, because if it was, analysis tools would probably pick up that as a warning. I meant it more as in that it can confuse more than help.

Veratil: `Con.1` just says don't enforce the rule of const by default for function parameters.
```
Function parameters passed by value are rarely mutated,
but also rarely declared const. To avoid confusion and lots
of false positives, don’t enforce this rule for function
parameters.

    void g(const int i) { ... }  // pedantic

Note that a function parameter is a local variable so
changes to it are local.
```
The comment says pedantic, but that is just an opinion honestly. Nothing about this says we shouldn't declare function parameters const.

What we are doing though is self-documentation and declaring to anyone looking that `xPosition` is a value that shall not be changed within the function. Like it says, if something should not change, declare it `const`. For such a large codebase in my opinion it's better to be explicit.

sakertooth: Fair, I do like the mindset of "const by default, mutable when you need to". Parameters passed by value can also be seen local to a function, so if we already agree that we should declare local variables `const` that are within the body, they can be applied to function parameters too since the scope is the same.

Did more looking into this. Some say its useful, some say it isn't. It's quite common for me to not have strong opinions on these kind of things, usually leading me to consider a credible source (i.e., the guidelines) as a source of truth and absolute correctness. But then again, I shouldn't really treat it as such, and actually like a guideline because not everything in there I fully agree with anyways (I just feel like some of them miss cruical details).

It just so happened that I did not really have a strong opinion for this, but I agree personally that the communication between the API and its users becomes clearer using `const`. In any case though, its definitely not something worth discussing over in many cases in terms of enforcing it or not, but it did have me thinking.

Veratil: Conversations like this are always useful though. 😄

PR #6390, opened 2022-05-02
Implement sample caching

sakertooth:
The goal of this PR is to implement an LRU cache for the ``SampleBuffer`` class. The data within the cache will also be able to share ``SampleBuffer``'s between users, allowing for improved memory efficiency.

The refactoring that previously underwent this PR has been moved into another PR. This should make this PR quite simple and straight to the point.


Rossmaxx:
I think it would be better to do the sample workflow/performance improvements first and then sample buffer can be done as a different pr at a later time. This is just my opinion.

sakertooth:
> I think it would be better to do the sample workflow/performance improvements first and then sample buffer can be done as a different pr at a later time. This is just my opinion.

For the sample workflow to be improved, the `SampleBuffer`, quite frankly, needs to be changed. Improvements in `SampleBuffer` lead to an improvement in sample workflow. They go hand in hand. Its just that this PR is not solely for sample caching. Theres more to it (such as performance improvements when visualizing samples).

Rossmaxx:
Oh i see. Btw does this pr fix the abnormal memory / cpu usage on sample track.

sakertooth:
> Oh i see. Btw does this pr fix the abnormal memory / cpu usage on sample track.

Indeed. I put a milestone video that shows me putting 18 sample tracks with the memory hovering around ~170 MB. The sample is only loaded once and are shared after that *throughout* the project. It will not matter if you are in the AFP, a sample clip, etc. It will load it one time and share it using the cache. The cache doesnt make duplicates of the samples either.

In a previous iteration of that milestone video I played the sample tracks but that obviously was not ideal, so I just left it be.

Rossmaxx:
i tried out this build and found some issues.
1) the sample track colour seems broken
2) the theming is broken while using custom themes

https://youtu.be/4UYxu4BxuGg here is the video where i did the test

the sample track performance seemed to have improved.
will do further testing and tell you more.

sakertooth:
> i tried out this build and found some issues.
> 
> 1. the sample track colour seems broken
> 2. the theming is broken while using custom themes
> 
> https://youtu.be/4UYxu4BxuGg here is the video where i did the test
> 
> the sample track performance seemed to have improved. will do further testing and tell you more.

Visualization of samples is on the back-burner for now. I will get to it soon though (unless somebody else would like to contribute).
I'm looking towards changing the playing of samples up a bit and making resampling faster (resampling takes time).

Spekular:
Why are you setting the minimum length of a clip to one bar again? This was intentionally changed in #4933 to allow e.g. the use of sample tracks for percussion. With sample caching implemented this would become even more feasible, but setting samples to a bar long makes it less so.

sakertooth:
> Why are you setting the minimum length of a clip to one bar again? This was intentionally changed in #4933 to allow e.g. the use of sample tracks for percussion. With sample caching implemented this would become even more feasible, but setting samples to a bar long makes it less so.

It was just that when I was importing certain samples of a small duration, it would make it seem as if it did not import at all. Reverting.

sakertooth:
The ambiguity between sample sharing and sample caching is somewhat problematic. While the samples are infact shared (no duplicates of the same ``SampleBufferV2`` derived from a file or Base64 are made), key entries of ``std::weak_ptr`` are cached, which come from the first ``std::shared_ptr`` using that ``SampleBufferV2``. Sharing comes from the usage of ``std::weak_ptr::lock``. 

Ergo, I think I will just stick to sample caching. Ultimately, the entries are cached, but the sample data itself is shared. Plus, I cannot think of a better fitting name for ``SampleBufferCache``.

PhysSong:
It seems like only `std::experimental::filesystem` is available(in `<experimental/filesystem>`) prior to GCC 8 on Linux and XCode 11 on macOS.
Also, on Windows, `std::filesystem::path::value_type` is `wchar_t`, so you will have to use `sf_wchar_open` on Windows(tested by `#ifdef LMMS_BUILD_WIN32`).

src/core/SampleBufferV2.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+void SampleBufferV2::loadFromDrumSynthFile(const std::filesystem::path& drumSynthFilePath)
+{
+	auto dsFilePathStr = drumSynthFilePath.native();
+	auto ds = DrumSynth();
+	auto samples = std::make_unique<int16_t>();
+	auto samplesRawPtr = samples.get();
+
+	// TODO: Remove QString::fromStdString when Qt is removed from DrumSynth
+	int numSamples = ds.GetDSFileSamples(QString::fromStdString(dsFilePathStr), samplesRawPtr,
```
PhysSong, line 195: I've changed `DrumSynth` to use `QFile` because `std::fstream` was unable to handle Unicode paths correctly on Windows, prior to C++17. It's possible to revert to `std::fstream` now, but that will conflict with #6078, so I'm not sure what is the best way to do this.

On Windows, `QString::fromStdWString` should be used instead, if we're going to keep using `QString` here.

src/core/SampleBufferV2.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+void SampleBufferV2::loadFromAudioFile(const std::filesystem::path& audioFilePath)
+{
+	SF_INFO sfInfo;
+	sfInfo.format = 0;
+
+	auto sndFileDeleter = [](SNDFILE* ptr) { sf_close(ptr); };
+	#ifdef LMMS_BUILD_WIN32
+		auto sndFile = std::unique_ptr<SNDFILE, decltype(sndFileDeleter)>(
+			sf_wchar_open(audioFilePath.c_str(), SFM_READ, &sfInfo), sndFileDeleter);
```
PhysSong, line 153: I forgot to mention that you have to define `ENABLE_SNDFILE_WINDOWS_PROTOTYPES` as 1 before including `sndfile.h` prior to libsndfile 1.1.0, in order to use `sf_wchar_open`.

Rossmaxx: Currently, some ogg files crash which Saker says has something to do with sndfile. Does this one fix it?

sakertooth: > I forgot to mention that you have to define `ENABLE_SNDFILE_WINDOWS_PROTOTYPES` as 1 before including `sndfile.h` prior to libsndfile 1.1.0, in order to use `sf_wchar_open`.

Since the plan is to use FFmpeg instead of sndfile, I'll just make the switch once it's added into the codebase.

src/core/SampleBufferV2.cpp
sakertooth: I am not sure what to do about this either. We are in the midst of removing Qt from the core, so I don't feel comfortable using ``QString`` in general. However, with the #6078 conflict you shown above, I am at a loss for course of action except waiting untill things are more settled.

sakertooth: I have added ``QString`` to and from ``std::experimental::filesystem::path`` conversion functions within ``SampleBufferV2``. That way, if a file path needs to be a ``QString``, the conversion can occur respecting portability with Windows and Unix. Ideally, these functions wouldn't be a part of the core, but I wasn't sure where to put them but in ``SampleBufferV2``.

src/core/SampleBufferV2.cpp
sakertooth: Just as a follow up to this: The plan is still to use FFmpeg, but for now sndfile is still being used. I'll check if the problem still remains but my course of immediate action against this bug is unknown.

PhysSong:
Note for the failure with MinGW: https://stackoverflow.com/a/33159746

sakertooth:
> Note for the failure with MinGW: https://stackoverflow.com/a/33159746

I think it's because the experimental ``filesystem`` implementation doesn't support the generic format observer functions for MinGW (GCC 7.3.0). The linux builds only have support for the experimental ``filesystem`` as well, but it doesn't fail and is GCC 7.5.0. So I'm pretty sure I have linked ``stdc++fs``, but I'm guessing GCC 7.3.0 doesn't support the generic format observers in its experimental ``filesystem`` implementation. Could be very wrong here, but I'm not sure why GCC 7.5.0 would be working (Linux GHA) and not GCC 7.3.0 (MinGW GHA).

MinGW had errors with not being able to find ``sf_wchar_open`` (I'm assuming because it's a Unix-like environment, even if the target is Windows), so I figured that using the generic format within ``sf_open`` worked well. If I can get MinGW to work with ``sf_wchar_open``, I might be able to get something to work. It's just very odd since the plan is to use FFMpeg in the future anyways, but I still want to get something that builds and functions at the very least.

src/core/SampleBufferV2.cpp
sakertooth: Hopefully my final update on this: I decided to keep the file paths as a ``QString``, just because it makes interfacing with the already existing Qt interface a lot easier. We can always switch when we decide to fully remove ``QString`` and are ready to use ``std::filesystem::path``. Feel free to resolve this if that's okay.

src/core/SampleBufferV2.cpp
sakertooth: Instead of ``sf_open`` and ``sf_wchar_open``, I went ahead and used sf_open_fd with ``QFile`` to simplify the handling of different file path formats on varying platforms, such as Windows.

sakertooth:
Closing in favor of #6703

PR #6399, opened 2022-05-08
Make AFP include the last sample frame

husamalhomsi:
Now AFP can graph and play the last sample frame.

I tested with the following [perfect loops](https://github.com/LMMS/lmms/files/8646530/perfectLoops.zip):

- 8frames.wav: now graphed with 8 points instead of 7, and there's no DC offset when looping it.
- 44100frames.wav: now labeled with a duration of 1000 ms instead of 999 ms, and there's no click when looping it.

This change could be problematic for existing projects that use AFP to loop tiny samples like 8frames.wav above.

### DIFF
--- a/plugins/AudioFileProcessor/AudioFileProcessor.cpp
+++ b/plugins/AudioFileProcessor/AudioFileProcessor.cpp
@@ -428,9 +428,9 @@ void AudioFileProcessor::loopPointChanged( void )
 
 void AudioFileProcessor::pointChanged( void )
 {
-	const f_cnt_t f_start = static_cast<f_cnt_t>( m_startPointModel.value() *	( m_sampleBuffer.frames() ) );
-	const f_cnt_t f_end = static_cast<f_cnt_t>( m_endPointModel.value() * ( m_sampleBuffer.frames() ) );
-	const f_cnt_t f_loop = static_cast<f_cnt_t>( m_loopPointModel.value() * ( m_sampleBuffer.frames() ) );
+	const f_cnt_t f_start = static_cast<f_cnt_t>( m_startPointModel.value() *	m_sampleBuffer.frames() );
+	const f_cnt_t f_end = static_cast<f_cnt_t>( m_endPointModel.value() * m_sampleBuffer.frames() );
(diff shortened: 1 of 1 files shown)


plugins/AudioFileProcessor/AudioFileProcessor.cpp
```diff
@@ -428,9 +428,9 @@ void AudioFileProcessor::loopPointChanged( void )
 
 void AudioFileProcessor::pointChanged( void )
 {
-	const f_cnt_t f_start = static_cast<f_cnt_t>( m_startPointModel.value() *	( m_sampleBuffer.frames()-1 ) );
-	const f_cnt_t f_end = static_cast<f_cnt_t>( m_endPointModel.value() * ( m_sampleBuffer.frames()-1 ) );
-	const f_cnt_t f_loop = static_cast<f_cnt_t>( m_loopPointModel.value() * ( m_sampleBuffer.frames()-1 ) );
+	const f_cnt_t f_start = static_cast<f_cnt_t>( m_startPointModel.value() *	( m_sampleBuffer.frames() ) );
+	const f_cnt_t f_end = static_cast<f_cnt_t>( m_endPointModel.value() * ( m_sampleBuffer.frames() ) );
+	const f_cnt_t f_loop = static_cast<f_cnt_t>( m_loopPointModel.value() * ( m_sampleBuffer.frames() ) );
```
PhysSong, line 433:
```suggestion
	const f_cnt_t f_start = static_cast<f_cnt_t>(m_startPointModel.value() * m_sampleBuffer.frames());
	const f_cnt_t f_end = static_cast<f_cnt_t>(m_endPointModel.value() * m_sampleBuffer.frames());
	const f_cnt_t f_loop = static_cast<f_cnt_t>(m_loopPointModel.value() * m_sampleBuffer.frames());
```

Sawuare: To minimize my style changes, I'll remove the redundant paraentheses but keep the extra spaces and tab. Is this okay?

PhysSong: I think you don't have to minimize style changes here because they are already in the line you're touching, but I will leave the decision up to you.

PR #6404, opened 2022-05-10
Refactor shared memory

DomClark:
Currently, every location where shared memory is used has `#ifdef`s to use either Qt or System V shared memory. This leads to a lot of duplicated, low-level code, and makes developing these parts more difficult as one needs to consider multiple platforms when doing so. This PR is an effort to extract the low-level shared memory code into a simple higher-level wrapper.

Features of the wrapper class:
* Uses RAII so shared memory is automatically detached or freed when necessary.
* Templated to provides typed access, automatically map const objects as read-only, and ensure the correct size is allocated.
* Smart-pointer-like interface for objects and arrays.
* Uses more modern POSIX shared memory instead of System V.
* Failure to create or attach shared memory throws an exception, which is a more appropriate way to handle potential runtime errors than an assertion, and more idiomatic C++ than testing validity after performing an operation.

Other changes:
* Dynamic shared memory keys are now chosen as UUIDs, rather than trying numerical keys in sequence until an unused one is found.
* Various names have been changed from something like `m_shm` to something like `m_audioBuffer`, to reflect the use of the memory rather than its type.

The API for creating/attaching shared memory is to pass the key to `create`/`attach` member functions on the wrapper, similar to `QSharedMemory`'s API. Other options I considered were to pass the key and a `create_shm`/`attach_shm` tag type to the constructor, or to use maker functions such as `createSharedMemory`/`attachSharedMemory`. If one of these sounds better, or you have another idea, of course mention it.

I've tested this on Windows and Linux Mint, wider testing appreciated of course.


DomClark:
The MSVC build is hitting [this compiler bug](https://developercommunity.visualstudio.com/t/struct-with-a-volatile-member-is-no-longer-a-pod/838625), which is fixed in v19.27 of the compiler, corresponding to Visual Studio 2019 version 16.7. I can remove the static assertion, or we can upgrade the compiler version in CI.

The macOS build is failing because shared memory functions there don't require linking with `-lrt`. I don't know what the situation is with other platforms that people build LMMS on, so I'm not sure whether to make this conditional on "is Linux", or "is not Apple".

PhysSong:
> I'm not sure whether to make this conditional on "is Linux", or "is not Apple".

Unfortunately, neither of them will work. OpenBSD doesn't have `librt` like macOS, but FreeBSD have `librt`.

DomClark:
I fixed the MSVC build by removing the `volatile` qualifiers from the `shmFifo::shmData` struct members. They weren't applied consistently anyway, and we were using them in a way that is deprecated in C++20.

For deciding when to link to `librt`, I tried using `find_library`, but that didn't work with the Wine VST hosts. I'm now using `check_library_exists` and passing the result to the remote plugin external projects, which seems to work.

The macOS build failed for a second reason - we were building GIG Player as C++98 when using Clang, because of some problem that used to exist in `gig.h`. I don't know what the problem was, but I've removed that logic and it builds successfully, so I guess it got fixed at some point.

DomClark:
I fixed the MSVC build by removing the `volatile` qualifiers from the `shmFifo::shmData` struct members. They weren't applied consistently anyway, and we were using them in a way that is deprecated in C++20.

For deciding when to link to `librt`, I tried using `find_library`, but that didn't work with the Wine VST hosts. I'm now using `check_library_exists` and passing the result to the remote plugin external projects, which seems to work.

The macOS build failed for a second reason - we were building GIG Player as C++98 when using Clang, because of some problem that used to exist in `gig.h`. I don't know what the problem was, but I've removed that logic and it builds successfully, so I guess it got fixed at some point.

src/common/SharedMemory.cpp
```diff
@@ -0,0 +1,230 @@
...
+	m_ptr{m_impl->get()}
+{ }
+
+SharedMemoryData::SharedMemoryData(std::string&& key, std::size_t size, bool readOnly) :
+	m_key{std::move(key)},
+	m_impl{std::make_unique<SharedMemoryImpl>(m_key, std::max(size, std::size_t{1}), readOnly)},
+	m_ptr{m_impl->get()}
+{ }
+
+SharedMemoryData::~SharedMemoryData() { }
```
sakertooth, line 222: I see no use in defining a empty deconstructor. I think it is best to let it be implicity defined, which is of course empty.
```suggestion
```

include/SharedMemory.h
```diff
@@ -0,0 +1,143 @@
...
+
+class SharedMemoryImpl;
+
+class SharedMemoryData
+{
+public:
+	SharedMemoryData() noexcept;
+	SharedMemoryData(std::string&& key, bool readOnly);
+	SharedMemoryData(std::string&& key, std::size_t size, bool readOnly);
+	~SharedMemoryData();
```
sakertooth, line 41: This deconstuctor does nothing given the implementation. It can be implicity defined.
```suggestion
```

include/SharedMemory.h
```diff
@@ -0,0 +1,143 @@
...
+
+	void attach(std::string key)
+	{
+		m_data = detail::SharedMemoryData{std::move(key), std::is_const_v<T>};
+	}
+
+	void create(std::string key)
+	{
+		m_data = detail::SharedMemoryData{std::move(key), sizeof(T), std::is_const_v<T>};
+	}
```
sakertooth, line 89: Wouldn't this create a copy of ``std::string`` and then move with that copy? Why not take in a ``std::string&&`` instead. Callers could then use ``std::move(key)``.
```suggestion
	void attach(std::string&& key)
	{
		m_data = detail::SharedMemoryData{key, std::is_const_v<T>};
	}

	void create(std::string&& key)
	{
		m_data = detail::SharedMemoryData{key, sizeof(T), std::is_const_v<T>};
	}
```

include/SharedMemory.h
```diff
@@ -0,0 +1,143 @@
...
+
+	void attach(std::string key)
+	{
+		m_data = detail::SharedMemoryData{std::move(key), std::is_const_v<T>};
+	}
+
+	void create(std::string key, std::size_t size)
+	{
+		m_data = detail::SharedMemoryData{std::move(key), size * sizeof(T), std::is_const_v<T>};
+	}
```
sakertooth, line 126:
```suggestion
	void attach(std::string&& key)
	{
		m_data = detail::SharedMemoryData{key, std::is_const_v<T>};
	}

	void create(std::string&& key, std::size_t size)
	{
		m_data = detail::SharedMemoryData{key, size * sizeof(T), std::is_const_v<T>};
	}
```

src/core/RemotePlugin.cpp
```diff
@@ -481,37 +470,10 @@ void RemotePlugin::hideUI()
...
-	{
-	}
-
-	m_shm = (float *) shmat( m_shmID, 0, 0 );
-#endif
-	m_shmSize = s;
-	sendMessage( message( IdChangeSharedMemoryKey ).
-				addInt( shm_key ).addInt( m_shmSize ) );
+	const size_t s = (m_inputCount + m_outputCount) * Engine::audioEngine()->framesPerPeriod();
+	m_audioBuffer.create(QUuid::createUuid().toString().toStdString(), s);
```
PhysSong, line 474: What if this call throws due to very unlikely failures?

src/common/SharedMemory.cpp
```diff
@@ -0,0 +1,230 @@
...
+	do
+	{
+		result = function();
+	}
+	while (result == -1 && errno == EINTR);
+	return result;
+}
+
+template<typename T, T Null>
+class NullableResource
```
PhysSong, line 65: This class looks like a good candidate to be contained in a separate header file for future uses.

src/common/SharedMemory.cpp
```diff
@@ -0,0 +1,230 @@
...
+	m_key{std::move(key)},
+	m_impl{std::make_unique<SharedMemoryImpl>(m_key, std::max(size, std::size_t{1}), readOnly)},
+	m_ptr{m_impl->get()}
+{ }
+
+SharedMemoryData::~SharedMemoryData() { }
+
+SharedMemoryData::SharedMemoryData(SharedMemoryData&& other) noexcept :
+	m_key{std::move(other.m_key)},
+	m_impl{std::move(other.m_impl)},
```
sakertooth, line 226: Consistency.
```suggestion
	m_key{std::exchange(other.m_key, "")},
	m_impl{std::exchange(other.m_impl, nullptr)},
```

include/SharedMemory.h
sakertooth: I also find it nicer to put the definitions in the cpp file, no matter the size. It makes the header file easier to read. Up to you though.

sakertooth (changes requested):
I see a couple of changes that can be made. The code overall looks good and modern however.

src/core/RemotePlugin.cpp
DomClark: The existing code didn't handle this very well (it just retried in a loop, assuming the only error would be key collisions, which shouldn't happen now with UUIDs), so I didn't bother doing it better. I'll change it to at least catch and log any exception thrown here.

src/common/SharedMemory.cpp
DomClark: Good idea, will move it to a new header `include/RaiiHelpers.h`.

src/common/SharedMemory.cpp
DomClark: This is only an internal helper class (hence it's in the `detail` namespace), so I don't really care about it having a nice moved-from state. Instead I think I'll change `m_ptr` just to copy without updating `other` (the current exchange is left over from an older design). Thanks for pointing this out!

include/SharedMemory.h
```diff
@@ -0,0 +1,143 @@
...
+#include <type_traits>
+
+namespace detail {
+
+class SharedMemoryImpl;
+
+class SharedMemoryData
+{
+public:
+	SharedMemoryData() noexcept;
```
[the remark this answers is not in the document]
sakertooth: ``But SharedMemoryImpl is an incomplete type at this point, so the destructor of std::unique_ptr<SharedMemoryImpl> cannot be called.``
Makes sense. Should've probably done some testing on my local machine before submitting this.

include/SharedMemory.h
```diff
@@ -0,0 +1,143 @@
...
+		using std::swap;
+		swap(a.m_key, b.m_key);
+		swap(a.m_impl, b.m_impl);
+		swap(a.m_ptr, b.m_ptr);
+	}
+
+	const std::string& key() const noexcept { return m_key; }
+	void* get() const noexcept { return m_ptr; }
+
+private:
```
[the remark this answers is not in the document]
sakertooth: > It would.
> 
> [class.base.init]/9:
> 
> > In a non-delegating constructor, if a given potentially constructed subobject is not designated by a _mem-initializer-id_ [...] the entity is default-initialized.
> 
> [dcl.init]/7:
> 
> > To _default-initialize_ an object of type `T` means:
> > 
> > * If `T` is a (possibly cv-qualified) class type, constructors are considered. The applicable constructors are enumerated ([over.match.ctor]), and the best one for the _initializer_ `()` is chosen through overload resolution. The constructor thus selected is called, with an empty argument list, to initialize the object.
> > * If `T` is an array type, each element is default-initialized.
> > * Otherwise, no initialization is performed.
> 
> `std::string` and `std::unique_ptr` have default constructors that initialise them to the empty string and `nullptr` respectively, which is the desired behaviour here (and indeed equivalent to your suggestion). On the other hand, `void*` is not a class type (or an array type), so no initialisation is performed by default-initialization and thus the explicit initialisation is required.

Ah, okay. I usually strive for just a consistent look across the board, so initalizing only ``m_ptr`` seemed unusal to me at first.

include/SharedMemory.h
sakertooth: For that last point, I was able to separate the function declaration and definition so I assumed that it could be done. I now realize that this cannot be done in separate translation units.

Issue #1215, opened 2014-10-15
Automatic FX-mixer strip management


unfa:
**The problem**
Right now, to use the FX-mixer the user has to manually add, assign and rename strips, before he can use the mixer. I've got used to do this, and I use minimal instrument names ("K" for "Kick" etc.) to cut down the time I spend just getting the mixer up and running. I think many operations can be automated and the user could just use the mixer creatively, not wasting time on setting it up every time from scratch.

**The solution**
The idea is to create, assign and rename one FX-mixer channel for every instrument the user adds. Since we have two instrument spaces (B+B and Song) I advise to make it first B+B instruments, then the Song instruments. Possibly we could use the colors to indicate the type of mixers strips (see #1214).

All strips should have an "automatic" flag (a boolean property). None of the automatic functions below should work if the strip has the "automatic" flag FALSE (manually created or modified strips should be kept intact).
1. When the user adds an instrument, create an FX-mixer strip and connect it with the created instrument; set the "auto" flag TRUE.
   If the instrument is added into B+B editor, scan the mixer to find the first Song instrument strip, and insert the new strip before it (shift all other strips). If the instrument is added into Song editor, add it at the end of the mixer.
2. When the user renames an instrument, rename the FX-mixer channel that it's connected to;
3. When the user moves an instrument, move the FX-mixer channel accordingly so the orders match in Sequencers and Mixer;
4. If the user deletes an instrument, delete the corresponding channel strip.
5. If the user creates or modifies (rename, move) a mixer strip, set the "automatic" flag to FALSE; Muting, moving the fader, modifying strips sends, and modifying the effects stack should not remove the "automatic" flag.
6. Let the user right click on a strip and set/unset the "automatic" flag so he can manually re-enable auto or disable it if he wants.
7. Let the user disable/enable this system in the options immediately (no restart required) possibly with a on/off button in the FX-mixer window.
8. If the system is enabled while the session is not empty, flag all channels as "manual" and only manage channels for new instruments.
9. Mark the automatically created and managed channels visually in the FX-mixer window, so the user knows what is what.

tresf:
@unfa can we close this out in lieu of #1572 which offers "Assign to new FX channel" option?

badosu:
@tresf I don't guess so, @unfa's request is an umbrella that encompasses many issues related to the workflow related to mixer management.

PR #6407, opened 2022-05-15
Automatic FX-mixer strip management #1215

spechtstatt:
A very first draft of an automatic mixer strip management as proposed by @unfa  which allows an automatic linking of tracks to fx mixer strips. 

Works so far for the song editor - bbeditor not tested yet and also the automatic sorting is not implemented yet.

AutoFXFeature_Draft01


spechtstatt:
Would be great to get some feedback about the functionality in general and also regarding my implementation approach. I tried to move some functionality to the FxMixer and use the FxMixerView as a kind of entry point for the track operations. 

If this makes sense I can work on resolving the conflicts.

spechtstatt:
### Open Tasks

- [ ] rebase 
- [ ] make buildable again
- [ ] test bbeditor 
- [ ] automatic sorting 
- [ ] add global config switch
- [ ] button for enabling/disabling the functionilty in the mixer

spechtstatt:
It would also be possible in the course of this PR to provide some additional information directly inside the mixer as it is proposed by #1214 because this PR is collecting this information anyway.

Like e.g. count of connections to the mixer channel or information about the area (pattern vs. song). Auto linked channels are currently identified by changing the text color to green. 

Maybe someone has some ideas for presenting all this information inside the UI.

src/gui/MixerView.cpp
```diff
@@ -185,11 +272,61 @@ int MixerView::addNewChannel()
...
+		{
+			list->append(i);
+		}
+	}
+
+	// add auto tracks in the order of the song tracks
+	mix->processAssignedTracks([&, list](Track * track, IntModel * model, MixerChannel * channel)
+	mutable {
+		(void) track;		
+		if (channel == NULL) return;
```
sakertooth, line 300:
```suggestion
		if (channel == nullptr) return;		
```

src/gui/MixerView.cpp
```diff
@@ -419,79 +547,86 @@ void MixerView::deleteChannel(int index)
...
+
+
+void MixerView::toggleAutoTrackLink(int index)
+{
+	Mixer * mix = Engine::mixer();
+	mix->toggleAutoTrackLink(index);
+	MixerChannel *  channel = mix->mixerChannel(index);
+	if (!channel->m_autoTrackLinkModel.value()) return;
+
+	Track * trackFound = NULL;
```
sakertooth, line 581:
```suggestion
	Track * trackFound = nullptr;
```

src/gui/MixerView.cpp
```diff
@@ -419,79 +547,86 @@ void MixerView::deleteChannel(int index)
...
-			channel = strack->mixerChannelModel()->value();
+			trackFound = track;
 		}
-		inUse[channel] = true;
-	}
+	});
 
-	//Check all channels except master, delete those with no incoming sends
-	for(int i = m_mixerChannelViews.size()-1; i > 0; --i)
+	if (trackFound != NULL)
```
sakertooth, line 591:
```suggestion
	if (trackFound != nullptr)
```

src/gui/MixerView.cpp
```diff
@@ -170,6 +170,93 @@ MixerView::~MixerView()
...
+
+		setCurrentMixerLine( channelIndex );
+	}
+}
+
+void MixerView::updateAfterTrackStyleModify(Track * track)
+{
+	Mixer * mix = Engine::mixer();
+	IntModel * model = mix->getChannelModelByTrack(track);
+	if (model != NULL)
```
sakertooth, line 194:
```suggestion
	if (model != nullptr)
```

src/gui/MixerView.cpp
```diff
@@ -170,6 +170,93 @@ MixerView::~MixerView()
...
+			setCurrentMixerLine(index);
+		}
+	}
+}
+
+void MixerView::updateAfterTrackMixerLineModify(Track * track)
+{
+	Mixer * mix = Engine::mixer();
+	IntModel * model = mix->getChannelModelByTrack(track);
+	if (model != NULL)
```
sakertooth, line 211:
```suggestion
	if (model != nullptr)
```

src/gui/MixerView.cpp
```diff
@@ -170,6 +170,93 @@ MixerView::~MixerView()
...
+		}
+		if (needUpdate) updateAutoTrackSortOrder();
+	}
+}
+
+void MixerView::updateAfterTrackMove(Track * track)
+{
+	Mixer * mix = Engine::mixer();
+	IntModel * model = Engine::mixer()->getChannelModelByTrack(track);
+	if (model != NULL)
```
sakertooth, line 234:
```suggestion
	if (model != nullptr)
```

src/gui/MixerView.cpp
```diff
@@ -170,6 +170,93 @@ MixerView::~MixerView()
...
+			updateAutoTrackSortOrder();
+		}
+	}
+}
+
+void MixerView::updateAfterTrackDelete(Track * track)
+{
+	Mixer * mix = Engine::mixer();
+	IntModel * model = mix->getChannelModelByTrack(track);
+	if ( model != NULL)
```
sakertooth, line 248:
```suggestion
	if ( model != nullptr)
```

src/gui/MixerView.cpp
```diff
@@ -185,11 +272,61 @@ int MixerView::addNewChannel()
...
 	updateMaxChannelSelector();
 
 	return newChannelIndex;
 }
 
+void MixerView::updateAutoTrackSortOrder()
+{
+
+	Mixer * mix = Engine::mixer();
+	QList<int> *list = new QList<int>();
```
sakertooth, line 285: Should this really be allocated on the heap? A simple ``std::vector`` on the stack would work fine here, no?

src/gui/MixerView.cpp
```diff
@@ -185,11 +272,61 @@ int MixerView::addNewChannel()
...
+		{
+			list->append(model->value());
+		}
+	});
+	return;
+
+
+	// bubblesort here because the list is normally almost ordered
+	int n = list->length();
+	bool swapped = false;
```
sakertooth, line 311: ``swapped`` is not reassigned any value in this do-while loop. Wouldn't this just run once? Unless the intent is something different, I think this can be simplified.

src/core/Mixer.cpp
```diff
@@ -368,74 +348,122 @@ void Mixer::deleteChannel( int index )
...
+		}
+	}
+}
+
+std::vector<int> Mixer::getUsedChannelCounts()
+{
+	std::vector<int> used(m_mixerChannels.size(), 0);
+	processAssignedTracks([&used](Track * track, IntModel * model, MixerChannel * channel)
+	mutable {
+		(void) track;
```
sakertooth, line 407: Is this necessary? I see you've done something like ``(void) <parameter>`` in other places as well, even when the parameter in question isn't used by the lambda.

src/core/Mixer.cpp
```diff
@@ -368,74 +348,122 @@ void Mixer::deleteChannel( int index )
...
+		return strk->mixerChannelModel();
+	}
+	return NULL;
+}
+
+bool Mixer::isAutoTrackLinkToggleAllowed(int index)
+{
+	if (mixerChannel( index )->m_autoTrackLinkModel.value()) return true;
 
+	std::vector<int> usedChannelCounts = getUsedChannelCounts();
```
sakertooth, line 375:
```suggestion
```

src/core/Mixer.cpp
```diff
@@ -368,74 +348,122 @@ void Mixer::deleteChannel( int index )
...
+	return usedChannelCounts[index] == 1;
```
sakertooth, line 376:
```suggestion
	return getUsedChannelCounts()[index] == 1;
```

sakertooth: I find this to be more compact. Up to you if you decide to make the change.

src/core/Mixer.cpp
```diff
@@ -368,74 +348,122 @@ void Mixer::deleteChannel( int index )
...
 }
 
+void Mixer::toggleAutoTrackLink(int index)
+{
+	m_mixerChannels[index]->m_autoTrackLinkModel.setValue(! m_mixerChannels[index]->m_autoTrackLinkModel.value());
+}
+
+IntModel * Mixer::getChannelModelByTrack(Track * track)
+{
+	if( track->type() == Track::InstrumentTrack )
```
sakertooth, line 358: You can also use a switch-case.

src/core/Mixer.cpp
```diff
@@ -285,46 +286,25 @@ void Mixer::deleteChannel( int index )
...
-	// go through every instrument and adjust for the channel index change
-	TrackContainer::TrackList tracks;
-	tracks += Engine::getSong()->tracks();
-	tracks += Engine::patternStore()->tracks();
 
-	for( Track* t : tracks )
+	processAssignedTracks( [index](Track * track, IntModel * model, MixerChannel * channel)
 	{
-		if( t->type() == Track::InstrumentTrack )
+		(void) track;
```
sakertooth, line 292: Why is this here? I'm not too sure of its purpose.

src/gui/MixerView.cpp
```diff
@@ -419,79 +547,86 @@ void MixerView::deleteChannel(int index)
...
-	//Populate inUse by checking the destination channel for every track
-	for (Track* t: tracks)
+	//Check all channels except master, delete those with no incoming sends
+	for(int i = m_mixerChannelViews.size()-1; i > 0; --i)
 	{
-		//The channel that this track sends to. Since master channel is always in use,
-		//setting this to 0 is a safe default (for tracks that don't sent to the mixer).
-		int channel = 0;
-		if (t->type() == Track::InstrumentTrack)
+		if ((inUse[i]==0) && Engine::mixer()->mixerChannel(i)->m_receives.isEmpty())
```
sakertooth, line 564:
```suggestion
		if ((inUse[i]==0) && mix->mixerChannel(i)->m_receives.isEmpty())
```

src/core/Mixer.cpp
```diff
@@ -285,46 +286,25 @@ void Mixer::deleteChannel( int index )
...
+		(void) channel;
```
sakertooth, line 293: Same with this one.

src/gui/MixerView.cpp
```diff
@@ -170,6 +170,93 @@ MixerView::~MixerView()
...
+	Mixer * mix = Engine::mixer();
+	IntModel * model = mix->getChannelModelByTrack(track);
+	if (model != NULL)
+	{
+		// check if there are more than one track pointing to the same mixer channel
+		// if yes disable the autotracklink
+		std::vector<bool> used(m_mixerChannelViews.size(), false);
+		bool needUpdate = false;
+		std::vector<int> usedChannelCounts = mix->getUsedChannelCounts();
+		for(unsigned long i = 0; i< usedChannelCounts.size();i++)
```
sakertooth, line 218:
```suggestion
		for(unsigned long i = 0; i < usedChannelCounts.size(); i++)
```

src/gui/widgets/MixerLine.cpp
```diff
@@ -244,14 +247,31 @@ void MixerLine::mouseDoubleClickEvent( QMouseEvent * )
 
 void MixerLine::contextMenuEvent( QContextMenuEvent * )
 {
-	QPointer<CaptionMenu> contextMenu = new CaptionMenu( Engine::mixer()->mixerChannel( m_channelIndex )->m_name, this );
+	Mixer * mix =  Engine::mixer();
+	QPointer<CaptionMenu> contextMenu = new CaptionMenu( mix->mixerChannel( m_channelIndex )->m_name, this );
```
sakertooth, line 251: Why not just:
```suggestion
	CaptionMenu* contextMenu = new CaptionMenu( mix->mixerChannel( m_channelIndex )->m_name, this );
```

src/gui/widgets/MixerLine.cpp
```diff
@@ -262,21 +282,44 @@ void MixerLine::contextMenuEvent( QContextMenuEvent * )
...
 	QMenu colorMenu(tr("Color"), this);
-	colorMenu.setIcon(embed::getIconPixmap("colorize"));
-	colorMenu.addAction(tr("Change"), this, SLOT(selectColor()));
-	colorMenu.addAction(tr("Reset"), this, SLOT(resetColor()));
-	colorMenu.addAction(tr("Pick random"), this, SLOT(randomizeColor()));
-	contextMenu->addMenu(&colorMenu);
+	if (!autoTrackLink)
+	{
+		colorMenu.setIcon(embed::getIconPixmap("colorize"));
+		colorMenu.addAction(tr("Change"), this, SLOT(selectColor()));
```
sakertooth, line 288: I find ``SLOT()`` to be redundant mainly because any function can act as a slot now. If you want to be consistent with the current codebase at the moment, then I would say it is not that big of an issue.

https://doc.qt.io/qt-5/signalsandslots.html#:~:text=You%20can%20also%20connect%20to%20functors%20or%20C%2B%2B11%20lambdas%3A

sakertooth (changes requested):
Just a few changes and a quick review from me. I had some questions as to what certain lines were doing, especially the ``(void) track`` and ``(void) channel`` ones.

src/gui/MixerView.cpp
```diff
@@ -185,11 +272,61 @@ int MixerView::addNewChannel()
...
+		if (channel == NULL) return;		
+		if (channel->m_autoTrackLinkModel.value())
+		{
+			list->append(model->value());
+		}
+	});
+	return;
+
+
+	// bubblesort here because the list is normally almost ordered
```
Veratil, line 309: why not `std::sort` ?

spechtstatt: Because I am C++ beginner :-)

I will look into it.

src/gui/MixerView.cpp
spechtstatt: The sorting approach is not working out as I thought - therefore the return - but I had no time yet to think again about it.

src/core/Mixer.cpp
spechtstatt: I just wanted to avoid the warning and signal that the parameter is not used by intent but it is not necessary.

src/gui/widgets/MixerLine.cpp
spechtstatt: Yes I know now but I just thought it is more consistent with what is already there.

src/gui/MixerView.cpp
sakertooth: I was thinking of ``std::sort`` as well, but I was not sure how the default swap behavior would be changed to use ``swapChannels`` instead.

spechtstatt: Because it is described as a non stable sorting algorithm (https://en.cppreference.com/w/cpp/algorithm/sort) I could imagine that this is maybe a quicksort.

When I was thinking about the requirements I considered the following: the items are already nearly sorted, there are relatively few items and the sorting algorithm should swap in a mostly linear fashion. 

There is a good animation for comparison here:
https://www.toptal.com/developers/sorting-algorithms/nearly-sorted-initial-order

Of course it could be that there is in fact no difference if only one single element is out of order as it would be in this case.

Monospace-V:
I noticed a few things, could you clarify these?
Setting an fx mixer colour before linking makes the FX mixer's colour permanent unless the track's colour is changed (not reset, changed). This is a **bug**.
Could we get a direct "Assign and link to new Mixer channel" option, so we don't have to do it twice? It seems useful. Also an "Link to Mixer channel" in the context menu of track itself?
After linking, colours are linked but can be edited only from the track's context menu. However, some of us color-code multiple channels consecutively in the mixer itself based on the instrument type, and it could get annoying having to search. Can this be made two-way, being able to change colour from the mixer too? Similar thing with mixer channel name- being able to edit it through mixer too.
I understand this PR's scope is limited to addition of the option and linking of colour and names, right?
Also, upon right click, the "Move left/right" seems to be broken. I can still move them with Alt-left/Alt-right, but this behaviour acts strange (exchanging of Mixer channel names?, not entirely clear what's happening). There are a lot of weird issues, this isn;t difficult to reproduce. Also experiment with move left/right after adding a new instrument while auto-track-link is on. This is a **bug**.

Thanks!

spechtstatt:
> I noticed a few things, could you clarify these?
> 

Yes, of course - thank you for your thoughts and remarks.

> * Setting an fx mixer colour before linking makes the FX mixer's colour permanent unless the track's colour is changed (not reset, changed). This is a **bug**.

I will look into this - I have added your point to my task list above

> * Could we get a direct "Assign and link to new Mixer channel" option, so we don't have to do it twice? It seems useful. Also an "Link to Mixer channel" in the context menu of track itself?

I think this is a good idea - we should only consider that it may not be used very often because of the automatic adding and assigning of mixer channels - so maybe it is enough to provide only the second option? 

> * After linking, colours are linked but can be edited only from the track's context menu. However, some of us color-code multiple channels consecutively in the mixer itself based on the instrument type, and it could get annoying having to search. Can this be made two-way, being able to change colour from the mixer too? Similar thing with mixer channel name- being able to edit it through mixer too.

I was somehow focused on the idea that the track should control the mixer line (@unfa's original suggestion was to disable the auto link if the mixer name changes)  but because it is in fact a management of a 1:1 relationship there is no real reason to not enable this from both sides - so, yes I think this is also a good idea. 

> * I understand this PR's scope is limited to addition of the option and linking of colour and names, right?

Principally yes - I was just thinking to display some information on the mixer lines because it is collected anyway (as I mentiond above) - like if the line is used at all, how many tracks are linked or if this is linked to a pattern or a song track.

> * Upon right click on a Mixer channel, the "Move left/right" seems to be broken. I can still move them with Alt-left/Alt-right, but this behaviour acts strange (exchanging of Mixer channel names?, not entirely clear what's happening). There are a lot of weird issues, this isn;t difficult to reproduce. Also experiment with move left/right after adding a new instrument while auto-track-link is on. This is a **bug**.

I have the whole sorting issue not really finished yet and just prepared the lock of the move operations and concentrated on the auto link updating. So it is very plausible that there are still flaws but thank you for giving me more detailed feedback about this. 


> 
> Thanks!

src/core/Mixer.cpp
spechtstatt: Yes - but normally I would not use a switch statement to differ between only two outcomes.

spechtstatt:
I was thinking that with this functionality enabled it maybe makes sense to always auto link channels by default when assigning them and if they are free.

Because the link would anyway automatically removed if more than one gets assigned. And it could automaticallylink again if there would be only one instrument left.

And it would use the track text/color by default because it is initiated from there.

allejok96:
Tested and worked good.

Makes sense to link by default when an FX is *created* via the menu. Currently LMMS copies the name. That's halfway there :)

Don't do this for assignment to empty FX from the menu, because you might have a template with a "sidechain" FX and you start by assigning one instrument and it does this linking thing and changes the name and color and you be like *hey what's going on how do I rename it back* instead of just ignoring it and continuing to assign like you should.

Likewise I'm skeptical to auto re-link an FX when it becomes only one instrument again, because deleting that instrument may delete the FX without the user realizing it, and effects and stuff may get lost.

Suggestion on toggle button for auto-creating mixer channels for new tracks:
bild

If there will be more than one option, we could go for a cogwheel button with a menu instead, like on tracks.

Agreeing with Monospace that it would be nice if the ability to set color went both ways. I'd actually like it if you could change the color of an FX channel and it would color multiple tracks in the song editor. But that may be out of context.

It doesn't feel that important to be able to change name of linked channels from the mixer tho.

Marking options grey in the context menu instead of just leaving them out feels better. I know there are lots of things left to be polished. Just mentioning it.

spechtstatt:
@allejok96: also thought about multiple tracks linked to a fx line but I found not really a good solution for it regarding the naming.

But maybe it is a solution to link only the color as you suggested.

If there would be a good way to bring both approaches together - of course it could behave different depending on the count of linked tracks (1 = link name & color, n= link only color) but I am not sure if this is a good idea especially if the count changes only temporary during reorganizing.

Regarding multiple options I would probably add a sort button there  because I think it is better to have it not in an automatic fashion. And also an options button where you can set up e.g. if and how the pattern editor should be auto linked and which sorting should be applied when pressing the sort button.

If the sorting is manual I will add the move menu entries again and yes you are absolutely right: it is normally better to grey them out instead of hiding them (don't know why I did it this way 😵)

spechtstatt:
@allejok96: also thought about multiple tracks linked to a fx line but I found not really a good solution for it regarding the naming.

But maybe it is a solution to link only the color as you suggested.

If there would be a good way to bring both approaches together - of course it could behave different depending on the count of linked tracks (1 = link name & color, n= link only color) but I am not sure if this is a good idea especially if the count changes only temporary during reorganizing.

Regarding multiple options I would probably add a sort button there  because I think it is better to have it not in an automatic fashion. And also an options button where you can set up e.g. if and how the pattern editor should be auto linked and which sorting should be applied when pressing the sort button.

If the sorting is manual I will add the move menu entries again and yes you are absolutely right: it is normally better to grey them out instead of hiding them (don't know why I did it this way 😵)

spechtstatt:
Because I got the impression that there are different needs regarding the auto linking it is now possible to define different settings like: 
- linking one track to one channel vs. linking multiple tracks to one channel
- linking name and color vs linking only the color
- enable the automatic deletion of auto linked channels if there is no track assigned vs. disabling the automatic deletion of auto linked channels
- add channels automatically for the pattern editor like the song editor vs. assigning always the first channel of the pattern editor to all subsequent added pattern tracks (if you want to use only one mixer channel for the pattern editor)
- enable and disable the automatic sorting and also if the song tracks should ordered first or the pattern editor tracks.

I am not sure if this is the right approach or if it would be better to define two or three named sets of settings which correspond to different ways of working - but maybe this is unrealistic. 

Another way would be to reduce the options and split them - e.g. while the sorting currently considers only auto track linked tracks it could be extracted by allowing sorting of all (normal and linked) assigned tracks. (If there are multiple assigned tracks the last  would determine the sort order).

But anyway: with the settings it should now be possible to play around with the various approaches.

spechtstatt:
Still some issues with the pattern editor.

PR #6409, opened 2022-05-21
Transpose midi clips in song editor

allejok96:
Lets you quickly transpose a single track, section of the song or the whole song without messing around with master pitch. Right click a midi clip or selection:

image

Side note: Undo is wonky like every other multi-track action, but a suggested fix is coming later.


allejok96:
Thanks @superpaik. The menu content depend on what clip you click.

Initially I thought it would be a good idea ignore beat clips (zero-length notes), because that would keep drums the same while transposing the melody. But I forgot to exclude beat clips from transposing when you right click on *another* clip... Now when I think about it, beat clips in song editor is a legacy thing so why give them special treatment? And I'm sure there are people who want to transpose beat clips too. So I'll add that option.

As for the other options I think Clear all notes could be completely removed from Song editor.

src/gui/MidiClipView.cpp
```diff
@@ -140,8 +142,37 @@ void MidiClipView::changeName()
...
+void MidiClipView::transposeSelection()
+{
+	int semitones = QInputDialog::getInt(this, tr("Transpose"), tr("Semitones to transpose:"), 0, -NumKeys, NumKeys, 1);
+	if (semitones == 0) { return; }
+
+	// TODO make this not crash
+	// Engine::getSong()->addJournalCheckPoint();
+
+	QSet<Track*> m_changedTracks;
+	for (ClipView* clipview: getClickedClips())
```
JohannesLorenz, line 154: QtCreator smashes a [warning](https://github.com/KDE/clazy/blob/1.11/docs/checks/README-range-loop-detach.md) here.

JohannesLorenz (changes requested):
I know you plan more changes for beat clips, but here goes my review for the current change set.

JohannesLorenz:
One minor thing from a user point of view: If the dialog would somehow tell you that you can only transpose upwards by 20 or downwards by `NumKeys-1-20`, this would give a better idea of what you *can* transpose, instead of saying: Here are `2*NumKeys-1` possible transposals, but many of them won't make sense. I imagine running a for loop over all selections and checking what is possible before displaying that dialogue would be fast enough.

allejok96:
I'm inclined to keep the current behavior where the transpose option is hidden from beat clips, because otherwise when you right click a clip in Pattern editor it will show "Transpose" and that doesn't make much sense...

If you so happen to have a beat clip in Song editor and select all clips and transpose, there's no harm in letting that beat clip be transposed too... At least then you have an obscure (read LMMS-ish) way to transpose even those clips in Song editor, as opposed to excluding them from transpose completely.

Or what do you say?

src/gui/MidiClipView.cpp
JohannesLorenz: Actually I think this warning can not be fixed. I can't make the `clipview` `const`... Does that warning make any sense?

JohannesLorenz:
> I'm inclined to keep the current behavior where the transpose option is hidden from beat clips

I don't have too much of an opinion. It's probably OK to hide it there, and if it's requested to show it later, we can still push a simple fix.

src/gui/MidiClipView.cpp
```diff
@@ -140,8 +142,37 @@ void MidiClipView::changeName()
...
+			auto track = clipview->getTrackView()->getTrack();
+			if (!m_changedTracks.contains(track))
+			{
+				track->addJournalCheckPoint();
+				m_changedTracks.insert(track);
+			}
+			mcv->getMidiClip()->transpose(semitones);
+		}
+	}
+}
```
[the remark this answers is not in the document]
allejok96: Thanks! Never thought of that.

src/gui/MidiClipView.cpp
```diff
@@ -140,8 +142,37 @@ void MidiClipView::changeName()
 
 
 
+void MidiClipView::transposeSelection()
+{
+	int semitones = QInputDialog::getInt(this, tr("Transpose"), tr("Semitones to transpose:"), 0, -NumKeys, NumKeys, 1);
```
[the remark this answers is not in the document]
allejok96: Fixed. Now it checks max/min possible transpose.

allejok96:
Transpose dialog only shows valid options now. No clipping.

I did this by adding `getBounds` and `transpose` to `NoteVector`. Maybe that's too Pythonic, but I figured it could be reused to reduce some code in PianoRoll in the future.

src/gui/MidiClipView.cpp
```diff
@@ -140,8 +142,61 @@ void MidiClipView::changeName()
...
+		{
+			auto track = clipview->getTrackView()->getTrack();
+			if (!m_changedTracks.contains(track))
+			{
+				track->addJournalCheckPoint();
+				m_changedTracks.insert(track);
+			}
+			auto clip = mcv->getMidiClip();
+			clip->notes().transpose(semitones);
+			emit clip->dataChanged();
```
JohannesLorenz, line 187: Is that line equal to
```c++
update();
if( getGUI()->pianoRoll()->currentMidiClip() == m_clip )
{
	getGUI()->pianoRoll()->update();
}
```
? I really can't find out where that signal leads to. At least, the piano roll is being refreshed now.

src/gui/MidiClipView.cpp
```diff
@@ -140,8 +142,61 @@ void MidiClipView::changeName()
...
+			{
+				lowest = std::min(low, lowest);
+				highest = std::max(high, highest);
+			}
+		}
+	}
+
+	int minStep = 0 - lowest;
+	int maxStep = NumKeys - 1 - highest;
+	int semitones = QInputDialog::getInt(this, tr("Transpose"), tr("Semitones to transpose:"), 0, minStep, maxStep, 1);
```
JohannesLorenz, line 168: Of course, this now restricts to only do transposing that does not change the distance between the notes (i.e. does not bash all notes into one key). If that was wanted, ignore this comment.

JohannesLorenz (approved):
Code looks mostly good, see comments.

From testing point of view: What I find interesting is the behavior if I select only non-MidiClips and then do right-click on a MidiClip while the non-MidiClip selection is still active. It still allows to transpose the right-clicked MidiClip. I guess that right-clicked MidiClip gets part of the selection at time of right-clicking, so it's wanted?

src/gui/MidiClipView.cpp
JohannesLorenz: Oh right, just reading your comments that it was intended. Solving my own comment.

src/gui/MidiClipView.cpp
allejok96: Yes, it updates both the clip view and piano roll if active

src/core/Note.cpp
```diff
@@ -232,3 +232,59 @@ bool Note::withinRange(int tickStart, int tickEnd) const
...
+
+/*! \brief Transpose all notes in the vector by X semitones
+ *
+ *  Notes will be hard-clipped to the MIDI note range. To prevent this use getBounds() prior to transposing.
+ *
+ *  \param semitones Semitones to transpose
+ */
+void NoteVector::transpose(int semitones) const
+{
+	if (empty() || !semitones) { return; }
```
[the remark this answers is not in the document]
allejok96: Can't hurt. Added check for `dataChanged()`. But by design there must be at least one clip with notes to show the transpose option and at least 1 semitone transpose before anything is performed so I left `setModified()` as is.

The journal issue (creating checkpoint for tracks where nothing changed) would be fixed by the Batch undo PR. The checkpoints would still be created, but restored as one.

allejok96:
> I guess that right-clicked MidiClip gets part of the selection at time of right-clicking, so it's wanted?

Kinda correct. `getClickedClips()` returns the selection *or* the clicked clip if it's outside the selection. The behavior is the same when setting clip colors.

JohannesLorenz (approved):

src/gui/MidiClipView.cpp
```diff
@@ -140,8 +142,64 @@ void MidiClipView::changeName()
...
```
DomClark, line 168: Personally, I would have written "Semitones to transpose *by*:", but I'll leave it up to you.

include/Note.h
```diff
@@ -244,7 +244,14 @@ class LMMS_EXPORT Note : public SerializingObject
...
 
 
-typedef QVector<Note *> NoteVector;
+
+
+class NoteVector: public QVector<Note*>
+{
+public:
+	bool getBounds(TimePos& start, TimePos& end, int& lower, int& upper) const;
+	bool transpose(int semitones) const;
```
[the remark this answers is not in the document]
allejok96: So I should make them static methods of `NoteVector` instead?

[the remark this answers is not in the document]
DomClark: I would leave `NoteVector` as a typedef for `QVector<Note*>`, and make them free (non-member) functions, so
```cpp
// See also my later comment for further changes to getBounds
bool getBounds(const NoteVector& notes, TimePos& start, TimePos& end, int& lower, int& upper);
bool transpose(const NoteVector& notes, int semitones);
```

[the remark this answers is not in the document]
allejok96: I don't see a lot of these kind of functions in header files (I want at least `getBounds` to be accessible), so I figured putting it in a class would make things cleaner. Or is there some other way to not have a ton of functions in the global scope?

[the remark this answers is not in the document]
DomClark: > Or is there some other way to not have a ton of functions in the global scope?

#6174?

> I don't see a lot of these kind of functions in header files (I want at least getBounds to be accessible), so I figured putting it in a class would make things cleaner.

I really wouldn't worry about it too much. You say you want to make the functions reusable - perhaps the reason there aren't many free functions in header files is because we haven't been very good at making things reusable up until now. If you look in namespace `std`, there are loads of free functions, but that doesn't cause anyone a problem. If you want, you can add `notes` or something to the names to make the purpose clearer without a class name for context (e.g. `boundsForNotes`, `transposeNotes`), but even without, you won't have problems with name collisions since parameter types distinguish functions in C++ (unlike in C, Python, etc.).

src/gui/MidiClipView.cpp
allejok96: You're the better English speaker!

src/core/Note.cpp
```diff
@@ -232,3 +232,61 @@ bool Note::withinRange(int tickStart, int tickEnd) const
...
+
+/*! \brief Get the outer bounds of the notes
+ *
+ *  \param start - Will be set to the first time position
+ *  \param end - Will be set to the last end position
+ *  \param lower - Will be set to the lowest key
+ *  \param upper - Will be set to the highest key
+ *  \return false if there are no notes
+ */
+bool NoteVector::getBounds(TimePos& start, TimePos& end, int& lower, int& upper) const
```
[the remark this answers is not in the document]
allejok96: Thanks for the tip! I didn't know about std::optional. That's great.

src/core/Note.cpp
```diff
@@ -232,3 +232,61 @@ bool Note::withinRange(int tickStart, int tickEnd) const
...
 		&& length().getTicks() != 0;
 }
+
+
+
+
+
+
+
+
```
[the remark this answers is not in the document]
allejok96: Removed

include/Note.h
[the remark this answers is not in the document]
allejok96: Ok, that makes sense. I moved out `getBounds` and renamed it `boundsForNotes`. And `transpose` is so simple that I just dropped it.

allejok96:
Fixed Dom's comments and rebased

src/core/Note.cpp
```diff
@@ -232,3 +232,32 @@ bool Note::withinRange(int tickStart, int tickEnd) const
...
+	{
+		// TODO should we assume that NoteVector is always sorted correctly,
+		// so first() always has the lowest time position?
+		start = std::min(start, note->pos());
+		end = std::max(end, note->endPos());
+		lower = std::min(lower, note->key());
+		upper = std::max(upper, note->key());
+	}
+
+	return NoteBounds{start, end, lower, upper};
```
JohannesLorenz, line 262: Just for my understanding: A negative side effect of returning `std::optional<NoteBounds>` instead of passing the four parameters via references (as "out parameters") is, using the return value might cause a copy? However, in C++17, it seems to be [forced to elide the copying](https://en.cppreference.com/w/cpp/language/copy_elision), which is why we do not need to worry?

JohannesLorenz (approved):
Code OK: Dom's fixes seems to be applied correctly. Testing also (still) works.

PhysSong (approved):
Ready to merge once you resolve conflicts.

src/core/Note.cpp
```diff
@@ -237,4 +237,33 @@ bool Note::withinRange(int tickStart, int tickEnd) const
...
 
+
+
+/*! \brief Get the start/end/bottom/top positions of notes in a vector
+ *
+ *  Returns no value if there are no notes
+ */
+std::optional<NoteBounds> boundsForNotes(const NoteVector& notes)
+{
+	if (notes.empty()) { return {}; }
```
DomClark, line 248: I wonder whether `return std::nullopt;` would be clearer here.

src/gui/clips/MidiClipView.cpp
```diff
@@ -144,8 +146,66 @@ void MidiClipView::changeName()
...
+void MidiClipView::transposeSelection()
+{
+	const auto selection = getClickedClips();
+
+	// Calculate the key boundries for all clips
+	int highest = 0;
+	int lowest = NumKeys - 1;
+	for (ClipView* clipview: selection)
+	{
+		if (auto mcv = dynamic_cast<MidiClipView*>(clipview))
```
DomClark, line 158: `qobject_cast` would be more efficient here (and further down in this file).

DomClark (approved):
Only two minor comments - the code is fine as is, so feel free to ignore them.

src/core/Note.cpp
PhysSong: I thought `{}` was enough, but `std::nullopt` looks clearer to me.

src/core/Note.cpp
allejok96: Interesting. Even prior to C++17
> compilers are permitted, but not required to omit the copy

I did some tests. With `optional` it's a different story, because the value seems to be constructed first and then copied into the `optional` object. To avoid it you must do

    return std::optional<NoteBounds>{std::in_place, start, end, lower, upper};

I'm feeling tempted to remove all the `optional` fuzz and just go with `operator bool()`

src/gui/clips/MidiClipView.cpp
allejok96: Thanks! I'll add it for the sake of good practice.

src/core/Note.cpp
allejok96: I decided to drop `std::optional` and go with `operator bool()` to return true if the NoteBound is non-zero. That way we get return value optimization like @JohannesLorenz in a clean and simple way. Thoughts?

src/core/Note.cpp
DomClark: At -O1 or above, GCC optimises out the extra copy and stores the constructor arguments straight into the optional. (See https://godbolt.org/z/aKoWcrGnY.) Even if it didn't, it's only copying 16 bytes in a hot memory region. An optional is better semantically than giving a special meaning to a particular value of `NoteBounds`, and I wouldn't sacrifice that unless profiling showed it to be especially problematic.

allejok96: Agreed that this is not a problem, disagreeing on semantics. But that's why you're the C++ guy and I'm not (cool site btw)

JohannesLorenz:
3 reviews and all rework done. Merge?

PR #6412, opened 2022-05-24
Add Tuner plugin

sakertooth:
Detects the pitch of an audio signal. This plugin uses aubio with the YinFFT algorithm for pitch detection by default. GTune was used as a reference plugin to compare results.

image


mirk0dex:
Shouldn't we merge this already?

sakertooth:
Closing. I don't know when I'll get back to this, if ever, and maybe someone with more knowledge with DSP should do this if its really wanted.

PR #6415, opened 2022-05-26
Correctly detect architecture on powerpc*

pkubaj:


PhysSong:
Thank you! I think you may also update the next `IF`~`ELSE` part and also `src/lmmsconfig.h` for PPC.

tresf:
There's been a request to backport this to `stable-1.2`.  Since the backport has conflicts and won't cherry-pick cleanrly, I'll request that this is done by the OP (or perhaps @barracuda156).  Please tag me in the Pull Request and I'll be happy to merge.

barracuda156:
@PhysSong Do we need RISC-V in 1.2? I have no idea if it will build and cannot verify that. PowerPC is needed.

tresf:
> Do we need RISC-V in 1.2? I have no idea if it will build and cannot verify that. PowerPC is needed.

Just PowerPC is fine.

barracuda156:
@tresf Added ARM, since otherwise Apple Silicon will be broken: fallback for an unknown CPU is 32 bit assumption.

tresf:
> @tresf Added ARM, since otherwise Apple Silicon will be broken: fallback for an unknown CPU is 32 bit assumption.

Thanks, merged.  We don't support Apple Silicon on this older version, but it's a nice consideration. 🍻

barracuda156:
> We don't support Apple Silicon on this older version

That’s perfectly fine. My initial plan is just to update the port to 1.2.2 with Qt4, since that provisionally may work for all supported systems, while having several versions is harder to implement and yet harder to maintain (and this is not a particularly easy port to begin with). In this context, not preventing aarch64 from building is helpful. If it still fails, well, someone will have to update it to a newer version (perhaps together with newer Intel systems).

PR #6418, opened 2022-05-29
Make VST sync always-on and non-global

### LINKED ISSUE #4730: Remove "Sync VST plugins to host playback" option
The "Sync VST plugins to host playback" option causes VST plugins to be informed of the following parameters, in addition to what is sent when the option is disabled:
* Time signature (defaults to 4/4 when the option is disabled)
* Playback state (defaults to playing)
* Playback position (calculated from tempo by default)
* Loop start, end, and enabled state (defaults to loop disabled)
* Start of current bar (calculated from playback position and time signature)
* Jumps in playback (defaults to continuous playback)

The option is disabled by default, and this seems to cause confusion for some users who wonder why their VST plugins won't stay in time with the rest of their song. This has come up a few times in `#support` on Discord, and I expect there are some other users who don't report a bug or ask a question and just assume LMMS is broken. I'm not aware of any other DAW that allows toggling VST sync (correct me if there is), and I'm not sure what the use case really is.
I think we should change it to be enabled by default in `stable-1.2` to reduce confusion and fix what could be perceived as a bug, and remove the option eventually in `master`.

DomClark:
Fixes #4730. Also fixes a bug I can't find an issue for, where VST sync doesn't work properly with multiple LMMS instances open because the shared memory key for the sync data is hardcoded.

Changes:
* Removed VST sync option from setup dialog.
* VST sync shared memory key is now a UUID, like the FIFO keys and the audio buffer keys.
* New `IdSyncKey` IPC message added to send the sync key to remote plugins - this is the first message sent and is now what triggers the remote plugin to send `IdHostInfoGotten`.
* Simplified some code now that it no longer needs to handle VST sync being turned off.


include/RemotePluginClient.h
```diff
@@ -265,6 +244,14 @@ bool RemotePluginClient::processMessage( const message & _m )
 		case IdUndefined:
 			return false;
 
+		case IdSyncKey:
+			m_vstSyncData.attach(_m.getString(0));
```
PhysSong, line 248: The same story with previous PR: how would failures be handled here? I'm asking because RemoteZynAddSubFx should still work in case of failures.

plugins/VstBase/RemoteVstPlugin.cpp
```diff
@@ -1801,86 +1773,71 @@ intptr_t RemoteVstPlugin::hostCallback( AEffect * _effect, int32_t _opcode,
...
 			SHOW_CALLBACK( "amc: audioMasterGetTime\n" );
 			// returns const VstTimeInfo* (or 0 if not supported)
 			// <value> should contain a mask indicating which
 			// fields are required (see valid masks above), as some
 			// items may require extensive conversions
 
-			// Shared memory was initialised? - see song.cpp
-			//assert( __plugin->m_vstSyncData != nullptr );
+			const auto syncData = __plugin->getVstSyncData();
+			assert(syncData != nullptr);
```
PhysSong, line 1784: You're making RemoteVstPlugin crash when VST sync data is available. It's acceptable, but could you give some explanations?

include/RemotePluginClient.h
DomClark: Shared memory has to be functional on the system for the remote plugins to be useful, since it is used to transfer audio to (and from) LMMS. The shared memory here is handled using the exact same mechanism as that for audio - generate a random UUID, use it to create a shared memory object, send the key to the remote plugin, and use it to attach the shared memory object. I don't see why we would expect this line to potentially fail, but still expect the audio buffer to attach reliably. On Windows, we even use shared memory for sending the IPC messages.

My inclination is that if any shared memory fails to attach, we treat that as a hard error and kill the remote plugin. Here I do that by letting the exception escape; this is admittedly not very elegant, and maybe a log message followed by a call to `exit` would be better, but it's intended behaviour. All other shared memory is essential, and I don't think it's worth complicating the code to allow it to hobble on in the rare case that a single piece of shared memory fails to attach. (Currently, we do try to continue if the audio buffer can't be attached - I intend to change this too, but it's out of scope here.)

plugins/VstBase/RemoteVstPlugin.cpp
DomClark: See my other comment for the reasons behind this. The assertion here is because it should not be possible for `syncData` to be null here - the `IdSyncKey` message should have been sent by LMMS, and if the shared memory failed to attach, the remote plugin should have terminated at that point. I generally try to use exceptions to handle runtime errors (such as failure to attach shared memory), and assertions to handle logic errors (such as here, where code elsewhere should ensure this condition is true).

include/RemotePluginClient.h
PhysSong: I'm okay with the decision, but as you said, proper logging and/or cleanup on failure will be better. It's definitely good to handle this in this PR, but you may do that in another PR if you want.

PhysSong (approved):

PR #6421, opened 2022-06-01
Fixes #6398: Colour of Master bus on FX Mixer retains when switching project

### LINKED ISSUE #6398: Colour of Master bus on FX Mixer retains when switching project
### Bug Summary
If a project has the master bus on the FX-Mixer set a certain colour, and then we open any other project where the master channel does not have a colour set, master channel remains that colour.

#### Steps to reproduce
1. Create a project and change the colour of the mixer's master channel. Alternatively, open a project that has a non-default colour on the mixer's master bus.
2. In the same session of LMMS, create a new project, or open any other project where master channel's colour is still default.
3. Look at the colour of master channel on mixer.

#### Expected behavior
Master channel resets to default colour

#### Actual behavior
Master channel remains coloured

#### Demonstration:

https://user-images.githubusercontent.com/76674645/167285887-deb4df1e-348a-41a1-b69f-46275cc2e5a8.mp4

#### Affected LMMS versions

LMMS 1.3 (only version with ability to change FX-channel colours)

spechtstatt:
Fixes #6398


DomClark (approved):

Monospace-V:
Can't replicate in any way. Seems fixed.

zonkmachine:
Merge?

JohannesLorenz (approved):
The changes are OK. However, looking at the comment about `m_hasColor` in the header, it's suggested to use `std::optional`. Indeed, `m_hasColor` is always used together with `m_hasColor`, so (instead of using `std::optional`) we could remove `m_hasColor` and instead use `QColor::isValid()`, which returns `false` after using the default CTOR? This should reduce future errors, where `m_hasColor` and `QColor::isValid()` are not in sync.

Spekular:
Optional should be available by now though, I believe we already bumped to the required C++ version.

PhysSong:
> > Optional should be available by now though, I believe we already bumped to the required C++ version.
> 
> True, but the we have redundant information in `std::optional` and `QColor` about validity.

I don't think two validity information represent the same thing. Invalid `QColor` may mean not only we haven't set a value yet, but also we set it to an invalid value manually for some reason.

JohannesLorenz:
I think they are equivalent in our current code. Do you see any use case in our code where an invalid `MixerChannel::m_color` and a not-set `MixerChannel::m_color` should be handled differently?

PhysSong:
> Do you see any use case in our code where an invalid `MixerChannel::m_color` and a not-set `MixerChannel::m_color` should be handled differently?

No for now, but if we allow setting colors by name, for example, `m_color` might be set to an invalid color. In that case, however, it might be good to treat invalid colors as not set.

Spekular:
In my opinion, a single (possibly invalid) member of type QColor doesn't imply that having no color is an acceptable state. An optional makes it clear that we expect there to be situations where there's no color, and it's not necessarily an error.

JohannesLorenz:
> In my opinion, a single (possibly invalid) member of type QColor doesn't imply that having no color is an acceptable state. An optional makes it clear that we expect there to be situations where there's no color, and it's not necessarily an error.

Look like the team is for using `std::optional` then. @spechtstatt do you think you could fix that? It's the last comment of this PR which has to be resolved. Alternatively, someone else could code it for you.

DomClark:
I feel like the switch to `std::optional` is a bit out of scope for what is otherwise a tiny change. There's a bunch of other modernisation to be done across the codebase, and I think it would fit better there. If the author is happy to add it here, then I won't object, but I think it's overcomplicating things to require it here.

spechtstatt:
Well - I am somehow with @DomClark in that it may be to much for this tiny bug fix but I tried anyway :-)

Not sure if I made it right though. Especially when calling the color selection dialog.

JohannesLorenz:
@spechtstatt The use of `std::optional` looks correct, but shouldn't it *replace* `m_hasColor`?

spechtstatt:
Ok - I see - somehow like the Nullables in C#.

Yes - makes sense.

PhysSong:
@spechtstatt It seems like somehow changed tabs to spaces.

spechtstatt:
Oh sorry.

I switched my Computer recently and I probably forgot to change it in the qt creator  settings.

PhysSong:
> @spechtstatt The use of `std::optional` looks correct, but shouldn't it _replace_ `m_hasColor`?

@spechtstatt Could you address that with the indentation fixes(revert to tabs)?

spechtstatt:
Yes of course.

src/core/Mixer.cpp
```diff
@@ -705,6 +704,7 @@ void Mixer::clearChannel(mix_ch_t index)
 	ch->m_volumeModel.setDisplayName( ch->m_name + ">" + tr( "Volume" ) );
 	ch->m_muteModel.setDisplayName( ch->m_name + ">" + tr( "Mute" ) );
 	ch->m_soloModel.setDisplayName( ch->m_name + ">" + tr( "Solo" ) );
+	ch->m_color = { };
```
JohannesLorenz, line 707: As Dom adviced me in Discord, it's probably more readable to write

```suggestion
	ch->m_color = std::nullopt;
```

This makes it clear to the reader that `ch->m_color` is an `std::optional`.

src/core/Track.cpp
```diff
@@ -639,14 +638,13 @@ void Track::toggleSolo()
...
 {
-	m_hasColor = true;
 	m_color = c;
 	emit colorChanged();
 }
 
 void Track::resetColor()
 {
-	m_hasColor = false;
+	m_color = { };
```
JohannesLorenz, line 647: Same as above:
```suggestion
	m_color = std::nullopt;
```

src/gui/MixerLine.cpp
```diff
@@ -442,7 +443,7 @@ void MixerLine::selectColor()
 // Disable the usage of color on this mixer line
 void MixerLine::resetColor()
 {
-	Engine::mixer()->mixerChannel( m_channelIndex )->m_hasColor = false;
+	Engine::mixer()->mixerChannel( m_channelIndex )->m_color = { };
```
JohannesLorenz, line 446: Same as above:
```suggestion
	Engine::mixer()->mixerChannel( m_channelIndex )->m_color = std::nullopt;
```

JohannesLorenz (commented):
Only minor and very easy-to-solve comments. Feel free to ignore.

src/core/Mixer.cpp
```diff
@@ -791,8 +790,8 @@ void Mixer::loadSettings( const QDomElement & _this )
 		m_mixerChannels[num]->m_name = mixch.attribute( "name" );
 		if( mixch.hasAttribute( "color" ) )
 		{
-			m_mixerChannels[num]->m_hasColor = true;
-            m_mixerChannels[num]->m_color->setNamedColor( mixch.attribute( "color" ) );
+			auto col = QColor(mixch.attribute( "color" ));
+			m_mixerChannels[num]->m_color = col;
```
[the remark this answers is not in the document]
spechtstatt: Yes - this was somehow a debugging artefact because the former setNamedColor did not work with the optional color.

include/Mixer.h
```diff
@@ -76,12 +76,10 @@ class MixerChannel : public ThreadableJob
 		void setColor (QColor newColor)
 		{
 			m_color = newColor;
-			m_hasColor = true;
 		}
 
 		// TODO C++17 and above: use std::optional instead
```
allejok96, line 81: Can remove this comment now

src/gui/MixerLine.cpp
```diff
@@ -94,6 +94,7 @@ MixerLine::MixerLine( QWidget * _parent, MixerView * _mv, int _channelIndex ) :
 	// mixer sends knob
 	m_sendKnob = new Knob( knobBright_26, this, tr( "Channel send amount" ) );
 	m_sendKnob->move( 3, 22 );
+
```
allejok96, line 97: Empty line slipped in here

allejok96 (approved):
Code looks good except for one warning during build

allejok96 (changes requested):

allejok96 (approved):
Confused by the commit log, but it builds fine and my comments seem to be fixed. Haven't test run tho.

JohannesLorenz (approved):
Tested. Bug is fixed an no regressions.

JohannesLorenz:
2 code reviews + 1 test review passed. Suggesting merge.

src/gui/MixerLine.cpp
```diff
@@ -435,7 +435,7 @@ void MixerLine::setStrokeInnerInactive( const QColor & c )
 void MixerLine::selectColor()
 {
 	auto channel = Engine::mixer()->mixerChannel( m_channelIndex );
-	auto new_color = ColorChooser(this).withPalette(ColorChooser::Palette::Mixer)->getColor(channel->m_color);
+	auto new_color = ColorChooser(this).withPalette(ColorChooser::Palette::Mixer)->getColor(*channel->m_color);
```
PhysSong, line 438: If `channel->m_color` doesn't have a value, the behavior is undefined. [`std::optional::value_or`](https://en.cppreference.com/w/cpp/utility/optional/value_or) might be useful, but it's hard to say what is the desired default value.

On the other hand(and out of scope), I found that using `ColorChooser` in this way looks weird: both `QColorDialog::setStandardColor`(called by `ColorChooser::withPalette`) and `QColorDialog::getColor` are static functions.

spechtstatt:
@allejok96: you mean the commit message? Yes: I was repeating the issue title again like in the first  commit which was probably not the best idea.

src/gui/MixerLine.cpp
allejok96: Here's a quick color reference just for fun

[Default highlight of MixerLine](https://github.com/LMMS/lmms/blob/0b27497be2ac93551c4e243e43d7f0811ac0b889/src/gui/MixerLine.cpp#L183)

    backgroundActive();

[Background of a widget](https://doc.qt.io/qt-5/qpalette.html#ColorRole-enum)

    palette().color(QPalette::Window);

[Widget's `color` in the stylesheet](https://doc.qt.io/qt-6/stylesheet-reference.html#color-prop)

    palette().color(foregroundRole());

[LMMS green (depending on theme naturally)](https://github.com/LMMS/lmms/blob/6ae075edf3c052a923074b46ae1115718a621acd/data/themes/default/style.css#L1009)

    qApp->palette().color(QPalette::WindowText);

JohannesLorenz:
@spechtstatt There is still one outstanding comment from PhysSong. Can you please try to fix it (or tell us if someone else should fix it)?

spechtstatt:
@JohannesLorenz: not sure how to fix this - any suggestions?

src/gui/MixerLine.cpp
DomClark: I think the default mixer line colour (as defined in the stylesheet) would be an intuitive choice - that way, the dialog opens with the current mixer line colour, just as it would if a custom one were set.

PhysSong: @DomClark's suggestion will look like this:
```suggestion
	auto new_color = ColorChooser(this).withPalette(ColorChooser::Palette::Mixer)->getColor(channel->m_color.value_or(backgroundActive()));
```

spechtstatt: Used DomClark suggestion based on PhySongs code example

src/gui/MixerLine.cpp
```diff
@@ -434,8 +434,8 @@ void MixerLine::setStrokeInnerInactive( const QColor & c )
 // Ask user for a color, and set it as the mixer line color
 void MixerLine::selectColor()
 {
-	auto channel = Engine::mixer()->mixerChannel( m_channelIndex );
-	auto new_color = ColorChooser(this).withPalette(ColorChooser::Palette::Mixer)->getColor(*channel->m_color);
+	auto channel = Engine::mixer()->mixerChannel( m_channelIndex );	
+    auto new_color = ColorChooser(this).withPalette(ColorChooser::Palette::Mixer)->getColor(channel->m_color.value_or(backgroundActive().color()));
```
PhysSong, line 438: Formatting issues:
```suggestion
	auto channel = Engine::mixer()->mixerChannel( m_channelIndex );
	auto new_color = ColorChooser(this).withPalette(ColorChooser::Palette::Mixer)->getColor(channel->m_color.value_or(backgroundActive().color()));
```

src/core/Mixer.cpp
```diff
@@ -741,7 +741,7 @@ void Mixer::saveSettings( QDomDocument & _doc, QDomElement & _this )
 		ch->m_soloModel.saveSettings( _doc, mixch, "soloed" );
 		mixch.setAttribute( "num", i );
 		mixch.setAttribute( "name", ch->m_name );
-		if( ch->m_hasColor ) mixch.setAttribute( "color", ch->m_color.name() );
+		if( ch->m_color.has_value()) mixch.setAttribute( "color", ch->m_color->name() );
```
Veratil, line 744:
```suggestion
		if (ch->m_color.has_value()) { mixch.setAttribute("color", ch->m_color->name()); }
```

src/core/Mixer.cpp
```diff
@@ -789,8 +789,7 @@ void Mixer::loadSettings( const QDomElement & _this )
 		m_mixerChannels[num]->m_name = mixch.attribute( "name" );
 		if( mixch.hasAttribute( "color" ) )
 		{
-			m_mixerChannels[num]->m_hasColor = true;
-			m_mixerChannels[num]->m_color.setNamedColor( mixch.attribute( "color" ) );
+			m_mixerChannels[num]->m_color = QColor(mixch.attribute( "color" ));
```
Veratil, line 792:
```suggestion
			m_mixerChannels[num]->m_color = QColor(mixch.attribute("color"));
```

src/core/Track.cpp
```diff
@@ -209,9 +208,9 @@ void Track::saveSettings( QDomDocument & doc, QDomElement & element )
 		element.setAttribute( "trackheight", m_height );
 	}
 	
-	if( m_hasColor )
+	if( m_color.has_value())
```
Veratil, line 211:
```suggestion
	if (m_color.has_value())
```

src/core/Track.cpp
```diff
@@ -209,9 +208,9 @@ void Track::saveSettings( QDomDocument & doc, QDomElement & element )
...
 	{
-		element.setAttribute( "color", m_color.name() );
+		element.setAttribute( "color", m_color->name() );
```
Veratil, line 213:
```suggestion
		element.setAttribute("color", m_color->name());
```

src/gui/MixerLine.cpp
```diff
@@ -174,9 +174,9 @@ void MixerLine::drawMixerLine( QPainter* p, const MixerLine *mixerLine, bool isA
 	int width = mixerLine->rect().width();
 	int height = mixerLine->rect().height();
 	
-	if( channel->m_hasColor && !muted )
+	if( channel->m_color.has_value() && !muted )
```
Veratil, line 177:
```suggestion
	if (channel->m_color.has_value() && !muted)
```

src/gui/MixerLine.cpp
```diff
@@ -174,9 +174,9 @@ void MixerLine::drawMixerLine( QPainter* p, const MixerLine *mixerLine, bool isA
...
 	{
-		p->fillRect( mixerLine->rect(), channel->m_color.darker( isActive ? 120 : 150 ) );
+		p->fillRect( mixerLine->rect(), channel->m_color->darker( isActive ? 120 : 150 ) );
```
Veratil, line 179:
```suggestion
		p->fillRect(mixerLine->rect(), channel->m_color->darker(isActive ? 120 : 150));
```

src/gui/MixerLine.cpp
```diff
@@ -446,7 +446,7 @@ void MixerLine::selectColor()
 // Disable the usage of color on this mixer line
 void MixerLine::resetColor()
 {
-	Engine::mixer()->mixerChannel( m_channelIndex )->m_hasColor = false;
+	Engine::mixer()->mixerChannel( m_channelIndex )->m_color = std::nullopt;
```
Veratil, line 449:
```suggestion
	Engine::mixer()->mixerChannel(m_channelIndex)->m_color = std::nullopt;
```

Veratil (changes requested):
Code style fixes and then we can merge :)

JohannesLorenz:
@spechtstatt Do you plan to continue this PR (solving the last open comments), or is it necessary that someone of the other devs takes over your PR?

spechtstatt:
Hi Johannes,

unfortunately, I currently have no time to work on the PR and this will continue for a while.

So, yes. I would be happy if someone could take over.

JohannesLorenz:
According to Veratil and Dom, this can now be merged, so I will merge it now.

Issue #6414, opened 2022-05-24
LMMS master loses Carla instrument configuration when loading old project file


Rua:
I have an old project file created in LMMS 1.2.2, which has a track whose instrument is a Carla rack with a loaded VST. If I load this project in the latest LMMS master, then the Carla rack loads, but many of the parameters of the VST instrument are set to 0.

#### Steps to reproduce

- Open LMMS 1.2.2 with a new project.
- Add a track with a Carla rack instrument.
- Load a VST plugin with Carla and set parameters as desired.
- Save the project and exit.
- Open LMMS master and open the previously saved project.
- Open the Carla rack GUI of the track in question.

#### Expected behavior

The VST parameters are preserved.

#### Actual behavior

The VST parameters are set to zero, including things like master volume, ADSR levels, filter cutoff etc. It depends on the instrument, but it happens with at least Dexed and TAL-U-No-LX.

#### Screenshot

![image](https://user-images.githubusercontent.com/3090610/170081212-3b62398a-81b9-442b-9c4a-2a437537215d.png)

#### Affected LMMS versions

3518d30

The bug does not occur in 1.3.0-alpha.1, so it must have been introduced at some point since.

zonkmachine:
Have you tested earlier builds of lmms master?

Rua:
No, how would I do that? Downloading and building each time will take a while.

zonkmachine:
> No, how would I do that? Downloading and building each time will take a while.

Yes. Pinpointing the precise commit that introduced an issue can take quite some time. '**git bisect**' is a command that speeds up the process.

The older 1.3 alpha is here:
https://github.com/LMMS/lmms/releases/tag/v1.3.0-alpha.1

Rua:
I have been able to determine that the problem appeared somewhere in 2021, but I am unable to investigate further because for some reason every build I make is now triggering a segmentation fault when I run it.

zonkmachine:
Does the alpha above work?

Rua:
If that's the same version that's available from the LMMS website, then yes. There is no bug in that one.

zonkmachine:
> every build I make is now triggering a segmentation fault when I run it, even from commits that previously worked.

Did you build it with debug flags?
Can you run it from gdb and get a backtrace when it crashes?

Rua:
```
(gdb) r
Starting program: /home/rua/code/lmms-bisect/build/lmms 
[Thread debugging using libthread_db enabled]
[...]

PR #6239, opened 2021-12-23
Rename FxMixer to Mixer

yashraj466:
This PR aims to rename FxMixer to Mixer, as decided in #6089 and #5592.

PR #6424, opened 2022-06-04
Correctly upgrade projects saved by LMMS forks

allejok96:
Check `version` (of file format) instead of `creatorversion` (of LMMS) when deciding if a project file needs to be upgraded.

This fixes the [currently broken](https://github.com/LMMS/lmms/pull/6309#issuecomment-1146224632) `demos/DnB.mmpz`. It got saved by a fork `1.3.0-alpha.1.199` back in #6239. The current master is at `1.3.0-alpha.1.191` so - funny enough - it will be automatically un-broken in 9 commits.

### DIFF
--- a/src/core/DataFile.cpp
+++ b/src/core/DataFile.cpp
@@ -1916,13 +1916,14 @@ void DataFile::loadData( const QByteArray & _data, const QString & _sourceFile )
 		if( !success ) qWarning("File Version conversion failure.");
 	}
 
+	// Perform upgrade routines
+	if (m_fileVersion < UPGRADE_METHODS.size()) { upgrade(); }
+
 	if (root.hasAttribute("creatorversion"))
 	{
 		// compareType defaults to All, so it doesn't have to be set here
 		ProjectVersion createdWith = root.attribute("creatorversion");
 		ProjectVersion openedWith = LMMS_VERSION;
-
-		if (createdWith < openedWith) { upgrade(); }
 
 		if (createdWith.setCompareType(ProjectVersion::Minor)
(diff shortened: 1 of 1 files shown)


zonkmachine:
Tested, fix confirmed!

src/core/DataFile.cpp
```diff
@@ -1916,14 +1916,15 @@ void DataFile::loadData( const QByteArray & _data, const QString & _sourceFile )
 		if( !success ) qWarning("File Version conversion failure.");
 	}
 
+	// Perform upgrade routines
+	if (m_fileVersion < UPGRADE_METHODS.size()) { upgrade(); }
```
DomClark, line 1920: `upgrade()` sets the `creatorversion` attribute to the current LMMS version, so calling it here means the next block of code can't see the original value of that attribute.

allejok96: Oh right, I didn't realize it touched the XML

DomClark (approved):

Rua:
Does this also fix #6414?

zonkmachine:
> Does this also fix #6414?

Can you please download a test bild (LmmsBot, above) and test this? That would be much appreciated!

Rua:
I tested it, unfortunately the issue still exists. I guess something else is going on with that one.

JohannesLorenz (approved):

allejok96:
Merge? Or a second real world test? I don't know how you usually do for smaller PRs.

Veratil:
Confirmed fixes the import of demos (not just `demos/DnB.mmpz`) with B+B tracks introduced from f56fc68b.

Okay to merge from me.

PR #6431, opened 2022-06-11
MidiRawClient: Fix setting useless param

JohannesLorenz:
MidiChannelPressure and MidiProgramChange have only one param, so only
set this one param for those.

### DIFF
--- a/src/core/midi/MidiClient.cpp
+++ b/src/core/midi/MidiClient.cpp
@@ -225,7 +225,7 @@ void MidiClientRaw::parseData( const unsigned char c )
 			m_midiParseData.m_midiEvent.setVelocity(m_midiParseData.m_buffer[1]);
 		case MidiChannelPressure:
 		case MidiProgramChange:
-			// set key (or for ChannelPressure/ProgramChange: set param 1)
+			// set key (or for ChannelPressure/ProgramChange: set 1st param)
 			m_midiParseData.m_midiEvent.setKey(m_midiParseData.m_buffer[0]);
 			break;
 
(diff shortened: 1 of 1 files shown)


src/core/midi/MidiClient.cpp
```diff
@@ -221,10 +221,12 @@ void MidiClientRaw::parseData( const unsigned char c )
 		case MidiNoteOff:
 		case MidiNoteOn:
 		case MidiKeyPressure:
+			// set velocity of key
+			m_midiParseData.m_midiEvent.setVelocity(m_midiParseData.m_buffer[1]);
```
PhysSong, line 225: There really should be an explicit comment or annotation when falling through in a `switch`-`case`.

src/core/midi/MidiClient.cpp
```diff
@@ -221,10 +221,12 @@ void MidiClientRaw::parseData( const unsigned char c )
...
 		case MidiChannelPressure:
 		case MidiProgramChange:
+			// set key (or for ChannelPressure/ProgramChange: set 1st param)
 			m_midiParseData.m_midiEvent.setKey(m_midiParseData.m_buffer[0]);
```
PhysSong, line 229: I think `m_midiParseData.m_midiEvent.setParam(0, m_midiParseData.m_buffer[0]);` is semantically more correct, but I'll leave the decision up to you.

PhysSong:
I think it would be better to eliminate the fall-through completely if we use `setParam` directly in two cases, which doesn't clearly represent the meaning for three cases above where it is for note key.

PhysSong (approved):

JohannesLorenz:
For the record, the CI error also occurs in other PRs.

This PR awaits 2nd approval and testing.

PhysSong:
@JohannesLorenz I think it's safe to merge without 2nd review.

PR #6432, opened 2022-06-11
Add non-Qt base64 encode/decode methods

Veratil:
Throwing this out to provide non-Qt base64 encode/decode methods. Threw them in the `lmms::base64` namespace for now.

Currently provided:
```cpp
std::string encode(const std::string&);
std::string decode(const std::string&);
char* encode(const char*, const size_t);
```

If we want to keep the `char*` type function, then I can add the decode method for that. If not we can delete it.

Also, I'm thinking it would be better to use move semantics for returning the `std::string`, but I'm not too versed in that aspect and how to make it work correctly.

### DIFF
--- a/include/base64.h
+++ b/include/base64.h
@@ -30,6 +30,8 @@
 #include <QString>
 #include <QVariant>
 
+#include <string>
+#include <cmath>
 
 namespace base64
 {
@@ -53,4 +55,12 @@ namespace base64
 
 }
 
+namespace lmms {
+	namespace base64 {
+		std::string encode(const std::string&);
+		std::string decode(const std::string&);
+		char* encode(const char*, const size_t);
+		//char* decode(const char*, const size_t);
+	}
+}
 #endif
--- a/src/core/base64.cpp
+++ b/src/core/base64.cpp
@@ -56,3 +56,159 @@ QVariant decode( const QString & _b64, QVariant::Type _force_type )
 
 } ;
 
+namespace lmms {
+	namespace base64 {
+		//! @brief base64 encode byte array data, returns pointer to result string
+		//!   or nullptr if length == 0; user is expected to free returned result pointer
+		//!
+		//! @param data : byte array of characters
+		//! @param length : length of data array in bytes
+		//! @return char* array of encoded data of length (ceil[length / 3] * 4)
+		char* encode(const char* data, const size_t length) {
+			const char map[] = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_";
+			const char pad = '=';
+			if (length == 0) {
+				return nullptr;
+			}
+			/*
+			Every 3 bytes of the origin string relates to 4 bytes of encoded string
+			length:         [0, 1, 2, 3, 4, 5, 6,  7,  8,  9, 10, 11, 12, ...]
+			result length:  [0, 4, 4, 4, 8, 8, 8, 12, 12, 12, 16, 16, 16, ...]
+			result padding: [0, 2, 1, 0, 2, 1, 0,  2,  1,  0,  2,  1,  0, ...]
+			*/
+			const size_t reserve = static_cast<size_t>(std::ceil(static_cast<double>(length) / 3.0) * 4.0);
+			const size_t offset = length % 3;
+			const size_t padding = offset ? 3 - offset : offset;
+			char* result = new char[reserve+1]; // +1 for \0
+			result[reserve] = '\0';
+			int result_index = 0;
+			for (int i = 0; i < length; i += 3) {
+				const char view[3] {
+					data[i],
+					(i + 1 < length ? data[i + 1] : '\0'),
+					(i + 2 < length ? data[i + 2] : '\0')
+				};
+				// create result chars, 8bit char -> 6bit map offset
+				result[result_index]     = map[view[0] >> 2];
+				result[result_index + 1] = map[((view[0] & 0x03) << 4) | (view[1] >> 4)];
+				result[result_index + 2] = map[((view[1] & 0x0F) << 2) | (view[2] >> 6)];
+				result[result_index + 3] = map[view[2] & 0x3F];
+				// next result index
+				result_index += 4;
+			}
+			// string will have trailing 'A' or 'AA' until replaced with pad char
+			for (int p = 0; p < padding; ++p) {
+				result[reserve - 1 - p] = pad;
+			}
+			return result;
+		}
+
+		//! @brief base64 encode std::string
+		//!
+		//! @param data : original std::string
+		//! @return base64 encoded std::string
+		std::string encode(const std::string& data) {
+			const char map[] = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_";
+			const char pad = '=';
+			if (data.length() == 0) {
+				return "";
+			}
+			std::string result;
+			result.reserve(static_cast<size_t>(std::ceil(static_cast<double>(data.length()) / 3.0) * 4.0));
+			const size_t offset = data.length() % 3;
+			const size_t padding = offset ? 3 - offset : offset;
+			for (int i = 0; i < data.length(); i += 3) {
+				const char view[3] {
+					data[i],
+					((i + 1 < data.length()) ? data[i + 1] : '\0'),
+					((i + 2 < data.length()) ? data[i + 2] : '\0')
+				};
+				result.push_back(map[view[0] >> 2]);
+				result.push_back(map[((view[0] & 0x03) << 4) | (view[1] >> 4)]);
+				result.push_back(map[((view[1] & 0x0F) << 2) | (view[2] >> 6)]);
+				result.push_back(map[view[2] & 0x3F]);
+			}
+			auto result_pad = result.rbegin();
+			for (int p = 0; p < padding; ++p, ++result_pad) {
+				*result_pad = pad;
+			}
+			return result;
+		}
+
+		//! @brief base64 decode std::string
+		//!
+		//! @param data : encoded base64 std::string
+		//! @return decoded std::string
+		std::string decode(const std::string& data) {
+			if (data.length() % 4 != 0) {
+				throw std::length_error("base64::decode : data length not a multiple of 4");
+			}
+			const std::map<char, uint32_t> map {
+				{'A',  0}, {'B',  1}, {'C',  2}, {'D',  3}, {'E',  4},
+				{'F',  5}, {'G',  6}, {'H',  7}, {'I',  8}, {'J',  9},
+				{'K', 10}, {'L', 11}, {'M', 12}, {'N', 13}, {'O', 14},
+				{'P', 15}, {'Q', 16}, {'R', 17}, {'S', 18}, {'T', 19},
+				{'U', 20}, {'V', 21}, {'W', 22}, {'X', 23}, {'Y', 24},
+				{'Z', 25}, {'a', 26}, {'b', 27}, {'c', 28}, {'d', 28},
+				{'e', 30}, {'f', 31}, {'g', 32}, {'h', 33}, {'i', 34},
+				{'j', 35}, {'k', 36}, {'l', 37}, {'m', 38}, {'n', 39},
+				{'o', 40}, {'p', 41}, {'q', 42}, {'r', 43}, {'s', 44},
+				{'t', 45}, {'u', 46}, {'v', 47}, {'w', 48}, {'x', 49},
+				{'y', 50}, {'z', 51}, {'0', 52}, {'1', 53}, {'2', 54},
+				{'3', 55}, {'4', 56}, {'5', 57}, {'6', 58}, {'7', 59},
+				{'8', 60}, {'9', 61}, {'-', 62}, {'_', 63}
+			};
+			const char pad = '=';
+			std::string result("");
+			if (data.length() == 0) {
+				return result;
+			}
+			result.reserve(static_cast<size_t>(std::ceil(static_cast<double>(data.length()) / 4.0) * 3.0));
+			// loop 4 characters at a time, except the last 4
+			for (int i = 0; i < data.length() - 4; i += 4) {
+				uint32_t container = (
+					map.at(data[i])     << 18
+					| map.at(data[i + 1]) << 12
+					| map.at(data[i + 2]) <<  6
+					| map.at(data[i + 3])
+				);
+				result.push_back(static_cast<char>((container & 0x00FF0000) >> 16));
+				result.push_back(static_cast<char>((container & 0x0000FF00) >>  8));
+				result.push_back(static_cast<char>(container & 0x000000FF));
+			}
+			// handle last 4 characters for padding check
+			const auto block_start = data.end() - 4;
+			// XX==
+			if (*(block_start + 2) == pad) {
+				uint32_t container = (
+					map.at(*block_start) << 2
+					| map.at(*(block_start + 1)) >> 4
+				);
+				result.push_back(static_cast<char>(container & 0x000000FF));
+			}
+			// XXX=
+			else if (*(block_start + 3) == pad) {
+				uint32_t container = (
+					map.at(*block_start)      << 10
+					| map.at(*(block_start + 1)) << 4
+					| map.at(*(block_start + 2)) >> 2
+				);
+				result.push_back(static_cast<char>((container & 0x0000FF00) >> 8));
+				result.push_back(static_cast<char>(container & 0x000000FF));
+			}
+			// XXXX (repeat of above loop)
+			else {
+				uint32_t container = (
+					map.at(*block_start)     << 18
+					| map.at(*(block_start + 1)) << 12
+					| map.at(*(block_start + 2)) <<  6
+					| map.at(*(block_start + 3))
+				);
+				result.push_back(static_cast<char>((container & 0x00FF0000) >> 16));
+				result.push_back(static_cast<char>((container & 0x0000FF00) >>  8));
+				result.push_back(static_cast<char>(container & 0x000000FF));
+			}
+			return result;
+		}
+	}
+}


sakertooth:
> Also, I'm thinking it would be better to use move semantics for returning the `std::string`, but I'm not too versed in that aspect and how to make it work correctly.

Generally, I don't think you would return ``T&&``. If ``T`` is returned, copy elision (or more specifcally, RVO) should kick in.

JohannesLorenz:
What would be the reason to elliminate Qt methods? FWIR, this is only useful for the GUI, which uses Qt anyways?

Veratil:
> What would be the reason to elliminate Qt methods? FWIR, this is only useful for the GUI, which uses Qt anyways?

Good question! Here's the reason this PR exists:
1. It's in core, we're removing Qt from core.
2. This was easy enough to throw together.

If base64 is unneeded in core then we just close this PR and move the file out of core folder into gui instead. 🙂

sakertooth:
> What would be the reason to elliminate Qt methods? FWIR, this is only useful for the GUI, which uses Qt anyways?

FWIW, my sample caching PR utilizes Base64 encoding/decoding within ``SampleBufferV2`` (eventually it will be just ``SampleBuffer``) when loading data from Base64. As ``SampleBufferV2`` is part of the core, this is a useful PR. This could've been part of the original Qt removal PR for simplicity's sake though.

Do note that this is one example of Base64 usage within the core. There is probably more.

PhysSong:
I suggest always returning `std::string` and removing code duplication using [`std::string_view`](https://en.cppreference.com/w/cpp/string/basic_string_view).

PhysSong:
Well, I mean, you can merge two `encode` functions using `std::string_view` and optionally add `std::string`/`const char*, size_t` variants.

Veratil:
> Well, I mean, you can merge two `encode` functions using `std::string_view` and optionally add `std::string`/`const char*, size_t` variants.

Oh! I thought you were only talking about the `std::string` ones. 😁

sakertooth:
I made some tweaks to this implementation [here](https://github.com/Veratil/lmms/pull/8). Took me a while because I was constantly refactoring it and trying to get something compact. Both functions should be functional and were tested with the test vectors specified in [RFC 4648](https://datatracker.ietf.org/doc/html/rfc4648#section-4) section 4.

JohannesLorenz:
Suggesting to add a few simple "unit tests", because it should be easy to test and helps us verifying it works.

Veratil:
> Suggesting to add a few simple "unit tests", because it should be easy to test and helps us verifying it works.

Done.

Also changed the current Qt-based encode/decode namespace to `lmms::gui::base64`, and kept the new in `lmms::base64`. This way we can adjust things to the new functions as needed.

PhysSong:
> Also changed the current Qt-based encode/decode namespace to `lmms::gui::base64`

Why in `lmms::gui`? We're removing Qt from core, but the base64 methods has nothing to do with GUI.

Also for the unit tests, I suggest iterating over original-encoded pairs and check encoding/decoding in the loop body, if possible.

sakertooth:
> > Suggesting to add a few simple "unit tests", because it should be easy to test and helps us verifying it works.
> 
> 
> 
> Done.
> 
> 
> 
> Also changed the current Qt-based encode/decode namespace to `lmms::gui::base64`, and kept the new in `lmms::base64`. This way we can adjust things to the new functions as needed.

Why not just just provide an overload for ``QString`` and put the Qt implementation there? No need for these widespread changes in my honest opinion.

Once ``QString`` has been removed in the Qt-removal PR, we can remove the Qt Base64 functions as well.

Veratil:
> Why in `lmms::gui`? We're removing Qt from core, but the base64 methods has nothing to do with GUI.

Whenever I left them in the same namespace I was getting errors about invalid overloads or something.

> Also for the unit tests, I suggest iterating over original-encoded pairs and check encoding/decoding in the loop body, if possible.

They're just tests, but okay. :)

tests/src/core/base64Test.cpp
```diff
@@ -0,0 +1,55 @@
...
+        std::vector<std::pair<std::string, std::string>> test_vectors{
+            {"", ""},
+            {"f", "Zg=="},
+            {"fo", "Zm8="},
+            {"foo", "Zm9v"},
+            {"foob", "Zm9vYg=="},
+            {"fooba", "Zm9vYmE="},
+            {"foobar", "Zm9vYmFy"},
+        };
+        for (auto vector : test_vectors)
```
PhysSong, line 47: Since we use C++17, something like this is also possible:
```suggestion
        for (const auto& [original, encoded] : test_vectors)
```

DomClark: Alternatively, one could use the technique described in https://doc.qt.io/qt-5/qttestlib-tutorial2-example.html, which has the advantage that the collection of pairs is exposed to the test framework. I'm not sure how well it plays with non-Qt types though, and it is a little more verbose.

src/core/base64.cpp
```diff
@@ -54,4 +54,134 @@ QVariant decode( const QString & _b64, QVariant::Type _force_type )
...
+ */
+std::string encode(std::string_view data)
+{
+	if (data.empty()) { return ""; }
+
+	// base64 encoded string
+	std::string result;
+
+	// number of chunks to process + padding
+	auto [numChunks, numTrailingBytes] = std::div(data.length(), numBytesPerChunk);
```
messmerd, line 72: The `std::div` function returns a `div_t` object whose `quot` and `rem` members are not guaranteed to be in any particular order ([see link](https://en.cppreference.com/w/cpp/numeric/math/div)), so using a structured binding here would result in unpredictable behavior.

src/core/base64.cpp
```diff
@@ -54,4 +54,134 @@ QVariant decode( const QString & _b64, QVariant::Type _force_type )
...
+		);
+	};
+	// Get the fourth base64 character offset from the char chunk
+	auto b64char4 = [](std::string_view chunk) {
+		return static_cast<std::string::value_type>(chunk.size() > 1 ? (chunk[2] & 0x3F) : 0);
+	};
+	for (int currentChunk = 0; currentChunk < numChunks; ++currentChunk)
+	{
+		std::string_view chunk = data.substr(currentChunk * numBytesPerChunk, numBytesPerChunk);
+		std::string output{
```
messmerd, line 113: Since this string is a fixed size of 4, you might get better performance in this loop by replacing `std::string` with `std::array` and using `result.append(output.data(), output.size());` down below.

messmerd (changes requested):

tests/src/core/base64Test.cpp
Veratil: I went the QTest route. Very cool how it works.

include/base64.h
```diff
@@ -53,4 +57,55 @@ namespace lmms::base64
 
 } // namespace lmms::base64
 
-#endif
+namespace lmms::base64 {
+	constexpr std::array<char, 64> map =
```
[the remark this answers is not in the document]
Veratil: Made both the map and reverse map inline. Whether it helps anything I don't know. 😁

src/core/base64.cpp
Veratil: Fixed! Thanks for showing that. I wasn't aware `div_t` wasn't in a guaranteed order! 😩

src/core/base64.cpp
Veratil: Changed to `std::array`. I think you're right, a fixed sized array could probably be more optimized than a `std::string`.

src/core/base64.cpp
```diff
@@ -54,4 +54,136 @@ QVariant decode( const QString & _b64, QVariant::Type _force_type )
...
+
+	/*
+		char1               char2               char3
+		7 6 5 4 3 2   1 0 | 7 6 5 4   3 2 1 0 | 7 6   5 4 3 2 1 0
+		5 4 3 2 1 0 | 5 4   3 2 1 0 | 5 4 3 2   1 0 | 5 4 3 2 1 0
+		b64c1         b64c2           b64c3           b64c4
+	*/
+	// Get the first base64 character offset from the char chunk
+	auto b64char1 = [](std::string_view chunk) {
+		return static_cast<std::string::value_type>(chunk[0] >> 2);
```
PhysSong, line 90: - Any reason for passing `std::string_view` by value?
- Given that `std::string` is an alias of `std::basic_string<char>`, `std::string::value_type` is always `char`.

src/core/base64.cpp
```diff
@@ -54,4 +54,136 @@ QVariant decode( const QString & _b64, QVariant::Type _force_type )
...
+			map[b64char2(chunk)],
+			pad,
+			pad
+		};
+		switch (chunk.length()) {
+			case 3:
+				output[3] = map[b64char4(chunk)];
+			case 2:
+				output[2] = map[b64char3(chunk)];
+			default: /* no-op */;
```
PhysSong, line 126: It might be good to add [`[[fallthrough]];`](https://en.cppreference.com/w/cpp/language/attributes/fallthrough) when falling through. I prefer explicitly `breaking` on `case 2` and `default`, but I'll leave it up to you.

Veratil: I was not aware of this, I'll add them.

src/core/base64.cpp
Veratil: * My understanding of `string_view` was that this is the way to do it. If not then I can change it. 👍
* My thought was in the instance for some reason `std::string` isn't based on `char`.

PhysSong: > My understanding of `string_view` was that this is the way to do it.

It turns out you're correct, indeed.

include/base64.h
```diff
@@ -53,4 +59,55 @@ namespace lmms::base64
 
 } // namespace lmms::base64
 
-#endif
+namespace lmms::base64 {
+	constexpr inline std::array<char, 64> map =
```
sakertooth, line 63: ``constexpr`` implies inline, so ``inline`` shouldn't be needed here.

include/base64.h
```diff
@@ -53,4 +59,55 @@ namespace lmms::base64
...
+	{
+		'A','B','C','D','E','F','G','H','I','J','K','L','M','N','O','P','Q','R','S','T','U','V','W','X','Y','Z',
+		'a','b','c','d','e','f','g','h','i','j','k','l','m','n','o','p','q','r','s','t','u','v','w','x','y','z',
+		'0','1','2','3','4','5','6','7','8','9',
+		'-','_'
+	};
+	const inline std::map<char, int> rmap {
```
sakertooth, line 70: Why not use ``constexpr`` here as well? If anything, it will be consistent with ``map``.

src/core/base64.cpp
```diff
@@ -54,4 +54,136 @@ QVariant decode( const QString & _b64, QVariant::Type _force_type )
 }
 
 
+//TODO C++20: It *may* be possible to make the following functions constepxr given C++20's constexpr std::string.
+
+/**
+ * @brief Base64 encodes data.
+ * @param data
+ * @return std::string containing the Base64 encoded data.
+ */
```
sakertooth, line 63: These doxygen comments may be more useful in the header file rather than the ``.cpp`` file as it is more easily accessible there.

PS: I know I was the one who added the C++20 TODO comment, but maybe integrate it better into the main doxygen comment with ``@todo``.

src/core/base64.cpp
```diff
@@ -54,4 +54,136 @@ QVariant decode( const QString & _b64, QVariant::Type _force_type )
...
+		result.append(output.data(), output.size());
+	}
+	return result;
+}
+
+/**
+ * @brief Decodes data in Base64.
+ * @param data
+ * @return std::string containing the original data.
+ */
```
sakertooth, line 137: Same as before:

> These doxygen comments may be more useful in the header file rather than the .cpp file as it is more easily accessible there.

src/core/base64.cpp
```diff
@@ -54,4 +54,136 @@ QVariant decode( const QString & _b64, QVariant::Type _force_type )
...
+			(chunk.size() > 1 ? ((chunk[2] & 0xC0) >> 6) : 0)
+		);
+	};
+	// Get the fourth base64 character offset from the char chunk
+	auto b64char4 = [](std::string_view chunk) {
+		return static_cast<std::string::value_type>(chunk.size() > 1 ? (chunk[2] & 0x3F) : 0);
+	};
+	for (int currentChunk = 0; currentChunk < numChunks; ++currentChunk)
+	{
+		std::string_view chunk = data.substr(currentChunk * numBytesPerChunk, numBytesPerChunk);
```
sakertooth, line 114:
```suggestion
		auto chunk = data.substr(currentChunk * numBytesPerChunk, numBytesPerChunk);
```

include/base64.h
```diff
@@ -53,4 +59,55 @@ namespace lmms::base64
...
+	constexpr int numBytesPerChunk = lcm / numBitsPerChar;
+	constexpr int numBase64CharPerChunk = lcm / numBitsPerBase64Char;
+	// double check math works and bit width matches
+	static_assert(numBytesPerChunk * numBitsPerChar == numBase64CharPerChunk * numBitsPerBase64Char);
+
+	auto encode(std::string_view data) -> std::string;
+	auto decode(std::string_view data) -> std::string;
+} // namespace lmms::base64
+
+#endif // _BASE64_H
```
sakertooth, line 113: This would've conflicted with the ``LMMS_`` header guard prefix change if it happened on the same line.

tests/src/core/base64Test.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+#include "QTestSuite.h"
+
+#include "base64.h"
+
+class Base64Test : QTestSuite
+{
+    Q_OBJECT
+private slots:
+    void create_test_data()
```
sakertooth, line 33:
```suggestion
    void createTestData()
```

tests/src/core/base64Test.cpp
```diff
@@ -0,0 +1,73 @@
...
+        // Test Vectors from RFC 4648 Section 10
+        QTest::newRow("empty string")  << ""       << "";
+        QTest::newRow("1 chunk 2 pad") << "f"      << "Zg==";
+        QTest::newRow("1 chunk 1 pad") << "fo"     << "Zm8=";
+        QTest::newRow("1 chunk 0 pad") << "foo"    << "Zm9v";
+        QTest::newRow("2 chunk 2 pad") << "foob"   << "Zm9vYg==";
+        QTest::newRow("2 chunk 1 pad") << "fooba"  << "Zm9vYmE=";
+        QTest::newRow("2 chunk 0 pad") << "foobar" << "Zm9vYmFy";
+    }
+    void b64_encode_data()
```
sakertooth, line 47:
```suggestion
    void b64EncodeData()
```

tests/src/core/base64Test.cpp
```diff
@@ -0,0 +1,73 @@
...
+    {
+        create_test_data();
+    }
+    void b64_encode()
```
sakertooth, line 51:
```suggestion
    void b64Encode()
```

tests/src/core/base64Test.cpp
```diff
@@ -0,0 +1,73 @@
...
+    {
+        using namespace lmms::base64;
+
+        QFETCH(QString, original);
+        QFETCH(QString, encoded);
+        QCOMPARE(QString(encode(original.toStdString()).c_str()), encoded);
+    }
+    void b64_decode_data()
```
sakertooth, line 59:
```suggestion
    void b64DecodeData()
```

tests/src/core/base64Test.cpp
```diff
@@ -0,0 +1,73 @@
...
+    {
+        create_test_data();
+    }
+    void b64_decode()
```
sakertooth, line 63:
```suggestion
    void b64Decode()
```

tests/src/core/base64Test.cpp
```diff
@@ -0,0 +1,73 @@
...
+    {
+        using namespace lmms::base64;
+
+        QFETCH(QString, original);
+        QFETCH(QString, encoded);
+        QCOMPARE(QString(encode(original.toStdString()).c_str()), encoded);
+    }
+    void b64_decode_data()
+    {
+        create_test_data();
```
sakertooth, line 61:
```suggestion
        createTestData();
```

tests/src/core/base64Test.cpp
```diff
@@ -0,0 +1,73 @@
...
+        QTest::newRow("1 chunk 2 pad") << "f"      << "Zg==";
+        QTest::newRow("1 chunk 1 pad") << "fo"     << "Zm8=";
+        QTest::newRow("1 chunk 0 pad") << "foo"    << "Zm9v";
+        QTest::newRow("2 chunk 2 pad") << "foob"   << "Zm9vYg==";
+        QTest::newRow("2 chunk 1 pad") << "fooba"  << "Zm9vYmE=";
+        QTest::newRow("2 chunk 0 pad") << "foobar" << "Zm9vYmFy";
+    }
+    void b64_encode_data()
+    {
+        create_test_data();
```
sakertooth, line 49:
```suggestion
        createTestData();
```

tests/CMakeLists.txt
```diff
@@ -20,6 +20,7 @@ ADD_EXECUTABLE(tests
 	$<TARGET_OBJECTS:lmmsobjs>
 
 	src/core/AutomatableModelTest.cpp
+	src/core/base64Test.cpp
```
sakertooth, line 23: Should be ``src/core/Base64Test.cpp``.

tests/src/core/base64Test.cpp
```diff
@@ -0,0 +1,73 @@
+/*
+ * base64Test.cpp
```
sakertooth, line 2:
```suggestion
 * Base64Test.cpp
```

include/base64.h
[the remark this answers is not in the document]
messmerd: That [might not be true](https://stackoverflow.com/a/49913226) for global variables, but regardless, I think all of these implementation details do not belong in the header anyway and should be moved to `base64.cpp`. Then they certainly will not need the `inline`.

include/base64.h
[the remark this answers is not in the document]
messmerd: Unfortunately, `std::map` does not have a constexpr constructor so `rmap` cannot be constexpr, but I agree with your other points.

sakertooth:
Hey @Veratil, I would like to fix the merge conflicts here and try implementing this again if you would allow it. Don't want #7095 to just be closed (its their first PR, so I would like to merge it at some point), so I'll leave the unit test for that PR.

PR #6434, opened 2022-06-12
Remove QVector from core

Veratil:
Removes [almost all] `QVector` from `src/core/`.

Only thing left is to change `InstrumentFunctionNoteStacking::ChordTable`. This will most likely require a bit more than changing `QVector` to `std::vector` since `ChordTable` inherits `QVector` and we don't want to inherit `std::vector`.

After changing a file I'd recompile lmms and open multiple demos and play them all to give a "working" test. No crashes. 👍

### DIFF
--- a/include/AudioEngine.h
+++ b/include/AudioEngine.h
@@ -27,10 +27,10 @@
 
 #include <QMutex>
 #include <QThread>
-#include <QVector>
 #include <QWaitCondition>
 #include <samplerate.h>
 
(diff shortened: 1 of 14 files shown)


sakertooth:
Shouldn't this be merged as its own commit in #6426?

Veratil:
> Shouldn't this be merged as its own commit in #6426?

It could! 😁

PR #6438, opened 2022-06-18
Add check-namespace script and rewrite verify script

JohannesLorenz:
This adds a (very simplified) namespace/ifdef checker. Also, this puts `check-strings` and the new `check-namespace` together.

Obviously, this PR fails its own `check-namespace` action, until PR #6174 will have been merged.


tests/scripted/check-namespace
```diff
@@ -0,0 +1,199 @@
...
+    'plugins/ReverbSC/base.h',
+    'plugins/ReverbSC/dcblock.c',
+    'plugins/ReverbSC/dcblock.h',
+    'plugins/ReverbSC/revsc.c',
+    'plugins/ReverbSC/revsc.h'
+}
+
+files = [Path(f) for f in result.stdout.splitlines() if
+            f not in exclude_files
+            and not f.startswith('plugins/LadspaEffect/calf')
```
irrenhaus3, line 81: I'd put the files filtered here into `exclude_files` for readability.

tests/scripted/check-namespace
```diff
@@ -0,0 +1,199 @@
...
+            and not f.startswith('plugins/LadspaEffect/caps')
+            and not f.startswith('plugins/LadspaEffect/cmt')
+            and not f.startswith('plugins/LadspaEffect/swh')
+            and not f.startswith('plugins/LadspaEffect/tap')
+            and not f.__contains__('portsmf')
+        ]
+
+namespaces = re.compile(r'^\s*namespace\s*\S*', re.MULTILINE)
+no_include_after_namespace = re.compile(
+    r'^(\s*#\s*(?:include|if|el|endif)[^\n]*)*(namespace [\w:]+ [{]|#(?:ifn?def(?: [\w_]+)|if|endif|el)[a-z]*(?: // ?[\w_]+)?|[}](?: // namespace [\w:]+)?|\n)*$')
```
irrenhaus3, line 91: This is a very complex regex, so I'm not sure I'm getting it right, but is the part at the beginning checking for any of `#include`, `#if`, `#elif` and `#endif` inside a namespace scope? If so, it's probably going to flag a lot of files that use conditional compilation in the middle of class and/or function implementations.

tests/scripted/check-namespace
```diff
@@ -0,0 +1,199 @@
...
+
+        # the remainder should be solely includes + >=0 namespaces
+        results_as_str = '\n'.join(results)
+        no_include_after_namespace.match(results_as_str) or error(cur_file,
+                                                                  f'File \'{cur_file}\' has includes inside/after a namespace: {results_as_str}')
+
+        if str(cur_file) == debugfile:
+            print(f'{cur_file}: {results}')
+
+        # now also remove #include and the namespaces - the resulting string must be empty
```
irrenhaus3, line 185: Doesn't the C++ compiler catch that already?

tests/scripted/check-namespace
JohannesLorenz: No, the first part allows all kind of macros (`#...`), including `#include`. And after that, only `namespace` and `#if...` (as well as closing }) can occur, but no `#include` anymore. So if this does not match, we have `#include` after the first occurence of `namespace` (which is why the variable is named like that).

tests/scripted/check-namespace
JohannesLorenz: It does, but if the string is not empty here, the script has an error. E.g. it could have removed `namespace xyz {`, but not the matching `} // namspace xyz`. I had these issues during development of the script, so I thought I leave this assertion here to make sure the script works.

tests/scripted/check-namespace
JohannesLorenz: I pushed a new version, can you please check if it's OK?

irrenhaus3 (approved):

tests/scripted/check-namespace
PhysSong: @JohannesLorenz Can you provide some examples where the regex matches(or not), and test this regex with the examples?

JohannesLorenz: @PhysSong I added the only test case I thought could make sense, please check [this link](https://github.com/JohannesLorenz/lmms/blob/f1f93b10fa0075d83f706c836b993c66568e17e0/tests/scripted/verify#L194-L199).

PhysSong: Looking into the regex, I found some potential issues:
- `namespace [\w:]+ `: What if someone writes in wrong formatting such as `namespace  lmms{`?
- ` // `: Similar, like `}//namespace lmms`.
- The regex should match multiple lines, but you haven't specified `re.MULTILINE` while using `^` and `$`.

Not sure if there are more, because I haven't tested this with many test cases yet.

tests/scripted/check-namespace
```diff
@@ -0,0 +1,202 @@
...
+        results = namespaces.findall(cur_file.read_text(errors='replace'))
+        results_as_str = '\n'.join(results)
+        # print(f'{cur_file}: {results_as_str}')
+
+        if str(cur_file) not in known_no_namespace_lmms:
+            has_namespace_lmms.search(results_as_str) or error(cur_file, f'File has no namespace lmms')
+
+        debugfile = ''  # 'include/RemotePluginBase.h' # debugging
+
+        # remove comments
```
allejok96, line 154: I think you can remove comments using one regex with lookaheads
```py
# Remove block comments (if it starts with //* it's not a block comment)
# Remove comments not preceded by #endif and not starting with "namespace"
cur_text = re.sub(r'(?<!/)/[*](.|\n)*?[*]/|(?!<#endif) *// *(?!namespace).*', r'', cur_text)
```

JohannesLorenz:
@allejok96 I merged your rewrite. It passes most tests, but sadly, it can not detect if the end comment of a macro/namespace does not match the start:

```c++
#ifdef XXX
#endif // XXX
namespace ABC {
} // namespace ABC
```

Do you think you could add this? Or should we ignore this for now?

allejok96:
Huh, I designed it with that in mind, apparently I failed :) maybe I'll have a little time this weekend to look at it.

JohannesLorenz:
> Huh, I designed it with that in mind, apparently I failed :) maybe I'll have a little time this weekend to look at it.

Thanks for the offer. I just fixed the verify script ot let it run again (temporarily). It might be easier for you if you rebase on my changes.

allejok96:
I just tested it on your branch, it works as intended. I never told you how it works though. Feel free to add more comments in the code. Here is an example file you might test against:

```cpp
#ifndef HEADERGUARD

// Pass - short with no ifdef inside
#ifdef SHORT_MACRO
#endif

// Fail - nested ifdef
#ifdef BAD_MACRO
#ifdef SHORT_SUB_MACRO
#endif
#endif

// Fill with 30 blanks and it will fail - too long
#ifdef BAD_LONG_MACRO
...
#endif

// Pass - no braces inside
namespace GoodNamespace {
class Example;
}

// Fail - contains braces
namespace BadNamespace {
class LongExample
{
public:
    LongClass() {}
};
}

// Pass (if file name is *.h) - first ifndef is concidered header guard and comment is not enforced
#endif 
```

allejok96:
@JohannesLorenz I opened a PR to fix a few edge cases, like `}` inside a comment

tests/scripted/check-namespace
```diff
@@ -0,0 +1,195 @@
...
+# types
+
+class BlockType:
+    pass
+
+
+IF_MACRO = BlockType()
+HEADER_GUARD = BlockType()
+CODE_BLOCK = BlockType()
+EXTERN = BlockType()
```
PhysSong, line 25: Why comparing this using object identity, not using `Enum`?

allejok96: I checked out Enum, but it didn't give me any better type checking than doing it this way, plus you'd have to write EnumType.MEMBER. Just didn't see the real win...

allejok96: I guess Enum prints nicer error messages though

PR #6439, opened 2022-06-19
clang-tidy: Apply modernize-use-override everywhere

irrenhaus3:
This declares all virtual functions in base classes `virtual` and all overriding functions in derived classes `override` instead of `virtual`, following C++ Core Guideline C.128 https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-override. This also affects destructors.


DomClark (approved):

JohannesLorenz (approved):
Code is OK. Also did some basic tests, everything works.

PR #6440, opened 2022-06-20
Don't send effEditIdle on CLI rendering

PhysSong:
This fixes Plugin-Alliance plugins crashing on CLI rendering, at least on Windows.

Most plugins seem to handle the cases correctly where `effEditIdle` is called without editor instantiation(`effEditOpen`), but PA plugins doesn't.

This PR fixes the issue by sending `effEditIdle` only the editor is created.

### DIFF
--- a/plugins/VstBase/RemoteVstPlugin.cpp
+++ b/plugins/VstBase/RemoteVstPlugin.cpp
@@ -2165,7 +2165,7 @@ void RemoteVstPlugin::idle()
 		return;
 	}
 	setProcessing( true );
-	if (!HEADLESS)
+	if (!HEADLESS && m_window)
 	{
 		pluginDispatch( effEditIdle );
 	}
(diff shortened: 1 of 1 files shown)


plugins/VstBase/RemoteVstPlugin.cpp
```diff
@@ -2165,7 +2165,10 @@ void RemoteVstPlugin::idle()
 		return;
 	}
 	setProcessing( true );
-	pluginDispatch( effEditIdle );
+	if (!HEADLESS && m_window)
```
DomClark, line 2168: Technically this could just test `m_window`, since that's the condition we actually care about, and `HEADLESS` implies `!m_window`. But we check `!HEADLESS && m_window` elsewhere in this file, and it's an equivalent condition, so I'm also happy for this to stay as it is.

DomClark (commented):
There's another place where `effEditIdle` is sent to the plugin here: https://github.com/LMMS/lmms/blob/2a09be4c955c257190f9c45ca9f549c6b21be89b/plugins/VstBase/RemoteVstPlugin.cpp#L2219
I think this should be changed too.

PhysSong:
@DomClark Good point! I also fixed that.

DomClark (approved):

PR #6446, opened 2022-06-23
clang-tidy: Apply modernize-redundant-void-arg everywhere.

irrenhaus3:
This refactor transforms function declarations (and friends) of the form `T foo(void);` into `T foo();`. The `void` keyword is redundant there, since the declaration is equivalent to one with an empty argument list.


sakertooth (approved):

JohannesLorenz:
Code is OK, but what about something like `sed 's#( )#()#'` ? :grin:

irrenhaus3:
You got it.

PR #6447, opened 2022-06-24
namespace-lmms: Add missing prefixes to types in AudioSoundIo.

irrenhaus3:
CI didn't catch this one, but @zonkmachine sure did! Now, 'tis fixed.

### DIFF
--- a/include/AudioSoundIo.h
+++ b/include/AudioSoundIo.h
@@ -70,7 +70,7 @@ public:
 		return QT_TRANSLATE_NOOP( "AudioDeviceSetupWidget", "soundio" );
 	}
 
-	class setupWidget : public AudioDeviceSetupWidget
+	class setupWidget : public gui::AudioDeviceSetupWidget
 	{
 	public:
 		setupWidget( QWidget * _parent );
@@ -84,8 +84,8 @@ public:
 	private:
 
 		AudioSoundIoSetupUtil m_setupUtil;
-		ComboBox * m_backend;
-		ComboBox * m_device;
+		gui::ComboBox * m_backend;
+		gui::ComboBox * m_device;
 
 		ComboBoxModel m_backendModel;
 		ComboBoxModel m_deviceModel;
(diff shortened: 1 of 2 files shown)


DomClark (approved):

include/AudioSoundIo.h
```diff
@@ -70,7 +70,7 @@ class AudioSoundIo : public AudioDevice
 		return QT_TRANSLATE_NOOP( "AudioDeviceSetupWidget", "soundio" );
 	}
 
-	class setupWidget : public AudioDeviceSetupWidget
+	class setupWidget : public gui::AudioDeviceSetupWidget
```
JohannesLorenz, line 73: Shouldn't this be in the `gui` namespace? This also makes all the `gui::` inside this classe obsolete...

irrenhaus3: This is a nested class inside the class `lmms::AudioSoundIo` and can't be put into a different namespace scope than its container.

JohannesLorenz: Un-nest it :grin:

In the end, I'm OK with both, though I think core/gui separation might be more important than nesting here.

irrenhaus3: Aye, unnesting it would make more sense for separation, but I should note that the other audio interface classes (`AudioDummy`, `AudioJack`, `AudioSdl`, etc.) follow the same pattern as this one, with the `setupWidget` nested inside the main class. Let's separate them in a separate PR for proper separation of these separate tasks.

zonkmachine:
@irrenhaus3 Thanks for the fast fix! There's still something a bit off there though.

```make
In file included from /home/zonkmachine/builds/lmms/build/src/lmmsobjs_autogen/DGKZTCOIDE/moc_AudioSoundIo.cpp:10,
                 from /home/zonkmachine/builds/lmms/build/src/lmmsobjs_autogen/mocs_compilation.cpp:14:
/home/zonkmachine/builds/lmms/build/src/lmmsobjs_autogen/DGKZTCOIDE/../../../../include/AudioSoundIo.h:74:9: error: expected class-name before ‘{’ token
   74 |         {
      |         ^

```
PS. Let it be understood. I wouldn't recognize a namespace if my life depended on it...

zonkmachine (approved):
Compiles fine

PR #6449, opened 2022-06-24
Update `.clang-tidy`

JohannesLorenz:
Should we update this file?

Check the two commits separately to understand what I changed.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,5 +1,12 @@
 ---
-Checks:          'bugprone-macro-parentheses,bugprone-macro-repeated-side-effects,modernize-use-override,readability-identifier-naming,readability-misleading-indentation,readability-simplify-boolean-expr,readability-braces-around-statements'
+Checks: >
+  bugprone-macro-parentheses,
+  bugprone-macro-repeated-side-effects,
+  modernize-use-override,
+  readability-identifier-naming,
+  readability-misleading-indentation,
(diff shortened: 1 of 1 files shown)


irrenhaus3:
`modernize-use-default` can screw up in some places because it's too eager. I had it happen in a scenario like 
```cpp
Foo():
 m_Member1(A_CONSTANT, ANOTHER_CONSTANT),
 m_Member2(YET_ANOTHER_CONSTANT)
 {}
```
which it replaced with 
```cpp
Foo():
 m_Member1(A_CONSTANT, ANOTHER_CONSTANT),
 m_Member2(YET_ANOTHER_CONSTANT)
 = default;
```
which is a syntax error.  
If that happens, it needs manual correction or `git restore`. If we can live with that, LGTM.

JohannesLorenz:
> If that happens, it needs manual correction or `git restore`. If we can live with that, LGTM.

As we agreed to only use `clang-tidy` for manual editing of single files (because, as we see here again, `clang-tidy` is too buggy for automation/CI), I think that's OK. A `git restore` does not seem necessary, IIRC you can just remove 1 or 2 lines in such a case.

irrenhaus3:
Aye, if we only run it manually anyway, go ahead and merge. :+1:

irrenhaus3 (approved):

PR #6450, opened 2022-06-25
clang-tidy: Apply modernize-use-equals-default everywhere

irrenhaus3:
This replaces all empty or trivial implementations of constructors, destructors and assignment operators with `= default;`, explicitly stating that these functions should be automatically generated by the compiler.  
I also ran `modernize-use-equals-delete` on everything, but this only hit 3rdparty submodules. Our code got 0 flags.


include/MidiEvent.h
```diff
@@ -66,16 +66,7 @@ class MidiEvent
...
-		m_type( other.m_type ),
-		m_metaEvent( other.m_metaEvent ),
-		m_channel( other.m_channel ),
-		m_data( other.m_data ),
-		m_sysExData( other.m_sysExData ),
-		m_sourcePort(other.m_sourcePort),
-		m_source(other.m_source)
-	{
-	}
+	MidiEvent( const MidiEvent& other ) = default;
```
irrenhaus3, line 69: Particularly impressed by this one here.

include/DetuningHelper.h
```diff
@@ -43,8 +43,7 @@ class DetuningHelper : public InlineAutomation
 	}
 
 	~DetuningHelper() override
-	{
-	}
+	= default;
```
JohannesLorenz, line 46: Many lines contain only `= default` now. Grep those and move them behind the actual function?

irrenhaus3: Done.

include/DummyEffect.h
```diff
@@ -61,8 +61,7 @@ class DummyEffectControls : public EffectControls
 	}
 
 	~DummyEffectControls() override
-	{
-	}
+	= default;
```
JohannesLorenz, line 64: Nope?

irrenhaus3: No idea what you're talking about. :smirk:

JohannesLorenz (approved):

JohannesLorenz:
Basic tests also successfull.

DomClark (approved):

PR #6451, opened 2022-06-25
clang-tidy: Apply modernize-use-emplace everywhere

irrenhaus3:
For containers with an `emplace` method that constructs new objects directly in the container, this check identifies places where we call `push_back` etc. with a temporary object that's being copied in, and replaces them with calls to `emplace_back` etc.

### DIFF
--- a/include/RemotePluginBase.h
+++ b/include/RemotePluginBase.h
@@ -412,7 +412,7 @@ public:
 		{
 			char buf[32];
 			sprintf( buf, "%d", _i );
-			data.push_back( std::string( buf ) );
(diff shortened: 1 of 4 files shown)


sakertooth (approved):

sakertooth:
Who knows when, but you might have to rerun this check due to the [removal of ``QVector`` from the core](https://github.com/Veratil/lmms/pull/6/) :sweat_smile:. You of course don't have to take any action right this second, but just a heads up in the future.

DomClark (approved):

Issue #4736, opened 2018-12-20
Possible ID collision when saving a project


jasp00:
https://github.com/LMMS/lmms/pull/2875#issuecomment-448790522:
> there might still be the potential here for ID collisions. Say a control with ID `EO_ID_MSB | 0xf00` is automated, then it will be saved with ID `0xf00`. Now, when the project is reloaded, it is possible for another control to be given ID `EO_ID_MSB | 0xf00`, which is now free, and if that were automated, an ID collision would occur when saving the file. The chances of this depend on the number of controls already automated, but `EO_ID_MSB` is 2^23 so they will tend to lie around 1 in a million.

@DomClark, do you want to handle this?

PhysSong:
I think I might be going to fix this later.

jasp00:
Fixing this should make #4723 unnecessary, since you will not need to look for alternate ids.

irrenhaus3:
Reviving this issue because it's been stale for a hot minute and I'm considering working on it.
Would it be worth using Qt-provided [UUIDs](https://doc.qt.io/qt-5/quuid.html) for this? Automation UUIDs would be saved in and restored from the project file and newly created IDs for... well, anything, really - would have only an infinitesimal probability of collision, since they'd span a 128-bit range. Would blow up project file size a bit since you'd now need to add 16 bytes per saved UUID, but would fix collision issues once and for all.

PR #6452, opened 2022-06-26
Journalling UUIDs

irrenhaus3:
Implements a UUID utility class and uses it to provide the type of `jo_id_t`. The intent is to remove any realistic possibility of collision between JournallingObject IDs.  
This addresses issue #4736 and supersedes PRs #2875 and #4723.  
  
**What's been tested to work so far?**
* Saving and loading a new project file with automation.
* Saving and loading a new project file with a LFO Controller connection.
  
**What doesn't work?**
* RemotePlugin
* Loading project files created with any previous version of LMMS (no update function for replacing int32-IDs with UUIDs).
  
**TODOs?**
* Remove workarounds from #4723
* Test more thoroughly (ideally automated)
* Fix anything that doesn't work, obviously.


allejok96:
Does this fix the bug of restoring ID of objects when undoing?

Steps to reproduce:

1. Create notes (you can undo each note)
2. Create a new clip on the same track (track is journalled as a whole)
3. Undo create clip (whole track is reloaded, clip gets new ID)
4. Unable to undo individual notes now (ID of clip does not match journal, so checkpoints are skipped)

PR #6462, opened 2022-07-12
Disable autoscroll by default

LostRobotMusic:
Close this if I'm wrong, but I'm pretty sure most users would prefer to have this disabled by default.

(Like always, I have no idea what I'm doing, but this stack of commits should be fixed by the squash-and-merge option I assume)


Monospace-V:
Personally, I tend to prefer by-default auto-scroll on, because like seeing the part of the song that's playing, more often than seeing a different part. Is it intuitive to have auto-scroll off by default? The only people the defaults matter to, are the new users who don't want to mess with defaults too much. The old users change their workflow as they get more comfortable with using LMMS.
I'd like to request a poll, if one hasn't been done already.

RiedleroD:
autoscroll doesn't save to project files (or templates, for that matter), so I have to change it every time I open a project or create a new one.

YohananDiamond:
Yeah, I think the best would be that it was a per-project option (and, since you can customize the default project template, a custom default for it could be set there)

allejok96:
Editing the default project template feels like a non-intuitive way to change a playback setting. If we just saved the state of the button when LMMS exits, would that be too much of a compromise? (Instead of saving it per-project)

Spekular:
I'm a big fan of the flexibility afforded by saving it per-project, personally. For production work I basically always want autoscroll off, but I have a few completed backing track projects where autoscroll is desireable (for playing along). A per-project setting accomodates that sort of thing very well.

PR #3940, opened 2017-11-05
Switch calf to submodule 

tresf:
First attempt at using a submodule for building Calf.

**Steps**
* [x] Update branch to reflect LMMS changes. Done.... we're somewhere around `0.0.19-rc3` now.
* [ ] Determine if we can get higher than `0.0.60` (`0.90.0` was released yesterday) despite the fact that upstream won't support the LADSPA format any longer. @JohannesLorenz says he's on this.

PR #6463, opened 2022-07-13
fixing "Alf42red-Mauiwowi" project file (by swapping it with version from 0.4x)

kamkamkamkamkamkamkamkam:
For a very long time, the version of "Alf42red-Mauiwowi" bundled with LMMS has been completely broken with all instances of ZynAddSubFX being silent for some unknown reason (despite the project file growing in size somehow). This has been a thing since I think 1.1x but it could've started in 1.2x. I replaced the project with a version from 0.4.15 (which was last modified in 0.4.8) and it plays fine in the latest build of LMMS


Veratil:
> Issue and fix confirmed. I'm curious about when this happened and why.

Probably some old/odd upgrade path that we've since fixed.

zonkmachine:
@kamkamkamkamkamkamkamkam Extra points for removing the unused channels on the mixer (right click any mixer channel -> Remove unused channels).

zonkmachine (approved):

Monospace-V:
Hey,
I can run the demo project in question perfectly on 1.2.0
image

and on 1.3:
image

and on 1.2.2, and 1.1.2, and somewhat ok on 0.4.

I've never noticed this issue.

kamkamkamkamkamkamkamkam:
> Hey, I can run the demo project in question perfectly on 1.2.0 image
> 
> and on 1.3: image
> 
> and on 1.2.2, and 1.1.2, and somewhat ok on 0.4.
> 
> I've never noticed this issue.

Are you sure about that? The mixer in the screenshots you sent suggest otherwise...

kamkamkamkamkamkamkamkam:
Also, is there a reason the checks are failing now suddenly?

zonkmachine:
*Edit* @kamkamkamkamkamkamkamkam Oh dear. I accidentally edited your post instead of making a quoted reply. Reverting under way...

 > Also, is there a reason the checks are failing now suddenly?

This seem to be related to the merging of https://github.com/LMMS/lmms/pull/6174 .

> @Monospace-V Are you sure about that? The mixer in the screenshots you sent suggest otherwise...

You say in the description above that the bugg affects ALL instances of zasfx. This isn't the case for me and doesn't seem to be the case for @Monospace-V . In my case it's only the first and third track that are silent.

It seem to be related to switching the ladspa effects to submodules. In this case the plugin **Calf Filter LADSPA**.
The pad on channel one has a filter on the mixer channel. Switch it off and there is sound.
On the third track, guitar, again, if I switch the filter on the instrument off, there is sound.
@tresf

zonkmachine:
Possible culprits: https://github.com/LMMS/lmms/pull/3940 and https://github.com/LMMS/lmms/pull/3987

Monospace-V:
In my case, the first and third track make sound, but very silent. If I amplify it over a few times it becomes audible.

zonkmachine:
> In my case, the first and third track make sound, but very silent. If I amplify it over a few times it becomes audible. @zonkmachine do see if this happens to be the case for you too?

Yes indeed! If I cram up the volume of my modest headset there is sound.

There could also be an issue related to the LFO Controllers. The calf filters are controlled in the same way and if you turn the **BASE** knob there is once more sound. This time from opening up the filters. If I right click the **Frequency** knob on the filters and choose **Set linear** the issue is fixed.

Monospace-V:
> There could also be an issue related to the LFO Controllers. The calf filters are controlled in the same way and if you turn the **BASE** knob there is once more sound. This time from opening up the filters. If I right click the **Frequency** knob on the filters and choose **Set linear** the issue is fixed.
> 

Did it work this way in previous versions of LMMS too? If yes, then this may have been how it was intended to sound. What version of LMMS is being tested with? 0.4 and 1.1, i can't see this project in demo files.

tresf:
> Possible culprits: https://github.com/LMMS/lmms/pull/3940 and https://github.com/LMMS/lmms/pull/3987

#3940 was essentially a direct clone of 0.60 and a partial rename (from `calf.so` to `veal.so` to avoid namespace conflicts with the ones from common repos).

#3987 was much more involved, was a full backport of the LADSPAv1 interface to the 0.90 Calf codebase, lead largely by @JohannesLorenz.  If this PR is the cause, the upgrade mappings are here if someone can help out: https://github.com/LMMS/lmms/pull/3987/files#diff-5c8fb43716d5630ce8bc7b97df7bd34fa05d92762129de919fd84e0670c27f7d.

zonkmachine:
Testing the project and fix further. Pad below means calf filter on mixer channel 3 and guitar means calf filter on the guitar track (track 3). What differs between versions of lmms-1.2.2 and lmms on master (f39b3d50224dcf92ec065a9d05bd39b34396a827) is whether the filter cutoff knob is linear or logarithmic and if that same knob is connected to a controller or not. This fix seem to be more complete applied to lmms-1.2.2 than to master. You need to also set the cutoff to linear and connect them to the controllers


1.2.2
broken demo
pad logarithmic controller 1 connected
guitar logarithmic controller 2 connected

fixed demo
pad linear controller 1 connected
guitar linear controller 2 connected

master
broken demo
pad logarithmic controller 1 connected
guitar logarithmic controller 2 connected

fixed demo
pad **logarithmic** controller **unconnected**
guitar **logarithmic** controller **unconnected**

zonkmachine (changes requested):
See: https://github.com/LMMS/lmms/pull/6463#issuecomment-1186593698
There seem to be more going on here and quite possibly other projects could be affected. What's missing now is controller connections on the filters and linear response instead of logarithmic. This should maybe be fixed by an upgrade routine instead of a manual fix.

zonkmachine:
> Did it work this way in previous versions of LMMS too? If yes, then this may have been how it was intended to sound. 

As you see from my reply above it seem to be a bit more complicated.

> What version of LMMS is being tested with? 0.4 and 1.1, i can't see this project in demo files.

I've tested on lmms-1.2.2 and on master with the file provided in this fix and also on newer, lmms-1,1.3 which also fixes this, at least on stable-1.2.2. I think the project degraded in at least two steps somewhere after lmms-1.1.3 .

kamkamkamkamkamkamkamkam:
> > In my case, the first and third track make sound, but very silent. If I amplify it over a few times it becomes audible. @zonkmachine do see if this happens to be the case for you too?
> 
> Yes indeed! If I cram up the volume of my modest headset there is sound.
> 
> There could also be an issue related to the LFO Controllers. The calf filters are controlled in the same way and if you turn the **BASE** knob there is once more sound. This time from opening up the filters. If I right click the **Frequency** knob on the filters and choose **Set linear** the issue is fixed.
I think this right here is the actual problem. When you open both instances of Calf Filter and set the "Frequency" knob to "Linear" the project then sounds exactly identical to how it's supposed to sound. I'm gonna try updating this PR with the regular project file with these changes.

PR #2671, opened 2016-03-12
Mallets 2

zonkmachine:
Fixes  #2455
mallets2

Changes from mallets.
 * The knobs work
 * Position was mirrored to reflect actual physics but it was a bit confusing as you get the same result turning the knob left/right. We only need to model from one edge to center.
 * Added a randomness knob which randomises the position and hardness knobs. At full on and with position/hardness in center positions the randomness sweeps across the whole range.
 * Tubular bells removed. Now in Borg Jr.
 * BandedWG ( Uniform bar - Tibetan bowl ) removed.

Issue #6464, opened 2022-07-17
mallets broken on master build


Rossmaxx:
### Bug Summary
i tried using mallets on the latest master, only to find out it's broken. 

#### Steps to reproduce
drop mallets on the song editor or the pattern editor and open instrument settings.

#### Expected behavior
mallets should play fine

#### Actual behavior
it gives the error 'stk not installed'. the instrument opens but doesn't give any sound.

#### Screenshot
![image](https://user-images.githubusercontent.com/74815851/179383502-571c79c1-7ee0-4e8d-8a9c-fa0887495412.png)

https://youtu.be/hLBVz5jSoTM

#### Affected LMMS versions
1.3 master - builds #6462 and #6374 (pre merge ci build)

Monospace-V:
Confirmed.
I faced a similar issue. To make Mallets work, I copied the data/STK/RawWaves file from my 1.2.0 install to where they're supposed to be in the filepath of my 1.3 install, then forgot about it.
I considered reporting it, but I have a slightly unconventional method of installation, and since this is actually a lack-of-files problem, I thought this specific issue might've been avoided if I installed normally.

musikBear:
Mallets issues: 
 Mallets no sound under Tubular Bells #2455 
 (LMMS 1.2.0rc6) Changing the type of Mallets plays weird sounds #4269 
 
Needs an overhaul

zonkmachine:
@musikBear Those are, quite obviously, unrelated issues. Please bring it up on Discord instead.

musikBear:
> @musikBear Those are, quite obviously, unrelated issues. Please bring it up on Discord instead.

It is meant to be a _summery of issues_ with Mallets

zonkmachine:
> > @musikBear Those are, quite obviously, unrelated issues. Please bring it up on Discord instead.
> 
> It is meant to be a _summery of issues_ with Mallets

I understand that. A summary of two issues, one that was closed in 2016 and one that was closed yesterday. It's off topic.

musikBear:
> > > @musikBear Those are, quite obviously, unrelated issues. Please bring it up on Discord instead.
> > 
> > 
> > It is meant to be a _summery of issues_ with Mallets
> 
> I understand that. A summary of two issues, one that was _closed in 2016 and one that was closed yesterday_. It's off topic.

lOOl i diddent see they were closed -sorry, no need for the summery. You are right. 👍

zonkmachine:
[...]

PR #6466, opened 2022-07-20
Mallets - Add random knob function

zonkmachine:
Add a random knob to liven up the sound a bit. Originally introduced as a part of https://github.com/LMMS/lmms/pull/2671.

The values of **Position** and **Hardness** in the first nine instruments and the values of **Modulator** and **Crossfeed** in Tubular Bells are randomly nudged by the value of **Random**.

### DIFF
--- a/plugins/Stk/Mallets/Mallets.cpp
+++ b/plugins/Stk/Mallets/Mallets.cpp
@@ -438,7 +438,7 @@ MalletsInstrumentView::MalletsInstrumentView( MalletsInstrument * _instrument,
 	m_spreadKnob->move( 190, 140 );
 	m_spreadKnob->setHintText( tr( "Spread:" ), "" );
 
-	m_randomKnob = new Knob(knobVintage_32, this);
+	m_randomKnob = new Knob( KnobType::Vintage32, this );
 	m_randomKnob->setLabel(tr("Random"));
 	m_randomKnob->move(190, 190);
 	m_randomKnob->setHintText(tr("Random:"),"");
--- a/plugins/Stk/Mallets/Mallets.cpp
+++ b/plugins/Stk/Mallets/Mallets.cpp
@@ -438,7 +438,7 @@ MalletsInstrumentView::MalletsInstrumentView( MalletsInstrument * _instrument,
 	m_spreadKnob->move( 190, 140 );
 	m_spreadKnob->setHintText( tr( "Spread:" ), "" );
 
-	m_randomKnob = new Knob( KnobType::Vintage32, this );
+	m_randomKnob = new Knob(KnobType::Vintage32, this);
 	m_randomKnob->setLabel(tr("Random"));
 	m_randomKnob->move(190, 190);
 	m_randomKnob->setHintText(tr("Random:"),"");
--- a/plugins/Stk/Mallets/Mallets.cpp
+++ b/plugins/Stk/Mallets/Mallets.cpp
@@ -305,18 +305,18 @@ void MalletsInstrument::playNote( NotePlayHandle * _n,
 		if(p < 9)
 		{
 			hardness += random * static_cast<float>(fast_rand() % 128) - 64.0;
-			hardness = std::clamp(0.0f, hardness, 128.0f);
+			hardness = std::clamp(hardness, 0.0f, 128.0f);
 
 			position += random * static_cast<float>(fast_rand() % 64) - 32.0;
-			position = std::clamp(0.0f, position, 64.0f);
+			position = std::clamp(position, 0.0f, 64.0f);
 		}
 		else if(p == 9)
 		{
 			modulator += random * static_cast<float>(fast_rand() % 128) - 64.0;
-			modulator = std::clamp(0.0f, modulator, 128.0f);
+			modulator = std::clamp(modulator, 0.0f, 128.0f);
 
 			crossfade += random * static_cast<float>(fast_rand() % 128) - 64.0;
-			crossfade = std::clamp(0.0f, crossfade, 128.0f);
+			crossfade = std::clamp(crossfade, 0.0f, 128.0f);
 		}
 
 		// critical section as STK is not thread-safe
--- a/plugins/Stk/Mallets/Mallets.cpp
+++ b/plugins/Stk/Mallets/Mallets.cpp
@@ -310,7 +310,7 @@ void MalletsInstrument::playNote( NotePlayHandle * _n,
 			position += random * static_cast<float>(fast_rand() % 64) - 32.0;
 			position = std::clamp(position, 0.0f, 64.0f);
 		}
-		else if(p == 9)
+		else if (p == 9)
 		{
 			modulator += random * static_cast<float>(fast_rand() % 128) - 64.0;
 			modulator = std::clamp(modulator, 0.0f, 128.0f);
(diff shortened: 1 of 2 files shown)


Rossmaxx:
i know this isnt a fix but i would appreciate if a fix for #6464 was included on this pr.

plugins/Stk/Mallets/Mallets.h
```diff
@@ -255,6 +256,7 @@ public slots:
 
 	ComboBox * m_presetsCombo;
 	Knob * m_spreadKnob;
+	Knob * m_randomKnob;
```
sakertooth, line 259:
```suggestion
	Knob* m_randomKnob;
```

plugins/Stk/Mallets/Mallets.cpp
```diff
@@ -469,7 +500,7 @@ QWidget * MalletsInstrumentView::setupModalBarControls( QWidget * _parent )
 	m_stickKnob->setLabel( tr( "Stick mix" ) );
 	m_stickKnob->move( 190, 90 );
 	m_stickKnob->setHintText( tr( "Stick mix:" ), "" );
-	
+
```
sakertooth, line 503:
```suggestion
```
Redundant I presume?

plugins/Stk/Mallets/Mallets.cpp
```diff
@@ -284,7 +287,7 @@ void MalletsInstrument::playNote( NotePlayHandle * _n,
 	}
 
 	int p = m_presetsModel.value();
-	
+
```
sakertooth, line 290:
```suggestion
```
I think this is redundant.

sakertooth (dismissed):
I suggest using ``<random>`` provided by the C++ standard library instead of ``rand``. It has far more control in terms of PRNG as you can specify the distribution explicitly as well as the random number engine, and it's just more idiomatic to C++ in general. See https://en.cppreference.com/w/cpp/numeric/random for details.

plugins/Stk/Mallets/Mallets.cpp
zonkmachine: I think you've only looked at the diff and not the whole file. The definition of the knobs follow a certain pattern and this hasn't changed. I just removed a trailing tab. A no on this one... ;)

plugins/Stk/Mallets/Mallets.h
zonkmachine: I couldn't find anything about this in the coding conventions.

sakertooth: Sorry, I'll add that these are just suggestions I would enact personally. I tend to follow a certain format, which puts the star right next to the type. If anything, feel free to resolve/ignore this if necessary.

plugins/Stk/Mallets/Mallets.cpp
sakertooth: > I think you've only looked at the diff and not the whole file. The definition of the knobs follow a certain pattern and this hasn't changed. I just removed a trailing tab. A no on this one... ;)

Ah, alright.

plugins/Stk/Mallets/Mallets.cpp
sakertooth: Given the previous discussion, I think this can be resolved as well without action.

zonkmachine:
* Rebased on latest master.
* rand() changed to fast_rand(). No advanced rand needed here.

zonkmachine:
As for the gui, that isn't my cup of tea. You could add some visual indication of how the random knob affects the synth. Like an arrow drawn from the random knob to the affected knobs (updated artwork). And/Or reorganize the knobs so the random knob is situated by the affected knobs. This is a later project.

Ps. The long text of the "Vibrato Frequency" knob disturbs it's position. What about renaming the knobs "Vibrato Gain" ad "Vibrato Frequency" to just "Gain" and "Frequency" and then the hover on message will show the whole line?

plugins/Stk/Mallets/Mallets.cpp
```diff
@@ -293,6 +296,29 @@ void MalletsInstrument::playNote( NotePlayHandle * _n,
...
+		const float random = m_randomModel.value();
+		float hardness = m_hardnessModel.value();
+		float position = m_positionModel.value();
+		float modulator = m_modulatorModel.value();
+		float crossfade = m_crossfadeModel.value();
+
+		if(p < 9)
+		{
+			hardness += random * static_cast<float>(fast_rand() % 128) - 64.0;
+			hardness = std::clamp(0.0f, hardness, 128.0f);
```
DomClark, line 308: Unlike `qBound`, `std::clamp` takes its arguments in the order value, minimum, maximum.

zonkmachine: A non trivial error I'd say. Thanks!

DomClark (approved):

plugins/Stk/Mallets/Mallets.cpp
```diff
@@ -293,6 +296,29 @@ void MalletsInstrument::playNote( NotePlayHandle * _n,
 			m_isOldVersionModel.value() ? 100.0 : 200.0;
 		const float vel = _n->getVolume() / velocityAdjust;
 
+		const float random = m_randomModel.value();
+		float hardness = m_hardnessModel.value();
+		float position = m_positionModel.value();
+		float modulator = m_modulatorModel.value();
+		float crossfade = m_crossfadeModel.value();
+
+		if(p < 9)
```
sakertooth, line 305:
```suggestion
		if (p < 9)
```

plugins/Stk/Mallets/Mallets.cpp
```diff
@@ -412,6 +438,11 @@ MalletsInstrumentView::MalletsInstrumentView( MalletsInstrument * _instrument,
 	m_spreadKnob->move( 190, 140 );
 	m_spreadKnob->setHintText( tr( "Spread:" ), "" );
 
+	m_randomKnob = new Knob( KnobType::Vintage32, this );
```
sakertooth, line 441:
```suggestion
	m_randomKnob = new Knob(KnobType::Vintage32, this);
```

plugins/Stk/Mallets/Mallets.cpp
```diff
@@ -293,6 +296,29 @@ void MalletsInstrument::playNote( NotePlayHandle * _n,
...
+
+		if(p < 9)
+		{
+			hardness += random * static_cast<float>(fast_rand() % 128) - 64.0;
+			hardness = std::clamp(hardness, 0.0f, 128.0f);
+
+			position += random * static_cast<float>(fast_rand() % 64) - 32.0;
+			position = std::clamp(position, 0.0f, 64.0f);
+		}
+		else if(p == 9)
```
sakertooth, line 313:
```suggestion
		else if (p == 9)
```

plugins/Stk/Mallets/Mallets.cpp
```diff
@@ -412,6 +438,11 @@ MalletsInstrumentView::MalletsInstrumentView( MalletsInstrument * _instrument,
 	m_spreadKnob->move( 190, 140 );
 	m_spreadKnob->setHintText( tr( "Spread:" ), "" );
 
+	m_randomKnob = new Knob( KnobType::Vintage32, this );
+	m_randomKnob->setLabel(tr("Random"));
+	m_randomKnob->move(190, 190);
+	m_randomKnob->setHintText(tr("Random:"),"");
```
sakertooth, line 444:
```suggestion
	m_randomKnob->setHintText(tr("Random:"), "");
```

sakertooth (commented):
Just some format updates. The overall code looks okay.

sakertooth (approved):

PR #2668, opened 2016-03-11
Borg Jr - FM synth based on stk::FM

zonkmachine:
borgjr

When I've been poking the code because of issues in Mallet #2455, I came up with this little FM synth. The mallet synths are based on three different sound engines and the one that is least buggy ( In our application n.b. ) is Tubular Bells and as it turns out it has siblings. https://ccrma.stanford.edu/software/stk/classstk_1_1FM.html

I tried them out and liked what I heard.

---

Some issues.
- Crash when using together with Mallets.
- ~~Like Mallets, changes to the settings doesn't update a note that is playing.~~ **_fixed_**
- ~~The ADSR doesn't seem to work on all of the engines. I think this is depending on the stk implication and not on Mallets/Borg Jr. ( _Confirmed. FMVoices :: noteOn writes over the same registers as the ADSR control change so it needs to be set after. Tried it but it doesn't sound very interesting. Could be something more missing there..._ )~~ **_fixed with the updated notes fix_**
- ~~FMVoices - Modulation 2 only goes to 127.9 to counter for a bug in stk.~~ **_fixed_**
- Spread isn't maybe the best on board effect for this one. Basically it's a short delay that only affects the left channel which makes it a bit unnatural sounding when you want to pan the sound to the right as the time shift in itself pans the sound to the left. I'm thinking either adding a delay to the right channel and then you can 'pan' the sound by delaying the signal or adding something else entirely in there like a chorus or phaser.
- I'm not an artist but I'm happy with the icon.

PR #6471, opened 2022-07-27
Mallets - Modalbar, proof of concept

zonkmachine:
mallets

I've attacked Mallets again. The instruments in Stk ModalBar (first nine instruments in Mallets) have settings for each resonator, there's four of them in each instrument, that can be manipulated individually. I've rewired the plugin to use these but had to scrap the instruments from Tubular Bells and below to pull it off because Mallets have been created to use Control Change messages which aren't implemented for the individual voices. Tubular Bells can be reimplemented like I did in an [earlier attempt.](https://github.com/LMMS/lmms/pull/2668) The ones below Tubular Bells, Uniform Bar to Tibetan Bowl, should in my opinion be recognized as the abominations they are and be removed.

TODO:
 - [ ] Replace rand()
 - [ ] Remove the per preset volume modifier (m_scalers)
 - [ ] Automating frequency (or the Fixed/Relative switch) leads to unpleasant results, noise and floating point errors, at fast frequencies. Setting the frequency models lowest relative values to non zero doesn't help to relieve this. A steep high pass filter first in the effect chain relieves this. They should probably be nonzero non the less. The frequency setting should perhaps be interpolated before calling stk.
 - [ ] Move the original presets to the lmms preset system
 - [ ] Rand:
   - [ ] The rand function nudges the values of hardness and position to give a bit of a 'feel' to the sound. The value could be represented on the knob as a semicircle outside of the present value, showing the values covered with rand included.
   - [ ] Hardness, Position and Rand should be on the same line.
 - [ ] Proper artwork
 - [ ] Coding conventions

Test project: [testmallets.mmp.zip](https://github.com/LMMS/lmms/files/9200690/testmallets.mmp.zip)

Edit: This is a proof of concept. Please test the functionality of these changes. I haven't been to particular about coding conventions and artwork. Also, the variables for the separate resonators should perhaps be arrays like in 3osc.


zonkmachine:
PS. This PR includes some tests I did which affects backward compatibility and some should probably be reverted but for a test of basic functionality of accessing the individual resonators this PR should work well. No artwork, just a black background at this point. Also included is the random function from PR #6466.

zonkmachine:
Found some issues. The resonances doesn't cover the full range and the position and hardness knobs seem to be acting up.

zonkmachine:
The preset system should probably be changed. You start with a sound, tweak it, and now the name and volume scaler will stick with the new sound for ever and ever. Not a good solution. It should probably be saved as lmms presets. The original presets need to be there though for backward compatibility and for easier sound tweaking.

zonkmachine:
I loved this one but it's a bit too hard to achieve through stk. At least via ModalBar. Closing for now.

PR #6476, opened 2022-07-31
Migrate CI to GitHub Actions

DomClark:
Every time some issue comes up with CI, moving to GitHub Actions is brought up. Well, here we go.


.github/workflows/checks.yml
```diff
@@ -0,0 +1,26 @@
...
+    - name: Run check-strings
+      run: tests/check-strings/check-strings
+  shellcheck:
+    runs-on: ubuntu-latest
+    container: koalaman/shellcheck-alpine:v0.4.6
+    steps:
+    - name: Check out
+      uses: actions/checkout@v3
+    - name: Run shellcheck
+      run: shellcheck $(find "./cmake/" -type f -name '*.sh' -o -name "*.sh.in")
```
JohannesLorenz, line 26: It would be nice if you could add a check for doc/completions/lmms (and remove the 3rd line from that file, which just explains that you should use shellcheck on this file - which is a job of the CI)

.github/workflows/build.yml
```diff
@@ -0,0 +1,246 @@
...
+        ccache -p
+        echo "[ccache stats]"
+        ccache -s
+  macos:
+    name: macos
+    runs-on: macos-11
+    env:
+      CMAKE_OPTS: -DUSE_WERROR=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo -DUSE_COMPILE_CACHE=ON
+      CCACHE_MAXSIZE: 500M
+      MAKEFLAGS: -j3
```
JohannesLorenz, line 68: It's funny that macos uses 3 jobs, and the other ones use 2. Was there a reason for the difference? Also, was there a reason to downgrade from 6 (in circleci) to 2/3?

JohannesLorenz (commented):
Only 2 minor comments.

Main issues:

* You're using artifacts, but in [your action tab](https://github.com/DomClark/lmms/actions), I can't find out how to download them
* The LMMS bot did not run for this PR. Can we somehow verify that it's still working?

.github/workflows/checks.yml
JohannesLorenz: (The third line of that file is wrong anyways: `-e bash` should be `-s bash` (and the parameter does not seem to be needed anymore)

JohannesLorenz: And a last one that could be added: `buildtools/update_locales`

.github/workflows/build.yml
```diff
@@ -0,0 +1,246 @@
...
+          ccache-${{ github.job }}-${{ github.ref }}-
+          ccache-${{ github.job }}-
+        path: ~/.ccache
+    - name: Configure
+      run: |
+        source /opt/qt5*/bin/qt5*-env.sh || true
+        mkdir build && cd build
+        cmake .. $CMAKE_OPTS -DCMAKE_INSTALL_PREFIX=./install
+    - name: Build
+      run: cd build && make
```
Veratil, line 41: I know this is our defined way to build, but I've been using `cmake --build build --parallel x` lately instead. I wonder if this would allow us to combine build jobs into a yaml anchor we could then just use all over.

Veratil (commented):
Just a few notes, nothing blocking.

.github/workflows/build.yml
```diff
@@ -0,0 +1,246 @@
...
+  linux:
+    name: linux
+    runs-on: ubuntu-18.04
+    container: lmmsci/linux.gcc:18.04
+    env:
+      CMAKE_OPTS: -DUSE_WERROR=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo -DUSE_COMPILE_CACHE=ON
+      CCACHE_MAXSIZE: 500M
+      MAKEFLAGS: -j2
+    steps:
+    - name: Update and configure Git
```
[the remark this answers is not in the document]
DomClark: Didn't realise there was a linter for YAML. Will update all files to be compliant.

.github/workflows/build.yml
DomClark: I'm matching the number of jobs to the number of CPUs specified at https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources.

.github/workflows/build.yml
```diff
@@ -0,0 +1,246 @@
...
+      uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+        submodules: recursive
+    - name: Cache ccache data
+      uses: actions/cache@v3
+      with:
+        key: ccache-${{ github.job }}-${{ github.ref }}-${{ github.run_id }}
+        restore-keys: |
+          ccache-${{ github.job }}-${{ github.ref }}-
```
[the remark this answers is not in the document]
DomClark: Yes - this is to help ensure the prefix matching of cache keys matches the whole of `github.ref`, and ignore cases where one happens to be a prefix of the other. It's not perfect of course - if there's a `-` in the ref, then things can still go wrong - but it helps, and I think it's clearer what the intent is. Also, this is in line with [GitHub's example](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#example-using-the-cache-action).

.github/workflows/build.yml
DomClark: Yeah, I prefer to use CMake like this too, and already used this style for the MSVC jobs. Unfortunately, `--parallel` was only introduced in CMake 3.12, so it isn't available for the Linux and MinGW jobs. I guess one could use `-- -j<x>`, or continue to use the `MAKEFLAGS` environment variable as we do right now. For now, I'll update the non-MSVC platforms to use `cmake --build build` too, and we can introduce the parallelism argument once the docker images are updated.

GitHub actions doesn't currently support YAML anchors (I tried to use them before and got an error message telling me not to), but it may in the future, and consistency never hurts.

DomClark:
> The LMMS bot did not run for this PR. Can we somehow verify that it's still working?

It only works for CircleCI and AppVeyor builds, and I removed the configuration files for those in this PR. In order to download artifacts from GitHub Actions, you need to be logged into GitHub, so posting raw links to artifacts won't work. However there are third-party services that can provide download links, such as [nightly.link](https://nightly.link/). The bot could be changed to link to that site, or as a temporary replacement, https://github.com/oprypin/nightly.link/blob/master/.github/workflows/pr-comment.yml might be worth trying. Eventually I think it would be nice to provide nightly and PR downloads via lmms.io, but that's getting more and more out of scope.

.github/workflows/build.yml
```diff
@@ -0,0 +1,266 @@
...
+          path: Qt
+      - name: Cache vcpkg dependencies
+        uses: actions/cache@v3
+        with:
+          key: vcpkg-${{ matrix.arch }}-${{ github.ref }}-${{ github.run_id }}
+          restore-keys: |
+            vcpkg-${{ matrix.arch }}-${{ github.ref }}-
+            vcpkg-${{ matrix.arch }}-
+          path: C:\vcpkg\installed
+      - name: Install Qt
```
PhysSong, line 213: I use [`jurplel/install-qt-action@v3`](https://github.com/jurplel/install-qt-action/tree/v3) in a private project, so this might help.

.github/workflows/build.yml
```diff
@@ -0,0 +1,266 @@
...
+          $qt_dir = Join-Path $target_dir $qt_version $qt_arch
+          $qt_bin_dir = Join-Path $qt_dir 'bin'
+          echo "Qt5_DIR=$qt_dir" >> $env:GITHUB_ENV
+          echo $qt_bin_dir >> $env:GITHUB_PATH
+      - name: Install dependencies
+        run: |
+          vcpkg install --triplet ${{ matrix.arch }}-windows --recurse `
+            fftw3 libsamplerate libsndfile lilv lv2 sdl2
+      - name: Set up build environment
+        uses: ilammy/msvc-dev-cmd@d8610e2b41c6d0f0c3b4c46dad8df0fd826c68e1
```
PhysSong, line 245: Why commit hash instead of tag name?

.github/workflows/build.yml
```diff
@@ -0,0 +1,266 @@
...
+          submodules: recursive
+      - name: Cache ccache data
+        uses: actions/cache@v3
+        with:
+          key: ccache-${{ github.job }}-${{ github.ref }}-${{ github.run_id }}
+          restore-keys: |
+            ccache-${{ github.job }}-${{ github.ref }}-
+            ccache-${{ github.job }}-
+          path: ~/Library/Caches/ccache
+      - name: Install dependencies
```
PhysSong, line 91: Maybe worth caching `~/Library/Caches/Homebrew`(package download directory) as well, like we do in CircleCI builds.

.github/workflows/build.yml
```diff
@@ -0,0 +1,266 @@
...
+        run: |
+          brew install ccache fftw pkg-config libogg libvorbis lame libsndfile \
+                       libsamplerate jack sdl libgig libsoundio lilv lv2 stk \
+                       fluid-synth portaudio fltk qt5 carla
```
PhysSong, line 95: Not important, but `qt5` is an alias of `qt@5` now.

.github/workflows/build.yml
DomClark: GitHub recommend hashes for security reasons (you can't point them somewhere else without finding a collision, unlike tags): https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

I don't think it's that likely to be an issue though, so I'm happy to switch to a tag if you think that would be better.

PhysSong: That makes sense.

.github/workflows/build.yml
DomClark: Do you think it's worth changing? I just copy-pasted this from the CircleCI configuration.

PhysSong: I prefer `qt@5` in new projects because it's the standard way to mark formula version in Homebrew, but `qt5` is also acceptable as long as it works.

.github/workflows/build.yml
DomClark: I used that initially, but had some trouble with it, so wrote that small bit of PowerShell instead.

I think the problem was that if you installed both 32- and 64-bit Qt, which we need for 32-bit VST plugins on 64-bit Windows, then it sets `Qt5_DIR` incorrectly (it always gets set to the 32-bit Qt directory, no matter which order you install them), which breaks the build. It should be easy enough to work around, but that was after some earlier frustration where I tried to use a newer version which handles caching itself, but it didn't work because that version hadn't been compiled yet. After multiple issues I just gave up.

I'm hoping to remove the Qt requirement for remote plugins on Windows soon, at which point this should all be much easier and I'll probably switch back to it.

PhysSong: > if you installed both 32- and 64-bit Qt, which we need for 32-bit VST plugins on 64-bit Windows, then it sets `Qt5_DIR` incorrectly

https://github.com/jurplel/install-qt-action/blob/d9c92014ea93bab0cd4502c676d5f210d5518eaf/action/src/main.ts#L325 is causing the problem. I think you may work around by installing 64bit Qt first and then installing 32bit Qt with `set-env` set to `false`.

.github/workflows/build.yml
DomClark: Initially I copied the CircleCI config, which caches `~/Library/Caches/Homebrew` and `/usr/local/Homebrew`, but it turned out that GitHub's macOS image has a bunch of stuff preinstalled in Homebrew, which made the cache use about 4GB of our 10GB quota.

If this is just the download directory, I guess it should only contain what we install ourselves, right? I wonder what proportion of the installation time is spent downloading, and how much faster GitHub's cache is. Also, will this grow in size over time as newer package versions are released? Sorry, I don't have much experience with Homebrew.

.github/workflows/build.yml
PhysSong: Note, you should drop manual Qt caching and rely on the mechanism provided by `jurplel/install-qt-action` if you want to try my suggestion.

DomClark: Ah, I see the newer version of that action with built-in caching has been built since I last tried it. Looking forward to being able to use it.

DomClark: I've tried using that action again but it still seems to be broken. Its `action.yml` tries to run `./action` as a step, which is intended to run the action in that directory within install-qt-action, but instead tries to run the action in that directory within LMMS. See jurplel/install-qt-action#153.

.github/workflows/build.yml
DomClark: The cache is still 3.3GB with just that one directory. I'm reverting it again.

.github/workflows/build.yml
PhysSong: I found that the action is broken in several different ways. :sweat_smile: I'm going to report the underlying issues.
In the meantime, you can try using `PhysSong/install-qt-action@v3`

DomClark: Works well now, thanks!

.github/workflows/build.yml
```diff
@@ -0,0 +1,249 @@
...
+        uses: actions/cache@v3
+        with:
+          key: vcpkg-${{ matrix.arch }}-${{ github.ref }}-${{ github.run_id }}
+          restore-keys: |
+            vcpkg-${{ matrix.arch }}-${{ github.ref }}-
+            vcpkg-${{ matrix.arch }}-
+          path: C:\vcpkg\installed
+      - name: Install 64-bit Qt
+        if: matrix.arch == 'x64'
+        uses: PhysSong/install-qt-action@v3
```
PhysSong, line 209: Could you add a comment which says this is a temporary workaround?

PhysSong (approved):

lukas-w:
I've been receiving random emails about failed builds from AppVeyor & CircleCI for years now with no way to turn them off. Finally those days are over. Thank you @DomClark :pray:

PR #6477, opened 2022-08-02
Core Refactor: Replace ``QVector`` with ``std::vector``

sakertooth:
Essentially what was done in Veratil's Qt removal PR, only this time with more modernization (particularly with iterators) and testing. Changes on the GUI side were made as necessary to make the code compile. The remaining ``QVector``'s, if any, in the codebase should be in the GUI, but it is likely that I may have missed something.

The "Phase 2" is mainly because there were changes that needed to be made in order to make the code compile that I had missed on the first run.

Edit: I've decided that I will also remove the remaining ``QVector``'s in the UI part of the codebase, but as a follow-up PR to this one.

### DIFF
--- a/include/AutomationClip.h
+++ b/include/AutomationClip.h
@@ -189,12 +189,12 @@ private:
 	void generateTangents();
 	void generateTangents(timeMap::iterator it, int numToGenerate);
 	float valueAt( timeMap::const_iterator v, int offset ) const;
-	
+
 	/**
-	 * @brief 
+	 * @brief
 	 * This function combines the song tracks, pattern store tracks,
 	 * and the global automation track all in one vector.
-	 * 
+	 *
 	 * @return std::vector<Track*>
 	 */
 	static std::vector<Track*> combineAllTracks();
--- a/include/AutomationClip.h
+++ b/include/AutomationClip.h
@@ -189,6 +189,15 @@ private:
 	void generateTangents();
 	void generateTangents(timeMap::iterator it, int numToGenerate);
 	float valueAt( timeMap::const_iterator v, int offset ) const;
+	
+	/**
+	 * @brief 
+	 * This function combines the song tracks, pattern store tracks,
+	 * and the global automation track all in one vector.
+	 * 
+	 * @return std::vector<Track*>
+	 */
+	static std::vector<Track*> combineAllTracks();
 
 	// Mutex to make methods involving automation clips thread safe
 	// Mutable so we can lock it from const objects
--- a/include/AutomationEditor.h
+++ b/include/AutomationEditor.h
@@ -27,6 +27,7 @@
 #define LMMS_GUI_AUTOMATION_EDITOR_H
 
 #include <QWidget>
+#include <array>
 
 #include "Editor.h"
 
@@ -180,7 +181,7 @@ private:
 	ComboBoxModel m_zoomingYModel;
 	ComboBoxModel m_quantizeModel;
 
-	static const std::vector<float> m_zoomXLevels;
+	static const std::array<float, 7> m_zoomXLevels;
 
 	FloatModel * m_tensionModel;
 
--- a/include/InstrumentFunctions.h
+++ b/include/InstrumentFunctions.h
@@ -119,7 +119,7 @@ public:
 	};
 
 
-	struct ChordTable : public QVector<Chord>
+	struct ChordTable
 	{
 	private:
 		ChordTable();
@@ -131,6 +131,7 @@ public:
 		};
 
 		static std::array<Init, NUM_CHORD_TABLES> s_initTable;
+		std::vector<Chord> m_chords;
 
 	public:
 		static const ChordTable & getInstance()
@@ -149,6 +150,11 @@ public:
 		const Chord & getChordByName( const QString & name ) const
 		{
 			return getByName( name, false );
+		}
+
+		const std::vector<Chord>& chords() const
+		{
+			return m_chords;
 		}
 	};
 
--- a/include/panning.h
+++ b/include/panning.h
@@ -31,6 +31,8 @@
 #include "Midi.h"
 #include "volume.h"
 
+#include <cmath>
+
 namespace lmms
 {
 
@@ -40,7 +42,7 @@ inline StereoVolumeVector panningToVolumeVector( panning_t _p,
 {
 	StereoVolumeVector v = { { _scale, _scale } };
 	const float pf = _p / 100.0f;
-	v.vol[_p >= PanningCenter ? 0 : 1] *= 1.0f - qAbs<float>( pf );
+	v.vol[_p >= PanningCenter ? 0 : 1] *= 1.0f - std::abs(pf);
 	return v;
 }
 
--- a/src/core/AutomatableModel.cpp
+++ b/src/core/AutomatableModel.cpp
@@ -473,7 +473,7 @@ float AutomatableModel::fittedValue( float value ) const
 
 void AutomatableModel::linkModel( AutomatableModel* model )
 {
-	bool containsModel = std::find(m_linkedModels.begin(), m_linkedModels.end(), model) != m_linkedModels.end();
+	auto containsModel = std::find(m_linkedModels.begin(), m_linkedModels.end(), model) != m_linkedModels.end();
 	if (!containsModel && model != this)
 	{
 		m_linkedModels.push_back( model );
@@ -505,8 +505,8 @@ void AutomatableModel::unlinkModel( AutomatableModel* model )
 
 void AutomatableModel::linkModels( AutomatableModel* model1, AutomatableModel* model2 )
 {
-	bool containsModel2 = std::find(model1->m_linkedModels.begin(), model1->m_linkedModels.end(), model2) != model1->m_linkedModels.end();
-	if (!containsModel2 && model1 != model2)
+	auto model1ContainsModel2 = std::find(model1->m_linkedModels.begin(), model1->m_linkedModels.end(), model2) != model1->m_linkedModels.end();
+	if (!model1ContainsModel2 && model1 != model2)
 	{
 		// copy data
 		model1->m_value = model2->m_value;
@@ -723,7 +723,7 @@ void AutomatableModel::reset()
 float AutomatableModel::globalAutomationValueAt( const TimePos& time )
 {
 	// get clips that connect to this model
-	auto clips = AutomationClip::clipsForModel( this );
+	auto clips = AutomationClip::clipsForModel(this);
 	if (clips.empty())
 	{
 		// if no such clips exist, return current value
@@ -734,12 +734,11 @@ float AutomatableModel::globalAutomationValueAt( const TimePos& time )
 		// of those clips:
 		// find the clips which overlap with the time position
 		std::vector<AutomationClip*> clipsInRange;
-
-		for (const auto clip : clipsInRange)
+		for (const auto& clip : clips)
 		{
-			int start = clip->startPosition();
-			int end = clip->endPosition();
-			if (start <= time && end >= time) { clipsInRange.push_back(clip); }
+			int s = clip->startPosition();
+			int e = clip->endPosition();
+			if (s <= time && e >= time) { clipsInRange.push_back(clip); }
 		}
 
 		AutomationClip * latestClip = nullptr;
@@ -754,18 +753,19 @@ float AutomatableModel::globalAutomationValueAt( const TimePos& time )
 		// if we find no clips at the exact time, we need to search for the last clip before time and use that
 		{
 			int latestPosition = 0;
-			for (const auto clip : clips)
+
+			for (const auto& clip : clips)
 			{
-				int end = clip->endPosition();
-				if (end <= time && end > latestPosition)
+				int e = clip->endPosition();
+				if (e <= time && e > latestPosition)
 				{
-					latestPosition = end;
+					latestPosition = e;
 					latestClip = clip;
 				}
 			}
 		}
 
-		if (latestClip)
+		if( latestClip )
 		{
 			// scale/fit the value appropriately and return it
 			const float value = latestClip->valueAt( time - latestClip->startPosition() );
--- a/src/core/AutomatableModel.cpp
+++ b/src/core/AutomatableModel.cpp
@@ -71,7 +71,7 @@ AutomatableModel::~AutomatableModel()
 {
 	while( m_linkedModels.empty() == false )
 	{
-		m_linkedModels.last()->unlinkModel( this );
+		m_linkedModels.back()->unlinkModel(this);
 		m_linkedModels.erase( m_linkedModels.end() - 1 );
 	}
 
@@ -473,7 +473,8 @@ float AutomatableModel::fittedValue( float value ) const
 
 void AutomatableModel::linkModel( AutomatableModel* model )
 {
-	if( !m_linkedModels.contains( model ) && model != this )
+	bool containsModel = std::find(m_linkedModels.begin(), m_linkedModels.end(), model) != m_linkedModels.end();
+	if (!containsModel && model != this)
 	{
 		m_linkedModels.push_back( model );
 
@@ -490,7 +491,7 @@ void AutomatableModel::linkModel( AutomatableModel* model )
 
 void AutomatableModel::unlinkModel( AutomatableModel* model )
 {
-	AutoModelVector::Iterator it = std::find( m_linkedModels.begin(), m_linkedModels.end(), model );
+	auto it = std::find(m_linkedModels.begin(), m_linkedModels.end(), model);
 	if( it != m_linkedModels.end() )
 	{
 		m_linkedModels.erase( it );
@@ -504,7 +505,8 @@ void AutomatableModel::unlinkModel( AutomatableModel* model )
 
 void AutomatableModel::linkModels( AutomatableModel* model1, AutomatableModel* model2 )
 {
-	if (!model1->m_linkedModels.contains( model2 ) && model1 != model2)
+	bool containsModel2 = std::find(model1->m_linkedModels.begin(), model1->m_linkedModels.end(), model2) != model1->m_linkedModels.end();
+	if (!containsModel2 && model1 != model2)
 	{
 		// copy data
 		model1->m_value = model2->m_value;
@@ -588,7 +590,7 @@ float AutomatableModel::controllerValue( int frameOffset ) const
 		return v;
 	}
 
-	AutomatableModel* lm = m_linkedModels.first();
+	AutomatableModel* lm = m_linkedModels.front();
 	if (lm->controllerConnection() && lm->useControllerValue())
 	{
 		return fittedValue( lm->controllerValue( frameOffset ) );
@@ -649,7 +651,7 @@ ValueBuffer * AutomatableModel::valueBuffer()
 		AutomatableModel* lm = nullptr;
 		if (hasLinkedModels())
 		{
-			lm = m_linkedModels.first();
+			lm = m_linkedModels.front();
 		}
 		if (lm && lm->controllerConnection() && lm->useControllerValue() &&
 				lm->controllerConnection()->getController()->isSampleExact())
--- a/src/core/AutomationClip.cpp
+++ b/src/core/AutomationClip.cpp
@@ -888,12 +888,8 @@ gui::ClipView * AutomationClip::createView( gui::TrackView * _tv )
 
 bool AutomationClip::isAutomated( const AutomatableModel * _m )
 {
-	TrackContainer::TrackList l;
-
-	auto& songTracks = Engine::getSong()->tracks();
-	l.insert(l.end(), songTracks.begin(), songTracks.end());
-
-	for (const auto& track : l)
+	auto l = combineAllTracks();
+	for (const auto track : l)
 	{
 		if (track->type() == Track::AutomationTrack || track->type() == Track::HiddenAutomationTrack)
 		{
@@ -924,15 +920,7 @@ bool AutomationClip::isAutomated( const AutomatableModel * _m )
 std::vector<AutomationClip *> AutomationClip::clipsForModel(const AutomatableModel* _m)
 {
 	std::vector<AutomationClip *> clips;
-	TrackContainer::TrackList l;
-
-	auto& songTracks = Engine::getSong()->tracks();
-	l.insert(l.end(), songTracks.begin(), songTracks.end());
-
-	auto& patternStoreTracks = Engine::patternStore()->tracks();
-	l.insert(l.end(), patternStoreTracks.begin(), patternStoreTracks.end());
-
-	l.push_back(Engine::getSong()->globalAutomationTrack());
+	auto l = combineAllTracks();
 
 	// go through all tracks...
 	for (const auto track : l)
@@ -994,15 +982,7 @@ AutomationClip * AutomationClip::globalAutomationClip(
 
 void AutomationClip::resolveAllIDs()
 {
-	TrackContainer::TrackList l;
-
-	auto& tracks = Engine::getSong()->tracks();
-	l.insert(l.end(), tracks.begin(), tracks.end());
-
-	auto& patternStoreTracks = Engine::patternStore()->tracks();
-	l.insert(l.end(), patternStoreTracks.begin(), patternStoreTracks.end());
-
-	l.push_back(Engine::getSong()->globalAutomationTrack());
+	auto l = combineAllTracks();
 	for (const auto& track : l)
 	{
 		if (track->type() == Track::AutomationTrack || track->type() == Track::HiddenAutomationTrack)
@@ -1181,6 +1161,20 @@ void AutomationClip::generateTangents(timeMap::iterator it, int numToGenerate)
 		}
 		it++;
 	}
+}
+
+std::vector<Track*> AutomationClip::combineAllTracks()
+{
+	std::vector<Track*> combinedTrackList;
+
+	auto& songTracks = Engine::getSong()->tracks();
+	auto& patternStoreTracks = Engine::patternStore()->tracks();
+
+	combinedTrackList.insert(combinedTrackList.end(), songTracks.begin(), songTracks.end());
+	combinedTrackList.insert(combinedTrackList.end(), patternStoreTracks.begin(), patternStoreTracks.end());
+	combinedTrackList.push_back(Engine::getSong()->globalAutomationTrack());
+
+	return combinedTrackList;
 }
 
 } // namespace lmms
--- a/src/core/AutomationClip.cpp
+++ b/src/core/AutomationClip.cpp
@@ -770,7 +770,7 @@ void AutomationClip::saveSettings( QDomDocument & _doc, QDomElement & _this )
 	_this.setAttribute( "prog", QString::number( progressionType() ) );
 	_this.setAttribute( "tens", QString::number( getTension() ) );
 	_this.setAttribute( "mute", QString::number( isMuted() ) );
-
+	
 	if( usesCustomClipColor() )
 	{
 		_this.setAttribute( "color", color().name() );
@@ -831,10 +831,10 @@ void AutomationClip::loadSettings( const QDomElement & _this )
 		}
 		else if( element.tagName() == "object" )
 		{
-			m_idsToResolve.push_back(element.attribute( "id" ).toInt());
+			m_idsToResolve.push_back(element.attribute("id").toInt());
 		}
 	}
-
+	
 	if( _this.hasAttribute( "color" ) )
 	{
 		useCustomClipColor( true );
@@ -1055,11 +1055,11 @@ void AutomationClip::objectDestroyed( jo_id_t _id )
 
 	for (auto objIt = m_objects.begin(); objIt != m_objects.end(); objIt++)
 	{
-		Q_ASSERT(!(*objIt).isNull());
-		if((*objIt)->id() == _id)
+		Q_ASSERT( !(*objIt).isNull() );
+		if( (*objIt)->id() == _id )
 		{
 			//Assign to objIt so that this loop work even break; is removed.
-			objIt = m_objects.erase(objIt);
+			objIt = m_objects.erase( objIt );
 			break;
 		}
 	}
--- a/src/core/AutomationClip.cpp
+++ b/src/core/AutomationClip.cpp
@@ -120,19 +120,19 @@ bool AutomationClip::addObject( AutomatableModel * _obj, bool _search_dup )
 {
 	QMutexLocker m(&m_clipMutex);
 
-	if( _search_dup && m_objects.contains(_obj) )
+	if (_search_dup && std::find(m_objects.begin(), m_objects.end(), _obj) != m_objects.end())
 	{
 		return false;
 	}
 
 	// the automation track is unconnected and there is nothing in the track
-	if( m_objects.isEmpty() && hasAutomation() == false )
+	if (m_objects.empty() && hasAutomation() == false)
 	{
 		// then initialize first value
 		putValue( TimePos(0), _obj->inverseScaledValue( _obj->value<float>() ), false );
 	}
 
-	m_objects += _obj;
+	m_objects.push_back(_obj);
 
 	connect( _obj, SIGNAL(destroyed(lmms::jo_id_t)),
 			this, SLOT(objectDestroyed(lmms::jo_id_t)),
@@ -184,7 +184,7 @@ const AutomatableModel * AutomationClip::firstObject() const
 	QMutexLocker m(&m_clipMutex);
 
 	AutomatableModel* model;
-	if (!m_objects.isEmpty() && (model = m_objects.first()) != nullptr)
+	if (!m_objects.empty() && (model = m_objects.front()) != nullptr)
 	{
 		return model;
 	}
@@ -865,9 +865,9 @@ QString AutomationClip::name() const
 	{
 		return Clip::name();
 	}
-	if( !m_objects.isEmpty() && m_objects.first() != nullptr )
+	if (!m_objects.empty() && m_objects.front() != nullptr)
 	{
-		return m_objects.first()->fullDisplayName();
+		return m_objects.front()->fullDisplayName();
 	}
 	return tr( "Drag a control while pressing <%1>" ).arg(UI_CTRL_KEY);
 }
@@ -889,9 +889,9 @@ gui::ClipView * AutomationClip::createView( gui::TrackView * _tv )
 bool AutomationClip::isAutomated( const AutomatableModel * _m )
 {
 	TrackContainer::TrackList l;
-	l += Engine::getSong()->tracks();
-	l += Engine::patternStore()->tracks();
-	l += Engine::getSong()->globalAutomationTrack();
+
+	auto& songTracks = Engine::getSong()->tracks();
+	l.insert(l.end(), songTracks.begin(), songTracks.end());
 
 	for (const auto& track : l)
 	{
@@ -924,13 +924,18 @@ bool AutomationClip::isAutomated( const AutomatableModel * _m )
 std::vector<AutomationClip *> AutomationClip::clipsForModel(const AutomatableModel* _m)
 {
 	std::vector<AutomationClip *> clips;
-	TrackContainer::TrackList tracks;
-	tracks += Engine::getSong()->tracks();
-	tracks += Engine::patternStore()->tracks();
-	tracks += Engine::getSong()->globalAutomationTrack();
+	TrackContainer::TrackList l;
+
+	auto& songTracks = Engine::getSong()->tracks();
+	l.insert(l.end(), songTracks.begin(), songTracks.end());
+
+	auto& patternStoreTracks = Engine::patternStore()->tracks();
+	l.insert(l.end(), patternStoreTracks.begin(), patternStoreTracks.end());
+
+	l.push_back(Engine::getSong()->globalAutomationTrack());
 
 	// go through all tracks...
-	for (const auto& track : tracks)
+	for (const auto track : l)
 	{
 		// we want only automation tracks...
 		if (track->type() == Track::AutomationTrack || track->type() == Track::HiddenAutomationTrack )
@@ -953,7 +958,7 @@ std::vector<AutomationClip *> AutomationClip::clipsForModel(const AutomatableMod
 						}
 					}
 					// if the clips is connected to the model, add it to the list
-					if( has_object ) { clips.push_back(a); }
+					if (has_object) { clips.push_back(a); }
 				}
 			}
 		}
@@ -989,9 +994,13 @@ AutomationClip * AutomationClip::globalAutomationClip(
 
 void AutomationClip::resolveAllIDs()
 {
-	TrackContainer::TrackList l = Engine::getSong()->tracks() +
-				Engine::patternStore()->tracks();
-	l += Engine::getSong()->globalAutomationTrack();
+	TrackContainer::TrackList l;
+
+	auto& tracks = Engine::getSong()->tracks();
+	auto& patternStoreTracks = Engine::patternStore()->tracks();
+	l.insert(tracks.end(), patternStoreTracks.begin(), patternStoreTracks.end());
+
+	l.push_back(Engine::getSong()->globalAutomationTrack());
 	for (const auto& track : l)
 	{
 		if (track->type() == Track::AutomationTrack || track->type() == Track::HiddenAutomationTrack)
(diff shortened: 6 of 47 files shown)


include/panning.h
```diff
@@ -40,7 +42,7 @@ inline StereoVolumeVector panningToVolumeVector( panning_t _p,
 {
 	StereoVolumeVector v = { { _scale, _scale } };
 	const float pf = _p / 100.0f;
-	v.vol[_p >= PanningCenter ? 0 : 1] *= 1.0f - qAbs<float>( pf );
+	v.vol[_p >= PanningCenter ? 0 : 1] *= 1.0f - std::abs(pf);
```
Veratil, line 45: Sneaking this in I see. 😁

sakertooth: Essentially, this commit was necessary so that the code could compile. I'm guessing ``qAbs`` was brought in by the ``QVector`` header.

I'm probably going to do a thorough removal of ``qAbs``, ``qMin``, ``qMax``, etc from the core later :smile:.

src/core/EffectChain.cpp
```diff
@@ -162,10 +162,10 @@ void EffectChain::removeEffect( Effect * _effect )
 
 void EffectChain::moveDown( Effect * _effect )
 {
-	if( _effect != m_effects.last() )
+	if (_effect != m_effects.back())
 	{
-		int i = m_effects.indexOf(_effect);
-		std::swap(m_effects[i + 1], m_effects[i]);
+		auto it = std::find(m_effects.begin(), m_effects.end(), _effect);
+		std::swap(*(it++), *it);
```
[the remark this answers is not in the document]
sakertooth: Ah, I don't think so. It should be ``*(++it)`` I believe. However, it may be cleaner to just use ``std::distance`` and ``std::find`` together to get the index and go from there. Looking back, it's not so easy to tell if this works as intended.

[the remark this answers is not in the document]
Veratil: I mean is it going to affect the next `it` as well? We don't want to increment the pointer just point to the next one.

include/panning.h
Veratil: Ah! Yeah it brought them in then.

src/core/EffectChain.cpp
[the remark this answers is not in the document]
sakertooth: This is most likely undefined behavior, since there is no way of knowing what arguments get evaluated first. I'll keep the original ``std::swap`` line and use ``std::distance`` and ``std::find`` to get the index instead.

Or, better yet, keep using ``it`` but with ``std::advance``.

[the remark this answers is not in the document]
Veratil: I'm rusty on pointers, why won't `*(it+1)` work?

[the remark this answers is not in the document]
sakertooth: ``std::find`` can be defined as (From the cppreference: https://en.cppreference.com/w/cpp/algorithm/find):
```
template< class InputIt, class T >
InputIt find( InputIt first, InputIt last, const T& value );
```

Evidently, the return type for ``std::find`` is of ``InputIt``, which must meet the named requirements of ``LegacyInputIterator``, so there is no guarantee that the iterator returned from ``std::find`` will support integral advances in this manner because a ``LegacyInputIterator`` can't be advanced this way (as seen here: https://en.cppreference.com/w/cpp/named_req/InputIterator). ``std::advance`` however, can advance any iterator fitting the requirements of ``LegacyInputIterator`` by an arbitary amount.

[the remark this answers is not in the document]
sakertooth: In other words, ``std::find`` may return an iterator that meets the bare requirements of a ``LegacyInputIterator``, which technically cannot be advanced that way.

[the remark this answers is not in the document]
PhysSong: > In other words, `std::find` may return an iterator that meets the bare requirements of a `LegacyInputIterator`, which technically cannot be advanced that way.

I think the template argument deduction will give `InputIt = std::vector<T>::iterator` which meets `LegacyRandomAccessIterator`, so itw won't be an issue.

If you're still worried, how about using [`std::next`](https://en.cppreference.com/w/cpp/iterator/next) and [`std::prev`](https://en.cppreference.com/w/cpp/iterator/prev) instead?

src/core/StepRecorder.cpp
```diff
@@ -91,7 +91,7 @@ void StepRecorder::notePressed(const Note & n)
 	StepNote* stepNote = findCurStepNote(n.key());
 	if(stepNote == nullptr)
 	{
-		m_curStepNotes.append(new StepNote(Note(m_curStepLength, m_curStepStartPos, n.key(), n.getVolume(), n.getPanning())));
+		m_curStepNotes.push_back(new StepNote(Note(m_curStepLength, m_curStepStartPos, n.key(), n.getVolume(), n.getPanning())));
```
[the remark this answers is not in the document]
PhysSong: Using `emplace_back` here wouldn't help much because the type of `m_curStepNotes` is `std::vector<Note*>`, not `std::vector<Note>`.

src/core/EffectChain.cpp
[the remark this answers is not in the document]
sakertooth: > > In other words, `std::find` may return an iterator that meets the bare requirements of a `LegacyInputIterator`, which technically cannot be advanced that way.
> 
> I think the template argument deduction will give `InputIt = std::vector<T>::iterator` which meets `LegacyRandomAccessIterator`, so itw won't be an issue.
> 
> If you're still worried, how about using [`std::next`](https://en.cppreference.com/w/cpp/iterator/next) and [`std::prev`](https://en.cppreference.com/w/cpp/iterator/prev) instead?

``std::next`` and ``std::prev`` do seem like better choices in this case though rather than ``std::advance``, since they return the iterator that was incremented/decremented. Thanks for bringing them to my attention.

And it probably would be fine, its just that that may not always be the case, as to which it serves good practice and a reminder (atleast for me) to write code that works given various scenarios that won't always give back the same result.

src/core/StepRecorder.cpp
```diff
@@ -288,26 +288,26 @@ void StepRecorder::removeNotesReleasedForTooLong()
 	int nextTimout = std::numeric_limits<int>::max();
 	bool notesRemoved = false;
 
-	QMutableVectorIterator<StepNote*> itr(m_curStepNotes);
-	while (itr.hasNext())
+	m_curStepNotes.erase(std::remove_if(m_curStepNotes.begin(), m_curStepNotes.end(), [&](StepNote* stepNote)
```
PhysSong, line 291: I'm not sure if this kind of side effects inside `std::remove_if` is okay.

src/core/StepRecorder.cpp
```diff
@@ -288,26 +288,26 @@ void StepRecorder::removeNotesReleasedForTooLong()
...
-			const int timeSinceReleased = stepNote->timeSinceReleased(); // capture value to avoid wraparound when calculting nextTimout
-			if (timeSinceReleased >= REMOVE_RELEASED_NOTE_TIME_THRESHOLD_MS)
+			const int timeSinceReleased = stepNote->timeSinceReleased();
+			shouldRemove = timeSinceReleased >= REMOVE_RELEASED_NOTE_TIME_THRESHOLD_MS;
+
+			if (shouldRemove) 
 			{
 				delete stepNote;
-				itr.remove();
 				notesRemoved = true;
```
PhysSong, line 302: We may remove this and check the return value of `std::remove_if` instead, if we stick to `std::remove_if`.

src/core/ConfigManager.cpp
```diff
@@ -334,27 +334,23 @@ void ConfigManager::addRecentlyOpenedProject(const QString & file)
...
-			if((*it).first == attribute)
+			if (setting.first == attribute)
 			{
-				return (*it).second ;
+				return setting.second;
 			}
 		}
 	}
-	static QString empty;
-	return empty;
```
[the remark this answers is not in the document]
sakertooth: ``ConfigManager::value`` was segfaulting given the ``QVector`` changes. Not exactly sure why, but I believe it had something to do with returning the empty static ``QString``, as to which I just returned "" instead.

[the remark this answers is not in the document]
PhysSong: Do you have a stack trace for the crash without this change? That might help finding the real issue.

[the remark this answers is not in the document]
PhysSong: Also, the 3-parameter overload may return a reference to a local variable after this change, as spotted by CI builds.

src/core/Song.cpp
```diff
@@ -848,7 +848,9 @@ AutomationClip * Song::tempoAutomationClip()
 
 AutomatedValueMap Song::automatedValuesAt(TimePos time, int clipNum) const
 {
-	return TrackContainer::automatedValuesFromTracks(TrackList{m_globalAutomationTrack} << tracks(), time, clipNum);
+	auto globalAutomationTracks = TrackList{m_globalAutomationTrack};
```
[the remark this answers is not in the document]
Veratil: Are you sure? The tracks are still set as hidden.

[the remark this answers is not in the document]
PhysSong: > Are you sure? The tracks are still set as hidden.

I mean, the next line is `globalAutomationTracks.insert(globalAutomationTracks.end(), tracks().begin(), tracks().end());`, so `globalAutomationTracks` will contain normal tracks after it.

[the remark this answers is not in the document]
Veratil: Why would that change their track type though?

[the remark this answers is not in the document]
sakertooth: > > Are you sure? The tracks are still set as hidden.
> 
> I mean, the next line is `globalAutomationTracks.insert(globalAutomationTracks.end(), tracks().begin(), tracks().end());`, so `globalAutomationTracks` will contain normal tracks after it.

I renamed it to ``trackList``. Couldn't think of a better one. I think it works.

src/core/ConfigManager.cpp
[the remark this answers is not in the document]
PhysSong: That sounds more like a dangling reference problem. It's a good idea to return a `std::optional` later, BTW.

src/core/AutomationClip.cpp
```diff
@@ -890,25 +890,20 @@ gui::ClipView * AutomationClip::createView( gui::TrackView * _tv )
...
+			const Track::clipVector & clips = track->getClips();
+			for (const auto clip : clips)
 			{
-				const AutomationClip * a = dynamic_cast<const AutomationClip *>( *j );
-				if( a && a->hasAutomation() )
+				const AutomationClip * a = dynamic_cast<const AutomationClip*>(clip);
+				if (a && a->hasAutomation())
 				{
-					for( objectVector::const_iterator k = a->m_objects.begin(); k != a->m_objects.end(); ++k )
+					for (const auto obj : a->m_objects)
```
PhysSong, line 904: GCC warns about copying `QPointer<AutomatableModel>`, but I think it doesn't hurt here.

src/core/AutomationClip.cpp
```diff
@@ -925,42 +920,38 @@ bool AutomationClip::isAutomated( const AutomatableModel * _m )
...
-				AutomationClip * a = dynamic_cast<AutomationClip *>( *j );
+				AutomationClip * a = dynamic_cast<AutomationClip*>(clip);
 				// check that the clip has automation
 				if( a && a->hasAutomation() )
 				{
 					// now check is the clip is connected to the model we want by going through all the connections
 					// of the clip
 					bool has_object = false;
-					for( objectVector::const_iterator k = a->m_objects.begin(); k != a->m_objects.end(); ++k )
+					for (const auto obj : a->m_objects)
```
PhysSong, line 946: Here as well...

src/core/ConfigManager.cpp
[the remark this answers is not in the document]
PhysSong: I found that [`QMap::operator[] const`](https://doc.qt.io/qt-5/qmap.html#operator-5b-5d-1) returns by value. For `QVector` which uses copy-on-write, this isn't a problem. For `std::vector`, however, it results in a dangling reference when a reference to its element is returned.

For future references, `std::map::operator[]` has no `const` version since `operator[]` may insert a default-constructed value for nonexistent keys. It's possible to use `std::map::at` which throws when the key is not found, though.

src/core/StepRecorder.cpp
```diff
@@ -288,27 +288,28 @@ void StepRecorder::removeNotesReleasedForTooLong()
...
+			else 
 			{
 				nextTimout = min(nextTimout, REMOVE_RELEASED_NOTE_TIME_THRESHOLD_MS - timeSinceReleased);
 			}
 		}
 	}
 
+	m_curStepNotes.erase(std::remove_if(m_curStepNotes.begin(), m_curStepNotes.end(), [](auto stepNote) 
+	{ 
+		return stepNote->timeSinceReleased() >= REMOVE_RELEASED_NOTE_TIME_THRESHOLD_MS;
```
PhysSong, line 310: Aren't we deleted stuff in the loop above? This may result in use after free.

sakertooth: Honestly don't know how I didn't see that :sweat_smile:. I'll write it differently.

src/core/AutomatableModel.cpp
```diff
@@ -732,17 +733,18 @@ float AutomatableModel::globalAutomationValueAt( const TimePos& time )
...
+		std::vector<AutomationClip*> clipsInRange;
+
+		for (const auto clip : clipsInRange) 
 		{
-			int s = ( *it )->startPosition();
-			int e = ( *it )->endPosition();
-			if( s <= time && e >= time ) { clipsInRange += ( *it ); }
+			int start = clip->startPosition();
+			int end = clip->endPosition();
+			if (start <= time && end >= time) { clipsInRange.push_back(clip); }
```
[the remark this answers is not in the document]
sakertooth: Probably should be a ``during`` function, especially if this kind of check happens in various parts around the code so we don't duplicate it. Will keep this in mind.

src/core/InstrumentFunctions.cpp
```diff
@@ -176,12 +176,11 @@ bool InstrumentFunctionNoteStacking::Chord::hasSemiTone( int8_t semi_tone ) cons
...
 
 
-InstrumentFunctionNoteStacking::ChordTable::ChordTable() :
-	QVector<Chord>()
+InstrumentFunctionNoteStacking::ChordTable::ChordTable()
 {
 	for (const auto& chord : s_initTable)
 	{
-		push_back(Chord(chord.m_name, chord.m_semiTones));
+		m_chords.emplace_back(chord);
```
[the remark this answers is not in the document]
sakertooth: Was just about to fix this after noticing the builds failed :laughing:. I came back to this PR a couple of days ago in the middle of fixing merge conflicts and I forgot to check over this error. Thank you for the quick reply.

sakertooth:
I've decided that I'm just going to replace ``QVector`` with ``std::vector`` everywhere. There is no reason to leave ``QVector`` in the codebase as a whole even if this PR has a greater focus on the core.

PhysSong:
> There is no reason to leave `QVector` in the codebase as a whole even if this PR has a greater focus on the core.

One of the biggest differences of `QVector` and `std::vector` that I'm aware of is `QVector` uses copy-on-write and can implicitly share data between multiple containers. I'm not sure how this will affect our codebase, though.

One more thing worth discussing here is: do we want to changes in UI code in the same PR or not?

sakertooth:
> > There is no reason to leave `QVector` in the codebase as a whole even if this PR has a greater focus on the core.
> 
> 
> 
> One of the biggest differences of `QVector` and `std::vector` that I'm aware of is `QVector` uses copy-on-write and can implicitly share data between multiple containers. I'm not sure how this will affect our codebase, though.
> 
> 
> 
> One more thing worth discussing here is: do we want to changes in UI code in the same PR or not?

I think changes to the UI code ultimately had to happen regardless, because I had to make sure the code would compile. However, for the 20 most recent commits I just sent, it might be worthwhile to split this PR up because quite frankly it is **huge**. I've had doubts that this PR will be merged anytime soon just because of how large it is.

Let me know if you want to split this PR personally as well, and I'll do so.

PhysSong:
> Let me know if you want to split this PR personally as well, and I'll do so. My only concern would be if we are going to run into merge conflicts, then I'm mostly not sure.

You can backup this branch and create a new PR from it once this is merged. This won't lead to significant merge conflicts.
One more reason for splitting is to review new changes separately, if required.

src/core/ConfigManager.cpp
```diff
@@ -347,18 +347,8 @@ const QString & ConfigManager::value(const QString & cls,
...
-
 
-
-const QString & ConfigManager::value(const QString & cls,
-				      const QString & attribute,
-				      const QString & defaultVal) const
-{
-	const QString & val = value(cls, attribute);
-	return val.isEmpty() ? defaultVal : val;
+	return val.value_or(defaultVal);
```
PhysSong, line 351: `val` is never set, so this will always be `defaultVal`?

sakertooth: > `val` is never set, so this will always be `defaultVal`?

Thanks for pointing this out. It should be fixed now.

src/core/ConfigManager.cpp
```diff
@@ -347,18 +346,8 @@ const QString & ConfigManager::value(const QString & cls,
...
-
-
-
-const QString & ConfigManager::value(const QString & cls,
-				      const QString & attribute,
-				      const QString & defaultVal) const
-{
-	const QString & val = value(cls, attribute);
-	return val.isEmpty() ? defaultVal : val;
+
```
PhysSong, line 349: Found a trailing whitespace here.
```suggestion

```

sakertooth: Unimportant, but it might be best if it would just not have the line. I guess the code feels more compact I suppose. If anything, it is one lesser line and doesn't really hurt readability.

src/core/ControllerConnection.cpp
```diff
@@ -186,8 +189,11 @@ void ControllerConnection::saveSettings( QDomDocument & _doc, QDomElement & _thi
 		}
 		else
 		{
-			int id = Engine::getSong()->controllers().indexOf( m_controller );
-			if( id >= 0 )
+			auto& controllers = Engine::getSong()->controllers();
+			auto it = std::find(controllers.begin(), controllers.end(), m_controller);
+			int id = std::distance(controllers.begin(), it);
+
+			if (id >= 0)
```
PhysSong, line 196: If `m_controller` is not found in `controllers`, `id` will be -1 in the previous code, but now it wouldn't.

src/gui/clips/ClipView.cpp
```diff
@@ -554,8 +554,9 @@ DataFile ClipView::createClipDataFiles(
 	dataFile.content().appendChild( clipParent );
 
 	// Add extra metadata needed for calculations later
-	int initialTrackIndex = tc->tracks().indexOf( t );
-	if( initialTrackIndex < 0 )
+
+	int initialTrackIndex = std::distance(tc->tracks().begin(), std::find(tc->tracks().begin(), tc->tracks().end(), t));
+	if (initialTrackIndex < 0)
```
PhysSong, line 559: Here as well.

include/AutomationClip.h
```diff
@@ -189,6 +189,15 @@ public slots:
 	void generateTangents();
 	void generateTangents(timeMap::iterator it, int numToGenerate);
 	float valueAt( timeMap::const_iterator v, int offset ) const;
+	
+	/**
+	 * @brief 
+	 * This function combines the song tracks, pattern store tracks,
+	 * and the global automation track all in one vector.
+	 *
```
PhysSong, line 197:
```suggestion
	 *
```

include/InstrumentFunctions.h
```diff
@@ -149,6 +150,11 @@ class InstrumentFunctionNoteStacking : public Model, public JournallingObject
 		{
 			return getByName( name, false );
 		}
+
+		const std::vector<Chord>& chords() const
```
PhysSong, line 154:
```suggestion
		const std::vector<Chord>& chords() const
```

src/core/AutomationClip.cpp
```diff
@@ -1173,4 +1164,18 @@ void AutomationClip::generateTangents(timeMap::iterator it, int numToGenerate)
 	}
 }
 
+std::vector<Track*> AutomationClip::combineAllTracks()
```
PhysSong, line 1167:
```suggestion
std::vector<Track*> AutomationClip::combineAllTracks()
```

src/core/Controller.cpp
```diff
@@ -86,10 +86,10 @@ Controller::Controller( ControllerTypes _type, Model * _parent,
 
 Controller::~Controller()
 {
-	int idx = s_controllers.indexOf( this );
-	if( idx >= 0 )
+	auto it = std::find(s_controllers.begin(), s_controllers.end(), this);
+	if (it != s_controllers.end())
```
PhysSong, line 90:
```suggestion
	if (it != s_controllers.end())
```

src/core/InstrumentFunctions.cpp
```diff
@@ -190,10 +189,12 @@ InstrumentFunctionNoteStacking::ChordTable::ChordTable() :
 
 const InstrumentFunctionNoteStacking::Chord & InstrumentFunctionNoteStacking::ChordTable::getByName( const QString & name, bool is_scale ) const
 {
-	for( int i = 0; i < size(); i++ )
+	for (const auto& chord : m_chords)
```
PhysSong, line 192:
```suggestion
	for (const auto& chord : m_chords)
```

src/core/InstrumentFunctions.cpp
```diff
@@ -190,10 +189,12 @@ InstrumentFunctionNoteStacking::ChordTable::ChordTable() :
 
 const InstrumentFunctionNoteStacking::Chord & InstrumentFunctionNoteStacking::ChordTable::getByName( const QString & name, bool is_scale ) const
 {
-	for( int i = 0; i < size(); i++ )
+	for (const auto& chord : m_chords) 
 	{
-		if( at( i ).getName() == name && is_scale == at( i ).isScale() )
-			return at( i );
+		if (chord.getName() == name && is_scale == chord.isScale())
```
PhysSong, line 194:
```suggestion
		if (chord.getName() == name && is_scale == chord.isScale())
```

src/core/InstrumentFunctions.cpp
```diff
@@ -210,9 +211,10 @@ InstrumentFunctionNoteStacking::InstrumentFunctionNoteStacking( Model * _parent
 	m_chordRangeModel( 1.0f, 1.0f, 9.0f, 1.0f, this, tr( "Chord range" ) )
 {
 	const ChordTable & chord_table = ChordTable::getInstance();
-	for( int i = 0; i < chord_table.size(); ++i )
+
+	for (const auto& chord : chord_table.chords())
```
PhysSong, line 215:
```suggestion
	for (const auto& chord : chord_table.chords())
```

src/core/InstrumentFunctions.cpp
```diff
@@ -308,9 +310,9 @@ InstrumentFunctionArpeggio::InstrumentFunctionArpeggio( Model * _parent ) :
 	m_arpModeModel( this, tr( "Arpeggio mode" ) )
 {
 	const InstrumentFunctionNoteStacking::ChordTable & chord_table = InstrumentFunctionNoteStacking::ChordTable::getInstance();
-	for( int i = 0; i < chord_table.size(); ++i )
+	for (auto& chord : chord_table.chords())
```
PhysSong, line 313:
```suggestion
	for (auto& chord : chord_table.chords())
```

src/core/Mixer.cpp
```diff
@@ -505,12 +508,22 @@ void Mixer::deleteChannelSend( mix_ch_t fromChannel, mix_ch_t toChannel )
 void Mixer::deleteChannelSend( MixerRoute * route )
 {
 	Engine::audioEngine()->requestChangeInModel();
+
+	auto removeFromMixerRoute = [route](MixerRouteVector& routeVec)
```
PhysSong, line 512:
```suggestion
	auto removeFromMixerRoute = [route](MixerRouteVector& routeVec)
```

src/core/PeakController.cpp
```diff
@@ -161,12 +161,11 @@ void PeakController::loadSettings( const QDomElement & _this )
 		effectId = m_loadCount++;
 	}
 
-	PeakControllerEffectVector::Iterator i;
-	for( i = s_effects.begin(); i != s_effects.end(); ++i )
+	for (const auto& effect : s_effects)
```
PhysSong, line 164:
```suggestion
	for (const auto& effect : s_effects)
```

src/core/PeakController.cpp
```diff
@@ -161,12 +161,11 @@ void PeakController::loadSettings( const QDomElement & _this )
 		effectId = m_loadCount++;
 	}
 
-	PeakControllerEffectVector::Iterator i;
-	for( i = s_effects.begin(); i != s_effects.end(); ++i )
+	for (const auto& effect : s_effects) 
 	{
-		if( (*i)->m_effectId == effectId )
+		if (effect->m_effectId == effectId)
```
PhysSong, line 166:
```suggestion
		if (effect->m_effectId == effectId)
```

src/core/PeakController.cpp
```diff
@@ -190,16 +189,14 @@ PeakController * PeakController::getControllerBySetting(const QDomElement & _thi
...
 
-	PeakControllerEffectVector::Iterator i;
-
 	//Backward compatibility for bug in <= 0.4.15 . For >= 1.0.0 ,
 	//foundCount should always be 1 because m_effectId is initialized with rand()
 	int foundCount = 0;
 	if( m_buggedFile == false )
 	{
-		for( i = s_effects.begin(); i != s_effects.end(); ++i )
+		for (const auto& effect : s_effects)
```
PhysSong, line 197:
```suggestion
		for (const auto& effect : s_effects)
```

src/core/StepRecorder.cpp
```diff
@@ -288,26 +288,26 @@ void StepRecorder::removeNotesReleasedForTooLong()
...
+			shouldRemove = timeSinceReleased >= REMOVE_RELEASED_NOTE_TIME_THRESHOLD_MS;
+
+			if (shouldRemove) 
 			{
 				delete stepNote;
-				itr.remove();
 				notesRemoved = true;
 			}
-			else
+			else
```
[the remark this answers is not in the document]
PhysSong:
```suggestion
			else
```

src/core/StepRecorder.cpp
```diff
@@ -288,27 +288,33 @@ void StepRecorder::removeNotesReleasedForTooLong()
...
 			}
-			else
+			else 
 			{
 				nextTimout = min(nextTimout, REMOVE_RELEASED_NOTE_TIME_THRESHOLD_MS - timeSinceReleased);
 			}
 		}
 	}
 
+	m_curStepNotes.erase(std::remove_if(m_curStepNotes.begin(), m_curStepNotes.end(), [](auto stepNote)
```
PhysSong, line 307:
```suggestion
	m_curStepNotes.erase(std::remove_if(m_curStepNotes.begin(), m_curStepNotes.end(), [](auto stepNote)
```

src/core/StepRecorder.cpp
```diff
@@ -288,27 +288,33 @@ void StepRecorder::removeNotesReleasedForTooLong()
...
 			{
 				nextTimout = min(nextTimout, REMOVE_RELEASED_NOTE_TIME_THRESHOLD_MS - timeSinceReleased);
 			}
 		}
 	}
 
+	m_curStepNotes.erase(std::remove_if(m_curStepNotes.begin(), m_curStepNotes.end(), [](auto stepNote) 
+	{
+		bool shouldRemove = stepNote->isReleased() && stepNote->timeSinceReleased() >= REMOVE_RELEASED_NOTE_TIME_THRESHOLD_MS;
+		if (shouldRemove)
```
PhysSong, line 310:
```suggestion
		if (shouldRemove)
```

src/tracks/MidiClip.cpp
```diff
@@ -115,8 +115,8 @@ void MidiClip::resizeToFirstTrack()
 		{
 			if (track != m_instrumentTrack)
 			{
-				unsigned int currentClip = m_instrumentTrack->
-					getClips().indexOf(this);
+				unsigned int currentClip = std::distance(m_instrumentTrack->getClips().begin(),
```
PhysSong, line 118:
```suggestion
				unsigned int currentClip = std::distance(m_instrumentTrack->getClips().begin(),
```

src/tracks/MidiClip.cpp
```diff
@@ -218,17 +218,14 @@ Note * MidiClip::addNote( const Note & _new_note, const bool _quant_pos )
 void MidiClip::removeNote( Note * _note_to_del )
 {
 	instrumentTrack()->lock();
-	NoteVector::Iterator it = m_notes.begin();
-	while( it != m_notes.end() )
+
+	m_notes.erase(std::remove_if(m_notes.begin(), m_notes.end(), [&](Note* note)
```
PhysSong, line 222:
```suggestion
	m_notes.erase(std::remove_if(m_notes.begin(), m_notes.end(), [&](Note* note)
```

src/tracks/MidiClip.cpp
```diff
@@ -354,16 +352,15 @@ void MidiClip::setType( MidiClipTypes _new_clip_type )
 
 void MidiClip::checkType()
 {
-	NoteVector::Iterator it = m_notes.begin();
-	while( it != m_notes.end() )
+	for (auto& note : m_notes)
```
PhysSong, line 355:
```suggestion
	for (auto& note : m_notes)
```

src/tracks/MidiClip.cpp
```diff
@@ -354,16 +352,15 @@ void MidiClip::setType( MidiClipTypes _new_clip_type )
 
 void MidiClip::checkType()
 {
-	NoteVector::Iterator it = m_notes.begin();
-	while( it != m_notes.end() )
+	for (auto& note : m_notes) 
 	{
-		if( ( *it )->length() > 0 )
+		if (note->length() > 0)
```
PhysSong, line 357:
```suggestion
		if (note->length() > 0)
```

PhysSong (commented):
Trailing whitespaces added in this PR:

src/core/ControllerConnection.cpp
```diff
@@ -186,11 +189,10 @@ void ControllerConnection::saveSettings( QDomDocument & _doc, QDomElement & _thi
...
 		{
-			int id = Engine::getSong()->controllers().indexOf( m_controller );
-			if( id >= 0 )
-			{
-				_this.setAttribute( "id", id );
-			}
+			const auto& controllers = Engine::getSong()->controllers();
+			const auto it = std::find(controllers.begin(), controllers.end(), m_controller);
+			const int id = it != controllers.end() ? std::distance(controllers.begin(), it) : -1;
+			_this.setAttribute("id", id);
```
PhysSong, line 195: Still not handling the `-1` case properly. I'd rather write it like this;
```suggestion
			if (it != controllers.end())
			{
						const int id = std::distance(controllers.begin(), it);
						_this.setAttribute("id", id);
			}
```

sakertooth: I was thinking the original code set the ``id`` attribute, even if the controller was not found in the vector. But, after taking another look, you're right, since it was only set when it was found.

Will fix shortly.

PhysSong: Oh, I put some extra indents inside the `if`. Sorry for bothering, but could you fix that as well?

sakertooth:
I will try to fix the conflicts within this PR soon, but the clang-tidy PRs that occurred are creating a lot more conflicts than what it would seem. That being said, it might take some time.

I also want to get rid of those "``QVector``  -> ``std::vector`` UI" commits I pushed and then reverted, as its just clutter (Assuming this will not cause any major problems. Otherwise, I will just leave it as is).

PhysSong:
@sakertooth Could you resolve merge conflicts? Once you do that, I will review this PR.

sakertooth:
> @sakertooth Could you resolve merge conflicts? Once you do that, I will review and test this PR.

@PhysSong Just got done doing that :+1:

src/gui/MixerView.cpp
```diff
@@ -476,12 +476,20 @@ bool MixerView::confirmRemoval(int index)
...
 void MixerView::deleteUnusedChannels()
 {
-	Mixer* mix = Engine::mixer();
+	TrackContainer::TrackList tracks;
+	
+	auto& songTracks = Engine::getSong()->tracks();
+	tracks.insert(tracks.end(), songTracks.begin(), songTracks.end());
+	
+	auto& patternStoreTracks = Engine::patternStore()->tracks();
+	tracks.insert(tracks.end(), patternStoreTracks.begin(), patternStoreTracks.end());
```
PhysSong, line 485: Unused variables, maybe artifacts after resolving conflicts?

src/gui/MixerView.cpp
```diff
@@ -476,12 +476,20 @@ bool MixerView::confirmRemoval(int index)
...
+
+	const auto mixer = Engine::mixer();
 
 	// Check all channels except master, delete those with no incoming sends
 	for (int i = m_mixerChannelViews.size() - 1; i > 0; --i)
 	{
-		if (!mix->isChannelInUse(i))
+		if (!mixer->isChannelInUse(i) && mixer->mixerChannel(i)->m_receives.empty())
```
PhysSong, line 492: Here as well.

src/gui/clips/ClipView.cpp
```diff
@@ -557,8 +557,11 @@ DataFile ClipView::createClipDataFiles(
 	dataFile.content().appendChild( clipParent );
 
 	// Add extra metadata needed for calculations later
-	int initialTrackIndex = tc->tracks().indexOf( t );
-	if( initialTrackIndex < 0 )
+
+	const auto initialTrackIt = std::find(tc->tracks().begin(), tc->tracks().end(), t);
+	const int initialTrackIndex = initialTrackIt != tc->tracks().end() ? std::distance(tc->tracks().begin(), initialTrackIt) : -1;
+
+	if (initialTrackIndex < 0)
```
PhysSong, line 564: How about checking `initialTrackIt != tc->tracks().end()` in `if` and then define `initialTrackIndex` after it?

src/tracks/MidiClip.cpp
```diff
@@ -477,9 +479,9 @@ MidiClip *  MidiClip::nextMidiClip() const
 
 MidiClip * MidiClip::adjacentMidiClipByOffset(int offset) const
 {
-	QVector<Clip *> clips = m_instrumentTrack->getClips();
+	std::vector<Clip *> clips = m_instrumentTrack->getClips();
 	int clipNum = m_instrumentTrack->getClipNum(this);
-	return dynamic_cast<MidiClip*>(clips.value(clipNum + offset, nullptr));
+	return (clipNum < 0 || clipNum > clips.size() - 1) ? nullptr : dynamic_cast<MidiClip*>(clips[clipNum + offset]);
```
PhysSong, line 484: I prefer splitting the conditional to a separate `if` statement which returns `nullptr`.

src/gui/MixerView.cpp
sakertooth: Not sure what happened, but I did a quick rebase with ``master`` to try and solve this.

src/gui/clips/ClipView.cpp
sakertooth: That would make the most sense since the ``if`` is already meant to check if it is not found.. :man_facepalming:. Thanks.

src/core/Mixer.cpp
```diff
@@ -835,15 +848,18 @@ void Mixer::validateChannelName( int index, int oldIndex )
...
 
 	// check if the destination mixer channel on any instrument or sample track is the index mixer channel
 	TrackContainer::TrackList tracks;
-	tracks += Engine::getSong()->tracks();
-	tracks += Engine::patternStore()->tracks();
+
+	auto& songTracks = Engine::getSong()->tracks();
+	auto& patternStoreTracks = Engine::patternStore()->tracks();
+	tracks.insert(tracks.end(), songTracks.begin(), songTracks.end());
+	tracks.insert(patternStoreTracks.end(), patternStoreTracks.begin(), patternStoreTracks.end());
```
DomClark, line 862:
```suggestion
	tracks.insert(tracks.end(), patternStoreTracks.begin(), patternStoreTracks.end());
```

src/core/ConfigManager.cpp
```diff
@@ -347,18 +346,7 @@ const QString & ConfigManager::value(const QString & cls,
...
-}
-
-
-
-const QString & ConfigManager::value(const QString & cls,
-				      const QString & attribute,
-				      const QString & defaultVal) const
-{
-	const QString & val = value(cls, attribute);
-	return val.isEmpty() ? defaultVal : val;
```
DomClark, line 361: The new behaviour is subtly different: if the value is present but empty, the old code returns the default, but the new code returns the empty value. I'm not sure whether this is actually a problem, but it's something to be aware of.

sakertooth: > The new behaviour is subtly different: if the value is present but empty, the old code returns the default, but the new code returns the empty value. I'm not sure whether this is actually a problem, but it's something to be aware of.

I'm not sure either, but I will err on the side of caution and try to match with the original behavior. Changing the behavior of how the code works isn't intended in this PR anyway.

sakertooth: So usually I would stay on the safe side with this, but I've given it more thought. If the value was found but empty, wouldn't it be more reasonable to return that empty value? An attribute set to an empty value in the configuration is still a valid value. Returning ``defaultVal`` would make sense if that attribute was not found, which is what is done here.

src/gui/editors/PianoRoll.cpp
```diff
@@ -406,7 +405,7 @@ PianoRoll::PianoRoll() :
 			InstrumentFunctionNoteStacking::ChordTable::getInstance();
 
 	m_scaleModel.addItem( tr("No scale") );
-	for( const InstrumentFunctionNoteStacking::Chord& chord : chord_table )
+	for (const InstrumentFunctionNoteStacking::Chord& chord : chord_table.chords())
```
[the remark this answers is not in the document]
sakertooth: Care to elaborate? This change seems relevant.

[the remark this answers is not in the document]
PhysSong: I missed that `.chords()` at the end and thought this is just a formatting change. What I meant is:
```suggestion
	for (const auto& chord : chord_table.chords())
```
The same below.

include/AutomationEditor.h
```diff
@@ -180,7 +180,7 @@ protected slots:
 	ComboBoxModel m_zoomingYModel;
 	ComboBoxModel m_quantizeModel;
 
-	static const QVector<float> m_zoomXLevels;
+	static const std::vector<float> m_zoomXLevels;
```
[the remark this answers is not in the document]
sakertooth: For such a small change, I don't mind it being slightly out of scope. Also, if I don't change it now, we might not remember to change it later.

src/gui/editors/PianoRoll.cpp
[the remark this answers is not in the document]
sakertooth: So that in case, I am assuming there is no action that is needed to be taken here, unless I am mistaken.

PhysSong (approved):

include/AutomationEditor.h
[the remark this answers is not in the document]
PhysSong: I also wanted to mention zoom levels in other editors but omitted them.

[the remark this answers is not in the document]
sakertooth: Should I add a TODO comment saying it should be changed in other editors?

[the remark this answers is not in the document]
IanCaio: Could open an issue, to keep track of that future change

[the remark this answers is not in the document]
sakertooth: Seems like the only editor thats not getting the ``std::array`` treatment for zoom levels currently is ``SongEditor``, though I might've missed one. Just a reminder that I do have another PR that targets the ``QVector`` to ``std::vector`` transition more closely for the UI (so the change is probably already done there), but I'm looking towards getting this PR merged first.

DomClark (approved):
LGTM. I haven't checked in-depth the places where the changed types are used, but the diff looks fine. It builds, the tests pass, and it seems to function as normal, so I'm happy for any fallout to be cleared up as we come across it in the future.

Issue #1836, opened 2015-03-08
Feat. req. ARP Sync timings


oeai:
we can change arpeggio period only in time (ms) right now
i propose to sync it with current tempo, so it was like 1\64 1\16 1\2 and so on
i thought that sync mode (Sort, Sync) can do that, but it's not, sync time is still in ms that is very hard to handle with current beat.

softrabbit:
Right-click the time knob and there should be a tempo sync submenu in the context menu.

curlymorphic:
This is already implemented.

If you right click on the time knob, Tempo Sync, choose timing. This could be handled in a better way though

tresf:
Just a guess, but he may be observing the failure to ever round-robin sync from start to finish in conjunction with a major chord and a standard 4/4 measure. (try it out)

I can draw a picture if you'd like.

curlymorphic:
@tresf you mean that a major chord is played in 3/4 time ish? maybe pics would help please.

Oh, now ive sat here for a bit,I can here it drifting.

tresf:
It behaves as described.  It's just math.  Count the number of notes in a major chord **up and then back down**, versus the number of slots... The alignment never works out on a standard measure because the C note only gets played once. :)

curlymorphic:
I agree that is incorrect, but changing that would break backwards compatibility.

tresf:
It is technically correct.  What we could possibly do is increase the chord progression by the appropriate notes to match, but make it an ARP checkbox setting perhaps?

curlymorphic:
> It is technically correct. What we could possibly do is increase the chord progression by the appropriate notes to match, but make it an ARP checkbox setting perhaps?

Yes thats a good idea. we would need to work out how to handle this for various note counts and time signatures.

oeai:
if we take abstract note and divide it into shares (quantitze) without counting current tempo, then this case can be solved with quantization (1-128).
the other way is in recalculating ms timings with tempo \ period change

zonkmachine:
@oeai I think it really sounds like you're describing the sync function we already have as described in the two first answers in this thread.
![temposync](https://cloud.githubusercontent.com/assets/6368949/11919424/c79fb82c-a752-11e5-9031-7c7ff251f677.png)

zonkmachine:
> I can draw a picture if you'd like

[...]

PR #6478, opened 2022-08-05
Arpeggiator - Random distribution

zonkmachine:
Implements different random sources in the arpeggiator, selectable from a new combobox, and adds a knob to control the behaviour.

 Exponential (default) - A negative value make the base note more likely and the top note less likely to be selected. The opposite
 is true for a positive value. Center (0) is equivalent to the old random behavior with equal representation.

 Normal - Normal distribution, a bell curve. A positive value will put the weight on the center note(s) and a negative value will put
 the weight on the base/top notes. Center (0) is equivalent to equal representation.

 Chase - Brownian motion for the keys. Nudges the last note with a random value whose extreme value is set by the knob. On a positive value the base/top notes are stop points and on a negative value the note is allowed to warp around between base/note.

---

I've also made some changes to 'range' that is intended for testing purposes only. For this reason the PR is a draft for now. The range of the arpeggio doesn't go the whole way to the full octave. If you play a scale on an instrument you usually complete a full octave and this is not the case on the lmms arpeggiator. I've tested other arpeggiators and we're not unique in this regard. 
I think the discussion here touches on the matter: https://github.com/LMMS/lmms/issues/1836#issuecomment-77753824
The 'Chase' function won't mind but the other additions is more at home in a full octave context.

https://www.youtube.com/watch?v=wegC5eXN6TU

### DIFF
--- a/include/InstrumentFunctionViews.h
+++ b/include/InstrumentFunctionViews.h
@@ -85,12 +85,14 @@ private:
 	Knob * m_arpRangeKnob;
 	Knob * m_arpRepeatsKnob;
 	Knob * m_arpCycleKnob;
+	Knob * m_arpRandShapeKnob;
 	Knob * m_arpSkipKnob;
 	Knob * m_arpMissKnob;
 	TempoSyncKnob * m_arpTimeKnob;
 	Knob * m_arpGateKnob;
 
 	ComboBox * m_arpDirectionComboBox;
+	ComboBox * m_arpRandomComboBox;
 	ComboBox * m_arpModeComboBox;
 
 } ;
--- a/include/InstrumentFunctions.h
+++ b/include/InstrumentFunctions.h
@@ -207,6 +207,9 @@ enum ArpRandomDistributions
 		return "arpeggiator";
 	}
 
+protected slots:
+
+	void updateNoteRange();
 
 private:
 	enum class ArpMode
--- a/include/InstrumentFunctions.h
+++ b/include/InstrumentFunctions.h
@@ -185,6 +185,14 @@ public:
 		Random
 	} ;
 
+enum ArpRandomDistributions
+	{
+		ArpRandomExp,
+		ArpRandomNormal,
+		ArpRandomChase,
+		NumArpRandomDistributions
+	} ;
+
 	InstrumentFunctionArpeggio( Model * _parent );
 	~InstrumentFunctionArpeggio() override = default;
 
@@ -213,13 +221,15 @@ private:
 	FloatModel m_arpRangeModel;
 	FloatModel m_arpRepeatsModel;
 	FloatModel m_arpCycleModel;
+	FloatModel m_arpRandShapeModel;
 	FloatModel m_arpSkipModel;
 	FloatModel m_arpMissModel;
 	TempoSyncKnobModel m_arpTimeModel;
 	FloatModel m_arpGateModel;
 	ComboBoxModel m_arpDirectionModel;
+	ComboBoxModel m_arpRandomModel;
 	ComboBoxModel m_arpModeModel;
-
+	int m_lastRandomNote;
 
 	friend class InstrumentTrack;
 	friend class gui::InstrumentFunctionArpeggioView;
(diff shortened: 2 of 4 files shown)


allejok96:
Works good! Here's a brain dump:

1. If a bell curve is displaced all the way to the edge (or even a little bit more) wouldn't it be pretty close to an exponential curve - at least by the sound of it? In that case we could have an OFFSET knob instead of the Exponential function.
3. CYCLE and MISS has no use for randomness, right? So it could be hidden or disabled or replaced by OFFSET when random is selected.
4. It would be nice with a more descriptive name for the RAND knob, since we can't see the curve. If the knob has different meaning for different modes, it should change its label and tooltip. For normal distribution it could be "Variance".
5. *If* we stick to different curves of random distribution, the icon should show each curve's shape. You could use something like `filter_bp` for the bell curve and `white_noise_wave_active` for chase.
6. I'd prefer a checkbox for looping around in chase mode instead of negative numbers.
8. If we had a LOOP checkbox  it *could* apply to the bell curve too. E.g. turn OFFSET to max and tick LOOP has the same effect as negative bell curve. That way the RAND knob could be positive only. Why on earth would you want that, you ask... Because when you can't see and understand what each knob does you tend to generalize their meaning. "This is the knob that makes it center around a note". Turning that negative doesn't make sense. Plus I quite like having the knob default all the way to the left and then being able to turn it up to hear the effect it has...
8. Since having the full octave breaks compatibility, maybe we can have a box for that too?

zonkmachine:
> 1\. If a bell curve is displaced all the way to the edge (or even a little bit more) wouldn't it be pretty close to an exponential curve - at least by the sound of it? In that case we could have an OFFSET knob instead of the Exponential function.

Yes, I thought about that but I was afraid people would get tired of too much knobs. I try to keep it as simple as possible. It would be dead simple to implement though.

> 2\. CYCLE and MISS has no use for randomness, right? So it could be hidden or disabled or replaced by OFFSET when random is selected.

CYCLE don't have a use for randomness but MISS do. Set the direction to Up or Down and add some randomness with MISS. I've thought about it but was afraid it would clutter the code to much. Would the knob be served by two different models that it switches between?

> 3\. It would be nice with a more descriptive name for the RAND knob, since we can't see the curve. If the knob has different meaning for different modes, it should change its label and tooltip. For normal distribution it could be "Variance".

You're right. I try and keep the label to four/five letters or the layout is moved around too much. What I think the arpeggiator lacks more than anything now is a manual so I think I'll have a go at documentation and perhaps make a video about it.

> 4\. _If_ we stick to different curves of random distribution, the icon should show each curve's shape. You could use something like `filter_bp` for the bell curve and `white_noise_wave_active` for chase.

Fixing (pushing later). I'm thinking for a permanent solution I'd like icons which the two graph extremes superimposed on each other for the normal and exponential randomness.

> 5\. I'd prefer a checkbox for looping around in chase mode instead of negative numbers.

Yes, or maybe rather keep them as two different menu choices. Real estate fixed...

> 6\. E.g. turn OFFSET to max and tick LOOP has the same effect as negative bell curve.

I have a piece going right now, where I glide seamlessly between the two extremes. I like to keep that the way it is. I like the idea with a second knob better then (comment 1).
 
> Since having the full octave breaks compatibility, maybe we can have a box for that too?

Yes, that's the easiest way to do it. I don't have any ideas that I favor really. A check box is a simple way to do it. I was thinking perhaps making the range knob have every second position being 'range +1 step'. 1, 1+1, 2, 2+1, etc. I don't know. Check box is tempting though.

Epic feedback! :tractor:

zonkmachine:
I've also played with the idea of adding back the original plain random and then inactivate RAND. That's probably a better basic setting.

zonkmachine:
The pattern on Cycle when looping down, down/up was fundamentally wrong from the beginning but I only noticed it now. When starting on a high note the start note would be shifted down when adding cycle to the equation. I've added a fix for this here and the only way I could solve it was to move Cycle and Repeats to within the original direction selection logic. https://github.com/LMMS/lmms/blob/a884205e82abfe36326099a910767cf9efec8671/src/core/InstrumentFunctions.cpp#L472-L505

PR #6480, opened 2022-08-09
clang-tidy: Apply modernize-use-auto everywhere

sakertooth:
This clang-tidy check uses ``auto`` in place of types in variable declarations where applicable. It does this with the goal of improving readability and maintainability, i.e. only used in places that yield a reasonable benefit.

### DIFF
--- a/plugins/VstEffect/VstEffectControls.cpp
+++ b/plugins/VstEffect/VstEffectControls.cpp
@@ -170,7 +170,7 @@ gui::EffectControlDialog* VstEffectControls::createView()
 void VstEffectControls::managePlugin()
 {
 	if ( m_effect->m_plugin != nullptr && m_subWindow == nullptr ) {
-		gui::ManageVSTEffectView * tt = new gui::ManageVSTEffectView( m_effect, this);
+		auto* tt = new gui::ManageVSTEffectView(m_effect, this);
 		ctrHandle = (QObject *)tt;
 	} else if (m_subWindow != nullptr) {
 		if (m_subWindow->widget()->isVisible() == false ) { 
@@ -221,8 +221,8 @@ void VstEffectControls::updateMenu()
     		to_menu->clear();
 
      		for (int i = 0; i < list1.size(); i++) {
-			QAction* presetAction = new QAction(this);
-			connect(presetAction, SIGNAL(triggered()), this, SLOT(selPreset()));
+				auto* presetAction = new QAction(this);
+				connect(presetAction, SIGNAL(triggered()), this, SLOT(selPreset()));
 
         		presetAction->setText(QString("%1. %2").arg(QString::number(i+1), list1.at(i)));
         		presetAction->setData(i);
@@ -291,13 +291,14 @@ void VstEffectControls::rolrPreset()
 void VstEffectControls::selPreset()
 {
 
-     QAction *action = qobject_cast<QAction *>(sender());
-     if (action)
-         if ( m_effect->m_plugin != nullptr ) {
-		lastPosInMenu = action->data().toInt();
-		m_effect->m_plugin->setProgram( lastPosInMenu );
-		//QWidget::update();
-	 }
+	auto* action = qobject_cast<QAction*>(sender());
+	if (action)
+		if (m_effect->m_plugin != nullptr)
+		{
+			lastPosInMenu = action->data().toInt();
+			m_effect->m_plugin->setProgram(lastPosInMenu);
+			// QWidget::update();
+		}
 }
 
 
--- a/plugins/VstEffect/VstEffectControls.cpp
+++ b/plugins/VstEffect/VstEffectControls.cpp
@@ -170,7 +170,7 @@ gui::EffectControlDialog* VstEffectControls::createView()
 void VstEffectControls::managePlugin()
 {
 	if ( m_effect->m_plugin != nullptr && m_subWindow == nullptr ) {
-		gui::ManageVSTEffectView * tt = new gui::ManageVSTEffectView( m_effect, this);
+		auto tt = new gui::ManageVSTEffectView(m_effect, this);
 		ctrHandle = (QObject *)tt;
 	} else if (m_subWindow != nullptr) {
 		if (m_subWindow->widget()->isVisible() == false ) { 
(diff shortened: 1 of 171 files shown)


JohannesLorenz (changes requested):
Only whitespace issues.

Of course, there are sometimes different uses in this PR of `auto` vs `auto*` when it comes to pointers, but this is just an inconsistency, not an error.

Another note: This PR might be better used before the modernize-loop PR, because in conflicting case, the modernize-loop PR has better solutions.

plugins/VstEffect/VstEffectControls.cpp
```diff
@@ -291,13 +291,14 @@ void VstEffectControls::rolrPreset()
 void VstEffectControls::selPreset()
 {
 
-     QAction *action = qobject_cast<QAction *>(sender());
```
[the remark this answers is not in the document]
sakertooth: Not sure I understand, but I ran clang-tidy with --format-style=file, so I figured it would make the necessary formatting changes. Looks like it removed the indent, but do you mean to keep it or something different?

plugins/Watsyn/Watsyn.cpp
```diff
@@ -331,14 +331,8 @@ void WatsynInstrument::playNote( NotePlayHandle * _n,
 {
 	if ( _n->totalFramesPlayed() == 0 || _n->m_pluginData == nullptr )
 	{
-		WatsynObject * w = new WatsynObject(
```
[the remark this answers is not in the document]
sakertooth: I find it odd how it doesn't entirely respect the ``.clang-format``. With so many files to deal with, it's somewhat plausible that it missed something relating to format that we couldn't catch. I'm planning to do a rerun anyways with ``RemoveStars`` set to ``true``, so I'll just come back to check if the formatting is okay after.

plugins/VstEffect/VstEffectControls.cpp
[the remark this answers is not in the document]
JohannesLorenz: Hmm the github editor made it look worse than it actually is. Anyways, I think `if(action)` must have `{}` due to our coding rules. Though you can also join both `if`-statements together.

Anyways, I can also live with how it is now. I just think it looks a bit weird.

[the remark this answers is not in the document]
sakertooth: > Hmm the github editor made it look worse than it actually is. Anyways, I think `if(action)` must have `{}` due to our coding rules. Though you can also join both `if`-statements together.
> 
> Anyways, I can also live with how it is now. I just think it looks a bit weird.

Ah, I see. At first I thought you were just referring to just the ``QAction *action = qobject_cast<QAction *>(sender());`` line. Looks like the code that follows has always been misformatted that way then. I don't think it would hurt for me to fix it and join the two ``if``-statements though.

JohannesLorenz (approved):
Code and test review passed.

src/gui/clips/ClipView.cpp
```diff
@@ -425,7 +425,7 @@ void ClipView::setColor(const QColor* color)
 void ClipView::dragEnterEvent( QDragEnterEvent * dee )
 {
 	TrackContentWidget * tcw = getTrackView()->getTrackContentWidget();
-	TimePos clipPos = TimePos( m_clip->startPosition() );
+	auto clipPos = TimePos(m_clip->startPosition());
```
JohannesLorenz, line 428: Maybe just `TimePos clipPos(m_clip->startPosition());`?

src/gui/clips/ClipView.cpp
```diff
@@ -465,7 +465,7 @@ void ClipView::dropEvent( QDropEvent * de )
 	if( m_trackView->trackContainerView()->allowRubberband() == true )
 	{
 		TrackContentWidget * tcw = getTrackView()->getTrackContentWidget();
-		TimePos clipPos = TimePos( m_clip->startPosition() );
+		auto clipPos = TimePos(m_clip->startPosition());
```
JohannesLorenz, line 468: same

src/gui/clips/ClipView.cpp
```diff
@@ -1207,7 +1205,7 @@ void ClipView::paste()
 	using namespace Clipboard;
 
 	// If possible, paste the selection on the TimePos of the selected Track and remove it
-	TimePos clipPos = TimePos( m_clip->startPosition() );
+	auto clipPos = TimePos(m_clip->startPosition());
```
JohannesLorenz, line 1208: same

src/gui/editors/PianoRoll.cpp
```diff
@@ -988,20 +988,20 @@ void PianoRoll::drawNoteRect( QPainter & p, int x, int y,
...
-	float const leftPanSpan = static_cast<float>(PanningRight - n->getPanning());
-	float const rightPanSpan = static_cast<float>(n->getPanning() - PanningLeft);
+	auto const panningRange = static_cast<float>(PanningRight - PanningLeft);
+	auto const leftPanSpan = static_cast<float>(PanningRight - n->getPanning());
+	auto const rightPanSpan = static_cast<float>(n->getPanning() - PanningLeft);
 
 	float leftPercent = qMin<float>( 1.0f, leftPanSpan / panningRange * 2.0f );
 	float rightPercent = qMin<float>( 1.0f, rightPanSpan / panningRange * 2.0f );
 
-	QColor col = QColor( noteCol );
```
JohannesLorenz, line 1004: `QColor col(noteCol);` ?

JohannesLorenz (approved):
Only minor comments/questions.

src/gui/clips/ClipView.cpp
sakertooth: Sure, but I'll use ``{}``, rather than ``()``, as per https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es23-prefer-the--initializer-syntax. I think we should start using this kind of initialization, given that we would want to adhere to the guidelines, and is just a better to initialize in general. I wouldn't consider this a hard requirement by any means however, but just something we should prefer when given the opportunity.

JohannesLorenz:
Rework is OK :+1:

DomClark (approved):

PR #6486, opened 2022-08-14
Blacklisting plugins depending on gui

zonkmachine:
Many LV2 plugins can be loaded in lmms that more or less lack functionality because of lack of gui support. Blacklist for now.


src/core/lv2/Lv2Manager.cpp
```diff
@@ -58,7 +58,77 @@ const std::set<const char*, Lv2Manager::CmpStr> Lv2Manager::pluginBlacklist =
...
+	"https://github.com/klangfreund/SpectrumAnalyser",
+	"https://github.com/klangfreund/lufsmeter",
+	"https://github.com/laixinyuan/StereoSourceSepartion",
+	"urn:juce:TalFilter2",
+	"urn:juce:Vex",
+
+	// unstable
+	"urn:juced:DrumSynth"
+
+	// need capacity to load files
```
JohannesLorenz, line 122: Just to understand it: To load files, you need something like a GUI button to load them, and that one is only inside the UI, right?

JohannesLorenz (approved):

src/core/lv2/Lv2Manager.cpp
```diff
@@ -58,7 +58,77 @@ const std::set<const char*, Lv2Manager::CmpStr> Lv2Manager::pluginBlacklist =
...
+	"http://lsp-plug.in/plugins/lv2/profiler_stereo",
+	"http://invadarecords.com/plugins/lv2/meter",
+	"http://guitarix.sourceforge.net/plugins/gxtuner#tuner",
+	"https://github.com/jpcima/ADLplug",
+	"https://github.com/HiFi-LoFi/KlangFalter",
+	"https://github.com/klangfreund/SpectrumAnalyser",
+	"https://github.com/klangfreund/lufsmeter",
+	"https://github.com/laixinyuan/StereoSourceSepartion",
+	"urn:juce:TalFilter2",
+	"urn:juce:Vex",
```
JohannesLorenz, line 117: With "worker", we will get ZynAddSubFX, too, so can you please add "http://zynaddsubfx.sourceforge.net" ?

src/core/lv2/Lv2Manager.cpp
zonkmachine: I guess that's what I mean. I don't know how it's supposed to work but I think you need to load a reverb profile into the plugin. I've moved them to the rest of the blacklisted 'needs gui'. No reason to separate them.

JohannesLorenz:
Rework looks good. Do you want to include plugins from the Worker PR here, too, or merge this PR as it is?

zonkmachine:
> Do you want to include plugins from the Worker PR here, too, or merge this PR as it is?

That would be those Gx plugins. Maybe wait for now and test them some more. I'd be surprised if we don't find more plugins that needs to be 'contained' later anyway. I think this PR can be merged as is.

JohannesLorenz:
OK, thanks for the contribution. Merging now.

PR #6498, opened 2022-08-26
PianoRoll: New edit mode "PitchShift" micro-tunes notes with single mouse clicks.

Taiten89:
This would make a great addition to the current microtuning capabilities of LMMS (i.e., Microtuner). This also enables users to experiment with microtonal intervals without the hassle of the automation editor. It was tested to work correctly with different setups of preexisting detuning data.

### DIFF
--- a/include/AutomationClip.h
+++ b/include/AutomationClip.h
@@ -177,6 +177,8 @@ public:
 	static int quantization() { return s_quantization; }
 	static void setQuantization(int q) { s_quantization = q; }
 
+	void shift(float amt);
+
 public slots:
 	void clear();
 	void objectDestroyed( lmms::jo_id_t );
--- a/include/PianoRoll.h
+++ b/include/PianoRoll.h
@@ -34,6 +34,7 @@
 #include "ComboBoxModel.h"
 #include "SerializingObject.h"
 #include "Note.h"
+#include "LcdSpinBox.h"
 #include "lmms_basics.h"
 #include "Song.h"
 #include "StepRecorder.h"
@@ -109,6 +110,7 @@ public:
 		ModeErase,
 		ModeSelect,
 		ModeEditDetuning,
+		ModeEditDetuningShift,
 		ModeEditKnife
 	};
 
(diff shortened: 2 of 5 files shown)


allejok96:
This is a really good idea with a lazy solution (I don't blame you).

1. The left/right click design doesn't feel that polished, or even similar to other elements of LMMS (knobs for example). It would be nicer with a mouse drag. I know it would take a lot of more coding, and PianoRoll is horrible horrible I know that too. 

1. Ability to erase/reset detune easily. Right mouse button?

1. When resizing notes, the detune line keeps its old size (see image). Quick fix would be to only give the detune pattern a single point.

bild

Now if even talk about mouse drag to detune 100 people will request full detune editing inside PianoRoll with that tool. At least ability to create slopes. How do we determine if that tool should edit the start or the end? What happens to the slope if we resize a note? How do we perform a really fine detune as opposed to 1:1 in the grid? The questions are numerous, and headbutting will be even worse than coding it (or maybe not considering the state of PianoRoll again).

To avoid said headbutting I suggest having microtuning as a third option for the velocity/panning area. That is a quick way to edit, gives you a good overview, ability to detune multiple notes at a time, and give you scroll wheel support

I'd implement it like so:

- The slider shows the detune of the first automation point.
-  Moving the slider should move all points (like your current implementation does).
-  By default the range should be 1-100 cents. If you want more you can move the note or use automation editor.
-  If detune exceeds +/- 100 cents (due to using the automation editor), we only show the modulo (e.g. a initial detune of +120 cents shows up as +20 in the microtuning field).

zonkmachine:
> 1. The left/right click design doesn't feel that polished, or even similar to other elements of LMMS (knobs for example). It would be nicer with a mouse drag. I know it would take a lot of more coding, and PianoRoll is horrible horrible I know that too.

Maybe some keyboard shortcuts?
It's `<Ctrl> + arrow up/down` to shift one octave up/down.
It's `<Shift> + arrow up/down` to shift one semi tone up/down.
Maybe `<Ctrl> + <Shift> + arrow up/down` to nudge tuning? Hold keys down a longer time and the speed of change increase/decrease.

CLandel89:
> To avoid said headbutting I suggest having microtuning as a third option for the velocity/panning area. That is a quick way to edit, gives you a good overview, ability to detune multiple notes at a time, and give you scroll wheel support

TBH, I like that approach more than my own, and: This could be completely unrelated to detuning. So the implementation should not touch automation models and instead modify the frequency passed to <code>processFrame</code>. That would save a LOT of hassle, for both users and developers.

PR #6507, opened 2022-09-09
Support Sf2 Player, Mallets, PortAudio, and Lame with MSVC

DomClark:
As in #6132, this is done by modifying the CMake code to consume these as imported targets, which can come either from vcpkg or from our own find modules. The FluidSynth port in vcpkg doesn't come with CMake support, but I have written a find module that produces an imported target anyway, as this is the preferred modern approach.

Changes:
* Added a find module for FluidSynth.
* Modified the find modules for STK, PortAudio, and LAME to create imported targets, first checking whether config mode produces such a target already.
* Moved the config mode preference logic for SDL2 from the root `CMakeLists.txt` to the find module, for consistency with the others. I feel this is better than cluttering the list file with logic related to finding packages, which is the whole point of find modules.
* Cleaned up the SDL2 find module since I was working on it anyway. I've removed all the unused bits, and modernised the code style.
* Set the sndio status for Windows builds to our standard unsupported platform message; previously it was blank, unlike all of the other optional feature statuses.
* Introduced a helper function for finding the appropriate static (import) libraries in vcpkg for different build types. This is useful when CMake integration is not provided. I used it for FluidSynth, but it could be useful for FFmpeg, for example.
* Added specific search hints for MinGW to SDL2. The config mode script for SDL2 on MinGW doesn't provide a target, but does offer directories to look in. We might as well take advantage of those.
* Set the host triplet for vcpkg to match the target triplet. Building FluidSynth requires building a code generation tool to generate some headers, and that tool shares a number of dependencies with FluidSynth. Matching the host and target triplets means only one version of these dependencies needs to be built, which cuts down the build time significantly.
* Set `CMAKE_NO_SYSTEM_FROM_IMPORTED` for MinGW builds. Some headers of libstdc++ are sensitive to the include search order, which can be changed by specifying system headers. Treating imported target headers as standard includes avoids modifying the order inadvertently.

I think this only leaves the LADSPA plugins, ZynAddSubFx, and Gig Player missing from MSVC builds now.


CMakeLists.txt
```diff
@@ -22,6 +22,14 @@ ENDIF(COMMAND CMAKE_POLICY)
 # Import of windows.h breaks min()/max()
 ADD_DEFINITIONS(-DNOMINMAX)
 
+# Treating imported target headers as system includes can change the header
+# search order if some targets have headers in the standard system include
+# directories, which can cause some headers not to be found due to the way
+# #include_next is used in the standard library.
+if(MINGW)
+	set(CMAKE_NO_SYSTEM_FROM_IMPORTED TRUE)
```
PhysSong, line 30: Does this mean the include directories from `IMPORTED` targets will use `-I` instead of `-isystem`, like when we add directories with `include_directories` or  `target_include_directories` without `SYSTEM` option? Then why do we need to turn on this only in MinGW builds?

cmake/modules/FindSDL2.cmake
```diff
@@ -68,110 +28,91 @@
...
+		endif()
+
+		# For threads, as mentioned Apple doesn't need this.
+		# In fact, there seems to be a problem if I used the Threads package
+		# and try using this line, so I'm just skipping it entirely for OS X.
+		if(NOT APPLE AND Threads_FOUND)
+			set_property(TARGET SDL2::SDL2 APPEND PROPERTY
+				INTERFACE_LINK_LIBRARIES "Threads::Threads"
+			)
+		endif()
```
PhysSong, line 101: This kind of stuff is skipped when SDL2 is found in `CONFIG` mode. Is it safe?

CMakeLists.txt
DomClark: That's exactly what it does. I've no idea why it's only needed for MinGW, as I would expect Linux builds to have the same problem. I guess the package headers and standard library headers are arranged differently. I'll investigate, but I don't think this is a problem for us.

cmake/modules/FindSDL2.cmake
DomClark: One would hope that the config script, since it's provided by SDL2 itself, would add whatever is necessary to use SDL2 on the current system. If not, I would consider that an upstream bug. Also, this was already skipped if SDL2 was found in config mode, since the root `CMakeLists.txt` didn't use module mode if config mode succeeded. The logic is the same, except it's all in the module now.

PhysSong: > One would hope that the config script, since it's provided by SDL2 itself, would add whatever is necessary to use SDL2 on the current system.

I checked `INTERFACE_LINK_LIBRARIES` of `SDL2::SDL2` in various build environment. Other than the SDL2 library itself, the property contains:
- Linux: nothing
- macOS(via Homebrew): nothing
- MinGW(via PPA): `-lmingw32 -lSDL2main -mwindows`

I'm not sure if every libraries in this CMake module are required, but I'm pretty sure this CMake module is doing some different from the upstream config script. Not sure which one(maybe both?) is doing wrong...

> Also, this was already skipped if SDL2 was found in config mode, since the root `CMakeLists.txt` didn't use module mode if config mode succeeded.

Alright, it's at least not a regression then.

PhysSong (approved):

PR #6512, opened 2022-09-17
Fix blank lines in language dropdown menu

### LINKED ISSUE #6506: Language setting shows a long~ empty list
Edit -> Setting 
When change language will show a long empty list, and I research the folder of locale found the problem.

under \LMMS\data\locale there are many files seems not read as effective language file, so it shows nothing.
remove all file which name start with 'qt' out of the folder (\LMMS\data\locale) will resolve this problem.

allejok96:
Fallback to English if there is no native language name, or to the filename of the translation file if there is no English name.

Fix #6506


src/gui/modals/SetupDialog.cpp
```diff
@@ -1352,4 +1357,4 @@ void SetupDialog::showRestartWarning()
 }
 
 
-} // namespace lmms::gui
\ No newline at end of file
+} // namespace lmms::gui
```
JohannesLorenz, line 1360: Is this a new trend to remove those blank lines? :grin:

src/gui/modals/SetupDialog.cpp
```diff
@@ -267,7 +267,12 @@ SetupDialog::SetupDialog(ConfigTabs tab_to_open) :
 		// Get locale extracted by filename.
 		fileNames[i].truncate(fileNames[i].lastIndexOf('.'));
 		m_languages.append(fileNames[i]);
-		QString lang = QLocale(m_languages.last()).nativeLanguageName();
+		QLocale locale(m_languages.last());
+		QString lang = locale.nativeLanguageName();
+		// If the language code is invalid, display the filename
+		if (locale.language() == QLocale::C) { lang = fileNames[i]; }
```
JohannesLorenz, line 273: So you display the `qt_` files as their filename? I understood this like they can not be used as a language,so why not hide them completely?

src/gui/modals/SetupDialog.cpp
allejok96: I think I have Qt creator set to add a newline at EOF when saving

src/gui/modals/SetupDialog.cpp
PhysSong: > So you display the `qt_` files as their filename? I understood this like they can not be used as a language,so why not hide them completely?

I think so.

allejok96: Good point, done

JohannesLorenz: I think the fix is invalid, because now, `m_languages` and `changeLang` will have different sizes and the indices don't match. The mismatching indices would bring errors, e.g. in 

```c++
	connect(changeLang, SIGNAL(currentIndexChanged(int)),
			this, SLOT(setLanguage(int)));
```

allejok96: Thanks for catching that! Fixed.

JohannesLorenz (approved):
Code looks good. Testing worked (except for the unrelated bug, that the newly selected language is unavailable in the GUI).

PR #6513, opened 2022-09-18
check-namespace: don't require comment after #else

allejok96:
Quick fix suggested by @JohannesLorenz


JohannesLorenz (approved):
Code is OK and fixes my problem in branch `lv2-worker`. IMO it can be merged without 2nd review.

@allejok96 I appended a commit that verifies your change. It shows no errors, but if I remove the `#else` in that new test, it does. So I think this should be OK.

PR #6517, opened 2022-09-27
Added setting to allow expanded sidebar

adam-eldredge:


Veratil:
Any linked issue? Can you give some more detail in what this PR is for? Before/after picture?

allejok96:
What happens if you decrease the size of the window? Will the text be cut off with `...` or perhaps it may be possible to revert to small buttons? (Yeah now I made everything complicated)

adam-eldredge:
As of now with these changes, the buttons that get cut off disappear. I think it would be a great idea to revert back to the smaller buttons, but as of right now I am super busy with classes. When I get more free time, I could look at it and see if I could figure out how to make that work.

PR #6522, opened 2022-10-01
Fix bug that shows only one effect in the effect chain

sakertooth:
#6481  Introduced a bug where only one effect would show up in the effect chain, even though you could audibly tell that the effect was still added. This PR fixes this bug.

### DIFF
--- a/src/gui/EffectRackView.cpp
+++ b/src/gui/EffectRackView.cpp
@@ -165,6 +165,7 @@ void EffectRackView::update()
 				view_map[i] = true;
 				break;
 			}
+			++i;
 		}
 		if( i >= m_effectViews.size() )
 		{


tresf:
Explanation from discord: 

> what happened was that the original code was a iterator-based loop that incremented both the iterator and some variable i using the comma operator
> 
> so `i` was incrementing alongside the iterator, but #6481 didn't increase `i` as it went through the loop

sakertooth:
Just going to explain this a bit more succinctly. The initial explanation I provided is rushed.

The original code was an iterator-based loop that not only incremented the iterator ``vit``, but also another integer variable ``i`` that was declared before the loop. This bug arose because the new ranged-based for-loop did not increment ``i`` as it was looping, leaving ``i`` to be 0 for the duration of the loop. With ``i`` staying at 0, this affected not only the ``if`` statement that used ``i`` within the loop, but also the code that followed the ranged-based for loop.

This "parallel" increment of ``vit`` and ``i`` was done through the use of the comma operator, which is probably why it was hard to see that there was an issue in the first place.

PR #6525, opened 2022-10-02
remove filepath from VST effect name

### LINKED ISSUE #5748: Effect modules show folder names instead of just the name of the effect
When you setup your effect folder location in settings, if that folder has folders within it, the names of these folders will show up in the Effect Module names        <folder name>\<effect name>
For example       SPACE\REVERB\Plate Reverb

Just in case this does not make sense, I am referring to the Effect modules that show up in the Mixer - Effect Chains as well as the ones in Instruments - Effects tab.
The white text name at the bottom of each effect module inserted into them.

Think I would prefer not to show the folder path within the names
This is more of just a preference, not sure if others will agree?

eliaslk:
changed line 73 to remove filepath from VST names in effect rack ( fixes #5748 )


DomClark (commented):
Thanks for the pull request!

A couple of tips:
* If you add `Fixes #5748` or similar to the PR description, GitHub will automatically close the associated issue when the PR is merged. See [the GitHub documentation](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).
* Not something to change for this pull request, but it's worth creating new branches for your pull requests, rather than working on the master branch. I won't explain it here (it'll take too long, and there are better explanations out there than I can write), but there are plenty of tutorials on how to use branches with Git and GitHub.
* This might just be personal preference, but I like to keep commit messages short and sweet. Firstly, I would put the issue number in the description instead (you'll want to do this anyway - see my first point). We already add the PR number in brackets to the message when we merge PRs, so having the issue there too can add a bit more noise and confusion. Also, the issue number is GitHub-specific, so in my opinion it's better to use the message to describe the fix in words, and save the GitHub bit for GitHub. Secondly, I wouldn't put the file name in the message: it's already apparent from the contents of the commit, and quickly becomes unmanageable for larger PRs that touch many files. How about something like "Remove file path from VST effect name"?

plugins/VstEffect/VstEffect.cpp
```diff
@@ -70,7 +70,7 @@ VstEffect::VstEffect( Model * _parent,
 		openPlugin( m_key.attributes["file"] );
 	}
 	setDisplayName( m_key.attributes["file"].section( ".dll", 0, 0 ).isEmpty()
-		? m_key.name : m_key.attributes["file"].section( ".dll", 0, 0 ) );
+		? m_key.name : m_key.attributes["file"].section( ".dll", 0, 0 ).section("\\", -1));
```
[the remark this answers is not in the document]
eliaslk: Hey there! Sorry for the late reply, and thank you for the useful pointers - I updated the commit message and comments, and I'll definitely keep in mind creating a new branch for future contributions.

I forgot that non-Windows machines use forward slashes, but that should be a pretty easy solve! Should be as simple as adding another .section("/", -1) to line 73. Do forward slashes need to be escaped in C++?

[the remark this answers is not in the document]
PhysSong: > Do forward slashes need to be escaped in C++?

No, `"/"` works fine.

sakertooth (commented):
You *could* simplify this so it is more robust and portable using ``std::filesystem::path::stem``, but since the support for the library is still experimental (and would probably make the PR more complicated than it needs to be), I think this is fine for now.

Also, feel free to edit the function call to match with the coding conventions (i.e., remove white space between opening parentheses)

plugins/VstEffect/VstEffect.cpp
```diff
@@ -70,7 +70,7 @@ VstEffect::VstEffect( Model * _parent,
 		openPlugin( m_key.attributes["file"] );
 	}
 	setDisplayName( m_key.attributes["file"].section( ".dll", 0, 0 ).isEmpty()
-		? m_key.name : m_key.attributes["file"].section( ".dll", 0, 0 ) );
+		? m_key.name : m_key.attributes["file"].section( ".dll", 0, 0 ).section("\\", -1).section("/", -1));
```
zonkmachine, line 73: It can also end with. `.so` with native Linux vst's.

Rossmaxx:
Mind if i take over?

plugins/VstEffect/VstEffect.cpp
Rossmaxx: What changes need to be done to add .so to the list?

zonkmachine: I don't know. There are qt functions to handle filename/paths, like https://doc.qt.io/qt-6/qfileinfo.html#baseName

zonkmachine: Oh, we're not stripping a part of the actual file name but an attribute of a file. Then I don't know.

Rossmaxx: Now since you said it, looks like a hacky way more than an actual fix. Also looks like something can break somewhere.

zonkmachine: If you 'grep' the lmms code for effGetEffectName, you may find some clues. Unfortunately I just borked my vst chain so I'm a bit blind sided there right now.

plugins/VstEffect/VstEffect.cpp
```diff
@@ -71,7 +71,7 @@ VstEffect::VstEffect( Model * _parent,
 		loaded = openPlugin(m_key.attributes["file"]);
 	}
 	setDisplayName( m_key.attributes["file"].section( ".dll", 0, 0 ).isEmpty()
```
yohannd1, line 73: From the previous conversations, I'm guessing this will work:
```suggestion
	? m_key.name : QFileInfo{m_key.attributes["file"]}.baseName();
```

yohannd1 (changes requested):

plugins/VstEffect/VstEffect.cpp
yohannd1: I think I messed up the suggestion LOL
But - from what I've tested, it works on empty strings as well, so we can just do:
```cpp
	auto autoName = QFileInfo{m_key.attributes["file"]}.baseName();
	setDisplayName(autoName.isEmpty() ? m_key.name : autoName);
```

(And don't forget to `#include <QFileInfo>`)

PR #6527, opened 2022-10-03
Ghost notes - create, toggle, repeat, paste

### LINKED ISSUE #5267: Ghost notes: Ghost icon inside piano roll.
Currently to set a piano roll pattern as a ghost, you have to do it from the song editor by clicking left on that pattern and executing the option `` set as ghost in piano-roll ''. good!!.
Establecer el patron fantasma
The problem is that when you modify the same pattern several times (either to make a correction or modify a note), you have to look for it and set it as many times as necessary through that menu option. This sometimes becomes a bit repetitive.

It would be useful to have this option but within the piano roll, (since basically this option `` photographs '' the written notes). something like:
fantasma dentro de piano roll
In this way we simplify that task, and every time you make a modification to the pattern you would only have to press that button without leaving the piano roll. Regards.

### LINKED ISSUE #6233: Don't open pianoroll when setting ghost notes
Most of the time you'd set ghost from one pattern to use it in another pattern, so the workflow would be:
1. Right click pattern A.
2. Set as ghost.
3. Double click pattern B.

This is interrupted by the fact that 2. opens the piano roll in your face, so if it overlaps song editor you have to close it to proceed.

This should be a good first issue, probably just one line :)

### LINKED ISSUE #6318: Piano Roll Ghost Note icon has no visual indication of being clickable
### Bug Summary
In LMMS 1.3 alpha, the piano-roll has a button with the purpose "Remove ghost notes". However, while most buttons light up upon hover, and darken upon click, this button stays the same colour, giving no indication it is anything more than an icon.

#### Steps to reproduce
Hover and click on ghost notes icon. Only indication it is clickable is the informative pop-up. And the popup only shows after a few seconds.

#### Expected behavior
Button reacts to hover and click.

#### Actual behavior
Button doesn't react

#### Screenshot
image mouse not there
image mouse hover
image mouse click

In all cases the button has no visual change.
I'm filing this as a bug because I believe it's an unintended oversight when the button was made, or maybe it was planned to be fixed later?

#### Affected LMMS versions

LMMS 1.3 alpha

allejok96:
Fixes #6318 fixes #6233 fixes #5267
bild

- Create ghost notes from within Piano Roll
- Toggle ghost notes
- Repeat ghost infinitely in all 4 directions
- Above mentioned settings saved in project file
- Raise ghost notes from the dead, a.k.a. paste them
- Set ghost from Song Editor shows a silly notification instead of raising Piano Roll window
- ~Clicking `Use selection as ghost` when there are no notes clears the ghost notes~

TODO

- [ ] Improve draw logic

A bit outdated video demo:

https://user-images.githubusercontent.com/7693838/193662842-6a45fae6-04c2-45ef-8965-79bba674f25e.mp4

\* goat notes


src/gui/editors/PianoRoll.cpp
```diff
@@ -628,47 +636,112 @@ void PianoRoll::markSemiTone(int i, bool fromMenu)
...
+	// TODO undo ghost note changes
+
+	m_ghostNotes.clear();
+	for (auto& note: notes)
+	{
+		m_ghostNotes.push_back(new Note(note->length(), note->pos(), note->key()));
+	}
+	if (auto bounds = boundsForNotes(m_ghostNotes))
+	{
+		m_ghostBounds = bounds.value();
```
PhysSong, line 656: It's safe to use `*bounds` here since you already checked for a value using `if`.

src/gui/editors/PianoRoll.cpp
```diff
@@ -3413,43 +3487,56 @@ void PianoRoll::paintEvent(QPaintEvent * pe )
...
 
 		// -- Begin ghost MIDI clip
-		if( !m_ghostNotes.empty() )
+		if (!m_ghostNotes.empty() && m_ghostVisible.value())
 		{
-			for( const Note *note : m_ghostNotes )
+			// Lambda function that takes a value x and rounds it to the next/previous value evenly divisable by q
+			// a.k.a. quantizing... (unlike integer division this is the same for negative numbers)
+			auto roundDown = [](int x, int q) -> int { return std::floor(static_cast<float>(x) / q) * q; };
+			auto roundUp = [](int x, int q) -> int { return std::ceil(static_cast<float>(x) / q) * q; };
```
PhysSong, line 3495: Maybe good candidates for extracting to a common place?

src/core/Note.cpp
```diff
@@ -254,10 +254,13 @@ std::optional<NoteBounds> boundsForNotes(const NoteVector& notes)
 
 	for (const Note* note: notes)
 	{
+		// Treat beat notes as 1 tick long
```
PhysSong, line 257: Why 1 tick?

src/gui/editors/PianoRoll.cpp
```diff
@@ -3413,43 +3487,56 @@ void PianoRoll::paintEvent(QPaintEvent * pe )
...
+			// Horizontal repeat
+			const bool loop = m_ghostRepeated.value();
+			// Round to whole bars
+			const int ghostLength = roundUp(m_ghostBounds.end - m_ghostBounds.start, TimePos::ticksPerBar());
+			// Offset of the leftmost visible repetition, relative to the original ghost notes
+			const int posOffBegin = loop ? roundDown(m_currentPosition - m_ghostBounds.start, ghostLength) : 0;
+			// Tick to the right of the visible area
+			const int rightmostTick = m_currentPosition + (width() - m_whiteKeyWidth) * TimePos::ticksPerBar() / m_ppb;
+			// Where to stop drawing
+			const int posEnd = loop ? rightmostTick : m_ghostBounds.end.getTicks();
```
PhysSong, line 3506: Why don't you repeat everything completely, i.e., rounding up `rightmostTick` as multiples of `m_ghostBounds.end.getTicks()`?

src/gui/editors/PianoRoll.cpp
```diff
@@ -3413,43 +3487,56 @@ void PianoRoll::paintEvent(QPaintEvent * pe )
...
+
+			// Vertical repeat
+			const bool stack = m_ghostStacked.value();
+			// Round to whole octaves
+			const int ghostRange = roundUp((m_ghostBounds.highest - m_ghostBounds.lowest + 1), KeysPerOctave);
```
PhysSong, line 3511: I'm not sure how vertical repeat in this way will be useful. Can the ghost notes be transposed in some way instead of this?

src/core/Note.cpp
allejok96: Even if they are drawn as 4 or something, they can be shorter than that depending on following notes. But they can never be shorter than 1.

If you ask me note->length should return 1 for beat notes, because logically notes must be >0 but it's out of scope and I know there's the note type PR... So I just fix that problem here meanwhile.

The problem being start must be < end for a non-empty bound.

src/gui/editors/PianoRoll.cpp
allejok96: No need to draw/iterate outside visible area

src/gui/editors/PianoRoll.cpp
allejok96: Personally I'd use it in the same way you use the key/scale marker but with ghost notes you get more flexibility and can have a chord progression laid out.

Plus it's nice how you automatically have the ghost there no matter if you write bass or high.

If your ghost has a wide range I suppose it can end up on the wrong octave even if repeat is on. In that case the temporary solution would be paste ghost/move/apply as ghost. But we could add transpose buttons in another PR, it's a good idea.

luzpaz:
@allejok96 any idea on an ETA for this ?

zonkmachine:
@luzpaz Are you on Discord?

clemetayer:
Hello, I am not really a maintainer so I don't really know if I can do much, but this is a really cool feature to have so if there is any help needed, maybe I could find some time some day.

PR #6539, opened 2022-10-10
Added automatic dark note on light background for midi clips.

### LINKED ISSUE #6526: Notes is barely visible if the Track color is light-colored
maybe make the notes black or dark-gray if the track color is light-colored

aakers24:
Closes #6526


Veratil:
Thanks! Can you share a screenshot of it working for various colors both light and dark?

aakers24:
MidiClipScreenshot

Veratil:
I feel the muted clips contrast on grey isn't as good as it could be. Not sure it's a problem, but what would happen if we tried to make those notes get slightly lighter? It would have an effect on the rest, but curious how much.

aakers24:
image
How does this look? I feel like it's more visible, but less obvious that they're muted. Maybe this is a happy medium? I played around with it and much more than this makes it hard for me to tell which are muted or not.

allejok96:
This is great. Personally I don't see an issue with the contrast, but if we change it, it should be done inside `style.css` and not hardcoded.

IMHO (not necessarily related to this PR) muted clips should always turn grey, no matter what color they have. Also when selecting clips they should always turn blue. Currently custom colored clips are just darkened when selected or muted... see `ClipView::getColorForDisplay`

zonkmachine (approved):

zonkmachine:
I stressed through the review yesterday and had a second look.

The demo track Greippi - Krem Kaakkuja has colored tracks so it lends itself to comparing before/after this fix.
Here are eight screenshots side by side. As you can see from the bottom row, the unmuted notes also changes by this approach. I'd prefer the old old unmuted look and the new muted. Can we have both?

Top row ------- default theme - unmuted/muted, classic theme - unmuted/muted
Bottom row -- default theme - unmuted/muted, classic theme - unmuted/muted

guimutedoldclassicmuted

zonkmachine (changes requested):
See comment https://github.com/LMMS/lmms/pull/6539#issuecomment-1722215372

In b224f4dafa4bbd94855ef13bed00a29e98cf4a9e the colors of the unmuted tracks are now darker/less visible.

src/gui/clips/MidiClipView.cpp
```diff
@@ -523,7 +523,7 @@ void MidiClipView::paintEvent( QPaintEvent * )
 		p.scale(width(), height() - distanceToTop - 2 * notesBorder);
 
 		// set colour based on mute status
-		QColor noteFillColor = muted ? getMutedNoteFillColor().lighter(150)
+		QColor noteFillColor = muted ? getMutedNoteFillColor().lighter(200)
 									 : (c.lightness() > 127 ? getNoteFillColor().darker(400) : getNoteFillColor());
```
zonkmachine, line 527: If you change the c.lightness test to be a bit higher, even more than `(c.lightness() > 180` then the switch to a darker tone would come much later and the old themes will show like they did before unless you were going John Lennon post Yoko and all tracks were set to white.

Here's what it looks like with `c.lightness() > 200` and the colors are just around the switch.
bright
bright2

zonkmachine (changes requested):

src/gui/clips/MidiClipView.cpp
zonkmachine: PS. 200 is an example and maybe a bit much. Maybe ~170 is a good value?

aakers24: I've made a new commit addressing this issue. I felt that 175 was the value to settle at because it prioritizes the original look and feel with the lighter notes, but going beyond that results in readability issues with some clip color values (Example attached).
Note-Readability

zonkmachine: Oki doki!

zonkmachine (approved):

zonkmachine:
Looks good to go! Caveat, I haven't tried it in any other themes than the ones we ship but this looks like a nice change to me.

DomClark (approved):
The code looks fine, and works with the default themes. A more thorough implementation could use the [luminance](https://en.wikipedia.org/wiki/Relative_luminance) of the colour rather than the HSL lightness, and compare it to the luminance of the note colour rather than a fixed constant.

PR #6544, opened 2022-10-21
Fix master pitch behavior to transpose instead of detune

LostRobotMusic:
Title says it all.  This is the expected behavior for a master pitch slider (and already how it works for VSTs and ZynAddSubFX), and also makes the usage of that slider compatible with microtuning.

### DIFF
--- a/src/core/NotePlayHandle.cpp
+++ b/src/core/NotePlayHandle.cpp
@@ -527,7 +527,7 @@ void NotePlayHandle::updateFrequency()
 	if (m_instrumentTrack->m_microtuner.enabled())
 	{
 		// custom key mapping and scale: get frequency from the microtuner
-		const int transposedKey = key() + masterPitch;
+		const auto transposedKey = key() + masterPitch;
 
 		if (m_instrumentTrack->isKeyMapped(transposedKey))
 		{
--- a/src/core/NotePlayHandle.cpp
+++ b/src/core/NotePlayHandle.cpp
@@ -527,13 +527,13 @@ void NotePlayHandle::updateFrequency()
 	if (m_instrumentTrack->m_microtuner.enabled())
 	{
 		// custom key mapping and scale: get frequency from the microtuner
-		const float detuneMaster = detune + masterPitch;
+		const int transposedKey = key() + masterPitch;
 
-		if (m_instrumentTrack->isKeyMapped(key()))
+		if (m_instrumentTrack->isKeyMapped(transposedKey))
 		{
-			const auto frequency = m_instrumentTrack->m_microtuner.keyToFreq(key(), baseNote);
-			m_frequency = frequency * powf(2.f, (detuneMaster + instrumentPitch / 100) / 12.f);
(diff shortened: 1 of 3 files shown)


src/core/NotePlayHandle.cpp
```diff
@@ -527,13 +527,13 @@ void NotePlayHandle::updateFrequency()
 	if (m_instrumentTrack->m_microtuner.enabled())
 	{
 		// custom key mapping and scale: get frequency from the microtuner
-		const float detuneMaster = detune + masterPitch;
+		const int transposedKey = key() + masterPitch;
```
DomClark, line 530: `auto` would be even better here:
```suggestion
		const auto transposedKey = key() + masterPitch;
```

DomClark (approved):

allejok96 (approved):
Tested and confirmed

he29-net:
The code looks good to me, but if I may, I would also suggest renaming it to "Global transposition" or something along those lines. Calling it "Master pitch" and talking about number of semitones in the tool-tip still leaves it ambiguous, which is what led to the mismatch between expectation and implementation in the first place. Calling it transposition and talking about key offset or key shift should be clearer.

LostRobotMusic:
@he29-net The changes have been made.  I also remembered to rename it in the instrument's tuning tab.

allejok96:
(A bit late to the game but isn't semitones a valid term even in microtonal contexts?)

he29-net:
> (A bit late to the game but isn't semitones a valid term even in microtonal contexts?)

Valid term? Sure. Appropriate to describe a key offset? Not quite, IMO.

The default piano mapping happens to have a sequence of notes of increasing pitch, but the custom key mapping can be completely arbitrary. It could go high to low, it could skip notes (so neighboring keys have a whole tone between them, not semitone), or it could form a weird sequence like 1 5 2 4 3. Essentially, key shift relates to the key map and semitones relate to the scale, so in this context they should not be used interchangeably.

he29-net:
> @he29-net The changes have been made. I also remembered to rename it in the instrument's tuning tab.

@LostRobotMusic May be also nice to change the comments and variables, there is still stuff like `m_useMasterPitchModel` (defined in `src/tracks/InstrumentTrack.cpp`). Nothing critical, just to keep the code base clean and consistent for future generations. :)

I tested it and the longer string in instrument window just barely fits. It's OK in English, but if a translation results in a longer string, it will force the window to resize horizontally, which does not play nice with the fixed-size LMMS instrument UIs. (Happened to me just a few hours ago when making changes for the Microtuner UI...)

Not really a problem caused by this PR, but it would be great if you could also add `tlabel->setWordWrap(true);` for the QLabel. That should prevent the issue.

Other than that I don't see any problems.

LostRobotMusic:
(Discussed on Discord, `m_useMasterPitchModel` will keep its current name since the master pitch is still keeping its original name as well.)

he29-net (approved):

allejok96:
> Appropriate to describe a key offset? Not quite, IMO.

Right now it says "notes". I think "keys" would be more appropriate.

PR #6545, opened 2022-10-21
Make shifting notes by octaves respect microtuning

LostRobotMusic:
Currently, holding Ctrl while pressing the up or down arrow shifts all selected notes in the Piano Roll by twelve semitones.  This is incredibly inconvenient for most tuning systems because their octave size is usually greater or less than one octave... and some tunings don't even use octaves at all, instead relying on the tritave as their equivalence interval.

This PR determines the user's effective octave size (if the current track has microtuning enabled) by looking at its keymap size (or the scale size when the keymap isn't in use), and shifts the notes by that number of semitones instead.


src/gui/editors/PianoRoll.cpp
```diff
@@ -1310,16 +1312,34 @@ void PianoRoll::keyPressEvent(QKeyEvent* ke)
...
+							int keymapSize = Engine::getSong()->getKeymap(m_midiClip->instrumentTrack()->microtuner()->currentKeymap())->getSize();
+							if (keymapSize > 0)
+							{
+								shiftSemiTone(keymapSize * direction);
+							}
+							else
+							{
+								// Determine octave size from the scale if the keymap isn't in use.
+								int scaleSize = Engine::getSong()->getScale(m_midiClip->instrumentTrack()->microtuner()->currentScale())->getIntervals().size() - 1;
+								shiftSemiTone(scaleSize * direction);
```
Veratil, line 1331:
```suggestion
						auto microtuner = m_midiclip->instrumentTrack()->microtuner();
						if (microtuner->enabled())
						{
							int keymapSize = Engine::getSong()->getKeymap(microtuner->currentKeymap())->getSize();
							if (keymapSize > 0)
							{
								shiftSemiTone(keymapSize * direction);
							}
							else
							{
								// Determine octave size from the scale if the keymap isn't in use.
								int scaleSize = Engine::getSong()->getScale(microtuner->currentScale())->getIntervals().size() - 1;
								shiftSemiTone(scaleSize * direction);
```
Let's shorten some of these lines just a little bit. :)

allejok96:
Would it be useful to move this to `Microtuner::octaveSize()` for re-usability (grid lines, note stacking, etc)? And also offloading `PianoRoll.cpp` some more code :stuck_out_tongue:

DomClark:
> Move this to `Microtuner::octaveSize()`

I definitely agree with this. I was going to leave a comment on #6527 that the vertical repeat should be based on the microtuner when it's enabled, and then this PR turned up with the appropriate code.

LostRobotMusic:
> Move this to `Microtuner::octaveSize()`

Done.

src/core/Microtuner.cpp
```diff
@@ -108,6 +108,17 @@ float Microtuner::keyToFreq(int key, int userBaseNote) const
 	return middleFreq * intervals[scaleDegree].getRatio() * pow(octaveRatio, keymapOctave + scaleOctave);
 }
 
+int Microtuner::octaveSize() const
+{
+	int keymapSize = Engine::getSong()->getKeymap(currentKeymap())->getSize();
```
he29-net, line 113: This should probably be a `const int`, just as a good practice.

he29-net:
Tried testing it with the default scales and keymaps and could not find and issue. So looks good to me.

(I wish I thought of this myself, instead of brainlessly hacking my scale experiments to always be a multiple of 12...)

src/gui/editors/PianoRoll.cpp
```diff
@@ -1310,16 +1312,25 @@ void PianoRoll::keyPressEvent(QKeyEvent* ke)
...
 				if( ( ke->modifiers() & Qt::ControlModifier ) && m_action == ActionNone )
 				{
-					// shift selection up an octave
+					// shift selection by one octave
 					// if nothing selected, shift _everything_
 					if (hasValidMidiClip())
 					{
-						shiftSemiTone( 12 * direction );
+						// An octave could potentially be greater or less than twelve semitones if the microtuner is in use.
+						auto microtuner = m_midiClip->instrumentTrack()->microtuner();
```
DomClark, line 1320: If you're changing things to be const, this could be as well.

DomClark (approved):

src/gui/editors/PianoRoll.cpp
```diff
@@ -61,11 +61,13 @@
 #include "GuiApplication.h"
 #include "gui_templates.h"
 #include "InstrumentTrack.h"
+#include "Keymap.h"
 #include "MainWindow.h"
 #include "MidiClip.h"
 #include "PatternStore.h"
 #include "PianoView.h"
 #include "PositionLine.h"
+#include "Scale.h"
```
DomClark, line 70: I don't think these extra headers need to be included any more.

he29-net (approved):
(Trying to find the "approve button", I suppose this is it?)

Issue #6487, opened 2022-08-16
Division with 0 in BandLimitedWave.h


zonkmachine:
Division with 0 in BandLimitedWave.h
https://github.com/LMMS/lmms/blob/9faa1d90c4b2eac2f19ca14f78354405ca9a04dc/include/BandLimitedWave.h#L120-L123

It seem to be used only in LB302 and Monstro which will crash in fpe enabled debugging.
So, std::max() with some suitable small number.

The function above, freqToLen() is also potentially dividing with 0 but doesn't seem to be used anywhere.
https://github.com/LMMS/lmms/blob/9faa1d90c4b2eac2f19ca14f78354405ca9a04dc/include/BandLimitedWave.h#L114-L117

luzpaz:
soft bump

zonkmachine:
> soft bump

Please stop bumping issues/PR's with no new information.

Issue #6173, opened 2021-09-30
Refactoring: Change `enum` types to `enum class` whereever possible.


irrenhaus3:
In the spirit of the C++ core guideline `Enum.3` (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Renum-class), all plain `enum`s that can be replaced with `enum class` should be thusly replaced. This avoids potiential name conflicts between enum values and other symbols in a scope where the enum values would be implicitly visible.  
  
An example in the codebase is the enum `Track::TrackTypes`, which contains values such as `BBTrack`, `SampleTrack`, etc. - those names are identical to the class names they represent, so when creating i.e. a `BBTrack` object somewhere in a subclass of `Track`, we end up with one of the two options:
`::BBTrack* bb_track = /* get a BB track */`
or
`class BBTrack* bb_track = /* get a BB track */`.  
  
By scoping the enum, `BBTrack` would always mean what you'd expect it to mean, and only `TrackTypes::BBTrack` would refer to the enum value.

Issue #6004, opened 2021-05-03
Beat And Bassline editor - scroll bar overlaps


JGHFunRun:
I was on the alpha, no clue if it occurs on stable, and noticed that when BnB's scroll bar shows up it overlaps the last note

![bruh-fix-this](https://user-images.githubusercontent.com/52897905/116921040-0c923380-ac19-11eb-8652-ef422b79e207.PNG)

zonkmachine:
Confirmed. This is not in lmms-1.2.2 which instead has a gap after the last beat step reserved for the scroll bar.

messmerd:
This appears to be fixed in master, though I don't know when it was fixed.

PR #6547, opened 2022-10-22
Some housekeeping/small bug fixes

ifndefJOSH:
I'd like to get more into LMMS development by starting with fixing smaller bugs and things like that as I get familiar with the codebase. This PR will contain some of my bug fixes and will be marked as a draft. Currently, there are changes here to address:

- Issue #6173 : in plugins, `enum` is changed to `enum class`
- Issue #6487  : creates a division floor to prevent division by 0
- (Solution not finished yet) Issue #6004 : Currently creates a 16px right margin in the scrollbox. Would rather have the margin be dynamic when the scrollbar appears and I know it's possible in Qt to get scrollbar width. Don't use this solution yet.
    - Also in Qt5 `setMargins()` is depricated, so I used `setContentMargins()`

### DIFF
--- a/data/themes/default/style.css
+++ b/data/themes/default/style.css
@@ -3,7 +3,7 @@
  ********************/
 
 /* most foreground text items */
-QLabel, QTreeWidget, QListWidget, QGroupBox, QMenuBar {
+QLabel, QTreeWidget, QListWidget, QGroupBox, QMenuBar, QCheckBox {
 	color: #d1d8e4;
 }
 
@@ -462,6 +462,10 @@ lmms--gui--AutomatableSlider::handle:horizontal {
 
 lmms--gui--EffectSelectDialog QScrollArea {
 	background: #262b30;
+}
+
+lmms--gui--SetupDialog QScrollArea {
+	border: 0px;
 }
 
 /* the inner boxes in LADSPA effect windows */
--- a/include/AudioDeviceSetupWidget.h
+++ b/include/AudioDeviceSetupWidget.h
@@ -2,6 +2,7 @@
  * AudioDeviceSetupWidget.h - Base class for audio device setup widgets
  *
  * Copyright (c) 2004-2015 Tobias Doerffel <tobydox/at/users.sourceforge.net>
+ * Copyright (c) 2023- Michael Gregorius
  *
  * This file is part of LMMS - https://lmms.io
  *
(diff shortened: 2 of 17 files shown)


zonkmachine:
As for caps, we don't really touch the code in **/plugins/LadspaEffect/** as it's 'upstream'. Caps will eventually be switched to a submodule (https://github.com/LMMS/lmms/pull/4027) so any changes there will be undone at that point.

src/gui/editors/TrackContainerView.cpp
```diff
@@ -89,8 +89,8 @@ TrackContainerView::TrackContainerView( TrackContainer * _tc ) :
 	m_tc->setHook( this );
 	//keeps the direction of the widget, undepended on the locale
 	setLayoutDirection( Qt::LeftToRight );
-	auto layout = new QVBoxLayout(this);
-	layout->setMargin( 0 );
+	QVBoxLayout * layout = new QVBoxLayout( this );
+	layout->setContentsMargins( 0, 0, RIGHT_SCROLLBOX_MARGIN, 0 );
```
Veratil, line 93:
```suggestion
	auto layout = new QVBoxLayout(this);
	layout->setContentsMargins(0, 0, RIGHT_SCROLLBOX_MARGIN, 0);
```
This doesn't need to change off `auto`, and adding whitespace is no-no. See [Coding guidelines](https://github.com/LMMS/lmms/wiki/Coding-Conventions#whitespace).

include/BandLimitedWave.h
```diff
@@ -101,6 +101,7 @@ class LMMS_EXPORT BandLimitedWave
 	BandLimitedWave() = default;
 	virtual ~BandLimitedWave() = default;
 
+	static const float DIVISION_FLOOR = 1e-4; // This can be changed
```
Veratil, line 104: We'll need to figure out a good small number here. I'm not sure if 0.0001 is suitable.

Veratil (changes requested):

src/gui/editors/TrackContainerView.cpp
ifndefJOSH: Sorry, the `auto` change must have been when I merged my commit with some upstream commits I hadn't pulled down

include/BandLimitedWave.h
ifndefJOSH: If we're dividing by a frequency, we could make it as high as 1

allejok96:
By changing to `enum class` you must make sure to write the full type everywhere the enum is used.

As you can see the build fails now. For example in `Monstro.h` you must change this:

```diff
- return BandLimitedWave::oscillate( _ph, _wavelen, BandLimitedWave::BLTriangle );
+ return BandLimitedWave::oscillate(_ph, _wavelen, BandLimitedWave::Waveform::BLTriangle);
```

There are some cases where enum class becomes really important (like when there is a class with the same name as in the enum), and it is generally a good idea to not have a ton of names in the same scope to avoid collision.

But on the other hand enum class makes code longer and switching to it introduces changes everywhere so unless there's a good reason for a particular enum to change, or we agree to change *all* enums in one go, I don't think we need to touch those right now. (That's just a personal opinion)

luzpaz:
bump @ifndefJOSH

DomClark:
The enum class changes I have now done in #6760.

PR #6550, opened 2022-10-26
Fix automation track unmute bugs

### LINKED ISSUE #6327: Import of project from old version with solo'ed track- automation tracks don't unmute
### Bug Summary
When opening a project made in an older version of LMMS, wherein an instrument track has been solo'ed before saving, 1.3 alpha upon un-soloing the soloed tracks, automation tracks do not unmute. (Related, PR 5547, where the feature was added that enabled this behaviour.)
ie,
Soloing instrument track and save in 1.2, then re-opening in 1.3 alpha and un-soloing, keeps automation tracks muted.

#### Steps to reproduce
1. Create project in 1.2, make sure instrument track and automation track present in project.
2. Solo an instrument track and save the project file.
3. (Reopening the project file in 1.2 and un-soloing the instrument, the automation track is unmuted and becomes active.)
4. Opening the project file in 1.3 alpha, however, keeps automation tracks muted.

#### Expected behavior
Automation tracks should return to original state of unmuted upon un-solo'ing.

#### Actual behavior
Automation tracks remain muted.

Presumably this can be solved with version-checking of a project file, and the way unmuting a non-solo track works when a specific track is solo'ed, or how automation tracks can be re-muted after soloing an instrument track, but upon un-solo'ing they return to their original state.

#### Screenshot
image
image
image
image

File:
[SoloTrack.zip](https://github.com/LMMS/lmms/files/8213324/SoloTrack.zip)

#### Affected LMMS versions
LMMS 1.3 alpha (by nature of bug)

jphinspace:
This is my first contribution to LMMS! Let me know if you have any feedback.

This change is to fix https://github.com/LMMS/lmms/issues/6327 - 1.2 projects with soloed tracks were not unmuting automation tracks properly when loaded into 1.3 and unsoloed.

While fixing this, I found a similar issue where you could mute an automation track, solo it, unsolo it, and the track would not revert to being muted. This change should fix that as well.

The issue seems to be the block associated with this PR https://github.com/LMMS/lmms/pull/5547#issuecomment-650464639
I don't think this check is necessary though, since we record the previous muted state of all tracks anyway whenever one is soloed. Soloing a different track will still leave automation tracks unmuted, and unsoloing a track will revert automation track mute statuses to whatever they were before. To me this seemed more intuitive.

Also I've changed a few var names (thought these might be more descriptive, open to reverting if anyone feels strongly though)

### DIFF
--- a/src/core/Track.cpp
+++ b/src/core/Track.cpp
@@ -582,37 +582,39 @@ void Track::toggleSolo()
 {
 	const TrackContainer::TrackList & tl = m_trackContainer->tracks();
 
-	bool soloBefore = false;
+	bool otherTrackAlreadySolo = false;
 	for (const auto& track : tl)
 	{
 		if (track != this)
 		{
 			if (track->m_soloModel.value())
 			{
-				soloBefore = true;
+				otherTrackAlreadySolo = true;
 				break;
 			}
 		}
 	}
 
-	const bool solo = m_soloModel.value();
+	const bool soloEnabled = m_soloModel.value();
 	// Should we use the new behavior of solo or the older/legacy one?
+	// Legacy: soloing a track mutes all other tracks including automation tracks
+	// New: soloing a track leaves other automation tracks in pre-solo state
 	const bool soloLegacyBehavior = ConfigManager::inst()->value("app", "sololegacybehavior", "0").toInt();
 
 	for (const auto& track : tl)
 	{
-		if (solo)
+		if (soloEnabled)
 		{
-			// save mute-state in case no track was solo before
-			if (!soloBefore)
+			// Save mute-states in case no track was solo before
+			if (!otherTrackAlreadySolo)
 			{
 				track->m_mutedBeforeSolo = track->isMuted();
 			}
-			// Don't mute AutomationTracks (keep their original state) unless we are on the sololegacybehavior mode
 			if (track == this)
 			{
 				track->setMuted(false);
 			}
+			// Don't mute AutomationTracks (keep their original state) unless we are on the soloLegacyBehavior mode
 			else if (soloLegacyBehavior || track->type() != AutomationTrack)
 			{
 				track->setMuted(true);
@@ -622,14 +624,10 @@ void Track::toggleSolo()
 				track->m_soloModel.setValue(false);
 			}
 		}
-		else if (!soloBefore)
+		// Revert all tracks to pre-solo values
+		else if (!otherTrackAlreadySolo)
 		{
-			// Unless we are on the sololegacybehavior mode, only restores the
-			// mute state if the track isn't an Automation Track
-			if (soloLegacyBehavior || track->type() != AutomationTrack)
-			{
-				track->setMuted(track->m_mutedBeforeSolo);
-			}
+			track->setMuted(track->m_mutedBeforeSolo);
 		}
 	}
 }


allejok96:
I haven't studied the code closely but I somehow think this belongs in `DataFile.cpp` in one of the upgrade methods. Like if `mutedBeforeSolo` is unset, give it a default value. Second option would be in `Track::loadSettings. That will make it more clear that this is just updating a legacy behavior, and not something we usually care about in our code.

allejok96:
Ignore what I said. The legacy behavior setting means the user can switch between behaviors so there is no way to make an upgrade method. And loadSettings is already the way I suggested, so this is probably the only place to fix this issue.

src/core/Track.cpp
```diff
@@ -622,14 +624,10 @@ void Track::toggleSolo()
 				track->m_soloModel.setValue(false);
 			}
 		}
-		else if (!soloBefore)
+		// Revert all tracks to pre-solo values
+		else if (!otherTrackAlreadySolo)
```
allejok96, line 628: Thinking out loud:

`soloEnabled` is true if *this* track is solo . `otherTrackAlreadySolo` is true if *any* track is solo. At first glance it seems like this function is only called when the solo LED is clicked. So you may think that if you uncheck the solo LED of a track, then `otherTrackAlreadySolo` must be false. *But* this function also calls itself through this line:

    track->m_soloModel.setValue(false);

And in that case a solo LED is unchecked while another solo LED is active. A bit confusing and ineffective but probably the simplest solution.

TL;DR there is nothing wrong here.

src/core/Track.cpp
```diff
@@ -622,14 +624,10 @@ void Track::toggleSolo()
...
 		{
-			// Unless we are on the sololegacybehavior mode, only restores the
-			// mute state if the track isn't an Automation Track
-			if (soloLegacyBehavior || track->type() != AutomationTrack)
-			{
-				track->setMuted(track->m_mutedBeforeSolo);
-			}
+			track->setMuted(track->m_mutedBeforeSolo);
```
allejok96, line 630: More thinking out loud:

If `soloLegacyBehavior` is false, then the automation track is not muted, and "restoring" it will just unmute it (no change).  So the if statement is not needed. This is correct.

(And for 1.2 projects `m_mutedBeforeSolo` defaults to false)

allejok96 (approved):
Looks good

DomClark:
Thanks for your contribution! It's really appreciated, so don't let the rest of this comment discourage you.

I don't think I agree with what behaviour would be most intuitive for the new solo mode. In my opinion, "soloing leaves automation tracks alone, but unsoloing restores them" is rather unintuitive, and "soloing/unsoloing ignores automation tracks" would be better (hence my original comment on the PR that introduced this bug).

The problem isn't unique to the 1.2->1.3 upgrade: the same behaviour occurs if a track is muted under the legacy mode, but unmuted under the new mode. It's just that upgrading effectively switches your mode from legacy to new. I think that the real bug (or source of unintuitive behaviour, at least) is that an unsolo operation uses the current solo mode, rather than the mode that was in effect during the solo operation.

As an alternative fix, I would consider making `m_mutedBeforeSolo` a `std::optional<bool>`. Upon unsoloing, if it has a value, then that value is restored, otherwise the muted state of the track is left alone. I'll let you fill in the details on how to set the value.

allejok96:
I'd imagine we need an upgrade method to adress the original bug if we choose to use `std::optional`.

@Dom your suggestion fixes this problem:

1. Solo
2. Mute an automation track manually
3. Un-solo
4. Automation track is un-muted

Mute changes during solo isn't preserved for any other tracks so I wouldn't raise an eyebrow if it isn't preserved for automation tracks either (even with the new solo mode on). So personally I'd go as-is, just to keep it simple if nothing else.

DomClark:
> I'd imagine we need an upgrade method to adress the original bug if we choose to use `std::optional`.

I think we'd be fine without - the way we did this last time is that we omit the attribute if the optional is empty. So all existing project files are still valid: they correspond to an optional that contains a value (and thus you get the desired behaviour).

> Mute changes during solo isn't preserved for any other tracks so I wouldn't raise an eyebrow if it isn't preserved for automation tracks either

That's true, but you could make a similar argument about consistency in favour of never implementing #5547 in the first place. Also, with #5547 in place (which I think was a useful change), automation track mute changes on unsolo can either be consistent with how other tracks behave on unsolo, or with how automation tracks behave on solo, but not both. So a consistency argument alone doesn't really clear anything up. We need to decide which is better, and I think we should aim for the most useful, intuitive workflow we can. Perhaps it's worth asking some users on Discord?

That said, I'm happy with the code in this PR if we do choose to go with the current approach.

allejok96:
Ah sorry @Dom I was thinking of the wrong thing. I mean we need an upgrade method if we load a project older than #5565 that doesn't save the mute states prior to solo. In that case it would default to null which would mean no change so #6327 would persist.

As for the logic your right. I'm just being lazy.

PR #6540, opened 2022-10-11
Add Phaser effect to LMMS

LostRobotMusic:
Replaces #5168, companion PR to #5873.  I already listed most of the features in the initial post, so I won't bother here.

I've fixed the bitrot issues that built up over time, and also added a few internal improvements (most notably 2x oversampling to every source of distortion).  I have no idea how to build a "good" oversampling filter, but for an effect of this nature I don't think absolute perfection is required in this aspect.  Just as long as most of the aliasing is removed.  I just used two lowpasses each for the input and output, cutoff at exactly Nyquist of the original sample rate.

A screenshot of the UI (thanks to H4CKTON3 for the assistance)
image

This is a very old video so it's out of date in some aspects, but it provides an easy-access showing of some of its features:
https://www.youtube.com/watch?v=_Zyd7OcZRO8

PR #6553, opened 2022-10-30
Add HIIR library to LMMS

LostRobotMusic:
A crucial library for high-performance oversampling within LMMS, licensed under WTFPL.

There are many situations within LMMS that require oversampling.  The quality of oversampling is heavily determined by the filters used in the process... the ideal is to have as steep of a filter slope as possible (to remove all aliasing without the filtering being audible at all) while also introducing minimal phase shifting (or else the filtering will audibly smear the audio).

Generic Butterworth lowpass filters are not adequate for this goal.  In order for the desired steep slope to be achieved, it would be necessary to stack a _very_ large number of these filters on top of each other, which results in tons of audible phase shifting and quite a hefty CPU demand.

An idealized FIR filter also won't work in many situations... the number of taps required to make the filter usable will result in large amounts of latency and quite a massive CPU demand once again.

HIIR solves this by using a different type of IIR filter called an elliptic filter, which is characterized by having an extremely steep slope with minimal filter taps.
image

Butterworth filters have such a gentle slope because they are designed to have minimal passband and stopband rippling.  Elliptic filters allow a small amount of rippling and are able to have a far steeper slope as a result.  Conveniently, this ripple is mostly only audible near the cutoff frequency, which is not an issue when you're filtering near Nyquist at frequencies too high for a human to hear.

The filters within HIIR are designed specifically with oversampling in mind, and will be the best things to use in almost all cases where oversampling is required.  It also comes with some other useful filters which work in similar ways, like an IIR Hilbert Transform which shifts the audio by 90 degrees (an essential ingredient for frequency shifting).

I'll be modifying #6540 to make use of this library once it is added.


sakertooth:
Why not add this library as a Git Submodule?

sakertooth (approved):

JohannesLorenz (approved):
A bit weird to add code that is not used, but I can't see errors in the code.

Veratil:
> A crucial library for high-performance oversampling within LMMS, licensed under WTFPL.

Just so we can see what the license is here without having to look it up. 👍 http://www.wtfpl.net/
```
        DO WHAT THE FUCK YOU WANT TO PUBLIC LICENSE 
                    Version 2, December 2004 

 Copyright (C) 2004 Sam Hocevar <sam@hocevar.net> 

 Everyone is permitted to copy and distribute verbatim or modified 
 copies of this license document, and changing it is allowed as long 
 as the name is changed. 

            DO WHAT THE FUCK YOU WANT TO PUBLIC LICENSE 
   TERMS AND CONDITIONS FOR COPYING, DISTRIBUTION AND MODIFICATION 

  0. You just DO WHAT THE FUCK YOU WANT TO.
```

src/CMakeLists.txt
```diff
@@ -188,6 +188,7 @@ SET(LMMS_REQUIRED_LIBS ${LMMS_REQUIRED_LIBS}
 	${SNDFILE_LIBRARIES}
 	${FFTW3F_LIBRARIES}
 	${EXTRA_LIBRARIES}
+	hiir
```
DomClark, line 191: I'd rather this be added as a dependency when and where it's needed, rather than having the core depend on it right now even though there's no code using it.

messmerd (approved):

DomClark (approved):

PR #6554, opened 2022-10-30
AM pitchshifter compatibility

### LINKED ISSUE #6312: AM Pitchshifter compatibility break
A user on Discord reports:

> I've compiled LMMS (git latest) and all works perfect, except one thing. When I open a 1.2.2 project which uses "AM pitchshifter", the value of pitch is always maxed out - newest version does not load proper value from the project file. I assume is that because of fact that new "AM pitchshifter" has more gauges than the old one

To reproduce: Create a project in 1.2.2 using the AM pitchshifter (with some amount of pitch shift set), save it, and load in a master build. The pitch shift will not load correctly.

I suspect this broke in https://github.com/LMMS/lmms/commit/fd36713314c23f43a5ce8e31be9e702e0683c485, but haven't bisected to confirm.

allejok96:
Extend the 1.3 upgrade routine to fix #6312 and shorten some of the code in there

### DIFF
--- a/src/core/DataFile.cpp
+++ b/src/core/DataFile.cpp
@@ -1322,44 +1322,32 @@ void DataFile::upgrade_1_3_0()
 					// Effect name changes
 
 					QDomElement attribute = attributes.item( k ).toElement();
-					if( attribute.attribute( "name" ) == "file" &&
-							( attribute.attribute( "value" ) == "calf" ||
-							attribute.attribute( "value" ) == "calf.so" ) )
+					const QString attrName = attribute.attribute("name");
+					const QString attrVal = attribute.attribute("value");
+					const QString plugin = attrName == "plugin" ? attrVal : "";
+
+					static const std::map<QString, QString> pluginNames = {
+						{"Sidechaincompressor", "SidechainCompressor"},
+						{"Sidechaingate", "SidechainGate"},
+						{"Multibandcompressor", "MultibandCompressor"},
+						{"Multibandgate", "MultibandGate"},
+						{"Multibandlimiter", "MultibandLimiter"},
+					};
+
+					if (attrName == "file" && (attrVal == "calf" || attrVal == "calf.so" ))
 					{
 						attribute.setAttribute( "value", "veal" );
 					}
-					else if( attribute.attribute( "name" ) == "plugin" &&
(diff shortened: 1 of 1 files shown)


src/core/DataFile.cpp
```diff
@@ -1314,44 +1314,34 @@ void DataFile::upgrade_1_3_0()
...
+					const QString attrVal = attribute.attribute("value");
+					const QString plugin = attrName == "plugin" ? attrVal : "";
+
+					const std::vector<std::pair<QString, QString>> pluginNames = {
+						{"Sidechaincompressor", "SidechainCompressor"},
+						{"Sidechaingate", "SidechainGate"},
+						{"Multibandcompressor", "MultibandCompressor"},
+						{"Multibandgate", "MultibandGate"},
+						{"Multibandlimiter", "MultibandLimiter"},
+					};
```
PhysSong, line 1327: How about using map-like data structure, if possible?

PhysSong (commented):
Generally looks good, one suggestion.

allejok96:
Huh why does it compile fine for me, but not here? Is it something I've forgot to `#include`?

PhysSong:
> Huh why does it compile fine for me, but not here? Is it something I've forgot to `#include`?

Likely, you can try adding `#include <unordered_map>` to the source file.

allejok96:
I hate c++

DomClark:
`std::unordered_map` by default requires that there be a specialisation of `std::hash` for the key type. Qt doesn't offer this until 5.14 (see https://github.com/qt/qtbase/commit/4469e36d7203a55a4e158a50f0e9effc3f2fa3c2). You have a couple of options:
* Implement the required hash object. I wouldn't specialise `std::hash`, as that may conflict with Qt's implementation on version that do supported; instead define a new struct and pass that in as the third template argument.
* Use `std::map` instead. This only requires its keys to be comparable.
* Use `QHash` instead. This works with Qt's own hash function.

Grissess:
It's perhaps worth noting that [Qt has its own `QMap`](https://doc.qt.io/qt-5/qmap.html); I'm not sure if it's better, worse, recommendable or not, but it probably plays much nicer with Qt's own core classes. Probably.

PhysSong (approved):

DomClark (approved):

PR #6564, opened 2022-11-08
clang-tidy: Apply modernize-avoid-c-arrays everywhere

sakertooth:
This clang-tidy warns against using C-arrays and suggests using either an ``std::array`` or ``std::vector``. Since the fix-its aren't supported for this check, all of these changes were done manually. Some C-style arrays were skipped as they made the code more verbose with no real benefit.


sakertooth:
Might revert the changes made in ``DrumSynth.cpp``, unless anyone objects. The changes were quite verbose, IMO.

plugins/Bitcrush/Bitcrush.cpp
```diff
@@ -35,8 +35,8 @@ const int OS_RATE = 5;
 const float OS_RATIO = 1.0f / OS_RATE;
 const float CUTOFF_RATIO = 0.353553391f;
 const int SILENCEFRAMES = 10;
-const float OS_RESAMPLE [5] = { 0.0001490062883964112, 0.1645978376763992, 0.6705063120704088,
-		0.1645978376763992, 0.0001490062883964112 };
+const auto OS_RESAMPLE = std::array{0.0001490062883964112f, 0.1645978376763992f, 0.6705063120704088f,
```
messmerd, line 38: Explicitly specifying this as a std::array of floats would probably be better than inferring that from integer suffixes.

plugins/VstEffect/VstEffect.cpp
```diff
@@ -86,15 +86,11 @@ bool VstEffect::processAudioBuffer( sampleFrame * _buf, const fpp_t _frames )
...
 	{
 		const float d = dryLevel();
-#ifdef __GNUC__
-		sampleFrame buf[_frames];
-#else
-		sampleFrame * buf = new sampleFrame[_frames];
-#endif
-		memcpy( buf, _buf, sizeof( sampleFrame ) * _frames );
+		auto buf = std::vector<sampleFrame>{static_cast<size_t>(_frames)};
+		memcpy(buf.data(), _buf, sizeof(sampleFrame) * _frames);
```
DomClark, line 90: Realtime safety again. Original code is also not good when `__GNUC__` is not defined, and should be changed to use `_alloca` instead, but is probably out of scope.

plugins/Compressor/Compressor.cpp
```diff
@@ -292,7 +292,7 @@ bool CompressorEffect::processAudioBuffer(sampleFrame* buf, const fpp_t frames)
 
 	for(fpp_t f = 0; f < frames; ++f)
 	{
-		sample_t drySignal[2] = {buf[f][0], buf[f][1]};
+		auto drySignal = std::array{buf[f][0], buf[f][1]};
 		sample_t s[2] = {drySignal[0] * m_inGainVal, drySignal[1] * m_inGainVal};
```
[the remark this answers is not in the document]
sakertooth: clang-tidy doesn't make any transformations with the modernize-avoid-c-arrays check (It's apparently dangerous to make the transformations in header files, not sure why exactly). The changes were done by me, so I probably just missed it while going through the list of C-styled arrays clang-tidy reported back. I'll try to rerun clang-tidy again to see if I missed anymore of them.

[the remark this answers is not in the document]
DomClark: Makes sense. 👍

src/gui/modals/ExportProjectDialog.cpp
```diff
@@ -159,8 +159,8 @@ void ExportProjectDialog::startExport()
 					static_cast<AudioEngine::qualitySettings::Interpolation>(interpolationCB->currentIndex()),
 					static_cast<AudioEngine::qualitySettings::Oversampling>(oversamplingCB->currentIndex()) );
 
-	const int samplerates[5] = { 44100, 48000, 88200, 96000, 192000 };
-	const bitrate_t bitrates[6] = { 64, 128, 160, 192, 256, 320 };
+	const auto samplerates = std::array{44100, 48000, 88200, 96000, 192000};
+	const auto bitrates = std::array{64, 128, 160, 192, 256, 320};
```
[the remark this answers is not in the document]
DomClark: I disagree with this. Leaving the template arguments to be automatically deduced is similar to using `auto` to declare variables, which we already try to do and has numerous benefits. The Core Guidelines [recommend automatically deducing template arguments](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#t44-use-function-templates-to-deduce-class-template-argument-types-where-feasible) (the guideline is a bit out of date, as it's written for C++14, but the concept is the same).

I don't think anyone would have complained if `bitrates` were an `int[]` originally (indeed, `samplerates` could use `sample_rate_t`). If it's that important to distinguish the types, then some sort of strict typedef (such as `enum class bitrate_t : std::uint16_t { };`) should be used, at which point specifying the element type alone won't work anyway.

In fact, there are a number of places in this PR where what used to be a `float[]` (or even a `double[]`) is now a `std::array<sample_t, N>` thanks to automatically deduced types, which is more appropriate for the situation and can avoid a runtime conversion.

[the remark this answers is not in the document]
sakertooth: I honestly never really considered the pros and cons of either approach. Dom brought to my attention that I can simplify the construction of these types using template deduction initially, which made the code simpler. There are other places in the code where I mixed the two approaches together, so I'm planning on just deducing wherever possible. I also try to follow the guidelines whenever possible. Going against it is probably not a good idea in the long run. It also helps to deduce in this case because the size doesn't have to be specified every time  

I still don't see this approach as "bad", per-se. The guidelines said that "Writing the template argument types explicitly can be tedious and unnecessarily verbose". I don't think specifying the type and the size made the code more verbose to the point where deducing is a hard recommendation.

src/gui/widgets/LedCheckBox.cpp
```diff
@@ -35,7 +35,7 @@ namespace lmms::gui
 {
 
 
-static const QString names[LedCheckBox::NumColors] =
+static const auto names = std::array<QString, 3>
 {
 	"led_yellow", "led_green", "led_red"
 } ;
```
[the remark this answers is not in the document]
sakertooth: Didn't know we could define our own literals. Thanks for bringing this feature to my attention :+1:

src/core/DataFile.cpp
```diff
@@ -89,18 +89,17 @@ const std::vector<ProjectVersion> DataFile::UPGRADE_VERSIONS = {
 	"1.2.0-rc3"        ,   "1.3.0"
 };
 
-DataFile::typeDescStruct
-		DataFile::s_types[DataFile::TypeCount] =
+std::array<DataFile::typeDescStruct, DataFile::TypeCount> DataFile::s_types =
```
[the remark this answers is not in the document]
sakertooth: Can you elaborate on this? Omitting the template arguments here gives a deduction error. The same occurs for ``InstrumentFunctions`` as well.
``no viable constructor or deduction guide for deduction of template arguments of array``

I'm not sure how I would go about deducing these types in this instance (or if it is necessary).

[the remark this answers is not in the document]
DomClark: It should work if you construct each array element with `DataFile::typeDescStruct`, rather than having raw initialiser lists (IIRC this was how you had it when I wrote that comment). I don't mind whether you put the type in the template parameters or the elements of the initialiser list; usually I prefer to have it on the elements, as then you can also infer the array size, but it does make it somewhat more verbose in this case.

You also need to remove the declarations from the header files when using `auto` to avoid getting redefinition errors. The variables in question are only used in their own source files, so there's no need to forward-declare them in headers (which forces you to specify the type explicitly, as there's no definition to deduce it from). They can be static variables in the source file only (even better, [define them in an anonymous namespace](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf22-use-an-unnamed-anonymous-namespace-for-all-internalnon-exported-entities)).

include/InstrumentFunctions.h
```diff
@@ -129,7 +129,7 @@ class InstrumentFunctionNoteStacking : public Model, public JournallingObject
 			ChordSemiTones m_semiTones;
 		};
 
-		static Init s_initTable[];
+		static std::vector<Init> s_initTable;
```
DomClark, line 132: I think this could just be a `std::array`, as it doesn't appear to be dynamically sized.

src/core/audio/AudioFileFlac.cpp
```diff
@@ -107,11 +107,11 @@ void AudioFileFlac::writeBuffer(surroundSampleFrame const* _ab, fpp_t const fram
 				buf[frame*channels() + channel] = qMax( clipvalue, _ab[frame][channel] * master_gain );
 			}
 		}
-		sf_writef_float(m_sf,static_cast<float*>(buf.get()),frames);
+		sf_writef_float(m_sf, static_cast<float*>(buf.data()), frames);
 	}
 	else // integer PCM encoding
 	{
-		std::unique_ptr<int_sample_t[]> buf{ new int_sample_t[frames*channels()] };
+		auto buf = std::make_unique<int_sample_t[]>(frames * channels());
```
DomClark, line 114: This could be a `std::vector`, like the earlier change in this file.

src/core/ProjectRenderer.cpp
```diff
@@ -234,12 +234,12 @@ void ProjectRenderer::updateConsoleProgress()
 	prog[cols] = 0;
 
 	const auto activity = (const char*)"|/-\\";
-	memset( buf, 0, sizeof( buf ) );
-	sprintf( buf, "\r|%s|    %3d%%   %c  ", prog, m_progress,
+	memset(buf.data(), 0, sizeof(buf));
```
DomClark, line 237: I think `.size()` would be better than `sizeof` here (replacing `memcpy` with `std::fill` would be even better, but you might consider that to be out of scope).

src/core/lv2/Lv2Proc.cpp
```diff
@@ -247,11 +247,11 @@ void Lv2Proc::copyModelsFromCore()
 				ev.time.frames(Engine::framesPerTick()) + ev.offset;
 			uint32_t type = Engine::getLv2Manager()->
 				uridCache()[Lv2UridCache::Id::midi_MidiEvent];
-			uint8_t buf[4];
-			std::size_t bufsize = writeToByteSeq(ev.ev, buf, sizeof(buf));
+			auto buf = std::array<uint8_t, 4>{};
+			std::size_t bufsize = writeToByteSeq(ev.ev, buf.data(), sizeof(buf));
```
DomClark, line 251: Again, I would prefer `.size()` over `sizeof`.

src/core/DataFile.cpp
```diff
@@ -89,18 +89,27 @@ const std::vector<ProjectVersion> DataFile::UPGRADE_VERSIONS = {
...
-	DataFile::typeDescStruct{ DataFile::SongProject, "song" },
-	DataFile::typeDescStruct{ DataFile::SongProjectTemplate, "songtemplate" },
-	DataFile::typeDescStruct{ DataFile::InstrumentTrackSettings, "instrumenttracksettings" },
-	DataFile::typeDescStruct{ DataFile::DragNDropData, "dnddata" },
-	DataFile::typeDescStruct{ DataFile::ClipboardData, "clipboard-data" },
-	DataFile::typeDescStruct{ DataFile::JournalData, "journaldata" },
-	DataFile::typeDescStruct{ DataFile::EffectSettings, "effectsettings" },
-	DataFile::typeDescStruct{ DataFile::MidiClip, "midiclip" }
-} ;
+	struct LMMS_EXPORT typeDescStruct
```
DomClark, line 94: I realise this is how it was before, so feel free to consider this out of scope, but `LMMS_EXPORT` can be removed since this struct is only used in this file. (Also, our coding conventions require types to start with an uppercase letter, and including "struct" in the name of a struct feels weird to me.)

src/core/DataFile.cpp
```diff
@@ -89,18 +89,27 @@ const std::vector<ProjectVersion> DataFile::UPGRADE_VERSIONS = {
...
+	{
+		DataFile::Type m_type;
+		QString m_name;
+	};
+
+	auto s_types = std::array<typeDescStruct, DataFile::TypeCount>
```
DomClark, line 100: Again, an improvement beyond what this PR was intended for, but this array can be const.

DomClark (approved):

plugins/BassBooster/BassBooster.cpp
```diff
@@ -102,7 +102,7 @@ bool BassBoosterEffect::processAudioBuffer( sampleFrame* buf, const fpp_t frames
 		m_bbFX.leftFX().setGain( gain );
 		m_bbFX.rightFX().setGain( gain);
 
-		sample_t s[2] = { buf[f][0], buf[f][1] };
+		auto s = std::array {buf[f][0], buf[f][1]};
```
PhysSong, line 105: Inconsistent formatting(not sure it's enforced or not).
```suggestion
		auto s = std::array{buf[f][0], buf[f][1]};
```

PhysSong (approved):

PR #6565, opened 2022-11-10
Show pull request number in version string

DomClark:
Could be useful, especially once PR builds are easier to download again. The commit hash has been really useful, but it's not always trivial to get from a hash to a PR (or determine that it's not part of one at all). As an example, builds from this PR have the version string `1.3.0-alpha.1.237+pr6565.gd3c0c8e84`.


Veratil:
Note: the bump to CMake 3.9 is for the [`CMAKE_MATCH_<N>` variable for regex results](https://cmake.org/cmake/help/latest/variable/CMAKE_MATCH_n.html).

PhysSong (approved):

sakertooth (approved):
Working as expected.

PR #6566, opened 2022-11-11
Only rebuild translations when ts files change

DomClark:
Currently, all the translations are compiled every time the build tool is run. This can generate a lot of annoying noise in the terminal. I have moved the `lrelease` command from `add_custom_target` to `add_custom_command`, so it should only run when its inputs change.


PhysSong (approved):

he29-net (approved):
I just tested it on my machine and it seems to work as intended (the "noise" is gone, and when I make a change to a .ts locale, the one changed file is picked up and rebuilt).

PR #6568, opened 2022-11-18
Fix several issues raised at Transifex

he29-net:
- change strings in spectrum analyzer to better work with different word orders
 - fix capitalization (Vst → VST)
 - fix typo (Mutliple → Multiple)
(Plus some small local code style updates.)

### DIFF
--- a/plugins/VstBase/VstPlugin.cpp
+++ b/plugins/VstBase/VstPlugin.cpp
@@ -579,32 +579,32 @@ void VstPlugin::savePreset()
 	QString presName = currentProgramName().isEmpty() ? tr(": default") : currentProgramName();
 	presName.replace("\"", "'"); // QFileDialog unable to handle double quotes properly
 
-	gui::FileDialog sfd( nullptr, tr( "Save Preset" ), presName.section(": ", 1, 1) + tr(".fxp"),
-		tr( "VST Plugin Preset (*.fxp *.fxb)" ) );
+	gui::FileDialog sfd(nullptr, tr("Save Preset"), presName.section(": ", 1, 1) + tr(".fxp"),
+		tr("VST Plugin Preset (*.fxp *.fxb)"));
 
-	if( p_name != "" ) // remember last directory
+	if (p_name != "") // remember last directory
 	{
-		sfd.setDirectory( QFileInfo( p_name ).absolutePath() );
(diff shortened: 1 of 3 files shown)


DomClark (approved):

plugins/VstBase/VstPlugin.cpp
```diff
@@ -586,7 +580,7 @@ void VstPlugin::savePreset()
 	presName.replace("\"", "'"); // QFileDialog unable to handle double quotes properly
 
 	gui::FileDialog sfd( nullptr, tr( "Save Preset" ), presName.section(": ", 1, 1) + tr(".fxp"),
-		tr( "Vst Plugin Preset (*.fxp *.fxb)" ) );
+		tr( "VST Plugin Preset (*.fxp *.fxb)" ) );
```
Veratil, line 583: You fix spacing on everything else but this one? 😂

he29-net: The `savePreset()` function was longer than `openPreset()`, so I did not want to touch it as I was afraid it would lead to a style-update chain reaction... :smile:

he29-net: There, the function should be fixed now. :)

Veratil: oh hah, you did the entire function. Just meant the line itself like the others. ;P

he29-net: > like the others

In the first commit I also updated the surrounding code in `openPreset()`, and a few lines around the "Mutliple" typo. I didn't want to leave it as a wild mix of two different styles, so I tried to at least update everything in related blocks of code. Sometimes it's just hard to decide where to stop. :))

liushuyu (approved):

plugins/VstBase/VstPlugin.cpp
Veratil: Usually just the line touched is what we go for on updating style. 👌 All good though!

he29-net: OK, I'll limit the changes next time. Probably makes sense; the diff would be easier to review as it would contain only relevant changes.

he29-net:
Reviews seem finished; anyone with write access wants to push the button? :)

PR #6580, opened 2022-11-30
Microtonality for SF2Player

LostRobotMusic:
SF2Player in its current state rounds microtuned notes to the closest 12EDO note, which obviously isn't desired behavior.  This PR makes use of `fluid_synth_activate_key_tuning` in order to adjust the number of cents each note is from MIDI note 0, to match the Hz of every note to the desired value.

### DIFF
--- a/plugins/Sf2Player/Sf2Player.cpp
+++ b/plugins/Sf2Player/Sf2Player.cpp
@@ -549,7 +549,6 @@ void Sf2Instrument::updateTuning()
 {
 	if (instrumentTrack()->microtuner()->enabledModel()->value())
 	{
-		int baseNote = instrumentTrack()->baseNoteModel()->value();
 		double * centArray = new double[128];
 		double lowestHz = powf(2.f, -69.f / 12.f) * 440.f;// Frequency of MIDI note 0, which is approximately 8.175798916 Hz
 		for (int i = 0; i < 128; ++i)
--- a/plugins/Sf2Player/Sf2Player.cpp
+++ b/plugins/Sf2Player/Sf2Player.cpp
@@ -644,6 +644,7 @@ void Sf2Instrument::reloadSynth()
 	updateReverbOn();
 	updateChorusOn();
 	updateGain();
+	updateTuning();
 
 	// Reset last MIDI pitch properties, which will be set to the correct values
 	// upon playing the next note
--- a/plugins/Sf2Player/Sf2Player.cpp
+++ b/plugins/Sf2Player/Sf2Player.cpp
@@ -183,12 +183,13 @@ Sf2Instrument::Sf2Instrument( InstrumentTrack * _instrument_track ) :
 	connect( &m_chorusDepth, SIGNAL( dataChanged() ), this, SLOT( updateChorus() ) );
 	
 	// Microtuning
-	connect(Engine::getSong(), SIGNAL(scaleListChanged(int)), this, SLOT(updateTuning()));
-	connect(Engine::getSong(), SIGNAL(keymapListChanged(int)), this, SLOT(updateTuning()));
-	connect(instrumentTrack()->microtuner()->enabledModel(), SIGNAL(dataChanged()), this, SLOT(updateTuning()), Qt::DirectConnection);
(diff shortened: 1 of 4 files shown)


plugins/Sf2Player/Sf2Player.cpp
```diff
@@ -537,6 +545,39 @@ void  Sf2Instrument::updateChorus()
...
+		double lowestHz = powf(2.f, -69.f / 12.f) * 440.f;// Frequency of MIDI note 0, which is approximately 8.175798916 Hz
+		for (int i = 0; i < 128; ++i)
+		{
+			// Get desired Hz of note
+			centArray[i] = instrumentTrack()->microtuner()->keyToFreq(i, DefaultBaseKey);
+			// Convert Hz to cents
+			centArray[i] = 1200.f * log2(centArray[i] / lowestHz);
+		}
+		
+		fluid_synth_activate_key_tuning(m_synth, 0, 0, "", centArray, true);
```
sakertooth, line 562: I forgot to mention changing this as well.
```suggestion
		fluid_synth_activate_key_tuning(m_synth, 0, 0, "", centArray.data(), true);
```

sakertooth (changes requested):

plugins/Sf2Player/Sf2Player.cpp
```diff
@@ -622,18 +663,19 @@ void Sf2Instrument::playNote( NotePlayHandle * _n, sampleFrame * )
...
 
 	const f_cnt_t tfp = _n->totalFramesPlayed();
+	
+	int masterPitch = instrumentTrack()->useMasterPitchModel()->value() ? Engine::getSong()->masterPitch() : 0;
+	int baseNote = instrumentTrack()->baseNoteModel()->value();
+	int midiNote = _n->midiKey() - baseNote + DefaultBaseKey + masterPitch;
 
-	if( tfp == 0 )
+	// out of range?
+	if( midiNote <= 0 || midiNote >= 128 )
```
DomClark, line 672: Not originally in scope, but while you're touching this bit of code, you could fix the range check to allow playing C-1:
```suggestion
	if (midiNote < 0 || midiNote >= 128)
```

plugins/Sf2Player/Sf2Player.cpp
```diff
@@ -537,6 +545,38 @@ void  Sf2Instrument::updateChorus()
...
+		double lowestHz = powf(2.f, -69.f / 12.f) * 440.f;// Frequency of MIDI note 0, which is approximately 8.175798916 Hz
+		for (int i = 0; i < 128; ++i)
+		{
+			// Get desired Hz of note
+			centArray[i] = instrumentTrack()->microtuner()->keyToFreq(i, DefaultBaseKey);
+			// Convert Hz to cents
+			centArray[i] = 1200.f * log2(centArray[i] / lowestHz);
+		}
+		
+		fluid_synth_activate_key_tuning(m_synth, 0, 0, "", centArray.data(), true);
```
DomClark, line 562: These tuning functions were introduced in FluidSynth 1.1.0, so the minimum required version in the root `CMakeLists.txt` should be bumped.

plugins/Sf2Player/Sf2Player.cpp
```diff
@@ -537,6 +545,38 @@ void  Sf2Instrument::updateChorus()
...
+	{
+		auto centArray = std::array<double, 128>{};
+		double lowestHz = powf(2.f, -69.f / 12.f) * 440.f;// Frequency of MIDI note 0, which is approximately 8.175798916 Hz
+		for (int i = 0; i < 128; ++i)
+		{
+			// Get desired Hz of note
+			centArray[i] = instrumentTrack()->microtuner()->keyToFreq(i, DefaultBaseKey);
+			// Convert Hz to cents
+			centArray[i] = 1200.f * log2(centArray[i] / lowestHz);
+		}
```
DomClark, line 560: It seems a bit odd to store everything as `double`s, but do the computation with `float`s. I would remove the `f` suffixes from the literals, and use `std::pow` instead. Also, I think it would be clearer to use a temporary variable for the note frequency, rather than using `centArray[i]` for two different things.

DomClark (commented):
A few more things I didn't spot the first time through - sorry for dragging this out.

DomClark (approved):

PhysSong (approved):

PR #6583, opened 2022-12-06
Fix unwanted scrolling when changing track name

### LINKED ISSUE #6343: Unwanted scroll when changing a track name in the Song-Editor and Pattern-Editor
### Bug Summary
When the Song-Editor has more tracks that those visible and you change the name of a track that is close to the end of the window, an unwanted scroll is produced. That happens also with Pattern-Editor. It happens also if you enter <ESC> without changing the track name.

#### Steps to reproduce
Identified while testing #6338 
You can load the project on #6334 and change the name of "Sandpaper Jam" track And track "Snr_snap" in Pattern-Editor

#### Expected behavior
Just a track rename.

#### Actual behavior
The track rename is ok but there is an unwanted scroll.

#### Screenshot
See attached image

#### Affected LMMS versions
Tested on various 1.3.*

unwanted-scroll-bug

sakertooth:
When changing the track name in the Song Editor, ``m_renameLineEdit`` gains focus and subsequently hides without losing focus after the track has been renamed. This causes the scroll area in the Song Editor to unnecessarily jump to the track that was renamed. To prevent this, ``m_renameLineEdit`` should lose focus before hiding.

Fixes #6343.

### DIFF
--- a/src/gui/tracks/TrackLabelButton.cpp
+++ b/src/gui/tracks/TrackLabelButton.cpp
@@ -105,6 +105,7 @@ void TrackLabelButton::renameFinished()
 {
 	if( !( ConfigManager::inst()->value( "ui", "compacttrackbuttons" ).toInt() ) )
 	{
+		m_renameLineEdit->clearFocus();
 		m_renameLineEdit->hide();
 		if( m_renameLineEdit->text() != "" )
 		{


superpaik:
It works as described (Tested on Minwin64-Win10)
However, there is one side-effect that maybe can be addressed. That is when the last row is not fully visible. See attached gif. It happens also for the first row. 
To me, the expected behaviour would be a jump, just when the row gains focus, up or down, to make it visible.
Test#6583

sakertooth:
> To me, the expected behaviour would be a jump, just when the row gains focus, up or down, to make it visible.

If you ever needed to edit the last row when it is not fully visible, the user can simply scroll to get it in view. I would say that the original situation of scrolling to something that is already in view can be a bit unwanted and unnecessary, but having it only scroll on the last row may be unnecessary as well.

superpaik:
I think it's better as it is now, because before there were an expected jump of 3 (i think) rows.
I agree with you that the user can scroll to make the row in view. I was only suggesting if by the look at the code you could easily make that last (or first) row visible. But I know that it's not always easy to implement.
So, LGTM.

DomClark (approved):

zonkmachine (approved):
Tested, works well. I prefer this over the current behaviour.

sakertooth:
> Tested, works well. I prefer this over the current behaviour.

Great. Anyone willing to merge this?

Issue #6346, opened 2022-03-26
Drag Beat+Bassline Pattern the other way in the Song-Editor


actuallyexeon:
### Enhancement Summary
When using a Beat+Bassline Pattern in the Song-Editor you can drag it from the right to left to shorten the pattern in that direction, but what would also be nice is to be able to go from left to right so you can have the other half or quarter of the pattern instead of the first half only.

#### Justification
This would be nice when making songs like when you need to implement a beat repeat that repeats a certain part but then goes back to normal and instead of having to duplicate all of the patterns in your Song-Editor you can just simply drag them from left to right to keep a certain part.

ToastedSadWitch:
hi, i would like to work on this :)
this is my first work on lmms so if you think it's too advanced, you can say no.
right now i figure out the gui part of the problem (there is already the starttimeoffset parameter we can make use of).
another major part would be the playback.
and a minor one the storing + compatibility with older version files.

zonkmachine:
> hi, i would like to work on this :)

Sure, have a go!
If you haven't joined our Discord channel yet, now is the time.
#dev-support is the first place to ask If you get stuck!

ToastedSadWitch:
thanks! i'm already on the discord as Posoppiser 👍

PR #6584, opened 2022-12-09
PatternClips now use startTimeOffset, too, and can be resized on their start

ToastedSadWitch:
I checked the gui part, playback, export, copying and storing.
When dropping a pattern clip with startTimeOffset into another pattern clip, the startTimeOffset is currently not copied. If this is no wanted behaviour, please tell me, so I can try to fix it.
Also the behaviour of resizing the start is faulty sometimes when colliding with the start of the song editor (clip may grow or there is a padding sometimes). As this also happens on the sample track, it seems to be a general deal with ResizeLeft action.
This commit is for issue #6346.

### DIFF
--- a/src/core/PatternClip.cpp
+++ b/src/core/PatternClip.cpp
@@ -60,6 +60,7 @@ void PatternClip::saveSettings(QDomDocument& doc, QDomElement& element)
 		element.setAttribute( "pos", startPosition() );
 	}
 	element.setAttribute( "len", length() );
+	element.setAttribute("off", startTimeOffset());
 	element.setAttribute( "muted", isMuted() );
 	if( usesCustomClipColor() )
 	{
@@ -78,6 +79,7 @@ void PatternClip::loadSettings(const QDomElement& element)
 		movePosition( element.attribute( "pos" ).toInt() );
 	}
 	changeLength( element.attribute( "len" ).toInt() );
+	setStartTimeOffset(element.attribute("off").toInt());
 	if( element.attribute( "muted" ).toInt() != isMuted() )
 	{
 		toggleMute();
--- a/src/gui/clips/ClipView.cpp
+++ b/src/gui/clips/ClipView.cpp
@@ -43,6 +43,8 @@
 #include "MidiClip.h"
 #include "MidiClipView.h"
 #include "Note.h"
+#include "PatternClip.h"
+#include "PatternStore.h"
 #include "SampleClip.h"
 #include "Song.h"
 #include "SongEditor.h"
@@ -502,10 +504,11 @@ void ClipView::dropEvent( QDropEvent * de )
 void ClipView::updateCursor(QMouseEvent * me)
 {
 	auto sClip = dynamic_cast<SampleClip*>(m_clip);
+	auto pClip = dynamic_cast<PatternClip*>(m_clip);
 
 	// If we are at the edges, use the resize cursor
 	if (!me->buttons() && !m_clip->getAutoResize() && !isSelected()
-		&& ((me->x() > width() - RESIZE_GRIP_WIDTH) || (me->x() < RESIZE_GRIP_WIDTH && sClip)))
+		&& ((me->x() > width() - RESIZE_GRIP_WIDTH) || (me->x() < RESIZE_GRIP_WIDTH && (sClip || pClip))))
 	{
 		setCursor(Qt::SizeHorCursor);
 	}
@@ -633,6 +636,7 @@ void ClipView::mousePressEvent( QMouseEvent * me )
 	if( !fixedClips() && me->button() == Qt::LeftButton )
 	{
 		auto sClip = dynamic_cast<SampleClip*>(m_clip);
+		auto pClip = dynamic_cast<PatternClip*>(m_clip);
 		const bool knifeMode = m_trackView->trackContainerView()->knifeMode();
 
 		if ( me->modifiers() & Qt::ControlModifier && !(sClip && knifeMode) )
@@ -677,7 +681,7 @@ void ClipView::mousePressEvent( QMouseEvent * me )
 					m_action = Resize;
 					setCursor( Qt::SizeHorCursor );
 				}
-				else if( me->x() < RESIZE_GRIP_WIDTH && sClip )
+				else if( me->x() < RESIZE_GRIP_WIDTH && (sClip || pClip) )
 				{
 					m_action = ResizeLeft;
 					setCursor( Qt::SizeHorCursor );
@@ -836,7 +840,6 @@ void ClipView::mouseMoveEvent( QMouseEvent * me )
 	if( m_action == Move )
 	{
 		TimePos newPos = draggedClipPos( me );
-
 		m_clip->movePosition(newPos);
 		newPos = m_clip->startPosition(); // Get the real position the Clip was dragged to for the label
 		m_trackView->getTrackContentWidget()->changePosition();
@@ -916,7 +919,8 @@ void ClipView::mouseMoveEvent( QMouseEvent * me )
 		else
 		{
 			auto sClip = dynamic_cast<SampleClip*>(m_clip);
-			if( sClip )
+			auto pClip = dynamic_cast<PatternClip*>(m_clip);
+			if( sClip || pClip )
 			{
 				const int x = mapToParent( me->pos() ).x() - m_initialMousePos.x();
 
@@ -948,12 +952,27 @@ void ClipView::mouseMoveEvent( QMouseEvent * me )
 					t = qMin<int>( m_initialClipEnd - minLength, m_initialClipPos + offset );
 				}
 
-				TimePos oldPos = m_clip->startPosition();
-				if( m_clip->length() + ( oldPos - t ) >= 1 )
+				TimePos positionOffset = m_clip->startPosition() - t;
+				if (m_clip->length() + positionOffset >= 1)
 				{
-					m_clip->movePosition( t );
-					m_clip->changeLength( m_clip->length() + ( oldPos - t ) );
-					sClip->setStartTimeOffset( sClip->startTimeOffset() + ( oldPos - t ) );
+					m_clip->movePosition(t);
+					m_clip->changeLength(m_clip->length() + positionOffset);
+					if (sClip)
+					{
+						sClip->setStartTimeOffset(sClip->startTimeOffset() + positionOffset);
+					}
+					else if (pClip)
+					{
+						// Modulus the start time offset as we need it only for offsets
+						// inside the pattern length. This is done to prevent a value overflow.
+						// The start time offset may still become larger than the pattern length
+						// whenever the pattern length decreases without a clip resize following.
+						// To deal safely with it, always modulus before use.
+						tick_t patternLength = Engine::patternStore()->lengthOfPattern(pClip->patternIndex())
+								* TimePos::ticksPerBar();
+						TimePos position = (pClip->startTimeOffset() + positionOffset) % patternLength;
+						pClip->setStartTimeOffset(position);
+					}
 				}
 			}
 		}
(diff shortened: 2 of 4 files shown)


superpaik:
Tested on W10-64. It works ok.
I was not able to reproduce that "When dropping a pattern clip with startTimeOffset into another pattern clip, the startTimeOffset is currently not copied. If this is no wanted behaviour, please tell me, so I can try to fix it." In my case, when dropped over another existing clip, it's replaced ok. In any case, I think that's the desired behaviour.

src/tracks/PatternTrack.cpp
```diff
@@ -113,8 +115,23 @@ bool PatternTrack::play( const TimePos & _start, const fpp_t _frames,
...
+			tick_t offset = fmod(static_cast<double>(clip->startTimeOffset()) / TimePos::ticksPerBar(), patLen)
+					* TimePos::ticksPerBar();
+			if (offset == 0)
+			{
+				lastPosition = clip->startPosition();
+				lastLen = clip->length();
+			}
+			else
+			{
+				lastPosition = clip->startPosition() - (patTicks - offset);
```
DomClark, line 132: This will affect the condition at the beginning of the loop that attempts to find the clip with the latest start time. I'm not sure how much that matters, but perhaps it would better to avoid it by storing `patTicks - offset` in a `lastOffset` variable or something, then adding that in the call to `play` below?

DomClark (commented):
Thanks for the pull request! This looks really useful. I've left a few comments on the code - feel free to quibble them if you think I'm talking nonsense, or ask for more detail if I haven't been clear enough.

src/gui/clips/ClipView.cpp
```diff
@@ -948,12 +952,27 @@ void ClipView::mouseMoveEvent( QMouseEvent * me )
...
+					if (sClip)
+					{
+						sClip->setStartTimeOffset(sClip->startTimeOffset() + posOff);
+					}
+					else if (pClip)
+					{
+						/**
+							modulus the start time offset as we need it only for offsets
+							inside the pattern length
+						*/
```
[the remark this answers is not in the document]
ToastedSadWitch: I guess "When submitting a PR, normal comments should be turned into doxygen comments." from the LMMS Coding Conventions confused me here.

src/gui/clips/ClipView.cpp
```diff
@@ -948,12 +952,27 @@ void ClipView::mouseMoveEvent( QMouseEvent * me )
...
+						bar_t patLen = Engine::patternStore()->lengthOfPattern(pClip->patternIndex());
+						TimePos pos = fmod(static_cast<double>(pClip->startTimeOffset() + posOff)
+								/ TimePos::ticksPerBar(), patLen)
+								* TimePos::ticksPerBar();
```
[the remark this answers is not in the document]
ToastedSadWitch: Yeah, I was thinking too complicated, I guess. Thank you!

src/tracks/PatternTrack.cpp
```diff
@@ -113,8 +115,23 @@ bool PatternTrack::play( const TimePos & _start, const fpp_t _frames,
 	{
 		if (!clip->isMuted() && clip->startPosition() >= lastPosition)
 		{
-			lastPosition = clip->startPosition();
-			lastLen = clip->length();
+			/**
+				calculate an offset in playback according to the start time offset of the clip
+			*/
+			bar_t patLen = Engine::patternStore()->lengthOfPattern(dynamic_cast<PatternClip*>(clip)->patternIndex());
```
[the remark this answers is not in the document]
ToastedSadWitch: I'm new to C++ and basically copied code here but this absolutely makes sense to me now 👍

src/tracks/PatternTrack.cpp
ToastedSadWitch: Now as you say, I am astonished it worked before, lol. It should be more self-explaining now. I also reduced the logic a little.

ToastedSadWitch:
build fails should be fixed with this

src/gui/clips/ClipView.cpp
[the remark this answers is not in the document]
DomClark: I see what you mean. I'll update the Coding Conventions to be clearer.

src/gui/clips/PatternClipView.cpp
```diff
@@ -115,12 +117,15 @@ void PatternClipView::paintEvent(QPaintEvent*)
 	
 	// bar lines
 	const int lineSize = 3;
+	bar_t t = Engine::patternStore()->lengthOfPattern(m_patternClip->patternIndex());
+	int offset = fmod(static_cast<double>(m_patternClip->startTimeOffset()) / TimePos::ticksPerBar(), t)
+			* pixelsPerBar();
```
DomClark, line 122: Just one more floating point calculation here that could be converted to use integers like the others.

ToastedSadWitch: I actually kept it because I did not introduce variable t. It is used a few lines later in the loop together with pixelsPerBar(). I also need to use pixelsPerBar() for conversion of the startTimeOffset to pixels. To work with ticks instead (and therefore not divide) would require either of two options:
a) dividing pixelsPerBar by ticksPerBar to then multiply with ticks - this basically shifts the problem; or
b) writing a second pixelsPerTick member which internally does not rely on division - otherwise it would just shift the problem again.
I don't know if it is possible to not use any division in this case without going deeper into code (float m_ppb and stuff).
Please tell me, if I missed a simple solution here.

ToastedSadWitch: I could drop the use fmod, though: `int offset = (m_patternClip->startTimeOffset() % (t * TimePos::ticksPerBar())) * (pixelsPerBar() / TimePos::ticksPerBar());`.
Would that be more preferable?

ToastedSadWitch: I will drop the use of fmod though and create a pixelsPerPattern variable to boost the loop.

DomClark: That makes sense - there's no need to overcomplicate things here, so the changes look good as you now have them.

src/gui/clips/PatternClipView.cpp
```diff
@@ -115,14 +115,18 @@ void PatternClipView::paintEvent(QPaintEvent*)
...
+			% pixelsPerPattern;
+	if (offset <= 1) {
+		offset += pixelsPerPattern;
+	}
+
 	p.setPen( c.darker( 200 ) );
 
-	bar_t t = Engine::patternStore()->lengthOfPattern(m_patternClip->patternIndex());
-	if (m_patternClip->length() > TimePos::ticksPerBar() && t > 0)
+	if (pixelsPerPattern > 0 && offset < width() - 2)
```
DomClark, line 127: I think the `offset < width() - 2` condition here is superfluous: the loop won't run anyway if that's not the case. But there's nothing wrong about it, so I'm happy for it to be left as-is.

DomClark (approved):

src/gui/clips/PatternClipView.cpp
ToastedSadWitch: No problem, I will delete it. I thought it might be good to not initiate an unnecessary loop but it shouldn't make a big difference in performance here. So it's better to keep it clean for others. 👍

src/core/PatternClip.cpp
```diff
@@ -60,6 +60,7 @@ void PatternClip::saveSettings(QDomDocument& doc, QDomElement& element)
 		element.setAttribute( "pos", startPosition() );
 	}
 	element.setAttribute( "len", length() );
+	element.setAttribute("off", startTimeOffset());
```
Veratil, line 63: I'd prefer we use the full description of the attribute, `offset`, instead of `off`. It can be confused with `off` (meaning turn off).

Granted this is in a file that no one should edit manually, but there's no reason we can't be descriptive.

src/gui/clips/ClipView.cpp
```diff
@@ -948,12 +952,27 @@ void ClipView::mouseMoveEvent( QMouseEvent * me )
 					t = qMin<int>( m_initialClipEnd - minLength, m_initialClipPos + offset );
 				}
 
-				TimePos oldPos = m_clip->startPosition();
-				if( m_clip->length() + ( oldPos - t ) >= 1 )
+				TimePos posOff = m_clip->startPosition() - t;
```
Veratil, line 955: Similar to previous comment, `posOffset` would be easier on quick reading.

src/tracks/PatternTrack.cpp
```diff
@@ -109,18 +109,26 @@ bool PatternTrack::play( const TimePos & _start, const fpp_t _frames,
...
 	TimePos lastPosition;
 	TimePos lastLen;
+	tick_t lastOffset = 0;
 	for (const auto& clip : clips)
 	{
 		if (!clip->isMuted() && clip->startPosition() >= lastPosition)
 		{
 			lastPosition = clip->startPosition();
 			lastLen = clip->length();
+			tick_t patLen = Engine::patternStore()->lengthOfPattern(static_cast<PatternClip*>(clip)->patternIndex())
```
Veratil, line 119: And again with `patLength`, or even better `patternLength`. :)

Veratil (commented):
Just some naming thoughts. Feel free to ignore for merge.

src/gui/clips/ClipView.cpp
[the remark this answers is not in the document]
DomClark: Finally got round to doing this. Let me know on Discord if any further tweaks are needed.

src/core/PatternClip.cpp
ToastedSadWitch: I just copied this line over from the SampleClip.cpp. I thought it would be good to be consistent across files here. Guess, I will just change both now before it breaks any files after a release.
And I indeed had to edit project files manually some time ago (sample paths weren't matching anymore).

src/tracks/PatternTrack.cpp
ToastedSadWitch: To be honest, when I write code, I normally always type things out, even if it gets like 20 characters wide. As this is my first contribution here, I tried to keep the style but if it is okay, I would be glad using more transparent namings in the future.

src/core/SampleClip.cpp
```diff
@@ -260,7 +260,7 @@ void SampleClip::saveSettings( QDomDocument & _doc, QDomElement & _this )
 	_this.setAttribute( "len", length() );
 	_this.setAttribute( "muted", isMuted() );
 	_this.setAttribute( "src", sampleFile() );
-	_this.setAttribute( "off", startTimeOffset() );
+	_this.setAttribute("offset", startTimeOffset());
```
DomClark, line 263: This attribute name needs to stay the same for backwards compatibility. (Alternatively, you could write a new upgrade routine to change it, but I'd rather we did that separately and fixed a bunch of other names while we were at it.) I'm also inclined to use "off" for now in `PatternClip` for consistency, but I don't mind so much about that.

ToastedSadWitch: undone

ToastedSadWitch:
Final fix from my side.

src/core/PatternClip.cpp
ToastedSadWitch: See newest review.

PhysSong: > Guess, I will just change both now before it breaks any files after a release.

There are at least three choices:
- Leave everything as-is
- Change `off` to `offset` only in `PatternClip`: inconsistent with `SampleClip`, but no compatibility issues
- Change both in `PatternClip` and `SampleClip`: offsets in new files will be ignored in old versions, the converse can be handled by either trying with `off` as a fallback or using `DataFile::upgrade`

Veratil:
@ToastedSadWitch since the vcpkg error isn't something on our side we can ignore that. :)

PR #6586, opened 2022-12-13
Add mixer dB markers

sakertooth:
This PR adds dB markers for mixer channels. This is especially useful when leveling mixes, and is generally useful when you want to simply estimate at what dB a fader is peaking at.

**Screenshot**
image

### DIFF
--- a/include/MixerDbMarkers.h
+++ b/include/MixerDbMarkers.h
@@ -1,0 +1,47 @@
+/*
+ * MixerDbMarkers.h - dB markers for mixer channels
+ *
+ * Copyright (c) 2022 saker <sakertooth@gmail.com>
+ *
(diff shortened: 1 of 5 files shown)


sakertooth:
Just seeing the meta issue for this: It seems like there's a lot more work to be done here.

For one, the mixer needs to move in logarithms, not linearly. The faders should also point to the top, as 0 dbFS is the ceiling for the audio signal.

superpaik:
I totally agree that the scale should by logarithmic, it would be more useful. As for the top, it depens as how the internal signal is managed, but I in other DAWs it goes up to +6dB too.

actuallyexeon:
I think it would be better if it was slimmer on the sides

sakertooth:
Closing due to #7295.

Issue #6383, opened 2022-04-20
Pause continues to play rest of note


Monospace-V:
### Bug Summary
I'm not entirely sure if this is intended behaviour, but pausing a song continues to play the rest of the length of the ongoing note instead of directly stopping sound, or perhaps going to Release.
Playing the rest of the note even upon pausing is unintuitive and confusing behaviour.

#### Steps to reproduce
1. In any instrument plugin, place a long note.
2. From Song Editor or Piano roll, play the start of the note
3. Pause the note. Don't stop it, pause it.
4. Rest of the note continues to play.

#### Expected behavior
Note goes silent.

#### Actual behavior
Rest of note continues to play.

#### Affected LMMS versions
1.2.0, 1.2.2, 1.3.0

_(A friend of mine who started using LMMS recently, typically uses pause instead of stop (while most of us use stop most of the time), notices these things more often than me. By the time I started filing bugs, I'd gotten used to this behaviour, and he's the reason I can point out unusual pause behaviour now- because I've gotten used to them and don't realise what's going on. New users often don't even understand the behaviour in the first place. To me, I already knew that pausing plays the rest of any ongoing note, but never stopped to think about it as unwanted or considered filing a report. This information isn't particularly useful to the report itself, but I think it's a telling bit of insight. What it tells, I know not; but there's definitely something to be gotten from this.)_

RiedleroD:
I believe this is intended - but I also think it should be changed. Looking back, especially long bass / pad notes were always very annoying when pausing. I usually use stop instead, but this throws the view back to the start in longer tracks when I forget to disable the thingy that does that.

musikBear:
"Intended" Naee.. But architecturely hard to avoid 
Funny 'Ha-haar' wise..
Pausing will actually not be different from stopping (space) if the transport is set to ">".
Playback will stop, and next press on space will start the playhead where it was ..so in fact :
Playhead was **Paused** :p

Issue #4410, opened 2018-06-08
Piano Roll plays previous paused pattern instead of current pattern


EtornaZ:
I'm on version RC6 of LMMS, with a Windows 10 64Bit computer.

When a box of notes is selected, played and then paused, and a new box of notes is selected and played, the old box of notes plays instead of the new box of notes. If you press the stop button and then press play, the new box of notes will play as intended.

GingkathFox:
I think he means chord groups @SecondFlight 
play part of a chord, pause, skip ahead to second chord, error happens? @EtornaZ

EtornaZ:
I'm sorry, I wasn't very clear at all. Here are some screenshots to hopefully clear the confusion up.
![image](https://user-images.githubusercontent.com/33297131/41160484-e0adc4f0-6afd-11e8-864b-8a7463cc7001.png)
I have a selection of notes in the Piano Roll above. I play them, and then pause the notes.
![image](https://user-images.githubusercontent.com/33297131/41160631-50327686-6afe-11e8-9227-7eaafff83101.png)
I now have a new set of notes selected above, but when I press play, the old selection of notes plays instead of the new selection. If I press stop and then replay, the new notes will play just fine.

GingkathFox:
_sneak peek at new track? lol_
Aight, that helps!

SecondFlight:
Thanks for the screenshots! So just to confirm, when you say "note selection" you're talking about patterns?

EtornaZ:
Yes. I didn't know the correct name for them until now, heh heh.

SecondFlight:
I'm using the FL term for them, I think it fits :)

GingkathFox:
Yeah, that fits lol

GingkathFox:
isnt there a debugging thing @SecondFlight? maybe u could use that @EtornaZ

SecondFlight:
I don't know @GingkathFox, I haven't been active here for very long. If it's reproducible then that wouldn't be needed. I haven't had time to attempt to reproduce yet, but if I remember I'll do that this evening.

GingkathFox:
ok!

musikBear:
Nahh Firstly, nothing is selected. Neither in piano-roll, nor in song-editor.
The *lightend* TCO shows that it is the one *currently opend in piano-roll!*
Even if you close piano-roll, it is still focused, and is still the one that are ready to play.
[...]

PR #6590, opened 2022-12-19
Fix issues #6383 and #4410 (pause issues).

madisonsilver:
The change to Song.cpp fixes issue #6383.

The change to PianoRoll.cpp fixes issue #4410.

### DIFF
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -858,14 +858,9 @@ void PianoRoll::setCurrentMidiClip( MidiClip* newMidiClip )
 	}
 
 	// force the song-editor to stop playing if it played a MIDI clip before
-	if( Engine::getSong()->isPlaying() &&
-		Engine::getSong()->playMode() == Song::Mode_PlayMidiClip )
+	if (Engine::getSong()->playMode() == Song::Mode_PlayMidiClip)
 	{
 		Engine::getSong()->playMidiClip( nullptr );
-	}
-	else if (Engine::getSong()->isPaused())
-	{
-		Engine::getSong()->playMidiClip(nullptr);
 	}
 
 	if(m_stepRecorder.isRecording())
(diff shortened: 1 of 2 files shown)


DomClark (approved):

src/gui/editors/PianoRoll.cpp
```diff
@@ -863,6 +863,10 @@ void PianoRoll::setCurrentMidiClip( MidiClip* newMidiClip )
 	{
 		Engine::getSong()->playMidiClip( nullptr );
 	}
+	else if (Engine::getSong()->isPaused())
```
PhysSong, line 866: I wonder if it's sufficient to do this in `Engine::getSong()->isPaused() && Engine::getSong()->playMode() == Song::Mode_PlayMidiClip`  cases. If it is, this can be merged with the `if` statement above.

PhysSong: This has a undesirable side-effect: Opening a MIDI clip in piano roll when the song is paused will change the playback state from paused to stopped, by calling `stop` in `playMidiClip`.
I think the correct solution is removing `Engine::getSong()->isPlaying()` from the `if` state above.

PhysSong (changes requested):

src/gui/editors/PianoRoll.cpp
PhysSong: To reproduce:
1. Add a MIDI clip
2. Set the playhead mode to `|<<` or `<<`
3. Play the song in the song editor
4. Pause
5. Open the MIDI clip
6. The playhead will jump back as if the stop button is pressed

PhysSong (approved):

RiedleroD:
wow, the second one was convoluted. Took me a bit to realise what was going on. Well, now I know why LMMS was playing notes from a different pattern seemingly at random.

I hope this explanation of the bug is sufficient, I gave my best: https://mastodon.social/@LMMS/109920642097039207

PR #6594, opened 2022-12-22
Resample SampleBuffer only once when loading from SampleClip

### LINKED ISSUE #6574: Anomaly in resampling of files on Sample Tracks
Currently the sample rate of the sample buffer gets saved into the project file with the value lmms was running at when saving, and I can see the purpose of doing this for samples recorded in lmms.
But the thing is that sample tracks in a project saved at 44.1 kHz get pitched down(!) when running at 48 kHz or higher, most probably because of audio files already getting resampled to the current sample rate on project loading too, resulting in resampling two times.

sakertooth:
The ``SampleBuffer``'s sample rate in ``SampleClip`` was altered twice during ``SampleClip::loadSettings``: first when ``setSampleFile`` was called, which set the sample rate of the ``SampleBuffer`` to the ``AudioEngine``'s sample rate (good), and a second time when calling ``setSampleRate``, which set it to the sample rate specified within the project file (bad). This led to the sample rate of the buffer being different than that of the project, resulting in it being pitched incorrectly on playback.

I believe more improvement can be done here (I am questioning the need of setting the sample rate of a ``SampleBuffer`` as it should match the sample rate of the project), but for the sake of keeping it simple, I have only implemented the fix for now.

Fixes #6574.

### DIFF
--- a/src/core/SampleClip.cpp
+++ b/src/core/SampleClip.cpp
@@ -292,6 +292,10 @@ void SampleClip::loadSettings( const QDomElement & _this )
 	if( sampleFile().isEmpty() && _this.hasAttribute( "data" ) )
 	{
 		m_sampleBuffer->loadFromBase64( _this.attribute( "data" ) );
+		if (_this.hasAttribute("sample_rate"))
+		{
+			m_sampleBuffer->setSampleRate(_this.attribute("sample_rate").toInt());
+		}
 	}
 	changeLength( _this.attribute( "len" ).toInt() );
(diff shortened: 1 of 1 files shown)


PhysSong:
The part is required for base64-saved audio data to have the correct sample rate, but it's affecting sample files in a wrong way as well. I think what you should do is moving the block next to `loadFromBase64`.

PhysSong (approved):

DomClark (approved):

PR #6595, opened 2022-12-22
Add MSVC support for GigPlayer

Veratil:
* Add FindGig.cmake
  * Tested on Windows 10 with vcpkg (pkg-config not present)
  * Tested on Fedora 36 with localinstall rpm's (pkg-config present)
* Fixes GigPlayer CMakeLists.txt for MSVC
* Patches GigPlayer for MSVC builds
  * Tested on Windows 10 (VS 17 2022)
  * Tested on Fedora 36 (gcc-12.2.1)

vcpkg port for testing can be found here: https://github.com/Veratil/vcpkg/tree/add-port-libgig


plugins/GigPlayer/CMakeLists.txt
```diff
@@ -3,14 +3,16 @@ if(LMMS_HAVE_GIG)
 	INCLUDE_DIRECTORIES(${GIG_INCLUDE_DIRS})
 
 	# Required for not crashing loading files with libgig
-	SET(GCC_COVERAGE_COMPILE_FLAGS "-fexceptions")
-	add_definitions(${GCC_COVERAGE_COMPILE_FLAGS})
+	if(NOT MSVC)
+		SET(GCC_COVERAGE_COMPILE_FLAGS "-fexceptions")
+		add_definitions(${GCC_COVERAGE_COMPILE_FLAGS})
+	endif(NOT MSVC)
```
PhysSong, line 9: This is a leftover from #4000. All occurences of `-fexceptions` can now be removed.

plugins/GigPlayer/GigPlayer.cpp
```diff
@@ -531,7 +539,11 @@ void GigInstrument::loadSample( GigSample& sample, sampleFrame* sampleData, f_cn
 	}
 
 	unsigned long allocationsize = samples * sample.sample->FrameSize;
+#ifndef _MSC_VER
 	int8_t buffer[allocationsize];
+#else
+	auto buffer = static_cast<int8_t*>(_alloca(allocationsize * sizeof(int8_t)));
```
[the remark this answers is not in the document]
Veratil: I had `const` originally and it would not compile, but I will try again after fixing the `&buffer` parts.

plugins/GigPlayer/CMakeLists.txt
```diff
@@ -3,14 +3,16 @@ if(LMMS_HAVE_GIG)
...
 
 	# disable deprecated check for mingw-x-libgig
-	if(LMMS_BUILD_WIN32)
+	if(LMMS_BUILD_WIN32 AND MINGW)
 		SET(GCC_GIG_COMPILE_FLAGS "-Wno-deprecated")
 		add_definitions(${GCC_GIG_COMPILE_FLAGS})
-	endif(LMMS_BUILD_WIN32)
+	endif(LMMS_BUILD_WIN32 AND MINGW)
```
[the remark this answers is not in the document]
Veratil: Oh good to know!

cmake/modules/FindGig.cmake
```diff
@@ -0,0 +1,34 @@
...
+if(PKG_CONFIG_FOUND)
+	pkg_check_modules(LIBGIG_PKG gig)
+endif()
+
+if(WIN32)
+	find_package(libgig ${Gig_FIND_VERSION} CONFIG QUIET)
+	if(libgig_FOUND)
+		get_target_property(libgig_Location libgig::libgig LOCATION)
+		get_target_property(libgig_Include_Path libgig::libgig INTERFACE_INCLUDE_DIRECTORIES)
+		get_filename_component(libgig_Path LibGig_Location PATH)
```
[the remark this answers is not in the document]
Veratil: This is beyond what I know of CMake, but if you have a reference of another done then I can work on converting it.

cmake/modules/FindGig.cmake
```diff
@@ -0,0 +1,34 @@
+# Attempt to find libigg using PkgConfig
+find_package(PkgConfig QUIET)
+if(PKG_CONFIG_FOUND)
+	pkg_check_modules(LIBGIG_PKG gig)
+endif()
+
+if(WIN32)
+	find_package(libgig ${Gig_FIND_VERSION} CONFIG QUIET)
```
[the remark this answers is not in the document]
Veratil: Should we create a new PR to convert the others? I pretty much copied this from `FindFluidsynth`.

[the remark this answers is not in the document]
Veratil: We can chat about my question in Discord later :)

PhysSong:
@Veratil BTW, are you going to submit a PR against the vcpkg upstream?

Veratil:
> @Veratil BTW, are you going to submit a PR against the vcpkg upstream?

I'm not too familiar with vcpkg so I was hoping Dom could check out the branch and see if there's anything I missed. You're welcome to do it, too, though. I see vcpkg has a very specific way to add new ports, so I'd have to go through and fix whatever I didn't do properly.

cmake/modules/FindGig.cmake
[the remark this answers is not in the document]
DomClark: `FindPortaudio.cmake` is probably closest to what you're trying to do, but `FindLame.cmake` and `FindSTK.cmake` are quite similar too. `FindFluidSynth.cmake` and `FindSDL2.cmake` are a bit more complex, but should serve as further examples.

PhysSong:
> > @Veratil BTW, are you going to submit a PR against the vcpkg upstream?
> 
> I'm not too familiar with vcpkg so I was hoping Dom could check out the branch and see if there's anything I missed. You're welcome to do it, too, though. I see vcpkg has a very specific way to add new ports, so I'd have to go through and fix whatever I didn't do properly.

I'm not sure if I'm doing things correctly, but here's what I'm doing: https://github.com/PhysSong/vcpkg/tree/libgig

plugins/GigPlayer/GigPlayer.cpp
```diff
@@ -576,14 +588,14 @@ void GigInstrument::loadSample( GigSample& sample, sampleFrame* sampleData, f_cn
 	{
 		sample.sample->SetPos( sample.pos );
 
-		unsigned long size = sample.sample->Read( &buffer, samples ) * sample.sample->FrameSize;
-		std::memset( (int8_t*) &buffer + size, 0, allocationsize - size );
+		unsigned long size = sample.sample->Read( buffer, samples ) * sample.sample->FrameSize;
+		std::memset( (int8_t*) buffer + size, 0, allocationsize - size );
```
sakertooth, line 592:
```suggestion
		const auto size = sample.sample->Read(buffer, samples) * sample.sample->FrameSize;
		std::memset(buffer + size, 0, allocationsize - size);
```

Perhaps this can be written a bit better?

Veratil: There are a lot of changes that should happen here in GigPlayer, but I decided to leave this as a minimum viable PR to get it to build on Windows. Definitely something that should be updated in a followup.

Rossmaxx:
What's the state of this pr now? Is the gig port going to vcpkg or would it be fine enough to submodule it and get it merged, then followed up by a submodule removal pr when libgig becomes available on vcpkg.

Veratil:
> What's the state of this pr now? Is the gig port going to vcpkg or would it be fine enough to submodule it and get it merged, then followed up by a submodule removal pr when libgig becomes available on vcpkg.

libgig still needs to be PR'd to vcpkg, then this could be merged. I just haven't done it.

Rossmaxx:
> libgig still needs to be PR'd to vcpkg, then this could be merged. I just haven't done it.

Is it ok if I open the pr before you?

Rossmaxx:
PR opened upstream [here](https://github.com/microsoft/vcpkg/pull/33255)

plugins/GigPlayer/GigPlayer.cpp
```diff
@@ -460,7 +464,11 @@ void GigInstrument::play( sampleFrame * _working_buffer )
 			// Output the data resampling if needed
 			if( resample == true )
 			{
+#ifndef _MSC_VER
 				sampleFrame convertBuf[frames];
+#else
+				const auto convertBuf = static_cast<sampleFrame*>(_alloca(frames * sizeof(sampleFrame)));
+#endif
```
DomClark, line 471: Try to avoid calling `_alloca` within a loop - it can easily lead to a stack overflow.

cmake/modules/FindGig.cmake
```diff
@@ -0,0 +1,35 @@
+
```
DomClark, line 1: There's a blank line here. Also, do you want to add a copyright notice to this file?

plugins/GigPlayer/GigPlayer.cpp
Veratil: What should we use instead? I don't recall where I got this from but I do remember MSVC did not like `convertBuf` as it's defined above.

cmake/modules/FindGig.cmake
Veratil: I wasn't sure what copyright to put since it's nearly a direct copy of some others.

plugins/GigPlayer/GigPlayer.cpp
DomClark: Lift the variable out of the loop if possible - it should be fine to reuse the same buffer for each iteration. There's another instance in this file where the array size is computed inside the loop - in this case, see if it's possible to compute an upper bound for the size.

Eventually, we should get rid of compiler-specific, non-standard stuff like this and preallocate buffers on the heap. However, that's a fairly significant change that's out of scope here, and needs to be applied elsewhere in the codebase too.

cmake/modules/FindGig.cmake
DomClark: I'm no lawyer, but I imagine something like this should be fine:
```cmake
# Copyright (c) <year> <name>
# Based on <file>, copyright (c) <name> <year>
#
# Redistribution and use is allowed according to the terms of the New BSD license.
# For details see the accompanying COPYING-CMAKE-SCRIPTS file.
```
Besides, if you, the author, are not sure about the copyright, some random person in the future wanting to use this in their project is going to have absolutely no clue.

plugins/GigPlayer/GigPlayer.cpp
PhysSong: There is another simple solution: move the loop body to another function to make sure `_alloca` isn't called in every iteration.

Rossmaxx:
Forgor to mention, it would be good to add libgig to the ci dependency list once vcpkg merges the port.

Rossmaxx:
After some testing, I found that libgig crashes on release builds but not on debug builds. I have such instances of crashes with other places too but might as well mention this here because i could get lmms to crash by opening a gig file on gig player using a release build.

DomClark:
> After some testing, I found that libgig crashes [...] Saw dom's comment on a line. Is it related?

Short of addressing the comment and seeing if that fixes it, there's no way to tell unless you can get a backtrace. If you need help with this, ask in `#dev-support` in Discord.

Rossmaxx:
happened to reproduce the crash and the prompt i'm getting is `noteEnd() triggered without a corresponding activate()!`

steps to reproduce: 
- drop a gig player on song editor and open a gig file with it. (it crashes here if built with release or relwithdebinfo)
- drop a few notes. 
- delete the track (crashes here if debug mode)

Rossmaxx:
> Forgot to mention, it would be good to add libgig to the ci dependency list once vcpkg merges the port.

Update: merged.

> happened to reproduce the crash again and the prompt i'm getting is `noteEnd() triggered without a corresponding activate()!`

Not related. Need to investigate the crash again.

Veratil:
> the prompt i'm getting is `noteEnd() triggered without a corresponding activate()!`

This is indeed unrelated, and a result of trying to activate too many notes at the same time afaict. The note start will queue and get dropped then the note end will process and trigger this.

superpaik:
> Inviting @superpaik for testing.

It builds and works ok. 
Tested on Win10 and MSVC 2019

What I did to build it:
- download the vcpkg libgig port from here: https://github.com/Veratil/vcpkg/tree/add-port-libgig
- download your code
- rebuild all

Rossmaxx:
Thanks for testing. 

> download the vcpkg libgig port from here: https://github.com/Veratil/vcpkg/tree/add-port-libgi

No need anymore, just use standard vcpkg, be sure to `git pull` from vcpkg. Libgig got merged today.

superpaik:
Retested with standard vcpkg, and everything Ok.

> Also,i noticed a crash while using libgig. Can you reproduce it?
Yes, it crashes when deleting the track. But I don't think it's related to gigplayer, because it crashes also if you do the same with the tripleOsc.

plugins/GigPlayer/GigPlayer.cpp
Rossmaxx: I found an even simpler solution for now (It's an ugly hack which should be replaced asap but for the current use case, didn't find anything simpler). 

We can add 
```cpp
#ifdef _MSC_VER
_freea(convertBuf)
#endif
```
Just before closing the loop. That way, it'll deallocate the `convertBuf`and avoid the memory leak. Tbh we should be rewriting this part but it's beyond the scope.

plugins/GigPlayer/GigPlayer.cpp
```diff
@@ -531,7 +539,11 @@ void GigInstrument::loadSample( GigSample& sample, sampleFrame* sampleData, f_cn
 	}
 
 	unsigned long allocationsize = samples * sample.sample->FrameSize;
+#ifndef _MSC_VER
 	int8_t buffer[allocationsize];
+#else
```
Rossmaxx, line 544: You can also add `_freea(buffer)` within the ifdef i posted in the other review comment.

Rossmaxx:
[libgig-comment.patch](https://github.com/LMMS/lmms/files/13266572/libgig-comment.patch)

I have a patch here which I did. You can apply this here if you are fine with this.

Veratil:
Closing in favor of #7162

PR #6602, opened 2023-01-04
Support per-note detuning and panning with Sf2 Player

DomClark:
See title. Panning only works with FluidSynth >= 2.0, unfortunately, since that's when the `GEN_CUSTOM_BALANCE` generator was introduced. There is a `GEN_PAN` generator that works with older versions, but the channel attenuation doesn't match that of LMMS's other native instruments.

I have introduced an `ArrayVector` class in this PR that is effectively a `std::vector` equivalent that stores its elements like `std::array` does. There are numerous third-party implementations out there, but I couldn't find one that was both maintained and didn't come as part of a massive library of things we don't need. (There's even a C++ standards proposal for one, but I don't believe it has been incorporated yet, so we probably won't see it until C++26 at the earliest.) I've wanted a container like this for a couple of purposes, and as the rest of this PR was quite small, I thought it would be a good place to introduce it. It allows vector-like operation without heap allocations, so is useful in real-time code.


Monospace-V:
I am wondering if older files with (non-working) detuning on sf2 player will play as intended? I believe previous behaviour was to play the note it happened to be at the start, and any 50 cents detuning rounded off.

image
(This would play an F.)

I have checked and found files where my SF2 player has pitch bend on the piano roll, possibly because I copied them over from a different instrument.
image
Is it possible to retain compatibility, or will all existing old files need to be manually modified for sounding like they did before?

YohananDiamond:
I'm not sure on how LMMS handles old version files, but maybe there could be a pop up that asks if you want one thing or the other (converting it to an accurate version of the pattern accordingly)

messmerd:
I think we should either bump the required FluidSynth version to 2.0 or later, or provide a warning in the CMake output when per-note panning is not supported.

The CMake output currently looks like:
```
* SoundFont2 player           : OK
```
But if we do not bump the FluidSynth version, it should read something like:
```
* SoundFont2 player           : OK (per-note panning enabled)

[...or...]

* SoundFont2 player           : OK (FluidSynth version < 2.0: Per-note panning disabled)
```

DomClark:
> I think we should either bump the required FluidSynth version to 2.0 or later, or provide a warning in the CMake output when per-note panning is not supported.

This sounds like a good idea. Linux, Mac, and MSVC builds use the FluidSynth version from their respective package managers, which is already at least 2.0 for Mac and MSVC. Linux builds are still on Ubuntu 18.04, which has 1.1.9, but when we upgrade to 20.04 (which we can now do), we'll get 2.1.1. MinGW builds are trickier, as these use a custom PPA which tobydox builds for us, and currently have version 1.1.8. I don't know whether we'll be able to switch to MSVC builds for Windows releases before we get to the RC stage for 1.3, or whether we should contact tobydox for an updated FluidSynth version.

> I am wondering if older files with (non-working) detuning on sf2 player will play as intended? [...] Is it possible to retain compatibility, or will all existing old files need to be manually modified for sounding like they did before?

It will certainly sound different from before. Which one is intended is a different question, and the answer may depend on the particular user and project. I'm inclined to avoid removing anything from the project automatically if the user might want it - it's a lot easier for them to remove it themselves than add it back. Plus, this depends on FluidSynth version, not just LMMS version, so sharing projects between identical LMMS versions could still sound inconsistent.

> Maybe there could be a pop up that asks if you want one thing or the other

I would rather not add user interaction to the project loading code - it would complicate command-line rendering for one thing, as well as adding some GUI dependency into the core, which is undesirable for technical reasons.

If removing unwanted automation does turn out to be tedious, I would instead suggest a context menu item or button within the song editor or piano roll to remove all per-note automation from a clip. Feedback on this idea, or better ideas, are welcome.

include/ArrayVector.h
```diff
@@ -0,0 +1,388 @@
...
+	{
+		assert(count <= N);
+		std::uninitialized_fill_n(begin(), count, value);
+	}
+
+	explicit ArrayVector(size_type count) noexcept(std::is_nothrow_default_constructible_v<T>) :
+		m_size{count}
+	{
+		assert(count <= N);
+		std::uninitialized_value_construct_n(begin(), count);
```
sakertooth, line 100: Wouldn't ``std::uninitialized_value_construct_n`` still allocate using ``new`` but just voidify the memory?

DomClark: It uses placement new, so there is no additional allocation involved.

plugins/Sf2Player/Sf2Player.cpp
```diff
@@ -683,10 +714,10 @@ void Sf2Instrument::playNote( NotePlayHandle * _n, sampleFrame * )
 		pluginData->midiNote = midiNote;
 		pluginData->lastPanning = 0;
 		pluginData->lastVelocity = _n->midiVelocity( baseVelocity );
-		pluginData->fluidVoice = nullptr;
 		pluginData->isNew = true;
 		pluginData->offset = _n->offset();
 		pluginData->noteOffSent = false;
+		pluginData->panning = std::clamp(PanningLeft, _n->getPanning(), PanningRight);
```
sakertooth, line 720: I believe ``_n->getPanning()`` should be the first argument?
```suggestion
		pluginData->panning = std::clamp(_n->getPanning(), PanningLeft, PanningRight);
```

tests/main.cpp
```diff
@@ -16,7 +16,7 @@ int main(int argc, char* argv[])
 	int failed = 0;
 	for (QTestSuite*& suite : QTestSuite::suites())
 	{
-		failed += QTest::qExec(suite, argc, argv);
+		if (QTest::qExec(suite, argc, argv) != 0) { ++failed; }
```
sakertooth, line 19: Probably out of scope?

plugins/Sf2Player/Sf2Player.cpp
DomClark: It should indeed. I think I can remove the clamp instead though - it looks like `Note` enforces the bound itself.

tests/main.cpp
DomClark: I can see that point of view, and can revert it if you want. However, I'm not sure it warrants its own PR. and this PR does at least touch the tests, so it's not entirely irrelevant.

I don't think I've mentioned the reason for the change, so I'll explain it here. The [documentation for `QTest::qExec`](https://doc.qt.io/qt-5/qtest.html#qExec) states "this function returns 0 if no tests failed, or a value other than 0 if one or more tests failed or in case of unhandled exceptions". As such, this is a more appropriate way to use it than the original, where the specific non-zero value was relevant. In practise, it seems to return the number of tests in the suite which failed, which is not useful when trying to count the number of failed suites, not tests. I made this fix when I got a nonsensical output while developing the tests included in this PR.

sakertooth: > In practise, it seems to return the number of tests in the suite which failed, which is not useful when trying to count the number of failed suites, not tests.

I do agree. It seems more realistic to count the number of suites that failed rather than the number of failed tests in each suite. Also, the documentation stating "this function returns 0 if no tests failed, or a value other than 0 if one or more tests failed or in case of unhandled exceptions" doesn't really guarantee that it will return a specific kind of number on failure (i.e., # of failed tests), just non-zero, and that should be the only thing we depend on. Feel free to keep the change, I don't really mind :+1:.

sakertooth (approved):

include/ArrayVector.h
```diff
@@ -0,0 +1,388 @@
...
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this program (see COPYING); if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA 02110-1301 USA.
+ *
+ */
+
+#ifndef ARRAY_VECTOR_H
+#define ARRAY_VECTOR_H
```
messmerd, line 26:
```suggestion
#ifndef LMMS_ARRAY_VECTOR_H
#define LMMS_ARRAY_VECTOR_H
```

include/ArrayVector.h
```diff
@@ -0,0 +1,388 @@
...
+	friend bool operator!=(const ArrayVector& l, const ArrayVector& r) { return !(l == r); }
+
+private:
+	alignas(T) std::byte m_data[std::max(N * sizeof(T), std::size_t{1})]; // Intentionally a raw array
+	size_type m_size = 0;
+};
+
+} // namespace lmms
+
+#endif // ARRAY_VECTOR_H
```
messmerd, line 388:
```suggestion
#endif // LMMS_ARRAY_VECTOR_H
```

messmerd (approved):
LGTM. No problems found while testing.

PR #5321, opened 2019-11-15
Show vst parameter string

ibuki2003:
VSTs gives us strings descript parameters.
This PR makes parameter knobs possible to show the patameter string instead of raw float value [0.0, 1.0]

PR #6603, opened 2023-01-05
Fix include guards and copyright statements

messmerd:
This PR corrects some minor issues with header files under `include/`
- Added namespace prefixes (`LMMS_` or `LMMS_GUI_`) to include guards of header files in the `include/` directory. This fixes a name conflict between the include guards of `include/Engine.h` and `plugins/ZynAddSubFx/zynaddsubfx/src/Nio/Engine.h` and possibly others. It will also prevent any future conflicts.
- The include guards of `Clip.h` and `ClipView.h` were still using their old "track content object" names from before the reorganization. This has been corrected. ([#5592](https://github.com/LMMS/lmms/issues/5592))
- Added missing copyright statement to `MidiCCRackView.h`.
- Added missing copyright statements to `CustomTextKnob.h` and `CustomTextKnob.cpp` according to author's comment in [#5321](https://github.com/LMMS/lmms/pull/5321#issuecomment-1371783964).
- Makes the number of newlines between copyright statement and include guard, between include guard and includes, and between closing include guard and EOF consistent for all headers


Veratil (commented):
LGTM

My only note is a few of the `#endif`'s lost their `// GUARD_H` comment. We don't really need them since they're EOF, but it doesn't hurt to have I think?

messmerd:
> My only note is a few of the `#endif`'s lost their `// GUARD_H` comment. We don't really need them since they're EOF, but it doesn't hurt to have I think?

There were only a few instances of those, so I just removed them for consistency with the rest of the headers.

messmerd:
Actually, I'll go ahead and add them to all the headers. It would be better to have them, and now's the best time to do it.

PhysSong:
Generally looks good.
I'm not sure if it's necessary, but I think it's a good chance to rename some include guards, ex. `LMMS_AUDIO_DEVICE_H` to `LMMS_AUDIO_DEVICE_H`. Historically, some header files used names in snake_case, but they were renamed to PascalCase.

include/aeffectx.h
```diff
@@ -22,9 +22,8 @@
  *
  */
 
-
-#ifndef AEFFECTX_H
-#define AEFFECTX_H
+#ifndef LMMS_AEFFECTX_H
```
DomClark, line 25: This one isn't in the LMMS namespace.

DomClark (approved):

Veratil:
Since @PhysSong has soft approval with comment, and @DomClark has approval. I'm merging this now.

PR #6606, opened 2023-01-07
Redesign ReverbSC and change its title.

tuckotb:
This pull request redesigns ReverbSC and changes its name from ReverbSC to Reverb (this is in favor of new algorithms, a dropdown can be implemented).

The background artwork was designed by @datkat21 in Figma.

In the image below the original ReverbSC is portrayed first, followed by the redesign, and lastly Dual Filter for comparison. The window is 144 by 144.

ReverbSC, Reverb, and Dual Filter


JGHFunRun:
I'd say wait on the name change until new algorithms are implemented, that said I do like the UI

Spekular:
I don't think it's a good idea to try to fit all (potential future) reverb algorithms into a single plugin. They're pretty likely to have different parameters, and trying to shoehorn support for that into one effect instead of having separate ones is unnecessarily complex.

actuallyexeon:
Yeah, I was thinking how I would add the new parameters and how I would change the artwork.png but I think it's just better to create a better algorithm and plugin instead.

superpaik:
I agree with the comment regarding the name. As for the artwork, I think that "IN" an "OUT" shouldn't be hardcode, you could use a label to do it (see the bitcrush effect) so they can be translated.

rdrpenguin04:
I'm not even sure what the SC means in the name; making a canonical "Reverb" plugin in LMMS seems like a very logical idea, especially given the way that Compressor has gone (combination Compressor and Limiter) and the way Cabinet already is (selectable Cabinet type rather than a larger number of plugins).

I personally agree with the rename, and I love the design :smile:

he29-net:
@rdrpenguin04 The 'SC' in ReverbSC refers to the author, Sean Costello. You may know him as the author of the "Valhalla" series of reverbs. I.e., this is not a simple generic reverb, but a specific (and very nice sounding) algorithm he developed. So dropping his initials from the name somehow does not seem fair to me. :)

Otherwise, the GUI redesign looks good.

As for making it the "default reverb plugin": it seems like a good idea, but all we have is a (very long) list of plugins; there is nothing that really says "if in doubt, use this one". What about, instead of relying on "simple names", we added a `recommended` flag for each plugin? Then we could select a recommended plugin for each category, add a '★' column to the effect list, and a simple sort on that column would bring all recommended plugins to the top. (Or, there is also the overkill option of having a full-blown rating system and sorting by the score... But that discussion is probably better left for its own issue, I'm getting distracted..)

rubiefawn:
I have some thoughts:

- There's an ongoing effort to replace raster assets with vector assets (see #7767, #7777), so if this PR is to move forward, `artwork.png` should be replaced with an equivalent `artwork.svg`.
- Like @superpaik said, the "IN" and "OUT" labels should not be hard-coded into the background image. These labels can be added to the in/out meters through code, which will allow them to be translated to other languages.
- Looking at the code, the in/out meters are actually faders (slider and all) and have "Gain" tooltips, but there doesn't seem to be any code that actually uses the values of these faders. This is confusing since a user might expect them to actually control input/output volume of the plugin, which they do not, and pre/post gain controls are also a feature that doesn't really make sense for reverb. What was the intent here?
- Furthermore, for some reason these faders are taken from the EQ plugin (see `CMakeLists.txt`) rather than using `Fader.h`. This seems odd to me.
- All but two of the new or modified lines in this PR with parentheses have whitespace padding inside the parentheses, which does not adhere to our current [coding conventions](https://github.com/LMMS/lmms/wiki/Coding-Conventions#braces) (e.g. `inFader->setHintText( tr( "Gain" ), "dBFS" )` should be `inFader->setHintText(tr("Gain"), "dBFS")`). Please make sure any lines you touch follow our coding conventions.
- It's been two years since the last update to this PR. Is there still interest?

rubiefawn:
If there is no further discussion before Oct 25, 2026 (one year after my previous comment, and more than three years after the most recent commit) I will close this PR.

PR #6607, opened 2023-01-07
Improve readability of text floats

michaelgregorius:
Improve the readability of text floats by removing the hard coded pixel font size. The widget is now composed of three `QLabels` which are used to display the title, text and pixmap. A small step to improve the problems described in #2510.

The adjusted version looks as follows. Please use the main menu as a reference for a "normal" text size:

https://user-images.githubusercontent.com/9293269/211168411-7fa46aa4-0701-4df5-ad7c-97215fa84699.mp4

## Technical details
Remove the methods `paintEvent` and `updateSize` because the widget is now updated automatically whenever the QLabels change state.

Merge both versions of `TextFloat::displayMessage` because one of them was only called by the other.

The default constructor of `TextFloat` now delegates to the "full" constructor. The latter is private because it is only used from within `displayMessage`.

The methods `setTitle`, `setText` and `setPixmap` show/hide their corresponding labels depending on whether there is text or a pixmap to show.

Adjust the class `Fader` so that the text float is shown to the right of the fader. Otherwise the new implementation would have covered the fader while it is being moved.

## Old version
For comparison, here's how the current implementation looks. The text is very small and hard to read.

https://user-images.githubusercontent.com/9293269/211168527-bd4e00ab-3c07-43ca-a19a-762d20af3ca6.mp4

### DIFF
--- a/data/themes/classic/style.css
+++ b/data/themes/classic/style.css
@@ -70,7 +70,7 @@ QToolTip {
 	color: #4afd85;
 }
 
-lmms--gui--TextFloat {
+lmms--gui--TextFloat, lmms--gui--SimpleTextFloat {
 	border-radius: 4px;
 	background: qlineargradient(spread:reflect, x1:0.5, y1:0.5, x2:0.5, y2:0, stop:0 rgba(0, 0, 0, 255), stop:1 rgba(50, 50, 50, 220));
 	opacity: 175;
--- a/data/themes/default/style.css
+++ b/data/themes/default/style.css
@@ -100,7 +100,7 @@ QToolTip {
 	color: #d1d8e4;
 }
 
-lmms--gui--TextFloat {
+lmms--gui--TextFloat, lmms--gui--SimpleTextFloat {
 	background: #040506;
 	color: #d1d8e4;
 }
--- a/include/Fader.h
+++ b/include/Fader.h
@@ -55,10 +55,11 @@
 
 #include "AutomatableModelView.h"
 
+
 namespace lmms::gui
 {
 
-class TextFloat;
+class SimpleTextFloat;
 
 
 class LMMS_EXPORT Fader : public QWidget, public FloatModelView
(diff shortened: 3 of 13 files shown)


superpaik:
I get that idea, but I don't see why the frame has to be so wide. I don't think it improves readability. Also, it is not applied to "float help text" (like the one that appears over icons), so we have an inconsistency between text floats.

michaelgregorius:
Hi @superpaik,

thanks for the feedback! The `TextFloat` in this branch is implemented using a horizontal layout that displays the icon on the left and the title and text on the right. The title and the text are part of another vertical layout. Both layouts currently have their default margin of 9.

The `TextFloat` is used in two contexts:
* As a temporary message that displays an icon, title and text. Usually shown at the bottom left of the main window.
* As a pseudo tool tip which only shows current values as a text. This text is shown close to the manipulated widget as long as an action occurs.

Personally I like the appearance in the temporary message case. The message is meant to be noticed by the user so it should have a certain size.

In my opinion the underlying problem is that the `TextFloat` is also used for the pseudo tool tip case in the first place. Perhaps it would be better to implement the pseudo tool tip cases with a separate widget that's meant specifically for these cases. Ideally it would inherit from `QToolTip` so that it receives the same styling as the tool tips that are shown when hovering over the buttons. That way they'd always look the same and be in sync. But I think this should be something for another issue/pull request.

superpaik:
Hi, @michaelgregorius.

I agree with you on temporary messages.
As for the second context, I agree that a different widget could be the best solution, but maybe a solution would be to apply different styles depending on whether the textfloat has icon, and/or title. I don't know if it's possible code-wise.

michaelgregorius:
I have now introduced the class `SimpleTextFloat` which simply shows a text and acts as a pseudo tool tip. The current state looks as follows:

https://user-images.githubusercontent.com/9293269/211633303-38b526e5-97d8-47d9-924a-a510867448a2.mp4

michaelgregorius:
@LMMS/developers, can I merge this?

superpaik:
I think that you should check your code to adapt it to coding conventions before merging it.

michaelgregorius:
> I think that you should check your code to adapt it to coding conventions before merging it.

Done!

PR #6612, opened 2023-01-16
Fix recent files menu on old KDE in a different way

### LINKED ISSUE #6587: Crash with segmentation fault if kwidgetsaddons 5.101.0 is installed
### Bug Summary

If kwidgetsaddons 5.101.0 (the newest release) is installed, lmms won't launch because of a segmentation fault during preparing ui. If using gdb to skip [this line](https://github.com/LMMS/lmms/blob/17aa7dbef5f4b1dbf8c7c335ef6f2a7759600f21/src/gui/MainWindow.cpp#L107), the segmentation fault won't happen.

It seems that the [related workaround code](https://github.com/LMMS/lmms/blob/17aa7dbef5f4b1dbf8c7c335ef6f2a7759600f21/src/gui/MainWindow.cpp#L81-L95) is outdated and recent updates of kwidgetsaddons break it.

Platform: Arch Linux

DM&WM: sddm+i3wm

#### Steps to reproduce

On Arch Linux, install kwidgetsaddons-5.101 and lmms using pacman, and then launch lmms.

#### Expected behavior

lmms will start normally.

#### Actual behavior

After the splash screen, lmms crashes with a segmentation fault.

#### Affected LMMS versions

Reproduced on v1.2.2 and v1.3.0-alpha.1

#### Logs

Backtrace, though not very helpful:

<details>
<pre>
Thread 1 "lmms" received signal SIGSEGV, Segmentation fault.
0x00007ffff688cadb in QCoreApplicationPrivate::sendThroughApplicationEventFilters (this=this@entry=0x555555956bb0, receiver=receiver@entry=0x5555560acc70, event=event@entry=0x7fffffffd540) at kernel/qcoreapplication.cpp:1172
1172	            if (obj->eventFilter(receiver, event))
(gdb) bt
#0  0x00007ffff688cadb in QCoreApplicationPrivate::sendThroughApplicationEventFilters(QObject*, QEvent*)
    (this=this@entry=0x555555956bb0, receiver=receiver@entry=0x5555560acc70, event=event@entry=0x7fffffffd540) at kernel/qcoreapplication.cpp:1172
#1  0x00007ffff7978b5a in QApplicationPrivate::notify_helper(QObject*, QEvent*) (this=0x555555956bb0, receiver=0x5555560acc70, e=0x7fffffffd540)
    at kernel/qapplication.cpp:3608
#2  0x00007ffff688cf98 in QCoreApplication::notifyInternal2(QObject*, QEvent*) (receiver=0x5555560acc70, event=0x7fffffffd540)
    at kernel/qcoreapplication.cpp:1064
#3  0x00007ffff79a53fa in QWidgetPrivate::setStyle_helper(QStyle*, bool) (this=<optimized out>, newStyle=0x0, propagate=<optimized out>)
    at kernel/qwidget.cpp:2659
#4  0x00007ffff79b0405 in QWidget::setParent(QWidget*, QFlags<Qt::WindowType>) (this=0x5555560acc70, parent=<optimized out>, f=...)
    at kernel/qwidget.cpp:10501
#5  0x00007ffff799d19a in QWidgetPrivate::init(QWidget*, QFlags<Qt::WindowType>) (this=0x555555d41c00, parentWidget=<optimized out>, f=...)
    at kernel/qwidget.cpp:1046
#6  0x0000555555759513 in MainWindow::MainWindow() (this=this@entry=0x555555ce6540) at /usr/include/qt/QtCore/qflags.h:120
#7  0x000055555574ccd9 in GuiApplication::GuiApplication() (this=this@entry=0x555555a499e0)
    at /usr/src/debug/lmms/lmms-1.2.2/src/gui/GuiApplication.cpp:116
#8  0x00005555556482ca in main(int, char**) (argc=<optimized out>, argv=<optimized out>) at /usr/src/debug/lmms/lmms-1.2.2/src/core/main.cpp:759
</pre>
</details>

PhysSong:
Fixes #6587.
This PR reverts #3872 and applies an alternative fix for #3741. In the previous approach, `KAcceleratorManager::setNoAccel` was called with the `MainWindow` instance. This, however, leads to a crash in certain cases as reported by #6587.
KXmlGui library registers a startup function for `QCoreApplication` which may install a `KCheckAccelerators` global event filter, using `KCheckAcceleratorsInitializer`. The event filter is supposed to be installed only if the application directly links to KXmlGui, but it was always installed prior to KXmlGui 5.55.0.

The new approach has two advantages:
- For buggy versions of KXmlGui, it removes the root cause of the issue completely
- For newer KXmlGui versions, it's effectively doing nothing.

As far as know, CentOS 7, Ubuntu 18.04 and Debian 10 ships with buggy versions of the library. Once all of them gets unsupported, we may entirely remove the workaround.

### DIFF
--- a/src/gui/MainApplication.cpp
+++ b/src/gui/MainApplication.cpp
@@ -39,6 +39,19 @@ MainApplication::MainApplication(int& argc, char** argv) :
 	QApplication(argc, argv),
 	m_queuedFile()
 {
+#if !defined(LMMS_BUILD_WIN32) && !defined(LMMS_BUILD_APPLE) && !defined(LMMS_BUILD_HAIKU)
+	// Work around a bug of KXmlGui < 5.55
+	// which breaks the recent files menu
+	// https://bugs.kde.org/show_bug.cgi?id=337491
+	for (auto child : children())
+	{
+		if (child->inherits("KCheckAcceleratorsInitializer"))
+		{
+			delete child;
+		}
+	}
+#endif
+
 #if defined(LMMS_BUILD_WIN32)
 	installNativeEventFilter(this);
 #endif
--- a/src/gui/MainWindow.cpp
+++ b/src/gui/MainWindow.cpp
@@ -33,7 +33,6 @@
 #include <QMenuBar>
 #include <QMessageBox>
 #include <QShortcut>
-#include <QLibrary>
 #include <QSplitter>
 
 #include "AboutDialog.h"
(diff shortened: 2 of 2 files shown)


Iron-Squid:
wheeeeeeeee

PR #6616, opened 2023-01-18
Add control+click selection in Piano Roll editor

martinvanzijl:
This makes Control+Click select work the same as Shift+Click in the Piano Roll editor.

This is part of #4602, namely: 

> Piano Roll --> Multiple selection of midi notes by pressing "Ctrl" inside the piano roll.

### DIFF
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -2225,7 +2225,7 @@ void PianoRoll::computeSelectedNotes(bool shift)
 	{
 		for( Note *note : m_midiClip->notes() )
 		{
-			// make a new selection unless they're holding shift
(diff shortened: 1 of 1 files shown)


zonkmachine:
Thanks @martinvanzijl and sorry for the late reply!
This works but I think what was intended was an analog to the selection in Song Editor where you can, in draw mode, press <Ctrl> and then add track segments to your selection by left clicking. I Piano Roll when you use <Ctrl> you can only use the selection tool (string/drag) to select multiple notes and not add/remove notes to the selection by clicking on individual notes.

martinvanzijl:
@zonkmachine Thanks for your comment, that's OK. I would still like to see this implemented since it improves usability.

In any case, I recommend updating the original ticket to clarify what they meant (track segments or individual MIDI notes).

PR #6629, opened 2023-02-03
Make song length only update periodically

LostRobotMusic:
Currently, when an LMMS project has a large number of clips, moving them around or copy+pasting them causes quite a massive amount of lag.  This lag multiplies when moving/pasting multiple clips, sometimes even causing LMMS to completely freeze for a full minute or more!

The reason for this is a tad embarrassing.  Every time a clip is moved or created, the song length is updated.  When the song length is updated, LMMS checks the length of every single track individually.  When a track's length is checked, it checks the length and location of every single clip it contains.

So, as an example, if your song has 1000 clips in it, and you want to move the song over to the right by 1 measure, LMMS will loop through _one million clips_, solely to calculate the length of the song.  This is obviously ridiculous.

This PR aims to fix this issue by changing as little of the code base as possible.  The `updateLength` function has been changed to simply set a boolean to true to let LMMS know that the song length should be checked again.  Then, approximately once per frame, LMMS checks the state of this boolean, and then updates the song length accordingly.  In the very rare cases where the song length needs to be updated immediately (e.g. upon project export), the new `recalcLength` function is used instead.

- [ ] TODO: Use MainWindow's `periodicUpdate` instead of a separate QTimer

### DIFF
--- a/src/core/Song.cpp
+++ b/src/core/Song.cpp
@@ -91,6 +91,7 @@ Song::Song() :
 	m_isCancelled( false ),
 	m_playMode( Mode_None ),
 	m_length( 0 ),
+	m_shouldUpdateLength( true ),
 	m_midiClipToPlay( nullptr ),
 	m_loopMidiClip( false ),
 	m_elapsedTicks( 0 ),
@@ -115,6 +116,10 @@ Song::Song() :
 			this, SLOT(masterVolumeChanged()), Qt::DirectConnection );
 /*	connect( &m_masterPitchModel, SIGNAL(dataChanged()),
 			this, SLOT(masterPitchChanged()));*/
+			
+	QTimer *timer = new QTimer(this);
+	connect(timer, &QTimer::timeout, [this](){Song::checkUpdateLength();});
+	timer->start(1000.f / 60.f);
 
 	qRegisterMetaType<lmms::Note>( "lmms::Note" );
 	setType( SongContainer );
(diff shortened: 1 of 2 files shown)


superpaik:
Tested on W10-64bits.
Works Ok. Less lags and visual "artifacts". Not only moving/pasting clips, but also inserting and deleting bars (shift+ctrl+insert/del)

src/core/Song.cpp
```diff
@@ -115,6 +116,10 @@ Song::Song() :
 			this, SLOT(masterVolumeChanged()), Qt::DirectConnection );
 /*	connect( &m_masterPitchModel, SIGNAL(dataChanged()),
 			this, SLOT(masterPitchChanged()));*/
+			
+	QTimer *timer = new QTimer(this);
+	connect(timer, &QTimer::timeout, [this](){Song::checkUpdateLength();});
```
allejok96, line 121: What about

```cpp
connect(this, &Song::updateLength, this, &Song::recalcLength, Qt::QueuedConnection);
```

similar to how `update()` and `repaint()` works, instead of the timer?

LostRobotMusic: I'm pretty sure this would run almost identically to how things are without the PR.  I don't think this throws out any redundant calls to updateLength.

PhysSong:
@LostRobotMusic Can this be closed in favor of #6741?

PhysSong:
Closing in favor of #6741.

PR #6637, opened 2023-02-08
Fix comparing int with bool

anytizer:


src/core/PatternClip.cpp
```diff
@@ -78,7 +78,7 @@ void PatternClip::loadSettings(const QDomElement& element)
 		movePosition( element.attribute( "pos" ).toInt() );
 	}
 	changeLength( element.attribute( "len" ).toInt() );
-	if( element.attribute( "muted" ).toInt() != isMuted() )
+	if( (bool)element.attribute( "muted" ).toInt() != isMuted() )
```
PhysSong, line 81: `static_cast` is preferred to a C-style cast.

src/tracks/MidiClip.cpp
```diff
@@ -424,7 +424,7 @@ void MidiClip::loadSettings( const QDomElement & _this )
 	{
 		movePosition( _this.attribute( "pos" ).toInt() );
 	}
-	if( _this.attribute( "muted" ).toInt() != isMuted() )
+	if( (bool)_this.attribute( "muted" ).toInt() != isMuted() )
```
PhysSong, line 427: Here as well.

PR #6154, opened 2021-09-09
Add ability to move Controllers in ControllerRack

ejaaskel:
This commit adds UI changes, signals, and handlers required for moving controllers up and down in a controller rack. The feature was requested in the issue #5773

The implementation is similar to the one used with EffectRacks. Quite small change, good for a first issue to solve (as it was mentioned in the issue itself)

- Added move up and move down buttons to context menu
- Added signals that'll message the requested move
- Added functions in the ControllerRackView that'll handle those signals

PR #6648, opened 2023-02-17
Draggable effects

LostRobotMusic:
This PR allows users to rearrange effects in the mixer by dragging them up or down, instead of needing to go into the context menu over and over and over again.  The effect's opacity is decreased while it is being dragged as a visual indication of which effect is currently being dragged.

### DIFF
--- a/include/EffectView.h
+++ b/include/EffectView.h
@@ -61,6 +61,11 @@ public:
 	}
 
 	static constexpr int DEFAULT_WIDTH = 215;
+	static constexpr int DEFAULT_HEIGHT = 60;
+	
+	void mouseMoveEvent(QMouseEvent* event) override;
+	void mousePressEvent(QMouseEvent* event) override;
+	void mouseReleaseEvent(QMouseEvent* event) override;
 
 public slots:
 	void editControls();
@@ -90,6 +95,8 @@ private:
 	Knob * m_gate;
 	QMdiSubWindow * m_subWindow;
 	EffectControlDialog * m_controlView;
+	
+	bool m_dragging;
 
 } ;
 
--- a/include/EffectView.h
+++ b/include/EffectView.h
@@ -98,7 +98,7 @@ private:
 	EffectControlDialog * m_controlView;
 	
 	bool m_dragging;
-	QGraphicsOpacityEffect * m_opacityEffect;
+	QGraphicsOpacityEffect* m_opacityEffect;
 
 } ;
 
--- a/include/EffectView.h
+++ b/include/EffectView.h
@@ -30,6 +30,7 @@
 #include "PluginView.h"
 #include "Effect.h"
 
+class QGraphicsOpacityEffect;
 class QGroupBox;
 class QLabel;
 class QPushButton;
@@ -97,6 +98,7 @@ private:
 	EffectControlDialog * m_controlView;
 	
 	bool m_dragging;
+	QGraphicsOpacityEffect * m_opacityEffect;
 
 } ;
 
--- a/src/gui/EffectView.cpp
+++ b/src/gui/EffectView.cpp
@@ -239,16 +239,14 @@ void EffectView::mouseReleaseEvent(QMouseEvent* event)
 
 void EffectView::mouseMoveEvent(QMouseEvent* event)
 {
-	if (m_dragging)
+	if (!m_dragging) {return;}
+	if (event->pos().y() < 0)
 	{
-		if (event->pos().y() < 0)
-		{
-			moveUp();
-		}
-		else if (event->pos().y() > EffectView::DEFAULT_HEIGHT)
-		{
-			moveDown();
-		}
+		moveUp();
+	}
+	else if (event->pos().y() > EffectView::DEFAULT_HEIGHT)
+	{
+		moveDown();
 	}
 }
 
--- a/src/gui/EffectView.cpp
+++ b/src/gui/EffectView.cpp
@@ -48,10 +48,10 @@ EffectView::EffectView( Effect * _model, QWidget * _parent ) :
 	PluginView( _model, _parent ),
 	m_bg( embed::getIconPixmap( "effect_plugin" ) ),
 	m_subWindow( nullptr ),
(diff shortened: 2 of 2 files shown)


src/gui/EffectView.cpp
```diff
@@ -47,9 +48,10 @@ EffectView::EffectView( Effect * _model, QWidget * _parent ) :
 	PluginView( _model, _parent ),
 	m_bg( embed::getIconPixmap( "effect_plugin" ) ),
 	m_subWindow( nullptr ),
-	m_controlView( nullptr )
+	m_controlView( nullptr ),
+	m_dragging( false )
```
Veratil, line 52:
```suggestion
	m_controlView(nullptr),
	m_dragging(false)
```

src/gui/EffectView.cpp
```diff
@@ -47,9 +48,10 @@ EffectView::EffectView( Effect * _model, QWidget * _parent ) :
...
 {
-	setFixedSize( EffectView::DEFAULT_WIDTH, 60 );
+	setFixedSize( EffectView::DEFAULT_WIDTH, EffectView::DEFAULT_HEIGHT );
```
Veratil, line 54:
```suggestion
	setFixedSize(EffectView::DEFAULT_WIDTH, EffectView::DEFAULT_HEIGHT);
```

Veratil (approved):
Just some missed code style. LGTM though.

superpaik:
It works ok on Win10-64b.
Some thoughts:
- the opacity effect it's not clear to me (default theme, not classic). Maybe decrease the opacity (make it more dim) or "disable" the controls. Or change the icon to a doble arrow (like in song editor) so it's clear that the drag action is engaged.
- it is possible to drag the effect over the limits of the Effects Chain visible area, which is great, but he content of the area is not "scrolled" accordingly so it's kind of wierd. I know it's QT an maybe it's not possible to update the visible area when dragging its content.
- There is another PR somehow related to this (#6154) Maybe it's worth considering merging on master both of them.

LostRobotMusic:
@superpaik I've changed the code to use a QGraphicsOpacityEffect instead of manually setting the QPainter opacity, so that the child widgets will have the opacity change applied to them as well.  That should fix the issue.

superpaik:
@LostRobotMusic Yes, much better now, thanks! But I also think that changing the cursor would be visually useful and to keep consistency with other drag&drops (I think you can "clone" the code from TrackView)

src/gui/EffectView.cpp
```diff
@@ -208,6 +214,42 @@ void EffectView::contextMenuEvent( QContextMenuEvent * )
...
+	{
+		if (event->pos().y() < 0)
+		{
+			moveUp();
+		}
+		else if (event->pos().y() > EffectView::DEFAULT_HEIGHT)
+		{
+			moveDown();
+		}
+	}
```
Veratil, line 249: We can reduce indention by inverting the if guard. :)
```suggestion
	if (!m_dragging) { return; }
	if (event->pos().y() < 0)
	{
		moveUp();
	}
	else if (event->pos().y() > EffectView::DEFAULT_HEIGHT)
	{
		moveDown();
	}
```

include/EffectView.h
```diff
@@ -90,6 +96,9 @@ public slots:
 	Knob * m_gate;
 	QMdiSubWindow * m_subWindow;
 	EffectControlDialog * m_controlView;
+	
+	bool m_dragging;
+	QGraphicsOpacityEffect * m_opacityEffect;
```
Veratil, line 101: I remember we discussed this somewhere, but I don't think we ever put it in the style guide itself. It might be part of the clang format check.
```suggestion
	QGraphicsOpacityEffect* m_opacityEffect;
```

src/gui/EffectView.cpp
```diff
@@ -116,7 +119,10 @@ EffectView::EffectView( Effect * _model, QWidget * _parent ) :
 			m_subWindow->hide();
 		}
 	}
-
+	
+	m_opacityEffect = new QGraphicsOpacityEffect(this);
```
Veratil, line 123: Does Qt `delete` this for us, or do we need one in the destructor?

sakertooth: Qt should handle ``m_opacityEffect``'s deletion, because its parent is the ``EffectView``.

sakertooth (approved):

messmerd:
I agree with superpaik that the cursor should change while dragging. Other than that, LGTM. I've tested it and there don't seem to be any issues.

LostRobotMusic:
The cursor now updates when dragging effects.

src/gui/EffectView.cpp
```diff
@@ -208,6 +214,43 @@ void EffectView::contextMenuEvent( QContextMenuEvent * )
...
 
 
+void EffectView::mousePressEvent(QMouseEvent* event)
+{
+	if (event->button() == Qt::LeftButton)
+	{
+		m_dragging = true;
+		m_opacityEffect->setOpacity(0.3);
+		QCursor c( Qt::SizeVerCursor );
+		QApplication::setOverrideCursor( c );
```
messmerd, line 224:
```suggestion
		QCursor c(Qt::SizeVerCursor);
		QApplication::setOverrideCursor(c);
```
Small style fix

src/gui/EffectView.cpp
```diff
@@ -208,6 +214,43 @@ void EffectView::contextMenuEvent( QContextMenuEvent * )
...
+		m_dragging = false;
+		m_opacityEffect->setOpacity(1);
+		QApplication::restoreOverrideCursor();
+		update();
+	}
+}
+
+void EffectView::mouseMoveEvent(QMouseEvent* event)
+{
+	if (!m_dragging) {return;}
```
messmerd, line 242:
```suggestion
	if (!m_dragging) { return; }
```
Might be better to have spaces for padding in the block

messmerd (approved):

superpaik:
LGTM.
Thanks!

PR #6654, opened 2023-02-24
Fix EQ effect warnings

superpaik:
Basically corrects two (math) errors that caused a lot of "QPainterPath::lineTo Invalid coordinates" warnings on the console.
- the first one is related to when there are no EQ bands active causing the warning when loading the EQ or changing some parameters.
- the second one is related to log10f(0) which cause a pole error, and the function is called every time an audio buffer is processed
I don't know why there are so many changes in the code with the first commit "Fix Warning". Just added an if around some existing code.


superpaik:
I close this PR so I'd try to make the changes in code more clear (sorry, this is my first PR, just learning ;-)

PR #6655, opened 2023-02-24
Fix EQ effect warnings

superpaik:
Basically corrects two (math) errors that cause a lot of "QPainterPath::lineTo Invalid coordinates" warnings on the console.

- [ ] the first one is related to when there are no EQ bands active causing a div by 0 and generating the warnings when loading the EQ or changing some parameters.

- [ ] the second one is related to log10f(0) which cause a pole error, returning and invalid x pixel position (the function is called every time audio buffer is processed)

I don't know why git tracks so many changes in the code with the first commit. Just added an if around some existing code.

### DIFF
--- a/plugins/Eq/EqCurve.cpp
+++ b/plugins/Eq/EqCurve.cpp
@@ -742,17 +742,17 @@ void EqCurve::paint( QPainter *painter, const QStyleOptionGraphicsItem *option, 
 		}
 		//compute a QPainterPath
 		m_curve = QPainterPath();
-		//only draw the EQ curve if there are activeHandles
+		//only draw the EQ curve if there are any activeHandles
 		if (activeHandles != 0)
 		{
-			for ( int x = 0; x < m_width ; x++ )
+			for (int x = 0; x < m_width; x++)
 			{
-				mainCurve[x] = ( ( mainCurve[x] / activeHandles ) ) - ( m_heigth/2 );
-				if ( x==0 )
+				mainCurve[x] = ((mainCurve[x] / activeHandles)) - (m_heigth / 2);
+				if (x == 0)
 				{
-					m_curve.moveTo( x, mainCurve[x] );
+					m_curve.moveTo(x, mainCurve[x]);
 				}
-				m_curve.lineTo( x, mainCurve[x] );
+				m_curve.lineTo(x, mainCurve[x]);
 			}
 		}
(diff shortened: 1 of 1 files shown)


JohannesLorenz:
> I don't know why git tracks so many changes in the code with the first commit. Just added an if around some existing code.

About that question: There is a cogwheel in the diff view. When you click on it, you can tell github to hide whitespace. If you reload, the diff looks much better :grin:

superpaik:
> > I don't know why git tracks so many changes in the code with the first commit. Just added an if around some existing code.
> 
> About that question: There is a cogwheel in the diff view. When you click on it, you can tell github to hide whitespace. If you reload, the diff looks much better 😁

Oh, didn't know that! Much better, now! Thanks!

plugins/Eq/EqCurve.cpp
```diff
@@ -65,6 +65,7 @@ QRectF EqHandle::boundingRect() const
 
 float EqHandle::freqToXPixel( float freq , int w )
 {
+	if (freq == 0) { return 0.0f; }
```
messmerd, line 68:
```suggestion
	if (typeInfo<float>::isEqual(freq, 0.0f)) { return 0.0f; }
```
This might be slightly better for comparing two floats for equality.

messmerd (approved):

JohannesLorenz:
What is the state here? Should we do a 2nd developer review? Or should it be tested? Or just be merged like it is?

superpaik:
> What is the state here? Should we do a 2nd developer review? Or should it be tested? Or just be merged like it is?

It's a simple PR, but I think it would be good to have a second revision in order to merge.

JohannesLorenz (approved):
Code is OK (Sorry for my late review!)

JohannesLorenz:
@superpaik This can be merged. Do you want it committed as multiple commits, or as one? Do you want to formulate commit messages?

superpaik:
> Code is OK (Sorry for my late review!)

Thanks. No worries.

> @superpaik This can be merged. Do you want it committed as multiple commits, or as one? Do you want to formulate commit messages?

I can't merge code to master, I don't have the needed role. Can you do that?
As for the commit message something like "Fix EQ invalid coordinate messages"

zonkmachine:
@superpaik Your commit messages were good so I just used them and hit the 'squash' button.

PR #6664, opened 2023-03-09
Change zoom in SongEditor to a Slider Zoom

superpaik:
- Added new styles and icons for horizontal slider (derived from vertical sliders)
- Adapt TimeLineWidget to work with new continous values of zoom, instead of fixed values.
- Change zoom from fixed values to a "continous" zoom with a Slider. 
Now Zooming can be done with:
- [ ] Slider: smooth zoom in and out, from the first visible beat.
- [ ] Ctrl + MouseWheel: zoom in and out, from the beat where the cursor is at. Increments/decrements of zoom value are done acording to "pixels per bar" and visible bars are realigned acordingly. 
- [ ] Crtl + Shift + MouseWheel: "slow" zoom in and out, from the beat where the cursor is at. Useful for detailed zooming.
- [ ] Crtl + 0: reset zoom to init value

SliderZoom


src/gui/editors/SongEditor.cpp
```diff
@@ -531,35 +535,64 @@ void SongEditor::keyPressEvent( QKeyEvent * ke )
...
+
+			} while (i++ < PredefinedZoom.size());
+		}
 
 		if(we->angleDelta().y() > 0)
 		{
-			z++;
+			(we->modifiers() & Qt::ShiftModifier)
+				? z = PredefinedZoom[(pos < (PredefinedZoom.size() - 1)) ? pos + 1 : pos]
+				: z = static_cast<int>(ceil((floor(z / 100.0f * DEFAULT_PIXELS_PER_BAR) + 1) * 100.0f / DEFAULT_PIXELS_PER_BAR));
```
messmerd, line 564:
```suggestion
			z = (we->modifiers() & Qt::ShiftModifier)
				? PredefinedZoom[std::min(pos + 1, PredefinedZoom.size() - 1)]
				: static_cast<int>(std::ceil((std::floor(z / 100.0f * DEFAULT_PIXELS_PER_BAR) + 1) * 100.0f / DEFAULT_PIXELS_PER_BAR));
```
The ternary operator should only be used to selected one of two values (see [coding conventions](https://github.com/LMMS/lmms/wiki/Coding-Conventions#ternary-operator)). `std::min` can simplify selecting the array index.

src/gui/editors/SongEditor.cpp
```diff
@@ -531,35 +535,64 @@ void SongEditor::keyPressEvent( QKeyEvent * ke )
...
 		}
 		else if(we->angleDelta().y() < 0)
 		{
-			z--;
+			(we->modifiers() & Qt::ShiftModifier) 
+				? z = PredefinedZoom[pos > 0 ? pos - 1 : pos]
+				: z = static_cast<int>(floor((ceil(z / 100.0f * DEFAULT_PIXELS_PER_BAR) - 1) * 100.0f / DEFAULT_PIXELS_PER_BAR));
```
messmerd, line 570:
```suggestion
			z = (we->modifiers() & Qt::ShiftModifier) 
				? PredefinedZoom[std::max(pos - 1, 0)]
				: static_cast<int>(std::floor((std::ceil(z / 100.0f * DEFAULT_PIXELS_PER_BAR) - 1) * 100.0f / DEFAULT_PIXELS_PER_BAR));
```
Similar to the review comment above. This still might not be the cleanest way of writing this logic, but it's a bit better.

src/gui/editors/SongEditor.cpp
```diff
@@ -531,35 +535,64 @@ void SongEditor::keyPressEvent( QKeyEvent * ke )
...
+		}
+		
+		z = qBound(ZOOM_MIN, z, ZOOM_MAX);
```
messmerd, line 573:
```suggestion
		z = std::clamp(z, ZOOM_MIN, ZOOM_MAX);
```
Try to avoid using Qt where possible

src/gui/editors/TimeLineWidget.cpp
```diff
@@ -273,9 +273,11 @@ void TimeLineWidget::paintEvent( QPaintEvent * )
 	for( int i = 0; x + i * m_ppb < width(); ++i )
 	{
 		++barNumber;
+		// round m_ppb to the nearest power of two
+		float ppbP2 = pow(2, round(log2(m_ppb)));
 		if( ( barNumber - 1 ) %
 			qMax( 1, qRound( 1.0f / 3.0f *
-				TimePos::ticksPerBar() / m_ppb ) ) == 0 )
+				TimePos::ticksPerBar() / ppbP2 ) ) == 0 )
```
messmerd, line 280: There are some code style issues here which could be fixed

messmerd (commented):
This is mostly a preliminary code style review. I have not thoroughly looked over the logic yet or tested the code.

src/gui/editors/TimeLineWidget.cpp
superpaik: I was only modifying the name of the variable and didn't change original code. Being that line of code part of a bigger instruction, should I modify the whole instruction (lines 278, 279 and 280) or only the one that is involved with the new code? I think that should be the right line of action.
As for the name of the new variable ppbP2 (stands for "pixels per bar at power of 2") I'm not sure what is the best coding convection here: pppP2 or pPBP2 (I just followed the existing variable name)

superpaik:
@messmerd thanks for the review! Sorry for the mistakes of coding conventions. I went through the code several times, but you know...

superpaik:
It seems that we are using different QT versions for building Windows/Mac and Linux/MinGW [(see building error)](https://github.com/LMMS/lmms/actions/runs/4377717620/jobs/7665024942#step:7:2313)
The function globalPosition() replaces the deprecated globalPos() which is the one that Linux/MinGW build is expecting.
Should I use a macro to check QT version in that case or it's a problem of the build?

src/gui/editors/TimeLineWidget.cpp
superpaik: I'd go for modifying the whole instruction and let the variable name as it is.

src/gui/editors/TimeLineWidget.cpp
```diff
@@ -273,9 +273,9 @@ void TimeLineWidget::paintEvent( QPaintEvent * )
 	for( int i = 0; x + i * m_ppb < width(); ++i )
 	{
 		++barNumber;
-		if( ( barNumber - 1 ) %
-			qMax( 1, qRound( 1.0f / 3.0f *
-				TimePos::ticksPerBar() / m_ppb ) ) == 0 )
+		// round m_ppb to the nearest power of two
+		float ppbP2 = pow(2, round(log2(m_ppb)));
+		if ((barNumber - 1) % qMax(1, qRound(1.0f / 3.0f * TimePos::ticksPerBar() / ppbP2)) == 0)
```
messmerd, line 278:
```suggestion
		const float ppbP2 = std::pow(2, std::lround(std::log2(m_ppb)));
		if ((barNumber - 1) % std::max(1l, std::lround((1.0f / 3.0f) * TimePos::ticksPerBar() / ppbP2)) == 0)
```
This removes Qt functions and fixes a compile error on my machine. You'll need to add `#include <cmath>` at the top of this file though. Maybe you could fix the problems with the [include order](https://github.com/LMMS/lmms/wiki/Coding-Conventions#include-order) while you're at it.

messmerd:
A few issues I noticed while testing:
- Right-clicking on the zoom slider provides a context menu with some unnecessary options such as automation and connecting to a controller. And when I click "Edit song-global automation", LMMS crashes.
- If I click and hold to the left or right of the slider bar, the zoom will adjust in that direction, but it's incredibly slow. I think the slider bar should immediately jump to that position instead.
- Maybe the zoom should be logarithmic, because it's hard to get the zoom just where I want it when the bar is all the way to the left (zoomed out), but the level of control seems to be too fine when the bar is all the way to the right (zoomed in).

Veratil:
> * And when I click "Edit song-global automation", LMMS crashes.

We should find where this context menu is adding it and removing it. This isn't something that should be automated.

messmerd:
A few more issues (mostly minor):
- Ctrl + MouseWheel seems to work, but if the mouse hovers over the Zoom percentage label, it loses focus and you can't zoom
- Ctrl + MouseWheel lets you zoom in/out even when hovering over the instruments on the left
- If you use Ctrl + MouseWheel to zoom, it creates the Zoom percentage label next to the mouse. If you then continue to hold down Ctrl and interact with clips by right-clicking, left-clicking, or double-clicking, the label will still remain. If you click on another window so that the Song-Editor loses focus, the label will remain until you use Ctrl+MouseWheel to zoom again or click on the label.
- Crtl + Alt + MouseWheel doesn't work

messmerd:
Thanks, it works better now.

> I've not been able to reproduce the first one. Maybe it's related to the platform?

I'm on Linux, so maybe it's related to that. It's not a very serious issue for me though

> I think that all these issues should be resolved in another PR. What do you think?

Yeah, that sounds best to me

It looks like if I use Ctrl+Drag on the slider, I can still cause the crash to occur. Maybe this could be fixed by adding a `disableMouseDragEvents()` method on `AutomatableSlider` or something.

When zooming in, it can move the horizontal scrollbar at the bottom of the song editor to the right and make some clips go out of sight. This is a little annoying imo, especially when the the scrollbar is all the way to the left and the clip on the first measure of the song goes out of sight. I'm not sure if there's any good solution to this though.

Is there a reason the smallest zoom value is 13% instead of something like 10%, 15%, etc?

messmerd:
> Sorry. I've been able to reproduce. Is that happening on the other sliders, like master volume or master pitch?

Not the crash. You can't reproduce the Ctrl+Click-and-drag of the zoom slider on your end? Dragging it to an automation track causes the crash, but we shouldn't be able to Ctrl+Drag the zoom slider in the first place since it isn't automatable.

> Unfortunatelly there is no disableMouseDragEvent on AutomatableSlider.

I meant that maybe we should add a method like that to `AutomatableSlider`. But since we're trying to remove all capabilities for automation from `AutomatableSlider`, it would be a better idea to just use a plain `QSlider` if possible, or if not, create a new wrapper class called `Slider` similar to `AutomatableSlider` but without automation.

> I can't see a neat solution but moving the cursor position (the zoom is based on current cursor position) which I think is a worse sideeffect

To be clear, I'm not totally against the current behavior. I just think it may be worth fiddling with it to see if there is any alternative behavior we like better. What would happen if it zoomed _as if_ the mouse was always located all the way to the left? Or if it zoomed as if the mouse was located wherever the playhead is?

superpaik:
> > Sorry. I've been able to reproduce. Is that happening on the other sliders, like master volume or master pitch?
> 
> Not the crash. You can't reproduce the Ctrl+Click-and-drag of the zoom slider on your end? Dragging it to an automation track causes the crash, but we shouldn't be able to Ctrl+Drag the zoom slider in the first place since it isn't automatable.

Sorry, I wasn't dragging it to an automation track. Yes, it crashes. Unfortunately, it happens with other controls as well (try to automate the current zoom or the snapping size combobox) It happens because in model.cpp we try to get the parentModel()->fullDisplayName(), and some of the controls' parents do not have that property or something.  The string is not null and is not empty, but when accessing to it, it generates an access violation. I haven't been able to correct it. I'll create an issue for that to see if someone can help with that.
 
> > Unfortunatelly there is no disableMouseDragEvent on AutomatableSlider.
> 
> I meant that maybe we should add a method like that to `AutomatableSlider`. But since we're trying to remove all capabilities for automation from `AutomatableSlider`, it would be a better idea to just use a plain `QSlider` if possible, or if not, create a new wrapper class called `Slider` similar to `AutomatableSlider` but without automation.

I found that the drag and drop events are handled by the model and not the widget. To correct that is as easy as not calling the model when crtl+click+drag. We can create a new variable inside AutomatableSlider like m_allowDragActions set by default to true, and when false just not call the model (inside mousePressEvent) Alternatively as you said, I can create another class, but it would be almost identical. I don't know what's the best approach, but I opted for the first one, just for simplicity of it. Just tell me if you think this is not a good solution and I try to create a new class.

> > I can't see a neat solution but moving the cursor position (the zoom is based on current cursor position) which I think is a worse side-effect
> 
> To be clear, I'm not totally against the current behavior. I just think it may be worth fiddling with it to see if there is any alternative behavior we like better. What would happen if it zoomed _as if_ the mouse was always located all the way to the left? Or if it zoomed as if the mouse was located wherever the playhead is?

The first option is what happens if you use the slider, the zoom is done by the first beat on the left. The second option I think that would have the same "problem" as with mouse position. I personally think is not a good option, because playhead tends to be on beat 1, and that can be confusing to the user. I prefer the idea of zooming around the mouse position. Part of the problem comes because we have an insane amount of zooming in, down to 1 pixel per beat. So I did and alternative approach which is zoom in just to the value where all the song is visible. That is something similar to the scroll bar, and how ableton handle zooming in.


Log Slider:
I've been in circles trying to get the slider (a new class Slider) to work properly with a log scale to no avail, and it's because the slider has an "inside" model that works only lineal but it's linked to the model which is forced to work in a log scale, so you get unexpected results (like the slider move faster than the mouse, or when set a value, the slider goes to a wrong place) Anyways, I've found a workaround to that that is creating an alternative logModel, not in the slider class but inside the SongEditor class, so internally the slider continues working in a linear fashion but externally -controlling the zoom and visually- it works as if it where logarithmic. That's why I decided to push-it in this PR.

superpaik:
After some thoughts, I've decided to change how zoomMin is calculated. Before was like 105% of the song (fit the song in the songEditor width) now it's 200% (so, double the size of the song). That is because we don't have "living" scrolling when clips are copied/moved to the right, making it difficult to copy/move all the song at once, what I believe to be a usual operation.

allejok96:
This is great! Some things I noticed while testing:

- Ctrl+scroll does not move the slider
- Scrolling the slider moves it but zoom level remains the same
- The variable minimum would make more sense if there was only ctrl+scroll, but now when there is a slider I expect to be able to drag it all the way down.
- Ctrl+scroll should change percent, not pixels. Scrolling above 1000% is painfully slow, and below 50% crazy fast.
- *Personal opinion:* holding shift should zoom slower, just to match behavior of other things like quantization (more accurate when holding shift)
- Maybe the TextFloat could be hidden when mouse hovers it to solve the problem with it getting in the way?
- To make this a good experience, song editor's scroll should not be limited to bars (other PR)

include/SongEditor.h
```diff
@@ -197,6 +205,13 @@ protected slots:
 
 	void updateSnapLabel();
 
+	void showZoomingSliderFloat();
```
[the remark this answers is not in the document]
allejok96: After some more testing I think the zoom float when scrolling is just annoying and not needed. We don't have that in PianoRoll or Audacity for example.

But I still think the slots should be in `SongEditor` next to their buddies.

src/gui/editors/SongEditor.cpp
```diff
@@ -531,41 +541,88 @@ void SongEditor::keyPressEvent( QKeyEvent * ke )
...
+	if ((we->modifiers() & Qt::ControlModifier) && (position(we).x() > m_trackHeadWidth))
 	{
-		int z = m_zoomingModel->value();
+		// calculate the min zoom based on song length and song editor size
+		int minZoom = calculateMinZoom();
+				
+		int z = m_zoomingLogModel->value();
+		if (we->angleDelta().y() > 0)
+			z = static_cast<int>(
+				std::ceil((std::floor(z / 100.0f * DEFAULT_PIXELS_PER_BAR) + 1) * 100.0f / DEFAULT_PIXELS_PER_BAR));
```
allejok96, line 554: z += std::max(1.0, z * 0.05)

or something like that has a nicer feeling IMO.

src/gui/editors/SongEditor.cpp
```diff
@@ -531,41 +541,88 @@ void SongEditor::keyPressEvent( QKeyEvent * ke )
...
-		int x = position(we).x() - m_trackHeadWidth;
-		// bar based on the mouse x-position where the scroll wheel was used
-		int bar = x / pixelsPerBar();
-		// what would be the bar in the new zoom level on the very same mouse x
-		int newBar = x / DEFAULT_PIXELS_PER_BAR / m_zoomLevels[z];
-		// scroll so the bar "selected" by the mouse x doesn't move on the screen
-		m_leftRightScroll->setValue(m_leftRightScroll->value() + bar - newBar);
+			int x = position(we).x() - m_trackHeadWidth;
+			// save the pixels up to the cursor next bar when starting zooming
+			if (!m_zvsStatus->isVisible()) { m_iniBar = static_cast<int>(x / pixelsPerBar()); }
```
allejok96, line 595: When zooming out and in quickly, you use `m_iniBar` to zoom back into the bar we were on before, instead of zooming in on the bar where the mouse is now.

This is a neat feature. But using TextFloat as a timer is confusing. And if you wait just a second, the feature breaks. IMO if the scroll speed is ok (so you don't accidentally zoom out too much) this is not a problem that needs to be fixed. Most other editors audio/video editors has this behavior.

src/gui/editors/SongEditor.cpp
```diff
@@ -839,15 +896,45 @@ void SongEditor::updatePositionLine()
...
 	realignTracks();
 	updateRubberband();
 	m_timeLine->setSnapSize(getSnapSize());
 	
-	emit zoomingValueChanged( m_zoomLevels[m_zoomingModel->value()] );
+	emit zoomingValueChanged(m_zoomingLogModel->value() / 100.0f);
+}
+
+
+int SongEditor::scaleFunction(int val, bool exp)
```
allejok96, line 911: This should be split in two separate functions. They don't have to be part of the `SongEditor` class. Since they rely on the `ZOOM_` constants the naming should give us a hint about that, like `linearZoom()` and `expZoom()`

src/gui/editors/SongEditor.cpp
```diff
@@ -1125,6 +1228,51 @@ void SongEditorWindow::lostFocus()
...
+	}
+	else
+	{
+		QWidget::keyPressEvent(ke);
+	}
+}
+
+void SongEditorWindow::keyReleaseEvent(QKeyEvent* ke)
+{
+	if (m_editor->m_zvsStatus->isVisible()) { m_editor->m_zvsStatus->hide(); }
```
allejok96, line 1272: You could clear `m_iniBar` here, but I still think it's a better idea to just skip that

allejok96 (changes requested):

src/gui/editors/TimeLineWidget.cpp
superpaik: ok

include/SongEditor.h
```diff
@@ -69,7 +70,8 @@ class SongEditor : public TrackContainerView
 	void saveSettings( QDomDocument& doc, QDomElement& element ) override;
 	void loadSettings( const QDomElement& element ) override;
 
-	ComboBoxModel *zoomingModel() const;
+	IntModel* zoomingLinearModelSlider() const;
```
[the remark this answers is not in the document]
superpaik: I was going in circles to set the slider work with a log scale and finally come across this solution of two models, when having just an integer variable holding the log value it's simpler. Thanks!

include/SongEditor.h
```diff
@@ -123,6 +125,11 @@ private slots:
 	int trackIndexFromSelectionPoint(int yPos);
 	int indexOfTrackView(const TrackView* tv);
 
+	int scaleFunction(int x, bool exp);
+	int calculateMinZoom();
+
+	bool zoomingFloatVisible;
```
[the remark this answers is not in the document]
superpaik: deleted

include/SongEditor.h
```diff
@@ -161,6 +167,8 @@ private slots:
 	int m_currentZoomingValue;
 	int m_trackHeadWidth;
 	bool m_selectRegion;
+	int m_iniBar;
+	int m_songEditorViewSize;
```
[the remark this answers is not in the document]
superpaik: You're right!

src/gui/editors/SongEditor.cpp
superpaik: I agree, two functions are better. I'll do that. 
But what do you mean about "They don't have to be part to the SongEditor class"?

Veratil: > But what do you mean about "They don't have to be part to the SongEditor class"?

The two functions don't need to be in the `SongEditor` class. They are independent from the class itself and can be reused elsewhere.

allejok96: > They are independent from the class itself

This means you can create a regular old-school function and put it at the bottom of `SongEditor.cpp` like so:

```cpp
int scaleFunction(int val, bool exp)
{
    ...
}
```

> can be reused elsewhere

Well, since they are kind of adapted to Song Editor I would just let them be local (unless you find somewhere to reuse them). This means you **don't** declare the functions in `SongEditor.h`.

src/gui/editors/SongEditor.cpp
```diff
@@ -42,7 +45,9 @@
 #include "DeprecationHelper.h"
 #include "embed.h"
 #include "GuiApplication.h"
+#include "qslider.h"
```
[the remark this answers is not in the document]
superpaik: You're right!

include/SongEditor.h
```diff
@@ -123,6 +125,11 @@ private slots:
 	int trackIndexFromSelectionPoint(int yPos);
 	int indexOfTrackView(const TrackView* tv);
 
+	int scaleFunction(int x, bool exp);
+	int calculateMinZoom();
+
+	bool zoomingFloatVisible;
+	TextFloat * m_zvsStatus;
```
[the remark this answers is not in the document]
superpaik: ok. changed the name to m_zoomStatus

include/SongEditor.h
```diff
@@ -197,6 +205,13 @@ protected slots:
 
 	void updateSnapLabel();
 
+	void showZoomingSliderFloat();
+	void updateZoomingSliderFloat(int new_val);
+	void hideZoomingSliderFloat();
+
+	void keyPressEvent(QKeyEvent* ke) override;
```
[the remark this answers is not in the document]
superpaik: Ok. Done.

include/AutomatableSlider.h
```diff
@@ -45,6 +45,16 @@ class AutomatableSlider : public QSlider, public IntModelView
 		return( m_showStatus );
 	}
 
+	void setAllowDragOptions(bool opt)
```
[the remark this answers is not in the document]
superpaik: I agree with you that this is trying to workaround a bug that should be addressed in another PR.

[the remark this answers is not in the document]
superpaik: I revert Automatableslider

include/SongEditor.h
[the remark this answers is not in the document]
superpaik: Ok. I agree. I remove zoomfloat for mouse scrolling.
But I don't get what you say about these slots. They are for the m_zoomingSlider slider that is in SongEditorWindow. Don't know what you mean but "their buddies"

src/gui/editors/SongEditor.cpp
superpaik: I removed TextFloat and reset m_iniBar in keyReleased as you mentions before.

src/gui/editors/SongEditor.cpp
superpaik: discussed in a previous commentary

src/gui/editors/SongEditor.cpp
superpaik: Not sure about how to do that without declaring the function in SongEditor.h  Can you point me to another class that has something similar?

Veratil: Put it at the top of the file and drop the `SongEditor::` part. :)

src/gui/clips/ClipView.cpp
```diff
@@ -314,10 +314,25 @@ void ClipView::updateLength()
 	}
 	else
 	{
-		setFixedWidth(
-		static_cast<int>( m_clip->length() * pixelsPerBar() /
-					TimePos::ticksPerBar() ) + 1 /*+
-						BORDER_WIDTH * 2-1*/ );
+		float w = 0;
```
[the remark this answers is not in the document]
superpaik: I was trying to solve those cases, when the clip was made bigger or smaller using ALT and zoom-in a lot (in some cases the clip even disappeared with original code). I tried different combinations and that was the best I could find. But it seems it's not correct in all cases ¿?

src/gui/editors/SongEditor.cpp
superpaik: I was following "..and put it at the bottom of ..." Now it works! Thanks!

include/SongEditor.h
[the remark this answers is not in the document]
allejok96: There are similar instances of `TextFloat` and slots to update them in `SongEditor` for the master pitch and volume. I suggest you place your slots and `TextFloat` there just to keep things organized.

(Alternatively you ditch the zoom float completely, since the numerical value doesn't really matter... What-you-see-is-what-you-get. But I don't know if other people would agree with me on that)

src/gui/editors/SongEditor.cpp
allejok96: Ah, the order matters, sorry :eyes:

include/SongEditor.h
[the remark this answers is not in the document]
superpaik: Those sliders are "weird" because they don't belong to SongEditor but MainWindow. I don't know why they are there. Anyway, I think that we shouldn't mix views with controllers, so I think it's better to keep them on SongEditorView. Don't you think?

src/gui/clips/ClipView.cpp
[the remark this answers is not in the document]
superpaik: Finally I've come with this solution that seems to work just fine, being 3 the minimum width needed to paint a clip (I don't know if there is a variable for this value)
		setFixedWidth(std::max(static_cast<int>(m_clip->length() * pixelsPerBar() / TimePos::ticksPerBar() + 1), 3));

src/gui/editors/SongEditor.cpp
```diff
@@ -57,14 +62,21 @@
...
+namespace
+{
+constexpr int ZOOM_MIN = 15;
+constexpr int ZOOM_MAX = 1600;
+constexpr int ZOOM_INI = 100;
 
-const QVector<float> SongEditor::m_zoomLevels =
-		{ 0.125f, 0.25f, 0.5f, 1.0f, 2.0f, 4.0f, 8.0f, 16.0f };
+// predefined zoom values used for Shift+Ctrl+mouseWheel
+constexpr auto PredefinedZoom = std::array{15, 20, 30, 40, 50, 60, 80, 100, 150, 200, 300, 400, 600, 800, 1000, 1200, 1400, 1600};
```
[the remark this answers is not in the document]
superpaik: Ok. I get rid of predefinedValues, and change it to be 10% with Ctrl and 1% with Ctrl+Shift.

src/gui/editors/SongEditor.cpp
superpaik: Unfortunately, I think it's has to be more complicated than that because zoom is in pixels and sometimes z*0.05 returns a value that doesn't increase the zoom at all.

src/gui/editors/SongEditor.cpp
```diff
@@ -57,14 +62,21 @@
 namespace lmms::gui
 {
 
+namespace
+{
+constexpr int ZOOM_MIN = 15;
+constexpr int ZOOM_MAX = 1600;
+constexpr int ZOOM_INI = 100;
```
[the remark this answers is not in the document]
superpaik: It uses "old" zoom scale. If changed, then I have to multiply by 100 in some other places (to control the slider, snapsize, rubberband, etc.) In balance, I can't see it to be less complicated. I made a quick change in the code and keeps crashing, so it need more not obvious changes.

src/gui/editors/SongEditor.cpp
superpaik: Done.

include/SongEditor.h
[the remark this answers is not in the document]
allejok96: You're absolutely right. Just because it was done wrong before doesn't mean we have to continue doing it wrong.

allejok96:
[I made some more suggestions](https://github.com/superpaik/lmms/pull/1) on how to simplify the zoom calculations

src/gui/editors/SongEditor.cpp
```diff
@@ -75,14 +87,15 @@ SongEditor::SongEditor( Song * song ) :
...
-	m_currentZoomingValue(m_zoomingModel->value()),
+	m_rubberbandPixelsPerBar(DEFAULT_PIXELS_PER_BAR),
 	m_trackHeadWidth(ConfigManager::inst()->value("ui", "compacttrackbuttons").toInt()==1
 					 ? DEFAULT_SETTINGS_WIDGET_WIDTH_COMPACT + TRACK_OP_WIDTH_COMPACT
 					 : DEFAULT_SETTINGS_WIDGET_WIDTH + TRACK_OP_WIDTH),
 	m_selectRegion(false)
 {
-	m_zoomingModel->setParent(this);
 	m_snappingModel->setParent(this);
+	m_zoomingModel->setParent(this);
```
allejok96, line 97: You can move this back up one line

src/gui/editors/SongEditor.cpp
```diff
@@ -170,9 +186,9 @@ SongEditor::SongEditor( Song * song ) :
 	connect( m_masterVolumeSlider, SIGNAL(sliderReleased()), this,
 			SLOT(hideMasterVolumeFloat()));
 
-	m_mvsStatus = new TextFloat;
-	m_mvsStatus->setTitle( tr( "Master volume" ) );
-	m_mvsStatus->setPixmap( embed::getIconPixmap( "master_volume" ) );
+	m_mvsStatus	= new TextFloat;
+	m_mvsStatus->setTitle(tr("Master volume"));
+	m_mvsStatus->setPixmap(embed::getIconPixmap("master_volume"));
```
allejok96, line 191: This is not a show-stopper (I want the code to look nice too) ... But in the future try to no edit lines when you don't actually need change them, because changes may introduce merge conflicts for other people, so try to keep the PR to the point.

I usually run `git gui` before committing and that allows me to see where I have added empty lines or changed whitespace etc for no good reason, and I can right click on those lines and select "revert line".

When doing follow-up commits to a PR like you are doing now, that :arrow_up: becomes harder to do. Then manual editing is perhaps your quickest option.

src/gui/widgets/AutomatableSlider.cpp
```diff
@@ -74,7 +74,7 @@ void AutomatableSlider::mousePressEvent( QMouseEvent * _me )
 	}
 	else
 	{
-		IntModelView::mousePressEvent( _me );
+		IntModelView::mousePressEvent(_me);
```
allejok96, line 77: Changes to `AutomatableSlider.h` and `AutomatableSlider.cpp` could be reverted completely

allejok96 (approved):
I think it looks good now, there's just a few unnecessary changes that could be reverted before merging, just to clean things up. But we need a few more people to test it out.

include/DeprecationHelper.h
```diff
@@ -64,6 +64,21 @@ inline QPoint position(QWheelEvent *wheelEvent)
 #endif
 }
 
+/**
+ * @brief globalPosition is a backwards-compatible adapter for
+ * QWheelEvent::globalPosition and globalPos functions.
+ * @param wheelEvent
+ * @Returns the position of the point in this event on the screen or virtual desktop.
+ */
+inline QPointF globalPosition(QWheelEvent* wheelEvent)
```
allejok96, line 73: This seems to be unused

src/gui/editors/SongEditor.cpp
superpaik: I try to. I have not knowingly changed it. I'll try to be more careful in the future 😄

src/gui/editors/SongEditor.cpp
superpaik: ok

src/gui/widgets/AutomatableSlider.cpp
superpaik: ok. reverted

superpaik:
@allejok96 Have you seen the two comments I made on your PR?

allejok96:
I'll take a break from this PR now in case you wonder, but you've done a good job @superpaik !

superpaik:
> I see, go ahead. But write that in a comment so it's more clear what the problem is.

Ok. I pushed new commit to do that.

> I'll take a break from this PR now in case you wonder, but you've done a good job @superpaik !

Thank you so much for your collaboration. You made the PR much more clear and simpler code-wise

src/gui/editors/SongEditor.cpp
```diff
@@ -170,9 +186,9 @@ SongEditor::SongEditor( Song * song ) :
 	connect( m_masterVolumeSlider, SIGNAL(sliderReleased()), this,
 			SLOT(hideMasterVolumeFloat()));
 
-	m_mvsStatus = new TextFloat;
-	m_mvsStatus->setTitle( tr( "Master volume" ) );
-	m_mvsStatus->setPixmap( embed::getIconPixmap( "master_volume" ) );
+	m_mvsStatus	= new TextFloat;
```
messmerd, line 189:
```suggestion
	m_mvsStatus = new TextFloat;
```
style

src/gui/editors/SongEditor.cpp
```diff
@@ -240,33 +256,23 @@ SongEditor::SongEditor( Song * song ) :
...
-			m_zoomingModel->findText( "100%" ) );
-	connect( m_zoomingModel, SIGNAL(dataChanged()),
-					this, SLOT(zoomingChanged()));
-	connect( m_zoomingModel, SIGNAL(dataChanged()),
-					m_positionLine, SLOT(update()));
-
-	//Set up snapping model, 2^i
-	for ( int i = 3; i >= -4; i-- )
+	// Set up snapping model
+	for (float bars: SNAP_SIZES)
```
messmerd, line 263:
```suggestion
	for (float bars : SNAP_SIZES)
```
style

src/gui/editors/SongEditor.cpp
```diff
@@ -240,33 +256,23 @@ SongEditor::SongEditor( Song * song ) :
...
 		}
-		else if ( i == 0 )
+		else if (bars == 1.0f)
 		{
 			m_snappingModel->addItem( "1 Bar" );
 		}
 		else
 		{
-			m_snappingModel->addItem( QString( "1/%1 Bar" ).arg( 1 << (-i) ) );
+			m_snappingModel->addItem(QString("1/%1 Bar").arg(1/bars));
```
messmerd, line 275:
```suggestion
			m_snappingModel->addItem(QString("1/%1 Bar").arg(1 / bars));
```
style

src/gui/editors/SongEditor.cpp
```diff
@@ -291,42 +297,40 @@ void SongEditor::loadSettings( const QDomElement& element )
...
-	int val = -m_snappingModel->value() + 3;
+	float snapSize = SNAP_SIZES[m_snappingModel->value()];
+
 	// If proportional snap is on, we snap to finer values when zoomed in
 	if (m_proportionalSnap)
 	{
-		val = val - m_zoomingModel->value() + 3;
+		// Finds the closest available snap size
+		const float optimalSize = snapSize * DEFAULT_PIXELS_PER_BAR / pixelsPerBar();
+		return *std::min_element(SNAP_SIZES.begin(), SNAP_SIZES.end(), [&](float a, float b)
```
messmerd, line 310:
```suggestion
		return *std::min_element(SNAP_SIZES.begin(), SNAP_SIZES.end(), [optimalSize](float a, float b)
```
Can capture by value instead of reference

src/gui/editors/SongEditor.cpp
```diff
@@ -837,17 +832,49 @@ void SongEditor::updatePositionLine()
...
 
+//! Convert zoom slider's value to bar width in pixels
+int SongEditor::calculatePixelsPerBar() const
+{
+	// What we need to raise 2 by to get MIN_PIXELS_PER_BAR and MAX_PIXELS_PER_BAR
+	double minExp = log2(MIN_PIXELS_PER_BAR);
+	double maxExp = log2(MAX_PIXELS_PER_BAR);
+	double exponent = m_zoomingModel->value() / double(ZOOM_STEPS) * (maxExp - minExp) + minExp;
+
+	double ppb = pow(2, exponent);
```
messmerd, line 843:
```suggestion
	static const double minExp = std::log2(MIN_PIXELS_PER_BAR);
	static const double maxExp = std::log2(MAX_PIXELS_PER_BAR);
	double exponent = m_zoomingModel->value() / static_cast<double>(ZOOM_STEPS) * (maxExp - minExp) + minExp;

	double ppb = std::pow(2, exponent);
```
Can use `static const` local variables because they only need to be calculated once.
Using `static_cast` instead of a functional-style cast.
Added `std::` prefix to math functions.

src/gui/editors/SongEditor.cpp
```diff
@@ -837,17 +832,49 @@ void SongEditor::updatePositionLine()
...
+
+
+
+//! Convert bar width in pixels to zoom slider value
+int SongEditor::calculateZoomSliderValue(int pixelsPerBar) const
+{
+	// What we need to raise 2 by to get MIN_PIXELS_PER_BAR and MAX_PIXELS_PER_BAR
+	double minExp = log2(MIN_PIXELS_PER_BAR);
+	double maxExp = log2(MAX_PIXELS_PER_BAR);
+	double exponent = log2(pixelsPerBar);
```
messmerd, line 857:
```suggestion
	static const double minExp = std::log2(MIN_PIXELS_PER_BAR);
	static const double maxExp = std::log2(MAX_PIXELS_PER_BAR);
	double exponent = std::log2(pixelsPerBar);
```
Can use `static const` local variables because they only need to be calculated once.
Added `std::` prefix to math functions.

src/gui/editors/TimeLineWidget.cpp
```diff
@@ -22,23 +22,29 @@
...
 #include "GuiApplication.h"
+#include "NStateButton.h"
 #include "TextFloat.h"
 
 namespace lmms::gui
 {
 
+namespace
+{
+	static const int MIN_BAR_LABEL_DISTANCE = 35;
```
messmerd, line 45:
```suggestion
	constexpr int MIN_BAR_LABEL_DISTANCE = 35;
```
`static` isn't needed within an anonymous namespace, and it can be made `constexpr` instead of just `const`.

messmerd (commented):
A few minor suggestions

messmerd:
I did a little bit of testing just now, since I hadn't tested any of the changes you've made over the past month.
It's looking pretty good and I'm almost ready to approve it.

A few notes:
- Is Ctrl+Shift+Scroll supposed to zoom slower now instead of faster? If so, maybe you could update the PR description.
- The Ctrl+Drag-to-automation-track crash still exists, but it also exists in master, so one of us could probably fix it in a later PR. We'd just need to create an issue for it if there isn't one already.
- When I use Ctrl+Alt+Scroll, it always zooms out regardless of scroll direction. I'm not sure why it does that. The PR description doesn't say anything about this key combination, and it's not a big deal, but I figured I'd mention it.

superpaik:
> * Is Ctrl+Shift+Scroll supposed to zoom slower now instead of faster? If so, maybe you could update the PR description.

Yes. I changed it according to @allejok96 comments to match behavior with other actions. PR description updated.

> * The Ctrl+Drag-to-automation-track crash still exists, but it also exists in master, so one of us could probably fix it in a later PR. We'd just need to create an issue for it if there isn't one already.

I'll create an issue.

> * When I use Ctrl+Alt+Scroll, it always zooms out regardless of scroll direction. I'm not sure why it does that. The PR description doesn't say anything about this key combination, and it's not a big deal, but I figured I'd mention it

Apparently it is a bug (or hidden behavior) of QT. When Alt is pressed, it mimics horizontal wheel movement. https://bugreports.qt.io/browse/QTBUG-91556 
I introduced a work-around with this suggestion https://bugreports.qt.io/browse/QTCREATORBUG-25612

messmerd:
The Alt button workaround works for me.

messmerd (approved):
I think this PR is ready to be completed now. Nice work!

src/gui/editors/SongEditor.cpp
```diff
@@ -837,17 +833,49 @@ void SongEditor::updatePositionLine()
...
+int SongEditor::calculatePixelsPerBar() const
+{
+	// What we need to raise 2 by to get MIN_PIXELS_PER_BAR and MAX_PIXELS_PER_BAR
+	static const double minExp = std::log2(MIN_PIXELS_PER_BAR);
+	static const double maxExp = std::log2(MAX_PIXELS_PER_BAR);
+	double exponent = m_zoomingModel->value() / static_cast<double>(ZOOM_STEPS) * (maxExp - minExp) + minExp;
+
+	double ppb = std::pow(2, exponent);
+
+	return static_cast<int>(std::round(ppb));
```
PhysSong, line 846: - You can replace `std::pow(2, exponent)` with `std::exp2(exponent)`. `log`-`exp` pair is also possible.
- If you're caching `minExp` and `maxExp`, why don't you calculate `1 / static_cast<double>(ZOOM_STEPS) * (maxExp - minExp)` as well?

src/gui/editors/SongEditor.cpp
```diff
@@ -291,42 +297,40 @@ void SongEditor::loadSettings( const QDomElement& element )
...
 {
-	// 1 Bar is the third value in the snapping dropdown
-	int val = -m_snappingModel->value() + 3;
+	float snapSize = SNAP_SIZES[m_snappingModel->value()];
+
 	// If proportional snap is on, we snap to finer values when zoomed in
 	if (m_proportionalSnap)
 	{
-		val = val - m_zoomingModel->value() + 3;
+		// Finds the closest available snap size
```
PhysSong, line 308: - Previously, it was possible to use finer or larger snap sizes by either zooming in with small snap size or zooming out with large snap size, but it's not possible now.
- I wonder why you have chosen nearest instead of rounding up/down.

superpaik: > * Previously, it was possible to use finer or larger snap sizes by either zooming in with small snap size or zooming out with large snap size, but it's not possible now.

Sorry, I think I don't understand. Do you mean that when ProportionalSnap is engaged the value should change while zooming in and out? Is the behaviour you expect any different as what you see in the gif? (I added values 1/32 and 1/64 to snapping values, because before it ended in 1/16, maybe that has something to do with your concern) 
SnapingZoom

>     * I wonder why you have chosen nearest instead of rounding up/down.
I have to say that this code was proposed by @allejok96, so I don't have the context as to why he chose that option. But comparing with master the behavior is not so different.

src/gui/editors/SongEditor.cpp
superpaik: Ok. I think that I've done what you suggested.

src/gui/editors/SongEditor.cpp
PhysSong: > Sorry, I think I don't understand. Do you mean that when ProportionalSnap is engaged the value should change while zooming in and out?

I mean, on the current `master`, I can make the snap size to 64 bars by setting to 12.5%+8Bars with proportional snap enabled, or 1/64 bars by 1600%+1/4Bar or 400%+1/16Bar.

allejok96: > I wonder why you have chosen nearest instead of rounding up/down.

The values we are trying to round to are increasing exponentially, so we can't just `round()`. @superpaik had figured it out earlier using `pow()`, but I removed it due to segfault in my brain (and because it wouldn't support snap sizes of 1/3, which the new design has opened the possibility to). If there is a c++ function that rounds to the closest value in a list of values then that would be the best solution.

PhysSong: > The values we are trying to round to are increasing exponentially, so we can't just `round()`

To clarify, I didn't mean `round()`, but minimum not less than/maximum not more than.

superpaik: I added a new variable PROPORTIONAL_SNAP_SIZES that hold bigger and smaller values than SNAP_SIZES, so the end result is similar to master (see gif, above this PR, below master). Is not exactly, but I can't see which one is better and why. To me, as long as they are consistent, both are ok.
SnapingZoom

superpaik:
Is there anything missing on my part in this PR? I don't think I have anything left to do.

PhysSong (approved):

RiedleroD:
bit late, ik, but would it be possible to set the zoom level directly via double-click on the slider, like almost all other controls in LMMS? I don't know if I'm going to remember Ctrl+0 for reset tbh.

superpaik:
> bit late, ik, but would it be possible to set the zoom level directly via popup on double-click on the slider, like almost all other controls in LMMS? I don't know if I'm going to remember Ctrl+0 for reset tbh.

I'm planning on adding some zoom actions (for zoom-to-selection, zoom-to-loop, zoom-to-fit, etc.) I'll have a look at your suggestion, it makes sense.

PR #6680, opened 2023-03-20
Fix FreeBoy CPU time bug

### LINKED ISSUE #6649: Freeboy crashes LMMS
Attachment: [freeboy-crash.zip](https://github.com/LMMS/lmms/files/10771605/freeboy-crash.zip)

If a second note plays earlier than first note finishes, freeboy crashes LMMS.

lmms: -/lmms/plugins/FreeBoy/game-music-emu/gme/Gb_Apu.cpp:131: void Gb_Apu::run_until(blip_time_t): Assertion `end_time >= last_time' failed.

messmerd:
Fixes #6649

The bug occurs because all `Gb_Apu_Buffer` objects, which are each stored in a `NotePlayHandle`'s `m_pluginData` member, share the same `m_time` CPU time value from `FreeBoyInstrument`. So if there are two different `NotePlayHandle` objects in use at the same time by the same `FreeBoyInstrument`, such as when transitioning to a new note, `FreeBoyInstrument::playNote(...)` called with `NotePlayHandle` **A** may reset `m_time` to zero prior to calling `papu->end_frame(FRAME_LENGTH)`. Then when `FreeBoyInstrument::playNote(...)` is called again, this time with `NotePlayHandle` **B**, it may call a `Gb_Apu` method which still remembers the `Gb_Apu::last_time` value as whatever `m_time` was before the call with `NotePlayHandle` **A** set it to zero. This `last_time` value is now invalid, which causes the assertion in `Gb_Apu::run_until(...)` to fail. For non-debug builds, this would be a silent error.

My solution to this problem is to have each `Gb_Apu_Buffer` store their own `m_time` variable rather than sharing a single one. This makes sense because each `Gb_Apu_Buffer` is a separate instance of the Game Boy APU, so they should be keeping track of their own CPU time values.

I also fixed the coding style in all the methods I edited and renamed `Gb_Apu_Buffer` to `GbApuWrapper` since that better reflects its purpose and follows the LMMS naming convention.

### DIFF
--- a/plugins/FreeBoy/FreeBoy.cpp
+++ b/plugins/FreeBoy/FreeBoy.cpp
@@ -2,7 +2,7 @@
  * FreeBoy.cpp - GameBoy papu based instrument
  *
  * Copyright (c) 2008 Attila Herman <attila589/at/gmail.com>
- *				Csaba Hruska <csaba.hruska/at/gmail.com>
+ *                    Csaba Hruska <csaba.hruska/at/gmail.com>
  *
  * This file is part of LMMS - https://lmms.io
  *
@@ -23,11 +23,12 @@
  *
  */
 
-#include <cmath>
-
-#include <QDomElement>
 #include "FreeBoy.h"
-#include "Gb_Apu_Buffer.h"
+
+#include <cmath>
+#include <QDomElement>
+
+#include "GbApuWrapper.h"
 #include "base64.h"
 #include "InstrumentTrack.h"
 #include "Knob.h"
@@ -45,8 +46,11 @@ namespace lmms
 {
 
 
-const blip_time_t FRAME_LENGTH = 70224;
-const long CLOCK_RATE = 4194304;
+namespace
+{
+constexpr blip_time_t FRAME_LENGTH = 70224;
+constexpr long CLOCK_RATE = 4194304;
+}
 
 extern "C"
 {
@@ -118,9 +122,7 @@ FreeBoyInstrument::FreeBoyInstrument( InstrumentTrack * _instrument_track ) :
 	m_trebleModel( -20.0f, -100.0f, 200.0f, 1.0f, this, tr( "Treble" ) ),
 	m_bassModel( 461.0f, -1.0f, 600.0f, 1.0f, this, tr( "Bass" ) ),
 
-	m_graphModel( 0, 15, 32, this, false, 1 ),
-
-	m_time(0)
+	m_graphModel( 0, 15, 32, this, false, 1 )
 {
 }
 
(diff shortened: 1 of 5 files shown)


superpaik (commented):
This is a comment for "Refactor wrapper class" commit. Sorry, I don't quite understand how git works yet.
Style. Functions should not have spaces between brackets and variables. Like in 
papu->setSampleRate( samplerate, CLOCK_RATE );
I'm not sure what's the coding convention for pointers and references, like in
long Gb_Apu_Buffer::readSamples(blip_sample_t* out, long count)
I'd say it should be 
long Gb_Apu_Buffer::readSamples(blip_sample_t * out, long count)

plugins/FreeBoy/FreeBoy.cpp
```diff
@@ -236,188 +236,192 @@ f_cnt_t FreeBoyInstrument::desiredReleaseFrames() const
...
 		framesleft -= count;
 	}
-	instrumentTrack()->processAudioBuffer( _working_buffer, frames + offset, _n );
+	instrumentTrack()->processAudioBuffer(workingBuffer, frames + offset, nph);
 }
 
 
 
-void FreeBoyInstrument::deleteNotePluginData( NotePlayHandle * _n )
+void FreeBoyInstrument::deleteNotePluginData( NotePlayHandle * nph )
```
superpaik: Style

plugins/FreeBoy/FreeBoy.cpp
```diff
@@ -236,188 +236,192 @@ f_cnt_t FreeBoyInstrument::desiredReleaseFrames() const
...
 {
-	delete static_cast<Gb_Apu_Buffer *>( _n->m_pluginData );
+	delete static_cast<Gb_Apu_Buffer *>( nph->m_pluginData );
```
superpaik: Refactor wrapper class

superpaik (approved):

plugins/FreeBoy/FreeBoy.cpp
messmerd: You can click the "Files Changed" tab in GitHub to see all my changes together before writing review comments. Looking at this PR commit-by-commit is just helpful for understanding what kinds of changes I made to the code, because the changes made in the 1st commit are what fix the bug, the 2nd commit does some refactoring, the 3rd fixes most of the style, etc.

Line 253 of FreeBoy.cpp is the only place I know of where I don't adhere to the style guidelines, and that was deliberate because I have another open PR which fixes that line and I want to get it into master right after this PR.

superpaik (approved):
LGTM

plugins/FreeBoy/GbApuWrapper.h
```diff
@@ -31,25 +32,30 @@ namespace lmms
...
+	~GbApuWrapper() = default;
 
-	void end_frame(blip_time_t);
+	blargg_err_t setSampleRate(long sampleRate, long clockRate);
+	void writeRegister(unsigned addr, int data) { Gb_Apu::write_register(fakeClock(), addr, data); }
+	long samplesAvail() const;
+	long readSamples(blip_sample_t* out, long count);
+	void trebleEq(const blip_eq_t& eq) { Gb_Apu::treble_eq(eq); }
+	void bassFreq(int freq);
+	void endFrame(blip_time_t endTime);
```
PhysSong, line 48: I'm not sure if it's better to match the original code style of the submoduled upstream library. @tresf Any ideas, since you wrote this class?

messmerd: Even with upstream changes, we'd still need a wrapper class for handling the CPU time. I don't think it's really worth it or should be within the scope of this simple bug fix PR. But after this PR is complete, I can look into upgrading to the latest version of the library and making significant structural changes like that.

tresf: Thanks... Yeah, my intentions aren't to delay anything.  I'm no longer an active contributor to LMMS, so I'm mostly brainstorming here. The code style guidelines are a wider project decision and since the codebase has recently received a huge effort to standardize all of the codebase, it's only natural that these types of exceptions need to be decided upon.

@JohannesLorenz / @PhysSong if the decision is to maintain the submodile style, is there a quick way make this process easier (or at least more straight forward) for contributors?

For example, the switch to the submodule deliberately changed the plugin name to match our internal conventions, however some wrapper code just doesn't benefit from forced styling due to divergence from upstream.

The project should have guidelines for this so that contributions aren't being held up by missing policy.

JohannesLorenz: I would say: our files -> our style. Their files -> their style. If we start changing style in their submodule, we risk ugly merge errors when we update from their upstream. If we use their style in our files, we risk issues with our code check/format tools and also difficulties amongst our developers. Of course, this can mean that some of our files use mixed styles in the same source file, but we have this already with STL or Qt. Though, it's just my opinion and I am open to other suggestions.

Veratil (approved):
Nothing else stands out to me, and renders shared in Discord sound much better and without clicks and static.

PR #6683, opened 2023-03-30
Add Dispersion effect

LostRobotMusic:
image

"Dispersion" is a convenient allpass filter plugin allowing for very high filter orders.  This causes it to act as a frequency-dependent delay, where the delay length is different for every frequency.  This is great for a very pleasing smearing of buzzy timbres, greatly exaggerating transients with a laser-like tone, and much more.  I've also added a Feedback knob as a very quick and convenient way to crank even more unique timbres out of the effect, as well as a built-in DC offset remover since the Feedback can sometimes get a bit out of control.

With the exception of the very high filter order and the new Feedback feature, this effect is essentially identical to Kilohearts's Disperser (which costs $60).

Here's a quick and simple demo:
https://www.youtube.com/watch?v=O9sqD773sNU


Veratil:
In the demo video there are clicks when you're changing the number on the left with the first and third track (I assume we just can't hear it in the second track). Is this expected?

LostRobotMusic:
The parameter will always have clicks since it's changing the number of filters being applied (it's unfortunately impossible to apply a partial filter).  However, I can attempt to reduce the clicking by flushing the filter buffers once they're no longer being used.  The parameter isn't meant to ever be automated, but that should at least make it less annoying to work with.  I'll try it and see how it sounds.

LostRobotMusic:
Alright, I resolved that issue.  There is still clicking unfortunately, but it is significantly more quiet and consistent and much less abrupt and distracting.

plugins/Dispersion/Dispersion.h
```diff
@@ -0,0 +1,74 @@
...
+
+#include "Effect.h"
+#include "DispersionControls.h"
+#include "lmms_math.h"
+#include <qmath.h>
+
+namespace lmms
+{
+
+constexpr int MAX_DISPERSION_FILTERS = 400;
```
Veratil, line 37: Just curious, what defines this limit?

Veratil (commented):
LGTM, but...

I'm hesitant about `qFastSin` and `qFastCos`. These are undocumented functions in Qt and are not supported and/or subject to change or be removed without warning.

plugins/Dispersion/Dispersion.h
LostRobotMusic: It was chosen arbitrarily.  I intended to increase it to 999 later on but forgot to, thanks for pointing it out.

LostRobotMusic:
I've replaced the trig functions with the standard ones, I didn't realize the Qt ones weren't officially supported.  LMMS probably needs to come with some sine/cosine approximations, given how often those calculations need to be made and how rarely perfect accuracy is actually needed...

superpaik:
LGTM and sounds ok.

Some thoughs:
- I think it'd be useful to have a label on the amount, so people know what that number stands for.
- I'm not sure what should be the name of the main class file (dispersion). Shouldn't it be called as the class inside, that is dispersioneffect? In the pluggins effects folder there are some that follow that rule and some that follow the other. For instance, in your last effect, phaser, the effect main file is called phasereffect.

- Bonus: Maybe you have tested it, but If you cranck up the feedback and stop playing you can get some crazy sounds out of it that are good for sound design (it's a pity we can not record from inside LMMS)

plugins/Dispersion/Dispersion.h
```diff
@@ -0,0 +1,73 @@
...
+	sample_t m_apY0[MAX_DISPERSION_FILTERS][2] = {{0}};
+	sample_t m_apY1[MAX_DISPERSION_FILTERS][2] = {{0}};
+	
+	float m_feedbackVal[2] = {0};
+	float m_integrator[2] = {0};
+	
+	int m_amountVal;
+
+	friend class DispersionControls;
+} ;
```
messmerd, line 68:
```suggestion
};
```
style

plugins/Dispersion/Dispersion.h
```diff
@@ -0,0 +1,73 @@
...
+
+
+} // namespace lmms
+
+#endif
```
messmerd, line 73:
```suggestion
#endif // LMMS_DISPERSION_H
```
style

plugins/Dispersion/DispersionControlDialog.h
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this program (see COPYING); if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA 02110-1301 USA.
+ *
+ */
+
+#ifndef DISPERSION_CONTROL_DIALOG_H
+#define DISPERSION_CONTROL_DIALOG_H
```
messmerd, line 26:
```suggestion
#ifndef LMMS_GUI_DISPERSION_CONTROL_DIALOG_H
#define LMMS_GUI_DISPERSION_CONTROL_DIALOG_H
```
Include guard style

plugins/Dispersion/DispersionControlDialog.h
```diff
@@ -0,0 +1,52 @@
...
+namespace gui
+{
+
+class DispersionControlDialog : public EffectControlDialog
+{
+	Q_OBJECT
+public:
+	DispersionControlDialog(DispersionControls* controls);
+	~DispersionControlDialog() override = default;
+} ;
```
messmerd, line 45:
```suggestion
};
```
style

plugins/Dispersion/DispersionControlDialog.h
```diff
@@ -0,0 +1,52 @@
...
+
+
+} // namespace gui
+
+} // namespace lmms
+
+#endif
```
messmerd, line 52:
```suggestion
#endif // LMMS_GUI_DISPERSION_CONTROL_DIALOG_H
```
style

plugins/Dispersion/DispersionControls.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+
+
+void DispersionControls::loadSettings(const QDomElement& _this)
+{
+	m_amountModel.loadSettings(_this, "amount");
+	m_freqModel.loadSettings(_this, "freq");
+	m_resoModel.loadSettings(_this, "reso");
+	m_feedbackModel.loadSettings(_this, "feedback");
+	m_dcModel.loadSettings(_this, "dc");
```
messmerd, line 55:
```suggestion
void DispersionControls::loadSettings(const QDomElement& elem)
{
	m_amountModel.loadSettings(elem, "amount");
	m_freqModel.loadSettings(elem, "freq");
	m_resoModel.loadSettings(elem, "reso");
	m_feedbackModel.loadSettings(elem, "feedback");
	m_dcModel.loadSettings(elem, "dc");
```
Naming conventions

plugins/Dispersion/DispersionControls.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+
+
+void DispersionControls::saveSettings(QDomDocument& doc, QDomElement& _this)
+{
+	m_amountModel.saveSettings(doc, _this, "amount");
+	m_freqModel.saveSettings(doc, _this, "freq");
+	m_resoModel.saveSettings(doc, _this, "reso");
+	m_feedbackModel.saveSettings(doc, _this, "feedback");
+	m_dcModel.saveSettings(doc, _this, "dc");
```
messmerd, line 67:
```suggestion
void DispersionControls::saveSettings(QDomDocument& doc, QDomElement& elem)
{
	m_amountModel.saveSettings(doc, elem, "amount");
	m_freqModel.saveSettings(doc, elem, "freq");
	m_resoModel.saveSettings(doc, elem, "reso");
	m_feedbackModel.saveSettings(doc, elem, "feedback");
	m_dcModel.saveSettings(doc, elem, "dc");
```
Naming conventions

plugins/Dispersion/DispersionControls.h
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this program (see COPYING); if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA 02110-1301 USA.
+ *
+ */
+
+#ifndef DISPERSION_CONTROLS_H
+#define DISPERSION_CONTROLS_H
```
messmerd, line 26:
```suggestion
#ifndef LMMS_DISPERSION_CONTROLS_H
#define LMMS_DISPERSION_CONTROLS_H
```
Include guard style

plugins/Dispersion/DispersionControls.h
```diff
@@ -0,0 +1,82 @@
...
+	IntModel m_amountModel;
+	FloatModel m_freqModel;
+	FloatModel m_resoModel;
+	FloatModel m_feedbackModel;
+	BoolModel m_dcModel;
+
+	friend class gui::DispersionControlDialog;
+	friend class DispersionEffect;
+
+} ;
```
messmerd, line 77:
```suggestion
};
```
style

plugins/Dispersion/DispersionControls.h
```diff
@@ -0,0 +1,82 @@
...
+
+
+} // namespace lmms
+
+#endif
```
messmerd, line 82:
```suggestion
#endif // LMMS_DISPERSION_CONTROLS_H
```
style

plugins/Dispersion/DispersionControls.cpp
```diff
@@ -0,0 +1,73 @@
...
+ * Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA 02110-1301 USA.
+ *
+ */
+
+
+#include <QDomElement>
+
+#include "DispersionControls.h"
+#include "Dispersion.h"
```
messmerd, line 29:
```suggestion
#include "DispersionControls.h"
#include "Dispersion.h"

#include <QDomElement>
```
Include order

messmerd (commented):
Mostly just a style review. Haven't built the plugin yet

plugins/Dispersion/Dispersion.cpp
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	for (fpp_t f = 0; f < frames; ++f)
+	{
+		sample_t s[2] = { buf[f][0] + m_feedbackVal[0], buf[f][1] + m_feedbackVal[1] };
+		
+		runDispersionAP(m_amountVal, apCoeff1, apCoeff2, s);
+		m_feedbackVal[0] = s[0] * feedback;
+		m_feedbackVal[1] = s[1] * feedback;
+
```
superpaik, line 107: When m_amountVal is zero I believe runDispersionAP does nothing to the samples, but afterwards you apply feedback to them. Is that intended? See the effect it has on the sound.
DispersionAnalysis

LostRobotMusic: Thanks for pointing that out.  I did initially leave that in intentionally, but I understand that if the filter count is at 0 the user is likely expecting the effect to not do anything.  I'll make sure the feedback is set to 0 when no filters are active.

LostRobotMusic:
I added support for negative feedback amounts which can create some new useful timbres.  I intended to try this from the beginning, but forgot to.

LostRobotMusic:
> * I'm not sure what should be the name of the main class file (dispersion). Shouldn't it be called as the class inside, that is dispersioneffect? In the pluggins effects folder there are some that follow that rule and some that follow the other. For instance, in your last effect, phaser, the effect main file is called phasereffect.

@superpaik I'd prefer this way over the way the Phaser does it currently.  I generally base my effects off of LMMS's Amplifier effect, which uses a class named AmplifierEffect and a file named Amplifier.

> * Bonus: Maybe you have tested it, but If you cranck up the feedback and stop playing you can get some crazy sounds out of it that are good for sound design (it's a pity we can not record from inside LMMS)

That's the magic of allpass filters.  I have a very complex frequency/pitch shifter effect on the way which abuses allpass filters to an even more extreme extent, you'll really enjoy it.

plugins/Dispersion/Dispersion.cpp
```diff
@@ -0,0 +1,166 @@
...
+				m_apX0[i][0] = m_apX0[i][1] = m_apX1[i][0] = m_apX1[i][1] = 0;
+				m_apY0[i][0] = m_apY0[i][1] = m_apY1[i][0] = m_apY1[i][1] = 0;
+			}
+		}
+		m_amountVal = amount;
+	}
+	
+	if (amount == 0)
+	{
+		feedback = 0;
```
superpaik, line 102: I think that you need to reset the model as well, because if not you can get a click on the first sample if m_feedbackVal has some value

plugins/Dispersion/DispersionControls.h
```diff
@@ -0,0 +1,82 @@
...
+	IntModel m_amountModel;
+	FloatModel m_freqModel;
+	FloatModel m_resoModel;
+	FloatModel m_feedbackModel;
+	BoolModel m_dcModel;
+
+	friend class gui::DispersionControlDialog;
+	friend class DispersionEffect;
+
+};
```
messmerd, line 77:
```suggestion
	friend class DispersionEffect;
};
```
Make CodeFactor happy

plugins/Dispersion/DispersionControls.h
```diff
@@ -0,0 +1,82 @@
...
+
+class DispersionControls : public EffectControls
+{
+	Q_OBJECT
+public:
+	DispersionControls(DispersionEffect* effect);
+	~DispersionControls() override = default;
+
+	void saveSettings(QDomDocument & _doc, QDomElement & _parent) override;
+	void loadSettings(const QDomElement & _this) override;
```
messmerd, line 50:
```suggestion
	void saveSettings(QDomDocument& doc, QDomElement& elem) override;
	void loadSettings(const QDomElement& elem) override;
```
I missed this in my initial review

plugins/Dispersion/Dispersion.h
```diff
@@ -0,0 +1,73 @@
...
+	
+	float m_sampleRate;
+	
+	sample_t m_apX0[MAX_DISPERSION_FILTERS][2] = {{0}};
+	sample_t m_apX1[MAX_DISPERSION_FILTERS][2] = {{0}};
+	sample_t m_apY0[MAX_DISPERSION_FILTERS][2] = {{0}};
+	sample_t m_apY1[MAX_DISPERSION_FILTERS][2] = {{0}};
+	
+	float m_feedbackVal[2] = {0};
+	float m_integrator[2] = {0};
```
messmerd, line 63:
```suggestion
	std::array<float, 2> m_feedbackVal{};
	std::array<float, 2> m_integrator{};
```
Can avoid using C arrays

plugins/Dispersion/Dispersion.h
```diff
@@ -0,0 +1,73 @@
...
+ *
+ */
+
+
+#ifndef LMMS_DISPERSION_H
+#define LMMS_DISPERSION_H
+
+#include "Effect.h"
+#include "DispersionControls.h"
+#include "lmms_math.h"
```
messmerd, line 31:
```suggestion
#include "Effect.h"

#include <array>

#include "DispersionControls.h"
#include "lmms_math.h"
```

plugins/Dispersion/Dispersion.cpp
```diff
@@ -0,0 +1,167 @@
...
+{
+	LMMS_STRINGIFY(PLUGIN_NAME),
+	"Dispersion",
+	QT_TRANSLATE_NOOP("PluginBrowser", "An allpass filter allowing for extremely high orders."),
+	"Lost Robot <r94231/at/gmail/dot/com>",
+	0x0100,
+	Plugin::Effect,
+	new PluginPixmapLoader("logo"),
+	nullptr,
+	nullptr,
```
messmerd, line 46:
```suggestion
	QT_TRANSLATE_NOOP("PluginBrowser", "An all-pass filter allowing for extremely high orders."),
	"Lost Robot <r94231/at/gmail/dot/com>",
	0x0100,
	Plugin::Effect,
	new PluginPixmapLoader("logo"),
	nullptr,
	nullptr
```
Typo

messmerd (commented):
Just a few more minor suggestions

messmerd:
I tested it and it seems to work fine.
However:
- If feedback is set to -1 and a note plays, the sound won't stop even if I stop playback. Maybe this is an issue with core LMMS, not the plugin
- The CPU usage when the all-pass filter amount is increased is very high, so hopefully there are some areas for performance improvement. Did you plan on using `fastSin`/`fastCos`?

plugins/Dispersion/DispersionControlDialog.cpp
```diff
@@ -0,0 +1,81 @@
...
+	QPalette pal;
+	pal.setBrush(backgroundRole(), PLUGIN_NAME::getIconPixmap("artwork"));
+	setPalette(pal);
+	setFixedSize(207, 50);
+
+	LcdSpinBox * m_amountBox = new LcdSpinBox(3, this, "Amount");
+	m_amountBox->setModel(&controls->m_amountModel);
+	m_amountBox->move(5, 10);
+	m_amountBox->setLabel(tr("AMOUNT"));
+	m_amountBox->setToolTip(tr("Number of allpass filters"));
```
messmerd, line 51:
```suggestion
	m_amountBox->setToolTip(tr("Number of all-pass filters"));
```
Typo

plugins/Dispersion/Dispersion.cpp
LostRobotMusic: I disagree, "allpass" without the hyphen is also correct and about equally common (same with lowpass and highpass).

LostRobotMusic:
> * If feedback is set to -1 and a note plays, the sound won't stop even if I stop playback. Maybe this is an issue with core LMMS, not the plugin

This is both intentional and unavoidable.  100% feedback means 100% of the output is fed back into the input, which causes an infinite loop.

> * The CPU usage when the all-pass filter amount is increased is very high, so hopefully there are some areas for performance improvement. Did you plan on using `fastSin`/`fastCos`?

"Amount" controls the number of filters being used, and running hundreds of filters is always going to be expensive.  Faster trig functions don't matter here since almost all of the performance usage is coming from `runDispersionAP` which is pure arithmetic.  Unfortunately, there isn't much room for SIMD optimizations or anything like that, since the filters are all run one after the other rather than in parallel.

plugins/Dispersion/Dispersion.h
Veratil: > Is there anything in the style guidelines against using C arrays?

Style guideline, no. This is a C++ guideline really.

> If I change these to STD arrays, I'd want to change the other arrays over (to std::array<std::array<sample_t, 2>, MAX_DISPERSION_FILTERS>) as well for consistency, which would hinder the readability of the code with no benefit.

Luckily C++ allows you to use `using name = type`. This is recommended to improve readability.
So you could do something like this:
```cpp
using Channels = std::array<sample_t, 2>;
using Filters = std::array<Channels, MAX_DISPERSION_FILTERS>;

Filters m_filters;
```

plugins/Dispersion/Dispersion.cpp
messmerd: While I have frequently seen it as "allpass", "all-pass" seems slightly more common and it's what Wikipedia uses. But more importantly, in other places in LMMS, such as the filter tab of instruments and the Dual Filter effect, "all-pass" (and "low-pass"/"high-pass") is used. Some LADSPA plugins use "allpass" but they are 3rd party in origin.

LostRobotMusic:
After a few hours of tinkering and profiling, I've managed to get the allpass filters to run approximately 17% faster.  There don't really seem to be many more potential optimizations that could possibly be made at this point.

plugins/Dispersion/Dispersion.cpp
```diff
@@ -130,24 +130,21 @@ bool DispersionEffect::processAudioBuffer(sampleFrame* buf, const fpp_t frames)
...
+        const sample_t currentInput = put[channel];
+        const sample_t filterOutput = apCoeff1 * (currentInput - m_state.y1[i])
+            + apCoeff2 * (m_state.x0[i] - m_state.y0[i]) + m_state.x1[i];
+        m_state.x1[i] = m_state.x0[i];
+        m_state.x0[i] = currentInput;
+        m_state.y1[i] = m_state.y0[i];
+        m_state.y0[i] = filterOutput;
+
+        put[channel] = filterOutput;
+    }
```
messmerd, line 147: Just need to use tabs here. Other than that, everything looks good to me

messmerd (approved):
LGTM

sakertooth (approved):
LGTM.

PR #6686, opened 2023-04-12
Fix LcdFloatSpinBox constructor and label alignment

he29-net:
Adds missing calls to `QWidget` constructors, so that `LcdFloatSpinBox` works correctly when not placed in a layout.

Plus two small label tweaks: fixed height ensures that appropriate vertical space is reserved for the label (without it I had issues when placing the spinbox in the main window toolbar), and replacing `horizontalAdvance` by `boundingRect` should very slightly improve vertical alignment (deprecation warning recommended replacing `width()` with `horizontalAdvance()` without any details, but in this case `boundingRect()` probably makes more sense).

### DIFF
--- a/src/gui/widgets/LcdFloatSpinBox.cpp
+++ b/src/gui/widgets/LcdFloatSpinBox.cpp
@@ -49,6 +49,7 @@ namespace lmms::gui
 
 
 LcdFloatSpinBox::LcdFloatSpinBox(int numWhole, int numFrac, const QString& name, QWidget* parent) :
+	QWidget(parent),
 	FloatModelView(new FloatModel(0, 0, 0, 0, nullptr, name, true), this),
 	m_wholeDisplay(numWhole, parent, name, false),
 	m_fractionDisplay(numFrac, parent, name, true),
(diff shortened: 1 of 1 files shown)


src/gui/widgets/LcdFloatSpinBox.cpp
```diff
@@ -240,9 +243,9 @@ void LcdFloatSpinBox::paintEvent(QPaintEvent*)
 	{
 		p.setFont(pointSizeF(p.font(), 6.5));
 		p.setPen(m_wholeDisplay.textShadowColor());
-		p.drawText(width() / 2 - horizontalAdvance(p.fontMetrics(), m_label) / 2 + 1, height(), m_label);
+		p.drawText(width() / 2 - p.fontMetrics().boundingRect(m_label).width() / 2 + 1, height(), m_label);
```
Veratil, line 246: If we're getting rid of `horizontalAdvance` can you check if we can remove `#include "DeprecationHelper.h"` too?

he29-net: On second thought, `QWheelEvent::position()` says "introduced in Qt 5.14", so it's probably better to keep the deprecation wrapper in use for now. `boundingRect()` should be fine, it existed in Qt 4.8. at the very least.

PhysSong (approved):

DomClark (approved):

Issue #6495, opened 2022-08-21
Ripple delete the notations in piano editor


anytizer:
### Enhancement Summary
To ripple delete notations; shifting all notes from right to left at the point of deletion.

#### Justification
1. It may be useful to delete the some of the notes in between, but without leaving the gap in the notations.
2. May be a keyboard shortcut has to be implemented to make a difference between regular delete and ripple delete.

![image](https://user-images.githubusercontent.com/5563341/185811184-613ae4b0-a2bc-41c0-9efe-be2545b6c94b.png)

In the picture, I want to delete the notes between first G4 to next G4 (selected with blue color). All note notes on the right would then shift towards left by 3 bars; resulting no **gap**.

Spekular:
I would propose shift + delete as the shortcut, and that this be implemented in the song editor as well.

This should just be a matter of finding all notes that start at or after the end of the selection, and shifting them left by the length of the selection (start of leftmost note to end of rightmost note), so I'll mark it as a good first issue.

anytizer:
Possibly the "opposite" to #6497 bulldozer's feature where notes shift towards left.

Monospace-V:
Possibly just backspace or shift-backspace? Since it's more associated with moving everything else back by as much as is deleted...

Spekular:
I was thinking shift since that's used for proportional resize, but I forgot that you need to hold control for that to move following notes. Perhaps shift+ctrl+delete would be appropriate then, which would avoid clashing with shift+delete for removing bars in the song editor.

allejok96:
Wouldn't shift delete be perfect since you're essentially doing what remove bar does (plus removing the selected notes)?

I know we're talking piano roll now, but if it turns out it clashes so badly with song editor, I'd say drop the shortkey for insert/remove bar, since we have a button for it, it's not something you frequently do and especially remove bar has quite an unpolished behavior.

hberg137:
Since it seems like this issue was not yet resolved, I am going to try and work on this issue (as a first issue) :)

hberg137:
I completed work on this and just made a pull request for this issue! #6692

Kartik-1311:
would propose to implement shift + delete keyboard shortcut for ripple delete. Anyone still working on this... would like to help him as a newbie.

vinodkrishnanr:
[...]

PR #6692, opened 2023-04-18
Added ripple delete

hberg137:
Pull request for ripple delete issue #6495 

Please see [the following video](https://www.youtube.com/watch?v=BJERT9EzJao) for demonstration/explanation.

### DIFF
--- a/src/gui/editors/PianoRoll.cpp
+++ b/src/gui/editors/PianoRoll.cpp
@@ -1458,9 +1458,47 @@ void PianoRoll::keyPressEvent(QKeyEvent* ke)
 
 		case Qt::Key_Backspace:
 		case Qt::Key_Delete:
-			deleteSelectedNotes();
-			ke->accept();
-			break;
+			{
+				// Get start pos and end pos of selection; used for shift + delete
+				auto selectedNotes = getSelectedNotes();
+				if (selectedNotes.empty()) { break; }
+				TimePos delStartPos = selectedNotes.first()->pos();
+				TimePos delEndPos = selectedNotes.last()->endPos();
+				TimePos delEndLen = selectedNotes.last()->length();
+
+				deleteSelectedNotes();
+
+				// Shift notes on the right of deleted notes to the left to fill deleted note gap
+				if (ke->modifiers() & Qt::ShiftModifier)
+				{
+					// Beat/Bassline editor notes have negative length; this adjusts for that
+					if (delEndLen < 0)
+					{
+						delEndPos -= delEndLen;
+					}
+					
+					NoteVector rightNotes;
+					for (Note* n : m_midiClip->notes())
+					{
+						if (n->pos() >= delEndPos)
+						{
+							rightNotes.push_back(n);
+						}
+					}
+
+					// In the case of the last deleted note having negative length...
+					// Assume that intended length is gap until the first note
+					// to the right of the deleted region in this case
+					if (delEndLen < 0 && !rightNotes.isEmpty())
+					{
+						delEndPos = rightNotes.first()->pos();
+					}
+
+					shiftPos(rightNotes, delStartPos - delEndPos);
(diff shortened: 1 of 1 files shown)


superpaik:
When the selection of notes is not "continuous" (Shift+Ctrl+Mouse) the behavior is not consistent. Not sure what the behavior should be. Perhaps treat each "continuous" sub-group individually.

hberg137:
@superpaik I was not sure about this behavior either when developing it. I assumed that if a note (or notes) was within the range of the selected notes but not actually selected, as what it sounds like you are alluding to above, the unselected note(s) should not be shifted upon ripple deleting the selected notes. So, sort of ignoring those notes was an intentional choice since this behavior was not specified in the feature request.

I think your suggestion on what to do in this case would make sense, however implementing it may be a bit tricky in trying to detect notes in between the gaps of selected notes. I think I would have to partition the selected notes into different regions, detect the unselected notes between them, and then shift the unselected notes between them based on the length of the partitions. It seems feasible, but again, a bit tricky, especially since I would probably have to restructure some of the code I already wrote.

Do you want to keep the behavior as it currently is or do you want me to try and make this adjustment?

superpaik:
@hberg137 As I said, not so sure what the best behavior should be. I recognize it's a weird use-case scenario. If it was intended and as you had a discussion already, maybe the current is the best possible behavior.

src/gui/editors/PianoRoll.cpp
```diff
@@ -1458,9 +1458,47 @@ void PianoRoll::keyPressEvent(QKeyEvent* ke)
...
-			deleteSelectedNotes();
-			ke->accept();
-			break;
+			{
+				// Get start pos and end pos of selection; used for shift + delete
+				auto selectedNotes = getSelectedNotes();
+				if (selectedNotes.empty()) { break; }
+				TimePos delStartPos = selectedNotes.first()->pos();
+				TimePos delEndPos = selectedNotes.last()->endPos();
+				TimePos delEndLen = selectedNotes.last()->length();
```
allejok96, line 1467: A prior note may still end after the last note depending on its length. You may find `boundsForNotes` helpful.

It will still be tricky with drum notes, but I imagine something like this:

```
area to delete     |xxxxxx|            |xxxxxxxxx|
normal note         ######    ####      ####      ###
drum note              #                    #  
```

In that case you just need to check if the drum note starts on the end position... I also see that `boundsForNotes` doesn't handle drum notes well. That needs to be fixed somehow.

src/gui/editors/PianoRoll.cpp
```diff
@@ -1458,9 +1458,47 @@ void PianoRoll::keyPressEvent(QKeyEvent* ke)
...
+
+					// In the case of the last deleted note having negative length...
+					// Assume that intended length is gap until the first note
+					// to the right of the deleted region in this case
+					if (delEndLen < 0 && !rightNotes.isEmpty())
+					{
+						delEndPos = rightNotes.first()->pos();
+					}
+
+					shiftPos(rightNotes, delStartPos - delEndPos);
```
allejok96, line 1497: You may disable journalling before moving to prevent the double undo

    m_midiClip->saveJournallingState(false);
    shiftPos(...);
    m_midiClip->restoreJournallingState();

allejok96:
>  if a note (or notes) was within the range of the selected notes but not actually selected it should not be shifted upon ripple deleting the selected notes

No please. This creates very weird results. I see two feasible alternatives...

1. Delete everything from start to end. You may justify it by the way shift+select works in file managers, or the way shift+delete deletes the whole line in some text editors.

1. Do nothing and show a warning like glue notes does. Less destruction but could be annoying in a case like this

bild

allejok96 (changes requested):

luzpaz:
Any progress on this ?

PR #6693, opened 2023-04-20
Fixed the bug where mono VSTs only output on the left channel

### LINKED ISSUE #6558: Mono VST plugins output only to the left channel
### Bug Summary
When using a mono VST plugin via VeSTige, sound is produced only in the left channel instead of both channels.

#### Steps to reproduce
Load this Windows VST in VeSTige https://github.com/bsutherland/JuceOPLVSTi/releases/download/1.0.1/JuceOPLVSTi.x64.dll and play any note.

#### Expected behavior
Same audio should play in both channels.

#### Actual behavior
Audio is hard-panned to the left.

#### Affected LMMS versions
Version 1.3.0-alpha.1.102+g89fc6c9 (Linux/x86_64, Qt 5.6.3, GCC 5.4.0 20160609).
AppImage downloaded from here https://github.com/LMMS/lmms/releases/download/v1.3.0-alpha.1/lmms-1.3.0-alpha.1.102%2Bg89fc6c9-linux-x86_64.AppImage

QuixoticJk:
Added a check for the amount of input channels and routed a single channel to both out channels if there is only one channel coming in, this should fix the problem of single inputs only getting processed into a single output channel

could still use some testing and might need some reformatting since its my first time using the lmms conventions and im still getting used to it

Fixes #6558


plugins/VstBase/RemoteVstPlugin.cpp
```diff
@@ -1087,21 +1088,41 @@ void RemoteVstPlugin::process( const sampleFrame * _in, sampleFrame * _out )
 		memset( m_outputs[i], 0, bufferSize() * sizeof( float ) );
 	}
 
-#ifdef OLD_VST_SDK
-	if( m_plugin->flags & effFlagsCanReplacing )
+	int numInputChannels = inputCount();
+	int numOutputChannels = outputCount();
+
```
superpaik, line 1093:
```suggestion

```

.travis.yml
messmerd, line 1: Why did you add this file?

plugins/VstBase/RemoteVstPlugin.cpp
messmerd, line 1: A lot of your changes don't follow our style guidelines. See: https://github.com/LMMS/lmms/wiki/Coding-Conventions

PhysSong:
I think this might not be a proper way to fix the issue which you want to fix. Can you provide an example case where this pull request helps?

messmerd:
> I think this might not be a proper way to fix the issue which you want to fix. Can you provide an example case where this pull request helps?

It looks like this PR is meant to fix #6558.

luzpaz:
@QuixoticJk please respond :point_up_2:

messmerd:
It seems that @QuixoticJk isn't responding so I will give them another week and then take over this PR.

zonkmachine:
Any suggestions for vsts to test this? In https://github.com/LMMS/lmms/issues/6558 is suggested JuceOPLVSTi but that one won't load for me.

messmerd:
I introduced a class called [`PluginPortConfig`](https://github.com/LMMS/lmms/pull/7199/files#diff-d965b73cba01403359847a8d5d3d5096170582e2f30ff2ab6346e2dfa1d18a7b) in my CLAP plugin PR as a general purpose solution for routing audio in and out of mono plugins. Maybe I should spin it off into its own PR to solve this problem with mono VSTs.

zonkmachine:
> I introduced a class called [`PluginPortConfig`](https://github.com/LMMS/lmms/pull/7199/files#diff-d965b73cba01403359847a8d5d3d5096170582e2f30ff2ab6346e2dfa1d18a7b) in my CLAP plugin PR as a general purpose solution for routing audio in and out of mono plugins. Maybe I should spin it off into its own PR to solve this problem with mono VSTs.

Yes. Do you mean merging this and then refactor it later?
PS. I'm still looking for plugins to test this PR/issue.

DomClark (changes requested):
As PhysSong suggests, the changes in this PR are incorrect. This puts what should be two sequential buffers into the left and right channels. The correct behaviour is to process a single buffer, and fill both channels with the same data.

messmerd:
This PR is superseded by #7247.

messmerd:
Closing in favor of #7247.

PR #6698, opened 2023-04-25
Add instrument plugins to Song Editor via right click

LostRobotMusic:
There seems to be a recurring trend of drag-and-drop completely breaking within LMMS.  It's been broken in Wayland for years, and it just recently broke for XOrg with GNOME as well in the new Fedora 38.

Being able to add instrument plugins via right click instead of drag-and-drop will allow people with this bug to still be able to make use of LMMS (though automation will still unfortunately be unusable...).
image

### DIFF
--- a/include/PluginBrowser.h
+++ b/include/PluginBrowser.h
@@ -63,13 +63,14 @@ public:
 	using PluginKey = Plugin::Descriptor::SubPluginFeatures::Key;
 	PluginDescWidget( const PluginKey & _pk, QWidget * _parent );
 	QString name() const;
-
+	void openInNewInstrumentTrack(QString value);
 
 protected:
 	void enterEvent( QEvent * _e ) override;
 	void leaveEvent( QEvent * _e ) override;
 	void mousePressEvent( QMouseEvent * _me ) override;
 	void paintEvent( QPaintEvent * _pe ) override;
+	void contextMenuEvent(QContextMenuEvent * e ) override;
 
 private:
 	constexpr static int DEFAULT_HEIGHT{24};
--- a/include/PluginBrowser.h
+++ b/include/PluginBrowser.h
@@ -70,7 +70,7 @@ protected:
 	void leaveEvent( QEvent * _e ) override;
 	void mousePressEvent( QMouseEvent * _me ) override;
 	void paintEvent( QPaintEvent * _pe ) override;
-	void contextMenuEvent(QContextMenuEvent * e) override;
+	void contextMenuEvent(QContextMenuEvent* e) override;
 
 private:
 	constexpr static int DEFAULT_HEIGHT{24};
--- a/include/PluginBrowser.h
+++ b/include/PluginBrowser.h
@@ -70,7 +70,7 @@ protected:
 	void leaveEvent( QEvent * _e ) override;
 	void mousePressEvent( QMouseEvent * _me ) override;
 	void paintEvent( QPaintEvent * _pe ) override;
-	void contextMenuEvent(QContextMenuEvent * e ) override;
+	void contextMenuEvent(QContextMenuEvent * e) override;
 
 private:
 	constexpr static int DEFAULT_HEIGHT{24};
--- a/src/gui/PluginBrowser.cpp
+++ b/src/gui/PluginBrowser.cpp
@@ -291,7 +291,7 @@ void PluginDescWidget::mousePressEvent( QMouseEvent * _me )
 }
 
 
-void PluginDescWidget::contextMenuEvent(QContextMenuEvent * e)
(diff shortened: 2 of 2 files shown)


src/gui/PluginBrowser.cpp
```diff
@@ -287,4 +291,30 @@ void PluginDescWidget::mousePressEvent( QMouseEvent * _me )
...
+
+void PluginDescWidget::openInNewInstrumentTrack(QString value)
+{
+	TrackContainer* tc = Engine::getSong();
+	
+	InstrumentTrack * it = dynamic_cast<InstrumentTrack*>(
+			Track::create(Track::InstrumentTrack,
+							tc));
+	InstrumentLoaderThread *ilt = new InstrumentLoaderThread(
+				this, it, value);
```
Veratil, line 315:
```suggestion
	auto it = dynamic_cast<InstrumentTrack*>(Track::create(Track::InstrumentTrack, tc));
	auto ilt = new InstrumentLoaderThread(this, it, value);
```

include/PluginBrowser.h
```diff
@@ -63,13 +63,14 @@ class PluginDescWidget : public QWidget
...
 	QString name() const;
-
+	void openInNewInstrumentTrack(QString value);
 
 protected:
 	void enterEvent( QEvent * _e ) override;
 	void leaveEvent( QEvent * _e ) override;
 	void mousePressEvent( QMouseEvent * _me ) override;
 	void paintEvent( QPaintEvent * _pe ) override;
+	void contextMenuEvent(QContextMenuEvent * e) override;
```
Veratil, line 73:
```suggestion
	void contextMenuEvent(QContextMenuEvent* e) override;
```

src/gui/PluginBrowser.cpp
```diff
@@ -287,4 +291,30 @@ void PluginDescWidget::mousePressEvent( QMouseEvent * _me )
 }
 
 
+void PluginDescWidget::contextMenuEvent(QContextMenuEvent * e)
```
Veratil, line 294:
```suggestion
void PluginDescWidget::contextMenuEvent(QContextMenuEvent* e)
```

src/gui/PluginBrowser.cpp
Veratil: Also it looks like this `ilt` pointer is hanging. What frees it?

LostRobotMusic: InstrumentLoaderThread is a QThread so Qt should free it.  TrackContainerView.cpp does the same thing.

Veratil: @PhysSong @DomClark Can you give some insight on this? Does the loader thread get freed from the heap somewhere?

sakertooth: ``InstrumentLoaderThread`` is a child of ``this`` (the ``PluginBrowser``), so the thread will be freed when its parent gets deleted.

src/gui/PluginBrowser.cpp
```diff
@@ -287,4 +291,24 @@ void PluginDescWidget::mousePressEvent( QMouseEvent * _me )
...
+	QMenu contextMenu(this);
+	contextMenu.addAction(
+		tr("Send to new instrument track"),
+		[=]{ openInNewInstrumentTrack(m_pluginKey.desc->name); }
+	);
+	contextMenu.exec(e->globalPos());
+}
+
+
+void PluginDescWidget::openInNewInstrumentTrack(QString value)
```
sakertooth, line 305: Any reason why ``value`` cant be a ``const`` ref?
```suggestion
void PluginDescWidget::openInNewInstrumentTrack(const QString& value)
```

src/gui/PluginBrowser.cpp
Veratil: Ah, gotcha! Everything I saw related to QThread wasn't saying anything about it freeing children (or even being allocated on the heap), but if the parent destructs it then cool. 👍

src/gui/PluginBrowser.cpp
Veratil: As I understand it, Qt objects are copy-on-write so this a shallow reference until writing and then creates a copy. If we're not changing the contents of value then using `QString` is fine.

PhysSong: I prefer `const QString&` to be consistent with the practice with `std::string`, but I don't consider this as a blocker.

allejok96 (approved):
Tested and looks good. Would perhaps be good with Send to active instrument track and Send to pattern editor, but as you said so many other thing depend on DnD so the fix should happen in Qt/Wayland. If it remains a problem we can start thinking about that later.

PR #6610, opened 2023-01-14
Refactor ``SampleBuffer``

sakertooth:
Did some heavy refactoring on ``SampleBuffer``. The main goals were to separate concerns of the class and improve maintainability and performance. This class shares little similarity with the current ``SampleBuffer``, so it is worth stating what this class is now and what the new classes introduced are.

## Description
* ``SampleBuffer`` is a mutable buffer class that stores audio data in the form of a ``std::vector<sampleFrame>``. It also stores the sample rate of the data as an ``int``, and optionally the file path of the audio file if it was created from one using ``std::optional<QString>``.
	* ``SampleBuffer`` adheres to the named requirements of ``Container`` and ``ReversibleContainer``.
	* When not shared across threads, it is fine to use a plain ``SampleBuffer`` as it will only be seen in one scope. When shared across threads however, ``SampleBuffer`` must be shared immutably using ``std::shared_ptr<const SampleBuffer>``.
	* Use ``SampleBuffer`` if the class is in a local scope and is not seen by other threads. If there's a chance it may be used by more than one thread (such as the UI & Audio thread) and is outside local scope, use ``std::shared_ptr<const SampleBuffer>``. The latter may not be real time safe for the audio thread if it happens to be the last owner of that ``const SampleBuffer`` object (as it will be responsible for deallocating the buffer), but for a refactoring PR, this is fine for now. Real-time safety can be handled in another PR.
* `Sample` is a thread-safe, mutable class that allows for manipulating a ``SampleBuffer`` independently from the buffer itself. As such, it holds most of the mutable state that was once in ``SampleBuffer``, as well as where the ``SampleBuffer`` can be played and visualized.
	* This class is useful for when a ``SampleBuffer`` is being shared immutably using ``std::shared_ptr<const SampleBuffer>`` and one sharer needs to set its sample rate, change its frequency and loudness, etc, for playback and/or visualization.
	* Use ``Sample`` if you plan on playing and/or visualizing a ``SampleBuffer``. Otherwise, you are better off using a ``SampleBuffer`` or ``std::shared_ptr<const SampleBuffer>``.
	* ``Sample::PlaybackState`` is a class that keeps track of the current position in playback for someone playing back the ``SampleBuffer`` through ``Sample``.
* ``SampleLoader`` handles opening and loading audio files that occur in the GUI. ``SampleLoader`` shows a dialog box when it fails to load a file. On failure to load a file, it returns an empty buffer in the form of a ``std::unique_ptr<SampleBuffer>``
[...]

PR #6703, opened 2023-05-04
Add cache for `SampleBuffer`

sakertooth:
This PR adds a cache for the `SampleBuffer` class, giving us two main benefits:
  * Faster load times by loading a sample from disk once and after that, from memory.
  * Potentially large memory savings by sharing samples between different users from one single data source.

Entries are automatically added to a `QFileSystemWatcher` once added to the cache, and when a file changes, it will evict it from the cache.


superpaik:
Test result on windows 10/64b, both mscv and mingw, loading a wav file into sample track (values are approximate)

|build | master | PR | master | PR | master | PR |
| ------------- | ------------- | ------------- | ------------- | ------------- | ------------- | ------------- |
| filesize  | 1 min  | 1 min  | 4 min  | 4 min  | 16 min  | 16 min  |
| msvc  | 2.58 s  | 2.6 s  | 11 s  | 10 s  | 38 s  | 46 s  | 
| mingw | 1.86 s | 1.8 s | 7 s  | 7 s | 27 s | 30 s  |

On the other hand, reloading the file is practically immediate on the PR. This can be positive if intended or negative if memory is not released when deleting a sample clip.

sakertooth:
> Test result on windows 10/64b, both mscv and mingw, loading a wav file into sample track (values are approximate)
> build 	master 	PR 	master 	PR 	master 	PR
> filesize 	1 min 	1 min 	4 min 	4 min 	16 min 	16 min
> msvc 	2.58 s 	2.6 s 	11 s 	10 s 	38 s 	46 s
> mingw 	1.86 s 	1.8 s 	7 s 	7 s 	27 s 	30 s
> 
> On the other hand, reloading the file is practically immediate on the PR. This can be positive if intended or negative if memory is not released when deleting a sample clip.

Yeah, the initial load of the samples should be similar to master, but subsequent loads of the same sample should be faster since it is likely to be in the cache.

superpaik:
Then is working perfectly! 
Especially useful when open a project that uses the same loop multiple times.

include/SampleBufferCache.h
```diff
@@ -0,0 +1,52 @@
...
+#include <memory>
+#include <optional>
+#include <unordered_map>
+#include <utility>
+#include <vector>
+
+namespace lmms
+{
+    class SampleBufferCache
+    {
```
superpaik, line 40: I think we don't have indentation inside namespace.

src/core/SampleBufferCache.cpp
```diff
@@ -0,0 +1,60 @@
...
+ * Boston, MA 02110-1301 USA.
+ *
+ */
+
+#include "SampleBufferCache.h"
+#include <cassert>
+
+namespace lmms
+{
+    SampleBufferCache::SampleBufferCache(int capacity) : m_capacity(capacity)
```
superpaik, line 30: Same as before. I think we don't have indentation inside namespace.

superpaik (approved):

src/core/SampleBuffer.cpp
```diff
@@ -90,6 +90,12 @@ SampleBuffer::SampleBuffer() :
 SampleBuffer::SampleBuffer(const QString & audioFile, bool isBase64Data)
 	: SampleBuffer()
 {
+	if (auto cachedBuffer = Engine::getSampleBufferCache()->get(audioFile.toStdString()); cachedBuffer.has_value())
+	{
+		*this = **cachedBuffer;
```
michaelgregorius, line 95: Doing this will likely create crashes in certain cases because this assignment will also copy the following raw pointers in `SampleBuffer`:
```
sampleFrame * m_origData;
sampleFrame * m_data;
```
When either `this` or `cachedBuffer` is deleted it will release the memory allocated to `m_origData` and `m_data`. Therefore the other `SampleBuffer` will be left with invalid data leading to a crash when it tries to access it.

src/core/SampleBuffer.cpp
```diff
@@ -90,6 +90,12 @@ SampleBuffer::SampleBuffer() :
 SampleBuffer::SampleBuffer(const QString & audioFile, bool isBase64Data)
 	: SampleBuffer()
 {
+	if (auto cachedBuffer = Engine::getSampleBufferCache()->get(audioFile.toStdString()); cachedBuffer.has_value())
```
michaelgregorius, line 93: In my opinion the caching and the update of the cache should be done outside of `SampleBuffer`. Things that are cached should not know that they are cached or know anything about the caching mechanism.

The `SampleBuffer` should ideally just be a low level buffer for samples and it should not know the context in which it is used. The fact that the low level object `SampleBuffer` accesses the high level object `Engine` shows that there is likely a problem with an indirection.

src/core/SampleBuffer.cpp
```diff
@@ -1521,6 +1542,7 @@ void SampleBuffer::loadFromBase64(const QString & data)
 
 	m_audioFile = QString();
 	update();
+	Engine::getSampleBufferCache()->put(data.toStdString(), this);
```
michaelgregorius, line 1545: The string `data` can be quite huge, especially if it contains audio data that's encoded as base64. Therefore it's not a good idea to use it as a key.

src/core/SampleBuffer.cpp
sakertooth: I was replanning to remove the LRU cache idea anyways, and just make it so that the project has a "Loaded Samples" Sidebar, and then we could just store and share the loaded samples in memory for as many samples the user needs. The user could then manage the cache on their own, so it isn't completely a black box. I should've drafted this earlier in that case.

And I would still need to continue working on #6610 since it would make a lot of what I'm trying to do here easier.

sakertooth:
Having an LRU cache was mostly a premature idea I had in mind. What I plan to do now is store the samples in an unbounded cache, and allowing the end user to manage the samples loaded into the project. Deleting a sample from the cache would then require it to be loaded from disk, and loading a sample already in the cache will be fast because it's already in memory.

michaelgregorius:
Hi @sakertooth,

if you manage all samples/audio regions in one place you should also consider to watch each of them using [QFileSystemWatcher](https://doc.qt.io/qt-6/qfilesystemwatcher.html) and reload them on a change. This way users could adjust the audio data in other applications and have the updated data directly available in LMMS. Classes which use that data will of course have to be aware that changes might occur, e.g. that an audio file might become shorter, and behave accordingly.

I think most DAWs, e.g. REAPER and Ardour, define regions (audio and MIDI)  that can be placed along the project. These regions then only reference the data so that one audio file can then potentially be used by several regions. Changing the audio file, e.g. outside of the DAW, will result in all regions that reference it using the changed data.

If I understand correctly LMMS makes use of virtual memory because it loads audio data directly into memory. This way the operating system will take care of memory blocks (pages) that have not been used for a while by swapping them out. One could also consider to implement some more direct form of memory management which also does not immediately load all the data into memory. I am not really an expert on that topic but I can imagine that memory mapped files would be loaded lazily into memory as the blocks are accessed.

sakertooth:
> if you manage all samples/audio regions in one place you should also consider to watch each of them using [QFileSystemWatcher](https://doc.qt.io/qt-6/qfilesystemwatcher.html) and reload them on a change. This way users could adjust the audio data in other applications and have the updated data directly available in LMMS.

Wouldn't this be considered a downside? If the user were to load a file and it was in the sample cache, and then move that same file into another directory, why should the DAW also move/remove the sample? I tested this behavior in other DAWs like FL, and it doesn't respond to file path changes at all. In FL, once you load a file from disk, it stays there. Deleting it from disk after the fact doesn't change anything about what FL previously loaded. And if the same sample was now in another directory and then loaded in, it would just be considered a new file altogether.

Unless you mean like a Refresh button of some kind, then that should be okay. However, I don't think this behavior should be automatic.

michaelgregorius:
My remark was not about the detection of removed/moved files but about the detection of changes within the files. The scenario is that a user keeps a file where it is but edits its content with an external audio editor. The basic premise is that an audio editor has more powerful tools to manipulate audio than a DAW. In REAPER I can change audio files with external editors and the changes are reflected immediately in the DAW. If a file is moved REAPER will show that the file is not available anymore.

I think it all boils down to how LMMS organizes/wants to organize large sets of audio data. If you want to keep the data safe in cases where the source file is moved to another location then this also implies that you either have to write a new file to a known location when the project is saved or you will have to include the audio data in the save file itself. The latter might not be a good idea with audio heavy projects and with regards to performance.

The latter question would also become relevant if audio recording is ever implemented in LMMS.

luzpaz:
Any progress on this ?

sakertooth:
> Any progress on this ?

I plan to continue with this PR soon.

src/core/SampleBuffer.cpp
sakertooth: What do you suggest should be done alternatively? I'm having trouble figuring out what key should be used for ``SampleBuffer``'s that originate from Base64 data. I'm also hoping that using ``std::string`` as a key for file paths are fine, since the data is not too large, but let me you know if you find that to be a problem as well.

michaelgregorius: In the current code the base64 path seems to be hit in two cases:
* In `AudioEngine::handleMetronome` when a play handle for a metronome sample is added (of which there are only two different ones)
* In `FileBrowserTreeWidget::previewFileItem` when a sample is previewed

The question is if it is really a good idea to fill the sample cache in these cases. Imagine a user that preview lots of files but uses only very few of them (if any). In that case the not used preview samples would "pollute" the cache with lots of data that's not really needed in the actual song.

src/core/SampleBufferCache.cpp
```diff
@@ -0,0 +1,84 @@
...
+	{
+		throw std::runtime_error{std::string{"Buffer with the same key already exists within the cache"}};
+	}
+
+	assert(m_cache.size() <= m_capacity);
+	if (m_cache.size() == m_capacity)
+	{
+		const auto leastRecentlyUsed = std::min_element(m_cache.begin(), m_cache.end(),
+			[](const auto& entryOne, const auto& entryTwo) { return entryOne.use_count() < entryTwo.use_count(); });
+		if (leastRecentlyUsed == m_cache.end()) { throw std::runtime_error{"Failed to evict buffer from cache"}; }
```
PhysSong, line 52: I think the only case where this can happen is `m_cache` is empty(`m_capacity == 0` or some race conditions). A cache with zero capacity doesn't make sense, so could we enforce non-zero capacity instead?

src/core/SampleBufferCache.cpp
```diff
@@ -0,0 +1,84 @@
...
+}
+
+auto SampleBufferCache::contains(const QString& key) const -> bool
+{
+	return m_cache.contains(key);
+}
+
+void SampleBufferCache::remove(const QString& key)
+{
+	if (m_cache.remove(key) == 0) { throw std::runtime_error{std::string{"Buffer not found"}}; }
```
PhysSong, line 71: Can we also have a non-throwing variant which probably returns a boolean value? Maybe `SampleBufferCache::add` too.

sakertooth: I was debating on wheter I would just return `bool` or throw an exception as I did not know which was better design wise.

PhysSong: It's also possible to add a non-throwing version with a different name such as `tryRemove`.

src/core/SampleBufferCache.cpp
```diff
@@ -0,0 +1,80 @@
...
+auto SampleBufferCache::add(const QString& key, Buffer buffer) -> bool
+{
+	if (buffer == nullptr || m_cache.contains(key)) { return false; }
+
+	assert(m_cache.size() <= m_capacity);
+	if (m_cache.size() == m_capacity)
+	{
+		const auto leastRecentlyUsed = std::min_element(m_cache.begin(), m_cache.end(),
+			[](const auto& entryOne, const auto& entryTwo) { return entryOne.use_count() < entryTwo.use_count(); });
+		m_cache.erase(leastRecentlyUsed);
```
messmerd, line 48:
```suggestion
		const auto leastUsed = std::min_element(m_cache.begin(), m_cache.end(),
			[](const auto& entryOne, const auto& entryTwo) { return entryOne.use_count() < entryTwo.use_count(); });
		m_cache.erase(leastUsed);
```
Wouldn't this be more accurate name since it's just comparing how many times it is used?

Though I'm not sure how well a least-used cache would perform. To have a least-recently-used cache, you may need to change the type of `m_cache` to something like:
```c++
QHash<QString, std::pair<Buffer, std::chrono::steady_clock::time_point>> m_cache;
```
And update the time every time the buffer is added or retrieved from the cache.

Also, when a buffer is erased from the cache, maybe you should also remove its path from the file system watcher. Same with `SampleBufferCache::remove()`.

src/core/SampleBufferCache.cpp
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	m_cache.insert(key, buffer);
```
messmerd, line 51: Could also add the path to the file system watcher here.

messmerd (commented):
Looks pretty good except for a few things

src/core/SampleBufferCache.cpp
sakertooth: I don't want to couple the watcher to the cache.

messmerd: Why not? The watcher is needed for the cache to function properly and you shouldn't have to remember to add the watcher in calling code.

sakertooth: SRP, but I think it's okay. I was thinking I couldn't test the cache if I coupled the watcher to it but I can still test without it.

src/core/SampleBufferCache.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+SampleBufferCache::SampleBufferCache(size_t capacity)
+	: m_capacity(capacity)
+{
+	if (capacity == 0) { throw std::runtime_error{"Capacity of cache cannot be zero"}; }
+}
+
+auto SampleBufferCache::add(const QString& key, Buffer buffer) -> bool
+{
+	if (buffer == nullptr || m_cache.contains(key)) { return false; }
```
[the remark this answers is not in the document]
sakertooth: I just realized this breaks all my tests since they use empty buffers.

I guess I will have to delete my tests, but I think there should be a better way. Maybe I will add the caching support directly within the `SampleLoader` create functions.

I don't think an empty buffer should signify a buffer that failed to load, even if we do return one in that case.

src/core/SampleBufferCache.cpp
sakertooth: I think we should just evict the ones that:
1. Aren't used at all
2. Been unused by anything for some time

src/core/SampleBufferCache.cpp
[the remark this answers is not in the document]
messmerd: Maybe you should make the `SampleLoader::createBufferFrom*` methods return nullptr on failure instead of `SampleBuffer::emptyBuffer()`. There doesn't seem to be much that you'd have to change in calling code.

`Sample`'s constructor from a `SampleBuffer` would need to check if it's null and create a sample buffer if so:

```c++
Sample::Sample(std::shared_ptr<const SampleBuffer> buffer)
	: m_buffer(buffer ? std::move(buffer) : std::make_shared<const SampleBuffer>())
	, m_startFrame(0)
	, m_endFrame(m_buffer->size())
	, m_loopStartFrame(0)
	, m_loopEndFrame(m_buffer->size())
{
}
```

Also, there might be a bug in `void SlicerT::updateFile(QString file)` when you move the `std::shared_ptr` from `SampleBuffer::emptyBuffer()`. It could leave the global empty buffer in a moved-from state.

With the changes I suggested above and this potential way to easily misuse the empty buffer, you might be better off removing `SampleBuffer::emptyBuffer()` entirely.

src/core/SampleBufferCache.cpp
Veratil: LRU by design will evict unused when another is loaded. We don't need to implement anything for 2.
Aren't used at all is harder to detect, but in essence is the same as 2, and will be evicted as the cache fills.

I'm debating whether the cache should even be an LRU instead of just an actual buffer cache (single load, multiple references).

sakertooth: > I'm debating whether the cache should even be an LRU instead of just an actual buffer cache (single load, multiple references).

This is what I was trying to say. I dont think LRU is a good eviction policy for this problem. I think it makes more sense just to evict ones that are not used by anyone at all and its been that way for some time (maybe a minute), probably using some timer.

Veratil: No don't use a timer, that's making the problem more complex. Evict when the reference count is 0 as then it's no longer used, `shared_ptr`-like.

src/core/SampleCache.cpp
```diff
@@ -0,0 +1,71 @@
...
+	if (s_entries.contains(filePath)) { return Entry{filePath, s_entries[filePath].lock()}; };
+
+	try
+	{
+		const auto buffer = std::make_shared<const SampleBuffer>(filePath);
+		if (!s_watcher.addPath(filePath)) { return std::nullopt; }
+		s_entries.insert(filePath, buffer);
+		return Entry{filePath, buffer};
+	}
+	catch (const std::runtime_error& error)
```
Veratil, line 60: Is this for catching possible `std::bad_alloc` thrown by `std::make_shared`?

sakertooth: It's for catching the failure to create a buffer from a file.

Veratil: I don't see a throw anywhere is why I ask. Perhaps I'm just missing it.

Veratil: D'oh! I thought it was something added with this PR.

Rossmaxx:
Good feature but I am having a few questions.

Where are the sample caches getting stored? In the disk or in memory? 
If in disk, it would be better to have a `samplecache` folder within the user lmms folder where everything gets cached on loading and tweaks to project files accordingly. 

Also, does this reduce duplication of samples?

sakertooth:
The buffers are cached in memory. Storing them on disk doesn't help at all because the file that we intend to cache after it has been loaded is on disk.

This act of just caching the results after we loaded the sample only helps with speeding up loading said sample into the project. Any memory saving  improvements are only made possible because of the kind of data we are caching. We are caching `shared_ptr`'s to these buffers, which then allows for the buffer to be shared with the cache.

What is really wanted is the sharing of these buffers. This cache only speeds up data retrieval by itself, but there are added RAM improvements only because of what we are caching and how anybody that needs a sample (think AFP, sample tracks, etc) can have some global place to go and get it if it's already been loaded.

This distinction I feel like hasn't been made and is now confusing this PR up by a lot.

sakertooth:
That being said, I have to go and fix this PR up.

src/core/SampleBufferCache.cpp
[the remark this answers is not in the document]
messmerd: Now that you are returning `nullptr` from `SampleLoader::createBufferFromFile()` when it fails (which is great, imo), I think you need to take a look at the call sites to make sure they are handling `nullptr` correctly. If `Sample` assumes its sample buffer is never null, you'll need to fix the `Sample::Sample(std::shared_ptr<const SampleBuffer> buffer)` constructor like I mentioned in my comment above.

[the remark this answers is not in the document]
Spekular: Is there a reason we don't use optional here? Then it would be a conpile-time error to not handle the empty case at call sites.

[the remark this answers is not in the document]
sakertooth: I think it would be better for `SampleCache::get` to just propagate the exception up near some GUI boundary, so that it can display the message box there.

Exceptions provide more detail to an error for the user instead of just returning `nullptr`.

src/core/SampleBufferCache.cpp
sakertooth: > No don't use a timer, that's making the problem more complex. Evict when the reference count is 0 as then it's no longer used, shared_ptr-like.

But when should this eviction happen? Cleaning up the cache periodically simply resembles garbage collection, so I don't think this will drive up complexity at all.

Cleaning up the cache periodically would ensure that the buffers are guaranteed to be removed at some point once no one is using them. If we only do after some capacity that was set initially, a buffer can be added, never used again, and no one decides to max out the cache by adding more items, so it will just be there for nothing.

Removing them right after no one is using them doesn't help either because it may become in use again very soon, so we should allow some time to pass first.

Veratil: > But when should this eviction happen?
> Cleaning up the cache periodically would ensure that the buffers are guaranteed to be removed at some point once no one is using them.

If it's reference counted it will clean itself upon having 0 references, correct? We know when someone is using it because the counter is >0.

sakertooth: > If it's reference counted it will clean itself upon having 0 references, correct? We know when someone is using it because the counter is >0.

Well yeah, I can add a custom deleter to the `shared_ptr` that will not only delete the buffer but also remove itself from the cache, but we forfeit any benefits if that sample will be accessed again after deletion without any changes being made on the filesystem.

Question becomes are we okay with that, and is it really a cache at that point?

messmerd: Maybe you could use the `std::shared_ptr` deleter to postpone the deletion of buffer it it is able to, else just delete it immediately.

Pseudocode:
```c++
// In the Entry class
struct BufferDeleter
{
	void operator()(SampleBuffer* ptr) const
	{
		if (!ptr) { return; }

		struct KeepAlive
		{
			std::shared_ptr<SampleBuffer> buffer;
		};

		auto temp = KeepAlive{ptr};

		// Now attempt to overwrite the Cache's std::weak_ptr value
		// for the filePath() key with `temp.buffer`.
		// While `temp` still exists, it will have sole
		// ownership of the buffer and the Cache will still be able
		// to use the buffer normally, but if time runs out and it is deleted,
		// the Cache will fail to lock the weak_ptr.

		if (/* if failed to overwrite cache */) { return; /* deletes buffer immediately */ }

		// Postpone buffer deletion (I don't know if this would work, but you get the idea)
		QTimer::singleShot(3000, [t = std::move(temp)](){});
	}
};

std::pair<QString, std::shared_ptr<const SampleBuffer, BufferDeleter>> m_entry;
```

src/core/SampleBufferCache.cpp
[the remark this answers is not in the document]
messmerd: This is one of those moments I wish we had `std::expected`

src/core/SampleBufferCache.cpp
sakertooth: Sorry, I was overthinking it. The cache can store `weak_ptr`'s to the buffer, but return `shared_ptr`'s on calls to `SampleCache::get`. When no more `shared_ptr`'s are using that buffer, the `weak_ptr` in the cache will expire. If someone where to request that sample again and the `weak_ptr` has expired, its a cache miss and we have to reload the buffer back in. It's also a cache miss if the entry just does not exist in the cache. Otherwise, its a cache hit.

This is all I have for this class now. I think it works nicely:
```cpp
SampleCache::SampleCache()
{
	QObject::connect(&s_watcher, &QFileSystemWatcher::fileChanged, [](const QString& path) { evict(path); });
}

auto SampleCache::get(const QString& filePath) -> std::shared_ptr<const SampleBuffer>
{
	if (s_entries.contains(filePath) && !s_entries[filePath].expired()) { return s_entries[filePath].lock(); }
	s_entries[filePath] = std::make_shared<const SampleBuffer>(filePath);
	return s_entries[filePath].lock();
}

void SampleCache::evict(const QString& filePath)
{
	s_entries.remove(filePath);
	s_watcher.removePath(filePath);
}
```

Veratil: Perhaps we should move the QFileSystemWatcher to the GUI side and connect it to the cache there?

sakertooth: > Perhaps we should move the QFileSystemWatcher to the GUI side and connect it to the cache there?

My initial idea was that the watcher would be separate from the cache, but @messmerd believed it should be together because you shouldn't forget to add the path to the watcher once you add it to the cache, which I agree with.

`QFileSystemWatcher` I dont think interferes with the core and GUI split. If you think about it, we can always replace the functionality of `QFileSystemWatcher` with another library that by no means is realted to the GUI. This is mostly functionality with watching changes on the filesystem, so it has nothing to do with the GUI.

But in terms of not using Qt in the core, it could be a problem, but not as much because in this case the functionality is so easily replaced with another standalone filesystem watcher library, so we can always change it when we are ready and it wont be a major blocking Qt dependency.

messmerd: @sakertooth 

Doesn't the code you posted delete buffers immediately once they're no longer used?
You said before:
> Removing them right after no one is using them doesn't help either because it may become in use again very soon, so we should allow some time to pass first.

Also, the `SampleCache::get()` method from the code you posted has a race condition with its `std::weak_ptr::expired()` usage, it searches `s_entries` twice rather than once, and I think it has some invalid behavior when assigning a `std::shared_ptr` temporary with sole ownership to the non-owning `std::weak_ptr`.

Here's something potentially better, though completely untested:
```c++
auto SampleCache::get(const QString& filePath) -> std::shared_ptr<const SampleBuffer>
{
	if (auto it = s_entries.find(filePath); it != s_entries.end())
	{
		if (auto buffer = it->lock()) { return buffer; }
	}

	auto buffer = std::make_shared<const SampleBuffer>(filePath);
	s_entries[filePath] = buffer;
	return buffer;
}
```

P.S. I will probably end up writing my own replacement for `QFileSystemWatcher` eventually since the CLAP host will need it.

sakertooth: > Doesn't the code you posted delete buffers immediately once they're no longer used?

I'm debating on potential implementations. Do we want to evict buffers immediately when they are not used? This is what @Veratil was originally thinking we should do, and I'm not entirely against it. I only mentioned that we would forgo any retrieval speed benefits if that same sample will be feteched again in a short amount of time after eviction, but we still get the benefits of:

1. Fast load times if the buffer is being used and is within the cache
2. Memory savings if the same conditions are met in point 1

So it is just a question of if we want to forgo that benefit in the intrest of making the implementation simpler.

> Also, the `SampleCache::get()` method from the code you posted has a race condition with its `std::weak_ptr::expired()` usage, it searches `s_entries` twice rather than once, and I think it has some invalid behavior when assigning a `std::shared_ptr` temporary with sole ownership to the non-owning `std::weak_ptr`.

To be honest, I was not considering thread safety into this because I could only ever envision the GUI thread requesting samples from the cache. What other thread? When the user wants a sample, it must be from the GUI, no?

And for assigning an r-value `shared_ptr` to ` weak_ptr`, yeah. I originally stored it as an l-value and returned it directly (while also assigning it to the `weak_ptr`) but I couldn't see what was wrong with that at first.

sakertooth: I understand your concerns better @messmerd. It actually does search twice, but its `O(2)` and not `O(n^2)` so its not a major concern, but still a useful optimization. Also, you're right that saying `!s_entries[filePath].expired()` is wrong assuming there are multiple threads because the `weak_ptr` may become expired right after we decided it wasn't, but it would only be a concern if multiple threads try and share the buffer, which I don't see happening?

I'll still apply the optimizations/fixes tough, along with some other things. I also decided that we should evict buffers immediately when they are no longer in use because the benefits we get from keeping that buffer for a minute or so isn't worth making the implementation more intricate. I'm still willing to negotiate different ideas though, but the two main benefits I listed above I feel is enough.

sakertooth:
Closing in favor of #7058.

PR #6705, opened 2023-05-08
Fix STK rawwave installation in Windows builds

### LINKED ISSUE #6464: mallets broken on master build

PhysSong:
Fixes #6464.
This PR also makes the rawwave installation handled by CMake scripts on macOS.


superpaik (approved):

tresf (approved):
Tested on MacOS, `rawwaves` exists, has the relevant `.raw` files, Mallets tested and working.

tresf:
> Fixes #6464. This PR also makes the rawwave installation handled by CMake scripts on macOS.

Curious, why not do the same for Linux?

https://github.com/LMMS/lmms/blob/b322f8022e9873ea0d09a7e3bb5cf785605fa793/cmake/linux/package_linux.sh.in#L110-L112

PhysSong:
> Curious, why not do the same for Linux?

Because LMMS uses system-wide installation of rawwaves when installed locally, and I didn't wanted to change that behavior. Please suggest me any ideas, if you have one.

tresf:
> Because LMMS uses system-wide installation of rawwaves when installed locally

I had forgotten about local installs.

> Please suggest me any ideas, if you have one.

This is a good question.  I suppose the installation would be copying over itself for local installs, however when using a custom install prefix, it would not be redundant. 

I suppose it's justified to keep this task in the `.AppImage` script for now, however it may be beneficial as a long -term goal to accommodate this in CMake.

Issue #6332, opened 2022-03-16
Hidden relationship between 'record-buttons' and Q-value.


musikBear:
### Enhancement Summary
I here suggest that we change Q-value to 1/192 automatically, if either of the two recording-buttons are clicked

#### Justification
Ooops i did it again.
'It' being played some really nice chords, and even had record invoked, so all should be good, but.. alas
Q-value was 1/8, and not a single chord was added as i played.
That made think. 
When either of the record-buttons is selected, wouldent best Q not in most circumstances be 1/192, eg LMMS' best free placing of notes?

Atm user need to remember to set the Q value to 1/192, and then start recording.
I forget that almost every time.
I think the few times where non-human placing of notes actually is wanted, should be the one where a special selection of Q-value also should be done. At that situation the mindset is focused on a special Q-value, and it would then be intuitive to also sét that value.
A inconvenience do exists if the user first set the Q he want, and then set record, because with my idea, LMMS would then revert to 1/192.
The user need to first set the record-option and then his preferred Q, someone may dislike that the sequence play a role.
Share Any thoughts on how that could be fixed.

#### Mockup
Not really possible to mock-up this functionality. LMMS already has the controls that are needed for this.
But here 
![image](https://user-images.githubusercontent.com/6662066/158695347-0fd6dfb0-70bd-43a8-af19-63b18bae6294.png)
are the buttons marked, that i want to 'link'

Spekular:
Rather than changing anything in the UI, I think recording should bypass the quantization setting entirely. If users want quantized recordings they can then press the quantize button.

musikBear:
> Rather than changing anything in the UI, I think recording should bypass the quantization setting entirely. If users want quantized recordings they can then press the quantize button.

Yes, that is also a way, as long as recording always pr default insert notes as they are being played, i would be fine with that as well 👍

PR #6714, opened 2023-05-19
Don't auto-quantize notes when recording MIDI input

irrenhaus3:
Relates to issue #6332.
I regularly ran into trouble with the piano roll quantization when recording a MIDI pattern from my controller (meaning I regularly forgot that I sometimes had to manually set quantization to 1/192 before hitting record), which led to me having to throw my pattern away and redo it. Pretty frustrating.
This PR makes only one change: When recording MIDI input in one of the two realtime modes, neither note positions nor lengths are quantized automatically. Makes life a lot easier, so I think it's a worthy change to merge upstream.


superpaik:
Sorry I didn't participate in the discussion of the topic, but what about having a variable in the MIDI Setting section so that the user can set the default behavior according to their preferences? Many people can't play the piano proficiently, and I think quantization is a great help (it's true that you can quantize later, but it might not be seen as a good alternative for some people).

irrenhaus3:
> it's true that you can quantize later, but it might not be seen as a good alternative for some people

I'm not a good piano player myself (like, at all, I just hit chords along with the song playback :smile:), but for me, the mode of recording freeform first and then moving & quantizing where needed is the only one that feels "right". Everything else feels like the software getting in the way instead of helping me with what I want to do.    
If we want a toggle for auto-quantization in the MIDI settings, I'd strongly advocate for the default to be "Off".

superpaik:
I'm not against having it to "Off" by default, but I think that for some people having auto-quantize while recording can be seen as a positive feature.

YohananDiamond:
Whatever is the default option, it'd be good if a per-user default value could be configured via the settings. Same with other options like follow-play-marker but I digress.

irrenhaus3:
I've added a checkbox for this in the MIDI settings. I'm not sure if that's the right place though, kinda feels like this should be a button in the piano roll editor.

superpaik:
> I've added a checkbox for this in the MIDI settings. I'm not sure if that's the right place though, kinda feels like this should be a button in the piano roll editor.

To me is ok to have it in MIDI settings, as per-user configuration better than as per-record configuration (I don't know if that makes sense ;-)

src/gui/editors/PianoRoll.cpp
```diff
@@ -4137,8 +4137,15 @@ void PianoRoll::finishRecordNote(const Note & n )
 					Note n1( n.length(), it->pos(),
 							it->key(), it->getVolume(),
 							it->getPanning() );
-					n1.quantizeLength( quantization() );
-					m_midiClip->addNote( n1 );
+
+					// TODO this value is probably better off cached somewhere;
+					//  currently doing a string parse everytime a note is released.
+					bool doQuantize = (ConfigManager::inst()->value("midi", "autoquantize").toInt() != 0);
+					if (doQuantize) {
```
sakertooth, line 4144:
```suggestion
					if (doQuantize)
					{
```

src/gui/editors/PianoRoll.cpp
```diff
@@ -4137,8 +4137,15 @@ void PianoRoll::finishRecordNote(const Note & n )
...
+						n1.quantizeLength(quantization());
+						n1.quantizePos(quantization());
+					}
+					m_midiClip->addNote(n1,false);
```
sakertooth, line 4148:
```suggestion
					m_midiClip->addNote(n1, false);
```

src/gui/modals/SetupDialog.cpp
```diff
@@ -723,10 +724,18 @@ SetupDialog::SetupDialog(ConfigTabs tab_to_open) :
 		m_assignableMidiDevices->setCurrentIndex(current);
 	}
 
+	// MIDI Recording tab
+	auto* midiRecordingTab = new TabWidget(tr("Behavior when recording"), midi_w);
+	int mwCounter=0;
```
sakertooth, line 729:
```suggestion
	int mwCounter = 0;
```

src/gui/modals/SetupDialog.cpp
```diff
@@ -723,10 +724,18 @@ SetupDialog::SetupDialog(ConfigTabs tab_to_open) :
...
+	addLedCheckBox(tr("Auto-quantize notes in Piano Roll"), midiRecordingTab, mwCounter,
+		m_midiAutoQuantize, SLOT(toggleMidiAutoQuantization(bool)), false);
```
sakertooth, line 731: I wonder if you could use a function pointer here instead of the ``SLOT`` macro?

src/gui/editors/PianoRoll.cpp
```diff
@@ -4137,8 +4137,15 @@ void PianoRoll::finishRecordNote(const Note & n )
 					Note n1( n.length(), it->pos(),
 							it->key(), it->getVolume(),
 							it->getPanning() );
-					n1.quantizeLength( quantization() );
-					m_midiClip->addNote( n1 );
+
+					// TODO this value is probably better off cached somewhere;
+					//  currently doing a string parse everytime a note is released.
+					bool doQuantize = (ConfigManager::inst()->value("midi", "autoquantize").toInt() != 0);
```
sakertooth, line 4143: Probably should make this a member of ``PianoRoll`` and make sure it gets updated with ``ConfigManager::valueChanged`` somewhere.
PS: You don't need the parentheses if you could plan on keeping this code.

src/gui/modals/SetupDialog.cpp
irrenhaus3: The addLedCheckbox convenience function takes a `const char*` resolved SLOT there, so a std::function / lambda would make this more messy. However, this is in conflict with #6779, which removes the LedCheckboxes from this dialog, so I could adapt this part in preparation for that merge.

sakertooth: Just tested it, yeah. It could work, but you would probably have to change the type from ``const char*`` to a function pointer type, which may break the other ones.

src/gui/editors/PianoRoll.cpp
irrenhaus3: Implemented and ready for testing.

sakertooth (approved):
Tested and working. Could easily see the improvement made at lower resolutions. Automatic updates of the auto quantization option also works as expected.

src/gui/editors/PianoRoll.cpp
```diff
@@ -248,6 +249,15 @@ PianoRoll::PianoRoll() :
...
 	connect( unmarkAllAction, &QAction::triggered, [this](){ markSemiTone(stmaUnmarkAll); });
 	connect( copyAllNotesAction, &QAction::triggered, [this](){ markSemiTone(stmaCopyAllNotesOnKey); });
+	connect( ConfigManager::inst(), &ConfigManager::valueChanged,
+		[this](QString const& cls, QString const& attribute, QString const& value)
+		{
+			if (cls != "midi" && attribute != "autoquantize")
+			{
+				return;
+			}
+			this->m_doAutoQuantization = (value.toInt() != 0);
```
IanCaio, line 259: Same suggestion as before, again, without being sure whether it has been decided to avoid implicit conversions. 
```suggestion
			this->m_doAutoQuantization = value.toInt();
```

src/gui/modals/SetupDialog.cpp
```diff
@@ -723,10 +724,18 @@ SetupDialog::SetupDialog(ConfigTabs tab_to_open) :
 		m_assignableMidiDevices->setCurrentIndex(current);
 	}
 
+	// MIDI Recording tab
+	auto* midiRecordingTab = new TabWidget(tr("Behavior when recording"), midi_w);
+	int mwCounter = 0;
+	addLedCheckBox(tr("Auto-quantize notes in Piano Roll"), midiRecordingTab, mwCounter,
```
IanCaio, line 730: It doesn't look necessary to use a counter here because there's a single CheckBox but if we are using wouldn't it be more consistent to the rest of the code to reuse `counter` as well?
```suggestion
	counter = 0;
	
	// MIDI Recording tab
	auto* midiRecordingTab = new TabWidget(tr("Behavior when recording"), midi_w);
	addLedCheckBox(tr("Auto-quantize notes in Piano Roll"), midiRecordingTab, counter,
```

IanCaio (changes requested):

src/gui/editors/PianoRoll.cpp
```diff
@@ -168,6 +168,7 @@ PianoRoll::PianoRoll() :
 	m_midiClip( nullptr ),
 	m_currentPosition(),
 	m_recording( false ),
+	m_doAutoQuantization(ConfigManager::inst()->value("midi", "autoquantize").toInt() != 0),
```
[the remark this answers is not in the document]
irrenhaus3: Preferring explicit test for better readability.

src/gui/editors/PianoRoll.cpp
irrenhaus3: Preferring explicit test for better readability.

src/gui/editors/PianoRoll.cpp
```diff
@@ -247,6 +248,15 @@ PianoRoll::PianoRoll() :
 	connect( markChordAction, &QAction::triggered, [this](){ markSemiTone(SemiToneMarkerAction::MarkCurrentChord); });
 	connect( unmarkAllAction, &QAction::triggered, [this](){ markSemiTone(SemiToneMarkerAction::UnmarkAll); });
 	connect( copyAllNotesAction, &QAction::triggered, [this](){ markSemiTone(SemiToneMarkerAction::CopyAllNotesOnKey); });
+	connect(ConfigManager::inst(), &ConfigManager::valueChanged,
+		[this](QString const& cls, QString const& attribute, QString const& value)
+		{
+			if (cls != "midi" && attribute != "autoquantize")
```
irrenhaus3, line 254: Logic bug; should be OR(~A,~B), or NAND(A,B), but not AND(~A,~B).
Condition is false if cls=="midi", even it attribute doesn't match, and vice versa, so we skip over the early return more often than intended.

irrenhaus3 (commented):
Found a bug.

src/gui/editors/PianoRoll.cpp
irrenhaus3: Fixed with 8c977c9

IanCaio (approved):
Code-wise it LGTM, I didn't manage to test it yet though.

zonkmachine (approved):
Tested, works well!

sakertooth (approved):
I'm okay with the new changes :+1:

PR #6717, opened 2023-05-23
Adding new FX channel while in solo 

### LINKED ISSUE #6311: FX-Mixer behaviour- adding a bus mute behaviour while one is solo'ed
### Enhancement Summary
Adding a mixer bus while one of the fx mixer channels is solo'ed, shouldn't reveal the new channel to be muted upon un-soloing.

#### Steps to reproduce (issue)
1. Create a channel or two in FX-Mixer
2. Solo one of the channels
3. Add a new channel (either by clicking the + or by Context Menu of Instrument > Assign to new FX-Mixer)
4. Un-solo the channel
5. Check the channels' muting

#### Justification
Adding a new mixer channel while a mixer channel is already solo'ed, leads to the new channel being muted once we un-solo the previous one. This is because manually unmuting a channel while another is solo'ed lets it play too, but the actual state of a new channel should be unmuted even if we add it while another is solo'ed.
This is as opposed to Song Editor, where upon adding a track while one is soloed it is unmuted, allowing it to be heard. However, when adding a track in SE, it is typically empty enough that mute/unmute doesn't make a difference in output.

I wasn't sure if this was an enhancement request or a bug, because while this is unintended behavior, the way it works prevents a different unintended behaviour from occurring instead.

superpaik:
Fixes #6311

When adding a new FX channel (A) and there is a solo'ed FX channel (B), that one (A) is muted. 
So when (B) is un-soloed (A) is un-muted.

### DIFF
--- a/src/core/Mixer.cpp
+++ b/src/core/Mixer.cpp
@@ -192,9 +192,9 @@ Mixer::Mixer() :
 	JournallingObject(),
 	m_mixerChannels()
 {
+	m_lastSoloed = -1;
 	// create master channel
 	createChannel();
-	m_lastSoloed = -1;
 }
 
 
@@ -223,6 +223,13 @@ int Mixer::createChannel()
 
 	// reset channel state
 	clearChannel( index );
+
+	// if there is a soloed channel, mute the new track
+	if (m_lastSoloed != -1 && m_mixerChannels[m_lastSoloed]->m_soloModel.value())
+	{
+		m_mixerChannels[index]->m_muteBeforeSolo = m_mixerChannels[index]->m_muteModel.value();
+		m_mixerChannels[index]->m_muteModel.setValue(true);
+	}
 
 	return index;
 }
--- a/src/core/Mixer.cpp
+++ b/src/core/Mixer.cpp
@@ -190,9 +190,9 @@ void MixerChannel::doProcessing()
 Mixer::Mixer() :
 	Model( nullptr ),
 	JournallingObject(),
-	m_mixerChannels()
+	m_mixerChannels(),
+	m_lastSoloed(-1)
 {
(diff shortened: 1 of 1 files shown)


src/core/Mixer.cpp
```diff
@@ -192,9 +192,9 @@ Mixer::Mixer() :
 	JournallingObject(),
 	m_mixerChannels()
 {
+	m_lastSoloed = -1;
```
PhysSong, line 195: Since you moved this line above `createChannel`, can you move this to the initializer list of this constructor as well?

superpaik: Done

src/core/Mixer.cpp
```diff
@@ -224,6 +224,13 @@ int Mixer::createChannel()
 	// reset channel state
 	clearChannel( index );
 
+	// if there is a soloed channel, mute the new track
+	if (m_lastSoloed != -1 && m_mixerChannels[m_lastSoloed]->m_soloModel.value())
+	{
+		m_mixerChannels[index]->m_muteBeforeSolo = m_mixerChannels[index]->m_muteModel.value();
```
PhysSong, line 230: I think `m_mixerChannels[index]->m_muteModel.value()` will always be `false` unless we put something between `clearChannel(index)` and here.

superpaik: Yes, that's true. But because it's set on the function clearChannel I wanted to use its value rather than presume it's false (even though we now is false) But maybe I'm overthinking things ;-)
If you think it's clearer/better to set it to false directly, I can change the code. No problem.

PhysSong: If you have such an intention, no problem.

PhysSong (approved):

PR #6718, opened 2023-05-26
Fix automation crash

### LINKED ISSUE #6711: Comboboxes in SongEditor allow automation but crash
### Bug Summary
Comboboxes for zooming and snapping in SongEditor allow automation, but when they are drag to an automation track LMMS crashes.

#### Steps to reproduce
Drag zoom combobox arrow in SongEditor to an automation track

#### Expected behavior
I don't see why those controllers (that are gui controllers) should be automatable, but if they are, they shouldn't cause a crash.
Similar controllers in PianoRoll or AutomationEditor do not cause a crash.

#### Actual behavior
It causes a crash.

#### Screenshot
DrowDown-Automate-Crash

#### Affected LMMS versions
1.2.2 and master

Hint:
As far as I found, it happens because in model.cpp we try to get the parentModel()->fullDisplayName(), and those controls' parents do not have that property or something. The string is not null and is not empty, but when accessing to it, it generates an access violation.

michaelgregorius:
Fix for the crash that occurs when the zooming or snapping model of the Song Editor is used for automation.

Resolves #6711

### DIFF
--- a/include/Model.h
+++ b/include/Model.h
@@ -37,35 +37,18 @@ class LMMS_EXPORT Model : public QObject
 {
 	Q_OBJECT
 public:
-	Model( Model * _parent, QString _display_name = QString(),
-					bool _default_constructed = false ) :
-		QObject( _parent ),
-		m_displayName( _display_name ),
-		m_defaultConstructed( _default_constructed )
-	{
-	}
+	Model(Model * _parent, QString _display_name = QString(),
+		  bool _default_constructed = false );
 
 	~Model() override = default;
 
-	bool isDefaultConstructed()
-	{
-		return m_defaultConstructed;
-	}
+	bool isDefaultConstructed() const;
 
-	Model* parentModel() const
-	{
-		return dynamic_cast<Model *>( parent() );
-	}
+	Model* parentModel() const;
 
-	virtual QString displayName() const
-	{
-		return m_displayName;
-	}
+	virtual QString displayName() const;
 
-	virtual void setDisplayName( const QString& displayName )
-	{
-		m_displayName = displayName;
-	}
+	virtual void setDisplayName(const QString& displayName);
 
 	virtual QString fullDisplayName() const;
 
--- a/include/Model.h
+++ b/include/Model.h
@@ -37,8 +37,8 @@ class LMMS_EXPORT Model : public QObject
 {
 	Q_OBJECT
 public:
-	Model(Model * parent, QString displayName = QString(),
-		  bool defaultConstructed = false );
+	Model(Model* parent, QString displayName = QString(),
+		  bool defaultConstructed = false);
 
 	~Model() override = default;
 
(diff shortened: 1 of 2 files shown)


lukas-w (approved):

superpaik:
Coding style: I think you should remove underscore on function parameters.
Other than that it works ok.

michaelgregorius:
Underscores and whitespace has been removed with commit fda938fc.

I have also simplified the logic of `Model::fullDisplayName` with commit 9fe7fbd6. This might also have made more obvious a potential problem if the parent has a non-empty name "xyz" but the child has an empty one. This will result in the name "xyz >" being returned.

include/Model.h
```diff
@@ -37,35 +37,18 @@ class LMMS_EXPORT Model : public QObject
...
 public:
-	Model( Model * _parent, QString _display_name = QString(),
-					bool _default_constructed = false ) :
-		QObject( _parent ),
-		m_displayName( _display_name ),
-		m_defaultConstructed( _default_constructed )
-	{
-	}
+	Model(Model * parent, QString displayName = QString(),
+		  bool defaultConstructed = false );
```
messmerd, line 41:
```suggestion
	Model(Model* parent, QString displayName = QString(),
		  bool defaultConstructed = false);
```

src/core/Model.cpp
```diff
@@ -27,22 +27,47 @@
 namespace lmms
 {
 
+Model::Model(Model * parent, QString displayName, bool defaultConstructed) :
```
messmerd, line 30:
```suggestion
Model::Model(Model* parent, QString displayName, bool defaultConstructed) :
```

src/core/Model.cpp
```diff
@@ -27,22 +27,47 @@
...
+}
+
+bool Model::isDefaultConstructed() const
+{
+	return m_defaultConstructed;
+}
+
+Model* Model::parentModel() const
+{
+	return dynamic_cast<Model *>(parent());
```
messmerd, line 44:
```suggestion
	return dynamic_cast<Model*>(parent());
```

src/core/Model.cpp
```diff
@@ -27,22 +27,47 @@
...
+}
+
+QString Model::displayName() const
+{
+	return m_displayName;
+}
+
+void Model::setDisplayName( const QString& displayName )
```
messmerd, line 52:
```suggestion
void Model::setDisplayName(const QString& displayName)
```

src/core/Model.cpp
```diff
@@ -27,22 +27,47 @@
...
+	m_displayName = displayName;
+}
+
 QString Model::fullDisplayName() const
 {
-	const QString & n = displayName();
-	if( parentModel() ) 
+	const QString n = displayName();
+
+	if(parentModel())
```
messmerd, line 61:
```suggestion
	if (parentModel())
```

messmerd (approved):
Just some very minor style suggestions.
Everything else looks good to me and it worked when I tested it. I've also confirmed that the logic you simplified is still correct.

michaelgregorius:
Thanks for the code review and double check @messmerd! I have implemented all your proposals with commit 0ebc1894.

PR #6723, opened 2023-06-04
Fix uninitialized variable "pCurPreset"

### LINKED ISSUE #6639: 'pCurPreset' is being used without being initialized.
Run-Time Check Failure - The variable 'pCurPreset' is being used without being initialized.

https://github.com/LMMS/lmms/blob/507fa239ddffd2ed1f186d0a51016c85d9c2b5a9/plugins/Sf2Player/PatchesDialog.cpp#L159

How to reproduce:
Open SF player. Choose a soundfont file. Choose wrench icon: "Choose patch".

Debugger stops at this point. After this stoppage, if you restart lmms:
`Cannot load library -\build\Debug\plugins\sf2player.dll: The specified module could not be found.` and the player won't show up in the instrument list next time.

michaelgregorius:
Fix the uninitialized variable `pCurPreset` by setting it to `nullptr`

Please note that it looks as if the `nullptr` is now being passed to the `function fluid_sfont_iteration_next_wrapper`. However, the function does not use the parameter anyway and then the variable is set to the result of that function.

Resolves #6639

### DIFF
--- a/plugins/Sf2Player/PatchesDialog.cpp
+++ b/plugins/Sf2Player/PatchesDialog.cpp
@@ -154,7 +154,7 @@ void PatchesDialog::setup ( fluid_synth_t * pSynth, int iChan,
 			fluid_preset_t preset;
 			fluid_preset_t *pCurPreset = &preset;
 #else
-			fluid_preset_t *pCurPreset;
+			fluid_preset_t *pCurPreset = nullptr;
 #endif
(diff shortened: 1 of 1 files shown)


irrenhaus3 (approved):

Veratil (approved):
Not needed because as said it's unused and assigned on the next line, but doesn't hurt to be explicit here.

PR #6725, opened 2023-06-05
Fix automated base notes and their automations (#6548)

### LINKED ISSUE #6548: Audio file transposed an octave down between versions
### Bug Summary
On 1.2, I created a project, and have used an ogg file sample in the beat/bassline editor.
On 1.3, trying to edit the project, it is transposed up by an octave (the basenote is shifted down).

I can't seem to reproduce this with the same sample file, yet, but the bug is constant on the mmp file. So, I have attached the mmp file and mentioned where the problem is.

#### Steps to reproduce
1. Extract this zipped folder's files:
[Puppet3 AudioBug.zip](https://github.com/LMMS/lmms/files/9846844/Puppet3.AudioBug.zip)
Do not open the file yet, as you will get an error because the sample is missing.
2. In the mmp file within the zipped folder, there is a single sample file leading to C:/users/username/mp3 mb/Snap.ogg
In a plaintext editor, edit this path to lead wherever you have placed Snap.ogg from the zipped folder on your system. This step is **not needed**, though: you can just check the position of the basenote with your eyes as it is seen even there. This is simply if you want to hear the difference.
3. Save it, and open in LMMS 1.2. In the B/B editor, check the sound of the snap by clicking the preview space, or play back a b/b with the snap used on a step. Alternatively, look at the position of the basenote on the Snap in the AudioFileProcessor (See Screenshots).
image
4. Now open it on LMMS 1.3 alpha and do the same.
5. Hear difference.

I cannot yet reproduce this on a blank slate. But it is reproducible within this file.

#### Expected behavior
They sound the same

#### Actual behavior
In 1.3, basenote is transposed an octave down so we hear file an octave up.

#### Screenshot
image
A5 vs A4.

#### Affected LMMS versions
1.2.0/1.2.2 -> 1.3 alpha

michaelgregorius:
Fix all base notes that are used in automations and their corresponding automation values. Base notes that are automated are stored as elements in the save file whereas non-automated base notes are stored as attributes. So far the method `upgrade_extendedNoteRange` only upgraded the non-automated base notes that are stored in attributes. This commit implements the upgrades of the automated ones which are stored in elements.

The fix works as follows:
* Collect all base note elements.
* Store their ids in a set so that we can later identify automations that reference them.
* Collect all automation pattern and check if they reference a base note.
* Adjust the values and out values of all automations that reference base notes.

Note: for many older files the out values will be introduced by the upgrade `method upgrade_automationNodes` and do not appear in the files themselves!


Veratil:
Will the upgrade function still be run if someone already had their project updated? I'm pretty sure our upgrade functions don't run on projects twice?

Spekular:
If it's in the same function the upgrade won't re-run. Breaking it out into a separate function would make it run, at the cost of DataFile.cpp being permanently uglier.

However, I'm not sure how likely it is that someone would upgrade to a 1.3 alpha/nightly, discover that instruments are playing in the wrong octave, and save over the file without fixing it. If they go back to the 1.2 project file, the fixed upgrade routine will run whenever they next try switching to 1.3. If they move to 1.3 anyway and manually fix the octave, the upgrade routine would be redundant. So maybe it's fine if it doesn't re-run?

Veratil:
> If it's in the same function the upgrade won't re-run. Breaking it out into a separate function would make it run, at the cost of DataFile.cpp being permanently uglier.

I don't think the files' beauty matters when we're trying to make sure things are fixed. 😉

> However, I'm not sure how likely it is that someone would upgrade to a 1.3 alpha/nightly, discover that instruments are playing in the wrong octave, and save over the file without fixing it.

I agree, but there are a number of people who have been working with 1.3 downloads. I wouldn't be surprised if there is more than one person that had this happen to their project.

> If they go back to the 1.2 project file, the fixed upgrade routine will run whenever they next try switching to 1.3. If they move to 1.3 anyway and manually fix the octave, the upgrade routine would be redundant. So maybe it's fine if it doesn't re-run?

That's what I'm trying to decide, do we want to just say "if you already opened past this commit then it's up to you to fix" or "we want to maintain proper upgrade routines no matter what".

he29-net:
I think the code may be missing one important step: checking whether the automation track drives a base note of an instrument affected by the original non-standard MIDI range (#1857), or some other instrument.

If the automated base note belongs to an affected instrument (see `auto affected = [](const QDomElement& instrument)` in the beginning of the update routine), it should not be touched. It already sounded an octave lower in <= 1.2.2, so it must stay the same in 1.3.x. Automation of all other instruments, on the other hand, should be updated to counteract the change in base note number.

And it gets even more unpleasant: if an automation track drives multiple base notes, it may be necessary to create two versions: one corrected, and one unchanged. The `affected` instruments should stay connected to the original, and the others should be re-connected to the new, corrected version.

There may also be some extreme edge cases, like if someone uses one automation track to drive several unrelated targets, like a base note and volume, or perhaps more realistically, base note and Master pitch or something. Hopefully, if changes are only ever made to a new copy of the automation track, I suppose it should leave all such "unreasonable scenarios" reasonably intact... Otherwise I don't even want to guess what to do with such cases, other than perhaps make sure the updated base note won't fall of range.

Spekular:
> I don't think the files' beauty matters when we're trying to make sure things are fixed. 😉

I'm mostly thinking about the precedent of going out of our way to support backwards compatibility between unstable versions. Of course 1.3 should properly upgrade 1.2 files, but how reasonable is it really to expect 1.3-[long commit hash a] to upgrade a 1.3-[long commit hash b] file that was saved in a broken state?

> That's what I'm trying to decide, do we want to just say "if you already opened past this commit then it's up to you to fix" or "we want to maintain proper upgrade routines no matter what".

I don't mean to compare "proper" vs "unproper"; what I wanted to point out was that any projects a user has *already* fixed after upgrading would be broken if the base note is shifted again by a new upgrade routine.

Veratil:
> I'm mostly thinking about the precedent of going out of our way to support backwards compatibility between unstable versions.

I was under the impression that we were moving forward with the idea that master is a "stable" version? People could expect that master will be kept as stable as possible.

> Of course 1.3 should properly upgrade 1.2 files, but how reasonable is it really to expect 1.3-[long commit hash a] to upgrade a 1.3-[long commit hash b] file that was saved in a broken state?

As reasonable as we want, honestly. There's old upgrade routines that are literally days between.

> I don't mean to compare "proper" vs "unproper"; what I wanted to point out was that any projects a user has _already_ fixed after upgrading would be broken if the base note is shifted again by a new upgrade routine.

This is a good point I didn't think of. So really either way we go we have the possibility of breaking something.

he29-net:
@michaelgregorius
>Can an automation track drive more that one modulation target? If yes, how can I set this up in LMMS to create a test file?

That should be easy to set up; just hold Ctrl and drag multiple controls, one after another, to the same automation clip. Then if you right-click on the clip, there should be an "# Connections →" entry, which contains a list of all connected controls.

>`QDomElement instrument = instruments.item(0).toElement();`
Can each track only have one instrument beneath it? Or is this a potential bug?

I am not 100 % sure, but I personally have no idea how I would put two instruments into one instrument track. I don't think it is possible in the GUI, so my guess is that if you try to create such an instrument track by editing the XML file, only the first instrument will play, or something breaks. Unless multiple instruments in one (non-BB / non-Pattern) track were supported in some ancient LMMS version I'm not aware of, and the core still happens to supports it...

>Depending on what `instrument.parentNode().toElement().attribute("basenote").toInt() + 12` does for a non-existing attribute "basenote" the example XML might have two basenotes after the fix: one attribute "basenote" and one element.

That's a good find; I had no idea that the `basenote` attribute could sometimes be missing and / or become an element instead, so I did not take that into account. I don't really know much about the internals of automation, so I'm not sure what would happen when two different `basenote` entries are present. Probably nothing good. ^^;

Given this oversight could lead to some potential weirdness, I would personally lean on the side of fixing and extending the current `DataFile::upgrade_extendedNoteRange` function, rather than creating a new upgrade routine just for the automation fix. But I have no insight to the number of users that may have already upgraded their projects from 1.2.2 to 1.3.x alpha, so I can't really help with deciding which approach would be less disruptive..

Spekular:
> If we decided to go with a new version please advise on how to do so. Adding the new method at the end of `DataFile::UPGRADE_METHODS` should be straight forward. But how is the version number itself bumped during the upgrade?

This is handled automatically in `DataFile::upgrade()` with the help of the `UPGRADE_METHODS` vector as of https://github.com/LMMS/lmms/pull/5660.

https://github.com/LMMS/lmms/blob/9f34c5cfa336a03d78a8783829357e15cd84107b/src/core/DataFile.cpp#L1915

michaelgregorius:
@he29-net, thanks to your help I was able to setup a more complex example file. With that file I now also came to the realization that a complete fix would indeed include a potential duplication of automation patterns into ones that contain unaffected elements and ones that contain affected ones.

I also found some other potential problems with that example file. Modulated elements, e.g. base notes, do not seem to be reset to their nominal values after playing. This means that if you modulate a base note with a linear automation, play the song and save the file then the base note is very likely saved with a `float` value. In my example file there's a "basenote" with a value of `57.200001`. I think this could create problems when they are cast to `int`, corrected and then written to the attributes/elements.

I also wonder why BB tracks are ignored? In my example file I have a BB track that container with a TripleOscillator that has an automated base note.

And why are the base notes fixed unconditionally but the notes are fixed conditionally?

Another potential problem in the upgrade method is that there are nested for loops that all have `i` as their running variable.

I agree that it might be better to fix `DataFile::upgrade_extendedNoteRange` and to not introduce a new version. If people work with alpha versions and/or development builds they should expect that things might break and not save over their save files saved with a stable version.

he29-net:
>And why are the base notes fixed unconditionally but the notes are fixed conditionally?

The notes are fixed so that the their labels in piano roll match the tones that will be played (assuming default 12 EDO tuning). Otherwise the update routine could have been be done by only changing base notes -- but the labels would be off.

Basically, the base notes are shifted one octave up, making _everything_ sound one octave lower. Then, notes of affected instruments are kept without change in order to preserve the buggy <= 1.2.2 behavior (while reflecting the fact they sound an octave lower in the label, i.e., old LMMS note label 57 = A4 becomes standard MIDI 57 = A3), while all the other notes are shifted one octave higher to counteract the previous unconditional base note change.

As for the ignored BB tracks, I think it's because the "container track" (that is being ignored according to the comment) does not contain any instruments. It just contains other, "proper" tracks, that are processed by the subsequent code.

michaelgregorius:
@he29-net, @Veratil, @Spekular,

I have pushed some more commits which fix the B+B tracks and which also implement the splitting of mixed automation tracks.

If an automation tracks automates base notes and other targets it is now split into two tracks. The first one contains all the automation for the other targets and the second one the corrected automation for all base note targets.

If an automation does not contain any base note automations it is not touched. If it only contains base note automations it is fixed in place.

I have used the following file to test the fixes: [MultiTargets-1.2.zip](https://github.com/LMMS/lmms/files/11688200/MultiTargets-1.2.zip). If you play that file in the current master you will find that most tracks sound different than in LMMS 1.2.2. If you load it with this branch it should sound exactly the same. The sound/behavior of presets has not been changed between this branch and master.

Can you please test as well?

he29-net (approved):
I finally finished going through the code, and apart from a few details (3x `while(!object.isNull())` missing a space after `while`, and a typo on line 1826 "because they container a track container"), the only small issue I see are the names of helper functions. `affected()`, `fixTrack()`, `fixAutomationPattern()` etc. would be now visible to all upgrade routines added in the future, since they are not just a local lambda like `affected()` was before.

Maybe they could be prefixed by `midiRange` or something. Or maybe it does not matter, because `DataFile.cpp` is not that complex so it won't cause much trouble. So I'm not insisting on fixing that; just wanted to point it out in case it was missed.

As for the TODO regarding float values on non-automated basenotes; it seems that the base note is loaded into an integer automatable model, so any decimal part would be discarded after a load from XML anyway? I also don't see how could a float make its way into the XML in the first place, if automation isn't involved. So I would say the current code is OK.

Other than that, I like how it is all structured. The upgrade routine got pretty complex, but separating the tasks into multiple functions and commenting on what's being done and why makes the process easy to follow and understand. :+1: I tested it with the supplied file about a week ago and I do not recall any issues, so as far as I'm concerned this PR should be good to go.

michaelgregorius:
Hi  @he29-net,

thanks for the review.!

I have fixed the `while`  statements and the typo with commit dac1f773470.

Commit 5335f6d8c62 then moves the whole upgrade code for the extended note range into its own class to hide all the helper methods. I have added the header file to the `src/core` folder because it is not part of the public interface and thus should not go into the "global" include directory.

Just as I am typing this I see that the mingw builds are now failing. 😅

src/core/UpgradeExtendedNoteRange.h
messmerd, line 1: This file should have gone under `include/` not `src/core/`.

michaelgregorius: Hi @messmerd,

yes, it seems like LMMS is currently missing a concept for these things.

If LMMS for example had a clean separation between the core and the GUI I would assume that only the core classes would be part of a public interface because they would be the most "interesting" and "open" ones to program again. Clients could for example use them to implement their own command line interfaces, reuse DSP algorithms, etc. I guess it would not be too helpful to have the interface of a very specialized GUI dialog in the include directory because what should clients really do with it?

Issue #6729, opened 2023-06-06
SID Filter image


anytizer:
Click on the blank area shown, the image appears.
Click again, and it vanishes.

![image](https://github.com/LMMS/lmms/assets/5563341/9a0057d6-49b1-49b7-9eff-3e807951a610)

plugins/SID/filter.png is missing.

PR #6731, opened 2023-06-08
Add SID filter image back (#6729)

michaelgregorius:
Add the image for the disabled SID filter back. According to a comment in the issue it seems to have gone missing during commit c1e6b313.


zonkmachine (approved):
Tested, works!

michaelgregorius:
Thanks for the check @zonkmachine! Merged,

PR #6738, opened 2023-06-14
Partially revert #6721, improve performance of Directory::addItems

### LINKED ISSUE #6734: sample search function broken on most recent nightly (g7c6ade573)
### Bug Summary
Searching in the samples tab only shows results in the currently selected folder and nothing else. No subfolders, and no other folders.
#### Steps to reproduce
Use the search function in the samples tab
#### Expected behavior
It shows results from everything in the samples folder
#### Actual behavior
Only shows results in the currently selected folder

#### Affected LMMS versions
1.3.0-alpha.1.302+g7c6ade573

LostRobotMusic:
#6721 was merged quite a bit faster than I expected, and unfortunately its incompatibility with the search bar wasn't noticed by anybody until I was messing with things the day after.  As an apology, I spent far too many hours on figuring out how I can optimize FileBrowser performance without lazy loading breaking the search bar.

This PR fixes that issue, and includes a full rewrite of Directory::addItems to massively increase its performance.  On my system, LMMS can load in over 10,000 directories with one file each in 16 seconds with the old (pre-#6721) behavior, and 4 seconds with this PR applied, so you can expect it to run around 4x as fast on average.  This will improve both FileBrowser reload times and LMMS opening times.

Fixes #6734.

### DIFF
--- a/src/gui/FileBrowser.cpp
+++ b/src/gui/FileBrowser.cpp
@@ -235,16 +235,19 @@ void FileBrowser::expandItems( QTreeWidgetItem * item, QList<QString> expandedDi
 	for (int i = 0; i < numChildren; ++i)
 	{
 		QTreeWidgetItem * it = item ? item->child( i ) : m_fileBrowserTreeWidget->topLevelItem(i);
-		auto d = dynamic_cast<Directory*>(it);
+		Directory *d = dynamic_cast<Directory *> ( it );
 		if (d)
 		{
-			d->update();
-			bool expand = expandedDirs.contains( d->fullName() );
-			d->setExpanded( expand );
-		}
-		if (m_recurse && it->childCount())
-		{
-			expandItems(it, expandedDirs);
+			if (m_recurse)
+			{
+				d->setExpanded(true);
+			}
+			bool expand = expandedDirs.contains(d->fullName());
+			d->setExpanded(expand);
+			if (m_recurse && it->childCount())
+			{
+				expandItems(it, expandedDirs);
+			}
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


Veratil (changes requested):
Looks great! A little bit of cleanup can be done still. 😄

src/gui/FileBrowser.cpp
```diff
@@ -1002,88 +1005,51 @@ void Directory::update()
...
+    treeWidget()->setUpdatesEnabled(false);
+
+    QStringList directories;
+    QStringList files;
+
+    QFileInfoList entries = thisDir.entryInfoList(QDir::Dirs | QDir::Files | QDir::NoDotAndDotDot);
+    for (const QFileInfo& entry : entries)
+    {
+        QString fileName = entry.fileName();
+        if (entry.isDir() && fileName[0] != '.')
```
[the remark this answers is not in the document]
LostRobotMusic: Oops, accidentally carried that over from the previous code, ha.

src/gui/FileBrowser.cpp
```diff
@@ -229,22 +229,21 @@ void FileBrowser::reloadTree()
...
 		{
-			d->update();
-			bool expand = expandedDirs.contains( d->fullName() );
-			d->setExpanded( expand );
-		}
-		if (m_recurse && it->childCount())
-		{
-			expandItems(it, expandedDirs);
+			if (m_recurse) { d->setExpanded(true); }
+			d->setExpanded(expandedDirs.contains(d->fullName()));
```
Veratil, line 242: I just realized this is doing the same thing, and it's possible that `m_recurse` is ignored in the case `expandedDirs.contains(d->fullName())` returns `false`.
Is this what we want to do here then?
```suggestion
			d->setExpanded(m_recurse || expandedDirs.contains(d->fullName()));
```

Veratil (changes requested):
One last thing I overlooked last time.

src/gui/FileBrowser.cpp
LostRobotMusic: Confusion over this exact area is actually exactly what caused the previous PR to break things.  Using "setExpanded(true)" and then "setExpanded(false)" immediately after actually does have an impact contrary to what you'd expect, since doing so goes through and caches all the files within that directory.  The change proposed here causes all directories to be expanded by default, since they aren't collapsed after being expanded anymore.

src/gui/FileBrowser.cpp
```diff
@@ -1002,88 +1005,51 @@ void Directory::update()
...
+        {
+            directories.append(fileName);
+        }
+        else if (entry.isFile() && fileName[0] != '.' && thisDir.match(m_filter, fileName.toLower()))
+        {
+            files.append(fileName);
+        }
+    }
+
+    directories.sort(Qt::CaseInsensitive);
```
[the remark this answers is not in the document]
messmerd: Sorting may have been important for getting the directories and files to appear in alphabetical order.

But if it actually isn't needed, then it should be possible to remove the `directories` (and `files`) variables altogether and just move the body of the directories for-loop into the entries for-loop above it (and the same thing for the files for-loop). That should make it simpler and more efficient.

And if sorting *is* actually needed, maybe the above strategy could be followed, and we could just pass the `QDir::Name` sort flag in the call to `thisDir.entryInfoList(...)`.

src/gui/FileBrowser.cpp
Veratil: Gotcha! All good then.

Veratil (commented):
Building off the comment by @messmerd here: https://github.com/LMMS/lmms/pull/6738#discussion_r1228988898

We'll need to test if sorting is still needed, but we can go to a single loop at least. Looking at the Qt docs we can add sorting as the second argument in the event we do need it.

LostRobotMusic:
Added sorting back in since it's required in the case where the user reloads the tree after new files/folders have been added.

Veratil (approved):

messmerd (approved):

PR #6741, opened 2023-06-20
Cache track length for Song Editor performance

LostRobotMusic:
Here for the same reason as #6629, calculating the song length is responsible for nearly 100% of LMMS's Song Editor lag.  While that other PR focuses on not calculating the song length a trillion times every time a clip is moved (as we should), this PR focuses on speeding up the song length calculation itself.  This should result in speed increases for moving and deleting and copy/pasting clips in any way (as well as a slight speed increase for opening large project files).

The performance gain will be roughly proportional to the number of tracks you have, since moving/creating/deleting clips will only need to calculate the length of the current tracks instead of the entire song.  So if you have 50 tracks, this'll run around 50x as fast as it did before, plus however long it takes for LMMS to check all of the cached track lengths (almost no time at all).


src/core/Track.cpp
```diff
@@ -543,13 +546,13 @@ void Track::removeBar( const TimePos & pos )
 
 
 
-/*! \brief Return the length of the entire track in bars
+/*! \brief Calculate the length of the entire track in bars
```
Veratil, line 549: Don't forget to update the description:
> Then we return that in bars by dividing by the number of ticks per bar.

src/core/Track.cpp
```diff
@@ -561,13 +564,15 @@ bar_t Track::length() const
 		}
 
 		const tick_t cur = clip->endPosition();
-		if( cur > last )
+		if (cur > last)
```
Veratil, line 567: We can shorten the if to a single line here.

src/core/Track.cpp
```diff
@@ -66,7 +66,8 @@ Track::Track( TrackTypes type, TrackContainer * tc ) :
 	m_simpleSerializingMode( false ),
 	m_clips(),        /*!< The clips (segments) */
 	m_color( 0, 0, 0 ),
-	m_hasColor( false )
+	m_hasColor( false ),
```
Veratil, line 69:
```suggestion
	m_hasColor(false),
```

src/core/Track.cpp
```diff
@@ -561,13 +564,15 @@ bar_t Track::length() const
...
+		if (cur > last)
 		{
 			last = cur;
 		}
 	}
 
-	return last / TimePos::ticksPerBar();
+	m_length = last / TimePos::ticksPerBar();
+	
+	if (Engine::getSong()) { Engine::getSong()->updateLength(); };
```
Veratil, line 575: I don't think we need the `if` here? We should always have a Song loaded I believe.

Veratil (changes requested):
LGTM so far. Just a few changes and a question.

src/core/Track.cpp
LostRobotMusic: I think you're right, I just got a bit nervous because `Track::removeClip` checks for a song for some reason.

Veratil: Yeah I'm not sure why. Maybe if doing a command line render? But even then we should have a song. Maybe @PhysSong has an idea?

PhysSong: > `Track::removeClip` checks for a song for some reason

The check was introduced in 2010 by commit 488337f6eb7d4c52d2d143d7fa23b5a3cca92f9f, but I have no ideas why this has been added.

Veratil: Alright, good to know that it's unknown. 😄

src/core/Clip.cpp
```diff
@@ -116,7 +116,7 @@ void Clip::movePosition( const TimePos & pos )
 void Clip::changeLength( const TimePos & length )
 {
 	m_length = length;
-	Engine::getSong()->updateLength();
+	if (getTrack()) { getTrack()->updateLength(); }
```
Veratil, line 119: I'm tempted to say let's remove this `if` and just call directly. We shouldn't ever have an instance of a clip without a track that I can think of.

LostRobotMusic: That's what I thought, but placing a note inside of a new clip consistently crashed LMMS, so it's clearly needed.

PhysSong: Do you have a backtrace for the crash? I can try myself if you don't.

Veratil: Ah, let's make a note of that in a comment then.
```cpp
// adding a note to a new clip will crash LMMS without this if check
// TODO: periodically check if the if is needed (a clip should always be connected to a track)
```

Veratil: Or @PhysSong will find the problem in a bit 😉
It's probably something creating a clip without a track somewhere.

LostRobotMusic: The traceback leads back to the Track's updateLength function, at `for (const auto& clip : m_clips)`.  There are several other parts in Clip.cpp that check for the Track's existence before doing things already.

Veratil (approved):

sakertooth (approved):

src/core/Track.cpp
LostRobotMusic: Found it.  LMMS segfaults when you close the program if you don't check for the Song here.  Turns out it is possible for these functions to run without an active Song, when LMMS is closing itself.

src/core/Track.cpp
```diff
@@ -551,11 +551,13 @@ void Track::removeBar( const TimePos & pos )
  */
 void Track::updateLength()
 {
+	if (!Engine::getSong()) { return; }
```
PhysSong, line 554: Won't this result in not updating `m_length` at all? Is this acceptable?

src/core/Track.cpp
PhysSong: > The check was introduced in 2010 by commit [488337f](https://github.com/LMMS/lmms/commit/488337f6eb7d4c52d2d143d7fa23b5a3cca92f9f), but I have no ideas why this has been added.

I now understand why it's required. The `deleteHelper` function sets `Engine::s_song` to `nullptr` before deleting the instance, so the check is required.
The intention of `deleteHelper` is to detect use-after-free, but it also prevents getting the instance being deleted via `Engine`, which may lead to crashes.

src/core/Track.cpp
LostRobotMusic: The song length is only used for determining the length of the horizontal scrollbar, and for determining the end of the song when exporting without loop markers.  So this shouldn't cause any problems whatsoever, unless it's somehow possible to export songs without there being an active Song.  I'm not sure how LMMS works in the command line, but I assume it still creates a Song when doing its magic stuff, right?

PhysSong: > I'm not sure how LMMS works in the command line, but I assume it still creates a Song when doing its magic stuff, right?

`main` calls `Engine::init` when exporting a projects from CLI, so yes.

PhysSong: >  So this shouldn't cause any problems whatsoever, unless it's somehow possible to export songs without there being an active Song.

True, but it's always good to keep the state of an object consistent.

Spekular:
LGTM aside from PhysSong's comment on `Track::updateLength()`, I agree that we should keep the track length accurate even if there's no song.

Spekular (approved):
LGTM!

LostRobotMusic:
This PR doesn't have the proper song length for exporting (it includes muted clips at the end), so merging should be held off for now.

PhysSong:
How about caching the length for normal uses(i.e., not for exporting) and then manually calculating the length for export?

michaelgregorius:
I did some measurements to evaluate the performance of this PR in the context of #5236 and had to merge master to do so. I thought that I can then as well push this work if it is already done.

Rossmaxx:
> How about caching the length for normal uses(i.e., not for exporting) and then manually calculating the length for export?

I think this is the only thing left to do.

PR #6742, opened 2023-06-22
Remove redundant nameChanged signal from InstrumentTrack

Veratil:
Noticed a bunch of these messages:
```
QMetaObject::indexOfSignal: signal nameChanged() from lmms::Track redefined in lmms::InstrumentTrack
```

This removes the signal from `InstrumentTrack` since it gets it from inheritance (`InstrumentTrack : public Track` and `nameChanged` is in public scope).

Also, since `Track::setName` emits `nameChanged` this removes a second emit of the same signal.


superpaik:
Tested ok on win10/64b

superpaik (approved):

PR #6746, opened 2023-06-25
Recursively unmute channels when soloing in the mixer

superpaik:
Implements/solve an existing TO-DO in the code of the function MixerChannel::unmuteForSolo() (//TODO: Recursively activate every channel, this channel sends to).

The problem is that when a channel sends audio to other channels, leaving that channel on "solo" doesn't make much sense (for example, you have an specific channel for "High Percs" and it is send to a "Drums Bus Channel" than then goes to master)
It is also true when other channels send audio to the one soloed (for example, if you set a "Drums Bus Channel" that receives from all drums channels - Kick, High Perc, Low Perc, etc.-)

Now when a channel is soloed (not including the master channel), all sends to and receives from this channel are unmuted so the audio is correctly send to master. This allows complex mixer channel routing (BUS, SENDs, etc.) to work properly and increases the mixer's usability.


superpaik:
The PR can be tested with the attached project.

The FX lines are routing using BUS lines that group instruments and SEND lines that are used for shared Reverb, Delay, etc. 
In the file, this is the setup:

Kicks -> Drums Bus -> Track Bus -> Master
Percussions -> Drums Bus -> Track Bus -> Master
		  -> Plate Send -> Track Bus -> Master
Perc Top End -> Drums Bus -> Track Bus -> Master
		    -> Delay Send -> Track Bus -> Master
Bass -> Bass Bus -> Track Bus -> Master
SubBass -> Bass Bus -> Track Bus -> Master
FX -> FX Bus -> Track Bus -> Master
     -> Room Send -> Track Bus -> Master
     -> Delay Send -> Track Bus -> Master
Drum Bus -> Track Bus -> Master
Bass Bus -> Track Bus -> Master
FX Bus -> Track Bus -> Master
Track Bus -> Master

As you can see in the video, 
- when a "instrument" channel is soloed (like Kicks channel), the corresponding BUS is unmuted, and also the SEND channels it sends audio to. So you can listen the instrument with all the effects you added in the SENDS
- when a "Bus" channel is soloed (like Drums Bus channel), all the lines that send audio to it are unmuted, and also the SENDS those channels send audio to. So you can listen this group, with all the effects you added in the SENDS.

There is a small issue. As you can see in the video when a SEND channel is soloed, other SEND channels may be unmuted. And that is because I cannot differentiate between BUS and SEND channels and behave different when is one or another. I don't see how to overcome this issue, and if that can be a stopper to this PR.
Test-unmuteSolo

[Test-unmuteSolo.zip](https://github.com/LMMS/lmms/files/11936201/Test-unmuteSolo.zip)

sakertooth:
Hey @superpaik, I'm looking to get this merged soon if possible. Are you able to fix the merge conflicts? I'm also wondering if the code could be simplified a bit (not sure about this as of right now, but just a thought).

> There is a small issue. As you can see in the video when a SEND channel is soloed, other SEND channels may be unmuted. And that is because I cannot differentiate between BUS and SEND channels and behave different when is one or another. I don't see how to overcome this issue, and if that can be a stopper to this PR.

I don't think this is much of an issue. IMO, after we isolate the signal paths a certain mixer channel can take to reach master, it should be fine as the user can mute whatever path they do not want after the fact.

superpaik:
Hi @sakertooth . I've been out of the project for a while now. Feel free to take it from here if you will.

michaelgregorius (approved):

sakertooth:
PR is good, just going to make small changes before merge.

sakertooth (approved):

